PR #542, opened 2020-09-02
SHOC property tests for lower level Assumed PDF functions

bogensch:
Adds property tests for all twelve lower level functions for the assumed_pdf routine in SHOC.  Note that none of these functions use the SHOCDataBase struct in the boilerplate code because they only need scalars for the inputs/outputs, unlike most other SHOC functions.

This routine also cleans up the shoc.F90 code slightly for assumed_pdf routines, because a few variables were being passed as output when they need only be local variables.


bogensch:
setting as WIP until Weaver error has been ameliorated.  @jgfouca any chance you can provide me with an expanded error log for the Weaver test?  Looks like the last 100 lines just gives me a bunch of P3 warnings and misses the error.  Thanks.

bogensch:
never mind, @jgfouca apparently I can't read.  I see the error

singhbalwinder (approved):
Looks good! Thanks Pete!

PR #543, opened 2020-09-03
improve database baseclass

jgfouca:
Apologies for the huge PR, recommend viewing with whitespace changes off.

The main goal was to make the PhysicsTestData base class API a bit better to reduce some of the work needed by subclasses. Also, to change p3_functions_f90 *Data structs to inherit PhysicsTestData. Also, updates the shoc_vertflux bfb test to make it a bit nicer. This should be the template for all future shoc bfb unit tests.


components/scream/src/physics/p3/p3_cloud_sed_impl.hpp
```diff
@@ -98,17 +98,17 @@ ::cloud_sedimentation(
...
             V_qc(pk).set(qc_gt_small, acn(pk)*tgamma(4 + bcn + mu_c(pk)) * dum / tgamma(mu_c(pk)+4));
             if (do_predict_nc) {
               V_nc(pk).set(qc_gt_small, acn(pk)*tgamma(1 + bcn + mu_c(pk)) * dum / tgamma(mu_c(pk)+1));
             }
-
-            const auto Co_max_local = max(qc_gt_small, -1,
-                                          V_qc(pk) * dt_left * inv_dz(pk));
-            if (Co_max_local > lmax)
-              lmax = Co_max_local;
           }
```
jgfouca, line 106: This was an actual bug in the c++ implementation. I discovered by accident when using better random values in the bfb unit test of cloud_sed.

jgfouca (commented):
Annotations

jgfouca:
Since this involved a lot of changes to memory management. I ran lots of tests with valgrind on, in addition to the standard tests.

This PR will likely conflict with a lot of ongoing development and should have high priority so that people are using the new coding style as soon as possible.

components/scream/src/physics/p3/p3_cloud_sed_impl.hpp
jeff-cohere: yikes!

components/scream/src/physics/p3/p3_functions_f90.hpp
```diff
@@ -479,34 +480,34 @@ void get_rain_dsd2_f(Real qr, Real* nr, Real* mu_r, Real* lamr, Real* cdistr, Re
...
 
-  CalcUpwindData(Int kts_, Int kte_, Int kdir_, Int kbot_, Int k_qxtop_, Int num_arrays_, Real dt_sub_,
-                 std::pair<Real, Real> rho_range, std::pair<Real, Real> inv_dz_range,
-                 std::pair<Real, Real> vs_range, std::pair<Real, Real> qnx_range);
+  CalcUpwindData(Int kts_, Int kte_, Int kdir_, Int kbot_, Int k_qxtop_, Int num_arrays_, Real dt_sub_);
 
   // deep copy
   CalcUpwindData(const CalcUpwindData& rhs);
 
+  CalcUpwindData& operator=(const CalcUpwindData& rhs) = delete;
```
jeff-cohere, line 502: Is it possible to delete the assignment operator in the base class and have subclasses follow suit? Is there any reason we wouldn't want to do this?

jeff-cohere (approved):
Nice work, Jim! The test code looks much less noisy. The random number generation stuff is awesome. I think it's much easier to understand.

bartgol (approved):
Looks good. I have a few comments, but feel free to pick and choose which (if any) are worth addressing.

xyuan (commented):
looks good to me

xyuan (approved):

components/scream/src/physics/p3/tests/p3_check_values_unit_tests.cpp
```diff
@@ -23,21 +23,20 @@ struct UnitWrap::UnitTest<D>::TestCheckValues {
...
-    std::make_pair( 1.000E+02, 5.000E+02), // temp
-  };
-
   CheckValuesData cvd_fortran[] = {
-    // kts_, kte_, timestepcount_, source_ind_, force_abort_, ranges
-    CheckValuesData(1,  72,   2,   100,   false,  ranges),
-    CheckValuesData(1,  72,   3,   100,   false,  ranges),
-    CheckValuesData(1,  72,   4,   100,   false,  ranges),
-    CheckValuesData(1,  72,   5,   100,   false,  ranges),
+    //          kts_, kte_, timestepcount_, source_ind_, force_abort_, ranges
```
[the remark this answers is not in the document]
jgfouca: Thx, left it in there on accident.

components/scream/src/physics/share/CMakeLists.txt
```diff
@@ -10,6 +10,7 @@ set(PHYSICS_SHARE_SRCS
   physics_share_f2c.F90
   physics_share.cpp
   physics_only_grids_manager.cpp
+  physics_test_data.cpp
```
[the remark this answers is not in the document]
jgfouca: It's needed for testing and cxx->f90 bridging. I think it would be pretty complex to try to build it selectively.

components/scream/src/physics/share/physics_test_data.cpp
```diff
@@ -0,0 +1,141 @@
...
+  m_totali(rhs.m_totali),
+  m_ptrs(rhs.m_ptrs.size()),
+  m_ptrs_i(rhs.m_ptrs_i.size()),
+  m_ptrs_c(rhs.m_ptrs_c.size()),
+  m_indices_c(rhs.m_indices_c.size()),
+  m_data(rhs.m_data),
+  m_idx_data(rhs.m_idx_data)
+{
+  const size_t expected_real_ptrs = m_ptrs.size() + m_ptrs_i.size() + m_ptrs_c.size();
+  EKAT_REQUIRE_MSG(real_ptrs.size() == expected_real_ptrs, real_ptrs.size() << " != " << expected_real_ptrs);
```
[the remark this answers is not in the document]
jgfouca: I will improve the beginning of the message. We may have to agree to disagree on `<<`. I personally love when assert macros support streams in the message since it is quite tedious in C++ to convert things to strings.

components/scream/src/physics/share/physics_test_data.cpp
```diff
@@ -0,0 +1,141 @@
...
+
+void PhysicsTestData::randomize(const std::vector<std::pair<void*, std::pair<Real, Real> > >& ranges)
+{
+  using range_type = std::remove_const<std::remove_reference<decltype(ranges)>::type >::type;
+
+  std::default_random_engine generator;
+
+  range_type ranges_copy(ranges);
+
+  for (auto& item : { std::make_pair(&m_ptrs, m_total) , std::make_pair(&m_ptrs_i, m_totali), std::make_pair(&m_ptrs_c, shcol) }) {
```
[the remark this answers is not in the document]
jgfouca: Thanks, that is nicer.

components/scream/src/physics/share/physics_test_data.hpp
```diff
@@ -0,0 +1,125 @@
...
+  SHOCGridData &operator=(const SHOCGridData &rhs) { PhysicsTestData::operator=(rhs); return *this; }
+};
+*/
+
+namespace scream {
+
+// Base class for common test data setup
+struct PhysicsTestData
+{
+  Int shcol, nlev, nlevi;
```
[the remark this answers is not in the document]
jgfouca: I think it's either "shared columns" or "shoc columns". It's the same concept as "ni" in P3 I think.

components/scream/src/physics/share/CMakeLists.txt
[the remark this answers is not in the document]
bartgol: Isn't cxx->f90 bridging only needed for tests?

components/scream/src/physics/share/physics_test_data.hpp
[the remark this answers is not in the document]
bartgol: If that's the case, then maybe we should change the name? Shared columns makes little sense to me (shared with whom, and in what sense?). And if it is "shoc_columns", then it's also not great, since it's in the common physics package (and what if we replace shoc with something else down the road?).

I would suggest something obvious, like `nlev`, which is the analogue of `nlev`. Uniform/consistent naming can help understand code better/faster, imho.

I understand that there are many files affected, but since this PR is already disruptive, a name change like that is probably best done here than in a later PR, so one only has conflicts once.

components/scream/src/physics/share/physics_test_data.cpp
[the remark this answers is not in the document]
bartgol: Do you think an EKAT macro like the following would be helpful? 
```
EKAT_REQUIRE_EQ (lhs, rhs, msg)
```
which checks lhs==rhs and prints msg if test fails (and then throws). And maybe similarly for <,<=,>,>=. That way we pass all the info needed withouth the need of streams. 

If so, I'll open a feature issue in EKAT.

Ps: it is quite handy when asserts support <<. My concern are that a) someone might think the return value of real_ptrs.size() supports <<, and b) if someone wraps the msg arg in the ekat macro in parentheses, we're in trouble. But these are just my fears. I won't argue with that being really handy.

components/scream/src/physics/p3/p3_cloud_sed_impl.hpp
ambrad: Was the bug that Co_max was < 0 if `!qc_gt_small.any()`? If so, that's b/c the team reduction returns `-DBL_MAX` (or `-FLT_MAX` for single precision) if the reduction variable isn't updated. In the micro-app, we handled this situation differently, demonstrated [here](https://github.com/E3SM-Project/scream-docs/blob/e83980e34c51b19a24d2209f7bd7785f3d483dfb/micro-apps/p3/p3_final_impl.hpp#L212). The idea is to save computation at every pk by instead just doing one test at the end for `Co_max < 0`.

ambrad: If that is indeed the bug, you could instead return the local test to the inside of the any() block and add a fast exit to generalized_sedimentation on Co_max < 0.

jgfouca:
Did upstream merge to resolve conflicts. Everyone (@xyuan , @singhbalwinder , @bogensch ), please no more porting PRs until this is in. 

@bartgol , @jeff-cohere , I know you already reviewed, but I had a bit a of a re-think (discussed with Luca offline) in order to make things even better. You should see f433ab9

components/scream/src/physics/p3/p3_cloud_sed_impl.hpp
jgfouca: @ambrad, the bug was that, when qc_gt_small.none(), it needed to skip the Co_max stuff, because that's what fortran does.

components/scream/src/physics/share/CMakeLists.txt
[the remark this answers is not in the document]
jgfouca: Yes, but unlike testing, the bridging code is not well-separated from the rest of the code. There would be link errors for all the *_c.f90 routines

components/scream/src/physics/share/physics_test_data.hpp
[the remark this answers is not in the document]
jgfouca: I agree. I'll do this in a subsequent PR.

jgfouca:
I'll just note, the code savings in existing code for this PR is a lot better than it appears by casual glance at the line counts for the PR. That's because I had to add a new ~350 of unit testing for PhysicsDataTest, so the actual savings is closer to 600 lines.

components/scream/src/physics/p3/p3_cloud_sed_impl.hpp
jgfouca: @ambrad , whoops, the exact opposite of what I just said. We need to always do the Co_max stuff, even when qc_gt_small.none()

ambrad: @jgfouca I agree. But that doesn't necessarily imply Co_max should be updated always. For future reference, I'm going to explain here as long as I looked at this again, but you needn't do anything in this PR.

I believe the bug you caught is that Co_max = -DBL_MAX when it is passed to generalized_sedimentation. That value is because Kokkos's team reduction ignores Co_max's initial value of 0 and rather initializes the reduction variable with -DBL_MAX. If !qc_get_small.any(), Co_max emerges as -DBL_MAX, when we want it to be 0.

Now, if Co_max is 0, then all the V_* velocities are 0. Then fluxdiv is everywhere 0 in the transport routine, so there is no change to the advected tracer, not even in a BFB sense, because the routine just adds and subtracts 0s.

In the micro-app -- line linked in my previous comment -- we handled the situation more efficiently. If Co_max < 0 on exit of the loop, we recognized this is because the loop never did anything; it has the value initialized by Kokkos. In the micro-app, we then broke from the overall advection loop because once a tracer is everywhere below the 'small' threshold, it will remain everywhere fixed in space thereafter.

There is a related issue that persists, though the main bug was fixed a couple of months ago. There are _incld variants of the tracers now; for a tracer q, q <= q_incld. ktop and kbot are found based on q >= qsmall, but then the V code chunk runs based on q_incld >= qsmall, which can be true when q < qsmall. So the range [ktop,kbot] is potentially a subset of the range implied by q_incld >= qsmall.

bartgol:
@jgfouca I saw very few cases where `DATA_COPY_CONS` and `ASSIGN` are called with `num_args` (in the former) different from `num_scalars` (in the latter). And I think those cases could be adjusted to make them equal (you were simply not init-ing some vars in the ctor).

So if you make them equal, then you could maybe get rid of two macros, and fuse them in one, which creates both copy ctor and assignment. Like this:

```
COPY_CTOR_AND_ASSIGN_OP (name, num_args, ...) \
  name (const name& rhs) : name(REP##num_args()) {
    *this = rhs;
    assignment_impl(rhs);
  }
  name& operator=(const name& rhs) {
    ASS#num_args(__VA_ARGS__);
    return *this;
  }
```

It might make it easier for those who write the derived classes, since they only have to remember to add one macro...

bartgol:
One last thing: when possible, I would add some disambigueting prefix for macros, like `SCREAM_`. I might be overthinking this too much, but I am not 100% sure other headers we may directly/indirectly include won't have `ASSIGN` or `ASS0`, or `REP1`...

components/scream/src/physics/share/CMakeLists.txt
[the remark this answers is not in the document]
bartgol: Sounds like a good topic for a future PR then.

bogensch (approved):

jeff-cohere:
> Did upstream merge to resolve conflicts. Everyone (@xyuan , @singhbalwinder , @bogensch ), please no more porting PRs until this is in.
> 
> @bartgol , @jeff-cohere , I know you already reviewed, but I had a bit a of a re-think (discussed with Luca offline) in order to make things even better. You should see [f433ab9](https://github.com/E3SM-Project/scream/commit/f433ab9584aa48b6810b5f2ca6fb016d31f0c811)

Hey @jgfouca . That commit looks like it really cleans things up further. Do you have a 1-2 sentence summary of your rethink, just to make sure I understand what's going on?

jgfouca:
@jeff-cohere,

I was concerned about the opportunity for users to make mistakes, in the previous impl, in the Ctor and CopyCtor of their *Data structs. If they decided to change the data layout of the struct in such a way that required a reorder of member initialization in the Ctor , they would need to remember to change the order in the CopyCtor accordingly. If not, things would compile but data would be in the wrong members, which would be a difficult thing to debug.

The key idea is to leverage the default cons and assignment in the CopyCtor so that you only need to specify members/layout in one place (the Ctor).

components/scream/src/physics/share/physics_test_data.hpp
```diff
@@ -0,0 +1,157 @@
...
+      {&zi_grid, &dz_zi}) {}               // list of (shcol x nlevi) members
+
+  DATA_COPY_CONS(SHOCGridData, 3); // 3 => number of arguments constructor expects
+  ASSIGN0(SHOCGridData); // 0 => number of scalars
+};
+*/
+
+// Convenience macros for up to 10 arguments, beyond that, you're on your own :)
+
+#define REP0()
```
jeff-cohere, line 44: Do you expect the `ASS*` and `REP*` macros to be used outside this file, or only by the preprocessor in expanding the `ASSIGN` and `DATA_COPY_CONS` macros below? If you don't need them for anything else, maybe we should #undef them after we define those macros.

jeff-cohere (changes requested):
Hey @jgfouca ,

I really like how this is shaping up. I made a few minor suggestions having to do with the `ASSIGN` and `DATA_COPY_CONS` macros. I don't have anything else to add to what others have mentioned. My "request changes" is just for you to decide whether to adopt what I've suggested.

components/scream/src/physics/share/physics_test_data.hpp
jgfouca: That is a good thought. I tried that myself and discovered it doesn't work because those macros need to be available when the ASSIGN macro is used.

components/scream/src/physics/share/CMakeLists.txt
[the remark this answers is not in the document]
jgfouca: Yep. I made an issue.

components/scream/src/physics/share/physics_test_data.hpp
jeff-cohere: Curses! In that case, maybe Luca's comment about prepending `SCREAM_` or maybe `PTEST_` to these macros to make them more specific (as I think Luca might have suggested) might make this less of an issue. After all, their actual names only appear in this file.

components/scream/src/physics/share/physics_test_data.cpp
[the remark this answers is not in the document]
jgfouca: I think it's probably better to just allow stream operations and not have a big increase in the number of macros, but I don't care to much one way or the other.

components/scream/src/physics/p3/p3_cloud_sed_impl.hpp
jgfouca: Made an issue for this.

jgfouca:
Deleting comments that are fixed or resolved to make this PR more readable.

components/scream/src/physics/share/physics_test_data.hpp
bartgol: Interesting. I was assuming that macros were expanded at file parsing in the order they appear. So I thought that the REP* macros would be immediately expanded in the _definition_ of  `DATA_COPY_CONS`, and the following expansions of `DATA_COPY_CONS` would not include `REP*` anymore.

bartgol: Oh oh oh, I see why. The CPP does not know which REP* will be called when it finds the definition of `DATA_COPY_CONS`, so it can't replace `REP*` yet. Makes sense.

jgfouca:
> @jgfouca I saw very few cases where `DATA_COPY_CONS` and `ASSIGN` are called with `num_args` (in the former) different from `num_scalars` (in the latter). And I think those cases could be adjusted to make them equal (you were simply not init-ing some vars in the ctor).

@bartgol , I noticed that too. I think the connection is somewhat coincidental with how P3 was handling it's data, and not an actual conceptual coupling. The number of ctor arguments (dimensions + scalars) and the number of scalars will only align if the struct is storing every dim as a scalar, which it really doesn't need to since the base class is already storing it. P3 only needed to store all of it because it was handling nk as kts and kte.

Anyway, I'm going to hold off on this particular recommendation for now in the interests of time and keeping macro offerings limited. Unless you feel pretty strongly about this, in which case I can do it.

jgfouca:
I think I addressed all the feedback. I will merge this once tests are passing unless I see more comments from people.

jeff-cohere (approved):

components/scream/src/physics/share/physics_test_data.hpp
jgfouca: Yeah, I "namespaced" all the macros with "PTD_" (for PhysicsTestData).

bartgol:
> > @jgfouca I saw very few cases where `DATA_COPY_CONS` and `ASSIGN` are called with `num_args` (in the former) different from `num_scalars` (in the latter). And I think those cases could be adjusted to make them equal (you were simply not init-ing some vars in the ctor).
> 
> @bartgol , I noticed that too. I think the connection is somewhat coincidental with how P3 was handling it's data, and not an actual conceptual coupling. The number of ctor arguments (dimensions + scalars) and the number of scalars will only align if the struct is storing every dim as a scalar, which it really doesn't need to since the base class is already storing it. P3 only needed to store all of it because it was handling nk as kts and kte.
> 
> Anyway, I'm going to hold off on this particular recommendation for now in the interests of time and keeping macro offerings limited. Unless you feel pretty strongly about this, in which case I can do it.

Uhm, then I think I don't really understand what "dimensions" and "scalars" are in this context. But that's ok, I should probably learn more about physics data structures before commenting further. Feel free to go ahead and merge then.

PR #546, opened 2020-09-04
fixed bug in P3 ice effective diameter needed by RRTMGP code

hassanbeydoun:
The PR addresses the bug @PeterCaldwell found on the calculation of the ice effective diameter in p3_interface. The multiplication by the density ratio has been removed.

### DIFF
--- a/components/cam/src/physics/cam/micro_p3_interface.F90
+++ b/components/cam/src/physics/cam/micro_p3_interface.F90
@@ -1179,7 +1179,7 @@ end subroutine micro_p3_readnl
    !! Effective radius for cloud ice
    rei(:ncol,top_lev:) = rei(:ncol,top_lev:) * 1e6_rtype  ! Rescale rei to be in microns
    !! Effective diameter for cloud ice
-   dei(:ncol,top_lev:) = rei(:ncol,top_lev:) * rho_qi(:ncol,top_lev:)/rho_h2os * 2._rtype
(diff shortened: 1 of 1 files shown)


PeterCaldwell (approved):
Perfect, thanks!

PeterCaldwell:
@bogensch - this bugfix will affect cloud forcing and hence tuning.

PeterCaldwell:
I've been creeped out all week because I knew we integrated this PR with only changes to the F90 version of P3... Why didn't BFB tests fail? The answer is that the change we made here is just within scream_p3_interface.F90 which isn't BFB tested. There is a components/scream/ version of scream_p3_interface, but it is just the ancient/crude coupling of F90 P3 into the AD and it never bothered with computing radiative properties. So there's nothing to be done here, but this is a reminder that our AD-P3 interface layer still needs to be done and we should worry about radiation calculations when we do make it. Pinging our AD man @AaronDonahue

AaronDonahue:
@PeterCaldwell, yes, good point.  Revamping our ad -> physics interfaces should be bumped up in priority.

PeterCaldwell:
> @PeterCaldwell, yes, good point. Revamping our ad -> physics interfaces should be bumped up in priority.

I definitely agree that AD->P3 interface is needed, but I don't think it is as urgent as IO, NH dynamics, and surface coupling (just because those tasks are already in-process and we don't need AD-P3 until we can run the model in realistic-ish settings).

PR #551, opened 2020-09-05
Property tests for SHOC third moment lowest level functions

bogensch:
Adds property tests for SHOC's third moment lowest level functions.

Note that this PR also renames the variable "isosqrt" (which implied square root of isotropy term) to "isosqrd" (which correctly implies isotropy term squared) in the shoc.F90 file.

### DIFF
--- a/components/scream/src/physics/shoc/shoc_functions_f90.hpp
+++ b/components/scream/src/physics/shoc/shoc_functions_f90.hpp
@@ -327,7 +327,8 @@ struct SHOCClipthirdmomsData : public PhysicsTestData {
   SHOCClipthirdmomsData(Int shcol_, Int nlevi_) :
     PhysicsTestData(shcol_, nlevi_,{&w_sec_zi, &w3}){}
 
-  SHOC_NO_SCALAR(SHOCClipthirdmomsData, 2);
+  PTD_STD_DEF(SHOCClipthirdmomsData, 2, 0);
+  PTD_DIM_RENAME(2, shcol, nlevi);
 
 };//SHOCClipthirdmomsData
 
--- a/components/scream/src/physics/shoc/shoc_functions_f90.hpp
+++ b/components/scream/src/physics/shoc/shoc_functions_f90.hpp
@@ -324,11 +324,10 @@ struct SHOCClipthirdmomsData : public PhysicsTestData {
   // In/out
   Real *w3;
 
-  SHOCClipthirdmomsData(Int shcol_, Int nlev_, Int nlevi_) :
-    PhysicsTestData(shcol_, nlev_, nlevi_,{},{&w_sec_zi, &w3}){}
+  SHOCClipthirdmomsData(Int shcol_, Int nlevi_) :
+    PhysicsTestData(shcol_, nlevi_,{&w_sec_zi, &w3}){}
 
-  PTD_DATA_COPY_CTOR(SHOCClipthirdmomsData, 3);
-  PTD_ASSIGN_OP(SHOCClipthirdmomsData, 0);
+  SHOC_NO_SCALAR(SHOCClipthirdmomsData, 2);
 
 };//SHOCClipthirdmomsData
 
--- a/components/scream/src/physics/shoc/shoc_functions_f90.hpp
+++ b/components/scream/src/physics/shoc/shoc_functions_f90.hpp
@@ -385,6 +385,22 @@ struct SHOCMixcheckData : public SHOCDataBase {
   { SHOCDataBase::operator=(rhs); return *this; }
 };//SHOCMixcheckData
 
+//Create data structure to hold data for clipping_diag_third_shoc_moments
+struct SHOCClipthirdmomsData : public SHOCDataBase {
+  // Inputs
+  Real *w_sec_zi;
+
+  // In/out
+  Real *w3;
+
+  SHOCClipthirdmomsData(Int shcol_, Int nlevi_) :
+    SHOCDataBase(shcol_, 0, nlevi_, {}, {&w_sec_zi, &w3}, {}) {}
+  SHOCClipthirdmomsData(const SHOCClipthirdmomsData &rhs) :
+    SHOCDataBase(rhs, {}, {&w_sec_zi, &w3}, {}) {}
+  SHOCClipthirdmomsData &operator=(const SHOCClipthirdmomsData &rhs)
+  { SHOCDataBase::operator=(rhs); return *this; }
+};//SHOCClipthirdmomsData
+
 struct SHOCSecondMomentSrfData : public SHOCDataBase {
   // Inputs
   Real *wthl, *uw, *vw;
@@ -453,6 +469,7 @@ void compute_conv_vel_shoc_length(SHOCConvvelData &d);
 void compute_conv_time_shoc_length(SHOCConvtimeData &d);
 void compute_shoc_mix_shoc_length(SHOCMixlengthData &d);
 void check_length_scale_shoc_length(SHOCMixcheckData &d);
+void clipping_diag_third_shoc_moments(SHOCClipthirdmomsData &d);
 void shoc_diag_second_moments_srf(SHOCSecondMomentSrfData& d);
 void linear_interp(SHOCLinearintData &d);
 void shoc_diag_second_moments_ubycond(SHOCSecondMomentUbycondData& d);
--- a/components/scream/src/physics/shoc/shoc_functions_f90.hpp
+++ b/components/scream/src/physics/shoc/shoc_functions_f90.hpp
@@ -401,6 +401,58 @@ struct SHOCClipthirdmomsData : public SHOCDataBase {
   { SHOCDataBase::operator=(rhs); return *this; }
 };//SHOCClipthirdmomsData
 
+struct SHOCAAdiagthirdmomsData
+{
+  // inputs
+  Real omega0, omega1, omega2, x0, x1, y0, y1;
+  
+  // outputs
+  Real aa0, aa1;
+
+};
+
+struct SHOCFtermdiagthirdmomsData
+{
+  // inputs
+  Real thedz, thedz2, bet2, iso, isosqrt, wthl_sec, wthl_sec_kc;
+  Real wthl_sec_kb, thl_sec, thl_sec_kc, thl_sec_kb, w_sec;
+  Real w_sec_kc, w_sec_zi, tke, tke_kc;
+  
+  // outputs
+  Real f0, f1, f2, f3, f4, f5;
+
+};
+
+struct SHOCOmegadiagthirdmomsData
+{
+  // inputs
+  Real buoy_sgs2, f3, f4;
+  
+  // outputs
+  Real omega0, omega1, omega2;
+
+};
+
+struct SHOCXYdiagthirdmomsData
+{
+  // inputs
+  Real buoy_sgs2, f0, f1, f2;
+  
+  // outputs
+  Real x0, y0, x1, y1;
+
+};
+
+struct SHOCFterminputthirdmomsData
+{
+  // inputs
+  Real dz_zi, dz_zt, dz_zt_kc, isotropy_zi, brunt_zi, thetal_zi;
+  
+  // outputs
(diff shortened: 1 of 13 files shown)


bartgol:
Unfortunately the blake jenkins job was stuck at the end without returning (you can see the line `OVERALL STATUS: PASS`). So I had to abort the build (which the AT interprets as a fail). I'm going to add the retest label, so this should be retested immediately (and should pass).

components/scream/src/physics/shoc/shoc_functions_f90.hpp
```diff
@@ -323,6 +323,84 @@ struct SHOCMixcheckData : public PhysicsTestData {
...
+//Create data structure to hold data for clipping_diag_third_shoc_moments
+struct SHOCClipthirdmomsData : public PhysicsTestData {
+  // Inputs
+  Real *w_sec_zi;
+
+  // In/out
+  Real *w3;
+
+  SHOCClipthirdmomsData(Int shcol_, Int nlev_, Int nlevi_) :
+    PhysicsTestData(shcol_, nlev_, nlevi_,{},{&w_sec_zi, &w3}){}
```
jgfouca, line 335: This doesn't look right. You just want shcol of w_sec and w3, right? Then you just need:
```
PhysicsTestData(shcol_, {&w_sec_zi, &w3}) {}
```

jgfouca (changes requested):
One minor change needed to take advantage of the improved Data base class.

components/scream/src/physics/shoc/shoc_functions_f90.hpp
bogensch: those arrays are actually of shape (shcol,nlevi).  I'm struggling how to do this correctly when there is data of (shcol,nlevi) but not data of (shcol,nlev).  The way I thought it should be done is:

SHOCClipthirdmomsData(Int shcol_, Int nlevi_):
  PhysicsTestData(shcol_, nlevi_, {&w_sec_zi, &w3}){}

and changing the 3 in PTD_DATA_COPY_CTOR to 2.  This compiles, but the test fails because nlevi is never assigned a value.  The only way I could get it to work was to assign a value for nlev (even though it's never used) and use the empty array.  Clearly I'm thinking about this wrong...

jgfouca: @bogensch , I see. You are not thinking about it wrong... I think I need to rethink the naming in PhysicsTestData.

jgfouca: @bogensch , once #570 is in, you can name the dimensions whatever you want.

bogensch: Thanks, i've merged now with master (I've pushed my changes if you want to look), but I'm still having trouble which is probably my fault.  Just to clarify should the correct syntax for this block now be:

  SHOCClipthirdmomsData(Int shcol_, Int nlevi_) :
    PhysicsTestData(shcol_, nlevi_,{&w_sec_zi, &w3}){}

  SHOC_NO_SCALAR(SHOCClipthirdmomsData, 2);
 
?  If so it complains that there is no d.nlevi() in the shoc_functions_f90.cpp file in:

void clipping_diag_third_shoc_moments(SHOCClipthirdmomsData &d) {
  shoc_init(d.nlev(), true);
  d.transpose<ekat::util::TransposeDirection::c2f>();
  clipping_diag_third_shoc_moments_c(d.nlevi(),d.shcol(),d.w_sec_zi,d.w3);
}

If I change d.nlevi() to d.nlev(), then I can get the test to compile/pass.  Does this seem right to you?

jgfouca: @bogensch , that makes sense. The SHOC_NO_SCALAR macro automatically renames the first dimension shcol, the second nlev, and the third nlevi since this is the common pattern in SHOC*Data structs. If you want a different naming scheme, put this in your struct instread of SHOC_NO_SCALAR:
```
  PTD_STD_DEF(SHOCClipthirdmomsData, 2, 0);
  PTD_DIM_RENAME(2, shcol, nlevi)
```

components/scream/src/physics/shoc/tests/shoc_aa_diag_third_moms_tests.cpp
```diff
@@ -0,0 +1,103 @@
...
+    
+    // Call the fortran implementation    
+    aa_terms_diag_third_shoc_moment(SDS); 
+    
+    // Check the result
+    REQUIRE(SDS.aa0 == aa0_test1a);
+    REQUIRE(SDS.aa1 < aa1_test1a);       
+    
+  }
+
```
jgfouca, line 81: All your new unit tests are not quite following the updated pattern. I expected to see an empty run_bfb function here:
```
static void run_bfb()
{
  // TODO
}
```

components/scream/src/physics/shoc/tests/shoc_aa_diag_third_moms_tests.cpp
```diff
@@ -0,0 +1,103 @@
...
+{
+  using TestStruct = scream::shoc::unit_test::UnitWrap::UnitTest<scream::DefaultDevice>::TestAAdiagThirdMoms;
+  
+  TestStruct::run_property();
+}
+
+TEST_CASE("shoc_aa_diag_third_moms_b4b", "shoc")
+{
+  using TestStruct = scream::shoc::unit_test::UnitWrap::UnitTest<scream::DefaultDevice>::TestAAdiagThirdMoms;
+
```
jgfouca, line 100: ... and a call to the empty run_bfb here. This will reduce warnings and make it easier for us to remember TODO items.

jgfouca (changes requested):
Empty run_bfbs needed for new unit tests, see inline for details.

components/scream/src/physics/shoc/shoc_functions_f90.hpp
bogensch: thanks!

components/scream/src/physics/shoc/tests/shoc_aa_diag_third_moms_tests.cpp
```diff
@@ -0,0 +1,108 @@
...
+namespace{
+
+TEST_CASE("shoc_aa_diag_third_moms_property", "shoc")
+{
+  using TestStruct = scream::shoc::unit_test::UnitWrap::UnitTest<scream::DefaultDevice>::TestAAdiagThirdMoms;
+  
+  TestStruct::run_property();
+}
+
+TEST_CASE("shoc_aa_diag_third_moms_b4b", "shoc")
```
jgfouca, line 102: Please use "bfb" instead of "b4b" everywhere.

components/scream/src/physics/shoc/tests/shoc_clip_third_moms_tests.cpp
```diff
@@ -0,0 +1,132 @@
...
+{
+  using TestStruct = scream::shoc::unit_test::UnitWrap::UnitTest<scream::DefaultDevice>::TestClipThirdMoms;
+  
+  TestStruct::run_property();
+}
+
+TEST_CASE("shoc_clip_third_moms_b4b", "shoc")
+{
+  using TestStruct = scream::shoc::unit_test::UnitWrap::UnitTest<scream::DefaultDevice>::TestClipThirdMoms;
+
```
jgfouca, line 129: Call TestStruct::run_bfb() here.

components/scream/src/physics/shoc/tests/shoc_aa_diag_third_moms_tests.cpp
```diff
@@ -0,0 +1,108 @@
...
+{
+  using TestStruct = scream::shoc::unit_test::UnitWrap::UnitTest<scream::DefaultDevice>::TestAAdiagThirdMoms;
+
+}
```
jgfouca, line 106: Call TestStruct::run_bfb() here.

components/scream/src/physics/shoc/tests/shoc_fterm_diag_third_moms_tests.cpp
```diff
@@ -0,0 +1,178 @@
...
+{
+  using TestStruct = scream::shoc::unit_test::UnitWrap::UnitTest<scream::DefaultDevice>::TestFtermdiagThirdMoms;
+  
+  TestStruct::run_property();
+}
+
+TEST_CASE("shoc_fterm_diag_third_moms_b4b", "shoc")
+{
+  using TestStruct = scream::shoc::unit_test::UnitWrap::UnitTest<scream::DefaultDevice>::TestFtermdiagThirdMoms;
+
```
jgfouca, line 175: Call TestStruct::run_bfb() here.

components/scream/src/physics/shoc/tests/shoc_fterm_input_third_moms_tests.cpp
```diff
@@ -0,0 +1,111 @@
...
+{
+  using TestStruct = scream::shoc::unit_test::UnitWrap::UnitTest<scream::DefaultDevice>::TestFtermInputThirdMoms;
+  
+  TestStruct::run_property();
+}
+
+TEST_CASE("shoc_fterm_input_third_moms_b4b", "shoc")
+{
+  using TestStruct = scream::shoc::unit_test::UnitWrap::UnitTest<scream::DefaultDevice>::TestFtermInputThirdMoms;
+
```
jgfouca, line 108: Call TestStruct::run_bfb() here.

components/scream/src/physics/shoc/tests/shoc_omega_diag_third_moms_tests.cpp
```diff
@@ -0,0 +1,105 @@
...
+{
+  using TestStruct = scream::shoc::unit_test::UnitWrap::UnitTest<scream::DefaultDevice>::TestOmegadiagThirdMoms;
+  
+  TestStruct::run_property();
+}
+
+TEST_CASE("shoc_omega_diag_third_moms_b4b", "shoc")
+{
+  using TestStruct = scream::shoc::unit_test::UnitWrap::UnitTest<scream::DefaultDevice>::TestOmegadiagThirdMoms;
+
```
jgfouca, line 102: Call TestStruct::run_bfb() here.

components/scream/src/physics/shoc/tests/shoc_w3_diag_third_moms_tests.cpp
```diff
@@ -0,0 +1,128 @@
...
+{
+  using TestStruct = scream::shoc::unit_test::UnitWrap::UnitTest<scream::DefaultDevice>::TestW3diagThirdMoms;
+  
+  TestStruct::run_property();
+}
+
+TEST_CASE("shoc_w3_diag_third_moms_b4b", "shoc")
+{
+  using TestStruct = scream::shoc::unit_test::UnitWrap::UnitTest<scream::DefaultDevice>::TestW3diagThirdMoms;
+
```
jgfouca, line 125: Call TestStruct::run_bfb() here.

components/scream/src/physics/shoc/tests/shoc_xy_diag_third_moms_tests.cpp
```diff
@@ -0,0 +1,111 @@
...
+{
+  using TestStruct = scream::shoc::unit_test::UnitWrap::UnitTest<scream::DefaultDevice>::TestXYdiagThirdMoms;
+  
+  TestStruct::run_property();
+}
+
+TEST_CASE("shoc_xy_diag_third_moms_b4b", "shoc")
+{
+  using TestStruct = scream::shoc::unit_test::UnitWrap::UnitTest<scream::DefaultDevice>::TestXYdiagThirdMoms;
+
```
jgfouca, line 108: Call TestStruct::run_bfb() here.

jgfouca (changes requested):
You forgot to call the empty run_bfbs; this is needed to remove the warnings. Also, lets be consistent and always use "bfb" instead of "b4b" to refer to "bit for bit".

jgfouca (approved):
Looks good! Thanks

PR #552, opened 2020-09-08
porting pblintd_cldcheck function

xyuan:
porting pblintd_cldcheck function

[BFB]

### DIFF
--- a/components/cam/src/physics/cam/shoc.F90
+++ b/components/cam/src/physics/cam/shoc.F90
@@ -1877,7 +1877,7 @@ end function w3_diag_third_shoc_moment
 
 subroutine clipping_diag_third_shoc_moments(&
            nlevi,shcol,w_sec_zi,& ! Input
-	   w3)                    ! Output
+	   w3)                    ! Input/Output
 
   ! perform clipping to prevent unrealistically large values from occuring
 
--- a/components/cam/src/physics/cam/shoc.F90
+++ b/components/cam/src/physics/cam/shoc.F90
@@ -4106,6 +4106,9 @@ subroutine pblintd_cldcheck(      &
                    shcol,nlev,nlevi, &                  ! Input
                    zi,cldn,          &                  ! Input
                    pblh)                                ! InOutput
+#ifdef SCREAM_CONFIG_IS_CMAKE
+    use shoc_iso_f, only: shoc_pblintd_cldcheck_f
+#endif
     !------------------------------Arguments--------------------------------
     ! Input arguments
     !
@@ -4122,8 +4125,14 @@ subroutine pblintd_cldcheck(      &
     !
     !---------------------------Local workspace-----------------------------
     !
-    integer  :: i                       ! longitude index
+    integer  :: i,k                       ! longitude index
     logical  :: cldcheck(shcol)      ! True=>if cloud in lowest layer
+#ifdef SCREAM_CONFIG_IS_CMAKE
+   if (use_cxx) then
+      call shoc_pblintd_cldcheck_f(shcol, nlev, nlevi, zi, cldn, pblh)
+      return
+   endif
+#endif
     !
     ! Final requirement on PBL heightis that it must be greater than the depth
     ! of the lowest model level if there is any cloud diagnosed in
--- a/components/cam/src/physics/cam/shoc.F90
+++ b/components/cam/src/physics/cam/shoc.F90
@@ -1647,7 +1647,7 @@ subroutine compute_diag_third_shoc_moment(&
   real(rtype) :: X0, Y0, X1, Y1, AA0, AA1
   real(rtype) :: zvar, x5var, iso, thedz, thedz2
   real(rtype) :: theterm, tsign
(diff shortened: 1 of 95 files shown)


components/scream/src/physics/shoc/shoc_pblintd_cldcheck_impl.hpp
```diff
@@ -0,0 +1,37 @@
...
+
+template<typename S, typename D>
+KOKKOS_FUNCTION
+void Functions<S,D>
+::shoc_pblintd_cldcheck(
+    const Scalar& zi, const Scalar& cldn, 
+    Scalar& pblh)
+{
+    //
+    // Final requirement on PBL heightis that it must be greater than the depth
```
jeff-cohere, line 19: heightis -> height is

components/scream/src/physics/shoc/shoc_pblintd_cldcheck_impl.hpp
```diff
@@ -0,0 +1,37 @@
...
+    // the lowest model level.  This is to deal with the inadequacies of the
+    // current "dry" formulation of the boundary layer, where this test is
+    // used to identify circumstances where there is marine stratus in the
+    // lowest level, and to provide a weak ventilation of the layer to avoid
+    // a pathology in the cloud scheme (locking in low-level stratiform cloud)
+    // If  any cloud is diagnosed in the lowest level, set pblh to 50 meters 
+    // higher than top interface of lowest level
+    //
+  
+    auto cldcheck = false;
```
jeff-cohere, line 30: It seems like we might want to do the parallel loop in this function and use masks instead of doing an if test on every element.

components/scream/src/physics/shoc/shoc_functions.hpp
```diff
@@ -86,6 +86,11 @@ struct Functions
     Scalar& thl_sec, Scalar& qw_sec, Scalar& wthl_sec, Scalar& wqw_sec,
     Scalar& qwthl_sec, Scalar& uw_sec, Scalar& vw_sec, Scalar& wtke_sec);
 
+  KOKKOS_FUNCTION
+  static void shoc_pblintd_cldcheck(
```
jeff-cohere, line 90: Right now, this function consists entirely of an if test applied to `cldn`, and a related assignment. We could pass it `uview_1d` arguments for the entire column and pull the loop inside this function, which would allow us to use a mask instead.

jeff-cohere (changes requested):
I think this function should operate at the column level and use a mask instead of cell-wise `if` tests.

components/scream/src/physics/shoc/shoc_functions.hpp
xyuan: ok, I will do it in that way though

xyuan: @jeff-cohere @ambrad In this way, we are going to endup repack the data to the same packsize in the inner loop, I was wondering whether this is efficient way to do, because the pblh is 1-dimensional data, but zi, and cldn are 2-dimensional.

components/scream/src/physics/shoc/shoc_pblintd_cldcheck_impl.hpp
ambrad: @jeff-cohere Just to be clear, this is on a single scalar, the nlev values.

components/scream/src/physics/shoc/shoc_functions.hpp
ambrad: @jeff-cohere see my comment above. Does this resolve your initial comment?

ambrad (approved):

components/scream/src/physics/shoc/shoc_functions.hpp
jeff-cohere: Right, yes. It only happens once per column. Sorry for the confusion!

jeff-cohere (approved):

components/cam/src/physics/cam/shoc.F90
```diff
@@ -4122,8 +4125,14 @@ subroutine pblintd_cldcheck(      &
     !
     !---------------------------Local workspace-----------------------------
     !
-    integer  :: i                       ! longitude index
+    integer  :: i,k                       ! longitude index
```
jgfouca, line 4128: I don't think this `k` is needed.

components/scream/src/physics/shoc/shoc_pblintd_cldcheck_impl.hpp
```diff
@@ -0,0 +1,37 @@
...
+    if (cldn >= sp(0.)) cldcheck = true;
```
jgfouca, line 31: Please use integer literals for 0 and 50. That lets you avoid sp.

jgfouca (changes requested):
Looks good. Some minor changes needed.

xyuan:
closing this PR, and redo it.

PR #553, opened 2020-09-08
port update_host_dse

tcclevenger:
Porting `update_host_dse()` to C++. and adding BFB test.


components/scream/src/physics/shoc/shoc_update_host_dse_impl.hpp
```diff
@@ -0,0 +1,40 @@
...
+  // Constants used
+  const auto lcond = C::LatVap;
+  const auto cp = C::CP;
+  const auto ggr = C::gravit;
+
+  Kokkos::parallel_for(Kokkos::TeamThreadRange(team, nlev_pack), [&] (const Int& k) {
+      auto range_pack = ekat::pack::range<IntSmallPack>(k*Spack::n);
+
+      Spack temp = (thlm(k)+(lcond/cp)*shoc_ql(k))/exner(k);
+      host_dse(k).set(range_pack < nlev, cp*temp+ggr*zt_grid(k)+phis);
```
ambrad, line 33: For simple slot-level operations, you don't need to mask with range_pack < nlev. Arithmetic in unused slots is fine. We just have to mask when doing things like reductions, BCs, shifted indices, and scans, although even in these cases not always.

ambrad (approved):
Approved subject to removal of mask and masked set.

jeff-cohere (approved):
Looks great!

tcclevenger:
@ambrad @jeff-cohere I finally have addressed the CPU test failing. Small issue: I needed a 1d Pack for a scalar input `phis` to my function `update_host_dse` (was using a `SmallPack`, which was fine for GPU but an issue on CPU).

ambrad:
This is for the host_to_device stuff, right, so just a bridge function issue? If so, that's fine.

tcclevenger:
> This is for the host_to_device stuff, right, so just a bridge function issue? If so, that's fine.

Correct

tcclevenger:
@ambrad Not sure if this needs to be reviewed again for the tester to rerun. I add the retest label yesterday.

ambrad (approved):

PR #554, opened 2020-09-09
P3 Bugfix: New rain evap impl

PeterCaldwell:
Replaces our MG2-like explicit-in-time rain evaporation parameterization with a WRF-like semi-analytic-integration-over-timestep version. This change seems to be critical for SCREAMv0 model stability.


PeterCaldwell:
Failing AD unit tests. My understanding is that this is due to a bug fixed in #556 .

bartgol:
@PeterCaldwell I don't know of any broken AD tests. The PR you link is changing something, but your branch should not be affected by that. All seems good on master.

I took a quick look at the p3 standalone test, and it fails because there is a (new?) input, "qv_prev", that is used but never registered. I believe there is a typo at line 73 of atmosphere_microphysics.cpp, where you declare the identifiers for the inputs: you declare the field as "q_prev", but later use it as "qv_prev". Since there is 1 place with "q_prev" and two with "qv_prev", democracy rules hints at the former being wrong?

Another remark/suggestion: our field names are treated as case-insensitive strings, so it's "safe" to use different cases for the same field. However, it might be misleading to other code developers/maintainers. I noticed you have "T_prev" and "t_prev". In the field repo, those are the same field (as I believe you intended), but it is probably best to be consistent, at least within the same file.

PeterCaldwell:
I'm going to stop working on this PR now and let it be integrated, but a follow-up PR will be needed because the initial timestep of an initial (rather than restart) run will get t_prev and qv_prev=0 in F90. @crterai had a fix in the F90 version which initialized these vals to a negative number, then checked each timestep for negative values and replaced them with current-step qv and T instead. I'm uncomfortable with doing that if/else every timestep and it would be awkward in C++ where we handle if statements by computing both branches and using masks over packs to choose which to use where. Also, the C++ initialization makes setting the first-step qv_prev and t_prev equal to beginning of step qv and T... this isn't exactly equal to the qv and T values seen by rain evap because of sequential-split timestepping, but should be close enough. Setting qv_prev and t_prev to beginning of step qv and T is difficult in the F90 because initialization routines don't have access to qv or T values. This discrepancy doesn't affect any tests because these initial values are set in the p3 interface layer which isn't BFB tested currently. In short - I'm not quite sure what to do here. Nothing? Try to pass qv and T to F90 p3_init?

There's also an issue about where t_prev and qv_prev should be set within each subsequent timestep. Right now we're following WRF-P3 by assigning _prev values at the end of microphysics but subtracting off the effect of the Bergeron process in evaporate_rain. I think the impact of *all* microphysical processes which don't handle vapor deposition would make more sense. I've been meaning to ask Hugh about this for months.

PeterCaldwell:
@JS-WRF-SBM - Just pinging you about our rain evap impl PR. I think it is BFB with what you're using but is easier to understand. Feel free to review it if you want (or not).

jgfouca (approved):
There's a lot here, but it looked good.

crterai (approved):
F90 side of things look good!

components/cam/src/physics/cam/bfb_math.inc
```diff
@@ -31,6 +32,7 @@
 #  define bfb_log(val) log(val)
 #  define bfb_log10(val) log10(val)
 #  define bfb_exp(val) exp(val)
+#  define bfb_expm1(val) (1.0 - exp(val))
```
PeterCaldwell, line 35: @ambrad and/or @jgfouca : expm1 doesn't seem to exist in fortran (at least on intel), so I'm defining it in the line above to be 1-exp(x) for E3SMv0 builds. This shouldn't affect scream tests even if this kludge implies a different algorithm because scream tests will always use the C++ version of expm1. My kludge will be less accurate in our SCREAMv0 runs when x is small, but I think that's ok because x=dt/tau_r in our case, which should never get too small for the applications we're currently using it for. Questions: 1). Does this hack seem ok to you? 2). I'm worried that the 1.0 here should be 1.0_rtype but that rtype isn't available here... which could cause casting to single precision and resulting loss of accuracy. Am I correct to worry about this? Thanks.

PeterCaldwell: Related point/reminder to self: going the expm1 route requires defining expm1 on packs in ekat. I'm waiting for confirmation that expm1 is good before issuing the 1-line ekat PR needed to get expm1 working in this PR.

ambrad: @PeterCaldwell but you should make it `(exp(val) - 1)`, i.e., opposite sign to what you have. Just use `1`, not `1.0`. `1` is exactly representable in FP, so no concerns here.

bartgol: Re: expm1 on packs. Yes, that'll need to be impl-ed for packs.

PeterCaldwell: @ambrad - doh. Thanks for the 1-exp(x) should be exp(x) -1 correction. I'm surprised you suggest using `1` instead of `1.0`: since this is fortran code, I thought type got cast to the crudest of the 2 inputs to '+' in fortran... You're probably right, I'm just sanity checking...

bartgol: the point is that, unlike, say, 1/3, 1 is exactly representable in binary format, regardless of precision, so it doesn't matter.

AaronDonahue:
@PeterCaldwell , not sure if this is related to all of the AT stuff but now this PR is looking to change 478(!) files.  I'm pretty sure that's a fluke.

PeterCaldwell:
Ok, failing as expected - p3_run_and_cmp fails in first step b/c not bfb, then in 2nd step because we added new variables the the list to compare vs what's on master. This causes an exception because the dimensions don't line up. Confirmed on Lassen and Quartz that using the new branch to create baselines fixes this 2nd issue.

PR #555, opened 2020-09-09
p3 variable rename for fortran and C++

xyuan:
p3 variable rename for fortran and C++

Naming guidelines and approach: https://acme-climate.atlassian.net/wiki/spaces/NGDNA/pages/1610647026/P3-C+Variable+Function+Renaming

[BFB]


xyuan:
Please review these names, some names was changed by me according to the rules in the table, not sure that is what we want.

jgfouca (approved):

jeff-cohere (approved):

components/scream/src/physics/p3/p3_inputs_initializer.cpp
```diff
@@ -61,7 +61,7 @@ void P3InputsInitializer::initialize_fields ()
 
   // Get host mirros' raw pointers
```
AaronDonahue, line 62: only a minor thing, not essential, but we could fix a typo in this comment "get host mirro**r**s' raw pointers".
Also, it looks like the white-space alignment below was meant to keep the "=" all aligned.  the rename of T_atm has shifted this.  Again not essential, but if there is an effort to keep the code clean and readable we may want to maintain that aesthetic.

components/scream/src/physics/p3/scream_p3_interface.hpp
```diff
@@ -13,7 +13,7 @@ extern "C"
 
 // Fortran routines to be called from C
 void p3_init_f90 ();
-void p3_standalone_init_f90 (Real* q, Real* T, Real* zi, Real* pmid, Real* dpres,
+void p3_standalone_init_f90 (Real* q, Real* t_atm, Real* zi, Real* pmid, Real* dpres,
```
AaronDonahue, line 16: In some cases I see T -> t_atm and sometimes T -> T_atm.  is there a distinction between the two options?  If not, should we commit to uppercase or lowercase "T" consistently throughout?

components/cam/src/physics/cam/micro_p3.F90
```diff
@@ -560,7 +560,7 @@ SUBROUTINE p3_main_part2(kts, kte, kbot, ktop, kdir, do_predict_nc, dt, inv_dt,
       if (qi(k).ge.qsmall) log_exitlevel = .false.
       !enddo
       if (log_exitlevel .and.                                                           &
-         (t(k).lt.zerodegc .and. qv_supersat_i(k).lt.-0.05_rtype)) goto 555   !i.e. skip all process rates
+         (t_atm(k).lt.T_zerodegc .and. qv_supersat_i(k).lt.-0.05_rtype)) goto 555   !i.e. skip all process rates
```
AaronDonahue, line 563: In these comparisons we have lowercase "t"_atm compared with uppercase "T"_zerodegc.  Does it make it more readable to stick with a single case "T" to represent anything related to temperature?

components/scream/src/physics/p3/p3_functions_f90.cpp
```diff
@@ -1812,11 +1812,11 @@ void calc_rime_density_f(
   Real local_rho_qm_cloud = *rho_qm_cloud_;
 
   Kokkos::parallel_for(1, KOKKOS_LAMBDA(const Int&) {
-      typename P3F::Spack t(t_), rhofaci(rhofaci_), table_val_qi_fallspd(table_val_qi_fallspd_), acn(acn_),
+      typename P3F::Spack T_atm(t_atm_), rhofaci(rhofaci_), table_val_qi_fallspd(table_val_qi_fallspd_), acn(acn_),
```
AaronDonahue, line 1815: Here's an example of the "T" case change being used back-to-back, but this isn't a pattern followed by other variables.

components/scream/src/physics/p3/p3_functions_f90.cpp
```diff
@@ -2901,7 +2904,7 @@ void p3_main_part2_f(
     mu_r_d              (temp_d[50]),
     lamr_d              (temp_d[51]),
     logn0r_d            (temp_d[52]),
-    cmeiout_d           (temp_d[53]),
+    qv2qi_depos_tend_d           (temp_d[53]),
```
AaronDonahue, line 2907: the white-space alignment is matched for `th_atm_d` on line 2878 but here it isn't...

AaronDonahue (commented):
The PR looks good, I had a series of questions about format that may be applicable to this PR.

components/scream/src/physics/p3/scream_p3_interface.hpp
xyuan: this is the argument variables, it use T_atm(t_atm), so I choose t_atm to differ it from T_atm inside the function.

components/cam/src/physics/cam/micro_p3.F90
PeterCaldwell: Yeah, we should use a consistent case. It would probably be best to change t_atm to T_atm everywhere. We might as well change th_atm to Th_atm as well for consistency.

components/scream/src/physics/p3/scream_p3_interface.hpp
PeterCaldwell: I thought we used "_" at the end where we had an argument we wanted to convert into a Pack later, so "T_atm(T_atm_)" would be the typical approach... but I'm confused why other arguments here don't do that. What does zi do to convert from argument to local Pack variable, for example? I'm confused because I don't see any argument assignment to packs in this function... I'm questioning your choice because I'd really like T_atm to be identical everywhere rather than switching cases for reasons that might be important but aren't immediately obvious...

components/scream/src/physics/p3/p3_functions_f90.cpp
PeterCaldwell: Yeah, let's use the underscore rather than the "T" case to indicate the local vs argument variable.

PeterCaldwell (changes requested):
Looks great subject to using a consistent case for T (like Aaron said). I'm amazed and impressed you managed to change all the T references to T_atm. I also really like that you put the link to the confluence page for this change in the initial comment of the PR. It was super handy for me!

xyuan:
> Looks great subject to using a consistent case for T (like Aaron said). I'm amazed and impressed you managed to change all the T references to T_atm. I also really like that you put the link to the confluence page for this change in the initial comment of the PR. It was super handy for me!

I committed a change from t_atm to T_atm for the temperature arguments variables according to your comment. @AaronDonahue @PeterCaldwell

AaronDonahue (approved):
Thanks for changing the t_atm to T_atm everywhere.  Looks good.

Just one question that can also be addressed in a future PR if we are just ready to get this one into the code, did we want to change the case on all `th_atm` as well?  I think Peter mentioned that.

xyuan:
@AaronDonahue  give me second, I wish to test it first, since there are a lot of rebased code. I will do it after the test finished.

xyuan:
@AaronDonahue @PeterCaldwell This PR is ready to merge now as the test passed

components/cam/src/physics/cam/micro_p3.F90
```diff
@@ -740,18 +740,18 @@ SUBROUTINE p3_main_part2(kts, kte, kbot, ktop, kdir, do_predict_nc, dt, inv_dt,
 
       call evaporate_rain(qr_incld(k),qc_incld(k),nr_incld(k),qi_incld(k), &
            cld_frac_l(k),cld_frac_r(k),qv(k),qv_prev(k),qv_sat_l(k),qv_sat_i(k), &
-           ab,abi,epsr,epsi_tot,t(k),t_prev(k),latent_heat_sublim(k),dqsdt,dt,&
+           ab,abi,epsr,epsi_tot,T_atm(k),t_prev(k),latent_heat_sublim(k),dqsdt,dt,&
            qr2qv_evap_tend,nr_evap_tend)
```
PeterCaldwell, line 744: T_atm should be t_atm above to keep case consistent?

components/cam/src/physics/cam/micro_p3.F90
```diff
@@ -2234,12 +2235,12 @@ subroutine ice_cldliq_wet_growth(rho,t,pres,rhofaci,    &
 
    real(rtype) :: qsat0, dum, dum1
 
-   if (qi_incld.ge.qsmall .and. qc_incld+qr_incld.ge.1.e-6_rtype .and. t.lt.zerodegc) then
-      qsat0=qv_sat( zerodegc,pres,0 )
+   if (qi_incld.ge.qsmall .and. qc_incld+qr_incld.ge.1.e-6_rtype .and. t_atm.lt.T_zerodegc) then
+      qsat0=qv_sat( T_zerodegc,pres,0 )
```
PeterCaldwell, line 2239: Also, can we change T_zerodegc to t_zerodegc so there's a consistent rule about capitalizing T?

PeterCaldwell: Also for T_rainfrz... maybe search + replace T_?

PeterCaldwell: Actually, nevermind. I see that the C++ uses uppercase T consistently for everything. Not worth worrying about the F90 since we will throw it away.

PeterCaldwell (approved):

PR #556, opened 2020-09-09
Upgrades to the grid class(es)

bartgol:
This PR improves a bit the AbstractGrid interface class, adding the possibility to get more info from the grid (e.g., the number of vertical levels). Additionally, it adds a SimpleGrid class, which makes testing of stand-alone parametrizations (or all physics, but no dynamics) a bit easier.

### DIFF
--- a/components/scream/data/scream_input.yaml
+++ b/components/scream/data/scream_input.yaml
@@ -16,4 +16,5 @@ Atmosphere Driver:
     Reference Grid: SE Physics
     Physics Only:
       Number of global columns: 1024
+      Number of vertical levels: 128
 ...
--- a/components/scream/data/scream_input.yaml
+++ b/components/scream/data/scream_input.yaml
@@ -6,14 +6,14 @@ Atmosphere Driver:
     Schedule Type: Sequential
     Process 0:
       Process Name: P3
-      Grid:         SE Physics
+      Grid:         Physics
       Can Initialize All Inputs: true
     Process 1:
       Process Name: SHOC
-      Grid:         SE Physics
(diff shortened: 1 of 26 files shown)


components/scream/data/scream_input.yaml
```diff
@@ -16,4 +16,5 @@ Atmosphere Driver:
     Reference Grid: SE Physics
```
AaronDonahue, line 16: Is SE Physics deprecated now as a reference grid?  I get the following error when unit testing:
```
  /p/lustre2/donahue5/E3SM_simulations/SCREAM/scream_development/
  rain_evap_tests/scream/components/scream/src/physics/share/
  physics_only_grids_manager.cpp:33: FAIL:
  grid_name=="Physics Only" || grid_name=="Physics"
  Error! Only 'Physics Only' (and, for convenience, 'Physics') grid supported
  for physics-only runs.
         Requested grid: SE Physics
```

which goes away if I delete SE.

AaronDonahue (changes requested):

PeterCaldwell:
My understanding from Aaron is that #554 depends on this fix, so I'd appreciate it if it was prioritized...

components/scream/data/scream_input.yaml
bartgol: The "Physics Only" grid manager does not produce a SEGrid, which is why "SE Physics" won't work. I will fix the ad input file.

AaronDonahue (approved):

AaronDonahue:
@bartgol , thanks for putting all of these changes in.

PR #496, opened 2020-08-06
Property tests for SHOC TKE lower level functions

bogensch:
Add property tests and boiler plate code for the lower level functions called by shoc_tke() function.  Specifically, for the following tests are added for these functions (more verbose descriptions of the tests found in the comments of the *.cpp test files):

1. integ_column_stability
    1. Stability test
    2. Dissipation test
2. compute_shr_prod
    1. Sheared profile test
    2. Unsheared profile test
3. adv_sgs_tke
    1. TKE decay/growth test
    2. Dissipation calculation test
4. isotropic_ts
    1. Stability test
    2. Dissipation test
5. eddy_diffusivities
    1. Boundary layer regime test
    2. Stable boundary layer test
    3. Unstable boundary layer test

PR #557, opened 2020-09-09
Property tests for SHOC TKE related upper level functions

bogensch:
Property test for SHOC upper level function call sgs_tke (of which all dependency function property tests have been completed in PR #496) and for function check_tke (which has no dependencies but is called directly from shoc_main).


jgfouca (changes requested):
I know this is WIP. Just a reminder to upstream merge/rebase and use the latest goodies and patterns.

jgfouca (approved):
Perfect!

PR #558, opened 2020-09-09
SHOC bug fix, dimensions of zi_grid corrected in length scale routine

bogensch:
Dimensions of zi_grid are corrected from (shcol,nlev) to (shcol,nlevi) in the upper level shoc_length routine.  This is a b4b fix on LC and Cori, presumably because the nlevi value of zi_grid is always zero in SCREAM and these compilers likely using zero when accessing this index, and thus expected to be a B4B PR


xyuan (approved):
good!

PR #570, opened 2020-09-10
Another upgrade/update to testing pattern

### LINKED ISSUE #549: Clean up dimension names in PhysicsTestData
The current names are shoc-specific ("shcol" for example). We should use the same approach as p3 to alias these dimension names (via a trivial member function) to a custom name if that's what the client wants.

jgfouca:
1) Allow flexible dimension renaming
2) Have generic, not SHOC-specific, dimension names in the PhysicsTestData base class
3) Add PTD_STD_DEF macro to reduce even more code duplication
4) Fix lots of warnings in shoc tests

Fixes #549

### DIFF
--- a/components/scream/src/physics/share/physics_test_data.hpp
+++ b/components/scream/src/physics/share/physics_test_data.hpp
@@ -29,13 +29,13 @@ struct SHOCGridData : public PhysicsTestData {
   // In/out
   Real *dz_zt, *dz_zi, *rho_zt;
 
-  SHOCGridData(Int shcol_, Int nlev_, Int nlevi_) :
-    PhysicsTestData(shcol_, nlev_, nlevi_,
-      {&zt_grid, &dz_zt, &pdel, &rho_zt},  // list of (shcol x nlev) members
-      {&zi_grid, &dz_zi}) {}               // list of (shcol x nlevi) members
+  SHOCGridData(Int dim1_, Int dim2_, Int dim3_) :
+    PhysicsTestData(dim1_, dim2_, dim3_,
+      {&zt_grid, &dz_zt, &pdel, &rho_zt},  // list of (dim1 x dim2) members
+      {&zi_grid, &dz_zi}) {}               // list of (dim1 x dim3) members
 
-  PTD_DATA_COPY_CTOR(SHOCGridData, 3); // 3 => number of arguments constructor expects
-  PTD_ASSIGN_OP(SHOCGridData, 0); // 0 => number of scalars
+  PTD_STD_DEF(SHOCGridData, 3, 0); // 3 => number of dimensions (1-3), 0 => number of scalars (0-10)
+  // If you have scalars, you'll have to add names after the count
 };
 */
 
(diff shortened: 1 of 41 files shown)


jgfouca:
For those doing conversions, please rebase/merge your branches once this PR is in (and please, no merges before this PR goes in).

jgfouca:
Recommend viewing with whitespace changes off. We still seem to be introducing a lot of trailing whitespaces in our work; we should try to avoid that. I have my editor set to highlight trailing whitespace in bright red, so I can advise on how to do this in emacs.

components/scream/src/physics/shoc/shoc_functions_f90.cpp
```diff
@@ -210,126 +210,126 @@ void shoc_energy_total_fixer(SHOCEnergytotData &d) {
...
   d.transpose<ekat::util::TransposeDirection::c2f>();
-  compute_conv_vel_shoc_length_c(d.nlev,d.shcol,d.pblh,d.zt_grid,
+  compute_conv_vel_shoc_length_c(d.nlev(),d.shcol(),d.pblh,d.zt_grid,
                                  d.dz_zt,d.thv,d.wthv_sec,d.conv_vel);
   d.transpose<ekat::util::TransposeDirection::f2c>();
 }
 
 void compute_conv_time_shoc_length(SHOCConvtimeData &d) {
-  shoc_init(d.nlev, true);
+  shoc_init(42, true); // fake nlev
```
jgfouca, line 299: SHOCConvtimeData doesn't have an nlev, so I assume I just pass whatever int I want?

bogensch: Yeah, in these instances (but apparently not this one!) I've been passing it "1" to sort of symbolize a single level.

bogensch (approved):

components/scream/src/physics/share/physics_test_data.hpp
```diff
@@ -72,25 +74,69 @@ struct SHOCGridData : public PhysicsTestData {
...
+#define PTD_ADD_3_3 6
+#define PTD_ADD_3_4 7
+#define PTD_ADD_3_5 8
+#define PTD_ADD_3_6 9
+#define PTD_ADD_3_7 10
+#define PTD_ADD_3_8 11
+#define PTD_ADD_3_9 12
+
+#define PTD_STD_DEF(name, dim, num_scalars, ...)        \
+  PTD_DATA_COPY_CTOR(name, PTD_ADD_##dim##_##num_scalars);  \
```
bartgol, line 118: I know I'm late to the party, but out of curiosity: why did you need the ADD_x_y macro? Didn't something like
```
PTD_DATA_COPY_CTOR(name, (dim + num_scalars) );
```
work?

bartgol: Nvm, I think I understand. The sum of the two needs to be used to append to other macro names, so it must be a simple numeral, not an expression. Got it.

PR #573, opened 2020-09-14
SHOC property test for upper level Length Scale function

bogensch:
Add property test for subroutine shoc_length, which serves as the upper level function for all of SHOC's length scale related routines.


jgfouca (approved):
Perfect!

PR #577, opened 2020-09-15
SHOC property tests for mid and upper level functions for Third Moments

bogensch:
Adds property tests for mid level function compute_diag_third_shoc_moments and upper level function diag_third_shoc_moments.  

Note that it was discovered that variable "shoc_mix" was being fed into the third moment routines but was never used.  Thus, this PR also eliminates unnecessarily passing this variable.

### DIFF
--- a/components/scream/src/physics/shoc/tests/shoc_diag_third_tests.cpp
+++ b/components/scream/src/physics/shoc/tests/shoc_diag_third_tests.cpp
@@ -20,31 +20,19 @@ namespace shoc {
 namespace unit_test {
 
 template <typename D>
-struct UnitWrap::UnitTest<D>::TestShocCompDiagThird {
+struct UnitWrap::UnitTest<D>::TestShocDiagThird {
 
   static void run_property()
   {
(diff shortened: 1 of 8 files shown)


components/scream/src/physics/shoc/tests/shoc_diag_third_tests.cpp
```diff
@@ -0,0 +1,244 @@
...
+    // Check to make sure there is at least one
+    //  positive w3 value for convective boundary layer
+    bool is_skew;
+    // Verify that boundary points are zero
+    for(Int s = 0; s < shcol; ++s) { 
+      is_skew = false; // Initialize
+      for(Int n = 0; n < nlevi; ++n) {
+        const auto offset = n + s * nlevi;
+
+	       // For this test make sure w3 has been clipped.
```
jgfouca, line 169: Some strange indentation going on here through line 212. Looks like some tabs are getting mixed-in too.

jgfouca (changes requested):

jgfouca (approved):

PR #578, opened 2020-09-16
[WIP]Ports adv_sgs_tke subroutine to C++/Kokkos

singhbalwinder:
This port builds Fortran->C++ bridge, prts the Fortran routine and adds a BFB test. I also created a SHOC constants file to store constants which are shared among SHOC routines.

*This PR also includes white space fixes, so it is better to turn them off while reviewing.

### DIFF
--- a/components/cam/src/physics/cam/shoc.F90
+++ b/components/cam/src/physics/cam/shoc.F90
@@ -2530,7 +2530,7 @@ subroutine shoc_assumed_pdf_compute_s(&
   real(rtype), intent(out) :: std_s
   real(rtype), intent(out) :: qn
   real(rtype), intent(out) :: C
-  
+
   ! local variables
   real(rtype) :: cthl, cqt
 
@@ -2834,6 +2834,11 @@ end subroutine compute_shr_prod
 subroutine adv_sgs_tke(nlev, shcol, dtime, shoc_mix, wthv_sec, &
      sterm_zt, tk, tke, a_diss)
 
+#ifdef SCREAM_CONFIG_IS_CMAKE
+  use shoc_iso_f, only: adv_sgs_tke_f
+#endif
+
+
   implicit none
 
   !intent -ins
@@ -2863,6 +2868,14 @@ subroutine adv_sgs_tke(nlev, shcol, dtime, shoc_mix, wthv_sec, &
   real(rtype) :: a_prod_bu, a_prod_sh
 
   real(rtype) :: Ck, Cs, Ce, Ce1, Ce2, Cee
+
+#ifdef SCREAM_CONFIG_IS_CMAKE
+  if (use_cxx) then
+     call adv_sgs_tke_f(nlev, shcol, dtime, shoc_mix, wthv_sec, &
+          sterm_zt, tk, tke, a_diss)
+  endif
+  return
+#endif
 
   Cs=0.15_rtype
   Ck=0.1_rtype
--- a/components/cam/src/physics/cam/shoc.F90
+++ b/components/cam/src/physics/cam/shoc.F90
@@ -2838,7 +2838,6 @@ subroutine adv_sgs_tke(nlev, shcol, dtime, shoc_mix, wthv_sec, &
   use shoc_iso_f, only: adv_sgs_tke_f
 #endif
 
-
   implicit none
 
   !intent -ins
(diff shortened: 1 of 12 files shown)


components/cam/src/physics/cam/shoc.F90
```diff
@@ -2864,9 +2868,17 @@ subroutine adv_sgs_tke(nlev, shcol, dtime, shoc_mix, wthv_sec, &
...
+  if (use_cxx) then
+     call adv_sgs_tke_f(nlev, shcol, dtime, shoc_mix, wthv_sec, &
+          sterm_zt, tk, tke, a_diss)
+     return
+  endif
+#endif
+
   Cs=0.15_rtype
   Ck=0.1_rtype
-  Ce=Ck**3/Cs**4
```
ambrad, line 2869: You should use `cube` and `pow<4>` for this, not `bfb_pow`. There's an important difference in the numerics of `pow` vs the others.

components/cam/src/physics/cam/shoc.F90
```diff
@@ -2885,7 +2897,7 @@ subroutine adv_sgs_tke(nlev, shcol, dtime, shoc_mix, wthv_sec, &
         a_prod_sh=tk(i,k)*sterm_zt(i,k)
 
         ! Dissipation term
-        a_diss(i,k)=Cee/shoc_mix(i,k)*tke(i,k)**1.5
+        a_diss(i,k)=Cee/shoc_mix(i,k)*bfb_pow(tke(i,k),1.5)
```
ambrad, line 2900: This hasn't been tested, but I think `sqrt(cube(.))` is the way to go here.

components/scream/src/physics/shoc/tests/shoc_pdf_compute_liqflux_tests.cpp
```diff
@@ -92,7 +92,7 @@ struct UnitWrap::UnitTest<D>::TestShocPdfCompLiqFlux {
     shoc_assumed_pdf_compute_liquid_water_flux(SDS);
 
     // Verify the result.  Liquid water flux should be positive
-    REQUIRE(SDS.wqls > 0);
+    //REQUIRE(SDS.wqls > 0); !BSINGH Temporarily commented out
```
singhbalwinder, line 95: I will remove this change, it was by accident.

components/cam/src/physics/cam/shoc.F90
ambrad: Btw, I don't know if anybody's impl'ed `pow<int>(.)` yet, but it's come up before and should be impl'ed. Can default to standard `pow` but then specialize for 2 (`x*x`), 3 (`x*x*x`), 4 (`pow<2>(pow<2>(x))`), with BFB wrappers for F90.

bartgol: We did an integer power in Hommexx, as part of our bfb utilities. At some point, it should be ported to Ekat.

jeff-cohere:
Sorry I haven't looked at this yet--I'm fighting fires elsewhere, and can look at this tomorrow.

components/scream/src/physics/shoc/shoc_adv_sgs_tke_impl.hpp
```diff
@@ -0,0 +1,79 @@
...
+    tke(k).set(tke(k)<0, 0);
+
+    // Compute buoyant production term
+    const auto a_prod_bu = (ggr/basetemp) * wthv_sec(k);
+
+    // Shear production term, use diffusivity from previous timestep
+    const auto a_prod_sh = tk(k) * sterm_zt(k);
+
+    // Dissipation term
+    a_diss(k) = Cee/shoc_mix(k)*pow(tke(k),sp(1.5));
```
jeff-cohere, line 55: Use `sqrt(cube(tke(k))` here as above.

jgfouca (approved):

components/cam/src/physics/cam/shoc.F90
jeff-cohere: Should we stick these into Ekat before this PR goes in? Or should we locally define macros for now? Or...?

singhbalwinder: I have been working on this PR and `pow` functions are the only one which are causing NBFB differences on GPU (as expected). Everything is working fine on the CPU side.
I am okay with waiting if that will be a cleaner solution and I think it will be. Other option is to use non-pow functions/Macros for now and switch to Ekat provided functions later.

xyuan (approved):

singhbalwinder:
Replaced by #764

PR #584, opened 2020-09-17
P3 bugfix: removed multiplication of cdist1 and cdistr by cloud and rain fraction

hassanbeydoun:
removed multiplication of cdist1 and cdistr by cloud and rain fraction respectively as suggested by @JS-WRF-SBM.  

Review uncovered the need to use in-cloud values in p3_main_part3. This change is implemented for cloud and rain water but doesn't make this change for ice phase yet, because doing so requires changing the p3_main_part3 interface. Removing multiplication by cloud fraction in cdist1 and cdistr also removes the need for cloud fraction to be passed to those functions. This will also be addressed in a follow-up PR.


JS-WRF-SBM (approved):
Yep, looks good.

PeterCaldwell:
get_cloud_dsd2 and get_rain_dsd2 are both called 2x. The first time is called with **in-cloud** values and the 2nd time (which is just used for diag_effc if I read correctly, see here: https://github.com/E3SM-Project/scream/blob/8abd7f8aedf4201857266bc864b294f2311028f8/components/cam/src/physics/cam/micro_p3.F90#L961 and here: https://github.com/E3SM-Project/scream/blob/8abd7f8aedf4201857266bc864b294f2311028f8/components/cam/src/physics/cam/micro_p3.F90#L975) is called with **cell-ave** values. I think this 2nd call should use in-cloud values because how do you fit a gamma distribution for cloud drop size where some fraction of the cell doesn't have clouds at all?

hassanbeydoun:
> get_cloud_dsd2 and get_rain_dsd2 are both called 2x. The first time is called with **in-cloud** values and the 2nd time (which is just used for diag_effc if I read correctly, see here:
> 
> https://github.com/E3SM-Project/scream/blob/8abd7f8aedf4201857266bc864b294f2311028f8/components/cam/src/physics/cam/micro_p3.F90#L961
> 
> and here:
> https://github.com/E3SM-Project/scream/blob/8abd7f8aedf4201857266bc864b294f2311028f8/components/cam/src/physics/cam/micro_p3.F90#L975
> 
> ) is called with **cell-ave** values. I think this 2nd call should use in-cloud values because how do you fit a gamma distribution for cloud drop size where some fraction of the cell doesn't have clouds at all?

I agree. all calls for the drop size distributions should use local particle concentration.

components/cam/src/physics/cam/micro_p3.F90
```diff
@@ -971,10 +978,14 @@ subroutine p3_main_part3(kts, kte, kbot, ktop, kdir, &
...
-
-         ze_rain(k) = nr(k)*(mu_r(k)+6._rtype)*(mu_r(k)+5._rtype)*(mu_r(k)+4._rtype)*           &
+         qr_incld = qr(k)/cld_frac_r(k)
+         nr_incld = nr(k)/cld_frac_r(k)
+         call get_rain_dsd2(qr_incld,nr_incld,mu_r(k),lamr(k),tmp1,tmp2,cld_frac_r(k))
+         nr(k) = nr_incld*cld_frac_r(k) !limiters might change nc_incld... enforcing consistency
+
+         !PMC potential bug: changed nr to nr_incld below. now ze is an in-cloud ze, while users
+         !may be assuming ze is a cell-averaged thing.
+         ze_rain(k) = nr_incld*(mu_r(k)+6._rtype)*(mu_r(k)+5._rtype)*(mu_r(k)+4._rtype)*           &
```
PeterCaldwell, line 988: What would users expect for ze (radar reflectivity)? Before this PR, stuff was a mishmash of in-cloud and cell-ave quantities. Now ze_rain is computed using all in-cloud values. Would people looking at output expect ze_rain to be cell ave? If so, can I just multiply it by the cloud fraction, or do I need to warp the drop size distribution in some way. @brhillman - do you have an opinion? @JS-WRF-SBM ? @crterai ? @hassanbeydoun ?

PeterCaldwell: Since I think ze_rain is just a diagnostic output, the easiest thing would be to just delete ze_rain (in which case this get_rain_dsd2 call may not be needed at all? Then whoever wants ze can implement what they want.

components/cam/src/physics/cam/micro_p3.F90
```diff
@@ -987,6 +998,7 @@ subroutine p3_main_part3(kts, kte, kbot, ktop, kdir, &
 
       ! ice:
 
+      !PMC potential bug: I think *all* of this qi stuff needs to be incld!!!
       call impose_max_total_Ni(ni(k),max_total_Ni,inv_rho(k))
```
PeterCaldwell, line 1002: @hassanbeydoun and @JS-WRF-SBM - I think all of this ice-phase stuff (from this comment to the end of p3_main_part3) should be in-cloud calculations. I haven't made the change because it requires passing cld_frac_i into p3_main_part3, which is a huge hassle in our C++ framework. Do you agree that this is needed? Kobby - did you already implement this in your version? My argument here is that qi distributions and resulting table lookups only make sense in an in-cloud format (and in fact, p3 and its table lookup was originally built to only be called when cloud fraction was 100%).

components/cam/src/physics/cam/micro_p3.F90
JS-WRF-SBM: I'm not following why after sedimentation where we update cell-avg values, you decided to go back to in-cld values here at the part3 routine.
The tables doesn't 'understand' what value are in the input, we just need to maintain consistency (and we are, before sedimentation)

Regarding Ze calc, I think that cell-avg is more coherent with what is actually being observed (volumetric scanning values), especially for high-res. But for low-res it might be less representable.
Also, if the riming and Ze_ice is cell-avg, Ze_rain should be the same (or the opposite way around).

PeterCaldwell: update regarding ze: I think the original implementation was probably correct but terribly confusing. It only makes sense to compute things involving parameters of our rain size distribution using in-rainy-area values because the rain size PDF doesn't exist/make sense outside of the cloudy region... so nr in the line above really should be nr_incld (like I changed it to be). But then we should convert the result to a cell ave value via ze_cell_ave = cld_frac_r*ze_incld + (1-cld_frac_r)*0. This ends up being ze = cld_frac_r*nc/cld_frac_r * (stuff above)... which is equal to what's written above. Seems like we need to add a clarifying comment here!!!

PeterCaldwell (approved):

PeterCaldwell:
setting to WIP again until AT is healthy. I expect this PR to pass.

PeterCaldwell:
I don't understand why this PR is passing p3_run_and_cmp... It should definitely be answer-changing???

AaronDonahue:
@hassanbeydoun , it looks like this PR might have a conflict to resolve before it can be merged.

PeterCaldwell:
@hassanbeydoun and @AaronDonahue - I resolved the merge conflicts.

PeterCaldwell:
I'm going to let the AT automerge this even though I still don't know why it's BFB because the PR content is innocuous and *not* changing answers is a pretty benign bug if it is a bug... if it is the right behavior, then merging it allows us to continue the DYAMOND campaign sooner.

PeterCaldwell (approved):

PeterCaldwell:
Only failure was on Blake: the mysterious p3_ut..._omp5 error which is a problem with master rather than with this PR.

PR #588, opened 2020-09-18
Both Shoc and P3 now call Murphy Koop

singhbalwinder:
Following is the summary of changes to make SHOC call Murphy Koop(MK) water vapor saturation scheme:
1. Created a new F90 file and moved `qv_sat` calculations from P3 to this new file
2. Modified SHOC code so that it always calls MK and removed "esatw_shoc" function from SHOC
3. On C++ side, moved some F90 files in CMakeList.txt from P3 to Share folder (so that both SHOC and P3 can use these files)
4. Moved "check temperature" calls from qv_sat to MK and polysvp1 functions (both C++ and F90)
5. Fixed qv_sat unit test to add additional argument of mask

### DIFF
--- a/components/cam/src/physics/cam/micro_p3.F90
+++ b/components/cam/src/physics/cam/micro_p3.F90
@@ -56,6 +56,8 @@ module micro_p3
        lookup_table_1a_dum1_c, &
        p3_qc_autocon_expon, p3_qc_accret_expon
 
+   use wv_sat_scream, only:qv_sat
+
   ! Bit-for-bit math functions.
 #ifdef SCREAM_CONFIG_IS_CMAKE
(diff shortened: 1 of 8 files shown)


bogensch (approved):
thanks for doing this.

PeterCaldwell (approved):
Looks good, thanks! What machines have you confirmed tests are BFB on? And have you confirmed that E3SM testing still works (e.g. that F cases still compile?)

singhbalwinder:
I have built and ran it on Compy and Summit (with CUDA). I have also run an F case (ne4 grid) on Compy and it compiles and runs fine. In my case, everything was BFB as my Fortran baselines were also using the modified code. This is going to be a non-BFB PR if we compare against the current master baselines (since SHOC calls MK now).

PR #593, opened 2020-09-20
porting pblintd_cldcheck function

xyuan:
redo the porting of pblintd_cldcheck function

https://github.com/E3SM-Project/scream/pull/552


components/scream/src/physics/shoc/shoc_functions_f90.cpp
```diff
@@ -579,6 +583,14 @@ void shoc_diag_second_moments_ubycond(SHOCSecondMomentUbycondData& d)
   d.transpose<ekat::util::TransposeDirection::f2c>();
 }
 
+void shoc_pblintd_cldcheck(SHOCPblintdCldCheckData& d)
+{
+  shoc_init(64, true);
```
jgfouca, line 588: Shouldn't init use nlev if it's available?

jgfouca (changes requested):
init needs fix, otherwise good

components/scream/src/physics/shoc/shoc_functions_f90.cpp
xyuan: yes, changes are made in the new commitment.

xyuan:
@jgfouca Hello, this case failed at the test, however, I am unable to reproduce this error on Summit, could you please get the full output of the test and send it to my email? I was unable to read the full information on github page. thanks a lot.

Issue #592, opened 2020-09-20
Inconsistancy in P3 rimed mass and volume 


JS-WRF-SBM:
Throughout the P3 code, the routine `calc_bulkRhoRime()` may update the `in-cld` values of rimed mass (`qrim`) and rimed volume (`brim`). For consistency, especially at the sedimentation section, the `cell-avg` values needs to be corrected as well, for example:
```
call calc_bulkRhoRime(qitot_incld(k),qirim_incld(k),birim_incld(k),rhop)
qirim(k) = qirim_incld(k)*icldm(k)
birim(k) = birim_incld(k)*icldm(k)
```

PeterCaldwell:
Thanks Koby! Addressed via #595

PR #595, opened 2020-09-21
P3 Bugfixes: p3_main_part3 ice should be incld and calc_bulkRhoRime should update cell-aves

PeterCaldwell:
This PR handles the interface-changing portion of #584 (ensuring that functions expecting in-cloud values aren't passed cell-aves) and addresses the issue noted in #592 where calc_bulkRhoRime changing in-cloud values but not the corresponding cell-aves, causing an inconsistency in model state. 

I'm just implementing the F90 changes so far. Corresponding C++ and bridging will be needed still (so this is a WIP PR).


xyuan:
@PeterCaldwell The C++ code is committed, it passed the test on Summit.

xyuan:
@PeterCaldwell The run_and_cmp failed in this PR, I think it is because of the baseline results, which are expected.

PeterCaldwell:
Thanks @xyuan . I noticed last night that p3_tests_ut_np1_omp5 failed for one of the cases, so I wanted the tests to rerun before integrating. It looks like this test failed again 3 hrs ago... but now I can't find the failure in last night's log files. Was I having a moment of prescience? Or am I just blind this morning? In any case, today's failure is on blake in full-debug mode. I'm unable to reproduce it on broadwell chipset. I figure we should run AT one more time to see if it is reproducible. Also, I think Blake is skylake - might be worth someone with access to skylake running this case with more verbose output to see if we can reproduce the fail.

xyuan:
This test results are available on the github website if you are unable to read it in your email. I will test it again on NERSC cori to see whether I can reproduce the error.

PeterCaldwell:
@xyuan - I think you mean I can look at https://github.com/E3SM-Project/scream/pull/595 to see the git log? That only shows the last 200 lines or so of the log. I'd love to see the full log file, but I think I need access to SNL machines to do so?

PeterCaldwell:
Passes all tests except p3_run_and_cmp, which it has small diffs for. This is the expected behavior.

PR #597, opened 2020-09-21
The AtmosphereProcess abstract class now maintains a timestamp.

### LINKED ISSUE #585: [AD] Timestamp helper functions
The current simulation timestamp information (i.e. total seconds since start-date) is stored by the AD.  Some processes (specifically Radiation and IO) will need access to this information.

Task: Create a set of helper functions to allow processes to access the timestamp information in various formats, such as seconds, hours, days, months, year, days in year.

jeff-cohere:
This change implements a commonly-used C++ pattern in which we override
protected methods (initialize_, run_, finalize_) that are wrapped by
public methods (initialize, run, finalize) that do the timestamp
arithmetic (and whatever else needs doing).

Closes #585

### DIFF
--- a/components/scream/src/share/atm_process/atmosphere_process.hpp
+++ b/components/scream/src/share/atm_process/atmosphere_process.hpp
@@ -81,7 +81,7 @@ public:
   //     that no other process is currently inside a call to 'run'.
   //   - the finalize method makes sure, if necessary, that all resources are freed.
   // NOTE: You don't override these methods. Override the protected methods
-  // NOTE: initialize_, run_, and finalize_ instead.
+  // NOTE: initialize_impl, run_impl, and finalize_impl instead.
   // The initialize/finalize method should be called just once per simulation (should
   // we enforce that? It depends on what resources we init/free, and how), while the
   // run method can (and usually will) be called multiple times.
@@ -89,15 +89,15 @@ public:
   // run/finalize is called.
   void initialize (const TimeStamp& t0) {
     t_ = t0;
-    initialize_(t_);
+    initialize_impl(t_);
   }
   void run        (const Real dt) {
     // Call the subclass's run method and update it afterward.
-    run_(dt);
+    run_impl(dt);
     t_ += dt;
   }
   void finalize   (/* what inputs? */) {
-    finalize_(/* what inputs? */);
+    finalize_impl(/* what inputs? */);
   }
 
   // These methods set fields in the atm process. Fields live on device and they are all 1d.
@@ -137,14 +137,14 @@ public:
 protected:
 
   // Override this method to initialize your subclass.
-  virtual void initialize_(const TimeStamp& t0) = 0;
+  virtual void initialize_impl(const TimeStamp& t0) = 0;
 
   // Override this method to define how your subclass runs forward one step
   // (of size dt). This method is called before the timestamp is updated.
-  virtual void run_(const Real dt) = 0;
+  virtual void run_impl(const Real dt) = 0;
 
   // Override this method to finalize your subclass.
-  virtual void finalize_(/* what inputs? */) = 0;
+  virtual void finalize_impl(/* what inputs? */) = 0;
 
   // This provides access to this process's timestamp.
   const TimeStamp& timestamp() const {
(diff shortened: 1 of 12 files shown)


components/scream/src/share/atm_process/atmosphere_process.hpp
```diff
@@ -80,14 +80,25 @@ class AtmosphereProcess : public ekat::util::enable_shared_from_this<AtmosphereP
...
   // we enforce that? It depends on what resources we init/free, and how), while the
   // run method can (and usually will) be called multiple times.
   // We should put asserts to verify that the process has been init-ed, when
   // run/finalize is called.
-  virtual void initialize (const TimeStamp& t0) = 0;
-  virtual void run        (const Real dt) = 0;
-  virtual void finalize   (/* what inputs? */) = 0;
+  void initialize (const TimeStamp& t0) {
+    t_ = t0;
+    initialize_(t_);
```
bartgol, line 92: I think `blah_` is very close to `blah`, and might be confusing. The field setter methods in this class use the `blah` vs `blah_impl` naming, so maybe we can stick with that convention?

bartgol (approved):
Looks good, thanks. I have a comment/suggestion for the init/run/final method names, but it might subjective. Feel free to push back.

jeff-cohere:
@bartgol , I made the changes you suggested. It builds and runs on my Linux machine, and I think it doesn't affect anything GPU-wise. How would you like to handle merging this? Should I just stick [WIP] into the title till the AT drama works itself out?

bartgol:
I think adding WIP was probably the best thing. Even if we're confident nothing should have broken on our target archs, it doesn't cost much to wait and see.

AaronDonahue:
@brhillman and @mfossitt , just a heads up that when this PR goes in there is a slight change the names of functions in the radiation and p3 interfaces that might impact your ongoing work.  Should be pretty straightforward to update your branch though.

AaronDonahue (approved):

bartgol (approved):

AaronDonahue:
Doe the `get_seconds` call return the total number of seconds in the simulation, or just the seconds portion of a yy-mm-dd-hh-mm-ss timestamp?

jeff-cohere:
> Doe the `get_seconds` call return the total number of seconds in the simulation, or just the seconds portion of a yy-mm-dd-hh-mm-ss timestamp?

It just returns the seconds field (ss) that you feed to it in the constructor.

AaronDonahue:
ahh ok, a little digging and I see that when you add `dt` to a timestamp it first adds to the seconds and then, if the total number of second `m_ss` is greater than 86400 then it will increase the day, month, year counter accordingly.  so `m_ss` is always bound by 86400 (# sec/day).

AaronDonahue:
maybe not in this PR, but the IO class would want a way to get total seconds from initialization.  It should be fairly easy to but in a helper function to get that information.  Or I can do that on the IO side.

jeff-cohere:
We could start by doing that on the I/O side and see whether anyone else makes use of it.

AaronDonahue:
sounds good

AaronDonahue:
@brhillman , while we are on the subject.  What format does radiation want the timestamp in?

jeff-cohere:
The test failures are odd--I wish we could get access to the full logs. I'm not sure whether the failures have to do with SNL's testing machine drama or whether they're real, since none of the changes have touched anything in P3 or SHOC (aside from their AtmosphereProcess subclasses).

bartgol:
The full log shows the same error reported in #600. PR #601 is already lined up to fix this. We might have to wait till the AT merges that one before anything else works. :/

bartgol:
> maybe not in this PR, but the IO class would want a way to get total seconds from initialization. It should be fairly easy to but in a helper function to get that information. Or I can do that on the IO side.

There is a utility function that can help you with that:

```
double operator- (const TimeStamp& ts, const TimeStamp& dt);
```

This allows you to do `double elapsed_sec = ts1 - ts2;`. Knowing the simulation initial timestamp and the current timestamp, you should be able to get what you need. I'm not sure we are really testing the TimeStamp operator overloads though...

bartgol (approved):

PR #598, opened 2020-09-22
[WIP]: Prescribed CCN option changes added into a new branch for PR

hassanbeydoun:
A new prescribed CCN option that reads CCN concentrations from a netcdf and uses the corresponding values to compute a cloud drop number concentration in p3 microphysics. A summary of the changes:
1. Added a new namelist parameter "do_prescribed_CCN"
2. monthly ccn variables mon_ccn_1 and mon_ccn_2 are initialized in p3_init. In p3_micro_tend, nccn_prescribed is calculated by interpolating between mon_ccn_1 and mon_ccn_2. At the start of each month, mon_ccn_1 and mon_ccn_2 are updated.
3. nccn_prescribed is fed into p3_main where it is used to compute nc. nc remains prognostic with this option. This option also bypasses reading in an ice nucleating particle concentration in the p3 ice_nucleation subroutine. 
4. The code assumes there are 12 prescribed CCN files representing each month of the year. The code looks for these files within the p3 microphysics lookup table directory to update mon_ccn_1 and mon_ccn_2.


components/cam/bld/namelist_files/namelist_definition.xml
```diff
@@ -2054,6 +2054,12 @@ Specify if cloud fraction is applied to microphysics tendencies.
 Default: TRUE
 </entry>
 
+<entry id="do_prescribed_CCN" type="logical" category="microphys"
+       group="micro_nl" valid_values="" >
+Specify if CCN concentration is read in from ncdf file within p3.
+Default: FALSE
```
AaronDonahue, line 2060: Here you say the default is "False" but it is actually true in `namelist_defaults_cam.xml`

run_scream.syrah_ne30_standard.csh=a_run_link
```diff
@@ -0,0 +1 @@
+/p/lustre2/beydoun1/syrah.default.test_prescribed_CCN_6.ne30_ne30
```
AaronDonahue, line 1: This link and the file `run_scream.syrah_ne30_standard.csh` comes from running the example sub script and shouldn't be in the PR.  I think you can just delete them.  Be sure to write `git rm` to remove them.

components/cam/src/physics/cam/micro_p3_interface.F90
```diff
@@ -699,6 +795,103 @@ subroutine get_cloud_fraction(its,ite,kts,kte,ast,qc,qr,qi,method, &
...
+
+         grid_id = cam_grid_id('physgrid')
+
+         call cam_grid_get_dim_names(grid_id, dim1name, dim2name)
+
+         call cam_pio_openfile(nccn_ncid,filename,PIO_NOWRITE)
+
+         allocate(ccn_values(pcols,pver,begchunk:endchunk))
+
+         call infld('CCN3',nccn_ncid,dim1name,'lev',dim2name,1,pcols,1,pver,begchunk,endchunk,&
```
AaronDonahue, line 869: Didn't we determine that using `lchnk` instead of begchunk and endchunk worked for `infld`?  I don't see that in this code.

components/cam/src/physics/cam/micro_p3_interface.F90
```diff
@@ -699,6 +795,103 @@ subroutine get_cloud_fraction(its,ite,kts,kte,ast,qc,qr,qi,method, &
...
 
+    subroutine get_prescribed_CCN(nccn_prescribed,micro_p3_lookup_dir,its,ite,kts,kte,pbuf,lchnk)
+
+       use pio,              only: file_desc_t,pio_nowrite
+       use cam_pio_utils,    only: cam_pio_openfile,cam_pio_closefile
+       use cam_grid_support, only: cam_grid_check, cam_grid_id
+       use cam_grid_support, only: cam_grid_get_dim_names
+       use ncdio_atm,        only: infld
+       use time_manager,     only: get_curr_date
+       use ppgrid,           only: begchunk, endchunk, pcols, pver, pverp, psubcols
```
AaronDonahue, line 806: I think as long as you include all of these ppgrid variables at the header of the module you don't have to include them in individual routines.

components/cam/src/physics/cam/micro_p3_interface.F90
```diff
@@ -623,6 +656,69 @@ subroutine micro_p3_init(pbuf2d)
       end if
    end if
 
+   if (do_prescribed_CCN) then !intialize mon_ccn_1 and mon_ccn_2
```
AaronDonahue, line 659: It seems like in the init step you simply repeat what you did whenever you call `get_prescribed_CCN` below.  Could you just call that routine instead of copying the same code here?

AaronDonahue (changes requested):
Nice work!  I have a few comments and a few changes, but overall it looks pretty good.

PeterCaldwell:
Since we're also changing ice nucleation with this PR now, maybe we should change "do_prescribed_CCN" to "do_SPA"?

PeterCaldwell:
Also, I added WIP to the title just because the autotester is going to be overburdened while it comes back online and we aren't ready to integrate this, so might as well not test it yet. Also, since this PR adds a new option which a). isn't exercised by default and b). is totally unknown to our C++ code, I think the autotester will notice any of these changes at all... it should pass.

components/cam/bld/namelist_files/namelist_defaults_cam.xml
```diff
@@ -1045,6 +1045,7 @@
 <micro_tend_output       	      > .true.  </micro_tend_output>
 <p3_qc_autoCon_Expon    		      >  2.47   </p3_qc_autoCon_Expon>
 <p3_qc_accret_Expon    		      >  1.15   </p3_qc_accret_Expon>
+<do_prescribed_CCN>                    > .true. </do_prescribed_CCN>
```
PeterCaldwell, line 1048: we should probably make the default behavior false for now so it doesn't change under people who aren't explicitly expecting the change.

components/cam/bld/namelist_files/namelist_definition.xml
PeterCaldwell: I just commented on this above! I say we leave this version as-is and change the previous mention to false.

components/cam/src/physics/cam/micro_p3.F90
```diff
@@ -405,10 +406,14 @@ SUBROUTINE p3_main_part1(kts, kte, kbot, ktop, kdir, do_predict_nc, dt, &
...
-          if (.not.(do_predict_nc)) then
+          if (.not.(do_predict_nc) .and. .not.(do_prescribed_CCN)) then
             nc(k) = nccnst*inv_rho(k)
           else
-            nc(k) = max(nc(k) + nc_nuceat_tend(k) * dt,0.0_rtype)
+            if (.not. (do_prescribed_CCN)) then
+               nc(k) = max(nc(k) + nc_nuceat_tend(k) * dt,0.0_rtype)
+            else
+               nc(k) = max(nc(k),nccn_prescribed(k))
+            endif
```
PeterCaldwell, line 416: less contorted logic here would be "if do_prescribed_CCN...else if do_predict_nc... else <do not do_predict_nc and not do_prescribed_ccn>"

components/cam/src/physics/cam/micro_p3_interface.F90
PeterCaldwell: Yeah, I like the idea of reusing get_prescribed_CCN for init. Less code to maintain.

components/cam/src/physics/cam/micro_p3_interface.F90
```diff
@@ -699,6 +795,103 @@ subroutine get_cloud_fraction(its,ite,kts,kte,ast,qc,qr,qi,method, &
...
+         deallocate(ccn_values)
+
+         current_month = month
+
+      end if
+
+      !interpolate between mon_ccn_1 and mon_ccn_1 to calculate nccn_prescribed
+      !based on current date
+
+      fraction_of_month = (day*3600.0*24.0 + tod)/(3600*24*30) !tod is in seconds
```
PeterCaldwell, line 885: I think you're assuming each month has 30 days here - which isn't true? I think it is safe to assume a 365 day calendar with no leap year (at least for SCREAMv0... Luca added leap-year support in the AD)... With a 365 day calendar, you can just create an array of days per month for each month of the year and replace 30 above with days_per_month(month)...

PeterCaldwell:
This PR looks great. As always, implementation is in practice much more complicated than I expected. 

One structural suggestion: I think you can avoid having to make changes to p3_main_part2, ice_nucleation, update_prognostic_liquid, etc. by:

1. enforcing do_predict_nc to be true when do_predict_CCN is true (by calling endrun if not true. Grep 'incompatible' in convect_shallow.F90 for examples)
2. using if do_predict_CCN... else if do_predict_nc (and possibly .not. do_predict_CCN if you're feeling explicit)... else... in p3_main_part1 (see my comment about contorted logic)

This approach would be more consistent with E3SM's past practices... I'm not sure whether we want 2 flags in our C++ impl, but for the F90 I think it would make more sense.

PeterCaldwell:
One other comment - we should be clear (in the PR description? In comments in the code?) that our input data needs to be already gridded to the grid we're using for our run.

PeterCaldwell:
Currently, this PR _reads in from a file_ a given month's climatological CCN concentration every time the simulated month changes. We could instead read all 12 months of data at initialization time. The benefit of the latter approach is that we don't need to keep reading data each month, which is slow(?). The drawback to the latter approach is that we're storing 10 more 3d fields all the time (ie there are 12 months and 2 months of data are needed at any one time for interpolating). Which is more computationally beneficial? @ambrad ? @jgfouca ?

ambrad:
Is it slow, though? If so, is it I/O bandwidth? If so, then seems like the two are equivalent, and reading all 12 mos at once would be painful for runs that are short or fail quickly.

components/cam/bld/namelist_files/namelist_defaults_cam.xml
hassanbeydoun: Changed to false by default.

components/cam/src/physics/cam/micro_p3.F90
hassanbeydoun: fixed.

components/cam/src/physics/cam/micro_p3_interface.F90
hassanbeydoun: Forgot to fix that! Fixed now.

components/cam/bld/namelist_files/namelist_definition.xml
hassanbeydoun: Fixed. Default is false now.

run_scream.syrah_ne30_standard.csh=a_run_link
hassanbeydoun: removed.

components/cam/src/physics/cam/micro_p3_interface.F90
hassanbeydoun: That didn't work unfortunately. I may play with it more in the future but at this point the approach here works so I suggest we stick to it.

components/cam/src/physics/cam/micro_p3.F90
```diff
@@ -407,11 +408,13 @@ SUBROUTINE p3_main_part1(kts, kte, kbot, ktop, kdir, do_predict_nc, dt, &
       !--- Apply droplet activation here (before other microphysical processes) for consistency with qc increase by saturation
       !    adjustment already applied in macrophysics. If prescribed drop number is used, this is also a good place to
       !    prescribe that value
-          if (.not.(do_predict_nc)) then
-            nc(k) = nccnst*inv_rho(k)
+          if (do_prescribed_CCN) then
+             nc(k) = max(nc(k),nccn_prescribed(k))
+          else if (do_predict_nc) then
+             nc(k) = max(nc(k) + nc_nuceat_tend(k) * dt,0.0_rtype)
           else
```
PeterCaldwell, line 415: Let's put a comment after else explaining what this case is: "!only remaining option is prescribed nc"

AaronDonahue:
@hassanbeydoun , is this now an old PR that was replaced by #642 ?  If so, can we delete it?

PeterCaldwell:
closed via #642

PR #599, opened 2020-09-22
[WIP] use bfb functions in shoc.F90

tcclevenger:
After this, I'm creating a PR for porting `compute_diag_third_shoc_moment` and I need to change some subroutines in shoc.F90 to use bfb_*() functions.

### DIFF
--- a/components/cam/src/physics/cam/shoc.F90
+++ b/components/cam/src/physics/cam/shoc.F90
@@ -1895,7 +1895,7 @@ subroutine clipping_diag_third_shoc_moments(&
 
       tsign = 1._rtype
       theterm = w_sec_zi(i,k)
-      cond = w3clip * bfb_sqrt(2._rtype * theterm**3)
+      cond = w3clip * bfb_sqrt(2._rtype * bfb_cube(theterm))
       if (w3(i,k) .lt. 0) tsign = -1._rtype
       if (tsign * w3(i,k) .gt. cond) w3(i,k) = tsign * cond
 
--- a/components/cam/src/physics/cam/shoc.F90
+++ b/components/cam/src/physics/cam/shoc.F90
@@ -1714,7 +1714,7 @@ subroutine fterms_input_for_diag_third_shoc_moment(&
   thedz2 = 1._rtype/(dz_zt+dz_zt_kc)
 
   iso       = isotropy_zi
-  isosqrd   = iso**2
+  isosqrd   = bfb_square(iso)
   buoy_sgs2 = isosqrd*brunt_zi
   bet2      = ggr/thetal_zi
 
@@ -1751,10 +1751,10 @@ subroutine f0_to_f5_diag_third_shoc_moment(&
   wsec_diff     = w_sec_kc    - w_sec
   tke_diff      = tke_kc      - tke
 
-  f0 = thedz2 * bet2**3 * iso**4 * wthl_sec * &
+  f0 = thedz2 * bfb_cube(bet2) * bfb_pow(iso,4._rtype) * wthl_sec * &
        thl_sec_diff
 
-  f1 = thedz2 * bet2**2 * iso**3 * (wthl_sec * &
+  f1 = thedz2 * bfb_square(bet2) * bfb_cube(iso) * (wthl_sec * &
        wthl_sec_diff + 0.5_rtype * &
        w_sec_zi*thl_sec_diff)
 
@@ -1788,6 +1788,7 @@ subroutine omega_terms_diag_third_shoc_moment(&
   !intent-out
   real(rtype), intent(out) :: omega0, omega1, omega2
 
+
   real(rtype), parameter :: a4=2.4_rtype/(3._rtype*c_diag_3rd_mom+5._rtype)
   real(rtype), parameter :: a5=0.6_rtype/(c_diag_3rd_mom*(3._rtype+5._rtype*c_diag_3rd_mom))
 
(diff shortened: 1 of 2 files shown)


components/cam/src/physics/cam/shoc.F90
```diff
@@ -1751,10 +1751,10 @@ subroutine f0_to_f5_diag_third_shoc_moment(&
   wsec_diff     = w_sec_kc    - w_sec
   tke_diff      = tke_kc      - tke
 
-  f0 = thedz2 * bet2**3 * iso**4 * wthl_sec * &
+  f0 = thedz2 * bfb_cube(bet2) * bfb_pow(iso,4._rtype) * wthl_sec * &
```
ambrad, line 1754: For 4, we want bfb_quad that evals to bfb_square(bfb_square(x)).

jeff-cohere, line 1754: I guess this call to `bfb_pow` may change once we get `pow<4>` into EKAT. Just a comment--no need to change anything.

components/cam/src/physics/cam/shoc.F90
```diff
@@ -1812,10 +1813,13 @@ subroutine x_y_terms_diag_third_shoc_moment(&
...
   real(rtype), intent(out) :: x0, y0, x1, y1
 
-  real(rtype), parameter :: a0=(0.52_rtype*c_diag_3rd_mom**(-2))/(c_diag_3rd_mom-2._rtype)
-  real(rtype), parameter :: a1=0.87_rtype/(c_diag_3rd_mom**2)
-  real(rtype), parameter :: a2=0.5_rtype/c_diag_3rd_mom
-  real(rtype), parameter :: a3=0.6_rtype/(c_diag_3rd_mom*(c_diag_3rd_mom-2._rtype))
+  ! local variables
+  real(rtype) :: a0, a1, a2, a3
+
+  a0 = (0.52_rtype*bfb_pow(c_diag_3rd_mom,-2._rtype))/(c_diag_3rd_mom-2._rtype)
```
ambrad, line 1819: Maybe this changed, but I believe bfb_square, etc, are macros, so you can keep these as parameters.

components/cam/src/physics/cam/shoc.F90
```diff
@@ -1891,7 +1895,7 @@ subroutine clipping_diag_third_shoc_moments(&
 
       tsign = 1._rtype
       theterm = w_sec_zi(i,k)
-      cond = w3clip * sqrt(2._rtype * theterm**3)
+      cond = w3clip * bfb_sqrt(2._rtype * theterm**3)
```
ambrad, line 1898: There's a cube here.

components/cam/src/physics/cam/shoc.F90
jeff-cohere: Ah, never mind. :-) Andrew's got the in-time solution.

components/cam/src/physics/cam/shoc.F90
jeff-cohere: They are still macros.

tcclevenger: I initially tried this, but I got a `Function ‘cxx_pow’ in initialization expression at (1) must be an intrinsic function` error.

components/cam/src/physics/cam/shoc.F90
tcclevenger: There doesn't appear to be a `bfb_quad` in bfb_math.inc. I can add one if you like, or I can change this to `bfb_square(bfb_square(iso))`.

jeff-cohere: I'd say go ahead and add `bfb_quad` if you haven't already.

tcclevenger:
@jeff-cohere @ambrad I've added `bfb_quad` to the bfb definitions. The only question still is surrounding the error when trying to use bfb_* in the parameter definitions.

components/cam/src/physics/cam/shoc.F90
bartgol: Yeah, cxx_pow requires a cuda kernel launch, so for sure not good for parameter in F90. But even on host, cxx_pow is not something that can be compiled away (it comes from another module).

And the compiler error seems to suggest it would not work even if cxx_pow was defined in the same module.

I guess no constexpr in F90...

ambrad: Ok, yeah, can't do it b/c of the C++ code path. I got that wrong in my initial comment.

ambrad (approved):
Be sure to check relative errors in shoc_run_and_cmp output carefully, both SP and DP.

jeff-cohere (approved):

tcclevenger:
> Be sure to check relative errors in shoc_run_and_cmp output carefully, both SP and DP.

shoc_run_and_cmp_ut_np1_omp16 passes on my machine even with this new change. Should this not be the case? I maybe be running the tests incorrectly..

tcclevenger:
Changed title to [WIP] until tester is fixed

PeterCaldwell:
If by "tester is fixed", you mean the shoc fails are dealt with - that's done now. Just merge master into your branch before you retest.

PeterCaldwell:
Oops - sorry folks. I just merged master into this branch thinking it was one of the other PRs I'm wrangling. I suspect this is a good thing for you, but it is unexpected so I thought I'd give a heads up.

tcclevenger:
@PeterCaldwell Thanks! I'll take it.. haha

tcclevenger:
@ambrad @jeff-cohere This also needs re-approval for autotester to run.

jeff-cohere (approved):

tcclevenger:
From the output it appears the relative difference of host_dse is quite high. Not sure what values we expect? Also, not sure why the GPU passes if CPU has such a high reltol for host_dse.

ambrad:
We expect rel diff to be just a little above std::numeric_limits<Real>::epsilon(). In the above, sometimes it is, but in the blake output, it looks erroneously high. You should investigate manually the failing build on blake.

tcclevenger:
I've rebased with master, and now testing on both Blake and Weaver, all tests (including the ones which were previously failing) pass.

ambrad (approved):

tcclevenger:
Working off a forked branch currently, which doesn't allow the tester to run without approval. Switching to branch off of the scream repo so I can debug without everyone getting emails. I'll request reviews once I've sorted out the issues.

PR #601, opened 2020-09-23
SHOC bug fix for Murphy/Koop implementation (and unrelated property test fix)

### LINKED ISSUE #600: Shoc tests failing in master
The error looks like the following:

```
/shoc/tests/shoc_pdf_compute_qs_tests.cpp:79: FAILED:
  REQUIRE( SDS.qs1 < qs1_test1 )
with expansion:
  0.622 < 0.622
```

@singhbalwinder and @tcclevenger first reported this kind of failure. @AaronDonahue also noticed it in LLNL machines nightlies, and I was able to reproduce it on my laptop. The error occurs not every time, which pointed to uninited memory. Valgrind seemd to confirm the suspicion, so the task is now to go up the update-tree and see which of the variables contributing to qs1 is not initialized.

bogensch:
This PR address two unrelated SHOC issues reported and closes #600 

1) With the implementation of Murphy Koop in PR #588 the 100 factor unit conversion is no longer needed since Murphy Koop returns in units of [Pa], unlike the old Flatau implementation.  This bug was uncovered by a failing SHOC property test.

2) It was also reported that the liquid water flux property test was failing when use with valgrind.  This was due to the grid mean vertical velocity not being initialized in any of the tests.  This has been fixed.

### DIFF
--- a/components/cam/src/physics/cam/shoc.F90
+++ b/components/cam/src/physics/cam/shoc.F90
@@ -2491,7 +2491,7 @@ subroutine shoc_assumed_pdf_compute_qs(&
   esval1_1=0._rtype
   esval1_2=0._rtype
 
-  esval1_1=MurphyKoop_svp(Tl1_1, liquid)*100._rtype
+  esval1_1=MurphyKoop_svp(Tl1_1, liquid)
   lstarn1=lcond
 
   qs1=0.622_rtype*esval1_1/max(esval1_1,pval-esval1_1)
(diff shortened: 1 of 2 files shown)


singhbalwinder (approved):
Thanks for fixing this. Sorry that I missed checking the units.

bartgol (approved):

PeterCaldwell:
Take a moment to rejoice - shoc property tests caught a bug!

bogensch:
@singhbalwinder I should have caught it in my review.... plus the complete lack of comments in my code about what that factor was meant for would have made it very difficult for anyone to catch this.

PeterCaldwell:
@bogensch - would you take a quick look at the test results above and confirm they're what you expect? Once you do, I'll integrate. It looks like we only fail shoc_run_and_cmp_ut_np1_omp* on all machines, which I think we expect because the change is non-bfb.

@jeff-cohere or @bartgol : It looks from the 'tail <log>' info above like Blake and Bowman are only failing with omp16. This seems weird because run_and_cmp should fail for all thread count. Are we only testing for 16 threads in shoc_run_and_cmp, or is the log looping over each thread count in such a way that other failures are lost to the tail? Also, the above log indicates we're running 150 steps. Wouldn't running 5 steps be sufficient?

bartgol:
The run_and_cmp tests only run with the maximum thread count. I don't know what those tests do under the hood though.

bogensch:
This is what I expect.  I did a DYCOMS-RF01 SCM test with this branch and results look as expected.

PR #576, opened 2020-09-15
[WIP] Establish a scorpio class for handling input/output

AaronDonahue:
This commit introduces the AtmosphereOutput class.  This class interfaces directly with the scorpio interface infrastructure and the field manager to control output data.

The capability to interface directly with the field manager enables output for the scream-ad.

The class and the underlying functions are tested with a new unit test.

This commit also moves the scorpio interface code from mct_coupling to a new directory, "io", in the share directory.

PR #604, opened 2020-09-24
Make field repository a shared pointer in the AD

AaronDonahue:
This commit makes the m_device_field_repo variable in the AD
a shared pointer.

This step is a precursor to storing the field repository inside
an output manager for IO purposes. See PR #576 

[BFB]


bartgol (approved):

AaronDonahue:
Though not technically WIP, this PR keeps failing the AT and from the output log is also spamming Luca with emails.  So I'm putting WIP to save on the AT.

bartgol (approved):

PR #608, opened 2020-09-24
SHOC Implicit diffusion interface lower level property tests

bogensch:
Adds property tests for lower level functions in the implicit diffusion interface (but not for the solver itself as those *.F90 will not be converted).  

Note that there were some b4b modifications made to shoc.F90 file to address:
- Two routines (impli_srf_stress_term, sfc_fluxes) only perform computations at the surface, but entire profiles were being read in.  Changed to only input surface level variables.
- Renames several variables with appended “_sfc” for clearer distinction that these are surface level variable.
- Routine “dp_inverse” passed in the never used “nlevi”, removed this.


jgfouca (approved):

singhbalwinder (approved):
Thanks Pete! I thought about making similar changes to SHOC.F90 while refactoring but I was a little reluctant on sending array slices in the Fortran code. Although it is done in the other part of the code, it requires the compiler to generate a temporary variable when we make that call. It is not a huge issue and we are anyway going to use C++ code in the future. I like this change as the functions code looks nice and also we are not sending nlev as arg when we only want to compute surface quantities.

singhbalwinder (approved):

PR #609, opened 2020-09-24
fix ekat calls

tcclevenger:
@bartgol It appears this function fell through the cracks on the EKAT update (must have been merged after edits). This PR updates the ekat calls.

### DIFF
--- a/components/scream/src/physics/shoc/shoc_functions_f90.cpp
+++ b/components/scream/src/physics/shoc/shoc_functions_f90.cpp
@@ -585,9 +585,9 @@ void shoc_diag_second_moments_ubycond(SHOCSecondMomentUbycondData& d)
 void shoc_pblintd_init_pot(SHOCPblintdInitPotData& d)
 {
   shoc_init(d.nlev(), true);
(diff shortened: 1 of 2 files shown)


bartgol (approved):
Interesting. The only scenario where AT testing infrastructure can fail is the following:

1) PR#1 is tested, and passes. No review yet, so not merged.
2) PR#2 is tested, passes, and get merged.
3) PR#1 is reviewed and accepted, and AT merges.

If PR#2 breaks something for PR#1, you have a problem, cause the AT won't retest it.

bartgol:
This is odd. I grepped in the branch, and did not notice any `d.transposeekat::util::TransposeDirection::f2c();` line. Perhaps test-all-scream is merging master not correctly? I can hardly imagine git being this sloppy at merging. I would expect a merge failure (which should abort the testing).

@jgfouca ^^

bartgol:
Nvm, probably the messy output is just caused by the different build reporting simultaneously.

The real issue is that master will never build successfully, so the AT will fail forever. I think this PR has to be merged forcefully...

PR #611, opened 2020-09-25
gen-boiler: new tool for generating boiler

jgfouca:
New tool for generating boilerplate along with a demonstration of that tool for linear_interp. Some linear_interp generated code had to be manually massaged because existing pieces did not exactly follow the standard pattern.

Feature highlights:
Supports generating the following pieces of code:
 ```
                        f90_c2f_bind     : The c to f90 fortran subroutine(<name>_c)
                        f90_f2c_bind     : The f90 to c fortran subroutine(<name>_f)
                        cxx_c2f_bind_decl: The c to f90 c function declaration(<name>_c)
                        cxx_c2f_glue_decl: The cxx to c function declaration(<name>(Data))
                        cxx_c2f_glue_impl: The cxx to c function implementation(<name>(Data))
                        cxx_c2f_data     : The cxx data struct definition(struct Data)
                        cxx_f2c_bind_decl: The f90 to cxx function declaration(<name>_f)
                        cxx_f2c_bind_impl: The f90 to cxx function implementation(<name>_f)
                        cxx_func_decl    : The cxx kokkos function declaration(<name>)
                        cxx_incl_impl    : The include of *impl.hpp file at bottom of main hpp
                        cxx_func_impl    : The cxx kokkos function stub implementation(<name>)
                        cxx_bfb_unit_decl: The cxx unit test struct declaration
                        cxx_bfb_unit_impl: The cxx bfb unit test implementation
                        cxx_eti          : The cxx explicit template instatiation file
                        cmake_impl_eti   : Make cmake aware of the ETI file if not cuda build
                        cmake_unit_test  : Make cmake aware of the unit test
```

Supports overwriting existing pieces or inserting new code.

Should be able to support both p3 and shoc, although only tested on shoc at the moment. P3 may need some minor formatting changes for full support.

After running the tool, the repo should be buildable without additional modifications! Of course, some of the implementations will need to be filled-in in order for the code to really work.

The tool is fully unit-tested for all non-trivial functions/methods.

### CODE REVIEW

**Comment by jgfouca** [comment]
The CLI, as described by the --help:

```
usage: 
gen-boiler <subroutine-name> [<subroutine-name> ...] [--piece] [--kernel] [--verbose]
OR
gen-boiler --help

EXAMPLES:
    # Generate all boilerplate p3 subroutine foo_bar 
    > gen-boiler foo_bar

    # Generate all boilerplate for shoc subroutine bar_baz and func_name 
    > gen-boiler bar_baz func_name

    # Generate only f90->cxx bindings for p3 subroutine foo_bar 
    > gen-boiler foo_bar --piece=f90_f2c_bind

    # Overwrite only f90->cxx and cxx->f90 bindings for p3 subroutine foo_bar 
    > gen-boiler foo_bar --piece=f90_f2c_bind --piece=f90_c2f_bind -o

Generate boiler-plate for converting a fortran subroutine to cxx.

This script is designed to facilitate f90->cxx conversion of p3 and shoc
subroutines, with the potential to be expanded to additional packages later.
This script should also help maintain the latest approved coding patterns.

To test this tool, run:
% python3 -m doctest gen_boiler.py

positional arguments:
  subs                  fortran subroutine name[s]. Must specify at least one

optional arguments:
  -h, --help            show this help message and exit
  -p , --piece          Select which pieces of boilerplate to generate.
                        Piece choices:
                        f90_c2f_bind     : The c to f90 fortran subroutine(<name>_c)
                        f90_f2c_bind     : The f90 to c fortran subroutine(<name>_f)
                        cxx_c2f_bind_decl: The c to f90 c function declaration(<name>_c)
                        cxx_c2f_glue_decl: The cxx to c function declaration(<name>(Data))
                        cxx_c2f_glue_impl: The cxx to c function implementation(<name>(Data))
                        cxx_c2f_data     : The cxx data struct definition(struct Data)
                        cxx_f2c_bind_decl: The f90 to cxx function declaration(<name>_f)
                        cxx_f2c_bind_impl: The f90 to cxx function implementation(<name>_f)
                        cxx_func_decl    : The cxx kokkos function declaration(<name>)
                        cxx_incl_impl    : The include of *impl.hpp file at bottom of main hpp
                        cxx_func_impl    : The cxx kokkos function stub implementation(<name>)
                        cxx_bfb_unit_decl: The cxx unit test struct declaration
                        cxx_bfb_unit_impl: The cxx bfb unit test implementation
                        cxx_eti          : The cxx explicit template instatiation file
                        cmake_impl_eti   : Make cmake aware of the ETI file if not cuda build
                        cmake_unit_test  : Make cmake aware of the unit test
                        
                        Default will generate ALL pieces
  --physics {shoc,p3}   
                        Limit generation to a specific set of physics. This should only be
                        needed to disambiguate cases where a subroutine has the same name in
                        multiple physics. This may also improve performance.
                        Default will scan ALL physics.
  -o, --overwrite       Overwrite existing boilerplate.
                        Default: False
  -k, --kernel          The generate cxx functions need the ability to launch kokkos kernels.
                        Default: False
  -s SOURCE_REPO, --source-repo SOURCE_REPO
                        The repo you wish to scan for fortran subroutines.
                        Default: /home/jgfouca/scream
  -t TARGET_REPO, --target-repo TARGET_REPO
                        The repo you wish to generate boilerplate into.
                        Default: /home/jgfouca/scream
  -d, --dry-run         Do a dry run instead of modifying files.
                        Default: False
  -v, --verbose         Produce extra output.
                        Default: False
```

**Review by ambrad — APPROVED** [review]
This looks nifty! I'm approving, but the primary code converters should weigh in before you merge this.

**Review by jeff-cohere — APPROVED** [review]
Nice! I like how this specifically encodes the cleaned-up boilerplate of the sort we've been using. I think this will allow everyone to focus a lot more on the actual ported code without the distraction of copying additional files around.

**Comment by jgfouca** [comment]
I will give the other reviewers one more day to review before I merge. I want to get this on master ASAP so we can speed up the code conversion process. I very much liked what this tool was able to do for me for linear_interp.

**Review by singhbalwinder — APPROVED** [review]
Thanks Jim!

**Review by tcclevenger — APPROVED** [review]

File: components/scream/scripts/gen_boiler.py [hunk 1]
```diff
@@ -0,0 +1,1878 @@
...
+TEST_CASE({sub}_bfb, "[{physics}_functions]")
+{{
+  using TestStruct = scream::{physics}::unit_test::UnitWrap::UnitTest<scream::DefaultDevice>::{test_data_struct};
+
+  TestStruct::run_bfb();
+}}
+
+}} // empty namespace
+""".format(physics=phys, sub=sub, test_data_struct=get_data_test_struct_name(sub), gen_code=gen_code),
+
```
PeterCaldwell, line 49: Personally, I'd prefer to have a row of "#=============" between each of these pieces to make it quicker to tell where one ends and the next starts. These templates are the core of the tool, so we should make them easy to understand. That said, I feel like most people don't like adding lots of #======== lines to their code, so feel free to ignore.

bogensch (approved):
Super slick, thanks!

PeterCaldwell:
Wow, this is really fantastic! Thanks!!! I feel like it's Christmas morning... but I've got a bunch of small questions/comments:

1. I checked out your branch on quartz and ran `./gen-boiler p3_evaporate_rain`, then `git status` and didn't see any changes even though no error message was generated. In fact, none of the arguments/subroutine names I added seemed to have any effect. What's expected here? I thought gen-boiler would change the code itself, but maybe it produces code snippets outside of the git repo? It would be good to document the expected effect a bit better. And for the code to be more expressive about what its doing so clueless people like me can figure out what is expected.
2. If I want to remake boilerplate for p3_evaporate_rain_impl.hpp, do I run gen-boiler with "evaporate_rain"? Or "p3_evaporate_rain"?
3. Does the code put stuff in the p3 or shoc directory based on the first few letters of the target function name (i.e. the p3_ in p3_evaporate_rain is critical)? Will it error out if not prepended by p3_ or shoc_?
4. I noticed that we have gen-boiler (with hyphen) but gen_boiler.py (with underscore). Is this inconsistency intended?
5. This is perhaps a bigger discussion, but I think we agreed at one point that we'd like _c and _f in function names to be replaced by _c2f and _f2c (respectively). Should we add this in the boilerplate generator now, or should we do that in a future PR?
6. Is there a unit test for gen-boiler? There probably is, but I didn't see it in my skim through the code...
7. I recall that Luca wanted to break up p3_functions_f90.cpp because it was so huge. It seems like having each function have its own file for this bit of boilerplate would make the boilerplate generator simpler. Did we want to make this change in this PR?
8. The help for gen-boiler says that you can generate a single "piece" of boilerplate at a time: "f2c", "c2f", or ???. It would be good to list all the possible pieces in the help message. Is the 3rd bit just the bfb testing? Or are there other pieces as well?
9. It looks like this code automatically generates the BFB test. This is really exciting/awesome! It seems like users would need to go through the BFB test code and specify the bounds for each of the input variables (if the defaults aren't good enough). Is that the situation?

components/scream/scripts/gen_boiler.py
bartgol: Would it make sense to put these templates inside files, like in `scream/scripts/code-templates` or something?

components/scream/scripts/gen_boiler.py
```diff
@@ -0,0 +1,1878 @@
...
+    "cxx_eti": lambda phys, sub, gen_code: gen_code
+}
+
+# piece map. maps the name of a piece of boilerplate that needs to be genereate to:
+#   (filepath (relative to cxx_root))
+FILEPATH, FILECREATE, INSERT_REGEX, ID_SELF_BEGIN_REGEX, ID_SELF_END_REGEX, DESC = range(6)
+PIECES = OrderedDict([
+    ("f90_c2f_bind", (
+        lambda phys, sub, gb: "{}_iso_c.f90".format(phys),
```
bartgol, line 81: Since we are (re-)establishing a pattern, could we call the bridge files `{}_f2c.f90` rather than `{}_iso_c.f90` (and similarly for the other way around)? I am never sure (and I might have gotten it wrong here too) which direction are `{}_iso_[c|f].f90` going. Imho, `f2c` and `c2f` are way clearer in explaining who calls what.

components/scream/scripts/gen_boiler.py
```diff
@@ -0,0 +1,1878 @@
...
+        "p3_init();"
+    ),
+    "shoc" : (
+        "components/cam/src/physics/cam/shoc.F90",
+        "components/scream/src/physics/shoc",
+        "shoc_init(d.nlev, true);"
+    ),
+}
+
+# A good set of arg data for unit testing
```
bartgol, line 242: I'm not sure I understand what these arrays are for. I don't see them used in the file either, only in the function help. Unless I'm missing some subtle mechanism?

components/scream/scripts/gen_boiler.py
```diff
@@ -0,0 +1,1878 @@
...
+UT_ARG_DATA = [
+    ("foo1", "real", "in", ("shcol",)),
+    ("foo2", "real", "in", ("shcol",)),
+    ("bar1", "real", "in", ("shcol","nlev")),
```
bartgol, line 246: Didn't we change shcol to ncols or something like that in a recent PR? Or maybe it's still in the todo list?

components/scream/scripts/gen-boiler
```diff
@@ -0,0 +1,99 @@
...
+    parser = argparse.ArgumentParser(
+        usage="""\n{0} <subroutine-name> [<subroutine-name> ...] [--piece] [--kernel] [--verbose]
+OR
+{0} --help
+
+\033[1mEXAMPLES:\033[0m
+    \033[1;32m# Generate all boilerplate p3 subroutine foo_bar \033[0m
+    > {0} foo_bar
+
+    \033[1;32m# Generate all boilerplate for shoc subroutine bar_baz and func_name \033[0m
```
bartgol, line 33: I think these helps are a bit confusing: this one says it creates stuff for shoc, the next for p3, but there's no mention of pr or shoc in the help. One (e.g., me) may wonder how does gen-boiler figure out where to put bar_baz...

bartgol: Edit: I tried adding `--physics shoc`, and it works. Perhaps we should add that in the help.

components/scream/scripts/gen_boiler.py
```diff
@@ -0,0 +1,1878 @@
...
+        lambda phys, sub, gb: "{}_{}.cpp".format(phys, sub),
+        lambda phys, sub, gb: create_template(phys, sub, gb, "cxx_eti"),
+        lambda phys, sub, gb: re.compile(".*"), # insert at top of file
+        lambda phys, sub, gb: re.compile(".*"), # start at top of file
+        lambda phys, sub, gb: get_namespace_close_regex("scream"), #end of file
+        lambda *x           : "The cxx explicit template instatiation file"
+    )),
+
+    ("cmake_impl_eti", (
+        lambda phys, sub, gb: "CMakeLists.txt",
```
bartgol, line 207: Reading the code I was having a hard time telling what sub, and gb were. After some time I realized gb=gen-boiler, and sub=soubroutine. Right?

Could we use 'func' instead of 'sub', or spell it out completely (subroutine)? The word 'sub' is quite vague until you link it to subroutine, and code maintainers (other than Jim) might have a hard time diving into the script.

components/scream/scripts/gen_boiler.py
```diff
@@ -0,0 +1,1878 @@
...
+        format(argtype=argtype, c_type=c_type, value=value, intent=intent_s, dimension=dimension_s, names=names_s)
+
+CXX_TYPE_MAP = {"real" : "Real", "integer" : "Int", "logical" : "bool"}
+###############################################################################
+def get_cxx_type(arg_datum):
+###############################################################################
+    """
+    Based on arg datum, give c++ bridge type
+
+    >>> get_cxx_type(("foo", "real", "in", ("100",)))
```
bartgol, line 812: I don't understand the examples. Where would the input strings come from? Would they come from parsing f90 decl lines? Or are the inputs just a random set of strings from which get_cxx_type find the only one (assume there's only one) that matches a known type? I'm guessing it's the former. It might be worth adding a couple of words saying that in the comments, like "the list of strings in the input is assumed to come from the parsing of a f90 type declaration. As such, we _expect only one_ string to match a known F90 type". Or something like that.

components/scream/scripts/gen_boiler.py
```diff
@@ -0,0 +1,1878 @@
...
+###############################################################################
+def get_kokkos_type(arg_datum):
+###############################################################################
+    """
+    Based on arg datum, give c++ kokkos type
+
+    Note: We can only guess at the correct types, especially whether an argument
+    should be packed data or not!
+
+    >>> get_kokkos_type(("foo", "real", "in", ("100",)))
```
bartgol, line 843: Same comment as for `get_cxx_type`.

components/scream/scripts/gen_boiler.py
```diff
@@ -0,0 +1,1878 @@
...
+    # over one of the dimensions
+    return "const uview_1d<{}>&".format(base_type) if is_view else "{}&".format(base_type)
+
+###############################################################################
+def gen_arg_cxx_decls(arg_data, kokkos=False):
+###############################################################################
+    """
+    Get all arg decls for a set of arg data
+
+    >>> gen_arg_cxx_decls([("foo", "real", "in", ("100",)), ("bar", "real", "in", None)])
```
bartgol, line 872: Same comment as for `get_cxx_type`.

components/scream/scripts/gen_boiler.py
```diff
@@ -0,0 +1,1878 @@
...
+
+    return False
+
+###############################################################################
+def gen_cxx_data_args(arg_data):
+###############################################################################
+    """
+    Based on data, generate unpacking of Data struct args
+
+    >>> gen_cxx_data_args([("foo", "real", "in", ("100",)), ("bar", "real", "in", None), ("baz", "integer", "inout", None), ("bag", "integer", "inout", ("100",))])
```
bartgol, line 909: Same comment as for `get_cxx_type`.

components/scream/scripts/gen_boiler.py
bartgol: After further reading the file, I _think_ these are only used in documentation examples. Might be worth saying it in the comment. Like "An example set of arg data to be used in the documentation of the methods below". So the user doesn't get confused about why `foo1` could ever be used in a real case (it got me thinking for a while).

bartgol (approved):
Huge piece of work Jim! I didn't check the actual python logic, I assume you tested it several times. Besides, not having developed anything in physics, I would not know how to check the behavior (including corner cases).

The only comments I had were regarding documentation. Being a complex script, if for any reason you can't maintain it anymore at some point, we need other people to quickly catch up with what the script does. In this respect, I think improving inline documentation is very important. I had a few suggestions, but I'm not marking this are "required changes". Some people might like this script to be available sooner than later, so comments can be added after merging. But I really think they should be added at some point.

components/scream/scripts/gen-boiler
jgfouca: Done.

components/scream/scripts/gen_boiler.py
jgfouca: I went with @PeterCaldwell 's suggestion of a comment separator to make these more readable. This file has gotten pretty big, so I will keep in mind @bartgol 's thought of perhaps offloading some items to different files.

components/scream/scripts/gen_boiler.py
jgfouca: Yes, that should be easy if others agree.

components/scream/scripts/gen_boiler.py
jgfouca: That's right. I'm conflicted about using longer names since I like that all these lambda are one line and long names would either make these lines really long or force me to introduce newlines.

components/scream/scripts/gen_boiler.py
jgfouca: I'm a bit confused, did you see the comment above?
```
# A good set of arg data for unit testing
```
Are you just looking for a better comment?

components/scream/scripts/gen_boiler.py
jgfouca: @bartgol , a core data structure in this program is the sub->args data base where args is a list of tuples (ARG_NAME, ARG_TYPE, ARG_INTENT, ARG_DIMS). These tuples are produced by parse_f90_args. Anywhere you see `arg_datum` it's referring to single one of these tuples. Anywhere you see `arg_data` it's referring to all the tuples for a sub. I will add comments.

components/scream/scripts/gen_boiler.py
bartgol: I saw it. But at first I thought those were actual data to be used in actual unit testing. But reading the rest of the file makes me think they are just a dummy set of arg data to be used in the comments section of each (python) function in the file, to provide example of the output produced by such functions. Is that right? If so, then yes, I'm looking for a better comment I guess.

components/scream/scripts/gen_boiler.py
bartgol: This is not a big deal, just trying to make the life of the maintainer who comes after you a bit easier.

What aboud `func`? I know functions and subroutines are different things in f90, but `sub` was not very clear at first. Or maybe `subrtn`? Anything that makes it unequivocal what we're taking about, I guess.

About long names: these names are only to be used inside this script, so it only affects the maintainer(s) of this tool.

Anyhow, as I said, not a big deal.

jgfouca:
@PeterCaldwell 

> 1. I checked out your branch on quartz and ran `./gen-boiler p3_evaporate_rain`, then `git status` and didn't see any changes even though no error message was generated. In fact, none of the arguments/subroutine names I added seemed to have any effect. What's expected here? I thought gen-boiler would change the code itself, but maybe it produces code snippets outside of the git repo? It would be good to document the expected effect a bit better. And for the code to be more expressive about what its doing so clueless people like me can figure out what is expected.

I've updated the documentation to recommend users add a `--physics` option to focus the scanning on a single file. gen-boiler will change the code unless `--dry-run` is provided. Didn't you get a warning saying it didn't find any subroutine `p3_evaporate_rain`?

> 2. If I want to remake boilerplate for p3_evaporate_rain_impl.hpp, do I run gen-boiler with "evaporate_rain"? Or "p3_evaporate_rain"?

Gen-boiler expects you to refer to a subroutine by it's exact name in the primary source file (micro_p3.F90 in this case), so it would be `evaporate_rain`.

> 3. Does the code put stuff in the p3 or shoc directory based on the first few letters of the target function name (i.e. the p3_ in p3_evaporate_rain is critical)? Will it error out if not prepended by p3_ or shoc_?

The physics package is selected by the `--phyiscs` argument (`--physics=p3` in this case). The subroutine name should not be adjusted.

> 4. I noticed that we have gen-boiler (with hyphen) but gen_boiler.py (with underscore). Is this inconsistency intended?

Yes. One is the CLI executable, so has flexibility for its name and I personally prefer hyphen over underscore. Hyphens are both easier to type and visually more appealing to me. gen_boiler.py is a python library that supports the CLI and python library files cannot have hyphens.

> 5. This is perhaps a bigger discussion, but I think we agreed at one point that we'd like _c and _f in function names to be replaced by _c2f and _f2c (respectively). Should we add this in the boilerplate generator now, or should we do that in a future PR?

Good question. It seems you agree with @bartgol on the name changes (I like it too). I would save this for a subsequent PR just for my sanity.

> 6. Is there a unit test for gen-boiler? There probably is, but I didn't see it in my skim through the code...

Yes, gen_boiler.py is filled with high-quality doctests. There a snippet in the --help on how to run them
```
To test this tool, run:                                                                                                                                                                                                                            
% python3 -m doctest gen_boiler.py 
```

> 7. I recall that Luca wanted to break up p3_functions_f90.cpp because it was so huge. It seems like having each function have its own file for this bit of boilerplate would make the boilerplate generator simpler. Did we want to make this change in this PR?

I think we do want this, but not as part of this PR.

> 8. The help for gen-boiler says that you can generate a single "piece" of boilerplate at a time: "f2c", "c2f", or ???. It would be good to list all the possible pieces in the help message. Is the 3rd bit just the bfb testing? Or are there other pieces as well?

If you look at the --help dump for the --piece argument:
```
  -p , --piece          Select which pieces of boilerplate to generate.
                        Piece choices:
                        f90_c2f_bind     : The c to f90 fortran subroutine(<name>_c)
                        f90_f2c_bind     : The f90 to c fortran subroutine(<name>_f)
                        cxx_c2f_bind_decl: The c to f90 c function declaration(<name>_c)
                        cxx_c2f_glue_decl: The cxx to c function declaration(<name>(Data))
                        cxx_c2f_glue_impl: The cxx to c function implementation(<name>(Data))
                        cxx_c2f_data     : The cxx data struct definition(struct Data)
                        cxx_f2c_bind_decl: The f90 to cxx function declaration(<name>_f)
                        cxx_f2c_bind_impl: The f90 to cxx function implementation(<name>_f)
                        cxx_func_decl    : The cxx kokkos function declaration(<name>)
                        cxx_incl_impl    : The include of *impl.hpp file at bottom of main hpp
                        cxx_func_impl    : The cxx kokkos function stub implementation(<name>)
                        cxx_bfb_unit_decl: The cxx unit test struct declaration
                        cxx_bfb_unit_impl: The cxx bfb unit test implementation
                        cxx_eti          : The cxx explicit template instatiation file
                        cmake_impl_eti   : Make cmake aware of the ETI file if not cuda build
                        cmake_unit_test  : Make cmake aware of the unit test
```

> 9. It looks like this code automatically generates the BFB test. This is really exciting/awesome! It seems like users would need to go through the BFB test code and specify the bounds for each of the input variables (if the defaults aren't good enough). Is that the situation?

I'm excited about this too. Yes, the user will need to provide reasonable inputs in the constructors for the Data structs.

components/scream/scripts/gen_boiler.py
PeterCaldwell: I agree that _f2c.f90 would be better than iso_c.f90!

components/scream/scripts/gen-boiler
PeterCaldwell: Yeah, that solves a big confusion of mine. Thanks.

PeterCaldwell:
Ok, thanks Jim for the explanation. Everything you say makes good sense. A couple follow-up comments:

`gen-boiler cows` returns without doing anything or printing anything. `gen-boiler cows --physics p3` tries to do a bunch of stuff and complains about cows not being defined. 

 It would be good to include an example with --physics in the examples. I tried to get the help by just typing `gen-boiler` without `--help`. This printed some help so I assumed that was all there was... but didn't print all the help messages.

jgfouca:
@PeterCaldwell , I changed all the examples to include `--physics`, which I think should be the preferred approach for nearly all use cases.

The default behavior of python's argparse library is to print only the usage, not the full --help when the user makes an error. I could look into doing a full --help when no args are provided or just leave it as is.

PeterCaldwell:
@jgfouca - Sounds good. I don't care whether you make a change to show the full help or not. Seems more natural to me, but if it doesn't to you as well, don't bother.

PR #612, opened 2020-09-25
change quartz back to broadwell

PeterCaldwell:
Recent mods switched components/scream/cmake/machine-files/quartz.cmake from being BDW architecture to SKL, causing "illegal instruction" errors when scream tests were run.


bartgol (approved):
Whoops! Copy+paste error. Sorry!

AaronDonahue:
should we add "AutoMerge" to get this change in? Looks like all tests passed.

PeterCaldwell:
Ok. Original tests failed for unclear reasons. Passed just because Luca set the "Integrate without Testing" flag. I figured that would imply Automerge so I deleted it... apparently we need both. So I just added Automerge back.

PR #613, opened 2020-09-25
use bfb_* in various shoc.F90 functions

tcclevenger:
non-BFB changes to shoc.F90 for C++ conversions

### DIFF
--- a/components/cam/src/physics/cam/shoc.F90
+++ b/components/cam/src/physics/cam/shoc.F90
@@ -1751,7 +1751,7 @@ subroutine f0_to_f5_diag_third_shoc_moment(&
   wsec_diff     = w_sec_kc    - w_sec
   tke_diff      = tke_kc      - tke
 
-  f0 = thedz2 * bfb_cube(bet2) * bfb_square(bfb_square(iso)) * wthl_sec * &
+  f0 = thedz2 * bfb_cube(bet2) * iso**4 * wthl_sec * &
        thl_sec_diff
 
   f1 = thedz2 * bfb_square(bet2) * bfb_cube(iso) * (wthl_sec * &
--- a/components/cam/src/physics/cam/shoc.F90
+++ b/components/cam/src/physics/cam/shoc.F90
@@ -1894,7 +1894,7 @@ subroutine clipping_diag_third_shoc_moments(&
 
       tsign = 1._rtype
       theterm = w_sec_zi(i,k)
-      cond = w3clip * bfb_sqrt(2._rtype * bfb_cube(theterm))
+      cond = w3clip * bfb_sqrt(2._rtype * (theterm**3))
       if (w3(i,k) .lt. 0) tsign = -1._rtype
       if (tsign * w3(i,k) .gt. cond) w3(i,k) = tsign * cond
 
--- a/components/cam/src/physics/cam/shoc.F90
+++ b/components/cam/src/physics/cam/shoc.F90
@@ -1751,7 +1751,7 @@ subroutine f0_to_f5_diag_third_shoc_moment(&
   wsec_diff     = w_sec_kc    - w_sec
   tke_diff      = tke_kc      - tke
 
-  f0 = thedz2 * bet2**3 * bfb_quad(iso) * wthl_sec * &
+  f0 = thedz2 * bfb_cube(bet2) * bfb_quad(iso) * wthl_sec * &
        thl_sec_diff
 
   f1 = thedz2 * bfb_square(bet2) * bfb_cube(iso) * (wthl_sec * &
--- a/components/cam/src/physics/cam/shoc.F90
+++ b/components/cam/src/physics/cam/shoc.F90
@@ -1815,7 +1815,7 @@ subroutine x_y_terms_diag_third_shoc_moment(&
 ! local variables
   real(rtype) :: a0, a1, a2, a3
 
-  a0=(0.52_rtype*bfb_pow(c_diag_3rd_mom,-2))/(c_diag_3rd_mom-2._rtype)
+  a0=(0.52_rtype*bfb_pow(c_diag_3rd_mom,-2._rtype))/(c_diag_3rd_mom-2._rtype)
   a1=0.87_rtype/bfb_square(c_diag_3rd_mom)
   a2=0.5_rtype/c_diag_3rd_mom
   a3=0.6_rtype/(c_diag_3rd_mom*(c_diag_3rd_mom-2._rtype))
--- a/components/cam/src/physics/cam/shoc.F90
+++ b/components/cam/src/physics/cam/shoc.F90
@@ -1751,10 +1751,10 @@ subroutine f0_to_f5_diag_third_shoc_moment(&
   wsec_diff     = w_sec_kc    - w_sec
   tke_diff      = tke_kc      - tke
 
-  f0 = thedz2 * bet2**3 * iso**4 * wthl_sec * &
+  f0 = thedz2 * bet2**3 * bfb_quad(iso) * wthl_sec * &
        thl_sec_diff
 
-  f1 = thedz2 * bfb_square(bet2) * iso**3 * (wthl_sec * &
+  f1 = thedz2 * bfb_square(bet2) * bfb_cube(iso) * (wthl_sec * &
        wthl_sec_diff + 0.5_rtype * &
        w_sec_zi*thl_sec_diff)
 
@@ -1894,7 +1894,7 @@ subroutine clipping_diag_third_shoc_moments(&
 
       tsign = 1._rtype
       theterm = w_sec_zi(i,k)
-      cond = w3clip * bfb_sqrt(2._rtype * (theterm**3))
+      cond = w3clip * bfb_sqrt(2._rtype * bfb_cube(theterm))
       if (w3(i,k) .lt. 0) tsign = -1._rtype
       if (tsign * w3(i,k) .gt. cond) w3(i,k) = tsign * cond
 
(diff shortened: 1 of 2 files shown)


jgfouca (approved):

components/cam/src/physics/cam/shoc.F90
```diff
@@ -1812,10 +1812,13 @@ subroutine x_y_terms_diag_third_shoc_moment(&
...
   real(rtype), intent(out) :: x0, y0, x1, y1
 
-  real(rtype), parameter :: a0=(0.52_rtype*c_diag_3rd_mom**(-2))/(c_diag_3rd_mom-2._rtype)
-  real(rtype), parameter :: a1=0.87_rtype/(c_diag_3rd_mom**2)
-  real(rtype), parameter :: a2=0.5_rtype/c_diag_3rd_mom
-  real(rtype), parameter :: a3=0.6_rtype/(c_diag_3rd_mom*(c_diag_3rd_mom-2._rtype))
+! local variables
+  real(rtype) :: a0, a1, a2, a3
+
+  a0=(0.52_rtype*(c_diag_3rd_mom**(-2._rtype)))/(c_diag_3rd_mom-2._rtype)
```
ambrad, line 1818: Just saw this. We want `x^(-k)` to be impl'ed as `1/x^k`, so in this case `(1/bfb_square(c_diag_3rd_mom))`. The general principle is we avoid `pow` and `**` in every case that makes sense.

tcclevenger:
@ambrad @jgfouca Here is what I have found out: All `bfb_square()` calls work and give identical output to the master version. But If I include just one `bfb_cube/quad()` call, the tester fails with quite large differences in many different variables..  Not really sure what is happening, let me know if you have any guesses? I'll continue to find more info.

ambrad:
Would you summarize the level of difference in the build having the most difference?

components/cam/src/physics/cam/shoc.F90
```diff
@@ -1751,10 +1751,10 @@ subroutine f0_to_f5_diag_third_shoc_moment(&
   wsec_diff     = w_sec_kc    - w_sec
   tke_diff      = tke_kc      - tke
 
-  f0 = thedz2 * bet2**3 * iso**4 * wthl_sec * &
+  f0 = thedz2 * bet2**3 * bfb_quad(iso) * wthl_sec * &
```
ambrad, line 1754: `**3` -> bfb_cube here

tcclevenger:
`host_dse` is the largest, a typical line:
`host_dse > tol 431 times. Max rel diff= 0.0968701 normalized by ref impl val=360745.`

ambrad:
Great, thanks. That's in an SP build, so that's not terrible, but it's also not great. It's also not the first time we've seen that level of discrepancy in the SP build.

1. I recommend you do a no-pack single-precision build on your dev workstation and see if you can reproduce that AT output (approximately); I believe you should be able to. Once you can, it would be useful to find exactly what single change leads to such a large discrepancy.
2. It would also be useful to look at the host_dse error in double precision. Is it ~1e-8?

tcclevenger:
> Great, thanks. That's in an SP build, so that's not terrible, but it's also not great. It's also not the first time we've seen that level of discrepancy in the SP build.

On Blake, I'm reproducing the error in both single and double precision, and the max diff is the same magnitude.

> 1. I recommend you do a no-pack single-precision build on your dev workstation and see if you can reproduce that AT output (approximately); I believe you should be able to. Once you can, it would be useful to find exactly what single change leads to such a large discrepancy.

I can't seem to reproduce this on my laptop in single or double precision, even with all pack/thread size set to 1.

ambrad:
> On Blake, I'm reproducing the error in both single and double precision, and the max diff is the same magnitude.

What are the build details? And can you isolate it to a single change on Blake?

tcclevenger:
Blake: here is a portion of cmake output setup:
```
SCREAM_DOUBLE_PRECISION: OFF
SCREAM_MIMIC_GPU: TRUE
SCREAM_PACK_SIZE: 16
SCREAM_SMALL_PACK_SIZE: 16
SCREAM_POSSIBLY_NO_PACK_SIZE: 16
```

On Blake, I isolated each of the (x**y), keeping everything as is on master and changing only one. For the last (150) timestep of the test: 
`(x**2) -> bfb_square()`: for each instance (and when all are bfb), the test passes with `tol=0`.
`(x**3) -> bfb_cube()`: for each instance, `host_dse max rel diff ~ 0.1`, all others at most `~ 1e-4`.
`(x**4) -> bfb_quad()`: for each instance, `host_dse max rel diff ~ 0.1`, and most others are actually higher, e.g., `shoc_ql2~178`, `vw_sec~1`

So it seems that all `bfb_cube/quad` calls result if large diffs.

ambrad:
Ok. But what single line modification (not **3 -> bfb_cube, but a single instance of it) causes such a large diff? If you can find that, then I'd recommend printing out the value of that line before and after the mod, i.e., compute each form and print out the values side by side. If each particular difference is small, then we can conclude that sensitivity elsewhere, such as in a threshold, is the issue for the overall large difference. Peter Bogenschutz would likely be interested in this finding.

tcclevenger:
> Ok. But what single line modification (not **3 -> bfb_cube, but a single instance of it) causes such a large diff? 

Sorry, I think I was confusing. I applied each of the instances of (x**3) individually (i.e., where it was the only change), and each lead to around the same magnitude of errors.

tcclevenger:
@jeff-cohere @jgfouca @ambrad Ok, this is ready to be reviewed. The tester fails/will fail with what appears to be quite large diffs when running on CPU.Talking with @ambrad, it seems as if this is not a bug and just a consequence of very small changes over many timesteps. 

To summarize the discussion above, changing any `x**2 -> bfb_square()` does not cause any diff in `./shoc_run_and_cmp`, however changing just one of the `x**3 -> bfb_cube()` or `x**4 -> bfb_quad()` causes diffs of the same magnitude as we see in the failed tests here. As a test, in line 1754 of `shoc.F90`, I compare the difference in `f0` from the original (using all `x**y` calls) and `f0` using `bfb_cube(bet2)`. This represents the diff we expect to come from one change to bfb_cube. Here are some results running `./shoc_run_and_cmp`:
```
abs(f0(original) - f0(using bfb_cube(bet2)) = 
 diff:   0.000000000000000E+000
 diff:   1.734723475976807E-018
 diff:   1.084202172485504E-019
 diff:   0.000000000000000E+000
 diff:   1.734723475976807E-018
 diff:   1.084202172485504E-019
 diff:   0.000000000000000E+000
 diff:   1.734723475976807E-018
 diff:   0.000000000000000E+000
 diff:   0.000000000000000E+000
.
.
.
 diff:   2.220446049250313E-016
 diff:   0.000000000000000E+000
 diff:   0.000000000000000E+000
 diff:   0.000000000000000E+000
 diff:   0.000000000000000E+000
 diff:   3.469446951953614E-018
 diff:   0.000000000000000E+000
 diff:   2.168404344971009E-019
 diff:   0.000000000000000E+000
 diff:   0.000000000000000E+000
```
Sumary: No single change is ever larger than `1e-15`, implying that `bfb_cube` is implemented correctly, and the change in output is as expected. @bogensch we would like your input as to if this is acceptable. All shoc property tests still pass.

jeff-cohere (approved):

ambrad:
Hi @tcclevenger, it looks like you're showing output for `abs(f0(original) - f0(using bfb_cube(bet2))`, which is an _absolute_ difference, but we need to know the _relative_ difference, which is something like `abs(f0(original) - f0(using bfb_cube(bet2))/(f0(original) == 0 ? 1 : abs(f0(original)))`. We'd like to see ~1e-16 for double precision and ~1e-8 for single.

tcclevenger:
@ambrad Ah yes, of course..

Single precision (grepping though all the results, all <3e-7)
```
abs(f0(original) - f0(using bfb_cube(bet2))/abs(f0(original) =  
 diff:   0.0000000E+00
 diff:   1.1366993E-07
 diff:   1.0463962E-07
 diff:   0.0000000E+00
 diff:   1.1366993E-07
 diff:   1.0463962E-07
 diff:   0.0000000E+00
 diff:   1.1366993E-07
 diff:   0.0000000E+00
 diff:   0.0000000E+00
.
.
.
 diff:   0.0000000E+00
 diff:   0.0000000E+00
 diff:   1.7644187E-07
 diff:   1.9924607E-07
 diff:   0.0000000E+00
 diff:   0.0000000E+00
 diff:   0.0000000E+00
 diff:   0.0000000E+00
 diff:   1.4482434E-07
 diff:   0.0000000E+00
```

Double precision (grepping though all the results, all <6e-16)
```
abs(f0(original) - f0(using bfb_cube(bet2))/abs(f0(original) =  
 diff:   2.845535090253680E-016
 diff:   0.000000000000000E+000
 diff:   1.950133928080147E-016
 diff:   0.000000000000000E+000
 diff:   0.000000000000000E+000
 diff:   1.950133928080147E-016
 diff:   0.000000000000000E+000
 diff:   0.000000000000000E+000
 diff:   0.000000000000000E+000
 diff:   0.000000000000000E+000
.
.
.
 diff:   0.000000000000000E+000
 diff:   4.394491755338432E-016
 diff:   0.000000000000000E+000
 diff:   0.000000000000000E+000
 diff:   0.000000000000000E+000
 diff:   0.000000000000000E+000
 diff:   0.000000000000000E+000
 diff:   1.734756739432805E-016
 diff:   0.000000000000000E+000
 diff:   0.000000000000000E+000
```

ambrad (approved):

PR #615, opened 2020-09-28
port shoc_energy_integrals

tcclevenger:
Port shoc_energy_integrals.

### DIFF
--- a/components/scream/src/physics/shoc/shoc_energy_integrals_impl.hpp
+++ b/components/scream/src/physics/shoc/shoc_energy_integrals_impl.hpp
@@ -1,0 +1,64 @@
+#ifndef SHOC_ENERGY_INTEGRALS_IMPL_HPP
+#define SHOC_ENERGY_INTEGRALS_IMPL_HPP
+
+#include "shoc_functions.hpp" // for ETI only but harmless for GPU
+
+namespace scream {
+namespace shoc {
+
+template<typename S, typename D>
+KOKKOS_FUNCTION
+void Functions<S,D>
+::shoc_energy_integrals(
+  const MemberType&            team,
+  const Int&                   nlev,
+  const uview_1d<const Spack>& host_dse,
+  const uview_1d<const Spack>& pdel,
+  const uview_1d<const Spack>& rtm,
+  const uview_1d<const Spack>& rcm,
+  const uview_1d<const Spack>& u_wind,
+  const uview_1d<const Spack>& v_wind,
+  Scalar&                      se_int,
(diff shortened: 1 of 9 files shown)


tcclevenger:
Failing tests will be resolved from https://github.com/E3SM-Project/EKAT/pull/56. Will need to update EKAT in scream.

components/scream/src/physics/shoc/shoc_energy_integrals_impl.hpp
```diff
@@ -0,0 +1,64 @@
...
+  Scalar&                      se_int,
+  Scalar&                      ke_int,
+  Scalar&                      wv_int,
+  Scalar&                      wl_int)
+{
+  using ExeSpaceUtils = ekat::ExeSpaceUtils<typename KT::ExeSpace>;
+  const auto ggr = C::gravit;
+
+  // Use serialized version of view_reduction below if doing BFB testing
+#ifdef EKAT_DEFAULT_BFB
```
tcclevenger, line 30: @bartgol Can you check my use of `EKAT_DEFAULT_BFB`. It seems the default scream build sets this value as true (since the tests are passing, and I have verified that they fail when I manually set to false). Is this acceptable, or should there be a `SCREAM_BFB` flag which sets `EKAT_BFB`?

singhbalwinder: This makes much more sense to me now. I probably didn't notice this `#ifdef earlier`

singhbalwinder (approved):
Looks great! Thanks @tcclevenger !

components/scream/src/physics/shoc/shoc_energy_integrals_impl.hpp
tcclevenger: This actually may be unnecessary since the default templates for `view_reduction` uses `EKAT_DEFAULT_BFB`. @bartgol are the testers compiled with `EKAT_DEFAULT_BFB=TRUE`?

ambrad: Once you figure out what cpp symbol is needed, I suggest you make a scream-physics top-level util that wraps `ekat::view_reduction` that passes the correct value of `Serialize`. That way, the app code doesn't get cluttered with this detail in each use. You can merge this PR and make the wrapper in a separate PR if you want, so I'll approve this PR now.

ambrad (approved):

jeff-cohere (approved):

PR #621, opened 2020-09-29
Added unit tests for grid classes.

### LINKED ISSUE #560: [AD] Create unit tests for grids

jeff-cohere:
This adds a unit test suite for subclasses of `AbstractGrid`, for use with the AD.

Closes #560

### DIFF
--- a/components/scream/src/share/tests/grid_tests.cpp
+++ b/components/scream/src/share/tests/grid_tests.cpp
@@ -12,8 +12,6 @@ namespace {
 
 using namespace scream;
 using namespace scream::ShortFieldTagsNames;
-using device_type = DefaultDevice;
-using kokkos_types = KokkosTypes<device_type>;
 
 TEST_CASE("simple_grid", "") {
 
@@ -27,9 +25,11 @@ TEST_CASE("simple_grid", "") {
   REQUIRE(grid.get_num_local_dofs() == num_cols / num_procs);
 
   auto lid_to_idx = grid.get_lid_to_idx_map();
+  auto host_lid_to_idx = Kokkos::create_mirror_view(lid_to_idx);
+  Kokkos::deep_copy(host_lid_to_idx, lid_to_idx);
   for (int i = 0; i < grid.get_num_local_dofs(); ++i) {
-    REQUIRE(lid_to_idx.extent_int(1) == 1);
-    REQUIRE(i == lid_to_idx(i, 0));
+    REQUIRE(host_lid_to_idx.extent_int(1) == 1);
+    REQUIRE(i == host_lid_to_idx(i, 0));
   }
 
   auto layout = grid.get_native_dof_layout();
@@ -51,17 +51,24 @@ TEST_CASE("se_cell_based_grid", "") {
 
   // Set up the degrees of freedom.
   SEGrid::dofs_list_type dofs("", num_elems*num_gp*num_gp);
+  auto host_dofs = Kokkos::create_mirror_view(dofs);
   SEGrid::lid_to_idx_map_type dofs_map("", num_elems*num_gp*num_gp, 3);
+  auto host_dofs_map = Kokkos::create_mirror_view(dofs_map);
   for (int ie = 0; ie < num_elems; ++ie) {
     for (int igp = 0; igp < num_gp; ++igp) {
       for (int jgp = 0; jgp < num_gp; ++jgp) {
         int idof = ie*num_gp*num_gp + igp*num_gp + jgp;
-        dofs_map(idof, 0) = ie;
-        dofs_map(idof, 1) = igp;
-        dofs_map(idof, 2) = jgp;
+        host_dofs(idof) = idof;
+        host_dofs_map(idof, 0) = ie;
+        host_dofs_map(idof, 1) = igp;
+        host_dofs_map(idof, 2) = jgp;
       }
     }
   }
+
+  // Move the data to the device and set the DOFs.
+  Kokkos::deep_copy(dofs, host_dofs);
+  Kokkos::deep_copy(dofs_map, host_dofs_map);
   grid.set_dofs(dofs, dofs_map);
 
   REQUIRE(grid.get_num_local_dofs() == num_elems*num_gp*num_gp);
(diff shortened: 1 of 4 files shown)


jeff-cohere:
Hey @bartgol , it looks like the build for this PR failed on Weaver. Can you send me the full log for the build?

bartgol (changes requested):
This looks good. There's one more thing I think we should test, and that is the ability to retrieve the correct dof when using native indices. Most importantly in the SE_Cell_Based grid (the dynamics grid), but might as well do it for all grids. I would do this:

- set the dof gids; for SE_Cell_Based, it doesn't matter if the shared dofs across elements have different gid, we're not testing how the gids are computed here. For SE_Cell_Based, we could set the dof as ie*NP*NP+igp*NP+jgp, while for the others, we can set dof i to be like `i+10` (offset to make sure local id doesn't match global id, for extra safety and generality).
- test lid to idx (all grids): MeshFree and SE_Node_Based should be identity maps, so simply check that, if `v=get_lid_to_idx(i)`, then `v.extent_int(1)==1`, and `v(i)==i`. For SE_Cell_Based, we need to check that `v.extent_int(1)==3`, and that `v(i,0)==ie, v(i,1)==igp, v(i,2)==jgp`, where ie, igp, jgp need to be computed with modular arithmetic from i.
- test idx to lid consistency (only SE_Cell_Based): retrieve the lid via the view returned by `get_idx_to_lid_map()`, and check that `get_dofs_gids(lid)==(ie*NP*NP+igp*NP+jgp)` (or whatever gid you assigned to (ie,igp,jgp)).

bartgol:
@jeff-cohere The error seems to be this:

```
grid_tests.cpp(79): error: device code does not support exception handling
```

jeff-cohere:
> @jeff-cohere The error seems to be this:
> 
> ```
> grid_tests.cpp(79): error: device code does not support exception handling
> ```

Thanks! That line is a call to `SEGrid::create_elgpgp_to_lid_map()`. Should that call just be wrapped in a `try` block, then, to prevent the device from needing to handle exceptions? I see references to `REQUIRE_THROWS` in other tests, but the way it's being called shouldn't actually generate a thrown exception.

jeff-cohere:
Also, I've implemented the index checking for the cell-based grid in a way that checks the reverse mapping, which I believe should be equivalent to the test you propose. But I'm happy to do it the way you suggest in addition to that if you think it makes a difference.

bartgol:
> > @jeff-cohere The error seems to be this:
> > ```
> > grid_tests.cpp(79): error: device code does not support exception handling
> > ```
> 
> Thanks! That line is a call to `SEGrid::create_elgpgp_to_lid_map()`. Should that call just be wrapped in a `try` block, then, to prevent the device from needing to handle exceptions? I see references to `REQUIRE_THROWS` in other tests, but the way it's being called shouldn't actually generate a thrown exception.

I checked out the branch, and line 79 is
```
   REQUIRE(i == idx_to_lid(ie, igp, jgp));
```
which is inside a parallel for. Perhaps you haven't pushed the latest version?

Re: REQUIRE_THROWS. That's a macro to use if you want to test that bad input generate an exception. I used it a few times to make sure all (well, some) of the `EKAT_REQUIRE_MSG(...)` inside various functions/methods were properly triggered. Of course, this only works on host, since CUDA doesn't handle exceptions.

jeff-cohere:
Yup, sorry. I noticed that when I rewound my latest changes (which then I lost because of some sort of murky git problem behind the scenes, possibly a bug encountered only on huge projects with lots of submodules). Reimplementing test additions now.

components/scream/src/share/tests/grid_tests.cpp
```diff
@@ -0,0 +1,108 @@
...
+        dofs_map(idof, 2) = jgp;
+      }
+    }
+  }
+  grid.set_dofs(dofs, dofs_map);
+
+  REQUIRE(grid.get_num_local_dofs() == num_elems*num_gp*num_gp);
+
+  auto layout = grid.get_native_dof_layout();
+  REQUIRE(layout.tags().size() == 3);
```
bartgol, line 64: No need to change this, but as a reference, `layout.rank()` would also return what you want.

jeff-cohere: Considering that this test is probably the only documentation for the grid layout tags, I'm inclined to keep it. :-)

components/scream/src/share/tests/grid_tests.cpp
```diff
@@ -0,0 +1,123 @@
...
+  SimpleGrid grid("my_grid", num_cols, num_levels, comm);
+  REQUIRE(grid.type() == GridType::MeshFree);
+  REQUIRE(grid.name() == "my_grid");
+  REQUIRE(grid.get_num_vertical_levels() == num_levels);
+  REQUIRE(grid.get_num_local_dofs() == num_cols / num_procs);
+
+  auto lid_to_idx = grid.get_lid_to_idx_map();
+  for (int i = 0; i < grid.get_num_local_dofs(); ++i) {
+    REQUIRE(lid_to_idx.extent_int(1) == 1);
+    REQUIRE(i == lid_to_idx(i, 0));
```
bartgol, line 32: I suspect this will fail on GPU, since lid_to_idx is a device view. You need to deep copy to host first.

components/scream/src/share/tests/grid_tests.cpp
```diff
@@ -0,0 +1,123 @@
...
+  REQUIRE(grid.get_num_local_dofs() == 0);
+
+  // Set up the degrees of freedom.
+  SEGrid::dofs_list_type dofs("", num_elems*num_gp*num_gp);
+  SEGrid::lid_to_idx_map_type dofs_map("", num_elems*num_gp*num_gp, 3);
+  for (int ie = 0; ie < num_elems; ++ie) {
+    for (int igp = 0; igp < num_gp; ++igp) {
+      for (int jgp = 0; jgp < num_gp; ++jgp) {
+        int idof = ie*num_gp*num_gp + igp*num_gp + jgp;
+        dofs_map(idof, 0) = ie;
```
bartgol, line 59: Same problem here: these are device views.

components/scream/src/share/tests/grid_tests.cpp
```diff
@@ -0,0 +1,123 @@
...
+              num_elems, num_gp, num_levels);
+  REQUIRE(grid.type() == GridType::SE_NodeBased);
+  REQUIRE(grid.name() == "se_node_grid");
+  REQUIRE(grid.get_num_vertical_levels() == num_levels);
+
+  // Set the degrees of freedom.
+  SEGrid::dofs_list_type dofs("", num_cols);
+  SEGrid::lid_to_idx_map_type dofs_map("", num_cols, 1);
+  for (int i = 0; i < num_cols; ++i) {
+    dofs(i) = i + 10;
```
bartgol, line 105: Device view.

components/scream/src/share/tests/grid_tests.cpp
```diff
@@ -0,0 +1,123 @@
...
+    dofs_map(i, 0) = i;
+  }
+  grid.set_dofs(dofs, dofs_map);
+  REQUIRE(grid.get_num_local_dofs() == num_cols);
+
+  auto dofs_gids = grid.get_dofs_gids();
+  auto lid_to_idx = grid.get_lid_to_idx_map();
+  for (int i = 0; i < dofs_gids.extent_int(0); ++i) {
+    REQUIRE(lid_to_idx.extent_int(1) == 1);
+    REQUIRE(i == lid_to_idx(i, 0));
```
bartgol, line 115: Device view.

bartgol (changes requested):
There's some device views that are accessed on host. It's annoying, but you have to fill host views, deep copy to device, and to test deep copy to host, then test the host mirror.

jeff-cohere:
Ah, okay. So all views created by the grid live on the device. I have to work for a bit on some other stuff, but I'll correct these oversights when I can.

jeff-cohere:
Hey @bartgol ,

I just saw the `TODO: template class on device type` comment in the header for `AbstractGrid`, which is what I was thinking about doing anyway, given the amount of noise the host <-> device copies are adding to the test. Do you mind if I go ahead and add a `DeviceType` template parameter to the grid classes and make it default to `DefaultDevice`?

bartgol:
Are you going to keep it limited to AbstractGrid? Meaning, GridsManager, AbstractRemapper, AtmosphereProcess, and others will continue to use exclusively default device versions, right?

I'm split on this issue, cause I wonder if it will generate confusion (offering host and device versions hints that we might need/use both). In principle, I try to avoid adding a library feature only to keep tests files smaller. But on the other hand, it's not really changing much for the downstream classes anyways.

I'll leave this up to you, I'm fine either way I guess.

jeff-cohere:
That's a good point. Here we're weighing the value of writing intelligible tests against the generation of additional code that exists essentially for testing. Maybe I'll just stick to mysterious mirrors and deep copies for now.

bartgol (approved):

PR #624, opened 2020-09-29
update ekat submodule

tcclevenger:
@bartgol I went to `external/ekat`, checked out master, pulled the latest update, and committed the changes in this scream branch. Let me know if this is not the correct way to update.


bartgol (approved):
Looks good to me.

PR #625, opened 2020-09-30
P3 Bugfix: protect tau_r from div-by-zero

PeterCaldwell:
tau_r=1/epsr was calculated even where qr_incld<qsmall => epsr was assigned to zero. tau_r wasn't actually used in this case, so I just moved the tau_r calculation within a qr_incld<qsmall branch (in both C++ and F90). 

This div-by-zero problem was never triggered by scream tests because epsr was always assumed to be a reasonable value. Balwinderfound this problem with e3sm-style tests in debug mode on compy. I was unable to reproduce them on quartz, but it makes conceptual sense that the original impl could cause issues.


singhbalwinder (approved):
Looks good!
I ran the whole test suite on Compy and all tests pass with this branch.

PR #628, opened 2020-09-30
Multiple groups for fields at field repo registration

AaronDonahue:
Make it possible to assign a field to more than one group when registering with the field repository.

Update "field" test to test the new functionality.

### DIFF
--- a/components/scream/src/share/field/field_repository.hpp
+++ b/components/scream/src/share/field/field_repository.hpp
@@ -140,10 +140,7 @@ template<typename ScalarType, typename Device>
 template<typename RequestedValueType>
 void FieldRepository<ScalarType,Device>::
 register_field (const identifier_type& id, const std::initializer_list<std::string>& groups_names) {
-  for (auto name : groups_names)
-  {
-    register_field<RequestedValueType>(id,name);
-  }
+  register_field<RequestedValueType>(id,std::set<std::string>(groups_names));
 }
 
 template<typename ScalarType, typename Device>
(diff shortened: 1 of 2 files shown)


components/scream/src/share/field/field_repository.hpp
```diff
@@ -133,6 +136,16 @@ register_field (const identifier_type& id, const std::string& group_name) {
   register_field<RequestedValueType>(id,group_name_set);
 }
 
+template<typename ScalarType, typename Device>
+template<typename RequestedValueType>
+void FieldRepository<ScalarType,Device>::
+register_field (const identifier_type& id, const std::initializer_list<std::string>& groups_names) {
+  for (auto name : groups_names)
```
bartgol, line 143: Since `register_field<T>(id,name)` in turn calls the overload with `std::set`, you're better off creating the set with all names from the initializer list:
```
   register_field<RequestedValueType>(id,std::set<std::string>(groups_name));
```

bartgol (changes requested):
Looks good. I only have one comment/request.

bartgol:
Note to FM devs: it would be nice to have only one register_field that takes a container of group names, rather than two. In fact, inside `register_field`, we only require to be able to iterate over the `group_names` input. Any std container would do.

However, templating over the container type (with a _second_ template arg) would require a default for the container type when trying to pass a non-default requested value type. Swapping the template args order won't make things better, since one would have to specify the container type every time a non-default requested value type is needed.

A solution would be to not give a default type for RequestedValueType, but that seems suboptimal.

For now, I think leaving two hard-coded overloads (initializer_list and set) is good enough.

components/scream/src/share/field/field_repository.hpp
AaronDonahue: Done.

bartgol (approved):

PR #629, opened 2020-09-30
Default Autotester EKAT BFB

tcclevenger:
Default `EKAT_DEFAULT_BFB=ON` for Autotester and simplify `ekat::view_reduction()`.

### DIFF
--- a/components/scream/scripts/test_all_scream.py
+++ b/components/scream/scripts/test_all_scream.py
@@ -53,14 +53,15 @@ class TestAllScream(object):
         expect(not (self._baseline_ref and self._baseline_dir),
                "Makes no sense to specify a baseline generation commit if using pre-existing baselines ")
 
-        self._custom_cmake_opts.append("EKAT_DEFAULT_BFB=ON")
-
-        self._tests_cmake_args = {"dbg" : [("CMAKE_BUILD_TYPE", "Debug")],
+        self._tests_cmake_args = {"dbg" : [("CMAKE_BUILD_TYPE", "Debug"),
+                                           ("EKAT_DEFAULT_BFB", "ON")],
                                   "sp"  : [("CMAKE_BUILD_TYPE", "Debug"),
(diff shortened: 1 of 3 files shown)


bartgol:
I'm actually having second thoughts on whether this option should be in jenkins_common.sh. I think it's best to put it inside `test_all_scream.py`, as a common config option for all builds.

The reason is that, in the future, we _might_ add a Release build to our test-all-scream builds, with end-to-end runs against baselines, and either no unit tests or tolerance-based unit tests. In that case, we'd want to test _exactly_ what runs in production, so without the BFB behaviors. Adding the option to jenkins_common.sh would make _all_ the builds have such option, but adding it manually to the three builds, would allow a fourth build (the Release one) to be run without such option.

@jgfouca what do you think? Am I overthinking this?

tcclevenger:
@bartgol @jgfouca I move the `EKAT_DEFAULT_BFB=ON` to `test_all_scream.py`. Let me know your thoughts. The other option is to put it in `jenkins_common.sh` (was 1st commit in the PR).

components/scream/scripts/test_all_scream.py
```diff
@@ -53,6 +53,8 @@ def __init__(self, cxx, kokkos=None, submit=False, parallel=False, fast_fail=Fal
         expect(not (self._baseline_ref and self._baseline_dir),
                "Makes no sense to specify a baseline generation commit if using pre-existing baselines ")
 
+        self._custom_cmake_opts.append("EKAT_DEFAULT_BFB=ON")
```
bartgol, line 56: I would not add it to the custom cmake opts, cause those are applied to all builds. I would instead add this to all dict entries below.

jgfouca (approved):
I approve pending @bartgol 's cmake_opts comment being addressed.

tcclevenger:
@jgfouca @bartgol All tests pass, let me know if you have any other comments. If not, I'll need one more approval.

bartgol (approved):

PR #630, opened 2020-09-30
more variable names modified

xyuan:
more variable names modified

[BFB]


jgfouca (approved):

components/scream/src/physics/p3/p3_calc_rime_density_impl.hpp
```diff
@@ -46,7 +46,7 @@ ::calc_rime_density(
       // Use mass-weighted mean size
       Spack D_c = (4 + mu_c) / lamc;
       Spack V_impact = abs(vtrmi1-Vt_qc);
-      Spack inv_Tc = 1/min(sp(-0.001), T_atm-T_zerodegc);
+      Spack inv_Tc = 1/min(sp(-0.001), T_atm-t_zerodegc);
       Spack Ri = max(1, min(sp(-0.5e+6) * D_c * V_impact * inv_Tc, 12));
```
PeterCaldwell, line 50: My goal was to have T always have the same case. You had "t_atm" and "T_zerodegc" in F90 in your previous PR, which is why I asked you to switch to "t_zerodegc". Since C++ has "T_atm", I'd like to keep T_Zerodegc as is. Same for other T_ variables.

xyuan:
It is already merged, so I will create another PR to do the task.

PeterCaldwell:
Ok, sounds good.

PR #232, opened 2020-02-20
SHOC refactor (PBL depth, energy calculation, substep)

bogensch:
This PR:
1) Moves the PBL depth diagnostic calculation into shoc.F90. 
2) Moves the energy fixer calculation from the SHOC interface to shoc.F90
3) Moves the SHOC substep loop from SHOC interface into shoc.F90

NOTE that this PR also removes all code in shoc_intr.F90 related to turbulent mountain stress (TMS).  do_tms always = .false. for E3SM/SCREAM runs.  This was legacy code from CAM5 days that got wrapped into shoc_intr.F90 when using clubb_intr.F90 as a template, thus is not needed.  

In addition, a bug where SHOC was not using the von karman constant consistent with the host model has also been fixed.  

This PR is not b4b (but non climate changing) because of step 1), where the PBL calculation was moved and due to order of operations this routine sees a slightly different state.  The von karman bug also contributes to this PR being non b4b.  Incremental implementations of steps 2) and 3) provided b4b answers with the last iteration, as expected.

PR #631, opened 2020-09-30
SHOC bug fix for theta_v definition in PBL height routine

bogensch:
Definition of virtual potential temperature was erroneously defined in the PBL height routine in SHOC.  This has been fixed in the F90 and C++ versions of the code.  shoc_run_and_cmp is an expected fail, all other tests should pass.  

This bug was discovered whilst developing property tests for the subroutine pblintd_init_pot (PR forthcoming).  Tracing back, this bug was introduced in PR #232 but was masked by the fact that 1) at that point the PBL height was not used interactively by SHOC (only a diagnostic) and 2) that PR addressed a separate answer changing bug.


bogensch:
Tests are passing and failing as expected.

xyuan (approved):

bogensch:
@PeterCaldwell this PR has been approved and failing/passing tests as expected.  I consider it ready to merge.

PR #632, opened 2020-09-30
use bfb function in lby_cond subroutine

xyuan:
use bfb function in lby_cond subroutine

[BFB]


xyuan:
@ambrad @jgfouca @jeff-cohere Could we let this PR merge to master? This is simple bfb_* function for fortran subroutine

ambrad (approved):

PR #633, opened 2020-10-01
Jgfouca/convert linear interp

jgfouca:
Converted linear interp. I'm quite happy with the implementation, it's actually smaller than the fortran impl because I was able to get it all into the same loop/||4.

Added a new random linear interp property test.


components/cam/src/physics/cam/shoc.F90
```diff
@@ -4248,7 +4260,7 @@ subroutine linear_interp(x1,x2,y1,y2,km1,km2,ncol,minthresh)
        end do
        k2 = km2
        do i = 1,ncol
-          y2(i,k2) = y1(i,km1) + (y1(i,km1)-y1(i,km1-1))*(x2(i,k2)-x1(i,km1))/(x1(i,km1)-x1(i,km1-1))
+          y2(i,k2) = y1(i,km1-1) + (y1(i,km1)-y1(i,km1-1))*(x2(i,k2)-x1(i,km1-1))/(x1(i,km1)-x1(i,km1-1))
```
jgfouca, line 4263: Note, I tweaked the f90 linear interp impl to be more cxx friendly. This change could result in roundoff-level changes, but they were small enough that they did not impact run_and_cmp, at least on mappy.

components/scream/src/physics/shoc/shoc_linear_interp_impl.hpp
```diff
@@ -13,10 +13,50 @@ namespace shoc {
...
+      ekat::index_and_shift<-1>(sy1, indx_pack, y1, y1s);
+      y2(k2) = y1s + (y1-y1s)*(x2(k2)-x1s)/(x1-x1s);
+    });
+  }
+  else if (km2 == km1+1) {
+
+    Kokkos::parallel_for(Kokkos::TeamThreadRange(team, km2_pack), [&] (const Int& k2) {
+      Spack x1, x1s, y1, y1s; // s->-1 shift
+      auto indx_pack = ekat::range<IntSmallPack>(k2*Spack::n);
+      indx_pack.set(indx_pack < 1, 1); // special shift for 0 boundary case
```
ambrad, line 44: I recommend you separate out the top and bottom from the interior to avoid this special casing in the interior.

ambrad (approved):
Very nice! Very cleanly vectorized code, and excellent new tests.

I left one suggestion you can optionally act on, either here or in another PR.

bogensch (approved):
nice!

jgfouca:
@bogensch , @ambrad, Apologies, I violated our policy of getting test-all-scream running before adding reviewers. I did run the full suite a couple commits prior, but neglected to re-run it before I submitted.

I had to increase the error margin for single-precision in the new linear_interp property test to an alarmingly high value. I confirmed two things, 1) that the same high-ish error margins existed prior to my change to shoc.F90 and 2) the high-ish error margins appear to be legitimately related to round-off amplification and not a real problem. In nearly all checks, SP passed with much tighter margins, there were just certain cases due to randomness (adjacent x1 and x2 points being very close to each other) that appeared to greatly amplify round-off changes.

ambrad (approved):

PR #636, opened 2020-10-01
porting shoc_diag_second_mom_lbycond

xyuan:
porting shoc_diag_second_mom_lbycond

[BFB]

### DIFF
--- a/components/scream/src/physics/shoc/shoc_diag_second_moments_lbycond_impl.hpp
+++ b/components/scream/src/physics/shoc/shoc_diag_second_moments_lbycond_impl.hpp
@@ -27,7 +27,7 @@ void Functions<S,D>
   const auto ufmin = 0.01;
 
   auto uf = sqrt(ustar2+sp(0.3)*wstar*wstar);
-  uf = max(ufmin,uf);
+  uf = std::max(ufmin,uf);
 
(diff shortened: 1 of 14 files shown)


components/scream/src/physics/shoc/shoc_diag_second_moments_lbycond_impl.hpp
```diff
@@ -0,0 +1,48 @@
...
+  // for the SHOC parameterization.
+  // The thermodymnamic, tracer, and momentum fluxes are set
+  // to the surface fluxes for the host model, while the
+  // thermodynamic variances and covariances are computed
+  // according to that of Andre et al. 1978.
+
+  const auto ufmin = 0.01;
+
+  auto uf = std::sqrt(ustar2+sp(0.3)*wstar*wstar);
+  uf = std::max(ufmin,uf);
```
ambrad, line 30: This could be the cause of the segfault. `std::max` is undefined on device, or at least has been in past Cuda versions. Try `ekat::impl::max` instead.

xyuan: Yes, I will replace it with ekat::impl::max<Scalar> function.

PR #642, opened 2020-10-02
Beydoun/prescribed ccn option new pr

hassanbeydoun:
New PR for new prescribed CCN option.


PeterCaldwell:
About failing tests: 

1. I only see *warning* messages, but gmake stops with an *error*. This confuses me.
2. The warning messages I see are related to SHOC, which can't be this PR's fault since it only changes P3 things. Can someone check whether master also fails tests now?
3. This PR changes the variables passed to/from various subroutines. Doing so will break P3 unit tests because the C++ testing functions will call the F90 with the old set of input/output variables.

I'm in meetings this morning so can't offer more concrete suggestions.

xyuan:
@PeterCaldwell why this branch is in draft mode?

xyuan:
@PeterCaldwell Now, it passed the test on Summit, hopefully it will work on AT:. This version of code is latest, most likely, I forgot the conversion of xlf in Fortran code, but I did it this time.

xyuan:
@PeterCaldwell We may need to regenerate the baseline since we add one new variables nccn_prescribed, it is working with the baseline data that manually generated.

components/scream/src/physics/p3/scream_p3_interface.F90
```diff
@@ -180,6 +181,7 @@ subroutine p3_main_f90 (dtime,zi,pmid,dpres,ast,ni_activated,nc_nuceat_tend,q,FQ
 
     integer(kind=c_int) :: it, its, ite, kts, kte
     logical(kind=c_bool) :: do_predict_nc = .true.
+    logical(kind=c_bool) :: do_prescribed_CCN = .true.
```
PeterCaldwell, line 184: This is kind of a rumination to myself rather than something to change: I was thinking you'd just implement the function interfaces for the C++, but you actually implemented the full capability within these functions. We should make sure to unit test both do_prescribed_CCN=.true. and .false. in this case. Maybe you have - I haven't checked yet. I was also thinking that we can't get nccn_prescribed values in C++ yet since we read them from netcdf... but we do that in p3_interface which we don't run in our P3 unit tests. Instead, we hardcode values to pass to our functions... so we can easily test nccn_prescribed values now.

components/scream/src/physics/p3/tests/p3_ice_nucleation_unit_tests.cpp
```diff
@@ -26,29 +26,30 @@ struct UnitWrap::UnitTest<D>::TestIceNucleation {
     using KTH = KokkosTypes<HostDevice>;
 
     constexpr bool   do_predict_nc = false;
+    constexpr bool   do_prescribed_CCN = false;
     constexpr Scalar inv_dt           = 9.558E-04;
```
PeterCaldwell, line 30: As I was saying before, we should probably test both do_predict_nc is false and is true.

components/scream/src/physics/p3/tests/p3_main_unit_tests.cpp
```diff
@@ -56,11 +56,11 @@ static void run_bfb_p3_main_part1()
...
-    P3MainPart1Data(1,  72,    1,   72,    1, false,         1.800E+03),
-    P3MainPart1Data(1,  72,    1,   72,    1, true,          1.800E+03),
-    P3MainPart1Data(1,  72,   72,    1,   -1, false,         1.800E+03),
-    P3MainPart1Data(1,  72,   72,    1,   -1, true,          1.800E+03),
+    //            kts, kte, ktop, kbot, kdir, do_predict_nc, do_prescribed_CCN,       dt
+    P3MainPart1Data(1,  72,    1,   72,    1, false,          true,           1.800E+03),
+    P3MainPart1Data(1,  72,    1,   72,    1, true,           true,           1.800E+03),
+    P3MainPart1Data(1,  72,   72,    1,   -1, false,          false,          1.800E+03),
+    P3MainPart1Data(1,  72,   72,    1,   -1, true,           false,          1.800E+03),
   };
```
PeterCaldwell, line 64: @hassanbeydoun - I think that do_predict_nc should only be true when do_prescribed_CCN is false and vice versa. Do the columns for do_predict_nc and do_prescribed_CCN just above and on the right hand panel make sense to you?

components/scream/src/physics/p3/tests/p3_main_unit_tests.cpp
```diff
@@ -143,11 +143,11 @@ static void run_bfb_p3_main_part2()
...
-    P3MainPart2Data(1,  72,    1,   72,    1, false,         1.800E+03),
-    P3MainPart2Data(1,  72,    1,   72,    1, true,          1.800E+03),
-    P3MainPart2Data(1,  72,   72,    1,   -1, false,         1.800E+03),
-    P3MainPart2Data(1,  72,   72,    1,   -1, true,          1.800E+03),
+    //            kts, kte, ktop, kbot, kdir, do_predict_nc, do_prescribed_CCN,       dt
+    P3MainPart2Data(1,  72,    1,   72,    1, false,         true,        1.800E+03),
+    P3MainPart2Data(1,  72,    1,   72,    1, true,          true,        1.800E+03),
+    P3MainPart2Data(1,  72,   72,    1,   -1, false,         false,       1.800E+03),
+    P3MainPart2Data(1,  72,   72,    1,   -1, true,          false,       1.800E+03),
   };
```
PeterCaldwell, line 151: Again, do these combinations of logicals make sense?

components/scream/src/physics/p3/tests/p3_main_unit_tests.cpp
```diff
@@ -328,9 +328,9 @@ static void run_bfb_p3_main_part3()
 static void run_bfb_p3_main()
 {
   P3MainData isds_fortran[] = {
-    //      its, ite, kts, kte,   it,        dt, do_predict_nc
-    P3MainData(1, 10,   1,  72,    1, 1.800E+03, false),
-    P3MainData(1, 10,   1,  72,    1, 1.800E+03, true),
+    //      its, ite, kts, kte,   it,        dt, do_predict_nc, do_prescribed_CCN
+    P3MainData(1, 10,   1,  72,    1, 1.800E+03, false, true),
+    P3MainData(1, 10,   1,  72,    1, 1.800E+03, true,  false),
   };
```
PeterCaldwell, line 334: Again, do these combos of logicals make sense?

components/scream/src/physics/p3/tests/p3_main_unit_tests.cpp
```diff
@@ -340,6 +340,7 @@ static void run_bfb_p3_main()
         {d.pres           , {1.00000000E+02 , 9.87111111E+04}},
         {d.dz             , {1.22776609E+02 , 3.49039167E+04}},
         {d.nc_nuceat_tend , {0              , 0}},
+        {d.nccn_prescribed, {0              , 0}},
```
PeterCaldwell, line 343: I think nccn_prescribed needs to be positive if do_prescribed_CCN is true. @hassanbeydoun - what's a good number here?

components/scream/src/physics/p3/p3_functions_f90.hpp
```diff
@@ -1037,13 +1037,14 @@ void calculate_incloud_mixingratios_f(Real qc, Real qr, Real qi, Real qm, Real n
 
 struct P3MainPart1Data : public PhysicsTestData
 {
-  static constexpr size_t NUM_ARRAYS = 39;
+  static constexpr size_t NUM_ARRAYS = 40;
```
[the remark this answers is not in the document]
xyuan: We have the nccn_prescribed data, see below @PeterCaldwell 

struct P3MainPart1Data : public PhysicsTestData
{
  static constexpr size_t NUM_ARRAYS = 40;

  // Inputs
  Int kts, kte, kbot, ktop, kdir;
  bool do_predict_nc, do_prescribed_CCN;
  Real dt;
  Real* pres, *dpres, *dz, *nc_nuceat_tend, *exner, *inv_exner, *inv_cld_frac_l, *inv_cld_frac_i, *inv_cld_frac_r, *latent_heat_vapor, *latent_heat_sublim,
     *latent_heat_fusion, *nccn_prescribed;

xyuan (commented):
@PeterCaldwell I put these options to cover do_predict_nc=true and false, it doesn't mean physically useful, so @hassanbeydoun please check it and let me know what needs to be tested.

components/scream/src/physics/p3/tests/p3_unit_tests.cpp
```diff
@@ -817,67 +817,67 @@ struct UnitWrap::UnitTest<D>::TestP3UpdatePrognosticLiq
...
        -8.8982E+07},
 
       {4.2614E-08, 1.8780E+07, 2.9903E-11, 1.3178E+04, 8.3182E+03, 0.0000E+00, 0.0000E+00, 0.0000E+00, 5.6939E+04,
-       true      , 1.0151E+00, 1.0911E+00, 2.5010E+06, 1.8000E+03, 2.9119E+02, 2.7224E-03, 7.6760E-05,-6.5494E+09,-7.6760E-05,
+       true      , true      , 1.0151E+00, 1.0911E+00, 2.5010E+06, 1.8000E+03, 2.9119E+02, 2.7224E-03, 7.6760E-05,-6.5494E+09,-7.6760E-05,
        -9.0523E+07},
 
       {4.5927E-08, 1.8780E+07, 3.2867E-11, 1.3440E+04, 8.4833E+03, 0.0000E+00, 0.0000E+00, 0.0000E+00, 5.6317E+04,
-       true      , 1.0263E+00, 1.0970E+00, 2.5010E+06, 1.8000E+03, 2.9143E+02, 2.4202E-03, 8.2728E-05,-6.5044E+09,-8.2728E-05,
+       true      , true      , 1.0263E+00, 1.0970E+00, 2.5010E+06, 1.8000E+03, 2.9143E+02, 2.4202E-03, 8.2728E-05,-6.5044E+09,-8.2728E-05,
```
PeterCaldwell, line 880: It seems like always using true here means we're only testing 1 branch of update prognostic liquid? This is a problem with the original implementation, not an issue we're adding to in this PR.

components/scream/src/physics/p3/tests/p3_main_unit_tests.cpp
hassanbeydoun: @PeterCaldwell, yes it makes sense since the code never forces these two logicals be of opposite value.

components/scream/src/physics/p3/p3_functions_f90.hpp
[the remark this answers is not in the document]
PeterCaldwell: Ok, you're right. I got confused because you added 2 new variables but only increased NUM_ARRAYS by 1. I see now that is because do_prescribed_CCN isn't an array, but rather a scalar.

components/cam/src/physics/cam/micro_p3_interface.F90
```diff
@@ -1067,7 +1237,9 @@ subroutine micro_p3_tend(state, ptend, dtime, pbuf)
          vap_ice_exchange(its:ite,kts:kte),& ! OUT sum of vap-ice phase change tendencies
          qv_prev(its:ite,kts:kte),         & ! IN  qv at end of prev p3_main call   kg kg-1
          t_prev(its:ite,kts:kte),          & ! IN  t at end of prev p3_main call    K
-         col_location(its:ite,:3)          & ! IN column locations
+         col_location(its:ite,:3),          & ! IN column locations
+         do_prescribed_CCN,                 & !IN  .true. = prescribe CCN
+         nccn_prescribed(its:ite,kts:kte)    & !IN prescribed CCN concentration
          )
```
xyuan, line 1244: In your original commitment, there is no "elapsed_s" argument also, I doubt that it is another issue caused your problem.

components/scream/src/physics/p3/tests/p3_ice_nucleation_unit_tests.cpp
```diff
@@ -29,26 +29,26 @@ struct UnitWrap::UnitTest<D>::TestIceNucleation {
...
-      {2.452E+02, 0.950E+00, 0.974E+03, 8.221E+02, 4.952E-02, inv_dt, do_predict_nc},
-      {2.352E+02, 0.950E+00, 0.823E+03, 7.221E+02, 1.952E-02, inv_dt, do_predict_nc},
-      {2.252E+02, 0.950E+00, 0.723E+03, 6.221E+02, 9.952E-02, inv_dt, do_predict_nc},
-
-      {1.990E+02, 1.069E+00, 0.674E+03, 5.221E+01, 6.952E-01, inv_dt, do_predict_nc },
-      {2.952E+02, 1.069E+00, 0.574E+03, 4.221E+01, 3.952E-01, inv_dt, do_predict_nc },
-      {2.852E+02, 1.069E+00, 0.423E+03, 3.221E+01, 1.952E-01, inv_dt, do_predict_nc },
-      {2.702E+02, 1.069E+00, 0.323E+03, 2.221E+01, 9.952E-01, inv_dt, do_predict_nc }
+      // temp,    inv_rho,   ni,     ni_activated,      qv_supersat_i,      inv_dt, do_predict_nc, do_prescribed_CCN
+      {2.106E+02, 8.852E-01, 0.974E+04, 9.221E+03, 5.100E-01, inv_dt, do_predict_nc, true },
```
jgfouca, line 33: Remember, the CXX test bundles all these inputs into a single pack and calls the CXX function once, so you cannot toggle flags like this. You'll need to make multiple sets of IceNucleationDatas if you want to do that.

jgfouca (changes requested):

PeterCaldwell (approved):

PeterCaldwell:
PR fails p3_run_and_cmp because it adds new variables to the variables stored in baseline, causing child abort. Other tests pass. Hassan verifies that the model runs in F compset config. I will merge after one final round of AT.

PR #640, opened 2020-10-01
Port 2 shoc_length functions

tcclevenger:
Port `compute_brunt_shoc_length` and `check_length_scale_shoc_length` to C++

PR #647, opened 2020-10-05
bfb_sqrt

tcclevenger:
Use bfb_sqrt for shoc.F90 function.


tcclevenger:
Small non-bfb change needed for other PR: https://github.com/E3SM-Project/scream/pull/640.

ambrad (approved):

Issue #566, opened 2020-09-10
[AD] Abstract interface to check field properties


AaronDonahue:
Create an abstract interface 'FieldPropertyCheck', which can perform a check on a field, to verify it has a certain property, and, if the property is not satisfied and the user requests it, try to fix it. A property could be as simple as being in a certain fixed bound, or even be dependent on other fields (e.g., |f1 - f2|<C), or even involve differential operators (e.g., |div(f)| < eps)

-secondary task:  Add FieldPropertyCheck's to the Field structure.  Should not put them in the header, since they require knowledge of the view, so they really belong to the Field.

PR #650, opened 2020-10-06
[WIP] Implemented a reshaping test for fields of Reals vs Packs.

### LINKED ISSUE #561: [AD] Unit tests for field and field repository

jeff-cohere:
Hey @bartgol ,

Here's a stab at the test you suggested in #561 . Is this the kind of thing you had in mind? I haven't done host <-> device copying yet--I'll make sure that's done when I'm more confident that I haven't misinterpreted something.

Closes #561

### DIFF
--- a/components/scream/src/share/tests/field_tests.cpp
+++ b/components/scream/src/share/tests/field_tests.cpp
@@ -126,6 +126,43 @@ TEST_CASE("field", "") {
     // Trying to reshape into something that the allocation cannot accommodate should throw
     REQUIRE_THROWS (f1.get_reshaped_view<Pack<Real,32>***>());
   }
+
+  // Check whether we can correctly manipulate a field with two scalar types.
+  SECTION ("reshape and reduce with multiple scalar types") {
+    Field<Pack<Real,4>,Device> f1 (fid);
+    f1.allocate_view();
+
+    auto v1d = f1.get_view();
+
+    // Reshape the field to a view of Reals, and fill it with ones.
+    // Here we use the field's layout to determine the extents of the relevant
+    // views, since the view's extents include padding introduced by Packs.
+    auto v3d_1 = f1.get_reshaped_view<Real***>();
+    auto layout = f1.get_header().get_identifier().get_layout();
+    int num_reals = 0;
+    for (int i = 0; i < layout.dim(0); ++i) {
+      for (int j = 0; j < layout.dim(1); ++j) {
+        for (int k = 0; k < layout.dim(2); ++k, ++num_reals) {
+          v3d_1(i, j, k) = 1;
+        }
+      }
+    }
+
+    // Reshape the field back to Packs and reduce to demonstrate that we get
+    // the desired sum.
+    auto v3d_2 = f1.get_reshaped_view<Pack<Real,4>***>();
+    int num_packs = 0;
+    for (int i = 0; i < layout.dim(0); ++i) {
+      for (int j = 0; j < layout.dim(1); ++j) {
+        for (int k = 0; k < layout.dim(2); ++k, ++num_packs) {
+          Real sum = ekat::reduce_sum(v3d_2(i, j, k));
+          if (num_packs <= num_reals/4) {
+            REQUIRE(ekat::reduce_sum(v3d_2(i, j, k)) == 4);
+          }
+        }
+      }
+    }
+  }
 }
 
 TEST_CASE("field_repo", "") {
--- a/components/scream/src/share/tests/field_tests.cpp
+++ b/components/scream/src/share/tests/field_tests.cpp
@@ -138,26 +138,30 @@ TEST_CASE("field", "") {
     // Here we use the field's layout to determine the extents of the relevant
     // views, since the view's extents include padding introduced by Packs.
     auto v3d_1 = f1.get_reshaped_view<Real***>();
+    auto host_v3d_1 = Kokkos::create_mirror_view(v3d_1);
     auto layout = f1.get_header().get_identifier().get_layout();
     int num_reals = 0;
     for (int i = 0; i < layout.dim(0); ++i) {
       for (int j = 0; j < layout.dim(1); ++j) {
         for (int k = 0; k < layout.dim(2); ++k, ++num_reals) {
-          v3d_1(i, j, k) = 1;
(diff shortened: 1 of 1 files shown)


bartgol:
Yes, this is the type of test I had in mind.

jeff-cohere:
I'll be able to test this more thoroughly tomorrow. Are there any other tests that you had in mind for #561 ? In any case, I've also started on #566 and can push that in before circling back around to the field tests.

bartgol:
@jeff-cohere I think you are doing this in the FieldPropertyCheck PR, right? Should we close this then? No point in testing this stuff twice...

jeff-cohere:
Good idea. I haven't merged this test into that branch, but it'd be easy to do.

components/scream/src/share/tests/field_tests.cpp
```diff
@@ -126,6 +126,47 @@ TEST_CASE("field", "") {
...
+    // Reshape the field back to Packs and reduce to demonstrate that we get
+    // the desired sum.
+    auto v3d_2 = f1.get_reshaped_view<Pack<Real,4>***>();
+    auto host_v3d_2 = Kokkos::create_mirror_view(v3d_2);
+    Kokkos::deep_copy(host_v3d_2, v3d_2);
+    int num_packs = 0;
+    for (int i = 0; i < layout.dim(0); ++i) {
+      for (int j = 0; j < layout.dim(1); ++j) {
+        for (int k = 0; k < layout.dim(2); ++k, ++num_packs) {
+          Real sum = ekat::reduce_sum(host_v3d_2(i, j, k));
```
bartgol, line 162: This result seems unused.

components/scream/src/share/tests/field_tests.cpp
```diff
@@ -126,6 +126,47 @@ TEST_CASE("field", "") {
...
+          if (num_packs <= num_reals/4) {
+            REQUIRE(ekat::reduce_sum(host_v3d_2(i, j, k)) == 4);
```
bartgol, line 164: Seems like you're checking that each pack is filled with ones. Might be more interesting to compute view reduction of v3d_2(i,j,:) and v3d_1(i,j,:), with start=0 and end=fid.get_layout().dim(2), and check that they are the same. This can also be done with random inputs, not necessarily all ones.

components/scream/src/share/tests/field_tests.cpp
```diff
@@ -126,6 +126,47 @@ TEST_CASE("field", "") {
     // Trying to reshape into something that the allocation cannot accommodate should throw
     REQUIRE_THROWS (f1.get_reshaped_view<Pack<Real,32>***>());
   }
+
+  // Check whether we can correctly manipulate a field with two scalar types.
+  SECTION ("reshape and reduce with multiple scalar types") {
+    Field<Pack<Real,4>,Device> f1 (fid);
```
bartgol, line 132: This is fine, but notice that 1) in the FieldRepository all fields are `Field<Real,Device>` (no pack scalar type), and 2) the extent `fid.get_layout().dim(2)` will be interpreted as the number of packs, rather than the "physical" dimension of the field. Perhaps it's best to stick with `Field<Real,Device>`, to avoid create confusion in users.

Currently, we don't offer a way to change the scalar type of a Field (only the underlying view can be reshaped). I suspect this is not a limitation, since packs are needed at the computational level, where the whole Field structure is no longer needed (when calculations start, one should already be using the stored view (possibly reshaped), not the field).

components/scream/src/share/tests/field_tests.cpp
```diff
@@ -126,6 +126,47 @@ TEST_CASE("field", "") {
...
+    // Reshape the field to a view of Reals, and fill it with ones.
+    // Here we use the field's layout to determine the extents of the relevant
+    // views, since the view's extents include padding introduced by Packs.
+    auto v3d_1 = f1.get_reshaped_view<Real***>();
+    auto host_v3d_1 = Kokkos::create_mirror_view(v3d_1);
+    auto layout = f1.get_header().get_identifier().get_layout();
+    int num_reals = 0;
+    for (int i = 0; i < layout.dim(0); ++i) {
+      for (int j = 0; j < layout.dim(1); ++j) {
+        for (int k = 0; k < layout.dim(2); ++k, ++num_reals) {
```
bartgol, line 146: Notice that, since the scalar type of the field is `Pack<Real,4>` the value of dim(2) will be the number of packs, not the number of reals. This is probably another argument in favor of changing the scalar type inside the Field declaration to be simply `Real`.

components/scream/src/share/tests/field_tests.cpp
```diff
@@ -126,6 +126,47 @@ TEST_CASE("field", "") {
...
+    // the desired sum.
+    auto v3d_2 = f1.get_reshaped_view<Pack<Real,4>***>();
+    auto host_v3d_2 = Kokkos::create_mirror_view(v3d_2);
+    Kokkos::deep_copy(host_v3d_2, v3d_2);
+    int num_packs = 0;
+    for (int i = 0; i < layout.dim(0); ++i) {
+      for (int j = 0; j < layout.dim(1); ++j) {
+        for (int k = 0; k < layout.dim(2); ++k, ++num_packs) {
+          Real sum = ekat::reduce_sum(host_v3d_2(i, j, k));
+          if (num_packs <= num_reals/4) {
```
bartgol, line 163: I don't get this if statement. At this point, `num_reals` should equal `fid.get_layout().size()`. Since the loops are the same as in the section above, what is this if checking?

bartgol (changes requested):
Thanks jeff. I have some comments/requests, let me know what you think.

components/scream/src/share/tests/field_tests.cpp
jeff-cohere: From this and your other comments, it sounds like I really don't understand everything that I need to in order to test reshaping of fields this way. Maybe we should go back to the drawing board and draft a test with specific inputs that produce specific expected outputs. That would actually allow me to document some of these things.

bartgol: Sure, and maybe I can write some better documentation.

jeff-cohere:
I'll redo this unit test, now that I have a better idea of how the field repo works.

PR #651, opened 2020-10-06
perf-analysis update for scream

jgfouca:
Change list:
1) Upgrades to perf-analysis script for more-flexible commands and better support for primary scream repo operations
2) Add more-precise timings for P3 CXX by excluding initialization
3) Add more-precise timings for P3 f90 by excluding initialization
4) Additional flags for controlling p3_run_and_cmp behavior

### DIFF
--- a/components/cam/src/physics/cam/micro_p3.F90
+++ b/components/cam/src/physics/cam/micro_p3.F90
@@ -1108,7 +1108,11 @@ contains
        diag_eff_radius_qi,rho_qi,do_predict_nc, &
        dpres,exner,qv2qi_depos_tend,precip_total_tend,nevapr,qr_evap_tend,precip_liq_flux,precip_ice_flux,cld_frac_r,cld_frac_l,cld_frac_i,  &
        p3_tend_out,mu_c,lamc,liq_ice_exchange,vap_liq_exchange, &
-       vap_ice_exchange,qv_prev,t_prev,col_location)
+       vap_ice_exchange,qv_prev,t_prev,col_location &
+#ifdef SCREAM_CONFIG_IS_CMAKE
+       , elapsed_s &
+#endif
+)
 
     !----------------------------------------------------------------------------------------!
     !                                                                                        !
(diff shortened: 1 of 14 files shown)


jgfouca:
With these changes, I was able to produce this plot, the first attempt to measure overall P3 performance:
AndyP3

ambrad (approved):
Are you intending to run an OpenMP loop over the F90 code? If not, then I'd leave F90 out of the figure. I know this isn't the point of the PR, but I suspect the code needs a fair bit of work to run well on the GPU. The target is 2.5-3x SKX with enough columns, like in the P3 micro-app.

PeterCaldwell (approved):
This is really great, thanks! A couple of general comments (which don't need to be addressed in this PR):

1. F90 results are all around 0 - is the F90 functionality broken? Given that this is a tool which we will run over and over until we get the performance we want, I'd suggest leaving the F90 code in (but fixing it) since we will eventually want it in the final product.
2. It is very interesting to me that skylake tracks v100 well over the entire range of cols/node. As Andrew said, this is probably just because we haven't done any performance tuning yet. Interesting to see how the sausage is made though.
3. In the title, would it make more sense to replace "number of timesteps" with "number of timesteps * repeats"? Isn't that really what we want to measure?

jgfouca:
@ambrad , @PeterCaldwell yeah, the f90 lines are pretty useless. They will always just be a line along the bottom of the chart since they are serial. I was a bit surprised that the f90 segf'd with columns > 128 on Blake and bowman. f90 worked fine up to 8192 columns on mappy. The f90 numbers consistently got worse as columns increase, since more columns reqs more memory and f90 has no parallelism.

@PeterCaldwell , the f90 performance is not zero, just appears to be on the chart because it's so much slower than the cxx.

jgfouca:
For reference, this is a plot for rain_sed:
andy34

jgfouca:
I don't understand the shoc_ut fails from AT. test-all-scream works for me on mappy and Blake. I will try an upstream merge and retest.

Issue #639, opened 2020-10-01
A simple getter interface in field repo to grab pointer to a field


AaronDonahue:
The field repository is currently a map of a field name to a separate map.  That map is from field identifier to the field itself.  This was done because field identifiers are unique but one field name can point to multiple field identifiers.  Sometimes we need a way to quickly get a field given it's name and the grid it is defined on.  This can be done by looping through the map of field identifiers to fields, but a simple getter function in the field repo could simplify this process.

AaronDonahue:
@brhillman , I've made an issue for the feature we discussed on the scream slack channel.

bartgol:
This is assuming that for each grid, there will only be *one* version of each field. This is probably true, but might be limiting in the future.

One scenario where I think there might be different copies is if some field should be used with different layout in different parts of the atm code. Say a field that has layout (ncols,ntracers,nlevs), which in some other parts of the code might be more efficiently used if stored as (ncols,nlevs,ntracers).

A few questions:
1. do we want to support such a case? Meaning, do we allow a field to be stored in two different ways? Or can we error out (due to unmet dependencies) and force one of the two to adapt to the other?
2. If we decide to support this scenario, how do we do it?

Answering NO to 1 would likely allow a way to do `field_repo.get_field(name,grid_name)`, with increased simplicity. However, what if we have a field defined at level midpoints as well as level interfaces. Right now, one can give the same name, and the FM would treat them as different fields. E.g, pressure (ncols,nlev) and (ncols,nlev+1) are different, despite being on the same grid and having the same name. This sparks more decisions to make:

- should we allow two fields with the same name to have different dimensions? The classic examples is field at midlevel or interfaces.
- If the answer is no, we are forcing people to use different names for midpoints and interface quantities. Are we ok with that? If so, then we should have pre-defined suffixes for this distinction. And no suffix should imply a pre-defined position (mid or int), but which one?
- if the answer is yes, then the map `field_repo.get_field(name,grid_name)` cannot be implemented.

Another solution to the mid-vs-int dilemma: should midpoints and interfaces be considered "different grids"? I was thinking not, but maybe that would solve the problem of same name, since pmid and pint would live on "different" grids...

AaronDonahue:
[...]

PR #653, opened 2020-10-06
Add a simple field getter function to field repo.

AaronDonahue:
Add a simple getter function to the field repository to retrieve
a field given it's name and grid.

Add to the field repo unit tests to test this new functionality.

Addresses #639 

[BFB]

### DIFF
--- a/components/scream/src/share/tests/field_tests.cpp
+++ b/components/scream/src/share/tests/field_tests.cpp
@@ -155,6 +155,9 @@ TEST_CASE("field_repo", "") {
   fid5.set_dimensions(dims3);
   fid6.set_dimensions(dims3);
 
+  fid2.set_grid_name("grid_1");
+  fid3.set_grid_name("grid_2");
+
   FieldRepository<Real,DefaultDevice>  repo;
 
   // Should not be able to register fields yet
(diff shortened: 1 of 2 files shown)


AaronDonahue:
@brhillman , once this PR goes in you should be able to retrieve a field from the field repo with the two string arguments "NAME" and "GRID".

components/scream/src/share/tests/field_tests.cpp
```diff
@@ -186,6 +189,16 @@ TEST_CASE("field_repo", "") {
...
+  // Check that get_field with a field name and grid name as arguments returns the appropriate field
+  // Using grid_1 should return fid2 field, using grid_2 should return fid3 field, using grid_3 should throw an error.
+  auto& f7 = repo.get_field("field_2","grid_1");
+  auto& f8 = repo.get_field("field_2","grid_2");
+  REQUIRE_THROWS( repo.get_field("field_2","grid_3") );
+  REQUIRE(f7.get_header().get_identifier()==fid2);
+  REQUIRE(f7.get_header().get_identifier()!=fid3);
+  REQUIRE(f8.get_header().get_identifier()!=fid2);
+  REQUIRE(f8.get_header().get_identifier()==fid3);
+
```
bartgol, line 201: We should also check that if there are two fields called `field_2` on `grid_2`, but with different layout, then `get_field(field_2,grid_2)` throws.

bartgol (changes requested):

components/scream/src/share/tests/field_tests.cpp
AaronDonahue: Yes, good idea.  I'll put that in.

bartgol (approved):

PR #656, opened 2020-10-06
remove extra constant

tcclevenger:
Constant is double defined, causing tests to fail.


bogensch (approved):
yikes, sorry about that!

tcclevenger:
> yikes, sorry about that!

@bogensch I think it's from one of mine that got merged today, actually.

bogensch:
from both of us... I just had a PR go in that added that variable today and your PR also added it.

tcclevenger:
@bogensch Not sure if the tests can actually pass given that master fails. This may have to be forcefully merged..

PR #657, opened 2020-10-06
[WIP] Write p3 parameters in namelist_defaults_cam to atm_in

crterai:
This PR makes sure that p3 parameters added to namelist_defaults_cam are not missed and are written to atm_in during the build.


crterai:
We decided that this is not necessary for SCREAM, since SCREAM compsets define the parameters. Instead, this should be done in E3SM when P3 makes it into the E3SM codebase.

PR #659, opened 2020-10-06
Port shoc_diag_obklen

tcclevenger:
Port shoc_diag_obklen to c++.


ambrad:
This PR looks like it's missing the shoc.F90 bridge call, but I might just be confused.

ambrad (approved):

jeff-cohere (approved):

tcclevenger:
> This PR looks like it's missing the shoc.F90 bridge call, but I might just be confused.

@ambrad Nope, you are not confused. 2nd time I've missed those..

PR #660, opened 2020-10-07
SHOC property tests for remaining second moment functions

bogensch:
Creates property tests and boiler plate for the remaining low, mid, and upper level SHOC property tests.  This PR:

- Fixes and issue where variables "qtraces" and "num_qtracers" were being fed into the second moment functions in shoc.F90 but never used.  Removed these.
- Revises the boiler plate for functions "diag_second_moments_uybcond" and "diag_second_moments_srf", which had previously been converted, so that variable names are consistent and to avoid confusion (i.e. in SHOC world "thl" and "thl_sec" mean two totally different things).
- Revises the .cpp testing files for "diag_second_moments_uybcond" and "diag_second_moments_srf" so they match the pattern established by other testing files.
-  Generates tests and boilerplate code for the functions "diag_second_moments_lbycond", "diag_second_moments", and "diag_second_shoc_moments".  For these functions the boilerplate was generated via the gen-boiler tool.


xyuan (approved):
looks good, thanks.

PR #661, opened 2020-10-07
Adding property checks to Fields.

### LINKED ISSUE #566: [AD] Abstract interface to check field properties

jeff-cohere:
This PR adds a new class, `FieldPropertyCheck`, that operates on a `Field`, checking to see whether it possesses a certain property (or set of properties), and returning `true` if so and `false` otherwise. A `FieldPropertyCheck` can also attempt to "repair" a `Field` so that it has the desired property or properties.

Additionally, we make it possible to add one or more `FieldPropertyCheck`s to a `Field`. The property checks are managed via shared pointers to make it less complicated for a number of fields to be checked for certain properties.

Finally, we add a `PointerList` type that manages pointers to types, making it possible to easily manipulate them with const- and non-const iterators, and not much else. We use this `PointerList` type to store pointers to the property checks for a `Field`.

I'm making this PR visible before adding unit tests for the property checks, to make sure I've not misunderstood something. That's why it's got the [WIP] tag.

## Property Checks Implemented

We've got three property checks coming in with this PR (and tested with unit tests):

1. A positivity check (`FieldPositivityCheck`): checks that all field values are positive. You can construct it with a lower bound if you want the check to be able to repair the field (and set all non-positive values to that lower bound).
2. A boundedness check (`FieldBoundednessCheck`): checks that all field values fall within [fmin, fmax]. Repairs fields by projecting all values outside that interval to their nearest endpoints.
3. A monotonicity check (`FieldMonotonicityCheck`): checks that a field is monotonically increasing or decreasing. This check can't repair a non-monotonic field without making some assumptions, so I didn't pick that fight.

Closes #566

### DIFF
--- a/components/scream/src/share/field/field.hpp
+++ b/components/scream/src/share/field/field.hpp
@@ -3,6 +3,7 @@
 
 #include "share/field/field_header.hpp"
 #include "share/field/field_property_check.hpp"
+#include "share/util/pointer_list.hpp"
 #include "share/scream_types.hpp"
 
 #include "ekat/ekat_type_traits.hpp"
@@ -42,9 +43,13 @@ public:
   // Statically check that RealType is not an array.
   static_assert(view_type::Rank==1, "Error! RealType should not be an array type.\n");
 
-  using prop_check_type           = FieldPropertyCheck<ScalarType, DeviceType>;
-  using prop_check_iterator       = std::vector<std::shared_ptr<prop_check_type> >::iterator;
-  using prop_check_const_iterator = std::vector<std::shared_ptr<prop_check_type> >::const_iterator;
+  // A Field maintains a list of shared_ptrs to FieldPropertyChecks that can
+  // determine whether it satisfies certain properties. We use the pointer_list
+  // class to provide simple access to the property checks.
+  using property_check_type = FieldPropertyCheck<ScalarType, device_type>;
+  using property_check_list = pointer_list<std::shared_ptr<property_check_type>,
+                                           const property_check_type>;
+  using property_check_iterator = typename property_check_list::const_iterator;
 
   // Constructor(s)
   Field () = default;
@@ -69,15 +74,18 @@ public:
   const_field_type get_const () const { return const_field_type(*this); }
 
   // Adds a propery check to this field.
-  void add_prop_check(std::shared_ptr<prop_check_type> prop_check) {
-    m_prop_checks.push_back(prop_check);
+  void add_property_check(std::shared_ptr<property_check_type> property_check) {
+    m_prop_checks.push_back(property_check);
   }
 
-  // Allows access to the set of property checks on the field.
-  field_prop_check_iterator field_prop_check_begin() { return m_prop_checks.begin(); }
-  field_prop_check_const_iterator field_prop_check_begin() const { return m_prop_checks.begin(); }
-  field_prop_check_iterator field_prop_check_end() { return m_prop_checks.end(); }
-  field_prop_check_const_iterator field_prop_check_end() const { return m_prop_checks.end(); }
+  // These (forward) iterators allow access to the set of property checks on the
+  // field.
+  property_check_iterator property_check_begin() const {
+    return m_prop_checks.begin();
+  }
+  property_check_iterator property_check_end() const {
+    return m_prop_checks.end();
+  }
 
   // Allows to get the underlying view, reshaped for a different data type.
   // The class will check that the requested data type is compatible with the
@@ -107,7 +115,7 @@ protected:
   bool                            m_allocated;
 
   // List of property checks for this field.
-  std::vector<FieldPropertyCheck<ScalarType, DeviceType> > m_prop_checks;
+  property_check_list m_prop_checks;
 };
 
 template<typename RealType>
--- a/components/scream/src/share/field/field.hpp
+++ b/components/scream/src/share/field/field.hpp
@@ -44,10 +44,10 @@ public:
   static_assert(view_type::Rank==1, "Error! RealType should not be an array type.\n");
 
   // A Field maintains a list of shared_ptrs to FieldPropertyChecks that can
-  // determine whether it satisfies certain properties. We use the pointer_list
+  // determine whether it satisfies certain properties. We use the PointerList
   // class to provide simple access to the property checks.
   using property_check_type = FieldPropertyCheck<ScalarType, device_type>;
-  using property_check_list = pointer_list<std::shared_ptr<property_check_type>,
+  using property_check_list = PointerList<std::shared_ptr<property_check_type>,
                                            property_check_type>;
   using property_check_iterator = typename property_check_list::const_iterator;
 
--- a/components/scream/src/share/field/field.hpp
+++ b/components/scream/src/share/field/field.hpp
@@ -46,7 +46,7 @@ public:
   // A Field maintains a list of shared_ptrs to FieldPropertyChecks that can
   // determine whether it satisfies certain properties. We use the PointerList
   // class to provide simple access to the property checks.
-  using property_check_type = FieldPropertyCheck<ScalarType, device_type>;
+  using property_check_type = FieldPropertyCheck<RealType>;
   using property_check_list = PointerList<std::shared_ptr<property_check_type>,
                                            property_check_type>;
   using property_check_iterator = typename property_check_list::const_iterator;
--- a/components/scream/src/share/field/field.hpp
+++ b/components/scream/src/share/field/field.hpp
@@ -48,7 +48,7 @@ public:
   // class to provide simple access to the property checks.
   using property_check_type = FieldPropertyCheck<ScalarType, device_type>;
   using property_check_list = pointer_list<std::shared_ptr<property_check_type>,
-                                           const property_check_type>;
+                                           property_check_type>;
   using property_check_iterator = typename property_check_list::const_iterator;
 
   // Constructor(s)
--- a/components/scream/src/share/field/field.hpp
+++ b/components/scream/src/share/field/field.hpp
@@ -49,7 +49,7 @@ public:
   using property_check_type = FieldPropertyCheck<RealType>;
   using property_check_list = PointerList<std::shared_ptr<property_check_type>,
                                            property_check_type>;
-  using property_check_iterator = typename property_check_list::const_iterator;
+  using property_check_iterator = typename property_check_list::iterator;
 
   // Constructor(s)
   Field () = default;
@@ -75,16 +75,16 @@ public:
 
   // Adds a propery check to this field.
   void add_property_check(std::shared_ptr<property_check_type> property_check) {
-    m_prop_checks.append(property_check);
+    m_prop_checks->append(property_check);
   }
 
   // These (forward) iterators allow access to the set of property checks on the
   // field.
   property_check_iterator property_check_begin() const {
-    return m_prop_checks.begin();
+    return m_prop_checks->begin();
   }
   property_check_iterator property_check_end() const {
-    return m_prop_checks.end();
+    return m_prop_checks->end();
   }
 
   // Allows to get the underlying view, reshaped for a different data type.
@@ -115,7 +115,7 @@ protected:
   bool                            m_allocated;
 
   // List of property checks for this field.
-  property_check_list m_prop_checks;
+  std::shared_ptr<property_check_list> m_prop_checks;
 };
 
 template<typename RealType>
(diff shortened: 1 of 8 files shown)


bartgol:
@jeff-cohere this looks nice. Quick question: are you planning on adding one or two basic property checks to this PR (like maybe check `f cmp a`, with `cmp` some comparison op)?, or do you think it's best to leave them for a second PR?

jeff-cohere:
I'll add a couple of property checks to the PR and use them in a unit test or two.

components/scream/src/share/field/field.hpp
```diff
@@ -93,6 +117,9 @@ class Field {
 
   // Keep track of whether the field has been allocated
   bool                            m_allocated;
+
+  // List of property checks for this field.
+  property_check_list m_prop_checks;
```
bartgol, line 122: I think we should put this in the FieldHeader. The header is "shared" across all copies of the Field, so that if someone adds a property check in the header, all other copies of Field will see it. On the other hands, if some atm proc adds a prop check to the field itself, all other copies of the field won't be affected.

components/scream/src/share/field/field_boundedness_check.hpp
```diff
@@ -0,0 +1,76 @@
...
+
+namespace scream
+{
+
+// This field property check returns true if all data in a given field are
+// bounded by the given upper and lower bounds (inclusively), and false if not.
+// It can repair a field that fails the check by clipping all unbounded values
+// to their closest bound.
+template<typename ScalarType, typename Device>
+class FieldBoundednessCheck: public FieldPropertyCheck<ScalarType, Device> {
```
bartgol, line 15: Very minor: BoundednessCheck makes me think we only check that the field is bounded (i.e., finite). Maybe `FieldBoundsCheck` instead?

components/scream/src/share/field/field_boundedness_check.hpp
```diff
@@ -0,0 +1,76 @@
...
+  // No default constructor -- we need lower and upper bounds.
+  FieldBoundednessCheck () = delete;
+
+  // Constructor with lower and upper bounds -- can repair fields that fail the check
+  // by overwriting nonpositive values with the given lower bound.
+  explicit FieldBoundednessCheck (ScalarType lower_bound,
+                                  ScalarType upper_bound) :
+    m_lower_bound(lower_bound),
+    m_upper_bound(upper_bound) {
+    EKAT_ASSERT_MSG(lower_bound < upper_bound,
```
bartgol, line 27: I can foresee one trying to force a field to be constant, without changing any of the rest of the code. Perhaps when trying to see the impact of an input field, or to debug something. He/she might use a FPC with can_repair=true to enforce this.

If we think that might be a legit use of FPC, we should allow lb==ub.

components/scream/src/share/field/field_boundedness_check.hpp
```diff
@@ -0,0 +1,76 @@
...
+      } else {
+        mm.min_val = ekat::impl::min(mm.min_val, host_view(i));
+        mm.max_val = ekat::impl::max(mm.max_val, host_view(i));
+      }
+    }, Kokkos::MinMax<ScalarType>(minmax));
+    return ((minmax.min_val >= m_lower_bound) && (minmax.max_val <= m_upper_bound));
+  }
+
+  bool can_repair() const override {
+    return true;
```
bartgol, line 50: Should we maybe make this a configurable option at construction time? Some atm procs might not be interested in fixing the pb, but just want to error out...

components/scream/src/share/field/field_positivity_check.hpp
```diff
@@ -0,0 +1,72 @@
...
+namespace scream
+{
+
+// This field property check returns true if all data in a given field are
+// positive, and false if not. If constructed with a specific lower bound, it
+// can repair a field that fails the check by setting all nonpositive values
+// to that lower bound. If no specific lower bound is given (i.e. if the
+// default constructor is used), no repairs can be made.
+template<typename ScalarType, typename Device>
+class FieldPositivityCheck: public FieldPropertyCheck<ScalarType, Device> {
```
bartgol, line 16: Do you think it's worth to have an impl independent on that of FieldBoundednessCheck? Why not inheriting from that one, and set ub=std::numeric_limits<ScalarType>::max()? It might do additional checks (the upper bound) so there may be a small perf hit. Might not be really noticeable though, and it would keep code repetition lower.

components/scream/src/share/tests/field_tests.cpp
```diff
@@ -241,4 +244,187 @@ TEST_CASE("field_repo", "") {
...
+      REQUIRE(iter->check(f1));
+    }
+
+    // Write a positive value to the middle of the array that causes the
+    // monotonicity check to fail.
+    host_view(host_view.extent(0)/2) = 1;
+    Kokkos::deep_copy(f1_view, host_view);
+    for (auto iter = f1.property_check_begin(); iter != f1.property_check_end(); iter++) {
+      REQUIRE(not iter->check(f1));
+    }
```
bartgol, line 382: To be extra pedantic, you could also check that repair throws:
```
REQUIRE_THROWS(iter->repair());
```

components/scream/src/share/field/field_positivity_check.hpp
bartgol: Nvm, I now see that this has a sligthly different way of repairing, by setting a value that is strictly larger than 0.

Follow up then: do you think it's worth making a similar repair fix for FieldBoundsCheck? Like, if you passs 3 or 4 scalars, we interpret the 3rd (and 4th, if present) as the "repair" values. If a repair value is provided, they must be strictly contained in the (lb,ub) range; also, if, say, a lb repair value is present, our check is > rather than >=.

Maybe this can be done in a follow up pr though...

bartgol (changes requested):
Looks very good. I had some comments. One (FPC inside FieldHeader rather than Field) is more important, the others are a bit more debatable.

components/scream/src/share/field/field.hpp
jeff-cohere: That sounds good. I'll move this and `add_property_check` to the FieldHeader class. Do you think we should keep the `property_check_begin()` and `property_check_end()` iterators on the Field class, or move them to the FieldHeader as well?

components/scream/src/share/field/field_boundedness_check.hpp
jeff-cohere: The boundedness check sets upper and lower bounds on the values in the field, which seems to me to be the same as what you think it means. A bounds check makes me think about checking that an index is valid for a field. Am I understanding you correctly?

components/scream/src/share/field/field_boundedness_check.hpp
jeff-cohere: I agree.

components/scream/src/share/field/field_boundedness_check.hpp
jeff-cohere: Sure, that makes sense.

components/scream/src/share/field/field_positivity_check.hpp
jeff-cohere: This is an interesting idea. But I like your suggestion of doing it in a follow-up PR when we might have some discussion of what people expect to be able to do.

components/scream/src/share/field/field.hpp
jeff-cohere: There's an issue with this that introduces the usual inheritance-related C++ warts. :-( FieldPropertyCheck is a template class with `ScalarType` and `Device` parameters, but `FieldHeader` is just a regular class. So now we have to make an otherwise-pointless `FieldPropertyCheckBase` class. :-( This is an anti-pattern, as far as I'm concerned, because it introduces issues that we have to understand on a per-case basis (such as: what if we want to check a property on fields having different scalar types and/or resident devices?).

If we can assume that properties are checked in a specific context (say, by the FieldRepo, in which we know our ScalarType is Real), we can simplify things a bit. Do you have any thoughts about this?

bartgol: Ah, that's a good point. I have no strong preference. That said, I think we're mixing a bit the templated and non-templated approaches in scream: fields are templated on scalar/device, but atm processes hard code a fixed one everywhere.

I think at some point we will _have_ to go toward a `FieldBase` class, which defines _some_ interfaces, but not all (in particular, no interface to get the actual view). I've seen the patter in other libraries, and albeit not great, it can be done.

For now, I would say leave FPC inside the Field class, so we can get the concept merged in. We'll have to go back on the field concept soon, and possibly change design (e.g., with a FieldBase class).

components/scream/src/share/field/field_boundedness_check.hpp
bartgol: My math mindsetting interpret "boundedness check" as "is the field finite or does it have singularities somewhere?". But that's probably my mindsetting. As I said, this was very minor. And you're the native speaker, so if you think Boundedness makes sense, it works for me too. :)

jeff-cohere: The finite/singularities question is relevant, for sure. From a "boneheaded" physicist's perspective, though, I would think that "a <= x <= b for all x" also implies boundedness of that type, for a and b finite. But I guess you're observing that my version of "boundedness" requires someone to supply the a and b, whereas yours does not.

What about FieldWithinIntervalCheck, then?

You've corrected my English in technical contexts before, so no need to defer on those grounds. :-)

components/scream/src/share/field/field.hpp
bartgol: I think a compromise to this issue is to store a shared_ptr<property_check_list>. This way, if AtmProc1 adds a property check, it will be visible also inside AtmProc2.

I am not 100% sure this is ever going to be used (meaning, one proc adds a check that another proc will also want to run), but I can't exclude it. And the fix (turning the list into a ptr to a list) is probably small enough to just do it.

jeff-cohere: Sounds good!

bartgol: Actually, I think we _do_ need this. In particular, we may have the checks done outside the atm proc (otherwise why storing them in the field?), in which case, we need to make sure that if AtmProcX adds a prop test, it will be stored not only in its local copy, but also in all the copies of that field.

bartgol (changes requested):
I think we might need prop checks to be shared across all copies of the field (i.e., turn the list into a ptr to a list), so we can enforce the properties from the AD rather than from inside the atm proc. E.g., if we subcycle a process, we might be ok enforcing a property only at the end of the subcycling, and therefore from _outside_ the atm proc.

bartgol (approved):

components/scream/src/share/field/field_monotonicity_check.hpp
```diff
@@ -0,0 +1,49 @@
...
+  // Overrides.
+
+  bool check(const Field<RealType>& field) const override {
+    auto view = field.get_view();
+    RealType sign;
+    Kokkos::parallel_reduce(view.extent(0), KOKKOS_LAMBDA(Int i, RealType& s) {
+      if ((i > 0) && (i < view.extent(0)-1)) {
+        auto diff1 = view(i) - view(i-1);
+        auto diff2 = view(i+1) - view(i);
+        s *= (diff1 * diff2 > 0) ? (diff1 * diff2) : 0;
```
AaronDonahue, line 30: More of just a question.  Does Kokkos initialize `s` to 1 or something.  My reading of this loop is that we want to multiply by 0 if non-monotonic and so that the product reduce at the end makes the value of sign=0.  But what does `s` start off as?

AaronDonahue (approved):
I just had a question, more for understanding than anything else.  Thanks Jeff for putting this functionality in the AD!

components/scream/src/share/field/field_monotonicity_check.hpp
jeff-cohere: Good question, and an example of how lots of hidden assumptions live within Kokkos. The `Kokkos::Prod<RealType>` operator uses the "multiplicative identity" for RealType, which is 1, so that's what the `sign` variable is initialized to (and `s` is applied to it). Similarly, there's a `Kokkos::Sum<RealType>` that uses the "additive identity", which is 0.

AaronDonahue: Ahhh, very cool and slick.

jeff-cohere:
@bartgol , this PR seems to be stalled. It's not stale, is it? I've been rebasing along with master branch. Is there some way we can kick the AT to get it retested and merged?

bartgol:
It's odd that no test happened in 4 days. Something must be off with the AT. I'll inspect, then get back to you.

PR #664, opened 2020-10-08
SHOC fix to apply tracer surface fluxes in sfc_flux routine

bogensch:
While making property tests it was discovered that SHOC was not applying the tracer surface fluxes to the implicit diffusion interface.  This PR fixes this and updates property tests and boiler plate codes.

Note that for SCREAMv0 simulations this PR should be b4b as tracer fluxes were NOT being applied in SHOC (i.e. wtracer_sfc array is being set to all zeros in shoc_intr.F90).  HOWEVER, shoc_run_and_cmp should fail since that test sets tracer surface fluxes to non zero values.

**NOTE** the expected result was that shoc_run_and_cmp should have only produced errors for the tracer variable... however all SHOC variables are different.  Through investigation it was uncovered that vapor inputs to the Obukhov and PBL height routines were using the qtracer variable and NOT SHOC prognostic/diagnostic variables.  This inconsistency will be fixed in a separate PR.

UPDATE: performed upstream merge with master as #664 should now have addressed the aforementioned bugfix.  Now, in shoc_run_and_cmp ONLY qtracers should be different and all other variables bfb.


bogensch:
Tests are passing/failing as expected.

singhbalwinder (approved):
Looks good to me!

bogensch:
@PeterCaldwell Tests are passing/failing as expected with this and I consider it ready to merge.  I verified that this PR is bfb for SCREAMv0 simulations, only failing for shoc_run_and_cmp for tracers as expected since tracers are fed a non-zero flux for that test.

PR #665, opened 2020-10-08
Change minthresh of a liner_interp call

tcclevenger:
I'm porting `diag_third_shoc_moments`, and ran across this bug when doing randomized testing. Since `thetal_zi` will be used in `clipping_diag_third_shoc_moments` as `bet2 = ggr/thetal_zi`, then `thetal_zi != 0`.

### DIFF
--- a/components/cam/src/physics/cam/shoc.F90
+++ b/components/cam/src/physics/cam/shoc.F90
@@ -1542,7 +1542,7 @@ subroutine diag_third_shoc_moments(&
   call linear_interp(zt_grid,zi_grid,isotropy,isotropy_zi,nlev,nlevi,shcol,0._rtype)
   call linear_interp(zt_grid,zi_grid,brunt,brunt_zi,nlev,nlevi,shcol,largeneg)
   call linear_interp(zt_grid,zi_grid,w_sec,w_sec_zi,nlev,nlevi,shcol,(2._rtype/3._rtype)*mintke)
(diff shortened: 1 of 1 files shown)


ambrad:
I'm not sure I agree. I believe the key line is the call https://github.com/E3SM-Project/scream/blob/9a78b3dc6306fdb52c411c9d01bfad2e04922ba1/components/cam/src/physics/cam/shoc.F90#L1635 This call is made for `k = 2:nlev`, which excludes `k = nlevi`. Thus, I think the smallest thetal_zi value that is used in the denominator is the interface level one above the ground, which is finite and order meters. Am I missing something?

tcclevenger:
@ambrad For some reason, an interior point of `thetal_zi` occasionally drops below zero when I use random positive values for `thetal` in bfb tests. Since `thetal_zi` is an interpolated from `thetal`, could it be the case that it drops just slightly below 0 given unlucky values (relatively large changes)?

This is happening for both the fortran and c++ implementation. I agree that it is strange, so I will study closer.

ambrad:
theta_l (liquid water potential temperature) cannot physically be <= 0, so I'm guessing the random numbers are generating an unphysical situation. You should feel free to mod random numbers to obey certain constraints. I believe we should not modify the parameterization unless it truly has something wrong w.r.t. what it can actually see at runtime.

tcclevenger:
Sounds good! I'll close this PR.

PR #667, opened 2020-10-08
port shoc_energy_fixer

tcclevenger:
Port `shoc_energy_fixer` to C++.

Note, I decided to implement `shoc_energy_total_fixer`, `shoc_energy_threshold_fixer`, and `shoc_energy_dse_fixer` all inside `shoc_energy_fixer` since these functions are called nowhere else in shoc.F90. Also, `pdel` was an unused input so I removed.

### DIFF
--- a/components/scream/src/physics/shoc/shoc_energy_fixer_impl.hpp
+++ b/components/scream/src/physics/shoc/shoc_energy_fixer_impl.hpp
@@ -1,0 +1,92 @@
+#ifndef SHOC_ENERGY_FIXER_IMPL_HPP
+#define SHOC_ENERGY_FIXER_IMPL_HPP
+
+#include "shoc_functions.hpp" // for ETI only but harmless for GPU
+
+namespace scream {
+namespace shoc {
+
+/*
+ * Implementation of shoc shoc_energy_fixer. Clients should NOT
+ * #include this file, but include shoc_functions.hpp instead.
+ */
+
+template<typename S, typename D>
+KOKKOS_FUNCTION
+void Functions<S,D>::shoc_energy_fixer(
+  const MemberType&            team,
+  const Int&                   nlev,
+  const Int&                   nlevi,
+  const Scalar&                dtime,
+  const Int&                   nadv,
+  const uview_1d<const Spack>& zt_grid,
+  const uview_1d<const Spack>& zi_grid,
+  const Scalar&                se_b,
+  const Scalar&                ke_b,
+  const Scalar&                wv_b,
+  const Scalar&                wl_b,
+  const Scalar&                se_a,
+  const Scalar&                ke_a,
+  const Scalar&                wv_a,
+  const Scalar&                wl_a,
+  const Scalar&                wthl_sfc,
+  const Scalar&                wqw_sfc,
+  const uview_1d<const Spack>& rho_zt,
+  const uview_1d<const Spack>& tke,
+  const uview_1d<const Spack>& pint,
+  const uview_1d<Spack>&       rho_zi,
+  const uview_1d<Spack>&       host_dse)
+{
+  // Constants
+  const auto cp = C::CP;
+  const auto lcond = C::LatVap;
+  const auto lice = C::LatIce;
+  const auto mintke = SC::mintke;
+  const auto ggr = C::gravit;
+
+  // Local variables
+  Scalar te_a = 0;
+  Scalar te_b = 0;
+  Scalar se_dis = 0;
+  Int shoctop = 0;
+
+  // Compute linear interpolation of data into rho_zi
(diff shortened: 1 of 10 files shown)


components/scream/src/physics/shoc/shoc_energy_fixer_impl.hpp
```diff
@@ -0,0 +1,92 @@
...
+  Scalar te_b = 0;
+  Scalar se_dis = 0;
+  Int shoctop = 0;
+
+  // Compute linear interpolation of data into rho_zi
+  linear_interp(team,zt_grid,zi_grid,rho_zt,rho_zi,nlev,nlevi,0);
+  team.team_barrier();
+
+  // Compute the host timestep
+  const Scalar hdtime = dtime*float(nadv);
```
jgfouca, line 58: You don't want this cast, just let cxx auto-promote nadv.

components/scream/src/physics/shoc/shoc_energy_fixer_impl.hpp
```diff
@@ -0,0 +1,92 @@
...
+  const auto s_rho_zi = ekat::scalarize(rho_zi);
+  const Scalar shf = wthl_sfc*cp*s_rho_zi(nlevi-1);
+  const Scalar lhf = wqw_sfc*s_rho_zi(nlevi-1);
+  te_a = se_a + ke_a + (lcond+lice)*wv_a +lice*wl_a;
+  te_b = se_b + ke_b + (lcond+lice)*wv_b + lice*wl_b;
+  te_b += (shf+lhf*(lcond+lice))*hdtime;
+
+  // Limit the energy fixer to find highest layer where SHOC is active.
+  // Find first level where tke is higher than lowest threshold.
+  const auto s_tke = ekat::scalarize(tke);
```
jgfouca, line 70: I think you should be able to vectorize (use packs) for this chunk of code.

jgfouca (changes requested):
Nice work! Just a couple minor changes needed.

tcclevenger:
@jgfouca Thanks Jim! Let me know what you think.

components/scream/src/physics/shoc/shoc_energy_fixer_impl.hpp
```diff
@@ -0,0 +1,99 @@
...
+  const Scalar lhf = wqw_sfc*rho_zi(nlevi_pack)[nlevi_indx];
+  te_a = se_a + ke_a + (lcond+lice)*wv_a +lice*wl_a;
+  te_b = se_b + ke_b + (lcond+lice)*wv_b + lice*wl_b;
+  te_b += (shf+lhf*(lcond+lice))*hdtime;
+
+  //Find which pack and pack index corresponds to shoctop
+  int shoctop_pack = shoctop/Spack::n;
+  int shoctop_indx = shoctop%Spack::n;
+
+  // Limit the energy fixer to find highest layer where SHOC is active.
```
ambrad, line 74: This is technically correct since these are all local variables, but you can probably work out a better parallel scheme using a reduction over an integer, the level index.

tcclevenger: @ambrad Let me know what you think about this. Maybe a reduction over scalar index instead of pack index would be better?

components/scream/src/physics/shoc/shoc_energy_fixer_impl.hpp
```diff
@@ -0,0 +1,105 @@
...
+    // Here we set all indices s.t. tke==mintke to nlev-2 since
+    // we require shoctop <= nlev-2
+    auto local_indices = ekat::range<IntSmallPack>(k*IntSmallPack::n);
+    local_indices.set(tke(k)==mintke, nlev-2);
+    const Int min_indx = ekat::min(local_indices);
+
+    if (min_indx < local_shoctop) {
+      local_shoctop = min_indx;
+    }
+  }, Kokkos::Min<int>(shoctop));
```
ambrad, line 84: This looks great. One little technical thing. I believe -- but check this, because I might be wrong -- that a team-reduce in which the reduction variable is never updated will return a pre-set value for the reduction value given by a Kokkos reduction_type (or whatever it's called) trait value. So you won't get `shoctop = 0` back like you want. I _think_ you need to explicitly check for this case and if it arises, set shoctop = 0 after the reduction.

tcclevenger: Correct. Looking in the Kokkos code, `Kokkos::reduction_identity<int>::min() =maximum-signed-int` so this shouldn't be an issue since all our indices will be `<=nlev-2`

ambrad: Oh, right. Yes, your code as is does handle this case. Great!

ambrad (approved):
Very nice!

PR #668, opened 2020-10-09
SCREAM: added support for mappy; removed support for bowman

bartgol:
Remove support for bowman (decommissioned next week), and add support for mappy (AMD EPYC).

Note: I am already turning off bowman runs for the AT. After this PR goes in, I will turn on mappy testing for the AT.


bartgol:
Putting this PR on hold. We'll wait for E3SM-Project/EKAT#62 to go in first, update ekat sumodule, then recycle ekat's mach file for mappy, to avoid cmake code redundancy.

PeterCaldwell:
It will be great to have EPYC testing because Chrysalis will probably become a target machine for us. I wonder whether ditching KNL testing will bite us because most of our E3SMv0 runs are on Cori and KNL-type architectures are the main reason for our packs. I suspect ditching is fine because a). our tests are for E3SMv1 rather than v0 and v1 probably won't be run much on KNL and b). even though packs/vectorization is most important on KNL, we can test it on any machine. Is this other folks' opinion too?

components/scream/scripts/jenkins/mappy_setup
```diff
@@ -0,0 +1,5 @@
+module load sems-env sems-python/3.5.2
+
+source ./scream/components/scream/scripts/jenkins/sandia_son_proxy
```
jgfouca, line 3: Mappy is currently on the SRN unfortunately.

jgfouca (changes requested):

components/scream/scripts/jenkins/mappy_setup
bartgol: the son proxy is just an alias to the srn one. So sonproxy or wwwproxy doesn't matter. The IT folks confirmed that to me. Your next question might then be "why do we have an alias at all?". Well, good question.

bartgol:
> It will be great to have EPYC testing because Chrysalis will probably become a target machine for us. I wonder whether ditching KNL testing will bite us because most of our E3SMv0 runs are on Cori and KNL-type architectures are the main reason for our packs. I suspect ditching is fine because a). our tests are for E3SMv1 rather than v0 and v1 probably won't be run much on KNL and b). even though packs/vectorization is most important on KNL, we can test it on any machine. Is this other folks' opinion too?

I am not ditching KNL completely. But it's taking me a while to get testing working on KNL, so I figured EPYC could be a good short term solution.

components/scream/scripts/jenkins/mappy_setup
jgfouca: @bartgol , is that a new change? Using the right proxy for your network definitely mattered in the past.

jgfouca: Just tested it, you're right!

jgfouca (approved):

components/scream/scripts/jenkins/mappy_setup
bartgol: Yes, I was confused too when they told me. I found out a couple of months ago.

PR #669, opened 2020-10-09
port shoc_length

tcclevenger:
Ports `shoc_length` and `compute_conv_time_shoc_length` to C++.

In `shoc_length`, I have rearranged inputs in F90 (C++ had different order than F90, I chose to match to C++). Also, `dz_zi` and `thetal` were unused and so I removed them.


components/scream/src/physics/shoc/shoc_length_impl.hpp
```diff
@@ -0,0 +1,56 @@
...
+  const uview_1d<const Spack>& zi_grid,
+  const uview_1d<const Spack>& dz_zt,
+  const uview_1d<const Spack>& dz_zi,
+  const uview_1d<const Spack>& wthv_sec,
+  const uview_1d<const Spack>& thetal,
+  const uview_1d<const Spack>& thv,
+  const uview_1d<Spack>&       thv_zi,
+  const uview_1d<Spack>&       brunt,
+  const uview_1d<Spack>&       shoc_mix)
+{
```
tcclevenger, line 30: I would like input on my use of `team_barrier()`. Basically, any function that has a `Kokkos::parallel_*(PerThread)` call, and that sets a value which will be used later in this functions, I call `team_barrier()` after. Is this correct?

components/scream/src/physics/shoc/shoc_compute_l_inf_shoc_length_impl.hpp
```diff
@@ -32,6 +32,7 @@ ::compute_l_inf_shoc_length(
                                 [&] (const int k) -> Spack {
     return ekat::sqrt(tke(k))*dz_zt(k);
   }, denom);
```
tcclevenger, line 34: Another `team_barrier()` question: Here, I did not previously call the barrier, but I am computing `numer` and `denom` by performing a reduce over threads, then using those values to compute a scalar (`l_inf`). Will it be required here? Or is this over kill?

tcclevenger (commented):
@jgfouca @ambrad @jeff-cohere I left some additional questions regarding `team_barrier()` Not sure if my implementation is overkill.

components/scream/src/physics/shoc/shoc_length_impl.hpp
jeff-cohere: I am the least experienced in these matters of the three of us, but it seems to me that you only need a barrier between A and B if values computed by a thread in A are needed by another thread in B. Of course, this treats the team dispatch mechanism as a black box, so I'm interested to hear what Jim and Andrew have to say.

ambrad: What Jeff wrote is correct. If you have two loops over the same indices and index k does not depend on k+i for i /= 0, then the two loops don't need a team_barrier between them b/c the index-thread map is identical in the two loops.

A reduction has an implicit team_barrier since each thread gets its own copy of the reduction value, so it too does not need a team_barrier after it.

This is the type of question for which writing a single-file sandbox code can be useful: to play around with scenarios and discover nuances.

ambrad (approved):

jeff-cohere (approved):

jgfouca (approved):

tcclevenger:
I've made one more commit. It contains the following:
1. Removing all but 1 `team_barrier()`. We keep one before `compute_brunt_shoc_length()` since it requires a shift calculations on `thv_zi` which we set above with `linear_interp()`.
2. Removed `dz_zi` and `thetal` from this function (unused parameters).
3. Added comments.

tcclevenger:
@bartgol I've rebased to master and I am still experiencing failure on Mappy. Will I need to just create a new PR off master?

bartgol:
@tcclevenger There was an issue with mappy mach specs. There's a chance #680 fixed it. If not, the output will at least be parse-able. For context: I mistakenly told our testing script to use mpirun to launch testing, which caused 32 concurrent runs of the testing script. I believe they clashed when it came to create baselines, but the output was so messed up it's hard to tell. I'll add retest to this PR, and we'll see.

tcclevenger:
@ambrad @jeff-cohere @jgfouca Thanks for approving this PR, I'll need another one after the rebase..

jeff-cohere (approved):

ambrad (approved):

PR #672, opened 2020-10-10
Revert "Beydoun/prescribed ccn option new pr"

PeterCaldwell:
Reverts E3SM-Project/scream#642


PeterCaldwell:
This PR has the same goal as #673 , but the other PR tries to fix the problem rather than revert... which is better. So closing this one.

PR #673, opened 2020-10-10
Beydoun/prescribed ccn option new pr

xyuan:
restore this branch

### DIFF
--- a/components/cam/src/physics/cam/micro_p3.F90
+++ b/components/cam/src/physics/cam/micro_p3.F90
@@ -2707,7 +2707,7 @@ subroutine rain_self_collection(rho,qr_incld,nr_incld,    &
       if (dum2.lt.dum1) then
          dum = 1._rtype
       else if (dum2.ge.dum1) then
-         dum = 2._rtype-bfb_exp(2300._rtype*(dum2-dum1))
+         dum = 2._rtype - bfb_exp(2300._rtype*(dum2-dum1))
       endif
 
       if (iparam.eq.1) then
@@ -2848,7 +2848,7 @@ subroutine prevent_ice_overdepletion(pres,t_atm,qv,latent_heat_sublim,inv_dt,   
    real(rtype) :: dumqv_sat_i, qdep_satadj
 
    dumqv_sat_i = qv_sat(t_atm,pres,1)
-   qdep_satadj = (qv-dumqv_sat_i)/(1._rtype+bfb_square(latent_heat_sublim)*dumqv_sat_i/(cp*rv*bfb_square(t_atm)))*inv_dt
+   qdep_satadj = (qv-dumqv_sat_i)/(1._rtype + bfb_square(latent_heat_sublim)*dumqv_sat_i/(cp*rv* bfb_square(t_atm) ))*inv_dt
    qidep  = qidep*min(1._rtype,max(0._rtype, qdep_satadj)/max(qidep, 1.e-20_rtype))
    qi2qv_sublim_tend  = qi2qv_sublim_tend*min(1._rtype,max(0._rtype,-qdep_satadj)/max(qi2qv_sublim_tend, 1.e-20_rtype))
 
--- a/components/scream/src/physics/p3/tests/p3_run_and_cmp.cpp
+++ b/components/scream/src/physics/p3/tests/p3_run_and_cmp.cpp
@@ -98,13 +98,10 @@ struct Baseline {
   {
     for (const bool do_prescribed_CCN : {true, false}) {
       for (const bool do_predict_nc : {true, false}) {
-        if (predict_nc == "both" || ( (do_predict_nc and predict_nc == "yes") || (!do_predict_nc and predict_nc == "no") )) {
-          //                 initial condit,     dt,  nsteps, prescribe or predict nc
-          params_.push_back({ic::Factory::mixed, 300, nsteps, do_predict_nc});
-        }
-        if (prescribed_ccn == "both" || ( (do_prescribed_CCN and prescribed_ccn == "yes") || (!do_prescribed_CCN and prescribed_ccn == "no") )) {
-          //                 initial condit,     dt,  nsteps, prescribe ccn
-          params_.push_back({ic::Factory::mixed, 300, nsteps, do_prescribed_CCN});
+        if ((predict_nc == "both"    || ( (do_predict_nc and predict_nc == "yes")         || (!do_predict_nc and predict_nc == "no") )) || 
+           (prescribed_ccn == "both" || ( (do_prescribed_CCN and prescribed_ccn == "yes") || (!do_prescribed_CCN and prescribed_ccn == "no") ))) {
+          //                 initial condit,     dt,  nsteps, prescribe or predict nc, prescribe CCN or not
+          params_.push_back({ic::Factory::mixed, 300, nsteps, do_predict_nc, do_prescribed_CCN});
         }
       }
     }
--- a/components/scream/src/physics/p3/tests/p3_run_and_cmp.cpp
+++ b/components/scream/src/physics/p3/tests/p3_run_and_cmp.cpp
@@ -93,13 +93,19 @@ static Int compare (const std::string& label, const Scalar* a,
 }
 
 struct Baseline {
-  Baseline (const Int nsteps=6, const Int ncol=3, const Int repeat=0, const std::string predict_nc="both") :
+  Baseline (const Int nsteps=6, const Int ncol=3, const Int repeat=0, const std::string predict_nc="both", const std::string prescribed_ccn="both") :
     m_ic_ncol(ncol), m_repeat(repeat)
   {
-    for (const bool do_predict_nc : {true, false}) {
-      if (predict_nc == "both" || ( (do_predict_nc and predict_nc == "yes") || (!do_predict_nc and predict_nc == "no") )) {
-        //                 initial condit,     dt,  nsteps, prescribe or predict nc
-        params_.push_back({ic::Factory::mixed, 300, nsteps, do_predict_nc});
+    for (const bool do_prescribed_CCN : {true, false}) {
+      for (const bool do_predict_nc : {true, false}) {
+        if (predict_nc == "both" || ( (do_predict_nc and predict_nc == "yes") || (!do_predict_nc and predict_nc == "no") )) {
+          //                 initial condit,     dt,  nsteps, prescribe or predict nc
+          params_.push_back({ic::Factory::mixed, 300, nsteps, do_predict_nc});
+        }
+        if (prescribed_ccn == "both" || ( (do_prescribed_CCN and prescribed_ccn == "yes") || (!do_prescribed_CCN and prescribed_ccn == "no") )) {
+          //                 initial condit,     dt,  nsteps, prescribe ccn
+          params_.push_back({ic::Factory::mixed, 300, nsteps, do_prescribed_CCN});
+        }
       }
     }
   }
@@ -188,12 +194,14 @@ private:
     Real dt;
     Int it;
     bool do_predict_nc;
+    bool do_prescribed_CCN;
   };
 
   static void set_params (const ParamSet& ps, FortranData& d) {
     d.dt = ps.dt;
     d.it = ps.it;
     d.do_predict_nc = ps.do_predict_nc;
+    d.do_prescribed_CCN = ps.do_prescribed_CCN;
   }
 
   std::vector<ParamSet> params_;
@@ -247,7 +255,8 @@ int main (int argc, char** argv) {
       "  -s <steps>      Number of timesteps. Default=6.\n";
       "  -i <cols>       Number of columns. Default=6.\n";
       "  -r <repeat>     Number of repetitions, implies timing run (generate + no I/O). Default=0.\n"
-      "  -p <predict_nc> yes|no|both. Default=both.\n";
+      "  -p <predict_nc>  yes|no|both. Default=both.\n";
+      "  -c <prescribed_ccn> yes|no|both. Default=both.\n";
     return 1;
   }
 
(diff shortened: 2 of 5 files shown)


xyuan:
@PeterCaldwell I restored this branch to fix this failure, now everything is working. I also add one more option to test prescribed_CCN which makes our life easy. Here is the test results,
bash-4.2$ jsrun -n 1 -a 1 -c 1 -g 1 ./p3_run_and_cmp_cxx -f -g -t 0
ExecSpace name: Cuda
ExecSpace initialized: yes
 avx  compiler GCC default FPE mask 0 (NONE) 
 #host threads 1
sizeof(Real) = 8
default pack size = 1

Generating to 08

bash-4.2$ jsrun -n 1 -a 1 -c 1 -g 1 ./p3_run_and_cmp_cxx  -t 0
ExecSpace name: Cuda
ExecSpace initialized: yes
 avx  compiler GCC default FPE mask 0 (NONE) 
 #host threads 1
sizeof(Real) = 8
default pack size = 1

Comparing with 08 at tol 0.0e+00

--- checking case # 2, timestep # 2 of 6 ---
--- checking case # 2, timestep # 3 of 6 ---
--- checking case # 2, timestep # 4 of 6 ---
--- checking case # 2, timestep # 5 of 6 ---
--- checking case # 2, timestep # 6 of 6 ---
--- checking case # 3, timestep # 1 of 6 ---
--- checking case # 3, timestep # 2 of 6 ---
--- checking case # 3, timestep # 3 of 6 ---
--- checking case # 3, timestep # 4 of 6 ---
--- checking case # 3, timestep # 5 of 6 ---
--- checking case # 3, timestep # 6 of 6 ---
--- checking case # 4, timestep # 1 of 6 ---
--- checking case # 4, timestep # 2 of 6 ---
--- checking case # 4, timestep # 3 of 6 ---
--- checking case # 4, timestep # 4 of 6 ---
--- checking case # 4, timestep # 5 of 6 ---
--- checking case # 4, timestep # 6 of 6 ---
--- checking case # 5, timestep # 1 of 6 ---
--- checking case # 5, timestep # 2 of 6 ---
--- checking case # 5, timestep # 3 of 6 ---
--- checking case # 5, timestep # 4 of 6 ---
--- checking case # 5, timestep # 5 of 6 ---
--- checking case # 5, timestep # 6 of 6 ---
--- checking case # 6, timestep # 1 of 6 ---
--- checking case # 6, timestep # 2 of 6 ---
--- checking case # 6, timestep # 3 of 6 ---
--- checking case # 6, timestep # 4 of 6 ---
--- checking case # 6, timestep # 5 of 6 ---
--- checking case # 6, timestep # 6 of 6 ---
--- checking case # 7, timestep # 1 of 6 ---
--- checking case # 7, timestep # 2 of 6 ---
--- checking case # 7, timestep # 3 of 6 ---
--- checking case # 7, timestep # 4 of 6 ---
--- checking case # 7, timestep # 5 of 6 ---
--- checking case # 7, timestep # 6 of 6 ---
--- checking case # 8, timestep # 1 of 6 ---
--- checking case # 8, timestep # 2 of 6 ---
--- checking case # 8, timestep # 3 of 6 ---
--- checking case # 8, timestep # 4 of 6 ---
--- checking case # 8, timestep # 5 of 6 ---
--- checking case # 8, timestep # 6 of 6 ---

xyuan:
@PeterCaldwell Test performed on NERSC passed also with baseline generated by the new HEAD

===============================================================================
Testing 'beydoun/prescribed_CCN_option_new_PR' for test 'full_debug'
===============================================================================
100% tests passed, 0 tests failed out of 60

Label Time Summary:
           = 3823.01 sec*proc (55 tests)
 driver    =   1.27 sec*proc (2 tests)
 p3        =  18.87 sec*proc (2 tests)
 rrtmgp    =   0.68 sec*proc (1 test)
physics    =  19.55 sec*proc (3 tests)
shoc       =   9.54 sec*proc (1 test)

Total Test time (real) =  64.76 sec
===============================================================================
Testing 'beydoun/prescribed_CCN_option_new_PR' for test 'full_sp_debug'
===============================================================================


100% tests passed, 0 tests failed out of 60

Label Time Summary:
           = 3589.43 sec*proc (55 tests)
 driver    =   1.69 sec*proc (2 tests)
 p3        =  17.93 sec*proc (2 tests)
 rrtmgp    =   0.75 sec*proc (1 test)
physics    =  18.69 sec*proc (3 tests)
shoc       =   9.90 sec*proc (1 test)

Total Test time (real) =  61.44 sec

===============================================================================
Testing 'beydoun/prescribed_CCN_option_new_PR' for test 'debug_nopack_fpe'
===============================================================================
100% tests passed, 0 tests failed out of 60

Label Time Summary:
           = 3797.45 sec*proc (55 tests)
 driver    =   1.17 sec*proc (2 tests)
 p3        =  18.44 sec*proc (2 tests)
 rrtmgp    =   0.57 sec*proc (1 test)
physics    =  19.00 sec*proc (3 tests)
shoc       =   9.31 sec*proc (1 test)

Total Test time (real) =  64.50 sec
**OVERALL STATUS: PASS**

PeterCaldwell:
I'm very confused by the title of this PR, which is identical to #642 which I merged on Friday. But this PR is actually different: It was created by xyuan and fixes the bugs in the previous PR. Perhaps it could be titled: "Fix bugs in #642"?

xyuan:
@PeterCaldwell  @AaronDonahue 
The fortran build error is fixed now, see below, it was caused by the bfb_* math function, and then a "comma" syntax error.

Time spent not building: 0.118072 sec
Time spent building: 673.464545 sec
MODEL BUILD HAS FINISHED SUCCESSFULLY

xyuan:
> I'm very confused by the title of this PR, which is identical to #642 which I merged on Friday. But this PR is actually different: It was created by xyuan and fixes the bugs in the previous PR. Perhaps it could be titled: "Fix bugs in #642"?

This PR is the restoration of the merged branch, so it is identical to the merged branch.

components/cam/src/physics/cam/micro_p3.F90
```diff
@@ -2848,7 +2848,7 @@ subroutine prevent_ice_overdepletion(pres,t_atm,qv,latent_heat_sublim,inv_dt,
    real(rtype) :: dumqv_sat_i, qdep_satadj
 
    dumqv_sat_i = qv_sat(t_atm,pres,1)
-   qdep_satadj = (qv-dumqv_sat_i)/(1._rtype+bfb_square(latent_heat_sublim)*dumqv_sat_i/(cp*rv*bfb_square(t_atm)))*inv_dt
+   qdep_satadj = (qv-dumqv_sat_i)/(1._rtype + bfb_square(latent_heat_sublim)*dumqv_sat_i/(cp*rv* bfb_square(t_atm) ))*inv_dt
```
PeterCaldwell, line 2851: The extra spaces you added here shouldn't matter, right? I like the new spacing better, but I don't think this is really relevant, right?

components/scream/src/physics/p3/tests/p3_run_and_cmp.cpp
```diff
@@ -93,13 +93,19 @@ static Int compare (const std::string& label, const Scalar* a,
...
   {
-    for (const bool do_predict_nc : {true, false}) {
-      if (predict_nc == "both" || ( (do_predict_nc and predict_nc == "yes") || (!do_predict_nc and predict_nc == "no") )) {
-        //                 initial condit,     dt,  nsteps, prescribe or predict nc
-        params_.push_back({ic::Factory::mixed, 300, nsteps, do_predict_nc});
+    for (const bool do_prescribed_CCN : {true, false}) {
+      for (const bool do_predict_nc : {true, false}) {
+        if (predict_nc == "both" || ( (do_predict_nc and predict_nc == "yes") || (!do_predict_nc and predict_nc == "no") )) {
+          //                 initial condit,     dt,  nsteps, prescribe or predict nc
+          params_.push_back({ic::Factory::mixed, 300, nsteps, do_predict_nc});
```
AaronDonahue, line 103: Doesn't the ParamSet struct have 5 elements now?  How does this line and the one below switch between whther `do_predict_nc` is being set or `do_prescribed_CCN`?

PeterCaldwell: I agree with Aaron - params_.push_back should contain both the do_predict_nc and do_prescribed_CCN state. This breaks the weird "loop over all options but use an if statement to decide whether or not to actually do anything" logic you've inherited here. I think fixing this will require rethinking how the logic works (which may be best to save for a follow-up PR).

I was initially very confused why the code doesn't crash when passed the wrong number of args. My recollection is that any entry to params_ which isn't assigned is given the value of 0. Thus you're only testing the "false" branch of the thing you don't specify.

components/cam/src/physics/cam/micro_p3.F90
xyuan: This is mistry to me when I added the space, this code can be built successfully, otherwise, it was failed. more specifically, it is the "+" operator that needs an additional space.

components/scream/src/physics/p3/tests/p3_run_and_cmp.cpp
xyuan: Yes, this is highly possibly the test code only test partial branch. The fix is committed.

components/cam/src/physics/cam/micro_p3.F90
xyuan: Where the error says,
PGF90-S-0038-Symbol, bfb_square, has not been explicitly declared (/autofs/nccs-svm1_home1/yuanx/scream_ccn_fix/components/cam/src/physics/cam/micro_p3.F90)
  0 inform,   0 warnings,   1 severes, 0 fatal for prevent_ice_overdepletion
gmake[2]: *** [cmake/atm/CMakeFiles/atm.dir/__/__/cam/src/physics/cam/micro_p3.F90.o] Error 2
gmake[2]: Leaving directory `/gpfs/alpine/cli115/proj-shared/yuanx/e3sm_scratch/ERS.ne4pg2_ne4pg2.F2010-SCREAM-LR.summit_pgi.20201012_113415_x0cbbb/bld/cmake-bld'
gmake[1]: *** [cmake/atm/CMakeFiles/atm.dir/all] Error 2
gmake[1]: Leaving directory `/gpfs/alpine/cli115/proj-shared/yuanx/e3sm_scratch/ERS.ne4pg2_ne4pg2.F2010-SCREAM-LR.summit_pgi.20201012_113415_x0cbbb/bld/cmake-bld'
gmake: *** [all] Error 2

PeterCaldwell: Ok, this makes no sense to me but since the code is working now and having spaces before/after the '+' operator looks more pleasant anyways, I say we just go with it.

components/scream/src/physics/p3/tests/p3_run_and_cmp.cpp
AaronDonahue: Thanks for putting the fix in xyuan.

components/scream/src/physics/p3/tests/p3_run_and_cmp.cpp
```diff
@@ -93,13 +93,16 @@ static Int compare (const std::string& label, const Scalar* a,
...
-      if (predict_nc == "both" || ( (do_predict_nc and predict_nc == "yes") || (!do_predict_nc and predict_nc == "no") )) {
-        //                 initial condit,     dt,  nsteps, prescribe or predict nc
-        params_.push_back({ic::Factory::mixed, 300, nsteps, do_predict_nc});
+    for (const bool do_prescribed_CCN : {true, false}) {
+      for (const bool do_predict_nc : {true, false}) {
+        if ((predict_nc == "both"    || ( (do_predict_nc and predict_nc == "yes")         || (!do_predict_nc and predict_nc == "no") )) || 
+           (prescribed_ccn == "both" || ( (do_prescribed_CCN and prescribed_ccn == "yes") || (!do_prescribed_CCN and prescribed_ccn == "no") ))) {
+          //                 initial condit,     dt,  nsteps, prescribe or predict nc, prescribe CCN or not
+          params_.push_back({ic::Factory::mixed, 300, nsteps, do_predict_nc, do_prescribed_CCN});
+        }
```
AaronDonahue, line 105: what happens when none of these options are true.  Does params_ just not add a new case?  Note, I think this question would go back to the original code before this PR as well.

jgfouca (approved):

PeterCaldwell:
The logic for testing predict_nc and prescribed_CCN still wasn't right in p3_run_and_cmp: it was going to test both branches of both conditionals all the time still. I just pushed a change which should fix that. Based on last Friday's experience, I don't want to integrate until my change is run through the AT. I'm also trying to run F-compset tests and they're stalling out... I'd like to sort that out before integrating.

components/scream/src/physics/p3/tests/p3_run_and_cmp.cpp
PeterCaldwell: I think you're right to question that. I rewrote the logic to solve the problem you're mentioning...

AaronDonahue (approved):
Just confirmed with the scream test suite on syrah that all tests PASS on this branch.

xyuan:
> Just confirmed with the scream test suite on syrah that all tests PASS on this branch.

awesome, so we could merge it to master.

PeterCaldwell:
Ok, so @AaronDonahue 's F-compset tests pass and AT is failing just for p3_run_and_cmp in a way that's expected: non-bfb for case 1 and 2 because master is currently running p3_main_part1 with its arguments scrambled so its output doesn't make sense. Then it crashes on case 3 because master only had 2 cases (this PR adds 2 more to ensure coverage for do_prescribed_CCN). This is all expected so I'll merge.

PR #675, opened 2020-10-12
non-bfb changes for shoc_assumed_pdf porting

tcclevenger:
Several changes to `bfb_*()` function for `shoc_assumed_pdf` porting.

All changes to `bfb_square()` and `bfb_sqrt()` cause no change to output from bfb testing, however, like other bfb changes, `bfb_cube` and `bfb_pow` can change the `shoc_run_and_cmp` output by a significant amount (~1e-1, `host_dse` is always the largest).

Again, I test all these by calculating the relative difference between variables containing `bfb_cube(x)` and those containing `x**3` everywhere it is used during ./shoc_run_and_cmp (likewise with `bfb_pow`). For all of these changes, the maximum relative difference of any one change is
`abs(old_val - new_val)/abs(old_val) < 7e-06(SP), 3e-14 (DP)`

### DIFF
--- a/components/cam/src/physics/cam/shoc.F90
+++ b/components/cam/src/physics/cam/shoc.F90
@@ -2190,7 +2190,7 @@ subroutine shoc_assumed_pdf_vv_parameters(&
   real(rtype) :: w_tol_sqd
   w_tol_sqd = bfb_square(2.e-2_rtype)
 
-  Skew_w=w3var/bfb_pow(w_sec,(3./2.))
+  Skew_w=w3var/bfb_sqrt(bfb_cube(w_sec))
 
   if (w_sec .le. w_tol_sqd) then
     Skew_w=0._rtype
--- a/components/cam/src/physics/cam/shoc.F90
+++ b/components/cam/src/physics/cam/shoc.F90
@@ -1997,7 +1997,7 @@ subroutine shoc_assumed_pdf(&
 
   thl_tol=1.e-2_rtype
   rt_tol=1.e-4_rtype
-  w_tol_sqd=(2.e-2_rtype)**2
+  w_tol_sqd=bfb_square(2.e-2_rtype)
   w_thresh=0.0_rtype
 
   ! Initialize cloud variables to zero
@@ -2033,9 +2033,9 @@ subroutine shoc_assumed_pdf(&
 
       ! Compute square roots of some variables so we don't
       !  have to compute these again
-      sqrtw2 = sqrt(w_sec(i,k))
-      sqrtthl = max(thl_tol,sqrt(thlsec))
-      sqrtqt = max(rt_tol,sqrt(qwsec))
+      sqrtw2 = bfb_sqrt(w_sec(i,k))
+      sqrtthl = max(thl_tol,bfb_sqrt(thlsec))
+      sqrtqt = max(rt_tol,bfb_sqrt(qwsec))
 
       !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
       !  FIND PARAMETERS FOR VERTICAL VELOCITY
@@ -2187,9 +2187,10 @@ subroutine shoc_assumed_pdf_vv_parameters(&
   real(rtype) :: sqrtw2t
 
   ! parameters
-  real(rtype), parameter :: w_tol_sqd=(2.e-2_rtype)**2
+  real(rtype) :: w_tol_sqd
+  w_tol_sqd = bfb_square(2.e-2_rtype)
 
-  Skew_w=w3var/w_sec**(3./2.)
+  Skew_w=w3var/bfb_pow(w_sec,(3./2.))
 
   if (w_sec .le. w_tol_sqd) then
     Skew_w=0._rtype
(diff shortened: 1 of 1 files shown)


components/cam/src/physics/cam/shoc.F90
```diff
@@ -1993,14 +1993,15 @@ subroutine shoc_assumed_pdf(&
...
+  real(rtype) :: sqrt2, sqrtpi
+  sqrt2=bfb_sqrt(2._rtype)
+  sqrtpi=bfb_sqrt(2._rtype*3.14_rtype)
 
   epsterm=rgas/rv
 
   thl_tol=1.e-2_rtype
   rt_tol=1.e-4_rtype
-  w_tol_sqd=(2.e-2_rtype)**2
+  w_tol_sqd=bfb_square(2.e-2_rtype)
```
tcclevenger, line 2004: Should `2.e` be changed to `2._rtype`?

components/cam/src/physics/cam/shoc.F90
```diff
@@ -2190,9 +2191,10 @@ subroutine shoc_assumed_pdf_vv_parameters(&
   real(rtype) :: sqrtw2t
 
   ! parameters
-  real(rtype), parameter :: w_tol_sqd=(2.e-2_rtype)**2
+  real(rtype) :: w_tol_sqd
+  w_tol_sqd = bfb_square(2.e-2_rtype)
 
-  Skew_w=w3var/w_sec**(3./2.)
+  Skew_w=w3var/bfb_pow(w_sec,(3./2.))
```
tcclevenger, line 2197: Should `3./2.` be changed to `3._rtype/2._rtype`?

tcclevenger (commented):
I've left 2 questions about other potential changes..

components/cam/src/physics/cam/shoc.F90
ambrad: Yes.

components/cam/src/physics/cam/shoc.F90
```diff
@@ -1993,14 +1993,15 @@ subroutine shoc_assumed_pdf(&
   real(rtype) :: qw_sec_zt(shcol,nlev)
 
   ! define these so they don't have to be computed more than once
-  real(rtype), parameter :: sqrt2 = sqrt(2._rtype)
-  real(rtype), parameter :: sqrtpi = sqrt(2._rtype*3.14_rtype)
+  real(rtype) :: sqrt2, sqrtpi
+  sqrt2=bfb_sqrt(2._rtype)
+  sqrtpi=bfb_sqrt(2._rtype*3.14_rtype)
```
ambrad, line 1998: @bogensch would it make sense in a separate PR to use pi to full precision here?

components/cam/src/physics/cam/shoc.F90
ambrad: Better: `bfb_sqrt(bfb_cube(w_sec))`.

ambrad (changes requested):

components/cam/src/physics/cam/shoc.F90
bartgol: Also, is this meant to be sqrt(pi) or is it really sqrt(2*pi)?

bogensch: It is meant to be sqrt(2*pi)... and yes, I agree that full precision pi should be used.

components/cam/src/physics/cam/shoc.F90
tcclevenger: Actually, I think I read this incorrect. This is `2e-2`.

components/cam/src/physics/cam/shoc.F90
tcclevenger: I'll create one.

PeterCaldwell: Could we please use the name `sqrt2pi` instead of `sqrtpi` for `sqrt(2*pi)` ? Otherwise the name is misleading.

tcclevenger: In  `physics_constants.hpp`, pi is defined `pi=3.14159265`. Shouldn't it be more digits (in both places)? Should I add a constant to `shoc.F90` similarly?

bartgol: Since pi is a universal constant, I would put it in `physics/share` at the very least. It might be actually belong into a more common package (see E3SM-Project/EKAT#48 as well).

Either way: why forcing full precision? This is F90, so it probably doesn't matter, but if the same design stays in C++, isn't there the chance that it will cause all downstream calculation to be in double precision? It seems to me that, unless there's a clear _need_ for pi in double precision, then its precision should match that of the rest of the library...

tcclevenger: Here is a [PR for adding pi to shoc.F90](https://github.com/E3SM-Project/scream/pull/677) (instead of using 3.14) and changing the name to `sqrt2pi`.

tcclevenger:
@bartgol @ambrad This branch is now rebased to master and the failing tests are expected. See initial comment on the PR for the rel_diff of the changes/

ambrad (approved):

ambrad:
@tcclevenger should I manually merge?

tcclevenger:
@ambrad yes

PR #677, opened 2020-10-12
add pi to shoc.F90

tcclevenger:
From https://github.com/E3SM-Project/scream/pull/675. This adds `pi` as a public variable. Also changes `sqrtpi` to `sqrt2pi`


components/cam/src/physics/cam/shoc.F90
```diff
@@ -29,6 +29,7 @@ module shoc
 logical :: use_cxx = .true.
 
 real(rtype), parameter, public :: largeneg = -99999999.99_rtype
+real(rtype), parameter, public :: pi = 3.14159265 ! Pi
```
bartgol, line 32: Isn't this less than double precision if rtype is double? Does `4*atan(1.0_rtype)` work inside a parameter expression in F90?

tcclevenger: I must have misunderstood. This is the value used in `physics/share/physics_constants.hpp`. 

`4*atan(1._rtype)` works as a parameter expression.

ambrad: I recommend writing `real(8), parameter :: pi = 3.141592653589793_rtype`. You need rtype to retain double precision and want full precision in the constant.

tcclevenger: Will this be bfb to `static constexpr Scalar Pi = 3.14159265` in `physics_constants.hpp`? Is this because we use `real(8)` here?

ambrad: I meant to write `rtype` above, not `r8`; I've fixed this. That aside, first, the number of digits to be written should be decided. Then `_rtype` should be suffixed regardless so there is no conversion of the literal to SP before the quantity is assigned to the `real(rtype)`.

tcclevenger:
@ambrad I changed the expression to match the number of digits in the c++ code. If I should include all the double precision digits, let me know and I can change the c++ version as well.

bartgol:
If you write out the number, then I think you should use all DP digits, so that it is represented correctly in a DP build. Otherwise in DP builds you would get 3.1415926500000000 instead of 3.1415926535897979.

tcclevenger:
@bartgol I added the full DP digits to shoc.F90. I then added a pi constant to `physics/shoc/shoc_constants.hpp`, since the one in `physics/share/physics_constants.hpp` is only defined up to SP and I imagine tests fail if I change that definition.

bartgol:
@tcclevenger ah. Personally, I don't like to have multiple definitions of pi, especially if they have different accuracy.

I think we _should_ make sure that in DP builds pi is correct to 16 digits. Possibly, defined in one place (at least one place per language). What do you think?

tcclevenger:
@bartgol I agree. I can change the `Pi` constant in `physics_constants.hpp` to have full precision, but I imagine multiple p3 bfb tests will fail? Are they comparing against a fortran version like shoc, or a baseline on master?

bartgol:
I don't know what the physics unit tests do. @jgfouca might.

tcclevenger:
@bartgol It looks like from my own machine we will have failing p3 tests if we change the value of pi (makes sense). I can think of 3 options:
1. Update p3 and p3 tests to use more accurate pi in double precision. Not sure how a more accurate pi is expected to change the computations there, and how significant it its.
2. Implement pi in shoc to match the precision in p3 (ie, have `pi=3.41592650000000` in DP)
3. Have two different definitions of pi. Either `p3-pi` and `shoc-pi`, or `sp-pi` and `dp-pi`.

Thoughts?

tcclevenger:
@ambrad I think I like that solution. Just updated this PR.

ambrad (approved):
Yeah, this works: SP specification of pi, but with _rtype suffix.

bartgol:
I had to do a master merge, because of missing mappy testing scripts.

tcclevenger:
@bartgol @ambrad 
Compiling successfully on all machines. Tests failing as expected from rel_diff since we are changing pi by `~1e-3`. The max rel_diff in the failing tests are around `1e-4`. This will need to be manually merged.

PR #678, opened 2020-10-13
Remove surface emissions from F2010 SCREAM use_cases

crterai:
Because SCREAM simulations are running with prescribed aerosols, surface emissions of aerosol or chemical species is unnecessary. These are removed in this PR.

### DIFF
--- a/components/cam/bld/namelist_files/use_cases/2010_scream_hr.xml
+++ b/components/cam/bld/namelist_files/use_cases/2010_scream_hr.xml
@@ -83,53 +83,6 @@
 <!-- Energy fixer options -->
 <ieflx_opt  > 0     </ieflx_opt>
 
-<!-- External forcing for BAM or MAM.  CMIP6 input4mips data -->
-<!-- COPIED FROM "CAM" USE CASE; TODO: IS THIS NECESSARY FOR SCREAM? -->
-<ext_frc_type         >CYCLICAL</ext_frc_type>
-<ext_frc_cycle_yr     >2010</ext_frc_cycle_yr>
-<so2_ext_file         >atm/cam/chem/trop_mozart_aero/emis/DECK_ne30/cmip6_mam4_so2_elev_1x1_2010_clim_c20190821.nc </so2_ext_file>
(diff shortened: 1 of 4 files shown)


components/cam/bld/namelist_files/use_cases/2010_scream_hr.xml
```diff
@@ -145,7 +98,7 @@
 <drydep_list            >'H2O2', 'H2SO4', 'SO2'</drydep_list>
 
 <!-- sim_year used for CLM datasets and SST forcings -->
-<sim_year>2015</sim_year>
+<sim_year>2010</sim_year>
```
brhillman, line 101: Doesn't this mean we are going to use 2010 SST data?

components/cam/bld/namelist_files/use_cases/2010_scream_hr_dyamond2.xml
```diff
@@ -151,7 +104,7 @@
 <drydep_list            >'H2O2', 'H2SO4', 'SO2'</drydep_list>
 
 <!-- sim_year used for CLM datasets and SST forcings -->
-<sim_year>2015</sim_year>
+<sim_year>2010</sim_year>
```
brhillman, line 107: As above, is this correct for DYAMOND?

components/cam/bld/namelist_files/use_cases/2010_scream_hr.xml
crterai: No, I think we still end up using this SST (sst_HadOIBl_bc_1x1_clim_c101029.nc).

components/cam/bld/namelist_files/use_cases/2010_scream_hr_dyamond2.xml
crterai: Yeah, this is what we'd been wanting to set (2010, rather than 2015).The IC file and CLM files are actually defined in the use_case files, so it actually won't have an impact.

components/cam/bld/namelist_files/use_cases/2010_scream_hr.xml
crterai: Also, changing this parameter came back with BFB results for ne4 tests run before and after this set of changes, which makes me think that setting sim_year has no impact on simulations. I wanted to set it to 2010 so that it's consistent with our forcing choices.

brhillman (approved):

PR #681, opened 2020-10-14
Quick fix to timestamp.to_string() routine.

AaronDonahue:
Added logic for the case when the hour, minute or second in the time
is between 1-9.

Also switches the timestamp string order to YYYY-MM-DD HH:MM:SS
i.e. year first not last.

Before: this would print single digit for the string
Now: prints 2 digits starting with 0.

This is helpful when creating filenames for writing output.

ex. If HH-MM-SS = 1-21-1
before: 1:21:1
now: 01:21:01

note that IO will strip the ":" for the filenames so in the above example we would have a string 1211 vs 012101.  The former could be 1-21-1, or 12-1-1, or 1-2-11.  There is no ambiguity in the latter.

### DIFF
--- a/components/scream/src/share/util/scream_time_stamp.cpp
+++ b/components/scream/src/share/util/scream_time_stamp.cpp
@@ -94,9 +94,13 @@ std::string TimeStamp::to_string () const {
   const int m = (ss % 3600) / 60;
   const int s = (ss % 3600) % 60;
   const std::string zero = "00";
+  const std::string single = "0";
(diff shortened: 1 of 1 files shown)


jeff-cohere (approved):
👍

AaronDonahue:
@jeff-cohere and @bartgol , just pushed a commit with a cleaner implementation (As suggested offline by Luca).  This probably needs a re-review for the autotester to be happy.

bartgol (approved):
Looks cleaner. Definitely immediate to parse for people who know a bit about std streams.

jeff-cohere (approved):

PR #682, opened 2020-10-14
porting lbycond function

xyuan:
porting lbycond function

[BFB]

### DIFF
--- a/components/scream/src/physics/shoc/shoc_diag_second_moments_lbycond_impl.hpp
+++ b/components/scream/src/physics/shoc/shoc_diag_second_moments_lbycond_impl.hpp
@@ -23,7 +23,7 @@ void Functions<S,D>
   // thermodynamic variances and covariances are computed
   // according to that of Andre et al. 1978.
 
-  const auto ufmin = 0.01;
+  const Scalar ufmin = 0.01;
 
   auto uf = std::sqrt(ustar2+sp(0.3)*wstar*wstar);
   uf = ekat::impl::max<Scalar>(ufmin,uf);
--- a/components/scream/src/physics/shoc/shoc_diag_second_moments_lbycond_impl.hpp
+++ b/components/scream/src/physics/shoc/shoc_diag_second_moments_lbycond_impl.hpp
@@ -40,7 +40,8 @@ void Functions<S,D>
   uw_sec   = uw_sfc;
   vw_sec   = vw_sfc;
 
-  wtke_sec = ekat::impl::max<Scalar>(sqrt(ustar2),ufmin)*ekat::impl::max<Scalar>(sqrt(ustar2),ufmin)*ekat::impl::max<Scalar>(sqrt(ustar2),ufmin);
+  auto max_val = ekat::impl::max(sqrt(ustar2),ufmin);
+  wtke_sec = max_val*max_val*max_val;
 }
 } // namespace shoc
 } // namespace scream
--- a/components/scream/src/physics/shoc/shoc_diag_second_moments_lbycond_impl.hpp
+++ b/components/scream/src/physics/shoc/shoc_diag_second_moments_lbycond_impl.hpp
@@ -40,7 +40,7 @@ void Functions<S,D>
   uw_sec   = uw_sfc;
   vw_sec   = vw_sfc;
 
-  auto max_val = ekat::impl::max(sqrt(ustar2),ufmin);
+  auto max_val = ekat::impl::max<Scalar>(sqrt(ustar2),ufmin);
   wtke_sec = max_val*max_val*max_val;
 }
 } // namespace shoc
--- a/components/scream/src/physics/shoc/shoc_diag_second_moments_lbycond_impl.hpp
+++ b/components/scream/src/physics/shoc/shoc_diag_second_moments_lbycond_impl.hpp
@@ -6,20 +6,42 @@
 namespace scream {
 namespace shoc {
 
-/*
- * Implementation of shoc diag_second_moments_lbycond. Clients should NOT
- * #include this file, but include shoc_functions.hpp instead.
- */
-
 template<typename S, typename D>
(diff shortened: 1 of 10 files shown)


components/scream/src/physics/shoc/shoc_diag_second_moments_lbycond_impl.hpp
```diff
@@ -6,20 +6,42 @@
...
+  qwthl_sec = sp(0.2)*sp(1.8)*(wthl_sfc/uf)*(wqw_sfc/uf);
+
+  // Vertical fluxes of heat and moisture, simply
+  // use the surface fluxes given by host model
+  wthl_sec = wthl_sfc;
+  wqw_sec  = wqw_sfc;
+  uw_sec   = uw_sfc;
+  vw_sec   = vw_sfc;
+
+  wtke_sec = ekat::impl::max<Scalar>(sqrt(ustar2),ufmin)*ekat::impl::max<Scalar>(sqrt(ustar2),ufmin)*ekat::impl::max<Scalar>(sqrt(ustar2),ufmin);
```
jeff-cohere, line 43: You can simplify this with `cube` unless that doesn't work here for some reason.

xyuan: cube function doesn't support Scalar argument at the moment.

tcclevenger: Would it be cleaner to do
```
max_val = ekat::impl::max(sqrt(ustar2),ufmin);
wtke_sec = max_val*max_val*max_val;
```
Just a thought, feel free to disregard if you prefer otherwise.

components/scream/src/physics/shoc/shoc_diag_second_moments_lbycond_impl.hpp
```diff
@@ -6,20 +6,42 @@
...
-}
+  // Purpose of this subroutine is to diagnose the lower
+  // boundary condition for the second order moments needed
+  // for the SHOC parameterization.
+  // The thermodymnamic, tracer, and momentum fluxes are set
+  // to the surface fluxes for the host model, while the
+  // thermodynamic variances and covariances are computed
+  // according to that of Andre et al. 1978.
+
+  const auto ufmin = 0.01;
```
tcclevenger, line 26: Using `auto` here evaluates to a `double` always? @jeff-cohere is this correct? Would this create any issues in SP for `uf = ekat::impl::max<Scalar>(ufmin,uf);` below?

tcclevenger (commented):
I left some comments, but most are questions I have and may not be incorrect for this PR.

components/scream/src/physics/shoc/shoc_diag_second_moments_lbycond_impl.hpp
jeff-cohere: I believe `auto` does produce `double` here, yeah. Good catch. Probably safer to use `Scalar` here.

components/scream/src/physics/shoc/shoc_diag_second_moments_lbycond_impl.hpp
xyuan: I followed the fortran code in C++, not sure whether we want to make these changes. I guess it depends how people likes

components/scream/src/physics/shoc/shoc_functions_f90.cpp
```diff
@@ -713,8 +713,9 @@ void shoc_pblintd_init_pot(SHOCPblintdInitPotData& d)
 
 void diag_second_moments_lbycond(DiagSecondMomentsLbycondData& d)
 {
-  shoc_init(1, true);  // Single level routine
-  diag_second_moments_lbycond_c(d.shcol(), d.wthl_sfc, d.wqw_sfc, d.uw_sfc, d.vw_sfc, d.ustar2, d.wstar, d.wthl_sec, d.wqw_sec, d.uw_sec, d.vw_sec, d.wtke_sec, d.thl_sec, d.qw_sec, d.qwthl_sec);
+  shoc_init(64, true);  // Single level routine
```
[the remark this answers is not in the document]
xyuan: Using of 64 is to make the loop that iterates more times in order to test the code.

components/scream/src/physics/shoc/shoc_diag_second_moments_lbycond_impl.hpp
xyuan: Right, it didn't create issues in the test with different precision, it may involve implicit type conversion here. thanks for this.

components/scream/src/physics/shoc/shoc_diag_second_moments_lbycond_impl.hpp
jgfouca: I recommend what @tcclevenger said

jgfouca (changes requested):
Just need to clean up the cube.

xyuan:
@jgfouca @jeff-cohere @tcclevenger reversion made to the cube function, and const auto variables.

jgfouca (approved):

jeff-cohere (approved):

jeff-cohere:
Strange. The tests all pass on mappy, but the AT still reports a failure there.

xyuan:
I am trying to reproduce it on NERSC using Intel compiler

xyuan:
@jgfouca Could you merge this PR to master? It passed the test, however it needs further inspection, not sure what that is? (see below).
All Jobs Finished; status = PASSED, However Inspection must be performed before merge can occur...

xyuan:
@jeff-cohere @jgfouca @tcclevenger Please further review the new commits, this PR passed the AT test, it needs your review of the new commitment before it can be merged.

PR #683, opened 2020-10-14
SHOC fix to feed updated water vapor to Obkuhov/PBL height routines

bogensch:
While making PR #664 it was discovered that SHOC was feeding the Okuhov length and PBL height routines vapor from the qtracer array.  While this is technically the water vapor, this is dangerous and a bad idea for several reasons.  

This PR computes the vapor based on updated SHOC prognostic total water mixing ratio and the diagnostic cloud water mixing ratio and feeds the routines this updated value.  This PR will not be B4B but produces scientifically similar results (as verified by SCM and a short ne30 simulation).

NOTE that this PR adds a new subroutine (shoc_compute_vapor), which will be added to the SHOC conversion table (and a property test PR will follow) upon the merging of this PR.


xyuan (approved):
This changes sounds good to me.

bogensch:
@PeterCaldwell this PR has been approved and tests are passing/failing as expected.  SCM and short ne30 tests show this change has minimal effect on simulations.  I consider it ready for manual merge.

PR #684, opened 2020-10-15
Prevent evaporate_rain from over-depleting qr

PeterCaldwell:
evaporate_rain wasn't checking whether it was removing more qr than existed to start with. Also added property unit test to check that this doesn't happen again.


JS-WRF-SBM (approved):
Yep, that's what I've added to my code.

PR #685, opened 2020-10-15
Adds support for scorpio in a standalone build

bartgol:
So far it's not used by any scream lib/exec. I manually tested it in another branch (the scorpio integration work by @AaronDonahue ), and it seemed to work. I'd like to get this PR in separated from scorpio-ad work, to keep that PR more focused.

Note: this PR should not affect any outstanding branch other than scorpio-ad integration.


AaronDonahue:
@bartgol , it looks like this PR still has some of the IO work that will be introduced in PR #576 .  But it's old code.  Is this going to result in any conflicts down the road?

bartgol:
@AaronDonahue argh! Let me fix this.

PR #687, opened 2020-10-16
Add routine to compute CAPE and CIN diagnostics

bogensch:
Adds routine to compute Convective Available Potential Energy (CAPE) and Convective Inhibition (CIN).  Assumes an undiluted parcel.  Results from this diagnostic have been validated to be reasonable in SCM and short ne30 simulations.  

NOTE:  This PR renames the preexisting "CAPE" diagnostic in the ZM convection scheme to "CAPE_ZM" and the preexisting "CIN" diagnostic in the UW shallow convection scheme to "CIN_SH".  The new diagnostics now take the names "CAPE" and "CIN".


bogensch:
@PeterCaldwell can you take a look where the call for this diagnostic was placed?  I was unsure where to place it so I put it in "diag_conv" in cam_diagnostics.F90.  Not sure if you were thinking of a better home?

PeterCaldwell:
RE placement of diagnostic: Since CAPE and CIN are predictors of the convection which should occur, one might expect them to be calculated using the model state *before* convection is triggered. Since deep convection is explicitly handled, however, it is a bit unclear when in the call sequence this would be. Before dynamics (which I think we can approximate by "after radiation")? Before SHOC? Or maybe it doesn't matter because our timesteps are 1.5 min at ne1024? Right now these diags are between microphys and radiation. I guess let's leave them as-is since I don't have a better idea...

you put CAPE and CIN calcs after SHOC and microphysics but before radiation. This seems as reasonable as anything.

PeterCaldwell:
@pressel - could you review this soon?

components/eam/src/physics/cam/cam_diagnostics.F90
```diff
@@ -2393,4 +2412,320 @@ subroutine diag_state_b4_phys_write (state)
...
+
+   !  Determine indicee of PBL height, first set up required height arrays
+      do i = 1, ncol
+         zs(i) = phis(i)*rgravit
+         zf(i,pver+1) = zi(i,pver+1) + zs(i)
+      end do
+
+      do k = 1,pver
+         do i = 1, ncol
+            z(i,k) = zm(i,k) + zs(i)
```
crterai, line 2510: @bogensch Is this to account for one of the heights being in terms of sea-level and the other being in terms of height above surface? In this case, is z supposed to be height wrt sea-level, zm is wrt to the surface, and pblh is wrt sea-level?

bogensch: Ah, yes, this section is unnecessarily complicated.  I will be pushing a commit here shortly to clean this up and with clearer comments.  In short, the index of the PBL height can be computed from standard zm and zi heights (wrt to surface).  A height computed wrt to sea-level still needs to be computed for later on in the code for the MSE.  I have verified that these new commits produce b4b CAPE and CIN diagnostics, as expected, for a SCM case over land.

components/eam/src/physics/cam/cam_diagnostics.F90
```diff
@@ -2393,4 +2412,320 @@ subroutine diag_state_b4_phys_write (state)
...
+         do i = 1,ncol
+            if (capeten(i,n) > cape(i)) then
+               cape(i) = capeten(i,n)
+               lel(i) = lelten(i,n)
+            end if
+         end do
+      end do
+
+   ! Compute CIN based on information of levels computed above, which
+   !  is the buoyancy integrated from surface to the lauching level
```
crterai, line 2685: Going by this, shouldn't the launch level be mx rather than lcl in L2693? 

But taking a step back, I understood CIN to be the energy required to lift a parcel from the launch level to the level of free convection, rather than from the surface to LCL. But perhaps there's a different definition?

crterai (commented):
The code looks good and consistent with the CAPE calculated in the zm_conv.F90. I found the added comments particularly helpful. 

I did have some questions about the definition of what CAPE and CIN should be though. Based on the formulation that was taken from the zm code, CAPE is defined as the integral from the launching level to the level of neutral buoyancy, rather than the level of free convection to the level of neutral buoyancy. 
https://glossary.ametsoc.org/wiki/Convective_available_potential_energy
I think the level of free convection can be located as the level above the launching level where buoy goes from negative to positive. 
If we leave it as is, then some comments at the head of the routine will be helpful to note how the CAPE and CIN here are different from other definitions.

bogensch:
Thanks @crterai for a thorough review, I agree about defining a LFC.  Mostly I was using the UW shallow convection basis for defining CIN and ZM's def for defining CAPE.  Neither of which seem to use LFC.  But thinking about it now, those CAPE's and CIN's were used as closures, so not sure if they defined them differently for parametization purposes.  In any case, for diagnostic purposes I think it should be done right and I will make this consistent with traditional definition.

bogensch:
@crterai This is ready for re-review.  I validated that the results with modifications are reasonable in SCM and the distributions/magnitudes are also reasonable in a short ne30 simulation.

crterai (approved):
I think the changes look good!

PR #689, opened 2020-10-16
Minor P3 fixes from Koby

PeterCaldwell:
This is the first of 2 PRs making sure we have all the bugfixes Koby has found. Changes here are trivially minor.

### DIFF
--- a/components/cam/src/physics/cam/micro_p3.F90
+++ b/components/cam/src/physics/cam/micro_p3.F90
@@ -1022,9 +1022,6 @@ contains
 
       ! ice:
 
-      !BUG BELOW: *all* of this qi stuff needs to be incld. will fix in future PR.
-      call impose_max_total_ni(ni(k),max_total_Ni,inv_rho(k))
-
       qi_not_small:  if (qi(k).ge.qsmall) then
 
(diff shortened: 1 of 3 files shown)


components/cam/src/physics/cam/micro_p3.F90
```diff
@@ -890,7 +889,11 @@ SUBROUTINE p3_main_part2(kts, kte, kbot, ktop, kdir, do_predict_nc, do_prescribe
          is_hydromet_present = .true.
       endif
 
-      call impose_max_total_ni(ni(k),max_total_ni,inv_rho(k))
```
PeterCaldwell, line 893: @JS-WRF-SBM - your code doesn't have this change. I think max_total_ni must be an in-cloud quantity... or at least it can't simultaneously be a limit for in-cloud and cell-ave!

AaronDonahue (approved):
Looks good to me.

components/cam/src/physics/cam/micro_p3.F90
JS-WRF-SBM: Thanks!

JS-WRF-SBM (approved):

Issue #607, opened 2020-09-24
Build scorpio in SCREAM standalone mode.


bartgol:
It will allow to test scorpio stuff in all our PR testings and nightlies, rather than require manual cime-builds runs.

PeterCaldwell:
Accomplished via #690 . Closing.

Issue #583, opened 2020-09-17
[AD] Infrastructure to call SCORPIO from SCREAM


AaronDonahue:
Formalizing this task as a git-issue.

Establish the framework for calling scorpio routines from scream.  In particular the AD.  This task has a number of goals:

1. C++/Fortran bridges to call SCORPIO routines.  -DONE
2. Establish a SCORPIO class in SCREAM to manage all output streams. 
3. Establish a SCORPIO class in SCREAM to manage all input streams.
4. Hook up to scream-AD

PeterCaldwell:
Implemented via #690 . Closing.

PR #690, opened 2020-10-19
Scorpio Interface with SCREAM and stand-alone scorpio build

AaronDonahue:
This commit introduces the infrastructure to access SCORPIO routines within SCREAM.

Includes work from @bartgol which allows scorpio to be built within scream in stand-alone mode.  This is a major addition which allows unit-tests that use scorpio functionality to be run.

An output manager class is introduced which manages all output streams.  The output manager class serves as the single POC between the AD and output.

An output class which is managed by the above output manager takes care of opening and writing to output files as well as tracking field progress for output options like average output or min/max.

An input class handles all input.  In this first implementation it is assumed that all input will be single timesnap.  Future development can address cases where input files may contain more than one timesnap of data.

This commit gives restart and restart history support..

Moves the io files out of the mct_coupling folder and into the more logical share/io folder.

Creates unit tests for IO.

All control of input and output is handled via parameter list which can be either created on the fly or passed via YAML file.

This commit is BFB in that it will not impact any of the other processes or the solution output.  However it introduces a new set of output options.

Address #607 and #583 

[BFB]


AaronDonahue:
I noticed via the list of files changed that somehow this commit wants to make changes to items in the physics folder.  I'm guessing this happened when merging master with a branch built on an older version.  I'm working to fix this.

AaronDonahue:
@bartgol , can you specifically comment on whether or not this PR should be changing the scorpio submodule hash?  Not sure if this is intentional or also just some legacy from an older version.

bartgol:
> @bartgol , can you specifically comment on whether or not this PR should be changing the scorpio submodule hash? Not sure if this is intentional or also just some legacy from an older version.

I think it should. I think scream master was using an old scorpio version, which didn't include, e.g., a cmake fix to allow scorpio to be built as a subproject.

components/scream/src/share/io/tests/scorpio_tests.cpp
```diff
@@ -379,9 +385,9 @@ TEST_CASE("scorpio_interface_input", "") {
   get_variable(infilename,"index_3d","test value for 3d field",3,vec_xyz, PIO_INT,"xyz-int");
 
   // Create data to be written
-  std::array<Int,1> xdim = {10};
-  std::array<Int,1> ydim = {5};
-  std::array<Int,1> zdim = {2};
+  // std::array<Int,1> xdim = {10};
```
jeff-cohere, line 388: Any reason for commenting these lines out instead of just deleting?

components/scream/src/share/io/tests/scorpio_ad_test.cpp
```diff
@@ -0,0 +1,61 @@
+#include <catch2/catch.hpp>
```
bartgol, line 1: This test needs to go in `scream/tests`. You cannot use `scream_control` from the `src/share` folder, since it's built _after_ the share folder.

components/scream/src/share/io/tests/scorpio_tests.cpp
```diff
@@ -37,12 +44,12 @@ TEST_CASE("scorpio_interface_output", "") {
 
   int nerr = 0; // Set a record of the number of errors encountered.
   /* Create the set of SCORPIO output files and their respective dimensions and variables. */
-  int compid=0;  // For CIME based builds this will be the integer ID assigned to the atm by the component coupler.  For testing we simply set to 0
+  // int compid=0;  // For CIME based builds this will be the integer ID assigned to the atm by the component coupler.  For testing we simply set to 0
```
bartgol, line 47: Since these tests are not cime based, I think compid is not needed. If so, let's remove thise line.

components/scream/src/share/io/tests/scorpio_tests.cpp
bartgol, line 388: Can these lines be removed?

components/scream/src/share/io/tests/scorpio_yaml_test.cpp
```diff
@@ -0,0 +1,309 @@
...
+  input_type input_instance(io_comm,input_params);
+  input_instance.pull_input(*repo,*upgm); 
+  //input_instance.run(*repo); 
+  auto index_1d_dev = repo->get_field(fid_index_1d).get_view(); 
+  auto index_2d_dev = repo->get_field(fid_index_2d).get_reshaped_view<Real**>();
+  for (int ii = 0;ii<dims3d[0];++ii)
+  {
+    for (int jj=0;jj<dims3d[1];++jj)
+    {
+      printf("Rank %2d: %12.2f  %12.2f\n",io_comm.rank(),index_1d_dev(ii),index_2d_dev(ii,jj));
```
bartgol, line 199: The device views are not readable from host.

components/scream/src/share/io/tests/scorpio_yaml_test.cpp
```diff
@@ -0,0 +1,309 @@
...
+    }
+  }
+  
+  
+  //Finished with PIO, finalize the system
+  m_output_manager.finalize(l_time);
+
+  // Finalize the grids manager.
+  (*upgm).clean_up();
```
bartgol, line 208: Use `->` syntax here.

components/scream/src/share/io/tests/scorpio_yaml_test.cpp
```diff
@@ -0,0 +1,309 @@
...
+
+void update_data_output(Kokkos::View<Real*> data_1d, Kokkos::View<Real**> data_2d, Kokkos::View<Real***> data_3d, Real* x, Real* y, Real* z, const std::vector<int> dims3d, const Real t, const dofs_list_type gids)
+{
+  for (int ii=0;ii<dims3d[0];ii++)
+  {
+    Int g_ii = gids(ii);
+    data_1d(ii) = 0.1 * cos(x[g_ii]+t);
+    for (int jj=0;jj<dims3d[1];jj++) 
+    {
+      data_2d(ii,jj) = sin(y[jj]+t) * data_1d(ii);
```
bartgol, line 278: Again, dev view are not readable from host.

bartgol (changes requested):
I think there's still some work to do. I left some comments.

components/scream/src/share/io/scorpio_input.hpp
```diff
@@ -0,0 +1,258 @@
...
+inline void AtmosphereInput::finalize() 
+{
+  using namespace scream;
+  using namespace scream::scorpio;
+  eam_pio_closefile(m_filename);
+} // finalize
+/* ---------------------------------------------------------- */
+inline void AtmosphereInput::add_identifier(const FieldRepository<Real, device_type>& field_repo, const std::string name)
+{
+  for (auto myalias=field_repo.begin();myalias!=field_repo.end();++myalias)
```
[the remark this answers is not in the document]
AaronDonahue: field_repo.get_field returns the field itself and here I need the field_id.  I guess would it make sense to put a field_id getter function in the field_repository code?

AaronDonahue:
@bartgol and @jeff-cohere , thanks for the comments.  I'll get to work addressing them and report back with any responses.

components/scream/src/share/io/output_manager.hpp
```diff
@@ -0,0 +1,179 @@
...
+  using namespace scorpio;
+  if (!param_set and !gm_set and !repo_set) { return; }
+  // Make sure params, repo and grids manager are already set
+  EKAT_REQUIRE_MSG(param_set,"Error! Output manager requires a parameter list to be set before initialization");
+  EKAT_REQUIRE_MSG(gm_set,   "Error! Output manager requires a grids manager to be set before initialization");
+  EKAT_REQUIRE_MSG(repo_set, "Error! Output manager requires a field repo to be set before initialization");
+  // First parse the param list for output.
+  // Starting with the stride.  NOTE: Only a stride of 1 is supported right now.  TODO: Allow for stride of any value.
+  Int stride = m_params.get<Int>("PIO Stride");
+  EKAT_REQUIRE_MSG(stride==1,"Error! Output only supports a PIO Stride of 1");  // TODO: Delete when no longer valid.
```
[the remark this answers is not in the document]
AaronDonahue: We want to eventually develop IO in scream to allow for PIO_STRIDE != 1, but that isn't in the infrastructure yet.  Hence the error.  I wanted to build the stride in such a way that takes a variable PIO_STRIDE into consideration without dealing with the gather and broadcast logic that will be needed when PIO_STRIDE>1 just yet.  But I don't want to somehow allow a user mistake to sneak past with a different PIO STRIDE.

I like the default value option.

components/scream/src/share/io/scorpio_input.hpp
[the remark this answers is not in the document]
bartgol: You can get the id from the field though. All I'm saying is that these two for loops with checks on grid names could be replaced with 

```
auto f = repo.get_field(name, gname);
auto fid = f.get_header().get_identifier();
```

components/scream/src/share/io/output_manager.hpp
```diff
@@ -0,0 +1,179 @@
...
+    it.run(*m_device_field_repo, *m_grids_manager, current_ts);
+  }
+}
+/*-----------------------------------------------------------------------------------------------*/
+inline void OutputManager::run(Real current_ts)
+{
+  if (no_output) { return; }
+  for (auto& it : m_output_streams)
+  {
+    it.run(*m_device_field_repo, *m_grids_manager, current_ts);
```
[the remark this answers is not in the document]
AaronDonahue: Should I do the same for the field repo?

components/scream/src/share/io/scorpio_output.hpp
```diff
@@ -0,0 +1,485 @@
...
+    auto g_view = field_repo.get_field(f_map.second).get_view();
+    auto l_view = m_view.at(name); //TODO: may want to fix this since l_view may not be actually updated.  Use g_view for Instant?
+    Int  f_len  = f_map.second.get_layout().size();
+    // The next two operations do not need to happen if the frequency of output is instantaneous.
+    if (m_avg_type != "Instant")
+    {
+      // Update local view given the averaging type.  TODO make this a switch statement?
+      if (m_avg_type == "Average")
+      {
+        for (int ii=0; ii<f_len; ++ii) {l_view(ii) = (l_view(ii)*(m_status["Avg Count"]-1) + g_view(ii))/(m_status["Avg Count"]);}
```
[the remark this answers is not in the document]
AaronDonahue: So if I understand correctly, `l_view` in this context is on Host and `g_view` is on Device, right?  Since we are keeping a persistent copy of `l_view` since it is updated from step to step the correct approach is to keep `l_view` on host and copy `g_view` to host for this operation.  Right?

components/scream/src/share/io/scorpio_output.hpp
```diff
@@ -0,0 +1,485 @@
...
+    n_dim_len = dof_len/num_cols;
+    // Given dof_len and n_dim_len it should be possible to create an integer array of "global output indices" for this
+    // field and this rank. For every column (i.e. gid) the PIO indices would be (gid * n_dim_len),...,( (gid+1)*n_dim_len - 1).
+    std::vector<Int> var_dof(dof_len);
+    Int dof_it = 0;
+    for (int ii=0;ii<num_cols;++ii)
+    {
+      for (int jj=0;jj<n_dim_len;++jj)
+      {
+        var_dof[dof_it] =  m_gids(ii)*n_dim_len + jj;
```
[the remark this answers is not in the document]
AaronDonahue: Is it possible to copy this information to host?

components/scream/src/share/io/output_manager.hpp
```diff
@@ -0,0 +1,179 @@
...
+
+#include "ekat/mpi/ekat_comm.hpp"
+#include "ekat/ekat_parameter_list.hpp"
+#include "share/field/field_repository.hpp"
+#include "share/grid/grids_manager.hpp"
+
+namespace scream
+{
+
+class OutputManager
```
jeff-cohere, line 15: Can you add some comments that describe this class, its purpose, and each of its methods? It's easiest by far to write this stuff down when one is working on it.

components/scream/src/share/io/output_manager.hpp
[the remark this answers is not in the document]
jeff-cohere: If each `output_type` is used only by this OutputManager and not shared amongst any other entities (which is probably true), then it's safe and convenient to pass them the grids manager and the field repo to use. If they were shared among different entities/output managers/whatever, it would make more sense to pass the grids manager and the field repo as arguments to their `run` methods, since an output_type would be `run`ning for more than one entity in that case.

components/scream/src/share/io/scorpio_input.hpp
```diff
@@ -0,0 +1,258 @@
...
+  std::map<std::string,FieldIdentifier>  m_fields;
+  std::map<std::string,Int>              m_dofs;
+  std::map<std::string,Int>              m_dims;
+  dofs_list_type                         m_gids;
+
+  bool m_is_init = false;
+
+private:
+
+}; // Class AtmosphereIntput
```
jeff-cohere, line 71: In*put

components/scream/src/share/io/scorpio_input.hpp
```diff
@@ -0,0 +1,258 @@
...
+#include "share/field/field_header.hpp"
+#include "share/field/field.hpp"
+#include "share/field/field_identifier.hpp"
+
+#include "share/grid/grids_manager.hpp"
+
+namespace scream
+{
+
+class AtmosphereInput
```
jeff-cohere, line 23: A few explanatory comments here and for each method would go a long way to help others that encounter this type.

components/scream/src/share/io/scream_scorpio_interface.cpp
```diff
@@ -89,11 +94,33 @@ void register_dimension(const std::string &filename, const std::string& shortnam
   register_dimension_c2f(filename.c_str(), shortname.c_str(), longname.c_str(), length);
 }
 /* ----------------------------------------------------------------- */
+void get_variable(const std::string &filename, const std::string& shortname, const std::string& longname, const int numdims, const std::vector<std::string>& var_dimensions, const int dtype, const std::string& pio_decomp_tag) {
+
+  /* Convert the vector of strings that contains the variable dimensions to a char array */
+  const char* var_dimensions_c[numdims];
```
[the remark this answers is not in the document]
jeff-cohere: Yeah, this is one of the biggest differences between C (where you can do this!) and C++.

components/scream/src/share/io/output_manager.hpp
```diff
@@ -0,0 +1,179 @@
...
+}; // class OutputManager
+/*===============================================================================================*/
+/* Short function to add a new output stream to the output manager.  By making this a stand-alone
+ * function it is possible to add an output stream after initialization has been called.         */
+inline void OutputManager::new_output(const ekat::ParameterList& params)
+{
+  output_type output_instance(pio_comm,params);
+  output_instance.init(*m_device_field_repo,*m_grids_manager);
+  m_output_streams.push_back(output_instance);
+  if( no_output ) {no_output = false;}
```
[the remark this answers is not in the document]
jeff-cohere: If `no_output` exists to disable output even if `m_output_streams` is not empty, the behavior of this method should be documented. If `no_output` is just a state variable that keeps track of whether `m_output_streams` is empty (for example), I'd advise getting rid of it.

components/scream/src/share/io/output_manager.hpp
```diff
@@ -0,0 +1,179 @@
...
+    out_params.set<Int>("restart_hist_N",restart_hist_N);
+    out_params.set<std::string>("restart_hist_OPTION",restart_hist_OPTION);
+    new_output(out_params);
+  }
+}
+/*===============================================================================================*/
+// Overload run to allow for passing a TimeStamp or just directly a Real
+inline void OutputManager::run(util::TimeStamp& current_ts)
+{
+  if (no_output) { return; }
```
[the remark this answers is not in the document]
jeff-cohere: See my above comment.

components/scream/src/share/io/scorpio_output.hpp
```diff
@@ -0,0 +1,485 @@
...
+ * 16) When dynamics is ready, write output from a dynamics run.
+ * 17) Remove the F90 layer and call PIO C routines directly.
+ * 18) Add option to restart to write a restart at finalization, maybe if OUT_N = -1
+ * 19) take advantage of shared pointers field_repo and gridsmanager to streamline init, run and finalize interface.
+ */
+
+namespace scream
+{
+
+class AtmosphereOutput
```
jeff-cohere, line 50: Again, a brief description of the type and its purpose, and comments for its methods would be nice, and probably easier to do sooner rather than later.

components/scream/src/share/io/scorpio_output.hpp
```diff
@@ -0,0 +1,485 @@
...
+  using namespace scream::scorpio;
+
+  // Nothing to do at the moment, but keep just in case future development needs a finalization step
+
+  m_status["Finalize"] += 1;
+} // finalize
+/* ---------------------------------------------------------- */
+inline void AtmosphereOutput::check_status()
+{
+  printf("IO Status for Rank %5d, File - %.40s: (Init: %2d), (Run: %5d), (Finalize: %2d)\n",m_comm.rank(),m_casename.c_str(),m_status["Init"],m_status["Run"],m_status["Finalize"]);
```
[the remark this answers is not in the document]
jeff-cohere: You can also break strings over lines in C++ using quotes as delimiters. 

```
"this string" == "this" "string"

"this string" == "this"
"string"
```

components/scream/src/share/io/scorpio_output.hpp
```diff
@@ -0,0 +1,485 @@
...
+      // map_it<Field::FieldHeader::FieldIdentifier,Field>
+      for (const auto& it : map_it) {
+        auto& field_id = it.first;
+        if (it.first.get_grid_name()!=m_grid_name) {
+          continue;
+        }
+        // ok, this field is on the correct mesh. add it to pio output
+        m_fields.emplace(name,field_id);
+        // check to see if all the dims for this field are already set to be registered.
+        std::map<std::string,Int>::iterator tag_loc;
```
jeff-cohere, line 334: You can declare this in the assignment on line 339 and use `auto`.

components/scream/src/share/io/scream_scorpio_interface.hpp
```diff
@@ -2,25 +2,24 @@
...
-const int PIO_INT = 4;
+static constexpr int PIO_INT = 4;
 
-using scream::Real;
-using scream::Int;
 namespace scream {
 namespace scorpio {
 
-  void eam_init_pio_subsystem(const int mpicom, const int compid, const bool local);
+  void eam_init_pio_subsystem(const int mpicom);
```
jeff-cohere, line 22: Are these intended to be used by other teams/people? It might be nice to document them with some comments if so.

AaronDonahue:
Setting to WIP for now since the AT is failing and there are plenty of comments to work on.  Thank you to @bartgol and @jeff-cohere for going through the code.  I'll ping when I've pushed a new version with your comments addressed.

AaronDonahue:
@bartgol and @jeff-cohere , I've pushed a new commit that addresses many of your questions and reviewer comments.  I still need to add robust comments to the scorpio_output class.

One major outstanding question that I've left for my next round of fixes is handling data that is available to host or device or both.  @bartgol, you had two comments about local views and global ids (gids) not being available to host or device, or at least not being properly handled here.  Could you give me some guidance on how to proceed there to make things correct and obtain the same functionality?

jeff-cohere:
@AaronDonahue, we store our actual index and numeric data on "the device" so that we don't have to keep copying it there, which  incurs a sizeable penalty. The cost we pay for this "optimization" is this: before you access data like this on the host (inside a test, or inside a loop that isn't a Kokkos parallel thingy), you must

1. create a mirror view on the host for each view that stores data on the device
2. do a `Kokkos::deep_copy` from the device view to the (mirrored) host view

Additionally, if you're changing data on the host, in a test environment, say, you need to deep-copy it back to the device when you're done.

This is confusing and complicated, and it's an example of how Kokkos doesn't hide all this performance portability stuff from us after all. But once you start thinking about data living on the host or the device, mirror views and deep copies make sense as tools for getting it back and forth.

Did I understand your question correctly? Was all this obvious to you already?

AaronDonahue:
@jeff-cohere , no I think this answers it, thank you.  I wasn't sure if mirror views and deep copies would do the trick.  I'll reach out if I run into any issues.

components/scream/src/share/io/scorpio_input.hpp
```diff
@@ -0,0 +1,258 @@
...
+    n_dim_len = dof_len/num_cols;
+    // Given dof_len and n_dim_len it should be possible to create an integer array of "global input indices" for this
+    // field and this rank. For every column (i.e. gid) the PIO indices would be (gid * n_dim_len),...,( (gid+1)*n_dim_len - 1).
+    std::vector<Int> var_dof(dof_len);
+    Int dof_it = 0;
+    for (size_t ii=0;ii<m_gids.size();++ii)
+    {
+      for (int jj=0;jj<n_dim_len;++jj)
+      {
+        var_dof[dof_it] =  m_gids(ii)*n_dim_len + jj;
```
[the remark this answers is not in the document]
AaronDonahue: I think it is now, see around line 166

components/scream/src/share/io/scorpio_output.hpp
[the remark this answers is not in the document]
AaronDonahue: I believe this has been fixed now, see about 10 lines up.

components/scream/src/share/io/scorpio_output.hpp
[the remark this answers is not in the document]
AaronDonahue: see line 153

components/scream/src/share/io/scorpio.hpp
```diff
@@ -0,0 +1,376 @@
...
+  }
+
+  // Take care of updating and possibly writing fields.
+  for (auto const& f_map : m_fields)
+  {
+    // Get all the info for this field.
+    auto name   = f_map.first;
+    auto g_view = field_repo.get_field(f_map.second).get_view();
+    auto l_view = m_view.at(name);
+    Int  f_len  = f_map.second.get_layout().size();
```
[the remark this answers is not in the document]
AaronDonahue: @bartgol and/or @jeff-cohere is there a way to know that if a value is in the padded region?  Like does it take on a particular value, or is there a mask I can look up?

Follow-up, how would I disable and then re-enable FPE's for IO?

components/scream/src/share/io/scorpio.hpp
```diff
@@ -0,0 +1,376 @@
...
+    // The next two operations do not need to happen if the frequency of output is instantaneous.
+    if (m_avg_type != "Instant")
+    {
+      // If this is the start of a new averaging flag then init the local view.
+      if (avg_init) { for (int ii=0; ii<f_len; ++ii) {l_view(ii) = g_view(ii); }
```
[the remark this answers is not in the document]
AaronDonahue: If I create the host mirror at init time do I need to Kokkos::deep_copy(l_view_host,l_view_dev) each time to get the updated values?

AaronDonahue:
@jgfouca , I've added you to review just for my changes to the `machines_spec.py` file.

components/scream/scripts/machines_specs.py
```diff
@@ -16,49 +16,49 @@
 
 MACHINE_METADATA = {
     "melvin"   : (["module purge", "module load sems-env", "module load sems-gcc/7.3.0 sems-openmpi/1.10.1 sems-gcc/7.3.0 sems-git/2.10.1 sems-cmake/3.12.2 sems-python/3.5.2"],
-                  "$(which mpicxx)",
+                  "$(which mpicxx) -c CMAKE_Fortran_COMPILER=$(which mpifort) -c CMAKE_C_COMPILER=$(which mpicc)",
```
jgfouca, line 19: @AaronDonahue , I think with @bartgol 's recent improvements to test_all_scream, these changes are no longer necessary. You should upstream-merge latest master into your PR, remove these cmake changes, and then confirm that things still work.

jgfouca (changes requested):
machine_spec changes related to cmake args can be removed I think

components/scream/scripts/machines_specs.py
AaronDonahue: @jgfouca , Ok, I rebased my branch and was able to confirm that things work for syrah (and I'm guessing quartz too).  But Lassen also needs `$(which mpicc)` which isn't a part of the PR that Luca put in.  Any thoughts on how to move forward?  Should we add mpicc in the same way we did mpifort?  Or have an extra option to pass other compiler flags for the test?  At which point I can just add the `-c CMAKE_C_COMPILER=$(which mpicc)` line.

AaronDonahue: @bartgol ^^

jgfouca: @AaronDonahue , if we are compiling C files then we probably need to add explicit C support just as @bartgol did for f90. For now, is there a way you can hardcode Lassen to add what you need? If not, I can quickly repeat what Luca did for f90 for C.

AaronDonahue: I may be able to hardcode it for lassen for now.  But it may be good to expand support for C if possible, just in case other machines have a similar requirement.

AaronDonahue:
@bartgol , it looks like this PR is still failing the AT.  But I am having a hard time figuring out where it is going wrong from the AT output on github.  Is it possible that the AT machines need to load netCDF modules?  Are the tests failing during the build or during one of the unit tests?  I feel like we may be close to getting this bottled up and in the code, hopefully the testing failures are straightforward to address.

bartgol:
> @bartgol , it looks like this PR is still failing the AT. But I am having a hard time figuring out where it is going wrong from the AT output on github. Is it possible that the AT machines need to load netCDF modules? Are the tests failing during the build or during one of the unit tests? I feel like we may be close to getting this bottled up and in the code, hopefully the testing failures are straightforward to address.

Ah, shoes, I forgot to push the commit on your branch to load the correct modules. I will do it now. Sorry.

components/scream/scripts/machines_specs.py
bartgol: Should this PR take care of mpicc support, or should it be another PR?

AaronDonahue:
@bartgol , no worries. I just wanted to make sure there wasn't something missing from me.

components/scream/scripts/machines_specs.py
AaronDonahue: It should probably be another PR.  This PR is getting pretty big.  On a side note, I am planning a small PR for machine support that will allow me to get the >1 mpi rank tests to work on syrah, quartz and lassen.  I could put that in and then @bartgol or @jgfouca you could add the mpicc support to it if you want.

AaronDonahue: Also, for lassen I just added the mpicc support to the lassen.cmake file.  So if my PR is just left as is we will have the lassen tests running and passing without any extra changes to test-all-scream.

bartgol: @jgfouca I can do the mpicc one. Should be a 2 min thing.

AaronDonahue:
@bartgol , thanks for fixing the configuration for cmake.  It looks like we are at least getting test failures.  Something to build on top of.

I wonder if we (I) should squash the commits before we merge.  There are 60 at the moment...

bartgol:
Imho, it's not the number of commits that has to be considered. It is the logic units of work. If there are multiple commits on the same work unit, then sure, go ahead and squash them. What a 'work unit' is, is kind of up to you though.

AaronDonahue:
@bartgol , it looks like the autotester has passed the build and we are getting actual test failures.  Although I am confused by the failures themselves.  There appears to be an inconsistent set of failures for the two new scorpio based tests; io_test and restart_test.  And a number of failures for tests that should be unaffected by this PR.

I checked on the LC machines that I do have access too and after merging this branch in with master the tests all seem to pass.  Do you have access to the verbose test failure logs?  Maybe I could figure out whats wrong if I knew why the failures were occurring.

AaronDonahue:
@bartgol , thanks for putting the cmake fix in.  Hopefully this solves our AT failure issues.  I have another question, while I was going through this PR carefully to track down the issue I noticed that this PR reports changes in the scorpio submodule, I'm guessing just a change to the scorpio hash.  I'm pretty sure we don't want to do this.

Am I misinterpreting all the scorpio stuff at the bottom of the "files changed" portion of the PR?   If we are changing the scorpio stuff then how do we remove that change?  Or alternatively, should I make a new PR where I carefully pull out the files from this branch that we actually want to update and make a new PR?

bartgol:
@AaronDonahue to change the submodule sha you want to use in the branch, you need to checkout the sha you want in the submodule, then `git add externals/scorpio` and commit the change.

I think at the beginning I _did_ want to change scorpio version, since we needed some mods that went in their master, and e3sm/scream master had not yet updated the scorpio version. It might be that that's no longer the case. Either way, I think it's ok to use whichever is more between the sha in this branch and the sha in scream's master.

As a general rule, when facing a conflict in submodule's version during a merge, one is probably ok using the most recent one. That's not true if the newer version is not backward compatible (that happened for kokkos a few times), but I don't think scorpio would do that.

AaronDonahue:
@bartgol , I still don't have complete access to Sandia machines.  So I can't investigate the AT fails for this PR.  But looking at them I can comment:

-- LC --
As a sanity check, I tested the branch on LC machines.  All tests pass.  So my theory is the AT failures are machine specific and may require some adjustment to the machine files, or machine specs.

-- Mappy --
Fails just for io_test.  Not sure what the issue is here since it looks like the `restart_test` passes and it is pretty similar.  It doesn't show in the log on git,  but can you see if `io_test` passes with 2 mpi ranks?  That should be tested too, and if that one passes and the single rank fails it can give us something to look at.

-- Weaver --
It looks like the two new IO tests fail when using 2 mpi ranks rather than 1.  The fails say "not run".  I ran into a similar issue with LC machines, I had to make sure the `mpirun` command was set correctly for the machine, and then add a flag to the machine file.  That may be what is happening on weaver too.

-- Blake --
In terms of the new IO tests, it looks like all 4 fail, so this may be an issue with netCDF or something like that.  I can't imagine what is causing the other fails for tests not at all impacted by this PR.  One theory would be that there is some old stuff in this PR that merges into master, but I don't see any evidence of that in the "files changed" tab.  Plus, Weaver and Mappy don't fail in these tests.  ???

bartgol:
@AaronDonahue Let me take a quick look, then I'll report back.

bartgol:
On Mappy, the error looks like this:

```
PIO: FATAL ERROR: Aborting... FATAL ERROR: NetCDF: Index exceeds dimension bound (/home/e3sm-jenkins/jenkins-ws/workspace/SCREAM_PullRequest_Autotester_Mappy/281/scream/externals/scorpio/src/clib/pio_darray_int.c: 1498)
Obtained 10 stack frames.
./io_test(print_trace+0x2f) [0x8236cb]
./io_test(piodie+0x138) [0x82388f]
./io_test(check_netcdf+0x1ba) [0x823ba4]
./io_test(pio_read_darray_nc_serial+0x1305) [0x859a03]
./io_test(PIOc_read_darray+0x604) [0x855407]
./io_test() [0x815b65]
./io_test(__piodarray_MOD_read_darray_1d_double+0x17b) [0x817283]
./io_test(__scream_scorpio_interface_MOD_grid_read_darray_1d_real+0x200) [0x78ccc8]
./io_test(grid_read_data_array_c2f_real+0x123) [0x788d12]
```
Is this an issue with netcdf configuration (i.e., config time max sizes of something)? I know exodus (an i/o format based on netcdf) needs netcdf configured/compiled with a certain set of dimensions (different from netcdf) default ones. Maybe something similar is happening here? Or maybe this is just a runtime index error. I have to manually debug the exec on mappy to find out.

bartgol:
Fun fact: when I manually run the tests from the folder where the AT built them, with the same env the AT used, tests pass.

AaronDonahue:
@bartgol , I've verified that when merged with master this PR passes the full suite of unit tests on syrah which uses an Intel compiler.  I think that it is ready to merge from that perspective.

There are two conflicts that will need to be resolved.  But since Blake isn't working we will need to manually merge anyway.  We can handle those conflicts at that point, they are just in two cmake related files that were changed in a PR 2 days ago.

I suppose the last issue is that there are some stale reviews which I believe I addressed in subsequent versions of the code.  @bartgol , @jeff-cohere  and @jgfouca , when you have a chance can you give the code another once over.  If all goes well we might be able to finally get this merged :-)

Thanks to everyone for their help getting this PR ready.

AaronDonahue:
@jeff-cohere , @bartgol Just pinging the two of you to give this PR one more once over review.  It has consistently passed mappy and weaver and I manually verified the intel build passes on the LC machine syrah using gather-all-data.  Luca updated the branch to fix all merge conflicts.  So a re-review is the only thing standing in the way of merging the PR once-and-for-all.

jeff-cohere:
I'm in a bunch of meetings this morning, but I should be able to look at this mid-day. Does that work?

AaronDonahue:
> I'm in a bunch of meetings this morning, but I should be able to look at this mid-day. Does that work?

Absolutely, thanks

components/scream/src/share/io/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
+set(SCREAM_SCORPIO_SRCS
+  scream_scorpio_interface.F90
+  scream_scorpio_interface.cpp
+  scream_scorpio_interface_iso_c2f.F90
+#  ${SCREAM_BASE_DIR}/../cam/src/physics/cam/physics_utils.F90
```
jeff-cohere, line 5: Do we forsee needing this in the future? If not, we can delete this line.

components/scream/src/share/io/output_manager.hpp
```diff
@@ -0,0 +1,271 @@
...
+ * The OutputManager Class handles all individual output streams.
+ * Rather than have the SCREAM-AD or unit tests call the scorpio_output
+ * class objects directly, the OutputManager stores these in a vector,
+ * m_output_streams.
+ *
+ * Similar to a typical atmospheric process object, the OutputManager
+ * has an init, run and finalize routine which is called during the AD
+ * during those respective steps.
+ *
+ * init():
```
jeff-cohere, line 25: Consider moving method descriptions so they are above their methods. This might encourage people to document their own code similarly. :-)

components/scream/src/share/io/output_manager.hpp
```diff
@@ -0,0 +1,271 @@
...
+  std::shared_ptr<const GridsManager>          m_grids_manager;
+
+  bool                                 param_set = false;
+  bool                                 gm_set    = false;
+  bool                                 repo_set  = false;
+  bool                        m_runtype_restart  = false;
+
+}; // class OutputManager
+/*===============================================================================================*/
+/* Short function to add a new output stream to the output manager.  By making this a stand-alone
```
jeff-cohere, line 134: Not a big deal, but: a "standalone function" is typically different from a class method--it's just a function and not part of a class definition. I think you're trying to say that allowing someone to add a stream outside of the constructor is a useful thing, right?

components/scream/src/share/io/output_manager.hpp
```diff
@@ -0,0 +1,271 @@
...
+ * Note: new_output requires a parameter list with all of the output control information for this
+ * stream.  See scorpio_output.hpp for more information on what the parameter list needs.         */
+inline void OutputManager::new_output(const ekat::ParameterList& params)
+{
+  auto output_instance = std::make_shared<output_type>(pio_comm,params,m_device_field_repo,m_grids_manager,m_runtype_restart);
+  output_instance->init();
+  m_output_streams.push_back(output_instance);
+}
+/* --------------------------------------------------------------------- */
+inline void OutputManager::new_output(const ekat::ParameterList& params, const bool runtype_restart)
```
jeff-cohere, line 145: You could give `runtype_restart` a default value and get rid of the one-argument version of this method.

components/scream/src/share/io/output_manager.hpp
jeff-cohere: Also, consider sticking this comment above the declaration of the method within the class.

components/scream/src/share/io/scream_scorpio_interface.hpp
```diff
@@ -2,43 +2,61 @@
...
-const int PIO_INT = 4;
+static constexpr int PIO_INT = 4;
 
-using scream::Real;
-using scream::Int;
 namespace scream {
 namespace scorpio {
-
-  void eam_init_pio_subsystem(const int mpicom, const int compid, const bool local);
+  /* All scorpio usage requires that the pio_subsystem is initialized. Happens only once per simulation */
```
jeff-cohere, line 23: Not important at all, but you should feel free to use C++-style comments (//) instead of C ones (/* */), even in C code.

components/scream/src/share/io/tests/scorpio_tests.cpp
```diff
@@ -132,44 +136,44 @@ TEST_CASE("scorpio_interface_output", "") {
   get_dof(ekat::size(x_data), myrank, numranks, dof_x[0], xstart,xstop);
   std::vector<Int> x_dof(dof_x[0]);
   for (int ii=xstart,cnt=0;ii<=xstop;++ii,++cnt) {
-    x_dof[cnt] = ii+1;  // Add one to index since C++ starts at 0 and Fortran starts at 1
+    x_dof[cnt] = ii;  // Add one to index since C++ starts at 0 and Fortran starts at 1
```
jeff-cohere, line 139: Is the comment still relevant?

components/scream/src/share/io/scorpio_output.hpp
```diff
@@ -0,0 +1,637 @@
...
+ *  RESTART FILE is an optional boolean parameter that specifies if this output stream is a restart output, which is treated differently.
+ *
+ *  Usage of this class is to create an output file, write data to the file and close the file.
+ *  This class keeps a running copy of data for all output fields locally to be used for the different averaging flags.
+ * --------------------------------------------------------------------------------
+ *  (2020-10-21) Aaron S. Donahue (LLNL)
+ */
+
+/* There are a number of things that need to be cleaned up - in no particular order -
+ *--1) Gather dimension data from grid manager or field repo.  We may need to use field repo since grid manager
```
jeff-cohere, line 82: Are the items in this list reflected in GitHub issues? That's a better place for a TODO. At the very least, we can erase the entries here that are --DONE--

components/scream/src/share/io/scorpio_input.hpp
```diff
@@ -0,0 +1,385 @@
...
+
+#include "share/grid/grids_manager.hpp"
+/*  The AtmosphereInput class handles all input streams to SCREAM.
+ *  It is important to note that there does not exist an InputManager,
+ *  like in the case of output.  So all input streams have to be managed
+ *  directly by the process that requires it.
+ *
+ *  The typical input call will be to the outward facing routine 'pull_input'.
+ *
+ *  pullinput calls the three typical init, run and finalize routines in order
```
jeff-cohere, line 26: Consider moving this comment to the method declaration in the class, and possibly describing the other methods there.

jeff-cohere (approved):
Looks good. I left several comments, mostly related to documentation. They all could be addressed in a later PR, though.

components/scream/src/share/io/tests/scorpio_tests.cpp
AaronDonahue: Actually, scorpio_tests will be deleted.  I kept it around for reference, but given that this PR is ready to merge I have deleted it in the most recent version.

components/scream/src/share/io/scream_scorpio_interface.hpp
AaronDonahue: Good point.  I think this file will be significantly revised when we do away from the C++ -> F90 bridge.  That would be a good time to also update the comments and probably change their format.

components/scream/src/share/io/output_manager.hpp
AaronDonahue: I think I will save this comment for a future PR.  I'm still not sure how `new_output` will be used.  I think that we could probably get rid of the single argument version but I don't want to go down a path that might take longer than expected and delay this PR anymore.

AaronDonahue:
@jeff-cohere , thank you for taking the time to re-review.  I've fixed almost all of your comments and responded to a couple (see above).  I've pushed the new branch and will test one more time on syrah (just to be super sure that the Intel builds work).

@bartgol , I'm not sure if you would also like to give the code a once-over review.  But assuming you are okay with things I think we should be ready to merge once my syrah tests are finished.

AaronDonahue:
Syrah passes after merging the latest branch with master.

bartgol:
> @bartgol , I'm not sure if you would also like to give the code a once-over review. But assuming you are okay with things I think we should be ready to merge once my syrah tests are finished.

Your assumption is correct.

AaronDonahue:
@bartgol , the AT just finished and it looks like the tests have gone back to passing on mappy and weaver.  Still failing on blake, but passing on syrah and lassen (on my end).  Should we merge?

bartgol:
Let me give one last look at the output on blake. I see non-io tests failing, and I don't want master to be broken for all subsequent AT runs.

bartgol:
Other PRs are being merged, and they pass on blake, so something might really be off with this PR.

AaronDonahue:
@bartgol , thats a good plan.  Wouldn't want this PR to bring the AT to its knees.  I wonder why completely independent tests are failing with this PR...

I should be getting my yubikey from Sandia today which will allow me to get access to blake, at which point I may be able to help with some of this detective work.

AaronDonahue:
on the plus side it looks like the same tests fail for each test setup, dbg, sp_dbg, fpe

components/scream/src/share/io/tests/io.cpp
```diff
@@ -0,0 +1,294 @@
...
+  {
+    REQUIRE(std::abs(f1_hst(ii)-(max_steps*dt+ii))<tol);
+    for (int jj=0;jj<num_levs;++jj)
+    {
+      REQUIRE(std::abs(f3_hst(ii,jj)-(ii+max_steps*dt + (jj+1)/10.))<tol);
+    }
+  }
+  for (int jj=0;jj<num_levs;++jj)
+  {
+    REQUIRE(std::abs(f2_hst(jj)-(max_steps*dt + (jj+1)/10.)<tol));
```
bartgol, line 152: My editor suggests that there might be an error here, since we're taking the abs of a bool. Should this perhaps be:

```
REQUIRE(std::abs(f2_hst(jj)-(max_steps*dt + (jj+1)/10.))<tol);
```
Notice the parentheses at the end: from `<tol))` to `)<tol)`.

bartgol (changes requested):

components/scream/src/share/io/tests/io.cpp
AaronDonahue: Yes, that change is correct.  Weird that you can take `std::abs(bool)` without throwing an error.  @bartgol , were you planning on pushing a version of the branch now that you may have figured out was making Blake go crazy?  If so, I can wait to put this change in.  Otherwise I'll push the fix this morning.

AaronDonahue: nevermind, I just noticed that the AT is showing passing, so your fix must already be active and on the blake side of things.  I'll re-inspect the tests and make sure there are no other weird things like this.  Then push the changes.

AaronDonahue: @bartgol , I pushed a fix for this error.

bartgol (approved):

AaronDonahue:
I think this PR is just waiting on @jgfouca to clear his requested changes.  Then we can merge.

jgfouca (approved):

PR #691, opened 2020-10-19
Simplified Field class and surrounding datatypes.

jeff-cohere:
This PR comes from a discussion I had with Luca last week about how we might simplify Fields and make it more obvious how they're intended to be used.

It removes the Device template parameter from the Field class and its associated types. This expresses our decision to always store Field data on DefaultDevice (CPU for CPU-enabled calculations, and GPU for GPU-enabled
ones).

We also change the remaining template parameter from ScalarType to RealType in the Field and related classes, to reflect the fact that Fields are always stored as 1D arrays of real-valued numbers (and never Packs).

Hopefully, these two changes simplify the Field class, and make it easier to understand the relationships between Fields and their (possible reshaped) views.

In order to make these changes, I had to make very minor edits in the atmosphere process types and their P3/SHOC/ZM/Radiation subclasses. Apologies for the inconvenience.

### DIFF
--- a/components/scream/src/control/atmosphere_driver.cpp
+++ b/components/scream/src/control/atmosphere_driver.cpp
@@ -81,7 +81,6 @@ void AtmosphereDriver::initialize (const ekat::Comm& atm_comm,
                                    const ekat::ParameterList& params,
                                    const util::TimeStamp& t0)
 {
-  using device_type = DefaultDevice;
   m_atm_comm = atm_comm;
   m_atm_params = params;
   m_current_ts = t0;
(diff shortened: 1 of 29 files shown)


AaronDonahue (approved):
Looks pretty good to me.  It will be nice to remove the ScalarType declaration which was just one more thing to mentally parse when trying to understand what is going on with fields.  Dido on DeviceType.  Thanks for the simplification Jeff.

Final note, there will likely be more changes needed in a few of the outstanding AD PR's so we will want to time things appropriately.

jeff-cohere:
Is there any way we can disable the Travis CI builds? Seems like we inherited these from somewhere else.

bartgol:
> Is there any way we can disable the Travis CI builds? Seems like we inherited these from somewhere else.

I thought you were one of those in favor of "more" testing. =)

That build is part of e3sm. I think that, so far, it only checks that a case is built (and maybe submitted), nothing else. It fails sometimes I believe due to time limit running out (although the build reports 14m, which is well below the 1h limit of the free version).

Luckily (in this time, though in general is not great), the AT ignores all the status checks that are not part of the AT framework. You can still "require" them on github, which prevents the PR from being merged, but the AT won't see those checks, and end up crashing if merge is disabled.

tl;dr: we can disable travis tests in the scream repo (without affecting the e3sm repo), but having it on does not prevent our CI framework from working (although it shows a red X next to some merged PRs, even though they passed all _our_ tests).

jeff-cohere:
Thanks for the explanation. I am in favor of more testing options, but I'm opposed to more noise :-), so I think we should disable Travis CI builds in SCREAM if they're not doing anything useful for us.

bartgol (approved):
Thanks for taking care of this.

bartgol:
> Thanks for the explanation. I am in favor of more testing options, but I'm opposed to more noise :-), so I think we should disable Travis CI builds in SCREAM if they're not doing anything useful for us.

I agree. Unless the testing master @jgfouca has any objection, I will disable it.

jgfouca:
@bartgol I think we can disable travis since we are relying on AT now.

bartgol:
> @bartgol I think we can disable travis since we are relying on AT now.

To be clear, this is not the travis testing we set up in scream long time ago. This was set up in e3sm, to perform some minimal deployment testing. But I agree that we don't need it.

singhbalwinder (approved):
Looks good to me! Thanks Jeff!

brhillman (approved):
Looks great. Love the simplification. Thank you for the ping on this.

PR #692, opened 2020-10-19
Big update to boiler plate

jgfouca:
Change list:
1) Add new PhysicsTestDataGeneric struct, a completely general struct for handling any assortment of MD arrays of Reals and Ints.
2) Implement PhysicsTestData as a specialization of PhysicsTestDataGeneric
3) Add more tests for new PhysicsTestData* capabilities
4) Update gen-boiler to handle more-general fortran routine data, it will use PhysicsTestDataGeneric instead of PhysicsTestData if it detects that the data is too complex for the simpler case (PhysicsTestData).
5) Test these new capabilities by generating boiler for update_prognostics_implicit, which is a shoc function whose data is too complex for PhysicsTestData.
6) Other fixes for gen-boiler problems discovered by @bogensch 
7) Add support for 3D data to PhysicsTestData if the 3d data is dim1xdim2xdim3


bogensch:
@jgfouca is this ready to review or should we wait until tests are passing?

components/scream/src/physics/shoc/tests/shoc_unit_tests_common.hpp
```diff
@@ -113,6 +113,7 @@ struct UnitWrap {
     struct TestDiagSecondShocMoments;
     struct TestPblintdCldCheck;
     struct TestComputeShocVapor;
+    struct TestUpdatePrognosticsImplici;
```
singhbalwinder, line 116: I think `t` is missing in this `struct` name. It is missing everywhere, so it is consistent. It doesn't matter code wise but it will be good to be consistent with the long name of the function.

jgfouca: @singhbalwinder , thanks, I will fix.

jgfouca:
These fails are unexpected since the PR seemed to work fine on mappy. I will investigate.

bogensch (approved):
Thanks for doing this!

PR #693, opened 2020-10-19
add c++ error function call

tcclevenger:
Have shoc.F90 function call std::erf instead of the version defined in `cam/src/utils/error_function.F90`.

### DIFF
--- a/components/cam/src/physics/cam/shoc.F90
+++ b/components/cam/src/physics/cam/shoc.F90
@@ -2615,7 +2615,7 @@ subroutine shoc_assumed_pdf_compute_s(&
   real(rtype), intent(out) :: C
 
   ! local variables
-  real(rtype) :: cthl, cqt, tmp_val
+  real(rtype) :: cthl, cqt, tmp_val, error_func_out
 
   ! Parameters
   real(rtype) :: sqrt2, sqrt2pi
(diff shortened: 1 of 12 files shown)


components/cam/src/physics/cam/shoc.F90
```diff
@@ -2735,6 +2737,31 @@ subroutine shoc_assumed_pdf_compute_buoyancy_flux(&
 
 end subroutine shoc_assumed_pdf_compute_buoyancy_flux
 
+! Calls the error function. If use_cxx=true, then we call std::erf()
+subroutine error_function(input, output)
```
bartgol, line 2741: You could make this a function instead of a subroutine, which would make its use a bit more similar to before (i.e., allow to do `c=error_function(blah)`).

Also, along the line of other bfb stuff, why not calling it simply bfb_erf?

components/scream/src/physics/shoc/shoc_error_function_impl.hpp
```diff
@@ -0,0 +1,25 @@
...
+ * Implementation of shoc error_function. Clients should NOT
+ * #include this file, but include shoc_functions.hpp instead.
+ */
+
+template<typename S, typename D>
+KOKKOS_FUNCTION
+void Functions<S,D>::error_function(const Scalar &input,
+                                    Scalar       &output)
+{
+  output = std::erf(input);
```
bartgol, line 19: I think this might not compile/work on gpu. Or, at the very least, not give bfb results with the host impl. I think you might need something similar to `cxx_sqrt` for erf.

tcclevenger: I think you are right. It compiles on GPU, but I believe my test doesn't test bfb

tcclevenger:
New pull request: https://github.com/E3SM-Project/scream/pull/695. New approach similar to `cxx_sqrt/bfb_sqrt`

PR #695, opened 2020-10-20
add bfb_erf

tcclevenger:
Adds `bfb_erf` for `shoc_assumed_pdf` porting.

### DIFF
--- a/components/eam/src/physics/cam/shoc.F90
+++ b/components/eam/src/physics/cam/shoc.F90
@@ -18,7 +18,7 @@ module shoc
 ! Bit-for-bit math functions.
 #ifdef SCREAM_CONFIG_IS_CMAKE
   use physics_share_f2c, only: cxx_pow, cxx_sqrt, cxx_cbrt, cxx_gamma, cxx_log, &
-                               cxx_log10, cxx_exp
+                               cxx_log10, cxx_exp, cxx_erf
 #endif
 
 implicit none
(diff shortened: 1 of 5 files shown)


tcclevenger:
@bartgol @jeff-cohere 
All CPU tests pass, GPU has `shoc_run_and_cmp_ut_np1_omp1` failing. Like other `bfb_*()` changes, `host_dse` sees the largest `Max rel diff ~ 0.1`. Again, I went to weaver and calculated the rel diff of the values computed with `bfb_erf` (only one instance, 
`C=0.5_rtype*(1._rtype+bfb_erf(s/(sqrt2*std_s)))`):
```
rel_diff = abs(new_C - old_C)/abs(old_C) < 8.6e-05 (SP) 2.3e-13 (DP)
```

components/eam/src/physics/cam/shoc.F90
```diff
@@ -2635,7 +2635,7 @@ subroutine shoc_assumed_pdf_compute_s(&
   C=0._rtype
 
   if (std_s .ne. 0.0_rtype) then
-    C=0.5_rtype*(1._rtype+erf(s/(sqrt2*std_s)))
+    C=0.5_rtype*(1._rtype+bfb_erf(s/(sqrt2*std_s)))
     IF (C .ne. 0._rtype) qn=s*C+(std_s/sqrt2pi)*exp(-0.5_rtype*bfb_square(s/std_s))
```
jeff-cohere, line 2639: There's an `exp` buried in here.

tcclevenger: Ahh good catch!

bartgol (approved):

jeff-cohere (approved):

PR #697, opened 2020-10-22
PR 3897 from E3SM

ambrad:
See https://github.com/E3SM-Project/E3SM/pull/3897

Making a separate PR here to get this in for the DYAMOND2 runs.

Testing:
* I ran e3sm_scream against baselines from today with compy_intel and all pass. I confirmed all ten tests are sensitive to the advance_volc_radiation_data routine.
* e3sm_atm_integration e3sm_developer pass except for the currently failing hommexx test.
* The standalone Homme suite, non-BFB build, passes on Blake (SKX) with -O3 -fp-model strict but is non-BFB if using -fp-model fast. The changes in imex_mod.F90 do not affect floating point arithmetic, but evidently they do induce the optimizer to change something.

[BFB]

### DIFF
--- a/components/eam/src/chemistry/utils/mo_util.F90
+++ b/components/eam/src/chemistry/utils/mo_util.F90
@@ -5,7 +5,7 @@ module mo_util
   implicit none
 
   private
-  public :: rebin
+  public :: rebin, rebin_fast
 
 contains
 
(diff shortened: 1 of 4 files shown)


ambrad:
I'm holding off on getting reviews until the inconsistent ref issue on master is resolved.

ambrad:
We'll probably keep master as is. This PR will bring in the proper refs for these commits from upstream. @bartgol I included you as a reviewer so you could OK that. @PeterCaldwell I included you so you can decide whether you want these speedups here before the corresponding E3SM PR gets merged into E3SM.

PeterCaldwell:
I can't see any reason why we wouldn't want these changes for DYAMOND2 but have a few clarifying questions:
1. do we have a sense of the impact of these changes on timing? If negligible, including in DYAMOND2 is less of a priority.
2. merging these changes into SCREAM now then having them show up in an e3sm upstream merge won't cause a lot of human work, right?
3. the test functions you added aren't called anywhere inside the e3sm model, right? I was surprised to see them because I haven't seen tests included in the code like that before (but it seems like a good idea).

bartgol (approved):
The fact that the PR doesn't show any other change other than in the files you touched in the last 4 commits is reassuring. The different sha's for the previous commits are probably due to something as little as a different commit message/author, that caused the sha of that chain of commits to be different. I suspect our next upstream merge will again show something like this (meaning, repeated commits), but hopefully that will be it.

All of this is annoying (the git history looks _much_ messier), but it is overall a benign issue.

P.S.: If anyone has a plausible theory for how this happened, I am really interested in hearing it.

singhbalwinder:
Thanks @ambrad for fixing the indexing issue. It was an oversight on my part. My guess is that it should make a non-negligible difference in performance but it is hard to say anything as E3SM has a lot of moving parts.

I left some comments/questions on E3SM PR. They are not important to address but it will be nice to add error checking if we are revisiting these codes. I added some comments for myself but this work was done in a rush so I forgot to address those comments to myself.

ambrad:
@singhbalwinder I removed the comments as you requested and updated both PRs.

@PeterCaldwell:

1. It speeds up the volcaero code by very roughly 3x. That code appears to have increased the CAM_run1 timer in the ne1024 runs by ~10%.
2. None. They have the same SHA1s, so a git merge will pass over them.
3. They are not hooked up b/c there isn't infrastructure for them. But better to have a unit test that can be hooked up into hypothetical future infrastructure than none at all. This file builds standalone if you add a def for r8 and a program main. That's how I run the unit test.

singhbalwinder:
Thanks @ambrad ! That is considerable speedup.

PeterCaldwell (approved):
This is great, thanks!

PR #702, opened 2020-10-23
Fix setting RUN_STARTDATE for SCREAM compsets

brhillman:
It looks like our setting of RUN_STARTDATE for the DYAMOND2 compsets somehow got lost in the shuffle of the recent upstream merge. Thus, when creating new cases using the DYAMOND compsets the run startdate would default to 0001-01-01, which is not what we want!

[B4B] (for everything other than DYAMOND compsets)

### DIFF
--- a/components/eam/cime_config/config_compsets.xml
+++ b/components/eam/cime_config/config_compsets.xml
@@ -576,6 +576,10 @@
 	<value  compset="C2R4_EAM">2004-01-01</value>
 	<value  compset="AR95_EAM">1995-07-18</value>
 	<value  compset="AR97_EAM">1997-06-19</value>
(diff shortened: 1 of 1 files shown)


PeterCaldwell (approved):
Looks good, thanks! I assume you tested that this fixed the problem.

PeterCaldwell:
This PR won't affect any of our tests (certainly not AT tests) so we might as well not run them. Let me know when you want me to merge manually or I'll just do it after @ndkeen 's review.

brhillman:
Yes, I did a no-run test and it fixed the RUN_STARTDATE.

ndkeen (approved):

PR #705, opened 2020-10-27
Clean up max/min code

tcclevenger:
Found the `ekat::max/min(Pack, Pack)` function during my last PR and realized I could clean up a lot of my past contributions. Also includes changes to `ekat::impl::max(a,b)` instead of `(a>b ? a : b)` type expressions, and some shortening of code.

### DIFF
--- a/components/scream/src/physics/shoc/shoc_compute_conv_time_shoc_length_impl.hpp
+++ b/components/scream/src/physics/shoc/shoc_compute_conv_time_shoc_length_impl.hpp
@@ -18,8 +18,7 @@ void Functions<S,D>::compute_conv_time_shoc_length(
   Scalar&       conv_vel,
   Scalar&       tscale)
 {
-  const Scalar tmp_val = (conv_vel > 0 ? conv_vel : 0);
-  conv_vel = std::pow(tmp_val, C::THIRD);
+  conv_vel = std::pow(ekat::impl::max(conv_vel, sp(0)), C::THIRD);
 
   if (conv_vel > 0) {
       tscale = pblh/conv_vel;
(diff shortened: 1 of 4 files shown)


jeff-cohere:
Did you mean to assign Jim and me as reviewers, Conrad?

tcclevenger:
> Did you mean to assign Jim and me as reviewers, Conrad?

I knew I would make that mistake one day..

jeff-cohere:
> > Did you mean to assign Jim and me as reviewers, Conrad?
> 
> I knew I would make that mistake one day..

Might as well get it over with!

jeff-cohere (approved):

components/scream/src/physics/shoc/shoc_compute_conv_time_shoc_length_impl.hpp
```diff
@@ -18,8 +18,7 @@ void Functions<S,D>::compute_conv_time_shoc_length(
   Scalar&       conv_vel,
   Scalar&       tscale)
 {
-  const Scalar tmp_val = (conv_vel > 0 ? conv_vel : 0);
-  conv_vel = std::pow(tmp_val, C::THIRD);
+  conv_vel = std::pow(ekat::impl::max(conv_vel, sp(0)), C::THIRD);
```
jgfouca, line 21: Don't need `sp` here.

jgfouca (changes requested):
Can remove sp, otherwise good

components/scream/src/physics/shoc/shoc_compute_conv_time_shoc_length_impl.hpp
tcclevenger: When I have max(conv_vel, 0), I get a 
```
error: no matching function for call to ‘max(scream::shoc::Functions<float, Kokkos::Device<Kokkos::Serial, Kokkos::HostSpace> >::Scalar&, int)
```
I think that `std::max` doesn't allow for mixed types.

jgfouca (approved):

components/scream/src/physics/shoc/shoc_compute_conv_time_shoc_length_impl.hpp
jgfouca: Ah, my bad.

PR #710, opened 2020-10-31
Populate RRTMGP optical property outputs

brhillman:
Populate existing optical property outputs that were not actually being set. These include:
```
EMIS
TOT_CLD_VISTAU
TOT_ICLD_VISTAU
LIQ_ICLD_VISTAU
ICE_ICLD_VISTAU
``` 
[B4B]

### DIFF
--- a/components/eam/src/physics/rrtmgp/radiation.F90
+++ b/components/eam/src/physics/rrtmgp/radiation.F90
@@ -1182,6 +1182,7 @@ contains
 
       ! Zero-array for cloud properties if not diagnosed by microphysics
       real(r8), target, dimension(pcols,pver) :: zeros
+      real(r8), dimension(pcols,pver) :: c_cldf  ! Combined cloud/snow fraction
 
       !----------------------------------------------------------------------
 
@@ -1213,6 +1214,13 @@ contains
          icswp => zeros
          des => zeros
       end if
+
+      ! Compute combined cloud and snow fraction
+      do icol = 1,size(cld,1)
+         do ilay = 1,size(cld,2)
+            c_cldf(icol,ilay) = max(cld(icol,ilay), cldfsnow(icol,ilay))
+         end do
+      end do
 
       ! Initialize clearsky-heating rates to make sure we do not get garbage
       ! for columns beyond ncol or nday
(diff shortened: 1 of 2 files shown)


components/eam/src/physics/rrtmgp/radiation.F90
```diff
@@ -809,7 +809,7 @@ subroutine radiation_init(state)
          end if
       end do
 
-      call addfld('EMIS', (/ 'lev' /), 'A', '1', 'Cloud longwave emissivity')
+      call addfld('EMIS', (/ 'lev' /), 'A', '1', 'Cloud 10.5 micron emissivity')
```
PeterCaldwell, line 812: maybe "10.5 micron (longwave) emissivity" for folks that don't think about wavelengths all the time?

components/eam/src/physics/rrtmgp/radiation.F90
```diff
@@ -1819,17 +1857,24 @@ subroutine output_cloud_optics_lw(state, tau)
...
       ! Check values
       call assert_valid(tau(1:state%ncol,1:pver,1:nlwbands), 'cld_tau_lw')
 
       ! Output
-      call outfld('CLOUD_TAU_LW', &
-                  tau(1:state%ncol,1:pver,1:nlwbands), &
-                  state%ncol, state%lchnk)
+      call outfld('CLOUD_TAU_LW', tau(1:ncol,1:pver,1:nlwbands), ncol, state%lchnk)
+      lw_band_index = get_band_index_lw(10.5_r8, 'micron')
+      emis(1:ncol,:) = 1._r8 - exp(-tau(1:ncol,:,lw_band_index))
```
PeterCaldwell, line 1876: This is probably a really dumb question, but as a sanity check: should this line be 1-exp(-tau/**cos(zenith)**)?

brhillman: No I don't think so. As this is longwave emissivity, the solar zenith angle doesn't factor in here at all.

components/eam/src/physics/rrtmgp/radiation.F90
brhillman: Or "longwave (10.5 micron) emissivity"? I can make that change, and a corresponding change for the SW properties as well. Just makes the names longer :-)

PeterCaldwell: I like "longwave (10.5 micron) emissivity". Might be longer, but it is just the "long name" field and it is oh so much more descriptive. And good idea to do the same thing for SW.

PeterCaldwell (approved):
Thanks Ben!

brhillman:
@PeterCaldwell this is good to go, passes `e3sm_scream` regression tests on mappy.

PR #711, opened 2020-10-31
Rename U10 Outfld

PeterCaldwell:
The output field U10 sounds like it should be the 10m U wind... but it is instead the 10m wind speed. This is a dangerous source of confusion, so we're renaming it WINDSPD_10M. It would be nice to make this change in E3SM and e3sm_diags, but we're just doing it in SCREAM repo for now since we want this change in upcoming DYAMOND2 runs.

Skipping AT since this change only affects F90 code not tested by it.


PeterCaldwell:
Confirmed that this change worked in ne4 case on cori-knl.

PR #714, opened 2020-11-02
Hack Koby's conservation checks into SCREAMv0

PeterCaldwell:
This PR implements 3 checks from Koby to make sure microphys doesn't over-deplete stuff. Because I don't have time to implement the C++ version of these changes, I just added a logical flag "Koby_fixes" in micro_p3.F90 which will be false by default to ensure BFB tests still pass but which we will set to true to use these changes. Obviously, we should port these changes to C++ soon so we can get rid of this weird hack in our code.

My expectation is that these overdepletion conditions don't happen often, so shouldn't affect model climate much, but will hopefully improve stability.

Additions by @jgfouca :
1) upgrades to gen-boiler to have better support for single-pack functions bfb unit test generation
2) upgrade gen-boiler to have support for auto-generation of _f glue impl for single pack functions
3) port ice_supersat_conservation, nc_conservation, nr_conservation, ni_conservation, and water_vapor_conservation to cxx
4) the _f impls for cloud_water_conservation, rain_water_conservation, and ice_water_conservation from long-ago were not done correctly (they were hardcoded to HostDevice). These have been fixed.
5) Reorganize p3_functions_f90.hpp to be compatible with gen-boiler; gen-boiler should now work for p3 functions
6) Remove the Koby bool, just always do it


components/eam/src/physics/cam/micro_p3.F90
```diff
@@ -830,6 +834,15 @@ SUBROUTINE p3_main_part2(kts, kte, kbot, ktop, kdir, do_predict_nc, do_prescribe
       call ice_water_conservation(qi(k), qidep, qinuc, qiberg, qrcol, qccol, qr2qi_immers_freeze_tend, qc2qi_hetero_freeze_tend, &
            dt, qi2qv_sublim_tend, qi2qr_melt_tend)
 
+      if (Koby_fixes) then
+         call nc_conservation(nc(k), nc_selfcollect_tend, dt, nc_collect_tend, nc2ni_immers_freeze_tend, &
+              nc_accret_tend, nc2nr_autoconv_tend)
+         call nr_conservation(nr(k),ni2nr_melt_tend,nr_ice_shed_tend,ncshdc,nc2nr_autoconv_tend,dt,nr_collect_tend,&
```
crterai, line 840: I think `nc2nr_autoconv_tend` should be replaced with ncautr, since `nc2nr_autoconv_tend` is the tendency of `nc`, while `ncautr` is the tendency of `nr`, at least according to notes in cloud_water_autoconversion. Another fix would be to use `0.5*nc2nr_autoconv_tend` since Delta(nc) = 2*Delta(nr).

PeterCaldwell: You're totally right that nc2nr_autoconv_tend is the number of cloud drops turned to rain by autoconversion and is NOT equal to ncautr=the number of rain drops caused by that conversion. ncautr partitions all the new mass into 25 micron radius drops while nc2nr_autoconv_tend removes the fraction of nc proportional to the autoconverted mass divided by the total mass. I'm dubious that autoconversion shouldn't conserve drop number though.

PeterCaldwell: Also, we should rename ncautr to nr_autoconv_tend at some point (but not in this PR)

PeterCaldwell: @crterai : I don't understand what you mean by "Another fix would be to use 0.5*nc2nr_autoconv_tend since Delta(nc) = 2*Delta(nr)." Are you saying that if I didn't want to use ncautr I could just divide the nc2rn_autoconv_tend value I'm using now by 2? I don't understand where you get this number...

components/eam/src/physics/cam/micro_p3.F90
```diff
@@ -2855,6 +2868,125 @@ subroutine prevent_ice_overdepletion(pres,t_atm,qv,latent_heat_sublim,inv_dt,
...
+  real(rtype), intent(in) :: cld_frac_i,qv,qv_sat_i,latent_heat_sublim,T_atm,dt
+  real(rtype), intent(inout) :: qidep,qinuc
+  
+  real(rtype) :: qv_sink, qv_avail, fract 
+
+  qv_sink = qidep + qinuc ! in [kg/kg] cell-avg values
+
+  if (qv_sink > qsmall .and. cld_frac_i > 1.0e-20_rtype) then
+     ! --- Available water vapor for deposition/nucleation
+     qv_avail = (qv - qv_sat_i) / &
```
PeterCaldwell, line 2905: @JS-WRF-SBM - I think qv_avail should be end-of-step qv, so the line above should be "qv_avail = (qv +(qi2qv_sublim_tend+qr2qv_evap_tend)*dt -qv_sat_i)/(stuff that was there before, or maybe abi). Do you agree? Further, I think that ice_supersat_conservation subroutine is always more strict than water_vapor_conservation so we could probably completely delete the latter function. Do you agree?

components/eam/src/physics/cam/micro_p3.F90
```diff
@@ -2855,6 +2868,125 @@ subroutine prevent_ice_overdepletion(pres,t_atm,qv,latent_heat_sublim,inv_dt,
...
+  !Make sure nc doesn't go below zero
+  
+  implicit none
+  
+  real(rtype), intent(in) :: nc,nc_selfcollect_tend,dt
+  real(rtype), intent(inout) :: nc_collect_tend,nc2ni_immers_freeze_tend,&
+                                nc_accret_tend,nc2nr_autoconv_tend
+  real(rtype) :: sink_nc, source_nc, ratio
+  
+  sink_nc = (nc_collect_tend + nc2ni_immers_freeze_tend + nc_accret_tend + nc2nr_autoconv_tend)*dt
```
PeterCaldwell, line 2932: nc_collect_tend + nc2ni_immers_freeze_tend are missing from update_prognostic_liquid. Either they should be added or these quantities should be deleted here.

components/eam/src/physics/cam/micro_p3.F90
```diff
@@ -2855,6 +2868,125 @@ subroutine prevent_ice_overdepletion(pres,t_atm,qv,latent_heat_sublim,inv_dt,
...
+  !Make sure nr doesn't go below zero
+
+  implicit none
+
+  real(rtype), intent(in) :: nr,ni2nr_melt_tend,nr_ice_shed_tend,ncshdc,nc2nr_autoconv_tend,dt
+  real(rtype), intent(inout) :: nr_collect_tend,nr2ni_immers_freeze_tend,nr_selfcollect_tend,nr_evap_tend
+  real(rtype) :: sink_nr, source_nr, ratio
+
+  sink_nr = (nr_collect_tend + nr2ni_immers_freeze_tend + nr_selfcollect_tend + nr_evap_tend)*dt 
+  source_nr = nr + (ni2nr_melt_tend + nr_ice_shed_tend + ncshdc + nc2nr_autoconv_tend)*dt
```
PeterCaldwell, line 2956: nc2nr_autoconv_tend should be ncautr here as well. 

nr_collect_tend and nr2ni_immers_freeze_tend, ni2nr_melt_tend, nr_ice_shed_tend, ncshdc all don't appear in update_prognostic_liquid so either should be added there, removed here, or I should figure out what's going on.

components/eam/src/physics/cam/micro_p3.F90
crterai: @PeterCaldwell As I understand it, autoconversion is the collision of two cloud droplets to form one rain droplet, so that's where the factor of 2 comes in. For every two cloud droplets lost in nc, one nr is gained during autoconversion. But for accuracy, I think it should be replaced with ncautr.

components/eam/src/physics/cam/micro_p3.F90
crterai: I agree, nc2nr_autoconv_tend should be ncautr here. 

The terms you listed are all ice processes, I think, and are oddly listed under update_prognostic_ice (perhaps the phase connotes liquid only processes vs any process including ice). So that's why they are missing there. I think it's okay as it is.

components/eam/src/physics/cam/micro_p3.F90
crterai: The nc tendency from ice phase processes are listed under update_prognostic_ice. I think it's okay as it is (see comment below for nr_conservation).

components/eam/src/physics/cam/micro_p3.F90
PeterCaldwell: Ok, you're right that ncautr is the appropriate variable.

components/eam/src/physics/cam/micro_p3.F90
PeterCaldwell: You're right, Chris. Thanks!

components/eam/src/physics/cam/micro_p3.F90
PeterCaldwell: Ok, switched to ncautr here. You're right about the quantities I'm worried about just showing up in the update_prognostic_ice instead of update_prognostic_liquid. I think it is kind of ambiguous whether a process taking liquid and converting it to ice should be in update_prognostic_liquid or ice... I think it is fine as is.

components/eam/src/physics/cam/micro_p3.F90
JS-WRF-SBM: @PeterCaldwell 

I agree that for the check of the available ice supersaturation we should use the 'full potential' of the water vapor, which seems to be at the beginning of the MP time step, rather then at the end of it -- isn't it?

I'm really concerned  from all these non-bounded evaporative processes (ice sublim / rain evap), particulary for LR -- this is why I tend to keep the check.

PeterCaldwell: @JS-WRF-SBM - I think I rewrote my comment while you were responding to it. Maybe the new version answers your questions, but to be clear:

1. In terms of the 'full potential' of the water vapor, I think an easier and equivalent way to think is "at the end of the timestep (i.e. after this function and the state update), the smallest ice supersaturation should be is 0%". Your current version rescales such that "in the absence of qv sources, supersaturation will be zero at the end of the step". But there are qv sources so your rescaling will result in positive supersaturation at the end of the timestep... it is too aggressive.

2. I'm arguing that if you ran your code with and without qv_conservation, you'd get BFB answers... qv_conservation just wastes flops and makes the code more complicated. This is something we should test empirically by doing runs with and without this fixer. I'm not sure when I'll find time to do that though.

JS-WRF-SBM: @PeterCaldwell 

-  The deposition and sublimation (sublimation being a source of `qv`) occurs alternately per grid point. So I wouldn't worry about that. If sublimation is enabled, both `qinuc` and `qidep` needs to be zero. Did we check that this is the case for `qinuc` which comes from outside the P3? I personally haven't checked that.
-  The rain evaporation requires `qv - qv_sat_l < 0.0` 
-   Not sure I'm fully following your argument above - but if you argue that limiting only the qv > 0 at the `update_liquid/ice` routines is enough? Please note that when you involve `qv_sat_i`  / `qv_sat_l` you have an intrinsic dependency on temperature from processes before P3. This is a more general condition compare to `qv > 0`

PeterCaldwell: @JS-WRF-SBM 
- good point that if you are sublimating or evaporating you shouldn't be doing vapor deposition or ice nucleation. There is a check in p3 that only gives qinuc positive values if ice supersaturation is > 5%. I concede that this means that your original implementation where qinuc and qidep are limited based on qv rather than qv+(sublim+evap)*dt is sufficient. I think this means you could have also omitted sublim + evap from your qv_limit function. I'm not going to make changes to what I've done because it isn't wrong (just adds something we know will be zero) and the logic for why it is ok to leave off these source terms is a bit convoluted. But mostly because I don't want to change the PR. 
- I'm confused about your last bullet. My argument here is basically that if you insist that qv is at least saturated with respect to ice, then you're already ensuring that qv>0 because qv_sat_i is always > 0. Thus there's no point in your qv_limiter function. It is true that abi has a mild dependence on T. I'm not sure where you're getting the qv_sat_l dependence - from rain evaporation? I agree that would have an impact, but I think it is in the noise compared to other assumptions we make.

jgfouca:
@PeterCaldwell , since this is your branch, I'm unable to ask for a review from you through github. Could you take a look at this when you get a chance?

jeff-cohere:
Hey @jgfouca . Anything in particular you'd like me to look for in this PR? Or just look it over and call out things that look strange?

jgfouca:
@jeff-cohere , no, nothing in particular.

components/eam/src/physics/cam/micro_p3.F90
```diff
@@ -2855,6 +2862,125 @@ subroutine prevent_ice_overdepletion(pres,t_atm,qv,latent_heat_sublim,inv_dt,
 
 end subroutine prevent_ice_overdepletion
 
+subroutine water_vapor_conservation(qv,qidep,qinuc,qi2qv_sublim_tend,qr2qv_evap_tend,dt)
```
jeff-cohere, line 2865: Can we get a one-line comment about exactly what this subroutine does and when to call it?

components/eam/src/physics/cam/micro_p3.F90
```diff
@@ -2855,6 +2862,125 @@ subroutine prevent_ice_overdepletion(pres,t_atm,qv,latent_heat_sublim,inv_dt,
...
+        fract = qv_avail / qv_sink
+        qinuc = qinuc * fract
+        qidep = qidep * fract
+     endif
+  endif
+
+  return
+end subroutine ice_supersat_conservation
+
+subroutine nc_conservation(nc, nc_selfcollect_tend, dt, nc_collect_tend, nc2ni_immers_freeze_tend, &
```
jeff-cohere, line 2915: For newcomers it would be exceedingly helpful to have a comment or two that describes each of these subroutines and their purpose in succinct language.

jeff-cohere (approved):
This looks great, @jgfouca . It seems like `gen_boiler` is already pulling its weight (with help from you, of course!). Thanks for the explanation of what you added to the original PR. It helped.

jgfouca:
Thanks, @jeff-cohere ! @PeterCaldwell , could you take a look at this and give me a brief description of these new methods so I can add a bit of documentation?

components/eam/src/physics/cam/micro_p3.F90
PeterCaldwell: Done.

components/eam/src/physics/cam/micro_p3.F90
PeterCaldwell: done.

PR #716, opened 2020-11-03
Some mods to the grids classes, notably SEGrid

bartgol:
This PR gets rid of the SE_CellBased and SE_NodeBased grid types, replacing them with SE, and the SimpleGrid becomes PointGrid. In general, it simplifies a bit the set of grids we deal with in scream (for now). The fact that a SE grid could be either a "dynamics"-like grid or a "physics"-like grid was probably confused. The SE_NodeBased grid was not really much more than a bunch of columns, with little to no importance on the fact that the columns were on the GLL points of a SE grid.


components/scream/src/share/field/field_monotonicity_check.hpp
```diff
@@ -39,7 +39,7 @@ class FieldMonotonicityCheck: public FieldPropertyCheck<RealType> {
     return false;
   }
 
-  void repair(Field<RealType>& field) const override {
+  void repair(Field<RealType>& /* field */) const override {
```
jeff-cohere, line 42: Just curious: did a compiler generate a warning for this unused parameter? I find these warnings to be distracting and usually disable them myself. Omitting unused parameters is Yet Another Thing that people must remember when using C++. :-)

components/scream/src/share/grid/point_grid.cpp
```diff
@@ -0,0 +1,60 @@
...
+ : m_grid_name   (grid_name)
+ , m_num_my_cols (dofs_gids.size())
+ , m_num_vl      (num_vertical_lev)
+ , m_dofs_gids   (dofs_gids)
+{
+  // The lid->idx map is the identity map.
+  m_lid_to_idx = decltype(m_lid_to_idx)("lid to idx",m_num_my_cols,1);
+
+  auto h_lid_to_idx = Kokkos::create_mirror_view(m_lid_to_idx);
+  std::iota(h_lid_to_idx.data(),h_lid_to_idx.data()+m_num_my_cols,0);
```
jeff-cohere, line 20: I learn something new about C++ every time I look at your code, Lucca. :-)

components/scream/src/share/grid/point_grid.hpp
```diff
@@ -0,0 +1,68 @@
...
+#include "share/grid/abstract_grid.hpp"
+#include "share/scream_types.hpp"
+
+#include "ekat/mpi/ekat_comm.hpp"
+
+namespace scream
+{
+
+/*
+ * A grid consisting of a bunch of points
```
jeff-cohere, line 13: Thank you for all the great code comments in this PR!

jeff-cohere (approved):
A very nice simplification. Approved, pending passing tests and all that.

components/scream/src/share/field/field_monotonicity_check.hpp
bartgol: In my vim editor, clang compiles the code in the background, and highlights warnings/errors. I think it uses pedantic flags. It really helps reduce the edit->compile cycles. This was one of them.

I agree that these warnings are not that important. Alas, I like to see my build output consisting just of 
```
[ 67%] Building CXX object ...
[ 68%] Built target ...
```
without any warning to pollute the output.

PR #717, opened 2020-11-03
Port dp_inverse and compute_tmpi

tcclevenger:
2 simple function, `compute_tmpi` and `dp_inverse`.

### DIFF
--- a/components/scream/src/physics/shoc/shoc_compute_tmpi_impl.hpp
+++ b/components/scream/src/physics/shoc/shoc_compute_tmpi_impl.hpp
@@ -1,0 +1,34 @@
+#ifndef SHOC_COMPUTE_TMPI_IMPL_HPP
+#define SHOC_COMPUTE_TMPI_IMPL_HPP
+
+#include "shoc_functions.hpp" // for ETI only but harmless for GPU
+
+namespace scream {
+namespace shoc {
+
+template<typename S, typename D>
+KOKKOS_FUNCTION
+void Functions<S,D>
+::compute_tmpi(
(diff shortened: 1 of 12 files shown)


jeff-cohere (approved):
Looks good to me!

components/scream/src/physics/shoc/shoc_compute_tmpi_impl.hpp
```diff
@@ -0,0 +1,34 @@
...
+  const uview_1d<Spack>&       tmpi)
+{
+  const auto ggr = C::gravit;
+
+  tmpi(0)[0] = 0;
+
+  const Int nlev_pack = ekat::npack<Spack>(nlevi);
+  Kokkos::parallel_for(Kokkos::TeamThreadRange(team, nlev_pack), [&] (const Int& k) {
+    auto range_pack = ekat::range<IntSmallPack>(k*Spack::n);
+    tmpi(k).set(range_pack > 0, dtime*(ggr*rho_zi(k))/dz_zi(k));
```
jgfouca, line 27: @tcclevenger , sorry, I'm just now looking at your PR. I'm confused by line 27: is there any chance for range_pack to fall below zero? If there is, I'm not seeing it.

tcclevenger: I think here we just want to exclude the first entry since the original for-loop in Fortran is k=2..nlev.

jgfouca: Ah, ok

PR #718, opened 2020-11-04
Add more structured support for f90 compilers in testing scripts

bartgol:
This PR is mainly focused on better structuring our compilers support in our testing scripts, but also does a couple more things. Here are the main points:

- cxx compiler is not longer captured without flag, and now needs `--cxx-compiler` (`-cxx` is a short version).
- added similar flag for f90 compiler: `--f90-compiler` (`-f90` is a short version).
- if compilers are not specified, test-all queries machine_specs for default compilers (which means `-m mach_name` must be passed).
- scream testing no longer allows to specify an existing kokkos installation, and instead always builds the internal one (via ekat).
- the `-k` flag in test-all has been rebranded as a short for the existing `--keep-tree`.
- gather-all still allows to specify an existing kokkos installation if `--scream-docs` is selected.
- using system modules for python on blake.

### DIFF
--- a/components/scream/scripts/gather_all_data.py
+++ b/components/scream/scripts/gather_all_data.py
@@ -50,19 +50,18 @@ class GatherAllData(object):
 
         repo = scream_docs_repo if self._scream_docs else scream_repo
 
-        # Compute kokkos location
-        if self._kokkos:
-            expect (self._scream_docs, "Error! Scream runs do not allow to specify a user-provided kokkos installation.")
-            kokkos_loc = pathlib.Path(self._kokkos.replace("$machine", machine).replace("$compiler", compiler))
-
         local_cmd = self._run
         # Do magic replacements here
         local_cmd = local_cmd.\
             replace("$compiler", compiler).\
-            replace("$kokkos", str(kokkos_loc)).\
             replace("$machine", machine).\
             replace("$scream_docs", str(scream_docs_repo)).\
             replace("$scream", str(scream_repo))
+        if self._kokkos:
+            # Compute kokkos location and do magic replacement
+            expect (self._scream_docs, "Error! Scream runs do not allow to specify a user-provided kokkos installation.")
+            kokkos_loc = pathlib.Path(self._kokkos.replace("$machine", machine).replace("$compiler", compiler))
+            local_cmd.replace("$kokkos", str(kokkos_loc)).\
 
         # Scream-docs tests may depend on scream's Kokkos and scripts, so update scream repo too if we're doing
         # a scream-docs test.
--- a/components/scream/scripts/gather_all_data.py
+++ b/components/scream/scripts/gather_all_data.py
@@ -52,9 +52,8 @@ class GatherAllData(object):
 
         # Compute kokkos location
         if self._kokkos:
+            expect (self._scream_docs, "Error! Scream runs do not allow to specify a user-provided kokkos installation.")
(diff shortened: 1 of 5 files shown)


components/scream/scripts/gather_all_data.py
```diff
@@ -50,20 +51,19 @@ def formulate_command(self, machine):
...
-            kokkos_loc = pathlib.Path(self._kokkos.replace("$machine", machine).replace("$compiler", compiler))
-        else:
-            kokkos_loc = pathlib.Path(scream_repo.parent.parent, "externals", "kokkos")
-
         local_cmd = self._run
         # Do magic replacements here
         local_cmd = local_cmd.\
-            replace("$compiler", compiler).\
-            replace("$kokkos", str(kokkos_loc)).\
+            replace("-cxx $cxx_compiler", cxx_compiler).\
```
jgfouca, line 57: These magic string replacements don't look right. I think you want:
```
replace("$cxx_compiler", cxx_compiler).\
replace("$f90_compiler", f90_compiler).\
```

For some context, in the help examples above, you have:
```
gather-all-data './scripts/test-all-scream --cxx-compiler $cxx_compiler --f90-compiler $f90_compiler -m $machine'
```
Which utilizes two magic strings `$cxx_compiler` and `$f90_compiler`.

I would confirm that the above use case still works with this PR.

components/scream/scripts/gather_all_data.py
```diff
@@ -50,20 +51,19 @@ def formulate_command(self, machine):
...
+            replace("-f90 $f90_compiler", f90_compiler).\
             replace("$machine", machine).\
             replace("$scream_docs", str(scream_docs_repo)).\
             replace("$scream", str(scream_repo))
+        if self._kokkos:
```
jgfouca, line 62: It seems like it might be easier to do the kokkos stuff before the other replacements so you don't have to repeat them.

jgfouca (changes requested):
Couple inline items.

components/scream/scripts/gather_all_data.py
bartgol: Ah, good catch.

components/scream/scripts/gather_all_data.py
```diff
@@ -50,17 +51,20 @@ def formulate_command(self, machine):
...
         repo = scream_docs_repo if self._scream_docs else scream_repo
 
-        # Compute kokkos location
+        # Compute kokkos location. scream itself won't allow this, and test-all-scream will error out,
+        # but one can use gather-all-data for plenty of other apps, so we need to support this here.
+        local_cmd = self._run
         if self._kokkos:
-            kokkos_loc = pathlib.Path(self._kokkos.replace("$machine", machine).replace("$compiler", compiler))
+            # Note: kokkos_loc can use other magic strings, like $cxx_compiler, or $machine
+            kokkos_loc = self._kokkos
```
jgfouca, line 59: If we don't want to support a custom kokkos for anything but scream-docs, you can put that expect check here.

bartgol: Well, I was thinking about it, but what if tomorrow someone decides to use gather-all for something else? I think gather-all might be able to accommodate that (with minor mods). test-all already errors out if you try to pass `-k blah` now, so you _would_ get an error.

jgfouca (approved):

Issue #699, opened 2020-10-22
[AD] Add two dictionaries of CF compliant variable longnames to be used to enforce CF compliance in SCREAM


AaronDonahue:
Variable longnames as stored in the Field Manager should be CF compliant.  This will allow the field manager to ensure that no two processes register the same variable under different names.  It will also establish a naming convention for variables that can be consistent across processes.  See https://cfconventions.org/standard-names.html

A dictionary of CF compliant long names will be needed in SCREAM.  The field manager can then query this dictionary for each new variable and throw an error if a variable is not listed.

There will be variables in the field manager that are not in the CF dictionary.  A second dictionary of SCREAM-specific variable names will be needed that can be expanded to include these variables.  By enforcing compliance with a second dictionary we can hedge against developers accidentally adding an unknown variable with no way of checking that it is in fact in the CF dictionary.

This is step 1 of getting CF compliance in SCREAM.

Addresses #527  and #528

jeff-cohere:
Hey @AaronDonahue ,

I know we discussed creating a YAML file in which to store the CF standard field names, and we have one of those now (generated by a script in the `jeff-cohere/cf-xml-to-yaml` branch `scripts/` directory). Is that the "deliverable" for this item?

The YAML file is huge, since there are a ton of possible CF field names (and aliases). I guess that doesn't matter as long as we just use it (and presumably a second one of our own) as input to generate validation code. I don't know that we want SCREAM to require these files as data.

bartgol:
@jeff-cohere yeah, I think it's fine. We are only going to use it to validate names (i.e., reject names that are not in that list).

PeterCaldwell:
I think this is half-done now via #719 , right? Should we narrow this issue down to just the "create a table of CF-like names for variables that aren't already in the CF table" (via modifying this PR or closing and creating a new one)?

jeff-cohere:
Yeah, and actually the (empty) file already exists for the SCREAM CF-like names (in `data/cf-scream-name-table.yaml`, alongside `data/cf-standard-name-table.yaml`). I think we can close this issue and just add names to that table as part of our ongoing work.

PR #719, opened 2020-11-04
Added a Python script that converts a CF standards XML file to a YAML file.

jeff-cohere:
This Python script can be used as part of a tool chain for downloading the CF conventions and generating a YAML file that we can use to accept/reject long field names and aliases for quantities that have counterparts in CF.

The resulting YAML file is the first of two "dictionaries" that we need. The second is one of our own making, that contains field names and aliases for quantities that we track but that do not appear in the CF conventions. It also establishes a format for these dictionaries.

I encourage anyone to try it out! Just download the CF Conventions XML file (using the URL included in the script's front matter ) and run the script on it.

```
wget http://cfconventions.org/Data/cf-standard-names/76/src/cf-standard-name-table.xml
scripts/cf-xml-to-yaml cf-standard-name-table.xml
```

I've also added a `build_cf_dictionary` target that does this for you. Just type `make build_cf_dictionary` in your build directory, and your scream/data source directory will be populated with the generated YAML file.

Addresses #699


AaronDonahue (approved):
Looks good to me.  And it would be great to have an easy way to generate the YAML file if a new CF XML comes online.

components/scream/scripts/cf-xml-to-yaml
```diff
@@ -0,0 +1,80 @@
+#! /usr/bin/env python
+
+"""
+Given an XML file containing the CF conventions for standardized field names
+(https://cfconventions.org/standard-names.html), this tool generates a YAML
+file with the same information.
+
+Usage: cf-xml-toyaml <cf-xml-file>
```
AaronDonahue, line 8: nitpicky, but should the text read `cf-xml-to-yaml`?  Notice the missing hyphen.

jeff-cohere: Whoops! Nice catch.

PeterCaldwell (approved):
looks good!

jeff-cohere:
Updated with a new build target, so I'm re-requesting approval. I think that ties off this PR.

PeterCaldwell:
The new version still looks good. My only question is why you put the new target in scream/components/scream/src/share/. I'm not sure where would be better, but I wouldn't think to look here to build a tool...

jeff-cohere:
> The new version still looks good. My only question is why you put the new target in scream/components/scream/src/share/. I'm not sure where would be better, but I wouldn't think to look here to build a tool...

The real "tool" is the script in the scripts/ directory. The target is just  CMake logic and can live in any `CMakeLists.txt` file. I guess I thought it should belong with the atmosphere driver code, since that's where we care about validating field names.

PeterCaldwell (approved):
Ok, so long as you thought about target location, I'm cool with it.

PR #720, opened 2020-11-04
porting shoc_diag_second_moments_impl function

xyuan:
porting shoc_diag_second_moments_impl function

[BFB]

### DIFF
--- a/components/scream/src/physics/shoc/shoc_diag_second_moments_impl.hpp
+++ b/components/scream/src/physics/shoc/shoc_diag_second_moments_impl.hpp
@@ -13,10 +13,68 @@ namespace shoc {
 
 template<typename S, typename D>
 KOKKOS_FUNCTION
-void Functions<S,D>::diag_second_moments(const Int& shcol, const Int& nlev, const Int& nlevi, const uview_1d<const Spack>& thetal, const uview_1d<const Spack>& qw, const uview_1d<const Spack>& u_wind, const uview_1d<const Spack>& v_wind, const uview_1d<const Spack>& tke, const uview_1d<const Spack>& isotropy, const uview_1d<const Spack>& tkh, const uview_1d<const Spack>& tk, const uview_1d<const Spack>& dz_zi, const uview_1d<const Spack>& zt_grid, const uview_1d<const Spack>& zi_grid, const uview_1d<const Spack>& shoc_mix, const uview_1d<Spack>& thl_sec, const uview_1d<Spack>& qw_sec, const uview_1d<Spack>& wthl_sec, const uview_1d<Spack>& wqw_sec, const uview_1d<Spack>& qwthl_sec, const uview_1d<Spack>& uw_sec, const uview_1d<Spack>& vw_sec, const uview_1d<Spack>& wtke_sec, const uview_1d<Spack>& w_sec)
+void Functions<S,D>::diag_second_moments(const MemberType& team, const Int& nlev, const Int& nlevi, 
+                     const uview_1d<const Spack>& thetal, const uview_1d<const Spack>& qw, const uview_1d<const Spack>& u_wind, 
+                     const uview_1d<const Spack>& v_wind, const uview_1d<const Spack>& tke, const uview_1d<const Spack>& isotropy, 
+                     const uview_1d<const Spack>& tkh, const uview_1d<const Spack>& tk, const uview_1d<const Spack>& dz_zi, 
+                     const uview_1d<const Spack>& zt_grid, const uview_1d<const Spack>& zi_grid, const uview_1d<const Spack>& shoc_mix,
+                     const uview_1d<Spack>& isotropy_zi, const uview_1d<Spack>& tkh_zi, const uview_1d<Spack>& tk_zi, 
+                     const uview_1d<Spack>& thl_sec, const uview_1d<Spack>& qw_sec, const uview_1d<Spack>& wthl_sec, const uview_1d<Spack>& wqw_sec, 
+                     const uview_1d<Spack>& qwthl_sec, const uview_1d<Spack>& uw_sec, const uview_1d<Spack>& vw_sec, const uview_1d<Spack>& wtke_sec, 
+                     const uview_1d<Spack>& w_sec)
 {
-  // TODO
-  // Note, argument types may need tweaking. Generator is not always able to tell what needs to be packed
+  // Purpose of this subroutine is to diagnose the second
+  //  order moments needed for the SHOC parameterization.
+  //  Namely these are variances of thetal, qw, and vertical
+  //  velocity.  In addition the vertical fluxes of thetal, qw,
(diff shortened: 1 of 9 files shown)


jeff-cohere (approved):

components/scream/src/physics/shoc/shoc_diag_second_moments_impl.hpp
```diff
@@ -13,10 +13,68 @@ namespace shoc {
...
+  // temp moisture covariance
+  const auto qwthl2tune = sp(1);
+
+  // vertical velocity variance
+  const auto w2tune = sp(1);
+
+  // Interpolate some variables from the midpoint grid to the interface grid
+  linear_interp(team, zt_grid, zi_grid, isotropy, isotropy_zi, nlev, nlevi, sp(0));
+  linear_interp(team, zt_grid, zi_grid, tkh,      tkh_zi,      nlev, nlevi, sp(0));
+  linear_interp(team, zt_grid, zi_grid, tk,       tk_zi,       nlev, nlevi, sp(0));
```
tcclevenger, line 47: `sp()` is unnecessary above since `linear_interp` will do the correct conversion of integers -> Scalar. It doesn't hurt anything though, just may look nicer. Up to you.

components/scream/src/physics/shoc/shoc_diag_second_shoc_moments_impl.hpp
```diff
@@ -13,10 +13,18 @@ namespace shoc {
 
 template<typename S, typename D>
 KOKKOS_FUNCTION
-void Functions<S,D>::diag_second_shoc_moments(const Int& shcol, const Int& nlev, const Int& nlevi, const uview_1d<const Spack>& thetal, const uview_1d<const Spack>& qw, const uview_1d<const Spack>& u_wind, const uview_1d<const Spack>& v_wind, const uview_1d<const Spack>& tke, const uview_1d<const Spack>& isotropy, const uview_1d<const Spack>& tkh, const uview_1d<const Spack>& tk, const uview_1d<const Spack>& dz_zi, const uview_1d<const Spack>& zt_grid, const uview_1d<const Spack>& zi_grid, const uview_1d<const Spack>& shoc_mix, const uview_1d<const Spack>& wthl_sfc, const uview_1d<const Spack>& wqw_sfc, const uview_1d<const Spack>& uw_sfc, const uview_1d<const Spack>& vw_sfc, const uview_1d<Spack>& thl_sec, const uview_1d<Spack>& qw_sec, const uview_1d<Spack>& wthl_sec, const uview_1d<Spack>& wqw_sec, const uview_1d<Spack>& qwthl_sec, const uview_1d<Spack>& uw_sec, const uview_1d<Spack>& vw_sec, const uview_1d<Spack>& wtke_sec, const uview_1d<Spack>& w_sec)
+void Functions<S,D>::diag_second_shoc_moments(const Int& shcol, const Int& nlev, const Int& nlevi,
```
tcclevenger, line 16: Is this conversion meant to be included in this PR?

components/scream/src/physics/shoc/shoc_diag_second_moments_impl.hpp
```diff
@@ -13,10 +13,68 @@ namespace shoc {
...
+  const auto thl2tune = sp(1);
+
+  // moisture variance
+  const auto qw2tune = sp(1);
+
+  // temp moisture covariance
+  const auto qwthl2tune = sp(1);
+
+  // vertical velocity variance
+  const auto w2tune = sp(1);
```
tcclevenger, line 42: @ambrad @bartgol Constants question: These tune parameters are defined at the top of `shoc.F90`, all set to 1, and used only in this function. Can they be removed in the C++ impl?
Further, they are the only values used for the `calc_shoc_varorcovar`'s `tunefac` parameter. Should this input be removed all together and 1.0 used in its place?

Not sure, it might still useful to have around unused variables for someone to tweak in the future..

tcclevenger (commented):
Just a few comments, otherwise looks good to me!

components/scream/src/physics/shoc/shoc_diag_second_moments_impl.hpp
bartgol: Also, 0 is exactly represented regardless of the numerical type.

components/scream/src/physics/shoc/shoc_diag_second_moments_impl.hpp
bartgol: That depends on whether there is a plan of tweaking them in the future or not. I am assuming someone _might_ want to try other values (otherwise they wouldn't have added the input in the functions below), so I'm inclined to keep them. Besides, I think it's more verbose like this than seeing `f(...,1.0,...)`, which leaves the readed wondering what 1.0 is.

If there is no plan to play with those values and we don't foresee that happening ever, then they should be removed from the fcn signatures altogether.

components/scream/src/physics/shoc/shoc_diag_second_shoc_moments_impl.hpp
xyuan: yes, this is necessary conversion.

components/scream/src/physics/shoc/shoc_diag_second_moments_impl.hpp
xyuan: This is local variable that only used in this function, so I won't recommend it to be shared with others.

components/scream/src/physics/shoc/shoc_diag_second_moments_impl.hpp
xyuan: I would keep them as it is

tcclevenger (approved):

components/scream/src/physics/shoc/shoc_diag_second_moments_impl.hpp
```diff
@@ -13,10 +13,68 @@ namespace shoc {
...
-  // TODO
-  // Note, argument types may need tweaking. Generator is not always able to tell what needs to be packed
+  // Purpose of this subroutine is to diagnose the second
+  //  order moments needed for the SHOC parameterization.
+  //  Namely these are variances of thetal, qw, and vertical
+  //  velocity.  In addition the vertical fluxes of thetal, qw,
+  //  u, v, TKE, and tracers are computed here as well as the
+  //  correlation of qw and thetal.
+
+  const auto thl2tune = sp(1);
```
jgfouca, line 33: We don't need `sp` for any of the integer literals in this function.

jgfouca (changes requested):
Looks good, just need to clean up some sp's

components/scream/src/physics/shoc/shoc_diag_second_moments_impl.hpp
xyuan: ok, I removed sp for integer literals.

jgfouca (approved):

PR #721, opened 2020-11-05
Set prescribed aerosol noise to zero and add bc_dep_to_snow_updates to scream compsets

crterai:
This small PR does two things.
1) Sets the daily random noise in the prescribed aerosol field to always be zero. 
2) Adds bc_dep_to_snow_updates to the scream compsets

### DIFF
--- a/components/eam/src/chemistry/utils/prescribed_aero.F90
+++ b/components/eam/src/chemistry/utils/prescribed_aero.F90
@@ -481,7 +481,10 @@ end subroutine spec_c_to_a
     ! random sampling is off (randn=0) in single column model setting. 
     if (single_column) then
       randn = 0._r8
-    endif
+   endif
+
+   ! turn off random sampling for scream simulations
+   randn = 0._r8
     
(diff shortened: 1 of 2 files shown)


components/eam/src/chemistry/utils/prescribed_aero.F90
```diff
@@ -481,7 +481,10 @@ subroutine rand_sample_prescribed_aero(state, pbuf2d)
     ! random sampling is off (randn=0) in single column model setting. 
     if (single_column) then
       randn = 0._r8
-    endif
+   endif
+
+   ! turn off random sampling for scream simulations
+   randn = 0._r8
```
bogensch, line 487: I originally thought that if we are always going to set randn = 0 via line 487 perhaps we should delete the duplicative single_column conditional above this .

But then I thought that if one day we remove the line 487 statement then we would likely forget about the single_column conditional completely.... So I think leaving as is is the safer option.

bogensch (approved):

PR #722, opened 2020-11-05
[WIP] Add warning before tau out of range crash location

crterai:
Simulations have been crashing with `ERROR: validate: tau values out of range`
Add additional clip_values call before the end_run call to catch any negative tau, which will print out original values and write out a warning.


PeterCaldwell:
@crterai - can we close this crusty old PR?

PR #723, opened 2020-11-05
[WIP] Add tau values in rte-rrtmgp error message

crterai:
To add tau values when the model crashes with tau values out of range. 
In the external module of rrtmgp (file: rte/mo_optical_props.F90), replace the following lines 
`err_message = "validate: tau values out of range"`
with
`err_message = "validate: tau values out of range - ",minval(this%tau)`

Three instances.


crterai:
Source of crash was located as hardware issue. Trying to put in a PR here (in scream) also doesn't show the changes made within the submodule, so closing this PR.

PR #725, opened 2020-11-09
Fix missing shoc_main intents and gen-boiler

jgfouca:


bogensch (approved):
Looks great, thanks for this.

PR #726, opened 2020-11-09
lc machine support 2020 11 - general mpirun flag addition

AaronDonahue:
Make changes needed to get tests working on LC machines.
    
    This commit adjusts the cmake flags to extend the generality in how
    mpirun is used in testing.  The immediate impact is that the two LC
    machine syrah and quartz will be able to run tests using srun, since
    mpirun is not available.
    
    There are also adjustments to fix lassen machine configurations.

[BFB]


bartgol (changes requested):
There's really no need to add a second layer of defaults. Meaning, you don't need `SCREAM_DEFAULT_MPI_NP_FLAG` or `SCREAM_DEFAULT_MPIRUN_EXE`. All you need is simply add
```
set (SCREAM_MPIRUN_EXE "srun" CACHE STRING "")
set (SCREAM_MPI_NP_FLAG "-n" CACHE STRING "")
```
to your machine file. Those will automatically take precedence over the corresponding `DEFAULT_` ones.

To be clear: those two vars are already CACHE vars, meaning that if you specify a value manually (via `-D` on command line, or `set(...)` in a machine file), then the default we set in our CMakeLists.txt files will be ignored, and your choices will be honored.

AaronDonahue:
> There's really no need to add a second layer of defaults. Meaning, you don't need `SCREAM_DEFAULT_MPI_NP_FLAG` or `SCREAM_DEFAULT_MPIRUN_EXE`. All you need is simply add
> 
> ```
> set (SCREAM_MPIRUN_EXE "srun" CACHE STRING "")
> set (SCREAM_MPI_NP_FLAG "-n" CACHE STRING "")
> ```
> 
> to your machine file. Those will automatically take precedence over the corresponding `DEFAULT_` ones.
> 
> To be clear: those two vars are already CACHE vars, meaning that if you specify a value manually (via `-D` on command line, or `set(...)` in a machine file), then the default we set in our CMakeLists.txt files will be ignored, and your choices will be honored.

Ahhh ok.  I'm testing that now then I'll make the change.  I was following the example of the summit.cmake file.  Should that also not have the _DEFAULT syntax?

AaronDonahue:
@bartgol , could you take another look?  I removed the _DEFAULT from the cmake file compiler flags for syrah and quartz.

bartgol:
@AaronDonahue Ah, I see what you mean.

Well, there's really no point in having a default value for the default value. In general, we should have only stuff like this in our CMakeLists.txt files:

```
set(DEFAULT_BLAH "blah")
set(SCREAM_BLAH ${DEFAULT_BLAH} CACHE STRING "")
```
Then, inside machine files we simply do
```
set(SCREAM_BLAH "my_blah" CACHE STRING "")
```

There's no need to introduce a second level of defaults.

AaronDonahue:
@bartgol , ok cool.  I can make that change and update the summit.cmake file as well.  I'll check out the rest of the cmake files too and adjust if needed.  I'll ping you when I'm done.

AaronDonahue:
ok done.  @bartgol , can you check and make sure this is what you were expecting?

jgfouca (approved):

bartgol (approved):

PR #731, opened 2020-11-11
Refactor tridiag solver

tcclevenger:
The solver in SHOC (`vd_shoc_decomp/vd_shoc_solver`) has a different order of operations that the bfb solver in `ekat_tridiag.hpp`. I have re worked the solver in shoc.F90 so that when the solver is added to C++ we can have a BFB version for testing. The new `vd_shoc_decomp` returns `du` (superdiag), `dl` (subdiag), and `d` (diag), instead of `ca, cc, denom, ze`, and now `vd_shoc_solve` only requires `du, dl, d, rhs`.

As expected, shoc_run_and_cmp fails. The diff on all machines is similar to other non-bfb changes, namely `host_dse` has the largest rel_diff `~ 8e-2`.

I did my own testing, comparing the output of the old solver to the new one. I ran shoc_run_and_cmp and comparing the largest relative diff for any single entry in the output, for any of the solver calls (`5+num_tracers` solves in each `update_prognostics_implicit` call), and found that over every timestep, the largest rel_diff for any single entry was ~ 1e-04(SP) and 1e-12 (DP), which were both in the first timestep. For subsequent timesteps, all largest diffs were ~1e-6 (SP) and 1e- 15(DP).

### DIFF
--- a/components/eam/src/physics/cam/shoc.F90
+++ b/components/eam/src/physics/cam/shoc.F90
@@ -3413,20 +3413,25 @@ subroutine vd_shoc_decomp( &
   du(:,nlev) = 0._rtype
   dl(:,1)    = 0._rtype
 
-  do k=1,nlev
+  ! Compute the diagonal and perform Thomas factorization. The diagonal
+  ! elements are a combination of du and dl (d=1-du-dl). Surface fluxes
+  ! are applied explicitly in the diagonal at the top level.
+  do i=1,shcol
+    d(i,1) = 1._rtype - du(i,1)
+  enddo
+  do k=2,nlev-1
     do i=1,shcol
-      ! Compute the diagonal. The diagonal elements are a combination
-      ! of du and dl (d=1-du-dl). Surface fluxes are applied explicitly
-      ! in the diagonal at the top level.
       d(i,k) = 1._rtype - du(i,k) - dl(i,k)
-      if (k .eq. nlev) d(i,k) = d(i,k) + flux(i)*dtime*ggr*rdp_zt(i,nlev)
 
-      ! Compute the Thomas factorization
-      if (k .ne. 1) then
-        dl(i,k) = dl(i,k)/d(i,k-1)
-        d (i,k) = d (i,k) - dl(i,k)*du(i,k-1)
-      endif
+      dl(i,k) = dl(i,k)/d(i,k-1)
+      d (i,k) = d (i,k) - dl(i,k)*du(i,k-1)
     enddo
+  enddo
+  do i=1,shcol
+    d(i,nlev) = 1._rtype - dl(i,nlev) + flux(i)*dtime*ggr*rdp_zt(i,nlev)
+
+    dl(i,nlev) = dl(i,nlev)/d(i,nlev-1)
+    d (i,nlev) = d (i,nlev) - dl(i,nlev)*du(i,nlev-1)
   enddo
 
   return
--- a/components/eam/src/physics/cam/shoc.F90
+++ b/components/eam/src/physics/cam/shoc.F90
@@ -693,10 +693,9 @@ subroutine update_prognostics_implicit( &
   real(rtype) :: flux_dummy(shcol)
   real(rtype) :: ksrf(shcol), wtke_sfc(shcol)
 
-  real(rtype) :: ca(shcol,nlev) ! superdiagonal for solver
-  real(rtype) :: cc(shcol,nlev) ! subdiagonal for solver
-  real(rtype) :: denom(shcol,nlev) ! denominator in solver
-  real(rtype) :: ze(shcol,nlev)
+  real(rtype) :: du(shcol,nlev) ! superdiagonal for solver
+  real(rtype) :: dl(shcol,nlev) ! subdiagonal for solver
+  real(rtype) :: d(shcol,nlev)  ! diagonal for solver
 
   ! linearly interpolate tkh, tk, and air density onto the interface grids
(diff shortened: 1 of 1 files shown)


jeff-cohere (commented):
This code looks good to me. I'm not sure what to think of the tolerances you mentioned on the first step, but otherwise I have no questions.

components/eam/src/physics/cam/shoc.F90
```diff
@@ -3482,49 +3458,46 @@ subroutine vd_shoc_solve(&
...
+  temp_dl(:,:) = dl(:,:)
+  temp_d (:,:) = d (:,:)
 
-  ! Calculate zf(k). Terms zf(k) and ze(k) are required in solution of
-  ! tridiagonal matrix defined by implicit diffusion equation.
-  ! Note that only levels ntop through nbot need be solved for.
-
-  do i=1,shcol
-    zf(i,nlev) = var(i,nlev) * denom(i,nlev)
+  ! Compute Thomas factorization
```
ambrad, line 3481: I think this factorization block can be moved to the vd_shoc_decomp routine. Overwrite dl and d there. Then these can be used in vd_shoc_solve repeatedly without temp_dl,d. That should be more efficient (save redundant work) and mimic what will happen in the C++ code.

tcclevenger: Good point! I moved it.

components/eam/src/physics/cam/shoc.F90
```diff
@@ -3390,54 +3390,49 @@ subroutine vd_shoc_decomp( &
...
+    d(i,nlev) = 1._rtype - dl(i,nlev) + flux(i)*dtime*ggr*rdp_zt(i,nlev)
   enddo
-
   do k=nlev-1,2,-1
     do i=1,shcol
-      denom(i,k) = 1._rtype/ &
-        (1._rtype + ca(i,k) + cc(i,k) - &
-    ca(i,k) * ze(i,k+1))
-      ze(i,k) = cc(i,k) * denom(i,k)
+      d(i,k) = 1._rtype - du(i,k) - dl(i,k)
```
ambrad, line 3423: Two things here:
1. You don't need to march backward in k b/c there are no deps on k+/-1 in this loop.
2. You can combine this loop with the next. In C++ that won't be true since the factorization is a separate routine, but in F90 it makes sense to do so.

tcclevenger: I also included other simplifications in addition to these.

components/eam/src/physics/cam/shoc.F90
```diff
@@ -3413,25 +3413,19 @@ subroutine vd_shoc_decomp( &
...
-  enddo
+      if (k .eq. nlev) d(i,k) = d(i,k) + flux(i)*dtime*ggr*rdp_zt(i,nlev)
 
-  ! Compute the Thomas factorization
-  do k=2,nlev
-    do i=1,shcol
-      dl(i,k) = dl(i,k)/d(i,k-1)
-      d (i,k) = d (i,k) - dl(i,k)*du(i,k-1)
+      ! Compute the Thomas factorization
+      if (k .ne. 1) then
```
ambrad, line 3425: Sorry to be picky, but we really shouldn't have `if` blocks inside the `i` loop. I recommend just moving the `k=1` case outside the `k` loop. A less attractive alternative, but not wholly unjustified, is to move the `i` loop into the `if` blocks.

tcclevenger: If we remove both if-statements out of the i-loop then I would need the factorization code twice:
```
do i=1,shcol
  d(i,1) = 1._rtype - du(i,1)
enddo
do k=2,nlev-1
  do i=1,shcol
    d(i,k) = 1._rtype - du(i,k) - dl(i,k)

    dl(i,k) = dl(i,k)/d(i,k-1)
    d (i,k) = d (i,k) - dl(i,k)*du(i,k-1)
  enddo
enddo
do i=1,shcol
  d(i,nlev) = 1._rtype - dl(i,nlev) + flux(i)*dtime*ggr*rdp_zt(i,nlev)

  dl(i,nlev) = dl(i,nlev)/d(i,nlev-1)
  d (i,nlev) = d (i,nlev) - dl(i,nlev)*du(i,nlev-1)
enddo
```

The code is _slightly_ more concise if we separate the diagonal and factorization computation:
```
do k=1,nlev
  do i=1,shcol
    d(i,k) = 1._rtype - du(i,k) - dl(i,k)
  enddo
enddo
do i=1,shcol
  d(i,nlev) = d(i,nlev) + flux(i)*dtime*ggr*rdp_zt(i,nlev)
enddo

do k=2,nlev
  do i=1,shcol
    dl(i,k) = dl(i,k)/d(i,k-1)
    d (i,k) = d (i,k) - dl(i,k)*du(i,k-1)
  enddo
enddo
```
What's your opinions? I'm ok with either.

ambrad: I'd say either use the first above or revert to two loops. I personally like the first one above. It's pretty standard to have to deal with boundary conditions, and they do tend to lead to slightly more verbose code if the code is written to cache well, move through memory linearly, and vectorize well.

tcclevenger: I went with the first above.

ambrad (approved):
Great! Please don't merge until @bogensch has a change to look at this PR.

jgfouca (approved):

jeff-cohere (approved):

bogensch (approved):

PR #732, opened 2020-11-11
Fixes memory leak in calcsize_diag

singhbalwinder:
This PR fixes a memory leak by removing "allocate" statement and changes all pointers to allocatable arrays in calsize_diag and wateruptake routines

### DIFF
--- a/components/eam/src/chemistry/utils/modal_aero_calcsize.F90
+++ b/components/eam/src/chemistry/utils/modal_aero_calcsize.F90
@@ -21,6 +21,8 @@ use constituents,     only: pcnst, cnst_name
 
 use ref_pres,         only: top_lev => clim_modal_aero_top_lev
 
+use shr_log_mod ,     only: errMsg => shr_log_errMsg
+
 #ifdef MODAL_AERO
 
 ! these are the variables needed for the diagnostic calculation of dry radius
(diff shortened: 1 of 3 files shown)


ndkeen (approved):
I've not tested this code, but it looks like the memory leak is addressed here. Not sure about the other changes, but in general, I like allocatable over pointer.

ambrad (approved):
This is fine. It looks like it's a verbatim part of https://github.com/E3SM-Project/E3SM/pull/3932/commits. If you wanted to keep things really clean w.r.t. the upstream E3SM repo, you could create a PR there with just this commit and then merge it to both E3SM and SCREAM, then later finish up PR 3932. But I'm fine with this copy-paste approach, as any potential conflict in a future upstream merge should be straightforward to fix.

PeterCaldwell (approved):
I don't fully understand this code, but what I do understand looks good. Thanks for doing this!

singhbalwinder:
Thank you all for your quick reviews. I tested it with `SMS_D.ne4pg2_ne4pg2.F2010-SCREAM-HR` test and the test ran fine with these changes.

@ambrad : I like your suggestion but I think creating multiple PRs for V2 will mostly likely take more time to integrate than a single PR. I took the code as is from the E3SM PR so (like you said) it will be easy to resolve conflicts if they arise during the upstream merge.

PeterCaldwell:
Subtle note here: we're just merging this PR into our "dyamond2" branch rather than into scream master... we'll wait for an e3sm upstream merge to bring in the memleak fix into the main repo. Thus we don't need to worry about merge issues later.

PR #733, opened 2020-11-11
Hack Koby conservation checks into 'dyamond2' branch

PeterCaldwell:
This PR is identical to #714 but will merge into dyamond2 branch rather than scream master. Also, #714 won't be integrated until it has a working C++ impl, while this PR will be merged immediately since the C++ code in the dyamond2 branch will never be used. Instead, the changes in this PR are implemented inside a "Koby_fixes" logical flag which is true by default. 

What this PR does is 1). make sure vapor deposition doesn't leave qv subsaturated with respect to ice and 2). make sure number concentrations don't go negative. These fixes came from Koby but were modified to fix with SCREAM variable names etc. 

I was expecting that these changes wouldn't be triggered often so wouldn't affect model climate, but they seem to decrease SWCF by 5 W/m2 (with little effect on LWCF). Thus BL clouds seem to depend heavily on these limiters, which should be fixed in longer-term research.

### DIFF
--- a/components/eam/src/physics/cam/micro_p3.F90
+++ b/components/eam/src/physics/cam/micro_p3.F90
@@ -818,7 +818,6 @@ contains
       call prevent_ice_overdepletion(pres(k), t_atm(k), qv(k), latent_heat_sublim(k), inv_dt, qidep, qi2qv_sublim_tend)
 
       if (Koby_fixes) then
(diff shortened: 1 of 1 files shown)


PeterCaldwell:
@crterai - this is an update of the Koby fix PR you already looked through in detail. Probably worth glancing at (or maybe better to just look at the changes I've made since you last approved in #714 ).

crterai (commented):
New changes look good!

PR #735, opened 2020-11-11
Get scream scripts working on cori

### LINKED ISSUE #436: Difficulty building scream unit tests on Cori-KNL
Running test-all-scream on Cori I run into issues.  Running `test_all_scream` and `gather_all_data` results in the same fail during the building of baselines:

```
RUN: cmake -DCMAKE_CXX_COMPILER=/opt/cray/pe/craype/2.5.18/bin/CC -C /global/u1/a/adonahue/Code/scream/components/scream/cmake/machine-files/cori.cmake -DCMAKE_BUILD_TYPE=Debug -DSCREAM_PACK_SIZE=1 -DSCREAM_SMALL_PACK_SIZE=1 -DSCREAM_BASELINES_ONLY=ON /global/u1/a/adonahue/Code/scream/components/scream
FROM: /global/u1/a/adonahue/Code/scream/components/scream/ctest-build/baselines/debug_nopack_fpe
WARNING: Failed to configure baselines:
CMake Error at ekat/cmake/YamlCpp.cmake:24 (add_subdirectory):
  The source directory

    /global/u1/a/adonahue/Code/scream/externals/yaml-cpp

  does not contain a CMakeLists.txt file.
Call Stack (most recent call first):
  ekat/CMakeLists.txt:26 (include)


CMake Warning (dev) at ekat/cmake/EkatSetCompilerFlags.cmake:91 (find_package):
  Policy CMP0074 is not set: find_package uses <PackageName>_ROOT variables.
  Run "cmake --help-policy CMP0074" for policy details.  Use the cmake_policy
  command to set the policy and suppress this warning.

  Environment variable CUDA_ROOT is set to:

    /usr/common/software/cuda/10.2.89

  For compatibility, CMake is ignoring the variable.
Call Stack (most recent call first):
  ekat/CMakeLists.txt:78 (SetCompilerFlags)
This warning is for project developers.  Use -Wno-dev to suppress it.
```

looking at the machine_specs file is looks like Cori doesn't load any modules in gather_all_data:

```
    "cori"   : (["eval $(../../cime/scripts/Tools/get_case_env)", "export OMP_NUM_THREADS=68"],
                "$(which CC)",
                "srun --time 02:00:00 --nodes=1 --constraint=knl,quad,cache --exclusive -q regular --account e3sm",
                68,
                68,
                ""),
```

jgfouca:
1) Remove dead symlink cori-knl.cmake and replace with cori.cmake
2) Fix subtle problem with cmake's openmpi lib/link detection on cori. The default behavior links `-lifcore` which leaves `for_abort` undefined. Linking `-lifport` instead resolves this problem. After lots of experimentation, I was unable to achieve this swap via cached cmake variables, so I had to add a special hack to our CMakeLists.txt.

On issue left unresolved is cori's poor performance when running test-all-scream. Despite some tweaks, I was unable to get things fast enough to use cori's debug queue (which has a 30m max) which means we have to use the regular queue which is much busier. The build performance is not great but the cmake configure commands are insanely slow when run on cori compute nodes. In the couple I measured, the configure command took longer (6m) than the entire build (5m).

Fixes #436


jgfouca:
Had to upstream merge to resolve conflicts.

components/scream/CMakeLists.txt
```diff
@@ -43,6 +43,11 @@ set (CMAKE_CXX_STANDARD 11)
 if (NOT CIME_BUILD)
   project(SCREAM CXX Fortran)
 
+  if (SCREAM_CORI_HACK)
```
bartgol, line 46: Maybe we should simply probe the env, like Kokkos does. We believe (I think) that this issue is caused by Cray compiler wrapper, we could simply do `if ($ENV{CRAYPE_VERSION})`, avoiding misterious cache vars.

bartgol:
@jgfouca the config times are quite hard to improve. Perhaps we could use `CMAKE_<LANG>_COMPILER_WORKS=ON`, just like cime does.

As for the rest, your config+build time is 11m. Is the run phase 20+ minutes? Or are we going above 30m due to the three different builds? If it's the former, we could limit the threads/ranks configurations (e.g., we could set SCREAM_TEST_THREAD_INC=4, or simply test 1 thread and $max_threads). If the latter, we could try to split the 3 builds into 3 separate jobs...

jgfouca:
@bartgol 

> Is the run phase 20+ minutes? Or are we going above 30m due to the three different builds?

No, the run phases are relatively fast (about 2m). Doing the three different builds in serial ( (6m config + 5m build) * 3) > 30m is what kills our chances of fitting into the 30m window. I will try a parallel run.

components/scream/CMakeLists.txt
jgfouca: That would be better except that we never established that it's a problem with Cray verses something specific to Cori (such as a bad  mixture of versions of cmake/intel etc). I don't think we have access to any other cray machine that works for scream so we can't prove our theory. In the absence of additional evidence, I'm reluctant to spread this very-hacky solution to other systems.

bartgol: Fair enough.

bartgol (approved):

jgfouca:
@bartgol , even with `-p`, test-all-scream doesn't quite finish in 30m.

bartgol:
> @bartgol , even with `-p`, test-all-scream doesn't quite finish in 30m.

Looks like we need to issue three separate gather-all commands. And being careful that baselines are already generated (otherwise one job may switch branch messing up the other job). Perhaps we should add a flag to test-all, that errors out if baselines are not present or expired (preventing test-all from changing the src tree).

PR #736, opened 2020-11-11
Update prescribed aerosol field for scream compsets

crterai:
New prescribed aerosols were created from a maint-1.0 version of E3SM using the F2010C5-CMIP6-LR compset. 
The PR points to these new set of prescribed aerosol fields.

### DIFF
--- a/components/eam/bld/namelist_files/use_cases/2000_cam5_av1c-04p2-MMF-1mom.xml
+++ b/components/eam/bld/namelist_files/use_cases/2000_cam5_av1c-04p2-MMF-1mom.xml
@@ -88,7 +88,7 @@
 
 <!-- Prescribed aerosols -->
 <prescribed_aero_datapath>atm/cam/chem/trop_mam/aero</prescribed_aero_datapath>
-<prescribed_aero_file>mam4_0.9x1.2_L72_2000clim_c170323.nc</prescribed_aero_file>
+<prescribed_aero_file>mam4_0.9x1.2_L72_E3SMv1_c201110.nc</prescribed_aero_file>
 <use_hetfrz_classnuc>.false.</use_hetfrz_classnuc>
 <prescribed_aero_type>CYCLICAL</prescribed_aero_type>
 <prescribed_aero_cycle_yr>01</prescribed_aero_cycle_yr>
--- a/components/eam/bld/namelist_files/use_cases/2000_cam5_av1c-04p2-MMF-1mom.xml
+++ b/components/eam/bld/namelist_files/use_cases/2000_cam5_av1c-04p2-MMF-1mom.xml
@@ -88,7 +88,7 @@
 
 <!-- Prescribed aerosols -->
 <prescribed_aero_datapath>atm/cam/chem/trop_mam/aero</prescribed_aero_datapath>
-<prescribed_aero_file>mam4_0.9x1.2_L72_E3SMv1_c201110.nc</prescribed_aero_file>
+<prescribed_aero_file>mam4_0.9x1.2_L72_2000clim_c170323.nc</prescribed_aero_file>
 <use_hetfrz_classnuc>.false.</use_hetfrz_classnuc>
 <prescribed_aero_type>CYCLICAL</prescribed_aero_type>
 <prescribed_aero_cycle_yr>01</prescribed_aero_cycle_yr>
--- a/components/eam/bld/namelist_files/use_cases/2000_scream_hr.xml
+++ b/components/eam/bld/namelist_files/use_cases/2000_scream_hr.xml
@@ -80,13 +80,13 @@
 <use_hetfrz_classnuc>.false.</use_hetfrz_classnuc>
 <aerodep_flx_type>'CYCLICAL'</aerodep_flx_type>
 <aerodep_flx_datapath>atm/cam/chem/trop_mam/aero</aerodep_flx_datapath>
-<aerodep_flx_file nlev="72">mam4_0.9x1.2_L72_2000clim_c170323.nc</aerodep_flx_file>
-<aerodep_flx_file nlev="128">mam4_0.9x1.2_L128_2000clim_c191106.nc</aerodep_flx_file>
+<aerodep_flx_file nlev="72">mam4_0.9x1.2_L72_E3SMv1_c201110.nc</aerodep_flx_file>
+<aerodep_flx_file nlev="128">mam4_0.9x1.2_L128_E3SMv1_c201110.nc</aerodep_flx_file>
 <aerodep_flx_cycle_yr>01</aerodep_flx_cycle_yr>
 <prescribed_aero_type>'CYCLICAL'</prescribed_aero_type>
 <prescribed_aero_datapath>atm/cam/chem/trop_mam/aero</prescribed_aero_datapath>
-<prescribed_aero_file nlev="72">mam4_0.9x1.2_L72_2000clim_c170323.nc</prescribed_aero_file>
-<prescribed_aero_file nlev="128">mam4_0.9x1.2_L128_2000clim_c191106.nc</prescribed_aero_file>
+<prescribed_aero_file nlev="72">mam4_0.9x1.2_L72_E3SMv1_c201110.nc</prescribed_aero_file>
+<prescribed_aero_file nlev="128">mam4_0.9x1.2_L128_E3SMv1_c201110.nc</prescribed_aero_file>
 <prescribed_aero_cycle_yr>01</prescribed_aero_cycle_yr>
 
 <!-- sim_year used for CLM datasets and SSTs forcings -->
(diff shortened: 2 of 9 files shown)


PeterCaldwell (approved):

bogensch (approved):

crterai:
I have three more use case files to update. Will push those shortly.

components/eam/bld/namelist_files/use_cases/2000_cam5_av1c-04p2-MMF-1mom.xml
```diff
@@ -88,7 +88,7 @@
 
 <!-- Prescribed aerosols -->
 <prescribed_aero_datapath>atm/cam/chem/trop_mam/aero</prescribed_aero_datapath>
-<prescribed_aero_file>mam4_0.9x1.2_L72_2000clim_c170323.nc</prescribed_aero_file>
```
brhillman, line 91: If these new files were made using the F2010 compset, do we really want to update compsets based on F2000?

components/eam/bld/namelist_files/use_cases/2000_scream_hr.xml
```diff
@@ -80,13 +80,13 @@
 <use_hetfrz_classnuc>.false.</use_hetfrz_classnuc>
 <aerodep_flx_type>'CYCLICAL'</aerodep_flx_type>
 <aerodep_flx_datapath>atm/cam/chem/trop_mam/aero</aerodep_flx_datapath>
-<aerodep_flx_file nlev="72">mam4_0.9x1.2_L72_2000clim_c170323.nc</aerodep_flx_file>
```
brhillman, line 83: Again I wonder if replacing F2000 prescribed aero with F2010 is a good idea here. As an aside, I think these "non-dyamond2-scream-compsets" should probably just be changed to "dyamond1 compsets", so this particular change is probably fine, so long as we then change this to a 2010_scream_hr_dyamond1.xml compset and update forcings. At that point though, it probably makes more sense to keep this an F2000 and base a new dyamond1 compset on the F2010.

components/eam/bld/namelist_files/use_cases/2000_scream_lr.xml
```diff
@@ -80,13 +80,13 @@
 <use_hetfrz_classnuc>.false.</use_hetfrz_classnuc>
 <aerodep_flx_type>'CYCLICAL'</aerodep_flx_type>
 <aerodep_flx_datapath>atm/cam/chem/trop_mam/aero</aerodep_flx_datapath>
-<aerodep_flx_file nlev="72">mam4_0.9x1.2_L72_2000clim_c170323.nc</aerodep_flx_file>
```
brhillman, line 83: Same comment as above about mixing F2000 and F2010.

components/eam/bld/namelist_files/use_cases/RCEMIP_EAMv1.xml
```diff
@@ -152,7 +152,7 @@
 
 <!-- Prescribed aerosols -->
 <prescribed_aero_datapath>atm/cam/chem/trop_mam/aero</prescribed_aero_datapath>
```
brhillman, line 154: I'm not 100% sure we want to update this here. But I defer to @hassanbeydoun .

components/eam/bld/namelist_files/use_cases/aquaplanet_MMF-1mom.xml
```diff
@@ -125,7 +125,7 @@
 
 <!-- Prescribed aerosols -->
 <prescribed_aero_datapath>atm/cam/chem/trop_mam/aero</prescribed_aero_datapath>
```
brhillman, line 127: Since the MMF aquaplanet compset is I think more consistent with F2000 forcings, I'm not sure this should be changed. @whannah1 might be interested in this.

brhillman (commented):
I left some comments/questions. My main concern is that this is updating some F2000-based compsets with prescribed aerosol data based on a F2010 compset. I'm not sure this makes the most sense, but I may be misunderstanding what's going on here.

components/eam/bld/namelist_files/use_cases/2000_cam5_av1c-04p2-MMF-1mom.xml
crterai: Fair point about merge conflicts in the future. I think @PeterCaldwell's reasoning behind changing these files as well was that we don't know what simulation produced the original prescribed aerosol file, so it's better to use one that's not quite consistent (2000 vs 2010). 
For the scream F2000 compset, I do think that we should use the newer aerosol field because we know what kind of model/simulation they came from. For the non-scream compsets, I don't have a strong opinion so I can revert them as well. Any thoughts, @PeterCaldwell?

PeterCaldwell: The merge conflict question is a good one... but irrelevant for the dyamond2 branch we're working on now because we will never merge upstream into it. The 2000 vs 2010 issue is legit, but I'm pretty sure that old prescribed_aero_file was created by @bogensch  from some random v1 alpha run with no real provenance... which is why I thought it would be better to switch to a maint1.0 tag even if we use 2000 instead of 2010. The other issue is that only a Luddite would use an F2000 compset these days so they deserve to get an inconsistent prescribed aerosol file. All in all, I don't care what we do for these F2000 cases (because I'm not a Luddite) and think whatever allows us to merge this PR and move on to other things is best. Sorry for suggesting you make a change that doesn't really matter, Chris!

components/eam/bld/namelist_files/use_cases/RCEMIP_EAMv1.xml
whannah1: This "should not" matter. In RCE the aerosol-cloud effects are disabled via the constant droplet number parameter, and the aerosol radiative effects are disabled directly in RRTMGP. I used prescribed aerosols here because there's not a "no aerosols" option.

components/eam/bld/namelist_files/use_cases/aquaplanet_MMF-1mom.xml
whannah1: All aqua planet compsets were loosely based on F2000 when it comes to tuning. Unlike RCE, I'm pretty sure the direct aerosol radiative effects are still present, but the cloud-aerosol effects are disabled. Either way, as long as the new prescribed aerosol data doesn't break these compsets, I wouldn't worry about how the subtle changes it might make to the solution.

PeterCaldwell: Isn't it alarmingly wrong to have spatially varying prescribed aerosols created under the assumption that we had continents in your aquaplanet config? Seems like you'd see outlines of the continents in the clear-sky TOA radiative fluxes...

whannah1: Agreed. I'll need to look back into this, but it doesn't  really hurt the utility of the aqua planet for testing/development. I know that I couldn't disable aerosol rad effects when I revived the aqua compset because I added that feature later for RCE, and only for RRTMGP. My immediate worry is that it would involve a lot of careful modifications to bypass all chemistry routines to implement a "no aerosol" option, which could be an annoying task. 

We could also consider creating a new prescribed aerosol file that just globally homogenizing the concentrations. That is probably way easier.

components/eam/bld/namelist_files/use_cases/2000_cam5_av1c-04p2-MMF-1mom.xml
PeterCaldwell: Ok, reverting back to old file: if people want to use an old F compset, they should expect to get an old/ratty aerosol file.

components/eam/bld/namelist_files/use_cases/2000_scream_hr.xml
PeterCaldwell: Again reverting to the old input file for F2000 compsets. Switching DYAMOND1 to have its own use case based on 2010 is a good idea... for another PR at a later date.

components/eam/bld/namelist_files/use_cases/2000_scream_lr.xml
PeterCaldwell: reverted.

components/eam/bld/namelist_files/use_cases/RCEMIP_EAMv1.xml
PeterCaldwell: Let's keep the F2010 value here because it shouldn't matter, RCE is no more related to year 2000 than year 2010, empirically, the impact of this switch seems to be negligible, and the 2010 value comes from a more credible/known simulation than the 2000 values did.

components/eam/bld/namelist_files/use_cases/aquaplanet_MMF-1mom.xml
PeterCaldwell: Ok, that makes sense. Since aquaplanet isn't really tied to year 2000 versus 2010 forcings, let's go with the new 2010 forcings which have better provenance.

crterai (commented):
Looks good to me

PR #740, opened 2020-11-16
use bfb_* calls

tcclevenger:
A few changes to `bfb_*()` calls.

### DIFF
--- a/components/eam/src/physics/cam/shoc.F90
+++ b/components/eam/src/physics/cam/shoc.F90
@@ -829,7 +829,7 @@ subroutine dp_inverse(nlev, shcol, rho_zt, dz_zt, rdp_zt)
 
 end subroutine dp_inverse
 
-pure function impli_srf_stress_term(shcol, rho_zi_sfc, uw_sfc, &
+function impli_srf_stress_term(shcol, rho_zi_sfc, uw_sfc, &
      vw_sfc, u_wind_sfc, v_wind_sfc) result (ksrf)
 
   !intent-ins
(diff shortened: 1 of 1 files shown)


components/eam/src/physics/cam/shoc.F90
```diff
@@ -900,8 +900,8 @@ pure function tke_srf_flux_term(shcol, uw_sfc, vw_sfc) result(wtke_sfc)
   do i = 1, shcol
      uw           = uw_sfc(i)
      vw           = vw_sfc(i)
-     ustar        = max(sqrt(sqrt(uw**2._rtype + vw**2._rtype)),ustarmin)
-     wtke_sfc(i) = ustar**3
+     ustar        = max(bfb_sqrt(bfb_sqrt(bfb_square(uw) + bfb_square(vw))),ustarmin)
+     wtke_sfc(i) = bfb_cube(ustar)
```
tcclevenger, line 904: @jeff-cohere @ambrad @jgfouca 
Like other changes of this type, the there is only test that fails (SP on Blake) and, if it is like the other changes I've done in the past, it only comes from this `bfb_cube` (`bfb_sqrt` and `bfb_square` have always been b4b in my changes). Unfortunately I can't seem to reserve an allocation on Blake yesterday or so far today to calculate the exact relative error of this call. Since both mappy and weaver have full passes I'm confident there is no error, but if anyone prefers I wait to actually test on Blake that is fine.

jeff-cohere (approved):
This looks okay to me. Just curious: did you have to remove the `pure` keyword from these functions because our BFB math functions aren't "pure"? Is this a limitation of C-interoperable functions/procedures?

ambrad (approved):

jgfouca (approved):

tcclevenger:
> This looks okay to me. Just curious: did you have to remove the `pure` keyword from these functions because our BFB math functions aren't "pure"? Is this a limitation of C-interoperable functions/procedures?

@jeff-cohere Yes, `pure` had to be removed for that reason. Not sure if it can be remedied.

PR #741, opened 2020-11-17
SCREAM: allow to specify a work dir for test-all-scream

bartgol:
Upon request from @AaronDonahue , I added the capability of requesting a specific folder for test-all-scream to work in.

This can be helpful for users on systems with limited disk quota, so that they can keep the src code in their home folder, but dump all the build/testing stuff in a scratch location.

The new option is to be used as `test-all-scream -w some_dir` or `test-all-scream --work-dir some_dir`. If not provided, we build in the same folder as we used to (i.e., `/path/to/scream-src-folder/ctest-build`).

### DIFF
--- a/components/scream/cmake/ctest_script.cmake
+++ b/components/scream/cmake/ctest_script.cmake
@@ -2,9 +2,9 @@ cmake_minimum_required(VERSION 3.9)
 
 set(CTEST_BUILD_NAME "scream_unit_tests_${BUILD_NAME_MOD}")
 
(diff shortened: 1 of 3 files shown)


bartgol:
In the spirit of keeping disk quota pressure low, I also pushed a commit that removes all files in the baselines directory, except for baselines themselves. This is done only if baselines are successfully generated. If not, every file is kept in the folder (for inspection purposes).

jgfouca (approved):

AaronDonahue:
Thanks for putting this in @bartgol!

bartgol:
Since I only fixed a typo since the last approval, I'm going to merge now without further review.

PR #743, opened 2020-11-18
Beydoun/spa optics 2

hassanbeydoun:
This PR adds the optics portion of simple-prescribed-aerosol (SPA). 3d_4d and 4d_4d infld options were added to ncdio_atm.F90 since aerosol optical properties are 3d (X,Z, and radiation bands) variables. SPA optics reads in 1 lw aerosol optical property and 3 sw aerosol optical properties.


components/eam/src/physics/rrtmgp/radiation.F90
```diff
@@ -103,6 +110,10 @@ module radiation
    ! zeroes out the aerosol optical properties if False
    logical :: do_aerosol_rad = .true.
 
+   ! Logical to indicate if aerosol optical properties are read in directly from
+   ! input file i.e., SPA:
+   logical :: do_SPA_optics = .false.
+
```
PeterCaldwell, line 116: Just to make sure I'm not missing something: you're initializing to false here but then overwriting with whatever was specified in the namelist, right?

hassanbeydoun: That's right.

components/eam/src/physics/rrtmgp/radiation.F90
```diff
@@ -413,14 +441,71 @@ end function radiation_nextsw_cday
...
+      character(len=20) :: mon_str
+      type(file_desc_t) :: nccn_ncid
+      integer :: year, month, day, tod, next_month, grid_id
+      logical :: found = .false.
+
+      write(mon_str,*) month_int
+
+      !assign base_file_name the name of the CCN file being used 
+      base_file_name = "spa_optics_file_"
+      optics_lookup_dir = "/p/lustre2/beydoun1/aerosol_optics"
```
PeterCaldwell, line 470: This is ok for right now, but is hardcoded for LC machines. We will need to move this file to inputdata and add it to the namelist specification (perhaps in another PR)

components/eam/src/physics/rrtmgp/radiation.F90
```diff
@@ -413,14 +441,71 @@ end function radiation_nextsw_cday
...
+
+      mon_str = adjustl(mon_str)
+
+     !retrieve the name of the relevant file by combining base file name !with
+     !!month and full file path:
+      filename = trim(optics_lookup_dir)//'/'//trim(base_file_name)//trim(mon_str)//'.nc'
+
+      write(iulog,*) 'file name is', filename
```
PeterCaldwell, line 478: these writeouts are probably not needed once debugging is over. If they are permanent, you should be more clear about what they are. E.g. "file name" => "Prescribed aerosol file name"...

components/eam/src/physics/rrtmgp/radiation.F90
AaronDonahue: It might be good to change this in the namelist now.  No one but Hassan will have access to this folder unless he changes the permissions.  @hassanbeydoun , you can follow the example of any of the input file variables in the namelist_definition and namelist_defaults files.  For example, the most recent would be the 'p3_lookup' tables.  You will also need to copy the aerosol optics files to the usr/climdat/... folder so that it is accessible by all.

components/eam/src/physics/rrtmgp/radiation.F90
```diff
@@ -871,6 +964,58 @@ subroutine radiation_init(state)
...
+         if (month==12) then
+            next_month = 1
+         else
+            next_month = month + 1
+         end if
+
+         allocate(aerosol_optical_property(pcols,pver,nlwbands,begchunk:endchunk))
+         call get_aerosol_optical_property_from_file(current_month,'AER_TAU_LW','lwband',nlwbands,aerosol_optical_property)
+         call pbuf_set_field(pbuf, aer_tau_bnd_lw_mon_1_idx,aerosol_optical_property)
+         deallocate(aerosol_optical_property)
```
AaronDonahue, line 982: The use of allocate and deallocate in this chunk of code I think is a way to re-initialize and recycle the `aerosol_optical_property` array.  I think you can also just get away with setting the whole array to a value each time, like `aerosol_optical_property(:,:,:,:) = 0.0` or something along those lines.

components/eam/src/physics/rrtmgp/radiation.F90
```diff
@@ -871,6 +964,58 @@ subroutine radiation_init(state)
...
+         allocate(aerosol_optical_property(pcols,pver,nlwbands,begchunk:endchunk))
+         call get_aerosol_optical_property_from_file(next_month,'AER_TAU_LW','lwband',nlwbands,aerosol_optical_property)
+         call pbuf_set_field(pbuf,aer_tau_bnd_lw_mon_2_idx,aerosol_optical_property)
+         deallocate(aerosol_optical_property)
+
+         allocate(aerosol_optical_property(pcols,pver,nswbands,begchunk:endchunk))
+         call get_aerosol_optical_property_from_file(current_month,'AER_TAU_SW','swband',nswbands,aerosol_optical_property)
+         call pbuf_set_field(pbuf, aer_tau_bnd_sw_mon_1_idx,aerosol_optical_property)
+         deallocate(aerosol_optical_property)
```
AaronDonahue, line 991: Of course if you decide to remove some of the deallocation steps you will need to make sure you keep the one that happens before you switch size from `lwbands` to `swbands`.

Alternatively, you can just declare `aerosol_optical_property_lw` and `aerosol_optical_property_sw` arrays at the top and not have to `allocate/deallocate` anything.

components/eam/src/physics/rrtmgp/radiation.F90
```diff
@@ -1180,6 +1325,24 @@ subroutine radiation_tend(state,   ptend,    pbuf,          cam_out, cam_in,  &
...
+      real(r8), pointer :: aer_tau_bnd_sw_mon_2(:,:,:)
+      real(r8), pointer :: aer_ssa_bnd_sw_mon_1(:,:,:)
+      real(r8), pointer :: aer_ssa_bnd_sw_mon_2(:,:,:)
+      real(r8), pointer :: aer_asm_bnd_sw_mon_1(:,:,:)
+      real(r8), pointer :: aer_asm_bnd_sw_mon_2(:,:,:)
+
+      real(r8), dimension(12):: days_per_month
+      nullify(aerosol_optical_property)
+      !fill days_per_month
+      days_per_month = (/31,28,31,30,31,30,31,31,30,31,30,31/)
```
AaronDonahue, line 1345: Maybe add a comment here that SPA optics is not compatible with Leap Year.  It goes without saying really, but if in sometime in the distant future someone does decide to add leap year in it's nice to give them a fighting chance to find all the changes they need to make.  Being able to grep "leap year" might be something that they do.

AaronDonahue (changes requested):
Looks good Hassan! I had a few comments and a followup on Peter's comment about the hard-coded path to the input files.

components/eam/src/physics/rrtmgp/radiation.F90
PeterCaldwell: Agreed. It is easy to do and won't be useable by anyone other than Hassan if he doesn't do this.

components/eam/src/physics/rrtmgp/radiation.F90
PeterCaldwell: Good idea, Aaron. I like the idea of just allocating both aerosol_optical_property_lw and aerosol_optical_property_sw at the top of the subroutine and just assigning new values for each property. I'm not sure whether zeroing out in between usage is needed - it is safer, but if `get_aerosol_optical_property_from_file` is guaranteed to produce a slab of data of the same size as the allocated chunk of memory (which I think is the case), there's no point and skipping the zeroing would be faster and cleaner.

components/eam/src/physics/rrtmgp/radiation.F90
PeterCaldwell: Good point. It may be worth adding the same comment in the microphysics SPA code as well.

PeterCaldwell:
Looks great, @hassanbeydoun ! I particularly like how `get_aerosol_optical_property_from_file` is reused to keep the SPA sections compact and easy to understand. I think there are just a couple easy fixes left:

1. replace hard-coded file location with standard-practice location. 
2. delete debugging print statements
3. only allocate/deallocate 1x
4. add comment about noleap calendar

components/eam/src/physics/rrtmgp/radiation.F90
```diff
@@ -467,33 +471,23 @@ subroutine get_aerosol_optical_property_from_file(month_int,aerosol_optical_prop
...
-      optics_lookup_dir = "/p/lustre2/beydoun1/aerosol_optics"
 
       mon_str = adjustl(mon_str)
 
      !retrieve the name of the relevant file by combining base file name !with
      !!month and full file path:
-      filename = trim(optics_lookup_dir)//'/'//trim(base_file_name)//trim(mon_str)//'.nc'
-
-      write(iulog,*) 'file name is', filename
+!      filename = trim(optics_lookup_dir)//'/'//trim(base_file_name)//trim(mon_str)//'.nc'
```
PeterCaldwell, line 479: Might as well delete this commented out line.

components/eam/src/physics/rrtmgp/radiation.F90
```diff
@@ -965,52 +966,38 @@ subroutine radiation_init(state,pbuf)
...
+
+         call get_aerosol_optical_property_from_file(current_month,'AER_SSA_SW','swband',nswbands,aerosol_optical_property_sw)
+         call pbuf_set_field(pbuf, aer_ssa_bnd_sw_mon_1_idx,aerosol_optical_property_sw)
+         call get_aerosol_optical_property_from_file(next_month,'AER_SSA_SW','swband',nswbands,aerosol_optical_property_sw)
+         call pbuf_set_field(pbuf,aer_ssa_bnd_sw_mon_2_idx,aerosol_optical_property_sw)
+
+         call get_aerosol_optical_property_from_file(current_month,'AER_G_SW','swband',nswbands,aerosol_optical_property_sw)
+         call pbuf_set_field(pbuf,aer_asm_bnd_sw_mon_1_idx,aerosol_optical_property_sw)
+         call get_aerosol_optical_property_from_file(next_month,'AER_G_SW','swband',nswbands,aerosol_optical_property_sw)
+         call pbuf_set_field(pbuf,aer_asm_bnd_sw_mon_2_idx,aerosol_optical_property_sw)
```
PeterCaldwell, line 1000: Do you need deallocate statements here?

components/eam/src/physics/rrtmgp/radiation.F90
```diff
@@ -1659,10 +1639,9 @@ subroutine radiation_tend(state,   ptend,    pbuf,          cam_out, cam_in,  &
                         else
                            next_month = month + 1
                         end if
-                        allocate(aerosol_optical_property(pcols,pver,nlwbands,begchunk:endchunk))
-                        call get_aerosol_optical_property_from_file(next_month,'AER_TAU_LW','lwband',nlwbands,aerosol_optical_property)
-                        aer_tau_bnd_lw_mon_2 = aerosol_optical_property(:,:,:,state%lchnk)
-                        deallocate(aerosol_optical_property)
+                        allocate(aerosol_optical_property_lw(pcols,pver,nlwbands,begchunk:endchunk))
+                        call get_aerosol_optical_property_from_file(next_month,'AER_TAU_LW','lwband',nlwbands,aerosol_optical_property_lw)
+                        aer_tau_bnd_lw_mon_2 = aerosol_optical_property_lw(:,:,:,state%lchnk)
```
PeterCaldwell, line 1644: Again, do you need to deallocate here?

run_scream.syrah_ne30_standard.csh
```diff
@@ -0,0 +1,1456 @@
+#! /bin/csh -fe
```
PeterCaldwell, line 1: I don't think you meant to include this file in the PR.

PeterCaldwell: Also for the other run_scream file you added too.

PeterCaldwell:
@hassanbeydoun - I just made some trivial comments to address. Once you fix them, please ping Aaron and Ben for a re-review. I'd also like to ask what runs you've done to test this capability. 

I notice that autotester is failing for this PR. I expected this because you've changed the variables being passed to/from functions so they're inconsistent with the C++. So we need to make the C++ version of these changes before we can merge to master. We could potentially merge to dyamond2 branch without making those changes, but I'd rather more thoroughly test the SPA implementation in this branch before doing any merging. I'll think about/try to find someone to add the C++ side.

components/eam/src/physics/rrtmgp/radiation.F90
hassanbeydoun: deleted. Missed it during last cleanup.

components/eam/src/physics/rrtmgp/radiation.F90
hassanbeydoun: it worked without deallocation but I added it just in case.

run_scream.syrah_ne30_standard.csh
hassanbeydoun: Deleted. thought I deleted it last time.

hassanbeydoun:
> @hassanbeydoun - I just made some trivial comments to address. Once you fix them, please ping Aaron and Ben for a re-review. I'd also like to ask what runs you've done to test this capability.
> 
> I notice that autotester is failing for this PR. I expected this because you've changed the variables being passed to/from functions so they're inconsistent with the C++. So we need to make the C++ version of these changes before we can merge to master. We could potentially merge to dyamond2 branch without making those changes, but I'd rather more thoroughly test the SPA implementation in this branch before doing any merging. I'll think about/try to find someone to add the C++ side.

AaronDonahue (approved):
Looks good, approving assuming the you have checked that the tests pass.

components/eam/src/physics/rrtmgp/radiation.F90
PeterCaldwell: I think this kind of stuff causes memory leaks. Maybe not in this case since this is just in init, but probably in the radiation_tend case where you allocate every month but never deallocate. In any case, fixed now.

PeterCaldwell (approved):

PeterCaldwell:
@brhillman - could you review?
@hassanbeydoun - it looks like the AT fail is happening to everyone so I'll manually merge once Ben reviews and I hear from you that sims with and without SPA look vaguely similar.

PeterCaldwell:
I just confirmed that default (NON-SPA) behavior passes ERP and SMS_D tests at ne4pg2 on cori and syrah. ERP flagged a memory leak on Cori which @ndkeen tracked down to the memory leak we had on master when this branch was created. Adding the relevant bugfix (not included in this PR) fixed the problem. Thanks Noel! 

I have not been able to test do_spa_optics =.true. because default prescribed optics files are not available yet. We will add them in a future PR.

PR #744, opened 2020-11-19
Fix a rare race condition in shoc_diag_second_moments_impl

### LINKED ISSUE #742: SHOC tests exibit seemingly random behavior, depending on CTEST_PARALLEL_LEVEL
Try to manually run `ctest -R shoc_tests` and `CTEST_PARALLEL_LEVEL=40 ctest -R shoc_tests`. The latter, on mappy as well as my laptop, passes, while the former has 3 tests failing. A midground with || level 4 has just 1 test failing.

jgfouca:
The tk_zi view was being used as output in one call and input
in a later call without a barrier in-between.

Also, cleanup some whitespace issues.

Fixes #742


bartgol (approved):
Great catch Jim!!!

PR #745, opened 2020-11-19
Machine file fix for the livermore computing machine lassen.

AaronDonahue:
Lassen builds started failing after the homme PR due to an incorrect
path to the blas and lapack libraries being used.  This fixes the
machine files.

[BFB]


components/scream/cmake/machine-files/lassen.cmake
```diff
@@ -5,3 +5,5 @@ include (${EKAT_MACH_FILES_PATH}/kokkos/cuda.cmake)
 set(NetCDF_C_PATHS /usr/gdata/climdat/lassen_builds/libs/netcdf/netcdf-c/netcdf-c-install/ CACHE STRING "")
 set(NetCDF_Fortran_PATHS /usr/gdata/climdat/lassen_builds/libs/netcdf/netcdf-f/netcdf-f-install/ CACHE STRING "")
 set(CMAKE_C_COMPILER /usr/tce/packages/spectrum-mpi/spectrum-mpi-rolling-release-gcc-7.3.1/bin/mpicc CACHE STRING "")
+set(BLAS_LIBRARIES /usr/tcetmp/packages/lapack/lapack-3.9.0-gcc-7.3.1/lib/libblas.so CACHE STRING "")
```
jeff-cohere, line 8: I realize this may be breaking with tradition, and I don't want to burden this PR with the responsibility of owning up to the issue, but: wouldn't it be less brittle for us to use environment variables set by modules on these systems? We can access these environment variables using `$ENV{VAR_NAME}`.

Maybe there are reasons we don't do this, or maybe it's something we could discuss, anyway.

jeff-cohere (approved):
Looks good to me. I left a comment that's not specific to this PR.

components/scream/cmake/machine-files/lassen.cmake
AaronDonahue: @jeff-cohere , I think this is a good point.  I should change a few of these to something more general so protect against module changes in the future.

PR #746, opened 2020-11-20
update ekat

tcclevenger:
Update changes made to `EKAT_DEFAULT_BFB`

### DIFF
--- a/components/scream/scripts/test_all_scream.py
+++ b/components/scream/scripts/test_all_scream.py
@@ -57,14 +57,14 @@ class TestAllScream(object):
                "Makes no sense to specify a baseline generation commit if using pre-existing baselines ")
 
         self._tests_cmake_args = {"dbg" : [("CMAKE_BUILD_TYPE", "Debug"),
-                                           ("EKAT_DEFAULT_BFB", "ON")],
+                                           ("EKAT_DEFAULT_BFB", "True")],
                                   "sp"  : [("CMAKE_BUILD_TYPE", "Debug"),
                                            ("SCREAM_DOUBLE_PRECISION", "False"),
-                                           ("EKAT_DEFAULT_BFB", "ON")],
+                                           ("EKAT_DEFAULT_BFB", "True")],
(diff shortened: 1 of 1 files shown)


bartgol:
The AT reported a fail on Weaver, despite all tests passing. I would give it another round of testing, and if it happens again, we should look at the testing scripts and the jenkins jobs, to see if something is wrong (perhaps as simple as a return code somewhere).

tcclevenger:
@bartgol How can we get more info on the weaver failure?

bartgol:
@tcclevenger It's bizarre. Build 690 today passed, but build 691 did not. They both reported 100% tests passing. Looking at the ctest logs, I can see mpi fabric initialization errors, like the following:

```
weaver5][[901,1],0][btl_openib_component.c:1699:init_one_device] error obtaining device attributes for mlx5_0 errno says Protocol not supported
```

However, both build 690 and 691 report this problem, and yet 690 passes, and 691 doesn't. I think my next best guess is to manually put some print statements in test_all_scream.py, to see a) which test is failing, b) if not all, then see if it's always the same one, and c) what's ctest return code, and see if that can tells us anything on what's wrong. I can also try to manually run ctest, in the same folders that the AT used, and see if I get a non-zero exit code.

I will work on this hopefully today, more likely tomorrow.

bartgol:
So, after manually adding some print statements, I was able to see that all builds on weaver have ctest reporting 255 as exit code. The fun thing is that ctest has exit codes 0,1,2,4,8,16,32,64 (corresponding to different things), so even if everything went wrong, they would add up to 127. Someone report an exit code of 255, but that's for the case where no tests were found (which is not our case). I'm trying to dig a bit deeper, and figure out what happens.

bartgol:
@tcclevenger See #754 for more details regarding weaver failures.

Personally, I think we can manually merge PRs (using admin powers) _if_ we see that tests pass, and the only issue is this misterious AT failure on weaver (with ctest reporting 100% pass).

tcclevenger:
@bartgol Ok, thanks for looking into this. I'll manually merge here.

PR #750, opened 2020-11-21
test-all-scream: A few minor fixes

jgfouca:
1) Restore the ability of test_all_scream to detect CUDA builds and skip
   fpe test if so.
2) Skip various not-None checks on self._machine since we now expect it to
   always be set.
3) Python strings containing `\` must be declared as raw or else this character
   will be interpreted as an escape character.

### DIFF
--- a/components/scream/scripts/test_all_scream.py
+++ b/components/scream/scripts/test_all_scream.py
@@ -3,8 +3,8 @@ from utils import run_cmd, run_cmd_no_fail, check_minimum_python_version, get_cu
     get_common_ancestor, merge_git_ref, checkout_git_ref, print_last_commit
 
 from machines_specs import get_mach_compilation_resources, get_mach_testing_resources, \
-                           get_mach_baseline_root_dir, setup_mach_env, \
-                           get_mach_cxx_compiler, get_mach_f90_compiler, get_mach_c_compiler
+    get_mach_baseline_root_dir, setup_mach_env, is_cuda_machine, \
+    get_mach_cxx_compiler, get_mach_f90_compiler, get_mach_c_compiler
 
 check_minimum_python_version(3, 4)
 
@@ -91,10 +91,9 @@ class TestAllScream(object):
                                   "fpe" : "debug_nopack_fpe"}
 
         if not self._tests:
-            is_cuda_machine = "OMPI_CXX" in os.environ
             # always do dbg and sp tests, do not do fpe test on CUDA
             self._tests = ["dbg", "sp"]
-            if not is_cuda_machine:
+            if not is_cuda_machine():
                 self._tests.append("fpe")
         else:
             for t in self._tests:
--- a/components/scream/scripts/test_all_scream.py
+++ b/components/scream/scripts/test_all_scream.py
@@ -50,6 +50,12 @@ class TestAllScream(object):
         #  Sanity checks and helper structs setup  #
         ############################################
 
+        # Unless the user claims to know what he/she is doing, we setup the env.
+        expect(self._machine is not None, "Machine is now required by test_all_scream")
+        if not self._preserve_env:
+            # Setup the env on this machine
+            setup_mach_env(self._machine)
+
         # Compute root dir
         if not self._root_dir:
             self._root_dir = pathlib.Path(__file__).resolve().parent.parent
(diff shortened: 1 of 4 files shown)


components/scream/scripts/test_all_scream.py
```diff
@@ -85,7 +91,11 @@ def __init__(self, cxx_compiler, f90_compiler, c_compiler, submit=False, paralle
                                   "fpe" : "debug_nopack_fpe"}
 
         if not self._tests:
-            self._tests = ["dbg", "sp", "fpe"]
+            is_cuda_machine = "OMPI_CXX" in os.environ
```
bartgol, line 94: I think we should not rely on OMPI_CXX. Ekat can automatically set that at config time, so one may not set it in the env. Even if now it is in the machine specs, if one adds a new GPU machine, he should not need to add OMPI_CXX to the env (since ekat auto sets that for you). Besides, OMPI_CXX is *only* good for openmpi. MPICH has a different env var (which ekat sets correctly).

Perhaps a better way to do this is to check if `nvcc` is present in the env PATH. I would assume nobody loads cuda when building a non-gpu version of scream...

components/scream/scripts/test_all_scream.py
```diff
@@ -475,7 +476,7 @@ def generate_baselines(self, test):
             return False
         else:
             # Clean up the directory, by removing everything but the 'data' subfolder
-            run_cmd_no_fail("find -maxdepth 1 -not -name data ! -path . -exec rm -rf {} \;",
+            run_cmd_no_fail(r"find -maxdepth 1 -not -name data ! -path . -exec rm -rf {} \;",
```
bartgol, line 479: There's a stranded 'r' right before the string.

components/scream/scripts/test_all_scream.py
```diff
@@ -50,6 +50,12 @@ def __init__(self, cxx_compiler, f90_compiler, c_compiler, submit=False, paralle
         #  Sanity checks and helper structs setup  #
         ############################################
 
+        # Unless the user claims to know what he/she is doing, we setup the env.
+        expect(self._machine is not None, "Machine is now required by test_all_scream")
```
bartgol, line 54: What if we defaulted the mach to `linux-generic`? It would save a bunch of keystrokes for the user who runs on a generic machine. As a debug check, we could print the mach (or maybe even more options) inside self._init, so we can always check the logs in case something goes wrong.

Also, _very minor_: maybe `test-all-scream` rather than `test_all_scream`.

bartgol (changes requested):
Thanks Jim. I have a few of comments (one typo, one suggestion, and one more important regards MPI distro portability).

components/scream/scripts/test_all_scream.py
jgfouca: > Python strings containing '' must be declared as raw or else this character will be interpreted as an escape character.

jgfouca:
@bartgol , this is ready for a re-review. I'm punting on defaulting to some generic machine since I'm pretty worried about surprising users with that behavior if they forget `-m`.

components/scream/scripts/test_all_scream.py
```diff
@@ -50,6 +50,20 @@ def __init__(self, cxx_compiler, f90_compiler, c_compiler, submit=False, paralle
...
         ############################################
 
+        # Probe machine if none was specified
+        if self._machine is None:
+            # We could potentially integrate more with CIME here to do actual
+            # nodename probing.
+            if "CIME_MACHINE" in os.environ and is_machine_supported(os.environ["CIME_MACHINE"]):
+                self._machine = os.environ["CIME_MACHINE"]
+            else:
+                expect(self._machine is not None, "Machine is now required by test-all-scream")
```
bartgol, line 60: Nit-picky here, but the condition is guaranteed to fail (given the enclosing if), so might as well put `expect(False,"...")`. Or even better use `raise`?

bartgol (approved):
I had a comment, feel free to disregard.

components/scream/scripts/test_all_scream.py
jgfouca: Done.

bartgol (approved):

PR #751, opened 2020-11-23
Port shoc solver functions

tcclevenger:
Implementation of the shoc solver in C++ using ekat_tridiag functions. 

These changes do not replace the fortran solver (i.e., no linking calls from functions in shoc.F90), only implementations of the functions in shoc_functions.hpp. The solver will be called from Fortran when I port update_prognostics_implicit next.

I added a bfb test which compares both the decomposition and solver to Fortran.

### DIFF
--- a/components/scream/src/physics/shoc/shoc_tridiag_solver_impl.hpp
+++ b/components/scream/src/physics/shoc/shoc_tridiag_solver_impl.hpp
@@ -1,0 +1,93 @@
+#ifndef SHOC_TRIDIAG_SOLVER_IMPL_HPP
+#define SHOC_TRIDIAG_SOLVER_IMPL_HPP
+
+#include "shoc_functions.hpp" // for ETI only but harmless for GPU
+#include "ekat/util/ekat_tridiag.hpp"
+
+namespace scream {
+namespace shoc {
+
+template<typename S, typename D>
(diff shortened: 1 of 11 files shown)


components/scream/src/physics/shoc/shoc_tridiag_solver_impl.hpp
```diff
@@ -0,0 +1,93 @@
...
+KOKKOS_FUNCTION
+void Functions<S,D>::vd_shoc_solve(
+  const MemberType&      team,
+  const uview_1d<Scalar>& du,
+  const uview_1d<Scalar>& dl,
+  const uview_1d<Scalar>& d,
+  const uview_2d<Spack>&  var)
+{
+#ifdef EKAT_DEFAULT_BFB
+  ekat::bfb(team, dl, d, du, var);
```
ambrad, line 79: I'd like the `tridiag` namespace to be added back to ekat, so here one would write `ekat::tridiag::bfb`, `ekat::tridiag::bfb`.

ambrad (approved):
Great!

I added a request that the tridiag namespace be added back to ekat and used here, but that can be done in a separate pair of PRs.

jeff-cohere (approved):

PR #756, opened 2020-11-24
Major update to boiler-related things

jgfouca:
Summary:

Before gen-boiler was created, I initially made a lot of effort to reduce every keystroke when creating boilerplate for SHOC porting. This led to many layers of syntactical sugar. Now that gen-boiler exists, this sugar is adding a lot of complexity for not much payoff. The new-and-improved PhysicsTestData is fully general and has a decent API that's not much more inconvenient to use than the sugar was, so I figured it was time to remove all this sugar. Unfortunately, nearly all the existing *Data structs were using this sugar and had to be regenerated en-masse. Further complicating things was that many of the names developers chose (through no fault of their own) were not 100% compliant with the names generated by gen-boiler, so I ended up blowing away and regenerating most of shoc_functions_f90.hpp and lots of shoc_functions_f90.cpp too. *Data name changes percolated to lots of test code which led to a fair bit of code churn unfortunately.

The good news is that, with these changes, our boiler is powerful, general, and decoupled from SHOC-specifics and should serve us better going into the future.

Change list:

- Gen boiler
  - Add support for fortran function syntax
  - Add support for multidimensional logicals
  - Remove support for SHOC-specific syntactic sugar
  - Improve use of whitespace when generating multiple items at once
- physics/share
  - Remove PhysicsTestData (sugar around PhysicsTestDataGeneral)
  - Rename PhysicsTestDataGeneral to PhyiscsTestData
  - Refactor PhysicsTestData to a much cleaner and more flexible implementation. If we need additional data types in the future, it should be super-easy.
  - Remove test cases that added no value other than testing branches in the sugar
- SHOC
  - shoc.F90 functions need to use btype logicals in order to be interoperable with CXX
  - full regen of shoc_functions_f90.hpp Data structs and glue decls. All *Data struct names are now compliant with gen-boiler
  - generate boiler for pblintd_height and impli_srf_stress_term to test these new changes
  - fix nastly little memory overrun error in shoc_diag_third_property
- P3
  - Change to use new PhysicsTestData


bartgol:
@jgfouca I removed myself as a reviewer, since my knowledge on the physics folder and the physics porting efforts is minimal at best. And not having ever used gen-boiler makes it hard for me to review even that part.

I'll leave it to other developers to review this PR.

jgfouca:
@bartgol , no problem. I only added you because I did some CS/software-design stuff in PhysicsTestData.

components/scream/src/physics/shoc/tests/shoc_diag_third_tests.cpp
```diff
@@ -176,7 +176,7 @@ struct UnitWrap::UnitTest<D>::TestShocDiagThird {
 
     // Load up new and increased TKE values
     for(Int s = 0; s < shcol; ++s) {
-      for(Int n = 0; n < nlevi; ++n) {
+      for(Int n = 0; n < nlev; ++n) {
```
jgfouca, line 179: The bug fix.

bartgol:
> @bartgol , no problem. I only added you because I did some CS/software-design stuff in PhysicsTestData.

jgfouca (commented):
Annotating the one bug fix.

jgfouca:
I should have been more clear about the reviews. There's a fair bit of churn here that is not interesting. I just want folks to be aware of the new capabilities and the expected code form for *Data structs in shoc_function_f90.

jeff-cohere (approved):
Looks solid. Definitely cleaner than before. Thanks for the hard work, Jim!

I didn't look closely at everything, but this seems good to me. The AT seems to like it, too.

bogensch:
sorry for not getting to this, was on vacation. but thanks!

PR #757, opened 2020-11-24
Implemented a small C++ program to search for CF-style field names.

### LINKED ISSUE #749: Implement python or small C++ program to query CF database
The script should allow one to query the database, in a few ways. The most important is to look for the CF name of a variable, given a "guess", like this:

```
  $ query-cf-database -s <name guess>
```
`-s` stands for search, but maybe `-f` for "find"?
E.g:

```
$ query-cf-database -s "surface_temperature"

   Input name: 'surface temperature'
   Best matches:
      surface_temperature
      sea_surface_temperature
      sea_ice_surface_temperature
      solid_earth_subsurface_temperature
      square_of_sea_surface_temperature
```
The script should report all matches with a similarity index above a certain threshold. This threshold should be configurable (to allow user to search "worse" matches if he/she cannot provide a decent guess), and should have a default determined by trial and error. I ran the Jaro and Jaro-Winkler similarity functions in EKAT on a "good" guess, and got this:
```
enter 1st string: evaporation water flux
enter 2nd string: water_evaporation_flux
similarity indices:
 j : 0.786284
 jw: 0.786284
```
but also
```
enter 1st string:temperature
enter 2nd string:pressure
similarity indices:
 j : 0.626263
 jw: 0.626263
```
so maybe we should pick something between 0.6 and 0.8 (if using J/JW indices).

Possible flags for the script:
- "-f" or "-t" <string>: search for var names similar to <string>.
- "-d": get description for a given variable; require exact var name, this would not do similarity search. It might help in case one is not 100% sure what the variable is about.
If using `-f` (or `-s`, or whatever it is), one could also specify:
- "-t <float number>": threshold for the index (between 0 and 1). As I said, we should default this to a value somewhere in 0.6-0.8
- "-n <int number>": max number of items to report. Default maybe to 5, or something not too big.
- "-b": ignore threshold, and report list of best matches (limited by -n option), regardless of how bad/good the matches are
- "-v": verbose mode, to show index for each match (could help user "learn" good values for future use).

Note: there are J/JW similarity index computation routines in `ekat_string_utils.*pp`. They are relatively small, so one should not have a hard time rewriting them in python. If you do, and want to test your mini script, in `tests/util/utils_cxx` in EKAT, there is a small dictionary of strings, with their expected J/JW similarity indices.

Note 2: it might be that J/JW indices are not great for our purposes. The fact that `evaporation water flux` and `water_evaporation_flux` have an index of 0.78 while `pressure` and `temperature` have 0.62 makes me think these indices may not be suitable for "sentences", where words are never misspelled, but might be rearranged.

jeff-cohere:
This PR introduces `query-cf-database`, a C++ program that lives in the now-not-quite-appropriately-named `scripts/` directory, and which searches two YAML files that store our standard and SCREAM-specific CF field names for matches to a supplied field name.

It also fixes some issues with the `cf-xml-to-yaml` script, and introduces a single phony SCREAM field to a `cf-scream-name-table.yaml` file in SCREAM's `data/` directory.

I'd understand if we wanted to rename the `scripts/` directory to something like `tools/`. :-)

Anyway, the options are a bit different but equivalent to what Luca suggested in #749. You can invoke the program (which gets built in the `scripts/` subdirectory of your build directory) with no arguments to get usage information.

Some interesting experiments to try (from the top of your `build/` directory):

```
./scripts/query-cf-database -v aerodynamic
./scripts/query-cf-database -v temperature
```

Closes #749


jeff-cohere:
Had to merge from `master` to resolve some conflicts in `ekat`, apparently. Sorry about that.

components/scream/scripts/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+# Generate the source file for the CF validator and build it.
+configure_file(
+  ${CMAKE_CURRENT_SOURCE_DIR}/query-cf-database.cpp.in
+  ${CMAKE_CURRENT_BINARY_DIR}/query-cf-database.cpp
+  @ONLY
+)
+add_executable(query-cf-database ${CMAKE_CURRENT_BINARY_DIR}/query-cf-database.cpp)
+target_link_libraries(query-cf-database ekat;${YAMLCPP_LIBARIES})
```
bartgol, line 8: Ekat already links against yaml, so there's no need to add YAMLCPP_LIBRARIES.

bartgol (approved):
Nice! I had one comment which is probably not a big deal anyways.

I noticed that the ekat version is changed. Usually, we do a separate PR for submodules update, for ease of tracking. But I'm not sure if it ever became a rule or we simply happened to do it that way. For what it's worth, given the small amount of changes in ekat and the fact that they are strictly linked to this PR, I"m fine with doing the submodule update as part of the PR.

jeff-cohere:
Okay. I'll try to keep EKAT updates separate from here on out. And I'll fix that one library link.

bartgol:
@jeff-cohere Have you thought about doing the whole script in python? It would not use the ekat function, but the impl of jaccard/jaro is quite easy we can duplicate the code. Having a python script would allow one to run this tool straight from the src folder, without the need to have a bld dir.

It's just a thought, feel free to push back.

jeff-cohere:
Yeah, a Python version would be pretty simple at this point. The biggest reason I didn't do it was that I didn't want to introduce the need to install Python modules, like for YAML (and/or Jaro and Jaccard indices, though as you say they're easy to implement).

But installing Python modules is pretty easy, too. :-)

bartgol (approved):

PR #761, opened 2020-11-25
dyamond2 specific: Fix bug where code looks for clm2.r file rather than elm.r for branch runs

crterai:
dyamond2 specific: Currently, branch runs will look for clm2.r files for branch or hybrid runs. This quick bug fix makes it so that it looks for elm.r files. 
The issue has been identified in e3sm (https://github.com/E3SM-Project/E3SM/issues/3905), so better for master to wait for an upstream merge.


PeterCaldwell (approved):
Looks pretty straightforward...

PR #762, opened 2020-11-25
porting pblintd_init function

xyuan:
porting pblintd_init function

[BFB]


jeff-cohere (approved):
Looks good.

components/scream/src/physics/shoc/shoc_functions_f90.cpp
```diff
@@ -2417,6 +2427,57 @@ void pblintd_height_f(Int shcol, Int nlev, Real* z, Real* u, Real* v, Real* usta
...
+    Scalar& z_s           = z_1d(nlev_pack)[nlev_indx];
+
+    SHOC::pblintd_init(z_s, check_s, rino_s, pblh_s);
+
+    check_d(i)   = check_s;
+    pblh_d(i)[0] = pblh_s;
+  });
+
+  Kokkos::Array<view_2d, 1> out_2d_views = {rino_2d};
+  ekat::device_to_host<int, 1>({rino}, {shcol}, {nlev}, out_2d_views, true);
```
jgfouca, line 2470: You don't need to use the API that takes vectors of sizes if they are all the same.

components/scream/src/physics/shoc/shoc_functions_f90.cpp
```diff
@@ -2417,6 +2427,57 @@ void pblintd_height_f(Int shcol, Int nlev, Real* z, Real* u, Real* v, Real* usta
...
+
+  Kokkos::Array<view_1d, 1> out_1d_views = {pblh_d};
+  ekat::device_to_host<int, 1>({pblh}, {shcol}, out_1d_views);
```
jgfouca, line 2473: Same as above.

components/scream/src/physics/shoc/shoc_functions_f90.cpp
```diff
@@ -2417,6 +2427,57 @@ void pblintd_height_f(Int shcol, Int nlev, Real* z, Real* u, Real* v, Real* usta
...
+
+  Kokkos::deep_copy(host_check_d, check_d);
```
jgfouca, line 2475: device_to_host now works for bools so you should use that instead. In fact, I think it has always works for 1d data.

jgfouca (changes requested):
Minor changes needed to _f.

components/scream/src/physics/shoc/shoc_functions_f90.cpp
jgfouca: I was wrong, device_to_host expects packed data.

jgfouca (approved):

PR #767, opened 2020-11-30
Adds some bfb function in shoc.F90

singhbalwinder:
Adds some required BFB functions in shoc.F90 to help with porting.

### DIFF
--- a/components/eam/src/physics/cam/shoc.F90
+++ b/components/eam/src/physics/cam/shoc.F90
@@ -2982,7 +2982,7 @@ subroutine compute_shr_prod(nlevi, nlev, shcol, dz_zi, u_wind, v_wind, sterm)
         ! calculate vertical gradient of u&v wind
         u_grad = grid_dz*(u_wind(i,km1)-u_wind(i,k))
         v_grad = grid_dz*(v_wind(i,km1)-v_wind(i,k))
-        sterm(i,k) = u_grad**2+v_grad**2
+        sterm(i,k) = bfb_square(u_grad)+bfb_square(v_grad)
      enddo
   enddo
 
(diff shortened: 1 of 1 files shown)


tcclevenger:
@singhbalwinder We expect fails here. My guess is from the `bfb_pow(tke(i,k),1.5_rtype)`call (`bfb_square` tends to be bfb). Let me check to make sue the diff is what we expect and then I will manually merge.

components/eam/src/physics/cam/shoc.F90
```diff
@@ -3054,7 +3054,7 @@ subroutine adv_sgs_tke(nlev, shcol, dtime, shoc_mix, wthv_sec, &
         a_prod_sh=tk(i,k)*sterm_zt(i,k)
 
         ! Dissipation term
-        a_diss(i,k)=Cee/shoc_mix(i,k)*tke(i,k)**1.5
+        a_diss(i,k)=Cee/shoc_mix(i,k)*bfb_pow(tke(i,k),1.5_rtype)
```
ambrad, line 3057: Here I recommend `bfb_sqrt(bfb_cube(tke(i,k)))`, but I've never actually demonstrated it should be preferred, so it's just a suggestion.

singhbalwinder: I can make this change if it is better than using `pow`. For C++ code, I think we have `ekat::sqrt` and `ekat::cube` which we can use to mimic this calculation.

ambrad: I did some investigating and have firmer evidence that `bfb_sqrt(bfb_cube(tke(i,k)))` is the way to go:

- I wrote a short program and looked at the assembly with `-xcore-avx512 -O3 -fp-model $model -fsource-asm -S -c`, where `$model` is either `fast` or `strict`. In the `fast` case, ifort on Compy ends up emitting code for the sqrt(x^3) formulation instead of pow, supporting that it's faster.
- Then I ran a bunch of trials with `strict` on random numbers drawn from U[0,1] and found that it's also slightly more accurate. I did this by computing these expressions in single precision and comparing with a double-precision truth value. But only slightly: The pow formulation is worse 10.3% of the time, the sqrt 7.2%, and it's a tie the rest. But slightly better and faster makes sqrt(x^3) the way to go.

tcclevenger: My apologies for being quick on the merge..

PR #768, opened 2020-11-30
Port update_prognostics_implicit

tcclevenger:
Ports `update_prognostics_implicit` subroutine to C++. The `update_prognostics_implicit` subroutine is the one that calls the tridiag solver.

### DIFF
--- a/components/scream/src/physics/shoc/shoc_update_prognostics_implicit_impl.hpp
+++ b/components/scream/src/physics/shoc/shoc_update_prognostics_implicit_impl.hpp
@@ -70,25 +70,25 @@ void Functions<S,D>::update_prognostics_implicit(
   // and tke flux calc (wtke_sfc)
   Scalar ksrf, wtke_sfc;
   {
-    const auto wsmin = 1;
-    const auto ksrfmin = 1e-4;
-    const auto ustarmin = 0.01;
+    const Scalar wsmin = 1;
+    const Scalar ksrfmin = 1e-4;
+    const Scalar ustarmin = 0.01;
 
-    const auto rho = rho_zi(last_nlevi_pack)[last_nlevi_indx];
-    const auto uw = uw_sfc;
-    const auto vw = vw_sfc;
+    const Scalar rho = rho_zi(last_nlevi_pack)[last_nlevi_indx];
+    const Scalar uw = uw_sfc;
+    const Scalar vw = vw_sfc;
 
-    const auto taux = rho*uw;
-    const auto tauy = rho*vw;
+    const Scalar taux = rho*uw;
+    const Scalar tauy = rho*vw;
 
-    const auto u_wind_sfc = u_wind(last_nlev_pack)[last_nlev_indx];
-    const auto v_wind_sfc = v_wind(last_nlev_pack)[last_nlev_indx];
+    const Scalar u_wind_sfc = u_wind(last_nlev_pack)[last_nlev_indx];
+    const Scalar v_wind_sfc = v_wind(last_nlev_pack)[last_nlev_indx];
 
-    const auto ws = ekat::impl::max(std::sqrt((u_wind_sfc*u_wind_sfc) + v_wind_sfc*v_wind_sfc), sp(wsmin));
-    const auto tau = std::sqrt(taux*taux + tauy*tauy);
-    ksrf = ekat::impl::max(tau/ws, sp(ksrfmin));
+    const Scalar ws = ekat::impl::max(std::sqrt((u_wind_sfc*u_wind_sfc) + v_wind_sfc*v_wind_sfc), wsmin);
+    const Scalar tau = std::sqrt(taux*taux + tauy*tauy);
+    ksrf = ekat::impl::max(tau/ws, ksrfmin);
 
-    const auto ustar = ekat::impl::max(std::sqrt(std::sqrt(uw*uw + vw*vw)), sp(ustarmin));
+    const Scalar ustar = ekat::impl::max(std::sqrt(std::sqrt(uw*uw + vw*vw)), ustarmin);
     wtke_sfc = ustar*ustar*ustar;
   }
 
--- a/components/scream/src/physics/shoc/shoc_update_prognostics_implicit_impl.hpp
+++ b/components/scream/src/physics/shoc/shoc_update_prognostics_implicit_impl.hpp
@@ -13,11 +13,156 @@ namespace shoc {
 
 template<typename S, typename D>
 KOKKOS_FUNCTION
-KOKKOS_FUNCTION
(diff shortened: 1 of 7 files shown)


jeff-cohere (approved):

components/scream/src/physics/shoc/tests/shoc_update_prognostics_implicit_tests.cpp
```diff
@@ -372,19 +381,14 @@ struct UnitWrap::UnitTest<D>::TestUpdatePrognosticsImplicit {
       UpdatePrognosticsImplicitData& d_f90 = f90_data[i];
       UpdatePrognosticsImplicitData& d_cxx = cxx_data[i];
       for (Int k = 0; k < d_f90.total(d_f90.thetal); ++k) {
-        REQUIRE(d_f90.total(d_f90.thetal) == d_cxx.total(d_cxx.thetal));
         REQUIRE(d_f90.thetal[k] == d_cxx.thetal[k]);
```
jgfouca, line 384: Why did you delete these `total` checks?

components/scream/src/physics/shoc/shoc_update_prognostics_implicit_impl.hpp
```diff
@@ -13,14 +13,159 @@ namespace shoc {
...
+  // compute 1/dp term, needed in diffusion solver
+  dp_inverse(team, nlev, rho_zt, dz_zt, rdp_zt);
+  team.team_barrier();
+
+  // compute terms needed for the implicit surface stress (ksrf)
+  // and tke flux calc (wtke_sfc)
+  Scalar ksrf, wtke_sfc;
+  {
+    const auto wsmin = 1;
+    const auto ksrfmin = 1e-4;
```
jgfouca, line 74: If you used `const Real XXX = literal;` instead of `const auto` some of the `sp()` calls below could be removed.

jgfouca (changes requested):
Looks good, just some minor questions.

components/scream/src/physics/shoc/shoc_update_prognostics_implicit_impl.hpp
```diff
@@ -13,14 +13,159 @@ namespace shoc {
...
+
+    const auto u_wind_sfc = u_wind(last_nlev_pack)[last_nlev_indx];
+    const auto v_wind_sfc = v_wind(last_nlev_pack)[last_nlev_indx];
+
+    const auto ws = ekat::impl::max(std::sqrt((u_wind_sfc*u_wind_sfc) + v_wind_sfc*v_wind_sfc), sp(wsmin));
+    const auto tau = std::sqrt(taux*taux + tauy*tauy);
+    ksrf = ekat::impl::max(tau/ws, sp(ksrfmin));
+
+    const auto ustar = ekat::impl::max(std::sqrt(std::sqrt(uw*uw + vw*vw)), sp(ustarmin));
+    wtke_sfc = ustar*ustar*ustar;
```
ambrad, line 92: These two lines could be optimized slightly in C++ and F90 since `ustar*ustar` undoes the second sqrt above. @bogensch just a quick check: is the sqrt(sqrt(...)) above correct, i.e., there should really be two sqrts?

components/scream/src/physics/shoc/shoc_update_prognostics_implicit_impl.hpp
```diff
@@ -13,14 +13,159 @@ namespace shoc {
...
+    const auto k_pack_indx = k%Spack::n;
+
+    X1(k,0)[0] = u_wind(k_view_indx)[k_pack_indx];
+    X1(k,1/Spack::n)[1%Spack::n] = v_wind(k_view_indx)[k_pack_indx];
+
+    X2(k,0)[0] = thetal(k_view_indx)[k_pack_indx];
+    X2(k,1/Spack::n)[1%Spack::n] = qw(k_view_indx)[k_pack_indx];
+    X2(k,2/Spack::n)[2%Spack::n] = tke(k_view_indx)[k_pack_indx];
+    for (Int p=0; p<num_tracer; ++p) {
+      X2(k,(3+p)/Spack::n)[(3+p)%Spack::n] = tracer(k, p/Spack::n)[p%Spack::n];
```
ambrad, line 124: Here you load the tracer array into another array. This is one of ~3 approaches to the problem, and I'll argue roughly the worst of the 3. Here are the approaches:
1. Tracers come in at the highest level with fast index k. Load them into a transposed tracer array for SHOC, with an extra 3 slots for other RHS. Use that array as the RHS; here, there would be no copy into another array. That's the approach I would take b/c it would save copying/transposes in the case of multiple SHOC iterations. Small win b/c we're expecting only one iteration in practice.
2. Tracers come in at the highest level with fast index k. Do nothing at the high level. Here, do a transpose at the same time as you copy. An acceptable alternative to 1.
3. Tracers come in at the highest level with fast index k. Load them into a transposed tracer array for SHOC. Then, here, copy to another array. This has needless copying. This is what's being done here.
So I'd like you to redesign this part to use approaches 1 or 2 or anything else you think of that's at least as good as either of those.

I really like how you've reduced this code to exactly two solves, the minimum number possible.

Separate point: Note here and in any tracer transpose code you end up writing that `scalarize`ing the standard arrays before this ||4 would save all the pack indexing arithmetic. You'd just write `X1(k,1/Spack::n)[1%Spack::n] = v_wind[k]`, for example.

components/scream/src/physics/shoc/tests/shoc_update_prognostics_implicit_tests.cpp
tcclevenger: Ah, I meant to pull them out of the for-loop (since they do not depend on k) but forgot to paste above.. They're back now.

ambrad (changes requested):

components/scream/src/physics/shoc/shoc_update_prognostics_implicit_impl.hpp
bogensch: Yes, that is correct.

jgfouca (approved):

tcclevenger:
@ambrad I agree with you and chose method 1). I resize the `tracer` view at the highest level to include 3 extra slots (using `Kokkos::resize(Kokkos::WithoutInitializing, tracer,...)`), then for the 2nd solve I store `thetal`, `qw`, and `tke` into `tracer` and solve. So no more copying the tracer values to a new view.

I did, however, keep the view `X1` to store `u_wind` and `v_wind` for the 1st solve.  A possible solution to having no additional views would be to pack `u_wind` and `v_wind` into a 3d view (named `wind`) at the highest level? Seems to me that this still requires copying the same amount of values, so I did not implement it. But maybe it could be advantageous.

Let me know if you have any more comments/suggestions. 

A further optimization (and maybe what you intended in your comment) is that both RHS views (the 1st solve with `u_wind/v_wind` and the 2nd solve with `tracer/thetal/qw/tke`) could all be packed together at the highest level (still in 2 views since the 1st solve would overwrite the values needed for the 2nd solve if they were all packed together into 1 view). A single scalar entry is only ever needed from `u_wind/v_wind/thetal/qw/tke` throughout this function so there would not be the need to pack and unpack values. Then, only at the highest level would there be a copy of values.

ambrad:
Looks great. You suggest a number of good ideas you could do now or in the future. Another is to put the 3 slots at the start of the X2 array, then use these instead of X1. Pack, decomp, solve, and unpack with just the first two columns of X2. Then do that again for 3 + tracers.

However, feel free to put anything like this off for future performance PRs. My primary interest in this PR was to establish the tracer memory pattern.

ambrad (approved):
Looks great!

tcclevenger:
@ambrad Ok great! I'll let this PR merge and think about improvements in future PRs.

PR #769, opened 2020-12-01
FieldInitializer modifications

bartgol:
This PR ensures that a FieldInitializer is able to init fields both on the grid that they are defined on, as well as on the reference grid (if those are the same, everything behaves as it did before).

This upgrade is necessary. Recall that the AD atm dag always starts from the ref grid, so the atm inputs are _all_ on the reference grid, and remappers are used to map fields onto different grids. So if an initializer inits fields only on a non-ref grid, we have 2 issues:
- the DAG check would fail (nobody is init-ing the corresponding field on the ref grid);
- even disabling the DAG checks, the remapper would overwrite the field on the non-ref grid (the one that the initializer inited) with whatever was in the corresponding ref-grid field (zeros or garbage, depending on how the view was inited by kokkos).

The solution in this PR is simple: if a field initializer inits F on a non-ref grid, when it's time to set the field in the initializer, we pass both F and F_ref (the copy on the ref grid) _and_ a remapper from the grid to the ref grid. If the initializer does not know how to handle the ref grid, it can simply init F, then use the remapper to init F_ref. Note: some initializers (like FieldValueInitializer) can init F_ref directly, without need of a remapper.

### DIFF
--- a/components/scream/src/physics/zm/zm_inputs_initializer.cpp
+++ b/components/scream/src/physics/zm/zm_inputs_initializer.cpp
@@ -47,8 +47,8 @@ struct GridOpts{
 unordered_map<string, GridOpts> opt_map;
 
 //Initializes struct GridOpts fields
-void set_grid_opts_helper(GridOpts O, string n, bool out, const Units* unit, int field_idx
-                          ){
+void set_grid_opts_helper(GridOpts O, string n, bool out, const Units* unit, int field_idx)
+{
   O.name = n;
   O.isOut = out;
   O.field_idx = field_idx;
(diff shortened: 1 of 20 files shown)


components/scream/src/physics/zm/zm_inputs_initializer.cpp
```diff
@@ -198,13 +197,38 @@ void ZMInputsInitializer::add_field (const field_type &f)
   m_fields_id.insert(id);
 }
 
+void ZMInputsInitializer::
```
AaronDonahue, line 200: Presumably we would need to create something like this for SHOC and RAD right?  Would it be simpler to make the Initializer a general class or object so that if we change the code we don't have to make the same change in 4+ places?

AaronDonahue (approved):
Looks good, just one question about organization of the process initializer.

components/scream/src/physics/zm/zm_inputs_initializer.cpp
bartgol: I'm not sure what you mean. There is already a FieldInitializer class, so that the AD can interact in a process-agnostic way.

PR #775, opened 2020-12-01
SCREAM: add options to test-all-scream for quick testing

bartgol:
This small upgrade to test-all-scream might help people to speed up development cycles while still relying on test-all-scream.

Currently, if one does a small modification to the source code and runs test-all-scream again, all the libraries/execs have to be rebuilt from scratch. This is the default behavior, which protects from dirty build dirs, which sometimes can cause things to work even if they shouldn't. However, for normal development, we don't want to rebuild the whole project, and incremental builds can save time.

I personally manually build scream outside the src tree, mostly because of test-all-scream is slow (for what said above). This new option makes test-all-scream more amenable for my needs.

The two options are `--quick-rerun` and `--quick-rerun-failed`. They both skip the config step in ctest, and simply build and test the project, but starting from the current status of the build dir (hence, only rebuilding what changed). The former reruns all tests, while the latter only reruns the tests that failed.

Note:
- if all tests passed in all the previous runs (meaning, there is no LastTestsFailed_XYZ file in the Testing/Temporary subdir of the build dir), then `--quick-rerun-failed` is the same as `--quick-rerun` (this follows from how ctest handles the `--rerun-failed`.
- if testing against a baseline commit, baselines will still be regenerated. To unleash the full potential of this option, provide a baseline dir, so that only the test phase is executed. Also, you can use this together with `-k`, which allows to test a dirty repo (with uncommited changes).

### DIFF
--- a/components/scream/scripts/test_all_scream.py
+++ b/components/scream/scripts/test_all_scream.py
@@ -563,7 +563,7 @@ class TestAllScream(object):
             # Note: make will still rerun cmake if some cmake file has changed
             ctest_config += "-DSKIP_CONFIG_STEP=TRUE "
             if self._quick_rerun_failed:
-              ctest_config += "--rerun-failed "
+                ctest_config += "--rerun-failed "
         else:
             # This directory might have been used also to build the model to generate baselines.
             # Although it's ok to build in the same dir, we MUST make sure to erase cmake's cache
@@ -588,8 +588,8 @@ class TestAllScream(object):
             # Create this test's build dir
             if test_dir.exists():
                 if not self._quick_rerun:
-                  shutil.rmtree(str(test_dir))
-                  test_dir.mkdir(parents=True)
+                    shutil.rmtree(str(test_dir))
+                    test_dir.mkdir(parents=True)
             else:
                 test_dir.mkdir(parents=True)
 
--- a/components/scream/scripts/test_all_scream.py
+++ b/components/scream/scripts/test_all_scream.py
@@ -21,7 +21,8 @@ class TestAllScream(object):
     def __init__(self, cxx_compiler, f90_compiler, c_compiler, submit=False, parallel=False, fast_fail=False,
                  baseline_ref=None, baseline_dir=None, machine=None, no_tests=False, keep_tree=False,
                  custom_cmake_opts=(), custom_env_vars=(), preserve_env=False, tests=(),
-                 integration_test="JENKINS_HOME" in os.environ, root_dir=None, work_dir=None, dry_run=False,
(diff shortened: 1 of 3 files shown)


bartgol:
Note: I discovered what I think is a bug in CTest, which causes CTest to not correctly pick up the most recent LastTestsFailed_XYZ.log file. This can cause ctest to run a wrong set of failed tests, namely, the one from the first LastTestsFailed* file that it process (I'm not sure how the files are ordered in the internal structures), rather than the one with the most recent modification time stamp. However, if you only have one LastTestsFailed* file in the folder (or if they all list the same tests), then the `--quick-rerun-failed` option should work as expected.

components/scream/scripts/test_all_scream.py
```diff
@@ -552,10 +558,17 @@ def run_test(self, test):
...
-        # This directory might have been used also to build the model to generate baselines.
-        # Although it's ok to build in the same dir, we MUST make sure to erase cmake's cache
-        # and internal files from the previous build (CMakeCache.txt and CMakeFiles folder)
-        run_cmd_no_fail("rm -rf CMake*", from_dir=test_dir)
+        if self._quick_rerun:
+            # Do not purge bld dir, and do not rerun config step.
+            # Note: make will still rerun cmake if some cmake file has changed
+            ctest_config += "-DSKIP_CONFIG_STEP=TRUE "
+            if self._quick_rerun_failed:
+              ctest_config += "--rerun-failed "
```
jgfouca, line 566: Mixing 2 and 4-space indents. We usually use 4 for python.

components/scream/scripts/test_all_scream.py
```diff
@@ -574,9 +587,11 @@ def run_all_tests(self):
 
             # Create this test's build dir
             if test_dir.exists():
-                shutil.rmtree(str(test_dir))
-
-            test_dir.mkdir(parents=True)
+                if not self._quick_rerun:
+                  shutil.rmtree(str(test_dir))
```
jgfouca, line 591: Mixing 2 and 4-space indents. We usually use 4 for python.

jgfouca (changes requested):
Looks real nice, just a couple minor cleanups. Did you run pylint on this? It would have caught the bad indents.

components/scream/scripts/test_all_scream.py
bartgol: Ah, good catch. I tried to set all indent to 4 (even though my vim uses 2), but I missed this. Thanks.

Ps: I didn't know about pylint. I see that sems python modules don't make it available. Is it a separate module/package?

jgfouca (approved):

PR #781, opened 2020-12-04
Adding some code comments to document AbstractRemapper interface methods

jeff-cohere:
These aren't comprehensive, but they do describe some important operations for registering/binding fields.

No code changes--only comments added.

### DIFF
--- a/components/scream/src/share/grid/remap/abstract_remapper.hpp
+++ b/components/scream/src/share/grid/remap/abstract_remapper.hpp
@@ -42,15 +42,6 @@ public:
   // Call this before you begin registering fields with this remapper.
   void registration_begins ();
 
-  // Specify that all given source fields be packed into the single "packed"
-  // target field identified by packed_tgt. The target field must have a layout
-  // with a sufficient rank to accommodate the fields being packed into it. In
-  // the inverse mapping, the target (source) fields are unpacked from the
-  // source (target) field in the same order in which they were packed.
-  // This method must be called before any other field registration method.
-  void register_packing (const std::vector<field_type>& src_fields,
-                         const identifier_type& packed_tgt);
-
   // This method registers a source field to be remapped to a target field.
   void register_field (const field_type& src, const field_type& tgt);
 
@@ -316,15 +307,6 @@ unregister_field (const identifier_type& src, const identifier_type& tgt) {
     --m_num_bound_fields;
   }
   m_fields_are_bound.erase(m_fields_are_bound.begin()+ifield);
-}
-
-template<typename RealType>
-void AbstractRemapper<RealType>::
-register_packing (const std::vector<field_type>& src_fields,
-                  const identifier_type& packed_tgt) {
-  EKAT_REQUIRE_MSG(m_state==RepoState::Open,
-                   "Error! You can only register fields to be packed during "
-                   "the registration phase.\n");
 }
 
 template<typename RealType>
--- a/components/scream/src/share/grid/remap/abstract_remapper.hpp
+++ b/components/scream/src/share/grid/remap/abstract_remapper.hpp
@@ -189,9 +189,10 @@ protected:
   virtual void do_remap_bwd () const = 0;
 
   // This helper function searches for a pair of source and target field
-  // identifiers in the remapper's set of bound fields, and returns the index to
-  // which they are bound, or -1 if no such pair of bound fields exists within
-  // the remapper.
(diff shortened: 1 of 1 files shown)


components/scream/src/share/grid/remap/abstract_remapper.hpp
```diff
@@ -143,15 +161,37 @@ class AbstractRemapper
...
+  // the protected data members of this class.
   virtual void do_remap_fwd () const = 0;
+
+  // Override this method to implement the backward/inverse remapping process
+  // using the protected data members of this class.
   virtual void do_remap_bwd () const = 0;
 
+  // This helper function searches for a pair of source and target field
+  // identifiers in the remapper's set of bound fields, and returns the index to
+  // which they are bound, or -1 if no such pair of bound fields exists within
```
bartgol, line 193: This function returns -1 if fields were not registered at all (via the field identifier). If they were registered, it will return the idx, even if they were not bound yet.

components/scream/src/share/grid/remap/abstract_remapper.hpp
```diff
@@ -178,9 +218,10 @@ class AbstractRemapper
...
   int           m_num_registered_fields;
 
-  // Whether fields have been provided. Recall that the user may register
-  // fields passing only an identifier, and actually bind fields only
-  // at a later moment.
+  // This vector stores a boolean whose elements are the indices of fields that
+  // have been bound to the remapper. This vector is necessary because the
+  // binding of fields is separate from their regіstration: recall that one may
+  // register a field using its identifier, and bind the actual field later.
   std::vector<bool>   m_fields_are_bound;
```
bartgol, line 225: std::vector<bool> is a weird beast. I am not sure if we should keep it as is, or replace with vector<int>. As of now, all use cases of it do things correctly, but there's a risk someone might use this in derived classes, and do something dangerous like `bool* b = &m_fields_are_bound[i];`.

Perhaps we could add a WARNING, telling devs that they should be aware of vector<bool> peculiar behavior in memory.

components/scream/src/share/grid/remap/abstract_remapper.hpp
```diff
@@ -178,9 +218,10 @@ class AbstractRemapper
   int           m_num_fields;
   int           m_num_registered_fields;
 
-  // Whether fields have been provided. Recall that the user may register
-  // fields passing only an identifier, and actually bind fields only
-  // at a later moment.
+  // This vector stores a boolean whose elements are the indices of fields that
```
bartgol, line 221: "This vector stores a boolean whose elements..." seems confusing. Maybe "A vector of booleans, whose elements ..."? I don't know, it sounds like 'whose' refer to the boolean(s) rather than the vector.

jeff-cohere: Good catch. I obviously did some mid-sentence editing that I didn't clean up. :-)

components/scream/src/share/grid/remap/abstract_remapper.hpp
jeff-cohere: This is one of those areas where C++'s complexity makes all sorts of nastiness possible. I'll add a comment about it.

components/scream/src/share/grid/remap/abstract_remapper.hpp
```diff
@@ -218,10 +219,15 @@ class AbstractRemapper
...
-  // have been bound to the remapper. This vector is necessary because the
-  // binding of fields is separate from their regіstration: recall that one may
-  // register a field using its identifier, and bind the actual field later.
+  // This vector maps the indices of registered fields to booleans that indicate
+  // whether these fields have been bound to the remapper. This vector is
+  // necessary because the binding of fields is separate from their
+  // regіstration: recall that one may register a field using its identifier,
+  // and bind the actual field later.
+  // NOTE: vector<bool> is a strange beast, and doesn't necessary behave as
+  // NOTE: expected. Use caution when manipulating this member, and don't
```
bartgol, line 228: behave 'as' expected.

bartgol: And is 'NOTE:' on all lines intended?

jeff-cohere: I did intend that with NOTE:, but have changed. `as` was already there, but I changed `necessary` to `necessarily`.

bartgol (approved):

PR #782, opened 2020-12-04
Allow AtmosphereProcess to expose dependency on field groups

bartgol:
An AtmosphereProcess (AP) can expose dependency on groups of fields, identified by a group name and a grid name. The dependency can be as 'required' group or 'updated' group. There is no 'computed' group, cause if an AP computes a bunch of fields out of other stuff, it should also know what they are. An 'updated' or 'required' group, on the other hand, is something that the AP can expect other processes to formally set up.

This PR required some changes to the AD flow and the remapper setup inside the AP group. Nothing too drastic, but some operations had to be rearranged and/or postponed, to allow for group registration to happen correctly.

The PR also include a bug fix on the FieldAllocProperties (see first commit).


components/scream/src/control/atmosphere_driver.cpp
```diff
@@ -181,14 +186,64 @@ void AtmosphereDriver::finalize ( /* inputs? */ ) {
...
+          auto f_units = fid.get_units();
+          auto src_layout = fid.get_layout();
+          auto tgt_layout = r->create_tgt_layout(src_layout);
+          FieldIdentifier new_fid(name,tgt_layout,f_units,grid);
+          m_field_repo->register_field(new_fid,group);
+        }
+      }
+    }
+  };
+  lambda( m_atm_process_group->get_required_groups() );
```
jeff-cohere, line 235: Any danger that this data might be needed before these lambdas finish? I'm guessing not, but the pedant in me has to ask.

components/scream/src/share/atm_process/atmosphere_process.hpp
```diff
@@ -103,25 +116,36 @@ class AtmosphereProcess : public ekat::enable_shared_from_this<AtmosphereProcess
...
+    ekat::error::runtime_check(
+        computes(f.get_header().get_identifier()),
+        "Error! This atmosphere process does not compute this field. "
+        "Something is wrong up the call stack. Please, contact developers.\n");
     set_computed_field_impl (f);
   }
 
+  virtual void set_required_group (const ci_string_pair& /* group_and_grid */,
+                                   const std::set<Field<const Real>>& /* group */) {
+    ekat::error::runtime_abort("Error! This atmosphere process does not require a group of fields.\n");
```
jeff-cohere, line 142: Maybe it would be good to explain that this is a constraint on process groups (if I've understood correctly)?

```
"Error! An atmosphere process group cannot require a group of fields.\n"
```

jeff-cohere: Same thing for `set_updated_group` below.

components/scream/src/share/field/field_repository.hpp
```diff
@@ -240,20 +244,48 @@ FieldRepository<RealType>::get_field (const std::string name, const std::string
...
-  EKAT_REQUIRE_MSG(f_matches.size()!=0, "Error! get_field: " + name + " not found on grid " + grid + ".\n");
   EKAT_REQUIRE_MSG(f_matches.size()==1, "Error! get_field: " + name + " found " + std::to_string(f_matches.size()) + " matches on grid " + grid + ".\n");
   // Use this field id to grab field itself.
   return get_field(f_matches[0]);
 }
 
+template<typename RealType>
+std::set<typename FieldRepository<RealType>::field_type>
+FieldRepository<RealType>::
+get_field_group (const std::string& group_name, const std::string& grid_name) const {
```
jeff-cohere, line 259: Does this mean that each field group contains fields that are all defined on a single grid?

jeff-cohere (approved):
This looks good. I left a few minor comments. I think my PR probably caused a merge conflict--sorry about that.

components/scream/src/control/atmosphere_driver.cpp
bartgol: You mean danger that data is erased before lambda finishes? I don't think so. I just used a lambda cause I didn't want to create a method for the AD. I'm lazy.

Do you think it's best to add a method?

components/scream/src/share/atm_process/atmosphere_process.hpp
bartgol: I just thought I'd add an empty default impl, so that all the atm proc (P3, SHOC, ...) do not need to have an empty impl of their methods. I can make the error message clearer, since it's a bit misleading. The issue is that if an atm proc exposes a non-trivial `get_required_groups` (i.e., returns a non-empty std::set), then it should also override this method to provide an impl of this method.

I'll update this.

components/scream/src/share/field/field_repository.hpp
bartgol: Well, if you declare that field F belongs to group G, then all "copies" of F (on all the different grids) that are registered in the repo will have the group G. However, when requesting groups as in or in/out variables, for each group you need to specify the grid you need it on. If group G is needed on grid "foo" and "bar", then one needs to ask for  two group-grid pairs: (G,foo) and (G,bar).

I can add better comments to the field repo class for this.

components/scream/src/control/atmosphere_driver.cpp
jeff-cohere: No, I don't think it matters.

jeff-cohere: If we want to be ultra paranoid about inconsistency, we can always add a barrier after the lambda. But I think it's unlikely that we need to worry about this.

bartgol: There should be no threading occurring in the lambda body, so yeah, this should be safe.

jeff-cohere (approved):
:+1:

jeff-cohere:
Should I wait for the AT to report? I assume you found and fixed a DAG-related issue.

bartgol:
> Should I wait for the AT to report? I assume you found and fixed a DAG-related issue.

Yeah, that's it. I just wanted an approval, so it gets merged whenever the AT manages to get its Successful Hierarchical Integration Testing together.

jeff-cohere (approved):

jeff-cohere (approved):

Issue #801, opened 2020-12-08
Wrong ctest_resource_file generated for non-parallel builds with 2+ builds


bartgol:
The first build gets the correct resource file (devices ids from 0 to ndevices-1), but the second build gets devices ids (ndevices,2*ndevices-1).

This is a problem only on GPU, since that's the only case where we _use_ those numbers (kokkos tries to initialize that gpu only).

Note: this does _not_ affect AT/nightly builds, since we always use `-p` in those jobs.

PR #802, opened 2020-12-08
SCREAM: fix to generation of ctest resource file

bartgol:
This PR fixes a bug that only shows its fangs in GPU builds _and_ if you run without the `-p` flag.

See #801 for more details.

### DIFF
--- a/components/scream/scripts/test_all_scream.py
+++ b/components/scream/scripts/test_all_scream.py
@@ -403,9 +403,13 @@ class TestAllScream(object):
         # on correct cores with taskset. On GPU, however, these ids are used to select
         # the kokkos device where the tests are run on.
 
-        it = itertools.takewhile(lambda name: name!=test, self._tests)
-        start = sum(self._testing_res_count[prevs] for prevs in it)
-        end   = start + self._testing_res_count[test]
+        if self._parallel:
+            it = itertools.takewhile(lambda name: name!=test, self._tests)
+            start = sum(self._testing_res_count[prevs] for prevs in it)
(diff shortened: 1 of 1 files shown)


components/scream/scripts/test_all_scream.py
```diff
@@ -403,9 +403,13 @@ def create_ctest_resource_file(self, test, build_dir):
         # on correct cores with taskset. On GPU, however, these ids are used to select
         # the kokkos device where the tests are run on.
 
-        it = itertools.takewhile(lambda name: name!=test, self._tests)
-        start = sum(self._testing_res_count[prevs] for prevs in it)
-        end   = start + self._testing_res_count[test]
+        if self._parallel:
+          it = itertools.takewhile(lambda name: name!=test, self._tests)
```
jgfouca, line 407: Need 4-space indents.

components/scream/scripts/test_all_scream.py
```diff
@@ -403,9 +403,13 @@ def create_ctest_resource_file(self, test, build_dir):
...
+          start = sum(self._testing_res_count[prevs] for prevs in it)
+          end   = start + self._testing_res_count[test]
+        else:
+          start = int(0)
```
jgfouca, line 411: Don't need to explicitly contruct 0 as an int, it's an int by default.

jgfouca: testing_res_count are ints too.

jgfouca (changes requested):
Minor stuff

jgfouca (approved):

Issue #759, opened 2020-11-24
Remove concept of Q (all tracers) from p3 hooks to the AD


bartgol:
This may have to wait till #758 is completed, but it should be done eventually.

As of now, having P3 accept a "packed" tracers array Q, and unpacking it internally, has a few drawbacks:

- atmosphere_microphysics.cpp needs to know qsize in order to size the Q array. It's mildly acceptable to have a SCREAM_QSIZE macro available (we need a compile time constant like that for Homme anyways).
- unpacking tracers might not be possible if the tracers layout is `(ncols,qsize,nlevs)`. We'd need them packed as `(qsize,ncols,nlevs)`.
- p3 is _assuming_ what each entry of Q represent. Even if the guesses are ok, all it takes is something being packed differently for the code to crash without any hint of what's wrong.

In short: the ad interface of P3 should assume all its required tracers are separate fields.

AaronDonahue:
I'm happy to include the solution to this in the p3-interface code.  It seems like that will be a natural place to do it.

bartgol:
Closing, cause I believe this is done already. @AaronDonahue , if I'm wrong, please, reopen.

PR #804, opened 2020-12-09
Refactor P3-AD interface to call only C++ version of P3

AaronDonahue:
This commit refactors the SCREAM-AD interface with P3 to use only the C++ implementation of P3.

This commit will also define a set of "universal" physics functions that can be accessed by any process.  These include, calculating Exner's formula given the pressure, converting temperature T into the potential temperature th, and the inverse, and calculating the width in meters for a specific level given the interface heights.

Addresses #759


AaronDonahue:
[WIP] for now as this is still a work in progress.

AaronDonahue:
TODO Item: Make unit tests for the new set of universal functions.

components/scream/src/physics/share/physics_functions.hpp
```diff
@@ -98,6 +98,19 @@ struct Functions
   //checks temperature for negatives and NaNs
   KOKKOS_FUNCTION
   static void check_temperature(const Spack& t_atm, const char* func_name, const Smask& range_mask);
+
+  // Computes exner function and potential temperature given the mid-level pressure
```
jeff-cohere, line 102: I think it would be nice to have a description of each of these "universal" functions with a bit more detail, including any assumptions about units, and the role of `range_mask`, since they could be used in various contexts.

AaronDonahue: Good point, Peter C suggested I break off the physics functions part into its own PR (see #805).  I'll make these changes there.

AaronDonahue:
switching to WIP to save on AT cycles.
1. Need to determine why p3_stand_alone and p3_shoc_stand_alone tests are failing.
2. Waiting on #805 before merging anyways.

components/scream/src/physics/share/physics_functions.hpp
AaronDonahue: handles in #805

AaronDonahue:
@jeff-cohere I've cleaned up the PR branch (note this is a different branch then the one I shared with you last week).  I'm going to wait for the AT before assigning reviewers.  But I don't anticipate any major deviations in this PR from the general structure here.

AaronDonahue:
Ahhh, it looks like the p3_shoc coupled is failing, I know the issue.  I will have it fixed by tomorrow morning.

components/scream/src/physics/p3/atmosphere_microphysics.cpp
```diff
@@ -32,52 +43,102 @@ void P3Microphysics::set_grids(const std::shared_ptr<const GridsManager> grids_m
   // Nevertheless, for output reasons, we like to see 'kg/kg'.
   auto Q = kg/kg;
   Q.set_string("kg/kg");
-
-  constexpr int NVL = SCREAM_NUM_VERTICAL_LEV;
-  constexpr int QSZ =  35;  /* TODO THIS NEEDS TO BE CHANGED TO A CONFIGURABLE */
+  auto nondim = m/m;
```
jeff-cohere, line 46: Maybe a silly question, but is there a "non-dimensional" unit supported by EKAT? "Meters per meter" sounds non-dimensional until I think about all the "kg / kg" and "mol / mol" stuff I've been seeing. @bartgol ?

AaronDonahue: not too silly.  This is kindof a hack, and maybe @bartgol  can comment on whether or not there is a legit nondim defined in EKAT.  kg/kg is a bit of a special unit because it represents kg/kg_air so it is "non-dimensional" but not unitless.  If that makes sense.

jeff-cohere: It does. I'm actually in favor of more carefully annotating these quantities, at least in our documentation, to avoid confusing non-experts. In EAGLES we deal with various of those ratios involving "air", aerosols, and gases.

components/scream/src/physics/p3/atmosphere_microphysics.hpp
```diff
@@ -84,6 +84,11 @@ class P3Microphysics : public AtmosphereProcess
 
   ekat::ParameterList     m_p3_params;
 
+  // Keep track of field dimensions and the iteration count
```
jeff-cohere, line 87: Do we explain what sort of "iteration" is going on anywhere? If not, can we add a little comment that does?

components/scream/src/physics/p3/atmosphere_microphysics.cpp
bartgol: I thought I added a `ekat::units::nondim` declaration, but I guess I didn't. Doing `auto nondim = kg/kg` obviously work, and does the same thing. The mixing ratio is indeed unitless, except that we hack its string so that it prints `kg/kg`, by doing something like `auto q = kg/kg;  q.set_string("kg/kg");`. However, internally, `q` and `nondim` store the same "dimensions".

That said, we can declare a `nondim` unit in ekat. Or you could create one less ambiguously like so `ekat::Units nondim(0,0,0,0,0,0,0);`.

components/scream/tests/scream_p3/p3_init_vals.txt
```diff
@@ -0,0 +1,216 @@
+0 0 0.0 1079.543354958876 100.0 34903.91666666666 0.0 0.0 0.0 1.0 0.0 1000000.0 0.0 1000000.0 0.0 1000000.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 1388.888888888889 7.197210072907482 0.0 0.0 0.0 0.0 0.0 0.0 1.0 1.0 1.0 0.0 0.0 0.0 0.0 0.0 0.0 150.0
```
jeff-cohere, line 1: Is it possible to include a 1-line header that describes the names of the entities? Maybe not for this PR, but to make folks' lives easier in the future.

components/scream/tests/scream_p3_shoc/p3_init_vals.txt
```diff
@@ -0,0 +1,216 @@
+0 0 0.0 1079.543354958876 100.0 34903.91666666666 0.0 0.0 0.0 1.0 0.0 1000000.0 0.0 1000000.0 0.0 1000000.0 0.0 0.0 0.0 0.0 0.0 0.0 0.0 1388.888888888889 7.197210072907482 0.0 0.0 0.0 0.0 0.0 0.0 1.0 1.0 1.0 0.0 0.0 0.0 0.0 0.0 0.0 150.0
```
jeff-cohere, line 1: Same comment as for p3_init_vals.txt.

components/scream/src/physics/p3/p3_inputs_initializer.cpp
```diff
@@ -45,88 +47,251 @@ add_field (const field_type &f, const field_type& f_ref,
...
+  std::string list_of_fields = "";
+  for (auto name : fields_to_init)
+  {
+    list_of_fields += name;
+    list_of_fields += ", ";
+    count += m_fields.count(name);
+  }
   
   if (count==0) {
     return;
```
jeff-cohere, line 96: Just curious--is this not an error condition? It's true that no initialization has to be done in this case, but is it "valid" to call this method under such conditions?

jeff-cohere (approved):
This looks good, @AaronDonahue . I left a few questions/suggestions here and there, but I see nothing that should prevent this from going in. Nice work.

components/scream/src/physics/p3/p3_inputs_initializer.cpp
```diff
@@ -45,88 +47,251 @@ add_field (const field_type &f, const field_type& f_ref,
...
+  while(getline(fid,tmp_line))
+  {
+    std::stringstream s(tmp_line);
+    std::string field;
+    std::vector<Real> field_vals;
+    while (getline(s,field,' '))
+    {
+      field_vals.push_back(std::stod(field));
+    }
+    int icol  = (int)field_vals[0];
```
bartgol, line 179: You could avoid having `field_vals`, and just do 

```
int  icol, k, ipack, ivec;
s >> icol >> k;
ipack = k / n;
ivec = k % n;

s >> h_qv(icol,ipack)[ivec];
s >> h_th_atm(icol,ipack)[ivec];
s >> h_pmid(icol,ipack)[ivec];
...
```
`std::stringstream` is designed precisely to support this kind of (generic) input/output operations.

Note: if you need to skip some values, you can always do 
```
std::string skip;
s >> skip >> skip >> h_qv(icol,ipack)[ivec];
...
```

bartgol (changes requested):
Great to see this getting close to be completed. I have a few comments/requests.

AaronDonahue:
@bartgol  and @jeff-cohere , thank you for the comments and review.  I'm going to work on these now.

components/scream/src/physics/p3/CMakeLists.txt
```diff
@@ -15,7 +15,6 @@ set(P3_SRCS
   ${SCREAM_BASE_DIR}/../eam/src/physics/cam/micro_p3.F90
```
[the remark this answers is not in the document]
AaronDonahue: I'm not sure why this doesn't cause a link error.  Perhaps @jgfouca has an idea.  I'm pretty sure this is included here for the C++/F90 BFB unit tests.  I don't think any of the code in the AD interface call micro_p3.F90 directly.

components/scream/src/physics/p3/atmosphere_microphysics.cpp
```diff
@@ -1,6 +1,8 @@
-#include "physics/p3/scream_p3_interface.hpp"
 #include "physics/p3/atmosphere_microphysics.hpp"
 #include "physics/p3/p3_inputs_initializer.hpp"
+#include "physics/p3/p3_main_impl.hpp"
+// Needed for p3_init, the only F90 code still used.
+#include "physics/p3/p3_f90.hpp"
```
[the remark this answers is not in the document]
AaronDonahue: TBH I don't know.  My understanding is that it sets up the table lookup values used within p3_main.  @jgfouca , does that sound correct?

components/scream/src/physics/p3/CMakeLists.txt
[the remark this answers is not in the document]
jeff-cohere: For what it's worth: if the function that needs `micro_p3_utils_init_c` isn't called itself, the linker is smart enough to eliminate the dependency.

components/scream/src/physics/p3/atmosphere_microphysics.cpp
AaronDonahue: I think actually might have:
```  static Units nondimensional () {
    return Units(RationalConstant::one());
  }```

File: components/scream/src/physics/p3/atmosphere_microphysics.cpp [hunk 10]
```diff
@@ -32,52 +43,102 @@ void P3Microphysics::set_grids(const std::shared_ptr<const GridsManager> grids_m
...
+  m_computed_fields.emplace("ni",     scalar3d_layout_mid, 1/kg, grid_name);
+  m_computed_fields.emplace("bm",     scalar3d_layout_mid, 1/kg, grid_name);
+  m_computed_fields.emplace("th_atm", scalar3d_layout_mid, K,    grid_name);
+  // Diagnostic Inputs: (only the X_prev fields are both input and output, all others are just inputs)
+  m_required_fields.emplace("nc_nuceat_tend",  scalar3d_layout_mid, 1/(kg*s), grid_name);
+  m_required_fields.emplace("nccn_prescribed", scalar3d_layout_mid, nondim,   grid_name);
+  m_required_fields.emplace("ni_activated",    scalar3d_layout_mid, 1/kg,     grid_name);
+  m_required_fields.emplace("inv_qc_relvar",   scalar3d_layout_mid, nondim,   grid_name);
+  m_required_fields.emplace("dp",              scalar3d_layout_mid, Pa,       grid_name);
+  m_required_fields.emplace("qv_prev",         scalar3d_layout_mid, Q,        grid_name);
```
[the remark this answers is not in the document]
AaronDonahue: We will need to store the two `_prev` fields in any restart file.  Given our current setup that would require these two fields to be registered with the field manager.

components/scream/src/physics/p3/atmosphere_microphysics.cpp
```diff
@@ -91,7 +152,10 @@ void P3Microphysics::initialize_impl (const util::TimeStamp& t0)
   //  - initable fields may not need initialization (e.g., some other atm proc that
   //    appears earlier in the atm dag might provide them).
 
-  std::vector<std::string> p3_inputs = {"q","T","FQ","ast","ni_activated","nc_nuceat_tend","pmid","dp","zi","qv_prev","T_prev"};
+  std::vector<std::string> p3_inputs = {"T_atm","ast","ni_activated","nc_nuceat_tend","pmid","dp","zi","qv_prev","T_prev",
```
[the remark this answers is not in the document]
AaronDonahue: Should the EKAT error message check be included in the `#ifndef`?

components/scream/src/physics/p3/p3_inputs_initializer.cpp
AaronDonahue: So I should switch this to EKAT_REQUIRE?

components/scream/src/physics/p3/p3_inputs_initializer.cpp
AaronDonahue: excellent, thank you for the suggestion.  I've cleaned this up and done this.

components/scream/src/physics/p3/p3_inputs_initializer.cpp
```diff
@@ -45,88 +47,251 @@ add_field (const field_type &f, const field_type& f_ref,
...
-  Kokkos::deep_copy(d_qv_prev,h_qv_prev);
-  Kokkos::deep_copy(d_t_prev,h_t_prev);
-
-  // If we are in charge of init-ing FQ as well, init it to 0.
-  if (m_fields.count("FQ")==1) {
-    // Init FQ to 0
-    auto d_FQ = m_fields.at("FQ").get_view();
-    Kokkos::deep_copy(d_FQ,Real(0));
+ // Initalize from text file 
+  std::ifstream fid("p3_init_vals.txt", std::ifstream::in);
```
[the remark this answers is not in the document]
AaronDonahue: done

components/scream/tests/scream_p3/p3_init_vals.txt
AaronDonahue: for sure, I made the change.

components/scream/src/physics/p3/p3_inputs_initializer.cpp
jeff-cohere: Sure, that works. We can always revisit if it turns out we're wrong about this.

components/scream/src/physics/p3/atmosphere_microphysics.cpp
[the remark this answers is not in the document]
bartgol: It doesn't seem to write in any C++ provided ptr though (or in any array that is manually copied to a C++ ptr). My guess is that we don't need it. Have you tried to not call the f90 stuff and see what happens?

components/scream/src/physics/p3/atmosphere_microphysics.cpp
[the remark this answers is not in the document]
bartgol: So is "T_prev" from the end of the last atm timestep? Or is it the end of the last time P3 ran (and if so, what if, say, P3 is subcycled)?

components/scream/src/physics/p3/atmosphere_microphysics.cpp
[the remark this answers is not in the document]
bartgol: You mean the one at the bottom of the fcn? It should be inside the ifndef. Anything past `p3_init();` should be ifdef-ed out.

components/scream/src/physics/p3/CMakeLists.txt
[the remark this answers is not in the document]
bartgol: Ah, it appears that our `physics_share` library compiles `eam/src/physics/cam/micro_p3_utils.f90`. I am a bit puzzled on why physics_share should depend on `micro_p3_utils.f90`. It seems to me that such file, as the name implies, is p3-specific.

@singhbalwinder digging in git history, it seems this dep was added when Murphy-Koop stuff was added. Do you know if this _really_ requires micro_p3_utils, or if it can be removed from `physics_share`?

components/scream/src/physics/p3/atmosphere_microphysics.cpp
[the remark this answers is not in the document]
jgfouca: @bartgol , @AaronDonahue , I haven't thought about this in a while, but my recollection is we didn't want to bother converting initialization since we don't care about it's performance. p3_init must be called in order for various tables to be read and initialized.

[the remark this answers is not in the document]
bartgol: TBC, I'm ok keeping initialization in F90. What puzzles me is that I can't see where the data is transferred from F90 to C. But if you know it is, than we're good.

components/scream/src/physics/p3/CMakeLists.txt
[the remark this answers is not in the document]
singhbalwinder: I moved MurphyKoop in `wv_sat_scream.F90` so that both P3 and SHOC can use MurphyKoop. MurphyKoop depends on some variables which P3 initializes (zerodegc and ep_2), therefore I had that dependency. I agree that it should be removed. 

For removing this dependency, we might need to re-shuffle some codes and add codes to initialize these variables (zerodegc and ep_2) in the shared codes.

components/scream/src/physics/p3/atmosphere_microphysics.cpp
```diff
@@ -138,36 +205,105 @@ void P3Microphysics::run_impl (const Real dt)
     Kokkos::deep_copy(m_p3_host_views_out.at(it.first),it.second.get_view());
   }
 
-  // Call f90 routine
-  p3_main_f90 (dt, m_raw_ptrs_in["zi"], m_raw_ptrs_in["pmid"], m_raw_ptrs_in["dp"], m_raw_ptrs_in["ast"], m_raw_ptrs_in["ni_activated"], m_raw_ptrs_in["nc_nuceat_tend"], m_raw_ptrs_out["q"], m_raw_ptrs_out["FQ"], m_raw_ptrs_out["T"], m_raw_ptrs_out["qv_prev"], m_raw_ptrs_out["T_prev"]);
-
   // Copy outputs back to device
   for (auto& it : m_p3_fields_out) {
     Kokkos::deep_copy(it.second.get_view(),m_p3_host_views_out.at(it.first));
   }
```
tcclevenger, line 211: Should your call to `auto elapsed_microsec = P3F::p3_main()` below come before you copy outputs back to device?

AaronDonahue: I believe that p3_main acts on device.  So we will need to make sure that host and device views are synced before calling p3_main.  Here I am removing the F90 version, which happened on host and replacing with the C++ version.  But, I may be misunderstanding how things work, it wouldn't be the first time, so I am open to some pushback.

components/scream/src/physics/p3/atmosphere_microphysics.hpp
AaronDonahue: I removed m_it since it can now be stored in `infrastructure`, see line 190.  I also added a comment describing what it is even used for.  Question, would it make more sense to add that comment where the `P3Infrastructure` structure is defined?

components/scream/src/physics/p3/atmosphere_microphysics.cpp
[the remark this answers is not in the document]
AaronDonahue: It should be from the last time `P3` is run.  But that is a good physics question, <- @hassanbeydoun or @PeterCaldwell .

components/scream/src/physics/p3/atmosphere_microphysics.cpp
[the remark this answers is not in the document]
AaronDonahue: great, done.

AaronDonahue:
@bartgol  and @jeff-cohere , I believe I have addressed your comments on this PR.  My changes were pretty robust so a few have disappeared from the code itself.  But I addressed the two biggest comments I think: a) Handling the structures passed to p3_main and b) How to deal with the local variables that are not stored in the field manager but are needed by P3. 

Can you take another look at the PR?  In particular, focus on the big changes I made to `atmosphere_microphysics.hpp`?

Last comment, while making these changes two big tasks occurred to me:
1. Still need a better unit test.
2. I have the pre-processing set up, but not post-processing.  Which will use the p3_main output to set variables needed by other processes.

The main question is:  Do we put this PR through now to get it out of the way and I begin work on these two tasks in a subsequent PR?  Or do we try to get these done here?  I'm partial to the former so that this PR doesn't get stale and it provides a good template for the SHOC stuff.

jeff-cohere:
> The main question is: Do we put this PR through now to get it out of the way and I begin work on these two tasks in a subsequent PR? Or do we try to get these done here? I'm partial to the former so that this PR doesn't get stale and it provides a good template for the SHOC stuff.

I haven't looked over your latest changes yet, but for the record: I'm in favor of getting this PR merged, and creating issues to track the 2 outstanding items you mentioned.

components/scream/src/physics/p3/atmosphere_microphysics.cpp
bartgol: Well, elapsed_microsec is not even used, so maybe we can remove it altogether?

components/scream/src/physics/p3/p3_inputs_initializer.cpp
```diff
@@ -45,88 +47,214 @@ add_field (const field_type &f, const field_type& f_ref,
...
+  EKAT_REQUIRE_MSG(!fid.fail(),"Error in p3_inputs_initializer.cpp loading p3_init_vals.txt file");
+  std::string tmp_line;
+  getline(fid,tmp_line);  // Read header and discard.
+  int icol_in_max = 0;
+  while(getline(fid,tmp_line))
```
jeff-cohere, line 160: Just a thought: using a file format that pairs field names with field data is more robust than assuming an ordering for the fields, since you can add and delete fields in the future without disrupting a sequence (and you can ignore any fields you don't need). Might not be worth fussing over, but I thought I'd mention it.

components/scream/src/physics/p3/atmosphere_microphysics.hpp
```diff
@@ -53,6 +68,85 @@ class P3Microphysics : public AtmosphereProcess
...
+      // Nothing else to initialize at the moment.
+    };
+    KOKKOS_INLINE_FUNCTION
+    void operator()(const int icol) const {
+      int ipack, ivec;
+      for (ipack=0;ipack<m_npack;ipack++) {
+        // Exner
+        const Spack opmid(pmid(icol,ipack));
+        const Smask opmid_mask(!isnan(opmid) and opmid>0.0);
+        auto oexner = physics_fun::get_exner(opmid,opmid_mask);
```
bartgol, line 102: I'm glad you enjoy so much the physics that you call it 'fun'. :)

components/scream/src/physics/p3/atmosphere_microphysics.hpp
jeff-cohere: Yes--if the iteration count is stored in a field for that struct, the documentation for it belongs there.

components/scream/src/physics/p3/atmosphere_microphysics.hpp
AaronDonahue: yeah man, got make people excited to work with the `p3` code.  In a more serious response, it turns out that `physics` already has a defined meaning in `atmosphere_microphysics.hpp` so I had to add the fun.

components/scream/src/physics/p3/p3_inputs_initializer.cpp
AaronDonahue: I agree, in this case I was able to generate the input file directly from the `p3_run_and_cmp` output, which is why I kept it intact here, just in case there was some change in p3_run_and_cmp that I wanted reflected in the input file, and I didn't want to have to do any more formatting - i.e. lazy.

that being said, I didn't really put my heart into it because I think the more robust unit test will touch this code and would be the appropriate time to clean things up for good.

components/scream/src/physics/p3/atmosphere_microphysics.cpp
AaronDonahue: ahhh I see, your comment is about `elapsed_microsec`.  It's true it isn't used, so @bartgol 's point is well-taken.  I'm not opposed to removing it, but that would require touching the unit testing code as well, which is maybe outside the scope of this PR (?)

bartgol: I am not sure why it would touch the UT code. TBC, I'm not suggesting to make p3_main return void. I'm just suggesting we don't need the elapsed_microsec var here.

AaronDonahue: Is it possible to call a routine that returns `non-void` without assigning it some named output variable?

jeff-cohere: Yep. Just omit the assignment.

AaronDonahue: oh cool, well, that's been done.

bartgol (approved):
Let's merge this, as Jeff said, and fix the remaining issues later. Having this in can help the rest of the AD tasks to move fwd.

AaronDonahue:
I forgot to add the "Automerge" label so I'm going to go ahead and merge this now.

PR #805, opened 2020-12-09
Introduce new universal functions to physics

AaronDonahue:
This commit introduces four new universal functions to the physics/share directory.

get_exner - Determines Exner's formula given a pressure value.
T_to_th   - Converts the temperature to the potential temperature.
th_to_T   - Converts the potential temperature to the temperature.
get_dz    - Calculates the level thickness in meters given the level interface heights.


bartgol:
I haven't checked yet, but I assume this will also be needed by dynamics and/or mct coupling.

AaronDonahue:
If that's true then we can bump these up higher.  I'll note that dynamics has a different Exner calculation than the rest of the model.  I noted this in an E3SM git-issue a long time ago.  But it never really gained traction.

bartgol:
No, I think physics/share is the correct place (they are...physics formulas, after all). It's ok to have dynamics depend on physics/share, since that's relatively general stuff.

bartgol:
If there's a discrepancy in exner calculation (not just non bfb difference, but a really different definition), then it should be made extremely clear.

Also, is th potential or virtual potential? I think homme uses vth rather than th, so that might be another source of differences.

AaronDonahue:
@bartgol , I'm not sure if this is still relevant, but the git-issue is here: https://github.com/E3SM-Project/E3SM/issues/2680

The TL;DR explanation is that the typical Exners formula definition uses a reference pressure P0 which is usually something like 1000hPa.  In the dp_coupling step in EAM the state%exner is set, but instead of P0 it uses the surface pressure.  Which is spatially varying, and in some cases can be very far from 1000hPa.

My investigation (although it could be outdated now) revealed that this state%exner in dynamics is only used for the surface coupling so it may need to be that way, but it probably shouldn't be called "exner".

bartgol:
@AaronDonahue We are using a new set of physics parametrizations, so we should be free to correct/change things in the pd coupling, right? For instance, can we force all scream physics to use Homme's version of exner?

PeterCaldwell:
@bartgol - if I recall correctly, HOMME's version of exner is the opposite of the definition used by the rest of the world (I think it's actually exner^{-1}) so I'd rather be inconsistent than use it!

bartgol:
> @bartgol - if I recall correctly, HOMME's version of exner is the opposite of the definition used by the rest of the world (I think it's actually exner^{-1}) so I'd rather be inconsistent than use it!

So long as the formulas are "compatible", meaning that one can be derived from the other _without any additional input_ (like `exner_cam = 1/exner_homme`), it's fine. In my mind, "it's the same formula". But if one uses surface pressure, while the other uses a constant p0, then they are intrinsically different.

As I think Mark suggested in the issue linked by Aaron, one can always rename one of the two "rexner", to avoid confusion.

P.S.: I'm an atmosphere noob, so I rely on what the almighty google can tell me. All the hits I get for 'exner pressure' report a formula like the one used by homme (Pi=(p/p0)^kappa).

PeterCaldwell:
@bartgol, @AaronDonahue  - well, it is good I prefaced my last comment with "if I recall correctly" because you're right: dynamics is defining exner correctly and physics is the one that's backward. The equation should be T = theta/exner... I guess the fix should be to make physics use the right version of exner...

AaronDonahue:
Great thing is that I can solve this right away in this PR and it should work elsewhere too.  I'll change the exner formula version in this PR.

@bogensch and @tcclevenger  , I know that SHOC uses the reciprocal version, when we hook SHOC up to the AD and decide to use this universal function we will not to take note of that in case anything else in the shoc_intr will need to be updated.  Note that for now the F90 code should be fine as it is doing everything in-house.

AaronDonahue:
Last note, going back to the definition in the dp_coupling.  There were two discrepancies.  One was the definition of exner, which I think we have settled.

The other is that in dp_coupling the reference pressure is the surface pressure (i.e. spatially varying), while in the textbook definition it is a fixed constant, in this case 1000hPA.  I only mention this in case we decide to have homme use this function in the future, it will definitely change things.

AaronDonahue:
adding WIP until I put the actual new tests in

AaronDonahue:
I think maybe the GPU run is failing.  I'll debug this.

bartgol:
Mappy is a CPU machine. And all the physics_tests_ut_npX_ompY are failing. I don't think it's a GPU issue only.

AaronDonahue:
> Mappy is a CPU machine. And all the physics_tests_ut_npX_ompY are failing. I don't think it's a GPU issue only.
> 
> Edit: though GPU is failing to build, so there's more to check there.

I think that the sp test is failing because of the sensitivity to the machine epsilon tolerance.

AaronDonahue:
I checked with syrah and lassen and can't recreate the fail.  Although, it looks pretty clear that physics_tests are failing across the board.  I think it's the tolerance for test failure.  Can't be sure until checking the verbose output.  I suppose one solution would be to bump up the tolerance and retest.

AaronDonahue:
good news, I can reproduce the error on syrah.  So now I have a way to hunt down the problem.

AaronDonahue:
got it.  It should pass now.

PeterCaldwell (approved):
This is a really great, very clean PR. Thanks!

AaronDonahue:
@PeterCaldwell , I think you need to re-review since I pushed an updated branch.
@ambrad , I've added you as a reviewer because I touched the definition of Tol in the physics constants.

This PR passed the last round of AT-tests, but I forgot to fix the comments per Peter's suggestion, so I pushed a new version that I'm expecting will pass as well.

ambrad:
@AaronDonahue I don't want to review based on just the name change Tol -> macheps, so I've removed my name. Thanks for changing the name. scream::physics::Constants in general needs an overhaul to remove some of the all-caps names, review entries like `masterproc    = true`, move entries like `CONS3` to param-specific constants files, rationalize definition policy (e.g., NSMALL at the bottom of the file), and probably other things I'm not seeing.

components/scream/src/physics/share/tests/physics_universal_unit_tests.cpp
```diff
@@ -0,0 +1,205 @@
...
+//-----------------------------------------------------------------------------------------------//
+  static void run()
+  {
+    using physics = scream::physics::Functions<Scalar, Device>;
+
+    int nerr = 0;
+    TeamPolicy policy(ekat::ExeSpaceUtils<ExeSpace>::get_default_team_policy(1, 1));
+    Kokkos::parallel_reduce("test_universal_physics", policy, KOKKOS_LAMBDA(const MemberType&, int& errors) {
+
+      errors = 0;
```
ambrad, line 155: Since this is a sum reduction, the error count should be accumulated, i.e., `error += ...`. Setting it to 0 has undefined and very likely incorrect results.

components/scream/src/physics/share/physics_universal_impl.hpp
```diff
@@ -0,0 +1,118 @@
...
+{
+  Spack result;
+  
+  static constexpr Scalar p0     = C::P0;
+  static constexpr Scalar rd     = C::RD;
+  static constexpr Scalar inv_cp = C::INV_CP;
+  const Spack exner = pow( P/p0, rd*inv_cp );
+  // Check that there are no obvious errors in the result.
+  auto is_nan_exner = isnan(exner) && range_mask;
+  EKAT_KERNEL_REQUIRE_MSG(!(is_nan_exner.any()), "Error in get_exner, Exner has NaN values.\n"); // exit with an error message
```
ambrad, line 42: REQUIRE implies this is always on, even in non-debug builds. Is that intended? For errors that are to be caught only in debug builds, use EKAT_KERNEL_ASSERT_MSG. (There tends to be a convention that ASSERT implies dependence on NDEBUG whereas REQUIRE or similarly named macros are always enabled.)

components/scream/src/physics/share/physics_universal_impl.hpp
```diff
@@ -0,0 +1,118 @@
...
+//   exner  is the exners formula, see definition above, unitless
+template <typename S, typename D>
+KOKKOS_FUNCTION
+typename Functions<S,D>::Spack
+Functions<S,D>::T_to_th(const Spack& T_atm, const Spack& exner, const Smask& range_mask)
+{
+  Spack result;
+  const Spack th_atm = T_atm/exner;
+  // Check that there are no obvious errors in the result.
+  check_temperature(th_atm,"T_to_th",range_mask);
```
ambrad, line 64: At some point there should be an empty def of check_temperature switched on NDEBUG.

ambrad (changes requested):
I ended up reviewing this. The main issue to fix is the `error` accumulation in the unit test's `parallel_reduce`. The other points I raise needn't be addressed in this PR.

components/scream/src/physics/share/tests/physics_universal_unit_tests.cpp
AaronDonahue: oh yes, good catch.  I'll make this change and push a the fix.

components/scream/src/physics/share/physics_universal_impl.hpp
AaronDonahue: TBH, I'm not sure.  My impulse is that we should never let a NaN exner value get through, debug or not debug build.  But if there is a good performance reason to only catch in DEBUG builds I'm happy to switch to ASSERT.  Presumably, anyone who is getting errors in a non-debug build should be checking with a debug build eventually.

bartgol: I agree with Andrew. I think kernels should only contain debug checks. Otherwise, where do we draw the line? If we check that one thing is not NaN, why not checking all the inputs every time? The cost of all those checks on all entries of the arrays can quickly pile up.(1)

Release builds should contain checks on scalar quantities (meaning quantities that are not arrays over the pb geo/time dimensions). E.g., if you compute some reduction (max/min) of a distributed quantity as part of the algo, it is ok to check the reduction value.

1. Note: I'm ok if you check all the inputs _in a debug build_. Debug runs can contain as many checks as we want.

AaronDonahue: ok cool. I'm happy to change this to ASSERT. I'll do that shortly and push a new version.

AaronDonahue: My apologies for going with REQUIRE, I noticed just now that I was using the `check_temperature` routine in `physics_saturation_impl.hpp` as an example.   But since that is a case where you would want an error if the temperature had incorrect values it makes sense that it would throw an error in debug and non-debug builds alike.

ambrad: Note that if you switch, you should make the error checking all inside the ASSERT call, like this:
```c++
EKAT_KERNEL_ASSERT_MSG(!(isnan(exner) && range_mask).any(), "msg");
```
That way, when NDEBUG is defined, the whole error-checking calculation is removed.

AaronDonahue:
@ambrad , pinging you again now that I've put in the changes you requested.  Thanks for taking the time to review the PR.

ambrad (approved):

PR #806, opened 2020-12-09
SCREAM: allow user to have a local mach specs file

bartgol:
I grew tired of typing `-m linux-generic --baseline-dir /path/to/my/baselines` when using test-all-scream on my laptop. I wanted a "faster" way to tell test-all about the baseline directory. So I added a new capability to test-all-scream.

You can now pass the `-l` (or `--local`) flag, and test-all-scream will look for `scream_mach_specs.py` inside the `~/.cime` folder (which you might already have if you run cime stuff on your local machine). This file must contain a list called MACHINE_METADATA, which contains what each entry of the dict in `scripts/machine_specs.py` does.

The local run is mutually exclusive with the `-m` flag (so you can't do `-m blake -l`). Also, an error is thrown if asking for `-l` but no specs file is found in `~/.cime`. In that folder, test-all-scream will also look for a cmake cache file called `scream_mach_file.cmake`, and error out if that is not found.

Now I can do `test-all-scream -l --baseline-dir AUTO` instead of `test-all-scream -m linux-generic --baseline-dir /my/very/long/path/to/baselines`.

Edit: also, using the scream_mach_specs.py file I don't need to manually setup my environment so that all required packages are in the PATH/LD_LIBRARY_PATH every time (remember that linux-generic in scripts/machine_specs.py has no env setup). Instead, I can put all the env setup commands in my local `~/.cime/scream_mach_specs.py` file, and simply invoke test-all-scream from a fresh terminal with no env/modules setup.

### DIFF
--- a/components/scream/scripts/machines_specs.py
+++ b/components/scream/scripts/machines_specs.py
@@ -12,7 +12,7 @@
 #       subdirs are located (full_debug, full_sp_debug, debug_no_fpe).
 
 from utils import expect, get_cpu_core_count, run_cmd_no_fail
-import os
+import os, sys
 import pathlib
 
 MACHINE_METADATA = {
@@ -82,6 +82,14 @@ MACHINE_METADATA = {
     "linux-generic-serial" : ([],["mpicxx","mpifort","mpicc"],"", get_cpu_core_count(), get_cpu_core_count(),""),
 }
 
+if (pathlib.Path("~/.cime/scream_mach_specs.py").expanduser().is_file()):
+    sys.path.append(str(pathlib.Path("~/.cime").expanduser()))
+    from scream_mach_specs import MACHINE_METADATA as LOCAL_MD
+    if len(LOCAL_MD) == 6:
+        MACHINE_METADATA["local"] = LOCAL_MD
+    else:
+        print("WARNING! File '~/.cime/scream_mach_specs.py' was found, but the MACHINE_METADATA in there is badly formatted. Ignoring it.")
+
 ###############################################################################
 def is_machine_supported(machine):
 ###############################################################################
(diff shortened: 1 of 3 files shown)


components/scream/scripts/test_all_scream.py
```diff
@@ -63,7 +65,12 @@ def __init__(self, cxx_compiler, f90_compiler, c_compiler, submit=False, paralle
             if "CIME_MACHINE" in os.environ and is_machine_supported(os.environ["CIME_MACHINE"]):
                 self._machine = os.environ["CIME_MACHINE"]
             else:
-                expect(False, "Machine is now required by test-all-scream")
+                expect(self._local,
+                "test-all-scream requires either the machine arg (-m $machine) or the -l flag,"
```
jgfouca, line 69: Strange indenting scheme here. I'd expect this line and the one below to have an addition 4 spaces of indent.

components/scream/scripts/machines_specs.py
```diff
@@ -82,6 +82,20 @@
     "linux-generic-serial" : ([],["mpicxx","mpifort","mpicc"],"", get_cpu_core_count(), get_cpu_core_count(),""),
 }
 
+###############################################################################
```
jgfouca, line 85: I think it would be easier to just modify MACHINE_METADATA directly if a local mach spec file is detected. Something like
```
MACHINE_METADATA = {...}
if pathlib.Path("~/.cime/scream_mach_specs.py").expanduser().is_file():
    sys.path.append(str(pathlib.Path("~/.cime").expanduser()))
    from scream_mach_specs import MACHINE_METADATA as LOCAL_MD
    if len(LOCAL_MD == 6):
        MACHINE_METADATA["local"] = LOCAL_MD
```

Then the rest of the file can remain unchanged.

jgfouca (changes requested):
Cool new feature. Couple suggestions.

components/scream/scripts/machines_specs.py
bartgol: good idea. it will keep the code simpler in the rest of the file.

jgfouca (approved):

PR #810, opened 2020-12-10
Add SCREAM AD interface to RRTMGP

brhillman:
Add SCREAM AD interface to the RRTMGP++ (C++ rewrite of RRTMGP) radiation package. This also adds a standalone test of RRTMGP++ to both test the implementation and demonstrate coupling to the AD. The test runs RRTMGP++ twice: once with a simple interface to the RRTMGP++ standalone code that is duplicated from the RRTMGP repo that compares against baseline fluxes, and once with the same test problem run through the AD. In the AD test, we then grab the computed fluxes from the Field Manager and compare against the same baseline fluxes as the standalone code.

The RRTMGP++ code uses Matt Norman's YAKL library for performance portability. Luckily, YAKL can wrap Kokkos pointers, and we make use of this in the interface between SCREAM and RRTMGP. YAKL is added as a submodule in the `externals` directory, and is built as a dependency to the RRTMGP++ code.

The build for the SCREAM interface to RRTMGP++ consists of two parts: building the core RRTMGP++ code as a library (named `rrtmgp`), and building the SCREAM-specific interface code as a second library to be used elsewhere in the model (`scream_rrtmgp`).


components/scream/CMakeLists.txt
```diff
@@ -42,7 +42,8 @@ if (Kokkos_ENABLE_CUDA)
   EkatSetNvccWrapper()
 endif()
 
-set (CMAKE_CXX_STANDARD 11)
+#set (CMAKE_CXX_STANDARD 11)
```
jeff-cohere, line 45: Is this an issue? @bartgol , are we cool with moving to C++14?

components/scream/src/physics/rrtmgp/CMakeLists.txt
```diff
@@ -1,50 +1,59 @@
...
+  ${SCREAM_BASE_DIR}/../eam/src/physics/rrtmgp/external/cpp/rrtmgp/mo_rrtmgp_constants.cpp
+  ${SCREAM_BASE_DIR}/../eam/src/physics/rrtmgp/external/cpp/rrtmgp/mo_rrtmgp_util_reorder.cpp
+  ${SCREAM_BASE_DIR}/../eam/src/physics/rrtmgp/external/cpp/rte/mo_rte_lw.cpp
+  ${SCREAM_BASE_DIR}/../eam/src/physics/rrtmgp/external/cpp/rte/mo_rte_sw.cpp
+  ${SCREAM_BASE_DIR}/../eam/src/physics/rrtmgp/external/cpp/rte/expand_and_transpose.cpp
+  ${SCREAM_BASE_DIR}/../eam/src/physics/rrtmgp/external/cpp/rte/kernels/mo_fluxes_broadband_kernels.cpp
+  ${SCREAM_BASE_DIR}/../eam/src/physics/rrtmgp/external/cpp/rte/kernels/mo_optical_props_kernels.cpp
+  ${SCREAM_BASE_DIR}/../eam/src/physics/rrtmgp/external/cpp/rte/kernels/mo_rte_solver_kernels.cpp
+  ${SCREAM_BASE_DIR}/../eam/src/physics/rrtmgp/external/cpp/examples/mo_load_coefficients.cpp
+  # This is a hack; I should not have to build this here, but we need it for the standalone RRTMGP
```
jeff-cohere, line 23: Is this comment referring to the line above? If so, it's customary (in my experience, at least) to place the comment above what it's referring to.

components/scream/src/physics/rrtmgp/CMakeLists.txt
```diff
@@ -1,50 +1,59 @@
...
 add_library(rrtmgp ${RRTMGP_SRCS})
+set_source_files_properties(${RRTMGP_SRCS} PROPERTIES COMPILE_FLAGS "${YAKL_CXX_FLAGS}")
 target_include_directories(rrtmgp PUBLIC ${SCREAM_INCLUDE_DIRS})
-target_include_directories(rrtmgp PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/../share)
-target_include_directories(rrtmgp SYSTEM PUBLIC ${SCREAM_TPL_INCLUDE_DIRS} ${CIMEROOT}/src/share/include)
-set_target_properties(rrtmgp PROPERTIES
-      Fortran_MODULE_DIRECTORY ${SCREAM_F90_MODULES})
-target_link_libraries(rrtmgp physics_share scream_share ${SCREAM_TPL_LIBRARIES})
+target_include_directories(rrtmgp PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/../common)
+#target_include_directories(rrtmgp SYSTEM PUBLIC ${SCREAM_TPL_INCLUDE_DIRS} ${CIMEROOT}/src/share/include)
```
jeff-cohere, line 41: We can get rid of stuff like this, unless you think we'll be adding it back in later.

components/scream/src/physics/rrtmgp/CMakeLists.txt
jeff-cohere: Ahh, looks like it might pertain to the stuff below. Maybe this just has to do with GitHub's code review interface.

components/scream/tests/rrtmgp/input.yaml
```diff
@@ -9,11 +9,12 @@ Atmosphere Processes:
...
+    Can Initialize All Inputs: true
 
 Grids Manager:
   Type: Physics Only
   Reference Grid: Physics
   Physics Only:
-    Number of global columns:   10
-    Number of vertical levels: 128
+    Number of global columns:  128
+    Number of vertical levels: 42
```
jeff-cohere, line 19: Is this right?

components/scream/src/physics/rrtmgp/scream_rrtmgp_interface.hpp
```diff
@@ -2,25 +2,59 @@
...
+                real2d &sw_flux_up, real2d &sw_flux_dn, real2d &sw_flux_dn_dir,
+                real2d &lw_flux_up, real2d &lw_flux_dn);
+        /*
+         * Perform any clean-up tasks
          */
-        extern void rrtmgp_init();
-        extern void rrtmgp_main();
         extern void rrtmgp_finalize();
+        /*
+         * Things that probably should not be used outside of the RRTMGP interface...
```
jeff-cohere, line 44: If you don't want to make these functions available to anyone outside of a certain context, you might consider placing them into a separate header only included by stuff within the interface.

components/scream/src/physics/rrtmgp/scream_rrtmgp_interface.cpp
```diff
@@ -1,13 +1,284 @@
...
+         * We want to initialize these once and use throughout the life of the
+         * program, so declare here and read data in during rrtmgp_initialize().
+         */
+        CloudOptics cloud_optics_sw;
+        CloudOptics cloud_optics_lw;
+
+        bool initialized = false;
+
         /* 
          * Define some dummy routines so we can start working on the interface
```
jeff-cohere, line 44: Are these still dummy routines?

components/scream/src/physics/rrtmgp/rrtmgp_inputs_initializer.hpp
```diff
@@ -0,0 +1,37 @@
...
+{
+public:
+
+  virtual ~RRTMGPInputsInitializer () = default;
+
+  // The name of the initializer
+  std::string name () const { return "RRTMGPInputsInitializer"; }
+  
+  // Initialize fields
+  void initialize_fields ();
```
jeff-cohere, line 18: You can annotate this (and other methods you've overridden from the `FieldInitializer` base class) with the `override` keyword to make it clearer that you're just implementing an interface here.

components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
```diff
@@ -1,25 +1,237 @@
 #include "ekat/ekat_assert.hpp"
 #include "physics/rrtmgp/scream_rrtmgp_interface.hpp"
 #include "physics/rrtmgp/atmosphere_radiation.hpp"
+#include "physics/rrtmgp/rrtmgp_inputs_initializer.hpp"
+#include "mo_gas_concentrations.h"
 
 namespace scream {
     RRTMGPRadiation::RRTMGPRadiation (const ekat::Comm& comm, const ekat::ParameterList& params) : m_rrtmgp_comm (comm), m_rrtmgp_params (params) {
```
jeff-cohere, line 8: It might be good to include the constructor for the `AtmosphereProcess` base class at the beginning of this initializer list, to make clear that this is how construction works (and to have an obvious place to change something if that base class constructor is altered).

jeff-cohere (approved):
Looks pretty good to me. I must admit I don't know much about the details of this process. I've left some comments for your consideration, but none prevent this PR from going in.

components/scream/src/physics/rrtmgp/CMakeLists.txt
brhillman: It refers to the hack below :-)

components/scream/src/physics/rrtmgp/CMakeLists.txt
brhillman: Thanks, done.

components/scream/src/physics/rrtmgp/scream_rrtmgp_interface.cpp
brhillman: They most certainly are NOT dummy routines! Thanks, fixed the comment.

components/scream/tests/rrtmgp/input.yaml
brhillman: For the purpose of the standalone test, I think yes; the input problem has 42 vertical levels, and is computed for 128 columns.

components/scream/src/physics/rrtmgp/scream_rrtmgp_interface.hpp
brhillman: Yeah that's not a bad idea. I'm somewhat on the fence on this, but in that strategy, would you recommend putting these prototypes in a separate header and then the function bodies in a separate source file? I.e., something like
```
scream_rrtmgp_interface_private.hpp
scream_rrtmgp_interface_private.cpp
```

components/scream/src/physics/rrtmgp/rrtmgp_inputs_initializer.hpp
brhillman: Ahhh, that's slick, I love clarity. I think I implemented what you're talking about in the latest commit, let me know if it looks funny. Thanks!

components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
brhillman: I'm confused; do you mean to call the base class constructor at the top of the RRTMGPRadiation constructor?

components/scream/CMakeLists.txt
brhillman: FYI, I believe this was to work around something in YAKL. I can try downgrading, Matt has since upgraded YAKL.

components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
brhillman: Like this?
```
    RRTMGPRadiation::RRTMGPRadiation (const ekat::Comm& comm, const ekat::ParameterList& params)                                                                    
        : AtmosphereProcess::AtmosphereProcess(), m_rrtmgp_comm (comm), m_rrtmgp_params (params) {
        /*
         * Anything that can be initialized without grid information can be initialized here.
         * I.e., universal constants, options, etc.
         */
          if (!m_rrtmgp_params.isParameter("Grid")) {
              m_rrtmgp_params.set("Grid",std::string("SE Physics"));
          }

          m_initializer = create_field_initializer<RRTMGPInputsInitializer>();
    }  // RRTMGPRadiation::RRTMGPRadiation
```

jeff-cohere: Yup, exactly.

components/scream/src/physics/rrtmgp/rrtmgp_inputs_initializer.hpp
jeff-cohere: Hmm, I don't see it. Did you just add the `override` keyword at the end of the method declaration (before the `;` above)?

components/scream/src/physics/rrtmgp/scream_rrtmgp_interface.hpp
jeff-cohere: That would certainly work. I don't think this is a big deal, but if you're worried about limiting visibility, it never hurts to move them out of sight.

components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
```diff
@@ -1,25 +1,237 @@
...
         /*
          * Anything that can be initialized without grid information can be initialized here.
          * I.e., universal constants, options, etc.
          */
+          if (!m_rrtmgp_params.isParameter("Grid")) {
+              m_rrtmgp_params.set("Grid",std::string("SE Physics"));
```
bartgol, line 14: `SE Physics` is no longer a valid name. The name is now `Physics GLL`. But in general, I would avoid setting the grid name if not defined.

Also, all other atm procs use the key `Grid Name`. You don't have to use the same (every atm proc can use whatever keys they want), but it might help reduce confusion.

components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
```diff
@@ -1,25 +1,237 @@
...
     void RRTMGPRadiation::set_grids(const std::shared_ptr<const GridsManager> grids_manager) {
 
+        using namespace ekat::units;
+
+        auto kgkg = kg/kg;
+        kgkg.set_string("kg/kg");
+        auto m3 = m * m * m;
+        m3.set_string("m3");
+        auto Wm2 = W / m / m;
+        Wm2.set_string("Wm2");
```
bartgol, line 29: The `set_string` method is usually not needed. I added it only cause ppl wanted to see `kg/kg` for the tracers mixing ratio. But I think for most units you might be ok with the default ones? Like, for `m3` you would see printed "m^3". For `Wm2` the string might help, otherwise you would get something like "s^-3 kg", which doesn't immediately struck as the same as W/m2.

components/scream/src/physics/rrtmgp/rrtmgp_inputs_initializer.hpp
brhillman: https://github.com/E3SM-Project/scream/pull/810/commits/273f1d148b78878308fbab05256b7dd415998fb8

components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
```diff
@@ -1,25 +1,237 @@
...
+        auto nondim = m/m;  // dummy unit for non-dimensional fields
+        auto micron = m / 1000000;
+
+        auto VL  = FieldTag::VerticalLevel;
```
bartgol, line 33: You could do `using namespace ShortFieldTagsNames;`. Those short names are already available in that namespace (see `field_tag.hpp`).

components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
```diff
@@ -1,25 +1,237 @@
...
+        auto COL = FieldTag::Column;
+        auto VAR = FieldTag::Variable;
+        auto CMP = FieldTag::Component;
+        constexpr int NVL = 42;  /* TODO THIS NEEDS TO BE CHANGED TO A CONFIGURABLE */
```
bartgol, line 37: The grid object now stores the number of levels. So you can do

```
const int nlevs = grid->get_num_vertical_levels();
```

components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
```diff
@@ -1,25 +1,237 @@
...
+
+        auto grid = grids_manager->get_grid("Physics");
+        const int num_dofs = grid->get_num_local_dofs();
+        //const int ncol = 128; //num_dofs;
+        const int ncol = num_dofs;
```
bartgol, line 42: Since you don't use num_dofs below, you could simply replace these two lines with
```
const int ncol = grid->get_num_local_dofs();
```

components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
```diff
@@ -1,25 +1,237 @@
...
+                !init_all_or_none || all_inited || all_uninited,
+                "Error! Some rrtmgp inputs were marked to be inited by RRTMGP, while others weren't.\n"
+                "       RRTMGP was requested to init either all or none of the inputs.\n"
+            );
+        }
+    }
+    void RRTMGPRadiation::run_impl      (const Real dt) {
+        // Get data from AD; RRTMGP wants YAKL views
+        // TODO: how can I just keep these around without having to create every time?
+        // They are just pointers, so should be able to keep them somewhere else and just associate them once?
```
bartgol, line 150: Yes. That's the direction where P3 is heading to. You could have a helper struct like
```
struct YaklArrays {
  using Array2d = yakl::Array<double,2,memDevice,yakl::styleFortran>;

  Array2d pmid;
  Array2d pint;
...
};
```
Then in RRTMGPRadiation you could store a YaklArrays object, and init the views once and for all during `initialize_impl`.

components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
```diff
@@ -1,25 +1,237 @@
...
+
+        // Get device views
+        auto d_pmid = m_rrtmgp_fields_in.at("pmid").get_view();
```
bartgol, line 153: One thing you should be aware of is that views _could_ be padded, to accommodate vectorization.

This is obviously not a priority, but if at some point vectorization is requested, and no checks are present, you'll suddendly complete garbage (likely nans).

One way to check that no padding was present is to check that, for instance, `d_pmid.size()` is equal to `pmid_layout.size()`, where `pmid_layout` is the field layout, which is found in `m_rrtmgp_fields_in.at("pmid").get_header().get_identifier().get_layout()`.

Alternatively, if you stored ncols already, you can directly check `d_pmid.size()==ncols*nlevs`, which is faster. If that test fails, it means someone requested an allocation for pmid that allowed for packing, and since the requested pack length did not divide the number of levels, the field repo added some padding.

Note: vector lengths are usually 4 or 8, which should divide the number of levels of a reasonable compset, but a) better not rely on "reasonable", and b) if mid points have no padding, then interface quantities will have padding.

components/scream/tests/rrtmgp/CMakeLists.txt
```diff
@@ -1,19 +1,35 @@
...
 
 # Copy yaml input file to run directory
 CONFIGURE_FILE(${CMAKE_CURRENT_SOURCE_DIR}/input.yaml
                ${CMAKE_CURRENT_BINARY_DIR}/input.yaml COPYONLY)
 
 # Test atmosphere processes
-CreateUnitTest(rrtmgp_stand_alone "rrtmgp_stand_alone.cpp" "${NEED_LIBS}" LABELS "rrtmgp;physics")
+CreateUnitTest(rrtmgp_stand_alone "${SRC}" "${NEED_LIBS}" LABELS "rrtmgp;physics")
+set_source_files_properties("${SRC}" PROPERTIES COMPILE_FLAGS "${YAKL_CXX_FLAGS}")
+target_include_directories(rrtmgp_stand_alone PUBLIC ${SCREAM_BASE_DIR}/../../externals/YAKL)
```
bartgol, line 15: Ugh, YAKL really needs to add its src folder to its target include directories, so we don't need to do this...

components/scream/tests/rrtmgp/rrtmgp_stand_alone.cpp
```diff
@@ -1,32 +1,171 @@
...
+     */
+    TEST_CASE("rrtmgp_scream_stand_alone", "") {
         using namespace scream;
         using namespace scream::control;
 
-        /* 
-         * Setup driver stuff
-         */
+        // Initialize yakl
+        yakl::init();
```
bartgol, line 155: Every TEST_CASE causes this file to be executed multiple times _within the same run_. Is YAKL ok with being inited/finalized multiple times during the same run?

Kokkos, for instance, is not happy if you try to re-initialize it after it's been finalized.

bartgol (commented):
Very *big* piece of work Ben! I have a few comments. Feel free to push back on any/all of them.

components/scream/src/physics/rrtmgp/rrtmgp_inputs_initializer.hpp
bartgol: I like using the override keyword. In 99% of the cases it doesn't catch bugs, but it definitely conveys the message.

Just one quick note: if I add it to one of the methods, then I have to add it to all the overridden methods, or else the compiler will issue a warning. Not a big deal, but worth remembering when trying to keep the compiler output under control.

components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
brhillman: Done. Thanks!

.gitmodules
```diff
@@ -15,10 +15,10 @@
 	path = externals/kokkos
 	url = git@github.com:E3SM-Project/kokkos.git
 	branch = e3sm/kokkos
-[submodule "components/cam/src/physics/rrtmgp/external"]
```
AaronDonahue, line 18: Was this not already fixed in the main scream branch?  How are we able to run F-Compsets if we are still pulling in a "cam" directory?

components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
```diff
@@ -1,25 +1,238 @@
...
+
+        auto grid = grids_manager->get_grid("Physics");
+        const int num_dofs = grid->get_num_local_dofs();
+        //const int ncol = 128; //num_dofs;
+        const int ncol = num_dofs;
+
+        int nswbands = 14;
+        int nlwbands = 16;
+        int ngas = 8;
+//        string1d gas_names("gas_names",ngas);
```
AaronDonahue, line 48: I assume that these commented lines are here because we want to use them in the future?

components/scream/src/physics/rrtmgp/rrtmgp_inputs_initializer.cpp
```diff
@@ -0,0 +1,184 @@
...
+        // opposite what RRTMGP expects (that is, (nx,ny,nz) -> (nz,ny,nx))
+        // Read in dummy Garand atmosphere; if this were an actual model simulation,
+        // these would be passed as inputs to the driver
+        // NOTE: set ncol to size of col_flx dimension in the input file. This is so
+        // that we can compare to the reference data provided in that file. Note that
+        // this will copy the first column of the input data (the first profile) ncol
+        // times. We will then fill some fraction of these columns with clouds for
+        // the test problem.
+        int ngas =  8;
+        int ncol = 128;
```
AaronDonahue, line 114: This appears to be a hard-coded value.  What would happen if we ran with a different number of columns or layers?

components/scream/src/physics/rrtmgp/tests/CMakeLists.txt
```diff
@@ -1,6 +1,32 @@
...
+    #target_include_directories(rrtmgp_tests PUBLIC ${SCREAM_BASE_DIR}/../eam/src/physics/rrtmgp/external/cpp/examples/all-sky)
+    #target_include_directories(rrtmgp_tests PUBLIC ${SCREAM_BASE_DIR}/../eam/src/physics/rrtmgp/external/cpp/examples/all-sky)
+    target_include_directories(rrtmgp_tests PUBLIC ${SCREAM_BASE_DIR}/../../externals/YAKL)
 endif()
+
+# Copy RRTMGP absorption coefficient lookup tables to local data directory
+FILE (MAKE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/data)
+CONFIGURE_FILE(${SCREAM_BASE_DIR}/../eam/src/physics/rrtmgp/external/rrtmgp/data/rrtmgp-data-sw-g224-2018-12-04.nc
+               ${CMAKE_CURRENT_BINARY_DIR}/data COPYONLY)
+CONFIGURE_FILE(${SCREAM_BASE_DIR}/../eam/src/physics/rrtmgp/external/rrtmgp/data/rrtmgp-data-lw-g256-2018-12-04.nc
```
AaronDonahue, line 23: Here we have some dependency on files in EAM.  How big are these?  Are they small enough that it would make sense to store them locally with SCREAM?

AaronDonahue (commented):
What, if any, considerations do we need to make for "packed" arrays?  In particular, how will YAKL respond to a variable that has packed values?

bartgol:
> What, if any, considerations do we need to make for "packed" arrays? In particular, how will YAKL respond to a variable that has packed values?

I think, for now, we can manually check that RRTMGP fields are _not_ padded. This is easy to check from the field:

```
  const auto& fid = f.get_header().get_identifier();
  const auto& layout = fid.get_layout();
  auto v = f.get_view();
  EKAT_REQUIRE_MSG (layout.size()==v.size(),
    "Error! Field " << fid.name () << " was padded (to allow packing), but currently RRTMGP does not work with padded views);
```

components/scream/CMakeLists.txt
brhillman: Just checked this, I can remove this dependence and we can decide independently of this PR whether or not to move to C++14.

components/scream/src/physics/rrtmgp/rrtmgp_inputs_initializer.hpp
brhillman: Awesome, thanks @bartgol ! I appreciate the learning opportunity!

components/scream/src/physics/rrtmgp/tests/CMakeLists.txt
brhillman: These are actually RRTMGP files (i.e., included in the external RRTMGP repo that we pull in as a submodule), not EAM files. I'm including them from the EAM directory because I elected to not pull in a separate clone of RRTMGP to the SCREAM directory. I think probably a better thing to do would be to move the RRTMGP submodule to the main "externals" directory.

components/scream/src/physics/rrtmgp/rrtmgp_inputs_initializer.cpp
brhillman: The test would fail :-) This is hard-coded for now just for the purpose of the standalone test. I will work on making this flexible.

components/scream/CMakeLists.txt
brhillman: Removed.

components/scream/src/physics/rrtmgp/CMakeLists.txt
```diff
@@ -1,50 +1,58 @@
-# Add RRTMGP source files. Right now, this builds the Fortran source for RRTMGP.
-# Once the C++ port of RRTMGP is complete we should switch over to that.
+# RRTMGP++ requires YAKL
+add_subdirectory(${SCREAM_BASE_DIR}/../../externals/YAKL ${CMAKE_CURRENT_BINARY_DIR}/YAKL_build)
+target_compile_options(yakl PUBLIC $<$<COMPILE_LANGUAGE:Fortran>:-fno-default-real-8 -fno-default-double-8>)
+target_compile_options(yakl PUBLIC $<$<COMPILE_LANGUAGE:CXX>:-std=c++14>)
```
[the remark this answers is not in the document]
brhillman: Turns out it's not needed at all with the updated YAKL.

components/scream/src/physics/rrtmgp/CMakeLists.txt
```diff
@@ -1,50 +1,58 @@
-# Add RRTMGP source files. Right now, this builds the Fortran source for RRTMGP.
-# Once the C++ port of RRTMGP is complete we should switch over to that.
+# RRTMGP++ requires YAKL
+add_subdirectory(${SCREAM_BASE_DIR}/../../externals/YAKL ${CMAKE_CURRENT_BINARY_DIR}/YAKL_build)
+target_compile_options(yakl PUBLIC $<$<COMPILE_LANGUAGE:Fortran>:-fno-default-real-8 -fno-default-double-8>)
```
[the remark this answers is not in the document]
brhillman: I think it has to do with overloaded procedures defined in YAKL. The pool allocator F90 interface overloads `real4` and `real8` parameters:
```
interface gator_deallocate
   ...
    module procedure :: gator_deallocate_real4_1d
    module procedure :: gator_deallocate_real4_2d
    module procedure :: gator_deallocate_real4_3d
    module procedure :: gator_deallocate_real4_4d
    module procedure :: gator_deallocate_real4_5d
    module procedure :: gator_deallocate_real4_6d
    module procedure :: gator_deallocate_real4_7d

    module procedure :: gator_deallocate_real8_1d
    module procedure :: gator_deallocate_real8_2d
    module procedure :: gator_deallocate_real8_3d
    module procedure :: gator_deallocate_real8_4d
    module procedure :: gator_deallocate_real8_5d
    module procedure :: gator_deallocate_real8_6d
    module procedure :: gator_deallocate_real8_7d
    ...
end interface
```
This will throw a compile-time error if `real4` is promoted to `real8`:
```
/ascldap/users/bhillma/codes/scream/branches/enable-scream-rrtmgp/externals/YAKL/gator_mod.F90:449:36:

  449 |   subroutine gator_allocate_real4_7d( arr , dims , lbounds_in )
      |                                    1
......
  562 |   subroutine gator_allocate_real8_7d( arr , dims , lbounds_in )
      |                                    2
Error: Ambiguous interfaces in generic interface 'gator_allocate' for ‘gator_allocate_real4_7d’ at (1) and ‘gator_allocate_real8_7d’ at (2)
/ascldap/users/bhillma/codes/scream/branches/enable-scream-rrtmgp/externals/YAKL/gator_mod.F90:799:38:

  799 |   subroutine gator_deallocate_real4_7d( arr )
      |                                      1
......
  835 |   subroutine gator_deallocate_real8_7d( arr )
      |                                      2
Error: Ambiguous interfaces in generic interface 'gator_deallocate' for ‘gator_deallocate_real4_7d’ at (1) and ‘gator_deallocate_real8_7d’ at (2)
```

components/scream/src/physics/rrtmgp/CMakeLists.txt
```diff
@@ -1,50 +1,58 @@
...
+target_compile_options(yakl PUBLIC $<$<COMPILE_LANGUAGE:CXX>:-std=c++14>)
+set_target_properties(yakl PROPERTIES CXX_STANDARD 14)
```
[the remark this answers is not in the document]
brhillman: Not needed at all. Removed.

components/scream/tests/rrtmgp/CMakeLists.txt
brhillman: You mean something like this in YAKL's CMakeLists.txt?
```
add_library(yakl STATIC ${YAKL_SOURCE})
target_include_directories(yakl PUBLIC ./)
```
I can make this change upstream if so.

bartgol: Yes, exactly that.

components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
brhillman: Yup good catch. Replaced.

components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
brhillman: Nah, here because I wasn't sure where to put them when I was developing. Removed.

components/scream/src/physics/rrtmgp/tests/CMakeLists.txt
brhillman: @AaronDonahue are you okay with this? I think moving the RRTMGP submodule up to the main "externals" directory should be an upstream thing, not a SCREAM thing, if we decide to do that.

brhillman:
@bartgol @AaronDonahue I implemented Luca's suggestion for requiring RRTMGP fields not be padded. I added a private function to do this since I needed to use it from both `set_required_fields_impl` and `set_computed_fields_impl`. @bartgol does this look correct? Could you school me on whether this function should be `private` or `protected`? This is the only place I imagine it will be used, I just hate copy and paste code :-)

bartgol:
> @bartgol @AaronDonahue I implemented Luca's suggestion for requiring RRTMGP fields not be padded. I added a private function to do this since I needed to use it from both `set_required_fields_impl` and `set_computed_fields_impl`. @bartgol does this look correct? Could you school me on whether this function should be `private` or `protected`? This is the only place I imagine it will be used, I just hate copy and paste code :-)

Yes, the helper function seems right to me.

Re: private vs protected. They are exactly the same if nobody will ever inherit from your class, which I think it is the case here.

components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
AaronDonahue: if you want to NVL or num_vertical_levels() to be 42 in your tests then you can adjust the `input.yaml` file to specify 42 levels.

components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
```diff
@@ -1,52 +1,272 @@
...
+        FieldLayout gas_layout          { {VAR,COL,VL}, {ngas,ncol,NVL} };
+        FieldLayout scalar2d_swband_layout { {CMP,COL}, {nswbands,ncol} };
+
+        // Set required (input) fields here
+        m_required_fields.emplace("pmid" , scalar3d_layout_mid, Pa, grid->name());
+        m_required_fields.emplace("pint", scalar3d_layout_int, Pa, grid->name());
+        m_required_fields.emplace("tmid" , scalar3d_layout_mid, K , grid->name());
+        m_required_fields.emplace("tint" , scalar3d_layout_int, K , grid->name());
+        m_required_fields.emplace("col_dry", scalar3d_layout_mid, kgkg, grid->name());
+        //m_required_fields.emplace("gas_names", gas_names_layout, nondim, grid->name());
```
AaronDonahue, line 62: any reason to keep this commented out code?

components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
```diff
@@ -1,52 +1,272 @@
...
+        m_required_fields.emplace("gas_vmr", gas_layout, kgkg, grid->name());
+        m_required_fields.emplace("sfc_alb_dir", scalar2d_swband_layout, nondim, grid->name());
+        m_required_fields.emplace("sfc_alb_dif", scalar2d_swband_layout, nondim, grid->name());
+        m_required_fields.emplace("mu0", scalar2d_layout, nondim, grid->name());
+        m_required_fields.emplace("lwp", scalar3d_layout_mid, kg/m3, grid->name());
+        m_required_fields.emplace("iwp", scalar3d_layout_mid, kg/m3, grid->name());
+        m_required_fields.emplace("rel", scalar3d_layout_mid, micron, grid->name());
+        m_required_fields.emplace("rei", scalar3d_layout_mid, micron, grid->name());
+
+        // Outputs; these needed to be added to both required and computed fields?
```
AaronDonahue, line 72: Output can just be "computed".  "required" would be for anything that acts like in/out.  For example, in the p3 interface we designate the `vap_liq_exchange` fields as just "computed" and it works great.

AaronDonahue: One caveat, which @bartgol can correct if I'm wrong, but the initializer can only initialize something that is considered "required".  The idea being that if something is strictly output then it probably doesn't need to be initialized at the beginning of the run.

components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
```diff
@@ -1,52 +1,272 @@
...
+        //  - we can request that RRTMGP either inits no inputs or all of the initable ones (as specified
+        //    at the previous point). The default is that RRTMGP must be in charge of init ing ALL or NONE
+        //    of its initable inputs.
+        // Recall that:
+        //  - initable fields may not need initialization (e.g., some other atm proc that
+        //    appears earlier in the atm dag might provide them).
+        std::vector<std::string> rrtmgp_inputs = {
+            "pmid","pint","tmid","tint","col_dry","gas_vmr","sfc_alb_dir","sfc_alb_dif",
+            "mu0","lwp","iwp","rel","rei",
+            "sw_flux_up", "sw_flux_dn", "sw_flux_dn_dir", "lw_flux_up", "lw_flux_dn"
```
AaronDonahue, line 103: I see here that you are initializing fields you call "output" above.  Are they actually being used as input/output?  If so, you may want to change the comment on line 72, otherwise future developers may be confused.

components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
AaronDonahue: On this point.  P3 will request pmid (for example) with packing.  It is necessary for passing the view to p3_main.

components/scream/src/physics/rrtmgp/rrtmgp_inputs_initializer.cpp
```diff
@@ -0,0 +1,189 @@
...
+        real2d iwp;
+        real2d rel;
+        real2d rei;
+
+        // NOTE: set ncol to size of col_flx dimension in the input file. This is so
+        // that we can compare to the reference data provided in that file. Note that
+        // this will copy the first column of the input data (the first profile) ncol
+        // times. We will then fill some fraction of these columns with clouds for
+        // the test problem.
+        // TODO: get this from the input data, or from the yaml file
```
AaronDonahue, line 121: I do this in the p3 inputs initializer here: https://github.com/E3SM-Project/scream/pull/804/files#diff-2358bb740d72ad4f39acb43909312d0599b402efd314715660dfd482ab348e62R230

components/scream/src/physics/rrtmgp/tests/CMakeLists.txt
AaronDonahue: Yes, I'm okay with this setup.  As long as there is an explanation.  I guess we would expect that at some point in the future we will have a similar set up in scream.  My understanding is that long-term we don't want scream to be dependent on anything in eam.  But for now this should be fine.  Maybe we can add a TODO comment or a git issue so we don't lose sight of this dependency.

AaronDonahue (commented):
@brhillman , this is a good looking PR.  I went through it again this morning with fresh eyes and had a few more comments.

components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
brhillman: Good catch. I think I included these when I was thinking I needed to register these with the AD here. Removed.

components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
brhillman: Thanks for catching this. I think this was required before because I was trying to initialize these in the input initializer. Removing that initialization code (since they are computed by radiation and do not need to be input) allowed me to remove them from the "required" list.

components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
brhillman: Only because I was hoping to find a way to keep track of which gas names corresponded to specific indices in the gas concentrations array. Right now I'm just assuming they don't change. I removed the commented out code though, because commented out code is silly.

components/scream/src/physics/rrtmgp/tests/CMakeLists.txt
brhillman: For now I just added a TODO comment. I will put moving the entire RRTMGP code to the top level externals directory on my list of things to do upstream. Alternatively I guess we could have an additional copy of the submodule in SCREAM? That seems silly though.

components/scream/src/physics/rrtmgp/CMakeLists.txt
```diff
@@ -1,50 +1,58 @@
...
 add_library(rrtmgp ${RRTMGP_SRCS})
+set_source_files_properties(${RRTMGP_SRCS} PROPERTIES COMPILE_FLAGS "${YAKL_CXX_FLAGS}")
 target_include_directories(rrtmgp PUBLIC ${SCREAM_INCLUDE_DIRS})
-target_include_directories(rrtmgp PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/../share)
-target_include_directories(rrtmgp SYSTEM PUBLIC ${SCREAM_TPL_INCLUDE_DIRS} ${CIMEROOT}/src/share/include)
-set_target_properties(rrtmgp PROPERTIES
-      Fortran_MODULE_DIRECTORY ${SCREAM_F90_MODULES})
-target_link_libraries(rrtmgp physics_share scream_share ${SCREAM_TPL_LIBRARIES})
+target_include_directories(rrtmgp PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/../common)
+target_include_directories(rrtmgp PUBLIC ${SCREAM_BASE_DIR}/../eam/src/physics/rrtmgp/external/cpp)
```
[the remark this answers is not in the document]
jeff-cohere: I like this suggestion.

components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
jeff-cohere: @brhillman , is this something that needs to be addressed still?

components/scream/src/physics/rrtmgp/mo_garand_atmos_io.cpp
```diff
@@ -0,0 +1,115 @@
...
+
+  int nlay = io.getDimSize("lay");
+  int nlev = io.getDimSize("lev");
+
+  p_lay = real2d("p_lay",ncol,nlay);
+  t_lay = real2d("t_lay",ncol,nlay);
+  p_lev = real2d("p_lev",ncol,nlev);
+  t_lev = real2d("t_lev",ncol,nlev);
+
+  real2d tmp2d;
```
jeff-cohere, line 20: This is a reflection of my own ignorance: is this a 2d array of reals that uses 1-based indexing?

components/scream/src/physics/rrtmgp/mo_load_coefficients.h
```diff
@@ -0,0 +1,9 @@
+#pragma once
```
jeff-cohere, line 1: I don't think this is part of the C++ standard. It does work in a lot of places, though.

components/scream/src/physics/rrtmgp/mo_garand_atmos_io.cpp
brhillman: Yes. It was decided to keep the 1-based indexing in the port of the RRTMGP code from Fortran to C++ to simplify the process. Kind of awkward, but is what it is.

components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
```diff
@@ -1,52 +1,249 @@
...
+        m_rrtmgp_host_views_out[name] = Kokkos::create_mirror_view(f.get_view());
+        m_raw_ptrs_out[name] = m_rrtmgp_host_views_out[name].data();
 
         // Add myself as provider for the field
-        //add_me_as_provider(f);
+        add_me_as_provider(f);
+
+        // Check to make sure that fields are not padded because we are not 
+        // sure how to handle that with RRTMGP yet
+        require_unpadded(f);
```
AaronDonahue, line 246: If this goes in as is we will want to prioritize getting RAD ready for padding since P3 already uses it and I will guess that the SHOC interface will too.  I'm in favor of getting this in sooner than later and putting in a patch for padding when we decide to hook RAD up to the rest of the atm. processes.  If that route is possible.

components/scream/src/physics/rrtmgp/atmosphere_radiation.hpp
```diff
@@ -58,14 +58,50 @@ namespace scream {
...
+            std::map<std::string,Real*>        m_raw_ptrs_out;
+
             std::shared_ptr<FieldInitializer> m_initializer;
 
             util::TimeStamp m_current_ts;
             ekat::Comm            m_rrtmgp_comm;
             ekat::ParameterList   m_rrtmgp_params;
+
+            // Keep track of number of columns and levels
+            int ncol;
```
AaronDonahue, line 87: This is totally a matter of preference, but I think that most headers use `m_` to start any variable that is local to that process.  Like the `m_raw_ptrs_in` above.

components/scream/src/physics/rrtmgp/mo_garand_atmos_io.cpp
```diff
@@ -0,0 +1,115 @@
...
+  // p_lay
+  io.read(tmp2d,"p_lay");
+  // for (int ilay=1 ; ilay <= nlay ; ilay++) {
```
AaronDonahue, line 23: Are we keeping these commented for loops in the code for any reason?

components/scream/src/physics/rrtmgp/mo_garand_atmos_io.cpp
```diff
@@ -0,0 +1,115 @@
...
+
+  tmp2d = real2d();     // Reset the tmp2d variable
+  for (int igas=1 ; igas <= ngas ; igas++) {
+    std::string vmr_name = "vmr_"+gas_names(igas);
+    if ( ! io.varExists(vmr_name) ) { stoprun("ERROR: gas does not exist in input file"); }
+    // Read in 2-D varaible
+    io.read(tmp2d,vmr_name);
+    // Create 1-D variable with just the first column
+    real1d tmp1d("tmp1d",nlay);
+    // for (int i=1 ; i <= nlay ; i++) {
```
AaronDonahue, line 73: same here for commented for loop

components/scream/src/physics/rrtmgp/mo_garand_atmos_io.cpp
```diff
@@ -0,0 +1,115 @@
...
+    });
+    // Call set_vmr with only the first column from the data file copied among all of the model columns
+    gas_concs.set_vmr( gas_names(igas) , tmp1d );
+  }
+
+  if ( io.varExists("col_dry") ) {
+    col_dry = real2d("col_dry",ncol,nlay);
+    tmp2d = real2d();     // Reset the tmp2d variable
+    io.read(tmp2d,"col_dry");
+    // for (int ilay=1 ; ilay <= nlay ; ilay++) {
```
AaronDonahue, line 85: and here

components/scream/src/physics/rrtmgp/rrtmgp_inputs_initializer.cpp
```diff
@@ -0,0 +1,179 @@
...
+            "Error! RRTMGPInputsInitializer is expected to init\n"
+            "       pmid, pint, tmid, tint, col_dry, gas_vmr, sfc_alb_dir, sfc_alb_dif,\n"
+            "       mu0, lwp, iwp, rel, rei,\n"
+            "       but only " + std::to_string(count) + " of those have been found.\n"
+            "       Please, check the atmosphere processes you are using,\n"
+            "       and make sure they agree on who's initializing each field.\n"
+        );
+  
+        // Get device views
+        auto d_pmid = m_fields.at("pmid").get_view();
```
AaronDonahue, line 76: For reference, if/when we switch to dealing with padded versions of these fields, there is an example of how IO deals with reading contiguous data into a padded array here: https://github.com/E3SM-Project/scream/blob/master/components/scream/src/share/io/scream_scorpio_interface.cpp#L158

AaronDonahue (approved):
I had a few more questions/comments.  But nothing that should stop this PR from going in.  Great work Ben, this was a pretty huge task.  I'm looking forward to playing with RRTMGP + Other Procs when the interface goes in.

components/scream/src/physics/rrtmgp/rrtmgp_inputs_initializer.cpp
```diff
@@ -0,0 +1,179 @@
...
+        auto d_pint = m_fields.at("pint").get_view();
+        auto d_tmid = m_fields.at("tmid").get_view();
+        auto d_tint = m_fields.at("tint").get_view();
+        auto d_col_dry = m_fields.at("col_dry").get_view();
```
bartgol, line 80: I'm assuming some of these are tracers, right? If so, I would strongly encourage to *not* use `get_view()`. Instead, use `get_reshaped_view<Real**>()` (for 2d fields). The reason is that tracers are now subviews of a bit Q tracers array, and calling get_view will _always_ return the big Q array, which is not what you want here.

components/scream/src/physics/rrtmgp/mo_load_coefficients.h
bartgol: Yeah, it's not standard, strictly speaking. I would stick with classic `#ifndef`-like header guards, since that's what happens in all the other parts of scream (and hommexx).

bartgol (changes requested):
After the bundled tracers PR went in, it became dangerous to use `get_view()` on a Field. I think that needs to be changed now.

components/scream/src/physics/rrtmgp/CMakeLists.txt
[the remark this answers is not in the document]
brhillman: Sorry I missed this before, this is a great suggestion, thanks! I tried to do this where possible, but the external RRTMGP repo assumes that full paths have been specified to the include directories. But, related to this, I actually separated the core, external RRTMGP build from the SCREAM RRTMGP interface build in a recent commit. This is to make it more clear what is our (SCREAM-specific) code and what is external RRTMGP code, and also because, ideally, RRTMGP provides a useable CMake build that we can just include here via `add_subdirectory`. That's currently not working quite right when building with CUDA, so I need to make a PR to the external RRTMGP repo. For now, I just made it more clear in our CMake build what is RRTMGP, and what is SCREAM RRTMGP interface (by building `rrtmgp` and `scream_rrtmgp` separately), and will update this once I fix the problem with CUDA in the external repo. Please do take a look at the updated `CMakeLists.txt` here though and make sure this makes sense.

components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
brhillman: I think I will leave these for now. It was helpful for me to see what each corresponded to explicitly.

components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
brhillman: These are checked in `set_required_field_impl` and `set_computed_field_impl` now.

components/scream/src/physics/rrtmgp/mo_garand_atmos_io.cpp
brhillman: This was external code that I duplicated solely for the purpose of using our own `simple_netcdf.hpp` library instead of YAKLs, so I'd prefer to keep the changes minimal to make it easier to diff against the original sources to see what we changed specifically. Matt likes the commented loops to help parse what the parallel_for is supposed to be doing.

components/scream/src/physics/rrtmgp/mo_load_coefficients.h
brhillman: Fixed.

components/scream/src/physics/CMakeLists.txt
```diff
@@ -1,5 +1,9 @@
 add_subdirectory(share)
 add_subdirectory(p3)
-add_subdirectory(rrtmgp)
+if ("${SCREAM_DOUBLE_PRECISION}")
```
bartgol, line 3: Minor: the cmake way of testing conditionals is `if (VAR_NAME)` (without dereferencing the variable). Dereferencing the variable could cause a config warning.

components/scream/src/physics/rrtmgp/scream_rrtmgp_interface.cpp
```diff
@@ -44,7 +44,7 @@ namespace scream {
          * objects. So we will initialize once, and then just update values
          * at runtime.
          */
-        GasConcs gas_concs;
+        //GasConcs gas_concs;
```
bartgol, line 47: Should this be removed then?

bartgol (approved):

components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
brhillman: Fixed, kept only where needed (kg/kg, W/m^2).

components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
brhillman: I'm going to punt on this for now.

components/scream/src/physics/rrtmgp/scream_rrtmgp_interface.cpp
brhillman: I'm leaning towards leaving this in. I'm still not 100% sure how to handle this and I think with Aaron's new initializer PR this might make more sense. I think it might be best to revisit after that PR goes in.

components/scream/src/physics/CMakeLists.txt
brhillman: Ah, cool, thanks! Fixed.

components/scream/src/physics/rrtmgp/rrtmgp_inputs_initializer.cpp
brhillman: Thanks Luca. This should be fixed now.

components/scream/src/physics/rrtmgp/atmosphere_radiation.hpp
brhillman: Oh, `m_` for "member variable". Cool, that's an easy fix, I'll implement.

components/scream/src/physics/rrtmgp/scream_rrtmgp_interface.cpp
brhillman: I'm leaning towards keeping it for now. I'm actually not 100% sure how I want to handle this, and I suspect it may change  with the new initializer, once that PR goes in. So will revisit when we revisit initialization of radiation fields.

AaronDonahue:
@brhillman Nicely done.  We will probably need to help with a few conflicts once #926 goes in, but otherwise I think we can get this merged soon.

One last question for the group (@bartgol , @brhillman , @PeterCaldwell , others), This PR has 95 commits in it, which if I'm not mistaken will make for a couple pages of git history only related to this PR.  Would it be worth it to squash all or some of these commits in case we ever have to do a code bisection in the future?

PeterCaldwell:
Yeah, nice work Ben! I'd vote for squashing and spending a bit of time to write a single coherent commit message.

brhillman:
@AaronDonahue @PeterCaldwell @bartgol squashing sounds fine to me. I can do that when rebasing after Aaron's PR goes in to resolve conflicts. Or can be done at merge time?

AaronDonahue:
@brhillman , I'm not an expert at squashing, but I think squashing after updating this branch off of master should be fine.  It may be just as easy to just merge master into this branch instead of rebasing.  my two cents :-) 

The initializer PR should go in soon, it just started a new round of AT.

brhillman:
@bartgol @AaronDonahue @jeff-cohere I have rewritten history (rebased and squashed). This probably deserves one last once over before merging. It is passing for me on mappy and weaver when I use the current HEAD for baselines, but will obviously fail on the autotester because current master does not generate baselines for these tests. 

I condensed the 100+ commits down to 10. I can go further into a single commit to keep the history cleaner if so desired. I think the painful rebase is out of the way now.

Also, Aaron and I decided to change how we handled the RRTMGP AD test for now. Rather than use the new initial condition, which very likely would have broken more things in RRTMGP, and in order to keep the tests consistent, we elected to continue to use the RRTMGP offline test, just run through the AD. In order to do this, we tell the AD to initialize the required inputs to zeros, and then between the `ad.initialize()` and `ad.run()` calls, I set the values consistent with the RRTMGP offline test. This means that we are using the same baseline for the RRTMGP offline and the RRTMGP AD tests, which I thought was a nice way of isolating the test to just test the AD interface to RRTMGP. A subsequent PR will add a *new* test that runs RRTMGP with an initial condition file constructed either from this test problem, or from output from a real SCREAMv0 simulation (as Aaron was preparing for before we decided this would be a more efficient way of getting this PR in and tested consistently).

bartgol:
I think 10 commits is a reasonable history for a PR this size. I don't think PRs should necessarily have 1 commit. I think of a commit as a logical unit of work (a new function, class, concept, fix).

components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
AaronDonahue: Should we make a git-issue so as to not forget about this?  It would probably be tagged as "low-priority" but still.

components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
AaronDonahue: We could probably make a git-issue for this too.  Maybe combined with the git-issue I suggested above, we could have a "remaining RAD tasks" issue with a couple of check-boxes.  This issue would be higher priority, but I think we should work on getting this PR merged first and then tackle this when we begin coupling RAD to the rest of the processes.

components/scream/src/physics/rrtmgp/simple_netcdf.hpp
```diff
@@ -0,0 +1,291 @@
+#ifndef SIMPLE_NETCDF_HPP
```
AaronDonahue, line 1: Not essential for this PR to go in, but should we include a comment here that lets any new developer know that this code is in place for needs particular to RRTMGP, and that any standard IO should use the scorpio interface?

I forsee a distant future where someone greps for netCDF and stumbles upon this code, and as a result bends over backwards to get their IO to work with this interface instead of scorpio.

components/scream/src/physics/rrtmgp/tests/rrtmgp_test_utils.cpp
```diff
@@ -0,0 +1,177 @@
...
+            return true;
+        } else {
+            return false;
+        }
+    }
+
+    // TODO: use YAKL intrinsics for this; this won't work on the GPU
+    bool all_equals(real2d &arr1, real2d &arr2) {
+        double tolerance = 0.01;
+        /*
```
AaronDonahue, line 25: Do we want to just delete these commented lines?

AaronDonahue (approved):
I had a few comments, but nothing that needs to impact this PR.  Just a few ideas to keep track of for future work.

components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
brhillman: Added a new issue to track. Thanks @AaronDonahue.

components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
brhillman: Thanks @AaronDonahue. I just created an entirely new issue. Contrary to the implication from this huge PR, I prefer to break these things into smaller chunks, and this seems like a well-defined unit of work to tackle in a single PR.

components/scream/src/physics/rrtmgp/tests/generate_baseline.cpp
```diff
@@ -0,0 +1,137 @@
...
+    rrtmgpTest::write_fluxes(
+        baseline, 
+        sw_flux_up, sw_flux_dn, sw_flux_dn_dir,
+        lw_flux_up, lw_flux_dn
+    );
+
+    // Clean up from test; this is probably not necessary, these things
+    // should be deallocated when they fall out of scope, but we should be
+    // good citizens and clean up our mess.
+    p_lay.deallocate();
```
jeff-cohere, line 110: It might be nice to know, as a matter of practice, whether these deallocations are actually necessary. C++ libraries are so full of leaky abstractions that it's made us all pretty superstitious about this stuff.

components/scream/src/physics/rrtmgp/simple_netcdf.hpp
brhillman: Sure, but I'd think that the fact that it exists in the `rrtmgp` subdirectory makes it pretty clear that it exists for RRTMGP and not necessarily for the rest of the model. Alternatively, we could probably rename this to something like rrtmgp_netcdf or something to make it clear that it exists for RRTMGP if you think it's existence in the `rrtmgp` subdirectory is not enough to clearly distinguish it from general SCREAM codes.

components/scream/src/physics/rrtmgp/tests/rrtmgp_test_utils.cpp
brhillman: I left them in hopes of revisiting this to provide a GPU-friendly assertion. Right now I need to make host copies to do this check.

jeff-cohere (approved):
Looks good, Ben!

components/scream/src/physics/rrtmgp/simple_netcdf.hpp
brhillman: Added comment anyways...

components/scream/src/physics/rrtmgp/tests/rrtmgp_test_utils.cpp
brhillman: Removed.

components/scream/src/physics/rrtmgp/tests/generate_baseline.cpp
brhillman: Ah yes, I’ll do some testing to confirm. I *did* have some problems with YAKL throwing segfaults on exit though if I didn’t clean things up appropriately at one point. This may not be a case of that, but, I suspect it is.

brhillman:
@bartgol what do we need to do to get this merged now that it's approved and up to date?

bartgol:
> @bartgol what do we need to do to get this merged now that it's approved and up to date?

This branch required a master merge/rebase in order for the AT to run on mappy. Aaron did it, so next AT run should only show the rrtmgp fails (due to lack of baselines). Once we see that (including mappy), we can manually merge.

bartgol:
All tests pass, except the rrtmgp ones that require baselines (not existing in master). Merging.

PeterCaldwell:
OMG finally. Hooray guys!!!

PR #819, opened 2020-12-11
Major warning cleanup

### LINKED ISSUE #524: Narrowing conversion warning in shoc_functions_f90
My compiler reports the following warning.

```
shoc_functions_f90.cpp:429:82: warning: narrowing conversion of ‘nlevi’ from ‘ekat::Int’ {aka ‘int’} to ‘long unsigned int’ inside { } [-Wnarrowing]
```

Apparently, this is not good. See [standard 8.5.4 item 3](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n4296.pdf):

```
  If a narrowing conversion (see below) is required to convert any of the arguments, the program is ill-formed.
```

jgfouca:
1. Remove many hundreds of warnings from scream
2. Update EKAT
3. Change _f implementation to use new host_to_device/device_to_host API that uses vectors instead of Kokkos::Array.
4. Whitespace clean up various files... probably a good idea to view PR changes with whitespace changes filtered out

With this PR, screams builds without warnings except for a few from zm_conv.F90 which I was less comfortable modifying. I'm adding a review request for @AaronDonahue since I did touch some zm stuff.

As a side note, there are still lots of warnings coming from eam.

Fixes #524

### DIFF
--- a/components/scream/src/physics/p3/tests/p3_ice_nucleation_unit_tests.cpp
+++ b/components/scream/src/physics/p3/tests/p3_ice_nucleation_unit_tests.cpp
@@ -30,28 +30,28 @@ struct UnitWrap::UnitTest<D>::TestIceNucleation {
     //Loop over logicals being both true and false. Use boolean:integer equivalence to loop.
     for (int do_predict_nc =0; do_predict_nc<2; do_predict_nc++){
       for (int do_prescribed_CCN = 0; do_prescribed_CCN<2; do_prescribed_CCN++){
-    
+
 	IceNucleationData self[max_pack_size] = {
 	  // temp,    inv_rho,   ni,     ni_activated,      qv_supersat_i,      inv_dt, do_predict_nc, do_prescribed_CCN
-	  {2.106E+02, 8.852E-01, 0.974E+04, 9.221E+03, 5.100E-01, inv_dt, do_predict_nc, do_prescribed_CCN },
-	  {2.052E+02, 8.852E-01, 0.874E+04, 8.221E+03, 4.100E-01, inv_dt, do_predict_nc, do_prescribed_CCN },
-	  {2.352E+02, 8.900E-01, 0.723E+04, 7.221E+03, 3.100E-01, inv_dt, do_predict_nc, do_prescribed_CCN },
-	  {2.212E+02, 9.900E-01, 0.623E+04, 6.221E+03, 2.100E-01, inv_dt, do_predict_nc, do_prescribed_CCN },
+	  {2.106E+02, 8.852E-01, 0.974E+04, 9.221E+03, 5.100E-01, inv_dt, do_predict_nc>0, do_prescribed_CCN>0 },
+	  {2.052E+02, 8.852E-01, 0.874E+04, 8.221E+03, 4.100E-01, inv_dt, do_predict_nc>0, do_prescribed_CCN>0 },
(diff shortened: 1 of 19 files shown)


components/scream/src/physics/p3/tests/p3_ice_nucleation_unit_tests.cpp
```diff
@@ -30,28 +30,28 @@ struct UnitWrap::UnitTest<D>::TestIceNucleation {
...
-	  {2.452E+02, 0.950E+00, 0.974E+03, 8.221E+02, 4.952E-02, inv_dt, do_predict_nc, do_prescribed_CCN },
-	  {2.352E+02, 0.950E+00, 0.823E+03, 7.221E+02, 1.952E-02, inv_dt, do_predict_nc, do_prescribed_CCN },
-	  {2.252E+02, 0.950E+00, 0.723E+03, 6.221E+02, 9.952E-02, inv_dt, do_predict_nc, do_prescribed_CCN },
-
-	  {1.990E+02, 1.069E+00, 0.674E+03, 5.221E+01, 6.952E-01, inv_dt, do_predict_nc, do_prescribed_CCN },
-	  {2.952E+02, 1.069E+00, 0.574E+03, 4.221E+01, 3.952E-01, inv_dt, do_predict_nc, do_prescribed_CCN },
-	  {2.852E+02, 1.069E+00, 0.423E+03, 3.221E+01, 1.952E-01, inv_dt, do_predict_nc, do_prescribed_CCN },
-	  {2.702E+02, 1.069E+00, 0.323E+03, 2.221E+01, 9.952E-01, inv_dt, do_predict_nc, do_prescribed_CCN }
+	  {2.106E+02, 8.852E-01, 0.974E+04, 9.221E+03, 5.100E-01, inv_dt, do_predict_nc>0, do_prescribed_CCN>0 },
+	  {2.052E+02, 8.852E-01, 0.874E+04, 8.221E+03, 4.100E-01, inv_dt, do_predict_nc>0, do_prescribed_CCN>0 },
```
bartgol, line 37: It is probably cleaner (and easier to read) to change the for loops to something like:
```
for (bool do_predict_nc : {false, true})
  for (bool do_prescribed_CCN : {false, true})
```
and then use them directly, without the `>0` comparison.

components/scream/src/physics/zm/zm_conv.F90
```diff
@@ -552,12 +540,11 @@ subroutine zm_convr(lchnk   ,ncol    , &
    integer ii
    integer k
    integer msg                      !  ic number of missing moisture levels at the top of model.
-   integer ierror
 
    real(r8) qdifr
    real(r8) sdifr
 
-   logical :: is_first_step
+   logical :: is_first_step = .false.
```
jgfouca, line 547: @AaronDonahue , please look at this line in particular. This logical was being used uninitialized.

bartgol (changes requested):
MUCH cleaner, thanks Jim! I have one suggestion only (see inline comment).

components/scream/src/physics/p3/tests/p3_ice_nucleation_unit_tests.cpp
jgfouca: Agree

jgfouca:
@bartgol , done.

bartgol (approved):

PeterCaldwell:
@jgfouca - thanks for your work on this, I'm super psyched about getting rid of those warnings which were eroding my confidence that we were doing things right and that we'd notice when we did things not right.

PR #820, opened 2020-12-13
bfb_pow for pblintd_height function

xyuan:
bfb_pow for pblintd_height function

[non-BFB]

### DIFF
--- a/components/eam/src/physics/cam/shoc.F90
+++ b/components/eam/src/physics/cam/shoc.F90
@@ -4326,7 +4326,7 @@ subroutine pblintd_height(&
     do k=nlev-1,nlev-npbl+1,-1
        do i=1,shcol
           if (check(i)) then
-             vvk = bfb_pow((u(i,k) - u(i,nlev)), 2.) + bfb_pow((v(i,k) - v(i,nlev)), 2.) + fac*bfb_pow(ustar(i), 2.)
+             vvk = bfb_square(u(i,k) - u(i,nlev)) + bfb_square(v(i,k) - v(i,nlev)) + fac*bfb_square(ustar(i))
              vvk = max(vvk,tiny)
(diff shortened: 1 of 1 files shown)


components/eam/src/physics/cam/shoc.F90
```diff
@@ -4326,7 +4326,7 @@ subroutine pblintd_height(&
     do k=nlev-1,nlev-npbl+1,-1
        do i=1,shcol
           if (check(i)) then
-             vvk = (u(i,k) - u(i,nlev))**2 + (v(i,k) - v(i,nlev))**2 +fac*ustar(i)**2
+             vvk = bfb_pow((u(i,k) - u(i,nlev)), 2.) + bfb_pow((v(i,k) - v(i,nlev)), 2.) + fac*bfb_pow(ustar(i), 2.)
```
tcclevenger, line 4329: Use `bfb_square(x)` instead of `bfb_pow(x,2)`.

tcclevenger (approved):

jeff-cohere (approved):

tcclevenger:
@bartgol I believe this is an example of a Blake failure with no actual tests failing (similar to https://github.com/E3SM-Project/scream/pull/746, but on blake not weaver..) Is this a case where we should manually merge? Or rerun tests?

bartgol:
> @bartgol I believe this is an example of a Blake failure with no actual tests failing (similar to #746, but on blake not weaver..) Is this a case where we should manually merge? Or rerun tests?

@tcclevenger I believe the issue is that your branch does not contain the mods to the scripts file that add netcdf information, so you need to merge master manually in your branch.


Long explanation: This is one limitation of our testing infrastructure: if a merged PR modified the scripts folder, all other prs *need* to merge master in order to see these mods. Our testing scripts _do_ merge master, but only _after_ the testing scripts have been loaded. In other words, the AT never sees the updated version (meaning, with master merged in) of the stuff in the scripts folder, and runs the version that was in your branch when it got cloned.

PR #825, opened 2020-12-14
Allow for bool randomization to be all true

jgfouca:
So, this is not really randomized, but a convenient way to initialize non-random data in the same call that randomizes the other members.


components/scream/src/physics/share/tests/physics_test_data_unit_tests.cpp
```diff
@@ -35,6 +35,20 @@ struct FakeClass1 : public PhysicsTestData
   PTD_STD_DEF(FakeClass1, 5, dim1, dim2, dim3, dim4, scalar);
 };
 
+struct AllTrue : public PhysicsTestData
```
bartgol, line 38: Is there already a use case for this class? Meaning, some code that needs this?

jgfouca: @bartgol , yes, @xyuan asked me for this.

bartgol (approved):

PR #826, opened 2020-12-14
Add some basic netcdf probing to scream CMakeLists.txt

jgfouca:
Leverage existence of nc-config/nf-config in PATH if they are there.

### DIFF
--- a/components/scream/CMakeLists.txt
+++ b/components/scream/CMakeLists.txt
@@ -129,18 +129,28 @@ if (SCREAM_CIME_BUILD)
 endif()
 set(DEFAULT_NUM_VERTICAL_LEV 72)
 
-find_path(DEFAULT_NetCDF_Fortran_PATHS nf-config)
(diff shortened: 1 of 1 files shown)


components/scream/CMakeLists.txt
```diff
@@ -129,6 +129,20 @@ if (SCREAM_CIME_BUILD)
 endif()
 set(DEFAULT_NUM_VERTICAL_LEV 72)
 
+find_path(DEFAULT_NetCDF_Fortran_PATHS nf-config)
+if (DEFAULT_NetCDF_Fortran_PATHS)
+  get_filename_component(DEFAULT_NetCDF_Fortran_PATHS ${DEFAULT_NetCDF_Fortran_PATHS} DIRECTORY)
```
bartgol, line 134: I think the best way to do this should be to run the command `nc-config --prefix`, which should _always_ be safe, regardless of how netcdf decides to organize its installation paths in the future.

bartgol (approved):

PR #827, opened 2020-12-15
set initial pblh to zero to avoid uninitialization issue

xyuan:
set initial pblh to zero to avoid uninitialization issue in corresponding c++ code

[BFB]

### DIFF
--- a/components/eam/src/physics/cam/shoc.F90
+++ b/components/eam/src/physics/cam/shoc.F90
@@ -4428,6 +4428,7 @@ subroutine pblintd_check_pblh(&
     ! PBL to exceed some maximum (npbl) number of allowable points
     !
     do i=1,shcol
+       pblh(i) = 0.0_rtype
        if (check(i)) pblh(i) = z(i,nlevi-npbl)
        pblh(i) = max(pblh(i),700.0_rtype*ustar(i))
     end do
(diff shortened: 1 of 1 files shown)


bogensch:
This is not correct.  Setting pblh to zero here will overwrite the calculation of pblh done in previous subroutine calls in "pblintd" and would have dire consequences on the model solution.  

Though I do see "pblh" is designated as "intent(out)" here when it should be "intent(inout)".

xyuan:
> This is not correct. Setting pblh to zero here will overwrite the calculation of pblh done in previous subroutine calls in "pblintd" and would have dire consequences on the model solution.
> 
> Though I do see "pblh" is designated as "intent(out)" here when it should be "intent(inout)".

yes, it need to be intent(inout) in these subroutines. it has been changed to intent(inout)

xyuan:
> > This is not correct. Setting pblh to zero here will overwrite the calculation of pblh done in previous subroutine calls in "pblintd" and would have dire consequences on the model solution.
> > Though I do see "pblh" is designated as "intent(out)" here when it should be "intent(inout)".
> 
> yes, it need to be intent(inout) in these subroutines. it has been changed to intent(inout)

@bogensch please review the changes, so I can move to get the porting PRs pass the autotest.

bogensch (approved):

xyuan:
@jgfouca could you merge this PR? Peter has approved for the changes.

PR #842, opened 2020-12-18
Fixes to get screamV1 building through CIME.

jgfouca:


jgfouca:
Note, the example testmod I added won't be usable until: E3SM updates its cime submodule and we do an upstream merge.

bartgol (approved):

PR #845, opened 2020-12-21
Fixes memory leak in calcsize_diag

singhbalwinder:
This PR fixes a memory leak by removing "allocate" statement and changes all pointers to allocatable arrays in calsize_diag and wateruptake routines


PeterCaldwell (approved):
Thanks Balwinder!

PR #848, opened 2020-12-22
Need to use kokkos arrays to pass collections of views to kernels

### LINKED ISSUE #846: Several compilation errors in p3_functions_f90.cpp
There are several suspicious uses of std::vector in `p3_functions_f90.cpp`. Take this snippet, around line 1430:

```
      view_1d_ptr_array fluxes_ptr, vs_ptr, qnx_ptr;
      for (int i = 0; i < N; ++i) {
        fluxes_ptr[i] = (const uview_1d*)(&fluxes_d[i]);
        vs_ptr[i]     = (const uview_1d*)(&vs_d[i]);
        qnx_ptr[i]    = (const uview_1d*)(&qnx_d[i]);
      }
      uview_1d urho_d(rho_d), uinv_rho_d(inv_rho_d), uinv_dz_d(inv_dz_d);
      P3F::calc_first_order_upwind_step<N>(urho_d, uinv_rho_d, uinv_dz_d, team, nk, kbot, k_qxtop, kdir, dt_sub, fluxes_ptr, vs_ptr, qnx_ptr);
    });

```
With gcc 8.2.0, I get the following error:

```
p3_functions_f90.cpp:1429:41: error: passing ‘const std::vector<Kokkos::View<ekat::Pack<double, 16>*, Kokkos::LayoutRight, Kokkos::Device<Kokkos::OpenMP, Kokkos::HostSpace> > >’ as ‘this’ argument discards qualifiers [-fpermissive]
       fluxes_ptr[i] = (uview_1d*)(&fluxes_d[i]);
```
The error is odd: if fluxes_d is const, then the const overload of `op[]` should be used. Funny enough, if I replace `fluxes_d[i]` with `fluxes_d.at(i)`, the code compiles.

Anyways, even if the code compiles, it should be wrong, since std::vector is not device-friendly, and yet is used inside a kernel.

@jgfouca I think this comes from your ekat PR that made `host_to_device` and `device_to_host` use std::vector instead of Kokkos::Array. However, if those vectors are used inside a kernel, we need to use a Kokkos::View.

Note: we can still use std::vector in the host<->device routines, but we then need to copy all entries into a view. Something like:
```
view_1d<decltype(fluxes_d[0])>  fluxes_d_view("",N);
auto host_temp = Kokkos::create_host_mirror(fluxes_d_view);
for (int i=0; i<N; ++i) {
  host_temp[i] = fluxes_d[i];
}
Kokkos::deep_copy(fluxes_d_view, host_temp);
```

jgfouca:
Fixes #846

### DIFF
--- a/components/scream/src/physics/p3/p3_functions_f90.cpp
+++ b/components/scream/src/physics/p3/p3_functions_f90.cpp
@@ -1421,14 +1421,21 @@ void calc_first_order_upwind_step_f_impl(
   ekat::host_to_device(ptr_to_arr((const Real**)vs, N)    , nk, vs_d);
   ekat::host_to_device(ptr_to_arr((const Real**)qnx, N)   , nk, qnx_d);
 
+  Kokkos::Array<view_1d, N> fluxes_a, vs_a, qnx_a;
+  for (Int i = 0; i < N; ++i) {
+    fluxes_a[i] = fluxes_d[i];
+    vs_a[i]     = vs_d[i];
+    qnx_a[i]    = qnx_d[i];
+  }
+
   // Call core function from kernel
   auto policy = ekat::ExeSpaceUtils<ExeSpace>::get_default_team_policy(1, nk_pack);
   Kokkos::parallel_for(policy, KOKKOS_LAMBDA(const MemberType& team) {
     view_1d_ptr_array fluxes_ptr, vs_ptr, qnx_ptr;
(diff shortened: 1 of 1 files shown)


bartgol (approved):
I confirm that warnings are gone on weaver, and it also compiles fine with gcc 8.2 with linux-generic mach file.

PR #851, opened 2021-01-05
use cbrt instead of bfb_pow

tcclevenger:
@ambrad Does this make sense to do computationally? It's useful to me since, when running tests, `std::pow` cannot handle negative values, and randomizing inputs can lead to those. However, if it is computationally slower then I can work around with the test inputs.

### DIFF
--- a/components/eam/src/physics/cam/shoc.F90
+++ b/components/eam/src/physics/cam/shoc.F90
@@ -4389,7 +4389,6 @@ subroutine pblintd_surf_temp(&
 
     !===================
     ! const parameter for Diagnosis of PBL depth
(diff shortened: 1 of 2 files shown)


ambrad (approved):
Great! cbrt is definitely preferable to pow.

bartgol:
Regardless of computational costs (this is only for bfb testing anyways, right?), I think cbrt is better: first, it's faster to visually parse, second cpp reference states it can be slightly more accurate. I don't know about performance, but as I said, given that it's just for bfb testing, performance is not a critical factor.

tcclevenger:
@bartgol @ambrad Great! Thanks.

tcclevenger:
@ambrad can I get another approval? Had to push the change to the c++ code..

ambrad (approved):

PR #854, opened 2021-01-06
Quick fix to a build error where the variable "thrd" was not defined in the bfb_math.inc file

AaronDonahue:
Fixes a build error that occurs when building scream through CIME related to a variable not being properly
defined in the bfb_math.inc file.

[BFB]

### DIFF
--- a/components/eam/src/physics/cam/bfb_math.inc
+++ b/components/eam/src/physics/cam/bfb_math.inc
@@ -29,7 +29,7 @@
 #  define bfb_erf(val) cxx_erf(val)
 #else
 #  define bfb_pow(base, exp) (base)**(exp)
-#  define bfb_cbrt(base) (base)**thrd
+#  define bfb_cbrt(base) (base)**(1.0/3.0)
 #  define bfb_gamma(val) gamma(val)
 #  define bfb_log(val) log(val)
 #  define bfb_log10(val) log10(val)
--- a/components/eam/src/physics/cam/bfb_math.inc
+++ b/components/eam/src/physics/cam/bfb_math.inc
@@ -29,7 +29,7 @@
 #  define bfb_erf(val) cxx_erf(val)
 #else
 #  define bfb_pow(base, exp) (base)**(exp)
-#  define bfb_cbrt(base) (base)**(1.0/3.0)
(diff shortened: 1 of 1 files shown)


tcclevenger (approved):

ambrad (changes requested):
I think the explicit type is needed, like this: `(1.0_real_kind/3.0_real_kind)`

AaronDonahue:
@ambrad , yes, thank you for the review.  I went about implementing the requested chage, but this brings up a few questions:
1. "real_kind" doesn't work because that is undefined where it used.  However "rtype" does.  So presumably we should use that.
2.  However, "rtype" just happens to be the variable name used by the three F90 codes that use bfb_math.inc.  In the future, if for any reason we use bfb_math.inc with a different code, that defines real_kind as "r8" for example (as is the case for a big chunk of EAM) we will run into compiler errors which can be a pain to fix.
3. Alternatively, we could simply always use double precision, i.e. 1.0d0/3.0d0.  I'm guessing that this will result in a minor hit to performance when using single precision for everything else.  But maybe that is acceptable (?)

4. Last question, somewhat tangential.  In bfb_math.inc there is also the defined function: 
                              #  define bfb_expm1(val) (exp(val) - 1)
Here we are relying on the compiler casting the "1" as the same type as "val".  Is this safe, or should we be giving the same treatment, i.e 1._rtype or 1.0d0?

AaronDonahue:
Update: Sorry for what seems like free-floating consciousness...

A simpler approach is to just define `thrd` in `shoc.F90`.  Up until now we've only been using cbrt with P3 which has `thrd` explicitly defined in the proper precision.  We can add that to shoc and be done with it.

This of course means that `thrd` needs to be explicitly defined by every module that uses bfb_math.inc.  Which means there is room for a bug, but given that soon we will switch to C++ only and that it's not likely that `thrd` will be defined in a way other than 1/3 it might be the best approach here. (?)

ambrad:
A bare 1 is fine. Main thing is to get 0.3333... to full precision. Since this is for the F90 code only, not for BFB tests, 1.0d0/3.0d0 is fine.

AaronDonahue:
@ambrad , excellent.  Thanks for responding.  I'll push the d0 change.

ambrad (approved):

PR #856, opened 2021-01-12
Port shoc_main

tcclevenger:
Ports `shoc_main` to cpp. Like with P3, `shoc_main()` in `shoc_main_impl.hpp` takes in Int/Scalar constants and all views are grouped in input/output structs, and returns microseconds elapsed in the outer `parallel_for`.

I've made a few comments below.


components/scream/src/physics/shoc/shoc_assumed_pdf_impl.hpp
```diff
@@ -152,13 +152,13 @@ void Functions<S,D>::shoc_assumed_pdf(
                  ekat::min(100,
                            ekat::max(0, (3*tmp_val_1*(1 - a*ekat::square(tmp_val_2) - (1-a)*ekat::square(tmp_val_1))
                                          - (Skew_thl - a*ekat::cube(tmp_val_2) - (1 - a)*ekat::cube(tmp_val_1)))
-                                         /(3*a*(tmp_val_1 - tmp_val_2))*thlsec)));
+                                         /(3*a*(tmp_val_1 - tmp_val_2))))*thlsec);
```
tcclevenger, line 155: This error was not caught in the shoc_assumed_pdf bfb test since the condition below was never triggered. I changed the input grid in the test so that the condition here is met.

components/scream/src/physics/shoc/shoc_assumed_pdf_impl.hpp
```diff
@@ -152,13 +152,13 @@ void Functions<S,D>::shoc_assumed_pdf(
...
       thl2_2.set(condition,
                  ekat::min(100,
                            ekat::max(0, (-3*tmp_val_2*(1 - a*ekat::square(tmp_val_2)
                                          - (1 - a)*ekat::square(tmp_val_1))
                                          + (Skew_thl - a*ekat::cube(tmp_val_2) - (1 - a)*ekat::cube(tmp_val_1)))
-                                         /(3*(1 - a)*(tmp_val_1 - tmp_val_2))*thlsec)));
+                                         /(3*(1 - a)*(tmp_val_1 - tmp_val_2))))*thlsec);
```
tcclevenger, line 161: Same as above.

components/scream/src/physics/shoc/shoc_energy_fixer_impl.hpp
```diff
@@ -50,6 +50,8 @@ void Functions<S,D>::shoc_energy_fixer(
   Scalar se_dis = 0;
 
   // Compute linear interpolation of data into rho_zi
+  // This calculation is needed for stand alone tests,
+  // but it redundant when calling shoc_main
   linear_interp(team,zt_grid,zi_grid,rho_zt,rho_zi,nlev,nlevi,0);
```
tcclevenger, line 55: This can be removed if `shoc_energy_fixer()` is only called from `shoc_main` (and not the stand alone bfb test) since rho_zt is already interpolated in a previous function.

components/scream/src/physics/shoc/tests/shoc_assumed_pdf_tests.cpp
```diff
@@ -408,6 +408,24 @@ struct UnitWrap::UnitTest<D>::TestShocAssumedPdf {
...
+        for (Int k=0; k<d.nlevi; ++k) {
+          const auto zi_k = upper - k*(upper-lower)/(d.nlevi-1);
+          d.zi_grid[k+s*d.nlevi] = zi_k;
+
+          if (k!=d.nlevi-1) {
+            const auto zi_kp1 = upper - (k+1)*(upper-lower)/(d.nlevi-1);
+            d.zt_grid[k+s*d.nlev] = 0.5*(zi_k + zi_kp1);
+          }
+        }
+      }
```
tcclevenger, line 428: See comment in shoc_assumed_pdf_impl.hpp

tcclevenger:
Update: SP shoc_main_bfb fail on Blake and Weaver and pass on Mappy.  SP/DP run_and_cmp pass on all machines. DP shoc_main_bfb pass on all machines.

tcclevenger:
Update: Only weaver, shoc_main_bfb, SP fails..

tcclevenger:
@bogensch In the shoc_main_bfb test, to keep values in `shoc_assumed_pdf` from triggering errors by dipping below 0, I needed to replace a purely random generation of `zi/zt_grid` with a decreasing set of points for each column. Like in the property tests, I set `zt_grid[k] = 0.5*(zi_grid[k]+zi_grid[k+1])`. Is this sufficient? Or should I be testing a more non-uniform zi_grid?

bogensch:
@tcclevenger yes that is sufficient and appropriate.

bogensch (approved):

components/scream/src/physics/shoc/shoc_functions.hpp
```diff
@@ -68,6 +69,112 @@ struct Functions
...
+    view_1d<Pack1d> wqw_sfc;
+    // Surface momentum flux (u-direction) [m2/s2]
+    view_1d<Pack1d> uw_sfc;
+    // Surface momentum flux (v-direction) [m2/s2]
+    view_1d<Pack1d> vw_sfc;
+    // Surface flux for tracers [varies]
+    view_2d<Spack>  wtracer_sfc;
+    // Exner function [-]
+    view_2d<Spack>  exner;
+    // Host model surface geopotential height
```
jeff-cohere, line 104: Do we know the units for this?

jeff-cohere (approved):
Looks good!

components/scream/src/physics/shoc/shoc_functions.hpp
bogensch: [m]

ambrad (approved):
Really nice. For future work:

1. You might see if you can trim down the number of local variables, which is basically workspace, by aliasing and reusing some.
2. You'll want to switch to using the workspace manager instead of declaring new Views before calling shoc_main_internal, as declaring new (memory-owning) Views at runtime is expensive. The workspace manager manages a pool of Views for use as workspace in parameterizations.

components/scream/src/physics/shoc/shoc_functions.hpp
tcclevenger: Everywhere in shoc.F90 its given the same description. Others will know better than me though.. @bogensch

tcclevenger:
>     1. You might see if you can trim down the number of local variables, which is basically workspace, by aliasing and reusing some.
>     2. You'll want to switch to using the workspace manager instead of declaring new Views before calling shoc_main_internal, as declaring new (memory-owning) Views at runtime is expensive. The workspace manager manages a pool of Views for use as workspace in parameterizations.

I agree on each point. After this is merged I will think about an optimization PR that includes these and maybe some of the solver packing/unpacking optimizations from the discussion in https://github.com/E3SM-Project/scream/pull/768

PR #862, opened 2021-01-25
Bug fixes and unit test for SurfaceCoupling class.

bartgol:
This PR fixes errors in the SC class, as well as adds a unit test for it. The UT simply exports and imports a field, checking that the result is equal to the input (at the surface only, obviously).

More still has to be done to make SC fully working in cime runs. In particular, all fields are still not present in the field repo, which causes the SC to throw errors. Skipping fields registration is not possible at the moment, since it would cause errors at run time (I haven't implemented a logic to skip fields, but i should add it at some point, since not all input fields from the cpl are needed by scream).

### DIFF
--- a/components/scream/src/control/atmosphere_driver.cpp
+++ b/components/scream/src/control/atmosphere_driver.cpp
@@ -147,11 +147,6 @@ void AtmosphereDriver::initialize (const ekat::Comm& atm_comm,
   // Initialize atm inputs
   init_atm_inputs ();
 
-  // Now we can inspect the dag, including also checking that
-  // all fields are correctly inited. Any unmet dependency in
-  // the dag at this point has to be treated as an error.
-  inspect_atm_dag ();
-
   // Set time steamp t0 to all fields
   for (auto& field_map_it : *m_field_repo) {
     for (auto& f_it : field_map_it.second) {
--- a/components/scream/src/control/tests/sc_tests.cpp
+++ b/components/scream/src/control/tests/sc_tests.cpp
@@ -1,0 +1,127 @@
+#include "share/grid/point_grid.hpp"
+#include "control/surface_coupling.hpp"
+
+#include <ekat/util/ekat_test_utils.hpp>
+
+#include <catch2/catch.hpp>
+
+TEST_CASE ("surface_coupling")
+{
+  // Some namespaces/aliases
+  using namespace scream;
+  using namespace ShortFieldTagsNames;
+  using namespace ekat::units;
+  using FL = FieldLayout;
+  using FID = FieldIdentifier;
+  using RPDF = std::uniform_real_distribution<Real>;
+  using rngAlg = std::mt19937_64;
+
+  // Some constants
+  constexpr int ncols = 4;
+  constexpr int nlevs = 8;
+  constexpr int nruns = 10;
+
+  // Create a comm
+  ekat::Comm comm (MPI_COMM_WORLD);
+
+  // The random numbers generator
+  std::random_device rd;
+  const unsigned int catchRngSeed = Catch::rngSeed();
+  const unsigned int seed = catchRngSeed==0 ? rd() : catchRngSeed;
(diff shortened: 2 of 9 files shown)


components/scream/src/control/tests/sc_tests.cpp
```diff
@@ -0,0 +1,127 @@
...
+  repo_out.register_field(s2d_id);
+  repo_out.register_field(s3d_id);
+  repo_out.register_field(v3d_id);
+  repo_out.registration_ends();
+
+  // Create two SC objects, to import and export
+  control::SurfaceCoupling importer(grid,repo_in);
+  control::SurfaceCoupling exporter(grid,repo_out);
+
+  importer.set_num_fields(4,0); // Recall that SC counts *scalar* fields, so vector3d counts as 2 fields
```
AaronDonahue, line 62: Could we create an `int` variable that records the count given here as `4`, that way any casual reader of the test is a) able to see the 4 must match the number registered above (+1 for the vector field) and b) is not related to the `ncols` value even farther up.

We could then use the `int` variable instead of hardcoded 4 here.  It also has the bonus that if the test is adjusted in the future to add another field we don't have to track down the hard-coded values too.

components/scream/src/control/tests/sc_tests.cpp
```diff
@@ -0,0 +1,127 @@
...
+  exporter.set_num_fields(0,4);
+
+  // Register fields in the importer/exporter
```
AaronDonahue, line 65: Could we have a brief comment here explaining where the integer arguments passed to register_import/export are coming from?

components/scream/src/control/tests/sc_tests.cpp
```diff
@@ -0,0 +1,127 @@
...
+  importer.register_import("s3d",1);
+  importer.register_import("v3d",2,0);
+  importer.register_import("v3d",3,1);
+  exporter.register_export("s2d",0);
+  exporter.register_export("s3d",1);
+  exporter.register_export("v3d",2,0);
+  exporter.register_export("v3d",3,1);
+
+  // Create a raw array big enough to contain all the 2d data for import/export
+  Real* raw_data = new Real[ncols*4];
```
AaronDonahue, line 76: is the 4 written here related to the total number of variables?  If so, then the `int` variable suggested in the earlier comment could be used here too.

AaronDonahue (commented):
Looks good @bartgol .  I had just a few questions before approving.

components/scream/src/control/atmosphere_driver.cpp
```diff
@@ -147,11 +147,6 @@ void AtmosphereDriver::initialize (const ekat::Comm& atm_comm,
   // Initialize atm inputs
   init_atm_inputs ();
 
-  // Now we can inspect the dag, including also checking that
-  // all fields are correctly inited. Any unmet dependency in
-  // the dag at this point has to be treated as an error.
-  inspect_atm_dag ();
```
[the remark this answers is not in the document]
bartgol: It is, but not here. The method `inspect_atm_dag` is public, and it can be used from outside. In case the SC is on (i.e., for CIME runs) it is called [here](https://github.com/E3SM-Project/scream/blob/7e806b8efcc7f325f631493b2c760efb8760d1e4/components/scream/src/mct_coupling/scream_cxx_f90_interface.cpp#L164).

components/scream/src/control/tests/sc_tests.cpp
bartgol: Sure, I can do this. Note, however, that one will still have to update the variable if the test is updated. But yeah, it is more likely to be remembered that way.

components/scream/src/control/tests/sc_tests.cpp
bartgol: Sure thing.

AaronDonahue (approved):

Issue #866, opened 2021-01-28
ni2nr_melt_tend tendency should be multiplied by nmltratio in conservation calc


PeterCaldwell:
From Koby:

The ni2nr_melt_tend tendency should be multiplied by nmltratio here:
https://github.com/E3SM-Project/scream/blob/1c3cb7dbb629898b218c7a8f36af910a1aecf2bf/components/eam/src/physics/cam/micro_p3.F90#L2951

in order to be consistent with usage in state updates like here: https://github.com/E3SM-Project/scream/blob/1c3cb7dbb629898b218c7a8f36af910a1aecf2bf/components/eam/src/physics/cam/micro_p3.F90#L3126

AaronDonahue:
@PeterCaldwell is this a bug?

I see `ni2nr_melt_tend` used twice when updating prognostic ice:
```
   nr = nr + (-nr_collect_tend-nr2ni_immers_freeze_tend+nmltratio*ni2nr_melt_tend+nr_ice_shed_tend+ncshdc)*dt
   ni = ni + (ni_nucleat_tend-ni2nr_melt_tend-ni_sublim_tend-ni_selfcollect_tend
```
note that for `nr` the `nmltratio` is used.  But for conservation should the source for nr be an equal sink for ni?

Follow-up, if `ncmltratio` is needed for the ice sink part then should it also be added to `ni_conservation` just like this issue is requesting it be added to `nr_conservation`?

PeterCaldwell:
@AaronDonahue - I'm confused by your question about a new bug. I see ni2nr_melt_tend showing up 1x in the nr equation and 1x in the ni equation... as expected for a variable being transferred from the ice to the rain category. What is your concern?

And in terms of nmltratio being included for nr but not ni: I think nmltratio is meant to make each melting snowflake break into several rain drops. I'll have to check the code though.

AaronDonahue:
yeah, the question was just about the use of `nmlratio` in one case but not the other.  So the impact on nr would be different on ni.  But I recognize that number concentration isn't as strict as mass.

I'm only asking because I'm in the code looking at it and want to make sure I don't fix one bug and leave behind another.

PeterCaldwell:
[...]

PR #867, opened 2021-01-28
ni2nr_melt_tend tendency should be multiplied by nmltratio in conserv…

AaronDonahue:
…ation calc

Quick bugfix to how tendencies for rain number conservation are calculated.

Addresses issue #866

This is likely non-BFB at some level as it does adjust the impact of one of the
rain-number concentration tendencies.

[NBFB]

### DIFF
--- a/components/eam/src/physics/cam/micro_p3.F90
+++ b/components/eam/src/physics/cam/micro_p3.F90
@@ -832,7 +832,7 @@ contains
 
       call nc_conservation(nc(k), nc_selfcollect_tend, dt, nc_collect_tend, nc2ni_immers_freeze_tend, &
            nc_accret_tend, nc2nr_autoconv_tend)
-      call nr_conservation(nr(k),ni2nr_melt_tend,nr_ice_shed_tend,ncshdc,nc2nr_autoconv_tend,dt,nr_collect_tend,&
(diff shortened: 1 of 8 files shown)


PeterCaldwell (changes requested):
Needs to be fixed on C++ side too... which should be similarly trivial.

AaronDonahue:
ahh yes, my bad :-)

AaronDonahue:
@PeterCaldwell , I believe this is now ready.  The AT shows fails in p3_run_and_cmp_X, but that should be expected since this is a non-BFB test (right?).  I performed `test-all-scream` on this branch for both syrah and lassen where the baseline is built using the same branch and those pass.  I also tested that the F-Compset will build and run using the SMS_D test.

PeterCaldwell (approved):
Looks great, thanks @AaronDonahue !!!

PeterCaldwell:
Just for posterity - I was creeped out that p3_run_and_cmp* was failing for "cases" 1 and 2 but passing for cases 3 and 4. Is this expected? I did a little sleuthing and found that cases 3 and 4 are predicted_nc=False cases so passing is the correct behavior: we're not allowing nc or ni to change.

Issue #815, opened 2020-12-11
[AD] implement mechanism for passing run options to SCREAMv1


PeterCaldwell:
Right now, the SE dycore expects options to be specified via a namelist. We were planning on having SCREAM use yaml files to specify options (and this seems to be working really well already for output variable requests). It is tempting to use the existing namelist structure from EAM since it has defaults and infrastructure that we could easily co-opt. @rljacob warns us that the eam approach is the result of decades of unplanned accretion of stuff and suggests that creating our own strategy would be better.

What will be a coherent, easy-to-manage strategy for passing runtime options to SCREAMv1 runs? The goal of this issue is to determine a strategy (probably via comments below).

bartgol:
Given the current status, I think the choice is between
 - namelists-only: atm_mct_comp will have to fwd all namelist entries to the driver;
 - mixed nameliists-yaml: still have a big atm_in namelist (easy to fwd to homme), but all the AD parameters are in a yaml file (whose name is either hardcoded, like 'atm_in.yaml', or read from the atm_in namelist).

Namelist pro: there's already a lot of cime infrastructure for creating namelists defaults, and integrating user changes to namelists defaults for a particular case.

YAML pro: it is highly flexible, and does not require the parser to know _a priori_ the structure of the namelist. E.g., you can have a variable number of sublistst.

An important piece for deciding this is whether we can "generate" a yaml file based on case configs (via buildnml), and if we can integrate user changes to the default.

PeterCaldwell:
I generally agree with Luca's assessment but:
1. I think most namelist variables are read directly by atm code rather than passed from atm_mct_comp?
2. using existing cime infrastructure is a benefit, but also a curse because it has a lot of legacy baggage. 
3. I like Luca's last comment about generating a yaml from case configs. Maybe more generally we can say that the big use case we need to handle is the ability to choose a compset and for that to automatically set all the default parameter settings, input files, etc but for the user to be able to override these defaults somehow.

bartgol:
I think generating a yaml file should not be hard (there's a yaml python module that can be used for read/write easily, though it's not standard, so it might not be available everywhere). The interesting piece is how we handle user mods. I don't know how cime setup/build/run works, so I may be saying something naive. If the user mods must be possible _after_ the namelist/yaml-file has been built, it is a bit tricky, but if they must be done before all the inputs are generated, then it is probavly not too hard.

rljacob:
[...]

PR #868, opened 2021-01-28
Add an python utility to ensure pyyaml availability

jgfouca:
SCREAM python coders who have files that need pyyaml should
put ensure_yaml() before their yaml impot. If YAML availability
is not detected, then pip will be used to install it for the
current user only and rehash the sys.path so yaml is available
for import within the same process.

Related to #815

### DIFF
--- a/components/scream/scripts/utils.py
+++ b/components/scream/scripts/utils.py
@@ -2,7 +2,7 @@
 Utilities
 """
 
-import os, sys, re, signal, subprocess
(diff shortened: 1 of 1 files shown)


jgfouca:
An alternative to this approach is to submodule https://github.com/yaml/pyyaml and import it from there. I've confirmed that this works, but without the build phase, you get the "pure python" version which has reduced performance.

bartgol (approved):

PR #869, opened 2021-02-03
Fields subviews

bartgol:
This PR introduces the possibility to create a field as a "subview" of another. The allowed scenarios are relatively limited, but should be enough to accommodate the use cases we need in scream.

In particular, we can
 -  subview a field by slicing its "leading" dimension (the slowest striding one);
 -  subview a field by slicing its second dimension, IF the input field rank is 3 or more (due to limits in kokkos).
 
 Both these operations can be repeated (so one could strip away, say, the first 3 dimensions). But in all likelihood, only one application of the subview will always be enough in scream.
 
Subviewing is as simple as
```
  // subview f1 at 4th entry along 2nd dim, and use f2_id as field identifier for f2
  auto f2 = f1.subfield(1, 3); // Keeps same name
  auto f3 = f1.subfield("new_name",1,3); // Different field name
```

Important detail (I need to improve the documentation): the view stored in f2 is _the same_ view that is stored in f1. The actual subviewing of it only happens during the call to `get_reshaped_view`. This means that if you do `f2.get_view()` you get the same thing that you would get with `f1.get_view()`. This may be important if `f2` is strided (like in this case, because it subviews f1 along the second dimension). In this case, doing something like
```
Kokkos::deep_copy(f2.get_view()),1.0); // likely not what you expected
```
will fill _the whole original view_ with ones, rather than just filling the subview. In order to get the actual subview, use `get_reshaped_view`:

```
auto v2 = f2.get_reshaped_view<Real***>(); // Will throw if f2 has rank!=3.
Kokkos::deep_copy(v2,1.0); // Ok, only the part of f1 corresponding to f2 gets filled with ones.
```

### DIFF
--- a/components/scream/src/share/field/field.hpp
+++ b/components/scream/src/share/field/field.hpp
@@ -16,6 +16,12 @@
 namespace scream
 {
 
+// Enum used when quering Field for a view on a specific mem space
+enum HostOrDevice {
+  Device = 0,
+  Host
+};
+
 template<typename FieldType>
 struct is_scream_field : public std::false_type {};
 
@@ -42,20 +48,10 @@ public:
   template<typename ViewT>
   using HM = typename ViewT::HostMirror;
 
-  // Given a type Space, check if it is the device mem/exec space,
-  // or even the device itself.
-  template<typename Space>
-  struct OnDevice {
-    static constexpr bool value =
-          std::is_same<Space,device_type::memory_space>::value ||
-          std::is_same<Space,device_type::execution_space>::value ||
-          std::is_same<Space,device_type>::value;
-  };
-
   // Given a device view type V and a mem/exec space, returns V if the space
   // is on device, otherwise returns the host mirror of V
-  template<typename DevViewT, typename Space>
-  using get_view_type = cond_t<OnDevice<Space>::value,DevViewT,HM<DevViewT>>;
+  template<typename DevViewT, HostOrDevice HD>
+  using get_view_type = cond_t<HD==Device,DevViewT,HM<DevViewT>>;
 
   template<typename DT>
   using view_type = typename kt::template view<DT>;
@@ -116,9 +112,9 @@ public:
         header_type& get_header ()       { return *m_header; }
   const std::shared_ptr<header_type>& get_header_ptr () const { return m_header; }
 
-  template<typename Space = device_type::memory_space>
-  const get_view_type<view_type<RT*>,Space>&
-  get_view () const { return  get_view_impl<Space>();   }
+  template<HostOrDevice HD = Device>
+  const get_view_type<view_type<RT*>,HD>&
+  get_view () const { return  get_view_impl<HD>();   }
 
   // Returns a const_field_type copy of this field
   const_field_type get_const () const { return const_field_type(*this); }
@@ -141,8 +137,8 @@ public:
   // The class will check that the requested data type is compatible with the
   // allocation. This allows each field to be stored as a 1d array, but then
   // be reshaped to the desired layout before being used.
-  template<typename DT, typename Space = device_type::memory_space>
-  get_view_type<uview_type<DT>,Space>
+  template<typename DT, HostOrDevice HD = Device>
+  get_view_type<uview_type<DT>,HD>
   get_reshaped_view () const;
 
   // If someone needs the host view, some sync routines might be needed.
@@ -180,16 +176,16 @@ public:
 
 protected:
 
-  template<typename Space>
-  const if_t<OnDevice<Space>::value,view_type<RT*>>&
+  template<HostOrDevice HD>
+  const if_t<HD==Device,view_type<RT*>>&
   get_view_impl   () const {
     EKAT_REQUIRE_MSG (is_allocated (),
         "Error! View was not yet allocated.\n");
     return  m_view_d;
   }
 
-  template<typename Space>
-  const if_t<!OnDevice<Space>::value,HM<view_type<RT*>>>&
+  template<HostOrDevice HD>
+  const if_t<HD==Host,HM<view_type<RT*>>>&
   get_view_impl   () const {
     ensure_host_view ();
     return  m_view_h;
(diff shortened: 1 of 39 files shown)


bartgol:
This PR is adding the subfield feature, but nowhere in scream (except unit tests) it is used. I would still argue that it is best to merge this now, and work on adding subfield in the FieldRepository (for tracers) in a second PR. I'm afraid I might start to touch quite a bit of code once I get out of the little nutshell where this feature lives.

Also, I need to wait for #E3SM-Project/EKAT81 to be merged, before I can merge this PR (assuming we're ok with updating EKAT within this PR...).

components/scream/src/share/tests/dummy_se_point_remapper.hpp
```diff
@@ -44,40 +44,46 @@ class DummySEPointRemapper : public AbstractRemapper<RealType>
...
+        return FieldLayout (TV{COL},IV{ncol});
       case 4:
-        return FieldLayout ({COL,VL},{ncol,tgt.dim(3)});
+        return FieldLayout (TV{COL,VL},IV{ncol,tgt.dim(3)});
       case 5:
-        return FieldLayout ({COL,CMP,VL},{ncol,tgt.dim(1),tgt.dim(4)});
+        return FieldLayout (TV{COL,CMP,VL},IV{ncol,tgt.dim(1),tgt.dim(4)});
       default:
-        return FieldLayout ({},{});
+        return FieldLayout (TV{},IV{});
```
jeff-cohere, line 63: Is this a meaningful field layout? Or does it indicate an error?

components/scream/src/share/tests/dummy_se_point_remapper.hpp
```diff
@@ -44,40 +44,46 @@ class DummySEPointRemapper : public AbstractRemapper<RealType>
...
+        return FieldLayout (TV{EL,GP,GP},IV{nele,4,4});
       case 2:
-        return FieldLayout ({EL,GP,GP,src.tag(1)},{nele,4,4,src.dim(1)});
+        return FieldLayout (TV{EL,GP,GP,src.tag(1)},IV{nele,4,4,src.dim(1)});
       case 3:
-        return FieldLayout ({EL,CMP,GP,GP,VL},{nele,src.dim(1),4,4,src.dim(2)});
+        return FieldLayout (TV{EL,CMP,GP,GP,VL},IV{nele,src.dim(1),4,4,src.dim(2)});
       default:
-        return FieldLayout ({},{});
+        return FieldLayout (TV{},IV{});
```
jeff-cohere, line 86: Same Q as line 63.

components/scream/src/share/field/field_tracking.hpp
```diff
@@ -61,8 +94,14 @@ class FieldTracking {
   void add_to_group (const std::string& group_name);
 
   // Set the time stamp for this field. This can only be called once, due to TimeStamp implementation.
+  // NOTE: if the field has 'children' (see below), their ts will be udpated too.
```
jeff-cohere, line 97: udpated -> updated

components/scream/src/share/field/field_tracking.hpp
```diff
@@ -28,8 +55,10 @@ class FieldTracking {
   using atm_proc_set_type = ekat::WeakPtrSet<AtmosphereProcess>;
 
   FieldTracking() = delete;
-  FieldTracking(const std::string&);
+  FieldTracking(const std::string& name);
```
jeff-cohere, line 58: worth using `explicit` here?

components/scream/src/share/tests/dummy_se_point_remapper.hpp
bartgol: The check at the top should throw an error before you hit this case. I think I added it to silence compiler warnings on CUDA maybe? I recall nvcc complaining for a fcn with no return if the default case was hit. I guess some compilers might not be smart enough to figure out that if you'd hit the default case, you'd have already errored out earlier on.

components/scream/src/share/field/field_tracking.hpp
bartgol: Good point.

components/scream/src/share/field/field.hpp
```diff
@@ -200,45 +263,89 @@ template<typename RealType>
...
+Field<RealType> Field<RealType>::
+subfield (const std::string& sf_name, const int idim, const int k) const {
+  const auto& id = m_header->get_identifier();
+  const auto& lt = id.get_layout();
+
+  // Sanity checks
+  EKAT_REQUIRE_MSG (is_allocated(),
+      "Error! Input field must be allocated in order to subview it.\n");
+  EKAT_REQUIRE_MSG (idim==0 || idim==1,
+        "Error! Subview dimension index must be either 0 or 1.\n");
```
jeff-cohere, line 328: This is to simplify the implementation in terms of assumptions made to provide padding for layouts, right? Not that it will surprise anyone, but we should probably document this limitation above the declaration of this method.

jeff-cohere (approved):
There's a lot of stuff here. I think the simplified FieldIdentifiers and the more specific FieldLayouts in the AbstractGrid interface will help people understand things better. I agree we should get these changes in with the knowledge that things will probably evolve a bit as they start being used.

components/scream/src/share/field/field.hpp
bartgol: You mean the idim=0,1 constraint? The reason is that if you subview at idim=2+, then there is no way to keep LayoutRight. In Kokkos, LayoutRight allows to store a single stride for the slowest dim, without hurting performance at all (LR computes offsets like `offset = i7 + dim(6)*(i6+dim(5)* (...)) + stride*i0`, where `stride=dim(1)*...*dim(7)` for contiguous views).

If you slice with idim=2, you can still fit in this offset strategy, but if you slice with idim=2+, you can't, and need LayoutStride. Besides the perf loss, I don't think we'd be able to implement it, since the return type would be different.

I can document better in the Field class, where the function is declared. Better yet, make the fcn private, and use a friend free function in field_utils.hpp, with ample documentation (the file field.hpp is too crowded).

bartgol:
@jeff-cohere I added host view capability to Field. In particular:

 - the field stores both a device and a host (1d) view.
 - both `get_view` and `get_reshaped_view` can accept a new template arg, representing the device or mem/exec space where the view is desired (defaults to device, for bwd compatibility).
 - `sync_to_host` and `sync_to_dev` allow to ensure the two are synced.
 - host mirror view is not created until needed, that is, if one of the sync methods is called, or one of the get view methods (with Host space template arg).
  - some internal typedefs have been changed a bit.
 
I also added some meta-utils at the top of the class. Some are a bit more general than just-for-Field, so at some point they might be factored out.

Tests pass on my laptop, I'll let the AT test on our machines.

jeff-cohere:
Looks like the AT machines might have had some trouble configuring SCREAM with the latest changes?

bartgol:
> Looks like the AT machines might have had some trouble configuring SCREAM with the latest changes?

I just figured out the pb (10 seconds ago). It looks like cmake's generator-expression, while recognizing `C_COMPILER_ID` and `CXX_COMPILER_ID`, does not recognize `Fortran_COMPILER_ID`, so this line (in EKAT's `EkatUtils.cmake`) is buggy:

```
  target_compile_options (${targetName} PRIVATE
    $<$<COMPILE_LANGUAGE:Fortran>:$<$<Fortran_COMPILER_ID:GNU>:-w> $<$<Fortran_COMPILER_ID:Intel>: -w>>)
```
I found online workarounds to emulate this line, but I also found that cmake 14+ does recognize that variable. So now I'm wondering whether to check if cmake 3.14 is available or to do the emulation, which would be something like
```
set (Fortran_COMPILER_ID ${CMAKE_Fortran_COMPILER_ID})
...

  target_compile_options (${targetName} PRIVATE
    $<$<COMPILE_LANGUAGE:Fortran>:$<$<STREQUAL:"${Fortran_COMPILER_ID}","GNU">:-w> $<$<STREQUAL:"${Fortran_COMPILER_ID}","Intel"l>: -w>>)
```

jeff-cohere:
What a tangled web we weave... :-)

jeff-cohere (approved):
Well done! Let's see how this plays in SCREAM! We'll definitely be trying it with EAGLES too, hopefully later in 2021.

Issue #864, opened 2021-01-26
[AD] Remaining tasks for P3-AD interface


AaronDonahue:
There are two remaining tasks to resolve before P3 is "fully" integrated into the AD:

- [x] post-processing of P3 output to prepare variables used by other processes.
  The pre-processing is already put in, but there are a number of steps that must follow the call to p3_main to update variables used by SHOC and Radiation.
- [ ] A more robust unit test.
  The current unit test just uses a text file input that is derived from p3_run_and_cmp output.  A more robust unit test - and p3_initializer setup - would be to utilize netCDF input and allow for pass/fail based on cprnc.

bartgol:
@AaronDonahue I'm tempted to call this issue solved. The remaining half part, namely to get cprnc-based tests, is probably more general, and has nothing to do with p3-ad interfaces.

AaronDonahue:
Yeah, we can probably call this resolved now that I've done a pretty thorough analysis of the p3-ad values compared with p3_run_and_cmp.

PR #870, opened 2021-02-03
P3 post-amble processing of variables

AaronDonahue:
Introduce post-processing of the output variables from P3.

This commit introduces a structure to the p3 interface header that similar to the pre-processing structure already in the code.  Typically, an atmospheric process does not produce variables needed by other processes in an appropriate format.  P3 is no exception.

This commit introduces a structure to the p3_interface header file which will update field managed variables that are needed in the next timestep and/or by other processes.  This action tracks with what is already done in micro_p3_interface.F90 in the EAM version of the code.

There are a number of variables that are missing, but noted in the comments.  It is likely that as new processes, such as SHOC and Radiation, are added we will need to update the post-processing code to provide the values these processes need.  But instead of guessing what those might be, this PR only focuses on the variables that are needed by Dynamics and P3.

This PR also fixes a bug that was discovered in how the `zi` variable is initialized in the p3_inputs_intializer code.

Addresses one part of #864 .

NBFB for p3_stand_alone test - doesn't touch the actual P3 code.

### DIFF
--- a/components/scream/src/physics/p3/atmosphere_microphysics.hpp
+++ b/components/scream/src/physics/p3/atmosphere_microphysics.hpp
@@ -74,6 +74,7 @@ public:
   // which can then be called during run_imple in the main .cpp code.
   // Structure to handle the local generation of data needed by p3_main in run_impl
   struct run_local_vars {
+    run_local_vars() = default; 
     run_local_vars(const int ncol, const int npack, view_2d_const pmid_, view_2d T_atm_, view_2d_const ast_, view_2d_const zi_) : 
       m_ncol(ncol),
       m_npack(npack),
@@ -92,10 +93,10 @@ public:
     {
       // Nothing else to initialize at the moment.
     };
+    // Functor for Kokkos loop to pre-process every run step
     KOKKOS_INLINE_FUNCTION
     void operator()(const int icol) const {
-      int ipack, ivec;
-      for (ipack=0;ipack<m_npack;ipack++) {
+      for (int ipack=0;ipack<m_npack;ipack++) {
         // Exner
         const Spack opmid(pmid(icol,ipack));
         const Smask opmid_mask(!isnan(opmid) and opmid>0.0);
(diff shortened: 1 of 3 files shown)


AaronDonahue:
Switching to WIP until I can figure out what causes the fails on weaver.

AaronDonahue:
@bartgol and @jeff-cohere , I found the bug and fixed it.  I think this PR is ready for review.

components/scream/src/physics/p3/atmosphere_microphysics.hpp
```diff
@@ -72,66 +72,54 @@ class P3Microphysics : public AtmosphereProcess
   // Most individual processes have a pre-processing step that constructs needed variables from
   // the set of fields stored in the field manager.  A structure like this defines those operations,
   // which can then be called during run_imple in the main .cpp code.
+  // NOTE: the use of the ekat command "set" to copy local values into the variables of interest.
```
jeff-cohere, line 75: Here, "set" ensures that only masked entries in a pack are copied from one array to the other. There's no circumstance in which an array assignment would lead to two arrays having elements that shared the same memory.

jeff-cohere: That said, what you're doing now is better than before. You were retrieving a partially-populated pack using `get_exner` and then copying the whole thing (uninitialized entries and all) to `exner(icol,ipack)`.

components/scream/src/physics/p3/atmosphere_microphysics.hpp
```diff
@@ -144,7 +132,86 @@ class P3Microphysics : public AtmosphereProcess
...
+  // p3_main.
+  struct p3_postamble {
+    p3_postamble() = default; 
+    // Functor for Kokkos loop to pre-process every run step
+    KOKKOS_INLINE_FUNCTION
+    void operator()(const int icol) const {
+      for (int ipack=0;ipack<m_npack;ipack++) {
+        // Update the atmospheric temperature and the previous temperature.
+        const Spack oexner(exner(icol,ipack));
+        const Smask oexner_mask(!isnan(oexner) and oexner>0.0);
```
jeff-cohere, line 170: Neither here nor there, but if you like the English word versions of the logical operators, you can also use `not` instead of `!`.

jeff-cohere (approved):
Looks good. I left some comments for you to consider regarding how to explain what your use of `set` is accomplishing with the Exner values.

PR #873, opened 2021-02-10
Create tracers as subviews of a large Q array

bartgol:
This PR introduce some mods to a bunch of the AD classes (field stack, atm proc stack, driver). The main goal is to allow the allocation of the tracers group (so far called "ADVECTED") as a single field `Q`, and then 'subview' each individual tracer from `Q`.

Note: This is sitll a WIP (e.g., no tests yet), but I'm not adding the WIP label, so that the AT will help me find GPU errors as I'm developing, so I can fix them while I'm still working on it.

### DIFF
--- a/components/scream/src/control/atmosphere_driver.cpp
+++ b/components/scream/src/control/atmosphere_driver.cpp
@@ -126,11 +126,11 @@ void AtmosphereDriver::initialize (const ekat::Comm& atm_comm,
   // Set all groups of fields
   for (const auto& it : m_atm_process_group->get_required_groups()) {
     auto group = m_field_repo->get_const_field_group(it.first,it.second);
-    m_atm_process_group->set_required_group(it,group);
+    m_atm_process_group->set_required_group(group);
   }
   for (const auto& it : m_atm_process_group->get_updated_groups()) {
     auto group = m_field_repo->get_field_group(it.first,it.second);
-    m_atm_process_group->set_updated_group(it,group);
+    m_atm_process_group->set_updated_group(group);
   }
 
   // Initialize the processes
@@ -150,7 +150,7 @@ void AtmosphereDriver::initialize (const ekat::Comm& atm_comm,
   // Set time steamp t0 to all fields
   for (auto& field_map_it : *m_field_repo) {
     for (auto& f_it : field_map_it.second) {
-      f_it.second.get_header().get_tracking().update_time_stamp(t0);
+      f_it.second->get_header().get_tracking().update_time_stamp(t0);
     }
   }
 
@@ -229,23 +229,23 @@ void AtmosphereDriver::register_groups () {
   using ci_string = ekat::CaseInsensitiveString;
   using ci_string_pair = std::pair<ci_string,ci_string>;
 
-  // Given a list of group-grid pairs (A,B), make sure there is a copy of A
-  // on grid B registered in the repo.
+  // Given a list of group-grid pairs (A,B), make sure there is a copy
+  // of each field in group A on grid B registered in the repo.
   auto lambda = [&](const std::set<ci_string_pair>& groups_grids) {
-    const auto& groups_to_fnames = m_field_repo->get_field_groups_names();
+    const auto& groups_info = m_field_repo->get_groups_info();
+
     for (const auto& gg : groups_grids) {
       const auto& group = gg.first;
-      const auto& grid  = gg.second;
-      EKAT_REQUIRE_MSG(groups_to_fnames.find(group)!=groups_to_fnames.end(),
-        "Error! Group '" << group << "' not found in the repo.\n");
+      const auto& grid = gg.second;
 
-      const auto& fnames = groups_to_fnames.at(group);
-      for (const auto& name : fnames) {
-        auto aliases = m_field_repo->get_aliases(name);
+      // Lambda helper fcn, that register field $name with group $group on grid $grid
+      // if not yet already registered
+      auto register_if_not_there = [&](const std::string& name) {
+        const auto& aliases = m_field_repo->get_aliases(name);
         EKAT_REQUIRE_MSG(aliases.size()>0,
           "Error! Something went wrong while looking for field '" << name << "' in the repo.\n");
 
-        // Check if a copy of this field on the right grid already is registered.
+        // Check if a copy of this field on the right grid is already registered.
         bool found = false;
         for (const auto& alias : aliases) {
           if (alias.first.get_grid_name()==grid) {
@@ -268,6 +268,25 @@ void AtmosphereDriver::register_groups () {
           FieldIdentifier new_fid(name,tgt_layout,f_units,grid);
           m_field_repo->register_field(new_fid,group);
         }
+      };
+
+      auto group_it = groups_info.find(group);
+      EKAT_REQUIRE_MSG(group_it!=groups_info.end(),
+        "Error! Group '" << group << "' not found in the repo.\n");
+
+      const auto& fnames = group_it->second->m_fields_names;
+      for (const auto& name : fnames) {
+        register_if_not_there(name);
+      }
+
+      if (group_it->second->m_bundled) {
+        // The group was allocated as a single bundled field, with each
+        // field in the group later subviewing the bundle.
+        // We need to ensure the bundle also exists on $grid
+        const auto& name = *fnames.begin();
+        const auto& f = m_field_repo->get_field(name,grid);
+        const auto& bundle_name = f.get_header().get_parent().lock()->get_identifier().name();
+        register_if_not_there(bundle_name);
       }
     }
   };
@@ -284,12 +303,12 @@ void AtmosphereDriver::init_atm_inputs () {
     // output of a remapper).
     auto& aliases = m_field_repo->get_aliases(input.name());
     for (auto& it : aliases) {
-      auto& f = it.second;
-      auto& hdr = f.get_header_ptr();
-      auto& id = hdr->get_identifier();
-      auto init_type = hdr->get_tracking().get_init_type();
-      if (init_type!=InitType::None) {
-        auto initializer = hdr->get_tracking().get_initializer().lock();
+      const auto& f = *it.second;
+      const auto& hdr = f.get_header();
+      const auto& id = hdr.get_identifier();
+      auto init_type = hdr.get_tracking().get_init_type();
+      if (init_type!=InitType::None && init_type!=InitType::Inherited) {
+        auto initializer = hdr.get_tracking().get_initializer().lock();
         EKAT_REQUIRE_MSG (static_cast<bool>(initializer),
                             "Error! Field '" + id.name() + "' has initialization type '" + e2str(init_type) + "',\n" +
                             "       but its initializer pointer is not valid.\n");
--- a/components/scream/src/control/atmosphere_driver.cpp
+++ b/components/scream/src/control/atmosphere_driver.cpp
@@ -110,7 +110,7 @@ void AtmosphereDriver::initialize (const ekat::Comm& atm_comm,
   m_field_repo->registration_begins();
   m_atm_process_group->register_fields(*m_field_repo);
   register_groups();
-  m_field_repo->registration_ends();
+  m_field_repo->registration_ends(m_grids_manager);
 
   // Set all the fields in the processes needing them (before, they only had ids)
   // Input fields will be handed to the processes as const
--- a/components/scream/src/control/atmosphere_driver.cpp
+++ b/components/scream/src/control/atmosphere_driver.cpp
@@ -226,7 +226,7 @@ void AtmosphereDriver::register_groups () {
 
   // Given a list of group-grid pairs (A,B), make sure there is a copy
   // of each field in group A on grid B registered in the repo.
-  auto lambda = [&](const std::set<GroupRequest>& groups_grids) {
+  auto has_group_fields_on_grid = [&](const std::set<GroupRequest>& groups_grids) {
     const auto& groups_info = m_field_repo->get_groups_info();
 
     for (const auto& gg : groups_grids) {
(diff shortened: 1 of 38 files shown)


bartgol:
@AaronDonahue @jeff-cohere I think this PR is getting somewhat ready for review. I will go through it and maybe add more tests/checks, but the bulk is there.

bartgol:
I tried to make some atm procs use the 'bundled' tracers feature, and it uncovered some issues. One I didn't fix it in this PR (since no current test fails), and opened issue #876 to track it.

The bundled tracers feature has several ramifications, and I believe it's best to merge this (if tests pass and reviewers approve), so that other work (e.g., SHOC interface) can move forward using this feature. Small-ish steps.

jeff-cohere:
Nice! I'm scrambling today to prep materials for a meeting, but I'll try to look at this tomorrow.

components/scream/src/control/atmosphere_driver.cpp
```diff
@@ -226,29 +222,29 @@ void AtmosphereDriver::finalize ( /* inputs? */ ) {
 }
 
 void AtmosphereDriver::register_groups () {
-  using ci_string = ekat::CaseInsensitiveString;
-  using ci_string_pair = std::pair<ci_string,ci_string>;
+  using GroupRequest = AtmosphereProcess::GroupRequest;
+
+  // Given a list of group-grid pairs (A,B), make sure there is a copy
+  // of each field in group A on grid B registered in the repo.
+  auto lambda = [&](const std::set<GroupRequest>& groups_grids) {
```
jeff-cohere, line 229: Maybe you could call this `has_registered_field_copy` (or something) instead of `lambda`, given how you've named the lambda helper below.

components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
```diff
@@ -44,30 +44,20 @@ void HommeDynamics::set_grids (const std::shared_ptr<const GridsManager> grids_m
...
-  const auto dyn_grid = grids_manager->get_grid(dyn_grid_name);
-  const int ne = dyn_grid->get_num_local_dofs()/(NGP*NGP);
+  m_dyn_grid = grids_manager->get_grid(dyn_grid_name);
+  const int ne = m_dyn_grid->get_num_local_dofs()/(NGP*NGP);
 
   // Sanity check for the grid. This should *always* pass, since Homme builds the grids
   EKAT_REQUIRE_MSG(get_num_local_elems_f90()==ne,
-                     "Error! The number of elements computed from the Dynamis grid num_dof()\n"
-                     "       does not match the number of elements internal in Homme.\n");
+      "Error! The number of elements computed from the Dynamis grid num_dof()\n"
```
jeff-cohere, line 59: Dynamics

components/scream/src/dynamics/homme/atmosphere_dynamics.hpp
```diff
@@ -47,9 +47,14 @@ class HommeDynamics : public AtmosphereProcess
   // Register all fields in the given repo
   void register_fields (FieldRepository<Real>& field_repo) const;
 
+  void set_required_group (const FieldGroup<const Real>& group);
```
jeff-cohere, line 50: Worth explaining in a comment that "TRACERS TENDENCY" is the required group, and "TRACERS" is the updated group for dynamics?

components/scream/src/physics/p3/p3_inputs_initializer.cpp
```diff
@@ -96,62 +96,37 @@ void P3InputsInitializer::initialize_fields ()
...
-  auto d_qr              = m_fields.at("qr").get_reshaped_view<Pack**>();
-  auto d_qi              = m_fields.at("qi").get_reshaped_view<Pack**>();
-  auto d_qm              = m_fields.at("qm").get_reshaped_view<Pack**>();
-  auto d_nc              = m_fields.at("nc").get_reshaped_view<Pack**>();
-  auto d_nr              = m_fields.at("nr").get_reshaped_view<Pack**>();
-  auto d_ni              = m_fields.at("ni").get_reshaped_view<Pack**>();
-  auto d_bm              = m_fields.at("bm").get_reshaped_view<Pack**>();
-  auto d_nccn_prescribed = m_fields.at("nccn_prescribed").get_reshaped_view<Pack**>();
-  auto d_inv_qc_relvar   = m_fields.at("inv_qc_relvar").get_reshaped_view<Pack**>();
+  auto h_T_atm           = m_fields.at("T_atm").get_reshaped_view<Pack**,Host>();
```
jeff-cohere, line 99: Is the comment above now incorrect?

components/scream/src/physics/shoc/atmosphere_macrophysics.hpp
```diff
@@ -49,9 +49,12 @@ class SHOCMacrophysics : public scream::AtmosphereProcess
   // Register all fields in the given repo
   void register_fields (FieldRepository<Real>& field_repo) const;
 
-  // Get the set of required/computed fields
+  void set_updated_group (const FieldGroup<Real>& group);
```
jeff-cohere, line 52: Worth mentioning "TRACERS" here?

components/scream/src/physics/zm/zm_inputs_initializer.cpp
```diff
@@ -243,15 +243,14 @@ void ZMInputsInitializer :: initialize_fields(){
...
-    Kokkos::View<Real*, Kokkos::LayoutRight, DefaultDevice> d_v = m_fields.at(zm_inputs[i]).get_view();
-    //Create and store host mirrors using device views
-    Kokkos::View<scream::Real*, Kokkos::LayoutRight, HostDevice> h_m = Kokkos::create_mirror_view(d_v);
-    //Create and store host mirrors raw pointers
-    // Real* r_p = h_m.data();
-    //Deep copy back to device
-    Kokkos::deep_copy(d_v, h_m);
+    const auto& f = m_fields.at(name);
+    
+    // TODO: Do something to init fields
```
jeff-cohere, line 250: This isn't something that has to be done in this PR, is it?

components/scream/src/share/atm_process/atmosphere_process.hpp
```diff
@@ -60,9 +61,26 @@ namespace scream
...
+    }
+
+    // Group name
+    ci_string name;
+    // Grid name
+    ci_string grid;
+    // Request an allocation that can accomodate a value type like Pack<Real,pack_size>
+    int       pack_size;
+
+    friend bool operator<(const GroupRequest&, const GroupRequest&);
```
jeff-cohere, line 82: Is this just for sorting requests?

components/scream/src/share/atm_process/atmosphere_process.hpp
```diff
@@ -60,9 +61,26 @@ namespace scream
 class AtmosphereProcess : public ekat::enable_shared_from_this<AtmosphereProcess>
 {
 public:
-  using TimeStamp       = util::TimeStamp;
-  using ci_string       = ekat::CaseInsensitiveString;
-  using ci_string_pair  = std::pair<ci_string,ci_string>; // Group and grid names
+  using TimeStamp      = util::TimeStamp;
+  using ci_string      = ekat::CaseInsensitiveString;
+
+  // A group request allow to request a group of fields.
```
jeff-cohere, line 67: Worth mentioning how GroupRequests are used? Okay if not--I'm just wondering if there's an easy example.

components/scream/src/share/atm_process/atmosphere_process.hpp
```diff
@@ -224,6 +258,22 @@ class AtmosphereProcess : public ekat::enable_shared_from_this<AtmosphereProcess
   TimeStamp t_;
 };
 
+// To allow using GroupRequest in std sorted containers
```
jeff-cohere, line 261: You answered my question!

components/scream/src/share/field/field_alloc_prop.cpp
```diff
@@ -85,16 +85,35 @@ subview (const int idim, const int k) const {
...
+
+  ekat::error::runtime_check(!m_committed, "Error! Cannot change allocation properties after they have been commited.\n");
+
+  const int vts = scalar_size*pack_size;
+  if (m_scalar_type_size==0) {
+    // This is the first time we receive a request. Set the scalar type properties
+    m_scalar_type_size = scalar_size;
+  } else {
+    // Make sure the new scalar_type coincides with the one already stored (check name and size)
+    runtime_check(scalar_size==m_scalar_type_size,
```
jeff-cohere, line 100: Is this the same function as in line 92?

components/scream/src/share/field/field_repository.hpp
```diff
@@ -160,6 +173,11 @@ template<typename RequestedValueType>
 void FieldRepository<RealType>::
 register_field (const identifier_type& id, const std::set<std::string>& groups_names) {
 
+  EKAT_REQUIRE_MSG (m_repo_state!=RepoState::Clean,
+      "Error! Repo state is not 'Open' yet. You must call registration_begins() first.\n");
+  EKAT_REQUIRE_MSG (m_repo_state!=RepoState::Closed,
+      "Error! Repo state is not 'Open' anumore. You must register field before calling registration_ends().\n");
```
jeff-cohere, line 179: anymore

components/scream/src/share/field/field_repository.hpp
```diff
@@ -232,72 +267,118 @@ has_field (const identifier_type& identifier) const {
...
+  auto it = m_field_groups.find(group_name);
+  if (it!=m_field_groups.end()) {
+    group.m_info = it->second;;
+    for (const auto& fname : group.m_info->m_fields_names) {
+      auto f = get_field_ptr(fname,grid_name);
+      group.m_fields[fname] = f;
+      // Fetch the bundle fields (if bundled) just once
+      if (group.m_info->m_bundled && group.m_bundle==nullptr) {
+        auto p = f->get_header().get_parent().lock();
+        EKAT_REQUIRE_MSG(p!=nullptr, "Error! Something is amiss with a bundled field group.\n");
```
jeff-cohere, line 307: Maybe "a bundled field is missing its parent"?

components/scream/src/share/field/field_repository.hpp
```diff
@@ -307,11 +388,193 @@ void FieldRepository<RealType>::registration_begins () {
 }
 
 template<typename RealType>
-void FieldRepository<RealType>::registration_ends () {
-  // Proceed to allocate fields
-  for (auto& map : m_fields) {
-    for (auto& it : map.second) {
-      it.second.allocate_view();
+void FieldRepository<RealType>::
+registration_ends (const std::shared_ptr<const GridsManager>& gm) {
```
jeff-cohere, line 392: Are we requiring a grids manager here because it may not have been created before registration began?

jeff-cohere (changes requested):
Well done, sir! I really like the more specific data structures and their guarantees. I've left several comments, mostly to do with documentation, but also a few conceptual questions. I'm not requesting changes so much as consideration. :-)

Nice work!

components/scream/src/physics/zm/zm_inputs_initializer.cpp
bartgol: No. I noticed that this chunk of code (even before my mods) was not actually init-ing anything. The person who did the zm work (which was a fair amount of work, tbh) took another job, so probably she never got to this. Or maybe zm does not really need to init fields, I'm not sure.

components/scream/src/share/atm_process/atmosphere_process.hpp
bartgol: Yes, we need to overload op< so that we can stuff GroupRequest's inside an std::set. I'll add a comment so it's clear why it's there.

bartgol: Actually, since it's a struct, everything is public, so this friendship declaration is superfluous/misleading. I'll purge it, and put the comment above the op< declaration.

components/scream/src/share/atm_process/atmosphere_process.hpp
bartgol: Right, we don't have a stand alone example, only the usage of GroupRequest's in shoc and homme. I'll add some more comments here.

components/scream/src/share/field/field_alloc_prop.cpp
bartgol: Ah, yes. I'll remove the using directive, and do fully qualified namespace call.

bartgol: Btw, we have both `EKAT_REQUIRE_MSG` (a macro) as well as `runtime_check` (a function) in ekat. If the check fails, the former throws an exception, while the latter aborts (finalizing kokkos and MPI, which will likely end up causing kokkos errors when views are deallocated at exit, after kokkos finalization). Perhaps at some point we can decide we're ok with just one, and remove the other altogether. I'm for keeping the exception macro and purging the abort fcn.

components/scream/src/share/field/field_repository.hpp
bartgol: Yeah, it can be phrased better.

components/scream/src/share/field/field_repository.hpp
bartgol: I originally modified the ctor of FieldRepository, so that one would have to pass a GM there. But I changed my mind, because it is only used in this fcn, so there's really no need for this class to _store_ a copy.

The GM is only needed to create a vector layout for the given grid, so that tracers can be allocated monolithically as a "vector" quantity.

I'll add a comment near the fcn declaration.

bartgol:
@jeff-cohere , thanks for the comments. I tried to improve documentation in the spots you mentioned. Let me know if it's clear or if it needs more rewording.

jeff-cohere (approved):
Thanks, Luca! These changes look great.

jeff-cohere (approved):

PR #875, opened 2021-02-10
Progress on CIME/SCREAM integration

### LINKED ISSUE #815: [AD] implement mechanism for passing run options to SCREAMv1

jgfouca:
Change list
1) Add PeterC's recommended SCREAM settings to scream_input.yaml
2) Convert theta.nl settings to a HOMME section of scream_input.yaml
3) Remove theta.nl from repo, it will now be generated by scream's buildnml
4) Remove missing include from scream_cxx_f90_interface.cpp
5) Add system for incorporating CIME XML settings into scream_input.yaml
6) Add capability for regenerating yams files while also keeping whitespace and comments in the regenerated file

Fixes #815 

To sum up, the strategy will be to encapsulate all scream input configuration in the scream/data/scream_input.yaml file, including Homme-related configuration items. This file will get copied to the case's RUNDIR by our buildnml script IF it doesn't exist there already. If the user wants to make temporary changes to this file, they can do it to $RUNDIR/data/scream_input.yaml. Our users should never need to touch $RUNDIR/data/theta.nl; it will be regenerated every time buildnml is run. If they want to tweak Homme settings, they should do it in the new HOMME section of the scream_input.yaml file. This approach will allow us to leverage existing nml infrastructure for homme without having to add application code for yaml. Also, I don't think there's any need for a YAML analog to user_nl_eam; the user can just make whatever changes they want in the $RUNDIR/data/scream_input.yaml.


jgfouca:
Tested both V0 and standalone. V1 runs further that it has before but crashes with:

```
  what():  /home/jgfouca/scream/components/scream/src/share/field/field_repository.hpp:253: FAIL:
map!=m_fields.end()
Error! get_field: surface_water_evaporation_flux not found.
```

Talking with @bartgol suggests this won't be an easy problem to solve, so I'm stopping here for now. We will need to discuss how to proceed after this PR.

jgfouca:
I should also mentioned that I confirmed by-hand that the theta.nl generated by buildnml is an exact match (modulo minor formatting changes) for the old scream/data/theta.nl that was in our repo.

bartgol (approved):

components/scream/data/scream_input.yaml
```diff
@@ -18,4 +18,118 @@ Atmosphere Driver:
...
+    rrtmgp_coefficients_file_sw : atm/cam/rad/rrtmgp_coefficients_sw_20181204.nc
+    solar_data_file             : atm/cam/solar/spectral_irradiance_Lean_1976-2007_ave_c20160517.nc
+    micro_p3_lookup_dir         : atm/cam/physprops
+
+    solar_data_ymd : 20000101
+    solar_data_forcing : constant
+
+    micro_p3_tableversion : 4
+
+  Output Files:
```
AaronDonahue, line 91: Maybe we should default to "None" as an option and if the output manager detects "none" then it skips everything else.  At the moment, it will likely throw an error because there isn't any viable information listed in the "Output Files" category.

components/scream/data/scream_input.yaml
```diff
@@ -18,4 +18,118 @@ Atmosphere Driver:
...
+      Ckh_s_def: 0.1 #using "moderate-stable"
+      Ckm_s_def: 0.1 #vals here as used by
+      Ck_s_min: 0.1  #DYAMOND2 SCREAMv0 run
+      #there will definitely eventually be more
+
+    P3:
+      mu_r_constant: 1.0
+      p3_qc_accret_expon: 1.15
+      p3_qc_autocon_expon: 2.47
+      #there will definitely eventually be more
```
AaronDonahue, line 58: We should create a git-issue after this PR goes in to have the respective teams audit the list of tuning parameters and add more if needed.

components/scream/data/scream_input.yaml
```diff
@@ -18,4 +18,118 @@ Atmosphere Driver:
...
       Dynamics Namelist File Name: ./data/theta.nl
+
+SCREAM:
+
+  #=====================
+  # I'm confused how we should set this. You shouldn't be able
+  # to change this once a run starts, but you should be able to choose it
+  # when you start a new run... What if your initial condition is for year 2000
+  # but you're doing a control run and want it to start at 01-01-0001?
+  Start Date: 01-01-2000
```
AaronDonahue, line 29: Can this be gathered from `case_run.xml` and thus replaced by the `./create_case` call?

AaronDonahue (approved):
It looks good to me, thank you @jgfouca for putting this together.  I had a few comments/questions but they won't effect this PR going in.  More of, "while we are on the subject" type of stuff.

components/scream/data/scream_input.yaml
jgfouca: @AaronDonahue , I'm not a YAML expert, but I think this will work fine. When a client asks for Output Files, they will get an empty map/dictionary.

jgfouca: Whoops, I just checked and the python version actually defaults to None: `('Output Files', None)`, so the behavior already matches what you want.

components/scream/data/scream_input.yaml
jgfouca: @AaronDonahue , if start date comes from the CIME XML env, I'm wondering if we even need this in the YAML? I guess maybe for non-CIME (aka standalone) runs? Normal CIME cases use this setting:

```
<entry id="RUN_STARTDATE" value="0001-01-01">
```

jgfouca: Ok, I think I know a good solution. Commit coming.

jgfouca:
@AaronDonahue , @bartgol , I made a significant push to this PR after your reviews. There's now a system to incorporate CIME XML settings into scream_input.yaml.

components/scream/cime_config/buildnml
```diff
@@ -49,9 +77,16 @@ def buildnml(case, caseroot, compname):
...
-    src    = os.path.join(case.get_value("SRCROOT"), "components/scream/data")
+    rundir   = case.get_value("RUNDIR")
+    target   = os.path.join(rundir, "data")
+    src      = os.path.join(case.get_value("SRCROOT"), "components/scream/data")
+    yaml_tgt = os.path.join(target, "scream_input.yaml")
+
+    # Specify yaml input changes that are driven by case env
+    # maps key-tuple to case env name to lookup
+    yaml_changes = {
+        ("SCREAM", "Start Date"): "RUN_STARTDATE",
```
bartgol, line 88: I guess we should later add info about the grid resolution as well. Later.

components/scream/data/scream_input.yaml
bartgol: Should we remove the comment, since you found a way to do it?

bartgol (approved):
Looks good.

jgfouca:
@AaronDonahue , @bartgol , sorry for spamming you. A did another (last?) significant push. Preserving whitespace and comments required a pretty big kludge.

components/scream/cime_config/buildnml
```diff
@@ -26,6 +26,50 @@ import yaml
 
 logger = logging.getLogger(__name__)
 
+###############################################################################
+def load_comments(filepath):
```
bartgol, line 30: Can you add a comment to these two functions explaining what they achieve and how they can be useful? For the sake of who will come next to maintain them. I'm having a hard time parsing them.

components/scream/data/scream_input.yaml
```diff
@@ -18,4 +18,113 @@ Atmosphere Driver:
     Reference Grid: Physics GLL
     Dynamics Driven:
       Dynamics Namelist File Name: ./data/theta.nl
+
+SCREAM:
+
+  Start Date: 01-01-2000 # For CIME cases, do NOT set this! It will be set by CIME's RUN_STARTDATE
+
+  # Model Timesteps
+  whole_model_dt: 3600.0 #aka phys/dyn coupling dt
```
AaronDonahue, line 27: I appreciate this comment.  It was super confusing to me what timesteps meant way back when I started running E3SM.

components/scream/data/scream_input.yaml
AaronDonahue: Okay great, then I will need to put in a hot fix to the output manager once this PR goes in.  I'll make a git-issue.  It shouldn't impact any of the unit tests since the ones that use I/O set Output Files.

AaronDonahue: We should adjust the comment that follows to something like.
"This should be a bracketed list of all the output control yaml files"
On the same note, I think we will eventually have a default yaml output control file for CIME runs that we will need to copy over to the run directory when the case is built.  We may need to think about this in a little more depth, but that conversation can wait until we have a CIME run working.

AaronDonahue (approved):

jgfouca:
OK, I think I have addressed all concerns.

PR #881, opened 2021-02-12
Adds a coupled dyamond2 compset

singhbalwinder:
A compset `A_SCREAM-HR-DYAMOND2-2010` is added which uses `2010_EAM%SCREAM-HR-DYAMOND2_ELM45%SPBC` compsets for atmosphere and land along with the A_WCYCL's compsets for MPAS ocean and sea-ice.


singhbalwinder:
I still need to make some changes so that the compset automatically picks up the right coupler files and sets the right coupling frequency. A test run on Cori is still running.

singhbalwinder:
The run blew up with the following error:
```
----------------------------------------------------------------------
Beginning MPAS-seaice Error Log File for task       0 of   12000
    Opened at 2021/02/12 12:11:56
----------------------------------------------------------------------

CRITICAL ERROR: rescale_mesh: inconsistent radii: sphere_radius: 1.00000000000000, earthRadius: 6371229.00000000
Logging complete.  Closing file at 2021/02/12 12:11:56
```
Stack:
```
 831: forrtl: error (76): Abort trap signal
  831: Image              PC                Routine            Line        Source
  831: e3sm.exe           0000000004286144  Unknown               Unknown  Unknown
  831: e3sm.exe           0000000003F960B0  Unknown               Unknown  Unknown
  831: e3sm.exe           00000000043E9A10  Unknown               Unknown  Unknown
  831: e3sm.exe           000000000483DEE1  Unknown               Unknown  Unknown
  831: e3sm.exe           00000000040CABD2  Unknown               Unknown  Unknown
  831: e3sm.exe           000000000409866B  Unknown               Unknown  Unknown
  831: e3sm.exe           000000000409F415  Unknown               Unknown  Unknown
  831: e3sm.exe           00000000020D37E2  mpas_log_mp_mpas_         828  mpas_log.f90
  831: e3sm.exe           0000000001D48FC1  seaice_mesh_mp_re         296  mpas_seaice_mesh.f90
  831: e3sm.exe           0000000001D4A8E7  seaice_mesh_mp_se          73  mpas_seaice_mesh.f90
  831: e3sm.exe           0000000001D225F1  seaice_initialize         112  mpas_seaice_initialize.f90
  831: e3sm.exe           00000000021968B7  seaice_core_mp_se         254  mpas_seaice_core.f90
  831: e3sm.exe           0000000001C16B47  ice_comp_mct_mp_i         568  ice_comp_mct.f90
  831: e3sm.exe           00000000004284A8  component_mod_mp_         257  component_mod.F90
  831: e3sm.exe           0000000000415825  cime_comp_mod_mp_        1439  cime_comp_mod.F90
  831: e3sm.exe           0000000000424F7B  MAIN__                    122  cime_driver.F90

```
I might be setting something incorrectly. I am not sure how to proceed but I will see if I can get some help from @jonbob .

jonbob:
@singhbalwinder - that's usually an indication that the compset and grid are inconsistent with the whole %SPUNUP and _ICG issue. I'm in the process of getting rid of that whole approach, but in the meantime let me help you get this going. What is the compset longname? And the mesh? If the compset has %SPUNUP for ocn and ice, then the grid needs to be the _ICG version. And if it doesn't have the %SPUNUP, then it can't use the _ICG mesh. I know that's a bit of a mess -- and why we're getting rid of it

singhbalwinder:
You are right. The compset long name has SPUNUP in it:
```
2010_EAM%SCREAM-HR-DYAMOND2_ELM45%SPBC_MPASSI%SPUNUP_MPASO%SPUNUP_MOSART_SGLC_SWAV`)
```
I used the above following WCYCL CMIP6 compset. 

I am using the following grid:
`ne1024pg2_r0125_oRRS18to6v3`

Should I just remove SPUNUP and try again?

jonbob:
Yes, try it that way or use the ne1024pg2_r0125_oRRS18to6v3_ICG grid, if it exists

singhbalwinder:
Thanks! I will give it a try.

cime_config/allactive/config_compsets.xml
```diff
@@ -353,6 +353,14 @@
   <lname>SSP585_EAM%CMIP6_ELM%CNPECACNTBC_MPASSI%SPUNUP_MPASO%SPUNUP_MOSART_SGLC_SWAV_BGC%BDRD</lname>
 </compset>
 
+<!-- SCREAM DYAMOND compsets -->
+<compset>
+  <alias>A_SCREAM-HR-DYAMOND2-2010</alias>
```
PeterCaldwell, line 358: This is a larger discussion point rather than a complaint about this compset: it is a bit oxymoronic to include 2010 *and* DYAMOND2 in the compset name since DYAMOND2 implies 2020, not 2010. Can we drop the 2010 from the end of this name? Note that our F compsets also have 2010 and DYAMOND2 in their name. @brhillman and @crterai ?

PeterCaldwell: @singhbalwinder - I think you can drop the "-2010" from the end of this compset name. Could you make that change and confirm it still runs? If so, I will integrate it today. It would be nice to have an initial condition/grid for this case, but I think that will take much more work (including get exec blessing to bug Luke and Jon) so should be a new PR.

brhillman: Certainly the 2010 can go away in the alias. And, really, we should probably take it out of the lname too, maybe replace "2010" with "2020" or "DYAMOND2". I think previously, we had the 2000, and later the 2010 in there because we were setting some parameters for the compset based on those things, like the use case files. Since we set specific use cases for the DYAMOND compset, I'm not sure this is needed. The other two places I can think of that would use this would be setting the run startdate, and CAM_CONFIG_OPTS. Our setting of CAM_CONFIG_OPTS for scream does not look for 2010 though, so that's not a problem, and RUN_STARTDATE is set by looking for the string `_EAM%SCREAM-[H,L]R-DYAMOND[1,2]`, so I think we are good there too. TL;DR, I think we can lose the 2010 altogether.

PeterCaldwell (approved):
Thanks Balwinder!!!

Issue #575, opened 2020-09-15
Add temporary buffers support for atmosphere processes


bartgol:
The AD won't likely use the WorkspaceManager from ekat. It will likely have something simple, like the FunctorsBuffersManager in Hommexx (which simply allocates a 1d buffer that accommodates all needs). Each atm process should then use the memory pointer provided by the AD. That's when the WSM from ekat will likely be used, i.e., created inside each atm processes, with memory provided from the AD.

Action items:
- add interfaces to atm process for getting the required memory size, and for setting a (device) memory ptr.
- make the AD query atm procs for needed temp memory, allocate memory, and finally set memory inside the atm procs.

tcclevenger:
@bartgol Is the plan to no longer use WSM inside the main functions (e.g. shoc_main()), but instead create and pass through a new structure which manages AD memory (the `FunctorsBuffersManager`-like structure)? Or is it just to use a`FunctorsBuffersManager`-like structure inside the AD for the non-FM variables passed to the main function, with shoc impl being completely agnostic to its existence? 

I think this task will be related to some temp variable work I need to do in shoc (making tracer transpose a workspace variable), so I may take this issue.

bartgol:
I don't really care how we achieve the result, but the logic should be roughly this: first the AD queries all atm procs for their temporary needs. Then, the AD allocates a single buffer that accommodates all needs, and pass the buffer back to all atm procs.

The second step could be achieved by having the AD store a single WSM (instantiated to accommodate the largest need among all atm procs), and provide it to all atm procs. However, different Atm Procs might use different Team policies, or even need temps on different grids. Ideally, e.g., I'd like Homme's FunctorsBufferManager to use the scratch mem provided by the AD.

We can discuss more via slack, and report here our conclusions, if you want, but my guess is that the AD should gather raw-memory needs from all atm procs, and provide back a raw-memory buffer. Each atm proc is then free to use this in whatever way they want. If we go this route, we need to provide a ctor of WSM that also takes in the raw mem buffer (without allocating it). It would be "risky", since WSM would have no way to check that the provided mem buffer is long enough. But it would be marked as a power-user ctor, with all the documentation needed to be aware of risks/assumptions.

PR #882, opened 2021-02-12
Refactor SHOC-AD interface

tcclevenger:
Refactors the shoc-ad interface to call c++ functions. The test `shoc_stand_alone` matches `shoc_run_and_cmp` closely, and I've tested that with the exact same inputs, the output variables are the same. 

There is currently no post-shoc_main computations done yet. The `shoc_preamble` is based on pre-shoc_main computations in `shoc_intr.F90`. 

Note on tracers: in `shoc_intr.F90`, the variables in `state%q` are used both as a monolith (stored into `edsclr_in` and passed to `shoc_main()` to be advected, and 3 individual tracers (`tke/shoc_ql/shoc_qv`) are used as individual inputs to `shoc_main()`. It appears these 3 are also in the monolithic array as well, so their values are update in 2 different ways. Here, I require `tke/shoc_ql/shoc_qv` as individual required/computed field, and when they are passed I tag them as "TRACERS" to be included in the Q group for advection. For the stand alone test, I store no additional tracers, but more can be added.

Other smaller changes/notes:
1. Adding/removing `const` keyword where possible/necessary.
2. Q is expected to be layout `(shcol,num_tracer,nlev)`, but shoc expects tracers in layout `(shcol,nlev,num_tracer)`. Here I make a copy of the tracer array and do the transpose in the `ShocPreamble`. In a separate optimization PR I will change the shoc impl to take the `(shcol,num_tracer,nlev)` layout and perform local transpose before solving.
3. Temporarily disabling p3_shoc_coupled test so that autotester can run


components/scream/src/physics/shoc/shoc_functions_f90.cpp
```diff
@@ -2247,7 +2247,9 @@ void update_prognostics_implicit_f(Int shcol, Int nlev, Int nlevi, Int num_trace
 
   view_3d
     tracer_d(temp_3d_d[0]);
-  Kokkos::resize(Kokkos::WithoutInitializing, tracer_d,
+
+  // update_prognostics_implicit() expects 3 extra tracer slots
+  Kokkos::resize(tracer_d,
                  shcol,nlev,ekat::npack<Spack>(num_tracer+3));
```
tcclevenger, line 2253: The changes in this file do not affect computations. `Kokkos::WithoutInitializing` is unnecessary.

components/scream/src/physics/shoc/atmosphere_macrophysics.hpp
```diff
@@ -57,6 +77,194 @@ class SHOCMacrophysics : public scream::AtmosphereProcess
   const std::set<FieldIdentifier>& get_computed_fields () const { return m_computed_fields; }
   std::set<GroupRequest> get_updated_groups () const { return m_inout_groups_req; }
 
+  /*--------------------------------------------------------------------------------------------*/
+  // Most individual processes have a pre-processing step that constructs needed variables from
+  // the set of fields stored in the field manager.  A structure like this defines those operations,
+  // which can then be called during run_imple in the main .cpp code.
```
jeff-cohere, line 83: Should run_imple be run_impl?

components/scream/src/physics/shoc/shoc_inputs_initializer.hpp
```diff
@@ -0,0 +1,112 @@
...
+
+public:
+
+  virtual ~SHOCInputsInitializer () = default;
+
+  // The name of the initializer
+  std::string name () const { return "SHOCInputsInitializer"; }
+
+  // Initialize fields
+  void initialize_fields ();
```
jeff-cohere, line 26: You can add the `override` keyword here to make clear this is an overridden method.

components/scream/src/physics/shoc/shoc_inputs_initializer.hpp
```diff
@@ -0,0 +1,112 @@
...
+
+  const std::set<FieldIdentifier>& get_inited_fields () const {
```
jeff-cohere, line 28: Also overridden.

components/scream/src/physics/shoc/shoc_inputs_initializer.hpp
```diff
@@ -0,0 +1,112 @@
...
+  using SHF            = shoc::Functions<Real, DefaultDevice>;
+  using Spack          = typename SHF::Spack;
+  using view_1d_scalar = typename SHF::view_1d<Real>;
+  using view_2d        = typename SHF::view_2d<Spack>;
+
+public:
+
+  virtual ~SHOCInputsInitializer () = default;
+
+  // The name of the initializer
```
jeff-cohere, line 22: Can be annotated with `override`.

components/scream/src/physics/shoc/shoc_inputs_initializer.hpp
```diff
@@ -0,0 +1,112 @@
...
+      else {
+        Real ra = (z1 - z0)/(th1 - th0);
+        pres(i, view_k)[pack_k] = pow(pow(p0, c0) + c0*c1*ra*log(th1/th0), 1.0/c0);
+      }
+    }
+  }
+
+protected:
+
+  void add_field (const field_type& f);
```
jeff-cohere, line 98: Annotate these methods with `override` if you like.

components/scream/src/physics/shoc/shoc_inputs_initializer.cpp
```diff
@@ -0,0 +1,354 @@
...
+  const auto& id = f.get_header().get_identifier();
+  const auto& id_ref = f_ref.get_header().get_identifier();
+
+  // To the AD, we only expose the fact that we init f_ref...
+  m_fields_id.insert(id_ref);
+
+  // ...but SHOC only knows how to init f...
+  m_fields.emplace(id.name(),f);
+
+  // ...hence, we remap to f_ref.
```
jeff-cohere, line 43: These are lovely comments. Thanks!

tcclevenger: I think we must credit @AaronDonahue for using it in p3_inputs_initializer...

jeff-cohere (approved):
Looks nice! I made some minor cosmetic suggestions.

tcclevenger:
> Looks nice! I made some minor cosmetic suggestions.

Thanks @jeff-cohere! I'll make the changes.

components/scream/src/physics/shoc/atmosphere_macrophysics.cpp
```diff
@@ -55,55 +115,269 @@ set_updated_group (const FieldGroup<Real>& group)
...
+          zt_grid("zt_grid",m_num_cols,nlev_packs),
+          zi_grid("zi_grid",m_num_cols,nlevi_packs),
+          wtracer_sfc("wtracer_sfc",m_num_cols,num_tracer_packs),
+          wm_zt("wm_zt",m_num_cols,nlev_packs),
+          exner("exner",m_num_cols,nlev_packs),
+          thlm("thlm",m_num_cols,nlev_packs),
+          qw("qw",m_num_cols,nlev_packs),
+          cloud_frac("cloud_frac",m_num_cols,nlev_packs);
+
+  view_3d tracers("tracers",m_num_cols,m_num_levs,num_tracer_packs);
```
ambrad, line 180: So, tbc, in the optimization follow-up PR, you won't allocate a 3D tracer array here?

components/scream/src/physics/shoc/atmosphere_macrophysics.cpp
```diff
@@ -55,55 +115,269 @@ set_updated_group (const FieldGroup<Real>& group)
...
 
-  Q->sync_to_dev();
+  // Output Variables
+  output.pblh     = m_shoc_fields_out["pblh"].get_reshaped_view<Pack1d*>();
+
+  view_2d shoc_ql2("shoc_ql2",m_num_cols,nlev_packs);
+  output.shoc_ql2 = shoc_ql2;
+
+  // Ouput (diagnostic)
+  view_2d shoc_mix("shoc_mix",m_num_cols,nlev_packs);
```
ambrad, line 227: Will all of these views ultimately disappear? This is time-stepping code, right, not one-time init? If so, we want to use the workspace manager for views like this so we don't bog down in view allocations and deallocations.

ambrad (approved):
I've left two comments regarding allocation of Views. We don't want to allocate Views during time stepping if we can avoid it; the workspace manager should provide these. Ultimately, the AD should provide the workspace manager to params for their tmp needs; this complements the permanent fields the AD manages.

I'm approving because I'm OK with incremental PRs rather than getting everything perfect in one go. Nice work!

bartgol:
I agree with @ambrad . Memory should be allocated only once during initialization. Unfortunately, the "global" atm-wide workspace/scratch-mem manager is not yet in the AD (added to the todo list long ago: #575). Since "getting scream to work" is a priority over "getting efficient memory handling", it is not necessary to do it in this PR. But I would add a "TODO" comment, so that we don't forget about these views.

components/scream/src/physics/shoc/atmosphere_macrophysics.cpp
```diff
@@ -22,30 +27,85 @@ void SHOCMacrophysics::set_grids(const std::shared_ptr<const GridsManager> grids
...
+  m_required_fields.emplace("u",        scalar3d_layout_mid, m/s,         grid_name);
+  m_required_fields.emplace("v",        scalar3d_layout_mid, m/s,         grid_name);
+  m_required_fields.emplace("wthv_sec", scalar3d_layout_mid, K*(m/s),     grid_name);
+  m_required_fields.emplace("tkh",      scalar3d_layout_mid, (m*m)/s,     grid_name);
+  m_required_fields.emplace("tk",       scalar3d_layout_mid, (m*m)/s,     grid_name);
+  m_required_fields.emplace("shoc_ql",  scalar3d_layout_mid, Qunit,           grid_name);
+
+  m_computed_fields.emplace("s",        scalar3d_layout_mid, J/kg,        grid_name);
+  m_computed_fields.emplace("tke",      scalar3d_layout_mid, (m*m)/(s*s), grid_name);
+  m_computed_fields.emplace("u",        scalar3d_layout_mid, m/s,         grid_name);
```
bartgol, line 90: Are we planning on having two separate fields for the two components of the horiz velocity? I am guessing most uses of horiz velocity will need both components, so a vector velocity might be better (localized mem access).

components/scream/src/physics/shoc/atmosphere_macrophysics.cpp
```diff
@@ -112,21 +386,27 @@ void SHOCMacrophysics::set_required_field_impl (const Field<const Real>& f) {
   // at the beginning of the run call. Other than that, there would be really
   // no need to store a scream field here; we could simply set the view ptr
   // in the Homme's view, and be done with it.
-  m_shoc_fields_in.emplace(f.get_header().get_identifier().name(),f);
+  const auto& name = f.get_header().get_identifier().name();
+  m_shoc_fields_in.emplace(name,f);
+  m_shoc_host_views_in[name] = Kokkos::create_mirror_view(f.get_view());
+  m_raw_ptrs_in[name] = m_shoc_host_views_in[name].data();
```
bartgol, line 392: Does shoc still need raw ptrs? Besides not being necessary, they are getting more dangerous to use now. Tracers can be subviews of a big Q array, but the subview is not contiguous in memory (the layout is (icol,iq,ilev)), so the ptr alone is not enough to access the array at the right location.

I'm assuming ptrs were used for calling f90. If that's still necessary, we need to allow tracers to be allocated not contiguously, OR have shoc copy them manually in/out to/from a contiguous array before/after calling f90.

components/scream/src/physics/shoc/atmosphere_macrophysics.cpp
```diff
@@ -112,21 +386,27 @@ void SHOCMacrophysics::set_required_field_impl (const Field<const Real>& f) {
   // at the beginning of the run call. Other than that, there would be really
   // no need to store a scream field here; we could simply set the view ptr
   // in the Homme's view, and be done with it.
-  m_shoc_fields_in.emplace(f.get_header().get_identifier().name(),f);
+  const auto& name = f.get_header().get_identifier().name();
+  m_shoc_fields_in.emplace(name,f);
+  m_shoc_host_views_in[name] = Kokkos::create_mirror_view(f.get_view());
```
bartgol, line 391: If you really need host views, do not create the mirror manually. The stored view might not be what you think for subviewed fields (like tracers). Instead, do (for a scalar 3d field)

`auto view_h = my_field.get_reshaped_view<Real**, Host>();`

The method `get_view` is too risky now (I can give you more details offline if you want), and I will likely remove it (or make it private, accessible only to some friends).

components/scream/src/physics/shoc/atmosphere_macrophysics.cpp
```diff
@@ -112,21 +386,27 @@ void SHOCMacrophysics::set_required_field_impl (const Field<const Real>& f) {
...
 void SHOCMacrophysics::set_computed_field_impl (const Field<Real>& f) {
   // Store a copy of the field. We need this in order to do some tracking updates
   // at the end of the run call. Other than that, there would be really
   // no need to store a scream field here; we could simply set the view ptr
   // in the Homme's view, and be done with it.
-  m_shoc_fields_out.emplace(f.get_header().get_identifier().name(),f);
+  const auto& name = f.get_header().get_identifier().name();
+  m_shoc_fields_out.emplace(name,f);
+  m_shoc_host_views_out[name] = Kokkos::create_mirror_view(f.get_view());
+  m_raw_ptrs_out[name] = m_shoc_host_views_out[name].data();
```
bartgol, line 406: Same comment about host views and host ptrs.

bartgol (changes requested):
Great work. I have some comments regarding host mirrors and host pointers.

components/scream/src/physics/shoc/atmosphere_macrophysics.cpp
tcclevenger: With these I'm not sure since they are output variables of shoc_main. It would depend if they are needed for any post-shoc_main computation below in `run_impl()`.

components/scream/src/physics/shoc/atmosphere_macrophysics.cpp
tcclevenger: Yes, I'll basically just pass the Q array to shoc_main and the transpose will be handled internally.

components/scream/src/physics/shoc/atmosphere_macrophysics.cpp
tcclevenger: Edit: I see Luca's comment below. Yes, they will. I will add a TODO.

components/scream/src/physics/shoc/atmosphere_macrophysics.cpp
tcclevenger: Yes, this is something I plan to do for the (ever growing) shoc optimization PR(s). It will also eliminate the need for a temporary variable in the tridiag solver call..

components/scream/src/physics/shoc/atmosphere_macrophysics.cpp
tcclevenger: You are correct, they are unnecessary. I'll remove.

components/scream/src/physics/shoc/atmosphere_macrophysics.cpp
tcclevenger: I believe these are also unnecessary, since I never access host_views anywhere.

tcclevenger:
Thanks @ambrad @bartgol! I've added a few TODO comments and remove some unnecessary structures.

bogensch (approved):

bartgol (approved):

Issue #880, opened 2021-02-12
[AD] Remaining scorpio tasks


AaronDonahue:
This issue makes it easier to understand what the remaining tasks for the scorpio interface are by combining them all together:

**High Priority**
- [x] Assure that I/O can handle variables that are packed - see #787 

**Medium Priority**
- [x] Get restarts working as expected - see #798 
- [x] Investigate the output file naming convention - see #789 
- [ ] YAML control of netCDF header - see #788 
- [ ] Automatic registration of dimensions as variables - see #786 
- [ ] Enable PIO_STRIDE - see #785 
- [x] Allow compile time adjustment of individual output control files during a CIME build - see #879
- [x] Add treatment for DIM field tag in I/O - see #930 
- [x] Switch input class to using grid rather than grid manager at construction - see #1109 

**Low Priority**
- [x] Make input able to handle files with more than one timesnap - see #799 
- [x] Make restart a separate sub-class of output class - see #797 
- [x] Better handling of output frequency - see #790 

**Wishlist**
- [x] Get rid of remaining F90 code in SCREAMv1 I/O interface - see #823 and #792 
- [x] Make an "averaging type" class for future flexibility - see #795 
- [ ] Create an I/O class that has output and input as sub-classes - see #794 
- [ ] Add option to write a restart at finalization regardless of restart frequency - see #793 
- [x] Allow for output on dynamics grid - see #791 
- [ ] Make it possible for SCREAM to output initial condition files.
- [ ] Allow SCREAM to load true non-hydrostatic IC files **and** hydrostatic IC files.

Issue #787, opened 2020-12-08
[AD] SCORPIO-Interface: Assure that I/O can handle variables that are packed


AaronDonahue:
In order to be compatible with the AD and full scale runs it is imperative that I/O properly handle packed arrays.

This task is to create a test that uses I/O with packed arrays and verifies that the input and output is correct.

This will most likely require some code refactoring to take advantage of EKAT routines for masked arrays.

AaronDonahue:
Addressed by #885

PR #885, opened 2021-02-16
Aarondonahue/io packed arrays

AaronDonahue:
Introduces infrastructure to support packed arrays for reading and writing from the field manager.

This commit expands the io_test to include a packed array and makes sure that output written and input read doesn't accidentally capture the padded values in the array.

This commit also cleans up a bunch of old unused I/O routines.

Addresses Issue #787 , which is also listed in #880 

[BFB]


components/scream/src/share/io/scream_scorpio_interface.cpp
```diff
@@ -150,58 +176,39 @@ void grid_read_data_array(const std::string &filename, const std::string &varnam
...
 };
 /* ----------------------------------------------------------------- */
+// Handling the writing of output for packed arrays
+void grid_write_data_array(const std::string &filename, const std::string &varname, const std::vector<int>& dims, const Int& dim_length, const Int& pack_extra, const Real* hbuf)
+{
+  if (pack_extra == 0) // then no xtra data, is contiguous
+  {
+    grid_write_data_array_c2f_real_1d(filename.c_str(),varname.c_str(),dim_length,hbuf);
+  } else {
+    std::vector<Real> hbuf_new;
```
bartgol, line 186: Since this chunk of code is the same as in the `read` routine, I think it's best to factor it out into an independent function. Something like
```
void remove_padding (const Real* src, Real* dst, const std::vector<int>& phys_dims, const int padding);
```
and then call
```
int dim_length = std::accumulate(dims.begin(),dims.end(),1,std::multiplies<int>());
std::vector<Real> hbuf_new(dim_length);
remove_padding(hbuf, hbuf_new.data(),dims,pack_extra);
```

components/scream/src/share/io/scream_scorpio_interface.cpp
```diff
@@ -138,9 +138,35 @@ void eam_pio_enddef(const std::string &filename) {
...
+    {
+      for (int jj=0;jj<dims.back();jj++)
+      {
+        hbuf[loc] = (hbuf_new[dloc]);
+        loc += 1;
+        dloc += 1;
+      }
+      loc += pack_extra;
+    }
+    EKAT_REQUIRE_MSG((dim_length == dloc),"Error read input for field " + varname + ", packed and unpacked data variables don't match in size.");
```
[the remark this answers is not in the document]
bartgol: Btw, since `dim_length=dims[0]*...*dims[n]`, you could do
```
auto slow_dim_len = dim_length / dims.back();
```

components/scream/src/share/io/tests/io.cpp
```diff
@@ -76,6 +76,20 @@ TEST_CASE("input_output_basic","io")
     m_output_manager.run(time);
   }
   m_output_manager.finalize();
+  // Because the next test for input in this sequence relies on the same fields.  We set all field values
+  // to nan to ensure no false-positive tests if a field is simply not read in as input.
+  for (auto it : out_fields)
+  {
+    auto f_dev  = field_repo->get_field(it,"Physics").get_view();
+    auto f_host = Kokkos::create_mirror_view( f_dev );
+    Kokkos::deep_copy(f_host,f_dev);
```
bartgol, line 85: You could simply do
```
Kokkos::deep_copy(f_dev,std::nan(""));
```
The `deep_copy` function is overloaded, and can either copy a src view into a dst view or a single value into a dst view.

components/scream/src/share/io/tests/io.cpp
```diff
@@ -195,36 +213,56 @@ std::shared_ptr<FieldRepository<Real>> get_test_repo(const Int num_lcols, const
...
   repo->register_field(fid2,{"output","restart"});
   repo->register_field(fid3,{"output","restart"});
+  repo->register_field<Pack>(fid4,{"output","restart"}); // Register field as packed
   repo->registration_ends();
 
+  // Make sure that field 4 is in fact a packed field
+  auto field4 = repo->get_field(fid4);
+  auto fid4_extent = field4.get_header().get_alloc_properties().get_last_extent();
+  auto fid4_final_dim = field4.get_header().get_identifier().get_layout().dims().back();
+  REQUIRE(fid4_extent-fid4_final_dim > 0);
```
bartgol, line 234: I think you need `>=0`. If pack sizes divides the field last dim, then `fid4_extent` and `fid4_final_dim` are the same.  In your case, you have nlevs=3 and a pack size 2, so it's fine. But if someone edits the test, it might be hard to catch what's going on.

You could make the test robust to edits, like this:
```
// Make sure that if ppl alter the test inputs, they see an error
EKAT_REQUIRE_MSG((last_dim % packsize)==0,
    "Error! Please, don't alter "nlevs" or "packsize" for this test.\n);
REQUIRE (extent-last_dim>0);
```

components/scream/src/share/io/tests/io.cpp
```diff
@@ -195,36 +213,56 @@ std::shared_ptr<FieldRepository<Real>> get_test_repo(const Int num_lcols, const
...
   Kokkos::deep_copy( f3_hst, f3_dev );
+  Kokkos::deep_copy( f4_hst, f4_dev );
   for (int ii=0;ii<num_lcols;++ii)
   {
     f1_hst(ii) = ii;
     for (int jj=0;jj<num_levs;++jj)
     {
       f2_hst(jj) = (jj+1)/10.0;
       f3_hst(ii,jj) = (ii) + (jj+1)/10.0;
+//      f4_hst(ii,jj) = (ii) + (jj+1)/10.0;
```
bartgol, line 256: Remove commented line.

bartgol (changes requested):

components/scream/src/share/io/tests/io.cpp
AaronDonahue: you bring up a good point.  What happens when packsize = 1?  Shouldn't extent = last_dim?

AaronDonahue: if so, then I think the REQUIRE shouldn't be there.  I'm essentially trying to avoid someone nullifying the packed test by accidentally making everything fit perfectly with nlev and packsize.

AaronDonahue:
@bartgol , I think I addressed all of your comments in this latest commit.  Except one about checking that the packed field in io.cpp is actually a padded field.  I responded to your comment with a question, hopefully we can figure out the right away to put in that check.

I also took a stab at adding the `get_padding` function to field allocation properties.  Can you take a look at that?

bartgol (approved):

PR #886, opened 2021-02-17
dump-scream-env: New tool for dumping SCREAM-approved env for a machine

jgfouca:
Designed for integration with bash functions for easy env loading.

A couple other minor pylint and whitespace fixes.


bartgol (commented):
I'm not sure why having it as an option for test-all-scream wasn't ok, but I don't have a strong feeling.

One thing I would change though is the script name: the script does not dump the env, but rather then command used to setup the env. So maybe `dump-scream-env-setup-cmd` or `get-env-setup-cmd'? Or something along that line?

Not a strong feeling though, so I'm not marking it as a request.

jgfouca:
@bartgol , the reason for removing it from test-all-scream was this:

```
% ./test-all-scream --print-mach-setup -m mappy
Last commit on ref 'master': 17d7dc6 Merge Pull Request #873 from E3SM-Project/scream/bartgol/bundled_tracers
Using commit origin/master to generate baselines
Note: no value passed for --ctest-parallel-level. Using the default for this machine: 48
Note: no value passed for --make-parallel-level. Using the default for this machine: 48
Warning: repo is not clean: M cime
 M components/elm/src/external_models/sbetr
 M components/mpas-source
 M externals/YAKL
 M externals/ekat
 M externals/scorpio
ERROR: Repo must be clean before running. If testing against HEAD or pre-built baselines, you can pass `--keep-tree` to allow non-clean repo.

* Wondering why I need my repo clean just to dump the env*

% ./test-all-scream --print-mach-setup -m mappy -k
Last commit on ref 'master': 17d7dc6 Merge Pull Request #873 from E3SM-Project/scream/bartgol/bundled_tracers
Using commit HEAD to generate baselines
Note: no value passed for --ctest-parallel-level. Using the default for this machine: 48
Note: no value passed for --make-parallel-level. Using the default for this machine: 48
WARNING! You have uncommitted changes in your repo.
         The PASS/FAIL status may depend on these changes
         so if you want to keep them, don't forget to create a commit.

Bash commands for machine 'mappy' setup:

  module purge && module load sems-env sems-python/3.5.2 sems-gcc/9.2.0 sems-cmake/3.12.2 sems-git/2.10.1 sems-openmpi/4.0.2 sems-netcdf

OVERALL STATUS: PASS
```

As you can see, the first run failed when it probably should not have. The second run worked but produced lots of extra output that would thwart any bash `eval` of the output like I'm doing with the `load_scream` bash functions in the documentation for this tool. 

We could go through the tedium of tweaking test-all-scream to behave differently when the user asks for print-mach-setup, but to me it seemed easier to just make a separate tool.

jgfouca:
@bartgol , I think you were right, the original name of the tool was not great. I changed it to `scream-env-cmd`. I think this better-describes what the tool does and is also more similar to `modulecmd` which is the name of the command the module env system uses for the same purpose.

jgfouca:
Marking as stale since there's no reason to re-run tests on the name change.

bartgol:
> Marking as stale since there's no reason to re-run tests on the name change.

FYI, the label `AT: STALE` is particular, in that it is actually set by the AT itself. It is set when the PR's last commit passed the tests, but a long time has passed since then (I think our config is set to 1 week). When set, the AT won't automerge the PR even if the last test was a success.

If you want to skip testing, you need to set `CI: Integrate Without Testing`.

bartgol (approved):

PR #887, opened 2021-02-18
Use workspace for local variables

tcclevenger:
In SHOC, change to use `ekat::Workspaces` instead of defining local variables for all local 2d views. This is just an memory/# of lines optimization, doesn't change any computations.

### DIFF
--- a/components/scream/src/physics/shoc/shoc_main_impl.hpp
+++ b/components/scream/src/physics/shoc/shoc_main_impl.hpp
@@ -43,9 +43,7 @@ void Functions<S,D>::shoc_main_internal(
   const uview_1d<const Spack>& exner,
   const Scalar&                phis,
   // Workspace/Local Variables
-  const Workspace&             workspace_nlev,
-  const Workspace&             workspace_nlevi,
-  const WorkspaceScalar&       workspace_nlev_scalar,
+  const Workspace&             workspace,
   const uview_2d<Spack>&       X1,
   // Input/Output Variables
   const uview_1d<Spack>&       host_dse,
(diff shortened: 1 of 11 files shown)


components/scream/src/physics/shoc/shoc_main_impl.hpp
```diff
@@ -313,9 +275,18 @@ Int Functions<S,D>::shoc_main(
...
   const auto policy = ekat::ExeSpaceUtils<ExeSpace>::get_default_team_policy(shcol, nlev_packs);
+
+  ekat::WorkspaceManager<Spack,  Device> workspace_mgr_nlev(nlev_packs, 9, policy);
+  ekat::WorkspaceManager<Spack,  Device> workspace_mgr_nlevi(nlevi_packs, 5, policy);
+  ekat::WorkspaceManager<Scalar, Device> workspace_mgr_nlev_scalar(nlev, 3,  policy);
+
   Kokkos::parallel_for(policy, KOKKOS_LAMBDA(const MemberType& team) {
     const Int i = team.league_rank();
 
+    auto workspace_nlev        = workspace_mgr_nlev.get_workspace(team);
```
ambrad, line 286: Have you spoken with Jim and Luca about having multiple workspace managers for different-sized arrays? I believe this is not the intent. The intent of the workspace manager is to have one pool of memory that can be carved up as needed; thus, the intent is to need just one workspace manager. The Workspace type itself I believe already permits taking views of different scalar types. The WorkspaceManager should be init'ed for size corresponding to nlevi, which then is sufficient to provision views of scalar and pack types for nlev and nlevi. Thus, I believe you should only need one WorkspaceManager, the workspace_mgr_nlevi one.

tcclevenger: I see where you can request views of different scalar types (`workspace.take<Spack/Scalar>()`), but I don't see where you can change the size of the view that you take . So this sometimes would result in receiving views which are 1 pack too large (for views which are supposed to be size `npack(nlev)`). Would this even be an issue though? The extra part would just hang around untouched? Or I could just consider a subview? 

@jgfouca I may be missing some workspace functionality which takes care of this..

ambrad: Right, take a subview. You might add convenience routines to Workspace to provision views smaller than the default size. But possibly large chunks of code are insensitive to 1D views that are larger than needed, so subviews often may not be needed. We've discussed in the past that loop limits should never be based on view size, and I think so far all or most of the code written, dycore and physics, has followed that guidance.

jgfouca:
@tcclevenger , is this ready for me to review?

tcclevenger:
> @tcclevenger , is this ready for me to review?

@jgfouca Not yet, I am seeing errors on weaver in my own tests, so I'm expecting the auto tester to fail. I'll update when I get weaver to pass.

tcclevenger:
@ambrad @jgfouca Only the tridiag solver asserts that the dimension of a workspace variable (du, dl, d) is exact (scalar view_1d views of size nlev). For this, I just take a subview, otherwise I do not worry about resizing any workspace variables.

New changes are ready for review.

ambrad (approved):
Nice!

PR #888, opened 2021-02-19
Fix release tests

### LINKED ISSUE #525: DEBUG build required for tests to pass
2 programs, which account for 17 tests (the shoc test below is run with OMP_NUM_THREADS=1...16) fail with CMAKE_BUILD_TYPE=RELEASE and CMAKE_BUILD_TYPE=RelWithDebInfo, but they pass with CMAKE_BUILD_TYPE=DEBUG.  Same results with both gcc6.5 and gcc7.3.

**Program 1:** ` ekat/tests/utils/debug_tools.cpp`, error from LastTest.log

```
/ascldap/users/pabosle/scream/externals/ekat/tests/utils/debug_tools_tests.cpp:150: FAILED:
  REQUIRE( run_fpe_tests () == num_expected_fpes )
with expansion:
  0 == 1

*) testing user-disabled fpes...
 tests mask: 57
   has FE_DIVBYZERO: 0
   has FE_INVALID:   1
   has FE_OVERFLOW:  1
  - Invalid arg threw.
  - Overflow threw.
-------------------------------------------------------------------------------
fpes
  user-requested-fpes
-------------------------------------------------------------------------------
/ascldap/users/pabosle/scream/externals/ekat/tests/utils/debug_tools_tests.cpp:159
...............................................................................

/ascldap/users/pabosle/scream/externals/ekat/tests/utils/debug_tools_tests.cpp:168: FAILED:
  REQUIRE( run_fpe_tests () == num_expected_fpes )
with expansion:
  2 == 3
```

**Program 2:** `scream/src/physics/tests/shoc_varorcovar.cpp`, error from LastTest.log

```
-------------------------------------------------------------------------------
shoc_varorcovar_property
-------------------------------------------------------------------------------
/ascldap/users/pabosle/scream/components/scream/src/physics/shoc/tests/shoc_varorcovar_tests.cpp:281
...............................................................................

/ascldap/users/pabosle/scream/components/scream/src/physics/shoc/tests/shoc_varorcovar_tests.cpp:205: FAILED:
  REQUIRE( SDS.varorcovar[offset] < 0.0 )
with expansion:
  100.0 < 0.0

===============================================================================
test cases:   48 |   47 passed | 1 failed
assertions: 2981 | 2980 passed | 1 failed

KokkosP: Kernel timing written to /home/pabosle/scream/components/scream/build_master/src/physics/shoc/tests/s1024454-24861.dat 
<end of output>
Test time =   0.22 sec
----------------------------------------------------------
Test Failed.
"shoc_tests_ut_np1_omp1" end time: Aug 25 16:38 MDT
"shoc_tests_ut_np1_omp1" time elapsed: 00:00:00
----------------------------------------------------------
```

jgfouca:
With this PR, running `ctest` will work for a Release build.

Handling the BFB unit tests was a bit awkward. I decided that the approach that preserved the most value was to continue to run these tests but disable the actual bfb comparisons. Most of the value of the test is lost but we at least know the code runs without crashing.

Same issue with the run_and_cmp tests. I decided the best thing to preserve at least a little value by continuing to run these test but setting the tol to 1e15. 

Other minor things:
1) SHOC did not support both cxx and f90 baseline generation like P3 does, so I added it
2) Fix a potential uninitialized value (according to CUDA compiler)

Fixes #525

### DIFF
--- a/components/scream/src/physics/share/tests/physics_saturation_unit_tests.cpp
+++ b/components/scream/src/physics/share/tests/physics_saturation_unit_tests.cpp
@@ -107,7 +107,11 @@ struct UnitWrap::UnitTest<D>::TestSaturation
     //: C::macheps is machine epsilon for single or double precision as appropriate. This will be
     //multiplied by a condition # to get the actual expected numerical uncertainty.
 
-    static constexpr Scalar tol = C::macheps;
+    static constexpr Scalar tol = C::macheps
+#ifdef NDEBUG
+      * 10000
+#endif
+;
 
     //PMC note: original version looped over pack dimension, testing each entry. This isn't
     //necessary b/c packs were created by copying a scalar up to pack size. Thus just evaluating
--- a/components/scream/src/physics/shoc/shoc_pblintd_height_impl.hpp
+++ b/components/scream/src/physics/shoc/shoc_pblintd_height_impl.hpp
@@ -42,7 +42,7 @@ void Functions<S,D>::pblintd_height(
   const auto nlev_p = (nlev-1)%Spack::n;
 
   // Compute rino values and find max index s.t. rino(k) >= ricr
-  Int max_indx;
+  Int max_indx = -1;
 
   const Int lower_pack_indx = (nlev-npbl)/Spack::n;
   const Int upper_pack_indx = (nlev-1)/Spack::n+1;
--- a/components/scream/src/physics/p3/tests/CMakeLists.txt
+++ b/components/scream/src/physics/p3/tests/CMakeLists.txt
@@ -2,43 +2,43 @@ include(ScreamUtils)
 
 set(NEED_LIBS p3 physics_share scream_share)
 set(P3_TESTS_SRCS
-  p3_tests.cpp
-  p3_unit_tests.cpp
-  p3_ice_tables_unit_tests.cpp
-  p3_table3_unit_tests.cpp
-  p3_back_to_cell_average_unit_tests.cpp
-  p3_prevent_ice_overdepletion_unit_tests.cpp
-  p3_find_unit_tests.cpp
-  p3_upwind_unit_tests.cpp
-  p3_calc_rime_density_unit_tests.cpp
-  p3_cldliq_imm_freezing_unit_tests.cpp
-  p3_rain_imm_freezing_unit_tests.cpp
-  p3_droplet_self_coll_unit_tests.cpp
-  p3_cloud_sed_unit_tests.cpp
-  p3_cloud_rain_acc_unit_tests.cpp
-  p3_ice_sed_unit_tests.cpp
-  p3_ice_collection_unit_tests.cpp
-  p3_rain_sed_unit_tests.cpp
-  p3_dsd2_unit_tests.cpp
-  p3_rain_self_collection_tests.cpp
-  p3_autoconversion_unit_tests.cpp
-  p3_ice_relaxation_timescale_unit_tests.cpp
-  p3_calc_liq_relaxation_timescale_unit_tests.cpp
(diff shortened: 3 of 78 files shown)


components/scream/src/physics/share/tests/physics_saturation_unit_tests.cpp
```diff
@@ -107,7 +107,11 @@ struct UnitWrap::UnitTest<D>::TestSaturation
     //: C::macheps is machine epsilon for single or double precision as appropriate. This will be
     //multiplied by a condition # to get the actual expected numerical uncertainty.
 
-    static constexpr Scalar tol = C::macheps;
+    static constexpr Scalar tol = C::macheps
+#ifdef NDEBUG
+      * 10000
```
jgfouca, line 112: This test needed a ridiculous tolerance in Release build; not sure why.

components/scream/src/physics/shoc/shoc_pblintd_height_impl.hpp
```diff
@@ -42,7 +42,7 @@ void Functions<S,D>::pblintd_height(
   const auto nlev_p = (nlev-1)%Spack::n;
 
   // Compute rino values and find max index s.t. rino(k) >= ricr
-  Int max_indx;
+  Int max_indx = -1;
```
jgfouca, line 45: Uninit warning came from this line.

jgfouca (commented):
Some annotations.

components/scream/src/physics/share/tests/physics_saturation_unit_tests.cpp
```diff
@@ -107,7 +107,11 @@ struct UnitWrap::UnitTest<D>::TestSaturation
     //: C::macheps is machine epsilon for single or double precision as appropriate. This will be
     //multiplied by a condition # to get the actual expected numerical uncertainty.
 
-    static constexpr Scalar tol = C::macheps;
+    static constexpr Scalar tol = C::macheps
```
bartgol, line 110: This might be personal, but I think having `ifdef`'s cutting through an expression is a bit hard to parse. I think having

```
#ifndef NDEBUG
static constexpr Real tol = blah;
#else
static constexpr Real tol = blah*10000;
#endif
```
is easier to parse.

bartgol (approved):
I think there's a lot of value in supporting ctest runs in scream in release mode. Even if they run much fewer checks. The share folder, for instance, can be tested regardless of build type.

As for the bfb checks, I don't really like the ubiquitous ifdef method, but I don't want to argue about it here. I think having this feature in is important.

For the record, I would use one of the following approaches:
- define `SCREAM_CHECK_EQUAL_REALS(a,b)` (or something along that line), which evaluates to `REQUIRE(a==b)` in non-release builds, and to nothing in release builds. At least we'd confine the ifdefs to one place.
- start using `REQUIRE(a==Approx(b))`, where Approx is a feature from Catch, and supports giving abs and rel tolerances.

The first solution could be implemented in this PR (if you want/have the time), but I'm ok if you want to merge as is, and perhaps defer to a second PR later; while the second option is more time consuming (what is the proper abs/rel tol? can we pick same tol for all checks or should each check have its own?), and is more a suggestion for future tests.

jgfouca:
Thanks @bartgol , I will merge and begin work on SCREAM_DEBUG_EQUAL.

bartgol:
@jgfouca Now that we can run tests in release mode, should we add a build to the test-all-scream? Like the release equivalent of full_debug (i.e., double precision, no FPEs)?

jgfouca:
@bartgol , yes, adding an opt build to test-all-scream is my next task.

PR #891, opened 2021-02-23
Add release tests to test-all-scream

### LINKED ISSUE #539: Should SCREAM have release mode nightlies?
We cannot (in general) do BFB testing cxx-vs-f90 in debug mode, due to potentially different optimizations/rounding (especially with GPU builds). However, we could still check against baselines.

Among the benefits, the most important I can see is that we could catch build/runtime errors that do not happen in debug mode (e.g., some optimizations that are unsafe due to algorithm/design details).

Pinging @jgfouca as the testing master.

jgfouca:
And some pylint fixes.

Fixes #539

### DIFF
--- a/components/scream/scripts/utils.py
+++ b/components/scream/scripts/utils.py
@@ -474,9 +474,10 @@ def ensure_pip():
 
         ensurepip.bootstrap(user=True)
 
-        site.main() # needed to "rehash" available libs
+        # needed to "rehash" available libs
+        site.main() # pylint: disable=no-member
 
-        import pip
+        import pip # pylint: disable=import-error
 
 ###############################################################################
 def ensure_yaml():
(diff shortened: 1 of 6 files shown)


components/scream/scripts/utils.py
```diff
@@ -474,9 +474,10 @@ def ensure_pip():
 
         ensurepip.bootstrap(user=True)
 
-        site.main() # needed to "rehash" available libs
+        # needed to "rehash" available libs
+        site.main() # pylint: disable=no-member
 
-        import pip
+        import pip # pylint: disable=import-error
```
bartgol, line 480: I know nothing about pylint; is this comment string parsed by pylint?

bartgol (approved):
Looks good.

A comment for the future: I wonder if we should start to think about the parallelization mechanism. If we go above 4 builds, we will have a pb on weaver (where there are 4 available devices). Currently, the python driver dispatches all builds in parallel, but with, say, 5 builds, one build will never be able to run anything on weaver (no available devices). And even at 4 (and actually even now with 3), some builds on weaver run with only 1 device. That means no "real" MPI testing, which is sub-optimal.

Ideally, I'd like the 3 (about to be 4) builds to "share" resources, but that is not possible. I might try to fiddle with cmake multiple builds feature, and see if we can somehow hijack it for our purposes.

components/scream/scripts/utils.py
jgfouca: Yes, it suppresses a pylint error. Pylint can have false positives.

jgfouca:
@bartgol , can you add a final review? I had to add a few commits to get this to work on Blake.

bartgol (approved):

Issue #892, opened 2021-02-24
Python script to create initial condition file for scream testing


bartgol:
Talking with @AaronDonahue (following on conversations happened during the v1 telecon), we decided that it would be best to have SCREAM set initial conditions _always_ through an input/restart file, even for scream-standalone mode, even when running just a small test (like dynamics-only, or homme+p3). This would allow us to get rid of the (frankly) clumsy FieldInitializer struct, and all its quirks and oddities.

This plan, however, brings up a question: what if someone wants to do a slightly different test? Say change something in the initial condition, or add a variable, or change a parametrization, or change resolution, or something else along these lines? We would like to have a lightweight python script that makes things easy. What we envisioned is something that allows me to do this:

```
# generate init condition netcdf file at ne4 with 128 levs and phys grid corresponding to gll nodes, with all vars inited to 0
./gen-input-file -o filename --ne 4 --pg=gll --nlev 128 --vars pressure temperature velocity water_vapor
# modify $filename by setting pressure=1, temperature to the result of given expression, and importing velocity from other_file
./mod-input-file -f filename --set-val pressure=1.0 --compute temperature="lat^2 - lon^2" --import velocity=other_file[:velocity_name_in_other_file]
```

This script is obviously meant to be used mostly for small test cases, supporting limited features (the fanciest being the `--compute` flag, which could be extended to use other fields currently stored, like `--compute exner="(p/p0)^1.3"`). Since its goal is for small testing purposes, the "quality" of generated data is not "that" important. The goal is simply to give devs a fast and simple way to generate an input file that will allow to bootstrap their testcase, avoiding invalid/incompatible data.

@PeterCaldwell @rljacob I suspect there might already be some tool for this in the e3sm ecosystem. If so, then we might simply hijack it for our purposes, perhaps wrapping it in a smaller script for the sake of simple scream-related usage.

PeterCaldwell:
I think I like this idea but I'm not sure I understand it. A couple thoughts:

1. the variables needed to run P3 standalone (for example) are different than the variables which would be needed to initialize the whole model (because many of P3's variables are calculated by processes called after the model starts but before p3 is called). Thus having a single utility for providing inputs to run any kind of simulation could end up getting really complicated/weird. 
2. this is a bit of a tangential thought, but it would be really good if all initial conditions and boundary conditions were online-interpolated to the target resolution. This would get rid of a huge quagmire in E3SM right now where you can't run at a new resolution because you never have the right files. 
[...]

PR #893, opened 2021-02-24
Various fixes to remap, dynamics, and initialization.

bartgol:
This branch was originally meant to revive the homme+p3 test, but after discussion with @AaronDonahue , we postponed that task until after we have completed the change regarding initial condition (see also issue #892 ).

However, several bugs and other things to fix were found during our try to revive homme+p3. This PR is meant to merge those fixes in.


components/scream/scripts/scream-env-cmd
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/env python3
+
+"""
+Dump the SCREAM-approved environment setup commands for a machine
```
AaronDonahue, line 4: Isn't this part of a different PR somewhere that Jim is doing?

bartgol: Ah, something must be off with the base branch. I'll rebase on master and force push.

bartgol: Ok, that seems to have fixed it.

components/scream/src/share/atm_process/atmosphere_process_dag.hpp
```diff
@@ -38,18 +39,25 @@ class AtmProcDAG {
   // If already stored, simply return its position
   int add_fid (const FieldIdentifier& fid);
 
+  int get_fid_id (const FieldIdentifier& fid) const;
```
jeff-cohere, line 42: Is this "field identifier ID" supposed to be an ID, or an index (of the identifier within a list, say)? If the latter, maybe we could call this `get_fid_index`.

components/scream/src/dynamics/homme/tests/homme_pd_remap_tests.cpp
```diff
@@ -268,35 +252,28 @@ TEST_CASE("remap", "") {
...
-        auto h_ss_3d_view = Kokkos::create_mirror_view(ss_3d_view);
-        auto h_vs_3d_view = Kokkos::create_mirror_view(vs_3d_view);
-        auto h_ts_3d_view = Kokkos::create_mirror_view(ts_3d_view);
+        auto h_s_2d_view = s_2d_field_dyn.get_reshaped_view<Homme::Real***,Host>();
+        auto h_v_2d_view = v_2d_field_dyn.get_reshaped_view<Homme::Real****,Host>();
+        auto h_s_3d_view = s_3d_field_dyn.get_reshaped_view<Homme::Real****,Host>();
+        auto h_v_3d_view = v_3d_field_dyn.get_reshaped_view<Homme::Real*****,Host>();
+        auto h_ss_3d_view = ss_3d_field_dyn.get_reshaped_view<Homme::Real*****,Host>();
+        auto h_vs_3d_view = vs_3d_field_dyn.get_reshaped_view<Homme::Real******,Host>();
+        auto h_tr_3d_view = tr_3d_field_dyn.get_reshaped_view<Homme::Real*****,Host>();
```
jeff-cohere, line 275: Honest question: should there be one more `*` here (`Homme:Real******`)?

components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
```diff
@@ -60,6 +60,17 @@ void HommeDynamics::set_grids (const std::shared_ptr<const GridsManager> grids_m
       "       does not match the number of elements internal in Homme.\n");
   const int nmf = get_homme_param<int>("num momentum forcings");
 
+  // Note: time levels are just an expedient used by Homme to
```
jeff-cohere, line 63: Nice explanation.

components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
```diff
@@ -127,29 +148,27 @@ set_required_group (const FieldGroup<const Real>& group)
...
-    Field<Real> Q(Q_fid);
-    Q.get_header().get_alloc_properties().request_allocation<Homme::Scalar>();
-    Q.allocate_view();
-    m_dyn_fields_out.emplace("Q",Q);
-  }
+  const auto& name = group.m_info->m_group_name;
+  EKAT_REQUIRE_MSG(name=="TRACERS",
+    "Error! We were not expecting a field group called '" << name << "\n");
+
+  EKAT_REQUIRE_MSG(group.m_info->size()>0,
```
jeff-cohere, line 155: Fun fact: this could also be written

```
EKAT_REQUIRE_MSG(not group.m_info->empty(),
...
```

jeff-cohere:
Nice! I'm getting excited about the dynamic tracers bookkeeping--we'll definitely need this for aerosols.

I approve. Since this is tagged with "AT:AUTOMERGE", I'll refrain from the green checkmark to let @AaronDonahue give any other feedback he might have.

components/scream/src/physics/p3/p3_inputs_initializer.cpp
```diff
@@ -76,21 +76,29 @@ void P3InputsInitializer::initialize_fields ()
...
   EKAT_REQUIRE_MSG(count!=0,"Error in p3_inputs_initializer: no fields have declared this initializer.  Check p3 interface."); 
 
-  EKAT_REQUIRE_MSG (count==(int)fields_to_init.size(),
-    "Error! P3InputsInitializer is expected to init " + std::to_string(fields_to_init.size()) + " fields:\n"
+  EKAT_REQUIRE_MSG (count==fields_to_init.size() || (count+1)==fields_to_init.size(),
+    "Error! P3InputsInitializer is expected to init the following fields:\n"
     "       " + list_of_fields + "\n"
-    "       Instead found " + std::to_string(count) + " fields.\n"
+    "       (possibly with the exception of dp). Instead, it is initializing these fields:\n"
+    "       " + inited_fields + "\n"
```
AaronDonahue, line 101: Nice update to the error message, much more helpful.

components/scream/src/physics/p3/p3_inputs_initializer.cpp
```diff
@@ -117,6 +124,12 @@ void P3InputsInitializer::initialize_fields ()
   auto h_nccn_prescribed = m_fields.at("nccn_prescribed").get_reshaped_view<Pack**,Host>();
   auto h_inv_qc_relvar   = m_fields.at("inv_qc_relvar").get_reshaped_view<Pack**,Host>();
 
+  // dp might be inited by Homme.
+  decltype(h_pmid) h_dp;
+  if (m_fields.count("dp")==1) {
+    h_dp = m_fields.at("dp").get_reshaped_view<Pack**,Host>();
+  }
+
```
AaronDonahue, line 132: So, presumably, a subsequent PR will get rid of this whole paradigm right?  And we won't have `p3_inputs_initializer` at all.

components/scream/src/share/field/field_repository.hpp
```diff
@@ -412,6 +414,13 @@ registration_ends (const std::shared_ptr<const GridsManager>& gm) {
     }
   }
 
+  // Homme (and possibly other places in EAM) hard-codes water vapor as the 1st
+  // tracer in Q. Therfore, find qv in the TRACERS group, and make sure it's the 1st
+  auto qv_pos = ekat::find(tr_gr.m_fields_names,"qv");
```
AaronDonahue, line 419: will this spit out an error if "qv" is not found at all?  On that same note, many of the unit tests probably won't have "qv", instead they'll have dummy variable names, but presumably we may create a test where dummy variable names are still treated like tracers.  

So long question short.  What will happen if we have some tracers and none of them are names "qv"?

components/scream/tests/scream_homme_dyn_p3/CMakeLists.txt
```diff
@@ -14,30 +14,43 @@ configure_file(${CMAKE_CURRENT_SOURCE_DIR}/input.yaml
...
+# Discretization/algorithm settings
 set (HOMME_TEST_NE 2)
-set (HOMME_TEST_NDAYS 1)
 set (HOMME_TEST_LIM 9)
 set (HOMME_TEST_QSIZE 6)
 set (HOMME_TEST_RSPLIT 3)
-set (HOMME_TEST_MOISTURE dry)
+set (HOMME_TEST_NDAYS 1)
+set (HOMME_TEST_TIME_STEP 600)
+SET (HOMME_THETA_FORM 1)
```
AaronDonahue, line 25: not important, but it looks like up above you specifically switched upper-case "SET" to lower-case "set" as a change and here you are switching between upper and lower case in this list of options.  Is there any difference in cmake?

AaronDonahue: same for configure_file below...

components/scream/src/dynamics/homme/tests/homme_pd_remap_tests.cpp
```diff
@@ -117,29 +113,31 @@ TEST_CASE("remap", "") {
   std::vector<FieldTag> v_3d_dyn_tags   = {EL,     CMP, GP, GP, VL};
   std::vector<FieldTag> ss_3d_dyn_tags  = {EL, TL,      GP, GP, VL};
   std::vector<FieldTag> vs_3d_dyn_tags  = {EL, TL, CMP, GP, GP, VL};
+  std::vector<FieldTag> tr_3d_dyn_tags  = {EL,     VAR, GP, GP, VL};
```
AaronDonahue, line 116: Will all of these references to "VL" need to be replaced now that #897 will be going in?

components/scream/src/dynamics/homme/tests/homme_pd_remap_tests.cpp
```diff
@@ -598,7 +572,577 @@ TEST_CASE("remap", "") {
...
+  // Note on prefixes: s=scalar, v=vector, ss=scalar state, vs=vector_state, ts=tracer state
+
+  // Create tags and dimensions
+  std::vector<FieldTag> s_2d_dyn_tags   = {EL,          GP, GP    };
+  std::vector<FieldTag> v_2d_dyn_tags   = {EL,     CMP, GP, GP    };
+  std::vector<FieldTag> s_3d_dyn_tags   = {EL,          GP, GP, VL};
+  std::vector<FieldTag> v_3d_dyn_tags   = {EL,     CMP, GP, GP, VL};
+  std::vector<FieldTag> ss_3d_dyn_tags  = {EL, TL,      GP, GP, VL};
+  std::vector<FieldTag> vs_3d_dyn_tags  = {EL, TL, CMP, GP, GP, VL};
+  std::vector<FieldTag> tr_3d_dyn_tags  = {EL,     VAR, GP, GP, VL};
```
AaronDonahue, line 664: same comment here for "VL"

AaronDonahue (commented):
Thanks for putting in all of these bugfixes @bartgol .  I had a few questions before I'm ready to hit "approve".

components/scream/src/physics/p3/p3_inputs_initializer.cpp
bartgol: Right. I'm already working on a branch. Purging all those files/lines feels so good. :)

components/scream/src/share/field/field_repository.hpp
bartgol: It shouldn't. This simply looks for it, and, if found, it makes sure `qv` is the 1st tracer. If not found, no big deal.

components/scream/tests/scream_homme_dyn_p3/CMakeLists.txt
bartgol: Ah, yes, thanks. HOMME uses mostly upper case, while scream seems to use only lower case (which I like better, btw). Copying and pasting from HOMME left some unchanged lines.

components/scream/src/dynamics/homme/tests/homme_pd_remap_tests.cpp
bartgol: Good point. I'll rebase on master to bring those in.

components/scream/src/share/atm_process/atmosphere_process_dag.hpp
bartgol: Good point. I'll change this.

components/scream/src/dynamics/homme/tests/homme_pd_remap_tests.cpp
bartgol: Well, no. Before, we did have 6 `*`, cause I was considering for input/output the tracers mass, which is a 'state' in homme, so it has time levels: `Qdp(ie,itl,iq,igp,jgp,ilev)`. Now we consider for input/output the tracers concentration, which in homme is computed "diagnostically" from `Qdp` and `dp`, and has layout `Q(ie,iq,igp,jgp,ilev)`.

components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
bartgol: Ah, right. I'll change to that, cause is faster to parse visually.

bartgol: I take it back. `*m_info` is a `FieldGroupInfo` struct, not a std set. I added the `size()` method, to avoid quering the list of names stored internally, but not an 'empty' one. But maybe it's a verbose method worth adding.

bartgol:
I decided to keep FT, FQ, and FM in `HommeDyamics`, even though, as I recall, we are moving toward a "if you need tendencies, back them out from the state" kind of approach. I decided that we can do that modificaiton in a following PR.

AaronDonahue (approved):

PR #896, opened 2021-02-26
SCREAM: first stab at creating simple nc files for AD tests initial conditions

bartgol:
Last night I spent ~1h on a stab at a script that creates initial conditions files, as discussed in #892. It is a much simpler version, in that it does not compute geometric data like lat/lon (too complicated for a 1h project, I realized), and it does not allow to compute fields based on math expressions (also too complicated for a 1h project). It also does not allow to 'import' a var from another file yet. That should not be too hard, maybe another 30-60min.

So far, the script allows to create a simple nc file, with 1d (mid or interface levels info), 2d (no levesl), or 3d (mid or interface levels) variables, scalar or vector. The user can provide a uniform scalar value for each field; if none is provided, the field is inited to 0.

The math expression feature might be easy-ish to add, infolving other existing vars in the file. However, allowing geometry-dependent expressions (like `v="lat+lon-z"`) seems to be more complicated, cause we have to make sure our dofs are listed in the same order as done by Homme. And doing stuff like `v="T + p*rho"` only becomes interesting if the vars in the rhs are not constants (otherwise one can do the math once), so only if we add the ability to import vars from another file (like an output of a simulation). If we do that, then we can maybe import also geometry fields from that output, which would be easier.

Remaining (reasonably quick) tasks:

1. add a variable group 'TRACERS', to mirror organization in the FM
2. add ability to import variable from other nc file.
3. (possibly) add ability to compute a var from other variables (only meaningful after 2. is done).

NOTE: the script requires the [netcdf-python](https://unidata.github.io/netcdf4-python/) package, which I simply installed with `pip install netCDF4` on my laptop in 3 seconds.


jgfouca (changes requested):
Overall, looks good. IMO the one method I commented on needs a refactor to reduce code duplication.

components/scream/scripts/nc_file_init.py
```diff
@@ -0,0 +1,255 @@
...
+        return ds
+
+    ###########################################################################
+    def add_variables(self,ds):
+    ###########################################################################
+        # re.match(r'^\w+$', string) is a very compact regexp check, to ensure
+        # the string only contains alphanumeric chars and underscores
+
+        # 1d variables
+        for item in self._1d_mid_s:
```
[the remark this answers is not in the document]
bartgol: I tried to see if I could delegate to a fcn the creation, but I couldn't find a satisfactory way, given the difference in variable names. I'll work on your idea, and see if I can make something more compact.

jgfouca (approved):
Much better!

bartgol:
It was brought to my attention that [NCO](http://research.jisao.washington.edu/data/nco/) already offers support for some nc file manipulation. It might be a good idea to see if we can have a python script piggy backing on that package for file manipulation.

The creation of an empty nc file, possibly with some vars with fixed value was easy enough to do it manually (meaning, with a small python script). But perhaps all other manipulation could be done with NCO.

bartgol:
I simplified the creation of empty variables a bit. Also, following Peter's suggestion, I integrated a bit with [NCO](http://nco.sourceforge.net/), which made some kind of variable manipulation easier.

With the current capabilities, if we can import geometry variables from another nc file (e.g., lat, lon, z), then we can build all sorts of input files using the `--compute-vars` option.

components/scream/scripts/nc_file_init.py
```diff
@@ -0,0 +1,359 @@
...
+        if vals_str=="":
+            # User did not specify values. Use all zeros.
+            return [0]*vec_dim
+        else:
+            try:
+                # Try to convert vals_str to a single float. If successful,
+                # the user passed a single value for all vector components
+                value = float(vals_str)
+                return [value]*vec_dim
+            except:
```
jgfouca, line 207: Minor point, python recommends against bare except statements. You should catch the exception type that can arise through semi-normal execution (in this case, it's the failure to convert str->float which is a ValueError) while leaving more surprising exceptions uncaught.

components/scream/scripts/nc_file_init.py
```diff
@@ -0,0 +1,359 @@
...
+                expect (close>open,"Error! Var values specification should be '..=val' or '..=[val1,...,valN]'.")
+                expect (close==len(vals_str)-1,"Error! Var values specification should be '..=val' or '..=[val1,...,valN]'.")
+                vals = vals_str[open+1:close].split(',')
+                expect (len(vals)==vec_dim,
+                        "Error! Var values specification has the wrong length: {} instead of {}."
+                        .format(len(vals),vec_dim))
+                try:
+                    values = [float(v) for v in vals]
+                    return values
+                except:
```
jgfouca, line 222: Same

jgfouca (changes requested):
Couple minor points regarding pythonic exception handling. Also, now that you're writing python, you should get familiar with pylint (I usually do `pylint --disable C --disable R $file`) which would have poked you about those bare except statements.

components/scream/scripts/nc_file_init.py
bartgol: Good point.

bartgol:
Ok, this is starting to look like a more complete tool now. It mostly relies on nco's tools for populating the nc file.

I split the script in two, since creating an empty nc file (with only dimensions an global attributes) is probably done once. Might as well keep the syntax (and arguments list) for the actual manipulation shorter.

The last few commits add the possibility of remapping fields from an existing nc file with different resolution. This feature relies on ncremap, which requires esmf, as well as map files.

As an exercise, I generated a small ne4np4 file, importing some homme inputs. I also remapped them to ne30 and back to ne4, to check that the tool worked without crashing. But more testing/usage might reveal other flaws.

PeterCaldwell:
Apologies for ignoring this task then dropping in with off-the-wall suggestions... One question is whether this PR should be in the scream repo or whether it belongs in scream-docs since it creates tools rather than affecting the scream .exe. I think I asked this before in a different context and we decided scream repo is the way to go... Another thought is that I think we should invite Charlie Zender to review your PR in case he sees easier ways to do what you're doing. I'll send him an email to provide context before we send him a review request - let me know when to do that.

bartgol:
@PeterCaldwell I wouldn't say you ignore this task. You gave me feedback (was it on slack?) and led me to find out about nco, which greatly simplified the script.

I don't know if the script belongs here on in scream-docs. I think perhaps scream-docs is more appropriate, but maybe scream makes it more practical? I don't have a strong opinion, either way is fine with me.

If Charlie wants to look at this, that's great. I don't expect him to look at the python code itself, but maybe he has some advice on what nco tool should be used for each use case.

PeterCaldwell:
Sounds good. Where to put helper scripts sounds like a good topic for tomorrow's v1 telecon. Let me know when to ask Charlie for a review...

jgfouca (approved):

components/scream/scripts/nc_file_init.py
```diff
@@ -0,0 +1,47 @@
...
+###########################################################################
+
+    expect (phys_grid=="gll", "Error! So far, only GLL physics grid supported.")
+
+    ofile = pathlib.Path(output_file).resolve().absolute()
+    expect ( (not ofile.exists()) or overwrite,
+            "Error! Output file '{}' already exists. Use -o to overwrite.".format(ofile))
+
+    ds = Dataset(ofile,"w",format="NETCDF4",persist=True)
+
```
PeterCaldwell, line 18: A weird point, but E3SM (and I think SCORPIO and therefore SCREAM) only supports netcdf3. I think we probably want 'NETCDF3_64BIT_OFFSET' but @wlin7 or @brhillman may remember differently.

bartgol: Interesting, I didn't know that.

czender (approved):
I looked at this PR and it looked good to me. My Python skills are infantile so I just tried to see if there were any obvious NCO speedups that could be obtained and I did not notice any. So, good job AFAICT.

bartgol:
@czender thanks for taking the time to look at it!

bartgol:
Merging, since all I changed since last approval was the netcdf format, following Peter's suggestion.

PR #897, opened 2021-03-01
SCREAM: created separate field tags for midpoints and interface levels

### LINKED ISSUE #648: Create separate FieldTag's for level midpoint and level interface
This helps distringuishing fields defined at level midpoints and fields defined at level interfaces.

Notice that these are already "different", even with the current VERTICAL_LEVEL tag, due to the different extent in the last dimension. However, having a different tag helps to reduce typos, and increases the differences between the two fields.

bartgol:
Two separate tags are helpful for I/O, where we need two different dimensions registered in the nc files. It also makes it simpler to spot errors, and makes the code a bit more verbose.

Note: the tags short names (which always match to the string you get with `e2str(tag)`) can be changed. I thought `LEV` and `ILEV` were clear enough, but we can use more verbose ones, like `MID_LEV` and `INT_LEV`, if you prefer.

Closes #648.

### DIFF
--- a/components/scream/src/control/tests/dummy_atm_proc.hpp
+++ b/components/scream/src/control/tests/dummy_atm_proc.hpp
@@ -56,10 +56,10 @@ public:
     std::vector<FieldTag> tags;
     std::vector<int> dims;
     if (m_grid->name()=="Point Grid A") {
-      tags = {COL,VL};
+      tags = {COL,LEV};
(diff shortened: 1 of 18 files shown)


bartgol:
Added WIP label, waiting for some time to debug this.

AaronDonahue (approved):
Looks good, thanks for putting this together.

PR #899, opened 2021-03-02
add shoc_init

tcclevenger:
Ports `shoc_init()` to calculate `npbl` for `shoc_main()`.

In shoc.F90, `npbl` is a global index, not an output of `shoc_init()`. For bfb testing purposes, I just included `shoc_init()` in the `shoc_main_bfb` test and made sure to hit each of the main scenarios, namely `npbl = 1`, `npbl = nlev`, `1 < npbl < nlev`.

### DIFF
--- a/components/scream/src/physics/shoc/shoc_main_impl.hpp
+++ b/components/scream/src/physics/shoc/shoc_main_impl.hpp
@@ -16,6 +16,49 @@ namespace shoc {
  */
 
 template<typename S, typename D>
+Int Functions<S,D>::shoc_init(
+  const Int&                  nbot_shoc,
+  const Int&                  ntop_shoc,
+  const view_1d<const Spack>& pref_mid)
+{
+  // This function calculates the maximum number of levels
+  // in pbl from surface
+
+  using ExeSpace = typename KT::ExeSpace;
+  view_1d<Int> npbl_d("npbl",1);
+
+  const auto policy = ekat::ExeSpaceUtils<ExeSpace>::get_default_team_policy(1, 1);
+  Kokkos::parallel_for(policy, KOKKOS_LAMBDA(const MemberType& team) {
+
+    const Scalar pblmaxp = SC::pblmaxp;
+
+    int npbl_val = 0;
+
+    const int begin_pack_indx = ntop_shoc/Spack::n;
+    const int end_pack_indx   = nbot_shoc/Spack::n+1;
+    Kokkos::parallel_reduce(Kokkos::TeamThreadRange(team, begin_pack_indx, end_pack_indx),
+                                                    [&] (const Int& k, Int& local_sum) {
+      const auto range = ekat::range<IntSmallPack>(k*Spack::n);
+      const auto in_range = (range >= ntop_shoc && range < nbot_shoc);
+      const auto condition = (in_range && pref_mid(k) >= pblmaxp);
+
+      IntSmallPack greater_than_pblmaxp(0);
+      greater_than_pblmaxp.set(condition, 1);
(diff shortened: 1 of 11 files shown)


components/scream/src/physics/shoc/shoc_main_impl.hpp
```diff
@@ -15,6 +15,49 @@ namespace shoc {
...
+  const auto policy = ekat::ExeSpaceUtils<ExeSpace>::get_default_team_policy(1, 1);
+  Kokkos::parallel_for(policy, KOKKOS_LAMBDA(const MemberType& team) {
+
+    const Scalar pblmaxp = SC::pblmaxp;
+
+    int npbl_val = 0;
+
+    const int begin_pack_indx = ntop_shoc/Spack::n;
+    const int end_pack_indx   = nbot_shoc/Spack::n+1;
+    Kokkos::parallel_reduce(Kokkos::TeamThreadRange(team, begin_pack_indx, end_pack_indx),
```
ambrad, line 39: A reduction that sums over the 1-mask of levels satisfying pref_mid >= pblbmaxp is fine, but perhaps a more natural alternative would be a reduction on the max level. Just mentioning it in case you don't know you can do team-level max-reductions.

ambrad (approved):

components/scream/src/physics/shoc/shoc_main_impl.hpp
tcclevenger: From the fortran code 
```
   npbl = 0
   do k=nbot_shoc,ntop_shoc,-1
      if (pref_mid(k) >= pblmaxp) then
         npbl = npbl + 1
      end if
   end do
   npbl = max(npbl,1)
```
A max-reduction assume that `pref-mid` is strictly increasing? Is this the case? If so, then I agree a max-reduction is more natural and I will change the bfb test to have an increasing `pref_mid` (currently it is only random values).

ambrad: Yes, pref_mid is monotonically increasing. Also, note that the use of npbl later, e.g., in loops of the sort `k=nlev-1,nlev-npbl+1,-1`, implies npbl refers to a contiguous set of levels extending from the surface.

tcclevenger: @ambrad changes done. Now calculating "max levels from surface in pbl" and changed the test to have monotonic pref_mid.

ambrad: Are you testing the case that the reduction would never update npbl_val? I believe the value you initialize of `npbl_val = 1` at line 35 will not be seen in the parallel_reduce, so on exit npbl_val would be undefined in the case that the F90 would give 0.

components/scream/src/physics/shoc/shoc_main_impl.hpp
```diff
@@ -15,6 +15,49 @@ namespace shoc {
...
+                                                    [&] (const Int& k, Int& local_max) {
+      auto range = ekat::range<IntSmallPack>(k*Spack::n);
+      const auto condition = (range >= ntop_shoc && range < nbot_shoc && pref_mid(k) >= pblmaxp);
+
+      auto levels_from_surface = nbot_shoc - range;
+      levels_from_surface.set(!condition, 1);
```
tcclevenger, line 45: @ambrad `levels_from_surface` is set to 1 for all indices which do not meet the condition, so that would be the minimum return. I test this case in the bfb test.

ambrad (approved):

PR #901, opened 2021-03-03
Make it possible to pull input from netCDF indendent of the field manager

AaronDonahue:
This PR revises the scorpio_input.hpp interface to allow for the reading of input from a file, independent of the field manager.

Is a first step towards finishing the final task in #864.


AaronDonahue:
@bartgol , It looks like the tests pass except on weaver, which is probably related to weaver not even running the nightly tests.  So I decided to ask for a review now so that things can go in once weaver is working again.

components/scream/src/share/io/scorpio_input.hpp
```diff
@@ -96,10 +96,19 @@ class AtmosphereInput
...
+    m_comm       = comm;
+    m_grid_name  = grid_name;
+    m_gm         = gm;
+  }
 
   // Main Functions
   void pull_input();
   void pull_input(const std::string name, view_type view_out);
+  void pull_input(const std::string filename, const std::string var_name, const std::vector<std::string> var_dims,
+                                        const bool has_columns, const std::vector<int> dim_lens, const int padding, Real* data);
```
bartgol, line 111: Indentation here is a bit funky.

components/scream/src/share/io/scorpio_input.hpp
```diff
@@ -143,6 +153,28 @@ class AtmosphereInput
  *  also outward facing but do not need those respective sections of the AD, 
  *  i.e. during ad.init, ad.run and ad.finalize.
  */
+/* ---------------------------------------------------------- */
+inline void AtmosphereInput::pull_input(const std::string filename, const std::string var_name, const std::vector<std::string> var_dims,
```
bartgol, line 157: Rather than forcing the user to figure out the padding, to pass it to the fcn, we could make this fcn templated on the ValueType, and have the last arg be `ValueType* data`. Then, the padding would be computed with
```
constexpr int pack_size = sizeof(ValueType) / sizeof(Real);
const int padding = PackInfo<pack_size>::padding(dim_lens.back());
...
auto data_real = reinterpret_cast<Real*>(data);
grid_read_data_array(filename,var_name,dim_lens,data_length,padding,data_real);
```
The `PackInfo` helper class is in `ekat_pack_utils.hpp`.

components/scream/src/share/io/scorpio_input.hpp
```diff
@@ -109,10 +118,12 @@ class AtmosphereInput
 
 protected:
   // Internal functions
-  void register_dimensions(const std::string name);
   void register_variables();
   void set_degrees_of_freedom();
   void register_views();
+  std::vector<std::string> get_vec_of_dims(const std::vector<FieldTag> tags);
+  std::string get_io_decomp(std::vector<std::string> vec_of_dims);
```
bartgol, line 125: Albeit not really important, as a C++ programmer it hurts to see std::vector's and std::string' passed by value instead of by reference...

components/scream/src/share/io/scorpio_input.hpp
bartgol: Btw, I'm assuming that `dim_lens` are the 'physical' dimensions, without the padding, right? If not, then scratch my comment above.

components/scream/src/share/io/scorpio_input.hpp
```diff
@@ -143,6 +153,28 @@ class AtmosphereInput
  *  also outward facing but do not need those respective sections of the AD, 
  *  i.e. during ad.init, ad.run and ad.finalize.
  */
+/* ---------------------------------------------------------- */
+inline void AtmosphereInput::pull_input(const std::string filename, const std::string var_name, const std::vector<std::string> var_dims,
+                                        const bool has_columns, const std::vector<int> dim_lens, const int padding, Real* data)
```
bartgol, line 158: Here too: pass std stuff by ref.

components/scream/src/share/io/scorpio_input.hpp
```diff
@@ -293,19 +298,39 @@ inline void AtmosphereInput::register_variables()
...
-    std::reverse(vec_of_dims.begin(),vec_of_dims.end()); // TODO: Reverse order of dimensions to match flip between C++ -> F90 -> PIO, may need to delete this line when switching to fully C++/C implementation.
+    std::vector<std::string> vec_of_dims = get_vec_of_dims(fid.get_layout().tags());
+    std::string io_decomp_tag           = get_io_decomp(vec_of_dims);
     get_variable(m_filename, name, name, vec_of_dims.size(), vec_of_dims, PIO_REAL, io_decomp_tag);  // TODO  Need to change dtype to allow for other variables.  Currently the field_repo only stores Real variables so it is not an issue, but in the future if non-Real variables are added we will want to accomodate that.
     //TODO: Should be able to simply inquire fromt he netCDF the dimensions for each variable.
   }
   if (m_is_rhist) { get_variable(m_filename,"avg_count","avg_count",1,{"cnt"}, PIO_INT, "cnt"); }
 } // register_variables
 /* ---------------------------------------------------------- */
+inline std::vector<std::string> AtmosphereInput::get_vec_of_dims(const std::vector<FieldTag> tags)
```
bartgol, line 309: Pass by ref.

components/scream/src/share/io/scorpio_input.hpp
```diff
@@ -293,19 +298,39 @@ inline void AtmosphereInput::register_variables()
...
+  std::vector<std::string> vec_of_dims;
+  for (auto& tag_ii : tags)
+  {
+    vec_of_dims.push_back(e2str(tag_ii));
+  }
+  std::reverse(vec_of_dims.begin(),vec_of_dims.end()); // TODO: Reverse order of dimensions to match flip between C++ -> F90 -> PIO, may need to delete this line when switching to fully C++/C implementation.
+  return vec_of_dims;
+}
+/* ---------------------------------------------------------- */
+inline std::string AtmosphereInput::get_io_decomp(std::vector<std::string> vec_of_dims)
```
bartgol, line 321: Pass by ref.

components/scream/src/share/io/scorpio_input.hpp
```diff
@@ -380,10 +381,41 @@ inline void AtmosphereInput::set_degrees_of_freedom()
     m_dofs.emplace(std::make_pair("avg_count",1));
   }
   /* TODO: 
-   * Adjust DOF to accomodate packing for fields 
    * Gather DOF info directly from grid manager
   */
 } // set_degrees_of_freedom
 /* ---------------------------------------------------------- */
+inline std::vector<Int> AtmosphereInput::get_var_dof(const int dof_len, const bool has_cols)
```
bartgol, line 388: Do all of these methods _need_ to be inlined? Usually, you inline _very small_ methods (a couple of lines), which are expected to be called quite a few times. If they are long methods and/or not critical for performance, we can simply put their impl in a cpp file. Helps to keep the headers smaller (i.e., easier/faster to parse).

bartgol (changes requested):
I had a few comments. Some are not too important though.

AaronDonahue:
@bartgol , thank you for the review.  This is why it pays to ask a c++ programmer to take a look.  I get to learn a better way to do things.  I'll make the changes you requested and ping you when I've pushed them.

components/scream/src/share/io/scorpio_input.hpp
AaronDonahue: should I pass all `std::` objects by reference?

bartgol: Usually, yes.

AaronDonahue:
@bartgol , I think I've addressed all of your comments in this latest version.  I took the liberty of applying the treatment you asked for in scorpio_input.hpp to the scorpio_output.hpp because they are fairly similar and as long as we have improving the input code on our mind we should tackle the output code because we probably won't think of it as we move to the next task.

As for the template setup, can you check what I've done there in particular?  It's the first time I've tried setting up a template, so I may have made something that works but isn't perfect.

components/scream/src/share/io/scorpio_input.cpp
```diff
@@ -0,0 +1,277 @@
...
+/* ---------------------------------------------------------- */
+void AtmosphereInput::pull_input(const std::string& name, view_type view_out)
+{
+/*  Run through the sequence of opening the file, reading input and then closing the file.  
+ *  Overloaded case to deal with just one output and to not put output into field repo.
+ *  This is an inefficient way to handle this special case because the same file would need
+ *  to be opened and closed multiple times if there is more than one variable.  TODO: Make this
+ *  a lot more efficient, most likely will need to overhaul the "pull_input" paradigm.
+ */
+  auto l_view = run(name);
```
bartgol, line 43: It seems like there is an unnecessary deep_copy here. I think you could remove the `pull_input` call, and have the `run` method return a `view_1d<const Real>`. You can construct it on the fly at the end of the run method (a simple (shallow) copy ctor call). Then, if the user wants to modify it, he/she can deep copy it into a non-const one; but if a const one is fine, you save a deep copy.

components/scream/src/share/io/scorpio_input.cpp
```diff
@@ -0,0 +1,277 @@
...
+  using namespace scream::scorpio;
+
+  // Cycle through all fields and register.
+  for (auto const& name : m_fields)
+  {
+    auto fmap = m_field_repo->get_field(name, m_grid_name);
+    // If the "averaging type" is instant then just need a ptr to the view.
+    auto view_d = fmap.get_view();
+    // Create a local copy of view to be stored by input stream.
+    auto view_copy_h = Kokkos::create_mirror_view( view_d );
```
bartgol, line 203: What's the point of `view_copy_h`? On GPU, creating a host mirror means allocating memory. But here you only use it for the extent, which is the same as `view_d`. You can just query `view_d` for the extent.

components/scream/src/share/io/scorpio_input.cpp
```diff
@@ -0,0 +1,277 @@
...
+    view_type view_copy("",view_copy_h.extent(0));
+    Kokkos::deep_copy(view_copy, view_d);
+    m_view_local.emplace(name,view_copy);
```
bartgol, line 206: Some users might need the host view from a field. Instead of creating two host mirrors, you should use the one inside fmap (btw, fmap is a field, not a map, the name is probably misleading). You could do this:

`m_view_local.emplace(name,fmap.get_view<Host>());`

Now, what's inside your m_view_local is the same that is in the host mirror inside fmap (saving a host view). And you can always sync them with `fmap.sync_to_host()` and `fmap.sync_to_dev()`.

components/scream/src/share/io/scorpio_input.cpp
```diff
@@ -0,0 +1,277 @@
...
+/* 
+ * Use information from the grids manager to determine which indices for each field are associated
+ * with this computational core.
+ */
+  using namespace scream;
+  using namespace scream::scorpio;
+  // Cycle through all fields and set dof.
+  for (auto const& name : m_fields)
+  {
+    auto fmap = m_field_repo->get_field(name, m_grid_name);
```
bartgol, line 226: Same: fmap is not a map, it's a single field. Name is confusing.

components/scream/src/share/io/scorpio_input.cpp
```diff
@@ -0,0 +1,277 @@
...
+{
+  std::vector<Int> var_dof;
+  int num_cols;
+  // Gather the column degrees of freedom for this variable
+  // Total number of values represented by this rank for this field is given by the dof_len.
+  // For a SCREAM Physics grid, only the total number of columns is decomposed over MPI ranks.  The global id (gid)
+  // is stored here as gids_hst. Thus, for this field, the total number of dof's in the other dimensions (i.e. levels)
+  // can be found by taking the quotient of dof_len and the length of gids_hst.
+  auto gids_dev = m_gm->get_grid(m_grid_name)->get_dofs_gids();
+  auto gids_hst = Kokkos::create_mirror_view( gids_dev );
```
bartgol, line 255: You might consider storing gids_hst in your class, so you don't do a deep copy every time you pull a variable.

And btw, it took me a minute to realize `hst` doesn't stand for "history" (which in the I/O context is also a thing). Since saving one char is not much, do you mind using `_host`  for host variables?

components/scream/src/share/io/scorpio_input.hpp
```diff
@@ -96,28 +97,52 @@ class AtmosphereInput
...
+  }
 
   // Main Functions
   void pull_input();
-  void pull_input(const std::string name, view_type view_out);
+  void pull_input(const std::string& name, view_type view_out);
+  void pull_input(const std::string& filename, const std::string& var_name, const std::vector<std::string>& var_dims,
+                  const bool has_columns, const std::vector<int>& dim_lens, const int padding, Real* data);
+
+  // Determine padding from the type of the variable.
```
bartgol, line 114: It's worth pointing out in a comment that this fcn _expects_ `var_dims` to be the "physical" dimensions of the field, as opposed to the "array" dimensions. That is, `var_dims` are the dimensions it would have if ValueType=Real (no packing).

bartgol (changes requested):
Sorry, I looked to the code in more detail, and found a few more things that might be improved.

components/scream/src/share/io/scorpio_input.hpp
AaronDonahue: done, I've added a comment.

components/scream/src/share/io/scorpio_input.cpp
AaronDonahue: Yes, good point.  I was trying to keep the names the same length for aesthetics :-).  I've changed to "host"

components/scream/src/share/io/scorpio_input.cpp
AaronDonahue: Good point.  I think the newest version has fixed this with a full revamping of the code.

AaronDonahue:
@bartgol , thank you for all the suggestions to making the code better.  I ended up doing a lot of rewriting to address the comments, I think a restructuring of scorpio_input and scorpio_output may be helpful further down the road.  But this PR at least addresses the comments you've made thus far.  When you have a chance can you re-review?

components/scream/src/share/io/scorpio_input.cpp
```diff
@@ -52,14 +66,12 @@ void AtmosphereInput::pull_input()
...
     // Get all the info for this field.
-    auto l_view_dev = fmap.get_view();
-    view_type l_view = m_view_local.at(name);
-    auto l_dims = fmap.get_header().get_identifier().get_layout().dims();
-    Int padding = fmap.get_header().get_alloc_properties().get_padding();
-    grid_read_data_array(m_filename,name,l_dims,m_dofs.at(name),padding,l_view.data());
-    Kokkos::deep_copy(l_view_dev,l_view);
+    auto l_dims = field.get_header().get_identifier().get_layout().dims();
+    Int padding = field.get_header().get_alloc_properties().get_padding();
+    grid_read_data_array(m_filename,name,l_dims,m_dofs.at(name),padding,field.get_view<Host>().data());
```
bartgol, line 73: Sorry, one more comment: if `field.get_view<Host>` returns a strided view, then passing the pointer is dangerous.

We can think about how to deal with strided views (like a single tracer) later. Meanwhile, we could put a check to make sure the field is not strided. The check would look like
```
EKAT_REQUIRE_MSG (field.get_header().get_alloc_properties().contiguous(), "Error! blah blah\n");
```

components/scream/src/share/io/scorpio_input.cpp
```diff
@@ -17,6 +17,10 @@ void AtmosphereInput::pull_input(const std::string& filename, const std::string&
   using namespace scream;
   using namespace scream::scorpio;
 
+  
+  auto gids_dev = m_gm->get_grid(m_grid_name)->get_dofs_gids();
```
bartgol, line 21: You already do this during the call to `init`, so you don't need to pull the dof gids again, no?

components/scream/src/share/io/scorpio_output.cpp
```diff
@@ -85,11 +85,10 @@ void AtmosphereOutput::init()
     rhist_in.init();
     for (auto name : m_fields)
     {
-      auto l_view = m_view_local.at(name);
-      rhist_in.pull_input(name,l_view);
+      auto l_view = rhist_in.pull_input(name);
+      Kokkos::deep_copy(m_view_local.at(name),l_view);
```
bartgol, line 89: Why the deep copy? Why not simply `m_view_local.at(name) = l_view'`?

components/scream/src/share/io/scorpio_output.cpp
```diff
@@ -304,12 +303,11 @@ void AtmosphereOutput::register_views()
...
   {
-    auto fmap = m_field_repo->get_field(name, m_grid_name);
+    auto field = m_field_repo->get_field(name, m_grid_name);
     // If the "averaging type" is instant then just need a ptr to the view.
-    auto view_d = fmap.get_view();
+    auto view_d = field.get_view();
     // Create a local copy of view to be stored by output stream.
-    auto view_copy_h = Kokkos::create_mirror_view( view_d );
-    view_type view_copy("",view_copy_h.extent(0));
+    view_type view_copy("",view_d.extent(0));
```
bartgol, line 310: I would advocate for using `field.get_view<Host>()` directly here (rather than creating view_copy), unless there is a reason to use a completely separate view.

bartgol: Also, do you _need_ the deep copy to host here? I would argue the deep copy is to be done only when the output is about to be written, not during registration...

components/scream/src/share/io/scorpio_output.cpp
```diff
@@ -304,12 +303,11 @@ void AtmosphereOutput::register_views()
   // Cycle through all fields and register.
   for (auto const& name : m_fields)
   {
-    auto fmap = m_field_repo->get_field(name, m_grid_name);
+    auto field = m_field_repo->get_field(name, m_grid_name);
     // If the "averaging type" is instant then just need a ptr to the view.
-    auto view_d = fmap.get_view();
+    auto view_d = field.get_view();
```
bartgol, line 308: Using `get_view` is dangerous if we're dealing with a subfield. We should put a check on `field`, to make sure it is not a subfield:

```
EKAT_REQUIRE_MSG (field.get_header().get_parent()==nullptr, "Error! Cannot  deal with subfield, for now.");
```

components/scream/src/share/io/scorpio_input.cpp
AaronDonahue: Oh yes, great point.  We will need to figure out how to handle this, and maybe get rid of `get_view` calls everywhere.  Probably a good git issue, so we don't lose sight of the goal.

components/scream/src/share/io/scorpio_input.cpp
AaronDonahue: This is a special case where `init` is not called because you want to do everything in one place to pull input without the field manager.  Long term, it might be good to simply get rid of any dependence on the field manager, although this would put the burden on the AD to make sure it iterates over all the important variables in the field manager for things like restarts.  Something we can discuss and then consider for a subsequent PR.

components/scream/src/share/io/scorpio_output.cpp
AaronDonahue: We need a separate view because this keeps track of "averaging", "min" and "max" type outputs too.  So we need to update a view that is independent of the field view.

components/scream/src/share/io/scorpio_output.cpp
AaronDonahue: I guess not, thanks :-)

components/scream/src/share/io/scorpio_output.cpp
AaronDonahue: per deep_copy.  That is a good point.  We could probably keep the l_view that we track on device.  I'll work on that.

components/scream/src/share/io/scorpio_output.cpp
AaronDonahue: yes, good catch.

components/scream/src/share/io/scorpio_output.cpp
AaronDonahue: I looked at the deep_copy, and I think we need to keep it.  Again, because the "view_copy" in this case refers to a separate view that will be used for output.  I suppose we could keep the view on device and just do a deep copy right before writing the output.

components/scream/src/share/io/scorpio_output.cpp
```diff
@@ -304,10 +304,13 @@ void AtmosphereOutput::register_views()
   for (auto const& name : m_fields)
   {
     auto field = m_field_repo->get_field(name, m_grid_name);
+    auto ts    = field.get_header().get_tracking().get_time_stamp();
+    EKAT_REQUIRE_MSG (ts.is_valid(), "Error in output, field " + name + " has not been set in the field manager yet");
```
bartgol, line 308: Well, the field _has_ been set in the FM. It's just that nobody has written on it yet...

components/scream/src/share/io/tests/io.cpp
```diff
@@ -262,6 +264,13 @@ std::shared_ptr<FieldRepository<Real>> get_test_repo(const Int num_lcols, const
       f4_host(ii,ipack)[ivec] = (ii) + (jj+1)/10.0;
     }
   }
+  // Update timestamp
+  util::TimeStamp time (0,0,0,0);
+  f1.get_header().get_tracking().update_time_stamp(time);
```
bartgol, line 269: If you want to init the ts of all the fields, you can simply do `repo->init_fields_time_stamp(time);`.

bartgol (approved):
I left two comments, but they are not really important. I removed the automerge label, so you can decide whether to implement them or not. If not, being approved, it can be merged away already.

AaronDonahue:
@bartgol great thanks! I made the changes you requested, might as well make the code as good as possible.  It may need a re-review for Automerge, just because I made changes.  Assuming it passes the AT I'll go ahead merge it myself without your re-review, since I just implemented the two fixes you posted.

bartgol (approved):

PR #924, opened 2021-03-08
hotfix to handle bug that with shoc_init being declared as a KOKKOS_FUNCTION

AaronDonahue:
This is a hotfix to the function shoc_init that caused a build error when using the LC machine lassen.


tcclevenger (approved):

AaronDonahue:
It looks like we failed on mappy, but its not clear it is an issue with the fix and not an issue with mappy in general.

tcclevenger:
> It looks like we failed on mappy, but its not clear it is an issue with the fix and not an issue with mappy in general.

Usually when I see error output like I just retest and it clears itself.

AaronDonahue:
ha, now its weaver and blake that fail and mappy that passes.  I'll RETEST again.

Issue #836, opened 2020-12-15
Read in input data needed for SCREAMv1 "initial" run


PeterCaldwell:
Create the scorpio code needed to read in the input datasets required for an "initial" (as opposed to a restart) run.

AaronDonahue:
I think this issue has been resolved.

PR #926, opened 2021-03-09
Initial conditions overhaul

bartgol:
This PR changes the way scream initializes its inputs, whether in standalone or CIME mode.

In particular:

- FieldInitializer classes are removed
- The AD init sequence is split into smaller functions, for ease of reading and using them.
- The AD creates an AtmosphereInput class to read in all the inputs to the atm processes.
- AtmosphereInput has been upgraded to handle also fields that are strided (it uses temporaries, not a great solution, but quick to implement). That allows to read in individual tracers from a file.

I created a `ic.nc` file on my laptop, but obviously the value `0.0` for all fields causes problems when computing exner in P3. I'm trying to set numerical values that are less nan/inf prone, but I haven't finished yet.

I also haven't tried to run the code in a CIME branch. I noticed shoc has some 2d input fields, whose name sound quite like surface-coupling related; so I suspect some atm inputs might _not_ need to be read from file in CIME runs, since they will be provided by the SC online. (@AaronDonahue does this sound correct?).

Still, I'm creating the PR in case some good soul wants to take a look and give feedback even if it's not ready.

Partly addresses #836 and #864


components/scream/src/share/io/CMakeLists.txt
```diff
@@ -1,4 +1,5 @@
 set(SCREAM_SCORPIO_SRCS
+  scorpio_input.cpp
```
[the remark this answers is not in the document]
bartgol: Good catch. I'll purge this line.

AaronDonahue:
strange that shoc_stand_alone is the only test that failed, and failed only on blake.

components/scream/src/control/atmosphere_driver.cpp
```diff
@@ -112,6 +141,129 @@ void AtmosphereDriver::initialize (const ekat::Comm& atm_comm,
...
+    m_output_manager.set_grids(m_grids_manager);
+    m_output_manager.set_repo(m_field_repo);
+  }
+  m_output_manager.init();
+
+  m_ad_status |= s_output_inited;
+}
+
+void AtmosphereDriver::
+initialize_fields (const util::TimeStamp& t0)
```
AaronDonahue, line 166: entirely unrelated, so take more as a comment that came to mind while reading this.  Is it possible to have every DAG output (.dot) file have a boiler plate comment in it that explains how to visualize it?  I know that when there are certain error messages that provide this info, but in the case explained in the comment below we won't get that error message, which then means the user has the DAG but has to be proficient in how to work with the `.dot` file to actually get any use out of it.

components/scream/src/share/io/scorpio_output.cpp
```diff
@@ -271,26 +279,30 @@ void AtmosphereOutput::register_dimensions(const std::string& name)
  *   field_repo: is a pointer to the field_repository for this simulation.
  *   name: is a string name of the variable who is to be added to the list of variables in this IO stream.
  */
+  using namespace scream;
```
AaronDonahue, line 282: I think I made this change, looking at the pattern in other places in these changes... Should these two lines be replaced with `using namespace scorpio`?

components/scream/tests/rrtmgp/input.yaml
```diff
@@ -14,6 +14,11 @@ Grids Manager:
   Type: Physics Only
   Reference Grid: Physics
   Physics Only:
-    Number of global columns:   10
-    Number of vertical levels: 128
+    Number of global columns: 384
+    Number of vertical levels: 72
+
+# The name of the file containing the initial conditions for this test.
+# TODO: Actually create an input file for rrtmgp.  Note that currently, rrtmgp is initialized using YAKL and not by reading input.
```
AaronDonahue, line 21: Just commenting that I need to fix this comment.  We now have an input file for rrtmgp, but it may still not get used because of YAKL.  Will need to check when the RAD PR goes in.

AaronDonahue (commented):
@bartgol , I had a few questions (mainly for myself), otherwise I think this is ready to merge.  You may also want to review the few places I made changes to the code.  Probably easiest just to check out my individual commits than go through all 80 files.

AaronDonahue:
Final comment, @bartgol , are we expecting to change the ekat and scorpio submodules with this PR?

AaronDonahue (approved):

Issue #818, opened 2020-12-11
[AD] Double check that processes don't use different names for the same variable


PeterCaldwell:
The AD allows processes to pass variables between themselves. A potential problem is that one process calls a variable one thing and another process calls the same variable something else. Thus each process wouldn't feel the effect of the other and we'd have multiple variables tracking the same thing. This issue will be complete once we have convinced ourselves that this isn't happening.

PeterCaldwell:
I think it may make most sense to wait on this task until we have all the interfaces between procs and the AD done. Otherwise we will probably have to redo this task. Do you agree, @bartgol ?

bartgol:
I think it's ok to wait until interfaces are ready. But I also think it's ok to start doing it now for what we have. E.g., Homme and P3 are relatively ripe; it might be that, once a couple of atm processes are "in sync", it becomes easier for the others to follow suite. I'm not sure.

Anyhow, the issue about field names hits pretty quickly as soon as one tries to write an atm test (like those in scream/tests) with more than one atm proc, with names that are not in sync. So, as soon as someone is finished with their atm process interface, and start writing a unit test that couples it to other atm procs, they will face this issue, and hopefully fix it.

PeterCaldwell:
Roger that. Doing an initial sweep and using the experience to avoid problems in our next sweep is a good idea. Don't we have a tool for checking whether names are similar? Maybe AI could solve this problem for us? Mostly kidding, I suspect it will take 2x more work to get an AI method which does a decent job of finding similar variables.

bartgol:
Yeah, no such tool, unfortunately. At least not that I can think of.

PeterCaldwell:
Plan is to pause on this task until SCREAMv1 is basically done, then to sweep all the interfaces with our eyeballs to confirm we don't have any issues. In general, it seems impossible to write a test to make sure folks use consistent variable names everywhere.

PeterCaldwell:
completed via https://acme-climate.atlassian.net/wiki/spaces/NGDNA/pages/1927512393/Field+Manager+Variable+Names

PR #931, opened 2021-03-10
Upgrades/fixes to query-cf-database

bartgol:
Miscellanea upgrades and fixes to the script. Most importantly, I added the possibility of not tokenizing an input string, so that searching, say, for `air_pressure` would not give a higher score to `sound_pressure_level_in_air` than to `air_pressure_at_mean_sea_level`.

Note: I ended up doing this while trying to uniform variable names in scream (as part of #818). In particular, while querying the cf database, I was not picking up names that I was seeing when I opened the file.


jeff-cohere (approved):
You've done it up real nice, Luca!

PR #932, opened 2021-03-10
Add README for scripts

jgfouca:
Also:
1) Improve --help dumps for key tools
2) Make cf-xml-to-yaml consistent with other tools
3) All tools should be using python3
4) perf-analysis: updates needed to support SCREAM build system changes, make sure mappy is using 46 threads
5) Instrument shoc_run_and_cmp to support timings (same as was done for P3). Since shoc_main does not do
a lot of initialization, it basically jumps right in to the interesting part, I don't think we need to limit
timings within shoc_main, we can just time the entire call.

### DIFF
--- a/components/scream/src/physics/shoc/tests/shoc_run_and_cmp.cpp
+++ b/components/scream/src/physics/shoc/tests/shoc_run_and_cmp.cpp
@@ -93,26 +93,59 @@ static Int compare (const std::string& label, const Scalar* a,
 
 struct Baseline {
 
-  // Number of iterations (nadv steps of size dtime per iteration).
-  const int num_iters = 10;
-
-  Baseline () {
-    //                 ic, shcol, nlev, num_qtracers, nadv, dtime
-    params_.push_back({ic::Factory::standard, 8, 72, 3, 15, 150});
+  Baseline (const Int num_iters, const Int shcol, const Int repeat) :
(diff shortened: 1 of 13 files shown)


components/scream/src/physics/shoc/tests/shoc_run_and_cmp.cpp
```diff
@@ -93,26 +93,59 @@ static Int compare (const std::string& label, const Scalar* a,
 
 struct Baseline {
 
-  // Number of iterations (nadv steps of size dtime per iteration).
-  const int num_iters = 10;
-
-  Baseline () {
-    //                 ic, shcol, nlev, num_qtracers, nadv, dtime
-    params_.push_back({ic::Factory::standard, 8, 72, 3, 15, 150});
+  Baseline (const Int num_iters, const Int shcol, const Int repeat) :
```
bartgol, line 96: Was it in your intent to modify also src code in scream? Or did this spill from some unstaged work in your tree?

jgfouca: @bartgol , this change was related to item 5 in the PR description. As I started to dive into improving the documentation of perf-analysis so that @tcclevenger can use it for measuring SHOC, I started trying to remember how I got it to work for measuring P3, starting poking around, and soon had shoc_run_and_cmp working.

bartgol (approved):
Very nice documentation!

PR #934, opened 2021-03-16
parallelize remap

tcclevenger:
Changes `PhysicsDynamics` remappers to parallelize over fields. No new test is written, relying on the one that already exists (shouldn't be an issue).
Main changes:
1. In both `do_remap_fwd/bwd`, the `for(num_fields)` was replaced with `parallel_for(TeamPolicy, num_fields)`.
2. All info needed on device is computed in a new function `initialize_device_variables()` which is called at the end of `do_registration_ends()`.
3. Added the pack logic to 2d remmaping methods (was just 3d). 
4. New function `set_dyn_to_zero()` sets the relevant portion of the dynamics view to 0. Called right before `local_remap_fwd_2d/3d()`. This function may be unnecessary if there is a `set_zero` like method on device, but I could not find it..

### DIFF
--- a/components/scream/src/dynamics/homme/physics_dynamics_remapper.hpp
+++ b/components/scream/src/dynamics/homme/physics_dynamics_remapper.hpp
@@ -38,6 +38,11 @@ public:
   using layout_type     = typename base_type::layout_type;
   using grid_ptr_type   = typename base_type::grid_ptr_type;
 
+  using device_type     = typename field_type::device_type;
+
+  template<typename T, int N>
+  using uview_nd = Unmanaged<typename KokkosTypes<device_type>::template view_ND<T,N>>;
+
   using pack_type = ekat::Pack<RealType,SCREAM_PACK_SIZE>;
   using small_pack_type = ekat::Pack<RealType,SCREAM_SMALL_PACK_SIZE>;
 
@@ -389,6 +394,11 @@ initialize_device_variables()
       EKAT_REQUIRE_MSG(dh.get_parent().lock() != nullptr,
                        "Error! If physics field has parent,"
                        "dynamics field must also have parent.");
+
+      const int ifield = this->find_field(ph.get_parent().lock()->get_identifier(), dh.get_parent().lock()->get_identifier());
+      EKAT_REQUIRE_MSG(ifield != -1,
+                       "Error! Parent must be registered for remapped subfields.");
+
       h_has_parent(i) = true;
       continue;
     }
@@ -463,11 +473,11 @@ set_dyn_to_zero(const MT& team) const
     case etoi(LayoutType::Scalar2D):
     {
       if (is_state_field_dev(i)) {
-        auto dyn = Unmanaged<KokkosTypes<DefaultDevice>::view<ScalarT****>> (reinterpret_cast<ScalarT*>(dyn_ptrs(i).get()),
-                                                                             dim_d[0],
-                                                                             dim_d[1],
-                                                                             dim_d[2],
-                                                                             dim_d[3]);
+        auto dyn = uview_nd<ScalarT,4> (reinterpret_cast<ScalarT*>(dyn_ptrs(i).get()),
+                                                                   dim_d[0],
+                                                                   dim_d[1],
+                                                                   dim_d[2],
+                                                                   dim_d[3]);
         auto v = ekat::subview_1(dyn,itl);
 
         Kokkos::parallel_for(Kokkos::TeamThreadRange(team, v.size()), [&](const int& k) {
@@ -478,8 +488,10 @@ set_dyn_to_zero(const MT& team) const
           v(k0, k1, k2) = 0;
         });
       } else {
-        auto dyn = Unmanaged<KokkosTypes<DefaultDevice>::view<ScalarT*[NP][NP]>> (reinterpret_cast<ScalarT*>(dyn_ptrs(i).get()),
-                                                                                  dim_d[0]);
+        auto dyn = uview_nd<ScalarT,3> (reinterpret_cast<ScalarT*>(dyn_ptrs(i).get()),
+                                        dim_d[0],
+                                        dim_d[1],
+                                        dim_d[2]);
 
         Kokkos::parallel_for(Kokkos::TeamThreadRange(team, dyn.size()), [&](const int& k) {
           int k0   = k%dim_d[0];
@@ -494,12 +506,12 @@ set_dyn_to_zero(const MT& team) const
     case etoi(LayoutType::Vector2D):
     {
       if (is_state_field_dev(i)) {
-        auto dyn = Unmanaged<KokkosTypes<DefaultDevice>::view<ScalarT*****>> (reinterpret_cast<ScalarT*>(dyn_ptrs(i).get()),
-                                                                              dim_d[0],
-                                                                              dim_d[1],
-                                                                              dim_d[2],
-                                                                              dim_d[3],
-                                                                              dim_d[4]);
+        auto dyn = uview_nd<ScalarT,5> (reinterpret_cast<ScalarT*>(dyn_ptrs(i).get()),
+                                        dim_d[0],
+                                        dim_d[1],
+                                        dim_d[2],
+                                        dim_d[3],
+                                        dim_d[4]);
         auto v = ekat::subview_1(dyn,itl);
 
         Kokkos::parallel_for(Kokkos::TeamThreadRange(team, v.size()), [&](const int& k) {
@@ -512,9 +524,11 @@ set_dyn_to_zero(const MT& team) const
           v(k0, k1, k2, k3) = 0;
         });
       } else {
-        auto dyn = Unmanaged<KokkosTypes<DefaultDevice>::view<ScalarT**[NP][NP]>> (reinterpret_cast<ScalarT*>(dyn_ptrs(i).get()),
-                                                                                   dim_d[0],
-                                                                                   dim_d[1]);
+        auto dyn = uview_nd<ScalarT,4> (reinterpret_cast<ScalarT*>(dyn_ptrs(i).get()),
+                                        dim_d[0],
+                                        dim_d[1],
+                                        dim_d[2],
+                                        dim_d[3]);
 
         Kokkos::parallel_for(Kokkos::TeamThreadRange(team, dyn.size()), [&](const int& k) {
           int k0   = k%dim_d[0];
@@ -531,12 +545,12 @@ set_dyn_to_zero(const MT& team) const
     case etoi(LayoutType::Scalar3D):
     {
       if (is_state_field_dev(i)) {
-        auto dyn = Unmanaged<KokkosTypes<DefaultDevice>::view<ScalarT*****>> (reinterpret_cast<ScalarT*>(dyn_ptrs(i).get()),
-                                                                              dim_d[0],
-                                                                              dim_d[1],
-                                                                              dim_d[2],
-                                                                              dim_d[3],
-                                                                              dim_d[4]);
+        auto dyn = uview_nd<ScalarT,5> (reinterpret_cast<ScalarT*>(dyn_ptrs(i).get()),
+                                        dim_d[0],
+                                        dim_d[1],
+                                        dim_d[2],
+                                        dim_d[3],
+                                        dim_d[4]);
         auto v = ekat::subview_1(dyn,itl);
 
         Kokkos::parallel_for(Kokkos::TeamThreadRange(team, v.size()), [&](const int& k) {
@@ -550,11 +564,11 @@ set_dyn_to_zero(const MT& team) const
         });
 
       } else {
-        auto dyn = Unmanaged<KokkosTypes<DefaultDevice>::view<ScalarT****>> (reinterpret_cast<ScalarT*>(dyn_ptrs(i).get()),
-                                                                             dim_d[0],
-                                                                             dim_d[1],
-                                                                             dim_d[2],
-                                                                             dim_d[3]);
+        auto dyn = uview_nd<ScalarT,4> (reinterpret_cast<ScalarT*>(dyn_ptrs(i).get()),
+                                        dim_d[0],
+                                        dim_d[1],
+                                        dim_d[2],
+                                        dim_d[3]);
 
         Kokkos::parallel_for(Kokkos::TeamThreadRange(team, dyn.size()), [&](const int& k) {
           int k0   = k%dim_d[0];
@@ -571,13 +585,13 @@ set_dyn_to_zero(const MT& team) const
     case etoi(LayoutType::Vector3D):
     {
       if (is_state_field_dev(i)) {
-        auto dyn = Unmanaged<KokkosTypes<DefaultDevice>::view<ScalarT******>> (reinterpret_cast<ScalarT*>(dyn_ptrs(i).get()),
-                                                                               dim_d[0],
-                                                                               dim_d[1],
-                                                                               dim_d[2],
-                                                                               dim_d[3],
-                                                                               dim_d[4],
-                                                                               dim_d[5]);
+        auto dyn = uview_nd<ScalarT,6> (reinterpret_cast<ScalarT*>(dyn_ptrs(i).get()),
+                                        dim_d[0],
+                                        dim_d[1],
+                                        dim_d[2],
+                                        dim_d[3],
+                                        dim_d[4],
+                                        dim_d[5]);
         auto v = ekat::subview_1(dyn,itl);
 
         Kokkos::parallel_for(Kokkos::TeamThreadRange(team, v.size()), [&](const int& k) {
@@ -592,12 +606,12 @@ set_dyn_to_zero(const MT& team) const
           v(k0, k1, k2, k3, k4) = 0;
         });
       } else {
-        auto dyn = Unmanaged<KokkosTypes<DefaultDevice>::view<ScalarT*****>> (reinterpret_cast<ScalarT*>(dyn_ptrs(i).get()),
-                                                                              dim_d[0],
-                                                                              dim_d[1],
-                                                                              dim_d[2],
-                                                                              dim_d[3],
-                                                                              dim_d[4]);
+        auto dyn = uview_nd<ScalarT,5> (reinterpret_cast<ScalarT*>(dyn_ptrs(i).get()),
+                                        dim_d[0],
+                                        dim_d[1],
+                                        dim_d[2],
+                                        dim_d[3],
+                                        dim_d[4]);
 
         Kokkos::parallel_for(Kokkos::TeamThreadRange(team, dyn.size()), [&](const int& k) {
           int k0   = k%dim_d[0];
@@ -957,14 +971,16 @@ local_remap_fwd_2d (const MT& team, const int num_cols, const VT1 lid2elgp, cons
   switch (phys_layout(i)) {
     case etoi(LayoutType::Scalar2D):
     {
-      auto phys = Unmanaged<KokkosTypes<DefaultDevice>::view<ScalarT*>> (reinterpret_cast<ScalarT*>(phys_ptrs(i).get()),
-                                                                         dim_p[0]);
+      auto phys = uview_nd<ScalarT,1> (reinterpret_cast<ScalarT*>(phys_ptrs(i).get()),
+                                       dim_p[0]);
 
       const auto tr = Kokkos::TeamThreadRange(team, num_cols);
       if (is_state_field_dev(i)) {
-        auto dyn = Unmanaged<KokkosTypes<DefaultDevice>::view<ScalarT**[NP][NP]>> (reinterpret_cast<ScalarT*>(dyn_ptrs(i).get()),
-                                                                                   dim_d[0],
-                                                                                   dim_d[1]);
+        auto dyn = uview_nd<ScalarT,4> (reinterpret_cast<ScalarT*>(dyn_ptrs(i).get()),
+                                        dim_d[0],
+                                        dim_d[1],
+                                        dim_d[2],
+                                        dim_d[3]);
 
         const auto f = [&] (const int icol) {
           const auto& elgp = Kokkos::subview(lid2elgp,p2d(icol),Kokkos::ALL());
@@ -972,8 +988,10 @@ local_remap_fwd_2d (const MT& team, const int num_cols, const VT1 lid2elgp, cons
         };
         Kokkos::parallel_for(tr, f);
       } else {
-        auto dyn = Unmanaged<KokkosTypes<DefaultDevice>::view<ScalarT*[NP][NP]>> (reinterpret_cast<ScalarT*>(dyn_ptrs(i).get()),
-                                                                                  dim_d[0]);
+        auto dyn = uview_nd<ScalarT,3> (reinterpret_cast<ScalarT*>(dyn_ptrs(i).get()),
+                                         dim_d[0],
+                                         dim_d[1],
+                                         dim_d[2]);
 
         const auto tr = Kokkos::TeamThreadRange(team, num_cols);
         const auto f = [&] (const int icol) {
@@ -986,16 +1004,18 @@ local_remap_fwd_2d (const MT& team, const int num_cols, const VT1 lid2elgp, cons
     }
     case etoi(LayoutType::Vector2D):
     {
-      auto phys = Unmanaged<KokkosTypes<DefaultDevice>::view<ScalarT**>> (reinterpret_cast<ScalarT*>(phys_ptrs(i).get()),
-                                                                          dim_p[0],
-                                                                          dim_p[1]);
+      auto phys = uview_nd<ScalarT,2> (reinterpret_cast<ScalarT*>(phys_ptrs(i).get()),
+                                       dim_p[0],
+                                       dim_p[1]);
 
       const auto tr = Kokkos::TeamThreadRange(team, num_cols*dim_p[1]);
       if (is_state_field_dev(i)) {
-        auto dyn = Unmanaged<KokkosTypes<DefaultDevice>::view<ScalarT***[NP][NP]>> (reinterpret_cast<ScalarT*>(dyn_ptrs(i).get()),
-                                                                                    dim_d[0],
-                                                                                    dim_d[1],
-                                                                                    dim_d[2]);
+        auto dyn = uview_nd<ScalarT,5> (reinterpret_cast<ScalarT*>(dyn_ptrs(i).get()),
+                                        dim_d[0],
+                                        dim_d[1],
+                                        dim_d[2],
+                                        dim_d[3],
+                                        dim_d[4]);
 
         const auto f = [&] (const int idx) {
           const int icol = idx/dim_p[1];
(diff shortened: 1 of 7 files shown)


components/scream/src/dynamics/homme/physics_dynamics_remapper.hpp
```diff
@@ -272,133 +314,496 @@ do_registration_ends ()
...
+          team.team_barrier();
 
-    const bool phys_pack_alloc = phys_alloc_prop.template is_compatible<pack_type>();
-    const bool dyn_pack_alloc  = dyn_alloc_prop.template  is_compatible<pack_type>();
-    const bool phys_small_pack_alloc = phys_alloc_prop.template is_compatible<small_pack_type>();
-    const bool dyn_small_pack_alloc  = dyn_alloc_prop.template  is_compatible<small_pack_type>();
+          local_remap_fwd_2d<pack_type>(team, num_cols, lid2elgp, p2d);
+        } else if (pack_alloc_property(i) == 1) {
+          set_dyn_to_zero<small_pack_type>(team);
+          team.team_barrier();
```
bartgol, line 715: Not that it matter much (I think), but the barrier here is redundant. There's no other code executed during this iteration, and there is already a guaranteed team barrier at the end of the iteration.

components/scream/src/dynamics/homme/physics_dynamics_remapper.hpp
```diff
@@ -272,133 +314,496 @@ do_registration_ends ()
...
-          remap_bwd_3d_impl<Real>(phys,dyn,itl);
+          set_dyn_to_zero<Real>(team);
+          team.team_barrier();
+
+          local_remap_fwd_3d<Real>(team, num_cols, lid2elgp, p2d);
         }
         break;
+      }
       default:
         ekat::error::runtime_abort("Error! Unhandled case in switch statement.\n");
```
bartgol, line 749: I think `runtime_abort` is not device friendly. You might need a `EKAT_KERNEL_ERROR_MSG` here.

components/scream/src/dynamics/homme/physics_dynamics_remapper.hpp
```diff
@@ -272,133 +314,496 @@ do_registration_ends ()
...
     }
-  }
+  };
+  const auto policy = ekat::ExeSpaceUtils<KT::ExeSpace>::get_default_team_policy(num_fields, 1);
```
bartgol, line 752: So, this is something worth checking going fwd, but I _think_ it might be best to ||ize over `num_fields*num_cols`. The fields might be unbalanced: if qsize>>1, remapping Q is much more work than remapping, say, T.

For 2d quantities, it leaves just 1 iteration in the local remap loop (which is a waste of threads), but 3d quantities can fully occupy 4 warps (at 128 levels). OTOH, I think surface pressure might be the only 2d field to remap (and I'm not 100% sure we need to remap it either), so it's ok to have some idle threads on 1 remapped fields, if we gain something on the 3d fields...

This can be adjusted in a second PR though.

bartgol: One side benefit of pulling the `icol` loop at the top level is that you only need switches to extract subviews, while the inner loop (for 3d fields, at the k level, that is) is the same, since the TTR would always happen on 1d views (you can subview elgp[0], time level (if any), elgp[1], and elgp[2]). That might skim a few lines of repeated code.

components/scream/src/dynamics/homme/physics_dynamics_remapper.hpp
```diff
@@ -589,396 +994,493 @@ setup_boundary_exchange () {
...
+      const auto tr = Kokkos::TeamThreadRange(team, num_cols*dim1*dim2);
+      const auto f = [&] (const int idx) {
+        const int icol = idx/(dim1*dim2);
+        const int dims[2] = { (idx/dim2)%dim1 , idx%dim2 };
 
         const auto& elgp = Kokkos::subview(lid2elgp,p2d(icol),Kokkos::ALL());
-        dyn(elgp[0],dims[ordering.first],dims[ordering.second],elgp[1],elgp[2]) = phys(icol,dims[0],dims[1]);
-      });
+
+        dyn(elgp[0],dims[ordering(i,0)],dims[ordering(i,1)],elgp[1],elgp[2]) = phys(icol,dims[0],dims[1]);
```
bartgol, line 1094: I can't think of any Tensor quantity to remap, but I'm ok keeping the capability. However, it might trim down this class if we _enforced_ the two dimensions to appear in the _same_ order. What do you think? The hardcore solution would be to completely remove support for Tensor2D and Tensor3D. Which I _think_ PD remapping doesn't really _need_.

components/scream/src/dynamics/homme/physics_dynamics_remapper.hpp
```diff
@@ -589,396 +994,493 @@ setup_boundary_exchange () {
...
+      };
+      Kokkos::parallel_for(tr, f);
       break;
     }
     default:
-      ekat::error::runtime_abort("Error! Invalid layout. This is an internal error. Please, contact developers\n");
+      ekat::error::runtime_abort("Error! Unhandled case in switch statement.\n");
```
bartgol, line 1100: Like before: this fcn is not device friendly.

components/scream/src/dynamics/homme/physics_dynamics_remapper.hpp
```diff
@@ -102,13 +101,56 @@ class PhysicsDynamicsRemapper : public AbstractRemapper<RealType>
...
-  void remap_bwd_3d_impl (const field_type& phys, const field_type& dyn, const int itl) const;
+  struct Pointer {
+    KOKKOS_FORCEINLINE_FUNCTION
+    Real* get() { return ptr; }
+
+    Real* ptr;
+  };
+
+  KokkosTypes<DefaultDevice>::view_1d<Pointer> phys_ptrs;
+  KokkosTypes<DefaultDevice>::view_2d<Int>     phys_dims;
```
bartgol, line 113: I am thinking that, since the 2nd dim is known at compile-time, we might be better off with

```
view_1d<Kokkos::Array<Int,4>> phys_dims;
```
This might save indexing time (is this true?) and also register count when using it (see below comment marked "ABOUT PHYS_DIMS").

components/scream/src/dynamics/homme/physics_dynamics_remapper.hpp
```diff
@@ -589,396 +994,493 @@ setup_boundary_exchange () {
...
-        const int dim = phys_dims[1];
-        Kokkos::parallel_for(RangePolicy(0,num_cols*dim),
-                             KOKKOS_LAMBDA(const int idx) {
-          const int icol = idx / dim;
-          const int idim = idx % dim;
+      auto phys = Unmanaged<KokkosTypes<DefaultDevice>::view<ScalarT**>> (reinterpret_cast<ScalarT*>(phys_ptrs(i).get()),
+                                                                          phys_dims(i,0),
+                                                                          phys_dims(i,1));
+
+      const int dim = phys_dims(i, 1);
```
bartgol, line 1040: ABOUT PHYS_DIMS:

If phys dims were a 1d view of Kokkos::Array, we could do
```
// Outside the switch:
const auto& dims = phys_dims(i);
...
// inside the loops
const icol = idx / dims[0];
...
const icol = idx / (dims[0]*dims[1]);
```
without having dim1,dim2,.. I _think_ this _might_ reduce register pressure (and maybe allow an extra block to fit on the SM).

components/scream/src/dynamics/homme/physics_dynamics_remapper.hpp
```diff
@@ -272,133 +314,496 @@ do_registration_ends ()
...
+  phys_layout    = decltype(phys_layout)    ("phys_layout",    num_fields);
+
+  dyn_ptrs   = decltype(dyn_ptrs)   ("dyn_ptrs",   num_fields);
+  dyn_dims   = decltype(dyn_dims)   ("dyn_dims",   num_fields, 6);
+  dyn_layout = decltype(dyn_layout) ("dyn_layout", num_fields);
+
+  pack_alloc_property = decltype(pack_alloc_property) ("phys_pack_alloc_property", num_fields);
+  itl_fwd             = decltype(itl_fwd)             ("itl_fwd",                  num_fields);
+  itl_bwd             = decltype(itl_bwd)             ("itl_bwd",                  num_fields);
+  ordering            = decltype(ordering)            ("ordering_new",             num_fields, 2);
```
bartgol, line 340: Somewhat similar to phys_dims: would it be beneficial to have ordering be a `view_1d<Kokkos::pair<int,int>>`? I suspect the answer is "yes". Perhaps `view<Int*[2]>` would have similar benefits, though I still suspect the pair approach would yield better register usage.

components/scream/src/dynamics/homme/physics_dynamics_remapper.hpp
bartgol: An alternative to `Kokkos::Array` would be a POD type:
```
struct PhysDims {
  int dim0;
  int dim1;
  int dim2;
  int dim3;
};
```
And you could even throw in `int size;` as a member, so to get rid of `phys_dims_size`. I might actually prefer this to `Kokkos::Array`.

bartgol (commented):
Nice work @tcclevenger ! Even if not a reviewer, I took the liberty to go through it. There's a couple of things that need to be fixed (namely the non-device-friendly calls), and others that can be improved (though don't really _have to_).

Let me know what your thoughts are. Some proposed mods are for perf (and can or even should be done in a different PR), others for style, others for readability.

components/scream/src/dynamics/homme/physics_dynamics_remapper.hpp
```diff
@@ -272,133 +314,496 @@ do_registration_ends ()
...
+    h_itl_fwd(i) = m_is_state_field[i] ? tl.n0 : -1;
+    h_itl_bwd(i) = m_is_state_field[i] ? tl.np1 : -1;
+
+    if (ph.get_identifier().get_layout().tags().size() > 1) {
+      EKAT_REQUIRE_MSG(dh.get_identifier().get_layout().tags().size() > 1,
+                       "Error! Dyn layout must also have at least 2 tags.");
+
+      h_ordering(i,0) =  (ph.get_identifier().get_layout().tag(1) ==
+                          dh.get_identifier().get_layout().tag(1) ? 0 : 1);
+      h_ordering(i,1) =  (ph.get_identifier().get_layout().tag(1) ==
```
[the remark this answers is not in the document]
tcclevenger: I think this is correct. The RHS are slightly different, one is `condition ? 0 : 1;` and the other is `condition ? 1 : 0;`. The original is 
```
Kokkos::pair<int,int> ordering;
if (phys_layout.tag(1)==dyn_layout.tag(1)) {
  ordering.first=0;
  ordering.second=1;
} else {
  ordering.first=1;
  ordering.second=0;
}
```
So if the tags are the same, then `(i,0)=0` and `(i,1)=1`.

components/scream/src/dynamics/homme/physics_dynamics_remapper.hpp
```diff
@@ -272,133 +314,496 @@ do_registration_ends ()
...
+      h_dyn_dims(i, dyn_dim.size()-1)   = ekat::npack<small_pack_type>(dyn_dim[dyn_dim.size()-1]);
+    } else {
+      h_pack_alloc_property(i) = 2;
+    }
+
+    h_itl_fwd(i) = m_is_state_field[i] ? tl.n0 : -1;
+    h_itl_bwd(i) = m_is_state_field[i] ? tl.np1 : -1;
+
+    if (ph.get_identifier().get_layout().tags().size() > 1) {
+      EKAT_REQUIRE_MSG(dh.get_identifier().get_layout().tags().size() > 1,
```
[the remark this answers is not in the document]
tcclevenger: Good point, I'll remove.

components/scream/src/dynamics/homme/physics_dynamics_remapper.hpp
```diff
@@ -272,133 +314,496 @@ do_registration_ends ()
...
-      Kokkos::deep_copy(dyn.get_view(),0.0);
+    for (int n=0; n<(int)dyn_dim.size(); ++n) {
+      h_dyn_dims(i, n) = dyn_dim[n];
     }
 
+    // Layout
+    const auto phys_lt = get_layout_type(ph.get_identifier().get_layout().tags());
     switch (phys_lt) {
       case LayoutType::Scalar2D:
+        h_phys_layout(i) = 0;
```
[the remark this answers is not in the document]
tcclevenger: Ahh I was looking for enum usages elsewhere, but I neglected the homme folder outside of scream.. This will make the code much nicer.

components/scream/src/dynamics/homme/physics_dynamics_remapper.hpp
```diff
@@ -272,133 +314,496 @@ do_registration_ends ()
...
+      h_pack_alloc_property(i) = 1;
+
+      // Adjust last dim to be number of packs
+      h_phys_dims(i, phys_dim.size()-1) = ekat::npack<small_pack_type>(phys_dim[phys_dim.size()-1]);
+      h_dyn_dims(i, dyn_dim.size()-1)   = ekat::npack<small_pack_type>(dyn_dim[dyn_dim.size()-1]);
+    } else {
+      h_pack_alloc_property(i) = 2;
+    }
+
+    h_itl_fwd(i) = m_is_state_field[i] ? tl.n0 : -1;
```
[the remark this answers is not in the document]
tcclevenger: I'll make a device friendly `m_is_state_field` and query that.

components/scream/src/dynamics/homme/physics_dynamics_remapper.hpp
```diff
@@ -272,133 +314,496 @@ do_registration_ends ()
...
+    {
+      if (itl>=0) {
+        auto dyn = Unmanaged<KokkosTypes<DefaultDevice>::view<ScalarT****>> (reinterpret_cast<ScalarT*>(dyn_ptrs(i).get()),
+                                                                             dyn_dims(i,0),
+                                                                             dyn_dims(i,1),
+                                                                             dyn_dims(i,2),
+                                                                             dyn_dims(i,3));
+        auto v = ekat::subview_1(dyn,itl);
+
+        const size_t N = v.extent(0)*v.extent(1)*v.extent(2);
```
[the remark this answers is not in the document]
tcclevenger: Yep, I switch all in this function to just ask for size..

components/scream/src/dynamics/homme/physics_dynamics_remapper.hpp
tcclevenger: Testing with OpenMP, it seems they are necessary.  The call `set_dyn_to_zero();` will iterate through the `dyn` entries differently than the call `local_remap_fwd_2d()` directly below, so I believe a `dyn` entry could be set correctly, and then set to 0. 

An option could be to iterate in `set_dyn_to_zero()` over all elements not touched in `local_remap_fwd_2d()`, but tat may be overly complicated for its benefit.

bartgol: Ah, shoes, I missed the call to `local_remap_fwd_2d`. Good point! Even if they iterated in the same way, it's dangerous to assume the same thread will access the same point in both calls.

components/scream/src/dynamics/homme/physics_dynamics_remapper.hpp
tcclevenger: Either way. If it's not needed then I say just remove.. But I'm not sure what all fields would need to be remapped in the future? 

If we keep it should we just `assert(phys_layout.tag(1) == dyn_layout.tag(1))`?

components/scream/src/dynamics/homme/physics_dynamics_remapper.hpp
tcclevenger: I like the struct option best, I think. Let me see how clean it is.

components/scream/src/dynamics/homme/physics_dynamics_remapper.hpp
bartgol: I think we can always add this code back if need be. We can always go back on the PR and see the code that got deleted. My guess, however, is that we won't need this, ever. So yeah, I am also in favor of purging it, making the class maintenance easier.

components/scream/src/dynamics/homme/physics_dynamics_remapper.hpp
bartgol: It's definitely more verbose.

tcclevenger: This issue with the struct is that at one place I need to access the last dimension, which would require a large if-block twice for both phys and dyn. What about a combo of the 2?
```
  struct Dims {
    int size;
    Kokkos::Array<int,6> dims;
  };
```

bartgol: Ah, cause you the idx to access is taken from a variable. I see. Then I guess what you suggest is a good solution.

components/scream/src/dynamics/homme/physics_dynamics_remapper.hpp
tcclevenger: This no longer exists (removed tensor3d).

components/scream/src/dynamics/homme/physics_dynamics_remapper.hpp
tcclevenger: Yeah, that makes sense if some field are much larger. I'll work on the implementation after this PR.

tcclevenger:
@bartgol I've compiled some basic timings:

Test: In the `homme_pd_remap` test, for both `remap` and `combo remap` test cases, I put a for loop around the `remap()` call (or the sequence of remap calls in combo) which runs 1000 times. Example:
```
auto start = std::chrono::steady_clock::now();
for (int i=0; i<1000; ++i) {
  // call(s) to remapper->remap()
}
auto finish = std::chrono::steady_clock::now();
auto duration = std::chrono::duration_cast<std::chrono::microseconds>(finish - start);
```
SCREAM was configured in Release with EKAT_BFB=OFF. For Blake and my workstaion, OPENMP=ON.

The results are suprising. Remap is now ~2x faster on my workstation, but 4-5x slower on Weaver, and varies widely on Blake..
Screenshot from 2021-03-24 15-24-23

Not sure if I am doing obviously wrong on the test machines..

bartgol:
@tcclevenger thanks for taking the time to take some timings. Very interesting results indeed.

For GPU, I think your team policy is the problem: looking [here](https://github.com/E3SM-Project/scream/blob/c4a37de3f7665e60989b8a770b2fc614df60c2e6/components/scream/src/dynamics/homme/physics_dynamics_remapper.hpp#L740), it seems like you are setting the team size to 1, which means you don't parallelize over columns/levels at all.

For the CPU, I'm puzzled, cause you're workstation showed a nice speedup. Was your workstation Intel too? And did you use same pack size on blake and your workstation? It might be that team_size=1 penalizes you on CPU too, but I'm not sure how much. It should, since we have a handful of fields, so with team_size=1 you end up using only a handful of threads. But then why the speedup on the workstation?

Anyways, I would try a bigger team size, and see what happens. And careful with `get_default_team_policy`, since its team size logic was written with the assumption that a team is handling a column, while here a team is handling a field (with multiple columns). We might play a bit with team size and vector length, but I think we should focus on getting a good choice for the remap of 3d fields. After all, we'll likely only use pd remap for `v, dp, theta, phi, Q` (or some quantities with the same layout). We might have to remap out also `ps` (a 2d field), or maybe replace `dp` (defined at levels midpoints) with `p_i` (at levels interfaces, with `p_i(top)==ps`). But the bulk of the work is the midpoints quantities (v,theta,Q, and possibly dp).

tcclevenger:
> For the CPU, I'm puzzled, cause you're workstation showed a nice speedup. Was your workstation Intel too? And did you use same pack size on blake and your workstation? It might be that team_size=1 penalizes you on CPU too, but I'm not sure how much. It should, since we have a handful of fields, so with team_size=1 you end up using only a handful of threads. But then why the speedup on the workstation?

I believe workstation is gcc (`mpicc --version` is gcc). Pack size is the same (default), and the team size is 7 for both my workstation and blake. 

For now though I'll focus on team size and start with the GPU to see if I can get speedup.

bartgol:
> For now though I'll focus on team size and start with the GPU to see if I can get speedup.

I think you should see *some* speedup. The previous impl used a simple RangePolicy, which means it could use as many threads as Cuda would allow, for a single field.

tcclevenger:
I actually when to Blake first.. Here are some interesting results.

New test: Inside do_remap_fwd, I timed 1000 outer teampolicy loops (over fields) with team_size=1,...,max_team_size (64 in Kokkos). There are 7 fields to remap in homme_pd_remap fwd test (5 3d fields, 2 2d fields):
```
const int max_threads =  KT::ExeSpace::concurrency();
for (int team_size=1; team_size<=max_threads; ++team_size) {
  const auto policy = ekat::ExeSpaceUtils<KT::ExeSpace>::get_team_policy_force_team_size(num_fields, team_size);

   // Start timer
   auto start = std::chrono::steady_clock::now();
   for (int t=0; t<1000; ++t) {
     // Actual remap
     Kokkos::parallel_for(policy, field_loop);
     Kokkos::fence();
    }
    auto finish = std::chrono::steady_clock::now();
    auto duration = std::chrono::duration_cast<std::chrono::microseconds>(finish - start);
    std::cout << "TeamSize: " << team_size << ": "
              << duration.count()*1e-6 << std::endl;
  }
```
Here are the minimum times, corresponding teamsize, and factor for 4 separate runs of the test (as well as the minimum time of 10 runs with the old method):
Time for old routine (min of 10 runs): 0.6505
- Run 1:
min time:  0.0848, Corresponding team_size: 19, Factor 0.130
- Run 2:
min time:  0.0877, Corresponding team_size: 20, Factor 0.135
- Run 3:
min time:  0.1133, Corresponding team_size: 24, Factor 0.174
- Run 4:
min time:  0.0895, Corresponding team_size: 21, Factor 0.138

These factors (I believe) makes sense for 7 fields (1/7=0.142..), but I'm not quite sure why a team size of 20 is the consensus team size, but I'm currently trying to get a better understanding of team size...

bartgol:
Well, it kind of makes sense. The test uses ne=2, which means 24 elements on the cubed sphere. I'm assuming you tried with only 1 rank, so 24 elems/rank. Some elements have more "owned" columns, while others have less. In particular, there's at least one element that owns all 16 columns, and there's at least one element that owns only 4 (no gll points on edges/corners). So with 24 team members, each member gets one element, but they might be unbalanced (the 16 columns element has 4x work than the 4 columns one), so a bit less might allow the thread that takes on the small element to also process another "small" element. So 20 is somewhere in the sweet spot, where each thread gets one element, and some threads get a couple, but probably they end up taking the smaller ones.

That's a bit "euristic", since the pd remap doesn't iterate over elements, but over physics columns. But still, if a thread ends up getting cols that are on the same element, it can get better caching of the dyn array.

One might wonder why larger team sizes don't pay off (basically putting two threads per element). I think it's because of better cache. CPU omp loops prefer to divide an array into chuncks, which mean prefer ||ization over the outermost index.

So I think these results make sense. And the fact that they hover around 1/7 for the total time proves that ||izing over fields is paying off.

components/scream/src/dynamics/homme/physics_dynamics_remapper.hpp
```diff
@@ -684,7 +684,14 @@ do_remap_fwd() const
         EKAT_KERNEL_ERROR_MSG("Error! Unhandled case in switch statement.\n");
     }
   };
-  const auto policy = ekat::ExeSpaceUtils<KT::ExeSpace>::get_default_team_policy(num_fields, 1);
+
+  const auto concurrency = KT::ExeSpace::concurrency();
+#ifdef KOKKOS_ENABLE_CUDA
+  const int team_size = std::min(1024, std::min(128*num_cols,32*(concurrency/nfields+31)/32));
+#else
+  const int team_size = (concurrency<num_fields ? 1 : concurrency/num_fields);
```
bartgol, line 692: If concurrency<num_fields (say, concurrency=6, and num_fields=7), a team size of 1 might not be great. When the last fields are processed (by a single thread), all other threads sit idle. Maybe we should set a minimum team size of 2, or 4?

Since we are planning on changing the team policy, it's not worth spending time on this though.

tcclevenger:
Here are the times for the new updates. The times for blake are very close, but team policy is changing so I think they are good enough for this PR.
Screenshot from 2021-03-25 16-57-26

components/scream/src/dynamics/homme/physics_dynamics_remapper.hpp
```diff
@@ -238,167 +293,467 @@ do_bind_field (const int ifield, const field_type& src, const field_type& tgt)
...
+
+    // If field has a parent, then its view has been subviewed. We
+    // do not want to remmap subviews, only the view of the parent,
+    // so we store this and no other info for these fields, then
+    // skip these fields during the remap.
+    if (ph.get_parent().lock() != nullptr) {
+      EKAT_REQUIRE_MSG(dh.get_parent().lock() != nullptr,
+                       "Error! If physics field has parent,"
+                       "dynamics field must also have parent.");
+      h_has_parent(i) = true;
```
bartgol, line 392: Should we check that, if has_parent is true, then the parent was also registered in the remapper? I'm wondering whether we might not realize we're not remapping the subfield b/c the parent was never registered...

components/scream/src/dynamics/homme/physics_dynamics_remapper.hpp
```diff
@@ -238,167 +293,467 @@ do_bind_field (const int ifield, const field_type& src, const field_type& tgt)
...
+{
+  const int i = team.league_rank();
+  const int itl = time_levels(0).first;
+  const auto& dim_d = dyn_dims(i).dims;
+
+  switch (phys_layout(i)) {
+    case etoi(LayoutType::Scalar2D):
+    {
+      if (is_state_field_dev(i)) {
+        auto dyn = Unmanaged<KokkosTypes<DefaultDevice>::view<ScalarT****>> (reinterpret_cast<ScalarT*>(dyn_ptrs(i).get()),
```
bartgol, line 466: These lines are quite long to parse. You might want to have in the class, something like

```
using device_type = typename field_type::device_type;
template<typename T, int N>
using uview_nd = Unmanaged<KokkosTypes<device_type>::view_ND<T,N>>;
```
And then, in the different fcns where you need them, simply do
```
auto dyn = uview_4d<ScalarT>(...);
```
It might be easier to parse.

bartgol (approved):
Nice work! I had a couple of comments, one cosmetic and the other more logical. I'm off tomorrow, so I'm gonna approve preemptively, disabling automerge, in case my comments are not relevant.

components/scream/src/dynamics/homme/physics_dynamics_remapper.hpp
tcclevenger: Yeah, that's a good idea.

components/scream/src/dynamics/homme/physics_dynamics_remapper.hpp
tcclevenger: And cleaner looking..

components/scream/src/dynamics/homme/physics_dynamics_remapper.hpp
```diff
@@ -589,396 +958,394 @@ setup_boundary_exchange () {
...
-      const int dim = phys_dims[1];
-      if (itl>=0) {
-        auto dyn  = dyn_field.template get_reshaped_view<ScalarT******>();
-        const int NumVerticalLevels = dyn.extent_int(5);
-        Kokkos::parallel_for(RangePolicy(0,num_cols*dim*NumVerticalLevels),
-                             KOKKOS_LAMBDA(const int idx) {
-          const int icol =  idx / (dim*NumVerticalLevels);
-          const int idim = (idx / NumVerticalLevels) % dim;
-          const int ilev =  idx % NumVerticalLevels;
+      auto phys = uview_nd<ScalarT,3> (reinterpret_cast<ScalarT*>(phys_ptrs(i).get()),
```
bartgol, line 1107: Not for this PR, but maybe when you do the team policy rework: you could make this even cleaner by having a routine

```
template<typename T, int N>
uview_nd<T,n> reshape (const Pointer& p, conost Dim& dim_specs) {
  // check inputs, maybe, and then reintepret case and create uview
}
```
and then in the code simply do
```
auto dyn = reshape<ScalarT,6>(dyn_ptr(i),dyn_dims(i));
```
It's mostly aesthetic, but might be easier to parse.

bartgol (approved):

Issue #936, opened 2021-03-16
Move remappers inside individual atmosphere processes


bartgol:
@AaronDonahue , @jeff-cohere , and I had a nice conversation about initialization sequence, which led us to realize that the infrastructure would be _a lot_ cleaner (without callbacks and hacky tricks) if the individual processes handled their remap. Now, it might seem like this would put more burden on the shoulders of the parametrization's developers. However, consider the following:

- we would still provide remappers, via `grids_manager->create_remapper(gridA,gridB)`;
- we are going to "hard-code" physics as the reference grid.

The first bullet ensures you're not alone if you decide to write a parametrization on a non-reference grid: you can still use the grids manager to retrieve a remapper from the physics grid (the reference one) to your grid. The second bullet point makes life easy for virtually all atmosphere processes: if you want to run on the physics grid, you don't need to worry about remapping inputs/outputs.

What are the advantages? Glad you asked.

- The class AtmosphereProcessGroup will become *much* slimmer. The code reduction in that class greatly outweigh the code increase in, say, the atmosphere dynamics code.
- The initialization sequence will be simplified. Right now, inputs are loaded on the ref grid (physics), but they are not remapped onto other grids; the remappers are invoked during the "run" phase only. This leads to, e.g., dynamics having to complete initialization during the first call to `run`. It is confusing (at best) to have initialization _not_ completed after the `initialize` call. With this modification, dynamics, e.g., will be storing a remapper, and will be able to remap what is needed for initialization onto the dyn grid.
- When developing an Atm Process it is easier t you have 10 clear steps to implement than

Issue #941, opened 2021-03-23
Allow remappers to receive field of const T as source/target fields (prevents one remap direction)


bartgol:
Once we move the remapper inside atm procs (see #936), we'll run into a const-correctness issue. Namely, required fields will only be handed to the atm proc as `Field<const Real>`. However, when registering them inside their inputs remapper, this will be a problem, since remappers are supposed to _always_ work both ways, and src/tgt fields need to be _both_ non const.

The solution is not too complicate:

- declare a `RemapDirection` enum, with `Forward, Backward, Bidirectional` values.
- add a template parameter to `AbstractRemapper`, of type `RemapDirection`, specifying in which direction the remapper can work.
- inside AbstractRemapper, differentiate the src and tgt field type, figuring out whether one of them is const. In particular, for Fwd remappers src field would be const and tgt non const, for Bwd remappers it would be the other way around, and for Bidirectional remappers they'd both be non-const (like they are now).
- in the remap method, allow fwd/bwd only if the template arg `RemapDirection` allows it.

Notice that, before moving remappers into atm procs, we did not need this, since all fields were handed over to AtmosphereProcessGroup (APG) (which is where the remappers were built) as non-const. The APG class would then hand them over to the individual atm proc with the correct constness, but was free to set them in the atm proc's input/output remapper since it had access to non-const versions.

Note: for homme, it just happens that during a normal  time step inputs and outputs are the same (state in and state out, with tendencies backed out), so the current impl would work. However, the current design is a bit flawed, and I'm afraid we'd run into serious headaches if we did not address this. If we decide to let it slip (since dynamics, which is the only process that currently needs remap, would work), this issue will serve as a reminder of what has to be fixed (and how) in case things change, and we have to address this issue.

bartgol:
I played with this issue a few months ago, but got stuck. After thinking about it, I came to the conclusion that we should instead have 3 reistration methods:
```
using nonconst_RT = std::remove_const<RealType>;
using nonconst_RT = std::add_const<RealType>;

using field_type            = Field<nonconst_RT>;
using const_field_type  = Field<const_RT>;

void register_field (const            field_type& src, const            field_type& tgt);
void register_field (const const_field_type& src, const            field_type& tgt);
void register_field (const            field_type& src, const const_field_type& tgt);
```
Then, at runtime, throw an error if the requested direction in `remap(bool forward)` requires to write on a const field.

PR #942, opened 2021-03-23
Move remappers inside atm procs

bartgol:
Addresses #936.

### DIFF
--- a/components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
+++ b/components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
@@ -1,4 +1,5 @@
 #include "atmosphere_dynamics.hpp"
+#include <memory>
 
 // HOMMEXX Includes
 #include "CaarFunctor.hpp"
@@ -52,6 +53,7 @@ void HommeDynamics::set_grids (const std::shared_ptr<const GridsManager> grids_m
 
   using namespace ShortFieldTagsNames;
   using namespace ekat::units;
+  using FID = FieldIdentifier;
 
   constexpr int NGP  = HOMMEXX_NP;
   constexpr int NVL  = HOMMEXX_NUM_PHYSICAL_LEV;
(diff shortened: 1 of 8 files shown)


bartgol:
The branch passes tests on my machine. However, there's one to do item, that I'm not sure whether should be in this PR or not.

I ended up changing Homme, to have all inputs being just the states. This is in line with the idea that Homme will back out tendencies. However, I haven't added logic to back them out (needs local copies of initial states, e.g). Should I add it here? Should I revert the changes so that Homme still uses FM,FQ,FT?

I'm tempted to leave the branch as-is. The next step is to make Homme do the right things for inputs/outputs. In particular, it should back out tendencies, as well as do T<->vTheta_dp conversions (in and out).

Note: one reason to remove FM, FT, and FQ as inputs is related to #941: the fields were going to be provided as `Field<const Real>` to Homme (since they're inputs), which prevented me from registering them inside a remapper (which, as things are now, requires both src and tgt to be non-const fields).

components/scream/src/share/atm_process/atmosphere_process_group.cpp
```diff
@@ -86,30 +86,13 @@ AtmosphereProcessGroup (const ekat::Comm& comm, const ekat::ParameterList& param
...
-  //    proper AbstractRemapper implementation for the reference-actual
-  //    grids pair. The APG stores internally a bunch of remappers,
-  //    which calls during the `run` call, to perform the remappings.
-
-  // Resizing the remappers vectors
-  m_inputs_remappers.resize(m_group_size);
-  m_outputs_remappers.resize(m_group_size);
+  // There is a catch when it comes to exposing a list of required/computed
+  // fields: if (and only if) the splitting is sequential, and contains
+  // 2+ processes, it is possible that some inputs of the non-first atm
```
jeff-cohere, line 91: Maybe rephrase: "..., it is possible that some inputs of atmosphere processes in the list are computed by processes that precede them."

components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
```diff
@@ -161,156 +146,70 @@ set_updated_group (const FieldGroup<Real>& group)
...
-  const int num_elems = state.num_elems();
-  const int num_tracers = tracers.num_tracers();
-  using Scalar = Homme::Scalar;
+  // Print homme's parameters, so user can see whether something wasn't set right.
+  ::Homme::Context::singleton().get<::Homme::SimulationParams>().print();
 
-  constexpr int NTL  = HOMMEXX_NUM_TIME_LEVELS;
-  constexpr int NVL  = HOMMEXX_NUM_LEV;
-  constexpr int NVLI = HOMMEXX_NUM_LEV_P;
+  // Now that all fields on the ref grid are set, create the fields on the
```
jeff-cohere, line 161: Is this just a leftover comment?

components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
```diff
@@ -161,156 +146,70 @@ set_updated_group (const FieldGroup<Real>& group)
...
-      auto ps_in = f.template get_reshaped_view<const Real*[NTL][NP][NP]>();
-      using ps_type = std::remove_reference<decltype(ps)>::type;
-      auto non_const_ptr = const_cast<Real*>(ps_in.data());
-      ps = ps_type(non_const_ptr,num_elems);
-    } else {
-      ekat::error::runtime_abort("Error! Unexpected field name. This is an internal error. Please, contact developers.\n");
-    }
-  }
+  // Finish homme initialization
+  // Homme::initialize_dp3d_from_ps_c();
```
jeff-cohere, line 168: Is this work still to be done?

jeff-cohere (approved):
This looks much simpler from the perspective of atmosphere processes, groups, and the DAG.

I think it's fine to back out the tendencies in Homme in another PR.

AaronDonahue:
I concur with Jeff, I think it would be okay to break the backing out of tendencies into a separate PR.

components/scream/src/share/atm_process/atmosphere_process_group.cpp
bartgol: Neat. Thanks!

components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
bartgol: Yup.

components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
bartgol: Ah, no. So, standalone Homme (from e3sm) does init dp3d from ps and from reference pressure levels thickness. However, in order to pick up ps from the initial conditions file, I would have had to expose ps as an input for HommeDynamics (otherwise it would not be given to Homme). Furthermore, even if I did that, `dp` would still need an I.C., according to the AD, since it's an input for the atm. Which would be confusing: you ask for an I.C. for `dp`, but it seems to not be refleted in the numbers (since the code will replace that with `dp=some_func(ps,ref_vert_coord)`.

I _think_ the easiest thing to do is to forget about `dp=some_func(ps,ref_vert_coord)`, and simply init `dp` with what's in the I.C. file. If you guys agree, I'll simply remove the commented line.

bartgol:
Sorry Aaron, I thought I asked also for your review, but I guess I didn't click correctly (or clicked twice).

components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
jeff-cohere: That sounds okay to me. What do you think, @AaronDonahue ?

jeff-cohere (commented):
Looks good to me, @bartgol . I'll wait to see what @AaronDonahue thinks about removing the comment related to `dp`'s initial conditions.

components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
AaronDonahue: I concur.  I'm guessing there will be some code cleanup in the future to address things like this.  For example if we decide it really doesn't make sense to include `dp` as an init'd variable.  But right now out goal is to get the code coupled and running.  So we can certainly revisit in the future if warranted.

jeff-cohere (approved):

bartgol:
> Looks good to me, @bartgol . I'll wait to see what @AaronDonahue thinks about removing the comment related to `dp`'s initial conditions.

Ah, shoot, I should have not added the automerge label then. Well, we can still think about dp I.C..

jeff-cohere:
> > Looks good to me, @bartgol . I'll wait to see what @AaronDonahue thinks about removing the comment related to `dp`'s initial conditions.
> 
> Ah, shoot, I should have not added the automerge label then. Well, we can still think about dp I.C..

Aaron already weighed in, so I think this PR is set.

bartgol:
Ah, I missed that. Great.

Issue #938, opened 2021-03-17
Calculate total st cloud fraction in SHOC (ast = max[alst,aist])


AaronDonahue:
Currently the shoc interface in the AD only handles liquid cloud fraction, alst.  But it is expected that macrophysics is responsible for computing the total cloud fraction as the maximum cloud fraction between liquid and ice.  This is later used by microphysics (P3).

PeterCaldwell:
I had expected this to already be in the C++ SHOC interface because it is in shoc_intr.F90: https://github.com/E3SM-Project/scream/blob/881f8b8717096d2ba91afcf48d26494afbb5c138/components/eam/src/physics/cam/shoc_intr.F90#L1010 . But it makes sense in retrospect that @tcclevenger  didn't include it because the F90 calls a function from the bowels of CAM to compute ice cloud fraction (aist).

It is also good that we don't have the existing F90 ice cloud fraction scheme coded up because it sucks and we have a git issue in the v0.1 campaign (#902 ) to replace it with a scheme that is just "aist=0 if qi<1e-5 else aist=1". Then we just need the additional "ast = max(aist, alst)" line to get the ast that microphysics wants.

PeterCaldwell:
Update: I just noticed that ast=max(aist,alst) is done in micro_p3_interface rather than shoc_intr as I originally thought: https://github.com/E3SM-Project/scream/blob/881f8b8717096d2ba91afcf48d26494afbb5c138/components/eam/src/physics/cam/micro_p3_interface.F90#L733

AaronDonahue:
@PeterCaldwell , I'm not sure I see where `ast` is calculated in micro_p3_interface.F90.  The cld_frac_l and cld_frac_i are both calculated from `ast` but not the other way around.  P3 doesn't have any sense of `alst` and `aist` as separate variables stored in PBUF.

AaronDonahue:
I think your original statement is correct, that it is handled in shoc.
https://github.com/E3SM-Project/scream/blob/881f8b8717096d2ba91afcf48d26494afbb5c138/components/eam/src/physics/cam/shoc_intr.F90#L1030

PeterCaldwell:
Oops. You're right. Thanks Aaron.

PeterCaldwell:
[...]

PR #944, opened 2021-03-23
Add cloud fraction calculation as an atmospheric process

AaronDonahue:
Establishes a new atmosphere process that can be used to calculate the ice cloud fraction and the total cloud fraction.

Addresses #938 

[BFB]


components/scream/src/physics/cld_fraction/atmosphere_cldfraction.cpp
```diff
@@ -0,0 +1,156 @@
...
+// =========================================================================================
+void CldFraction::initialize_impl (const util::TimeStamp& t0)
+{
+  m_current_ts = t0;
+}
+
+// =========================================================================================
+void CldFraction::run_impl (const Real dt)
+{
+/* Luca, do we need any of the following code?
```
AaronDonahue, line 63: @bartgol , I left a question in the code here for you.  This is copied from other atm. processes, but I'm not sure I understand what is happening here and if its needed for this code.

AaronDonahue:
For this stand alone test it was relatively easy to create a baseline file to test against.  I compared the baseline files locally using `cprnc`, but there isn't yet infrastructure for incorporating something like a BFB netCDF diagnostic in our tests.
1. Is this something we want to look at doing sooner than later?
2. More generally, should baselines be stored in the same directory as the tests themselves?  Or should we have some sort of `make baselines` equivalent for netCDF type tests?

AaronDonahue:
@bartgol and @PeterCaldwell , I'm adding you as reviewers.  Only the mappy test failed, but from the log I'm not sure if that might be more an issue that mappy had instead of a problem with the branch.

tcclevenger:
> @bartgol and @PeterCaldwell , I'm adding you as reviewers. Only the mappy test failed, but from the log I'm not sure if that might be more an issue that mappy had instead of a problem with the branch.

@AaronDonahue I'm seeing the same thing on one of my PRs..

components/scream/src/physics/cld_fraction/atmosphere_cldfraction.cpp
bartgol: No. I'm assuming you copied that from P3. These lines were for when the interface called f90 stuff. We can purge these lines (also from P3 and shoc).

components/scream/src/physics/cld_fraction/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
+set(CIMEROOT ${SCREAM_BASE_DIR}/../../cime)
```
bartgol, line 1: We don't need genf90, so lines 1-7 can be eliminated.

components/scream/src/physics/cld_fraction/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
...
+list(APPEND CMAKE_MODULE_PATH ${CIMEROOT}/src/CMake)
+
+set(GENF90 ${CIMEROOT}/src/externals/genf90/genf90.pl)
+set(ENABLE_GENF90 True)
+include(genf90_utils)
+include(Sourcelist_utils)
+
+set(CLDFRAC_SRCS
+  atmosphere_cldfraction.cpp
```
bartgol, line 10: Picky, but: since you spell `cld_fraction` in the folder name (2 "words", as opposed to single "word" `cldfraction`), you might do the same in the file: `atmosphere_cld_fraction.cpp`.

components/scream/src/physics/cld_fraction/atmosphere_cldfraction.cpp
```diff
@@ -0,0 +1,156 @@
...
+#include "ekat/util/ekat_units.hpp"
+
+#include <array>
+
+namespace scream
+{
+  using namespace cldfrac;
+// =========================================================================================
+CldFraction::CldFraction (const ekat::Comm& comm, const ekat::ParameterList& params)
+ : m_cldfraction_comm (comm)
```
bartgol, line 13: Doesn't look like you need the comm, so you might as well not store it.

components/scream/src/physics/cld_fraction/atmosphere_cldfraction.cpp
```diff
@@ -0,0 +1,156 @@
...
+  // Nevertheless, for output reasons, we like to see 'kg/kg'.
+  auto Q = kg/kg;
+  Q.set_string("kg/kg");
+  Units nondim(0,0,0,0,0,0,0);
+
+  const auto& grid_name = m_cldfraction_params.get<std::string>("Grid");
+  auto grid  = grids_manager->get_grid(grid_name);
+  m_num_cols = grid->get_num_local_dofs(); // Number of columns on this rank
+  m_num_levs = grid->get_num_vertical_levels();  // Number of levels per column
+  m_nk_pack  = ekat::npack<Spack>(m_num_levs);
```
bartgol, line 34: I don't see this used anywhere. Should it be purged?

components/scream/src/physics/cld_fraction/atmosphere_cldfraction.hpp
```diff
@@ -0,0 +1,106 @@
...
+  using host_view_in_type   = host_view_type<const_field_type::RT>;
+  using host_view_out_type  = host_view_type<      field_type::RT>;
+
+  std::map<std::string,host_view_in_type>   m_cldfraction_host_views_in;
+  std::map<std::string,host_view_out_type>  m_cldfraction_host_views_out;
+
+  std::map<std::string,const Real*>  m_raw_ptrs_in;
+  std::map<std::string,Real*>        m_raw_ptrs_out;
+
+  util::TimeStamp     m_current_ts;
```
bartgol, line 93: The base class already stores a time stamp (`t_`). Maybe you can use that one directly. Having two timestamps might be confusing.

components/scream/src/physics/cld_fraction/atmosphere_cldfraction.hpp
```diff
@@ -0,0 +1,106 @@
...
+  template<typename T>
+  using host_view_type = field_type::get_view_type<view_type<T>,Host>;
+
+  using host_view_in_type   = host_view_type<const_field_type::RT>;
+  using host_view_out_type  = host_view_type<      field_type::RT>;
+
+  std::map<std::string,host_view_in_type>   m_cldfraction_host_views_in;
+  std::map<std::string,host_view_out_type>  m_cldfraction_host_views_out;
+
+  std::map<std::string,const Real*>  m_raw_ptrs_in;
```
bartgol, line 90: Looks like you don't use raw ptrs, so you can purge these. They are used in parametrizations that need to call F90, but you do everything in c++.

Same thing for host views.

components/scream/src/physics/cld_fraction/atmosphere_cldfraction.hpp
```diff
@@ -0,0 +1,106 @@
...
+  void finalize_impl   ();
+
+  // Setting the fields in the atmospheric process
+  void set_required_field_impl (const Field<const Real>& f);
+  void set_computed_field_impl (const Field<      Real>& f);
+
+  std::set<FieldIdentifier> m_required_fields;
+  std::set<FieldIdentifier> m_computed_fields;
+
+  std::map<std::string,const_field_type>  m_cldfraction_fields_in;
```
bartgol, line 75: Same comment about spelling. `cld` and `fraction` are two words, so `cld_fraction` is probably a better name (definitely easier for the eye to parse).

components/scream/src/physics/cld_fraction/cldfraction_functions.hpp
```diff
@@ -0,0 +1,73 @@
...
+
+  using Scalar = ScalarT;
+  using Device = DeviceT;
+
+  template <typename S>
+  using BigPack = ekat::Pack<S,SCREAM_PACK_SIZE>;
+  template <typename S>
+  using SmallPack = ekat::Pack<S,SCREAM_SMALL_PACK_SIZE>;
+
+  using IntSmallPack = SmallPack<Int>;
```
bartgol, line 28: Never used, so can be purged.

components/scream/tests/scream_cldfraction/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
+INCLUDE (ScreamUtils)
```
bartgol, line 1: Does a cld_fraction standalone test makes sense? What are we testing here? If the impl is all we care, we could just put tests in `physics/cld_fraction/tests`, without bothering the with AD. Seems like a lot of work to setup a test.

bartgol (changes requested):
There's some code that can be cleaned up. When copying an impl from another folder, it's a good thing to do to go over the pieces, and see what's really used, so we can keep the code base smaller for maintenance (and readability) purposes.

bartgol:
> > @bartgol and @PeterCaldwell , I'm adding you as reviewers. Only the mappy test failed, but from the log I'm not sure if that might be more an issue that mappy had instead of a problem with the branch.
> 
> @AaronDonahue I'm seeing the same thing with Mappy on one of my PRs..

That's my fault. I temporarily hacked the mappy jenkins job to debug why PR 945 was failing, but then I forgot to bring it back to its original form. I did that now, so mappy should start to actually _test_ again. Add the retest label to trigger retest immediately (rather than wait 4h).

components/scream/src/physics/cld_fraction/atmosphere_cldfraction.cpp
```diff
@@ -0,0 +1,157 @@
...
+  // Layout for 3D (2d horiz X 1d vertical) variable defined at mid-level and interfaces 
+  FieldLayout scalar3d_layout_mid { {COL,LEV}, {m_num_cols,m_num_levs} };
+
+  // Set of fields used strictly as input
+  m_required_fields.emplace("qi",   scalar3d_layout_mid, Q,      grid_name);
+  m_required_fields.emplace("alst", scalar3d_layout_mid, nondim, grid_name);
+
+  // Set of fields used strictly as output
+  m_computed_fields.emplace("ast",   scalar3d_layout_mid, nondim, grid_name);
+  m_computed_fields.emplace("aist",  scalar3d_layout_mid, nondim, grid_name);
```
PeterCaldwell, line 48: Can we use cld_frac_l, cld_frac_i, and cld_frac_tot as liquid, ice, and total cloud fractions rather than aist, alst, and ast?

components/scream/src/physics/cld_fraction/atmosphere_cldfraction.cpp
```diff
@@ -0,0 +1,157 @@
...
+  // and the ice mass mixing ratio. 
+  auto qi   = m_cldfraction_fields_in["qi"].get_reshaped_view<const Pack**>();
+  auto alst = m_cldfraction_fields_in["alst"].get_reshaped_view<const Pack**>();
+  auto aist = m_cldfraction_fields_out["aist"].get_reshaped_view<Pack**>();
+  auto ast  = m_cldfraction_fields_out["ast"].get_reshaped_view<Pack**>();
+
+  cldfracF::cldfraction_main(m_num_cols,m_num_levs,qi,alst,aist,ast);
+
+  // Get a copy of the current timestamp (at the beginning of the step) and
+  // advance it,
```
PeterCaldwell, line 91: Why does the cldfraction process care about the timestep?

components/scream/src/physics/cld_fraction/atmosphere_cldfraction.cpp
```diff
@@ -0,0 +1,157 @@
...
+// =========================================================================================
+void CldFraction::finalize_impl()
+{
+  // Do nothing
+}
+
+// =========================================================================================
+void CldFraction::register_fields (FieldRepository<Real>& field_repo) const {
+  std::set<ci_string> q_names =
+    { "qv","qc","qr","qi","qm","nc","nr","ni","bm"};
```
PeterCaldwell, line 109: most of these variables aren't used by cldfraction, so why register them here?

components/scream/src/physics/cld_fraction/cldfraction_main_impl.hpp
```diff
@@ -0,0 +1,88 @@
...
+  const uview_1d<const Spack>& alst,
+  const uview_1d<const Spack>& aist,
+  const uview_1d<Spack>&       ast)
+{
+  team.team_barrier();
+  const Int nk_pack = ekat::npack<Spack>(nk);
+  Kokkos::parallel_for(
+    Kokkos::TeamThreadRange(team, nk_pack), [&] (Int k) {
+      auto icecld = aist(k) > alst(k);
+      ast(k).set(Smask(true), alst(k));
```
PeterCaldwell, line 78: Smask should be true for cells we're actually doing calculations in and false for cells included only as padding, right? Or is it false where we do calculations? Smask is a confusing name. In any case, where is Smask calculated before it is used here?

components/scream/tests/scream_cldfraction/CMakeLists.txt
PeterCaldwell: But that work is already done, right? And the more tests the better? I think this test just confirms that the code doesn't crash... Seems like a fine start.

Other tests could be: can we ever get cld_fraction>1 or <0? If qi>>1e-5 everywhere, is ice cldfrac=1 everywhere?. if qi=0 everywhere, is ice cldfrac=0 everywhere? If alst>aist, is ast=aist in cldfraction_calc_totalfrac? and vice versa.

components/scream/src/physics/cld_fraction/cldfraction_main_impl.hpp
```diff
@@ -0,0 +1,88 @@
...
+
+    cldfraction_calc_totalfrac(team,nk,oalst,oaist,oast);
+  });
+  Kokkos::fence();
+} // cldfraction_main
+/*-----------------------------------------------------------------*/
+template <typename S, typename D>
+KOKKOS_FUNCTION
+void Functions<S,D>
+::cldfraction_calc_icefrac(
```
PeterCaldwell, line 46: Do we need the cldfraction_ bit in the front of this and the next name? Isn't it obvious that this is a cloud fraction function from the rest of its name and because it is in the cldfrac namespace?

components/scream/src/physics/cld_fraction/atmosphere_cldfraction.cpp
AaronDonahue: I guess the only question is about updating the field time stamp.  Since cld_fraction will be the only process to update cld_frac_i and cld_frac_tot we need to have it grab the timestamp and use it in line 95.  @bartgol , am I correct here?

components/scream/src/physics/cld_fraction/atmosphere_cldfraction.cpp
AaronDonahue: Yes, I debated that myself.  I'm happy to drop it down to just `qi`, I left it in there this long just in case in the future we decide to use a scheme that relies on other tracer information.  But it is probably not unreasonable to expect the developer to also make this list longer if/when that happens.

components/scream/src/physics/cld_fraction/atmosphere_cldfraction.hpp
AaronDonahue: Does this also apply to P3 and SHOC which also don't call F90 code from the AD?

components/scream/src/physics/cld_fraction/cldfraction_main_impl.hpp
AaronDonahue: Smask is being constructed on demand here.  So `Smask(true)` passes a mask array of the appropriate size for the packed array `alst(k)`.  To answer your first question, this was my way to get `ast = max(alst,aist)`.  If there is a better way I'm happy to change it.  TBH, looking at this again, I think it would be just easier to set `ast(k)=alst(k)`, that should effectively copy the values of `alst` to `ast` and the mask properties, right?  << @bartgol

components/scream/tests/scream_cldfraction/CMakeLists.txt
AaronDonahue: Yes, the last step would be to put in unit tests.  I haven't done this yet.

components/scream/src/physics/cld_fraction/atmosphere_cldfraction.cpp
PeterCaldwell: It seems like if all FM variables need a timestamp, it would be easier to just pass the timestep they should use into all processes and have the timestep update occur at the top level rather than in individual processes??? I feel like I'm missing something here.

components/scream/src/physics/cld_fraction/atmosphere_cldfraction.cpp
PeterCaldwell: Yeah, let's not proactively add stuff people *might* want someday.

components/scream/src/physics/cld_fraction/cldfraction_main_impl.hpp
PeterCaldwell: Ok, your explanation make sense to me. And yeah, might as well ast(k)=alst(k) rather than using the mask.

PeterCaldwell:
2 other big-picture points:

1. it would be nice to make the 1e-5 qi threshold for being in-cloud a tuning parameter rather than a hardcoded constant. I'm fine with doing this in a later PR once we start tuning more, or you could put it in physics_constants now to annoy @bartgol !
2. we should also implement the qi threshold cld_frac_i method in the F90 code. My inclination is to just stuff an F90 version of calc_ice_cldfrac into shoc_intr.F90 and to make it a namelist option (as opposed to creating a separate process in F90 to mirror our C++ approach). I guess we could have BFB tests for the F90 versus C++ method, but it doesn't really seem worth it???

components/scream/src/physics/cld_fraction/atmosphere_cldfraction.hpp
bartgol: Yeah, I would remove those there as well (not in this PR, obviously). We did decide to drop completely F90 support, at some point, right? Not sure if it's worth spending effort (and risk bugs) for something we don't actively use anymore...

components/scream/tests/scream_cldfraction/CMakeLists.txt
bartgol: Well, Aaron already put in the effort, so I guess it's ok to keep them. And I guess it does test something more than just an individual function test, namely that hooks to the AD are correctly set up. Alright, I did not think that through.

components/scream/src/physics/cld_fraction/atmosphere_cldfraction.cpp
bartgol: I vote for `liq_cld_frac`, `ice_cld_frac`, and `tot_cld_frac`.

components/scream/src/physics/cld_fraction/atmosphere_cldfraction.cpp
bartgol: Well, we are not yet _really_ using the time stamp, but _ideally_ one should use the timestamp to figure out when a variable was last updated (at least on rank, we are not syncing time stamps across MPI, maybe we should, I don't know).

If we update all time stamps from the top level, the functionality kind of loses value. If each atm proc updates the ts of only what got updated during that call, maybe it can be more useful. That is, assuming that _some_ variables are not updated during _some_ calls. I'm thinking of something like "ah, it was night, so rad didn't update blah this time step". Maybe this never happens though, and everything is always updated. I don't know. If that's the case (i.e., everything always updated), then I think we could (should?) remove the time stamp from the fields.

Put it another way: what use can you envision from having the ability of checking the time stamp of a field?

components/scream/src/physics/cld_fraction/cldfraction_main_impl.hpp
bartgol: Two things:
- calling `a.set(Mask(true),b)` is the same as doing `a = b` (and might actually be better to avoid the mask, not sure);
- you can simply do `ast(k) = max(alst(k), aist(k)`, which should give you the same result as the current impl.

bartgol: This makes me wonder though: is `ast=max(aist,alst)`? I thought it was `ast=aist+alst`...

bartgol: Side note: the terms `Smask` and `Spack` always confused me. It's my Italian brain, probably: S-xyz in italian often means the opposite of xyz, so Smask sounds like 'unmask' to me... One day I'll propose for a different typedef. I haven't come up with one yet...

components/scream/src/physics/cld_fraction/cldfraction_main_impl.hpp
bartgol: Maybe we should rename the enclosing class: `Functions`->`CldFractionFunctions`.

components/scream/src/physics/cld_fraction/atmosphere_cldfraction.cpp
AaronDonahue: I would vote for a name that includes the words `liq` and `ice` instead of `l` and `i` since they are more descriptive. 

But more to the point of this PR, changing the field managed name would require us to change the FM name in SHOC and P3, and update the initial condition files for their stand-alone tests.  All of which I am happy to do, but do we want that to be in this PR, or a subsequent PR?

Alternatively, we establish these names here just for cld_fraction and then update shoc and p3 interfaces when it comes time to couple them.

@bartgol , @PeterCaldwell thoughts?

components/scream/src/physics/cld_fraction/atmosphere_cldfraction.hpp
AaronDonahue: This is also a holdover from copying from the p3 interface.  I think it isn't used there either.  Should we do an audit of all interfaces for these unused variables?

components/scream/src/physics/cld_fraction/atmosphere_cldfraction.cpp
AaronDonahue: It does look like I still need a `get_comm` getter function.  Since it is just an integer might as well store it so that the function makes sense.  See compile time error:

```
/usr/tce/packages/gcc/gcc-7.3.1/rh/usr/include/c++/7/ext/new_allocator.h(136): error: object of abstract class type "scream::CldFraction" is not allowed:
            pure virtual function "scream::AtmosphereProcess::get_comm" has no overrider
          detected during:
```

components/scream/src/physics/cld_fraction/CMakeLists.txt
AaronDonahue: I've made these changes everywhere.  In the code should we also split the term?  In particular, should the namespace be called `cld_fraction` instead of `cldfrac`?

components/scream/src/physics/cld_fraction/cldfraction_main_impl.hpp
AaronDonahue: I've made the change, `ast=max(aist,alst)` is much simpler.  For some reason I thought this would fail on a packed array.

AaronDonahue:
@bartgol and @PeterCaldwell , I pushed a newer version which addresses most of the review comments.  Feel free to take a look at the changes and my responses.

There are still a few outstanding questions that I need to work on once I get feedback from you.
1. @PeterCaldwell , do we want to implement this in the F90 code?  Out of curiosity is this because we also want SCREAMv0 to use this scheme?  Or is it just to have another BFB test.
2. The variable names: cld_frac_x vs X_cld_frac where x = [i,l,t] and X = [ice,liq,tot].  I asked about this in my response to the review comments.  It would be good to nail down both the name and how we want to implement.
3. `qi` -> ice_cloud trigger parameter.  I agree with @PeterCaldwell that this should probably be a parameter that can change.  A few implementation details,
    a. We could make this a runtime option, if so then it needs to be passed from the `input.yaml` file down into the process itself.
    b. or we can assign this to the universal physics constants header file as something like `min_icecld_qi_threshold`, or some better name.  Side note, this brings up a different infrastructure question about setting it up so that any of the physics_constants can be overwritten by changes in the input.yaml file.  That might make it easier to impact tuning parameters.  We could also have sub-structures within the physics constants that are labeled by scheme, if organization is a question (see Slack convo on this file from a few days ago).
4. timestamp discussion.  What do we want to do with this?

components/scream/src/physics/cld_fraction/cldfraction_main_impl.hpp
PeterCaldwell: I like that idea.

components/scream/src/physics/cld_fraction/cldfraction_main_impl.hpp
PeterCaldwell: About my thumbs down emoji - I mean that it is max(alst,aist) rather than alst+aist... I realized after "reacting" that thumbs down might be construed as saying I thought it was a dumb question rather than reacting "NOOOO!!!!" like I meant to. :-)

components/scream/src/physics/cld_fraction/cldfraction_main_impl.hpp
AaronDonahue: I went ahead and did this for the most recent version

components/scream/src/physics/cld_fraction/CMakeLists.txt
bartgol: I think so. It's more consistent.

components/scream/src/physics/cld_fraction/atmosphere_cldfraction.cpp
bartgol: Ah, shoot. Nobody is actually using that method, so far. I'm pretty sure I put it there for parallel splitting, but I can't remember.

components/scream/src/physics/cld_fraction/atmosphere_cldfraction.hpp
bartgol: Yes.

AaronDonahue: Okay great.  Maybe a subsequent clean-up PR instead of this one.

components/scream/src/physics/cld_fraction/atmosphere_cldfraction.cpp
bartgol: Well, I don't think this PR should change fields names in other parts of the code. That said, I don't think there's any test that uses both shoc/p3 _and_ IceCldFraction, right? So whatever name you pick here, is not going to cause conflicts.

But I guess it's ok to keep the same name as shoc/p3, and change it in the name-changing PR.

components/scream/src/physics/cld_fraction/atmosphere_cldfraction.cpp
PeterCaldwell: > what use can you envision from having the ability of checking the time stamp of a field?

I'm not sure why we're doing this. I think the idea was that we could check for race conditions when procs were parallel split? Like "don't update the state until all the expected processes are updated"? Otherwise I'm not sure when we'd use this. I don't think we will check that SW radiation isn't called at night, for example. Overall, I'm unsure this timestamp thing will be that helpful. Maybe I'm just suffering from a case of the grumpies.

> If we update all time stamps from the top level, the functionality kind of loses value.

I'm not sure I agree. As I see it, we can pass all processes the current timestep and if they update a variable, they assign that new timestamp to it. If not, they don't.

components/scream/src/physics/cld_fraction/cldfraction_main_impl.hpp
bartgol: If that's the case, when we rename variables, I vote for `max_cld_frac` for the "total" one. Total always makes me thing of a sum.

components/scream/src/physics/cld_fraction/atmosphere_cldfraction.cpp
PeterCaldwell: Yeah, I like Luca's names much better than mine. Maybe cld_frac_liq and cld_frac_ice, though, to emphasize that these are all cloud fractions first and flavors next. This is consistent with the naming convention for qi, qc, nc, ni, etc.

PeterCaldwell:
Replies to @AaronDonahue:
1. F90 version: we do want to use this parameterization in v0. A BFB test would be nice, but might be hard considering the code structure for ice cldfrac in F90 vs C++ is totally different??? In any case, I'd like to assign you to implement the F90 version and welcome you to decide for yourself whether to put it in this PR or a new one. 
2. cld_frac naming convention: I replied inline that I like cld_frac_ice etc.
3. tuning parameter: I think we should set the convention that tuning parameters are namelist settings rather than physics_constants. 
4. timestamps: this seems like a bigger issue which we should decide on independent of this PR. So I suggest we keep things as-is but talk about it in the v1 telecon or in a new git issue.

components/scream/src/physics/cld_fraction/cldfraction_main_impl.hpp
PeterCaldwell: I'm not sure the total cloud fraction always needs to be the max of ice or liquid cloud fractions, so I don't like max_cld_frac. It can never be the sum of cld_frac_liq and cld_frac_ice, though, because if both liquid and ice cloud fractions were 100%, our total cloud fraction would be 200%! I guess it could be something like cld_frac_tot = min( cld_frac_liq+cld_frac_ice, 1) though. That would be "minimally overlapped" total cloud fraction...

components/scream/src/physics/cld_fraction/atmosphere_cldfraction.cpp
PeterCaldwell: Yeah, maybe the easiest thing to do is to just leave (N)ast(Y) cld frac names in this PR and to do a subsequent PR that just switches everything to cld_frac_ice and so on later. Though like Luca says, naming conventions are compartmentalized for each proc, so they don't need to be consistent.

bartgol:
> 
>     1. @PeterCaldwell , do we want to implement this in the F90 code?  Out of curiosity is this because we also want SCREAMv0 to use this scheme?  Or is it just to have another BFB test.

If it's not used in v0, then we should not implement it in F90. Doing it "just to have another BFB test" would be nonsense, since the test would only verify that you wrote the same code twice.

>     2. The variable names: cld_frac_x vs X_cld_frac where x = [i,l,t] and X = [ice,liq,tot].  I asked about this in my response to the review comments.  It would be good to nail down both the name and how we want to implement.

I think "ice" and "liq" are specifying what the word "cld_fraction" refers to, and "cld" is specifying what "frac" refers to, so ice/liq should come first, followed by cld, and frac. At least that's my understanding of English. Like "killer whale lover" vs "whale lover killer" vs "whale killer lover"...

>     3. `qi` -> ice_cloud trigger parameter.  I agree with @PeterCaldwell that this should probably be a parameter that can change.  A few implementation details,
>        a. We could make this a runtime option, if so then it needs to be passed from the `input.yaml` file down into the process itself.
>        b. or we can assign this to the universal physics constants header file as something like `min_icecld_qi_threshold`, or some better name.  Side note, this brings up a different infrastructure question about setting it up so that any of the physics_constants can be overwritten by changes in the input.yaml file.  That might make it easier to impact tuning parameters.  We could also have sub-structures within the physics constants that are labeled by scheme, if organization is a question (see Slack convo on this file from a few days ago).
> 

If it's a physics _constant_, it should not be overwritten from input file (if you need to change it, change the src code). If you want the ability to overwrite it for certain runs, then maybe it's not a constant...

>     4. timestamp discussion.  What do we want to do with this?

Use the one from the base class (`t_`), and set it in all output fields. We should prob think on what we want to achieve with timestamps.

bartgol:
Seems like Peter and I agree on all items except the name one. I'm a believer that `ice_cld_frac` sounds more english than `cld_frac_ice`, since you probably say "the ice cloud fraction" rather than "the cloud fraction ice", when talking. Just like you say "the wind velocity" instead of "the velocity of the wind". But he outranks me, so... :)

PeterCaldwell:
I don't care that much about cld_frac_ice versus ice_cld_frac. I'm just trying to come up with/follow rules to make it easier to decide how to name whatever comes next. Let's leave it up to Aaron to make the deciding call, knowing that Luca and I don't really care either way.

components/scream/src/physics/cld_fraction/atmosphere_cldfraction.hpp
AaronDonahue: Okay cool, I've made a git issue, #948 , so that we don't lose sight of the task.

components/scream/src/physics/cld_fraction/atmosphere_cldfraction.hpp
AaronDonahue: Also could be handled by solving #948

components/scream/src/physics/cld_fraction/atmosphere_cldfraction.cpp
AaronDonahue: I'm in favor of changing the variable names to the better names and leave the field managed names for the cleanup we already see on the horizon.  That way coupling between this process and others is immediately possible, and the task of cleaning up the field managed name won't delay progress anywhere else.

components/scream/src/physics/cld_fraction/atmosphere_cldfraction.cpp
bartgol: > I'm not sure I agree. As I see it, we can pass all processes the current timestep and if they update a variable, they assign that new timestamp to it. If not, they don't.

I agree. I guess I misunderstood what you meant with "have the timestep update occur at the top level rather than in individual processes".  I thought you meant by the AD, or something. If the individual procs update the ts of what they update, then it does have some meaning.

bartgol: > I'm not sure why we're doing this. I think the idea was that we could check for race conditions when procs were parallel split? Like "don't update the state until all the expected processes are updated"? Otherwise I'm not sure when we'd use this. I don't think we will check that SW radiation isn't called at night, for example. Overall, I'm unsure this timestamp thing will be that helpful. Maybe I'm just suffering from a case of the grumpies.

Yeah, I'm not sure either. It _might_ help if we had 'related' vars (like T and Theta), and you want to check that the one you want to use is "more recent" than the other (otherwise you might be using stale data). I guess I assumed some vars might go through a few timesteps without being "actually" updated (e.g., solve rad only every other atm step, or something). But I don't think that will happen.

For now, we can keep the ts, I guess, but we should think about whether it can help or not. If not, we should purge it.

AaronDonahue:
@PeterCaldwell , per your comment
"
F90 version: we do want to use this parameterization in v0. A BFB test would be nice, but might be hard considering the code structure for ice cldfrac in F90 vs C++ is totally different??? In any case, I'd like to assign you to implement the F90 version and welcome you to decide for yourself whether to put it in this PR or a new one.
"

the code `components/eam/src/physics/cam/cldfrc2m.F90` already appears to have multiple options for calculating `aist` and `shoc_intr.F90` already calculates `ast=max(aist,alst)`.  Would it make the most sense for me to simply add another `iceopt` that uses this approach to find "aist"?  Rather than implement in a completely different module?

PeterCaldwell:
> Would it make the most sense for me to simply add another iceopt to cldfrc2m

Oh yeah, that makes way more sense than what I was saying! Thanks for thinking of it.

AaronDonahue:
> > Would it make the most sense for me to simply add another iceopt to cldfrc2m
> 
> Oh yeah, that makes way more sense than what I was saying! Thanks for thinking of it.

Alright, i went ahead and did that in this commit.  You can check out the latest commit or look at the changes I made in the `eam` sub-directory.

AaronDonahue:
This is very nearly ready for re-review.  I'm testing the F90 code at the moment and would like to adjust the C++ stand-alone test to capture the cases Peter mentioned above.

components/eam/bld/namelist_files/namelist_definition.xml
```diff
@@ -2490,9 +2488,16 @@ Default: 700.e2
        group="cldfrc_nl" valid_values="" >
 Scheme for ice cloud fraction: 1=wang & sassen, 2=schiller (iciwc),  
 3=wood & field, 4=Wilson (based on smith), 5=modified slingo (ssat & empyt cloud)
```
PeterCaldwell, line 2490: might as well fix "empyt" -> "empty" here.

components/eam/src/physics/cam/cldfrc2m.F90
```diff
@@ -953,7 +962,8 @@ subroutine aist_vector(qv_in, T_in, p_in, qi_in, ni_in, landfrac_in, snowh_in, a
      Kc = 75._r8
    ! Wilson & Ballard closure ( Option.4. smaller = more ice clouds)
    ! Slingo modified (option 5)
-     minice = 1.e-12_r8
+!ASD     minice = 1.e-12_r8 - Should be a namelist option now
+     if (iceopt.eq.7) minice = 1.e-5_r8
```
PeterCaldwell, line 966: Seems like we should specify minice in the compset defaults for our case rather than having it be a namelist option, but overriding the namelist if iceopt=7...

AaronDonahue:
@bartgol and @PeterCaldwell I've tested this PR for both the c++ and F90 versions.  It all seems to check out.  Can you re-review so that it can go in?

components/eam/bld/namelist_files/namelist_definition.xml
```diff
@@ -2489,10 +2487,17 @@ Default: 700.e2
...
+3=wood & field, 4=Wilson (based on smith), 5=modified slingo (ssat & empty cloud)
+6=Gettelman (based on Heymsfield), 7=all-or-nothing
 Default: 5
 </entry>
 
+<entry id="cldfrc_minice" type="real" category="conv"
+       group="cldfrc_nl" valid_value="" >
+Minimum ice mass mixing ratio used for calculating the ice cloud fraction,
+Used for cldfrc_iceopt = [4,5,6,7]
+Default: 1e-12
```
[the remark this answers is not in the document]
PeterCaldwell: I just pushed a fix rather than complain.

PeterCaldwell:
I looked through the F90 in detail and skimmed the C++. It all looks good to me. I'm not approving just so @bartgol  has a chance to check it out before auto-merging. Let me know if you want me to approve now. 

One thing that seems odd: do we mean to commit cld_fraction_standalone_ne4_ne4.nc ? Isn't this the output from a test?

AaronDonahue:
> I looked through the F90 in detail and skimmed the C++. It all looks good to me. I'm not approving just so @bartgol has a chance to check it out before auto-merging. Let me know if you want me to approve now.
> 
> One thing that seems odd: do we mean to commit cld_fraction_standalone_ne4_ne4.nc ? Isn't this the output from a test?

Ahh yes, that netCDF file is a holdover from before I made changes to the stand-alone test.  I'll delete that so the PR is cleaner.  Good catch.

components/scream/src/physics/cld_fraction/cld_fraction_main_impl.hpp
```diff
@@ -29,13 +29,13 @@ ::main(
     const Int i = team.league_rank();
 
     const auto oqi   = ekat::subview(qi,   i);
-    const auto oalst = ekat::subview(alst, i);
-    const auto oaist = ekat::subview(aist, i);
-    const auto oast  = ekat::subview(ast,  i);
+    const auto oliq_cld_frac = ekat::subview(liq_cld_frac, i);
```
bartgol, line 32: Definitely not for this PR, but I think at some point we need to change the names of the single-col vars. What is the `o` in front supposed to mean?

components/scream/src/physics/cld_fraction/cld_fraction_main_impl.hpp
```diff
@@ -47,15 +47,15 @@ ::calc_icefrac(
...
 {
   team.team_barrier();
   const Int nk_pack = ekat::npack<Spack>(nk);
   Kokkos::parallel_for(
     Kokkos::TeamThreadRange(team, nk_pack), [&] (Int k) {
       auto icecld = qi(k) > 1e-5;
-      aist(k) = 0.0;
-      aist(k).set(icecld, 1.0);
+      ice_cld_frac(k) = 0.0;
+      ice_cld_frac(k).set(icecld, 1.0);
```
bartgol, line 58: Note to self: we should add a method for Pack like `set(Mask m, Scalar val_if_m_true, Scalar val_if_m_false)`. This would allow to merge the two lines above into one single for loop.

components/scream/src/physics/cld_fraction/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
...
+  cld_fraction_functions.hpp
+  cld_fraction_main_impl.hpp
+)
+
+add_library(cld_fraction ${CLDFRAC_SRCS})
+target_include_directories(cld_fraction PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/../share)
+target_link_libraries(cld_fraction physics_share scream_share)
+target_compile_options(cld_fraction PUBLIC)
+
+#if (NOT SCREAM_LIB_ONLY)
```
bartgol, line 17: So I guess there's no unit test of cld_fraction? In particular, no bfb testing of v1-vs-v0? Is that just something we plan to do later, or are we ok without it?

components/scream/src/physics/cld_fraction/atmosphere_cld_fraction.cpp
```diff
@@ -0,0 +1,135 @@
...
+  m_computed_fields.emplace("ast",   scalar3d_layout_mid, nondim, grid_name);
+  m_computed_fields.emplace("aist",  scalar3d_layout_mid, nondim, grid_name);
+
+  // Set of fields used as input and output
+  // - There are no fields used as both input and output.
+
+}
+
+// =========================================================================================
+void CldFraction::initialize_impl (const util::TimeStamp& t0)
```
bartgol, line 55: Since `t0` is not used, you should comment it, to avoid compiler warnings. Simply put it inside comments, like `/* t0 */`.

components/scream/src/physics/cld_fraction/atmosphere_cld_fraction.cpp
```diff
@@ -0,0 +1,135 @@
...
+}
+
+// =========================================================================================
+void CldFraction::register_fields (FieldRepository<Real>& field_repo) const {
+  std::set<ci_string> q_names =
+    { "qi" };
+
+  for (const auto& fid : m_required_fields) {
+    const auto& name = fid.name();
+    if (q_names.count(name)>0) {
```
bartgol, line 95: Since there's only one entry in `q_name`, you could do something more readabe:

```
if (name=="qi")
```
and get rid of `q_names`.

components/scream/src/physics/cld_fraction/atmosphere_cld_fraction.cpp
```diff
@@ -0,0 +1,135 @@
...
+      field_repo.register_field<Pack>(fid,"TRACERS");
+    } else {
+      field_repo.register_field<Pack>(fid);
+    }
+  }
+}
+
+void CldFraction::set_required_field_impl (const Field<const Real>& f) {
+  // Store a copy of the field. We need this in order to do some tracking checks
```
bartgol, line 112: This comment is Homme-specific. It should be removed.

components/scream/src/physics/cld_fraction/atmosphere_cld_fraction.cpp
```diff
@@ -0,0 +1,135 @@
...
+  // in the Homme's view, and be done with it.
+  const auto& name = f.get_header().get_identifier().name();
+  m_cld_fraction_fields_in.emplace(name,f);
+
+  // Add myself as customer to the field
+  add_me_as_customer(f);
+}
+
+void CldFraction::set_computed_field_impl (const Field<      Real>& f) {
+  // Store a copy of the field. We need this in order to do some tracking updates
```
bartgol, line 124: Same here: comment to be removed.

components/scream/src/physics/p3/atmosphere_microphysics.cpp
```diff
@@ -206,7 +206,6 @@ void P3Microphysics::initialize_impl (const util::TimeStamp& t0)
 // =========================================================================================
 void P3Microphysics::run_impl (const Real dt)
 {
-  // std::array<const char*, num_views> view_names = {"q", "FQ", "T", "zi", "pmid", "dpres", "ast", "ni_activated", "nc_nuceat_tend"};
 
   std::vector<const Real*> in;
```
bartgol, line 210: Can we remove these two as well, while we're at it? They're never used. Probably a leftover of when P3 called the F90 impl.

components/scream/tests/scream_cld_fraction/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+INCLUDE (ScreamUtils)
+
+SET (cld_fractionLib cld_fraction)
```
bartgol, line 3: I would remove this intermediate variable. I assume you got the idea from Homme dynamics, but there the issue is that we may build several Homme dynamics libs (they have num levs and qsize hard coded in). Here it's just confusing, I think.

bartgol (changes requested):
It looks good. I had a few comments/requests.

components/scream/src/physics/cld_fraction/cld_fraction_main_impl.hpp
AaronDonahue: ??? I don't know.  I took this syntax from the p3_main_impl code.

components/scream/src/physics/cld_fraction/CMakeLists.txt
AaronDonahue: @PeterCaldwell can comment.  I made the decision to let the cld_fraction_stand_alone to take care of these sorts of tests.  As for BFB, the change I made to F90 was more surgical.  I think we may still be able to make a BFB test, but it would be a fair bit of work.  Not sure if it is high priority.

AaronDonahue:
@bartgol , Thanks for re-reviewing.  I think I've addressed all of your comments.

AaronDonahue:
I think I may need to rebase this PR too...

AaronDonahue:
okay, rebased

components/scream/src/physics/cld_fraction/CMakeLists.txt
PeterCaldwell: I'm conflicted about whether we need a BFB test for cld_fraction. On one hand, BFB tests are always good. On the other hand, our impl is only 3 lines of code or so, so it seems unlikely we'd screw it up. And the F90 version is entwined in a much bigger ice cloud fraction scheme that we don't want to port, while the F90 total cloud fraction calculation is stuffed into shoc_intr.F90. This makes creating a BFB test annoying. Thus I really don't think it should be part of this PR, but we might want to create a follow-on PR to do the BFB testing.

components/scream/src/physics/cld_fraction/cld_fraction_main_impl.hpp
PeterCaldwell: I think the o was just someone's lazy attempt to have multiple versions of the same variable name (to include both scalar and vector versions). I've never figured out what the o means. Maybe output? I agree this should be cleaned up... in another PR.

components/scream/src/physics/cld_fraction/CMakeLists.txt
AaronDonahue: At the risk of adding one more thing to the "TODO" pile, we could certainly open up a git-issue to add in a BFB test.  I'm in favor of not creating one now just because there are a number of higher priority items to address in pursuit of v1 and I'm pretty confident (given that its only a few lines) that the implementation of ice cloud fraction is correct.  As Peter pointed out, a BFB test with the F90 would probably take a few days and delay this PR going in.

bartgol: Sounds good to me. If the F90 impl is buried deep into other F90 routines, then it's not a trivial task, and one needs to weigh the benefits and costs. Aaron, you could remove those lines from the CMakeLists.txt, and add a 2 lines comment explaining why this folder does not have a 'tests' subdir, explaining what we discussed here.

bartgol (approved):

components/scream/src/physics/cld_fraction/CMakeLists.txt
AaronDonahue: done

PeterCaldwell (approved):

bartgol (approved):

components/scream/src/physics/cld_fraction/cld_fraction_main_impl.hpp
AaronDonahue: Conrad had a nice approach in the shoc code to use `_s` trailing after the variable name, presumably meaning "subview".

AaronDonahue:
@bartgol or @PeterCaldwell , because I pushed a rebase change this needs to be "rereviewed" before it can be merged.  Can either of you take one last look?

PeterCaldwell (approved):

PR #949, opened 2021-03-26
Use vector-valued wind in SHOC

tcclevenger:
In SHOC, instead of `u_wind` and `v_wind`, now just `horiz_wind`.

For BFB testing, in the `blah_f()` functions in `shoc_functions_f90.cpp` I just create an extra `horiz_wind` view and copy both `u_wind` and `v_wind` view into `horiz_wind`. This is not efficient, but I believe these functions are only really used for BFB testing so this was just easier. I can change it if needed.

### DIFF
--- a/components/scream/src/physics/shoc/atmosphere_macrophysics.cpp
+++ b/components/scream/src/physics/shoc/atmosphere_macrophysics.cpp
@@ -72,8 +72,6 @@ void SHOCMacrophysics::set_grids(const std::shared_ptr<const GridsManager> grids
   // Input/Output variables
   m_required_fields.emplace("s",        scalar3d_layout_mid, J/kg,        grid_name);
   m_required_fields.emplace("tke",      scalar3d_layout_mid, (m*m)/(s*s), grid_name);
-  m_required_fields.emplace("u",        scalar3d_layout_mid, m/s,         grid_name);
-  m_required_fields.emplace("v",        scalar3d_layout_mid, m/s,         grid_name);
   m_required_fields.emplace("wthv_sec", scalar3d_layout_mid, K*(m/s),     grid_name);
   m_required_fields.emplace("tkh",      scalar3d_layout_mid, (m*m)/s,     grid_name);
   m_required_fields.emplace("tk",       scalar3d_layout_mid, (m*m)/s,     grid_name);
@@ -81,8 +79,6 @@ void SHOCMacrophysics::set_grids(const std::shared_ptr<const GridsManager> grids
 
   m_computed_fields.emplace("s",        scalar3d_layout_mid, J/kg,        grid_name);
   m_computed_fields.emplace("tke",      scalar3d_layout_mid, (m*m)/(s*s), grid_name);
-  m_computed_fields.emplace("u",        scalar3d_layout_mid, m/s,         grid_name);
-  m_computed_fields.emplace("v",        scalar3d_layout_mid, m/s,         grid_name);
   m_computed_fields.emplace("wthv_sec", scalar3d_layout_mid, K*(m/s),     grid_name);
   m_computed_fields.emplace("tkh",      scalar3d_layout_mid, (m*m)/s,     grid_name);
   m_computed_fields.emplace("tk",       scalar3d_layout_mid, (m*m)/s,     grid_name);
@@ -93,28 +89,42 @@ void SHOCMacrophysics::set_grids(const std::shared_ptr<const GridsManager> grids
 
   // Tracer group
   m_inout_groups_req.emplace("TRACERS",grid->name());
+
+  // Wind group
+  m_inout_groups_req.emplace("horiz_wind",grid->name());
 }
 // =========================================================================================
 void SHOCMacrophysics::
 set_updated_group (const FieldGroup<Real>& group)
 {
-  EKAT_REQUIRE_MSG(group.m_info->size() >= 3,
-                   "Error! Shoc requires at least 3 tracers (tke, shoc_qv, shoc_ql) as inputs.");
-
   const auto& name = group.m_info->m_group_name;
 
-  EKAT_REQUIRE_MSG(name=="TRACERS",
+  EKAT_REQUIRE_MSG(name=="TRACERS" || name=="horiz_wind",
     "Error! We were not expecting a field group called '" << name << "\n");
 
   EKAT_REQUIRE_MSG(group.m_info->m_bundled,
-      "Error! Shoc expects bundled fields for tracers.\n");
+      "Error! Shoc expects bundled fields for tracers and wind.\n");
 
-  // Add Q bundle as in/out field
-  m_shoc_fields_in["Q"]  = *group.m_bundle;
-  m_shoc_fields_out["Q"] = *group.m_bundle;
+  if (name=="TRACERS") {
+    EKAT_REQUIRE_MSG(group.m_info->size() >= 3,
+                     "Error! Shoc requires at least 3 tracers (tke, shoc_qv, shoc_ql) as inputs.");
 
-  // Calculate number of advected tracers
-  m_num_tracers = m_shoc_fields_in["Q"].get_header().get_identifier().get_layout().dim(1);
+    // Add Q bundle as in/out field
+    m_shoc_fields_in["Q"]  = *group.m_bundle;
+    m_shoc_fields_out["Q"] = *group.m_bundle;
+
+    // Calculate number of advected tracers
+    m_num_tracers = m_shoc_fields_in["Q"].get_header().get_identifier().get_layout().dim(1);
+  }
+
+  if (name=="horiz_wind") {
+    EKAT_REQUIRE_MSG(group.m_info->size() == 2,
+                     "Error! Shoc requires exactly 2 wind vectors (u, v) as inputs.");
+
+    // Add V bundle as in/out field
+    m_shoc_fields_in["V"]  = *group.m_bundle;
+    m_shoc_fields_out["V"] = *group.m_bundle;
+  }
 }
 
 // =========================================================================================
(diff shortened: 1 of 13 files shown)


tcclevenger:
@AaronDonahue @PeterCaldwell @bartgol I've pinged you guys to discuss changing the input files for the shoc_stand_alone test (the only current failing test). Aaron has offered to update the inputs files, but I want to make sure there is agreement on how I'm handling wind in `atmosphere_macrophysics.cpp` before anyone puts in any work.

In `atmosphere_macrophysics.cpp` I now treat the old `u/v_wind` as a group `V` with tag `horiz_wind`. It's then handled just like the tracers. 

An issue I already see with this is that order matters, i.e., `u` must come before `v` in `V` for `shoc_main()` to compute correctly. A few options/questions:
- We could add a comment int the code explaining this.
- Maybe it's already expected from homme that this is the layout for `V`?
- The structure be changed where, instead of a group, having a required/computed field `horiz_wind` as a `view_3d(num_cols, 2, nlevs)`
- other?

Maybe there are even more issues that I'm not aware of with my impl so any comments are helpful.

AaronDonahue:
@tcclevenger , thank you for bringing this up.  Could I offer an "other", could we use "bundling" like we do for tracers?  That would allow shoc to grab the velocities as sub-views and not have to worr

AaronDonahue:
@tcclevenger , thanks for bringing this up.  Can I ask a follow-up question about the velocities in SHOC?

In the old system where you had `u` and `v` as separate fields did shoc still want them as a single vector `[u,v]`?  Or is the plan now to load `V` as a vector quantity and then extract `u` and `v`?

If the latter, would it be possible/beneficial to employ the "bundling" infrastructure that we are using for tracers and let `u` and `v` be subviews of `V`?  That would remove the concern about which comes first.  Although, I am guessing that homme already expects `u` to be the first velocity component.

tcclevenger:
> In the old system where you had `u` and `v` as separate fields did shoc still want them as a single vector `[u,v]`? Or is the plan now to load `V` as a vector quantity and then extract `u` and `v`?

Yes, the old system wanted u and v separate throughout SHOC. This PR combines them together inside SHOC as 
`horiz_wind = [u,v]`.  So `u(i,k)` entry in the old way is now `horiz_wind(i,0,k)`.

PeterCaldwell:
I think anyone who has dealt with a wind vector before would expect u to come before v... so I think it is fine that you're making that assumption in `horiz_winds`. I do think it is worth adding a comment somewhere in the code just clarifying that point in case it confuses someone sometime though. 

Regarding Aaron's question about getting u and v as subviews rather than using indexing to get u and v components directly from `horiz_winds`: I think this is a "6 of one, half dozen of the other" thing... Since you've already changed the code to use vector winds everywhere, let's just stick with that.

tcclevenger:
@PeterCaldwell I added the following check: When `v` is registered into group `V`, I assert the `u` has already been registered with the Error message that `u` must be registered first.

bartgol:
> @PeterCaldwell I added the following check: When `v` is registered into group `V`, I assert that `u` has already been registered with the Error message that `u` must be registered first.

I'm confused. If you handle a vector wind, why do you still need to subview u/v?

If you'd like to still use a scalar u/v, you can simply expose the vector wind as requirement, then, when the wind is passed to the atm proc via `set_required_field_impl`, you can do

```
void MyAtmProc::set_required_field_impl(const Field<const Real>& f) {
  auto fh = f.get_header();
  if (fh.get_identifier().name()=="horiz_wind") {
    // Assuming I'm on physics grid, horiz_wind has layout (COL,CMP,LEV)
    my_fields["u"] = f.subfield("my_u",1,0);
    my_fields["v"] = f.subfield("my_v",1,0);
  }
}
```
In other words, there's no need to put u/v in the FM, or even to create a "V" group. One can simply create u/v on the fly when the vector wind is passed. All the metadata will be synced, meaning that the fact that `horiz_wind` has `my_u` and `my_v` as subfield would be visible to the outside world.

tcclevenger:
@bartgol So if shoc only takes in `horiz_wind` as an input, should we just require `horiz_wind` in atmospheric_macrophysics?
That is, instead of treating `horiz_wind` like `TRACERS`, just treat it like a 3-dim required field?

More general question: in SHOC, the wind is often needed separately, e.g. in shoc_energy_integrals(),
```
ekat::square(u_wind(k))+ekat::square(v_wind(k))
```
changed to (in this PR)
```
ekat::square(horiz_wind(0,k))+ekat::square(horiz_wind(1,k))
```
This is typical of the use throughout SHOC. The only place where this is not the case is for the tridiag solve where wind is used as a block RHS. However, there we have to preform a transpose and make num_rhs the fastest index (i.e, `horiz_wind(nlevs,2)` instead of how it is passed to shoc: `horiz_wind(2,nlevs)`) , so we wouldn't be able to pass `horiz_wind` directly there either. 
So there is no real benefit to having wind be vector-valued, other than it would match homme. Is this a good enough reason for this PR, or should we just keep wind in shoc as is?

bartgol:
Since SHOC uses both u and v component of the wind, I think it makes sense to simply request for the vector component. As @PeterCaldwell pointed out, virtually all user will understand that `horiz_wind=[u,v]` rather than `[v,u]`.

Also, I don't think there's ant need to change SHOC's impl functions; they can keep `u_wind` and `v_wind`, we just need to change how the 1d views are created (subviewing `horiz_wind` rather than the two fields `u` and `v`). But the impl of all shoc's functions can stay the same (i.e., expect `u_wind` and `v_wind` in the interface).

tcclevenger:
@bartgol I changed atmosphere_macrophysics to just request the wind vector like you suggested (no groups). And I reverted back to using u_wind and v_wind inside shoc. Now shoc_main gets horiz_wind and then subviews correctly for all internal functions.

If we are ok with how things look then we can update the input file for shoc_standalone and test.

components/scream/src/physics/shoc/atmosphere_macrophysics.cpp
```diff
@@ -35,18 +35,21 @@ void SHOCMacrophysics::set_grids(const std::shared_ptr<const GridsManager> grids
...
   FieldLayout scalar2d_layout_col{ {COL}, {m_num_cols} };
-  FieldLayout scalar1d_layout_lev{ {LEV},  {m_num_levs} };
+  FieldLayout scalar2d_layout_lev{ {LEV},  {m_num_levs} };
 
   // Layout for 3D (2d horiz X 1d vertical) variable defined at mid-level and interfaces
   FieldLayout scalar3d_layout_mid { {COL,LEV}, {m_num_cols,m_num_levs} };
   FieldLayout scalar3d_layout_int { {COL,ILEV}, {m_num_cols,m_num_levs+1} };
 
+  // Layout for horiz_wind field
+  FieldLayout horiz_wind_layout { {COL,VAR,LEV}, {m_num_cols,2,m_num_levs} };
```
bartgol, line 45: Ok, I may have created more confusion than other by having both `VAR` (for Variable) and `CMP` (for Component) as field tags. In my mind, `VAR` was for variables like the tracers, where each "component" represent a completely different variable, possibly unrelated to the other ones, while `CMP` was for a component of a "physically" vector quantity. So in this case, I would have used `CMP` for  the velocity, rather than `VAR`.

The problem is that a field with layout `(COL,VAR,LEV)` is considered different from a field with layout `(COL,CMP,LEV)`, which causes issues with the atm dag. 

Should I get rid of one between `VAR` and `CMP`? It might be that the benefit of having both is minimal, while the risk of mismatched tags is non trivial. If so, I would remove `VAR` and keep `CMP`, which is more generic/flexible, I think. 

Tagging also @AaronDonahue and @jeff-cohere for their opinion on this.

components/scream/src/physics/shoc/atmosphere_macrophysics.cpp
```diff
@@ -70,61 +73,49 @@ void SHOCMacrophysics::set_grids(const std::shared_ptr<const GridsManager> grids
...
+  // Add Q bundle as in/out field
+  m_shoc_fields_in["Q"]  = *group.m_bundle;
+  m_shoc_fields_out["Q"] = *group.m_bundle;
 
-    // Add V bundle as in/out field
-    m_shoc_fields_in["V"]  = *group.m_bundle;
-    m_shoc_fields_out["V"] = *group.m_bundle;
-  }
+  // Calculate number of advected tracers
+  m_num_tracers = m_shoc_fields_in["Q"].get_header().get_identifier().get_layout().dim(1);
```
bartgol: This is perfectly fine, but for the record, another way (perhaps more verbose) is
```
m_num_tracers = group.m_info->size();
```

components/scream/src/physics/shoc/atmosphere_macrophysics.cpp
AaronDonahue: Yes, I would agree that we should only have one "generic" sounding dimension tag and that if we had to choose between VAR and CMP that CMP would be the way to go.  VAR does sound like "variable" which might be confusing.  While CMP, component, is more general.

components/scream/src/physics/shoc/atmosphere_macrophysics.cpp
tcclevenger: Yes, much better.

tcclevenger:
@AaronDonahue I think this should be ready to add the new input file with new wind data layout.

components/scream/src/physics/shoc/atmosphere_macrophysics.cpp
jeff-cohere: Also agree that one component tag is simpler. Let's just make sure to document how CMP is expected to be used and give some examples like the ones above (tracers, vector/tensor components, etc).

components/scream/src/physics/shoc/atmosphere_macrophysics.cpp
```diff
@@ -35,19 +35,22 @@ void SHOCMacrophysics::set_grids(const std::shared_ptr<const GridsManager> grids
 
   // Layout for 2D (1d horiz X 1d vertical) variable
   FieldLayout scalar2d_layout_col{ {COL}, {m_num_cols} };
-  FieldLayout scalar1d_layout_lev{ {LEV},  {m_num_levs} };
+  FieldLayout scalar2d_layout_lev{ {LEV},  {m_num_levs} };
```
AaronDonahue, line 38: This is a bit nitpicky, but wouldn't vertical level be considered a 1d scalar variable?

AaronDonahue (approved):
Looks good to me, I had just one nitpicky variable name comment that doesn't need to be addressed.

components/scream/src/physics/shoc/atmosphere_macrophysics.cpp
tcclevenger: ahh that's why I had it the other way. I'll change back.

components/scream/src/physics/shoc/atmosphere_macrophysics.cpp
bartgol: Ok, let me open an issue about it, so we don't forget.

bartgol (approved):

PR #954, opened 2021-03-31
Port Homme's ColumnOps framework to scream

bartgol:
Porting the column interpolation/differentiation utilities from Homme to SCREAM. Requires some adaptation, since HOMME has compile-time number of levels, and also assumes these utilities are called from within a TeamThreadRange.

### DIFF
--- a/components/homme/src/share/cxx/Config.hpp
+++ b/components/homme/src/share/cxx/Config.hpp
@@ -43,4 +43,11 @@
 # define HOMMEXX_CUDA_MAX_WARP_PER_TEAM 1
 #endif
 
+#if defined KOKKOS_COMPILER_GNU
+// See https://github.com/kokkos/kokkos-kernels/issues/129
+# define ConstExceptGnu
+#else
+# define ConstExceptGnu const
+#endif
+
 #endif // HOMMEXX_CONFIG_HPP
--- a/components/homme/src/share/cxx/utilities/BfbUtils.hpp
+++ b/components/homme/src/share/cxx/utilities/BfbUtils.hpp
@@ -118,7 +118,6 @@ bfb_pow_impl (ScalarType val, ExpType e) {
     Kokkos::abort("Cannot take powers of negative numbers.\n");
   }
   if (e<-1.0 || e>1.5) {
-    printf("Bad exponent: %3.15f\n",1);
     Kokkos::abort("bfb_pow x^a impl-ed with only -1.0<a<1.5 in mind.\n");
   }
   if (val==ScalarType(0)) {
--- a/components/homme/src/theta-l_kokkos/cxx/DirkFunctorImpl.hpp
+++ b/components/homme/src/theta-l_kokkos/cxx/DirkFunctorImpl.hpp
@@ -582,7 +582,9 @@ struct DirkFunctorImpl {
     };
     parallel_for(Kokkos::TeamThreadRange(kv.team, 1), f2);
     const auto f3 = [&] (const int km1) {
-      const auto k = km1 + 1;
+      // The following is morally a const var, but there are issues with
+      // gnu and std=c++14. The macro ConstExceptGnu is defined in share/cxx/Config.hpp.
+      ConstExceptGnu auto k = km1 + 1;
       const auto kr = [&] (const int i) {
         dpnh_dp_i(k,i) = ((pnh(k,i) - pnh(k-1,i))/
                           ((dp3d(k-1,i) + dp3d(k,i))/2));
@@ -721,7 +723,9 @@ struct DirkFunctorImpl {
     };
     parallel_for(pt1, f1);
     const auto f2 = [&] (const int km1) {
-      const auto k = km1 + 1;
+      // The following is morally a const var, but there are issues with
+      // gnu and std=c++14. The macro ConstExceptGnu is defined in share/cxx/Config.hpp.
+      ConstExceptGnu  auto k = km1 + 1;
       const auto kmid = [&] (const int i) { // middle Jacobian rows
         const auto b = 2*a/(dp3d(k-1,i) + dp3d(k,i));
         dl(k,i) = b*(pnh(k-1,i)/dphi(k-1,i));
--- a/components/scream/CMakeLists.txt
+++ b/components/scream/CMakeLists.txt
@@ -48,7 +48,7 @@ if (Kokkos_ENABLE_CUDA)
   EkatSetNvccWrapper()
 endif()
 
-set (CMAKE_CXX_STANDARD 11)
+set (CMAKE_CXX_STANDARD 14)
 
 if (NOT SCREAM_CIME_BUILD)
   project(SCREAM CXX Fortran)
--- a/components/scream/src/physics/p3/p3_upwind_impl.hpp
+++ b/components/scream/src/physics/p3/p3_upwind_impl.hpp
@@ -32,7 +32,10 @@ void Functions<S,D>
     // Add 1 to make [kmin, kmax). But then the extra term (Spack::n -
     // 1) to determine pack index cancels the +1.
     kmax = (kmax_scalar + Spack::n) / Spack::n;
-  const Int k_top_pack = k_top / Spack::n;
+
+  // The following is morally a const var, but there are issues with
+  // gnu and std=c++14. The macro ConstExceptGnu is defined in ekat_kokkos_types.hpp.
+  ConstExceptGnu Int k_top_pack = k_top / Spack::n;
 
   // calculate fluxes
   Kokkos::parallel_for(
--- a/components/scream/src/physics/shoc/shoc_assumed_pdf_impl.hpp
+++ b/components/scream/src/physics/shoc/shoc_assumed_pdf_impl.hpp
@@ -258,7 +258,7 @@ void Functions<S,D>::shoc_assumed_pdf(
       Spack s1(0), std_s1(0), qn1(0), C1(0), ql1(0),
             s2(0), std_s2(0), qn2(0), C2(0), ql2(0);
       {
-        const Scalar sqrt2(std::sqrt(2)), sqrt2pi(std::sqrt(2*pi));
+        const Scalar sqrt2(std::sqrt(Scalar(2.0))), sqrt2pi(std::sqrt(2*pi));
 
         // First plume
         const Spack cthl1=((1 + beta1*qw1_1)/ekat::square(1 + beta1*qs1))*(cp/lcond)*
--- a/components/scream/src/physics/shoc/shoc_assumed_pdf_impl.hpp
+++ b/components/scream/src/physics/shoc/shoc_assumed_pdf_impl.hpp
@@ -81,7 +81,9 @@ void Functions<S,D>::shoc_assumed_pdf(
   linear_interp(team,zi_grid,zt_grid,wqw_sec,wqw_sec_zt,nlevi,nlev,largeneg);
   linear_interp(team,zi_grid,zt_grid,qw_sec,qw_sec_zt,nlevi,nlev,0);
 
-  const Int nlev_pack = ekat::npack<Spack>(nlev);
+  // The following is morally a const var, but there are issues with
+  // gnu and std=c++14. The macro ConstExceptGnu is defined in ekat_kokkos_types.hpp.
+  ConstExceptGnu Int nlev_pack = ekat::npack<Spack>(nlev);
   Kokkos::parallel_for(Kokkos::TeamThreadRange(team, nlev_pack), [&] (const Int& k) {
 
     // Store active pack entries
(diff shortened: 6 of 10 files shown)


PeterCaldwell:
I think this is a great victory in the campaign to only have one function which does a particular thing. Thanks.

bartgol:
@jgfouca @jeff-cohere @tcclevenger 

I marked all of you as reviewers since this PR contains some technical stuff, and extra sets C++ expert eyes can help.

One thing I am questioning is the impl of `compute_interface_values`. I think it's not really a good impl, since it weighs neighboring cells equally, even if one cell is much larger than the other (and therefore its midpoint is "further away" from the interface).

I have a couple of variations, and I'm not sure which one to use:

- do a weighted avg of v_m to recover v_i, using layers thickness dx_m as weight:
```
 v_i(k+1/2) = ( v_m(k)*dx(k+1) + v_m(k+1)*dx(k) ) / ( dx_m(k) + dx_m(k+1) )
```
This is a convex combination of v_m(k) and v_m(k+1), which weighs more the midpoint that is "closer" to the interface. The downside of this forumula is that `v_i ( v_m (v_i)) != v_i`.

- do a scan sum, fixing x_i(top) or x_i(bot), by solving the inverting the formula
```
v_m(k) = ( v_i(k+1/2) + v_i(k-1/2) ) / 2
```
This would guarantee that `v_i ( v_m ( v_i ) ) == v_i`.

Option 1 is faster to implement, and does not require scan sum or reductions, but option two is appealing b/c it's the inverse of the int->mid interpolation.

That said, how often will we have to compute v_i from v_m? I'm not talking about v_i = scan(dv_m) (which is well defined), but a _bare_ interpolation from midpoints to interfaces.

@PeterCaldwell @AaronDonahue you might want to share your thoughts too.

bartgol:
Ok, I changed the implementation of the mid->int kernel, following option 2 above. Some details:

- the impl is different for packsize 1 or N>1.
- the scan sum that arises from inverting the midpoint interpolation rule contains alternating positive/negative values. To avoid accuracy loss, I perform the scan sum in two chunks: scan sum all even and odd entries separately, then combine the two results. Ironically, for this task it is easier the case with pack size N>1, since a packed scan is intrinsically keeping odd and even entries separated (save for the last pack, which is handled "unpacked"). I also perform the final pack reduction (from 1 pack down to 1 scalar) in a odd/even split way.
- the tests verify that v_int(v_mid(v_int))==v_int. 

An important point: all tests use integer values, since I did not put any thought into a serialization option. It should be a separate task, and most likely not as urgent as getting these utilities available for atm processes interfaces.

bartgol:
Also, I apologize for how messy the code is. Unfortunately, there are a few combinations to keep into account, like pack_size 1 vs N>1, or whether num packs is the same for midpoints and interfaces, or the scan direction.

If you guys think it could help, I can split the impl completely based on the scenario, like this:

```
compute_blah (...) {
  if (pack_size==1) {
    compute_blah_ps_1(...);
  } else if (num_int_packs==num_mid_packs) {
    compute_blah_ps_N_same_npacks (...);
  } else {
    compute_blah_ps_N_different_npacks (...);
  }
}
```
This will increase the number of lines, but it might make the code inside each function cleaner, I'm not 100% sure.

jeff-cohere:
> If you guys think it could help, I can split the impl completely based on the scenario, like this:

I like this idea. If it's not much work to do this, I think it might clarify what's going on. @jgfouca , @tcclevenger , what do you think?

jeff-cohere:
Whoops! Just noticed it was me that closed this earlier. Sorry about that.

bartgol:
Ok, I cleaned up a bit the implementation. I did not introduce a new method for each case (ps1, psN with int_packs=mid_packs, psN with int_packs>mid_packs), but I put the code for each case inside an if statement, which made code easier to follow, and also removed some lines of code that were there cause the fcn had to handle two scenarios at once. We still have those lines of code, but they are confined to the if branch that they belong to (e.g., some calls to `team_single`).

I also removed the implementation of scan sum for alternating sign midpoint values. I tried it on a sequence where I thought it might improve accuracy, but to my dismay it was no better than the standard scan sum. Better for who maintains the code, I guess.

jeff-cohere:
It's strange to me that the Blake and Mappy jobs are failing, because all of their tests pass. Does this indicate a specific problem with the PR, or is it just one of those AT things?

Weaver still seems to have a compilation error.

tcclevenger:
> I like this idea. If it's not much work to do this, I think it might clarify what's going on. @jgfouca , @tcclevenger , what do you think?

I agree, makes it a bit easier to follow.

bartgol:
> It's strange to me that the Blake and Mappy jobs are failing, because all of their tests pass. Does this indicate a specific problem with the PR, or is it just one of those AT things?
> 
> Weaver still seems to have a compilation error.

It might be that they fail in the build phase. We should probably improve our output in test-all-scream, to make that clear. I'll check the actual logs.

bartgol:
Ah, I forgot to update the ekat submodule, to bring in the Pack's methods overload for volatile cv, which caused build failures. With ekat updated, I get a pass on mappy for dbg build. I'll let the AT do the rest.

bartgol:
I verified that with the changes in E3SM-Project/EKAT#102, and changing scream's CXX standard to 14, this branch compiles on weaver.

I will put the PR on hold, and as soon as EKAT's PR is merged, I'll update the submodule in this branch, and retest.

bartgol:
Turns out that the team-level parallel_scan only works for team sizes that are powers of 2. I _think_ this will always be fine for us, since our team size is likely going to be num_levels, which is a pow of 2 for scream.

Unfortunately, I cannot change the block's dimension (aka, the team size) from within the || iteration, so to get around this limitations, in case of team sizes not pow of 2 one would need to either serialize the || scan or roll our own || scan. Needless to say, both "solutions" are not great. IMHO, we can simply check that the column_scan op is not used for team sizes not pow of 2. If someone wants to use it with a funky number of levels, he/she would be forced to "bump up" the team size (and have idle threads), or reduce it.

I don't think this is a limitation that should make us lose sleep at night...

PeterCaldwell:
@bartgol - I think this is what you're suggesting: check that team size is a factor of 2 and if so use the existing parallel scan. If not, throw an error which can either be fixed by hacking (padding the team size) or by implementing a 'roll your own' parallel or serial implementation. If so, I agree that this is the best approach.

One other question - you're just grabbing the impl from HOMME and making it available to SCREAM, right? So HOMME already has this limitation and seems to survive fine? Do we anticipate that SCREAM would use this impl with a different team policy than HOMME, or are we already in a situation where powers of 2 are needed?

bartgol:
> @bartgol - I think this is what you're suggesting: check that team size is a factor of 2 and if so use the existing parallel scan. If not, throw an error which can either be fixed by hacking (padding the team size) or by implementing a 'roll your own' parallel or serial implementation. If so, I agree that this is the best approach.

Basically, yes. I think these routines won't be used much outside the preamble/postamble of each process, so they should not be huge bottlenecks, even if their impl is suboptimal.
 
> One other question - you're just grabbing the impl from HOMME and making it available to SCREAM, right? So HOMME already has this limitation and seems to survive fine? Do we anticipate that SCREAM would use this impl with a different team policy than HOMME, or are we already in a situation where powers of 2 are needed?

Well, HOMME has a different grid structure, which calls for a different parallelization scheme. In Homme, we use 3 layers of parallelism: over elements (possibly times number of variables, like tracers), then over np*np GLL points, and finally over levels. In SCREAM (physics), we use 2 layers of parallelism: over columns, and over levels. Focusing just on GPU (on CPU there are virtually no constraints to implementation), we have (pseudo code)

HOMME:
```
pfor (nelems) { // outer pfor
  pfor (np*np) { // A "TeamThreadRange", in kokkos terms
     pscan(nlevs) {...} // A "ThreadVectorRange" in kokkos terms
  }
}
```
SCREAM:
```
pfor (ncols) { // outer pfor
  pscan (nlevs) { // A "TeamThreadRange", in kokkos terms
  }
}
```
Note: in scream, we use a team policy with vector_length=1.

Kokkos offers a parallel scan for `ThreadVectorRange` "without" limitations. The quotes are b/c kokkos already enforces vector length to be a pow of 2 (due to hardware reasons), so that is somewhat "hidden", and always verified. The parallel scan for `TeamThreadRange` is different, since team sizes need not be pow of 2 (in general). 

I think it is fine to assume we will always use team policies with team_size=2^n (note: we check, just to give the user a better error msg than kokkos). Thinking about it, this is not a _real_ limitation. CUDA works with warps of threads, each thread in the warp executing the same instruction.So even for pfor, if nlev!=2^n, we would still need to use a team_policy with team_size=2^n, having some threads sitting idle.

Bottom line: I am not concerned. This _looks_ like a limitation of parallel_scan, but I think in the end is not. Yes, pscan _cannot_ run if team_size!=2^n, but we would never use team_size!=2^n on CUDA anyways.

PeterCaldwell:
Ok, yeah that makes sense. Thanks for the explanation.

From: Luca Bertagna ***@***.***>
Reply-To: E3SM-Project/scream ***@***.***>
Date: Tuesday, April 6, 2021 at 8:59 AM
To: E3SM-Project/scream ***@***.***>
Cc: Peter Caldwell ***@***.***>, Mention ***@***.***>
Subject: Re: [E3SM-Project/scream] Port Homme's ColumnOps framework to scream (#954)


@bartgol<https://urldefense.us/v3/__https:/github.com/bartgol__;!!G2kpM7uM-TzIFchu!jvlyb9SMdP2Z1UHtFuqH2q5tFdXBeM8lB7EnL-8odoOj1Idhme_Ys4JMY1wFIrbq0z3G$> - I think this is what you're suggesting: check that team size is a factor of 2 and if so use the existing parallel scan. If not, throw an error which can either be fixed by hacking (padding the team size) or by implementing a 'roll your own' parallel or serial implementation. If so, I agree that this is the best approach.

Basically, yes. I think these routines won't be used much outside the preamble/postamble of each process, so they should not be huge bottlenecks, even if their impl is suboptimal.

One other question - you're just grabbing the impl from HOMME and making it available to SCREAM, right? So HOMME already has this limitation and seems to survive fine? Do we anticipate that SCREAM would use this impl with a different team policy than HOMME, or are we already in a situation where powers of 2 are needed?

Well, HOMME has a different grid structure, which calls for a different parallelization scheme. In Homme, we use 3 layers of parallelism: over elements (possibly times number of variables, like tracers), then over np*np GLL points, and finally over levels. In SCREAM (physics), we use 2 layers of parallelism: over columns, and over levels. Focusing just on GPU (on CPU there are virtually no constraints to implementation), we have (pseudo code)

HOMME:

pfor (nelems) { // outer pfor

  pfor (np*np) { // A "TeamThreadRange", in kokkos terms

     pscan(nlevs) {...} // A "ThreadVectorRange" in kokkos terms

  }

}

SCREAM:

pfor (ncols) { // outer pfor

  pscan (nlevs) { // A "TeamThreadRange", in kokkos terms

  }

}

Note: in scream, we use a team policy with vector_length=1.

Kokkos offers a parallel scan for ThreadVectorRange "without" limitations. The quotes are b/c kokkos already enforces vector length to be a pow of 2 (due to hardware reasons), so that is somewhat "hidden", and always verified. The parallel scan for TeamThreadRange is different, since team sizes need not be pow of 2 (in general).

I think it is fine to assume we will always use team policies with team_size=2^n (note: we check, just to give the user a better error msg than kokkos). Thinking about it, this is not a real limitation. CUDA works with warps of threads, each thread in the warp executing the same instruction.So even for pfor, if nlev!=2^n, we would still need to use a team_policy with team_size=2^n, having some threads sitting idle.

Bottom line: I am not concerned. This looks like a limitation of parallel_scan, but I think in the end is not. Yes, pscan cannot run if team_size!=2^n, but we would never use team_size!=2^n on CUDA anyways.

bartgol:
Good news: kokkos 3.3 seems to pass tests on Weaver (after a minor mod). As soon as the PR goes in in ekat, I can update the submodule here.

One big caveat: gcc seems to have an annoying bug with c++14, which manifests at least with 7.2 and 7.4 (I haven't tried newer versions yet): some const variables inside lambda functions cause a parsing error. Ekat was already aware of this, since someone (@jgfouca  maybe) introduced the `ConstExceptGnu` macro, to get around this. However, now that c++14 will be the default (as required by kokkos 3.3), we might see more of this bug. It's not a big deal, we simply have to use the ConstExceptGnu macro when needed. If we notice too many occurrences, we can test newer versions of gcc, and see if the bug disappear (and if so, ask the testbeds maintainers to provide such version of gcc on weaver).

bartgol:
Now that the EKAT pr for the kokkos submodule update went in, this PR can again be tested and reviewed.

bartgol:
@jeff-cohere @jgfouca @tcclevenger Looks like this is ready for review.

jeff-cohere (approved):
Very cool. Nice job with the comments and the organization. It's not easy to make fancy C++ intelligible.

tcclevenger (approved):
Looks great to me!

bartgol:
Given the two approvals, I'm gonna go ahead and merge this, so that we can start using the kernels for pre/post processing inside atm processes.

Issue #957, opened 2021-03-31
Remove tkh from SHOC interface


PeterCaldwell:
shoc_main doesn't need to return tkh - it isn't used outside of SHOC.

PR #972, opened 2021-04-01
make tkh local in c++

tcclevenger:
Makes `tkh` a local variable in c++ `shoc_main` impl. It's an input/output currently, but the first time it is used (`eddy_diffusivities()`) all its values are reset so it isn't required as an input, and from issue https://github.com/E3SM-Project/scream/issues/957 it is also not needed as an output.

The way I get around testing is just to skip any comparison of fortran `tkh` to c++ `tkh` (see line 94 of `shoc_run_and_cmp.cpp`). This is hacky, but I think the alternative is to remove `tkh` as input/output of fortran code and remove any reference in the test data? Which I have no problem doing.


PeterCaldwell:
Seems like a reasonable plan to me since we're going to ditch the F90 eventually anyways. This is also a good paradigm for the other variable removal tasks, like @AaronDonahue has....

PeterCaldwell:
Looks great to me. I'm not approving just so it doesn't get automerged before Jim and Aaron get a chance to check it over.

jgfouca (approved):

AaronDonahue (approved):
Looks good to me.

PR #973, opened 2021-04-01
Change field manager short names so that they are consistent across processes

AaronDonahue:
This commit will update the field repository short names for all processes currently in SCREAM to match the agreed upon names here: https://acme-climate.atlassian.net/wiki/spaces/NGDNA/pages/1927512393/Field+Manager+Variable+Names?focusedCommentId=2538635319#comment-2538635319

This will also ensure that all processes use the same shortname for variables they share, facilitating multi-process coupling in the future.

Addresses #818


AaronDonahue:
This PR changes the names of variables in each of the processes.  
@bartgol  - HOMME
@tcclevenger  - SHOC
@brhillman  - RRTMGP
Could each of you take a look at the changes I made to your specific process to make sure I didn't miss anything or make a mistake?

tcclevenger (approved):
SHOC looks perfect to me. Thanks @AaronDonahue !

bartgol (approved):
Looks good. I still vote for `T_mid` instead of `T_atm_mid`. If other temperatures are needed (say, the ocean), then _those_ can use specifiers (e.g., `T_ocn`). All other vars do not have `_atm`, so having one seems off (and triggers my developer OCD, I guess). But I am going to approve in case I'm outvoted on this.

AaronDonahue:
> Looks good. I still vote for `T_mid` instead of `T_atm_mid`. If other temperatures are needed (say, the ocean), then _those_ can use specifiers (e.g., `T_ocn`). All other vars do not have `_atm`, so having one seems off (and triggers my developer OCD, I guess). But I am going to approve in case I'm outvoted on this.

This has been addressed in the latest commit.

AaronDonahue:
I rebased to rebased to resolve conflicts, so it will need a re-review.  Since this is a SHOC conflict, can you rereview @tcclevenger

tcclevenger (approved):

components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
```diff
@@ -36,25 +36,25 @@ namespace scream {
         FieldLayout scalar2d_swband_layout { {COL,SWBND}, {m_ncol,m_nswbands} };
 
         // Set required (input) fields here
-        m_required_fields.emplace("pmid" , scalar3d_layout_mid, Pa, grid->name());
+        m_required_fields.emplace("p_mid" , scalar3d_layout_mid, Pa, grid->name());
         m_required_fields.emplace("pint", scalar3d_layout_int, Pa, grid->name());
```
brhillman, line 40: Is it too much to ask to change `pint` and `tint` to `p_int` and `T_int` for consistency with the change in `pmid` and `tmid`?

AaronDonahue: Great suggestion.  I can put that into this PR

tcclevenger: There is a `pint` in shoc as well.

bartgol: Being Friday, "pint"s don't sound that bad after all.. ;)

PeterCaldwell: p_int is supposed to be a calculated rather than a FM variable so why is it showing up in the FM at all?

AaronDonahue: I think the idea is to do thinks in pieces.  We don't yet have the infrastructure to calculate pint.  Once that is in we will remove pint from the FM.  So from one perspective, maybe we don't change these now because they may be replaced/removed in the future (?).  But we should keep them in the FM at least, until we are ready to calculate them.

bartgol: @PeterCaldwell I think this PR is only changing var names, to make sure all atm procs agree on names. The work to ensure atm procs only pass aroudn vars that are listed on the confluence FM page will come after this (in several separate PRs).

AaronDonahue: So, if the plan is too eventually replace these variables with local variables that are derived, would it make sense to go ahead and merge this PR now and take care of the variable names when we get to that point?

PeterCaldwell: Yeah, the explanation of creating the function for p_int in a future PR makes sense. I just don't think it's worth worrying about p_int vs pint if we're going to delete it eventually anyways. Merge away.

bartgol (approved):

tcclevenger:
@AaronDonahue When running the shoc standalone test on the current master branch, I get 
```
PIO ERROR: could not find variable T_mid in file shoc_standalone_ne4_ne4.nc
```
 I noticed this when the auto tester for PR https://github.com/E3SM-Project/scream/pull/974 was passing even though I had not updated the input file, so not sure if the auto tester is calling the standalone tests properly.

PR #974, opened 2021-04-01
add shoc postprocessing

tcclevenger:
Adds post processing to shoc interface.


AaronDonahue:
@tcclevenger , I think this PR and #973 will have a bunch of conflicts.  So we may want to time merging.  Any thoughts?

tcclevenger:
> @tcclevenger , I think this PR and #973 will have a bunch of conflicts. So we may want to time merging. Any thoughts?

Yes, It actually was an accident that I created mine. I plan to wait until yours is merged before requesting reviews.

bartgol:
For once, I'm happy to see the AT failures...

components/scream/src/physics/shoc/atmosphere_macrophysics.cpp
```diff
@@ -164,49 +167,52 @@ void SHOCMacrophysics::initialize_impl (const util::TimeStamp& t0)
           exner("exner",m_num_cols,nlev_packs),
           thlm("thlm",m_num_cols,nlev_packs),
           qw("qw",m_num_cols,nlev_packs),
-          cloud_frac("cloud_frac",m_num_cols,nlev_packs);
+          qv_copy("qv_copy",m_num_cols,nlev_packs),
+          qc_copy("qc_copy",m_num_cols,nlev_packs),
+          tke_copy("tke_copy",m_num_cols,nlev_packs);
```
tcclevenger, line 173: These tracers are treated as separate input/output SHOC. Here, I need to make a copy since they will also be updated in the `Q` bundle. Then in the postamble, I copy these tracer copies back into the `Q` bundle. Not sure if this is the correct thing to do..

tcclevenger:
@AaronDonahue and @bartgol Could y'all take a look? I left a comment about the tracers which are treated separately, which I'm having to make a copy of for the shoc_main call.

components/scream/src/physics/shoc/atmosphere_macrophysics.cpp
bartgol: Why can't you just pass qv,qc,tke? Do you need to "freeze" the value of these at the beginning of shoc step? I.e., do you need to "discard" whatever shoc does to those three tracers when handling Q? If so, then maybe shoc should request a subset of the tracers... But I thought SHOC needed to advect _all_ tracers (at least for now, without aerosols in scream).

tcclevenger: Yes, I believe I need to discard the values updated in `Q`.

In shoc_intr.F90, `tke` (and the `qv`,`qc`) is copied from `state%Q(tke_indx)` and updated as a separate input/output and as a part of `Q` in `shoc_main(...,tke,...Q,...)`. Then, when the tendencies are updated after `shoc_main()`,  `physics_tend%Q(tke_indx)` uses `tke` to update its value instead of the version in `Q` bundle. 

Should the shoc interface not recognize tke/qc/qv as tracers and keep them out of the bundle? Maybe change the name to "advected tracers" instead of just "tracers"?

PeterCaldwell: I'm not sure I understand this situation, but I'd really like to avoid having multiple copies of the same variables floating around (because it is confusing and invites errors). I'm pretty sure qv, qc, and tke need to be advected by the dycore, so it makes sense for them to be included in the field manager's "tracers" array. I think these 3 variables need to be handled specially withing SHOC - tke is centrally important in SHOC for determining other turbulent moments and SHOC advects qt = qv+qc rather than qv and qc independently. It also repartitions qt into qv and qc components using its PDF. So SHOC needs these variables to *not* be in the array that it applies turbulent diffusion to. I don't know if we want to handle this by only turbulent-diffusing the first N indices in the tracers array, or whether we want to advect all terms than overwrite qv, qc, and tke later, or whether we want to create a new subset "shoc_tracers" array which doesn't include qc, qv, and tke. Hopefully this made sense.

bartgol: It is not desirable to do work for nothing (i.e., advect but then overwrite), _especially_ on CPU. I suggest we make the shoc_tracers group (or maybe, "turbulent_diffusion" or something like that), which inherits everything in the "tracers" group, but excludes "qv,", "qc", "tke". We'll let the FM allocate the tracers in a way that makes shoc's tracers a _contiguous_ subset of the tracers indices.

tcclevenger: From your comment on slack, should we keep it as is here and fix this separately?

bartgol: Yes. I don't want to hold this PR back waiting for the "optimal" version. We should do things incrementally. I'm already working on allowing atm procs to request "subsets" of groups, so we don't risk to leave this behind.

components/scream/src/physics/shoc/atmosphere_macrophysics.cpp
```diff
@@ -164,49 +167,52 @@ void SHOCMacrophysics::initialize_impl (const util::TimeStamp& t0)
           exner("exner",m_num_cols,nlev_packs),
           thlm("thlm",m_num_cols,nlev_packs),
           qw("qw",m_num_cols,nlev_packs),
-          cloud_frac("cloud_frac",m_num_cols,nlev_packs);
+          qv_copy("qv_copy",m_num_cols,nlev_packs),
+          qc_copy("qc_copy",m_num_cols,nlev_packs),
```
bartgol, line 171: Can you add a TODO comment (here, or maybe in the hpp file), reminding us to remove the `*_copy` vars once shoc can request a subgroup of the tracers?

bartgol (approved):

components/scream/src/physics/shoc/atmosphere_macrophysics.cpp
PeterCaldwell: I like Luca's plan.

bartgol:
I'm gonna merge this, since other PRs need the p3/shoc IC nc files fixes in order to pass.

components/scream/src/physics/shoc/atmosphere_macrophysics.cpp
tcclevenger: I'll make a small PR.

Issue #975, opened 2021-04-02
[AD] At initialilzation make it possible to assign a variable to copy another variable


AaronDonahue:
There are some variables in SCREAM that are set equal to other variables at initialization.  In particular, in P3 the two variables `T_prev_micro_step` and `qv_prev_micro_step` are set to `T_mid` and `qv` respectively at initialization.  Rather than be forced to include separate entries for `T_prev_micro_step` and `qv_prev_micro_step` in the initial netCDF file, we want to make it possible to have these two fields simply copy `T_mid` and `qv_mid` at initialization.

The idea is simple, instead of just having the option to assign a field to a constant value, we can also assign a string argument in the `input.yaml` file.  If a string argument is detected then we copy the data for the field matching that string name.

The recipe to accomplish this is as follows:
1. Adjust EKAT so that it is possible to query the "type" of a node,  see: https://github.com/E3SM-Project/EKAT/pull/97
2. if type=double than assign the field a constant value everywhere (already possible)
3. if type=string than copy the data from field="string" to the field in question.

AaronDonahue:
Handled by PR #978

PR #978, opened 2021-04-02
Introduce new option for intialization, allowing a string argument

AaronDonahue:
This PR will make it possible to pass a string as an parameter list value in the control YAML file for any field that needs to be intialized.  Currently, only a double precision value is possible, which causes the field to be initialized uniformly to that value.

Allowing a string to be passed will open up the door to designating another field which can be copied into the specific field in question.

An example would be in initializing the variable `T_prev_micro_step` in P3-Microphysics.  For an initial run this is usually set to equal the initial temperature, `T_mid`.  Currently we would have to store a copy of `T_prev_micro_step` within the netCDF initial file, this PR will make it possible to simply designate `T_mid` as having the same initial values as `T_prev_micro_step`.

It will also allow a user to take a netCDF file with perhaps different names and still populate the appropriate fields in SCREAM.

Addresses #975 

Depends on EKAT PR: https://github.com/E3SM-Project/EKAT/pull/97

### DIFF
--- a/components/scream/src/control/atmosphere_driver.cpp
+++ b/components/scream/src/control/atmosphere_driver.cpp
@@ -196,7 +196,7 @@ initialize_fields (const util::TimeStamp& t0)
     // First, check if the input file contains constant values for some of the fields
     if (ic_pl.isParameter(name)) {
       // The user provided a constant value for this field. Simply use that.
-      initialize_one_field(name, ic_pl);
+      initialize_one_field<double>(name, ic_pl);
     } else {
       // The field does not have a constant value, so we expect to find it in the nc file
       ic_fields.set(ekat::strint("field",ifield+1),name); 
@@ -230,6 +230,7 @@ initialize_fields (const util::TimeStamp& t0)
   m_ad_status |= s_fields_inited;
 }
 
+template<typename T>
 void AtmosphereDriver::initialize_one_field(const std::string& name, const ekat::ParameterList& ic_pl)
 {
   printf("ASD - name = %s\n",name.c_str());  //ASD - DELETE
@@ -243,7 +244,7 @@ void AtmosphereDriver::initialize_one_field(const std::string& name, const ekat:
   if (layout.is_vector_layout()) {
     const auto idim = layout.get_vector_dim();
     const auto vec_dim = layout.dim(idim);
-    const auto& values = ic_pl.get<std::vector<Real>>(name);
+    const auto& values = ic_pl.get<std::vector<T>>(name);
     EKAT_REQUIRE_MSG (values.size()==static_cast<size_t>(vec_dim),
         "Error! Initial condition values array for '" + name + "' has the wrong dimension.\n"
         "       Field dimension: " + std::to_string(vec_dim) + "\n"
@@ -257,7 +258,7 @@ void AtmosphereDriver::initialize_one_field(const std::string& name, const ekat:
       printf("    - [%d] = %f\n",comp,values[comp]);    // ASD-DELETE
     }
   } else {
-    const auto& value = ic_pl.get<double>(name);
+    const auto& value = ic_pl.get<T>(name);
     printf("    - val = %f\n",value);  //ASD - DELETE
     f.set_value(value);
   }
--- a/components/scream/src/control/atmosphere_driver.cpp
+++ b/components/scream/src/control/atmosphere_driver.cpp
@@ -198,11 +198,11 @@ initialize_fields (const util::TimeStamp& t0)
     if (ic_pl.isParameter(name)) {
       // The user provided a constant value for this field. Simply use that.
       if (ic_pl.isType<double>(name) or ic_pl.isType<std::vector<double>>(name)) {
-        initialize_one_field(name, ic_pl);
+        initialize_constant_field(name, ic_pl);
       } else if (ic_pl.isType<std::string>(name)) {
         ic_fields_to_copy.push_back(name);
       } else {
-        EKAT_REQUIRE_MSG (false, "ERROR: invalid assignment for variable " + name + ", only double or string arguments are allowed");
+        EKAT_REQUIRE_MSG (false, "ERROR: invalid assignment for variable " + name + ", only scalar double or string, or vector double arguments are allowed");
       }
     } else {
       // The field does not have a constant value, so we expect to find it in the nc file
@@ -237,7 +237,7 @@ initialize_fields (const util::TimeStamp& t0)
     const auto& name_src = ic_pl.get<std::string>(name_tgt);
     auto f_tgt = m_field_repo->get_field(name_tgt,m_grids_manager->get_reference_grid()->name());
     auto f_src = m_field_repo->get_field(name_src,m_grids_manager->get_reference_grid()->name());
-    f_tgt.set_value(f_src);
+    f_tgt.deep_copy(f_src);
   }
 
   m_current_ts = t0;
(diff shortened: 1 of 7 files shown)


components/scream/src/control/atmosphere_driver.cpp
```diff
@@ -190,35 +190,19 @@ initialize_fields (const util::TimeStamp& t0)
...
-          f_i.set_value(values[comp]);
-        }
+      if (ic_pl.isType<double>(name) or ic_pl.isType<std::vector<double>>(name)) {
+        initialize_one_field(name, ic_pl);
+      } else if (ic_pl.isType<std::string>(name)) {
+        ic_fields_to_copy.push_back(name);
       } else {
-        const auto& value = ic_pl.get<double>(name);
-        f.set_value(value);
+        EKAT_REQUIRE_MSG (false, "ERROR: invalid assignment for variable " + name + ", only double or string arguments are allowed");
```
bartgol, line 205: Technically, argument is double for scalars and vector<double> for vectors. Not sure if this error msg should be fixed.

components/scream/src/share/field/field.hpp
```diff
@@ -148,6 +148,10 @@ class Field {
   template<HostOrDevice HD = Device>
   void set_value (const RT value);
 
+  // Copy the data from one field to this field
+  template<HostOrDevice HD = Device>
+  void set_value (const field_type& field_src);
```
bartgol, line 153: I would change the name of this method. `set_value` makes me think that you set a single value. I would use something like `assign`, or `import`, or 'deep_copy'. Maybe deep_copy is more inline with what we are doing, and it's the same name kokkos uses.

components/scream/src/share/field/field.hpp
```diff
@@ -464,6 +468,62 @@ sync_to_dev () const {
...
+          " to field " + get_header().get_identifier().name() + " during initialization.  Layouts don't match.");
+  const auto  rank = layout.rank();
+  // Note: we can't just do a deep copy on get_view<HD>(), since this
+  //       field might be a subfield of another. So check the header
+  //       first, to see if we have a parent. If not, deep copy is fine.
+  //       If we do, we need to get the reshaped view first.
+  const auto parent = get_header().get_parent();
+  if (parent.lock()==nullptr) {
+    // No parent. Deep copying to get_view<HD>() is safe.
+    Kokkos::deep_copy (get_view<HD>(),field_src.get_view<HD>());
```
bartgol, line 488: What if `field_src` has a parent? This code would crash.

I think we might as well get rid of the parent check, and simply call `get_reshaped_view` on both `*this` and `field_src`.

components/scream/src/share/field/field.hpp
```diff
@@ -464,6 +468,62 @@ sync_to_dev () const {
...
 
+template<typename RealType>
+template<HostOrDevice HD>
+void Field<RealType>::
+set_value (const field_type& field_src) {
+  const auto& layout     = get_header().get_identifier().get_layout();
+  const auto& layout_src = field_src.get_header().get_identifier().get_layout();
+  EKAT_REQUIRE_MSG(layout==layout_src,
+       "ERROR: Unable to copy field " + field_src.get_header().get_identifier().name() + 
+          " to field " + get_header().get_identifier().name() + " during initialization.  Layouts don't match.");
```
bartgol, line 479: Let's not say "during initialization". We don't know if/when other people might be using this method.

components/scream/src/share/field/field.hpp
bartgol: If we go with `deep_copy`, we should also change the name of the method above. Kokkos overloads the `deep_copy` function, offering a copy from view and a copy from scalar value. We would be more aligned with that nomenclature.

Not necessarily in this PR though.

AaronDonahue:
@bartgol , the tests all passed so I think its a good time for review.

components/scream/src/control/tests/ad_tests.cpp
```diff
@@ -31,6 +31,16 @@ TEST_CASE ("group_requirements","[!throws]")
   // Init and run a single time step
   util::TimeStamp init_time(0,0,0,0.0);
   ad.initialize(atm_comm,ad_params,init_time);
+  // Verify that after initialization field "E" matches field "A"
+  auto& field_repo = ad.get_field_repo();
+  const auto& view_A = field_repo.get_field("A","Point Grid A").get_reshaped_view<const Real**, Host>();
+  const auto& view_E = field_repo.get_field("E","Point Grid A").get_reshaped_view<const Real**, Host>();
+  for (int icol=0;icol<num_cols;++icol) {
```
bartgol, line 38: I implemented `views_are_equal` in `field_utils.hpp` to do this kind of checks. But it was before we introduced subfields, so I think that routine needs to be upgraded.

Not for this PR though. But I opened an issue to track this.

components/scream/src/control/atmosphere_driver.cpp
```diff
@@ -190,35 +190,19 @@ initialize_fields (const util::TimeStamp& t0)
...
-            "       Array dimenions: " + std::to_string(values.size()) + "\n");
-
-        // Extract a subfield for each component. This is not "too" expensive, expecially
-        // considering that this code is executed during initialization only.
-        for (int comp=0; comp<vec_dim; ++comp) {
-          auto f_i = f.get_component(comp);
-          f_i.set_value(values[comp]);
-        }
+      if (ic_pl.isType<double>(name) or ic_pl.isType<std::vector<double>>(name)) {
+        initialize_one_field(name, ic_pl);
```
bartgol, line 201: The name could be more specific, namely telling we init with constant value. E.g., `initialize_constant_field(...);`.

bartgol (changes requested):
Mostly good. There's one issue with `set_value`'s impl. The other comments are more subjective maybe, so feel free to push back on some (all?).

components/scream/src/share/field/field.hpp
```diff
@@ -471,63 +471,58 @@ sync_to_dev () const {
...
-set_value (const field_type& field_src) {
+deep_copy (const field_type& field_src) {
   const auto& layout     = get_header().get_identifier().get_layout();
   const auto& layout_src = field_src.get_header().get_identifier().get_layout();
   EKAT_REQUIRE_MSG(layout==layout_src,
        "ERROR: Unable to copy field " + field_src.get_header().get_identifier().name() + 
-          " to field " + get_header().get_identifier().name() + " during initialization.  Layouts don't match.");
+          " to field " + get_header().get_identifier().name() + ".  Layouts don't match.");
   const auto  rank = layout.rank();
   // Note: we can't just do a deep copy on get_view<HD>(), since this
```
bartgol, line 481: This comment has to be updated now. We could leave just the 1st sentence, and remove the other ones.

components/scream/src/share/field/field.hpp
```diff
@@ -471,63 +471,58 @@ sync_to_dev () const {
...
   //       field might be a subfield of another. So check the header
   //       first, to see if we have a parent. If not, deep copy is fine.
   //       If we do, we need to get the reshaped view first.
   const auto parent = get_header().get_parent();
```
bartgol, line 485: I think this var is now unused, and can be removed.

components/scream/src/share/field/field.hpp
```diff
@@ -471,63 +471,58 @@ sync_to_dev () const {
...
+      }
+      break;
+    case 4:
+      {
+        auto v     = get_reshaped_view<RT****,HD>();
+        auto v_src = field_src.get_reshaped_view<RT****,HD>();
+        Kokkos::deep_copy(v,v_src);
+      }
+      break;
+    default:
```
bartgol, line 517: Would you mind adding also the rank-5 case? The rank-4 is probably unused on physics grid, and meant for dyn grid. But a 3d vector quantity on dyn grid would have rank 5 (EL,CMP,GP,GP,LEV). Since it's almost the same code, we might as well add it.

bartgol (changes requested):

components/scream/src/share/field/field.hpp
AaronDonahue: oh yeah, of course it does.  Sorry for the lazy "response to reviewer comments" commit.

components/scream/src/share/field/field.hpp
AaronDonahue: done.

bartgol (approved):

PR #980, opened 2021-04-05
SCREAM: fix call to mpi_abort in scream_scorpio_interface.F90

### LINKED ISSUE #979: Wrong variable name in shoc standalone input file
This is causing master failures. I'm not sure how it got past the AT.

See comment at bottom of #973 for details.

bartgol:
Fixes #979 

Note: merging this will cause shoc standalone test to fail. Should that failure be fixed in this PR? I say no. Let's merge this, and then turn the AT back on, so we make all other open PRs fail again.

Also, I'm skipping testing for this PR, since the shoc test will fail.

### DIFF
--- a/components/scream/src/share/io/scream_scorpio_interface.F90
+++ b/components/scream/src/share/io/scream_scorpio_interface.F90
@@ -725,13 +725,14 @@ contains
     integer,           intent(in)    :: retVal
 
     type(pio_file_list_t), pointer :: curr => NULL()
+    integer :: ierr
 
     if (retVal .ne. PIO_NOERR) then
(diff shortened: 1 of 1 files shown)


tcclevenger (approved):
This is just going to be manually merged?

bartgol:
Yeah, or we could put all other PRs on hold, and turn the AT back on. But it should skip testing anyways, so the only difference would be a green check mark instead of a yellow dot next to the PR title on github... I'm ok with manual merge now. What do you think?

tcclevenger:
I'm good with just merging..

PR #981, opened 2021-04-05
Fixes valgrind errors in shoc_uts and p3_standalone

jgfouca:
Also some minor whitespace and comment fixes.

With this PR, all valgrind tests should pass on cdash tomorrow.

### DIFF
--- a/components/scream/src/physics/shoc/tests/shoc_pblintd_height_tests.cpp
+++ b/components/scream/src/physics/shoc/tests/shoc_pblintd_height_tests.cpp
@@ -102,7 +102,7 @@ struct UnitWrap::UnitTest<D>::TestPblintdHeight {
 
     // Save result for checking with next test
     Real pblh_save[shcol];
-    for(Int s = 0; s < shcol-1; ++s) {
+    for(Int s = 0; s < shcol; ++s) {
       pblh_save[s] = SDS.pblh[s];
     }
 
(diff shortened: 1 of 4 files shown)


components/scream/src/physics/shoc/tests/shoc_pblintd_height_tests.cpp
```diff
@@ -102,7 +102,7 @@ struct UnitWrap::UnitTest<D>::TestPblintdHeight {
 
     // Save result for checking with next test
     Real pblh_save[shcol];
-    for(Int s = 0; s < shcol-1; ++s) {
+    for(Int s = 0; s < shcol; ++s) {
```
jgfouca, line 105: @bogensch , I think this was a simple copy/paste error and you did not actually intend to cut this loop off 1 iteration early.

bartgol (approved):

components/scream/src/physics/shoc/tests/shoc_pblintd_height_tests.cpp
bogensch: Good catch Jim, yes this was a copy/paste error.

bogensch (approved):

bartgol:
The failures are b/c of the IC nc files (see #979 and #980). We fixed the IC files as part of #974, which is about to go in. Once it goes in, tests should pass.

Issue #955, opened 2021-03-31
Compute dry static energy in SHOC from T and z


PeterCaldwell:
v1 SHOC currently assumes "s" (dry static energy) is passed as an input. It should instead be computed in the SHOC interface via DSE = cp*T_atm_mid + g*(z_mid + surf_geopotential). Inclusion of surf_geopotential is needed because z_mid is actually a relative geopotential assuming the surface height is 0. 

It might make sense for this DSE calculation to be a universal function rather than stuffed into SHOC itself in case other future parameterizations want DSE.

tcclevenger:
See PR https://github.com/E3SM-Project/scream/pull/983

PR #983, opened 2021-04-06
add get_dse function

tcclevenger:
Compute dry static energy (DSE) instead of requiring as an input in the field manager. Function is added to `physics/share/physics_functions.hpp`.

Addresses Issue https://github.com/E3SM-Project/scream/issues/955.

### DIFF
--- a/components/scream/src/physics/share/physics_universal_impl.hpp
+++ b/components/scream/src/physics/share/physics_universal_impl.hpp
@@ -106,7 +106,30 @@ Functions<S,D>::get_dz(const Spack& zi_top, const Spack& zi_bot, const Smask& ra
   return result;
 }
 //-----------------------------------------------------------------------------------------------//
+// Compute dry static energy (DSE).
+// The result unit is in J/kg
+// The inputs are
+//   T_mid is the atmospheric temperature. Units in K.
+//   z_mid is the geopotential height above surface at midpoints. Units in m.
+//   surf_geopotential is the surface geopotential height. Units in m.
+template <typename S, typename D>
+KOKKOS_FUNCTION
+typename Functions<S,D>::Spack
+Functions<S,D>::get_dse(const Spack& T_mid, const Spack& z_mid, const Real surf_geopotential, const Smask& range_mask)
+{
+  static constexpr Scalar cp  = C::CP;
+  static constexpr Scalar ggr = C::gravit;
 
+  Spack result;
+  const Spack dse = cp*T_mid + ggr*z_mid + surf_geopotential;
+  // Check that there are no obvious errors in the result.
+  EKAT_KERNEL_ASSERT_MSG(!((isnan(dse) && range_mask).any()), "Error in get_dse, dse has NaN values.\n"); // exit with an error message
+  EKAT_KERNEL_ASSERT_MSG(!(((dse <= 0) && range_mask).any()), "Error in get_dse, dse has negative values.\n"); // exit with an error message
+  // Set the values of the result
+  result.set(range_mask,dse);
+  return result;
+}
+//-----------------------------------------------------------------------------------------------//
 
 } // namespace physics
 } // namespace scream
--- a/components/scream/src/physics/share/physics_universal_impl.hpp
+++ b/components/scream/src/physics/share/physics_universal_impl.hpp
@@ -125,7 +125,7 @@ Functions<S,D>::get_dse(const Spack& T_mid, const Spack& z_mid, const Real surf_
   // Check that there are no obvious errors in the result.
   EKAT_KERNEL_ASSERT_MSG((isnan(T_mid) && range_mask).none(), "Error in get_dse, T_mid has NaN values.\n"); // exit with an error message
   EKAT_KERNEL_ASSERT_MSG((isnan(z_mid) && range_mask).none(), "Error in get_dse, z_mid has NaN values.\n"); // exit with an error message
-  EKAT_KERNEL_ASSERT_MSG((isnan(surf_geopotential) && range_mask).none(), "Error in get_dse, surf_geopotential has NaN values.\n"); // exit with an error message
+  EKAT_KERNEL_ASSERT_MSG(!isnan(surf_geopotential), "Error in get_dse, surf_geopotential has NaN values.\n"); // exit with an error message
   EKAT_KERNEL_ASSERT_MSG(((dse <= 0) && range_mask).none(), "Error in get_dse, dse has negative values.\n"); // exit with an error message
 
   return dse;
--- a/components/scream/src/physics/share/physics_universal_impl.hpp
+++ b/components/scream/src/physics/share/physics_universal_impl.hpp
@@ -120,14 +120,15 @@ Functions<S,D>::get_dse(const Spack& T_mid, const Spack& z_mid, const Real surf_
   static constexpr Scalar cp  = C::CP;
   static constexpr Scalar ggr = C::gravit;
 
-  Spack result;
-  const Spack dse = cp*T_mid + ggr*z_mid + surf_geopotential;
+  const Spack dse(range_mask, cp*T_mid + ggr*z_mid + surf_geopotential);
+
   // Check that there are no obvious errors in the result.
-  EKAT_KERNEL_ASSERT_MSG(!((isnan(dse) && range_mask).any()), "Error in get_dse, dse has NaN values.\n"); // exit with an error message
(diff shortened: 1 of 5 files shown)


components/scream/src/physics/share/tests/physics_universal_unit_tests.cpp
```diff
@@ -181,6 +214,11 @@ struct UnitWrap::UnitTest<D>::TestUniversal
         Real zi_bot = (pow(k,2)+k)/2.0;
         Real zi_top = zi_bot + (k+1); 
         dz_tests(zi_top,zi_bot,errors);
+        // DSE Test
+        Real temp = k+1;
+        Real height = k+1;
+        Real surface_height = k+1;
+        dse_tests(temp,height,surface_height,errors);
```
tcclevenger, line 221: Not sure what values make sense for the test.. They just need to be >0 to pass all checks..

components/scream/src/physics/share/physics_universal_impl.hpp
```diff
@@ -106,7 +106,30 @@ Functions<S,D>::get_dz(const Spack& zi_top, const Spack& zi_bot, const Smask& ra
...
+typename Functions<S,D>::Spack
+Functions<S,D>::get_dse(const Spack& T_mid, const Spack& z_mid, const Real surf_geopotential, const Smask& range_mask)
+{
+  static constexpr Scalar cp  = C::CP;
+  static constexpr Scalar ggr = C::gravit;
 
+  Spack result;
+  const Spack dse = cp*T_mid + ggr*z_mid + surf_geopotential;
+  // Check that there are no obvious errors in the result.
+  EKAT_KERNEL_ASSERT_MSG(!((isnan(dse) && range_mask).any()), "Error in get_dse, dse has NaN values.\n"); // exit with an error message
```
bartgol, line 126: I think it makes more sense to check the inputs for any nan, no? If this throws, we still don't know what caused it. We might instead check T_mid, z_mid, and the surf geopot, and print exactly which var is nan.

This is only done in dbg mode anyways.

components/scream/src/physics/share/physics_universal_impl.hpp
```diff
@@ -106,7 +106,30 @@ Functions<S,D>::get_dz(const Spack& zi_top, const Spack& zi_bot, const Smask& ra
...
+  EKAT_KERNEL_ASSERT_MSG(!(((dse <= 0) && range_mask).any()), "Error in get_dse, dse has negative values.\n"); // exit with an error message
```
bartgol, line 127: Mask has the `none()` method, which is probably what you want here. Its impl is precisely `!any()`, but its probably easier to read this check if you use it.

components/scream/src/physics/share/physics_universal_impl.hpp
```diff
@@ -106,7 +106,30 @@ Functions<S,D>::get_dz(const Spack& zi_top, const Spack& zi_bot, const Smask& ra
...
+template <typename S, typename D>
+KOKKOS_FUNCTION
+typename Functions<S,D>::Spack
+Functions<S,D>::get_dse(const Spack& T_mid, const Spack& z_mid, const Real surf_geopotential, const Smask& range_mask)
+{
+  static constexpr Scalar cp  = C::CP;
+  static constexpr Scalar ggr = C::gravit;
 
+  Spack result;
+  const Spack dse = cp*T_mid + ggr*z_mid + surf_geopotential;
```
bartgol, line 124: Why this temporary? Why not doing right away `result = cp*T_mid...`?

components/scream/src/physics/shoc/atmosphere_macrophysics.cpp
```diff
@@ -141,11 +139,11 @@ void SHOCMacrophysics::initialize_impl (const util::TimeStamp& t0)
   auto qc               = m_shoc_fields_out["qc"].get_reshaped_view<Spack**>();
   auto qv               = m_shoc_fields_out["qv"].get_reshaped_view<Spack**>();
   auto tke              = m_shoc_fields_out["tke"].get_reshaped_view<Spack**>();
-  auto s                = m_shoc_fields_out["s"].get_reshaped_view<Spack**>();
   auto Q                = m_shoc_fields_out["Q"].get_reshaped_view<Spack***>();
   auto cldfrac_liq      = m_shoc_fields_out["cldfrac_liq"].get_reshaped_view<Spack**>();
   auto sgs_buoy_flux    = m_shoc_fields_out["sgs_buoy_flux"].get_reshaped_view<Spack**>();
   auto inv_qc_relvar    = m_shoc_fields_out["inv_qc_relvar"].get_reshaped_view<Spack**>();
+  auto phis             = m_shoc_fields_in["phis"].get_reshaped_view<const Pack1d*>();
```
bartgol, line 146: This might be a bit more general than just this PR: I noticed shoc uses Pack1d=Pack<Scalar,1>; What's the point in using this vs just Scalar?

components/scream/src/physics/share/tests/physics_universal_unit_tests.cpp
bartgol: Why not using random numbers in something like [0.001,1000.0]? You can use `ekat::genRandArray(&temp,1,engine,pdf)`. See, e.g., [field tests](https://github.com/E3SM-Project/scream/blob/440e098e9a16ee5feb41a68c56fee69d9f6dfecc/components/scream/src/share/tests/field_tests.cpp#L493), which also shows how to give the ability to pass the rng seed to the test exec (in case you're debugging and want to match someone else's buggy numbers).

bartgol (changes requested):
Looks good. I have a couple of suggestions, but feel free to push back.

components/scream/src/physics/shoc/atmosphere_macrophysics.cpp
tcclevenger: Yes, this is all over SHOC. Seemed strange to me when I started working on shoc conversion, but I never asked about it.. I can create a separate PR changing these in shoc.

components/scream/src/physics/share/physics_universal_impl.hpp
tcclevenger: Goop point, I simplified this.

bartgol (approved):

bartgol (approved):

components/scream/src/physics/share/physics_universal_impl.hpp
```diff
@@ -106,7 +106,31 @@ Functions<S,D>::get_dz(const Spack& zi_top, const Spack& zi_bot, const Smask& ra
...
+//   surf_geopotential is the surface geopotential height. Units in m.
+template <typename S, typename D>
+KOKKOS_FUNCTION
+typename Functions<S,D>::Spack
+Functions<S,D>::get_dse(const Spack& T_mid, const Spack& z_mid, const Real surf_geopotential, const Smask& range_mask)
+{
+  static constexpr Scalar cp  = C::CP;
+  static constexpr Scalar ggr = C::gravit;
+
+  const Spack dse(range_mask, cp*T_mid + ggr*z_mid + surf_geopotential);
```
tcclevenger, line 123: @PeterCaldwell On the Field manager variable page I see two versions of DSE:
```
dse = cp*T_mid + g*(z_mid+surf_geopotential) (in the table)
```
and, from a Mark Taylor comment on that page, 
```
state%s(:ncol,k) = state%t(:ncol,k  )*cpairv_loc(:ncol,k,state%lchnk)&
                       + gravit*state%zm(:ncol,k) + state%phis(:ncol) 
```
Not sure the source of the latter..

Do you know which is correct? `dse = cp*T_mid + g*(z_mid + surf_geopotential) ` or
`dse = cp*T_mid + g*z_mid + surf_geopotential`

PeterCaldwell: @tcclevenger - so the question is whether to multiply grav by phis or not? The answer to this depends on whether phis is geopotential HEIGHT (which I was assuming and should have said in my impl) or actual geopotential (which it appears Mark was assuming). To answer this, we need to check how phis is used elsewhere in the code.

PeterCaldwell: check shco_update_host_dse_impl.hpp. It says `host_dse(k) = cp*temp+ggr*zt_grid(k)+phis` so Mark's version is right.

tcclevenger: Ok, awesome. Thanks!

PR #984, opened 2021-04-06
update pd remapper

tcclevenger:
For remap_bwd, I use a `TeamPolicy` over `ncols*nfields`. This speeds up things _slightly_. For remap_fwd I keep it the same since changing to `TeamPolicy` over `ncols*nfields` slows things down because the `set_dyn_to_zero()` call must be inside a `TeamPolicy` over `nfields`.

I also created a `reshape` function for views (as suggested in https://github.com/E3SM-Project/scream/pull/934)

### DIFF
--- a/components/scream/src/dynamics/homme/physics_dynamics_remapper.hpp
+++ b/components/scream/src/dynamics/homme/physics_dynamics_remapper.hpp
@@ -179,42 +179,31 @@ public:
 
     switch (dims.size) {
       case 1:
-      {
         return uview_nd(reinterpret_cast<ScalarT*>(p.get()),
                         dims.dims[0]);
-      }
       case 2:
-      {
         return uview_nd(reinterpret_cast<ScalarT*>(p.get()),
                         dims.dims[0],
                         dims.dims[1]);
-      }
       case 3:
-      {
         return uview_nd(reinterpret_cast<ScalarT*>(p.get()),
                         dims.dims[0],
                         dims.dims[1],
                         dims.dims[2]);
-      }
       case 4:
-      {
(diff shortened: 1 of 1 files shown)


components/scream/src/dynamics/homme/physics_dynamics_remapper.hpp
```diff
@@ -166,13 +163,71 @@ class PhysicsDynamicsRemapper : public AbstractRemapper<RealType>
...
+
+  template<typename ScalarT, int N>
+  KOKKOS_FUNCTION
+  Unmanaged<typename KokkosTypes<device_type>::template view_ND<ScalarT,N>>
+  reshape (Pointer& p, const Dims& dims) const {
+    using uview_nd = Unmanaged<typename KokkosTypes<device_type>::template view_ND<ScalarT,N>>;
+
+    switch (dims.size) {
+      case 1:
+      {
```
bartgol, line 182: MInor: you only need braces inside a `case` if you declare local variables, otherwise you don't. No braces would get rid of a few lines (and help fit more code in one screen).

components/scream/src/dynamics/homme/physics_dynamics_remapper.hpp
```diff
@@ -460,24 +515,21 @@ initialize_device_variables()
 }
 
 template<typename RealType>
-template <typename ScalarT, typename MT>
+ template <typename ScalarT, typename MT>
```
bartgol, line 518: There seems to be an extra space here.

bartgol (approved):

tcclevenger:
@bartgol Thanks! I'll need your approval again..

bartgol (approved):

Issue #966, opened 2021-03-31
Decide what optical property scheme we want to use and how to make that scheme consistent with P3 assumptions


PeterCaldwell:
lamc and mu_c are parameters for the gamma PDF for subgrid variability in qc. They were used in older versions of radiation's subcolumn generator but aren't part of the current RRTMGP setup. This isn't necessarily a good idea, but is a topic for a longer-term discussion. For our initial SCREAMv1 impl, we can delete lamc and mu_c from the p3_main interface since they aren't used outside P3. These variables are used a lot in P3 so we can't delete them completely.

AaronDonahue:
Should already be addressed by #987

Issue #959, opened 2021-03-31
Remove nevapr, precip_total_tend, and qr_evap_tend from P3 interface


PeterCaldwell:
nevapr, precip_total_tend, and qr_evap_tend were needed by MAM aerosols... which we're not porting to C++. In addition, these names don't really make sense or describe what they actually did. And the way they were implemented seems wrong. So we should just delete all mention of them.

AaronDonahue:
@PeterCaldwell , Should we remove these variables from all of P3?  Including the F90, SCREAMv0 code?

PeterCaldwell:
good question. I'm inclined to say we should follow Conrad's lead and skip F90 changes just out of laziness. Also, I think Koby wants these things in F90 for low res... not that he's using our code for that, but someday he might.

AaronDonahue:
okay cool, sounds good.

AaronDonahue:
Solved by #987

PR #987, opened 2021-04-07
Update the p3 pre- and post- ambles to facilitate coupling with other processes

AaronDonahue:
This PR updates the pre- and post- ambles in P3 microphysics so that variables expected by other atmosphere processes are in the appropriate format. 

These changes also reflect decisions made about which variables will be carried in the field manager, and which variables will be computed on the fly in the pre-amble of a process.

Finally this PR will remove unused variables from P3, addressing issue #959 and #966.  Note, we have decided to keep these variables in the F90 code, as a result, this PR introduces a structure to the P3 interface that will hold "deprecated" variables that are no longer used in the cpp implementation but must be kept as long as our testing call F90 code, or until these variables are removed from the F90 code.

### DIFF
--- a/components/scream/src/physics/p3/atmosphere_microphysics.cpp
+++ b/components/scream/src/physics/p3/atmosphere_microphysics.cpp
@@ -100,12 +100,6 @@ void P3Microphysics::set_grids(const std::shared_ptr<const GridsManager> grids_m
   m_computed_fields.emplace("micro_liq_ice_exchange", scalar3d_layout_mid, nondim, grid_name);
   m_computed_fields.emplace("micro_vap_liq_exchange", scalar3d_layout_mid, nondim, grid_name);
   m_computed_fields.emplace("micro_vap_ice_exchange", scalar3d_layout_mid, nondim, grid_name);
-  // Deprecated: (these should be removed from the AD eventually)
-//ASD  m_computed_fields.emplace("precip_total_tend", scalar3d_layout_mid, mm,     grid_name);
-//ASD  m_computed_fields.emplace("qr_evap_tend",      scalar3d_layout_mid, mm/s,   grid_name);
-//ASD  m_computed_fields.emplace("nevapr",            scalar3d_layout_mid, nondim, grid_name);
-//ASD  m_computed_fields.emplace("mu_qc",             scalar3d_layout_mid, nondim, grid_name);
-//ASD  m_computed_fields.emplace("lambda_qc",         scalar3d_layout_mid, nondim, grid_name);
 
 }
 
@@ -194,11 +188,6 @@ void P3Microphysics::initialize_impl (const util::TimeStamp& t0)
   history_only.vap_liq_exchange = m_p3_fields_out["micro_vap_liq_exchange"].get_reshaped_view<Pack**>();
   history_only.vap_ice_exchange = m_p3_fields_out["micro_vap_ice_exchange"].get_reshaped_view<Pack**>();
   // Deprecated -- These are fields actively being deleted, but are still needed for F90 BFB tests.
-//ASD  diag_outputs.nevapr           = m_p3_fields_out["nevapr"].get_reshaped_view<Pack**>();
-//ASD  diag_outputs.qr_evap_tend = m_p3_fields_out["qr_evap_tend"].get_reshaped_view<Pack**>();
-//ASD  diag_outputs.precip_total_tend  = m_p3_fields_out["precip_total_tend"].get_reshaped_view<Pack**>();
-//ASD  diag_outputs.mu_c               = m_p3_fields_out["mu_qc"].get_reshaped_view<Pack**>();
-//ASD  diag_outputs.lamc               = m_p3_fields_out["lambda_qc"].get_reshaped_view<Pack**>();
   view_2d nevapr("nevapr",m_num_cols,nk_pack);
   view_2d qr_evap_tend("qr_evap_tend",m_num_cols,nk_pack);
   view_2d precip_total_tend("precip_total_tend",m_num_cols,nk_pack);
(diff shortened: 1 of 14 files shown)


components/scream/src/physics/p3/atmosphere_microphysics.cpp
```diff
@@ -240,7 +242,7 @@ void P3Microphysics::run_impl (const Real dt)
 
   // Run p3 main
   P3F::p3_main(prog_state, diag_inputs, diag_outputs, infrastructure,
-                                       history_only, m_num_cols, m_num_levs);
+                                       history_only, deprecated, m_num_cols, m_num_levs);
```
PeterCaldwell, line 245: If I understand correctly, you're still outputting these variables we don't want. Do we really need to do this? I thought Conrad's strategy was to still produce these variables in F90, but not to actually compare them against anything anymore... so the C++ wouldn't need to carry them anymore. Am I missing something?

components/scream/src/physics/p3/p3_functions_f90.cpp
```diff
@@ -3265,15 +3263,15 @@ Int p3_main_f(
   std::vector<const Real*> ptr_array = {
     pres, dz, nc_nuceat_tend, nccn_prescribed, ni_activated, dpres, exner, cld_frac_i, cld_frac_l, cld_frac_r, inv_qc_relvar,
     qc, nc, qr, nr, qi, qm, ni, bm, qv, th_atm, qv_prev, t_prev, diag_eff_radius_qc, diag_eff_radius_qi,
-    rho_qi, mu_c, lamc, qv2qi_depos_tend, precip_total_tend, nevapr, qr_evap_tend, liq_ice_exchange,
-    vap_liq_exchange, vap_ice_exchange, precip_liq_flux, precip_ice_flux, precip_liq_surf, precip_ice_surf
+    rho_qi, qv2qi_depos_tend,
+    liq_ice_exchange, vap_liq_exchange, vap_ice_exchange, precip_liq_flux, precip_ice_flux, precip_liq_surf, precip_ice_surf
   };
 
   //PMC - hardcoding the index for each variable is very brittle :-(.
```
AaronDonahue, line 3270: @PeterCaldwell , the following changes address this brittle-ness, I think.  I can delete this comment from you if you think that it isn't brittle anymore.

AaronDonahue:
Since this PR removes variables from the p3-output, I think it is expected that `p3_run_and_cmp` will fail when compared with baselines from master.

AaronDonahue:
@PeterCaldwell , when you have a chance could you take one more look at this PR to verify that it has addressed your questions?  I also had one question for you about a comment of yours in the code that I might be able to delete now, see above.

components/scream/src/physics/p3/p3_functions_f90.cpp
PeterCaldwell: Looks good.

components/scream/src/physics/p3/p3_iso_c.f90
```diff
@@ -143,6 +139,10 @@ subroutine p3_main_c(qc,nc,qr,nr,th_atm,qv,dt,qi,qm,ni,bm,   &
 
     real(kind=c_real), dimension(its:ite,kts:kte,49)   :: p3_tend_out
     real(kind=c_real), dimension(its:ite,3) :: col_location
+    real(kind=c_real), dimension(its:ite,kts:kte)      :: mu_c, lamc
+    real(kind=c_real), dimension(its:ite,kts:kte)      :: precip_total_tend
+    real(kind=c_real), dimension(its:ite,kts:kte)      :: nevapr
```
PeterCaldwell, line 144: Isn't nevapr supposed to be deleted everywhere?

AaronDonahue: oh yes, I think this PR was opened before the cleanup PR went in.  I can clean that up

AaronDonahue: it isn't deleted in the F90 code, right?  At least not in this PR.  I can go ahead and make that change too if you would like.

PeterCaldwell (approved):
Looks good, thanks Aaron!

AaronDonahue:
I had to rebase this branch because of conflicts with master.  I'll let the pre-merge checks go through and then go ahead and manually merge once I verify that the p3_run_and_cmp tests are still the only ones that fail.

AaronDonahue:
update: they didn't pop up as conflicts so I didn't see them until now, but there was a change to how we add fields in a processes that caused the build failures.  I'll fix this now.

PeterCaldwell (approved):

AaronDonahue:
@bartgol , I'm not sure why the pre-merge inspection is still pending for this PR.  Peter definitely reapproved it as the last action on the PR so presumably this check should be done.  I can't merge the pull request until it is.  Am I missing something?

bartgol:
> @bartgol , I'm not sure why the pre-merge inspection is still pending for this PR. Peter definitely reapproved it as the last action on the PR so presumably this check should be done. I can't merge the pull request until it is. Am I missing something?

The pre-merge inspection is performed _after_ testing passed. As opposed to the pre-test inspection (which is usually skipped since we're part of the scream team), which happens before testing.

Once tests pass, the AT will do the pre-merge inspection, which amounts to check whether there is a review approval for the current tip of the branch _and_ no reviews that requested changes.

AaronDonahue:
ahhh, ok.  Sorry I misunderstood.  Tests won't pass.  I'm guessing I don't have the ability to override the AT and merge a PR.  Are you able to do that @bartgol?

Test fail explanation: This PR changes the arguments to P3 by removing a few, we expect p3_run_and_cmp to fail, which it does.

PR #990, opened 2021-04-09
Remove VAR field tag

### LINKED ISSUE #952: Have just one FieldTag for components of vector fields
Currently, we have VAR (for Variable) and CMP (for Component). When I added them, I was thinking VAR is more appropriate for arrays that are bundling together unrelated variables (e.g., tracers), while CMP was for component of "physically" vector fields.

However, having two tags can be confusing, and could lead to someone using, say, (COL,VAR,LEV) for horizontal velocity or (COL,CMP,LEV) for tracers, which would cause dag issues.

I think the most general tag is CMP, so I would keep that one, and remove VAR.

jgfouca:
Fixes #952

### DIFF
--- a/components/scream/src/share/field/field_layout.cpp
+++ b/components/scream/src/share/field/field_layout.cpp
@@ -116,14 +116,14 @@ LayoutType get_layout_type (const std::vector<FieldTag>& field_tags) {
       if ( (tags[1]==LEV || tags[1]==ILEV) && (tags[0]==CMP || tags[0]==TL)) {
         result = LayoutType::Vector3D;
       } else if ((tags[0]==CMP1 && tags[1]==CMP2) ||
-                 (tags[1]==CMP && tags[0]==TL)) {
+                 (tags[0]==TL   && tags[1]==CMP )) {
         result = LayoutType::Tensor2D;
       }
       break;
     case 3:
       // The only scenarios are:
       //  1) <CMP1, CMP2, LEV|ILEV>
-      //  3) <TL,  CMP, LEV|ILEV>
+      //  2) <TL,  CMP, LEV|ILEV>
       if (tags[2]==LEV || tags[2]==ILEV) {
         if ((tags[0]==CMP1 && tags[1]==CMP2) ||
             (tags[0]==TL && tags[1]==CMP)) {
@@ -131,7 +131,7 @@ LayoutType get_layout_type (const std::vector<FieldTag>& field_tags) {
         }
       }
   }
-  
+
   return result;
 }
 
--- a/components/scream/src/share/field/field_layout.cpp
+++ b/components/scream/src/share/field/field_layout.cpp
@@ -112,25 +112,21 @@ LayoutType get_layout_type (const std::vector<FieldTag>& field_tags) {
       // Possible scenarios:
       //  1) <CMP|TL,LEV|ILEV>
       //  2) <TL,CMP>
-      //  3) <TL>
-      //  4) <CMP1,CMP2>
+      //  3) <CMP1,CMP2>
       if ( (tags[1]==LEV || tags[1]==ILEV) && (tags[0]==CMP || tags[0]==TL)) {
         result = LayoutType::Vector3D;
       } else if ((tags[0]==CMP1 && tags[1]==CMP2) ||
-                 (tags[1]==CMP && tags[0]==TL) ||
-                  tags[0]==TL) {
+                 (tags[1]==CMP && tags[0]==TL)) {
         result = LayoutType::Tensor2D;
       }
       break;
     case 3:
       // The only scenarios are:
       //  1) <CMP1, CMP2, LEV|ILEV>
-      //  2) <TL,  LEV|ILEV>
       //  3) <TL,  CMP, LEV|ILEV>
       if (tags[2]==LEV || tags[2]==ILEV) {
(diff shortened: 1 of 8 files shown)


components/scream/src/share/field/field_layout.cpp
```diff
@@ -103,36 +101,36 @@ LayoutType get_layout_type (const std::vector<FieldTag>& field_tags) {
...
       }
       break;
     case 2:
       // Possible scenarios:
-      //  1) <CMP|VAR|TL,LEV|ILEV>
-      //  2) <VAR|TL,CMP>
-      //  3) <TL,VAR>
+      //  1) <CMP|TL,LEV|ILEV>
+      //  2) <TL,CMP>
+      //  3) <TL>
```
bartgol, line 115: Actually, option 3 goes away completely.

components/scream/src/share/field/field_layout.cpp
```diff
@@ -103,36 +101,36 @@ LayoutType get_layout_type (const std::vector<FieldTag>& field_tags) {
...
       //  4) <CMP1,CMP2>
-      if ( (tags[1]==LEV || tags[1]==ILEV) && (tags[0]==CMP || tags[0]==VAR || tags[0]==TL)) {
+      if ( (tags[1]==LEV || tags[1]==ILEV) && (tags[0]==CMP || tags[0]==TL)) {
         result = LayoutType::Vector3D;
       } else if ((tags[0]==CMP1 && tags[1]==CMP2) ||
-                 (tags[1]==CMP && (tags[0]==VAR || tags[0]==TL)) ||
-                 (tags[0]==TL && tags[1]==VAR)) {
+                 (tags[1]==CMP && tags[0]==TL) ||
+                  tags[0]==TL) {
```
bartgol, line 121: To be a 2d tensor, there should be 2 tags left, so tags[0]==TL alone is wrong.

As stated above, option 3 goes away, so this last condition should be removed.

components/scream/src/share/field/field_layout.cpp
```diff
@@ -103,36 +101,36 @@ LayoutType get_layout_type (const std::vector<FieldTag>& field_tags) {
...
         result = LayoutType::Tensor2D;
       }
       break;
     case 3:
       // The only scenarios are:
       //  1) <CMP1, CMP2, LEV|ILEV>
-      //  2) <TL,  VAR, LEV|ILEV>
-      //  3) <TL|VAR,  CMP, LEV|ILEV>
+      //  2) <TL,  LEV|ILEV>
```
bartgol, line 128: Another case where a case goes away completely: <TL, LEV|ILEV> is not a rank-3 layout, so it should go away.

components/scream/src/share/field/field_layout.cpp
```diff
@@ -103,36 +101,36 @@ LayoutType get_layout_type (const std::vector<FieldTag>& field_tags) {
...
+      //  3) <TL,  CMP, LEV|ILEV>
       if (tags[2]==LEV || tags[2]==ILEV) {
         if ((tags[0]==CMP1 && tags[1]==CMP2) ||
-            (tags[0]==TL && (tags[1]==VAR || tags[1]==CMP)) ||
-            (tags[0]==VAR && tags[1]==CMP)) {
+            (tags[0]==TL && tags[1]==CMP) ||
+            tags[1]==CMP) {
```
bartgol, line 133: And here, the last condition has to be removed.

bartgol (changes requested):
Looks good. There is only a couple of conditions to be fixed in the `get_layout_type` routine.

jgfouca:
@bartgol , done

components/scream/src/share/field/field_layout.cpp
```diff
@@ -112,25 +112,21 @@ LayoutType get_layout_type (const std::vector<FieldTag>& field_tags) {
...
       //  2) <TL,CMP>
-      //  3) <TL>
-      //  4) <CMP1,CMP2>
+      //  3) <CMP1,CMP2>
       if ( (tags[1]==LEV || tags[1]==ILEV) && (tags[0]==CMP || tags[0]==TL)) {
         result = LayoutType::Vector3D;
       } else if ((tags[0]==CMP1 && tags[1]==CMP2) ||
-                 (tags[1]==CMP && tags[0]==TL) ||
-                  tags[0]==TL) {
+                 (tags[1]==CMP && tags[0]==TL)) {
```
bartgol, line 119: Would you mind swapping the two checks in the second one? Meaning, check tags[0] and then tags[1]? It might be a tad confusing to read them bwd. I am not sure why I wrote it that way in the first place.

components/scream/src/share/field/field_layout.cpp
```diff
@@ -112,25 +112,21 @@ LayoutType get_layout_type (const std::vector<FieldTag>& field_tags) {
...
         result = LayoutType::Tensor2D;
       }
       break;
     case 3:
       // The only scenarios are:
       //  1) <CMP1, CMP2, LEV|ILEV>
-      //  2) <TL,  LEV|ILEV>
       //  3) <TL,  CMP, LEV|ILEV>
```
bartgol, line 126: Would you mind also changing `3)` to `2)`, to avoid ppl wondering if we're missing something?

jgfouca:
@bartgol  done

bartgol (approved):
Thanks!

PR #992, opened 2021-04-09
Greatly reduce warnings when building SCREAM

jgfouca:
Down from many hundreds to 3.

rrtmgp needed the most work.

### DIFF
--- a/components/scream/cmake/tpls/Scorpio.cmake
+++ b/components/scream/cmake/tpls/Scorpio.cmake
@@ -56,10 +56,8 @@ macro (CreateScorpioTarget CREATE_FLIB)
       # Not a CIME build. Add scorpio as a subdir
       add_subdirectory (${E3SM_EXTERNALS_DIR}/scorpio ${CMAKE_BINARY_DIR}/externals/scorpio)
       EkatDisableAllWarning(pioc)
+      EkatDisableAllWarning(piof)
       EkatDisableAllWarning(gptl)
-      if (CREATE_FLIB)
-        EkatDisableAllWarning(pioc)
-      endif()
     endif()
   endif ()
 
--- a/components/scream/cmake/tpls/Scorpio.cmake
+++ b/components/scream/cmake/tpls/Scorpio.cmake
@@ -56,8 +56,12 @@ macro (CreateScorpioTarget CREATE_FLIB)
       # Not a CIME build. Add scorpio as a subdir
       add_subdirectory (${E3SM_EXTERNALS_DIR}/scorpio ${CMAKE_BINARY_DIR}/externals/scorpio)
       EkatDisableAllWarning(pioc)
-      EkatDisableAllWarning(piof)
       EkatDisableAllWarning(gptl)
+      # target piof is not always created by scorpio depending on CMake settings
+      # if it is created, we don't want to see warnings from it when building SCREAM
+      if (TARGET piof)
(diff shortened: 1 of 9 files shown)


components/scream/cmake/tpls/Scorpio.cmake
```diff
@@ -56,10 +56,8 @@ macro (CreateScorpioTarget CREATE_FLIB)
       # Not a CIME build. Add scorpio as a subdir
       add_subdirectory (${E3SM_EXTERNALS_DIR}/scorpio ${CMAKE_BINARY_DIR}/externals/scorpio)
       EkatDisableAllWarning(pioc)
+      EkatDisableAllWarning(piof)
```
jgfouca, line 59: I think mistakes were made here, `pioc` warnings were disabled twice  if CREATE_FLIB is on. I think maybe the original coder intended to disable `piof` if CREATE_FLIB is on, but that's not right either because `piof` gets built even if CREATE_FLIB is off.

components/scream/src/physics/share/tests/physics_universal_unit_tests.cpp
```diff
@@ -100,14 +100,14 @@ struct UnitWrap::UnitTest<D>::TestUniversal
...
-      printf("T to th test: abs[ T_new (%.3e) - T_in (%.3e) ]=%e is larger than the tol=%e\n",T_new,T_in,std::abs(T_new[0]-T_in),tol);
+      printf("T to th test: abs[ T_new (%.3e) - T_in (%.3e) ]=%e is larger than the tol=%e\n",T_new[0],T_in,std::abs(T_new[0]-T_in),tol);
       errors++;
     }
     // and vice versa
     const Spack T_temp = physics::th_to_T(T,exner,Smask(true));
     const Spack th_new = physics::T_to_th(T_temp,exner,Smask(true));
-    if (std::abs(T_new[0]-T_in)>tol) {
-      printf("T to th test: abs[ th_new (%.3e) - T_in (%.3e) ]=%e is larger than the tol=%e\n",th_new,T_in,std::abs(th_new[0]-T_in),tol);
+    if (std::abs(th_new[0]-T_in)>tol) {
```
jgfouca, line 109: Based on the text of the printf, I think a copy paste error was made here, so I fixed it.

jgfouca (commented):
Annotations

components/scream/cmake/tpls/Scorpio.cmake
bartgol: `piof` should not be created unless CREATE_FLIB is on.

bartgol: Ah, wait, it appears scorpio _does_ build flib by default. I believe then the correct code should have been:

```
add_subdirectory(...)
EkatDisableAllWarning(pioc)
if (PIO_ENABLE_FORTRAN) # ON by default in scorpio
  EkatDisableAllWarning(piof)
endif()
```
Without the if, someone _might_ disable f90 scropio (if we ever support pure C scorpio), and then cmake could complain that piof is not a target. Alternatively, we might change the if to the following
```
if (TARGET piof)
...
endif()
```

bartgol (changes requested):
Looks great! I only had a comment for the cmake logic when building scorpio.

bartgol (approved):

bartgol (approved):

PR #993, opened 2021-04-12
Update shoc timing structure

tcclevenger:
Small changes to the shoc timings to avoid counting variable initialization. This makes shoc match the structure of p3.

### DIFF
--- a/components/eam/src/physics/cam/shoc.F90
+++ b/components/eam/src/physics/cam/shoc.F90
@@ -192,9 +192,9 @@ subroutine shoc_main ( &
      w_sec, thl_sec, qw_sec, qwthl_sec,&  ! Output (diagnostic)
      wthl_sec, wqw_sec, wtke_sec,&        ! Output (diagnostic)
      uw_sec, vw_sec, w3,&                 ! Output (diagnostic)
-     wqls_sec, brunt, shoc_ql2,&          ! Output (diagnostic)
+     wqls_sec, brunt, shoc_ql2 &          ! Output (diagnostic)
 #ifdef SCREAM_CONFIG_IS_CMAKE
-     elapsed_s &
+     , elapsed_s &
 #endif
      )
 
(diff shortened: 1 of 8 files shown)


tcclevenger:
Here are some initial timings for shoc using these changes. I recorded the timings using the `./perf_analysis` scripts (11 repeats, ignore the first run, timing represents the average of the last 10 runs). I tested on Blake and Weaver. Factors are ~15x more columns/sec for CPU and ~35x more columns/sec for GPU. Are these close to expected?
Screenshot from 2021-04-12 13-08-05

components/eam/src/physics/cam/shoc.F90
```diff
@@ -192,7 +192,11 @@ subroutine shoc_main ( &
      w_sec, thl_sec, qw_sec, qwthl_sec,&  ! Output (diagnostic)
      wthl_sec, wqw_sec, wtke_sec,&        ! Output (diagnostic)
      uw_sec, vw_sec, w3,&                 ! Output (diagnostic)
-     wqls_sec, brunt, shoc_ql2)           ! Output (diagnostic)
+     wqls_sec, brunt, shoc_ql2,&          ! Output (diagnostic)
```
jeff-cohere, line 195: Does the comma at the end of this line mess things up when `SCREAM_CONFIG_IS_CMAKE` isn't defined?

components/scream/src/physics/shoc/shoc_main_impl.hpp
```diff
@@ -384,6 +384,7 @@ Int Functions<S,D>::shoc_main(
 
     shoc_output.pblh(i)[0] = pblh_s;
   });
+  Kokkos::fence();
```
ambrad, line 387: Nice. I was looking for this. :)

ambrad (approved):
Great!

For the timing stuff, I recommend a few things independent of this PR, at your convenience:
- Measure single-thread C++ vs single-thread F90 on an architecture with AVX2 or AVX512 instructions. Note that on SKX (like blake) there are tradeoffs in -xcore-avx2 vs -xcore-avx512; the latter uses wider registers but throttles the clock. So it's worth running both F90 and C++ with each of these flags.
- Test 48 threads as well as 96 on blake.
- It can be useful in the figure to give a better indication of speedup of the C++ w.r.t. to serial F90. You could added a text label to the F90 curve with the y-value for the final point, for example.
- You probably should add a C++ serial curve, too, again with a text label.
- If the parallel speedup on SKX isn't close to at least 40, I'd recommend doing an analysis to determine at least the following:
  - Is the C++ substantially slower than the F90 when run serially?
  - If so, isolate and fix hot spots.
  - If not, then sweep over thread count to see where the fall-off from perfect scaling occurs.

jeff-cohere (approved):

bartgol (approved):

PR #995, opened 2021-04-14
Physics universal function to calculate the virtual temperature

AaronDonahue:
This PR introduces a new physics universal function to calculate the virtual temperature from the atmosphere temperature and water vapor mass mixing ratio.


tcclevenger (approved):
Looks good to me.

PR #997, opened 2021-04-15
Upgrades to the FieldManager

bartgol:
The main goal of this PR is to upgrade the FM capabilities when it comes to handling groups of fields. In particular, this PR aims at making the FM able to accommodate multiple requests for bundled groups, even if those requests overlap. Clearly, there are sets of requests that are not all honorable (e.g., the three groups `G1=(A,B), G2=(B,C), G3=(A,C)` cannot all be bundled), but the FM can now accommodate as many as possible. E.g., the groups  `G1=(A,B,C), G2=(A,B,C,D,E), G3=(C,D), G4=(C,D,E,F), G5=((D,E,F,G)` can all be bundled, by allocating the bundle `(A,B,C,D,E,F,G)`, but it might not be as obvious if the entries in each group were scrambled.

The PR also introduces the concept of `FieldRequest`, which parallels that of `GroupRequest`. By entering all the specs for a request in a struct (including field id, pack size, and list of groups the field belongs to), the AD can now autonomously register the fields, by simply querying the atm procs for a list of their requests. This allows to _not_ hand over the field managers to the atm procs, resulting in more strict control of the AD over what the atm procs can do.

Remaining items:

- [x] ensure `qv` is the first field in the group "tracers"
- [x] propagate changes to dynamics/control/physics folders
- [x] stress test FM bundling capabilities
- [ ] now that we have multiple FM (one per grid), we might have to run some checks and/or enforce that we don't end up with different group infos on different grids. Unless we can verify that such scenario can still be acceptable, and handled correctly.


components/scream/src/share/util/scream_utils.hpp
```diff
@@ -0,0 +1,191 @@
...
+//             algorithm has failed, and we can return.
+//     - if G has a portion that does not intersect with any sublist S, add it
+//       as a new sublist at the end. This corresponds to operation (c) from above.
+
+// Regardless of the order in which we process the individual lists, if there is
+// an arrangement A of T's that allows to have G1,...,Gn as a contiguous sublist
+// of A, the above algorithm is guaranteed to find it.
+
+template<typename T>
+std::list<T> contiguous_superset (const std::list<std::list<T>>& groups)
```
bartgol, line 61: I'm not sold on the name. What the fcn does is, given `lol=list<list<T>>` find an ordering of all the elements `ordered=list<T>` such that each list in `lol` is a contiguous sublist of `ordered`.

bartgol:
@AaronDonahue @tcclevenger @jeff-cohere 

Although I still have tests failing (will get to those tonight), the bulk of this PR is done. I am not turning on AT testing yet (given the failures), but it might be worth starting to look at it, to see if something could be done better.

components/scream/src/share/util/scream_utils.hpp
jeff-cohere: Can you give a contextual example of how this is used?

bartgol: Take the tracers. There's a big large Q array, of dimensions (ncols,ntracers,nlevs), containing all the tracers. But there are different kind of tracers, and some may require certain updates that others don't. E.g., shoc needs to do apply some sort of turbulent scheme to some of them, but not all; radiation, on the other hand, needs some of them for optical property stuff. All the atm procs that use some (but not all) of the tracers can benefit from having their subset of tracers contiguous within Q. Here, contiguous refers to the tracer index, so that their subset of tracers looks like `Q(:, my_start:my_end, :)`, which is not contiguous in memory, but once subviewed at a particular col idx is.

So, in order to accommodate (if possible) all requests for contiguous sub-groups, I need to rearrange the order of the tracers in Q, in a way that all the sub-groups are a contiguous set of slices. This utility is doing this: given some sets of objects (of the same type, variable sizes), find a way to list the elements in their union that allows each of these sets to be a contiguous subset of it.

bartgol: Note: I'm using `list` instead of `set` in the impl, for code reasons, not 'conceptual' reasons. From the conceptual point of view, those lists are like sets (meaning, no repeated entries, and the order of elements doesn't matter). But I can control how entries in a list are ordered, while set does not let me rearrange them.

bartgol:
Ok, tests now pass on my laptop. A summary of what this PR does:

1. Creates FieldRequest objects, parallel to GroupRequests.
2. AtmosphereProcess exposes FieldRequest objects instead of FieldIdentifier's in the calls to `get_required_fields` and `get_computed_fields`.
3. Remove `register_fields` method from AtmosphereProcess. The AD can use the FieldRequest info to register the fields in the FM (or FMs). This reduces interaction of Atm procs with AD internals, which removes room for hacks (e.g., storing a ptr to the FM, to later access fields that were not requested).
4. Enhance GroupRequest, to allow creating requests that "depend" on other requests. E.g., create request for group "A", which contains all the fields in group "B", _except_ fields "f1,...,fn".
5. Add FM's capability to handle "overlapped" bundled fields. In particular, the FM will "rearrange" fields in groups, to accommodate as many requests for bundled groups as possible. E.g., given G1=(f1,f2), G2=(f2,f3), the FM will allocate the group G=(f1,f2,f3), so that both G1 and G2 are "contiguous" sub-bundles of G.
6. Add testing for FM overlapped bundles feature.

components/scream/src/physics/cld_fraction/atmosphere_cld_fraction.cpp
```diff
@@ -33,18 +34,18 @@ void CldFraction::set_grids(const std::shared_ptr<const GridsManager> grids_mana
...
-  add_required_field("qi",   scalar3d_layout_mid, Q,      grid_name);
-  add_required_field("cldfrac_liq", scalar3d_layout_mid, nondim, grid_name);
+  constexpr int ps = Pack::n;
+  add_field<Required>("qi",   scalar3d_layout_mid, Q,      grid_name,"tracers",ps);
+  add_field<Required>("cldfrac_liq", scalar3d_layout_mid, nondim, grid_name,ps);
 
   // Set of fields used strictly as output
-  add_computed_field("cldfrac_tot",   scalar3d_layout_mid, nondim, grid_name);
-  add_computed_field("cldfrac_ice",  scalar3d_layout_mid, nondim, grid_name);
+  add_field<Computed>("cldfrac_tot",   scalar3d_layout_mid, nondim, grid_name,ps);
```
AaronDonahue, line 47: I like the use of the template for `add_field`

components/scream/src/share/field/field_request.hpp
```diff
@@ -0,0 +1,225 @@
+#ifndef SCREAM_FIELD_REQUEST_HPP
+#define SCREAM_FIELD_REQUEST_HPP
+
+#include "share/field//field_identifier.hpp"
+
+namespace scream {
+
+// Whether a the bundling of a field group (see below) is needed, optional, or not needed.
```
AaronDonahue, line 8: minor typo, not important, but might as well clean up "whether a the"

AaronDonahue:
@brhillman and @tcclevenger, just pinging you to keep an eye on this PR.  It's a pretty big overhaul of the interface with fields and I imagine once it goes in you will need to rebase any branches you are working on for ongoing development.  

I'm also noting just for myself in this comment that #987 will have to be rebased if this goes in first, or @bartgol you may need to make adjustments to the code if 987 goes in first.  Is there actually a better order to handle that?

components/scream/src/share/util/scream_utils.hpp
jeff-cohere: Cool. I was confused before because I thought that the existence of this function was an assertion that such an ordering could always be found, but I see now that it returns an empty list if this is not the case. And using a `list` does indeed make sense if you want control of the ordering, since `set` is an ordered set and `unordered_set` is indifferent to ordering.

If elements in a group have a canonical ordering, I think `contiguous_superset` is a decent name. To me, It doesn't really seem like a partitioning, since partitions are disjoint and the function returns a single list from which overlapping subsets may be selected.

My only question is: are there circumstances of interest to SCREAM where one could find no such ordering that would cause trouble for the caller? Another way to ask this question is: does the possibility of no such ordering place an undue burden on the caller to handle edge cases, or is this unusual in practice?

bartgol:
I obviously can adjust this PR if other PRs go in first causing conflicts. And I can also help with fixing other branches, if need be.

components/scream/src/share/field/field_layout.cpp
```diff
@@ -135,5 +135,23 @@ LayoutType get_layout_type (const std::vector<FieldTag>& field_tags) {
   return result;
 }
 
+std::string print (const FieldLayout& layout)
```
jeff-cohere, line 138: In Python, this function would be called `str` or `repr`.

jeff-cohere: Actually `__str__` or `__repr__`, but the `str` and `repr` functions are what the caller would use to access it. Just another perspective.

components/scream/src/share/field/field_request.hpp
```diff
@@ -0,0 +1,225 @@
...
+ * A typical example is an AP that needs to advect tracers (like Dynamics does):
+ * it treats tracers agnostically, and does not really care how many there are.
+ * So the AP exposes this need as a GroupRequest. Later, it will be provided
+ * with a FieldGroup, which allows to access all the fields in the group
+ * individually, and, if the allocation permits it, as a single N+1 dimensional
+ * field. For more details about the FieldGroup struct, see field_group.hpp.
+ */
+struct GroupRequest {
+  // Main constructor method
+  GroupRequest (const std::string& name_, const std::string& grid_, const int ps, const Bundling b,
```
jeff-cohere, line 50: Might be good to give brief descriptions of the constructor parameters.

components/scream/src/share/field/field_request.hpp
```diff
@@ -0,0 +1,225 @@
...
+ * 
+ * The request contains at least a FieldIdentifier, but can also contain
+ * a pack size, and a list of names of groups that the field should belong to.
+ */
+
+struct FieldRequest {
+  using Units = ekat::units::Units;
+  using FID   = FieldIdentifier;
+
+  // Main constructor
```
jeff-cohere, line 165: Brief parameter descriptions here would be good, too.

components/scream/src/share/util/scream_utils.hpp
bartgol: Honestly, I don't know. I think there might be such cases once we get aerosols. Also, SL transport in Homme, as I understand, will distinguish between active and passive tracers.

If we ever end up with a set of tracers that does not allow all the subgroups to be contiguous, and if these subgroups _really need_ to be bundled, then we will need some re-thinking of the impl of the atm procs issuing those requests.

I believe if that happens, we will have to pick 1+ of these procs, and turn its request for bundled fields into a "preferred" bundling. The FM does a few tries when building bundled groups: first, all the 'Required' and 'Preferred' bundling requests are considered; if that fails, it starts removing one 'Preferred' request at a time, until it succeeds (which it will, at the latest when there's only one request left). So if we ever hit the case where all atm procs issue a GroupRequest with 'Required' bundling, we'll have to go through 1+ of these atm procs, and change the request to 'Preferred'. However, that means that the atm proc will have to be able to handle a contiguous bundle (like `Q(:,start:end,:)`) as well as completely independent fields.

So, long story short, I think it's possible, as the number of parametrizations grows, that we'll hit a case where we can't bundle all groups. If that happens, 1+ of the parametrizations will have to implement a "fall-back" strategy, where it can deal with separate fields rather than a single N+1 dimensional field.

bartgol: Tbc, I don't think we should implement the 'fall-back' strategy now. Homme and SHOC (the only current users of groups) can still issue a GroupRequest with bundling=Required. We'll cross that bridge if/when we get there.

components/scream/src/share/field/field_manager.hpp
```diff
@@ -57,32 +63,12 @@ class FieldManager {
 
   // Change the state of the database
   void registration_begins ();
+  void register_field (const FieldIdentifier& fid) { register_field(FieldRequest(fid)); }
```
jeff-cohere, line 66: Worth calling out that this overload assumes a pack size of 1? Just curious.

components/scream/src/share/util/scream_utils.hpp
```diff
@@ -0,0 +1,193 @@
...
+    copy.sort();
+    return copy;
+  };
+  auto unique = [] (const std::list<T>& l) {
+    auto copy = l;
+    copy.unique();
+    return copy;
+  };
+
+  std::list<std::list<T>> lol;
```
jeff-cohere, line 98: 🤣

components/scream/src/share/util/scream_utils.hpp
bartgol: Another note: Homme currently is adding a constraint, namely that qv must be the 1st tracer. There really isn't a need for that, since accessing a single tracer at iq=0 is the same as accessing at iq=23. At some point, I'd like to change homme, to store `qv_idx`, and using that every time it needs qv (which is needed for T<->T_virt conversions). However, that's a non-trivial change in Homme, since it might be time-consuming to track down all the places where Q is accessed only for qv's values. So for now I opted for hacking the FM, so that if qv is present _and_ the "tracers" group is present, we manage to put qv at the front of the list.

Forcing qv at the front _might_ be a constraint that causes the FM to fail to honor all requests. I don't think it will, but it's worth keeping this "pointless" constraint in mind.

Note: the logic for putting qv at the front of the list might fail if there is a group G that is a superset of the tracers. I don't think that will happen any time soon, but it's worth keeping in mind (I think I added a comment in the code).

components/scream/src/share/field/field_manager.hpp
bartgol: Actually, I think I left this ctor to accommodate the use in `field_tests.cpp`. However, I'm trhinking of removing it, and force the user to build a FieldRequest (which can be done on the fly).

components/scream/src/share/util/scream_utils.hpp
bartgol: I couldn't resist. :)

components/scream/src/physics/shoc/atmosphere_macrophysics.cpp
```diff
@@ -49,51 +49,46 @@ void SHOCMacrophysics::set_grids(const std::shared_ptr<const GridsManager> grids
   // Define fields needed in SHOC.
   // Note: shoc_main is organized by a set of 5 structures, variables below are organized
   //       using the same approach to make it easier to follow.
+  std::set<ci_string> q_names =
```
jeff-cohere, line 52: Is `q_names` actually used anywhere now?

bartgol: ah, shoot, I thought I deleted it. I copied and pasted it from the `register_fields` method, so I could see which ones needed the "tracers" group, but then forgot to delete the line. Good catch!

jeff-cohere (approved):
*Whew*! Quite a lot of stuff going on here. But it looks good--I had only minor questions. I think the clarity of the `FieldRequest` and `GroupRequest` types offsets the complexity of accommodating more nouns.

I'm approving, subject to the NetCDF issue on Mappy being sorted out.

components/scream/src/share/field/field_manager.hpp
bartgol: Turns out, it wasn't used anywhere at all. Purged.

components/scream/src/share/field/field_layout.cpp
bartgol: I opted for `to_string`, mimicking ekat's util for `Units` (and some std utils).

jeff-cohere (approved):
👍

PR #1004, opened 2021-04-19
Change shoc 1d views to have type Scalar

tcclevenger:
A bit of code clean up. 1d views in shoc were type `view<ekat::Pack<Scalar,1>>`, now they are type `view<Scalar>`. Just simplifies the code a bit.

### DIFF
--- a/components/scream/src/physics/shoc/shoc_functions.hpp
+++ b/components/scream/src/physics/shoc/shoc_functions.hpp
@@ -39,7 +39,6 @@ struct Functions
   using IntSmallPack = SmallPack<Int>;
   using Pack = BigPack<Scalar>;
   using Spack = SmallPack<Scalar>;
-  using Pack1d = ekat::Pack<Scalar, 1>;
 
   using Mask  = ekat::Mask<Pack::n>;
   using Smask = ekat::Mask<Spack::n>;
@@ -74,9 +73,9 @@ struct Functions
     SHOCInput() = default;
 
     // Grid spacing of host model in x direction [m]
-    view_1d<const Pack1d> host_dx;
+    view_1d<const Scalar> host_dx;
     // grid spacing of host model in y direction [m]
-    view_1d<const Pack1d> host_dy;
+    view_1d<const Scalar> host_dy;
     // heights, for thermo grid [m]
     view_2d<const Spack>  zt_grid;
     // heights, for interface grid [m]
@@ -92,19 +91,19 @@ struct Functions
     // large scale vertical velocity [m/s]
     view_2d<const Spack>  w_field;
     // Surface sensible heat flux [K m/s]
-    view_1d<const Pack1d> wthl_sfc;
+    view_1d<const Scalar> wthl_sfc;
     // Surface latent heat flux [kg/kg m/s]
-    view_1d<const Pack1d> wqw_sfc;
+    view_1d<const Scalar> wqw_sfc;
     // Surface momentum flux (u-direction) [m2/s2]
-    view_1d<const Pack1d> uw_sfc;
+    view_1d<const Scalar> uw_sfc;
     // Surface momentum flux (v-direction) [m2/s2]
-    view_1d<const Pack1d> vw_sfc;
+    view_1d<const Scalar> vw_sfc;
     // Surface flux for tracers [varies]
     view_2d<const Spack>  wtracer_sfc;
     // Exner function [-]
     view_2d<const Spack>  exner;
     // Host model surface geopotential height
-    view_1d<const Pack1d> phis;
+    view_1d<const Scalar> phis;
   };
 
   // This struct stores input/outputs views for shoc_main.
@@ -140,7 +139,7 @@ struct Functions
     SHOCOutput() = default;
 
     // planetary boundary layer depth [m]
-    view_1d<Pack1d> pblh;
+    view_1d<Scalar> pblh;
     // cloud liquid mixing ratio variance [kg^2/kg^2]
     view_2d<Spack>  shoc_ql2;
   };
--- a/components/scream/src/physics/shoc/shoc_functions_f90.cpp
+++ b/components/scream/src/physics/shoc/shoc_functions_f90.cpp
@@ -7,6 +7,8 @@
 #include "ekat/ekat_pack_kokkos.hpp"
 #include "ekat/kokkos/ekat_subview_utils.hpp"
 
+#include "share/util/scream_deep_copy.hpp"
+
 #include <random>
 
 using scream::Real;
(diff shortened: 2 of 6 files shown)


components/scream/src/physics/shoc/shoc_functions.hpp
```diff
@@ -74,9 +73,9 @@ struct Functions
     SHOCInput() = default;
 
     // Grid spacing of host model in x direction [m]
-    view_1d<const Pack1d> host_dx;
+    view_1d<const Scalar> host_dx;
```
jeff-cohere, line 76: This might be a silly question, but why do we sometimes call this data type a `Scalar` and other times a `Real`?

components/scream/src/physics/shoc/shoc_functions_f90.cpp
```diff
@@ -1799,7 +1826,7 @@ void check_length_scale_shoc_length_f(Int nlev, Int shcol, Real* host_dx, Real*
   // Sync to device
   std::vector<view_1d> temp_1d_d(2);
   std::vector<view_2d> temp_2d_d(1);
-  ekat::host_to_device({host_dx,host_dy}, shcol, temp_1d_d);
+  copy_to_device({host_dx,host_dy}, shcol, temp_1d_d);
   ekat::host_to_device({shoc_mix}, shcol, nlev, temp_2d_d, true);
```
jeff-cohere, line 1830: Any reason not to use `copy_to_device` here too?

jeff-cohere (approved):
Looks good to me.

components/scream/src/physics/shoc/shoc_functions_f90.cpp
tcclevenger: `copy_to_device` is a function I wrote for 1d scalar views. `shoc_mix` is a packed view, so it still requires the ekat function.

components/scream/src/physics/shoc/shoc_functions.hpp
tcclevenger: Good question.. I'm not quite sure why the shoc `Functions` structure allows for a different scalar type that Real in scream_types..

components/scream/src/physics/shoc/shoc_functions_f90.cpp
```diff
@@ -843,6 +843,42 @@ void pblintd(PblintdData& d)
 
 // end _c impls
 
+// Helper function to copy 1d data to and from Device
+template<typename ViewT>
+void copy_to_device(const std::vector<typename ViewT::value_type const*>& data,
```
bartgol, line 848: So, the only big difference between this fcn and what's in ekat (beside the fact that you have a fixed size for all views, which is a minor difference) is that your views are `view<Real*>` rather than `view<Pack<T,N>*>`, right?

Would it make sense to put these utilities in `scream/src/share/util`? Something like `scream_deep_copy.hpp`. I suspect other parts of scream might need this at some point.

Also (but not for this PR, I think), we might have a simpler version:
```
template<typename T>
void deep_copy(const view_1d<T>& dst, typename ViewT::value_type const* src);
template<typename T>
void deep_copy(typename ViewT::value_type const* dst, const view_1d<T>& src);

template<typename T>
void deep_copy(const view_2d<T>& dst, typename ViewT::value_type const* src, const int src_stride);
template<typename T>
void deep_copy(typename ViewT::value_type const* dst, const view_2d<T>& src, const int dst_stride);
```
and have the 'vector' version simply call this. We should also use enalbe_if to provide an overload if T=Pack<S,N>, cause in that case we need the 'physical' length, to avoid reading/writing out of bounds.
The single view version can be maybe a tad easier to visually parse, when syncing 1 or 2 views:
```
view_1d a(ncols);
view_2d b(ncols,nlevs), c(ncols,nlevs);
deep_copy(a, a_data);
deep_copy(b, b_data, nlevs);
deep_copy(c, c_data, nlevs);
```
I feel like creating the vector kind of adds a bit of clunkiness, and doesn't save many lines of code for a handful of views.

components/scream/src/physics/shoc/shoc_functions_f90.cpp
```diff
@@ -1047,7 +1080,7 @@ void update_host_dse_f(Int shcol, Int nlev, Real* thlm, Real* shoc_ql, Real* exn
   std::vector<const Real*> ptr_array = {thlm,  shoc_ql, exner, zt_grid, host_dse};
 
   // Sync to device
-  ekat::host_to_device({phis}, shcol, temp_1d_d);
+  copy_to_device({phis}, shcol, temp_1d_d);
```
bartgol, line 1083: Follow up on the comment above: here's an example where having to create a vector makes the code a bit less obvious, imho. It'd be easier to parse if we could do
```
view_1d phis_d(shcol);
copy_to_device(phis_d,phis);
```

components/scream/src/physics/shoc/shoc_functions_f90.cpp
```diff
@@ -1934,25 +1959,25 @@ void shoc_diag_obklen_f(Int shcol, Real* uw_sfc, Real* vw_sfc, Real* wthl_sfc, R
...
+  using Spack   = typename SHOC::Spack;
   using Scalar  = typename SHOC::Scalar;
-  using view_2d = typename SHOC::view_2d<Pack1>;
-  using view_1d = typename SHOC::view_1d<Pack1>;
+  using view_2d = typename SHOC::view_2d<Spack>;
+  using view_1d = typename SHOC::view_1d<Scalar>;
 
-  std::vector<Int> dim1(2, shcol);
-  std::vector<Int> dim2  = {nlevi,  nlev};
+  std::vector<Int> dim1 = {shcol, shcol};
```
bartgol, line 1979: This change seems to be unrelated to this PR. Was it a bug you caught?

bartgol (approved):
Looks good. I have some comments, but nothing major.

jgfouca:
@tcclevenger , just curious, why are shoc views not using SIMD Packs?

bartgol:
> @tcclevenger , just curious, why are shoc views not using SIMD Packs?

The ones he changed are 2d fields on the surface (which means 1d view of size ncols). For those, packing does not buy you anything. And if by chance we turned one of those into a strided subview of a 3d field (i.e., 2d view), it might even be wrong.

components/scream/src/physics/shoc/shoc_functions_f90.cpp
tcclevenger: Yeah, I didn't mean to include that. I deleted that line (evidentially) and wrote it back slightly different. I'll change back.

components/scream/src/physics/shoc/shoc_functions_f90.cpp
tcclevenger: Yes, I think putting this in `scream/src/share/util` makes sense. I'll move it there.

tcclevenger:
@bartgol I moved the copy _to* functions to scream/../utils. In a later PR I'll do the impl of the simpler version

jgfouca:
@bartgol , gotcha. So these views have 1 scalar per column.

jgfouca (approved):

Issue #1002, opened 2021-04-15
Physics universal function to calculate layer height from surface - z_int, z_mid & dz


AaronDonahue:
A function is needed to determine geopotential height offset (height above the surface in meters) at cell interfaces (`z_int`) and cell midpoints (`z_mid`). As well as layer thickness, `dz`.

There is guidance on what this algorithm looks like here:

https://acme-climate.atlassian.net/wiki/spaces/NGDNA/pages/2605384045/Vertical+Coordinate?focusedCommentId=2605417001#comment-2605417001

bartgol:
I think #1019 took care of this. Closing.

PR #1006, opened 2021-04-19
Create `get_dz` function to determine dz using the equation of state

AaronDonahue:
This PR overloads the `get_dz` function in the universal physics to allow for a calculation of `dz` given the equation of state.

Addresses part of #1002


AaronDonahue:
@bartgol , it looks like the tests are passing except the "release" tests on mappy.  I'll investigate those fails, but in the meantime, when you have a chance could you make sure the code itself looks good?

AaronDonahue:
Handled in PR #1019

Issue #835, opened 2020-12-15
Finalize C++ version of Simple Prescribed Aerosol (SPA)


PeterCaldwell:
Once #834 is done, use it as a template to make our C++ code also use SPA forcings appropriately.

bartgol:
@AaronDonahue I think this was done by #1153 , right? I mean, we still miss horiz interpolation, but assuming one has the aerosols nc file at the same res of the state, then SPA _is_ working, right? Can we close?

AaronDonahue:
I'm all for closing issues, and #1153 did get us most of the way there.  That being said,this issue is to finalize the code and that isn't finished yet.  I think I can reference this issue in the next SPA PR so we don't lose sight of it and go ahead and close the issue when that PR is merged.

bartgol:
@AaronDonahue can this be closed now that 1260 went in?

Issue #834, opened 2020-12-15
Finalize F90 version of Simple Prescribed Aerosol (SPA)


PeterCaldwell:
- [ ] complete 5 yr ne30 simulations with and without prescribed aerosol and create e3sm_diags to confirm that the difference between these runs isn't large.
- [ ] Create SPA forcing files (including both CCN and aerosol optics properties for each climatological month)
- [ ] Put SPA forcing files in the E3SM inputdata server
- [ ] Modify SCREAM code to use these new files by default.

PR #1009, opened 2021-04-19
Beydoun/spa update

hassanbeydoun:
PR for reading in all SPA files during initialization in both micro_p3_interface.F90 and radiation.F90. Aaron had already done the former but I removed the 12 mon_ccn's and replaced them with a single mon_ccn with an additional month dimension. I also fixed a units error when using nccn_prescribed in micro_p3.F90.


PeterCaldwell:
@hassanbeydoun - It looks like the autotester is failing in `p3_tests_ut_np1_omp*`
`REQUIRE( isds_fortran[i].nc[k] == isds_cxx[i].nc[k] )
with expansion:
184189.6610748996 == 0.5671007917`
which means that nc computed by p3 is being changed by this PR. I don't think it should be because the default behavior for the tests is to NOT to use SPA, right?

hassanbeydoun:
> @hassanbeydoun - It looks like the autotester is failing in `p3_tests_ut_np1_omp*`
> `REQUIRE( isds_fortran[i].nc[k] == isds_cxx[i].nc[k] ) with expansion: 184189.6610748996 == 0.5671007917`
> which means that nc computed by p3 is being changed by this PR. I don't think it should be because the default behavior for the tests is to NOT to use SPA, right?

@PeterCaldwell yes the default should be not using SPA...

PeterCaldwell:
One general comment/question: there are a bunch of places in this PR where arithmetic on floats is applied using numbers which aren't explicitly defined as r8. Is this ok? For example `1-cldfrac` instead of `1._r8 - cldfrac`? This is maybe more of a question for @AaronDonahue than Hassan.

components/eam/src/physics/rrtmgp/radiation.F90
```diff
@@ -558,9 +554,11 @@ subroutine radiation_init(state,pbuf)
...
       integer :: year, month, day, tod, next_month
       real(r8), pointer :: aerosol_optical_property(:,:,:,:)
-      real(r8), pointer :: aerosol_optical_property_lw(:,:,:,:)
-      real(r8), pointer :: aerosol_optical_property_sw(:,:,:,:)
+      real(r8), pointer :: aerosol_optical_property_lw(:,:,:,:,:)
+      real(r8), pointer :: aerosol_optical_property_sw(:,:,:,:,:)
 
+      nullify(aerosol_optical_property_lw)
+      nullify(aerosol_optical_property_sw)
       !-----------------------------------------------------------------------
```
PeterCaldwell, line 562: Why would you need to nullify a pointer immediately after creating it? How could it possibly be assigned at this point?

components/eam/src/physics/rrtmgp/radiation.F90
```diff
@@ -1511,44 +1491,26 @@ subroutine radiation_tend(state,   ptend,    pbuf,          cam_out, cam_in,  &
...
                      !populate with corresponding pbuf
                      !variables
-                     call pbuf_get_field(pbuf,aer_tau_bnd_sw_mon_1_idx, aer_tau_bnd_sw_mon_1)
-                     call pbuf_get_field(pbuf,aer_tau_bnd_sw_mon_2_idx, aer_tau_bnd_sw_mon_2)
-                     call pbuf_get_field(pbuf,aer_ssa_bnd_sw_mon_1_idx, aer_ssa_bnd_sw_mon_1)
-                     call pbuf_get_field(pbuf,aer_ssa_bnd_sw_mon_2_idx, aer_ssa_bnd_sw_mon_2)
-                     call pbuf_get_field(pbuf,aer_asm_bnd_sw_mon_1_idx, aer_asm_bnd_sw_mon_1)
-                     call pbuf_get_field(pbuf,aer_asm_bnd_sw_mon_2_idx, aer_asm_bnd_sw_mon_2)
-
+                     call pbuf_get_field(pbuf,aer_tau_bnd_sw_mon_idx, aer_tau_bnd_sw_mon)
```
PeterCaldwell, line 1494: Maybe I'm just not seeing things correctly, but shouldn't this stuff be inside an "if spa" block? We don't need to do it if SPA isn't called, right?

components/eam/src/physics/rrtmgp/radiation.F90
```diff
@@ -1511,44 +1491,26 @@ subroutine radiation_tend(state,   ptend,    pbuf,          cam_out, cam_in,  &
...
+                     call pbuf_get_field(pbuf,aer_ssa_bnd_sw_mon_idx, aer_ssa_bnd_sw_mon)
+                     call pbuf_get_field(pbuf,aer_asm_bnd_sw_mon_idx, aer_asm_bnd_sw_mon)
+
+                     if (month==12) then
+                         next_month = 1
+                     else
+                         next_month = month + 1
+                     end if
                      if (current_month .ne. month) then
```
PeterCaldwell, line 1503: I'm weirded out by the logic above and below this comment. If current_month and month are always set to the same thing, why not delete month and only use current_month? And is next_month getting set to the right value if current_month .ne. month?

components/eam/src/physics/rrtmgp/radiation.F90
```diff
@@ -1656,25 +1618,19 @@ subroutine radiation_tend(state,   ptend,    pbuf,          cam_out, cam_in,  &
                      call get_curr_date(year,month,day,tod)
                      !populate with corresponding pbuf
                      !variables
-                     call pbuf_get_field(pbuf,aer_tau_bnd_lw_mon_1_idx, aer_tau_bnd_lw_mon_1)
-                     call pbuf_get_field(pbuf,aer_tau_bnd_lw_mon_2_idx, aer_tau_bnd_lw_mon_2)
-                     if (current_month .ne. month) then
-                        aer_tau_bnd_lw_mon_1 = aer_tau_bnd_lw_mon_2
-                        if (month==12) then
+                     call pbuf_get_field(pbuf,aer_tau_bnd_lw_mon_idx, aer_tau_bnd_lw_mon)
+                     if (month==12) then
```
PeterCaldwell, line 1622: current month and next month were already defined above in this same function, right? I suspect these definitions aren't needed (though they're cheap and perhaps nice to remind people where these variables were defined?)

run_scream.syrah_ne30_control.csh=a_run_link
```diff
@@ -0,0 +1 @@
+/p/lustre2/beydoun1/syrah.default.run_SPA_new_optics.ne30_ne30
```
PeterCaldwell, line 1: this file should be deleted. It snuck into git by accident.

components/eam/src/physics/rrtmgp/radiation.F90
AaronDonahue: The above definition appears to be nested within an `if` statement.  Hence the repeat.  However, since the same action is taken regardless of the result of the `if` statement we might as well move this chunk of code to just before the `if` statement.

components/eam/src/physics/rrtmgp/radiation.F90
```diff
@@ -1656,25 +1618,19 @@ subroutine radiation_tend(state,   ptend,    pbuf,          cam_out, cam_in,  &
...
                            next_month = 1
                         else
                            next_month = month + 1
-                        end if
-                        allocate(aerosol_optical_property_lw(pcols,pver,nlwbands,begchunk:endchunk))
-                        call get_aerosol_optical_property_from_file(next_month,'AER_TAU_LW','lwband',nlwbands,aerosol_optical_property_lw)
-                        aer_tau_bnd_lw_mon_2 = aerosol_optical_property_lw(:,:,:,state%lchnk)
-                        deallocate(aerosol_optical_property_lw)
+                     end if
+                     if (current_month .ne. month) then
```
AaronDonahue, line 1627: It's my understanding that we don't really use `current_month` as a universal variable anymore.  It's only use is below for the interpolation, which could also simply use `month` right?  If so, I think we can get rid of `current_month` in it's current form everywhere.  If we like the term `current_month` then we can use that instead of `month` inside this subroutine.

components/eam/src/physics/rrtmgp/radiation.F90
AaronDonahue: I think the idea here is that month is set by asking EAM to provide the actual month for this timestep.  current_month was meant to keep track of what month the last time the interpolation was done.  It made more sense when we were loading files on the fly, since you would only load new data if the "current_month" has changed, which would mean your currently loaded data was outdated.
As for `next_month`, if we set `current_month` before calculating `next_month` then it would be okay to use current_month in this calculation.  But since everything is being treated as dependent on `month` which is the only thing actually updated by timestep at this point, the code is using month here.  In one of my other comments I advocate for removing `current_month` unless it has some other use outside of the interpolation.

components/eam/src/physics/rrtmgp/radiation.F90
```diff
@@ -1340,20 +1328,12 @@ subroutine radiation_tend(state,   ptend,    pbuf,          cam_out, cam_in,  &
...
+      real(r8), pointer :: aer_tau_bnd_lw_mon(:,:,:,:)
+      real(r8), pointer :: aer_tau_bnd_sw_mon(:,:,:,:)
+      real(r8), pointer :: aer_ssa_bnd_sw_mon(:,:,:,:)
+      real(r8), pointer :: aer_asm_bnd_sw_mon(:,:,:,:)
 
       real(r8), dimension(12):: days_per_month
-      nullify(aerosol_optical_property_lw)
-      nullify(aerosol_optical_property_sw)
       !fill days_per_month, SPA doesn't recognize leap year
       days_per_month = (/31,28,31,30,31,30,31,31,30,31,30,31/)
```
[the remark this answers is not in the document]
PeterCaldwell: Yeah, just hardcoding days_per_month makes me nervous. I think we want SPA to be as simple as possible so I don't think we want to add sophisticated "grab this info from a central repository depending on calendar" stuff here, but adding a comment saying this is also hardcoded in rrtmgp and that SPA only works with a noleap calendar would be good and easy to do.

components/eam/src/physics/rrtmgp/radiation.F90
PeterCaldwell: Ok, that's what I thought. I agree we should remove current_month.

AaronDonahue:
> > @hassanbeydoun - It looks like the autotester is failing in `p3_tests_ut_np1_omp*`
> > `REQUIRE( isds_fortran[i].nc[k] == isds_cxx[i].nc[k] ) with expansion: 184189.6610748996 == 0.5671007917`
> > which means that nc computed by p3 is being changed by this PR. I don't think it should be because the default behavior for the tests is to NOT to use SPA, right?
> 
> @PeterCaldwell yes the default should be not using SPA...

It looks like in `p3_tests` though there is 2/4 tests that do trigger prescribed_CCN, which is being fundamentally changed in this PR.  See `p3_main_unit_tests.cpp` this chunk of code:

```
  P3MainPart1Data isds_fortran[] = {
    //            kts, kte, ktop, kbot, kdir, do_predict_nc, do_prescribed_CCN,       dt
    P3MainPart1Data(1,  72,    1,   72,    1, false,          true,           1.800E+03),
    P3MainPart1Data(1,  72,    1,   72,    1, true,           true,           1.800E+03),
    P3MainPart1Data(1,  72,   72,    1,   -1, false,          false,          1.800E+03),
    P3MainPart1Data(1,  72,   72,    1,   -1, true,           false,          1.800E+03),
  };
```

And note that in this PR line 413 in `components/eam/src/physics/cam/micro_p3.F90` has 

```
          if (do_prescribed_CCN) then
             nc(k) = max(nc(k),nccn_prescribed(k)*1.0e6*inv_rho(k))
          else if (do_predict_nc) then
```

where the 1.0e6*inv_rho(k) is new.

PeterCaldwell:
replying to @AaronDonahue 's brilliant observation about the 1e6 bugfix in do_prescribed_CCN: ah, that makes total sense why the autotester is failing. Did we mean to apply the 1e6 bugfix in "do_prescribed_CCN", or was that intended to be a bugfix in "do_SPA" cases?

hassanbeydoun:
> replying to @AaronDonahue 's brilliant observation about the 1e6 bugfix in do_prescribed_CCN: ah, that makes total sense why the autotester is failing. Did we mean to apply the 1e6 bugfix in "do_prescribed_CCN", or was that intended to be a bugfix in "do_SPA" cases?

The syntax in SPA-CCN is "do_prescribed_CCN" not "do_SPA".

PeterCaldwell:
> The syntax in SPA-CCN is "do_prescribed_CCN" not "do_SPA".
Ok, sorry for totally not being on top of the naming conventions. That makes sense. I'm confused though - I know our tests cover do_predict_nc (having checked the actual name this time) but I don't remember extending them to include do_prescribed_CCN. I guess it's good that they do include do_prescribed_CCN and makes sense that those tests would fail with this new bugfix. So all is well.

AaronDonahue:
> > replying to @AaronDonahue 's brilliant observation about the 1e6 bugfix in do_prescribed_CCN: ah, that makes total sense why the autotester is failing. Did we mean to apply the 1e6 bugfix in "do_prescribed_CCN", or was that intended to be a bugfix in "do_SPA" cases?
> 
> The syntax in SPA-CCN is "do_prescribed_CCN" not "do_SPA".

@hassanbeydoun , how does that align with the option in RAD at this line: https://github.com/E3SM-Project/scream/pull/1009/files#diff-b7b0ba62bfed1a061af2d6d58136768865459f91a35b4cbc128af731f3cd57bcR1489

for reference this is line 1489 in `components/eam/src/physics/rrtmgp/radiation.F90`

where `do_SPA_optics` is being used.

PeterCaldwell:
Good point because it validates my previous poor reading skills, @AaronDonahue ! Seriously though, I thought we were going to merge do_prescribed_CCN and do_SPA_optics into a single "do_SPA" flag once we had both completed?

components/eam/src/physics/cam/micro_p3_interface.F90
```diff
@@ -373,13 +348,10 @@ subroutine get_prescribed_CCN_from_file(micro_p3_lookup_dir,month_int, ccn_value
       write(mon_str,*) month_int
 
       !assign base_file_name the name of the CCN file being used 
-      base_file_name = "prescribed_CCN_file_"
+      base_file_name = "spa_file_"
```
AaronDonahue, line 351: Have we updated these filenames on the machines we are running on?

components/eam/src/physics/cam/micro_p3_interface.F90
```diff
@@ -847,13 +814,12 @@ subroutine get_prescribed_CCN(nccn_prescribed,micro_p3_lookup_dir,its,ite,kts,kt
...
-      call pbuf_get_field(pbuf,mon_ccn_idx(month), mon_ccn_1)
 
-      call pbuf_get_field(pbuf,mon_ccn_idx(next_month), mon_ccn_2)
+      call pbuf_get_field(pbuf,mon_ccn_idx, mon_ccn)
 
-      fraction_of_month = (day*3600.0*24.0 + tod)/(3600*24*days_per_month(current_month)) !tod is in seconds
+      fraction_of_month = (day*3600.0*24.0 + tod)/(3600*24*days_per_month(month)) !tod is in seconds
 
-      nccn_prescribed = mon_ccn_1*(1-fraction_of_month) + mon_ccn_2*(fraction_of_month)
+      nccn_prescribed = mon_ccn(:,:,month)*(1-fraction_of_month) + mon_ccn(:,:,next_month)*(fraction_of_month)
```
AaronDonahue, line 822: So here it looks like we are loading the full range of CCN values from pbuf and then just using `month` and `next_month` for the interpolation.  But in the radiation code changes we are still using the old approach.  Would it make sense (or is it even possible) to make these consistent?

components/eam/src/physics/rrtmgp/radiation.F90
```diff
@@ -1656,25 +1606,16 @@ subroutine radiation_tend(state,   ptend,    pbuf,          cam_out, cam_in,  &
...
-                        else
-                           next_month = month + 1
-                        end if
-                        allocate(aerosol_optical_property_lw(pcols,pver,nlwbands,begchunk:endchunk))
-                        call get_aerosol_optical_property_from_file(next_month,'AER_TAU_LW','lwband',nlwbands,aerosol_optical_property_lw)
-                        aer_tau_bnd_lw_mon_2 = aerosol_optical_property_lw(:,:,:,state%lchnk)
-                        deallocate(aerosol_optical_property_lw)
-                        current_month = month
+                     call pbuf_get_field(pbuf,aer_tau_bnd_lw_mon_idx, aer_tau_bnd_lw_mon)
+                     if (month==12) then
```
AaronDonahue, line 1610: I think you can move both instances of this `if` statement outside of the `if do_SPA_optics` call and clean up the code.  That being said, not really important.

AaronDonahue (commented):
Looking good, I still have a few more comments/questions.

components/eam/src/physics/cam/micro_p3_interface.F90
AaronDonahue: Oh nevermind, it looks like you are using the same approach.

components/scream/src/physics/p3/p3_main_impl.hpp
```diff
@@ -185,7 +185,7 @@ ::p3_main_part1(
       // prescribe that value
 
       if (do_prescribed_CCN) {
-         nc(k).set(not_drymass, max(nc(k), nccn_prescribed(k)));
+         nc(k).set(not_drymass, max(nc(k)*1.0e6*inv_rho(k), nccn_prescribed(k)));
```
AaronDonahue, line 188: @hassanbeydoun , the tests failed, but I think it is because this change doesn't match the F90 change.  Note the F90 change above is

```
nc(k) = max(nc(k),nccn_prescribed(k)*1.0e6*inv_rho(k))
```

There the `nccn_prescribed` is multiplied by the conversion factor, not `nc`

AaronDonahue:
Does this PR address issue #834 ?

components/scream/src/physics/p3/p3_main_impl.hpp
hassanbeydoun: Oops! Thanks for pointing this out Aaron. Fixed now.

components/eam/src/physics/cam/micro_p3_interface.F90
hassanbeydoun: Yes, I'm in the process of making these files available for all machines.

components/eam/src/physics/rrtmgp/radiation.F90
hassanbeydoun: But then it will get called regardless of whether it will be used or not?

hassanbeydoun:
> Does this PR address issue #834? or issue #835?

yes I believe it does.

components/eam/src/physics/rrtmgp/radiation.F90
AaronDonahue: oh wait, yes, you are correct.  I thought that this call was in both cases of an if statement, but now I see there are two separate if statements.  Ignore my suggestion :-)

hassanbeydoun:
> Good point because it validates my previous poor reading skills, @AaronDonahue ! Seriously though, I thought we were going to merge do_prescribed_CCN and do_SPA_optics into a single "do_SPA" flag once we had both completed?

We should keep the two flags for now just so we can continue to test out the different components of SPA separately.

PeterCaldwell (approved):
Looks good. Will need another PR to allow different file names for different horizontal resolutions and files for those different resolutions will need to be generated and put onto the ANL inputdata server.

PR #1010, opened 2021-04-20
Update the temperature conversion universal functions

AaronDonahue:
This PR updates the temperature universal conversion functions to support whole-column operations.

This PR updates the universal physics tests to take advantage of column-based operations.

This PR updates the pre and post amble for P3 to reflect the changes to the temperature conversion functions.

### DIFF
--- a/components/scream/src/physics/share/physics_functions.hpp
+++ b/components/scream/src/physics/share/physics_functions.hpp
@@ -114,8 +114,14 @@ struct Functions
   //   T_mid is the atmospheric temperature, units in K
   //   Exners function, unitless.  Exners function can be derived using `get_exner` defined above.
   KOKKOS_FUNCTION
-  static Spack T_to_th(const Spack& T_atm, const Spack& exner, const Smask& range_mask);
+  static Spack get_potential_temperature(const Spack& T_mid, const Spack& exner, const Smask& range_mask);
 
+  KOKKOS_FUNCTION
+  template <typename InputProvider>
+  static void get_potential_temperature(const int nlev,
+                                        const InputProvider& T_mid,
+                                        const InputProvider& exner,
+                                        const view_1d<Scalar>& T_potential);
   // Converts potential temperature into temperature
   // The result is the temperature, units in K
   // The inputs are
--- a/components/scream/src/physics/share/physics_functions.hpp
+++ b/components/scream/src/physics/share/physics_functions.hpp
@@ -153,7 +153,7 @@ struct Functions
   KOKKOS_FUNCTION
   static Spack get_dse(const Spack& T_mid, const Spack& z_mid, const Real surf_geopotential, const Smask& range_mask);
 
-  // Comput virtual temperature
+  // Compute virtual temperature
   // The result unit is in K
   // The inputs are
   //   T_mid is the atmospheric temperature.  Units in K.
--- a/components/scream/src/physics/share/physics_functions.hpp
+++ b/components/scream/src/physics/share/physics_functions.hpp
@@ -112,29 +112,29 @@ struct Functions
   // The result is the potential temperature, units in K
   // The inputs are
   //   T_mid is the atmospheric temperature, units in K
-  //   Exners function, unitless.  Exners function can be derived using `get_exner` defined above.
+  //   p_mid is the atmospheric pressure, units in Pa.  Used with Exners function using `get_exner` defined above.
   KOKKOS_FUNCTION
-  static Spack get_potential_temperature(const Spack& T_mid, const Spack& exner, const Smask& range_mask);
+  static Spack get_potential_temperature(const Spack& T_mid, const Spack& p_mid, const Smask& range_mask);
 
   KOKKOS_FUNCTION
   template <typename InputProvider>
   static void get_potential_temperature(const int nlev,
                                         const InputProvider& T_mid,
-                                        const InputProvider& exner,
+                                        const InputProvider& p_mid,
                                         const view_1d<Scalar>& T_potential);
   // Converts potential temperature into temperature
   // The result is the temperature, units in K
   // The inputs are
   //   Potential Temperature, units in K
-  //   Exners function, unitless.  Exners function can be derived using `get_exner` defined above.
+  //   p_mid is the atmospheric pressure, units in Pa.  Used with Exners function using `get_exner` defined above.
   KOKKOS_FUNCTION
-  static Spack get_potential_temperature_inv(const Spack& th_mid, const Spack& exner, const Smask& range_mask);
+  static Spack get_potential_temperature_inv(const Spack& th_mid, const Spack& p_mid, const Smask& range_mask);
 
   KOKKOS_FUNCTION
   template <typename InputProvider>
   static void get_potential_temperature_inv(const int nlev,
                                              const InputProvider& th_mid,
-                                             const InputProvider& exner,
+                                             const InputProvider& p_mid,
                                              const view_1d<Scalar>& T_mid);
   // Determine the physical thickness of a vertical layer
   // The result is dz, units in m
--- a/components/scream/src/physics/share/physics_functions.hpp
+++ b/components/scream/src/physics/share/physics_functions.hpp
@@ -160,6 +160,13 @@ struct Functions
   //   qv    is the water vapor mass mixing ratio.  Units in kg/kg
   KOKKOS_FUNCTION
   static Spack get_virtual_temperature(const Spack& T_mid, const Spack& qv, const Smask& range_mask);
+
+  KOKKOS_FUNCTION
+  template <typename InputProvider>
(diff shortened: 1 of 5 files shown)


AaronDonahue:
@bartgol , it looks like this is nearly there.  It is still failing on Weaver during the build, so there must be something GPU specific that is failing.  I am investigating at the moment.  In the meantime, it should be okay to review this for its structure, to ensure that this is the direction we want to go for revamping the universal physics functions.

components/scream/src/physics/share/physics_functions.hpp
```diff
@@ -112,18 +112,30 @@ struct Functions
   // The result is the potential temperature, units in K
   // The inputs are
   //   T_mid is the atmospheric temperature, units in K
-  //   Exners function, unitless.  Exners function can be derived using `get_exner` defined above.
+  //   p_mid is the atmospheric pressure, units in Pa.  Used with Exners function using `get_exner` defined above.
   KOKKOS_FUNCTION
-  static Spack T_to_th(const Spack& T_atm, const Spack& exner, const Smask& range_mask);
+  static Spack get_potential_temperature(const Spack& T_mid, const Spack& p_mid, const Smask& range_mask);
```
bartgol, line 117: For what is worth, I thought `T_to_th` was fine. Maybe something a tad more verbose, like `T_to_Theta`.

components/scream/src/physics/share/physics_functions.hpp
```diff
@@ -112,18 +112,30 @@ struct Functions
...
   //   Potential Temperature, units in K
-  //   Exners function, unitless.  Exners function can be derived using `get_exner` defined above.
+  //   p_mid is the atmospheric pressure, units in Pa.  Used with Exners function using `get_exner` defined above.
   KOKKOS_FUNCTION
-  static Spack th_to_T(const Spack& th_atm, const Spack& exner, const Smask& range_mask);
+  static Spack get_potential_temperature_inv(const Spack& th_mid, const Spack& p_mid, const Smask& range_mask);
 
+  KOKKOS_FUNCTION
+  template <typename InputProvider>
+  static void get_potential_temperature_inv(const int nlev,
```
bartgol, line 135: Personally, I'm not a fan of the fcn name. If nothing, b/c Theta is a fcn of two variables. How about `compute_temperature` or `Theta_to_T`?

components/scream/src/physics/share/physics_universal_impl.hpp
```diff
@@ -46,44 +46,75 @@ Functions<S,D>::get_exner(const Spack& P, const Smask& range_mask)
...
+template <typename S, typename D>
+KOKKOS_FUNCTION
+template <typename InputProvider>
+void Functions<S,D>::get_potential_temperature(const int nlev,
+                                               const InputProvider& T_mid,
+                                               const InputProvider& p_mid,
+                                               const view_1d<S>& T_potential)
+{
+  Kokkos::parallel_for("get_potential_temperature",nlev,[&](const int& ilev) {
+    T_potential[ilev] = get_potential_temperature(T_mid[ilev],p_mid[ilev],Smask(true))[0];
```
bartgol, line 79: I'm pretty sure you don't want the `[0]` at the end. Otherwise, you are stuffing the 1st entry of the result into all entries of the output pack.

components/scream/src/physics/share/physics_universal_impl.hpp
```diff
@@ -46,44 +46,75 @@ Functions<S,D>::get_exner(const Spack& P, const Smask& range_mask)
...
+template <typename S, typename D>
+KOKKOS_FUNCTION
+template <typename InputProvider>
+void Functions<S,D>::get_potential_temperature_inv(const int nlev,
+                                           const InputProvider& th_mid,
+                                           const InputProvider& p_mid,
+                                           const view_1d<Scalar>& T_mid)
+{
+  Kokkos::parallel_for("get_potential_temperature_inv",nlev,[&](const int& ilev) {
+    T_mid[ilev] = get_potential_temperature_inv(th_mid[ilev],p_mid[ilev],Smask(true))[0];
```
bartgol, line 115: Same here: you don't want the `[0]` at the end.

components/scream/src/physics/share/physics_universal_impl.hpp
```diff
@@ -153,6 +184,18 @@ Functions<S,D>::get_virtual_temperature(const Spack& T_mid, const Spack& qv, con
...
+template <typename S, typename D>
+KOKKOS_FUNCTION
+template <typename InputProvider>
+void Functions<S,D>::get_virtual_temperature(const int nlev,
+                                             const InputProvider& T_mid,
+                                             const InputProvider& qv,
+                                             const view_1d<S>& T_virtual)
+{
+  Kokkos::parallel_for("get_potential_temperature_inv",nlev,[&](const int& ilev) {
+    T_virtual[ilev] = get_virtual_temperature(T_mid[ilev],qv[ilev],Smask(true))[0];
```
bartgol, line 196: Here as well.

components/scream/src/physics/share/tests/physics_universal_unit_tests.cpp
```diff
@@ -248,22 +157,60 @@ struct UnitWrap::UnitTest<D>::TestUniversal
...
+      // Inputs:
+      //   temp is the atmospheric temperature. Units in K.
+      //   qv   is the atmospheric water vapor mass mixing ratio.  Units in kg/kg
+      //========================================================
+      physics::get_virtual_temperature(num_levs,temp,qv,T_virtual);
+      // Check the values from the tests above against expected values over the entire column
       for (int k=0;k<num_levs;++k)
       {
+        // temperature conversion check
+        Real exp_T_pot = physics::get_potential_temperature(temp(k),pressure(k),Smask(true))[0];
```
bartgol, line 195: I'm not sure if these tests are meaningful or not. It seems to me that all they are doing is serializing the num_levs loop, and comparing the output of the column version at level k with the single-level output. Even if the test passes, we are not testing much (namely, only that we wrote the TeamThreadRange correctly).

Imho, we should compute a whole column worth of solution, then do all checks on host, using directly REQUIRE (no reduction of nerr). The 'inv' functions can easily be tested by checking that their combo gives back the initial input. For each individual fcn we'd need to compare the output with the math formulation, or test some properties.

To avoid testing the result of a fcn against its mathematical formula, we could do stuff like testing that get_exner(p=p0) returns 1 (up to a tolerance). Or that get_exner(2*p)/get_exner(p) = 2^(rd/cp). But these might be as complicated as testing that get_exner(p)==(p/p0)^(rd/cp)...

In general, btw, we need to be aware that tolerance tests are bound to fail, no matter what. For every tol you pick, there will be some combination of the inputs that can cause ulp errors to cascade and cause an error>tol.

Note: in Homme we implemented a bfb pow, using taylor expansion. Notice that this is _not_ bfb in the sense that f90 matches cxx, but rather that the impl only requires +,-,*,/, which with math optimization disabled lead to bfb results. It worked well for Homme, and maybe we can apply it in scream too, if tol tests start to be a burden.

bartgol (changes requested):

components/scream/src/physics/share/physics_functions.hpp
AaronDonahue: I'm happy to change it back if there is consensus on the names.  I know that `theta` is the typical term for potential temperature but I thought it was a bit ambiguous in the context of a function name.  Hence my decision to explicitly state potential temperature.  Is there any opposition to an even more verbose name like `get_potential_temp_from_temp`?  Again, totally happy to keep `T_to_Theta` if that is likely to be clear to users.  @PeterCaldwell , any thoughts?

components/scream/src/physics/share/physics_functions.hpp
AaronDonahue: Similar response as the `T_to_Theta` discussion above.

bartgol: If we keep `get_potential_temperature`, which is fine, I would call this fcn `get_temperature`. The "_inv" suffix only makes sense for fcns of one variable only.

components/scream/src/physics/share/physics_functions.hpp
bartgol: We can keep `get_potential_temperature`.

PeterCaldwell: Oh jeez, asking Aaron "Verbose" Donahue to name functions could be dangerous. ;-). I like long names only to the extend that we need them to be self-describing. Another goal I have is to name things in a consistent way. So having `get_potential_temperature` to convert T to theta but `Th2T` to convert from theta to T would make me cry. I think `get_potential_temperature` is a fine name. `T_to_Th` is too brief and `get_potential_temp_from_temp` is too long.

AaronDonahue:
Handled in PR #1019

Issue #1008, opened 2021-04-19
Make field manager use wet mixing ratios, physics procs use dry mixing ratios as needed


PeterCaldwell:
Updated plan: we actually decided the field manager should always use *wet* mixing ratios because this makes more sense for weird variables like TKE that would be totally counterintuitive if they were per unit dry air. We want to have the field manager always do the same thing to keep code (and assumptions) simple. Most physics processes make more sense for dry mixing ratios, however. We will handle this by creating universal functions for converting between dry and wet and using these functions liberally in process interfaces.

OLD TEXT BELOW: 

> @bogensch , @brhillman , and @PeterCaldwell agree that it would be better to use mixing ratio (water mass divided by mass of dry air) rather than specific humidity (water mass divided by total mass). One reason for this is that a lot of empirical formulations used in P3 microphysics are based on mixing ratio rather than specific humidity. Another reason is that the variable q typically refers to a mixing ratio rather than specific humidity, so it is really confusing to use q as the core variable for moisture in our model and have it refer to specific humidity. Finally, mixing ratio is the more typical variable in the atm sci world. 
> 
> Note, however, that this is *different* from what EAM does - it assumes q is a specific humidity.
> 
> P3 and SHOC are already formulated assuming mixing ratio. In particular, our qsat (saturation calculation) already uses mixing ratio. RRTMGP uses total cell mass (qv*pres) as its water variable and doesn't have the interface to qv set up yet - so will need to perform the correct conversion.
> 
> My understanding is that HOMME operates on total mass qv*pres and converts to specific humidity for passing to physics. That is probably the only concrete task for this issue.

PeterCaldwell:
@mt5555 - I'm assigning you because I know you probably understand the issue. Feel free to delegate to someone else. @bartgol in particular maybe was the person who converted qdp to qv in hommexx so may be able to quickly fix this.

PeterCaldwell:
Also - I talk about qv above just for concreteness and convenience. I suspect this issue will also affect qc, qi, nc, ni, etc.

oksanaguba:
+1 for 'dry' water forms, would make relaxing const moist approximation easier.

mt5555:
You wrote SCREAM v0.1 above - but I think you mean SCREAM v1?

some initial comments:

[...]

Issue #998, opened 2021-04-15
Redo midpoints->interfaces interpolation utility in `scream_column_ops.hpp`


bartgol:
While the int->mid interpolation is well defined (average `f` at level interfaces, to obtain value of f at the level midpoints), the reverse direction is not. The current implementation "inverts" the int->mid interpolation rule, adding a boundary condition. It has the attractive property that the chained interpolation `mid->int->mid` is an identity.

However, this interpolation is not what Homme does for pressure, and using it for the pressure might violate surface boundary conditions in Homme. For more details on the topic, see [this confluence page](https://acme-climate.atlassian.net/wiki/spaces/NGDNA/pages/2579530246/NH+surface+pressure).

So for now, until we have a better idea of what properties the mid->int interpolation should preserve/satisfy, we should use a linear interpolation for "internal" interfaces, and fix b/c values at top/bottom. The linear interpolation is simply a convex combination of the midpoints on left/right of the interface, using the distances of the midpoints from the interface as weights. This translates to

```
Xi (k) = ( Xm(k)*h(k-1) + Xm(k-1)*h(k) ) / ( h(k)+h(k-1) )
```
where `h(k)` is the width of the k-th vertical level, and Xm/Xi are the vectors of values of quantity X at interfaces/midpoints respectively.

Note: this interpolation does _not_ verify `Xm(Xi(Xm))=Xm`, but it is a very reasonable linear interpolation strategy.

bartgol:
Note: we should probably _not_ delete the current "compatible" implementation, but mark it with a special name. Perhaps we could use the names `compute_interface_values_linear` and `compute_interface_values_compatible`, or something like that.

bartgol:
Removing the priority from this one, since we don't really know if we'll end up needing a mid->int routine.

bartgol:
@tcclevenger tagging you since you're working on it. Please, link this issue once a PR ready.

PR #1012, opened 2021-04-21
Compute and apply radiative heating

### LINKED ISSUE #953: Radiation needs to update temperature
The radiation interface needs to update the temperature. For the test problem, only the fluxes are computed, but before we couple with the rest of the model we need to add the calculation of the radiative heating terms and apply these to the temperature field.

brhillman:
Compute and apply radiative heating in the radiation interface code. This computes the atmosphere heating rate from the fluxes and applies the heating to the temperature field. This makes temperature a "computed" field in the radiation interface. Since radiation was not coupled with any other processes yet, this is BFB. Unit tests are added to test the routine added that computes the heating, and a very rudimentary test is added that verifies that temperature is updated. Fixes #953.

### DIFF
--- a/components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
+++ b/components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
@@ -1,6 +1,7 @@
 #include "ekat/ekat_assert.hpp"
 #include "physics/rrtmgp/scream_rrtmgp_interface.hpp"
 #include "physics/rrtmgp/atmosphere_radiation.hpp"
+#include "physics/rrtmgp/rrtmgp_heating_rate.hpp"
 #include "cpp/rrtmgp/mo_gas_concentrations.h"
 #include "YAKL/YAKL.h"
 
@@ -149,6 +150,22 @@ void RRTMGPRadiation::run_impl (const Real /* dt */) {
     sw_flux_up, sw_flux_dn, sw_flux_dn_dir,
     lw_flux_up, lw_flux_dn
   );
+
+  // Compute and apply heating rates
+  auto sw_heating  = real2d("sw_heating", m_ncol, m_nlay);
+  auto lw_heating  = real2d("lw_heating", m_ncol, m_nlay);
+  auto rad_heating = real2d("rad_heating", m_ncol, m_nlay);
+  rrtmgp::compute_heating_rate(
+    sw_flux_up, sw_flux_dn, p_lev, sw_heating
+  );
+  rrtmgp::compute_heating_rate(
+    lw_flux_up, lw_flux_dn, p_lev, lw_heating
+  );
+  parallel_for(Bounds<2>(m_nlay,m_ncol), YAKL_LAMBDA(int ilay, int icol) {
+    rad_heating(icol,ilay) = sw_heating(icol,ilay) + lw_heating(icol,ilay);
+    t_lay(icol,ilay) = t_lay(icol,ilay) + rad_heating(icol,ilay) * dt;
+  });
+
 }
 
 void RRTMGPRadiation::finalize_impl  () {
--- a/components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
+++ b/components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
@@ -51,6 +51,7 @@ void RRTMGPRadiation::set_grids(const std::shared_ptr<const GridsManager> grids_
   add_field<Required>("eff_radius_qi", scalar3d_layout_mid, micron, grid->name());
 
   // Set computed (output) fields
+  add_field<Computed>("T_mid"     , scalar3d_layout_mid, K  , grid->name());
   add_field<Computed>("SW_flux_dn", scalar3d_layout_int, Wm2, grid->name());
   add_field<Computed>("SW_flux_up", scalar3d_layout_int, Wm2, grid->name());
   add_field<Computed>("SW_flux_dn_dir", scalar3d_layout_int, Wm2, grid->name());
@@ -84,7 +85,6 @@ void RRTMGPRadiation::run_impl (const Real /* dt */) {
   // Get device views
   auto d_pmid = m_rrtmgp_fields_in.at("p_mid").get_reshaped_view<const Real**>();
   auto d_pint = m_rrtmgp_fields_in.at("pint").get_reshaped_view<const Real**>();
-  auto d_tmid = m_rrtmgp_fields_in.at("T_mid").get_reshaped_view<const Real**>();
   auto d_tint = m_rrtmgp_fields_in.at("tint").get_reshaped_view<const Real**>();
   auto d_gas_vmr = m_rrtmgp_fields_in.at("gas_vmr").get_reshaped_view<const Real***>();
   auto d_sfc_alb_dir = m_rrtmgp_fields_in.at("surf_alb_direct").get_reshaped_view<const Real**>();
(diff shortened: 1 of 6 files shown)


brhillman:
Failed on Weaver so labeling WIP until addressed.

components/scream/src/physics/rrtmgp/rrtmgp_heating_rate.hpp
```diff
@@ -0,0 +1,30 @@
...
+                yakl::Array<T,2,myMem,myStyle> const &plev   , yakl::Array<T,2,myMem,myStyle> &heating_rate
+            ) {
+            using physconst = scream::physics::Constants<Real>;
+            auto ncol = flux_up.dimension[0];
+            auto nlay = flux_up.dimension[1]-1;
+            parallel_for(Bounds<2>(nlay,ncol), YAKL_LAMBDA(int ilay, int icol) {
+                heating_rate(icol,ilay) = (
+                    flux_up(icol,ilay+1) - flux_up(icol,ilay) - 
+                    flux_dn(icol,ilay+1) + flux_dn(icol,ilay)
+                ) * physconst::gravit / (physconst::Cpair * (plev(icol,ilay+1) - plev(icol,ilay)));
```
AaronDonahue, line 25: Instead of approximating `dp` as the difference between the layer widths, could we use `pseudo_density` here?  The main reason I ask, is the current plan is to carry `pseudo_density` in the FM and back out `p_int` using a conversion routine.  

Slightly related question, does it matter if we are talking about wet pressure thickness, or dry pressure thickness?  There is currently an ongoing conversation, #1008 , that is examining these things so I want to make sure that anything we are doing with `dp` takes that discussion into account.

components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
```diff
@@ -51,6 +52,7 @@ void RRTMGPRadiation::set_grids(const std::shared_ptr<const GridsManager> grids_
   add_field<Required>("eff_radius_qi", scalar3d_layout_mid, micron, grid->name());
 
   // Set computed (output) fields
+  add_field<Computed>("T_mid"     , scalar3d_layout_mid, K  , grid->name());
```
AaronDonahue, line 55: There is now a new `add_field` type, `Updated` which stands for fields that are both required and computed.  Since `T_mid` is now both you could remove the `Required` declaration above and switch this to `Updated`.

AaronDonahue (changes requested):
Looks great Ben!  I had one requested change (see `Updated` comment) and one general question (see `dp` conversation).  After those we should be good to go.

components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
brhillman: Fixed, thanks for the tip!

components/scream/src/physics/rrtmgp/rrtmgp_heating_rate.hpp
brhillman: Hey @AaronDonahue, this was ported directly over from how we do things in SCREAMv0...since this seems to be an on-going conversation as to how we should handle this, i.e., start with making everything BFB as opposed to simultaneously changing both software and algorithms, I added a TODO comment above this describing the situation. I think at the least, maybe we decide to swap this over anyways, but putting it in this way minimizes the changes in this PR, and gives us something to compare against in a follow-up PR. Thoughts? I will happily swap this over to `dp` now though if you and others want...

AaronDonahue: Good point, we can probably continue to keep `p_int` in the FM for now and convert over to `dp` once we have a PR that can back `p_int` out from other variables.  We will probably need `homme` to make sure that `p_int` is updated somewhere when the pressure is changed.  Most likely in homme.  I think we can keep this as is then.

PeterCaldwell: Since the radiation interface is already not going to be BFB with SCREAMv0 (right?), why bother keeping p_int(i,k) - p_int(i,k) - which I think is actually wrong - instead of using pseudo_density like we want to?

Similarly (and not really related to this PR), why not delete p_int from the field manager and use whatever midpoint-to-interface code we have in universal functions to get T_int? Again, since we're already not BFB in rad interface?

brhillman: `p_int(i,k) - p_int(i,k)` is certainly wrong, it's `p_int(i,k+1) - p_int(i,k)`, which I believe is correct if we are using the hydrostatic pressure (which, if I'm not mistaken, is what was agreed to be provided in the field repo). Can you elaborate on why you think this is problematic? I'd be happy to change it here if you want, was just trying to minimize changes in this PR and relative to v0. Perhaps as a starting point we should change v0 to use `dp` as well.

PeterCaldwell: Sorry, I meant `p_int(i,k+1) - p_int(i,k)`. This *is* equal to dp and dp is equal to pseudo_density=rho*g*dz *if* we use hydrostatic pressure, but not for nonhydrostatic pressure (which we hope to transition to soon). Thus we are trying to convert all uses of `p_int(i,k+1) - p_int(i,k)` as a proxy for density into calls to pseudo_density now to avoid errors later. Does that make sense?

bartgol: FYI, Homme interface layer_does_ compute `p_int` (well, not in master yet, but will be soon). Regarding using dp vs computing p_int increments, they should be _exactly_ the same. At least if pack_size=1 (with larger pack sizes, the scan sum of dp to recover p_int is done in a way that exploits vector ops, so dp might not be bfb with p_int increments in that case).

Finally, we _do_ have a routine mid->int, but so far it's the version that has the oscillations issue (maybe we should purge the routine, so to avoid surprises when using it). So I don't think we want to rely on the general routine (yet). Issue #998 proposes a more "interpolatory" way of doing the conversion, but nobody has taken up the issue (I kind of put it on hold following our decision to put p_int in the FM, since other than radiation's T_int case there was no other need for this interpolation routine).

brhillman: Yeah, I get it, and I agree with the plan to move in that direction. My motivation for not doing it here was to minimize changes in a single PR, and also since we are not coupling with dynamics yet, I don't actually have a pseudo-density in the test problem we are running through the radiation interface, so I'll just have to compute one in the same way we do here but up further in the code in the test driver and pretend it came from somewhere meaningful. No problem though I'll make the change today and we can put it to rest. Of course, that does mean we are passing an additional field through the field repo to radiation, since I believe the current computation of `T_int` uses `p_int` for pressure weighting.

brhillman: Okay, this should be fixed to use `pseudo_density` now.

PeterCaldwell: Ok, so this PR will use pseudo_density instead of dp but will keep the current local T_int calculation using p_int rather than the mid->int universal function which might be subject to oscillations. That all sounds good.

brhillman:
Rebased.

components/scream/src/physics/rrtmgp/rrtmgp_heating_rate.hpp
bartgol: TBC, I was not advocating for changing the interpolation formula in this PR. I was just stating what was out there, and how different choices compared to each other. I fully support using p_int to do the in-house interpolation as it is done in v0.

AaronDonahue:
The weaver tests that are failing here match the tests failing for weaver on master.  So they are probably not related to the PR changes itself.  We'll have to fix those before this (and probably any other) PR will pass the AT.

AaronDonahue (approved):
Thanks for getting this done @brhillman

PeterCaldwell:
@AaronDonahue and @bartgol - should we manually merge this PR? Do we know what's wrong on weaver and have an ETA on a fix?

AaronDonahue:
@PeterCaldwell We don't know whats wrong with Weaver yet.  I noted this issue in my CDash update on Slack though.  There are a couple of AT issues outstanding that we probably want to look at.

PR #1015, opened 2021-04-22
SHOC updates

bogensch:
This PR contains various SHOC updates.  Specifically, this PR:

1. Brings SHOC’s tunable parameters out to the namelist for F90 version.  The default values remain hardcoded in the cpp version as that is not yet supported.  The default parameters are set the same in the F90 and cpp version to ensure that SHOC tests pass.
2. Minor SHOC adjustments/fixes.  Both F90 and cpp versions (and property tests) have been modified:
    1. Fixes an issue where the shear production of TKE was missing a coefficient, leading to abnormally large values of TKE in areas of even weak shear.  
    2. Revises the clipping for w3, which only really happens during spin-up, and sets to a default neutral value for safety measures.
    3. Removes the calculation of the overturning timescale in the length scale calculation and prescribes this values.  Because of this, two routines were removed (and their associated tests and boiler plate). 
3. Modifies the way the SHOC time step is set.  
    1. Similar to dynamics, the default SHOC time step is now set to -1, which tells SHOC to then set the SHOC time step equal to the macrophysics/microphysics time step (determined by dtime/cld_macmic_num_steps).
    2. The default SHOC time steps (which weren’t actually used in the model because they were not valid), are removed from the F90 SCREAM compsets.  All configurations will now set it equal to the macro/micro timestep, as intended, by default.
    3. The user is still allowed to specify the SHOC time step and the checks to make sure the time step is valid are simplified from the original nonsensical logic.  The requirement is that the SHOC time step should divide evenly into the macro/micro time step.  The previous version would force the time step to work, which was horrible from a provenance perspective.

SHOC run and compare is an expected fail since this PR is non B4B.


bogensch:
Ready for review.  The shoc run and compare failures on blake, weaver, mappy are all expected fails because this PR is non B4B.

mchinita:
Hi @bogensch, I see that you changed the eddy turnover timescale used to calculate SHOC's mixing length. On that note, I have another comment related to the mixing length calculation: in the calculation of L_inf in the subroutine compute_l_inf_shoc_length(), the integral is done from the surface to the top of the domain, however, TKE is not 0 above the PBL but rather 0.0004 m2/s2 (mintke). This produces a l_inf on the order of magnitude of 3 instead of 1 (using BOMEX as an example) and it impacts SHOC's mixing length accordingly. Is there a reason for this or should the integral stop near the PBL top since TKE ≠ 0 all the way to the top of the domain?

tcclevenger (approved):
Sorry for the late review, I completely let this slip my mind..

It looks good to me!

Edit: except for the conflicting files. Ping me when you rebase and I'll approve right away.

bogensch:
thanks @mchinita for the insightful comment.  Yes, I am planning on looking into this in the near future.

AaronDonahue (approved):
Approving, assuming you've checked that the shoc_run_and_cmp test and the shoc_tests work with these changes.  Adding/removing variables and keeping the F90 and C++ consistent can be tricky.

bogensch:
thanks @AaronDonahue .  The shoc_run_and_cmp test is an expected fail because this PR is not b4b.  However, yes, I spent a lot of time to ensure that all shoc_tests (b4b tests between f90 and cpp and property tests) all passed.

PeterCaldwell:
So it seems like Blake has a config error after yesterday's upgrade: error on Blake is "srun: error: Couldn't find the specified plugin name for mpi/mvapich looking at all files". @jgfouca - can you look at this?

I'm also confused why rrtmgp_tests_ut_np1_omp1 fails on weaver in release mode. This PR shouldn't touch rrtmgp, right? Is that test just randomly failing for other PRs?

@AaronDonahue and @bogensch - are you convinced enough based on previous test results that this PR can be integrated by hand (in which case I will)?

AaronDonahue:
Let me go ahead and run the test suite on all of the LC machines and report back.

AaronDonahue:
```
I'm also confused why rrtmgp_tests_ut_np1_omp1 fails on weaver in release mode. This PR shouldn't touch rrtmgp, right? Is that test just randomly failing for other PRs?
```

For this, the Weaver is failing for all the tests that have baselines, which includes the rrtmgp stand alone tests.  So that fail is not related to this PR.  See my Slack correspondence on the ngd-scream channel for more details on the issues with Weaver.

bogensch:
Yeah, I would say to hold off on merging until the rrtmgp test is good; that wasn't happening before i did an upstream merge.  @AaronDonahue I was just about to reconfirm the tests on LC; so no worries about doing this.  this is my problem.

bogensch:
@PeterCaldwell I am convinced based on previous test results (and with Aaron's intel that the rrtmgp test fail is not related to this PR) that this can be integrated.  I'll let you decide whether you want to wait for another of tests to start to see if Blake passes/fails as expected.

mchinita:
Hi all, I hope I'm not writing this in the wrong "place". My apologies, if I'm. Nevertheless, I think there is a small typo in L195 in shoc.F90.
At the moment, it reads:
  if (present(qw2tune_in)) length_fac=length_fac_in
Instead of:
   if (present(length_fac_in)) length_fac=length_fac_in
Right?

bogensch:
Thanks @mchinita, great find!  I will issue a PR to correct this very soon.

PR #1017, opened 2021-04-23
For Cori, add "module load perl5-extras"

ndkeen:
For Cori, add "module load perl5-extras".
This will find the missing perl module and allow cases to work again.
Same fix applied to e3sm master https://github.com/E3SM-Project/E3SM/pull/4243
As no cases are working on Cori thought it would be good to make a quick fix.
[bfb]


AaronDonahue (approved):
Looks great! Thanks Noel

PR #1019, opened 2021-04-23
Create a common physics functions structure

AaronDonahue:
This PR will commit a new common physics functions structure to handle most physics conversions functions needed in SCREAM, including,

geopotential calculations
temperature conversion functions
dry static energy calculation

The purpose of this work is to replace and expand the current universal physics functions in the code which are restrictive in their definitions.

Tests are updated with more robust property tests for each function.

We will, however, retain all functions currently in the code that are meant for use with the saturation adjustment calculation given that these functions are used throughout the code.  So adjusting their signature will be left to a future PR.

### DIFF
--- a/components/scream/src/physics/share/common_physics_impl.hpp
+++ b/components/scream/src/physics/share/common_physics_impl.hpp
@@ -148,9 +148,9 @@ template<typename DeviceT>
 template<typename ScalarT, typename InputProviderT, typename InputProviderQ>
 KOKKOS_FUNCTION
 void PhysicsFunctions<DeviceT>::get_virtual_temperature(const MemberType& team,
-                                                      const InputProviderT& T_mid,
-                                                      const InputProviderQ& qv,
-                                                      const view_1d<ScalarT>& T_virtual)
+                                                        const InputProviderT& T_mid,
+                                                        const InputProviderQ& qv,
+                                                        const view_1d<ScalarT>& T_virtual)
 {
   Kokkos::parallel_for(Kokkos::TeamThreadRange(team,T_virtual.extent(0)),
                        [&] (const int k) {
@@ -158,6 +158,40 @@ void PhysicsFunctions<DeviceT>::get_virtual_temperature(const MemberType& team,
   });
 }
 
+//-----------------------------------------------------------------------------------------------//
+// Compute dry static energy (DSE).
+// The result unit is in J/kg
+// The inputs are
+//   T_mid is the atmospheric temperature. Units in K.
+//   z_mid is the geopotential height above surface at midpoints. Units in m.
+//   surf_geopotential is the surface geopotential height. Units in m.
+template<typename DeviceT>
+template<typename ScalarT>
+KOKKOS_FUNCTION
+ScalarT PhysicsFunctions<DeviceT>::get_dse(const ScalarT& T_mid, const ScalarT& z_mid, const Real surf_geopotential)
+{
+  using C = scream::physics::Constants<ScalarT>;
+  static constexpr ScalarT cp (C::CP);
+  static constexpr ScalarT ggr(C::gravit);
+
+  ScalarT dse = cp*T_mid + ggr*z_mid + surf_geopotential;
+  return dse;
+}
+
+template<typename DeviceT>
+template<typename ScalarT, typename InputProviderT, typename InputProviderZ>
+KOKKOS_FUNCTION
+void PhysicsFunctions<DeviceT>::get_dse(const MemberType& team,
+                                        const InputProviderT& T_mid,
+                                        const InputProviderZ& z_mid,
+                                        const Real surf_geopotential,
+                                        const view_1d<ScalarT>& dse)
+{
+  Kokkos::parallel_for(Kokkos::TeamThreadRange(team,dse.extent(0)),
+                       [&] (const int k) {
+    dse(k) = get_dse(T_mid(k),z_mid(k),surf_geopotential); 
+  });
+}
 //-----------------------------------------------------------------------------------------------//
 } // namespace physics
 } // namespace scream
--- a/components/scream/src/physics/share/common_physics_impl.hpp
+++ b/components/scream/src/physics/share/common_physics_impl.hpp
@@ -32,7 +32,12 @@ ScalarT PhysicsFunctions<DeviceT>::get_exner(const ScalarT& p_mid)
   static constexpr ScalarT rd(C::RD);
   static constexpr ScalarT inv_cp(C::INV_CP);
 
-  return pow( p_mid/p0, rd*inv_cp );
+  ScalarT exner = pow( p_mid/p0, rd*inv_cp );
+  // Simple check to make sure no obvious errors occurred
+  EKAT_REQUIRE_MSG(!isnan(exner),"Error in get_exner, exner(p = " + std::to_string(p_mid) + ") = NaN"); 
+  EKAT_REQUIRE_MSG(!(exner<0),   "Error in get_exner, exner(p = " + std::to_string(p_mid) + ") = " + std::to_string(exner) + " < 0");
+  // Return exner 
+  return exner;
 }
 // For operations on a full column at a time:
 template<typename DeviceT>
@@ -42,14 +47,118 @@ void PhysicsFunctions<DeviceT>::get_exner(const MemberType& team,
                       const InputProviderP& p_mid, 
                       const view_1d<ScalarT>& exner)
 {
-
   Kokkos::parallel_for(Kokkos::TeamThreadRange(team,exner.extent(0)),
                        [&] (const int k) {
     exner(k) = get_exner(p_mid(k)); 
   });
-
 }
 
+//-----------------------------------------------------------------------------------------------//
+// Converts temperature to potential temperature using Exners function:
+//   theta = T_atm/exner(p_mid),
+// where
+//   theta  is the potential temperature, K
+//   T_mid  is the temperature, K
+//   p_mid  is the pressure, Pa.  Used for exners formula, see definition above, unitless
+template<typename DeviceT>
+template<typename ScalarT>
+KOKKOS_FUNCTION
+ScalarT PhysicsFunctions<DeviceT>::get_theta_from_T(const ScalarT& T_mid, const ScalarT& p_mid)
+{
+  ScalarT exner = get_exner(p_mid);
+  ScalarT theta = T_mid/exner;
+  // Simple check to make sure no obvious errors occurred
+  EKAT_REQUIRE_MSG(!isnan(theta),"Error in get_theta_from_T, theta(T = " + std::to_string(T_mid) + ", p = " + std::to_string(p_mid) + ") = NaN"); 
+  EKAT_REQUIRE_MSG(!(theta<0),"Error in get_theta_from_T, theta(T = " + std::to_string(T_mid) + ", p = " + std::to_string(p_mid) + ") = " + std::to_string(theta) + " < 0"); 
+  // Return theta
+  return theta;
+}
+// For operations on a full column at a time:
+template<typename DeviceT>
+template<typename ScalarT, typename InputProviderT, typename InputProviderP>
+KOKKOS_FUNCTION
+void PhysicsFunctions<DeviceT>::get_theta_from_T(const MemberType& team,
+                                                 const InputProviderT& T_mid,
+                                                 const InputProviderP& p_mid,
+                                                 const view_1d<ScalarT>& theta)
+{
+  Kokkos::parallel_for(Kokkos::TeamThreadRange(team,theta.extent(0)),
+                       [&] (const int k) {
+    theta(k) = get_theta_from_T(T_mid(k),p_mid(k)); 
+  });
+}
+
+//-----------------------------------------------------------------------------------------------//
+// Converts potential temperature to temperature using Exners function:
+//   T_mid = theta*exner(p_mid),
+// where
+//   theta  is the potential temperature, K
+//   T_mid  is the temperature, K
+//   p_mid  is the pressure, Pa.  Used for exners formula, see definition above, unitless
+template<typename DeviceT>
+template<typename ScalarT>
+KOKKOS_FUNCTION
+ScalarT PhysicsFunctions<DeviceT>::get_T_from_theta(const ScalarT& theta, const ScalarT& p_mid)
+{
+  ScalarT exner = get_exner(p_mid);
+  ScalarT T_mid = theta*exner;
+  // Simple check to make sure no obvious errors occurred
+  EKAT_REQUIRE_MSG(!isnan(T_mid),"Error in get_T_from_theta, T(theta = " + std::to_string(theta) + ", p = " + std::to_string(p_mid) + ") = NaN"); 
+  EKAT_REQUIRE_MSG(!(T_mid<0),"Error in get_T_from_theta, T(theta = " + std::to_string(theta) + ", p = " + std::to_string(p_mid) + ") = " + std::to_string(T_mid) + " < 0"); 
+  // Return T
+  return T_mid;
+}
+// For operations on a full column at a time:
+template<typename DeviceT>
+template<typename ScalarT, typename InputProviderT, typename InputProviderP>
+KOKKOS_FUNCTION
+void PhysicsFunctions<DeviceT>::get_T_from_theta(const MemberType& team,
+                                                 const InputProviderT& theta,
+                                                 const InputProviderP& p_mid,
+                                                 const view_1d<ScalarT>& T_mid)
+{
+  Kokkos::parallel_for(Kokkos::TeamThreadRange(team,T_mid.extent(0)),
+                       [&] (const int k) {
+    T_mid(k) = get_T_from_theta(theta(k),p_mid(k)); 
+  });
+}
+
+//-----------------------------------------------------------------------------------------------//
+  // Compute virtual temperature
+  // The result unit is in K
+  // The inputs are
+  //   T_mid is the atmospheric temperature.  Units in K.
+  //   qv    is the water vapor mass mixing ratio.  Units in kg/kg
+template<typename DeviceT>
+template<typename ScalarT>
+KOKKOS_FUNCTION
+ScalarT PhysicsFunctions<DeviceT>::get_virtual_temperature(const ScalarT& T_mid, const ScalarT& qv)
+{
+  using C = scream::physics::Constants<ScalarT>;
+  static constexpr ScalarT ep_2(C::ep_2);
+  ScalarT T_virtual = T_mid*(qv+ep_2)/(ep_2*(1.0+qv));
+  // Simple check to make sure no obvious errors occurred
+  EKAT_REQUIRE_MSG(!isnan(T_virtual),"Error in get_virtual_temperature, T_virtual(T = " + std::to_string(T_mid) + ", qv = " + std::to_string(qv) + ") = NaN"); 
+  EKAT_REQUIRE_MSG(!(T_virtual<0),"Error in get_virtual_temperature, T_virtual(T = " + std::to_string(T_mid) + ", qc = " + std::to_string(qv) + ") = " + std::to_string(T_virtual) + " < 0");
+  // Return T_virtual
+  return T_virtual;
+}
+
+template<typename DeviceT>
+template<typename ScalarT, typename InputProviderT, typename InputProviderQ>
+KOKKOS_FUNCTION
+void PhysicsFunctions<DeviceT>::get_virtual_temperature(const MemberType& team,
+                                                      const InputProviderT& T_mid,
+                                                      const InputProviderQ& qv,
+                                                      const view_1d<ScalarT>& T_virtual)
+{
+  Kokkos::parallel_for(Kokkos::TeamThreadRange(team,T_virtual.extent(0)),
+                       [&] (const int k) {
+    T_virtual(k) = get_virtual_temperature(T_mid(k),qv(k)); 
+  });
+}
+
+//-----------------------------------------------------------------------------------------------//
 } // namespace physics
 } // namespace scream
 
--- a/components/scream/src/physics/share/common_physics_impl.hpp
+++ b/components/scream/src/physics/share/common_physics_impl.hpp
@@ -193,6 +193,49 @@ void PhysicsFunctions<DeviceT>::get_dse(const MemberType& team,
   });
 }
 //-----------------------------------------------------------------------------------------------//
+// Determines the vertical layer thickness using the equation of state:
+//   dz = - (-psuedo_density)*Rd*T_virtual / (p_mid*g)
+//     note the extra negative sign because the psuedo_density in the model is measured in the positive direction.
+// where
+//   dz             is the vertical layer thickness, m
+//   psuedo_density is the pressure level thickness, Pa
+//   T_virtual      is the virtual temperature - calculated using a separate function from this suite, K
+//   p_mid          is the avgerage atmosphere pressure over the level, Pa
+//   g              is the graviational constant, m s-2
+//   Rd             is the universal gas constant for dry air, J/kg/K
+//   T_mid          is the atmospheric temperature, K - needed for T_virtual
+//   qv             is the water vapor mass mixing ratio, kg/kg - needed for T_virtual
+template<typename DeviceT>
+template<typename ScalarT>
+KOKKOS_FUNCTION
+ScalarT PhysicsFunctions<DeviceT>::get_dz(const ScalarT& psuedo_density, const ScalarT& p_mid, const ScalarT& T_mid, const ScalarT& qv)
+{
+  using C = scream::physics::Constants<ScalarT>;
+  static constexpr ScalarT Rd  (C::RD);
+  static constexpr ScalarT ggr (C::gravit);
+  // Need to first back out virtual temperature
+  ScalarT T_virtual = get_virtual_temperature(T_mid,qv);
+  // Now can back out the vertical layer thickness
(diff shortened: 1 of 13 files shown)


components/scream/src/physics/share/common_physics_impl.hpp
```diff
@@ -0,0 +1,165 @@
...
+ScalarT PhysicsFunctions<DeviceT>::get_exner(const ScalarT& p_mid)
+{
+  using C = scream::physics::Constants<ScalarT>;
+  static constexpr ScalarT p0(C::P0);
+  static constexpr ScalarT rd(C::RD);
+  static constexpr ScalarT inv_cp(C::INV_CP);
+
+  ScalarT exner = pow( p_mid/p0, rd*inv_cp );
+  // Simple check to make sure no obvious errors occurred
+  EKAT_REQUIRE_MSG(!isnan(exner),"Error in get_exner, exner(p = " + std::to_string(p_mid) + ") = NaN");
```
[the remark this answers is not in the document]
bartgol: Also, `isnan(exner)` will return a mask in case of packs. So the check here is tricky. What you wrote works for Real but not for Pack, but `isnan(exner).any()` works for Pack but not for Real.

We could either remove any check inside the fcns (and leaving the responsibility of checking input to the user _before_ calling the physics function), or we should add some utilities in ekat, that work with both Mask and bool. Something like

```
template<typename C>
bool any (const C& c);
template<typename C>
bool none (const C& c);
template<typename C>
bool all (const C& c);
```
which we specialize for `C=Mask` and `C=bool`.

components/scream/src/physics/share/common_physics_impl.hpp
```diff
@@ -0,0 +1,165 @@
...
+//   p_mid  is the pressure, Pa.  Used for exners formula, see definition above, unitless
+template<typename DeviceT>
+template<typename ScalarT>
+KOKKOS_FUNCTION
+ScalarT PhysicsFunctions<DeviceT>::get_theta_from_T(const ScalarT& T_mid, const ScalarT& p_mid)
+{
+  ScalarT exner = get_exner(p_mid);
+  ScalarT theta = T_mid/exner;
+  // Simple check to make sure no obvious errors occurred
+  EKAT_REQUIRE_MSG(!isnan(theta),"Error in get_theta_from_T, theta(T = " + std::to_string(T_mid) + ", p = " + std::to_string(p_mid) + ") = NaN");
```
bartgol, line 71: Here too: I find more insightful to check inputs than outputs.

bartgol: Also, `to_string` is not overloaded for packs.

bartgol (commented):
I think the design looks good. I wrote a few comments. You are probably already aware of some, but I wrote them anyways.

components/scream/src/physics/share/tests/physics_universal_unit_tests.cpp
```diff
@@ -178,105 +179,162 @@ struct UnitWrap::UnitTest<D>::TestUniversal
...
+      // Exner property tests:
+      // get_exner(p0) should return 1.0
+      // get_exner(0.0) should return 0.0
+      // get_exner(2*p0) should return 2**(rd/cp)
+      // get_exner(pressure) should work.
+      EKAT_REQUIRE_MSG(physicscommon::get_exner(p0)==1.0,"Error in get_exner(p0) property test");
+      EKAT_REQUIRE_MSG(physicscommon::get_exner(0.0)==0.0,"Error in get_exner(0.0) property test");
+      EKAT_REQUIRE_MSG(physicscommon::get_exner(2*p0)==pow(2.0,rd*inv_cp),"Error in get_exner(2*p0) property test");
+      EKAT_REQUIRE_MSG(physicscommon::get_exner(4.0)/physicscommon::get_exner(2.0)==pow(2.0,rd*inv_cp),"Error in get_exner(4)/get_exner(2) property test");
+      REQUIRE_NOTHROW(physicscommon::get_exner(team,pressure,exner));
```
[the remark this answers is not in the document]
AaronDonahue: Is there a better way to make sure the function actually works?  Maybe just do the call and of course if it throws and error it throws an error, and just remove the `REQUIRE_NOTHROW`?

components/scream/src/physics/share/common_physics_impl.hpp
AaronDonahue: In response to checking inputs rather than outputs.  My reasoning was that it would be helpful to make sure that the function doesn't return a NaN even if passed a reasonable set of inputs.  Think for example if you had a function that accepted `qc` as input, which could be 0.0 but some adjustment to the code accidentally divided by `qc` rather than multiplied.  The reasonable input of 0.0 would produce NaN as output which is precisely the type of thing we want to catch right?

AaronDonahue: But checking both input and output should be simple and satisfy both concerns

components/scream/src/physics/share/common_physics_impl.hpp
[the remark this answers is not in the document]
AaronDonahue: I'd vote for the EKAT option.  It may be that in this case we can just move the checks to outside the function, but I can forsee cases in the future where having in-function checks would be desirable.  Should I go ahead and work on that in EKAT?

components/scream/src/physics/share/common_physics_impl.hpp
AaronDonahue: > Also, `to_string` is not overloaded for packs.

Is this something we can overload in EKAT?  Would we even want to?  I guess in this case it isn't essential to provide all of the information I am adding into the ERROR message.

bartgol: I don't think this fcn should bother about output. As long as inputs are valid, the fcn is good to go. If the user has constraints on the output values, _the user_ should check those. The checks on the inputs are to verify that it is "legal" to call the fcn. And most likely they are enough to guarantee a valid output. But again, what "valid" means for the output, depends on how it is used. And in general, that depends on how it used later.

components/scream/src/physics/share/tests/physics_universal_unit_tests.cpp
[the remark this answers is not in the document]
bartgol: Yeah, if it throws, the code should already crash.

FWIW, I find `REQUIRE_NOTHROW` a bit pointless, in that regard (while `REQUIRE_THROW` is much more useful).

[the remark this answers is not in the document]
AaronDonahue: Right, I'm guessing the only usage would be that the test would report that an error was thrown on that requirement, which would be easier to track down then a case where the test errors out with something unintelligible.

AaronDonahue:
@bartgol this PR should be ready for review, granted it still needs to pas the AT which unfortunately has had some problems lately.

Two things to keep in mind while reviewing,
1. I'm not 100% committed to function names, feel free to suggest different names.
2. In your first read through you expressed hesitancy about the inline error checking (like making sure no NaN or negative temperature).  I didn't add any of these in all new functions but didn't delete the ones already there.  If you could advise on what is the proper path I can go ahead and streamline all functions to follow the same pattern.  i.e. no in-function checks, or only check inputs, etc...

AaronDonahue:
Pushed a rebased version to remove merge conflicts...

bartgol:
>     2. In your first read through you expressed hesitancy about the inline error checking (like making sure no NaN or negative temperature).  I didn't add any of these in all new functions but didn't delete the ones already there.  If you could advise on what is the proper path I can go ahead and streamline all functions to follow the same pattern.  i.e. no in-function checks, or only check inputs, etc...

I think we should keep the checks out of the impl, since a) they assume packs (making it hard to have a single impl templated on ScalarT), and b) they can be a burden if one simply wants to mask out the padding of the last level.

Regarding b, I am using get_exner in HommeDynamics, and I find myself having to create a mask just to prevent checks on the padding on the last level pack, while for all other packs the mask is true (for all pack entries).

Instead, I can check inputs _before_ calling the function, in a way that does not add any variable in Release builds:
```
  using PT = Pack<Real,N>;
  EKAT_REQUIRE_MSG ( (isnan(p) && range<PT>(ipack*N)<num_levs).none(),
    "Error! Pressure is nan.");
  const auto e = get_exner(p);
```
If I have to pass the range mask to get_exner, I have create the var (regardless of build type), and I doubt I can hope it gets optimized away.

AaronDonahue:
Okay great, thanks for the explanation Luca.  I'll remove the error checks from the impl.  I also now recall I need to move these files to the common share directory instead of leaving them in physics/share.  I'll rebase after the hot-fix goes in and make these changes too and commit a new version.

bartgol:
Yeah, with these in physics/share forced me to modify our CMakeLists.txt, to parse the physics folder _before_ the dynamics one, so that the target physics_share was available to link against. Moving the fcns in src/share will allow to remove this constraint, and dynamics can simply link against scream_share.

AaronDonahue:
@jeff-cohere , with Luca on vacation can I ask you to take a look and review this PR?

jeff-cohere:
> @jeff-cohere , with Luca on vacation can I ask you to take a look and review this PR?

Sure, I'll take a look tomorrow!

components/scream/src/physics/share/common_physics_impl.hpp
jeff-cohere: I agree with Luca about checking input only within the implementation of a function. However, I do think we should check function output in unit tests for the function--that's really where such checks belong, and will allow us to track bugs and fix them. To me, checking the output of a function given valid input blurs the boundary between model logic and tests. I do agree with the sentiment that we need to check output somewhere, though!

AaronDonahue: @jeff-cohere , yes.  I went with this suggestion for the latest version.  But it would be great for you to keep an eye on the route I took in case there is an even better way.  At the moment there are no checks inside the functions themselves.  Instead, the unit tests do all the typical checking of the function output.

components/scream/src/share/util/scream_column_ops.hpp
```diff
@@ -79,218 +87,102 @@ class ColumnOps {
...
-        team_parallel_for(team,NUM_MID_PACKS,shift_and_avg);
-      }
-    }
+    compute_midpoint_values_impl<CM,InputProvider,ScalarT>(team,num_mid_levels,x_i,x_m,alpha,beta);
   }
 
   // Compute X at level interfaces, given X at level midpoints and top or bot bc.
   // Note: with proper bc, then x_int(x_mid(x_int))==x_int.
-  template<bool FixTop, typename InputProvider = DefaultProvider>
+  template<bool FixTop, typename InputProvider, typename ScalarT>
```
jeff-cohere, line 142: Is it worth describing the FixTop template parameter and what it does, in a comment?

components/scream/src/share/util/scream_column_ops.hpp
```diff
@@ -79,218 +87,102 @@ class ColumnOps {
...
                           const InputProvider& x_i,
-                          const view_1d<pack_type>& dx_m,
+                          const view_1d<ScalarT>& dx_m,
+                          const scalar_type alpha = one(),
+                          const scalar_type beta = zero())
+  {
+    compute_midpoint_delta<CombineMode::Replace>(team,num_mid_levels,x_i,dx_m,alpha,beta);
+  }
+
+  // Given X at level interfaces, compute dX at level midpoints.
```
jeff-cohere, line 171: Maybe describe CombineMode and its supported options (unless they aren't a fixed set).

components/scream/src/share/util/scream_column_ops.hpp
```diff
@@ -607,6 +636,68 @@ class ColumnOps {
...
   }
 
+  // +--------------------------------------------+
+  // |        Kokkos::parallel_X wrappers         |
+  // +--------------------------------------------+
+
+  // Runs the input lambda with a TeamThreadRange parallel for over [0,count) range
+  template<typename Lambda>
+  KOKKOS_INLINE_FUNCTION
+  static void team_parallel_for (const TeamMember& team,
```
jeff-cohere, line 646: Question: why do we use the `static` keyword here? Is it to prevent this stuff from being used elsewhere? Is it needed in order to inline the function? If the former, we could place these functions into an anonymous namespace.

Not a big deal--just wondering about the reasoning.

components/scream/src/share/util/scream_common_physics_functions.hpp
```diff
@@ -0,0 +1,163 @@
...
+
+  template <typename S>
+  using view_1d = typename KT::template view_1d<S>;
+
+  //--------------------------------------------------------------------//
+  //  Functions
+  //--------------------------------------------------------------------//
+
+  // Computes exner function
+  // The result is exners formula, and is dimensionless
```
jeff-cohere, line 25: These comments are great. I like to use the square bracket convention for units, which seems to be used elsewhere in physics and earth science, and looks like this:

```
// The input is mid-level pressure [Pa]
```

jeff-cohere: Why `get_exner`, though? Why not just `exner`, given that this is already within the `PhysicsFunctions` struct/"namespace"?

components/scream/src/share/util/scream_common_physics_impl.hpp
```diff
@@ -0,0 +1,281 @@
+#ifndef COMMON_PHYSICS_IMPL_HPP
+#define COMMON_PHYSICS_IMPL_HPP
+
+#include "share/util/scream_common_physics_functions.hpp"
+#include "physics/share/physics_constants.hpp"
+
+namespace scream {
+  
+//-----------------------------------------------------------------------------------------------//
+// Applies Exners Function which follows:
```
jeff-cohere, line 10: Why not put these comments in the header describing the interface? None of them are related to the implementation of the function itself

components/scream/src/share/util/scream_common_physics_impl.hpp
```diff
@@ -0,0 +1,281 @@
...
+KOKKOS_FUNCTION
+ScalarT PhysicsFunctions<DeviceT>::get_dz(const ScalarT& psuedo_density, const ScalarT& p_mid, const ScalarT& T_mid, const ScalarT& qv)
+{
+  using C = scream::physics::Constants<Real>;
+  static const ScalarT Rd  (C::RD);
+  static const ScalarT ggr (C::gravit);
+  // Need to first back out virtual temperature
+  ScalarT T_virtual = get_virtual_temperature(T_mid,qv);
+  // Now can back out the vertical layer thickness
+  ScalarT dz = psuedo_density*Rd*T_virtual / (p_mid*ggr);
```
jeff-cohere, line 237: psuedo -> pseudo, btw

jeff-cohere (commented):
Things look pretty good to me. I have a few comments/questions. I'll hold off on approving till the testing system is in better shape, if you don't mind, @AaronDonahue

AaronDonahue:
> Things look pretty good to me. I have a few comments/questions. I'll hold off on approving till the testing system is in better shape, if you don't mind, @AaronDonahue

Thanks for the comments Jeff.  I don't mind holding off another day while testing gets sorted out.

AaronDonahue:
@jeff-cohere , I think I've addressed your comments in this latest version of the code.  The only comments I didn't address where for the changes in column_ops.  @bartgol made those changes in the code so I would want to give him a chance to respond.

components/scream/src/share/util/scream_column_ops.hpp
bartgol: In principle, we don't care about static here, since these methods are only used from within the class, not from outside. However, you cannot call a non-static method from a static one, and the main methods of the class are static.

jeff-cohere: Ahhhh, okay. I missed that these appeared in a class declaration(!). Wrong `static`. Thanks  Luca.

components/scream/src/share/tests/common_physics_functions_tests.cpp
```diff
@@ -0,0 +1,298 @@
...
+    ekat::genRandArray(temperature_packed,engine,pdf_temp);
+    ekat::genRandArray(height_packed,engine,pdf_height);
+    ekat::genRandArray(surface_height,engine,pdf_surface);
+    ekat::genRandArray(qv_packed,engine,pdf_qv);
+    ekat::genRandArray(pressure_packed,engine,pdf_pres);
+    ekat::genRandArray(pseudo_density_packed,engine,pdf_dp);
+
+    // Construct a simple set of `dz` values for testing the z_int function
+    auto dz_for_testing_host = Kokkos::create_mirror(dz_for_testing_packed);
+    for (int k = 0;k<num_levs;k++)
```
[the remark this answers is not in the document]
bartgol: Once you have the scalarized view, you could even do:
```
std::iota(dz_for_testing_host.data(),dz_for_testing_host.data()+num_levs,1); // gen sequence 1,2,...
std::reverse(dz_for_testing_host.data(),dz_for_testing_host.data()+num_levs); // revert sequence (b/c of atm lev ordering)
```
Here it's best to leave the manual for loop. I just thought I mentioned std library functions, in case you were not aware of those.

components/scream/src/share/tests/common_physics_functions_tests.cpp
```diff
@@ -0,0 +1,298 @@
...
+      // calculate_virtual_temperature(temperature,qv) should work for Scalar and for Spack
+      // calculate_temperature_from_virtual_temperature should work for Scalar and for Spack
+      physicscommon::calculate_virtual_temperature(team,temperature,qv,T_virtual);
+      physicscommon::calculate_temperature_from_virtual_temperature(team,T_virtual,qv,T_mid_from_virt);
+      physicscommon::calculate_virtual_temperature(team,temperature_packed,qv_packed,T_virtual_packed);
+      physicscommon::calculate_temperature_from_virtual_temperature(team,T_virtual_packed,qv_packed,T_mid_from_virt_packed);
+      Kokkos::parallel_for(Kokkos::TeamThreadRange(team,num_levs), [&] (const int k)
+      {
+        // T and TH conversion TEST
+        EKAT_KERNEL_REQUIRE(theta(k)==physicscommon::calculate_theta_from_T(temperature(k),pressure(k)));
```
bartgol, line 189: Since you are already doing it for other views, I would avoid doing checks on device. Instead, deep copy to host, and then use the REQUIRE macro on the host mirror.

bartgol (changes requested):
Mostly good. I have a few comments.

components/scream/src/share/util/scream_column_ops.hpp
bartgol: I wish C++ had not overloaded the static keyword...

components/scream/src/share/util/scream_common_physics_impl.hpp
```diff
@@ -0,0 +1,263 @@
...
+// The inputs are
+//   T_virtual is the virtual temperature.  Units in [K].
+//   qv        is the water vapor mass mixing ratio.  Units in [kg/kg]
+template<typename DeviceT>
+template<typename ScalarT>
+KOKKOS_FUNCTION
+ScalarT PhysicsFunctions<DeviceT>::calculate_temperature_from_virtual_temperature(const ScalarT& T_virtual, const ScalarT& qv)
+{
+  using C = scream::physics::Constants<Real>;
+  static const ScalarT ep_2(C::ep_2);
```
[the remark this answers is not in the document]
AaronDonahue: I originally had `ScalarT ep_2 = C:ep_2` but it doesn't work when ScalarT is a pack, since C::ep_2 is not of the same type.  This way, C::ep_2 is the constructor for ep_2 irregardless if ScalarT is a Real or a pack.

components/scream/src/share/util/scream_common_physics_impl.hpp
```diff
@@ -0,0 +1,263 @@
...
+
+namespace scream {
+  
+//-----------------------------------------------------------------------------------------------//
+// Computes exner function
+// The result is exners formula, and is [dimensionless]
+// The input is mid-level pressure, and has units of [Pa]
+template<typename DeviceT>
+template<typename ScalarT>
+KOKKOS_FUNCTION
```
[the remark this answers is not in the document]
AaronDonahue: Does this apply to all of the functions?

components/scream/src/share/util/scream_common_physics_impl.hpp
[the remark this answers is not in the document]
AaronDonahue: per `ep_2`: I believe this is because the variable is generally written as the greek letter epsilon, but that is already in use so this is a convenient shorthand.  In this case I am taking it to match the name in physics_constants.hpp.  I'm open to changing the name locally, or if there is agreement in the group changing it in physics_constants.hpp as well.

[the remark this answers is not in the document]
AaronDonahue: `static constexpr ScalarT ep_2 = C::ep_2;` this actually breaks in the build with an error 
```
error: expression must have a constant value
```

components/scream/src/share/tests/common_physics_functions_tests.cpp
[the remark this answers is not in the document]
AaronDonahue: So be "scalarizing" the view I should be able to replace `dz_for_testing_host(ipack)[ivec]` with `dz_for_testing_host[k]`?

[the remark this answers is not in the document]
bartgol: Yes.

AaronDonahue:
@bartgol , thank you for all of the comments.  I think I've addressed all of them either in the latest version of the branch or directly responding to your comments in github.  I didn't make the changes to the ColumnOps code.  I left that for you so that you can also respond to Jeff's comments, and you probably know better than I what you want to do/say.

components/scream/src/share/tests/common_physics_functions_tests.cpp
```diff
@@ -20,7 +20,7 @@ struct UnitWrap::UnitTest<D>::TestUniversal
   static void run()
   {
     using physicscommon      = scream::PhysicsFunctions<Device>;
-    using physicscommon_host = scream::PhysicsFunctions<HostDevice>;
+    using physicscommon = scream::PhysicsFunctions<HostDevice>;
```
bartgol, line 23: This line must be removed. You can't redefine `physicscommon` to be something else.

bartgol:
@AaronDonahue  I'm gonna update documentation, then push a final commit. So I'm gonna hold off the merge a tad longer.

AaronDonahue:
@jeff-cohere , would you mind taking another look at this PR when you have a chance?  It's been updated pretty heavily by Luca and myself.  It would be great to get your feedback on it.

jeff-cohere:
> @jeff-cohere , would you mind taking another look at this PR when you have a chance? It's been updated pretty heavily by Luca and myself. It would be great to get your feedback on it.

Of course! I'll do it this morning. I didn't want to do it at the end of yesterday because it looked pretty worked over.

AaronDonahue:
Awesome, thanks Jeff!

bartgol:
I removed myself as reviewer, since it's a conflict of interests...

components/scream/src/share/util/scream_column_ops.hpp
```diff
@@ -44,26 +45,42 @@ namespace scream {
...
+  using pack_type = ekat::Pack<scalar_type,PackSize>;
+
+  template<typename ScalarT>
+  static constexpr bool is_simd () {
+    return ekat::ScalarTraits<ScalarT>::is_simd;
+  }
+
+  template<typename ScalarT>
+  KOKKOS_FUNCTION
+  static constexpr int pack_size () {
```
jeff-cohere, line 80: Is this for computing the pack size of another type of Pack for easier conversion between packs of different types?

components/scream/src/share/tests/common_physics_functions_tests.cpp
```diff
@@ -0,0 +1,326 @@
...
+template<typename ScalarT, int NumLevels>
+struct ChecksHelpers {
+
+  static bool is_non_negative (const ScalarT& s, const int k) {
+    return not ( k<NumLevels && (s<0 || isnan(s)) );
+  }
+  static bool equal (const ScalarT& lhs, const ScalarT& rhs) {
+    return lhs==rhs;
+  }
+  static bool approx_equal (const ScalarT lhs, const ScalarT rhs,
```
jeff-cohere, line 22: Neither here nor there, but I've used `nearly_equal` for this type of function. Same number of letters without abbreviation.

components/scream/src/share/util/scream_column_ops.hpp
bartgol: Well, `ScalarT` may or may not be a Pack. You can call any of these methods with ScalarT=Real or ScalarT=Pack<Real,N>. So I needed a helper function that gave me "the number of Real's" in a ScalarT.

I think eventually I should change this, doing what I did in PhysicsFunctions, namely, do not template the class on the scalar type, but template only the individual functions. I think that might make the code a tad cleaner, but it might take some coding time, so I'm punting on this.

jeff-cohere: Good to know that `ScalarT` describes both actual scalars and packs. I think of packs as a "vector" type, so this wasn't obvious to me.

jeff-cohere (approved):
Looks great! I was able to follow your conversation and figure out your motivations in reworking the functions.

components/scream/src/share/util/scream_column_ops.hpp
bartgol: Yeah, tbh, I am not fond of how we use the term "scalar" in scream/ekat. I am thinking that `ValueType` is often more appropriate than `ScalarType`.

PR #1024, opened 2021-04-27
Fix all remaining compiler warnings

jgfouca:
Also, fix a valgrind error for column_ops_ut_np1_omp1


components/scream/src/share/util/scream_column_ops.hpp
```diff
@@ -426,7 +426,7 @@ class ColumnOps {
 
           // Note: for the last interface, this_pack_end==1, so we will *not* access
           //       dx_m(LAST_INT_PACK) (which would be OOB).
-          const auto this_pack_end = pack_info::vec_end(num_mid_levels,k);
+          const auto this_pack_end = pack_info::vec_end(num_mid_levels+1,k);
```
jgfouca, line 429: valg fix is here.

bartgol (approved):
Nice catch!

Issue #1020, opened 2021-04-23
exner is still incorrectly defined in p3


PeterCaldwell:
Matt Norman pointed out that https://github.com/E3SM-Project/scream/blob/90c85ecac8f6e1c482f06fb281613b41f7f85124/components/eam/src/physics/cam/micro_p3.F90#L452 and https://github.com/E3SM-Project/scream/blob/90c85ecac8f6e1c482f06fb281613b41f7f85124/components/eam/src/physics/cam/micro_p3.F90#L452 of micro_p3.F90 use inv_exner where the standard definition of exner=T/theta (e.g. https://en.wikipedia.org/wiki/Exner_function) suggests NOT the inverse of exner is needed.  I don't think there's a numerical error here because exner is also defined wrong in https://github.com/E3SM-Project/scream/blob/888c8fb92c4a5ed2301c03d7fc798b8789ced47e/components/eam/src/physics/cam/micro_p3_interface.F90#L1038 and everyone knows 2 wrongs make a right. We should fix this. Note that @AaronDonahue and I struggled with EAM misdefining exner in the past... It looks like this screw up managed to snare us yet again.

AaronDonahue:
On the plus side this error is not replicated in SCREAMv1 - see #1019 definition of exner for example.

AaronDonahue:
And just for reference, I remember Peter and I talking about this a lot in the early porting of P3.  And based on how exner is defined in the dynamics->physics coupling we must have adopted that approach, even though it is also wrong, https://github.com/E3SM-Project/scream/blob/888c8fb92c4a5ed2301c03d7fc798b8789ced47e/components/eam/src/dynamics/se/dp_coupling.F90#L555
I'm surprised that we went with this definition instead of the correct definition.  Oh well.  Good time to fix it though.

PeterCaldwell:
I recall @bartgol convincing us both that the dynamics definition was correct... though upon further review it is clearly the reciprocal of what it is supposed to be (noting in particular that cappa=+R/cp on https://github.com/E3SM-Project/scream/blob/888c8fb92c4a5ed2301c03d7fc798b8789ced47e/components/eam/src/control/physconst.F90#L89 . It's also weird that dynamics uses phys_state(lchnk)%pint(i,pver+1) as p0. I think this means that the actual surface pressure rather than 1000 mb is used in dynamics exner. That's definitely non-standard.

AaronDonahue:
[...]

PR #1027, opened 2021-04-27
Rework SHOC to use tracer transpose

tcclevenger:
This PR does 2 main things:
1. Changes SHOC to match the FM layout for tracers, `(num_col, num_tracer, num_levs)` (currently shoc treats tracers as fastest index).
2. Take all 2d local variables in SHOC from the workspace manager using the new function in ekat. This is applicable for both RHS of the tridiag solves in `update_prognostics_implicit`

PR #1028, opened 2021-04-27
Update exner variable names in P3 and SHOC

AaronDonahue:
This PR updates the variable names in P3 and SHOC to reflect the true formula, which is really the inverse of Exner's formula.

Addresses issue #1020.


AaronDonahue:
@PeterCaldwell and @bogensch, this is a simple PR to just update the name of the exner variable in the F90 version of P3 and SHOC.  It's a quick fix to issue #1020 , I could also make the PR more extensive by making the same cosmetic changes to the C++ code if desired.

It would probably be a good idea considering that the same confusion for what the variable `exner` represents applies to the C++ code and we want that code to be as clean as possible.  Let me know if you think I should do that in this PR.

tcclevenger:
@AaronDonahue Noticing the exact same weaver failure on my PR https://github.com/E3SM-Project/scream/pull/1027

bartgol:
> @AaronDonahue Noticing the exact same weaver and mappy failure on my PR #1027

@jgfouca and I are currently looking at it. Something is off with the test scripts, it seems. The error suggests the baselines dirs are not purged before starting to generate baselines, even though the script src code clearly has a line that should do it. Assuming this is not a fluke, we'll push a fix asap.

bogensch:
@AaronDonahue thanks for doing this.  I still can't believe my poor choice of variable names for "exner" (but sort of comforted/disturbed that I wasn't the only one that did this).  

I guess I would vote to make the same cosmetic changes in the cpp code in this PR to avoid the confusion in the future that this has caused us, unless there is a good reason for doing it later that is not obvious to me.

PeterCaldwell:
@AaronDonahue - Did you find all the exners to change by search/replacing everywhere in the p3 and shoc files in eam? 
I agree that we definitely want the C++ code to use the correct names for exner and we might as well do that in this PR since those name changes will also be BFB.

PeterCaldwell:
Comment on corrections in C++: I had originally thought that we defined Exner appropriately in C++ (which is why I didn't suggest we fix the C++ when we first discussed this PR). That appears to not be the case based on lines like `th_atm(k).set(drymass, th_atm(k) - exner(k) * qc(k) * latent_heat_vapor(k) * inv_cp);` - the exner term is converting the T from condensational heating into a theta tendency and exner=T/theta so we should have an inv_exner here.

AaronDonahue:
> Comment on corrections in C++: I had originally thought that we defined Exner appropriately in C++ (which is why I didn't suggest we fix the C++ when we first discussed this PR). That appears to not be the case based on lines like `th_atm(k).set(drymass, th_atm(k) - exner(k) * qc(k) * latent_heat_vapor(k) * inv_cp);` - the exner term is converting the T from condensational heating into a theta tendency and exner=T/theta so we should have an inv_exner here.

We are using exner appropriately in the SCREAMv1 code, but the inner-workings of the code are still the old way.  The easiest solution is to pass `inv_exner` to the main code but keep the incorrect names.  But as noted, it is way better and less confusing to just use the correct variable name everywhere.

I'll will plug away at that in between SCREAMv1 builds and let you both know when the PR is ready for re-review.

AaronDonahue:
@tcclevenger , you were mostly responsible for the SHOC port to C++ right?  If so, do you recall if you ever changed the usage of `exner` in the body of the main code?  I was looking through the C++ code and I found cases where it looked like exner was correctly formulated and other cases where it looks like it was used as `inv_exner`.  I want to update the C++ code to make it consistent with the F90 code but since there are so many C++ files to go through I want to make sure I'm doing it correctly from the beginning.

tcclevenger:
> @tcclevenger , you were mostly responsible for the SHOC port to C++ right? If so, do you recall if you ever changed the usage of `exner` in the body of the main code? I was looking through the C++ code and I found cases where it looked like exner was correctly formulated and other cases where it looks like it was used as `inv_exner`. I want to update the C++ code to make it consistent with the F90 code but since there are so many C++ files to go through I want to make sure I'm doing it correctly from the beginning.

It is only used 1 place in shoc that I see: `update_host_dse()`. There is used `y = x/exner;`

AaronDonahue:
@bogensch , @tcclevenger and @PeterCaldwell .  I've been working on this from time to time and I'm hesitant to update the shoc cpp code with the exner name change.  Mainly because I've noticed that exner is correctly defined and used in some cases, but in other cases we still call shoc_main from F90 which would presumably be using inv_exner instead of exner.  I think it might be a bit more of a tangled mess to audit the use of the exner variable name in this case.  One of you is welcome to take a look though, since you may be more familiar with the shoc code.

PeterCaldwell:
But @bogensch already approved it so I assume he thinks exner is used correctly in SHOC? Also, I'm impressed it actually passed all unit tests. Maybe AT is back in the game?

bogensch:
thanks @AaronDonahue, I will look into this.

AaronDonahue:
@bogensch and @PeterCaldwell , since I rebased this PR off of a newer version of master it needs to be reinspected before it can be merged.  Can both of you give one last quick look at the PR?

components/eam/src/physics/cam/micro_p3.F90
```diff
@@ -815,7 +815,7 @@ SUBROUTINE p3_main_part2(kts, kte, kbot, ktop, kdir, do_predict_nc, do_prescribe
 
       call prevent_ice_overdepletion(pres(k), t_atm(k), qv(k), latent_heat_sublim(k), inv_dt, qidep, qi2qv_sublim_tend)
 
-      call ice_supersat_conservation(qidep,qinuc,cld_frac_i(k),qv(k),qv_sat_i(k),latent_heat_sublim(k),th_atm(k)/exner(k),dt, &
+      call ice_supersat_conservation(qidep,qinuc,cld_frac_i(k),qv(k),qv_sat_i(k),latent_heat_sublim(k),th_atm(k)/inv_exner(k),dt, &
            qi2qv_sublim_tend, qr2qv_evap_tend)
```
PeterCaldwell, line 819: prob better to multiply by exner instead of dividing by inv_exner? ;-) . Though maybe you're trying to keep things BFB?

components/scream/src/physics/p3/p3_main_impl.hpp
```diff
@@ -618,7 +618,7 @@ ::p3_main_part2(
 
     prevent_ice_overdepletion(pres(k), T_atm(k), qv(k), latent_heat_sublim(k), inv_dt, qv2qi_vapdep_tend, qi2qv_sublim_tend, range_mask, not_skip_all);
 
-    ice_supersat_conservation(qv2qi_vapdep_tend,qv2qi_nucleat_tend,cld_frac_i(k),qv(k),qv_sat_i(k),latent_heat_sublim(k),th_atm(k)/exner(k),dt,qi2qv_sublim_tend,qr2qv_evap_tend, not_skip_all);
+    ice_supersat_conservation(qv2qi_vapdep_tend,qv2qi_nucleat_tend,cld_frac_i(k),qv(k),qv_sat_i(k),latent_heat_sublim(k),th_atm(k)/inv_exner(k),dt,qi2qv_sublim_tend,qr2qv_evap_tend, not_skip_all);
```
PeterCaldwell, line 622: As I commented above for F90, why not th*exner instead of th/inv_exner?

components/eam/src/physics/cam/micro_p3.F90
AaronDonahue: Agreed.  I can test to make sure switching to `*exner` is still BFB.  I think it would definitely make the code make more sense.

PeterCaldwell (approved):
Thanks Aaron. Lots of work here...

PR #1029, opened 2021-04-28
SCREAM: fix in test-all-scream

bartgol:
There was a problem with the creation of the baselines directories. The fcn 'crete_test_dirs' was called from the wrong place, leading to a wrong path. Testing on some PRs is currently failing because of this.


jgfouca (approved):

bartgol:
The weaver build failed cause jenkins got disconnected from the executor, but by manually inspecting the output stored on the executor, I was able to verify that tests passed. Since blake and mappy builds passed, I'm gonna merge this.

PR #1031, opened 2021-04-30
Hot-fix for namelist variables that caused SCREAMv0 to fail building

AaronDonahue:
Hot-fix for namelist variables that caused SCREAMv0 to fail during the build.

In particular, build-namelist could not find default namelist variables and
thus threw an error.  The namelist variables in question only had default values
assigned when microp_scheme=mg2.

These namelist variables were introduced during the upstream merge, and understandably
had not been tested when the microphysics scheme was P3 not MG2.

### DIFF
--- a/components/eam/bld/build-namelist
+++ b/components/eam/bld/build-namelist
@@ -813,8 +813,10 @@ add_default($nl,'use_hetfrz_classnuc');
 add_default($nl,'hist_hetfrz_classnuc');
 add_default($nl,'gw_convect_hcf');
 add_default($nl,'linoz_psc_t');
-add_default($nl,'micro_mg_dcs_tdep');
-add_default($nl,'micro_mincdnc');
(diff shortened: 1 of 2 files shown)


brhillman (approved):
Just noting that these should go upstream to E3SM too, but I think it's good to merge the hotfix now so that SCREAM works again before we do another upstream merge.

AaronDonahue:
> Just noting that these should go upstream to E3SM too, but I think it's good to merge the hotfix now so that SCREAM works again before we do another upstream merge.

Yes, good point @brhillman , the upstream PR has been submitted: https://github.com/E3SM-Project/E3SM/pull/4266

oksanaguba:
what's the difference between `-phys default` and `-phys cam5`?

AaronDonahue:
`cam5` is no-longer supported in E3SM

PR #1032, opened 2021-04-30
Small bugfixes from upstream-merge in EAM

AaronDonahue:
This commit fixes a couple of bugs that were introduced from the upstream-merge
that caused the build to throw an error.

In one case a function was declared twice from a use module statement.

In the other case, the keyword for the optional "abort" argument has
changed, which of course didn't touch two lines of code unique to the
SCREAM code base.


brhillman (approved):
Looks good to me. Thanks for finding these!

PR #1033, opened 2021-04-30
use ekat::scalarize more in SHOC

tcclevenger:
Adds a few instances of using `ekat::scalarize` instead of using pack index logic. Only an attempt to make the code more readable.


ambrad (approved):

jgfouca (approved):
These look like good changes that improve readability and should not impact performance.

jeff-cohere (approved):
These changes are great!

PR #1034, opened 2021-05-03
Bugfix to namelist defaults in EAM for HR compset with ne4

AaronDonahue:
This commit fixes a bug where the default ncdata entry for the case
when the ne4np4 grid is used with 128 levels was deleted.
This caused the build to fail when running the HR compsets at coarse resolution
because no default ncdata value was detected.

The entry for ne4np4 w/ 128 levels was inadvertantly deleted during an upstream
merge.

[BFB]


crterai (approved):
Looks good! Thanks, Aaron.

PR #1036, opened 2021-05-04
Add flag to register SPA variables with PBUF only when SPA is active

AaronDonahue:
This commit makes sure that PBUF variables intended for use only when SPA is active are only registered with the PBUF when SPA is active.

This remedies an issue where the out-of-the-box SCREAMv0 had difficulty
writing restart files given the size of the SPA PBUF variables footprint.

A subsequent PR will have to address the issue writing a restart when SPA
is active.

[BFB]


hassanbeydoun (approved):
Looks good. Thanks for quickly getting to this Aaron!

PR #1037, opened 2021-05-05
Add option to add mappy scream+cime testing via label

jgfouca:
The lable is CIME. Also, change the 'CI: Integrate Without Testing' to 'AT: Integrate Without Testing'


jgfouca:
Pinging @bartgol so he'll see this when he gets back from vacation.

bartgol (approved):
Nice feature!

PR #1038, opened 2021-05-05
Fix units for all physics processes so that they match.

AaronDonahue:
This commit updates the units for all of the field managed variables
in the src/physics directory so that they match.  This is needed for
1. Accuracy, some units were not correct which can lead to confusion.
2. Coupling between processes, the AD will throw an error if a field
   is registered with the field manager with more than one set of units.

There is also some cosmetic changes, like fixing indentation and changing the
name of some variables to match the names we have agreed upon in our
field managed variables group discussion.

[BFB]


AaronDonahue:
@tcclevenger and @brhillman , I put you two down as reviewers to check and make sure I didn't make any mistakes with units, and my cosmetic changes to variable names.

tcclevenger (commented):
@AaronDonahue Looks good to me. I'll let Ben approve so he has a chance to look at it before it merges.

AaronDonahue:
@brhillman , when you get a chance can you check the changes I've made to the RRTMGP interface and stand-alone test code to make sure it all looks good.  Then I think we can merge.

components/scream/tests/scream_shoc/CMakeLists.txt
```diff
@@ -8,5 +8,5 @@ CreateUnitTest(shoc_stand_alone "shoc_stand_alone.cpp" "${NEED_LIBS}" LABELS "sh
 # Copy yaml input file to run directory
 CONFIGURE_FILE(${CMAKE_CURRENT_SOURCE_DIR}/input.yaml
                ${CMAKE_CURRENT_BINARY_DIR}/input.yaml COPYONLY)
-CONFIGURE_FILE(${CMAKE_CURRENT_SOURCE_DIR}/shoc_standalone_ne4_ne4.nc
-               ${CMAKE_CURRENT_BINARY_DIR}/shoc_standalone_ne4_ne4.nc COPYONLY)
+CONFIGURE_FILE(${CMAKE_CURRENT_SOURCE_DIR}/shoc_standalone_ne4_np4.nc
```
bartgol, line 11: While we're at it, I think the e3sm convention is to _not_ use an underscore between neX and phys grid specs. That is, `blah_ne4np4` rather than `blah_ne4_np4`.

components/scream/src/physics/shoc/atmosphere_macrophysics.cpp
```diff
@@ -53,23 +53,23 @@ void SHOCMacrophysics::set_grids(const std::shared_ptr<const GridsManager> grids
...
-
-  add_field<Updated> ("T_mid",            scalar3d_layout_mid, nondim, grid_name, ps);
-  add_field<Updated> ("qv",               scalar3d_layout_mid, Qunit,  grid_name, "tracers", ps);
+  add_field<Required>("pref_mid",         pref_mid_layout,     Pa,      grid_name, ps);
+  add_field<Required>("z_int",            scalar3d_layout_int, m,       grid_name, ps);
+  add_field<Required>("z_mid",            scalar3d_layout_mid, m,       grid_name, ps);
+  add_field<Required>("omega",            scalar3d_layout_mid, Pa/s,    grid_name, ps);
+  add_field<Required>("surf_sens_flux",   scalar2d_layout_col, W/(m*m), grid_name, ps);
+  add_field<Required>("surf_latent_flux", scalar2d_layout_col, W/(m*m), grid_name, ps);
+  add_field<Required>("surf_u_mom_flux",  scalar2d_layout_col, N/(m*m), grid_name, ps);
```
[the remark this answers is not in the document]
AaronDonahue: This came from following the units we put down on the Confluence page for fields in the field manager.  Perhaps we should leave it as is for now and if we change the units in Confluence we can submit a quick PR to make the two consistent.

components/scream/tests/scream_shoc/CMakeLists.txt
AaronDonahue: Ahhh yes, good point.  I think we should make the change now and start the good habit.  I'll make the fix and push the changes so we can merge this PR.

AaronDonahue:
@bartgol , @tcclevenger and @brhillman .  This PR may need to be "re-reviewed" now that I've pushed a new commit.  But it should be ready to merge.

tcclevenger (approved):

PR #1039, opened 2021-05-06
Should not clean here, breaks AUTO baselines

jgfouca:
Do a full clean at the beginning of generate_all_baselines deletes all pre-existing baselines that we were going to use later in the main testing. There is already code in the __init__ to ensure the baseline directories exist.


jgfouca:
Pinging @bartgol so he sees this when he gets back.

jgfouca:
Standalone `scripts-tests` works. This is somewhat urgent so I'm forcing the merge.

PR #1043, opened 2021-05-10
Add support for building cprnc

jgfouca:
There has been interest in potentially using the CIME tool, cprnc, in SCREAM. The simplest way to offer it is to build it as an external. I little bit of hacky stuff is need to spoof the CIME Macros.cmake file which the cprnc CMake system expects.

### DIFF
--- a/components/scream/CMakeLists.txt
+++ b/components/scream/CMakeLists.txt
@@ -306,6 +306,8 @@ add_custom_target(baseline_cxx)
 add_subdirectory(src)
 if (NOT SCREAM_LIB_ONLY)
   add_subdirectory(tests)
+  include(BuildCprnc)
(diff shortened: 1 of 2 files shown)


jgfouca:
Getting this to work on Blake required fixing the nf-config used by blake. It was not reporting the correct -L lib dir and was also missing the -lnetcdff.

bartgol (approved):

jgfouca:
A similar fix was needed for the weaver netcdf installation.

bartgol:
Why was this not a problem so far?

jgfouca:
@bartgol , up until this PR, we only used `nf-config --prefix`, we were not using `nf-config --flibs`

PR #1044, opened 2021-05-11
SHOC namelist bug fix

bogensch:
Clears up inconsistency in the namelist settings for the length scale tuning factor.  b4b PR because the tuning parameters used in the tests are the default.


AaronDonahue (approved):
Looks good.

PR #1045, opened 2021-05-12
Using tracer data routines to read and interpolate SPA input files.

hassanbeydoun:
New method for reading in aerosol optical properties and CCN for SPA. 

Following namelist settings would invoke these changes:
```
spa_datapath = <path to the spa input file>
spa_file = <input file name>
spa_type = 'CYCLICAL'
spa_cycle_yr = 1
do_prescribed_ccn = .true.
```


components/eam/src/chemistry/utils/read_spa_data.F90
```diff
@@ -0,0 +1,93 @@
+module read_spa_data
+
+!BSINGH - I didn't add any comments here due to time constrain but I will add comments later
+
+use tracer_data
```
singhbalwinder, line 5: Use `only` with the `use` statement

components/eam/src/chemistry/utils/read_spa_data.F90
```diff
@@ -0,0 +1,93 @@
...
+private
+
+public :: read_spa_data_init, read_spa_data_adv, read_spa_data_register
+
+type(trfld), pointer :: spa_fields(:)
+type(trfile)         :: spa_file
+logical              :: rmv_file = .false.
+
+!fields to read from the file
+character(len=16), parameter :: pbuf_names(59) = &
```
singhbalwinder, line 18: May be split these apart and use these as in the radiation code and then combine them in one array. This will save us from writing these all over again in the radiation code and it will be easy to maintain.

Also, use a parameter variable to store `59`

components/eam/src/chemistry/utils/read_spa_data.F90
```diff
@@ -0,0 +1,93 @@
...
+       'AER_TAU_LW_0 ', 'AER_TAU_LW_1 ', 'AER_TAU_LW_2 ', 'AER_TAU_LW_3 ', 'AER_TAU_LW_4 ', &
+       'AER_TAU_LW_5 ', 'AER_TAU_LW_6 ', 'AER_TAU_LW_7 ', 'AER_TAU_LW_8 ', 'AER_TAU_LW_9 ', &
+       'AER_TAU_LW_10', 'AER_TAU_LW_11', 'AER_TAU_LW_12', 'AER_TAU_LW_13', 'AER_TAU_LW_14', &
+       'AER_TAU_LW_15', &
+       'AER_TAU_SW_0 ', 'AER_TAU_SW_1 ', 'AER_TAU_SW_2 ', 'AER_TAU_SW_3 ', 'AER_TAU_SW_4 ', &
+       'AER_TAU_SW_5 ', 'AER_TAU_SW_6 ', 'AER_TAU_SW_7 ', 'AER_TAU_SW_8 ', 'AER_TAU_SW_9 ', &
+       'AER_TAU_SW_10', 'AER_TAU_SW_11', 'AER_TAU_SW_12','AER_TAU_SW_13', &
+       'CCN3         ']
+
+character(len=16), parameter :: specifier(59) = pbuf_names(59)
```
singhbalwinder, line 34: May be get rid of `pbuf_names` variable and use `specifier` only everywhere or vice versa.

components/eam/src/chemistry/utils/read_spa_data.F90
```diff
@@ -0,0 +1,93 @@
...
+  endsubroutine read_spa_data_register
+
+
+  subroutine read_spa_data_init
+
+    !Tracer data routine init
+    allocate (spa_file%in_pbuf(size(specifier)))
+    spa_file%in_pbuf(:) = .true.
+
+!    call trcdata_init( specifier, 'unfied_SPA_file_lat_lon.nc', '', '/compyfs/sing201/lat_lon', spa_fields, spa_file, &
```
singhbalwinder, line 61: Handle at least the file name and path using the namelist. Explicitly specify rest of the arguments and what they represent.

components/eam/src/chemistry/utils/read_spa_data.F90
```diff
@@ -0,0 +1,93 @@
...
+         rmv_file, 1, 0, 0, 'CYCLICAL')
+
+
+  end subroutine read_spa_data_init
+
+
+  subroutine read_spa_data_adv( state, pbuf2d )
+
+    !advance fields in time and interpolate (space and time)
+    use tracer_data,  only : advance_trcdata
```
singhbalwinder, line 73: remove variables/methods which are not used.

components/eam/src/physics/cam/micro_p3_interface.F90
```diff
@@ -420,6 +421,9 @@ subroutine micro_p3_init(pbuf2d)
     prec_pcw_idx = pbuf_get_index('PREC_PCW') !! from physpkg 
     snow_pcw_idx = pbuf_get_index('SNOW_PCW') !! from physpkg 
 
+    !BSINGH- Get indices for SPA treatment
```
singhbalwinder, line 424: remove `BSINGH`

components/eam/src/physics/cam/physpkg.F90
```diff
@@ -274,6 +275,7 @@ subroutine phys_register
        call prescribed_volcaero_register()
        call prescribed_ozone_register()
        call prescribed_aero_register()
+       call read_spa_data_register()
```
singhbalwinder, line 278: call only if SPA is active

components/eam/src/physics/cam/physpkg.F90
```diff
@@ -809,6 +812,7 @@ subroutine phys_init( phys_state, phys_tend, pbuf2d, cam_out )
     call prescribed_ozone_init()
     call prescribed_ghg_init()
     call prescribed_aero_init()
+    call read_spa_data_init()
```
singhbalwinder, line 815: call only if SPA is active

components/eam/src/physics/cam/physpkg.F90
```diff
@@ -2785,6 +2790,7 @@ subroutine phys_timestep_init(phys_state, cam_out, pbuf2d)
   call prescribed_ozone_adv(phys_state, pbuf2d)
   call prescribed_ghg_adv(phys_state, pbuf2d)
   call prescribed_aero_adv(phys_state, pbuf2d)
+  call read_spa_data_adv(phys_state, pbuf2d)
```
singhbalwinder, line 2793: call only if SPA is active

components/eam/src/physics/rrtmgp/radiation.F90
```diff
@@ -203,8 +203,30 @@ module radiation
    integer :: cldfsnow_idx = 0
 
    !needed for SPA
-   integer :: aer_tau_bnd_lw_mon_idx, aer_tau_bnd_sw_mon_idx,&
-              aer_ssa_bnd_sw_mon_idx, aer_asm_bnd_sw_mon_idx 
+   integer :: aer_tau_bnd_lw_idx(nlwbands), aer_tau_bnd_sw_idx(nswbands),&
+              aer_ssa_bnd_sw_idx(nswbands), aer_asm_bnd_sw_idx(nswbands)
+   !array of SPA pbuf indices
+   character(len=16), parameter :: aer_asm_sw_names(14) = &
```
singhbalwinder, line 209: We can use these arrays in read_spa_data.F90, instead of specifying these fields again there. Use parameter variables instead of numeric literals.

singhbalwinder (approved):
I added these comments for myself. I will address these in the future if we need to merge this PR ASAP.

components/eam/src/chemistry/utils/read_spa_data.F90
PeterCaldwell: Maybe this is just what you're saying, but it looks to me like you created a char vector called pbuf_names then created an identical copy of it called specifier. Both are parameter types so can't be changed, so why do that?

components/eam/src/chemistry/utils/read_spa_data.F90
```diff
@@ -0,0 +1,93 @@
...
+
+implicit none
+
+private
+
+public :: read_spa_data_init, read_spa_data_adv, read_spa_data_register
+
+type(trfld), pointer :: spa_fields(:)
+type(trfile)         :: spa_file
+logical              :: rmv_file = .false.
```
PeterCaldwell, line 15: It would be good to include a comment here saying what rmv_file means. I suspect it means "don't delete the file after pulling its data"?

components/eam/src/chemistry/utils/read_spa_data.F90
```diff
@@ -0,0 +1,93 @@
...
+
+  subroutine read_spa_data_init
+
+    !Tracer data routine init
+    allocate (spa_file%in_pbuf(size(specifier)))
+    spa_file%in_pbuf(:) = .true.
+
+!    call trcdata_init( specifier, 'unfied_SPA_file_lat_lon.nc', '', '/compyfs/sing201/lat_lon', spa_fields, spa_file, &
+!         rmv_file, 1, 0, 0, 'CYCLICAL')
+    call trcdata_init( specifier, 'unfied_SPA_file_lat_lon.nc', '', '/global/cscratch1/sd/bsingh/users/hassan', spa_fields, spa_file, &
```
PeterCaldwell, line 63: super minor comment, but it looks like there's a typo in the file name: unfied instead of unified?

components/eam/src/chemistry/utils/read_spa_data.F90
PeterCaldwell: I think you just need only: trfld,trfile at the module level since individual functions have their own `use tracer_data, only` commands.

components/eam/src/physics/cam/physpkg.F90
PeterCaldwell: I think CCN3 is only being defined in pbuf if SPA is active, but is being grabbed all the time. Probably need `if SPA` conditional around these other calls too?

components/eam/src/physics/cam/physpkg.F90
PeterCaldwell: I agree, but I'm surprised that the other function calls here aren't within if statements. Maybe the typical paradigm is to always call but to return without doing anything if that condition isn't met?

components/eam/src/physics/rrtmgp/radiation.F90
```diff
@@ -976,26 +990,15 @@ subroutine radiation_init(state,pbuf)
...
-         do month = 1,12
-            call get_aerosol_optical_property_from_file(month,'AER_TAU_LW','lwband',nlwbands,aerosol_optical_property_lw(:,:,:,month,:))
-         end do
-         call pbuf_set_field(pbuf,aer_tau_bnd_lw_mon_idx,aerosol_optical_property_lw)
-         do month = 1,12 
-            call get_aerosol_optical_property_from_file(month,'AER_TAU_SW','swband',nswbands,aerosol_optical_property_sw(:,:,:,month,:))
+         do band_index = 0,nswbands
+            aer_tau_bnd_sw_idx(band_index) = pbuf_get_index(aer_tau_sw_names(band_index))
+            aer_ssa_bnd_sw_idx(band_index) = pbuf_get_index(aer_ssa_sw_names(band_index))  
+            aer_asm_bnd_sw_idx(band_index) = pbuf_get_index(aer_asm_sw_names(band_index))
```
PeterCaldwell, line 996: Stuffing the 4d vector needed by rad with the pointers to the 3d vectors available from tracer_data is really slick. We should make sure it doesn't cause big performance issues due to weird striding. I don't think it will since we're pulling from contiguous assignments to pbuf, but just in case...

PeterCaldwell:
Looks good. I suspect this PR will break non-SPA runs because it tries to grab pbuf items only defined when SPA is active. I think the basic approach is sound and cleanup shouldn't take too long.

components/eam/src/chemistry/utils/read_spa_data.F90
PeterCaldwell: I think this comment has been implemented, right?

singhbalwinder: yes, I removed all variables/methods which are not used.

PeterCaldwell (approved):

PeterCaldwell (approved):

PR #1046, opened 2021-05-12
Fix last valgrind error in homme standalone test

bartgol:
The conversion from C to F90 string of the namelist filename in homme_params_mod.F90 was not done correctly.
The correct version was copied from convert_c_string from [here](https://github.com/E3SM-Project/scream/blob/a8af66be21d47eaa8df60b9ccc01b221cd653774/components/scream/src/share/io/scream_scorpio_interface_iso_c2f.F90#L209).


jgfouca (approved):

jgfouca:
Very nice! Can't wait to have a clean SCREAM dashboard!

PR #1047, opened 2021-05-12
add data copy to radiation-ad

tcclevenger:
In `atmosphere_radiation.cpp` and `rrtmgp_stand_alone.cpp`, copy data to and from YAKL arrays. Previously, no copy was being performed, and (since YAKL arrays there were assuming `ncol` as the fastest index) the Kokkos views in the FM had incorrect entries.

A consequence of this PR is that the FM can now contain padded views.

This PR also updates the RRTMGP submodule and corrects calls to RRTMGP to have correct inputs based on the updates.

### DIFF
--- a/components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
+++ b/components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
@@ -7,6 +7,10 @@
 
 namespace scream {
 
+  using KT = KokkosTypes<DefaultDevice>;
+  using ExeSpace = KT::ExeSpace;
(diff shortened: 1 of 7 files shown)


tcclevenger:
@brhillman Can you take a look?

AaronDonahue (commented):
Looks good to me, I'll let @brhillman officially approve so the AT doesn't merge before he can get eyes on it.

brhillman (approved):
Sorry for the delay. This looks good to me. At some point we might want to encapsulate the copies in a separate routine, but no need to do that now.

PR #1048, opened 2021-05-12
add memory buffer to AD

tcclevenger:
Adds a AD-wide memory buffer which each process can request an allocation.

Changes:
- Adds memory buffer manager struct: `ATMBufferManager`.
- Each ATM process can request memory (in bytes) and initialize any local variables through the buffer manager. This is called for each process during the initialization phase in `atmosphere_process_group.cpp`.
- `WorkspaceManager` memory is also included in the request. The WSM variable is changed to be an input into the `_main()` functions instead of constructed internally

### DIFF
--- a/components/scream/src/physics/p3/atmosphere_microphysics.cpp
+++ b/components/scream/src/physics/p3/atmosphere_microphysics.cpp
@@ -100,9 +100,13 @@ int P3Microphysics::requested_buffer_size_in_bytes() const
   const Int nk_pack = ekat::npack<Spack>(m_num_levs);
 
   // Number of Reals needed by local views in the interface
-  const int interface_request = Buffer::num_1d_scalar*m_num_cols*sizeof(Real) +
-                                Buffer::num_2d_vector*m_num_cols*nk_pack*sizeof(Spack) +
-                                Buffer::num_2d_scalar*m_num_cols*3*sizeof(Real);
+  const int interface_request =
+      // 1d view scalar, size (ncol)
+      Buffer::num_1d_scalar*m_num_cols*sizeof(Real) +
+      // 2d view packed, size (ncol, nlev_packs)
+      Buffer::num_2d_vector*m_num_cols*nk_pack*sizeof(Spack) +
+      // 2d view scalar, size (ncol, 3)
+      m_num_cols*3*sizeof(Real);
 
   // Number of Reals needed by the WorkspaceManager passed to p3_main
   const auto policy       = ekat::ExeSpaceUtils<KT::ExeSpace>::get_default_team_policy(m_num_cols, nk_pack);
--- a/components/scream/src/physics/p3/atmosphere_microphysics.cpp
+++ b/components/scream/src/physics/p3/atmosphere_microphysics.cpp
@@ -95,6 +95,79 @@ void P3Microphysics::set_grids(const std::shared_ptr<const GridsManager> grids_m
 }
 
 // =========================================================================================
+int P3Microphysics::requested_buffer_size_in_bytes() const
+{
+  const Int nk_pack = ekat::npack<Spack>(m_num_levs);
+
+  // Number of Reals needed by local views in the interface
+  const int interface_request = Buffer::num_1d_scalar*m_num_cols*sizeof(Real) +
+                                Buffer::num_2d_vector*m_num_cols*nk_pack*sizeof(Spack) +
+                                Buffer::num_2d_scalar*m_num_cols*3*sizeof(Real);
+
+  // Number of Reals needed by the WorkspaceManager passed to p3_main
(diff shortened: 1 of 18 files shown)


components/scream/src/physics/p3/atmosphere_microphysics.cpp
```diff
@@ -94,6 +94,79 @@ void P3Microphysics::set_grids(const std::shared_ptr<const GridsManager> grids_m
...
 
+// =========================================================================================
+int P3Microphysics::requested_buffer_size_in_bytes() const
+{
+  const Int nk_pack = ekat::npack<Spack>(m_num_levs);
+
+  // Number of Reals needed by local views in the interface
+  const int interface_request = Buffer::num_1d_scalar*m_num_cols*sizeof(Real) +
+                                Buffer::num_2d_vector*m_num_cols*nk_pack*sizeof(Spack) +
+                                Buffer::num_2d_scalar*m_num_cols*3*sizeof(Real);
```
bartgol, line 105: It seems 1d are (ncols) quantities, and 2d  are (ncols,nlevs). And scalar vs vector is to say non-packed vs packed? But then, why are 2d scalars of size ncols*3?

components/scream/src/physics/p3/atmosphere_microphysics.hpp
```diff
@@ -206,6 +212,29 @@ class P3Microphysics : public AtmosphereProcess
   }; // p3_postamble
   /* --------------------------------------------------------------------------------------------*/
 
+  // Structure for storing local variables initialized using the ATMBufferManager
+  struct Buffer {
```
bartgol, line 216: Food for thought: later, we might see if it's beneficial to make a Buffer struct storing values within a column, in order to achieve better data locality. Then store a view of Buffer instead of a single Buffer.

components/scream/src/physics/p3/atmosphere_microphysics.cpp
tcclevenger: Yeah, this was a random local variable called `col_location`. It's the only 2d local scalar view. I'll change this to not be under the generic category `num_2d_scalar`.

components/scream/src/physics/p3/atmosphere_microphysics.cpp
```diff
@@ -94,6 +94,83 @@ void P3Microphysics::set_grids(const std::shared_ptr<const GridsManager> grids_m
...
+  const Int nk_pack = ekat::npack<Spack>(m_num_levs);
+
+  // Number of Reals needed by local views in the interface
+  const int interface_request =
+      // 1d view scalar, size (ncol)
+      Buffer::num_1d_scalar*m_num_cols*sizeof(Real) +
+      // 2d view packed, size (ncol, nlev_packs)
+      Buffer::num_2d_vector*m_num_cols*nk_pack*sizeof(Spack) +
+      // 2d view scalar, size (ncol, 3)
+      m_num_cols*3*sizeof(Real);
```
jeff-cohere, line 109: So there are 3 2D scalars? Is it worth adding a `Buffer::num_2d_scalar` to reflect this? Or am I misunderstanding?

jeff-cohere (approved):
This looks good to me. Approved (subject to passing tests, etc).

components/scream/src/physics/p3/atmosphere_microphysics.hpp
tcclevenger: I'm guessing this would involve passing the buffer to `shoc_main()` instead of aliasing input/output views in `SHOCMacrophysics::initialize_impl()`?

Side (but maybe related) note: There are a lot of shoc variables labeled `diagnostic output` which are never used outside of `shoc_main()`. What is the point of keeping these around as global variables in internal shoc, instead of having them as local variables int the workspace manager?

bartgol: Not sure about shoc, but in P3 the short answer is yes. I see that diag_outputs contain both Field's views and Buffer's views. That's somewhat of a mix, and prevents us from passing a view_1d<Buffer>. Your 2nd question is probably linked to if/how we can do this. Imho, shoc and p3 (as well as every other atm proc that can be written as a single big kernel) should not have buffer views of length (ncols,...), but rather rely completely on WSM slots. There's no point in allocating (ncols,nlev) views if you don't plan to stick them in the FM (or, as it's the case for Homme, need these views to persist across 2+ kernels).

bartgol (approved):

components/scream/src/physics/p3/atmosphere_microphysics.hpp
tcclevenger: Is this worth focusing on now? I wouldn't mind continuing working on it.

bartgol: I don't think so. Memory access on GPU is relatively fast, so locality has a lesser impact than it may have on CPU. I would add this to v1.1 or v2 stuff.

PR #1049, opened 2021-05-13
Should never change global git configs

jgfouca:
Limit the change to the local repo and only do it for e3sm-jenkins.

The prior implementation inadvertently changed my personal git config such
that all my subsequent commits were done as authoer "Jenkins Jenkins".


bartgol (approved):
Good catch!

PR #1050, opened 2021-05-13
Compute cell area in PointGrid and use for dx/dy in SHOC

tcclevenger:
Instead of `host_dx/dy` being a FM variable, we now compute cell area in the PointGrid and use this to calculate `host_dx/dy` in SHOC pre-processing. The cell area is estimated to be `4*pi*R^2/num_cols`m.

### DIFF
--- a/components/scream/src/physics/shoc/atmosphere_macrophysics.cpp
+++ b/components/scream/src/physics/shoc/atmosphere_macrophysics.cpp
@@ -27,8 +27,13 @@ void SHOCMacrophysics::set_grids(const std::shared_ptr<const GridsManager> grids
 
   const auto& grid_name = m_shoc_params.get<std::string>("Grid");
   auto grid = grids_manager->get_grid(grid_name);
+
   m_num_cols = grid->get_num_local_dofs(); // Number of columns on this rank
   m_num_levs = grid->get_num_vertical_levels();  // Number of levels per column
+
+  // TODO: In preprocessing, we assume area is in meters. This may not
+  //       always be the case.
+  m_cell_area = grid->get_geometry_data("area"); // area of each cell
 
   // Define the different field layouts that will be used for this process
   using namespace ShortFieldTagsNames;
@@ -141,7 +146,9 @@ void SHOCMacrophysics::initialize_impl (const util::TimeStamp& t0)
   const int nlevi_packs = ekat::npack<Spack>(m_num_levs+1);
   const int num_tracer_packs = ekat::npack<Spack>(m_num_tracers);
 
-  view_1d wpthlp_sfc("wpthlp_sfc",m_num_cols),
+  view_1d host_dx("host_dx",m_num_cols),
+          host_dy("host_dy",m_num_cols),
+          wpthlp_sfc("wpthlp_sfc",m_num_cols),
           wprtp_sfc("wprtp_sfc",m_num_cols),
           upwp_sfc("upwp_sfc",m_num_cols),
           vpwp_sfc("vpwp_sfc",m_num_cols);
@@ -162,15 +169,15 @@ void SHOCMacrophysics::initialize_impl (const util::TimeStamp& t0)
           qc_copy("qc_copy",m_num_cols,nlev_packs),
           tke_copy("tke_copy",m_num_cols,nlev_packs);
 
-  shoc_preprocess.set_variables(m_num_cols,m_num_levs,m_num_tracers,T_mid,
-                                z_int,z_mid,p_mid,pseudo_density,omega,phis,surf_sens_flux,surf_latent_flux,
-                                surf_u_mom_flux,surf_v_mom_flux,qv,qv_copy,qc,qc_copy,tke,tke_copy,
+  shoc_preprocess.set_variables(m_num_cols,m_num_levs,m_num_tracers,m_cell_area,
+                                T_mid,z_int,z_mid,p_mid,pseudo_density,omega,phis,surf_sens_flux,surf_latent_flux,
+                                surf_u_mom_flux,surf_v_mom_flux,qv,qv_copy,qc,qc_copy,tke,tke_copy,host_dx,host_dy,
                                 s,rrho,rrho_i,thv,dz,zt_grid,zi_grid,wpthlp_sfc,wprtp_sfc,upwp_sfc,vpwp_sfc,
                                 wtracer_sfc,wm_zt,exner,thlm,qw);
 
   // Input Variables:
-  input.host_dx     = m_shoc_fields_in["host_dx"].get_reshaped_view<const Real*>();
-  input.host_dy     = m_shoc_fields_in["host_dy"].get_reshaped_view<const Real*>();
+  input.host_dx     = shoc_preprocess.host_dx;
+  input.host_dy     = shoc_preprocess.host_dy;
   input.zt_grid     = shoc_preprocess.zt_grid;
   input.zi_grid     = shoc_preprocess.zi_grid;
   input.pres        = p_mid;
--- a/components/scream/src/physics/shoc/atmosphere_macrophysics.cpp
+++ b/components/scream/src/physics/shoc/atmosphere_macrophysics.cpp
@@ -146,8 +146,7 @@ void SHOCMacrophysics::initialize_impl (const util::TimeStamp& t0)
   const int nlevi_packs = ekat::npack<Spack>(m_num_levs+1);
   const int num_tracer_packs = ekat::npack<Spack>(m_num_tracers);
 
-  view_1d host_dx("host_dx",m_num_cols),
-          host_dy("host_dy",m_num_cols),
+  view_1d cell_length("cell_length",m_num_cols),
           wpthlp_sfc("wpthlp_sfc",m_num_cols),
           wprtp_sfc("wprtp_sfc",m_num_cols),
           upwp_sfc("upwp_sfc",m_num_cols),
@@ -171,13 +170,13 @@ void SHOCMacrophysics::initialize_impl (const util::TimeStamp& t0)
 
   shoc_preprocess.set_variables(m_num_cols,m_num_levs,m_num_tracers,m_cell_area,
                                 T_mid,z_int,z_mid,p_mid,pseudo_density,omega,phis,surf_sens_flux,surf_latent_flux,
-                                surf_u_mom_flux,surf_v_mom_flux,qv,qv_copy,qc,qc_copy,tke,tke_copy,host_dx,host_dy,
+                                surf_u_mom_flux,surf_v_mom_flux,qv,qv_copy,qc,qc_copy,tke,tke_copy,cell_length,
                                 s,rrho,rrho_i,thv,dz,zt_grid,zi_grid,wpthlp_sfc,wprtp_sfc,upwp_sfc,vpwp_sfc,
                                 wtracer_sfc,wm_zt,exner,thlm,qw);
 
   // Input Variables:
-  input.host_dx     = shoc_preprocess.host_dx;
-  input.host_dy     = shoc_preprocess.host_dy;
+  input.dx          = shoc_preprocess.cell_length;
+  input.dy          = shoc_preprocess.cell_length;
   input.zt_grid     = shoc_preprocess.zt_grid;
   input.zi_grid     = shoc_preprocess.zi_grid;
   input.pres        = p_mid;
--- a/components/scream/src/physics/shoc/atmosphere_macrophysics.hpp
+++ b/components/scream/src/physics/shoc/atmosphere_macrophysics.hpp
@@ -206,7 +206,7 @@ public:
 
     // Assigning local variables
     void set_variables(const int ncol_, const int nlev_, const int num_qtracers_,
-                       const view_1d_const& area_,
+                       const view_1d_const_double& area_,
                        const view_2d_const& T_mid_, const view_2d_const& z_int_,
                        const view_2d_const& z_mid_, const view_2d_const& p_mid_, const view_2d_const& pseudo_density_,
                        const view_2d_const& omega_,
--- a/components/scream/src/physics/shoc/atmosphere_macrophysics.hpp
+++ b/components/scream/src/physics/shoc/atmosphere_macrophysics.hpp
@@ -149,6 +149,9 @@ public:
       const int nlev_v = (nlev-1)/Spack::n;
       const int nlev_p = (nlev-1)%Spack::n;
 
+      host_dx(i) = sqrt(area(i));
+      host_dy(i) = host_dx(i);
+
       wpthlp_sfc(i) = surf_sens_flux(i)/(cpair*rrho_i(i,nlev_v)[nlev_p]);
       wprtp_sfc(i)  = surf_latent_flux(i)/rrho_i(i,nlev_v)[nlev_p];
       upwp_sfc(i)   = surf_u_mom_flux(i)/rrho_i(i,nlev_v)[nlev_p];
@@ -161,6 +164,7 @@ public:
 
     // Local variables
     int ncol, nlev, num_qtracers;
+    view_1d_const area;
     view_2d_const T_mid;
     view_2d_const z_int;
     view_2d_const z_mid;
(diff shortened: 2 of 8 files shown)


components/scream/src/physics/shoc/atmosphere_macrophysics.cpp
```diff
@@ -141,7 +146,9 @@ void SHOCMacrophysics::initialize_impl (const util::TimeStamp& t0)
   const int nlevi_packs = ekat::npack<Spack>(m_num_levs+1);
   const int num_tracer_packs = ekat::npack<Spack>(m_num_tracers);
 
-  view_1d wpthlp_sfc("wpthlp_sfc",m_num_cols),
+  view_1d host_dx("host_dx",m_num_cols),
+          host_dy("host_dy",m_num_cols),
+          wpthlp_sfc("wpthlp_sfc",m_num_cols),
```
bartgol, line 151: What do `wpthlp_sfc` and the other `_sfc` names stand for? Any chance we can use a few more chars and make it clearer?

bartgol: I know this stuff was here before this PR, so it's not a deal breaker. But since it would be a simple find+replace op, if there's the will and opportunity to make it more readable, we might as well take it.

components/scream/src/physics/shoc/atmosphere_macrophysics.hpp
```diff
@@ -167,50 +171,55 @@ class SHOCMacrophysics : public scream::AtmosphereProcess
...
-    view_1d       wprtp_sfc;
-    view_1d       upwp_sfc;
-    view_1d       vpwp_sfc;
-    view_2d       wtracer_sfc;
-    view_2d       wm_zt;
-    view_2d       exner;
-    view_2d       thlm;
-    view_2d       qw;
-    view_2d       cloud_frac;
+    view_1d_const_double area;
```
bartgol, line 174: Side comment: I'm baffled that github shows this huge diff, when in reality is a small diff. Took me a while to realize you simply added three lines...

components/scream/src/share/grid/point_grid.cpp
```diff
@@ -1,4 +1,6 @@
 #include "share/grid/point_grid.hpp"
+#include "ekat/kokkos/ekat_kokkos_utils.hpp"
+#include "../../physics/share/physics_constants.hpp"
```
bartgol, line 3: I would avoid relative paths when including. You can use the path relative to the `src` folder: 

```
#include "physics/share/physics_constants.hpp"
```

bartgol: At some point we should move the constants to `src/share`, cause it's a bit ugly when the share folder includes stuff in the physics one. Yes, it's only headers, so it does not create target dependencies, but still..

components/scream/src/share/grid/point_grid.cpp
```diff
@@ -121,6 +123,36 @@ create_point_grid (const std::string& grid_name,
 
   grid->set_dofs(dofs_gids);
 
+  using device_type       = DefaultDevice;
```
bartgol, line 126: For consistence safety, it might be better to get these types (except the constants) from the grid type itself:

```
using geo_view_type = PointGrid::geo_view_type;
using KT                    = PointGrid::kokkos_types;
```
They are defined [here](https://github.com/E3SM-Project/scream/blob/e3a2be3e60c78610121cd91eaa4765dc79cdb500/components/scream/src/share/grid/abstract_grid.hpp#L40).

Also, KT and kokkos_types are redundant here.

bartgol (commented):
Looks good. I have some comments, nothing big, feel free to push back.

components/scream/src/physics/shoc/atmosphere_macrophysics.hpp
tcclevenger: I think the issue is that to line up all the view declarations, I needed each view to have an additional space..

components/scream/src/physics/shoc/atmosphere_macrophysics.hpp
```diff
@@ -28,16 +28,17 @@ class SHOCMacrophysics : public scream::AtmosphereProcess
...
+  using IntSmallPack         = typename SHF::IntSmallPack;
+  using Smask                = typename SHF::Smask;
+  using view_1d              = typename SHF::view_1d<Real>;
+  using view_1d_const        = typename SHF::view_1d<const Real>;
+  using view_1d_const_double = typename SHF::view_1d<const double>;
+  using view_2d              = typename SHF::view_2d<SHF::Spack>;
+  using view_2d_const        = typename SHF::view_2d<const Spack>;
+  using sview_2d             = typename KokkosTypes<DefaultDevice>::template view_2d<Real>;
+  using view_3d              = typename SHF::view_3d<Spack>;
+  using view_3d_const        = typename SHF::view_3d<const Spack>;
```
PeterCaldwell, line 42: Not a complaint about this PR, but just commenting on something that makes me a bit uncomfortable: SHF usually means sensible heat flux to atm scientists. It looks like this is being used to denote something very different here, which will confuse atm people reading the code. Also, I'm personally not clear what SHF means here.

components/scream/src/physics/shoc/atmosphere_macrophysics.hpp
bartgol: Ah, I see. Ok, github didn't go nuts then. Thanks!

components/scream/src/physics/shoc/atmosphere_macrophysics.hpp
bartgol: I'm guessing ShocFunctions?

components/scream/src/share/grid/point_grid.cpp
```diff
@@ -121,6 +123,36 @@ create_point_grid (const std::string& grid_name,
...
+  const double rearth    = C::r_earth;
+  const double pi        = C::Pi;
+  const double cell_area = 4*pi*rearth*rearth/num_my_cols;
+
+  const auto policy = ekat::ExeSpaceUtils<KT::ExeSpace>::get_default_team_policy(num_my_cols, num_vertical_lev);
+  Kokkos::parallel_for("area_loop", policy, KOKKOS_LAMBDA (const KT::MemberType& team) {
+    const int i = team.league_rank();
+    area(i) = cell_area;
+    lon(i) = std::nan("");
+    lat(i) = std::nan("");
```
PeterCaldwell, line 150: Am I understanding correctly that we are adding lat and lon to geo_view object now because we know it will be useful, but we are punting on actually giving it values at this point, assigning these things to NaN instead? Or is lat and lon getting assigned somewhere else?

components/scream/src/physics/shoc/atmosphere_macrophysics.hpp
bartgol: I think a lot of types are duplicated between, say, P3Functions and ShocFunctions. I am wondering whether a `Types` struct in `src/share/scream_types.hpp` might be beneficial to avoid redundancies, and also to have somewhat more intuitive names, like

```
using view_3d = typename Types::view_3d<Spack>;
```

tcclevenger: Yeah, I think this is confusing, but it's used all over shoc. I wouldn't mind implementing this shared types struct. Or at the least maybe use a more discriptive name (`SHOCFunc` or even just `SHOC`).

components/scream/src/share/grid/point_grid.cpp
bartgol: This function has no info about the sphere, so it can't setup lat/lon. It would require a massive amount of code to figure out the lat/lon coords of N uniformly distributed points on the sphere... On the other hand, the DynamicsDrivenGridsManager can extract lat/lon info from Homme, and will set them in the geo_data of the phys grid.

So we will lack lat/lon info for standalone shoc tests, but we will have that info for actually coupled (i.e., with dynamics) runs.

bartgol: Note: rrtmgp might actually need lat/lon info (for day/night/angle stuff), I am not sure. If so, in order to allow small standalone tests, we will have to either a) find a way to build a small PointGrid that _does_ have lat/lon info, or b) simply load small grids from file (with lat/lon info stored).

components/scream/src/physics/shoc/atmosphere_macrophysics.cpp
PeterCaldwell: _sfc stands for surface. I personally don't have a problem with that. I agree that wpthlp is a mouthful. It stands for $\overline{w'\theta_l'}$, though, which is hard to write in English at all. Maybe vert_liq_pot_T_flux? I think these names were created by @bogensch so maybe he has an opinion on this.

components/scream/src/share/grid/point_grid.cpp
tcclevenger: I think that when the dynamics grid is used it will have lat defined, but we don't need it for the PointGrid and so we just set to NaN so that if someone tries to use it they will get an exception that shows it isn't set.

components/scream/src/physics/shoc/atmosphere_macrophysics.cpp
tcclevenger: From what I understand these are surface values. I could replace with `_surface`.

components/scream/src/physics/shoc/atmosphere_macrophysics.hpp
PeterCaldwell: Yeah, I agree a Types struct would be nice. Seems like something we should work on for SCREAMv1.1 rather than changing what we've already done now?

bartgol: I already changed my mind, in the sense that P3, SHOC, and whatnot might have slightly different needs, so maybe a common types struct won't give us much. However, on a slightly related topic, I advocated in the past for splitting P3Functions into a P3Types struct and a functions namespace (`Types` is only 3 chars less then `Functions`, so not big saving probably, but there were other reasons). One day I'll bring it up again to haunt you with philosophical discussions on C++ and compilation units...

PeterCaldwell:
One other general comment: I like the cell area = `4*pi*R^2/num_cols` approach, but note that it will fail horribly when we start doing planar doubly periodic runs. I think that's ok, but we should make sure we don't forget to extend this when we get around to adding doubly-periodic mode in C++. I expect that at that time dx and dy will need to be specified from the user somehow. @bogensch, tagging you because you should be aware of this detail.

bartgol:
> One other general comment: I like the cell area = `4*pi*R^2/num_cols` approach, but note that it will fail horribly when we start doing planar doubly periodic runs. I think that's ok, but we should make sure we don't forget to extend this when we get around to adding doubly-periodic mode in C++. I expect that at that time dx and dy will need to be specified from the user somehow. @bogensch, tagging you because you should be aware of this detail.

We could do something along the line of what @celdredsandia did in Homme, where we save a scaling factor (and/or its inverse), which defaults to `4*pi*rearth^2` (for the sphere), but it can be configured at run time for planar geometries.

components/scream/src/physics/shoc/atmosphere_macrophysics.cpp
bogensch: wpthlp was intended to stand for w_prime_thl_prime, but since it's not clear to others outside of turbulence I have no objections to something like vert_liq_pot_T_flux

bogensch:
Yes, the F90 version of DP-SCREAM does not use the standard calculation based on a sphere to calculate dx/dy, a special routine is called that uses the dx/dy computed from planar dycore routines.

tcclevenger:
Thanks everyone! @bartgol I removed the local include path and eliminated the redundant typedefs

components/scream/src/physics/shoc/atmosphere_macrophysics.cpp
```diff
@@ -162,15 +169,15 @@ void SHOCMacrophysics::initialize_impl (const util::TimeStamp& t0)
...
+  shoc_preprocess.set_variables(m_num_cols,m_num_levs,m_num_tracers,m_cell_area,
+                                T_mid,z_int,z_mid,p_mid,pseudo_density,omega,phis,surf_sens_flux,surf_latent_flux,
+                                surf_u_mom_flux,surf_v_mom_flux,qv,qv_copy,qc,qc_copy,tke,tke_copy,host_dx,host_dy,
                                 s,rrho,rrho_i,thv,dz,zt_grid,zi_grid,wpthlp_sfc,wprtp_sfc,upwp_sfc,vpwp_sfc,
                                 wtracer_sfc,wm_zt,exner,thlm,qw);
 
   // Input Variables:
-  input.host_dx     = m_shoc_fields_in["host_dx"].get_reshaped_view<const Real*>();
-  input.host_dy     = m_shoc_fields_in["host_dy"].get_reshaped_view<const Real*>();
+  input.host_dx     = shoc_preprocess.host_dx;
```
bartgol, line 179: Sorry for the late comment, but I was going through this once again, and stumbled into "host_dx". I am guessing 'host' stands for "host_atm_model", but in a GPU-friendly implementation, the word "host" can be a bit ambiguous. Would `column_dx` or `col_dx` or simply `dx` be an acceptable name? Pinging also @PeterCaldwell @bogensch .

Also, is it really needed to keep two views, one for dx and one for dy, considering we set dy=dx? Is there a plan to use a "non-square columns" assumption?

PeterCaldwell: Great points, Luca. In terms of dx and dy being combined into a single variable: I think I asked the same question and the answer is that this is a feature for the future. Folks might want to use SHOC in models with non-uniform x and y in the future.

In terms of "host" - I totally agree that this is confusing in a GPU code. I think `input_dx` would be more appropriate in this setting. @bogensch ?

bartgol: @tcclevenger what do you think of making dy a shallow copy of dx for now then? Granted, it's not _a ton_ of memory we'd save, but it's also a simple change. Keeping two aliasing views will make transitioning to dy!=dx easier later on (just need to change the way we set dy up, keeping the code that uses dy the same), while at the same time saving a bit of memory.

bogensch: Yeah, if host_dx is confusing then i'm fine with changing it.  I think I prefer "column_dx" over "input_dx", but both work.

PeterCaldwell: Ok, sounds good. I like the shallow copy idea and I don't really care what the name is. That said, since all variables are column variables (right?), I'm confused what value "column_" adds. Can we just call them dx and dy?

tcclevenger: No problem! I went ahead and changed `host_dx/dy` to `dx/dy` throughout shoc src code (I didn't change variable names related to testing, but can). 

For the shallow copy, instead of having both `dx/dy` as preprocessing variables, I just use `cell_length = sqrt(area)` with a comment that we would need to compute `dx/dy` there instead if `dx!=dy`.

components/scream/src/physics/shoc/atmosphere_macrophysics.hpp
```diff
@@ -28,16 +28,17 @@ class SHOCMacrophysics : public scream::AtmosphereProcess
...
-  using view_2d_const  = typename SHF::view_2d<const Spack>;
-  using sview_2d = typename KokkosTypes<DefaultDevice>::template view_2d<Real>;
-  using view_3d = typename SHF::view_3d<Spack>;
-  using view_3d_const = typename SHF::view_3d<const Spack>;
+  using Spack                = typename SHF::Spack;
+  using IntSmallPack         = typename SHF::IntSmallPack;
+  using Smask                = typename SHF::Smask;
+  using view_1d              = typename SHF::view_1d<Real>;
+  using view_1d_const        = typename SHF::view_1d<const Real>;
+  using view_1d_const_double = typename SHF::view_1d<const double>;
```
bartgol, line 36: Side comment: if hard coding double for geometric info ends up making the code cumbersome, we can turn the geo data into Real's as well. After all, we are not "viewing" some array inside Homme (which uses double across the board), so we can easily store the data with whatever precision...

bartgol (approved):

PR #1051, opened 2021-05-13
Fix scorpio decomp logic

bartgol:
When initializing the decomposition for a given variable in scorpio, we don't need the dof gids, but rather the _offsest_ of the dofs on this rank wirh respect to the global array. 

E.g., if we have 4 global columns with gids [2, 5, 6, 10], and if rank 0 has col gids [2, 5] and rank 1 has col gids [6, 10], then rank 0 needs to pass [0, 1] to PIOc_InitDecomp, while rank 1 needs to pass [2, 3] (assuming this is for a 2d var, like surf_pressure).

Note: I only fixed `AtmosphereInput`, but a similar fix is needed for `AtmosphereOutput`.

Note: this PR depends on E3SM-Project/EKAT#108. So the branch needs to be updated once that PR is merged in ekat's master.


AaronDonahue (approved):
Looks great! Thanks Luca.

PR #1052, opened 2021-05-14
Infrastructure to generate stand-alone input data from SCREAMv0 runs

AaronDonahue:
This commit introduces the infrastructure to produce output files from SCREAMv0 runs that can be used as initial files for the stand-alone tests.

The current strategy employed in this PR is to addendum the variable name with "inX" where X is the process we care about.  There would still need to be some post-processing of the data to strip the "inX" from the variable name.  Additionally, SCREAMv0 will write output with the column and level dimensions flipped, so these need to be flipped back for use with SCREAM.  Both of these tasks can be accomplished using the NCO toolbox.

The SCREAM init fields are not included as default output, so as to not increase the memory footprint of typical runs.  Instead, the user will need to use the fincl namelist option.


AaronDonahue:
It might finally be time to merge this into master so that we can grab fields SCREAM cares about from a SCREAMv0 run more generally.

components/eam/src/dynamics/se/stepon.F90
```diff
@@ -85,7 +85,7 @@ subroutine stepon_init(dyn_in, dyn_out )
   use cam_history,            only: register_vector_field
   use gravity_waves_sources,  only: gws_init
   use phys_control,           only: use_gw_front
-  use cam_history_support,    only: max_fieldname_len
+  use cam_history_support,    only: max_fieldname_len, add_hist_coord
```
brhillman, line 88: `add_hist_coord` does not look like it is used here. I suggest not including it in the `use` statement if it is not used in this routine.

components/eam/src/physics/cam/micro_p3_interface.F90
```diff
@@ -1048,6 +1091,29 @@ subroutine micro_p3_tend(state, ptend, dtime, pbuf)
...
+    call outfld("inv_qc_relvar_inP3",      relvar,                 pcols, lchnk) 
+    call outfld("bm_inP3",                 state%q(:,:,ixrimvol),  pcols, lchnk) 
+    call outfld("nc_inP3",                 state%q(:,:,ixnumliq),  pcols, lchnk) 
+    call outfld("ni_inP3",                 state%q(:,:,ixnumice),  pcols, lchnk) 
+    call outfld("nr_inP3",                 state%q(:,:,ixnumrain), pcols, lchnk) 
+    call outfld("qi_inP3",                 state%q(:,:,ixcldice),  pcols, lchnk) 
+    call outfld("qm_inP3",                 state%q(:,:,ixcldrim),  pcols, lchnk) 
+    call outfld("qr_inP3",                 state%q(:,:,ixrain),    pcols, lchnk) 
+    call outfld("cldfrac_tot_inP3",        ast,                    pcols, lchnk) 
+    call outfld("cldfrac_tot_inRAD",       ast,                    pcols, lchnk) ! Radiation needs this one too in SCREAM
```
brhillman, line 1110: If radiation is using this, shouldn't it be output from radiation?

components/eam/src/physics/cam/zm_conv_intr.F90
```diff
@@ -173,6 +174,74 @@ subroutine zm_conv_init(pref_edge)
     call addfld ('ZMICUD',     (/ 'lev' /), 'A', 'm/s', 'ZM in-cloud U downdrafts')
     call addfld ('ZMICVU',     (/ 'lev' /), 'A', 'm/s', 'ZM in-cloud V updrafts')
     call addfld ('ZMICVD',     (/ 'lev' /), 'A', 'm/s', 'ZM in-cloud V downdrafts')
+
+    ! Add fields for ZM stand-alone tests
```
brhillman, line 178: I'm usually not a fan of leaving commented out code, this will likely go out of date, and may never end up being used if we abandon the fortran version of the code.

components/eam/src/physics/cam/micro_p3_interface.F90
AaronDonahue: I don't think radiation is using this in F90.  I'll have to check.  If it isn't then I figured it was probably more than we needed to have radiation grab it from the pbuf just to output it there.  That being said it is a little weird to grab something for RAD from P3.  So maybe I should just add it in the RAD interface anyways.

components/eam/src/physics/cam/zm_conv_intr.F90
AaronDonahue: Fair point.  I struggled with keeping this or not.  Happy to delete.

components/eam/src/physics/rrtmgp/radiation.F90
```diff
@@ -2744,6 +2792,7 @@ subroutine get_gas_vmr(icall, state, pbuf, gas_names, gas_vmr)
 
       ! initialize
       gas_vmr(:,:,:) = 0._r8
+      if (present(write_mmr).and.write_mmr) allocate(gas_out(pcols,pver))
```
brhillman, line 2795: http://www.cs.rpi.edu/~szymansk/OOF90/bugs.html#1

See "Danger with Optional Arguments", this could potentially cause an out of bounds memory reference.

components/eam/src/physics/rrtmgp/radiation.F90
```diff
@@ -617,6 +624,28 @@ subroutine radiation_init(state,pbuf)
       call add_hist_coord('swband', nswbands, 'Shortwave wavenumber', 'cm-1', sw_band_midpoints)
       call add_hist_coord('lwband', nlwbands, 'Longwave wavenumber', 'cm-1', lw_band_midpoints)
 
+
+      ! Add outfields to be used for RRTMGP stand-alone testing.
+      call cnst_get_ind('CLDLIQ',ixcldliq)
+      call cnst_get_ind('CLDICE',ixcldice)
+      do igas = 1,size(active_gases)
+        active_gases_index(igas) = igas
```
brhillman, line 632: This does not appear to be used.

components/eam/src/physics/rrtmgp/radiation.F90
```diff
@@ -178,6 +178,7 @@ module radiation
       'H2O', 'CO2', 'O3 ', 'N2O', &
       'CO ', 'CH4', 'O2 ', 'N2 ' &
    /)
+   real(r8), target :: active_gases_index(size(active_gases))
```
brhillman, line 181: This does not appear to be used.

components/eam/src/physics/rrtmgp/radiation.F90
```diff
@@ -617,6 +624,28 @@ subroutine radiation_init(state,pbuf)
       call add_hist_coord('swband', nswbands, 'Shortwave wavenumber', 'cm-1', sw_band_midpoints)
       call add_hist_coord('lwband', nlwbands, 'Longwave wavenumber', 'cm-1', lw_band_midpoints)
 
+
+      ! Add outfields to be used for RRTMGP stand-alone testing.
```
brhillman, line 628: General comment, probably no need to change this since it appears to be pretty ubiquitous in E3SM, but these do not appear to be used in this routine, only later when/if these are actually written out. I kind of prefer to keep this query closer to its use, it makes for more readable code, unless this saves computational cost by only initializing once.

components/eam/src/physics/rrtmgp/radiation.F90
AaronDonahue: ahhh good catch! Thanks

AaronDonahue:
closing for now, until we finalize our strategy for getting IC in SCREAMv1 runs

PR #1054, opened 2021-05-17
Bugfix: p3 interface no minimum cloud fraction being set.

AaronDonahue:
Fixes a bug in the p3 interface where the cloud fraction values were not being set to the mincld minimum.  This could potentially cause issues with division by zero when finding the in-cloud mass values.

### DIFF
--- a/components/scream/src/physics/p3/atmosphere_microphysics.hpp
+++ b/components/scream/src/physics/p3/atmosphere_microphysics.hpp
@@ -88,9 +88,9 @@ public:
         th_atm(icol,ipack).set(oT_atm_mask,oth);
         // Cloud fraction
         // Set minimum cloud fraction - avoids division by zero
-        cld_frac_l(icol,ipack).set(Smask(true),mincld);
-        cld_frac_i(icol,ipack).set(Smask(true),mincld);
-        cld_frac_r(icol,ipack).set(Smask(true),mincld);
+        cld_frac_l(icol,ipack) = mincld;
+        cld_frac_i(icol,ipack) = mincld;
+        cld_frac_r(icol,ipack) = mincld;
         // Update cloud fractions based on the total cloud fraction
         const Spack oast(ast(icol,ipack));
         const Smask oasti_mask(!isnan(oast) and oast>mincld);
--- a/components/scream/src/physics/p3/atmosphere_microphysics.hpp
+++ b/components/scream/src/physics/p3/atmosphere_microphysics.hpp
@@ -86,12 +86,18 @@ public:
         const Smask oT_atm_mask(!isnan(oT_atm) and oT_atm>0.0);
         auto oth = physics_fun::T_to_th(oT_atm,oexner,oT_atm_mask);
         th_atm(icol,ipack).set(oT_atm_mask,oth);
-        // Cloud fraction and dz
+        // Cloud fraction
+        // Set minimum cloud fraction - avoids division by zero
+        cld_frac_l(icol,ipack).set(Smask(true),mincld);
+        cld_frac_i(icol,ipack).set(Smask(true),mincld);
+        cld_frac_r(icol,ipack).set(Smask(true),mincld);
+        // Update cloud fractions based on the total cloud fraction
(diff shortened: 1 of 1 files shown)


components/scream/src/physics/p3/atmosphere_microphysics.hpp
```diff
@@ -86,12 +86,18 @@ class P3Microphysics : public AtmosphereProcess
         const Smask oT_atm_mask(!isnan(oT_atm) and oT_atm>0.0);
         auto oth = physics_fun::T_to_th(oT_atm,oexner,oT_atm_mask);
         th_atm(icol,ipack).set(oT_atm_mask,oth);
-        // Cloud fraction and dz
+        // Cloud fraction
+        // Set minimum cloud fraction - avoids division by zero
+        cld_frac_l(icol,ipack).set(Smask(true),mincld);
+        cld_frac_i(icol,ipack).set(Smask(true),mincld);
+        cld_frac_r(icol,ipack).set(Smask(true),mincld);
```
PeterCaldwell, line 93: Apologies if I'm being dense, but I think you're setting all the cld_fracs to mincld everywhere since Smask(true) is always true(?)... so why not just set cld_fracs = mincld here? Also, I'm confused what makes padded values masked out here.

AaronDonahue: @PeterCaldwell , my understanding is that because all three `cld_frac` values in this case are packed arrays there is no definition for `cld_frac_x = mincld` <- @bartgol correct me if I'm wrong.  In other words, changing the syntax produces a compiler error.  As for padded values, yes this means that even the padded values for cld_frac_x will be changed to mincld.  My understanding here is that the padded values won't be used so it shouldn't break anything.  That being said, if it is more correct to keep the padded values to always be NaN I'm happy to improve the syntax to check for padded values and not do anything with them.

bartgol: If `a` is a Pack<T,N>, `a = b;` is valid if `b` is a Pack<S,N> (with T constructible from S) , or a scalar S (with, again, T constructible from S). So `cld_frac_x = mincld` is valid.

AaronDonahue: Okay great, I misunderstood when `.set` is required.  I'll make that change and push a new version.

bartgol (approved):

components/scream/src/physics/p3/atmosphere_microphysics.hpp
PeterCaldwell: Ah that looks much better. Sorry to be a grumpkin, but I'd also like @bartgol 's opinion about whether it's kosher to assign non- NaN values to entries in a pack that shouldn't normally be used for computations. My concern is that if we made a mistake and used these padding values in real calculations, having those padding values be NaN would make error detection easier.

bartgol: @PeterCaldwell  That is a fair question. I think there are other places where we do this, so I'm not too concerned about this specific case. In general, I'm fine with assigning non-NaN to the padding, if it makes the calculation easier. The alternative is to compute a mask every time, just for the sake of "preserving" the garbage at the end, but I find that a tad too involved, and it might actually make a) the code less readable, and b) perhaps cost a few pointless flops/registers.

tl;dr: I think it's ok to write in the garbage part of the column if that makes the code easier to read.

bartgol: Note: of course, things are a bit different if we _read_ the garbage and use it inside some conditionals. Obviously, we later discard the values generated by that garbage, but in the process of using them, we can create valgrind errors. These were, in fact, a big chunk of the valgrind errors in shoc/p3 that we had to get rid of for our valgrind nightly test.

PeterCaldwell: Ok, thanks @bartgol ! That makes sense.

PeterCaldwell (approved):

PR #1057, opened 2021-05-20
Adopt common physics functions for all processes

AaronDonahue:
This PR removes all references to the universal physics functions and replaces them with the common physics functions which was recently introduced into the code by PR #1019 .


components/scream/src/physics/p3/atmosphere_microphysics.cpp
```diff
@@ -106,18 +106,20 @@ void P3Microphysics::initialize_impl (const util::TimeStamp& t0)
...
   //       variables a local view is constructed.
   const Int nk_pack = ekat::npack<Spack>(m_num_levs);
-  auto pmid  = m_p3_fields_in["p_mid"].get_reshaped_view<const Pack**>();
-  auto T_atm = m_p3_fields_out["T_mid"].get_reshaped_view<Pack**>();
-  auto ast   = m_p3_fields_in["cldfrac_tot"].get_reshaped_view<const Pack**>();
-  auto zi    = m_p3_fields_in["z_int"].get_reshaped_view<const Pack**>();
+  auto pmid           = m_p3_fields_in["p_mid"].get_reshaped_view<const Pack**>();
+  auto pseudo_density = m_p3_fields_in["pseudo_density"].get_reshaped_view<const Pack**>();
+  auto T_atm          = m_p3_fields_out["T_mid"].get_reshaped_view<Pack**>();
+  auto ast            = m_p3_fields_in["cldfrac_tot"].get_reshaped_view<const Pack**>();
```
bartgol, line 112: Isn't the var name `cld_frac_tot`? Meaning, isn't there a `_` between `cld` and `frac`, since they are two words?

components/scream/src/physics/p3/atmosphere_microphysics.hpp
```diff
@@ -79,13 +79,21 @@ class P3Microphysics : public AtmosphereProcess
         // Exner
         const Spack opmid(pmid(icol,ipack));
         const Smask opmid_mask(!isnan(opmid) and opmid>0.0);
```
bartgol, line 81: We can remove the isnan check. `opmid>0.0` is already false for entries that are nan, since every nan compares false to anything,regardless of the rhs and regardless of whether the comparison is ==,>,<,<=, or >=.

bartgol: On a side note: is it even possible that pmid<=0 ?

components/scream/src/physics/p3/atmosphere_microphysics.hpp
```diff
@@ -79,13 +79,21 @@ class P3Microphysics : public AtmosphereProcess
         // Exner
         const Spack opmid(pmid(icol,ipack));
```
bartgol, line 80: Ok, this is not strictly speaking related to this PR, but while we're at it: we should create refs to views' entries, to save a copy:
```
const auto& opmid = pmid(icol,ipack);
```

components/scream/src/physics/p3/atmosphere_microphysics.hpp
```diff
@@ -79,13 +79,21 @@ class P3Microphysics : public AtmosphereProcess
...
         const Smask opmid_mask(!isnan(opmid) and opmid>0.0);
-        auto oexner = physics_fun::get_exner(opmid,opmid_mask);
+        auto oexner = PF::exner_function(opmid);
```
bartgol, line 82: While we're at it, can we avoid `oblah` names, where possible? I get that `pmid` already exists, so we can't call the pack `pmid` (although p_mid works), but for other stuff like exner, can we just call it `exner`? I find the `o` in front a bit confusing, since I keep thinking it stands for something, like 'output', or something.

components/scream/src/physics/p3/atmosphere_microphysics.hpp
```diff
@@ -79,13 +79,21 @@ class P3Microphysics : public AtmosphereProcess
...
         const Spack oT_atm(T_atm(icol,ipack));
         const Smask oT_atm_mask(!isnan(oT_atm) and oT_atm>0.0);
-        auto oth = physics_fun::T_to_th(oT_atm,oexner,oT_atm_mask);
+        auto oth = PF::calculate_theta_from_T(oT_atm,opmid);
         th_atm(icol,ipack).set(oT_atm_mask,oth);
+        // DZ
+        const Spack opseudo_density(pseudo_density(icol,ipack));
+        const Spack oqv(qv(icol,ipack));
+        const Smask odz_mask(opmid_mask and oT_atm_mask and 
+                             !isnan(opseudo_density) and opseudo_density>0.0 and
```
bartgol, line 93: Here too: no need to check XYZ for nans if you `and` it with `XYZ>0`.

components/scream/src/physics/p3/atmosphere_microphysics.hpp
```diff
@@ -79,13 +79,21 @@ class P3Microphysics : public AtmosphereProcess
         // Exner
         const Spack opmid(pmid(icol,ipack));
         const Smask opmid_mask(!isnan(opmid) and opmid>0.0);
-        auto oexner = physics_fun::get_exner(opmid,opmid_mask);
+        auto oexner = PF::exner_function(opmid);
         inv_exner(icol,ipack).set(opmid_mask,1.0/oexner);
         // Potential temperature
         const Spack oT_atm(T_atm(icol,ipack));
         const Smask oT_atm_mask(!isnan(oT_atm) and oT_atm>0.0);
```
bartgol, line 86: Isn't the T>0 check a bit superfluous? If T<=0 (assuming it's even possible), I believe the code should have already crashed somewhere else...

components/scream/src/physics/p3/atmosphere_microphysics.hpp
bartgol: But again, dp cannot be <=0, so this mask is pointless maybe?

components/scream/src/physics/p3/atmosphere_microphysics.hpp
```diff
@@ -174,10 +176,10 @@ class P3Microphysics : public AtmosphereProcess
         // Update the atmospheric temperature and the previous temperature.
         const Spack opmid(pmid(icol,ipack));
         const Smask opmid_mask(!isnan(opmid) and opmid>0.0);
```
bartgol, line 178: Same considerations about checks and `oblah` names as in the preamble hold for the postprocessing too.

bartgol (changes requested):
Looks good, but I have a few comments.

components/scream/src/physics/p3/atmosphere_microphysics.hpp
AaronDonahue: I suppose a better question for these, and other similar mask declarations is, would it make sense to just go with the `X<0` which will catch the NaNs too as you mentioned.  Even if the code should have crashed by now because `X<0` is totally non-physical?  Or should we stick with the `isnan` check which should mask padded values in most cases and of course mask any NaNs that pop up?

components/scream/src/physics/p3/atmosphere_microphysics.cpp
bartgol: Ah, you're right. For some reason I thought we agreed on using `_` to separate words, much like we do in the code (and I believe that's also the CF convention).

Then I don't think this needs to be addressed in here. Though I might want to bring this up again soon.

components/scream/src/physics/p3/atmosphere_microphysics.hpp
bartgol: I'm not sure I understand what you are suggesting. X<0 (like X<=0, X==0, X>0, X>=0) will be false for NaNs, btw. What I was suggesting is to simply do `m = X>0;`, since `m[i]` will be false if `X[i]<=0` as well as if `X[i]` is NaN, hence doing two checks in one.

bartgol: That said, we should not check for NaNs, I think. Otherwise, where to we draw the line? Should we always check every single input for NaNs? If your goal was to simply mask out the padding at the end, then my suggestion is to not do it. We don't care what's in the padding, so even if it's NaNs, that's fine.

bartgol: Yes, it can. Creating a `Smask(true)` is pointless. Just do `X(icol,ilev) = PF::compute_blah(...)`.

components/scream/src/physics/p3/atmosphere_microphysics.hpp
AaronDonahue: in the set_variables function below should we also be passing by reference?

bartgol: No. Rule of thumb: builtin types should always be passed by value. Everything that amounts to more than one int/Real/char/pointer is usually better suited to be passed by const reference.

Note: I'm only referring to input values (that the fcn should not modify). If you expect the fcn to modify the argument, then always use reference or pointer, regardless of whether it is a builtin type or not.

AaronDonahue:
@bartgol , thank you for all the great comments.  I believe I addressed them with these latest commits.  Can you take a look?
@tcclevenger , I also made some adjustments to the SHOC pre/post ambles.  Can you take a look at those for me and make sure I didn't inadvertently break anything

components/scream/src/physics/p3/atmosphere_microphysics.hpp
```diff
@@ -76,28 +76,25 @@ class P3Microphysics : public AtmosphereProcess
     KOKKOS_INLINE_FUNCTION
     void operator()(const int icol) const {
       for (int ipack=0;ipack<m_npack;ipack++) {
+        // The ipack slice of input variables
+        const Spack pmid_pack(pmid(icol,ipack));
```
bartgol, line 80: All of these these should be const references, not hard copies. Like
```
const Spack pmid_pack = pmid(icol,ipack);
...
```

components/scream/src/physics/p3/atmosphere_microphysics.hpp
```diff
@@ -171,24 +160,18 @@ class P3Microphysics : public AtmosphereProcess
...
```
bartgol, line 164: These three as well: `const Spack& pmid_pack = ...`.

components/scream/src/physics/p3/atmosphere_microphysics.hpp
```diff
@@ -171,24 +160,18 @@ class P3Microphysics : public AtmosphereProcess
...
-        auto oT = physics_fun::th_to_T(oth_atm,oexner,oth_atm_mask);
-        T_atm(icol,ipack).set(oth_atm_mask,oT);
-        T_prev(icol,ipack).set(oth_atm_mask,T_atm(icol,ipack));
+        T_atm(icol,ipack) = PF::calculate_T_from_theta(th_atm_pack,pmid_pack);
+        T_prev(icol,ipack) = T_atm(icol,ipack);
         // Update qv_prev
-        const Spack oqv(qv(icol,ipack));
-        const Smask oqv_mask(!isnan(oqv) and oqv>0.0);
-        qv_prev(icol,ipack).set(oqv_mask,oqv);
+        qv_prev(icol,ipack) = qv_pack;
```
bartgol, line 171: You don't really need `qv_pack` here, since you only access it once. You could just do
```
qv_prev(icol,ipack) = qv(icol,ipack);
```

bartgol: Note: imho, a temp variable (even if it's a const ref) is helpful if you use it multiple times (avoids doing the view indexing math every time you need to access the entry) or if it makes the code shorter (like `const auto& tmp = my_very_long_view_name_that_doesnt_fit_in_the_next_line(icol,ipack)`).

bartgol (changes requested):
I think you missed some mods related to const refs vs copies.

AaronDonahue:
@bartgol , doh, thanks for catching the few remaining missed const references.  I've made the changes and cleaned up the code a bit more per your suggestions.  Can you take another look now?

tcclevenger (approved):
SHOC portion looks good to me.

components/scream/src/physics/shoc/atmosphere_macrophysics.hpp
```diff
@@ -96,9 +96,7 @@ class SHOCMacrophysics : public scream::AtmosphereProcess
       Kokkos::parallel_for(Kokkos::TeamThreadRange(team, nlev_packs), [&] (const Int& k) {
         // Exner
         const Spack p_mid_ik(p_mid(i,k));
```
tcclevenger, line 98: These could also be changed from hard copies

components/scream/src/physics/shoc/atmosphere_macrophysics.hpp
```diff
@@ -111,21 +109,17 @@ class SHOCMacrophysics : public scream::AtmosphereProcess
         qv_copy(i,k) = qv(i,k);
 
         // Temperature
+        const Spack T_mid_ik(T_mid(i,k));
```
tcclevenger, line 112: ...and here.

components/scream/src/physics/p3/atmosphere_microphysics.hpp
```diff
@@ -76,24 +76,22 @@ class P3Microphysics : public AtmosphereProcess
...
-        dz(icol,ipack) = PF::calculate_dz(pseudo_density_pack, pmid_pack, T_atm_pack, qv_pack);
+        dz(icol,ipack) = PF::calculate_dz(pseudo_density(icol,ipack), pmid_pack, T_atm_pack, qv(icol,ipack));
         // Cloud fraction
         // Set minimum cloud fraction - avoids division by zero
-        cld_frac_l(icol,ipack) = ekat::max(cld_frac_t(icol,ipack),mincld);
-        cld_frac_i(icol,ipack) = ekat::max(cld_frac_t(icol,ipack),mincld);
-        cld_frac_r(icol,ipack) = ekat::max(cld_frac_t(icol,ipack),mincld);
+        cld_frac_l(icol,ipack) = ekat::max(cld_frac_t_pack,mincld);
+        cld_frac_i(icol,ipack) = ekat::max(cld_frac_t_pack,mincld);
+        cld_frac_r(icol,ipack) = ekat::max(cld_frac_t_pack,mincld);
```
bartgol, line 94: Since the rhs is the same for all three, you could use `ekat::max` only in the first:
```
auto& frac_l = cld_frac_i(icol,ipack);
frac_l = ekat::max(...);
cld_frac_i(icol,ipack) = frac_i;
cld_frac_r(icol,ipack) = frac_i;
```

components/scream/src/physics/p3/atmosphere_microphysics.hpp
```diff
@@ -160,15 +158,11 @@ class P3Microphysics : public AtmosphereProcess
     KOKKOS_INLINE_FUNCTION
     void operator()(const int icol) const {
       for (int ipack=0;ipack<m_npack;ipack++) {
-        // The ipack slice of input variables
-        const Spack pmid_pack(pmid(icol,ipack));
-        const Spack th_atm_pack(th_atm(icol,ipack));
-        const Spack qv_pack(qv(icol,ipack));
         // Update the atmospheric temperature and the previous temperature.
-        T_atm(icol,ipack) = PF::calculate_T_from_theta(th_atm_pack,pmid_pack);
+        T_atm(icol,ipack)  = PF::calculate_T_from_theta(th_atm(icol,ipack),pmid(icol,ipack));
```
bartgol, line 162: To prevent accessing T_atm twice (with twice as much indexing arithmetic), you could exploit the fact that op= returns a reference:

```
T_prev(icol,ipack) = T_atm(icol,ipack) = PF::calculate_T_from_theta(...);
```

bartgol:
Looks like the AT beat me. Still suggestions to keep in mind for later.

PR #1058, opened 2021-05-21
Pass cloud mixing ratios to radiation instead of in-cloud mass

### LINKED ISSUE #965: Compute LWP and IWP in radiation from qc and qi
Radiation needs profiles of LWP and IWP = liquid water and ice water paths just within a single cell. Water paths typically have units of kg water/m2 and can (I think) be computed from qc and qi via LWP = qc*air_density*(z_int[k-1] - z_int[k]) and analogous for ice phase.

brhillman:
Pass cloud water and ice mixing ratios into radiation interface instead of expecting in-cloud water and ice mass (i.e., what is called in-cloud liquid and ice water path in EAM). This requires radiation to compute these quantities from the mixing ratios, but it makes more sense to keep this localized to radiation and just pass standard fields through the field manager, rather than requiring microphysics to compute things that are only needed by the radiation code. Fixes #965.

### DIFF
--- a/components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
+++ b/components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
@@ -50,8 +50,9 @@ void RRTMGPRadiation::set_grids(const std::shared_ptr<const GridsManager> grids_
   add_field<Required>("surf_alb_direct", scalar2d_swband_layout, nondim, grid->name());
   add_field<Required>("surf_alb_diffuse", scalar2d_swband_layout, nondim, grid->name());
   add_field<Required>("cos_zenith", scalar2d_layout, nondim, grid->name());
-  add_field<Required>("lwp", scalar3d_layout_mid, kg/m3, grid->name());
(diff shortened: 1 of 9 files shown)


brhillman:
@tcclevenger I wrote this using YAKL for now, since that was what most of the rad interface was using for consistency with RRTMGP, but as we discussed the other day it might make more sense for this extra code to support the interface to be written using kokkos before doing the copy to YAKL arrays. But, I wanted to get a solution to this issue up sooner rather than later, and this was already written.

brhillman:
Add WIP label for now since failing on weaver.

brhillman:
Passing on weaver now.

AaronDonahue (approved):
Looks great to me Ben! Thanks for putting this together.

tcclevenger (approved):
This looks good to me. When I finish the memory buffer for the AD I'll probably end up touching this code, and maybe then we can discuss moving to Kokkos, but it may not be of any benefit.

brhillman:
Thanks @tcclevenger. That sounds good.

PR #1059, opened 2021-05-21
Implementing hooks of HommeDynamics process to the AD

bartgol:
Include pre/post process (like convert T<->VTheta*dp).


components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
```diff
@@ -54,129 +55,410 @@ void HommeDynamics::set_grids (const std::shared_ptr<const GridsManager> grids_m
...
+  const auto nlevs  = m_dyn_grid->get_num_vertical_levels();
+  const auto policy = ekat::ExeSpaceUtils<KT::ExeSpace>::get_default_team_policy(num_elems*NP*NP,NVL);
+  const int n0 = c.get<Homme::TimeLevel>().n0;
+  const int nm1 = c.get<Homme::TimeLevel>().nm1;
+  const int np1 = c.get<Homme::TimeLevel>().np1;
+
+  const auto& hvcoord = c.get<Homme::HybridVCoord>();
+  const auto ps0 = hvcoord.ps0 * hvcoord.hybrid_ai0;
+
+  // Need two tepomraries, for pi_mid and pi_int
```
tcclevenger, line 357: Typo

tcclevenger (approved):
Looks good to me! There is a typo I pointed out, but maybe I can fix that with the memoryBuffer stuff to come later so that the tester doesn't need to re-run.

bartgol:
> Looks good to me! There is a typo I pointed out, but maybe I can fix that with the memoryBuffer stuff to come later so that the tester doesn't need to re-run.

Yeah, let's do it in your PR, to speed up things.

bartgol:
@tcclevenger I still get a nan layer thickness from Homme when the pack size is larger than 1. I will keep investigating why, but in the meantime, I'm going to push a commit with this test disabled, so that we can move on, and start setting up a Homme+Physics test. Does that sound ok?

bartgol:
@tcclevenger Bug found and fixed. The homme test now runs on my laptop. Once the AT tests pass, if you give the final approve, we can merge.

ambrad:
@bartgol I see that you modify some nhxx files in this PR. It might make future upstream merges harder if we modify nhxx files in the SCREAM fork rather than E3SM main.

bartgol:
> @bartgol I see that you modify some nhxx files in this PR. It might make future upstream merges harder if we modify nhxx files in the SCREAM fork rather than E3SM main.

Yes, some mods were needed for homme too. It seems to be a common pattern, every time I need to integrate scream and homme. I made sure to mark all commits with the prefix `HOMME:`, to make it easy to extract them later, and perhaps create a PR in the upstream.

tcclevenger (approved):

PR #1060, opened 2021-05-24
[WIP] Correction to P3 ice overdepletion check routine

JS-WRF-SBM:


components/eam/src/physics/cam/micro_p3.F90
```diff
@@ -2836,33 +2838,51 @@ subroutine back_to_cell_average(cld_frac_l,cld_frac_r,cld_frac_i,
...
+
+   real(rtype), intent(inout) :: qi2qv_sublim_tend,qr2qv_evap_tend
+
+   real(rtype) :: dumqv_sat_i, qdep_satadj, qrevp_satadj
+   real(rtype) :: qtmp_all,ttmp_all,qv_sat_l,q_sink,t_sink,dumqv_sat_l,qv_source_evp,qi2qv_sublim_satadj
+
+   ! ... available water vapor 
+   qtmp_all = qv - (qidep + qinuc)*dt + (qi2qv_sublim_tend + qr2qv_evap_tend)*dt 
+   ! ... corresponding temperature
+   ttmp_all = t_atm + ((qidep-qi2qv_sublim_tend+qinuc)*latent_heat_sublim*inv_cp + (-qr2qv_evap_tend*latent_heat_vapor*inv_cp))*dt
```
PeterCaldwell, line 2863: Good point that T should also be changed by microphysics during the timestep. This fix is a clear improvement.

components/eam/src/physics/cam/micro_p3.F90
```diff
@@ -2836,33 +2838,51 @@ subroutine back_to_cell_average(cld_frac_l,cld_frac_r,cld_frac_i,
...
+
+   ! ... water vapor needed for liquid saturation is the minimal requirment for both liquid and ice sturation
+   qv_sat_l = qv_sat(ttmp_all,pres,0)
+
+   ! ... if water vapor mass exceeds liquid saturation, we limit only source terms (e.g., sublimation, rain evp) 
+   if(qtmp_all > qv_sat_l)then
```
PeterCaldwell, line 2869: if statements don't work for GPU calculations. The original max and min approach is probably needed to ensure BFB agreement between Kokkos and F90 implementations.

PeterCaldwell:
Ok, I agree with Koby that qidep does not need to be adjusted in the routine modified in this PR because it is already adjusted in ice_supersat_conservation. qidep does still need to be an input (rather than an inout variable as before) because it affects end of step qv. 

I also agree that the previous treatment of ice_overdepletion was insufficient because it ignored the simultaneous impact from other processes in setting end-of-step qv. In particular, qr2qv_sublim_tend should also be reduced to prevent end-of-step liquid saturation. 

One thing ice_supersat_conservation and the reformulated ice_overdepletion don't do is ensure that sublimation doesn't happen if start-of-step qv is supersaturated and deposition doesn't happen for subsaturated condtions... but that check is baked into the original calculation of these variables in ice_deposition_sublimation.

PeterCaldwell:
One other thing this PR does is move ice_overdepletion from being the first conservation check to the last. Because later conservation checks change tendencies set by earlier conservation checks, earlier checks have a larger chance of having their relationship corrupted by the end of the checking process. Thus the most important checks should be saved for last. I don't think supersaturation is more important than non-negative mass so I'm dubious of this part of the PR. Since ice_overdepletion just reduces the amount of qi and qr converted to vapor, however, I don't think it is in any danger of driving any species negative... it will just make condensate masses set by prior conservation routines to be exactly zero at the end of a step a bit larger than zero. Thus I'm approving that change too even though I don't think it's warranted.

PeterCaldwell:
One thing I do want to change from Koby's implementation is the details of how prevent_ice_overdepletion work. Koby first limits evaporation by evap/(evap+sublim)*[correction needed to achieve liquid saturation if sublim was 0]. Then he limits sublimation by sublim/(evap+sublim)*[correction needed to achieve ice saturation if evap was 0]. This neatly gets around the problem that the latent heating coefficient for evaporation is different than for sublimation and gives the exact right answer if either sublimation or evaporation were zero, but I think it gets some details wrong. First, we should never enforce ice saturation adjustment because being supersaturated with respect to ice is fine. Second, the sequential nature of evaporation and sublimation corrections breaks the ratio between the originally-calculated evap and sublimation values. Although the whole conservation correction process is artificial, the relationship between process rates is one thing we try to preserve. 

Instead, I think it is better to adjust both evap and sublim by 'frac' in order to obtain liquid saturation at the end of the microphysics step (counted as the time just before sedimentation is called). These reductions in evap and sed causes T to change according to the latent heats of vaporization and sublimation, respectively, but these adjustments are easy to calculate. Linearizing the effect of these T changes on liquid saturation makes life easier at the expense of a bit of accuracy. 

Because all of these computations are somewhat complicated, I ran my impl versus Koby's in python to check their accuracy. In the graphic below, the top panels show the predicted evap and sublimation rates before (black lines at top) and after correction for a variety of qv values (along x axis). T_atm=263K and pres=850mb for all calculations. You can see that my version (blue) and Koby's (red) are somewhat different but have similar behavior. Note also that evap and sublim are always set to 0 for qv>2 g/kg because that's above liquid saturation by itself - additional qv sources would only make non-conservation worse. The bottom left plot (which only shows qv vals<2 g/kg) plots the ratio of evap to sublim tendencies after correction. This should be the same ratio as for the input data (=2). Koby's impl clearly does not conserve this. The bottom right panel shows the difference between qv and liquid saturation at the end of the step. Because the whole point of this overdepletion check is to set qv equal to liquid saturation, this value should be zero. My method has error because it linearizes the latent-heating effect. as qv gets bigger, that error increases because more condensational/sublimational correction is needed. Still, my error is generally less than Koby's. Note, however, that Koby's method also has relatively small error despite its empirical nature. 

Screen Shot 2021-08-17 at 10 17 16 AM

Based on this analysis, I'm going to implement my method for the PR.

JS-WRF-SBM:
> Ok, I agree with Koby that qidep does not need to be adjusted in the routine modified in this PR because it is already adjusted in ice_supersat_conservation. qidep does still need to be an input (rather than an inout variable as before) because it affects end of step qv.
> 
> I also agree that the previous treatment of ice_overdepletion was insufficient because it ignored the simultaneous impact from other processes in setting end-of-step qv. In particular, qr2qv_sublim_tend should also be reduced to prevent end-of-step liquid saturation.
> 
> One thing ice_supersat_conservation and the reformulated ice_overdepletion don't do is ensure that sublimation doesn't happen if start-of-step qv is supersaturated and deposition doesn't happen for subsaturated condtions... but that check is baked into the original calculation of these variables in ice_deposition_sublimation.

Thanks for the constructive discussion.
Definitely the `ice_deposition_sublimation` s/r takes care of zero out the sublimation process rate when needed.

JS-WRF-SBM:
> One other thing this PR does is move ice_overdepletion from being the first conservation check to the last. Because later conservation checks change tendencies set by earlier conservation checks, earlier checks have a larger chance of having their relationship corrupted by the end of the checking process. Thus the most important checks should be saved for last. I don't think supersaturation is more important than non-negative mass so I'm dubious of this part of the PR. Since ice_overdepletion just reduces the amount of qi and qr converted to vapor, however, I don't think it is in any danger of driving any species negative... it will just make condensate masses set by prior conservation routines to be exactly zero at the end of a step a bit larger than zero. Thus I'm approving that change too even though I don't think it's warranted.

Good point.
Please remember we have clipping mass/number conditions at the beginning of the time step.
In addition, bulk microphysics separates to modes (cloud,rain,ice), which is inherently non-correlated to some degree; the microphysical equations are solved primarily for the mass, and the number is evaluated proportionally to the mean size.
Lastly, the contribution of T, QV to the in- / out- cloud RH in terms of radiative forcing, seems quite important compared to LWC.

JS-WRF-SBM:
> First, we should never enforce ice saturation adjustment because being supersaturated with respect to ice is fine

Peter, thanks for the in-depth analysis.
Definitely my sequential breakup is not ideal.

One point regarding your comment: "First, we should never enforce ice saturation adjustment because being supersaturated with respect to ice is fine" >>  sure, but the point here is that evaporative processes (source of qv) per time step cannot lead to supersaturation, nor condensational process (sink of qv) can lead to subsaturation. 
For the rain evap case, it relates to the fact the model doesn't handle saturation over liquid (.e.g. the satadj is inherent to CLUBB, etc ..). But for the ice sublim case, being an evaporative process at constant T and P, it shouldn't lead to supersaturation above ice, as opposed to radiative-cooling, mixing or uplifting which usually change simultaneously  qv and either T (e / es)  and/or P.
This is why I enforced satadj to sublimation.

Let me know what do you think.
Please share your version when ready.

PeterCaldwell:
>the contribution of T, QV to the in- / out- cloud RH in terms of radiative forcing, seems quite important compared to LWC.

Yes, vapor is relatively transparent to radiation while condensate is relatively opaque, so getting the correct amount of condensate is important. But having negative water or clipping spurious water into existence either causes the model to crash or violates mass and energy conservation. Also, the conservation corrections we do are totally artificial and hopefully occur infrequently and make small changes, so it's hard to get too worked up about them. The correct thing to do is to run with a timestep small enough to avoid reliance on conservation fixes. If we insist on doing something we know is wrong, how much effort should we put on it?

In any case, I think we're both happy with the plan of moving ice_overdepletion to the location you want it, so it's not worth dwelling on...

PeterCaldwell:
>evaporative processes (source of qv) per time step cannot lead to supersaturation, nor condensational process (sink of qv) can lead to subsaturation.

This is a good point. I guess there are **3** conditions we're trying to satisfy:
1. sublimation doesn't push qv into ice-supersaturated territory
2. evaporation doesn't push qv into liquid-saturated territory
3. the totality of processes doesn't push end-of-step qv into liquid-supersaturated territory.

One issue with conditions 1 and 2 is that it is unclear whether they should apply to the beginning-of-step or end-of-step model state. Any top-level process has the right to expect and can easily check whether qv is liquid-supersaturated, but once we sum over microphysical processes it becomes impossible to say whether rain evap or ice sublimation was too strong. I have a hard time getting too worked up about this check, but feel like maybe we should apply it where we first calculate these quantities.

To get into details, qi2qv_sublim_tend is computed from `epsi * oabi * (qv - qv_sat_i)` in ice_deposition_sublimation. oabi is the psychrometric correction and epsi is 1/(ice relaxation timescale) so without timestepping discretization it would never undersaturate. But if dt is big and epsi is small, our forward Euler timestepping could result in undersaturation. I think  qr2qv_evap_tend is similar. Perhaps we should replace epsi with min(1.0,epsi*dt)/dt where evap and sublim are first calculated to ensure these procs never overdo what they're supposed to? This would be limiting the beginning-of-step state rather than the end-of-state tendencies like you have, but it is a lot simpler.

PeterCaldwell:
After further thought, I believe
1. we do need to ensure no liquid supersaturation (which is what my version does). Even if evap and sublim are limited to not themselves cause supersaturation, the sum of evap and sublim could still cause supersaturation. 
2. we should also ensure evap and sublimation don't individually cause liquid and ice supersaturation (respectively). The formulation for these processes should itself ensure this doesn't happen, but time discretization error could corrupt this intent.

PeterCaldwell:
In terms of how to ensure that evap and sublim don't individually cause supersaturation:
1. evaporate_rain already includes a check to ensure saturation isn't exceeded (look around line 3477 of micro_p3.F90)
2. the original prevent_ice_overdepletion includes a comment saying that a separate function for enforcing non-supersaturation is only needed because the original version of P3 allowed for multiple ice categories and we need to make sure the sum of those categories doesn't exceed ice supersaturation. Since we only allow 1 category in SCREAM-P3, it makes more sense to limit sublimation in the routine that actually calculates it... subject to the caveat below.

PeterCaldwell:
Question for @JS-WRF-SBM - the original prevent_ice_overdepletion code acted on start-of-step model state. Your new implementation acts on a state including the qv sinks over the course of the step but ignoring the qv sources. I'm unclear whether beginning of step or end of step qv is the best to use for these calculations, but this hybrid include-some, ignore-some processes approach seems suboptimal. Am I not understanding? 

On a perhaps related note, I think your scheme boils down to new_sublim_tend = old_sublim_tend/(old_sublim_tend+old_evap_tend)*(ice saturation adjustment including sink tendencies and ignoring sources). I don't understand why limiting sublimation to avoid supersaturation would be scaled by the fraction of qv source due to sublimation. Can you defend this implementation?

JS-WRF-SBM:
@PeterCaldwell 
Thanks for the brainstorming.
I'm trying to address your points here:
-  

> Question for @JS-WRF-SBM - the original prevent_ice_overdepletion code acted on start-of-step model state. Your new implementation acts on a state including the qv sinks over the course of the step but ignoring the qv sources. I'm unclear whether beginning of step or end of step qv is the best to use for these calculations, but this hybrid include-some, ignore-some processes approach seems suboptimal. Am I not understanding?
> 
> On a perhaps related note, I think your scheme boils down to new_sublim_tend = old_sublim_tend/(old_sublim_tend+old_evap_tend)*(ice saturation adjustment including sink tendencies and ignoring sources). I don't understand why limiting sublimation to avoid supersaturation would be scaled by the fraction of qv source due to sublimation. Can you defend this implementation?

@PeterCaldwell 
-  If we trust CLUBB to clear any liquid-saturation, we definitely need to enforce liquid-satadj only at the end of the time step like you did
-  As for ice-saturation, this should be free to evolve at the beginning of the time step due to mixing, radiation, advection. 
-  My approach is that individual process tendency shouldn't by itself determine the end value, except the magnitude of the relative process (mass/number tendencies). At the end of the time step we combine the leading terms (or all terms) to enforce limits so we maintain proportionality (e.g., determine same weight w.r.t to how single process changes temperature/water vapor simultaneously).
-  As for the time step: for a significant cloud mass the relaxation time for liquid and ice would likely be shorter then 1-min integration timestep used by SCREAM 3km, but the formulation is for unit of time and by multiplying 'dt' at the end of the timestep we likely allow more mass to evaporate/condensate. I do not see any problem with that; on the contrary, this is why such bulk scheme can run across multiple scales (3km or 100km), with only enforcing mass/number conservation. Otherwise, I missed your point about time stepping 
-  As for my treatment: I guess that was my way to treat the sequential nature of the code by imposing maximal possible source of `qv' that can drive ice-supersat.

PR #1061, opened 2021-05-24
[WIP] Total, rain and ice reflectivity are added to the P3 std output

JS-WRF-SBM:


components/eam/src/physics/cam/micro_p3.F90
```diff
@@ -991,6 +991,11 @@ subroutine p3_main_part3(kts, kte, kbot, ktop, kdir, &
    real(rtype)    :: qm_incld     !in-cloud qm
    real(rtype)    :: bm_incld     !in-cloud bm
 
+   real(rtype), dimension(kts:kte) :: ze_rain, ze_ice
+
+   ze_rain = 1.e-22_rtype
```
PeterCaldwell, line 996: Shouldn't these initial values be 10*log10(1e-22*1e18)=40 for consistency with the min value that's actually imposed when calculations are done?

PeterCaldwell:
This PR requires changes on the C++ side since it changes the number of outputs from `p3_main`. Note that it changes the variable names but not the actual arguments to `p3_main_part3` so no changes are needed for that proc. 

The harder way to implement needed changes would be to also have C++ also compute/pass these new output variables. The easier way is just to have the F90 unit test code read but discard these new outputs. This latter approach would mean that v1 couldn't output these new variables... but I think that's fine for the time being.

JS-WRF-SBM (commented):
I hope you can change vars name in case needed.

wrliugit:
Hi, I tried this PR in order to have some radar reflectivity results and found a minor bug. In components/eam/src/physics/cam/micro_p3.F90, lines 699-701, there should be a space. Sepcifically ,this means changing the line from 
call add_default('diag_equiv_reflectivity', 1, '')
to
call add_default('diag_equiv_reflectivity', 1, ' '). 
The code works after this change. Thanks.

AaronDonahue:
@PeterCaldwell , is this still a bug we need to fix?

PeterCaldwell:
I think this PR adds radar reflectivity as a microphysics output in v0 but we never integrated it because we don't have the same variable for v1. It is a good addition but we never had the time to prioritize adding it to v1. I realize now that we should have just made these outputs available but not output by default (by deleting the "add_default" lines). In that case, cprnc tests will pass and everything should be good. It would also be nice to have these outputs in v1, but again... not the most pressing to-do item.

PeterCaldwell:
Ok, I've modified this PR so all it does is add v0 output variables diag_equiv_reflectivity, diag_ze_rain, and diag_ze_ice *when the user explicitly asks for them*. This should solve the problem that C++/F90 cprnc comparisons failed because of new variables in F90 that we didn't add to v1. Before integrating, we should sanity check that CIME v0 tests still pass and that a v0 run asking for these 3 outputs succeeds (which it should because Weiran has been running with this change for a while). I haven't run these tests yet. I'd also like a review from @wrliugit or @hassanbeydoun to make sure the reflectivity formulae make sense in p3_micro_part3. I'm pretty sure the piping in the rest of the PR is correct.

hassanbeydoun (approved):
Looks good!

AaronDonahue:
@PeterCaldwell , can we turn this into a git-issue and fold it into the I/O epic about fields we want to add as possible outputs?

PeterCaldwell:
Unless @hassanbeydoun cares, I think we should just close this issue. Apparently nobody cares about it enough to complain about it not being done for years.

AaronDonahue:
@PeterCaldwell sounds good.  I'm going to close this for now and someone can reopen.

PR #1062, opened 2021-05-24
Changing the P3 lookup table from v4 to v4.1.1

JS-WRF-SBM:
@PeterCaldwell 

The P3 lookup table #1 was updated.
1. I've updated the original table from `v4` to `v4.1` based on the WRF P3 repository some time after starting using the SCREAM version (~1 yr ago).
2. Later on, the `v4.1` lookup table generation code had a bug in the ice aggregation part (based on personal communication with Hugh and Jason). The code was corrected. The updated tables were given the (arbitrary) number `v4.1.1`
3. Note that `p3_lookup_table_2` has no change, so on `Compy` I've just changed the name to `p3_lookup_table_2.dat-v4.1.1`

If needed, I can upload the old vs new comparison of the 1st lookup table generation code


components/eam/src/physics/cam/micro_p3_interface.F90
```diff
@@ -190,6 +190,10 @@ subroutine micro_p3_readnl(nlfile)
   select case (trim(micro_p3_tableversion))
     case ('4')
        ! Version 4 is valid
+    case ('4.1')
+      ! Version 4.1 is valid
+    case ('4.1.1')
+      ! Version 4.1.1 is valid
```
PeterCaldwell, line 196: Do we really want to continue supporting v4 and v4.1 if we think they are wrong? I'd advocate for only supporting v4.1.1.

PeterCaldwell:
Thanks for this PR, @JS-WRF-SBM ! My main question with this PR is whether the meaning of the variables we're reading in has been changed from 1 table to the next, which would require code changes in SCREAM in addition to just updating the table version. I'm pretty sure this is not the case, but it is an important sanity check. We should be able to answer this question quickly from viewing the code that produces the table lookup data for v4 and v4.1.1 - could you post that? Perhaps in a git gist?

More generally, I've been wanting to provide links to the code which generates the table lookups in the comments of the code that introduce and/or use those tables. Is this generator code under version control in WRF? Or should we add it to our own git repo?

PeterCaldwell:
Another comment is that eventually I think these table lookup files should move from the SCREAM git repo into the E3SM inputdata repo to be more consistent with E3SM protocol. This isn't something to do in this PR, I'm just mentioning it so we don't forget.

JS-WRF-SBM:
@PeterCaldwell 

- You can probably keep only the `v4.1.1`. I kept the previous version for consistency.
- The table has the same span for the output fields; the change is local in the ice-aggregation process.
- AFAIK the WRF version for table #1 includes now the 3rd moment for ice, so we do not want that. Both versions of the generation code for table #1 are included here. The notation `bf` is before correction, where `af` is after correction.
Using a diff tool between the two codes shoes the difference

[p3_crrct_generation_code_table1.gz](https://github.com/E3SM-Project/scream/files/6562657/p3_crrct_generation_code_table1.gz)

PeterCaldwell:
@AaronDonahue this PR finally passes except for p3_run_and_cmp as expected. Could you review it now?

AaronDonahue (approved):
The p3_run_and_cmp fails are expected since the baselines will change with this PR.  Approving.

PR #1063, opened 2021-05-26
SCREAM: define and enable default fpes based on SCREAM_FPE

bartgol:
Up to now, we were actually *never* enabling FPEs in scream. EKAT does not really define default fpes (see github issue https://github.com/E3SM-Project/EKAT/issues/114), so our fpe builds ended up being just...builds.

Scream should take control of defining its own default fpes, as well as enabling them when a scream session starts. The fix is simple, but it had to be manually tested. In fact, since FPE are not a language feature, but an OS feature, we cannot "capture" them. There _is_, to be fair, a way to test the FPE mask (see [here](https://github.com/E3SM-Project/EKAT/blob/master/tests/utils/debug_tools_tests.cpp) to see what EKAT does for that), but it's rather clunky, and messes with OS calls. However, if we think there would be value in it, we can run a test that verifies that the macro SCREAM_FPE indeed enables our default fpe mask, but running some buggy code and verifying that the FPE is raised.

### DIFF
--- a/components/scream/src/share/scream_session.cpp
+++ b/components/scream/src/share/scream_session.cpp
@@ -1,14 +1,27 @@
 #include "scream_session.hpp"
 #include "scream_config.hpp"
 
+#include "ekat/ekat_assert.hpp"
 #include "ekat/ekat_session.hpp"
+#include "ekat/util/ekat_feutils.hpp"
 
 #include <iostream>
 
 namespace scream {
 
+static int get_default_fpes () {
+#ifdef SCREAM_FPE
+  return (FE_DIVBYZERO |
+          FE_INVALID   |
+          FE_OVERFLOW);
+#else
+  return 0;
+#endif
+}
+
 void initialize_scream_session (bool print_config) {
   ekat::initialize_ekat_session(false);
+  ekat::enable_fpes(get_default_fpes());
 
   if (print_config) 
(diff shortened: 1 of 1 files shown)


jeff-cohere (approved):

jeff-cohere (commented):
Seems like a lot of test are failing, so maybe we should hold off on merging this till the AT issues are fixed.

bartgol:
> Seems like a lot of test are failing, so maybe we should hold off on merging this till the AT issues are fixed.

Well, I don't think they are AT issues, but actually legit FPEs. The only build that fails is the fpe one (and since weaver doesn't do it, it's the only one passing).

If we want to keep the fpe build, we have some work to do, to clean up all the fpes that flew under the radar so far. I guess we can do it as part of this PR.

jeff-cohere:
Oh, okay. I just looked at the first log and saw a bunch of things failing without any info. I think it would be good to address these issues in this PR, then. It may not be a trivial amount of work, though!

bartgol:
Yeah, I'm building right now on blake, to see what kind of fpes (and how many) we hit. There are a lot of tests, but some are just different threads/ranks configurations, so it's probably more scary than it actually is (fingers crossed).

jgfouca:
Replaced by #1090

PR #1065, opened 2021-05-26
Update ekat to include new WSM ctor

tcclevenger:
Updates ekat to include https://github.com/E3SM-Project/EKAT/pull/109


bartgol (approved):
For the record, so long as we're bringing in a limited set of changes from ekat (e.g., 1-2 PRs) I'm fine updating ekat inside the PR that needs the new ekat feature(s) (i.e., without the need of an ekat-update dedicated PR). I think dedicated submodule/upstream update prs are very useful if it's been a while since the last update, and lots of stuff has changed.

PR #1071, opened 2021-06-09
Guarantee baseline dir cleanup

### LINKED ISSUE #1064: Another baselines issue in test-all-scream
It appears test-all-scream has hit yet another corner case, and needs another fix. Several of the last AT runs on weaver (as well as the nightlies) reported this error:

```
WARNING: Failed to configure baselines:
CMake Error: The source "/home/e3sm-jenkins/weaver/workspace/SCREAM_PullRequest_Autotester_Weaver/1456/scream/components/scream/CMakeLists.txt" does not match the source "/home/e3sm-jenkins/weaver/workspace/SCREAM_PullRequest_Autotester_Weaver/1451/scream/components/scream/CMakeLists.txt" used to generate cache.  Re-run cmake with a different source directory.
```

The problem is that the baselines folder were not cleaned up when baselines were generated; the only case where this can happen, from what I can tell, is if `make baselines` returns a non-zero exit code, otherwise test-all-scream should remove all the folder content other than the `data` subfolder. I'm not sure why we don't always clean up the folder after baselines generation though.

OTOH, test-all-scream _could_ still work if the baselines folders were nuked when the scripts detects that new baselines are needed. Currently, that doesn't seem to happen (obviously, given the error).

I think there are one or two things to adjust:

1. make generate_baselines clean up the directory (except for the `data` subfolder) _regardless_ of whether `make baselines` returned 0 or not;
2. when calling `create_tests_dirs` for the baselines dir, nuke the content of all the baselines folders if any of them is deemed expired.

Note: if this happens again while the fix is still been worked on, the workaround is to login on the testbed where the error happens, and manually nuke the baselines folders.

jgfouca:
Fixes #1064

### DIFF
--- a/components/scream/scripts/test_all_scream.py
+++ b/components/scream/scripts/test_all_scream.py
@@ -565,33 +565,40 @@ remove existing baselines first. Otherwise, please run 'git fetch $remote'.
         print("Generating baseline for test {} with config '{}'".format(self._test_full_names[test], cmake_config))
         print("===============================================================================")
 
-        # We cannot just crash if we fail to generate baselines, since we would
-        # not get a dashboard report if we did that. Instead, just ensure there is
-        # no baseline file to compare against if there's a problem.
-        stat, _, err = run_cmd("{} {}".format(cmake_config, self._root_dir), from_dir=test_dir, verbose=True, dry_run=self._dry_run)
-        if stat!= 0:
-            print ("WARNING: Failed to configure baselines:\n{}".format(err))
-            return False
+        success = True
 
-        cmd = "make -j{} && make -j{} baseline".format(self._compile_res_count[test],self._testing_res_count[test])
-        if self._parallel:
-            start, end = self.get_taskset_id(test)
-            cmd = "taskset -c {}-{} sh -c '{}'".format(start,end,cmd)
+        try:
+            # We cannot just crash if we fail to generate baselines, since we would
+            # not get a dashboard report if we did that. Instead, just ensure there is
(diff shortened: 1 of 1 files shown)


components/scream/scripts/test_all_scream.py
```diff
@@ -565,33 +565,40 @@ def generate_baselines(self, test, commit):
...
-        stat, _, err = run_cmd(cmd, from_dir=test_dir, verbose=True, dry_run=self._dry_run)
+        try:
+            # We cannot just crash if we fail to generate baselines, since we would
+            # not get a dashboard report if we did that. Instead, just ensure there is
+            # no baseline file to compare against if there's a problem.
+            stat, _, err = run_cmd("{} {}".format(cmake_config, self._root_dir),
+                                   from_dir=test_dir, verbose=True, dry_run=self._dry_run)
+            if stat != 0:
+                print ("WARNING: Failed to configure baselines:\n{}".format(err))
+                success = False
```
bartgol, line 578: Shouldn't we immediately jump to the finally statement if we hit this case? There's no point in running `make` and `make baseline`, right?

bartgol (commented):
Good catch Jim! I have a question regarding early termination of baselines generation if cmake fails. See inline comment.

components/scream/scripts/test_all_scream.py
jgfouca: Yep. Good catch

bartgol (approved):

PR #1072, opened 2021-06-11
Upgrade perf infrastructure

jgfouca:
Cleanup of the run_and_cmp (both shoc and p3) and make them more configurable.

These exes should be the gold standard for measuring performance for standalone scream.

Still need to figure out why @bartgol was seeing different numbers when he was using p3_tests for measuring p3 performance.

### DIFF
--- a/components/scream/scripts/perf_analysis.py
+++ b/components/scream/scripts/perf_analysis.py
@@ -173,7 +173,7 @@ class PerfAnalysis(object):
 
         cmd = self.formulate_cmd(test_exe)
         results = []
-        with open("{}.perf.log".format(test_exe), "w") as fd:
+        with open("{}.perf.log".format(os.path.split(test_exe)[1].split(" ")[0]), "w") as fd:
             fd.write(cmd + "\n\n")
             fd.write("ENV: \n{}\n\n".format(run_cmd_no_fail("env")))
             for _ in range(self._num_runs):
@@ -199,7 +199,7 @@ class PerfAnalysis(object):
     def perf_analysis(self):
     ###############################################################################
         if self._use_existing:
-            expect(os.path.isdir("p3") and os.path.exists("CMakeCache.txt"),
+            expect(os.path.exists("CMakeCache.txt"),
                    "{} doesn't look like a build directory".format(os.getcwd()))
 
         else:
--- a/components/scream/src/physics/p3/p3_f90.cpp
+++ b/components/scream/src/physics/p3/p3_f90.cpp
@@ -35,7 +35,7 @@ FortranData::FortranData (Int ncol_, Int nlev_)
   do_predict_nc = true;
   do_prescribed_CCN = true;
   dt = -1; // model time step, s; set to invalid -1
-  it = 1;  // seems essentially unused
+  it = 1;
   // In/out
   qc              = Array2("cloud liquid water mixing ratio, kg/kg", ncol, nlev);
   nc              = Array2("cloud liquid drop number, #/kg", ncol, nlev);
--- a/components/scream/src/physics/p3/tests/p3_run_and_cmp.cpp
+++ b/components/scream/src/physics/p3/tests/p3_run_and_cmp.cpp
@@ -263,7 +263,7 @@ int main (int argc, char** argv) {
       "  -f                  Use fortran impls instead of c++. Default False.\n"
       "  -t <tol>            Tolerance for relative error. Default 0.\n"
       "  -s <steps>          Number of timesteps. Default=6.\n"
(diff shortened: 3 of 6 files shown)


components/scream/scripts/perf_analysis.py
```diff
@@ -173,7 +173,7 @@ def run_test(self, test_cmd):
 
         cmd = self.formulate_cmd(test_exe)
         results = []
-        with open("{}.perf.log".format(test_exe), "w") as fd:
+        with open("{}.perf.log".format(os.path.split(test_exe)[1].split(" ")[0]), "w") as fd:
```
bartgol, line 176: Just courious: why the second split? Shouldn't `path.split(fname)[1]` already return the file name?

components/scream/src/physics/p3/p3_f90.cpp
```diff
@@ -35,7 +35,7 @@ FortranData::FortranData (Int ncol_, Int nlev_)
   do_predict_nc = true;
   do_prescribed_CCN = true;
   dt = -1; // model time step, s; set to invalid -1
-  it = 1;  // seems essentially unused
+  it = 1;
```
bartgol, line 38: Is it used then? It seems to me it's used in `p3_run_and_cmp`, but not in here (which would make sense).

components/scream/src/physics/p3/tests/p3_run_and_cmp.cpp
```diff
@@ -261,21 +259,25 @@ int main (int argc, char** argv) {
...
-      "  -t <tol>        Tolerance for relative error. Default 0.\n"
-      "  -s <steps>      Number of timesteps. Default=6.\n"
-      "  -i <cols>       Number of columns. Default=6.\n"
-      "  -r <repeat>     Number of repetitions, implies timing run (generate + no I/O). Default=0.\n"
-      "  -p <predict_nc>  yes|no|both. Default=both.\n"
+      "  -g                  Generate baseline file. Default False.\n"
+      "  -f                  Use fortran impls instead of c++. Default False.\n"
+      "  -t <tol>            Tolerance for relative error. Default 0.\n"
+      "  -s <steps>          Number of timesteps. Default=6.\n"
+      "  -d <seconds>        Length of timestep. Default=300.\n"
```
bartgol, line 266: Very minor, but the letter `d` for `seconds` is a bit weird. Maybe we could use `-n` for the number of steps and `-s` for the step length?

components/scream/src/physics/shoc/tests/shoc_run_and_cmp.cpp
```diff
@@ -244,20 +239,28 @@ int main (int argc, char** argv) {
...
-      "  -f          Use fortran impls instead of c++.\n"
-      "  -t <tol>    Tolerance for relative error.\n"
-      "  -s <steps>  Number of timesteps. Default=10.\n"
-      "  -i <cols>   Number of columns(shcol). Default=8.\n"
-      "  -r <repeat> Number of repetitions, implies timing run (generate + no I/O). Default=0.\n";
+      "  -g                Generate baseline file.\n"
+      "  -f                Use fortran impls instead of c++.\n"
+      "  -t <tol>          Tolerance for relative error.\n"
+      "  -s <steps>        Number of timesteps. Default=10.\n"
+      "  -d <seconds>      Length of timestep. Default=150.\n"
```
bartgol, line 246: Similar comment on `-d` as for p3.

bartgol (approved):
Looks good! I'm approving, but I removed the automerge label so you have time to look at the comments first. But they are mostly minor, so I'm approving, and you can just ignore them.

components/scream/scripts/perf_analysis.py
jgfouca: path.split splits the path from the executable and args, the second split isolates the executable from the args.

components/scream/src/physics/p3/p3_f90.cpp
jgfouca: It is. It's used when the Data is passed to p3_main.

components/scream/src/physics/p3/tests/p3_run_and_cmp.cpp
jgfouca: I agree. Awkward for sure. I intended `-d` to be short for `--dt`. I'll see if I can some up with something better. I may just remove the short name for this option.

bartgol: Ah, I see. I think having just `-dt` (and `--dt` if you want both) is probably fine.

PR #1074, opened 2021-06-14
Add dynamics output fields

oksanaguba:
Adds mu, pnh, z3 (phi/g), w on gll grid.

### DIFF
--- a/components/eam/src/dynamics/se/stepon.F90
+++ b/components/eam/src/dynamics/se/stepon.F90
@@ -14,7 +14,7 @@ module stepon
    use constituents,   only: pcnst, cnst_name, cnst_longname
    use cam_abortutils, only: endrun
    use ppgrid,         only: begchunk, endchunk
-   use physconst,      only: zvir, cappa
+   use physconst,      only: zvir, cappa, gravit
    use physics_types,  only: physics_state, physics_tend
    use dyn_comp,       only: dyn_import_t, dyn_export_t
    use perf_mod,       only: t_startf, t_stopf, t_barrierf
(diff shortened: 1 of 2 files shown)


components/eam/src/dynamics/se/stepon.F90
```diff
@@ -417,6 +423,37 @@ subroutine stepon_run2(phys_state, phys_tend, dyn_in, dyn_out )
...
+
+   if (hist_fld_active('DYN_PNH')) then
+      do ie=1,nelemd
+         ! time level ntQ is not used
+         call get_field(dyn_in%elem(ie),'pnh',tmp_dyn(:,:,:,ie),hvcoord,tl_f,1)
+         call outfld('DYN_PNH',tmp_dyn(1,1,1,ie),npsq,ie)
+      enddo
+   endif
+
+   if (.not. theta_hydrostatic_mode) then
```
mt5555, line 435: Lets remove this check.  and remove this variable from the use statement above.  In hydrostatic mode, get_field_i() should do the correct thing for these variables (or abort if it can not).  that's preferable to silently not outputting them if the user requested them.

mt5555 (changes requested):

mt5555:
should this go into E3SM and get pulled into SCREAM?

oksanaguba:
I have no preference about that.

oksanaguba:
The last commit ran in nonhydro regime as expected with output of all 4 fields, the same output in hydro regime produced abort as expected, hydro regime with mu, pnh output only ran as expected.

mt5555 (approved):

oksanaguba:
closing and moving to e3sm

oksanaguba:
https://github.com/E3SM-Project/E3SM/pull/4343

PR #1075, opened 2021-06-16
Blake and weaver do not use sandia's proxy

jgfouca:
Also remove unused white.cmake


jgfouca:
Hotfix, skipping CI testing.

PR #1082, opened 2021-06-22
Some Compy machine file updates

singhbalwinder:
The code compiles fine now on Compy with these new updates.


bartgol (approved):
Looks good. I added the `AT: Integrate Without Testing` label, since there's no point in running the AT. Feel free to merge when the shiny green button appears.

PR #1083, opened 2021-06-22
SCREAM: add ability to specify version when installing python pkg

bartgol:
I was getting weird yaml files when configuring scream via CIME, and could not figure out why things where working fine for others.

Eventually, I found out the version of my pyyaml was quite old. The utility `ensure_yaml` was not checking versions, so since _a_ version of pyyaml was installed, it did not proceed to download/install a newer one.

To fix this, we can now specify (optionally) a version for the packages we need. Our python scripts will then proceed to pip install the package if the import fails (like they did before) _or_ if the import succeeds, but the version loaded is older than the minimum version (if specified).

### DIFF
--- a/components/scream/scripts/utils.py
+++ b/components/scream/scripts/utils.py
@@ -315,7 +315,7 @@ def ensure_pip():
     _ = import_module("pip")
 
 ###############################################################################
-def pip_install_lib(libname, min_version=None, pipname=None):
+def pip_install_lib(min_version=None, pipname=None):
 ###############################################################################
     """
     Ask pip to install a version of a package which is >= min_version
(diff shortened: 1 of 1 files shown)


components/scream/scripts/utils.py
```diff
@@ -298,39 +298,71 @@ def get_cpu_core_count():
...
+    import ensurepip
+    ensurepip.bootstrap(user=True)
+
+    # needed to "rehash" available libs
+    site.main() # pylint: disable=no-member
+
+    _ = import_module("pip")
+
+###############################################################################
+def pip_install_lib(min_version=None, pipname=None):
```
jgfouca, line 318: I'm wondering if the way min_version is used by this function is what we want. Pip should give you the most recent version that it can, so I don't think we should be asking for the minimum. One thing we could do is verify that the version pip gave us is above the minimum like you did in the function below.

jgfouca (changes requested):

components/scream/scripts/utils.py
bartgol: I guess a different way could be to do `pip install --upgrade $pipname` if version is present but not at the right version.

jgfouca: Yeah, that would be a nice addition.

jgfouca (approved):

PR #1085, opened 2021-06-23
Introduces a calculate_density function to the common physics functions

AaronDonahue:
Introduces a calculate_density function to the common physics functions framework.

The density calculation takes advantage of the definition of pseudo_density
passed to physics by the dynamics.  Psuedo_density assumes a shallow atmosphere,
i.e. cell area doesn't change with cell height.  As a result the expression for
density (rho) is a rather simple expression in terms of pseudo_density and dz.

This commit also introduces a set of new unit tests to test the calculate_density
function.

The calculate_density function will be needed for surface coupling, which requires the density of the bottom
layer of the atmosphere.

[BFB]

### DIFF
--- a/components/scream/src/share/tests/common_physics_functions_tests.cpp
+++ b/components/scream/src/share/tests/common_physics_functions_tests.cpp
@@ -115,7 +115,8 @@ void run(std::mt19937_64& engine)
           vmr("volume_mixing_ratio",num_mid_packs),
           mmr("mass_mixing_ratio",num_mid_packs),
           wetmmr("wet_mass_mixing_ratio",num_mid_packs),
-          drymmr("dry_mass_mixing_ratio",num_mid_packs);
+          drymmr("dry_mass_mixing_ratio",num_mid_packs),
+          density("density",num_mid_packs);
 
   auto dview_as_real = [&] (const view_1d& v) -> rview_1d {
     return rview_1d(reinterpret_cast<RealType*>(v.data()),v.size()*pack_size);
@@ -131,6 +132,7 @@ void run(std::mt19937_64& engine)
        pdf_pres(0.0,PC::P0),
        pdf_temp(200.0,400.0),
        pdf_height(0.0,1e5),
+       pdf_dz(1.0,1e5),
        pdf_surface(100.0,400.0),
        pdf_mmr(0,0.99);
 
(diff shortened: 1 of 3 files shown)


AaronDonahue:
Adding WIP label pending a determination about which formula we want to use, pseudo_density or ideal gas law.

AaronDonahue:
@jeff-cohere , per our SCREAMv1 telecon I've assigned you to review the PR and I've removed the WIP status since we've settled on the algorithm we want to use.

jeff-cohere:
@AaronDonahue , how do you want to handle the conflict(s) in `common_physics_functions_tests.cpp`? Would you like to merge the most recent changes in from the master branch?

components/scream/src/share/tests/common_physics_functions_tests.cpp
```diff
@@ -112,7 +112,8 @@ void run(std::mt19937_64& engine)
           dz("dz",num_mid_packs),
           z_int("z_int",num_int_packs),
           vmr("volume_mixing_ratio",num_mid_packs),
-          mmr("volume_mixing_ratio",num_mid_packs);
+          mmr("volume_mixing_ratio",num_mid_packs),
```
jeff-cohere, line 115: should this read `"mass_mixing_ratio"`?

AaronDonahue: yes, good catch.  Actually Balwinder caught this and changed it in his PR that just went in.  But I think this means there will be conflict, so I'll rebase and repush the branch.

components/scream/src/share/tests/common_physics_functions_tests.cpp
```diff
@@ -326,6 +347,7 @@ void run(std::mt19937_64& engine)
   for (int k=0; k<num_mid_packs; ++k) {
 
     // Make sure all results don't contain invalid numbers
+    REQUIRE( Check::is_non_negative(density_host(k),k) );
```
jeff-cohere, line 350: Anything else you'd like to test here? Is it worth checking anything based on the pseudo-density?

jeff-cohere (commented):
Approved, subject to tests passing. If you can think of any other tests for the density function, go ahead and stick them in. But it's such a simple function, I don't think it's strictly necessary.

components/scream/src/share/tests/common_physics_functions_tests.cpp
AaronDonahue: Nothing comes to mind, the density function is based on pseudo_density and I put in a comparison to the ideal-gas-law formulation up above to check consistency that way.

jeff-cohere:
Looks like there was a problem generating one of the baselines on Blake.

jeff-cohere (approved):

PR #1086, opened 2021-06-23
DP-SCREAM nightly test fix

bogensch:
Remove unneeded and problematic test modifier from DP SCREAM tests, which was causing the nightly tests to fail.

### DIFF
--- a/cime_config/tests.py
+++ b/cime_config/tests.py
@@ -392,8 +392,8 @@ _TESTS = {
             "REP.ne4pg2_ne4pg2.F2010-SCREAM-LR",
             "PEM.ne4pg2_ne4pg2.F2010-SCREAM-HR",
             "PEM.ne4pg2_ne4pg2.F2010-SCREAM-LR",
-            "SMS_R_Ln10.ne4_ne4.FDPSCREAMA97.dp-scream",
-            "ERS_R_Ln10.ne4_ne4.FDPSCREAMA97.dp-scream",
+            "SMS_R_Ln10.ne4_ne4.FDPSCREAMA97",
+            "ERS_R_Ln10.ne4_ne4.FDPSCREAMA97",
             )
     },
 


AaronDonahue:
This PR is failing the AT on mappy, but from the error log it doesn't look like these changes would trigger such a fail.  In fact, this change shouldn't impact unit tests at all.  I added the new "CIME" label which should test the CIME tests.

AaronDonahue (approved):

AaronDonahue:
The fail on mappy seems to indicate that there is an issue updating one of the submodules, which shouldn't be impacted by this PR.  I've rebased off of master to see if that moves this PR past the mappy fail.

AaronDonahue:
@jgfouca , I'm trying to figure out why this PR is failing on mappy.  From the error message it looks like it is at least during one of the FCompset tests since we are using the new "CIME" label.  So that is related to code touched in this PR.  One of the messages in the build log,

```
FATAL: Unable to delete script file /tmp/jenkins1541393845631214849.sh
```

makes me wonder if we are still running into one of those issues we've had on mappy in the past with deleting baselines.  This doesn't look like a baseline delete, but it is trying to delete a file and that may be causing the fail.

jgfouca:
@AaronDonahue , that's an internal Jenkins error. I used sudo to remove this file. Lets retest this and see what happens.

jgfouca:
@bogensch , @AaronDonahue , the internal Jenkins error is gone, which is good, but I'm concerned about this fail in the CIME testing:

```
FAIL ERS_R_Ln10.ne4_ne4.FDPSCREAMA97.mappy_gnu (phase COMPARE_base_rest)
```

I dug into the logs and found this:

```
 ANRAIN   (ncol,lev,time)  t_index =      2     2
        386    16200  (    73,    63,     1) (     1,     1,     1) (   222,    67,     1) (     2,    69,     1)
               16200   3.093962158203125E+03   0.000000000000000E+00 1.2E-07  2.980322751682252E-04 1.3E-04  0.000000000000000E+00
               16200   3.093962158203125E+03   0.000000000000000E+00          2.981558209285140E-04          2.941730414818267E-09
               16200  (    73,    63,     1) (     1,     1,     1)
          avg abs field values:    6.791586279869080E-01    rms diff: 4.9E-09   avg rel diff(npos):  1.3E-04
                                   6.791586279869080E-01                        avg decimal digits(ndif):  4.7 worst:  0.0
 RMS ANRAIN                           4.9240E-09            NORMALIZED  7.2501E-09

...

 AQRAIN   (ncol,lev,time)  t_index =      2     2
        363    16200  (   175,    65,     1) (     1,     1,     1) (    76,    68,     1) (     2,    69,     1)
               16200   3.804887782621336E-08   0.000000000000000E+00 2.5E-14  2.859786296660616E-10 1.3E-04  0.000000000000000E+00
               16200   3.804887782621336E-08   0.000000000000000E+00          2.860041092844767E-10          3.449282935597532E-17
               16200  (   175,    65,     1) (     1,     1,     1)
          avg abs field values:    4.242858842040675E-11    rms diff: 2.2E-15   avg rel diff(npos):  1.3E-04
                                   4.242882087335254E-11                        avg decimal digits(ndif):  4.3 worst:  0.0
 RMS AQRAIN 

...

 FREQR   (ncol,lev,time)  t_index =      2     2
        450    16200  (    73,    64,     1) (     1,     1,     1) (     2,    67,     1) (     2,    69,     1)
               16200   4.097498604096472E-04   0.000000000000000E+00 3.0E-05  3.999999898951501E-05 5.7E-03  0.000000000000000E+00
               16200   4.097498604096472E-04   0.000000000000000E+00          7.000000186963007E-05          1.999999949475750E-05
               16200  (    73,    64,     1) (     1,     1,     1)
          avg abs field values:    1.003868692350807E-05    rms diff: 1.7E-06   avg rel diff(npos):  5.7E-03
                                   1.032303771353327E-05                        avg decimal digits(ndif):  0.7 worst:  0.0
 RMS FREQR                            1.7311E-06            NORMALIZED  1.7003E-01

...

 A total number of    439 fields were compared
          of which      3 had non-zero differences
```

So, restart runs are giving us different results than non-restart runs in the fields ANRAIN, AQRAIN, FREQR

AaronDonahue:
@jgfouca , yes, this looks like a fail we should investigate further.  @bogensch reported that the tests pass on his machine, not sure why they would fail elsewhere, but it is something we should investigate.  Peter B. is on vacation, but it is something we can take a look at together when he gets back.

AaronDonahue:
@jgfouca , I wonder if we could still merge this PR now and let the new set of fails get into master?  We won't lose sight of them since they will keep popping up on CDash but at least CDash will move from showing build failures to runtime failures which will be more informative.  Since very few PR's use the AT:CIME Label this shouldn't impact development but will at least let us move on from this PR to a subsequent fix.

Thoughts?

jgfouca:
@AaronDonahue , if that's what you want, I'm happy to do it. I can use my admin powers to merge this PR. Please confirm and I will do that.

bogensch:
@jgfouca sorry for my radio silence.  Thanks for digging into the logs.  It looks like this is proof that restarts in DP-SCREAM are not completely air tight since they are failing on Jenkins and I have more tedious debugging to do.  I support this PR going in if @AaronDonahue does and I would issue the fix in separate PR.

AaronDonahue:
@jgfouca or @PeterCaldwell , I don't have the authority to override the AT and force merge this PR.  Could one of you do that?  I'm in favor of putting this in so that the test fails point to the actual problem (see the conversation above) and having @bogensch  put that fix in as a separate PR.

jgfouca:
@AaronDonahue done.

PR #1087, opened 2021-06-24
P3 - Recent Fixes/Changes in P3

JS-WRF-SBM:
@PeterCaldwell @AaronDonahue 

There are 3 changes in P3 microphysics as described below (following Hugh Morrison and Jason Milbrandt fixes):
1. Ice self-collection: since the lookup table is normalized by `Nitot`, the self-collection process which is double integral over the ice category sizes needs a normalization of `Nitot^2`. This was overlooked by the authors since the `Nitot` normalisations was applied to the few latest versions.
2. Ice collection efficiency: as a result of change (1) which forces larger ice particles, the ice collection efficiency - being a tunable profile depend on temperature - was changed toward lower values. Independent of the ice self-collection bug, there are publications which support these low values of ice collection efficiency in cold temperature, so it is justified.
3. The `nmltratio` tuning parameter was changed to 1.0, thus eliminating it's role in the transition between ice number to rain number during melting.

Overall, the changes in the climatology of a 1-deg simulation are small since the authors likely managed to balance (1) and (2) to some extent. We do see small change in the effective radius, and larger changes in the mean-mass radius (as expected).
Ice mass decreases a little and so is the ice number. We do see small changes also in the rain number concentration due to (3).

I would be interested in seeing the 3km convective-permitting test of these since I would argue the 1-deg still mask some of the changes magnitude, and in case needed we can use the collection efficiency as a tuning profile.


PeterCaldwell:
Ok, passes for all tests except p3_run_and_cmp - which is expected since we are changing some formulae and hence need new baselines. @AaronDonahue - can you review and integrate this now?

AaronDonahue (approved):

AaronDonahue:
This is non-BFB PR, now that CDash is reporting all-green I'm going to manually merge and we will need new baselines generated.

PR #1088, opened 2021-06-24
Functions to convert dry to wet mixing ratios and vice versa

singhbalwinder:
Adds functions and tests which converts dry to wet and wet to dry mixing ratios. Some simple tests are also being added.

### DIFF
--- a/components/scream/src/share/tests/common_physics_functions_tests.cpp
+++ b/components/scream/src/share/tests/common_physics_functions_tests.cpp
@@ -216,13 +216,12 @@ void run(std::mt19937_64& engine)
   REQUIRE( Check::equal(PF::calculate_dse(zero,zero,surf_height),ScalarT(surf_height)) );
   REQUIRE( Check::equal(PF::calculate_dse(ScalarT(inv_cp),ScalarT(1/g),surf_height),ScalarT(surf_height+2.0)) );
 
-  // DRYMMR to WETMMR (and vice versa) property tests
+  // WETMMR to DRYMMR (and vice versa) property tests
   wetmmr0 = pdf_mmr(engine);// get initial inputs for wetmmr_from_drymmr and drymmr_from_wetmmr functions
   qv0  = pdf_qv(engine);  // This is an input for mmr_tests, so it won't be modified by mmr tests
   // mmr_test1: For zero drymmr, wetmmr should be zero
   // mmr_test2: For zero wetmmr, drymmr should be zero
   // mmr_test3: Compute drymmr from wetmmr0 and then use the result to compute wetmmr, which should be equal to wetmmr0
-
 
   REQUIRE( Check::equal(PF::calculate_wetmmr_from_drymmr(zero,qv0),zero) ); //mmr_test1
   REQUIRE( Check::equal(PF::calculate_drymmr_from_wetmmr(zero,qv0),zero) ); //mmr_test2
--- a/components/scream/src/share/tests/common_physics_functions_tests.cpp
+++ b/components/scream/src/share/tests/common_physics_functions_tests.cpp
@@ -101,7 +101,8 @@ void run(std::mt19937_64& engine)
           pressure("pressure",num_mid_packs),
           pseudo_density("pseudo_density",num_mid_packs),
           dz_for_testing("dz_for_testing",num_mid_packs),
-          mmr_for_testing("mass_mixing_ratio_for_testing",num_mid_packs);
+          mmr_for_testing("mass_mixing_ratio_for_testing",num_mid_packs),
+          wetmmr_for_testing("wet_mass_mixing_ratio_for_testing",num_mid_packs);
   // Output views
   view_1d exner("exner",num_mid_packs),
           theta("theta",num_mid_packs),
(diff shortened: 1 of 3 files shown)


singhbalwinder:
I added some naive tests but if you guys think I can add more tests, please let me know.

components/scream/src/share/tests/common_physics_functions_tests.cpp
```diff
@@ -274,6 +290,13 @@ void run(std::mt19937_64& engine)
     // Compute vmr from mmr and vice versa
     PF::calculate_vmr_from_mmr(team,h2o_mol,qv,mmr_for_testing,vmr);
     PF::calculate_mmr_from_vmr(team,h2o_mol,qv,vmr,mmr);
+
+    // Compute wetmmr assuming mmr is dry
+    PF::calculate_wetmmr_from_drymmr(team,mmr,qv,wetmmr);//get wetmmr from mmr, assuming mmr is drymmr
+
+    // Convert wetmmr computed above to drymmr (i.e. mmr)
+    PF::calculate_drymmr_from_wetmmr(team,wetmmr,qv,mmr);//convert it back to drymmr, i.e. mmr
```
AaronDonahue, line 298: The variable `mmr` is already being used for the vmr/mmr tests and is still being used later in the test code, see line 342.  Could we create a new variable, call it `mmrdry` that you could use for the wet/dry tests so that these don't conflict?

components/scream/src/share/tests/common_physics_functions_tests.cpp
```diff
@@ -274,6 +290,13 @@ void run(std::mt19937_64& engine)
     // Compute vmr from mmr and vice versa
     PF::calculate_vmr_from_mmr(team,h2o_mol,qv,mmr_for_testing,vmr);
     PF::calculate_mmr_from_vmr(team,h2o_mol,qv,vmr,mmr);
+
+    // Compute wetmmr assuming mmr is dry
+    PF::calculate_wetmmr_from_drymmr(team,mmr,qv,wetmmr);//get wetmmr from mmr, assuming mmr is drymmr
```
AaronDonahue, line 295: These tests look great, and thanks for catching my comment error in the vmr,mmr portion.

If you follow the example of vmr->mmr and vice versa there is also a few checks below where we deep copy to host the output from these operations on the view and do some checks, like not negative.  Could you add that in too?

You could also follow the vmr/mmr example and check that `wetmmr` is approx_equal to `mmr` for all data in the view.  See line 342.

AaronDonahue (changes requested):
Looks great! Thank you Balwinder.  Just one set of changes requested in the tests themselves and then I think this will be good to go.

components/scream/src/share/tests/common_physics_functions_tests.cpp
singhbalwinder: Thanks! It is a good idea. I created a new variable named `drymmr0`. My tests were converting mmr back to its original value, so I didn't create a new variable earlier. You are right, it is not safe to rely on that.

components/scream/src/share/tests/common_physics_functions_tests.cpp
singhbalwinder: I was converting it back to `mmr` here as well which is tested downstream so I didn't add the code for wet and dry mmr. More testing is always better, so I followed your suggestion to add checks for both dry and wet mmr. I inserted a loop for assigning `mmr` to `drymmr`. Let me know if there is a better way to handle this assignment.

components/scream/src/share/tests/common_physics_functions_tests.cpp
```diff
@@ -317,6 +349,9 @@ void run(std::mt19937_64& engine)
 
     // Check vmr(mmr(vmr))==mmr (up to roundoff tolerance)
     REQUIRE ( Check::approx_equal(mmr_host(k),mmr_for_testing_host(k),test_tol) );
+
+    // Check drymmr==mmr (up to roundoff tolerance)
+    REQUIRE ( Check::approx_equal(drymmr_host(k),mmr_for_testing_host(k),test_tol) );
```
AaronDonahue, line 354: @singhbalwinder , yes, these changes look better.  Although I still think it would be good to have the mmr variables in the wet/dry mmr tests be completely independent from the mmr/vmr variables.  I would suggest making a `drymmr_for_testing` variable, like what is done with `mmr_for_testing` and then testing against that here.  If you do that, you will need to use `drymmr_for_testing` in line 300 above.  That way you are testing that if you start with `drymmr_for_testing` and then convert to `wetmmr` and then back to `drymmr` the `drymmr_for_testing` and `drymmr` match.

AaronDonahue: You will also want to follow the example on line 141 to generate random values for `drymmr_for_testing`.

Also, a total side note, but the mmr/vmr conversions actually assume that mmr is wet given that this will be the scream default.   So the use of the `mmr` variable as "dry" could cause confusion.  But here, if we make the variables independent we won't have to worry about that.

AaronDonahue (changes requested):
Just a follow up on the new test variables.

components/scream/src/share/tests/common_physics_functions_tests.cpp
singhbalwinder: Thanks for the very thoughtful suggestions. I have created variables which are completely independent for these functions. I have also changed the tests so that they assume wet mixing ratios to start with. I went through the code again but let me know if I need to change anything or if there are any typos/misleading comments.

AaronDonahue (approved):
Excellent, thanks for being so quick to adjust the PR @singhbalwinder .  It looks great now.

components/scream/src/share/util/scream_common_physics_functions.hpp
```diff
@@ -236,6 +260,20 @@ struct PhysicsFunctions
                              const Real surf_geopotential,
                              const view_1d<ScalarT>& dse);
 
+  template<typename ScalarT, typename InputProviderX, typename InputProviderQ>
+  KOKKOS_INLINE_FUNCTION
+  static void calculate_wetmmr_from_drymmr (const MemberType& team,
+                             const InputProviderX& drymmr,
+                             const InputProviderQ& qv,
+                             const view_1d<ScalarT>& wetmmr);
```
PeterCaldwell, line 268: I'm probably just being dense here, but why is the output wetmmr being declared const here?

AaronDonahue: I believe that since we are passing by reference (&) that the reference is being declared as `const`, but the data in wetmmr can still be changed.

PeterCaldwell: Ok, that's what I suspected. Thanks.

PeterCaldwell (approved):
Thanks Balwinder, looks good!

PR #1090, opened 2021-06-24
Jgfouca/fix fpe testing

jgfouca:
Change list:
1) Fixes SCREAM_FPE so that FPEs are actually on when it is defined (merged this from a previous branch by @bartgol )
2) Adds a test to scripts-tests to ensure that FPEs are thrown when SCREAM_FPE is on
3) Fix all FPEs in P3, none of these was hard or concerning
4) Fix all FPEs in SHOC, there was a floating point overflow that appears to be a real issue. I will annotate it inline.
5) Fix all FPEs in rrtmgp. One of these was very concerning and required a change to the submodule in addition to disabling some tests that started breaking when I put this fix in.
6) Add new SCREAM_AUTOTESTER cmake setting. Allows behavior changes when a run is an AT run such as disabling certain tests.
7) Disable shoc_cld_p3_rrtmgp


components/eam/src/physics/cam/micro_p3.F90
```diff
@@ -2936,7 +2936,7 @@ subroutine nr_conservation(nr,ni2nr_melt_tend,nr_ice_shed_tend,ncshdc,nc2nr_auto
   !recall that melting number is scaled by nmltratio to account for ice crystals melting then
   !rapidly evaporating. Thus ni removal from melting is *not* scaled by nmltratio...
   source_nr = nr + (ni2nr_melt_tend*nmltratio + nr_ice_shed_tend + ncshdc + nc2nr_autoconv_tend)*dt
-  if(sink_nr > source_nr) then
+  if(sink_nr > source_nr .and. sink_nr.gt.0._rtype) then
```
jgfouca, line 2939: I think these should probably be != 0.

components/scream/scripts/scripts-tests
```diff
@@ -336,13 +336,25 @@ class TestTestAllScream(TestBaseOuter.TestBase):
...
+            self.skipTest("Skipping full run")
+
     def test_run_diff_captured(self):
         if self._full:
             cmd = self.get_cmd("./test-all-scream -e SCREAM_FORCE_RUN_DIFF=True -m $machine -b HEAD -k -t dbg", self._machine, dry_run=False)
             run_cmd_assert_result(self, cmd, from_dir=TEST_DIR, expect_works=False)
         else:
             self.skipTest("Skipping full run")
 
+
```
jgfouca, line 356: Don't need these newlines.

components/scream/src/physics/p3/p3_ni_conservation_impl.hpp
```diff
@@ -17,7 +17,7 @@ void Functions<S,D>::ni_conservation(const Spack& ni, const Spack& ni_nucleat_te
 {
   const auto sink_ni = (ni2nr_melt_tend + ni_sublim_tend + ni_selfcollect_tend)*dt;
   const auto source_ni = ni + (ni_nucleat_tend+nr2ni_immers_freeze_tend+nc2ni_immers_freeze_tend)*dt;
-  const auto mask = sink_ni > source_ni && context;
+  const auto mask = sink_ni > source_ni && context && sink_ni > 0;
```
jgfouca, line 20: != 0

components/scream/src/physics/p3/p3_nr_conservation_impl.hpp
```diff
@@ -17,7 +17,7 @@ void Functions<S,D>::nr_conservation(const Spack& nr, const Spack& ni2nr_melt_te
 {
   const auto sink_nr = (nr_collect_tend + nr2ni_immers_freeze_tend + nr_selfcollect_tend + nr_evap_tend)*dt;
   const auto source_nr = nr + (ni2nr_melt_tend*nmltratio + nr_ice_shed_tend + ncshdc + nc2nr_autoconv_tend)*dt;
-  const auto mask = sink_nr > source_nr && context;
+  const auto mask = sink_nr > source_nr && context && sink_nr > 0;
```
jgfouca, line 20: != 0

components/scream/src/physics/share/physics_test_data.hpp
```diff
@@ -40,23 +40,23 @@ struct SHOCGridData : public PhysicsTestData {
...
-#define PTD_ZEROES5 PTD_ZEROES4, 0
-#define PTD_ZEROES6 PTD_ZEROES5, 0
-#define PTD_ZEROES7 PTD_ZEROES6, 0
-#define PTD_ZEROES8 PTD_ZEROES7, 0
-#define PTD_ZEROES9 PTD_ZEROES8, 0
-#define PTD_ZEROES10 PTD_ZEROES9, 0
-#define PTD_ZEROES11 PTD_ZEROES10, 0
-
-#define PTD_ZEROES(a) PTD_ZEROES##a
+#define PTD_ONES0
```
jgfouca, line 43: We just want to initialize the Data struct with garbage, it's going to be assigned in the next line anyway. 1 garbage is less likely to be a problem since some of the Data constructors have inv_* members that divide.

components/scream/src/physics/shoc/shoc_assumed_pdf_impl.hpp
```diff
@@ -263,11 +270,17 @@ void Functions<S,D>::shoc_assumed_pdf(
         std_s1 = ekat::sqrt(ekat::max(0,
                                       ekat::square(cthl1)*thl2_1
                                       + ekat::square(cqt1)*qw2_1 - 2*cthl1*sqrtthl2_1*cqt1*sqrtqw2_1*r_qwthl_1));
+        const auto std_s1_not_small = std_s1 > 2.0E-155;
```
jgfouca, line 273: Here is the overflow problem on the C++ side. With very small std_s1, the division below can overflow max float.

components/scream/src/physics/shoc/shoc_assumed_pdf_impl.hpp
```diff
@@ -284,15 +297,26 @@ void Functions<S,D>::shoc_assumed_pdf(
...
-        qn2.set(!equal && std_s2 != 0 && C2 != 0, s2*C2+(std_s2/sqrt2pi)*ekat::exp(-sp(0.5)*ekat::square(s2/std_s2)));
-        qn2.set(!equal && std_s2 == 0 && s2 > 0, s2);
+        const auto nequal = !equal;
+        if (nequal.any()) {
+          std_s2.set(nequal,
+                     ekat::sqrt(ekat::max(0,
+                                          ekat::square(cthl2)*thl2_2
+                                          + ekat::square(cqt2)*qw2_2 - 2*cthl2*sqrtthl2_2*cqt2*sqrtqw2_2*r_qwthl_1)));
+          s2.set(nequal, qw1_2-qs2*((1 + beta2*qw1_2)/(1 + beta2*qs2)));
+          const auto std_s2_not_small = std_s2 > 2.0E-155;
```
jgfouca, line 307: Another overflow check here.

jgfouca (commented):
Annotations.

components/scream/src/physics/rrtmgp/tests/rrtmgp_tests.cpp
```diff
@@ -113,11 +113,11 @@ int main(int argc, char** argv) {
         lw_flux_up_ref, lw_flux_dn_ref
     );
     int nerr = 0;
-    if (!rrtmgpTest::all_equals(sw_flux_up_ref , sw_flux_up )) nerr++;
-    if (!rrtmgpTest::all_equals(sw_flux_dn_ref , sw_flux_dn )) nerr++;
-    if (!rrtmgpTest::all_equals(sw_flux_dir_ref, sw_flux_dir)) nerr++;
-    if (!rrtmgpTest::all_equals(lw_flux_up_ref , lw_flux_up )) nerr++;
-    if (!rrtmgpTest::all_equals(lw_flux_dn_ref , lw_flux_dn )) nerr++;
+    // if (!rrtmgpTest::all_equals(sw_flux_up_ref , sw_flux_up )) nerr++;
```
jgfouca, line 116: @brhillman , had to disable these once the rrtmgp FPE was fixed.

components/scream/tests/uncoupled/rrtmgp/rrtmgp_stand_alone.cpp
```diff
@@ -278,8 +278,8 @@ namespace scream {
         REQUIRE(rrtmgpTest::all_equals(sw_flux_up_ref    , sw_flux_up_test  ));
         REQUIRE(rrtmgpTest::all_equals(sw_flux_dn_ref    , sw_flux_dn_test    ));
         REQUIRE(rrtmgpTest::all_equals(sw_flux_dn_dir_ref, sw_flux_dn_dir_test));
-        REQUIRE(rrtmgpTest::all_equals(lw_flux_up_ref    , lw_flux_up_test    ));
-        REQUIRE(rrtmgpTest::all_equals(lw_flux_dn_ref    , lw_flux_dn_test    ));
+        //REQUIRE(rrtmgpTest::all_equals(lw_flux_up_ref    , lw_flux_up_test    ));
```
jgfouca, line 281: @brhillman , had to disable these once the rrtmgp FPE was fixed too.

jgfouca (commented):
Some additional notes for Ben.

components/eam/src/physics/cam/micro_p3.F90
ambrad: Does the original code actually cause an FPE? I think source_nr must be >= 0. Thus, if sink_nr > source_nr, then sink_nr > 0, causing no issues in source_nr/sink_nr.

components/scream/src/physics/p3/p3_main_impl.hpp
```diff
@@ -799,12 +799,14 @@ ::p3_main_part3(
       get_cloud_dsd2(qc_incld, nc_incld, mu_c(k), rho(k), nu(k), dnu, lamc(k), ignore1, ignore2, qc_gt_small);
 
       nc(k).set(qc_gt_small,nc_incld*cld_frac_l(k)); //cld_dsd2 might have changed incld nc... need consistency.
-      diag_eff_radius_qc(k)       .set(qc_gt_small, sp(0.5) * (mu_c(k) + 3) / lamc(k));
-      qv(k)              .set(qc_small, qv(k)+qc(k));
-      th_atm(k)              .set(qc_small, th_atm(k)-inv_exner(k)*qc(k)*latent_heat_vapor(k)*inv_cp);
-      vap_liq_exchange(k).set(qc_small, vap_liq_exchange(k) - qc(k));
-      qc(k)              .set(qc_small, 0);
-      nc(k)              .set(qc_small, 0);
+      if (qc_gt_small.any()) {
```
ambrad, line 802: This code would be improved by one overall `if (qc_gt_small.any())`. That would then naturally account for the /0 check.

components/eam/src/physics/cam/micro_p3.F90
jgfouca: @ambrad , yes. My methodology was:
1) Build SCREAM with SCREAM_FPE
2) RUN ctest
3) Pick one of the failing tests
4) Run it through gdb, find FPE line, fix it, run it again, repeat until test passes

So, I don't think I touched anything in this PR unless it was to fix an FPE. I can double-check this particular change.

components/scream/src/physics/p3/p3_main_impl.hpp
```diff
@@ -799,12 +799,14 @@ ::p3_main_part3(
       get_cloud_dsd2(qc_incld, nc_incld, mu_c(k), rho(k), nu(k), dnu, lamc(k), ignore1, ignore2, qc_gt_small);
 
       nc(k).set(qc_gt_small,nc_incld*cld_frac_l(k)); //cld_dsd2 might have changed incld nc... need consistency.
```
ambrad, line 801: There are two chunks of code here that depend on two different masks. Setting up two `if (mask.any())` blocks would be natural and increase performance. As a side effect, it would take care of the /0 case.

components/scream/src/physics/p3/p3_main_impl.hpp
```diff
@@ -822,12 +824,14 @@ ::p3_main_part3(
       //Note that integrating over the drop-size PDF as done here should only be done to in-cloud
       //quantities but radar reflectivity is likely meant to be a cell ave. Thus nr in the next line
       //really should be cld_frac_r * nr/cld_frac_r. Not doing that since cld_frac_r cancels out.
-      ze_rain(k).set(qr_gt_small, nr(k)*(mu_r(k)+6)*(mu_r(k)+5)*(mu_r(k)+4)*
-                     (mu_r(k)+3)*(mu_r(k)+2)*(mu_r(k)+1)/pow(lamr(k), sp(6.0))); // once f90 is gone, 6 can be int
+      if (qr_gt_small.any()) {
+        ze_rain(k).set(qr_gt_small, nr(k)*(mu_r(k)+6)*(mu_r(k)+5)*(mu_r(k)+4)*
```
ambrad, line 828: Same here.

components/scream/src/physics/p3/p3_ni_conservation_impl.hpp
ambrad: See my comment above re: the same code in the F90. Sorry if I'm just confused.

components/eam/src/physics/cam/micro_p3.F90
jgfouca: @ambrad , I stand corrected. This check was not necessary.

components/scream/src/physics/p3/p3_nr_conservation_impl.hpp
ambrad: Same as sink_ni.

components/scream/src/physics/p3/p3_ni_conservation_impl.hpp
jgfouca: @ambrad , this change does appear to be necessary. When I remove it, I get FPEs in all the p3_tests.

components/scream/src/physics/p3/p3_nr_conservation_impl.hpp
jgfouca: @ambrad , this change does appear to be necessary. When I remove it, I get FPEs in all the p3_tests.

components/scream/src/physics/shoc/shoc_assumed_pdf_impl.hpp
```diff
@@ -146,17 +146,19 @@ void Functions<S,D>::shoc_assumed_pdf(
...
+                             ekat::max(0, (3*tmp_val_1*(1 - a*ekat::square(tmp_val_2) - (1-a)*ekat::square(tmp_val_1))
+                                           - (Skew_thl - a*ekat::cube(tmp_val_2) - (1 - a)*ekat::cube(tmp_val_1)))
+                                       /(3*a*(tmp_val_1 - tmp_val_2))))*thlsec);
+        thl2_2.set(condition,
+                   ekat::min(100,
                            ekat::max(0, (-3*tmp_val_2*(1 - a*ekat::square(tmp_val_2)
                                          - (1 - a)*ekat::square(tmp_val_1))
                                          + (Skew_thl - a*ekat::cube(tmp_val_2) - (1 - a)*ekat::cube(tmp_val_1)))
                                          /(3*(1 - a)*(tmp_val_1 - tmp_val_2))))*thlsec);
+      }
```
ambrad, line 161: Looks like there is the same opportunity here to do one big `condition.any()` check to either run or not run a big chunk of code. That will then naturally solve the /0 problem.

components/eam/src/physics/cam/micro_p3.F90
jgfouca: @ambrad sorry for the confusion; this change does appear to be necessary (I tested my non-FPE build on accident when I said it was not necessary).

components/scream/src/physics/shoc/shoc_assumed_pdf_impl.hpp
ambrad: If possible, use a numeric_limits<Real> function so this threshold works for single precision too.

components/eam/src/physics/cam/micro_p3.F90
ambrad: Is source_nr going negative? If it is, we should bring this to the attention of Aaron. Otherwise, I'm not seeing how this code triggers an FPE.

jgfouca: @ambrad , I remove the `.and. sink_nr.gt.0._rtype` and added a print inside the if block. The values printed immediately before the FPE crash were:

```
sink_nr= -0.00000000000000000000000000000000E+00 source_nr= -3.63189856026829738233629996102536E-01
```

ambrad:
I believe dz_zi(k) is a fatal error, so the dz_zi cases might just be an artifact of the testing inputs. If so, it's definitely fair to make the dz testing inputs always > 0.

components/eam/src/physics/cam/micro_p3.F90
ambrad: @AaronDonahue is source_nr < 0 and sink_nr = 0 a physically valid state? If not, we need to figure out what's leading to it.

components/eam/src/physics/cam/shoc.F90
```diff
@@ -2855,9 +2855,9 @@ subroutine shoc_assumed_pdf_compute_s(&
   qn=0._rtype
   C=0._rtype
 
-  if (std_s .ne. 0.0_rtype) then
+  if (std_s .gt. 2.0E-155_rtype) then
```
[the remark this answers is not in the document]
bartgol: You could use `tiny(X)` (with X any number of kind rtype). It seems to be part of the fortran standard, at least f95 and later.

components/scream/src/physics/p3/p3_ni_conservation_impl.hpp
bartgol: I think, as stated by Andrew in another comment, that we might have to check with Aaron. A negative sink is a source, but since we have two distinct vars for sink and sources, I would expect both of them to be positive. If we are not "wasting" time to clip them to 0, then I understand, and this check is fine, but if they were not supposed to get negative in the first place, then we should do a `!=0` check (so that a <0 val can still generate errors, and reveal bugs).

components/scream/src/physics/p3/tests/p3_rain_sed_unit_tests.cpp
```diff
@@ -52,20 +52,21 @@ static void run_bfb_rain_vel()
     {5.6298E-05, 1.1129E+00, 1.6576E+02},
     {1.0000E-02, 1.0774E+00, 1.9436E+02},
 
-    {0.0       , 1.3221E+00, 6.2964E+05},
+    {1.1030E-04, 1.3221E+00, 6.2964E+05},
```
bartgol, line 55: Maybe there was a reason to test 0.0? Like stress-testing some conditionals?

components/scream/src/physics/shoc/shoc_assumed_pdf_impl.hpp
```diff
@@ -146,17 +146,19 @@ void Functions<S,D>::shoc_assumed_pdf(
...
-                                         - (Skew_thl - a*ekat::cube(tmp_val_2) - (1 - a)*ekat::cube(tmp_val_1)))
-                                         /(3*a*(tmp_val_1 - tmp_val_2))))*thlsec);
-      thl2_2.set(condition,
-                 ekat::min(100,
+      if (condition.any()) {
+        thl2_1.set(condition,
+                   ekat::min(100,
+                             ekat::max(0, (3*tmp_val_1*(1 - a*ekat::square(tmp_val_2) - (1-a)*ekat::square(tmp_val_1))
+                                           - (Skew_thl - a*ekat::cube(tmp_val_2) - (1 - a)*ekat::cube(tmp_val_1)))
+                                       /(3*a*(tmp_val_1 - tmp_val_2))))*thlsec);
```
bartgol, line 154: This expression is quite complicated. Is it guaranteed that if condition is true, then the denominator is nonzero? To me, it's hard to tell. Might be that with this mod the test passes, but I can't tell if it just "happens" to pass, or if it will "always" pass, even when we add more scream tests (e.g., fully coupled scream, with real data).

components/scream/src/physics/shoc/shoc_assumed_pdf_impl.hpp
```diff
@@ -284,15 +297,26 @@ void Functions<S,D>::shoc_assumed_pdf(
...
-                   ekat::sqrt(ekat::max(0,
-                                        ekat::square(cthl2)*thl2_2
-                                        + ekat::square(cqt2)*qw2_2 - 2*cthl2*sqrtthl2_2*cqt2*sqrtqw2_2*r_qwthl_1)));
-        s2.set(!equal, qw1_2-qs2*((1 + beta2*qw1_2)/(1 + beta2*qs2)));
-        C2.set(!equal && std_s2 != 0, sp(0.5)*(1 + ekat::erf(s2/(sqrt2*std_s2))));
-        C2.set(!equal && std_s2 == 0 && s2 > 0, 1);
-        qn2.set(!equal && std_s2 != 0 && C2 != 0, s2*C2+(std_s2/sqrt2pi)*ekat::exp(-sp(0.5)*ekat::square(s2/std_s2)));
-        qn2.set(!equal && std_s2 == 0 && s2 > 0, s2);
+        const auto nequal = !equal;
+        if (nequal.any()) {
```
bartgol, line 301: You could do `if (not equal.none())`. Mighr help save a register on GPU.

jgfouca:
Hi all,

I'm going to need some help getting this across the finish line. Unhiding our FPEs has revealed some concerns/errors in the code that are real. Here are the points on which I'd like some help/feedback.
1) @AaronDonahue , in P3 for nr_conservation and ni_conservation, we are sometimes seeing the source=0 and the sink<0. Is this valid/expected? This appears to only happen for unit tests, not p3_run_and_cmp, so maybe the unit test inputs are bad?
2) @bogensch , in SHOC for shoc_assumed_pdf_compute, the value of std_s is getting small enough that a subsequent division by std_s is causing a result that overflows a double. Through printing and trial&error, I found that limiting std_s to be no smaller than `2.0E-155` (just treat it like zero if it is) fixes this problem. I don't know if this is how we want to handle this situation. This appears to only happen for unit tests, not shoc_run_and_cmp, so maybe the unit test inputs are bad?
3) @bartgol , when I initially developed this PR, I did not have `-DSCREAM_DYNAMICS_DYCORE=HOMME` , so I did not notice until the AT returned that there are some FPEs coming from Homme as well. I had to do a fair bit of digging but I found out some info that is hopefully useful:
The failing stack trace is as follows:
```
test
  DynamicsDrivenGridsManager::build_grids
    homme_grid_mod::init_grids_f90
      dyn_grid_mod::dyn_grid_init
        prim_driver_base::prim_init1_geometry
          geometry_mod::set_area_correction_map0
            cube_mod::cube_init_atomic
              cube_mod::metric_atomic (FPE on this line: elem%Dinv = elem%Dinv / sqrt(alpha) )
```
The FPE occurs because `alpha` is zero. `alpha` is zero because in geometry_mod::set_area_correction_map0, the value that becomes `alpha` lower in stack is computed via `area(1) = domain_size/area(1)  ! ratio correction`. The value is zero because `domain_size` is zero. `domain_size` comes from the `physical_constants` module and it should have been initialized by `namelist_mod::readnl` but apparently this subroutine was never called. Based on this, it seems like the homme_pd_remap_tests need to call `namelist_mod::readnl`. If this is correct, what's the easiest way to do this? I see a couple nl files living alongside this test `theta-ndays.nl  theta-nmax.nl`.

AaronDonahue:
Per:

```
@AaronDonahue , in P3 for nr_conservation and ni_conservation, we are sometimes seeing the source=0 and the sink<0. Is this valid/expected? This appears to only happen for unit tests, not p3_run_and_cmp, so maybe the unit test inputs are bad?
```

Looking at the formulation of both source and sink in nr_conservation and ni_conservation, it should not be possible to get sink<0.  In both cases the sink term is a sum of what should be positive tendencies.  You mentioned that you see negative sink in the unit tests?  I'll start there and investigate what is going on.  I'll comment back here if/when I find something.

AaronDonahue:
@jgfouca , I think I found the source of the issue with negative source/sinks.  One of the test cases in p3_tests was not setting `dt` in the FortranDataIterator, which was defaulting to `dt=-1`.  The fix looks like:

```
diff --git a/components/scream/src/physics/p3/p3_f90.cpp b/components/scream/src/physics/p3/p3_f90.cpp
index 7cbeb8541b..8a1277c846 100644
--- a/components/scream/src/physics/p3/p3_f90.cpp
+++ b/components/scream/src/physics/p3/p3_f90.cpp
@@ -167,6 +167,7 @@ int test_p3_init () {
 
 int test_p3_ic (bool use_fortran) {
   const auto d = ic::Factory::create(ic::Factory::mixed);
+  d->dt = 300.0;
   p3_init();
   p3_main(*d, use_fortran);
   P3GlobalForFortran::deinit();
```

AaronDonahue:
I have also found another location where we can get negative sink values, see https://github.com/E3SM-Project/scream/blob/master/components/scream/src/physics/p3/p3_rain_self_collection_impl.hpp#L35

because we use 2 - exp(X) it is possible for X to be large enough to cause this to go negative.  This bug is also present in the F90 code, as expected.  I think this is something worth investigating.

AaronDonahue:
Depdending on the results of Issue #1094 , this is a possible fix that removes the remaining sink/source<0 issues, according to my tests:

```
diff --git a/components/scream/src/physics/p3/p3_rain_self_collection_impl.hpp b/components/scream/src/physics/p3/p3_rain_self_collection_impl.hpp
index 361c87a1b3..3520e8db32 100644
--- a/components/scream/src/physics/p3/p3_rain_self_collection_impl.hpp
+++ b/components/scream/src/physics/p3/p3_rain_self_collection_impl.hpp
@@ -34,8 +34,16 @@ void Functions<S,D>
     if (dum2_gt_dum1.any()) {
       dum.set(dum2_gt_dum1, 2 - exp(2300 * (dum2-dum1)));
     }
+    const auto dum_neg = dum < 0 && qr_incld_not_small;
+    if (dum_neg.any()) {
+      dum.set(dum_neg,0.0);
+    }
```

and

```
diff --git a/components/eam/src/physics/cam/micro_p3.F90 b/components/eam/src/physics/cam/micro_p3.F90
index 2468b643d4..45ddb02360 100644
--- a/components/eam/src/physics/cam/micro_p3.F90
+++ b/components/eam/src/physics/cam/micro_p3.F90
@@ -2719,6 +2719,8 @@ subroutine rain_self_collection(rho,qr_incld,nr_incld,    &
       else if (dum2.ge.dum1) then
          dum = 2._rtype - bfb_exp(2300._rtype*(dum2-dum1))
       endif
+      ! Clip dum so that it can't be <0
+      dum = max(dum,0.0);
```

jgfouca:
@AaronDonahue , your fix worked great! I've removed the extra checks I added to ni|nr_conservation. I also added a check to p3|shoc_main to ensure that parameters that are not automatically set by the p3|shoc:::FortranData Factories are set to something reasonable by the time p3|shoc_main is called.

jgfouca:
@bartgol , my latest push takes care of the homme FPE. Please review.

components/scream/src/physics/rrtmgp/tests/rrtmgp_tests.cpp
brhillman: This is a little worrisome. We should be able to keep these lines and just update the baselines, but let me check with the RRTMGP offline tests and make sure the changes still give the right answers to machine precision before we do that.

bartgol (approved):

brhillman (commented):
I think the RRTMGP fixes look good now. I put the changes in a PR upstream to our RRTMGP repo, assuming everything checks out I think we should switch over to main once I get that PR integrated.

components/scream/src/physics/p3/tests/p3_rain_sed_unit_tests.cpp
jgfouca: @bartgol , yes, I'm sure that was the intention here. The "real" calls to compute_rain_fall_velocity occur within a block of code in rain_sedimentation that's already protected by qr_small checks. I did not want to add a check for the sake of unit testing that would be redundant and potentially reduce performance in the context of the full system.

components/scream/src/physics/shoc/shoc_assumed_pdf_impl.hpp
jgfouca: Good question. I think if `w1_2 - w1_1 > w_thresh` then `tmp_val_1 - tmp_val_2` should never be zero.

components/eam/src/physics/cam/shoc.F90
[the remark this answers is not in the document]
PeterCaldwell: Hmm. One question is why std_s would ever approach 0. s is the dry static energy, so is constructed from a sum of temperature and vapor variances... I'm not sure why these would ever be zero. So: are we sure we're getting overflow in realistic cases, or just with random data for unit tests? 

Beyond the question of whether the inputs causing the problem are reasonable, I think your hack is fine as a temporary solution. A better solution might be to notice that the problem calculation always appears as R=s/std_s. The right thing to do would be to use L'Hospital's rule to decide whether s or std_s dominates in this case. I think s should be physically bounded by the temperature and water vapor values seen on planet Earth, while std_s could get arbitrarily large (and is a standard deviation so is always positive), so my bet on L'Hospital is that R will go to infinity as std_s goes to zero. 

Note as well that R is only used in erf(R) and exp(-R) types of expressions, which quickly converge to 1 or 0 as R gets big. This means that we can replace the overflowing R with a big number without impacting the solution. So maybe replace `if (std_s .gt. 2.0E-155_rtype)` with `std_s = min(std_s, tiny(rtype) )`? I think this would give you the C=1 and qn=s  that you're prescribing in the `else` for the `std_s > 2e-155` conditional. So you get rid of the overflow *and* the conditional!

components/scream/src/physics/shoc/shoc_calc_shoc_varorcovar_impl.hpp
```diff
@@ -33,11 +33,14 @@ ::calc_shoc_varorcovar(
...
-
-    // Compute the variance or covariance
-    varorcovar(k).set(range_pack1 > 0 && range_pack1 < nlev, tunefac*(isotropy_zi(k)*tkh_zi(k))*grid_dz2*(invar1_sm1 - invar1_s)*(invar2_sm1 - invar2_s));
+    const auto dz_zi_ne_zero = dz_zi(k) != 0;
+    if (dz_zi_ne_zero.any()) {
+      const Spack grid_dz = 1 / dz_zi(k);
+      const Spack grid_dz2 = ekat::square(grid_dz); // vertical grid diff squared
+
+      // Compute the variance or covariance
+      varorcovar(k).set(range_pack1 > 0 && range_pack1 < nlev && dz_zi_ne_zero, tunefac*(isotropy_zi(k)*tkh_zi(k))*grid_dz2*(invar1_sm1 - invar1_s)*(invar2_sm1 - invar2_s));
```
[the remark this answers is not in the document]
jgfouca: @ambrad , you're right. The issue here was not so much that dz_zi was zero, it is that we were not protecting these blocks of code with `any()` checks on the range masks.

components/eam/src/physics/cam/shoc.F90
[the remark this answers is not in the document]
jgfouca: @PeterCaldwell , note that `tiny(rtype)` would be somewhere in the range of 1e-308 for doubles, so much much smaller than 2e-155. Since it is the square of `s/std_s` that is overflowing, we'd need to check against something closer to `sqrt(tiny(rtype))`, depending on size of s.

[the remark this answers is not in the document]
PeterCaldwell: Ok, I suggested using tiny(rtype) rather than 2e-155 because hardcoded numbers look ugly rather than for any good reason. The bigger point I was trying to make is that it shouldn't matter whether you use 2e-155 or 1e-30 or tiny(rtype) because all of them will get mapped to approximately identical values of erf(R) or exp(-R). So feel free to replace my ill-advised tiny statements with 2e-155 or whatever value you like.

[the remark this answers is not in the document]
bartgol: On a different note, shoc.F90 is not going to be supported in scream for much longer, so maybe whatever hack works is fine?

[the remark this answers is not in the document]
bartgol: But given that you need cxx-f90 bfb-ness, and we want to avoid errors in cxx even when the f90 goes away (if ever), then I encourage a formulation based on a not-hard-coded number. Jim, I think something like `bfb_sqrt(tiny(rtype))` as you suggested is fine.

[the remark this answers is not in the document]
jgfouca: @PeterCaldwell , I was able to come up with a less-hacky min value, `bfb_sqrt(tiny(1._rtype)) * 100`. Trying to get rid of the conditional entirely led to some subtle non-bfbness between f90 and cxx that I'm not sure I want to track down. The cxx version of shoc_assumed_pdf stuff has diverged significantly from fortran.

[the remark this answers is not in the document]
PeterCaldwell: ok, that's fine with me.

jgfouca:
It looks like new FPEs have made it onto master (undetected) which is caused AT to fail my PR (it does an upstream merge of master, bringing in the new FPE problems which get now get detected on my PR). Consulting with @brhillman since the FPE is in the new test `shoc_cld_p3_rrtmgp_ut_np1_omp1`.

brhillman:
The current issue (after merging in master) with the shoc_p3_cld_rrtmgp test is that mu0/cos_zenith has columns that are zero-valued. This is totally reasonable, mu0 <= 0 just means the sun is at or below the horizon, but the solar radiative transfer code doesn’t know what to do with this. Currently, RRTMGP assumes only daytime columns are passed to the code so that a bunch of conditionals are avoided within the kernels. In order to deal with this, and to save some computations, the strategy in EAM has always been to subset the inputs to daytime-only columns before passing to the solar radiative transfer codes, and then to expand the fluxes back to all columns afterwards. The alternative is to add conditionals to simply pass over the columns with mu0 <= 0, which would avoid data copies and make the higher level code cleaner but would mean more computations relative to the subsetting approach, and might mean fairly intrusive modifications to the RRTMGP code (not a huge deal). Regardless, I think both of these strategies are beyond of the scope of this PR, so we should disable this test from the autotester, merge the PR, and just let the test register a fail on the dashboard until we fix it (which should be short order, but I don’t think we want to make these side effect changes in Jim’s PR, nor hold up the PR any more).

jgfouca:
Thx, @brhillman

jgfouca:
At long last this is ready to merge.

PR #1099, opened 2021-06-30
SCREAM: fix usage of CudaWrap structs

### LINKED ISSUE #1013: CudaWrap's second template argument is pointless
The [CudaWrap struct](https://github.com/E3SM-Project/scream/blob/fdb39f830ad687a11cbad1027edcff4572e082bb/components/scream/src/physics/share/physics_share.cpp#L20) has a second template argument (DeviceT) that is completely pointless, since it's never used.

I believe this template arg should be used to construct the RangePolicy for the parallel_reduce call, or else Kokkos will automatically use the default execution space. Usually, that would be ok, but if we are trying to use the CudaWrap struct with a different device (e.g., with `HostDevice`), it would not work.

We should either remove the `DeviceT` template arg (and assume we're always ok with Kokkos' default exec space), or use it in the parallel_reduce, like so:

```
Kokkos::parallel_reduce(Kokkos::RangePolicy<typename DeviceT::execution_space>(1), KOKKOS_LAMBDA ...
```

bartgol:
Explicitly build the dummy range policy, to use the template arg DeviceT (avoiding Kokkos picking the default device).

Fixes #1013.

Note: the code as is works fine but is a bit confusing the fact that DeviceT is never used.


jgfouca (approved):
Looks fine. Wondering why this code is duplicated in both p3 and shoc.

bartgol:
Good question. All the fcns are guarded by namespaces, so it's fine for cxx linking. So long as they are actually the same code, even f90 linking would work (the linker should simply pick the 1st symbol it finds).

That said, we should probably remove the p3 version (since p3 should link against physics/share). I'm gonna try to do that here, and see if it still passes the tests.

jgfouca (approved):
Thx!

PR #1100, opened 2021-07-01
Changes equal to approx equal for wet<-->dry mmr test

singhbalwinder:
Strict equal comparison was failing on some nightly test machines. This PR changes `equal` to `approx_equal` to see if the tests pass. A new test has been added where strict `equal` test is used with qv=2^k+1 (where k is a random number)

### DIFF
--- a/components/scream/src/share/tests/common_physics_functions_tests.cpp
+++ b/components/scream/src/share/tests/common_physics_functions_tests.cpp
@@ -98,6 +98,7 @@ void run(std::mt19937_64& engine)
   view_1d temperature("temperature",num_mid_packs),
           height("height",num_mid_packs),
           qv("qv",num_mid_packs),
+          rand_int("rand_int",num_mid_packs),
           pressure("pressure",num_mid_packs),
           pseudo_density("pseudo_density",num_mid_packs),
           dz_for_testing("dz_for_testing",num_mid_packs),
@@ -136,9 +137,14 @@ void run(std::mt19937_64& engine)
        pdf_surface(100.0,400.0),
        pdf_mmr(0,0.99);
 
+  //contruct random integers
+  using IPDF = std::uniform_int_distribution<int>;
+  IPDF pdf_rand_int(1,100);
+
   ekat::genRandArray(dview_as_real(temperature),     engine,pdf_temp);
   ekat::genRandArray(dview_as_real(height),          engine,pdf_height);
   ekat::genRandArray(dview_as_real(qv),              engine,pdf_qv);
+  ekat::genRandArray(dview_as_real(rand_int),        engine,pdf_rand_int);
   ekat::genRandArray(dview_as_real(pressure),        engine,pdf_pres);
   ekat::genRandArray(dview_as_real(pseudo_density),  engine,pdf_dp);
   ekat::genRandArray(dview_as_real(mmr_for_testing), engine,pdf_mmr);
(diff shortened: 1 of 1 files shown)


components/scream/src/share/tests/common_physics_functions_tests.cpp
```diff
@@ -238,15 +238,28 @@ void run(std::mt19937_64& engine)
...
   // mmr_test2: For zero wetmmr, drymmr should be zero
-  // mmr_test3: Compute drymmr from wetmmr0 and then use the result to compute wetmmr, which should be equal to wetmmr0
+  // mmr_test3: Compute drymmr from wetmmr0 and then use the result to compute wetmmr, which should be approximately
+  //            equal to wetmmr0
+
+  // mmr_test4: [to test mathematical properties of the function] Compute drymmr from wetmmr0 and then use the result to
+  //            compute wetmmr. "qv" should be in the form of 2^k+1 (1 is added as we have [1-qv] in the numerator or
+  //            denominator of this function), so that the results of wet->dry->wet are exactly
+  //            the same as the initial input mmr(wetmmr0). This is a mathematical property test, so physically invalid qv
+  //            values are also acceptable
```
bartgol, line 248: Did you mean "unphysical values"?

bartgol: Also, I would add a `TODO` comment, stating that if we ever add inputs checks in the conversion functions, this test might have to be removed. Just so that devs don't freak out if/when they add the check and start seeing tests failing... :)

components/scream/src/share/tests/common_physics_functions_tests.cpp
```diff
@@ -238,15 +238,28 @@ void run(std::mt19937_64& engine)
...
   tmp = PF::calculate_drymmr_from_wetmmr(wetmmr0,qv0);//get drymmr from wetmmr0
   tmp = PF::calculate_wetmmr_from_drymmr(tmp, qv0);//convert it back to wetmmr0
-  REQUIRE( Check::equal(tmp,wetmmr0) );// wetmmr0 should be equal to tmp
+  REQUIRE( Check::approx_equal(tmp,wetmmr0,test_tol) );// wetmmr0 should be equal to tmp
+
+  //mmr_test4
+  ScalarT qv0_2k_m1 = pow(2,qv0)+1; // 2^qv0+1 (qv0 is just used as a random number here)
+  tmp = PF::calculate_drymmr_from_wetmmr(wetmmr0,qv0_2k_m1);//get drymmr from wetmmr0
+  tmp = PF::calculate_wetmmr_from_drymmr(tmp, qv0_2k_m1);//convert it back to wetmmr0
+  REQUIRE( Check::equal(tmp,wetmmr0) );// wetmmr0 should be exactly equal to tmp
```
bartgol, line 262: The fact that this is working (according to the AT) is puzzling. `qv0` is not an integer, so `2^qv0` is not an integer power of two, which means multiplying/dividing by it is not done _exactly_ in fp arithmetic.

bartgol (changes requested):
Ok, I ran 1M tests with wetmmr and qv0 drawn from their respective pdfs, and it failed 345 times, with error equal to machine precision (that's 0.0345% of the times). Then I tried again, but with qv drawn from a uniform int distribution in [-10,10], and it passed all 1M times.

I think we need to address this, to avoid random failures in our nightlies, even though we fail very few times.

singhbalwinder:
Thanks @bartgol for you careful review and running these tests. I have modified the code to use integers instead of reals.

components/scream/src/share/tests/common_physics_functions_tests.cpp
```diff
@@ -98,6 +98,7 @@ void run(std::mt19937_64& engine)
   view_1d temperature("temperature",num_mid_packs),
           height("height",num_mid_packs),
           qv("qv",num_mid_packs),
+          rand_int("rand_int",num_mid_packs),
```
bartgol, line 101: This view seems unused.

components/scream/src/share/tests/common_physics_functions_tests.cpp
singhbalwinder: I have added a *Warning* to alert developers that it might fail if a check is added to the function. It is a little unsatisfactory as someone has to go into the code and read this comment but I am not sure of there is a better way to display this info.

components/scream/src/share/tests/common_physics_functions_tests.cpp
singhbalwinder: You are right (thanks!), I removed it in a new commit. It looks like, when packs are used, all numbers (real or integers) in a pack are same. Is that a known issue?

bartgol (approved):

components/scream/src/share/tests/common_physics_functions_tests.cpp
bartgol: Yeah, I don't see any better way. The string `TODO` (all uppercase) is sometimes highlighted by some text editors, but that's about it, and it's not much more helpful than a normal warning. I think this is fine.

AaronDonahue (approved):

PR #1106, opened 2021-07-08
Enable core dumps on blake

jgfouca:


jgfouca:
This change will only be temporary while we are debugging the intermittent Blake FPE for shoc_tests_ut_np1_omp8

jgfouca:
@tcclevenger , @bogensch , the core file allowed me to find the line that Blake is dying on with an FPE:

```
#0  0x0000000000a56ab4 in ekat::operator>=<double, 1, double> ...
  at .../scream/externals/ekat/src/ekat/ekat_pack.hpp:484
#1  ... ::shoc_init(...)
    at .../scream/components/scream/src/physics/shoc/shoc_main_impl.hpp:42
   ...
```

Line `shoc_main_impl.hpp:42` has two `>=` operators in it, the first `range >= ntop_shoc` I can see is fine. The second `pref_mid(k) >= pblmaxp` , I can see that pblmaxp is fine which leaves `pref_mid(k)` as the problem. Gdb on a core dump won't let me probe this value further.

I can see this value getting initialized in `shoc_main_tests.cpp` via this block of code:
```
      // pref_mid must be monotonically increasing                                                                                                                                                                                                
      upper = pref_mid_range.first;
      lower = pref_mid_range.second;
      for (Int k=0; k<d.nlev; ++k) {
        d.pref_mid[k] = upper - k*(upper-lower)/(d.nlev-1);
      }
```

I don't see anything wrong there, and there's no FPE coming from this code, which leads me to wonder what could the contents of pref_mid(k) possibly be to cause on FPE on a `>=` operation?

jgfouca:
Some variable values in the broken stackframe:
```
k=72
range=72
ntop_shoc=0
nbot_shoc=72
```
So maybe we were running off the end of pref_mid?

jgfouca:
Just confirmed that pref_mid is size 72, so `pref_mid(k)` with k=72 is indeed running off the end of this array. Wondering why valgrind or Kokkos bounds checking did not catch this.

jgfouca:
Tests passing, so hopefully this fixes the last FPE.

jgfouca:
I will re-disable core dumps unless this FPE comes back.

bogensch (approved):

PR #1112, opened 2021-07-12
A number of changes that will be needed for SPA and RRTMGP:

AaronDonahue:
1. Make it possible to initialize the latitude and longitude in the
reference grid outside of dynamics.  In particular, make it possible
to gather that information from an input file.
   - SPA will need the lat/lon of the physics grid columns in order
     conduct the horizontal interpolation, at least in the case when
     weights will be calculated online.
   - RRTMGP will need the lat/lon of the physics grid columns when
     calculating the solar zenith angle.

2. Fix a bug in the time manager that only popped up when the time
went past one month.

3. Introduce the ability to get the Julian day from the time manager.
   - this will be needed for rrtmgp when calculating the solar zenith
     angle.
   - this will be needed by SPA when conducting the temporal interpolation
     on data.

4. Cleaned up the stand-alone tests to work with the new lat/lon infrastructure code.

5. Delete the build for p3+shoc test, it wasn't really doing anything and can be considered
   replaced by the full physics test shoc_cld_p3_rrtmgp.


tcclevenger (commented):
Seems to have a SP fail during building. Otherwise, looks good. I'll approve once the bug is fixed.

tcclevenger (approved):

AaronDonahue:
Flagging the conversation in #1050 which references this usage of `double` in the definition of a grid, that will be changed in this PR.

components/scream/src/control/atmosphere_driver.cpp
```diff
@@ -291,7 +291,19 @@ initialize_fields (const util::TimeStamp& t0)
...
   }
 
-  if (ifield>0) {
+  // Check the status of lat and lon values on reference grid:
+  // This option allows the user to set lat or lon in their own
+  // test or run setup code rather than by file.
+  bool skip_init_lat = true;
+  bool skip_init_lon = true;
+  if (ic_pl.isParameter("skip_init_lat")) {
+    bool skip_init_lat     = ic_pl.get<bool>("skip_init_lat");
```
bartgol, line 300: @AaronDonahue This is likely not what you want. You are not changing the value of the previous variable, but you are creating one with the same name, so `skip_init_lat` will always be true.

Your compiler should have issued a warning here, something like `Declaration shadows a local variable`.

Issue #1108, opened 2021-07-09
Mismatch between dynamics library and the HOMMEXX setup for SCREAM stand-alone runs


AaronDonahue:
The cmake macro `CreateDynamicsLib` creates a unique homme library, and is defined such that if a library has already been defined that matchs the "HOMME_TARGET", "NP", "PLEV" and "QSIZE_D" that the link points to that library rather than create another new one.

There appears to be a problem though, when setting up a new scream stand-alone test that has the same homme dynamics setup.  

For example, the `homme_stand_alone` test has the following in CMakeLists.txt:
```
# Get or create the dynamics lib
#                 HOMME_TARGET   NP PLEV QSIZE_D
CreateDynamicsLib("theta-l_kokkos"  4   72   35)
```

If I create a separate test, call it `homme_physics_stand_alone` and add the same line to my CMakeLists.txt file the macro detects that the `scream_theta-l_kokkos_4_72_35` library already exists and set the link to the previously established link, ala
```
  if ("${hommeLibName}" IN_LIST DynamicsLibsCreated)
    # This dynamics lib was built already somewhere in the project. Nothing to do
    set (dynLibName scream_${hommeLibName})
  else ()
  ```

Unfortunately, during the build of the new test we encounter the following compile-time error:
```
In file included from components/scream/../homme/src/share/cxx/ExecSpaceDefs.hpp(15),
                 from components/scream/../homme/src/share/cxx/Types.hpp(14),
                 from components/scream/tests/coupled/homme_physics/homme_physics_stand_alone.cpp(14):
components/scream/../homme/src/share/cxx/Dimensions.hpp(46): error: identifier "PLEV" is undefined
    static constexpr const int NUM_PHYSICAL_LEV = PLEV;
```

**Comment by AaronDonahue** [comment]
Following up on this issue:
I investigated the directories for both the `homme_stand_alone` and `homme_physics_stand_alone` tests and noticed that there are a few files missing.  I'm guessing that `Dimensions.hpp` is gathering data (including PLEV) from files it expects to find in the test directory and isn't finding them so it fails.

```
[donahue5@quartz380:homme_physics]$ ls
CMakeFiles  CTestTestfile.cmake  Makefile  cmake_install.cmake	homme_physics_stand_alone_modules  homme_shoc_cld_p3_rad_init_ne4np4.nc  input.yaml  namelist.nl  vcoord
```

vs.

```
[donahue5@quartz380:homme]$ ls
[...]

PR #1115, opened 2021-07-13
Introduce a Dynamics + Physics stand alone test

AaronDonahue:
This PR introduces the `homme_physics_stand_alone` test which builds on top of the `shoc_cld_p3_rrtmgp` test to include dynamics as well.

There are a few bugfixes that were discovered over the course of getting this test to work.  

- In particular a fix to the AD logic related to group requests - original bugfix credited to @bartgol 
- A bugfix to the shoc interface in how the tracers group was registered.


AaronDonahue:
Currently a WIP PR because it hasn't been fully vetted, and there is a hacky band-aid put on the homme_stand_alone test that will need to be addressed once #1108 is sorted out.

AaronDonahue:
@jeff-cohere I've requested you review mostly for the changes I made to the field property check code in this PR.  Of course, I absolutely welcome your comments on the other changes as well.

@tcclevenger This PR would bring in that commit that you gave to me which fixes the issue with `qv` being the first tracer in a group.  So if it goes in, it may impact the work you are doing on the surface coupling.  just a heads up.

tcclevenger (approved):
Looks good!

components/scream/src/share/io/scorpio_output.cpp
```diff
@@ -26,7 +26,7 @@ void AtmosphereOutput::init()
   m_is_restart = m_params.get<bool>("RESTART FILE",false);  // optional, default to false
 
   // Gather data from grid manager:  In particular the global ids for columns assigned to this MPI rank
-  EKAT_REQUIRE_MSG(m_grid_name=="Physics","Error with output grid! scorpio_output.hpp class only supports output on a Physics grid for now.\n");
+  EKAT_REQUIRE_MSG(m_grid_name=="Physics" || m_grid_name=="Physics GLL","Error with output grid! scorpio_output.hpp class only supports output on a Physics grid for now.\n");
```
jeff-cohere, line 29: Might it be helpful to explicitly list the valid "Physics grids" in this error message?

jeff-cohere (approved):
Looks good to me! The `name` on the `FieldPropertyCheck` interface probably helped quite a bit with debugging.

components/scream/src/share/io/scorpio_output.cpp
AaronDonahue: Good suggestion, I've made the change.

AaronDonahue:
I'm going to go ahead and merge this.  It had passed review before I fixed a comment per reviewer suggestion.  It's pending review again just because I pushed a new version, but that change was minimal.

components/scream/src/physics/share/physics_constants.hpp
```diff
@@ -139,6 +139,7 @@ Scalar Constants<Scalar>::get_gas_mol_weight(ci_string gas_name) {
   } else if (gas_name == "n2" ) {
     return 28.0134;
   }
+  return std::numeric_limits<Scalar>::quiet_NaN();
```
bartgol, line 142: Would be slightly better to use `ekat::ScalarTraits<Scalar>::invalid()`.

components/scream/src/share/field/field_property_checks/field_nan_check.hpp
```diff
@@ -0,0 +1,53 @@
...
+      }
+    }, num_nans);
+    return (num_nans == 0);
+  }
+
+  bool can_repair() const override {
+    return false;
+  }
+
+  void repair(Field<non_const_RT>& field) const override {
```
bartgol, line 43: This will yield a compiler warning, since `field` is unused. Can be suppressed by doing
```
void repair(Field<non_const_RT>& /* field */) const override {
```
or by removing the var name altogether.

components/scream/src/share/tests/field_tests.cpp
```diff
@@ -684,6 +685,34 @@ TEST_CASE("field_property_check", "") {
...
+    Kokkos::deep_copy(f1_view, host_view);
+    for (auto iter = f1.property_check_begin(); iter != f1.property_check_end(); iter++) {
+      REQUIRE(iter->check(f1));
+    }
+
+    // Assign a NaN value to the field, make sure it fails the check,
+    Int midpt = host_view.extent(0)/2;
+    host_view(midpt) = std::numeric_limits<Real>::quiet_NaN();
+    Kokkos::deep_copy(f1_view, host_view);
+    for (auto iter = f1.property_check_begin(); iter != f1.property_check_end(); iter++) {
```
bartgol, line 711: Minor: the loop here is pointless, since you have only one check. You could just do

```
const auto& nan_check = *f1.property_check_begin();
REQUIRE (nan_check->check(f1));
```

components/scream/tests/coupled/homme_physics/CMakeLists.txt
```diff
@@ -0,0 +1,99 @@
...
+# Get or create the dynamics lib
+#                 HOMME_TARGET   NP PLEV QSIZE_D
+CreateDynamicsLib("theta-l_kokkos"  4   72   35)
+
+set (NEED_LIBS cld_fraction shoc p3 scream_rrtmgp rrtmgp ${NETCDF_C} ${dynLibName} scream_control scream_share physics_share yakl)
+
+# Add source files
+set (SRC shoc_cld_p3_rrtmgp.cpp ${SCREAM_BASE_DIR}/src/physics/rrtmgp/tests/rrtmgp_test_utils.cpp)
+
+# Test Homme standalone
```
bartgol, line 20: This is not Homme standalone. Copy+paste?

components/scream/tests/coupled/homme_physics/CMakeLists.txt
```diff
@@ -0,0 +1,99 @@
...
+if (NOT SCREAM_AUTOTESTER AND NOT EKAT_ENABLE_VALGRIND)
+  CreateUnitTest(homme_physics_stand_alone "homme_physics_stand_alone.cpp" "${NEED_LIBS}" EXE_ARGS "< namelist.nl" LABELS "dynamics;driver;shoc;cld;p3;rrtmgp;physics")
```
bartgol, line 22: This might be just me, but `standalone` seems inappropriate if the test has 2+ processes. You probably intended it as "not through cime", but all the cmake tests we have are "not through cime", so we'd have to add it to all the tests...

Some possible alternative names: `homme_scream_physics` or `homme_p3_shoc_cld_rrtmgp`.

components/scream/tests/coupled/homme_physics/CMakeLists.txt
```diff
@@ -0,0 +1,99 @@
...
+  set_target_properties(homme_physics_stand_alone PROPERTIES COMPILE_FLAGS "${YAKL_CXX_FLAGS}")
+  target_include_directories(homme_physics_stand_alone PUBLIC
+    ${SCREAM_BASE_DIR}/src/physics/rrtmgp
```
bartgol, line 25: Doesn't the `scream_rrtmgp` target already add these folders to its include dirs? If so, these are unnecessary, since liinking(*) against `scream_rrtmgp` already takes care of propagating the include folders.

(*) Recall that, when what is linked is actually a cmake target generated in the current project (or imported),  `target_link_libraries` does more than just linking, namely propagates compiler flags/definitions, include directories, link directories, dependencies,...

components/scream/tests/coupled/p3_shoc/CMakeLists.txt
```diff
@@ -1,20 +0,0 @@
-INCLUDE (ScreamUtils)
```
bartgol, line 1: I'm curious: why was this test removed? Was it hard to maintain maybe?

AaronDonahue: @bartgol , I removed it because it was essentially broken and unused up till now.  It would have taken a bit of effort to get it working again.  I'm happy to bring it back if we want, but in the meantime I deleted it since it wasn't actually being run and was thus sort of a zombie test that could confuse future developers.  Since shoc+cld+p3+rrtmgp already tests all the physics coupled I felt comfortable moving forward without this test in place.

We may want permutations of the 4 physics tests in the future.  That would make it easier to identify the cause if the full physics breaks for some reason.

components/scream/tests/coupled/homme_physics/CMakeLists.txt
AaronDonahue: TBH, I'm not sure.  I recall adding all of these include directories after the test wouldn't work.  When I created the CMake for this test I started as lean as possible and added pieces as the build or run would fail.

components/scream/tests/coupled/homme_physics/CMakeLists.txt
AaronDonahue: Yes, that's a good point.  I'm down for removing "stand_alone" from all of the "coupled" tests and just making the name fit all of the procs in it.  So in this case `homme_p3_shoc_cld_rrtmgp` would be the most descriptive.

components/scream/tests/coupled/homme_physics/CMakeLists.txt
AaronDonahue: ahhh yes, definitely copy/paste :-)

components/scream/tests/coupled/p3_shoc/CMakeLists.txt
bartgol: No, I'm fine without it. I don't see the need to add permutations of all the processes supported in scream.

components/scream/tests/coupled/homme_physics/CMakeLists.txt
bartgol: I just checked, and rrtmgp _does_ attach the include dirs to the target as PUBLIC, so there's no need to add them again here. I would remove completely the `target_include_directories` command.

AaronDonahue:
I'll note for future reference, since the original thread was "resolved", if `shoc+cld+p3+rrtmgp` does fail we have the easy flexibility in the `input.yaml` file to turn on only certain processes.  Which would be equivalent to the permutations I hinted at, and much cleaner.

components/scream/src/share/tests/field_tests.cpp
AaronDonahue: per this comment, if I clean this up here, should I also clean it up for the other property check tests?  See positivity_check, monotonicity_check and within_interval_check.

bartgol: You don't have to clean up anything, this can stay the way it is...

PR #1117, opened 2021-07-14
Update EKAT

jgfouca:
Bring in a new EKAT version, this will enable Kokkos bounds checking.

The bounds checking uncovered some problems in p3_standalone where some views were being sized (ncol, nk_pack) where they should have been sized (ncol, nk_pack_p1) where `nk_pack_p1 = num_packs(nlev + 1)`. This error had been hidden by a couple factors, first, we lost Kokkos bounds checking sometime in the last year when we updated the Kokkos submodule. Second, when pack_size>1, nk_pack_p1 and nk_pack might be the same. I only noticed this error when building for FPEs (pack_size=1).


AaronDonahue (approved):
I'm surprised we didn't catch the microphysics bugs earlier on.  Thanks for fixing them.

jgfouca:
@ambrad , added you as a reviewer due to the homme change. The most recent commit fixes both tests `homme_pd_remap_ut_np1_omp1` and `homme_stand_alone_ut_np1_omp1`. The policy in question was not using the EKAT Kokkos utils and had a somewhat hardcoded team max size, so I don't think that was an opportunity for a broader refactor, but feel free to push back on that.

components/scream/src/dynamics/homme/physics_dynamics_remapper.hpp
```diff
@@ -695,7 +695,11 @@ do_remap_fwd() const
 
   const auto concurrency = KT::ExeSpace::concurrency();
 #ifdef KOKKOS_ENABLE_CUDA
+#ifdef KOKKOS_ENABLE_DEBUG
```
ambrad, line 698: OK, but none of this special-case code should be here. That's not the subject of this PR, though.

I'll just note that this code is working around the fact that EKAT's get_default_team_policy is currently specific to the SCREAM physics. That should be moved to SCREAM's physics directory's utilities. EKAT itself should provide basic utitlies like max team size. Then SCREAM can use that to develop a few basic team policies for different parts of the computation.

ambrad: @tcclevenger it looks like you introduce the original code. At some point I recommend you review https://github.com/E3SM-Project/E3SM/blob/master/components/homme/src/share/cxx/ExecSpaceDefs.hpp and its associated cpp file, as well as the similar code in https://github.com/E3SM-Project/EKAT/blob/master/src/ekat/kokkos/ekat_kokkos_utils.hpp, to work out a coherent approach to making TeamPolicys. EKAT should focus on providing basic utils like max team size as a function of debug build flag. SCREAM should use these utils to develop a set of TeamPolicy constructors for each use case. App code like this should never have KOKKOS_ENABLE_CUDA and similar details.

ambrad (approved):
The primary issues are orthogonal to this PR, so I'll approve this PR since it fixes the current practical problems.

components/scream/src/dynamics/homme/physics_dynamics_remapper.hpp
tcclevenger: @ambrad Ok, I'll look into a better way for defining these things. And then implement in scream.

PR #1118, opened 2021-07-14
Fix #1108.

### LINKED ISSUE #1108: Mismatch between dynamics library and the HOMMEXX setup for SCREAM stand-alone runs

ambrad:
Also fix some std namespace usage and static constexpr usage.

Fixes #1108

### DIFF
--- a/components/scream/src/dynamics/homme/CMakeLists.txt
+++ b/components/scream/src/dynamics/homme/CMakeLists.txt
@@ -98,6 +98,8 @@ macro (CreateDynamicsLib HOMME_TARGET NP PLEV QSIZE)
     message (FATAL_ERROR "Error! So far, SCREAM only supports 'theta-l_kokkos' as dynamics target.")
   endif()
 
+  add_definitions(-DHAVE_CONFIG_H)
+
   # Create the lib name, and proceed building only if not already built
   set (hommeLibName "${HOMME_TARGET}_${NP}_${PLEV}_${QSIZE}")
   if ("${hommeLibName}" IN_LIST DynamicsLibsCreated)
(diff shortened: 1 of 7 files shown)


AaronDonahue (approved):
Thank you for the quick fix @ambrad.

PR #1126, opened 2021-07-22
Final touches to get things building and running on LC machines

AaronDonahue:
This commit should fix all the issues we are seeing on building/running SCREAM on Livermore Computing machines.


jgfouca (approved):

components/scream/CMakeLists.txt
```diff
@@ -8,7 +8,7 @@ else()
   set (SCREAM_BINARY_DIR ${CMAKE_CURRENT_BINARY_DIR})
 endif()
 
-if (${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.12.0")
+if (${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.16.0")
```
AaronDonahue, line 11: @jgfouca , could this line be causing the `mappy` failures?

jgfouca: @AaronDonahue , I bet it is. Let me add a commit to this PR.

jgfouca (approved):

AaronDonahue:
@jgfouca , I'm not sure why this is still failing the AT, but in all honesty, I think that one file I changed that could have impacted mappy and blake could just be reverted.  I changed it because I thought it was needed to enforce that the `cmake` version be new enough for Kokkos.  But CDash is demonstrating that these changes are not needed everywhere, since the the sandia machines are not running in to an issue with cmake.

jgfouca (approved):

PR #1128, opened 2021-07-26
Add import/export in SurfaceCouplingfor SCREAM variable

tcclevenger:
Updates the `SurfaceCoupling` import and export to include special scream cases.

### DIFF
--- a/components/scream/src/control/surface_coupling.cpp
+++ b/components/scream/src/control/surface_coupling.cpp
@@ -1,6 +1,7 @@
 #include "control/surface_coupling.hpp"
 
 #include "share/field/field_utils.hpp"
+#include "share/util/scream_common_physics_functions.hpp"
 
 namespace scream {
 namespace control {
@@ -12,7 +13,17 @@ SurfaceCoupling (const field_mgr_ptr& field_mgr)
 {
   auto grid = m_field_mgr->get_grid();
 
-  m_num_cols  = grid->get_num_local_dofs();
+  m_num_cols = grid->get_num_local_dofs();
+  m_num_levs = grid->get_num_vertical_levels();
+
+  using namespace ShortFieldTagsNames;
+  using namespace ekat::units;
+
+  FieldLayout layout { {COL}, {m_num_cols} };
+  Units nondim(0,0,0,0,0,0,0);
+  auto grid_name = grid->name();
+  FieldIdentifier id ("dummy_field", layout, nondim, grid_name);
+  dummy_field = Field<const double>(id);
 
   EKAT_REQUIRE_MSG(grid->type()==GridType::Point,
       "Error! Surface coupling only implemented for 'Point' grids.\n"
@@ -20,21 +31,32 @@ SurfaceCoupling (const field_mgr_ptr& field_mgr)
 }
 
 void SurfaceCoupling::
-set_num_fields (const int num_imports, const int num_exports)
+set_num_fields (const int num_total_imports, const int num_scream_imports, const int num_exports)
 {
   EKAT_REQUIRE_MSG(m_state==RepoState::Clean,
                      "Error! You can only set the number of fields in SurfaceCoupling once.\n");
 
-  m_scream_imports_dev = decltype(m_scream_imports_dev)("",num_imports);
+  m_scream_imports_dev = decltype(m_scream_imports_dev)("",num_scream_imports);
   m_scream_exports_dev = decltype(m_scream_exports_dev)("",num_exports);
 
   m_scream_imports_host = Kokkos::create_mirror_view(m_scream_imports_dev);
   m_scream_exports_host = Kokkos::create_mirror_view(m_scream_exports_dev);
 
+  // These fields contain computation needed for some export fields
+  qv_sfc    = decltype(qv_sfc)   ("", m_num_cols);
+  Sa_ptem   = decltype(Sa_ptem)  ("", m_num_cols);
+  Sa_u      = decltype(Sa_u)     ("", m_num_cols);
+  Sa_v      = decltype(Sa_v)     ("", m_num_cols);
+  Sa_dens   = decltype(Sa_dens)  ("", m_num_cols);
+  zero_view = decltype(zero_view)("", m_num_cols);
+
   // These will be incremented every time we register an import/export.
   // At registration end, they should match the length of the corresponding view above.
   m_num_imports = 0;
   m_num_exports = 0;
+
+  // We keep up with the total number of imports for indexing input data
+  m_num_total_imports = num_total_imports;
 
   m_state = RepoState::Open;
 }
@@ -50,38 +72,48 @@ register_import(const std::string& fname,
                       "       Did you forget to call set_num_fields(..) ?\n");
   EKAT_REQUIRE_MSG (m_state!=RepoState::Closed, "Error! Registration phase has already ended.\n");
 
-  // Check that we still have room
-  EKAT_REQUIRE_MSG(m_num_imports<m_scream_imports_host.extent_int(0),
-                     "Error! Imports view is already full. Did you call 'set_num_fields' with the wrong arguments?\n");
+  if (fname == "unused" || fname == "RRTMGP")
+  {
+    // Do nothing
+  } else {
+    // Check that we still have room
+    EKAT_REQUIRE_MSG(m_num_imports<m_scream_imports_host.extent_int(0),
+                       "Error! Imports view is already full. Did you call 'set_num_fields' with the wrong arguments?\n");
 
-  // Get the field, and check that is valid
-  import_field_type field = m_field_mgr->get_field(fname);
-  EKAT_REQUIRE_MSG (field.is_allocated(), "Error! Import field view has not been allocated yet.\n");
-  
-  EKAT_REQUIRE_MSG(cpl_idx>=0, "Error! Input cpl_idx is negative.\n");
+    // Get the field, and check that is valid
+    import_field_type field = m_field_mgr->get_field(fname);
 
-  // Check that this cpl_idx wasn't already registered
-  for (int i=0; i<m_num_imports; ++i) {
-    EKAT_REQUIRE_MSG(cpl_idx!=m_scream_imports_host(i).cpl_idx,
-                       "Error! An import with cpl_idx " + std::to_string(cpl_idx) + " was already registered.\n");
+    EKAT_REQUIRE_MSG (field.is_allocated(), "Error! Import field view has not been allocated yet.\n");
+
+    EKAT_REQUIRE_MSG(cpl_idx>=0, "Error! Input cpl_idx is negative.\n");
+
+    // Check that this cpl_idx wasn't already registered
+    for (int i=0; i<m_num_imports; ++i) {
+      EKAT_REQUIRE_MSG(cpl_idx!=m_scream_imports_host(i).cpl_idx,
+                         "Error! An import with cpl_idx " + std::to_string(cpl_idx) + " was already registered.\n");
+    }
+
+    auto& info = m_scream_imports_host(m_num_imports);
+
+    EKAT_ASSERT_MSG (field.get_header().get_parent().lock() == nullptr,
+                     "Error! " + fname + "has a parent. Currently SurfaceCoupling does not support "
+                     "importing or exporting fields which have parents since get_view().data() is unsafe.");
+
+    // Set view data ptr
+    info.data = field.get_view().data();
+
+    // Set cpl index
+    info.cpl_idx = cpl_idx;
+
+    // Get column offset and stride
+    get_col_info (field.get_header_ptr(), vecComp, info.col_offset, info.col_stride);
+
+    // Store the identifier of this field, for debug purposes
+    m_imports_fids.insert(field.get_header().get_identifier());
+
+    // Update number of imports stored
+    ++m_num_imports;
   }
-
-  auto& info = m_scream_imports_host(m_num_imports);
-
-  // Set view data ptr
-  info.data = field.get_view().data();
-
-  // Set cpl index
-  info.cpl_idx = cpl_idx;
-
-  // Get column offset and stride
-  get_col_info (field.get_header_ptr(), vecComp, info.col_offset, info.col_stride);
-
-  // Store the identifier of this field, for debug purposes
-  m_imports_fids.insert(field.get_header().get_identifier());
-
-  // Update number of imports stored
-  ++m_num_imports;
 }
 
 void SurfaceCoupling::
@@ -99,10 +131,6 @@ register_export (const std::string& fname,
   EKAT_REQUIRE_MSG(m_num_exports<m_scream_exports_host.extent_int(0),
                      "Error! Exports view is already full. Did you call 'set_num_fields' with the wrong arguments?\n");
 
-  // Get the field, and check that is valid
-  export_field_type field = m_field_mgr->get_field(fname);
-  EKAT_REQUIRE_MSG (field.is_allocated(), "Error! Export field view has not been allocated yet.\n");
-
   EKAT_REQUIRE_MSG(cpl_idx>=0, "Error! Input cpl_idx is negative.\n");
 
   // Check that this cpl_idx wasn't already registered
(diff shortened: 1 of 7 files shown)


tcclevenger (commented):
@bartgol I left a few comments and a question about the existing test. I'm marking as WIP since I already plan to make changes after your comments. Fell free to critique the overal structure and suggest any changes.

Note that `do_import` and `do_export` remain identical, everything is take care of in the registration. Still in the registration there is only ever a loop over columns so it should not be too expensive.

components/scream/src/control/surface_coupling.cpp
```diff
@@ -12,30 +13,51 @@ SurfaceCoupling (const field_mgr_ptr& field_mgr)
...
 
-  m_num_cols  = grid->get_num_local_dofs();
+  m_num_cols = grid->get_num_local_dofs();
+  m_num_levs = grid->get_num_vertical_levels();
+
+  using namespace ShortFieldTagsNames;
+  using namespace ekat::units;
+
+  FieldLayout layout { {COL}, {m_num_cols} };
+  Units nondim(0,0,0,0,0,0,0);
```
bartgol, line 23: Minor (and probably subjective): I think most people will be confused if this ctor is used. You could do something less puzzling by doing
```
auto nondim = Units::nondimensional();
```

components/scream/src/control/surface_coupling.cpp
```diff
@@ -12,30 +13,51 @@ SurfaceCoupling (const field_mgr_ptr& field_mgr)
...
+  dummy_field = Field<const double>(id);
 
   EKAT_REQUIRE_MSG(grid->type()==GridType::Point,
       "Error! Surface coupling only implemented for 'Point' grids.\n"
       "       Input grid type: " + e2str(grid->type()) + "\n");
 }
 
 void SurfaceCoupling::
-set_num_fields (const int num_imports, const int num_exports)
+set_num_fields (const int num_total_imports, const int num_scream_imports, const int num_exports)
```
bartgol, line 34: Another subjective one: 'total' is a very overloaded word; we may call the first one `num_cpl_imports`, which emphasizes it's the number of imports offered by the coupler.

components/scream/src/control/surface_coupling.cpp
```diff
@@ -12,30 +13,51 @@ SurfaceCoupling (const field_mgr_ptr& field_mgr)
...
   m_scream_imports_host = Kokkos::create_mirror_view(m_scream_imports_dev);
   m_scream_exports_host = Kokkos::create_mirror_view(m_scream_exports_dev);
 
+  // These fields contain computation needed for some export fields
+  qv_sfc    = decltype(qv_sfc)   ("", m_num_cols);
+  Sa_ptem   = decltype(Sa_ptem)  ("", m_num_cols);
+  Sa_u      = decltype(Sa_u)     ("", m_num_cols);
+  Sa_v      = decltype(Sa_v)     ("", m_num_cols);
+  Sa_dens   = decltype(Sa_dens)  ("", m_num_cols);
+  zero_view = decltype(zero_view)("", m_num_cols);
```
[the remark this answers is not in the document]
bartgol: Yes, this makes sense to me.

components/scream/src/control/surface_coupling.cpp
```diff
@@ -113,17 +141,124 @@ register_export (const std::string& fname,
...
 
-  // Get column offset and stride
-  get_col_info (field.get_header_ptr(), vecComp, info.col_offset, info.col_stride);
+    if (fname == "qv") {
 
-  // Store the identifier of this field, for debug purposes
-  m_exports_fids.insert(field.get_header().get_identifier());
+      const auto qv = m_field_mgr->get_field(fname).get_reshaped_view<const Real**>();
+      const auto policy = KokkosTypes<device_type>::RangePolicy(0, m_num_cols);
+      Kokkos::parallel_for(policy, KOKKOS_LAMBDA(const int& i) {
```
bartgol, line 185: Why do you copy data from the FM field into qv_sfc at registration time? I would expect this operation to be done every time that do_export is called...

components/scream/src/control/surface_coupling.cpp
```diff
@@ -113,17 +141,124 @@ register_export (const std::string& fname,
...
+      // Set view data ptr
+      info.data = zero_view.data();
+
+    } else if (fname == "Sa_ptem") {
+
+      const auto T_mid = m_field_mgr->get_field("T_mid").get_reshaped_view<const Real**>();
+      const auto p_mid = m_field_mgr->get_field("p_mid").get_reshaped_view<const Real**>();
+      const auto policy = KokkosTypes<device_type>::RangePolicy(0, m_num_cols);
+      Kokkos::parallel_for(policy, KOKKOS_LAMBDA(const int& i) {
+        Sa_ptem(i) = PF::calculate_theta_from_T(T_mid(i, last_entry), p_mid(i, last_entry));
```
bartgol, line 205: Same comment here (and for other fields below) as for qv.

components/scream/src/control/tests/sc_tests.cpp
```diff
@@ -145,12 +145,263 @@ TEST_CASE ("surface_coupling")
     v3d_imp.sync_to_host();
     for (int icol=0; icol<ncols; ++icol) {
       REQUIRE (s2d_exp_h(icol)==s2d_imp_h(icol));
-      REQUIRE (s3d_exp_h(icol,0)==s3d_imp_h(icol,0));
-      REQUIRE (v3d_exp_h(icol,0,0)==v3d_imp_h(icol,0,0));
+      REQUIRE (s3d_exp_h(icol,nlevs-1)==s3d_imp_h(icol,nlevs-1));
+      REQUIRE (v3d_exp_h(icol,0,0)==v3d_imp_h(icol,0,0)); // Is this correct?
       REQUIRE (v3d_exp_h(icol,0,1)==v3d_imp_h(icol,0,1));
```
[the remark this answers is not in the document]
bartgol: Yes. I think we discussed this when we last chatted. Throughout the SC code, I accidentally assumed that the surface is at ilev=0. Since I made the mistake both in the SC impl and in the unit test, things worked out. Once you fixed the SC impl, this test needs to be fixed as well.

components/scream/src/control/tests/sc_tests.cpp
```diff
@@ -145,12 +145,263 @@ TEST_CASE ("surface_coupling")
...
+    REQUIRE (export_raw_data[2  + icol*num_exports] == z_mid_h          (icol,    nlevs-1)); // 3rd export
+    REQUIRE (export_raw_data[3  + icol*num_exports] == horiz_winds_h    (icol, 0, nlevs-1)); // 4th export
+    REQUIRE (export_raw_data[4  + icol*num_exports] == horiz_winds_h    (icol, 1, nlevs-1)); // 5th export
+    REQUIRE (export_raw_data[5  + icol*num_exports] == p_mid_h          (icol,    nlevs-1)); // 6th export
+    REQUIRE (export_raw_data[7  + icol*num_exports] == qv_h             (icol,    nlevs-1)); // 8th export
+    REQUIRE (export_raw_data[9  + icol*num_exports] == precip_liq_surf_h(icol            )); // 10th export
+
+    // These exports should be set to 0
+    const auto eps = std::numeric_limits<Real>::epsilon();
+    REQUIRE (abs(export_raw_data[8  + icol*num_exports]) < eps); // 9th export
```
bartgol, line 402: Why not check that they are exactly 0, rather than a tolerance test? If I understand correctly, they should exactly be 0, no?

components/scream/src/mct_coupling/scream_cpl_indices.F90
```diff
@@ -7,6 +7,7 @@ module scream_cpl_indices
 
   ! Focus only on the ones that scream imports/exports (subsets of x2a and a2x)
   integer, parameter, public :: num_required_imports = 21
+  integer, parameter, public :: num_scream_imports   = 4
```
bartgol, line 10: We might add "_cpl_" to the other ones, like `num_cpl_required_imports`, to stress the distinction between mct and scream sides.

components/scream/src/mct_coupling/scream_cxx_f90_interface.cpp
```diff
@@ -146,20 +138,46 @@ void scream_setup_surface_coupling (
...
+
+void scream_init_atm (const int& start_ymd,
+                      const int& start_tod)
+{
+  using namespace scream;
+  using namespace scream::control;
+
+  fpe_guard_wrapper([&](){
+    // First of all, initialize the scream session
+    scream::initialize_scream_session();
```
bartgol, line 161: I think this should be now moved inside scream_create_atm_instance, right?

components/scream/src/control/surface_coupling.cpp
```diff
@@ -324,6 +459,11 @@ get_col_info(const std::shared_ptr<const FieldHeader>& fh,
     me = p;
     p = me->get_parent().lock();
   }
+
+  // Surface corresponds to k=nlevs-1
+  // Hacky, but need to ask Luca about vector3d
+  if (lt==LayoutType::Scalar3D)
+    col_offset+= m_num_levs-1;
```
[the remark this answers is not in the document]
bartgol: Yeah, we need vector3d support. Should be fairly easy to do though, right?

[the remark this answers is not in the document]
bartgol: I'm actually looking at how col_offset is computed, and I am not 100% sure it is correct. In particular, for vector fields, it's init-ed to the vector dimension, I believe because I was thinking that the surface was the 1st entry in a column. Would you mind double checking that logic?

components/scream/src/control/tests/sc_tests.cpp
```diff
@@ -145,12 +145,263 @@ TEST_CASE ("surface_coupling")
...
+    // Check exports
+
+    // These exports are direct values from a scream field
+    REQUIRE (export_raw_data[0  + icol*num_exports] == T_mid_h          (icol,    nlevs-1)); // 1st export
+    REQUIRE (export_raw_data[2  + icol*num_exports] == z_mid_h          (icol,    nlevs-1)); // 3rd export
+    REQUIRE (export_raw_data[3  + icol*num_exports] == horiz_winds_h    (icol, 0, nlevs-1)); // 4th export
+    REQUIRE (export_raw_data[4  + icol*num_exports] == horiz_winds_h    (icol, 1, nlevs-1)); // 5th export
+    REQUIRE (export_raw_data[5  + icol*num_exports] == p_mid_h          (icol,    nlevs-1)); // 6th export
+    REQUIRE (export_raw_data[7  + icol*num_exports] == qv_h             (icol,    nlevs-1)); // 8th export
+    REQUIRE (export_raw_data[9  + icol*num_exports] == precip_liq_surf_h(icol            )); // 10th export
```
bartgol, line 398: I was wondering if there's any way to check the fields that are computed from scream fields, such as `Sa_ptem`. I guess it would entail writing a lot of code, which would boil dow to re-do what is done inside the SC class, so maybe not much interesting?

bartgol (commented):
Nice work. I put some comments here and there. I think the biggest things are 1) figuring out the correct offset for vector3d fields (and, in general, check that offset/stride are computed correctly), and 2) make the do_export routine compute the "derived" fields.

components/scream/src/control/tests/sc_tests.cpp
tcclevenger: That is what I thought, but I kept getting a failure with `0.0 == 0.0` message. I tried casting both to `Real` and it still didn't fix the issue. Not sure what I am doing wrong.

components/scream/src/control/surface_coupling.cpp
tcclevenger: Ahh didn't realize there was a simpler way. I'll definitely change.

components/scream/src/control/surface_coupling.cpp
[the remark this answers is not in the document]
tcclevenger: Yeah, I'll go through it.

components/scream/src/mct_coupling/scream_cxx_f90_interface.cpp
tcclevenger: After this PR is finished, my plan was to have a separate one with all the stuff required for CIME to run. I can include those changes here if you'd rather keep them together.

tcclevenger: Maybe it makes sense to combine since this PR already includes interface changes..

components/scream/src/control/tests/sc_tests.cpp
bartgol: Uhn, interesting. Setting `a=0;` should make `a==0.0` yield true, since 0 is exactly representable in FP (single or double). Might be worth spending 10 minutes to see why it failed, to make sure we are indeed setting a=0...

components/scream/src/mct_coupling/scream_cxx_f90_interface.cpp
bartgol: Up to you. If the SC work can be self-contained (it seems you are adding scream-standalone testing for it), then we can break it up into 2 prs..

tcclevenger:
@bartgol I updated changes to `SurfaceCoupling` (working on the mct-coupling/interface still). I addressed the comments and changed the original test to include a vector2d example to test the new `col_offset` logic.

components/scream/src/control/tests/sc_tests.cpp
```diff
@@ -139,18 +154,276 @@ TEST_CASE ("surface_coupling")
...
+  coupler.register_export("Sa_dens",          6);
+  coupler.register_export("tracers",          7, 0);
+  coupler.register_export("set_zero",         8);
+  coupler.register_export("precip_liq_surf",  9);
+  coupler.register_export("set_zero",         10);
+  coupler.register_export("set_zero",         11);
+  coupler.register_export("set_zero",         12);
+
+  // Complete setup of importer/exporter, providing raw_data
+  double* import_raw_data = new Real[ncols*num_total_imports];
```
bartgol, line 352: It looks like you are not deallocating these. Not a huge deal, since the OS will do it, but still...

components/scream/src/control/tests/sc_tests.cpp
```diff
@@ -139,18 +154,276 @@ TEST_CASE ("surface_coupling")
...
+  // Create necessary fields for export. Tracers qc and qr are unnecessary, but
+  // are included to verify that subviewed fields (qv) are correctly handled
+  const auto nondim = Units::nondimensional();
+  FID T_mid_id           ("T_mid",           scalar3d_layout, K,      grid_name);
+  FID p_mid_id           ("p_mid",           scalar3d_layout, Pa,     grid_name);
+  FID z_mid_id           ("z_mid",           scalar3d_layout, m,      grid_name);
+  FID horiz_winds_id     ("horiz_winds",     vector3d_layout, m/s,    grid_name);
+  FID pseudo_density_id  ("pseudo_density",  scalar3d_layout, Pa,     grid_name);
+  FID qv_id              ("qv",              scalar3d_layout, nondim, grid_name);
+  FID qc_id              ("qc",              scalar3d_layout, nondim, grid_name);
```
bartgol, line 238: Is there a reason to register these? I don't see them used here.

components/scream/src/control/surface_coupling.cpp
```diff
@@ -222,6 +290,34 @@ void SurfaceCoupling::do_export ()
...
+    Kokkos::parallel_for(policy, KOKKOS_LAMBDA(const int& i) {
+      const auto dz = PF::calculate_dz(pseudo_density(i, last_entry), p_mid(i, last_entry),
+                                       T_mid(i, last_entry), qv(i, last_entry));
+
+      Sa_ptem(i) = PF::calculate_theta_from_T(T_mid(i, last_entry), p_mid(i, last_entry));
+      Sa_u(i)    = horiz_winds(i, 0, last_entry);
+      Sa_v(i)    = horiz_winds(i, 1, last_entry);
+      Sa_dens(i) = PF::calculate_density(pseudo_density(i, last_entry), dz);
+    });
+    Kokkos::deep_copy(zero_view, 0.0);
```
bartgol, line 319: This can probably be done right after the view is constructed.

components/scream/src/control/surface_coupling.cpp
```diff
@@ -268,43 +364,57 @@ get_col_info(const std::shared_ptr<const FieldHeader>& fh,
...
+  // Compute initial offset. We use get_last_extent() as it
+  // accounts for padding (if any).
   col_offset = vecComp;
-  if (vector) {
-    dims_prod = dims[1];
+  if (layout3d) {
+    if (lt==LayoutType::Vector3D) {
+      col_offset *= fh->get_alloc_properties().get_last_extent();;
+    }
+    col_offset += m_num_levs-1;
```
bartgol, line 384: I am not sure if we ever export/import from/into an interface field. Maybe @AaronDonahue can tell us. However, we could just use `fh->get_identifier().get_layout().dims().back()`. This assumes that the number of levels is the last dimension, but I think that's safe. Or, alternatively, be super safe and do:
```
const auto& layout = fh->get_identifier().get_layout();
const bool at_midpoints = layout.has_tag(VL);
const int num_levs = at_midpoints ? m_num_levs : m_num_levs+1;
col_offset += num_levs-1;
```

components/scream/src/control/tests/sc_tests.cpp
tcclevenger: I just included in case there was some issue with more than one field in a group. I guess this is being tested though in the first test. I'll remove.

components/scream/src/control/surface_coupling.cpp
```diff
@@ -268,43 +364,57 @@ get_col_info(const std::shared_ptr<const FieldHeader>& fh,
...
+  int dims_prod = 1;
+    if (vector) {
+      dims_prod = dims[1];
+    }
+
   std::shared_ptr<const FieldHeader> me = fh;
   std::shared_ptr<const FieldHeader> p  = me->get_parent().lock();
   while (p!=nullptr) {
+
+    // TODO: Currently this is not used as we require fields to not have parents in
```
bartgol, line 414: Do you think we should just remove this code? I mean, if we don't support subfields, we might as well remove code that we'll not run. It is not that complicated, so we'll probably be able to recreate it if need be...

tcclevenger: Alternatively it would be simple enough to remove the "no parent" requirement and give the SC-user the option of just handing the child field with no vector component (now that the offset is correct). I think it may be too trivial not for me to try real quick. If it becomes a hassle I'll just delete this section of code.

tcclevenger:
@bartgol Errors have been fixed. 

New stuff: mainly the col_stride/col_offset. In `get_col_info` now check for a single parent (no grandparents). I added a case to the test as well.

components/scream/src/control/surface_coupling.cpp
```diff
@@ -379,60 +369,49 @@ get_col_info(const std::shared_ptr<const FieldHeader>& fh,
...
-  std::shared_ptr<const FieldHeader> me = fh;
-  std::shared_ptr<const FieldHeader> p  = me->get_parent().lock();
-  while (p!=nullptr) {
+    const auto parent_lt = get_layout_type(parent->get_identifier().get_layout().tags());
 
-    // TODO: Currently this is not used as we require fields to not have parents in
-    //       register_import/export() functions. It is possible, but the logic here
-    //       would need to be tested. We will end up erroring out before making it here.
+    // If the field has a parent field, it should be the case that the
+    // parent field is a vectored field. Currently we conly expect
```
bartgol, line 398: It seems like the last part of the comment got cut. Also, "conly"

bartgol: Also, is "vectored" a word?

components/scream/src/control/surface_coupling.cpp
```diff
@@ -379,60 +369,49 @@ get_col_info(const std::shared_ptr<const FieldHeader>& fh,
...
+    EKAT_REQUIRE_MSG(parent_lt==LayoutType::Vector3D,
+                     "Error! SurfaceCoupling expects all subfields to have parents"
+                     " with LayoutType::Vector3D.\n");
 
-    const auto& idx = me->get_alloc_properties().get_subview_idx();
+    const auto& idx = fh->get_alloc_properties().get_subview_idx();
 
     // Recall: idx = (idim,k) = (dimension where slice happened, index along said dimension).
     // Field class only allows idim=0,1. But we should never be in the case of idim=0, here.
     // If we have idim=0, it means that the parent field did not have CL has tag[0]...
```
bartgol, line 407: While at it, would you mind fixing this comment? I think the last sentence was supposed to be "... did not have COL as tag[0]" (couple of typos).

bartgol (commented):
Looks good. One question though: now that you test exporting `qv` directly, we are not testing the registration of any vector field anymore, right? Also, since vector fields work fine, should we get rid of the locals `Sa_u` and `Sa_v`, and work directly with `horiz_winds`? We should be able to support that, right? It would be a good way to test vector fields (unless we already are, and I'm missing it).

components/scream/src/control/surface_coupling.cpp
tcclevenger: Hmm, I'll fix that. And yeah, now that I think about it, I don't think vector vectored is a word... I'll fix that too.

tcclevenger:
@bartgol We test vector fields of size `(ncols, 2, nlevs)` in the "surface_coupling" test. But I completely agree that doing `horiz_winds` with vector components is much better. I'll add it.

tcclevenger:
Trying to rebase to master? I've run test_all_scream on weaver and it's not showing this failure. Not sure what could have affected GPU performance in my last commit..

bartgol:
The weaver and blake failures are quite weird. They look like some sort of drop of connection with the node. Let's retest.

tcclevenger:
@bartgol Test now passes.

components/scream/src/mct_coupling/scream_cxx_f90_interface.cpp
bartgol: So, are you going to do it in this PR?

tcclevenger: I'll split it up.

components/scream/src/control/surface_coupling.cpp
tcclevenger: Ahh, forgot about this as well.

tcclevenger: Actually never mind, I had included it.

bartgol (approved):

tcclevenger:
@bartgol I rebased and cleaned up the commit log. I also looked through the changes and verified they are all there, and then changed the test to check `REQUIRE(a == 0)` (this must have been fix when I fix the SP testing fails?)

Anyways, I'll need one last approval.

components/scream/src/control/surface_coupling.hpp
```diff
@@ -138,10 +153,13 @@ class SurfaceCoupling {
 
   field_mgr_ptr m_field_mgr;
 
-  int           m_num_imports;
+  int           m_num_scream_imports;
+  int           m_num_cpl_imports;
   int           m_num_exports;
+  int           m_num_custom_exports;
```
bartgol, line 159: It seems that this var is never set/updated/used?

components/scream/src/control/surface_coupling.hpp
```diff
@@ -41,8 +41,14 @@ class SurfaceCoupling {
...
   explicit SurfaceCoupling (const field_mgr_ptr& field_mgr);
 
-  // This allocates some service views
-  void set_num_fields (const int num_imports, const int num_exports);
+  // This allocates some service views. Since not all imported
+  // data is used by SCREAM, we distinguish between cpl
+  // imports and SCREAM imports for book keeping.
+  void set_num_fields (const int num_cpl_imports, const int num_scream_imports, const int num_exports);
+
+  // Version of the above function when num_total_imports = num_scream_imports
```
bartgol, line 49: Might also update the comment `num_total_imports` -> `num_cpl_imports`.

components/scream/src/control/surface_coupling.hpp
tcclevenger: Yep, this got added back in all the rebases.. I'll remove.

bartgol (approved):

components/scream/src/mct_coupling/scream_cxx_f90_interface.cpp
tcclevenger: Fix in this PR: https://github.com/E3SM-Project/scream/pull/1152

PR #1129, opened 2021-07-26
Only compute SW radiative fluxes for daytime columns

### LINKED ISSUE #1102: Radiation interface needs to do solar radiative transfer only on daytime columns
Currently, our radiative interface code passes all columns (both day and night) to the RRTMGP solvers. This is a problem in the shortwave, because the codes are only valid for daytime columns (cosine of solar zenith angle greater than 0) and the RRTMGP codes assume that only daytime values are passed. The approach in EAM is to subset the inputs to the solar code to daytime-only values ahead of time. We should probably do the same here, although we *could* consider adding conditionals to the shortwave kernels to simply pass over the nighttime columns. This second approach may degrade performance more than the cost of the data copies to subset the data ahead of time though.

brhillman:
RRTMGP expects inputs to the shortwave (SW) solvers to be daytime only. Thus, no checks are made that the cosine of the solar zenith angle is greater than 0 (that the sun is above the horizon). Thus, floating point exceptions are encountered if nighttime columns are passed, due to division by the cosine of the solar zenith angle in the RRTMGP codes. To prevent this, we need to subset the inputs to the SW code to contain only daytime columns, and then expand the outputs back to all columns (filling in zeros for the fluxes where columns are not sunlit). This is consistent with the EAM treatment. Fixes #1102

### DIFF
--- a/components/scream/tests/coupled/shoc_cld_p3_rrtmgp/CMakeLists.txt
+++ b/components/scream/tests/coupled/shoc_cld_p3_rrtmgp/CMakeLists.txt
@@ -14,12 +14,10 @@ SET (NEED_LIBS shoc cld_fraction p3 scream_rrtmgp rrtmgp ${NETCDF_C} scream_cont
 # Add source files
 set (SRC shoc_cld_p3_rrtmgp.cpp ${SCREAM_BASE_DIR}/src/physics/rrtmgp/tests/rrtmgp_test_utils.cpp)
 
-# Test atmosphere processes. Temporarily disabled for AT because it has FPEs. Also
-# disabled if valgrind is ON because it takes too long.
-if (NOT SCREAM_AUTOTESTER)
-  CreateUnitTest(shoc_cld_p3_rrtmgp "${SRC}" "${NEED_LIBS}" LABELS "shoc;cld;p3;rrtmgp;physics")
-  set_target_properties(shoc_cld_p3_rrtmgp PROPERTIES COMPILE_FLAGS "${YAKL_CXX_FLAGS}")
-  target_include_directories(shoc_cld_p3_rrtmgp PUBLIC
+# Test atmosphere processes.
+CreateUnitTest(shoc_cld_p3_rrtmgp "${SRC}" "${NEED_LIBS}" LABELS "shoc;cld;p3;rrtmgp;physics")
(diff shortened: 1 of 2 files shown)


brhillman:
Added WIP flag until I test and debug on GPU.

AaronDonahue (changes requested):
We will also want to reactivate the `shoc_cld_p3_rrtmgp` test once this PR goes in.  So we should remove the conditional at this line: https://github.com/E3SM-Project/scream/blob/master/components/scream/tests/coupled/shoc_cld_p3_rrtmgp/CMakeLists.txt#L19

On the same note, if the homme_physics fix goes in earlier than this PR we can remove the equivalent conditional in that CMakeLists.txt file.

brhillman:
Removed WIP flag since this is passing my local tests on Weaver with the new commits. @AaronDonahue I also removed the conditional around the coupled test for the autotester.

components/scream/tests/coupled/shoc_cld_p3_rrtmgp/CMakeLists.txt
```diff
@@ -14,12 +14,10 @@ SET (NEED_LIBS shoc cld_fraction p3 scream_rrtmgp rrtmgp ${NETCDF_C} scream_cont
 # Add source files
 set (SRC shoc_cld_p3_rrtmgp.cpp ${SCREAM_BASE_DIR}/src/physics/rrtmgp/tests/rrtmgp_test_utils.cpp)
 
-# Test atmosphere processes. Temporarily disabled for AT because it has FPEs. Also
-# disabled if valgrind is ON because it takes too long.
-if (NOT SCREAM_AUTOTESTER)
-  CreateUnitTest(shoc_cld_p3_rrtmgp "${SRC}" "${NEED_LIBS}" LABELS "shoc;cld;p3;rrtmgp;physics")
-  set_target_properties(shoc_cld_p3_rrtmgp PROPERTIES COMPILE_FLAGS "${YAKL_CXX_FLAGS}")
-  target_include_directories(shoc_cld_p3_rrtmgp PUBLIC
+# Test atmosphere processes.
```
AaronDonahue, line 17: @brhillman , can you remove this flag for the `homme_shoc_cld_p3_rrtmgp` coupled test as well?

AaronDonahue (changes requested):

AaronDonahue (approved):
Thanks for making this happen Ben.

PR #1132, opened 2021-07-27
Replace SROF with MOSART for the F compsets

wlin7:
All the SCREAM F compsets are using stub component
SROF as the river model. This PR sets to use MOSART
as the active river model for all SCREAM F compsets,
except for FDPSCREAMA97. Enabling MOSART however has
no impact on the atm simulation results.

[BFB]


PeterCaldwell (approved):
Ok, approving the switch to having SCREAMv0.1 (and perhaps eventually SCREAMv1) sims use an active MOSART river runoff model rather than the SROF stub. This decision is based on the understanding (based on ne30) that MOSART computational time is negligible and its output addition will be minimal. These assumptions may not hold at ne1024, in which case we'll undo this PR. Our motivation for turning on MOSART is that we expect analysts could be very interested in looking at runoff output at 3 km.

wlin7:
This PR cause diffs with mapping files in seq_maps.rc and new file mosart_in. It also causes diff in two files in cpl.hi: 
```
x2l_Flrr_volr: River channel total water volume
x2l_Flrr_volrmch: River channel main channel water volume
```

PR #1136, opened 2021-07-28
Hommexx: Fix a Cuda 10, c++14, gcc 7 internal error.

ambrad:


AaronDonahue:
@ambrad , as recommended I tried to run this branch on lassen before approving.  Unfortunately, it didn't fix the issue.  That being said, I noted that it was failing in `test-all-scream` during the baseline generation step.  So I did two experiments:
1. I copy/pasted the command line and tried it directly myself, rather than through `test-all-scream`.  That produced the same error.
2. Then I removed the `-DSCREAM_BASELINES_ONLY=ON ` from the cmake line and retried, and that worked.  So it seems to be isolated to baseline generation.

For reference, here is the full line that `test-all-scream` is using: 
```
cmake -C /p/gpfs1/donahue5/E3SM_simulations/SCREAM/nightly_tests/lassen/scream/components/scream/cmake/machine-files/lassen.cmake -DNetCDF_C_PATHS=/usr -DCMAKE_BUILD_TYPE=Debug -DSCREAM_PACK_SIZE=1 -DSCREAM_SMALL_PACK_SIZE=1 -DEKAT_DEFAULT_BFB=True -DEKAT_DISABLE_TPL_WARNINGS='ON' -DCMAKE_CXX_COMPILER=/usr/tce/packages/spectrum-mpi/spectrum-mpi-rolling-release-gcc-7.3.1/bin/mpicxx -DCMAKE_C_COMPILER=/usr/tce/packages/spectrum-mpi/spectrum-mpi-rolling-release-gcc-7.3.1/bin/mpicc -DCMAKE_Fortran_COMPILER=/usr/tce/packages/spectrum-mpi/spectrum-mpi-rolling-release-gcc-7.3.1/bin/mpifort -DSCREAM_DYNAMICS_DYCORE=HOMME -DSCREAM_BASELINES_ONLY=ON /p/gpfs1/donahue5/E3SM_simulations/SCREAM/nightly_tests/lassen/scream/components/scream
```

AaronDonahue:
For complete coverage I am going to retry building the current master (w/o this branch merged) to isolate whether or not this branch made any difference in the behavior.  I'll report back when I'm done with that.

AaronDonahue:
Update, without this PR the normal build fails.  So it looks like this should be the solution we need but I can't figure out why the `-DSCREAM_BASELINES_ONLY=ON` would make a difference.

AaronDonahue (approved):
I have confirmed that these changes fix the issue we see on lassen.

ambrad:
There seems to be a problem with Jenkins, so I recommend manually merging this PR.

AaronDonahue:
@ambrad , okay cool.  I'll go ahead and do that.

PR #1141, opened 2021-07-30
HOMME: remove pointless cmake targets for cxx (homme) targets.

bartgol:
There are cmake targets built inside the preqx/theta folders, which are not used by any test. They can be either executables or libraries (depending on how HOMME was configured). Not building these can save some compilation time. More importantly, though, these targets (which are handled internally by homme, with no inputs from scream) seem to be responsible for some internal compiler errors on CUDA, so removing them is actually a priority, to get the AT working (though we could probably have also manipulated the target property to remove it from the 'all' target, but that seems a bit more involved and intrusive).

Note: this commit is replicated in the upstream PR in E3SM.

### DIFF
--- a/components/homme/src/preqx_kokkos/CMakeLists.txt
+++ b/components/homme/src/preqx_kokkos/CMakeLists.txt
@@ -242,20 +242,3 @@ MACRO(PREQX_KOKKOS_SETUP)
   SET(USE_OPENACC FALSE)
 
 ENDMACRO(PREQX_KOKKOS_SETUP)
-
-# Build exec only if HOMME_BUILD_EXECS is true
-IF (HOMME_BUILD_EXECS)
-  PREQX_KOKKOS_SETUP()
-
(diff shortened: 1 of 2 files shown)


tcclevenger (approved):

tcclevenger:
I'm guessing this is the cause of failures in https://github.com/E3SM-Project/scream/pull/1128? Can't seem to replicate on Weaver, so hopefully it is..

bartgol:
> I'm guessing this is the cause of failures in #1128? Can't seem to replicate on Weaver, so hopefully it is..

I believe so. These targets seem to only have issues in the release build, btw, so if you tried debug, you probably could not replicate the errors.

bartgol:
Weaver had a weird jenkins error (probably connection with the executor was dropped), but from the jenkins log I could see that the build phase went through. Adding the retest tag, but this should indeed fix the internal compiler error issue.

PR #1142, opened 2021-07-30
Change exner to inv_exner in SHOC

tcclevenger:
It appears that C++ shoc is using the `inv_exner`, but calling it `exner`. Example: 
```
update_host_dse C++:
  temp = (thlm + (lcond/cp)*ql)/exner 
  host_dse = cp*temp + ggr*zt + phis

update_host_dse F90:
  temp = (thlm + (lcond/cp)*ql)/inv_exner 
  host_dse = cp*temp + ggr*zt + phis
```
This PR changes the name.

Also, in `atmosphere_macrophysics.hpp` `exner` was being computed and passed into `shoc_main()` where it is used as `inv_exner`. This PR corrects this.

Lastly, we expect `shoc_run_and_cmp` to fail since we needed to change it to input `inv_exner`(see comments).

### DIFF
--- a/components/scream/src/physics/shoc/atmosphere_macrophysics.hpp
+++ b/components/scream/src/physics/shoc/atmosphere_macrophysics.hpp
@@ -102,9 +102,15 @@ public:
       const auto sub_z_int = ekat::subview(z_int, i);
       const auto s_z_int = ekat::scalarize(sub_z_int);
       Kokkos::parallel_for(Kokkos::TeamThreadRange(team, nlev_packs), [&] (const Int& k) {
-        // Exner
+        const auto range = ekat::range<IntSmallPack>(k*Spack::n);
+        const Smask in_nlev_range = (range < nlev);
+
+        // Inverse of Exner. Assert that exner != 0 when in range before computing.
         const Spack p_mid_ik(p_mid(i,k));
-        exner(i,k)  = PF::exner_function(p_mid_ik);
+        const Spack exner_ik = PF::exner_function(p_mid_ik);
+        const Smask nonzero = (exner_ik != 0);
+        EKAT_KERNEL_ASSERT((nonzero || !in_nlev_range).any());
+        inv_exner(i,k).set(nonzero, 1.0/exner_ik);
 
         tke(i,k) = ekat::max(sp(0.004), tke(i,k));
 
@@ -206,7 +212,7 @@ public:
     view_1d              vpwp_sfc;
     view_2d              wtracer_sfc;
     view_2d              wm_zt;
-    view_2d              exner;
+    view_2d              inv_exner;
     view_2d              thlm;
     view_2d              qw;
     view_2d              cloud_frac;
@@ -224,7 +230,7 @@ public:
                        const view_2d& s_, const view_2d& rrho_, const view_2d& rrho_i_,
                        const view_2d& thv_, const view_2d& dz_,const view_2d& zt_grid_,const view_2d& zi_grid_, const view_1d& wpthlp_sfc_,
                        const view_1d& wprtp_sfc_,const view_1d& upwp_sfc_,const view_1d& vpwp_sfc_, const view_2d& wtracer_sfc_,
-                       const view_2d& wm_zt_,const view_2d& exner_,const view_2d& thlm_,const view_2d& qw_)
+                       const view_2d& wm_zt_,const view_2d& inv_exner_,const view_2d& thlm_,const view_2d& qw_)
     {
       ncol = ncol_;
       nlev = nlev_;
@@ -263,7 +269,7 @@ public:
       vpwp_sfc = vpwp_sfc_;
       wtracer_sfc = wtracer_sfc_;
       wm_zt = wm_zt_;
-      exner = exner_;
+      inv_exner = inv_exner_;
       thlm = thlm_;
       qw = qw_;
     } // set_variables
@@ -361,7 +367,7 @@ public:
     uview_2d<Spack> zi_grid;
     uview_2d<Spack> wtracer_sfc;
     uview_2d<Spack> wm_zt;
-    uview_2d<Spack> exner;
+    uview_2d<Spack> inv_exner;
     uview_2d<Spack> thlm;
     uview_2d<Spack> qw;
     uview_2d<Spack> s;
--- a/components/scream/src/physics/shoc/atmosphere_macrophysics.hpp
+++ b/components/scream/src/physics/shoc/atmosphere_macrophysics.hpp
@@ -110,7 +110,7 @@ public:
         const Spack exner_ik = PF::exner_function(p_mid_ik);
         const Smask nonzero = (exner_ik != 0);
         EKAT_KERNEL_ASSERT((nonzero || !in_nlev_range).any());
-        inv_exner(i,k).set(nonzero, 1.0/exner_ik);
+        inv_exner(i,k).set(nonzero, 1/exner_ik);
 
         tke(i,k) = ekat::max(sp(0.004), tke(i,k));
 
--- a/components/scream/src/physics/shoc/shoc_ic_cases.cpp
+++ b/components/scream/src/physics/shoc/shoc_ic_cases.cpp
@@ -54,7 +54,7 @@ void flip_vertically(FortranData& d)
   flip_vertically(d.presi);
   flip_vertically(d.pdel);
   flip_vertically(d.w_field);
-  flip_vertically(d.exner);
+  flip_vertically(d.inv_exner);
   flip_vertically(d.host_dse);
   flip_vertically(d.tke);
   flip_vertically(d.thetal);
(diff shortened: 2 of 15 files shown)


components/scream/src/physics/shoc/atmosphere_macrophysics.hpp
```diff
@@ -102,9 +102,15 @@ class SHOCMacrophysics : public scream::AtmosphereProcess
...
+        const auto range = ekat::range<IntSmallPack>(k*Spack::n);
+        const Smask in_nlev_range = (range < nlev);
+
+        // Inverse of Exner. Assert that exner != 0 when in range before computing.
         const Spack p_mid_ik(p_mid(i,k));
-        exner(i,k)  = PF::exner_function(p_mid_ik);
+        const Spack exner_ik = PF::exner_function(p_mid_ik);
+        const Smask nonzero = (exner_ik != 0);
+        EKAT_KERNEL_ASSERT((nonzero || !in_nlev_range).any());
+        inv_exner(i,k).set(nonzero, 1.0/exner_ik);
```
tcclevenger, line 113: In the AD, we were passing `exner`, now we change the name to `inv_exner`, compute the inverse of `exner`, and pass this to `shoc_main`. Internally, `shoc_main` now matches in both cpp and f90.

components/scream/src/physics/shoc/shoc_ic_cases.cpp
```diff
@@ -202,8 +202,8 @@ FortranData::Ptr make_standard(const Int shcol, Int nlev, Int num_qtracers) {
     // Compute pressure differences and host_dse / exner.
     for (Int k = 0; k < nlev; ++k) {
       d.pdel(i, k) = std::abs(d.presi(i, k+1) - d.presi(i, k));
-      d.exner(i, k) = pow(d.pres(i, k)/consts::P0, consts::Rair/consts::Cpair);
-      d.host_dse(i, k) = consts::Cpair * d.exner(i, k) * d.thv(i, k) +
+      d.inv_exner(i, k) = 1.0/pow(d.pres(i, k)/consts::P0, consts::Rair/consts::Cpair);
+      d.host_dse(i, k) = consts::Cpair * d.thv(i, k)/d.inv_exner(i, k) +
```
tcclevenger, line 206: This change causes the `shoc_run_and_cmp` tests to fail. The baselines are created with the old version where the `inv_exner` passed to `shoc_main()` (F90) actually was `exner`. Since C++ and F90 were doing the same computation, they matched, even though the value for `inv_exner` (in f90) and the computation with `exner` (c++) didn't make sense.

components/scream/src/physics/shoc/tests/shoc_energy_update_dse_tests.cpp
```diff
@@ -38,8 +38,8 @@ struct UnitWrap::UnitTest<D>::TestShocUpdateDse {
 
     // Liquid water potential temperature [K]
     static constexpr Real thlm[nlev] = {350, 325, 315, 310, 300};
-    // Exner function [-]
-    static constexpr Real exner[nlev] = {0.1, 0.3, 0.5, 0.7, 1.0};
+    // Inverse of the Exner function [-]
+    static constexpr Real inv_exner[nlev] = {0.1, 0.3, 0.5, 0.7, 1.0};
```
tcclevenger, line 42: @bogensch Do these values actually make sense for `inv_exner`? If I change to the inverse of these values this test (`shoc_energy_host_dse_property`) will fail with `REQUIRE(host_dse[k+1] - host_dse[k] < 0)`.

components/scream/src/physics/shoc/tests/shoc_main_tests.cpp
```diff
@@ -133,8 +133,10 @@ struct UnitWrap::UnitTest<D>::TestShocMain {
       thetal[n] = pot_temp - (LatVap/Cpair)*shoc_ql[n];
       // Virtual potential temperature
       thv[n] = pot_temp * (1 + 0.61*qv - shoc_ql[n]);
-      // Compute exner function
-      exner[n] = std::pow(pres[n]/p0,Rair/Cpair);
+      // Compute the inverse of the exner function
+      const Real exner = std::pow(pres[n]/p0,Rair/Cpair);
+      REQUIRE(exner > 0);
+      inv_exner[n] = 1.0/exner;
```
tcclevenger, line 139: This property test (`shoc_main_property`) actually passes with the new `inv_exner`.

tcclevenger (commented):
I've left some comments. These changes are essentially just name change, however, changing the input data for `shoc_run_and_cmp` test causes expected failures.

components/scream/src/physics/shoc/tests/shoc_energy_update_dse_tests.cpp
bogensch: @tcclevenger yes, this is wrong. BUT, I already fixed this for PR #1028 , when I pushed the change into @AaronDonahue branch (https://github.com/E3SM-Project/scream/commit/8cf79bbfb256e9c411bd73d58ac5a48597fa86f0https://github.com/E3SM-Project/scream/commit/8cf79bbfb256e9c411bd73d58ac5a48597fa86f0https://github.com/E3SM-Project/scream/commit/8cf79bbfb256e9c411bd73d58ac5a48597fa86f0https://github.com/E3SM-Project/scream/commit/8cf79bbfb256e9c411bd73d58ac5a48597fa86f0).  However, for some reason this did not make it onto master?  I see on May 7th on PR #1028 @AaronDonahue did a force push of this branch so may be something went wrong there?  In any case, this should be fixed.  @tcclevenger should do you want me to modify/fix these and push to your branch?

AaronDonahue: Ahhhhi imagine this was a case where we had to rebase off of a new master. Which requires a force push. Somehow the last commit didn't make it in. My apologies for that error

jgfouca (changes requested):
All the added instances of `1.0 / ` should be changed to `1 /`. It may not cause any problems in this PR, but it's a good habit to get into.

components/scream/src/physics/shoc/tests/shoc_energy_update_dse_tests.cpp
tcclevenger: @bogensch Yes, that would be good if you could just push the changes here.

jgfouca (approved):

components/scream/src/physics/shoc/tests/shoc_energy_update_dse_tests.cpp
AaronDonahue: Not sure if the fix is in yet, but we can just cherry-pick the git hash that Peter B posted.

tcclevenger:
The only failed tests are expected. I'm adding WIP to keep the auto tester from running. Probably pick up monday.

components/scream/src/physics/shoc/tests/shoc_energy_update_dse_tests.cpp
bogensch: @AaronDonahue or you can just admit this is part of your sneaky plan to sabotage my career ;) .  Kidding, I should have have noticed myself when I signed off on PR #1028 .  @tcclevenger I pushed the corrected tests.

components/scream/src/physics/shoc/shoc_ic_cases.cpp
AaronDonahue: Is the assumption being made here that `phis` is always 0.0?  See this line: https://github.com/E3SM-Project/scream/blob/ef54333db4acd190dabcadb299fb506f2f98b56e/components/eam/src/physics/cam/shoc.F90#L3798

AaronDonahue: I guess so, looking at the line just below.

components/scream/src/physics/shoc/shoc_ic_cases.cpp
```diff
@@ -202,8 +202,8 @@ FortranData::Ptr make_standard(const Int shcol, Int nlev, Int num_qtracers) {
     // Compute pressure differences and host_dse / exner.
```
AaronDonahue, line 202: Minor:  should the line change to "host_dse * inv_exner"?

AaronDonahue (approved):
Looks good Conrad.  Thanks for fixing this bug so quickly.  I had a few questions, but they were mostly minor, so this should be good to go once you are happy with the shoc_run_and_cmp test failing like you expect.

components/scream/src/physics/shoc/shoc_ic_cases.cpp
tcclevenger: Yes it should.

tcclevenger:
Ok, the only test failing is expected. Any objections to me merging? The only non-cosmetic/name changes are to test input values.

bogensch (approved):
thanks for doing this @tcclevenger , looks good to me!

AaronDonahue:
If there are no objections to merging, I can go ahead and do the manual merge.  I'll give it until just after lunchtime, in case some one wants to object :-)

AaronDonahue:
No objections by lunchtime, I'm going to merge.

PR #1143, opened 2021-07-30
Allow output of subfields

bartgol:
This PR allows to output fields that are subfields of another field (which implies that they are strided). For instance, it allows to output a single tracer, rather than the whole tracer array Q.


AaronDonahue (approved):
Thank you Luca for both solving this issue and improving the code in general.

PR #1144, opened 2021-08-02
Jgfouca/improve data randomization

### LINKED ISSUE #550: Manage randomization in unit test input better
Some thoughts:
* Do we want different random (seed=sec since epoch) values every time the test is run?
* Or do we want repeatability? Even if we are running different subsets of tests? I think this would require every test to select a seed.
* Do we want a hybrid where we can get either of the above based on user input?

jgfouca:
I was starting to think the rest of my career would be debugging SHOC random input tests... but I think everything is working now.

Changes
1) Add common infrastructure, `setup_random_test` to make it easy to have randomized tests
2) Handle corner case in p3_upwind to avoid FPE
3) Fix a few P3 tests that were passing different values for scalar inputs to tests that were testing packed data. These values should be held constant.
4) shoc TestData structs that have grid data are now all subclasses of ShocTestGridDataBase, this should ensure all their randomize calls produce non-garbage grid data.
5) Change ShocMainData randomization to produce better inputs.
6) Fix out-of-bounds indexing in shoc_varorcovar_property test.

Fixes #550


components/scream/src/physics/p3/p3_upwind_impl.hpp
```diff
@@ -106,7 +106,7 @@ ::generalized_sedimentation (
 {
   // compute dt_sub
   const Int tmpint1 = static_cast<int>(Co_max + 1);
-  const Scalar dt_sub = ekat::impl::min(dt_left, dt_left/tmpint1);
+  const Scalar dt_sub = tmpint1 == 0 ? dt_left : ekat::impl::min(dt_left, dt_left/tmpint1);
```
jgfouca, line 109: P3 upwind change is here. Would like to get some eyes on this for sure.

components/scream/src/physics/p3/tests/p3_ice_supersat_conservation_tests.cpp
```diff
@@ -17,12 +18,15 @@ struct UnitWrap::UnitTest<D>::TestIceSupersatConservation {
...
+    auto engine = setup_random_test();
+
     IceSupersatConservationData f90_data[max_pack_size];
 
     // Generate random input data
     // Alternatively, you can use the f90_data construtors/initializer lists to hardcode data
     for (auto& d : f90_data) {
-      d.randomize();
+      d.randomize(engine);
+      d.dt = f90_data[0].dt; // hold this fixed, it is not packed data
```
jgfouca, line 29: When testing calls that take packed data, if those calls also take scalars, those scalars should be held constant.

components/scream/src/physics/share/physics_test_data.hpp
```diff
@@ -255,7 +256,51 @@ class PhysicsTestData
   // the member to a range.
   // Example, to use a -1 to 1 range for wthl member:
   // d.randomize({ {d.wthl, {-1, 1}} });
-  void randomize(const std::vector<std::pair<void*, std::pair<Real, Real> > >& ranges = {});
+  template <typename Engine>
+  void randomize(Engine& engine, const std::vector<std::pair<void*, std::pair<Real, Real> > >& ranges = {})
```
jgfouca, line 260: Basically the same as the original code but needed to be moved to header in order to take a template Engine argument.

components/scream/src/physics/shoc/shoc_functions_f90.hpp
```diff
@@ -801,6 +850,122 @@ struct ShocMainData : public PhysicsTestData {
                     num_qtracers(num_qtracers_), dtime(dtime_), nbot_shoc(nbot_shoc_), ntop_shoc(ntop_shoc_) {}
 
   PTD_STD_DEF(ShocMainData, 8, shcol, nlev, nlevi, num_qtracers, dtime, nadv, nbot_shoc, ntop_shoc);
+
+  template <size_t N>
+  Real interpolate_data(const std::array<Real, N>& ref_elevations,
```
jgfouca, line 855: This and compute_column_pressure are copied from shoc_ic_cases.cpp with minor changes to do all columns in a single call.

jgfouca (commented):
Left some annotations for points of interest since this is a big PR.

bartgol (approved):

bartgol:
@jgfouca I only have one request: could you please rewrite the git history, and squash/fixup the commits that simply say `progress`? Or, alternatively, keep them but with a more explanatory message? While you're at it, feel free to squash more commits, if they are logically a single unit. This will help keeping a git history that is easier to navigate and read. Thanks!

components/scream/src/physics/p3/p3_upwind_impl.hpp
ambrad: Courant number. Co_max should be >= 0 on input. I believe all the callers to generalized_sedimentation have code that assures Co_max >= 0. Is it just a unit test that causes tempint1 0?

Separately, what Luca writes is correct: that min doesn't make sense. Seems to have been in the F90 from the start, so it was just copied over.

Taken together, here you could write simply
```C++
assert(Co_max >= 0);
const Scalar dt_sub = dt_left/tmpint1;
```

components/scream/src/control/tests/sc_tests.cpp
```diff
@@ -33,13 +33,7 @@ TEST_CASE ("surface_coupling")
...
   // The random numbers generator
-  std::random_device rd;
-  const unsigned int catchRngSeed = Catch::rngSeed();
-  const unsigned int seed = catchRngSeed==0 ? rd() : catchRngSeed;
-  if (comm.am_i_root()) {
-    std::cout << "seed: " << seed << (catchRngSeed==0 ? " (catch rng seed was 0)\n" : "\n");
-  }
-  rngAlg engine(seed);
+  auto engine = setup_random_test(&comm);
   RPDF pdf(0.0,1.0);
```
tcclevenger, line 37: @jgfouca As of yesterday evening there is a new test case "recreate_mct_coupling" in this file on master that uses the same RNG. So it will need this change as well.

components/scream/src/physics/p3/p3_upwind_impl.hpp
jgfouca: @ambrad , yes. I think this problem came up in one of the random-input bfb tests.

components/scream/src/physics/p3/tests/p3_nc_conservation_tests.cpp
```diff
@@ -52,7 +56,7 @@ struct UnitWrap::UnitTest<D>::TestNcConservation {
         nc_selfcollect_tend[s] = cxx_device(vs).nc_selfcollect_tend;
       }
 
-      Functions::nc_conservation(nc, nc_selfcollect_tend, cxx_device(0).dt, nc_collect_tend, nc2ni_immers_freeze_tend, nc_accret_tend, nc2nr_autoconv_tend);
+      Functions::nc_conservation(nc, nc_selfcollect_tend, cxx_device(offset).dt, nc_collect_tend, nc2ni_immers_freeze_tend, nc_accret_tend, nc2nr_autoconv_tend);
```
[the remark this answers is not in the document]
jgfouca: @bartgol , this is related to the packed vs. non-packed inputs:

> When testing calls that take packed data, if those calls also take scalars, those scalars should be held constant.

Since we are holding the scalars constant, there's no reason to hardcode to the 0th dt, they're all the same. The behavior is the same either way.

components/scream/src/physics/p3/p3_upwind_impl.hpp
bartgol: Then maybe there is some random input that needs better constraints for the pdf?

components/scream/src/control/tests/sc_tests.cpp
jgfouca: @tcclevenger , thanks for the heads-up. I've done another upstream merge and changed this test to use setup_random_test.

jgfouca:
@bartgol , unfortunately, I've done a number of upstream merges to this branch which makes re-writing old history problematic. I will try to remember to squash these progress commits in the future.

bogensch (approved):
Thanks for doing this, Jim.  I only looked at the SHOC related changes and they all look good to me.

PR #1147, opened 2021-08-03
Some changes to how we extract a kokkos view from a Field.

### LINKED ISSUE #1145: Remove all use of `get_view` from Field, and possibly remove that method too.
This is causing way too many bugs, especially since we started allowing subfields (which, for technical reasons, do not store the subfield view inside, but the view of the 'parent' field).

The last example is the field property checks: the checks are implemented in a rank-agnostic way, by looping over the entries in `field.get_view()`. This is already somewhat wrong due to the fact that the view might contain padding (for pack sizes larger than 1), but it's flat out bad in case of subfields.

I'm adding this to v1, since it's showed up twice in the last week (slowing down development, while tracking down seemingly odd behaviors).

Note: I'm not pointing the finger and blaming people for these bugs. It's just a confusing thing, which I always wanted to get rid of, and is now showing its dangerousness in all its might.

bartgol:
This PR attempts to get rid (or, as it ended up being, better handle) the ability to extract the 1d view from a Field. The reason is that the apparently innocuous method `get_view` was getting dangerous to use, and was indeed already used in contexts where it produced buggy behavior. The most dangerous case is when the current field is indeed the subfield of another field. For technical reason, in this case we store the same 1d view as the 'parent' field, doing the correct reshaping (and striding) only when the N-d view is requested (via the old `get_reshaped_view` method). Originally, I was thinking of getting rid of `get_view` altogether. However, it is handy in a few selected scenarios, where we are already ensuring that the field is indeed _not_ the subfield of another field. Therefore, I kept this capability around, albeit with some modifications.

This PR basically does the following:

* change `get_view` to get_flattened_view`, to emphasize that we are flattening a field (even though the view is _always_ stored as a 1d view internally).
* change `get_reshaped_view` to `get_view`, since that should be the most common case, so let's give it the "simple" name, rather than the more involved one.
* make `get_flattened_view` throw if the field is a subfield of another field, since the resulting 1d view would _not_ be what the user thinks it is. If for some weird logic, you indeed wanted the stored 1d view, you can get that from the parent field (assuming you have access to it).

To the reviewers: I am considering making `get_flattened_view` throw also if the field contains padding, since the 1d view will contain garbage (the padding) at selected location in the middle of the 1d array. I'm hesitant to do that, since the places that use `get_flattened_view` are already paying attention to how they access the field (e.g., reshaping back to a N-d view later on), and use this method as a way to quickly grab a pointer to the data. Perhaps, we can go through these few uses (mainly: surface coupling, scorpio I/O, and pd remapper), and see if we can switch to using `get_view<DT>` (the old 'reshaped'), and if we can, then we can make `get_flattened_view` throw in case of padding (or straight on remove the method). What do you think?

Closes #1145 .


components/scream/src/share/field/field_property_checks/field_monotonicity_check.hpp
```diff
@@ -23,10 +23,10 @@ class FieldMonotonicityCheck: public FieldPropertyCheck<RealType> {
   // Overrides.
 
   // The name of the field check
-  std::string name () const { return "Monotonicity Field Check"; }
+  std::string name () const override { return "Monotonicity Field Check"; }
 
   bool check(const Field<const_RT>& field) const override {
-    auto view = field.get_view();
+    auto view = field.get_flattened_view();
```
AaronDonahue, line 29: A general comment for all of the field checks.  Now that `get_flattened_view` will throw an error if a field is actually a subfield we will probably want to rethink this approach.  At this point it would not be possible to field check any of the tracers.

Since we are not performing the field checking by default we don't need to fix this yet.  But will want to address this sooner than later.

AaronDonahue (commented):
@bartgol thank you for cleaning the get_view command up.  I know it's been on the TODO for awhile.  I have one general comment.  I'm not sure if we want to put the fix in this PR or leave it for a subsequent PR.

components/scream/src/control/surface_coupling.cpp
```diff
@@ -148,13 +148,13 @@ register_export (const std::string& fname,
   if (m_field_mgr->has_field(fname)) {
 
     // If fname is a field in the the field manager, then set the data and column info based on this field.
-    // Note that the field must not have a parent as get_view() would be unsafe.
+    // Note that the field must not have a parent as get_flattened_view() would be unsafe.
```
tcclevenger, line 151: This comment is actually incorrect since we allow for fields with a parent. Got left in I guess from a previous version of SurfaceCoupling that I cycled through. Could you remove it? Just the part `// Note that the field...`

components/scream/src/control/surface_coupling.cpp
```diff
@@ -148,13 +148,13 @@ register_export (const std::string& fname,
...
     export_field_type field = m_field_mgr->get_field(fname);
 
     EKAT_REQUIRE_MSG (field.is_allocated(), "Error! Export field view has not been allocated yet.\n");
 
     // Set view data ptr
-    info.data = field.get_view().data();
+    info.data = field.get_flattened_view().data();
```
tcclevenger, line 157: I believe that now when `qv` is registered in the surface coupling unit test this should throw since it is a subfield. The logic in surface coupling will need to change (currently we register fields with parents, and `get_col_info` checks if it is subfield and computes the correct stride).

tcclevenger (commented):
This looks mostly good to me. I like that we keep `get_flattened_view` around, but that make it clear that it can't be a subview. The only issue I see is that we allow for subfields in the `SurfaceCoupling::register_import/export`, and internally take care of striding correctly. This will need to change.

jeff-cohere:
@bartgol , I like the idea of going through the use cases for `get_flattened_view` and seeing whether they can be converted to use `get_view`, which allows `get_flattened_view` to throw on padded fields. This is a good example of clarifying the intent of `get_flattened_view`. We always have to be careful about how much power we give developers in C++, because past a certain point, the abstractions can lose all meaning.

components/scream/src/control/surface_coupling.cpp
tcclevenger: Looking at the testing log, this is exactly the test that fails.

components/scream/src/share/field/field.hpp
```diff
@@ -135,7 +128,27 @@ class Field {
...
+  // this field may be a subfield of another one. In that case, the stored
+  // view would be that of the parent field, which would need to be read
+  // in a strided fashion. Furthermore, if the field has been allocated
+  // in a way that allows packing, there might be padding. However, the
+  // padding is inserted in the last physical dimension of the field,
+  // so when one sees the 1d view, the padding is interleaved in the array.
+  // For these reasons, this must be regarded as a *dangerous* power-user
+  // method. Do not use unless you know what you are doing.
+  // WARNING: this method throws if this field is a subfield. It does *NOT*
+  //          throw if there is padding, so careful how you access/use the view.
```
jeff-cohere, line 143: Maybe this is a bad idea, but something to consider, anyway:

We could have an argument to this called `allow_padding`, set to `false` by default. If this argument is `true`, it would allow this method to be called on fields with padding. Otherwise it would throw.

components/scream/src/share/field/field_property_checks/field_monotonicity_check.hpp
bartgol: Yes, that's true. I think it won't be long till we actually need that fix. If you think it's fine, I can add it to this PR.

components/scream/src/share/field/field.hpp
bartgol: Something along that line might actually not be a bad idea.

components/scream/src/control/surface_coupling.cpp
bartgol: Ah, right. I will have to rework the register method then, since I don't want to allow `get_flattened_view()` for subfields. Will it work if I extract the parent, and  grab the pointer from that? Meaning, does the current impl of `get_col_info` correctly offset a subfield from the start _of the parent field's_ view?

bartgol:
What if I renamed `get_flattened_view()` to something less user-friendly, like `get_internal_storage()`, and make it return a raw pointer rather than a 1d view? This should be enough to discourage its usage, limiting it to very selected places, with all the caveats needed.

components/scream/src/control/surface_coupling.cpp
tcclevenger: Yes, `get_col_info` assumes that a `subview.data()` is exactly the `parent_view.data()`. So something like
```
if (field has a parent) {
  info.data = parent_field.get_flattened_view.data();
} else {
  info.data = field.get_flattened_view.data();
}
```
should work.

jeff-cohere:
If you think that's practical, maybe `data()` would be better, since it matches the same construction in Kokkos.

tcclevenger:
Is there anywhere that actually uses `get_flattened_view()` as a view itself? Seems to always be `get_flattened_view().data()`. So I would be in favor of it returning a raw pointer.

components/scream/src/share/field/field_property_checks/field_monotonicity_check.hpp
AaronDonahue: I'm happy to have it added here if it's not too much work.

bartgol:
I'm still getting errors for the restart tests.

AaronDonahue:
> I'm still getting errors for the restart tests.

Let me see if I can track down the cause.

components/scream/src/share/io/tests/restart.cpp
```diff
@@ -66,12 +66,30 @@ TEST_CASE("restart","io")
...
+            for (int i=0; i<fl.dim(0); ++i) {
+              v(i) += dt;
+            }
+          }
+          break;
+        case 2:
+          {
+            auto v = f.get_view<Real**,Host>();
+            for (int i=0; i<fl.dim(0); ++i) {
+              for (int j=0; j<fl.dim(0); ++j) {
```
AaronDonahue, line 85: Shouldn't this be `fl.dim(1)`?

AaronDonahue (approved):

PR #1148, opened 2021-08-03
Add radiation to SurfaceCoupling

tcclevenger:
To add radiation to SurfaceCoupling, we need the FieldManager to split the input field 
`surf_alb_direct` (size `(ncol, nswbands)`) into `sfc_alb_dir_vis` and `sfc_alb_dir_nir` (size`(ncol)` each) and similarly, `surf_alb_diffuse` into `sfc_alb_dif_vis` and `sfc_alb_dif_nir`. Then, using wave information in the rrtmgp interface, we can combine these 4 new fields into the band-by-band version originally in the FM.

In the SCREAM RRTMGP interface, we add a routine `compute_band_by_band_surface_albedos` which combines the broadband albedos into band-by-band albedos. The inputs to `rrtmgp_main` do not change. Now the `SurfaceCoupling::register_import()` can just ask for these new fields, so only changes to the unit test are required here.


AaronDonahue:
@tcclevenger , we will probably need to update the initial condition file for rrtmgp variables in the `shoc_cld_p3_rrtmgp` and `homme_shoc_cld_p3_rrtmgp` "coupled" tests too.  Originally, these are generated directly from a SCREAMv0 run.  I could regenerate these, although I would need to know which variables represent the new variables in the F90 code.  Otherwise, if there is an easy way to break single variables into multiple variables, which you have done in the PR I could manipulate the data offline and recreate an appropriate init-condition file.  In the meantime I would suspect that the `shoc_cld_p3_rrtmgp` test will fail the AT.

tcclevenger:
@AaronDonahue I'll keep this as WIP so the AT isn't annoying.

components/scream/src/physics/rrtmgp/tests/rrtmgp_test_utils.cpp
```diff
@@ -40,15 +40,20 @@ namespace rrtmgpTest {
...
+            real1d &mu0,
             real2d &lwp, real2d &iwp, real2d &rel, real2d &rei, real2d &cld) {
 
         // Setup boundary conditions, solar zenith angle, etc
         // NOTE: this stuff would come from the model in a real run
 
         // Ocean-ish values for surface albedos, just for example
-        memset(sfc_alb_dir , 0.06_wp );
-        memset(sfc_alb_dif , 0.06_wp );
+        memset(sfc_alb_dir_vis , 0.06_wp );
```
brhillman, line 52: Hmm it’s a bit of a bummer that the rrtmgp_standalone test requires us to change this here, as this would prevent us from testing with band-varying surface albedos in the future, and also causes this test to diverge a little more from what was in the RRTMGP repo. I don’t really see an easy way around this though, and I don’t see any plans to change how the model handles surface albedos, so this is probably fine.

components/scream/tests/uncoupled/rrtmgp/rrtmgp_stand_alone.cpp
```diff
@@ -132,13 +134,25 @@ namespace scream {
...
             lwp, iwp, rel, rei, cld
         );
-        //
+
+        // Set surface albedos to 0.06. This is the initial condition used
+        // is other rrtmgp tests.
+        memset(sfc_alb_dir_vis , 0.06_wp );
+        memset(sfc_alb_dir_nir , 0.06_wp );
+        memset(sfc_alb_dif_vis , 0.06_wp );
+        memset(sfc_alb_dif_nir , 0.06_wp );
```
tcclevenger, line 154: @brhillman These few lines in rrtmgp_standalone (137-141 above as well) are the tradeoff for not changing any of the other rrtmgp tests.

brhillman (approved):
I think this is in good shape, once we generate the new initial conditions for the test. Thanks for doing this!

tcclevenger:
@AaronDonahue Thanks for the .nc file. 

@brhillman @AaronDonahue It's ready for review.

brhillman (approved):

PR #1149, opened 2021-08-05
WIP: P3 bugfix: as hydrometeors sediment, recalculate the top of the layer with condensate

PeterCaldwell:
P3 currently computes the top and bottom layers in each column which contain condensate mass before substepping sedimentation to ensure CFL stability. The generalized_sedimentation routine moves the bottom layer containing condensate down one level each time it is called, ensuring sedimentation is able to fall to the surface without getting caught. The top layer is never updated even though condensate can fall out of it. This would be fine except that the top layer actually containing condensate can fall all the way to the surface (i.e. you started with condensate but it all rained out during the CFL substepping). When this happens, the code crashes because it can't calculate a CFL number when nothing is sedimenting. Recalculating the precipitating layer top every substep also may result in less layers to process and hence a performance gain... but finding the top every substep probably offsets that improvement. 

This bugfix should be BFB since it just skips cells which didn't have anything in them anyways.

### DIFF
--- a/components/eam/src/physics/cam/micro_p3.F90
+++ b/components/eam/src/physics/cam/micro_p3.F90
@@ -3745,53 +3745,69 @@ subroutine rain_sedimentation(kts,kte,ktop,kbot,kdir,   &
       dt_left   = dt  !time remaining for sedi over full model (mp) time step
       prt_accum = 0._rtype  !precip rate for individual category
 
+      !find bottom. Note that generalized_sedimentation moves k_qxbot down 1
+      !level every time it is called (as appropriate for an explicit advection
+      !scheme. Thus bottom is accurately defined for the rest of this scheme.
+      do k = kbot,k_qxtop,kdir
+         if (qr(k).ge.qsmall) then
+            k_qxbot = k
+            exit
+         endif
+      enddo
+
       substep_sedi_r: do while (dt_left.gt.1.e-4_rtype)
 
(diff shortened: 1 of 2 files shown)


PeterCaldwell:
Ugh, I just realized that both F90 and C++ need code to make sure all calculations are skipped if find_bottom says there's no condensate in the column. Also I've implemented changes for rain sed, but probably need to make identical changes for ice and cloud sed.

bartgol:
> I noticed this problem after p3_stand_alone failed for #1062 due to all rain sedimenting out of the domain, causing the CFL in C++ to go negative. F90 seems to default to 0 CFL in that case.

Was CFL=0 set by F90 conceptually wrong, meaning, did it ignore that some rain might have completely left the column? If so, can this have impacted dyamond simulations?

PeterCaldwell:
> Was CFL=0 set by F90 conceptually wrong, meaning, did it ignore that some rain might have completely left the column? If so, can this have impacted dyamond simulations?
I don't think CFL=0 in F90 was wrong, but not recomputing the bottom layer with precip was wrong in all our previous simulations. I have a vague sense that there was probably always enough precip in lower levels that precip getting trapped in the atm didn't happen very often. We will see what the impact is when I finish the fixes.

PeterCaldwell:
Closing because Jim inadvertently fixed this big in #1144 by setting C++ min CFL to 0 instead of -1!

PR #1151, opened 2021-08-06
Homme should not try to manage its own flags

jgfouca:
When being built for SCREAM


jgfouca:
HOMME exists in a number of contexts: standalone, as a subcomponent of E3SM, and as a subcomponent of SCREAM. I think the approach in this PR makes it the most clear that Homme manages its own flags when in standalone or part of E3SM, but not when part of SCREAM.

Without this PR, the flags for Homme were kind of a mess when building as part of SCREAM. The compile lines were a mix of debug and optimization flags, as well as mixed stdlibs. This caused summit to be unable to build. With this PR, the flags look a lot better and summit builds and runs homme_shoc_cld_p3_rrtmgp_ut_np1_omp1 in both debug and release.

components/homme/CMakeLists.txt
```diff
@@ -159,11 +159,14 @@ IF (HOMME_ENABLE_COMPOSE)
   SET (HOMME_USE_KOKKOS TRUE)
 ENDIF ()
 
-IF (HOMME_USE_CXX)
-  MESSAGE(STATUS "This configuration of HOMME requires a C++ compiler")
-  ENABLE_LANGUAGE(CXX)
-  SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
-ENDIF ()
+IF (NOT "${PROJECT_NAME}" STREQUAL "SCREAM")
```
bartgol, line 162: I think this may be an issue down the road. E.g., when scream is merged in e3sm, I'm not sure if this check would fail. Also, it would be perfectly legal for homme devs to declare homme as a project, regardless of whether there is a host project or not (cmake lets you do that, and it should arguably be done here).

- at the top of this file, set the bool HOMME_STANDALONE based on whether PROJECT_NAME is defined or not.
- in the rest of the file, use `if (HOMME_STANDALONE)` to isolate these chunks of code that should only be run in standalone mode.

bartgol (changes requested):
I have two requests. The first is to isolate changes to the homme folder in a separate commit, so that I can cherry pick it and add it to the upstream PR with all the scream-related changes in homme.

The second request is to be more robust with the check for an existing host project (see inline comment).

components/homme/CMakeLists.txt
jgfouca: @bartgol , the situation is complicated. I think Homme wants to manage its own flags when it is a subcomponent of E3SM, right?

bartgol: Yes, but when building homme in EAM, the cmake system is not used. EAM uses homme via the perl/bash/python scripts.

jgfouca: @bartgol , thanks, I did not know that.

components/homme/CMakeLists.txt
```diff
@@ -23,6 +25,8 @@ ELSE ()
   # By default, do not build executables/tests, and build 'library-style' targets
   OPTION (HOMME_BUILD_EXECS "Whether Homme should build executables and tests." OFF)
   OPTION (HOMME_BUILD_LIBS  "Whether Homme should build a library for each dycore target." ON)
+
+  SET(HOMME_STANDALONE TRUE)
```
bartgol, line 29: Copy+paste error?

components/homme/CMakeLists.txt
```diff
@@ -470,7 +479,7 @@ ADD_CUSTOM_TARGET(clean-all
    COMMAND ${CMAKE_COMMAND} -P "${CMAKE_CURRENT_SOURCE_DIR}/cmake/clean-cmake-files.cmake"
 )
 
-IF (NOT "${PROJECT_NAME}" STREQUAL "HOMME")
+IF (NOT HOMME_STANDALONE)
```
bartgol, line 482: Since scream does not use these flags, should we completely remove these lines?

components/scream/src/dynamics/homme/CMakeLists.txt
```diff
@@ -37,25 +37,16 @@ set (BUILD_HOMME_THETA_KOKKOS ON CACHE BOOL "")
...
 set (HOMMEXX_VECTOR_SIZE ${SCREAM_PACK_SIZE} CACHE STRING "")
 
 # configure Homme folder
 add_subdirectory(${HOMME_SOURCE_DIR} ${HOMME_BINARY_DIR})
 
-# *APPEND* scream flags to homme flags
-set (HOMME_C_FLAGS       ${HOMME_C_FLAGS} ${TEMP_C_FLAGS})
-set (HOMME_CXX_FLAGS     ${HOMME_CXX_FLAGS} ${TEMP_CXX_FLAGS})
-set (HOMME_Fortran_FLAGS ${HOMME_Fortran_FLAGS} ${TEMP_F_FLAGS})
+# Use same flags as SCREAM
```
bartgol, line 46: If we use the same flags in homme that we use in scream, all uses of HOMME_<LANG>_FLAGS can be removed, I think.

components/homme/CMakeLists.txt
jgfouca: Yikes! Indeed.

components/scream/src/dynamics/homme/CMakeLists.txt
jgfouca: I was wondering that too. I'll go ahead and remove

jgfouca:
@bartgol , done.

components/homme/CMakeLists.txt
```diff
@@ -159,11 +163,14 @@ IF (HOMME_ENABLE_COMPOSE)
...
-  MESSAGE(STATUS "This configuration of HOMME requires a C++ compiler")
-  ENABLE_LANGUAGE(CXX)
-  SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
-ENDIF ()
+IF (HOMME_STANDALONE)
+  # Homme needs to setup its own flags
+  IF (HOMME_USE_CXX)
+    MESSAGE(STATUS "This configuration of HOMME requires a C++ compiler")
+    ENABLE_LANGUAGE(CXX)
+    SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
```
bartgol, line 171: I think this might be not right for standalone either. Kokkos uses c++14, and either way, the standard should be set via the apposite cmake commands, rather than stuffed in the flags.

But there's no need to fix it here, so it's fine.

components/scream/src/dynamics/homme/CMakeLists.txt
bartgol: Are you going to remove these lines (and all the occurrences of HOMME_<LANG>_FLAGS below) ?

components/homme/CMakeLists.txt
jgfouca: I agree, this is a hacky approach. I will let homme devs decide what to do.

bartgol:
@jgfouca Unfortunately the mkl flag is added to the big bunch of CMAKE_<LANG>_FLAGS, rather than added as compile option to all generated cmake targets.

The cleanest solution is to modify this behaivor, and have homme add `-mkl` (if mkl is enabled) to each target's compile option, and remove it from the CMAKE_<LANG>_FLAGS. I can do this on Monday, if you want.

jgfouca:
@bartgol , I just checked that this still builds and runs on summit. This is ready for final review.

bartgol:
@jgfouca Ok, let me run homme testing, to make sure this is fine.

bartgol:
@jgfouca It's a bit complicated to check this version of homme, since it doesn't really compile. Turns out we messed up a few things, which got however fixed in the upstream PR containing all scream changes, which is (knock on wood) quite close to be integrated into master (figuring out the last few test fails).

My proposal: can we put this PR on hold, do an upstream merge once E3SM-Project/E3SM#4365 goes in, and then rebase this PR on master?

jgfouca:
@bartgol , sure. I don't think there's any rush to merge this. I can do my summit work on this branch.

bartgol:
Great. I added the WIP label, so the AT doesn't keep writing useless messages.

jgfouca:
@bartgol , the upstream merge has been done on master and I merged latest scream master into this branch.

jgfouca:
It looks like homme_shoc_cld_p3_rrtmgp_ut_np1_omp1 still builds and runs on summit.

bartgol:
Upon discussion with @jgfouca , we decided to hold off this PR, and add its changes to another PR in the upstream. Adding the WIP label, so the AT doesn't keep posting stuff.

PR #1153, opened 2021-08-09
Include SPA interface for SCREAM, part 1

AaronDonahue:
This commit introduces the SPA interface into the SCREAM code and establishes a simple unit test.  This is only part 1 of a series of PR's because it only handles temporal and vertical interpolation.  The final two steps will include the horizontal interpolation component and including more unit and property tests.


components/scream/src/physics/spa/spa_functions_impl.hpp
```diff
@@ -0,0 +1,169 @@
...
+      using C = scream::physics::Constants<Real>;
+      static constexpr auto P0 = C::P0;
+      const Int nk_pack = ekat::npack<Spack>(pressure_state.nlevs);
+      Kokkos::parallel_for(
+        Kokkos::TeamThreadRange(team, nk_pack), [&] (Int k) {
+          p_src_sub(k)    = ps_src * pressure_state.hybm(k) + P0 * pressure_state.hyam(k);
+          ccn3_src_sub(k) = ccn3_beg_sub(k) + slope * (ccn3_end_sub(k) - ccn3_beg_sub(k));
+      });
+      team.team_barrier();
+      Kokkos::parallel_for(
```
ambrad, line 79: This should have a TeamThreadRange argument.

components/scream/src/physics/spa/spa_functions_impl.hpp
```diff
@@ -0,0 +1,169 @@
...
+        Kokkos::parallel_for(
+          Kokkos::TeamThreadRange(team, nk_pack), [&] (Int k) {
+          aer_g_sw_src_sub(k)   = aer_g_sw_beg_sub(k)   + slope * (aer_g_sw_end_sub(k)   - aer_g_sw_beg_sub(k));
+          aer_ssa_sw_src_sub(k) = aer_ssa_sw_beg_sub(k) + slope * (aer_ssa_sw_end_sub(k) - aer_ssa_sw_beg_sub(k));
+          aer_tau_sw_src_sub(k) = aer_tau_sw_beg_sub(k) + slope * (aer_tau_sw_end_sub(k) - aer_tau_sw_beg_sub(k));
+        });
+        team.team_barrier();
+      });
+      Kokkos::parallel_for(
+        nlwbands, [&] (int n) {
```
ambrad, line 102: Same here.

components/scream/src/physics/spa/spa_functions_impl.hpp
```diff
@@ -0,0 +1,169 @@
...
+
+        const auto& aer_ssa_sw_beg_sub         = ekat::subview(data_beg.AER_SSA_SW, i, n);
+        const auto& aer_ssa_sw_end_sub         = ekat::subview(data_end.AER_SSA_SW, i, n);
+        const auto& aer_ssa_sw_src_sub         = ekat::subview(aer_ssa_sw_src, i, n);
+  
+        const auto& aer_tau_sw_beg_sub         = ekat::subview(data_beg.AER_TAU_SW, i, n);
+        const auto& aer_tau_sw_end_sub         = ekat::subview(data_end.AER_TAU_SW, i, n);
+        const auto& aer_tau_sw_src_sub         = ekat::subview(aer_tau_sw_src, i, n);
+
+        Kokkos::parallel_for(
```
ambrad, line 93: Since this is nested in a team loop, this should be ThreadVectorRange. Or perhaps the outer loop is meant to be a kernel launch.

components/scream/src/physics/spa/spa_functions_impl.hpp
```diff
@@ -0,0 +1,169 @@
...
+        });
+        team.team_barrier();
+      });
+      Kokkos::parallel_for(
+        nlwbands, [&] (int n) {
+        const auto& aer_tau_lw_beg_sub         = ekat::subview(data_beg.AER_TAU_LW, i, n);
+        const auto& aer_tau_lw_end_sub         = ekat::subview(data_end.AER_TAU_LW, i, n);
+        const auto& aer_tau_lw_src_sub         = ekat::subview(aer_tau_lw_src, i, n);
+
+        Kokkos::parallel_for(
```
ambrad, line 107: Same nesting issue here.

components/scream/src/physics/spa/spa_functions_impl.hpp
```diff
@@ -0,0 +1,169 @@
...
+      VertInterp.setup(team,
+                       ekat::subview(p_src,i),
+                       ekat::subview(pressure_state.pmid,i));
+      team.team_barrier();
+      VertInterp.lin_interp(team,
+                            ekat::subview(p_src,i),
+                            ekat::subview(pressure_state.pmid,i),
+                            ekat::subview(ccn3_src,i),
+                            ekat::subview(data_out.CCN3,i));
+      Kokkos::parallel_for(
```
ambrad, line 134: Same issues here.

components/scream/src/physics/spa/spa_functions_impl.hpp
```diff
@@ -0,0 +1,168 @@
...
+      const auto& aer_ssa_sw_end_sub         = ekat::subview(data_end.AER_SSA_SW, i);
+      const auto& aer_ssa_sw_src_sub         = ekat::subview(aer_ssa_sw_src, i);
+  
+      const auto& aer_tau_sw_beg_sub         = ekat::subview(data_beg.AER_TAU_SW, i);
+      const auto& aer_tau_sw_end_sub         = ekat::subview(data_end.AER_TAU_SW, i);
+      const auto& aer_tau_sw_src_sub         = ekat::subview(aer_tau_sw_src, i);
+
+      Kokkos::parallel_for(
+        Kokkos::ThreadVectorRange(team, nswbands), [&] (Int n) {
+        aer_g_sw_src_sub(n,k)   = aer_g_sw_beg_sub(n,k)   + slope * (aer_g_sw_end_sub(n,k)   - aer_g_sw_beg_sub(n,k));
```
bartgol, line 95: Same here: save one read by factoring `xyz_beg_sub`.

components/scream/src/physics/spa/spa_functions_impl.hpp
```diff
@@ -0,0 +1,168 @@
...
+      const auto& aer_ssa_sw_beg_sub         = ekat::subview(data_beg.AER_SSA_SW, i);
+      const auto& aer_ssa_sw_end_sub         = ekat::subview(data_end.AER_SSA_SW, i);
+      const auto& aer_ssa_sw_src_sub         = ekat::subview(aer_ssa_sw_src, i);
+  
+      const auto& aer_tau_sw_beg_sub         = ekat::subview(data_beg.AER_TAU_SW, i);
+      const auto& aer_tau_sw_end_sub         = ekat::subview(data_end.AER_TAU_SW, i);
+      const auto& aer_tau_sw_src_sub         = ekat::subview(aer_tau_sw_src, i);
+
+      Kokkos::parallel_for(
+        Kokkos::ThreadVectorRange(team, nswbands), [&] (Int n) {
```
bartgol, line 94: This threading is not optimal for GPU. In particular, the ThreadVectorRange should be on the fastest index. The way it is now, we have all threads in the warp working on separate bands, which are far away in memory, causing non-coalesced access. You should swap `nk_pack` and `nswbands` in the two ||fors.

components/scream/src/physics/spa/spa_functions_impl.hpp
```diff
@@ -0,0 +1,168 @@
...
+    const Int nk_pack = ekat::npack<Spack>(pressure_state.nlevs);
+    Kokkos::parallel_for(
+      Kokkos::TeamThreadRange(team, nk_pack), [&] (Int k) {
+        p_src_sub(k)    = ps_src * pressure_state.hybm(k) + P0 * pressure_state.hyam(k);
+        ccn3_src_sub(k) = ccn3_beg_sub(k) + slope * (ccn3_end_sub(k) - ccn3_beg_sub(k));
+    });
+    team.team_barrier();
+    Kokkos::parallel_for(
+      Kokkos::TeamThreadRange(team, nk_pack), [&] (int k) {
+      const auto& aer_g_sw_beg_sub           = ekat::subview(data_beg.AER_G_SW, i);
```
bartgol, line 81: These subviews don't depend on k, so you could take them _before_ entering the ||for, rather than at every k iteration. Or, alternatively, subview both i and k. The latter is actually not possible as the code is, since it is not slicing the slowest indices. However, read the comment below: if you swap k and n in the two ||fors, then you can slice away column and band index.

components/scream/src/physics/spa/spa_functions_impl.hpp
```diff
@@ -0,0 +1,168 @@
...
+    team.team_barrier();
+    Kokkos::parallel_for(
+      Kokkos::TeamThreadRange(team, nk_pack), [&] (int k) {
+      const auto& aer_tau_lw_beg_sub         = ekat::subview(data_beg.AER_TAU_LW, i);
+      const auto& aer_tau_lw_end_sub         = ekat::subview(data_end.AER_TAU_LW, i);
+      const auto& aer_tau_lw_src_sub         = ekat::subview(aer_tau_lw_src, i);
+
+      Kokkos::parallel_for(
+        Kokkos::ThreadVectorRange(team, nlwbands), [&] (Int n) {
+        aer_tau_lw_src_sub(n,k) = aer_tau_lw_beg_sub(n,k) + slope * (aer_tau_lw_end_sub(n,k) - aer_tau_lw_beg_sub(n,k));
```
bartgol, line 109: Same comment re: convex combination.

bartgol: Also, same same comment about order of || fors.

AaronDonahue:
@bartgol can you review this PR now?  It has passed the AT.

I also did an offline verification of the output from the SPA interface against a python code that handled the equivalent vertical and temporal interpolation and both solutions appear to agree very closely, on the relative order of 1e-7 in the L-inf norm and 1e-5 in the L2 norm.  Not sure if this is close enough for us to feel comfortable.  But we would expect C++ and python to not be BFB.

components/scream/src/physics/spa/spa_functions_impl.hpp
```diff
@@ -0,0 +1,168 @@
...
+                     ekat::subview(pressure_state.pmid,i));
+    team.team_barrier();
+    VertInterp.lin_interp(team,
+                          ekat::subview(p_src,i),
+                          ekat::subview(pressure_state.pmid,i),
+                          ekat::subview(ccn3_src,i),
+                          ekat::subview(data_out.CCN3,i));
+    Kokkos::parallel_for(
+      Kokkos::TeamThreadRange(team, nlwbands), [&] (int n) {
+      VertInterp.lin_interp(team,
```
bartgol, line 136: I don't think you can call lin_interp from _within_ a TeamThreadRange loop. From what I see in ekat, lin interp dispatches _its own_ TeamThreadRange loop.

@jgfouca Do you think it makes sense (and is it possible?) to make lin_interp templated on the policy, so that one can call it with either a TTR (and vec_length=1) or from a TVR?

bartgol (changes requested):
I think there are some things that need to be fixed (on the kokkos side), and some other open for debate.

components/scream/src/physics/spa/atmosphere_prescribed_aerosol.cpp
```diff
@@ -0,0 +1,216 @@
...
+  FieldLayout scalar3d_swband_layout { {COL,SWBND, LEV}, {m_num_cols, m_nswbands, m_num_levs} }; 
+  FieldLayout scalar3d_lwband_layout { {COL,LWBND, LEV}, {m_num_cols, m_nlwbands, m_num_levs} }; 
+
+  // Set of fields used strictly as input
+  constexpr int ps = Pack::n;
+  add_field<Required>("p_mid"      , scalar3d_layout_mid, 1/kg,   grid_name, ps);
+  add_field<Required>("hyam"       , scalar1d_layout_mid, nondim, grid_name, ps); // TODO: These fields should  be loaded from file and not registered with the field manager.
+  add_field<Required>("hybm"       , scalar1d_layout_mid, nondim, grid_name, ps); // TODO: These fields should  be loaded from file and not registered with the field manager.
+
+  add_field<Required>("PS_beg"     , scalar2d_layout,     Pa,     grid_name, ps); // TODO: These fields should  be loaded from file and not registered with the field manager.
```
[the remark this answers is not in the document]
AaronDonahue: We could wait, the idea here was to get a version of SPA that works in case we have parallel tasks that could benefit from SPA being at least able to be run in some way, while we wait on the SCORPIO changes we need to do this correctly.

components/scream/src/physics/spa/spa_functions.hpp
```diff
@@ -0,0 +1,128 @@
...
+  using view_1d_ptr_array = typename KT::template view_1d_ptr_carray<S, N>;
+
+  /* ------------------------------------------------------------------------------------------- */
+  // SPA structures to help manage all of the variables:
+  struct SPATimeState {
+    SPATimeState() = default;
+    // The current month
+    Int current_month;
+    // Julian Date for the beginning of the month
+    Real t_beg_month;
```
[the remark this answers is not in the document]
AaronDonahue: yeah, for sure.  I stole this terminology from EAM which uses it in the same context for radiation.  But we can also just call it "days_from_beggining_of_month" or something like that.

components/scream/src/share/util/scream_time_stamp.cpp
```diff
@@ -116,13 +116,21 @@ double TimeStamp::get_julian_day () const {
...
     julianday += dpm(yy,m);
   }
   return julianday;
 }
 
+int TimeStamp::get_dpm () const {
+  return dpm(m_yy,m_mm);
+}
+
+int TimeStamp::get_dpm (const int yy, const int mm) const {
```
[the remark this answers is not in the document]
AaronDonahue: Is it possible to call `dpm` from outside the class?  i.e. `TimeStamp.dpm()` or something like that.  I just needed a way to query how many days are in the current month for the temporal interpolation and I recall running into an issue trying to access `dpm` from the spa interface, hence this wrapper function.

AaronDonahue:
@bartgol , thank you for the review comments. I'll get to work on putting in the changes you suggested.

components/scream/src/physics/spa/spa_functions.hpp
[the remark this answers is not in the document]
bartgol: We can use `julian_date`, but perhaps we should add an explanation where `get_julian_date` is declared, and put a note here saying "See comments in blah.hpp for the def of Julian date".

components/scream/src/physics/spa/spa_functions_impl.hpp
bartgol: TBC: this code is not _incorrect_, it is just _inefficient_, since the reads/writes are not coalesced.

components/scream/src/physics/spa/spa_functions.hpp
```diff
@@ -0,0 +1,128 @@
...
+    // Current simulation Julian Date
+    Real t_now;
+    // Number of days in the current month, cast as a Real
+    Real days_this_month;
+  }; // SPATimeState
+
+  struct SPAPressureState {
+    SPAPressureState() = default;
+    // Number of vertical levels for the data
```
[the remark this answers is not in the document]
AaronDonahue: We could use `ncols/nlevs_spa` for the spa data.  My original intent was to have everything stored in a SPA data structure to be spa data and refer to spa data.  So ncols, nlevs, in this context is by definition a reference to SPA data.  In the actual interface I pass the simulation ncols and nlevs.

components/scream/src/physics/spa/spa_functions_impl.hpp
AaronDonahue: Yes, thank you for the comment, I was unsure of the most efficient way to set up this parallelism.  I've followed your suggestion and switch the order of the loops.

components/scream/src/physics/spa/spa_functions_impl.hpp
AaronDonahue: @jgfouca , I'll wait for your comment on this before making any changes.

components/scream/src/physics/spa/spa_functions.hpp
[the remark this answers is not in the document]
bartgol: I guess we can leave `ncols` and `nlevs` here, since this struct only stores one resolution. But in places where there are two res handled, we should make an effort to make as clear as possible which is which..

components/scream/src/physics/spa/spa_functions_impl.hpp
jgfouca: @bartgol , @AaronDonahue , you're right, LinInterp expects to be able to dispatch its own TeamThreadRange kernel. I'm actually surprised that this code worked. I will look at changing the interface.

jgfouca: Another thought, would it make sense to change these team thread range kernels to regular for-loops?

jgfouca: Spoke with @bartgol offline and we agreed to pursue the LinInterp templating

AaronDonahue: @jgfouca , thanks for getting back on this.  So, should we wait on the change in EKAT and then incorporate the updated EKAT in this PR?

jgfouca: See https://github.com/E3SM-Project/EKAT/pull/124

components/scream/src/share/util/scream_time_stamp.cpp
```diff
@@ -116,13 +116,21 @@ double TimeStamp::get_julian_day () const {
 
 double TimeStamp::get_julian_day (const int yy, const int mm, const int dd, const double ss) const {
   // Initialize Julian Day
-  double julianday = (dd+1) + ss;
+  double julianday = dd + ss;
```
bartgol, line 119: I just noticed: this adds things quite unrelated: the day of the month and the seconds of the day. Perhaps you wanted to do something like `double julianday = dd*3600 + ss;`?

AaronDonahue: ahh yes, good catch.  I think I was thrown off by ss being a double that maybe it was fraction of day.  But it's double to allow for `dt<1s`.  I've fixed this in the newest version to `dd+ss/86400` since we are calculating Julian "day" in units of days.

components/scream/src/physics/spa/spa_functions_impl.hpp
```diff
@@ -0,0 +1,168 @@
...
+  Int nlevs,
+  Int nswbands,
+  Int nlwbands)
+{
+  // Gather time stamp info
+  auto& t_now = time_state.t_now;
+  auto& t_beg = time_state.t_beg_month;
+  auto& t_len = time_state.days_this_month;
+
+  // For now we require that the Data in and the Data out have the same number of columns.
```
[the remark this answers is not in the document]
AaronDonahue: I've added robust comments above and changed the name of the variable to `ncols_atm` and `nlevs_atm` respectively.

components/scream/src/physics/spa/spa_functions.hpp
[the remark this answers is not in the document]
AaronDonahue: I've added in a comment that hopefully will save the next person who looks at this code.

components/scream/src/physics/spa/spa_functions_impl.hpp
jgfouca: @AaronDonahue , yes, lets wait and use the new LinInterp API which better-supports your use case here. I actually don't think the code you have here will work on GPU, but I could be wrong.

AaronDonahue: huh, it seems to pass the tests on Weaver, so I wonder if it's not actually running that section of code

AaronDonahue:
@bartgol , while we are waiting on the EKAT PR, can you re-review to confirm I addressed all of your other comments? I also had a response to one question that I would like your feedback on.

components/scream/src/share/util/scream_time_stamp.cpp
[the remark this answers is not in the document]
bartgol: You need `dpm` to be declared (but not necessarily defined) in the hpp file, if you want it to be available to other files.

components/scream/src/physics/spa/spa_functions_impl.hpp
bartgol: I don't think kokkos (or cuda) checks that you are nesting the || layers correctly. You can probably write a TTR inside a TVR inside a TTR, and kokkos/cuda would not complain (but you'd probably get garbage out of it).

components/scream/src/physics/spa/spa_functions_impl.hpp
```diff
@@ -0,0 +1,168 @@
...
+    /* Reconstruct the vertical pressure profile for the data and time interpolation
+     * of the data */
+    {
+    using C = scream::physics::Constants<Real>;
+    static constexpr auto P0 = C::P0;
+    const Int nk_pack = ekat::npack<Spack>(pressure_state.nlevs);
+    Kokkos::parallel_for(
+      Kokkos::TeamThreadRange(team, nk_pack), [&] (Int k) {
+        p_src_sub(k)    = ps_src * pressure_state.hybm(k) + P0 * pressure_state.hyam(k);
+        ccn3_src_sub(k) = ccn3_beg_sub(k) + slope * (ccn3_end_sub(k) - ccn3_beg_sub(k));
```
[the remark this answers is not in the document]
bartgol: You could still do this mod across the file.

AaronDonahue:
Okay, with the new EKAT PR in this PR should be ready.  @bartgol can you re-review just to confirm?

components/scream/src/physics/spa/spa_functions_impl.hpp
jgfouca: @AaronDonahue , now that you have the latest EKAT, you'll need a slight modification to all the lin_interp calls that happen within a TeamThreadRange.  I will add some comments to help you.

components/scream/src/physics/spa/spa_functions_impl.hpp
```diff
@@ -0,0 +1,285 @@
...
+  // This is done using the EKAT linear interpolation routine, see /externals/ekat/util/ekat_lin_interp.hpp
+  // for more details. 
+  using LIV = ekat::LinInterp<Real,Spack::n>;
+  Real minthreshold = 0.0;  // Hard-code a minimum value for aerosol concentration to zero.
+
+  LIV VertInterp(ncols_atm,pressure_state.nlevs,nlevs_atm,minthreshold);
+  /* Parallel loop strategy:
+   * 1. Loop over all simulation columns (i index)
+   * 2. Where applicable, loop over all aerosol bands (n index)
+   */
```
jgfouca, line 218: You'll need to use a custom policy.
```
const Int most_bands = std::max(nlwbands, nswbands);
LIV::TeamPolicy policy(ncols_atm, ekat::OnGpu<LIV::ExeSpace>::value ? most_bands: 1, VertInterp.km2_pack());
```

components/scream/src/physics/spa/spa_functions_impl.hpp
```diff
@@ -0,0 +1,285 @@
...
+    /* Conduct vertical interpolation for the 2D variable CCN3 */
+    VertInterp.lin_interp(team,
+                          ekat::subview(p_src,i),
+                          ekat::subview(pressure_state.pmid,i),
+                          ekat::subview(ccn3_src,i),
+                          ekat::subview(data_out.CCN3,i));
+    /* Conduct vertical interpolation for the LW banded data - nlwbands (n index) */
+    Kokkos::parallel_for(
+      Kokkos::TeamThreadRange(team, nlwbands), [&] (int n) {
+      VertInterp.lin_interp(team,
```
jgfouca, line 237: You'll need to set up a ThreadVectorRange here:
```
const auto& tvr = Kokkos::ThreadVectorRange(team, VertInterp.km2_pack());
```
... and pass it as the second arg to lin_interp.

components/scream/src/physics/spa/spa_functions_impl.hpp
```diff
@@ -0,0 +1,285 @@
...
+                            ekat::subview(p_src,i),
+                            ekat::subview(pressure_state.pmid,i),
+                            ekat::subview(aer_tau_lw_src,i,n),
+                            ekat::subview(data_out.AER_TAU_LW,i,n));
+    });
+    /* Conduct vertical interpolation for the SW banded data - nswbands (n index) */
+    Kokkos::parallel_for(
+      Kokkos::TeamThreadRange(team, nswbands), [&] (int n) {
+      VertInterp.lin_interp(team,
```
jgfouca, line 246: You'll need to set up a ThreadVectorRange here:
```
const auto& tvr = Kokkos::ThreadVectorRange(team, VertInterp.km2_pack());
```
... and pass it as the second arg to lin_interp.

jgfouca (changes requested):

components/scream/src/physics/spa/spa_functions_impl.hpp
jgfouca: @bartgol , can you confirm this is the correct policy?

components/scream/src/physics/spa/spa_functions_impl.hpp
AaronDonahue: @jgfouca , Does this mean that all `lin_interp` calls need an extra argument now?

jgfouca: @AaronDonahue , just when you are using a custom boundary, so the lin_interp calls that occur inside a TeamThreadRange kernel.

jgfouca (approved):

components/scream/src/physics/spa/spa_functions_impl.hpp
jgfouca: @AaronDonahue , I gave you bad code (doesn't work on CUDA). You'll need to capture `VertInterp.km2_pack()` in a constant before the kernel is launched. You can do this right after the VertInterp object is constructed.

jgfouca (changes requested):

components/scream/src/physics/spa/spa_functions_impl.hpp
AaronDonahue: @bartgol  said there may be a fix coming in EKAT.  Should I just wait on that?  Or would I still need to capture VertInterp.km2_pack() as a constant?

bartgol: I would wait for the fix. It will allow a cleaner code, without the need for the local temporary.

jgfouca (approved):
Nevermind, I will just change EKAT to support calls to km2_pack() from device.

AaronDonahue:
@bartgol and @jgfouca , I just now updated the EKAT sub-module to capture the lin_interp hot-fix.  Can you both take another look through and see if this PR is ready to go in?

jgfouca:
@AaronDonahue , what about the weaver fail?

AaronDonahue:
ahhh it looks like sp failed.  The log doesn't say which test, but it would make sense that the test would be the new spa test.  I'll take a look.

bartgol:
> ahhh it looks like sp failed. The log doesn't say which test, but it would make sense that the test would be the new spa test. I'll take a look.

Since it doesn't report the name of a test failed, it means there were errors in either the config or build stage. I guess we could improve a bit test-all-scream, to be more verbose on this regard.

AaronDonahue:
I think the SPA test is timing out. This happened once before and was because my for loops were incorrect. Maybe there was an mpi barrier that was inside a for loop. Is it possible my implementation of Lin interp after the hot fix is now doing something similar? @jgfouca

components/scream/src/physics/spa/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
...
+)
+
+set(CLDFRAC_HEADERS
+  atmosphere_prescribed_aerosol.hpp
+  spa_functions.hpp
+  spa_functions_impl.hpp
+)
+
+add_library(spa ${CLDFRAC_SRCS})
+target_include_directories(spa PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/../share)
```
bartgol, line 13: You would need this only if you include stuff from physics/share without the "physics/share" part, like
```
#include "physics_constants.hpp"
``` 
We don't do that (and you don't do it in this PR, if I'm not mistaken). So you can/should remove this line.

components/scream/src/physics/spa/atmosphere_prescribed_aerosol.cpp
```diff
@@ -0,0 +1,193 @@
...
+// =========================================================================================
+void SPA::finalize_impl()
+{
+  // Do nothing
+}
+
+void SPA::set_required_field_impl (const Field<const Real>& f) {
+
+  const auto& name = f.get_header().get_identifier().name();
+  m_spa_fields_in.emplace(name,f);
```
bartgol, line 178: This method will have to be removed, following what's done in #1165 , which I think will go in before this PR, given it's been approved and passed testing, while I'm gonna request changes here, :)

components/scream/src/physics/spa/atmosphere_prescribed_aerosol.hpp
```diff
@@ -0,0 +1,125 @@
...
+protected:
+
+  // The three main overrides for the subcomponent
+  void initialize_impl (const util::TimeStamp& t0);
+  void initialize_spa_impl ();
+  void run_impl        (const Real dt);
+  void finalize_impl   ();
+
+  // Setting the fields in the atmospheric process
+  void set_required_field_impl (const Field<const Real>& f);
```
bartgol, line 88: To be removed from the header too, obv.

components/scream/src/physics/spa/atmosphere_prescribed_aerosol.hpp
```diff
@@ -0,0 +1,125 @@
...
+  void set_computed_field_impl (const Field<      Real>& f);
+
+  // Computes total number of bytes needed for local variables
+  int requested_buffer_size_in_bytes() const;
+
+  // Set local variables using memory provided by
+  // the ATMBufferManager
+  void init_buffers(const ATMBufferManager &buffer_manager);
+
+  std::map<std::string,const_field_type>  m_spa_fields_in;
```
bartgol, line 98: To be removed too.

components/scream/src/physics/spa/atmosphere_prescribed_aerosol_hack.cpp
```diff
@@ -0,0 +1,40 @@
+#include "atmosphere_prescribed_aerosol.hpp"
```
bartgol, line 1: I'm ok with this file, very temporarily though (if possible). As soon as the PR goes in, I would like to remove the hack, and fix it, if possible.

components/scream/src/physics/spa/spa_functions_impl.hpp
```diff
@@ -0,0 +1,294 @@
...
+    const Int nk_pack = ekat::npack<Spack>(pressure_state.nlevs);
+    Kokkos::parallel_for(
+      Kokkos::TeamThreadRange(team, nk_pack), [&] (Int k) {
+        // Reconstruct vertical temperature profile using the hybrid coordinate system.
+        p_src_sub(k)    = ps_src * pressure_state.hybm(k) + P0 * pressure_state.hyam(k);
+        // Time interpolation for CCN3
+        ccn3_src_sub(k) = linear_interp(ccn3_beg_sub(k),ccn3_end_sub(k),t_norm);
+    });
+    team.team_barrier();
+    }
```
bartgol, line 163: What's the point of these scope blocks? I don't think they are needed, for they don't help with temps scoping. Besides, the indentation is off. Can we remove them?

components/scream/src/physics/spa/spa_functions_impl.hpp
```diff
@@ -0,0 +1,294 @@
...
+   * 2. Where applicable, loop over all aerosol bands (n index)
+   */ 
+  const Int most_bands = std::max(nlwbands, nswbands);
+  typename LIV::TeamPolicy band_policy(ncols_atm, ekat::OnGpu<typename LIV::ExeSpace>::value ? most_bands : 1, VertInterp.km2_pack());
+  Kokkos::parallel_for("vertical-interp-spa",
+    band_policy,
+    KOKKOS_LAMBDA(typename LIV::MemberType const& team) {
+    const int i = team.league_rank();
+    /* Setup the linear interpolater for this column. */
+    VertInterp.setup(team,
```
bartgol, line 227: This call to `setup` and the following one to `lin_interp` will dispatch a TTR, but given that the vector length is greater than 0, there will be 32 threads running the setup/lin_interp at the same time. I think everything "works", since both `setup` and `lin_interp` don't contain lines like `some_view(k) += blah`, but always do += on thread-local variables, and always use plain assignment when it's time to write into the view. So every thread will write on the output views, but they will all write the same thing.

That said, I _think_ this is not good programming. As soon as someone makes a tiny change to setup/lin_interp, all the code that used them with TTR boundaries, but with a policy with vector length>1 could start failing, and the bug would be _very hard_ to find. Furthermore, 32 threads all trying to write at the same memory location is not very performant (not coalesced access).

I suggest doing somethign like this:
```
Kokkos::single(Kokkos::PerTeam(team),[&](){
  const auto tvr = Kokkos::ThreadVectorRange(team, VertInterp.km2_pack());
  VectInterp.setup(team,tvr,...);
});
```
and similarly for the 2d lin_interp call.

components/scream/src/physics/spa/spa_functions_impl.hpp
bartgol: I think it is correct.

bartgol: @jgfouca  

I find it annoying to have to do the switch based on OnGpu<ExeSpace> though. In Hommexx, we handled this from our `get_default_team_policy` util, which used a ThreadPreferences input struct to specify the max team size (usually the length of the 2nd nested for) and the max vector size (the length of the 3rd nested loop), whether to prefer threads over vectors, and whether to prefer large teams even if the outer ||for had several iters. The routine then, based on thread preferences and arch, would spit out the team policy you needed.

We might add something like that to ekat, at some point. You can find what I'm talking about in `components/homme/src/share/cxx/ExecSpaceDefs.*pp`.

components/scream/src/share/util/scream_time_stamp.cpp
bartgol: @AaronDonahue This still seems wrong to me...

bartgol (changes requested):
I think there are still a few things to clean up. Some cmake stuff, some related to #1165, some related to having a TTR without a TVR inside, and one related to time stamp math. See inline comments.

components/scream/src/physics/spa/atmosphere_prescribed_aerosol_hack.cpp
AaronDonahue: Agreed

components/scream/src/physics/spa/spa_functions_impl.hpp
AaronDonahue: I wanted to follow up on this since the comment is a bit old.  Is there anything left to do here?  Should I be deleting the `tvr` variables being passed to the `lin_interp` calls?

bartgol: No, no, you need to keep those `tvr`'s.

components/scream/src/physics/spa/spa_functions_impl.hpp
```diff
@@ -0,0 +1,297 @@
...
+    if (team.team_rank()==0) {
+      const auto tvr = Kokkos::ThreadVectorRange(team, VertInterp.km2_pack());
+    
+      VertInterp.lin_interp(team,tvr,
+                            ekat::subview(p_src,i),
+                            ekat::subview(pressure_state.pmid,i),
+                            ekat::subview(ccn3_src,i),
+                            ekat::subview(data_out.CCN3,i));
+    }
+    team.team_barrier();
```
bartgol, line 239: This team barrier is pointless. Other team members can get started with the 3d interpolation, they don't need to wait for the 2d interpolation to be done. The only barrier needed is the one after the setup, since all team members need to waif for the setup to be done.

bartgol (changes requested):
One last micro-change needed, then this is good to go.

components/scream/src/physics/spa/spa_functions_impl.hpp
AaronDonahue: Should be fixed now.  Go ahead and re-review, and if it's ready then we can let the AT merge it.  Thanks for all of the review comments.

bartgol (approved):

PR #1154, opened 2021-08-09
Remove the dependence of the scorpio input and output on the grids manager

AaronDonahue:
In the I/O interface the grids_manager was only really needed to
determine the global degree's of freedom each rank is responsible
for.  Which is in turn used with PIO to read/write input/output.

In reality, it should be possible to pass an arbitrary grid with this
information, which would in principle allow I/O to handle files that
are not on the simulation grid.

In particular, this is a necessary precursor to full implementation
of SPA in SCREAM.

### DIFF
--- a/components/scream/src/share/io/output_manager.hpp
+++ b/components/scream/src/share/io/output_manager.hpp
@@ -110,12 +110,12 @@ protected:
   ekat::Comm                                   atm_comm;
   ekat::Comm                                   pio_comm; 
   ekat::ParameterList                          m_params;
-  std::shared_ptr<const FieldManager<Real>> m_device_field_manager;
+  std::shared_ptr<const FieldManager<Real>>    m_device_field_manager;
   std::shared_ptr<const GridsManager>          m_grids_manager;
 
   bool                                 param_set = false;
   bool                                 gm_set    = false;
-  bool                                 fm_set  = false;
+  bool                                 fm_set    = false;
   bool                        m_runtype_restart  = false;
 
 }; // class OutputManager
@@ -126,14 +126,18 @@ protected:
  * stream.  See scorpio_output.hpp for more information on what the parameter list needs.         */
 inline void OutputManager::new_output(const ekat::ParameterList& params)
 {
-  auto output_instance = std::make_shared<output_type>(pio_comm,params,m_device_field_manager,m_grids_manager,m_runtype_restart);
+  auto grid_name = params.get<std::string>("GRID");
+  auto grid = m_grids_manager->get_grid(grid_name);
+  auto output_instance = std::make_shared<output_type>(pio_comm,params,m_device_field_manager,grid,m_runtype_restart);
   output_instance->init();
   m_output_streams.push_back(output_instance);
 }
 /* --------------------------------------------------------------------- */
 inline void OutputManager::new_output(const ekat::ParameterList& params, const bool runtype_restart)
 {
-  auto output_instance = std::make_shared<output_type>(pio_comm,params,m_device_field_manager,m_grids_manager,runtype_restart);
+  auto grid_name = params.get<std::string>("GRID");
+  auto grid = m_grids_manager->get_grid(grid_name);
+  auto output_instance = std::make_shared<output_type>(pio_comm,params,m_device_field_manager,grid,runtype_restart);
   output_instance->init();
   m_output_streams.push_back(output_instance);
 }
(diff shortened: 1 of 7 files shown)


components/scream/src/share/io/output_manager.hpp
```diff
@@ -110,12 +110,12 @@ class OutputManager
   ekat::Comm                                   atm_comm;
   ekat::Comm                                   pio_comm; 
   ekat::ParameterList                          m_params;
-  std::shared_ptr<const FieldManager<Real>> m_device_field_manager;
+  std::shared_ptr<const FieldManager<Real>>    m_device_field_manager;
```
bartgol, line 113: MInor: there is no Device vs Host field manager. Each field contains both a host and a device view. The name 'device_field_manager' gives the false feeling that it has only device stuff, or that there is such a thing as a 'host_field_manager'. I would just call it `m_field_manager`.

components/scream/src/share/io/output_manager.hpp
```diff
@@ -110,12 +110,12 @@ class OutputManager
...
   std::shared_ptr<const GridsManager>          m_grids_manager;
 
   bool                                 param_set = false;
   bool                                 gm_set    = false;
-  bool                                 fm_set  = false;
+  bool                                 fm_set    = false;
```
bartgol, line 118: Minor: these bools are kind of redundant. Since gm and fm are pointers, one can simply check if the pointers are null. E.g., replace `if (gm_set)` with `if (m_grids_manager)`.

components/scream/src/share/io/scorpio_input.cpp
```diff
@@ -18,7 +18,7 @@ void AtmosphereInput::
 pull_input(const std::string& filename, const std::string& var_name,
            const std::vector<std::string>& var_dims,
            const bool has_columns, const std::vector<int>& dim_lens,
-           const int padding, Real* data)
+           const int padding, const grid_ptr_type& grid, Real* data)
```
bartgol, line 21: It appears this input is never used. Do we need it?

components/scream/src/share/io/scorpio_input.hpp
```diff
@@ -158,6 +187,8 @@ class AtmosphereInput
 
   std::shared_ptr<const FieldManager<Real>>   m_field_mgr;
   std::shared_ptr<const GridsManager>         m_grid_mgr;
+  grid_ptr_type                               m_grid;
+  bool                                        m_grid_set = false;
```
bartgol, line 191: Same comment as for the output  manager class: the bool is superfluous, and can be replaced with a direct check on the ptr.

components/scream/src/share/io/output_manager.hpp
bartgol: To give more context, checking the pointer directly does not require one to make sure that the bool and the pointer are "in sync" (a minor bug that can creep in during refactors, or if another method is added).

components/scream/src/share/io/scorpio_output.cpp
```diff
@@ -13,7 +13,7 @@ void AtmosphereOutput::init()
   // Parse the parameters that controls this output instance.
   // See the comments at the top for more details.
   m_casename        = m_params.get<std::string>("FILENAME");
-  m_grid_name       = m_params.get<std::string>("GRID","Physics");  // optional, default to Physics.
+  m_grid_name       = m_params.get<std::string>("GRID","Physics");  // optional, default to Physics., TODO: I don't think this is used anymore.
```
bartgol, line 16: Then let's remove it. If the code stops compiling or starts crashing, you can add it back. In general, let's aim at keeping classes lean. If something is suspected to be unused, let's remove it. It's relatively easy to add small things like this back, if need be.

components/scream/src/share/io/scorpio_output.hpp
```diff
@@ -140,6 +167,8 @@ class AtmosphereOutput
   ekat::Comm                                  m_comm;
   std::shared_ptr<const FieldManager<Real>>   m_field_mgr;
   std::shared_ptr<const GridsManager>         m_grid_mgr;
+  grid_ptr_type                               m_grid;
+  bool                                        m_grid_set = false;
```
bartgol, line 171: Here too: consider removing this bool, and simply query the pointer.

components/scream/src/share/io/scorpio_output.hpp
```diff
@@ -140,6 +167,8 @@ class AtmosphereOutput
   ekat::Comm                                  m_comm;
   std::shared_ptr<const FieldManager<Real>>   m_field_mgr;
   std::shared_ptr<const GridsManager>         m_grid_mgr;
+  grid_ptr_type                               m_grid;
```
bartgol, line 170: I'm confused by seeing both a grids manager and a grid. If you need just one grid, then there's no need to keep the grids manager around. OTOH, if you need more than one grid, then what's `m_grid`?

AaronDonahue:
Closing, after consultation with @bartgol we are going to go a different route.

PR #1155, opened 2021-08-10
Introduce a restart test that runs actual physics through the AD.

AaronDonahue:
This commit establishes the test stored in /tests/coupled/ but it is not
complete yet.  Currently the test mimics the typical shoc_cld_p3_rrtmgp
test but also includes output and the writing of restart files.

This commit also forces all fields that are "required" to be added
to the "restart" group so that they are automatically written to
the restart file.


AaronDonahue:
This should be on hold until #1162 goes in.

bartgol:
Closing, in favor of #1251

PR #1152, opened 2021-08-09
[WIP] Update mct-coupling code

tcclevenger:
This updates the remaining `mct-coupling` code require to get CIME to build. Answers remaining comments from https://github.com/E3SM-Project/scream/pull/1128.

PR #1156, opened 2021-08-10
mct-coupling changes

tcclevenger:
This updates the remaining mct-coupling code require to get CIME to build and begin to run. Answers remaining comments from #1128.

@bartgol I think it would be nice to have these mct-coupling changes in master. This is stripped of any CIME related config code. The difference in this PR and https://github.com/E3SM-Project/scream/pull/1152 is that now the code runs up until the DIRK failure (which I do not suspect is because of SurfaceCoupling).

Main additions:
1. YAKL must be initialized as in the rrtmgp tests
2. I created a `vec_comp_a2x/x2a` in `scream_cpl_indices.F90` so that I can pass exact field name "horiz_winds" through the coupler.
3. The new RRTMGP fields are now considered in mct-coupling.

### DIFF
--- a/components/scream/src/mct_coupling/scream_cpl_indices.F90
+++ b/components/scream/src/mct_coupling/scream_cpl_indices.F90
@@ -6,13 +6,13 @@ module scream_cpl_indices
   private
 
   ! Focus only on the ones that scream imports/exports (subsets of x2a and a2x)
-  integer, parameter, public :: num_required_imports = 21
-  integer, parameter, public :: num_scream_imports   = 4
-  integer, parameter, public :: num_required_exports = 12
-  integer, parameter, public :: num_optional_imports = 0
-  integer, parameter, public :: num_optional_exports = 1
-  integer, parameter, public :: num_imports = num_required_imports + num_optional_imports
-  integer, parameter, public :: num_exports = num_required_exports + num_optional_exports
+  integer, parameter, public :: num_required_cpl_imports = 21
+  integer, parameter, public :: num_scream_imports       = 8
+  integer, parameter, public :: num_required_exports     = 12
+  integer, parameter, public :: num_optional_cpl_imports = 0
+  integer, parameter, public :: num_optional_exports     = 1
+  integer, parameter, public :: num_cpl_imports = num_required_cpl_imports + num_optional_cpl_imports
+  integer, parameter, public :: num_exports     = num_required_exports + num_optional_exports
 
   integer(kind=c_int), public, allocatable, target :: index_x2a(:)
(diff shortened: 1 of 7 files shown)


components/scream/src/mct_coupling/scream_cpl_indices.F90
```diff
@@ -117,6 +125,11 @@ subroutine scream_set_cpl_indices (x2a, a2x)
     scr_names_x2a(20) = 'unused'
     scr_names_x2a(21) = 'unused'
 
+    ! No scream imports currently require vector component info. Set to -1.
```
bartgol, line 128: Side question: I saw `surf_u_mom_flux` and `surf_v_mom_flux`, and thought "why, aren't those two components of a vector?" So I grepped for them, and found that they are two distinct fields in SHOC. Since you are a shoc guru too, do you think they can be fused into a vector field in SHOC?

components/scream/src/mct_coupling/scream_cxx_f90_interface.cpp
```diff
@@ -80,6 +81,9 @@ void scream_create_atm_instance (const MPI_Fint& f_comm,
     // First of all, initialize the scream session
     scream::initialize_scream_session();
 
+    // Initialize yakl
+    if(!yakl::isInitialized()) { yakl::init(); }
```
bartgol, line 85: This might be personal, but I don't see why the cxx-f90 interface should do this explicitly. Imho, it should be done in one of the following places:

- inside `initialize_scream_session`: this is where also kokkos gets (indirectly) inited.
- inside rrtmgp interface (or any place using yakl): since rrtmgp is probably the only yakl customer (at the very least, they are not many), and since yakl offers the `isInitialized()` guard, maybe it should be up to rrtmgp (and possibly other yakl customers) to check/init yakl.

I guess the latter is more of a totalitarian kokkos regime approach, while the former is more of a peaceful coexistence and equal treatment approach. So I'm fine with the former, I guess.

bartgol: Question: does yakl need to be finalized as well?

tcclevenger: I agree it should be moved, and I think I like the Kooks-totalitarian approach of putting it into the interface since it is the only place that uses yakl..

components/scream/src/mct_coupling/scream_cpl_indices.F90
tcclevenger: Absolutely.

bartgol (approved):

PR #1158, opened 2021-08-11
Move YAKL init and finalize

tcclevenger:
Move `yakl::init` and `yakl::finalize` into the rrtmgp interface (calling `init()` only if it has not been initialized previously).

### DIFF
--- a/components/scream/src/mct_coupling/scream_cxx_f90_interface.cpp
+++ b/components/scream/src/mct_coupling/scream_cxx_f90_interface.cpp
@@ -81,9 +81,6 @@ void scream_create_atm_instance (const MPI_Fint& f_comm,
     // First of all, initialize the scream session
     scream::initialize_scream_session();
 
-    // Initialize yakl
-    if(!yakl::isInitialized()) { yakl::init(); }
-
     // Create the context
     auto& c = ScreamContext::singleton();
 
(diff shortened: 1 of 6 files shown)


tcclevenger:
@brhillman @bartgol Maybe move finalizing yakl into `rrtmgp::rrtmgp_finalize`? The issue is that customers may want to use yakl after they use rrtmgp.

So maybe the solution should just be to leave the AD the same as it was, and for mct-coupling, add it to `finalize_scream_session`?

tcclevenger:
> If there is a chance that someone else uses yakl in e3sm, that might still be a problem. To be fair, though, the same could be said for kokkos. So at this point I would say let's move it in `finalize_scream_session`. At some point we need to think about the fact that kokkos initialization _might_ have to be handled directly by e3sm.

What about the same for `initialize_scream_session`?

bartgol:
> > If there is a chance that someone else uses yakl in e3sm, that might still be a problem. To be fair, though, the same could be said for kokkos. So at this point I would say let's move it in `finalize_scream_session`. At some point we need to think about the fact that kokkos initialization _might_ have to be handled directly by e3sm.
> 
> What about the same for `initialize_scream_session`?

You mean move yakl init inside `initialize_scream_session`? I'm ok with that.

PeterCaldwell:
I'm not sure if this is relevant or too late to make a difference, but I'm unclear whether YAKL will ever be used for anything outside of RRTMGP. I think it would be worth asking Matt Norman and Ben directly before getting wrapped around the spindle about how to support it outside RRTMGP... I can ping them via slack if you want, but it might make more sense for one of you to do it since you have a more detailed understanding of the situation...

tcclevenger:
> I'm not sure if this is relevant or too late to make a difference, but I'm unclear whether YAKL will ever be used for anything outside of RRTMGP. I think it would be worth asking Matt Norman and Ben directly before getting wrapped around the spindle about how to support it outside RRTMGP... I can ping them via slack if you want, but it might make more sense for one of you to do it since you have a more detailed understanding of the situation...

Definitely not too late. I'll ping in slack.

tcclevenger:
To summarize, for now Ben and Matt are OK with this being added to the radiation interface.

bartgol (approved):

PR #1162, opened 2021-08-13
Refactor scorpio interfaces and classes

bartgol:


bartgol:
I think this is a good point to stop, and integrate these mods, before moving forward. @AaronDonahue @jeff-cohere , here's a summary of the mods:

- Fix a bug related to how FieldHeader and FieldTracking were tracking parent/children (coming from subfields). The bug was subtle: basically, [`shared_from_this`](https://en.cppreference.com/w/cpp/memory/enable_shared_from_this/shared_from_this) cannot be called from the ctor, since the shared_ptr control block is not yet built, so the returned shared ptr is always null. While fixing this (basically, do the parent-children link _after_ the ctor returns), I thought I would abstract this parent-child link setup/cleanup with a common class, since the logic was the same in both header and tracking classes.
Note: FieldTracking needs the parent/children link _only_ to be able to update the children fields time stamps whenever the parent time stamp is updated. If we think this feature is not useful, we can simplify the FieldTracking class.
- all scorpio I/O classes are now on a single grid. We currently don't need 2+ grids but will soon. The idea will be to have 1 output manager per grid. We will always have very few grids (2 or maybe 3), so this is probably ok. 
- scorpio input: reworked a bit its impl. Can work with a FieldManager (grabs everything from there) or with a vector of 1d views and one of field layouts.
- scorpio output: more rework than for input. Unlike the input class, it only works with FieldManager (we might want to extend to save raw view+layout, perhaps for geometry data). Removed the m_status map, since the only counter really useful is the number of steps since the last output (and only if avg!=Instant). I started to introduce the word "checkpoint" or "output checkpoint" for output history restart files, but I don't think I made it super consistent throughout. Let me know what you think of the wording, and how it can be improved.

I already have a few things in mind for what's next with scorpio (e.g., output (and restart!) on dyn grid, or output non-field quantities like lat/lon). We can discuss those at the telecon.

components/scream/src/control/atmosphere_driver.cpp
```diff
@@ -215,20 +217,28 @@ void AtmosphereDriver::create_fields()
...
-  if (m_atm_params.isSublist("Output Manager")) {
-    auto& out_params = m_atm_params.sublist("Output Manager");
-    m_output_manager.set_params(out_params);
-    m_output_manager.set_comm(m_atm_comm);
-    m_output_manager.set_grids(m_grids_manager);
-    m_output_manager.set_field_mgr(get_ref_grid_field_mgr());
+  // For each grid in the grids manager, grab a homonymous sublist of the
+  // "Output Managers" atm params sublist (might be empty), and create
+  // an output manager for that grid.
+  // in the parameter list with that name. If so, create an output
```
AaronDonahue, line 226: This comment reads as though something is missing.

components/scream/src/control/atmosphere_driver.cpp
```diff
@@ -215,20 +217,28 @@ void AtmosphereDriver::create_fields()
...
+  // manager for that grid
+  auto& out_mgrs_params = m_atm_params.sublist("SCREAM Output");
+  for (auto it : m_grids_manager->get_repo()) {
+    auto& out_params = out_mgrs_params.sublist(it.first);
+    auto fm = m_field_mgrs.at(it.first);
+    // The restart params are the same for all grid, so they are specified
```
AaronDonahue, line 232: nitpicky: grid->grids

components/scream/src/control/atmosphere_driver.cpp
```diff
@@ -296,29 +306,29 @@ initialize_fields (const util::TimeStamp& t0)
...
+  ic_reader_params.set("FIELDS",ic_fields_names);
 
   // Check the status of lat and lon values on reference grid:
   // This option allows the user to set lat or lon in their own
   // test or run setup code rather than by file.
   bool skip_init_lat = true;
   bool skip_init_lon = true;
   if (ic_pl.isParameter("skip_init_lat")) {
-    skip_init_lat     = ic_pl.get<bool>("skip_init_lat");
+    skip_init_lat = ic_pl.get<bool>("skip_init_lat");
```
jeff-cohere, line 323: Is `skip_init_{lat/lon}` a legacy name? It's less confusing to use terms that indicate things should be done rather than skipped. Skipping things implies assumptions that everyone knows. But if this name is historical, I guess we're stuck with it.

components/scream/src/control/atmosphere_driver.hpp
```diff
@@ -85,7 +85,8 @@ class AtmosphereDriver
   // which is handy for scream standalone tests.
   void initialize (const ekat::Comm& atm_comm,
                    const ekat::ParameterList& params,
-                   const util::TimeStamp& t0);
+                   const util::TimeStamp& t0,
+                   const bool restarted_run = false);
```
jeff-cohere, line 89: Can we add some comments above this method that describe the parameters, since we're adding a new one? One can guess at what they mean, but we spend so much time guessing at things this way already. :-)

components/scream/src/share/io/scorpio_output.hpp
```diff
@@ -81,113 +60,124 @@
...
+  int m_nsteps_since_last_output = 0;
 
-}; // Class AtmosphereOutput
+  // When this equals m_checkpoint_freq, it's time to write a history restart file.
+  int m_nsteps_since_last_checkpoint = 0;
 
-//// ====================== IMPLEMENTATION ===================== //
-inline void AtmosphereOutput::check_status()
+  // When this equals m_out_max_steps, it's time to close the out file, and open a new one.
+  int m_num_snapshots_in_file = 0;
```
AaronDonahue, line 162: I like these new names, they make more sense.

components/scream/src/share/io/scorpio_output.hpp
```diff
@@ -81,113 +60,124 @@
...
+};
+
+inline void AtmosphereOutput::combine (const Real& new_val, Real& curr_val) const
```
AaronDonahue, line 165: We should add a comment to this function that explains that it controls updating variables based on the output type and is where any new output types would go.  For example if someone wanted to add standard deviation as an option, or something like that.

components/scream/src/share/util/scream_family_tracking.hpp
```diff
@@ -0,0 +1,170 @@
+#ifndef SCREAM_FAMILY_TRACKING_CLASS
+#define SCREAM_FAMILY_TRACKING_CLASS
+
+#include "ekat/std_meta//ekat_std_enable_shared_from_this.hpp"
```
jeff-cohere, line 4: You've got a double slash here.

components/scream/src/share/io/scream_output_manager.cpp
```diff
@@ -0,0 +1,123 @@
...
+  auto output = std::make_shared<output_type>(m_io_comm,params,m_field_mgr,
+                                              m_runtype_restart, model_restart_output);
+  m_output_streams.push_back(output);
+}
+/*===============================================================================================*/
+void OutputManager::make_restart_param_list(ekat::ParameterList& params)
+{
+  // Given the unique nature of restart files, this routine sets up the specific requirements.
+
+  //TODO change this to some sort of case_name, TODO: control so suffix is .r.nc
```
AaronDonahue, line 94: Probably a bigger TODO which is on the list of SCORPIO tasks is to create a generic case name for all runs that will be used for the output file names.  For example in EAM I think the pattern for all outputs is to have the name MACHINE.RESOLUTION.COMPSET.PROCESS.DATE.nc (or something along those lines)

components/scream/src/share/util/scream_family_tracking.hpp
```diff
@@ -0,0 +1,170 @@
...
+
+#include <list>
+#include <memory>
+#include <type_traits>
+
+namespace scream
+{
+
+/*
+ * Class that handles tracking of parent/children instances of the same class type
```
jeff-cohere, line 15: Neat. Unless we want to introduce a "family" as a first-class concept, maybe we could call this `TypeTracking` or `TypeBasedTracking`?

components/scream/src/share/util/scream_family_tracking.hpp
```diff
@@ -0,0 +1,170 @@
...
+// =================== IMPLEMENTATION ====================== //
+
+template<typename DerivedType>
+FamilyTracking<DerivedType>::FamilyTracking ()
+{
+  // Note: we cannot put the static assert in the class decl, cause DerivedType
+  //       is still incomplete at that point.
+  static_assert (std::is_base_of<tracking_type,derived_type>::value,
+      "Error! Do not try to instantiate FamilyTracking<T> if T does not inherit from FamilyTracking.\n"
+      "       This class is trying to use the Curiously Recurring Template Pattern (CRTP).\n");
```
jeff-cohere, line 75: Do or do not. There is no try (with the exception of exceptions).

components/scream/src/share/io/scream_output_manager.cpp
```diff
@@ -0,0 +1,123 @@
...
+  params.set<std::string>("AVERAGING TYPE","Instant");
+
+  if (not params.isParameter("GRID")) {
+    params.set<std::string>("GRID",m_field_mgr->get_grid()->name());
+  }
+  auto& freq_params = params.sublist("FREQUENCY");
+  freq_params.get<Int>("OUT_MAX_STEPS",1);
+
+  // Grab the restart fields from the field manager.
+  // If a developer wants a field to be stored in the restart file than they must add the "restart" group tag
```
AaronDonahue, line 108: Is there a way we can use the DAG or the actions during the DAG to add this label by default to all "required" fields?  I have a hacky way to do that in a separate WIP PR, see https://github.com/E3SM-Project/scream/pull/1155/files#diff-ffc785b3f4b9ce9afd2510be2ae5df1eaa5f92d3b18d0a7ca67f0b7e70aea5a2R132
But I'm sure this can be done better.

components/scream/src/share/io/scream_scorpio_interface.F90
```diff
@@ -1018,23 +1086,38 @@ end subroutine get_new_pio_atm_file
...
+    type(hist_var_t), pointer          :: var
+    integer                            :: ierr,var_size,ndims,i
+    logical                            :: found
 
     call lookup_pio_atm_file(trim(filename),pio_atm_file,found)
     call get_var(pio_atm_file,varname,var)
+
+    ! We don't want the extent along the 'time' dimension
+    if (var%has_t_dim) then
+      ndims = SIZE(var%dimlen)-1
```
AaronDonahue, line 1107: The if and else are the same statements here.

components/scream/src/share/io/scream_scorpio_interface.F90
```diff
@@ -1054,23 +1137,39 @@ end subroutine grid_write_darray_1d_real
...
+    logical                            :: found
+    real(rtype), dimension(:), pointer :: var_data
 
     call lookup_pio_atm_file(trim(filename),pio_atm_file,found)
     call get_var(pio_atm_file,varname,var)
+    
+    ! We don't want the extent along the 'time' dimension
+    if (var%has_t_dim) then
+      ndims = SIZE(var%dimlen)-1
+    else
```
AaronDonahue, line 1160: Same here, copy/paste

components/scream/src/control/atmosphere_driver.cpp
bartgol: I agree, I'll swap that around.

components/scream/src/control/atmosphere_driver.cpp
bartgol: Ah, yes, I change my mind when writing it, and left chunks of the first draft around. Will clean up.

components/scream/src/control/atmosphere_driver.hpp
bartgol: Good point.

components/scream/src/share/io/tests/io.cpp
```diff
@@ -34,6 +38,77 @@ std::shared_ptr<UserProvidedGridsManager>   get_test_gm(const ekat::Comm& io_com
 ekat::ParameterList                         get_om_params(const Int casenum, const ekat::Comm& comm);
 ekat::ParameterList                         get_in_params(const std::string type, const ekat::Comm& comm);
 
+// TEST_CASE("ciccio")
```
AaronDonahue, line 41: This is commented out, should we delete, or is this a TODO for the next PR?

components/scream/src/share/field/field_header.hpp
```diff
@@ -98,16 +98,11 @@ class FieldHeader : public ekat::enable_shared_from_this<FieldHeader> {
...
-  // OTOH, if other fields are sub-view of this field, we keep a pointer to them
-  std::weak_ptr<FieldHeader>              m_parent;
-  std::list<std::weak_ptr<FieldHeader>>   m_children;
-
   // Extra data associated with this field
   extra_data_type                 m_extra_data;
 };
 
-// Use this free function to exploit features of enable_from_this
+// Use this free functions to exploit features of enable_from_this
```
jeff-cohere, line 105: Does this comment apply to the two functions below?

components/scream/src/share/util/scream_family_tracking.hpp
bartgol: Argh, I saw that and wanted to get rid of it, then I forgot. You've got a 25/20 vision, sir.

components/scream/src/share/field/field_manager.hpp
```diff
@@ -382,7 +382,7 @@ registration_ends ()
   //     we must be able to allocate C bundled.
   //  5) For each cluster, call the function contiguous_superset from scream_utils.hpp
   //     (see that file for details). If the fcn fails to find an ordering of the cluster's
-  //     field that accommodate all bundler request, it will return an empty list.
+  //     field that accommodate all bundled request, it will return an empty list.
```
jeff-cohere, line 385: bundled requests

components/scream/tests/uncoupled/p3/p3_standalone_output.yaml
```diff
@@ -7,28 +7,6 @@ FREQUENCY:
...
-#  field 14: qi
-#  field 15: qm
-#  field 16: qr
-#  field 17: qv
-#  field 7: nevapr - deprecated
-#  field 7: qr_evap_tend - deprecated
-#  field 11: precip_total_tend - deprecated
-#  field 4: lambda_qc - deprecated
-#  field 6: mu_qc - deprecated
+FIELDS: [T_mid, T_prev_micro_step, qv_prev_micro_step, eff_radius_qc, eff_radius_qi,
```
AaronDonahue, line 10: We can add qv, qc, ... to this output file, now that you've fixed outputting tracers.

components/scream/src/share/io/scorpio_input.hpp
```diff
@@ -33,43 +25,34 @@
...
  *  -----
  *  Input Parameters
  *    FILENAME: STRING
  *    GRID: STRING
- *    FIELDS
- *      Number of Fields: INT
- *      field_1: STRING
- *      ...
- *      field_N: STRING
+ *    FIELDS: [field_name_1, field_name_2, ..., field_name_N]
```
jeff-cohere, line 42: If people would rather have one field name per line (easier to insert/delete fields), one can use hyphens:

```
FIELDS:
    - field_name_1
    - field_name_2
    ...
    - field_name_N
```

components/scream/src/share/io/scorpio_input.hpp
```diff
@@ -33,43 +25,34 @@
  *  this class to facilitate cases where reading input over some number of simulation
  *  timesteps is possible.
  *
- *  At construction time All input instances require
+ *  At construction time input instances require
```
jeff-cohere, line 28: You might stick a comma after `time` to make this read easier.

components/scream/src/share/io/scorpio_input.hpp
```diff
@@ -80,94 +63,74 @@ namespace scream
...
-                   const std::shared_ptr<const GridsManager>& grid_mgr)
-    : m_comm      (comm)
-    , m_grid_mgr  (grid_mgr)
-    , m_grid_name (grid_name)
-  {
-    // Nothing to do here
-  }
+  AtmosphereInput (const ekat::Comm& comm,
+                   const ekat::ParameterList& params);
+  AtmosphereInput (const ekat::Comm& comm,
```
jeff-cohere, line 77: Do any of these constructors need explanation?

components/scream/src/share/util/scream_family_tracking.hpp
bartgol: I struggled with the name. I used 'family' only cause the class stores "parent" and "children", so I kept the analogy rolling. But a better name might help.

What about `PointerGraphTracking` or something with the word "graph" or "tree" in it?

bartgol: Technically, the graph might not be a tree, since I do not check that my parent is not me, nor one of my children (or grandchildren or...). In scream, the use case is quite limited, so I didn't feel the need for this check (though if used with circular deps, there would def be bugs). If we moved this file to ekat (which might be a possibility, given that it has no scream-based dep), then I might add checks.

components/scream/src/share/io/scream_output_manager.cpp
bartgol: Yes, I wanted to get rid of this comment, and add code in the AD that deduces the restart group. I didn't get to that. If you want, I can do it here, otherwise we can do a separate PR (it's a very self-contained task, so I'd vote for PR, and assign to whomever got time).

components/scream/src/share/io/scream_scorpio_interface.F90
bartgol: Ctrl+C - Ctrl+V

components/scream/src/share/io/scream_output_manager.cpp
AaronDonahue: We can keep it in PR 1155 since it's one of the first places it will be used explicitly and then when it's ready to go you can review if there is a better way then what I did.

components/scream/src/share/io/tests/io.cpp
bartgol: Oh, lol! I added that while debugging, and completely forgot. I am thankful I didn't follow my usual habit of filling debug messages/vars/names with "colorful" italian words... :D

components/scream/src/share/io/scream_scorpio_interface.F90
AaronDonahue: or for us VI users "yy" - "p"

components/scream/src/share/field/field_header.hpp
bartgol: Ah, well, yes and no. You do need those two fcns to trigger `enable_shared_from_this` extra capabilities. But the second should be the only way to create a subfield header...except that I left the subfield header ctor declared (but never defined!).

I will revert back to singular, and make a separate comment for the 2nd fcn.

components/scream/src/share/io/tests/io.cpp
AaronDonahue: Your debug messages are helping me prepare for the next time I watch soccer with my in-laws...

components/scream/src/share/field/field_manager.hpp
bartgol: Ah, dang. I felt a bit like a "Jeff" for catching these typos, but I clearly can't outperform the master. Good catch!

components/scream/tests/uncoupled/p3/p3_standalone_output.yaml
bartgol: Yes, that might add a bit of testing to the IO infrastructure.

components/scream/src/share/io/scorpio_input.hpp
bartgol: Does that still map to a sequence, like the syntax I used in the comment? Nice!

components/scream/src/share/io/scorpio_input.hpp
bartgol: Probably. I will add some.

components/scream/src/share/io/scorpio_input.hpp
jeff-cohere: Yup. It's alternate syntax for a YAML sequence.

components/scream/src/share/util/scream_family_tracking.hpp
jeff-cohere: If it should be a tree to avoid circular refs, it seems like this check might be useful. And it may have a home in EKAT--the curiously recurring template pattern is not too bad a tax to pay for this kind of tracking.

components/scream/src/share/io/scream_output_manager.cpp
bartgol: Let's add it to the TODO list then. It should be item 817191, right? =)

Jokes aside, a good v1.1 issue.

components/scream/src/share/io/scream_output_manager.hpp
```diff
@@ -0,0 +1,92 @@
...
+
+  void setup (const ekat::Comm& io_comm, const ekat::ParameterList& params,
+              const std::shared_ptr<const fm_type>& field_mgr,
+              const bool runtype_restart);
+  void run(util::TimeStamp& current_ts);
+  void finalize();
+
+protected:
+  // Add an output stream.
+  void new_output(const ekat::ParameterList& params, const bool model_restart_output);
```
jeff-cohere, line 72: Consider `add_output` for a "verbier" method name.

components/scream/src/share/io/scorpio_output.cpp
```diff
@@ -1,102 +1,143 @@
...
     rpointer_file.open("rpointer.atm");
     std::string filename;
     bool found = false;
-    std::string testname = m_casename+"."+m_avg_type+"."+m_out_units+"_x"+std::to_string(m_out_frequency);
-    while (rpointer_file >> filename)
-    {
-      if (filename.find(testname) != std::string::npos)
-      {
+    std::string testname;
+    if (m_params.isParameter("RESTART HISTORY FILENAME")) {
```
jeff-cohere, line 96: Are all these parameter names written down somewhere?

components/scream/src/share/io/scorpio_output.cpp
```diff
@@ -1,102 +1,143 @@
...
+    // If the history restart file is not found, we must HOPE that it is because
+    // the last model restart step coincided with a model output step, in which case
+    // a restart history file is not written.
+    // TODO We have NO WAY of checking this from this class, but the OutputManager
+    //      *might* be able to figure out if this is the case, and perhaps set
+    //      "Restart History" to false in the input parameter list.
+    //      For now, simply print a warning.
+    if (found) {
+      // If restart data is needed, the history restart file *must* be found.
+      EKAT_REQUIRE_MSG(found,
```
jeff-cohere, line 118: `found` is always `true` in this context.

components/scream/src/share/io/scorpio_output.cpp
bartgol: I doubt it. I will add some docs in the output header.

components/scream/src/share/util/scream_family_tracking.hpp
bartgol: Ok. I might do this in a second time, cause I don't want to delay this PR too much, if that's ok. I will work on it right after it gets merged though.

components/scream/src/share/io/scorpio_output.cpp
bartgol: Good catch.

bartgol:
@AaronDonahue @jeff-cohere I think I addressed most of your comments. I tried to add/improve the documentation, explaining the meaning of fcn args as well as input plist entries.

Would you mind taking another look?

components/scream/tests/uncoupled/homme/CMakeLists.txt
```diff
@@ -6,7 +6,7 @@ CreateDynamicsLib("theta-l_kokkos"  4   72   35)
 
 set (NEED_LIBS ${dynLibName} scream_control scream_share)
 
-# Test Homme standalone
+# Test Homme stand alone
```
jeff-cohere, line 9: Maybe okay to leave as "standalone". Standalone (or stand-alone) is an adjective, like "standalone HOMME", whereas to "stand alone" is a verb phrase.

Not a big deal, though.

jeff-cohere (approved):
Looks good to me!

components/scream/tests/uncoupled/homme/CMakeLists.txt
bartgol: I was just trying to make it so that we use the same version. There was some "standalone" and some "stand_alone". I tossed a coin, and picked one.

components/scream/tests/uncoupled/p3/p3_stand_alone_output.yaml
```diff
@@ -0,0 +1,12 @@
+%YAML 1.1
+---
+Casename: p3_stand_alone_output
+Averaging Type: Instant
+Grid: Physics
+Max Snapshots Per File: 1
+Fields: [T_mid, T_prev_micro_step, qv, qc, qr, qv_prev_micro_step, eff_radius_qc, eff_radius_qi,
```
AaronDonahue, line 7: nitpicky, and not necessary, but this might be a good place to use the `-` example for making long lists of fields.

AaronDonahue (approved):
Looks good!  Thank you @bartgol

components/scream/tests/uncoupled/p3/p3_stand_alone_output.yaml
bartgol: Done

components/scream/tests/uncoupled/homme/CMakeLists.txt
bartgol: Done

bartgol:
Adding WIP till I figure out why the test randomly hangs. Meanwhile, I found some bugs that got me wondering how is master passing. E.g., see bug fix in 1e0a082. The dofs offset we computed were completely wrong. How does pio not crash in master?

components/scream/tests/uncoupled/p3/p3_standalone_output.yaml
```diff
@@ -1,34 +1,22 @@
...
-#  field 7: nevapr - deprecated
-#  field 7: qr_evap_tend - deprecated
-#  field 11: precip_total_tend - deprecated
-#  field 4: lambda_qc - deprecated
-#  field 6: mu_qc - deprecated
+Casename: p3_standalone_output
+Averaging Type: Instant
+Grid: Physics
+Max Snapshots Per File: 1
+Fields:
```
jeff-cohere, line 7: Looks great!

components/scream/src/share/io/scream_scorpio_interface.F90
```diff
@@ -249,80 +251,108 @@ subroutine eam_pio_enddef(filename)
...
-    allocate(curr%coord)
-    hist_coord => curr%coord
-    pio_atm_file%DimCounter = pio_atm_file%DimCounter + 1  ! AaronDonahue - Is this even still needed?
-    ! Register this dimension
-    hist_coord%name      = trim(shortname)
-    hist_coord%long_name = trim(longname)
-    hist_coord%dimsize   = length
-    if (length.eq.0) then
-      ierr = PIO_def_dim(pio_atm_file%pioFileDesc, trim(shortname), pio_unlimited , hist_coord%dimid)
+          ! call errorHandle("PIO Error: Could not register dimension"//trim(shortname)//", already exists in file: "//trim(filename),-999)
```
AaronDonahue, line 322: Should we remove this comment?

components/scream/src/share/io/scream_scorpio_interface.F90
```diff
@@ -249,80 +251,108 @@ subroutine eam_pio_enddef(filename)
...
+    ! If the dim was not found, create it
+    if (.not. dim_found) then
+      allocate(prev%next)
+      curr => prev%next
+      allocate(curr%coord)
+      hist_coord => curr%coord
+      pio_atm_file%DimCounter = pio_atm_file%DimCounter + 1  ! AaronDonahue - Is this even still needed?
```
AaronDonahue, line 329: If DimCounter isn't used currently then I would say no, it isn't needed.

components/scream/src/share/io/scream_scorpio_interface.F90
```diff
@@ -351,53 +381,79 @@ subroutine get_variable(pio_atm_filename,shortname,longname,numdims,var_dimensio
...
     curr => pio_atm_file%var_list_top
+
     do while (associated(curr))
       if (associated(curr%var)) then
-        if (trim(curr%var%name)==trim(shortname)) call errorHandle("PIO Error: Could not register variable "//trim(shortname)//", already exists in file: "//trim(pio_atm_filename),-999)
-      end if
+        if (trim(curr%var%name)==trim(shortname)) then
+          var_found = .true.
+          exit
+        ! if (trim(curr%var%name)==trim(shortname)) call errorHandle("PIO Error: Could not register variable "//trim(shortname)//", already exists in file: "//trim(filename),-999)
```
AaronDonahue, line 405: Again, should we delete this comment?

jeff-cohere (approved):
Looks great, @bartgol.

AaronDonahue (approved):

AaronDonahue:
I approved the PR, looks good, and is a ton of work!  I had a few comments, but they shouldn't hold up merging this PR.

I did have a follow-up question, does this PR address the discrepancy we noticed regarding the DOF indices passed to PIO?  I'm guessing that will be the subject of a later PR?

AaronDonahue (approved):

PR #1163, opened 2021-08-17
Update machine specs for LC machine syrah to avoid timing out during nightly tests

AaronDonahue:


AaronDonahue:
Merging this manually, it's just a machine file change that impacts LC only.

PR #1164, opened 2021-08-18
Move conversion of nc_prescribed out of p3_main code and into the interface

AaronDonahue:
Previously, the nc_prescribed values passed to p3_main, gathered from SPA,
had units of #/cm3.  These needed to be converted to the units of nc, #/kg.

This was done inside of the p3_main routine, which could cause confusion in
the future if we adopt a different aerosol scheme to SPA.  Instead, the
unit conversion is now done in the p3 interface and explicitly done when
SPA is active.

This commit also introduces non-zero values for nccn_prescribed in the
p3_run_and_cmp tests.  In testing the changes in this test it was discovered
there wasn't any nccn_prescribed coverage for the p3_run_and_cmp tests.

[BFB]

### DIFF
--- a/components/eam/src/physics/cam/micro_p3_interface.F90
+++ b/components/eam/src/physics/cam/micro_p3_interface.F90
@@ -744,7 +744,8 @@ end subroutine micro_p3_readnl
                               rho_h2os, &
                               qsmall, &
                               mincld, & 
-                              inv_cp 
+                              inv_cp, &
+                              grav=g 
 
     !INPUT/OUTPUT VARIABLES
     type(physics_state),         intent(in)    :: state
@@ -1046,6 +1047,15 @@ end subroutine micro_p3_readnl
        ! ccn is uniformly distributed throughout the cell, but P3 computes in-cloud values assuming cell-averages are comprised 
        ! of zero values outside cloud. Preemptively multiplying by cldfrac here is needed to get the correct in-cloud ccn value in P3
        nccn_prescribed = ccn_trcdat*cld_frac_l
+       ! SPA produced ccn is in units of #/cm3 rather than #/kg, so we need to
+       ! convert
+       if (is_spa_active) then
+          do icol = 1,ncol
+             do k = 1,pver
+                nccn_prescribed(icol,k) = nccn_prescribed(icol,k)*1.0e6_rtype * (dz(icol,k)*grav)/state%pmid(icol,k)
+             end do
+          end do
+       end if 
     end if
 
     ! CALL P3
--- a/components/eam/src/physics/cam/micro_p3_interface.F90
+++ b/components/eam/src/physics/cam/micro_p3_interface.F90
@@ -744,8 +744,7 @@ end subroutine micro_p3_readnl
                               rho_h2os, &
                               qsmall, &
                               mincld, & 
-                              inv_cp, &
-                              grav=g 
+                              inv_cp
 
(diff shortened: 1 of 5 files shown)


AaronDonahue:
The fails in the AT are expected, `p3_run_and_cmp`

components/eam/src/physics/cam/micro_p3_interface.F90
```diff
@@ -1046,6 +1047,15 @@ subroutine micro_p3_tend(state, ptend, dtime, pbuf)
        ! ccn is uniformly distributed throughout the cell, but P3 computes in-cloud values assuming cell-averages are comprised 
        ! of zero values outside cloud. Preemptively multiplying by cldfrac here is needed to get the correct in-cloud ccn value in P3
        nccn_prescribed = ccn_trcdat*cld_frac_l
+       ! SPA produced ccn is in units of #/cm3 rather than #/kg, so we need to
```
PeterCaldwell, line 1050: Isn't the point that we want to change the SPA input files so CCN is in #/kg and we don't need to do this calculation either in C++ or in F90????

components/scream/src/physics/p3/p3_main_impl.hpp
```diff
@@ -185,10 +185,11 @@ ::p3_main_part1(
       // prescribe that value
 
       if (do_prescribed_CCN) {
```
PeterCaldwell, line 187: I was trying to argue this morning that this entire weird "we're in microphysics but we're going to do aerosol activation now" chunk of code below should be deleted in C++. If SPA does aerosol, we should expect the state after SPA has the correct nc value.

components/eam/src/physics/cam/micro_p3_interface.F90
AaronDonahue: Yes. If we have the SPA files ready for SCREAMv0 to run without this conversion than we can add them to this PR and remove this line.  I meant for this PR to clean up the interface without impacting SCREAMv0 runs, while we wait for the SPA data.

If we want, we can wait for the data to come available before integrating this PR.

components/scream/src/physics/p3/p3_main_impl.hpp
AaronDonahue: Yeah, that would probably simplify things a lot.  For example we could set nccn_prescribed to a constant value in the interface if we are not using SPA.  I'm guessing we don't plan on predictingNC right?

AaronDonahue:
adding WIP until we have the new SPA data.

components/scream/src/physics/p3/p3_main_impl.hpp
AaronDonahue: @PeterCaldwell , can you comment further on if you want me to take care of purging `do_prescribed_CCN` conditionals in this PR?

components/eam/src/physics/cam/micro_p3_interface.F90
PeterCaldwell: Aren't the new SPA files available now so we can delete this?

PeterCaldwell (approved):
Looks good. In the end, this PR doesn't do much since we decided to keep nc_prescribed in the p3 code.

AaronDonahue:
@hassanbeydoun , I've changed the spa data file name.  Can you re-review one more time so that the PR can go in after the AT passes?

hassanbeydoun (approved):

PR #1165, opened 2021-08-20
Automated field checking during each run step

AaronDonahue:
Introduce automatic field property checking to atm_proc class.
    
This comimt introduces to new functions to the AtmosphereProcess class, check_required_fields and check_computed_fields.
    
The purpose of these two functions is to run the suite of field property checks for each field that is either required or computed by a specific process.
    
The AtmosphereProcess instances control what happens for those checks by overriding the check_required_fields_impl and check_computed_fields_impl.
    
At the run stage, the check_required_fields_impl is called just prior to run_impl to ensure that the state heading into the process run sequence is valid.
    
Similarly after run_impl, the check_computed_fields_impl is called to ensure that all fields computed or updated by the process are valid.
    
These new routines have been implemented for all processes currently in SCREAM.

[BFB]


AaronDonahue:
@bartgol and @jeff-cohere - This PR is WIP for now, but I hope to get it ready pretty soon.  I mainly posted it now instead of later so I can get some feedback on my approach, before I commit to heavily down this road.

I put this together after finding it a bit messy to use the field property checker to investigate the homme+physics fail that Conrad reported.  I wanted to start checking in more granularity and figure out if SHOC was the problem or if SHOC was just catching the error.  I decided a clean infrastructure for robust use of the property checker would be a lot nicer than always hacking something in when I needed it.

Any thoughts or comments are totally appreciated.

components/scream/src/share/atm_process/atmosphere_process.hpp
```diff
@@ -335,6 +346,8 @@ class AtmosphereProcess : public ekat::enable_shared_from_this<AtmosphereProcess
   virtual void set_required_field_impl (const Field<const Real>& f) = 0;
   virtual void set_computed_field_impl (const Field<      Real>& f) = 0;
 
+  virtual void check_required_fields_impl () = 0;
```
jeff-cohere, line 349: Are these always required for every subclass? You can give them an empty body here by default if you don't think these checks apply universally. Of course, that invites maintainers not to implement them, so maybe this is better.

AaronDonahue: ahhh, so if i replace this line with ` virtual void check_required_fields_impl ();` then they won't be required to be defined by every subclass?  I think that is probably a better approach.  Most people will use existing process interfaces as templates, which will all have these defined, so I think the risk of a process skipping is low.

bartgol: You also need to give an implementation, so
```
virtual void check_required_fields_impl () {}
```

jeff-cohere: Great! To do this, try

```
virtual void check_required_fields_impl() {}
```

to make sure it has an empty body.

jeff-cohere: Also, if these checks are not allowed to change the state of an atmosphere process, make sure they're declared  with `const` at the end, e.g.

```
virtual void check_required_fields_impl() const {}
```

This rules out the checking as a source of error (unless you want to allow field checks to repair the fields in situ).

AaronDonahue: Would the `repair` feature of the field property checks not work then?

AaronDonahue: nvmd, you said that directly :-)

AaronDonahue: This is a good suggestion, the `check_required_fields` should probably not be allowed to "repair".  But `check_computed_fields` might.

bartgol:
It seems to me that the implementation of the check methods is quite general, in the sense that does not use any information about the specific process. If that's the intent, we could move the impl in the base class. Currently, we don't store the in/out fields in the base class. But this would be an example of why we could. We could store the fields during the `set_computed_field` and `set_required_field` (and similarly for the groups). Then, move the impl of the checks to the base class.

Notice that this would mean that _all_ atm process, by default, run all checks on all inputs/outputs. Of course, they can override the method to "skip" some checks, if they so desire.

Having all fields/groups stored in the base class might also remove (globally) a few lines of code, since virtually all atm procs store some map of the form `std::map<std::string,Field<const Real>>` for their inputs, and similarly for their outputs...

AaronDonahue:
> It seems to me that the implementation of the check methods is quite general, in the sense that does not use any information about the specific process. If that's the intent, we could move the impl in the base class. Currently, we don't store the in/out fields in the base class. But this would be an example of why we could. We could store the fields during the `set_computed_field` and `set_required_field` (and similarly for the groups). Then, move the impl of the checks to the base class.
> 
> Notice that this would mean that _all_ atm process, by default, run all checks on all inputs/outputs. Of course, they can override the method to "skip" some checks, if they so desire.
> 
> Having all fields/groups stored in the base class might also remove (globally) a few lines of code, since virtually all atm procs store some map of the form `std::map<std::string,Field<const Real>>` for their inputs, and similarly for their outputs...

I like this suggestion.  I think it would save on developers having to add another set of defined `impl` that would most likely be a copy/paste of all the other processes.  I went with the strategy I did because I saw that the "required" and "computed" fields weren't exposed to the base class, but if that is easy to fix I say we go with that.

components/scream/src/physics/shoc/atmosphere_macrophysics.cpp
```diff
@@ -346,16 +349,26 @@ void SHOCMacrophysics::initialize_impl (const util::TimeStamp& t0)
...
+  m_fields_out["Q"].add_property_check(positivity_check);
+
 }
 
 // =========================================================================================
 void SHOCMacrophysics::run_impl (const Real dt)
 {
   // Copy inputs to host. Copy also outputs, cause we might "update" them, rather than overwrite them.
-  for (auto& it : m_shoc_fields_in) {
+  for (auto& it : m_fields_in) {
```
bartgol, line 368: As for P3, this seems to be pointless now that we run shoc on device.

components/scream/src/share/atm_process/atmosphere_process.hpp
```diff
@@ -332,16 +389,23 @@ class AtmosphereProcess : public ekat::enable_shared_from_this<AtmosphereProcess
     f.get_header_ptr()->get_tracking().add_customer(weak_from_this());
   }
 
-  virtual void set_required_field_impl (const Field<const Real>& f) = 0;
-  virtual void set_computed_field_impl (const Field<      Real>& f) = 0;
+  virtual void set_required_field_impl (const Field<const Real>& f) {};
```
bartgol, line 392: We should add a comment on top of these two methods, explaining that subclasses _can_ override this if they need to do additional stuff (like additional book keeping or setup), but it is not necessary.

components/scream/src/share/atm_process/atmosphere_process.hpp
```diff
@@ -332,16 +389,23 @@ class AtmosphereProcess : public ekat::enable_shared_from_this<AtmosphereProcess
...
+  virtual void set_computed_field_impl (const Field<      Real>& f) {};
 
-private:
+  virtual void check_required_fields_impl () const {}
```
bartgol, line 395: Here too: the base class already runs all checks, so in general there's no need to override these. But derived class _may_ want to do more stuff. E.g., repair the field if the checks don't pass, or, for a AdaptiveAtmProcess, we may think of "re-doing" the timestep, subcycling with dt/2.

components/scream/tests/coupled/homme_physics/CMakeLists.txt
```diff
@@ -16,9 +16,10 @@ set (NEED_LIBS cld_fraction shoc p3 scream_rrtmgp rrtmgp ${NETCDF_C} ${dynLibNam
 
 # Add source files
 set (SRC shoc_cld_p3_rrtmgp.cpp ${SCREAM_BASE_DIR}/src/physics/rrtmgp/tests/rrtmgp_test_utils.cpp)
-
-CreateUnitTest(homme_shoc_cld_p3_rrtmgp "homme_shoc_cld_p3_rrtmgp.cpp" "${NEED_LIBS}" EXE_ARGS "< namelist.nl" LABELS "dynamics;driver;shoc;cld;p3;rrtmgp;physics")
-set_target_properties(homme_shoc_cld_p3_rrtmgp PROPERTIES COMPILE_FLAGS "${YAKL_CXX_FLAGS}")
+if (NOT SCREAM_AUTOTESTER) # TODO: This test should be reinstated once the homme+physics test is fixed.
```
bartgol, line 19: Wouldn't this cause nightlies to fail? Can we comment the test altogether then? I get the desire to have the AT passing, but this would cause the nightlies to be constantly red, for a test we know should fail.

bartgol (changes requested):
I think we're close to merge this. I have a few more comments.

components/scream/tests/coupled/homme_physics/CMakeLists.txt
AaronDonahue: That's the point, we want the constant reminder from the nightlies that there is a known failed test that still needs to be fixed.  But we don't want the the AT to hold everyone else up.

jeff-cohere (approved):

components/scream/src/physics/p3/atmosphere_microphysics.cpp
```diff
@@ -314,7 +315,7 @@ void P3Microphysics::run_impl (const Real dt)
   // advance it, updating the p3 fields.
   auto ts = timestamp();
   ts += dt;
-  for (auto& f : m_p3_fields_out) {
+  for (auto& f : m_fields_out) {
```
bartgol, line 318: TODO: this seems like another good candidate for stuff to move in the base class, since we kind of do it in all subclasses.

components/scream/src/share/atm_process/atmosphere_process.hpp
```diff
@@ -332,16 +389,37 @@ class AtmosphereProcess : public ekat::enable_shared_from_this<AtmosphereProcess
     f.get_header_ptr()->get_tracking().add_customer(weak_from_this());
   }
 
-  virtual void set_required_field_impl (const Field<const Real>& f) = 0;
-  virtual void set_computed_field_impl (const Field<      Real>& f) = 0;
+  // The base class already registers the required and computed fields in
+  // the set_required/computed_field main routines.  These impl definitions
+  // provide a way for subclasses to define extra field bookkeeping that
+  // is not already covered by the `set_required/computed_field`.  It is
+  // not essential for subclasses to overwrite these two routines, but it
```
bartgol, line 396: Typo: overwrite->override. No need to change it here though (it would trigger another AT round, and require another approval).

bartgol (approved):
Looks good. Thanks for addressing all comments!

bartgol:
Since everything passes, and reviewers approved, I'm gonna merge this.

PR #1168, opened 2021-08-23
Allow scorpio I/O on dynamics grid

bartgol:
This PR allows to perform scorpio I/O on the dyn grid (and, more generally, on a grid with repeated dofs). This required to add the ability to store a ptr to a "unique" grid inside a grid. The output manager, uses this grid to build an auxiliary field manager (on the unique grid) to use for I/O, and a remapper (from grid to unique-grid) to be run right before running the output stream.

This is work in progress. In particular, there is no capability to _load_ fields directly from the dyn grid (which is the feature needed to have a BFB restart of HOMME).

NOTE: this PR is built on top of #1162 , so even if it appears as if it contains many commits, the list will get shorter once that PR goes in (I will also rebase on master after that happens, if necessary). In particular, only the commits starting at 5599335 are peculiar to this PR.

### DIFF
--- a/components/scream/src/dynamics/homme/dynamics_driven_grids_manager.cpp
+++ b/components/scream/src/dynamics/homme/dynamics_driven_grids_manager.cpp
@@ -15,6 +15,7 @@ namespace scream
 DynamicsDrivenGridsManager::
 DynamicsDrivenGridsManager (const ekat::Comm& comm,
                             const ekat::ParameterList& p)
+ : m_comm (comm)
 {
   if (!is_parallel_inited_f90()) {
     // While we're here, we can init homme's parallel session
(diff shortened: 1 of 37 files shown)


bartgol:
This PR introduces the possibility of performing IO with fields defined on the dynamics grid (but must be stored in a FieldManager). The goal is achieved by the following changes:

- For each grid, allow storing a corresponding "unique grid", meaning a grid where each DOF appears once _globally_ (that is, a mesh that would have no repeated dofs running with 1 MPI rank).
- Allow creation of both a physics GLL grid and a physics FV grid (the gll-fv remap mod in Homme still forces to have only one instance of FV grid, but I doubt that's a limitation, at least for now).
- IO is performed on the unique grid associated with the input FieldManager's grid (the two may be the same grid). If the unique grid is indeed a different grid, a remap is performed to/from the unique grid before/after IO for output/input respectively.

components/scream/src/dynamics/homme/dynamics_driven_grids_manager.cpp
```diff
@@ -139,14 +131,17 @@ build_grids (const std::set<std::string>& grid_names,
 
 void DynamicsDrivenGridsManager::build_dynamics_grid () {
   if (m_grids.find("Dynamics")==m_grids.end()) {
+    // Build the Physics GLL grid, to be used as 'unique grid' in the dyn grid
+    build_physics_grid("Physics GLL");
```
AaronDonahue, line 135: Just a question on syntax, while I'm looking at this code.  We have "Physics GLL" and "Physics" used throughout the code.  Would it make sense to never use "Physics" and require the user to be specific what kind of physics grid is in use?  So have "Physics GLL", "Physics Point", and eventually "Physics PG"

components/scream/src/dynamics/homme/dynamics_driven_grids_manager.cpp
```diff
@@ -198,10 +194,10 @@ build_physics_grid (const std::string& name) {
 
     // Get dimensions and create "empty" grid
     const int nlev  = get_nlev_f90();
-    const int nlcols = get_num_local_columns_f90 ();
-    const int ngcols = get_num_global_columns_f90 ();
+    const int nlcols = get_num_local_columns_f90 (pg_type % 10);
```
AaronDonahue, line 197: Does this mean all grids are PG type grids?  Do we plan to only support 10 grid codes, hence the modulo?  If so, why would we allow something to be 11 and map to 1, for example.  Wouldn't we error out?  I'm probably missing something here.

components/scream/src/dynamics/homme/dynamics_driven_grids_manager.cpp
bartgol: Sure, we can do that. I can remove any use of `Physics`, and maybe call the non-GLL phys grid `Physics FV` rather than `PG`, They are all phys grids, after all, we are distinguishing them by the numerical algorithm they are used with.

bartgol: Actually, there are two uses of 'Physics' in scorpio in/out, which can be purged (we now can do I/O on dyn grid too). Other than those, the places that have `Physics` as a grid name are either tests building a PointGrid (and they give it the name "Physics"), or the PhysicsOnlyGridManager, which one could argue it should be called "PointGridOnlyGridManager" maybe.

I'm gonna move the PhysicsOnlyGridManager to share/grid, and use it for io tests too.

AaronDonahue (approved):
I had a couple comments, but they shouldn't hold up this PR.  This was a mammoth amount of work.  Thanks @bartgol

AaronDonahue (approved):

bartgol:
It looks like the failures on mappy are due to OpenMPI having not being able to create a scratch folder in /tmp. Apparently, the chances of this happening are miniscule (see [this github issue](https://github.com/open-mpi/ompi/issues/8510)for more details), so I'm just gonna retest.

PR #1173, opened 2021-08-31
Redo scream_cpl_indices for SurfaceCoupling

tcclevenger:
Simplify `scream_cpl_indices.F90`. Previously not all cpl fields were present. Now, I just query the cpl array that contains the import/export data for its size, explicitly define all the fields needed by scream, and set all other fields to "unused" (import) or "set_zero" (export). 

Also update SC test to match this.

### DIFF
--- a/components/scream/src/mct_coupling/scream_cpl_indices.F90
+++ b/components/scream/src/mct_coupling/scream_cpl_indices.F90
@@ -6,20 +6,13 @@ module scream_cpl_indices
   private
 
   ! Focus only on the ones that scream imports/exports (subsets of x2a and a2x)
-  integer, parameter, public :: num_required_cpl_imports = 21
   integer, parameter, public :: num_scream_imports       = 8
-  integer, parameter, public :: num_required_exports     = 12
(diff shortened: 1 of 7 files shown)


components/scream/src/mct_coupling/scream_cpl_indices.F90
```diff
@@ -47,183 +39,69 @@ subroutine scream_set_cpl_indices (x2a, a2x)
     !
     integer :: i,idx
 
+    ! total number of import/export var in data
+    num_cpl_imports = size(x2a%rAttr,1)
+    num_cpl_exports = size(a2x%rAttr,1)
```
tcclevenger, line 44: Main change: these are no longer explicitly defined. We now query the x2a/a2x arrays.

tcclevenger: The main changes are in this file, but the diff is hard to read.. May be easier to look at the new file as a whole.

tcclevenger:
@AaronDonahue I also verified manually on weaver. Still need to ask for reviews though

bartgol (approved):

PR #1174, opened 2021-08-31
Bugfix to ftype=2 option in the homme dynamics interface.

AaronDonahue:
Switch homme_shoc_cld_p3_rrtmgp coupled test to use ftype=2 rather
than ftype=0.

Bump up the number of iterations for the homme_shoc_cld_p3_rrtmgp
test because with ftype=2 it is able to run many more iterations.


AaronDonahue:
Note: all tests pass on lassen.

components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
```diff
@@ -677,6 +677,7 @@ void HommeDynamics::homme_pre_process (const Real dt) {
         fq /= dt;
         fq *= dp;
       });
+      Kokkos::fence();
```
bartgol, line 680: I'm curious: was this fence necessary (i.e., did tests fail without), or did you add it just for symmetry with the other `case` clause? I suspect these fences are not necessary, at least not at the moment. They might become relevant in case of async tasks and/or multiple cuda streams though, so it's good to have them (they don't hurt).

bartgol (approved):

PR #1175, opened 2021-09-01
Compute solar zenith angle for radiation

brhillman:
Compute solar zenith angle each time radiation is called, based on lat/lon coordinates, calendar day, and time of day. This adds a bridge to the Fortran functions that compute the Earth orbital parameters and solar zenith angle in share/utils/shr_orb_mod.F90.


components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
```diff
@@ -174,6 +176,9 @@ void RRTMGPRadiation::init_buffers(const ATMBufferManager &buffer_manager)
 
 void RRTMGPRadiation::initialize_impl(const util::TimeStamp& /* t0 */) {
   using PC = scream::physics::Constants<Real>;
+  // Determine orbital year
+  m_orbital_year = m_rrtmgp_params.get<Real>("Orbital Year",1990);
```
jeff-cohere, line 180: If `1990` is a default value, do we want to document this somewhere? Should it have a default value?

AaronDonahue: Good question, it is the default value used by EAM.  We can definitely document this at this line.  As for whether or not we actually want a default value I would defer to @brhillman .  Presumably, if we don't have a default value we would make it so that the code throws an error if no Orbital Year is provided?

components/scream/src/physics/rrtmgp/share/CMakeLists.txt
```diff
@@ -0,0 +1,53 @@
...
+set(ENABLE_GENF90 True)
+include(genf90_utils)
+include(Sourcelist_utils)
+
+# NOTE: I should be able to do this:
+#    # Note that this sets the share_sources variable to the list of source files we need to build
+#    add_subdirectory(${SCREAM_BASE_DIR}/../../share/util ${CMAKE_CURRENT_BINARY_DIR}/share_util_build)
+# However, that did not seem to add the fortran sources derived from the genf90 templates to the targets.
+# Instead, try explicitly adding sources and process the genf90 templates ourselves.
+# TODO: Figure out why the CMakeLists.txt and the cmake macros there do not handle this correctly.
```
jeff-cohere, line 17: As far as I understand, CMake is not designed to dynamically generate lists of source files. Dynamically generating these lists requires a more aggressive set of criteria for regenerating CMake's makefiles. Otherwise you end up thinking you can just type `make` when in fact you must rerun `cmake` to get consistent results.

components/scream/src/physics/rrtmgp/share/shr_orb_mod_c.hpp
```diff
@@ -0,0 +1,14 @@
+#ifndef SHR_ORB_MOD_C_HPP
+#define SHR_ORB_MOD_C_HPP
+extern "C" void shr_orb_params_c(
```
jeff-cohere, line 3: I still think our "_c" and "_f90" convention is backwards.

components/scream/src/physics/rrtmgp/tests/rrtmgp_unit_tests.cpp
```diff
@@ -1,8 +1,10 @@
 #include "catch2/catch.hpp"
 #include "physics/rrtmgp/rrtmgp_heating_rate.hpp"
 #include "physics/rrtmgp/scream_rrtmgp_interface.hpp"
+//#include "physics/rrtmgp/zenith.hpp"
```
jeff-cohere, line 4: Remove?

jeff-cohere (approved):
Looks good to me. I left a few minor comments/questions.

components/scream/src/physics/rrtmgp/tests/CMakeLists.txt
```diff
@@ -9,7 +9,7 @@ endif()
 
 # Required libraries
 find_library(NETCDF_C netcdf HINTS ${NetCDF_C_PATHS}/lib)
-set (NEED_LIBS ${NETCDF_C} rrtmgp scream_rrtmgp scream_share yakl)
+set (NEED_LIBS ${NETCDF_C} rrtmgp scream_rrtmgp scream_share yakl share_util)
```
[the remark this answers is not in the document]
bartgol: P.S: I don't know how to comment a line not involved in the PR, but I kind of want to: below I see

```
set_target_properties(generate_baseline PROPERTIES COMPILE_FLAGS "${YAKL_CXX_FLAGS}")
```
This will tell CMake to add ${YAKL_CXX_FLAGS} to the compile flags of _all languages_. I assume some of those flags might be c++ specific. In this case, one should use generator expressions (and also, use target_compile_options, which is the "cmake-y" way for this stuff):
```
target_compile_options(generate_baselines PUBLIC $<$<COMPILE_LANGUAGE:CXX>:${YAKL_CXX_FLAGS}>)
```

components/scream/src/physics/rrtmgp/zenith.hpp
```diff
@@ -0,0 +1,58 @@
...
+ *
+ *     shr_orb_params()
+ *     shr_orb_decl()
+ *     shr_orb_avg_cosz()
+ *
+ * That should at least get us a baseline, and then we can work on porting
+ * shr_orb_mod.F90 to C++.
+ */
+
+extern void shr_orb_params(double calday, double eccen, double mvelpp, double lambm0,
```
bartgol, line 21: The keyword `extern` along is somewhat pointless, since functions already have external linkage by default.

bartgol: By the way, I do not see any definition for `shr_orb_params`. I do see a `bind(C,name='shr_orb_params_c')` in a f90 module, but there's a name mismatch. Also, assuming that `shr_orb_params_c` is the function definition corresponding to dhis function declaration, you should declare this function as `extern "C"`, not just `extern`. The "C" part is crucial, since it tells the linker to link this function using C name mangling, even if this file is included in a cpp file compiler with c++ compilers.

bartgol (changes requested):
Looks ok. There's one issue with the C-F90 bridges, I think.

components/scream/src/physics/rrtmgp/share/shr_orb_mod_c.hpp
brhillman: Apologies, I’ll fix.

components/scream/src/physics/rrtmgp/atmosphere_radiation.hpp
```diff
@@ -59,6 +63,8 @@ class RRTMGPRadiation : public AtmosphereProcess {
   int m_nlay;
   view_1d_real m_lat;
   view_1d_real m_lon;
+  // The orbital year, used for zenith angle calculations:
+  Int m_orbital_year;
```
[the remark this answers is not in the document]
brhillman: EAM uses a perpetual orbital year in many configurations (so the year that’s used for orbital parameters always stays the same). This is usually logic that is handled in the coupler. For now, prescribing a constant orbital year consistent with the EAM configurations we want to emulate.

components/scream/src/physics/rrtmgp/share/CMakeLists.txt
```diff
@@ -0,0 +1,53 @@
...
+    ${SCREAM_BASE_DIR}/../../share/util/shr_sys_mod.F90
+    ${SCREAM_BASE_DIR}/../../share/util/shr_const_mod.F90
+    ${SCREAM_BASE_DIR}/../../share/util/shr_log_mod.F90
+    ${SCREAM_BASE_DIR}/../../share/util/shr_abort_mod.F90
+    ${SCREAM_BASE_DIR}/../../share/util/shr_mpi_mod.F90
+    ${SCREAM_BASE_DIR}/../../share/util/shr_strconvert_mod.F90
+)
+# Process genf90 template files. This adds a custom command (and hence target) for each f90 source
+# that needs to be built from the genf90 template files listed in GENF90_SOURCE.
+foreach (SRC_FILE ${GENF90_SOURCE})
```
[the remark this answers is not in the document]
brhillman: I agree, but I don’t know how to avoid this here. I’m only using this because that’s how these sources are compiled in EAM’s cmake, and the implementation is taken directly from (this is compiling such “templated” F90 sources). If you have a better idea that does not entail us completely rewriting all of the shr modules in this PR, I’d be happy to implement it.

components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
brhillman: We actually probably want to change this to allow for a dynamic orbital year. The configurations we are running right now use a perpetual orbital year, and for F-cases this looks like it’s set to 1990 in the EAM compsets for some reason. A better idea is to set the default to something that would trigger grabbing the year from the current timestamp at runtime.

components/scream/src/physics/rrtmgp/share/CMakeLists.txt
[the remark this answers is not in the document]
bartgol: Oh, sorry, I wasn't trying to say "do it another way". It was just a stand-alone rant. :)

However, in a full CIME build, I wonder if these modules may conflict with shared libs built by E3SM. I am mildly positive CIME will build this stuff at some point, and then we may have trouble at link time. Perhaps we should check if this is a CIME build, and, if not, proceed to build these files?

components/scream/src/physics/rrtmgp/share/CMakeLists.txt
brhillman: Okay, I think I get what you’re saying. I think I was mostly upset that I could not simply do:
```
 add_subdirectory(${SCREAM_BASE_DIR}/../../share/util ${CMAKE_CURRENT_BINARY_DIR}/share_util_build)
```
but if I understand correctly that’s because of how the CMakeLists.txt is structured in that folder, and that’s probably because of the use of genf90.

components/scream/src/physics/rrtmgp/share/shr_orb_mod_c.hpp
brhillman: Okay I think I fixed this. I renamed the bridges with _c2f since it creates a C interface that calls F90 code. Let me know if this isn’t quite what you were looking for.

components/scream/src/physics/rrtmgp/tests/rrtmgp_unit_tests.cpp
brhillman: Removed.

components/scream/src/physics/rrtmgp/CMakeLists.txt
```diff
@@ -10,6 +10,10 @@ add_subdirectory(${SCREAM_BASE_DIR}/../../externals/YAKL ${CMAKE_CURRENT_BINARY_
 target_compile_options(yakl PUBLIC $<$<COMPILE_LANGUAGE:Fortran>:-fno-default-real-8 -fno-default-double-8>)
 EkatDisableAllWarning(yakl)
 
+# We will need the shr_orb_mod source too; this builds library share_util
+enable_language(Fortran)
```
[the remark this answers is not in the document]
brhillman: This should be fixed now.

components/scream/src/physics/rrtmgp/share/CMakeLists.txt
```diff
@@ -0,0 +1,53 @@
...
+list(APPEND CMAKE_MODULE_PATH ${CIMEROOT}/src/CMake)
+set(GENF90 ${CIMEROOT}/src/externals/genf90/genf90.pl)
+set(ENABLE_GENF90 True)
+include(genf90_utils)
+include(Sourcelist_utils)
+
+# NOTE: I should be able to do this:
+#    # Note that this sets the share_sources variable to the list of source files we need to build
+#    add_subdirectory(${SCREAM_BASE_DIR}/../../share/util ${CMAKE_CURRENT_BINARY_DIR}/share_util_build)
+# However, that did not seem to add the fortran sources derived from the genf90 templates to the targets.
```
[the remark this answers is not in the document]
brhillman: Removed this comment.

components/scream/src/physics/rrtmgp/zenith.hpp
brhillman: This was a temporary file as I was figuring out how to implement this that I mistakenly left behind after deciding to just use the fortran codes for the time being. This file has since been removed. `shr_orb_params` is in the top level share source, called from the bridged defined in this PR. I do have `extern “C”` defined for these. Please do verify that I’ve taken care of this properly.

components/scream/src/physics/rrtmgp/tests/CMakeLists.txt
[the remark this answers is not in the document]
brhillman: Thanks Luca. This should all be fixed (including the suggestion about using the generator expressions).

bartgol (approved):

components/scream/src/physics/rrtmgp/tests/CMakeLists.txt
```diff
@@ -9,7 +9,7 @@ endif()
 
 # Required libraries
 find_library(NETCDF_C netcdf HINTS ${NetCDF_C_PATHS}/lib)
-set (NEED_LIBS ${NETCDF_C} rrtmgp scream_rrtmgp scream_share yakl share_util)
+set (NEED_LIBS ${NETCDF_C} scream_rrtmgp)
```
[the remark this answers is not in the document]
brhillman: Thanks for pointing this out. Testing this now.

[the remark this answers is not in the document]
brhillman: Fixed.

components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
brhillman: @AaronDonahue this should be fixed now, I set a default to -9999, and changed to code to grab the current year from the timestamp in the event this value is negative (the default). Can you take a look at this?

AaronDonahue:
It looks like just a couple of test fails in the AT.  It appears like the `debug_nofpe` test suite is where the fail pops up.  I'll take a look and try to figure out why.

bartgol (approved):

PR #1177, opened 2021-09-07
Add needed team_barrier() to Homme interface

tcclevenger:
When running through CIME with `OMP_NUM_THREADS >1`, there was an internal Homme `NaN` check triggered for  `vTh_dp`. These team_barriers after calculating `p_mid` and before `vTh_dp` fixed the values. Once this was fixed, the 2nd team_barrier was necessary to stop SHOC from erroring out early with negative temperature.

Interestingly enough, these doesn't matter for the homme_physics test.


bartgol (approved):
Yes, these are definitely needed.  I'm not sure why standalone homme does not show the same behavior...

PR #1180, opened 2021-09-09
Fix area/lat/lon computations in phys_grid_mod

tcclevenger:
In phys_grid_mod, an assumption was made that element global ids were contiguous in mpi ranks (i.e., rank0 owned dofs `1,...,n1`, rank1 owned `n1+1,...,n2`, etc..). This is not the case. and so indexing needs to change when computing area and lat/lon. Now we order global geo arrays with all rank0 elements, then rank1 elements, etc. where the global id of each element is not necessarily its global index in the arrays. This matches the way dofs are handled in the same file:
```
      ! Fill local dofs part
      idof = 1
      do ie=1,nelemd
        do icol=1,elem(ie)%idxP%NumUniquePts
          i = elem(ie)%idxP%ia(icol)
          j = elem(ie)%idxP%ja(icol)

          dofs_l(idof) = elem(ie)%gdofP(i,j) ! idof+offset != elem(ie)%GlobalID necessarily
          idof = idof + 1
        enddo
      enddo

      call MPI_Allgatherv( MPI_IN_PLACE, 0, MPI_DATATYPE_NULL, &
                                       dofs_g, g_ncols, g_offsets, MPIinteger_t, par%comm, ierr)
```
How area is computed:
```
      start = 1
      do ie=1,nelemd
        areaw = 1.0_c_double / elem(ie)%rspheremp(:,:)
        ncols = elem(ie)%idxP%NumUniquePts
        call UniquePoints(elem(ie)%idxP, areaw, area_l(start:start+ncols-1))
        start = start + ncols
      enddo

      call MPI_Allgatherv(MPI_IN_PLACE, 0, MPI_DATATYPE_NULL, &
                          area_g, g_ncols, g_offsets, MPIreal_t, &
                          par%comm, ierr)
```

### DIFF
--- a/components/scream/src/dynamics/homme/dynamics_driven_grids_manager.cpp
+++ b/components/scream/src/dynamics/homme/dynamics_driven_grids_manager.cpp
@@ -166,6 +166,18 @@ void DynamicsDrivenGridsManager::build_dynamics_grid () {
     Kokkos::deep_copy(lat,h_lat);
     Kokkos::deep_copy(lon,h_lon);
 
+#ifndef NDEBUG
+    // Check that latitude and longitude are valid
+    using KT = KokkosTypes<DefaultDevice>;
+    const auto policy = ekat::ExeSpaceUtils<KT::ExeSpace>::get_default_team_policy(ndofs,nlev);
+    Kokkos::parallel_for(policy, KOKKOS_LAMBDA(const KT::MemberType& team) {
+      const Int i = team.league_rank();
+
+      EKAT_ASSERT_MSG(!isnan(lat(i)), "Error! NaN values detected for latitude.");
+      EKAT_ASSERT_MSG(!isnan(lon(i)), "Error! NaN values detected for longitude.");
+    });
+#endif
+
     // Set dofs and geo data in the grid
     dyn_grid->set_dofs (dofs, elgpgp);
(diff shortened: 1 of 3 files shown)


bartgol (commented):
Can we add a test of some sort? I'm not sure where, perhaps a new tiny exec inside `dynamics/homme/tests`? I'm not sure what we could test though...

tcclevenger:
@bartgol Yeah, I was struggling to think of a test.. I'll keep thinking.

bartgol (approved):

bartgol:
> @bartgol Yeah, I was struggling to think of a test.. I'll keep thinking.

I guess there's not much to test. You are simply rearranging entries in the global area/lat/lon arrays. We keep the global arrays around, in case we need more than one phys grid type (e.g., a pg2, and a twin-cols pg2). However, we could also pre-compute the on-rank gids (and geo data) for all possible physics grids (only one N is allowed for pgN, so that means there can be 4 physics grids at most: physics GLL, pgN, and their twin cols counterparts), and get rid of the global arrays.

Anyhow, that's just to say that the way we order the entries in the global area/lat/lon is 100% arbitrary, so long as we then go read the right entries when it's time to grab the rank local stuff. So I don't think there's anything to test here.

PR #1181, opened 2021-09-10
Changes to GroupRequest logic, AtmosphereProcess interfaces, and AD handling of initial conditions.

bartgol:
I apologize for the size of the PR, but the more I tried to accommodate something, the more I realized it was by far easier if I modified something else.

This PR started with one goal: fix how Homme creates the Q_prev copy of the tracers group. It ended up doing more. Here's a tentative list of what it does:

- reworked implementation of `contiguous_superset`. I think the number of LOC increased, but it is imho easier to follow now.
- changed names/impl of GroupRequests (GR) derived on other GRs. E.g., Relationship->DerivationType, Child->Subset, Parent->Superset.
- make FM correctly handle copy of groups: Copy only allowed on the same grid, and the individual fields are not created to avoid clashes with the subfields of original group.
- move most impl of AtmosphereProcess (AP) class into a cpp file (the hpp was getting too long, making hard to see the class interfaces at a glance).
- store all fields/groups in/out in the base AP class, provide getters for derived class (so they can't add more fields to the list, which could mess up some infrastructure logic, like that there is 1-1 correspondence between m_fields_in and the required fields requests).
- allow derived classes to not override any set_blah_impl method, unless *really* needed. All basic bookkeping is already in the AP class
- make AD load IC for field on any grid. This does not yet support loading fields from nc file directly on dyn grid. But at least one can init a dyn grid field to a constant or equal to another field on the phys grid (via syntax `field_1: field_2, Physics_GLL`).

There might be more, but it's late, and I need to go to bed. I wanted to put this in before taking off for the long weekend.

A couple of things that I did not address:
- Homme needs Q_prev only for ftype 0, but currently it always creates it. Fix this by introducing some checks on ftype.
- I realized we have some pb when creating a subgroup of a group, asking for it to be bundled. E.g., SHOC wants a subset of the tracers, and would like them bundled. However, we never create the subfield `Q_shoc=Q(:,1:num_shoc_tr,:)`, since we can only handle subfield that completely slice away a dimension (rather than taking an extent). I have an idea in mind on how to extend this, to allow, say, `subfield(const int idim, const int k0, const int n)`, with `k0` being the 1st index along dimension `idim` to subview, and `n` being the number of slices. But it may take some time, so we need to see how big of a priority it is (I suspect we want it in the long run though).

### DIFF
--- a/components/scream/src/control/tests/ad_tests.yaml
+++ b/components/scream/src/control/tests/ad_tests.yaml
@@ -5,7 +5,7 @@ Debug:
 
 Initial Conditions:
   Point Grid:
-    Initial Conditions File: should_not_be_neeeded_and_code_will_throw_if_it_tries_to_open_this_nonexistent_file.nc
+    Filename: should_not_be_neeeded_and_code_will_throw_if_it_tries_to_open_this_nonexistent_file.nc
     A: 1.0
     B: 2.0
     C: 3.0
--- a/components/scream/src/control/tests/ad_tests.yaml
+++ b/components/scream/src/control/tests/ad_tests.yaml
@@ -4,12 +4,13 @@ Debug:
   Atmosphere DAG Verbosity Level: 5
 
 Initial Conditions:
-  Initial Conditions File: should_not_be_neeeded_and_code_will_throw_if_it_tries_to_open_this_nonexistent_file.nc
-  A: 1.0
-  B: 2.0
-  C: 3.0
-  D: [4.0, 5.0]
-  E: "A"
+  Point Grid:
+    Initial Conditions File: should_not_be_neeeded_and_code_will_throw_if_it_tries_to_open_this_nonexistent_file.nc
+    A: 1.0
+    B: 2.0
+    C: 3.0
+    D: [4.0, 5.0]
+    E: "A"
 
 Atmosphere Processes:
   Number of Entries: 3
@@ -17,7 +18,7 @@ Atmosphere Processes:
 
   Process 0:
     Process Name: Dummy
-    Sub Name: A to BC
+    Sub Name: A to Group
     Grid Name: Point Grid
   Process 1:
     Process Name: Dummy
--- a/components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
+++ b/components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
@@ -724,7 +724,7 @@ void HommeDynamics::import_initial_conditions () {
   using PF = PhysicsFunctions<DefaultDevice>;
 
   const auto dp3d      = get_internal_field("dp3d",dgn).get_view<Pack*****>();
-  const auto ps        = get_internal_field("ps",dgn).get_view<Pack****>();
+  const auto ps        = get_internal_field("ps",dgn).get_view<Real****>();
   const auto v         = get_internal_field("v",dgn).get_view<Pack******>();
   const auto w_i       = get_internal_field("w_i",dgn).get_view<Pack*****>();
   const auto phinh_i   = get_internal_field("phinh_i",dgn).get_view<Pack*****>();
--- a/components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
+++ b/components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
@@ -27,7 +27,6 @@
 #include "physics/share/physics_constants.hpp"
 #include "share/util/scream_common_physics_functions.hpp"
 #include "share/util//scream_column_ops.hpp"
-#include "share/field/field_property_checks/field_positivity_check.hpp"
 
 // Ekat includes
 #include "ekat/ekat_assert.hpp"
@@ -315,10 +314,6 @@ void HommeDynamics::initialize_impl (const util::TimeStamp& /* t0 */)
 
   // Complete homme model initialization
   prim_init_model_f90 ();
-
-  // Set positivity check on the tracers
-  // auto positivity_check = std::make_shared<FieldPositivityCheck<Real> >();
-  // get_group_out("Q",m_ref_grid->name()).m_bundle->add_property_check(positivity_check);
 }
 
 void HommeDynamics::run_impl (const Real dt)
--- a/components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
+++ b/components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
@@ -1,4 +1,5 @@
 #include "atmosphere_dynamics.hpp"
+#include <string>
 
 // HOMMEXX Includes
 #include "Context.hpp"
@@ -62,7 +63,6 @@ void HommeDynamics::set_grids (const std::shared_ptr<const GridsManager> grids_m
 
   using namespace ShortFieldTagsNames;
   using namespace ekat::units;
-  using FID = FieldIdentifier;
   using FL  = FieldLayout;
 
   constexpr int NGP  = HOMMEXX_NP;
@@ -99,7 +99,7 @@ void HommeDynamics::set_grids (const std::shared_ptr<const GridsManager> grids_m
 
   // Note: the field T_mid will contain Temperature (at midpoints) at entry/exit.
   //       However, Homme uses VTheta*dp as state variable. Instead of keeping both
-  //       T and VTheta on dyn grid, vtheta_dp, we remap T_mid from phys directly
+  //       T and VTheta*dp on dyn grid, we remap T_mid from phys directly
   //       into the n0 time-slice of Homme's vtheta_dp, and then do the conversion
   //       T_mid->VTheta_dp in place.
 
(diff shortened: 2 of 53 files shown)


components/scream/src/share/field/field_group.hpp
```diff
@@ -102,12 +104,43 @@ struct FieldGroup {
...
+             >::type
+           >
+  FieldGroup<RealType>& operator= (const FieldGroup<RT>& src) {
+    m_info = src.m_info;
+    copy_fields<RT> (src);
+  }
+
   const std::string& grid_name () const {
-    EKAT_REQUIRE_MSG(m_fields.size()>0,
+    EKAT_REQUIRE_MSG(m_fields.size()>0 || m_bundle,
```
jeff-cohere, line 134: Doesn't this predicate mean that `EKAT_REQUIRE_MSG` will report the error below if `m_bundle` is set? If so, what does it mean that we're checking `m_bundle` on line 137?

components/scream/src/share/atm_process/atmosphere_process.hpp
```diff
@@ -82,192 +87,67 @@ class AtmosphereProcess : public ekat::enable_shared_from_this<AtmosphereProcess
...
+  void set_computed_field (const Field<Real>& f);
+  void set_required_group (const FieldGroup<const Real>& group);
+  void set_updated_group (const FieldGroup<Real>& group);
+
+  // These methods checks that all the in/out fields of this atm process are valid.
+  // For each field, these routines run all the property checks stored in the field.
+  // Derived classes can perform additional checks (or repairs) by overriding the
+  // corresponding check_xyz_fields_impl method(s).
+  // Note: We don't want inputs to be 'repaired', since they might break assumptions
+  //       in other atm procs (e.g., some atm procs might haave a "backup" copy).
```
tcclevenger, line 122: If you have other changes to make, here's a typo "haave".

components/scream/src/share/atm_process/atmosphere_process.hpp
```diff
@@ -366,64 +248,105 @@ class AtmosphereProcess : public ekat::enable_shared_from_this<AtmosphereProcess
...
+
+  FieldGroup<const Real>& get_group_in(const std::string& group_name, const std::string& grid_name);
+  FieldGroup<const Real>& get_group_in(const std::string& group_name);
+  FieldGroup<Real>& get_group_out(const std::string& group_name, const std::string& grid_name);
+  FieldGroup<Real>& get_group_out(const std::string& group_name);
+
+  // These methods set up an extra pointer in the m_[fields|groups]_[in|out]_pointers,
+  // for convenience of use (e.g., use a short name for a field/group).
+  // Note: these methods do *not* create a copy of the field/group. Also, notice that
+  //       these methors need to be created *after* set_fields_and_groups_pointers().
```
tcclevenger, line 305: Typo: "methods"

components/scream/src/share/field/field_request.hpp
```diff
@@ -2,7 +2,7 @@
 #define SCREAM_FIELD_REQUEST_HPP
 
 #include "share/field//field_identifier.hpp"
-#include "share/scream_types.hpp"
+#include "share/util//scream_utils.hpp"
```
tcclevenger, line 5: Maybe this is a feature I don't know, but should there be 2 slashes here?

tcclevenger (commented):
This looks great, makes the code much nicer. I've looked at most of the files and have a few comments (mostly just typos, so feel free to ignore if it needs no changes before merging). I'll take a look at the rest later today or Monday if it still isn't merged.

components/scream/src/share/field/field_group.hpp
bartgol: Uhm, it reports the error if neither of the conditions pass. The grid name is pulled from any of the fields in the group, or from the bundled field. If neither are set, we can't retrieve the grid name. Since a group may have `m_fields` set, but not the bundle, or viceversa, at 137 we check which one is set, and grab the grid name from the first one that is available.

components/scream/src/share/field/field_request.hpp
bartgol: I've seen that same typo happening multiple times on my PRs. It might be the autocomplete and me having a race condition for adding the slash, I am not sure... But no, it's not a feature. :)

components/scream/tests/uncoupled/rrtmgp/input.yaml
```diff
@@ -20,27 +20,28 @@ Grids Manager:
 
 # The name of the file containing the initial conditions for this test.
```
jeff-cohere, line 21: Seems like this comment might no longer be accurate.

components/scream/tests/uncoupled/homme/input.yaml
```diff
@@ -4,12 +4,15 @@ Debug:
...
 
 Initial Conditions:
-  Initial Conditions File: homme_standalone_ne4np4.nc
-  # Set X_prev = X
-  T_mid_prev: T_mid
-  horiz_winds_prev: horiz_winds
-  w_int: 0.0 # Start from zero vert velocity
-  w_int_prev: 0.0
+  Physics GLL:
+    Initial Conditions File: homme_standalone_ne4np4.nc
```
jeff-cohere, line 8: What about just `File`? It's already in the `Initial Conditions` section/block

bartgol: I agree.

components/scream/src/share/field/field_group.hpp
jeff-cohere: Got it. Thanks!

components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
```diff
@@ -120,72 +120,70 @@ void HommeDynamics::set_grids (const std::shared_ptr<const GridsManager> grids_m
...
-                              &tracers,Relationship::Alias);
-    add_group<Updated>(tracers_prev);
-  }
+  // We need a temp also for tracers, but at the time this fcn is called, we still don't
+  // know the size of the tracers group. We can create that temp later.
+  create_internal_field("FM",{COL,CMP,LEV},{ncols,3,NVL},m_ref_grid->name());
+  create_internal_field("FT",{COL,    LEV},{ncols,  NVL},m_ref_grid->name());
 
-  // Create the PD and DP remappers
+  // Dynamics backs out tendencies from the states, and pass those to Homme.
```
jeff-cohere, line 167: Not a huge deal, but `pass` -> `passes`

components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
```diff
@@ -120,72 +120,70 @@ void HommeDynamics::set_grids (const std::shared_ptr<const GridsManager> grids_m
...
 }
 
 void HommeDynamics::
-set_updated_group (const FieldGroup<Real>& group)
+set_updated_group_impl (const FieldGroup<Real>& group)
 {
   const auto& name = group.m_info->m_group_name;
-  const int ftype = get_homme_param<int>("ftype");
-  EKAT_REQUIRE_MSG(name=="tracers" || (ftype==0 && name=="tracers_prev"),
+  // const int ftype = get_homme_param<int>("ftype");
```
jeff-cohere, line 185: Any reason to keep this comment?

jeff-cohere (approved):
Nice work, Luca. This PR does clear things up quite a bit.

Since we're introducing and clarifying more concepts (groups, importing, etc), it would be good if we wrote down a description of these concepts and their interaction somewhere (Confluence, maybe, for the moment?). This stuff needs to go into whatever we put together for a "SCREAM developer guide". I know Peter C has mentioned the need for documentation in recent days. Maybe we're getting to the point where we need to think about it more seriously.

components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
bartgol: Nope.

bartgol:
@jeff-cohere I agree. I will make this my next todo list item. Once the PR is in (and the homme forcing fix PR in e3sm is up), I will carve time to write down some documentation. Pinky promise. ;)

jeff-cohere:
> @jeff-cohere I agree. I will make this my next todo list item. Once the PR is in (and the homme forcing fix PR in e3sm is up), I will carve time to write down some documentation. Pinky promise. ;)

Let me know how I can help. You don't have to document all of this yourself!

bartgol:
> > @jeff-cohere I agree. I will make this my next todo list item. Once the PR is in (and the homme forcing fix PR in e3sm is up), I will carve time to write down some documentation. Pinky promise. ;)
> 
> Let me know how I can help. You don't have to document all of this yourself!

When I start, I will ping you, and we can see if we can thread the execution.

jeff-cohere (approved):
Looks good!

tcclevenger (approved):
Nice work!

components/scream/src/control/tests/ad_tests.yaml
```diff
@@ -4,20 +4,21 @@ Debug:
...
-  A: 1.0
-  B: 2.0
-  C: 3.0
-  D: [4.0, 5.0]
-  E: "A"
+  Point Grid:
+    Filename: should_not_be_neeeded_and_code_will_throw_if_it_tries_to_open_this_nonexistent_file.nc
+    A: 1.0
+    B: 2.0
+    C: 3.0
```
AaronDonahue, line 11: Not relevant to this PR, but I'm mentioning now because it just reminded me.  The YAML parser interprets any number starting with `-` as a string not a double.  So it's impossible to assign a negative value in this way.  We may want to fix that at some point.

jeff-cohere: Do you mean that if you tell the parser to interpret negative numbers as doubles, that operation fails? The parser can interpret anything as a string, and you tell it to cast to a specific data type, right?

jeff-cohere: Or are we using something on top of yaml-cpp to parse our YAML files?

components/scream/src/physics/p3/atmosphere_microphysics.cpp
```diff
@@ -179,21 +179,21 @@ void P3Microphysics::init_buffers(const ATMBufferManager &buffer_manager)
...
-  m_fields_out["qc"].add_property_check(positivity_check);
-  m_fields_out["qr"].add_property_check(positivity_check);
-  m_fields_out["qi"].add_property_check(positivity_check);
-  m_fields_out["qm"].add_property_check(positivity_check);
-  m_fields_out["nc"].add_property_check(positivity_check);
-  m_fields_out["nr"].add_property_check(positivity_check);
-  m_fields_out["ni"].add_property_check(positivity_check);
-  m_fields_out["bm"].add_property_check(positivity_check);
+  // auto positivity_check = std::make_shared<FieldPositivityCheck<Real> >();
+  // get_field_out("qv").add_property_check(positivity_check);
```
AaronDonahue, line 186: Did we mean to keep these commented out?

components/scream/src/physics/shoc/atmosphere_macrophysics.cpp
```diff
@@ -350,11 +346,11 @@ void SHOCMacrophysics::initialize_impl (const util::TimeStamp& t0)
   // Set field property checks for the fields in this process
   auto T_interval_check = std::make_shared<FieldWithinIntervalCheck<Real> >(150, 500);
   auto positivity_check = std::make_shared<FieldPositivityCheck<Real> >();
-  m_fields_out["T_mid"].add_property_check(T_interval_check);
-  m_fields_out["qv"].add_property_check(positivity_check);
-  m_fields_out["qc"].add_property_check(positivity_check);
-  m_fields_out["tke"].add_property_check(positivity_check);
-  m_fields_out["Q"].add_property_check(positivity_check);
+  get_field_out("T_mid").add_property_check(T_interval_check);
+  // get_field_out("qv").add_property_check(positivity_check);
```
AaronDonahue, line 350: Same question here, did we mean to keep these commented out?

components/scream/src/control/tests/ad_tests.yaml
bartgol: Really?!? Wow, I didn't know that. There should be a way to fix this.

components/scream/src/physics/p3/atmosphere_microphysics.cpp
bartgol: I guess I can remove them. I wasn't sure what to do with the FPC's in Homme, so I commented for now. You are working on fixing these checks, right? Does it bother you to keep them in master just for a couple more days, till you get your PR ready?

components/scream/src/control/tests/ad_tests.yaml
bartgol: Interesting. In ekat, the unit test for yaml parser correctly reads stuff like `My Int: -2` as a negative int. But `-2.0` is read as a string. I guess I should fix this right away in ekat.

components/scream/src/share/tests/field_tests.cpp
```diff
@@ -530,33 +530,33 @@ TEST_CASE("multiple_bundles") {
...
   FieldRequest d_req(d_id,SL{"group1","group3"});
   FieldRequest e_req(e_id,SL{"group1","group2"});
-  FieldRequest f_req(f_id,SL{"group1","group3"});
+  FieldRequest f_req(f_id,SL{"group1"});
 
   GroupRequest g1_req ("group1",grid_name,Bundling::Required);
   GroupRequest g2_req ("group2",grid_name,Bundling::Required);
-  // group3 = group3 + group2
-  GroupRequest g3_req ("group3",grid_name,4,Bundling::Required,&g2_req,Relationship::Child);
+  // group3 += group2
```
AaronDonahue, line 542: Do we mean to keep this commented line?

bartgol: Ah yeah. This line is simply explaining how group3 is "extended" by the following group request. Maybe I can improve the comment.

AaronDonahue:
Looks good @bartgol .  As for the property checks, I think we can revisit the ones set up in each individual process.  I envisioned that individual atm. processes may have had some strict criteria for incoming or outgoing fields so we would expect to see those property checks added in each process.  Also, for simple testing we can't rely on another process to add a property check.  Consider for example the negative tracers property, if we are running SHOC stand-alone we would still want to catch those.  I think we should approve this for now and add a v1.1 task to figure out how we want to give individual processes the power to check fields of interest.

AaronDonahue (approved):

bartgol:
> Looks good @bartgol . As for the property checks, I think we can revisit the ones set up in each individual process. I envisioned that individual atm. processes may have had some strict criteria for incoming or outgoing fields so we would expect to see those property checks added in each process. Also, for simple testing we can't rely on another process to add a property check. Consider for example the negative tracers property, if we are running SHOC stand-alone we would still want to catch those. I think we should approve this for now and add a v1.1 task to figure out how we want to give individual processes the power to check fields of interest.

I agree. Each process should store FPC's _internally_, and run them before/after run. The FPC's stored in the field should only be used if we foresee the need to run said check before/after all (or at least many) atm procs.

PR #1182, opened 2021-09-10
Quick fix to ensure that quartz can run multi-rank jobs

AaronDonahue:


AaronDonahue:
@bartgol or @jgfouca , could either of you make sure this change makes sense to get around the `srun` issue on quartz with binding mpi ranks.  This was the fix that Luca suggested and I tested it on quartz and it worked.  But always good to have a fresh set of eyes in case i missed something.

bartgol (approved):

PR #1183, opened 2021-09-13
Setup "reparing" instances of negative tracer mass after the dynamics step

AaronDonahue:
- Repairing tracers after dynamics

This addresses the issue where some tracers appear to be negative after dynamics, albeit essentially well within epsilon of 0.0.  Here we clip these values.

 - Make it so that the dynamics grid reports lat/lon in units of degrees
to the grids manager.

This is needed because everywhere else in the code lat/lon is assumed to be in degrees.  The immediate impact is that rrtmgp is currently calculating the zenith angle assuming lat/lon is in degrees.  So in the Homme+physics test rrtmgp has no daylight columns.

 - Clean up the call to tracers_forcing in the dynamics interface.

This is needed to ensure that ftype=2 is a valid option.

### DIFF
--- a/components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
+++ b/components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
@@ -818,29 +818,6 @@ void HommeDynamics::import_initial_conditions () {
 // =========================================================================================
 void HommeDynamics::
 check_computed_fields_impl () {
-//ASD  using KT = KokkosTypes<DefaultDevice>;
-//ASD  constexpr int N = sizeof(Homme::Scalar) / sizeof(Real);
-//ASD  using Pack = ekat::Pack<Real,N>;
-//ASD  using ColOps = ColumnOps<DefaultDevice,Real>;
-//ASD
-//ASD  const auto ncols = m_ref_grid->get_num_local_dofs();
-//ASD  const auto nlevs = m_ref_grid->get_num_vertical_levels();
-//ASD  const auto npacks= ekat::PackInfo<N>::num_packs(nlevs);
-//ASD
-//ASD  const auto& Q_view  = m_ref_grid_fields.at("Q").get_view<Pack***>();
-//ASD
-//ASD  using ESU = ekat::ExeSpaceUtils<KT::ExeSpace>;
-//ASD  const auto policy = ESU::get_thread_range_parallel_scan_team_policy(ncols,Q_view.extent(1));
-//ASD
-//ASD  Kokkos::parallel_for(policy, KOKKOS_LAMBDA (const KT::MemberType& team) {
-//ASD    const int& icol = team.league_rank();
-//ASD    Kokkos::parallel_for(Kokkos::TeamThreadRange(team,Q_view.extent(1)),
-//ASD                         [&](const int qidx) {
-//ASD      auto q_chk  = ekat::subview(Q_view,icol,qidx);
-//ASD      
-//ASD    });
-//ASD    team.team_barrier();
-//ASD  });
     auto& field = m_fields_out["Q"];
     for (auto& pc : field.get_property_checks()) {
       if (!pc.check(field) and pc.can_repair()) {
--- a/components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
+++ b/components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
@@ -27,6 +27,7 @@
 #include "physics/share/physics_constants.hpp"
 #include "share/util/scream_common_physics_functions.hpp"
 #include "share/util//scream_column_ops.hpp"
+#include "share/field/field_property_checks/field_lower_bound_check.hpp"
 
 // Ekat includes
 #include "ekat/ekat_assert.hpp"
@@ -818,12 +819,27 @@ void HommeDynamics::import_initial_conditions () {
 // =========================================================================================
 void HommeDynamics::
 check_computed_fields_impl () {
-    auto& field = m_fields_out["Q"];
-    for (auto& pc : field.get_property_checks()) {
-      if (!pc.check(field) and pc.can_repair()) {
-        pc.repair(field);
-      }
-    }
+    // Note: We are seeing near epsilon negative values in a handful of places,
+    // The strategy is to 
+    // 1. First check that no values are sufficiently negative as to require an error.
+    //    i.e. below some tolerance.
+    // 2. Clip all negative values to zero.
+    // TODO: Construct a more robust check that compares the value of Q against an
+    // average value or maximum value over each column.  That way we can use a relative
+    // error as our threshold, rather than an arbitrary tolerance.
+   
+    // Grab the pointer to the tracer group. 
+    const auto& rgn = m_ref_grid->name();
+          auto& Q   = *get_group_out("Q",rgn).m_bundle;
+    // Create a local copy of a lower bound check to ensure we are not encountering truly
+    // bad negative tracer values.
+    Real tol = -1e-20;
+    auto lower_bound_check = FieldLowerBoundCheck<Real>(tol);
+    lower_bound_check.check(Q);
+    // Now repair negative tracers using a lower bounds check at 0.0
+    auto lower_bound_repair = FieldLowerBoundCheck<Real>(0.0);
+    lower_bound_repair.repair(Q);
+
   
 }
 // =========================================================================================
--- a/components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
+++ b/components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
@@ -455,21 +455,7 @@ void HommeDynamics::homme_pre_process (const Real dt) {
       break;
     case ForcingAlg::FORCING_2:
       // Hard adjustment of qdp
-      Kokkos::parallel_for(Kokkos::RangePolicy<>(0,Q.size()),KOKKOS_LAMBDA(const int idx) {
-        const int ie = idx / (qsize*NP*NP*NVL);
-        const int iq = (idx / (NP*NP*NVL)) % qsize;
-        const int ip = (idx / (NP*NVL)) % NP;
-        const int jp = (idx / NVL) % NP;
-        const int k  =  idx % NVL;
-
-        // fq is currently storing q_new
-              auto& qdp = Qdp(ie,n0_qdp,iq,ip,jp,k);
-        const auto& dp  = dp3d(ie,n0,ip,jp,k);
-        const auto& fq  = FQ(ie,iq,ip,jp,k);
-
-        qdp = fq*dp;
-      });
-      Kokkos::fence();
+      ff.tracers_forcing(dt,n0,n0_qdp,true,params.moisture);
       break;
     default:
       EKAT_ERROR_MSG ("Error! Unexpected/unsupported forcing algorithm.\n"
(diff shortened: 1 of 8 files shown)


AaronDonahue:
@bartgol , this PR has the fix to ftype=2 and "repairs" the O(eps) negative tracers we were observing.  It passed the AT so I think it's ready for review when you have a chance.

components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
```diff
@@ -516,7 +516,7 @@ void HommeDynamics::initialize_impl (const util::TimeStamp& /* t0 */)
       m_ref_grid_fields.at("w_int").get_header().get_tracking().get_providers().size()==1,
       "Error! Someone other than Dynamics is trying to update the vertical velocity.\n");
 
-  auto positivity_check = std::make_shared<FieldPositivityCheck<Real> >();
+  auto positivity_check = std::make_shared<FieldPositivityCheck<Real> >(0.0);
```
bartgol, line 519: We should add a comment here, that clarifies that we are repairing the field if needed.

Also, we need to make sure we are not clipping stuff that is not morally zero. We know that _right now_ the negative values are benign, but once this clipping goes in, we may not spot others.

I think we should do something better here. I see 2 options:

1) make the check `f >-tol`, with tol not too big, to make it menaingful, and not too small to make it throw with not-so-small benign neg values. Do not allow repair. Then add a second check that does simple positivity check, and allow repair.
2) implement a more robust check, which enforces `f(icol,:) > -10*eps*abs(max(f(icol,:)))`, and also can repair.

The latter is more robust, and we should be the long term solution. Even if we manually rolled it in the atm dyn interface (without writing a FPC class). However, the former is probably faster to implement. Notice that the 1st is "safe", since checks are internally stored in the order in which they are added, so we can always enforce that a FPC is run before another, by adding it first (if FPC's were stored as an ordered container, such as `std::set`, this would not be possible).

tl;dr: can we add a FieldWithinIntervalCheck _before_ the positivity check, with lower bound `-1e-10`, upper bound `std;:numeric_limits<Real>::max()`, and `can_repair=false`?

components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
```diff
@@ -815,5 +801,16 @@ create_dyn_field (const std::string& name,
   m_dyn_grid_fields[name] = f;
 }
 // =========================================================================================
+void HommeDynamics::
+check_computed_fields_impl () {
+    auto& field = m_fields_out["Q"];
+    for (auto& pc : field.get_property_checks()) {
+      if (!pc.check(field) and pc.can_repair()) {
```
bartgol, line 808: This is repairing _all_ prop checks that fail, rather than just the one that we know is dynamics fault. Do we want this?

Suggestion: what if we store the FPC in HommeDynamics, and simply run it on `m_fields_out["Q"]`? I would still store a FPC in the field (a FieldWithinIntervalChek, to be precise), without possibility to repair. Then store a second FPC in HommeDynamics (a FieldPositivityCheck) and run it on Q, repairing neg values.

components/scream/src/dynamics/homme/interface/homme_grid_mod.F90
```diff
@@ -118,8 +119,8 @@ subroutine get_dyn_grid_data_f90 (gids_ptr, elgpgp_ptr, lat_ptr, lon_ptr) bind(c
           elgpgp(3,idof) = ip-1
 
           gids(idof) = INT(elem(ie)%gdofP(ip,jp),kind=c_int)
-          lat(idof)  = elem(ie)%spherep(ip,jp)%lat
-          lon(idof)  = elem(ie)%spherep(ip,jp)%lon
+          lat(idof)  = elem(ie)%spherep(ip,jp)%lat * 180.0/pi
```
bartgol, line 122: Did you intend to bring the radians->degrees change in this PR? Or was it by accident?

bartgol (changes requested):

components/scream/src/dynamics/homme/interface/homme_grid_mod.F90
AaronDonahue: I did, I made sure to explain why in the description of the PR.  I can change the title of the PR though so it doesn't get lost in the history where this change was made.

components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
AaronDonahue: Good idea.  I like option 2, but it seems like option 1 would be the fastest solution for now.  I can go with that but put in a comment about how we would like to implement option 2 eventually.

components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
AaronDonahue: It would repair all the ones that "can repair".  But yes, arguably we may want to allow for a different tracer property check that can only be repaired somewhere else.

I like your suggestion of a local FPC that we store in the interface and then run locally on Q instead though.  It's cleaner, and avoids us accidentally repairing something we shouldn't later down the road.

components/scream/src/dynamics/homme/interface/homme_grid_mod.F90
bartgol: No, I should just learn to fully read PRs description before starting the review... :)

components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
bartgol: Right. The only reason to store FPC's inside Field's, is to allow _other_ atm procs to run the checks. Perhaps only atm proc A knows the correct limits of a field, so it sets up the check, and everyone can then run it.

But in this case, we _don't_ want other atm procs to run this check. It should be a very local quick fix, to work around finite precision issues...

AaronDonahue:
@bartgol , I think the latest commits address your comments.  I've introduced a lower and upper bound field check and use the two part strategy you recommended where we first check that we don't have anything less than a negative tolerance, and then repair the values to 0.0.

I've also put in a TODO comment to apply your more robust recommendation down the road.

components/scream/src/share/field/field_property_checks/field_lower_bound_check.hpp
```diff
@@ -0,0 +1,52 @@
...
+
+  // Overrides.
+
+  // The name of the field check
+  std::string name () const override { return "Lower Bound Check of " + std::to_string(m_lower_bound); }
+
+  protected:
+
+  // Lower bound
+  RealType m_lower_bound;
```
bartgol, line 46: Why do you store this variable? You already have a `m_lower_bound` in the base class, you should just use that.

components/scream/src/share/field/field_property_checks/field_upper_bound_check.hpp
```diff
@@ -0,0 +1,50 @@
...
+
+  // Overrides.
+
+  // The name of the field check
+  std::string name () const override { return "Upper Bound Check of " + std::to_string(m_upper_bound); }
+
+protected:
+
+  // Upper bound
+  RealType m_upper_bound;
```
bartgol, line 45: Same here: do not store another upper bound, simply use the one in the base class.

components/scream/src/share/tests/field_tests.cpp
```diff
@@ -727,6 +729,82 @@ TEST_CASE("field_property_check", "") {
       REQUIRE(p.check(f1));
     }
   }
+
+  // Check that the values of a field are above a lower bound
+  SECTION ("field_lower_bound_check") {
+    Field<Real> f1(fid);
+    auto lower_bound_check = std::make_shared<FieldLowerBoundCheck<Real> >(-1.0);
+    REQUIRE(lower_bound_check->can_repair());
+    f1.add_property_check(lower_bound_check);
```
bartgol, line 738: Since we're just checking the new FPC, let's keep it simple: let's not store the FPC in the field, and avoid the for loop below (see next comment).

components/scream/src/share/tests/field_tests.cpp
```diff
@@ -727,6 +729,82 @@ TEST_CASE("field_property_check", "") {
...
+    f1.allocate_view();
+    const int num_reals = f1.get_header().get_alloc_properties().get_num_scalars();
+
+    // Assign in-bound values to the field and make sure it passes the lower_bound check
+    auto f1_data = f1.get_internal_view_data<Host>();
+    for (int i = 0; i<num_reals; ++i) {
+      f1_data[i] = std::numeric_limits<Real>::max() - i*1.0; 
+    }
+    f1.sync_to_dev();
+    for (auto& p : f1.get_property_checks()) {
```
bartgol, line 748: Here: no need for the for loop. Simply do

```
f1.sync_to_dev();
REQUIRE(lower_bound_check->check(f1));
```

components/scream/src/share/tests/field_tests.cpp
```diff
@@ -727,6 +729,82 @@ TEST_CASE("field_property_check", "") {
...
+      REQUIRE(p.check(f1));
+    }
+
+    // Assign out-of-bounds values to the field, make sure it fails the check,
+    // and then repair the field so it passes.
+    for (int i = 0; i<num_reals; ++i) {
+      f1_data[i] = -2.0*(i+1);
+    }
+    f1.sync_to_dev();
+    for (auto& p : f1.get_property_checks()) {
```
bartgol, line 758: Here too: no loop, just manually invoke the methods on `lower_bound_check`.

components/scream/src/share/tests/field_tests.cpp
```diff
@@ -727,6 +729,82 @@ TEST_CASE("field_property_check", "") {
...
+      REQUIRE(f1_data[i] == -1.0);
+    }
+  }
+
+  // Check that the values of a field are above below an upper bound
+  SECTION ("field_upper_bound_check") {
+    Field<Real> f1(fid);
+    auto upper_bound_check = std::make_shared<FieldUpperBoundCheck<Real> >(1.0);
+    REQUIRE(upper_bound_check->can_repair());
+    f1.add_property_check(upper_bound_check);
```
bartgol, line 776: Same thing as for the lower bound check.

bartgol (changes requested):
Mostly good. I have just a couple of requests.

components/scream/src/share/field/field_property_checks/field_lower_bound_check.hpp
AaronDonahue: I ran into issues accessing the lower bound in the base class for the `name` override above.  If it's not important I can just simplify the name override and remove this variable.  Same for the upper bound check.

components/scream/src/share/tests/field_tests.cpp
AaronDonahue: Sounds good

AaronDonahue:
@bartgol , thanks for the extra comments to clean up the code.  I've made the changes and am ready for a re-review.

components/scream/src/share/field/field_property_checks/field_lower_bound_check.hpp
```diff
@@ -0,0 +1,46 @@
...
+// This field property check returns true if all data in a given field are
+// bounded below by the given lower bounds, and false if not.
+// It can repair a field that fails the check by clipping all unbounded values
+// to the lower bound.
+// Note: The lower bound check is a sub-class of the within interval check with
+// the upper bound set to the numeric limit.
+template<typename RealType>
+class FieldLowerBoundCheck: public FieldWithinIntervalCheck<RealType> {
+public:
+  using non_const_RT = typename FieldPropertyCheck<RealType>::non_const_RT;
```
bartgol, line 20: This typedeff is unused, and can be removed.

components/scream/src/share/field/field_property_checks/field_upper_bound_check.hpp
```diff
@@ -0,0 +1,46 @@
...
+// This field property check returns true if all data in a given field are
+// bounded above by the given upper bound, and false if not.
+// It can repair a field that fails the check by clipping all unbounded values
+// to the upper bound.
+// Note: The upper bound check is a sub-class of the within interval check with
+// the upper bound set to the numeric limit.
+template<typename RealType>
+class FieldUpperBoundCheck: public FieldWithinIntervalCheck<RealType> {
+public:
+  using non_const_RT = typename FieldPropertyCheck<RealType>::non_const_RT;
```
bartgol, line 20: Unused.

bartgol (changes requested):

bartgol (approved):

PR #1186, opened 2021-09-14
YAKL should only send messages for one MPI rank

brhillman:
Update to latest YAKL, which implements a fix so that only one MPI rank sends messages on initialization. A number of recent updates to YAKL are also included, which required updating the RRTMGP submodule as well.


tcclevenger (approved):
Thanks for doing this, Ben!

PR #1190, opened 2021-09-16
hotfix to lassen cmake file to make jsrun command work after LC policy change

AaronDonahue:


AaronDonahue:
This was just a hotfix to an LC machine so I went ahead and merged it

PR #1191, opened 2021-09-16
Add needed team_barrier()

tcclevenger:
Running through CIME, again I saw the issues with `OMP_NUM_THREAD > 1`. Still not sure why this bug is not caught in the scream coupled/uncoupled tests.

### DIFF
--- a/components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
+++ b/components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
@@ -516,6 +516,7 @@ void HommeDynamics::homme_post_process () {
     auto p_int = ekat::subview(p_int_view,icol);
 
     ColOps::column_scan<true>(team,nlevs,dp,p_int,ps0);
+    team.team_barrier();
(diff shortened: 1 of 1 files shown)


bartgol (approved):
I might have brought this back with the field group refactor PR. I had some conflicts in this file, and probably messed up this thing (which I thought you already fixed).

Re: why this doesn't show up in our unit tests. Our AD tests are all single threaded. The Homme tests are a tad long, so I don't want to add another copy of them. Perhaps we can switch the homme+physics test to N threads (on CPU)?

tcclevenger:
> I might have brought this back with the field group refactor PR. I had some conflicts in this file, and probably messed up this thing (which I thought you already fixed).
> 
> Re: why this doesn't show up in our unit tests. Our AD tests are all single threaded. The Homme tests are a tad long, so I don't want to add another copy of them. Perhaps we can switch the homme+physics test to N threads (on CPU)?

Ahh ok, that makes me feel better that they are not threaded. Yeah, maybe it would be a good idea to have that one theaded, since it has already caused some issues.

PR #1192, opened 2021-09-16
Switch the default ftype for the homme+physics test back to 0.

AaronDonahue:
We will likely switch back to 2 in the future, but there is a bugfix
working it's way through homme and e3sm.  So this is a temporary change
to get back to having a homme+physics test that works for at least a few
steps.


bartgol (approved):

AaronDonahue:
Huh, not sure why the `release` test on blake failed for the `ColumnOps` test.  I'm going to try to retest.

PR #1196, opened 2021-09-16
More print only rank0

tcclevenger:
More fixes for only output on rank 0. Requires ekat to update.


tcclevenger:
Some reasons having issues with updating ekat. Just made a new PR for convenience.

PR #1198, opened 2021-09-16
More print only rank0 and removing virtual get_comm()

tcclevenger:
More fixes for only output on rank 0. Requires ekat to update.

### DIFF
--- a/components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
+++ b/components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
@@ -37,8 +37,7 @@ namespace scream
 {
 
 HommeDynamics::HommeDynamics (const ekat::Comm& comm, const ekat::ParameterList& params)
- : m_params        (params)
- , m_dynamics_comm (comm)
+  : AtmosphereProcess(comm, params)
 {
   // Nothing to do here
 }
--- a/components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
+++ b/components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
@@ -287,7 +287,7 @@ void HommeDynamics::initialize_impl (const util::TimeStamp& /* t0 */)
   const auto& Q_dyn_track = get_group_out("Q",dgn).m_bundle->get_header().get_tracking();
   for (auto wp : Q_dyn_track.get_providers()) { 
     auto sp = wp.lock();
-    printf("Q is provided by: %s\n",sp->name().c_str());
+    if (get_comm().am_i_root()) printf("Q is provided by: %s\n",sp->name().c_str());
   }
   EKAT_REQUIRE_MSG (
       rho_track.get_providers().size()==1,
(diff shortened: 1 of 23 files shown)


components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
```diff
@@ -287,7 +287,7 @@ void HommeDynamics::initialize_impl (const util::TimeStamp& /* t0 */)
   const auto& Q_dyn_track = get_group_out("Q",dgn).m_bundle->get_header().get_tracking();
   for (auto wp : Q_dyn_track.get_providers()) { 
     auto sp = wp.lock();
-    printf("Q is provided by: %s\n",sp->name().c_str());
+    if (get_comm().am_i_root()) printf("Q is provided by: %s\n",sp->name().c_str());
```
bartgol, line 290: I think I added this for loop while I was debugging (trying to figure out who else was computing Q). Can you please remove it altogether?

bartgol (changes requested):
Looks good. I have a cleanup favor to ask though.

components/scream/src/mct_coupling/scream_cxx_f90_interface.cpp
```diff
@@ -64,6 +64,11 @@ const scream::control::AtmosphereDriver& get_ad () {
   return c.get<scream::control::AtmosphereDriver>();
 }
 
+const ekat::Comm& get_ekat_comm () {
```
bartgol, line 67: I think we should instead add `get_comm()` to the AtmosphereDriver class.

tcclevenger: Yeah, I think there is a comment somewhere about removing every instance of `m_[insert-process]_comm` and instead opt for a base class comm. Then we would have a `AtmosphereDriver::get_comm()`. I can do that as a part of this PR.

tcclevenger:
@bartgol I removed all the `m_[process name]_comm` and `m_[process name]_param` from `AtmosphereProcess` derived classes. Instead now there are 2 `AtmosphereProcess` constructors: default constructor and one that takes `comm, params` as input. Each derived class has the option to call 
```
DerivedClass (const ekat::Comm& comm, const ekat::ParameterList& params)
  : AtmosphereProcess(comm, params)
{ ... }
```
Now only `AtmosphereProcess` has `get_comm()`.

components/scream/src/share/atm_process/atmosphere_process.hpp
```diff
@@ -72,6 +72,12 @@ class AtmosphereProcess : public ekat::enable_shared_from_this<AtmosphereProcess
   template<typename T>
   using str_map = std::map<std::string,T>;
 
+  // Default base constructor
+  AtmosphereProcess () = default;
```
bartgol, line 76: Question: do we need/want a default ctor? Who uses it?

bartgol (commented):
Looks good. Only one question regarding atm proc ctor.

components/scream/src/share/atm_process/atmosphere_process.hpp
tcclevenger: No one currently. My only thought was if a process (or test) doesn't care about params and comms then they would only have to define their own constructor. But maybe this is not something that would happen. I can remove it.

bartgol: Uhm, I think devs of such an atm proc would not have a hard time creating dummy comm/params. OTOH, if someone accidentally uses the def ctor, there might be issues finding the bug. I think we can remove it, and if need be, add it back later.

bartgol:
I believe the failure we see on weaver is already in master (I manually verified that master can fail on weaver1). I will merge this.

bartgol (approved):

tcclevenger:
Thanks @bartgol!

PR #1202, opened 2021-09-17
Fix some cuda-memcheck errors for the phys-dyn remapper.

bartgol:
There were quite a few access to this->blah members. It seemed easier to replace the remap lambdas with functor-like calls.


tcclevenger (approved):
Looks great!

PR #1204, opened 2021-09-20
Add testing profiles feature

bartgol:
This allows us to lower the number of time steps for particular builds and/or on particular machines. E.g., for valgrind runs, 3 timesteps are probably enough to detect memory errors.


bartgol:
I actually ended up making this branch a bit fancier. Each test can now set some of its inputs depending on the currently enabled testing profile. For now, I defined 3 testing profiles: short, medium, long. Only one can be active in each build. The meaning of short/medium/long is completely up to the individual test. Each test can then do something like

```
SetVarDependingOnTestProfile(my_var 5 10 50)
```
which will result in `my_var` being set to 5, 10, or 50 if the test profile is 'short', 'medium', or 'long' respectively. This var can then be used inside any cmake configured file, like any of our input.yaml files or cmake-generated namelists (must not use `COPYONLY`!) to obtain a test with the desired level of intensity.

The default testing profile is MEDIUM. The few places where I added the profile feature use it to set the number o ftimesteps, with MEDIUM yielding whatever was currently used in master.

jgfouca (approved):
This looks great!

components/scream/tests/coupled/shoc_cld_p3_rrtmgp/shoc_cld_p3_rrtmgp.cpp
```diff
@@ -22,7 +22,7 @@ TEST_CASE("shoc-stand-alone", "") {
   REQUIRE_NOTHROW ( parse_yaml_file(fname,ad_params) );
 
   // run params:
-  const auto& num_iters = ad_params.get<int>("Number of Iterations",5);
+  const auto& num_iters = ad_params.get<int>("Number of Iterations");
```
AaronDonahue, line 25: Have you confirmed that we don't need this fix in all of the other AD tests?  I'm pretty sure I added this to each test, but maybe in an open PR?

bartgol: I just saw it, and removed it. I don't want tests that should use testing profile logic to use the wrong iter number just because of some typos.

bartgol: Other tests that do get `num_iters` from the parameter list do not use the testing profiles feature, so it doesn't matter. However, we _could_ make all tests (at least the ad ones) use the feature.

AaronDonahue: I like the idea of using this tool for as many tests as we can.  Of course it's most important to take care of the most expensive ones first.  I'm working on testing in a separate PR  so I could extend this tool to all the other tests there if we want to get this PR in ASAP.

AaronDonahue (approved):

PR #1206, opened 2021-09-21
SL and physgrid on GPU

ambrad:
Passes e3sm_scream against baselines on chrysalis.

Optional whether to merge this; we can also wait for the merge to upstream once the v2 tag is made.

[BFB]


ambrad:
@bartgol here's the SL/physgrid GPU PR if you end up wanting to merge it before it goes into E3SM. This PR fixes two trivial conflicts. It passes e3sm_scream against baselines on chrysalis.

PR #1209, opened 2021-09-22
scripts-tests upgrade

jgfouca:
Greatly speeds up scripts-test (2-3 hours down to 12 minutes) by using an environment variable (`SCREAM_FAKE_ONLY`) to reduce what gets built and tested. The only things needed to test the mechanics of our testing scripts is `scream_share` and `fake_test`. We also need a minimal P3 build to test diffing mechanics.

Add stubs for mpirank/thread spreading. @bartgol will fill these in once his work is done.

Add tests to check return codes for jenkins_scripts in both AT and nightly modes. Fix return code for nightlies to reflect if there were failures.


components/scream/scripts/scripts-tests
```diff
@@ -409,7 +409,10 @@ class TestTestAllScream(TestBaseOuter.TestBase):
         else:
             self.skipTest("Skipping full run")
 
-    def test_jenkins_process(self):
+    def test_mpirank_and_thread_spreading(self):
```
bartgol, line 412: What's the point of this one? Are we putting it there as a placeholder, with the idea of expanding it in a subsequent PR?

components/scream/src/share/CMakeLists.txt
```diff
@@ -1,44 +1,55 @@
...
-    COMMAND ${PROJECT_SOURCE_DIR}/scripts/cf-xml-to-yaml ${CF_CONVENTIONS_XML_FILE}
-    WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}/data
-    COMMENT "Downloading CF conventions XML file and building YAML dictionary...")
-endif()
+if (NOT $ENV{SCREAM_FAKE_ONLY})
+  # The "build_cf_dictionary" target downloads an XML file containing valid field
+  # names and aliases based on the CF conventions, and transforms it into a YAML
+  # file that we use to validate our own fields against it. This target is only
+  # available if CMake finds wget on your system.
+  find_program(WGET wget)
```
bartgol, line 39: Side note: it seems to me that this target should be part of our scripts folder, or maybe a "tools" folder. @jeff-cohere , do you recall why we put it in the share folder, building it as part of `make all`?

bartgol (approved):

components/scream/scripts/scripts-tests
jgfouca: > Add stubs for mpirank/thread spreading. @bartgol will fill these in once his work is done.

bartgol: lol, I really need to learn to read the PR message before starting the review... :)

components/scream/src/share/CMakeLists.txt
jeff-cohere: I agree with you that it should be in our scripts folder. I was probably just trying to get the CF field name stuff to work, and my first thought (which I didn't get past) was "well, we probably want to have CF field names available in general!"

bartgol: Ok, I'll add it to my todo list.

bartgol (approved):

bartgol (approved):

PR #1214, opened 2021-09-23
Fix env checking in CMake for SCREAM_FAKE_ONLY

jgfouca:


bartgol (approved):

bartgol:
This is very-high priority. Will merge immediately.

PR #1216, opened 2021-09-23
New compute_interface_values() impl

tcclevenger:
This PR changes the name/impl for `compute_interface_values()` to be `compute_interface_values_linear()`:
```
x_int(k) = [x_mid(k)*dz(k-1) + x_mid(k-1)*dz(k)]/[dz(k-1) + dz(k)]
```
with `x_int(0)` and `x_int(num_levs)` given by the user as BCs.

If `dz` is a constant view, then this is just the average of the surrounding `x_mid` values. In this case, `x_int(x_mid(x_int) = x_int` (given appropriate BCs).

I've changed the name of the old method to `compute_interface_values_compatible()`


tcclevenger:
@brhillman @AaronDonahue The failure is coming from `rrtmgp_standalone`:
```
REQUIRE( rrtmgpTest::all_close(lw_flux_up_ref , lw_flux_up_test , 0.001) )
```
with values `arr1 = 457.139174, arr2 = 455.938201`.

This must be because we are changing the values for `T_int`. I'm not sure how much change is acceptable, or how to go about changing the baselines if it is acceptable. Thoughts?

AaronDonahue:
@tcclevenger and @brhillman , thanks for both your work on this.  It looks like it passed the AT.  I'm going to assign reviewers so we can get this into master.

components/scream/src/control/tests/sc_tests.cpp
```diff
@@ -262,6 +262,7 @@ TEST_CASE ("recreate_mct_coupling")
   FID sfc_alb_dir_nir_id  ("sfc_alb_dir_nir",    scalar2d_layout, nondim,  grid_name);
   FID sfc_alb_dif_vis_id  ("sfc_alb_dif_vis",    scalar2d_layout, nondim,  grid_name);
   FID sfc_alb_dif_nir_id  ("sfc_alb_dif_nir",    scalar2d_layout, nondim,  grid_name);
+  FID surf_lw_flux_up_id  ("surf_lw_flux_up",    scalar2d_layout, nondim,  grid_name);
```
AaronDonahue, line 265: Just a note, I think there was some discussion about changing this to `sfc_lw_flux_up`.  Is that still what we want to do?  @brhillman ?

components/scream/src/share/tests/column_ops.cpp
```diff
@@ -131,7 +133,98 @@ TEST_CASE("column_ops_ps_1") {
...
+    // Run column kernel
+    Kokkos::parallel_for(policy,
+                         KOKKOS_LAMBDA(const member_type& team){
+      const int icol = team.league_rank();
+      auto v_i = ekat::subview(v_int,icol);
+      auto v_m = ekat::subview(v_mid,icol);
+      auto dz  = ekat::subview(dz_mid,icol);
+
+      column_ops::compute_midpoint_values(team,num_levs,v_i,v_m);
+      column_ops::compute_interface_values_linear(team,num_levs,v_m,dz,bc_top,bc_bot,v_i);
```
bartgol, line 216: I think you need a barrier between the two colum_ops calls, no?

bartgol (commented):
Looks good. I only have a couple of comments.

components/scream/src/share/tests/column_ops.cpp
tcclevenger: Yeah, you're right. Getting lucky with a pass again..

tcclevenger: I just added them

bartgol (approved):

components/scream/src/control/tests/sc_tests.cpp
brhillman: Just noting for future, yes I think we want to be consistent in naming. Will fix in a subsequent PR.

PR #1223, opened 2021-09-29
Add initial export after SC setup

tcclevenger:
In mct-coupling, directly after `SurfaceCoupling` is setup, do an initial `do_export()`, as Ice model requires the initial density from SCREAM before the first SCREAM run.

Other changes

- Adding needed rrtmgp data files to `data/` folder for CIME runs
- Update `data/scream_input.yaml` to include all processes, and remove unused inputs (maybe they are needed, but I comment them out in my CIME runs)
- Add `cubed_sphere_map` to homme interface params so that the value can be changed in input files


bartgol:
I don't see where this option is used and/or tested. Are you planning to add a test and to start using it in the mct interface?

tcclevenger:
@bartgol I'm using it in mct-coupling interface, but I should add a test in `sc_ut` as well..

bartgol:
> @bartgol I'm using it in mct-coupling interface, but I should add a test in `sc_ut` as well..

Is the use in the mct-coupling interface in a different PR? I can't find it here.

tcclevenger:
> > @bartgol I'm using it in mct-coupling interface, but I should add a test in `sc_ut` as well..
> 
> Is the use in the mct-coupling interface in a different PR? I can't find it here.

And that would be because I forgot to add the changes where I actually do the export.... Now it there.

tcclevenger:
@bartgol Still should add a test. I'll let you know when it's ready to be looked at.

tcclevenger:
@bartgol I added a test and changed the format a bit, namely, moved the decision to skip export fields to the `register_export()` call and added that info to `scream_cpl_indices.F90`

tcclevenger:
Setting to WIP. Strangely the data on device on Weaver doesn't seem to be set to -1 in the test (`std::fill_n(raw_data,ncols*num_fields,-1);`.. I'll have to investigate tomorrow.

tcclevenger:
@bartgol @AaronDonahue I added a test to `sc_ut` and added an `update_pressure()` function to the homme AD which calculates `p_mid` and `p_int` during initialization (so that the SC has something to export to ice).

components/scream/src/control/surface_coupling.cpp
```diff
@@ -303,6 +307,10 @@ void SurfaceCoupling::do_export ()
     });
   }
 
+  // Deep copy fields to device in case host data was updated
+  // in between registration_end() and do_export()
```
bartgol, line 311: I don't understand this: why copying host data to device, given that the exports views will be written on by the next pfor? I assume this has something to do with the init phase, but I can't understand it. Btw, assuming that's the case, shouldn't you do this only if `init_phase` is true?

components/scream/src/control/surface_coupling.hpp
```diff
@@ -56,7 +56,8 @@ class SurfaceCoupling {
                         const int vecComp = -1);
   void register_export (const std::string& fname,
                         const int cpl_idx,
-                        const int vecComp = -1);
+                        const int vecComp = -1,
+                        const bool export_during_init = true);
```
bartgol, line 60: Can we add a 2-line comment explaining why we have the `export_during_init` arg? Doesn't have to go into too much detail, just reminding the reader about mct needs.

components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
```diff
@@ -814,6 +819,40 @@ void HommeDynamics::import_initial_conditions () {
...
+
+  const auto dp_view    = get_field_out("pseudo_density").get_view<Pack**>();
+  const auto p_int_view = get_field_out("p_int").get_view<Pack**>();
+  const auto p_mid_view = get_field_out("p_mid").get_view<Pack**>();
+
+  using ESU = ekat::ExeSpaceUtils<KT::ExeSpace>;
+  const auto policy = ESU::get_thread_range_parallel_scan_team_policy(ncols,npacks);
+
+  Kokkos::parallel_for(policy, KOKKOS_LAMBDA (const KT::MemberType& team) {
+   const int& icol = team.league_rank();
```
bartgol, line 844: Indentation is off.

components/scream/src/dynamics/homme/interface/homme_params_mod.F90
```diff
@@ -118,6 +118,7 @@ subroutine init_params_f90 (nl_fname_c) bind(c)
     namelist /ctl_nl/test_case, &
       u_perturb,                &
       partmethod,               &         ! mesh partitioning method
+      cubed_sphere_map,         &
```
bartgol, line 121: Have you tried running a CIME case? The entry `cubed_sphere_map` is not in `data/scream_input.yaml`, which is used to generate the input namelist. I have still to understand namelist completely, but it's my understanding that if a namelist entry is not found in the input file, then random errors happen. Maybe I'm wrong, but if you could run a case through valgrind, and  verify that there are no errors, and the correct params are printed, that'd be great.

components/scream/src/mct_coupling/atm_comp_mct.F90
```diff
@@ -55,7 +55,8 @@ subroutine atm_init_mct( EClock, cdata, x2a, a2x, NLFilename )
                                   scream_init_atm
     use scream_cpl_indices, only: scream_set_cpl_indices, num_cpl_exports, &
                                   num_cpl_imports, num_scream_imports, &
-                                  scr_names_x2a, scr_names_a2x, index_x2a, index_a2x, vec_comp_x2a, vec_comp_a2x
+                                  scr_names_x2a, scr_names_a2x, index_x2a, index_a2x, vec_comp_x2a, vec_comp_a2x, &
+                                  can_be_exported_during_init
```
bartgol, line 59: Idea: can't we figure out (from scream) the fields that must be skipped during the init phase just by looking at the time stamp? Meaning, if we init a field (either during IC setup or during an atm proc init phase, like p_mid/p_int in Homme), we _should_ set the time stamp of the field to `t0`. If an export field has time stamp still set to invalid, then it cannot be exported (obv, during a _normal_ export, we must have no invalid field).

bartgol: On a second thought, this would blindly skip any field that has invalid timestamp at init phase, but we might want to be able to _ensure_ that the ones that _should_ be exported are exported. So probably not a great idea? I was trying to see if we can get the AD to automatically do stuff, without making the F90 interface heavier.

bartgol (commented):
Looks good. I have some comments/questions.

Note: Github tells me that two new NC files for rrtmgp in the data folder are quite big (31MB and 14MB). They are not huge, but they might make cloning the repo a tad slower on some system. Since this is not a PR that modifies rrtmgp, I'm wondering if these files were not meant to be included or if it's on purpose.

components/scream/src/control/surface_coupling.cpp
tcclevenger: The issue I was having is that, for the test, my cpl data all is -1. If I do not perform a deep copy here, that -1 data gets overwritten to 0 during the deep copy from device to host (only on GPU). So my test was failing because for the initial export, the fields I wasn't exporting I was expecting to be -1, but were 0.

2 things:
1. Will we even care about preserving cpl_data? Or is it fine to send things that are not touched to 0 in the export? I guess this is what we do for fields that aren't relevant to scream..
2. Even if that is not the case, the comment is still incorrect. registration_end() doesn't matter. This is just to preserve any existing data. Should only be done on the init phase. Probably unnecessary on CPU.

components/scream/src/dynamics/homme/interface/homme_params_mod.F90
tcclevenger: I added the entry into scream_input.yaml as a part of this PR and it works through CIME.

The question would be if it is ok for the standalone test (which don't include it in their .yaml files). I'll run those.

@singhbalwinder says that it is necessary for us to set this to 0, maybe he can elaborate here. I copied this from 
`scream/components/homme/src/share/namelist_mod.F90`

components/scream/src/control/surface_coupling.cpp
tcclevenger: I could just change the test actually to store the data before registration end...

tcclevenger: Is this even necessary then? Shouldn't `create_mirror_view()` do a deep copy if the spaces are different?

bartgol: It creates a view on the mirror space, but it doesn't perform the copy. But there is `create_mirror_view_and_copy`, which maybe is what you want?

components/scream/src/mct_coupling/atm_comp_mct.F90
bartgol: Well, in general, if a quantity is not initialized by scream, it can't be exported. So simply looking at the time stamp might be a valid approach (perhaps we could print a msg to screen when we skip a field during the init phase).

However, this would be a bit cumbersome. We can punt on this, and polish later, if we find a decent solution.

components/scream/src/control/surface_coupling.cpp
bartgol: You could move the creation/setup of the SurfaceCoupling in the unit test _inside_ the nruns for loop... Not pretty, maybe, but for such small tests, it's not expensive anyways...

bartgol:
> @bartgol Regarding the files: there are 4 rrtmgp ones I had to add. They are referenced in the rrtmgp interface
> 
> ```
>         /*
>          * Names of input files we will need.
>          */
>         std::string coefficients_file_sw = "./data/rrtmgp-data-sw-g224-2018-12-04.nc";
>         std::string coefficients_file_lw = "./data/rrtmgp-data-lw-g256-2018-12-04.nc";
>         std::string cloud_optics_file_sw = "./data/rrtmgp-cloud-optics-coeffs-sw.nc";
>         std::string cloud_optics_file_lw = "./data/rrtmgp-cloud-optics-coeffs-lw.nc";
> ```
> 
> The tests use CMakeLists to copy these files. I was struggling to get MCT to do the same.. So copying them in worked. But obviously not ideal. I can remove them from this PR and work on CMake doing the job.

Don't sweat on it. Keep them in for now.

components/scream/src/dynamics/homme/interface/homme_params_mod.F90
```diff
@@ -175,6 +176,7 @@ subroutine init_params_f90 (nl_fname_c) bind(c)
     ndays =  0
     nmax  = 12
     moisture = 'dry'
+    cubed_sphere_map = 2
```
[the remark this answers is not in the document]
tcclevenger: This is a bit lower in the file
```
! set map
    if (cubed_sphere_map<0) then
       cubed_sphere_map=2  ! theta model default = element local
    endif
```

PR #1224, opened 2021-09-29
Add cuda-memcheck nightly to weaver

jgfouca:


components/scream/scripts/test_all_scream.py
```diff
@@ -234,6 +236,8 @@ def __init__(self, cxx_compiler=None, f90_compiler=None, c_compiler=None,
             "opt" : [("CMAKE_BUILD_TYPE", "Release")],
             "valg" : [("CMAKE_BUILD_TYPE", "Debug"),
                       ("EKAT_ENABLE_VALGRIND", "True")],
+            "cmc"  : [("CMAKE_BUILD_TYPE", "Debug"),
+                      ("EKAT_ENABLE_CUDA_MEMCHECK", "True")],
```
bartgol, line 240: I think you need to update ekat in your branch. I don't see this option in the version of ekat in this branch.

bartgol (changes requested):
Looks good. But I think it needs an ekat update.

components/scream/scripts/test_all_scream.py
jgfouca: Good call. I thought it was already in our submodule for some reason.

bartgol (approved):

PR #1225, opened 2021-09-29
add scripts ctest driver

jgfouca:
will allow for nightly testing + cdash submission

### DIFF
--- a/components/scream/cmake/ctest_script_scripts_tests.cmake
+++ b/components/scream/cmake/ctest_script_scripts_tests.cmake
@@ -1,0 +1,42 @@
+cmake_minimum_required(VERSION 3.9)
+
+set(CTEST_BUILD_NAME "scripts_tests")
+
+set(CTEST_SOURCE_DIRECTORY "${SCREAM_ROOT}/scripts/tests")
+set(CTEST_BINARY_DIRECTORY "${BUILD_WORK_DIR}")
+
+if(NOT DEFINED dashboard_model)
+  set(dashboard_model Experimental)
+endif()
+if(NOT DEFINED dashboard_track)
+  set(dashboard_track E3SM_SCREAM)
+endif()
+
+set(CTEST_CMAKE_GENERATOR "Unix Makefiles")
+
+ctest_start(${dashboard_model} TRACK ${dashboard_track})
+
+ctest_configure(RETURN_VALUE CONFIG_ERROR_CODE)
+
+if (CONFIG_ERROR_CODE)
+  set (TEST_FAILS TRUE)
+else ()
+
+  # Need this code so that build errors don't get buried
+  if (DEFINED ENV{CTEST_PARALLEL_LEVEL})
+    ctest_test(PARALLEL_LEVEL $ENV{CTEST_PARALLEL_LEVEL} RETURN_VALUE TEST_ERROR_CODE)
+  else()
+    ctest_test(RETURN_VALUE TEST_ERROR_CODE)
+  endif()
+  if (TEST_ERROR_CODE)
+    set(TEST_FAILS TRUE)
+  endif()
+endif ()
+
+if (NOT NO_SUBMIT)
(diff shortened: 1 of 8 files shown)


components/scream/cmake/ctest_script_scripts_tests.cmake
```diff
@@ -0,0 +1,42 @@
...
+ctest_start(${dashboard_model} TRACK ${dashboard_track})
+
+ctest_configure(RETURN_VALUE CONFIG_ERROR_CODE)
+
+if (CONFIG_ERROR_CODE)
+  set (TEST_FAILS TRUE)
+else ()
+
+  # Need this code so that build errors don't get buried
+  if (DEFINED ENV{CTEST_PARALLEL_LEVEL})
```
bartgol, line 26: Since we don't use parallel level (we hard code 1), can we remove the CTEST_PARALLEL_LEVEL logic altogether?

components/scream/scripts/scripts_ctest_driver.py
```diff
@@ -0,0 +1,81 @@
...
+            shutil.rmtree(str(self._work_dir))
+
+        self._work_dir.mkdir(parents=True)
+
+        setup_mach_env(self._machine)
+
+    ###############################################################################
+    def generate_ctest_config(self, extra_configs):
+    ###############################################################################
+        result = "CTEST_PARALLEL_LEVEL=1 CIME_MACHINE={} ".format(self._machine)
```
bartgol, line 57: Here too: no point in setting CTEST_PARALLEL_LEVEL=1, since we later hard code `-j1` anyways.

components/scream/scripts/tests/CTestConfig.cmake
```diff
@@ -0,0 +1 @@
+../../CTestConfig.cmake
```
bartgol, line 1: Not familiar with how ctest/cmake parse this file, but shouldn't this be something like `include(../../CTestConfig.cmake)`?

components/scream/scripts/tests/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+cmake_minimum_required(VERSION 3.3)
+cmake_policy(SET CMP0057 NEW)
+
+project(SCRIPTS_TESTS)
+
+enable_testing()
+include(CTest)
+
+add_test(NAME scripts_tests_tas COMMAND ${CMAKE_CURRENT_SOURCE_DIR}/../gather-all-data -l -m $ENV{CIME_MACHINE} "${CMAKE_CURRENT_SOURCE_DIR}/../scripts-tests -f -m \$machine TestTestAllScream")
```
bartgol, line 9: Why are there two tests? Are they always both run?

jgfouca: One has to be run with gather-all-data to ensure a compute node, the other has to be on a login node.

components/scream/scripts/tests/CTestConfig.cmake
jgfouca: I think GitHub is just telling you this is a symlink.

components/scream/cmake/ctest_script_scripts_tests.cmake
jgfouca: Good call, this was overriding the `-j1`

components/scream/scripts/tests/CTestConfig.cmake
bartgol: Ah you're right. It even said it when I clicked on "View File", but I missed it.

components/scream/scripts/tests/CMakeLists.txt
bartgol: And we need to run both? Shouldn't one run either or, depending on whether it's already on a compute node?

bartgol: Clarified by Jim offline: we need both.

bartgol (approved):

PR #1226, opened 2021-09-29
Remove subgrid enhancement information from SHOC for P3 process rates

crterai:
The subgrid enhancement coefficients are leading to very high conversion efficiency of cloud water to rain water and causing an increase in popcorn cumulus in the model. Because subgrid/subcloud variability of cloud water is difficult to diagnose and we do not expect the subgrid enhancement to be that large in the 3km model, they are set to 1. 
This change is a non-BFB change.


PeterCaldwell:
Thanks for this, Chris! I'd actually like to leave the old implementation commented out rather than deleting all mention of it it. My thinking is that for low-res simulations, we will probably still need this capability. if we truly didn't need this subgrid stuff ever again, we should have also ripped out all mentions of subgrid_variance_scaling. We may eventually want to make the ability to turn this on/off a namelist option, but I think for now commenting it out is the best.

crterai:
@PeterCaldwell I've updated the PR to keep the changes as comments in both F90 and C++

PeterCaldwell (approved):
Perfect :-) Thanks!!!

PeterCaldwell:
I'll merge this once @hassanbeydoun approves. Chris - would you mind kicking off an ne4 run with this branch just to make sure we're not missing something subtle?

hassanbeydoun (approved):

PR #1230, opened 2021-10-01
Some mods to GridsManager classes

bartgol:
Mainly:

- Moved `phisics/share/physics_only_grids_manager.*pp` to `share/grid/mesh_free_grids_manager.*pp`.
- Make MeshFreeGridsManager also handle a simple SEGrid, for ease of testing.
- Simplified `build_grid` method of GridsManager, and added `build_all_grids` (for ease of testing).
- Changed tests inputs to adopt MeshFreeGridsManager's new syntax.


components/scream/src/dynamics/homme/dynamics_driven_grids_manager.cpp
```diff
@@ -50,6 +52,11 @@ DynamicsDrivenGridsManager (const ekat::Comm& comm,
 
   // TODO: add other rebalancing?
 
+  // Get the ref grid name
+  m_ref_grid_name = p.get<std::string>("Reference Grid");
+  EKAT_REQUIRE_MSG (ekat::contains(gn,m_ref_grid_name),
```
jeff-cohere, line 57: What does it mean that a valid reference grid name **contains** `m_ref_grid_name`?

components/scream/src/share/tests/atm_process_tests.yaml
```diff
@@ -4,17 +4,17 @@ Number of Entries: 2
 Schedule Type: Sequential
 
 Process 0:
```
jeff-cohere, line 6: Maybe not appropriate for this PR, but recall the list/sequence YAML construction here. It could be more convenient than a bunch of keys named `Process N`.

components/scream/src/share/tests/grid_tests.cpp
```diff
@@ -44,84 +44,28 @@ TEST_CASE("point_grid", "") {
 }
 
 TEST_CASE("se_grid", "") {
-  // Assume a "strip" of elements:
```
jeff-cohere, line 47: Is this schematic no longer valid?

components/scream/src/share/grid/mesh_free_grids_manager.cpp
```diff
@@ -0,0 +1,178 @@
...
+    // Assume a "strip" of elements:
+    //
+    //  *---*---*     *---*
+    //  |   |   | ... |   *
+    //  *---*---*     *---*
+    //
+    // partitioned evenly across ranks. Use this config to establish
+    // a unique numbering for each element, using the following (np=4 here):
+    //
+    //   1--5--9-13  13-17-21-25
```
jeff-cohere, line 58: Okay, I see where this ended up. Disregard my other comment.

jeff-cohere (commented):
I asked some questions, but nothing of consequence. Okay to merge pending passing tests IMO.

components/scream/src/dynamics/homme/dynamics_driven_grids_manager.cpp
bartgol: Well, I don't want one to request XYZ as a reference grid, when XYZ is not even a valid grid name...

components/scream/src/share/tests/atm_process_tests.yaml
bartgol: I think I have an open issue, where I plan to change the syntax, and turn it into something like
```
  Schedule Type: Sequential:
  Processes: [Homme, SHOC]
  Homme:
     ...
  SHOC:
    ...
```
I also don't really like `Process N`, especially if there are multiple proc groups, and you see `Process 0` in different nested sublists. I'll go check on the issues to make sure this is on the radar for v1.1.

components/scream/src/share/tests/grid_tests.cpp
bartgol: You mean the test name? Or the  deleted comment? If the latter, I moved it to the mesh free grids manager file, to specify how it creates the SE grid.

components/scream/src/share/grid/mesh_free_grids_manager.cpp
bartgol: Ok, I see your comment. Disregard my other comment to your other comment. :)

components/scream/src/dynamics/homme/dynamics_driven_grids_manager.cpp
```diff
@@ -107,10 +112,7 @@ build_grids (const std::set<std::string>& grid_names,
       pg_codes.push_back(code);
     }
   }
-  // Also check the ref grid (and retrieve its code)
-  EKAT_REQUIRE_MSG (supported_grids().count(reference_grid)==1,
-                    "Error! Grid '" +reference_grid + "' is not supported by this grid manager.\n");
-  pg_codes.push_back(m_grid_codes.at(reference_grid));
+  pg_codes.push_back(m_grid_codes.at(m_ref_grid_name));
```
AaronDonahue, line 115: This is really just a general naming convention comment.  I foresee some confusion in the future when we bring in the PhysGrid stuff, since this is typically referred to as "PG".  So we might want to start thinking about how we thing that grid will be names within the context of scream.

components/scream/src/physics/register_physics.hpp
```diff
@@ -13,8 +13,6 @@
 #include "physics/cld_fraction/atmosphere_cld_fraction.hpp"
 #include "physics/rrtmgp/atmosphere_radiation.hpp"
```
AaronDonahue, line 14: This is actually a general comment about the `register_physics.hpp` file.  If one were to use it, does that require that the executable also link to every process registered?  In other words, for a shoc+cld+p3 test we would still need to link to the rrtmgp library in CMakeLists.txt because register_physics would be looking for the RRTMGP information?

If so, how do we plan to handle things when we expand the number of processes to include say more than one microphysics scheme or similar.  Would we still need to include all of the known libraries in each test's CMake file?

AaronDonahue (approved):
Just a few comments that don't relate to holding this PR up.   Otherwise, looks good to me.

AaronDonahue:
I removed "Automerge" to give Jeff a chance to sign off on the PR before it is merged.

components/scream/src/dynamics/homme/dynamics_driven_grids_manager.cpp
bartgol: I think PG should refer to any physics grid, and what is current called "PG" should probably be called FV phys grid (the Finite Volume physics grid), as opposed to the GLL phys grid.

I understand that people are used to call the FV grid just "PG", by now, and "names battles" are a waste of energy. But fwiw, I would argue PG should just mean "physics grid". We can then decide that pgN means FV grid for N>0 and GLL grid for N=0 (like it's done in the code).

AaronDonahue: I'm in agreement on the naming.  I'm not sure we can win the names battle, but we do have some agency at the ground level.  In a few months, when we begin work on implementing the FV grid we can lean into a naming convention there that makes more sense.  I just mention it now so that it is on the radar.

components/scream/src/physics/register_physics.hpp
bartgol: Including the file by itself does not add linking requirements. It is ok to declare functions that are never defined or linked, so long as they are never invoked (though it is bad coding).

However, if one were to actually call `register_physics()`, then all the physics libraries would have to be linked. That means that single-parametrization AD test (or, in general, tests that do not use all parametrizations) should not call `register_physics()`, or else the test would need to link against all physics parametrizations (which causes unneeded build deps).

The comment in the file hints at how one could still use `register_physics()` for single-parametrization tests: guard each include (and `register_product` call) by something like `#ifdef HAS_P3`, and add `HAS_P3` to the compile definitions of each lib. That way, when the lib is "linked" (*) via `target_link_libraries`, is `HAS_BLAH` are propagated to the test, and its include/register sections are compiled.

(*) recall that `target_link_libraries(foo PUBLIC bar)` does more than just linking `bar` to `foo`: it propagates public compiler flags/definitions from `bar` to `foo`.

components/scream/src/dynamics/homme/dynamics_driven_grids_manager.cpp
bartgol: In homme, the fv grid is already called fv. The module is `gllvfremap`, underscoring that the two grids are GLL and FV. Imho, we can simply keep the naming convention used in Homme also in scream.

AaronDonahue (approved):

AaronDonahue (approved):

PR #1231, opened 2021-10-01
Quick fix of typo impacting timer

### LINKED ISSUE #1228: Typo in timer name causes many warnings to e3sm.log
In `components/eam/src/dynamics/se/dyn_comp.F90` it looks like we have a typo in the timer name for the `stop`:

```
           call t_startf('prim_run_subcycle')
           call prim_run_subcycle(dyn_state%elem,hybrid,nets,nete,&
               tstep, single_column_in, TimeLevel, hvcoord, n)
           call t_stopf('prim_run_sybcycle') !<---
```

Which all still works, but will print many (depending on how often this is called, of course) warnings to e3sm.log. Simulation still OK, but might be slowed down and the timing data in that timer will be inaccurate.

```
1: GPTLstopf thread 0: timer for a:prim_run_sybcycle had not been started.
1: GPTLstopf thread 5: timer for a:prim_run_sybcycle had not been started.
1: GPTLstopf thread 1: timer for a:prim_run_sybcycle had not been started.
```

Last night I ran a 48-step test of our next ne1024 case using the new repo to find this. It printed that message 8250179 times, so that's about 170k lines of output that are written to e3sm.log every step -- likely would have caused some issues.

I just verified this error is not in e3sm. It looks like just a typo during the last upstream merge and is only in scream repo.

crterai:
Quick fix of typo in timer to warning messages in e3sm.log. that fills up the log file and slows down the F90 ne1024 simulations. 
Fixes #1228

### DIFF
--- a/components/eam/src/dynamics/se/dyn_comp.F90
+++ b/components/eam/src/dynamics/se/dyn_comp.F90
@@ -394,7 +394,7 @@ CONTAINS
            call t_startf('prim_run_subcycle')
            call prim_run_subcycle(dyn_state%elem,hybrid,nets,nete,&
                tstep, single_column_in, TimeLevel, hvcoord, n)
-           call t_stopf('prim_run_sybcycle')
+           call t_stopf('prim_run_subcycle')
          end do
        endif
 


bartgol (approved):
Given that it's Friday, and other reviewers might already be gone for the weekend, and given the triviality of the changes, I am gonna go ahead and approve, adding automerge.

jgfouca (approved):

ndkeen (approved):

PR #1232, opened 2021-10-04
Disable BFB checks between f90 and cxx when cuda-memcheck is on

jgfouca:
The BFB math layer is extremely expensive when cuda-memcheck is on, so we can't use it and therefore we can't expect BFB results. BFB testing is already done in the dbg test, so it adds no value to repeat it in the cuda-memcheck test.

Also, this PR improves how scripts-tests verifies the behavior of bfb unit tests and baseline diffs.


jgfouca:
Most of the changes in this PR are changing BFB tests from

```
#ifndef NDEBUG
   * do bfb checks
```

to

```
    if (SCREAM_BFB_TESTING) {
```

SCREAM_BFB_TESTING will be false if NDEBUG or EKAT_ENABLE_CUDA_MEMCHECK. This change impacted a lot of lines and also impacted indenting of many code regions, but is not interesting, so reviewing that stuff can be skipped if you approve of the idea.

components/scream/cmake/ctest_script.cmake
```diff
@@ -45,10 +45,19 @@ else ()
       endif()
   else()
     if (NOT BUILD_ONLY)
+
+      if (DEFINED ENV{SCREAM_FORCE_RUN_DIFF})
+        set(INCLUDE_REGEX "p3_run_and_cmp")
+      elseif (DEFINED ENV{SCREAM_FORCE_RUN_DIFF_BFB_UNIT})
+        set(INCLUDE_REGEX "p3_tests_ut")
```
bartgol, line 52: I'm not following here. Is the SCREAM_FORCE_RUN_DIFF option only forcing a diff in baselines tests, while SCREAM_FORCE_RUN_DIFF_BFB_UNIT forces a diffs in the unit tests only? Why isn't `RUN_DIFF` including also unit tests? Maybe I'm confused by the var names... And why is this logic applied only to P3?

bartgol: On a side note, perhaps a better way to handle this would be through labels? A regex would have to be modified every time we have a new test that needs to be added, while using specific labels would not require the ctest script to be updated every time.

components/scream/scripts/scripts-tests
```diff
@@ -360,6 +360,27 @@ class TestTestAllScream(TestBaseOuter.TestBase):
         else:
             self.skipTest("Skipping full run")
 
+    def test_run_diff_opt_not_captured(self):
```
bartgol, line 363: Coluld you add a 1 line comment for each of these fcns, explaining what they are testing?

components/scream/src/physics/p3/p3_rain_sed_impl.hpp
```diff
@@ -152,9 +152,12 @@ ::rain_sedimentation(
       });
     }
     Kokkos::single(
-     Kokkos::PerTeam(team), [&] () {
-      precip_liq_surf += prt_accum * C::INV_RHO_H2O * inv_dt;
-    });
+      Kokkos::PerTeam(team), [&] () {
+        precip_liq_surf += prt_accum * C::INV_RHO_H2O * inv_dt;
+#if defined(SCREAM_FORCE_RUN_DIFF) || defined(SCREAM_FORCE_RUN_DIFF_BFB_UNIT)
```
bartgol, line 157: Can you add a comment line here, explaining to puzzled devs that we are just trying to trigger an error, to see if the testing system can catch it?

components/scream/src/physics/p3/tests/p3_run_and_cmp.cpp
```diff
@@ -219,9 +219,6 @@ struct Baseline {
       const auto& f = fdi.getfield(i);
       ekat::write(&f.dim, 1, fid);
       ekat::write(f.extent, f.dim, fid);
-#ifdef SCREAM_FORCE_RUN_DIFF
```
bartgol, line 222: Why is this no longer needed? What's forcing the run diff now?

components/scream/src/physics/shoc/tests/shoc_diag_second_mom_srf_test.cpp
```diff
@@ -115,16 +115,16 @@ struct UnitWrap::UnitTest<D>::TestSecondMomSrf {
...
     }
 
-#ifndef NDEBUG
-    static constexpr Int num_runs = sizeof(mom_srf_data_f90) / sizeof(SHOCSecondMomentSrfData);
-    for (Int i = 0; i < num_runs; ++i) {
-      Int shcol = mom_srf_data_cxx[i].shcol;
-      for (Int k = 0; k < shcol; ++k) {
-        REQUIRE(mom_srf_data_f90[i].ustar2[k] == mom_srf_data_cxx[i].ustar2[k]);
-        REQUIRE(mom_srf_data_f90[i].wstar[k]  == mom_srf_data_cxx[i].wstar[k]);
+    if (SCREAM_BFB_TESTING) {
```
bartgol, line 118: This code is ifdeff-ed out. Is there a reason to keep it? Should we take this chance to purge it? Tagging also @bogensch , which git blame says wrote these lines...

components/scream/cmake/ctest_script.cmake
jgfouca: @bartgol , yes, we want SCREAM_FORCE_RUN_DIFF to only run the p3_run_and_cmp test. If we ran all the tests, the BFB ones would fail too since we "sabotage" rain_sed in order to get a diff. This logic applies to p3 because we leverage p3 to test diffs. In the main CMakeLists.txt file:

```
if ($ENV{SCREAM_FORCE_RUN_DIFF})
  add_definitions(-DSCREAM_FORCE_RUN_DIFF)
  set(SCREAM_FAKE_NEED_P3 ON)
endif()
if ($ENV{SCREAM_FORCE_RUN_DIFF_BFB_UNIT})
  add_definitions(-DSCREAM_FORCE_RUN_DIFF_BFB_UNIT)
  set(SCREAM_FAKE_NEED_P3 ON)
endif()
```

So, if SCREAM_FAKE_ONLY is on, we build almost none of SCREAM, but then we don't have enough to test that diffs are detected, so we add p3 if it looks like that's what we are trying to test.

Similar logic would be needed with labels. `INCLUDE_REGEX` is used to limit what tests are run by ctest so, if we are trying to detect a baseline diff, we only run the baseline test (p3_run_and_cmp). I hope that makes sense.

components/scream/scripts/scripts-tests
jgfouca: Yes

components/scream/src/physics/p3/p3_rain_sed_impl.hpp
jgfouca: yes

components/scream/src/physics/p3/tests/p3_run_and_cmp.cpp
jgfouca: Because we want the difference to be "authentic", IE coming from the real p3 code. This is why the code was introduced to p3_rain_sed_impl above.

components/scream/src/physics/shoc/tests/shoc_diag_second_mom_srf_test.cpp
jgfouca: Don't care either way.

jgfouca:
@bartgol , added better comments as you requested.

components/scream/cmake/ctest_script.cmake
bartgol: Ok, regex or label would not change much, so that's fine.

bartgol (approved):
I removed the automerge since I am approving, but want to give @bogensch a chance to respond to the comment about purging the ifdeffed SHOC code. If he's fine purging, I would like to purge it.

bartgol (approved):

components/scream/src/physics/shoc/tests/shoc_diag_second_mom_srf_test.cpp
bogensch: @bargol I'd prefer to keep these lines of code.  Not sure why they were iffdeff-ed in the first place.

bartgol: Should they be compiled/run then?

bogensch: I think they should be?  I guess after looking at it again I'm a little less confident at giving an answer.  While I wrote the property test section of the code for this file, I didn't write the b4b testing section (for any functions) so I'm less familiar.  I'm surprised it said I wrote these lines.  Thus I'm not clear what the issue was with this block of code in the first place to iffdeff them out...  but ustar2 and wstar are the main outputs of this routine, so I would think they should be checked for bit-for-bit ness.

bartgol: Ok. If you feel like this should be kept (and possibly cleaned up/enhanced), can you please go ahead and open an issue, explaining what these lines should do/test. So hopefully we won't forget.

bartgol (approved):

PR #1233, opened 2021-10-04
Scorpio read input interface allow for specific time level to be read

AaronDonahue:
This PR introduces the ability to read input at a specific time level in files that may contain more than one time level of data.

A new test is introduced that first writes output for every time step in a 10 time step run and then reads each time snap individually and checks that it matches.

This is a necessary precursor to the SPA phase 2 work.

This PR also fixes a bug where it wasn't possible to write multiple time steps to output.  Per the above test, this is now tested as well.

### DIFF
--- a/components/scream/src/share/io/scorpio_input.cpp
+++ b/components/scream/src/share/io/scorpio_input.cpp
@@ -136,13 +136,31 @@ set_grid (const std::shared_ptr<const AbstractGrid>& grid)
 /* ---------------------------------------------------------- */
 void AtmosphereInput::read_variables ()
 {
+  // The default read input is to use the last step set for this
+  // file by eam_update_timesnap.
+  // To trigger the default in the read_variables routine we pass
+  // a negative value for the time level.
+  read_variables(-999);
+}
+/* ---------------------------------------------------------- */
+// Note: The timelevel argument provides a way to control which
+//       time snap to read input from in the file.  If a negative
+//       number is provided the routine will read input at the
+//       last time level set by running eam_update_timesnap.
+void AtmosphereInput::read_variables (const int timelevel)
+{
   EKAT_REQUIRE_MSG (m_is_inited,
       "Error! The init method has not been called yet.\n");
 
   for (auto const& name : m_fields_names) {
 
     // Read the data
-    scorpio::grid_read_data_array(m_filename,name,m_host_views_1d.at(name).data());
+    if (timelevel < 0)
+    {
+      scorpio::grid_read_data_array(m_filename,name,m_host_views_1d.at(name).data());
+    } else {
+      scorpio::grid_read_data_array(m_filename,name,timelevel,m_host_views_1d.at(name).data());
+    }
 
     // If we have a field manager, make sure the data is correctly
     // synced to both host and device views of the field.
--- a/components/scream/src/share/io/scorpio_input.cpp
+++ b/components/scream/src/share/io/scorpio_input.cpp
@@ -134,20 +134,11 @@ set_grid (const std::shared_ptr<const AbstractGrid>& grid)
 }
 
 /* ---------------------------------------------------------- */
-void AtmosphereInput::read_variables ()
-{
-  // The default read input is to use the last step set for this
-  // file by eam_update_timesnap.
-  // To trigger the default in the read_variables routine we pass
-  // a negative value for the time level.
-  read_variables(-999);
-}
-/* ---------------------------------------------------------- */
-// Note: The timelevel argument provides a way to control which
+// Note: The time_index argument provides a way to control which
 //       time snap to read input from in the file.  If a negative
 //       number is provided the routine will read input at the
 //       last time level set by running eam_update_timesnap.
-void AtmosphereInput::read_variables (const int timelevel)
+void AtmosphereInput::read_variables (const int time_index)
 {
   EKAT_REQUIRE_MSG (m_is_inited,
       "Error! The init method has not been called yet.\n");
@@ -155,12 +146,7 @@ void AtmosphereInput::read_variables (const int timelevel)
   for (auto const& name : m_fields_names) {
 
     // Read the data
-    if (timelevel < 0)
-    {
-      scorpio::grid_read_data_array(m_filename,name,m_host_views_1d.at(name).data());
-    } else {
-      scorpio::grid_read_data_array(m_filename,name,timelevel,m_host_views_1d.at(name).data());
-    }
+    scorpio::grid_read_data_array(m_filename,name,time_index,m_host_views_1d.at(name).data());
 
     // If we have a field manager, make sure the data is correctly
     // synced to both host and device views of the field.
--- a/components/scream/src/share/io/scorpio_input.hpp
+++ b/components/scream/src/share/io/scorpio_input.hpp
@@ -136,6 +136,7 @@ public:
 
   // Read fields that were required via parameter list.
   void read_variables ();
+  void read_variables (const int timelevel);
   int read_int_scalar (const std::string& name);
   void finalize();
 
--- a/components/scream/src/share/io/scorpio_input.hpp
+++ b/components/scream/src/share/io/scorpio_input.hpp
@@ -135,8 +135,7 @@ public:
             const std::map<std::string,FieldLayout>&  layouts);
 
   // Read fields that were required via parameter list.
-  void read_variables ();
-  void read_variables (const int timelevel);
+  void read_variables (const int time_index = -1);
   int read_int_scalar (const std::string& name);
   void finalize();
 
--- a/components/scream/src/share/io/scream_scorpio_interface.F90
+++ b/components/scream/src/share/io/scream_scorpio_interface.F90
@@ -216,7 +216,7 @@ module scream_scorpio_interface
   end interface grid_read_data_array
 !----------------------------------------------------------------------
   interface grid_write_data_array
-    module procedure grid_write_darray_1d_real
+    module procedure grid_write_darray_1d
   end interface
 !----------------------------------------------------------------------
 contains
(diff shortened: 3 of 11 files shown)


components/scream/src/share/io/scorpio_input.hpp
```diff
@@ -132,10 +133,12 @@ class AtmosphereInput
   //  - layouts: the layout of the vars (used to reshape the views).
   void init(const std::shared_ptr<const grid_type>& grid,
             const std::map<std::string,view_1d_host>& host_views_1d,
-            const std::map<std::string,FieldLayout>&  layouts);
+            const std::map<std::string,FieldLayout>&  layouts,
+            const std::map<std::string, std::string>& dtypes);
 
   // Read fields that were required via parameter list.
   void read_variables ();
+  void read_variables (const int timelevel);
```
bartgol, line 141: We can simply have a single method, with a default for timelevel: `void read_variables (const int timelevel = -1);`

Also, can we use `time_index`, rather than `time_level`? The word level makes me think of something with a fixed upper bound (like the vert levels, or the time levels in homme used for RK stages storage).

components/scream/src/share/io/scorpio_input.cpp
```diff
@@ -33,10 +33,11 @@ AtmosphereInput (const ekat::Comm& comm,
                  const ekat::ParameterList& params,
                  const std::shared_ptr<const grid_type>& grid,
                  const std::map<std::string,view_1d_host>& host_views_1d,
-                 const std::map<std::string,FieldLayout>&  layouts)
+                 const std::map<std::string,FieldLayout>&  layouts,
+                 const std::map<std::string, std::string>& dtypes)
```
bartgol, line 37: This first draft of dtype feature is probably not going to work, and will need some revision: you are passing views with a Real data type, so anyt value of dtype different from "Real" would be erroneous. Given that, this is not really an "added feature".

I would hold off any mod that has to do with dtype, until we have a better design in mind. This PR should focus only on the time index feature.

components/scream/src/share/io/tests/io.cpp
```diff
@@ -122,6 +122,7 @@ TEST_CASE("input_output_basic","io")
   auto avg_params = get_in_params("Average",io_comm);
   auto min_params = get_in_params("Min",io_comm);
   auto max_params = get_in_params("Max",io_comm);
+  auto multi_params = get_in_params("Multisnap",io_comm);
```
bartgol, line 125: You should update the cocmment above. We no longer expect 4 files, but 5. 1 file each for average, instantaneous, min and max data, and multiple time snapshots.

components/scream/src/share/io/tests/io.cpp
```diff
@@ -122,6 +122,7 @@ TEST_CASE("input_output_basic","io")
...
   Real tol = pow(10,-6);
```
bartgol, line 126: This seems like a very loose tolerance for double precision. Can we try to set `tol=100*std::numeric_limits<Real>::epsilon()`? It is already quite loose, given that no catastrophic cancellation should happen.

components/scream/src/share/io/scorpio_input.cpp
```diff
@@ -135,14 +136,32 @@ set_grid (const std::shared_ptr<const AbstractGrid>& grid)
...
 
   for (auto const& name : m_fields_names) {
 
     // Read the data
-    scorpio::grid_read_data_array(m_filename,name,m_host_views_1d.at(name).data());
+    if (timelevel < 0)
+    {
+      scorpio::grid_read_data_array(m_filename,name,m_host_views_1d.at(name).data());
+    } else {
+      scorpio::grid_read_data_array(m_filename,name,timelevel,m_host_views_1d.at(name).data());
```
bartgol, line 163: It seems to me that you only need one version of this subroutine (the one with timelevel). (see comments to scorpio interfaces)

components/scream/src/share/io/scream_scorpio_interface.cpp
```diff
@@ -17,7 +17,8 @@ extern "C" {
   void register_file_c2f(const char*&& filename, const int& mode);
   void set_decomp_c2f(const char*&& filename);
   void set_dof_c2f(const char*&& filename,const char*&& varname,const Int dof_len,const Int *x_dof);
-  void grid_read_data_array_c2f_real(const char*&& filename, const char*&& varname, Real *&hbuf);
+  void grid_read_data_array_c2f(const char*&& filename, const char*&& varname, Real *&hbuf);
+  void grid_read_data_array_wtime_c2f(const char*&& filename, const char*&& varname, Real *&hbuf, const Int timelevel);
```
bartgol, line 21: Keep just this one (if you really want, default time index to -1), and get rid of the other one.

components/scream/src/share/io/scream_scorpio_interface.cpp
```diff
@@ -131,7 +132,11 @@ void count_pio_atm_file() {
 }
 /* ----------------------------------------------------------------- */
 void grid_read_data_array(const std::string &filename, const std::string &varname, Real *hbuf) {
-  grid_read_data_array_c2f_real(filename.c_str(),varname.c_str(),hbuf);
+  grid_read_data_array_c2f(filename.c_str(),varname.c_str(),hbuf);
+}
+/* ----------------------------------------------------------------- */
+void grid_read_data_array(const std::string &filename, const std::string &varname, const int timelevel, Real *hbuf) {
```
bartgol, line 138: Same here: keep one fcn only

components/scream/src/share/io/scream_scorpio_interface.hpp
```diff
@@ -59,6 +59,7 @@ namespace scorpio {
 
   /* Read data for a specific variable from a specific file. */
   void grid_read_data_array (const std::string &filename, const std::string &varname, Real* hbuf);
+  void grid_read_data_array (const std::string &filename, const std::string &varname, const int timelevel, Real* hbuf);
```
bartgol, line 62: And here.

components/scream/src/share/io/scream_scorpio_interface_iso_c2f.F90
```diff
@@ -270,6 +270,23 @@ subroutine grid_read_data_array_c2f_real(filename_in,varname_in,var_data_ptr) bi
     call convert_c_string(varname_in,varname)
     call grid_read_data_array(filename,varname,var_data_ptr)
 
-  end subroutine grid_read_data_array_c2f_real
+  end subroutine grid_read_data_array_c2f
+!=====================================================================!
+  subroutine grid_read_data_array_wtime_c2f(filename_in,varname_in,var_data_ptr,timelevel) bind(c)
```
bartgol, line 275: Keep only this (remove `_wtime` from fcn name), and remove the previous impl.

components/scream/src/share/io/scream_scorpio_interface.F90
```diff
@@ -1216,61 +1218,84 @@ subroutine grid_write_darray_1d_real(filename, varname, var_data_ptr)
...
+  subroutine grid_read_darray_1d(filename, varname, var_data_ptr, timelevel)
 
     ! Dummy arguments
-    character(len=*), intent(in) :: filename       ! PIO filename
-    character(len=*), intent(in) :: varname
-    type (c_ptr),     intent(in) :: var_data_ptr
+    character(len=*), intent(in)  :: filename       ! PIO filename
+    character(len=*), intent(in)  :: varname
+    type (c_ptr),     intent(in)  :: var_data_ptr
+    integer, optional, intent(in) :: timelevel
```
bartgol, line 1262: Don't make it optional. Force to pass it.

components/scream/src/share/io/scream_scorpio_interface.F90
```diff
@@ -1216,61 +1218,84 @@ subroutine grid_write_darray_1d_real(filename, varname, var_data_ptr)
...
     logical                            :: found
-    real(rtype), dimension(:), pointer :: var_data
+    real(rtype), dimension(:), pointer         :: var_data_real
+    integer(kind=c_int), dimension(:), pointer :: var_data_int
 
     call lookup_pio_atm_file(trim(filename),pio_atm_file,found)
     call get_var(pio_atm_file,varname,var)
+
+    ! Set the timesnap we are reading
+    if (present(timelevel)) then
```
bartgol, line 1276: Instead of `if(present(time_index))`, you can do `if (time_index .gt. 0)`.

bartgol (changes requested):
I think this PR is mixing two purposes, one of which is only a stump. In particular, the time index feature is well defined, tested, and probably fully functional (at least, to the extents that we foresee using it for now). The other is the dtype feature, which, however, is not tested, and might require some serious thinking.

IMHO, the dtype proto-feature should be left out from this PR, and all its use cases should be put in a separate branch. Do you have any strong reason to leave the dtype proto-feature mixed up with the time index PR?

I also have a suggestion about the time index functions duplications. Other than that, it looks good.

AaronDonahue:
@bartgol , thank you for the review.  I struggled with whether or not to make the time level (time index) feature optional or not.  Based on your comments I'll switch it over to being always an argument.

As for the dtype stuff.  I added that after the fact, and would be happy to remove it, or make it a separate PR.  I will need something like this for SPA, because I need to have a way to designate that I am reading Integer input rather than Real input.  Perhaps there is a better way to implement this which we can discuss offline and as a part of a different PR.

bartgol:
> @bartgol , thank you for the review. I struggled with whether or not to make the time level (time index) feature optional or not. Based on your comments I'll switch it over to being always an argument.

Yeah, if you make it a defaulted arg in the IO classes, then you only need one impl everywhere, which is prob easier to maintain.
 
> As for the dtype stuff. I added that after the fact, and would be happy to remove it, or make it a separate PR. I will need something like this for SPA, because I need to have a way to designate that I am reading Integer input rather than Real input. Perhaps there is a better way to implement this which we can discuss offline and as a part of a different PR.

I would keep this PR focused on the new feature. My issue with the dtype stuff is that it might take some work to figure out how to handle int (or whatever) as input views. So given that the feature is not really working, I am inclined in removing it. At the very least, I would remove it from the IO classes interfaces. But at that point, you can simply remove it everywhere, and just do it in a second PR altogether.

AaronDonahue:
@bartgol , alright.  I think I've made all the suggested changes.  Would you mind rereviewing after the PR goes through another round of AT?

components/scream/src/share/io/scream_scorpio_interface.F90
```diff
@@ -1216,61 +1218,84 @@ subroutine grid_write_darray_1d_real(filename, varname, var_data_ptr)
...
+    integer(kind=c_int), dimension(:), pointer :: var_data_int
 
     call lookup_pio_atm_file(trim(filename),pio_atm_file,found)
     call get_var(pio_atm_file,varname,var)
+
+    ! Set the timesnap we are reading
+    if (time_index .gt. 0) then
+      call PIO_setframe(pio_atm_file%pioFileDesc,var%piovar,int(time_index,kind=pio_offset_kind))
+    else
+      call PIO_setframe(pio_atm_file%pioFileDesc,var%piovar,int(max(1,pio_atm_file%numRecs),kind=pio_offset_kind))
```
bartgol, line 1279: It seems that `numRecs` is the number of records on file, while `time_index` is the time index since the beginning of the simulation. I am wondering what happens when we limit the number of snapshots on file. Say you put 10 snapshots per file, at the 11th time step, `time_index=11`, but `numRecs=0` on the file. Is it ok to write to file a time index 11, withuot any previous time slice? I guess so, but I want to make sure we are not overlooking some index arithmetics that scorpio might assume under the hood.

bartgol (commented):
Looks good. I have one comment/question about time index when we write to file.

components/scream/src/share/io/scream_scorpio_interface.F90
AaronDonahue: My guess is that if you ask for a time_index that is outside of the bounds then PIO would return an error.  We can check for that.

I'll note, this is currently only being used for reading, so no writing.  I'm not sure what the policy is for writing when you specify a time index.  If PIO will fill in all missing time_indices with something, or if it will throw an error. 

I'm also not sure if we will ever use a similar syntax for writing, since we can currently already write multiple time snaps per file and we follow the convention of just bumping the numRecs up by one every time we write.

bartgol: You're right this is a read routine. I would then like to add a check before the call to `PIO_setframe`. Even if PIO would error out, I think we can do a better job with the error message, which would make life easier for debugging.

We should check that `0<=time_index<num_time_slices_on_file` (or add 1 on both sides if using netcdf numbering).  Then, if the check fails, we should print a nice msg printing the requested slice, and the valid interval.

AaronDonahue: Okay, this will actually require a bit more work I think.  numRecs is set manually by the interface and only changes when `eam_update_timestep` is called.  Perhaps we should remove it entirely from the read_input call.  And instead query the size of the time dimension directly from PIO and then put in your check.

bartgol: Taking this offline.

bartgol (approved):

PR #1234, opened 2021-10-04
[WIP] Increasing the maximal rain size in P3 from 0.8-mm to 2.0-mm

JS-WRF-SBM:
The original number-weighted maximal rain size was limited in old versions of the P3 scheme to 0.8-mm.
This diameter size is smaller then typical stratiform precipitation, let alone convective precip. 
For a 3-km resolution cloud microphysics this maximal size is likely to be reached many times along a simulation.

The rain maximal size is affected by the ice category melting, rain-collection + breakup, as well as size sorting (e.g., sedimentation). Although, there weren't major updates in the corresponding processes in P3, the current maximal size of P3 is 2-mm.

### DIFF
--- a/components/eam/src/physics/cam/micro_p3.F90
+++ b/components/eam/src/physics/cam/micro_p3.F90
@@ -52,7 +52,7 @@ module micro_p3
        clbfact_dep,iparam, isize, densize, rimsize, rcollsize, ice_table_size, collect_table_size, &
        get_latent_heat, T_zerodegc, pi=>pi_e3sm, dnu, &
        T_rainfrz, T_icenuc, T_homogfrz, iulog=>iulog_e3sm, &
-       masterproc=>masterproc_e3sm, calculate_incloud_mixingratios, mu_r_constant, &
+       masterproc=>masterproc_e3sm, calculate_incloud_mixingratios, mu_r_constant, inv_Drmax, &
        lookup_table_1a_dum1_c, &
        p3_qc_autocon_expon, p3_qc_accret_expon
 
@@ -1857,7 +1857,7 @@ contains
        mu_r = mu_r_constant
        lamr   = bfb_cbrt(cons1*nr*(mu_r+3._rtype)*(mu_r+2._rtype)*(mu_r+1._rtype)/(qr))  ! recalculate slope based on mu_r
        lammax = (mu_r+1._rtype)*1.e+5_rtype   ! check for slope
-       lammin = (mu_r+1._rtype)*1250._rtype   ! set to small value since breakup is explicitly included (mean size 0.8 mm)
+       lammin = (mu_r+1._rtype)*inv_Drmax     ! set to small value since breakup is explicitly included (mean size 2.0 mm)
 
        ! apply lambda limiters for rain
        if (lamr.lt.lammin) then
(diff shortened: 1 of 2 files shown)


JS-WRF-SBM:
> Thanks @JS-WRF-SBM . A couple of questions:
> 
> 1. How did you choose 2 mm as the max drop size? It seems like 0.8 was probably chosen as a tuning parameter. I agree 0.8 is unrealistically small, but 2mm seems kind of arbitrary. I know much bigger drops have been observed (e.g. https://www.washington.edu/news/2004/07/22/record-rain-uw-scientists-find-some-of-the-biggest-raindrops-ever/).
> 2. What is the impact on climate at ne30 of making this change?
> 3. I'm going to change the PR to just hardcode the max drop size rather than making it a namelist parameter just because I don't want to spend a lot of time on implementing this change right now. We will switch it back to a namelist parameter after our v1 release.

Hi @PeterCaldwell 

1. The maximal rain diameter size allowed through the v4.1/v5.1.6/v5.3 lookup table is 5-mm. Theoretically, rain drops could be up to 5/6 -mm, but these are not stable drops, for which we introduce raindrop breakup which eliminated these large drop and set a smaller, more stable, modal size. 
In the last 2 versions of P3, the authors chose a larger value (0.8mm >> 2mm). Since the relevant physical processes which determines the rain size are rather identical between these versions, there is no reason to limit the rain size:    
https://github.com/wrf-model/WRF/blob/eed56d74b865af4ce9f98ea029226acfe52b0569/phys/module_mp_p3.F#L207
2. For the `ne30` run: the LWCF bias is reduced overall, and the bias in sfc precip associated with the Indian / Indian ocean / African continent is mostly eliminated. Over these areas ZM is rather strongly coupled with P3.
3. Yeah, I was just thinking you must see some more meaningful sfc precip / radiative forcing sensitivity in your 3-km run.  
Good luck!

hassanbeydoun:
Hi @JS-WRF-SBM. Just a quick clarification. The 0.8 mm max value corresponds to the maximum mean drop diameter not the max drop diameter possible. At a threshold rain mixing ratio of 1e-2 kg/kg air, a lambda of 1250 results in about 3000 drops per kg air. If I plot n(D)*dD against diameter for 100 um size bins, I find that my max diameter (diameter at which n(D)*dD = 1) is just under 5 mm. if I use a lambda of 500, I get a max diameter of 7 mm. 
rain_dsds
This is not to say that adjusting lam_min is not a good idea. I've looked at that MWR paper you referred to and if later versions of WRF adjusted lam_min, we should do so as well. Thanks!

hassanbeydoun (approved):
Thanks for the change. Looks good!

JS-WRF-SBM:
> Hi @JS-WRF-SBM. Just a quick clarification. The 0.8 mm max value corresponds to the maximum mean drop diameter not the max drop diameter possible. At a threshold rain mixing ratio of 1e-2 kg/kg air, a lambda of 1250 results in about 3000 drops per kg air. If I plot n(D)*dD against diameter for 100 um size bins, I find that my max diameter (diameter at which n(D)*dD = 1) is just under 5 mm. if I use a lambda of 500, I get a max diameter of 7 mm. rain_dsds This is not to say that adjusting lam_min is not a good idea. I've looked at that MWR paper you referred to and if later versions of WRF adjusted lam_min, we should do so as well. Thanks!

Hi @hassanbeydoun,
Yeah, in bulk schemes we cannot literally speak about maximal rain size b/c we integrate the rain particle size distribution, and thus only a certain modal size of interest can be used as a threshold per time step (e.g., mean-mass, mean-volume, etc ..). 
In addition, we apply rain breakup so we cannot literally reach max drop diam possible (which is likely not a stable drop). 
The 'maximal rain size' here was attributed to the minimal slope (lambda) and to the name `inv_Drmax` which the authors chose.
Thanks for the comments!

PR #1239, opened 2021-10-05
Clean up context mask in p3

PeterCaldwell:
This PR:

1. Fixes a bug in run_bfb_p3_main_part2 where nk was getting set to a default value of -1 in p3_main_part2_f which caused range_mask to always be false. Range_mask was computed appropriately in full model runs, but not when part2 was called directly for a unit test. Fix was just adding nk as an input variable in the part2 call in p3_functions_f90.cpp.
2. Includes `|| !range_mask` as a condition in the `skip_all` logical calculation: range_mask is false wherever a cell is included solely to pad out our pack arrays, so you certainly shouldn't be doing calculations with it. 
3. For the 4 functions which took both range_mask and not_skip_all as inputs, only pass not_skip_all.


PeterCaldwell:
I've now confirmed that the bug in range_mask in p3_main_part2 testing previously had no impact because range_mask is only used for qv_sat, which ignored range_mask except for in checks that T wasn't nan.

jgfouca (approved):

PR #1240, opened 2021-10-06
Clean up saturation conservation checks: "Koby fix #2"

PeterCaldwell:
This PR replaces #1060. It modifies the checks that make sure ice processes don't push qv above liquid saturation or below ice saturation. In particular it: 

1. *moves* these checks from being the first conservation checks to being the last. This probably isn't necessary, but since these checks just reduce conversion from qi and qr to qv, and since qv is typically much greater than 0, they shouldn't break any earlier mass>=0 fixes.

2. *modifies* ice_deposition_sublimation to prevent deposition and sublimation from pushing qv above or below ice saturation (respectively). Deposition/sublimation rates are computed as (qv - q_sat_i)/(abi*tau_i) where abi is the psychrometric condensational correction and tau_i is the ice supersaturation removal timescale. So ensuring deposition/sublimation don't cause qv to become super- or sub-saturated is equivalent to replacing abi*tau_i with max(abi*tau_i,dtime). Modified ice_deposition_sublimation unit test to also confirm no super and subsaturation and do other property tests. 

3. Now that ice_deposition_sublimation avoids oversaturation, we also *delete* the prevent_ice_overdepletion function, whose purpose was to perform exactly this calculation. Because the original version of P3 supported multiple ice categories which collectively could super-or sub-saturate even if an individual pass through ice_deposition_sublim didn't, this correction needed to wait for the sum over all categories to be calculated. Since SCREAM only supports one ice category, it now makes more sense to perform that check where deposition and sublimation are first calculated. 

4. *Add* "prevent_liq_supersaturation" and rewrite it to ensure that the sum of rain evap and sublimation don't leave end-of-step qv supersaturated with respect to liquid. Even though both rain evap and sublimation throttled to ensure they don't by themselves do this, their net effect could. This is a check which should have been in the original P3 but wasn't. Added bfb and property tests ensuring this check is working correctly.


components/scream/src/physics/p3/p3_functions.hpp
```diff
@@ -689,11 +680,9 @@ struct Functions
 
   // TODO (comments)
   KOKKOS_FUNCTION
-  static void ice_deposition_sublimation(const Spack& qi_incld,
-    const Spack& ni_incld, const Spack& T_atm, const Spack& qv_sat_l, const Spack& qv_sat_i,
-    const Spack& epsi, const Spack& abi, const Spack& qv, Spack& qv2qi_vapdep_tend,
-    Spack& qi2qv_sublim_tend, Spack& ni_sublim_tend, Spack& qc2qi_berg_tend,
-    const Smask& context = Smask(true));
+  KOKKOS_FUNCTION
```
jgfouca, line 683: Looks like `KOKKOS_FUNCTION` was duplicated here.

components/scream/src/physics/p3/p3_functions_f90.cpp
```diff
@@ -1314,40 +1363,34 @@ void update_prognostic_liquid_f(Real qc2qr_accret_tend_, Real nc_accret_tend_, R
   *nr_    = t_h(5);
 }
 
-void ice_deposition_sublimation_f(Real qi_incld_, Real ni_incld_, Real T_atm_, Real qv_sat_l_,
-				  Real qv_sat_i_, Real epsi_, Real abi_, Real qv_,
-				  Real* qv2qi_vapdep_tend_, Real* qi2qv_sublim_tend_, Real* ni_sublim_tend_, Real* qc2qi_berg_tend_)
+void ice_deposition_sublimation_f(Real qi_incld, Real ni_incld, Real t_atm, Real qv_sat_l, Real qv_sat_i, Real epsi, Real abi, Real qv, Real inv_dt, Real* qidep, Real* qi2qv_sublim_tend, Real* ni_sublim_tend, Real* qiberg)
 {
-  using P3F = Functions<Real, DefaultDevice>;
+#if 0
```
jgfouca, line 1368: This entire region is disabled by `#if 0`

components/scream/src/physics/p3/p3_functions_f90.cpp
```diff
@@ -3385,5 +3402,29 @@ void ni_conservation_f(Real ni, Real ni_nucleat_tend, Real nr2ni_immers_freeze_t
   *ni_sublim_tend = t_h(2);
 }
 
+void prevent_liq_supersaturation_f(Real pres, Real t_atm, Real qv, Real latent_heat_vapor, Real latent_heat_sublim, Real dt, Real qidep, Real qinuc, Real* qi2qv_sublim_tend, Real* qr2qv_evap_tend)
+{
+#if 0
```
jgfouca, line 3407: This entire region is disabled by `#if 0`

jgfouca (changes requested):

JS-WRF-SBM (changes requested):
Other than applying the time-scale limit operator, the Fortran side looks good to me.

components/scream/src/physics/p3/p3_functions_f90.cpp
PeterCaldwell: @jgfouca - 
1. it's odd these `#if 0` comments are here - I suspect they must have been added by gen-boiler because I certainly didn't add these lines...
2. I don't think the newest versions of our tests use <func>_f any more. I meant to ask you whether I can actually just delete <func>_f references...

components/scream/src/physics/p3/p3_functions.hpp
PeterCaldwell: Ok, fixed.

components/scream/src/physics/p3/p3_functions_f90.cpp
jgfouca: @PeterCaldwell , you're right. Since we now have a full cxx mode for P3, there's no need for f90 p3 routines to call their cxx counterparts. I will issue a PR to clean this up

components/eam/src/physics/cam/micro_p3.F90
```diff
@@ -2878,64 +2878,70 @@ end subroutine ice_supersat_conservation
...
+  real(rtype), intent(in) :: qv
+  real(rtype), intent(in) :: latent_heat_sublim,latent_heat_vapor
+  real(rtype), intent(in) :: dt
+  real(rtype), intent(in) :: qidep,qinuc
+  real(rtype), intent(inout) :: qi2qv_sublim_tend,qr2qv_evap_tend
+
+  real(rtype) :: qv_sinks, qv_sources, qv_endstep, T_endstep, qsl, A, frac
+
+  qv_sources = qi2qv_sublim_tend + qr2qv_evap_tend
+  if (qv_sources == 0._rtype) then
```
JS-WRF-SBM, line 2898: @PeterCaldwell 

It's a bit risky to equate to a floating point (12 digit) number.
A condition is a better practice (or equate to an integer conditional switch).

components/scream/src/physics/p3/p3_functions_f90.cpp
PeterCaldwell: Ok, I'll delete the 2 _f functions you noticed are `#if 0` in this PR, but leave the rest of them for you to deal with later.

components/eam/src/physics/cam/micro_p3.F90
PeterCaldwell: Yeah, testing caught this issue too. Fixed.

components/scream/src/physics/p3/tests/p3_prevent_liq_supersaturation_tests.cpp
```diff
@@ -0,0 +1,196 @@
...
+
+    constexpr Scalar inv_cp       = C::INV_CP;
+    constexpr Scalar rv           = C::RV;
+
+    //Start with reasonable values
+    //============================
+    Spack pres(100000);
+    Spack t_atm(270);
+    Spack qv(1.7e-3);
+    Spack latent_heat_vapor(2.5e6);
```
AaronDonahue, line 35: I'm not sure it's important for the test,  but I believe that the latent_heat values are also in the physics constants.

PeterCaldwell: I think my attempts to access physics constants in this context failed and I figured a unit test doesn't need to be exact... So I think I will punt on fixing this.

PR #1244, opened 2021-10-07
SCREAM: fix compiler error related to v0 access to bfb math functions

bartgol:
Should fix failures of CIME cases tests in nighlies.


jgfouca:
Force merging because I need this for the upstream PR.

PR #1247, opened 2021-10-07
Remove unused _f functions in P3

jgfouca:
For functions without array/view data, we no longer need `_f` functions. I'm not 100% sure we want to remove all this stuff because it could potentially be useful if we ever hit a really hard non-BFB problem between f90 and cxx that was not caught by the BFB unit tests, but it's not every day you can remove 2000+ lines of code without losing any capability.

I suppose we could always revert this PR if we really needed. Also, I don't think we will keep the P3 f90 implementation forever.


PeterCaldwell:
one thought is that I still want to be able to call our C++ functions from python. My recollection was that these _c or _f functions may be useful for that, so maybe we don't want to delete these things yet?

Another thought is that since these functions don't hurt anything right now, perhaps getting rid of them shouldn't be a high priority?

jgfouca:
@PeterCaldwell , I'm not familiar with python interfaces for p3. Is there any documentation on that?

jeff-cohere:
Hey @PeterCaldwell ,

I don't recall where the Python project was left, but if we wanted to use Python's foreign function interface (FFI) to call P3 directly, having C wrapper functions would be handy. If we're currently giving these wrapper functions an `_f` suffix, though, maybe we could just rename them to use `_c` to be a bit less insane about it.

Every language has to go through C to talk to any other language, so having C wrappers for physics functions can be a nice thing from a tooling perspective.

jeff-cohere:
I don't have a strong opinion about keeping these bridge functions around. It's very easy to wrap a C function around a C++ function if and when we need to do so. And being stingy about code we support is good for keeping maintenance costs down.

bartgol:
Spring is still a month away, but I already feel the spring-cleaning vibes. So what should we do with this PR? I am still in favor of merging it (after rebasing/retesting it, ofc).

@jgfouca @PeterCaldwell @jeff-cohere thoughts?

jeff-cohere:
I'm for merging it if everyone is comfortable doing so.

PeterCaldwell:
Thanks for the spring cleaning, Luca. After rereading the previous discussion, I vote for removing these _f functions based on the following logic:
1. the current _f functions haven't been exercised in ages so probably don't work anyways. We shouldn't keep code around that we're not regularly testing
2. the _f functions might be somewhat useful for interfacing with python, but we'd probably want to write our own wrappers since the needs of F90 and the needs of python aren't the same (c2py files, anyone?). We might as well just grab what information exists for these wrappers from an old hash rather than dragging all these files along in master though.
3. I expect external developers will want to hook their F90 parameterizations into our C++ version of the model in the not-too-distant future. Having an example _f file would be useful for that. But it sounds like some _f functions which include array/view data wouldn't be deleted so there's still an example to point to.

PeterCaldwell:
Oh, and I also (still) think that merging this PR is a good thing to do when someone needs a brain break but isn't a top priority.

jgfouca:
@bartgol , i can look at this and merge it.

jgfouca:
OK, this PR is ready to be reviewed and merged.

bartgol (approved):
Nice! I suspect this will have a non-negligible impact p3 compilation time on weaver.

jeff-cohere (approved):

PeterCaldwell:
Thanks for this, @jgfouca !!!

PR #1248, opened 2021-10-07
Add test to confirm valgrind errors are caught

jgfouca:
We were actually not catching these after the test-launcher PR went in.

Also, add some stubs for the upcoming thread/rank-spread test.

### DIFF
--- a/components/scream/tests/generic/infra/resource_spread.cpp
+++ b/components/scream/tests/generic/infra/resource_spread.cpp
@@ -1,0 +1,15 @@
+#include "ekat/kokkos/ekat_kokkos_types.hpp"
+
+#include <catch2/catch.hpp>
+
+#include <iostream>
+
+namespace scream {
+
+TEST_CASE("rank_and_thread_spread", "[fake_infra_test]")
+{
+  // TODO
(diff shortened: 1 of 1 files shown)


components/scream/tests/generic/infra/CMakeLists.txt
```diff
@@ -4,3 +4,10 @@ set(NEED_LIBS ekat scream_share)
 
 # Add the fake test
 CreateUnitTest(fake_test "fake_test.cpp" "${NEED_LIBS}")
+# CreateUnitTest(resource_spread_thread "resource_spread.cpp" "${NEED_LIBS}"
```
bartgol, line 7: Leftover from debugging? Or leftover from a different branch?

components/scream/tests/generic/infra/resource_spread.cpp
```diff
@@ -0,0 +1,15 @@
+#include "ekat/kokkos/ekat_kokkos_types.hpp"
```
bartgol, line 1: Is this file something to leave for another PR? Seems not completed...

Issue #837, opened 2020-12-15
Read in the appropriate data required for SCREAMv1 restart run


PeterCaldwell:
Make whatever code changes are needed to restart using SCREAMv1. We may want to do this before we set up the code for an "initial" run since new runs pull inputdata from a variety of sources while restarts get all their data from one source. We might be able to restart from a SCREAMv0 file (after variable name translation)?

Issue #791, opened 2020-12-08
[AD] SCORPIO-Interface: Allow for output on dynamics grid


AaronDonahue:
This would be a nice feature to not lose sight of.  The ability to write fields used by dynamics on the native dynamics mesh. 

This is low priority because it is not essential for any release and may require a fair bit of thought on how to make sure that the written field is on the dynamics grid.

Most likely there will be a need to remap from physics->dynamics, or store a local copy of the field while it is on the dynamics grid that will be used by I/O later.  Or call the output from dynamics.

In short - there would need to be some design choices before this task could be implemented.

bartgol:
Done in #1168 .

PR #1251, opened 2021-10-08
Add full atmosphere restart test

bartgol:
This PR aims at automatizing the AD capability of creating a restart file.

- [x] Fix atm proc group class to correctly expose input fields/groups
- [x] Make the AD create the "RESTART" group
- [x] Create a test that runs twice (much like the io 'restart' test): 1st time for N time step, second time restarts from N/2 timesteps.
- [x] Compare output of the two aforementioned tests.
- [x]  Make restart BFB (requires saving dyn grid state to restart file).


bartgol:
Update: the test runs, but there are still diffs between the monolithic 2 steps run, and the restarted run. I disabled all physics, and still get diffs, so (part of) the issue is in how the homme interface restarts homme's state. Interestingly, running the output managers right after init during the restarted run, yields a bfb output with the 1st timestep of the monolithic run.

bartgol:
Update: I have made changes in the grid stack, in particular in the SEGrid. It can now store either Continuous Galerkin (CG) or Discontinuous Galerkin dofs (DG) gids. The former corresponds to the old behavior, where corresponding edge dofs across bordering elements have the same gid, while the latter allows (in fact, forces) such dofs to have different gids. This will allow me to use a DG grid to write pio files with native dyn layout, while CG grid fields can be remapped to a PhysicsGLL grid (as before). In practice, all dyn fields will live on a DG grid. The CG grid is only used by the PD remapper, since it makes it possible to form the local map for the remapping.

The current tests now pass, but none of them writes/loads data to/from file with native dynamics layout (the only test that does dyn IO is `dyn_grid_io`, but it remaps to PhysicsGLL grid in the process). The next step is to hack IO classes to allow writing fields in native SEGrid layout.

AaronDonahue:
I think this PR will address #791

AaronDonahue:
I think this will also address #837

bartgol:
@AaronDonahue @tcclevenger @jeff-cohere This PR is probably ready for review. Some tests might still fail on some machines (they all pass on my laptop), but at least we can get a head start on the design review.

I pinged all of you, but I don't expect each of you to review all changes (there's a lot here!). As a guideline, I would like
- @AaronDonahue : review I/O stuff and tests
- @tcclevenger : review homme-related stuff
- @jeff-cohere : review share folder stuff (infrastructure, and abstract classes design)

Of course, if you want to review all of the PR, that'd be great! But I know it's a relatively large one, so I don't expect you to.

Thanks!

components/scream/src/share/atm_process/atmosphere_process.hpp
```diff
@@ -149,11 +159,14 @@ class AtmosphereProcess : public ekat::enable_shared_from_this<AtmosphereProcess
...
-  // Whether this atm proc requested the field/group as in/out/inout, via a FieldRequest/GroupRequest.
-  bool requires_field (const FieldIdentifier& id) const;
-  bool computes_field (const FieldIdentifier& id) const;
-  bool requires_group (const std::string& name, const std::string& grid) const;
-  bool computes_group (const std::string& name, const std::string& grid) const;
+  // The base class does not store internal fields.
+  virtual const std::list<field_type>& get_internal_fields  () const { return m_internal_fields; }
+
+  // Whether this atm proc requested the field/group as in/out, via a FieldRequest/GroupRequest.
+  bool is_required_field (const FieldIdentifier& id) const;
```
jeff-cohere, line 166: Not a big deal at all, but is there a reason you decided to change these method names? To me, `atm_proc.requires_field(f)` is a nice example of  the `noun.relates_to(other_noun)` idiom.

components/scream/src/share/atm_process/atmosphere_process_group.hpp
```diff
@@ -65,13 +65,17 @@ class AtmosphereProcessGroup : public AtmosphereProcess
   void set_required_field (const Field<const Real>& field);
   void set_required_group (const FieldGroup<const Real>& group);
 
+  // Gather internal fields from all processes in the group
+  void gather_internal_fields ();
```
jeff-cohere, line 69: Any rules about when this can/can't be called?

components/scream/src/share/field/field_manager.hpp
```diff
@@ -69,6 +70,11 @@ class FieldManager {
   void registration_ends ();
   void clean_up ();
 
+  // Adds a pre-existing field to the FieldManager.
```
jeff-cohere, line 73: Maybe "Adds a field to be managed by the FieldManager, that can be used by atmospheric processes." I was confused to read that the "pre-existing" field must not exist already. :-)

components/scream/src/share/field/field_property_checks/field_nan_check.hpp
```diff
@@ -107,6 +107,27 @@ class FieldNaNCheck: public FieldPropertyCheck<RealType> {
...
+      case 6:
+        {
+          auto v = field.template get_view<const_RT******>();
+          const int dim1 = dims[1];
+          const int dim2 = dims[2];
+          const int dim3 = dims[3];
+          const int dim4 = dims[4];
+          const int dim5 = dims[5];
+          Kokkos::parallel_reduce(dim0*dim1*dim2*dim3*dim4*dim5, KOKKOS_LAMBDA(int idx, int& result) {
+            const int i = ((((idx / dim5) / dim4) / dim3) / dim2) / dim1;
```
jeff-cohere, line 119: Do you think it's worth having an `unflatten(index, rank, dims, indices)` method on the field or layout object with the following signature, in which `dims` and `indices` are of length `rank`, and `i`, `j`, `k`, etc are stored in `indices` as a result of the method?

```
void unflatten(int flattened_index, int rank, int dims[], int indices[])
```

components/scream/src/share/field/field_utils.hpp
```diff
@@ -1,7 +1,9 @@
 #ifndef SCREAM_FIELD_UTILS_HPP
 #define SCREAM_FIELD_UTILS_HPP
 
-#include "field.hpp"
+#include "share/field/field.hpp"
```
jeff-cohere, line 4: On second thought, maybe `unflatten` could go here as a free function, but we'd have to add the field or layout as an arg.

components/scream/src/share/grid/abstract_grid.hpp
```diff
@@ -127,12 +102,24 @@ class AbstractGrid : public ekat::enable_shared_from_this<AbstractGrid>
   const lid_to_idx_map_type& get_lid_to_idx_map () const;
 
   // Set/get geometric views. The setter is virtual, so each grid can check if "name" is supported.
```
jeff-cohere, line 104: This function is no longer virtual, so this comment should be removed.

components/scream/src/share/grid/abstract_grid.hpp
```diff
@@ -127,12 +102,24 @@ class AbstractGrid : public ekat::enable_shared_from_this<AbstractGrid>
...
 
-  void set_unique_grid (const std::shared_ptr<const AbstractGrid>& unique_grid);
+  // Derived classes can override these methods, which are called inside the
+  // set_dofs and set_lid_to_idx_map methods respectively, to verify that the
+  // views have been set to something that satisfies any requirement of the grid type.
+  // This class already checks the extents of the view, but derived classes can add
+  // some extra consistency check.
+  // NOTE: these methods are called *at the end* of the respective set_* method,
+  //       so that other methods can be called (e.g., get_global_max_dof_gid()).
+  virtual void check_dofs_list () const {}
```
jeff-cohere, line 117: Do you have a one-liner that describes what this checks, and what it does in case the check passes/fails?

jeff-cohere: Also, I don't know where the conversation in the C++ community is regarding the use of exceptions as super-fancy `if` tests, but if we want to programmatically check things and correct them if they're deficient, I would prefer to have this return a boolean that indicates whether the check passed or failed. If, on the other hand, failing this check is always a fatal or uncorrectable error, the exception is more appropriate

components/scream/src/share/io/scorpio_input.hpp
```diff
@@ -93,6 +93,7 @@ class AtmosphereInput
   //               variables from the input filed will be read into.
   //               Fields can be padded/strided.
   // It calls init(field_mgr) at the end.
+  // TODO: is comm superfluous, considering we can get it from the grid in the field_mgr?
```
jeff-cohere, line 96: This is a good question! I think the answer is the same as the answer to "is it invalid for this comm to be different from the one used by the grid in the field_mgr?"

components/scream/src/share/io/scorpio_input.hpp
```diff
@@ -120,35 +122,33 @@ class AtmosphereInput
...
-  //               Fields can be padded/strided.
-  void init();
 
-  // Sets up the scorpio metadata to preare for reading
-  // Inputs:
-  //  - host_views_1d: the 1d flattened views where data will be read into.
-  //                   These views must be contiguous (no padding/striding).
-  //  - layouts: the layout of the vars (used to reshape the views).
-  void set_parameters (const ekat::ParameterList& params);
+  void set_parameters (const ekat::ParameterList& params, const std::string& grid_name = "");
```
jeff-cohere, line 141: What does it mean for the grid name to be left blank here?

components/scream/src/share/io/scorpio_input.cpp
```diff
@@ -279,6 +284,48 @@ void AtmosphereInput::read_variables (const int time_index)
               }}}
               break;
             }
+          case 4:
```
jeff-cohere, line 287: This smells like something that might go in `field_utils` as well.

components/scream/src/share/util/scream_time_stamp.cpp
```diff
@@ -139,7 +149,8 @@ TimeStamp& TimeStamp::operator+=(const int seconds) {
   auto& mm = m_date[1];
   auto& yy = m_date[0];
 
-  EKAT_REQUIRE_MSG (seconds>=0, "Error! Cannot rewind time, sorry.\n");
+  EKAT_REQUIRE_MSG (seconds>0, "Error! Time must move forward.\n");
```
jeff-cohere, line 152: Indeed.

components/scream/src/physics/spa/CMakeLists.txt
```diff
@@ -1,14 +1,4 @@
-set(CLDFRAC_SRCS
```
AaronDonahue, line 1: wow, good catch Luca, this is clearly copy/paste gone wrong.

components/scream/src/share/io/tests/io.cpp
```diff
@@ -374,7 +378,7 @@ ekat::ParameterList get_in_params(const std::string type,
       + "." + t_first_write.to_string() + ".nc";
 
   in_params.set<std::string>("Filename",filename);
-  in_params.set<vos_type>("Fields",{"field_1", "field_2", "field_3", "field_packed"});
+  in_params.set<vos_type>("Fields Names",{"field_1", "field_2", "field_3", "field_packed"});
```
jeff-cohere, line 381: This might be too much of a pain, but what do you think about `"Field Names"` instead of `"Fields Names"`?

components/scream/src/share/io/scorpio_input.cpp
```diff
@@ -86,112 +106,97 @@ set_field_manager (const std::shared_ptr<const fm_type>& field_mgr,
...
+void AtmosphereInput::
+register_fields_specs() {
+  for (auto const& name : m_fields_names) {
+    auto f = m_field_mgr->get_field(name);
+    const auto& fh  = f.get_header();
+    const auto& fap = fh.get_alloc_properties();
+    const auto& fid = fh.get_identifier();
+    const auto& fl  = fid.get_layout();
+
+    // Store tha layout
```
AaronDonahue, line 177: "the"

components/scream/src/share/io/scorpio_output.hpp
```diff
@@ -28,14 +29,20 @@
  *  Averaging Type:               STRING
  *  Max Snapshots Per File:       INT                   (default: 1)
  *  Fields:
- *     GRID_NAME_1:               ARRAY OF STRINGS
- *     GRID_NAME_2:               ARRAY OF STRINGS
+ *     GRID_NAME_1:
+ *        Fields Names:            ARRAY OF STRINGS
```
AaronDonahue, line 33: extra whitespace before "ARRAY OF STRINGS"

jeff-cohere (commented):
Really good work here, Luca. I've left several comments for your consideration. I stuck mostly to the `src/share` directory but also gave the tests a glance. Let me know if you would like feedback on anything that I haven't commented on.

components/scream/src/share/scream_types.hpp
```diff
@@ -31,6 +31,12 @@ enum class RepoState {
   Closed
 };
 
+// The type of this run
```
AaronDonahue, line 34: this is great.

components/scream/tests/coupled/dynamics_physics/homme_shoc_cld_p3_rrtmgp/homme_shoc_cld_p3_rrtmgp_output.yaml
```diff
@@ -5,58 +5,58 @@ Averaging Type: Instant
...
-    - ni_activated
-    - nc_nuceat_tend
-    - eff_radius_qc
-    - eff_radius_qi
-    - sfc_alb_dir_nir
-    - sfc_alb_dir_vis
-    - sfc_alb_dif_nir
-    - sfc_alb_dif_vis
-    - surf_lw_flux_up
+    Fields Names:
```
AaronDonahue, line 8: Does this change need to be made to the other uncoupled and coupled  tests?

AaronDonahue:
This is an epic PR.  My first read through I didn't spot anything that jumped out at me.  I left a few minor comments along the way.  I have a few meetings now, I'm hoping to go through it again after just to make sure since it is so big.

tcclevenger (commented):
Nice work, @bartgol. Like Aaron, first look through nothing huge stood out to me. I'll end up looking back through it later to make sure. I found a typo in the homme stuff that I commented on.

components/scream/src/share/atm_process/atmosphere_process.hpp
bartgol: I didn't have an equivalent for "internal" fields... So to keep the impl for required, computed, internal somewhat parallel, I switched to this wording. I can change to `has_blah_field(f)`, with blah in `required, computed, internal` maybe?

components/scream/src/share/atm_process/atmosphere_process_group.hpp
bartgol: Ah, yes, this must be called before any attempt to query the list of internal fields for the group. I will add a note in the comment.

components/scream/src/share/field/field_property_checks/field_nan_check.hpp
bartgol: Uhm, it might be useful, but I'm worried about how to use it from device. I am not 100% sure, but I don't think device lambdas can capture host static arrays (such as dims). Though we could declare `int indices[6];` locally, and call `unflatten(idx, 6, dim0, dim1,dim2,dim3,dim4,dim5,indices);`.

If the kokkos layout stored a 1d view for the (physical) extents, then we could probably do it more easily:
```
const auto& layout = f.get_header().get_layout();
const auto extents = layout.extents(); // or dims_view(), or whatever
Kokkos::parallel_for(layout.size(),KOKKOS_LAMBDA(const int idx){
  int indices[6];
  unflatten_indices<6> (idx, extents, indices);
  ...
});
```
Imho, it's a good v1.1 issue.

bartgol: We could offer also an overload where indices are N separate ints, for ease of use. Like:

```
int indices[6];
int i,j,k,l,m,n;
unflatten<6>(idx,extents,indices); // with array
unflatten<6>(idx,extents,i,j,k,l,m,n);
v(i,j,k,l,m,n) = blah; // Concise
v(indices[0], indices[1], indices[2], indices[3], indices[4], indices[5]) = blah; // Lost in parsing...
```

components/scream/src/share/field/field_utils.hpp
bartgol: Could also be a (kokkos-friendly) method of `FieldLayout`.

components/scream/src/share/grid/abstract_grid.hpp
bartgol: That's a good point. I think I can easily turn them into methods returning bool's. The only place that uses them will always throw if dofs are bad, but returning true/false is definitely more flexible.

components/scream/src/share/io/scorpio_input.hpp
bartgol: True. I can see how this could be potentially buggy, so maybe I should remove the comm from the inputs. And if one uses that basic ctor (comm+params) and calls `init(field_mgr,grids_mgr)` later, we should check that the stored comm matches the one in the field mgr (and grids mgr)

components/scream/src/share/io/scorpio_input.hpp
bartgol: This is a poor design choice on my end. It all stems from how the parameter list is shaped, and allowing "simple" syntax for unit tests.

The general use case requires the fields names to be specified like so:
```
Fields:
  Grid Name 1:
     Fields Names: [a, b, c, ...]
  Grid Name 2:
     Fields Names: [a2, b2, ...]
```
But for single-grid tests, it seemed silly to request the user to know/set the grid name, so I allowed the syntax
```
Fields Names: [a, b, c, ...]
```
directly in the root parameter list. Now, the class offers a minimal ctor, where only comm and params are passed. This can be used if a) you don't want to read fields, only some global attributes (e.g., some counters), or if b) you need to read in some info first, before being able to pass in the grids (e.g., spa first reads ncols from nc file, then creates the views, then calls `init`, passing in the newly created views).

So the ctor where the field mgr (and the grid) is passed can call `set_parameters` passing the grid name, and therefore read the sublist `Fields->GridName->Field Names`. OTOH, the other ctors will not pass a grid name, so `set_parameters` will try to look for the `Field Names` entry, and if not found leave `m_fields_names` unset.

bartgol: NOTE: I'm not fond of how this ended up. An alternative, is to store the parameter list internally, and query it later for the fields names (if/when the grid name is already available).

components/scream/src/share/io/scorpio_input.cpp
bartgol: Maybe, yes. The src view is not a field, but we could have a method `void deep_copy(Field<Real>& f, view_1d_type<Real> src);`. I think the `deep_copy` method of `Field` could be implemented in terms of this too...

components/scream/src/share/io/tests/io.cpp
bartgol: Not a pain at all. I actually wondered quite a bit what was the correct English formulation between the two. If you say `Field Names` is the right one, I can switch them all.

components/scream/tests/coupled/dynamics_physics/homme_shoc_cld_p3_rrtmgp/homme_shoc_cld_p3_rrtmgp_output.yaml
bartgol: I thought I changed all the yaml files. But I now see I missed some, probably due to the rebase following your tests folder changing PR. I'll fix those.

components/scream/src/share/io/scorpio_input.cpp
bartgol: OTOH, implementing `Field::deep_copy(src)` in terms of the new util would create a circular dep between `field.hpp` and `field_utils.hpp`... So I might punt on this.

components/scream/src/share/field/field_utils.hpp
bartgol: I went for this file, with some help from FieldLayout.

bartgol:
@jeff-cohere @AaronDonahue @tcclevenger I addressed most of your comments. Jeff, I still haven't addressed the `is_required_field` comment. Let me know if there's an option you like better than the current one. I would really like to keep a symmetry between required, computed, and internal.

components/scream/src/share/atm_process/atmosphere_process.hpp
jeff-cohere: Aha, makes sense. I like `atm_proc.has_blah_field(f)`, actually. If that's okay with you, let's do it!

jeff-cohere:
> @jeff-cohere @AaronDonahue @tcclevenger I addressed most of your comments. Jeff, I still haven't addressed the `is_required_field` comment. Let me know if there's an option you like better than the current one. I would really like to keep a symmetry between required, computed, and internal.

Looks great! I think `is_{required,computed,internal}_field` is a good solution. Thanks for all the hard work, dude!

jeff-cohere (approved):

tcclevenger (approved):

bartgol:
A few last fixes were needed. Things should work now.

bartgol:
There's still one failure on blake. The model restart does not appear to be BFB in release mode. I will investigate.

AaronDonahue (approved):
I did my second read-through.  Epic PR.  It looks like its failing the AT so we will want to check on that, but otherwise I think this is ready to go.

bartgol:
Update: if Homme is disabled, then the restart test passes on blake (release build). OTOH, the test with Homme fails even with pack size 1 (so it's not pack-related).

bartgol:
I take it back. Pack size 1 seems to pass on blake, hinting at some non BFB issue related to vectorization. This is somewhat odd, considering that pack size 16 is also tested on mappy, where it passes. It might be there's an operation done on packed data during homme's post processing but on unpacked data during restart bootstrap (or viceversa), which only intel happens to expose.

bartgol:
@AaronDonahue I think I fixed the restart test. If you re-approve, we can merge if/when the AT passes.

AaronDonahue (approved):
Great! Glad you found the issue

bartgol:
Spoke too  soon...But I swear I got it passing on blake at some point. Not sure what happened.

bartgol:
As soon as #1349 is merged, this should pass. @AaronDonahue @jeff-cohere @tcclevenger , if you want to give a final review, there's only a handful of commits since your reviews, so it should be quick (if you limit the review to the last commits).

tcclevenger (approved):
Looks good to me.

jeff-cohere (approved):
Looks good, Luca.

AaronDonahue (approved):
Looks good.  Glad we can finally get this merged.

bartgol:
#1349 went in, so now I'm manually testing on mappy again, to verify everything still works.

bartgol:
I fixed some issues related to hydro-vs-nonhydro mode for Homme. I want to test again on weaver before merging, but it's currently down, so I'll likely pick this up on Monday.

AaronDonahue:
@bartgol , does this PR need to re-approved and tested on mappy?  With #1349 merged it should be ready to go right?

AaronDonahue (approved):

bartgol:
@AaronDonahue Well, almost. There was still an issue, and tests failed everywhere. I fixed it with the last commit, but it could not be tested cause weaver was offline. Let's give the AT one last round of testing before merging. Meanwhile, you can review the last commit, and approve/comment.

bartgol:
@AaronDonahue Can I merge this?

bartgol:
Retesting, since this PR became stale.

AaronDonahue:
@bartgol , yes.  I suppose once retesting is finished and passes I say go for it.

components/scream/src/share/util/scream_time_stamp.cpp
```diff
@@ -65,8 +65,9 @@ TimeStamp::TimeStamp(const std::vector<int>& date,
   const auto sec  = time[2];
 
   // Check the days and seconds numbers are non-negative.
-  EKAT_REQUIRE_MSG (mm>0   && mm<=12, "Error! Month out of bounds.\n");
-  EKAT_REQUIRE_MSG (dd>0   && dd<=days_in_month(yy,mm), "Error! Day out of bounds.\n");
+  EKAT_REQUIRE_MSG (yy>=0   && yy<=9999, "Error! Year out of bounds.\n");
```
AaronDonahue, line 68: Do we really need an upper bound on the year?  I'm sure we won't be running simulations for a millennium, so maybe this isn't an issue.  But it feels like an unnecessary limitation.  Who knows what the future holds in terms of simulations.

AaronDonahue (approved):
Just one comment that shouldn't hold this PR going in.

components/scream/src/share/util/scream_time_stamp.cpp
bartgol: Well, I wanted to put a check that caught possibly uninitialized inputs. No really compelling reasons though.

PR #1252, opened 2021-10-11
Add new compset

tcclevenger:
Adds new compset for CIME testing.


components/scream/data/scream_input.yaml
```diff
@@ -20,7 +20,16 @@ Atmosphere Driver:
     Process 4:
       Process Name: RRTMGP
       Grid: Physics GLL
-      active_gases: ["h2o", "co2", "o3", "n2o", "co" , "ch4", "o2", "n2"]
+      active_gases:
```
AaronDonahue, line 23: I like this use of the YAML syntax.  We may consider doing this for the unit tests too, but not in this PR.

PR #1258, opened 2021-10-12
Remove z from FM

tcclevenger:
Removes `z_int` and `z_mid` from the Field Manager, instead calculating them on the fly in SHOC AD and `SurfaceCoupling`.

Changes:
1. Added `PhysicsFunctions::calculate_z_mid()`. All this does is call `ColumnOps::compute_midpoint(z_int)`.
2. Removed unused field request for `z_int` in P3 AD.
3. Change `PhysicsFunctions::calculate_z_int()` to be templated on `ViewT` instead of explicitly stating the view type. I was running into issues of it expecting a `MemoryManaged` view, but shoc ad and SC want to pass `ekat::subview()` which are `Unmanaged`.

### DIFF
--- a/components/scream/src/control/surface_coupling.cpp
+++ b/components/scream/src/control/surface_coupling.cpp
@@ -1,7 +1,6 @@
 #include "control/surface_coupling.hpp"
 
 #include "share/field/field_utils.hpp"
-#include "share/util/scream_common_physics_functions.hpp"
 
 namespace scream {
 namespace control {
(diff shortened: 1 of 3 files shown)


components/scream/src/physics/shoc/atmosphere_macrophysics.hpp
```diff
@@ -119,32 +111,36 @@ class SHOCMacrophysics : public scream::AtmosphereProcess
...
-        auto range_pack1 = ekat::range<IntSmallPack>(k*Spack::n);
-        auto range_pack2 = range_pack1;
-        range_pack2.set(range_pack1 > nlev, 1);
-        ekat::index_and_shift<1>(s_z_int, range_pack2, zi_k, zi_kp1);
-        dz(i,k).set(range_pack1 < nlev, zi_k - zi_kp1);
+        // Vertical layer thickness
+        dz(i,k) = PF::calculate_dz(pseudo_density(i,k), p_mid(i,k), T_mid(i,k), qv(i,k));
 
-        zt_grid(i,k) = z_mid(i,k) - z_int(i, nlevi_v)[nlevi_p];
         rrho(i,k) = (1/ggr)*(pseudo_density(i,k)/dz(i,k));
```
bartgol, line 121: Very minor: we might as well precompute 1/ggr outside the loops...

components/scream/src/physics/shoc/atmosphere_macrophysics.hpp
```diff
@@ -89,23 +89,15 @@ class SHOCMacrophysics : public scream::AtmosphereProcess
...
         const auto range = ekat::range<IntSmallPack>(k*Spack::n);
         const Smask in_nlev_range = (range < nlev);
 
         // Inverse of Exner. Assert that exner != 0 when in range before computing.
-        const Spack p_mid_ik(p_mid(i,k));
-        const Spack exner_ik = PF::exner_function(p_mid_ik);
-        const Smask nonzero = (exner_ik != 0);
+        const Spack exner = PF::exner_function(p_mid(i,k));
+        const Smask nonzero = (exner != 0);
         EKAT_KERNEL_ASSERT((nonzero || !in_nlev_range).any());
```
bartgol, line 99: What's the reason behind the kernel assert above? I would expect (nonzero || !in_nlev_range).all(), no?

components/scream/src/physics/shoc/atmosphere_macrophysics.cpp
```diff
@@ -281,9 +283,12 @@ void SHOCMacrophysics::initialize_impl (const util::TimeStamp& /* t0 */)
   auto tke_copy    = m_buffer.tke_copy;
   auto shoc_ql2    = m_buffer.shoc_ql2;
 
-  shoc_preprocess.set_variables(m_num_cols,m_num_levs,m_num_tracers,m_cell_area,
-                                T_mid,z_int,z_mid,p_mid,pseudo_density,omega,phis,surf_sens_flux,surf_latent_flux,
-                                surf_mom_flux,qv,qc,tke,tke_copy,cell_length,
+  // For now, set z_int(i,nlevs) = z_surf = 0
+  const Real z_surf = 0.0;
```
bartgol, line 287: You already have a `z_surf` member of this class. This will generate a warning. Either delete the class member, or use that (instead of the temporary).

bartgol: Nvm, I think the member was inside SHOCPreprocess.

components/scream/src/share/util/scream_column_ops.hpp
```diff
@@ -431,6 +431,9 @@ class ColumnOps {
                     const view_1d<ScalarT,MT>& x_i,
                     const scalar_type& s0 = zero())
   {
+    EKAT_KERNEL_ASSERT_MSG(pack_size<ScalarT>() <= num_mid_levels,
```
bartgol, line 434: Did you check whether it works for pack_size==nlevs?

components/scream/src/share/util/scream_common_physics_functions.hpp
```diff
@@ -351,15 +351,26 @@ struct PhysicsFunctions
   // Note: because this function does an integral it cannot be run just on a single level.  It requires
   // the full column wise integration.
   //-----------------------------------------------------------------------------------------------//
-  template<typename ScalarT, typename InputProviderZ>
+  template<typename InputProviderZ, typename ViewT>
```
bartgol, line 354: Templating on the view type is fine, but as a reference, another valid approach is to do
```
template<typename ScalarT, typename InputProviderZ, typename MT = Kokkos::MemoryManaged>
static void calculate_z_int (...., const view_1d<ScalarT, MT>& z_int);
```
since `view_1d` in KokkosTypes supports also an optional memory traits template parameter (which defaults to MemoryManaged).

components/scream/src/physics/shoc/atmosphere_macrophysics.hpp
```diff
@@ -89,23 +89,15 @@ class SHOCMacrophysics : public scream::AtmosphereProcess
...
-        inv_exner(i,k).set(nonzero, 1/exner_ik);
+        inv_exner(i,k).set(nonzero, 1/exner);
```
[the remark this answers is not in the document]
bartgol: Actually, it doesn't really matter, I guess. Unless one uses the same result many times (in which case, one can simply do  `auto inv_exner = 1/exner;`...).

components/scream/src/control/surface_coupling.cpp
```diff
@@ -316,21 +316,23 @@ void SurfaceCoupling::do_export (const bool init_phase)
...
-      });
+      // Compute vertical layer thickness
+      PF::calculate_dz(team, pseudo_density_i, p_mid_i, T_mid_i, qv_i, dz_i);
       team.team_barrier();
 
-      // Compute vertical layer heights
-      PF::calculate_z_int(team,m_num_levs,dz,0.0,z_int);
+      // Compute vertical layer heights. Use z_int(nlevs) = z_surf = 0.0.
+      const Real z_surf = 0.0;
+      PF::calculate_z_int(team, m_num_levs, dz_i, z_surf, z_int_i);
```
bartgol, line 329: Since you are using `KOKKOS_LAMBDA`, you can't access class members from inside. Here, `m_num_levs` causes a `this->m_num_levs` access on device. You need a local copy before the parallel for.

bartgol (changes requested):
I saw one last thing to address, then I think it's good.

bartgol (approved):

PR #1259, opened 2021-10-13
Change handling of dt and time stamps

bartgol:
This PR does a few things:

* Refactors a bit the TimeStamp class: clean up, and different internal storage.
* Changes the time step to be an integer (same thing that is done in EAM).
* Changes HommeDynamics implementation, fixing dt handling (dt_atm != dt_homme, and dt_atm may not be the same as dt_homme*rsplit*qsplit).
* [Edit] Improves output for most system tests, and make them configurable (can choose number of iterations based on testing profile).

### DIFF
--- a/components/scream/src/control/tests/dummy_atm_proc.hpp
+++ b/components/scream/src/control/tests/dummy_atm_proc.hpp
@@ -67,7 +67,7 @@ public:
 
 protected:
 
-  void initialize_impl (const util::TimeStamp&) {
+  void initialize_impl () {
     // Do nothing
   }
 
--- a/components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
+++ b/components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
@@ -265,7 +265,7 @@ void HommeDynamics::init_buffers(const ATMBufferManager &buffer_manager)
   fbm.allocate(buffer_manager.get_memory(), buffer_manager.allocated_bytes()/sizeof(Real));
 }
 
-void HommeDynamics::initialize_impl (const util::TimeStamp& /* t0 */)
+void HommeDynamics::initialize_impl ()
 {
   const auto& dgn = m_dyn_grid->name();
   const auto& rgn = m_ref_grid->name();
--- a/components/scream/src/dynamics/homme/atmosphere_dynamics.hpp
+++ b/components/scream/src/dynamics/homme/atmosphere_dynamics.hpp
@@ -65,7 +65,7 @@ protected:
   // Updates p_mid
   void update_pressure ();
 
-  void initialize_impl (const util::TimeStamp& t0);
+  void initialize_impl ();
 protected:
   void run_impl        (const int dt);
   void finalize_impl   ();
--- a/components/scream/src/physics/cld_fraction/atmosphere_cld_fraction.cpp
+++ b/components/scream/src/physics/cld_fraction/atmosphere_cld_fraction.cpp
@@ -52,7 +52,7 @@ void CldFraction::set_grids(const std::shared_ptr<const GridsManager> grids_mana
 }
 
 // =========================================================================================
-void CldFraction::initialize_impl (const util::TimeStamp& /* t0 */)
+void CldFraction::initialize_impl ()
 {
   // Set property checks for fields in this process
   auto frac_interval_check = std::make_shared<FieldWithinIntervalCheck<Real> >(0,1);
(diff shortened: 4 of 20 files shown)


components/scream/src/share/tests/utils_tests.cpp
```diff
@@ -38,3 +39,34 @@ TEST_CASE("field_layout") {
...
+  REQUIRE (ts1.get_seconds()==30);
+
+  // Julian day = day_of_year.fraction_of_day, with day_of_year=0 at Jan 1st.
+  REQUIRE (ts1.get_julian_day()==(284 + (17*3600+8*60+30)/86400.0));
+
+  REQUIRE (ts1.get_date_string()=="2021-10-12");
+  REQUIRE (ts1.get_time_string()=="17:08:30");
+  REQUIRE (ts1.to_string()=="2021-10-12 17:08:30");
+
+  // Comparisons
```
AaronDonahue, line 62: Could you also make a few tests which check that adding 31 days, for example, switches the month.  Or adding 365 days switches the year?  Would that be helpful?

components/scream/tests/uncoupled/spa/CMakeLists.txt
```diff
@@ -1,14 +1,19 @@
...
+set (NEED_LIBS spa scream_control scream_share)
 CreateUnitTest(spa_stand_alone "spa_stand_alone.cpp" "${NEED_LIBS}" LABELS "spa;physics")
 
-# Copy yaml input file to run directory
-CONFIGURE_FILE(${CMAKE_CURRENT_SOURCE_DIR}/input.yaml
-               ${CMAKE_CURRENT_BINARY_DIR}/input.yaml COPYONLY)
-CONFIGURE_FILE(${CMAKE_CURRENT_SOURCE_DIR}/spa_init_ne2np4.nc
+# Set AD configurable options
+SetVarDependingOnTestProfile(NUM_STEPS 2 24 48)
+set (ATM_TIME_STEP 1800)
```
AaronDonahue, line 9: I think the spa_stand_alone timestep was 864000s before.

components/scream/tests/uncoupled/spa/CMakeLists.txt
```diff
@@ -1,14 +1,19 @@
...
+# Create the test
+set (NEED_LIBS spa scream_control scream_share)
 CreateUnitTest(spa_stand_alone "spa_stand_alone.cpp" "${NEED_LIBS}" LABELS "spa;physics")
 
-# Copy yaml input file to run directory
-CONFIGURE_FILE(${CMAKE_CURRENT_SOURCE_DIR}/input.yaml
-               ${CMAKE_CURRENT_BINARY_DIR}/input.yaml COPYONLY)
-CONFIGURE_FILE(${CMAKE_CURRENT_SOURCE_DIR}/spa_init_ne2np4.nc
+# Set AD configurable options
+SetVarDependingOnTestProfile(NUM_STEPS 2 24 48)
```
AaronDonahue, line 8: I like this functionality for controlling tests locally.

components/scream/src/share/tests/utils_tests.cpp
bartgol: Yeah, that's actually useful. I'll add it.

components/scream/src/dynamics/homme/interface/homme_params_mod.F90
```diff
@@ -524,10 +479,43 @@ function get_homme_bool_param_f90 (param_name_c) result(param_value) bind(c)
...
+    !
+    integer (kind=c_int) :: atm_dt
+    !
+    ! Local(s)
+    !
+    integer (kind=c_int) :: nsplit_out
+    integer :: ierr, nstep_factor
+
+    nsplit_out = -1
+    ierr = compute_nsplit(par, dt_remap_factor, dt_tracer_factor, nsplit_out, nstep_factor, tstep, atm_dt)
```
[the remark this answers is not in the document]
bartgol: Yeah, unfortunately no. The `dt_xyz_factor` states how many dts are in a xyz step. So `dt_remap_factor` states how many dt's are in a remap, and same for tracer. So those two are fully internal in homme.

We actually had a discussion in the homme slack channel. It seems that the "ideal" scenario would be for atm to simply pass dt_atm to homme, and let homme handle all the subcycling internally (including computing nsplit,rsplit,qsplit, or their "factor" counterparts). That would allow atm to be fully agnostics to the fact that dyn might have certain dt restrictions. But in a way, scream _is_ agnostic to that: it's just HommeDynamics that knows about homme split factors, and it's somewhat semantical the distinction of whether HommeDynamics belongs to scream or Homme.

PR #1261, opened 2021-10-13
Add test for resource spreading

jgfouca:
Only works on mappy for now. Not sure how to do this on GPUs to ensure proper spreading across resources.

One thing that bothers me a little bit is that SCREAM developers no longer get good spreading by default when using `ctest` manually. You have to have a resource file in order to get proper spreading. @bartgol , maybe that's something we can build into CMake/Ctest for SCREAM?


components/scream/scripts/scripts-tests
```diff
@@ -484,8 +500,19 @@ class TestTestAllScream(TestBaseOuter.TestBase):
...
         """
-        Test that MPI ranks and threads are properly spread across a machine's computing resources (TODO)
+        Test that MPI ranks and threads are properly spread across a machine's computing resources
         """
+        if self._full and self._machine == "mappy":
+            cmd = self.get_cmd("./test-all-scream -x -e OMP_DISPLAY_ENV=True -e SCREAM_TEST_THREAD_SPREAD=True -m $machine -b HEAD -k -t dbg --ctest-parallel-level=40 ", self._machine, dry_run=False)
+            output = run_cmd_assert_result(self, cmd, from_dir=TEST_DIR)
+            test_omp_spread(self, output, 40)
+
+            cmd = self.get_cmd("./test-all-scream -x -e OMP_DISPLAY_ENV=True -e SCREAM_TEST_RANK_SPREAD=True -m $machine -b HEAD -k -t dbg --ctest-parallel-level=40 ", self._machine, dry_run=False)
```
bartgol, line 510: Why the `-k` flag?

components/scream/scripts/scripts-tests
```diff
@@ -125,6 +125,22 @@ def test_branch_switch(test_obj, repo, branch_name):
...
+    for line in lines:
+        if "OMP_PLACES = '" in line:
+            places_str = line.split("=")[1].strip().strip("'")
+            places = [int(item.strip("{}")) for item in places_str.split(",")]
+            all_places.extend(places)
+
+    all_places.sort()
+    test_obj.assertEqual(len(all_places), num_places)
+    for idx, place in enumerate(all_places):
+        test_obj.assertEqual(idx, place)
```
bartgol, line 142: What if `all_places` contains duplicates? Shouldn't we make it unique first?

Edit: Ah, is it b/c when this test is run, only the `resource_spread_rank` and `resource_spread_thread` are used?

Edit2: Wait, those tests still add up to 40+40=80 threads used, which is likely to be more than the machine has, which means that my comment might still be correct.

bartgol: Also, I think this is testing that all corese 0-N are used, but it's not testing that each test spreads the threads on different cores.

components/scream/scripts/test-all-scream
```diff
@@ -122,6 +122,9 @@ OR
     parser.add_argument("--ctest-parallel-level", action="store", type=int, default=0,
         help="Force to use this value for CTEST_PARALLEL_LEVEL. If not provided, use default for given machine.")
 
+    parser.add_argument("-x", "--extra-verbose", action="store_true",
```
bartgol, line 125: Probably personal: any reason not to use the common `-v --verbose` flag, like most programs? "extra" makes me think we also have an option for "normal verbose", and "not verbose"...

components/scream/scripts/test_all_scream.py
```diff
@@ -526,8 +528,9 @@ def generate_ctest_config(self, cmake_config, extra_configs, test):
         test_dir = self.get_test_dir(self._work_dir,test)
         num_test_res = self.create_ctest_resource_file(test,test_dir)
         cmake_config += " -DSCREAM_TEST_MAX_TOTAL_THREADS={}".format(num_test_res)
+        verbosity = "-V --output-on-failure" if not self._extra_verbose else "-VV"
```
bartgol, line 531: I know this was already the case in master, but why do we need `-V` when we already have `--output-on-failure`? On my laptop, `-V --output-on-failure` is the same as `-V`, with all test output being printed. Also, I don't get any difference between `-V` and `-VV` on my laptop. CTest docs are not very clear on which "extra" output should be shown by `-VV` that is not shown by `-V`.

Can we just do `verbosity = "-V" if self._extra_verbose else "--output-on-failure"`?

components/scream/tests/generic/infra/CMakeLists.txt
```diff
@@ -4,10 +4,11 @@ set(NEED_LIBS ekat scream_share)
...
-# CreateUnitTest(resource_spread_rank "resource_spread.cpp" "${NEED_LIBS}"
-#   THREADS 4
-#   MPI_RANKS 1 4 1)
+CreateUnitTest(resource_spread_thread "resource_spread.cpp" "${NEED_LIBS}"
+  THREADS 1 4 1
+  MPI_RANKS 4)
+
+CreateUnitTest(resource_spread_rank "resource_spread.cpp" "${NEED_LIBS}"
+  THREADS 4
+  MPI_RANKS 1 4 1)
```
bartgol, line 14: Hard coding the max threads/ranks number might give headaches to some extra diligent developer who runs these tests on a small machine (perhaps b/c they suspect something messed up spreading). These two sets of tests have a max of 16 threads needed. Although that's small for clusters, some very diligent dev might run them on a workstation with 10 cores, and get config errors. We could use `SCREAM_TEST_MAX_RANKS` and `SCREAM_TEST_MAX_THREADS`, to give them the ability to stay within machine bounds.

bartgol (commented):
I have some comments. Some b/c I don't fully understand, some are suggestions which you may as well discard.

components/scream/scripts/scripts-tests
bartgol: Ah, I didn't realize we changed the repo during the scripts test.

components/scream/scripts/scripts-tests
bartgol: Ah, I see. Would be nice to also run a test where we ensure that separate tests are spread out. Not sure if you want to add that here though.

components/scream/scripts/scripts-tests
jgfouca: @bartgol , we don't. But it is possible the user could have a dirty repo when they run `scripts-tests`

PR #1203, opened 2021-09-20
Aarondonahue/add cprnc testing to adtests

AaronDonahue:
This commit will expand all tests that run through the ad, i.e. those located in `components/scream/tests`, to run with multiple ranks and check for BFB between the outputs.  Most of the changes are in the cmake structure and in the control files to accomodate running with multiple ranks.

However, there is now a rrtmgp standalone test that runs multiple steps through the AD.  The original rrtmgp standalone test compared the results after 1 step to a known "all-sky" baseline.  That test is retained, but renamed to `rrtmgp_standalone_unit`, while `rrtmgp_standalone` now refers to a simulation that is initialized by file and run for multiple steps.

PR #1269, opened 2021-10-20
add cprnc testing to adtests

AaronDonahue:
This commit will expand all tests that run through the ad, i.e. those located in components/scream/tests, to run with multiple ranks and check for BFB between the outputs. Most of the changes are in the cmake structure and in the control files to accomodate running with multiple ranks.

However, there is now a rrtmgp standalone test that runs multiple steps through the AD. The original rrtmgp standalone test compared the results after 1 step to a known "all-sky" baseline. That test is retained, but renamed to rrtmgp_standalone_unit, while rrtmgp_standalone now refers to a simulation that is initialized by file and run for multiple steps.

Replaces #1203


components/scream/src/share/io/scorpio_output.cpp
```diff
@@ -106,8 +106,12 @@ void AtmosphereOutput::init()
 /* Overload the run routine to accept a TimeStamp or floating point for the input time */
 void AtmosphereOutput::run(const util::TimeStamp& time)
 {
-  // Pass the time in seconds and as a string to the run routine.
-  run_impl(time.get_seconds(),time.to_string());
+  if (m_filename_with_time_string) {
```
bartgol, line 109: Perhaps it would be easier to add a `run ()` method, without timestamp input? I don't want to give users the option of accidentally disabling time string in the filename during normal runs. Adding a `run()` interface is probably safer, in that the AD will _never_ call this method, ensuring that no file overwrite happens in production runs.

AaronDonahue: but for the tests I want to run using this method I am running through the AD.

AaronDonahue:
We have broken this PR into two separate PR's: #1289  and #1290 , so this PR can be closed.

PR #1281, opened 2021-10-27
DP-SCREAM large scale forcing bug fix

bogensch:
If the selected DP-SCREAM case does not include the 3D large scale (LS) tendencies in the IOP forcing file then the effects of LS vertical advection need to computed.  In DP-SCREAM this is done using an Eulerian calculation.  However, the thermal expansion/compression term due to LS vertical advection was missing for temperature, thus this PR adds it.

This bug fix is B4B for non DP runs but is answer changing for select (but not all) DP-SCREAM runs.  It will be b4b for the ARM97 case, which is used for the SCREAM nightly tests.  Hence, the "B4B" flag.

### DIFF
--- a/components/eam/src/control/forecast.F90
+++ b/components/eam/src/control/forecast.F90
@@ -315,6 +315,11 @@ subroutine forecast(lat, psm1, psm2,ps, &
          (qminus(1,k,m) - qminus(1,k-1,m)))
      enddo
 
+     ! thermal expansion term due to LS vertical advection
+     do k=1,plev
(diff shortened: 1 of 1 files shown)


hassanbeydoun (approved):
Just read through this and looks good. Sorry for the delay!

AaronDonahue (approved):
Re-approving since it was approved by Hassan earlier and I just rebased to get it passing the AT.

AaronDonahue:
As expected, this branch failed the two known failures for the F-Compset suite on mappy.  All other tests pass.  Considering the conversation above with @bogensch that we don't expect this PR to impact the know ERP_DP failure on mappy I am going to manually merge.

PR #1284, opened 2021-10-27
Big refactor to taskset handling in test-all-scream

jgfouca:
The prior implementation assumed an unrestricted test-all-scream process
with access to the entire machine. lassen breaks this assumption because
compute node have logical cores 0-8 reserved. This required a new implementation
that did not make that assumption.

We now use the python tpl "psutil" to get a list of all logical cores
available to test-all-scream and we use that list to manage tasksetting.

With the availability of psutil, we no longer need to hardcode resource
counts in the MACHINE_METADATA dict. We can accurately probe for what's
available.

With these changes, we can safely limit test-all-scream to any (contiguous) subset
of the machine's cores and things should work.


components/scream/scripts/test_all_scream.py
```diff
@@ -187,11 +194,11 @@ def __init__(self, cxx_compiler=None, f90_compiler=None, c_compiler=None,
         if self._parallel:
             # We need to be aware that other builds may be running too.
             # (Do not oversubscribe the machine)
-            make_physical_cores_per_test = get_available_cpu_count(logical=False) // len(self._tests)
             log_per_phys = logical_cores_per_physical_core()
 
-            make_jobs_per_test  = make_physical_cores_per_test * log_per_phys
-            ctest_jobs_per_test = ctest_max_jobs // len(self._tests)
+            # Avoid splitting physical cores across test types
+            make_jobs_per_test  = ((make_max_jobs // log_per_phys) * log_per_phys) // len(self._tests)
```
bartgol: I'm not sure what's the point of dividing-multiply by log_per_phys. All it does is making sure that the result is divisible by log_per_phys, but that's not a real issue, right? And when we divide by `len(self._tests)` we can still end up with a physical core split across builds. Perhaps you wanted to do this:
```
ctest_jobs_per_test = ( (ctest_max_jobs // log_per_phys) // len(self._tests) ) * log_per_phys
```
so that you split physical cores across builds, then add back the hw firepower.

jgfouca: @bartgol , there was a bug here and it looks like you may have reviewed the outdated code.

bartgol (approved):

bartgol (approved):

bartgol (approved):

bartgol (approved):

PR #1288, opened 2021-10-29
Convert wet mixing ratios to dry mixing ratios for P3 (Fortran and C++)

singhbalwinder:
The wet mixing ratios from the scream's state vector (or prognostic state) are converted to dry mixing ratios for P3. P3 requires mixing ratios to be dry. The output from P3 are dry mixing ratios, which are then converted back to the wet mixing ratios.

[NBFB]


components/eam/src/physics/cam/physics_utils.F90
```diff
@@ -32,6 +33,53 @@ module physics_utils
 
 #endif
 
-    contains
+contains
+
+  pure function calculate_drymmr_from_wetmmr(ncols, wetmmr, qv_wet) result (drymmr)
+    !Compute drymmr for any wetmmr constituent using wet water vapor mixing ratio(qv_wet)
```
PeterCaldwell, line 39: since common usage of "mixing ratio" is what we call drymmr and wetmmr is really the specific humidity, maybe edit comment to "Compute drymmr (mass of water divided by mass of dry air; commonly known as mixing ratio) for any wetmmr constituent (mass of water divided by mass of dry air plus water vapor; see specific humidity) using qv_wet (mass of vapor divided by mass of dry air plus vapor).

components/eam/src/physics/cam/physics_utils.F90
```diff
@@ -32,6 +33,53 @@ module physics_utils
...
+    real(rtype), intent(in) :: qv_wet(:,:) !water vapor wet mass mixing ratio
+
+    !return variable
+    real(rtype) :: drymmr(pcols,pver) !dry mmr of a constituent
+
+    !Compute drymmr
+    drymmr(:ncols,:) = wetmmr(:ncols,:)/(1.0_rtype - qv_wet(:ncols,:))
+
+    !Since pcols can be > ncols, assign uninitialized columns using "huge"
+    drymmr(ncols+1:pcols,:) = huge(1.0_rtype)
```
PeterCaldwell, line 57: seems more typical and better to just initialize everything to huge, then overwrite the values we actually want to use. That way if you screw up your indexing, the code tells you something is wrong.

components/eam/src/physics/cam/physics_utils.F90
```diff
@@ -32,6 +33,53 @@ module physics_utils
...
+    real(rtype), intent(in) :: qv_dry(:,:)!water vapor dry mass mixing ratio
+
+    !return variable
+    real(rtype) :: wetmmr(pcols,pver) !wet mmr of a constituent
+
+    !Compute wetmmr
+    wetmmr(:ncols,:) = drymmr(:ncols,:)/(1.0_rtype + qv_dry(:ncols,:))
+
+    !Since pcols can be > ncols, assign uninitialized columns using "huge"
+    wetmmr(ncols+1:pcols,:) = huge(1.0_rtype)
```
PeterCaldwell, line 81: Same as above: initialize everywhere to huge, then overwrite where we actually will use the entries.

singhbalwinder:
Thanks Peter for the review. I have resolved your comments.

PeterCaldwell:
@singhbalwinder - I've now looked through the code and I think your changes look perfect. Thanks! We can't integrate this PR because it only has the F90 changes, so fails the AT. Did you mean to expand this PR to include the C++ once I confirm that it is right?

components/scream/src/physics/p3/atmosphere_microphysics.hpp
```diff
@@ -161,6 +189,23 @@ class P3Microphysics : public AtmosphereProcess
...
+         *DRY-TO-WET MMRs:
+         *-----------------
+         *Since the host model (or AD) needs wet mixing ratio tendencies(state vector has wet mixing ratios),
+         *we need to convert dry mixing ratios from P3 to wet mixing ratios
+         *NOTE: water vapor mixing ratio argument in calculate_wetmmr_from_drymmr function has to be dry water vapor mixing
+         *ratio
+         *----------------------------------------------------------------------------------------------------------------------
+         */
+
+        qc(icol,ipack) = PF::calculate_wetmmr_from_drymmr(qc(icol,ipack), qv(icol,ipack));
```
bartgol, line 202: In the preamble, you had both `qc` and `qc_dry` as member variables, while here you use the same variable. I don't have a preference, but it might be easier to use the same approach in pre and post process.

Note: that's regardless of whether qc is an alias to qc_dry in memory.

bartgol (commented):
Looks good. I had a couple of comments, but nothing very serious.

components/eam/src/physics/cam/micro_p3_interface.F90
```diff
@@ -890,12 +894,12 @@ subroutine micro_p3_tend(state, ptend, dtime, pbuf)
...
     call pbuf_get_field(pbuf,    prec_pcw_idx,  prec_pcw)
     call pbuf_get_field(pbuf,    snow_pcw_idx,  snow_pcw)
-    !============================ 
+    !============================
     ! All internal PBUF variables
     ! INPUTS
     call pbuf_get_field(pbuf,      relvar_idx,    relvar                                                   )
     call pbuf_get_field(pbuf,      t_prev_idx,    t_prev                                                   )
-    call pbuf_get_field(pbuf,     qv_prev_idx,    qv_prev                                                  )
+    call pbuf_get_field(pbuf,     qv_prev_idx,    qv_prev_wet                                                  )
```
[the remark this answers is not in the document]
singhbalwinder: Here I just changed the name of the PBUF variable to explicitly mention that it is a `wet` mmr. I am using qv_prev_dry internally in the P3 code.

components/eam/src/physics/cam/physics_utils.F90
```diff
@@ -32,6 +33,56 @@ module physics_utils
 
 #endif
 
-    contains
+contains
+
+  pure function calculate_drymmr_from_wetmmr(ncols, pver, wetmmr, qv_wet) result (drymmr)
+    !Computes drymmr (mass of a constituent divided by mass of dry air; commonly known as mixing ratio)
+    !for any wetmmr constituent (mass of a constituent divided by mass of dry air plus water
```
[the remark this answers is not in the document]
singhbalwinder: These terms (wet mmr and dry mmr) can be confusing. Peter came up with the above sentences to clarify these terms. I think it is better to keep it here for future reference.

singhbalwinder:
Hi @AaronDonahue , @bartgol and @PeterCaldwell : I think this PR is complete and ready for review after the tests pass.

bartgol (approved):

components/eam/src/physics/cam/micro_p3_interface.F90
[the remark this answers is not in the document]
PeterCaldwell: I think perhaps what Luca is asking is whether we don't just want to store qv_prev_dry in the field manager directly since (if I recall correctly, no other process touches it. The benefit would be saving the cost of a conversion to/from wet, but the cost would be violation of our "FM only stores wet mixing ratios" rule, which could be a dangerous precedent. My inclination is to vote for storing qv_prev_wet (as you do now).

bartgol:
@singhbalwinder can we merge this?

AaronDonahue (approved):

AaronDonahue:
@bartgol I added the CIME label just to make sure it behaves correctly for our CIME based tests.  It should fail them all with DIFF, but we should make sure there are not any build or run failures before merging.

PeterCaldwell (approved):
Thanks Balwinder!

AaronDonahue:
Okay, the AT reported much like we would expect for Blake and Mappy.  Passed on blake and just DIFFs on mappy.  Weaver failed though.  From the looks of it it failed in the build and not when running the test.  I think we will want to retest just to be certain there isn't something wrong on the GPU side.
@singhbalwinder , did you have a chance to test your branch using a GPU machine?

bartgol:
@AaronDonahue the previous AT run passed on Weaver. The last round it looks like it got stuck in the queue and then timed out. Given that it passed before though, I'm ok merging this.

bartgol:
@bartgol @jgfouca this will require a baseline bless tomorrow.

singhbalwinder:
@AaronDonahue : The standalone scream tests were passing on the AT machines, so I didn't run this branch on a GPU machine.

PR #1289, opened 2021-11-01
Make all tests that run through the AD run on multi-ranks.

AaronDonahue:
Additionally, make it so that each run produces output labeled
by the number of mpi-ranks used.

This commit introduces a number of new tests that determine whether
or not an atmospheric process will be BFB when run with 1 or more MPI
ranks.

It introduces a new rrtmgp stand alone test that runs a set number
of steps for radiation only.  Note, this commit changes the old
rrtmgp_standalone test to the rrtmgp_standalone_unit test.

Over the course of preparing this commit a bug was found in the rrtmgp
interface which is also fixed here.

This commit establishes a CprncTest macro which will compare two netCDF
output files.

For clarity, the word stand_alone was removed from the shoc_cld_p3_rrtmgp
and homme_shoc_cld_p3_rrtmgp tests, since they are not really stand alone
tests.

Finally there is a minor change to the cmake file for lassen, and a change
to the scorpio output interface that allows a user to not have the timestamp
attached to an output file name.


AaronDonahue:
This PR replaces most of #1269

components/scream/src/physics/rrtmgp/scream_rrtmgp_interface.cpp
```diff
@@ -390,7 +390,9 @@ namespace scream {
             auto p_lay_host = p_lay.createHostCopy();
             bool top_at_1 = p_lay_host(1, 1) < p_lay_host(1, nlay);
             auto t_lev_host = t_lev.createHostCopy();
```
bartgol, line 392: Is this var unused now then?

components/scream/tests/coupled/shoc_cld_p3_rrtmgp/shoc_cld_p3_rrtmgp_output.yaml
```diff
@@ -0,0 +1,62 @@
+%YAML 1.1
+---
+Casename: shoc_cld_p3_rrtmgp_output_np${MPI_RANKS}
```
bartgol, line 3: Side Q: Having to create multiple input files just for the sake of differentiate the output files depending on the number of ranks is getting annoying. What do you think about making the output class incorporate npN in the file name? We could control whether we want that or not with a simple input file parameter, just like `Timestamp in Filename`...

If the idea sounds good, I'm ok either merging this and doing it in another PR as well as doing it in this PR.

components/scream/tests/uncoupled/shoc/CMakeLists.txt
```diff
@@ -1,19 +1,48 @@
 include (ScreamUtils)
 
 # Create the test
+SET (TEST_LABELS "shoc;physics")
 set (NEED_LIBS shoc scream_control scream_share)
-CreateUnitTest(shoc_standalone "shoc_standalone.cpp" "${NEED_LIBS}" LABELS "shoc;physics")
+CreateUnitTest(shoc_standalone "shoc_standalone.cpp" "${NEED_LIBS}" LABELS ${TEST_LABELS}
+  MPI_RANKS 1 ${SCREAM_TEST_MAX_RANKS}
+  PROPERTIES FIXTURES_SETUP shoc_main
```
bartgol, line 8: Subjective: `shoc_main` (as well as `p3_main`. `spa_main`,... in other tests) is a somewhat overloaded string. I don't want other devs down the road to think this string is related to the funciton `shoc_main`. Perhaps we could use something like `generate_output_nc_files` or something quite obvious like that?

bartgol (approved):
I removed AUTOMERGE so you have time to consider my comments before the AT merges it. If you disagree with the comments, you can just merge manually upon AT success.

bartgol:
Ah, crap, I must have removed AUTOMERGE a second too late.

PR #1294, opened 2021-11-03
Fix IO classes' handling of var offsets.

### LINKED ISSUE #1293: Make IO classes use dofs GIDs as offsets for scorpio
This issue might be at the core of why tests involving homme (which, in practice, means tests involving a non-trivial mesh) are not BFB when running with 2+ ranks vs 1 rank.

Short explanation: when Homme chops the mesh, the global dofs [0,N) are not split linearly among ranks, and some rank might not have contiguous gids. However, IO classes computes the dofs offsets for PIO as `offsets=[S, S+1, ..., S+num_local_dofs-1]`, with `S = sum_{rank<my_rank} num_local_dofs(rank)`.

NOTE: using the dof GIDs as offsets in PIO _requires_ that for a problem with N global dofs, the global gids are the interval [0,N). I think we can safely assume that (and check it).

bartgol:
Fixes #1293

### DIFF
--- a/components/scream/src/share/io/scorpio_input.cpp
+++ b/components/scream/src/share/io/scorpio_input.cpp
@@ -415,7 +415,7 @@ get_var_dof_offsets(const FieldLayout& layout)
     auto dofs_h = Kokkos::create_mirror_view(dofs);
     Kokkos::deep_copy(dofs_h,dofs);
 
-    // Precompute tis *before* the loop, since it involves expensive collectives.
+    // Precompute this *before* the loop, since it involves expensive collectives.
     // Besides, the loop might have different length on different ranks, so
     // computing it inside might cause deadlocks.
     auto min_gid = m_grid->get_global_min_dof();
--- a/components/scream/src/share/io/scorpio_input.cpp
+++ b/components/scream/src/share/io/scorpio_input.cpp
@@ -124,7 +124,7 @@ set_grid (const std::shared_ptr<const AbstractGrid>& grid)
       "Error! I/O only supports grids which are 'unique', meaning that the\n"
       "       map dof_gid->proc_id is well defined.\n");
   EKAT_REQUIRE_MSG (
-      (grid->get_global_max_dof()-grid->get_global_min_dof()+1)==grid->get_num_global_dofs(),
+      (grid->get_global_max_dof_gid()-grid->get_global_min_dof_gid()+1)==grid->get_num_global_dofs(),
       "Error! In order for IO to work, the grid must (globally) have dof gids in interval [gid_0,gid_0+num_global_dofs).\n");
 
   EKAT_REQUIRE_MSG(m_comm.size()<=grid->get_num_global_dofs(),
@@ -418,7 +418,7 @@ get_var_dof_offsets(const FieldLayout& layout)
     // Precompute this *before* the loop, since it involves expensive collectives.
     // Besides, the loop might have different length on different ranks, so
     // computing it inside might cause deadlocks.
-    auto min_gid = m_grid->get_global_min_dof();
+    auto min_gid = m_grid->get_global_min_dof_gid();
     for (int icol=0; icol<num_cols; ++icol) {
       // Get chunk of var_dof to fill
       auto start = var_dof.begin()+icol*col_size;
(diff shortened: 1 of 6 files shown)


bartgol:
@AaronDonahue @tcclevenger Can you try to merge this branch into your local branches, and see if it fixes anything related to the 1rank-vs-Nranks issue(s)?

components/scream/src/share/io/scorpio_input.cpp
```diff
@@ -123,6 +123,9 @@ set_grid (const std::shared_ptr<const AbstractGrid>& grid)
   EKAT_REQUIRE_MSG (grid->is_unique(),
       "Error! I/O only supports grids which are 'unique', meaning that the\n"
       "       map dof_gid->proc_id is well defined.\n");
+  EKAT_REQUIRE_MSG (
```
AaronDonahue, line 126: I think I know the answer, but might as well ask.  How will this work for the future when we might want non-contiguous grids.  I.e. twin column.

components/scream/src/share/io/scorpio_input.cpp
```diff
@@ -397,24 +400,35 @@ get_var_dof_offsets(const FieldLayout& layout)
...
-
-    // Compute offsets of all my dofs
-    std::iota(var_dof.begin(), var_dof.end(), offset*col_size);
+    int col_size = layout.size() / num_cols;
+
+    auto dofs = m_grid->get_dofs_gids();
+    auto dofs_h = Kokkos::create_mirror_view(dofs);
+    Kokkos::deep_copy(dofs_h,dofs);
+
+    // Precompute tis *before* the loop, since it involves expensive collectives.
```
AaronDonahue, line 418: tis -> this

AaronDonahue, line 418: tis -> this

components/scream/src/share/io/scorpio_input.cpp
bartgol: The contiguity is at the _global_ level, not necessarily local. Even now, with 2+ ranks, the physics GLL grid does not have contiguous column gids.

Another way to look at this, is that running with 1 rank you must get gids `[g0,...,g0+num_dofs)`.

components/scream/src/share/io/scorpio_output.cpp
```diff
@@ -565,24 +569,38 @@ std::vector<int> AtmosphereOutput::get_var_dof_offsets(const FieldLayout& layout
...
-    m_comm.scan(&num_cols,&offset,1,MPI_SUM);
-    offset -= num_cols;
-
-    // Compute offsets of all my dofs
-    std::iota(var_dof.begin(), var_dof.end(), offset*col_size);
+    auto dofs = m_grid->get_dofs_gids();
+    auto dofs_h = Kokkos::create_mirror_view(dofs);
+    Kokkos::deep_copy(dofs_h,dofs);
+
+    // Precompute tis *before* the loop, since it involves expensive collectives.
```
AaronDonahue, line 590: copy/paste "tis"  -> "this"

components/scream/src/share/tests/grid_tests.cpp
```diff
@@ -29,6 +29,8 @@ TEST_CASE("point_grid", "") {
   REQUIRE(grid->get_num_local_dofs()  == num_local_cols);
   REQUIRE(grid->get_num_global_dofs() == num_global_cols);
   REQUIRE(grid->is_unique());
+  REQUIRE(grid->get_global_min_dof()==0);
```
AaronDonahue, line 32: is it possible to add a multi-rank test for say 2 ranks (when the machine makes it possible) that ensures that rank 1 has dofs [0,...,N/2-1] and rank 2 has dofs [N/2,...,N-1]?

components/scream/src/share/io/scorpio_input.cpp
AaronDonahue: ahhh okay, now I understand.  Great.

AaronDonahue (approved):
I had a couple comments, but nothing major.  Thank you @bartgol for getting this fix in so quickly!

components/scream/src/share/tests/grid_tests.cpp
bartgol: We can definitely add multi-rank tests for grids, it might help test a few more things. That said, we don't _need_ the point grid to have dofs partitioned that way, meaning that there should not be any place in the code that assumes that.

To be fair, we also don't assume that the min gid is 0, so maybe we should check that `max_gid-min_gid==num_global_gids`, which is the property we actually _need_ in IO.

PR #1297, opened 2021-11-03
Do not set CTEST_PARALLEL_LEVEL for scripts-ctest-driver

jgfouca:
It runs on login, not compute, nodes so the probed resources won't
always be accurate.


jgfouca:
Already tested that it works on weaver and mappy.

bartgol (approved):

PR #1285, opened 2021-10-27
Fix F2000SCREAMv1 test

tcclevenger:
Various updates required to get the F200SCREAMv1 nightly test to run.

PR #1299, opened 2021-11-04
Pull V1 tests into their own suite

jgfouca:
This is part of the effort to get V1 using gnu9 on mappy.


tcclevenger (approved):
Thanks @jgfouca!

This is still expected to fail, but PR https://github.com/E3SM-Project/scream/pull/1285 will address this.

PR #1300, opened 2021-11-04
CIME label testing needs to include V1 tests now

jgfouca:


tcclevenger (approved):

AaronDonahue:
@tcclevenger or @jgfouca , can either of you check that the F2000-SCREAM did actually run on mappy AT?  I checked the log and don't see it, but maybe the log is truncated.  We expect the mappy test to fail since the 2 DP tests have DIFF's, we just want to make sure we are capturing the F2000-SCREAM compset too.

bartgol (approved):

bartgol:
@AaronDonahue has a good point. The v1 compset did not run, since `jenkins_common_impl.sh` has the `-e` bash flag at the top, which cause the script to return immediately if any of the commands return a nonzero code.

@jgfouca `jenkins_common_impl.sh` seems to "store" the number of failures, but shouldn't the `-e` flag kill it immediately as soon as one of the tests fails? If so, we should modify this, since we don't want, say, cime tests to not run if we have a failure in test-all; all non-bfb PRs that change scream baselines would not get cime baselines tested otherwise. I wonder if this actually already happened...

jgfouca:
@bartgol , the `-e` was intentional to speed up AT return times. Do we not want that anymore?

bartgol:
I think `-e` is great if you can consider _every_ failure as bad. In that case, you don't care what happens next, since you need to go ahead and fix the failures, and test again. But `-e` can be a problem if there are _expected_ diffs, since you don't get to run the rest of the tests, to verify they still pass.

It's like if ctest quit immediately upon a run_and_cmp failure, without the chance to verify unit tests still pass.

bartgol (approved):

jgfouca:
Looks like both suites of tests ran.

tcclevenger:
Thanks @jgfouca! I'll merge my other branch into master and test.

PR #1304, opened 2021-11-08
Allow changing a subfield's slice idx at runtime

bartgol:
Currently, subfields are constructed by providing the dimension along which the field has to be subviewed, as well as the index along said dimension. These two indices are then frozen for the lifetime of the subfield.

This PR allows one to change the slice index of the subfield at runtime (a "dynamic" subfield). This is particularly useful for subviewing dynamics state fields, which have a "TimeLevel" dimension, which is simply used to store stages during the timestepping, and the "end-of-step" field could be stored in any of the time levels.

With dyn subfields, one can register both the state and a subfield of the state. The latter can be passed to several other routines (or atm procs), and all dynamics needs to do is move the index of the subview. Something like this:

```
field_type u_with_tl (...); // a 6-dimensional field
field_type u_no_tl = u_with_tl.subfield("u",1,0,true);
...
// Change tl where 'u_no_tl' points to
u_no_tl.get_header().get_alloc_prop().reset_subview_idx(new_tl);
```
Notice that this doesn't break anything in the downstream code _provided that they don't pre-extract the N-dim view_. That is, the downstream code cannot do
```
u_no_tl_view = u_no_tl.get_view<Real*****>();
```
until it's actually time to use the data. Pre-storing such view would cause it to always point to the same slice of `u_with_tl`. If instead they call `get_view<Real*****>()` every time they need to access the view, the correct up-to-date slice will be extracted from the parent field.

NOTE: this feature is to help outputing dyn fields with the native dyn grid layout (i.e., without remapping to phys grid first), in a subsequent PR.

### DIFF
--- a/components/scream/src/share/field/field.hpp
+++ b/components/scream/src/share/field/field.hpp
@@ -190,11 +190,13 @@ public:
   //     to store a stride for the slowest dimension.
   //   - If dynamic = true, it is possible to "reset" the slice index (k) at runtime.
   field_type subfield (const std::string& sf_name, const ekat::units::Units& sf_units,
-                       const int idim, const int k, const bool dynamic = false) const;
-  field_type subfield (const std::string& sf_name, const int idim, const int k, const bool dynamic = false) const;
+                       const int idim, const int index, const bool dynamic = false) const;
+  field_type subfield (const std::string& sf_name, const int idim,
+                       const int index, const bool dynamic = false) const;
   field_type subfield (const int idim, const int k, const bool dynamic = false) const;
 
   // If this field is a vector field, get a subfield for the ith component.
+  // If dynamic = true, it is possible to "reset" the component index at runtime.
   // Note: throws if this is not a vector field.
   field_type get_component (const int i, const bool dynamic = false);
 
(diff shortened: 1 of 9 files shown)


components/scream/src/share/field/field.hpp
```diff
@@ -188,14 +188,15 @@ class Field {
...
-  field_type subfield (const std::string& sf_name, const int idim, const int k) const;
-  field_type subfield (const int idim, const int k) const;
+                       const int idim, const int k, const bool dynamic = false) const;
+  field_type subfield (const std::string& sf_name, const int idim, const int k, const bool dynamic = false) const;
+  field_type subfield (const int idim, const int k, const bool dynamic = false) const;
 
   // If this field is a vector field, get a subfield for the ith component.
   // Note: throws if this is not a vector field.
-  field_type get_component (const int i);
+  field_type get_component (const int i, const bool dynamic = false);
```
jeff-cohere, line 199: Worth repeating the comment above to explain this new argument?

components/scream/src/share/field/field_request.hpp
```diff
@@ -233,10 +234,23 @@ struct FieldRequest {
    : FieldRequest(FID(name,layout,u,grid),std::list<std::string>{group},ps)
   { /* Nothing to do here */ }
 
+  FieldRequest (const FID& fid, const FieldRequest& parent, int idim, int k, bool dynamic)
```
jeff-cohere, line 237: I wonder if we shouldn't rename `k` to `index` here, to emphasize that the slice index is not necessarily the vertical pack index.

jeff-cohere (commented):
This is cool. I've left a couple of remarks for you to consider, essentially only related to comments. I'll leave this as a comment so @AaronDonahue can weigh in.

jeff-cohere (approved):

PR #1306, opened 2021-11-09
Remove updated groups from atm proc interface

bartgol:
The asymmetry between fields and groups has puzzled more than one person so far. This PR makes the treatment of field groups more similar to that of fields.

Additionally, fix the treatment of required fields/groups in the AtmosphereProcessGroup class.

### DIFF
--- a/components/scream/src/control/atmosphere_driver.cpp
+++ b/components/scream/src/control/atmosphere_driver.cpp
@@ -26,29 +26,34 @@ namespace control {
  *  2) Create the grid manager, and query the atm procs for the grids they need. The GM will then
  *     proceed to build those grids (and only those grids).
  *  3) The GM is passed back to the atm procs, which can grab the needed grids, from which they can
- *     get the information needed to complete the setup of the FieldIdentifiers of their fields
- *     (both required and computed). Their field identifiers MUST be completed upon return from
+ *     get the information needed to complete the setup of the FieldRequest and GroupRequest for
+ *     the required/computed fields/groups. Their requests MUST be completed upon return from
  *     the 'set_grids' method.
  *     Note: at this stage, atm procs that act on non-ref grid(s) should be able to create their
  *           remappers. The AD will *not* take care of remapping inputs/outputs of the process.
- *  4) Register all fields from all atm procs inside the field manager
+ *  4) Register all fields and all groups from all atm procs inside the field managers, and proceed
+ *     to allocate fields. Each field manager (there is one FM per grid) will take care of
+ *     accommodate all requests for packing as well as (if possible) bundling of groups.
+ *     For more details, see the documentation in the share/field/field_request.hpp header.
  *  5) Set all the fields into the atm procs. Before this point, all the atm procs had were the
- *     FieldIdentifiers for their input/output fields. Now, we pass actual Field objects to them,
- *     where both the data (Kokkos::View) and metadata (FieldHeader) inside will be shared across
- *     all processes using the field. This allow data and metadata to be always in sync.
- *     Note: output fields are passed to an atm proc as read-write (i.e., non-const data type),
- *           while input fields are passed as read-only (i.e., const data type). Yes, the atm proc
- *           could cheat, and cast away the const, but we can't prevent that. However, in debug builds,
- *           we store 2 copies of each field, and use the extra copy to check, at run time, that
- *           no process alters the values of any of its input fields.
+ *     FieldIdentifiers for their input/output fields and FieldGroupInfo for their input/output
+ *     field groups. Now, we pass actual Field and FieldGroup objects to them, where both the
+ *     data (Kokkos::View) and metadata (FieldHeader) inside will be shared across all copies
+ *     of the field. This allow data and metadata to be always in sync.
+ *     Note: output fields/groups are passed to an atm proc as read-write (i.e., non-const data type),
+ *           while input ones are passed as read-only (i.e., const data type). Yes, the atm proc
+ *           could cheat, and cast away the const, but we can't prevent that.
  *  6) All the atm inputs (that the AD can deduce by asking the atm proc group for the required fiedls)
- *     are initialized, by reading values from an initial conditions netcdf file.
+ *     are initialized. For restart runs, all fields are read from a netcdf file (to allow BFB
+ *     restarts), while for initial runs we offer a few more options (e.g., init a field to
+ *     a constant, or as a copy of another field). During this process, we also set the initial
+ *     time stamp on all the atm input fields.
  *     If an atm input is not found in the IC file, we'll error out, saving a DAG of the
  *     atm processes, which the user can inspect (to see what's missing in the IC file).
  *  7) All the atm process are initialized. During this call, atm process are able to set up
- *     all the internal structures that they were not able to init previously. They can also
- *     utilize their input fields to perform initialization of some internal data structure.
- *  8) Finally, set the initial time stamp on all fields, and perform some debug structure setup.
+ *     all the internal structures that they were not able to init previously. For instance,
+ *     they can set up remappers from the reference grid to the grid they operate on. They can
+ *     also utilize their input fields to perform initialization of some internal data structure.
  *
  */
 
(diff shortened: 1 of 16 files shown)


components/scream/src/control/atmosphere_driver.cpp
```diff
@@ -26,29 +26,34 @@ namespace control {
...
- *     get the information needed to complete the setup of the FieldIdentifiers of their fields
- *     (both required and computed). Their field identifiers MUST be completed upon return from
+ *     get the information needed to complete the setup of the FieldRequest and GroupRequest for
+ *     the required/computed fields/groups. Their requests MUST be completed upon return from
  *     the 'set_grids' method.
  *     Note: at this stage, atm procs that act on non-ref grid(s) should be able to create their
  *           remappers. The AD will *not* take care of remapping inputs/outputs of the process.
- *  4) Register all fields from all atm procs inside the field manager
+ *  4) Register all fields and all groups from all atm procs inside the field managers, and proceed
+ *     to allocate fields. Each field manager (there is one FM per grid) will take care of
```
AaronDonahue, line 35: nitpicky, but I think we either want "take care of" or "accomodate"

components/scream/src/control/atmosphere_driver.cpp
```diff
@@ -26,29 +26,34 @@ namespace control {
  *  2) Create the grid manager, and query the atm procs for the grids they need. The GM will then
  *     proceed to build those grids (and only those grids).
  *  3) The GM is passed back to the atm procs, which can grab the needed grids, from which they can
- *     get the information needed to complete the setup of the FieldIdentifiers of their fields
- *     (both required and computed). Their field identifiers MUST be completed upon return from
+ *     get the information needed to complete the setup of the FieldRequest and GroupRequest for
```
AaronDonahue, line 29: Do we explain in these comments what the distinction is between a field and a group?  If not, perhaps one line would make it easier to understand for any newcomers.

components/scream/src/control/tests/ad_tests.cpp
```diff
@@ -31,15 +32,14 @@ TEST_CASE ("group_requirements","[!throws]")
...
-  // Verify that after initialization field "E" matches field "A"
+
+  // Verify that the atm proc group has 1 required field and 0 required groups.
+  const auto& apg = ad.get_atm_processes();
+  REQUIRE (apg->get_num_processes()==3);
+  REQUIRE (apg->get_fields_in().size()==1);
+  REQUIRE (apg->get_groups_in().size()==0);
+
   auto field_mgr = ad.get_ref_grid_field_mgr();
-  const auto& view_A = field_mgr->get_field("A").get_view<const Real**, Host>();
```
AaronDonahue, line 36: I'm sure there is a good reason, but why do we remove this test?

components/scream/src/control/tests/ad_tests.yaml
```diff
@@ -9,8 +9,6 @@ Initial Conditions:
     A: 1.0
     B: 2.0
     C: 3.0
-    D: [4.0, 5.0]
```
AaronDonahue, line 12: By removing this are we no longer testing if it is possible to pass a vector quantity to initialize a vector field?

components/scream/src/share/atm_process/atmosphere_process.hpp
```diff
@@ -115,10 +111,14 @@ class AtmosphereProcess : public ekat::enable_shared_from_this<AtmosphereProcess
   // corresponding set_xyz_impl method(s).
   // Note: this method will be called *after* set_grids, but *before* initialize.
   //       You are *guaranteed* that the views in the field/group are allocated by now.
-  void set_required_field (const Field<const Real>& f);
-  void set_computed_field (const Field<Real>& f);
-  void set_required_group (const FieldGroup<const Real>& group);
-  void set_updated_group (const FieldGroup<Real>& group);
+  // Note: you are *unlikely* to need to override these method. In 99.99% of the cases,
```
AaronDonahue, line 114: "methods"

components/scream/src/share/atm_process/atmosphere_process.hpp
```diff
@@ -167,7 +167,7 @@ class AtmosphereProcess : public ekat::enable_shared_from_this<AtmosphereProcess
   enum RequestType {
     Required,
     Computed,
-    Updated
+    Updated   // For convenience, triggers Required+Computed
```
AaronDonahue, line 170: I like this comment

components/scream/src/share/atm_process/atmosphere_process_group.cpp
```diff
@@ -144,25 +144,162 @@ void AtmosphereProcessGroup::finalize_impl (/* what inputs? */) {
...
+  }
+
+  EKAT_REQUIRE_MSG (first_proc_that_needs_f>=0,
+    "Error! This atmosphere process does not require the input field.\n"
+    "    field id: " + f.get_header().get_identifier().get_id_string() + "\n"
+    "   atm process: " + this->name() + "\n"
+    "Something is wrong up the call stack. Please, contact developers.\n");
+
+  // Loop over all the fields in the group, and see if they are all computed
+  // by a previous process. If so, then this group is not a "required" group
```
AaronDonahue, line 165: Is there a better way to say this? I think the copious use of the word "group" is a bit confusing.

components/scream/src/share/atm_process/atmosphere_process_group.cpp
```diff
@@ -144,25 +144,162 @@ void AtmosphereProcessGroup::finalize_impl (/* what inputs? */) {
 }
 
 void AtmosphereProcessGroup::
-set_required_group_impl (const FieldGroup<const Real>& group)
-{
+set_required_field (const Field<const Real>& f) {
+  // Find the first process that requires this group
```
AaronDonahue, line 148: This assumes sequential splitting right?  Will this be an issue when we develop parallel splitting in the model?

AaronDonahue (commented):
I left a few comments, they shouldn't hold up the merge but because the Automerge is triggered I'm reviewing as a "comment" in case we want to respond before it is merged.

components/scream/src/control/tests/ad_tests.cpp
bartgol: I just wanted to keep the test small, I guess, so I removed a couple of fields. But I can see there's value in checking initialization procedures. I'll bring some stuff back.

components/scream/src/control/atmosphere_driver.cpp
bartgol: Added some pointers

components/scream/src/control/tests/ad_tests.yaml
bartgol: Added back testing for that feature.

components/scream/src/share/atm_process/atmosphere_process_group.cpp
bartgol: Added a check.

components/scream/src/share/atm_process/atmosphere_process_group.cpp
```diff
@@ -144,25 +144,174 @@ void AtmosphereProcessGroup::finalize_impl (/* what inputs? */) {
...
 void AtmosphereProcessGroup::
-set_required_group_impl (const FieldGroup<const Real>& group)
-{
+set_required_field (const Field<const Real>& f) {
+  if (m_group_schedule_type==ScheduleType::Parallel) {
+    // In parallel splitting, all required fields are *actual* inputs,
+    // and the base class impl is fine.
+    AtmosphereProcess::set_required_field(f);
+  }
+
```
AaronDonahue, line 153: Thank you for adding the check, could we add another check that the ScheduleType is Sequential and throw an error if not?  I'm thinking that in the future we may implement other ScheduleTypes we've never heard of and putting in as many hooks as possible that make sure we don't accidentally break the code without a map to find out where would be great.

bartgol: Do you really think there will be a schedule type other than Sequential or Parallel?

AaronDonahue: none specifically come to mind.  I guess from one perspective, anything we come up with will just be a combination of those two elemental coupling approaches.  I suppose if we decide to do something exotic it will be important that whatever that is be not a new ScheduleType per-se but some set of operations that are Sequential or Parallel.

bartgol: Well, the way I see it, a "mixed" approach will be likely consist of sub-groups, so each individual sub-group will still be sequential or parallel.

AaronDonahue (approved):

PR #1307, opened 2021-11-09
Beydoun/precip ice fraction 2

hassanbeydoun:
PR adds diagnostic output for liquid cloud fraction, ice cloud fraction, and precipitating ice fraction. These new hydrometeor fractions are purely diagnostic and have no impact on internal code calculations.

### DIFF
--- a/components/eam/src/physics/cam/shoc_intr.F90
+++ b/components/eam/src/physics/cam/shoc_intr.F90
@@ -409,8 +409,9 @@ end function shoc_implements_cnst
     call addfld('BRUNT',(/'lev'/), 'A', 's-1', 'Brunt frequency')
     call addfld('RELVAR',(/'lev'/), 'A', 'kg/kg', 'SHOC cloud liquid relative variance')
     call addfld('ICE_CLOUD_FRAC',(/'lev'/), 'A', '', 'Ice number aware cloud fraction')
-    call addfld('PRECIP_ICE_FRAC',(/'lev'/), 'A', '', 'Precipitating ice fraction')
+    call addfld('PRECIPITATING_ICE_FRAC',(/'lev'/), 'A', '', 'Precipitating ice fraction')
     call addfld('LIQ_CLOUD_FRAC',(/'lev'/), 'A', '', 'Liquid cloud fraction')
+    call addfld('TOT_CLOUD_FRAC',(/'lev'/), 'A', '', 'total cloud fraction')
 
     call add_default('SHOC_TKE', 1, ' ')
     call add_default('WTHV_SEC', 1, ' ')
@@ -434,8 +435,9 @@ end function shoc_implements_cnst
     call add_default('BRUNT',1,' ')
     call add_default('RELVAR',1,' ')
     call add_default('ICE_CLOUD_FRAC',1,' ')
-    call add_default('PRECIP_ICE_FRAC',1,' ')
+    call add_default('PRECIPITATING_ICE_FRAC',1,' ')
     call add_default('LIQ_CLOUD_FRAC',1,' ')
+    call add_default('TOT_CLOUD_FRAC',1,' ')
 
     ! ---------------------------------------------------------------!
     ! Initialize SHOC                                                !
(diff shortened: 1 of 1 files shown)


components/eam/src/physics/cam/shoc_intr.F90
```diff
@@ -1069,11 +1079,28 @@ subroutine shoc_tend_e3sm( &
     enddo
    
     !  Probably need to add deepcu cloud fraction to the cloud fraction array, else would just 
-    !  be outputting the shallow convective cloud fraction 
-
+    !  be outputting the shallow convective cloud fraction
+ 
+    !  Add liq, ice, and precipitating ice fractions here. These are purely
+    !  diagnostic outputs and do not impact the rest of the code. The qi threshold for
+    !  setting ice_cloud_fraction to 1 currently set 1e-8 kg/kg but can be
```
PeterCaldwell, line 1086: Putting specific values of tuning parameters in comments often leads to incorrect comments later. I'd suggest just saying "is a tunable parameter".

PeterCaldwell: Also, does the 1e-8 we use here need to be synced to the threshold used for the actual ice cloud fraction calculation used by P3? If so, we should probably make those values a single parameter set somewhere like micro_p3_utils.F90...

PeterCaldwell: And maybe also mention that the d(qi)/d(ni) slope is also a tuning parameter?

components/eam/src/physics/cam/shoc_intr.F90
```diff
@@ -1069,11 +1079,28 @@ subroutine shoc_tend_e3sm( &
...
+    !  changed based on what we think is sensible. 
+
+    liq_cloud_frac = 0.0_r8
+    ice_cloud_frac = 0.0_r8
+    precip_ice_frac = 0.0_r8
+ 
     do k=1,pver
       do i=1,ncol
         cloud_frac(i,k) = min(ast(i,k)+deepcu(i,k),1.0_r8)
```
PeterCaldwell, line 1095: we should probably add a total cloud fraction output which is min(1, max(ice_cloud_frac,liq_cld_frac)+deepcu).

components/eam/src/physics/cam/shoc_intr.F90
```diff
@@ -408,6 +408,9 @@ subroutine shoc_init_e3sm(pbuf2d, dp1_in)
     call addfld('CONCLD',(/'lev'/),  'A',        'fraction', 'Convective cloud cover')
     call addfld('BRUNT',(/'lev'/), 'A', 's-1', 'Brunt frequency')
     call addfld('RELVAR',(/'lev'/), 'A', 'kg/kg', 'SHOC cloud liquid relative variance')
+    call addfld('ICE_CLOUD_FRAC',(/'lev'/), 'A', '', 'Ice number aware cloud fraction')
+    call addfld('PRECIP_ICE_FRAC',(/'lev'/), 'A', '', 'Precipitating ice fraction')
```
PeterCaldwell, line 412: I feel like precip_ice_frac will get misinterpreted a lot - is it a kind of precipitating ice mass? I'd like to rename it, but I struggle a bit with what a better name would be. Perhaps "PRECIPITATING_ICE_FRAC"?

PeterCaldwell (commented):
Looks good. My only comment is that we should probably make the threshold for having ice cloud fraction 0 or 1 a parameter so it can be consistently used by both the outfld cloud fraction calculation and the ice cloud fraction calculation used by P3... It would be awkward to have those 2 numbers be inconsistent.

I also think this PR can be integrated without any changes to C++ code (though we should probably make those changes as well at some point) because you don't change the interfaces to any functions...

components/eam/src/physics/cam/shoc_intr.F90
bogensch: I second this!

PeterCaldwell (approved):
Looks good, thanks!

PR #1308, opened 2021-11-09
Move responsibility of handling output files to OutputManager

bartgol:
Moving file management to the OutputManager, we allow writing more than one output stream on the same file (e.g., for both dyn and phys grids output). Before, two separate instances of AtmosphereOutput were both acting as if they owned the file (hence, updating time, closing files,...), which is a problem when 2+ instances write on the same file. Additionally, rpointer files (particularly during unit tests) could easily get corrupted.

Other miscellanea changes:
- bug fixes in the `share/io` folder (one related to padding, one related to the F90 interface)
- added ability to subtract time stamps, to get elapsed seconds.
- storing/loading initial date/tim in nc files. Allows to restart time dimension from where it stopped (rather than start from 0 again).


components/scream/src/share/util/scream_time_stamp.cpp
```diff
@@ -200,6 +205,72 @@ TimeStamp operator+ (const TimeStamp& ts, const int dt) {
...
+      diff += spd*days_in_month(y1,m);
+    }
+    diff += spd*(d1-1);
+    diff += ts1.sec_of_day();
+  } else if (m1>m2) {
+    // Rest of the day and month in ts2's current month
+    diff += spd - ts2.sec_of_day();
+    diff += spd*(days_in_month(y2,m2)-d2);
+
+    // Whole montsh in (m2,m1)
```
AaronDonahue, line 251: "months"

components/scream/src/share/io/scorpio_input.cpp
```diff
@@ -25,7 +25,11 @@ AtmosphereInput (const ekat::Comm& comm,
                  const std::shared_ptr<const gm_type>& grids_mgr)
  : AtmosphereInput(comm,params)
 {
-  init (field_mgr,grids_mgr);
+  // Sets the intermal field mgr, and possibly sets up the remapper
```
AaronDonahue, line 28: "internal"

components/scream/src/share/io/scorpio_output.hpp
```diff
@@ -27,7 +27,12 @@
  *  Casename:                     STRING
  *  Averaging Type:               STRING
  *  Max Snapshots Per File:       INT                   (default: 1)
- *  Fields:                       ARRAY OF STRINGS
+ *  Grids:                        ARRAY OF STRINGS
```
AaronDonahue, line 30: I'm a bit confused here, from the code above I understood that one can also just list "Fields" at the level of "Grids" like is currently done.  Maybe I misunderstood the checks though.

components/scream/tests/coupled/homme_physics/homme_shoc_cld_p3_rrtmgp.cpp
```diff
@@ -63,11 +63,16 @@ TEST_CASE("scream_homme_physics", "scream_homme_physics") {
   //       during RRTMGPRatiation::finalize_impl, after RRTMGP has deallocated
   //       all its arrays.
   ad.initialize(atm_comm,ad_params,t0);
-  printf("Start time stepping loop...       [  0%%]\n");
+
+  if (atm_comm.am_i_root()) {
```
AaronDonahue, line 67: Thank you for this change.

AaronDonahue (commented):
Just a couple of questions, but other than that it looks great!  There was a lot of work that went into this and I think it cleans the code up considerably.  Thanks!

components/scream/src/share/io/scorpio_output.hpp
bartgol: I will go over this documentation, and make sure it's consistent.

bartgol: Ah, yeah, it is possible, but the AD (or, better, the OutputManager) was not really supporting it, making it effectively pointless. I'll add this possibility back.

AaronDonahue (approved):

AaronDonahue (approved):

PR #1309, opened 2021-11-10
Update F2000v1 compset to use 24 procs

tcclevenger:
Previously, test only used 1 proc. Now, test with 24 procs.

### DIFF
--- a/cime_config/tests.py
+++ b/cime_config/tests.py
@@ -419,7 +419,7 @@ _TESTS = {
     "e3sm_scream_v1" : {
         "time"  : "03:00:00",
         "tests" : (
-            "SMS_D_Ln2_P1x1.ne4_ne4.F2000SCREAMv1.scream-30min_ts",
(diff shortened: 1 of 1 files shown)


AaronDonahue (approved):

AaronDonahue:
The fails on mappy are expected.  although I still don't see the `gnu9` tests in the mappy read outs.  I assume that the gnu9 test suite did run (i.e. F2000SCREAMv1) and didn't report a fail.  If so we can manually merge this PR.  @jgfouca , @bartgol or @tcclevenger can one of you confirm that mappy will pass the F2000SCREAMv1 ran through the AT and passed?

tcclevenger:
@AaronDonahue Looked at the log, failed tests are the expected ones, the F2000Screamv1 gives DIFF test status like before. Good to merge?

AaronDonahue:
Excellent, then I will manually merge.

tcclevenger:
Thanks @AaronDonahue!

Issue #1301, opened 2021-11-05
For initial runs, allow atm procs to "bootstrap" some initial condition (IC) fields


bartgol:
In EAM, Homme bootstraps some fields, like dp3d and phi, from ps (assuming hydrostatic state). Currently, SCREAM _requires_ all IC fields to be loaded from file (either set to constant via yaml file, or loaded from netcdf file). It would be nice if SCREAM could use files like `cam.i` for ICs. These files contain `ps, U, V, T, q`. In order to use these files, we need to allow some runtime features:

- Allow scorpio to read vars by renaming them. We can allow specification of fields names in the yaml IO file as `Fields: [ ps, T_mid=T, ...]`, which our IO class can interpret as "the FM field T_mid is to be read/written from/to nc file as T".
- Allow atm procs to claim "bootstrapping abilities for certain fields. E.g., Homme could claim to be able to bootstrap "pseudo_density" and "phi_int". The flow would be roughly the following:
  1. Each atm proc could register itself as a "FieldInitializer" when the field is set in the process (via `set_[required|computed]_field`).
  2. During the call to `initialize_fields()`, the AD would _not_ add those fields to the list of fields that scorpio needs to load from nc file.
  3. Upon loading all other fields, the AD `atm_proc->initialize_fields()` for any atm proc that claimed to be an initializer of some field. The atm proc can use any other field at this point, since they have been loaded from file.
  4. Option: in case atm proc 1 (AP1) wants to bootsrap field F1, but the bootstrap requires field F2, which is bootstrapped by atm proc 2, we could make the request for being an initializer a bit more complex. Namely, when AP1 registers itself as initializer for F1, it can list all other fields needed for the bootstrapping, like `F1.add_me_as_initializer("F2")`. Then, the AD can make sure that a) there are no circular dependencies, and b) the bootstrapping happens in the correct order.

Issue #1056, opened 2021-05-19
Read Topography input data into the grids manager


AaronDonahue:
The surface topography needs to be read in from netCDF and stored somewhere for use by dynamics and physics.  The natural candidate would be to store this information in the grids manager.

The motivation for storing this information in the grids manager is:
1. It is assumed to be needed for all simulations.  So not process specific.
2. It is static, i.e. it doesn't change with time.

AaronDonahue:
@bartgol , do we do this now?  @tcclevenger do you know?  Can we close this issue?

bartgol:
Yes.

Issue #958, opened 2021-03-31
Get surface geopotential from the grid manager rather than passing it through the field manager


PeterCaldwell:
Currently some processes ask the field manager for PHIS. But PHIS is a time-invariant property of the model grid, so makes better logical sense to be supplied by the grid manager. Tasks:

- [ ] Make a function associated with the grid which returns surface geopotential
- [ ] Modify processes asking for PHIS to instead use this feature. It seems like PHIS should be defined during initialization time then held in memory forever afterwards? Though maybe it would be more memory efficient to just recalculate it every timestep?
- [ ] quit calling it PHIS, which only makes sense to E3SM afficionados. Call it surf_geopot or something instead...

bartgol:
How would one recompute phis? `phi=-gz`, so phis is basically the topography, right? If you want to recompute it, you need to store the topography in the grid. One of the two must be available. I'm ok with storing surface elevation, instead of surface geopotential (it might be even clearer). But one of the two needs to be in the grid.

PeterCaldwell:
Yeah, I like storing surface height rather than surface geopotential. And  like you say, I don't envision "recomputing" geopotential because it is really just something that needs to be read in from a file. I was just imagining that the grid would include surface elevation data and would have a method to just return that data when asked. Though I guess if grid is an object, we could just as easily ask  for grid.surface_height as asking for surface_height = grid.get_surface_height() - which is what I was originally thinking of above.

bartgol:
Well, that assumes that we have a `surface_height` member of the base abstract class. We already store "geometry data" views, in a `std::map<std::string, view1d_type>`. So we don't even have to change the base class interface. We simply need to make sure our grids store such a view.

AaronDonahue:
@bartgol or @tcclevenger , based on the conversation in #1313 is this still even a relevant issue?  Can we close it?

bartgol:
I think we can close, yes. The topography can def stay in the FM.

PR #1313, opened 2021-11-16
Set topography views inside Homme.

tcclevenger:
Set topography views inside Homme. This requires:
1. Linking FM phis to homme elem%state%phis to ElementGeometry::m_phis
2. Calling `compute_gradphis_c()` (which requires adding link in homme/src/theta-l_kokkos/theta_f2c_mod.F90)

Edit: No longer adding anything to Homme. Instead, using `use derivative_mod_base, only: gradient_sphere` to compute `gradphis`

### DIFF
--- a/components/homme/src/theta-l_kokkos/theta_f2c_mod.F90
+++ b/components/homme/src/theta-l_kokkos/theta_f2c_mod.F90
@@ -87,10 +87,6 @@ interface
     type (c_ptr) , intent(in) :: phis_ptr, gradphis_ptr
   end subroutine init_geopotential_c
 
-  ! Uses phis to compute gradphis
-  subroutine compute_gradphis_c () bind(c)
-  end subroutine
-
   ! Initializes C++ diagnostics arrays with ptrs provided from f90
   subroutine init_diagnostics_c (elem_state_q_ptr, elem_accum_qvar_ptr, elem_accum_qmass_ptr, &
                                  elem_accum_q1mass_ptr, elem_accum_iener_ptr,                 &
--- a/components/homme/src/theta-l_kokkos/theta_f2c_mod.F90
+++ b/components/homme/src/theta-l_kokkos/theta_f2c_mod.F90
@@ -87,6 +87,10 @@ interface
     type (c_ptr) , intent(in) :: phis_ptr, gradphis_ptr
   end subroutine init_geopotential_c
 
+  ! Uses phis to compute gradphis
+  subroutine compute_gradphis_c () bind(c)
+  end subroutine
+
   ! Initializes C++ diagnostics arrays with ptrs provided from f90
   subroutine init_diagnostics_c (elem_state_q_ptr, elem_accum_qvar_ptr, elem_accum_qmass_ptr, &
                                  elem_accum_q1mass_ptr, elem_accum_iener_ptr,                 &
--- a/components/scream/src/dynamics/homme/atmosphere_dynamics.hpp
+++ b/components/scream/src/dynamics/homme/atmosphere_dynamics.hpp
@@ -45,6 +45,9 @@ public:
   // Note: field_mgrs[grid_name] is the FM on grid $grid_name
   void register_fields (const std::map<std::string,std::shared_ptr<FieldManager<Real>>>& field_mgrs) const;
 
+  // Retrieves an internal field, given field name and grid name.
+  const Field<Real>& get_internal_field (const std::string& name, const std::string& grid) const;
+
 #ifndef KOKKOS_ENABLE_CUDA
   // Cuda requires methods enclosing __device__ lambda's to be public
 protected:
@@ -66,6 +69,7 @@ protected:
   void update_pressure ();
 
   void initialize_impl ();
+
 protected:
   void run_impl        (const int dt);
   void finalize_impl   ();
@@ -88,10 +92,6 @@ protected:
                               const std::vector<FieldTag>& tags,
                               const std::vector<int>& dims,
                               const std::string& grid);
-
-  // Retrieves an internal field, given field name and grid name.
-  Field<Real>& get_internal_field (const std::string& name, const std::string& grid);
-
 
   // Some helper fields. WARNING: only one copy for each internal field!
   std::map<std::string,field_type>  m_internal_fields;
--- a/components/scream/src/dynamics/homme/interface/homme_driver_mod.F90
+++ b/components/scream/src/dynamics/homme/interface/homme_driver_mod.F90
@@ -77,9 +77,10 @@ contains
 
   subroutine prim_copy_cxx_to_f90 (copy_phis)
     use iso_c_binding,     only: c_ptr, c_loc
-    use homme_context_mod, only: tl
-    use dimensions_mod,    only: nlevp
-    use theta_f2c_mod,     only: cxx_push_results_to_f90
+    use homme_context_mod, only: tl, elem
+    use dimensions_mod,    only: nlevp, nelemd, np
+    use kinds,             only: real_kind
+    use theta_f2c_mod,     only: cxx_push_results_to_f90, init_geopotential_c, compute_gradphis_c
     use element_state,     only: elem_state_v, elem_state_w_i, elem_state_vtheta_dp,     &
                               elem_state_phinh_i, elem_state_dp3d, elem_state_ps_v,   &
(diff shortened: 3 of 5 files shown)


AaronDonahue:
I think this partly addresses Issue #958

tcclevenger:
> I think this partly addresses Issue #958 and #1056

Yeah, it ends up being a different approach (not using the grid). Since https://github.com/E3SM-Project/scream/issues/1301 is about having have homme bootstrap `phi` from T,qv,dp,ps.

components/homme/src/theta-l_kokkos/theta_f2c_mod.F90
```diff
@@ -87,6 +87,10 @@ subroutine init_geopotential_c (ie, phis_ptr, gradphis_ptr) bind(c)
     type (c_ptr) , intent(in) :: phis_ptr, gradphis_ptr
   end subroutine init_geopotential_c
 
+  ! Uses phis to compute gradphis
+  subroutine compute_gradphis_c () bind(c)
```
bartgol, line 91: We should try to avoid modifying stuff in the homme folder. If we need to do so, we need to go through the process of "PR in E3SM" -> "merge in E3SM master" -> "upstream merge", which can be long.

bartgol: In this case, I agree that it's appropriate for this to be in the theta_f2c mod. But fortunately, this module is not _really_ needed; it's just for convenience, so one can declare all interfaces in one place, and use them via `use` directive. For this function, we can add the interface where we use it in scream...

components/scream/src/dynamics/homme/interface/homme_driver_mod.F90
```diff
@@ -77,9 +77,10 @@ end subroutine prim_init_data_structures_f90
 
   subroutine prim_copy_cxx_to_f90 (copy_phis)
     use iso_c_binding,     only: c_ptr, c_loc
-    use homme_context_mod, only: tl
-    use dimensions_mod,    only: nlevp
-    use theta_f2c_mod,     only: cxx_push_results_to_f90
+    use homme_context_mod, only: tl, elem
+    use dimensions_mod,    only: nlevp, nelemd, np
+    use kinds,             only: real_kind
+    use theta_f2c_mod,     only: cxx_push_results_to_f90, init_geopotential_c, compute_gradphis_c
```
bartgol, line 83: Namely, in this subroutine, you could add the `interface` block inside this subroutine (or at the top of the module, as you wish). That way, we don't have to modify the homme folder...

bartgol (changes requested):
That looked simple enough! Unfortunately, we need to keep out hands off of the homme component folder, whenever possible.

AaronDonahue:
So with this PR will we have topography in our runs?  Or do we need a follow-up PR for that?

components/scream/src/dynamics/homme/interface/homme_driver_mod.F90
```diff
@@ -113,6 +119,19 @@ subroutine prim_copy_cxx_to_f90 (copy_phis)
...
       ! Set phis=phi(bottom)
       elem_state_phis(:,:,:) = elem_state_phinh_i(:,:,nlevp,tl%n0,:)
+
+      ! Set geopotential views
+      elem_state_phis_local_ptr       = c_loc(elem_state_phis_local)
+      elem_derived_gradphis_local_ptr = c_loc(elem_gradphis_local)
+      do ie=1,nelemd
+        elem_state_phis_local = elem(ie)%state%phis
+        elem_gradphis_local   = elem(ie)%derived%gradphis
+        call init_geopotential_c (ie-1, elem_state_phis_local_ptr, elem_derived_gradphis_local_ptr)
```
bartgol, line 129: An alternative is to use `derivative_mod` here, and do
```
elem_gradphis_local = gradient_sphere(elem_state_phis_local,deriv,elem(ie)%Dinv)
```
so that you don't need to call `compute_gradphis_c` anymore (simply do the calculation in F90).

components/homme/src/theta-l_kokkos/theta_f2c_mod.F90
tcclevenger: Yeah then, much better to add it in SCREAM. Didn't realize I could do that.

components/scream/src/dynamics/homme/interface/homme_driver_mod.F90
tcclevenger: Ok, I'll try that out.

tcclevenger:
@AaronDonahue This PR makes it to where the homme `elem(ie)%state%phis != 0` (like before) but matches the `phis` in the FM. Not sure if there is any thing that needs to be "turned on" etc in addition to this.. But now homme has non-zero topography if the IC contain non-zero topography.

components/scream/tests/uncoupled/homme/homme_standalone.cpp
```diff
@@ -69,6 +73,29 @@ TEST_CASE("scream_homme_standalone", "scream_homme_standalone") {
...
+    constexpr int NVL = HOMMEXX_NUM_PHYSICAL_LEV;
+
+    Kokkos::parallel_for(Kokkos::RangePolicy<>(0,nelem*NP*NP),
+                         KOKKOS_LAMBDA (const int idx) {
+      const int ie = idx/(NP*NP);
+      const int ip = (idx/NP)%NP;
+      const int jp = idx%NP;
+      EKAT_KERNEL_ASSERT(phinh_i(ie,n0,ip,jp,NVL) == phis(ie,ip,jp));
+    });
+  }
```
tcclevenger, line 97: I add a check that the internal dyn field `phinh_i(nlevs)` matches internal homme `phis`.

components/scream/src/dynamics/homme/atmosphere_dynamics.hpp
```diff
@@ -45,6 +45,9 @@ class HommeDynamics : public AtmosphereProcess
   // Note: field_mgrs[grid_name] is the FM on grid $grid_name
   void register_fields (const std::map<std::string,std::shared_ptr<FieldManager<Real>>>& field_mgrs) const;
 
+  // Retrieves an internal field, given field name and grid name.
+  const Field<Real>& get_internal_field (const std::string& name, const std::string& grid) const;
```
tcclevenger, line 49: To do the `phis` check in `homme_standalone.cpp`, I need to access an internal field, and it needed to be const. So I've changed it a bit. Is this ok?

tcclevenger (commented):
@bartgol Left some comments and questions. 

I went with the `use derivative_mod_base, only: gradient_sphere` option and added a check to the homme standalone test which asserts the `ElementGeometry::m_phis` matches `phinh_i(...,nlevs)` in the homme ad interface.

components/scream/src/dynamics/homme/atmosphere_dynamics.hpp
bartgol: Yeah, that's ok. I made it non-const originally, cause it was all used inside the class only, but this does not create issues.

(Btw: I'm changing this stuff in the restart branch, to make it more structured, so no constraint right here).

components/scream/tests/uncoupled/homme/homme_standalone.cpp
```diff
@@ -69,6 +73,29 @@ TEST_CASE("scream_homme_standalone", "scream_homme_standalone") {
...
   ad.initialize(atm_comm,ad_params,t0);
 
+  // Check that topography data from the FM matches Homme.
+  {
+    auto& geo = Homme::Context::singleton().get<Homme::ElementsGeometry>();
+    auto phis = geo.m_phis;
+
+    const auto& atm_process_group = ad.get_atm_processes();
+    const auto& process = atm_process_group->get_process(0);
+    const scream::HommeDynamics* homme_process = static_cast<const scream::HommeDynamics*>(&(*process));
```
bartgol, line 83: When dealing with shared_ptr, it's best to use the provided std cast function:
```
auto homme_process = std::dynamic_pointer_cast<const HommeDynamics>(process);
EKAT_REQUIRE_MSG (process, "Error! Cast to HommeDynamics failed.\n");
```

bartgol (changes requested):

bartgol (approved):

AaronDonahue (approved):

PR #1314, opened 2021-11-17
Add new coupled tests that include SPA

AaronDonahue:
Now that both phases of SPA have been integrated into SCREAM we can create coupled tests that include SPA.  This PR adds two new tests to the coupled scream test suite; `homme_shoc_cld_spa_p3_rrtmgp` and `shoc_cld_spa_p3_rrtmgp`.

Each is just an expansion of the original `homme+physics` and `full physics` tests to include SPA in the appropriate sequential location, just before P3.

There are also a few minor changes that were needed to get these tests working.


bartgol:
Question: since we don't have combinations of all physics packages, should we simply _add_ spa to the existing tests? If not, should we also add tests that run with just 1-2 physics packages (along side homme)? Like, should we add a homme+shoc+p3 test, without radiation?

I see some benefit in having reduce processes combination sets, since they can allow checking that everything works. OTOH, it requires that we provide stubs for missing parts that were supposedly provided by the removed processes.

AaronDonahue:
@bartgol , Yes.  I agree.  We should discuss which tests we want and either have a new PR or change this PR.

components/scream/tests/coupled/CMakeLists.txt
```diff
@@ -1,9 +1,9 @@
 # NOTE: if you have baseline-type tests, add the subdirectory OUTSIDE the following if statement
 if (NOT ${SCREAM_BASELINES_ONLY})
-  if ("${SCREAM_DYNAMICS_DYCORE}" STREQUAL "HOMME")
-    add_subdirectory(homme_physics)
-  endif()
   if (SCREAM_DOUBLE_PRECISION)
-    add_subdirectory(shoc_cld_p3_rrtmgp)
+    add_subdirectory(shoc_cld_spa_p3_rrtmgp)
```
bartgol, line 4: We should have probably called this test/folder `scream_physics` or `full_physics`, so it would have stayed the same, despite us adding something every time. If you feel like changing it, I'd be ok with that. Otherwise, nvm.

components/scream/tests/coupled/CMakeLists.txt
```diff
@@ -1,9 +1,9 @@
...
+    if ("${SCREAM_DYNAMICS_DYCORE}" STREQUAL "HOMME")
+      add_subdirectory(homme_shoc_cld_spa_p3_rrtmgp)
```
bartgol, line 6: I guess the name of this folder follows the other one. I'm not super fond of listing all atm procs in the name, since it makes the name long and hard to parse. But I won't raise a wall here.

components/scream/tests/coupled/homme_shoc_cld_spa_p3_rrtmgp/CMakeLists.txt
```diff
@@ -6,12 +6,22 @@ CreateDynamicsLib("theta-l_kokkos"  4   72   35)
...
-CreateUnitTest(homme_shoc_cld_p3_rrtmgp "homme_shoc_cld_p3_rrtmgp.cpp" "${NEED_LIBS}"
+set (NEED_LIBS cld_fraction shoc spa p3 scream_rrtmgp ${dynLibName} scream_control scream_share physics_share)
+CreateUnitTest(homme_shoc_cld_spa_p3_rrtmgp "homme_shoc_cld_spa_p3_rrtmgp.cpp" "${NEED_LIBS}"
                LABELS ${TEST_LABELS} 
                MPI_RANKS 1 ${SCREAM_TEST_MAX_RANKS}
-               PROPERTIES FIXTURES_SETUP homme_shoc_cld_p3_rrtmgp_generate_output_nc_files
+               PROPERTIES FIXTURES_SETUP homme_shoc_cld_spa_p3_rrtmgp_generate_output_nc_files
 )
+# Make sure the SPA data directory is correctly defined
+if (NOT EXISTS ${SCREAM_SPA_DATA_DIR})
```
bartgol, line 16: Uhm, does this mean we cannot locally test homme+physics if there's no spa data folder? This might be a valid reason for having two folders: `homme_physics` and `homme_physics_no_spa` or something. I'd like people to be able to locally test the full PD coupling scream, even if they don't have spa data...

components/scream/tests/coupled/shoc_cld_spa_p3_rrtmgp/CMakeLists.txt
```diff
@@ -1,12 +1,22 @@
...
-CreateUnitTest(shoc_cld_p3_rrtmgp shoc_cld_p3_rrtmgp.cpp "${NEED_LIBS}" LABELS ${TEST_LABELS}
+SET (TEST_LABELS "shoc;cld;spa;p3;rrtmgp;physics")
+SET (NEED_LIBS shoc cld_fraction spa p3 scream_rrtmgp scream_control scream_share physics_share)
+CreateUnitTest(shoc_cld_spa_p3_rrtmgp shoc_cld_spa_p3_rrtmgp.cpp "${NEED_LIBS}" LABELS ${TEST_LABELS}
   MPI_RANKS 1 ${SCREAM_TEST_MAX_RANKS}
-  PROPERTIES FIXTURES_SETUP shoc_cld_p3_rrmtgp_generate_output_nc_files
+  PROPERTIES FIXTURES_SETUP shoc_cld_spa_p3_rrmtgp_generate_output_nc_files
 )
+# Make sure the SPA data directory is correctly defined
+if (NOT EXISTS ${SCREAM_SPA_DATA_DIR})
```
bartgol, line 11: Same here: a user without spa data folder cannot run coupled physics test. We should have a full physics test without spa.

bartgol (changes requested):

AaronDonahue:
@bartgol , these are great comments.  I agree, we shouldn't shut a user down for running unit tests just because they haven't downloaded the SPA dataset and remap file.  I think we should flag cases where SPA data is missing and just not build the SPA tests.  And, we should definitely bring back the coupled test suites that don't have SPA.  I push a new version soon.

components/scream/tests/coupled/dynamics_physics/homme_shoc_cld_p3_rrtmgp/CMakeLists.txt
```diff
@@ -6,7 +6,7 @@ CreateDynamicsLib("theta-l_kokkos"  4   72   35)
 
 # Create the test
 SET (TEST_LABELS "dynamics;driver;shoc;cld;p3;rrtmgp;physics")
-set (NEED_LIBS cld_fraction shoc p3 scream_rrtmgp ${dynLibName} scream_control scream_share physics_share)
+set (NEED_LIBS cld_fraction shoc spa p3 scream_rrtmgp ${dynLibName} scream_control scream_share physics_share)
```
bartgol, line 9: Minor, but you do not need the spa library. Not a big deal, but during dev cycles, this dep can trigger rebuild of spa, even if I'm just trying to run this test.

components/scream/tests/coupled/dynamics_physics/homme_shoc_cld_spa_p3_rrtmgp/homme_shoc_cld_spa_p3_rrtmgp.cpp
```diff
@@ -0,0 +1,82 @@
+#include "catch2/catch.hpp"
+
+// The AD
+#include "control/atmosphere_driver.hpp"
+
+// Dynamics includes
+#include "dynamics/register_dynamics.hpp"
+#include "dynamics/homme/interface/scream_homme_interface.hpp"
```
bartgol, line 8: Minor: given what I see in the file, you don't need this include.

components/scream/tests/coupled/physics_only/shoc_cld_p3_rrtmgp/CMakeLists.txt
```diff
@@ -2,10 +2,10 @@ INCLUDE (ScreamUtils)
 
 # Create the test
 SET (TEST_LABELS "shoc;cld;p3;rrtmgp;physics")
-SET (NEED_LIBS shoc cld_fraction p3 scream_rrtmgp scream_control scream_share physics_share)
+SET (NEED_LIBS shoc cld_fraction spa p3 scream_rrtmgp scream_control scream_share physics_share)
```
bartgol, line 5: Same here: no need for spa.

components/scream/tests/coupled/physics_only/shoc_cld_spa_p3_rrtmgp/shoc_cld_spa_p3_rrtmgp.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+// Boiler plate, needed for all runs
+#include "control/atmosphere_driver.hpp"
+#include "share/atm_process/atmosphere_process.hpp"
+
+#include "physics/register_physics.hpp"
+#include "share/grid/mesh_free_grids_manager.hpp"
+
+// EKAT headers
+#include "ekat/ekat_pack.hpp"
```
bartgol, line 11: There's no use of Pack in this file, so this header can be removed.

components/scream/tests/coupled/physics_only/shoc_cld_spa_p3_rrtmgp/shoc_cld_spa_p3_rrtmgp.cpp
```diff
@@ -0,0 +1,64 @@
+#include <catch2/catch.hpp>
+
+// Boiler plate, needed for all runs
+#include "control/atmosphere_driver.hpp"
+#include "share/atm_process/atmosphere_process.hpp"
```
bartgol, line 5: You are not explicitly using anything from this header, so this include can be removed.

bartgol (approved):
I only have some `-luca-pedantic` comments.

components/scream/tests/coupled/dynamics_physics/homme_shoc_cld_p3_rrtmgp/CMakeLists.txt
AaronDonahue: we actually do, if we use `register_physics.hpp` in the main file.  Because register_physics still registers SPA even if it's not used.

bartgol: Ugh, then we should not call register_phyisics, and rather do the registration manually.

bartgol (approved):

PR #1315, opened 2021-11-17
Add an aquaplanet compset for SCREAMv0

### LINKED ISSUE #878: Build error with ne4 F case in physpkg.F90
This build error is reproducible with compset FC5AV1C-L and res ne4_ne4 on
Cori, Theta and anlworkstation.

[GNU compiler]
```
components/eam/src/physics/cam/physpkg.F90:850:42:

     call radiation_init(phys_state,pbuf2d)
                                          1
Error: More actual than formal arguments in procedure call at (1)
```
[Intel compiler]
```
components/eam/src/physics/cam/physpkg.F90(850): error #6784: The number of actual arguments cannot be greater than the number of dummy arguments.   [RADIATION_INIT]
    call radiation_init(phys_state,pbuf2d)
---------^
```
This error is only reproducible with scream (not reproducible with E3SM).

brhillman:
This adds two new aquaplanet compsets for SCREAMv0 (F-SCREAM-LR-AQP1 and F-SCREAM-HR-AQP1). Two new use case files were needed to set namelist options, copied largely from the existing EAM use cases for AQP configurations. These compsets are needed to generate initial conditions for SCREAMv1 tests until SCREAMv1 correctly deals with topography, but will be useful in their own right as well. Fixes #878

### DIFF
--- a/components/eam/src/physics/rrtmg/radiation.F90
+++ b/components/eam/src/physics/rrtmg/radiation.F90
@@ -76,6 +76,9 @@ logical :: spectralflux  = .false. ! calculate fluxes (up and down) per band.
 
 logical :: use_rad_dt_cosz  = .false. ! if true, uses the radiation dt for all cosz calculations !BSINGH - Added for solar insolation calc.
 
+! Flag to indicate whether to read optics from spa netcdf file
+! NOTE: added for consistency with RRTMGP; this is non-functioning for RRTMG!
+logical :: do_spa_optics = .false.
 
 character(len=4) :: diag(0:N_DIAG) =(/'    ','_d1 ','_d2 ','_d3 ','_d4 ','_d5 ','_d6 ','_d7 ','_d8 ','_d9 ','_d10'/)
 
@@ -117,7 +120,8 @@ subroutine radiation_readnl(nlfile, dtime_in)
 
    ! Variables defined in namelist
    namelist /radiation_nl/ iradsw, iradlw, irad_always, &
-                           use_rad_dt_cosz, spectralflux
+                           use_rad_dt_cosz, spectralflux, &
+                           do_spa_optics
 
    ! Read the namelist, only if called from master process
    ! TODO: better documentation and cleaner logic here?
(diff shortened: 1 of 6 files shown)


components/eam/src/physics/rrtmg/radiation.F90
```diff
@@ -362,13 +372,13 @@ end function radiation_nextsw_cday
 
 !================================================================================================
 
-  subroutine radiation_init(phys_state)
```
AaronDonahue, line 365: Will these changes conflict with an upstream merge of E3SM?

AaronDonahue:
Does this also remedy issue #878 ?

components/eam/src/physics/rrtmg/radiation.F90
brhillman: Sorry I let this go. I think it only creates a conflict if E3SM also modified these lines. This is a very SCREAM-specific thing though, I don't think any of the SPA infrastructure is in E3SM upstream, so I'm not sure if this change should get pushed back upstream. Thoughts?

brhillman:
> Does this also remedy issue #878 ?

Yes. Updated description.

AaronDonahue:
@brhillman , is there a reason this PR is still WIP?  Were you planning a follow up PR to create a SCREAMv1 Aquaplanet compset?

components/eam/src/physics/rrtmg/radiation.F90
bartgol: We do have scream-specific things in other parts, like Homme. One way to isolate them is to do something like
```
#ifdef SCREAM
  ! scream impl
#else
  ! eam impl
#endif
```

bartgol:
If I understand correctly, some F90 mods are to make the impl of RRTMG match the impl of RRTMGP. Is that really necessary though? I don't think we need rrtmg in scream anyway, no? I get that it's nice to keep the two interfaces close, but maybe we don't need to?

brhillman:
@AaronDonahue sorry, this was WIP because I was waiting to hear about resolution on your review comments. There's no reason to keep this WIP.

@bartgol one reason to fix the RRTMG impl here is so that we can run BFB comparisons between EAM and SCREAM repos (which are non-BFB at the moment for some reason). Other's have also wanted to run EAM compsets with our SCREAM repo for various reasons, and that's impossible without the RRTMG mod in this PR.

AaronDonahue (approved):

PR #1316, opened 2021-11-17
changed ice mixing ratio threshold to 1e-5 kg/kg for setting ice clou…

hassanbeydoun:
…d fraction to 1


PeterCaldwell (approved):
Looks good.

PR #1319, opened 2021-11-18
Add ne256pg2 to r0125 atm2rof maps

crterai:
Adds the map files for running SCREAM using the ne256pg2_r0125_oRRS18to6v3 grid. Specifically, it adds the atm2rof mapping files that are currently missing in config_grids.xml.


ndkeen (approved):
I tested this with a few different cases to ensure it runs. Runs for at least 24 days.

PR #1323, opened 2021-11-18
Update the capabilities of SCREAM's scream_input.yaml file

jgfouca:
We add some customized syntax to increase the expressiveness
of the items in this file:

'${VAR}' will be used to refer to env variables in the CIME case

'<cond : key1 => val1 : key2 => val2 : elseval>' will be used to express conditional
statements. If cond matches key1, then the expression evaluates to val1; if cond
matches key2, then the expression evaluates to val2; if it matches neither, then
the expression evaluates to elseval. The elseval component of this expression is optional.

This should give SCREAM the ability to express what it needs in this file
without having to resort to lots of hardcoded special handling.

### DIFF
--- a/components/scream/data/scream_input.yaml
+++ b/components/scream/data/scream_input.yaml
@@ -55,8 +55,6 @@ SCREAM:
     input_root                  : "${DIN_LOC_ROOT}"
 
 
-  Test Var: "<${ATM_GRID} : ne4np4 => 42 : 420>"
-
 # Note: HOMME settings will be translated into data/namelist.nl
 HOMME:
   ctl_nl:
--- a/components/scream/data/scream_input.yaml
+++ b/components/scream/data/scream_input.yaml
@@ -49,11 +49,13 @@ Atmosphere Driver:
 
 SCREAM:
 
-  Start Date: 01-01-2000 # For CIME cases, do NOT set this! It will be set by CIME's RUN_STARTDATE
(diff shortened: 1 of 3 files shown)


components/scream/data/scream_input.yaml
```diff
@@ -49,12 +49,14 @@ Atmosphere Driver:
...
 
-  Start Date: 01-01-2000 # For CIME cases, do NOT set this! It will be set by CIME's RUN_STARTDATE
+  Start Date: "${RUN_STARTDATE}"
 
   Input Files:
-    input_root                  : default # this will default to DIN_LOC_ROOT for the current machine
+    input_root                  : "${DIN_LOC_ROOT}"
 
 
+  Test Var: "<${ATM_GRID} : ne4np4 => 42 : 420>"
```
bartgol, line 58: Might add a second example `Test Var 2: ...` where you show how to use 2+ matching values for the cond var. So it's clearer that this is not (just) a ternary op.

jgfouca: @bartgol , this `Test Var` item should not have been left in this file to begin with. I temporarily inserted it just to test that switch statement syntax worked. I added some comments to the top of buildnml to hopefully clarify how the user can use switch statements.

bartgol (approved):
Since you told me offline that this kind of logic is almost verbatim taken from another project where it was tested in depth, I'm ok with not having any scripts test.

PR #1325, opened 2021-11-24
Update SHOC default tunable parameters

bogensch:
This PR updates certain SHOC tunable parameters to their optimal/intended values.  Specifically:

- SHOC scalar variances tunable parameters are now set to 1.0
- SHOC stable boundary layer diffusivities are now set 0.1 (these were the values that were used for the DYAMOND 2 run in Caldwell et al. 2021 but they were never updated in master).

F90 and C++ codes have both been updated to reflect this.

This PR will change answers, therefore SHOC run and compare tests (F90 and C++) are expected fails.

### DIFF
--- a/components/eam/bld/namelist_files/namelist_defaults_eam.xml
+++ b/components/eam/bld/namelist_files/namelist_defaults_eam.xml
@@ -792,9 +792,9 @@
 <do_shoc_sgs                 > .false. </do_shoc_sgs>
 <do_shoc_sgs   shoc_sgs="1"  > .true. </do_shoc_sgs>
 <shoc_timestep               > -1 </shoc_timestep>
-<shoc_thl2tune               > 0.15D0 </shoc_thl2tune>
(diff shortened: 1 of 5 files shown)


AaronDonahue:
This PR is failing the AT but it looks like it is just the `shoc_run_and_cmp` tests, which would make sense for a non-BFB change like this.  Is this PR ready to review and merge now? - @bogensch

PeterCaldwell:
mappy full_sp_debug is consistently failing for this PR, which makes no sense since Peter just changed the values for some constants. I asked Peter B to see whether rebasing off master fixes the issue. It would also be useful to see which specific unit test is failing for this case.

bogensch:
@PeterCaldwell I was able to get the tests to fail on quartz/syrah with debug and single precision turned on, with this branch only (master passes).  I'm investigating.  But all this suggests we should hold off on merging until I figure out what the heck is going on.

PeterCaldwell (approved):

PR #1330, opened 2021-12-03
Add units to new diagnostic cloud fraction outputs

bogensch:
Add units to cloud diagnostics outfield calls from PR #1307.  Leaving these fields blank will mean a units attribute will not be generated in the netcdf file for these particular variables, which could cause problems for analysis code that automatically reads in units for all variables.  Plus, units is good in general for provenance.


hassanbeydoun (approved):
Good catch, thanks Peter B!

PR #1332, opened 2021-12-07
SHOC surface sensible heat flux fix

### LINKED ISSUE #1333: SHOC interface (cpp code) surface flux index bug
https://github.com/E3SM-Project/scream/blob/4b5842e40bc0108a765e405c932087279a542e58/components/scream/src/physics/shoc/atmosphere_macrophysics.hpp#L162-L165

When computing the surface fluxes in the SHOC cpp interface there appears to be a index bug when attempting to use the interface level density (rho_i) at the surface.  Instead of "rrho_i(i,nlev_v)[nlev_p]" this needs to be "rrho_i(i,nlevi_v)[nlevi_p]" for the lines referenced.  The current cpp implementation is at odds with the F90 implementation and using density at the wrong level.

Of course, with my ignorance of cpp code, there is the possibility I am misreading something?...

bogensch:
Takes into account exner function when passing surface sensible heat flux to SHOC.  PR is answer changing.  However, all tests pass because the SHOC interface is NOT exercised in the SHOC run and compare tests.

Also this PR closes #1333

### DIFF
--- a/components/scream/src/physics/shoc/atmosphere_macrophysics.cpp
+++ b/components/scream/src/physics/shoc/atmosphere_macrophysics.cpp
@@ -247,6 +247,7 @@ void SHOCMacrophysics::initialize_impl ()
   //       variables a local view is constructed.
   const auto& T_mid            = get_field_out("T_mid").get_view<Spack**>();
   const auto& p_mid            = get_field_in("p_mid").get_view<const Spack**>();
+  const auto& p_int            = get_field_in("p_int").get_view<const Spack**>();
   const auto& pseudo_density   = get_field_in("pseudo_density").get_view<const Spack**>();
   const auto& omega            = get_field_in("omega").get_view<const Spack**>();
   const auto& surf_sens_flux   = get_field_in("surf_sens_flux").get_view<const Real*>();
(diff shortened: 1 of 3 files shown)


bogensch:
@AaronDonahue and @tcclevenger if you could focus on the cpp changes (since I am super uncomfortable with the cpp code) to make sure I did them correctly and consistently, it would be greatly appreciated.

components/scream/src/physics/shoc/atmosphere_macrophysics.hpp
```diff
@@ -159,7 +159,11 @@ class SHOCMacrophysics : public scream::AtmosphereProcess
 
       const int nlev_v = (nlev-1)/Spack::n;
       const int nlev_p = (nlev-1)%Spack::n;
+      const auto& exner_int = PF::exner_function(p_int(i,nlevi_v)[nlevi_p]);
+      const auto& inv_exner_int_surf = 1/exner_int;
+
       wpthlp_sfc(i) = surf_sens_flux(i)/(cpair*rrho_i(i,nlev_v)[nlev_p]);
```
tcclevenger, line 165: Should all the below `*_sfc(i) =` lines be changed to use `(nlevi_v)[nlevi_p]`?
```suggestion
      wpthlp_sfc(i) = surf_sens_flux(i)/(cpair*rrho_i(i,nlevi_v)[nlevi_p]);
```

components/scream/src/physics/shoc/atmosphere_macrophysics.cpp
```diff
@@ -287,7 +288,7 @@ void SHOCMacrophysics::initialize_impl ()
   const Real z_surf = 0.0;
 
   shoc_preprocess.set_variables(m_num_cols,m_num_levs,m_num_tracers,z_surf,m_cell_area,
-                                T_mid,p_mid,pseudo_density,omega,phis,surf_sens_flux,surf_latent_flux,
+                                T_mid,p_mid,p_int,pseudo_density,omega,phis,surf_sens_flux,surf_latent_flux,
                                 surf_mom_flux,qv,qc,tke,tke_copy,z_mid,z_int,cell_length,
                                 dse,rrho,rrho_i,thv,dz,zt_grid,zi_grid,wpthlp_sfc,wprtp_sfc,upwp_sfc,vpwp_sfc,
                                 wtracer_sfc,wm_zt,inv_exner,thlm,qw);
```
tcclevenger, line 294: A few lines down there is 
```
input.presi  = get_field_in("p_int").get_view<const Spack**>();
```

Could you change this to 
```
input.presi  = p_int;
```
since we are now declaring `p_int`?

tcclevenger (commented):
Just a comment about using `nlevi` indexing. And a cosmetic request inside `SHOCMacrophysics::initialize_impl ()`. Otherwise the c++ looks good to me.

tcclevenger (approved):

PR #1334, opened 2021-12-07
Make vertically intergated cloud fractions use TOT_CLOUD_FRAC instead…

hassanbeydoun:
… of CLOUD

### DIFF
--- a/components/eam/src/physics/cam/cloud_cover_diags.F90
+++ b/components/eam/src/physics/cam/cloud_cover_diags.F90
@@ -25,7 +25,6 @@ subroutine cloud_cover_diags_init(sampling_seq)
   character(len=*), intent(in) :: sampling_seq
   logical :: history_amwg         ! output the variables used by the AMWG diag package
 
-  !call addfld ('CLOUD',(/ 'lev' /), 'A','1','Cloud fraction'                        , sampling_seq=sampling_seq)
   call addfld ('CLDTOT',horiz_only,    'A','1','Vertically-integrated total cloud'     , sampling_seq=sampling_seq)
   call addfld ('CLDLOW',horiz_only,    'A','1','Vertically-integrated low cloud'       , sampling_seq=sampling_seq)
   call addfld ('CLDMED',horiz_only,    'A','1','Vertically-integrated mid-level cloud' , sampling_seq=sampling_seq)
@@ -35,7 +34,6 @@ subroutine cloud_cover_diags_init(sampling_seq)
   call phys_getopts(history_amwg_out           = history_amwg  )
  
   if (history_amwg) then
-!      call add_default ('CLOUD   ', 1, ' ')
       call add_default ('CLDTOT  ', 1, ' ')
       call add_default ('CLDLOW  ', 1, ' ')
       call add_default ('CLDMED  ', 1, ' ')
@@ -70,7 +68,6 @@ subroutine cloud_cover_diags_out(lchnk, ncol, cld, pmid, nmxrgn, pmxrgn )
   call outfld('CLDMED  ',clmed  ,pcols,lchnk)
   call outfld('CLDHGH  ',clhgh  ,pcols,lchnk)
 
-  !call outfld('CLOUD   ',cld    ,pcols,lchnk) 
 
 end subroutine cloud_cover_diags_out
 
--- a/components/eam/src/physics/cam/cloud_cover_diags.F90
+++ b/components/eam/src/physics/cam/cloud_cover_diags.F90
@@ -25,7 +25,7 @@ subroutine cloud_cover_diags_init(sampling_seq)
   character(len=*), intent(in) :: sampling_seq
   logical :: history_amwg         ! output the variables used by the AMWG diag package
 
-  call addfld ('CLOUD',(/ 'lev' /), 'A','1','Cloud fraction'                        , sampling_seq=sampling_seq)
+  !call addfld ('CLOUD',(/ 'lev' /), 'A','1','Cloud fraction'                        , sampling_seq=sampling_seq)
   call addfld ('CLDTOT',horiz_only,    'A','1','Vertically-integrated total cloud'     , sampling_seq=sampling_seq)
   call addfld ('CLDLOW',horiz_only,    'A','1','Vertically-integrated low cloud'       , sampling_seq=sampling_seq)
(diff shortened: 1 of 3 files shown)


components/eam/src/physics/cam/cloud_diagnostics.F90
```diff
@@ -53,12 +53,13 @@ subroutine cloud_diagnostics_register
     use phys_control,  only: phys_getopts
     use physics_buffer,only: pbuf_add_field, dtype_r8, dtype_i4
 
-    character(len=16) :: rad_pkg, microp_pgk
+    character(len=16) :: rad_pkg, microp_pgk, macrop_pgk
```
PeterCaldwell, line 56: I think you mean pkg instead of pgk?

hassanbeydoun: I just copied 'microp_pgk' which was already there.

components/eam/src/physics/cam/cloud_diagnostics.F90
```diff
@@ -181,6 +186,7 @@ subroutine cloud_diagnostics_init()
 
 
     if (history_amwg) then
+       call add_default ('CLOUD   ', 1, ' ')
```
PeterCaldwell, line 189: Let's make our new TOT_CLD_FRAC the default output rather than CLOUD...

PeterCaldwell: Though I guess that would give bad output if we ever didn't use SHOC... And we never do history_amwg for high-res SCREAM runs. So never mind. Don't bother with this change.

PeterCaldwell:
Looks good. I'm going to wait to see if tests are BFB. I could imagine the AD-level tests using cprnc failing because we changed the definition of some variables.

components/eam/src/physics/cam/cloud_diagnostics.F90
```diff
@@ -284,6 +291,7 @@ subroutine cloud_diagnostics_calc(state,  pbuf)
 
     itim_old = pbuf_old_tim_idx()
     call pbuf_get_field(pbuf, cld_idx, cld, start=(/1,1,itim_old/), kount=(/pcols,pver,1/) )
+    call outfld('CLOUD',cld(:,:),pcols,lchnk)
```
bogensch, line 294: I guess I'm a little confused why the CLOUD outfld call was moved from cloud_cover_diags to here because it seems you are outputting the exact same variable.... Or was your intention to output "tot_cloud_frac" here instead of the E3SM "cld"?

components/eam/src/physics/cam/shoc_intr.F90
```diff
@@ -580,7 +583,7 @@ subroutine shoc_tend_e3sm( &
    real(r8) :: ice_cloud_frac(pcols,pver)          ! ice number aware cloud fraction, 0 or 1
    real(r8) :: precipitating_ice_frac(pcols,pver)        ! precipitating ice fraction, 0 or 1
    real(r8) :: liq_cloud_frac(pcols,pver)     
-   real(r8) :: tot_cloud_frac(pcols,pver)         
+   !real(r8) :: tot_cloud_frac(pcols,pver)
```
bogensch, line 586: I'm in favor of removing this line rather than commenting it out.

components/eam/src/physics/cam/cloud_diagnostics.F90
hassanbeydoun: Previously, the CLOUD oufld call was done in the cloud_cover_diags_out subroutine. But since we are now using that same subroutine for either CLOUD or TOT_CLOUD_FRACTION, I didn't want TOT_CLOUD_FRACTION to overwrite CLOUD in the output. Does that make sense?

components/eam/src/physics/cam/cloud_diagnostics.F90
PeterCaldwell: Yeah, but micro_pgk was obviously a typo!

components/eam/src/physics/cam/cloud_cover_diags.F90
```diff
@@ -35,7 +35,7 @@ subroutine cloud_cover_diags_init(sampling_seq)
   call phys_getopts(history_amwg_out           = history_amwg  )
  
   if (history_amwg) then
-      call add_default ('CLOUD   ', 1, ' ')
+!      call add_default ('CLOUD   ', 1, ' ')
```
PeterCaldwell, line 38: this could also probably be deleted rather than commented - we get the same output from the new loc of CLOUD outfld so will never go back to this old ver.

components/eam/src/physics/cam/cloud_cover_diags.F90
```diff
@@ -70,7 +70,7 @@ subroutine cloud_cover_diags_out(lchnk, ncol, cld, pmid, nmxrgn, pmxrgn )
   call outfld('CLDMED  ',clmed  ,pcols,lchnk)
   call outfld('CLDHGH  ',clhgh  ,pcols,lchnk)
 
-  call outfld('CLOUD   ',cld    ,pcols,lchnk) 
+  !call outfld('CLOUD   ',cld    ,pcols,lchnk)
```
PeterCaldwell, line 73: also delete commented out thing.

components/eam/src/physics/cam/shoc_intr.F90
PeterCaldwell: I agree with Peter.

components/eam/src/physics/cam/cloud_diagnostics.F90
PeterCaldwell: Ah yes, Hassan is right. We call cloud_cover_diags_out below with *either* CLOUD or TOT_CLOUD_FRACTION as an input depending on whether we're using SHOC or not. So the variable CLOUD in cloud_cover_diags_out could refer to either the old or the new 3d cloud fraction. But we want to preserve the ability to write out both the old CLOUD (because that's the weirdly-defined cloud that the model actually uses) and also the new TOT_CLOUD_FRAC (because that's what will keep humans from rioting). So Hassan needed to move where the old CLOUD was output.

PeterCaldwell (approved):
Ahh. Just right. Thanks Hassan!!!

PR #1336, opened 2021-12-08
Add diagnostics specific for DP-SCREAM

bogensch:
Adds a subroutine specific to DP-SCREAM to compute and output resolved turbulence statistics.  These are computed in a manner so they are consistent and comparable with each of SHOC's SGS statistics.


components/eam/src/dynamics/se/se_single_column_mod.F90
```diff
@@ -505,4 +510,150 @@ subroutine iop_domain_relaxation(elem,hvcoord,hybrid,t1,dp,nelemd_todo,np_todo,d
...
+      ! Sum each variable for each level
+      global_shared_buf(ie,1) = global_shared_buf(ie,1) + SUM(thetal(:,:,k))
+      global_shared_buf(ie,2) = global_shared_buf(ie,2) + SUM(qw(:,:,k))
+      global_shared_buf(ie,3) = global_shared_buf(ie,3) + SUM(elem(ie)%state%v(:,:,1,k,t1))
+      global_shared_buf(ie,4) = global_shared_buf(ie,4) + SUM(elem(ie)%state%v(:,:,2,k,t1))
+    enddo
+
+    ! Compute the domain mean
+    call wrap_repro_sum(nvars=4, comm=hybrid%par%comm)
+    domain_thetal(k) = global_shared_sum(1)/(dble(nelem)*dble(np)*dble(np))
```
crterai, line 582: It looks like each column is weighted equally. my impression was that some nodes were larger than others. If so, do the variables need to be area-weighted in this case?

components/eam/src/dynamics/se/se_single_column_mod.F90
```diff
@@ -505,4 +510,150 @@ subroutine iop_domain_relaxation(elem,hvcoord,hybrid,t1,dp,nelemd_todo,np_todo,d
...
+
+    ! Compute liquid water potential temperature
+    do k=1,nlev
+      thetal(:,:,k) = thetal(:,:,k) - (latvap/Cp)*(elem(ie)%state%Q(1:np,1:np,k,2))
+      qw(:,:,k) = elem(ie)%state%Q(1:np,1:np,k,1) + elem(ie)%state%Q(1:np,1:np,k,2)
+    enddo
+
+    ! Compute air density
+    do k=1,nlev
+      rho(:,:,k) = pres(k)/(Rgas*temperature(:,:,k))
```
crterai, line 604: The dry air density seems to be used. I want to verify that this is what you'd want, rather than moist air density.

crterai (commented):
I had three minor questions, but if those all are what you'd wanted, I can approve.

components/eam/src/dynamics/se/se_single_column_mod.F90
```diff
@@ -505,4 +510,150 @@ subroutine iop_domain_relaxation(elem,hvcoord,hybrid,t1,dp,nelemd_todo,np_todo,d
...
+
+      ! Sum each variable for each level
+      global_shared_buf(ie,1) = global_shared_buf(ie,1) + SUM(thetal(:,:,k))
+      global_shared_buf(ie,2) = global_shared_buf(ie,2) + SUM(qw(:,:,k))
+      global_shared_buf(ie,3) = global_shared_buf(ie,3) + SUM(elem(ie)%state%v(:,:,1,k,t1))
+      global_shared_buf(ie,4) = global_shared_buf(ie,4) + SUM(elem(ie)%state%v(:,:,2,k,t1))
+    enddo
+
+    ! Compute the domain mean
+    call wrap_repro_sum(nvars=4, comm=hybrid%par%comm)
```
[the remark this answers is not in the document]
bogensch: Yes, that is exactly right.  I've verified this is the case with some simple tests.

components/eam/src/dynamics/se/se_single_column_mod.F90
bogensch: This is what I intended.

components/eam/src/dynamics/se/se_single_column_mod.F90
bogensch: For the planar grid (which these diagnostics are specific for, for DP-SCREAM), this is not an issue.

crterai (approved):
Thanks for answering all my questions.

PR #1337, opened 2021-12-08
SCREAM: bug fix in column ops packed scan sum

bartgol:
The packed version of scan sum had a bug in the case where the number of packs is the same for both midpoints and interface quantities (e.g., 72 levels with packsize 16, yields 5 packs for both mid/int quantities).

Thanks @brhillman and @AaronDonahue for spotting this issue!


brhillman:
Is there a test we can add that would have caught this issue?

AaronDonahue (approved):

PR #1340, opened 2021-12-09
set SCREAM to run with adjust_ps=.false.

mt5555:
dont reproduce buggy behavior used in E3SM v1/v2

non-BFB, with minor impacts on climate

### DIFF
--- a/components/homme/src/share/prim_driver_base.F90
+++ b/components/homme/src/share/prim_driver_base.F90
@@ -1590,7 +1590,7 @@ contains
   if (dt_remap_factor==0) then
      adjust_ps=.true.   ! stay on reference levels for Eulerian case
   else
-     adjust_ps=.true.   ! Lagrangian case can support adjusting dp3d or ps
+     adjust_ps=.false.  ! Lagrangian case can support adjusting dp3d or ps
   endif
 #else
   adjust_ps=.true.      ! preqx requires forcing to stay on reference levels


PeterCaldwell:
Hey @ndkeen - can you confirm this is the PR you were expecting/approve the PR? I had assumed the change was more complicated than literally just setting adjust_ps=.false. ! Even I could do that!

Also, I'm confused why this PR is merging into E3SM-Project:master instead of just master (line just under the PR title above). Is this a PR to SCREAM or to E3SM?

Also, we probably need to make a corresponding change to the C++ to avoid causing all our tests to fail?

oksanaguba:
Following https://github.com/E3SM-Project/E3SM/pull/4533 , adjust_ps=false is not tested . I would suggest we make a namelist variable (defaulted to true), add a standalone test with forcing (bfb type), and merge this to e3sm master first.

oksanaguba:
since we will have a bfb test, we will have a namelist variable in cxx, too.

bartgol:
This is a change in the homme folder. I think our current "policy" for non-scream folder changes it to make a change in the upstream, and bring it in with the next upstream-merge. I'm not sure 

I understand that E3SM v1/v2 don't use this option, so maybe it should be guarded by an `#ifdef SCREAM`. Or maybe even better, we could add a specific #cmakedefine entry in config.h.cmake.in (in theta-l_kokkos). Something like `HOMME_TRACERS_FORCING_ADJUST_PS`, so that we can start testing it in Homme without having to define the SCREAM macro (which might trigger additional chunks of the code).
 
> Also, I'm confused why this PR is merging into E3SM-Project:master instead of just master (line just under the PR title above). Is this a PR to SCREAM or to E3SM?

Unfortunately, for PRs across forks, github seems to name the branches as `$user:branch_name` (for user accounts) or `$organization:branch_name`. It would be much clearer if it used `$user:repo_name:branch_name` or `$organization:repo_name:branch_name`. On github defense, each user/org can only have one fork of a given repo, and I guess they don't see often a user/org forking one of their own repos (like scream does), so the notation they have is _usually_ unambiguous.

> Also, we probably need to make a corresponding change to the C++ to avoid causing all our tests to fail?

Good point. Perhaps it's another reason to do this change in the upstream.

mt5555:
I'd prefer not to make this a namelist option, since that touches several files just for the sake of preserving buggy behavoir.  I'd prever the #ifdef SCREAM idea.  I will attempt to follow this git best practice:

Apply this change to the common parent of SCREAM and E3SM.  submit a PR to both E3SM and SCREAM.  It can then be merged into SCREAM right away  (our goal), and that wont produce conflicts when it finally makes it way into E3SM.

mt5555:
@oksanaguba raised a complex issue with regards to testing Fortran bfb with cxx code that would break if we try to put this PR in E3SM (with an #ifdef SCREAM) and then merge into SCREAM.  Hence I closed those PRs and I am reopining this one.  

I propose we allow for a short term violation of the upstream merge policy, and allow this simple 1 line change to go into the SCREAM repo only.   This will better reflect current practice as to how we run SCREAM v0.1.  

I will make an issue describing the work that needs to be done to allow proper testing of the cxx code (which is done in the E3SM repo and will not be affected by this PR)

bartgol:
I guess every rule has exceptions. I'm ok with this tiny mod in the SCREAM repo, if it's needed asap by v0.

oksanaguba (approved):

PeterCaldwell:
I'd really like this PR integrated ASAP. Are there any reasons why we can't do that? I don't understand the "complex issue" with testing Oksana brought up so I don't want to inadvertently cause huge problems...

mt5555:
it's been approved. Noel can merge anytime.

ndkeen (approved):

PeterCaldwell (approved):

PeterCaldwell:
Merging this even though it doesn't pass tests b/c 1) the fails are clearly due to mappy and weaver being offline or otherwise unhealthy 2). it is passing on Blake, 3). this is a 1 line change which switches to a config we've been running for months by making manual changes, so we're pretty sure it is fine. My only worry was that it causes AT fails due to it only changing F90 and not C++, but the Blake pass and conceptual expectations suggest this will be fine.

PR #1341, opened 2021-12-09
set SCREAM to run with adjust_ps=.false.

mt5555:
dont reproduce buggy behavior used in E3SM v1/v2

[BFB]
In the SCREAM model, will be non-BFB, with minor impacts on climate

this is a branch from the e3sm and scream merge-base, and thus this branch applies cleanly to both repos and wont produce future conflicts.   PR to upstream:  https://github.com/E3SM-Project/E3SM/pull/4692

### DIFF
--- a/components/homme/src/share/prim_driver_base.F90
+++ b/components/homme/src/share/prim_driver_base.F90
@@ -1591,7 +1591,7 @@ contains
      adjust_ps=.true.   ! stay on reference levels for Eulerian case
   else
 #ifdef SCREAM
-     adjust_ps=.false.  ! Lagrangian case should adjusting dp3d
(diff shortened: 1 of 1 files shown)


bartgol:
Sorry, I missed the fact that this is for SCREAMv0. The macro `SCREAM` is currently only present when we build Homme from SCREAMv1, not from EAM. I suspect it should not be too hard to add `-DSCREAM` in the eam build system, if the compset is a SCREAM compset. Unfortunately, there's no "scream" config option in EAM, but we might emulate by doing

```
if ($macrophys_pkg eq 'shoc_sgs' and $microphys_pkg eq 'p3'){ 
    $cfg_cppdefs .= ' -DSCREAM';
}
```
Since neither shoc nor p3 are in the e3sm repo, maybe we could just check one of them.

PR #1344, opened 2021-12-10
Bartgol/upstream merge 12092021

bartgol:
There were only a handful of conflicts:

- `components/eam/src/physics/rrtmgp/radiation.F90`: the upstream changes were mostly cleanup (unused lines removed), but scream contained some added subroutines. @brhillman you should make sure I didn't mess up.
- `components/homme/src/theta-l_kokkos/prim_driver_mod.F90`: some added subroutines in the upstream, for C<->F syncing (picked all mods from upstream).
- `cime_config/machines/cmake_macros/intel_[quartz|syrah].cmake`: seems like the upstream removed some redundant flags.


jgfouca (approved):
Thanks, Luca!

bartgol:
Manually tested on mappy, and it passes.

brhillman (approved):

bartgol:
Given that:

- previous tests on blake/weaver passed
- manual tests on mappy for standalone pass
- scream was already non-bfb against e3sm (for eam cases) before this PR
- weaver is offline, and AT might get stuck "testing" other PRs for 30m each (waiting for weaver jobs to start)

I'm gonna manually merge this PR. With @brhillman we decided that tracking the reason why our repo's eam runs diff from e3sm repo's run is *critical*, but does not have to happen in this PR (given the third point above). He's already working on tracking the diffs, as a separate effort, without paying attention to where/when the diffs originated.

PR #1345, opened 2021-12-10
DP-SCREAM mpi communicator for Data ocean initialization fix

bogensch:
Add mpi rank when reading in data ocean namelist to prevent it for doing this for all processes.  This fix does not change the solution, but prevents the log files from being inundated by lots of messages when many processors are used.  Previously only the SCM (which always uses one processor) only used this code so the communicator was not included for this call.


ndkeen (approved):
This is fine, but note you aren't adding "mpi rank", you are adding the mpi communicator (mpicom) so that the routine can check it's present and only do certain things on rank0

PeterCaldwell:
BFB, just changes 1 line which only affects the F90 code and only changes log file output. Mappy still showing fails because it isn't mounted correctly right now.

PR #1346, opened 2021-12-11
Update YAKL and RRTMGP to fix memory leak on GPU

brhillman:
The YAKL pool allocator had a bug that resulted in a memory leak when running on GPUs. This updates to the latest YAKL, which includes a fix for this. RRTMGP is also updated because the YAKL API is changed. This removes the requirement to export `GATOR_DISABLE=1` when running on summit.

### DIFF
--- a/components/scream/src/physics/rrtmgp/mo_garand_atmos_io.h
+++ b/components/scream/src/physics/rrtmgp/mo_garand_atmos_io.h
@@ -3,7 +3,6 @@
 
 #include "cpp/rrtmgp_const.h"
 #include "cpp/rrtmgp/mo_gas_concentrations.h"
-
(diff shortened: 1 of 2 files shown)


AaronDonahue (approved):

AaronDonahue:
@brhillman , it seems like this PR is just one manual test on mappy away from being able to be merged.  Is that true?  Would you like to run `test-all-scream` on mappy with this PR and report back if it passes? then I can manually merge.

brhillman:
@AaronDonahue tests passed, but I needed to do a rebase to resolve merge conflicts. Should be good to go now but I'm going to run the CIME tests to make sure we do not need to bless the v0 baselines.

AaronDonahue (approved):

PR #1347, opened 2021-12-14
Reset default cosp_ncolumns to 10

wlin7:
The hardwired default for cosp_ncolumns is 50. This PR resets it to
10 in namelist_defaults_eam.xml, and the new default is grid 
independent for all config with phys='default'. When cosp_lite is true,
cosp_ncolumns is always set to 10. cosp_ncolumns is also set to 10 when
any of the following is true: cosp_passive, cosp_active, cosp_isccp,
and cosp_amwg. Note that setting cosp_lisccp_sim = .true. does not lead to
cosp_isccp = .true. during runtime. But it does serve to set lisccp_sim to
true, which is the actual control for everything regarding the ISCCP simulator.

[BFB] would impact cosp output for simulations that not effectively 
set default to 10, for example, when not using cosp_lite or cosp_amwg.

### DIFF
--- a/components/eam/bld/build-namelist
+++ b/components/eam/bld/build-namelist
@@ -915,6 +915,12 @@ if ($rad_pkg eq 'rrtmg' or $chem =~ /waccm_mozart/) {
 # COSP simulator
 if ($cfg->get('cosp')) {
     add_default($nl, 'docosp', 'val'=>'.true.');
+
+    # following serves to always set a default when no matching value found in  namelist_defaults_eam.xml
+    my $val = undef;
+    $val = get_default_value('cosp_ncolumns');
+    $val = 50 if (! defined($val) );
+    add_default($nl, 'cosp_ncolumns','val'=>"$val");
 }
 
 
--- a/components/eam/bld/namelist_files/namelist_defaults_eam.xml
+++ b/components/eam/bld/namelist_files/namelist_defaults_eam.xml
@@ -1804,5 +1804,6 @@ with se_tstep, dt_remap_factor, dt_tracer_factor set to -1
 <so4_sz_thresh_icenuc phys="default"> 0.080e-6   </so4_sz_thresh_icenuc>
 <micro_mg_berg_eff_factor phys="default" clubb_sgs="1"> 0.7D0      </micro_mg_berg_eff_factor>
 <cldfrc_dp1  phys="default"> 0.018D0    </cldfrc_dp1>
+<cosp_ncolumns phys="default" hgrid="ne1024np4"> 10 </cosp_ncolumns>
 
 </namelist_defaults>
--- a/components/eam/bld/namelist_files/namelist_defaults_eam.xml
+++ b/components/eam/bld/namelist_files/namelist_defaults_eam.xml
@@ -1804,6 +1804,6 @@ with se_tstep, dt_remap_factor, dt_tracer_factor set to -1
 <so4_sz_thresh_icenuc phys="default"> 0.080e-6   </so4_sz_thresh_icenuc>
 <micro_mg_berg_eff_factor phys="default" clubb_sgs="1"> 0.7D0      </micro_mg_berg_eff_factor>
 <cldfrc_dp1  phys="default"> 0.018D0    </cldfrc_dp1>
-<cosp_ncolumns phys="default" hgrid="ne1024np4"> 10 </cosp_ncolumns>
+<cosp_ncolumns phys="default"> 10 </cosp_ncolumns>
 
 </namelist_defaults>


components/eam/bld/build-namelist
```diff
@@ -915,6 +915,12 @@ if ($rad_pkg eq 'rrtmg' or $chem =~ /waccm_mozart/) {
 # COSP simulator
 if ($cfg->get('cosp')) {
     add_default($nl, 'docosp', 'val'=>'.true.');
+
+    # following serves to always set a default when no matching value found in  namelist_defaults_eam.xml
```
brhillman, line 919: Is this extra code necessary? I actually think this can all be replaced by a single line in `namelist_defaults_eam.xml` that sets an overall default value (with no `hgrid` match) for `cosp_ncolumns`, and then overrides for `hgrid=ne1024np4`. I.e.,
```
<cosp_ncolumns                                 > 50 </cosp_ncolumns>
<cosp_ncolumns phys="default" hgrid="ne1024np4"> 10 </cosp_ncolumns>
```

components/eam/bld/namelist_files/namelist_defaults_eam.xml
```diff
@@ -1804,5 +1804,6 @@ with se_tstep, dt_remap_factor, dt_tracer_factor set to -1
 <so4_sz_thresh_icenuc phys="default"> 0.080e-6   </so4_sz_thresh_icenuc>
 <micro_mg_berg_eff_factor phys="default" clubb_sgs="1"> 0.7D0      </micro_mg_berg_eff_factor>
 <cldfrc_dp1  phys="default"> 0.018D0    </cldfrc_dp1>
+<cosp_ncolumns phys="default" hgrid="ne1024np4"> 10 </cosp_ncolumns>
```
PeterCaldwell, line 1807: I'd rather delete `hgrid="ne1024np4"` in the line above so we always use 10 subcolumns unless we make an explicit change. Otherwise I'm sure we're going to think we're doing the same thing at ne512 or 256 and wonder why we get different timings or results...

wlin7: Thanks for chiming in, @PeterCaldwell . The code changes largely to allow a resolution dependent of the parameter (and to preserve existing cosp tests). It makes senses at least for SCREAM simulations, we want it to be 10.

components/eam/bld/build-namelist
wlin7: Very true, @brhillman , adding an extra entry in namelist_defaults_eam.xml would have equivalent effect.  I tested both before implementing this version. We can simplify it if to use a grid-independent value of 10, as  Peter prefers.

No matter what, the extra call of add_default for cosp_ncolumns is always needed.

wlin7: Ben, if to remove explicit grid dependence, I would inline to keep the changes in build-namelist. The phys='default' will still be used, to differ from use of other potential physics package (such as MMF)

components/eam/bld/namelist_files/namelist_defaults_eam.xml
brhillman: @wlin7 actually no tests use COSP without `cosplite`, so this would not affect any existing tests. It appears that COSP is never used in E3SM without setting `cosp_ncolumns = 10`.

components/eam/bld/build-namelist
brhillman: Sorry, I should have been more clear. I think you're right that `add_default` is needed, but I don't think this part is:
```
    my $val = undef;
    $val = get_default_value('cosp_ncolumns');
    $val = 50 if (! defined($val) );
```
Most of the rest of the settings in `build-namelist` seem to just call `add_default`.

ndkeen (approved):
Approving, but I've not had a chance to follow the work or test -- trusting others

components/eam/bld/namelist_files/namelist_defaults_eam.xml
wlin7: Thanks for checking on the existing cost tests, @brhillman . I remove the explicit grid dependence.

Actually, in our SCREAM simulations so far, with cosp_lisccp_sim, there is no internal mechanism to reset the hardwired default of 50. If setting cosp_isccp = .true. instead, it would.

components/eam/bld/build-namelist
wlin7: These are still needed, unless we either remove the 'phys' modifider, or add another entry in namelist_defaults_eam.xml to cover all scenarios.

PeterCaldwell:
@brhillman and @wlin7 - any reason why I don't merge this now? I isn't passing on mappy because mappy is still not fully functional.

PeterCaldwell:
@bartgol and @jgfouca - this is another PR the v0 team needs which is failing on mappy but doesn't change stuff C++ even has access to. I suggest merging without worrying about testing on mappy.

jgfouca:
@PeterCaldwell , done

PR #1349, opened 2021-12-15
Fix bug in pblintd_height()

tcclevenger:
Bug with the `Kokkos::TeamThreadRange` range when upper and lower index were equal. Adds a varying `npbl` to tests using `pblintd_height()`.

### DIFF
--- a/components/scream/src/physics/shoc/shoc_pblintd_height_impl.hpp
+++ b/components/scream/src/physics/shoc/shoc_pblintd_height_impl.hpp
@@ -52,7 +52,7 @@ void Functions<S,D>::pblintd_height(
   const Int upper = nlev-1;
   const Int length = upper-lower;
   const Int lower_pack_indx = lower/Spack::n;
-  const Int upper_pack_indx = lower_pack_indx + (length+Spack::n-1)/Spack::n;
+  const Int upper_pack_indx = lower_pack_indx + (length+Spack::n)/Spack::n;
   Kokkos::parallel_reduce(Kokkos::TeamThreadRange(team, lower_pack_indx, upper_pack_indx),
                           [&] (const Int& k, Int& local_max) {
     auto indices_pack = ekat::range<IntSmallPack>(k*Spack::n);
--- a/components/scream/src/physics/shoc/shoc_pblintd_height_impl.hpp
+++ b/components/scream/src/physics/shoc/shoc_pblintd_height_impl.hpp
@@ -45,15 +45,19 @@ void Functions<S,D>::pblintd_height(
   const auto s_thv  = ekat::scalarize(thv);
   const auto s_rino = ekat::scalarize(rino);
 
-  // Compute rino values and find max index s.t. rino(k) >= ricr
-  Int max_indx = Kokkos::reduction_identity<Int>::max();
+  // Compute range for reduction. Only run the
+  // reduction if there is a valid range
+  const Int lower_indx = nlev-npbl;
+  const Int upper_indx = nlev-1;
+  if (lower_indx >= upper_indx) {
+    return;
+  }
+  const Int lower_pack_indx = lower_indx/Spack::n;
+  const Int upper_pack_indx = upper_indx/Spack::n;
 
-  const Int lower = nlev-npbl;
-  const Int upper = nlev-1;
-  const Int length = upper-lower;
-  const Int lower_pack_indx = lower/Spack::n;
-  const Int upper_pack_indx = lower_pack_indx + (length+Spack::n)/Spack::n;
-  Kokkos::parallel_reduce(Kokkos::TeamThreadRange(team, lower_pack_indx, upper_pack_indx),
+  // Compute rino values and find max index k s.t. rino(k) >= ricr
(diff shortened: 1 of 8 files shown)


bartgol (approved):

components/scream/src/physics/shoc/shoc_pblintd_height_impl.hpp
```diff
@@ -61,33 +64,32 @@ void Functions<S,D>::pblintd_height(
...
-    indices_pack.set(rino(k)<ricr || !in_range, -1);
+    indices_pack.set(rino(k)<ricr || !in_range, Kokkos::reduction_identity<Int>::max());
 
     // Local max index s.t. rino(k)>=ricr and k in [nlev-npbl, nlev-1)
     if (local_max < ekat::max(indices_pack))
       local_max = ekat::max(indices_pack);
 
   }, Kokkos::Max<Int>(max_indx));
 
+  std::cout << max_indx << std::endl;
```
bartgol, line 83: This should be removed (not desired, and won't work on gpu).

bartgol (changes requested):

components/scream/src/physics/shoc/shoc_pblintd_height_impl.hpp
tcclevenger: Looks like I missed one..

bartgol (approved):

tcclevenger:
Investigating the fail on Blake. Interestingly `./shoc_tests` fails in `pblintd_bfb` and `pblintd_height_bfb` every time I manually run it, but running `./shoc_tests pblintd_bfb` and `./shoc_tests pblintd_height_bfb` pass every time I manually run it..

tcclevenger:
> Investigating the fail on Blake. Interestingly `./shoc_tests` fails in `pblintd_bfb` and `pblintd_height_bfb` every time I manually run it, but running `./shoc_tests pblintd_bfb` and `./shoc_tests pblintd_height_bfb` pass every time I manually run it..

It is even triggered running without OpenMP.

tcclevenger:
Simple indexing issue. Should pass now.

components/scream/src/physics/shoc/shoc_pblintd_height_impl.hpp
```diff
@@ -52,7 +52,7 @@ void Functions<S,D>::pblintd_height(
   const Int upper = nlev-1;
   const Int length = upper-lower;
   const Int lower_pack_indx = lower/Spack::n;
-  const Int upper_pack_indx = lower_pack_indx + (length+Spack::n-1)/Spack::n;
+  const Int upper_pack_indx = lower_pack_indx + (length+Spack::n)/Spack::n;
```
bartgol, line 55: Doesn't this make it so that there's always at least one iteration? The interval [L,U) , with U>=L+1, always runs at least once...

bartgol: But maybe that's the point? Is upper supposed to be "the last" level to be run on?

tcclevenger: Ahh yeah it does. Ok, so there has to be a way to avoid that. But the previous index was one short (I think always). I mean, we could add a check if `upper <= lower`.

tcclevenger: @bartgol So I think the original issue was not that we were executing the `parallel_reduce()` when we shouldn't, but that when we did execute and never found an `if (inrange.any())`, then `max_indx = Kokkos::recuct_id<Int>::max()` was returned instead of the `-1` we were comparing against. I think we could run the `parallel_reduce` over all levels and still get the correct answer since we have the `Mask` checking for `in_range.any()`. 
As long as we use ` Kokkos::recuct_id<Int>::max()`

I think the other issue is unrelated, so you could test this out in your restart branch.

tcclevenger: So it turns out the new indexing was still wrong in cases with 16 pack size. I reverted back to the old way, but added an `if (u==l) return` so that we don't reduce when there is nothing to reduce. I do still believe that is ok to do, but might as well skip the work.

bartgol: So, you want to run the reduction on all levels, regardless? That might be ok on GPU, but it might be suboptimal on CPU. That said, correctness is the first concern, so whatever gets the correct results, can be merged, and performance be tackled with a follow up PR.

bartgol (approved):

components/scream/src/physics/shoc/shoc_pblintd_height_impl.hpp
bartgol: nvm, I looked at the code,, and it seems we do the ||reduce only on the packs we need, so it's fine. Thanks!

bartgol:
Manually running tests on mappy now.

bartgol:
All tests passing on mappy, I'll merge.

PR #1350, opened 2021-12-16
P3 F90 diagnostic bug fix (fixes chronically failing DP-SCREAM Mappy tests)

bogensch:
DP-SCREAM F90 CIME tests (ERP and ERS) have been chronically failing on Mappy with gnu compiler.  This PR fixes the underlying issue which was related to an output error for ANRAIN, AQRAIN, and FREQR diagnostics (these were the only three variables that had differences in the said ERP and ERS test fails).

Lines 1317-1319 are the culprits for the fails where indices bounds needs to be added to match the "where" statement search.  Thus, the fails were NOT DP specific but indicative of a general output bug with P3.  The DP tests were only catching them because DP tests output higher frequency *eam.h0* files, whereas the standard LR and HR tests never output *eam.h0* files.  

While not culprits of the fails, lines 1359-1369 fixes and inconsistency error where "psetcols" was being used in the outfld call.  "psetcols" should only be used if the code is enabled for sub-column use (which P3 implementation is not).  In our case, psetcols = pcols (as far as I can tell), but is a consistency fix.

Finally, the DP tests have been modified to use SPA (instead of the old prescribed aerosol model), to be consistent with SCREAMv0.1/v1.0

### DIFF
--- a/cime_config/machines/config_machines.xml
+++ b/cime_config/machines/config_machines.xml
@@ -658,12 +658,12 @@
       </arguments>
     </mpirun>
     <module_system type="module" allow_error="false">
-      <init_path lang="python">/usr/share/Modules/init/python.py</init_path>
-      <init_path lang="perl">/usr/share/Modules/init/perl.pm</init_path>
-      <init_path lang="sh">/usr/share/Modules/init/sh</init_path>
-      <init_path lang="csh">/usr/share/Modules/init/csh</init_path>
-      <cmd_path lang="python">/usr/bin/modulecmd python</cmd_path>
-      <cmd_path lang="perl">/usr/bin/modulecmd perl</cmd_path>
+      <init_path lang="python">/projects/sems/install/rhel7-x86_64/sems/v2/lmod/lmod/8.3/gcc/10.1.0/zbzzu7k/lmod/lmod/init/env_modules_python.py</init_path>
+      <init_path lang="perl">/projects/sems/install/rhel7-x86_64/sems/v2/lmod/lmod/8.3/gcc/10.1.0/zbzzu7k/lmod/lmod/init/perl</init_path>
+      <init_path lang="sh">/projects/sems/install/rhel7-x86_64/sems/v2/lmod/lmod/8.3/gcc/10.1.0/zbzzu7k/lmod/lmod/init/sh</init_path>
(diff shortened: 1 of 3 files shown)


AaronDonahue (approved):

AaronDonahue:
@bartgol or @jgfouca , if you have a moment would you mind manually testing this PR on mappy? the FSCREAM-Compset tests since those are the ones that should be impacted.  If it passes I think we can merge.

jgfouca:
@AaronDonahue , I can try. I will have to set up an intermediary repository somewhere in sandia because mappy cannot pull from github.

jgfouca:
@AaronDonahue , wow, all the submodules need to be proxied too :/

AaronDonahue:
> @AaronDonahue , wow, all the submodules need to be proxied too :/

Ugggg, what a nightmare.

jgfouca:
@AaronDonahue , I was able to proxy everything and get a test going. I had to add a commit to this PR to get things to get modules to work again on mappy. I'm at 8/13, all 8 PASS.

jgfouca:
OK, run complete. I'm not going to sweat the DIFFs since we haven't run or updates baselines in a while on mappy. The two fails:

```
FAIL ERP_R_Ln10.ne4_ne4.FDPSCREAMA97.mappy_gnu 
FAIL SMS_R_Ln10.ne4_ne4.FDPSCREAMA97

Reason:
 getMetaSchedule: tmpP:            1           1
 initializing elements...
 ERROR: CHKRC
 ERROR timeaddmonths():  MM out of range
#0  0x127059a in ???
#1  0x127075d in ???
#2  0x539a77 in ???
#3  0x5da936 in ???
#4  0x5e4870 in ???
#5  0x5cff57 in ???
#6  0x7017a7 in ???
#7  0x4c0898 in ???
#8  0x4b9dc0 in ???
#9  0x42fbd4 in ???
#10  0x4203dd in ???
#11  0x411451 in ???
#12  0x7f24b6de9554 in ???
#13  0x411bd8 in ???
#14  0xffffffffffffffff in ???
```

jgfouca:
Forgot to mention @bogensch , ^^^

jgfouca:
Interesting, when I run the tests standalone, they seem to work? Let me test things a bit more.

jgfouca:
OK, everything passed but with DIFFs. I will merge.

PR #1351, opened 2021-12-17
SCREAM: update sems modules and flags for mappy

bartgol:
SEMS modules changed on mappy, and gcc 10 needs some extra flags to work around some stricter checks that MPI would fail.

Internal scream tests pass on mappy, but I'm still trying to figure out CIME testing, given the compiler changes.


bartgol:
Update: @jgfouca is on it, talking with SNL IT folks constantly, to see what's the best course of action. Turns out the issues are deeper than I thought, and they can't be fixed with a simple module switch.

jgfouca:
We will want this merged at some point (when CIME and the rest of E3SM is ready for the new system).

jgfouca:
Closing for now. We could be waiting a while for SEMS TPLs v2 to meet E3SM's needs.

PR #1354, opened 2021-12-22
Move field property check outside the Field class

bartgol:
This is part of a series of PRs that I am planning to do in order to simplify the Field class stack. These PRs are meant to simplify the class usage (from a development point of view).

This PR moves the storage of FieldPropertyCheck (FPC) outside the Field class, and inside the AtmosphereProcess class. Each class instance can request FPC's for its own in/out fields (or groups), and the base class will run those checks inside the `run` method, before and/or after the call to `run_impl` (depending on the FPC request specs).

Edit: also, the user can disable property checks for input/output fields on a per-atm-proc basis, via yaml file, by setting `Enable Input Fields Checks: false` or `Enable Output Fields Checks: false` respectively, inside each atm proc yaml file section. E.g:
```
  Process 0:
    Name: Homme
    Enable Output Fields Checks: true  # true is the default anyways
    ...
  Process 1:
    Name: SHOC
    Enable Output Fields Checks: false
```
Homme will run all the checks that were created for the output fields (if any), while SHOC would not run any check.

NOTE: I have not added the possibility to create additional checks via yaml file. I don't think it would be hard though.


components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
```diff
@@ -315,6 +315,25 @@ void HommeDynamics::initialize_impl ()
 
   // Complete homme model initialization
   prim_init_model_f90 ();
+
```
jeff-cohere, line 318: What do you think about putting these inside a `add_property_checks` method, as you've done for all the above operations? That might make for an easier setting in which someone can digest the assumptions.

One could argue that we could provide an `add_property_checks` method as part of the `AtmosphereProcess` interface to encourage people to think about it in this way, but I don't know if that's necessary.

jeff-cohere: Yeah, it's definitely not needed, with the `add_property_check` method already there. I like this reorganization, Luca.

components/scream/src/share/atm_process/atmosphere_process.hpp
```diff
@@ -362,11 +357,51 @@ class AtmosphereProcess : public ekat::enable_shared_from_this<AtmosphereProcess
...
+
   // This process's copy of the timestamp, which is set on initialization and
   // updated during stepping.
   TimeStamp m_time_stamp;
 };
 
+// ====================== IMPLEMENTATION ======================= //
+
+template<RequestType RT, template<typename> class FPC, typename... Args>
+void AtmosphereProcess::add_property_check (const FieldIdentifier& fid, const Args... args) {
```
jeff-cohere, line 373: What do you think about making this function idempotent? In other words: if you try to add a property check that's already been added this way, the method does nothing. Might make things easier to analyze.

jeff-cohere (approved):
This all looks good. I've left some comments for your consideration.

components/scream/src/share/atm_process/atmosphere_process.hpp
bartgol: I was indeed trying to work on a way to ensure that the same FPC (**for the same field**) is not added twice. It would require a couple more maps to be scanned, but I opted for the easy (lazy?) way, and didn't do it.

To be fair, though, now that the FPC's are added/run inside each Atm Proc separately, I find it harder to have duplicates. It would probably only happen in case of copy+paste, when a dev is trying to add a FPC for a bunch of fields, but forgets to change the name of the field in one of them.

bartgol:
@jeff-cohere I had to rebase on master, so this will need another approval (nothing changed though).

jeff-cohere (approved):
Happy new year!

jeff-cohere:
Looks like Weaver is unhappy with these tests (for both debug and release builds):

```
36:homme_shoc_cld_p3_rrtmgp_ut_np1_omp1
37:model_baseline
38:model_initial
39:model_restarted
40:model_restart_check
41:homme_shoc_cld_spa_p3_rrtmgp_ut_np1_omp1
```

bartgol:
Yeah, I was looking at the fails. Apparently, the lower bound check for tracers after Homme is failing. The tracers _could_ end up with tiny negative values, which were expected, and to be clipped to 0, while "large" negative values should be a true error. So we check that Q>-1e-20, before repairing to 0. However, the min is now at -8e-18, so the check fails. I'm not sure why this happens now, and not before. All this PR did was moving the place that stores the checks.

jeff-cohere:
Could the minimum (negative) tracer values have changed as the result of a PR that we picked up in the rebase?

AaronDonahue:
> Looks like Weaver is unhappy with these tests (for both debug and release builds):
> 
> ```
> 36:homme_shoc_cld_p3_rrtmgp_ut_np1_omp1
> 37:model_baseline
> 38:model_initial
> 39:model_restarted
> 40:model_restart_check
> 41:homme_shoc_cld_spa_p3_rrtmgp_ut_np1_omp1
> ```

@jeff-cohere and @bartgol , it looks like some of these tests are failing in master.  So it is possible we picked something up in the rebase.  Probably want to fix `master` first before worrying about what went wrong in this PR.

bartgol:
Yeah, discarding a return value is an easy bug to introduce in the code.

bartgol:
Ugh, it turns out that the tracers get quite bad on GPU. At the end of the 2nd iteration, homme's check for tracers craps out, with quite large (in magnitude) negative values.

First iteration (all good, gets repaired to 0):
```
 Bounds Interval: [-1e-17,1.79769e+308]
 Field  Interval: [-2.30756e-20,3.5961e+07]
```
Second iteration:
```
 Bounds Interval: [-1e-17,1.79769e+308]
 Field  Interval: [-6.52592e-09,3.06056e+07]
```
The field interval is consistent between runs, but it points to some issues.

It's worth noticing that this is likely a bug that has been existing in our code for a while, it simply got unnoticed due to the field check return value being previously discarded.

bartgol:
@AaronDonahue @jeff-cohere The tests that are still failing are not, per se, due to this PR. The errors were already in master, this PR simply exposed them. However, fixing the errors might take time, and it would definitely be something orthogonal to this PR.

Given that this PR adds flexibility for turning on/off checks within each atm proc, I would like to disable the field checks (at least in homme), and merge this PR. Would you be ok with it?

jeff-cohere:
Sounds good to me!

components/scream/src/share/atm_process/atmosphere_process.cpp
```diff
@@ -20,14 +20,18 @@ void AtmosphereProcess::initialize (const TimeStamp& t0, const RunType run_type)
 }
 
 void AtmosphereProcess::run (const int dt) {
-  // Make sure required fields are valid
-  check_required_fields();
+  if (m_params.get("Enable Input Fields Checks", true)) {
```
jeff-cohere, line 23: This is super nitpicky, but `"Enable Input Field Checks"` might be a slightly better way to express this option.

components/scream/src/share/atm_process/atmosphere_process.cpp
```diff
@@ -20,14 +20,18 @@ void AtmosphereProcess::initialize (const TimeStamp& t0, const RunType run_type)
...
+    // Run any check on required fields that has been stored in this AP
+    check_required_fields();
+  }
 
   // Let the derived class do the actual run
   run_impl(dt);
 
-  // Make sure computed fields are valid
-  check_computed_fields();
+  if (m_params.get("Enable Output Fields Checks", true)) {
```
jeff-cohere, line 31: Same here--just change `Fields` -> `Field`

jeff-cohere (approved):
Looks great! I left a very minor suggestion for a couple of configuration setting names.

bartgol:
Mappy is having some issues with ssh. I manually tested the branch, and tests pass. Given the approval by Jeff, and the fact that tests pass on blake and weaver, I'm gonna manually merge to get things going.

PR #1356, opened 2021-12-23
Remove temporary override of tensor HV setting for ne4

brhillman:
Remove temporary override of namelist defaults to set tensor hyperviscosity for ne4. This override appears to have been put in 11 months ago, with the note that it was a temporary fix for an upstream merge, but I think we ultimately want to use tensor HV, and this is one area that makes EAM compsets non-BFB between the E3SM and SCREAM repos. If we make this change, then the only other thing that appears to be non-BFB between E3SM and SCREAM repos for F-cases is the setting of `adjust_ps`.

[non-B4B] (for v0 tests, should be B4B for v1 tests)

### DIFF
--- a/components/eam/bld/namelist_files/namelist_defaults_eam.xml
+++ b/components/eam/bld/namelist_files/namelist_defaults_eam.xml
@@ -1363,13 +1363,6 @@
 <hypervis_order >     2 </hypervis_order>
 <hypervis_scaling> 3.0 </hypervis_scaling>
 
-<!-- special TEMPORARY defaults for scream for upstream merge  -->
-<nu hgrid="ne4np4"> 5.0e17  </nu>
-<hypervis_scaling hgrid="ne4np4" > 0 </hypervis_scaling>
-
-
-
-
 <!-- -1.0 indicates use the value for nu -->
 <nu_q> -1.0 </nu_q>
 <nu_p> -1.0 </nu_p>


mt5555 (approved):

oksanaguba (approved):

bartgol:
@brhillman I think the issue with this PR is that it is not up to date with master (like 1381). Given how small this PR is, and how sure we are that it should not impact v1, I'm inclined to use the force, and merge it. Is that ok with you?

bartgol (approved):
Approving the rebase, so we get a non-brute-force merge.

brhillman:
@bartgol I don’t understand why the mappy CIME tests keep failing. It looks like a problem checking out submodules from the AT log above, but this PR doesn’t seem to touch anything related to this.

bartgol:
> @bartgol I don’t understand why the mappy CIME tests keep failing. It looks like a problem checking out submodules from the AT log above, but this PR doesn’t seem to touch anything related to this.

The reason why it needs to check out submodules is that the AT tries to merge origin/master into the branch. `test-all-scream` does that too, but it restores the initial git sha upon completing, so before running the CIME tests we need to re-do the merge.

I am not sure why this causes errors though. @jgfouca , thoughts?

jgfouca:
@bartgol , @brhillman , I see the problem. Fix coming

jgfouca:
@brhillman , I upstreamed the fix into this branch.

AaronDonahue:
This PR is expected to result in DIFFs for the SCREAM-HR and SCREAM-LR tests, which is what the AT reports.  I'm going to manually merge.

PR #1360, opened 2021-12-23
Make adjust_ps consistent with SCREAM

brhillman:
Make setting of adjust_ps consistent with SCREAM repo. We want to allow
adjust_ps=.false. for SCREAM, but retain adjust_ps=.true. for E3SM to
remain BFB for v2. This change was originally made downstream in SCREAM,
but causes
conflicts between the two repos, and prevents EAM compsets from being
BFB between the E3SM and SCREAM repos. Moving this change upstream,
protected in an ifdef, allows the two repos to be consistent, and keeps
EAM compsets BFB between the SCREAM and E3SM repos. This also sets the
precedent for allowing changes upstream in E3SM that should apply only
to SCREAM configurations, protected in ifdef SCREAM. To enable, we
simply need to add the -cppdefs '-DSCREAM' flag to CAM_CONFIG_OPTS
for SCREAM compsets.

[BFB]


brhillman:
Note: this applies the identical commit made upstream in the E3SM repo, created from the most recent common ancestor of the two repos, to avoid merge conflicts in our next upstream merge. This PR should NOT go in until after #1359 goes in.

AaronDonahue (approved):

AaronDonahue (approved):

PR #1362, opened 2022-01-04
Move AD WSM setup to initialization

tcclevenger:
The WSM is being constructed inside run_impl() of both shoc and p3 AD. This PR move WSM setup into initialize_impl(). Also, for shoc, `shoc_init()` is moved to initialization.


tcclevenger:
The issue here is that when the WSM is used in back-to-back iterations, the internal counters are not reset and so, during the 2nd iteration, internally the WSM is complaining that data is already used when in fact we can freely overwrite the values used in the last iteration. https://github.com/E3SM-Project/EKAT/pull/176 adds a function to the WSM to reset these internal values.

bartgol (approved):

AaronDonahue:
It looks like this is approved and passed the AT.  Should we add the Automerge label and get it in?  @tcclevenger or @bartgol ?

tcclevenger:
Again, Thanks @AaronDonahue!

PR #1363, opened 2022-01-04
Mappy setup was not safe to call if already called

jgfouca:
The purge makes it recallable.

This should fix the scripts-test fails on the dashboard.


bartgol (approved):
I'm not sure why the purge fixes things. In particular, I'm not sure why loading sems-env causes the load error for the downstream packages. It's almost like loading sems-env causes a re-load of all packages, but in the wrong order... But whatever, this fix is perfectly fine.

PR #1364, opened 2022-01-06
New scripts to help EKAT transition to better test system

jgfouca:


bartgol (approved):

jgfouca:
Merging. The only that could possibly impact the AT testing is the more-robust way of getting the directory of the current script which I already tested.

Issue #1305, opened 2021-11-08
Change as many bool as possible to enums (where useful)


bartgol:
Bool's can be cryptic when found at the call place. Take the function
```
void allocate_vector (std::vector<int>& v, const int n, const bool init_to_random);
```
As the name says, if the last input is `true`, we init the entries to random values. Pretty clear. However, this is how the code looks at the call site:
```
std::vector<int> vec;
allocate_vector(vec,10,true);
```
Not so clear anymore. The issue can be overcome by inserting inline comments, such as
```
allocate_vector(vec,10, /* init_to_random = */ true);
```
but I would argue that's a bit clunky. Compare that with
```
enum InitType {
  InitDefault,
  InitRandom
};
void allocate_vector(std::vector<int>& vec, int n, InitType);
...
std::vector<int> vec;
allocate_vector(vec,10,InitRandom);
```

It gets even better if you use strong enums:
```
enum class InitType {
  Default,
  Random
};
void allocate_vector(std::vector<int>& vec, int n, InitType);
void allocate_vector_bad(std::vector<int>& vec, int n, bool init_to_random);
...
std::vector<int> v;
allocate_vector (v,10,InitType::Random); // ok, works
allocate_vector (v,InitType::Random,10); // Compiler error
allocate_vector_bad(v,10,true);                // ok, works
allocate_vector_bad(v,true,10);                // ARG, this works too!
```
The last example might be silly, but I _did_ run into something similar with a very well developed linear algebra package, where a mistake at the call site ended up passing a pointer to a bool. Had that bool been a strong enum, the compiler would have barked, and the bug caught right away.

bartgol:
Possibly, also tackle #1179 .

PR #1370, opened 2022-01-14
Field class refactor

bartgol:
This PR removes template arguments from Field.

There have been instances where a change was not possible, due to the incompatibility of Field<const T> with Field<T>. While that's generally healthy, in some cases we store all Field's in a vector/list of non-const fields, which prevents from adding const fields. An example is arising in another branch, where I am adding (limited) output capabilities to each atm proc (for debug purposes): in order to create an output object, I need a field manager, but I can't add const fields to the field manager (which stores all fields as non-const).

Another example is remappers: right now, registration of fields in remappers requires non-const fields (both for src and tgt). If an atm process needs to remap inputs (that are not also outputs), it will only have access to a const field, which cannot be used with remappers. The alternative would be to store 2 copies of each field, one as const, and one non-const, and offer more registration method: const_src/nonconst_tgt, nonconst_src/const_tgt, and nonconst_src/nonconst_tgt (both const makes no sense here), which makes the syntax heavier.

The big downside of this PR is that we need to query the field identifier for the data type, then call the corresponding templated method for the correct data type. To be fair, if we only assumed Real fields, then the overhead would be virtually non existent, and if we want to allow mixed precision, we would have to support some sort of similar scenario. To be clear, this is what one needs to do to call some function that needs a templated arg for the scalar type:
```
void my_func (const Field& f) {
   const auto& dt = f.get_header().get_identifier().data_type();
   if (dt=="float") {
      my_func_impl<float>(f);
   } else if (dt=="double") {
      my_func_impl<double>(f);
   } else if (dt=="int") {
      my_func_impl<int>(f);
   } else {
     // error out
   }
}

template<typename ST>
void my_func_impl (const Field& f) {
   const auto& fl = f.get_header().get_identifier().get_layout();
   switch (fl.rank()) {
      case 1:
       {
          auto v1d = f.get_view<ST*>();
          //do stuff on v1d
       }
       case 2:
       {
          auto v2d = f.get_view<ST**>();
          //do stuff on v2d
       }
       ...
   }
}
```

Edit: I thought about breaking the commit into multiple ones, but I didn't think it would have added much. Most of the files are simply replacing `Field<Real>` with `Field`, or breaking a header of a formerly templated class into header-cpp.


bartgol:
@jeff-cohere @tcclevenger This might look like a huge PR, but most of the file changes are stuff like `Field<Real>` -> `Field`. The bulk of the changes are in the `share/field` folder.

jeff-cohere:
Hey Luca,

Is there any reason not to add direct accessors to the `Field` for more commonly used metadata, like `data_type` and `rank`? It feels to me in the above code like we are asking everyone to become intimately familiar with the workings of the `Field` and its headers/identifiers just to understand basic field properties.

This kind of thing is always a bit controversial because of the many interpretations of "don't repeat yourself", but I find a little redundancy can make it easier for people to do the right thing and harder to do the wrong thing, and to me that is the essence of good design.

bartgol:
> Hey Luca,
> 
> Is there any reason not to add direct accessors to the `Field` for more commonly used metadata, like `data_type` and `rank`? It feels to me in the above code like we are asking everyone to become intimately familiar with the workings of the `Field` and its headers/identifiers just to understand basic field properties.
> 
> This kind of thing is always a bit controversial because of the many interpretations of "don't repeat yourself", but I find a little redundancy can make it easier for people to do the right thing and harder to do the wrong thing, and to me that is the essence of good design.

I think it can go either way. As you said, I didn't do it to avoid code repetition, but we can definitely add some common metadata getter shortcuts.And in fact, data type and rank are somewhat needed in order to do a correct reshape of the stored array. I guess the name of the field is also a very common thing we query for.

I can do it.

components/scream/src/share/field/field_property_checks/field_lower_bound_check.hpp
```diff
@@ -26,14 +21,12 @@ class FieldLowerBoundCheck: public FieldWithinIntervalCheck<RealType> {
   // can repair fields that fail the check by overwriting nonpositive values
   // with the given lower bound. If can_repair is false, the check cannot
   // apply repairs to the field.
-  explicit FieldLowerBoundCheck (const_RT lower_bound,
+  explicit FieldLowerBoundCheck (const double lower_bound,
```
jeff-cohere, line 24: Any reason not to use `Real` here?

components/scream/src/share/field/field_property_checks/field_positivity_check.hpp
```diff
@@ -1,196 +1,26 @@
...
   // Default constructor -- cannot repair fields that fail the check.
-  FieldPositivityCheck () : m_lower_bound(-1) {}
-
-  // Constructor with lower bound -- can repair fields that fail the check
-  // by overwriting nonpositive values with the given lower bound.
-  explicit FieldPositivityCheck (const_RT lower_bound) :
-    m_lower_bound(lower_bound) {
-    EKAT_ASSERT_MSG(lower_bound >= 0, "lower_bound must be positive.");
+  explicit FieldPositivityCheck (const bool can_repair = true) :
+    FieldLowerBoundCheck (std::numeric_limits<double>::epsilon(),can_repair)
```
jeff-cohere, line 17: Consider `Real` instead of `double` here (unless we are sticking with double precision)

jeff-cohere:
I like how clean the Field code is without the template parameter.

One more question. Instead of storing the data type as a string, why not use an enum? We already have to query a set of supported data types, so adding support for more already requires code changes. And string comparisons aren't going to kill us, but this is good stuff for enums, right?

components/scream/src/share/field/field_property_checks/field_positivity_check.hpp
bartgol: As I said in the other comment, the FPC could be used on a double field (even if Real=float). For that case, a lower bound of `std::numeric_limits<Real>::epsilon()` would be too large.

bartgol:
> I like how clean the Field code is without the template parameter.
> 
> One more question. Instead of storing the data type as a string, why not use an enum? We already have a TypeMap that lists supported data types, so adding support for more already requires code changes. And string comparisons aren't going to kill us, but this is good stuff for enums, right?

That's a good point. Enums are usually the way to go. I think I picked strings because I was afraid of clashing with actual type names; I didn't find satisfactory names for the enum entries. Obviously, `int` doesn't work. `Int` is an alias to int. I thought about adding `_t`, but is misleading (in the std lib it hints to a type). Maybe add `int_v, float_v, double_v` or `int_idx, float_idx, double_idx`? I'm not sure. None seems appealing to me. If you have some good ideas, I can implement it real quick though.

components/scream/src/share/field/field_property_checks/field_lower_bound_check.hpp
jeff-cohere: That makes sense. No reason to use Real here unless we mind the implicit cast from Real to double in the constructor.

jeff-cohere:
I guess this raises the issue of whether we have a naming convention for enumerated type fields. Google's style guide treats them like constants and puts a `k` prefix in front of them. We could use something like `int_type` or `real_type`, etc, I suppose. Playing with "dt" for data type seems dangerous, because dt usually means "the length of the time step."

I don't know. Naming is hard! If I think of something, I'll let you know.

bartgol:
> I guess this raises the issue of whether we have a naming convention for enumerated type fields. Google's style guide treats them like constants and puts a `k` prefix in front of them. We could use something like `int_type` or `real_type`, etc, I suppose. Playing with "dt" for data type seems dangerous, because dt usually means "the length of the time step."
> 
> I don't know. Naming is hard! If I think of something, I'll let you know.

Yeah, `dt` wasn't the best of choices. I thought there was no confusion, since it was a local var in a context where there's no dt (as in time step) involved, but it's definitely not pretty.

I'll keep the PR on hold, and give us a day to think of enum names. If we don't find something satisfactory, I'll merge, and piggy back the data type enum task on #1305 .

components/scream/src/share/scream_types.hpp
```diff
@@ -37,6 +37,14 @@ enum class RunType {
   Restarted
 };
 
+// An enum for specifying fields data type
+enum class DataType {
```
jeff-cohere, line 41: @bartgol, what's this? Is this left over from your original `_t` enum type idea? Seems like we could start from this and change if we think of something better.

bartgol: Yep, that's it. I forgot about this. I am not fond of `_t` since my std-mind set parses that as a typename, while this is an enum value. Some other options:

- `I`, `F`, `D`: maybe not clear enough
- `int_type`, `float_type`, `double_type`: not too different from `_t` (`blah_type` is often a typename, not a value), but maybe a bit better
- `int_data`, `float_data`, `double_data`: like the above, but less ambiguity with typenames.
- `type_int`, `type_float`, `type_double`: flipping the words seems less confusing?
- `IntType`, `FloatType`, `DoubleType`: no conflicts with existing typenames.

Maybe the last one is my favorite. But there may be better options.

jeff-cohere: I like the last one too. Why don't we go with that? If we need to further "scope" it, we can always make it an anonymous enum within the `Field` class itself (but no need this PR).

bartgol: Being an enum class, I think it's already scoped enough.

bartgol:
@jeff-cohere I implemented the enum for the data type. Let me know what you think.

jeff-cohere (approved):
Much better! Thanks, Luca.

tcclevenger (approved):

AaronDonahue:
@bartgol , it looks like this PR is nearly there except for the `field_ut` test failing on all three AT machines.

bartgol:
Yeah, I'm taking a look at it now.

bartgol:
Ok, this should be working now.

jeff-cohere (approved):

PR #1371, opened 2022-01-18
Add the possibility to subcycle any atm process

bartgol:
The user simply has to specify 'Number of Subcycles' in the parameter list section of the atm proc to subcycle. The specified number must divide the time step exactly.

All checks for in/out fields are performed at the end of the subcycling. I assumed this is ok. If you think it's desirable to check in and/or out fields within each subcycling iteration, I can add that capability relatively easily, I think. (Edit: see comment at the bottom)

Example: to subcycle P3-cld-fraction-SHOC within the homme-p3-cld-shoc-rrtmgp test, say, 3 times, one needs to have in the input yaml file something like this
```
Atmosphere Processes:
  Number of Entries: 3
  Schedule Type: Sequential
  Process 0:
    Process Name: Homme
  Process 1:
    Process Name: Group
    Number of Entries: 3
    Schedule Type: Sequential
    Number of Subcycles: 3
    Process 0:
      Process Name: SHOC
      Grid: Physics GLL 
    Process 1:
      Process Name: CldFraction
      Grid: Physics GLL 
    Process 2:
      Process Name: P3
      Grid: Physics GLL 
  Process 2:
      Process Name: RRTMGP
      Grid: Physics GLL
      active_gases: ...
```
If `Number of Subcycles` is set to 1 in the list above, the internal grouping of P3-cld-SHOC is irrelevant, and is equivalent to have the outer group of size 5 (as it is done now in the ad coupled tests).

Edit: notice that, adding pre/post checks in the options of any of the inner processes, would run the checks at each subcycle iteration. If you need/want checks only at the beginning/end of the subcycling, you need to request the checks inside the group section. E.g.,
```
    Process 1:
       Process Name: Group
       Enable Output Field Checks: true
       ...
```
will run checks on all the output of the P3-cld-SHOC group, at the end of the subcycles.

### DIFF
--- a/components/scream/src/share/atm_process/atmosphere_process.cpp
+++ b/components/scream/src/share/atm_process/atmosphere_process.cpp
@@ -4,6 +4,7 @@
 
 #include <set>
 #include <stdexcept>
+#include <string>
 
 namespace scream
 {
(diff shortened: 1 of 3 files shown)


AaronDonahue (approved):
Nice, clean and simple implementation.  Thanks @bartgol

tcclevenger (approved):

PR #1373, opened 2022-01-20
Scale Faxa_rainl in SC::do_export

tcclevenger:
To match EAM, we should be scaling `precip_liq_surf` by `RHO_H2O = 1000` when exporting to `Faxa_rainl`.

I chose to treat `Faxa_rainl` as a computed field, which is `1000*precip_liq_surf`. Another solution could be to store a view of constant multiples and apply that directly during export (much like the new "sign switching" import stuff works with -1.

Selecting "Hide whitespace" when looking at the files makes seeing the changes easier.

### DIFF
--- a/components/scream/src/control/surface_coupling.cpp
+++ b/components/scream/src/control/surface_coupling.cpp
@@ -51,10 +51,11 @@ set_num_fields (const int num_cpl_imports, const int num_scream_imports,
   z_mid = decltype(z_mid) ("", m_num_cols, m_num_levs);
 
   // These fields contain export data
-  Sa_z         = decltype(Sa_z)      ("", m_num_cols);
-  Sa_ptem      = decltype(Sa_ptem)   ("", m_num_cols);
-  Sa_dens      = decltype(Sa_dens)   ("", m_num_cols);
-  zero_view    = decltype(zero_view) ("", m_num_cols);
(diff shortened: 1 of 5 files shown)


AaronDonahue (approved):
Thanks for getting this so quickly Conrad.  Looks good to me.

bartgol (approved):

AaronDonahue:
We expect this PR to be non-BFB for the F2000SCREAMv1 compset, so once the AT returns a DIFF for this test we can manually merge.  Provided everything else passes.

AaronDonahue:
@bartgol or @jgfouca , According to the AT this PR passed Blake the first time it was tested and then failed the second time.  According to the FAIL message in the second it looks like the build timed out, so a not a fail due to the PR right?

The only other fail is the expected DIFF on mappy in the SCREAMv1 F-Compset test.  If we agree the blake fail is a one-off thing then I can manually merge this PR.

jgfouca:
@AaronDonahue you can manually merge if you are confident things would have worked on Blake were it not for the timeout. It would probably be better to just request an AT retest.

AaronDonahue:
The RETEST returned all passing except the expected DIFF.  I'll manually merge now.

tcclevenger:
@AaronDonahue Sorry, I was off Friday. thanks for retesting and checking the mappy fail!

PR #1378, opened 2022-01-24
Fix scream v1 compiler flags when building from CIME

jgfouca:
We do expect CIME to have the correct CMAKE_BUILD_TYPE, but,
unlike other components, we do not need to clear CMAKE*FLAGS
before configuring scream because scream does not rely on
the CIME build system to manage its flags.

This problem got noticed when scream developers noticed that, in their debug CIME SCREAM V1 cases, none of the scream cxx files were getting built with `-g`.

### DIFF
--- a/components/CMakeLists.txt
+++ b/components/CMakeLists.txt
@@ -91,9 +91,24 @@ endif()
 
 project(E3SM C CXX Fortran)
 
-# We do want CMAKE_BUILD_TYPE to be set, but we do NOT want CMake to 
+# Include function definitions
+include(${CMAKE_CURRENT_SOURCE_DIR}/cmake/cmake_util.cmake)
+include(${CMAKE_CURRENT_SOURCE_DIR}/cmake/build_mpas_model.cmake)
+include(${CMAKE_CURRENT_SOURCE_DIR}/cmake/build_scream.cmake)
+include(${CMAKE_CURRENT_SOURCE_DIR}/cmake/build_model.cmake)
+
+set(CMAKE_VERBOSE_MAKEFILE TRUE)
+
+if(USE_CUDA)
+  enable_language(CUDA)
+endif()
+
+# Scream manages its own flags
+build_scream()
+
+# We do want CMAKE_BUILD_TYPE to be set, but we do NOT want CMake to
 # decide what optimization flags to append, based on build type,
-# so make all the following empty
+# for components who rely on CIME for build flags, so make all the following empty.
 set (CMAKE_C_FLAGS_RELEASE "")
 set (CMAKE_CXX_FLAGS_RELEASE "")
 set (CMAKE_Fortran_FLAGS_RELEASE "")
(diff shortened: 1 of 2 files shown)


components/CMakeLists.txt
```diff
@@ -102,25 +117,12 @@ set (CMAKE_C_FLAGS_DEBUG "")
...
-
-# Include function definitions
-include(${CMAKE_CURRENT_SOURCE_DIR}/cmake/cmake_util.cmake)
-include(${CMAKE_CURRENT_SOURCE_DIR}/cmake/build_mpas_model.cmake)
-include(${CMAKE_CURRENT_SOURCE_DIR}/cmake/build_scream.cmake)
-include(${CMAKE_CURRENT_SOURCE_DIR}/cmake/build_model.cmake)
-
 set(BUILDCONF ${CASEROOT}/Buildconf)
 
-# Do any MPAS/scream builds first since they manage their own flags
```
jgfouca, line 117: MPAS does not manage its own flags so this comment was wrong.

components/CMakeLists.txt
```diff
@@ -91,9 +91,24 @@ endif()
...
+set(CMAKE_VERBOSE_MAKEFILE TRUE)
+
+if(USE_CUDA)
+  enable_language(CUDA)
+endif()
+
+# Scream manages its own flags
+build_scream()
+
+# We do want CMAKE_BUILD_TYPE to be set, but we do NOT want CMake to
```
bartgol, line 109: Gets me wondering: what's the point of setting the build type, if we do not want cmake to use it? Maybe, we could set `CMAKE_<LANG>_<BUILD_TYPE>` inside `cime_config/machines/cmake_macros/$mach.cmake`?

components/CMakeLists.txt
```diff
@@ -102,25 +117,12 @@ set (CMAKE_C_FLAGS_DEBUG "")
...
-
 set(BUILDCONF ${CASEROOT}/Buildconf)
 
-# Do any MPAS/scream builds first since they manage their own flags
 build_mpas_models()
-build_scream()
 
 # Set global cmake settings
 set(CMAKE_MODULE_PATH ${CIMEROOT}/src/CMake)
-set(CMAKE_VERBOSE_MAKEFILE TRUE)
```
bartgol, line 123: Thank you for removing this. :)

components/scream/CMakeLists.txt
```diff
@@ -127,7 +127,7 @@ endif ()
 
 # Add CUDA as a language for CUDA builds
 if (CUDA_BUILD)
-    enable_language(CUDA)
+  enable_language(CUDA)
```
bartgol, line 130: Not for this PR, but... we don't actually use CUDA as a cmake language. That is, we don't have any source file, for which we tell CMake that the language is CUDA, and that CMAKE_CUDA_COMPILER needs to be used to build it...

bartgol: Imho, it would be a nice EKAT issue, to set compilers "the right way". That means, CMAKE_CXX_COMPILER=mpicxx, CMAKE_CUDA_COMPILER=nvcc. But I fear that the solution would be "set LANG=CUDA for all cxx files on GPU builds", since virtually all cxx files eventually include some kokkos header...

bartgol: Being very cmake-style, we should actually set `CMAKE_CXX_COMPILER=g++`, and then call `find_package(MPI)`, and link against MPI.

bartgol: Sorry for the long thread, I just had a stream of thoughts. I feel like all of these changes to make scream "very cmake compliant" should be tried in EKAT at some point, to experiment.

bartgol (approved):
Sneaky bug! I have some comments, but mostly just discussion topics...

components/CMakeLists.txt
jgfouca: @bartgol , we do want CMake to use it for SCREAM.

components/CMakeLists.txt
jgfouca: Not removed, just moved farther up the file so that SCREAM builds are verbose and recorded in e3sm.bldlog like all the other components.

components/CMakeLists.txt
bartgol: True. I do think, however, that the whole e3sm should make use of the build type. The `$mach.cmake` files seem like an appropriate way to add default debug/release flags (mostly, optimization level and/or debug flag). Users should still be able to a) override optimization levels, and b) append additional flags.

components/CMakeLists.txt
bartgol: Ah. Oh well. =)

components/CMakeLists.txt
jgfouca: @bartgol I agree, there is a lot of legacy stuff built in to CIME that predates the cmake effort, so there's a lot of non standard stuff that still needs to be cmake-ified. This is on my longterm todo list.

jgfouca:
Merging. I'm not worried about the weaver timeout. Nothing in this PR could be responsible for that.

PR #1380, opened 2022-01-26
Bootstrap phi_int from phis in Homme, for initial runs

bartgol:
This PR aims at removing the need for `phi_int` in our IC files. With these changes, SCREAM now does what EAM does: read in phis, and compute phi using Homme's EOS.

@tcclevenger I had to do some changes to the PD remapper. After the Field PR, I ran into the first use case for the de-templetization of the FIeld class. Namely, Homme does not have a non-cons version of phis, but needs to remap it to the physics grid, to compute phi via vertical integral. After the field refactor, this is possible, but the remapper needed to handle the case of a read-only field, which meant that it could not extract a `Real*` from the FIeld (only a `const Real*`).


components/scream/scripts/populate_nc_file.py
```diff
@@ -391,6 +396,69 @@ def remove_variables(self):
         if len(self._rvars)>0:
             Nco().ncks(str(self._ofile),output=str(self._ofile),exclude=True,force=True,variable=",".join(self._rvars))
 
+    ###########################################################################
+    def slice_variables(self):
```
bartgol, line 400: I used this functionality to grab `phi_int(nlev+1)` from the current IC file.

components/scream/src/dynamics/homme/dynamics_driven_grids_manager.cpp
```diff
@@ -181,8 +181,8 @@ void DynamicsDrivenGridsManager::build_dynamics_grid () {
     Kokkos::parallel_for(policy, KOKKOS_LAMBDA(const KT::MemberType& team) {
       const Int i = team.league_rank();
 
-      EKAT_KERNEL_ASSERT_MSG(!isnan(lat(i)), "Error! NaN values detected for latitude.");
-      EKAT_KERNEL_ASSERT_MSG(!isnan(lon(i)), "Error! NaN values detected for longitude.");
+      EKAT_KERNEL_ASSERT_MSG(!ekat::impl::is_nan(lat(i)), "Error! NaN values detected for latitude.");
```
bartgol, line 184: SCREAM would not build on my laptop. This can indeed happen when using `isnan` (and similar std functions), depending on the way in which stdlib headers are included. Using ekat's wrapper is always the safe choice.

components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
```diff
@@ -142,10 +142,11 @@ void HommeDynamics::set_grids (const std::shared_ptr<const GridsManager> grids_m
   // Note: qv is needed to transform T<->Theta
   add_field<Updated> ("horiz_winds",   FL({COL,CMP, LEV},{ncols,2,nlev_mid}),m/s,   rgn,N);
   add_field<Updated> ("T_mid",         FL({COL,     LEV},{ncols,  nlev_mid}),K,     rgn,N);
-  add_field<Updated> ("phi_int",       FL({COL,    ILEV},{ncols,  nlev_int}),Pa/rho,rgn,N);
   add_field<Updated> ("pseudo_density",FL({COL,     LEV},{ncols,  nlev_mid}),Pa,    rgn,N);
   add_field<Updated> ("ps",            FL({COL         },{ncols           }),Pa,    rgn);
   add_field<Required>("qv",            FL({COL,     LEV},{ncols,  nlev_mid}),Q,     rgn,"tracers",N);
+  add_field<Required>("phis",          FL({COL         },{ncols           }),Pa/rho,rgn);
+  add_field<Computed>("phi_int",       FL({COL,    ILEV},{ncols,  nlev_int}),Pa/rho,rgn,N);
```
tcclevenger, line 149: So is the `phi_int` field still needed as even a computed field? I don't think any other AD process or SC uses it.

components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
```diff
@@ -1025,6 +1028,8 @@ void HommeDynamics::initialize_homme_state () {
   const auto ps0 = hvcoord.ps0 * hvcoord.hybrid_ai0;
 
   const auto policy = ESU::get_thread_range_parallel_scan_team_policy(nelem*NGP*NGP,npacks_mid);
+  const auto phis = m_helper_fields.at("phis_dyn").get_view<const Real***>();
```
tcclevenger, line 1031: Since there still is a `phis` and `phis_dyn`, should this be called `const auto phis_dyn =` to avoid confusion? Feel free to ignore.

tcclevenger (commented):
Just a couple of comments, but looks good to me. The PD remap changes are nice and easy to follow.

components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
bartgol: Right, it's not really needed. Just like `w_int`, tbf. I have wondered whether I should purge them or leave them. In terms of memory, it's "only" 2 phys grid fields. It adds 2 fields to remap during PD remap though. Doesn't seem like a huge waste to me. But I guess waste can add up over time.

@AaronDonahue what are your thoughts? Should we remove `phi_int` and `w_int` on the phys grid?

components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
bartgol: Yeah, let's prevent any future confusion.

components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
PeterCaldwell: We will probably want to write out w_int and maybe phi_int for analysis, but we can just do that on the dynamics grid anyway, right? And better to output stuff as close to its native environment as possible....

AaronDonahue: I'm with @PeterCaldwell .  I don't think we should get in the habit of remapping variables to physics that are not actually needed in physics.  If the need arises we can make that happen then.  But EAM is pretty rife with clutter of fields added to PBUF that are not used and fields added as possible output that may be outdated.  So as Luca pointed out, it's only 2 now but those can quickly add up and add more confusion than anything else.

AaronDonahue (approved):

components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
bartgol: @PeterCaldwell good call, I forgot about that feature (despite being the one that added it!). Then I will remove the phys grid version of phi_int and w_int from HommeDynamics. If one is just interested in outputing them, they can request to do so from file (even if they want it outputed on the phys grid!).

bartgol:
There's a conflicts of units between Homme's verison of `phis` (in `Pa / (kg/m3) = m2/s2`) and SHOC's version of `phis` (in `m`). SHOC seems to be interested in the `geopotential height` rather than the `geopotential` itself. Since `phi` is usually the name of the latter, I would change SHOC's field name to `surf_geopot_height`, or something like that.

tcclevenger:
@bartgol is this ready to be reviewed again?

bartgol:
@tcclevenger yes, thank you.

components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
```diff
@@ -139,13 +139,17 @@ void HommeDynamics::set_grids (const std::shared_ptr<const GridsManager> grids_m
...
+
   // Note: qv is needed to transform T<->Theta
   add_field<Updated> ("horiz_winds",   FL({COL,CMP, LEV},{ncols,2,nlev_mid}),m/s,   rgn,N);
   add_field<Updated> ("T_mid",         FL({COL,     LEV},{ncols,  nlev_mid}),K,     rgn,N);
-  add_field<Updated> ("phi_int",       FL({COL,    ILEV},{ncols,  nlev_int}),Pa/rho,rgn,N);
   add_field<Updated> ("pseudo_density",FL({COL,     LEV},{ncols,  nlev_mid}),Pa,    rgn,N);
   add_field<Updated> ("ps",            FL({COL         },{ncols           }),Pa,    rgn);
   add_field<Required>("qv",            FL({COL,     LEV},{ncols,  nlev_mid}),Q,     rgn,"tracers",N);
+  add_field<Required>("phis",          FL({COL         },{ncols           }),m2/s2, rgn);
+  add_field<Computed>("phi_int",       FL({COL,    ILEV},{ncols,  nlev_int}),m2/s2, rgn,N);
```
tcclevenger, line 152: What was the conclusion of `phi_int` hanging around?

tcclevenger (approved):
Looks good!

components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
bartgol: Ah, right, I said I would remove it, along with w_int. Commit coming.

tcclevenger: Sorry, didn't think about automerge being on...

PR #1381, opened 2022-01-26
Add a SCREAMv1 aquaplanet compset

brhillman:
Add a SCREAMv1 aquaplanet compset (F2000-SCREAMv1-AQP1). Changes are made to `scream_input.yaml` to pull the proper initial condition from the machine-specific inputdata location. However, this is not yet configured to work with `check_inputdata` to automatically download the initial condition file if it does not already exist locally, so for now we'd need to setup each machine manually. Also adds stdout message to print current timestep in AD. This should be considered a placeholder until we implement more robust logging capabilities.

### DIFF
--- a/components/scream/cime_config/buildnml
+++ b/components/scream/cime_config/buildnml
@@ -31,7 +31,7 @@ import yaml
 logger = logging.getLogger(__name__)
 
 CIME_VAR_RE = re.compile(r'[$][{](\w+)[}]')
-SWITCH_RE   = re.compile(r'<\s*(\w+)\s*:')
+SWITCH_RE   = re.compile(r'<\s*([^:]+)\s*:')
 
 ###############################################################################
 def do_cime_vars(entry, case):
--- a/components/scream/data/scream_input.yaml
+++ b/components/scream/data/scream_input.yaml
@@ -38,7 +38,7 @@ Atmosphere Driver:
 
   Initial Conditions:
     Physics GLL:
-      Filename: ./data/init_ne4np4.nc
+      Filename: "<${COMPSET} : .*SCREAM%AQUA.* => /sems-data-store/ACME/bhillma/scream/data/init_files/ne4np4_L72/aqua/homme_shoc_cld_spa_p3_rrtmgp_init_ne4np4_L72.nc : ./data/init_ne4np4.nc>"
       # Set X_prev = X
       T_mid_prev: T_mid
       horiz_winds_prev: horiz_winds
--- a/components/scream/data/scream_input.yaml
+++ b/components/scream/data/scream_input.yaml
@@ -38,7 +38,7 @@ Atmosphere Driver:
 
   Initial Conditions:
     Physics GLL:
-      Filename: "<${COMPSET} : .*SCREAM%AQUA.* => /sems-data-store/ACME/bhillma/scream/data/init_files/ne4np4_L72/aqua/homme_shoc_cld_spa_p3_rrtmgp_init_ne4np4_L72.nc : ./data/init_ne4np4.nc>"
+      Filename: "<${COMPSET} : .*SCREAM%AQUA.* => ${DIN_LOC_ROOT}/atm/scream/init/scream-aquaplanet_init_ne4np4_L72_20211202.nc : ./data/init_ne4np4.nc>"
       # Set X_prev = X
       T_mid_prev: T_mid
       horiz_winds_prev: horiz_winds
--- a/components/scream/src/control/atmosphere_driver.cpp
+++ b/components/scream/src/control/atmosphere_driver.cpp
@@ -802,6 +802,11 @@ void AtmosphereDriver::run (const int dt) {
   // Make sure the end of the time step is after the current start_time
   EKAT_REQUIRE_MSG (dt>0, "Error! Input time step must be positive.\n");
 
+  // Print current timestamp information
+  if (m_atm_comm.am_i_root()) {
+    std::cout << "Atmosphere step = " << m_current_ts.get_num_steps() << "; model time = " << m_current_ts.get_date_string() << " " << m_current_ts.get_time_string() << std::endl;
+  }
+
   if (m_surface_coupling) {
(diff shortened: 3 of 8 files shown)


components/scream/src/control/atmosphere_driver.cpp
```diff
@@ -802,6 +802,11 @@ void AtmosphereDriver::run (const int dt) {
   // Make sure the end of the time step is after the current start_time
   EKAT_REQUIRE_MSG (dt>0, "Error! Input time step must be positive.\n");
 
+  // Print current timestamp information
```
AaronDonahue, line 805: Probably worth putting in the PR description that you also add a timestamp to typical output.  Might be worth having this be a runtime option, it would be pretty straightforward to have a YAML key for it that defaults to TRUE.

brhillman: Yeah sure. I could also remove it if you think it doesn't belong here.

bartgol: Imho, this can be addressed when tackling the output logging. Rather than add a yaml entry for each piece of information, I think it will be easier to add a "log level" option, and internally we decide what level is required to output each piece of information.

brhillman: @bartgol @AaronDonahue do you want me to remove this for now then? I find it pretty useful to see what step the model made it to without looking at the output netcdf files, which first requires hand configuring that output.

bartgol: I think you can keep it where it is.

AaronDonahue: agreed, we can always replace when we get more robust logging in place

AaronDonahue:
I'm not sure why this PR is giving a DIFF in all of the tests.  Should we rebase and try again?

brhillman:
@AaronDonahue yes I'll investigate.

brhillman:
This passes baselines for a local merge, so I'm adding the retest label.

components/scream/cime_config/buildnml
```diff
@@ -47,7 +47,7 @@ import yaml
 logger = logging.getLogger(__name__)
 
 CIME_VAR_RE = re.compile(r'[$][{](\w+)[}]')
-SWITCH_RE   = re.compile(r'<\s*(\w+)\s*:')
+SWITCH_RE   = re.compile(r'<\s*([^:]+)\s*:')
```
AaronDonahue, line 50: This is the only change in the whole PR that I could see even possibly touching SCREAMv0 tests.  @jgfouca and/or @brhillman .  Do you have any thoughts on what is causing the DIFFs when we run the AT?

AaronDonahue (approved):

components/scream/cime_config/buildnml
jgfouca: @AaronDonahue , scream is not currently using this syntax in its YAML files. This change should not be causing DIFFs.

jgfouca: I see a new test was added. That would definitely cause a DIFF (no baselines).

components/scream/data/scream_input.yaml
```diff
@@ -38,7 +38,7 @@ Atmosphere Driver:
 
   Initial Conditions:
     Physics GLL:
-      Filename: ./data/init_ne4np4.nc
+      Filename: "<${COMPSET} : .*SCREAM%AQUA.* => ${DIN_LOC_ROOT}/atm/scream/init/scream-aquaplanet_init_ne4np4_L72_20211202.nc : ./data/init_ne4np4.nc>"
```
AaronDonahue, line 41: Okay cool. Now we pass all v0 tests but still have a DIFF in the original SCREAMV1 test. Is it possible this logical is not working correctly?  Just trying to track down the last unexpected DIFF in the AT results.

brhillman: It appears to be working correctly, I just ran this test manually and it gets the proper (old) initial condition. But, then I ran this test on current master, and the v1 test is DIFF'ing on current SCREAM master for me. @bartgol was also tracking down another unexpected DIFF from the v1 tests with #1356 and suspected something funny with baselines.

AaronDonahue:
> @brhillman We would have to merge master in the branch, since v0/v1 tests do not auto merge master in the PR branch before testing it (*), like we do with test-all-scream. So whatever change caused the baselines to change, is not in this branch. However, instead of merging master, we could _locally_ do it on mappy, run the v1 tests, and then admin-merge this PR once they pass. Let me know if you prefer update the PR merging master, and let the AT do the job, or if you want me to manually test. Or if you want to do the manual thing on mappy.
> 
> (*) Jim is working on adding this to our jenkins testing script.

@brhillman , didn't you already manually test awhile back, before the rebase?  Probably could just do that again and then merge this.

brhillman:
> > @brhillman We would have to merge master in the branch, since v0/v1 tests do not auto merge master in the PR branch before testing it (*), like we do with test-all-scream. So whatever change caused the baselines to change, is not in this branch. However, instead of merging master, we could _locally_ do it on mappy, run the v1 tests, and then admin-merge this PR once they pass. Let me know if you prefer update the PR merging master, and let the AT do the job, or if you want me to manually test. Or if you want to do the manual thing on mappy.
> > (*) Jim is working on adding this to our jenkins testing script.
> 
> @brhillman , didn't you already manually test awhile back, before the rebase? Probably could just do that again and then merge this.

Lots of testing! @bartgol and I have been going back and forth about this and another open PR. It looks like testing finally resolved. Luca, correct me if I'm wrong, but looks like the only diff is the new aquaplanet test this PR adds, so I think this is ready for manual merge.

bartgol:
@brhillman I'm assuming the v1 diff is expected? If so, then we can merge.

AaronDonahue:
Yes, now it looks we are finally whole.  As Jim said, a new test will get flagged.  Since this PR adds the aquaplanet test I think this is expected.  I'm going to manually merge.

PR #1382, opened 2022-01-26
Refactor handling of scream_input.yaml

jgfouca:
A fair bit of effort was made to preserve the comments in the repo
file for screan_input.yaml during the processing of that file for
the case. The idea was that users would be interacting with this
file in their cases and we wanted them to have access to our comments
directly to assist them.

YAML does not have any internal mechasisms for preserving comments
when parsing/dumping YAML content, so I had added some hacky code
to read comments separately and reestablish them in the file manually.

After a few months of developers creating CIME SCREAM V1
cases, it seems this approach has been a source of syntax errors
in this case's version of this file, so this PR is a rethink of
the approach.

Instead of relying solely on inline comments in the repo's scream_input.yaml,
we provide a separate SCREAM_INPUT_README file for full documentation
on how to use scream_input.yaml.

### DIFF
--- a/components/scream/data/SCREAM_YAML_README
+++ b/components/scream/data/SCREAM_YAML_README
@@ -1,0 +1,51 @@
+
+INTRO:
+
+The scream_input.yaml is the key file for configuring a SCREAM run. This file will be
+processed and copied to $case/run/scream_input.yaml by scream's buidnml script, which
+is called during case.setup. Note, this is for runtime coniguration
+only. Cmake/build-time configuration should be done through SCREAM_CMAKE_OPTIONS.
+
+For inline comments, see the version of scream_input.yaml that lives in the repo
+(components/scream/data/scream_input.yaml)
+
+Note, the $case/run/scream_input.yaml will NEVER be overwritten by subsequent
+calls to case.setup/buildnml in order to avoid blowing away potential local
+modifications. To force a regeneration of this file, it should be removed from the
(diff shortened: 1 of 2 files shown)


components/scream/data/SCREAM_YAML_README
```diff
@@ -0,0 +1,51 @@
...
+
+SYNTAX:
+
+This file supports some special syntax in addition to basic YAML:
+'${VAR}' will be used to refer to env variables in the CIME case
+
+'<switch_val : key1 => val1 : key2 => val2 : elseval>' will be used to express conditional
+statements. If switch_val matches key1, then the expression evaluates to val1; if switch_val
+matches key2, then the expression evaluates to val2; if it matches neither, then
+the expression evaluates to elseval. The elseval component of this expression is optional.
```
bartgol, line 34: Might be worth pointing out that the user can have an arbitrary number of `key => val` sections.

bartgol (approved):
Neat. I like the idea of putting docs in a separate file. I have one suggestion, but very minor.

AaronDonahue (approved):

jgfouca:
The AT seems increasingly flaky. If you take the union of passed tests in the 2 AT runs of the PR, everything is passing, so I'm going to declare this PR as PASS.

components/scream/data/SCREAM_YAML_README
jgfouca: Done.

PR #1383, opened 2022-01-27
SCREAM: add some fine-grain switches when testing with jenkins

bartgol:
A couple of upgrades:

* Can skip scream standalone testing setting the `AT: Skip Stand-Alone Testing` on the PR
* Can set `SCREAMv0` or `SCREAMv1` PR labels to run CIME cases for v0 or v1 only, respectively. Useful when making changes that _for sure_ only affect one of the two. The `CIME` label is still supported for backward compatibility, and is equivalent to set both `SCREAMv0` and `SCREAMv1`.

@jgfouca I'm not sure how to test the new mods, other than trying to add the labels that trigger the new behaviors. I guess we can set/unset labels, and ask the AT to retest, to ensure this PR works. But maybe there's a way to use the scripts tests?

Probably separate topic: should we make v1 testing ON by default, and add a `AT: Skip v1 Testing` label instead? This PR keeps the default behavior the same as master (test SA scream, do not test CIME scream). But we should start thinking about whether these defaults need to change.

### DIFF
--- a/components/scream/scripts/jenkins/jenkins_common_impl.sh
+++ b/components/scream/scripts/jenkins/jenkins_common_impl.sh
@@ -3,10 +3,12 @@
 # Check if the env var PR_LABELS is defined, and contains something meaningful
 IFS=';' read -r -a labels <<< "$PR_LABELS";
 
-# default values
+# default values. By default, test only Stand-Alone (SA) on all machines
 skip_testing=0
 test_scripts=0
-test_cime=0
+test_v0=0
+test_v1=0
+test_SA=1
 skip_mappy=0
 skip_weaver=0
 skip_blake=0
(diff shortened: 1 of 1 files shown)


jgfouca (changes requested):
I like these enhancements. I think this is a good opportunity for some additional cleanups. One thing I've wanted for a while is for any label that affects AT behavior to have an `AT: ` prefix. So `AT: CIME` instead of just `CIME`. 

One potential point of confusion is that, if I have both `CIME` and `SCREAMv1` labels, both v1 and v0 will be tested. Maybe we should get rid of `CIME` and just have the v0 and v1 labels.

I hate to say it, but this script is getting complicated enough that it might need some real testing. I can look into some options here... I have some ideas.

jgfouca (approved):

AaronDonahue:
@bartgol , it looks like this PR is basically ready, except it fails with a DIFF in the SCREAMv1 test on mappy-cime.  Why would these changes make a DIFF?  Is it possible we just need to rebase and retest?

bartgol:
It's likely a matter of rebase. Let me update the branch.

jgfouca (approved):

PR #1387, opened 2022-01-27
Upgrades to scripts-tests framework

bartgol:
CMake/CTest offers a property for tests, `WILL_FAIL`, which makes the ctest test pass if the underlying test fails (a negation of the test outcome, if you will).

We can use this feature to ensure that we get failures with, e.g., OOB access in kokkos views. We do still need to use scripts-tests to detect a build failure, but any runtime failure can be tested with ctest native feature.

I haven't added a similar test for p3 baseline check. The way it is currently implemented, we would have to build p3 twice, which can be expensive on GPU archs. @jgfouca what would you think if I removed

```
#if defined(SCREAM_FORCE_RUN_DIFF) || defined(SCREAM_FORCE_RUN_DIFF_BFB_UNIT)
        precip_liq_surf *= 2;
#endif
```
from `p3_rain_sed_impl.hpp`, and instead do something similar in p3_run_and_cmp.cpp, like
```
  #if defined(SCREAM_FORCE_RUN_DIFF_BFB_UNIT)
      const auto num = std::abs(a[i] - b[i]);
  #else
      const auto num = std::abs(a[i]*1.2 - b[i]);
  #endif
```
and something similar in a p3 unit test? That way we could ensure diff and baseline diff failures happen during any normal test run (not just scripts tests runs).


jgfouca (changes requested):
I like the general idea that we should prefer ctests over scripts-tests where it makes sense. That said, I have some concerns here.

Let's take the valgrind test. There are a lot of layers of code between `test-all-scream ... -t valg` and the cmake code you added to `fail_check/CMakeLists.txt`. If those layers don't work correctly, then this approach will not detect it. For example, if there is a problem in test-all-scream and `-t valg` makes a build that does not enable EKAT_ENABLE_VALGRIND, we would appear to get 100% PASS even though there's problem.

In other words, the tests you removed test both the scripts infrastructure and the core SCREAM cmake, while the tests you added only test core SCREAM.

On the other hand, I see the problem with the scripts-tests is that test-all-scream can pass/fail for any reason, not necessarily the thing we think we are testing.

Not sure what the best approach is here... I suppose the most-thorough would keep the scripts-tests and actually test the output coming from test-all-scream to make sure the reason for the pass/fail is what we think it should be.

bartgol:
That's a good point. Perhaps we can use scripts-test to ensure that some key CMake variables are set (inspecting CMakeCache.txt, for instance). If those are set, then whether or not testing works can be handled inside cmake.

We could actually use this paradigm for more than just valg. E.g., to ensure that `-t opt` yields CMAKE_BUILD_TYPE=RELEASE in the cache (that is, the `-t opt` flag is doing what we wanted).

So I see 3 layers of things to test:
- verify that test-all can generate the correct cmake configuration
- verify that a certain cmake config yields ctest pass/fails where we expect
- verify that test-all correctly picks up ctest's config/build/run exit code

The first and latter must be done with scripts-tests. But the middle one can (should?) be done internally with cmake/ctest features.

jgfouca:
@bartgol , that's a good plan.

bartgol:
I think I reached a good status. I was able to do things a bit more cleanly (wihtout, e.g., the need to create fake baselines), and I think I lowered the scripts tests runtime quite a bit. I was running on mappy, while mappy_e3sm_master was running (I taskset-ed the scripts test on cores that jenkins was using less often, which btw caused failures of the spread tests), and I ran scripts-tests in about 10 minutes or so (I didn't run with time, so I had to manually keep an eye on the process with top/ps). From jenkins, I see that our nightly scripts tests take about 30min to run, so this might be a good improvement.

Main ingredients of the speedup:
- Run most of the tests to detect failures during normal builds, as additional tests (little compilation/testing overhead)
- Avoid building _all_ of the src folder and most of the test folder if SCREAM_FAKE_ONLY is in the env
- Run some of the scripts tests with SCREAM_FAKE_ONLY on and/or test-all-scream options such as `--no-tests` or `--config-only` (the latter, a new one), to speed up each call to test-all. E.g, when testing baseline handling logic, we do not need to actually build baselines; we only need test-all to run its logic to see if new baselines are needed, and if, so, to generate the baselines dir, with the baseline_sha file. Building/running tests/baselines execs is not needed.
- When running test_full (the most expensive test), add the `-p` flag to test-all, to squeeze the last drop of perf out of the CPU cores. Note: I did not have this option when I timed scripts-tests on mappy, so performance might be even a bit better.

@jgfouca I have two questions for you:
- While moving the "FORCE_DIFF" tests from scripts-tests to inside our normal cmake builds, I did not move the tests that check no diff are triggered in opt mode. I did not quite get that test. Why shouldn't diffs trigger a fail in release mode?
- Do we really need to "really" run test-all-scream in test_full? What are we actually testing? We should not care whether tests pass or not, since that's a responsibility of the normal nightlies to check. In scripts-tests we simply want to ensure the scripts are doing what we think they are. What's the goal of running a real (i.e., no FAKE) test-all-scream?

jgfouca:
@bartgol 

> While moving the "FORCE_DIFF" tests from scripts-tests to inside our normal cmake builds, I did not recreate the tests that check no diff are triggered in opt mode. I did not quite get that test. Why shouldn't diffs trigger a fail in release mode?

Things are not BFB between fortran and cxx in release mode.

> Do we really need to "really" run test-all-scream in test_full? What are we actually testing? We should not care whether tests pass or not, since that's a responsibility of the normal nightlies to check. In scripts-tests we simply want to ensure the scripts are doing what we think they are. What's the goal of running a real (i.e., no FAKE) test-all-scream?

It looks like test_full is used to hit some use cases:
```
    CMDS_TO_TEST = [
        "./test-all-scream -m $machine -b HEAD -k",
        "./test-all-scream -m $machine -b HEAD -k -t dbg",
        "./test-all-scream --baseline-dir $results -c EKAT_DISABLE_TPL_WARNINGS=ON -i -m $machine --submit --dry-run", # always dry run                                                                                                           
    ]
```
The test runs these commands and stores all output which can then serve as a baseline to check against. This is to check that behavior is exactly identical to the version used in the baseline, at least as far as it can be determined from the output.

components/scream/scripts/scripts-tests
```diff
@@ -289,208 +289,191 @@ class TestTestAllScream(TestBaseOuter.TestBase):
...
-            test_cmake_cache_contents(self, "full_debug", "SCREAM_FPE", "FALSE")
-            if is_cuda_machine(self._machine):
-                test_cmake_cache_contents(self, "full_debug", "SCREAM_PACK_SIZE", "1")
-            else:
-                test_cmake_cache_contents(self, "full_debug", "SCREAM_PACK_SIZE", "16")
+        Test the 'dbg' test in test-all-scream. It should set certain CMake values
+        """
+        options = "-b HEAD -k -t dbg --config-only"
+        cmd = self.get_cmd("./test-all-scream -m $machine {}".format(options),
+                            self._machine, dry_run=False)
```
jgfouca, line 299: We should test that `Kokkos_ENABLE_DEBUG_BOUNDS_CHECK` is on.

jgfouca (changes requested):
Looks good, just one minor request.

jgfouca:
@bartgol most of the work is in the base class. Look for compare/generate in the code.

jgfouca (approved):

jgfouca (approved):

PR #1388, opened 2022-01-27
Update homme namelist options for SCREAMv1 run.

AaronDonahue:
This commit changes the default homme namelist options as defined in
the scream_input.yaml file to match those of a recent simulation which
was able to run 30 days.

### DIFF
--- a/components/scream/data/scream_input.yaml
+++ b/components/scream/data/scream_input.yaml
@@ -7,18 +7,28 @@ Atmosphere Driver:
     Schedule Type: Sequential
     Process 0:
       Process Name: Homme
-      Grid:         Dynamics
+      Enable Output Fields Checks: false
+      Enable Input Fields Checks: false
+      Grid: Dynamics
     Process 1:
       Process Name: SHOC
+      Enable Output Fields Checks: false
+      Enable Input Fields Checks: false
       Grid: Physics GLL
     Process 2:
       Process Name: CldFraction
+      Enable Output Fields Checks: false
+      Enable Input Fields Checks: false
       Grid: Physics GLL
     Process 3:
       Process Name: P3
+      Enable Output Fields Checks: false
+      Enable Input Fields Checks: false
       Grid: Physics GLL
     Process 4:
       Process Name: RRTMGP
+      Enable Output Fields Checks: false
+      Enable Input Fields Checks: false
       Grid: Physics GLL
       active_gases:
       - h2o
@@ -29,6 +39,12 @@ Atmosphere Driver:
       - ch4
       - o2
       - n2
+
+  Time Stepping:
+    Time Step: 300
+    Start Time: [12, 30, 00]      # Hours, Minutes, Seconds
+    Start Date: [2000, 2, 1]    # Year, Month, Day
+    Number of Steps: 288
 
   Grids Manager:
     Type: Dynamics Driven
@@ -61,31 +77,32 @@ HOMME:
     partmethod        : 4
     cubed_sphere_map  : 0
     topology          : cube
-    test_case         : jw_baroclinic
     u_perturb         : 1
     rotate_grid       : 0
     ne                : 4
     statefreq         : 9999
     mesh_file         : /dev/null
     tstep             : 300
-    nu                : 7.0e+15
-    nu_div            : 1.0e+15
-    nu_p              : 7.0e+15
-    nu_q              : 7.0e+15
-    nu_s              : -1
-    nu_top            : 2.5e+5
+    nu                :  .000000034
+    nu_div            : -1 
+    nu_p              : .000000034
+    nu_q              : .000000034
+    nu_s              : .000000034
+    nu_top            : 
     se_ftype          : 0
     limiter_option    : 9
-    vert_remap_q_alg  : 1
-    hypervis_scaling  : 0
+    vert_remap_q_alg  : 10
+    hypervis_scaling  : 3.0
     hypervis_order    : 2
-    hypervis_subcycle : 2
-    hypervis_subcycle_tom : 0
-    theta_hydrostatic_mode : false
-    theta_advect_form : 0
-    tstep_type        : 10
-    dt_remap_factor   : 3
+    hypervis_subcycle : 1
+    hypervis_subcycle_tom : 1
+    theta_hydrostatic_mode : true
+    theta_advect_form : 1
+    tstep_type        : 5
+    dt_remap_factor   : 1
     dt_tracer_factor  : 1
+    transport_alg     : 12
+    moisture:           moist
 
   vert_nl:
     vfile_mid : ./data/scream-72m.ascii
--- a/components/scream/data/scream_input.yaml
+++ b/components/scream/data/scream_input.yaml
@@ -63,6 +63,10 @@ Atmosphere Driver:
     Dynamics:
       tracers: tracers, Physics GLL
 
+  # The parameters for I/O control
+  Scorpio:
+    Output YAML Files: ["data/screamv1_output.yaml"]
+
 SCREAM:
 
   Start Date: "${RUN_STARTDATE}"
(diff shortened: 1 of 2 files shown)


components/scream/data/scream_input.yaml
```diff
@@ -7,18 +7,28 @@ Atmosphere Driver:
     Schedule Type: Sequential
     Process 0:
       Process Name: Homme
-      Grid:         Dynamics
+      Enable Output Fields Checks: false
```
brhillman, line 10: Do we want to disable these checks altogether, or just loosen the tolerances? I was able to find some success by lowering the temperature threshold minimum from 150 to 140 in P3 and SHOC.

components/scream/data/scream_input.yaml
```diff
@@ -61,31 +77,32 @@ HOMME:
...
     statefreq         : 9999
     mesh_file         : /dev/null
     tstep             : 300
-    nu                : 7.0e+15
-    nu_div            : 1.0e+15
-    nu_p              : 7.0e+15
-    nu_q              : 7.0e+15
-    nu_s              : -1
-    nu_top            : 2.5e+5
+    nu                :  .000000034
```
brhillman, line 86: I think this should use scientific notation.

components/scream/data/scream_input.yaml
```diff
@@ -61,31 +77,32 @@ HOMME:
...
+    nu_div            : -1 
+    nu_p              : .000000034
```
brhillman, line 88: These can be `-1` to adopt values from `nu`.

components/scream/data/scream_input.yaml
```diff
@@ -61,31 +77,32 @@ HOMME:
...
+    nu_q              : .000000034
+    nu_s              : .000000034
+    nu_top            :
```
brhillman, line 91: Did you mean to leave `nu_top` empty here?

components/scream/data/scream_input.yaml
```diff
@@ -61,31 +77,32 @@ HOMME:
...
-    tstep_type        : 10
-    dt_remap_factor   : 3
+    hypervis_subcycle : 1
+    hypervis_subcycle_tom : 1
+    theta_hydrostatic_mode : true
+    theta_advect_form : 1
+    tstep_type        : 5
+    dt_remap_factor   : 1
     dt_tracer_factor  : 1
+    transport_alg     : 12
```
brhillman, line 104: Was this working for you? It looks to me like `transport_alg` option hasn't been added to our HOMME namelist read routine.

brhillman (changes requested):
Thanks for putting this together Aaron. I have some minor edits and questions but overall this is a big improvement.

components/scream/data/scream_input.yaml
AaronDonahue: Good question.  I think your AQ PR drops the threshold right?  I was hoping to make this a simple change on one file.  But I can also drop the tolerance in the tests if you want.

components/scream/data/scream_input.yaml
AaronDonahue: I've had limited success with scientific notation and the yaml parser.  Sometimes it grabs it as a string, maybe when we use `e-X`, maybe the `-` throws it off for some reason.  I'll have to check.

components/scream/data/scream_input.yaml
AaronDonahue: You would think, but when I do that I get an error with the nu values being -1.  So I think our namelist parser still has some kinks.

components/scream/data/scream_input.yaml
AaronDonahue: ooops, good catch.  I'll fix that

components/scream/data/scream_input.yaml
bartgol: I forgot to follow up on this, but on GPU the output fields end up failing the checks _really badly_. That is likely a bug that I still have to fix.

To give an idea: homme _can_ generate negative tracers, but they are very tiny numbers (like -1e-17), due to roundoff during remap. Those can easily and safely be fixed by setting them to 0. However, we first check that we don't get "too negative values". Well, on GPU I get -1e-9, which is definitely too negative.

I have to investigate what's happening. Meanwhile, disabling the checks can help keeping cdash green.

components/scream/data/scream_input.yaml
AaronDonahue: I think you are right.  This option may not be used.  To be safe I copied everything I had in the new namelist.  But I can remove this.

components/scream/data/scream_input.yaml
brhillman: Okay, that's fine. I guess it just makes me a little uncomfortable to disable checks because we fail them...

components/scream/data/scream_input.yaml
brhillman: I guess if it passes that's fine, but including this one for me was throwing zeros into the rest of the fields.

components/scream/data/screamv1_output.yaml
```diff
@@ -0,0 +1,65 @@
+%YAML 1.1
+---
+Casename: SCREAMv1_output_hydrostatic
+Averaging Type: Instant
+Max Snapshots Per File: 1000
```
brhillman, line 5: Maybe we don't care so much about this for this PR, but with 1000 snapshots per file and outputs every 12 steps, won't we get file splits at weird times of day? Would it make more sense to configure to keep 1-file-per-day, or something like that?

components/scream/data/scream_input.yaml
```diff
@@ -61,31 +81,31 @@ HOMME:
...
     hypervis_order    : 2
-    hypervis_subcycle : 2
-    hypervis_subcycle_tom : 0
-    theta_hydrostatic_mode : false
-    theta_advect_form : 0
-    tstep_type        : 10
-    dt_remap_factor   : 3
+    hypervis_subcycle : 1
+    hypervis_subcycle_tom : 1
+    theta_hydrostatic_mode : true
```
bartgol, line 103: Should we use non-hydro by default? I feel like stuff in the data folder might be used as a template by other people, without paying attention to some parameters. Besides, non-hydro might stress test more sections of the code?

Note: non hydro requires tstep_type 10.

bartgol (approved):
Looks good. I have one comment regarding the default tsteptype/hydro-mode choice.

components/scream/data/screamv1_output.yaml
AaronDonahue: The 1000 is just the max amount of steps the file can hold before a new one is created.

AaronDonahue: For sure, we could do that. I figured for the analysis we are doing now having one big file is the easiest to visualize

components/scream/data/scream_input.yaml
AaronDonahue: Yeah we can do that. This name list setup is for the one case that ran 31 days. And that one was hydrostatic. I've tried nongydrostatic but get errors. Although we may have addressed those recently. So probably worth trying again.

brhillman: Actually, we probably want both. We do this for SCREAMv0, we have a LR test that uses hydro, and a HR test that uses non-hydro. Note that Jim's yaml parser can query compset and we can set this appropriately. I vote we keep this as Aaron has it for this PR, and then add a subsequent PR with another v1 compset that tests the HR (non-hydro) config.

AaronDonahue:
I mainly put this PR in to make it easy for other folks to get a run going. My hope is that as we fix bugs we will keep updating the defaults.

brhillman (approved):

components/scream/data/scream_input.yaml
bartgol: Our YAML parser should not have issues digesting scientific format.

components/scream/data/scream_input.yaml
bartgol: Sounds good.

AaronDonahue:
As expected this PR produces a DIFF in the v1 tests.  I'm going to wait to manually merge until after we resolve the DIFF for our nightly v1 test on mappy.

bartgol:
If this was expected to produce DIFFs, there's no point waiting to merge. I mean, even if the current diffs are wrong and have to be fixed, and we go ahead and fix them, this PR will still yield the same result (diff).

I can inspect the current diffs even if this gets merged...

AaronDonahue:
Okay, now that we have a clean nightly test for SCREAMv1 I'm going to manually merge this change.

PR #1390, opened 2022-01-31
Add units and long-name to SCREAM netcdf output

PeterCaldwell:
SCREAMv1 netcdf output files lacked info about the units of each variable as well as a long name explaining them. This PR adds that.

### DIFF
--- a/components/scream/src/share/io/scream_output_manager.cpp
+++ b/components/scream/src/share/io/scream_output_manager.cpp
@@ -213,7 +213,7 @@ void OutputManager::run(const util::TimeStamp& timestamp)
       register_dimension(filename,"time","time",0);
 
       // Register time as a variable.
-      register_variable(filename,"time","time",1,{"time"},  PIO_REAL,"time");
+      register_variable(filename,"time","time","sec",1,{"time"},  PIO_REAL,"time");
 
       // Make all output streams register their dims/vars
       for (auto& it : m_output_streams) {
--- a/components/scream/src/share/io/scream_output_manager.cpp
+++ b/components/scream/src/share/io/scream_output_manager.cpp
@@ -213,7 +213,7 @@ void OutputManager::run(const util::TimeStamp& timestamp)
       register_dimension(filename,"time","time",0);
 
       // Register time as a variable.
-      register_variable(filename,"time","time","sec",1,{"time"},  PIO_REAL,"time");
+      register_variable(filename,"time","time","s",1,{"time"},  PIO_REAL,"time");
 
       // Make all output streams register their dims/vars
       for (auto& it : m_output_streams) {
--- a/components/scream/src/share/io/scorpio_output.cpp
+++ b/components/scream/src/share/io/scorpio_output.cpp
@@ -1,6 +1,7 @@
 #include "share/io/scorpio_output.hpp"
 #include "ekat/std_meta/ekat_std_utils.hpp"
 #include "share/io/scorpio_input.hpp"
+#include "ekat/util/ekat_units.hpp"
 
 #include "ekat/util/ekat_string_utils.hpp"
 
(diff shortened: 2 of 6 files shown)


AaronDonahue (approved):
Looks good to me.  Probably not important for this PR, but it might be good to also read the units of an input file and throw an error if they are different.  Just mentioning now since it comes to mind.

components/scream/src/share/io/scream_output_manager.cpp
```diff
@@ -213,7 +213,7 @@ void OutputManager::run(const util::TimeStamp& timestamp)
       register_dimension(filename,"time","time",0);
 
       // Register time as a variable.
-      register_variable(filename,"time","time",1,{"time"},  PIO_REAL,"time");
+      register_variable(filename,"time","time","sec",1,{"time"},  PIO_REAL,"time");
```
bartgol, line 216: EKAT uses "s" for seconds, which is the standard symbol. All units printed by ekat should have 's'. Do climate app prefer 'sec'?

PeterCaldwell:
"Probably not important for this PR, but it might be good to also read the units of an input file and throw an error if they are different. Just mentioning now since it comes to mind."
@AaronDonahue - I don't really understand this comment. Are you saying I should create a unit test which compares the units of an input file for a restart against the output file units? Do we expect them to be identical? I guess they should be right now, but we may have to change output units later to satisfy CF conventions... I'm not sure.

bartgol:
I think Aaron wants our scorpio input to throw if the nc file a) does not store units or b) stores different units (or maybe just the latter). Currently, we never inspect the units of fields in the IC file.

IMHO, it could become a big task (might require a few scorpio interface changes), so better suited for a separate PR.

components/scream/src/share/io/scream_output_manager.cpp
PeterCaldwell: Good point. I've fixed this to use "s" instead for consistency.

AaronDonahue:
Yes, @bartgol is right, and I don't think it should go into this PR.  I just mentioned it because it came to my mind while reviewing this PR.

PeterCaldwell:
Oh, one question for @AaronDonahue and @bartgol - this PR is WIP because it sets long name in the .nc file equal to just the variable name. I did this because it looked to me like the FM doesn't currently store long names for the variables. Am I mistaken? Is this something we want to change? Changing the FM to add long names might be above my pay grade...;-)

AaronDonahue:
@PeterCaldwell , as far as I know there is not yet a longname attribute assigned to a field, but it shouldn't be that much work to put one in.  Would you like Luca or I to make that happen and add it to this PR?  Or let this one go in and we make a separate PR?

bartgol:
Yeah, the Field stack structures don't store that info. I had an old branch where I added something in that direction. I will revive it once I run out of more urgent tasks.

That said, for now, we can set longname=shortname=field_id.name(), whenever we need.

AaronDonahue:
> Yeah, the Field stack structures don't store that info. I had an old branch where I added something in that direction. I will revive it once I run out of more urgent tasks.
> 
> That said, for now, we can set longname=shortname=field_id.name(), whenever we need.

oh yeah, good point Luca.  If I understand correctly, we can add longname to the field attributes and just have it set to shortname for now.  Then later we can add something more sophisticated and that will automatically update anywhere else that uses the longname attribute.

PeterCaldwell:
So the suggestion is to add longname to the FM in this PR, but to initialize all longname values to the shortname to minimize the work now. Then to create a new PR when we get around to it to provide real longname values? 

I wonder if we could just grab longnames from the initial condition file?

bartgol:
> So the suggestion is to add longname to the FM in this PR, but to initialize all longname values to the shortname to minimize the work now. Then to create a new PR when we get around to it to provide real longname values?
> 
> I wonder if we could just grab longnames from the initial condition file?

No, my suggestion is to not bother adding long name to the Field stack classes in this PR. Whenever you need a short/long name in scorpio, let's just use the field name twice. Later, we can work on adding long name support.

PeterCaldwell:
"let's just use the field name twice. Later, we can work on adding long name support."

Ok, that's what I already do, so I think this PR is good to go.

bartgol (approved):

PeterCaldwell:
@AaronDonahue - you said yesterday you thought this PR would fail because it changes the output... that doesn't seem to be the case. I guess cprnc or whatever else we test doesn't care about units? Just pinging you in case there's something sketchy going on.

AaronDonahue:
@PeterCaldwell , yeah.  After we chatted I reconsidered my theory.  So I was happy you turned off WIP and let the tester do its thing.

PR #1391, opened 2022-02-01
Major refactor to input-file handling in SCREAM

jgfouca:
This is a large scale refactor with wide impacts. The idea is to move away from having input files bundled directly in the repo and instead use the input file download system that is supported by E3SM/CIME.

Change list:
1) New SCREAM_MACHINE cmake cache var, optional. Machines that are also supported by E3SM should use this setting to ensure that SCREAM_INPUT_ROOT is automatically set to E3SM's DIN_LOC_ROOT config_machines setting.
2) New SCREAM_INPUT_ROOT cmake cache var. This will be derived using SCREAM_MACHINE and the `query-cime` script if SCREAM_MACHINE is known. Otherwise, this setting will need to be provided by a cache file or on the cmake command line.
3) big upgrades to scream's `buildnml` script to use CIME's native `check_input_data` capabilities.
4) New scream script `check-input` that is a thin hook into CIME's `check_input_data` capabilities that is usable in standalone scream.
5) New `GetInputFile` cmake function for SCREAM that uses `check-input` in standalone mode to ensure input files are present.
6) Changes to scream_input.yaml to use this new system to acquire all input files needed for CIME SCREAM V1 cases.
7) Changes all over SCREAM's CMake system to use `GetInputFile` instead of `configure_file`.
8) Some other minor cleanups.


AaronDonahue (commented):
@jgfouca , thank you for this major set of changes.  I can't speak to the python or cmake changes directly so I defer my approval to other reviewers.

I do have one question though.  It looks like all the initial condition files for all of the tests have been removed.  I imagine this is because we will expect the CIME infrastructure to go ahead and download all files to the DIN_LOCAL_ROOT location.  My question is, what about for users who just want to work on a local machine, like a laptop.  Presumably they would have to have access to a server to grab those files, and they would have to define a DIN_LOCAL_ROOT for their machine?

PeterCaldwell:
Just to make sure I understand what's going on here: the goal of this PR is to use E3SM's capability to check if the input files we need exist locally and to download the needed files if not, right? If so, do we need to move all our input files to the inputdata server on Anvil/Blues in conjunction with this PR?

And this PR is totally separate from the problem we've been having lately where hommexx config options are ignored, right?

components/scream/CMakeLists.txt
```diff
@@ -4,8 +4,6 @@ if (NOT DEFINED PROJECT_NAME)
   set(SCREAM_CIME_BUILD FALSE)
 else()
   set(SCREAM_CIME_BUILD TRUE)
-  set (SCREAM_SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR})
```
bartgol, line 7: This might come down to taste, but why removing these? As it is now, scream does not call `project()` in a CIME build, so these two won't get declared in a CIME build (`project()` sets them for you). They are handy variables, since they are [standard cmake vars](https://cmake.org/cmake/help/latest/variable/PROJECT-NAME_SOURCE_DIR.html) (if `project()` is called). Users are 'used' to see `<PROJECT-NAME>_SOURCE_DIR` and `<PROJECT-NAME>_BINARY_DIR`. On the other hand, `SCREAM_BIN_DIR` or `SCREAM_BASE_DIR` are less obvious.

Btw, imho, we should _always_ call project (cmake encourages it), so that they are always defined.

components/scream/cime_config/buildnml
```diff
@@ -215,6 +229,25 @@ def buildnml(case, caseroot, compname):
 
             fd.write("/\n")
 
+    # Create input data list
+    input_data_list_file = "{}/Buildconf/scream.input_data_list".format(caseroot)
+    if os.path.exists(input_data_list_file):
+        os.remove(input_data_list_file)
+
+    input_file = scream_input["Atmosphere Driver"]["Initial Conditions"]["Physics GLL"]["Filename"]
```
bartgol, line 237: Sort of unrelated: this is an ugly implementation detail, which I hope to get rid of. Now that we can put both dynamics and physics fields on the same nc file, the initial conditions should be all in the same file, regardless of the grids where fields live. So I hope to soon rearrange the IC section as
```
  Initial Conditions:
    Filename: blah.nc
```
which has the advantage of not requiring buildnml to know what the grids names are, though I suspect we could infer the physics grid name (like "physics GLL" vs "PG2") from the CASE object.

components/scream/cime_config/buildnml
```diff
@@ -106,21 +106,32 @@ def refine_type(entry):
         return entry
 
 ###############################################################################
-def process_leaves(yaml_entry, case):
+def process_leaf(leaf_str, case):
```
bartgol, line 109: Is there a place where you explain briefly what "node" and "leaf" we are talking about? If not, can you add a two liner?

bartgol: Also, a quick explanation of what the 'processing' is, maybe? I think you are processing a single line, replacing cime vars and the if-then-else switches, but might worth explain it (for the sake of maintaining the code).

components/scream/cime_config/buildnml
```diff
@@ -185,16 +196,19 @@ def buildnml(case, caseroot, compname):
         if not os.path.isdir(target):
             os.mkdir(target)
 
-        for item in os.listdir(src):
+        for gitem in [os.path.basename(yaml_tgt), "*.ascii"]:
```
bartgol, line 199: What files are we processing here? Are these the vertical coordinate files (like `scream-128m.ascii')? Or are there others? If it's the vcoord files, we should prob init them differently.

@AaronDonahue I think EAM puts hyai/hyam/hybi/hybm in the nc initial conditions file, right? Perhaps we should do the same, so that we do not need to provide the vertical coordinate file. The latter is what is done in when running homme by itself (with file name passed via namelist); I pulled that detail in, but probably we don't want it, right?

@jgfouca tbc, it's ok to deal with those files like this, for now. I just want to understand if there are potentially other ascii file we are trying to copy here, cause, if not, I know we can remove this once we deal with vert coords directly from the IC netcdf file...

components/scream/cime_config/buildnml
```diff
@@ -215,6 +229,25 @@ def buildnml(case, caseroot, compname):
 
             fd.write("/\n")
 
+    # Create input data list
+    input_data_list_file = "{}/Buildconf/scream.input_data_list".format(caseroot)
```
bartgol, line 233: How is this file generated? I'm guessing CIME does it, but based on what info from our end?

components/scream/src/physics/spa/tests/CMakeLists.txt
```diff
@@ -15,5 +15,4 @@ CreateUnitTest(spa_one_to_one_remap_test "spa_one_to_one_remap_test.cpp" "${NEED
   MPI_RANKS 1 ${SCREAM_TEST_MAX_RANKS}
 )
 
-CONFIGURE_FILE(${CMAKE_CURRENT_SOURCE_DIR}/spa_data_for_testing.nc
-               ${CMAKE_CURRENT_BINARY_DIR}/spa_data_for_testing.nc COPYONLY)
+GetInputFile(init/spa_data_for_testing.nc ${CMAKE_CURRENT_BINARY_DIR}/spa_data_for_testing.nc)
```
bartgol, line 18: I'm all for removing from our repo files needed for cime runs, and force a download (if not present locally). But this file is not for CIME stuff, only for internal testing. I can see it both ways, but my first thought would be "stuff for internal scream tests can live in the repo". It would make developing on local laptops/machines easier, no?

components/scream/tests/uncoupled/CMakeLists.txt
```diff
@@ -7,13 +7,11 @@ if (NOT ${SCREAM_BASELINES_ONLY})
...
-  if (DEFINED SCREAM_SPA_DATA_DIR)
-    add_subdirectory(spa)
-  endif()
+  add_subdirectory(spa)
   if (SCREAM_DOUBLE_PRECISION)
-      add_subdirectory(rrtmgp)
+    add_subdirectory(rrtmgp)
   else()
-      message(STATUS "WARNING: RRTMGP only supported for double precision builds; skipping")
+    message(STATUS "WARNING: RRTMGP only supported for double precision builds; skipping")
```
bartgol, line 14: While at it, can you remove the word "WARNING" from this? CTest catches it as a warning for the config phase, when reporting to cdash. Since there's nothing the user can do to avoid skipping rrtmgp for sp runs, we could simply notify the user:

```
message (STATUS: "RRTMGP only supported for double precision builds. Skipping")
```

components/scream/cmake/ScreamUtils.cmake
```diff
@@ -159,3 +159,27 @@ function(CreateUnitTest test_name test_srcs scream_libs)
...
+  if (NOT SCREAM_CIME_BUILD)
+    if (NOT SCREAM_INPUT_ROOT)
+      message(FATAL_ERROR "No SCREAM_INPUT_ROOT set, cannot try to locate input file ${src_path}")
+    endif()
+    set(full_src_path ${SCREAM_INPUT_ROOT}/atm/scream/${src_path})
+    if (NOT EXISTS ${full_src_path})
+      execute_process(COMMAND ${SCREAM_SRC_DIR}/../scripts/check-input ${SCREAM_INPUT_ROOT} atm/scream/${src_path}
+        RESULT_VARIABLE check_input_status
+        OUTPUT_VARIABLE check_input_output
+        ERROR_VARIABLE  check_input_errput
```
bartgol, line 176: Since you print them both later, you might use the same var for both OUTPUT_VARIABLE and ERROR_VARIABLE. In that case, cmake redirects both output and errput to that var.

bartgol (commented):
Big undertaking Jim! I like it.

My biggest concern regards files needed for local testing. How does this affect day to day development on non-cime machines (like one's laptop)? Also, these files might be subject to frequent changes (I just changed all IC files in a very recent PR). How easy/hard is it to change these files on servers? I guess we can change them on our sems data store, but it requires a SNL guy to do it. And we would not like to do it until the PR is ready...but it can't be ready without testing...which needs the files on the server (a catch22 situation).

jgfouca:
@AaronDonahue , @PeterCaldwell , @bartgol ,

> what about for users who just want to work on a local machine, like a laptop

That should still be doable. If the laptop is known to CIME, then add SCREAM_MACHINE=$machname to either the cmake cache file that you use or the cmake command line. If that is not the case, then add SCREAM_INPUT_ROOT instead in the same manner. You can point it to any directory in your filesystem.

> do we need to move all our input files to the inputdata server on Anvil/Blues in conjunction with this PR?

I've already added all the ones that we use.

> And this PR is totally separate from the problem we've been having lately where hommexx config options are ignored, right?

Yes, not related.

> Also, these files might be subject to frequent changes (I just changed all IC files in a very recent PR). How easy/hard is it to change these files on servers? 

That's a good question. The E3SM policy has been to always require a new filename when modifications are needed. Historically, the CIME input downloader has not been smart enough to know when a file has changed on the remote side, so it was not smart enough to know when a re-download was needed. Recently, CIME has added a chksum capability that can help verify that the downloaded file matches what's on the server. That requires some additional work to set up, so I did not do that yet. If we think we will be frequently updating these files, then it may be worth the effort.

components/scream/CMakeLists.txt
jgfouca: If you look around lines 94-98, variables with identical purpose were already being set. This just removes the duplication.

jgfouca: Oh, I see what you mean. Maybe we should rename the 94-98 ones to use the standard convention? At least until we refactor things to call `project`?

components/scream/cime_config/buildnml
jgfouca: That's fine. It will just require a pretty easy change to buildnml when you change this.

components/scream/cime_config/buildnml
jgfouca: Done. Commit coming soon.

components/scream/cime_config/buildnml
jgfouca: @bartgol , I initially put this in to avoid copying the .nc files in `scream/data` because I was moving towards having SCREAM get them from DIN_LOC_ROOT instead. Now that the .nc files have been remove from our repo, this change is no longer necessary, so I will revert it.

components/scream/cime_config/buildnml
jgfouca: @bartgol , the code writing to this file is right below this line.

components/scream/src/physics/spa/tests/CMakeLists.txt
jgfouca: This is related to your general comments above. This should not make it any harder to do local work; just pick a SCREAM_INPUT_ROOT, stick it in your cache file, and you should be good. The only exception to this would be if you do not currently have an internet connect and you are in needed of downloading some new files.

components/scream/tests/uncoupled/CMakeLists.txt
jgfouca: DOne.

components/scream/cmake/ScreamUtils.cmake
jgfouca: Done.

bartgol:
How does one upload a new file on E3SM servers? Do we have a local server at SNL? I want to avoid the scenario where only one or two people on this project can upload a new file (or new version of it) on the server, especially if we are in the midst of continuous development, involving frequent changes to those files.

jgfouca:
@bartgol ,

> How does one upload a new file on E3SM servers?

It is quite easy if you have a blues/anvil account. You get on blues and copy files to `/lcrc/group/acme/public_html/inputdata/atm/scream`.

I don't know how common it is for SCREAM devs to have such an account. I assume it is pretty common.

jgfouca:
@bartgol , I forgot to mention, we can add our own servers by customizing `cime_config/config_inputdata.xml`

PeterCaldwell:
>> Also, these files might be subject to frequent changes (I just changed all IC files in a very recent PR). How easy/hard is it to change these files on servers?
> 
> That's a good question. The E3SM policy has been to always require a new filename when modifications are needed. Historically, the CIME input downloader has not been smart enough to know when a file has changed on the remote side, so it was not smart enough to know when a re-download was needed. Recently, CIME has added a chksum capability that can help verify that the downloaded file matches what's on the server. That requires some additional work to set up, so I did not do that yet. If we think we will be frequently updating these files, then it may be worth the effort.

My understanding is that the creation date should always be included in the name of any file uploaded to the inputdata repository. Therefore, any time you want to update to use a new file version you need to make a PR to ask to use that new file and anyone with an older version of the code would still be able to run because they would grab the old version of the changed file as appropriate for their repo state. The only downside here is that the inputdata repo could get bloated with old files we don't use. But that's not our problem.

In other words, I don't think we need to worry about chksum or other exotic coding.

jgfouca:
@PeterCaldwell , good point.

jgfouca:
Tested this by hand on Blake and weaver. I think this is ready. @bartgol , do you approve now?

components/scream/CMakeLists.txt
bartgol: Yeah, that was my idea. But if you think we might do another PR soon, where we call project no matter what, then it can wait.

components/scream/cime_config/buildnml
bartgol: LOL!

bartgol (approved):

PR #1392, opened 2022-02-02
Revised theta_l definition to SHOC

bogensch:
This PR assigns a less approximate definition of theta_l (liquid water potential temperature) to SHOC.  The effect is a much reduced work done by the energy fixer in certain columns.  DP-SCREAM simulations shows that this modification (tested for five cases at resolutions from 100 m to 5 km) has very minor impact on the model solution.  The F90 and cpp code have both been modified.

This PR is not b4b, thus SHOC run and compare is expected to fail.  

Fixes https://github.com/E3SM-Project/E3SM/issues/4600

### DIFF
--- a/components/scream/src/physics/shoc/atmosphere_macrophysics.hpp
+++ b/components/scream/src/physics/shoc/atmosphere_macrophysics.hpp
@@ -114,7 +114,7 @@ public:
 
         // Temperature
         const auto& theta_zt = PF::calculate_theta_from_T(T_mid(i,k),p_mid(i,k));
-        thlm(i,k) = theta_zt - (latvap/cpair)*qc(i,k);
+        thlm(i,k) = theta_zt-(theta_zt/T_mid(i,k))*(latvap/cpair)*qc(i,k);
         thv(i,k)  = theta_zt*(1 + zvir*qv(i,k) - qc(i,k));
 
         // Vertical layer thickness
--- a/components/eam/src/physics/cam/shoc.F90
+++ b/components/eam/src/physics/cam/shoc.F90
@@ -3801,7 +3801,8 @@ subroutine update_host_dse(&
 
   do k=1,nlev
     do i=1,shcol
-      temp = (thlm(i,k)+(lcond/cp)*shoc_ql(i,k))/inv_exner(i,k)
+!      temp = (thlm(i,k)+(lcond/cp)*shoc_ql(i,k))/inv_exner(i,k)
+      temp = (thlm(i,k)/inv_exner(i,k))+(lcond/cp)*shoc_ql(i,k)
       host_dse(i,k) = cp*temp+ggr*zt_grid(i,k)+phis(i)
     enddo
   enddo
--- a/components/eam/src/physics/cam/shoc.F90
+++ b/components/eam/src/physics/cam/shoc.F90
@@ -3801,7 +3801,6 @@ subroutine update_host_dse(&
 
   do k=1,nlev
     do i=1,shcol
-!      temp = (thlm(i,k)+(lcond/cp)*shoc_ql(i,k))/inv_exner(i,k)
       temp = (thlm(i,k)/inv_exner(i,k))+(lcond/cp)*shoc_ql(i,k)
       host_dse(i,k) = cp*temp+ggr*zt_grid(i,k)+phis(i)
     enddo
--- a/components/eam/src/physics/cam/shoc_intr.F90
+++ b/components/eam/src/physics/cam/shoc_intr.F90
@@ -609,7 +609,7 @@ end function shoc_implements_cnst
    real(r8) :: minqn, rrho(pcols,pver), rrho_i(pcols,pverp)    ! minimum total cloud liquid + ice threshold    [kg/kg]
    real(r8) :: cldthresh, frac_limit
    real(r8) :: ic_limit, dum1
-   real(r8) :: inv_exner_surf
+   real(r8) :: inv_exner_surf, pot_temp
  
    real(r8) :: wpthlp_sfc(pcols), wprtp_sfc(pcols), upwp_sfc(pcols), vpwp_sfc(pcols)
    real(r8) :: wtracer_sfc(pcols,edsclr_dim)
@@ -777,7 +777,9 @@ end function shoc_implements_cnst
        um(i,k) = state1%u(i,k)
        vm(i,k) = state1%v(i,k)
        
-       thlm(i,k) = state1%t(i,k)*inv_exner(i,k)-(latvap/cpair)*state1%q(i,k,ixcldliq)
+       !thlm(i,k) = state1%t(i,k)*inv_exner(i,k)-(latvap/cpair)*state1%q(i,k,ixcldliq)
+       pot_temp = state1%t(i,k)*inv_exner(i,k)
+       thlm(i,k) = pot_temp-(pot_temp/state1%t(i,k))*(latvap/cpair)*state1%q(i,k,ixcldliq)
        thv(i,k) = state1%t(i,k)*inv_exner(i,k)*(1.0_r8+zvir*state1%q(i,k,ixq)-state1%q(i,k,ixcldliq)) 
  
        tke_zt(i,k) = max(tke_tol,state1%q(i,k,ixtke))
--- a/components/eam/src/physics/cam/shoc_intr.F90
+++ b/components/eam/src/physics/cam/shoc_intr.F90
@@ -777,7 +777,6 @@ end function shoc_implements_cnst
        um(i,k) = state1%u(i,k)
        vm(i,k) = state1%v(i,k)
        
-       !thlm(i,k) = state1%t(i,k)*inv_exner(i,k)-(latvap/cpair)*state1%q(i,k,ixcldliq)
        pot_temp = state1%t(i,k)*inv_exner(i,k)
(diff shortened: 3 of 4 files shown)


oksanaguba:
hi Peter, did you mention a fixer in this context? it seems to me there is no internal energy fixer in shoc yet. however, i tried this change in clubb (only in clubb_intr), the fixer does not look good.

components/scream/src/physics/shoc/atmosphere_macrophysics.hpp
```diff
@@ -113,7 +113,7 @@ class SHOCMacrophysics : public scream::AtmosphereProcess
 
         // Temperature
         const auto& theta_zt = PF::calculate_theta_from_T(T_mid(i,k),p_mid(i,k));
-        thlm(i,k) = theta_zt - (latvap/cpair)*qc(i,k);
+        thlm(i,k) = theta_zt-(theta_zt/T_mid(i,k))*(latvap/cpair)*qc(i,k);
```
oksanaguba, line 116: Where would this code be in e3sm? Maybe that is a change i am missing. 

Also, if you are aware of other changes in e3sm that use thlm definition, please, let me know. So far i used it in 2 places in clubb_intr.

bogensch: You should only need to modify clubb_intr.F90 for consistent CLUBB modification (to my knowledge anyway).  If you can point me to your branch I can double check that your modifications are consistent.

oksanaguba:
es3m branch is oksanaguba/clubb-shflx . Look for NEWTHETAL directive.

oksanaguba:
I fixed the typo you found and got this run, but the fixer does not look good with it (3 ncl plots at the very end):
https://acme-climate.atlassian.net/wiki/spaces/COM/pages/3214606402/CLUBB+SHFLX+fix

Is it correct that we need only 2 modifications in clubb_intr? My understanding is that switching to the new definition of thetal does not affect how surface heat flux is applied in clubb (primitively, because theta, which absorbs temperature tendency from SHFLX, in thetal did not change from the old definition to the new). Could it be that the new definition requires changes in vapor flux or when phase change occurs in clubb, because the new thetal definition has a new multiplier in ql term?

bogensch:
Hi @oksanaguba, I'm a big confused.  On the page you share the right most panel is the one with the new Thlm, right?  If so, that looks to me like a substantial global reduction of the fixer being active and hence a big improvement.  It looks like all three plots you show are on the same color scale.... but maybe I'm misinterpreting something?

adamrher:
Hi All. I'm a bit confused. It looked to me like Peter is fixing the thl calc in shoc_intr:

SHOC old: thl = T/exner - ql*L/cp
SHOC new: thl = 1/exner * (T - ql*L/cp)

But the new version (which is the definition of thetal I am used to) is what we already use in defining thlm in clubb_intr:

```
       thlm(i,k) = ( state1%t(i,k) &
                     - (latvap/cpairv(i,k,lchnk))*state1%q(i,k,ixcldliq) ) &
                   * inv_exner_clubb(i,k)
```

Am I not understanding this?

oksanaguba:
@adamrher just to clarify, e3sm is not using definition of thetal that cesm is using, see https://github.com/E3SM-Project/E3SM/issues/4600 . this is an attempt to use a 'less approximate' definition of thetal in scream (and maybe e3sm, though there is no PR for it).

oksanaguba:
@bogensch  -- i was confused by the red in equatorial areas. i re-plotted with a different color bar, see https://acme-climate.atlassian.net/wiki/spaces/COM/pages/3214606402/CLUBB+SHFLX+fix for green to be zero, and i agree the new run looks the best overall.

however, i still have a question whether we need to touch more code. if it is correct that clubb conserves global column integral of thetal, and thetal now has a new multiplier in term L/cp*QL, then should the code that deal with QFLX and/or phase transitions be affected? thanks!

adamrher:
@oksanaguba thanks for clarifying. That's all from me!

oksanaguba:
@bogensch on confluence, i uploaded climo for the run with new thetal.

bogensch:
@oksanaguba these modifications should be sufficient.

PeterCaldwell:
Hi @oksanaguba - is there a reason why you haven't approved this PR? It would be useful to get it integrated soon...

oksanaguba:
@PeterCaldwell sorry did not notice that i am a reviewer. if this is a case, could @bogensch please point me to 1) where phase transitions are, i still would like to trace how conservative computations regarding the ql term in thetal definitions are dealt with and 2) where in shoc is the local energy fixer? 

thanks!

oksanaguba:
Copying Peter B's message from slack:
```
 In SHOC the cloud liquid condensate is computed in the “shoc_assumed_pdf” routine in components/eam/src/physics/cam/shoc.F90. The energy fixer is done in the three routines “update_host_dse”, “shoc_energy_integrals”, “shoc_energy_fixer” in the same file.
```
I could not follow the code in shot_assumed_pdf, unfortunately. I typed a simple example where (I think) introducing an Exner factor in the ql term in theta, like in this PR, affects conservation. Note that I am only assuming which conservation show operates with, it could be a column conservation instead. Correct me if this example is not relevant, please:
Screen Shot 2022-02-09 at 1 21 57 PM

oksanaguba:
I looked at the energy fixer and did not see anything. it is using dse instead of enthalpy, but since only difference of TE matters and since (I think) shoc/clubb does not change geopotential, gz term cancels out.

PeterCaldwell:
Ok, here's my summary of this PR: 
1. The goal is to change from the "worse" to the "better" approximation of theta_l described in https://glossary.ametsoc.org/wiki/Liquid_water_potential_temperature.
2. Because SHOC exclusively uses theta_l internally, conversion between dse and theta_l is only needed to pass theta_l **into** SHOC and to extract dse**out of** SHOC. 
3. Passing theta_l **into** SHOC is handled in shoc_intr.F90 for F90 and atmosphere_macrophysics.hpp for C++
4. Passing dse **out of** SHOC is handled in shoc.F90 for F90 and shoc_update_host_dse_impl.hpp for C++. It would be conceptually preferable if dse was converted within the shoc **interface** layer to make my assertion in bullet 2 that "SHOC exclusively uses theta_l internally" more clear, but note that dse is calculated in `update_host_dse` on line https://github.com/E3SM-Project/scream/blob/29844a4b9b7605cf8c88b233159253cb2c1c972c/components/eam/src/physics/cam/shoc.F90#L548 , which is **after** comments clearly saying that SHOC has already ended. Note as well that the functions after this point either use use dse exclusively or just compute diagnostics. I think it would be better to put this stuff in the shoc_interface layer because we want actual parameterizations to be model independent and dse is EAM-specific... but that's clearly a topic for another PR. 
5. @oksanaguba is concerned about the implications of this PR on energy conservation. Since conversion to theta_l, SHOC calculations, then conversion back to dse are all performed before the energy fixer is called, and since the energy fixer is still in the same dse units it started in at the beginning of this PR, I dont' see how this PR could affect the fixer except by changing how strong its impact is. The proper response to this last concern is to just empirically check whether fixer magnitude is increasing. I think @bogensch told me that the whole motivation for this PR was to remove anomalously large fixer tendencies, so I think we're good there. This is not to say that energy conservation and energy definition aren't   real problems, just that this PR isn't the place to tackle that - we need a big campaign.

oksanaguba (approved):

oksanaguba:
@PeterCaldwell thanks for you explanation.

I agree that the goal of this PR is to use a better definition of thetal, not energy conservation. It wasn't my intention to evaluate energy conservation here, I asked a fixer question to follow the code, mostly, for my own benefit. 

My other question, about phase changes, is directly related to the definition of thetal in shoc. My question was this: if we are changing definition of thetal in shoc, does it mean that calculations around a latent heat release are affected by the new definition? It may be, as Peter B. said before, that internally clubb and shoc already use the proper definition of thetal. I must add, I forgot about his comment when I wrote my question about phase changes. If there is nothing to check there, it would still be of a great benefit for me if you guys point to me where these phase changes happen in shoc.

Also, from a reviewer point of view, if would be great if we could see in this PR description latlon plots of how the shoc fixer is reduced. There was no quantifiable data presented. When I ran the same changes in e3sm, by an eye ball norm, it may have been a case that the fixer is better, but there are questionable areas at the equator. 

However, I do not mean to delay this PR, so I approved it. In future, feel free to override my comments regarding any PRs in scream.

thanks again!

PeterCaldwell:
Thanks @oksanaguba ! Good point that it would be useful to include a graphic or two in this PR illustrating the impact on the SHOC energy fixer of making this change. 

Also, I think I finally understand your question about phase change. My thinking (illustrated in my framing of the PR above) is that SHOC is framed entirely in theta_L so questions about how we convert back to dse are irrelevant within the main SHOC calculations. Kind of like if you live entirely in one country you don't ever need to worry about translating your speech to another language. This would be true if saturation mixing ratio was computed from theta_L, but I think it can only be calculated from T... which requires another translation within the SHOC code. I'm also not sure how this is done in SHOC. It certainly wasn't updated in this PR to be consistent with the new theta_L definition. @bogensch - is there an inconsistency here?

AaronDonahue:
Just rebased and set this to RETEST

jgfouca:
@AaronDonahue , cprnc seems to be working fine on today's master, so this is likely an issue of this branch being out of date. I would expect your rebase and retest to work.

AaronDonahue:
ahhh that did it.  DIFFs across the field after rebase.  I think we can confidently manually merge now.

PR #1397, opened 2022-02-02
Compute and expose surface radiative fluxes

brhillman:
Compute and expose surface radiative fluxes from RRTMGP to the Atmosphere Driver. The surface models expect broadband UV+visible (vis) and near infrared (nir) radiative fluxes, but we neglected to implement these fluxes in our initial RRTMGP-to-AD interface, and hence we have been neglecting to pass surface radiative fluxes from the atmosphere to the surface models. Modifications here compute these special broadband quantities from the band-resolved radiative fluxes computed by RRTMGP, and add these fluxes to the FieldManager so that the AD can then export them to the surface models (exporting of those fluxes will need to be handled in a subsequent PR, this PR only deals with the radiation interface responsibility of adding the computed fluxes to the FieldManager). In order to accomplish this, the following modifications are made:
  - Use the RRTMGP `FluxesByband` flux class instead of the `FluxesBroadband` class, which includes not just the broadband integrated fluxes but also the band-resolved fluxes computed by RRTMGP
  - Register YAKL arrays in the radiation buffer to hold the band-resolved fluxes as well as computed surface fluxes
  - Pass band-resolved fluxes back out of `rrtmgp_main` to be used by `run_impl`
  - Add a function to compute surface fluxes from the band-resolved 3D fluxes
  - Register fields in the FM for the new surface fluxes, and copy computed surface fluxes into the FM in `run_impl`

### DIFF
--- a/components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
+++ b/components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
@@ -40,7 +40,7 @@ void RRTMGPRadiation::set_grids(const std::shared_ptr<const GridsManager> grids_
   kgkg.set_string("kg/kg");
   auto m3 = m * m * m;
   auto Wm2 = W / m / m;
-  Wm2.set_string("Wm2");
+  Wm2.set_string("W/m2)");
   auto nondim = m/m;  // dummy unit for non-dimensional fields
   auto micron = m / 1000000;
 
--- a/components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
+++ b/components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
@@ -90,6 +90,21 @@ void RRTMGPRadiation::set_grids(const std::shared_ptr<const GridsManager> grids_
   add_field<Computed>("LW_flux_up", scalar3d_layout_int, Wm2, grid->name(), ps);
   add_field<Computed>("LW_flux_dn", scalar3d_layout_int, Wm2, grid->name(), ps);
 
+  // Translation of variables from EAM
+  // --------------------------------------------------------------
+  // EAM name | EAMXX name       | Description
+  // --------------------------------------------------------------
+  // soll       sfc_flux_dir_nir   solar near-IR direct flux
+  // sols       sfc_flux_dir_vis   solar UV/visible direct flux
+  // solld      sfc_flux_dif_nir   solar near-ID diffuse flux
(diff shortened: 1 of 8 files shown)


components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
```diff
@@ -90,14 +90,31 @@ void RRTMGPRadiation::set_grids(const std::shared_ptr<const GridsManager> grids_
...
 
+  // Translation of variables from EAM
+  // --------------------------------------------------------------
+  // EAM name | EAMXX name       | Description
+  // --------------------------------------------------------------
+  // soll       sfc_flux_dir_nir   solar near-IR direct flux
+  // sols       sfc_flux_dir_vis   solar UV/visible direct flux
+  // solld      sfc_flux_dif_nir   solar near-ID diffuse flux
+  // solsd      sfc_flux_dif_vis   solar UV/visible diffuse flux
+  // --------------------------------------------------------------
```
tcclevenger, line 101: Should we add a description for `sfc_flux_sw_net` and `sfc_flux_lw_dn`? If this PR passes test and we don't want to delay getting it in, I can add it to the SC PR that follows this. 

```
// EAM name | EAMXX name       | Description
// netsw           sfc_flux_sw_net             ?
// flwds            sfc_flux_lw_dn               ?
```

brhillman: Oh yeah, I had it in here but took it out because I moved where I set these to the kokkos parallel for block below. I'll add a comment back in.

brhillman: Okay done.

tcclevenger (approved):

brhillman:
Know the cause of the Weaver fail, testing my fix now before pushing and adding RETEST.

tcclevenger:
Seems that everything passes.. I was going to look at the AT console output.

tcclevenger:
Looks like one of the weaver build has an error
```
jenkins/weaver/workspace/SCREAM_PullRequest_Autotester_Weaver/2576/scream/components/scream/src/physics/rrtmgp/scream_rrtmgp_interface.cpp(170): error: calling a __device__ function("atomicAdd") from a __host__ __device__ function("operator()") is not allowed
```

tcclevenger (approved):

tcclevenger (approved):

PR #1399, opened 2022-02-03
SCREAM: refactor of field property checks

bartgol:
This PR refactors a bit the concept of property checks. The biggest difference is that a property check is no longer linked to a single field. One can implement a property check involving 2+ fields. Also, make the property checks explicitly define what field they operate on, and which ones of those can be repaired. If none can be repaired, the repair function would throw.

@AaronDonahue This PR should allow you to define a "WaterMass" check. The new framework returns an enum for the `check()` method, which can be `Pass`, `Fail`, or `Warn`. The first two do what you expect them to do, while the latter prints a warning, but does not cause a crash. You could add two property checks for water mass conservation, the first with strict tolerance and returning Warn if mass is not conserved, while the second with a lose tolerance but returning a Fail if mass is not conserved.

Note: I did not implement any logic for the AD to set "global" property checks, that could be part of a follow up PR. For instance, we could have the AD build the "water mass" check, and set it in the atm process group, which would add it to each of the processes in the group. This would allow the check to be run after each atm process is run. Note that a property check in an atm process cannot run the 'repair' method, unless all the 'repairable' fields are computed by the atm process. This is to avoid bypassing the dag, having a global property check (like water mass conservation) updating a field that wasn't updated by the atm process.


components/scream/src/physics/cld_fraction/atmosphere_cld_fraction.cpp
```diff
@@ -55,9 +55,9 @@ void CldFraction::set_grids(const std::shared_ptr<const GridsManager> grids_mana
 void CldFraction::initialize_impl (const RunType /* run_type */)
 {
   // Set property checks for fields in this process
-  auto frac_interval_check = std::make_shared<FieldWithinIntervalCheck>(0,1);
-  add_property_check<Computed>(get_field_out("cldfrac_ice").get_header().get_identifier(),frac_interval_check);
-  add_property_check<Computed>(get_field_out("cldfrac_tot").get_header().get_identifier(),frac_interval_check);
+  using FWIC = FieldWithinIntervalCheck;
+  add_post_run_property_check<FWIC>(get_field_out("cldfrac_ice"),0,1);
```
jeff-cohere, line 59: I like this economical syntax, which feeds the arguments of `add_x_property_check` to the specified property check constructor. Just curious: do you think anyone who uses it might find it confusing, as opposed to passing an instance as the second argument instead?

bartgol: I am not sure. I provided both constructors, so people can decide for themselves. I like the economy version too, so I use that when I'm the one coding. But I guess the two steps version leaves less room for interpretation, so some might like that.

components/scream/src/share/atm_process/atmosphere_process.cpp
```diff
@@ -158,90 +158,64 @@ void AtmosphereProcess::set_computed_group (const FieldGroup& group) {
...
-          }
-        }
-      }
+void AtmosphereProcess::run_property_checks_post () const {
+  constexpr auto Pass = PropertyCheck::CheckResult::Pass;
+  constexpr auto Warn = PropertyCheck::CheckResult::Warn;
+  constexpr auto Fail = PropertyCheck::CheckResult::Fail;
+
+  // Run all pre-condition property checks
+  for (const auto& pc : m_property_checks_pre) {
```
jeff-cohere, line 198: This should probably be `m_property_checks_post`, right?

bartgol: Argh, good catch!

components/scream/src/share/atm_process/atmosphere_process.cpp
```diff
@@ -412,6 +388,49 @@ get_internal_field(const std::string& field_name) const {
   return get_internal_field_impl(field_name);
 }
 
+void AtmosphereProcess::
+add_property_check (const prop_check_ptr& pc)
```
jeff-cohere, line 392: how about `add_invariant_property_check` or `add_inv_property_check`, to indicate that this checks an invariant instead of a pre/post condition (while we're using design-by-contract jargon)?

bartgol: I think that would be the most appropriate word. Do you think most people would get the meaning (i.e., invariant=pre+post)?

jeff-cohere: I think if they looked up "invariant" and "precondition", they'd find the definition. An invariant is a predicate that holds for the entire lifetime of an object.

components/scream/src/share/property_checks/check_and_repair_wrapper.hpp
```diff
@@ -0,0 +1,90 @@
...
+
+  // Constructor(s)
+  explicit CheckAndRepairWrapper (base_ptr_type check,
+                                  base_ptr_type repair = nullptr)
+   : m_check  (check)
+   , m_repair (repair)
+  {
+    // repair can be null, but check cannot
+    EKAT_REQUIRE_MSG (m_check,
+      "Error in CheckAndRepair constructor!\n"
```
jeff-cohere, line 36: Might want to say `CheckAndRepairWrapper` here

components/scream/src/share/property_checks/field_lower_bound_check.hpp
```diff
@@ -0,0 +1,43 @@
...
+{
+
+// Convenience implementation of check for interval [U,-\infty). The class
+// inherits from FieldWithinIntervalCheck, and sets lower bound to -infinity
+class FieldLowerBoundCheck: public FieldWithinIntervalCheck {
+public:
+  // Constructor with lower bound. By default, this property check
+  // can repair fields that fail the check by overwriting nonpositive values
+  // with the given lower bound. If can_repair is false, the check cannot
+  // apply repairs to the field.
```
jeff-cohere, line 18: What about `allow_failures`?

components/scream/src/share/property_checks/field_nan_check.cpp
```diff
@@ -87,20 +96,22 @@ bool FieldNaNCheck::check_impl(const Field& field) const {
       }
       break;
     default:
-      EKAT_ERROR_MSG ("Error! Unsupported field rank.\n");
+      EKAT_ERROR_MSG ("Error! Unsupported f rank.\n");
```
jeff-cohere, line 99: Could say `field` instead of `f`. Also, why not print the field's name and its (unsupported) rank?

components/scream/src/share/property_checks/field_nan_check.cpp
```diff
@@ -87,20 +96,22 @@ bool FieldNaNCheck::check_impl(const Field& field) const {
...
-      return check_impl<int>(field);
+      return check_impl<int>();
     case DataType::FloatType:
-      return check_impl<float>(field);
+      return check_impl<float>();
     case DataType::DoubleType:
-      return check_impl<double>(field);
+      return check_impl<double>();
     default:
       EKAT_ERROR_MSG ("Error! Field data type not supported.\n");
```
jeff-cohere, line 116: Maybe print the field's name here too?

components/scream/src/share/property_checks/field_upper_bound_check.hpp
```diff
@@ -1,28 +1,30 @@
...
   // Constructor with upper bound. By default, this property check
   // can repair fields that fail the check by overwriting nonpositive values
   // with the given upper bound. If can_repair is false, the check cannot
   // apply repairs to the field.
-  explicit FieldUpperBoundCheck (const double upper_bound,
-                                 bool can_repair = true) : 
-    FieldWithinIntervalCheck(-std::numeric_limits<double>::max(), upper_bound, can_repair)
+  FieldUpperBoundCheck (const Field& field,
+                        const double upper_bound,
+                        const bool can_repair = true,
```
jeff-cohere, line 21: If you decide to describe allow_failures in the comments for the lower bound check constructor, do it here too.

components/scream/src/share/property_checks/field_within_interval_check.cpp
```diff
@@ -83,30 +102,34 @@ bool FieldWithinIntervalCheck::check_impl (const Field& field) const
       }
       break;
     default:
-      EKAT_ERROR_MSG ("Error! Unsupported field rank.\n");
+      EKAT_ERROR_MSG ("Error! Unsupported f rank.\n");
```
jeff-cohere, line 105: Same comment here as above for the lower bound check.

components/scream/src/share/property_checks/field_within_interval_check.cpp
```diff
@@ -83,30 +102,34 @@ bool FieldWithinIntervalCheck::check_impl (const Field& field) const
...
+      return check_impl<int>();
     case DataType::FloatType:
-      return check_impl<float>(field);
+      return check_impl<float>();
     case DataType::DoubleType:
-      return check_impl<double>(field);
+      return check_impl<double>();
     default:
-      EKAT_ERROR_MSG ("Error! Field data type not supported.\n");
+      EKAT_ERROR_MSG ("Error! f data type not supported.\n");
```
jeff-cohere, line 121: field instead of f?

components/scream/src/share/property_checks/field_within_interval_check.cpp
```diff
@@ -189,29 +212,31 @@ void FieldWithinIntervalCheck::repair_impl(Field& field) const
       }
       break;
     default:
-      EKAT_ERROR_MSG ("Error! Unsupported field rank.\n");
+      EKAT_ERROR_MSG ("Error! Unsupported f rank.\n");
```
jeff-cohere, line 215: See above comments.

components/scream/src/share/property_checks/field_within_interval_check.cpp
```diff
@@ -189,29 +212,31 @@ void FieldWithinIntervalCheck::repair_impl(Field& field) const
...
-      repair_impl<float>(field);
+      repair_impl<float>();
       break;
     case DataType::DoubleType:
-      repair_impl<double>(field);
+      repair_impl<double>();
       break;
     default:
-      EKAT_ERROR_MSG ("Error! Field data type not supported.\n");
+      EKAT_ERROR_MSG ("Error! f data type not supported.\n");
```
jeff-cohere, line 239: field instead of f

components/scream/src/share/property_checks/field_within_interval_check.hpp
```diff
@@ -1,74 +1,66 @@
...
-  // Overrides.
-
-  // The name of the field check
+  // The name of the property check
   std::string name () const override {
     // NOTE: std::to_string does not do a good job with small numbers (like 1e-9).
     std::stringstream ss;
-    ss << "Within Interval [" << m_lower_bound << ", " << m_upper_bound << "] Check";
+    ss << fields().front().name()
+       << " within wnterval [" << m_lower_bound << ", " << m_upper_bound << "]";
```
jeff-cohere, line 37: wnterval -> interval

components/scream/tests/coupled/dynamics_physics/homme_shoc_cld_p3_rrtmgp/input.yaml
```diff
@@ -20,7 +20,7 @@ Atmosphere Processes:
   Schedule Type: Sequential
   Process 0:
     Process Name: Homme
-    Enable Output Field Checks: false
+    Enable Pre Run Property Checks: false
```
jeff-cohere, line 23: Now that I think of it: should we call these `Precondition Property Checks`? And `Postcondition Property Checks`?

jeff-cohere (approved):
Looks good, Luca! I left several remarks for you to consider, but nothing very significant.

components/scream/src/share/atm_process/atmosphere_process.cpp
bartgol: Then I think the fully spelled out `add_invariant_property_check` might be better, since `inv` might not easily hint to "invariant"...

components/scream/src/share/property_checks/field_lower_bound_check.hpp
bartgol: `allow_failures` only applies to the check phase. If f<0, check would return Warn, but repair would still attempt to repair the field.

In a way, allow_failure is more for the benefit of AtmosphereProcess. Maybe I should remove this enum from the PropertyCheck class, and add `allow_failure` to the methods `add_x_property_chec` of AtmosphereProcess?

components/scream/src/share/property_checks/field_nan_check.cpp
bartgol: Ah, yeah. I did a find+replace earlier, but accidentally edited also comments. :)

components/scream/src/share/property_checks/field_lower_bound_check.hpp
jeff-cohere: That sounds good to me.

jeff-cohere (approved):
Excellent!

Issue #1255, opened 2021-10-12
Turn SurfaceCoupling into an atmosphere process


bartgol:
This is not required, but would have a few benefits:

- The DAG construction would not have to treat the SC as a special case
- The initialization of atm inputs would be more linear, without the need to check if an atm input is by chance imported by the SC.
- The SC export method would not have to check whether this is an AD run.

PeterCaldwell:
I'm psyched about this change. I think it would make the code much more logical/intuitive for new users.

oksanaguba:
Is it the same as removing application of QFLX and SHFLX from clubb/shoc? I would vote for such change, too. it is a better design to separate it from clubb and would give us more control. not sure if fluxes need to be sub cycled and diffused by clubb if they are too strong.

bartgol:
@oksanaguba This is unrelated to that, I think. This issue is only a perspective change of the software design. That said, we can always change what happens inside the SC class, to incorporate more/less stuff.

PeterCaldwell:
Luca is right - this is just about software engineering. I'm not sure surface fluxes can be separated from clubb/shoc since they serve as the lower boundary condition for those schemes.

whannah1:
Does this mean that surface fluxes would be calculated in the atmosphere? Or would the surface component models still own these calculations?

tcclevenger:
From what I see so far, nothing should really change from a surface model perspective. Currently inside 
`scream/src/mct-coupling/atm_comp_mct.F90` function `atm_init_mct()`, we initialize the atm (`scream_init_atm()`) and then initialize the current `SurfaceCoupling` class (`scream_setup_surface_coupling()`) where we pass in data pointers from the surface model to import/export to/from. Then scream's `AtmosphereDriver` is responsible for call `do_import()` and `do_export()` before an after each atm process is called. 

All this issue would change is that we would not have a separate `SurfaceCoupling` class, and instead it would be itself an atm process, and therefore setup is taken care of in `scream_init_atm()` and calling import/export function is still done by the driver, but done by calling the atm processes, not independently calling the `SurfaceCoupling` class.

PR #1400, opened 2022-02-03
Change name of surf_latent_heat to surf_evaporation to match coupler.

AaronDonahue:
This commit change the name of the field manager variable
surf_latent_heat to be surf_evaporation.  Which is what the actual
variable passed by the component coupler represents.

This commit also changes the units of the variable in the FM to
match the units in the component coupler.

### DIFF
--- a/components/scream/data/scream_output.yaml
+++ b/components/scream/data/scream_output.yaml
@@ -41,7 +41,7 @@ Fields:
       - p_mid
       - phis
       - pseudo_density
-      - surf_evaporation
+      - surf_evap
       - surf_mom_flux
       - surf_sens_flux
       - cldfrac_ice
--- a/components/scream/data/scream_output.yaml
+++ b/components/scream/data/scream_output.yaml
@@ -41,7 +41,7 @@ Fields:
       - p_mid
       - phis
       - pseudo_density
-      - surf_latent_flux
+      - surf_evaporation
       - surf_mom_flux
       - surf_sens_flux
       - cldfrac_ice
--- a/components/scream/src/control/surface_coupling.cpp
+++ b/components/scream/src/control/surface_coupling.cpp
@@ -112,7 +112,7 @@ register_import(const std::string& fname,
 
     // For import fluxes, we must change the sign as cpl and atm interprete the direction differently.
     if (fname == "surf_mom_flux"    || fname == "surf_sens_flux" ||
-        fname == "surf_latent_flux" || fname == "surf_lw_flux_up") {
+        fname == "surf_evaporation" || fname == "surf_lw_flux_up") {
       m_cpl_scream_sign_change_host(m_num_scream_imports) = -1;
     } else {
       m_cpl_scream_sign_change_host(m_num_scream_imports) = 1;
--- a/components/scream/src/control/surface_coupling.cpp
+++ b/components/scream/src/control/surface_coupling.cpp
@@ -112,7 +112,7 @@ register_import(const std::string& fname,
 
     // For import fluxes, we must change the sign as cpl and atm interprete the direction differently.
     if (fname == "surf_mom_flux"    || fname == "surf_sens_flux" ||
-        fname == "surf_evaporation" || fname == "surf_lw_flux_up") {
+        fname == "surf_evap"        || fname == "surf_lw_flux_up") {
       m_cpl_scream_sign_change_host(m_num_scream_imports) = -1;
     } else {
       m_cpl_scream_sign_change_host(m_num_scream_imports) = 1;
--- a/components/scream/src/control/tests/sc_tests.cpp
+++ b/components/scream/src/control/tests/sc_tests.cpp
@@ -257,7 +257,7 @@ TEST_CASE ("recreate_mct_coupling")
 
   // Create import fields
   const auto nondim = Units::nondimensional();
-  FID surf_evaporation_id ("surf_evaporation",   scalar2d_layout, kg/(m*m)/s, grid_name);
+  FID surf_evap_id        ("surf_evap",   scalar2d_layout, kg/(m*m)/s, grid_name);
   FID surf_sens_flux_id   ("surf_sens_flux",     scalar2d_layout, W/(m*m), grid_name);
   FID surf_mom_flux_id    ("surf_mom_flux",      vector2d_layout, W/(m*m), grid_name);
   FID sfc_alb_dir_vis_id  ("sfc_alb_dir_vis",    scalar2d_layout, nondim,  grid_name);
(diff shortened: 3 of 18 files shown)


PeterCaldwell:
Looks good to me. I prefer the name `surf_evap` because I think our goal is to be as short as possible so long as our meaning is unambigious... but I don't care too much.

PeterCaldwell:
I'd approve, but I want Peter B to get a chance to look at it before it auto-merges.

AaronDonahue:
> Looks good to me. I prefer the name `surf_evap` because I think our goal is to be as short as possible so long as our meaning is unambigious... but I don't care too much.

I can make that change, super easy, and we might as well do it right the first time.

bogensch:
Looks good to me!

bogensch (approved):

AaronDonahue:
@bartgol (pinging you because you wrote the test) , this is failing the AT for just the suite of `model_restart` tests.  I'm thinking this is because those tests expect the old variable, `surf_latent_heat` to be present in the netCDF init file and it's missing.  Does this sound correct?  If so then I just need to update the name in that file I should be good to go.

bartgol:
I think you're right, but let me hop on the testbeds, and manually inspect the cprnc output.

bartgol:
Yeah

```
  PIO ERROR: could not find variable surf_evap in file     homme_physics_ne2np4.nc
```

AaronDonahue:
Okay, so we need to change the file.  I'm going to hold off on that until Jims big refactor PR goes in #1391 .  Then I can update the file on blues and change its name so it will be automatically downloaded.

AaronDonahue:
Chatted with @jgfouca on Slack and we decided that I'll make the init condition file change here and he can take care of the changes in his PR.

AaronDonahue:
@bogensch , do you mind re-approving.  I think this needs to be reapproved after I made a new commit.

bogensch (approved):

AaronDonahue:
This one is pending a rebase and a file upload to blues, since we need a different initial condition file for the model_restart tests.

AaronDonahue:
I don't have easy access to mappy so I'm speculating here, but I think that this PR is failing for the SCREAMv1 CIME based runs because the initial condition files those cases are using have the variable `latent_heat_flux` instead of `surf_evap`.  That being said, isn't this information coming from the surface coupler?  I wouldn't expect SCREAM to need to grab this info from the initial condition file anyway.  The easy fix would be to change the name of `latent_heat_flux` to `surf_evap` in our two initial condition files for SCREAMv1.  But I wanted to check on the behavior in general in case it is something we expect.

@bartgol or @tcclevenger , do you have any thoughts?  Also, if one of you have a chance could you hop on mappy and check that my speculation is correct?

tcclevenger:
@AaronDonahue I can hop on mappy real quick and check.

tcclevenger:
The error is 
```
PIO ERROR: could not find variable surf_evap in file /sems-data-store/ACME/inputdata/atm/scream/init/init_ne4np4.nc
```

Does that `.nc` file need to be updated? I don't think this has to do with SC since it does not deal with `latent_heat_flux/surf_evap`.

tcclevenger:
Also, `latent_heat_flux/surf_evap` isn't in the SC. We do have `surf_latent_flux` (SCREAM name)/`Faxx_evap` (surface model name). I'm assuming those are not the same?

AaronDonahue:
> The error is
> 
> ```
> PIO ERROR: could not find variable surf_evap in file /sems-data-store/ACME/inputdata/atm/scream/init/init_ne4np4.nc
> ```
> 
> Does that `.nc` file need to be updated?

Yeah, that was my theory.  I had to update the homme+physics model restart test to have `surf_evap` instead of `surf_latent_flux`

AaronDonahue:
> > The error is
> > ```
> > PIO ERROR: could not find variable surf_evap in file /sems-data-store/ACME/inputdata/atm/scream/init/init_ne4np4.nc
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Does that `.nc` file need to be updated?
> 
> Yeah, that was my theory. I had to update the homme+physics model restart test to have `surf_evap` instead of `surf_latent_flux`

@tcclevenger , yes sorry, I meant `surf_latent_flux`.

AaronDonahue:
> Also, `latent_heat_flux/surf_evap` isn't in the SC. We do have `surf_latent_flux` (SCREAM name)/`Faxx_evap` (surface model name). I'm assuming those are not the same?

Yes, @tcclevenger , sorry, I mean `surf_latent_flux`.  I guess the question still applies, if `surf_latent_flux` is one of the import variables do we know why SCREAM is even looking for it in the first place.  Note that with this PR it has been changed to `surf_evap` and that is what it can't find in the IC file.

tcclevenger:
So the mystery is that something that is specific to the SCREAMv1 run is asking for `surf_evap`? I'm running in debug to see if I can get a more useful error that points me to where this is begin queried.

tcclevenger:
When we register an import, we need to call
```
SurfaceCoupling::register_import(fname) {
  .
  . 
  .
  Field field = m_field_mdr->get_field(fname)
  .
  .
  .
}
```
I'm guessing this is what is being called.

bartgol:
> > Also, `latent_heat_flux/surf_evap` isn't in the SC. We do have `surf_latent_flux` (SCREAM name)/`Faxx_evap` (surface model name). I'm assuming those are not the same?
> 
> Yes, @tcclevenger , sorry, I mean `surf_latent_flux`. I guess the question still applies, if `surf_latent_flux` is one of the import variables do we know why SCREAM is even looking for it in the first place. Note that with this PR it has been changed to `surf_evap` and that is what it can't find in the IC file.

Are you saying that scream is looking in the IC file for the imported fields? Yeah, that sounds wrong...

AaronDonahue:
Yeah, I think it probably doesn't use them.  But looks for them all the same.  We will want to fix that.

bartgol:
It's probably an issue with the DAG. I think addressing #1255 would fix this, since the SC imported fields would not appear as inputs for the ATM group anymore.

Note: we do print a correct DAG, including the surface coupling, but when we check which fields need an Initial Condition ([here](https://github.com/E3SM-Project/scream/blob/32ffa4b57f28591ad208b9aa61198c7e7ccd43a0/components/scream/src/control/atmosphere_driver.cpp#L557)), we do not keep into account the surface coupling...

tcclevenger:
@bartgol Yeah, that's what I figured I would move to next. Figured out why p3 wasn't scaling (loading decently large tables on host, which can be done during initialization).

AaronDonahue:
Closing because this branch is stale and #1751 does the same thing.

PR #1401, opened 2022-02-03
Change output netcdf time units to days

PeterCaldwell:
This small PR changes netcdf output to use units "days since <start of simulation>". It also changes global attributes to match EAM conventions for variables that always have the same text.

### DIFF
--- a/components/scream/src/share/io/scream_output_manager.cpp
+++ b/components/scream/src/share/io/scream_output_manager.cpp
@@ -213,7 +213,7 @@ void OutputManager::run(const util::TimeStamp& timestamp)
       register_dimension(filename,"time","time",0);
 
       // Register time as a variable.
-      auto time_units="s since " + m_case_t0.get_date_string() + " " + m_case_t0.get_time_string();
+      auto time_units="days since " + m_case_t0.get_date_string() + " " + m_case_t0.get_time_string();
       register_variable(filename,"time","time",time_units,1,{"time"},  PIO_REAL,"time");
 
       // Make all output streams register their dims/vars
@@ -249,7 +249,7 @@ void OutputManager::run(const util::TimeStamp& timestamp)
     }
 
     // Update time and nsteps in the output file
-    pio_update_time(filename,timestamp.seconds_from(m_case_t0));
+    pio_update_time(filename,timestamp.seconds_from(m_case_t0)/86400.0);
     if (m_is_model_restart_output) {
       // Only write nsteps on model restart
       set_int_attribute_c2f(filename.c_str(),"nsteps",timestamp.get_num_steps());
@@ -258,7 +258,7 @@ void OutputManager::run(const util::TimeStamp& timestamp)
 
   // Run the output streams
   for (auto& it : m_output_streams) {
-    // Note: filename might referencing an invalid string, but it's only used
+    // Note: filename might reference an invalid string, but it's only used
     //       in case is_write_step=true, in which case it will *for sure* contain
     //       a valid file name.
     it->run(filename,is_write_step,m_output_control.nsteps_since_last_write);
--- a/components/scream/src/share/io/scream_output_manager.cpp
+++ b/components/scream/src/share/io/scream_output_manager.cpp
@@ -213,7 +213,8 @@ void OutputManager::run(const util::TimeStamp& timestamp)
       register_dimension(filename,"time","time",0);
 
       // Register time as a variable.
-      register_variable(filename,"time","time",1,{"time"},  PIO_REAL,"time");
+      auto time_units='s since '+m_case_t0.get_date_string()+' '+m_case_t0.get_time_string();
+      register_variable(filename,"time","time",time_units,1,{"time"},  PIO_REAL,"time");
 
       // Make all output streams register their dims/vars
       for (auto& it : m_output_streams) {
@@ -229,10 +230,7 @@ void OutputManager::run(const util::TimeStamp& timestamp)
       if (is_checkpoint_step) { 
         set_int_attribute_c2f (filename.c_str(),"avg_count",m_output_control.nsteps_since_last_write);
       }
-      auto t0_date = m_case_t0.get_date()[0]*10000 + m_case_t0.get_date()[1]*100 + m_case_t0.get_date()[2];
-      auto t0_time = m_case_t0.get_time()[0]*10000 + m_case_t0.get_time()[1]*100 + m_case_t0.get_time()[2];
-      set_int_attribute_c2f(filename.c_str(),"start_date",t0_date);
-      set_int_attribute_c2f(filename.c_str(),"start_time",t0_time);
+
       filespecs.is_open = true;
     }
 
--- a/components/scream/src/share/io/scream_output_manager.cpp
+++ b/components/scream/src/share/io/scream_output_manager.cpp
@@ -249,7 +249,7 @@ void OutputManager::run(const util::TimeStamp& timestamp)
     }
 
     // Update time and nsteps in the output file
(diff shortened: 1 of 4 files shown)


components/scream/src/share/io/scream_output_manager.cpp
```diff
@@ -213,7 +213,8 @@ void OutputManager::run(const util::TimeStamp& timestamp)
       register_dimension(filename,"time","time",0);
 
       // Register time as a variable.
-      register_variable(filename,"time","time","s",1,{"time"},  PIO_REAL,"time");
+      auto time_units="days since " + m_case_t0.get_date_string() + " " + m_case_t0.get_time_string();
+      register_variable(filename,"time","time",time_units,1,{"time"},  PIO_REAL,"time");
```
bartgol, line 217: I don't think it's correct to do this. If you do this, then you should modify how the timestamp is passed to scorpio for writing. Right now we have (at line 251 of `scream_output_manager.cpp`):
```
     pio_update_time(filename,timestamp.seconds_from(m_case_t0));
```
If you wish time units to be in days since case start, you should probably change that to
```
     pio_update_time (filename, timestamp.seconds_from(m_case_t0)/3600);
```
Or, maybe better, add a `days_from` method to TimeStamp.

bartgol: Also: are you assuming that `days from t0` will always be an integer? Or do you want to accommodate fractional values?

AaronDonahue: days from t0 would definitely have to be a double.  if we have output every timestep it's a guarantee that we will have fractional day output.

components/scream/src/share/io/scream_scorpio_interface.F90
```diff
@@ -637,17 +637,21 @@ subroutine eam_pio_createHeader(File)
...
-    retval=pio_put_att (File, PIO_GLOBAL, 'title', 'SCORPIO TEST')
-    retval=pio_put_att (File, PIO_GLOBAL, 'logname','THE GIT LOG HASH')
-    retval=pio_put_att (File, PIO_GLOBAL, 'host', 'THE HOST')
-    retval=pio_put_att (File, PIO_GLOBAL, 'Version', &
-           '0')
-    retval=pio_put_att (File, PIO_GLOBAL, 'revision_Id', &
-           'None')
-    retval=pio_put_att (File, PIO_GLOBAL, 'initial_file', 'NONE FOR NOW')
-    retval=pio_put_att (File, PIO_GLOBAL, 'topography_file', 'NONE FOR NOW')
+    retval=pio_put_att (File, PIO_GLOBAL, 'source', 'E3SM Atmosphere Model Version 4')
```
AaronDonahue, line 640: @PeterCaldwell , thank you for doing this.  It has bothered me for awhile but I've kept putting off the change.

AaronDonahue:
Just to note, when I first saw the name of this PR I thought maybe this was going to unlock the ability to designate output as daily, rather than just nsteps.  I think we might actually be there when it comes to the maturity of the scorpio interface.  And now that we are starting to do full runs it would be helpful to be able to get daily, monthly, yearly output.  Especially if we start looking for averages.

components/scream/src/share/io/scream_output_manager.cpp
PeterCaldwell: Yeah, I agree that days_from would be better than seconds_from. Would you mind making that change, Aaron? I couldn't find where seconds_from was defined by grepping or I would have made that change myself...

PeterCaldwell: Regarding Luca's second question of "should days from t0 be an integer?" - no. It already isn't because I was dividing seconds (which is an integer) by 86400.0 and passing the result to register_variable, which expects a PIO_REAL as an argument.

AaronDonahue: ahhh, but I think Luca is concerned because `seconds_from` returns an integer.  I'll put in the `days_from` and it will reuturn a double.

AaronDonahue: done

bartgol: Yeah, doing `int / int` needs care, since the result gets rounded. That's what I meant.

components/scream/src/share/util/scream_time_stamp.cpp
```diff
@@ -94,6 +94,10 @@ int TimeStamp::seconds_from (const TimeStamp& ts) const {
   return *this-ts;
 }
 
+double TimeStamp::days_from (const TimeStamp& ts) const {
+  return ((double)*this.seconds_from(ts)/86400.0);
```
bartgol, line 98: Fun fact: if one of the two operands is a double, the other is automatically promoted to double. So this works just as well:
```
  return seconds_from(ts) / 86400.0;
```
since the denominator is double.

AaronDonahue: should I change it?  Or just stick with what's there, since both work the same way?

PeterCaldwell: I don't understand the current implementation. Is this taking this.seconds_from(ts) and multiplying it by double to promote it to a double from an int, then dividing it by a double which would - as Luca notes - have an identical effect? If so, then yeah - it would probably be better to just divide by a double to get the type you want.

I'm also a bit confused why we create a new function `days_from` which just divides `seconds_from` by 86400.0... which is exactly what I had already done in the code directly. I'm not a fan of creating functions which just create extra steps and obscure what we're doing, so I'd rather keep the code as I originally implemented it unless I'm missing something.

AaronDonahue: That's alright with me.  We can just revert the branch to the commit before this one.  I put in the implementation based on this comment from above,
"
Yeah, I agree that days_from would be better than seconds_from. Would you mind making that change, Aaron? I couldn't find where seconds_from was defined by grepping or I would have made that change myself...
"

bartgol: Peter, the bottom line is code clarity. Reading `ts.seconds_from(t0) / 86400.0` requires (a tiny bit) more mental parsing than `ts.days_from(t0)`. It communicates "what" you want, hiding a bit "how" you get it.

I'm not sure if it's a personal taste though. Since it's used only in one place, and it's a quite simple and obvious calculation (we all know there are 86400 seconds in a day), it might be not that important.

PeterCaldwell: Yeah, sorry for asking you to do extra work. I had assumed that you had created `seconds_from` yourself and were going to create a `days_from` routine as an equivalent low-level function. Since using `seconds_from` then dividing by 86400 is kind of a kludge, I figured it would be better to instead create a function that returns days. But if you're just going to use the same kludge as I had, there's not point!

PeterCaldwell: Yeah, ok Luca. I don't care which method we use. If either of you have a preference, let's go with that. Maybe "what we already have" is best? In which case, I think if one of you approve the code will auto-merge?

bartgol: yeah, this is a tiny detail. I don't really care either. :)

bartgol (approved):

bartgol (approved):

bartgol (approved):

AaronDonahue:
failed on weaver for some reason.  So retesting only on Weaver

PR #1402, opened 2022-02-03
Add surface radiative fluxes to surface coupling

tcclevenger:
Recently implemented surface radiative fluxes need to be exported in SurfaceCoupling.

### DIFF
--- a/components/scream/src/control/tests/sc_tests.cpp
+++ b/components/scream/src/control/tests/sc_tests.cpp
@@ -274,6 +274,12 @@ TEST_CASE ("recreate_mct_coupling")
   FID pseudo_density_id  ("pseudo_density",  scalar3d_layout, Pa,     grid_name);
   FID qv_id              ("qv",              scalar3d_layout, nondim, grid_name);
   FID precip_liq_surf_id ("precip_liq_surf", scalar2d_layout, m/s,    grid_name);
+  FID sfc_flux_dir_nir_id("sfc_flux_dir_nir", scalar2d_layout, W/(m*m), grid_name);
+  FID sfc_flux_dir_vis_id("sfc_flux_dir_vis", scalar2d_layout, W/(m*m), grid_name);
+  FID sfc_flux_dif_nir_id("sfc_flux_dif_nir", scalar2d_layout, W/(m*m), grid_name);
+  FID sfc_flux_dif_vis_id("sfc_flux_dif_vis", scalar2d_layout, W/(m*m), grid_name);
+  FID sfc_flux_sw_net_id ("sfc_flux_sw_net",  scalar2d_layout, W/(m*m), grid_name);
+  FID sfc_flux_lw_dn_id  ("sfc_flux_lw_dn",   scalar2d_layout, W/(m*m), grid_name);
 
   // NOTE: if you add fields above, you will have to modify these counters too.
   const int num_cpl_imports    = 30;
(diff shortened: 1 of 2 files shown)


components/scream/src/control/tests/sc_tests.cpp
```diff
@@ -274,6 +274,12 @@ TEST_CASE ("recreate_mct_coupling")
   FID pseudo_density_id  ("pseudo_density",  scalar3d_layout, Pa,     grid_name);
   FID qv_id              ("qv",              scalar3d_layout, nondim, grid_name);
   FID precip_liq_surf_id ("precip_liq_surf", scalar2d_layout, m/s,    grid_name);
+  FID sfc_flux_dir_nir_id("sfc_flux_dir_nir", scalar2d_layout, W/(m*m), grid_name);
```
brhillman, line 277: So these units look great, but we named them a little differently in `atmosphere_radiation.cpp`, which @AaronDonahue pointed out is a bug (well, misleading, does not affect answers) and we need to fix it. For this PR, is there any harm in the face that here `sfc_flux_dir_vis` has units of `W/(m*m)` but in `atmosphere_radiation.cpp` we mistakenly labeled these with `Wm2`, which has the string "Wm2" for some reason?

tcclevenger: Isn't the actual units for "Wm2" in `atmosphere_radiation.cpp` actually `W/m/m`, so they are the same?

Either way, this test doesn't use the units at all, or interact with the AD, it just creates random field manager (that _just so happen_ to correspond to the AD fields), so it wouldn't matter if all the units were incorrect. But also, ideally they would be consistent.

brhillman: Yes, but the "string" is wrong in `atmosphere_radiation.cpp`. I for some reason set it to "Wm2". I could make the change in my pending PR to set these to "W/(m*m)" for consistency?

brhillman: This is currently what we have in `atmosphere_radiation`:
```
  auto Wm2 = W / m / m;
  Wm2.set_string("Wm2");
```

tcclevenger: Ahh I see. Yeah, that would be fine.

brhillman: Also just noticed there's no string representation here, so I think this comment is irrelevant.

tcclevenger:
@AaronDonahue @brhillman Because the export vars are in the FM, this is very simple. Just need to make the link between surface model and atm variable names in `scream_cpl_indices.F90`, and to tell SC not to export the values when `do_export()` is called during initialization (since these are field computed in rrtmgp). I added the `Cime` label since it actually uses this code.

AaronDonahue (approved):
It looks good to me.  Thanks for putting this together @tcclevenger

AaronDonahue:
This PR produced DIFFs in the CIME based tests.  This is expected since we should be now passing new fluxes to the surface and have correspondingly DIFFs in the fluxes coming back to the ATM.  That being said.  We've already merged one non-BFB PR today.  So we may want to manually merge this next week.  @PeterCaldwell or @bartgol , thoughts?

Note, the non-BFB PR I refer to is the one that changed time from secs to days.  I expect that CPRNC will report a diff because the time vector should be way different in any output.

bartgol:
If everyone is ok with red cdash over the weekend, we can bless on monday, test on monday morning, and if diffs are still only the expected ones, then merge (and rebless again next day)

AaronDonahue:
> If everyone is ok with red cdash over the weekend, we can bless on monday, test on monday morning, and if diffs are still only the expected ones, then merge (and rebless again next day)

I'm okay with a red CDash.  I don't think anyone even checks over the weekend.

AaronDonahue:
I'm going to manually merge, so that we can start next week with this bug fixed :-)

tcclevenger:
Thanks @AaronDonahue!

PR #1404, opened 2022-02-04
Quick fix to the default settings for scream_output.yaml

AaronDonahue:
This commit fixes an issue where long CIME runs would not produce
multiple output files when a file was full.  By not storing the
timestamp in the filename, each "new" file would have the same name
as the last file and then overwrite it.

This should not be turned on by default.

### DIFF
--- a/components/scream/data/scream_output.yaml
+++ b/components/scream/data/scream_output.yaml
@@ -59,6 +59,5 @@ Fields:
 Output Control:
   Frequency: 12
   Frequency Units: Steps
-  Timestamp in Filename: false
   MPI Ranks in Filename: true
 ...


bartgol:
Why do you want your output file to replace the previous?

brhillman:
> Why do you want your output file to replace the previous?

You *don't*! But that's the default now. This hot-fix makes the default to *not* continually overwrite your output file.

brhillman:
@bartgol maybe I misunderstood your question; do you mean why do we even have the option of *not* putting the timestamp in the filename?

brhillman (approved):
This looks great, although I wonder if we ultimately want to remove the option to not include the timestamp info in the file? In any event, we shouldn't be setting this false here, so this change is definitely needed. Thanks for fixing this!

bartgol:
> @bartgol maybe I misunderstood your question; do you mean why do we even have the option of _not_ putting the timestamp in the filename?

No, I wasn't asking that. Although except for unit testing, it's probably not that useful. I was asking why this should be the default?

brhillman:
@bartgol did I answer your question? It should not be the default, but for some reason it was. Aaron's PR here will change the default to always including the timestamp information, so output files are not overwritten.

bartgol:
Aaaah, sorry, I parsed github diff backwards. I thought there was no option in the yaml file, and the PR added it with 'false'.

Now the question is whether the default should be to include the timestamp, and to that regard I think the answer should be "yes".

PeterCaldwell:
Sanity check: by timestamp, we mean the first model timestep of the period captured in that file, right? Rather than the wallclock timestamp?

bartgol:
> Sanity check: by timestamp, we mean the first model timestep of the period captured in that file, right? Rather than the wallclock timestamp?

Yes.

PeterCaldwell:
Yeah, I don't think this should be an option. It should always be on. When would we not want it?

bartgol:
> Yeah, I don't think this should be an option. It should always be on. When would we not want it?

You are probably right. I think the rationale was to allow simpler/shorter names in our unit testing. But it seems this may be more of a source of headaches than a handy feature...

@AaronDonahue , assuming others agree, would you mind removing this option altogether from our output classes as part of this PR? Unless this is a hotfix that we desperately need merged today...

AaronDonahue:
@luca, I don't mind at all. And no, it's not a high priority hotfix. I just made the PR quickly because Ben and I were chatting about it and I didn't want to forget. 

I think removing it will require some changes to the stand alone tests. Nothing complicated, just adding the time stamp to the cprnc checks.

AaronDonahue:
@PeterCaldwell , Luca is right. We put the option in because the stand alone tests were getting bogged down with the time stamp names and it meant that the cprnc scripts had to be aware of what the time stamp would be. So if we couldn't easily change the test on the fly. But in truth I don't see us doing that often anyways. So it's probably not needed as an option.

bartgol:
Uhm. I'm actually torn. It makes unit tests much easier to handle, but otoh a) we won't use it in practice and b) we are unit testing a setting not used in real runs (though it might be an orthogonal setting to what the unit tests are testing).

PeterCaldwell:
Ah, if it's useful for unit tests I think it's fine to keep. But yeah, testing a config we don't use isn't useful. Do whatever is easiest/makes sense to you.

PR #1407, opened 2022-02-08
Enable setting fixed orbital parameters

brhillman:
Enable setting fixed orbital parameters. This is needed for aquaplanet
configurations, in which we need to set orbital parameters to perpetual
equinox conditions (by setting eccentricity and obliquity to zero). This
exposes these parameters to the AD through YAML options, and adds
default values to the scream/data/scream_input.yaml file based on
compset regex. The new behavior now checks if non-negative orbital
parameters are passed via the yaml file, and if they are, bypasses the
computation of orbital parameters based on year in favor of using the
fixed parameters. Note that ALL parameters must be non-negative, or else
all parameters will be computed instead in shr_orb_params.

[Non-B4B] (only for aquaplanet CIME test on mappy, v1 standalone is all B4B)

### DIFF
--- a/components/scream/data/scream_input.yaml
+++ b/components/scream/data/scream_input.yaml
@@ -39,10 +39,21 @@ Atmosphere Driver:
       - ch4
       - o2
       - n2
+      # Year to use for orbital calculations. If positive, use a fixed year for
+      # orbital computations, in the same way EAM F-compsets are configured. If
+      # negative, compute using current model year. This setting will be IGNORED
+      # if orbital parameters below are set to non-negative values in favor of
+      # using fixed orbital parameters (e.g., for aquaplanet compsets)
+      Orbital Year: -9999 # Negative means compute from current year
+      # Orbital parameters; if positive, use orbital parameters specified
+      # and ignore orbital year setting.
+      Orbital Eccentricity: "<${COMPSET} : .*SCREAM%AQUA.* => 0 : -9999>"  # Eccentricity for orbital computations
+      Orbital Obliquity:    "<${COMPSET} : .*SCREAM%AQUA.* => 0 : -9999>"  # Obliquity for orbital computations
+      Orbital MVELP:        "<${COMPSET} : .*SCREAM%AQUA.* => 0 : -9999>"  # Vernal Equinox Mean Longitude of Perihelion
 
   Time Stepping:
     Time Step: 300
-    Start Time: [12, 30, 00]      # Hours, Minutes, Seconds
+    Start Time: [12, 30, 00]    # Hours, Minutes, Seconds
     Start Date: [2000, 2, 1]    # Year, Month, Day
     Number of Steps: 288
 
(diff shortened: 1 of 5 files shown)


brhillman:
Some notes on this. I discovered the problem when trying to reconcile surface fluxes between v0 and v1, and the incident solar radiation was obviously different between v0 and v1 aquaplanet compsets similarly configured. E.g., the following figure shows the downwelling solar flux at the top of the model in v0 (left) and v1 (right) aquaplanet simulations:
SCREAMv1-old_vs_SCREAMv0 FDS
I discovered that we were not properly setting the orbital parameters for perpetual equinox, hence the sun being in the incorrect spot. This PR fixes this:
SCREAMv1-fix_vs_SCREAMv0 FDS
There is still a slight discrepancy between v0 and v1 incoming solar radiation, but the seasonality is fixed. I am still not sure what is causing the remaining discrepancy, but it may be related to the fact that we add a layer above model top in EAM to account for gaseous absorption between model top and top of atmosphere, which I do not believe we have implemented in SCREAM yet.

components/scream/data/scream_input.yaml
```diff
@@ -39,10 +39,21 @@ Atmosphere Driver:
       - ch4
       - o2
       - n2
+      # Year to use for orbital calculations. If positive, use a fixed year for
```
AaronDonahue, line 42: @jgfouca - I seem to remember something about our CIME setup not keeping comments from our YAML.  Is this an issue just with buildnl or would this chunk of comments in this PR also be impacted?

AaronDonahue (approved):
Looks good.  Thanks Ben for finding this issue and fixing it!  I had just one comment, which is more a question for Jim about copying comments in a YAML control file.

AaronDonahue:
I think we can manually merge this once we get a response from @jgfouca to my comment above.

components/scream/data/scream_input.yaml
jgfouca: @AaronDonahue , these comments will not appear in $case/run/scream_input.yaml. The new README file for documenting scream_input.yaml tells the user to look at this file in the repo if they want to see comments. I think that's the best we can do unless we want to return to the hacky system I had for trying to preserve these comments. It's possible I can come up with a smarter design than relying purely on line numbers which failed any time the processing of the file by buildnml changed the line count.

AaronDonahue: Oh no worries.  I'm happy for the comments to get deleted in the actual case.  I just wanted to make sure that the existence of comments wouldn't cause an issue.  This is great.

AaronDonahue:
I'm going to manually merge - the FAIL on mappy is expected as this is non-BFB for the Aquaplanet testcase.

components/scream/data/scream_input.yaml
bartgol: Imho, it's ok to keep comments only in the template input file.

AaronDonahue: Absolutely.  Agreed.

PR #1409, opened 2022-02-09
Move table init out of p3_main()

tcclevenger:
Initializing these tables caused a bottle neck in the `p3_main()` call, making it to where we did not see strong scaling over MPI ranks (since the table dimensions are column independent). Now the initialization is done outside of `p3_main()`.

### DIFF
--- a/components/scream/src/physics/p3/atmosphere_microphysics.cpp
+++ b/components/scream/src/physics/p3/atmosphere_microphysics.cpp
@@ -270,6 +270,12 @@ void P3Microphysics::initialize_impl (const RunType /* run_type */)
       prog_state.qi, prog_state.qm, prog_state.ni,prog_state.bm,qv_prev,
       diag_outputs.diag_eff_radius_qc,diag_outputs.diag_eff_radius_qi);
 
+  // Load tables
+  P3F::init_kokkos_ice_lookup_tables(lookup_tables.ice_table_vals, lookup_tables.collect_table_vals);
+  P3F::init_kokkos_tables(lookup_tables.vn_table_vals, lookup_tables.vm_table_vals,
+                          lookup_tables.revap_table_vals, lookup_tables.mu_r_table_vals,
+                          lookup_tables.dnu_table_vals);
+
   // Setup WSM for internal local variables
   const auto policy = ekat::ExeSpaceUtils<KT::ExeSpace>::get_default_team_policy(m_num_cols, nk_pack);
   workspace_mgr.setup(m_buffer.wsm_data, nk_pack, 52, policy);
--- a/components/scream/src/physics/p3/p3_functions.hpp
+++ b/components/scream/src/physics/p3/p3_functions.hpp
@@ -77,6 +77,22 @@ struct Functions
   template <typename S>
   using view_2d = typename KT::template view_2d<S>;
 
+  // lookup table values for rain shape parameter mu_r
+  using view_1d_table = typename KT::template view_1d_table<Scalar, C::MU_R_TABLE_DIM>;
+
+  // lookup table values for rain number- and mass-weighted fallspeeds and ventilation parameters
+  using view_2d_table = typename KT::template view_2d_table<Scalar, C::VTABLE_DIM0, C::VTABLE_DIM1>;
+
+  // ice lookup table values
+  using view_ice_table    = typename KT::template view<const Scalar[P3C::densize][P3C::rimsize][P3C::isize][P3C::ice_table_size]>;
+
+  // ice lookup table values for ice-rain collision/collection
+  using view_collect_table = typename KT::template view<const Scalar[P3C::densize][P3C::rimsize][P3C::isize][P3C::rcollsize][P3C::collect_table_size]>;
+
+  // droplet spectral shape parameter for mass spectra, used for Seifert and Beheng (2001)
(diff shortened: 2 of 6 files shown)


components/scream/src/physics/p3/p3_functions.hpp
```diff
@@ -244,9 +262,16 @@ struct Functions
...
+  static void init_kokkos_tables_new(
+    const uview_2d<Scalar>& vn_table_vals, const uview_2d<Scalar>& vm_table_vals, const uview_2d<Scalar>& revap_table_vals,
+    const uview_1d<Scalar>& mu_r_table_vals, const uview_1d<Scalar>& dnu);
+
   static void init_kokkos_ice_lookup_tables(
     view_ice_table& ice_table_vals, view_collect_table& collect_table_vals);
 
+  static void init_kokkos_ice_lookup_tables_new(
+    const uview_ND<Scalar,4>& ice_table_vals, const uview_ND<Scalar,5>& collect_table_vals);
+
```
tcclevenger, line 274: @bartgol This obviously is not the final form, but I am having an issue that the original functions (`init_kokkos_ice_lookup_tables()` and `init_kokkos_tables()`) cannot accept unmanaged views, but the memory buffer in `atmosphere_microphysics` cannot define these `view_*_table` which are type `view<Scalar[x][y]...[z]>`.

1. Is that correct? Am I wrong to assume buffer views need to be unmanaged?
2. I don't think we want two versions of each `init_tables()`. Should `atmosphere_microphysics` just define its own version locally? Should we change all impl to the `init_tables(unmanaged_view_type....)`?

components/scream/src/physics/p3/p3_functions_f90.cpp
```diff
@@ -3235,13 +3241,25 @@ Int p3_main_f(
...
+  view_2d_table vn_table_vals, vm_table_vals, revap_table_vals;
+  view_ice_table ice_table_vals;
+  view_collect_table collect_table_vals;
+  view_dnu_table dnu;
+  P3F::init_kokkos_ice_lookup_tables(ice_table_vals, collect_table_vals);
+  P3F::init_kokkos_tables(vn_table_vals, vm_table_vals, revap_table_vals, mu_r_table_vals, dnu);
+
+  P3F::P3LookupTables lookup_tables{mu_r_table_vals, vn_table_vals, vm_table_vals, revap_table_vals,
+                                    ice_table_vals, collect_table_vals, dnu};
+
```
tcclevenger, line 3255: This was originally in `p3_main()`, now we move it out and hand `p3_main()` another input structure `P3LookupTables`

components/scream/src/physics/p3/atmosphere_microphysics.cpp
```diff
@@ -270,6 +304,20 @@ void P3Microphysics::initialize_impl (const RunType /* run_type */)
...
+                                 m_buffer.dnu_table_vals);
+
+  lookup_tables.mu_r_table_vals    = m_buffer.mu_r_table_vals;
+  lookup_tables.vn_table_vals      = m_buffer.vn_table_vals;
+  lookup_tables.vm_table_vals      = m_buffer.vm_table_vals;
+  lookup_tables.revap_table_vals   = m_buffer.revap_table_vals;
+  lookup_tables.dnu                = m_buffer.dnu_table_vals;
+  lookup_tables.ice_table_vals     = m_buffer.ice_table_vals;
+  lookup_tables.collect_table_vals = m_buffer.collect_table_vals;
+
```
tcclevenger, line 320: Here is where we need the new version of the `init_*_tables()` functions.

components/scream/src/physics/p3/p3_functions.hpp
bartgol: I don't think you want to use the atm memory buffer. You need these tables to not be overwritten by others, right? Why not just declaring them as local vars to `P3Microphysics`? Or maybe in `P3Functions::Infrastructure`?

tcclevenger: Of course.. Ok, I'll make them local..

bartgol: Nvm, just read the comment below about `P3LookupTables`.

So now I'm confused. Why do you need unmanaged views?

components/scream/src/physics/p3/atmosphere_microphysics.cpp
```diff
@@ -270,6 +270,12 @@ void P3Microphysics::initialize_impl (const RunType /* run_type */)
       prog_state.qi, prog_state.qm, prog_state.ni,prog_state.bm,qv_prev,
       diag_outputs.diag_eff_radius_qc,diag_outputs.diag_eff_radius_qi);
 
+  // Load tables
+  P3F::init_kokkos_ice_lookup_tables(lookup_tables.ice_table_vals, lookup_tables.collect_table_vals);
+  P3F::init_kokkos_tables(lookup_tables.vn_table_vals, lookup_tables.vm_table_vals,
+                          lookup_tables.revap_table_vals, lookup_tables.mu_r_table_vals,
+                          lookup_tables.dnu_table_vals);
+
```
tcclevenger, line 278: Call `init_*_tables()` functions during `initialize_impl()`, instead of P3 main.

jgfouca (approved):

PR #1410, opened 2022-02-09
Changes needed to get a scream v1 CIME case to build on summit

jgfouca:
Case used was SMS_D_Ln2_P24x1.ne4_ne4.F2000-SCREAMv1-AQP1.summit_gnugpu.scream-30min_ts

Change list:
1) Update spdlog ekat subm in order to avoid flag quoting issues
2) Add KOKKOS_DEPENDENCE to rrtmgp builds so that the kokkos_launch_compiler
   knows to inject nvcc
3) Change config_machines.xml env for summit to match scream/scripts/machines_specs.
   Needed to fix this error:
     #error The version of CUB in your include path is not compatible with
     this release of Thrust. CUB is now included in the CUDA Toolkit, so
     you no longer need to use your own checkout of CUB. Define
     THRUST_IGNORE_CUB_VERSION_CHECK to ignore this.


components/scream/src/physics/rrtmgp/CMakeLists.txt
```diff
@@ -38,6 +38,9 @@ target_include_directories(rrtmgp PUBLIC
     ${EAM_RRTMGP_DIR}/external/cpp/rrtmgp/kernels
 )
 
+target_compile_definitions(rrtmgp PRIVATE KOKKOS_DEPENDENCE)
```
bartgol, line 41: Would you mind adding a couple of lines above this to explain what compile/link option is? I'm sure people will be wondering what it is, and why other places don't have it...

jgfouca: @bartgol , yes, when the AT finishes.

jgfouca:
@bartgol , I added a comment. The problem in the AT was related to diffs that have been blessed for PRs that occurred after this branch was forked. I've done an upstream merge to resolve this.

bartgol (approved):

PR #1411, opened 2022-02-11
Homme dynamics to use homme's readnl function

bartgol:
This PR is built on top of E3SM-Project/E3SM#4788, which allows SCREAM to access some homme functionalities. As such, merging this PR would also do an upstream merge, unless we wait for that PR to go in E3SM/master first, and then do an upstream merge.

Main changes:
- HommeDynamics now parses the namelist for Homme using Homme's `readnl` subroutine.
- The vertical coordinates values are no longer read in from ascii files (whose name was stored in the namelist file). Instead, we grab them from an netcdf file, specified in the input yaml file. For now, it is not assumed that this file is the IC/restart file. If this is deemed undesirable (i.e., we want to _enforce_ the vcoord file to be the IC file), we need to make the AD pass IC file name to the atm processes constructor. That's doable, and not too invasive, but didn't want to do it unless deemed a must.
- The IC files need to be updated on the servers, to ensure they store the hybrid vcoord data. Some already do, but at least one (the homme standalone IC file) doesn't.

UPDATE:

This PR should be ready to test. IC files do not need to be updated on the servers, since homme+physics IC file already stores the vertical coordinates, and the homme_standalone test has been modified to use the IC file for homme+physics.


bartgol:
The existing `model_restart` failure persists in this PR. I am having a hard time debugging it though, since valgrind reports lots of false positives in gfortran's namelist parsing routines.

Using a Serial backend on weaver still causes the test to fail. I wonder if this has anything to do with the compiler and/or our manually built libraries on weaver.

bartgol:
Talking with @AaronDonahue , we decided it's best to get this PR merged, despite the restart test failure on CUDA. The PR does an upstream merge, as well as makes HommeDynamics use homme's native readnl subroutine to parse the input namelist. Since the failure is present in master as well, it seems more beneficial to merge this PR and debug on a separate branch, rather than keeping the PR open waiting for a bug fix (which may take a long time to arrive).

bartgol:
@AaronDonahue I'm skipping standalone testing for the next round, since on all platform, it passed (except for weaver's restart test).

AaronDonahue (approved):
I reviewed the SCREAM changes and assume the other changes are all upstream merge changes with no conflicts.

AaronDonahue (approved):

bartgol:
@AaronDonahue I tested V1 for this branch on mappy, againt the common ancestor with master (CAM, pun intended). I verified that, hacking the branch so that it loads vcoords from ascii file, it is BFB with CAM. That suggests that the non BFB diffs are due to roundoff difference between vcoords from nc file and ascii file (they are roundoff diffs, since ncdump shows 10 digits accuracy).

I think this PR is ready to be merged now.

AaronDonahue:
@bartgol , excellent.  I'm happy to manually merge or let you do it.

PR #1412, opened 2022-02-11
Correct qc computation in shoc ad

tcclevenger:
The tracer `qc` must be treated like `tke` as it is a separate input into `shoc_main()`, updated separately from the tracer bundle.

Currently, `qc` is being updated as both `shoc_ql` and as a part of the tracer bundle inside `shoc_main()`. Here, we make those separate inputs and tell the AD to overwrite the tracer bundle with `shoc_ql`.


tcclevenger:
@AaronDonahue yes, should be b4b

PR #1413, opened 2022-02-14
Add ne1024pg2_ICOS10 grid and supporting files

wlin7:
Enable coupling of ne1024pg2 with globally uniform 7.5km icosahedral 
mesh for ocean and sea ice. Grids and supporting files are added.  The 
grid alias is ne1024pg2_ICOS10, which uses r0125 for river.

Also added the new land surfdata file that makes use of invert-distance map
for grid dataset (MODIS_0.5x0.5_to_ne1024pg2) as well as a land spun-up
file for DYAMOND1 that  uses the new surfdata and the ICOS10 mask.

[BFB]

### DIFF
--- a/cime_config/config_grids.xml
+++ b/cime_config/config_grids.xml
@@ -4333,8 +4333,8 @@
     </gridmap>
 
     <gridmap ocn_grid="ICOS10" rof_grid="r0125">
-      <map name="ROF2OCN_ICE_RMAPNAME">cpl/cpl6/map_r0125_to_ICOS10_nco.211212.nc</map>
-      <map name="ROF2OCN_LIQ_RMAPNAME">cpl/cpl6/map_r0125_to_ICOS10_nco.211212.nc</map>
+      <map name="ROF2OCN_ICE_RMAPNAME">cpl/cpl6/map_r0125_to_ICOS10_smoothed.r50e100.220302.nc</map>
+      <map name="ROF2OCN_LIQ_RMAPNAME">cpl/cpl6/map_r0125_to_ICOS10_smoothed.r50e100.220302.nc</map>
     </gridmap>
 
     <!--- river flooding variables -->
--- a/cime_config/config_grids.xml
+++ b/cime_config/config_grids.xml
@@ -1457,6 +1457,16 @@
       <mask>oRRS18to6v3</mask>
     </model_grid>
 
+    <model_grid alias="ne1024pg2_ICOS10">
+      <grid name="atm">ne1024np4.pg2</grid>
+      <grid name="lnd">ne1024np4.pg2</grid>
+      <grid name="ocnice">ICOS10</grid>
+      <grid name="rof">r0125</grid>
+      <grid name="glc">null</grid>
+      <grid name="wav">null</grid>
+      <mask>ICOS10</mask>
+    </model_grid>
+
     <model_grid alias="ne1024np4_360x720cru_oRRS15to5" compset="(DOCN|XOCN|SOCN|AQP1)">
       <grid name="atm">ne1024np4</grid>
       <grid name="lnd">360x720cru</grid>
@@ -2447,6 +2457,8 @@
       <ny>1</ny>
       <file grid="atm|lnd" mask="oRRS18to6v3">$DIN_LOC_ROOT/share/domains/domain.lnd.ne1024pg2_oRRS18to6v3.200212.nc</file>
       <file grid="ice|ocn" mask="oRRS18to6v3">$DIN_LOC_ROOT/share/domains/domain.ocn.ne1024pg2_oRRS18to6v3.200212.nc</file>
+      <file grid="atm|lnd" mask="ICOS10">$DIN_LOC_ROOT/share/domains/domain.lnd.ne1024pg2_ICOS10.211018.nc</file>
+      <file grid="ice|ocn" mask="ICOS10">$DIN_LOC_ROOT/share/domains/domain.ocn.ne1024pg2_ICOS10.211018.nc</file>
       <desc>ne1024np4.pg2 is Spectral Elem 3km grid w/ 2x2 FV physics grid per element:</desc>
     </domain>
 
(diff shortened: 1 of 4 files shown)


wlin7:
Hi @jonbob , I am not able to find you in the drop-down reviewer list, but do need you to take a look of this PR.
For ROF2OCN maps, I am going with the unsmoothed mono for now. We can switch to your smoothed maps anytime you think they are ready. Thanks.

PeterCaldwell:
Looks good! Thanks Wuyin. My only question is whether we also want to enable ne1024np4 simulations in this PR. I'm leaning towards no since our production runs will use pg2. np4 would be nice, but aren't urgent and it's unclear who would use them.

jonbob:
@wlin7 - if you're going to run a fully-active case, that rof=>ocn file is not correct. I think the ones I provided are fine, but I'll keep poking at them to be sure. I am sure, however, that simple remapping for rof=>ocn will not get reasonable results

wlin7:
Thanks @jonbob . We will eventually need the smoothed maps. No hurry now. We will not be running coupled simulations with this grid in the foreseeable future.

wlin7:
> Looks good! Thanks Wuyin. My only question is whether we also want to enable ne1024np4 simulations in this PR. I'm leaning towards no since our production runs will use pg2. np4 would be nice, but aren't urgent and it's unclear who would use them.

Peter, I also think not to consider in this PR. We can do that if it is needed one day. To support bi-grid ne1024np4, it will again take a while to get all the mapping files together in order to generate land surface data.

wlin7 (commented):
Reviewed the changes, all files involved have a copy on inputdata server.

wlin7:
@PeterCaldwell , @crterai , @brhillman , can one of you approve this PR to let merge proceed? All tests passed. Thanks.

brhillman:
Nothing here appears to be specific to SCREAM, and ne1024 exists in E3SM upstream; why not push this PR upstream? This will very possibly create merge conflicts with future upstream merges.

cime_config/config_grids.xml
```diff
@@ -3332,6 +3351,15 @@
       <map name="OCN2ATM_SMAPNAME">cpl/gridmaps/ne1024pg2/map_oRRS18to6v3_to_ne1024pg2_nco.200212.nc</map>
     </gridmap>
 
+    <gridmap atm_grid="ne1024np4.pg2" ocn_grid="ICOS10">
+      <!-- 3km atm /  7.5km ocean.  downscale atm->ocn -->
+      <map name="ATM2OCN_FMAPNAME">cpl/gridmaps/ne1024pg2/map_ne1024pg2_to_ICOS10_nco.211018.nc</map>
```
brhillman, line 3356: Why use nco as opposed to mono maps here? Were the mono maps bad?

cime_config/config_grids.xml
```diff
@@ -4279,6 +4307,11 @@
       <map name="ROF2OCN_LIQ_RMAPNAME">cpl/cpl6/map_r0125_to_oRRS15to5_nn.160720.nc</map>
     </gridmap>
 
+    <gridmap ocn_grid="ICOS10" rof_grid="r0125">
+      <map name="ROF2OCN_ICE_RMAPNAME">cpl/cpl6/map_r0125_to_ICOS10_nco.211212.nc</map>
```
brhillman, line 4311: @jonbob do you have the smoothed versions of these maps yet? I am a little uncomfortable leaving this as-is with the expectation that we will somehow remember that these maps will be inappropriate for coupled cases in the future.

jonbob:
@brhillman - I did give a smoothed map to @wlin7 but I had to use a slightly different workflow than normal because both of the grids are so large. I think Wuyin preferred the one he included, but I think the smoothed one is fine. So I'm OK either way. I'll try again with the usual tools I use and see if they can be coaxed

jonbob:
Hi @brhillman -- sorry that wasn't clear. The runoff maps are simply different from all our other maps, so I expect them to fail the typical tests. I'm happy to explain more if you're interested? And we probably need to make a specific test just for them.

cime_config/config_grids.xml
wlin7: No, @brhillman . I didn't try mono map. I went for a quicker map generation with nco, as using TempestRemap was expected to take much longer between two high resolution grids. Do you have any concern of using the nco map?

jonbob:
@brhillman - sorry, I had to restart the process to make the runoff map but it's crunching along right now. If you need the mono maps I'm happy to see if tempest has any issues with the two large grids

wlin7:
> @brhillman - I did have some better luck with the runoff tools, so the last step is running right now. It should finish in a couple of hours?, and then I think it makes sense to include these more correct rof2ocn files in this PR. I think part of Wuyin's concern is that these files won't pass the ncks check_map tests, but I don't think we should expect them to since these are not typical remapping files. I'll let you and @wlin7 know once the last step finishes.

Hi  @jonbob , I would be happy to replace the rof2ocn files with your new smoothed one. the ncks check_map just raised a flag (some very big weights implying potential consistency issue). If we can be certain it is going to be fine for the actual mapping, we should go for it.

wlin7:
> Nothing here appears to be specific to SCREAM, and ne1024 exists in E3SM upstream; why not push this PR upstream? This will very possibly create merge conflicts with future upstream merges.

This is a very fair question, Ben. When we wanted to enable this grid to run SCREAM, we didn't know (and didn't try to coordinate a schedule) for upstream merge. Nevertheless, it should be rare, if ever, it would cause conflicts in future upstream merges, as we don't expect to run EAM at this resolution and there wouldn't be such addition in EAM's config files. If we do need to mirror some changes here in EAM, we will use the settings in SCREAM as the reference to avoid conflicts.

jonbob:
Thanks @wlin7 - the new one should be finished later today, but I'm afraid it will fail the ncks check_map test for the same reasons the old one does. My guess is that virtually all runoff maps will fail that test, because they really aren't meant to be simple maps between two overlapping grids.

jonbob:
@brhillman, @wlin7 - I did manage to create a new runoff mapping file using the older set of tools (with just a little coaxing). I'll put it in the inputdata repo this morning in case you want to point at it. The file is:
````
map_r0125_to_ICOS10_smoothed.r50e100.220302.nc
````
and is located in the usual directory (~inputdata/cpl/cpl6/) for rof2ocn mapping.

I'm happy to have more of a discussion about why mapping files between runoff and ocean are unique and won't pass the usual "ncks check_map" test, if needed.

wlin7:
> map_r0125_to_ICOS10_smoothed.r50e100.220302.nc

Thanks, @jonbob . I am going to use this.

brhillman (approved):
Thanks for switching the ROF maps!

brhillman (approved):

bartgol:
@wlin7 @brhillman The fail on mappy is due to the known issue with `spa_main`. If this PR is ready, I can manually merge it.

bartgol:
Great, merging.

PR #1415, opened 2022-02-15
Add SPA to the set of default processes run for all SCREAMv1 F-Case

AaronDonahue:
This commit adds SPA to the set of processes in the default
scream_input.yaml file.  This will make sure that all F-Compset
runs also include SPA.

NOTE: This PR temporarily hardcodes the default case to be on an ne4
mesh.  SPA requires different mapping files depending on the resolution
of the simulation.  We will need to update the build infrastructure
to accomodate switching files depending on the resolution of the simulation.

### DIFF
--- a/components/scream/data/scream_input.yaml
+++ b/components/scream/data/scream_input.yaml
@@ -26,7 +26,7 @@ Atmosphere Driver:
       Enable Input Fields Checks: false
       Grid: Physics GLL
       SPA Remap File: none 
-      SPA Data File: @SCREAM_INPUT_ROOT@/scream/spa_file_unified_and_complete_ne4_scream.nc
+      SPA Data File: ${DIN_LOC_ROOT}/scream/spa_file_unified_and_complete_ne4_scream.nc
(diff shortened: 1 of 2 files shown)


PeterCaldwell (approved):
Looks great, thanks!

AaronDonahue:
Not sure this why this is failing mappy with a FAIL and not just a DIFF.  I'm testing on quartz to see if the changes broke the build in some way.

AaronDonahue:
This PR shouldn't have DIFFs in the SCREAMv0 runs.  Is it possible that because we just blessed new baselines that the SCREAMv0 test simulations are providing a false fail?  @bartgol or @jgfouca

bartgol:
I think so. @jgfouca may correct me, but I don't think our CIME testing performs an automatic master merge before testing, so if master had a non-BFB pr merged, you do need to rebase. We are thinking of adding the auto-merge before testing to our CI framework though. We just need to find the time.

jgfouca:
@bartgol , i forgot about that issue. Let me do that real quick.

AaronDonahue:
Cool, I'll go ahead and rebase since it's easy.

AaronDonahue:
Merging because DIFFs are expected

PR #1416, opened 2022-02-15
Aarondonahue/add diagnostic class

AaronDonahue:
This commit will introduce a diagnostic class to the EAMXX infrastructure.  

The diagnostic class is a sub-class of the atmosphere process class with a few limitations.  The purpose of the diagnostic class is to make it possible to define a diagnostic that can be used for property checking or as simulation output, without having to add a new field to field manager.

The unique properties of the diagnostic class are as follows:
1. Only allowed to access fields in the field manager as "required", meaning a diagnostic class object cannot change a field managed variable.
2. Has the "get_diagnostic" routine which provides a Field as output that contains all of the diagnostic data and metadata.
3. That field is also stored internally to the diagnostic class, but in contrast to a field managed variable will use shared memory.  In other words, the memory footprint of a diagnostic class should be relatively small, and importantly, the data for a diagnostic class cannot be trusted unless the diagnostic has just been run, that is because all diagnostic objects share the same field memory.

This commit does not make diagnostics callable from the output streams, that will be the subject of a subsequent PR.

### DIFF
--- a/components/scream/src/diagnostics/CMakeLists.txt
+++ b/components/scream/src/diagnostics/CMakeLists.txt
@@ -1,5 +1,5 @@
 set(DIAGNOSTIC_SRCS
- potential_temperature.cpp
+  potential_temperature.cpp
 )
 
 set(DIAGNOSTIC_HEADERS
--- a/components/scream/src/diagnostics/CMakeLists.txt
+++ b/components/scream/src/diagnostics/CMakeLists.txt
@@ -1,0 +1,14 @@
+set(DIAGNOSTIC_SRCS
+ potential_temperature.cpp
+)
+
+set(DIAGNOSTIC_HEADERS
+  potential_temperature.hpp
+) 
+
+add_library(diagnostics ${DIAGNOSTIC_SRCS})
+target_include_directories(diagnostics PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/../share)
+target_link_libraries(diagnostics scream_share)
+target_compile_options(diagnostics PUBLIC)
+
+add_subdirectory(tests)
--- a/components/scream/src/diagnostics/CMakeLists.txt
+++ b/components/scream/src/diagnostics/CMakeLists.txt
@@ -2,13 +2,8 @@ set(DIAGNOSTIC_SRCS
   potential_temperature.cpp
 )
 
-set(DIAGNOSTIC_HEADERS
-  potential_temperature.hpp
-) 
-
 add_library(diagnostics ${DIAGNOSTIC_SRCS})
 target_include_directories(diagnostics PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/../share)
-target_link_libraries(diagnostics scream_share)
-target_compile_options(diagnostics PUBLIC)
+target_link_libraries(diagnostics PUBLIC scream_share)
 
 add_subdirectory(tests)
--- a/components/scream/src/diagnostics/potential_temperature.hpp
+++ b/components/scream/src/diagnostics/potential_temperature.hpp
@@ -1,0 +1,94 @@
+#ifndef EAMXX_POTENTIAL_TEMP_DIAGNOSTIC_HPP
+#define EAMXX_POTENTIAL_TEMP_DIAGNOSTIC_HPP
+
+#include "share/atm_process/atmosphere_diagnostic.hpp"
+#include "share/util/scream_common_physics_functions.hpp"
+
+namespace scream
+{
+
+/*
+ * This diagnostic will produce the potential temperature.
+ */
+
+class PotentialTemperatureDiagnostic : public AtmosphereDiagnostic
+{
+public:
+  template <typename S>
+  using SmallPack     = ekat::Pack<S,SCREAM_SMALL_PACK_SIZE>;
+
+  using Spack         = SmallPack<Real>;
+  using Pack          = ekat::Pack<Real,Spack::n>;
+  using PF            = scream::PhysicsFunctions<DefaultDevice>;
+  using KT            = KokkosTypes<DefaultDevice>;
+  using view_2d       = typename KT::template view_2d<Spack>;
+  using view_2d_const = typename KT::template view_2d<const Spack>;
+
+  // Constructors
+  PotentialTemperatureDiagnostic (const ekat::Comm& comm, const ekat::ParameterList& params);
+
+  // Set type to diagnostic
+  AtmosphereProcessType type () const { return AtmosphereProcessType::Diagnostic; }
+
+  // The name of the diagnostic
+  std::string name () const { return "Potential Temperature"; } 
+
+  // Get the required grid for the diagnostic
+  std::set<std::string> get_required_grids () const {
+    static std::set<std::string> s;
+    s.insert(m_params.get<std::string>("Grid"));
+    return s;
+  }
+
+  // Set the grid
+  void set_grids (const std::shared_ptr<const GridsManager> grids_manager);
+
+  // Actual diagnostic calculation 
+  struct run_diagnostic_impl {
+    run_diagnostic_impl() = default;
+    // Functor for Kokkos loop
+    KOKKOS_INLINE_FUNCTION
+    void operator()(const int icol) const {
+      for (int jpack=0;jpack<m_npack;jpack++) {
+        const Spack& T_mid_ij(T_mid(icol,jpack));
+        const Spack& p_mid_ij(p_mid(icol,jpack));
+        output(icol,jpack) = PF::calculate_theta_from_T(T_mid_ij,p_mid_ij);
+      }
+    }
+    int m_ncol, m_npack;
+    view_2d_const T_mid;
+    view_2d_const p_mid;
+    view_2d       output;
+    // assign variables to this structure
+    void set_variables(const int ncol, const int npack,
+      const view_2d_const& pmid_, const view_2d_const& tmid_, const view_2d& output_)
+    {
+      m_ncol = ncol;
+      m_npack = npack;
+      // IN
+      T_mid = tmid_;
+      p_mid = pmid_;
+      // OUT
+      output = output_;
+    }
+  }; // struct run_diagnostic_impl
+
+protected:
+
+  // The three main overrides for the subcomponent
+  void initialize_impl (const RunType run_type);
+  void run_impl        (const int dt);
+  void finalize_impl   ();
+
+  // Keep track of field dimensions
+  Int m_num_cols; 
+  Int m_num_levs;
+
+  // Structure to run the diagnostic
+  run_diagnostic_impl  run_diagnostic;
+
+}; // class PotentialTemperatureDiagnostic
+
+} //namespace scream
+
+#endif // EAMXX_POTENTIAL_TEMP_DIAGNOSTIC_HPP
--- a/components/scream/src/diagnostics/potential_temperature.hpp
+++ b/components/scream/src/diagnostics/potential_temperature.hpp
@@ -48,12 +48,12 @@ public:
     run_diagnostic_impl() = default;
     // Functor for Kokkos loop
     KOKKOS_INLINE_FUNCTION
-    void operator()(const int icol) const {
-      for (int jpack=0;jpack<m_npack;jpack++) {
-        const Spack& T_mid_ij(T_mid(icol,jpack));
-        const Spack& p_mid_ij(p_mid(icol,jpack));
-        output(icol,jpack) = PF::calculate_theta_from_T(T_mid_ij,p_mid_ij);
-      }
+    void operator() (const int& idx) const {
+      const int icol  = idx / m_npack;
+      const int jpack = idx % m_npack;
+      const Spack& T_mid_ij(T_mid(icol,jpack));
+      const Spack& p_mid_ij(p_mid(icol,jpack));
+      output(icol,jpack) = PF::calculate_theta_from_T(T_mid_ij,p_mid_ij);
     }
     int m_ncol, m_npack;
     view_2d_const T_mid;
--- a/components/scream/src/diagnostics/tests/CMakeLists.txt
+++ b/components/scream/src/diagnostics/tests/CMakeLists.txt
@@ -2,8 +2,7 @@
 if (NOT ${SCREAM_BASELINES_ONLY})
   include(ScreamUtils)
 
-  CreateUnitTest(dummy_diag "dummy_diag_test.cpp" "scream_share;diagnostics")
   # Test potential temperature diagnostic
-  CreateUnitTest(potential_temperature "potential_temperature_test.cpp" "scream_share;diagnostics")
+  CreateUnitTest(potential_temperature "potential_temperature_test.cpp" "scream_share;diagnostics" LABELS "diagnostics" )
 
 endif()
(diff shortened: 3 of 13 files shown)


components/scream/src/diagnostics/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
...
+)
+
+set(DIAGNOSTIC_HEADERS
+  potential_temperature.hpp
+) 
+
+add_library(diagnostics ${DIAGNOSTIC_SRCS})
+target_include_directories(diagnostics PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/../share)
+target_link_libraries(diagnostics scream_share)
+target_compile_options(diagnostics PUBLIC)
```
bartgol, line 12: What's this line for? It seems to me it can be removed, since it adds no compile options...

components/scream/src/diagnostics/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
...
+  potential_temperature.cpp
+)
+
+set(DIAGNOSTIC_HEADERS
+  potential_temperature.hpp
+) 
+
+add_library(diagnostics ${DIAGNOSTIC_SRCS})
+target_include_directories(diagnostics PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/../share)
+target_link_libraries(diagnostics scream_share)
```
bartgol, line 11: It's best to link libraries with `PUBLIC`, so that flags, include dirs, compile options, link options, etc. are forwarded to downstream targets.

components/scream/src/diagnostics/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
+set(DIAGNOSTIC_SRCS
+  potential_temperature.cpp
+)
+
+set(DIAGNOSTIC_HEADERS
```
bartgol, line 5: This var is never used. Can be removed.

components/scream/src/diagnostics/tests/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+# NOTE: tests inside this if statement won't be built in a baselines-only build
+if (NOT ${SCREAM_BASELINES_ONLY})
```
bartgol, line 2: CMake style: booleans should not be dereferenced with `${..}`. Depending on CMake policy settings, you may get a warning, suggesting you to change this into `if (NOT SCREAM_BASELINES_ONLY)`.

components/scream/src/diagnostics/tests/potential_temperature_test.cpp
```diff
@@ -0,0 +1,277 @@
...
+TEST_CASE("potential_temp_test", "potential_temp_test]"){
+  // Run tests for both Real and Pack, and for (potentially) different pack sizes
+  using scream::Real;
+  using Device = scream::DefaultDevice;
+
+  constexpr int num_runs = 5;
+
+  auto engine = scream::setup_random_test();
+
+  std::cout << " -> Number of randomized runs: " << num_runs << "\n\n";
```
bartgol, line 251: You are mixing cout and printf. I'm ok with either one, but let's keep it consistent.

components/scream/src/share/atm_process/atmosphere_diagnostic.hpp
```diff
@@ -0,0 +1,56 @@
...
+
+  // The type of subcomponent
+  AtmosphereProcessType type () const { return AtmosphereProcessType::Diagnostic; }
+
+  // Getting the diagnostic output
+  Field& get_diagnostic (const Real dt);
+
+protected:
+
+  void set_computed_field_impl (const Field& f);
```
bartgol, line 45: Unfortunately, since these methods are virtual in the base class, they can still be overridden in the derived classes. You can prevent that, by adding the keyword `final` right before the semicolon.

HOWEVER, the base class also declares `set_computed_xyz` as virtual, even though most atm procs don't need to override those (notable exception is the atm proc group class). So even if you put `final` here, they may still override those methods.

The best solution is to override the `set_computed_xzy` methods (not the `_impl` ones), and mark them with `final`.

components/scream/src/share/atm_process/atmosphere_diagnostic.hpp
```diff
@@ -0,0 +1,56 @@
...
+  // Constructor(s)
+  explicit AtmosphereDiagnostic (const ekat::Comm& comm, const ekat::ParameterList& params);
+
+  virtual ~AtmosphereDiagnostic () = default;
+
+  // The type of subcomponent
+  AtmosphereProcessType type () const { return AtmosphereProcessType::Diagnostic; }
+
+  // Getting the diagnostic output
+  Field& get_diagnostic (const Real dt);
```
bartgol, line 41: Since you don't want people to modify your field, you might mark this as `const` and return a const ref.

components/scream/src/share/atm_process/atmosphere_diagnostic.cpp
```diff
@@ -0,0 +1,28 @@
...
+AtmosphereDiagnostic (const ekat::Comm& comm, const ekat::ParameterList& params)
+  : AtmosphereProcess(comm,params)
+{
+  // Nothing to do here
+}
+
+// Function to retrieve the diagnostic output which is stored in m_diagnostic_output
+Field& AtmosphereDiagnostic::get_diagnostic (const Real dt) {
+  run(dt);
+  return m_diagnostic_output;
```
bartgol, line 16: Safety guard:
```
return m_diagnostic_output.get_const();
```
which forces customers to only extract const data from the field (cannot be written on*).


\* Well, they can still take the const data out, cast away const, overwrite the data, then do an evil laugh, but what you gonna do?

components/scream/src/share/tests/atm_process_tests.cpp
```diff
@@ -429,4 +606,117 @@ TEST_CASE ("subcycling") {
...
+  diag_fail->set_grids(gm);
+
+  // Create fields for ap1
+  for(const auto& req : ap1->get_required_field_requests()) {
+    Field f(req.fid);
+    f.allocate_view();
+    f.deep_copy(0);
+    f.get_header().get_tracking().update_time_stamp(t0);
+    // Set the required and computed fields for the process and for the diagnostic(s)
+    ap1->set_required_field(f.get_const());
```
bartgol, line 665: This seems like an overkill. Can't you just do `f.deep_copy(2);`, and do away with the AddTwo process?

components/scream/src/diagnostics/potential_temperature.hpp
```diff
@@ -0,0 +1,94 @@
...
+  // Set the grid
+  void set_grids (const std::shared_ptr<const GridsManager> grids_manager);
+
+  // Actual diagnostic calculation 
+  struct run_diagnostic_impl {
+    run_diagnostic_impl() = default;
+    // Functor for Kokkos loop
+    KOKKOS_INLINE_FUNCTION
+    void operator()(const int icol) const {
+      for (int jpack=0;jpack<m_npack;jpack++) {
```
bartgol, line 52: This for loop wont be performant on GPU, since access is not coalesced. One possibility is to incorporate `m_npacks` in the range policy size, remove this for loop, and retrieve icol/ipack as
```
void operator() (const int& idx) const {
   const int icol    = idx / m_ncol;
   const int jpack = idx % m_ncol;
   ...
}
```

components/scream/src/diagnostics/tests/potential_temperature_test.cpp
```diff
@@ -0,0 +1,277 @@
+#include "catch2/catch.hpp"
+
+#include "share/grid/mesh_free_grids_manager.hpp"
+#include "share/atm_process/atmosphere_diagnostic.hpp"
+#include "diagnostics/potential_temperature.hpp"
+
+#include "physics/share/physics_constants.hpp"
```
bartgol, line 7: Unused? Let's remove it.

components/scream/src/diagnostics/tests/potential_temperature_test.cpp
```diff
@@ -0,0 +1,277 @@
+#include "catch2/catch.hpp"
+
+#include "share/grid/mesh_free_grids_manager.hpp"
+#include "share/atm_process/atmosphere_diagnostic.hpp"
```
bartgol, line 4: Redundant, can be removed.

components/scream/src/diagnostics/tests/potential_temperature_test.cpp
```diff
@@ -0,0 +1,277 @@
...
+
+
+  // A world comm
+  ekat::Comm comm(MPI_COMM_WORLD);
+
+  // Create a grids manager - single column for these tests
+  auto gm = create_gm(comm,1,num_levs);
+
+  // Kokkos Policy
+  TeamPolicy policy(ekat::ExeSpaceUtils<ExecSpace>::get_default_team_policy(1, 1));
```
bartgol, line 128: For such a trivial policy, you don't need to bring in ExeSpaceUtils (and include its header). You can just do
```
TeamPolicy policy(1,1);
```

components/scream/src/diagnostics/tests/potential_temperature_test.cpp
```diff
@@ -0,0 +1,277 @@
...
+
+  // Input (randomized) views
+  view_1d temperature("temperature",num_mid_packs),
+          pressure("pressure",num_mid_packs);
+  // Output views
+  view_1d theta("theta",num_mid_packs);
```
bartgol, line 134: This view seems to be unused

components/scream/src/diagnostics/tests/potential_temperature_test.cpp
```diff
@@ -0,0 +1,277 @@
...
+
+  auto dview_as_real = [&] (const view_1d& v) -> rview_1d {
+    return rview_1d(reinterpret_cast<RealType*>(v.data()),v.size()*pack_size);
+  };
+  auto hview_as_real = [&] (const typename view_1d::HostMirror& v) -> typename rview_1d::HostMirror {
```
bartgol, line 139: This lambda seems to be unused.

components/scream/src/diagnostics/tests/potential_temperature_test.cpp
```diff
@@ -0,0 +1,277 @@
...
+  const auto& diag_out = diag->get_diagnostic(100.0);
+  REQUIRE(views_are_equal(diag_out,zero_f));
+  }
+  //  - theta=T when p=p0
+  {
+  Kokkos::parallel_for("", policy, KOKKOS_LAMBDA(const MemberType& team) {
+    const int i = team.league_rank();
+    const auto& T_sub = ekat::subview(T_mid_v,i);
+    const auto& p_sub = ekat::subview(p_mid_v,i);
+    Kokkos::deep_copy(T_sub,temperature);
```
bartgol, line 213: You cannot call deep_copy from inside a pfor. This should make compiler crap out (at least on GPU).

bartgol: I think you're better off doing this for loop on the host. Something like
```
const auto& T_mid_f = input_fields["T_mid"];
auto T_mid_v = T_mid_f.get_view<Real**>();
for (int icol=0; icol<ncols;++icol) {
  auto T_col = ekat::subview(T_mid_v,icol);
  Kokkos::deep_copy(T_col,temperature);
}
```

bartgol (changes requested):
Lots of good work! I think there are a few things to fix, and some to polish.

components/scream/src/diagnostics/CMakeLists.txt
AaronDonahue: I stole this line from somewhere else.  So not sure.

components/scream/src/diagnostics/potential_temperature.hpp
AaronDonahue: okay cool.  I copy/pasted this from elsewhere.  Should I clean up those other CMakeLists.txt files?

components/scream/src/diagnostics/tests/CMakeLists.txt
AaronDonahue: okay cool.  I copy/pasted this from elsewhere.  Should I clean up those other CMakeLists.txt files?

components/scream/src/diagnostics/tests/potential_temperature_test.cpp
AaronDonahue: we use it to get P0 in the actual test.

components/scream/src/diagnostics/tests/potential_temperature_test.cpp
AaronDonahue: this is copy/pasted directly from our common_physics_unit_tests.

AaronDonahue: should I fix there too?

components/scream/src/share/atm_process/atmosphere_diagnostic.hpp
```diff
@@ -38,13 +38,12 @@ class AtmosphereDiagnostic : public AtmosphereProcess
   AtmosphereProcessType type () const { return AtmosphereProcessType::Diagnostic; }
 
   // Getting the diagnostic output
-  Field& get_diagnostic (const Real dt);
+  Field get_diagnostic (const Real dt);
 
+  void set_computed_field (const Field& f);
```
bartgol, line 43: Can you mark the methods with `final`, so that derived class cannot override them again?

components/scream/src/diagnostics/tests/potential_temperature_test.cpp
```diff
@@ -122,23 +119,19 @@ void run(std::mt19937_64& engine)
   ekat::Comm comm(MPI_COMM_WORLD);
 
   // Create a grids manager - single column for these tests
-  auto gm = create_gm(comm,1,num_levs);
+  const int ncols = 1;
+  auto gm = create_gm(comm,ncols,num_levs);
 
   // Kokkos Policy
-  TeamPolicy policy(ekat::ExeSpaceUtils<ExecSpace>::get_default_team_policy(1, 1));
+  TeamPolicy policy(ncols, 1);
```
bartgol, line 126: Wait, I said to build the policy manually cause I saw you had `(1,1)`. But now you set the league size to `ncols`. And actually, later on seems like you need also a team size larger than 1. So I would go back to the ekat util, and do
```
auto policy = ekat::ExeSpaceUtils<ExecSpace>::get_default_team_policy(ncols, npacks);
```

components/scream/src/share/tests/atm_process_tests.cpp
bartgol: I think you might have missed this comment.

bartgol (changes requested):
I think there's still a couple of things to address.

components/scream/src/diagnostics/tests/potential_temperature_test.cpp
bartgol: This is a test, so not too big of a deal. However, a team size of 1 might hide some bugs, so I would recommend a team policy (ncols,npacks).

bartgol (approved):

AaronDonahue:
There is nothing in this PR that should impact the F-Compset simulations.  I'm not sure why mappy is reporting DIFFs for both of those tests.  Could this be related to the need for new baselines in the master branch?  I merged a non-BFB earlier this morning.  If so, I think it is safe to manually merge this PR.

AaronDonahue:
Merging because we understand why the DIFFs are there for SCREAMv1.  Not due to this PR, instead they are from the non-BFB PR merged this morning.

PR #1417, opened 2022-02-16
add snowl to CPL

PeterCaldwell:
A mistake in (my) understanding led us to not pass frozen hydrometeors to surface components. This PR adds precip_ice_surf to the field manager (and adds precip_ice_surf and precip_liq_surf) as default history outputs!), then passes precip_ice_surf to the a2x_Faxa_snowl variable in CPL. 

This will be non-bfb for C++ runs using the whole model but should be BFB for all unit tests. Note in particular that sc_tests.cpp (the unit tests for the coupler) is in the process of being reworked and as it stands wouldn't benefit from including snowl anyways... so I didn't include snowl in any tests. It might also fail for tests using default history output since I've added a few more history variables.

### DIFF
--- a/components/scream/src/control/atmosphere_driver.cpp
+++ b/components/scream/src/control/atmosphere_driver.cpp
@@ -802,11 +802,6 @@ void AtmosphereDriver::run (const int dt) {
   // Make sure the end of the time step is after the current start_time
   EKAT_REQUIRE_MSG (dt>0, "Error! Input time step must be positive.\n");
 
-  // Print current timestamp information
-  if (m_atm_comm.am_i_root()) {
-    std::cout << "Atmosphere step = " << m_current_ts.get_num_steps() << "; model time = " << m_current_ts.get_date_string() << " " << m_current_ts.get_time_string() << std::endl;
-  }
-
   if (m_surface_coupling) {
     // Import fluxes from the component coupler (if any)
     m_surface_coupling->do_import();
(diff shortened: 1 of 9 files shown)


components/scream/src/control/atmosphere_driver.cpp
```diff
@@ -802,11 +802,6 @@ void AtmosphereDriver::run (const int dt) {
   // Make sure the end of the time step is after the current start_time
   EKAT_REQUIRE_MSG (dt>0, "Error! Input time step must be positive.\n");
 
-  // Print current timestamp information
-  if (m_atm_comm.am_i_root()) {
-    std::cout << "Atmosphere step = " << m_current_ts.get_num_steps() << "; model time = " << m_current_ts.get_date_string() << " " << m_current_ts.get_time_string() << std::endl;
-  }
```
tcclevenger, line 808: What is the reason for removing this? I'll be honest, I did think we had something printing out timestep info..

tcclevenger (commented):
Other than the comment, it looks good to me.

components/scream/src/control/atmosphere_driver.cpp
PeterCaldwell: Wow, I have no recollection of doing this. Not just in this PR, but ever in my life! I'll revert it.

AaronDonahue: maybe the version of master you built upon was older???  Weird only this was deleted.

tcclevenger:
@PeterCaldwell Looks like I lied and the surface coupling test actually needs to be updated. I have a fix, do you want me to push it to this branch?

PeterCaldwell:
Sure, if you have a fix go ahead and update.

PeterCaldwell:
@AaronDonahue - I think this PR is failing only where expected now, but the mappy CIME test output is a bit hard to parse so I thought I'd ask for a 2nd opinion. If it's ok, could you click merge?

AaronDonahue:
I'm going to merge this.  The two DIFFs are expected.

PR #1418, opened 2022-02-16
Update jenkins script to better handle out-of-date PRs

jgfouca:
If CIME tests are requested by the AT, do an upstream merge
before the CIME tests are run so that we know any DIFF is
caused by this PR and not just that the PR is old.

Use a new simple CLI hook for git_merge_ref so we can reuse
this nice functionality without having to duplicate it in bash.

### DIFF
--- a/components/scream/scripts/jenkins/jenkins_common_impl.sh
+++ b/components/scream/scripts/jenkins/jenkins_common_impl.sh
@@ -104,28 +104,39 @@ if [ $skip_testing -eq 0 ]; then
     echo "SCREAM Stand-Alone tests were skipped, since the Github label 'AT: Skip Stand-Alone Testing' was found.\n"
   fi
 
-  # scripts-tests is pretty expensive, so we limit this testing to mappy
-  if [[ $test_scripts == 1 && "$SCREAM_MACHINE" == "mappy" ]]; then
-    # JGF: I'm not sure there's much value in these dry-run comparisons
-    # since we aren't changing HEADs
-    ./scripts/scripts-tests -g -m $SCREAM_MACHINE
-    if [[ $? != 0 ]]; then fails=$fails+1; fi
-    ./scripts/scripts-tests -c -m $SCREAM_MACHINE
-    if [[ $? != 0 ]]; then fails=$fails+1; fi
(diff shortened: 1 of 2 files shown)


components/scream/scripts/jenkins/jenkins_common_impl.sh
```diff
@@ -104,28 +104,39 @@ if [ $skip_testing -eq 0 ]; then
...
+    fi
 
-  if [[ $test_v1 == 1 && "$SCREAM_MACHINE" == "mappy" ]]; then
-    ../../cime/scripts/create_test e3sm_scream_v1 --compiler=gnu9 -c -b master
-    if [[ $? != 0 ]]; then fails=$fails+1; fi
+    if [[ $test_v0 == 1 || $test_v1 == 1 ]]; then
+      # AT CIME runs may need an upstream merge in order to ensure that any DIFFs
+      # are caused by this PR and not simply because the PR is too far behind master
+      if [ -n "$PULLREQUESTNUM" ]; then
+        ./scripts/git-merge-ref origin/master
```
bartgol, line 127: If the auto-merge fails (e.g., due to conflicts), shouldn't we immediately return, without testing?

bartgol: Or is the `-e` at the top ensuring the script returns immediately?

bartgol: Actually, the lines below seem to _assume_ the script does not exit immediately if an error occurs, no?

bartgol (commented):
Looks good, I only have one-ish question.

components/scream/scripts/jenkins/jenkins_common_impl.sh
jgfouca: @bartgol , if upstream merges were causing conflicts, then test-all-sandia would have already failed. But you're right that maybe we should not rely on that.

bartgol: @jgfouca One can add the `AT: Skip Stand-Alone Testing` label, to skip test-all completely.

PR #1419, opened 2022-02-21
Enable aerosol optics for RRTMGP

brhillman:
Enable aerosol optics for RRTMGP. This adds the SPA optical property inputs to the required fields for the `atmosphere_radiation` process, as well as adding the necessary code to merge these optical properties into the total optical properties that RRTMGP will see.

### DIFF
--- a/components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
+++ b/components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
@@ -490,6 +490,21 @@ void RRTMGPRadiation::run_impl (const int dt) {
     sfc_alb_dif_vis, sfc_alb_dif_nir,
     sfc_alb_dir, sfc_alb_dif);
 
+  // Dummy aerosol optics
+  // Setup some dummy aerosol optical properties
+  auto aer_tau_sw = real3d("aer_tau_sw", m_ncol, m_nlay, m_nswbands);
+  auto aer_ssa_sw = real3d("aer_ssa_sw", m_ncol, m_nlay, m_nswbands);
+  auto aer_asm_sw = real3d("aer_asm_sw", m_ncol, m_nlay, m_nswbands);
+  auto aer_tau_lw = real3d("aer_tau_lw", m_ncol, m_nlay, m_nlwbands);
+  parallel_for(Bounds<3>(m_nswbands,m_nlay,m_ncol), YAKL_LAMBDA(int ibnd, int ilay, int icol) {
+      aer_tau_sw(icol,ilay,ibnd) = 0;
+      aer_ssa_sw(icol,ilay,ibnd) = 0;
+      aer_asm_sw(icol,ilay,ibnd) = 0;
+  });
+  parallel_for(Bounds<3>(m_nlwbands,m_nlay,m_ncol), YAKL_LAMBDA(int ibnd, int ilay, int icol) {
+      aer_tau_lw(icol,ilay,ibnd) = 0;
+  });
+
   // Run RRTMGP driver
   rrtmgp::rrtmgp_main(
     m_ncol, m_nlay,
@@ -497,6 +512,8 @@ void RRTMGPRadiation::run_impl (const int dt) {
     gas_concs,
     sfc_alb_dir, sfc_alb_dif, mu0,
     lwp, iwp, rel, rei,
+    aer_tau_sw, aer_ssa_sw, aer_asm_sw,
+    aer_tau_lw,
     sw_flux_up, sw_flux_dn, sw_flux_dn_dir,
     lw_flux_up, lw_flux_dn, 
     sw_bnd_flux_up, sw_bnd_flux_dn, sw_bnd_flux_dir,
--- a/components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
+++ b/components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
@@ -57,6 +57,8 @@ void RRTMGPRadiation::set_grids(const std::shared_ptr<const GridsManager> grids_
   FieldLayout scalar2d_layout     { {COL   }, {m_ncol    } };
   FieldLayout scalar3d_layout_mid { {COL,LEV}, {m_ncol,m_nlay} };
   FieldLayout scalar3d_layout_int { {COL,ILEV}, {m_ncol,m_nlay+1} };
+  FieldLayout scalar3d_swband_layout { {COL,SWBND,LEV}, {m_ncol, m_nswbands, m_nlay} };
+  FieldLayout scalar3d_lwband_layout { {COL,LWBND,LEV}, {m_ncol, m_nlwbands, m_nlay} };
 
   constexpr int ps = SCREAM_SMALL_PACK_SIZE;
   // Set required (input) fields here
@@ -82,6 +84,12 @@ void RRTMGPRadiation::set_grids(const std::shared_ptr<const GridsManager> grids_
       add_field<Required>(it,scalar3d_layout_mid,kgkg,grid->name(), ps);
     }
   }
+  // Required aerosol optical properties from SPA
+  add_field<Required>("aero_tau_sw", scalar3d_swband_layout, nondim, grid->name(), ps);
+  add_field<Required>("aero_ssa_sw", scalar3d_swband_layout, nondim, grid->name(), ps);
+  add_field<Required>("aero_g_sw"  , scalar3d_swband_layout, nondim, grid->name(), ps);
+  add_field<Required>("aero_tau_lw", scalar3d_lwband_layout, nondim, grid->name(), ps);
+
   // Set computed (output) fields
   add_field<Updated >("T_mid"     , scalar3d_layout_mid, K  , grid->name(), ps);
   add_field<Computed>("SW_flux_dn", scalar3d_layout_int, Wm2, grid->name(), ps);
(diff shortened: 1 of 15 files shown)


brhillman:
Note, this is WIP for the moment because it looks like we do not yet have the SPA optics in our input files.

PeterCaldwell:
@brhillman - you mean that spa optics aren't in our initial condition files, right? Aren't we using the exact same spa-specific files for v1 as we did for v0? And haven't we been running with spa in v0 for a long time now?

AaronDonahue:
> @brhillman - you mean that spa optics aren't in our initial condition files, right? Aren't we using the exact same spa-specific files for v1 as we did for v0? And haven't we been running with spa in v0 for a long time now?

@brhillman and I talked about this on Slack and I think the tests are running and passing, but Ben was confused as to why we didn't need SPA forcing in the IC file.  I think we've sorted it out and Ben has moved on to looking at the output to see if it makes sense.

components/scream/tests/uncoupled/spa/spa_stand_alone.cpp
```diff
@@ -7,6 +7,9 @@
 #include "share/grid/mesh_free_grids_manager.hpp"
 #include "share/atm_process/atmosphere_process.hpp"
 
+#include "physics/rrtmgp/atmosphere_radiation.hpp"
```
AaronDonahue, line 10: Why does SPA stand alone need rrtmgp and YAKL?

components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
```diff
@@ -115,8 +123,10 @@ int RRTMGPRadiation::requested_buffer_size_in_bytes() const
                                 Buffer::num_2d_nlay*m_ncol*m_nlay*sizeof(Real) +
                                 Buffer::num_2d_nlay_p1*m_ncol*(m_nlay+1)*sizeof(Real) +
                                 Buffer::num_2d_nswbands*m_ncol*m_nswbands*sizeof(Real) +
-                                Buffer::num_3d_nswbands*m_ncol*(m_nlay+1)*m_nswbands*sizeof(Real) +
-                                Buffer::num_3d_nlwbands*m_ncol*(m_nlay+1)*m_nlwbands*sizeof(Real);
+                                Buffer::num_3d_nlev_nswbands*m_ncol*(m_nlay+1)*m_nswbands*sizeof(Real) +
```
AaronDonahue, line 126: I don't see where there is an (m_nlay+1) sized array being added to the buffer.  I only see the four buffer arrays of size `m_ncol*m_nlay*m_ns/lwbands`.  Am I missing something?

components/scream/tests/uncoupled/spa/spa_stand_alone.cpp
brhillman: Removed.

brhillman:
I put some quick plots together from two simulations, one with current master, and one with this branch rebased on the same commit on master. These are two-timestep sims with instantaneous outputs. Shown below are the shortwave and longwave fluxes at the top of the model:
SW_flux_up_TOM
SW_flux_dn_TOM
LW_flux_up_TOM
LW_flux_dn_TOM
The only differences in these plots should be the effect of aerosol on the radiative fluxes. We see that the *difference* in SW flux up is mostly positive, and the difference in LW flux up is mostly negative. This is consistent with the addition of a scattering medium in the SW leading to an increase in reflected SW flux at the top of the atmosphere (more of the incoming solar radiation is being reflected back to space, so SW flux up increases) and an absorbing medium in the LW (more of the terrestrial radiation emitted by the earth is absorbed by the atmosphere and then remitted at a lower temperature, leading to less outgoing longwave radiation in the presence of aerosol). Note that SW flux down at the TOA is unchanged, as we would expect since this just gives the solar insolation before the atmosphere has any opportunity to interact with it. LW flux down at the TOA is zero because we do not consider LW radiation emitted by the sun.

I think all of these effects are consistent with what we would expect from adding the optical effects of aerosol on the radiative transfer.

components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
brhillman: The `num_3d_nlev_nswbands` fields (`m_nlay+1`) are the existing 3d fluxes that were already in the buffer, the `num_3d_nlay_nswbands` fields (`m_nlay`) are the new ones that hold the aerosol optical properties that were added in this PR. I changed the name for the existing ones from `num_3d_nswbands` to `num_3d_nlev_nswbands` to distinguish between the number with `m_nlay` and `m_nlay+1` vertical levels. Does this make sense, or am I still missing something?

AaronDonahue:
@brhillman , excellent, thanks for the plots.  Follow-up sanity question.  Have you looked at the bounds on the rebased version results to confirm they are within a reasonable range of values?  We had concerns about that before the latest SPA PR went in.

components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
AaronDonahue: ahhh wait I see.  I didn't catch the nuance between the before and after.  I thought this was a whole new line added and then looked at the new variables being added below and couldn't figure out where the memory was used.  But now I see, these were already there and this isn't a new line, just a change of name.  Thanks for clarifying.

brhillman:
> @brhillman , excellent, thanks for the plots. Follow-up sanity question. Have you looked at the bounds on the rebased version results to confirm they are within a reasonable range of values? You had concerns about that before the latest SPA PR went in.

@AaronDonahue yes, the ranges appear to be the same in the output file as they are in the input file (which was not the case before). It does look like `AER_G_SW` in the input file is slightly larger than 1 though, as we identified before (1.00757), so we should check that. Maybe just a remap artifact when that input file was created. Also @bartgol did open a branch that added checks on these, we may want to merge that at some point? But note that those checks would currently fail with `AER_G_SW` being slight larger than 1.

bartgol:
> @AaronDonahue yes, the ranges appear to be the same in the output file as they are in the input file (which was not the case before). It does look like `AER_G_SW` in the input file is slightly larger than 1 though, as we identified before (1.00757), so we should check that. Maybe just a remap artifact when that input file was created. Also @bartgol did open a branch that added checks on these, we may want to merge that at some point? But note that those checks would currently fail with `AER_G_SW` being slight larger than 1.

That's a "big" overshoot. Definitely not something that can happen just because of roundoffs, I would say. It would be good to know which part of SPA causes that (namely time or vertical interpolation). The time  one seems so simple that one may be tempted to rule it out, but I am wondering whether there's some issue related to t_beg/t_end/t_len. E.g., use the length of the wrong month, which could make the den be 30 instead of 31 (and 31/30~1.03, so it could be enough to cause that kind of overshoots).

@AaronDonahue, as part of your spa test fixes, I suggest we add some additional checks on time data correctness.

AaronDonahue:
@bartgol , the overshoot that Ben is talking about is in the raw data file, so not related to the EAMxx SPA infrastructure.

AaronDonahue:
Merging now.  The fails on `mappy` are the expected DIFFs because this is non-BFB.  Ben confirmed with manually testing himself on `mappy` that the fails are expected.

AaronDonahue (approved):

brhillman:
The plots above also demonstrate the expected diffs.

AaronDonahue:
Ben rebased and rechecked.  Merging now

bartgol:
I strongly suggest we fix our raw input files then, to ensure that inputs are within the physically meaningful bounds. That way, we can enable rrtmgp input field checks, and be covered against buggy data in the future.

Note: we _could_ make the rrtmgp inputs checks a bit more loose (e.g., check that `aer_g_sw<1.01`), but I'm worried we might forget about it, leave it like that, and not be able to spot bugs later on (e.g., bug in vert/time interpolation usage that cause overshooting).

AaronDonahue:
I'm happy with fixing them.  I guess it would be good to check the actual ne30 raw data, which is the source for everything.  Can horizontal remapping bump things out of bounds?

bartgol:
> I'm happy with fixing them. I guess it would be good to check the actual ne30 raw data, which is the source for everything. Can horizontal remapping bump things out of bounds?

I don't know the specifics of horiz remap, but given e3sm needs for property preservation, I would assume it should not happen? Besides, the horiz grid is topologically closed, with no boundary, which suggests no extrapolation happens (I would assume)

AaronDonahue:
Yeah, @bartgol , it looks like the max is already in the ne30 version of the data: see below
```
Checking variable AER_G_SW
min = 0.000000, max = 1.007571
Checking variable AER_SSA_SW
min = 0.000000, max = 1.000000
Checking variable AER_TAU_SW
min = 0.000000, max = 0.183401
min = 0.000000, max = 0.039404
```

AaronDonahue:
We don't see the same in the raw-raw data which I think is used by EAMv0
```
AER_G_SW_13
min = 0.000000, max = 0.860235
AER_G_SW_12
min = 0.000000, max = 0.845013
AER_G_SW_11
min = 0.000000, max = 0.820269
AER_G_SW_10
min = 0.000000, max = 0.817363
AER_G_SW_9
min = 0.000000, max = 0.809853
AER_G_SW_8
min = 0.000000, max = 0.808941
AER_G_SW_7
min = 0.000000, max = 0.808412
AER_G_SW_6
min = 0.000000, max = 0.809544
AER_G_SW_5
min = 0.000000, max = 0.814798
AER_G_SW_4
min = 0.000000, max = 0.825059
AER_G_SW_3
min = 0.000000, max = 0.843522
AER_G_SW_2
min = 0.000000, max = 0.889769
AER_G_SW_0
min = 0.000000, max = 0.863277
AER_G_SW_1
min = 0.000000, max = 1.000000
```

AaronDonahue:
So maybe the mapping from this data set to the ne30 data set had extapolation.  Which presumably shouldn't happen.

PeterCaldwell:
Really dumb question, but - was the original SPA data from an ne120 run? I'm confused how the raw ne30 data could not have values >1 but the v1 data does have values above 1. Why would we interpolate at all between ne30 and ne30?

AaronDonahue:
Okay, I made an issue #1470 to discuss what is going on with the SPA data itself.  In the meantime I will merge this PR.

PR #1422, opened 2022-02-23
Adapt SCREAM to the changes regarding test-launcher in EKAT

bartgol:
This PR depends on E3SM-Project/EKAT#179, so it will be updated once that PR is merged.

The PR also brings in recent changes regarding MPI, though it could do more: all the compilers for scream could be changed from the MPI wrappers to the native ones (e.g., use `g++` instead of `mpicxx`), which simplifies things when building for CUDA. Perhaps something for another PR.


bartgol:
Pause testing, since there's a bug in EKAT, which causes PR 179 in EKAT to fail.

jgfouca (approved):

bartgol:
This is so old, it would require work to bring back. Closing for now.

PR #1423, opened 2022-02-23
Changes related to the handling of MPI, following EKAT update

bartgol:
EKAT has recently changed how it deals with MPI: rather than requiring mpi compiler wrappers, it links to MPI (using libs/flags found with `find_package(MPI REQUIRED)`).

EKAT is now capable of building with "raw" compilers rathern than the MPI wrappers (e.g. gcc instead of mpicc). However, building with raw compilers poses a lot of issues with upstream deps, which are not explicitly compiling/linking against MPI (relying on mpi wrappers adding those flags).

Nevertheless, turns out that (probably unrelated to this, and more related to Kokkos changes), we can do away with the `ekat_mpicxx` wrapper, which simplifies a bit our life on CUDA machines, and some cmake logic can be pruned.


bartgol:
This branch has evolved in something different, closing and reopening a new PR.

PR #1424, opened 2022-02-23
Change SPA tests to use standard input file system

### LINKED ISSUE #1420: Spa remap file and data file are not fetched with the GetInputFile macro
All our init files are in `${SCREAM_INPUT_ROOT}/inputdata/atm/scream/`, while the spa remap/data files are in `${SCREAM_INPUT_ROOT}/inputdata/scream/`. I'm not sure if we should move them to `${SCREAM_INPUT_ROOT}/inputdata/atm/scream/data` or keep them where they are. I would vote for moving them. That way, we could keep the `GetInputFile` macro as is, with no mods whatsoever.

On testing platform, these files are present because we had manually added them, but they should put in the e3sm input data server(s), and downloaded if needed.

@jgfouca I assigned this to you, cause you wrote the infrastructure and I thought this might be a quick fix for you. If you're swamped, let me know, and I can take a stab at it.

jgfouca:
Fixes #1420

### DIFF
--- a/components/scream/data/scream_input.yaml
+++ b/components/scream/data/scream_input.yaml
@@ -25,8 +25,8 @@ Atmosphere Driver:
       Enable Output Fields Checks: false
       Enable Input Fields Checks: false
       Grid: Physics GLL
-      SPA Remap File: none 
-      SPA Data File: ${DIN_LOC_ROOT}/scream/spa_file_unified_and_complete_ne4_scream.nc
+      SPA Remap File: none
+      SPA Data File: ${DIN_LOC_ROOT}/atm/scream/init/spa_file_unified_and_complete_ne4_scream.nc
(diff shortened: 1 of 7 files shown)


bartgol (approved):
Thanks for cleaning this up Jim!

Maybe a todo item for the future: we have plenty of tests using the same input files. Right now, they are all copied to each test's bin dir. We might just use a single copy from `${SCREAM_BINARY_DIR}/data`, or maybe even the file inside `DIN_LOC_ROOT` (though that might be a different fs, potentially slower, so perhaps bin dir is better).
I think that might make our binary dirs smaller (which can be an issue on testbeds and/or if one has several builds inside their home).

jgfouca:
@bartgol , yep my thoughts exactly. I have an item on my todo list to see if we can shrink the size of our build dirs by reducing duplicate copies or maybe even eliminating copies by using symlinks.

PR #1429, opened 2022-02-24
Fast-forwarded EKAT to the latest version to get the A100 machine file.

jeff-cohere:
This PR just gives SCREAM the latest version of EKAT.


jgfouca (approved):

jeff-cohere:
That didn't go so well! Can someone with access to the full log see what's going wrong?

jgfouca:
@jeff-cohere , it failed on all 3 machines, so it will likely fail on whatever machine you do SCREAM development on too. Let me know if that is not the case.

jeff-cohere:
Thanks @jgfouca . The last time I was actively working on SCREAM I was using a local machine, but it seems like the model has come far enough along that we only seem to support the Big Iron these days. I'll try this out on Cori (and/or revive my failed attempt to gain access to Compy) and see what happens.

ndkeen (approved):
I'm only approving that I see `cmake/machine-files/kokkos/nvidia-a100.cmake` is coming in

jeff-cohere:
Seems like some work has to be done on the SCREAM side in order for the new EKAT to function properly. More soon.

bartgol:
Updating EKAT is a bit of a mess, due to some recent changes in EKAT that need some fix in SCREAM. I was working on fixing SCREAM, but realized we broke SCORPIO. So I went back to EKAT, and did a overhaul of how we set flags, which seems to fix problems in SCORPIO (in a way that we can then manage from EKAT/SCREAM). E3SM-Project/ekat#197 should do that. Once merged, `bartgol/ekat-update` should do all the fixes in SCREAM that are needed to a) handle recent MPI mods, and b) don't make SCORPIO upset.

jgfouca:
@jeff-cohere , it should be fairly easy to support SCREAM development on any machine, as long as the machine has the required dependencies are there (MPI, netcdf, etc). Let me know if that's the not the case. It becomes easier if you set up a machine cache file for the machine you're working on and even easier if you add the machine to `scripts/machines_specs.py`.

bartgol:
@jeff-cohere #1433 should take care of this as well. So I think we can close this one.

jeff-cohere:
Thanks for the info, @jgfouca . I'll do some more poking around once the SCREAM side is fixed up with EKAT.

And I agree that the other PR is a good way to go, @bartgol . I'll close this one.

PR #1430, opened 2022-02-25
SCREAM: fix cmake var name

bartgol:
The Scorpio.cmake script (which establishes whether we should add scorpio as a subdir, or wrap existing, pre-built, libraries) was using a legacy name for a CMake var, causing to always build scorpio (also in CIME builds).


ndkeen (approved):
With this change, I can verify the build continues much further.

PR #1431, opened 2022-02-25
Emergency fix for mappy env

jgfouca:
Modules we were using got removed.


jgfouca:
standalone and CIMEv0 tests seem to work. CIMEv1 has problems with `-lpiof` that I don't thing are related to this PR.

bartgol (approved):

PR #1433, opened 2022-02-25
Update EKAT submodule, and fix SCREAM build configuration issues

bartgol:
This PR updates the EKAT submodule with the version from E3SM-Project/ekat#197. There were a few issues related to flags and MPI, which caused some issues in SCREAM. The EKAT pr fixes them, and this PR implements the few changes needed to build with the new version of EKAT.

Closes #1429 .

### DIFF
--- a/components/scream/CMakeLists.txt
+++ b/components/scream/CMakeLists.txt
@@ -22,20 +22,12 @@ list(APPEND CMAKE_MODULE_PATH
      ${CMAKE_CURRENT_SOURCE_DIR}/cmake
      ${CMAKE_CURRENT_SOURCE_DIR}/cmake/tpls
      ${EKAT_CMAKE_PATH}
-     ${EKAT_CMAKE_PATH}/mpi
      ${EKAT_CMAKE_PATH}/pkg_build
 )
 if (SCREAM_CIME_BUILD)
   list(APPEND CMAKE_MODULE_PATH
        ${CMAKE_CURRENT_SOURCE_DIR}/cmake/cime)
 endif ()
-
-include(EkatMpiUtils)
-
-# We should avoid cxx bindings in mpi; they are already deprecated,
-# and can cause headaches at link time, cause they require -lmpi_cxx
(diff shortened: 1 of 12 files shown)


jgfouca (approved):

jeff-cohere (approved):

bartgol:
Ok, I spent the whole day (and good chunk of night) tracking down the reason for the failures.

The failures are all in the release build on blake, and consist of all (and only) the tests involving p3. In particular, some NaN generates in the rain evaporation function. However, adding cout statements makes the tests pass. This usually points to memory corruption, but valgrind does not report any issue.

I therefore bisected the updates in ekat, and the candidate culprit is a set of mods that should be completely fine. In particular, the problem seems to be generated by [this](https://github.com/E3SM-Project/EKAT/blob/0e7b913ecf36a79f5e21846bf7345b1a3b37d871/cmake/pkg_build/EkatBuildEkat.cmake#L83) line. The line was introduced to allow suppressing ekat internal warnings, when ekat is built inside a larger project. The directory included is already added to the target public include dirs [here](https://github.com/E3SM-Project/EKAT/blob/0e7b913ecf36a79f5e21846bf7345b1a3b37d871/src/ekat/CMakeLists.txt#L81), so I'm not even sure the line in `EkatBuildEkat.cmake` should affect the target build (system includes are scanned after normal ones, and since a copy of that folder is already added as normal include, the system one should have no effect).

@jeff-cohere @jgfouca I can ninja fix this in EKAT, and then update the PR, and see if it fixes the tests. Tbf, I'm a bit worried that this has nothing to do with the failure. Alas, when valgrind doesn't help, and git bisect points to that, I'm afraid there's not much to do. Unless there's something else that looks off to you, or you have any other suggestion of what I could try.

jeff-cohere:
@bartgol, in the absence of any other info, I think your ninja fix is a good idea. At least then we can definitively know whether or not that seemingly innocuous line is related to the issue. Sorry this has been such a pain!

jeff-cohere (approved):
Nice work!

PR #1435, opened 2022-02-26
Fix creation of cmake targets wrapping scorpio libs in CIME builds

bartgol:
This should fix the link error related to `piof` that @ndkeen was seeing when building V1.

I try to build v1 on my machine with master, and was able to reproduce the link error. With this fix, the link error disappears.

Upon fixing the link error, I also added an if, to skip SPA unit tests in a SCREAM_LIB_ONLY build (which is the default in a CIME build).


ndkeen (approved):
I can confirm that with these changes (and my local changes for PM), it will complete the build.

PR #1446, opened 2022-03-02
Introduce a unit test for spa_main

AaronDonahue:
This new test focuses on the spa_main routine and will check the
temporal and vertical interpolation of SPA data.  It has three parts:

1. It checks that if the timestamp is at the beginning of the month
   with the same pressure profile as the SPA data the interpolated
   data matches the SPA input data for the beginning of the month.
2. It checks the same as with (1) except when the timestamp is at
   the end of the month.
3. It picks a few days into the month and checks that the interpolated
   data remains within the bounds of the actual SPA data.
   Note, it forces the vertical interpolation to handle cases when the
   target pressure profile is outside of the data pressure profile by
   constructing a target pressure profile with a surface pressure that
   is 5% bigger.

This commit also changes where the hybrid coordinate system is read for
input SPA data, rather than using the local hybrid coordinate system, which
could be different than the SPA data, it grabs the hybrid levels directly
from the SPA data file.

Finally, using padding for the SPA source data, this commit takes care of an
issue where extrapolated data was far out of reasonable bounds for SPA data.
The new approach more closely mirrors what is currently being done in the F90
implementation of the SPA interpolation.

### DIFF
--- a/components/scream/src/physics/spa/spa_functions_impl.hpp
+++ b/components/scream/src/physics/spa/spa_functions_impl.hpp
@@ -92,7 +92,7 @@ template <typename S, typename D>
 void SPAFunctions<S,D>
 ::spa_main(
   const SPATimeState& time_state,
-  const SPAPressureState& pressure_state,
+  const view_2d<const Spack>& p_tgt,
   const SPAData&   data_beg,
   const SPAData&   data_end,
   const SPAOutput& data_out,
@@ -107,7 +107,6 @@ void SPAFunctions<S,D>
   auto& t_len = time_state.days_this_month;
 
   const int nlevs_src = data_beg.nlevs;
-  const auto& p_tgt = pressure_state.pmid;  //TODO: since this is the only place pressure state is used, should just pass this one view.
 
   // For now we require that the Data in and the Data out have the same number of columns.
   EKAT_REQUIRE(ncols_tgt==data_beg.ncols);
--- a/components/scream/src/physics/spa/spa_functions_impl.hpp
+++ b/components/scream/src/physics/spa/spa_functions_impl.hpp
@@ -160,7 +160,7 @@ void SPAFunctions<S,D>
     Kokkos::parallel_for(
       Kokkos::TeamThreadRange(team, nk_pack), [&] (Int k) {
         // Reconstruct vertical temperature profile using the hybrid coordinate system.
-        p_src_sub(k)    = ps_src * pressure_state.hybm(k) + P0 * pressure_state.hyam(k);
+        p_src_sub(k)    = ps_src * data_beg.hybm(k)  + P0 * data_beg.hyam(k);
         // Time interpolation for CCN3
         ccn3_src_sub(k) = linear_interp(ccn3_beg_sub(k),ccn3_end_sub(k),t_norm);
     });
@@ -438,7 +438,7 @@ void SPAFunctions<S,D>
   ekat::Comm self_comm(MPI_COMM_SELF);
 
   // We have enough info to start opening the file
-  std::vector<std::string> fnames = {"PS","CCN3","AER_G_SW","AER_SSA_SW","AER_TAU_SW","AER_TAU_LW"};
+  std::vector<std::string> fnames = {"hyam","hybm","PS","CCN3","AER_G_SW","AER_SSA_SW","AER_TAU_SW","AER_TAU_LW"};
   ekat::ParameterList spa_data_in_params;
   spa_data_in_params.set("Field Names",fnames);
   spa_data_in_params.set("Filename",spa_data_file_name);
@@ -464,18 +464,24 @@ void SPAFunctions<S,D>
   //   then we will use the horizontal interpolation structure, spa_horiz_interp, to
   //   interpolate PS_v onto the simulation grid: PS_v -> spa_data.PS
   //   and so on for the other variables.
+  view_1d<Real> hyam_v("hyam",spa_horiz_interp.source_grid_nlevs);
+  view_1d<Real> hybm_v("hybm",spa_horiz_interp.source_grid_nlevs);
   view_1d<Real> PS_v("PS",spa_horiz_interp.source_grid_ncols);
   view_2d<Real> CCN3_v("CCN3",spa_horiz_interp.source_grid_ncols,spa_horiz_interp.source_grid_nlevs);
   view_3d<Real> AER_G_SW_v("AER_G_SW",spa_horiz_interp.source_grid_ncols,nswbands,spa_horiz_interp.source_grid_nlevs);
   view_3d<Real> AER_SSA_SW_v("AER_SSA_SW",spa_horiz_interp.source_grid_ncols,nswbands,spa_horiz_interp.source_grid_nlevs);
   view_3d<Real> AER_TAU_SW_v("AER_TAU_SW",spa_horiz_interp.source_grid_ncols,nswbands,spa_horiz_interp.source_grid_nlevs);
   view_3d<Real> AER_TAU_LW_v("AER_TAU_LW",spa_horiz_interp.source_grid_ncols,nlwbands,spa_horiz_interp.source_grid_nlevs);
+  auto hyam_v_h          = Kokkos::create_mirror_view(hyam_v);
+  auto hybm_v_h          = Kokkos::create_mirror_view(hybm_v);
   auto PS_v_h            = Kokkos::create_mirror_view(PS_v);
   auto CCN3_v_h          = Kokkos::create_mirror_view(CCN3_v);
   auto AER_G_SW_v_h      = Kokkos::create_mirror_view(AER_G_SW_v);
   auto AER_SSA_SW_v_h    = Kokkos::create_mirror_view(AER_SSA_SW_v);
   auto AER_TAU_SW_v_h    = Kokkos::create_mirror_view(AER_TAU_SW_v);
   auto AER_TAU_LW_v_h    = Kokkos::create_mirror_view(AER_TAU_LW_v);
+  Kokkos::deep_copy(hyam_v_h,          hyam_v);                     
+  Kokkos::deep_copy(hybm_v_h,          hybm_v);                     
   Kokkos::deep_copy(PS_v_h,            PS_v);                     
   Kokkos::deep_copy(CCN3_v_h,          CCN3_v);                   
   Kokkos::deep_copy(AER_G_SW_v_h,      AER_G_SW_v);               
@@ -491,6 +497,7 @@ void SPAFunctions<S,D>
   grid->set_dofs(dof_gids);
 
   using namespace ShortFieldTagsNames;
+  FieldLayout scalar1d_layout { {LEV}, {spa_horiz_interp.source_grid_nlevs} };
   FieldLayout scalar2d_layout_mid { {COL}, {spa_horiz_interp.source_grid_ncols} };
   FieldLayout scalar3d_layout_mid { {COL,LEV}, {spa_horiz_interp.source_grid_ncols, spa_horiz_interp.source_grid_nlevs} };
   FieldLayout scalar3d_swband_layout { {COL,SWBND, LEV}, {spa_horiz_interp.source_grid_ncols, nswbands, spa_horiz_interp.source_grid_nlevs} }; 
@@ -498,6 +505,11 @@ void SPAFunctions<S,D>
   std::map<std::string,view_1d_host> host_views;
   std::map<std::string,FieldLayout>  layouts;
   // Define each input variable we need
+  host_views["hyam"] = view_1d_host(hyam_v_h.data(),hyam_v_h.size());
+  layouts.emplace("hyam", scalar1d_layout);
+  host_views["hybm"] = view_1d_host(hybm_v_h.data(),hybm_v_h.size());
+  layouts.emplace("hybm", scalar1d_layout);
+  //
   host_views["PS"] = view_1d_host(PS_v_h.data(),PS_v_h.size());
   layouts.emplace("PS", scalar2d_layout_mid);
   //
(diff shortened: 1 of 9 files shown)


components/scream/src/physics/spa/spa_functions_impl.hpp
```diff
@@ -566,29 +589,42 @@ ::update_spa_data_from_file(
...
     int pack = kk / Spack::n; 
     int kidx = kk % Spack::n;
-    hyam_h(pack)[kidx] = hyam_v_h(kk);
-    hybm_h(pack)[kidx] = hybm_v_h(kk);
+    hyam_h(pack)[kidx] = hyam_v_h(kk-1);
+    hybm_h(pack)[kidx] = hybm_v_h(kk-1);
   }
+  const int pack = (spa_data.nlevs-1) / Spack::n;
+  const int kidx = (spa_data.nlevs-1) % Spack::n;
+  hyam_h(pack)[kidx] = 1e5;
```
bartgol, line 625: Since p_mid(end) will _never_ be 10x larger than p_src(end), hybm(end)=10 is probably way more than enough.

That said, 1e5 works.

bartgol: Actually, hold on. It's hybm that has to be large, and hyam=0, since the formula is
```
p_ref(k) = ps*hybm(k) + p0*hyam(k);
```

bartgol (changes requested):
Looks like a good cleanup! I like using `_tgt` instead of `_atm`, since it makes clear that's the output.

I think there's a bug in setting the padding for hyam/hybm.

Also, the PR title suggests you added a unit test for spa, but I don't see any new test (that's fine, I just want to understand if I'm missing something).

components/scream/src/physics/spa/spa_functions_impl.hpp
```diff
@@ -425,6 +428,24 @@ ::get_remap_weights_from_file(
   Kokkos::deep_copy(spa_horiz_interp.target_grid_loc, target_grid_loc_h);
 }  // END get_remap_weights_from_file
 /*-----------------------------------------------------------------*/
+/* Note: In this routine the SPA source data is padded in the vertical.
+ * 
+ * The padding sets the first value in a vertical column of data to 0.0.
+ * Thus the top of the model source data is always 0.0.
+ * The final value is set to match the actual surface value.  Thus for
+ * each column:
+ *   src_data = [ 0.0, actual_spa_data, actual_spa_date[-1] ]
```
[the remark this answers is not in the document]
AaronDonahue: > 

The new test is in `src/physics/spa/tests/spa_main_test.cpp`

AaronDonahue:
> Looks like a good cleanup! I like using `_tgt` instead of `_atm`, since it makes clear that's the output.
> 
> I think there's a bug in setting the padding for hyam/hybm.
> 
> Also, the PR title suggests you added a unit test for spa, but I don't see any new test (that's fine, I just want to understand if I'm missing something).

The new test is in src/physics/spa/tests/spa_main_test.cpp

components/scream/src/physics/spa/spa_functions_impl.hpp
AaronDonahue: I went with hyam because no matter what p0=10000, since its a reference surface pressure.  PS can be anything.  presumably it should never be small.  But P0 is a guarantee

bartgol:
Wow, that file showed only minor changes while I was reviewing. I must have accidentally clicked on the menu to select the commits to diff. I'll re-review.

components/scream/src/physics/spa/tests/CMakeLists.txt
```diff
@@ -14,5 +14,13 @@ CreateUnitTest(spa_one_to_one_remap_test "spa_one_to_one_remap_test.cpp" "${NEED
   LABELS "spa"
   MPI_RANKS 1 ${SCREAM_TEST_MAX_RANKS}
 )
+CreateUnitTest(spa_main_test "spa_main_test.cpp" "${NEED_LIBS}"
```
bartgol, line 17: The operations in spa_main are all rank-local, right? Is there any feature that is tested by running with 2+ ranks? If not, I would remove the multi-rank tests.

components/scream/src/physics/spa/spa_functions_impl.hpp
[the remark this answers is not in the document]
AaronDonahue: good idea

components/scream/src/physics/spa/spa_functions_impl.hpp
```diff
@@ -566,29 +589,42 @@ ::update_spa_data_from_file(
...
       // input are in real N-D views.  So we need to back out the indices for the target
       // data.
       int pack = kk / Spack::n; 
       int kidx = kk % Spack::n;
-      ccn3_h(tgt_col,pack)[kidx] += CCN3_v_h(src_col,kk)*src_wgt;
+      // Note, we want to pad the actual source data such that
+      //   Y[0] = 0.0, note this is handled by the deep copy above
+      //   Y[nlevs-1] = Y[nlevs-2]
+      // and the data level index is kk-1 for kk = 1,nlevs-2
+      const int ksrc = kk < spa_data.nlevs-1 ? kk-1 : kk-2;
```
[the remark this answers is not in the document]
AaronDonahue: I can switch to the inner levels if that is easier to understand.

components/scream/src/physics/spa/spa_functions_impl.hpp
```diff
@@ -566,29 +589,42 @@ ::update_spa_data_from_file(
...
+      ccn3_h(tgt_col,pack)[kidx] += CCN3_v_h(src_col,ksrc)*src_wgt;
       for (int n=0; n<nswbands; n++) {
-        aer_g_sw_h(tgt_col,n,pack)[kidx]   += AER_G_SW_v_h(src_col,n,kk)*src_wgt;
-        aer_ssa_sw_h(tgt_col,n,pack)[kidx] += AER_SSA_SW_v_h(src_col,n,kk)*src_wgt;
-        aer_tau_sw_h(tgt_col,n,pack)[kidx] += AER_TAU_SW_v_h(src_col,n,kk)*src_wgt;
+        aer_g_sw_h(tgt_col,n,pack)[kidx]   += AER_G_SW_v_h(src_col,n,ksrc)*src_wgt;
```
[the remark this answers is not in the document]
AaronDonahue: It's hidden in the review, but if you go up to line 585 you'll see that we are looping over all of the weights.  Which makes sense, this is applying the horizontal interpolation.  So we would expect 1+ weights per point.

components/scream/src/physics/spa/spa_functions_impl.hpp
AaronDonahue: That being said, if switching to PS makes more sense I'm happy with that.

components/scream/src/physics/spa/tests/spa_main_test.cpp
```diff
@@ -0,0 +1,381 @@
...
+  auto aer_sw_g_bnds   = view_3d<Real>("",2,ncols,nswbands);
+  auto aer_sw_ssa_bnds = view_3d<Real>("",2,ncols,nswbands);
+  auto aer_sw_tau_bnds = view_3d<Real>("",2,ncols,nswbands);
+  auto aer_lw_tau_bnds = view_3d<Real>("",2,ncols,nlwbands);
+  auto ccn3_bnds_h       = Kokkos::create_mirror_view(ccn3_bnds      );
+  auto aer_sw_g_bnds_h   = Kokkos::create_mirror_view(aer_sw_g_bnds  );
+  auto aer_sw_ssa_bnds_h = Kokkos::create_mirror_view(aer_sw_ssa_bnds);
+  auto aer_sw_tau_bnds_h = Kokkos::create_mirror_view(aer_sw_tau_bnds);
+  auto aer_lw_tau_bnds_h = Kokkos::create_mirror_view(aer_lw_tau_bnds);
+  for (size_t dof_i=0;dof_i<dofs_gids_h.size();dof_i++) {
```
bartgol, line 269: It appears to me that you are doing both time and vertical interpolation, and checking that minmax(X_out) is within minmax of both X_beg and X_end.

I think that's a weak form test. I think you should test time and vertical interpolation separately. For time, you need a pointwise min(X_beg,X_end) < X_out < max(X_beg,X_end), which is a stronger condition than minmax over column begin bounded by minmax of beg/end. 

For vertical interpolation, instead, I would pick X_beg=X_end (for simplicity). I would also consider writing a helper function that computes min/max of a 1d view:
```
template<typename ViewT>
std::pair<Real,Real> minmax (const ViewT& v_out);
```
and loop over idof (and nbands), calling minmax on subviews of src and tgt fields, and checking their values. This might remove some code duplication, and make the intent of the test (min(in)<output<max(in) over column) more obvious to the reader.

bartgol (commented):
I have a couple more comments.

components/scream/src/physics/spa/spa_functions_impl.hpp
bartgol: Ah, yeah, you're right, it's the same.

components/scream/src/physics/spa/spa_functions_impl.hpp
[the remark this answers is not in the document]
bartgol: Yeah, I think something like

```
for (int k=0; k<nlev; ++k)
  f_padded(k+1) = f(k);
}
f_padded(0) = 0;
f_padded(nlev) = f(nlev-1);
```
is a bit more readable.

AaronDonahue:
I still have some fixes to do given the reviews.  So I'll work on those.  I also wanted to comment on the AT results.  The new test should not have interfered with the existing SPA tests, which appear to be failing.  I'll jump on blake today and try to narrow down what the issue is.

components/scream/src/physics/spa/tests/CMakeLists.txt
AaronDonahue: Good point, the reading of data should be tested on multiple ranks, but that is handled by it's own unit test, so I can delete this one running on multiple ranks.

AaronDonahue:
I'm not sure why this keeps failing mappy for the CIME SCREAMv1 build?  I just tested on quartz and it was able to build.

bartgol:
```
/home/e3sm-jenkins/jenkins-ws/workspace/SCREAM_PullRequest_Autotester_Mappy/2117/scream/externals/ekat/src/ekat/util/ekat_catch_main.cpp:3:10: fatal error: catch2/catch.hpp: No such file or directory
    3 | #include "catch2/catch.hpp"
      |          ^~~~~~~~~~~~~~~~~~
```

I'm gonna guess the submodule wasn't properly updated (though I have no idea why).

bartgol:
Yup:

```
$ pwd
/home/e3sm-jenkins/jenkins-ws/workspace/SCREAM_PullRequest_Autotester_Mappy/2117/scream
$ git status
HEAD detached from 4219add
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)
  (commit or discard the untracked or modified content in submodules)

	modified:   externals/ekat (modified content)

no changes added to commit (use "git add" and/or "git commit -a")
```
@jgfouca maybe something is off with our jenkins_common scripts?

bartgol:
Ah, yes, jenkins common is not updating submodules after merging `origin/master`. I will put up a ninja fix

bartgol:
@AaronDonahue I suspect the following happened.

Your branch is a bit old, and the last ekat update changed ekat's submodules. When switching branches back/forth in test-all-scream, the Catch2 subdir is added/removed from the index, but it stays in the tree. This can cause issues when checking out the new ekat version, which should create the Catch2 subfodler (for the sub-submodule). Perhaps rebasing on master can help.

AaronDonahue:
ahhh okay.  I can rebase

AaronDonahue:
@bartgol , I will need a re-review on both of my PR's since I rebased them.

components/scream/src/physics/spa/tests/spa_main_test.cpp
```diff
@@ -0,0 +1,432 @@
...
+    const auto& ccn3_in = ekat::subview(ccn3_beg,dof_i);
+    compute_max_min(ccn3_in,1,spa_data_beg.nlevs-1,ccn3_bnds_h(0,dof_i),ccn3_bnds_h(1,dof_i));
+    for (int n=0;n<nswbands;n++) {
+      const auto& sw_g_in = ekat::subview(aer_g_sw_beg,dof_i,n);
+      compute_max_min(sw_g_in,1,spa_data_beg.nlevs-1,aer_sw_g_bnds_h(0,dof_i,n),aer_sw_g_bnds_h(1,dof_i,n));
+      const auto& sw_ssa_in = ekat::subview(aer_ssa_sw_beg,dof_i,n);
+      compute_max_min(sw_ssa_in,1,spa_data_beg.nlevs-1,aer_sw_ssa_bnds_h(0,dof_i,n),aer_sw_ssa_bnds_h(1,dof_i,n));
+      const auto& sw_tau_in = ekat::subview(aer_tau_sw_beg,dof_i,n);
+      compute_max_min(sw_tau_in,1,spa_data_beg.nlevs-1,aer_sw_tau_bnds_h(0,dof_i,n),aer_sw_tau_bnds_h(1,dof_i,n));
```
bartgol, line 376: Shouldn't the end point be `nlevs`? Is `spa_data_beg.nlevs=actual_nlevs+2`?

bartgol (commented):
I have some comments regarding test legibility, and how to hide common lengthy (and somewhat hard to quickly parse) stuff inside some lambdas for the unit test.

I'll let you decide whether that's useful or too time consuming.

components/scream/src/physics/spa/tests/spa_main_test.cpp
```diff
@@ -352,57 +350,25 @@ TEST_CASE("spa_read_data","spa")
...
-        REQUIRE(aer_tau_lw_h(dof_i,n,kpack)[kidx] <= aer_lw_tau_bnds_h(1,dof_i,n) );
-      }
+    const auto& ccn3_in  = ekat::subview(ekat::scalarize(ccn3_beg),dof_i);
+    const auto& ccn3_out = ekat::subview(ekat::scalarize(ccn3_h),dof_i);
+    check_bounds( ccn3_in, ccn3_out );
+    for (int n=0;n<nswbands;n++) {
+      auto ssv = [&](const view_3d<Spack>& v, const int dof_i, const int n) -> col_type {
+        return ekat::subview(ekat::scalarize(v),dof_i,n);
+      };
+      check_bounds (ssv(aer_ssa_sw_h,dof_i,n),ssv(aer_ssa_sw_beg,dof_i,n));
```
bartgol, line 369: Aren't you missing a similar line for all other band quantities?

AaronDonahue: duh

AaronDonahue: done

components/scream/src/physics/spa/tests/spa_main_test.cpp
```diff
@@ -366,7 +366,15 @@ TEST_CASE("spa_read_data","spa")
       auto ssv = [&](const view_3d<Spack>& v, const int dof_i, const int n) -> col_type {
         return ekat::subview(ekat::scalarize(v),dof_i,n);
       };
+      check_bounds (ssv(aer_g_sw_h,dof_i,n),ssv(aer_g_sw_beg,dof_i,n));
       check_bounds (ssv(aer_ssa_sw_h,dof_i,n),ssv(aer_ssa_sw_beg,dof_i,n));
+      check_bounds (ssv(aer_tau_sw_h,dof_i,n),ssv(aer_tau_sw_beg,dof_i,n));
+    }
+    for (int n=0;n<nlwbands;n++) {
+      auto ssv = [&](const view_3d<Spack>& v, const int dof_i, const int n) -> col_type {
```
bartgol, line 374: Since you added dof_i and n in the lambda signature, you could move the definition out of the dof_i loop, so you create one lambda only once.

bartgol (approved):

AaronDonahue:
This PR has been approved and the DIFFs reported on `mappy` are expected since this is non-BFB in SPA.  I'm going to manually merge.

PR #1450, opened 2022-03-04
Update EKAT submodule version

AaronDonahue:
This updates EKAT so that SCREAM has access to EKAT PR 204 which
fixes an issue with building using livermore computing machines.

### DIFF
--- a/components/scream/src/physics/spa/spa_functions_impl.hpp
+++ b/components/scream/src/physics/spa/spa_functions_impl.hpp
@@ -210,9 +210,8 @@ void SPAFunctions<S,D>
   // This is done using the EKAT linear interpolation routine, see /externals/ekat/util/ekat_lin_interp.hpp
   // for more details. 
   using LIV = ekat::LinInterp<Real,Spack::n>;
-  Real minthreshold = 0.0;  // Hard-code a minimum value for aerosol concentration to zero.
 
-  LIV VertInterp(ncols_atm,pressure_state.nlevs,nlevs_atm,minthreshold);
+  LIV VertInterp(ncols_atm,pressure_state.nlevs,nlevs_atm);
   /* Parallel loop strategy:
    * 1. Loop over all simulation columns (i index)
    * 2. Where applicable, loop over all aerosol bands (n index)


jeff-cohere (approved):

AaronDonahue:
Huh, weird.  It looks like there are DIFFs for the `homme_shoc_cld_spa_p3_rrtmgp` test on weaver and no where else.  Could something in the updated EKAT being causing non-BFB behavior?  @bartgol ? @jeff-cohere ?

AaronDonahue:
Adding the CIME label because I forgot to have that test on mappy.  Not testing on blake again because that passed.

jeff-cohere:
The list of changed files includes an updated Kokkos submodule, which is probably our switch to Kokkos 3.5. I thought we had already brought this into SCREAM in our prior EKAT submodule update, but I checked in my own SCREAM master clone, and it doesn't look like we did. Updating to Kokkos 3.5 from 3.3 could definitely affect BFB stuff on Weaver. That's my guess.

AaronDonahue:
I think model restart doesn't check against baselines.  Instead it run N steps directly, and then runs a restart at N/2 steps out to N steps and checks that the final result of that run matches the original.  So this would mean our non-BFB is in running with a restart.

bartgol (approved):

AaronDonahue:
I'm setting this to RETEST, the last time it ran the only fails where the expected ones on weaver.  So if they fail again and everything else is okay I'm going to manually merge this PR

AaronDonahue:
The fails on weaver are the known issue with `model_restart` failing.  The fails on mappy for CIME cases are because of this change in EKAT: https://github.com/E3SM-Project/EKAT/commit/80ed5b6496772996fd2b0d61109706d8476758f9 which impacts SPA.  I'll push a change to this PR and then retest.  If only the weaver `model_restart` fails pop up I can manually merge.

AaronDonahue:
I'm really at a loss on why this is failing on mappy with build fails.  @bartgol or @jgfouca , since you have access to mappy can you check the build logs and see if its obvious where things are going awry?

bartgol:
Did any non-bfb V1 pr get merged recently? If so, baselines need to be blessed.

AaronDonahue:
but they are failing the BUILD.

bartgol:
Seems to be the same problem as #1446 .

AaronDonahue:
should I rebase this one too then?

AaronDonahue:
@jgfouca or @bartgol , can you take a look again at the build fails here?  I rebased the branch and retested and I'm still getting build fails on mappy for the SCREAMv1 tests.  Maybe I should just delete this PR and submit a fresh one?

bartgol:
Creating a new PR should not change anything. I have to manually run the script on mappy, and see.

AaronDonahue:
@jgfouca , for sure.  I actually thought I addressed that in this PR.  There's the one change to the SPA API.  But maybe I missed something?

jgfouca:
@AaronDonahue , the issue was actually the opposite of what I said. It looks like SPA was updated based on the latest EKAT but this PR did not bring EKAT submodule all the way up to the latest commit that had those changes.

AaronDonahue:
@jgfouca , ahhhh ,aha.  Thanks for catching that.

jeff-cohere (approved):

bartgol (approved):

PR #1451, opened 2022-03-04
Set defaults for planar HOMME elements to 0

bogensch:
Set the default values for the planar HOMME elements to 0.  Previously, they were set to -1, however this causes an abort message if SCREAM is run in RRM mode (as per the logic of lines 988 to 995 of components/homme/src/share/namelist_mod.F90) due to the presence of a mesh_file being defined.


PeterCaldwell:
What does planar RRM even mean?

bogensch:
there is no such thing as planar RRM.  But DP-SCREAM default values are currently breaking RRM mode.  This fixes that.

PeterCaldwell:
Ok phew. Thanks.

bartgol:
@bogensch @PeterCaldwell FYI, the fails on weaver are a known issue, present also in master, so you can ignore those.

PeterCaldwell:
I'm not going to review because I have no idea what the proposed change does. But it's only 2 lines so should be easy for the right person. I suggest @mt5555 or @celdredsandia as alternative reviewers, but maybe @bogensch has a better idea.

oksanaguba:
Does this change need to go into scream and not e3sm?

oksanaguba:
i see that dpcrm mode does not exist in e3sm.

mt5555 (approved):

celdredsandia (approved):

AaronDonahue:
@bogensch , the mappy fails look like a bunch of namelist fails.  Is this expected?

AaronDonahue:
I rebased and set to RETEST.  Since there was no other change and the previous version was already approved I am going to approve so that if this passes the AT it will be automatically merged.

AaronDonahue (dismissed):

bogensch:
It looks like a NLFAIL.  Is that just because there are differences in the namelist?  If so that would be expected since the default values are set differently.

AaronDonahue:
@brhillman , do you know what kinds of fails we expect when we have a namelist change?  Have you ever seen fails like this from the CIME testing infrastructure?  @jgfouca ?

```
NLFAIL ERP.ne4pg2_ne4pg2.F2010-SCREAM-HR.mappy_gnu.eam-double_memleak_tol (but otherwise OK) RUN
    Case dir: /ascldap/users/e3sm-jenkins/acme/scratch/ERP.ne4pg2_ne4pg2.F2010-SCREAM-HR.mappy_gnu.eam-double_memleak_tol.C.20220324_063345_zvwbgu
NLFAIL ERP.ne4pg2_ne4pg2.F2010-SCREAM-LR.mappy_gnu.eam-double_memleak_tol (but otherwise OK) RUN
    Case dir: /ascldap/users/e3sm-jenkins/acme/scratch/ERP.ne4pg2_ne4pg2.F2010-SCREAM-LR.mappy_gnu.eam-double_memleak_tol.C.20220324_063345_zvwbgu
```

bogensch:
I haven't been able to reproduce this fail on any machine that I have access to.  If I could get more verbose error from this fail that would be helpful.

mt5555:
doesn't NLFAIL mean the new namelist diffs from the baseline (as expected for this PR), and the baseline namelist has tto be updated.

AaronDonahue:
Yeah, I think so.  I just wanted to confirm before manually merging since we haven't run into these fails before with EAMxx PRs

AaronDonahue:
@bogensch , this PR has gotten pretty old, but I think it is basically ready to merge right?

bogensch:
yes it is

AaronDonahue:
@bogensch , is this expected to be non-BFB for the typical SCREAMv1 simulations?  I see DIFFs being reported on mappy.

bogensch:
No, it's expected to be B4B.  Zero idea what could be causing this.  If there is access to more verbose output that would help.

AaronDonahue (approved):

PR #1432, opened 2022-02-25
WIP: Put tech docs in scream/docs/

PeterCaldwell:
Tech docs were spread between scream/components/scream/docs (i.e. mixed in with the code) and scream-docs/design-docs/. This PR moves all our technical documentation into scream/docs. A PR to scream-docs repo will remove the documentation there. This PR also fiddles around with formatting and so on to get all the existing documentation to compile both in standalone mode and as a single unified tech doc. All this documentation is in latex.

@bogensch - can you check scream/docs/physics/shoc/main.pdf to make sure I didn't ruin your doc?

@AaronDonahue - scream/docs/control/main.tex includes everyone originally in SCREAM as authors on each section by default. Would you mind deleting everyone who hasn't contributed significantly to the driver? You may have to add some people too (like Conrad).

PR #1454, opened 2022-03-05
Add PSL to cpl export

PeterCaldwell:
PSL (sea level pressure) was all zero in cpl.hi files. This PR adds a method for computing PSL following the methodology described in #1432 and fixes the zero PSL problem. 

One question I have is whether we should keep calling the variable psl, or whether we should take this opportunity to switch to a more self-describing name like `p_sea_level`. Any opinions?


components/scream/src/control/surface_coupling_export.cpp
```diff
@@ -22,21 +22,25 @@ void SurfaceCoupling::do_export (const bool init_phase)
   const bool scream_ad_run =
       (m_field_mgr->has_field("qv") && m_field_mgr->has_field("T_mid") &&
        m_field_mgr->has_field("p_mid") && m_field_mgr->has_field("pseudo_density") &&
-       m_field_mgr->has_field("precip_liq_surf") && m_field_mgr->has_field("precip_ice_surf"));
+       m_field_mgr->has_field("precip_liq_surf") && m_field_mgr->has_field("precip_ice_surf") &&
+       m_field_mgr->has_field("p_int") && m_field_mgr->has_field("phis") ); //PMC added
```
PeterCaldwell, line 26: Note to self: delete comments flagging where you changed code before integrating!

components/scream/src/control/surface_coupling_export.cpp
```diff
@@ -59,16 +64,23 @@ void SurfaceCoupling::do_export (const bool init_phase)
...
+      // Compute vertical layer heights (relative to ground surface rather than from sea level).
+      // Use z_int(nlevs) = z_surf = 0.0.
       const Real z_surf = 0.0;
       PF::calculate_z_int(team, num_levs, dz_i, z_surf, z_int_i);
       team.team_barrier();
       PF::calculate_z_mid(team, num_levs, z_int_i, z_mid_i);
       team.team_barrier();
 
+      // Calculate air temperature at bottom of cell closest to the ground for PSL
+      // ?would be better to do this in a function so it can be better tested?
```
PeterCaldwell, line 76: I'm not particularly comfortable with getting surface T by extrapolating the 2 lowest midpoint points. EAMf90 computes this by assuming a lapse rate of 6.5 K/km, which is perhaps more numerically stable but is otherwise way more terrible. Assuming we want to use this formulation, should we be interpolating in Z or in pressure? Should we use column_scan or create a standalone function to do this computation rather than hardcoding it in here (where it isn't well tested)?

components/scream/src/share/util/scream_common_physics_functions_impl.hpp
```diff
@@ -384,6 +384,78 @@ void PhysicsFunctions<DeviceT>::calculate_mmr_from_vmr(const MemberType& team,
   });
 }
 
+//+++++++++++PMC
+template<typename DeviceT>
+template<typename ScalarT>
+KOKKOS_INLINE_FUNCTION
+void PhysicsFunctions<DeviceT>::lapse_T_for_psl(const ScalarT& T_ground, const ScalarT& p_ground, const ScalarT& phi_ground,
```
PeterCaldwell, line 391: This function should only be used by calculate_psl and some unit tests. I wasn't sure whether I should treat it differently (e.g. making it private or something) because of this.

components/scream/src/share/util/scream_common_physics_functions_impl.hpp
```diff
@@ -384,6 +384,78 @@ void PhysicsFunctions<DeviceT>::calculate_mmr_from_vmr(const MemberType& team,
...
+ */
+
+  using C = scream::physics::Constants<Real>;
+  constexpr Real gravit = C::gravit;
+
+  //Get preliminary surface and sea level temperature to decide on lapse rate
+  auto T_sl = T_ground + 0.0065*phi_ground/gravit; //start by assuming lapse rate is 6.5 K/km
+  T_ground_tmp= T_ground; //make copy b/c may need to modify later
+
+  if (T_ground_tmp<=290.5 && T_sl>290.5){
```
PeterCaldwell, line 407: I was surprised that these functions are called from inside a loop over columns so - since we're only dealing with surface variables - these functions will only ever be called on scalars... not ScalarT=pack, but an actual scalar. It seems like this might limit these functions' usability (e.g. for writing output). More importantly, since these functions are called on scalars, I'm assuming it's fine to have as many if/else statements as I feel like. Another thing which could be a problem is perhaps I should be declaring things Real instead of ScalarT and not templating on ScalarT?

AaronDonahue:
@PeterCaldwell , I like `p_sea_level` or even `pres_sea_level` over `psl`.  If now is the time to change it I vote for that.

components/scream/src/share/util/scream_common_physics_functions_impl.hpp
bartgol: I think it's fine to keep templated on ScalarT. The code should not compile for a pack type, so there should be no risk of sneaking in bugs with unintentional behaviors.

That said, hard coding Real is perfectly fine too.

PeterCaldwell: Changed to hard-coded Real in order to be more explicit about what these variables are.

PeterCaldwell:
@AaronDonahue and @bartgol - this PR passes now for everything except restart on Weaver. Could you review it? I decided not to change PSL to p_sea_level now because I want to move on to other things and I'm still irrationally fond of the name PSL. I suspect we will change the name as part of a SCREAMv1.1 cleanup effort.

AaronDonahue:
@tcclevenger , I added you as a reviewer just for the surface coupler interface changes.  If you want to check those.

components/scream/src/control/tests/sc_tests.cpp
```diff
@@ -253,6 +253,7 @@ TEST_CASE ("recreate_mct_coupling")
   FL scalar2d_layout{ {COL          }, {ncols          } };
   FL vector2d_layout{ {COL, CMP     }, {ncols, 2       } };
   FL scalar3d_layout{ {COL, LEV     }, {ncols,    nlevs} };
+  FL scalar3d_int_layout{ {COL, LEV     }, {ncols,    nlevs+1} };
```
AaronDonahue, line 256: I think we have a tag `ILEV` for when layouts are defined at the interface instead of midpoint.  So this should be `{COL, ILEV}`

components/scream/src/control/surface_coupling_export.cpp
```diff
@@ -22,21 +22,25 @@ void SurfaceCoupling::do_export (const bool init_phase)
...
   if (scream_ad_run) {
     const int last_entry = m_num_levs-1;
     const auto& qv              = m_field_mgr->get_field("qv").get_view<const Real**>();
     const auto& T_mid           = m_field_mgr->get_field("T_mid").get_view<const Real**>();
     const auto& p_mid           = m_field_mgr->get_field("p_mid").get_view<const Real**>();
+    const auto& p_int           = m_field_mgr->get_field("p_int").get_view<const Real**>();
     const auto& pseudo_density  = m_field_mgr->get_field("pseudo_density").get_view<const Real**>();
     const auto& precip_liq_surf = m_field_mgr->get_field("precip_liq_surf").get_view<const Real*>();
     const auto& precip_ice_surf = m_field_mgr->get_field("precip_ice_surf").get_view<const Real*>();
+    const auto& phis = m_field_mgr->get_field("phis").get_view<const Real*>();
```
AaronDonahue, line 36: Not at all important, but since I left a comment elsewhere I figured I would leave this comment too.  For code readability we could align the `=` sign for this line with all other lines above it.  Like you did with the addition of `p_int`.

AaronDonahue (commented):
Looks good! Just a couple comments which should be simple fixes.

components/scream/src/control/tests/sc_tests.cpp
tcclevenger: That's my bad. I'll fix it.

components/scream/src/control/surface_coupling_export.cpp
tcclevenger: @PeterCaldwell I'll align this as well with my test fix (@AaronDonahue comment above)

tcclevenger:
@AaronDonahue @bartgol I'll let one of y'all approve so I'm not approving my changes. SC looks good to me otherwise.

PeterCaldwell:
I just realized that we haven't run CIME tests yet for this PR, so I just added that label...

AaronDonahue (approved):

AaronDonahue:
The only FAILs in the AT for this look like the expected DIFFs in the two SCREAMv1 CIME based runs.  Given that this PR is non-BFB.  I'm going to manually merge now.

AaronDonahue (approved):

components/scream/src/share/tests/common_physics_functions_tests.cpp
```diff
@@ -64,6 +64,79 @@ auto cmvdc (const ViewT& v_d) -> typename ViewT::HostMirror {
   return v_h;
 }
 
+template<typename DeviceT>
+void run_scalar_valued_fns()
+{
+  /*
+  Most common physics functions operate simultaneously on packs of vertical indices. The functions tested here
```
bartgol, line 71: To be clear, the common physics functions come in two flavors: one templated on ScalarT, without assumptions on what that scalar is; and one that works on a full column of ScalarT (which simply runs a ||for loop calling the former). As such, one can call them with `Scalar=Pack<Real,N>` (as we often do), as well as with `Scalar=Real`.

That said, since these functions are always gonna be called with 2d inputs/outputs, it's perfectly fine to test them separately.

components/scream/src/share/tests/common_physics_functions_tests.cpp
```diff
@@ -64,6 +64,79 @@ auto cmvdc (const ViewT& v_d) -> typename ViewT::HostMirror {
...
+  // If T_ground = 0, T_ground_tmp should be 255/2 and lapse should be 0.0065 Really cold case.
+  // If T_ground = 300 K with phi_ground>0 m2/s2, lapse = 0 and T_ground_tmp=0.5*(290.5+T_ground). Really hot case.
+  // If T_ground = 290 K and phi_ground=10000 m2/s2, T_ground_tmp=T_ground and lapse
+  //    is such that T_sl=T_ground+lapse*phi_ground/gravit is within roundoff of 290.5 K. Marginally hot case.
+  // If T_ground = 280 K and phi_ground=100 m2/s2, T_ground_tmp=T_ground and lapse=6.5 K/km (typical conditions)
+  RealType lapse;
+  RealType T_ground_tmp;
+  RealType T_ground=0;
+  RealType phi_ground=100;
+  RealType p_ground=100000; //can't use p0 because that's a pack
```
bartgol, line 96: p0 (from physics/share) is not a pack. It can be, if `Scalar=Pack<Real,N>`, but it can be a normal real with `Scalar=Real`. But since we have p0 in `src/physics/share`, we should not use it, or else we create dependency on a downstream pkg.

components/scream/src/share/tests/common_physics_functions_tests.cpp
```diff
@@ -64,6 +64,79 @@ auto cmvdc (const ViewT& v_d) -> typename ViewT::HostMirror {
...
+  PF::lapse_T_for_psl(T_ground, p_ground,  phi_ground, lapse, T_ground_tmp );
+  REQUIRE( Check::approx_equal(T_ground_tmp,0.5*255.,test_tol) );
+  REQUIRE( Check::equal(lapse,0.0065) );
+  T_ground=300;
+  PF::lapse_T_for_psl(T_ground, p_ground,  phi_ground, lapse, T_ground_tmp );
+  REQUIRE( Check::approx_equal(T_ground_tmp,0.5*(290.5+T_ground),test_tol) );
+  REQUIRE( Check::equal(lapse,0) );
+  T_ground=290;
```
bartgol, line 104: Do you think it would add value to randomize the inputs? Perhaps over an interval that goes a tad beyond normal values (for robustness)?

components/scream/src/share/util/scream_common_physics_functions_impl.hpp
```diff
@@ -384,6 +384,74 @@ void PhysicsFunctions<DeviceT>::calculate_mmr_from_vmr(const MemberType& team,
...
+    Choose lapse rate and effective ground temperature to use for sea-level pressure calculation.
+    This function should only be used by calculate_psl and is separated from that function solely 
+    to improve specificity of unit testing.
+ */
+
+  using C = scream::physics::Constants<Real>;
+  constexpr Real gravit = C::gravit;
+
+  //Get preliminary surface and sea level temperature to decide on lapse rate
+  auto T_sl = T_ground + 0.0065*phi_ground/gravit; //start by assuming lapse rate is 6.5 K/km
```
bartgol, line 402: Beware that, sInce 0.0065 is a double (that's the default floating point type), you are performing calculations in double precision here. `T_sl` is a double, for instance, regardless of the precision of the other inputs.

Of course, when stuff is stored in `lapse` or `T_ground_tmp` it is converted back to `Real`, so the final precision is the normal scream one. But I don't know if you want the intermediate calculations do be _always_ in double precision.

bartgol: In general, if you want your hard coded numbers to match the default precision of scream (i.e., Real), you should use the `sp` function. E.g.,

```
auto t_sl = T_ground + sp(0.0065) * phi_ground/gravit;
```
Or even just `Real(0.0065)`, which is what the `sp` function does internally.

Note: for integer numbers, you can also avoid using the `.`. So long as they are used in arithmetic expressions involving Real's, the integer will be promoted to a Real. If, otoh, you write

```
auto blah = 2.0 * some_real;
```
Then `some_real` is promoted to double, and the type of `blah` is `double`.

bartgol:
I removed automerge, since I think @tcclevenger (who worked a lot on the SC class) should have a look at this before it's merged.

tcclevenger (approved):

components/scream/src/share/tests/common_physics_functions_tests.cpp
PeterCaldwell: The new functions added here are different than other common physics functions because they *only* operate on reals. It would be nonsensical to test them on packs because all common physics functions are called within a loop over the horizontal and these functions are also only defined for a single vertical layer. I originally had these functions templated on ScalarT but then I got worried that someone might test them on Packs, get a fail (because what does a pack even mean in this context), and spend a lot of time debugging.

components/scream/src/share/tests/common_physics_functions_tests.cpp
PeterCaldwell: Again, problem is asserting p0 is a pack instead of "can be" a pack.

components/scream/src/share/tests/common_physics_functions_tests.cpp
PeterCaldwell: I definitely don't want to randomize the input for the property tests I've added here - I'm testing specific logic branches here. I wondered about *also* adding some randomized tests, but decided that they wouldn't add much value.

components/scream/src/share/tests/common_physics_functions_tests.cpp
PeterCaldwell: Not relevant to this PR, but note that p0 *is* already used in this test file so the dependency on a downstream package is already present. I'm not clear why that would be a problem.

components/scream/src/share/util/scream_common_physics_functions_impl.hpp
PeterCaldwell: Good catch. I meant to sp that number. I'll do that.

PeterCaldwell:
Need another re-approval b/c I did some cleanup.

tcclevenger (approved):

AaronDonahue:
Just checking @PeterCaldwell that you have reviewed all of @bartgol 's comments are comfortable merging this now.  Is that true?

PeterCaldwell:
I got a bit overexcited about making changes and am seeing build fails on Weaver and Mappy now but I can't debug them (b/c I don't have machine access and because I'm in meetings) so I don't know whether they are real or not. Check out the latest AT tests...

bartgol:
@PeterCaldwell The fail on weaver seems to be a build fail. I can take a look.

PeterCaldwell:
Ok, I think PSL is back to behaving as expected. Can you check over and integrate, @AaronDonahue ?

AaronDonahue:
Looks good.  Now it's passing the weaver build and tests and the only fails I see are the expected mappy DIFFs since this is non-BFB.

AaronDonahue:
Note, we have already merged Ben's non-BFB rrtmgp PR.  Are we comfortable merging both of these non-BFB PR's the same day?  Just doing my diligence before manually merging. @PeterCaldwell? @bartgol ?

PeterCaldwell:
I don't think there's a rush - we can do it first thing tomorrow AM.

tcclevenger:
@bartgol Did you have a chance to look at this? If not and you don't have time, I can real quick. It may involve SC stuff.

AaronDonahue:
@tcclevenger and @bartgol , I was planning on manually merging this today now that it looks like it is passing the AT (except the non-BFB DIFFs that we expect).  But if you want one more look at it, I can hold off.

tcclevenger:
Oh I thought it had a build fail on weaver. It's good to merge for me.

PeterCaldwell:
@tcclevenger - I'm unclear what problem you suggest Luca look at - the only failures we currently see are (according to @AaronDonahue ) expected cprnc diffs on mappy due to the addition of an additional output variable. I think this is ready to merge (so long as someone takes one last glance at the code). 

I was thinking that Aaron was waiting to do this until new baselines were created for the current master, which @jgfouca is doing... but maybe I'm wrong.

AaronDonahue:
okay cool.  I'm going to manually merge now.

PR #1456, opened 2022-03-07
Add CIME support for weaver

jgfouca:
This was harder than I thought it would be and I was not able to get `e3sm_scream_v1` to PASS, I was only able to get builds to work. Tests fail due to running out of memory on the device.


tcclevenger:
@bartgol No, I have checked memory consumption.

jgfouca:
@bartgol , both `e3sm_scream_v1` test cases have debug mode enabled.

PR #1459, opened 2022-03-08
Change weaver to use netcdf mode for PIO

jgfouca:
Fixes SMS_D_Ln2_P24x1.ne4_ne4.F2000-SCREAMv1-AQP1, and it actually passes now!

SMS_D_Ln2_P24x1.ne4_ne4.F2000SCREAMv1.weaver_gnugpu.scream-30min_ts still fails with an error coming from elm (land):

```
 check_downscale_consistency ERROR: column-level forcing differs from gridcell-level forcing for urban point
 c, g =         4183         280
 ENDRUN:ERROR in /home/jgfouca/scream/components/elm/src/main/atm2lndMod.F90 at line 440
```

At this point, I may have to ask for outside assistance since I don't know the first thing about elm.

### DIFF
--- a/cime_config/machines/config_pio.xml
+++ b/cime_config/machines/config_pio.xml
@@ -52,6 +52,7 @@
       <value mach="userdefined">netcdf</value>
       <value mach="melvin">netcdf</value>
       <value mach="mappy">netcdf</value>
+      <value mach="weaver">netcdf</value>
       <value mach="eastwind">netcdf</value>
       <value mach="constance">netcdf</value>
       <value mach="cascade">netcdf</value>


bartgol:
Jim, unrelated (but also related) question: on weaver, we ended up using our own gcc/8.5 compiler built from scratch, with all the packages built with that (including mpi). We needed that b/c existing modules had either internal compiler errors (gcc<=7) or were not supported by nvcc (gcc>=9). How does cime manage to build?

jgfouca:
@bartgol , if you look at the original big commit I did to support weaver, you'll see I added all that stuff to config_machines.xml for weaver.

bartgol:
> @bartgol , if you look at the original big commit I did to support weaver, you'll see I added all that stuff to config_machines.xml for weaver.

Aaah, sorry, I miserad the pr title. I read `to use netcdf *module*`, and I was wondering how that was gonna be compatible with our build, so I wondered what was cime doing. Ok then.

bartgol (approved):

jgfouca:
Ah, 😆

PeterCaldwell:
@wlin7 - Do you understand the error Jim reports above for realistic-terrain runs on GPU?
```
 check_downscale_consistency ERROR: column-level forcing differs from gridcell-level forcing for urban point
 c, g =         4183         280
 ENDRUN:ERROR in /home/jgfouca/scream/components/elm/src/main/atm2lndMod.F90 at line 440
```

PR #1461, opened 2022-03-09
Fix model_restart test on GPU

### LINKED ISSUE #1453: model_restart test (seemingly randomly) failing on weaver after 1433 got merged
What I have so far:

- The test seems to have somewhat random failures, but I observed (though after running 1k times memories are mixing) that the test passes the first time it's run right after cmake ran, while subsequent calls to ctest exibit failures.
- Saving the output file `model_output_baseline.***.nc` after the 1st run after cmake, and after a subsequent run, shows sizeable diffs in almost all fields. This suggests that some input file is modified (though I didn't see any diffs pre/post 1st run), or that the presence of some files trigger different behavior in the executable. Recall, the test should be DETERMINISTIC, since it uses fixed (i.e., no random) inputs.
- The test fails also without ctest parallel execution.
- Reverting master to before 1433 was merged makes the test consistently pass
- While I don't know if this is part of the problem, the kokkos CXX flags (as reported by kokkoscore.dir/flags.make) are different pre/post 1433 merge. In particular, before that PR, our debug bfb flags are not present, while they are present afterword. That is
```
$ cat /path/to/pre-merge/build/kokkoscore.dir/flags | grep CXX_FLAGS
CXX_FLAGS = -g -fopenmp -march=znver2 -mtune=znver2 -w -std=gnu++14
$ cat /path/to/post-merge/build/kokkoscore.dir/flags | grep CXX_FLAGS
CXX_FLAGS =  -ffp-contract=off -Wall -g -g -O0 -fopenmp-simd -fopenmp -march=znver2 -mtune=znver2 -w -std=gnu++14
```

bartgol:
The test was failing since, after the last ekat update (which moved CUDA flags out of CMAKE_CXX_FLAGS, in favor of a per-target basis) we were no longer setting `--fmad=false` on rrtmgp.

This PR also updates EKAT.

Fixes #1453

### DIFF
--- a/components/scream/src/physics/rrtmgp/CMakeLists.txt
+++ b/components/scream/src/physics/rrtmgp/CMakeLists.txt
@@ -1,8 +1,11 @@
+include(EkatUtils)
+include(EkatSetCompilerFlags
+)
 # Add CUDA flags for YAKL
 set(YAKL_CXX_FLAGS "")
 if (CUDA_BUILD)
-    set(YAKL_ARCH "CUDA")
-    list(APPEND YAKL_CXX_FLAGS -DYAKL_ARCH_CUDA --expt-extended-lambda --expt-relaxed-constexpr)
+  set(YAKL_ARCH "CUDA")
+  list(APPEND YAKL_CXX_FLAGS -DYAKL_ARCH_CUDA --expt-extended-lambda --expt-relaxed-constexpr)
 endif()
 
 # RRTMGP++ requires YAKL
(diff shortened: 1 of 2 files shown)


bartgol:
@jgfouca I updated EKAT, and fixed a bug, which was revealed when I built with cuda RDC enabled.

bartgol:
The fail on Mappy is due to #1446. Since this PR is orthogonal, I'm going to manually merge.

Issue #1449, opened 2022-03-04
Process Name vs. Process Type


AaronDonahue:
This thought occurred to me during our conversation about runtime options, yaml parsing, xml parsing, etc.

Currently when subcycling we use a process group if we want to subcycle 2+ processes together.  One of the entries is "Process Name", which has to be "Group".  If we have multiple groups of tightly subcycled processes they will all have the name "Group".  Which makes it difficult for something like `yamlchange` to find the proper named process and change the sub-category.

See #1371 for an example.

Can we have something like `Process Name` which can be any unique string and `Process Type` which would be "Group", "SHOC", etc.

bartgol:
Yeah, that's doable.

Issue #843, opened 2020-12-21
Improve format of Atm Proc sequence in yaml input file


bartgol:
Currently, the sequence is specified as:

```
  Atmosphere Processes:
    Number of Entries: 3
    Schedule Type: Sequential
    Process 0:
      Process Name: Homme
      [homme options]
    Process 1:
      Process Name: P3
      [p3 options]
    Process 2:
      Process Name: SHOC
      [shoc options]
```
It would be nice to specify them in a more compact way, like this
```
  Atmosphere Processes:
    Processes: seq(Homme, P3, SHOC)
    Homme:
       [homme options]
    P3: 
       [p3 options]
    SHOC:
       [shoc options]
```
This will help reduce the complexity of input files. E.g., it will reduce the number of lines, and the indentation levels.

Note: in case of parallel splitting, one would use the following syntax:
```
   Processes: par(50 Homme, 50 seq(P3, SHOC))
   Homme:
     ...
   P3:
     ...
   SHOC:
     ...
```
The `par` and `seq` keywords are to express parallel/sequential splitting, and the leading integer is the percentage of atm ranks to dedicate to the current proc in the parallel splitting (with a runtime check that enforces zero remainder in N_atm*50/100).

bartgol:
@jeff-cohere This is the issue I was thinking about for the yaml file atm proc specs change. The `par/seq` thingy could be deferred till we actually have a parallel splitting. I think the named processes sublists would already increase readability.

AaronDonahue:
@luca, i feel like our current format fixes this right?

bartgol:
Right.

PR #1462, opened 2022-03-09
Prototype of new namelists.xml 

jgfouca:
This approach leverages a single XML file to define namelist and yaml inputs for SCREAM/EAMxx.

## Discussion items:

### XML format for stuff that becomes YAML 

I originally had the YAML specification set up in XML such that the names of things were expressed as attributes instead of element names:

The original YAML:
```
  Time Stepping:
    Time Step: 300
    Start Time: [12, 30, 00]    # Hours, Minutes, Seconds
    Start Date: [2000, 2, 1]    # Year, Month, Day
    Number of Steps: 288
```

My first approach:
```
  <dict name="Time Stepping">
    <item name="Time Step">300</item>
    <list name="Start Time">12, 30, 0</list>
    <list name="Start Date">2000, 2, 1</list>
    <item name="Number of Steps">288</item>
  <\dict>
```
VS what I have currently:
```
  <Time Stepping>
    <Time Step>300</Time Step>
    <Start Time>12, 30, 0</Start Time>
    <Start Date>2000, 2, 1</Start Date>
    <Number of Steps>288</Number of Steps>
  <\Time Stepping> 
```
There are some pros to the first approach:
* There's no issue of spaces in the name. It's not legal XML to have spaces in element names, so the latter block is not valid XML. We will need to map spaces to underscores and have intentional underscores be double underscores.
* We can be explicit about the types of things instead of inferring them (dict/list/item).

The big drawback is that this format is very different from what's in namelist_defaults_eam.xml , so will be unfamiliar to users. Maybe we could have the format of YAML content be different than namelist content, but then we aren't even consistent across our own namelist.xml file.

### Do we need a namelist_definition.xml file?

EAM has both a namelist_defaults_eam.xml file and an ever bigger namelist_definition.xml file. Entries in the namelist definition file look like this:
```
<entry id="ncdata" type="char*256" input_pathname="abs" category="initial_conditions"
       group="cam_inparm" valid_values="" >                                                                                                                                                                                                      
Full pathname of initial atmospheric state dataset (NetCDF format).                                                                                                                                                                              
Default: set by build-namelist.                                                                                                                                                                                                                  
</entry>
```
I am currently thinking we do not want to have a definition file as it's too cumbersome and contains info that can be inferred. For SCREAM, this parameter would be
```
<cam_inparm>
  <ncdata>${DIN_LOC_ROOT}/path/to/data</ncdata>
</cam_inparm>
```
We can infer the group from the containing element (cam_inparm) and the type from the contents (it's not a valid int, float, or logical, so must be a string). This approach is a small divergence from EAM, but it may be worth it to avoid that awful namelist_definition file.

### How to handle attribute selectors

EAM seems to support a lot of convenient (but magical and undocumented) attribute selectors for defaults, example:
```
<ncdata dyn="se" hgrid="ne4np4"   nlev="128">some/path/data.nc</ncdata>
```
You can see "dyn", "hgrid", and "nlev" being used as selectors. I am thinking we want to limit the magic and just support selectors that can be xmlqueried, but that makes the above selection much more verbose:
```
<ncdata ATM_DYN_TARGET="se" GRID="a%ne4np4.*"   SCREAM_CMAKE_OPTIONS=".*SCREAM_NUM_VERTICAL_LEV 128.*" >some/path/data.nc</ncdata>
```
We can see that the removal of the magic selectors comes at a cost of more typing and uglier XML.


components/scream/cime_config/namelist_defaults_scream.xml
```diff
@@ -0,0 +1,130 @@
...
+      <Time Stepping>
+        <Time Step>300</Time Step>
+        <Start Time>12, 30, 0</Start Time>
+        <Start Date>2000, 2, 1</Start Date>
+        <Number of Steps>288</Number of Steps>
+      <\Time Stepping>
+      <Grids Manager>
+        <Type>Dynamics Driven</Type>
+        <Reference Grid>Physics GLL</Reference Grid>
+        <Dynamics Driven>
```
bartgol, line 66: Along the topic of simplifying our YAML files structure, this is another sublist that need not to exist. Unlike atm procs, we only build _one_ grids manager per run, so we don't need to have a sublist for each grids manager type, and we can just put all the params in the "Grids Manager" list.

components/scream/cime_config/namelist_defaults_scream.xml
```diff
@@ -0,0 +1,130 @@
...
+          <Enable Input Fields Checks>False</Enable Input Fields Checks>
+          <Grid>Physics GLL</Grid>
+        <\Process 1>
+        <Process 2>
+          <Process Name>CldFraction</Process Name>
+          <Enable Output Fields Checks>False</Enable Output Fields Checks>
+          <Enable Input Fields Checks>False</Enable Input Fields Checks>
+          <Grid>Physics GLL</Grid>
+        <\Process 2>
+        <Process 3>
```
bartgol, line 28: I think this format has the disadvantage that, in order to xmlchange/yamlchange/paramchange an option for a particular process, you need to know where that process is in the DAG. E.g., to change "SPA Remap File", you would have to know that SPA is in the sublist "Process 3".

I think we might benefit from some changes in our yaml structure, so that you don't need to know this kind of detail. E.g., for the param above, one should be able to do just "Atmosphere Driver"->"Processes"->"SPA".

rljacob:
The namelist_definition file is supposed to define everything that might possibly be in the namelist.  From build-namelist:
```
# Create a namelist definition object.  This object provides a method for verifying that the
# output namelist variables are in the definition file, and are output in the correct
# namelist groups.  Requires a namelist definition file...
#
# The namelist definition file contains entries for all namelist variables that
# can be output by build-namelist. ....
```

bartgol:
> There are some pros to the first approach:
> 
> * There's no issue of spaces in the name. It's not legal XML to have spaces in element names, so the latter block is not valid XML. We will need to map spaces to underscores and have intentional underscores be double underscores.
> * We can be explicit about the types of things instead of inferring them (dict/list/item).
> 
> The big drawback is that this format is very different from what's in namelist_defaults_eam.xml , so will be unfamiliar to users. Maybe we could have the format of YAML content be different than namelist content, but then we aren't even consistent across our own namelist.xml file.

XML is by definition a syntax-heavy language. If we stick with it, we might as well use a format that is more powerful (option 1), so that we can do some type checking, as you mentioned. Also, minor thing, but some text editors highlight strings, so one would be able to distinguish the param name from the XML noise.

> ### Do we need a namelist_definition.xml file?

The way I see it, namelist_definition.xml is useful for 1) documentation of what parameters are, and 2) enforcing valid parameter values. I don't know if there's any other place where the meaning of each parameter is stored. In a way, this serves as documentation for the model parameters. I suspect we might want to keep this, perhaps with less attributes. We could, tbf, merge it with nameslist_defaults.xml, but I suspect it would bloat the size of the xml file, and most regular users are probably only interested in the defaults, rather than the definition.

> ### How to handle attribute selectors
> EAM seems to support a lot of convenient (but magical and undocumented) attribute selectors for defaults, example:
> 
> ```
> <ncdata dyn="se" hgrid="ne4np4"   nlev="128">some/path/data.nc</ncdata>
> ```
> 
> You can see "dyn", "hgrid", and "nlev" being used as selectors. I am thinking we want to limit the magic and just support selectors that can be xmlqueried, but that makes the above selection much more verbose:
> 
> ```
> <ncdata ATM_DYN_TARGET="se" GRID="a%ne4np4.*"   SCREAM_CMAKE_OPTIONS=".*SCREAM_NUM_VERTICAL_LEV 128.*" >some/path/data.nc</ncdata>
> ```
> 
> We can see that the removal of the magic selectors comes at a cost of more typing and uglier XML.

I don't know. If `ncdata` depends on `nlev`, which itself depends on xmlqueriable params, I think it's fine. However, that means that `nlev` needs to be resolved before `ncdata`, so the xml entries need to be parsed in a specific order. This might not be too hard to do, meaning that we don't need to build a DAG of parameters. Something like (pseudo-code)
```
unresolved = all_params()
while not unresolved.empty():
  for (param : unresolved) {
    if (item in unresolved for item in param.dependencies()):
        continue
        # Might also check that param is not itself a dependency of item.
    else:
      resolve_param(param)
```
would do...assuming that there is no cyclic dependency, as I hinted in the pseudocode (e.g., nlev depends on ncdata and ncdata depends on nlev).

components/scream/cime_config/namelist_defaults_scream.xml
PeterCaldwell: I think it's reasonable for people doing xmlchanges to know what process they're trying to change - if they didn't, they don't have any business changing things! That said, I don't think we can expect them to know that SPA is "Process 3" - is there any way to list the processes by their *name* and to have process *number* be a property of each entry?

PeterCaldwell:
Regarding @rljacob 's comment that "The namelist_definition file is supposed to define everything that might possibly be in the namelist" - are you saying that we need to do this just because EAM does?

It seems to me like "everything that might possibly be in the namelist" should have a default setting and therefore should be set in our new namelist defaults file. I could see how namelist defaults wasn't sufficient in EAM because some settings could only have default values in use_case files, but I'm hoping to get rid of those.

components/scream/cime_config/namelist_defaults_scream.xml
bartgol: Yes, I've been thinking about this for a while, but put it on the (way) back burner. See #843 . That, in addition to #1449 lays a reasonable path for how to make the declaration of atm procs specs independent of their organization/order.

jgfouca: Good thoughts, and I did not address the user-mods aspect of things in this PR. If we like the general look of this prototype, we would not be using xmlchange since that only supports CIME case settings. It would likely be a (hopefully simple) homegrown script. Since it is homegrown, we would be free to add smarts to it so that names to things would not have to be fully qualified, ex:

`eamxx-config-change SPA::SPA_Remap_File $mynewfile`

bartgol:
I think of namelist_definition.xml as a "documentation" of the model. By checking that a specified parameter is in there, we force developers that add new options to document them somewhere (and hopefully they will document parameter "xyz" with something more than "value for xyz"). A sort of first barrier against lack of documentation.

In a way, this file could be seen as the model-parameter equivalent of the cf-database for the model input/output fields.

PeterCaldwell:
### About "magical attribute selectors:
Just to state the obvious: it's the existence of these input selectors that made us switch from yaml back to xml in the first place. Our dependence on these selectors will probably increase as we get rid of the multiple pathways for changing things (e.g. both namelist_defaults and use_cases => just namelist_defaults). Thus I think we probably want to make setting attribute selectors as easy to implement and easy to read as possible. I suspect using magic will help in that goal. Like Jeff said, perhaps we can just make the magic less mystical?

jgfouca:
@PeterCaldwell , @jeff-cohere , @rljacob , we could potentially add these convenience selectors to the definitions.xml file (assuming we want to keep that file).

jgfouca:
Just a general thought, having whitespaces in the names of things really mucks things up. Not only does it make the names unusable as XML elements, but it will also complicate any shell commands since the shell will interpret the space as starting a new argument. There are workarounds of course, but it reduces the usability of the system.

YAML made it easy to support items with whitespaces in names, so we took advantage of that in scream_input.yaml, but now that we are considering expanding the system to XML and CLI scripts, it probably makes sense to convert all these to an underscore-separator naming system.

jeff-cohere:
> Just a general thought, having whitespaces in the names of things really mucks things up. Not only does it make the names unusable as XML elements, but it will also complicate any shell commands since the shell will interpret the space as starting a new argument. There are workarounds of course, but it reduces the usability of the system.
> 
> YAML made it easy to support items with whitespaces in names, so we took advantage of that in scream_input.yaml, but now that we are considering expanding the system to XML and CLI scripts, it probably makes sense to convert all these to an underscore-separator naming system.

Yeah, as long as we're going with XML, maybe we should adopt the `snake_case` convention that seems most prevalent in XML dialects.

bartgol:
We can switch all our keys to snake_case in yaml too. I am probably responsible for the switch to Non Snake Case, since, as JIm said, yaml made that possible, and it seemed more "legible" to me. But the xmlchange/yamlchange feature is a compelling reason to abandon it, and revert to snake case.

jgfouca:
Another quick thought, how important is compatibility with EAM? Do we want users to be able to copy/paste stuff from eam_defaults.xml to scream_defaults.xml?

rljacob:
I believe namelist_definitions is used to error check namelist files including user_nl files.  So if you have a typo in user_nl, or try to add a line for a param that isn't in a namelist, the build-namelist can flag it.

jgfouca:
ATTN reviewers: this prototype is now fully functional and runs e3sm_scream_v1 fine on mappy. Please take a fresh look at this.

Some design decisions:
1) Parameters with selectors must have a default (no selectors) entry. Example:
```
          <Filename>some_file</Filename>
          <Filename COMPSET=".*SCREAM%AQUA.*">some_other_file</Filename>
```
If the first entry with no attributes is not present, then that is an error in the current implementation. Another potential approach is to just omit the parameter entirely if none of its entries' selectors match.
2) The namelist/yaml files are always recreated when buildnml is called. This means any direct modifications will be lost and therefore the user must always make changes through editing namelist_scream.xml (or running atm-config-chg).
3) I decided against having a definitions.xml file. Pretty much everything we need to know about a parameter can be inferred by looking at namelist_defaults_scream.xml. If documentation is needed, we can put XML comments in that file. This approach should simply the implementation quite a bit compared to what's currently in eam.
4) I have not yet tackled how to define and implement "convenience selectors", in other words, the ability to do this:
```
<ncdata dyn="se" hgrid="ne4np4"   nlev="128">some/path/data.nc</ncdata>
```
instead of 
```
<ncdata ATM_DYN_TARGET="se" GRID="a%ne4np4.*"   SCREAM_CMAKE_OPTIONS=".*SCREAM_NUM_VERTICAL_LEV 128.*" >some/path/data.nc</ncdata>
```

jeff-cohere (commented):
This looks good, Jim! I'd probably have to run it to get a better idea of how it works, but I like where we're headed.

jgfouca:
ATTN reviewers: a first crack at the implementation of "easy selectors" is done.

ndkeen (approved):
I will just have to try it

jeff-cohere:
> Discussion of another design point related to #1493. Do we want this new system to do validation (making sure all options are known and of the right type etc)? Also, if yes, do we want to do the validation at the python level (buildnml) or at the C++ level? Doing it an the python level would likely require a good deal of additional XML, specifically the namelist_definitions.xml file that I was trying to avoid.

@jgfouca , the operative question to me is: what is the source of truth/knowledge we use to decide how to validate input? One argument in favor of validating at the Python level is that we can create a document that people can look at (and argue over) that clearly articulates the "spec" against which we validate. If we validate within the C++ code itself, it might be more difficult to establish a correspondence between such a "spec" and the model.

Just my thoughts.

jgfouca:
@jeff-cohere , the only way to know for sure whether a value is known and has a valid value is in the C++/fortran code that is using the file. The definitions.xml would serve as documentation and for catching errors earlier, but it would have to be kept in sync with the main code.

PeterCaldwell:
I have an irrational(?) hatred of definitions.xml. Whenever I've added a new parameter in the past, I've added it to namelist_defaults.xml, then been flummoxed why I need to add it to namelist_definitions.xml as well. I feel like it would be cleaner if adding a parameter could be done in a single location. Though I guess changing an input/output variable in the rest of the C++ requires changing a billion files so maybe I shouldn't freak out about this extra function change for xml. 

I had the idea of enforcing legal parameter names at the C++ level by creating a copy of the yaml input data when the model starts running, then adding functionality to the yaml parser commands that deletes each parameter from the copied yaml data as it is read in. Then if any yaml input parameters are left after all the parser calls, the model crashes with a message about the remaining variables not being used. I think Luca said this wouldn't work - and that's fine - but I still don't understand quite why. Feel free to ignore.

PeterCaldwell:
Replies to Jim's overview above (which for convenience I note is here: https://github.com/E3SM-Project/scream/pull/1462#issuecomment-1078482559):

1. I think requiring each parameter to have a default to fall back on if none of the specific selectors matches is the way to go. I want our xml file to be the only place parameters can possibly be set rather than having default parameter values lurking in the bowels of the code in case no default parameter is passed to the exe.
2. Yup, users shouldn't touch yaml and nml files. We might consider putting these files in their own directory named something like `do_not_touch?` or `auto_generated_inputs`?
3. Yeah, I like not having definitions.xml, though we may need to add it for ensuring typing mistakes don't result in silently ignoring intended settings.

mt5555:
one feature (which some may consider a bug):  there are variables in namelist_definitions.xml which are not in namelist_defaults.xml.   This means they do not show up in the namelist, unless set in user_nl_eam.    This means developers can add them to user_nl_eam for testing / debugging, but the average user wont ever see them, and thus wont be tempted to change them.

rljacob:
That "feature" also allows CIME to check that entries put in user_nl_eam are valid.

bartgol:
I love purging files and/or lines of code, but I am surprisingly in favor of keeping namelist_definitiions.xml in EAMxx, actually. For a few reasons:

- Having definitions in the same name of defaults would make the defaults file much bigger and hard to parse. Rob also pointed out tha some parameters may not have a default (though I don't fully understand how that works). I would keep the defaults file _very_ minimal, without any metadata ( bounds, description, units...) in it. The avg user does not need to know how the default is set, after all, so they will rarely inspect this file (naive hope?).
- It forces people to write a minimal documentation. Yes, they can provide empty strings, but at least they get "pinged" by CIME, reminding them to add an entry in the definition xml file. I want to believe most people will spend 10 seconds coming up with some sort of description.
- YAML files do not enforce type. If you write `blah: 2`, the parser adds an entry of type `int` in the ParameterList. Later, when you try to access it with `params.get<double>("blah")` you will get an error, but you won't know exactly how to fix it, since the yaml file is generated automatically by CIME.
- Parsing the C++ code to see what the meaning of a parameter is (and maybe infer its bounds) is super challenging, especially if the parameter has a "good old" unintelligible name (like `swclq`...which I just made up, but I bet you thought for a second it might have been real), or it's used once in a very generic expression, where other variables also have generic names, making it hard to guess what it is (this should not happen in a well written code, but alas, I still have to see (let alone write!) a well written code).

Our buildnml python script can easily pre-load the definitions file, to extract valid bounds (if any), as well as the type, and a) check that the required value (if not defaulted) is valid, and b) set the yaml file correctly (e.g., using decimal point for floating point numbers).

components/scream/cime_config/buildnml
```diff
@@ -47,52 +48,144 @@ def do_cime_vars(entry, case):
...
+                        selector_regex = ".*" # Just grab the whole thing
+
+                    selectors[selector_name] = (selector_env, selector_regex)
+
             else:
-                compare = re.compile(components[0].strip())
-                if compare.match(value):
-                    best_val = do_switches(components[-1]).strip()
-                    break
+                expect(child.tag == "file", "Unexpected child.tag {}".format(child.tag))
```
bartgol, line 166: This only works now that we have two things to parse (an input.yaml file and a namelist.nl file) in the defaults. I suspect this will change very soon. But if you'd rather keep this check here for now, maybe add a comment warning, that this is only the case now, and might change in the future?

components/scream/cime_config/namelist_defaults_scream.xml
```diff
@@ -0,0 +1,154 @@
...
+        <partmethod nlev="72">420</partmethod>
+  -->
+  <selectors>
+    <selector name="hgrid" case_env="ATM_GRID"/>
+    <selector name="dyn" case_env="ATM_DYN_TARGET"/>
+    <selector name="nlev" case_env="SCREAM_CMAKE_OPTIONS" regex=".*SCREAM_NUM_VERTICAL_LEV ([0-9]+).*"/>
+  </selectors>
+
+  <file name="data/scream_input.yaml" format="yaml">
+    <Atmosphere__Driver>
```
bartgol, line 30: I will work to generate a more "plain" tree that can be more easily parsed by buildnml, and easier to display in the defaults file. I think `buildnml` can then put the blocks back together, and generate the current YAML structure again. Tbc, I'm thinking of

```
<atm_procs_names>homme, p3shoc, spa, rad<atm_procs_names>
<homme>
  <Enable__Output__Fields__Checks>False</Enable__Output__Fields__Checks>
  <Grid>dynamics</Grid>
  <!-- other params for homme -->
</homme>
<p3shoc>
  <Process__Type>Gruop</Process__Type>
  <Schedule__Type>Sequential</Schedule__Type>
  <Processes__Names>p3, shoc</Processes__Names>
  ...
</p3shoc>
<p3>
...
</p3>
<spa>
 <SPA__Data__File hgrid="ne4">spa_ne4_file.nc</SPA__Data__File>
 <SPA__Data__File hgrid="ne30">spa_ne30_file.nc</SPA__Data__File>
</spa>
...
<Grids__Manager>
 ...
</Grids__Manager>
<Initial__Conditions>
  <Start__Date> ... </Start_Date>
  <Input__Files>blah, blah2, ... </Input_Files>
</Initial__Conditions>
<Scorpio__Output>
  <Output__YAML__Files> a.yaml, b.yaml </Output__YAML__Files>
</Scorpio__Output>
<Model__Restart>
  <Output__Control>
     <Frequency>10</Frequency>
     <Frequency__Units>Steps</Frequency__Units>
  </Output__Control>
  <Casename>my_restart</Casename>
</Model__Restart>
...
```
This way, our `atmchange` python util will allow for one (maybe two?) indirection layers, like
```
./atmchange Initial__Conditions::Start__Date="2000-01-01.00:00:00"
./atmchange Model__Restart::Output__Control::Frequency=10
...
```
Then, buildnml can put together a yaml file, since we know which sections need to be present, and how they are nested. Better yet, I could flatten a bit our YAML file, without all those layers of nesting.

Note: I think what you did was the best choice, since we don't want this to be a single PR. Adding the defaults reflecting the current input files structure is the best first step. Next, comes structuring the xml files in a more legible/skinny way. Just making a note here.

components/scream/scripts/atm-config-chg
```diff
@@ -0,0 +1,120 @@
...
+    expect(change.count("=") > 0,
+           "Bad change request '{}', expected format is 'param=val'".format(change))
+
+    param_namespace, new_value = change.split("=", maxsplit=1)
+
+    for child in element:
+        child_name = child.tag
+
+        has_children = False
+        for _ in child:
```
bartgol, line 41: Can this also be `has_children = len(child)>0`?

components/scream/scripts/atm-config-chg
```diff
@@ -0,0 +1,120 @@
...
+
+###############################################################################
+def change_matches(param_namespace, parent, child_name):
+###############################################################################
+    param_ns_re = re.compile(".*" + param_namespace + "$")
+    m = param_ns_re.match(parent + "::" + child_name)
+    return m is not None
+
+###############################################################################
+def perform_change(element, change, parent, any_change):
```
bartgol, line 30: If I understand correctly, if a parameter has 2+ namespaces, say `A::B::param_name`, the change has to be issued as
```
atm-config-chg B::param_name=blah
```
Is that correct? If so, any reason not to force the user to do
```
atm-config-chg A::B::param_name=blah
```
?

components/scream/scripts/yaml-to-xml
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env python3
```
bartgol, line 1: Was this just to generate an xml from our current `scream_input.yaml` file? Or do you see it being used during case.xyz? If it's the former, do we need to keep it? Once we have our `namelist_defaults_scream.xml` file, we should not need to manually create a yaml file anymore, right? Even less so, generate an xml from a yaml...

bartgol (approved):
This is great Jim. I have a few questions/comments, most of them pointing maybe at a need to put a few comments here and there, to explain the data organization and the big picture of some functions.

I think this is a great starting point, and I will get start (this week?) on a simplification of our yaml inputs, which should, hopefully, allow a skinnier, less nested, and more readable namelist_defaults.xml file.

bartgol:
Imho, we should not iterate too many times on the current impl (assuming people have no big/drastic concern). I think it's best to merge soon, and immediately open a round-2 pr, tackling other aspects, and/or polishing the current implementation. This way, other people can (in theory) start using this, and quickly point out strengths/weaknesses of the approach.

jeff-cohere (approved):
I agree with Luca. There's a lot of good ideas floating around and stuff we could haggle over, but there's no reason not to get started and see how things work in practice.

components/scream/cime_config/buildnml
```diff
@@ -47,52 +48,144 @@ def do_cime_vars(entry, case):
...
+    else:
+        with open(src, "r") as fd:
+            tree = ET.parse(fd)
+            root = tree.getroot()
+
+        selectors = None
+        selectors_elem = None
+        for child in root:
+            if selectors is None:
+                expect(child.tag == "selectors", "Unexpected child.tag {}".format(child.tag))
```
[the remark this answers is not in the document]
jgfouca: What this means is that we expect the first big organizational block of XML in `namelist_defaults_scream.xml` to be the block that defines our "easy selectors". This block is currently mandatory as I think we agreed that we will want "easy selectors" like those available in eam's nml config files. If we did decide to go with a definitions.xml file, this would be moved there.

components/scream/cime_config/buildnml
jgfouca: I'm not sure I understand the problem here. `namelist_defaults_scream.xml` is used to produce program inputs will which always be in the form of one or more files. Currently, every major organizational block in this XML (with the exception of the easy selectors) should be defining an input file.

components/scream/cime_config/namelist_defaults_scream.xml
jgfouca: Yes, I think your refactors will simplify the situation. Also, we won't need these annoying double underscores once we switch to "snake case" or whatever you call: this_naming_convention.

components/scream/scripts/atm-config-chg
jgfouca: It will match however many namespaces you provide. "B::param" and "A::B::param" will both match. The fewer namespace qualifications the user provides, the more chance of an ambiguity, which is an error. If there is no ambiguity, then the user can use shorter names to refer to things.

components/scream/scripts/yaml-to-xml
jgfouca: Yes, we can probably remove this once the PR is ready.

components/scream/cime_config/buildnml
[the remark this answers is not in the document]
bartgol: Ah, I see now. The `if selectors is None:` line forces this check to run only on the first entry of the XML. Got it.

components/scream/cime_config/namelist_defaults_scream.xml
bartgol: Yes, I think that's snake_case.

components/scream/cime_config/buildnml
```diff
@@ -72,20 +75,27 @@ def ordered_dump(data, item, Dumper=yaml.SafeDumper, **kwds):
 ###############################################################################
 def evaluate_selectors(element, case, ez_selectors):
 ###############################################################################
+    """
+    Evaluate and remove selectors from the unprocessed XML nml file in the repo.
+
+    Elements with selectors are removed. If the selector evaulates to True, then
+    the corresponding element text becomes the new text value of the original
```
bartgol, line 82: Does the script check that no two unprocessed XML entries have a selector evaluating to true? E.g, if `hgrid=ne4`, and `nlevs=72`

```
<my_value hgrid="ne4"> Jim </my_value>
<my_value nlevs="72"> Not Jim </my_value>
```
would buildnml spot the error or will it simply pick the last one that ends up evaluating to true? Or, worse, both?

jgfouca:
OK, all conflicts have been resolved and I think I have addressed all of the non-controversial feedback. I am changing the settings to let AT hit this and merge it if tests pass.

jgfouca:
As a side note for further discussion, I will point out that there's no official testing of any of this new capability. I did a few simple tests by hand and, other than that, just relied on the e3sm_scream_v1 suite (and a quick glance at the produced nml/yaml files). 

The core of the new functionality is only a couple hundred lines of mostly straightforward python, so maybe it is ok to skip intensive testing?

components/scream/cime_config/buildnml
jgfouca: @bartgol . Good question. The current implementation is that multiple matches are allowed and the match that occurs latest has the highest precedence.

jgfouca: To answer your question, "Not Jim" would be set for my_value.

bartgol: I think that's probably ok for now, so long as whoever adds new entries to the defaults.xml file pays good attention, ensuring that no duplicates are possible after selector resolution. But perhaps, since we have the chance to improving things, we want to be safe down the road, and either prevent this, or at least warn the user.

bartgol: Actually, warning the user might be not too hard, right? We can scan the current content of the fully resolved xml, and verify that the new entry is not there, no?

Tbc, I'm ok merging this without implementing this safeguard, but I do think we should add it at some point.

jgfouca: We also need to keep in mind the behavior of eam's namelist XML system. I'm OK with diverging from them when it suits our needs, but I'd prefer to stay compatible with them if it's all the same to us. It's clear from looking at their file that it works the way I described above. For example, just looking within the top few lines of this file, I see:

```
<dtime dyn="fv"                    >1800</dtime>
<dtime dyn="fv"    waccmx="1">300 </dtime>
```

This pattern only makes sense if multiple matches are supported with the later match taking precedence. I suppose its also possible that precedence is derived from the number of selectors matched.

jgfouca:
@bartgol , there were definitely some bugs in our first version of `buildnml`. I will see if I can think of a nice way to add some tests.

PeterCaldwell:
> As a side note for further discussion, I will point out that there's no official testing of any of this new capability.

Some obvious/trivial tests would be to check that xml defaults for all entries don't fall into the case where there's no default for some conditions or multiple acceptable solutions for some typical cases (as noted above). Making sure options passed to xml file are either valid or raise an error is another thing we've discussed.

Of course, the above are tests of the stuff people add to the functionality rather than of the functionality itself. Having unit tests checking what happens when you pass xml with those bad conditions would be useful, but I'm not sure how to accomplish that - create special "test" xml entries to query?

components/scream/cime_config/buildnml
PeterCaldwell: I've always intuitively expected these entries to choose the most specific option (ie. the one with the most selectors). But I guess choosing the "most recent match" by just running through each line and reassigning the variable every time you have a match might be easier. If we establish an expectation that the simplest case will be first and the most complex last, most recent match could be ok.

bartgol: Ah, good point. There might be two selectors that both eval to true, but one is more specific than the other. We can live with that for now, but boy, that _does_ look like a source of sneaky bugs!

Unfortunately, our buildnml would become quite complicated if we had to keep track of "independent" duplicates. We'd have to store some extra lists, to check if one option is a "specialization" of the other, in which case, all is good, and possibly throw an error if two xml entries have selectors that eval to true, but are "independent".

Given that EAM still works the way it does, I _assume_ this scenario never happened, so I'm ok with living with it.

jgfouca:
Oof, yet another discussion point: 

In our system, outer elements function as namespaces, so this allows reuse of names for inner elements:
```
  <file name="test.yaml" format="yaml">
    <group1>
      <testval>41</testval>
    </group1>
    <group2>
      <testval>hi</testval>
    </group2>
  </file>
```

We currently must support this approach because there is reuse of inner names going on in our YAML file. This approach defeats the definitions.xml idea, at least the way it's done in eam, which assumes a very flat defaults.xml file. I suppose we could duplicate the element structure of defaults.xml in the definitions.xml:

```
  <file name="test.yaml" format="yaml">
    <group1>
      <entry id="testval" type="int">Documentation here</entry>
    </group1>
    <group2>
     <entry id="testval" type="string" valid_values="hi,there,buddy">Documentation here</entry>
    </group2>
  </file>
```

... but this comes at additional cost/tedium in the definitions.xml file and more code to support all this. Perhaps it is worth it to have documentation and rigorous checking at the buildnml level?

bartgol:
Do you know who does the check against the definitions.xml file? Is it CIME (in which case we might be in trouble) or is it EAM (in which case we can find a solution)?

jgfouca:
@bartgol , it is definitely not in CIME. I believe EAM does this stuff somewhere in the huge perl files under `eam/bld`. Since several other components follow a similar pattern, it's possible there's a shared perl lib somewhere that they all use. In any event, we will not be using the same library since we are avoiding perl.

components/scream/cime_config/buildnml
jgfouca: Oof, and yet another idea: we could enforce that selections are more specific as additional selection elements are added:

```
<foo>default</foo>
<foo s1="a">1</foo>
<foo s1="a" s2="b">2</foo>
```

^^^ This would be ok since the selections are increasingly "specific" for each sequential selection element.

```
<foo>default</foo>
<foo s1="a" s2="b">2</foo>
<foo s1="a">1</foo>
```

^^^ This would be an error. Something like "specificity inversion for foo".

components/scream/cime_config/buildnml
jgfouca: I forgot to follow up on this. If I'm understanding you right, buildnml is already doing what you described. There are two expected attributes for the `file` element, `name` and `format`. The `name` can be something like `data/scream_input.yaml`, which tells buildnml where to put the file relative to RUNDIR.

jgfouca:
I've added a pretty good set of doctests, so we can at least tuck away that discussion point. As usual, making the tests was a bit tedious, but I am glad that I did it because I found a number of small issues.

PeterCaldwell:
> In our system, outer elements function as namespaces, so this allows reuse of names for inner elements
It seems like a no-brainer to me that definitions.xml would need to follow the same hierarchical organization that defaults.xml does since that hierarchy is fundamental to the meaning of the parameters. Is it such a bad thing for definitions.xml to be hierarchical? And can't the python code we will call to check the supplied parameters be recursive over the levels of hierarchy? I envision it being something like  "loop over all entries in a level of hierarchy; for each entry, check that it exists in definitions and that it has an entry, then call this script to check over all its children, then move to the next entry"

jgfouca:
@PeterCaldwell , 

> Is it such a bad thing for definitions.xml to be hierarchical?

No, not a huge deal. It would just be a bit different from eam and would require additional coding.

PeterCaldwell:
Yeah, I agree that getting rid of "Process 0" sorts of entries would be good and I don't think it is a big deal to need to register all the valid atm processes - that list shouldn't grow too rapidly. I'm not clear where in the code the <GROUPS> stuff would go though. Is this in a section of defaults.xml set aside from all the rest of the info about SHOC and P3, or is the P3 and SHOC stuff contained within this <GROUPS> tag? Like 
```
<GROUPS>
  <P3__SHOC>
    <scheduling>sequential</scheduling>
    <P3>
      p3_parameter1='blah'
    <P3>
    <SHOC>
      shoc_parameter1='boo'
    <SHOC>
   </P3__SHOC>
</GROUPS>
```

bartgol:
> One other thought: namelist_defaults_eam.xml has lots of weird parameters left over from cases folks probably ran decades ago. A good example is
> 
> ```
> <megan_factors_file>atm/cam/chem/trop_mozart/emis/megan21_emis_factors_c20130304.nc</megan_factors_file>
> ```
> 
> It would be a bummer if all of the cruft like this that builds up in our repo-version of defaults.xml file gets copied into the xml file we create for a specific run case. Perhaps we _should_ let parameters like this that are only needed for particular configurations not have a default value outside those cases. Then if a parameter doesn't match any criteria, it simply isn't added to the case.xml file. Is this what we already do?

Whoops! I accidentally clicked on "edit" instead of "quote reply". Here's my reply.

This is a more general question: if parameter `FOO` does not have any selector, does/should it always end up in the intermediate xml file? As of now, I think the answer is that it does. But we should not copy stuff that is not needed by the current case... I'm not sure how to enforce this though. I think the only way to do it, is to make these sort of inputs have a selector that only evaluates to true for that particular case.

bartgol:
> Yeah, I agree that getting rid of "Process 0" sorts of entries would be good and I don't think it is a big deal to need to register all the valid atm processes - that list shouldn't grow too rapidly. I'm not clear where in the code the stuff would go though. Is this in a section of defaults.xml set aside from all the rest of the info about SHOC and P3, or is the P3 and SHOC stuff contained within this tag? Like
> 
> ```
> <GROUPS>
>   <P3__SHOC>
>     <scheduling>sequential</scheduling>
>     <P3>
>       p3_parameter1='blah'
>     <P3>
>     <SHOC>
>       shoc_parameter1='boo'
>     <SHOC>
>    </P3__SHOC>
> </GROUPS>
> ```

That section was to be found in the intermediate XML file. The defaults one, only contains sections for SHOC and P3. Possibly, a "generic" section for a GROUP, which tells buildnml what kind of params a GROUP_XYZ section should contain.

bartgol:
New idea: @jgfouca  how easy would it be to do the following in our defaults file:
```
<atm_proc_base>
  <number__of__subcycles>1</number__of__subcycles>
  <enable__precondition__checks>false</enable__precondition__checks>
  <enable__postcondition__checks>false</enable__postcondition__checks>
</atm_proc_base>

<atm_proc_group inherit="atm_proc_base">
  <schedule__type>sequential</schedule__type>
</atm_proc_group>

<SHOC inherit="atm_proc_base">
    <enable__precondition__checks>true</enable__precondition__checks>
</SHOC>
```

Rationale: we have a hierarchy of atm proc classes. Some options are common to _all_ atm procs (e.g., number of subcycles). We might be ok with specifying a default value for all of them, once and for all. The defaults for a derived class can "inherit" the defaults of the base class, and possibly override some.

jgfouca:
@bartgol , it would not be too hard, I'm just worried that rules of using the file will become too complex for users to understand. That said, I really hate copy/paste design, so I'm willing to add an "inherit" concept if we all agree on it.

jgfouca:
@bartgol , @PeterCaldwell ,

My feeling is this. If we do not want an item to appear in the $case/namelists.xml file, we can add a "omit_if_empty" metadata attribute to parameter. Example:

```
<unwanted_param omit_if_empty="true"></unwanted_param>
<unwanted_param unlikely_selector="almost never">some value</unwanted_param>
```

With the above, `unwanted_param` will not get added to the $case xml unless `unlikely_selector` evaluates to True.
```

**Comment by bartgol** [comment]
> @bartgol , @PeterCaldwell ,
> 
> My feeling is this. If we do not want an item to appear in the $case/namelists.xml file, we can add a "omit_if_empty" metadata attribute to parameter. Example:
> 
> ```
> <unwanted_param omit_if_empty="true"></unwanted_param>
> <unwanted_param unlikely_selector="almost never">some value</unwanted_param>
> ```
> 
> With the above, `unwanted_param` will not get added to the $case xml unless `unlikely_selector` evaluates to True.

If you already have the selector, do you really need the first line? I mean, if the selector is false, `unwanted_param` will not enter the processed xml file, no?

**Comment by jgfouca** [comment]
@bartgol , currently, it is illegal to have a parameter with "selection" elements but no "default" element.

**Comment by bartgol** [comment]
> @bartgol , currently, it is illegal to have a parameter with "selection" elements but no "default" element.

Is that an XML limitation, a python XML module limitation, or something that CIME and/or buildnml enforces?

**Comment by jgfouca** [comment]
@bartgol , none of those, just a rule we (I?) decided to adopt for us (scream).

**Comment by jeff-cohere** [comment]
> @bartgol , none of those, just a rule we (I?) decided to adopt for us (scream).

@jgfouca , how are you feeling about this rule in the light of this conversation? Do you feel like it's more clear to use `omit_if_empty` than to just omit parameters without default elements?

**Comment by jgfouca** [comment]
@jeff-cohere , it is probably better to not have to support another metadata flag.

**Comment by bartgol** [comment]
> > @bartgol , it would not be too hard, I'm just worried that rules of using the file will become too complex for users to understand. That said, I really hate copy/paste design, so I'm willing to add an "inherit" concept if we all agree on it.
> 
> I'm also a bit worried about hiding defaults with "inheritance" (adding more assumptions to how these XML files are interpreted), but I get why this could be useful. On the other hand, a defaults file that explicitly gives defaults for everything isn't necessarily "copy/paste design"--it's a catalog of all parameters and their defaults. Arguably, the debate about which way to go rests on whether you want to make life easier for developers or for users.

Well, consider the example I made. The entries `Number of Subcycles`, `Enable Precondition Checks`, and `Enable Postcondition Checks` are options of the base class, and they are handled _completely_ by the base class. The XML is just reflecting that hierarchy. Also in C++ we have the base class establishing some defaults for the parameters.

As the code grows in complexity, classes will grow too, with new configurable options, and as SCREAM moves to EAMxx, there will be maybe 10+ atm procs types, each of which needs to set 3+ parameters that perhaps rarely (if ever) require a different default value in the XML.

**Comment by bartgol** [comment]
Perhaps the word "inherit" can be changed into something that is less overloaded, though all alternatives I can think of also have 1+ other meanings ("include", "derive")

**Comment by jeff-cohere** [comment]
Having thought about this more, I think "inherit" is actually appropriate here, even if you don't have in mind the relationship between a C++ base class and its subclasses. A parameter set for any given atmosphere process "inherits" defaults from its archetype. So I guess that syntax works for me.

**Comment by PeterCaldwell** [comment]
As I started this comment, I was strongly in favor of changing our "all xml entries must have a default" to "all xml entries lacking a default are understood to be omitted from the generated xml/yaml/nml file unless matching a specific case". As I think about it more, though, I think being didactic/clear about what to expect in those intermediate files is really important. The code is going to get chaotic and it might be hard to tell from just looking at an entry whether it has a default...Actually, I'm changing my mind again. If each entry starts either with or without a default value, it should always be easy to tell whether it is intended to always be included or not. And I think if the documentation we write for this xml stuff clearly states that entries lacking a default are omitted unless matching a specific case, any user smart enough to create an entry should be able to figure out how to include/exclude stuff. I take you along with me on my thought journey in case it spurs ideas of your own.

**Comment by PeterCaldwell** [comment]
I'm wondering whether we want to integrate this PR now. It seems like conversation has died down?

**Comment by bartgol** [comment]
I'm in favor of integrating this, unless @jgfouca is working on fixes/polishing.

**Comment by jgfouca** [comment]
@PeterCaldwell , @bartgol , @jeff-cohere , @AaronDonahue , 

I think this is all ready to go. I have left block inheritance for a future PR.  I think the main point of interest will be the documentation at the top of `namelist_defaults_scream.xml` which explains how the file works.

File: components/scream/cime_config/buildnml [hunk 10]
```diff
@@ -3,19 +3,27 @@
 """
 Namelist creator for E3SM's SCREAM component
 
-This script is mostly for processing the scream_input.yaml file.
+This scripts takes the namelist_defaults_scream.xml file from the repo
+and produces the processed $case/namelist_scream.xml. From the processed
+file, raw input files are produced which should not be modified by users.
```
PeterCaldwell, line 8: Maybe add a comment here saying that users can modify $case/namelist_scream.xml? Upon first read, it sounded like defaults-> case xml file -> raw input files without any opportunity for users to modify defaults. Also, it might be easier for readers to understand if "raw input files" was "scream_input.yaml and namelist.nml"... though I get that you're trying to avoid comments which will become wrong when we modify the code to read in xmls.

components/scream/cime_config/buildnml
```diff
@@ -31,68 +39,524 @@ import yaml
...
+    >>> do_cime_vars('hi ${foo}${bar} there', case)
+    'hi 12 there'
+    >>> do_cime_vars('hi ${baz_bar} there', case)
+    'hi blah there'
+    >>> do_cime_vars('hi ${baz2_bar2} there', case)
+    'hi 4 there'
+    >>> do_cime_vars('hi ${invalid} there', case)
+    Traceback (most recent call last):
+      ...
+    CIME.utils.CIMEError: ERROR: Cannot resolve yaml entry 'hi ${invalid} there', CIME has no value for 'invalid'
```
PeterCaldwell, line 77: is it still a yaml entry, or an xml entry now?

components/scream/cime_config/buildnml
```diff
@@ -31,68 +39,524 @@ import yaml
...
+    def get_value(self, key):
+        if key in self._kv_dict:
+            return self._kv_dict[key]
+        else:
+            return None
 
 ###############################################################################
 def do_cime_vars(entry, case):
 ###############################################################################
+    """
```
PeterCaldwell, line 60: Would a more descriptive function name be "dict_entries_to_string"?

components/scream/cime_config/namelist_defaults_scream.xml
```diff
@@ -0,0 +1,210 @@
...
+      case_env : The case value (thing you can xmlquery)
+      regex    : OPTIONAL, how to pick out the piece of the value you are interested in
+                 By default, the entire value is grabbed. If the regex does not match
+                 the case value, then the selector will evaluate to False.
+
+    Example using the nlev selector
+      <partmethod nlev="128">42</partmethod>
+      <partmethod nlev="72">420</partmethod>
+
+    Selector info will not be duplicated in the processed $case/namelist_scream.xml file.
```
PeterCaldwell, line 33: So what happens if 2 selectors evaluate to true? You keep the last entry in that case rather than throwing an error, right? Might be worth saying that here.

components/scream/cime_config/namelist_defaults_scream.xml
```diff
@@ -0,0 +1,210 @@
...
+        Parameter element: An XML element representing an individual input parameter.
+                           This element should not have child elements. It should have element text
+                           representing the desired value. These elements can have
+                           one or more selectors.
+        Selector: An XML attribute that checks a property of the case. See the documentation
+                  above the <selectors> element above for more details.
+        Metadata: An XML attribute that represents metadata for an input parameter
+
+      Rules:
+        1) Parameters can have any number of parameter elements. A parameter element
```
PeterCaldwell, line 57: Maybe change to "Parameters can have any number of parameter elements - one for each different value of a selector"? As written, it sounded a bit to me like "parameter elements can have any number of parameter elements"

components/scream/cime_config/namelist_defaults_scream.xml
```diff
@@ -0,0 +1,210 @@
...
+           with no selectors, which means it controls the default param value, must
+           come before parameter elements with selectors. If multiple selection elements
+           are matched, then the match the occurs last takes precendence.
+        2) The first parameter element encounter for a parameter will define the default metadata
+           for that parameter. All subsequent parameter elements will assume the same
+           metadata as the first unless an explicit override occurs.
```
PeterCaldwell, line 63: It sounds a bit odd that the *last* selector's value takes precedence but the *first* selector defines metadata.

components/scream/cime_config/namelist_defaults_scream.xml
```diff
@@ -0,0 +1,210 @@
...
+        3) For a parameter whose elements all have selectors, if none of the selectors
+           are matched, then the parameter will be omitted from the inputs.
```
PeterCaldwell, line 65: I suggest "... if none of the selectors are matched, then the parameter will be omitted from $case/scream_inputs.yaml". "from the inputs" sounds too vague to me.

components/scream/cime_config/namelist_defaults_scream.xml
```diff
@@ -0,0 +1,210 @@
...
+        4) Parameter types will be inferred. You can override the inferred type
+           via the 'type' metadata attribute. Types are:
+             logical: will be inferred from any case of true/false
+             integer: will be inferred from any sequence of 0-9
+             real   : will be inferred from any numeric expression with a decimal point or scientific notation
+             string : everything else
+        5) A parameter can be set to any value of the correct type unless
+           the 'valid_values' metadata attribute is set. If desired, this
+           metadata should be set up as a comma-separated list of values.
```
PeterCaldwell, line 74: Do you mean that the set of all metadata can be specified as a comma-separated list (e.g. "type=integer, valid_values={0,1,2},..." or that "valid_values=0,1,2" is ok? I assume the latter, in which case I'd suggest replacing "this metadata" with "valid_values"

components/scream/cime_config/namelist_defaults_scream.xml
```diff
@@ -0,0 +1,210 @@
...
+           TODO support numeric checks (<, >, ...)??
+        6) Parameter metadata can be overriden in subsequent parameter elements, but
+           by default, metadata will be the same as the first parameter element. The
+           final metadata for a parameter will reflect what's in the last matching
+           parameter element. User changes to the case XML file must be compatible
```
PeterCaldwell, line 79: I think rules 2 and 6 are somewhat redundant or at least very related. Perhaps merge them? In particular, I think rule 2 says that final metadata for a parameter will reflect what's in the last matching parameter element _which explicitly sets that metadata variable_. 

This entry also made me realize that I keep getting confused by the difference between a "parameter element" and a "parameter". Maybe add a definition for parameter to your "lingo" section to clarify that a particular parameter can have many parameter elements?

components/scream/cime_config/namelist_defaults_scream.xml
```diff
@@ -0,0 +1,210 @@
...
+        7) You can inline case values into parameter values via this syntax:
+           <param>MY GRID IS ${ATM_GRID}</param>
+           This would result in parameter 'param' having the value 'MY GRID IS ne4ne4'
```
PeterCaldwell, line 82: This example confused me because creating a phrase is far from our typical usage. Perhaps switch to an example more related to our work? Like "<init_file> scream_init_${ATM_GRID}.nc</param}"?

components/scream/cime_config/namelist_defaults_scream.xml
```diff
@@ -0,0 +1,210 @@
...
+        8) Any parameter value with a comma will be assumed to be a list.
+           TODO allow this behavior to be overridden?
```
PeterCaldwell, line 84: When would we want commas to _not_ imply a list? I can only think of strings... so wouldn't strings naturally be protected by their quotes? E.g. `<param> "I love dogs, pigs, and chickens"</param>`

components/scream/cime_config/namelist_defaults_scream.xml
```diff
@@ -0,0 +1,210 @@
...
+
+  -->
+
+  <file name="data/scream_input.yaml" format="yaml">
+    <Atmosphere__Driver>
+      <Atmosphere__Processes>
+        <Number__of__Entries>6</Number__of__Entries>
+        <Schedule__Type>Sequential</Schedule__Type>
+        <Process__0>
+          <Process__Name>Homme</Process__Name>
```
PeterCaldwell, line 97: dumb question, but why are all the underscores double-underscores in this file? Should this be added to the rules?

bartgol: That's a workaround to the fact that xml does not support spaces in the element tag. Two underscores signal buildnml that it should replace them with a space when generating the yaml file. Adding a single underscore would have been problematic, since it is something we _could_ get in a valid name.

bartgol:
I want to point out that there's a follow-up PR that is changing quite a bit the formatting of our XML file, and how we generate both the intermediate XML as well as the nml/yaml files. I would not dwindle too long on fixing things in this PR, since things might change in the second iteration anyways.

components/scream/cime_config/namelist_defaults_scream.xml
```diff
@@ -0,0 +1,210 @@
...
+<namelist_defaults>
+
+  <!--
+    The following block defines convenience selectors. Case settings
+    can be used as selectors, but they are often not covenient due to
+    long names or having to process their values through regular expressions
+    to get the subpiece you are interested in. The convenience selectors
+    allow you to express this logic and capture it with a short name.
+
+    The follow attributes are how one defines a selector
```
jeff-cohere, line 22: follow -> following

components/scream/cime_config/namelist_defaults_scream.xml
jeff-cohere: Would it be worth printing something to a log or the terminal when this occurs, to prevent someone from doing it unintentionally?

components/scream/cime_config/namelist_defaults_scream.xml
bartgol: Well, I think it's cleaner to change our formatting in yaml files, from `Atmosphere Driver` to `atmosphere_driver`, so we don't need any shenanigan like that...

components/scream/cime_config/namelist_defaults_scream.xml
jeff-cohere: Maybe we could mention in the definition of "parameter element" that the value in a parameter element is the "parameter" associated with it. This is one of those things that seems obvious when you read it, but that doesn't mean it's obvious before you do.

components/scream/cime_config/namelist_defaults_scream.xml
jeff-cohere: Another downstream PR, right?

jeff-cohere (approved):
This looks good to me. There are a few nips and tucks that could be done, but doing them in a followup PR is also fine.

components/scream/cime_config/namelist_defaults_scream.xml
bartgol: Yes, separate one. But coming soon, so I don't think documenting the "__" vs " " issue is necessary.

jeff-cohere:
Whoops! Didn't mean to terminate our conversation. I forgot AUTOMERGE was set. Sorry, all.

jgfouca:
@jeff-cohere , no worries, I was planning for follow-on PRs.

components/scream/cime_config/buildnml
jgfouca: This function gets values from the CIME Case object, it only looks like a dict lookup because of the MockCase implementation.

components/scream/cime_config/namelist_defaults_scream.xml
jgfouca: That behavior is documented lower in this file, but it couldn't hurt to have it here too.

components/scream/cime_config/namelist_defaults_scream.xml
jgfouca: Yeah, it is a bit subtle, but I think it's consistent. The first "matched" (either no selectors, or all selectors are True) parameter element defines the default metadata for the parameter, but subsequent matched elements can override the default and these overrides will take precedence.

components/scream/cime_config/namelist_defaults_scream.xml
jgfouca: We can't assume the parameter is for scream_inputs.yaml. I can clarify that it will be eliminated from the processed case XML file.

components/scream/cime_config/namelist_defaults_scream.xml
jgfouca: A agree, we always want commas to denote a a list. We can revisit this if that is ever not true.

components/scream/cime_config/namelist_defaults_scream.xml
jgfouca: A temporary workaround. Our scream_input.yaml took advantage of the fact that YAML allows keys to have whitespace in them, e.g. "My Var". Unfortunately, in XML, "<My Var>" is not allowed, so we have to jump through some hoops for everything to work.

PR #1465, opened 2022-03-10
Fix SPA timestate structure and spa_main to address test fails

AaronDonahue:
This commit addresses two fails that show up in our testing on mappy.
In particular, we observe a valgrind fail and a fail in spa_main_test
for the release build.

1. The valgrind fail is due to a conditional on uninitialized memory.
   In this commit we initialize the SPA TimeState structure by setting
   the month to an impossible month value of -1 at construction.  That
   way in spa_update_data when it checks what month it is, there is
   initialized memory to compare with.
2. The spa_main_test has been updated to use the approx_equal function
   that is used in the common physics functions tests when comparing two
   values that should be equal.  Now we check the two values are equal to
   within a specific tolerance.

Additionally, for code cleanup, this commit adopts the routine defined in
common physic functions tests that creates a host mirror and immediately
deep copies.  Which cleans up the code a bit.

### DIFF
--- a/components/scream/src/physics/spa/spa_functions.hpp
+++ b/components/scream/src/physics/spa/spa_functions.hpp
@@ -61,8 +61,8 @@ struct SPAFunctions
     SPATimeState() = default;
     // Whether the timestate has been initialized.
     bool inited = false;
-    // The current month
-    Int current_month;
+    // The current month, initialize to an impossible month like -1.
+    Int current_month = -1;
     // Julian Date for the beginning of the month, as defined in
     //           /src/share/util/scream_time_stamp.hpp
     // See this file for definition of Julian Date.
--- a/components/scream/src/physics/spa/tests/spa_main_test.cpp
+++ b/components/scream/src/physics/spa/tests/spa_main_test.cpp
@@ -32,10 +32,34 @@ using Spack = SPAFunc::Spack;
 // Helper Functions
 std::pair<Real,Real> compute_max_min(const view_1d<Real>::HostMirror& input, const int start, const int end);
 
+template<typename ScalarT>
+struct ChecksHelpers {
+
+  static bool is_non_negative (const ScalarT& s) {
+    return not ( (s<0 || std::isnan(s)) );
+  }
+  static bool equal (const ScalarT& lhs, const ScalarT& rhs) {
+    return lhs==rhs;
+  }
+  static bool approx_equal (const ScalarT lhs, const ScalarT rhs,
+                            const ScalarT tol) {
+    using std::abs;
+    return not ( abs(lhs-rhs)>=tol );
+  }
+};
+
+// Helper function. Create Mirror View and Deep-Copy (CMVDC)
+template<typename ViewT>
+auto cmvdc (const ViewT& v_d) -> typename ViewT::HostMirror {
+  auto v_h = Kokkos::create_mirror_view(v_d);
+  Kokkos::deep_copy(v_h,v_d);
+  return v_h;
+}
 TEST_CASE("spa_read_data","spa")
 {
   using C = scream::physics::Constants<Real>;
   static constexpr auto P0 = C::P0;
+  static constexpr auto test_tol = C::macheps*1e1;
 
  /* ====================================================================
   *                  Test Setup, create structures, etc.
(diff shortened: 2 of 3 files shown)


bartgol:
@AaronDonahue I removed the labels to skip blake/weaver. I think we should make sure that whatever changes you add don't compromise results on other machines. I think those labels are to be used sparingly, to skip testing on a machine that we know is broken, or to test some change that can possibly impact just a specific machine (like a machine file change).

components/scream/src/physics/spa/tests/spa_main_test.cpp
```diff
@@ -59,7 +59,7 @@ TEST_CASE("spa_read_data","spa")
 {
   using C = scream::physics::Constants<Real>;
   static constexpr auto P0 = C::P0;
-  static constexpr auto test_tol = C::macheps*1e1;
+  static constexpr auto test_tol = C::macheps*1e2;
```
bartgol, line 62: It looks like this tolerance is only used in places where a tolerance should not even be needed (namely, the  case where there's no time/vertical interpolation).

It's true that that's the check that is failing, but I think the tolerance should not be the issue. In fact, I believe those tests should pass with tol=0, even in finite precision arithmetic. That's because `y2=y1+dy/dx*(x-x1)==y1` if x==x1, regardless of roundings. Similarly for time interpolation. Allowing a tolerance might actually be a bad idea (covering up failing comparisons that should not fail).

bartgol: Note, for time interpolation, `t_norm = (t_now - t_beg) / t_len`, so for the case t=t_beg we clearly expect exact results. One may argue that for `t=t_end` we could get `t_norm` slightly different from 1.0. That's true for general doubles, but, at least in this test, we should have t_now, t_beg, and t_end integers, so they can be represented exactly. That is, when we do
```
ts += util::days_in_month(ts.get_year(),ts.get_month())*86400;
spa_time_state.t_now = ts.frac_of_year_in_days();
```
we should have `t_now -t_beg` be _exactly_ `t_len`. So even for the case where t equals the end of the month (or, as it's probably better to say, for the beg of the next month), we should have _exact_ time interpolation.

bartgol: It might also be beneficial to add some checks on the time state. SInce tests pass on other archs, I suspect they are ok, but better stress test the test. E.g., stuff like
```
REQUIRE (time_state.days_this_month==31); // it's January
REQUIRE (time_state.t_beg_month==0);
REQUIRE (time_state.t_now==0); // for 1st test
REQUIRE (time_state.t_now==31); // for 2nd test
```

AaronDonahue:
> @AaronDonahue I removed the labels to skip blake/weaver. I think we should make sure that whatever changes you add don't compromise results on other machines. I think those labels are to be used sparingly, to skip testing on a machine that we know is broken, or to test some change that can possibly impact just a specific machine (like a machine file change).

@bartgol , Agreed.  I only pinged those because the first round of AT passed on `weaver` and `blake` and I was trying to figure out which tolerance was needed to pass on `mappy`.  So I didn't want to waste AT cycles testing on those machines since the only new changes were a higher tolerance, which would have passed too.  I wasn't quite ready to have this reviewed, but if I labeled the PR as WIP it wouldn't have ran the AT.  Since I don't have access to `mappy` and I couldn't get a recreator on my local machines: `blake`, `lassen` and `quartz` I was kinda stuck relying on the AT to let me know how bad the test was failing.

components/scream/src/physics/spa/tests/spa_main_test.cpp
AaronDonahue: Maybe this is a more severe fail than I thought.  I can go ahead and delete most of this PR and just keep the `valgrind` fix in.  Which is the initialization of the `TimeState.month`.  And spend a bit more time investigating why the `==` fails.  I agree we would expect it to pass in cases where we have no time interpolation.  It's perplexing because it only fails in `release` build on `mappy`.  It passes in `release` build on all the machines I have access too.  But they are all intel compilers, so it is likely related to compiler.

bartgol:
> Since I don't have access to `mappy` and I couldn't get a recreator on my local machines: `blake`, `lassen` and `quartz` I was kinda stuck relying on the AT to let me know how bad the test was failing.

is there a gcc modules stack on quartz?

components/scream/src/physics/spa/spa_functions.hpp
```diff
@@ -61,8 +61,8 @@ struct SPAFunctions
     SPATimeState() = default;
     // Whether the timestate has been initialized.
     bool inited = false;
-    // The current month
-    Int current_month;
+    // The current month, initialize to an impossible month like -1.
+    Int current_month = -1;
```
AaronDonahue, line 65: @bartgol , this is the only change I added to address the valgrind fail for the `spa_main` test.  Can you add it to #1486 ?  Then this PR can just be closed since it won't be needed.  TBH - since I don't have access to mappy I'm not sure if this will solve the valgrind fail, but I tracked the fail down to a conditional that uses `current_month` so I believe this will address the issue.

AaronDonahue:
Fixed by #1486

PR #1471, opened 2022-03-11
SCREAM: disable baselines logic for certain builds in test-all-scream

bartgol:
For fpe/valg/cmc/cov builds we are not interested in comparing against baselines. Instead, we simply want to run the code to check for possible errors/leaks or to verify coverage.

### DIFF
--- a/components/scream/CMakeLists.txt
+++ b/components/scream/CMakeLists.txt
@@ -214,6 +214,12 @@ set(SCREAM_BASELINES_ONLY FALSE CACHE BOOL "Whether building only baselines-rela
 # has the sole purpose of detecting memory errors. For these builds, we can disable baselines tests
 set(SCREAM_ENABLE_BASELINES_TESTS TRUE CACHE BOOL "Whether to run baselines-related tests")
 
+if (SCREAM_BASELINES_ONLY AND NOT SCREAM_ENABLE_BASELINES_TESTS)
+  message (FATAL_ERROR
+    "Makes no sense to set SCREAM_BASELINES_ONLY=ON,\n"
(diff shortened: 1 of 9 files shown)


jgfouca (approved):
Good enhancement!

bartgol:
@jgfouca Note: the `rrtmgp_allsky_baseline.nc` baselines where a bit tricky. After talking with @brhillman , we agreed that those baselines are different from, say, p3/shoc run_and_cmp ones. In particular, rrtmgp is not comparing output of an old scream version with the current version (to check for answer-changing mods). Instead, the baselines are generated running raw rrtmgp, and later used to check the output of scream::rrtmgp (i.e., rrtmgp run through the AD). In this respect, these tests are akin to f90-cxx comparison, and "baselines" need not to be stored in scream's baseline repo (of the current machine), but can be generated on the fly.

This allows to run these rrtmgp tests also in valgrind/cuda-memcheck builds, without having to ensure baselines are present on the machine. For instance, we can run a quick valgrind test on blake (where we don't have baselines for valgrind) without having to first build master (or the baseline ref) to generate baselines.

PeterCaldwell:
Maybe we should call the RRTMGP "baselines" something different to avoid confusion? Perhaps "compare_against_raw" or something like that?

jgfouca (approved):

bartgol:
> Maybe we should call the RRTMGP "baselines" something different to avoid confusion? Perhaps "compare_against_raw" or something like that?

We might. I don't have a strong preference here.

bartgol:
Only the expected release build fail on mappy, due to spa_main_test (fix coming soon). Merging.

PR #1472, opened 2022-03-11
SCREAM: fix compilation error on CUDA

bartgol:
#1411 (merged ~4h ago) introduced a cmake bug for our v1 build on CUDA.

The `-Xcudafe xyz` flag for the `scream_control` lib was PUBLIC, and not language specific, so when linking control to the `atm` lib, it caused F90 compilation errors.


tcclevenger:
This is great to see. I've been debugging this error for the past few hours...

tcclevenger (approved):

bartgol:
Note: the mappy failure is due to #1411 , which introduced non BFB diffs in the vert coordinates. I ran master CIME tests on mappy, and launcher the bless job. Now the baselines should be up to date, and we can retest this PR.

bartgol:
This fails on mappy just b/c of spa_main (which is a separate issue, already being tackled elsewhere). Merging.

PR #1473, opened 2022-03-11
Apply scaling to incoming solar flux to account for orbital eccentricity

brhillman:
Apply scaling to incoming solar flux in radiation interface to account for changing earth-sun distance due to orbital eccentricity. This follows the implementation in EAM to apply the eccentricity factor computed in `shr_orb_mod` to the incoming solar flux read from the RRTMGP input data. This factor was already computed in `RRTMGPRadiation::run_impl` in the call to `shr_orb_decl`, but we needed an additional argument added to `rrtmgp_main` and `rrtmgp_sw` to pass this factor down to where it is applied. The `input_unit.yaml` file for the test `rrtmgp_standalone_unit` is modified to set eccentricity (and the other orbital parameters) to zero to keep from scaling fluxes in this test, since we are comparing against a reference problem that does not include any scaling in that test. This will be BFB for all v1 standalone tests, but diff for the v1 CIME tests.

### DIFF
--- a/components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
+++ b/components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
@@ -544,7 +544,7 @@ void RRTMGPRadiation::run_impl (const int dt) {
     lw_flux_up, lw_flux_dn, 
     sw_bnd_flux_up, sw_bnd_flux_dn, sw_bnd_flux_dir,
     lw_bnd_flux_up, lw_bnd_flux_dn, 
-    get_comm().am_i_root()
+    eccf, get_comm().am_i_root()
   );
 
   // Compute and apply heating rates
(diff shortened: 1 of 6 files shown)


brhillman:
Comparison showing downward SW flux at the top of the model in SCREAMv0, SCREAMv1 before this fix, and SCREAMv1 with fix applied in this PR. Note that to get the correct scaling, I needed to set the `Orbital Year`  in the `scream_input.yaml` file in the case directory to `Orbital Year = 1990` to be consistent with what is used in the F2010-SCREAM-LR compset in v0 (yes, we use a perpetual year 1990 orbital year for the solar forcing for a compset labelled 2010...I do not know why).
apply_toa_scaling SW_flux_dn

AaronDonahue (approved):
Looks good to me.  Thanks for putting this in Ben.

brhillman:
@AaronDonahue yes this is expected, aquaplanet uses constant orbital parameters that specify no eccentricity or obliquity, so there will be no scaling of incoming solar irradiance in this case and hence the changes here will not affect aquaplanet.

AaronDonahue:
Great! Then I'm going to manually merge, because the only fail is expected.

PR #1477, opened 2022-03-15
Finalize scream session for CIME runs

tcclevenger:
SCREAM was not being finalized in `mct_coupling`. Thus, CUDA was not being finalized. With this fix, CIME runs of SCREAMv1 run on weaver.

### DIFF
--- a/components/scream/src/mct_coupling/scream_cxx_f90_interface.cpp
+++ b/components/scream/src/mct_coupling/scream_cxx_f90_interface.cpp
@@ -213,6 +213,9 @@ void scream_finalize (/* args ? */) {
 
     // Clean up the context
     scream::ScreamContext::singleton().clean_up();
+
+    // Finalize scream session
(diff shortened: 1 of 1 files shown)


ndkeen:
Noting that I also needed this for runs on PM and have used this edit for all attempts.

tcclevenger:
I am skipping the testing on weaver and blake since they don't do a screamv1 run through cime.

bartgol (approved):
There's a hidden issue that we might have to tackle at some point, in order to run a model coupled with ocn and/or land ice. The issue is that the scream finalization routine also finalizes Kokkos. However, ocn and land-ice _might_ use Kokkos as well, and finalizing Kokkos before they're done with it, might be a problem.

I think this is fine for now, but it's something to keep in mind. We really need to start a discussion on how to handle Kokkos uniformly in E3SM probably...

AaronDonahue (approved):

PR #1481, opened 2022-03-16
Add arg to jsrun for summit

tcclevenger:
Adds jsrun arg `--smpiargs="-gpu"`. Needed for CIME runs on summit to not fail.

@jgfouca The 2nd commit is something that I'm curious if it is a typo. From [the open mpi FAQ](https://www.open-mpi.org/faq/?category=running#:~:text=OMPI_COMM_WORLD_RANK) it seems as if the variable should have an "I". But I could be missing something..


jgfouca (approved):

jgfouca:
Skipping testing because these changes cannot impact any machine on which we do CI testing.

PR #1483, opened 2022-03-16
Change FieldAllocProp, to store alloc size as 64 bit int

### LINKED ISSUE #1480: Field alloc prop should store alloc size as a size_t
Currently, `FieldAllocProp` uses `int` for all sizes. While this is ok for stuff like `m_last_extent` or `m_pack_size_max`, it is clearly not ok for the allocation size, since large arrays can easily lead to more than 2GB allocations.

bartgol:
This avoids overflows when running very large cases on a very small number of ranks.

Close #1480


tcclevenger (approved):
Looks good to me.

bartgol:
Only the usual spa_main_test fails. Merging.

PR #1484, opened 2022-03-17
Some changes to build on Perlmutter with gnu v10

ndkeen:
Only real change is to add the compiler flag for gnu v10 to ignore the type mismatch errors.


components/scream/cmake/machine-files/perlmutter.cmake
```diff
@@ -1,11 +1,25 @@
 # Load all kokkos settings from Ekat's mach file
 set (EKAT_MACH_FILES_PATH ${CMAKE_CURRENT_LIST_DIR}/../../../../externals/ekat/cmake/machine-files)
-include (${EKAT_MACH_FILES_PATH}/kokkos/nvidia-a100.cmake)
-include (${EKAT_MACH_FILES_PATH}/kokkos/cuda.cmake)
+IF (USE_CUDA OR KOKKOS_ENABLE_CUDA)
```
bartgol, line 3: Does this actually work? IIRC, the machine file is loaded by cmake _before_ any command line option (the `-DBLAH=xyz`) are parsed. That means that there's no way to specify KOKKOS_ENABLE_CUDA in order to cause a switch in the mach file. At least _I don't think_ this might be a portable solution.

But I will have to check.

components/scream/cmake/machine-files/perlmutter.cmake
```diff
@@ -1,11 +1,25 @@
...
+else()  
+  include (${EKAT_MACH_FILES_PATH}/kokkos/amd-zen2.cmake)
+  #include (${EKAT_MACH_FILES_PATH}/kokkos/openmp.cmake)
+  include (${EKAT_MACH_FILES_PATH}/kokkos/serial.cmake)
+endif()
 
 #option(Kokkos_ARCH_AMPERE80 "" ON)
 set(CMAKE_CXX_FLAGS "-DTHRUST_IGNORE_CUB_VERSION_CHECK" CACHE STRING "" FORCE)
 
+if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
```
bartgol, line 15: This is another thing I'm not sure it's portable. The compiler ID is set by cmake only after the language has been enabled (e.g., during the call to `project()`), so I suspect this cannot be set by the time the mach file is parsed...

Again, it might be this works, but I am not confident it's portable. I should check.

components/scream/cmake/machine-files/perlmutter.cmake
ndkeen: I was lazy and used OR on both of those, but it looks like it's only USE_CUDA being set (so could remove the other)

For cime cases at least, with `--compiler=gnu`:
```
-- NDK Perlmutter USE_CUDA=FALSE KOKKOS_ENABLE_CUDA=
-- NDK Perlmutter CMAKE_CXX_COMPILER_ID=GNU CMAKE_Fortran_COMPILER_VERSION=10.3.0
```

and with `--compiler=gnugpu`:
```
-- NDK Perlmutter USE_CUDA=TRUE KOKKOS_ENABLE_CUDA=
-- NDK Perlmutter CMAKE_CXX_COMPILER_ID=GNU CMAKE_Fortran_COMPILER_VERSION=10.3.0
```

bartgol: Yes, for CIME cases it might be ok, since Macros.cmake sets USE_CUDA. I am not sure where those print statements were, but I tried on my machine to put a print of `CMAKE_Fortran_COMPILER_VERSION` right before and right after the call to `project` in `components/CMakeLists.txt`, and they print, respectively, "" and "8.5.0". Was the second print statement before the call to project? If so, then I'm puzzled.

bartgol:
@jgfouca what if stand-alone scream did the same thing that CIME does, and _include_ the mach file in our CMakeLists.txt file? Right now, our test scripts pre-load the mach file via `-C $machfile`, which has some limitations. E.g., in a pre-loadable cache file, you are only allowed to set CACHE variables. But also, we cannot do something like what Noel is trying to do: set certain vars based on other cache vars. It _would_ be nice to do

```
if (SOME_CACHE_VAR)
  # DO X
else ()
  # DO Y
endif()
```
but we cannot do that in a CACHE file, since `Kokkos_ENABLE_CUDA` is not set yet. If the mach file was included by our main CMakeLists (before calling `project()`, so we can set compilers stuff too), then the user could do

```
test-all-scream ... -c "SOME_CACHE_VAR=ON"
```
and trigger "DO X".

components/scream/cmake/machine-files/perlmutter.cmake
ndkeen: Those were in this file I'm updating `components/scream/cmake/machine-files/perlmutter.cmake`.  But I think your point is that we have (at least) 2 ways to build -- with cime and without. As I'm only testing with cime, might be problematic otherwise. Is there a way I can test on how it's being built within this file?  Like `IF (CIME_BUILD)`?  If so, does it make sense to use something like that here? I mean, in general, you can see what I'm trying to do here -- what's the best way?

bartgol: Yeah, exactly, we are doing 2 different things in standalone vs CIME builds.

If this is a CIME build, the mach file is loaded _before_ entering the scream dir, but _after_ the call to `project(E3SM...)`. So you could probably check the content of `PROJECT_NAME`. This should be "E3SM" for a CIME build, and empty for a standalone build (since we use the mach file as a cache to pre-load, which happens before scream's call to `project()`).

bartgol: I think what you are doing is fine. I also think we should prob switch to use the mach file in the same way that CIME does (i.e., including rather than using as a cache pre-load). I think for now checking PROJECT_NAME should be ok.

ndkeen: I added a change

bartgol (approved):
I think the if guard with PROJECT_NAME should make this work also for standalone builds.

jgfouca (approved):

ndkeen:
How to move forward?

AaronDonahue:
It looks like this has been approved so we should be able to set it up to merge.  The AT fails because of an old test (unrelated) which has been fixed.  If we want to make sure this is a clean merge we can rebase and retest.  I'll go ahead and do that.

AaronDonahue:
I'm going to approve this branch so it can be merged.  It was already approved by @jgfouca before I rebased, so presumably it doesn't really need to be re-reviewed.

PR #1486, opened 2022-03-17
Make the implementation of spa_main more modular, to improve testing granularity

### LINKED ISSUE #1463: Spa unit tests fail in release mode
The spa_main_test fails in release mode, due to roundoff diffs:

```
  REQUIRE( ccn3_h(dof_i,kpack)[kidx] == ccn3_beg(dof_i,kpack_pad)[kidx_pad] )
with expansion:
  3.9652588367 == 3.9652588367
```

The test also failed in the PR unit test. It's worth to point out that, if the AT encounters a fail in standalone tests, it sill proceeds to run CIME tests. However, on the PR page, the standalone tests result are hidden, and one might miss info regarding standalone tests output.

@AaronDonahue I'm gonna assign you, since you are the spa wizard.

@jgfouca we should improve jenkins output, so that standalone fails are not hidden. To be fair,
```
+ [[ 100 != 0 ]]
+ fails=1+1
+ [[ 2 > 0 ]]
+ exit 1
```
is showing that 2 commands failed (and in this case they are standalone and v1 testing), but we could have the script state things clearly at the end.

### LINKED ISSUE #1469: SPA creates ncol*nlev and ncol*nlev*nbands temporaries at each time step
At the beginning of `spa_main`, we have the snippet

```  
    // Set up temporary arrays that will be used for the spa interpolation.
    view_2d<Spack> p_src("p_mid_src",ncols_atm,pressure_state.nlevs),
                   ccn3_src("ccn3_src",ncols_atm,pressure_state.nlevs);
    view_3d<Spack> aer_g_sw_src("aer_g_sw_src",ncols_atm,nswbands,pressure_state.nlevs),
                   aer_ssa_sw_src("aer_ssa_sw_src",ncols_atm,nswbands,pressure_state.nlevs),
                   aer_tau_sw_src("aer_tau_sw_src",ncols_atm,nswbands,pressure_state.nlevs),
                   aer_tau_lw_src("aer_tau_lw_src",ncols_atm,nlwbands,pressure_state.nlevs);
```
These are expensive allocations, which could instead be done once during initialization (and possibly using the ATM buffer, to limit memory consumption).

bartgol:
This PR modifies a tiny bit the SPAXyz structures, and splits `spa_main` into separate function calls (and separate kernels). The testing for the individual sub-functions is done on completely random data, to increase robustness. The only subroutine not tested is `compute_source_pressure_levels`, since I have no idea how to test it without copy-pasting the code itself (which is not really a test). However, it's small enough that I'm confident we can live with it.

Fixes #1469 
Fixes #1463 
Closes #1465

### DIFF
--- a/components/scream/src/physics/spa/atmosphere_prescribed_aerosol.cpp
+++ b/components/scream/src/physics/spa/atmosphere_prescribed_aerosol.cpp
@@ -61,62 +61,64 @@ void SPA::set_grids(const std::shared_ptr<const GridsManager> grids_manager)
   add_field<Computed>("aero_tau_sw",    scalar3d_swband_layout, nondim, grid_name,ps);
   add_field<Computed>("aero_tau_lw",    scalar3d_lwband_layout, nondim, grid_name,ps);
 
-  // Set of fields used as input and output
-  // - There are no fields used as both input and output.
+  // Init output data structure
+  SPAData_out.init(m_num_cols,m_num_levs,m_nswbands,m_nlwbands,false);
 }
 // =========================================================================================
 int SPA::requested_buffer_size_in_bytes() const
 {
-  const Int num_mid_packs    = ekat::npack<Spack>(m_num_levs);
-  const Int num_int_packs = ekat::npack<Spack>(m_num_levs+1);
+  using PackInfo = ekat::PackInfo<Spack::n>;
+  const int num_mid_packs = PackInfo::num_packs(m_num_levs);
 
+  // We have two 2d mid view (p_mid_src, spa_data.ccn), three
+  // 3d swbands mid views (aer_g_sw, aer_ssa_sw, aer_tau_sw),
+  // and one 3d lwbands mid view (aer_tau_lw)
   // Number of Reals needed by local views in the interface
-  const int interface_request =
-      // 1d view scalar, size (ncol)
-      Buffer::num_1d_scalar*m_num_cols*sizeof(Real) +
-      // 2d view packed, size (ncol, nlev_packs)
-      Buffer::num_2d_vector*m_num_cols*num_mid_packs*sizeof(Spack) +
-      Buffer::num_2dp1_vector*m_num_cols*num_int_packs*sizeof(Spack);
+  const int num_reals = m_num_cols*num_mid_packs*(2 + 3*m_nswbands + m_nlwbands);
 
-  // Number of Reals needed by the WorkspaceManager
-  const auto policy       = ekat::ExeSpaceUtils<KT::ExeSpace>::get_default_team_policy(m_num_cols, num_mid_packs);
-  const int wsm_request   = WSM::get_total_bytes_needed(num_mid_packs, 3, policy);
-
-  return interface_request + wsm_request;
+  return num_reals*sizeof(Real);
 }
 
 // =========================================================================================
 void SPA::init_buffers(const ATMBufferManager &buffer_manager)
 {
-  EKAT_REQUIRE_MSG(buffer_manager.allocated_bytes() >= requested_buffer_size_in_bytes(), "Error! Buffers size not sufficient.\n");
+  EKAT_REQUIRE_MSG(
+      buffer_manager.allocated_bytes() >= requested_buffer_size_in_bytes(),
+      "Error! Buffers size not sufficient.\n");
 
-  Real* mem = reinterpret_cast<Real*>(buffer_manager.get_memory());
+  using PackInfo = ekat::PackInfo<Spack::n>;
 
-  // 1d scalar views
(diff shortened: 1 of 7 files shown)


components/scream/src/physics/spa/spa_functions_impl.hpp
```diff
@@ -90,195 +90,187 @@ void SPAFunctions<S,D>
...
+  const SPAInput&   data_beg,
+  const SPAInput&   data_end,
+  const SPAInput&   data_tmp,
+  const SPAOutput&  data_out)
+{
+  // Number of bands must match.
+  EKAT_REQUIRE(data_end.data.nswbands==data_beg.data.nswbands);
+  EKAT_REQUIRE(data_end.data.nlwbands==data_beg.data.nlwbands);
+
+  // For now we require input and output to have same geometric sizes
```
AaronDonahue, line 103: Minor, but you probably mean "beginning and ending" data instead of "input and output"

bartgol: Actually, I mixed up things. I need
- beg/end _must_ have same geo/bands dims (meaning spa data has same grid for all months), otherwise it's a mess.
- input/output to have same ncols, since we don't do horiz interp in spa_main (for now)
- input/output must have same bands dims (i don't even know why we wouldn't, or how we would handle that).

I'll fix the checks/comments.

components/scream/src/physics/spa/atmosphere_prescribed_aerosol.cpp
```diff
@@ -155,14 +175,21 @@ void SPA::initialize_impl (const RunType /* run_type */)
...
   // Update the local time state information and load the first set of SPA data for interpolation:
   auto ts = timestamp();
   SPATimeState.inited = false;
   SPATimeState.current_month = ts.get_month();
   SPAFunc::update_spa_timestate(m_spa_data_file,m_nswbands,m_nlwbands,ts,SPAHorizInterp,SPATimeState,SPAData_start,SPAData_end);
+
+  // NOTE: we *assume* hybrid v coordinates don't change with time.
+  //       IF this ever ceases to be the case, you need to remove these
+  //       lines, and have spa_main interpolate those during the call
+  //       to performe_time_interpolation.
```
AaronDonahue, line 190: I suspect that if we ever allow the hybrid coordinates to change with time then we would just store them in SPAData_start and SPAData_end and when we construct pmid_src we would have to first construct pmid_start and pmid_end with the respective hya/b and then do time_interpolation between those two.

Not that I see that happening anytime soon.

components/scream/src/physics/spa/spa_functions_impl.hpp
```diff
@@ -90,195 +90,204 @@ void SPAFunctions<S,D>
...
+      data_end.data.nlevs==data_tmp.data.nlevs,
+      "Error! SPAInput data structs must have the same number of columns/levels.\n");
+
+  // Output must have same number of bands
+  EKAT_REQUIRE_MSG (
+      data_end.data.nswbands==data_out.nswbands &&
+      data_end.data.nlwbands==data_out.nlwbands,
+      "Error! SPAInput and SPAOutput data structs must have the same number of SW/LW bands.\n");
+
+  // While horiz interp is not supported, we also require matching ncols
```
AaronDonahue, line 119: This isn't actually accurate.  We do support horizontal interpolation, we just don't do it in `spa_main` because it would be wasteful to do horizontal interpolation every timestep.  Instead, when we read from file the `data_beg` and `data_end` data we conduct the horizontal interpolation right away.  I don't think we will ever want to do horizontal interpolation in `spa_main` for that reason.

components/scream/src/physics/spa/spa_functions_impl.hpp
```diff
@@ -90,195 +90,204 @@ void SPAFunctions<S,D>
...
+  EKAT_REQUIRE_MSG (
+      data_end.data.ncols==data_out.ncols,
+      "Error! While horizontal interpolation is not supported in spa_main,\n"
+      "       SPAInput and SPAOutput data structs must have the same number columns.\n");
+
+  // Step 1. Perform horizontal interpolation (todo)
```
AaronDonahue, line 125: See comment above.  Maybe "Step 1. Perform horizontal interpolation (already done in `update_spa_data`"

AaronDonahue (commented):
Thank you @bartgol for doing such a great job cleaning up the spa interface and spa routines.  These changes make it much cleaner and improve the testing.  I don't have any changes to recommend except a couple comments about the comments.

components/scream/src/physics/spa/spa_functions_impl.hpp
bartgol: Ah, I probably misread the comment in the original spa main implementation. That makes sense.

AaronDonahue (approved):

PR #1489, opened 2022-03-18
Make default timestep 5min for SCREAMv1 CIME cases

brhillman:
Change the default timestep for SCREAMv1 CIME cases to 5 minutes for all resolutions. This is currently required because SHOC and P3 assume a 5 min timestep, and we do not yet have the ability to substep these processes in EAMxx. This will be non-BFB for the SCREAMv1 CIME cases because it changes the timestep for those tests, and also renames the tests to remove the testmod that was used to force the timestep to 30 minutes (which we know was not quite optimal for SHOC and P3).

### DIFF
--- a/driver-mct/cime_config/config_component_e3sm.xml
+++ b/driver-mct/cime_config/config_component_e3sm.xml
@@ -408,6 +408,8 @@
       <!-- a shorter timestep is needed for stability in MMF  -->
       <!-- Use for all "low res" grids (i.e. ne4, ne30, ne45) -->
       <value compset="_EAM%MMF">72</value>
+      <!-- Shorter timestep needed for SCREAMv1 because we are not substepping SHOC and P3 yet -->
+      <value compset="_SCREAM.*">288</value>
     </values>
     <group>run_coupling</group>
     <file>env_run.xml</file>
--- a/cime_config/tests.py
+++ b/cime_config/tests.py
@@ -447,8 +447,8 @@ _TESTS = {
     "e3sm_scream_v1" : {
         "time"  : "03:00:00",
         "tests" : (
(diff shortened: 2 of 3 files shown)


driver-mct/cime_config/config_component_e3sm.xml
```diff
@@ -408,6 +408,8 @@
       <!-- a shorter timestep is needed for stability in MMF  -->
       <!-- Use for all "low res" grids (i.e. ne4, ne30, ne45) -->
       <value compset="_EAM%MMF">72</value>
+      <!-- Shorter timestep needed for SCREAMv1 because we are not substepping SHOC and P3 yet -->
+      <value compset="_SCREAM.*">288</value>
```
PeterCaldwell, line 412: So this is using a 5 min dt for all SCREAM compsets regardless of model grid? And previously we defaulted to a half hour dt if we didn't explicitly choose dt? Yeah, this sounds like a good change until we get substepping working.

PeterCaldwell (approved):

driver-mct/cime_config/config_component_e3sm.xml
brhillman: Right. Well, almost. The previous behavior would key off the entries above that set based on resolution, since no defaults were set for SCREAM compsets. So actually for ne4 it would default to a 2-hour timestep based on this entry:
```
      <value compset=".+" grid="a%ne4np4">12</value>
```
we were just overriding that with the `scream-30min_ts` testmod to get 30 minutes. For ne30, we would have gotten 30 minutes because the overall default if no xml keys are matched was 48.

PeterCaldwell: 👍

brhillman:
I believe this is just throwing the expected DIFFs in the SCREAMv1 CIME tests, and should be ready to merge.

PR #1491, opened 2022-03-18
Use 10 tracers and field checking by default

PeterCaldwell:
Since SCREAMv1 only supports prescribed aerosol, it only has 10 tracers. We were initializing space for 35 of them. This PR fixes that. Also, scream_input.yaml has an option for checking that each process doesn't return bad data (e.g. concentrations <0). We had turned this off for a while because of some overly sensitive value, but I think that's fixed. So turning it back on. I'm not positive that field checking won't cause test fails... using the AT to confirm that.

### DIFF
--- a/components/scream/data/scream_input.yaml
+++ b/components/scream/data/scream_input.yaml
@@ -7,38 +7,40 @@ Atmosphere Driver:
     Schedule Type: Sequential
     Process 0:
       Process Name: Homme
-      Enable Output Fields Checks: true
-      Enable Input Fields Checks: true
+      Enable Postcondition Checks: true
+      Enable Precondition Checks: true
       Grid: Dynamics
-      Vertical Coordinate Filename: "<${COMPSET} : .*SCREAM%AQUA.* => ${DIN_LOC_ROOT}/atm/scream/init/scream-aquaplanet_init_ne4np4_L72_20211202.nc : ${DIN_LOC_ROOT}/atm/scream/init/init_ne4np4.nc>"
+      Vertical Coordinate Filename: "<${COMPSET} : .*SCREAM%AQUA.* =>
+      ${DIN_LOC_ROOT}/atm/scream/init/scream-aquaplanet_init_ne4np4_L72_20211202.nc
+      : ${DIN_LOC_ROOT}/atm/scream/init/init_ne4np4.nc>"
       Moisture: moist
     Process 1:
       Process Name: SHOC
-      Enable Output Fields Checks: true
-      Enable Input Fields Checks: true
+      Enable Postcondition Checks: true
+      Enable Precondition Checks: true
       Grid: Physics GLL
     Process 2:
       Process Name: CldFraction
-      Enable Output Fields Checks: true
-      Enable Input Fields Checks: true
+      Enable Postcondition Checks: true
+      Enable Precondition Checks: true
       Grid: Physics GLL
     Process 3:
       Process Name: SPA
-      Enable Output Fields Checks: true
-      Enable Input Fields Checks: true
+      Enable Postcondition Checks: true
+      Enable Precondition Checks: true
       Grid: Physics GLL
       SPA Remap File: none
       SPA Data File: ${DIN_LOC_ROOT}/atm/scream/init/spa_file_unified_and_complete_ne4_scream.nc
       # TODO: Fix the SPA remap file and data file to be dependent on the grid resolution.  For now we hardcode the ne4 case.
     Process 4:
       Process Name: P3
-      Enable Output Fields Checks: true
-      Enable Input Fields Checks: true
+      Enable Postcondition Checks: true
+      Enable Precondition Checks: true
       Grid: Physics GLL
     Process 5:
       Process Name: RRTMGP
-      Enable Output Fields Checks: true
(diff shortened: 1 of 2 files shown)


PeterCaldwell:
So I don't forget over the weekend - I'd also like this PR to add print statements with the name of each process as we enter it. This is a temporary modification that would really help us figure out where the model is crashing and would be taken out when we're ready for production runs. I think the place to make this addition is in the run_sequential function within components/scream/src/share/atm_process/atmosphere_process_group.cpp but I can't figure out how to extract a processes' name. I think L363 of atmosphere_process_dag.cpp provides an example but I haven't had time to grok it yet.

components/scream/data/scream_input.yaml
```diff
@@ -7,38 +7,38 @@ Atmosphere Driver:
     Schedule Type: Sequential
     Process 0:
       Process Name: Homme
-      Enable Output Fields Checks: false
-      Enable Input Fields Checks: false
+      Enable Output Fields Checks: true
```
bartgol, line 10: I apologize, I should have update this script when the syntax changed.

Would you mind changing all these entries into:
```
Enable Postcondition Checks: ...
Enable Precondition Checks: ...
```
?

bartgol (commented):
So sorry, but there's one thing to change, due to a chance in syntax that I forgot to propagate to this file.

components/scream/data/scream_input.yaml
PeterCaldwell: Ah, thanks for the heads up. Done.

AaronDonahue:
The only fail seems to be the known `spa_main_test` fail on `mappy`, which has now been fixed.  I think we can rebase this and retest and it should go through.  I'll take care of that.

AaronDonahue (approved):
Luca is on vacation, and it looks like Peter addressed his only comment, so I am going to approve.

PeterCaldwell:
I'm actually going to WIP this PR again because I added log statements for each time we enter a new proc and I haven't quite fully tested them yet. I'd also like to confirm a case where fields are going out of bounds actually gets caught... Thanks for remembering this though Aaron!

components/scream/src/share/atm_process/atmosphere_process_group.cpp
```diff
@@ -155,6 +155,9 @@ void AtmosphereProcessGroup::run_sequential (const Real dt) {
 
   for (auto atm_proc : m_atm_processes) {
     // Run the process
+    //if (m_comm.am_i_root()) {
+    printf("Entering %s\n",atm_proc->name() );
+    //}
```
PeterCaldwell, line 160: @jeff-cohere - this is my attempt to add a log message to SCREAM (it doesn't seem to work in CIME runs and I commented out the i_am_root bit bit as a test and didn't bother to un-comment it). Would using your ekat logging just involve a single line change here, or is there more work to be done since SCREAM doesn't init the logger?

jeff-cohere: EKAT has some custom logging tools, but you can also just use spdlog. The [README](https://github.com/gabime/spdlog) for that project has some good examples.

Tagging @bartgol in case he has thoughts.

PeterCaldwell:
Note that I decided that adding log messages should be its own PR and removed logging from this PR just now... I think it's ready to integrate.

AaronDonahue (approved):

AaronDonahue:
Okay cool.  I re-approved and will let the AT take over and merge

PeterCaldwell:
I still haven't hit a case where field checking actually caught a problem though. Could someone who actually knows how this checking work do a quick review to make sure the current implementation is sufficient? @AaronDonahue or @brhillman ???

AaronDonahue:
Sure, I can do that.  Shouldn't be hard, just set the T_mid value to something outside of bounds and it should get caught.

components/scream/src/share/atm_process/atmosphere_process_group.cpp
bartgol: Technically, `atm_proc->name()` returns a C++ string, while `printf` only takes native C types, and `%s` is for C-style string (aka, a `char*`). I'm not sure if this is the issue (the compile might just pass the first part of a string object, which might as well be the char* stored inside), but it's worth pointing out. So the correct call should be
```
printf ("Entering %s\n", atm_proc->name().c_str());
```

bartgol:
I'm a bit puzzled that you had to change `true` to `1` in the input file. That might point to some issues with the yaml parser. I will double check.

Tbc, if you have `blah: 1` in the input file, it gets interpreted and stored as an int, which means that later you must access it from the parameter list as an int. We are not doing that though (see, e.g., [here](https://github.com/E3SM-Project/scream/blob/63e5cd03785ad05fa72639188c1334e88b585a59/components/scream/src/share/atm_process/atmosphere_process.cpp#L32)). So the fact that setting the entries to `1` instead of `true` works, is somewhat puzzling/worrying...

AaronDonahue:
@bartgol , I've encountered this before with the parser.  I believe it interprets anything that looks like a string as a string.  So the word "true" or "false" is interpreted as a string.

Issue #1492, opened 2022-03-19
Integer overflow with variables holding sizes with ne30 (perhaps change some int to size_t)


ndkeen:
With ne30, there is at least one location where we need a long instead of int to represent memory size.

`components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp`

Printing value of mem during this function:
```
0: ndk RRTMGPRadiation::init_buffers mymem=   395814688 after 2d
0: ndk RRTMGPRadiation::init_buffers mymem=   594499664 after 3d arrays
0: ndk RRTMGPRadiation::init_buffers mymem= -2110987824 after 3d nswbands
0: ndk RRTMGPRadiation::init_buffers mymem= -1202713648 after 3d nlwbands
0: ndk RRTMGPRadiation::init_buffers mymem= -1191826800 after 3d surf alb
0: ndk RRTMGPRadiation::init_buffers mymem=   431868816 after 3d aero -- all
0: ndk RRTMGPRadiation::init_buffers used_mem=431868816
```

And then there is an assert that is OK because *both* locations are integer overflowing:

```
  int used_mem = (reinterpret_cast<Real*>(mem) - buffer_manager.get_memory())*sizeof(Real);
  EKAT_REQUIRE_MSG(used_mem==requested_buffer_size_in_bytes(), "Error! Used memory != requested memory for RRTMGPRadiation.");
```

This may or may not be impacting other things

PeterCaldwell:
I think @bartgol is on vacation all this week and @jgfouca is on vacation today (monday) so we might not get a quick answer, but this is the 2nd time we've seen integer overflow in assigning memory size and I'm wondering whether we need a campaign to find and upgrade *all* integer memory references to 64B...

ndkeen:
I can make a branch with these changes -- it's all pretty straight-forward (especially for authors), but I suspect there will be some opinions on how best to do it.  It may also be what's causing the failure with GPU as may need to be consistent with the gators.

PeterCaldwell:
That would be great. I'd actually suggest you not just make a branch, but make a WIP PR (i.e. start the title of the PR with "[WIP]: " which causes the autotester to not try to test or integrate it. That way we can easily poke through the code modifications you propose and comment on them...

ndkeen:
I made a branch here `ndk/scream/use-size_t-for-allocation-vars` that attempts to use `size_t` in a few places associated with storing a size.

ndkeen:
[...]

PR #1498, opened 2022-03-22
For integer types storing sizes in bytes, use size_t

ndkeen:
For ne30 cases, some integers were overflowing. Using size_t can help.

Fixes https://github.com/E3SM-Project/scream/issues/1492


PeterCaldwell:
This is great, thanks Noel! It looks like you've went through the whole v1 code and found all the places where ints were being used as buffer size, then changed them into sizeT types. This is exactly what I was hoping someone would do. I think when we first talked about this, someone (Luca) was saying we should only make this change in cases where we knew overflow would occur, but I think we should do this everywhere because who knows what resolution we will move to in the future and because the error messages we get when things go wrong are very hard to track down...

My one question is - how confident are you that you found all the cases of using int32 to set memory buffer size?

jgfouca (approved):
Looks good. Can we use the compiler to help us find narrowing conversions?

PeterCaldwell:
"Can we use the compiler to help us find narrowing conversions?" - can you expand on this, @jgfouca ? I'm not sure how I'd do this...

jgfouca:
@PeterCaldwell , I'm concerned that this type of problem will be common in the code:

```
int var_with_wrong_type = api_that_returns_size_t();
```

So, even though the API is set to use the correct type, the clients will continue to use the wrong type and the compiler will not catch this mistake. I just wrote a small test program and even `-Wall` does not catch this.

jgfouca:
Ok, it looks like `-Wconversion` does what we want, at least for clang:

```
clang++ -Wconversion testNarrowingConversion.cc
testNarrowingConversion.cc:12:13: warning: implicit conversion loses integer precision: 'size_t' (aka 'unsigned long') to 'int' [-Wshorten-64-to-32]
  int foo = my_api();
      ~~~   ^~~~~~~~
```

We probably have a crazy number of conversion warnings in our code, so we'd have to be clever in how we used this warning. Maybe modify the API header and recompile with the warning on.

ndkeen:
afaik, there isn't a compiler flag to detect overflow, but I'm testing the use of a gnu-specific pragma

PeterCaldwell:
@jgfouca - it's not really the narrowing conversion that's the problem, right? It's that the narrowing conversion doesn't bother to complain when it can't squeeze the sizeT integer into the int32 it's asked to. For example, the narrowing isn't a problem with ne4 because ne4 data volumes are small enough to fit into int32... 

Are you suggesting we look for narrowing conversions because they are easier to ask the compiler to to than to catch overflows like Noel is attempting?

PeterCaldwell:
Ok. I'm inclined to merge this PR and to keep issue #1492 open (or create a more pointed issue) while we try to figure out whether there are any other instances we missed here.

PeterCaldwell (approved):
just re-approving after fixing merge conflicts.

ndkeen:
With todays master after upstream merge (March 23), which does NOT have these particular `size_t` changes, I am able to run ne30 (at least with 4 nodes and 16 MPI's).  So that indicates we may not need these changes for ne30 and I'm now wondering why the branch I had been working in seemed to require this change. I still think these (or similar) changes are a step in the right direction.

PR #1501, opened 2022-03-22
Fix scorpio cmake settings for scream

jgfouca:
shr_pio_mod was not getting picked up.

### DIFF
--- a/components/scream/cmake/tpls/Scorpio.cmake
+++ b/components/scream/cmake/tpls/Scorpio.cmake
@@ -18,21 +18,23 @@ macro (CreateScorpioTargets)
         message (FATAL_ERROR "Something is off: pioc was not yet imported but piof was.")
       endif()
 
-      set (SCORPIO_LIB_DIR ${INSTALL_SHAREDPATH}/lib)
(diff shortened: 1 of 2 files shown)


jgfouca:
Adding @bartgol so he can see this when he gets back, not expecting a review.

PeterCaldwell:
So will this PR make SCREAM-CIME runs take whatever netcdf output type CIME tries to set for it? In other words, this PR will override the hardcoded value implemented in #1500?

AaronDonahue:
Actually, this PR is really part 2 of a 2 part fix.  In #1500 we make it so in standalone mode the default is 64 bit and in a CIME build it will take the value passed by CIME, stored in the shared modules.  But it turns out that the ifdef defined there wasn't correct.  So this PR fixes that, but that also requires some cmake changes.  So with 1500 alone we will always default to 64 bit.  With this PR we will get that control from CIME

Jim wanted a 2-part PR because this one should probably be reviewed by Luca, while #1500 can probably be reviewed by someone else and will allow Noel to move past the issue he is seeing on Perlmutter.

AaronDonahue (approved):
Approving for now to get this into master - as Jim said it would be good for @bartgol to take a look to see what we changed.

PR #1511, opened 2022-03-28
Add clone method to Field class

### LINKED ISSUE #1510: Field::deep_copy(const Field& src) doesn't work if the two fields have different allocations
@brhillman pointed this out. Turns out that the deep copy implementation assumes that the views from the two fields have the same dimensions. However, when using this feature to store "old" values, it is a bit cumbersome to have to do

```
  Field f_copy (f_src.get_header().get_identifier);
  // --- These lines are a bit obscure, and most importantly only needed to be able to use deep_copy
  const auto& ap_src = f_src.get_header().get_alloc_properties();
        auto& ap_copy = f_copy.get_header().get_alloc_properties();
  ap_copy.request_allocation(ap_src.get_largest_pack_size());
  //-----------------
  f_copy.allocate();
  ...
  f_copy.deep_copy(f_src);
```
It makes no sense to force `f_copy` to accommodate the same pack size as `f_src`, _just so that we can use deep_copy_. The code is also not very self-explanatory.

We could do a few things:
- make Field::deep_copy check whether the two fields have same padding, and, if not, manually copy (via parfor).
- add a "clone" method, that creates a Field identical to the input one, taking care of setting alloc props internally (so that the user doesn't have to see it)

I think the latter is probably more appealing, since it would allow to simply do
```
Field f_copy = f_src.clone();
```
and be done.

bartgol:
Makes it way easier to create temporary copies of a field (e.g., when checking if/how a field changes before/after a function call).

Fixes #1510

### DIFF
--- a/components/scream/src/share/tests/field_tests.cpp
+++ b/components/scream/src/share/tests/field_tests.cpp
@@ -199,6 +199,13 @@ TEST_CASE("field", "") {
     REQUIRE(f2.is_allocated());
     REQUIRE(fap2.get_alloc_size()==fap1.get_alloc_size());
     REQUIRE(views_are_equal(f1,f2));
+
+    // Changing f2 should leave f1 unchanged
+    f2.deep_copy<Real>(0.0);
+    REQUIRE (field_max<Real>(f2)==0.0);
+    REQUIRE (field_min<Real>(f2)==0.0);
+    REQUIRE (field_max<Real>(f1)==3.0);
+    REQUIRE (field_min<Real>(f1)==3.0);
   }
 
   SECTION ("deep_copy") {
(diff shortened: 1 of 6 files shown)


components/scream/src/share/tests/field_tests.cpp
```diff
@@ -186,6 +186,28 @@ TEST_CASE("field", "") {
...
+    f1.allocate_view();
+    f1.deep_copy(3.0);
+
+    Field f2 = f1.clone();
+    auto& fap2 = f2.get_header().get_alloc_properties();
+    REQUIRE(f2.is_allocated());
+    REQUIRE(fap2.get_alloc_size()==fap1.get_alloc_size());
+    REQUIRE(views_are_equal(f1,f2));
+
+    // Changing f2 should leave f1 unchanged
```
brhillman, line 203: Is it guaranteed that changing f1 similarly leaves f2 unchanged?

brhillman (approved):
Thanks for implementing this, this will be handy to have!

components/scream/src/share/tests/field_tests.cpp
bartgol: I think so.

PR #1513, opened 2022-03-29
Be consistent in referring to cori as cori-knl

jgfouca:
We need to use the CIME name for this machine, 'cori-knl'
in order for CIME cases to work.


ndkeen (approved):
ok. if needed, can also add cori-haswell

PR #1515, opened 2022-03-30
Convert wet mixing ratios to dry mixing ratios for SHOC (Fortran and C++)

singhbalwinder:
The wet mixing ratios from the scream's state vector (or prognostic state) are converted to dry mixing ratios for SHOC (except TKE). SHOC requires mixing ratios to be dry. The output from SHOC are dry mixing ratios, which are then converted back to the wet mixing ratios.

[NBFB]


singhbalwinder:
I think I have all the needed changes in the code. It will be great if you guys can quickly review it and let me know if there are ways to improve the code or add missing comments. In the meantime, I will merge master into the branch and resolve conflicts. Thanks!!

components/scream/src/physics/shoc/atmosphere_macrophysics.hpp
```diff
@@ -286,7 +318,33 @@ class SHOCMacrophysics : public scream::AtmosphereProcess
...
+        const int num_qtracer_packs = ekat::npack<Spack>(num_qtracers);
+        Kokkos::parallel_for(Kokkos::TeamThreadRange(team, num_qtracer_packs), [&] (const Int& q_iter) {
+            const auto is_not_tke_index = q_iter != tke_index;
+            if(is_not_tke_index){PF::calculate_wetmmr_from_drymmr(qtracers(i,q_iter,k),qv_dry);}
+          });
+
+
+        //convert qc from dry to wet mmr[kg/kg(wet-air)]
+        qc(i,k) = PF::calculate_wetmmr_from_drymmr(qc(i,k), qv_dry);
+        qv(i,k) = PF::calculate_wetmmr_from_drymmr(qw(i,k),qv_dry) - qc(i,k);
```
bartgol, line 347: `qv` and `qc` are slices of `qtracers`. As such, you are already converting them in the main loop.

bartgol (changes requested):
There are a few things to change, when it comes to the ||for's that do the conversion. There's one main challenge: you have 3 indices, basically completely independent of each other, so you could ||ize over ncols*ntrac*nlevs. HOWEVER, during the for loop you'd be also updating `qv`, so some iterations would use the original (dry) `qv`, while others would use the new (wet) one.

I see two solutions:
- store a copy of `qv` before dry->wet conversion, and always use that in the calls to PF::calc_wet_from_dry.
- run a 1st kernel to convert all tracers _except for_ `qv`. Then run a second loop to convert qv.

The latter could be done with team ||ism in a single main kernel. In pseudo code
```
// policy has league size ncols
pfor (policy,LAMBDA(...){
   icol = team.league_rank();
   qv = subview (qtracers, icol, qv_idx); // qv is the 1st, but can use qv_idx for clarity
   pfor(TTR(nlevs_packs*(num_tracers-1)),LAMBDA(int idx) {
      // +1 b/c we skip qv, which is the 1st. This is _assuming_ qv comes first,
      // which is true now, and likely will be, for quite some time
      int iq    = idx / nlevs_packs + 1;
      int ilev = idx % nlevs_packs 
      PF::convert_to_wet( qtracer(icol,iq,ilev), qv(icol,ilev).. )
   });
   
   // A second loop, over cols only, to convert qv
  pfor (TTR (nlevs_packs), LAMBDA(int k) {
    PF::convert_to_wet ( qv(k), .... );
  });
});
```

components/eam/src/physics/cam/shoc_intr.F90
```diff
@@ -348,46 +356,73 @@ subroutine shoc_init_e3sm(pbuf2d, dp1_in)
...
        lq(ixnumliq) = .false.
        edsclr_dim = edsclr_dim-1
-    endif 
+    endif
+
+    !SHOC needs all its water constituents in terms of "dry" mmr
+    !but the state vector has all its scalars in terms of "wet" mmr
+    !identify scalar which need conversion between wet and dry mmrs
+    do icnst = 1, pcnst
+       lq_dry_wet_cnvr(icnst) = .false. ! lets assume .false. by default
```
PeterCaldwell, line 404: didn't you already initialize the whole array to false? This seems redundant?

components/scream/src/physics/shoc/atmosphere_macrophysics.hpp
```diff
@@ -107,7 +107,33 @@ class SHOCMacrophysics : public scream::AtmosphereProcess
         // TODO: remove *_copy views once SHOC can request a subset of tracers.
         tke_copy(i,k) = tke(i,k);
 
-        qw(i,k) = qv(i,k) + qc(i,k);
+        /*---------------------------------------------------------------------------------
+         *Wet to dry mixing ratios:
+         *-------------------------
+         *Since tracers from the host model (or AD) are  wet mixing ratios and SHOC expects
+         *these tracers in dry mixing ratios, we convert the wet mixing ratios to dry mixing
+         *ratios for all the tracers except TKE (a comment about why leave TKE??).
```
PeterCaldwell, line 115: comment about TKE: "... except TKE, which is always defined in units of J/(kg of total air) and therefore should be considered to be "wet" when stored in the field manager and also "wet" when used in SHOC.

PeterCaldwell:
Thanks for your hard work on this, Balwinder! I made some comments, but you will see they're just about streamlining rather than about whether your code is working correctly...

oksanaguba:
I have trivial questions, if anyone has time: 

It seems that clubb uses wet ratios, not dry. What is the difference in shoc that needs dry tracers?

Does P3 still need wet ratios or can it also be switched?

bartgol:
> Does P3 still need wet ratios or can it also be switched?

I don't know anything about the other questions, but P3 is using dry ratios. @singhbalwinder put in the mods that do the wet->dry conversion before running p3, and the dry->wet after.

components/scream/src/physics/shoc/atmosphere_macrophysics.cpp
```diff
@@ -285,9 +287,27 @@ void SHOCMacrophysics::initialize_impl (const RunType /* run_type */)
   // For now, set z_int(i,nlevs) = z_surf = 0
   const Real z_surf = 0.0;
 
-  shoc_preprocess.set_variables(m_num_cols,m_num_levs,m_num_tracers,z_surf,m_cell_area,
+  // Find index of qv (water vapor, kg/kg-of-wet-air) and tke (units??) in the qtracer 3d view
+  // These indices are later used for converting tracers from wet mmr to dry mmr and vice-versa
+  auto tke_index = -1; // initialize index of tke in qtracers view
```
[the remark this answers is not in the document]
singhbalwinder: Thanks, Luca! This is a better way to find indices. I will replace my code with this.

components/scream/src/physics/shoc/atmosphere_macrophysics.hpp
```diff
@@ -286,7 +318,33 @@ class SHOCMacrophysics : public scream::AtmosphereProcess
...
+         */
+
+        //Units of all tracers below will become [kg/kg(wet-air)] for mass and [#/kg(wet-air)]
+        //for number (except TKE)
+
+        const auto qv_dry = qtracers(i,qv_index,k);// store qv "dry" mmr
+
+        //iterate over all tracers and convert dry mmr to wet mmr (except TKE)
+        const int num_qtracer_packs = ekat::npack<Spack>(num_qtracers);
+        Kokkos::parallel_for(Kokkos::TeamThreadRange(team, num_qtracer_packs), [&] (const Int& q_iter) {
```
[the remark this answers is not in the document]
singhbalwinder: Ok, I will move the code to convert wet<->dry before everything in its own TTR as you suggested.

components/eam/src/physics/cam/physics_utils.F90
```diff
@@ -9,7 +9,8 @@ module physics_utils
 
   implicit none
   private
-  public:: calculate_drymmr_from_wetmmr, calculate_wetmmr_from_drymmr
+  public:: calculate_drymmr_from_wetmmr, calculate_wetmmr_from_drymmr, &
+       calculate_drymmr_from_wetmmr_1gridcell, calculate_wetmmr_from_drymmr_1gridcell
```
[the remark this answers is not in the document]
singhbalwinder: I am calling `1gridcell` version inside the subroutine which computes it for the entire array. P3 uses the code for the entire array but SHOC uses the `1 gridcell` version. I can create an interface to make the calls look similar if that will look more consistent.

This code now looks more like C++ code where we have two versions, one for packs and other without packs.

components/eam/src/physics/cam/shoc_intr.F90
```diff
@@ -830,26 +867,45 @@ subroutine shoc_tend_e3sm( &
...
-   
-   !  Do the same for tracers 
+   enddo
+
+   !obtain wet mmr from the state vector
+   qv_wet = state%q(:,:,1)
+
+   !  Do the same for tracers
+   !Track which scalars need a conversion to wetmmr after SHOC main call
+   convert_back_to_wet(:) = .false.
```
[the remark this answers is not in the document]
singhbalwinder: I initially was using `lq_dry_wet_cnvr` (which is dimensioned as `pcnst`) but then I realized that shoc main loops over `edsclr_dim`, which is `pcnst - number of aerosols`. Therefore, I had to create a new logical array. Currently, `lq_dry_wet_cnvr` is used just in the  init code to make sure we have the all the right species to convert but `convert_back_to_wet` is a local array (at runtime) to track which species are converted to dry and would need to be converted back to wet.

components/scream/src/physics/shoc/atmosphere_macrophysics.hpp
singhbalwinder: Thanks!

singhbalwinder:
@oksanaguba : E3SM store water species (qv, qc etc.) in wet mmr but aerosols in dry. The parameterizations do the conversion internally (AFAIK). In scream, we assume all tracers are wet in AD (including aerosols, although they are not implemented yet), so these functions are provided to convert species from wet to dry and vice versa.

SHOC and P3 both needs dry tracers. @bogensch and @AaronDonahue might know  why these parameterizations need dry mmrs.

components/eam/src/physics/cam/physics_utils.F90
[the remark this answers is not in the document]
PeterCaldwell: Oh ok. No need to create an interface. I just didn't realize that p3 was still using the vector form.

components/eam/src/physics/cam/shoc_intr.F90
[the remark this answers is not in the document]
PeterCaldwell: Ok, that makes sense. Thanks!

PeterCaldwell:
Apologies for what's going to be an annoying post... if the only reason SHOC needs dry mixing ratios is for consistency with saturation mixing ratio computed within SHOC, it would be a lot easier to just call dry_to_wet_mmr on saturation mixing ratio rather than converting all input/output variables to/from dry mixing ratio. 

I started looking through https://agupubs.onlinelibrary.wiley.com/doi/10.1002/jame.20018 to see if I could figure out whether wet or dry is needed. My top-level view is that SHOC's view of water species is based on Reynold's averaged Navier-Stokes of an arbitrary scalar, so should hold equally well for mixing ratio or specific humidity. Note in particular that most equations in this paper are of the form dq/dt = K*q*...  converting units on both sides of the equation will have no effect. One caveat is that if we convert between mass per volume to mass per mass of air by using air density, then we need to be clear whether air density is measured in terms of total air mass or dry air mass. If rho_air = pres* Rd*T and if pres is doesn't include vapor loading and Rd is *dry* rather than total, this would be a reason to use dry variables in SHOC. I'm also a bit unsure of whether quantities like the Brunt Vaisala frequency require moist or dry variables, but I think this is easy to sort out.

PeterCaldwell:
@singhbalwinder - what's the status here? I think you have all the info you need to finish this task now?

components/eam/src/physics/cam/shoc_intr.F90
singhbalwinder: You are right, this is redundant. I can remove it but I kept it there for safety to guard against any change in this variable between its initialization to this code block.

AaronDonahue: @PeterCaldwell ,are you okay with this PR now?  It finally passed the AT so I think we can get it merged, but before I start that process I wanted to make sure you are happy with the response...

bartgol (approved):

components/eam/src/physics/cam/shoc_intr.F90
```diff
@@ -389,6 +397,33 @@ subroutine shoc_init_e3sm(pbuf2d, dp1_in)
        edsclr_dim = edsclr_dim-1
     endif 
 
+    !SHOC needs all its water constituents in terms of "dry" mmr
+    !but the state vector has all its scalars in terms of "wet" mmr
+    !identify scalar which need conversion between wet and dry mmrs
+    do icnst = 1, pcnst
+       lq_dry_wet_cnvr(icnst) = .false. ! lets assume .false. by default
+       do idw = 1, size(dry_wet_scalars)
+          if(trim(adjustl(stateq_names(icnst))) == trim(adjustl(dry_wet_scalars(idw))) )then
```
AaronDonahue, line 406: @singhbalwinder , where is this variable defined?  `stateq_names`.  I don't see it in any modules or being declared.

PR #1521, opened 2022-04-04
Second iteration of namelist defaults & Co.

bartgol:
This PR builds on top of #1462, and is a first attempt to "flatten" the nested structure of our input file.

So far the PR is only touching the input files, without the corresponding changes in the source code, to correctly parse the different nested structure.

The major change this PR does is the organization of the defaults xml file. Instead of grouping inputs by the output file, we store a few layers of metadata. In particular, there are a few sections in the defaults file that do _not_ enter the processed XML file, or enter it in a different location. For instance, the `generated_files` section lists what files we generate, and which sections of the XML will they contain. Another different layer is the one storing the atm processes (APs) defaults. Here's a snippet:

```
  <!-- Defaults for various atmosphere processes -->
  <atmosphere_processes>

    <!-- Basic options for each atm process -->
    <atm_proc_base>
      <Num__Subcycle__Iterations>1</Num__Subcycle__Iterations>
      <Enable__Precondition__Checks>true</Enable__Precondition__Checks>
      <Enable__Postcondition__Checks>true</Enable__Postcondition__Checks>
    </atm_proc_base>

    <!-- Basic options for each atm process group -->
    <atm_proc_group inherit="atm_proc_base">
      <Schedule__Type>sequential</Schedule__Type>
    </atm_proc_group>

    <!-- Basic options for each "physics" atm process -->
    <physics_proc_base inherit="atm_proc_base">
      <Grid>Physics GLL</Grid>
    </physics_proc_base>

    <!-- Homme dynamics -->
    <homme inherit="atm_proc_base">
      <Grid>Dynamics</Grid>
      <Vertical__Coordinate_Filename>${DIN_LOC_ROOT}/atm/scream/init/init_ne4np4.nc</Vertical__Coordinate_Filename>
      <Vertical__Coordinate_Filename COMPSET=".*SCREAM%AQUA.*">${DIN_LOC_ROOT}/atm/scream/init/scream-aquaplanet_init_ne4np4_L72_20211202.nc</Vertical__Coordinate_Filename>
      <Moisture>moist</Moisture>
    </homme>

    <!-- P3 microphysics -->
    <p3 inherit="physics_proc_base"/>
    ...
  </atmosphere_processes>
```
Some comments:
- The "inherit" keyword, allows an AP to include all entries of the "parent" AP, which helps to keep the defaults file short (e.g., you don't have to put a default for `Number__Of__Subcycles` in each AP defaults section). Of course, an AP can "override" the default of its parent.
- After processing the CIME case selectors, buildnml proceeds to build the section of the "Atmosphere Driver" input yaml file by combining these blocks, based on the content of `atm_procs_list` in the `Atmosphere Driver` section. This string can be used to group processes, by using parentheses. For instance `(homme,(shoc,p3))` is to be used to tell the AD to create a group G=(shoc,p3), and then set the whole atm proc group to (homme,G). The generated sublist for a group is called `group.$ap1_$ap2[  ]_$apN.`.
- I have not set a way to set defaults for specific "groups" of processes. E.g., consider the case `(homme,(shoc,p3)`, where one might want to subcycle `(shoc,p3)` three times. Right now, the only way to do that, is to generate the processed XML file, _and then_ do something like

```
./atm-config-chg  Atmosphere__Driver::group.homme_group.shoc_p3..::group.shoc_p3.::Number__Of__Subcycles=3
```
This is suboptimal on a few levels: 1) we would like to set this default in the namelist_default xml file, and 2) it requires knowledge on the nested structure of the XML file. I envision fixing 1) by allowing to add the following to namelist_default.xml:
```
<group name="shoc_p3">
  <Number__Of__Subcycles>3</Number__Of__Subcycles>
</group>
```
and have buldnml check for special defaults whenever parsing a new group of processes. For 2) I went this route since it made the intermediate XML more similar to the structure of the YAML file. In the CXX code, we create APs with a factory design, where we pass the parameters of the process to build only. So we kind of need to have a nested structure. Perhaps, we can create the nested structure only when turning the XML into a yaml file, though it might require some extra logic in the buildnml function.


bartgol:
@jgfouca I think I made some progress, and got something more manageable. I tested the `type` and `valid_values`  of atm-config-chg, and it works nicely:

```
$ ./atm-config-chg rrtmgp::Orbital__Year=s
ERROR: Could not use 's' as type 'integer'
$ ./atm-config-chg rrtmgp::Orbital__Year=10
$ ./atm-config-chg homme_ctl_nl::se_ftype=1
ERROR: Invalid value '1' for element 'se_ftype'. Value not in the valid list ('[0, 2]')
$ ./atm-config-chg homme_ctl_nl::se_ftype=2
$ 
```

This PR changes also how the input YAML file is organized, so I need to also modify our source code, so that it finds the expected data. I am tempted to also change our `Current Convention` for keys into `current_convention`, so that we can do away with the "__" trick in the XML keys. But if you think it's better to do that in a separate PR, I can punt on that.

bartgol:
This is how the intermediate file looks like (added .txt otherwise github woudl not upload). As you can see, there are still some things to work out, but the format is more amenable for atm-config-chg: notice that the atm procs are grouped as "(homme, ((shoc,cld,spa,p3),rrtmgp))", with in the old way of specifying atm procs would have led to a lot of nesting.

[namelist_scream.xml.txt](https://github.com/E3SM-Project/scream/files/8492988/namelist_scream.xml.txt)

jgfouca:
Looks good to me. We'll need updated documentation for how the file works and get approval from the other SCREAM members.

bartgol:
> Looks good to me. We'll need updated documentation for how the file works and get approval from the other SCREAM members.

Yes, and I haven't updated the testdoc framework, I simply commented it out for now.

bartgol:
I think this PR is somewhat ready. It is big, I have to admit, so multiple reviewers can help, and we can split the work. E.g., @jgfouca is probably the best candidate to review the python/xml stuff, while @AaronDonahue can focus on the src and test folder?

bartgol:
A few comments.

For the XML/buildnml/atm-config-chg stack

- Our defaults XML is now organized on a "concept" structure, rather than based on how information is split across input files (which should be hidden to the user).
- Atm procs stuff is nested in our input files, but the defaults need not to be nested. We can "build" the nesting format on the fly, based on the atm_procs_list entry.
- Atm procs groups can either have a "name", with a corresponding section in the defaults XML file (e.g., the "physics" group) or be declared on the fly, via "(a,b,...,z)" syntax. So, for instance, say you want a compset that changes SPA with MAM for aerosols. You have a few options to achieve this. 1) change the main atm_procs_list group to be "(homme, (shoc, cld, MAM, p3), rrtmgp)", without the need of adding a group to the defaults XML file; 2) declare a "my_mac_aero_mic" entry in the XML defaults, along the line of the existing "mac_aero_mic"  group, with a changed list of procs; 3) more simply, one could add selectors to the `atm_procs_list` entry of the existing "mac_aero_mic" group in the defaults XML file, to switch based on compset. Either way, one will have to add a "MAM" entry to the defaults XML file.
- Some entries can be "locked", so that once the intermediate XML file is generated, one cannot change them. The list of atm procs is the main example (and reason) for this feature: changing the list of procs would require to re-parse the defaults, and basically re-generate the intermediate XML file (which might even overwrite existing mods already done with atm-config-chg). For this reason, we "lock" this entry, and if one wants to change it, their only way is to declare a new compset, and add a selector in the defaults XML file that uses the desired new value for that compset.
- type and valid values metadata attributes are kept in the intermediate XML file. This allows atm-config-chg to perform checks on required new values, without the need to re-parse the defaults XML file.
- Some entries/nesting levels of the defaults XML file have been removed, to allow a tree that has as little nested levels as possible.
- Atm procs are not nested in the defaults XML file, but _are_ already nested in the intermediate XML file (depending on the atm procs list string). However, one does not need to know _where_ a particular atm proc is nested in order to change its parameters. E.g., `atm-config-chg shoc::Num__Subcycle__Iterations=3` works regardless of _where_ shoc is located in the atm proc group tree.
- Added the "Debug" section to the defaults XML file.

In the src and test folder stuff.
- input yaml files have been modified to reflect the structure that buildnml creates.
- added some logging to the AD initialization (helped me debug)
- Fix #1563 .

bartgol:
To give an idea, this is the structure of the _intermediate_ XML file I get for the aquaplanet case (I fixed indentation, which, unfortunately, the xml python module is not great at doing right when dumping to file).
[namelist_scream.xml.txt](https://github.com/E3SM-Project/scream/files/8544831/namelist_scream.xml.txt)

components/scream/cime_config/buildnml
```diff
@@ -199,281 +194,264 @@ def evaluate_selectors(element, case, ez_selectors):
...
-            expect("format" in child.attrib, "file element missing required 'format' attribute")
-
-            metadata[child.attrib["name"]] = evaluate_selectors(child, case, selectors)
-
-    if selectors_elem is not None:
-        root.remove(selectors_elem)
-
-    return ET.tostring(root, encoding="unicode"), metadata
+    #  """
+    #  >>> case = MockCase({'ATM_GRID':'ne4ne4', 'SCREAM_CMAKE_OPTIONS':'FOO=ON SCREAM_NUM_VERTICAL_LEV 128 BAR=OFF'})
```
jgfouca, line 205: All these good tests are commented out. I assume you didn't want to worry about them as you refactored things, but they should be re-enabled and fixed-up now that you are done.

Also, the new "inherit" and "locked" mechanisms will need tests.

components/scream/cime_config/buildnml
```diff
@@ -487,257 +465,170 @@ def create_raw_xml_file(case, caseroot):
...
-    interleave = 'hi'
-    /
-    &group2
-    testval = .false.
-    testval2(:) = 1, 2, 3, 4
-    interleave(:) = hi, there
-    /
-    <BLANKLINE>
-    """
+    #  """
```
jgfouca, line 502: More tests need to be re-enabled.

jgfouca (changes requested):
I approve the python changes under the condition that doctest unit tests are restored and new tests to cover the new features are added.

bartgol:
@jgfouca Adding back tests allowed me to spot errors, so thank you for pushing me to do it.

I added a bunch of tests, but I find it complicated to add tests for a few of the top-level functions, such as `create_input_data_list_file`, which needs to open a file from the `$caseroot/Buildconf/` folder. Can you take a look and see if the amount of testing is enough, and, if not, give me some directions of where/how to test more?

jgfouca:
@bartgol , you only need to test the lower level functions that have complex implementations. Don't feel obligated to test anything that wasn't already being tested in the original version.

jgfouca:
@bartgol , I'm still a bit concerned that a lot of testing was removed from `_create_raw_xml_file_impl` and `_dump_to_nml_impl` and does not appear to have been recreated/moved.

bartgol:
> @bartgol , I'm still a bit concerned that a lot of testing was removed from `_create_raw_xml_file_impl` and `_dump_to_nml_impl` and does not appear to have been recreated/moved.

Some of it has been added to eamxx_buildnml_impl. I had some issues with the dump to nml, since I did not want to create files to compare against. But I now notice `_dump_to_nml_impl` simply produces a string, which I can easily test. I also pruned the test that checked

``` 
CIME.utils.CIMEError: ERROR: No metadata for child 'alien_nation'
```
Since the metadata is no longer extracted from the xml, but left in there until it's time to dump to file. I guess I could verify that the intermediate file has the metadata.

bartgol:
I'm gonna bring back some tests in the morning. Stay tuned!

components/scream/cime_config/namelist_defaults_scream.xml
```diff
@@ -87,136 +100,158 @@ tweaking their $case/namelist_scream.xml file.
...
+    <!-- CLD fraction -->
+    <cldFraction inherit="physics_proc_base"/>
+
+    <!-- Simple Prescribed Aerosols (SPA) -->
+    <spa inherit="physics_proc_base">
+      <SPA__Remap__File>none</SPA__Remap__File>
+      <SPA__Data__File>${DIN_LOC_ROOT}/atm/scream/init/spa_file_unified_and_complete_ne4_scream.nc</SPA__Data__File>
+    </spa>
+
+    <!-- Simple Prescribed Aerosols (SPA) -->
```
AaronDonahue, line 168: copy/paste error - should be "radiation"

components/scream/cime_config/namelist_defaults_scream.xml
```diff
@@ -87,136 +100,158 @@ tweaking their $case/namelist_scream.xml file.
...
+
+  <!-- List of nc files for loading inputs on specified grids -->
+  <Initial__Conditions>
+    <Filename>${DIN_LOC_ROOT}/atm/scream/init/init_ne4np4.nc</Filename>
+    <Filename COMPSET=".*SCREAM%AQUA.*">${DIN_LOC_ROOT}/atm/scream/init/scream-aquaplanet_init_ne4np4_L72_20211202.nc</Filename>
+  </Initial__Conditions>
+
+  <!-- List of yaml files containing I/O output specs -->
+  <Scorpio>
+    <Output__YAML__Files>data/scream_output.yaml, </Output__YAML__Files>
```
AaronDonahue, line 204: will the trailing `,` in the list cause an error?

components/scream/cime_config/namelist_defaults_scream.xml
```diff
@@ -87,136 +100,158 @@ tweaking their $case/namelist_scream.xml file.
...
+  </Scorpio>
+
+  <!-- Debug options for scream -->
+  <Debug>
+    <Atmosphere__DAG__Verbosity__Level>0</Atmosphere__DAG__Verbosity__Level>
+    <Atm__Log__Level valid_values="trace,debug,info,warn,error">info</Atm__Log__Level>
```
AaronDonahue, line 210: I like the `valid_values` control.

components/scream/cime_config/namelist_defaults_scream.xml
```diff
@@ -87,136 +100,158 @@ tweaking their $case/namelist_scream.xml file.
...
+
+  <!-- Grids manager specs -->
+  <Grids__Manager>
+    <Type>Dynamics Driven</Type>
+    <Reference__Grid>Physics GLL</Reference__Grid>
+    <Dynamics__Namelist__File__Name>./data/namelist.nl</Dynamics__Namelist__File__Name>
+  </Grids__Manager>
+
+  <!-- List of nc files for loading inputs on specified grids -->
+  <Initial__Conditions>
```
AaronDonahue, line 197: Have we removed all the reference to RESTARTS?  If so, will we run into an issue with no "metadata" being found for the restart section if we try to manually add it?

components/scream/cime_config/namelist_defaults_scream.xml
```diff
@@ -87,136 +100,158 @@ tweaking their $case/namelist_scream.xml file.
...
+    <Filename>${DIN_LOC_ROOT}/atm/scream/init/init_ne4np4.nc</Filename>
+    <Filename COMPSET=".*SCREAM%AQUA.*">${DIN_LOC_ROOT}/atm/scream/init/scream-aquaplanet_init_ne4np4_L72_20211202.nc</Filename>
+  </Initial__Conditions>
+
+  <!-- List of yaml files containing I/O output specs -->
+  <Scorpio>
```
AaronDonahue, line 203: Have we removed all the reference to RESTARTS?  If so, will we run into an issue with no "metadata" being found for the restart section if we try to manually add it?

components/scream/src/mct_coupling/scream_cxx_f90_interface.cpp
```diff
@@ -111,7 +106,7 @@ void scream_create_atm_instance (const MPI_Fint f_comm, const int atm_id,
     auto& ad = c.create<AtmosphereDriver>();
 
     ad.set_comm(atm_comm);
-    ad.set_params(ad_params);
+    ad.set_params(scream_params);
```
AaronDonahue, line 109: should we call this `eamxx_params`?

components/scream/cime_config/namelist_defaults_scream.xml
bartgol: Interesting. The generated `scream_input.yaml` is correct. I actually suspect that without ',' it might cause issues, since it would no longer look like an array, so it would be parsed as a string, causing issues when we access it from the parameter list asking for a vector of strings...

bartgol: This should be probably fixed though.

components/scream/tests/coupled/dynamics_physics/model_restart/input_baseline.yaml
```diff
@@ -12,41 +12,36 @@ Time Stepping:
...
-    Filename: ${SCREAM_DATA_DIR}/init/homme_physics_ne2np4.nc
     aero_g_sw: 0.0
     aero_ssa_sw: 0.0
     aero_tau_sw: 0.0
     aero_tau_lw: 0.0
 
-Atmosphere Processes:
-  Number of Entries: 5
+atmosphere_processes:
+  atm_procs_list: (homme,shoc,CldFraction,p3,rrtmgp)
```
AaronDonahue, line 24: Nice, does this approach mean that if we wanted to rearrange the order we could just change this entry and nothing else?

AaronDonahue (commented):
I had a few comments, nothing major, but I'll wait to approve so that the AT doesn't take over.

components/scream/cime_config/namelist_defaults_scream.xml
bartgol: Ok, 9fc3e489b4 should have addressed this.

components/scream/cime_config/namelist_defaults_scream.xml
bartgol: @jgfouca idea. It is already in master. I just added a few usages here and there.

components/scream/cime_config/namelist_defaults_scream.xml
bartgol: Ah, good catch!

bartgol:
@jgfouca I ran into an issue that I do not know how to handle. Consider the xml entries

```
    <Model__Restart>
      <Casename>${CASE}</Casename>
      <Output__Control>
        <Frequency__Units>INVALID</Frequency__Units>
        <Frequency__Units REST_OPTION="nstep*">Steps</Frequency__Units>
      </Output__Control>
    </Model__Restart>
```
If the user does `./xmlchange REST_OPTION=ndays` _after_ `namelist_scream.xml` has been generated, we have no way of getting that info in our XML file.

So the general issue is: what do we do if the user does some XML changes that should impact the atm XML logic? Can we inject info on the cime's XML files timestamps in our `namelist_scream.xml` file, and re-generate if they have changed (warning the user that all pre-exsisging atm-config-chg changes are lost)?

components/scream/tests/coupled/dynamics_physics/model_restart/input_baseline.yaml
bartgol: Yes.

bartgol: Well, actually, we do _not_ allow changing this entry via `atm-config-chg`, since changing this would require a full re-generation of it (b/c we might need to go grab other defaults that are not currently in the XML).

The idea behind atm-config-chg is that once namelist_scream.xml is generated, you no longer need to look at namelist_defaults_scream.xml. So if you need sections from the defaults that are not in the current XML, it is an invalid use.

If you look at the defaults XML, you will notice that this entry has the attribute `locked="true"`, which makes atm-config-chg throw an error if you try to change it.

Atm proc order needs to be defined via compset. At least for now.

components/scream/src/mct_coupling/scream_cxx_f90_interface.cpp
bartgol: I guess. There's so much scream->eamxx to do... We can of course change this, but we can also wait until we change _all_ occurrences.

components/scream/cime_config/buildnml
```diff
@@ -120,9 +111,115 @@ def evaluate_selectors(element, case, ez_selectors):
...
+    ... </namelist_defaults>
+    ... '''
+    >>> bad3 = ET.fromstring(xml_bad3)
+    >>> evaluate_selectors(bad3,case,selectors_good)
+    Traceback (most recent call last):
+    evaluate_selectors.<locals>.BadDefaultsOrdering: ERROR: child 'var1' element without selectors occured after other parameter elements for this parameter
     """
-    metadata = {} # elem_name -> (type, valid_values)
+
+    class BadSelectorName (Exception):
```
jgfouca, line 214: Do we really need custom exception types? This is only useful if we are catching and handling multiple types of exceptions, e.g.

```
try:
    * some action *
except CustomExc1:
    * handle exc1 error condition *
except CustomExc2:
    * handle exc2 error condition *
```

We aren't doing that anywhere as far as I've looked, so the custom exceptions are not necessary. Checking the exception string in doctest is adequate for ensuring the expected error was hit.

Another problem is that these new exceptions are not SystemExit exception and so will give the user an ugly stacktrace that is likely useless to them. They only need to see the msg describing the XML mistake they made.

components/scream/cime_config/buildnml
```diff
@@ -176,304 +278,106 @@ def evaluate_selectors(element, case, ez_selectors):
...
-    ...   </selectors>
-    ...   <file name="test.yaml" format="yaml">
-    ...     <testval valid_values="1,2,3">1</testval>
-    ...     <testval hgrid="ne4ne4" valid_values="1,2,3,4">2</testval>
-    ...     <testval hgrid="no_match" valid_values="1,2,3,4,5">3</testval>
-    ...     <testval2 valid_values="1,2,3">1</testval2>
-    ...     <testval2 hgrid="ne4ne4">2</testval2>
-    ...     <testval2 hgrid="no_match">3</testval2>
-    ...   </file>
+    ...     <selectors/>
```
jgfouca, line 317: This is a good new test, but I still don't understand why some of the tests above were deleted instead of adapted to work with this new iteration? I can see some of the selector tests were moved but it looks like some were just lost.

components/scream/cime_config/buildnml
```diff
@@ -176,304 +278,106 @@ def evaluate_selectors(element, case, ez_selectors):
...
+    ...     </atmosphere_processes_defaults>
     ... </namelist_defaults>
     ... '''
-    >>> _create_raw_xml_file_impl(case, test_metadata_override)[1]
-    {'test.yaml': {'testval': ['integer', '1,2,3,4'], 'testval2': ['integer', '1,2,3']}}
+    >>> import xml.etree.ElementTree as ET
+    >>> defaults = ET.fromstring(xml)
+    >>> generated = _create_raw_xml_file_impl(case,defaults)
+    >>> aps = get_child(generated,"atmosphere_processes")
+    >>> for c in aps:
```
jgfouca, line 341: Instead of probing objects in the doctests, why not just dump the raw xml and compare all of it at once?

components/scream/cime_config/buildnml
```diff
@@ -487,272 +391,118 @@ def create_raw_xml_file(case, caseroot):
...
+    ...     'real': 2.0,
+    ...     'string': 's'
+    ... }
+    >>> print(_dump_to_nml_impl(good2))
+    Traceback (most recent call last):
+    _dump_to_nml_impl.<locals>.MixedNestingLevels: ERROR: Error! _dump_to_nml_impl cannot mix nested and non-nested dicts.
+
     """
+
+    class MixedNestingLevels (Exception):
```
jgfouca, line 477: Again, no need for a custom exception type.

jgfouca (changes requested):
Minor stuff

components/scream/cime_config/buildnml
bartgol: What do you mean with "dump"? I don't want doctest to create a file. I guess ordered_dump can dump to a dictionary?

jgfouca (approved):

AaronDonahue (approved):
Thank you for addressing all my comments.

bartgol:
Blake and weaver timed out, but they passed in the previous round of testing (and no commit was pushed after that round). Merging.

PR #1525, opened 2022-04-05
Fix subcycling logic for AtmosphereProcess

bartgol:
This PR fixes the last few issues with subcycling an atmosphere process. I manually verified it works, by hacking the input.yaml file for the full homme+physics test, subcycling SHOC+Cld+SPA+P3 3 times, and printing out the dt inside Homme, P3, and SHOC. I got:

```
Run HOMME with dt = 1800
Run SHOC with dt = 600
Run P3 with dt = 600
Run SHOC with dt = 600
Run P3 with dt = 600
Run SHOC with dt = 600
Run P3 with dt = 600
```

I haven't added a specific test, since I wanted to ask first: should I add a dummy test (subcycling some dummy procs) or should we modify one of our coupled tests, so that 1+ (but not all) atm procs are subcycled?

@PeterCaldwell @AaronDonahue thoughts?

### DIFF
--- a/components/scream/src/control/atmosphere_driver.cpp
+++ b/components/scream/src/control/atmosphere_driver.cpp
@@ -752,7 +752,9 @@ void AtmosphereDriver::initialize_atm_procs ()
 {
   // Initialize memory buffer for all atm processes
   m_memory_buffer = std::make_shared<ATMBufferManager>();
-  m_atm_process_group->initialize_atm_memory_buffer(*m_memory_buffer);
+  m_memory_buffer->request_bytes(m_atm_process_group->requested_buffer_size_in_bytes());
+  m_memory_buffer->allocate();
+  m_atm_process_group->init_buffers(*m_memory_buffer);
 
   const bool restarted_run = m_atm_params.sublist("Initial Conditions").get<bool>("Restart Run",false);
 
--- a/components/scream/src/physics/spa/atmosphere_prescribed_aerosol.cpp
+++ b/components/scream/src/physics/spa/atmosphere_prescribed_aerosol.cpp
@@ -132,7 +132,7 @@ void SPA::init_buffers(const ATMBufferManager &buffer_manager)
   m_buffer.spa_temp.PS = decltype(m_buffer.spa_temp.PS)(r_mem,ncols);
   r_mem += m_buffer.spa_temp.PS.size();
 
-  int used_mem = (r_mem - buffer_manager.get_memory())*sizeof(Real);
+  size_t used_mem = (r_mem - buffer_manager.get_memory())*sizeof(Real);
   EKAT_REQUIRE_MSG(used_mem==requested_buffer_size_in_bytes(),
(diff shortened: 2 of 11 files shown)


tcclevenger (commented):
This looks good to me! What is the reason for changing the mem buffer?

PeterCaldwell:
I think a dummy test would be good. For example, could you run p3+shoc standalone 3x, then run p3+shoc standalone with 3 subcycles and confirm that you get the same answer? Or were you thinking of truly dummy procs that just return 1 or 2 or something? I think we will naturally have some of the coupled tests use subcycling because it will allow them to run with a longer timestep and therefore run faster. Like I think ne4 should be using a 2 hr timestep instead of a 5 min timestep, with shoc+p3 substepped down to 5 min and shoc further substepped to 2.5 min(?).

bartgol:
> I think a dummy test would be good. For example, could you run p3+shoc standalone 3x, then run p3+shoc standalone with 3 subcycles and confirm that you get the same answer? Or were you thinking of truly dummy procs that just return 1 or 2 or something? I think we will naturally have some of the coupled tests use subcycling because it will allow them to run with a longer timestep and therefore run faster. Like I think ne4 should be using a 2 hr timestep instead of a 5 min timestep, with shoc+p3 substepped down to 5 min and shoc further substepped to 2.5 min(?).

I was thinking of a really dummy test, since this is a basic infrastructure test. When SCREAM->EAMxx, I want the feature to be tested even without p3 or shoc. But I guess for now we can just use something SCREAM-specific. I think what you suggest (p3+SHOC monolithic-vs-subcycled) is relatively easy to set up. I'll do that.

bartgol:
@PeterCaldwell I added a small shoc+p3 test, that runs 3 steps either "monolithically" (3 atm steps) or "subcycling-style" (1 atm step, subcycled 3 times), and then compares nc output files.

bartgol:
Blake is currently experiencing some issue, with commands taking a long time to complete on the login node, causing timeout during the git clone phase of our jenkins jobs. I manually tested this PR on a compute node, and all tests pass. Mappy and weaver are already passing.

@tcclevenger If/when you are ok with this PR, we can manually merge it.

bartgol:
Blake seems more responsive today. Let's give the AT another chance, so we get a nice shiny checkmark near the PR.

AaronDonahue (commented):
Looks good, I just have one question about the setup of a sub-cycled test.  Do we need to add the process type `Group` to `register_physics` in order to tightly couple a subset of processes together?  I recall I had to do that when I was experimenting with this.

bartgol:
Today is weaver's turn to be slow, with several long jobs running, causing AT to timeout.

@tcclevenger can you please re-review? There was a test added after your review. If you are ok with how things are, I'll manually merge.

tcclevenger (approved):

tcclevenger:
@bartgol I approve if you want to manually merge

PR #1526, opened 2022-04-05
Set non-exported values to 0.0 in SurfaceCoupling::do_export()

tcclevenger:
This solves an issue on GPU CIME runs. For the exports that are skipped during the initial `do_export()` call, we were doing nothing. But it just so happens that  the `cpl_exports_ptr`on GPU contains NaNs instead of zeros. This was causing fails in other models using the exports. The `"do_initial_export"` test in `sc_tests` has been modified to catch that fields are set to 0.

### DIFF
--- a/components/scream/src/control/surface_coupling_export.cpp
+++ b/components/scream/src/control/surface_coupling_export.cpp
@@ -102,6 +102,8 @@ void SurfaceCoupling::do_export (const bool init_phase)
     bool do_export = (!init_phase || info.do_initial_export);
     if (do_export) {
       cpl_exports_view_d(icol,info.cpl_idx) = info.data[offset];
+    } else {
+      cpl_exports_view_d(icol,info.cpl_idx) = 0.0;
     }
   });
 
--- a/components/scream/src/control/tests/sc_tests.cpp
+++ b/components/scream/src/control/tests/sc_tests.cpp
@@ -685,12 +685,12 @@ TEST_CASE ("do_initial_export")
     // Perform export with init_phase=true
     exporter.do_export(true);
 
-    // Check that only f1 was exported
+    // Check f1 was exported, but f2 was set to 0
     f1_exp.sync_to_host();
     f2_exp.sync_to_host();
     for (int icol=0; icol<ncols; ++icol) {
       REQUIRE (raw_data[0 + icol*num_fields] == f1_exp_h(icol));
-      REQUIRE (raw_data[1 + icol*num_fields] == -1);
+      REQUIRE (raw_data[1 + icol*num_fields] == 0.0);
     }
 
     // Set all raw_data back to -1


PeterCaldwell:
Awesome catch, Conrad!

tcclevenger:
> @tcclevenger why does the view contain NaN's? We never init it to anything, so we should be using Kokkos default behavior, which, IIRC, is to init to 0. We do have a policy of default-init to NaN for Pack, but the import/export views in the SC class are not packed.

The issue seems to be with what was in `a2x%rAttr` data ptr given externally. When we set up the export view in SC we just construct with the pointer, which should be preserving the values there.

bartgol:
> > @tcclevenger why does the view contain NaN's? We never init it to anything, so we should be using Kokkos default behavior, which, IIRC, is to init to 0. We do have a policy of default-init to NaN for Pack, but the import/export views in the SC class are not packed.
> 
> @bartgol The issue seems to be with what was in `a2x%rAttr` data ptr given externally. When we set up the export view in SC we just construct with the pointer, which should be preserving the values there.

Ah, right, we do create the device view as a device-mirror of the host one. Then it makes sense.

However, if there are models using these exports...is it correct that scream skips them? Is setting them to 0 safe? Maybe @PeterCaldwell could chime in, if we knew which field in particular this/these was/were.

tcclevenger:
@bartgol @PeterCaldwell The issue is actually with fields that are not exported during initialization. The field that caught the error was `Faxa_rain`.

bartgol:
> @bartgol @PeterCaldwell The issue is actually with fields that are not exported during initialization (but are exported after each time step). The field that caught the error was `Faxa_rain`.

Still, if some other component caught the NaN, it might be that we should compute a good value even during initialization?

tcclevenger:
> Still, if some other component caught the NaN, it might be that we should compute a good value even during initialization?

Yes. Also are we certain that the fields that scream does not use should also be set to 0? I can output the values in that data ptr to see if each field is nan, or if some have actual values.

bartgol:
> > Still, if some other component caught the NaN, it might be that we should compute a good value even during initialization?
> 
> Yes. Also are we certain that the fields that scream does not use should also be set to 0? I can output the values in that data ptr to see if each field is nan, or if some have actual values.

That's a good point. If they are used downstream, setting them to 0 might be dangerous. Imho, if you don't compute something, NaN is the valimost logical value. If the field ends up being used, then we need to provide valid values (perhaps 0 _is_ valid, but we need to be sure).

PeterCaldwell:
In this page I compare cpl.hi files from v0 and v1. This should be an empirical evaluation of what values these variables will take: https://acme-climate.atlassian.net/wiki/spaces/NGDNA/pages/3282501670/SCREAMv1+ne4+diagnostics . The interesting thing about these plots is that they *don't* include any variable whose biggest difference between v0 and v1 is <0.01% of the v1 mean value or <1e-12 (whichever is bigger). I think this means that v0 is zeros where v1 is zeros.

Not particularly useful, but just because I spent time digging it up, here's the confluence page where we decided what variables we needed to export: https://acme-climate.atlassian.net/wiki/spaces/NGDNA/pages/1927512393/Field+Manager+Variable+Names

PeterCaldwell:
I am also creeped out that the surface models are actually using the export variables we don't think we need, though some of the variables - like Faxa_rainc - are zero because SCREAM doesn't have convective rain. Other components probably take "rain = rainc+rainl", so setting rainc to zero is the right thing to do. Which reminds me: @tcclevenger - I think there is no Faxa_rain. Do you mean Faxa_rainc? I wouldn't be surprised if Faxa_rainl is zero for the initial step of a run that's just starting, but it's not ok for Faxa_rainl to be zero on a restart run...

bartgol:
@PeterCaldwell thanks for triple checking this. If it makes sense to set to 0, that's great. I just wanted to be sure we don't set to 0 without making sure 0 is the _right_ value. Given your comments, I agree that we should be fine with setting to 0 (assuming the "rain" vs "rainc" vs "rainl" doubt gets cleared).

tcclevenger:
@PeterCaldwell actually is is `Faxa_rainl`. I don't think SC knows about `Faxa_rainc`.

PeterCaldwell:
Yeah, Faxa_rainc is one of those variables we want to automatically get set to zero instead of nan! Like I said before, I'm not freaked out about rainl being zero in the first timestep. I assume that the init routine that is setting these things to zero is only called for a new initialization rather than a restart run, so I think setting to zero is the right thing to do.

PeterCaldwell:
More generally, I think it's fine to set a bunch of stuff to zero for the initial step because that's kind of a garbage step where everything is getting spun up. For example, I think all the clouds need to condense in that step. We just want the assumptions we make there to not be so bad that the model crashes.

bartgol:
> Yeah, Faxa_rainc is one of those variables we want to automatically get set to zero instead of nan! Like I said before, I'm not freaked out about rainl being zero in the first timestep. I assume that the init routine that is setting these things to zero is only called for a new initialization rather than a restart run, so I think setting to zero is the right thing to do.

Good point! @tcclevenger can we confirm that setting fields to 0 only happens for initial runs, and not for restarted runs? 

Actually, this is a broader issue: we do not have any v1 restart test. We do have the `model_restart` test, but that's only for standalone scream. There's definitely some work to do in our mct interface for restarted runs, and I don't think we're doing it. I'll open an issue.

bartgol (approved):

tcclevenger:
@PeterCaldwell I was able to run indefinitely (I think I went 1/2 month? ~5K atmosphere steps).

bartgol:
> @bartgol No, it is not done during restart. It's only done in `atm_init_mct()` in `atm_comp_mct.F90`.

`atm_init_mct` is run for all types of simulation. Internally that subroutine needs to handle restarts. See [here](https://github.com/E3SM-Project/scream/blob/d78716ee3df34d94e23d4b853d60ffe0adb66621/components/eam/src/cpl/atm_comp_mct.F90#L274) for what EAM does (while we are currently not).

bartgol:
@PeterCaldwell I'm 99.9% confident that v1 does not, in fact do restarts. I have to work on adding the logic, based on CIME case parameters.

tcclevenger:
@PeterCaldwell Nope, weaver still fails with the "divide by zero" error in SnowSnicarMod.F90`.

PR #1527, opened 2022-04-06
Fix units on surf_latent_flux

PeterCaldwell:
Several months ago we figured out that latent heat flux was expected to be in flux units (kg/m2/s) but its units label was energy units (W/m2). This 2-line PR just changes the units to the correct value.

The only thing I want to triple check is that SHOC is actually expecting LHF in flux units.


bogensch (approved):
tripled checked.

PR #1528, opened 2022-04-06
Generate p3 tables at build time

### LINKED ISSUE #1505: Read error (array_io_read failed) with ne30 v1scream case
Haven't had a chance to look at this closer.

New error with ne30, CPU-only on PM.  This is using 4 nodes, 16 MPI's.  I see the error with OPT build, but not with DEBUG build. 
```
15: array_io_read failed with: /pscratch/sd/n/ndk/wacmy/s17-mar23/externals/ekat/src/ekat/util/ekat_file_utils.hpp:24: FAIL:
15: nread == sz
15: read: nread = 1023 sz = 3000
15: WARNING: SPA Remap File has been set to 'NONE', assuming that SPA data and simulation are on the same grid - skipping horizontal interpolation p3_iso_c::p3_init: One or more table files exists but gave a read error.
 0: WARNING: SPA Remap File has been set to 'NONE', assuming that SPA data and simulation are on the same grid - skipping horizontal interpolation p3_iso_c::p3_init: One or more table files exists but gave a read error.
 0: array_io_read failed with: /pscratch/sd/n/ndk/wacmy/s17-mar23/externals/ekat/src/ekat/util/ekat_file_utils.hpp:24: FAIL:
 0: nread == sz
 0: read: nread = 1023 sz = 3000
15: terminate called after throwing an instance of 'std::logic_error'
15:   what():  /pscratch/sd/n/ndk/wacmy/s17-mar23/components/scream/src/physics/p3/p3_f90.cpp:122: FAIL:
15: info == 0
15: p3_init_c returned info -1
```

```
template<typename T>
void read (T* v, size_t sz, const FILEPtr& fid) {
  size_t nread = fread(v, sizeof(T), sz, fid.get());
  EKAT_REQUIRE_MSG(nread == sz, "read: nread = " << nread << " sz = " << sz);
}
```

could we maybe add the filename that is being read when this error happens?

bartgol:
This PR sparked from discussion in #1505. There, we discussed a few approaches, but since it was quick to do, I just went ahead and did it.

This PR makes P3 compute the mu_r, revap, vn, and vm tables at build time. Before, they were computed if the file was not found, but due to concurrent checks from multiple ranks in CIME runs, it was not very safe.

Specifically, this PR creates a cmake target (`p3_create_tables`), built as part of `all`, which runs _serially_ the `p3_init` function, ensuring tables are written.

One caveat: if you run `make` from a folder that is not a parent of `src/physics/p3`, the `p3_create_tables` target is not built. I tried to think of a clear solution, but the only way seems to be to add a target dependency, which could be cumbersome.

Closes #1505.

### DIFF
--- a/components/scream/tests/coupled/dynamics_physics/homme_shoc_cld_p3_rrtmgp/CMakeLists.txt
+++ b/components/scream/tests/coupled/dynamics_physics/homme_shoc_cld_p3_rrtmgp/CMakeLists.txt
@@ -69,6 +69,7 @@ configure_file(${HOMME_SOURCE_DIR}/test/vcoord/acme-72m.ascii
                ${CMAKE_CURRENT_BINARY_DIR}/vcoord COPYONLY)
 
 # RRTMGP data files
+file (MAKE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/data)
 GetInputFile(init/rrtmgp-data-sw-g224-2018-12-04.nc
   ${CMAKE_CURRENT_BINARY_DIR}/data COPYONLY)
 GetInputFile(init/rrtmgp-data-lw-g256-2018-12-04.nc


PeterCaldwell:
Why not just stuff these files in the inputdata server? And why would we ever `make` from a directory that isn't a parent?

bartgol:
> Why not just stuff these files in the inputdata server? And why would we ever `make` from a directory that isn't a parent?

We can put these in the inputdata server. Since the tables are generated based on code in the repo, it seemed more self-consistent to have the build system generate the tables. One tiny advantage of this approach: if anything _at all_ changes in p3 that can cause changes in the tables, we would need to upload new files on the server, and the name change (input files need to be versioned) would require to change all places that use them. This way, we do not need to update the server files if p3 implementation changes. Granted, maybe we will rarely change p3 implementation, though I saw this line below in micro_p3.F90.
```
    mu_r_loop: do ii = 1,10   !** change 10 to 9, since range of mu_r is 0-8  CONFIRM
```
Bottom line: if you think we can write these files in stone, then we can probably put them in the repo. We can keep the target `p3_create_tables`, but _not_ run it as part of 'all'.

As for the second question: sometimes I build only what I need. E.g., I run cmake, then cd into `tests/coupled/physics_only/shoc_cld_p3_rrtmgp`, and make only in that folder, so that I do not have to build Homme (which is a bit time consuming).

PeterCaldwell:
Ok, that all makes sense. I agree that we should merge this PR in the short term and perhaps later decide to additionally add these tables to the inputdata server.

PeterCaldwell (approved):

bartgol:
Turns out it is very easy to just download files, and use the `p3_tables_setup` exec to just "test" that the downloaded tables match what the p3 src code would generated.

@jgfouca If you upload the 8 tables (mu_r, revap, vm, vn, once for double precision and once for single precision) on the server, I can push a second branch where we just download them, and run tests to verify they match what p3 would generate.

jgfouca:
@bartgol , are these files really independent of build/architecture? Won't they be affected by sizeof<Real>? I think regenerating them is safer.

bartgol:
> @bartgol , are these files really independent of build/architecture? Won't they be affected by `sizeof<Real>`? I think regenerating them is safer.

They are def affected by the size of Real (which is why there should be 2 copy of each, one for single and one for double). I did not think about architecture. I know they are always generated in F90, so no CUDA shenanigans there. If we generate them in a debug build (with optimizations off) we should also get consistent values across archs, but I'm not 100% sure.

I guess it's fine to re-generate them then (even though, having "data" be different across archs is a bit unsettling, semantically speaking).

jgfouca (approved):

PR #1529, opened 2022-04-06
Improve scorpio output class

### LINKED ISSUE #1522: Fix time-hog IO routines
As of now, IO is a time hog for two reasons:

- We tally the averages on host, instead of doing it on device
- We tally the averages also for INSTANT output on non-write steps (a complete waste of time)

Edit: I'm adding another reason, still part of being a time-hog. @ndkeen reported that a 2.5GB file takes about 20s to write, which is very slow. We should make sure that 1) we use scorpio correctly (e.g., set up rearrangers, sub-communicator, and whatnot), and 2) we waste as little time as possible in our interface layer to scorpio.

bartgol:
This PR addresses #1522, reducing time wasting and moving certain operations on device.

Closes #1522 .


AaronDonahue (approved):
This PR looks good too.

PR #1531, opened 2022-04-07
Minor fixes in property check classes

bartgol:
Namely:
- Make `can_repair = false` the default (where repair is possible at all). Repairing is an intrusive operation, and should be explicitly requested by the user. Besides, when adding checks for inputs, fields are read-only, so repairs are not possible. The default behavior should be ok everywhere.
- Fix a check in PropertyCheck. The check should throw if input field is read-only but has to be repaired, but it was throwing for all read-only fields (regardless of whether repair was requested or not).


jeff-cohere (approved):
Seems like a good idea!

PR #1532, opened 2022-04-07
Add bounds checks for atmospheric processes

wlin7:
Add additional bounds checks for p3, shoc, spac, cloud fraction, radiation and dynamics.

[BFB]


components/scream/src/physics/p3/atmosphere_microphysics.cpp
```diff
@@ -175,7 +175,20 @@ void P3Microphysics::init_buffers(const ATMBufferManager &buffer_manager)
 void P3Microphysics::initialize_impl (const RunType /* run_type */)
 {
   // Set property checks for fields in this process
-  add_invariant_check<FieldWithinIntervalCheck>(get_field_out("T_mid"),140, 500);
+  add_invariant_check<FieldWithinIntervalCheck>(get_field_out("T_mid"),140.0, 500.0);
```
wlin7, line 178: Haven't been able to verified if these checks are functioning properly. Purposely resetting the T_mid bound checks between 140 and 270 didn't trigger any warning or fatal error in a multi-step run.

PeterCaldwell:
This is really great @wlin7 - exactly what I was hoping you'd do. Now if only we can figure out if/why the bounds don't seem to be actually triggering errors. Maybe add an "if outside bounds, print val" statement just before a bounds check you expect to fail?

wlin7:
Don't fully understand what said in the template in atmospher_process.hpp. 
It has  `CheckFailHandling CFH = CheckFailHandling::Fatal`. I took it as it would already trigger fatal error if checked failure.

wlin7:
Some info on the bounds used. The bounds are set to 2x to 10x of max absolute values found in literatures (for effective radius of cloud liquid and ice ), IFS instantaneous data (T, q, u, v, qc, qi). Surface precipitation rate upper bound amounts to 3600 mm/hr, which is more than 10x of world hourly precipitation record.

bartgol:
@wlin7 the checks also need to be _activated_ from input file. Inside the atm proc section, you need to ensure that you have
```
Enable Precondition Checks: true
Enable Postcondition Checks: true
```
If not specified, the default should be to run them, but still worth checking.

bartgol:
Ah, another very important thing, which probably is the culprit here!

PR #1531 is changing this, but as of now, in master, the property checks can repair a field by default If you want to error out, you need to also pass "false" as the last arg to the ctor:

```
add_postcondition_check<FieldWithinIntervalCheck>(get_field_out("precip_liq_surf"),0.0, 0.001,false);
```
The bool at the end specifies whether the property check can "repair" the field or not. Unfortunately, the default now is true for all checks (except for "FieldNanCheck", which has no "reasonable" way to repair the field).

PeterCaldwell:
Ugh, that's a bad default behavior! Glad we're changing that.

wlin7:
Thanks for the tips, @bartgol . Currently, the scream_input.yaml does enable all the checks. 
The repair option makes perfect senses to me. When check() not passed, the code always goes to check for pc->can_repair() , and silently moves to next check if so. Only when not repairable, it would give warning or crash out depending the CFH flag.

I am going to give it a try anyway and update the codes.

wlin7:
Latest commit added a warning messsage to stdout (e3sm.log) when a repair takes place after a check failed. Previously it went silently.

components/scream/src/physics/spa/atmosphere_prescribed_aerosol.cpp
```diff
@@ -190,6 +191,15 @@ void SPA::initialize_impl (const RunType /* run_type */)
...
   m_buffer.spa_temp.hyam = SPAData_start.hyam;
   m_buffer.spa_temp.hybm = SPAData_start.hybm;
+
+  // Set property checks for fields in this process
+  add_postcondition_check<FieldWithinIntervalCheck>(get_field_out("nc_activated"),0.0,1.0e6,false);
+  // upper bound set to 1.01 as max(g_sw)=1.00757 in current ne4 data assumingly due to remapping
+  add_postcondition_check<FieldWithinIntervalCheck>(get_field_out("aero_g_sw"),0.0,1.01,false);
+  add_postcondition_check<FieldWithinIntervalCheck>(get_field_out("aero_g_sw"),0.0,1.01,false);
+  add_postcondition_check<FieldWithinIntervalCheck>(get_field_out("aero_ssa_sw"),0.0,1.0,false);
+  add_postcondition_check<FieldWithinIntervalCheck>(get_field_out("aero_tau_sw"),0.0,1.0,false);
```
brhillman, line 201: Optical depth can in general be greater than one. I’d probably make these `FieldLowerBoundCheck`s instead. Is there a reason to limit max values?

components/scream/src/physics/spa/atmosphere_prescribed_aerosol.cpp
```diff
@@ -190,6 +191,15 @@ void SPA::initialize_impl (const RunType /* run_type */)
   //       to performe_time_interpolation.
   m_buffer.spa_temp.hyam = SPAData_start.hyam;
   m_buffer.spa_temp.hybm = SPAData_start.hybm;
+
+  // Set property checks for fields in this process
+  add_postcondition_check<FieldWithinIntervalCheck>(get_field_out("nc_activated"),0.0,1.0e6,false);
+  // upper bound set to 1.01 as max(g_sw)=1.00757 in current ne4 data assumingly due to remapping
+  add_postcondition_check<FieldWithinIntervalCheck>(get_field_out("aero_g_sw"),0.0,1.01,false);
```
brhillman, line 198: I think in general asymmetry parameter can be [-1, 1].

components/scream/src/physics/spa/atmosphere_prescribed_aerosol.cpp
```diff
@@ -190,6 +191,15 @@ void SPA::initialize_impl (const RunType /* run_type */)
...
+  add_postcondition_check<FieldWithinIntervalCheck>(get_field_out("aero_g_sw"),0.0,1.01,false);
```
brhillman, line 199: Duplicated check for `aero_g_sw`?

components/scream/src/share/atm_process/atmosphere_process.cpp
```diff
@@ -197,12 +197,16 @@ void AtmosphereProcess::run_postcondition_checks () const {
     auto check = pc->check();
     if (check) {
       continue;
-    }
+    }
```
brhillman, line 200: Extra space after ending brace was inserted here, please remove.

components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
```diff
@@ -270,6 +271,10 @@ void RRTMGPRadiation::initialize_impl(const RunType /* run_type */) {
   gas_concs.init(gas_names_yakl_offset,m_ncol,m_nlay);
   rrtmgp::rrtmgp_initialize(gas_concs);
 
+  // Set property checks for fields in this process
+
```
brhillman, line 275: We should add precondition checks for the surface albedos read in from the surface models here (which should be between 0 and 1).

components/scream/src/physics/spa/atmosphere_prescribed_aerosol.cpp
wlin7: Not for aerosol optical depth, though. AODVIS is typically around 0.13.

components/scream/src/physics/spa/atmosphere_prescribed_aerosol.cpp
wlin7: There are from the SPA prescribed input file. min (aer_g_sw)=0. I wonder if the definition is the same.

components/scream/src/share/atm_process/atmosphere_process.cpp
wlin7: Done.

components/scream/src/physics/spa/atmosphere_prescribed_aerosol.cpp
wlin7: Ben, you are right that in theory the aerosol asymmetry parameter can range from -1 (for 100% backscattering) to 1 (100% forward scattering).  Measurement derived and modeled values (and similarly SPA prescribed) fall within [0,1]. ne4 file however contains some overshooting (as large as 1.0076), while ne30 file are all with [0,1].

components/scream/src/physics/spa/atmosphere_prescribed_aerosol.cpp
wlin7: The around 0.13 value is a global mean in observation. Local values can be larger. In the ne30 SPA file, the max is 0.18. The max is much smaller in the ne4 SPA file. We could indeed see much larger value in ultra high resolution, if not directly mapping from ne30.

components/scream/src/physics/spa/atmosphere_prescribed_aerosol.cpp
brhillman: Right but I don’t think we should set these checks to only span the values in our current input file, I think they should represent the theoretical limits. If, for example, we had something more sophisticated than SPA that actually computed values (which we will if EAMxx is ever going to provide the driver for EAMv4), then you can imagine asymmetry parameter < 0. These checks should allow for that, not just verify the current input data we have.

brhillman: The overshooting in the ne4 data is a problem we should correct as well, probably rather than fudging the bounds checks to allow input data that we know is suspect.

wlin7: Agreed but for a later date, esp. for the upper bound. Otherwise ne4 tests cannot proceed until the ne4 SPA data is updated. Also worth a thought on the purpose of this bound check in SPA. For example, since we know data generated by ne30 model are all > 0, a presence of negative value, though within theoretical limit of the asymmetry parameter, it could indicate an improper hi-order mapping (if used in certain occasion).

components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
bartgol: There was a problem in the `PropertyCheck` class, which was fixed in #1531 . If your branch does not contain the commits in that branch, you may want to update the branch first. You have to manually merge master anyways, since github reports the PR as not mergeable, due to conflicts.

wlin7: Great you already fixed it, @bartgol . The 2nd bullet of the commit message for #1531 states exactly the problem I am having. I am going to rebase my branch with master.

PeterCaldwell:
Hi @wlin7 - what's the status here? Is this PR ready to merge?

brhillman (approved):

brhillman:
I think @wlin7 needs to merge or rebase on master for an automerge, since there are conflicts to resolve.

wlin7:
This PR is now ready. With the several recent fixes (PropertyCheck class, p3_tables) by @bartgol , all the new bounds checks worked well in the SMS_D_Ln2_P24x1.ne4_ne4.F2000SCREAMv1.cori-knl_intel test.  Instead of rebasing, master was merged into the branch for this PR. The only previous conflict was the change in the call to rrtmgp_initialize in atmosphere_radiation.cpp, which is resolved to be consistent with the current master.

AaronDonahue:
Are the fails we are seeing in this PR due to the new bounds checks?  Or something else.  Either way, it looks like we have a few things that need to be fixed.  @wlin7 ?

brhillman:
Looks like anything involving SPA is failing.

AaronDonahue:
yes, it might be worth checking the error logs to see if things are going out of bounds or if it's something else.

brhillman:
I did a manual test on mappy and seeing things like this:
```
/home/bhillma/codes/scream/branches/wlin/atm/add_bounds_checks/components/scream/tests/uncoupled/spa/spa_stand_alone.cpp:27: FAILED:
  {Unknown expression after the reported line}
due to unexpected exception with message:
  /home/bhillma/codes/scream/branches/wlin/atm/add_bounds_checks/components/
  scream/src/share/atm_process/atmosphere_process.cpp:231: FAIL:
  false
  Error! Failed post-condition check (cannot be repaired).
    - Atm process name: Simple Prescribed Aerosols (SPA)
    - Property check name: aero_ssa_sw within interval [0, 1]
    - Atmosphere process MPI Rank: 1
```

brhillman:
It looks like the max value for `aero_ssa_sw` ends up comparing as slightly greater than 1. Printing out the max value in the property check code reports 1 even though the test fails, but if I increase the tolerance to, e.g., 1.0 + 1e-8 in the check, the test passes.

bartgol:
> It looks like the max value for `aero_ssa_sw` ends up comparing as slightly greater than 1. Printing out the max value in the property check code reports 1 even though the test fails, but if I increase the tolerance to, e.g., 1.0 + 1e-8 in the check, the test passes.

To print doubles, one should use `std::cout << std::setprecision(17) << ...`, to ensure no truncation is done to print.

brhillman:
> To print doubles, one should use `std::cout << std::setprecision(17) << ...`, to ensure no truncation is done to print.

Thanks for the tip @bartgol with that, I see that `aero_ssa_sw` is just *slightly* out of bounds:
```
ERROR: FieldWithinIntervalCheck aero_ssa_sw min = 5.9476225260351741e-08; max = 1.0000000000000009
```

wlin7:
Thanks for debugging, @brhillman . Push a commit to add 1e-8 to the upper bound of aero_ssa_sw.  Hopefully this will allow the SPA auto tests to pass. The out-of-bound condition didn't exist in the spa's ne4 data for the full model test. Where does it come from in the spa alone?

wlin7:
@brhillman , what is the command-line format to run the standalone tests? Next time I should also run some.

PeterCaldwell:
I like this page as a reference for running unit tests: https://acme-climate.atlassian.net/wiki/spaces/NGDNA/pages/1217528411/Running+SCREAMv1+Unit+Tests+on+LLNL+Machines+Quartz+Syrah+and+Lassen 

Not sure whether it answers your question...

wlin7:
Thanks, Peter. That is what I was looking for.

bartgol:
@wlin7 I just added a note on that page on how to set up a couple of files that allow you to run test-all-scream on your own laptop/workstation, without the need to use a supported scream machine. Of course, your machine might not catch the errors you see in the PR (e.g., your machine will likely be a CPU machine, so you won't catch GPU errors), but it should be enough to catch "simple" errors (e.g., bad path to file, read OOB, ...). It's a 5-10 min investment, but can pay off in the long run.

I always (well, let's say "often") run `test-all-scream -l` on my laptop, and catch most of the bugs before the AT.

brhillman:
This is failing single precision tests, I assume because the input data has values very slightly above 1. When these are truncated to single precision, it looks like things get a little worse. Adding a print of the min/max when checks are failed, I see this in the single precision tests:
```
FieldWithinIntervalCheck: aero_ssa_sw; min = 0; max = 1.0000002384185791
```
So the tolerance we added to the check just wasn't large enough to handle the single precision case I guess. I wonder if just fixing the input data would solve this, without having to continue to loosen the tolerance. This field really should NOT have values even a little bit greater than 1.

brhillman:
Okay, another thing I missed is that the unit tests are using ne2 I think, so this is doing a remap (using a mono map) from ne4 to ne2. So perhaps the remap is causing the values to go slightly out of bounds.

PeterCaldwell:
> the unit tests are using ne2 I think, so this is doing a remap (using a mono map) from ne4 to ne2

Ah that makes sense...

brhillman:
We *could* make the SPA optics fields "repairable", so that the checks will do the limiting for us. That would at least prevent `RRTMGP_EXPENSIVE_CHECKS` from later complaining about the slightly out of bounds values if ever used.

PeterCaldwell:
> We could make the SPA optics fields "repairable"
that's what we did for v0, right?

bartgol:
Is the value slightly above 1 going to mess up some calculations? Like some illegal operation along the line of `sqrt(1-x)`? If that's the case, then yes, we should repair fields. But if the value above 1 is still numerically ok, then I would not bother repair the field.

Now, for unit tests, we can have "coarse" tolerances (like 100*eps), so long as we are confident that this is ok. If this sort of remap is to be done in real runs too, then maybe we should reconsider the type of remapping (again, assuming the oscillations are not acceptable).

Another option is to make the post-condition checks for SPA use `CheckFailHandling::Warning` instead of `CheckFailHandling::Error`. This should print a message, so that we can verify that the out-of-bounds value is not "too" large.

brhillman:
Yeah, I think so. Let me try that. The danger would be that if something were to go wrong with the interp, we wouldn't really know about it, and might be just truncating garbage to the interval [0,1]. That's not what's going on here though.

brhillman:
The problem with NOT repairing the fields is that if `RRTMGP_EXPENSIVE_CHECKS` are enabled, then RRTMGP will abort the run if these values are even slightly out of bounds. I think we want to be replacing most of those checks with explicit checks we handle in the interface, but note that Peter is right that we DO repair these fields in v0, to avoid just that issue.

brhillman:
Okay, I think this is the thing to do. I changed the can-repair flag to true, and the tests pass. When I finish the PR to add more info to the warning/error messages, we will get a report of min/max values for this, similar to this:
```
WARNING: Post-condition property check failed and repaired.
  - Property check name: aero_g_sw within interval [0, 1]
  - Atmosphere process name: Simple Prescribed Aerosols (SPA)
  - minval = 0
  - maxval = 1.0000002384185791
```

bartgol:
The `CheckAndRepairWrapper` class was designed for this kind of situation, but looking at the code I don't think it does what we need.

In theory, the class should run a "coarse" check (crapping out if, say, `aero_blah > 1 + 100*eps`), and use a "strict" repair (setting `aero_blah = 1`). Unfortunately, I made some errors in the impl, so it repairs only if the check craps out, which is not what we want (we want to fix, regardless). However, this would entail running two property checks: one that checks for "not-too-large" values, and one that does the clip.

I think I should re-write the class, so that it does the check and the repair with one sweep.

PeterCaldwell:
> In theory, the class should run a "coarse" check (crapping out if, say, aero_blah > 1 + 100*eps), and use a "strict" repair (setting aero_blah = 1)

It would be cool to only repair values that are roundoff bad instead of really bad. I was thinking repair would just set *any* bad value to its nearest acceptable bound and print a warning message. Is that what the code does now? Perhaps we can live with that in the short-term (i.e. at least long enough to integrate this PR)?

bartgol:
> > In theory, the class should run a "coarse" check (crapping out if, say, aero_blah > 1 + 100*eps), and use a "strict" repair (setting aero_blah = 1)
> 
> It would be cool to only repair values that are roundoff bad instead of really bad. I was thinking repair would just set _any_ bad value to its nearest acceptable bound and print a warning message. Is that what the code does now? Perhaps we can live with that in the short-term (i.e. at least long enough to integrate this PR)?

Say you put the "checker" to report pass if `x<1+100eps` everywhere, and the "fixer" to set `x=min(x,1)`. Right now, the class runs the fixer _only if_ the checker reports a fail. That's clearly not what we want. What I want, is to

1. run checker and report FAIL if `x>1+100*eps`.
2. run the fixer if the checker _did not_ fail.

This is somewhat pointing to a design flaw of `PropertyCheck`. I think what the class should report with the `check` method should not just be pass-vs-fail, but rather "pass", "fail, but fixable", "fail, catastrophic". I _could_ work on changing the property check interface, as well as all property checks, to allow for "tolerances". Imho, it would make the checks more powerful.

brhillman:
I guess I'm confused, because looking at it, it looks like it's just repairing no mater what if `check()` fails. Isn't that what this is doing?
```
// Check the property, and if not satisfied, proceed to repair.
// The default impl is to run check() and repair() in sequence. If you can
// perform both in a single call, for performance reason, you should
// override this method.
void PropertyCheck::check_and_repair () const {
  EKAT_REQUIRE_MSG(can_repair(),
      "Error! This property check cannot repair^R.\n"
      "  - PropertyCheck name: " + name() + "\n");
  if (not check()) {
    repair();
  }
}
```

bartgol:
> I guess I'm confused, because looking at it, it looks like it's just repairing no mater what if `check()` fails. Isn't that what this is doing?
> 
> ```
> // Check the property, and if not satisfied, proceed to repair.
> // The default impl is to run check() and repair() in sequence. If you can
> // perform both in a single call, for performance reason, you should
> // override this method.
> void PropertyCheck::check_and_repair () const {
>   EKAT_REQUIRE_MSG(can_repair(),
>       "Error! This property check cannot repair^R.\n"
>       "  - PropertyCheck name: " + name() + "\n");
>   if (not check()) {
>     repair();
>   }
> }
> ```

Yes, that's what it's doing. Which is not what we need here. This class has a design flaw, and should probably _not_ be used.

brhillman:
I get what you're saying, and I like the proposal to check for both roundoff-level issues and catastrophic issues, but I think what we need right now is just the limiter on these values that will cause problems, which sounds like the simpler impl you have now accomplishes. We DO still get a warning that the check failed, and that `PropertyCheck` is repairing the field. For greater safety, if it makes you feel better, could we add two sets of checks to the SPA interface, one with 100*eps as the limit with repair = false, and one with [0,1] with repair = true?

PeterCaldwell:
I agree with Ben. I think the existing implementation is good enough for now. The more sophisticated checker/fixer is cool and worth doing, but I don't think it should be a super high priority.

brhillman:
Okay, I pushed a commit for this "temporary" fix then. To be clear, the behavior now is that `PropertyCheck` will ALWAYS repair the SPA optics, no matter how bad they are, which is not what we want long term but should suffice until we can revamp `CheckAndRepairWrapper` to check for both roundoff-level and catastrophic bounds violations. This will at least enable the same "bumpers" we have on the v0 code to keep RRTMGP from complaining about the SPA optics values.

bartgol (approved):

PR #1536, opened 2022-04-11
Add radiation super-stepping

brhillman:
Add option for radiation super-stepping (only running radiation every N steps as opposed to every step). This is enabled by setting `rad_frequency` (units of number of steps) to a value greater than 1 in the RRTMGP section of `input.yaml`. This is done by default for the `rrtmgp_standalone` test to test this functionality, where `rad_frequency = 3`.

### DIFF
--- a/components/scream/src/physics/rrtmgp/tests/rrtmgp_unit_tests.cpp
+++ b/components/scream/src/physics/rrtmgp/tests/rrtmgp_unit_tests.cpp
@@ -1,5 +1,5 @@
 #include "catch2/catch.hpp"
-#include "physics/rrtmgp/rrtmgp_heating_rate.hpp"
+#include "physics/rrtmgp/rrtmgp_utils.hpp"
 #include "physics/rrtmgp/scream_rrtmgp_interface.hpp"
 #include "YAKL.h"
 #include "physics/share/physics_constants.hpp"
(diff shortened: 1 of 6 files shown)


components/scream/src/physics/rrtmgp/tests/rrtmgp_unit_tests.cpp
```diff
@@ -403,3 +403,18 @@ TEST_CASE("rrtmgp_test_compute_broadband_surface_flux") {
...
+    REQUIRE(scream::rrtmgp::radiation_do(2, 0) == true);
+    REQUIRE(scream::rrtmgp::radiation_do(2, 1) == false);
+    REQUIRE(scream::rrtmgp::radiation_do(2, 2) == true);
+    REQUIRE(scream::rrtmgp::radiation_do(2, 3) == false);
+
+    REQUIRE(scream::rrtmgp::radiation_do(3, 0) == true);
+    REQUIRE(scream::rrtmgp::radiation_do(3, 1) == false);
+    REQUIRE(scream::rrtmgp::radiation_do(3, 2) == false);
+    REQUIRE(scream::rrtmgp::radiation_do(3, 3) == true);
+    REQUIRE(scream::rrtmgp::radiation_do(3, 4) == false);
```
AaronDonahue, line 419: might be worth adding a `radiation_do(3,6) == true` check.  To test when nsteps > the rad frequency and we want to update radiation.

AaronDonahue (approved):
Looks good, I had one comment but it isn't critical.  I also had one question come to mind, what would happen if a user set radiation to subcycle, for some reason?  I suppose we would want the radiation frequency to be 1 in that case so that it will actually subcycle.  I guess you could see some weird cases where you set the supercycle to 2 and the subcycle to 3 so you end up with a radiation timestep of 2/3*dt_phys...  I think infrastructure-wise it would still work, but I wonder aloud if we want (or can even have) some checks that we are not subcycling and supercycling at the same time.

brhillman:
@AaronDonahue interesting question. I think with how I've implemented `radiation_do` here (querying number of atmos steps taken and seeing if divisible by `rad_frequency`), even if the AD is subcycling, `radiation_do` would only be true every `rad_frequency` steps, since I would assume that for a subcycle step, the overall value of nsteps is the same (is that correct?). So, I think this would mean that, if you set subcycling to 2, and rad supercycling to 3, then radiation would still only run every third step, but on each third step it would run *twice*. This is probably not what anyone would ever want, so a check would probably be appropriate. This seems like something that would be the responsibility of radiation, if radiation is  handling its own supercycling. The alternative, which I think we've considered as a wish list item, is to implement supercycling at the driver level rather than having radiation handle it. Not sure it would ever make sense to supercycle other physics processes, but I guess it would be nice to have one approach to handle both sub/super-cycling. That would prevent this kind of nonsense I would think.

AaronDonahue:
Passes `mappy` and `weaver`, testing manually on `quartz` now to substitute for a `blake` test.

AaronDonahue:
This PR passed all tests on `quartz`.  I'm going to manually merge.

Issue #1540, opened 2022-04-12
Problem at build with p3_tables_setup exe on Cori with optimized build


ndkeen:
Using a scream master of Apr12th, can reproduce with `SMS_Ld1_P16x1.ne4_ne4.F2000SCREAMv1.cori-knl_gnu` (ie OPT build) after commenting out abort as noted here https://github.com/E3SM-Project/scream/issues/1539.

Note that after upstream merge that pushed gnu version to 10.3 on cori, we need the same flag that treats certain errors as warnings for now, which I will get into repo soon via this branch `ndk/scream/cori-allow-gnuv10-builds` 

```
[ 98%] Generating /global/cscratch1/sd/ndk/e3sm_scratch/cori-knl/bio-apr11/f30cpu.F2000SCREAMv1.ne30_ne30.bio-apr11.gnu.12s.n011b16x1.RECe.N576.ts150.s8.barr/run/data/mu_r_table_vals.dat8, /global/cscratch1/sd/ndk/e3sm_scratch/cori-knl/bio-apr11/f30cpu.F2000SCREAMv1.ne30_ne30.bio-apr11.gnu.12s.n011b16x1.RECe.N576.ts150.s8.barr/run/data/revap_table_vals.dat8, /global/cscratch1/sd/ndk/e3sm_scratch/cori-knl/bio-apr11/f30cpu.F2000SCREAMv1.ne30_ne30.bio-apr11.gnu.12s.n011b16x1.RECe.N576.ts150.s8.barr/run/data/vm_table_vals.dat8, /global/cscratch1/sd/ndk/e3sm_scratch/cori-knl/bio-apr11/f30cpu.F2000SCREAMv1.ne30_ne30.bio-apr11.gnu.12s.n011b16x1.RECe.N576.ts150.s8.barr/run/data/vn_table_vals.dat8
[...]

PR #1545, opened 2022-04-14
For Cori, add flag to allow building with gnuv10 and above.

ndkeen:
Only affects Cori and only for GNU compiling.
Add the `-fallow-argument-mismatch` flag only for gnu v10 and above (same as was done on Perlmutter).
With last upstream merge, the gnu compiler was bumped to v10.3 on Cori.

I tested with DEBUG build, but cannot test out-of-the-box with OPT due to https://github.com/E3SM-Project/scream/issues/1540 (but I did test with work-around).


bartgol:
I've seen this issue before in E3SM, and also in netcdf-fortran github repo. I thought it got fixed in E3SM already...

bartgol (approved):

bartgol:
Skipping AT testing, since changes won't affect any machine used by the AT.

ndkeen:
Well, first it's more like a work-around than a fix. And it was me who made the changes in E3SM -- and now here.

PR #1546, opened 2022-04-14
Improve documetation of nml XML

jgfouca:
This was driven by comments that were stranded in the last PR.

### DIFF
--- a/components/scream/cime_config/buildnml
+++ b/components/scream/cime_config/buildnml
@@ -3,9 +3,11 @@
 """
 Namelist creator for E3SM's SCREAM component
 
-This scripts takes the namelist_defaults_scream.xml file from the repo
+This script takes the namelist_defaults_scream.xml file from the repo
 and produces the processed $case/namelist_scream.xml. From the processed
 file, raw input files are produced which should not be modified by users.
+The user is allowed to modify $case/namelist_scream.xml either directly
+via text editor or via atm-config-chg.
 
 This script can be internally tested standalone via: buildnml --test
 
@@ -74,13 +76,13 @@ def do_cime_vars(entry, case):
     >>> do_cime_vars('hi ${invalid} there', case)
     Traceback (most recent call last):
       ...
-    CIME.utils.CIMEError: ERROR: Cannot resolve yaml entry 'hi ${invalid} there', CIME has no value for 'invalid'
+    CIME.utils.CIMEError: ERROR: Cannot resolve XML entry 'hi ${invalid} there', CIME has no value for 'invalid'
     """
     m = CIME_VAR_RE.search(entry)
     while m:
         cime_var = m.groups()[0]
         value = case.get_value(cime_var)
-        expect(value is not None, "Cannot resolve yaml entry '{}', CIME has no value for '{}'".format(entry, cime_var))
+        expect(value is not None, "Cannot resolve XML entry '{}', CIME has no value for '{}'".format(entry, cime_var))
         entry = entry.replace("${{{}}}".format(cime_var), str(value))
         m = CIME_VAR_RE.search(entry)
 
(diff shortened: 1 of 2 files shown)


components/scream/cime_config/buildnml
```diff
@@ -3,9 +3,11 @@
 """
 Namelist creator for E3SM's SCREAM component
 
-This scripts takes the namelist_defaults_scream.xml file from the repo
+This script takes the namelist_defaults_scream.xml file from the repo
 and produces the processed $case/namelist_scream.xml. From the processed
 file, raw input files are produced which should not be modified by users.
+The user is allowed to modify $case/namelist_scream.xml either directly
```
bartgol, line 9: Do we really want to let them modify it directly? I mean, we can't prevent this from happening, but we should *highly* discourage it. If the user accidentally modifies the structure of the file, we might generate something completely off, with an invalid xml format being our best-case scenario.

PeterCaldwell:
Looks great! I'm not approving now just to give others a chance to comment before auto-merge.

jeff-cohere:
> Looks great! I'm not approving now just to give others a chance to comment before auto-merge.

At least we learn from each other's goofs.

components/scream/cime_config/namelist_defaults_scream.xml
```diff
@@ -31,6 +31,8 @@ tweaking their $case/namelist_scream.xml file.
       <partmethod nlev="72">420</partmethod>
 
     Selector info will not be duplicated in the processed $case/namelist_scream.xml file.
+    If multiple parameter elements have selectors that evaulate to True, then the last
```
jeff-cohere, line 34: evaulate -> evaluate

bartgol: I think the comment might be misleading for some one. It's not the last selector within a single element that takes precedence, but the last element among those (with the same name) that have all selectors that eval to true. I think a mini example might help. Something tiny, like
```
If bar=3 and baz=4, then
   <foo bar="3">my_foo</foo>
   <foo baz="4">your_foo</foo>
would generate an xml with content
   <foo>your_foo</foo> 
```
And maybe we should add a warning, cause this can lead to defaults being inadvertently shadowed, depending on the order in scream_namelists_defaults.xml

jeff-cohere:
Looks good to me, too. I found one typo.

components/scream/cime_config/namelist_defaults_scream.xml
bartgol: Actually, we can probably remove this sentence, since you go over the rules later, and rule 1 clearly states this fact already.

jgfouca: Done.

bartgol (commented):
Looks good. Got a couple of comments, but nothing to hold this back.

jgfouca:
What the heck? How did AT fail with AT: Integrate without testing?

bartgol:
> What the heck? How did AT fail with AT: Integrate without testing?

And all builds passed!

jgfouca:
@jeff-cohere , I just realized you had a lot of comments in the older PR that I did not notice since they were under "hidden items". I'm going through them now.

bartgol:
> What the heck? How did AT fail with AT: Integrate without testing?

Ah, you pushed a commit to the PR right after the builds finished, but before the AT was done with the PR, so it marked it as failed, to prevent merging.

jgfouca:
Ah, unlucky race condition.

components/scream/cime_config/buildnml
jgfouca: Remember, with the changes introduced yesterday to the prototype, the metadata of parameters, computed from the repo XML file, must be obeyed in the $case file. So it's not completely left up to the user to not make a mistake.

components/scream/cime_config/namelist_defaults_scream.xml
jgfouca: I removed the sentence; it wasn't even using the right terminology.

jgfouca:
OK, as far as I can tell, this is all ready. Can someone approve?

jeff-cohere (approved):

PR #1548, opened 2022-04-14
Jgfouca/upstream merge 2022 04 14

jgfouca:
There were only a couple minor conflicts in CMake files, but there were a number of files that needed fixes due to the big CIME reorganization.


bartgol (approved):

AaronDonahue:
@jgfouca, the ELM fix that LC needs went into E3SM master yesterday.  Not sure if it makes sense with this PR, but can we update the upstream merge to include yesterdays master?  Or we can always do a second one once this is merged...

jgfouca:
@AaronDonahue , doing a second upstream merge on the same branch is easy. Doing it now.

AaronDonahue (approved):

PR #1555, opened 2022-04-19
Have dynamic expose omega as a computed field

bartgol:
Also build a single version of the dyn library for all our unit tests, with qsize_d=10 (before we built one with qsize_d=35 and one with qsize_d=4).

### DIFF
--- a/components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
+++ b/components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
@@ -166,6 +166,7 @@ void HommeDynamics::set_grids (const std::shared_ptr<const GridsManager> grids_m
   add_field<Required>("phis",          FL({COL         },{ncols           }),m2/s2, rgn);
   add_field<Computed>("p_int",         FL({COL,    ILEV},{ncols,  nlev_int}),Pa,    rgn,N);
   add_field<Computed>("p_mid",         FL({COL,     LEV},{ncols,  nlev_mid}),Pa,    rgn,N);
+  add_field<Computed>("omega",         FL({COL,     LEV},{ncols,  nlev_mid}),Pa/s,  rgn,N);
   add_group<Updated>("tracers",rgn,N, Bundling::Required);
 
   // Dynamics grid states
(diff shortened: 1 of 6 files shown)


AaronDonahue (approved):
Looks good!  Thanks for getting on the fix so quickly Luca.  I'm approving, but I a question comes to mind.  Did you test that with these changes SHOC gets different results?  I suppose we will see that by expecting SCREAMv1 tests to be non-BFB...

bartgol:
> Looks good! Thanks for getting on the fix so quickly Luca. I'm approving, but I a question comes to mind. Did you test that with these changes SHOC gets different results? I suppose we will see that by expecting SCREAMv1 tests to be non-BFB...

Yes, and I have some issues with what I saw. First, I manually removed omega from the input file, and verified the dyn+physics test still passes, so I'm confident that SHOC is indeed using the omega field produced by dynamics (rather than relying on the input file). Then I cprnc-ed the output with current master, and found 40 fields differ.

Some fields that differ are the spa fields, which I thought were "inputs", but I guess we save them after interpolation on the vert grid, which could be affected by using a different omega.

PeterCaldwell:
> Some fields that differ are the spa fields, which I thought were "inputs", but I guess we save them after interpolation on the vert grid, which could be affected by using a different omega.

spa reads in variables on pressure levels, then interpolates them to the model grid. So any non-bfb change would change them, right? I don't think the issue is specific to omega (since vert interp doesn't care about omega)...

AaronDonahue:
@bartgol , I'm confused why this didn't fail the mappy tests with DIFF in the SCREAMv1 tests.  Doesn't adding the `SCREAMv1` label cause the CIME tests to be run?

bartgol:
> > Some fields that differ are the spa fields, which I thought were "inputs", but I guess we save them after interpolation on the vert grid, which could be affected by using a different omega.
> 
> spa reads in variables on pressure levels, then interpolates them to the model grid. So any non-bfb change would change them, right? I don't think the issue is specific to omega (since vert interp doesn't care about omega)...

But different omega will lead to changes in SHOC output, which will change Homme forcing, which will change dp... So I think we can expect the change in omega to impact spa output too. At least after 2 timesteps.

PeterCaldwell:
> But different omega will lead to changes in SHOC output, which will change Homme forcing, which will change dp... So I think we can expect the change in omega to impact spa output too. At least after 2 timesteps.

Yeah, I totally agree. That's what I was trying to say...

PeterCaldwell:
> @bartgol , I'm confused why this didn't fail the mappy tests with DIFF in the SCREAMv1 tests. Doesn't adding the SCREAMv1 label cause the CIME tests to be run?

I agree with Aaron - I think we didn't expect this to be BFB but it was because we didn't pass the CIME flag. This is a bit weird...

PR #1556, opened 2022-04-20
Better memory management for SPA data read

AaronDonahue:
This commit improves the infrastructure for pulling SPA data from a file which will solve an issue we have been seeing with out-of-memory errors on certain machines, in particular `quartz` and `cori`.

This commit changes the data read to just pull spa data for the columns that a particular rank is interested in.  Before this change each rank would pull all of the data, which was an issue for the ne30 dataset.

### DIFF
--- a/components/scream/src/physics/spa/spa_functions_impl.hpp
+++ b/components/scream/src/physics/spa/spa_functions_impl.hpp
@@ -491,14 +491,15 @@ void SPAFunctions<S,D>
           SPAInput&             spa_data)
 {
   // Ensure all ranks are operating independently when reading the file, so there's a copy on all ranks
-  ekat::Comm self_comm(MPI_COMM_SELF);
+  auto comm = spa_horiz_interp.m_comm;
 
   // We have enough info to start opening the file
   std::vector<std::string> fnames = {"hyam","hybm","PS","CCN3","AER_G_SW","AER_SSA_SW","AER_TAU_SW","AER_TAU_LW"};
   ekat::ParameterList spa_data_in_params;
   spa_data_in_params.set("Field Names",fnames);
   spa_data_in_params.set("Filename",spa_data_file_name);
-  AtmosphereInput spa_data_input(self_comm,spa_data_in_params);
+  spa_data_in_params.set("Skip_Grid_Checks",true);  // We need to skip grid checks because multiple ranks may want the same column of source data.
+  AtmosphereInput spa_data_input(comm,spa_data_in_params);
 
   // Note that the SPA data follows a conventional GLL grid format, albeit at a different resolution than
   // the simulation.  For simplicity we can use the scorpio_input object class but we must construct a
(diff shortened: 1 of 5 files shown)


components/scream/src/physics/spa/spa_functions_impl.hpp
```diff
@@ -599,7 +608,7 @@ ::update_spa_data_from_file(
   Kokkos::deep_copy(target_grid_loc_h, spa_horiz_interp.target_grid_loc);
   for (int idx=0;idx<spa_horiz_interp.length;idx++) {
```
bartgol, line 609: Unrelated, but why are we doing this big interp loop on host? This seems like a beefy loop, so might benefit from device acceleration. Although, being done only once per month (right?), we should prob check how much it costs first.

components/scream/src/physics/spa/tests/spa_one_to_one_remap_test.cpp
```diff
@@ -62,6 +64,12 @@ TEST_CASE("spa_one_to_one_remap","spa")
   SPAFunc::SPAHorizInterp spa_horiz_interp;
   spa_horiz_interp.m_comm = spa_comm;
   SPAFunc::set_remap_weights_one_to_one(ncols,min_dof,dofs_gids,spa_horiz_interp);
+  // Make sure one_to_one remap has the correct unique columns
+  REQUIRE(spa_horiz_interp.num_unique_cols==my_ncols);
+  for (int ii=0;ii<spa_horiz_interp.num_unique_cols;ii++) {
+    // Also check that each entry is uniquely placed in the vector
+    REQUIRE(spa_horiz_interp.source_local_col_map.count(dofs_gids_h(ii))==1);
```
bartgol, line 71: std::map does not allow for duplicates, so I'm not sure this check does anything. The count can be either 0 or 1. The only way that the size of the map is right (you check it before the loop) but `count` returns 0 is if your `source_local_col_map` contains gids that are not in the grid, which is really impossible.

components/scream/src/physics/spa/tests/spa_read_remap_from_file_test.cpp
```diff
@@ -54,6 +54,12 @@ TEST_CASE("spa_read_remap_data","spa")
   REQUIRE(test_total_ncols == tgt_grid_ncols_total);
 
   SPAFunc::get_remap_weights_from_file(remap_file_name,tgt_grid_ncols_total,0,dofs_gids,spa_horiz_interp);
+  // Make sure one_to_one remap has the correct unique columns
+  REQUIRE(spa_horiz_interp.num_unique_cols==src_grid_ncols);
+  for (int ii=0;ii<src_grid_ncols;ii++) {
+    // Also check that each entry is uniquely placed in the vector
+    REQUIRE(spa_horiz_interp.source_local_col_map.count(ii)==1);
```
bartgol, line 61: Same comment as for the other test: is this check adding any value?

components/scream/src/physics/spa/spa_functions_impl.hpp
```diff
@@ -520,27 +527,29 @@ ::update_spa_data_from_file(
...
   using namespace ShortFieldTagsNames;
   FieldLayout scalar1d_layout { {LEV}, {source_data_nlevs} };
-  FieldLayout scalar2d_layout_mid { {COL}, {spa_horiz_interp.source_grid_ncols} };
-  FieldLayout scalar3d_layout_mid { {COL,LEV}, {spa_horiz_interp.source_grid_ncols, source_data_nlevs} };
-  FieldLayout scalar3d_swband_layout { {COL,SWBND, LEV}, {spa_horiz_interp.source_grid_ncols, nswbands, source_data_nlevs} }; 
-  FieldLayout scalar3d_lwband_layout { {COL,LWBND, LEV}, {spa_horiz_interp.source_grid_ncols, nlwbands, source_data_nlevs} };
+  FieldLayout scalar2d_layout_mid { {COL}, {num_local_cols} };
+  FieldLayout scalar3d_layout_mid { {COL,LEV}, {num_local_cols, source_data_nlevs} };
+  FieldLayout scalar3d_swband_layout { {COL,SWBND, LEV}, {num_local_cols, nswbands, source_data_nlevs} }; 
+  FieldLayout scalar3d_lwband_layout { {COL,LWBND, LEV}, {num_local_cols, nlwbands, source_data_nlevs} };
```
bartgol, line 552: I put the comment here, cause github doesn't let me comment below.

I think this function is way more expensive that it should be. Besides the grid construction, we have lots of views allocations, as well as the AtmosphereInput construction/setup (with all the scorpio stuff). I think we should be able to avoid these.

Yes, this routine is only called once per month (actually twice), but still. Why making the call more expensive? Do we know how much does a SPA call cost when the month has changed, compared to when we're in the middle of it? That might tell us if we're dramatically inefficient (and need to fix this) or if the overhead is negligible in the grand scheme of things.

components/scream/src/physics/spa/spa_functions_impl.hpp
```diff
@@ -599,7 +608,7 @@ ::update_spa_data_from_file(
   Kokkos::deep_copy(target_grid_loc_h, spa_horiz_interp.target_grid_loc);
   for (int idx=0;idx<spa_horiz_interp.length;idx++) {
     auto src_wgt = weights_h(idx);
-    int  src_col = source_grid_loc_h(idx);
+    int  src_col = spa_horiz_interp.source_local_col_map[source_grid_loc_h(idx)];
```
bartgol, line 611: Something seems overly complicated here. Do we need to store the src grid gids, if all you need is their local id? Instead of the map, you could store a vector of LIDs (of size `spa_horiz_interp.length`), specifying the LID of the current src col in the unique grid dof list:
```
int src_col = spa_horiz_interp.source_local_col_lid[idx];
```

bartgol (approved):
I have a few improvement suggestions, but I don't want to prevent merging if the current status prevents a memory hog.

That said, I do think SPA could be improved here and there.

bartgol (approved):

PR #1562, opened 2022-04-22
Add error message output from property checks

### LINKED ISSUE #1426: Make PropertyCheck::check return also a string, along with the check (bool) result.
The string can be used by the property check to give more context about the test that failed. E.g., for a FieldPositivityCheck, it could return the lowest value, and perhaps also its position (like a col id and level id). The AtmosphereProcess class would then print this message (along with other info already currently displayed) in case of failed test.

brhillman:
Add error message output from property checks. The return type for the `PropertyCheck::check()` function is changed to a `struct` containing the fields `pass` and `msg` to report both error status and error message. The pre and post-condition checks in `atmosphere_process` are correspondingly updated to report this information when a check fails. Fixes #1426.

[B4B]


bartgol (approved):
Much more flexible!

PR #1569, opened 2022-04-26
Add extra RRTMGP bounds checks and debug info

brhillman:
Add extra RRTMGP bounds checks and debug info, so that min/max values are reported when cloud or aerosol optics go outside expected ranges.

[B4B]

### DIFF
--- a/components/scream/src/physics/rrtmgp/rrtmgp_utils.hpp
+++ b/components/scream/src/physics/rrtmgp/rrtmgp_utils.hpp
@@ -5,6 +5,9 @@
 #include "cpp/rrtmgp_const.h"
 #include "YAKL.h"
 
+using yakl::intrinsics::maxval;
+using yakl::intrinsics::minval;
+
 namespace scream {
     namespace rrtmgp {
         // Provide a routine to compute heating due to radiative fluxes. This is
@@ -32,7 +35,7 @@ namespace scream {
             });
         }
 
-        bool radiation_do(const int irad, const int nstep) {
+        inline bool radiation_do(const int irad, const int nstep) {
             // If irad == 0, then never do radiation;
             // Otherwise, we always call radiation at the first step,
             // and afterwards we do radiation if the timestep is divisible
@@ -43,6 +46,26 @@ namespace scream {
                 return ( (nstep == 0) || (nstep % irad == 0) );
             }
         }
+
+
+        // Verify that array only contains values within valid range, and if not
+        // report min and max of array
+        template <class T> bool check_range(T x, Real xmin, Real xmax, std::string msg) {
+            bool pass = true;
+            auto _xmin = minval(x);
+            auto _xmax = maxval(x);
+            if (_xmin < xmin or _xmax > xmax) {
+                pass = false;
+                std::cout 
+                    << std::string(msg) 
+                    << ": one or more values outside range"
+                    << "; minval = " << _xmin 
+                    << "; maxval = " << _xmax << "\n";
+            }
+            return pass;
+        }
+
     }
+
 }
 #endif
--- a/components/scream/src/physics/rrtmgp/tests/rrtmgp_unit_tests.cpp
+++ b/components/scream/src/physics/rrtmgp/tests/rrtmgp_unit_tests.cpp
@@ -425,3 +425,19 @@ TEST_CASE("rrtmgp_test_radiation_do") {
     REQUIRE(scream::rrtmgp::radiation_do(3, 5) == false);
     REQUIRE(scream::rrtmgp::radiation_do(3, 6) == true);
 }
+
+TEST_CASE("rrtmgp_test_check_range") {
+    // Initialize YAKL
+    if (!yakl::isInitialized()) { yakl::init(); }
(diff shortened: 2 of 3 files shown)


brhillman:
@PeterCaldwell I tagged you here to see if this was what you were wanting to see from better radiation debugging info when values go outside expected ranges. The way this is setup now is that warning messages will be printed when optical properties go outside expected ranges, but this does not implement bumpers. Bumpers would be easy to implement here though if we want that, hence this is labeled WIP for the time being. This does not require extra debug flags right now, but the impl of the min/max values may not be performant on GPU so we might want to do that too.

components/scream/src/physics/rrtmgp/rrtmgp_utils.hpp
```diff
@@ -43,6 +46,26 @@ namespace scream {
...
+            bool pass = true;
+            auto _xmin = minval(x);
+            auto _xmax = maxval(x);
+            if (_xmin < xmin or _xmax > xmax) {
+                pass = false;
+                std::cout 
+                    << std::string(msg) 
+                    << ": one or more values outside range"
+                    << "; minval = " << _xmin 
+                    << "; maxval = " << _xmax << "\n";
```
PeterCaldwell, line 63: Shouldn't you also print out what the allowable range was?

brhillman: Done.

components/scream/src/physics/rrtmgp/rrtmgp_utils.hpp
```diff
@@ -43,6 +46,27 @@ namespace scream {
...
+
+        // Verify that array only contains values within valid range, and if not
+        // report min and max of array
+        template <class T> bool check_range(T x, Real xmin, Real xmax, std::string msg) {
+            bool pass = true;
+            auto _xmin = minval(x);
+            auto _xmax = maxval(x);
+            if (_xmin < xmin or _xmax > xmax) {
+                pass = false;
+                std::cout
```
bartgol, line 59: MInor (if you have time and agree): you could add an optional arg to the signature, which can be used to redirect output somewhere else:
```
template<typename T>
bool check_range (T x, Real xmin, Real xmax, const std::string&, const std::ostream& out = std::cout);
```
The default behavior stays the same (output to screen), but now one can do
```
std::stringstream ss;
check_range (my_val, my_min, my_max, my_msg, ss);

write_to_file (ss);
```

components/scream/src/physics/rrtmgp/tests/rrtmgp_unit_tests.cpp
```diff
@@ -425,3 +425,19 @@ TEST_CASE("rrtmgp_test_radiation_do") {
...
+    // Initialize YAKL
+    if (!yakl::isInitialized()) { yakl::init(); }
+    // Create some dummy data and test with both values inside valid range and outside
+    auto dummy = real2d("dummy", 2, 1);
+    // All values within range
+    memset(dummy, 0.1);
+    REQUIRE(scream::rrtmgp::check_range(dummy, 0.0, 1.0, "dummy") == true);
+    // At least one value below lower bound
+    parallel_for(1, YAKL_LAMBDA (int i) {dummy(i, 1) = -0.1;});
+    REQUIRE(scream::rrtmgp::check_range(dummy, 0.0, 1.0, "dummy") == false);
```
bartgol, line 439: If you pass the extra arg to `check_range`, you could even check that the error msg is what you expect it to be.

bartgol (approved):
Looks good. I have one minor suggestion for an improvement, but nothing worth blocking this.

components/scream/src/physics/rrtmgp/rrtmgp_utils.hpp
brhillman: Oh that's a great idea, thanks for the suggestion! I'll implement, I want to add a few additional checks anyways.

bartgol:
This PR might be good to go. Let's wait for #1582 to be merged (and fix the v1 build error), then we can re-test.

brhillman:
Huh, failed SHOC sp test on blake:
```
Build type full_sp_debug failed. Here's a list of failed tests:
86:shoc_tests_omp13
```
That doesn't really make sense.

AaronDonahue:
There is a particular seed that will fail shoc_tests.

PeterCaldwell:
> Huh, failed SHOC sp test on blake

What Aaron said. Just try another round of AT?

PeterCaldwell:
Maybe we should make the tolerance a bit larger for SP for that test?

AaronDonahue:
> Maybe we should make the tolerance a bit larger for SP for that test?

If I recall for the lassen fail we got `nanf=nanf` which will always fail.  So its not a tolerance thing if its the same issue.

PeterCaldwell:
> If I recall for the lassen fail we got nanf=nanf which will always fail

Yikes. If there are realistic situations where SHOC crashes, we should probably fix that. If these crashes are because we are feeding SHOC unrealistic input... maybe we should stop that...

brhillman:
Another round of testing it is...

PR #1570, opened 2022-04-26
Add memory usage information to atm log

bartgol:
With some suggestion from @ndkeen I put together a super small, very minimal, memory usage utility function, and use it in the driver, to report mem usage at the end of each time step. It can use either the `getrusage` function or check `/proc/self/statm` for the information. The two methods seem to produce _very_ different answers (like 180MB vs 900MB), but they can both be used to determine if the memory usage is increasing.

NOTE: CIME already has this feature, but this PR allows us to get this info also for standalone runs, which might be easier to debug.


jgfouca (dismissed):

bartgol:
Note: this might require to update some mach files, in case the system does not support neither `getrusage` nor `/proc/self/statm`.

jgfouca (approved):

PR #1573, opened 2022-04-27
Add python script to query atm configs, similar to xmlquery

bartgol:
Also, rename `atm-config-chg` to `atmchange`, to mirror the name of the case `xmlchange` and `xmlquery` tools.

Note: as of now (04/27/22 at 12:15pm) this PR looks huge, but that's b/c it's built on top of another currently open PR (#1521). As soon as that one is merged, almost all commits listed below will "disappear" from this page.


jgfouca (dismissed):
Looks great!

AaronDonahue:
I think this PR should be ready... The mappy fail is likely because 2010 was a new test and maybe we didn't have baselines?  Should we retest and let the AT take it over?

bartgol:
@AaronDonahue The PR was indeed likely failing b/c lack of baselines. Jim and I were investigating why the baselines were not correctly blessed (didn't figure it out, remains a mistery, though the next bless job went through correctly).

I had to rebase to fix the conflicts anyways, so AT must run once more.

bartgol:
I had to solve some conflicts with master, but didn't change the substance of the PR. Since Jim approved, and it's passing, I'm gonna merge now.

PR #1574, opened 2022-04-28
Report indices where check fails in field property checks

bartgol:
This should help identify where things go wrong in the model.

Notes:
- Currently, we simply report the rank-local indices of one entry of the field where the check failed. Ideally, we should also provide coordinates of this point (e.g., lat/lon and level). We need to slightly change the ctor of the class, to also accept a grid or a grid manager, and retrieve the info later on.
- FieldWithinIntervalCheck actually reports min/max and their location _regardless_ of pass/fail status. I'm not sure if this will ever be useful, but it didn't cost anything (compared to the pfor), so I thought we might as well store the message anyways.

@PeterCaldwell I've been told this might be needed to debug v1, so I am going to submit the PR even without the first bulled upgrade. In the meantime, one can probably retrieve the global coords of the point running with 1 rank (since local col idx = col global id, and one can then check coords of that col in the nc file).


PeterCaldwell:
So if I understand this PR correctly, it returns not just the max and min values when an out-of-bounds event is detected, but also the i,j,k for that biggest and smallest value. And these values are rank local so if we wanted to know the global rank, we would need to run with a single MPI? If I'm right about the single MPI, this indexing will be of limited use because I don't think we can fit ne30 into the memory of a single node. This PR still has value by telling us the vertical index though, and knowing the other indices doesn't hurt. 

I'm confused about "FieldWithinIntervalCheck actually reports min/max and their location regardless of pass/fail status" - I don't understand whether you're saying that we'll always print all the ranks max and min forever more, or just if someone chooses to do so. I can't imagine we'd ever want to log the max and min vals on all MPIs all the time, but I could easily imagine we might want to log the biggest and smallest values on ranks that exceed some threshold without stopping the run or correcting the values.

bartgol:
I don't think we want to pollute atm.log with a list of GIDs on each rank. I will just go ahead and do the col id mods once I'm done with the gptl fix.

bartgol:
@AaronDonahue @PeterCaldwell I added column GID and lat/lon information to the output. One needs to create the property check passing the grid. If not, no location or lat/lon is reported (just that the test failed, and the values). The output should be something like

```
  "FieldNaNCheck failed.
    - entry (1,1,2)
    - lat/lon: (1.000000, -1.000000)
```

PeterCaldwell:
> One needs to create the property check passing the grid.

So to be clear, when this PR is merged we will have the ability to print lat/lon/etc but someone needs to search/replace all instances of pre/post condition checks to add the grid to their calls before we will actually start seeing this info?

components/scream/src/share/property_checks/field_nan_check.cpp
```diff
@@ -112,10 +130,40 @@ PropertyCheck::CheckResult FieldNaNCheck::check_impl() const {
   }
 
   PropertyCheck::CheckResult check_result;
-  check_result.pass = (num_invalid == 0);
+  check_result.pass = (invalid_idx < 0);
   check_result.msg = "";
   if (not check_result.pass) {
-    check_result.msg = std::string("FieldNaNCheck failed; ") + std::to_string(num_invalid) + " invalid values found\n";
+    auto indices = unflatten_idx(layout.dims(),invalid_idx);
```
brhillman, line 136: Is this going to end up reporting ALL of the indices that fail the invalid check? That might end up creating a very large string below?

components/scream/src/share/property_checks/field_within_interval_check.cpp
```diff
@@ -114,13 +170,65 @@ PropertyCheck::CheckResult FieldWithinIntervalCheck::check_impl () const
...
+    has_latlon = m_grid->has_geometry_data("lat") && m_grid->has_geometry_data("lon");
+    if (has_latlon) {
+      lat = m_grid->get_geometry_data_host("lat");
+      lon = m_grid->get_geometry_data_host("lon");
+    }
+  }
+
+
+  check_result.msg += "  minimum:\n";
+  check_result.msg += "    - value: " + std::to_string(minmaxloc.min_val) + "\n";
```
PeterCaldwell, line 207: @bartgol - One thing I'm noticing today is that the min and max values are currently getting printed as floats instead of in exponential notation. This makes it hard to tell how bad roundoff errors are: we're getting messages like "x is outside [0,1]. Min=0.0000 and max=1.0000". Would it be easy to use exponential notation for just the min and max (and not the lat/lon)?

bartgol: Ah, yeah, good point. I can take care of that.

components/scream/src/share/property_checks/field_nan_check.cpp
bartgol: Only one of the columns, yeah. And I think it's the last. Unfortunately, keeping track of all the nan's would be a bit more difficult. Do you think it would be necessary?

AaronDonahue: In terms of NaN, just finding out there is one of them is probably all we need, in my opinion.  Mins/maxes are a different story.  I could see a case where the user would like to know if the list of min/max values is scattered or clustered.  But that is also probably a lot of work.  I would be content to get at least one of the offending columns printed to screen with the location info.

PeterCaldwell: Just knowing one location where bad data is occurring would be super helpful. Knowing all the locations would of course be nice, but let's not make perfect be the enemy of the good. We can make another PR some time in the (far) future if we decide that we need all locations. My feeling is that we won't because we just want a general sense of where things are going bad, which is probably all in one area. If it's not, you have bigger problems to deal with.

bartgol:
@PeterCaldwell @brhillman @AaronDonahue Tests pass, so this is ready for a final review.

PeterCaldwell:
I just hacked p3 T_mid invariant_condition check to crash if greater than 300 K and ran an ne4 test with this branch and got the following error:
```
Error! Failed pre-condition check (cannot be repaired).
  - Atm process name: Microphysics
  - Property check name: T_mid within interval [140, 300]
  - Message: Check failed.
  T_mid within interval [140, 300]
  minimum:
    - value: 187.783
  maximum:
    - value: 303.485
```
I'm confused why I don't see the lat/lon/lev as part of the resulting error message. I also modified post_condition for qv and also didn't see the lat/lon/lev info. Am I missing something?

bartgol:
@PeterCaldwell it might be that you are creating the property check _without_ passing the grid pointer.

I didn't want current branches to start having trouble, so I left the ability of creating FPC's without passing the grid pointer, in which case, no lat/lon info is reported. However, I realize this might be confusing. I will remove the possibility to do so, and force ppl to pass a grid at construction time. I will make sure all existing FPC's created in our atm procs follow this pattern.

PeterCaldwell:
> it might be that you are creating the property check without passing the grid pointer

Ah shoot, I forgot that you and I discussed this previously. I should have rephrased my last question as "none of our existing tests pass the grid pointer"! Yeah, it would be nice if this PR could make that happen.

bartgol:
> > it might be that you are creating the property check without passing the grid pointer
> 
> Ah shoot, I forgot that you and I discussed this previously. I should have rephrased my last question as "none of our existing tests pass the grid pointer"! Yeah, it would be nice if this PR could make that happen.

Ok, every test should report lat/lon now.

PeterCaldwell:
One last comment before approving: currently, the vertical level is gleaned from the line `- entry: (1,1)`. It's unclear to users whether the horiz index is the first entry and vertical is the second entry, or vice versa. I don't think this can be easily fixed because "entry" is written out by looping over all dimensions of a given variable so we don't know a priori what the ordering will be. Instead, users should be savvy enough to know SCREAM's index ordering. 

I'm adding this comment just in case it is actually easy to add this info... otherwise we should just integrate.

bartgol:
> One last comment before approving: currently, the vertical level is gleaned from the line `- entry: (1,1)`. It's unclear to users whether the horiz index is the first entry and vertical is the second entry, or vice versa. I don't think this can be easily fixed because "entry" is written out by looping over all dimensions of a given variable so we don't know a priori what the ordering will be. Instead, users should be savvy enough to know SCREAM's index ordering.
> 
> I'm adding this comment just in case it is actually easy to add this info... otherwise we should just integrate.

Ah, we can print the layout of the field (in terms of tags, not on-process dimensions). That should be easy, and prob informative.

bartgol:
@PeterCaldwell the last commit should be addressing your request.

PeterCaldwell (approved):

AaronDonahue:
@PeterCaldwell , do you mind re-reviewing this?  It looks like it passed the AT and may be ready for merge.

PR #1586, opened 2022-04-28
scream v1 hires test suite

jgfouca:
Please advise on the test cases.

### DIFF
--- a/cime_config/tests.py
+++ b/cime_config/tests.py
@@ -453,11 +453,11 @@ _TESTS = {
             )
     },
 
-    "e3sm_scream_v1_hires" : {
+    "e3sm_scream_v1_medres" : {
         "time"  : "03:00:00",
         "tests" : (
             "SMS_D_Ln2.ne30_ne30.F2000SCREAMv1",
-            "SMS_D_Ln2.ne30_ne30.F2000-SCREAMv1-AQP1",
+            "SMS_Ln2.ne30_ne30.F2000-SCREAMv1-AQP1",
             )
     },
 
--- a/cime_config/tests.py
+++ b/cime_config/tests.py
@@ -453,6 +453,15 @@ _TESTS = {
             )
     },
 
+    "e3sm_scream_v1_hires" : {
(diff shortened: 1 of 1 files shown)


jgfouca:
Some comments: I reused the test cases from the regular v1 test suite, bumped the grids to `ne30_ne30` and removed the hardcoded 24 proc counts. I think we want to let CIME/machine config select the proc layout for these bigger tests.

cime_config/tests.py
```diff
@@ -453,6 +453,15 @@
             )
     },
 
+    "e3sm_scream_v1_hires" : {
+        "time"  : "03:00:00",
+        "tests" : (
+            "SMS_D_Ln2.ne30_ne30.F2000SCREAMv1",
+            "SMS_D_Ln2.ne30_ne30.F2000-SCREAMv1-AQP1",
+            )
```
PeterCaldwell, line 461: Maybe have one of these tests include _D_ and the other not? ie have one debug build and one optimized build? Otherwise, not much to complain about!

bartgol:
I totally read "scream v1 hires test suite" as "scream v1 has a job opening for a new testsuite"...

Side note: is ne30 considered "hi-res"? I thoguht ne4 was ultra-low-res, ne30 was low-res, and ne256 was hi-res.

PeterCaldwell:
> is ne30 considered "hi-res"? I thoguht ne4 was ultra-low-res, ne30 was low-res, and ne256 was hi-res.

Luca makes a good point. I agree that ne30 is low res... just not as low as ne4.

jgfouca:
@PeterCaldwell , @bartgol , I am happy to change the grid. Would `ne256_ne256` work? I almost never use grids that high so I'm unfamiliar with them.

jgfouca:
@PeterCaldwell , is "medres" (medium res) ok? So "e3sm_scream_v1_medres"? I don't want to put "ne30" in the name since that would commit us to only using ne30 grids.

PeterCaldwell:
Yeah, medres would be fine with me.

PR #1587, opened 2022-04-28
Adds the XML defaults to run an ne30 case for CIME

AaronDonahue:
This commit will make it possible to build an ne30 testcase for
SCREAMv1 and run out-of-the-box.

It does this by adding the `hgrid` conditional to some of the XML
entries.

### DIFF
--- a/components/scream/cime_config/namelist_defaults_scream.xml
+++ b/components/scream/cime_config/namelist_defaults_scream.xml
@@ -163,8 +163,8 @@ tweaking their $case/namelist_scream.xml file.
     <!-- Simple Prescribed Aerosols (SPA) -->
     <spa inherit="physics_proc_base">
       <SPA__Remap__File>none</SPA__Remap__File>
-      <SPA__Data__File hgrid="ne4np4">${DIN_LOC_ROOT}/atm/scream/init/spa_file_unified_and_complete_ne4_20220428.nc</SPA__Data__File>
-      <SPA__Data__File hgrid="ne30np4">${DIN_LOC_ROOT}/atm/scream/init/spa_file_unified_and_complete_ne30_20220428.nc</SPA__Data__File>
+      <SPA__Data__File hgrid="ne4np4">${DIN_LOC_ROOT}/atm/scream/init/spa_file_unified_and_complete_ne4_scream.nc</SPA__Data__File>
+      <SPA__Data__File hgrid="ne30np4">${DIN_LOC_ROOT}/atm/scream/init/spa_file_unified_and_complete_ne30_scream.nc</SPA__Data__File>
     </spa>
 
     <!-- Radiation -->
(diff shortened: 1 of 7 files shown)


AaronDonahue:
I'm setting this to WIP now because the last thing I think it needs is an audit of the homme namelist parameters.  I changed the obvious one - `se_ne`, but I imagine there are others.

@oksanaguba , @mt5555 , @bartgol or others.  Can you let me know if there are homme namelist parameters that need to be adjusted given ne4 vs. ne30?

PeterCaldwell:
on the v1 tasks page, Mark wrote: 
- Correct dycore defaults for all resolutions:  se_ftype=2, cube_sphere_map=2, nu=3.4e-8, hypervis_scaling=3.0.   ( some runs were using older E3SM V1 or SCREAM V0 settings for these parameters)
- dycore resolution dependent settings:  dtime, se_tstep,  nu_top (recommended defaults given on  [EAM's HOMME dycore](https://acme-climate.atlassian.net/wiki/spaces/DOC/pages/1044644202) ).

AaronDonahue:
@PeterCaldwell , great thanks!  We will have to stick with ftype=0 for the time being, until we get that sorted out in our interface.  But we can change the default once we get that working.

mt5555:
settings look good to me.  i'll update that confluence page with the note about some settings being postponed ( SL transport, NH mode, cube_sphere_map and se_ftype).

AaronDonahue:
@mt5555 , I see in the Confluence page conflicting recommendations for `nu`.  It says `nu=3.4e-8` is recommended for all runs, but in the table below it says ne30 should have `nu=1e15`.  Which one should I follow?

mt5555:
3.4e-8 is correct.  That other value is was used in E3SM V1, and in E3SM V2 only for ne30, for backwards compatibility with V1.  

When you wrote "the table below", which table is that?  lets correct that table.

AaronDonahue:
@mt5555 - I left a comment on the Confluence page and tagged you in it where the "table below" is.

mt5555:
thanks!  fixed.

components/scream/cime_config/namelist_defaults_scream.xml
```diff
@@ -162,7 +163,8 @@ tweaking their $case/namelist_scream.xml file.
     <!-- Simple Prescribed Aerosols (SPA) -->
     <spa inherit="physics_proc_base">
       <SPA__Remap__File>none</SPA__Remap__File>
-      <SPA__Data__File>${DIN_LOC_ROOT}/atm/scream/init/spa_file_unified_and_complete_ne4_scream.nc</SPA__Data__File>
+      <SPA__Data__File hgrid="ne4np4">${DIN_LOC_ROOT}/atm/scream/init/spa_file_unified_and_complete_ne4_20220428.nc</SPA__Data__File>
+      <SPA__Data__File hgrid="ne30np4">${DIN_LOC_ROOT}/atm/scream/init/spa_file_unified_and_complete_ne30_20220428.nc</SPA__Data__File>
     </spa>
```
PeterCaldwell, line 168: Don't we have some ne2 tests which need a SPA regrid file? Do we need to include the capability for that in this PR? I guess not because this PR doesn't break tests?

components/scream/cime_config/namelist_defaults_scream.xml
```diff
@@ -233,34 +237,41 @@ tweaking their $case/namelist_scream.xml file.
...
   <!-- Homme control namelist -->
   <ctl_nl>
-    <cubed_sphere_map>0</cubed_sphere_map>
+    <cubed_sphere_map>2</cubed_sphere_map>
     <disable_diagnostics>False</disable_diagnostics>
-    <dt_remap_factor>1</dt_remap_factor>
-    <dt_tracer_factor>1</dt_tracer_factor>
+    <dt_remap_factor>2</dt_remap_factor>
+    <dt_tracer_factor hgrid="ne4np4">6</dt_tracer_factor>
+    <dt_tracer_factor hgrid="ne30np4">6</dt_tracer_factor>
```
PeterCaldwell, line 250: why not just have one entry for tracer_factor and don't condition on hgrid since you have the same value for all grids?

PeterCaldwell: Same with hypervis_subcycle_q?

mt5555: I missed this the  first time I looked.  These are the settings for SL transport.  I thought that wasn't ready yet for EAMxx, so scream v1 should be running with Eulerian transport?

PeterCaldwell: Yes, that's correct. We should be using Eulerian transport settings for v1.

AaronDonahue: @mt5555 should I be looking at the defaults for a different section of the Confluence page then?

AaronDonahue: @bartgol , per Peter's question.  I was going to ask, is it possible to have one XML setting be set to another.  In this case, according to the homme recommendations `hypervis_subcycle_q` should be set to match `dt_tracer_factor`.

AaronDonahue:
Have we agreed to keep the default timesteps to match EAM?  So 1800s for ne30, 7200s for ne4?

PeterCaldwell:
Yeah, let’s just integrate it keeping the EAM timesteps.

From: AaronDonahue ***@***.***>
Date: Friday, April 29, 2022 at 8:56 AM
To: E3SM-Project/scream ***@***.***>
Cc: Caldwell, Peter M. ***@***.***>, Mention ***@***.***>
Subject: Re: [E3SM-Project/scream] Adds the XML defaults to run an ne30 case for CIME (PR #1587)

Have we agreed to keep the default timesteps to match EAM? So 1800s for ne30, 7200s for ne4?

components/scream/cime_config/namelist_defaults_scream.xml
PeterCaldwell: I wasn't asking if we could link different variables to be the same - that seems too complicated and not really necessary. Instead, I was just noticing that you had entries for hypervis_subcycle_q and dt_tracer_factor which were different for different horiz resolution... but you assigned these quantities to the same value regardless of their resolution. Might as well not condition on resolution at all. Would save 2 lines of xml and avoid setting weird precedents for people who don't know what they're doing.

AaronDonahue: Yes, your question just reminded me of something I wanted to ask in general.  We can certainly save two lines by keeping the default as is.  I arranged it this way because in the Confluence page these were one of those namelist options that were supposed to change by resolution.  It just so happen that ne4 and ne30 have the same value, but when we go to ne120+ they should change.  Plus, I was ready to be told that ne4 might be wrong - I took that value from a SCREAMv0-HR ne4 run.  I'm happy either way.  I guess ultimately, I'll wait to hear back from HOMME folks about what changes, if any, are needed for using the Eulerian dycore.  I can consolidate if/when I make those changes.

AaronDonahue:
@PeterCaldwell, @bartgol and @mt5555 , this is ready for re-review.  I've removed the WIP label.

AaronDonahue:
For some strange reason, my newest commit didn't post to the PR.  So it is some kind of limbo.  I'm going to delete this PR and open a fresh one which should hopefully reflect the most up to date version of the branch.

PR #1596, opened 2022-05-02
Adjust default SPA data files to reflect the new set of data files.

AaronDonahue:
The old SPA data files had the wrong units for CCN3. which led to a
model with little or no clouds.  This commit adopts updated SPA data
files with the correct units.

This commit also adjusts the bounds for property checks for `nc` in
`P3` and for `nccn_activated` in `SPA` taking into account that with
the correct units these values are expected to be higher.

### DIFF
--- a/components/scream/cime_config/namelist_defaults_scream.xml
+++ b/components/scream/cime_config/namelist_defaults_scream.xml
@@ -162,7 +162,7 @@ tweaking their $case/namelist_scream.xml file.
 
     <!-- Simple Prescribed Aerosols (SPA) -->
     <spa inherit="physics_proc_base">
-      <SPA__Remap__File hgrid="ne4np4">${DIN_LOC_ROOT}/atm/scream/init/map_ne30_to_ne4.nc</SPA__Remap__File>
+      <SPA__Remap__File hgrid="ne4np4">${DIN_LOC_ROOT}/atm/scream/init/map_ne30_to_ne4_mono_20220502.nc</SPA__Remap__File>
       <SPA__Remap__File hgrid="ne30np4">none</SPA__Remap__File>
       <SPA__Data__File>${DIN_LOC_ROOT}/atm/scream/init/spa_file_unified_and_complete_ne30_20220428.nc</SPA__Data__File>
     </spa>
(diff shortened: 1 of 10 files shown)


AaronDonahue:
Worth noting, this PR will fail for an `ne30` run with the following error after 31 timesteps.  Although, if I recall correctly we don't expect even the current master to run ad-infinitum.

```
terminate called after throwing an instance of 'std::logic_error'
  what():  /g/g17/donahue5/Code/e3sm/scream/scream_sandbox/components/scream/src/share/atm_process/atmosphere_process.cpp:250: FAIL:
false
Error! Failed post-condition check (cannot be repaired).
  - Atm process name: Microphysics
  - Property check name: qv within interval [1e-13, 0.2]
  - Atmosphere process MPI Rank: 21
  - Error message: FieldWithinIntervalCheck failed; min = 0.000000; max = 0.091150
```

PeterCaldwell (approved):

hassanbeydoun (dismissed):

PeterCaldwell:
```
Error! Failed post-condition check (cannot be repaired).
  - Atm process name: Microphysics
  - Property check name: qv within interval [1e-13, 0.2]
  - Atmosphere process MPI Rank: 21
  - Error message: FieldWithinIntervalCheck failed; min = 0.000000; max = 0.091150
```
This error should clearly be fixed by making this P3 post-condition check repair negative qv... Like SHOC, P3 removes qv without worrying about whether it makes qv negative. v0 already applies this correction. 

Also, it would be more useful if min and max values were provided in exponential rather than f notation.

AaronDonahue:
@PeterCaldwell , I think this is the last commit.  Everything should work now.  If you could re-review that would be great.

PeterCaldwell (approved):

PeterCaldwell (dismissed):

AaronDonahue:
I tested this locally and it passed.  I will try again on blake.  In the meantime, is it possible that the files needed for the tests are just not being downloaded from the E3SM repo?  I can't really tell without getting a look at the fail messages from the AT.

AaronDonahue:
Are the issues on weaver due to the netCDF version of the new SPA data file?  I'm running into similar issues testing on lassen, the version of netcdf built there doesn't appear to support cdf5 files.  @bartgol or @jgfouca , you both have access to the AT logs, can you confirm that?  Does it pass unit testing on mappy?

jgfouca:
@AaronDonahue , that new ne30 file is not used in any CI testing.

AaronDonahue:
> @AaronDonahue , that new ne30 file is not used in any CI testing.

But we do have a new ne4 SPA data file that we are using in unit testing.

jgfouca:
@AaronDonahue , ah ok. I thought you were thinking that my PR from this morning might have broken things. I will look at the logs.

AaronDonahue:
Huh, strange.  Now `mappy` is outright failing during the CIME tests in the AT for this branch, but the only changes I made impact stand alone testing.  `weaver` and `blake` now both pass.  @jgfouca , can you tell why `mappy` is failing?

jgfouca:
@AaronDonahue something strange is happening with SEMS TPL modules across many sandia machines. I just checked and mappy appears to be working again. We should do a retest.

PeterCaldwell (approved):

Issue #1588, opened 2022-04-28
SHOC v1 input grid size does not have correct units


bogensch:
SHOC requires an estimation of the local grid size (dx & dy) to properly set the maximum allowable value for the dissipation length scale.  In v1 it is being fed in with improper units and needs to be converted to meters.

The offending line of code, line 159 assumes that "area" is in meters when it is in fact in radians:
https://github.com/E3SM-Project/scream/blob/0b56109b076bf0ca76aed88824edfcb31480515c/components/scream/src/physics/shoc/atmosphere_macrophysics.hpp#L155-L159

Lines of code (or a routine) is needed to convert area from radians to meters.  One can use the subroutine "grid_size", that v0 uses (in shoc_intr.F90), as a reference for this:
https://github.com/E3SM-Project/scream/blob/0b56109b076bf0ca76aed88824edfcb31480515c/components/eam/src/physics/cam/shoc_intr.F90#L1171-L1204

AaronDonahue:
@bogensch , so based on this, once we get the proper `grid_dx` and `grid_dy` we will want to replace `cell_length(i)=sqrt(area(i));` with `cell_length(i)=grid_dx(i)`, right?  Since we assume dx=dy?

bogensch:
correct

AaronDonahue:
@bogensch and @PeterCaldwell , can I get your feedback on something for this problem.  It looks like we do store an estimate of area in meters defined as follows:
```
  // Estimate cell area for a uniform grid by taking the surface area
  // of the earth divided by the number of columns
  const Real rearth    = C::r_earth;
  const Real pi        = C::Pi;
  const Real cell_area = 4*pi*rearth*rearth/num_my_cols;
  ```
We are using this in `shoc_main` but not in the `hpp` file that @bogensch notes above.

My first question is, should we be using the grid area estimate that Peter cites above from the F90 code, instead of this approximation?
Follow-up, given our deadline, would it be acceptable for now to just use this approximation for now, which would be a one line change, and then adopt the more advanced approach in the future? 

Just want to make sure I use my time wisely.

AaronDonahue:
Note, before I found this I was able to port the F90 routine to cpp and test it.  But it would take a little more work to make sure the `hpp` has access to the latitude needed for the calculation.

AaronDonahue:
I may also be able to redefine `cell_area` to use the more advanced formula...

AaronDonahue:
[...]

PR #1598, opened 2022-05-02
Fix bug in SHOC with calculation of cell grid length

AaronDonahue:
This commit will update the calculation of a physics grid cell area from HOMME to be in the units of meters-squared rather than radians-squared when used by EAMxx.  Note that the area reported by HOMME is also scaled to the unit-sphere, so the function used here also rescales the area using the radius of Earth.

This commit will introduce a new common physics function which calculates the cell length of a physics column given the area in radians^2 and the latitude of that cell.  New tests are introduced to unit test this function.

This will fix a bug described in #1588 .

### DIFF
--- a/components/scream/src/share/tests/common_physics_functions_tests.cpp
+++ b/components/scream/src/share/tests/common_physics_functions_tests.cpp
@@ -156,10 +156,12 @@ void run_scalar_valued_fns(std::mt19937_64& engine)
 
   // Get dx from grid cell area property tests:
   RealType area, lat, grid_dx;
-  REQUIRE( Check::equal(PF::calculate_dx_from_area(0.0,1.0),0.0) );
-  area = (pi/180.0)*(pi/180.0);
-  REQUIRE( Check::equal(PF::calculate_dx_from_area(area,0.0), coeff_1-coeff_2+coeff_3) );
-  REQUIRE( Check::equal(PF::calculate_dx_from_area(area,pi/2.0), coeff_1+coeff_2+coeff_3) );
+  area = 0.0; lat = 1.0;
+  REQUIRE( Check::equal(PF::calculate_dx_from_area(area,lat),0.0) );
+  area = (pi/180.0)*(pi/180.0); lat = 0.0;
+  REQUIRE( Check::equal(PF::calculate_dx_from_area(area,lat), coeff_1-coeff_2+coeff_3) );
+  lat = pi/2.0;
+  REQUIRE( Check::equal(PF::calculate_dx_from_area(area,lat), coeff_1+coeff_2+coeff_3) );
   
   lat     = pdf_lat(engine);
   area    = pdf_area(engine);
--- a/components/scream/src/share/tests/common_physics_functions_tests.cpp
+++ b/components/scream/src/share/tests/common_physics_functions_tests.cpp
@@ -164,16 +164,9 @@ void run_scalar_valued_fns(std::mt19937_64& engine)
   lat = pi/2.0;
   REQUIRE( Check::equal(PF::calculate_dx_from_area(area,lat), coeff_1+coeff_2+coeff_3) );
   lat = pi/4.0;
-  // Note, pi is not defined up to double precision, so in order to compare with output from
-  // calculate_dx_from_area we need to define the following variables to approximate the output
-  // expected from the use of cosine in the function.  In other words this provides finite
-  // precision values to use in the test.
-  auto cos_pi_4 = std::cos(pi/4.0);
-  auto cos_pi_2 = std::cos(pi/2.0);
-  auto cos_pi   = std::cos(pi);
-  REQUIRE( Check::approx_equal(PF::calculate_dx_from_area(area,lat), coeff_1-coeff_2*cos_pi_2 + coeff_3*cos_pi, test_tol) );
+  REQUIRE( Check::approx_equal(PF::calculate_dx_from_area(area,lat), coeff_1 - coeff_3, test_tol) );
   lat = pi/8.0;
-  REQUIRE( Check::approx_equal(PF::calculate_dx_from_area(area,lat), coeff_1-cos_pi_4*coeff_2 + coeff_3*cos_pi_2, test_tol) );
+  REQUIRE( Check::approx_equal(PF::calculate_dx_from_area(area,lat), coeff_1-std::sqrt(2.0)/2.0*coeff_2, test_tol) );
   lat     = pdf_lat(engine);
   area    = pdf_area(engine);
   REQUIRE( Check::equal(PF::calculate_dx_from_area(area,lat),PF::calculate_dx_from_area(area,-lat)) );
--- a/components/scream/src/share/tests/common_physics_functions_tests.cpp
+++ b/components/scream/src/share/tests/common_physics_functions_tests.cpp
@@ -164,9 +164,16 @@ void run_scalar_valued_fns(std::mt19937_64& engine)
   lat = pi/2.0;
   REQUIRE( Check::equal(PF::calculate_dx_from_area(area,lat), coeff_1+coeff_2+coeff_3) );
   lat = pi/4.0;
-  REQUIRE( Check::approx_equal(PF::calculate_dx_from_area(area,lat), coeff_1-coeff_3, test_tol) );
+  // Note, pi is not defined up to double precision, so in order to compare with output from
+  // calculate_dx_from_area we need to define the following variables to approximate the output
+  // expected from the use of cosine in the function.  In other words this provides finite
+  // precision values to use in the test.
+  auto cos_pi_4 = std::cos(pi/4.0);
+  auto cos_pi_2 = std::cos(pi/2.0);
+  auto cos_pi   = std::cos(pi);
+  REQUIRE( Check::approx_equal(PF::calculate_dx_from_area(area,lat), coeff_1-coeff_2*cos_pi_2 + coeff_3*cos_pi, test_tol) );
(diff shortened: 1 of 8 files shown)


AaronDonahue:
This PR is ready, although I have a few outstanding questions which might cause us to change it a bit.
1. For standalone tests that do not involve `homme` we will still be approximating the area of the cell by assuming uniform sized cells over all of Earth.  Should we fix this?  If so, I think the way to do it is to have standalone tests that don't have `homme` have to read the cell area from file.
2. I am calculating area by first backing out cell_length and than squaring it.  Later, SHOC will go ahead and take the sqrt of that value.  Would it make more sense to create function that just calculates area?  I went this route so that I could copy exactly what SHOC is doing... but I'm open to changing it.

PeterCaldwell:
> For standalone tests that do not involve homme we will still be approximating the area of the cell by assuming uniform sized cells over all of Earth.

I think it's fine to use approx area for standalone calculations

PeterCaldwell:
> I am calculating area by first backing out cell_length and than squaring it. Later, SHOC will go ahead and take the sqrt of that value. Would it make more sense to create function that just calculates area?

I think you're asking whether you should create a function that computes the lengthscale rather than having homme compute a lengthscale to get the area, then have shoc convert area back to that lengthscale. Is this right? I think it's best to allow shoc to compute the lengthscale from area using its own conception of lengthscale and for homme to compute the cell area using its own vision of what that would be. In other words, I think the way you've done it now is best even if it is a bit convoluted. I could be convinced otherwise though...

components/scream/src/share/util/scream_common_physics_functions_impl.hpp
```diff
@@ -6,6 +6,24 @@
...
+ScalarT PhysicsFunctions<DeviceT>::calculate_dx_from_area(const ScalarT& area, const ScalarT& lat)
+{
+  using C = scream::physics::Constants<Real>;
+
+  static constexpr auto coeff_1 = C::earth_ellipsoid1;
+  static constexpr auto coeff_2 = C::earth_ellipsoid2;
+  static constexpr auto coeff_3 = C::earth_ellipsoid3;
+  static constexpr auto pi      = C::Pi; 
+
+  auto mpdeglat = coeff_1 - coeff_2 * std::cos(2.0*lat) + coeff_3 * std::cos(4.0*lat);
```
bartgol, line 21: Any chance we can use this opportunity to switch to a name that means something also to normal humans, instead of "mpdeglat"? :)

AaronDonahue: Sure can... @bogensch , do you have an idea for a better variable name?

bogensch: First, I would vote to include the comments above this code block that the F90 version has, that should help make this variable name (and code block) descriptive.  If you desire a more verbose name in addition to this then m_per_degree_lat might be good.

AaronDonahue:
I had to switch the strategy, converting the units directly in the homme interface led to issues with the coupler because it expects the original units for the data passed from the ATM.  So instead I opted to assume the area is always in radians-squared and convert directly in the SHOC interface.  Which means I had to update the approximate area calculated in the physics grid when it is not Dynamics Driven.

AaronDonahue:
I confirmed with @bogensch that cell_lengths reported for an ne30 run looked like they were on the correct order in size.

bogensch (approved):

AaronDonahue:
@bogensch and/or @bartgol can you re-review this PR.  It looks like its passing the AT, although it didn't test SCREAMv1 tests which should be non-BFB so I flagged those tests and a retest.  Either way though, it should be good to go now once those tests reveal a DIFF.

bogensch (dismissed):

components/scream/src/share/tests/common_physics_functions_tests.cpp
```diff
@@ -144,7 +153,22 @@ void run_scalar_valued_fns()
...
+  REQUIRE( Check::equal(PF::calculate_dx_from_area(area,lat),0.0) );
+  area = (pi/180.0)*(pi/180.0); lat = 0.0;
+  REQUIRE( Check::equal(PF::calculate_dx_from_area(area,lat), coeff_1-coeff_2+coeff_3) );
+  lat = pi/2.0;
+  REQUIRE( Check::equal(PF::calculate_dx_from_area(area,lat), coeff_1+coeff_2+coeff_3) );
+  
+  lat     = pdf_lat(engine);
+  area    = pdf_area(engine);
+  grid_dx = (coeff_1 - coeff_2 * std::cos(2.0*lat) + coeff_3 * std::cos(4.0*lat)) * std::sqrt(area)*(180.0/pi);
+  REQUIRE( Check::equal(PF::calculate_dx_from_area(area,lat),grid_dx) );
```
bartgol, line 169: Imho, this is not a real property test, in that we are spelling out the very same formula we have in the fcn definition (probably via copy/paste as well). This only checks that the formula you wrote here is the same as the formula you wrote there.

Instead, I'd like to check more values of lat: pi/4 and maybe pi/8 for instance. Also check that f(lat)==f(-lat). Another thing could be to check that f(area=4) == 2*f(area=1) (this might require some tol, not sute).

AaronDonahue: excellent suggestion.  I'll put that in.

components/scream/src/share/util/scream_common_physics_functions_impl.hpp
```diff
@@ -6,6 +6,27 @@
...
+
+  static constexpr auto coeff_1 = C::earth_ellipsoid1;
+  static constexpr auto coeff_2 = C::earth_ellipsoid2;
+  static constexpr auto coeff_3 = C::earth_ellipsoid3;
+  static constexpr auto pi      = C::Pi; 
+
+  // Now find meters per degree latitude
+  // Below equation finds distance between two points on an ellipsoid, derived from expansion
+  // taking into account ellipsoid using World Geodetic System (WGS84) reference 
+  auto m_per_degree_lat = coeff_1 - coeff_2 * std::cos(2.0*lat) + coeff_3 * std::cos(4.0*lat);
```
bartgol, line 24: This is much more self-documenting than mpdeglat. Thanks!

AaronDonahue: credit goes to @bogensch :-)

components/scream/src/share/tests/common_physics_functions_tests.cpp
```diff
@@ -158,15 +158,19 @@ void run_scalar_valued_fns(std::mt19937_64& engine)
...
+  area = (pi/180.0)*(pi/180.0);
+  lat = 0.0;
   REQUIRE( Check::equal(PF::calculate_dx_from_area(area,lat), coeff_1-coeff_2+coeff_3) );
   lat = pi/2.0;
   REQUIRE( Check::equal(PF::calculate_dx_from_area(area,lat), coeff_1+coeff_2+coeff_3) );
-  
+  lat = pi/4.0;
+  REQUIRE( Check::approx_equal(PF::calculate_dx_from_area(area,lat), coeff_1-coeff_3, test_tol) );
+  lat = pi/8.0;
+  REQUIRE( Check::approx_equal(PF::calculate_dx_from_area(area,lat), coeff_1-std::sqrt(2)/2*coeff_2, test_tol) );
```
bartgol, line 169: If this check fails (due to roundoffs), we can probably remove it (or maybe replace with lat=pi, which should be the same as lat=0).

bartgol (approved):

AaronDonahue:
I'm a bit confused about the fails on weaver.  `p3_run_and_cmp` fails, which shouldn't be impacted by these changes...

AaronDonahue:
I'm checking manually on blake too.

bartgol:
P3 uses pi, and this pr changes the value (at least for double precision). The fact that single precision passes, supports this idea. You can quickly test by reverting (locally) the change of pi, and verify that p3_run_and_cmp passes.

AaronDonahue:
@bartgol , thanks for the idea.  I just confirmed that the changing of Pi did cause all of the fails.  Do we need to change where Pi is defined elsewhere to get these tests back to passing?

bartgol:
> @bartgol , thanks for the idea. I just confirmed that the changing of Pi did cause all of the fails. Do we need to change where Pi is defined elsewhere to get these tests back to passing?

No, we want to keep this new definition for the pi constant. We simply have to force merge, once we are sure that the diffs are only due to increasing the accuracy of pi.

AaronDonahue:
Okay, I can confirm the fails are due to the change in`pi` so I will force merge.  Note, I only confirmed on blake, I'm assuming that we would not expect different behavior on GPUs right?  @bartgol

bartgol:
> Okay, I can confirm the fails are due to the change in`pi` so I will force merge. Note, I only confirmed on blake, I'm assuming that we would not expect different behavior on GPUs right? @bartgol

You could easily double check on weaver as well (they share the same OS, so you already have the mods ready).

AaronDonahue:
I've never accessed Weaver. I guess I could abuse the AT and just revert Pi and see if everything passes, then put that commit back and manually merge...

bartgol:
> I've never accessed Weaver. I guess I could abuse the AT and just revert Pi and see if everything passes, then put that commit back and manually merge...

Running a quick check now

AaronDonahue:
Luca confirmed on Slack that weaver passes without Pi change.  I'm manually merging.

PR #1599, opened 2022-05-03
Use file that exists input repo for ne30

jgfouca:
screami_ne30np4L72_20220329.nc does not exist in input repo.
screami_ne30np4L72_20220413.nc does exist and has a similar name with a newer
date stamp, so I assume that is the one we should be using.


jgfouca:
With this PR, my summit runs of e3sm_scream_v1_medres at least make it to RUN phase.

AaronDonahue (approved):

AaronDonahue:
I'm surprised the other datestamp has disappeared.  I've been using it for awhile... hmmm.  Thanks for catching this @jgfouca .

jgfouca:
I forgot to add skip testing. We don't test this resolution in any CI testing.

PeterCaldwell:
@brhillman - I have a vague recollection that the earlier timestamp had a bug in it and we should always use the new timestamp. Can you confirm?

Issue #1592, opened 2022-05-02
pre-condition error with GNU compiler at ne120 -- Property check name: sfc_alb_dir_vis within interval


ndkeen:
Error with GNU compiler on cori after turning off SPA. Must still be init or on the first step.

```
1585: terminate called after throwing an instance of 'std::logic_error'
1585:   what():  /global/cfs/cdirs/e3sm/ndk/se02-apr27/components/scream/src/share/atm_process/atmosphere_process.cpp:244: FAIL:
1585: false
1585: Error! Failed pre-condition check (cannot be repaired).
1585:   - Atm process name: Radiation
1585:   - Property check name: sfc_alb_dir_vis within interval [0, 1]
1585:   - Atmosphere process MPI Rank: 1585
1585:   - Message: FieldWithinIntervalCheck failed; min = 1.000000; max = 1.000000
```

PeterCaldwell:
This is clearly a roundoff error: sfc_alb_dir_vis should be between 0 and 1 and is... 1. A quick fix would be to change false to true (i.e. repair out-of-bounds values rather than crash) in https://github.com/E3SM-Project/scream/blob/master/components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp#L283 . You might have to do this for other sfc_albedo quantities too. A better(?) fix might be to increase the bounds on that line of code from 0 to 1 to an upper bound of 1+eps for small eps.

PeterCaldwell:
@brhillman - do you have an opinion about whether repairing, setting bigger bounds, or something else should be our permanent fix?

PeterCaldwell:
Note this is kinda the same issue as #1575

ndkeen:
Yes, I saw that it was maybe the same thing as 1575, but this is without SPA and run the same way as the intel attempt https://github.com/E3SM-Project/scream/issues/1591

PeterCaldwell:
Yeah, it's fine that they are separate issues, but the obvious solution in this issue is to allow repair since errors are just roundoff... but #1575 shows that doing so is dangerous because we *also* have cases where albedo is way outside the reasonable range. 

I'm beginning to wonder whether some of our surface components -> atm component mapping files are screwed up...

ndkeen:
Can close this specific issue with #1596

Issue #1575, opened 2022-04-28
Fail precondition check for sfc_alb_dir_vis with ne120


ndkeen:
With April 27th master, tried a ne120 cime case and see this error after 1st step on PM:

```
  1: terminate called after throwing an instance of 'std::logic_error'
  1:   what():  /global/cfs/cdirs/e3sm/ndk/se02-apr27/components/scream/src/share/atm_process/atmosphere_process.cpp:244: FAIL:
  1: false
  1: Error! Failed pre-condition check (cannot be repaired).
  1:   - Atm process name: Radiation
  1:   - Property check name: sfc_alb_dir_vis within interval [0, 1]
  1:   - Atmosphere process MPI Rank: 1
  1:   - Message: FieldWithinIntervalCheck failed; min = 0.700549; max = 1.254542
```

Will run again to try getting snapshot after 1st step.

PeterCaldwell:
Ok, it's great that the condition checks spit out the max and min values now. I was hoping to see that we were just roundoff-level outside of bounds but an albedo of 1.25 is far from roundoff. Yeah, it would be useful to see the output after step 1: I'm curious whether this excessive value is a single location or widespread.

@brhillman - do you know if albedo is taken directly from the surface schemes (i.e. is this not our fault)?

Also, do we know if EAM was employing limiters which just forced albedo into the acceptable range?

ndkeen:
Currently unable to write snapshots for ne120 (ne30 ok) due to 
```
0: PIO: FATAL ERROR: Aborting... FATAL ERROR: NetCDF: One or more variable 
  sizes violate format constraints 
  (file = SCREAMv1_output_hydrostatic.INSTANT.Steps_x1.np128.0001-01-01.000230.nc) (/global/cfs/cdirs/e3sm/ndk/se02-apr27/externals/scorpio/src/clib/pioc_support.c: 3921)
```
which I thought was something we already fixed

brhillman:
We do clip in EAM, but the tolerance is set to `0.01`:
```
      call handle_error(clip_values( &
         albedo_dir, 0._r8, 1._r8, trim(subname) // ': albedo_dir', tolerance=0.01_r8) &
      )
      call handle_error(clip_values( &
         albedo_dif, 0._r8, 1._r8, trim(subname) // ': albedo_dif', tolerance=0.01_r8) &
      )
```
[...]

Issue #1597, opened 2022-05-02
Albedo out of bounds due to remapping from surface components


PeterCaldwell:
We're struggling with model fails due to albedo>1. Sometimes they are roundoff level, but they've been as big as 1.2. If errors were at roundoff level, we could just correct the problem... but that would hide the fact that sometimes something more evil seems to be happening.

Wuyin says albedo was well behaved in a run with ocean and atm on ne4 grid but out of bounds when atm=ne4 and ocn=oqu240. It seems unlikely that the physical calculation of albedo wouldn't be clipped at 1, so we suspect the problem is with regridding. Have we ever seen these albedo-out-of-bounds errors in a case where ocean was on the same grid as atm? I think we're mainly seeing them in ne120 runs, where we are using a different ocean grid because we don't have the right inputdata...

Open questions:
1. do we have any examples where this is happening where atm2ocn is the identity grid? Or anything other than ne120?
2. what map are we using in these cases that fail? for ne120, it looks like we're using map_ne120np4_to_oRRS18to6v3_aave.170111.nc or map_ne120np4_to_oRRS18to6v3_conserve.170111.nc
3. Do we expect these maps to be conservative? I don't know their details...

wlin7:
The aave maps are conservative. 
Just did a mappping test of the map files used in the ne120 run (grid ne120_r0125_oRRS18to6v3).

With unity values everywhere on oRRS18to6v3, the remapped values on ne120 grids using map_oRRS18to6v3_to_ne120np4_aave.170111.nc never exceed 1.0.

Similarly with unity values on r0125 land grids, using map file map_r0125_to_ne120np4_mono.200508.nc, the remapped data on ne120 grids has a max of 1.000000238. 

The large albedo values as large as 1.2+ do not appear to be due to map files.

PeterCaldwell:
So you're saying that mono maps *do* create spurious maxima though? Was the 1.000000238 generated by remapping a file that was one everywhere on the r0125 grid? I would have thought that you'd need gradients in the source grid to have overshoots on the destination grid.

wlin7:
Yeah, the source data is 1.0 everywhere on r0125 grid. Don't have an explanation why overshoot happens. This is from finer to coarser grid remapping.  A remapping from ne120 to r0125, using the corresponding map file map_ne120np4_to_r0125_mono.20200508.nc, no overshooting occurs.

PeterCaldwell:
[...]

PR #1600, opened 2022-05-03
Adjust rrtmgp albedo checks to repair albedo's out of range.

AaronDonahue:
This commit allows rrtmgp to repair albedos that are out of the 0-1
range.  It is a temporary fix to an issue we are seeing with ne120 runs.
When surface coupling becomes it's own atmosphere process the appropriate
place to do the repair would be as a post-condition check there.

This commit includes a WARNING comment alerting users to the above.

Should address #1575 and #1592

### DIFF
--- a/components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
+++ b/components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
@@ -67,10 +67,15 @@ void RRTMGPRadiation::set_grids(const std::shared_ptr<const GridsManager> grids_
   add_field<Required>("p_mid" , scalar3d_layout_mid, Pa, grid->name(), ps);
   add_field<Required>("p_int", scalar3d_layout_int, Pa, grid->name(), ps);
   add_field<Required>("pseudo_density", scalar3d_layout_mid, Pa, grid->name(), ps);
-  add_field<Required>("sfc_alb_dir_vis", scalar2d_layout, nondim, grid->name());
-  add_field<Required>("sfc_alb_dir_nir", scalar2d_layout, nondim, grid->name());
-  add_field<Required>("sfc_alb_dif_vis", scalar2d_layout, nondim, grid->name());
-  add_field<Required>("sfc_alb_dif_nir", scalar2d_layout, nondim, grid->name());
+  // WARNING! We are switch surface albedo variables to "Updated" to allow rrtmgp
+  // property checks to repair them as a precondition check.  TODO: Change back
+  // to "Required" when surface coupling is it's own process and can handle repair
+  // locally.
+  add_field<Updated>("sfc_alb_dir_vis", scalar2d_layout, nondim, grid->name());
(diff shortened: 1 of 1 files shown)


brhillman:
Do we know why albedo goes out of range yet? It goes WELL out of range for ne120, which makes me think maybe we should understand why before just truncating it.

PeterCaldwell (dismissed):

PeterCaldwell:
> Do we know why albedo goes out of range yet? It goes WELL out of range for ne120, which makes me think maybe we should understand why before just truncating it.

No, we're still trying to figure out why albedo goes out of range. But I don't want Noel to be stuck while we figure that out. So this is a temporary fix to allow him to get further while we push on #1597

components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
```diff
@@ -67,10 +67,15 @@ void RRTMGPRadiation::set_grids(const std::shared_ptr<const GridsManager> grids_
   add_field<Required>("p_mid" , scalar3d_layout_mid, Pa, grid->name(), ps);
   add_field<Required>("p_int", scalar3d_layout_int, Pa, grid->name(), ps);
   add_field<Required>("pseudo_density", scalar3d_layout_mid, Pa, grid->name(), ps);
-  add_field<Required>("sfc_alb_dir_vis", scalar2d_layout, nondim, grid->name());
-  add_field<Required>("sfc_alb_dir_nir", scalar2d_layout, nondim, grid->name());
-  add_field<Required>("sfc_alb_dif_vis", scalar2d_layout, nondim, grid->name());
-  add_field<Required>("sfc_alb_dif_nir", scalar2d_layout, nondim, grid->name());
+  // WARNING! We are switch surface albedo variables to "Updated" to allow rrtmgp
```
bartgol, line 70: Imho, we should just add FPC's in the SC class (it's not complicated). This is a hack that I'm afraid might be forgotten, and left there...

AaronDonahue: I'll let @PeterCaldwell chime in.  We decided this route to make for the quickest implementation so that Noel could start using it right away.  It would also be simple to undo, hence the WARNING in the code.  My hope is that this comment is direct enough that we won't see it and move on (like so much EAM code), and that we will revert when we make SFC CPL an atmosphere process.  I would like to think that this will be on our minds when we do eventually get the SFC CPL as ATM PROC PR in.

bartgol: We should keep an issue or a confluence page, tracking all the "temporary hacks" we put in the code for the sake of getting stuff to run. I am always wary of "we'll fix this later" promises. I've been making that kind of promise myself in the past... ;)

PeterCaldwell (approved):
Let's merge this for the short term and undo it upon conclusion of the investigation into albedos.

AaronDonahue:
@bartgol , I added issue #1605 to track this and added it to the SCREAMv1 and SCREAMv1.1 projects so we don't lose sight of it as a priority.

PR #1602, opened 2022-05-04
Fixing SCREAMv1 restart run from cime

wlin7:
This is just to initiate the PR and discussion towards fixing screamv1 restart issue. Changes will be evolving

Summary of the problems identified:

1. The logic for deciding has_restart_data is not suitable for history output restart. Because of the use of OR (||) operator (frequency > 1), has_restart_data would almost always be true even when m_avg_typ == INSTANT. When history output averaging typs is INSTANT, no history output restart is needed.
2. Even when Averaging Type is 'Average' -- set in scream_output.yaml, no history output restart file (.rhist) is generated when hitting the restart step. 1&2 cause restart run to be aborted due to failing to find the .rhist file -- unnecessary in the case of m_avg_type == INSTANT and missing in the case of m_avg_type == AVERAGE
3. When forcing to skip reading history output restart (set "Perform Restart" to false as default), the restart run would proceed but then fail to pass the T_mid bound check in Dynamics (T_mid below lower bound, as low as 104 when running on ne4 grid)


components/scream/src/share/io/scream_output_manager.cpp
```diff
@@ -90,7 +90,7 @@ setup (const ekat::Comm& io_comm, const ekat::ParameterList& params,
     }
   }
 
-  const auto has_restart_data = (m_avg_type!=OutputAvgType::Instant || m_output_control.frequency>1);
+  const auto has_restart_data = (m_avg_type!=OutputAvgType::Instant && m_output_control.frequency>1);
```
wlin7, line 93: has_restart_data is used for both Checkpoint_control and history output_control. Not sure if the original logic is necessary for Checkpoint. It definitely needs to be changed for history output control. And for history output restart, the 2nd condition is not necessary, and  inappropriate if "Frequency Units" is not 'Steps".

PeterCaldwell: I don't really know what any of these variables mean, but I conjecture that the intent was that we should require the existence of a history restart file if output isn't instantaneous and if we're not restarting at the start of an averaging interval for output.

PeterCaldwell:
> Even when Averaging Type is 'Average' -- set in scream_output.yaml, no history output restart file (.rhist) is generated

@AaronDonahue should check into this in case the fix is trivial, though I think it's fine if we can only write instantaneous output in the v1 release. Which I guess means that a reasonable solution for parts 1 and 2 of PR would be "comment out the check for restart history files"

AaronDonahue:
> > Even when Averaging Type is 'Average' -- set in scream_output.yaml, no history output restart file (.rhist) is generated
> 
> @AaronDonahue should check into this in case the fix is trivial, though I think it's fine if we can only write instantaneous output in the v1 release. Which I guess means that a reasonable solution for parts 1 and 2 of PR would be "comment out the check for restart history files"

Let me check this.  We should be able to write `average`, `min` and `max` output and we actually check that these work with restart history files in the IO unit tests.

bartgol:
> We should be able to write `average`, `min` and `max` output and we actually check that these work with restart history files in the IO unit tests.

We only check restart for "Average". The other averaging type are not tested in the output restart test.

wlin7:
> > > the restart run would proceed but then fail to pass the T_mid bound check in Dynamics (T_mid below lower bound, as low as 104 when running on ne4 grid)
> > 
> > 
> > This is the real crux of this PR. I thought we had unit tests ensuring that writing/reading restarts was BFB?
> 
> That is a good point. We do test history restart, but we only test 'Average' averaging type. We could add a test for a history restart of 'Instant' averaging type (and maybe another with 'Average' averaging type, with initial run stopping exactly at an output step, so that no hist restart is generated), to make sure we don't attempt to read hist restart when not needed.

wlin7:
> > > the restart run would proceed but then fail to pass the T_mid bound check in Dynamics (T_mid below lower bound, as low as 104 when running on ne4 grid)
> > 
> > 
> > This is the real crux of this PR. I thought we had unit tests ensuring that writing/reading restarts was BFB?
> 
> That is a good point. We do test history restart, but we only test 'Average' averaging type. 

Does setting "Perform Restart" to false has spillover effect on restarting model state? I changed the default to false in the following statement in scream_output_manager.cpp.

`bool perform_history_restart = has_restart_data && restart_pl.get("Perform Restart",false);`

The inline comments say ok to skip reading history restart, and I assumed it only impacted the creation of ensuing new history output file.

bartgol:
> > > > the restart run would proceed but then fail to pass the T_mid bound check in Dynamics (T_mid below lower bound, as low as 104 when running on ne4 grid)
> > > 
> > > 
> > > This is the real crux of this PR. I thought we had unit tests ensuring that writing/reading restarts was BFB?
> > 
> > 
> > That is a good point. We do test history restart, but we only test 'Average' averaging type.
> 
> Does setting "Perform Restart" to false has spillover effect on restarting model state? I changed the default to false in the following statement in scream_output_manager.cpp.
> 
> `bool perform_history_restart = has_restart_data && restart_pl.get("Perform Restart",false);`
> 
> The inline comments say ok to skip reading history restart, and I assumed it only impacted the creation of ensuing new history output file.

`Perform Restart` does not impact whether _future_ history restart files will be written. It only allows to forget about the fact that, say, we were in the middle of an 'average' step during the previous run, and start with a fresh new average.

This parameter only impacts the restart of the model output (on a per-output-file-basis). It does not impact the restart of the model itself.

bartgol:
> BTW, @bartgol , where/how to set "Perform Restart" in runtime namelist files? I tried it in namelist_scream.xml (which then went to scream_input.yaml) and scream_output.yaml without success, so I just hardwired it in the code to proceed with the restart run.

You should be able to get a history restart by default. If you don't want it, you can try adding in the proper section of `scream_output.yaml`. I see that there is no 'Restart' sublist in the default one, so you would have to add it (e.g., at the bottom):
```
Restart:
  Perform Restart: false
...
```
Notice that you don't need this if you _do_ want to perform history restart, since that's the default behavior.

Are you saying that modifying `$rundir/data/scream_output.yaml`, say disabling restart, does not cause the history restart to be disabled?

Also, to be clear: setting `Perform Restart: false` in the output yaml file does _not_ affect whether the model is restarted. The model is automatically restarted if the case t0 is "older" than the current run t0.

wlin7:
Thanks. I added it just after the line of "Max Snapshots Per File:" and it didn't take effect. I will try again.

As for the default, it seems to be hardwired to "true" as set in the following line
`bool perform_history_restart = has_restart_data && restart_pl.get("Perform Restart",true);`

bartgol:
> Thanks. I added it just after the line of "Max Snapshots Per File:" and it didn't take effect. I will try again.
 
Ah, yes, that's not the right place. It has to be in the `Restart` sublist.

> As for the default, it seems to be hardwired to "true" as set in the following line `bool perform_history_restart = has_restart_data && restart_pl.get("Perform Restart",true);`

Yes, I assumed that this would be the most common use case. In other words, I assumed it is rare that someone restarts a simulation in the middle of an averaging period without wanting to restart also the averaging process...

bartgol:
@wlin7 I removed the WIP tag, so that we can start testing this PR. Even if you have other things to fix, I believe this bugfix is already worth testing and integrating.

bartgol (dismissed):

PeterCaldwell:
To be clear, my understanding is that this PR only works if time is specified in nsteps rather than days, months, etc. I think that's fine for EAMv1 because other aspects of the code also only work for nsteps right now.

bartgol:
I now recall why I had `||` instead of `&&`. The issue is the counter of steps since the last output. Say frequency=2, and you write a restart after 1 timestep. Upon restart, we need the output stream to know that it doesn't have to wait 2 steps before output, but only 1, since it was already in the middle of an output interval.

That said, I think the mod in this PR is ok. I just need to add some logic to communicate the step information to the output. Testing the fix now.

bartgol:
Ok, the last mod _should_ fix the restart. To summarize

- the current failure in the AT is due to the fact that the counter `nsteps_since_last_write` (also called `avg_count` in the output classes) was not restarted.
- since INSTANT output does not generate any rhist file, we switched to restart `avg_count` based on the timestamp nstep counter and out freq (via mod arithmetic).
- since we don't really want to allow changing out freq through restarts (e.g., out freq=10, but then restart with out freq=3, would lead to funky behavior with avg_type=AVERAGE), I decided to not store avg_count in the file, and simply deduce it from run_t0, as described at the previous point.


These mods should fix the first two issues described by @wlin7 . I am not sure about the third. I'm checking standalone tests now (still one fail, in io unit tests). But I was thinking to merge, and fix remaining restart issues for CIME runs in #1619.

bartgol:
@AaronDonahue @PeterCaldwell since I also contributed to the PR, it would be more appropriate if someone else reviewed it.

PeterCaldwell:
> But I was thinking to merge, and fix remaining restart issues for CIME runs in https://github.com/E3SM-Project/scream/pull/1619. 

Yeah, we should definitely merge this without fixing the 3rd issue, which I think needs some digging still. I don't think issue 3 will be fixed in #1619 , but that's fine.

wlin7:
Continue digging on the restart problem #3: failure in atm dynamics process, first caught during bound checks for T_mid (being too cold).   Comparison between startup and restart runs show both runs read the same variables that are on physics GLL grids. The restart run also read several variables on dyn grids: v_dyn, vtheta_dp_dyn, dp3d_dyn, etc.. It is these variables that appears suspicious.

I am testing with ne4_ne4 grid. These dyn-grid variables in the restart file only has 4 spectral elements, while 96 are expected for ne4. A sample restart file on nersc at 
`/global/homes/w/wlin/gscrc/e3sm_scratch/cori-knl/20220502.F2000SCREAMv1.ne4_ne4/run/20220502.F2000SCREAMv1.ne4_ne4.INSTANT.[Steps_x5.0001-01-01.002500.r.nc](http://steps_x5.0001-01-01.002500.r.nc`

The dyn-grid variable has dimensions like: `double v_dyn(time, elem, dim2, gp, gp, lev)`, where in this ne4 restart file, `elem=4, and gp=4`

If elem=4 is indeed wrong, it would explain that the data in the restart file are insufficient to be assigned to the corresponding arrays at runtime, causing each of these variables to be partially improperly initialized and the calculations to go awry. Confirmation needed from our experts.

Then the related questions: if the correct elem should be 96 in this ne4 case, how it ends up with elem=4 in the restart file? And if the ne4 model need to read data size of 96 spectral elements, why it didn't complain when only get data of 4 elements?

I suspect the dyn-grid variables are only for one of the MPI rank. Note that the ne4_ne4 test was run with 24 MPI tasks.

wlin7:
Running with just one mpi task, the restart run went well; and the simulation results "appear" BFB. With initial run of 9 steps, restart check point at 5th step, and restart run for 4 steps, cpl.hi at step 9 are BFB. INSTANT history output for the overlapping  4 steps are BFB except for the following three variables (and the maximum absolute difference):  
 ```
   micro_liq_ice_exchange   (4.1827053e-12)
    micro_vap_liq_exchange  (2.9696711e-11)
    micro_vap_ice_exchange   (6.1400191e-11)
```
Another issue with history output file: the history output during the restart run did not go to the one created during the startup run that has not reached `Max Snapshot Per File`. A new file was created to output the new history snapshots. To maintain consistency and in some case for the ease of post-processing, the restart run  should continue to write to the latest history output file (of the same type, in terms of fields saved, output frequency, averaging type) until reaching `Max Snapshot Per File`. But we can live with it for now.

**Comment by bartgol** [comment]
Nicely done @wlin7 ! That was very good debugging!

Yes, we only test model restart with 1 rank in standalone mode. What you found strongly suggests to test also with 2+ ranks. I'm going to check why dyn fields are written with the wrong dimensions.

**Comment by PeterCaldwell** [comment]
Yeah, great work Wuyin!!!

**Comment by wlin7** [comment]
Hi @bartgol , agreed restart test with 2+ rank is also needed. I will leave it to you to fix it. Does the rank 0 gather the data then write, or all rank writes to the same file via PIO. I would assume the latter. And for the restart run that didn't complain about data size being read, I guess it is also at per rank level thus data size read meets the required.

**Comment by bartgol** [comment]
> Does the rank 0 gather the data then write, or all rank writes to the same file via PIO. I would assume the latter.
Yes, we use PIO.

> And for the restart run that didn't complain about data size being read, I guess it is also at per rank level thus data size read meets the required.
Yes, all ranks have the same number of elements, so I'm guessing they all end up reading the first $nelem/$nranks ones.

It should be an easy fix. Might require a few lines of code, but not conceptually complicated.

**Comment by bartgol** [comment]
I am working on the fixes. Do you guys want to merge this PR for now, since it does address some issues already? Or do you want to wait for all restart issues to be fixed? @AaronDonahue @PeterCaldwell @wlin7

**Comment by wlin7** [comment]
@bartgol , we can merge this one now. This PR was started to facilitate discussion towards fixing restart issues. Including the fix you are working on, we will need some more fixes to make restart works as designed; no need to wait and put all of them together.

BTW, your changes in this PR, the logics related to perform_restart and has_restart_data look good to me. I am not able to add myself as a reviewer to formally approve it, though.

**Comment by PeterCaldwell** [comment]
Yeah, I prefer to integrate this PR sooner and have a separate PR for the MPI dynamics variables bugfix. I like smaller PRs.

File: components/scream/src/share/io/scream_output_manager.cpp [hunk 2]
```diff
@@ -111,30 +111,38 @@ setup (const ekat::Comm& io_comm, const ekat::ParameterList& params,
...
+  // our averaging in a "restart" file (e.g., the current avg).
   // Note: the user might decide *not* to restart the output, so give the option
   //       of disabling the restart. Also, the user might want to change the
   //       casename, so allow to specify a different casename for the restart file.
   if (m_is_restarted_run) {
     // Allow to skip history restart, or to specify a casename for the restart file
     // that is different from the casename of the current output.
     auto& restart_pl = m_params.sublist("Restart");
-    bool perform_history_restart = has_restart_data && restart_pl.get("Perform Restart",true);
+    bool perform_history_restart = restart_pl.get("Perform Restart",true);
```
AaronDonahue, line 122: What if we don't have history restart data?  For example, a typical restart run in E3SM would be with monthly restarts and monthly average data.  Assuming that our run finishes normally, we wouldn't need to restart the history file since we would be starting a new month.  Doesn't this check make sure we don't throw an error in this case?

AaronDonahue:
@wlin7 , I just have one question about one change before I can approve.

components/scream/src/share/io/scream_output_manager.cpp
wlin7: This PR only fixes the issue when avg_type for history output is Instant. Fixes for multiple places will be needed for the other avg_type or output frequency units (such as monthly).

wlin7:
> @wlin7 , I just have one question about one change before I can approve.

@AaronDonahue , is it the question just above your latest comment? I have replied. We need more fixes to have flexible restart options.

components/scream/src/share/io/scream_output_manager.cpp
bartgol: Good point. We can check also the value of `m_output_control.nsteps_since_last_write` a few lines below before attempting to read data.

bartgol: @AaronDonahue The commit I just pushed should fix this.

AaronDonahue (approved):

PR #1604, opened 2022-05-04
Allow unexpected resolutions and add ne120 defaults in XML

AaronDonahue:
This commit adds the default settings for running an ne120np4 CIME based SCREAMv1 case.

It also updates the filename for the initial condition file for ne30.

### DIFF
--- a/components/scream/cime_config/namelist_defaults_scream.xml
+++ b/components/scream/cime_config/namelist_defaults_scream.xml
@@ -153,9 +153,11 @@ tweaking their $case/namelist_scream.xml file.
     <!-- Homme dynamics -->
     <homme inherit="atm_proc_base">
       <Grid>Dynamics</Grid>
+      <Vertical__Coordinate__Filename>UNSET</Vertical__Coordinate__Filename>
       <Vertical__Coordinate__Filename hgrid="ne4np4">${DIN_LOC_ROOT}/atm/scream/init/init_ne4np4.nc</Vertical__Coordinate__Filename>
       <Vertical__Coordinate__Filename hgrid="ne30np4">${DIN_LOC_ROOT}/atm/scream/init/screami_ne30np4L72_20220503.nc</Vertical__Coordinate__Filename>
       <Vertical__Coordinate__Filename hgrid="ne120np4">${DIN_LOC_ROOT}/atm/scream/init/screami_ne120np4L72_20220503.nc</Vertical__Coordinate__Filename>
+      <Vertical__Coordinate__Filename hgrid="ne512np4">${DIN_LOC_ROOT}/atm/scream/init/screami_ne512np4L128_20220505.nc</Vertical__Coordinate__Filename>
       <Vertical__Coordinate__Filename hgrid="ne4np4" COMPSET=".*SCREAM%AQUA.*">${DIN_LOC_ROOT}/atm/scream/init/scream-aquaplanet_init_ne4np4_L72_20211202.nc</Vertical__Coordinate__Filename>
       <Moisture>moist</Moisture>
     </homme>
@@ -171,9 +173,11 @@ tweaking their $case/namelist_scream.xml file.
 
     <!-- Simple Prescribed Aerosols (SPA) -->
     <spa inherit="physics_proc_base">
+      <SPA__Remap__File>UNSET</SPA__Remap__File>
       <SPA__Remap__File hgrid="ne4np4">${DIN_LOC_ROOT}/atm/scream/init/map_ne30_to_ne4_mono_20220502.nc</SPA__Remap__File>
       <SPA__Remap__File hgrid="ne30np4">none</SPA__Remap__File>
       <SPA__Remap__File hgrid="ne120np4">${DIN_LOC_ROOT}/atm/scream/init/map_ne30np4_to_ne120np4_mono_20220502.nc</SPA__Remap__File>
+      <SPA__Remap__File hgrid="ne512np4">UNSET</SPA__Remap__File>
       <SPA__Data__File>${DIN_LOC_ROOT}/atm/scream/init/spa_file_unified_and_complete_ne30_20220428.nc</SPA__Data__File>
     </spa>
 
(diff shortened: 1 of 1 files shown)


AaronDonahue:
I'm opening this PR now and marking it as WIP since it will conflict with other PR's that will soon be merged.  I wanted to open it early so I can ask HOMME folks to comment on HOMME namelist options.  As far as I can tell there are only a few minor changes needed to accommodate the ne120 resolution simulations.  @mt5555  or @oksanaguba , can one of you confirm?

PeterCaldwell:
@wlin7 - I think we should change this PR to use ne120_ne120. I don't quite understand what needs to be done to make that work... Can you comment? I guess this PR would depend on your upstream https://github.com/E3SM-Project/E3SM/pull/4932 ?

wlin7:
> @wlin7 - I think we should change this PR to use ne120_ne120. I don't quite understand what needs to be done to make that work... Can you comment? I guess this PR would depend on your upstream [E3SM-Project/E3SM#4932](https://github.com/E3SM-Project/E3SM/pull/4932) ?

@PeterCaldwell , this PR is for generic ne120np4 atm grid and does not depend on my PR. When ne120_ne120 is used for a case, the setting in the [upstream #4932](https://github.com/E3SM-Project/E3SM/pull/4932) will kick in to complete the setup for land. This settings in this PR for scream atm will be equally applicable when running on ne120_oRRS18to6v3 or tri-grid -- though tri-grid has its own potential issue.

PeterCaldwell:
> this PR is for generic ne120np4 atm grid and does not depend on my PR

Ah yes, that makes sense. Thanks.

AaronDonahue:
I'm retesting this PR, it should pass.  It will impact ne30 runs too because it updates the ne30 input files.  But for the most part it just adds ne120 defaults so we can compile that case out of the box.  @PeterCaldwell, @crterai  and @ndkeen , I listed you as reviewers because I think you are the three who are most familiar with setting up and running an ne120 case with SCREAMv1.  Let me know if I missed anything important.

components/scream/cime_config/namelist_defaults_scream.xml
```diff
@@ -238,7 +247,8 @@ tweaking their $case/namelist_scream.xml file.
     ${DIN_LOC_ROOT}/atm/scream/tables/vn_table_vals.dat8,
     ${DIN_LOC_ROOT}/atm/scream/tables/vm_table_vals.dat8,
     ${DIN_LOC_ROOT}/atm/scream/init/spa_file_unified_and_complete_ne30_20220428.nc,
-    ${DIN_LOC_ROOT}/atm/scream/init/map_ne30_to_ne4_mono_20220502.nc
+    ${DIN_LOC_ROOT}/atm/scream/init/map_ne30_to_ne4_mono_20220502.nc,
+    ${DIN_LOC_ROOT}/atm/scream/init/map_ne30np4_to_ne120np4_mono_20220502.nc
```
PeterCaldwell, line 251: Do we want to include IC files in this list of things to download? Note that trying to download all files for all resolutions won't be a workable long-term strategy (but is probably ok for v1).

bartgol: Yes, that's a good point. We should use cime vars to download just _one_ IC file, depending on the grid resolution.

PeterCaldwell: Well, right now we're not downloading any _IC_ file, but we are downloading all _remap_ files. I'm inclined to have folks manually download both IC files and mapping files for SCREAMv1 since we're all power users and aren't running on that many machines.

PeterCaldwell:
@AaronDonahue - I've made a bunch of changes to your PR. Could you check that they're ok?

Also - I'm pretty sure ne512 files aren't on blues yet. I guess we should put them there (though I don't think anyone has tried the remap file yet).

components/scream/cime_config/namelist_defaults_scream.xml
```diff
@@ -248,7 +248,12 @@ tweaking their $case/namelist_scream.xml file.
     ${DIN_LOC_ROOT}/atm/scream/tables/vm_table_vals.dat8,
     ${DIN_LOC_ROOT}/atm/scream/init/spa_file_unified_and_complete_ne30_20220428.nc,
     ${DIN_LOC_ROOT}/atm/scream/init/map_ne30_to_ne4_mono_20220502.nc,
-    ${DIN_LOC_ROOT}/atm/scream/init/map_ne30np4_to_ne120np4_mono_20220502.nc
+    <!-- ${DIN_LOC_ROOT}/atm/scream/init/map_ne30np4_to_ne120np4_mono_20220502.nc -->
```
AaronDonahue, line 251: Do we just not need these files to be downloaded anymore?

components/scream/cime_config/namelist_defaults_scream.xml
```diff
@@ -238,7 +247,13 @@ tweaking their $case/namelist_scream.xml file.
     ${DIN_LOC_ROOT}/atm/scream/tables/vn_table_vals.dat8,
     ${DIN_LOC_ROOT}/atm/scream/tables/vm_table_vals.dat8,
     ${DIN_LOC_ROOT}/atm/scream/init/spa_file_unified_and_complete_ne30_20220428.nc,
-    ${DIN_LOC_ROOT}/atm/scream/init/map_ne30_to_ne4_mono_20220502.nc
+    ${DIN_LOC_ROOT}/atm/scream/init/map_ne30_to_ne4_mono_20220502.nc,
+    <!-- ${DIN_LOC_ROOT}/atm/scream/init/map_ne30np4_to_ne120np4_mono_20220502.nc -->
+    <!-- ${DIN_LOC_ROOT}/atm/scream/init/map_ne30np4_to_ne512np4_mono_20220506.nc -->
+    ${DIN_LOC_ROOT}/atm/scream/init/init_ne4np4.nc
```
AaronDonahue, line 253: This will need a comma at the end, otherwise it throws an obscure error during the `case.submit` command that is hard to track.

bartgol: Are you sure? Current master does not have a comma after the last input file, and it seems to run just fine.

components/scream/cime_config/namelist_defaults_scream.xml
PeterCaldwell: I noticed that we were downloading remap files for all resolutions in order to run at any particular resolution, but weren't downloading any initial condition at all. I decided to set us up to download all data for all resolutions, then commented out the most biggest resolutions because I worry that someone might want to run ne4 and ned up waiting hours for ne512 IC to download... What I was hoping for was to have all the files show up in namelist_scream.xml, but to have ne120 and ne512 be commented out. Then folks can uncomment/download them as needed. Of course, it would be better to only download the files we need, but that's a v1.1 task.

AaronDonahue:
Considering the discussion in #1614 should we add a trimmed set of "default outputs" to this PR?  @PeterCaldwell ?  Also, I think this might be nearly ready to review/merge.  I just had one fix needed.  Which I can do too.

components/scream/cime_config/namelist_defaults_scream.xml
AaronDonahue: I'm okay with that.  super important to make sure that every item is trailed by a comma.  Maybe we should make sure that the last item in the entry is something we will for sure keep uncommented in the list and have it be the only thing without a comma.  That comma is subtle and tricky.

components/scream/cime_config/namelist_defaults_scream.xml
AaronDonahue: This is the second-to-last entry.

bartgol: Ah, I cannot read.

jgfouca:
@AaronDonahue , I can regenerate baselines with current master. I assume this is for screamV1

AaronDonahue:
@jgfouca , yes, for SCREAMv1

PeterCaldwell (approved):

jgfouca:
@AaronDonahue , ok, the baselines should be up to date now.

PeterCaldwell (approved):

ndkeen:
will se_tstep need to be `real` instead of `integer`?

bartgol:
Yes. Why do you say "instead of integer" though? Are we assuming it is an int somewhere?

PR #1612, opened 2022-05-05
Add a diagnostic for every common physics function defined to date.

AaronDonahue:
This commit will add diagnostics for each common physics function.  The total list is as follows:

AtmosphereDensity: The density of the atmosphere (3D)
DryStaticEnergy: The dry static energy, as used by SHOC (3D)
SeaLevelPressure: The sea level pressure, as used by surface coupling (2D)
VerticalLayerInterface: The z-height for the interfaces of each vertical layer (3D)
        ""           Midpoint: The z-height for the midpoints of each vertical layer (3D)
        ""           Thickness: The dz vertical layer thickness (3D)
VirtualTemperature: The virtual temperature (3D)
Exner: The exners function (3D)


AaronDonahue:
Note, I haven't created unit tests for all the new diagnostics yet.  I've just verified by looking at output from a physics only stand alone test case.  Unit tests will be needed eventually, but since this isn't as high a priority I'm not sure when I will get to them.  In the meantime, eval folks can access this branch if they want to produce output like z-coordinate height, density, potential temperature, dry-static-energy, etc.

components/scream/src/diagnostics/atm_density.hpp
```diff
@@ -0,0 +1,104 @@
...
+    static std::set<std::string> s;
+    s.insert(m_params.get<std::string>("Grid"));
+    return s;
+  }
+
+  // Set the grid
+  void set_grids (const std::shared_ptr<const GridsManager> grids_manager);
+
+  // Actual diagnostic calculation 
+  struct run_diagnostic_impl {
```
bartgol, line 47: This regards all the diagnostics in the PR.

Is this inner functor class necessary? It seems to add an unnecessary layer of complexity. Can't you just have the operator() in the AtmDensityDiagnostic class?

components/scream/src/diagnostics/diagnostics_interface.hpp
```diff
@@ -0,0 +1,42 @@
...
+
+inline void register_diagnostics () {
+  // Note, we use the atmosphere process factory to register diagnostics
+  // because each diagnostic is a sub-class of an atmopshere process
+  auto& proc_factory = AtmosphereProcessFactory::instance();
+ 
+  proc_factory.register_product("Potential Temperature",&create_atmosphere_process<PotentialTemperatureDiagnostic>); 
+}
+
+class DiagnosticInterface
```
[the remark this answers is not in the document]
AaronDonahue: Good catch, I thought I had deleted these.  They must have somehow still got added when using `git add *` on my part.  I'll remove them.

AaronDonahue:
This PR will include the changes to the potential temperature diagnostic from #1656 , so if this one goes in we can change 1656 to just focus on the fix to how restarts are handled.

AaronDonahue:
I suspect that the DIFFs we are seeing on mappy are because the baselines have not been updated since the this morning, when a separate PR with DIFFs was merged.  @bartgol  can you review this PR so that it is ready to merge after the baselines are set?

components/scream/src/diagnostics/tests/atm_density_test.cpp
```diff
@@ -0,0 +1,187 @@
...
+      ekat::genRandArray(dview_as_real(pseudodensity), engine, pdf_pseudodens);
+      ekat::genRandArray(dview_as_real(pressure),      engine, pdf_pres);
+      ekat::genRandArray(dview_as_real(watervapor),    engine, pdf_qv);
+      Kokkos::deep_copy(T_sub,temperature);
+      Kokkos::deep_copy(pseudo_sub,pseudodensity);
+      Kokkos::deep_copy(p_sub,pressure);
+      Kokkos::deep_copy(qv_sub,watervapor);
+    } 
+
+    // Run diagnostic and compare with manual calculation
```
bartgol, line 142: I am not sure I like this kind of test. It's basically a lot of boiler plate code, and the important part is kind of a copy paste of the same code in the diagnostic run method. It's somewhat of a burden in terms of lines of code to maintain vs lines of code tested.

Alas, I'm not really sure how to test all these diagnostics classes, which do come down to a lot of boilerplate code that calls a 5 lines function...

bartgol (approved):
Seems ok. After a couple of files, they all started to look the same, so I hope I didn't miss anything big. I only saw a leftover debug comment.

I'm a bit unhappy with tests that are basically a copy+paste of the code inside another class/function, and compare the results. But I can't come up with anything more meaningful.

components/scream/src/diagnostics/tests/atm_density_test.cpp
AaronDonahue: I agree, my hope is that we won't have to update all the tests at once in the future.  So the repeated boilerplate code won't be too much of a burden.  I also struggled with how to best test these new diagnostics.  I think they are a special case though, since they are just output from defined physics functions, i.e. they are not new functions themselves.  Since the common physics functions are, themselves, robustly tested anyway I didn't see a need to replicate those tests here.  Instead, if the diagnostic produces the same output as the common physics function does then the diagnostic is doing exactly what we want it too.

Other diagnostics should be more rigorously unit tested.

bartgol (approved):

PR #1613, opened 2022-05-05
Add resolution-dep memory usage to atm log

bartgol:
Account for:
- field manager variables (without metadata)
- grids data (dofs, lat, lon,...)
- atm scratch buffer
- scorpio output views (for averaging or dealing with padding).


bartgol:
Turns out, the failures I was seeing were only due to a corrupted input file on the ANL server (which we should fix). Letting the AT run its course now.

bartgol:
> > Turns out, the failures I was seeing were only due to a corrupted input file on the ANL server (which we should fix). Letting the AT run its course now.
> 
> So this still needs to be fixed, right? By replacing the spa_data_for_testing.nc file on the blues/ANL server with a more modern version (from NERSC?). With a datestamp in the name and accompanying PR to read the new file? I can do that if so.

The PR per se doesn't have to be fixed. We need to fix the data on the ANL server. I guess you could also create a new  file `spa_data_for_testing_$TIMESTAMP.nc`, and upload that on one of the servers. But I feel like that's something orthogonal to this PR.

components/scream/src/control/atmosphere_driver.cpp
```diff
@@ -909,12 +909,7 @@ void AtmosphereDriver::initialize_atm_procs ()
   stop_timer("EAMxx::init");
   m_atm_logger->info("[EAMXX] initialize_atm_procs ... done!");
 
-#ifdef SCREAM_HAS_MEMORY_USAGE
-  long long my_mem_usage = get_mem_usage(MB);
-  long long max_mem_usage;
-  m_atm_comm.all_reduce(&my_mem_usage,&max_mem_usage,1,MPI_MAX);
-  m_atm_logger->info("[EAMxx::init] memory usage: " + std::to_string(max_mem_usage) + "MB");
-#endif
+  report_res_dep_memory_foorprint ();
```
AaronDonahue, line 912: Should this be "footprint"?

components/scream/src/control/atmosphere_driver.cpp
```diff
@@ -1079,5 +1074,63 @@ check_ad_status (const int flag, const bool must_be_set)
   }
 }
 
+void AtmosphereDriver::report_res_dep_memory_foorprint () const {
```
AaronDonahue, line 1077: Same here with "foorprint"

AaronDonahue (commented):
Just a few places where "foorprint" is used instead of "footprint" probably want to fix those before merging.

ndkeen (commented):
i was also going to suggest making this into a function. i think a nice generic name is best as we can use this for various things (ie not just memory use). Something like log_scream_diagnostic("cool name");

ndkeen:
No, the string would just be helpful to know where you are (like a timer string name). The function would print a one-line message. For now it can be memory use. At a later date, we could add other diagnostics, such that nothing needs to change in the code.

bartgol:
> No, the string would just be helpful to know where you are (like a timer string name). The function would print a one-line message. For now it can be memory use. At a later date, we could add other diagnostics, such that nothing needs to change in the code.

Well, we can already do that, using the logger. The AD, as well as all atm procs, can access it. The AD can do
```
m_atm_logger->debug ("some debug message")
m_atm_logger->info ("some info message")
m_atm_logger->warn ("some warning message")
```
while atm procs can do
```
log (LogLevel::debug,"my debug stuff");
log (LogLevel::info,"my info stuff");
log (LogLevel::warn,"my warning stuff");
```
The log level of the atm logger (set via `Debug->Atm Log Level`) decides what makes it to the atm.log file (or to screen, for standalone runs). Other log levels are available. In order of increasing importance: `trace`, `debug`, `info`, `warn`, `error`, `critical`. If the logger log level is <= than the requested level, the msg gets logged.

AaronDonahue (approved):

PR #1617, opened 2022-05-06
SCREAM: use same value of PI in SHOC for v1 and v0 code bases

bartgol:
The macro `M_PI` is defined in some C header, so we can't include it in Fortran. The simplest solution is to just spell out pi with enough decimal digits for double precision, and let the compiler truncate it to whatever Real/rtype are, in both C++ and Fortran.


jgfouca (approved):

bartgol:
The fails are expected. Merging.

PR #1620, opened 2022-05-07
Allow changing the list of atm processes in a group

bartgol:
@brhillman This should allow to change the list of atm processes without the need to create an ad-hoc compset.

Removing an atm proc was already possible. What this PR does is to allow to completely change the atm procs list and/or organization.

Note: I am still testing this.

### DIFF
--- a/components/scream/cime_config/buildnml
+++ b/components/scream/cime_config/buildnml
@@ -382,6 +382,9 @@ def _create_raw_xml_file_impl(case, xml):
     atm_procs.tag = "atmosphere_processes"
     xml.append(atm_procs);
 
+    # 5. Keep the atm procs defaults, so one can change atm procs via atmchange
+    xml.append(atm_procs_defaults)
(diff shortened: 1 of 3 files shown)


jgfouca (dismissed):

AaronDonahue:
@bartgol and/or @jgfouca , is this still a functionality we want/need?  I'm seeing conflicts w/ master since it's a few months old.  Should we revive?  Or close?

bartgol:
This PR would easily allow to swap a param for another (e.g., swap in MG2 for P3) or to remove one (e.g., remove SPA) from an existing compset. Or grouping atm procs differently (e.g., to try a || split, which, granted, we still don't have implemented) without the need of a different compset.

I think it might be useful, but I don't run the model, so I don't know if this is really useful or if it's just interesting from an aesthetic point of view... If you think it's not really going to be useful, we can close it.

PR #1628, opened 2022-05-10
Slimmed down output request

PeterCaldwell:
This PR just removes unnecessary output variables from scream_output.yaml.

The one question I have is whether anyone has a recent copy of our DAG, or more generally whether our switch to xmls has blocked our ability to write out the DAG.

### DIFF
--- a/components/scream/data/scream_output.yaml
+++ b/components/scream/data/scream_output.yaml
@@ -27,10 +27,10 @@ Fields:
       - cldfrac_liq
       - cldfrac_ice
       - cldfrac_tot
+      - horiz_winds
       - tke
       - eddy_diff_mom
       - sgs_buoy_flux
-      - inv_qc_relvar
       - pbl_height
       - LW_flux_up
       - LW_flux_dn
--- a/components/scream/data/scream_output.yaml
+++ b/components/scream/data/scream_output.yaml
@@ -7,8 +7,8 @@ Fields:
   Physics GLL:
     Field Names:
       - T_mid
+      - PotentialTemperature
       - qv
-      - T_prev_micro_step
       - qc
       - qr
       - qi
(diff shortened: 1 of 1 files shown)


components/scream/data/scream_output.yaml
```diff
@@ -19,15 +19,15 @@ Fields:
...
+      - nc_activated
+      - ni_activated
+      - nc_nuceat_tend
       - cldfrac_liq
+      - cldfrac_ice
       - cldfrac_tot
+      - tke
       - eddy_diff_mom
       - sgs_buoy_flux
       - inv_qc_relvar
```
crterai, line 33: are we using `inv_qc_relvar` for any of our calculations now? If not, I think we can remove.

PeterCaldwell: Both of your suggestions are great, @crterai . I've just pushed a commit with both of them.

crterai (dismissed):
Looks good to me!

bogensch (dismissed):

AaronDonahue:
@jgfouca , I'm seeing this error a lot, I also got it when grabbing the latest version of master locally:

```
error: The following untracked working tree files would be overwritten by checkout:
	CIME/non_py/externals/genf90/ChangeLog
	CIME/non_py/externals/genf90/genf90.pl
```

I suspect it is related to the upstream merge.  Do we need to clean the repos that the AT works in to stop this error from popping up?  I mention it here because it is in the `mappy` error log.

AaronDonahue:
@jgfouca , nvmd, I see on Slack that @bartgol has already commented about this.

AaronDonahue:
@crterai or @bogensch , can one of you re-review this PR, I had to update it with `master` so that it would pass the AT, which means it needs a new review.

crterai (approved):
Still looks good!

PR #1629, opened 2022-05-10
Use grid-mean condensate for cloud optics

brhillman:
Use grid-mean condensate for cloud optics when not doing subcolumn sampling of cloud state. This adds a new yaml option to RRTMGP, `do_subcol_sampling`, to enable future support for implementing subcolumn sampling and to control whether or not we want to use grid-mean or in-cloud condensate amount to compute cloud optics. When not doing subcolumn sampling, we assume that each column is either entirely cloudy or entirely clear at a given level (i.e., layer cloud fraction is 0 or 1), and thus we want to use the grid-mean (total) cloud condensate amount to compute cloud optical properties. When doing subcolumn sampling of cloud state (not currently supported), we would want to use in-cloud values to compute cloud optical properties before assigning to subcolumns (i.e., using qc / cldfrac). The default for `do_subcol_sampling` is set to `false` for now, since subcolumn sampling is not yet implemented, and the code will throw an error if the user tries to select `do_subcol_sampling = true`. This change is non-BFB, since cloud properties are scaled differently for the computation of cloud optics.

[Non-B4B]

### DIFF
--- a/components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
+++ b/components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
@@ -263,6 +263,10 @@ void RRTMGPRadiation::initialize_impl(const RunType /* run_type */) {
   // Determine whether or not we are using a fixed solar zenith angle (positive value)
   m_fixed_solar_zenith_angle = m_params.get<Real>("Fixed Solar Zenith Angle", -9999);
 
+  // Whether or not to do MCICA subcolumn sampling
+  m_do_subcol_sampling = m_params.get<bool>("do_subcol_sampling",false);
+  EKAT_REQUIRE_MSG(not m_do_subcol_sampling, "Error! RRTMGP does not yet support do_subcol_sampling = true");
+
   // Initialize yakl
   if(!yakl::isInitialized()) { yakl::init(); }
 
@@ -483,7 +487,6 @@ void RRTMGPRadiation::run_impl (const int dt) {
         p_del(i+1,k+1)       = d_pdel(i,k);
         qc(i+1,k+1)          = d_qc(i,k);
         qi(i+1,k+1)          = d_qi(i,k);
-        cldfrac_tot(i+1,k+1) = d_cldfrac_tot(i,k);
         rel(i+1,k+1)         = d_rel(i,k);
         rei(i+1,k+1)         = d_rei(i,k);
         p_lev(i+1,k+1)       = d_pint(i,k);
@@ -529,6 +532,40 @@ void RRTMGPRadiation::run_impl (const int dt) {
 
     gas_concs.set_vmr(name, tmp2d);
   }
+
+  // Set layer cloud fraction.
+  //
+  // If not doing subcolumn sampling for mcica, we want to make sure we use grid-mean
+  // condensate for computing cloud optical properties, because we are assuming the
+  // entire column is completely clear or cloudy. Thus, in this case we want to set
+  // cloud fraction to 0 or 1. Note that we could choose an alternative threshold
+  // criteria here, like qc + qi > 1e-5 or something.
+  //
+  // If we *are* doing subcolumn sampling for MCICA, then keep cloud fraction as input
+  // from cloud fraction parameterization, wherever that is computed.
+  auto do_subcol_sampling = m_do_subcol_sampling;
+  if (not do_subcol_sampling) {
+    const auto policy = ekat::ExeSpaceUtils<ExeSpace>::get_default_team_policy(m_ncol, m_nlay);
+    Kokkos::parallel_for(policy, KOKKOS_LAMBDA(const MemberType& team) {
+      const int i = team.league_rank();
+      Kokkos::parallel_for(Kokkos::TeamThreadRange(team, nlay), [&] (const int& k) {
+        if (d_cldfrac_tot(i,k) > 0) {
+          cldfrac_tot(i+1,k+1) = 1;
+        } else {
+          cldfrac_tot(i+1,k+1) = 0;
+        }
+      });
+    });
+  } else {
+    const auto policy = ekat::ExeSpaceUtils<ExeSpace>::get_default_team_policy(m_ncol, m_nlay);
+    Kokkos::parallel_for(policy, KOKKOS_LAMBDA(const MemberType& team) {
+      const int i = team.league_rank();
+      Kokkos::parallel_for(Kokkos::TeamThreadRange(team, nlay), [&] (const int& k) {
+        cldfrac_tot(i+1,k+1) = d_cldfrac_tot(i,k);
+      });
+    });
+  }
+  Kokkos::fence();
 
   // Compute layer cloud mass (per unit area)
   auto lwp = m_buffer.lwp;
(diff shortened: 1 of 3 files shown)


components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
```diff
@@ -531,6 +534,40 @@ void RRTMGPRadiation::run_impl (const int dt) {
...
+  auto do_subcol_sampling = m_do_subcol_sampling;
+  if (not do_subcol_sampling) {
+    const auto policy = ekat::ExeSpaceUtils<ExeSpace>::get_default_team_policy(m_ncol, m_nlay);
+    Kokkos::parallel_for(policy, KOKKOS_LAMBDA(const MemberType& team) {
+      const int i = team.league_rank();
+      Kokkos::parallel_for(Kokkos::TeamThreadRange(team, nlay), [&] (const int& k) {
+        if (d_cldfrac_tot(i,k) > 0) {
+          cldfrac_tot(i+1,k+1) = 1;
+        } else {
+          cldfrac_tot(i+1,k+1) = 0;
```
AaronDonahue, line 556: could we just initialize `cldfrac_tot` to 0 everywhere and then not need the `else` part of this conditional?

AaronDonahue (dismissed):
Looks good, I had one question that was really just a style question, that won't change the functionality.

AaronDonahue:
Note, this PR is failing `mappy` in SCREAMv1 tests with a DIFF.  Which is expected since it is non-BFB

components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
```diff
@@ -263,6 +263,10 @@ void RRTMGPRadiation::initialize_impl(const RunType /* run_type */) {
   // Determine whether or not we are using a fixed solar zenith angle (positive value)
   m_fixed_solar_zenith_angle = m_params.get<Real>("Fixed Solar Zenith Angle", -9999);
 
+  // Whether or not to do MCICA subcolumn sampling
+  m_do_subcol_sampling = m_params.get<bool>("do_subcol_sampling",false);
+  EKAT_REQUIRE_MSG(not m_do_subcol_sampling, "Error! RRTMGP does not yet support do_subcol_sampling = true");
```
crterai, line 268: What is the logic that determines the error message? If `m_do_subcol_sampling` is `not` true, then write out the message? That doesn't seem right, so I must be missing something.

crterai (commented):
It looks like the PR will include a new namelist parameter of whether to use subcolumn sampling (MCICA), which by default is set to false. And when it's false, it will use all or nothing clouds to compute the radiation (based on whether cldfrc>0). Based on that, it looks good to me. 
I had a couple questions regarding where do_subcol_sampling is set to false by default and the logic to trigger the error message, but I'm less familiar with the C++ side of the model, so if Aaron thinks they look okay, I'm happy with this being merged.

components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
AaronDonahue: ahhh, I think you are right @crterai , I'm surprised this would pass any tests without throwing an error right away.

components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
brhillman: Yeah we probably could! I can do that but it does mean a separate kernel.

components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
```diff
@@ -263,6 +263,10 @@ void RRTMGPRadiation::initialize_impl(const RunType /* run_type */) {
   // Determine whether or not we are using a fixed solar zenith angle (positive value)
   m_fixed_solar_zenith_angle = m_params.get<Real>("Fixed Solar Zenith Angle", -9999);
 
+  // Whether or not to do MCICA subcolumn sampling
+  m_do_subcol_sampling = m_params.get<bool>("do_subcol_sampling",false);
```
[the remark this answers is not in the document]
brhillman: Yup!

components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
AaronDonahue: Ugg, now I'm so confused, but I think Ben is right.  So if `m_do_subcol_sampling` is FALSE than `not m_do_subcol_sampling` would be TRUE.  The REQUIRE does nothing if the input is TRUE.  So yeah, I take back my previous comment.

components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
brhillman: It's not more involved, it's just a `memset(cldfrac_tot, 0.0)`, but my comment was that I think in either case this probably launches a kernel under the hood to do the "work" of setting to zero, especially since this is a device array/view.

components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
brhillman:
```
#define IMPL_THROW(condition, msg, exception_type)                      \
  do {                                                                  \
    if ( ! (condition) ) {                                              \
      std::stringstream _ss_;                                           \
      _ss_ << __FILE__ << ":" << __LINE__ << ": FAIL:\n" << #condition; \
      _ss_ << "\n" << msg;                                              \
      throw exception_type(_ss_.str());                                 \
    }                                                                   \
  } while(0)

#define EKAT_REQUIRE_MSG(condition, msg)              IMPL_THROW(condition, msg, std::logic_error)
```

brhillman: If this is too confusing, we could change the flag to something like `disable_subcol_sampling`, so the assertion would look like
```
EKAT_REQUIRE_MSG(disable_subcol_sampling, "ERROR! Need to disable subcol sampling for now.");
```
Or, we could take subcol sampling out of it and use a flag that just says `use_gridmean_clouds` and then the assertion would be
```
EKAT_REQUIRE_MSG(use_gridmean_clouds, "ERROR: need to use grid-mean clouds for cloud optics until MCICA sampling is implemented");
```

crterai: Ahh, I understand now. It throws an error if the first condition is not satisfied. Then it looks good to me.

crterai (dismissed):
My questions were resolved.

components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
AaronDonahue: I think this is good, I think I just confused myself :-)

AaronDonahue:
@brhillman . I think this branch just needs the whole `merge master in` treatment after the upstream merge went in.  Then I suspect we'll see passing everywhere except the expected DIFFs in SCREAMv1.

brhillman:
Rebased.

crterai (approved):

AaronDonahue (approved):

AaronDonahue:
The last time this ran through the AT it passed all tests except the expected DIFFs in the SCREAMv1 tests.  I think we can manually merge this, unless there are objections...

PR #1630, opened 2022-05-10
Make "all or nothing" ice cloud fraction default in SCREAMv0 compsets

bogensch:
PR #944 added an "all or nothing" ice cloud fraction scheme to v0/v1 SCREAM codes.  In SCREAMv1 this is used by default.  In SCREAMv0 this was never set as the default ice cloud fraction scheme, which makes it very easy to unintentionally use the unwanted relative humidity based definition when running v0.  This PR sets the default to the all or nothing ice cloud fraction scheme in the SCREAMv0 compsets.

### DIFF
--- a/components/eam/bld/namelist_files/namelist_defaults_eam.xml
+++ b/components/eam/bld/namelist_files/namelist_defaults_eam.xml
@@ -1144,7 +1144,7 @@
 <cldfrc_iceopt  phys="default"                > 5 </cldfrc_iceopt>
 
 <cldfrc_minice                             > 1.0e-12 </cldfrc_minice>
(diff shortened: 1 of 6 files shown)


crterai (approved):
Looks good to me. Thanks for making this fix!

AaronDonahue:
@crterai , can you confirm that you approve of this PR?  The AT reported the expected DIFFs in SCREAMv0 runs, so once you confirm I can manually merge.

crterai (approved):
looks good!

AaronDonahue:
I'm going to manually merge now that @crterai has re-reviewed and the only fails are DIFFs in SCREAMv0 runs - which are expected.

PR #1632, opened 2022-05-11
Work-around for slurm settings on perlmutter

ndkeen:
For perlmutter, and only for compiler=gnugpu, we need a different `gpu_bind` settings for MMF and non-MMF runs. We think this might be a system issue and NERSC is aware, but for now, this works.

Also added a newline for a print statement.

Fixes https://github.com/E3SM-Project/E3SM/issues/4834

### DIFF
--- a/cime_config/machines/config_batch.xml
+++ b/cime_config/machines/config_batch.xml
@@ -374,18 +374,23 @@
     <directives>
       <directive> --constraint=gpu</directive>
     </directives>
-    <directives compiler="nvidiagpu">
-      <directive> --gpus-per-task=1</directive>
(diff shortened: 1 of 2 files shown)


ndkeen:
Can confirm that `SMS_P8x1_Ln12.ne30_ne30.F2000SCREAMv1.perlmutter_gnugpu` will work out-of-box with this change.

Note that current ne30 default is to use 1 node on PM which isn't enough memory and will fail. We might be making some memory use improvements to allow it to fit on 1 node.
Also we still need `gnugpu` as compiler, though I will be changing that to the default soon.

jgfouca (approved):

jgfouca:
Skipping testing since there is no potential impact to any CI machine.

PR #1634, opened 2022-05-11
dtime updates

mt5555:
Correct highres dtime defaults for SCREAM:

ne256  600s
ne512  200s 
ne1024 100s

### DIFF
--- a/driver-mct/cime_config/config_component_e3sm.xml
+++ b/driver-mct/cime_config/config_component_e3sm.xml
@@ -384,9 +384,9 @@
       <value compset=".+" grid="a%ne60np4">96</value>
       <value compset=".+" grid="a%ne120np4">96</value>
       <value compset=".+" grid="a%ne240np4">144</value>
-      <value compset=".+" grid="a%ne256np4">288</value>
-      <value compset=".+" grid="a%ne512np4">576</value>
-      <value compset=".+" grid="a%ne1024np4">1152</value>
+      <value compset=".+" grid="a%ne256np4">144</value>
+      <value compset=".+" grid="a%ne512np4">432</value>
+      <value compset=".+" grid="a%ne1024np4">864</value>
       <value compset=".+" grid="a%ne0np4_arm_x8v3" >144</value>
       <value compset=".+" grid="a%ne0np4_conus_x4v1" >96</value>
       <value compset=".+" grid="a%ne0np4_northamericax4v1" >48</value>


mt5555:
Should this go in E3SM first?

jgfouca (approved):

jgfouca:
Your PR didn't do this, but I'm wondering what is the value of all these compset selectors that match everything `compset=".+"`? I think that should be equivalent to just not having a compset selector at all.

mt5555:
Jim, just to confirm, since this is cime_config code - we can merge direct to SCREAM, or should it go to E3SM first?
(or branch off comment parent and merge to both?)

jgfouca:
@mt5555 , thanks for bringing that up. The changes look like that could affect any COMPSET, not just SCREAM ones, so we should keep this in SCREAM.

jgfouca:
Alternatively, we can change all those COMPSET selectors so that only SCREAM COMPSETs are matched.

mt5555:
They only impact ne256, 512 and 1024 - which are not E3SM resolutions, so this is safe to merge into E3SM. 

And if someone did want to run E3SM at these resolutions, these would be the correct settings.

PeterCaldwell:
> They only impact ne256, 512 and 1024 - which are not E3SM resolutions, so this is safe to merge into E3SM. And if someone did want to run E3SM at these resolutions, these would be the correct settings.

That sounds like sound reasoning to me. So I guess we make this PR in E3SM repo?

bartgol:
Is there any reason to do this PR in E3SM? It is my understanding that nobody will run those resolution without EAMxx. They can just wait for EAMxx to be merged back in E3SM, so we can do the PR here only.

jgfouca:
@bartgol , since this file lives in E3SM too, and we want to avoid conflicts, it should probably go into E3SM unless it is needed in SCREAM today.

mt5555:
since it's such a simple change, I was thinking of branching off commen parent, then merge the same branch to both. (so we dont have to wait for the e3sm merge).

jgfouca:
@mt5555 , up to you. I did an upstream merge for SCREAM earlier this week and it was trivial. The delay would be waiting for all the integration testing to go through for E3SM if you went that route.

bartgol:
> @mt5555 , up to you. I did an upstream merge for SCREAM earlier this week and it was trivial. The delay would be waiting for all the integration testing to go through for E3SM if you went that route.

I think he mant to issue the same PR in both repos _at the same time_. That's better than doing a PR in E3SM followed by an upstream merge in scream.

mt5555:
Yes, I'm curious to try this. theoretically it should avoid conflicts when
we do have the next merge from E3SM.

jgfouca (dismissed):

mt5555:
@jgfouca , can you redo your review?  I force updated after rebasing to the e3sm/scream merge base.  I think that's why it now says "review required"?

bartgol (approved):
Since this is trivial, I can go ahead and approve without waiting for jim, so this moves along.

AaronDonahue:
The AT is down, so this is being held up.  But nothing about this PR should impact the AT tests to begin with.  I'm going to manually merge unless someone objects...

PR #1637, opened 2022-05-12
Initialize SHOC v1 turbulence consistent with v0

bogensch:
In SHOC v0 various input turbulence quantities are initialized uniformly (TKE, eddy diffusivity, SGS buoyancy flux) with either 0 or the minimum tolerance value (TKE).  

Currently, v1 initializes these quantities via initialization file with values from a spun-up simulation.  While there is nothing technically wrong with this, it is an inconsistency with v0 which makes doing apples-to-apples comparison between v0 and v1 difficult (especially diagnosing v1 crashes that occur during spin up period).  In addition, I am more comfortable allowing SHOC turbulence to spin-up in conjunction with the rest of the model since I'm not 100% sure of the impacts of doing it the v1 way.

Confirmed that changes in this PR perform scientifically as expected in ne30 SCREAMv1.

### DIFF
--- a/components/scream/src/physics/shoc/atmosphere_macrophysics.cpp
+++ b/components/scream/src/physics/shoc/atmosphere_macrophysics.cpp
@@ -335,7 +335,7 @@ void SHOCMacrophysics::initialize_impl (const RunType run_type)
   
   // Some SHOC variables should be initialized uniformly if an Initial run
   if (run_type==RunType::Initial){
-//    Kokkos::deep_copy(sgs_buoy_flux,0.0);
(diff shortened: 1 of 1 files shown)


AaronDonahue (approved):
Good bug-catching Peter!

AaronDonahue:
I think this can go ahead and be manually merged.  The only FAILS in the AT are the expected DIFFs in the SCREAMv1 simulations.

AaronDonahue:
I'm going to go ahead and manually merge this now.

PR #1641, opened 2022-05-13
Update summit env to GCC 9.3

jgfouca:
GCC was old (7.5) which and we were seeing internal compiler failures
in some debug builds.


ndkeen (approved):
Looks ok, but haven't tested it

bartgol (approved):

AaronDonahue:
@jgfouca , should we just go ahead and merge this manually?  The AT is down and I see it is approved and even has "integrate without testing" set.

PR #1642, opened 2022-05-13
Create scream high resolution test suite

jgfouca:
Haven't tried to do much with this suite other than ensure that it at least builds on mappy. Now that the medium res suite is "working" (model runs to completely but times out due to not releasing resources) on summit, I want a higher resolution suite to try.

### DIFF
--- a/cime_config/tests.py
+++ b/cime_config/tests.py
@@ -465,8 +465,8 @@ _TESTS = {
     "e3sm_scream_v1_hires" : {
         "time"  : "03:00:00",
         "tests" : (
-            "SMS_D_Ln2.ne120_ne120.F2000SCREAMv1",
-            "SMS_Ln2.ne120_ne120.F2000-SCREAMv1-AQP1",
+            "SMS_D_Ln12.ne120_ne120.F2000SCREAMv1",
+            "SMS_Ln12.ne120_ne120.F2000-SCREAMv1-AQP1",
             )
     },
 
--- a/cime_config/tests.py
+++ b/cime_config/tests.py
@@ -462,6 +462,13 @@ _TESTS = {
             )
     },
 
+    "e3sm_scream_v1_hires" : {
+        "time"  : "03:00:00",
+        "tests" : (
+            "SMS_D_Ln2.ne120_ne120.F2000SCREAMv1",
+            "SMS_Ln2.ne120_ne120.F2000-SCREAMv1-AQP1",
+            )
+    },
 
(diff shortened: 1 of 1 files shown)


ndkeen (commented):
I think we should be running more steps for the larger res (at least 12?).
I don't think we have aqua planet compset at this res?
3 hours seems way too long.

jgfouca:
@ndkeen , I bumped to 12 steps. I will await others' thoughts on compsets.

AaronDonahue (approved):
Looks good to me.

PeterCaldwell:
- There's a slight chance that the inputdata needed to run ne120 isn't on the blues server yet... I've been working on moving it there from NERSC but had some login issues. 
- I agree with Noel that we probably don't have the initial condition file needed to run aquaplanet at ne120. The compset should be resolution independent, but the IC doesn't exist. Ben's kind of slammed with creating ne1024 and 256 ICs, so I don't think we should expect the aquaplanet config any time soon. I'd delete that test for the moment.
- We should use F2010-SCREAMv1 rather than F2000SCREAMv1 compsets. F2000 is deprecated. 
- It's a bit odd that you use ne120_ne120 rather than ne120_r0125_oRRS18to6v3 as the grid. My understanding is that ne120_ne120 uses CICE instead of MPAS-sea-ice (because the latter needs to run on MPAS grids). We need to use MPAS at ne1024 because CICE doesn't scale (at least, that's my recollection), but the trigrid config we've been using at ne120 has a CPL bug in albedo which has been giving us problems. Eventually we want to move to ne120_oRRS18to6v3 but we don't have the land inputdata for that config yet. All of this is to say that I think what you have is fine for now, but keep in mind that we will need to change it later.
- Combining my point 2 and point 4, I'd suggest having a debug compiled test and a opt-compiled test. Maybe make them identical compset (F2010-SCREAMv1) but have one use ne120_ne120 and the other use ne120_r0125_oRRS18to6v3?

ndkeen:
Oh yes I missed that.  I think we want `ne120_r0125_oRRS18to6v3` -- `ne120_ne120` might not even work.

bartgol:
Taking myself off the reviewer cause I don't know much about what's good for our compsets. Peter knows more about this.

jgfouca:
@PeterCaldwell , look OK now?

PeterCaldwell (approved):
Yup, looks great. Like Noel said, I'm not sure the ne120_ne120 case will run at all but I hope it does and it would be good to test.

PR #1647, opened 2022-05-17
Fix nccn in P3, F90 interface has conversion not in cpp version.

AaronDonahue:
This commit matches a conversion of the nccn values that exists in
the F90 interface but is missing in the cpp interface.

This commit adjusts the incoming nc_activated values to be cell-average
values when calculating nccn, which is an input for p3_main.

This commit will also update the minimum threshold ice cloud to be present.


AaronDonahue:
This will implement the bugfix that @hassanbeydoun identified in the p3 interface.  I'm setting as WIP right not because I haven't tested it.  I'll first test on my local machine and then go ahead and let the AT test it.

AaronDonahue:
@hassanbeydoun , can you grab this branch and confirm it fixes the issue with nc_activated being too low?

PeterCaldwell:
I struggle to understand what's going on in this PR. It seems like all we need to do is multiply nccn by cldfrac, so I don't understand why this PR leaves us carrying around an additional variable. Can you explain a bit what's going on, @AaronDonahue ? Also, the variable names seem somewhat confusing. nccn is the number of CCN, I think, but it also seems to be equal to nc_activated, which I would think is the number of nc activated in this step (which would be nccn_newstep - nccn_oldstep).

AaronDonahue:
@PeterCaldwell 

The main reason we have a new variable, `nccn`, is the decision to keep nc_activated as an input only variable to P3.  If we wanted to just take `nc_activated` and scale it by the cloud fraction we would have to change it to "Updated".  Which is fine if we want to make that change.

in terms of `nc_activated`, this is the variable produced by SPA when interpolating `CCN3`.  If it's a bad name we can change that, right now it follows the convention used in SCREAMv0.  But you are correct, nccn and nc_activated are basically the same, with the former just being scaled by the cloud fraction.  I'm happy to change the variable name to something like `nc_activated_incld`, although I think @hassanbeydoun was saying that terminology doesn't actually make any sense, there is no "in-cloud" nc_activated...

Let me know if there are any changes you want, I'm happy to do them.  This current implementation is essentially a copy of what is done in SCREAMv0, so it can be adjusted.

bartgol:
Can't we just replace all uses of `nccn` with `nc_activated / cldfrac` (or whatever the formula is)? Unless we use nccn _many_ times, it's ok to replace it with the simple expression and do away with nccn, I would say.

AaronDonahue:
`nccn` is an argument for p3_main.  The question there is, do we want to change p3_main?  If so then we are out of sync with the F90 code, so we'll need to change it there too.

PeterCaldwell:
I like Luca's idea, which I don't think Aaron grokked. If I understand correctly, Luca's idea is to not change anything in the atm_microphysics interface at all and instead to just multiply or divide nccn by cldfrac inside p3_main wherever it is used... which I think is just one place. That way we would get around the input-only status of nccn because we wouldn't ever be redefining it. If I understand correctly, nccn is equal to CCN3 which is just the CCN concentration, not the number of new activated drops. So I like using nccn and deleting all mention of nc_activated. I think nc_activated was created by this PR, so that should be easy. If it's not, let's leave that correction for v1.1.

bartgol:
> I like Luca's idea, which I don't think Aaron grokked. If I understand correctly, Luca's idea is to not change anything in the atm_microphysics interface at all and instead to just multiply or divide nccn by cldfrac inside p3_main wherever it is used... which I think is just one place. That way we would get around the input-only status of nccn because we wouldn't ever be redefining it. If I understand correctly, nccn is equal to CCN3 which is just the CCN concentration, not the number of new activated drops. So I like using nccn and deleting all mention of nc_activated. I think nc_activated was created by this PR, so that should be easy. If it's not, let's leave that correction for v1.1.

Yes, that's my idea. But to Aaron's point, if the point where nccn is needed is in a function that does not have cldfrac as input, then the particular function interface has to be changed.

That said, I still think it's cleaner to change the interface of a handful of function, rather than add another variable just to store a super simple calculation that is later only used once.

AaronDonahue:
@PeterCaldwell - extra points for use of the word "grok"

But in more seriousness.  I'm actually fairly flexible on where we adjust the incoming CCN3 value by the cloud fraction.  This PR follows the example of the F90 code, but that may not be how we want things to work in the actual cpp code.  So now is the opportunity to change.

To the point of new variables:
1. `nc_activated` is not new in this PR, it is the variable produced by SPA and gathered from the field manager by P3.  What is new is creating a local variable in the memory buffer (in this case `nccn`) that will hold `nc_activated` multiplied by the cld fraction.  @PeterCaldwell , I'm happy to make the change in the variable name but as you point out, it is a bit more involved so maybe best left to v1.1.  To that point, inside `p3_main` this variable is called `nc_prescribed`, maybe that's a better name overall?
2. @bartgol , the `nc_activated`/`nccn` variable is used inside `p3_main_part1` (see `p3_main_part1_impl.cpp`) which also has the `inv_cld_fraction` passed to it.  So presumably we could just locally use the old approach and divide by `inv_cld_fraction` in the one location it is used.  It would turn this PR into a single line change.
3. My last point, which I only mention to ensure we don't make a change without all counter-points addressed, is that if we locally adjust by cloud fraction then we are setting the assumption that our main P3 cpp code assumes that the prescribed nc value needs to be scaled by cloud fraction.  @hassanbeydoun may speak to this best, but is this a fair assumption from the perspective of the next person who may want to add an aerosol forcing to P3, or use our cpp code in their new model?  In other words, does this assumption still hold under the SCREAM design assumption that all "main" codes are blackboxes where the input/outputs are not specific to EAM.

bartgol:
> 3. My last point, which I only mention to ensure we don't make a change without all counter-points addressed, is that if we locally adjust by cloud fraction then we are setting the assumption that our main P3 cpp code assumes that the prescribed nc value needs to be scaled by cloud fraction.  @hassanbeydoun may speak to this best, but is this a fair assumption from the perspective of the next person who may want to add an aerosol forcing to P3, or use our cpp code in their new model?  In other words, does this assumption still hold under the SCREAM design assumption that all "main" codes are blackboxes where the input/outputs are not specific to EAM, or in this case SPA.

Perhaps this (together with point 1) calls for a very specific variable name. Mabe instead of `nc_activated` one can use `nc_activated_over_cldfrac`. This is probably too verbose, but at least leaves no room for misunderstandings.

A different alternative (maybe v1.2) would be to make SPA produce `nc` rather than `nc/cldfrac`. That would leave no room to misunderstanding maybe?

AaronDonahue:
So based on the above comments... are we in favor of 
1. Adopting the name `nccn` everywhere, including in the SPA interface and
2. Dividing by `inv_cld_frac` within the `p3_main` code
??

@PeterCaldwell , @bartgol , @hassanbeydoun ?

hassanbeydoun:
> So based on the above comments... are we in favor of
> 
> 1. Adopting the name `nccn` everywhere, including in the SPA interface and
> 2. Dividing by `inv_cld_frac` within the `p3_main` code
>    ??
> 
> @PeterCaldwell , @bartgol , @hassanbeydoun ?

Sounds good to me!

PeterCaldwell:
Yeah, I think so. nccn: definitely better. Divide by inv_cld_frac: yes, and in fact inv_cld_frac is already available in p3_main_part1. I think you want inv_cld_frac_l since this is liquid droplet number...

bartgol:
I am ok with whatever name you choose. I just don't want to add variables (even if they use scratch mem) just to store A*B.

AaronDonahue:
@bartgol and @PeterCaldwell , I have updated this PR with the new plan,
1. Simpler implementation - divide by inv_cld_frac_l inside p3_main_part1
2. Better variable name - nc_activated => nccn
3. Also fix the ice cloud fraction threshold.

One last thing that came to me while making these changes.  We also have the variable `ni_activated` which is kind of like `nc_activated`.  So two questions:
1. Should that be changed to a different name?
2. More importantly, a quick grep of `ni_activated` didn't reveal WHO provides ni_activated.  It's "required" by P3 but doesn't seem to come from anywhere.  Who should be producing ni_activated?  SPA?  @hassanbeydoun or @PeterCaldwell , any thoughts?

PeterCaldwell:
Ah, I like the new version much better. Do we need to make the same corrections in v0 in order to avoid non-bfb diffs in p3 regression tests?

Regarding ni_activated - I think the name is fine... this is the total number of activated drops. I also think its implementation is fine, though I'd like a second opinion from Hassan. My recollection is that ni_activated is something the user can specify in the initial condition file (along with the do_prescribed_CCN flag) to force the model to use a prescribed ni. That's why no process is responsible for producing it - only a user would create it.

hassanbeydoun:
> @bartgol and @PeterCaldwell , I have updated this PR with the new plan,
> 
> 1. Simpler implementation - divide by inv_cld_frac_l inside p3_main_part1
> 2. Better variable name - nc_activated => nccn
> 3. Also fix the ice cloud fraction threshold.
> 
> One last thing that came to me while making these changes. We also have the variable `ni_activated` which is kind of like `nc_activated`. So two questions:
> 
> 1. Should that be changed to a different name?
> 2. More importantly, a quick grep of `ni_activated` didn't reveal WHO provides ni_activated.  It's "required" by P3 but doesn't seem to come from anywhere.  Who should be producing ni_activated?  SPA?  @hassanbeydoun or @PeterCaldwell , any thoughts?

ni_activated is unused in v1. having SPA be true tells p3 to calculate ice nucleation internally. If you want to be consistent with names, rename ni_activated ninp (INP being analogous to CCN) but might not be worth since its unused.

PeterCaldwell:
Ok, I think Hassan's and my comments (written at the same time) say the same thing. Let's ignore ni_activated for the purposes of v1.

AaronDonahue:
@PeterCaldwell and @bartgol , I need some advice on how to get this PR into the passing category.  This change makes the F90 and CPP implementation of p3_main_part1 no longer B4B.  This is because the F90 relies on the interface scaling by cloud fraction and in this PR we are having the cpp code do the scaling by cloud fraction.  This leads to FAILs in the p3_tests, understandably.  

I could scale the cpp data (or the F90 data) in the tests themselves, to mimic what is done on the interface level, but I would expect we could still be not exactly B4B.  The alternative would be to do something different in p3_main_part1 if we are doing B4B testing, or to remove the tests altogether.  I think the latter is not what we want.

Do either of you have any other ideas?  Comments?

PeterCaldwell:
I expected that you would change the F90 to do the same thing as we're doing now in the C++. Is that not possible or difficult?

AaronDonahue:
> I expected that you would change the F90 to do the same thing as we're doing now in the C++. Is that not possible or difficult?

I can do that.  I didn't know if that would disrupt the SCREAMv0 simulations, but I suppose if in the end we are B4B then it shouldn't be an issue.

hassanbeydoun:
> > I expected that you would change the F90 to do the same thing as we're doing now in the C++. Is that not possible or difficult?
> 
> I can do that. I didn't know if that would disrupt the SCREAMv0 simulations, but I suppose if in the end we are B4B then it shouldn't be an issue.

I don't see why it would disrupt v0. I think the same change can be made there.

AaronDonahue:
This is largely fixed.  The last step is to update the initial condition files for p3 tests to have `nccn` instead of `nc_activated` as the init variable.

AaronDonahue:
Is there any opposition, while I'm at it, to changing the initial condition file used by all "stand-alone" tests to use the same file?  Currently we have different tests using their own specific files, but I believe they can all use the homme_shoc_cld_spa_p3_rrtmgp_init file and still work fine.  That would make it easier to update an init file when we need too, rather than having to update a handful of files and uploading all of those to blues, we would have just the one.

PeterCaldwell:
I thought we were trying to move to using @brhillman 's "screami" initial condition files for everything instead of the older homme_shoc_cld_spa_p3_rrtmgp_init files? But I agree, much better to use a single initial condition file for all our unit tests.

AaronDonahue:
Our standalone tests are all ne2.  So that may be why.  No reason we can't remap Bens init condition file to ne2 though...

AaronDonahue:
Leaving this at WIP until I can manually check that CIME based runs also work.  Then I'll have the new files uploaded to blues and let that AT go for it.

AaronDonahue:
WIP until I can upload two new files to blues.

components/scream/cime_config/namelist_defaults_scream.xml
```diff
@@ -267,10 +267,11 @@ tweaking their $case/namelist_scream.xml file.
...
     <!-- ${DIN_LOC_ROOT}/atm/scream/init/map_ne30np4_to_ne512np4_mono_20220506.nc, -->
-    <!-- ${DIN_LOC_ROOT}/atm/scream/init/screami_ne512np4L128_20220505.nc, -->
-    <!-- ${DIN_LOC_ROOT}/atm/scream/init/screami_ne120np4L72_20220503.nc, -->
-    ${DIN_LOC_ROOT}/atm/scream/init/screami_ne30np4L72_20220503.nc,
-    ${DIN_LOC_ROOT}/atm/scream/init/init_ne4np4.nc
+    ${DIN_LOC_ROOT}/atm/scream/init/screami_ne30np4L72_20220524.nc,
+    ${DIN_LOC_ROOT}/atm/scream/init/screami_ne4np4L72_20220524.nc,
+    ${DIN_LOC_ROOT}/atm/scream/init/screami_ne120np4L72_20220524.nc,
+    ${DIN_LOC_ROOT}/atm/scream/init/screami_ne512np4L128_20220525.nc,
+    scream-aquaplanet_init_ne4np4_L72_20220524.nc
```
PeterCaldwell, line 274: isn't this last line missing DIN_LOC_ROOT stuff?

PeterCaldwell (dismissed):

components/scream/cime_config/namelist_defaults_scream.xml
AaronDonahue: yes, I'll fix that.  thanks for catching it.

PeterCaldwell (approved):

PeterCaldwell:
Just copying `https://svn-ccsm-inputdata.cgd.ucar.edu/trunk/inputdata` into a web browser reveals it *isn't* down. I suspect something got screwed up with the SNL firewall again. Could you check, @jgfouca ? 

Maybe we should just copy the file where it is needed to get things rolling again this time?

PeterCaldwell (approved):

AaronDonahue:
The DIFFs are expected because this is a non-BFB PR, so I am going to manually merge.

PR #1648, opened 2022-05-17
Fix bug in SPA when running simulations with levels not = 72

### LINKED ISSUE #1646: SPA file load error with 128 levels
At both ne4 and ne30, CIME v1 runs using initial conditions with 128 levels fail with:
```
PIO: FATAL ERROR: Aborting... An error occured, Reading variable (hyam, varid=10) from file 
(/global/cfs/cdirs/e3sm/inputdata/atm/scream/init/spa_file_unified_and_complete_ne30_20220428.nc, ncid=20) failed with 
PIO_IOTYPE_PNETCDF iotype. The low level (PnetCDF) I/O library call failed to read the variable (Number of regions = 1, iodesc id
 = 513, Bytes to read on this process = 128). NetCDF: Start+count exceeds dimension bound (err=-57).
```
Note that this spa file always has 72 vertical levels - which is the same number of levels all previous v1 runs used. It is supposed to interpolate in the vertical to other grids, but perhaps it's not doing so?

128-level initial conditions on NERSC are: 
/global/cfs/cdirs/e3sm/bhillma/scream/data/init/screami_ne4np4L128_20220512.nc
/global/cfs/cdirs/e3sm/bhillma/scream/data/init/screami_ne30np4L128_20220512.nc

AaronDonahue:
This commit fixes a bug in the SPA interface where it was assumed that
the SPA data and the simulation would have the same number of vertical
levels.  Since SPA data is on a 72 level grid this meant we could not
run at 128 levels.

This commit fixes the definition of how many levels are in the SPA data
to only come from reading that info from the SPA data file.  Presumably
this would also allow us to use SPA data with a number of levels different
than the current 72.

Fixes #1646


AaronDonahue:
I also added a unit test for running SPA with 128 level simulation but that is currently not active because the necessary files are not on the Blues server.  Once those are added we can add a commit that makes this test active.

AaronDonahue:
huh, so strangely, it looks like with 128 levels the homme+physics test is not BFB for multiple ranks.  I'm going to investigate a bit on blake, which is failing.

AaronDonahue:
I've tracked this down to p3 rain sedimentation, which is unrelated to these changes.  I think this is an issue that has been present but only pops up now that we are attempting 128 levels.

bartgol:
Yeah, when possible, we should test our code with both `nlevs%pack_size==0` and `nlevs%pack_size!=0` (obviously, only on CPU). We might have limited our testing to only one of these.

AaronDonahue:
I just manually tested on blake for `full_debug` and the tests that fail the AT pass.  So I am going to retest.

AaronDonahue:
@bartgol , when you get a chance can you take a look at these changes?

AaronDonahue:
oh wait, nvmd.  The tests run, but are non-BFB.  I'm going to investigate.

bartgol:
@AaronDonahue I'll wait till you're done fixing this, then will review.

AaronDonahue:
Update on these fails.  It looks like with 128 `homme+physics` fails B4B comparison but `physics_only` and `homme_standalone` both look like they pass.  I'm testing `homme+shoc` to test the hypothesis that maybe there is an issue in the homme->physics remapping with 128 levels.

AaronDonahue:
Not sure why this is failing, I manually tested on blake and it passed for me.

bartgol:
Yes, mappy filled up the hard drive b/c of all the case scratch folders. I freed half a TB, which should get us going for a bit. Long term, we need to create a jenkins job that cleans up old case scratch folders.

AaronDonahue:
I'm not sure why this DIFFs for ERP and ERS.  Does anyone know if DIFF's mean a DIFF with some known baseline, like SMS, or these should be interpreted as FAILs?

bartgol:
> I'm not sure why this DIFFs for ERP and ERS. Does anyone know if DIFF's mean a DIFF with some known baseline, like SMS, or these should be interpreted as FAILs?

The mac_mic PR went in, but we did not re-bless. I can bless, and then we can re-test.

bartgol (approved):

PR #1652, opened 2022-05-20
Fix memory access error in unflatten_idx util

bartgol:
This should take care of the valgrind errors, and likely also of the cuda memcheck errors.

On my workstation, I ran a few tests that failed before (with valgrind) and they pass. Since tests take long, I'm willing to just merge this, and let the nightlies verify if the fails are gone.


AaronDonahue (approved):
I agree. I'll approve and we can see how the nightlies respond.

PR #1655, opened 2022-05-20
Fix usage of parallel reduce in field property checks

bartgol:
The reducers were templated on the wrong space (device instead of host).


PeterCaldwell:
FYI Aaron is helping out with little Luca's class this AM, so won't review for a few hours.

AaronDonahue (approved):

PR #1664, opened 2022-05-23
Fix v1 mac mic subcycles parameter

### LINKED ISSUE #1657: resolution dependent Num__Subcycle__Iterations
Num__Subcycle__Iterations in namelist_defaults_scream.xml should vary depending on resolution. We should just copy what EAMf90 uses for each resolution.

### LINKED ISSUE #1663: v1 macmic equivalent does not work
I believe the reason is that in the XML file one sees
```xml
<Num__Subcycle__Iterations>18</Num__Subcycle__Iterations>
```
In the emitted YAML one then sees
```yaml
Num Subcycle Iterations: 18
```
But in the C++ one sees
https://github.com/E3SM-Project/scream/blob/6adad54bdc6bb95ed3463719d36ebdb765dd00d6/components/scream/src/share/atm_process/atmosphere_process.cpp#L27-L29
If this last is switched to "Num Subcycle Iterations", then the macmic behavior works.

If I've gotten this right, I have a suggestion to help prevent this kind of mistake in the future. In the YAML parser, keep a table of queries for specific entires. At the end of initialization, output a warning if there are entries in the input YAML that have not been queried by any process. This isn't necessarily an error, so it shouldn't end the simulation, but an informative warning would be helpful. In the opposite direction, one might also output a list of fields processes have queried that do not have entires in the input YAML.

bartgol:
Namely:

* Make parameter default value resolution dependent
* Fix name of parameter, matching what's used in the C++ code.

I took the resolution dependent values from `components/eam/namelist_files/use_cases/2010_scream_hr.xml`. Please, double check that they are indeed the right ones.

Fix #1657 
Fix #1663

### DIFF
--- a/components/scream/cime_config/namelist_defaults_scream.xml
+++ b/components/scream/cime_config/namelist_defaults_scream.xml
@@ -195,7 +195,7 @@ tweaking their $case/namelist_scream.xml file.
 
     <mac_aero_mic inherit="atm_proc_group">
       <atm_procs_list>(shoc,cldFraction,spa,p3)</atm_procs_list>
-      <Number__of__Subcycles hgrid="ne4np4">6</Number__of__Subcycles>
+      <Number__of__Subcycles hgrid="ne4np4">24</Number__of__Subcycles>
       <Number__of__Subcycles hgrid="ne30np4">6</Number__of__Subcycles>
       <Number__of__Subcycles hgrid="ne120np4">3</Number__of__Subcycles>
       <Number__of__Subcycles hgrid="ne256np4">3</Number__of__Subcycles>
--- a/components/scream/cime_config/namelist_defaults_scream.xml
+++ b/components/scream/cime_config/namelist_defaults_scream.xml
@@ -133,7 +133,7 @@ tweaking their $case/namelist_scream.xml file.
 
     <!-- Basic options for each atm process -->
     <atm_proc_base>
-      <Num__Subcycle__Iterations constraints="gt 0">1</Num__Subcycle__Iterations>
+      <Number__of__Subcycles constraints="gt 0">1</Number__of__Subcycles>
       <Enable__Precondition__Checks type="logical">true</Enable__Precondition__Checks>
       <Enable__Postcondition__Checks type="logical">true</Enable__Postcondition__Checks>
     </atm_proc_base>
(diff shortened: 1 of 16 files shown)


AaronDonahue:
I am going to defer to @brhillman, @PeterCaldwell or @crterai on this one since I think they have more experience running at high resolution.  But if these come directly from the SCREAMv0 settings they are probably correct.

Note, the only fails for this test are DIFFs for two ne4 cases.  I would expect this to be B4B in ne4 since the number of subcycles stays the same.  Any reason they could end up being DIFF?

bogensch:
SHOC should definitely have dt <= 5 min.

bartgol:
> Note, the only fails for this test are DIFFs for two ne4 cases. I would expect this to be B4B in ne4 since the number of subcycles stays the same. Any reason they could end up being DIFF?

Yes, due to the different name used in xml and C++ code, we were not really using the input value of subcycle (6) in `mac_aero_mic`, but rather use the default value (1) set in `atm_proc_base`. This PR fixes that, so that now mac_aero_mic is _really_ subcycled, and therefore the diffs.

ambrad:
Suggestion based on "SHOC should definitely have dt <= 5 min." Add a REQUIRE to the SHOC interface init routine to check that this condition holds.

AaronDonahue:
@ambrad good suggestion, can we add that check to this PR.  Follow-up, should P3 have a similar check?

ambrad:
I don't know about P3. Perhaps Peter or Chris know the dt limit for P3 that matters in practice.

PeterCaldwell:
No, I don't think P3 cares that much what its timestep is. It's best if it has the same timestep as SHOC, but I don't think we need a require.

PeterCaldwell (approved):
We should add default ATM_NCPL for ne256, 512, and 1024, but let's do that in a separate PR because it will take some discussion to decide what they should be.

bartgol:
My bad, I forgot a semicolon and wasted an AT run.

AaronDonahue:
@PeterCaldwell , should we keep the ones in this PR as is and then update in a subsequent PR if they need to be changed?

PeterCaldwell:
> should we keep the ones in this PR as is and then update in a subsequent PR if they need to be changed?

 Your question made me realize the numbers I provided above were from driver-moab/... instead of driver-mct/... version of cime_config/config_component_e3sm.xml . The mct version (which we use) has ATM_NCPL values for ne1024 and other high resolutions. So neglect my earlier comment - this PR is good, no follow-up is needed.

AaronDonahue (approved):

AaronDonahue:
I'm approving given @PeterCaldwell 's comment above.  Since our other PR's are failing in the aquaplanet test and this should fix that I'm working to shepherd it through quickly.

PeterCaldwell:
I just updated SHOC substeps in my table above (https://github.com/E3SM-Project/scream/pull/1664#issuecomment-1134859397) confirming that all the values we use now are good. We could use 2 rather than 3 subcycles for ne256, but I think what we have is fine.

PeterCaldwell (approved):

PeterCaldwell:
@bartgol - can I add the automerge flag for this?

bartgol:
@PeterCaldwell Yes, but the AT is now failing across the board. I'll have to check what's going on.

AaronDonahue (approved):

AaronDonahue:
The only FAILs are DIFFs for SCREAMv1 tests on mappy, which given that we are now running with actual subcycles are expected DIFFs.  I think this one is ready to merge - @bartgol

bartgol:
Looks like we're ready to merge.

PR #1667, opened 2022-05-23
Use YAKL_DEBUG for DEBUG builds

brhillman:
Add the `YAKL_DEBUG` CPP macro when building SCREAM with DEBUG mode (when `CMAKE_BUILD_TYPE` is `DEBUG`). This adds additional YAKL debugging.

### DIFF
--- a/components/scream/src/physics/rrtmgp/CMakeLists.txt
+++ b/components/scream/src/physics/rrtmgp/CMakeLists.txt
@@ -14,6 +14,12 @@ add_subdirectory(${SCREAM_BASE_DIR}/../../externals/YAKL ${CMAKE_BINARY_DIR}/ext
 target_compile_options(yakl PUBLIC $<$<COMPILE_LANGUAGE:CXX>:${YAKL_CXX_FLAGS}>)
 EkatDisableAllWarning(yakl)
 
+# For debug builds, set -DYAKL_DEBUG                                                                                       
+string(TOLOWER "${CMAKE_BUILD_TYPE}" CMAKE_BUILD_TYPE_ci)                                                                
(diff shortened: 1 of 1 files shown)


components/scream/src/physics/rrtmgp/CMakeLists.txt
```diff
@@ -14,6 +14,12 @@ add_subdirectory(${SCREAM_BASE_DIR}/../../externals/YAKL ${CMAKE_BINARY_DIR}/ext
 target_compile_options(yakl PUBLIC $<$<COMPILE_LANGUAGE:CXX>:${YAKL_CXX_FLAGS}>)
 EkatDisableAllWarning(yakl)
 
+# For debug builds, set -DYAKL_DEBUG                                                                                       
+string(TOLOWER "${CMAKE_BUILD_TYPE}" CMAKE_BUILD_TYPE_ci)
```
bartgol, line 18: FYI, we are already computing this in scream's main CMakeLists.txt, storing the result in `SCRAM_DEBUG`.

bartgol (approved):
Looks fine. I pointed out there's already a var in scream storing whether it is a debug build or not, but there's nothing wrong with doing the same thing here.

components/scream/src/physics/rrtmgp/CMakeLists.txt
brhillman: Is it better to use SCREAM_DEBUG here then? I thought it might be more appropriate to use the CMAKE intrinsic CMAKE_BUILD_TYPE directly, rather than relying on the SCREAM-specific flag to make this more portable.

PR #1668, opened 2022-05-23
CIME fixes on Summit: Additional queues and explicit env vars

sarats:
This resolves the following issues:

* Unable to select debug queue
* Environment variables are set multiple times with the last invocation doing the right option. Avoided by changing the CPU env vars conditional on compilers.

We should merge into SCREAM repo for expediency and resolve conflicts with upstream E3SM master in the future.

[BFB]


bartgol (approved):
I was wondering why xmlchange would not allow me to change JOB_QUEUE to debug. Thanks for the fixes!

PR #1669, opened 2022-05-24
Fix atmosphere diagnostic class

bartgol:
The main change of this PR is to allow calling `get_diagnostic` without computing the diagnostic. The get method should only return a handle to the Field computed. If the user needs to compute the diagnostic, they can call `run`.

Another minor mod is to remove the dt argument from `run`, since it's pointless and never used. Technically, the new method `void run()` is simply _shadowing_ the base class one. If one access a diagnostic via pointer to the base class, the base class method will be available. But for the main practical purpose of diagnostics (i.e., I/O), the user will have a pointer to an AtmosphereDiagnostic, which is fine.

### DIFF
--- a/components/scream/src/diagnostics/tests/potential_temperature_test.cpp
+++ b/components/scream/src/diagnostics/tests/potential_temperature_test.cpp
@@ -101,7 +101,6 @@ void run(std::mt19937_64& engine)
 
   using KT         = ekat::KokkosTypes<DeviceT>;
   using ExecSpace  = typename KT::ExeSpace;
-  using TeamPolicy = typename KT::TeamPolicy;
   using MemberType = typename KT::MemberType;
   using view_1d    = typename KT::template view_1d<ScalarT>;
   using rview_1d   = typename KT::template view_1d<RealType>;
@@ -112,9 +111,6 @@ void run(std::mt19937_64& engine)
 
   constexpr int num_levs = 32; // Number of levels to use for tests.
   const     int num_mid_packs = pack_info::num_packs(num_levs);
-
-  using Check = ChecksHelpers<ScalarT,num_levs>;
-
 
   // A world comm
   ekat::Comm comm(MPI_COMM_WORLD);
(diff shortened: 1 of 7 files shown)


ambrad:
@bartgol this isn't a strong push back but more of just a check. I'm a little worried about shadowing methods since I think a lot of the time folks are used to either calling a unique fn or calling a virtual method. It might be easy to miss that one is calling a method whose resolution depends on whether one has the concrete or base type in the calling context. I've also noticed you're using the pattern of non-virtual `run` and then virtual `run_impl`, with I think so far the assumption that only the base class has `run`. Is it possible you might be breaking an established pattern that could lead to confusion later?

bartgol:
> @bartgol this isn't a strong push back but more of just a check. I'm a little worried about shadowing methods since I think a lot of the time folks are used to either calling a unique fn or calling a virtual method. It might be easy to miss that one is calling a method whose resolution depends on whether one has the concrete or base type in the calling context. I've also noticed you're using the pattern of non-virtual `run` and then virtual `run_impl`, with I think so far the assumption that only the base class has `run`. Is it possible you might be breaking an established pattern that could lead to confusion later?

That's a fair point. I am not fully convinced of my approach, so might be worth to discuss it a bit.

The AtmDiag class inherits from AtmProc out of convenience, so that lots of the code for initializing and interacting with it is already present. However, a diagnostic does not require a dt to be computed, and it makes no sense (and could be confusing) to force the user to pass one. Hence, my approach of shadowing. Which, as you point out, could still cause confusion downstream.

Perhaps the bad choice was to have AtmDiag inheriting from AtmProc in the first place. A diagnostic is virtually only used in IO and possibly (not right now, but I can see how it could help) inside surface coupling. Having it implemented as if it was an atm process might be the bad decision here.

Perhaps, it would be worth implementing the diagnostic as its own class. It might cause some code duplication with AtmProc, but I suspect not a ton (lots of code in AtmProc does not make sense in AtmDiag).

I will revert the shadowing by adding a `using`, leaving the overload with no inputs. Assuming people use only the `run()` method instead of `run(dt)`, if/when we revisit the base class, we should have a seamless transition from the point of view of the customer of the class. And if we don't revisit the design, at least we won't break anything.

bartgol:
> 4\. It follows also that we may foresee wanting to be able to set pre/post condition checks on a diagnostic.

No, the default is to just grab the diagnostic field when `get_diagnostic` is called. If you look at the output class mods, the `get_field` method takes a bool `eval_diagnostic` specifying whether the diagnostic should be run (if the field ends up being a diagnostic). That defaults to `false`, but during the output class `run` method, `get_field` is called with `eval_diagnostic=true`, which forces diagnostic evaluation (if the field is a diagnostic and not a FM field).

AaronDonahue (approved):

Issue #1506, opened 2022-03-24
List of unit tests that fail with specific random seeds


AaronDonahue:
The shoc_tests test suite fails for the `sp` test suite on `lassen` when the random seed is 376024247.

The fail occurs in the `update_prognostics_implicit_bfb` TEST_CASE with the following:

```
SCREAM_SOURCE/components/scream/src/physics/shoc/tests/shoc_update_prognostics_implicit_tests.cpp:394: FAILED:
  REQUIRE( d_f90.thetal[k] == d_cxx.thetal[k] )
with expansion:
  nanf == nanf
```

The line to recreate is:

`./shoc_tests --rng-seed 376024247`

bartgol:
Have you tried running `./shoc_tests --rng-seed 376024247` on another machine, to verify the problem are not the inputs? Might be that's a very unlucky inputs combination. You could also verify the same thing on lassen, but with a dp build.

AaronDonahue:
@bartgol , just tried... 
- `lassen` debug build passes
- `quartz` debug build passes
- `quartz` full_sp FAILS
So it looks like a bad seed for SP builds.

AaronDonahue:
Adding to this issue, `p3_main` test fails on blake for the `sp` test suite with seed 653534356

```

 For test p3_main, random number generator seed: 653534356
    Note: catch rng seed was 0 (default). We interpret that as a request to pick a random seed.
    To reproduce a previous run, use --rng-seed N to provide the rng seed.


~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
p3_tests is a Catch v2.13.8 host application.
Run with -? for options

-------------------------------------------------------------------------------
p3_main
-------------------------------------------------------------------------------
/home/e3sm-jenkins/blake/workspace/SCREAM_NightlyTests_Blake/scream/components/scream/src/physics/p3/tests/p3_main_unit_tests.cpp:463
...............................................................................

/home/e3sm-jenkins/blake/workspace/SCREAM_NightlyTests_Blake/scream/components/scream/src/physics/p3/tests/p3_main_unit_tests.cpp:218: FAILED:
  REQUIRE( isds_fortran[i].qi[k] == isds_cxx[i].qi[k] )
with expansion:
  0.0f == 0.0f
```

see the CDash report for May 19, 2022 for more details.

[...]

PR #1672, opened 2022-05-24
Change machine names for Perlmutter

ndkeen:
perlmutter -> pm-gpu
and now we have a new machine: pm-cpu


ndkeen:
Not sure why this failed

AaronDonahue:
For the fail on `blake` I'm going to guess you hit one of those random seeds that actually cause a fail.  I would retest and see what happens.  As for `mappy` it looks like that one just went off the rails.

ndkeen:
OK thanks.  I also just realized I neglected to `git add` the pm-cpu.cmake file.  I just did that -- does AT automatically detect and retest?

bartgol:
> OK thanks. I also just realized I neglected to `git add` the pm-cpu.cmake file. I just did that -- does AT automatically detect and retest?

Yes. Every time the sha of the branch changes, the AT tests the PR (regardless of whether it already passed or not).

bartgol:
> For the fail on `blake` I'm going to guess you hit one of those random seeds that actually cause a fail. I would retest and see what happens. As for `mappy` it looks like that one just went off the rails.

Aren't "those random seeds" becoming a bit more frequent lately?

AaronDonahue:
@bartgol , yes.  I've been keeping track of them in issue #1506 .  I think we will want to figure out what is going on there sooner or later.

bartgol (approved):

PR #1673, opened 2022-05-24
SCREAM: add cmake checks for SCREAM_INPUT_ROOT

bartgol:
* Error out if not present or not a directory
* Warn if the directory is not writable, since input files might have to be downloaded.

This should at least help users to write new machine files. If SCREAM_INPUT_ROOT does not get configured properly, cmake prints error/warning messages that can help the user to fix their configuration (or mach file).


ambrad (approved):

bartgol:
Skipping mappy/weaver since they passed and are the ones usually taking longer. I will remove the labels if pushing new commits.

PR #1677, opened 2022-05-25
SCREAM: allow to store special name for partitioned dim in a grid

bartgol:
This allows to use different dim names for pgN and GLL physics grids, so that we can store both fields in the same nc file.

I manually tested this by hacking a bit the dyn_grid_io test, to spoof pg2 data (dof gids, area, lat/lon), and verified the generated nc file has ncols_d instead of ncols as dim name.

Until pg2 is actually supported, this will have little to no impact. But since I had an idea in mind of how to handle `ncols` vs `ncols_d` I decided to flesh it out now.

If AT chokes on this, we can set it to WIP.

### DIFF
--- a/components/scream/src/dynamics/homme/homme_grids_manager.cpp
+++ b/components/scream/src/dynamics/homme/homme_grids_manager.cpp
@@ -119,6 +119,14 @@ build_grids ()
 
   this->alias_grid(pg_name,"Physics");
 
+  // If pg_type is not GLL or physics uses rebalance, we need to change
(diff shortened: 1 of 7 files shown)


ambrad:
I suggest closing this and waiting until the pg2 PR goes in to redesign things.

bartgol:
Once #1877 is merged, this becomes a 2 (simple) commits PR, and should allow outputing dyn grid fields in case pg2 is used for physics.

bartgol:
Closing, since @tcclevenger might simply pick what's needed out of this PR for his topo-file work.

PR #1684, opened 2022-05-27
Add configuration for ne256np4_oRRS18to6v3 v0 grid

brhillman:
Add configuration for ne256np4 bigrid with oRRS16to6v3 ocean/ice grid for SCREAMv0. Note that this currently needs an increase on the tolerance of `EPS_AGRID=1e-9`.

[B4B]

### DIFF
--- a/cime_config/config_grids.xml
+++ b/cime_config/config_grids.xml
@@ -1427,6 +1427,16 @@
       <mask>tx0.1v2</mask>
     </model_grid>
 
+    <model_grid alias="ne256np4_oRRS18to6v3">
+      <grid name="atm">ne256np4</grid>
+      <grid name="lnd">ne256np4</grid>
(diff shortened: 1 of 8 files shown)


PeterCaldwell:
What is the physical meaning of eps_agrid?

brhillman:
> What is the physical meaning of eps_agrid?

It’s the tolerance on the atm grid coordinate check. I think the check is to see if the coordinates from the initial condition match the coordinates from the domain file mapped back to the atmosphere grid? I’m not 100% sure. @wlin7 probably knows more.

brhillman:
I think the need for larger eps_agrid is due to a known issue: https://github.com/E3SM-Project/E3SM/issues/4437

ndkeen (approved):
Would need to test to make comment

PR #1685, opened 2022-05-27
Add default rad_frequency to v1 and updates to v0 scream nml

crterai:
This PR 
* adds `rrtmgp::rad_frequency` to `namelist_defaults_scream.xml` and provides default values for ne4, ne30, ne120, ne256, ne512, and ne1024 simulations. 
* makes consistent changes to scream v0 compsets (at least for higher rez sims)
* makes the se_tstep changes consistent with updated NCPL for ne1024.

### DIFF
--- a/components/eam/bld/namelist_files/use_cases/2010_scream_hr.xml
+++ b/components/eam/bld/namelist_files/use_cases/2010_scream_hr.xml
@@ -113,8 +113,12 @@
 <prescribed_aero_cycle_yr>01</prescribed_aero_cycle_yr>
 
 <!-- Do radiation every five minutes for ne1024 -->
-<iradsw hgrid="ne1024np4"> 4 </iradsw>
-<iradlw hgrid="ne1024np4"> 4 </iradlw>
+<iradsw hgrid="ne256np4"> 3 </iradsw>
+<iradlw hgrid="ne256np4"> 3 </iradlw>
+<iradsw hgrid="ne512np4"> 3 </iradsw>
+<iradlw hgrid="ne512np4"> 3 </iradlw>
+<iradsw hgrid="ne1024np4"> 3 </iradsw>
+<iradlw hgrid="ne1024np4"> 3 </iradlw>
 
 <!-- Settings related to the SCREAM NH dycore -->
 <theta_hydrostatic_mode>.false.</theta_hydrostatic_mode>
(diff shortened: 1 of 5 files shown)


PeterCaldwell:
Awesome, thanks for taking the initiative on this Chris!

components/eam/bld/namelist_files/use_cases/2010_scream_hr.xml
```diff
@@ -123,6 +127,11 @@
 <se_tstep hgrid="ne256np4"> 37.5 </se_tstep>
 <dt_tracer_factor hgrid="ne256np4"> 8 </dt_tracer_factor>
 <hypervis_subcycle_q hgrid="ne256np4"> 8 </hypervis_subcycle_q>
+<!-- Settings specific for ne1024 configuration -->
+<se_tstep hgrid="ne1024np4"> 8.333333333333333d0 </se_tstep>
+<dt_remap_factor hgrid="ne1024np4"> 6 </dt_remap_factor>
```
brhillman, line 132: I think dt_remap_factor should be 2, not 6

components/eam/bld/namelist_files/use_cases/2010_scream_hr_dyamond2.xml
```diff
@@ -129,6 +133,11 @@
 <se_tstep hgrid="ne256np4"> 37.5 </se_tstep>
 <dt_tracer_factor hgrid="ne256np4"> 8 </dt_tracer_factor>
 <hypervis_subcycle_q hgrid="ne256np4"> 8 </hypervis_subcycle_q>
+<!-- Settings specific for ne1024 configuration -->
+<se_tstep hgrid="ne1024np4"> 8.333333333333333d0 </se_tstep>
+<dt_remap_factor hgrid="ne1024np4"> 6 </dt_remap_factor>
+<dt_tracer_factor hgrid="ne1024np4"> 6 </dt_tracer_factor>
```
brhillman, line 139: Same here. But actually, these should go in `namelist_defaults_eam.xml` I think, not in the use case files.

components/eam/bld/namelist_files/use_cases/2010_scream_hr.xml
```diff
@@ -113,8 +113,12 @@
 <prescribed_aero_cycle_yr>01</prescribed_aero_cycle_yr>
 
 <!-- Do radiation every five minutes for ne1024 -->
-<iradsw hgrid="ne1024np4"> 4 </iradsw>
-<iradlw hgrid="ne1024np4"> 4 </iradlw>
+<iradsw hgrid="ne256np4"> 3 </iradsw>
```
brhillman, line 116: Actually, these should probably all go in namelist_defaults_eam.xml too now that I think about it; then we only have to change them in one place.

crterai: Yeah, having less copying and pasting makes sense.

components/eam/bld/namelist_files/use_cases/2010_scream_hr.xml
crterai: taking the whole se_tstep out since, #1686 will do that in namelist_defaults_eam.xml

brhillman (approved):
Looks great!

bartgol:
Adding retest, since master baselines have changed, but I suspect that the diffs are expected, no? If so, should we just merge away?

bartgol:
@crterai This PR is listed as Non-BFB, but the AT passed on both maps (standalone, v0 and v1) as well as weaver (standalone only), while Blake got some connection issue this round. Is the PR really actually BFB, or is the PASS unexpected (and therefore needs to be investigated)?

bartgol:
Ah, I see now that almost all changes affect only CIME coupled runs, for ne30 and higher resolutions. The only change mentioning ne4 sets rad frequency to 1, which I assume is what was already used. If that's the case, then forget about my previous comment.

PR #1689, opened 2022-05-27
Fast-forwarding EKAT to latest rev.

jeff-cohere:
This is just an update to SCREAM's EKAT submodule that hopefully allows it to build under less restrictive circumstances (see [EKAT PR 221](https://github.com/E3SM-Project/EKAT/pull/221) for details).


tcclevenger (approved):
Sorry, got pulled away after the meeting and forgot.

PeterCaldwell:
huh. I think the blake monolith fail is expected/due to a known issue on master. The mappy ERS and ERP diffs indicate that either this PR is answer changing - which I didn't expect - or new baselines are needed. @AaronDonahue - can you poke at this and see if you think these fails are justified? I'd like this PR to go in because it will help us with running on Summit.

bartgol:
> huh. I think the blake monolith fail is expected/due to a known issue on master. The mappy ERS and ERP diffs indicate that either this PR is answer changing - which I didn't expect - or new baselines are needed. @AaronDonahue - can you poke at this and see if you think these fails are justified? I'd like this PR to go in because it will help us with running on Summit.

#1647 went in 3d ago, and master baselines were not blessed afterwards (our nightlies are failing too). I'll bless and retest.

PR #1693, opened 2022-05-27
This commit fixes the restart_vs_monolithic unit test.

AaronDonahue:
Restarted runs in the test inadvertantly overwrote the rpointer.atm
file so when tests were run in parallel they would sometimes fail.

### DIFF
--- a/components/scream/tests/coupled/dynamics_physics/model_restart/input_restarted.yaml
+++ b/components/scream/tests/coupled/dynamics_physics/model_restart/input_restarted.yaml
@@ -50,9 +50,4 @@ Grids Manager:
 # List all the yaml files with the output parameters
 Scorpio:
   Output YAML Files: ["model_restart_output.yaml"]
-  Model Restart:
-    Casename: model_restart
-    Output Control:
-      Frequency:       1
-      Frequency Units: Steps
 ...


AaronDonahue:
@singhbalwinder - this PR should fix the test that was failing.

singhbalwinder:
Thanks @AaronDonahue !

bartgol:
These changes do fix the problem, but I think they are actually shadowing it. What you probably need is to set `MPI Ranks in Filename` in the `Output Control` section, so that different tests write different names in the rpointer file. The restart test will then try to match the name _including_ the number of ranks, so that the np1 and np4 tests filename inside rpointer will be different.

Right now, np1 and np4 are generating the same nc files, which means one could override the file generated by the other. Sure, other tests are already ensuring np1 is bfb with np4, but we should not mix the tests.

I think the mod in this PR is ok (the restarted run does not need to write a restart file), but we should also make it so that np1 and np4 tests generated _different_ restart files, by asking that MPI ranks are _in_ the filename.

AaronDonahue:
> These changes do fix the problem, but I think they are actually shadowing it. What you probably need is to set `MPI Ranks in Filename` in the `Output Control` section, so that different tests write different names in the rpointer file. The restart test will then try to match the name _including_ the number of ranks, so that the np1 and np4 tests filename inside rpointer will be different.
> 
> Right now, np1 and np4 are generating the same nc files, which means one could override the file generated by the other. Sure, other tests are already ensuring np1 is bfb with np4, but we should not mix the tests.
> 
> I think the mod in this PR is ok (the restarted run does not need to write a restart file), but we should also make it so that np1 and np4 tests generated _different_ restart files, by asking that MPI ranks are _in_ the filename.

@bartgol , good suggestion, I'll go ahead and make that change and then let you give it the final review so it can merge.

AaronDonahue:
@luca, actually... won't that mean that the `rpointer.atm` file will still get polluted with different filenames for a restart file.  That would mean that np1 could potentially restart from an np4 restart file just by the luck of the draw.

bartgol:
> @luca, actually... won't that mean that the `rpointer.atm` file will still get polluted with different filenames for a restart file? That would mean that np1 could potentially restart from an np4 restart file just by the luck of the draw. It seems like having these runs be sequential is the only safe bet.

IIRC, the restart logic checks that in the string `npX` the number X matches the current comm size. This allows np1 and np4 tests to run at the same time (though it would not allow to have np4 restarted from np1 rest file).

AaronDonahue:
But then the result would be throwing an error right?  `np1` could presumably read the `rpointer.atm` file and see a file name with `np4` in it.  Or vice versa.  Since both cases are writing to the same `rpointer.atm` file, and only the top entry is used.

AaronDonahue:
the current issue is that since the `restart` run is writing a restart there is a chance that the `rpointer.atm` file will have at the top of it a filename that isn't correct for the test.  In this case, you can have one of the tests read a restart created by a separate restarted run.  This fixes that case, but putting MPI ranks in the name of the restart file itself should create a similar issue.

bartgol:
I stand corrected. I thought I added that logic in the utility `find_filename_in_rpointer`, but I did not. Therefore, what you said is true.

That said, the np1 and np4 restart file _should_ be the same (we do test np1 vs np4 output in the homme+physics tests), so whether we read the np1 or np4 file should not matter. But it could, in a way, check for two things at the same time, which is unhealthy.

You can set the tests to be mutually exclusive, but I don't know how to ensure that both initial_np1 and restarted_np1 run _before_ their np4 counterparts (or vice versa), unless we make the np4 tests _depend_ on the np1 ones...

bartgol (approved):

PR #1696, opened 2022-05-30
Do not overwrite default output yaml file if present

### LINKED ISSUE #1683: screamv1 output choices get reset when submitting run in default
When submitting a v1 simulation, even if changes are made to the `./run/data/scream_default_output.yaml` file before submitting, they currently (`f1c36273`) get rewritten with the default choices, so they aren't reflected in the simulation. 

Current workaround is to copy over a different named output file, `./run/data/scream_output.yaml` and also changing the OUTPUT__YAML__FILE with `./atmchange Scorpio::Output__YAML__Files=data/scream_output.yaml so that it looks for the `scream_output.yaml` that was copied over.

bartgol:
Do not grab default output from src dir every time buildnml runs, so that we do not overwrite changes to the file in rundir/data (if already present) that the user may have done.

WARNING: this means that CIME vars in the default output yaml file are only expanded at case.setup phase. If the user then does some xmlchange, the changes are not reflected on the output yaml file. IMHO, that's ok. The file `scream_default_output.yaml` is intended to be used in our tests, as well as an "exemplar" of how to create a yaml file to request some output.

Note: we still do expand CIME vars in the output yaml files, every time that buildnml runs. However,this only matters if the user edits the file to contain some more CIME vars.

Fix #1683

### DIFF
--- a/components/scream/cime_config/buildnml
+++ b/components/scream/cime_config/buildnml
@@ -600,24 +600,14 @@ def buildnml(case, caseroot, compname):
 
     rundir     = case.get_value("RUNDIR")
     screamroot = os.path.join(case.get_value("SRCROOT"), "components/scream")
-    srcdata    = os.path.join(screamroot,"data")
     rundata    = os.path.join(rundir, "data")
 
     #
-    # Copy scream/data to rundir/data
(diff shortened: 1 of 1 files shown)


crterai (approved):
Looks good to me. Thanks for making this change!

PR #1702, opened 2022-05-31
SCREAM: remove some stranded calls to timers routines in homme interface

### LINKED ISSUE #1692: Timing files: Misc issues
Starting this thread as I think SCREAM handles timing files a bit differently. 

I see HommeTime.0 and HommeTime_stats files but we don't have the corresponding 
e3sm_timing and e3sm_timing_stats that are written out by default E3SM.

I have various jobs that sometimes appeared to hang while writing timing checkpoints as well as the directory `timing/checkpoints` does not exist. 

Other jobs print an error at the end and perhaps related cause a job to not release resources even after successful completion. Just a guess, there might be an issue with post-run provenance as well.
```
No timing file found in run directory
ERROR: /gpfs/alpine/cli115/scratch/sarat/repos/scream/cime/scripts/ne30-64nodes/timing/e3sm_timing_stats.2093349.220527-191932 does not exists
```

bartgol:
These calls were copied from standalone homme tests. In SCREAM, we should not handle timers and timing files from within homme interfaces.

I'm not sure if this addresses #1692 completely or only partially. It definitely addresses two points (`t_stopf` without corresponding `t_startf` and `HommeTime` file). But I don't think it will address the lack of `timing/e3sm_timing_stats.xyz` file.

Still, marking as
Closes #1692.

### DIFF
--- a/components/scream/src/dynamics/homme/interface/homme_driver_mod.F90
+++ b/components/scream/src/dynamics/homme/interface/homme_driver_mod.F90
@@ -7,7 +7,6 @@ module homme_driver_mod
   use iso_c_binding, only: c_ptr, c_int, c_double, c_bool, C_NULL_CHAR
 
   use parallel_mod,  only: abortmp
-  use perf_mod,      only: t_initf, t_finalizef, t_prf, t_startf, t_stopf
 
   implicit none
   private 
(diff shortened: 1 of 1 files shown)


sarats (approved):
Thanks! This makes sense. Hopefully with this fix, the expected e3sm timing files get finalized.

Did you happen to run a test?

bartgol:
> Thanks! This makes sense. Hopefully with this fix, the expected e3sm timing files get finalized.
> 
> Did you happen to run a test?

No, I did not. My thought was "let's check what happens in the Autotester runs"...

PeterCaldwell:
> My thought was "let's check what happens in the Autotester runs"...

I think we need to test on Summit since it is the only machine that fails to exit properly?

bartgol:
Our Perlmutter nightly tests were not running, due to a change in the machine name about 10 days ago or so. I don't actually think they ever successfully built and run a scream case on PM, to be honest.

Issue #1676, opened 2022-05-24
v1 initial run is not BFB-invariant to PE layout


AaronDonahue:
The homme_shoc_cld_spa_p3_rrtmgp test in the SCREAM unit test suite fails when comparing solutions from running with a different number of ranks, when the grid is switched to ne4.

Note this does not have the same fails when the grid is the default ne2.

AaronDonahue:
Testing master with hash: `11787720eb1936d5ee2963097885164bca30e087` (May 24th) and switching the `homme_shoc_cld_spa_p3_rrtmgp` test to ne4 yields 

```
The following tests FAILED:
	202 - homme_shoc_cld_spa_p3_rrtmgp_np1_vs_np3_bfb (Failed)
	203 - homme_shoc_cld_spa_p3_rrtmgp_np1_vs_np4_bfb (Failed)
```

Note, this means that np1 vs. np2 passed.  I confirmed this manually.

AaronDonahue:
Putting the diffs to master here for reproducibility:
```
diff --git a/components/scream/tests/coupled/dynamics_physics/homme_shoc_cld_spa_p3_rrtmgp/CMakeLists.txt b/components/scream/tests/coupled/dynamics_physics/homme_shoc_cld_spa_p3_rrtmgp/CMakeLists.txt
index dd045cc..c21e536 100644
--- a/components/scream/tests/coupled/dynamics_physics/homme_shoc_cld_spa_p3_rrtmgp/CMakeLists.txt
+++ b/components/scream/tests/coupled/dynamics_physics/homme_shoc_cld_spa_p3_rrtmgp/CMakeLists.txt
@@ -29,7 +29,7 @@ configure_file(${CMAKE_CURRENT_SOURCE_DIR}/homme_shoc_cld_spa_p3_rrtmgp_output.y
 
 # Set homme's test options, so that we can configure the namelist correctly
 # Discretization/algorithm settings
-set (HOMME_TEST_NE 2)
+set (HOMME_TEST_NE 4)
 set (HOMME_TEST_LIM 9)
 set (HOMME_TEST_REMAP_FACTOR 3)
 set (HOMME_TEST_TRACERS_FACTOR 1)
@@ -64,7 +64,7 @@ set (TEST_INPUT_FILES
   init/spa_init_ne2np4.nc
   init/map_ne4np4_to_ne2np4_mono.nc
   init/spa_file_unified_and_complete_ne4_20220428.nc
[...]

PR #1703, opened 2022-06-01
Fixes an undefined variable in shoc interface

singhbalwinder:


bartgol (approved):

bartgol:
@jgfouca I'm not sure what the phase `phase COMPARE_base_rest` means, and whether it can be caused by non BFB changes to the source code. The name sounds like a fail when comparing output of peLayout1 with that of peLayout2, which is odd.

@singhbalwinder was the previous pr (the wet/dry stuff in shoc) supposed to cause nonBFB diffs in v0 tests?

AaronDonahue:
My guess is that the phrase `phase COMPARE_base_rest` is a fail when comparing restarted vs monolithic since that phrase pops up for the ERP tests.  Is that a test that changes the number of ranks?  If so then it may be due to the discussion in issue #1676 .

As for the non-BFB, I'm surprised this is not non-BFB across the board for SCREAMv0.  It looks like it only is for one of the many tests.  @singhbalwinder , any thoughts?

singhbalwinder:
It should definitely be non-BFB for the tests which invoke SHOC, which I am assuming would include all the v0 CIME tests. I was not able to run CIME tests on Compy and Summit so far but I will try them on Cori to investigate this.

jgfouca:
@bartgol , `COMPARE_base_rest` should not be impacted by non-bfb changes because it is not a baseline check. It is checking that a restarted run and a non-restart run produce the exact same results.

AaronDonahue (approved):

AaronDonahue:
@jgfouca , I have a question before I manually merge this PR.   Do ERP tests change PE-Layout?  I know ERS doesn't, but isn't that what the "P" in the test name refers to?

AaronDonahue:
Follow-up question, we are seeing DIFFs in the SCREAMv1 tests.  This PR shouldn't be non-BFB in v1.  I suspect this is because #1706 went in so now `master` is non-BFB with the baselines right?  Is there a way to confirm that the DIFFs we are seeing in the AT results are not due to this PR?

AaronDonahue:
> So, yes, it looks like it uses half the PEs for one of the runs.

Okay, then these fails are probably related to issue #1676 .

singhbalwinder:
@AaronDonahue : To find out if v1 tests are failing due to #1706 or not, I can merge master into this PR. That way, this PR will include non-BFB effects of 1706.

AaronDonahue:
@singhbalwinder , I think that's what the AT already does.  I think the baseline comparison is against the baselines that are saved on mappy, the ones we occasionally bless after the nightlies come in.

singhbalwinder:
Ok, thanks! That makes sense.

bartgol:
@AaronDonahue v1 tests were passing in the second to last AT run. The fact that they fail now is probably due to something else that got merged in master between the two runs.

singhbalwinder:
> As for the non-BFB, I'm surprised this is not non-BFB across the board for SCREAMv0. It looks like it only is for one of the many tests. @singhbalwinder , any thoughts?

Just for the record, I ran SMS test with and without shoc's wet<->dry conversion (PR #1515), and the baseline comparisons showed a FAIL as expected (i.e. adding wet<->dry conversions do change the model answers):
```
PASS SMS.ne4pg2_ne4pg2.F2010-SCREAM-LR.cori-knl_intel CREATE_NEWCASE
PASS SMS.ne4pg2_ne4pg2.F2010-SCREAM-LR.cori-knl_intel XML
PASS SMS.ne4pg2_ne4pg2.F2010-SCREAM-LR.cori-knl_intel SETUP
PASS SMS.ne4pg2_ne4pg2.F2010-SCREAM-LR.cori-knl_intel SHAREDLIB_BUILD time=39
PASS SMS.ne4pg2_ne4pg2.F2010-SCREAM-LR.cori-knl_intel NLCOMP
PASS SMS.ne4pg2_ne4pg2.F2010-SCREAM-LR.cori-knl_intel MODEL_BUILD time=141
PASS SMS.ne4pg2_ne4pg2.F2010-SCREAM-LR.cori-knl_intel SUBMIT
PASS SMS.ne4pg2_ne4pg2.F2010-SCREAM-LR.cori-knl_intel RUN time=268
FAIL SMS.ne4pg2_ne4pg2.F2010-SCREAM-LR.cori-knl_intel BASELINE shoc_wet_dry_baseline: DIFF
PASS SMS.ne4pg2_ne4pg2.F2010-SCREAM-LR.cori-knl_intel MEMCOMP
PASS SMS.ne4pg2_ne4pg2.F2010-SCREAM-LR.cori-knl_intel TPUTCOMP
PASS SMS.ne4pg2_ne4pg2.F2010-SCREAM-LR.cori-knl_intel MEMLEAK
PASS SMS.ne4pg2_ne4pg2.F2010-SCREAM-LR.cori-knl_intel SHORT_TERM_ARCHIVER
```

Issue #1374, opened 2022-01-21
Bad interaction between CIME cmake system and SCREAM cmake machine cache files


jgfouca:
Tagging @bartgol , @AaronDonahue , and @PeterCaldwell so you see this.

This issue was uncovered by Aaron here: https://acme-climate.atlassian.net/wiki/spaces/NGDNA/pages/3243900990/SCREAMv1+runs+on+LC+quartz

Notice in step 2, he uncached a number of cmake settings. This fixed the problem (the `CMAKE_CXX_FLAGS` were not being added to compilations) because, when building from CIME, the SCREAM machine cache files are not loaded as cache files normally are (via the `-C` flag); instead, they are included here in `build_scream.cmake`:

```
function(build_scream)

  if (COMP_NAMES MATCHES ".*scream.*")

    # Include machine file here                                                                                                                                                                                                                   
    message("Found scream component")
    include(${CMAKE_SOURCE_DIR}/scream/cmake/machine-files/${MACH}.cmake)

    add_subdirectory("scream")
  endif()

endfunction(build_scream)
```

The consequence of this approach is that any CACHE variable that's already been set (by CIME) will not be reset in the scream MACH.cmake unless you add `FORCE` to the set.

Proposals:
1) Quick fix: We look at our SCREAM machine cache files and add `FORCE` to any cache variable that have been set by CIME. I believe it is the intent of SCREAM to manage its own flags not really on anything coming from E3SM, so this should be a viable approach. The downside is that we'll need to forever keep this in mind when making changes to our SCREAM machine files and be aware of the potential need to add `FORCE`. This will generally only be needed for CMAKE*FLAGS variables. Currently, only syrah and quartz need these in their machine files.
2) Big refactor: we (SCREAM) rely on E3SM to provide flag information. This probably means we also rely on E3SM's build of Kokkos instead of our own.

Since 1) seems to be quite easy, I will submit a PR that does it while we consider 2).

bartgol:
@jgfouca Another option is to have the mach files set `SCREAM_CXX_FLAGS` in the cache. Later, we then append scream flags to cmake cxx flags.

[...]

PR #1704, opened 2022-06-01
Update to latest RRTMGPXX with better CPU performance

brhillman:
Update RRTMGP and YAKL submodules to take advantage of improvements for better CPU performance.

[B4B]


bartgol (approved):

AaronDonahue:
@brhillman , what is the status of this PR?  It looks like it was approved but failed the AT 22 days ago.  I added RETEST to see if it passes now.

PeterCaldwell:
This PR is failing because RRTMGPxx needs updated cuda and gcc modules to work with these perf enhancements. @tcclevenger is supposed to be working on that, but he might be done and needs Ben's help to fix the remaining issues?

We definitely want these fixes to go in.

tcclevenger:
> This PR is failing because RRTMGPxx needs updated cuda and gcc modules to work with these perf enhancements. @tcclevenger is supposed to be working on that, but he might be done and needs Ben's help to fix the remaining issues?

We know where these fail, but I don't yet have a solution (haven't got a change to look at it since Weaver came back online). I'll try and get to it next week.

PeterCaldwell:
Thanks Conrad. I think it would also be useful for @brhillman to look at it, if you can point him to your branch...

tcclevenger:
> Thanks Conrad. I think it would also be useful for @brhillman to look at it, if you can point him to your branch...

Here is the commit which updates to cuda11: https://github.com/E3SM-Project/scream/commit/4f373cebbb352b425e3c531ee3960d42b68b0e1f

ambrad:
I think one thing that should be done before this PR is brought in is to add the commit (with conflict resolution) to the new rte-rrtmgp version that gives us GPU determinism: https://github.com/E3SM-Project/rte-rrtmgp/commit/55782e43dab02bab9f90f37ba7cd68e7118d2c64. If we don't add that commit, all of our ERS, ERP, and PEM tests will fail on GPU machines.

PeterCaldwell:
Note that our plan is to pause on this task until after the Sept 1 simulations start.

AaronDonahue:
I'm adding the WIP label to signify that this is actively being worked on.  Feel free to drop the label when we resume it.

bartgol (dismissed):

jgfouca:
@brhillman , looks like this is failing AT in a few places.

brhillman:
Thanks @jgfouca I will take a look tonight

tcclevenger:
Couple of things:

1. The current weaver fail in this PR is from YAKL and EAMxx now having different signatures in `target_link_libraries()` CMake calls. 
Fix is here: 1df4ad5dff7d3dcfd6865df53a2cb14f0e239b26

2. With that fix, there will then be the issue that YAKL needs to run on nodes with updated drivers (supporting CUDA11), whereas the AT runs on `-q rhel7W` nodes (supporting CUDA10). @bartgol @jgfouca is it ok to let the AT use the `rhel8` nodes?
If so, fix here: 68912062e8a84c58c9ab49d7146dea4a27869cd5

3. Oddly enough, using these nodes is enough to have EAMxx build, run, and pass tests (I tested "ctest -R rrtmgp") without actually using CUDA11 (AT loads CUDA10). @bartgol @jgfouca Is there any reason not to update to CUDA11 on weaver? 
If not, update here: df26915114ca7e19ddf7f5921521e745761a0d29

brhillman:
Thanks @tcclevenger would you mind merging your fix to this branch? I actually thought @bartgol had made this fix on another branch that I thought was already merged to master but maybe not.

jgfouca:
@tcclevenger , the only reason I can think of to not use rhel8 is if there's higher general demand for those nodes which could result in longer waits in the queue. I don't know of any reason not to move to CUDA11.

bartgol:
rhel8 nodes are more used, but there's also more of them, while rhel7W only has 2 nodes. As for CUDA version, we already use cuda11 on PM, but we still use cuda10 on Summit, where most of our production runs currently are. While I don't see major issues switching, I also think it's nice to test all our PR with the same cuda version that production runs are using...

bartgol:
> 1. The current weaver fail in this PR is from YAKL and EAMxx now having different signatures in `target_link_libraries()` CMake calls.
>    Fix is here: [1df4ad5](https://github.com/E3SM-Project/scream/commit/1df4ad5dff7d3dcfd6865df53a2cb14f0e239b26)
>    Not a CMake expert, so I don't know if there are other reasons to need that `PUBLIC` tag, but I think that should be fine..

CMake does not allow mixing the bare and decorated interfaces (without and with PUBLIC/PRIVATE/INTERFACE respectively). I don't know what's the reason behind it, at least in the case with bare mixed with PUBLIC, since bare is _equivalent_ to PUBLIC. That said, given that not specifying anything is equivalent to using PUBLIC and we were using PUBLIC, removing it is fine.

bartgol:
> @bartgol is there any reason to use cuda10 over cuda11 on summit though? I didn’t realize we were still using cuda10 on summit and I think 11 should be available there as well.

Ah, I stand corrected. Cuda10 is used with pgi compilers. gnugpu uses cuda11. So yes, we _should_ switch to 11 also on weaver.

brhillman:
Great, let’s update to cuda11 on weaver and get this thing merged then

tcclevenger:
@brhillman Yeah, I'll merge it all in.

ambrad:
ERS_Ln362.ne30pg2_ne30pg2.F2010-SCREAMv1.ascent_gnugpu passes.

tcclevenger:
Seems that `AT: Skip blake` tag isn't working properly since it just timed out waiting for blake..

bartgol:
> Seems that `AT: Skip blake` tag isn't working properly since it just timed out waiting for blake..

that's odd. let me check our scripts.

tcclevenger:
I don't quite understand the weaver fail, but it looks like it has to do with the new modules and/or testing on rhel8.

jgfouca:
@tcclevenger , I can look

jgfouca:
Actually, it looks like the problem may be a bad interaction between our python run_cmd and module. The python-3 module loaded fine from weaver_setup.

jgfouca:
Ok, the problem appears to be that module setup is not being preserved by the bsub command. Modules are initialize by the login scripts.

Test:
```
% cat myscript
#! /bin/bash 

module list
```

```
% bsub -I -q rhel8 -n 1 ./myscript
...
./myscript: line 3: module: command not found
```

If I add `-l` to the bash #! (will cause login stuff to be run)
```
% bsub -I -q rhel8 -n 1 ./myscript
...
Currently Loaded Modulefiles:
 1) python/3.7.3   2) git/2.10.1 
```

There are a couple solutions that should be pretty easy that I will look into.

jgfouca:
I have added `skip mappy` since my commit cannot impact mappy.

jgfouca:
Hmm, now missing psutil. I can fix that i think

jgfouca:
Ok, the downside to adding `-L /bin/bash` to the bsub command is that all current modules are lost. So, instead of that, I'm just going to source the file that sets up the `module` function in the environment command for weaver.

ambrad:
Regarding this build error:
```
/home/projects/ppc64le-pwr9-nvidia/spack-installs/cuda/11.2.2/gcc/8.3.1/base/uflzqtv/include/thrust/system/cuda/config.h:78:2: error: #error The version of CUB in your include path is not compatible with this release of Thrust. CUB is now included in the CUDA Toolkit, so you no longer need to use your own checkout of CUB. Define THRUST_IGNORE_CUB_VERSION_CHECK to ignore this.
#error The version of CUB in your include path is not compatible with this release of Thrust. CUB is now included in the CUDA Toolkit, so you no longer need to use your own checkout of CUB. Define THRUST_IGNORE_CUB_VERSION_CHECK to ignore this.
^~~~~
```
I'm wondering if you need to actually merge the upstream PR first, then merge master into this PR. That's just a guess.

tcclevenger:
> Regarding this build error:
> 
> ```
> /home/projects/ppc64le-pwr9-nvidia/spack-installs/cuda/11.2.2/gcc/8.3.1/base/uflzqtv/include/thrust/system/cuda/config.h:78:2: error: #error The version of CUB in your include path is not compatible with this release of Thrust. CUB is now included in the CUDA Toolkit, so you no longer need to use your own checkout of CUB. Define THRUST_IGNORE_CUB_VERSION_CHECK to ignore this.
> #error The version of CUB in your include path is not compatible with this release of Thrust. CUB is now included in the CUDA Toolkit, so you no longer need to use your own checkout of CUB. Define THRUST_IGNORE_CUB_VERSION_CHECK to ignore this.
> ^~~~~
> ```
> 
> I'm wondering if you need to actually merge the upstream PR first (with the YAKL CMake fix one-liner), then merge master into this PR. That's just a guess.

That is strange. It is as if 
```set(CMAKE_CXX_FLAGS "-DTHRUST_IGNORE_CUB_VERSION_CHECK" CACHE STRING "" FORCE)```
in `weaver.cmake` isn't being applied..

**Comment by tcclevenger** [comment]
@ambrad That is the error one gets if `set(CMAKE_CXX_FLAGS "-DTHRUST_IGNORE_CUB_VERSION_CHECK" CACHE STRING "" FORCE)` isn't included.. It is as if it did not pick up the change in `weaver.cmake.

**Comment by ambrad** [comment]
@tcclevenger you can check the e3sm.bldlog file to see if it's being dropped. I've had issues in the past with getting the scream/ekat cmake layers to accept flag modifications in machine files. If the flag is indeed missing in e3sm.bldlog, @bartgol might have insight into what to do.

**Comment by jgfouca** [comment]
@ambrad , i am going to try merging the upstream PR into this branch.

**Comment by ambrad** [comment]
@jgfouca, OK, but I think Conrad has identified and is tracking down the true issue. My wild guess was wildly wrong.

**Comment by jgfouca** [comment]
@ambrad , it appears to be working. No build error yet 5 minutes in. Seems hard to imagine that $machine.cmake files are being ignored... we've been using those for years.

**Comment by ambrad** [comment]
@jgfouca OK. I don't think the machine file is being ignore, but I think CMAKE_SET_FLAGS is overwritten at some point and so is effectively ignored. In the CIME-level machine files, one sees e.g. this:
```cmake
string(APPEND CPPDEFS " -DTHRUST_IGNORE_CUB_VERSION_CHECK")
```
and this formulation seems to get through.

jgfouca:
The merge of the upstream PR seems to fix things, at least for dbg test.

tcclevenger:
So the build folder `/home/e3sm-jenkins/weaver/workspace/SCREAM_PullRequest_Autotester_Weaver/4063/scream/components/eamxx/ctest_build.` for the failed AT run is empty.. 

Doing manual testing, not including the CXX_FLAG definitely triggers the error. @jgfouca Did the dbg test completely build? In my experience the error is thrown a while after starting the build.

bartgol:
We don't use `CPPDEFS` in standalone builds, just like any other CIME flag (topic of discussion: should we?)

That said, we do have the macro set in our standalone mach files, but not for weaver:
```
$ git grep CUB_VERSION
cmake/machine-files/alvarez.cmake:set(CMAKE_CXX_FLAGS "-DTHRUST_IGNORE_CUB_VERSION_CHECK" CACHE STRING "" FORCE)
cmake/machine-files/ascent.cmake:set(CMAKE_CXX_FLAGS "-DTHRUST_IGNORE_CUB_VERSION_CHECK" CACHE STRING "" FORCE)
cmake/machine-files/gcp.cmake:set(CMAKE_CXX_FLAGS "-DTHRUST_IGNORE_CUB_VERSION_CHECK" CACHE STRING "" FORCE)
cmake/machine-files/pm-cpu.cmake:set(CMAKE_CXX_FLAGS "-DTHRUST_IGNORE_CUB_VERSION_CHECK" CACHE STRING "" FORCE)
cmake/machine-files/pm-gpu.cmake:set(CMAKE_CXX_FLAGS "-DTHRUST_IGNORE_CUB_VERSION_CHECK" CACHE STRING "" FORCE)
cmake/machine-files/summit.cmake:  set(CMAKE_CXX_FLAGS "-DTHRUST_IGNORE_CUB_VERSION_CHECK" CACHE STRING "" FORCE)
```

bartgol:
Btw, issue #1374 has been open for ages, and it might be time to address it. Maybe, we should just pulled the trigger, and make SCREAM handle mach files in the same way as e3sm (i.e., include them in the CMakeLists.txt, rather than preload them on cmd line via -c).

tcclevenger:
@bartgol It is added to weaver.cmake in the PR

jgfouca:
@bartgol 

```
% cat components/eamxx/cmake/machine-files/weaver.cmake 
set (EKAT_MACH_FILES_PATH ${CMAKE_CURRENT_LIST_DIR}/../../../../externals/ekat/cmake/machine-files)
include (${EKAT_MACH_FILES_PATH}/weaver.cmake)
set (BLAS_LIBRARIES /ascldap/users/projects/e3sm/scream/libs/openblas/install/weaver/gcc/8.5.0/lib/libopenblas.so CACHE STRING "")
set (LAPACK_LIBRARIES /ascldap/users/projects/e3sm/scream/libs/openblas/install/weaver/gcc/8.5.0/lib/libopenblas.so CACHE STRING "")
set(SCREAM_INPUT_ROOT "/home/projects/e3sm/scream/data" CACHE STRING "")
set(CMAKE_CXX_FLAGS "-DTHRUST_IGNORE_CUB_VERSION_CHECK" CACHE STRING "" FORCE) # <----------
```

jgfouca:
@tcclevenger , yes, it completely built and all the tests passed except one (baseline diff).

jgfouca:
OK, this is the diff I saw when I ran. Is this expected?

jgfouca:
Note, this PR now includes the upstream merge PR.

ambrad:
Is this diff against a baseline? As long as the C++ vs F90 is BFB, then all is well. A Cuda version change can certainly cause a diff against a baseline.

jgfouca:
@ambrad, yes:

```
./shoc_run_and_cmp_cxx -b /home/projects/e3sm/scream/pr-autotester/master-baselines/weaver/full_debug/data/shoc_run_and_cmp.baseline
```

ambrad:
@jgfouca or @tcclevenger if this isn't already established, would you generate a baseline with the F90 code and then run the C++/GPU version against it to verify BFBness. If that succeeds, this PR is ready to merge.

jgfouca:
@ambrad , sure. I assume debug build? I'm already testing that at the moment.

ambrad:
@jgfouca yes, debug.

jgfouca:
@ambrad , @bartgol , both p3_run_and_cmp_cxx and shoc_run_and_cmp_cxx are passing with baselines generated from fortran. The BFB unit tests are also all passing (also compares against fortran).

One confusing thing is that the p3_run_and_cmp_cxx is generating some concerning output:
```
23: qv > tol 117 times. Max rel diff= 0.2 normalized by ref impl val=0.05.
23: th_atm > tol 216 times. Max rel diff= 0.2 normalized by ref impl val=1079.54.
23: pres > tol 216 times. Max rel diff= 0.2 normalized by ref impl val=98711.1.
23: dz > tol 216 times. Max rel diff= 0.2 normalized by ref impl val=34903.9.
23: nccn_prescribed > tol 216 times. Max rel diff= 0.2 normalized by ref impl val=1e+08.
23: inv_qc_relvar > tol 216 times. Max rel diff= 0.2 normalized by ref impl val=1.
23: qc > tol 24 times. Max rel diff= 0.2 normalized by ref impl val=1.2016e-05.
23: nc > tol 24 times. Max rel diff= 0.2 normalized by ref impl val=9.99096e+07.
23: qr > tol 38 times. Max rel diff= 0.2 normalized by ref impl val=0.000187795.
23: nr > tol 38 times. Max rel diff= 0.2 normalized by ref impl val=135710.
23: qi > tol 52 times. Max rel diff= 0.2 normalized by ref impl val=0.000409097.
23: ni > tol 52 times. Max rel diff= 0.2 normalized by ref impl val=216385.
23: qm > tol 30 times. Max rel diff= 0.2 normalized by ref impl val=0.000385423.
23: bm > tol 30 times. Max rel diff= 0.2 normalized by ref impl val=4.28247e-07.
23: precip_liq_surf > tol 3 times. Max rel diff= 0.2 normalized by ref impl val=2.47897e-07.
23: diag_eff_radius_qc > tol 216 times. Max rel diff= 0.2 normalized by ref impl val=1e-05.
23: diag_eff_radius_qi > tol 216 times. Max rel diff= 0.2 normalized by ref impl val=0.000139314.
23: rho_qi > tol 52 times. Max rel diff= 0.2 normalized by ref impl val=900.
23: dpres > tol 216 times. Max rel diff= 0.2 normalized by ref impl val=1388.89.
23: inv_exner > tol 216 times. Max rel diff= 0.2 normalized by ref impl val=7.19721.
23: qv2qi_depos_tend > tol 38 times. Max rel diff= 0.2 normalized by ref impl val=6.88692e-07.
23: precip_liq_flux > tol 38 times. Max rel diff= 0.2 normalized by ref impl val=0.0120278.
23: cld_frac_r > tol 216 times. Max rel diff= 0.2 normalized by ref impl val=1.
23: cld_frac_l > tol 216 times. Max rel diff= 0.2 normalized by ref impl val=1.
23: cld_frac_i > tol 216 times. Max rel diff= 0.2 normalized by ref impl val=1.
23: liq_ice_exchange > tol 8 times. Max rel diff= 0.2 normalized by ref impl val=2.14424e-06.
23: vap_liq_exchange > tol 15 times. Max rel diff= 0.2 normalized by ref impl val=6.00805e-07.
23: vap_ice_exchange > tol 38 times. Max rel diff= 0.2 normalized by ref impl val=6.88692e-07.
23: qv_prev > tol 111 times. Max rel diff= 0.2 normalized by ref impl val=0.05.
23: t_prev > tol 216 times. Max rel diff= 0.2 normalized by ref impl val=299.396.
23: Ref impl failed.
```
... even though it is reporting pass.

jgfouca:
Disregard previous message. That was output coming from `./p3_run_and_cmp_cxx_fail` which my regex was picking up. I think this is good to merge.

jgfouca:
Just need someone to approve...

ambrad:
@jgfouca great, I agree. Merge when you're ready.

ambrad (approved):

PR #1706, opened 2022-06-02
Use non-hydrostatic mode by default for SCREAMv1

brhillman:
Use non-hydrostatic mode by default for SCREAMv1. As SCREAMv1 is meant to target higher (cloud-permitting) resolutions, and does not yet have a deep convective parameterization, it makes the most sense to use non-hydrostatic mode for these simulations by default. We might condition this on either compset or resolution in the future once we have a deep convective parameterization available for the lower resolutions.

[non-B4B] (for CIME cases).


mt5555 (approved):

bartgol:
V1 diffs are expected. I think you have the merge powers. If not, Aaron, Peter, I, Jim, and someone else too can help.

PR #1690, opened 2022-05-27
Summit machine configuration and jsrun fixes

sarats:
This PR fixes jsrun arguments (--nrs etc.) for small number of tasks. Previously, jsrun always tried to use 6 resource sets per node even when NTASKS<6. This is an immediate fix that helps debugging at smaller scale. 
I would plan for a more robust fix (other edge cases - non-multiples of 6) in the future.

This PR also includes the best known modules (gcc and cuda) and customizes various CIME settings specifically for SCREAM workloads. Specifically, we use 1 process per GPU (so CPU_PER_RS=GPU_PER_RS=1)

Testing:
I was able to get the latest master to build and run a ne30 case successfully on 32 nodes (192 tasks).

Following timers are `wallmax` from `HommeTime_stats` file.
```
Timer name   : seconds
CPL:ATM_RUN  : 64.616
CPL:RUN_LOOP : 73.535
```

jsrun:
```
jsrun -X 1 --nrs 192 --rs_per_host 6 --tasks_per_rs 1 -d plane:1 --cpu_per_rs 1 --gpu_per_rs 1 --bind packed:smt:1 --latency_priority gpu-cpu --stdio_mode prepended   --smpiargs="-gpu" /gpfs/alpine/cli115/proj-shared/sarat/e3sm_scratch/n1/bld/e3sm.exe   >> e3sm.log.$LID 2>&1
 
 ```
 
cc @PeterCaldwell @mt5555

PR #1711, opened 2022-06-02
Jgfouca/upstream merge 2022 06 02

jgfouca:
No conflicts.

I noticed this problem:
`FAIL ERS_Ln9.ne30_ne30.F2010-SCREAMv1.mappy_gnu9 (phase COMPARE_base_rest)`

But this appears to be on master too.

### DIFF
--- a/cime_config/machines/config_machines.xml
+++ b/cime_config/machines/config_machines.xml
@@ -3492,6 +3492,8 @@
     <MAX_TASKS_PER_NODE compiler="pgigpu">18</MAX_TASKS_PER_NODE>
     <MAX_TASKS_PER_NODE compiler="gnugpu">6</MAX_TASKS_PER_NODE>
     <MAX_TASKS_PER_NODE compiler="ibmgpu">42</MAX_TASKS_PER_NODE>
+	<!-- Need to activate following attribute after CIME update from upstream -->
+    <!-- <MAX_GPUS_PER_NODE>6</MAX_GPUS_PER_NODE> -->
     <MAX_MPITASKS_PER_NODE>84</MAX_MPITASKS_PER_NODE>
     <MAX_MPITASKS_PER_NODE compiler="pgigpu">18</MAX_MPITASKS_PER_NODE>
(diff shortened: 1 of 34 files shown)


sarats:
We have some customized settings for getting things running on Summit which will be overridden by upstream, so need to manage changes to config_machines.xml.

cime_config/machines/config_machines.xml
```diff
@@ -3492,6 +3492,8 @@
     <MAX_TASKS_PER_NODE compiler="pgigpu">18</MAX_TASKS_PER_NODE>
     <MAX_TASKS_PER_NODE compiler="gnugpu">6</MAX_TASKS_PER_NODE>
     <MAX_TASKS_PER_NODE compiler="ibmgpu">42</MAX_TASKS_PER_NODE>
+	<!-- Need to activate following attribute after CIME update from upstream -->
```
sarats, line 3495: Did I miss something? Are these 2 lines the only changes to the machine config?

jgfouca: Yes, it seems upstream did not do much to modify config_machines.xml since the last update.

jgfouca: Wasn't the config_machines change made to both SCREAM and E3SM? In that case, the change would already be in SCREAM and so would not show up here.

PeterCaldwell: I think Sarat's confusion is that he merged SCREAM PR #1690 the other day (https://github.com/E3SM-Project/scream/pull/1690/files) and it included a lot of changes to config_machines.xml which I would have thought would show up as diffs here. But maybe I'm just not understanding git merge.

bartgol: Yeah, I don't think that matters. The diffs you see here are the changes that _these_ commits add to current master. Whether upstream/HEAD and scream/HEAD had _lots_ of diffs is a separate issue.

sarats: Yes, my changes were the only ones to this file in both SCREAM and E3SM and looks like the SCREAM-specific changes survive (MAX_MPI_TASKS "gnugpu" = 6). So we're good.

sarats (approved):

Issue #1707, opened 2022-06-02
Implement "no aerosol" option in scream v1


hassanbeydoun:
v0 had an option of turning off aerosols by setting do_aerosol_rad and micro_aerosolactivation to false. The former bypasses aerosol calculations in radiation.F90 and sets aerosol optical properties to zeros while the latter sets nc to a constant value saved in p3_utils. Having this option in v1 will allow comparisons with v0 without SPA.  

The v1 p3 code still uses do_prescribed_CCN and do_predict_nc so we can set those to false and have our constant drop number option going. The radiation however does not seem to have a "do_aerosol_rad" equivalent so more work might be needed there. 

@AaronDonahue @brhillman @PeterCaldwell let me know if you have any thoughts. I'll post updates as I have them.

PeterCaldwell:
Additional context - I see this innocuous issue as the path forward for getting ne512 and ne1024 working in the short term (by bypassing the memory bottleneck in SPA init). So I think it is high priority. In my mind, this should be easy but requires changes to our user interface that I don't know how to do - basically, we need to create an option that bundles a couple of user interface options: do_prescribed_CCN=False, do_predict_nc=False, and aerosol optical properties should all be initialized to zero. We also need to specify the prescribed nc value we want to use... I'm not sure if that is possible yet. 

Do you have comments on how this could be done, @jgfouca , @bartgol , @AaronDonahue ?

AaronDonahue:
I've been chatting with @hassanbeydoun on Slack about this...  Some thoughts so far:
1. Adding `do_prescribed_CCN` and `do_predict_nc` as runtime options should be fairly straightforward.  They can just be added to the parameter list in the `namelist_defaults.xml` file for SCREAM.
2. We will need to put some checks into rrtmgp to essentially grab 0 valued arrays when aerosol is turned off.  Note, P3 already has some infrastructure for handling cases when aero is turned off so we just need to pass the correct logical there.
3. I think, that in cases where all of the `aero_blah_sw/lw` variables are going to be unused, we shouldn't even add them to the field manager.  Probably the same with `nccn`.  I think we will just need to add conditionals in the p3 and rrtmgp interfaces to either get those fields when active, or later on assign those fields (as they are used) to some zero array.

jgfouca:
[...]

PR #1715, opened 2022-06-03
Add `do_aerosol_rad` as an option for SCREAMv1

AaronDonahue:
This commit will add runtime parameters to P3 and RRTMGP interfaces to allow the user to turn on and off the prescribed aerosol.

For P3 the two new parameters are
`do_prescribed_ccn`
`do_predict_nc`
Which can be turned on and off.  In particular this is useful if SPA is active or not.

For RRTMGP the new parameter is
`do_aerosol_rad`
which should be turned off whenever SPA is not active to avoid passing incorrect non-zero aerosol forcing when it is not expected.

Additionally, this commit removes aerosol fields as "required" by P3 and RRTMGP if they won't be used.

Addresses #1707

### DIFF
--- a/components/scream/cime_config/namelist_defaults_scream.xml
+++ b/components/scream/cime_config/namelist_defaults_scream.xml
@@ -164,8 +164,8 @@ tweaking their $case/namelist_scream.xml file.
 
     <!-- P3 microphysics -->
     <p3 inherit="physics_proc_base">
-      <do__prescribed__ccn>true</do__prescribed__ccn>
-      <do__predict__nc>true</do__predict__nc>
+      <do_prescribed_ccn>true</do_prescribed__ccn>
+      <do_predic__nc>true</do_predict__nc>
     </p3>
 
     <!-- SHOC macrophysics -->
@@ -200,7 +200,7 @@ tweaking their $case/namelist_scream.xml file.
       <rad_frequency hgrid="ne256np4">3</rad_frequency>
       <rad_frequency hgrid="ne512np4">3</rad_frequency>
       <rad_frequency hgrid="ne1024np4">3</rad_frequency>
-      <do__aerosol__rad>true</do__aerosol__rad>
+      <do_aerosol_rad>true</do_aerosol_rad>
     </rrtmgp>
 
     <mac_aero_mic inherit="atm_proc_group">
--- a/components/scream/cime_config/namelist_defaults_scream.xml
+++ b/components/scream/cime_config/namelist_defaults_scream.xml
@@ -164,8 +164,8 @@ tweaking their $case/namelist_scream.xml file.
 
     <!-- P3 microphysics -->
     <p3 inherit="physics_proc_base">
-      <do_prescribed_ccn>true</do_prescribed__ccn>
-      <do_predic__nc>true</do_predict__nc>
+      <do_prescribed_ccn>true</do_prescribed_ccn>
+      <do_predict_nc>true</do_predict_nc>
     </p3>
 
     <!-- SHOC macrophysics -->
--- a/components/scream/cime_config/namelist_defaults_scream.xml
+++ b/components/scream/cime_config/namelist_defaults_scream.xml
@@ -200,6 +200,7 @@ tweaking their $case/namelist_scream.xml file.
       <rad_frequency hgrid="ne256np4">3</rad_frequency>
       <rad_frequency hgrid="ne512np4">3</rad_frequency>
       <rad_frequency hgrid="ne1024np4">3</rad_frequency>
+      <do__aerosol__rad>true</do__aerosol__rad>
     </rrtmgp>
 
     <mac_aero_mic inherit="atm_proc_group">
--- a/components/scream/cime_config/namelist_defaults_scream.xml
+++ b/components/scream/cime_config/namelist_defaults_scream.xml
@@ -163,7 +163,10 @@ tweaking their $case/namelist_scream.xml file.
     </homme>
 
     <!-- P3 microphysics -->
-    <p3 inherit="physics_proc_base"/>
+    <p3 inherit="physics_proc_base">
+      <do__prescribed__ccn>true</do__prescribed__ccn>
+      <do__predict__nc>true</do__predict__nc>
+    </p3>
 
     <!-- SHOC macrophysics -->
     <shoc inherit="physics_proc_base"/>
--- a/components/scream/data/scream_default_output.yaml
+++ b/components/scream/data/scream_default_output.yaml
@@ -6,45 +6,71 @@ Max Snapshots Per File: 744 # One output every 31 days
 Fields:
   Physics GLL:
     Field Names:
-      - T_mid
-      - PotentialTemperature
-      - qv
-      - qc
-      - qr
-      - qi
-      - qm
-      - nc
-      - nr
-      - ni
-      - bm
-      - precip_liq_surf
-      - precip_ice_surf
-      - eff_radius_qc
-      - eff_radius_qi
-      - nccn
-      - ni_activated
-      - nc_nuceat_tend
+      # HOMME
+      - horiz_winds
+      - ps
+      - pseudo_density
+      - omega
+      - p_int
+      - p_mid
+      # SHOC
       - cldfrac_liq
+      - eddy_diff_mom
+      - horiz_winds
+      - sgs_buoy_flux
+      - tke
+      - inv_qc_relvar
+      - pbl_height
+      # CLD
       - cldfrac_ice
       - cldfrac_tot
-      - horiz_winds
-      - tke
-      - eddy_diff_mom
-      - sgs_buoy_flux
-      - pbl_height
-      - LW_flux_up
-      - LW_flux_dn
-      - SW_flux_up
-      - SW_flux_dn
+      # SPA
+      - aero_g_sw
+      - aero_ssa_sw
+      - aero_tau_lw
+      - aero_tau_sw
+      - nccn
+      # P3
+      - bm
+      - nc
+      - ni
+      - nr
+      - qi
+      - qm
+      - qr
+      - T_prev_micro_step
+      - qv_prev_micro_step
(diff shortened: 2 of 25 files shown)


AaronDonahue:
WIP for now because it still needs polishing, but opening the PR so that we can discuss what (if any) polishing needs to be done.

AaronDonahue:
This PR will also ensure that if we turn SPA off that P3 and RRTMGP won't grab static values for prescribed aerosols from the initial condition file and run with them.  Turning off `do_prescribed_ccn` and `do_aerosol_rad` will remove the prescribed aerosol fields from the field manager entirely.

components/scream/cime_config/namelist_defaults_scream.xml
```diff
@@ -163,7 +163,10 @@ tweaking their $case/namelist_scream.xml file.
     </homme>
 
     <!-- P3 microphysics -->
-    <p3 inherit="physics_proc_base"/>
+    <p3 inherit="physics_proc_base">
+      <do__prescribed__ccn>true</do__prescribed__ccn>
```
jgfouca, line 167: @AaronDonahue , you don't need double underscores here. It looks like the real name of this parameter is `do_prescribed_ccn`. The double underscores are for represented spaces (which need to be represented as double underscores because XML does not allow spaces in element tag names), so you'd only need to do it the way you did it if the real name of the parameter was `do prescribed ccn`

components/scream/cime_config/namelist_defaults_scream.xml
```diff
@@ -197,6 +200,7 @@ tweaking their $case/namelist_scream.xml file.
       <rad_frequency hgrid="ne256np4">3</rad_frequency>
       <rad_frequency hgrid="ne512np4">3</rad_frequency>
       <rad_frequency hgrid="ne1024np4">3</rad_frequency>
+      <do__aerosol__rad>true</do__aerosol__rad>
```
jgfouca, line 203: I don't think you need the double underscores here either.

jgfouca (changes requested):

AaronDonahue:
@brhillman , I've added you as a reviewer mainly to check the RRTMGP interface changes.
@hassanbeydoun , Can you check my P3 changes?
@jgfouca and @bartgol , can you check my implementation?  Maybe there is a more elegant way?

AaronDonahue:
Somehow my fat fingers accidentally pressed the proper combination of hot keys on the keyboard to close this PR.

components/scream/cime_config/namelist_defaults_scream.xml
AaronDonahue: This should be resolved now, can you check again?

components/scream/cime_config/namelist_defaults_scream.xml
```diff
@@ -163,7 +163,10 @@ tweaking their $case/namelist_scream.xml file.
     </homme>
 
     <!-- P3 microphysics -->
-    <p3 inherit="physics_proc_base"/>
+    <p3 inherit="physics_proc_base">
+      <do_prescribed_ccn>true</do_prescribed__ccn>
```
jgfouca, line 167: I still see some double underscores here. There's one in the second separator in the closing tag of `</do_prescribed__ccn>` and both of the second separators for `do_predic__nc`

jgfouca (changes requested):

components/scream/cime_config/namelist_defaults_scream.xml
AaronDonahue: ha - I was fixing 2 fast 2 furious, but not quite as cool

AaronDonahue:
@jgfouca , there was one typo I missed so I had to push a new commit and re-request your review.

jgfouca (dismissed):

AaronDonahue:
I've tracked down the weaver fail to being an issue of non-BFB for any new run.  I ran the same exact case twice, with no change, and saved the NC output from each simulation and they are non-BFB.  So it is no wonder that restarts are also non-BFB.  Must be something that happens at initialization.  I'll investigate.  The diffs are on the O(1e-16) and are in the `sfc_flux_*` variables in rrtmgp.

AaronDonahue:
The `weaver` fails can be attributed to what is discussed in issue #1727 and are an issue in master.  This PR just happened to introduce the right variables to the output stream.  Presumably the fails on `mappy` for restart tests could also be attributed to the same issue.  Someone with access to the `mappy` logs could look up which variables are diff'ing.  @jgfouca , if you have a moment could you take a look?

jgfouca:
@AaronDonahue , for ERS_Ln9.ne4_ne4.F2000-SCREAMv1-AQP1, 3 variables were involved in the restart diff:

```
micro_liq_ice_exchange   (lev,ncol,time)  t_index =      1     1
         57    62352  (    38,   577,     1) (    58,   148,     1) (    51,   592,     1) (    46,     7,     1)
               62352   1.765754383302761E-07  -6.180488868091096E-07 5.2E-12  5.231048397393170E-12 9.1E-04  3.965303146512797E-17
               62352   1.765754383302761E-07  -6.180488868091096E-07          0.000000000000000E+00          0.000000000000000E+00
               62352  (    38,   577,     1) (    58,   148,     1)
          avg abs field values:    2.242275170379575E-09    rms diff: 2.1E-14   avg rel diff(npos):  9.1E-04
                                   2.242275086212496E-09                        avg decimal digits(ndif):  0.0 worst:  0.0
 RMS micro_liq_ice_exchange           2.0949E-14            NORMALIZED  9.3428E-06

 micro_vap_ice_exchange   (lev,ncol,time)  t_index =      1     1
        674    62352  (    54,   338,     1) (    62,   137,     1) (    38,    14,     1) (    46,     5,     1)
               62352   3.445601327367164E-08  -2.145529799697381E-07 1.2E-09 -1.176800569518160E-09 1.1E-02 -1.085005060089990E-10
               62352   3.445601327367164E-08  -2.145529799697381E-07          0.000000000000000E+00          0.000000000000000E+00
               62352  (    54,   338,     1) (    62,   137,     1)
          avg abs field values:    9.668803709890597E-10    rms diff: 1.6E-11   avg rel diff(npos):  1.1E-02
                                   9.661360135271322E-10                        avg decimal digits(ndif):  0.0 worst:  0.0
 RMS micro_vap_ice_exchange           1.6328E-11            NORMALIZED  1.6894E-02

 micro_vap_liq_exchange   (lev,ncol,time)  t_index =      1     1
         59    62352  (     1,     1,     1) (    63,   583,     1) (    48,   527,     1) (    46,     7,     1)
               62352   0.000000000000000E+00  -8.526749879044721E-08 9.8E-13 -9.842232964701706E-13 9.5E-04 -2.568533104942590E-16
               62352   0.000000000000000E+00  -8.526749879044721E-08          0.000000000000000E+00          0.000000000000000E+00
               62352  (     1,     1,     1) (    63,   583,     1)
          avg abs field values:    4.016252781545810E-10    rms diff: 3.9E-15   avg rel diff(npos):  9.5E-04
                                   4.016252610657429E-10                        avg decimal digits(ndif):  0.0 worst:  0.0
 RMS micro_vap_liq_exchange           3.9428E-15            NORMALIZED  9.8172E-06
```

AaronDonahue:
Okay, I was able to reproduce the issue but also I get the same error on `master`, so it may be unrelated to this PR.

PeterCaldwell:
@wlin7 reported months ago that micro_liq_ice_exchange, micro_vap_ice_exchange, and micro_vap_liq_exchange were non-BFB upon restarts. These variables are only used for output, so our strategy at the time was to just quit outputting them by default. Apparently you dug the issue back up (which illustrates that ignoring issues isn't a good long-term strategy). We should fix it, but not in this PR.

jgfouca:
@PeterCaldwell , @AaronDonahue , so should I use my admin privileges to merge this?

AaronDonahue:
If we want to make it so that the tests pass we can just merge a commit that removes those three variables from the default output. I'm out sick today, but @jgfouca you probably could easily grep for them and delete them from the output.yank files. Including the one in components/scream/data. Then the AT should be able to merge.

PeterCaldwell:
What Aaron said.

jgfouca (approved):

jgfouca:
I will try to get a successful AT run

components/scream/cime_config/namelist_defaults_scream.xml
```diff
@@ -163,7 +163,10 @@ tweaking their $case/namelist_scream.xml file.
     </homme>
 
     <!-- P3 microphysics -->
-    <p3 inherit="physics_proc_base"/>
+    <p3 inherit="physics_proc_base">
+      <do_prescribed_ccn>true</do_prescribed_ccn>
```
PeterCaldwell, line 167: @hassanbeydoun - this should be false for default runs to continue having the same behavior as current master, right? So using SPA requires do_prescribed_ccn=false, do predict_nc=true, and do_aerosol_rad=true? And no-SPA requires the opposite for all 3 of these variables?

jgfouca:
@AaronDonahue , @PeterCaldwell , the mappy run was fine other than the CIME V1 diffs which are expected. Peter, it looks like you have a pending question. I will wait to force merge until I hear from you.

PeterCaldwell:
Thanks Jim. Yeah, I didn't expect CIMEv1 diffs because my understanding is that this PR implements a "stealth" feature rather than changing default behavior. I think I understand the reason (which Aaron had also pointed out), but need confirmation from @hassanbeydoun .

components/scream/cime_config/namelist_defaults_scream.xml
PeterCaldwell: ah yes, that's correct. I just wasn't distinguishing between ccn and nc.

PeterCaldwell:
@jgfouca - can you tell me which variables in the mappy fails had diffs?

PeterCaldwell:
> No fields that both shared were different.

Ok, that makes more sense and is what I expect. Please force merge away.

Issue #1734, opened 2022-06-10
Pruned list of default outputs in CIME-EAMxx runs


AaronDonahue:
Currently the default set of output for an EAMxx run may lead to unnecessarily big files with output variables we don't actually need.  So this issue is to start a discussion about which variables we want to keep in the default.  The current list, organized by process that changes it, is:
```
      # HOMME
      - horiz_winds
      - ps
      - pseudo_density
      - omega
      - p_int
      - p_mid
      # SHOC
      - cldfrac_liq
      - eddy_diff_mom
      - horiz_winds
      - sgs_buoy_flux
      - tke
      - inv_qc_relvar
      - pbl_height
      # CLD
      - cldfrac_ice
      - cldfrac_tot
      # SPA
      - aero_g_sw
      - aero_ssa_sw
      - aero_tau_lw
      - aero_tau_sw
      - nccn
      # P3
      - bm
      - nc
      - ni
      - nr
      - qi
      - qm
      - qr
      - T_prev_micro_step
      - qv_prev_micro_step
      - eff_radius_qc
      - eff_radius_qi
      - precip_ice_surf
      - precip_liq_surf
      # SHOC + P3
      - qc
      - qv
      # SHOC + P3 + RRTMGP + HOMME
      - T_mid
      # RRTMGP
      - sfc_alb_dif_nir
      - sfc_alb_dif_vis
      - sfc_alb_dir_nir
      - sfc_alb_dir_vis
      - LW_flux_dn
      - LW_flux_up
      - SW_flux_dn
      - SW_flux_dn_dir
      - SW_flux_up
      - rad_heating_pdel
      - sfc_flux_lw_dn
      - sfc_flux_sw_net
      # Diagnostics
      - PotentialTemperature
```

AaronDonahue:
I think there is agreement we should remove the SPA section of variables since they are just regridded variables from the SPA data files anyway.

AaronDonahue:
We probably don't need the T_prev_micro_step and qv_prev_micro_step either, those are basically P3 internal variables.

PeterCaldwell:
I agree with both of Aaron's suggestions above. in addition, I'd remove
```
- sfc_alb_dif_nir
- sfc_alb_dif_vis
- sfc_alb_dir_nir
- SW_flux_dn_dir
```

AaronDonahue:
@PeterCaldwell , but keep
```
      - sfc_alb_dir_vis
```

[...]

PR #1735, opened 2022-06-10
Remove unneeded default output for our EAMxx CIME runs.

AaronDonahue:
The current set of default output for CIME runs includes everything
produced by all processes.  But many of these are not used for post
processing and analysis and thus don't need to be written by default.

A user can always include them before simulation if they are needed
for debugging.

### DIFF
--- a/components/scream/data/scream_default_output.yaml
+++ b/components/scream/data/scream_default_output.yaml
@@ -7,19 +7,18 @@ Fields:
   Physics GLL:
     Field Names:
       # HOMME
-      - horiz_winds
       - ps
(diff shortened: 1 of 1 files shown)


AaronDonahue:
Addresses the discussion in #1734

ndkeen (dismissed):

AaronDonahue:
I believe the DIFFs in the AT are expected since the new output files will be missing certain variables.

AaronDonahue:
The DIFFs on mappy are expected.  I'm going to manually merge this PR

PR #1738, opened 2022-06-13
Add PEM test to med res suite

jgfouca:
We want a test to catch the non-BFB-when-PES-changes problem
that we have been struggling with.

### DIFF
--- a/cime_config/tests.py
+++ b/cime_config/tests.py
@@ -457,7 +457,6 @@ _TESTS = {
         "tests" : (
             "ERP_D_Ln9.ne4_ne4.F2010-SCREAMv1",
             "ERS_Ln9.ne4_ne4.F2000-SCREAMv1-AQP1",
-            "PEM_Ln9.ne4_ne4.F2010-SCREAMv1"
             )
     },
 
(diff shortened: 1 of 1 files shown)


jgfouca:
I am a bit hesitant to merge this since it would be a failing test that would show up on the dashboard (not too bad) and block PRs due to AT fails (very bad).

PeterCaldwell:
Great point, Jim. Perhaps we should add it to a suite that is used for nightlies but not for AT?

jgfouca:
@PeterCaldwell , we do run the medres suite on perlmutter. We could put it there but it would not get much visibility.

PeterCaldwell:
Maybe this task is a bit contrived - I think if we distill down to the essence of what we're looking for, it's that we don't want to forget PEM exists. We definitely want to add PEM to our test suite, but only once we actually start passing tests. I think adding PEM to medres now is fine as a reminder that we want to put it somewhere more visible later but don't want to totally forget about it now...

jgfouca:
@PeterCaldwell , OK I moved it to medres and increased the grid to ne30.

PeterCaldwell (approved):

PR #1743, opened 2022-06-15
Fix scripts-tests of jenkins script

jgfouca:
Cannot run full CIME cases when SCREAM_FAKE_ONLY is on. This should fix the broken dashboard build for scripts-tests on mappy.


jgfouca:
It looks like this finally made it through AT, so just needs approval from either @bartgol or @AaronDonahue

AaronDonahue:
I was going to let @bartgol review since this is more his expertise and I think he may be back from vacation now.

bartgol (approved):

PR #1744, opened 2022-06-15
Add a default output filename that matches the E3SM casename

AaronDonahue:
This commit will provide for a default output filename if one is not provided in the output control YAML file.  For typical CIME based runs the E3SM case name, ${CASE}, will be gathered by the AD and used as the default output file name.  If no case name can be found the default is "EAMxx".  

In cases where there are more than one output stream that will use the default file name there is a running tally, `h0`, `h1`, `h2`, etc. to keep track of each unique file.

If an output control file provides a case name already, as with our default output and all unit tests, then that case name will be used for the filename instead of the E3SM case name.  In other words, this PR shouldn't change any of the current output file names.

### DIFF
--- a/components/scream/cime_config/namelist_defaults_scream.xml
+++ b/components/scream/cime_config/namelist_defaults_scream.xml
@@ -47,7 +47,7 @@ tweaking their $case/namelist_scream.xml file.
       <sections>ctl_nl</sections>
     </file>
     <file name="data/scream_input.yaml" format="yaml">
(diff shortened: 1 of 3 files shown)


jgfouca (approved):

AaronDonahue:
@ndkeen , I confirmed this on one of my machines.  If I do
```
./atmchange Output__YAML__Files=data/scream_default_output.yaml,$SCRATCH_PATH/scream_output_experiment.yaml
```

PR #1746, opened 2022-06-16
Reverting the revert which reverted SHOC dry<->wet conversion and ERP bug fix

singhbalwinder:
This PR reverts a commit to include SHOC wet<->dry conversion codes into master. It also fixes the ERP test failure.


components/eam/src/physics/cam/shoc_intr.F90
```diff
@@ -81,11 +82,18 @@ module shoc_intr
...
+
+  !lq_dry_wet_cnvr is true for all the water constituents used  by SHOC.
+  !These water constituents will participate in dry/wet mmr conversion
+  logical      :: lq_dry_wet_cnvr(pcnst)
  
   logical            :: history_budget
   integer            :: history_budget_histfile_num  
   logical            :: micro_do_icesupersat
+
+  !Store names of the water constituents used by SHOC
```
AaronDonahue, line 94: We probably need a logical here to switch between P3 and MG2.  I know that in SCREAM we only are committed to use P3, but in case in the future someone wants to experiment with MG2+SHOC this will throw an error or cause other issues since some of these constituents are not in MG2.

components/scream/src/physics/shoc/atmosphere_macrophysics.hpp
```diff
@@ -88,6 +89,33 @@ class SHOCMacrophysics : public scream::AtmosphereProcess
...
+       *Since tracers from the host model (or AD) are  wet mixing ratios and SHOC expects
+       *these tracers in dry mixing ratios, we convert the wet mixing ratios to dry mixing
+       *ratios for all the tracers *except* for qv [which is done after all the other tracers]
+       *and TKE [which is always defined in units of J/(kg of total air) and therefore
+       *should be considered to be "wet" when stored in the FM and also "wet" when used in
+       *SHOC].
+       *----------------------------------------------------------------------------------
+       */
+      Kokkos::parallel_for(Kokkos::TeamThreadRange(team, num_qtracers-2), [&] (const Int& idx) {
+          int iq   = idx / nlev_packs;
```
AaronDonahue, line 106: doesn't this loop only go for `idx` in `1...num_qtracers-2`?  If so, I don't see how we can grab `ilev` from `idx`.  Instead it would seem like `idx` is `iq`.  Am I missing something here?

components/scream/src/physics/shoc/atmosphere_macrophysics.hpp
```diff
@@ -214,12 +247,15 @@ class SHOCMacrophysics : public scream::AtmosphereProcess
     view_2d              cloud_frac;
 
     // Assigning local variables
-    void set_variables(const int ncol_, const int nlev_, const int num_qtracers_, const Real z_surf_,
+    void set_variables(const int ncol_, const int nlev_, const int num_qtracers_,
+                       const view_1d_int_const& convert_wet_dry_idx_d_,
+                        const Real z_surf_,
```
AaronDonahue, line 252: has an extra white-space.  Not a big deal though, just mentioning.

components/scream/src/physics/shoc/atmosphere_macrophysics.hpp
```diff
@@ -313,13 +355,44 @@ class SHOCMacrophysics : public scream::AtmosphereProcess
...
+       *DRY-TO-WET MMRs:
+       *-----------------
+       *Since the host model (or AD) expects wet mixing ratios, we need to convert dry
+       *mixing ratios from SHOC to wet mixing ratios except for qv[which will be converted
+       *in the following "parallel for" after all the other tracers] and TKE [which is
+       already in wet mmr].
+       *---------------------------------------------------------------------------------
+       */
+
+      Kokkos::parallel_for(Kokkos::TeamThreadRange(team, num_qtracers-2), [&] (const Int& idx) {
```
AaronDonahue, line 369: Same comment here as the loop in the preprocess.

AaronDonahue (changes requested):

AaronDonahue:
@singhbalwinder , I had a few comments in my review.

components/scream/src/physics/shoc/atmosphere_macrophysics.hpp
```diff
@@ -328,22 +401,26 @@ class SHOCMacrophysics : public scream::AtmosphereProcess
...
+      num_qtracers = num_qtracers_;
+      convert_wet_dry_idx_d = convert_wet_dry_idx_d_;
       rrho = rrho_;
       qv = qv_;
       qw = qw_;
       qc = qc_;
       qc_copy = qc_copy_;
       tke = tke_;
       tke_copy = tke_copy_;
+      qtracers = qtracers_;
```
bartgol, line 423: Not a perf/mem issue, but it is somewhat confusing to see qtracers as well as qv/qw/qc as inputs to the function. You can of course store a copy of qc/qv/qw, but you might just grab qv/qw/qc from qtracers. Of course, you would have to pass the indices of qv/qw/qc in order to properly subview qtracers.

Note: this is not a _big_ deal. It's just that when I see `f(x,y)` i always think that `x` and `y` are _independent_ variables. If `x` _has_ to be a subview of `y`, then it's probably more self-documenting to have `f(idx_x,y)`, and extract `x=y(:,idx_x,:)` inside the function.

PeterCaldwell:
@singhbalwinder - what's the status on this? In particular, I thought that some of the water leak problems we've been having were due to wet/dry conversions, but now I see that we're still not using the wet/dry code improvements yet? I guess the problem is that we are alreadyusing wet/dry fix in P3 and this PR is about SHOC?

components/eam/src/physics/cam/physics_utils.F90
```diff
@@ -1,88 +1,88 @@
 ! physics_utils: common elements for SCREAM physics models.
 module physics_utils
-
+
```
AaronDonahue, line 3: @singhbalwinder , did you add all of these extra spaces for a reason?  It seems to me this may put EAMxx in conflict with upstream E3SM for no real reason.

components/scream/src/physics/shoc/atmosphere_macrophysics.cpp
```diff
@@ -298,10 +300,36 @@ void SHOCMacrophysics::initialize_impl (const RunType run_type)
...
   }
+  // Find index of qv (water vapor, kg/kg(wet-air) and tke (J/kg(wet-air)) in the qtracer 3d view
+  // These indices are later used for converting tracers from wet mmr to dry mmr and vice-versa
+  auto qv_index  = tracer_info->m_subview_idx.at("qv");
+  auto tke_index = tracer_info->m_subview_idx.at("tke");
+
+  //view to store indices of tracers which will participate in wet<->dry conversion; we are excluding
+  //"tke" [as it is not "water based" tracer] and "qv"[as "qv" (before conversion) is needed for
+  //computing conversion for all other tracers] from qtracers view
+  using view_1d    = typename KT::template view_1d<Int>;
```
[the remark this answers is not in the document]
AaronDonahue: ~~We still need to change this to `view_1d_int`~~

I think we can just delete this line, you define `view_1d_int` in the hpp file.

components/scream/src/physics/shoc/atmosphere_macrophysics.cpp
```diff
@@ -298,10 +300,36 @@ void SHOCMacrophysics::initialize_impl (const RunType run_type)
...
+
+  //device view containing indices of tracers participating in wet<->dry conversion (2 tracers, qv and tke, are excluded)
+  view_1d convert_wet_dry_idx_d("convert_wet_dry_idx_d",m_num_tracers-2);
```
AaronDonahue, line 314: is this the variable that should be `view_1d_int`

components/scream/src/physics/shoc/atmosphere_macrophysics.hpp
```diff
@@ -32,6 +32,7 @@ class SHOCMacrophysics : public scream::AtmosphereProcess
   using Spack                = typename SHF::Spack;
   using IntSmallPack         = typename SHF::IntSmallPack;
   using Smask                = typename SHF::Smask;
+  using view_1d_int_const    = typename KT::template view_1d<const Int>;
```
AaronDonahue, line 35: Do we want a `const` view of int's to use or a non-const?  I feel like if I understand the usage in the cpp it should be non-const since you later set it to a specific value.  Maybe you need both a `view_1d_int` and a `view_1d_const`.

components/scream/src/physics/shoc/atmosphere_macrophysics.hpp
AaronDonahue: This comment still stands

components/scream/src/physics/shoc/atmosphere_macrophysics.hpp
```diff
@@ -176,6 +207,7 @@ class SHOCMacrophysics : public scream::AtmosphereProcess
     // Local variables
     int ncol, nlev, num_qtracers;
     Real z_surf;
+    view_1d_int_const convert_wet_dry_idx_d;
```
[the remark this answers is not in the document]
AaronDonahue: this comment also still stands.

components/scream/src/physics/shoc/atmosphere_macrophysics.hpp
AaronDonahue: same here

components/eam/src/physics/cam/shoc_intr.F90
```diff
@@ -81,11 +83,18 @@ module shoc_intr
...
+  character(len=6), parameter :: dry_wet_scalars(9) = ['Q     ','CLDLIQ', &
```
AaronDonahue, line 96: Can you confirm that SHOC uses all of these tracers?  Including RAINQM, CLDRIM, NUMRAI, BVRIM?  If so then we have a little more work cut out for us because SHOC should be agnostic to the microphysics scheme and CLDRIM and BVRIM are specific to P3 only.  MG2 has a SNOW constituent instead.  So if we actually need these here then we need a conditional to check the `microp_scheme` and act accordingly.  Otherwise the conditional loop down below will fail with MG2 to capture SNOW.

AaronDonahue (changes requested):
@singhbalwinder , I had some more comments and changes requested.

components/eam/src/physics/cam/physics_utils.F90
singhbalwinder: I was working on Cori with a different editor which added these white spaces. I will fix these in my next commit.

components/eam/src/physics/cam/shoc_intr.F90
singhbalwinder: My original idea was to have this array and add all the "possible" water constituents (which includes water constituents from all the microphysics schemes such as MG) in this array. Later in the code, we can check if a water constituent in the Q array exists in this list or not. If it exists, it can participate in the wet/dry conversion. 

From your comment above, I agree with you that it may be too restrictive. Another idea is to let all constituents in the Q array participate in the wet/dry conversion except for the TKE. Do you think that is a viable solution? That will let SHOC run with any microphysics scheme.

bartgol: Perhaps we can hard-code P3 stuff for scream. But for EAMxx, we might want to have a "water_constituent" group, so that SHOC can be used without assuming anything on which constituents are actually in the model.

singhbalwinder: Yeah, it will be better to have such a tag or group for the C++ codes. Currently C++ codes just exclude TKE from the dry/wet conversion.

We can use similar approach (of excluding TKE) in the F90 code.

AaronDonahue: Yeah, I think if we go for the exclusion approach it would be safer.  I think you can grab a string array of the constituent names and then search through them and exclude TKE.  That way if we have SNOW (MG2) or RIME (P3) the SHOC process won't care.  It will just grab all constituents except TKE.  Does that make sense?

components/scream/src/physics/shoc/atmosphere_macrophysics.hpp
singhbalwinder: @AaronDonahue : Thanks for catching this. I added an additional kokkos parallel-for to loop over levels.

components/scream/src/physics/shoc/atmosphere_macrophysics.hpp
singhbalwinder: Fixed this the same way (added an additional parallel-for)

components/scream/src/physics/shoc/atmosphere_macrophysics.hpp
AaronDonahue: @singhbalwinder , I think this question still applies.

components/scream/src/physics/shoc/atmosphere_macrophysics.hpp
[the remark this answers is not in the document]
AaronDonahue: @singhbalwinder , same with this comment.

components/scream/src/physics/shoc/atmosphere_macrophysics.hpp
AaronDonahue: @singhbalwinder , still not a big deal, but as long as you may be making other changes.

components/scream/src/physics/shoc/atmosphere_macrophysics.hpp
AaronDonahue: @bartgol , do you still have this comment?

AaronDonahue (approved):
Looks good Balwinder.  I had one other comment about a const vs. non-const view that would be good to confirm.  Otherwise this should be ready to go in.

AaronDonahue (approved):
thanks for all of your work on this @singhbalwinder

AaronDonahue:
Looks like this is ready, it just needs a manual merge. We did merge the non-BFB IO PR today. @bartgol do you have any opinion on also manually merging this PR?

bartgol:
> Looks like this is ready, it just needs a manual merge. We did merge the non-BFB IO PR today. @bartgol do you have any opinion on also manually merging this PR?

Given the CIME usage of cprnc, the PR was actually BFB! So we can merge a non BFB pr whenever.

singhbalwinder:
Maybe something has changed recently regarding testing but I think this PR should be NBFB for the CIME tests. This PR changes just the interface codes for SHOC. CIME tests invoke these interface codes, so they should fail at compare baseline step. Tests on the C++ side, do not test the interface codes for BFBness, so they should all pass.

AaronDonahue:
@singhbalwinder , I forgot to bring this up at the meeting.  Did you get any feedback on your question as to why we are not seeing a DIFF in the CIME tests?  @bartgol any thoughts?

singhbalwinder:
From the last testing results above, the CIME tests did fail with a baseline DIFF: `DIFF ERP.ne4pg2_ne4pg2.F2010-SCREAM-HR.mappy_gnu.eam-double_memleak_tol (phase BASELINE)`, which is what I expected. The rest of the tests are all PASS.

AaronDonahue:
But shouldn't this PR be non-BFB for all tests?  Like even SMS_D, and all SCREAMv1?  We are changing the input values for shoc.

singhbalwinder:
You are right. I was more focused on the fails...and didn't look at the CIME v1 tests. I will run the v1 test and try to see what is going on there.

singhbalwinder:
I have pushed changes that should affect the CIME v1 simulations as well. I expect it to fail now. Thanks, @AaronDonahue for catching this!

AaronDonahue (approved):
thank for the fix Balwinder

AaronDonahue:
This is now DIFF in the expected SCREAMv1 tests on mappy.

PR #1747, opened 2022-06-16
Add new SCREAM_TEST_LEVEL concept

jgfouca:
Current SCREAM_TEST_LEVELS:
```
0) "AT" : stands for autotester. This is the default. Test at this level
   are expected to always pass and not take a long time to run.
1) "NIGHTLY" : These are tests that run nightly. They can take longer and
   fails are possible but should be addressed in a reasonable time frame.
2) "EXPERIMENTAL" : These tests are not run by any automated test process
   and do not appear on our dashboard. This is for testing future features
   that are not fully developed yet.
```

```
Usage:
  Command line:
    cmake ... -DSCREAM_TEST_LEVEL=NIGHTLY ...

  Cmake:
    Making a level-restricted test:
      CreateUnitTest(${test_name} ${test_src} MINIMUM_TEST_LEVEL ${SCREAM_TEST_LEVEL_NIGHTLY})

    Making general level-restricted cmake logic:
      # Code to run only if test level is at least nightly
      if (SCREAM_TEST_LEVEL GREATER_EQUAL SCREAM_TEST_LEVEL_NIGHTLY)
        # do nightly/experimental stuff
      endif()

  test-all-scream:
    ./test-all-scream --test-level=nightly ...
```

Note that, in test-all-scream, test-level is not pinned to a specific build type (dbg, opt, etc)
but instead can be set for all test types.

Change list:
1) Add --test-level option to test-all-scream
2) Add support for MINIMUM_TEST_LEVEL option to Scream's CreateUnitTest/CreateUnitTestFromExec Cmake macro
  - This is the first Scream-only option (EKAT is unaware of test level) for our Cmake test API
3) Add tests for this new feature in scripts-tests and corresponding cmake tests in meta-tests/test_level_check
4) Remove SCREAM_ENABLE_EXPERIMENTAL cmake option since that functionaly is now available in this new concept


components/scream/tests/uncoupled/CMakeLists.txt
```diff
@@ -13,7 +13,7 @@ if (NOT SCREAM_BASELINES_ONLY)
   else()
     message(STATUS "RRTMGP only supported for double precision builds; skipping")
   endif()
-  if (SCREAM_ENABLE_EXPERIMENTAL)
+  if (SCREAM_TEST_LEVEL GREATER_EQUAL SCREAM_TEST_LEVEL_EXPERIMENTAL)
```
AaronDonahue, line 16: Have we checked if the code will build at the experimental level with this flag.  I think the ZM sub-directory is so old and unsupported that it may not even build now.  Maybe we can comment this out and add a comment that ZM is no longer supported even experimentally.

AaronDonahue (commented):
@jgfouca , thank you for getting on top of this so quickly.  This is great.  I had just one question about ZM and a separate question:
Can we add the PEM SCREAMv1 test to the nightly level in this PR?  Or maybe it already has been set as nightly?

jgfouca:
@AaronDonahue , this new capability is for SCREAM standalone tests only, IE tests defined via cmake. It won't have any impact on our CIME test suites. Fortunately, we already have all that we need to manage our CIME test suites; we just need to do it.

components/scream/tests/uncoupled/CMakeLists.txt
jgfouca: I did not check that and I'm not worried about it in this PR unless you think it is urgent to remove ZM. The new code is functionally identical to the old, so ZM problems will not be created by this PR.

components/scream/CMakeLists.txt
```diff
@@ -250,6 +247,28 @@ endif()
...
+#
+
+# Constants
+set(SCREAM_TEST_LEVEL_AT "0")
+set(SCREAM_TEST_LEVEL_NIGHTLY "1")
+set(SCREAM_TEST_LEVEL_EXPERIMENTAL "2")
+
+set(SCREAM_TEST_LEVEL "AT" CACHE STRING "The test level to run. Default is AT. NIGHTLY will run additional tests but is not guaranteed to PASS. EXPERIMENTAL will run even more tests with failures being more likely")
+
+if (SCREAM_TEST_LEVEL STREQUAL "AT")
```
bartgol, line 261: This might be personal taste, but I think that a "test level" suggests that the different values add an increasingly higher amount of testing. This is the case here, but it might be hard for users to know which level is higher. Also, "AT" might not be very self-explanatory for users.

What about
- "FAST" instead of "AT": immediately hints to the fact that tests are relatively quick.
- "FULL", or "STANDARD", or "NORMAL" instead "NIGHTLY"
- "EXTENSIVE" instead of  "EXPERIMENTAL": this should hint to the fact that it covers everything, and might be quite costly.

bartgol (commented):
There's already another test-level-related cmake variable, SCREAM_TEST_PROFILE (SHORT, MEDIUM, or LONG). I wonder if there could be confusion between the two.

Maybe, SCREAM_TEST_PROFILE (which is used to switch the space/time resolution) could be renamed to something hinting at resolution. Maybe SCREAM_TEST_RESOLUTION, with values "LOW", "MEDIUM", and "HIGH"? And each test level is free to set the default value for the test resolution. E.g., for "AT"/"FAST" test level, we would set a "LOW" resolution, while for 'NIGHTLY"/"STANDARD"/"NORMAL" level we would set "MEDIUM" resolution.

Note: the resolution may or may not impact the tests. In particular, tests need to use the `SetVarDependingOnTestProfile` macro in order to change the test "intensity".

components/scream/CMakeLists.txt
jgfouca: @bartgol , the intent of the SCREAM_TEST_LEVEL is more geared towards how crucial it is that the tests always passes rather than how fast it runs. While it is true that we generally want AT tests to be pretty fast, it would not be unreasonable to put some slower tests at the AT level if that's what we needed to ensure PRs did not break important functionality:

AT: this test must always pass
Nightly: this test should always pass. If it doesn't than we see it on our dashboard and we're working on it.
Experimental: no guarantees. Useful for future/experimental features

jgfouca:
I just pushed an important commit. My initial PR forgot to enable nightly mode in the jenkins script for nightly runs. I also removed SCREAM_AUTOTESTER from our CMake which was not used anywhere and is now redundant with SCREAM_TEST_LEVEL=AT.

jgfouca:
Summing up some info:

With this PR, we now have two Cmake vars controlling test "levels", SCREAM_TEST_PROFILE and SCREAM_TEST_LEVEL. 

SCREAM_TEST_PROFILE is used to control test parameters (NUM_STEPS) for a handful of our tests. The primary use case is for reducing steps when valgrind or cuda-memcheck is on. Potential values are SHORT, MEDIUM, and LONG.

SCREAM_TEST_LEVEL is the new variable and it is not used much at the moment. The only current use is to disable tests in `scream/tests/uncoupled/zm` unless the test level is set to EXPERIMENTAL. Conceptually, it is different from SCREAM_TEST_PROFILE because it expresses how critical it is that a test is passing. I think it is valid to say that, for example, a certain MEDIUM profile test could be at test level "AT" (IE critical, must pass).

@AaronDonahue , @PeterCaldwell , I'd be interested to hear your thoughts on this. Is this situation confusing? Are the names of these two variables clear? Are these two concepts different-enough to remain separate or should we merge them into a single variable?

bartgol:
I think the two names "profile" and "level" are somewhat confusing. In particular, I think "profile" is somewhat ambiguous now that there are two knobs. I can see someone thinking that "AT", "Nightly", and "Experimental" should be the values of the test "profile", while "short", "medium", and "long" should be the values for "level".

I think we should take this chance to rename vars, if needed. I think "profile" might be renamed to something hinting at resolution, or size, like `SCREAM_TESTS_GRID_RES` or `SCREAM_TESTS_SIZE`. For the new var, `SCREAM_TEST_LEVEL` is fine (provided we change 'profile' into something else), or maybe something like `SCREAM_TESTING_TYPE` (or maybe use `SCREAM_TESTING_PROFILE` for this knob).

PeterCaldwell:
Sorry for not weighing in on this sooner, Jim. I kept struggling to get my head around the need, then wandering off to other tasks. 

First, I see computational efficiency and expected robustness as orthogonal dimensions of a test. I think both of these dimensions are useful to have control over. 

Second, I agree with Luca that our current names for these dimensions are problematic. I'm not sure I have good suggestions though. It would be nice if SCREAM_TEST_PROFILE could be SCREAM_TEST_CHEAPNESS or SCREAM_TEST_EFFICIENCY or something, but I think the idea is that we could potentially stick a lot of parameters into this variable rather than just efficiency ones. If so, maybe SCREAM_TEST_PARAM_SETTINGS or something? SCREAM_TEST_LEVEL is supposed to reflect the expectation or robustness of the test. So maybe SCREAM_TEST_EXPECTATION or SCREAM_TEST_ROBUSTNESS? Or go really long with SCREAM_TEST_PASSEXPECTATION?

jgfouca:
No worries, @PeterCaldwell , this is not urgent.

> First, I see computational efficiency and expected robustness as orthogonal dimensions of a test. I think both of these dimensions are useful to have control over.

Yes, I strongly agree here. 

I think SCREAM_FLEX_TEST_SIZE (replaces SCREAM_TEST_PROFILE) and SCREAM_TEST_LEVEL would be concise and clear enough. @bartgol ?

The name SCREAM_FLEX_TEST_SIZE is intended to portray the concept that a few of our tests have a flexible test size (or test steps) which can be tuned by setting this parameter. I initially was thinking SCREAM_TEST_SIZE, but that implies that it applies to a wide set of tests when the reality is that only a small number of tests have tunable size.

bartgol:
Sounds good.

Side comment, completely orthogonal to this PR: I've been thinking for a while to put all of SCREAM config options in a separate cmake file, like `cmake/ScreamConfigOptions.cmake`, so that the user can find all of them in one place, hopefully with a somewhat decent description. Right now, we have all of them scattered throughout our main CMakeList.txt, which is 460 lines long. Also, some config options could be relegated to subfolders (e.g. `SCREAM_RRTMGP_DEBUG`), since they only make sense in the context of those sub-components.

PeterCaldwell:
@jgfouca - SCREAM_TEST_LEVEL seems a bit ambiguous to me still - level can equally well refer to level of computational efficiency as level of robustness. And SCREAM_FLEX_TEST_SIZE also doesn't make sense to folks that don't already know what it is - I was like "is this where the weight lifters go to compare biceps?". In that context I might prefer SCREAM_TEST_SIZE with a note in Luca's new ScreamConfigOptions.cmake saying that not all tests are affected by this setting.

jgfouca:
@PeterCaldwell 

> is this where the weight lifters go to compare biceps 

:)  ok we can stick with SCREAM_TEST_SIZE then

>  SCREAM_TEST_LEVEL seems a bit ambiguous to me still

I agree. I'm wondering what the standard industry jargon is for referring to "test levels"? "test types"? "test tranche"?

PeterCaldwell:
> I'm wondering what the standard industry jargon is for referring to "test levels"? "test types"? "test tranche"?

This got me wondering what would happen if I googled it. But then what to google? I ended up querying "test dependability" which didn't return anything useful, but might itself be a good name?

jgfouca:
@PeterCaldwell , this concept is too nuanced for google I think. I've asked some fellow SE/CS people in my org and the best they've come up with is "test category".

jgfouca:
Also,

```
there is a software engineering precedent for "The Four Levels of Testing" which refer to Unit, Integration, System, and Acceptance
```

So I think it's either SCREAM_TEST_LEVEL or SCREAM_TEST_CATEGORY

bartgol:
I think we could rebrand `SCREAM_TEST_PROFILE` for what Jim called `SCREAM_TEST_LEVEL`. I think a "profile" is probably an ok word to distinguish different kind of testing.

As for the current meaning of `SCREAM_TEST_PROFILE` (what it actually does right now in master), I think "flex" is good, but may be cleared if fully expanded, like`SCREAM_FLEXIBLE_TESTS_RESOLUTION`. Or maybe `SCREAM_DRIVER_TESTS_RESOLUTION`, since we basically only use that var for tests running through the AD...

jgfouca:
SCREAM_TEST_TIER is another suggestion.

jgfouca:
@PeterCaldwell , I'll leave the final call to you. There's a lot of suggestions above. I still think SCREAM_TEST_LEVEL and SCREAM_TEST_SIZE are the best pair. SCREAM_TEST_TIER and SCREAM_TEST_CATEGORY would be my second choices for SCREAM_TEST_LEVEL.

PeterCaldwell:
I like SCREAM_TEST_SIZE and SCREAM_TEST_CATEGORY best, but if SCREAM_TEST_LEVEL sounds much better you you CS types I'm fine with that as well.

bartgol:
Ok, the let's converge to Jim's idea of replacing `SCREAM_TEST_PROFILE` with `SCREAM_FLEXIBLE_TESTS_SIZE` (more spelled out might help) and keep `SCREAM_TEST_LEVEL` for the new feature. @jgfouca  thoughts?

PeterCaldwell:
Ok. I leave the final decision to Jim.

jgfouca:
OK, ready for final review.

bartgol (approved):

jgfouca:
Weaver and blake both passed. Mappy lost its connection midway through the run, so retesting with blake/weaver skipped.

bartgol (approved):

PR #1754, opened 2022-06-23
Use Homme get_default_policy instead of raw kokkos policy constructor in RemapFunctor

jgfouca:
This fixes the "Team size too large" error we have been seeing in our GPU debug runs.

### DIFF
--- a/components/homme/src/share/cxx/RemapFunctor.hpp
+++ b/components/homme/src/share/cxx/RemapFunctor.hpp
@@ -544,10 +544,10 @@ private:
   typename std::enable_if<OnGpu<ExecSpace>::value == true,
                           Kokkos::TeamPolicy<ExecSpace, FunctorTag> >::type
   remap_team_policy(int num_exec) {
-    constexpr int num_threads = 16;
-    constexpr int num_vectors = 32;
-    return Kokkos::TeamPolicy<ExecSpace, FunctorTag>(num_exec, num_threads,
-                                                     num_vectors);
+    ThreadPreferences tp;
+    tp.max_threads_usable = 16;
+    tp.max_vectors_usable = 32;
+    return Homme::get_default_team_policy<ExecSpace, FunctorTag>(num_exec, tp);
   }
 
   template <typename FunctorTag>


oksanaguba:
Should not this go into e3sm instead?

jgfouca:
@ambrad , this fix allowed me to get past the "team size" error and I then saw the same errors as I was getting for the optimized runs. Not really a great situation, but this did seem to help.

@oksanaguba I also did a grep and only found a few calls to Kokkos::TeamPolicy in `homme/src/share/cxx/mpi/BoundaryExchange.cpp`, none of which had the hardcoded threads/vector that RemapFunctor had.

ambrad:
What errors did you see on Summit? ne30 should be running cleanly on Summit.

jgfouca:
@ambrad , I was working with debug/opt versions of this case `SMS_Ln9.ne30_ne30.F2010-SCREAMv1.summit-gpu_gnugpu`. It looks like the default layout is to try to run it on one node, which is too few:

```
#BSUB  -nnodes 1 
...
total tasks: 42
...
jsrun -X 1 --nrs 6 --rs_per_host 6 --tasks_per_rs 7 -d plane:7 --cpu_per_rs 7 --gpu_per_rs 1
```

ambrad:
I agree. I don't think anyone has set up good defaults for Summit.

jgfouca:
Good news:
```
./create_test SMS_D_Ln9_P256.ne30_ne30.F2010-SCREAMv1 --compiler=gnugpu --wait
PASS SMS_D_Ln9_P256.ne30_ne30.F2010-SCREAMv1.summit-gpu_gnugpu RUN
    Case dir: /gpfs/alpine/cli115/proj-shared/acmetest/e3sm_scratch/SMS_D_Ln9_P256.ne30_ne30.F2010-SCREAMv1.summit-gpu_gnugpu.20220623_125512_c996da
```

IE this PR appears to be all that is needed to get debug runs working again.

ambrad:
Ok, cool. Let's close this PR and I'll make one to the upstream repo. I actually want to try to do it it a litte differently, as get_default_team_policy has an optional arg that should make it so inline KOKKOS_ENABLE_DEBUG logic isn't needed while still getting the same opt-build team config.

PR #1759, opened 2022-06-28
Restore surface fluxes to default output for CIME runs.

AaronDonahue:


PeterCaldwell (approved):

AaronDonahue:
Given that this PR adds fields to the default output, I think we can say that the DIFFs are due to the new fields and this is ready for manual merge.

PR #1762, opened 2022-06-29
Set up infrastructure to write output with a frequency other than steps.

AaronDonahue:
This commit adds a timestamp to the IOControl structure which will
keep a record of the timestamp of the last time output was written.

This can then be used with the current timestamp to determine the number
of secs, minutes, hours, days, months and years that have passed since the
last output was written.


bartgol:
I was working on the same issue in another branch, but never got to finish it.

One thing that I was considering was to separate the "write frequency" from the "sampling frequency". E.g., we might want to save a daily average of F, but averaging F every hour (without averaging every timestep). So write frequency would be 1day, but sampling frequency would be 1h (with a 5min timestep, this would be 12x reduction in I/O averaging operations). I think this needs to be encoded in the IOControl structure.

AaronDonahue:
> 

I like this suggestion.  It would certainly cut down on the number of IO operations we need to do per simulation and I imagine that for things like monthly output hourly sampling would be good enough since we are averaging out the noise anyway.  @PeterCaldwell may have some thoughts as well.

AaronDonahue:
@bartgol , what do you think of doing this change in two parts?  The first part would be to review and integrate what I have here so that other users can start requesting daily or monthly output.  The second step would be to update this interface to use an enum like you have in your branch and allow for sampling frequency and output frequency to be designated.

components/scream/src/share/io/scream_io_utils.hpp
```diff
@@ -44,11 +45,58 @@ inline OutputAvgType str2avg (const std::string& s) {
 struct IOControl {
   // A non-positive frequency can be used to signal IO disabled
   int frequency = -1;
-  int nsteps_since_last_write;
+  int nsteps_since_last_write;  // Needed when updating output data, such as with the OAT::Average flag
+  util::TimeStamp& timestamp_since_last_write;
```
bartgol, line 49: I wouldn't call it "since" last write. This is the timestamp _of_ the last write. If I read "since", I think that this is a timestamp difference (which is dangerous to use, since depending on when t0 is, t1-t0 may consist of a different number of months/years).

bartgol (changes requested):

AaronDonahue:
@bartgol , The unit tests still need some adjusting.  For now we only test up to a frequency output of DAYS.  Months and Years will take some more infrastructure.  In the meantime, I think the heart of the code, i.e. the io_control stuff, is ready for review.

components/scream/src/share/io/tests/io.cpp
```diff
@@ -251,116 +264,265 @@ TEST_CASE("input_output_basic","io")
...
-  for (int jj=0;jj<num_levs;++jj) {
-    avg_val = (max_steps+1)/2.0*dt + (jj+1)/10.;  //note x0=(jj+1)/10 in this case.
-    REQUIRE(std::abs(f2_host(jj)-avg_val)<tol);
-  }
-  avg_input.finalize();
+
+  // Get a fresh new field manager
+  auto field_manager = get_test_fm(grid);
+  const auto& out_fields = field_manager->get_groups_info().at("output");
+  auto reset_fields = [&] () {
```
bartgol, line 396: This utility is no longer needed, since you run _one_ avg type at a time in the run function.

components/scream/src/share/io/tests/io.cpp
```diff
@@ -251,116 +264,265 @@ TEST_CASE("input_output_basic","io")
...
+  auto f4 = field_manager->get_field("field_packed");
+  auto f1_host = f1.get_view<Real*,Host>();
+  auto f2_host = f2.get_view<Real*,Host>();
+  auto f3_host = f3.get_view<Real**,Host>();
+  auto f4_host = f4.get_view<Real**,Host>();
+
+
+  if (output_type == "instant") {
+    // Check instant output
+    AtmosphereInput ins_input(input_params,field_manager);
```
bartgol, line 437: The creation of the reader, the call to `read_variables`, and the sync's to host can all be done once for all output types, before the if statement.

components/scream/src/share/io/tests/io.cpp
```diff
@@ -251,116 +264,265 @@ TEST_CASE("input_output_basic","io")
...
+    }
+    ins_input.finalize();
+    reset_fields();
+  } else if (output_type == "average") {
+    // Check average output
+    AtmosphereInput avg_input(input_params,field_manager);
+    avg_input.read_variables();
+    f1.sync_to_host();
+    f2.sync_to_host();
+    f3.sync_to_host();
```
bartgol, line 467: Any reason why f4 and the diag output are not checked for output_type other than "instant"?

components/scream/src/share/io/tests/io.cpp
```diff
@@ -501,5 +661,41 @@ get_diagnostic_input(const ekat::Comm& comm, const std::shared_ptr<GridsManager>
...
+}
+/*========================================================================================================*/
+TEST_CASE("input_output_basic","io")
+{
+  std::vector<std::string> output_types =
+    { "instant","average", "max", "min"};
+  std::vector<std::string> output_freq =
+    { "nsteps", "nsecs", "nmins", "nhours", "ndays"};
+  for (const auto& of : output_freq) {
+    printf("Testing output freq %s\n",of.c_str());
```
bartgol, line 689: Would you mind wrapping these in some if statement, so that only one rank prints them?

components/scream/src/share/io/tests/io.cpp
```diff
@@ -501,5 +661,41 @@ get_diagnostic_input(const ekat::Comm& comm, const std::shared_ptr<GridsManager>
...
+    for (const auto& ot : output_types) {
+      printf("  Testing output type %s...",ot.c_str());
+      run(ot,of);
+      printf("Done!\n");
+    } 
+    printf("  Testing output type multisnap...");
+    run_multisnap(of);
+    printf("Done!\n");
+  }
+  printf("DONE\n");
```
bartgol, line 699: This last print seems pointless.

bartgol (commented):
When you output, say, every 10min, and avg_type=AVERAGE, does it mean that you average every single timestep? That is, for non-instant averaging, is the sampling frequency 1 timestep?

components/scream/src/share/io/tests/io.cpp
AaronDonahue: ahhh good catch.  I think that f4 was just to add a test for a packed variable so I only originally had it testing in one case.  But it was trivial to have it tested in all cases.

bartgol (approved):

AaronDonahue:
epic fails for this PR.  I'll investigate

cime_config/tests.py
```diff
@@ -459,6 +459,7 @@
         "time"  : "01:00:00",
         "tests" : (
             "ERP_D_Ln9.ne4_ne4.F2010-SCREAMv1",
+            "ERS_Ld5.ne4_ne4.F2010-SCREAMv1",
```
bartgol, line 462: Why adding a new ERS test?

AaronDonahue: I instead replaced the ERP_D test with an Lh4 instead of Ln9.  The purpose is to confirm that we have BFB restarts when the frequency is not number of steps.

AaronDonahue:
Alright, this is now passing everywhere except one DIFF on mappy.  Looking at which case failed it is the ERP_D_Ln4 case, note, this is the case that is added by this PR so that is not a surprise.  @bartgol  when you get a chance can you re-review?  If all is well I can manually merge.

components/scream/cime_config/namelist_defaults_scream.xml
```diff
@@ -280,9 +280,9 @@ tweaking their $case/namelist_scream.xml file.
       <Casename>./${CASE}.scream</Casename>
       <Output__Control>
         <Frequency>${REST_N}</Frequency>
+        <Frequency__Units>${REST_OPTION}</Frequency__Units>
         <Frequency REST_OPTION="none">0</Frequency>
-        <Frequency__Units>INVALID</Frequency__Units>
-        <Frequency__Units REST_OPTION="nstep*">Steps</Frequency__Units>
+        <Frequency__Units REST_OPTION="none">INVALID</Frequency__Units>
```
bartgol, line 285: Perhaps we should remove this line. If `REST_OPTION=none`, wouldn't the output manager figure out that no output is required? I assume it would. And if not, we could easily add a few lines to the class.

Seems silly to add this extra line just to convert an arbitrary string "none" to another arbitrary string "INVALID" (and the latter is, I think, never used anyways).

components/scream/data/scream_default_output.yaml
```diff
@@ -60,4 +60,7 @@ Output Control:
   Timestamp in Filename: true
   AVG Type in Filename: false
   Frequency in Filename: false
+Checkpoint Control:
+  Frequency: ${REST_N}
```
bartgol, line 64: Aren't `REST_XYZ` options for the model restart? Shouldn't we instead use `HIST_OPTION` and `HIST_N` from cime?

components/scream/src/share/io/scream_io_utils.hpp
```diff
@@ -44,12 +45,54 @@ inline OutputAvgType str2avg (const std::string& s) {
...
+      } else if (frequency_units == "nmins") {
+        ret = (ts_diff >= 60) && (ts_diff % (frequency*60) == 0);
+      } else if (frequency_units == "nhours") {
+        ret = (ts_diff >= 3600) && (ts_diff % (frequency*3600) == 0);
+      } else if (frequency_units == "ndays") {
+        ret = (ts_diff >= 86400) && (ts_diff % (frequency*86400) == 0);
+      } else if (frequency_units == "nmonths" || frequency_units == "nyears") {
+        // For months and years we need to be careful, can't just divide ts_diff by a set value.
+        // First we make sure that if we are the same day of the month and at the same time of day.
+        if (ts.get_day() == timestamp_of_last_write.get_day() &&
```
bartgol, line 74: This line here is _assuming_ that timestepping will end up on the same time of day every day. Is it safe to assume that atm timesteps will _always_ be such that there's an integer number of steps in a day? I think the answer is yes, due to ATM_NCPL cime entry, but wanted to double check.

components/scream/src/share/io/scream_output_manager.cpp
```diff
@@ -93,22 +94,30 @@ setup (const ekat::Comm& io_comm, const ekat::ParameterList& params,
...
+    m_checkpoint_file_specs.num_snapshots_in_file = 0;
+    m_checkpoint_file_specs.filename_with_time_string = pl.get("Timestamp in Filename",true);
+    m_checkpoint_file_specs.filename_with_mpiranks    = pl.get("MPI Ranks in Filename",false);
+    m_checkpoint_file_specs.filename_with_avg_type    = pl.get("AVG Type in Filename",true);
+    m_checkpoint_file_specs.filename_with_frequency   = pl.get("Frequency in Filename",true);
+  } else {
+    // If there is no restart data or there is but no checkpoint control sublist then we initialize
+    // the checkpoint control so that it never writes checkpoints.
+    m_checkpoint_control.frequency  = 0;
+    m_checkpoint_control.frequency_units = "never";
```
bartgol, line 118: We could use "none", to allow using CIME's `HIST_OPTION/REST_OPTION` entries directly.

bartgol: Also, if you change IOControl struct declaration in `scream_io_utils.hpp` to have
```
struct IOControl {
...
   std::string frequency_units = "none";
};
```
so that all IOControl objects are created with units="none" (hence, disabled) by default. Also, frequency is already inited to -1 in the struct decl by default, so I'm guessing setting it to 0 here is pointless.

components/scream/data/scream_default_output.yaml
AaronDonahue: No I think we still want `REST_N`.  If we use `HIST_N` then we will never need to checkpoint because that would correspond to when we write output.  But if HIST_N != REST_N and we write a restart file we will want a corresponding checkpoint to accompany it, in case we restart the run from that file.

components/scream/src/share/io/scream_io_utils.hpp
bartgol: Uhm, you also check secofday, so it should pick the right time of day. I'm ok without adding a check. You don't have dt in this routine after all, so you can't do it here.

components/scream/src/share/io/scream_output_manager.cpp
AaronDonahue: Good call, "never" is actually an acceptable entry for HIST/REST_OPTION, but it looks like "none" is the default from CIME so best to go with that.

AaronDonahue:
Failed for the new surface coupling test - I fixed that.

@tcclevenger and @bartgol , I noticed that the new surface coupling standalone test was put into the `/coupled/` sub-directory rather than `uncoupled`.  I moved it to `uncoupled` in my latest commit, since it is just a test of a single process.  Do you agree with this move?  If not, I can move it back.

bartgol:
I'll let Conrad answer that, but I suspect he put under coupled b/c there are _two_ atm processes in the test: SCImport and SCExport.

tcclevenger:
Yeah, @bartgol is correct, that is why it is there. But I understand that is confusing.. so maybe it is better that it is uncoupled? I don't have much of a preference..

bartgol (approved):
I'll let @tcclevenger  decide about whether the SC test should be a coupled or uncoupled test. Other than that, it looks good.

tcclevenger:
I think it's fine as an "uncoupled" test, just renamed to "surface_coupling" (since all the other uncoupled tests are, e.g., "shoc_only").

AaronDonahue:
> I think it's fine as an "uncoupled" test, just renamed to "surface_coupling" (since all the other uncoupled tests are, e.g., "shoc_only").

@tcclevenger - done.  I'll let one more round of AT go through and if all is well I'll go ahead and manually merge this.

tcclevenger (approved):

AaronDonahue:
had to fix a conflict, letting the AT have one more shot at it, then I will manually merge if all goes well.

tcclevenger (approved):

AaronDonahue:
All recent commits were just to handle conflicts in the branch, now the AT is passing everywhere except the one DIFF which is expected on mappy for a new CIME test this PR introduces.  I'm planning on merging unless there is some push back - @tcclevenger or @bartgol ?

tcclevenger:
Sounds good to me.

PR #1764, opened 2022-07-01
Make initialization easier for EAMxx

brhillman:
Make initialization and generating initial condition files for EAMxx. This adds initialization for a number of fields we do not need to read from file for initial runs, either to constants (most physics fields) or idealized profiles (for pseudo-density). This is more similar to how EAM initializes, and allows us to derive EAMxx initial conditions almost entirely from EAM/SCREAMv0 initial conditions *without* running SCREAMv0 with special outputs. Stripped down initial condition files generated from EAM “cami” style ICs will be added and used as defaults in a subsequent PR. This will be non-BFB for any test that uses HOMME, since initialization of pseudo-density is changed to be computed assuming hydrostatic initial state from hybrid coefficients instead of being read from file, and CIME tests will be non-BFB also because initialization of physics fields have changed to be set to constants initially.

### DIFF
--- a/components/scream/tests/coupled/dynamics_physics/model_restart/input_baseline.yaml
+++ b/components/scream/tests/coupled/dynamics_physics/model_restart/input_baseline.yaml
@@ -14,11 +14,14 @@ Time Stepping:
 Initial Conditions:
   Filename: ${SCREAM_DATA_DIR}/init/screami_ne2np4L72_20220524.nc
   Restart Run: false
-  Physics GLL:
-    surf_evap: 0.0
-    surf_sens_flux: 0.0
-    precip_liq_surf_mass: 0.0
-    precip_ice_surf_mass: 0.0
+  surf_evap: 0.0
+  surf_sens_flux: 0.0
+  precip_liq_surf_mass: 0.0
+  precip_ice_surf_mass: 0.0
+  aero_g_sw: 0.0
+  aero_ssa_sw: 0.0
+  aero_tau_sw: 0.0
+  aero_tau_lw: 0.0
 
 atmosphere_processes:
   atm_procs_list: (homme,physics)
--- a/components/scream/cime_config/namelist_defaults_scream.xml
+++ b/components/scream/cime_config/namelist_defaults_scream.xml
@@ -232,7 +232,7 @@ tweaking their $case/namelist_scream.xml file.
   <!-- List of nc files for loading inputs on specified grids -->
   <Initial__Conditions>
     <Filename>UNSET</Filename>
-    <Filename hgrid="ne4np4">/sems-data-store/ACME/bhillma/scream/data/init/screami_ne4np4L72_20220701.nc</Filename>
+    <Filename hgrid="ne4np4">${DIN_LOC_ROOT}/atm/scream/init/screami_ne4np4L72_20220701.nc</Filename>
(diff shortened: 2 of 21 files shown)


brhillman:
Note: we are setting surface fluxes to zero here, because we want to prevent the AD from trying to read them from file, but we know that surface coupling will take care of them by importing from the surface models. This is kind of a hack; ultimately, we probably want Surface Coupling to be a separate “process” in AD-speak that can be tagged as a “provider” for these fields so that we do not have to pretend to initialize them to constants.

components/scream/tests/coupled/dynamics_physics/model_restart/input_baseline.yaml
```diff
@@ -14,9 +14,12 @@ Time Stepping:
 Initial Conditions:
   Filename: ${SCREAM_DATA_DIR}/init/screami_ne2np4L72_20220524.nc
   Restart Run: false
-  Physics GLL:
-    surf_evap: 0.0
-    surf_sens_flux: 0.0
+  surf_evap: 0.0
```
bartgol, line 17: I'm puzzled why we need quite a few "init to 0" lines in the XML defaults for v1, while here we only need a handful.

AaronDonahue (dismissed):
It looks good to me.  Just a question though, this PR will not solve our issue of designating some fields as being initialized elsewhere, instead we are taking advantage of the ability to initialize a variable to a constant using the input.yaml file, right?

bartgol:
> It looks good to me. Just a question though, this PR will not solve our issue of designating some fields as being initialized elsewhere, instead we are taking advantage of the ability to initialize a variable to a constant using the input.yaml file, right?

Assuming I understand your question correctly, then yes. The AD has no knowledge of the fact that some fields will be inited inside the `initialize_impl` method of some atm process. We are in fact hijacking the "init to 0" feature to fool the AD to think those fields are inited, while they may in fact be inited later.

In order to do things cleanly we need to first allow to try reading N fields from nc file, without erroring out if one of them is not in the file (I am currently working on that feature). Once that's possible, the AD will use the feature to init "as many fields as possible" from nc file, call initialize impl for all processes, and then, as one of the last steps of the init sequence, verify that all the inputs in the atm DAG have been inited. If there's still some uninited fields at that point, then it's a legit error.

components/scream/tests/coupled/dynamics_physics/model_restart/input_baseline.yaml
brhillman: I think because these aren't using the stripped down input files.

PeterCaldwell:
This PR just passed for everything except baseline DIFFS for mappy SMS, ERP, and ERS tests... which I think is expected, right @brhillman ? 

I also see that we've got some merge conflicts to resolve. Can you do that ASAP, Ben?

brhillman:
@bartgol I just pushed a rebase that should address all the conflicts, but feel free to merge #1786 first if you need.

bartgol:
The standalone tests fail b/c of the FPE build. Interestingly, only the `shoc_cld_spa_p3_rrtmgp` test fails, everything else passes. I'll double check the inputs, maybe we forgot to init some inputs.

bartgol:
The PR is non-bfb, and is failing baseline cmp, as expected. Merging.

PR #1771, opened 2022-07-07
Add clear-sky fluxes

brhillman:
Add calculation for clear-sky radiative fluxes, and add corresponding fields to the field manager so that they can be output. Note that this will increase the cost of radiation somewhat, since the flux solvers need to be called an additional time (before cloud optical properties are merged in, and after). New fields added to FM for purposes of output are:
```
SW_clearsky_flux_up
SW_clearsky_flux_dn
SW_clearsky_flux_dn_dir
LW_clearsky_flux_up
LW_clearsky_flux_dn
```


brhillman:
Clearsky fluxes are consistent with implementation in EAM. SW and LW cloud radiative effect (difference in net TOA fluxes between allsky and clearsky fluxes) are comparable between the implementation here in SCREAMv1 and EAM. From a short 1-day simulation using both the SCREAMv1 code on this branch and EAM:
image

crterai (approved):
From what I can tell, it looks good to me. Especially, if tests show all-sky fluxes are B4B, then I see no issues.

Issue #1767, opened 2022-07-07
Reset accumulated variables in a standardized way


PeterCaldwell:
In SCREAMv1, we apply hacks in surface coupling to zero out variables which are accumulated over substeps in time. This can lead to confusion for someone who is tinkering with surface coupling without understanding other parts of the code (e.g. "why is precip zeroed here?") or for folks adding other accumulated variables to the model (e.g. "where is this variable reset between steps?"). See https://github.com/E3SM-Project/scream/issues/1755#issuecomment-1176886931 for discussion of this approach. 

We should in the future develop a tool in the AD for zeroing out accumulated variables between steps. Luca suggests
> we could think of marking some fields as "single-step" quantities, which, at the beginning of an atm step, must be reset to a value that is an "identity" for the operation that they undergo. E.g., for accumulated fields, reset to 0, while for fields that are "max" operation (like f = max(f, some_new_value)), reset to -infinity. The AD could easily take care of "resetting" these fields at the beginning of the time step.

Issue #1755, opened 2022-06-23
precipitation variables in P3 are incorrect when p3 is subcycled


AaronDonahue:
The two variables that track precipitation in P3 are `precip_liq_surf` and `precip_ice_surf`.  These are designated as "Computed" when added as fields in P3.  Which means the old `precip_X_surf` values are overwritten everytime P3 is called.

Those variables are not actually used until surface coupling is called.

In cases where P3 is subcycled we need to keep a cumulative value of `precip_liq_surf` and `precip_ice_surf` to pass to the surface.  Otherwise total water mass will not be conserved.

Note - this issue most likely impacts any "flux" type variable that could be substepped.

AaronDonahue:
Summary of conversation via Slack with @bartgol -

The solution should be a mechanism by which and field that is a "flux" is able to accumulate until it is a) used or b) needs to be reset.  This can be accomplished by the following strategy

1. Change all fluxes in atmosphere processes from "Computed" to "Updated".  That way old values are retained and fluxes can be accumulated.
2. Add local flux calculations to the memory buffer for a process, that way we can track the local addition to the flux and simply add it to the overall flux in post-processing.
3. We need a way to check if a flux should be reset to 0.0.
  -  We could have the process that uses the flux reset it. 
  -  We could keep track of the timestamp for the beginning of timestep and check if this call to run_impl is at the beginning, if so reset the flux, if not, accumulate it.

The second approach is probably the safest, it avoids resetting a flux that may be used in more than one place, and avoids a separate process having to "know" to reset the flux.

ambrad:
What does EAM do in this case?

bartgol:
Imho, storing the atm timestep initial timestamp in each atm process, and have p3 check if `timestamp()==atm_timestep_start()` at the beginning of run_impl (and zero out fluxes if true), is the fastest way to achieve the desired result. Of course, all fluxes computation in p3 needs to be changed from `flux_blah = ...` to `flux_blah += ...`.

AaronDonahue:
```
  // Note: if we are being subcycled from the outside, the host will set
  //       do_update=false, and we will not update the timestamp of the AP
  //       or that of the output fields.
  void set_update_time_stamps (const bool do_update);
```

@bartgol , does this comment and line mean that when subcycled the timestamp internal to this atmosphere process will not be updated be subcycle step?  Note this can be found on line 124 of `/components/scream/src/share/atm_process/atmosphere_process.hpp`

[...]

PR #44, opened 2019-02-06
Add ctest script for driving dashboard testing

jgfouca:

PR #43, opened 2019-02-05
Add P3 to EAM

AaronDonahue:
This commit introduces two new files,

micro_p3_interface.F90
micro_p3_utils.F90

and makes fundamental changes to the file,

micro_p3.F90

The interface code sets up an interface between the E3SM infrastructure
and the main P3 subroutines.  Similar in structure to the interface
for MG, the interface code initializes P3, initializes the PBUF variables
associated with P3, converts physics state variables to work with P3,
and conducts any post-processing of P3 output to be compatible with E3SM.

The utils code includes utility functions necessary to make P3 work,
including the loading of all constants used in P3.

There was an inconsistentcy between how E3SM defines the ratio
between temperature and potential temperature (exner) which has been
fixed.

The main micro_p3 routines have been adjusted to be consistent with
E3SM and to recieve the appropriate input/output variables needed to
interface with E3SM.

The microp_driver has been changed to call P3 routines when P3 microphysics
is chosen as the microphysics scheme.

Lastly, there were some minor configuration changes needed to get this
version of E3SM running properly on LC machines.

PR #1773, opened 2022-07-07
Fix how precipitation is handled so that it is still correct with subcycling.

AaronDonahue:
This commit adjusts how precip_liq_surf and precip_ice_surf is handled in EAMxx to accommodate subcycling.

Now, the precipitation flux provided by P3 is converted into a total precipitation mass (units=kg), and the precipitation mass
is accumulated every p3 run step.

In the surface coupling export, where precipitation is used, the total mass is converted to a mass flux for use by the component coupler.  After the precipitation is exported it is no longer needed this time step, so surface coupling is responsible for resetting the accumulated precipitation to 0.0.

This commit is non-BFB in that precipitation output variables will now be in units of kg rather than m/s.

Addresses issue #1755 
Will be improved when issue #1767 is eventually addressed.

### DIFF
--- a/components/scream/src/control/tests/sc_tests.cpp
+++ b/components/scream/src/control/tests/sc_tests.cpp
@@ -284,8 +284,8 @@ TEST_CASE ("recreate_mct_coupling")
   FID horiz_winds_id     ("horiz_winds",     vector3d_layout, m/s,    grid_name);
   FID pseudo_density_id  ("pseudo_density",  scalar3d_layout, Pa,     grid_name);
   FID qv_id              ("qv",              scalar3d_layout, nondim, grid_name);
-  FID precip_liq_surf_id ("precip_liq_surf", scalar2d_layout, m/s,    grid_name);
-  FID precip_ice_surf_id ("precip_ice_surf", scalar2d_layout, m/s,    grid_name);
+  FID precip_liq_surf_id ("precip_liq_surf_mass", scalar2d_layout, kg,    grid_name);
+  FID precip_ice_surf_id ("precip_ice_surf_mass", scalar2d_layout, kg,    grid_name);
   FID sfc_flux_dir_nir_id("sfc_flux_dir_nir", scalar2d_layout, W/(m*m), grid_name);
   FID sfc_flux_dir_vis_id("sfc_flux_dir_vis", scalar2d_layout, W/(m*m), grid_name);
   FID sfc_flux_dif_nir_id("sfc_flux_dif_nir", scalar2d_layout, W/(m*m), grid_name);
--- a/components/scream/src/control/tests/sc_tests.cpp
+++ b/components/scream/src/control/tests/sc_tests.cpp
@@ -26,6 +26,7 @@ TEST_CASE ("surface_coupling")
   using FR = FieldRequest;
   using GR = GroupRequest;
   using RPDF = std::uniform_real_distribution<Real>;
+  using IPDF = std::uniform_int_distribution<int>;
 
   // Some constants
   constexpr int ncols = 4;
@@ -38,6 +39,10 @@ TEST_CASE ("surface_coupling")
   // The random numbers generator
   auto engine = setup_random_test(&comm);
   RPDF pdf(0.0,1.0);
+  IPDF dt_pdf(1,1800);
+
+  // Set dt for test
+  const int dt = dt_pdf(engine);
 
   // Create a grid
   auto grid = create_point_grid("my grid",ncols*comm.size(), nlevs, comm);
@@ -180,7 +185,7 @@ TEST_CASE ("surface_coupling")
     std::fill_n(raw_data,ncols*num_fields,-1);
 
     // Perform export
-    exporter.do_export();
+    exporter.do_export(dt);
 
     // Perform import
     importer.do_import();
@@ -221,7 +226,7 @@ TEST_CASE ("recreate_mct_coupling")
    * This test aims to recreate the import/export of fields inside the mct-coupling
    * for CIME runs. Much of the code inside surface coupling is hard-coded for these runs.
    */
-
+  
   // Some namespaces/aliases
   using namespace scream;
   using namespace ShortFieldTagsNames;
@@ -231,6 +236,7 @@ TEST_CASE ("recreate_mct_coupling")
   using FR     = FieldRequest;
   using GR     = GroupRequest;
   using RPDF   = std::uniform_real_distribution<Real>;
+  using IPDF = std::uniform_int_distribution<int>;
   using C = scream::physics::Constants<Real>;
 
   // Some constants
@@ -244,6 +250,10 @@ TEST_CASE ("recreate_mct_coupling")
   // The random numbers generator
   auto engine = setup_random_test(&comm);
   RPDF pdf(0.0,1.0);
+  IPDF dt_pdf(1,1800);
+
+  // Set dt for test
+  const int dt = dt_pdf(engine);
 
   // Create a grid
   auto grid = create_point_grid("my_grid",ncols*comm.size(), nlevs, comm);
@@ -397,6 +407,15 @@ TEST_CASE ("recreate_mct_coupling")
   auto p_int_h            = p_int_f.get_view<Real**,Host>();
   auto phis_h             = phis_f.get_view<Real*,Host>();
 
+  // We need a copy of precip to check export is correct, since precip will be zeroed out after do_export is called.
+  // TODO: This step can be undone once we have the AD take care of resetting precip fluxes.
+  auto precip_liq_surf_copy_f = precip_liq_surf_f.clone(); 
+  auto precip_ice_surf_copy_f = precip_ice_surf_f.clone(); 
+  auto precip_liq_surf_copy_d = precip_liq_surf_copy_f.get_view<Real*>();
+  auto precip_ice_surf_copy_d = precip_ice_surf_copy_f.get_view<Real*>();
+  auto precip_liq_surf_copy_h = precip_liq_surf_copy_f.get_view<Real*,Host>();
+  auto precip_ice_surf_copy_h = precip_ice_surf_copy_f.get_view<Real*,Host>();
+
   // Create SC object and set number of import/export fields
   control::SurfaceCoupling coupler(fm);
   coupler.set_num_fields(num_cpl_imports, num_scream_imports, num_cpl_exports);
@@ -497,6 +516,10 @@ TEST_CASE ("recreate_mct_coupling")
     ekat::genRandArray(pseudo_density_d,engine,pdf);
     ekat::genRandArray(precip_liq_surf_d,engine,pdf);
     ekat::genRandArray(precip_ice_surf_d,engine,pdf);
+    // Copy precip_liq/ice_surf to the copy we need for comparison
+    Kokkos::deep_copy(precip_liq_surf_copy_d,precip_liq_surf_d);
+    Kokkos::deep_copy(precip_ice_surf_copy_d,precip_ice_surf_d);
+    // TODO: These deep copies won't be needed when the AD handles resetting precip
     ekat::genRandArray(sfc_flux_lw_dn_d,engine,pdf);
     ekat::genRandArray(sfc_flux_dir_nir_d,engine,pdf);
     ekat::genRandArray(sfc_flux_dir_vis_d,engine,pdf);
@@ -520,7 +543,7 @@ TEST_CASE ("recreate_mct_coupling")
     coupler.do_import();
 
     // Perform export
-    coupler.do_export();
+    coupler.do_export(dt);
 
     // Sync host to device
     surf_evap_f.sync_to_host();
@@ -569,8 +592,8 @@ TEST_CASE ("recreate_mct_coupling")
       REQUIRE (export_raw_data[4 + icol*num_cpl_exports]  == T_mid_h          (icol,    nlevs-1)); // 5th export
       REQUIRE (export_raw_data[6 + icol*num_cpl_exports]  == qv_h             (icol,    nlevs-1)); // 7th export
       REQUIRE (export_raw_data[7 + icol*num_cpl_exports]  == p_mid_h          (icol,    nlevs-1)); // 8th export
-      REQUIRE (export_raw_data[15 + icol*num_cpl_exports] == C::RHO_H2O*precip_liq_surf_h(icol));  // 16th export
-      REQUIRE (export_raw_data[16 + icol*num_cpl_exports] == C::RHO_H2O*precip_ice_surf_h(icol));  // 17th export
+      REQUIRE (export_raw_data[15 + icol*num_cpl_exports] == precip_liq_surf_copy_h(icol) / dt  ); // 16th export
+      REQUIRE (export_raw_data[16 + icol*num_cpl_exports] == precip_ice_surf_copy_h(icol) / dt  ); // 17th export
       REQUIRE (export_raw_data[17 + icol*num_cpl_exports] == sfc_flux_lw_dn_h(icol));              // 18th export
       REQUIRE (export_raw_data[18 + icol*num_cpl_exports] == sfc_flux_dir_nir_h(icol));            // 19th export
       REQUIRE (export_raw_data[19 + icol*num_cpl_exports] == sfc_flux_dir_vis_h(icol));            // 20th export
@@ -598,6 +621,10 @@ TEST_CASE ("recreate_mct_coupling")
       REQUIRE (export_raw_data[33 + icol*num_cpl_exports] == 0);
       REQUIRE (export_raw_data[34 + icol*num_cpl_exports] == 0);
       REQUIRE (export_raw_data[35 + icol*num_cpl_exports] == 0); // 35th export
+
+      // precipitation should be set to zero.  TODO: Won't be needed when AD handles resetting precipitation.
+      REQUIRE (precip_liq_surf_h(icol) == 0 ); 
+      REQUIRE (precip_ice_surf_h(icol) == 0 ); 
     }
   }
 
@@ -621,6 +648,7 @@ TEST_CASE ("do_initial_export")
   using FL = FieldLayout;
   using FID = FieldIdentifier;
   using RPDF = std::uniform_real_distribution<Real>;
+  using IPDF = std::uniform_int_distribution<int>;
 
   // Some constants
   constexpr int ncols = 4;
@@ -633,6 +661,10 @@ TEST_CASE ("do_initial_export")
   // The random numbers generator
   auto engine = setup_random_test(&comm);
   RPDF pdf(0.0,1.0);
+  IPDF dt_pdf(1,1800);
+
+  // Set dt for test
+  const int dt = dt_pdf(engine);
 
   // Create a grid
   auto grid = create_point_grid("my grid",ncols*comm.size(), nlevs, comm);
@@ -683,7 +715,7 @@ TEST_CASE ("do_initial_export")
     ekat::genRandArray(f2_exp_d,engine,pdf);
 
     // Perform export with init_phase=true
-    exporter.do_export(true);
+    exporter.do_export(dt,true);
 
     // Check f1 was exported, but f2 was set to 0
     f1_exp.sync_to_host();
@@ -697,7 +729,7 @@ TEST_CASE ("do_initial_export")
     std::fill_n(raw_data,ncols*num_fields,-1);
 
     // Perform export with init_phase=false (default)
-    exporter.do_export();
+    exporter.do_export(dt);
 
     // Check that both f1 and f2 were exported
     f1_exp.sync_to_host();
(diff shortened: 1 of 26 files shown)


AaronDonahue:
@PeterCaldwell , I had to change the postcondition checks for precipitation in this PR.  Could you check that the bounds I put are reasonable when precipitation is in the units of `kg`?

PeterCaldwell:
what happens in standalone tests including all the procs except surface coupling? I assume they're fine because they don't substep P3, but if they did, I don't think they'd ever zero out accumulated variables. Right? Maybe we can make runs that try to substep but don't have surface coupling error out? Not necessarily in this PR - we could create an issue to fix this.

components/scream/src/control/tests/sc_tests.cpp
```diff
@@ -397,6 +407,15 @@ TEST_CASE ("recreate_mct_coupling")
   auto p_int_h            = p_int_f.get_view<Real**,Host>();
   auto phis_h             = phis_f.get_view<Real*,Host>();
 
+  // We need a copy of precip to check export is correct, since precip will be zeroed out after do_export is called.
+  // TODO: This step can be undone once we have the AD take care of resetting precip fluxes.
+  auto precip_liq_surf_copy_f = precip_liq_surf_f.clone(); 
+  auto precip_ice_surf_copy_f = precip_ice_surf_f.clone(); 
+  auto precip_liq_surf_copy_d = precip_liq_surf_copy_f.get_view<Real*>();
+  auto precip_ice_surf_copy_d = precip_ice_surf_copy_f.get_view<Real*>();
+  auto precip_liq_surf_copy_h = precip_liq_surf_copy_f.get_view<Real*,Host>();
```
PeterCaldwell, line 416: I'm not sure you need _copy_f and _copy_d since these are just used for storing data you're never going to change, then are only using for tests on host. Am I missing something? Note as well that other files use this same paradigm and should also be changed (in the unlikely event that my criticism is well placed).

components/scream/src/physics/p3/atmosphere_microphysics.cpp
```diff
@@ -96,12 +96,13 @@ void P3Microphysics::set_grids(const std::shared_ptr<const GridsManager> grids_m
...
   // Diagnostic Outputs: (all fields are just outputs w.r.t. P3)
-  add_field<Computed>("precip_liq_surf",    scalar2d_layout,     m/s,    grid_name);
-  add_field<Computed>("precip_ice_surf",    scalar2d_layout,     m/s,    grid_name);
+  add_field<Computed>("precip_liq_surf",    scalar2d_layout,     kg,     grid_name);
+  add_field<Computed>("precip_ice_surf",    scalar2d_layout,     kg,     grid_name);
   add_field<Computed>("eff_radius_qc",      scalar3d_layout_mid, micron, grid_name, ps);
   add_field<Computed>("eff_radius_qi",      scalar3d_layout_mid, micron, grid_name, ps);
 
   // History Only: (all fields are just outputs and are really only meant for I/O purposes)
+  // TODO: question - should these be averaged over subcycle as well?
```
PeterCaldwell, line 105: Yeah, these should also be averaged over subcycles. If we don't have bandwidth for doing that now, you could just delete them (since they are only used for output) and create a git issue to re-add them correctly. Though I guess that takes re-wiring... Just creating a git issue and putting a warning comment in the code here would be fine.

components/scream/src/physics/p3/atmosphere_microphysics.cpp
```diff
@@ -198,8 +205,8 @@ void P3Microphysics::initialize_impl (const RunType /* run_type */)
   add_postcondition_check<FieldWithinIntervalCheck>(get_field_out("nr"),m_grid,0.0,1.e9,false);
   add_postcondition_check<FieldWithinIntervalCheck>(get_field_out("ni"),m_grid,0.0,1.e9,false);
   add_postcondition_check<FieldWithinIntervalCheck>(get_field_out("bm"),m_grid,0.0,1.0,false);
-  add_postcondition_check<FieldWithinIntervalCheck>(get_field_out("precip_liq_surf"),m_grid,0.0,0.001,false);
-  add_postcondition_check<FieldWithinIntervalCheck>(get_field_out("precip_ice_surf"),m_grid,0.0,0.001,false);
+  add_postcondition_check<FieldWithinIntervalCheck>(get_field_out("precip_liq_surf"),m_grid,0.0,100.0,false);
+  add_postcondition_check<FieldWithinIntervalCheck>(get_field_out("precip_ice_surf"),m_grid,0.0,100.0,false);
```
PeterCaldwell, line 209: our new precip variables will have magnitude which scales by dt, so it's hard to say how big they should be. To solve this, I'd suggest converting them to the old flux-form precip rates for these condition checks. In other words, replace the line above with `...get_field_out("precip_liq_surf")/dt/C::RHO_LIQ, m_grid,0.0,0.001,false)`

PeterCaldwell:
A couple random thoughts:
1. if we called surface coupling export before writing output, the precip output would all be zeroes right? And don't we call export at the end of each atm step?
2. I'd greatly prefer if precip_liq_surf and precip_ice_surf got renamed to precip_mass_liq_surf and precip_mass_ice_surf to reflect the fact that they're not fluxes. I think this can be done via simple search/replace but I'm ok with skipping that change for now in the interest of time.

components/scream/src/physics/p3/atmosphere_microphysics.hpp
```diff
@@ -246,9 +247,11 @@ class P3Microphysics : public AtmosphereProcess
         diag_eff_radius_qc(icol,ipack) *= 1e6;
         diag_eff_radius_qi(icol,ipack) *= 1e6;
       } // for ipack
+      precip_liq_surf(icol) += precip_liq_surf_local(icol) * PC::RHO_H2O * m_dt;
+      precip_ice_surf(icol) += precip_ice_surf_local(icol) * PC::RHO_H2O * m_dt;
```
PeterCaldwell, line 251: dumb question, but isn't precip in m/s = precip mass in kg *1/RHO_H2O/dt? I don't understand why you're multiplying here.

PeterCaldwell: also, maybe call _local _flux instead?

PeterCaldwell: Oh, my 2nd comment answers my first: The RHS is m/s and the LHS is in kg. Nevermind...

components/scream/src/physics/p3/atmosphere_microphysics.hpp
```diff
@@ -246,9 +247,11 @@ class P3Microphysics : public AtmosphereProcess
         diag_eff_radius_qc(icol,ipack) *= 1e6;
         diag_eff_radius_qi(icol,ipack) *= 1e6;
       } // for ipack
+      precip_liq_surf(icol) += precip_liq_surf_local(icol) * PC::RHO_H2O * m_dt;
```
PeterCaldwell, line 250: I'd rather _local was _flux to clarify what it means...

components/scream/src/physics/p3/atmosphere_microphysics.cpp
```diff
@@ -288,7 +297,8 @@ void P3Microphysics::initialize_impl (const RunType /* run_type */)
   p3_postproc.set_variables(m_num_cols,nk_pack,prog_state.th,pmid,T_atm,t_prev,
       prog_state.qv, prog_state.qc, prog_state.nc, prog_state.qr,prog_state.nr,
       prog_state.qi, prog_state.qm, prog_state.ni,prog_state.bm,qv_prev,
-      diag_outputs.diag_eff_radius_qc,diag_outputs.diag_eff_radius_qi);
+      diag_outputs.diag_eff_radius_qc,diag_outputs.diag_eff_radius_qi,
+      diag_outputs.precip_liq_surf,diag_outputs.precip_ice_surf,precip_liq_surf,precip_ice_surf);
```
PeterCaldwell, line 301: Apologies if I'm just not reading the code deeply enough, but my understanding was that p3_main would only know about precip_flux variables and the AD would only know about precip_mass variables and this interface layer is the only place that would know about and translate between these versions. I don't see how this layer is doing that. Maybe renaming the variables _mass and _flux like I suggested above would help!

PeterCaldwell:
> Why do the units have to change from m/s, a flux, to kg, an accumulated mass? EAM's macmic loop uses m/s.

If you've got a couple free months you can read all about this in https://github.com/E3SM-Project/scream/issues/1755 . 

The TL;DR version is that kg units are convenient for accumulating precip over multiple substeps: the precip over N substeps is just the sum over those substeps. Accumulating fluxes is awkward because if you have 2 mm/day in both substep 1 and substep 2, the accumulated flux is the average rather than the sum. But if you average the first 2 substeps and then the 3rd substep has 1mm/day precip, you can't just average the 2-step average with the new average because the weights are different. So you end up doing something like "substep contribution to total precip is substep precip mass * dt_substep/dt_whole_step" which is more complicated and doesn't really make sense in a single substep. Sorry, my explanation is almost as long as the git issue!

ambrad:
> So you end up doing something like "substep contribution to total precip is substep precip mass * dt_substep/dt_whole_step" which is more complicated and doesn't really make sense in a single substep

Thanks for the pointer to the issue. Just to state the obvious, one can do the multiplication by dt_substep/dt_whole_step at the end. And then dt_substep/dt_whole_step = 1/nsubstep. So in summary one accumulates N times and then divides by N.

components/scream/src/control/tests/sc_tests.cpp
AaronDonahue: @tcclevenger can correct me if I'm wrong, but I'm pretty sure the REQUIRE statements are done on HOST but the actual export is done on DEVICE so we need both views, and to make sure they are synced.

components/scream/src/mct_coupling/scream_cxx_f90_interface.cpp
```diff
@@ -149,7 +149,7 @@ void scream_setup_surface_coupling (
 
     // After registration we need to export all fields (except those that require scream computations)
     // as other models will need these values.
-    sc->do_export(true);
+    sc->do_export(0.0,true);
```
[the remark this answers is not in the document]
AaronDonahue: Good point, my understanding is that do_export with a true flag passed is just for initialization and shouldn't do anything with fields that are produced by scream, thus we shouldn't do anything with precip.  That being said, it is worth confirming.  Although maybe forcing an error with a divide by zero would be a good way to make sure we don't accidentally pass an unset precip to the surface models.

components/scream/src/physics/p3/atmosphere_microphysics.cpp
AaronDonahue: I should be able to add them without too much difficulty.

components/scream/src/physics/p3/atmosphere_microphysics.cpp
AaronDonahue: Yeah, that's what I thought we may need to do.  I didn't do something scaled by `dt` originally because `dt` is not passed to the initialize_impl, so the information is unknown when the postcondition check is set.  I feel like making it so this is possible may be more work than this PR is set out to do.  Maybe I could suppress these checks for now?  @PeterCaldwell ?

components/scream/src/physics/p3/atmosphere_microphysics.hpp
AaronDonahue: So do I, but there is already a variable in P3 and in the memory buffer of the interface called `precip_liq/ice_flux`.  I didn't want to go down the rabbit hole of whether or not that variable could be (a) changed in name or (b) used instead of the precip_liq/ice_surf that P3 creates.  I suspect the latter.  The _flux versions are just 3D variables that record the precipitation at every layer, including the surface.

components/scream/src/physics/p3/atmosphere_microphysics.cpp
AaronDonahue: The naming convention is probably to blame for the confusion.  What I've done here is create a local precip_liq/ice_surf (the flux leaving the surface) variable that will be used inside p3_main which is stored in the memory buffer of the interface.  Then I pass to the post-process structure the local and field managed versions so the field managed version can accumulate the precip from this substep.  The `diag_outputs` versions are the ones used by p3_main.  The other two are pointing to the field managed.

PeterCaldwell: Ok, that was my expectations. Using _flux and _mass would help clarify.

components/scream/src/physics/p3/atmosphere_microphysics.cpp
PeterCaldwell: Ok.

components/scream/src/physics/p3/atmosphere_microphysics.cpp
PeterCaldwell: Let's instead set the max values to a very big number. I have a vague recollection that "Inf" or something is an allowable value? I think there's a lot of benefit in making sure precip isn't negative even if we don't know how big the max can be.

components/scream/src/control/tests/sc_tests.cpp
PeterCaldwell: Ah, ok. I see now that the "_copy" version is needed to check that the data exported is the expected value of precip *before* sfc cpl zeroed things out. And I assume you get_view is assigning the pointer to the data rather than copying the data itself so we can't skip the _copy_f lines by directly assigning surf_f to copy_d and copy_h, right?

components/scream/src/physics/p3/atmosphere_microphysics.hpp
PeterCaldwell: Uh oh. It doesn't bug me that we already have `precip_liq_flux` in the FM since `precip_liq_**surf**_flux` is a different variable. Optimal would be to name `precip_liq_flux`=> `precip_liq_**3d**_flux` to disambiguate, but maybe that's too much to ask. 

The bigger problem is that *all* fluxes should be averaged over their substeps and it sounds like we missed this one (well, two since there's also `precip_ice_flux`). Is it actually used for anything other than output?

components/scream/src/physics/p3/atmosphere_microphysics.cpp
AaronDonahue: Okay, that just means switch the check to a lower-bound check rather than an interval check.

components/scream/src/control/tests/sc_tests.cpp
AaronDonahue: Yes, we should be able to do it that way.  I think in order to do it I will have to first define 1d views on host and device in the test so that I can create and copy them.  I figured it was the same amount of code so I went with the simpler option since this is a test that will probably get replaced when @tcclevenger changes surface coupling to it's own process.  I'm happy either way.

components/scream/src/physics/p3/atmosphere_microphysics.hpp
AaronDonahue: TBH I didn't dig too deep to see what it is used for.  I can make surf_flux the naming convention.

components/scream/src/physics/p3/atmosphere_microphysics.cpp
AaronDonahue: I just realized, there is no easy mechanism yet to zero these out every timestep.  So maybe we add a warning here in the comments and a git-issue to resolve it later.

components/scream/src/physics/p3/atmosphere_microphysics.cpp
PeterCaldwell: Oh yeah, you could just do that ;-).

components/scream/src/physics/p3/atmosphere_microphysics.hpp
PeterCaldwell: 👍

components/scream/src/physics/p3/atmosphere_microphysics.cpp
PeterCaldwell: Ah yes, because these variables aren't used by surface coupling! How logical!!! Yeah, let's just add a warning in the code and a v1.1 issue in git.

PeterCaldwell:
Mappy is failing baselines - I think this is expected because we've redefined the units to precip. 
Blake had indigestion - retesting.

PeterCaldwell:
Passes all tests except CIME diffs on mappy, which are expected because we changed the units for precip variables. Manually integrating.

bartgol:
> Standalone tests were also failing, actually. We'll ninja fix this.

bartgol:
Haha, ok, I'll close an eye, but don't tell my boss!

There are a couple of signs of standalone fails. The first is the top of the log (say, on mappy):
```
213:homme_shoc_cld_spa_p3_rrtmgp_128levels_np3
214:homme_shoc_cld_spa_p3_rrtmgp_128levels_np4
215:homme_shoc_cld_spa_p3_rrtmgp_128levels_np1_vs_np2_bfb
216:homme_shoc_cld_spa_p3_rrtmgp_128levels_np1_vs_np3_bfb
217:homme_shoc_cld_spa_p3_rrtmgp_128levels_np1_vs_np4_bfb
Error(s) occurred during test phase
OVERALL STATUS: FAIL
Starting analysis on mappy with cmd: cd /home/e3sm-jenkins/jenkins-ws/workspace/SCREAM_PullRequest_Autotester_Mappy/2805/scream/components/scream && module purge && module load sems-archive-env sems-archive-gcc/9.2.0 sems-archive-cmake/3.19.1 sems-archive-git/2.10.1 sems-archive-openmpi/4.0.2 sems-archive-netcdf/4.7.3/parallel && export OMP_PROC_BIND=spread && true &&  ./scripts/test-all-scream --baseline-dir AUTO $compiler -c EKAT_DISABLE_TPL_WARNINGS=ON  -p -i -m mappy
RUN: cd /home/e3sm-jenkins/jenkins-ws/workspace/SCREAM_PullRequest_Autotester_Mappy/2805/scream/components/scream && module purge && module load sems-archive-env sems-archive-gcc/9.2.0 sems-archive-cmake/3.19.1 sems-archive-git/2.10.1 sems-archive-openmpi/4.0.2 sems-archive-netcdf/4.7.3/parallel && export OMP_PROC_BIND=spread && true &&  ./scripts/test-all-scream --baseline-dir AUTO $compiler -c EKAT_DISABLE_TPL_WARNINGS=ON  -p -i -m mappy
FROM: /home/e3sm-jenkins/jenkins-ws/workspace/SCREAM_PullRequest_Autotester_Mappy/2805/scream/components/scream
mappy failed
```
The second is a bit subtle, and you must know the scripts content to know it signals a standalone fail.
```
fails=1+1
```
right after the v1 tests, means that the number of fails was 1, and since v1 failed is getting updated to 2. But the only thing tested before v1 was standalone.

Anyhow, I think the fix is a matter of updating input files:
```
18:23:28 [EAMXX] initialize_fields ...
18:23:28   [EAMXX] Run  start time stamp: 2021-10-12-45000
18:23:28   [EAMXX] Case start time stamp: 2021-10-12-45000
18:23:28   [EAMXX] set_initial_conditions ...
18:23:28      -49                                                             PIO ERROR: could not find variable precip_ice_surf_mass in file /sems-data-store/ACME/inputdata/atm/scream/init/screami_ne2np4L72_20220524.nc
```
Does this sound right? I didn't realize we were adding an atm input.

bartgol:
> [...] Or do we need to capture more lines of the log file in these PR pages?

Even though there were signs in the log (as I explained above), I do think we should improve our jenkins scripts, by printing an ALL CAPS message at the end, which states which test suites had failures. It would be nice to also list the tests (within each testsuite) that failed, but that's probably not as important as scream-ing out that standalone failed, or v0, or v1, or something else. Especially as the number of tests increases, we run the risk of hiding some important output more than 500 lines before the end.

PeterCaldwell:
I thought we *did* print out which tests had fails. Grabbing the output from some other PR, I see

>43/45 Test #43: shoc_p3_subcycled ................   Passed    1.52 sec
      Start 44: shoc_p3_monolithic
44/45 Test #44: shoc_p3_monolithic ...............   Passed    1.39 sec
      Start 45: check_subcycling
45/45 Test #45: check_subcycling .................   Passed    0.06 sec
100% tests passed, 0 tests failed out of 45

It seems like maybe there's a bug in our AT scripts which causes these difficult-to-understand messages when tests fail? I recall that our scripts used to report fails in the same way as the passes I copied above...

PeterCaldwell:
I think we can fix this via `ncrename -h -O -v old_variable_name,new_variable_name filename.nc`. Do we have a list of which files need to be corrected?

bartgol:
It's actually simpler, I think. Notice that _some_ tests involving p3 passed (e.g., p3_standalone). The precip fields are inited to 0 in the passing tests, so I can do the same in the failing ones.

bartgol:
Anyhow, I think I managed to fix all failing tests. I will submit quick pr in a second.

bartgol:
Wait, I thought those variables were "time-step-local" accumulation variables. Aren't they?

PeterCaldwell:
Oh yeah, you're right. Surface precip flux is not a prognostic variable. I got mixed up with qr and nr, which are prognostic. So setting to zero should be fine.

PR #1776, opened 2022-07-08
Switch the pressure coordinate in P3 to be dry instead of wet.

AaronDonahue:
Because the mass-mixing-ratios in P3 are dry, the pressure coordinates
are inconsistent, which leads to a mass leak during sedimentation in P3.

This commit introduces the pdel_dry and pmid_dry variables which are passed
to p3_main which represent the conversion of the wet pressure coordinates to
a dry set.

This commit also introduces a helper function for standalone tests that will
allow the user to calculate the total global water mass.  This can be used to
ensure that any mass leaks are sufficiently small, as shown in the application
of this new function in the p3_standalone test.

### DIFF
--- a/components/scream/src/physics/p3/atmosphere_microphysics.hpp
+++ b/components/scream/src/physics/p3/atmosphere_microphysics.hpp
@@ -77,13 +77,16 @@ public:
         const Spack& pmid_pack(pmid(icol,ipack));
         const Spack& T_atm_pack(T_atm(icol,ipack));
         const Spack& cld_frac_t_pack(cld_frac_t(icol,ipack));
+        // Ensure that the pseudo_density and pmid used by P3 is dry
+        pdel_dry(icol,ipack) = pseudo_density(icol,ipack) * (1-qv(icol,ipack));
+        pmid_dry(icol,ipack) = pmid_pack * (1-qv(icol,ipack));
         // Exner
-        const auto& exner = PF::exner_function(pmid_pack);
+        const auto& exner = PF::exner_function(pmid_dry(icol,ipack));
         inv_exner(icol,ipack) = 1.0/exner;
         // Potential temperature
-        th_atm(icol,ipack) = PF::calculate_theta_from_T(T_atm_pack,pmid_pack);
+        th_atm(icol,ipack) = PF::calculate_theta_from_T(T_atm_pack,pmid_dry(icol,ipack));
         // DZ
-        dz(icol,ipack) = PF::calculate_dz(pseudo_density(icol,ipack), pmid_pack, T_atm_pack, qv(icol,ipack));
+        dz(icol,ipack) = PF::calculate_dz(pdel_dry(icol,ipack), pmid_dry(icol,ipack), T_atm_pack, qv(icol,ipack));
         // Cloud fraction
         // Set minimum cloud fraction - avoids division by zero
         cld_frac_l(icol,ipack) = ekat::max(cld_frac_t_pack,mincld);
@@ -120,6 +123,7 @@ public:
         // ^^ Ensure that qv is "wet mmr" till this point ^^
         //NOTE: Convert "qv" to dry mmr in the end after converting all other constituents to dry mmr
         qv(icol, ipack)      = PF::calculate_drymmr_from_wetmmr(qv(icol,ipack),qv(icol,ipack));
+
 
         // update rain cloud fraction given neighboring levels using max-overlap approach.
         for (int ivec=0;ivec<Spack::n;ivec++)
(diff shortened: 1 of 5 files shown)


AaronDonahue:
I'm not sure if the second commit should really be part of this PR.  I included it in case someone else wants to work with this branch and wants quick access to a mass checker.  Perhaps once we are comfortable with the PR as is we can simply remove the commit that includes the mass checker helpers.

AaronDonahue:
note, this PR fixes the issue we are seeing with sedimentation, but there appears to still be a leak tied to `p3_main_part2`.  This can be confirmed by commenting out `p3_main_part2` in `p3_main_impl.hpp` and rerunning the p3_standalone test on this branch.  You see a precipitous drop in the mass leak.  Commenting out sedimentation on the other hand doesn't make any sizeable difference.

AaronDonahue:
Note, so far all of the changes have been on the C++ side.  We still need to make the equivalent changes to the EAM F90 side for the p3 comparison tests to pass.

components/scream/src/physics/p3/atmosphere_microphysics.hpp
```diff
@@ -77,13 +77,16 @@ class P3Microphysics : public AtmosphereProcess
         const Spack& pmid_pack(pmid(icol,ipack));
         const Spack& T_atm_pack(T_atm(icol,ipack));
         const Spack& cld_frac_t_pack(cld_frac_t(icol,ipack));
+        // Ensure that the pseudo_density and pmid used by P3 is dry
+        pdel_dry(icol,ipack) = pseudo_density(icol,ipack) * (1-qv(icol,ipack));
+        pmid_dry(icol,ipack) = pmid_pack * (1-qv(icol,ipack));
```
ambrad, line 82: pmid_dry in general cannot simply be multiplied by (1-qv); rather, one has to do a scan over pdel_dry.

ambrad: By the way, if your wiring plan was simply to add two or three fields -- pmiddry, pdeldry, and possibly pmiddry -- for general use by any parameterization in homme_post_process, I support that. That's exactly what EAM does. In particular, it might be natural to add these to HommeDynamics::update_pressure.

components/scream/src/share/util/standalone_helper_functions.hpp
```diff
@@ -0,0 +1,96 @@
+#ifndef SCREAM_STANDALONE_TEST_UTILS_HPP
+#define SCREAM_STANDALONE_TEST_UTILS_HPP
+
+#include "share/field/field_manager.hpp"
+#include "share/grid/mesh_free_grids_manager.hpp"
+
+namespace scream {
+
+// Helper function to check the total water mass
+Real calculate_water_mass(const std::shared_ptr<const GridsManager>& grids_mgr,const std::shared_ptr<FieldManager>& field_mgr, const Real dt, const bool use_precip) {
```
ambrad, line 10: Here I suggest checking per column rather than over all columns that happen to be on the process, since conservation should hold in each column.

components/scream/src/physics/p3/atmosphere_microphysics.hpp
PeterCaldwell: > pmid_dry in general cannot simply be multiplied by (1-qv); rather, one has to do a scan over pdel_dry.

Yes, I agree that dry pressure at a given level needs should be the weight of all the dry pressure thicknesses above it. As written, all levels except for the one computed are still wet pressure thickness. I feel dumb for saying, but I'm not sure how to implement the needed cumulative sum on packs which are already inside a parallel for loop... There's also the issue that pdel gives the mass between top and bottom of a cell so cumsum over pdel gives interface pressures which then need to be converted to midlevel pressures...

ambrad: @PeterCaldwell see e.g. HommeDynamics::update_pressure re: the loop mechanics; and note that I suggest adding pmiddry, pdeldry, and possibly pintdry to that routine, following EAM's example.

AaronDonahue: @ambrad , thanks for catching that.  That was an oversight on my part.  In reference to what is done in EAM, I see this line https://github.com/E3SM-Project/scream/blob/master/components/eam/src/physics/cam/physics_types.F90#L1472 which appears to be where EAM sets the dry coordinates.  It also appears to be making the same mistake.  Can we say then that EAM is not doing things correctly either?  If so, I should probably change this in EAM but that would be an upstream PR.

AaronDonahue: oh wait, nevermind, I didn't pay attention to the next few lines. :-)

PR #1780, opened 2022-07-12
SurfaceCoupling as an ATM process

tcclevenger:
Adds Surface Coupling as an ATM process. 

New:
- 2 new ATM process types: `SurfaceCouplingImporter` and `SurfaceCouplingExporter`
- `SCDataManager` class which stores the cpl data coming from the mct-coupler (or standalone test) so that the AD can pass to ATM processes (similar to the `ATMBufferManager` class)
- New interface in the AD for connecting SCData and the SC atm processes
- Standalone test in `tests/coupled/surface_coupling_only/` which replicates the data structure used in CIME runs

Structure
- Call `ad.setup_surface_coupling_data_manager()` (either inside `mct-coupling/` or in the `standalone_test.cpp`) for both import and export to store cpl data inside `SCDataManager` object in the AD. 
- During `initializa_atm_procs()`, the AD calls `m_atm_process_group->setup_surface_coupling_processes()` for both import and export which passes `SCDataManager` object to the import/export atm processes so they can setup internals.
- The `SurfaceCouplingImporter` and `SurfaceCouplingExporter` ATM processes class define `initialize_impl()` and `run_impl()` as usual.


components/scream/src/control/surface_coupling_utils.hpp
```diff
@@ -0,0 +1,144 @@
...
+// 2D vector field (ncols, num_comp):
+//     col_offset=vec_comp,                              col_stride=num_comp
+// 3D scalar field (ncols, nlevs):
+//     col_offset=nlevs-1,                               col_stride=nlevs_with_padding
+// 3D vector field (ncols, num_comp, nlevs):
+//     col_offset=vec_comp*nlevs_with_padding + nlevs-1, col_stride=nlevs_with_padding*num_comp
+//
+// Field with parents are allowed if the parent has layout 3D vector (ncols, num_comp, nlevs),
+// and the field was sliced along num_comp. See comments below for details.
+inline void get_col_info_for_surface_values(const std::shared_ptr<const FieldHeader>& fh,
```
bartgol, line 73: This seems a bit long for an inline function. You could just put it in a cpp file.

components/scream/src/share/atm_process/SCDataManager.hpp
```diff
@@ -0,0 +1,105 @@
...
+struct SCDataManager {
+
+  template<typename DevT, typename DataT>
+  using view_1d = typename KokkosTypes<DevT>::template view_1d<DataT>;
+  template<typename DevT, typename DataT>
+  using view_2d = typename KokkosTypes<DevT>::template view_2d<DataT>;
+
+  using name_t = char[32];
+
+  SCDataManager() {}
```
bartgol, line 20: Since you do it for the destructor, you might use the "default" keyword here as well.

components/scream/src/control/atmosphere_driver.cpp
```diff
@@ -883,6 +908,15 @@ void AtmosphereDriver::initialize_atm_procs ()
 
   const bool restarted_run = m_case_t0 < m_run_t0;
 
+  if (m_surface_coupling_import_data_manager) {
+    m_atm_process_group->setup_surface_coupling_processes(SurfaceCouplingTransferType::Import,
```
bartgol, line 912: The `setup_surface_coupling_processes` method seems to just be needed by the SC processes, so injecting it into every atm proc seems a bit hacky. What do you think about the following:

- add it only as a method of the SC classes
- have the AD build the SC import/export atm processes
- the AD passes the SCDataMgr to the import/export atm procs when it creates them
- the AD stores the atm procs as a new atm proc group: [sc_import, atm_proc_group, sc_export]

The latter would require a new ctor for atm proc group, that takes in a set of pre-built atm processes, but should not be too hard.

The overall point being to avoid making base classes more complex only for the sake of a handful of derived ones.

bartgol (commented):
Looks like good progress. I have a few comments up for consideration.

components/scream/src/control/atmosphere_driver.cpp
tcclevenger: I agree. I think the 3rd option (pass SCDataMgr to import/export) would be quite simple. Then in `initialize_impl()` these procs could setup the data. As I look at it now, there is no reason for the separate `setup_surface_coupling_processes()`..

tcclevenger:
@bartgol Check my latest commit (where I remove the generic sc setup functions in AtmosphereProcess class). I went with route 1 , "add it only as a method of the SC classes", where the AD calls setup functions for each SC atm process. For this, I need to cast from Base class ptr to Derived class ptr (via `dynamic_ptr_cast`) to call derived-class-only functions. I often make simple mistakes with ptr casts, so please correct anything inefficient you see.

tcclevenger:
Depending on how the DIFF is calculated, this PR may introduce some. For example, `sfc_alb_diff_dir` was a `Computed` field in `atmosphere_radiation` on master, but in this PR it is a `Required` field in `atmosphere_radiation` and a `Computed` field in `SurfaceCouplingImporter`. Not sure if that would trigger anything..

tcclevenger:
I see we have some failures though in CIME tests. investigating.

bartgol:
> Depending on how the DIFF is calculated, this PR may introduce some. For example, `sfc_alb_diff_dir` was a `Computed` field in `atmosphere_radiation` on master, but in this PR it is a `Required` field in `atmosphere_radiation` and a `Computed` field in `SurfaceCouplingImporter`. Not sure if that would trigger anything..

CIME tests result only depends on the output nc files. A DIFF means that the generated NC files are different. Who computes each field should not be relevant.

tcclevenger:
Found the DIFF in the SCasATMPro PR. Basically, in `AD::run()` on master:
```
1. run all atm_proc->run_impl()

2. output_mgr.run() to update output streams

3. run surface_coupling->do_export() where precip_ice(liq)_surf_mass:=0
```
But on this branch, 
```
1. run all atm_proc->run_impl(), where export is done at the end, where precip_ice(liq)_surf_mass:=0

2. output_mgr.run() to update output streams
```
So `precip_ice(liq)_surf_mass=0` in the output on this branch.

@bartgol  Should I add AD routine that sets `precip_ice(liq)_surf_mass=0` instead of the export doing this? This seems to be a hacky version of what https://github.com/E3SM-Project/scream/issues/1767 is suggesting...

bartgol:
Yes, let's take it out of the SC, and put it at the end of the AD's run method. We'll work on that issue at some point.

bartgol (approved):

tcclevenger:
Forgot about the restart. Needed to set to zero there.

tcclevenger:
Bad news, our friend the ICE is back... and this time in an external lib....

From `SMS_D_Ln9.ne4_ne4.F2010-SCREAMv1-noAero.summit_gnugpu`:
```
/autofs/nccs-svm1_home1/tccleve/E3SM/SCREAM/scream/externals/
      ekat/extern/spdlog/include/spdlog/detail/circular_q.h:30:3: 
internal compiler error: Segmentation fault
```

ambrad:
The spdlog version hasn't changed, right? What is the cpp file that is being compiled when that ICE occurs?

bartgol:
> The spdlog version hasn't changed, right? What is the cpp file that is being compiled when that ICE occurs?

Not in this PR, but notice that Jeff put in a PR that changed the version of spdlog in ekat master. Might be worth a shot updating the ekat submodule and see if it goes away...

ambrad:
Oh, that must be it. Jeff was fixing something on summit related to spdlog. So, Conrad, merge master into this PR (locally; no need to push that merge), updated the submodules, and try again.

tcclevenger:
Updating ekat solves that ICE, then another one exists in Homme boundary exchange. I have a branch that fixes these
Branch: https://github.com/E3SM-Project/scream/tree/tcclevenger/fix_ice_on_summit
Commit: a5ae4a4bf7dc37119c34164e7ea921563621e439 (Homme) 7d86c57e92cfb45815d6d60b75b42b1606bd4d0a (EKAT)

tcclevenger:
What is the process for submitting a change in `components/homme/`? I'm a bit confused by that..

bartgol:
You issue a PR in e3sm.

ambrad:
This is on Summit, right? We run v1 via CIME there all the time. Is this Homme BE ICE perhaps due to an out-of-date environment?

tcclevenger:
I don't think it would be outdated env. I'm running through `./create_test` and letting it load modules. On master branch, this works fine, it's only on the PRs branch.

ambrad:
You might merge master into this branch, then, as the CIME config for Summit on your branch may be outdated.

tcclevenger:
I rebased 3 days ago, but I can try again.

ambrad:
3 days ago is fine. @bartgol am I missing something? It doesn't make sense to me that this branch should have an ICE in a Homme cpp file and a refactor of a Homme _cpp_ file (not hpp file) should fix it. This PR is entirely orthogonal to that part of the build.

bartgol:
Uhm, you're right. The Homme cpp files are not affected by any change in this PR. I wonder if there has been a change in compile flags, that sneaked in. @tcclevenger can you compare flags.make for the `theta-l_kokkos_4_128_10` target (I believe that's the name) for both master and your branch on summit? You might have to look around to find its `.dir` folder, but I suspect it's in `$bindir/tests/uncoupled/homme/CMakeFIles`.

tcclevenger:
@bartgol You were right, on master, updating Kokkos in ekat results in the same homme ICE.

Weirdly though, on this PRs branch, if I update only spdlog and not kokkos, I get the ICE from spdlog..

bartgol:
@ambrad given what Conrad said, I think we can argue that the ICE is caused by an update of Kokkos in #EKAT/233. I think we can probably proceed with the upstream PR (that breaks the cpp file in two).

ambrad:
Ok. But I would caution against updating Kokkos just as we're sprinting to get the science run going by 1 Sep. Nothing wrong with bringing in the Homme part in an upstream PR, as long as that was tested with the current Kokkos version.

bartgol:
Hold on, your comment on slack made me rethink my position. This PR does not update EKAT, so my conjecture about this being due to a kokkos update is probably not right. If master builds, this branch should also build.

@tcclevenger when you said `on master, updating Kokkos in ekat results in the same homme ICE`, did you mean you had to manually change kokkos in master? Or was current master (with submodules all up to date) exhibiting the ICE?

tcclevenger:
@bartgol On SCREAM/master, I manually updated ekat/extern/kokkos/ to what is currently on EKAT/master, and then got the ICE. But SCREAM/master (with submodules up-to-date) runs fine.

ambrad:
Ok, Conrad's comment explains things. EKAT's Kokkos was inadvertantly updated. So for now we can't merge EKAT into SCREAM. But this PR is fine to merge if, with SCREAM master's current modules, everything is fine.

tcclevenger:
@ambrad As far as the AT is concerned, this PR is fine with current master modules. To run on Summit, one would need to update EKAT (for the spdlog update), and include the ICE fix in the E3SM/Homme PR.

bartgol:
EKAT/master is about to get updated, restoring kokkos 3.6 (as it is used in scream). I suspect that version of EKAT will fix this PR (will have the spdlog update, as well as the _right_ kokkos (3.6)). Conrad is checking that. If that works, we can update ekat in this PR, test again, and merge.

ambrad:
@bartgol isn't the spdlog fix already in SCREAM master's EKAT version? If so, this PR should not include any module updates. 

My best guess at reconstructing these confusing events is Conrad was testing this PR with the EKAT module checked out to EKAT master (which is currently broken) rather than to SCREAM master's version.

ambrad:
Adding myself as a reviewer. To protect Fort SCREAM, I won't approve until we have a clear statement of what happens when we take current SCREAM master (with modules up to date with `git submodule update --init --recursive), merge this PR branch into it, and run on Summit.

tcclevenger:
The issues stemmed from me running with EKAT having Kokkos 3.5 and SCREAM having Kokkos 3.6. After rebasing my branch here to scream master and updating submodules, both ICE mentioned (Homme boundary layer and spdlog) are gone. This branch PASSes with both SCREAM's version of EKAT (old spdlog) and with EKAT master (new spdlog), both have Kokkos 3.6 (which matches scream/external/kokkos)

I'm doing one last check of the values by creating a baseline on master for the lowres tests, and comparing with my branch. When I verify they pass, I'll update here.

bartgol:
Great! Once (and if) you verify that everything is good as is, please, go ahead and close the homme pr in the upstream repo.

ambrad:
Am I correct that the BFB label applies to this PR?

tcclevenger:
@ambrad yes you are

bartgol:
@tcclevenger So this PR does not need any additional testing, and can be merged, correct? To be clear, I'm assuming that:

- this branch (with master merged in) passes on all AT machines
- this branch (with master merged in) builds/runs on summit

bartgol:
You don't need to rebase. You can just put `AT: RETEST`, and let AT run again.

tcclevenger:
Yeah, might as well retest.

ambrad (approved):
The ICE issue was resolved and the PR is BFB, so I'm approving.

tcclevenger:
Some diffs of GPU after 9 steps. Investigating.

bartgol:
> Some diffs of GPU CIME runs. Investigating.

If the diffs are in the same tests that failed the nightlies, they might be due to GPU variability, as andrew suggested on slack.

Note: I'm not saying "don't even look into that". Just "if it's similar to current master, it might be out of the scope of the PR".

tcclevenger:
Yeah, values seem close. But I don't see a GPU machine that runs the CIME tests in nightlies. 

I'm generating a cime baseline with master for some of the lowres examples, and then comparing on my branch.

bartgol:
> Yeah, values seem close. But I don't see a GPU machine that runs the CIME tests in nightlies.

https://my.cdash.org/viewTest.php?buildid=2203497

This is a GPU v1 build.

tcclevenger:
Yesterday when trying to create baselines, completely new clone of scream, master branch no changes... ICE appeared in the old SC code..

I'm _fairly_ confident that the diffs I saw before that ICE were expected (seemed similar magnitude to nightly assent ones, but on different tests). I'd like to run again on the same test to be sure, but maybe we can just merge.

Going to try again this morning with a fresh scream clone. Maybe I was doing things dumb at the end of the day.

tcclevenger:
Ok, ran was able to test that one of the SCREAMv1 tests matched master. Tested another one and it and diffs, but around the same order of magnitude off as the ascent nightly test. So I am convinced I'm not doing anything off.

@ambrad @bartgol I'm good to merge if y'all are. Let me know if you want any more tests run.

ambrad:
Fine with me. What tests in particular diffed on Summit?

tcclevenger:
> https://my.cdash.org/viewTest.php?buildid=2203497

Diff: SMS_D_Ln9.ne4_ne4.F2010-SCREAMv1-noAero
Pass: ERP_D_Ln9.ne4_ne4.F2010-SCREAMv1

I'm checking an ne30 one now.

tcclevenger:
> Fine with me. What tests in particular diffed on Summit?

Diff: SMS_D_Ln9.ne4_ne4.F2010-SCREAMv1-noAero
Pass: ERP_D_Ln9.ne4_ne4.F2010-SCREAMv1

I'm checking an ne30 one now.

ambrad:
@tcclevenger you don't need to. Was just wondering what diffs you had seen. If this is BFB on CPU and you're seeing random diffs on Summit, I'm fine with merging this PR.

bartgol:
Agreed.

tcclevenger:
Great! Nice to have this finally merge..

PR #1395, opened 2022-02-02
Fix dimensions of dry/wet mixing variables in p3

singhbalwinder:
The dimensions of dry/wet conversion variables are set to (:ncol,:pver) to possibly fix issue #1393
.
BFB
Fixes #1393

PR #1788, opened 2022-07-14
Revert dry to wet mass mixing ratio conversion commits in P3 

crterai:
Until a new fix goes in, this PR reverts PR #1395 and #1288 to revert the dry to wet mass mixing ratio conversion in P3, since we found it to be a source of a water leak. 

NBFB

### DIFF
--- a/components/eam/src/physics/cam/micro_p3_interface.F90
+++ b/components/eam/src/physics/cam/micro_p3_interface.F90
@@ -1005,16 +1005,16 @@ end subroutine micro_p3_readnl
     !---------------------------------------------------------------------------------------
     qv_wet_in   = state%q(:,:,1) ! Get "wet" water vapor mixing ratio from state
     !Compute dry mixing ratios for all the constituents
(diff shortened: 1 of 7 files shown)


bartgol (approved):

singhbalwinder (approved):
Looks good to me!

bartgol:
FYI, when you see something like
```
Build type full_debug failed. Here's a list of failed tests:
Build type full_sp_debug failed. Here's a list of failed tests:
Build type debug_nopack_fpe failed. Here's a list of failed tests:
Build type release failed. Here's a list of failed tests:
```
which is puzzling cause one thinks "so no tests failed?", the reason is that ctest failed at build time. This is another thing that could maybe be made clearer in our testing scripts.

PeterCaldwell:
Now that @AaronDonahue is back, let's not integrate this PR for a day or so in hopes that he can fix the actual issue: https://github.com/E3SM-Project/scream/pull/1776.

AaronDonahue:
addressed by #1792

PR #1789, opened 2022-07-14
Fix homme restart of nstep

### LINKED ISSUE #1785: restart homme nstep appears to be physics nstep
In the following lines, it appears that Homme's `nstep` variable is set directly to the `:nstep` field from the restart file. The latter is the number of physics time steps. Thus, it appears that these lines do not account for the number of dynamics steps per physics time step.
https://github.com/E3SM-Project/scream/blob/6e4dd419ce107ae1d3801729c07419fab5b74b1a/components/scream/src/dynamics/homme/atmosphere_dynamics.cpp#L372-L375

bartgol:
The main goal of this PR is to fix the restart of Homme's internal step counter 'nstep'. This is different from the atm counter nstep, in case anyone of `nsplit`, `dt_remap_factor`, `dt_tracers_factor` is larger than 1.

In the process, I made a couple of small additions to the atm proc class and to the output manager class:
- atm proc: expose a list of "extra restart data". This is a map `data_name -> (data_type,data_value)`, where the data is supposed to be "simple", like a builtin type. The data value is stored in an `ekat::any` container, so it has pointer semantic. Note: names are supposed to be unique: if 2 atm proc have the same name for some restart data, the AD will crap out.
- output manager: allow to set a list of global attributes to save. This is the same map structure as above. The map is set at setup time (though it can in principle be set in the middle of a run), and the manager will internally loop over it whenever it hits a write step.

This feature(s) are used to restart Homme's nstep counter correctly. I manually ran `ERS_Ln9.ne4_ne4.F2010-SCREAMv1` on mappy, and verified in the log of the restarted test that nstep is 30 (correct) rather than 5 (current master incorrect value).

That said, I _think_ that the internal counter is only useful for Homme to determine if it's time to copy some data back to host, in view of diagnostics output (meaning the internal diagnostic output to screen performed by Homme). I don't think that the incorrect value of nstep had any real consequence on the correctness of the data.

Fix #1785

### DIFF
--- a/components/scream/src/control/atmosphere_driver.cpp
+++ b/components/scream/src/control/atmosphere_driver.cpp
@@ -133,7 +133,7 @@ init_scorpio(const int atm_id)
 
 void AtmosphereDriver::create_atm_processes()
 {
-  m_atm_logger->info("[EAMXX] create_atm_processes  ...");
+  m_atm_logger->info("[EAMxx] create_atm_processes  ...");
   start_timer("EAMxx::init");
   start_timer("EAMxx::create_atm_processes");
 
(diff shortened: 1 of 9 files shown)


ambrad:
Is this going in soon? This affects restart with SL because the nstep stuff matters more for the nontrivial time stepping done in SL. Thanks.

bartgol:
Let's retest. Should pass, now that we fixed master.

tcclevenger (approved):

PR #1790, opened 2022-07-15
Add trcmix bridge to scream

### LINKED ISSUE #1768: Trace gases should be initialized how EAM does
Currently we prescribe trace gas constituents by including values in the initial condition file, rather than initializing how EAM does. As far as I can tell, EAM allows prescribing surface values, and then extrapolating in the vertical and horizontal dimensions. The list of radiatively-active gas constituents in a typical F-case are:
```
h2o
o2
o3
ch4
co2
n2o
cfc11
cfc12
n2o
```
H2O comes from the water vapor tracer, so nothing needs to be done there. O2, CH4, CO2, N2O, CFC11, and CFC12 appear to be prescribed as surface values and then extrapolated. The code that does this appears to be `trcmix` in `components/eam/src/physics/cam/ghg_data.F90`, with surface values set as in `components/eam/src/physics/chem_surfvals.F90`.

jgfouca:
Fixes #1768


components/scream/src/physics/share/scream_trcmix.cpp
```diff
@@ -0,0 +1,23 @@
...
+void trcmix_cf2(const char* name, Int ncol, Int pcols, Int pver, Real* clat, Real* pmid, Real* q);
+
+}
+
+namespace scream {
+namespace physics {
+
+void trcmix(const char* name, Int ncol, Int pcols, Int pver, Real* clat, Real* pmid, Real* q)
+{
+  //d.transpose<ekat::TransposeDirection::c2f>();
```
jgfouca, line 17: I will implement the transposes once I know the API you guys want.

components/scream/src/physics/share/scream_trcmix.hpp
```diff
@@ -0,0 +1,18 @@
...
+#include "share/scream_types.hpp"
+
+//
+// Bridge function to call trcmix from CXX
+//
+
+namespace scream {
+namespace physics {
+
+void trcmix(const char* name, Int ncol, Int pcols, Int pver, Real* clat, Real* pmid, Real* q);
```
jgfouca, line 13: Here is the API from SCREAM's cxx point of view. Do you want the arrays to be raw pointers, as they are here, or Kokkos Views?

jgfouca (commented):
@brhillman , I've made progress... just need feedback for a few things to wrap this up.

ambrad:
For clarity: This PR is meant to handle the initialization associated with this part of scream_input.yaml:
```yaml
    rrtmgp:
      active_gases:
      - h2o
      - co2
      - o3
      - n2o
      - co
      - ch4
      - o2
      - n2
```
The h2o entry can be ignored; it is already handled by the initialization of qv, the first advected tracer. rrtmgp forms its list of active gases based on this YAML here: https://github.com/E3SM-Project/scream/blob/dcc877ad91baefc2aa4d63d2706b81bcf982444d/components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp#L30-L38
Then it builds fields for these here: https://github.com/E3SM-Project/scream/blob/dcc877ad91baefc2aa4d63d2706b81bcf982444d/components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp#L87-L93

components/scream/src/physics/share/scream_trcmix.F90
```diff
@@ -0,0 +1,285 @@
...
+
+  real(rtype), public, parameter :: rmwn2o = mwn2o/mwdry ! ratio of molecular weight n2o   to dry air
+  real(rtype), public, parameter :: rmwch4 = mwch4/mwdry ! ratio of molecular weight ch4   to dry air
+  real(rtype), public, parameter :: rmwf11 = mwf11/mwdry ! ratio of molecular weight cfc11 to dry air
+  real(rtype), public, parameter :: rmwf12 = mwf12/mwdry ! ratio of molecular weight cfc12 to dry air
+  real(rtype), public, parameter :: rmwco2 = mwco2/mwdry ! ratio of molecular weights of co2 to dry air
+
+
+  ! Default values for namelist variables -- now set by build-namelist
+  real(rtype) :: o2mmr = .23143_rtype               ! o2 mass mixing ratio
```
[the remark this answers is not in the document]
brhillman: I think this is how the concentrations for these are specified. I think we can easily add these in (probably in the radiation interface?) as input parameters and read from yaml.

components/scream/src/physics/share/scream_trcmix.F90
```diff
@@ -0,0 +1,285 @@
...
+  ! use physconst,      only: mwdry, mwch4, mwn2o, mwf11, mwf12, mwco2
+  ! use chem_surfvals,  only: chem_surfvals_get, chem_surfvals_co2_rad
+
+  use physics_utils, only: rtype
+
+  implicit none
+  save
+
+  ! Copied from physconst
+  real(rtype), public, parameter :: mwdry       = 28.966_rtype ! molecular weight dry air ~ kg/kmole
```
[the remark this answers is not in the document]
brhillman: That seems reasonable. I don't think anyone else will use these constants, but seems appropriate to have them in physics_constants.hpp.

components/scream/src/physics/share/scream_trcmix.hpp
brhillman: I'm not sure it matters. I'm imagining this routine being called from `RRTMGPRadiation::initialize_impl` to populate the gas concentrations stored in the Field Manager, so maybe it makes sense to have `trcmix` accept the kokkos views directly?

brhillman:
> > Right, so I think in the second block, we will change the `Required` to `Computed`, and then add a block to call the `trcmix` routine that Jim has ported in `RRTMGPRadiation::initialize_impl` to initialize the values of these gas concentrations at init time.
> 
> It should probably still have the 2d fields as Required, though, right?

Not sure I follow. I think the idea is that rad will initialize all of these trace gases (with the exception of h2o obviously), so doesn't that make them all "Computed"?

bartgol:
> > > Right, so I think in the second block, we will change the `Required` to `Computed`, and then add a block to call the `trcmix` routine that Jim has ported in `RRTMGPRadiation::initialize_impl` to initialize the values of these gas concentrations at init time.
> > 
> > 
> > It should probably still have the 2d fields as Required, though, right?
> 
> Not sure I follow. I think the idea is that rad will initialize all of these trace gases (with the exception of h2o obviously), so doesn't that make them all "Computed"?

Ah, sorry, I thought Rad would have to consume 2d fields as inputs (the surface values), and extrapolate 3d as outputs. If so, the AD needs to auto-load the 2d fields from IC file, then the 2d fields need to be listed as Required.

brhillman:
@bartgol actually it’s just scalar values set for surface emissions I believe, which are extrapolated in both horizontal and vertical. In EAM I think those scalar values are specified by namelist, so here we will be able to set them by input YAML and then Jim’s code here will use those to populate the 3D gas arrays.

jgfouca:
@brhillman , OK, I think this PR now has all the needed changes.

components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
```diff
@@ -88,7 +89,7 @@ void RRTMGPRadiation::set_grids(const std::shared_ptr<const GridsManager> grids_
     if (it == "h2o") { /* Special case where water vapor is called h2o in radiation */
       // do nothing, qv has already been added.
     } else {
-      add_field<Required>(it,scalar3d_layout_mid,kgkg,grid_name, ps);
+      add_field<Computed>(it,scalar3d_layout_mid,kgkg,grid_name, ps);
```
brhillman, line 92: I think we still need o2 and o3 to be "required", since those will still be read from file. Once we can remap online, I think they can also be computed and read by this process.

components/scream/src/physics/share/physics_constants.hpp
```diff
@@ -32,6 +32,7 @@ struct Constants
   static constexpr Scalar RhoIce        = 917.0;        // Ice density at 0 C from Wallace+Hobbes 1977
   static constexpr Scalar MWH2O         = 18.016;
   static constexpr Scalar MWdry         = 28.966;
+  static constexpr Scalar o2mmr         = 0.23143;      // o2 mass mixing ratio
```
brhillman, line 35: I think this should probably an input variable (read from scream_input.yaml) rather than a physics constant. I think this is the O2 mixing ratio that's used to populate the 3d distribution of O2 in the atmosphere. So it's not a fundamental constant.

components/scream/cime_config/namelist_defaults_scream.xml
```diff
@@ -189,6 +189,12 @@ tweaking their $case/namelist_scream.xml file.
     <!-- Radiation -->
     <rrtmgp inherit="physics_proc_base">
       <active_gases>h2o, co2, o3, n2o, co, ch4, o2, n2</active_gases>
+      <ch4vmr>-1.0</ch4vmr>
```
brhillman, line 192: It looks like this is setting the default values for these gases. I think we want to pull these from the F2010 use case from EAM and set here.

components/scream/src/physics/share/scream_trcmix.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+  {
+    std::vector<trcmix_view1d<const Real>> stuff = {clat};
+    ekat::device_to_host(
+      {hclat.data()},
+      pcols,
+      reinterpret_cast<std::vector<trcmix_view1d<const P>>&>(stuff));
+  }
+  {
+    std::vector<trcmix_view2d<const Real>> stuff = {pmid};
```
bartgol, line 51: It seems that all you are trying to do is get a host view of the field, with sizes (pcols,pver). At this point, why not requesting a host view as input to start with? The Field class offers that possibility:
```
auto pmid = p_mid.get_view<const Real**,Host>();
```
At that point, you could roll your own nested for loops to crop `pmid` to just the physical (pcols,pver) dimensions.

I think `ekat::device_to_host` is good if you have real (pun intended) non trivial packs. Also, I suspect our impl of that fcn _assumes_ that the output host view has sizes equal to the actual allocation (that is, no padding is used).

jgfouca:
@brhillman and I were going in circles trying to debug f90/cxx bridge issues. I got fed up and just ported the f90 code to cxx.

AaronDonahue:
@jgfouca or @brhillman , it looks like this PR needs a rebase or a merge from master to resolve conflicts.

jgfouca:
@brhillman , I know you are actively working on this branch. Is it OK if I do an upstream merge?

brhillman:
@jgfouca @AaronDonahue sorry just finished my changes, this code is doing what we expect it to now. Time for an upstream to resolve conflicts I guess.

brhillman:
Lots of conflicts, think I got them testing again now.

brhillman:
Alright, lots of changes have been made. The impl here now appears to be consistent with EAM/SCREAMv0, and passes unit tests for me on mappy. This WILL be non-BFB for CIME tests of course, because we computing greenhouse gases differently than we were before, and in fact the previous implementation had a few bugs, both in the input files and in how we were treating the values read in. This now gives comparable numbers for greenhouse gases at runtime with a companion SCREAMv0 run. @bartgol @jgfouca @AaronDonahue I'd appreciate a second look at this since I've made some big changes.

ambrad:
Have you tested this yet in a v1 pg2 simulation?

brhillman:
@ambrad thanks for the hint, working on a fix for pg2 now.

ambrad:
@brhillman I think you should just have to modify the list here: https://github.com/E3SM-Project/scream/blob/394305bd7f35740de2c4d6d4cef887be90cf35ec/components/scream/src/dynamics/homme/atmosphere_dynamics_fv_phys.cpp#L230-L236 to have just whateve 3D fields must still be read from the IC file. If none, then see the `// TODO [rrtmgp active gases]` comment for instructions on removing all that code.

brhillman:
Thanks @ambrad I had found that this morning, but I think I'm running into issues because the name we store in the FM has changed. So I tried to modify this as follows:
```
void fv_phys_rrtmgp_active_gases_init (const ekat::ParameterList& p) {
  const auto& v = p.sublist("atmosphere_processes").sublist("physics")
    .sublist("rrtmgp").get<std::vector<std::string>>("active_gases");
  for (const auto& e : v)
    if (e == "o3") //(e != "h2o")
      s_tgw.active_gases.push_back(e + "_volume_mix_ratio");
}
```
as `o3` is the only gas that needs to be init'ed from file now, but we keep it as `o3_volume_mix_ratio` in the FM and in the initial condition file. This, however, gives me the following:
```
Backtrace for this error:
terminate called after throwing an instance of 'std::logic_error'
  what():  /home/bhillma/codes/scream/branches/add_trcmix/components/scream/src/share/field/field_manager.cpp:154: FAIL:
ptr!=nullptr
Error! Field identifier 'o3_volume_mix_ratio[Physics PG2] <double:COL,LEV>(16,72) [moles/mole]' not found.
```
(note that I also corrected the units).

ambrad:
@brhillman I'll run your branch and see what is amiss.

ambrad:
@brhillman where can I find screami_ne4np4L72_20220818.nc? Or can I just switch back to the old IC file?

brhillman:
@ambrad it's on blues, it should download automatically if you try to run, but I just fixed permissions on the file on the server to make sure it has read permission for all. If you previously tried and failed, please try again. You cannot switch back to the original IC because we renamed `o3` to `o3_volume_mix_ratio`.

ambrad:
The permission change worked. Thanks!

ambrad:
Looks like the issue was just matching units. This commit works for me: https://github.com/ambrad/scream/commit/558e025035944d28baa3200cdface68a4a5dcc06

brhillman:
Ah, slick, I didn't realize our FM machinery was matching up units. Thanks for the fix, merging your commit in.

brhillman:
> I noticed the IC file has o3 listed as kg/kg still. Not sure if that's intended.

Already on it, thanks for pointing it out! Renaming the attributes in the IC files now.

components/scream/cime_config/namelist_defaults_scream.xml
```diff
@@ -205,8 +205,20 @@ tweaking their $case/namelist_scream.xml file.
...
-      <active_gases>h2o, co2, o3, n2o, co, ch4, o2, n2</active_gases>
       <column_chunk_size>1280</column_chunk_size>
+      <!-- Radiatively active gases; surface values set to F2010 settings taken from EAM  -->
+      <!-- Note that h2o concentrations are just taken from qv, o3 is prescribed for now, -->
+      <!-- o2 is hard-coded as a constant, CFCs are ignored                               -->
+      <active_gases>h2o, co2, o3, n2o, co, ch4, o2, n2</active_gases>
+      <ch4vmr      > 1807.851e-9  </ch4vmr>
+      <co2vmr      > 388.717e-6   </co2vmr>
+      <!-- If co2vmr_rad is set and greater than 0, it will override any values of co2vmr for rad calcs -->
+      <co2vmr_rad  > -1.0         </co2vmr_rad>
```
bartgol, line 216: So, if co2vm_rad>0, then rrtmgp uses this instead of co2vmr? Why not just changing the value of co2vmr?

components/scream/cime_config/namelist_defaults_scream.xml
```diff
@@ -251,13 +263,13 @@ tweaking their $case/namelist_scream.xml file.
...
-    <Filename hgrid="ne512np4">${DIN_LOC_ROOT}/atm/scream/init/screami_ne512np4L128_20220525.nc</Filename> 
-    <Filename hgrid="ne1024np4">${DIN_LOC_ROOT}/atm/scream/init/screami_ne1024np4L128_20220713.nc</Filename> 
-    <Filename hgrid="ne4np4" COMPSET=".*SCREAM%AQUA.*">${DIN_LOC_ROOT}/atm/scream/init/scream-aquaplanet_init_ne4np4_L72_20220524.nc</Filename>
-    <Filename hgrid="ne30np4" COMPSET=".*SCREAM%AQUA.*">${DIN_LOC_ROOT}/atm/scream/init/scream-aquaplanet_init_ne30np4_L128_20211202.nc</Filename> <!-- This will need to be updated to the new nccn name -->
+    <Filename hgrid="ne4np4">${DIN_LOC_ROOT}/atm/scream/init/screami_ne4np4L72_20220823.nc</Filename>
+    <Filename hgrid="ne30np4">${DIN_LOC_ROOT}/atm/scream/init/screami_ne30np4L72_20220823.nc</Filename>
+    <Filename hgrid="ne120np4">${DIN_LOC_ROOT}/atm/scream/init/screami_ne120np4L72_20220823.nc</Filename>
+    <Filename hgrid="ne512np4">${DIN_LOC_ROOT}/atm/scream/init/screami_ne512np4L128_20220823.nc</Filename> 
+    <Filename hgrid="ne1024np4">${DIN_LOC_ROOT}/atm/scream/init/screami_ne1024np4L128_era5_20131001_20220823.nc</Filename>
```
bartgol, line 270: I'm curious, why does this file have two timestamps?

components/scream/src/dynamics/homme/atmosphere_dynamics_fv_phys.cpp
```diff
@@ -230,9 +230,10 @@ static TraceGasesWorkaround s_tgw;
 void fv_phys_rrtmgp_active_gases_init (const ekat::ParameterList& p) {
   const auto& v = p.sublist("atmosphere_processes").sublist("physics")
     .sublist("rrtmgp").get<std::vector<std::string>>("active_gases");
-  for (const auto& e : v)
-    if (e != "h2o")
-      s_tgw.active_gases.push_back(e);
+  for (const auto& e : v) {
+    if (e == "o3")
```
bartgol, line 234: So h2o was not an active gas before (was it the same as qv?), but now it is? Just making sure this did not get lost in the fixes.

brhillman: This section is only for the gases that need to be read from file and are not handled elsewhere. h2o is just qv, for which the dyn<->phys mapping happens elsewhere, and the rest of the active gases are computed within the radiation interface, so now we only need this special workaround code for o3 now (this too will go away when we can do simple online remap and can use the o3 climatology file that EAM uses for prescribed ozone cases, or if/when we port the linear ozone scheme to EAMxx).

components/scream/cime_config/namelist_defaults_scream.xml
brhillman: The first one is the start date, the second is the creation date. I.e., this is an initial condition derived from reanalysis to simulate October 1, 2013. I'm open to a better convention here!

components/scream/cime_config/namelist_defaults_scream.xml
brhillman: That's a good question. Short answer is "that's what EAM does". I think this is to support overriding the value of co2vmr *just* for the rad calculations. I don't really see what the point of this is, so I think I'll just remove this...

components/scream/src/dynamics/homme/atmosphere_dynamics_fv_phys.cpp
bartgol: So you could basically just remove the for loop and do something like
```
if (ekat::contains(v,"o3")) {
  s_tgw.active_gases.push_back( .. );
}
```

components/scream/cime_config/namelist_defaults_scream.xml
brhillman: Removed.

components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
```diff
@@ -590,23 +609,65 @@ void RRTMGPRadiation::run_impl (const int dt) {
...
+        scream::physics::trcmix(
+          name, m_lat, d_pmid, d_vmr,
+          m_co2vmr_rad, m_co2vmr, m_n2ovmr, m_ch4vmr, m_f11vmr, m_f12vmr
+        );
+        // Back out volume mixing ratios
+        const auto air_mol_weight = PC::MWdry;
+        const auto policy = ekat::ExeSpaceUtils<ExeSpace>::get_default_team_policy(m_ncol, m_nlay);
+        Kokkos::parallel_for(policy, KOKKOS_LAMBDA(const MemberType& team) {
+          const int i = team.league_rank();
+          Kokkos::parallel_for(Kokkos::TeamThreadRange(team, m_nlay), [&] (const int& k) {
```
bartgol, line 652: `m_nlay` is an implicit use of `this`. You should use `nlay`.

components/scream/src/physics/share/scream_trcmix.cpp
```diff
@@ -0,0 +1,118 @@
+#include "scream_trcmix.hpp"
+#include "physics_constants.hpp"
+
+#include "ekat/util/ekat_math_utils.hpp"
+#include "ekat/kokkos/ekat_kokkos_utils.hpp"
+
+#include <tuple>
```
bartgol, line 7: this seems unused.

components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
brhillman: Ah yeah you're right. Will fix, thanks for finding.

brhillman: Fixed.

components/scream/src/dynamics/homme/atmosphere_dynamics_fv_phys.cpp
brhillman: Yes, replaced.

components/scream/src/physics/share/scream_trcmix.cpp
brhillman: Indeed, removed.

components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
```diff
@@ -728,7 +728,7 @@ void RRTMGPRadiation::run_impl (const int dt) {
     // Compute band-by-band surface_albedos. This is needed since
     // the AD passes broadband albedos, but rrtmgp require band-by-band.
     rrtmgp::compute_band_by_band_surface_albedos(
-      ncol, m_nswbands,
+      ncol, nswbands,
```
bartgol, line 731: This is fine, but FYI, the `m_lay` issue was b/c you used `this` inside a KOKKOS_LAMBDA. It's ok to use `m_xyz` here, since you're not _inside_ a `__device__ [=]` kind of lambda.

bartgol (dismissed):

components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
brhillman: True, this just ended up getting changed because I found a case of using `m_nswbands` inside a YAKL device lambda. Would it be better to use the `m_nswbands` version here? I think it probably doesn't matter I was just trying to be consistent in my use...

brhillman:
Some quick ugly plots to verify impl here. n2o comparison between a v0 run instrumented with additional outfld calls to output the gas volume mixing ratios and a v1 run with this branch, showing that indeed n2o is correctly set to a constant at the bottom level and extrapolated as done in v0:
image
o3 comparison, in this case the values are different because o3 is time-dependent in v0, but is prescribed via input file in v1 for the time being (I did verify separately that the o3 values in v1 are consistent with the inputs we specify):
image
co2 comparison, which shows that co2 is being set to a constant everywhere consistent with v0. Note that the value is slightly different because the v0 configuration appeared to still be using the F2000 value, while I pulled the slightly higher F2010 value for our v1 compset from a F2010 EAM configuration. Apparently, CO2 concentration should be set based on the  CCSM_CO2_PPMV xml variable, so I'm not sure why our F2010-SCREAM-LR compset is not picking up the same value. Anyways, the comparison:
image

bartgol (approved):

components/scream/cime_config/namelist_defaults_scream.xml
```diff
@@ -237,13 +247,13 @@ tweaking their $case/namelist_scream.xml file.
...
-    <Filename hgrid="ne30np4" COMPSET=".*SCREAM%AQUA.*">${DIN_LOC_ROOT}/atm/scream/init/scream-aquaplanet_init_ne30np4_L128_20211202.nc</Filename> <!-- This will need to be updated to the new nccn name -->
+    <Filename hgrid="ne4np4">${DIN_LOC_ROOT}/atm/scream/init/screami_ne4np4L72_20220823.nc</Filename>
+    <Filename hgrid="ne30np4">${DIN_LOC_ROOT}/atm/scream/init/screami_ne30np4L72_20220823.nc</Filename>
+    <Filename hgrid="ne120np4">${DIN_LOC_ROOT}/atm/scream/init/screami_ne120np4L72_20220823.nc</Filename>
+    <Filename hgrid="ne512np4">${DIN_LOC_ROOT}/atm/scream/init/screami_ne512np4L128_20220823.nc</Filename> 
+    <Filename hgrid="ne1024np4">${DIN_LOC_ROOT}/atm/scream/init/screami_ne1024np4L128_era5_20131001_20220823.nc</Filename> 
+    <Filename hgrid="ne4np4" COMPSET=".*SCREAM%AQUA.*">${DIN_LOC_ROOT}/atm/scream/init/screami_aquaplanet_ne4np4L72_20220823.nc</Filename>
+    <Filename hgrid="ne30np4" COMPSET=".*SCREAM%AQUA.*">${DIN_LOC_ROOT}/atm/scream/init/screami_aquaplanet_ne30np4L128_20220823.nc</Filename> <!-- This will need to be updated to the new nccn name -->
```
AaronDonahue, line 256: @brhillman , should we make similar changes for the `vertical_coordinate_filename` field?  Otherwise we'll still have to download the huge old ne1024 init condition file just to satisfy the vertical coordinate setting.

brhillman: Ah shoot, you're right. I'll fix. Side note, maybe it would be good to not have this redundancy in here? It's one more place to make a mistake like this.

brhillman: Fixed.

brhillman:
Looks the fails on mappy now are just the expected diffs in the CIME tests. This PR should be ready, pending results from @AaronDonahue test with ne1024.

AaronDonahue:
@bartgol , I think this PR is next on the docket for non-BFB merging.  Do you mind taking one more look before I manually merge?  I'm taking a look too.

AaronDonahue (approved):

bartgol (approved):
Not much changed since my last review. Let's get this party moving!

brhillman:
Blake testing passed, mappy testing just shows the expected diffs, I think weaver is expected to fail today is that correct @AaronDonahue @bartgol ? If true I think this is ready for manual merge and baseline blesses.

bartgol:
> Blake testing passed, mappy testing just shows the expected diffs, I think weaver is expected to fail today is that correct @AaronDonahue @bartgol ? If true I think this is ready for manual merge and baseline blesses.

Weaver is back online. I would wait for another round.

brhillman:
@AaronDonahue @bartgol weaver and blake both passed, mappy has expected diffs.

PR #1792, opened 2022-07-18
Add dry pressure coordinates to the homme/eamxx interface

AaronDonahue:
This commit adds dry pressure coordinates to the homme interface with EAMxx in the `update_pressure()` routine.

This will allow physics processes that need to use dry coordinates to have access to them.  In particular, P3 uses dry mass-mixing-ratios for the constituents and thus needs the dry pressure coordinates when calculating densities.  This commit applies the new dry pressure coordinates in the P3 interface.

There is also a minor bugfix to the P3 microphysics update of ice prognostics that was discovered over the course of this development.

Replaces #1776


AaronDonahue:
WIP until I update the initial condition files for the standalone tests to accomodate dry pressure coordinates.

AaronDonahue:
@bartgol and @ambrad, I think this is ready for review. The fail on mappy with DIFFs is expected now that we switched to dry pressure. 

@bartgol would you mind commenting specifically on the choice to setup a single cmake variable to define the initial condition file for our stand alone tests? I'm not sure if my approach is the correct one. It would be nice though to not have to change 20+ files every time we update the name of that file though.

ambrad (dismissed):

bartgol (approved):

AaronDonahue:
It looks like `restarted_vs_monolithic` is failing on mappy.  I'll need to fix that and push a new commit.  I'll probably need re-approval after that.

ambrad (dismissed):

components/eam/src/physics/cam/micro_p3.F90
```diff
@@ -3154,7 +3154,7 @@ subroutine update_prognostic_ice(qc2qi_hetero_freeze_tend,qccol,qc2qr_ice_shed_t
    if (do_predict_nc) then
       nc = nc + (-nc_collect_tend-nc2ni_immers_freeze_tend)*dt
    endif
-   qr = qr + (-qrcol+qi2qr_melt_tend-qr2qi_immers_freeze_tend+qc2qr_ice_shed_tend)*dt
+   qr = qr + (-qrcol-qr2qi_immers_freeze_tend+qc2qr_ice_shed_tend)*dt
```
AaronDonahue, line 3157: Pinging these changes so they show up in the comments.  I'm going to remove the commits that make this bugfix.  Is it causing `p3_run_and_cmp` to fail and is holding up the PR.  Since they are orthogonal to the main thrust of this fix they can be removed.  But we should submit an issue that points to the bug.

ambrad (approved):

bartgol (approved):

AaronDonahue:
Now that this is approved I am going to manually merge.

PR #1794, opened 2022-07-18
Fix regex for ELM use cases for SCREAM compsets

brhillman:
Fix bigrid configuration for ne1024. Two fixes needed: first, fix a regular expression for ELM use cases for SCREAM compsets. This previously would not match the SCREAMv1 compsets because the regex had EAM in the compset name. This would leave fsurdat unset for resolutions that do not have defaults (i.e., ne1024np4). Also cleans up the setting of use cases for the rest of the SCREAM compsets. Second, we neglected to set the vertical coordinate file for ne1024 in `namelist_defaults_scream.xml`. This is non-BFB because previously the SCREAMv1 compsets would not match any ELM use cases, leaving the ELM use case UNSET.

### DIFF
--- a/components/elm/cime_config/config_component.xml
+++ b/components/elm/cime_config/config_component.xml
@@ -83,12 +83,10 @@
       <value compset="SSP370(?:SOI)?_EAM%CMIP6_ELM">2015-2100_SSP370_transient</value>
       <value compset="2010(?:SOI)?_EAM%CMIP6-LR_ELM">2010_CMIP6LR_control</value>
       <value compset="2010(?:SOI)?_EAM%CMIP6-HR_ELM">2010_CMIP6HR_control</value>
-      <value compset="2010(?:SOI)?_EAM.*SCREAM-LR" >2010_defscream_control</value>
-      <value compset="2010(?:SOI)?_EAM.*SCREAM-HR" >2010_defscream_control</value>
-      <value compset="2010(?:SOI)?_EAM.*SCREAM-LR-DYAMOND2" >2010_dy2scream_control</value>
-      <value compset="2010(?:SOI)?_EAM.*SCREAM-HR-DYAMOND2" >2010_dy2scream_control</value>
-      <value compset="2000(?:SOI)?_EAM.*SCREAM-LR" >2000_control</value>
-      <value compset="2000(?:SOI)?_EAM.*SCREAM-HR" >2000_control</value>
+      <!-- Use cases for SCREAM compsets -->
+      <value compset="2010(?:SOI)?_.*SCREAM"                    >2010_defscream_control</value>
+      <value compset="2010(?:SOI)?_EAM.*SCREAM-[L,H]R-DYAMOND2" >2010_dy2scream_control</value>
+      <value compset="2000(?:SOI)?_EAM.*SCREAM"                 >2000_control</value>
       <!-- superfast chemistry compsets (use CMIP6 land configuration)-->
       <value compset="1850(?:SOI)?_EAM%AR5sf_ELM">1850_CMIP6_control</value>
       <value compset="20TR(?:SOI)?_EAM%AR5sf_ELM">20thC_CMIP6_transient</value>
--- a/components/elm/cime_config/config_component.xml
+++ b/components/elm/cime_config/config_component.xml
@@ -85,7 +85,7 @@
       <value compset="2010(?:SOI)?_EAM%CMIP6-HR_ELM">2010_CMIP6HR_control</value>
       <!-- Use cases for SCREAM compsets -->
       <value compset="2010(?:SOI)?_.*SCREAM"                    >2010_defscream_control</value>
-      <value compset="2010(?:SOI)?_EAM.*SCREAM-[L,H]R-DYAMOND2" >2010_dy2scream_control</value>
+      <value compset="2010(?:SOI)?_EAM.*SCREAM-[LH]R-DYAMOND2" >2010_dy2scream_control</value>
       <value compset="2000(?:SOI)?_EAM.*SCREAM"                 >2000_control</value>
       <!-- superfast chemistry compsets (use CMIP6 land configuration)-->
       <value compset="1850(?:SOI)?_EAM%AR5sf_ELM">1850_CMIP6_control</value>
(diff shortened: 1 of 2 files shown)


components/elm/cime_config/config_component.xml
```diff
@@ -83,12 +83,10 @@
...
       <value compset="2010(?:SOI)?_EAM%CMIP6-HR_ELM">2010_CMIP6HR_control</value>
-      <value compset="2010(?:SOI)?_EAM.*SCREAM-LR" >2010_defscream_control</value>
-      <value compset="2010(?:SOI)?_EAM.*SCREAM-HR" >2010_defscream_control</value>
-      <value compset="2010(?:SOI)?_EAM.*SCREAM-LR-DYAMOND2" >2010_dy2scream_control</value>
-      <value compset="2010(?:SOI)?_EAM.*SCREAM-HR-DYAMOND2" >2010_dy2scream_control</value>
-      <value compset="2000(?:SOI)?_EAM.*SCREAM-LR" >2000_control</value>
-      <value compset="2000(?:SOI)?_EAM.*SCREAM-HR" >2000_control</value>
+      <!-- Use cases for SCREAM compsets -->
+      <value compset="2010(?:SOI)?_.*SCREAM"                    >2010_defscream_control</value>
+      <value compset="2010(?:SOI)?_EAM.*SCREAM-[L,H]R-DYAMOND2" >2010_dy2scream_control</value>
```
crterai, line 88: I'm not familiar with the `[L,H]` syntax to point to both L or H options. In L23 above, it looks like there's no comma (`[LH]`). Can you verify this is what we want/tested?

brhillman: Ah good catch, I think what I wrote here matches "L", ",", or "H". I'll fix and retest.

brhillman: Okay, fixed.

bogensch (approved):

crterai (approved):
Looks good to me!

brhillman:
Note this is failing baseline compare because SCREAMv1 CIME cases are non-BFB due to these cases previously not picking up an ELM use case, but now they pick up the 2010_defscream_control use case like SCREAMv0 compsets do.

bartgol:
@brhillman should we manually merge then (and re-bless)?

AaronDonahue:
@brhillman, I added retest because the last time it was tested was 15 days ago.  I suspect we'll get the same behavior.  Can you confirm that once we see the expected DIFF's it will be ready to manually merge?  

Note manual merging folks, we have an issue, #1828 , that is tracking the order we merge non-BFB PR's.  I have this one set to go in after PG2 is merged.  Which will be hopefully today.

AaronDonahue:
Looks like this is ready to go with just the expected DIFFs.  I'm going to manually merge.

ambrad:
Does it make sense that the pg2 CIME cases passed? Are the pg2 cases not picking up something that is needed?

bartgol:
So, is it safe to bless baselines?

ambrad:
I'd say go ahead and do that, but we should definitely resolve the points I raised above. @brhillman, any chance my musings above are actually correct?

bartgol:
Ok, blessed.

brhillman:
Sorry, missed these messages. Yes, okay to bless, diffs were expected. Lost track of this PR on vacation.

brhillman:
To resolve @ambrad musings, I think pg2 passed because the defaults are consistent with what the use cases pointed to. The differences in the namelists for ne4np4 after vs before this PR were as follows:
``` 
<  finidat = ' '
---
>  finidat = '/global/cfs/cdirs/e3sm/inputdata/lnd/clm2/initdata_map/clmi.ICRUCLM45.ne4_oQU240.67a0f98.0001-01-04-00000.nc'
13c13
<  fsurdat = '/global/cfs/cdirs/e3sm/inputdata/lnd/clm2/surfdata_map/surfdata_ne4np4_simyr2010_c210908.nc'
---
>  fsurdat = '/global/cfs/cdirs/e3sm/inputdata/lnd/clm2/surfdata_map/surfdata_ne4np4_simyr2000_c160614.nc'
```
So after this PR, for ne4np4, with the `2010_defscream_control` use case being picked up, finidat is set to empty, fsurdat is set to the 2010 version (rather than the resolution default of the 2000 version). Note that finidat is set empty because the use case file sets `sim_year=2010`, rather than 2000, and there's no resolution default for `sim_year=2010`. That explains the diffs for ne4np4, which we expected.

The ne4pg2 namelist shows no diff before/after the PR. Looking at the use case file, we have:
```
<?xml version="1.0"?>

<namelist_defaults>

<use_case_desc>Conditions to simulate 2010 land-use for default scream</use_case_desc>

<sim_year>2010</sim_year>

<!-- Temporary until fsurdat files for ne4pg2 and ne30pg2 are produced for F2010 simulations  -->
<fsurdat hgrid="ne4np4.pg2" >lnd/clm2/surfdata_map/surfdata_ne4pg2_simyr2000_c190620.nc</fsurdat>
<fsurdat hgrid="ne30np4.pg2" >lnd/clm2/surfdata_map/surfdata_ne30pg2_simyr2000_c190408.nc</fsurdat>

<sim_year_range>constant</sim_year_range>

<stream_year_first_ndep phys="clm4_5" use_cn=".true." ndepsrc="stream" >2010</stream_year_first_ndep>
<stream_year_last_ndep  phys="clm4_5" use_cn=".true." ndepsrc="stream" >2010</stream_year_last_ndep>

<stream_year_first_pdep phys="clm4_5" use_cn=".true." ndepsrc="stream" >2010</stream_year_first_pdep>
<stream_year_last_pdep  phys="clm4_5" use_cn=".true." ndepsrc="stream" >2010</stream_year_last_pdep>

<stream_year_first_popdens phys="clm4_5" use_cn=".true." ndepsrc="stream" >2010</stream_year_first_popdens>
<stream_year_last_popdens  phys="clm4_5" use_cn=".true." ndepsrc="stream" >2010</stream_year_last_popdens>

</namelist_defaults>
```
So the use case explicitly sets the fsurdat file to the 2000 one here, but this is already the resolution default in `namelist_defaults.xml`:
```
<fsurdat hgrid="ne4np4.pg2" sim_year="2000" >
lnd/clm2/surfdata_map/surfdata_ne4pg2_simyr2000_c190620.nc</fsurdat>
<fsurdat hgrid="ne4np4.pg2" sim_year="2010" >
lnd/clm2/surfdata_map/surfdata_ne4pg2_simyr2000_c190620.nc</fsurdat>
```
`finidat` is empty for both ne4pg2 cases because there is no resolution default `finidat` for ne4pg2 for either `sim_year=2000` or `sim_year=2010`.

brhillman:
Looking at this, I actually think we want to remove those temporary overrides for the `fsurdat` for ne4pg2 and ne30pg2 in the use case file. These are resolution defaults anyways specified in `namelist_defaults.xml`, no need to additionally specify in the use case. That will be a separate PR though.

PR #1795, opened 2022-07-19
Add ifdef guard around gptl init to reduce warning messages

### LINKED ISSUE #1791: warning GPTLinitialize: has already been called is written too often (once per MPI)
`GPTLinitialize: has already been called`
It doesn't take a very large problem for this to become really annoying to see in e3sm.log.
A job with 85 nodes using 128 MPI's per node means we get this 10880 times.

In `share/timing/gptl.c`

```
  if (initialized)
    return GPTLerror ("%s: has already been called\n", thisfunc);
```

For some cases I can change this to `return 0;` to avoid the messages, but this causes trouble for other cases when it tries to do timing summary work at end of run and crashes.

ndkeen:
Add `#ifdef SCREAM_CONFIG_IS_CMAKE` around gptl init.

Fixes #1791

### DIFF
--- a/components/scream/src/share/util/scream_timing.cpp
+++ b/components/scream/src/share/util/scream_timing.cpp
@@ -5,8 +5,12 @@
 namespace scream {
 
 void init_gptl (bool& was_already_inited) {
+#ifdef SCREAM_CONFIG_IS_CMAKE
+  was_already_inited = true;
+#else
   auto ierr = GPTLinitialize();
   was_already_inited = (ierr!=0);
+#endif
 }
 void finalize_gptl () {
   GPTLfinalize();
--- a/components/scream/src/share/util/scream_timing.cpp
+++ b/components/scream/src/share/util/scream_timing.cpp
@@ -1,11 +1,12 @@
 #include "share/util/scream_timing.hpp"
+#include "scream_config.h"
 
 #include <gptl.h>
 
 namespace scream {
 
 void init_gptl (bool& was_already_inited) {
-#ifdef SCREAM_CONFIG_IS_CMAKE
+#ifdef SCREAM_CIME_BUILD
   was_already_inited = true;
 #else
   auto ierr = GPTLinitialize();


components/scream/src/share/util/scream_timing.cpp
```diff
@@ -5,8 +5,12 @@
 namespace scream {
 
 void init_gptl (bool& was_already_inited) {
+#ifdef SCREAM_CONFIG_IS_CMAKE
```
bartgol, line 8: This is incorrect for standalone SCREAM builds. `SCREAM_CONFIG_IS_CMAKE` is always defined, regardless of CIME-vs-standalone configuration, so the `else` branch would never be executed (while we do need it to be executed for standalone runs).

The correct thing to do is to check for `SCREAM_CIME_BUILD`. Notice that you also need to add an include for `scream_config.h`, so that the macro (if defined) will be available.

bartgol (changes requested):

components/scream/src/share/util/scream_timing.cpp
ndkeen: Would it make sense to ensure all files have `scream_config.h`?

bartgol: I would prefer not. I don't think it's a good habit to include what you don't need. If someone adds a line that needs config.h macros, they should explicitly include the file (without relying on indirect inclusion).

ndkeen: But then we could easily arrive at same scenario here where we think a macro is being used in the file and it's not.

bartgol: Imho, if someone adds `#ifdef BLAH`, then they are responsible to ensure that the macro `BLAH` is defined when they expect it to be defined.

In general, I'm not in favor of Include-What-You-Use. That said, I understand that this might come down to a matter of taste. If the general consensus is to include config.h everywhere "just in case", I can live with that.

Pinging some other AD devs for their thoughts: @jgfouca @jeff-cohere @tcclevenger @AaronDonahue .

ndkeen: In general, sure I agree to only include what you need. But this is a special case -- far from "just in case" -- this is something fundamental if we are building for stand-alone testing or with CIME.

In fact, are we not at a point where the default logic should be *with* CIME?  And make changes to do something different when building stand-alone?  `#ifdef SCREAM_STANDALONE_TESTING` ?

jeff-cohere: I think questions like this can only really be answered in terms of costs: what does including `scream_config.h` everywhere cost us versus time spent fumbling with macros that may or may not be defined? Maybe there's a principle of least astonishment in here, too.

bartgol: I think it's fine to have `SCREAM_STANDALONE` in the config.h, and have code check for that (so that the default behavior in case config.h is not included is to think in CIME terms). Basically trade `SCREAM_CIME_BUILD` everywhere (including CMake code) for `SCREAM_STANDALONE` (with opposite behavior, of course).

As for the costs, in our case, there's probably little drawback in including config.h everywhere. I personally see two, but they are arguably personal and tiny: 1) I tend to _assume_ that files included (like symbols delcared, inputs to fcns...) are needed/used somewhere, and 2) certain files might not need to depend on config.h, so they need not to be rebuilt during continuous development when config.h is regenerated (e.g., b/c of branch/config changes).

ndkeen: I committed a change to branch

bartgol (approved):

PR #1799, opened 2022-07-20
Upgrade PD remapper to handle subfields

bartgol:
This PR upgrades the PD remapper to handle subfields just like any other field. In doing so, it simplifies a bit the implementation (e.g., no switches based on whether a dyn field is a state; or just one BoundaryExchange object needed). Also, it removes the need of having `Homme::TimeLevel` already set up. This might be important to be able to use the PD remapper during initialization _before_ `HommeDynamics::initialize` is called (which is where homme initialization was completed).

For instance, I think we can now use a PD remapper while loading IC fields. I also would like to use a PD remapper inside another branch, where I'm changing the handling of topography (phis), from a field-manager-managed variable, to a field stored inside the grid.

### DIFF
--- a/components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
+++ b/components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
@@ -354,14 +354,14 @@ void HommeDynamics::initialize_impl (const RunType run_type)
   create_helper_field("FT_dyn",{EL,    GP,GP,LEV},{nelem,      NGP,NGP,nlevs},dgn);
   create_helper_field("FM_dyn",{EL,CMP,GP,GP,LEV},{nelem,    3,NGP,NGP,nlevs},dgn);
   create_helper_field("Q_dyn" ,{EL,CMP,GP,GP,LEV},{nelem,qsize,NGP,NGP,nlevs},dgn);
+
   // Tendencies for temperature and momentum computed on physics grid
   create_helper_field("FT_phys",{COL,LEV},    {ncols,  nlevs},pgn);
-  create_helper_field("FM_phys",{COL,CMP,LEV},{ncols,3,nlevs},pgn);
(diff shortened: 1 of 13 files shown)


tcclevenger (approved):
Much nicer!

ambrad:
Is this going in soon? I ask b/c of some anticipated conflicts with the pg2 work.

bartgol:
> Is this going in soon? I ask b/c of some anticipated conflicts with the pg2 work.

It can wait.

bartgol:
@tcclevenger this is probably ready for review. The fail I was getting was sneaky, and forced me to put a limitation on the remapping of subfields, which I wanted to discuss with you.

The limitation concerns "dynamic subfields". These are subfields of a field F, where the slicing index can be changed at runtime. We use these to subview homme's states (which have timelevels). By changing the slice they are subviewing at runtime, we don't need to create separate fields for each timelevel slice (the reason this works is b/c subfields internally store the parent's field view, and only subview it at the proper index when `get_view` is called).

Now, the issue is the BoundaryExchange structure in Homme: it needs to be setup by passing an actual view. This means that once we setup the bex object, we cannot move the slice where the subfield is subviewing the parent field. Since the bex is only needed for P->D remapping, dynamic subfields can still be used for D->P remapping (which, luckily, is the case we have).

I think the logic is not too involved (the code is prob more complicated than it should): if you have subfields, if the subview_info changes after setup, then P->D remap will crap out. Everything else works:
- no subfields: works (like before)
- subfields with static subview info: works
- subfields with dynamic subview info, but the subview info is never changed: works (this is, e.g., the case for the IC remapper)
- subfields with dynamic subview info, but only in the physics fields: works (I don't think this is currently used, dynamic subfields are used on the dyn grid)
- subfields with dynamic subview info on the dyn grid, but only D->P remap: works.

tcclevenger (dismissed):
This definitely looks good, and I don't think it's become too complicated.

tcclevenger (approved):

PR #1801, opened 2022-07-20
Fix to water path diagnostics and tests

bartgol:
One mathematical bug and a bunch of GPU-related bugs.

### DIFF
--- a/components/scream/src/diagnostics/ice_water_path.cpp
+++ b/components/scream/src/diagnostics/ice_water_path.cpp
@@ -57,11 +57,12 @@ void IceWaterPathDiagnostic::run_impl(const int /* dt */)
   const auto& qi_mid             = get_field_in("qi").get_view<const Pack**>();
   const auto& pseudo_density_mid = get_field_in("pseudo_density").get_view<const Pack**>();
 
+  const auto num_levs = m_num_levs;
   Kokkos::parallel_for("IceWaterPathDiagnostic",
                        default_policy,
                        KOKKOS_LAMBDA(const MemberType& team) {
     const int icol = team.league_rank();
-    Kokkos::parallel_reduce(Kokkos::TeamThreadRange(team, m_num_levs), [&] (const Int& idx, Real& lsum) {
+    Kokkos::parallel_reduce(Kokkos::TeamThreadRange(team, num_levs), [&] (const Int& idx, Real& lsum) {
       const int jpack = idx / Pack::n;
       const int klev  = idx % Pack::n;
       lsum += qi_mid(icol,jpack)[klev] * pseudo_density_mid(icol,jpack)[klev]/gravit;
--- a/components/scream/src/diagnostics/liquid_water_path.cpp
+++ b/components/scream/src/diagnostics/liquid_water_path.cpp
@@ -57,11 +57,12 @@ void LiqWaterPathDiagnostic::run_impl(const int /* dt */)
   const auto& qc_mid             = get_field_in("qc").get_view<const Pack**>();
   const auto& pseudo_density_mid = get_field_in("pseudo_density").get_view<const Pack**>();
 
+  const auto num_levs = m_num_levs;
   Kokkos::parallel_for("LiqWaterPathDiagnostic",
                        default_policy,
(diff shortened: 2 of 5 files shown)


AaronDonahue (approved):
This looks good to me, thank you @bartgol for fixing all of my GPU errors.  It looks like the AT is reporting this is failing mappy on the new test.  Since it basses on blake perhaps the AT found a random seed that makes the new test fail.  Probably worth checking out.

bartgol:
> This looks good to me, thank you @bartgol for fixing all of my GPU errors. It looks like the AT is reporting this is failing mappy on the new test. Since it basses on blake perhaps the AT found a random seed that makes the new test fail. Probably worth checking out.

It fails on both blake and mappy, and fails consistently, so it's not a fluke.

Besides, the water path classes should work with whatever input, since they do not do special ops that cause crash with certain values (e.g., square roots).

bartgol:
Btw, is there a compelling reason to keep 4 separate classes for the water paths? They seem to all do the exact same operation, just on a different field...

AaronDonahue:
> > This looks good to me, thank you @bartgol for fixing all of my GPU errors. It looks like the AT is reporting this is failing mappy on the new test. Since it basses on blake perhaps the AT found a random seed that makes the new test fail. Probably worth checking out.
> 
> It fails on both blake and mappy, and fails consistently, so it's not a fluke.
> 
> Besides, the water path classes should work with whatever input, since they do not do special ops that cause crash with certain values (e.g., square roots).

Ahhh I see that in the blake dropdown it shows the fails, the AT summary flags blake as passing.

AaronDonahue:
> Btw, is there a compelling reason to keep 4 separate classes for the water paths? They seem to all do the exact same operation, just on a different field...

No, it would be cleaner to have one single "water_path" class with liquid, ice, vapor, rain and eventually rimed ice as a subclass.  I wasn't sure how to do this in a clean way, but if you are up to the challenge I am all for it.

bartgol:
> Ahhh I see that in the blake dropdown it shows the fails, the AT summary flags blake as passing.

Well, no. The lat bit you see
```
./scream/components/scream/scripts/jenkins/jenkins_cleanup.sh
[SCREAM_PullRequest_Autotester_Blake] $ /bin/bash -le /tmp/jenkins7794054993994217919.sh
POST BUILD TASK : SUCCESS
END OF POST BUILD TASK : 0
Sending e-mails to: [lbertag@sandia.gov](mailto:lbertag@sandia.gov)
Finished: SUCCESS
```
is reporting the success of the "post-build action", which is what runs _after_ the job has completed. We use it to clean up old builds, to keep the filesystem usage under control.

The actual status of the testing is a bit above:
```
OVERALL STATUS: FAIL
```
That's a string that _we_ print from our scripts. That's what you need to look at.

AaronDonahue:
aahhhh okay.  I scroll with my eyes too quickly.

bartgol:
> > Btw, is there a compelling reason to keep 4 separate classes for the water paths? They seem to all do the exact same operation, just on a different field...
> 
> No, it would be cleaner to have one single "water_path" class with liquid, ice, vapor, rain and eventually rimed ice as a subclass. I wasn't sure how to do this in a clean way, but if you are up to the challenge I am all for it. The only difference between all five is which tracer to grab as input.

I don't think we should use derived classes. I think we should just have a "WaterPath" class, store the name of the tracer, and use that name during run_impl, to compute the water path of that tracer.

AaronDonahue:
Screen Shot 2022-07-20 at 11 08 50 AM

bartgol:
> I don't see anything in this read out to indicate failure though. Maybe I'm missing something. The `mappy` entry clearly has FAILED, where here I see PASSED for blake.

Ah, I see. This is part of the bug that is addressed by #1800 . Tbh, I never look at the summary. I always click on the individual builds, so I can also see if the fail (if any) came from v1 or standalone.

AaronDonahue:
I usually start with that drop down so I know which machines to look at in more detail.  Saves me O(seconds), which add up over time... I think. :-)

AaronDonahue (approved):

PR #1804, opened 2022-07-20
Fixes the sc_ut test which was failing on GPU's

AaronDonahue:
This commit adds a sync to host for the precipitation field copies
that are used in the sc_ut tests.  Without the sync to host this test
was failing on GPU.  Note that the failures are from changes made previously
in how precipitation is being handled in EAMxx.


bartgol (approved):

bartgol:
We have two sets of bugs in master. This PR fixes one, but not the other, so it would never pass the AT. It is however clear from the logs that the SP build passes, so the SC bug is fixed. I'm merging manually.

PR #1811, opened 2022-07-21
Make the handling of the diagnostic timestamp more robust.

### LINKED ISSUE #1810: Output field not initialized error when adding LiqWaterPath output and related fields
The ability to add vertically integrated cloud liquid, vapor, ice, and rain water paths was added with PR#1761. With master `35de2d4`, when `LiqWaterPath` is requested as an output, the following error results in a crash. Although note tested, I suspect it also affects `VapWaterPath`, `IceWaterPath`, and `RainWaterPath`.
```
5376: terminate called after throwing an instance of 'std::logic_error'
5376:   what():  /global/u1/t/terai/SCREAM/scream/master/components/scream/src/share/io/scorpio_output.cpp:193: FAIL:
5376: field.get_header().get_tracking().get_time_stamp().is_valid()
5376: Error! Output field 'LiqWaterPath' has not been initialized yet
5376: .
```

bartgol:
Before, we were setting the diag timestamp in `scorpio_output.cpp`, as soon as we created it, by grabbing the timestamp of one of the diagnostic required fields. However, we never checked that the timestamp was valid. In certain cases, additionally, all the required fields might have an invalid timestamp. This can happen if all the required fields are _computed_ in the ATM, so that none of them has a valid timestamp when the IO class is created at init time.

Fix #1810

### DIFF
--- a/components/scream/src/share/atm_process/atmosphere_diagnostic.hpp
+++ b/components/scream/src/share/atm_process/atmosphere_diagnostic.hpp
@@ -43,13 +43,15 @@ public:
   // Getting the diagnostic output
   Field get_diagnostic () const;
 
-  // Avoid shadowing the base class method, but allow calling without passing a dummy dt
-  using AtmosphereProcess::run;
-  void run () { this->run_impl(0); } // Pass null dt, but should not be used anyways.
-
   void set_computed_field (const Field& f) final;
   void set_computed_group (const FieldGroup& group) final;
+
+  void compute_diagnostic ();
 protected:
+
+  virtual void compute_diagnostic_impl () = 0;
+
+  void run_impl (const int /* dt */);
 
   // Diagnostics are meant to return a field
   Field m_diagnostic_output;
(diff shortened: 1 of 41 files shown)


components/scream/src/share/atm_process/atmosphere_diagnostic.hpp
```diff
@@ -43,14 +43,16 @@ class AtmosphereDiagnostic : public AtmosphereProcess
...
-
   void set_computed_field (const Field& f) final;
   void set_computed_group (const FieldGroup& group) final;
+
+  void compute_diagnostic ();
 protected:
 
+  virtual void compute_diagnostic_impl () = 0;
+
+  void run_impl (const int /* dt */);
```
AaronDonahue, line 54: is it safe to say that we don't actually ever use `run_impl` now?

AaronDonahue (dismissed):

components/scream/src/share/atm_process/atmosphere_diagnostic.hpp
bartgol: Well, probably. But until we refactor the diag class, making it independent of AtmosphereProcess, we don't want to risk someone calling the run methods. Since it's easy to fwd a call to run to a call to compute_diagnostic, we might as well do it.

Long term, I think Diagnostic should not be an atm process.

AaronDonahue: I agree

AaronDonahue (approved):

PR #1814, opened 2022-07-22
test-all-scream significant refactor

jgfouca:
Refactored items:
1) Tests become a proper object that can store test state
2) Valgrind and cuda-mem-check become test modifiers instead of test types
3) Use python f-strings instead of .format
4) 'opt' test no longer is set to do baseline checking. This made no sense
   since release mode is not BFB

### DIFF
--- a/components/scream/scripts/jenkins/jenkins_common_impl.sh
+++ b/components/scream/scripts/jenkins/jenkins_common_impl.sh
@@ -88,7 +88,7 @@ if [ $skip_testing -eq 0 ]; then
 
     # Add a valgrind and coverage tests for mappy for nightlies
     if [[ $is_at_run == 0 && "$SCREAM_MACHINE" == "mappy" ]]; then
-      ./scripts/gather-all-data "./scripts/test-all-scream -t dbg --valgrind ${TAS_ARGS}" -l -m $SCREAM_MACHINE
+      ./scripts/gather-all-data "./scripts/test-all-scream -t dbg --mem-check ${TAS_ARGS}" -l -m $SCREAM_MACHINE
       if [[ $? != 0 ]]; then fails=$fails+1; fi
 
       ./scripts/gather-all-data "./scripts/test-all-scream -t cov ${TAS_ARGS}" -l -m $SCREAM_MACHINE
@@ -97,7 +97,7 @@ if [ $skip_testing -eq 0 ]; then
 
     # Add a cuda-memcheck test for weaver for nightlies
     if [[ $is_at_run == 0 && "$SCREAM_MACHINE" == "weaver" ]]; then
-      ./scripts/gather-all-data "./scripts/test-all-scream -t dbg --cuda-mem-check ${TAS_ARGS}" -l -m $SCREAM_MACHINE
+      ./scripts/gather-all-data "./scripts/test-all-scream -t dbg --mem-check ${TAS_ARGS}" -l -m $SCREAM_MACHINE
       if [[ $? != 0 ]]; then fails=$fails+1; fi
     fi
   else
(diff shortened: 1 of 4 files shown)


jgfouca:
@AaronDonahue adding you since we discussed this briefly on slack. This PR fixes the issue of undocumented test types in the `--help`. I also decided that valgrind and cuda-memcheck should be modifiers instead of test types. I think this is better because it would allow the user the flexibility to do things like a valgrind single-precision test (with valgrind as a test type, this was not possible).

ambrad:
@jgfouca would you expand on this statement?
> 'opt' test no longer is set to do baseline checking. This made no sense since release mode is not BFB

My understanding is that baseline checking means that a run of a test is checked against reference output. The reference output is updated manually when a non-BFB PR is merged. According to this understanding, every build type should do baseline checking.

bartgol (changes requested):
Looks good. I have a suggestion about memcheck/valgrind.

As for baselines, we should be able to do _some_ baselines checking also in opt mode. The only potentially non deterministc ops are sum-reductions (max/min reductions should be ok). I see P3 has no such reduction, so even in OPT should generate the same result across runs, no?

ambrad:
@bartgol SCREAM as a whole, and thus all subcomponents, is run-to-run deterministic in all builds at this point, and we want to maintain that property.

jgfouca:
@ambrad , @bartgol , my comment regarding opt testing and baselines was based on this:

`src/share/scream_types.hpp`:
```
// We cannot expect BFB results between f90 and cxx if optimizations are on.                                                                                                                                                                   
// Same goes for cuda-memcheck because it makes the bfb math layer prohibitively                                                                                                                                                               
// expensive and so must be turned off.                                                                                                                                                                                                        
#if defined (NDEBUG) || defined (EKAT_ENABLE_CUDA_MEMCHECK)
static constexpr bool SCREAM_BFB_TESTING = false;
#else
static constexpr bool SCREAM_BFB_TESTING = true;
#endif
```

`src/physics/p3/tests/p3_run_and_cmp.cpp`:
```
scream::Real tol = SCREAM_BFB_TESTING ? 0 : std::numeric_limits<Real>::infinity();
```

`src/physics/shoc/tests/shoc_run_and_cmp.cpp`
```
scream::Real tol = SCREAM_BFB_TESTING ? 0 : std::numeric_limits<Real>::infinity();
```

The comment in scream_types.hpp for SCREAM_BFB_TESTING is incomplete. It is true that SCREAM_BFB_TESTING controls the f90/cxx bfb unit testing, but it also sets the default tolerance for the shoc/p3 run_and_cmp programs to infinity, which essentially disables the baseline checks. Maybe we shouldn't be doing that anymore?

In any event, I will turn baselines back on in opt mode so we will preserve the original (before the PR) behavior.

ambrad:
I see. To be clear, then, when you refer to baselines you mean only the standalone p3/shoc_run_and_cmp baseline checking, which also effectively is F90 vs C++ checking? Whatever you do is fine, then. I just wanted to check that this change doesn't affect CIME baseline testing in opt builds.

jgfouca:
@ambrad , correct, no impacts to anything CIME related. This is for internal testing and SCREAM_BFB_TESTING turns off the bfb unit test checking (comparing f90 and cxx) and the run_and_cmp_(f90|cxx) baseline checking which is NOT cross-language (run_and_cmp_cxx checks cxx runs against cxx baselines).

@bartgol , I looked over some old commits and the rationale for having baseline checking on for Release was to ensure that things (run_and_cmp, bfb unit tests) at least ran without exception/segfault even if exact results were not checked. By that logic, it seems like the FPE test should also do baseline checking too? It was set to off for some reason.

bartgol:
I think we should change the name then. `baseline` makes me think of run-to-run comparison of the _same_ implementation (be that f90 or cxx). The tests that have no hope to pass in opt are f90 vs cxx. Perhaps we should find a better name for those. I would vote for something very explicit, like `f90_vs_cxx`. Or maybe `cross_impl_testing` (more general, can be used for cxx-GPU vs cxx-CPU).

jgfouca:
@bartgol , I think I've addressed all your feedback for this PR. I did not do any further investigation into the deeper issue of whether run_and_cmp tolerances should be infinite in Release mode.

bartgol (approved):
We can address the issue of `run-to-run` vs `cross-impl` bfb-ness in a second PR.

PR #1820, opened 2022-07-26
Fix FieldGroup cloning method

bartgol:
The `get_const` method was not copying the individual group subfields if the "bundled" field was present. This caused the IC reader to "not see" some inputs that were needed from the IC file.


ambrad:
I suggest adding a unit test that shows the original failure and then certifies the fix.

tcclevenger (approved):

bartgol:
I believe this bugfix is exposing another bug in P3, related to liq_ice, vap_liq, and vap_ice exchange variables. In particular, the `model_restart` test fails, with the vars in the restarted run being 0 at some location (while the vars are !=0 at those locations for the monolithic run).

I suspect master would also fail if we ran more than 1+1 steps (which we probably should) for the test.

ambrad:
> I suspect master would also fail if we ran more than 1+1 steps (which we probably should) for the test.

I see that nontrivial v1 ERS/P tests pass with master. Does that contradict this statement, or am I misunderstanding the issue?

AaronDonahue:
@ambrad , the three variables in question are just diagnostics and are not part of the default output for CIME based tests.  So the non-BFB behavior isn't getting picked up in these tests.

components/scream/src/physics/p3/atmosphere_microphysics_run.cpp
```diff
@@ -24,6 +24,10 @@ void P3Microphysics::run_impl (const int dt)
   workspace_mgr.reset_internals();
 
   // Run p3 main
+  get_field_out("micro_liq_ice_exchange").deep_copy(0.0);
```
AaronDonahue, line 27: Luca, does this fix the issue we were seeing with restarts and these fields?

bartgol: Yes. At least it fixed it on this branch for the `model_restart` test.

bartgol:
@AaronDonahue This PR only fails v1 tests, with DIFF in baseline phase. This is expected, since we are now reading all the tracers (rather than init-ing them to 0).

We can merge at any point after pg2.

bartgol:
@AaronDonahue I believe this PR is next in line among the non-bfb ones.

AaronDonahue:
@bartgol , the FAILS on `mappy-CIME` runs look legit.  Can you check those out before we manually merge.

bartgol:
Ok, should be fixed now. The automatic merge of the defaults generated a duplicated entry in the XML file.

AaronDonahue (approved):

bartgol:
The diffs are expected (pg2 does not diff, bc the IC are read differently). I'm gonna merge and bless.

PR #1821, opened 2022-07-26
Jgfouca/upstream merge 2022 07 26

jgfouca:
Conflict summary:

### File: components/eam/bld/build-namelist

```
<<<<<<< HEAD
    if ($chem eq 'none' or $chem eq 'spa' or $chem eq 'waccm_ghg' or $chem eq 'super_fast_llnl' or $chem eq 'waccm_mozart' or $chem eq 'waccm_mozart_sulfur') {
        # If no chemistry then there must be prescribed aerosols unless physics
        # package is adiabatic or ideal.
        if ($phys eq 'default') {
            $prescribed_aero_model = 'modal';
        }
        elsif ($phys eq 'cam4' or $phys eq 'cam3') {
            $prescribed_aero_model = 'bulk';
        }
=======
    if ($chem eq 'none' or $chem eq 'waccm_ghg' or $chem eq 'super_fast_llnl' or $chem eq 'waccm_mozart' or $chem eq 'waccm_mozart_sulfur') {
        # If no chemistry then there must be prescribed aerosols unless physics
        # package is adiabatic or ideal.
        if ($phys eq 'default') {
            $prescribed_aero_model = 'modal';
        }
>>>>>>> upstream/master
```

Solution: add `$chem eq 'spa'` to upstream if statement

### File: components/eam/src/physics/cam/check_energy.F90

```
<<<<<<< HEAD
    elseif (icldliq > 1  .and.  icldice > 1 .and. irain > 1 .and. microp_scheme == 'P3') then
       ! In the case where the micro-physics scheme is P3 there is no snow
       ! constituent and isnow = -1.  So we still calculate the rest of the
       ! consituents and ensure that ws = 0.0.  NOTE! This change will likely
       ! lead to conflicts with upstream E3SM any time check_energy is
       ! changed...  The most important thing is to avoid any SNOW calculations
       ! when P3 is the microphysics scheme.
       do i = 1, ncol
          call energy_helper_eam_def_column(u(i,:),v(i,:),T(i,:),q(i,1:pver,1:pcnst),&
                                   ps(i),pdel(i,:),phis(i), &
                                   ke(i),se(i),wv(i),wl(i),wi(i),wr(i),ws(i),te(i),tw(i) )
          ws(i) = 0.0_r8
       enddo
    else
       call endrun('energy_helper...column is not implemented if water forms do not exist')
    endif
=======
    !else
    !   call endrun('energy_helper...column is not implemented if water forms do not exist')
    !endif
>>>>>>> upstream/master
```

Solution:

```
    !if (icldliq > 1  .and.  icldice > 1 .and. irain > 1 .and. isnow > 1) then
       do i = 1, ncol
          call energy_helper_eam_def_column(u(i,:),v(i,:),T(i,:),q(i,1:pver,1:pcnst),&
                                   ps(i),pdel(i,:),phis(i), &
                                   ke(i),se(i),wv(i),wl(i),wi(i),wr(i),ws(i),te(i),tw(i) )
          ! In the case where the micro-physics scheme is P3 there is no snow
          ! constituent and isnow = -1.  So we still calculate the rest of the
          ! consituents and ensure that ws = 0.0.  NOTE! This change will likely
          ! lead to conflicts with upstream E3SM any time check_energy is
          ! changed...  The most important thing is to avoid any SNOW calculations
          ! when P3 is the microphysics scheme.
          if (microp_scheme == 'P3') then
             ws(i) = 0.0_r8
          endif
       enddo
    !else
    !   call endrun('energy_helper...column is not implemented if water forms do not exist')
    !endif
```

In other words, take upstream except add the P3 check after the energy_helper call

```
<<<<<<< HEAD
    do k = 1, pver
       wr = wr + q(k,irain)*pdel(k)/gravit
       ! Recall, P3 does not have a snow constituent.
       if (.not.microp_scheme == 'P3') then
         ws = ws + q(k,isnow)*pdel(k)/gravit
       end if
    end do
=======
    if (irain > 1 .and. isnow > 1) then
       do k = 1, pver
          wr = wr + q(k,irain)*pdel(k)/gravit
          ws = ws + q(k,isnow)*pdel(k)/gravit
       end do
    end if
>>>>>>> upstream/master
```

Solution: Take upstream but add the P3 check before changing ws

### File: components/eam/src/physics/cam/phys_control.F90

```
<<<<<<< HEAD
      rrtmg_temp_fix 
=======
      rrtmg_temp_fix, ideal_phys_option
>>>>>>> upstream/master
```

Solution: Take upstream


ambrad (approved):
The radiation and build-namelist conflict resolutions look fine. Aaron should assess the P3-related ones. Ben doesn't need to review this.

AaronDonahue:
@jgfouca , I'm concerned about the P3 energy conservation fixes.  If P3 is active than isnow will not be >1 but irain will be and we want to still do the calculation just ignoring snow.  With the conflict fixes above I think we will just always skip adding rain to the check with P3.

components/eam/src/physics/cam/check_energy.F90
```diff
@@ -1228,18 +1228,21 @@ subroutine energy_helper_eam_def_column(u,v,T,q,ps,pdel,phis, &
...
+    end if
 
-    do k = 1, pver
-       wr = wr + q(k,irain)*pdel(k)/gravit
-       ! Recall, P3 does not have a snow constituent.
-       if (.not.microp_scheme == 'P3') then
-         ws = ws + q(k,isnow)*pdel(k)/gravit
-       end if
-    end do
+    if (irain > 1 .and. isnow > 1) then
```
AaronDonahue, line 1238: I think we will need to break this if up into an `if (irain>1)` and a `if (isnow>1 && microp_scheme!="P3")` set of conditionals.

AaronDonahue: maybe add an "else ws(i) = 0` and then we can remove the `P3` conditional up above.

AaronDonahue (changes requested):

jgfouca:
@AaronDonahue , ok, see if you like my latest commit.

components/eam/src/physics/cam/check_energy.F90
```diff
@@ -1146,28 +1146,24 @@ subroutine energy_helper_eam_def(u,v,T,q,ps,pdel,phis, &
...
           call energy_helper_eam_def_column(u(i,:),v(i,:),T(i,:),q(i,1:pver,1:pcnst),&
                                    ps(i),pdel(i,:),phis(i), &
                                    ke(i),se(i),wv(i),wl(i),wi(i),wr(i),ws(i),te(i),tw(i) )                             
+          ! In the case where the micro-physics scheme is P3 there is no snow
+          ! constituent and isnow = -1.  So we still calculate the rest of the
+          ! consituents and ensure that ws = 0.0.  NOTE! This change will likely
+          ! lead to conflicts with upstream E3SM any time check_energy is
+          ! changed...  The most important thing is to avoid any SNOW calculations
+          ! when P3 is the microphysics scheme.
+          if (microp_scheme == 'P3') then
```
AaronDonahue, line 1160: Looks much better, not to be nitpicky but we can probably move this conditional to down below with the P3 check.  Then this chunk of code won't be too out of sync with upstream master.

AaronDonahue (changes requested):
@jgfouca , looks much better.  Just one more minor fix I would suggest.

jgfouca:
@AaronDonahue , done. Note that this changes behavior a bit. The `ws` field won't be zero'd in the subroutine `energy_helper_eam_def`; it will be zero'd in `energy_helper_eam_def_column`.

jgfouca:
Ah, I see `energy_helper_eam_def` calls `energy_helper_eam_def_column`. So behavior is not really changed.

AaronDonahue (approved):
@jgfouca , thanks for making the changes.  I can see that it will change the behavior a little bit but I think functionally we will get the same behavior, with ws being zero for all columns.

PR #1825, opened 2022-07-27
This PR ensures that SCORPIO time indices are 0-based in C++ and 1-based in Fortran.

### LINKED ISSUE #1815: time_index in scream-scorpio interface starts at 1 in the cpp interface not 0
In a conversation with @jeff-cohere we figured out that there may be some confusion with C++ developers when using the scream/scorpio interface with respects to the time index.  

The natural way to set the first time snap in a file is to use 1, but in C++ we think of 0 as being the first entry in an array.  In the current implementation of the time_index we assume the user knows to use "1" to represent the first time slice, and so on.

So far this isn't an issue since only SPA uses this feature and it passes the current month which follow 1-12.  But future users may not recognize this nuance.  This is made more confusing by the default value for the time slice is -1, giving the impression that 0 is the first real value that can be passed.

I'm making this issue so we don't lose sight of this inconsistency and to open up discussion on what approach we want to use.

Note the fix is easy.
1. If we want to keep the time_index starts with 1 convention then we should add comments to that effect and change the default to 0.
2. If we want to change to 0 then there are a few updates to the spa functions needed and we need to pass the time_index+1 to the F90 interface.

jeff-cohere:
Fixes #1815

### DIFF
--- a/components/scream/src/share/io/scream_scorpio_interface.hpp
+++ b/components/scream/src/share/io/scream_scorpio_interface.hpp
@@ -56,8 +56,8 @@ namespace scorpio {
   void pio_update_time(const std::string &filename, const Real time);
 
   // Read data for a specific variable from a specific file. To read data that
-  // isn't associated with a time index, set time_index to -1. Otherwise use
-  // the proper zero-based time index.
+  // isn't associated with a time index, or to read data at the most recent
+  // time, set time_index to -1. Otherwise use the proper zero-based time index.
   void grid_read_data_array (const std::string &filename, const std::string &varname,
                              const int time_index, void* hbuf);
   /* Write data for a specific variable to a specific file. */
--- a/components/scream/src/share/io/scream_scorpio_interface.hpp
+++ b/components/scream/src/share/io/scream_scorpio_interface.hpp
@@ -55,10 +55,14 @@ namespace scorpio {
   /* Called each timestep to update the timesnap for the last written output. */
   void pio_update_time(const std::string &filename, const Real time);
 
-  /* Read data for a specific variable from a specific file. */
-  void grid_read_data_array (const std::string &filename, const std::string &varname, const int time_index, void* hbuf);
+  // Read data for a specific variable from a specific file. To read data that
+  // isn't associated with a time index, set time_index to -1. Otherwise use
+  // the proper zero-based time index.
+  void grid_read_data_array (const std::string &filename, const std::string &varname,
+                             const int time_index, void* hbuf);
   /* Write data for a specific variable to a specific file. */
-  void grid_write_data_array(const std::string &filename, const std::string &varname, const Real* hbuf);
+  void grid_write_data_array(const std::string &filename, const std::string &varname,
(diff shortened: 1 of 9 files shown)


AaronDonahue (approved):
looks good to me.

AaronDonahue:
@bartgol , I'd be curious to get your eyes on this.  It's a simple fix but touches IO code.

jeff-cohere:
Let me circle round and fix those tests.

components/scream/src/share/io/scream_scorpio_interface.hpp
```diff
@@ -55,10 +55,14 @@ namespace scorpio {
   /* Called each timestep to update the timesnap for the last written output. */
   void pio_update_time(const std::string &filename, const Real time);
 
-  /* Read data for a specific variable from a specific file. */
-  void grid_read_data_array (const std::string &filename, const std::string &varname, const int time_index, void* hbuf);
+  // Read data for a specific variable from a specific file. To read data that
```
bartgol, line 58: IIRC, putting an "invalid" time index was also a way to signal "read the last time index, if time is one of the dimensions". If so, perhaps we could make this clear in the comment.

jeff-cohere:
Oof, I really wish it were easier to build SCREAM locally. I'm willing to pitch it to make it more so, especially since it's getting to be time for us to start plugging in aerosol stuff.

bartgol:
> Oof, I really wish it were easier to build SCREAM locally. I'm willing to pitch it to make it more so, especially since it's getting to be time for us to start plugging in aerosol stuff.

It should be relatively easy. Are you referring to the need for mach specs/file for building with `test-all-scream` on a local machine?

jeff-cohere:
Yeah, I'm referring to the amount of inference needed by the installer. Is there in fact a definitive written process for building SCREAM and running its tests locally? I see several Confluence pages on the topic, of varying age and relevance, but nothing that reflects the current process. I had a working build a few weeks back, but then had to update my MPI installation and am back in dev-env jail.

Right now, I'm struggling with strange Fortran-related MPI errors trying to get pNetCDF built. The intersection of environments that can handle all the tech we're using is pretty small. :-)

AaronDonahue:
Not a solution at present, but there are a couple of interns who have been working on a SPACK build of scream standalone which presumably will make it very easy to port to any machine.  They are working out the kinks, but have a working branch in SPACK that seems to run on in-house machines not yet supported in the SCREAM code.

bartgol:
@jeff-cohere You should ask to the folks that were doing the documentation...oh wait, that's us.

I assume you already created `~/.cime/scream_mach_specs.py` and `~/.cime/scream_mach_file.cmake`? Those are the files that test-all-scream checks when you pass the `-l/--local` flag.

jeff-cohere:
Hey guys. Seems like this one's ready to go. Sorry about the confusion with the tests.

bartgol (approved):
Nice. I like keeping the "classic" index base for each language (0 in C/C++ and 1 in Fortran).

PR #1826, opened 2022-07-27
mem_check tests need to use different build dirs

jgfouca:
... to avoid colliding with other tests.


jgfouca:
I think this was the source of the apparent race condition in the testing. With the original refactor PR, the mem_check tests started using the same build directory as the full_debug tests. I'm still don't 100% understand the issue, because the mem_check tests were never concurrent with the other tests.

bartgol (approved):

PR #1830, opened 2022-08-02
Bugfix to check_energy in EAM

AaronDonahue:
This commit fixes a minor bug in check_energy.F90 in EAM where the
water snow content incorrectly references the ith column when it is
passed as just a scalar to a sub-function.


bartgol (approved):

jgfouca:
Thanks @AaronDonahue !

AaronDonahue:
It looks like all we get are NL fails, which I think we expected.  If that sounds good I'm going to manually merge.

AaronDonahue:
@bartgol or @jgfouca , if it is easy do we want to fire up the nightlies for SCREAMv0 on `mappy` to ensure everything is good and so not to hold up any incoming PR's?

PR #1831, opened 2022-08-02
EAMxx: Integrate pg2 and SL transport into SCREAMv1.

ambrad:
The core operations (files ComposeTransport*, GllFvRemap*) are in Hommexx and
were brought in through upstream merges. This PR makes EAMxx-side interface
changes to enable these capabilities in SCREAMv1.

Grid selectors in namelist_defaults_scream.xml make it so that any compset
specification with pg2 in the grid alias also triggers SL transport.

This PR adds three new ERS/P v1 tests.

A number of things still need to be done:

1. fvphyshack is a global variable that works around some initialization issues:
   1a. active_gases for rrtmgp,
   1b. some grid/field manager stuff in the AD.
1a can be backed out when the active_gases are fully initialized separately from
the IC file. 1b will require some discussion and bits of AD redesign or
correction of what I missed.

2. I've noticed qv min in the Homme state output can be <0, unlike in
E3SMv2. I've traced this preliminarily to Hommexx's vertical remap (but perhaps
only indirectly) but need to do more analysis and work out a fix. Eventually we
should be able to remove the lower-bound check in dynamics when using SL.

3. There seems to be a PE sensitivity somewhere in (very preliminary diagnosis)
postcondition checking on the GPU, I think unrelated to this PR. The new ERP
tests this PR includes should pass, but others I've tried do not.

[BFB] except for new ERS/P v1 tests

### DIFF
--- a/components/scream/src/control/atmosphere_driver.cpp
+++ b/components/scream/src/control/atmosphere_driver.cpp
@@ -14,6 +14,8 @@
 #include "ekat/ekat_parse_yaml_file.hpp"
 #include "ekat/std_meta/ekat_std_utils.hpp"
 
+#include "control/fvphyshack.hpp"
+
 #include <fstream>
 
 namespace scream {
@@ -108,6 +110,13 @@ set_params(const ekat::ParameterList& atm_params)
   create_logger ();
 
   m_ad_status |= s_params_set;
+
+  const auto hgn = "Physics PG2";
+  fvphyshack = m_atm_params.sublist("Grids Manager").get<std::string>("Reference Grid") == hgn;
+  if (fvphyshack) {
+    // See the [rrtmgp active gases] note in dynamics/homme/atmosphere_dynamics_fv_phys.cpp.
+    fv_phys_rrtmgp_active_gases_init(m_atm_params);
+  }
 }
 
 void AtmosphereDriver::
@@ -236,13 +245,14 @@ void AtmosphereDriver::create_fields()
             "   grid name: " + req.grid + "\n");
         // Given request for group A on grid g1 to be an Import of
         // group B on grid g2, register each field in group B in the
-
         // field manager on grid g1.
         const auto& fm = m_field_mgrs.at(req.grid);
         const auto& rel_fm   = m_field_mgrs.at(req.src_grid);
         const auto& rel_info = rel_fm->get_groups_info().at(req.src_name);
 
-        auto r = m_grids_manager->create_remapper(req.src_grid,req.grid);
+        // In the case of pg2, can't use create_remapper. But the remapper is
+        // used only for a convenience function, anyway, create_tgt_fid.
+        auto r = fvphyshack ? nullptr : m_grids_manager->create_remapper(req.src_grid,req.grid);
         // Loop over all fields in group src_name on grid src_grid.
         for (const auto& fname : rel_info->m_fields_names) {
           // Get field on src_grid
@@ -250,9 +260,19 @@ void AtmosphereDriver::create_fields()
 
           // Build a FieldRequest for the same field on greq's grid,
           // and add it to the group of this request
-          auto fid = r->create_tgt_fid(f->get_header().get_identifier());
-          FieldRequest freq(fid,req.name,req.pack_size);
-          fm->register_field(freq);
+          if (fvphyshack) {
+            const auto& sfid = f->get_header().get_identifier();
+            auto dims = sfid.get_layout().dims();
+            dims[0] = fm->get_grid()->get_num_local_dofs();
+            FieldLayout fl(sfid.get_layout().tags(), dims);
+            FieldIdentifier fid(sfid.name(), fl, sfid.get_units(), req.grid);
+            FieldRequest freq(fid,req.name,req.pack_size);
+            fm->register_field(freq);
+          } else {
+            const auto fid = r->create_tgt_fid(f->get_header().get_identifier());
+            FieldRequest freq(fid,req.name,req.pack_size);
+            fm->register_field(freq);
+          }
         }
 
         // Now that the fields have been imported on this grid's GM,
@@ -323,11 +343,18 @@ void AtmosphereDriver::create_fields()
   // the internal fields/groups, and mark them as part of the RESTART group.
   for (const auto& f : m_atm_process_group->get_fields_in()) {
     const auto& fid = f.get_header().get_identifier();
-    auto fm = get_field_mgr(fid.get_grid_name());
+    const auto& gn = fid.get_grid_name();
+    if (fvphyshack and gn == "Physics CGLL") {
+      // CGLL is used only in the IC field for Homme.
+      continue;
+    }
+    auto fm = get_field_mgr(gn);
     fm->add_to_group(fid.name(),"RESTART");
   }
   for (const auto& g : m_atm_process_group->get_groups_in()) {
-    auto fm = get_field_mgr(g.grid_name());
+    const auto& gn = g.grid_name();
+    auto fm = get_field_mgr(gn);
+    if (fvphyshack and gn == "Physics CGLL") continue; // see above
     if (g.m_bundle) {
       fm->add_to_group(g.m_bundle->get_header().get_identifier().name(),"RESTART");
     } else {
@@ -338,7 +365,9 @@ void AtmosphereDriver::create_fields()
   }
   for (const auto& f : m_atm_process_group->get_internal_fields()) {
     const auto& fid = f.get_header().get_identifier();
-    auto fm = get_field_mgr(fid.get_grid_name());
+    const auto& gn = fid.get_grid_name();
+    if (fvphyshack and gn == "Physics CGLL") continue; // see above
+    auto fm = get_field_mgr(gn);
     fm->add_to_group(fid.name(),"RESTART");
   }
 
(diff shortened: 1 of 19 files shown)


ambrad:
Everything passes except the expected DIFFs on Mappy for the new pg2 ERS/P tests: 
```
# old tests PASS
PASS ERP_D_Ln9.ne4_ne4.F2010-SCREAMv1.mappy_gnu9 RUN
PASS ERS_Ln9.ne4_ne4.F2000-SCREAMv1-AQP1.mappy_gnu9 RUN
PASS SMS_D_Ln9.ne4_ne4.F2010-SCREAMv1-noAero.mappy_gnu9 RUN
# new tests DIFF since there are no baselines
DIFF ERP_Ln22.ne4pg2_ne4pg2.F2010-SCREAMv1.mappy_gnu9 (phase BASELINE)
DIFF ERS_D_Ln22.ne4pg2_ne4pg2.F2010-SCREAMv1.mappy_gnu9 (phase BASELINE)
```

ambrad:
@bartgol @AaronDonahue see my comment above summarizing that the AT results, after a few fixes, are now as expected.

components/scream/src/control/atmosphere_driver.cpp
```diff
@@ -323,11 +348,18 @@ void AtmosphereDriver::create_fields()
   // the internal fields/groups, and mark them as part of the RESTART group.
   for (const auto& f : m_atm_process_group->get_fields_in()) {
     const auto& fid = f.get_header().get_identifier();
-    auto fm = get_field_mgr(fid.get_grid_name());
+    const auto& gn = fid.get_grid_name();
+    if (fvphyshack and gn == "Physics CGLL") {
```
bartgol, line 352: I suspect this (as well as below) might have to be changed to "Physics GLL", to match what's used elsewhere.

components/scream/src/dynamics/homme/atmosphere_dynamics.hpp
```diff
@@ -10,6 +10,8 @@
 namespace scream
 {
 
+class FieldManager;
```
bartgol, line 13: Looks like this is not needed?

bartgol (commented):
Looks good. I only have one comment on "Physics CGLL" vs "Physics GLL", but other than that I think it's good to go.

components/scream/src/control/atmosphere_driver.cpp
ambrad: Good catch. That actually indicates that that workaround is not needed, assuming the restart file looks fine. I'll check and delete the workaround if it's fine.

components/scream/src/dynamics/homme/atmosphere_dynamics.hpp
ambrad: True. I'll remove it.

components/scream/src/control/atmosphere_driver.cpp
ambrad: The restart file has no sign of the temporary CGLL fields, so the hack above (corrected) isn't needed. I don't actually understand why that's true, as I thought add_field'ing the temporary CGLL fields as Required would mean they'd appear in the restart file.

bartgol: That should be the case. Are the fields you refer to those that you later remove (with the newly added feature "remove_field" in AtmosphereProcess)?

bartgol (approved):

components/scream/src/control/atmosphere_driver.cpp
ambrad: Oh, I see why. I had two layers of hacks, but only one was needed. The part of the code with the comment "// Don't save CGLL fields from ICs to the restart file." strips those fields from the restart file, so this first hack, whether written correctly or not, isn't needed.

ambrad: > That should be the case. Are the fields you refer to those that you later remove (with the newly added feature "remove_field" in AtmosphereProcess)?

That's not a great feature and probably something we should discuss that motivates a more flexible init process. In any case, that field removal I think would lead to a restart-file-write error later on if it weren't for the chunk of code starting with "// Don't save CGLL fields from ICs to the restart file", which makes the restart I/O object not see these fields.

components/scream/src/dynamics/homme/atmosphere_dynamics_fv_phys.cpp
```diff
@@ -0,0 +1,320 @@
...
+    get_group_in("tracers", gn).m_bundle->get_view<const Real***>().data(),
+    nelem, npg, nq, nlev);
+  
+  gfr.run_fv_phys_to_dyn(time_idx, T, uv, q);
+  Kokkos::fence();
+  gfr.run_fv_phys_to_dyn_dss();
+  Kokkos::fence();
+}
+
+// [rrtmgp active gases] This is to address issue #1782. It supports option 1 in
```
AaronDonahue, line 219: Would it be helpful to add the string "TODO" to this comment so that it is easily identifiable?

components/scream/src/dynamics/homme/atmosphere_dynamics_fv_phys.cpp
```diff
@@ -0,0 +1,320 @@
...
+  kgkg.set_string("kg/kg");
+  const auto& rgn = m_cgll_grid->name();
+  const auto& pgn = m_phys_grid->name();
+  const auto rnc = m_cgll_grid->get_num_local_dofs();
+  const auto pnc = m_phys_grid->get_num_local_dofs();
+  const auto nlev = m_cgll_grid->get_num_vertical_levels();
+  constexpr int ps = SCREAM_SMALL_PACK_SIZE;
+  for (const auto& e : s_tgw.active_gases) {
+    add_field<Required>(e, FieldLayout({COL,LEV},{rnc,nlev}), kgkg, rgn, ps);
+    // 'Updated' so that it gets written to the restart file.
```
AaronDonahue, line 258: It is strange that we need "Updated" to ensure a write to the restart file.  My understanding is that "Required" should be enough, because that essentially says I need this field to be set before I run.  Does FV actually change the active gases so that it is indeed an "updated" variable?  If not and this is a hack to get it written in the restart file then there might be something we need to fix.  @bartgol any thoughts?

AaronDonahue (approved):
I had a few questions, but nothing major.  Looks good, excited to get PG2 into EAMxx

components/scream/src/control/atmosphere_driver.cpp
```diff
@@ -14,6 +14,9 @@
 #include "ekat/ekat_parse_yaml_file.hpp"
 #include "ekat/std_meta/ekat_std_utils.hpp"
 
+#include "scream_config.h" // for SCREAM_CIME_BUILD
+#include "control/fvphyshack.hpp"
```
AaronDonahue, line 18: I forgot to add this comment...

From the filename I'm guessing this is meant to be a hack that won't be needed in the future?  If so, should we have a comment somewhere making that clear?  If nothing more than to make it easy to track down which code is a hack.

components/scream/src/dynamics/homme/atmosphere_dynamics_fv_phys.cpp
```diff
@@ -0,0 +1,320 @@
...
+    t ? t->horiz_winds.data() : get_field_out("horiz_winds", gn).get_view<Real***>().data(),
+    nelem, npg, 2, nlev);
+  const auto q = Homme::GllFvRemap::Phys3T(
+    t ? t->tracers.data() : get_group_out("tracers", gn).m_bundle->get_view<Real***>().data(),
+    nelem, npg, nq, nlev);
+  const auto dp = Homme::GllFvRemap::Phys2T(
+    get_field_out("pseudo_density", gn).get_view<Real**>().data(),
+    nelem, npg, nlev);
+
+  gfr.run_dyn_to_fv_phys(time_idx, ps, phis, T, omega, uv, q, &dp);
```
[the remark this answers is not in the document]
bartgol: `&dp` passes the address of `dp`, so you end up passing a pointer. This allows to do two different things, based on whether the passed pointer is valid or null.

components/scream/src/dynamics/homme/atmosphere_dynamics_fv_phys.cpp
bartgol: I suspect it's b/c `Required` fields are read-only, while here we need to read the gases from a GLL grid and remap them to pg2. So on top of being Required (for restart reasons), the pg2 version needs to also be Computed (to be writable). Hence, Updated.

components/scream/src/control/atmosphere_driver.cpp
bartgol: I think it's ok. `fvphyshack` is quite verbose anyways, and all the code that (may) need to be revisited is enclosed in `if (fvphyshack)`, which is easy to grep for.

components/scream/src/dynamics/homme/atmosphere_dynamics_fv_phys.cpp
ambrad: I can do that. Is "TODO [rrtmgp active gases]" fine? The bracket thing I do just for myself to group related comments in a grep-able way.

components/scream/src/dynamics/homme/atmosphere_dynamics_fv_phys.cpp
ambrad: Right. Required for restart + Computed b/c I'm mapping from CGLL IC input to pg2 for simulation. I can add to the comment to clarify this.

components/scream/src/control/atmosphere_driver.cpp
ambrad: Since I'm adding a commit with comment additions, I'll add a comment about fvphyshack.

components/scream/src/dynamics/homme/atmosphere_dynamics_fv_phys.cpp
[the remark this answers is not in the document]
AaronDonahue: ahh okay cool, thanks.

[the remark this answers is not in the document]
ambrad: Right. EAMxx requires dp_fv whereas EAM doesn't. Thus, the API already supported not providing it but using it internally. One of the upstream changes I made was to add an output dp as an optional argument.

components/scream/src/dynamics/homme/atmosphere_dynamics_fv_phys.cpp
AaronDonahue: Absolutely. I like the TODO string because it jumps out at the developer.

ambrad:
@AaronDonahue I added a commit with comments.

AaronDonahue (approved):
Thanks Andrew

PR #1837, opened 2022-08-04
Changes to PropertyCheck classes

### LINKED ISSUE #1833: Post-condition positivity check sets fields to eps
I'm investigating a few different issues in SCREAMv1 that are unrelated in terms of code but related in terms of clues. In the process I've stumbled upon something I believe we definitely do not want.

The dynamics sets a repairable post-condition on the tracers using the positivity checker. That checker then makes every value equal to eps (~2.22e-16) that was originally <= eps. Thus, fields that normally can be and often are 0 in large regions of the earth (true of many of the microphysics fields) at any given time end up being >= eps after every dynamics call.

My guess is this influences climate a little and probably water conservation over time.

Assuming everyone agrees this is a problem, the positivity checker should be replaced by a nonnegativity checker. IMO, a positivity checker that hardcodes eps as the lower bound is easily misused and so should not be available. The domain expert should be the one to set lower bounds other than 0.

### LINKED ISSUE #1834: Issue with CheckAndRepairWrapper
I think Homme's use of CheckAndRepairWrapper (CARW) is wrong. And I'm also wondering if that class should exist at all.

Originally, the intent of CARW was to allow different tolerances during check and repair phases. The goal was, say for a LB check, to crap out if field F is "significantly below LB", and repair values "just below LB". But that's not what's happening. So long as CARW stores a PropCheck repair object, the check method will never cause the overall CARW to fail, and the repair object will always clip to LB.

I think I have to change the concept of [CheckResult](https://github.com/E3SM-Project/scream/blob/2e0f5dcf50dac46ebff6c26b4fd6cd03c99b79a3/components/scream/src/share/property_checks/property_check.hpp#L45): rather than storing a `bool pass;`, it should store an enum, with possible values along the line of `Pass, CatastrophicFail, RepairableFail`. PC classes like FieldWithinIntervalCheck should then store two LB/UB, to determine if something is within reasonable distance from the bound (and therefore fixable) or completely off.

E.g., Homme could use a FieldLowerBoundCheck with LB1=-1e-17, LB2=0. If min(F)<LB1, return `CatastrophicFail`, if LB1<min(F)<LB2, return `RepairableFail`, otherwise return `Pass`.

bartgol:
A few changes:

- Remove `FieldPositivityCheck`. It was only used where a `FieldLowerBoundCheck` (with lb=0) was more appropriate, since it avoids introducing spurious non-zero values.
- Changed return type of `PropertyCheck::check()`. The return type contains an enum, with possible values `Pass` (all good), `Fail` (no hope, crash immediately), `Repairable` (check failed, but we can fix it).
- Added "repairable" bounds to `FieldWithinIntervalCheck` (FWIC). The new bounds are no-tighter than the bounds. The class returns `Pass` if `lb<=F<=ub`, `Fail` if `F<lb_repairable` or `F>ub_repairable`, and `Repairable` otherwise. The "repairable" bounds default to an infinitely large tolerance, so that, if repair is enabled, all values outside `[lb,ub]` are clipped to `lb`/`ub`. 
- Remove `CheckAndRepairWrapper`. It was meant to allow two different bounds in check and repair phases, but it was ill-designed. Besides, the aforementioned mod to FWIC makes the wrapper no longer needed.

I expect this to be NON-BFB, since it changes how values are clipped in certain field reparations.

Fixes #1833
Fixes #1834


ambrad:
@bartgol I suggest that you run the old code with the correction to the positivity checks (0 instead of eps) and generate baselines, then compare this PR's results to those baselines. That way, you can be sure that the rest of the redesign doesn't change answers. Recall that we like non-BFB (for existing tests) PRs to be as short as possible, with a second PR bringing in BFB stuff.

components/scream/src/physics/shoc/atmosphere_macrophysics.cpp
```diff
@@ -380,13 +380,18 @@ void SHOCMacrophysics::initialize_impl (const RunType run_type)
...
   // Set field property checks for the fields in this process
-  add_postcondition_check<FieldWithinIntervalCheck>(get_field_out("T_mid"),m_grid,140.0,500.0,false);
-  add_postcondition_check<FieldWithinIntervalCheck>(get_field_out("qv"),m_grid,1e-13,0.2,true);  // This is a quick fix to make sure qv doesn't go negative.  TODO item is to take care of this in a more clever way.
-  add_postcondition_check<FieldWithinIntervalCheck>(get_field_out("qc"),m_grid,0.0,0.1,false);
-  add_postcondition_check<FieldWithinIntervalCheck>(get_field_out("horiz_winds"),m_grid,-400.0,400.0,false);
-  add_postcondition_check<FieldPositivityCheck>(get_field_out("pbl_height"),m_grid);
-  add_postcondition_check<FieldWithinIntervalCheck>(get_field_out("cldfrac_liq"),m_grid,0.0,1.0,false);
-  add_postcondition_check<FieldPositivityCheck>(get_field_out("tke"),m_grid);
+  auto eps = std::numeric_limits<Real>::epsilon();
+  using FWIC = FieldWithinIntervalCheck;
```
jeff-cohere, line 384: Not a big deal, but maybe `WithinInterval` and `LowerBound` might be more eye-catching than `FWIC` and `FLBC` in the lines that follow?

components/scream/src/physics/spa/atmosphere_prescribed_aerosol.cpp
```diff
@@ -189,19 +188,15 @@ void SPA::initialize_impl (const RunType /* run_type */)
...
-    add_postcondition_check<CheckAndRepairWrapper>(check, repair);
-  }
-  // upper bound set to 1.01 as max(g_sw)=1.00757 in current ne4 data assumingly due to remapping
-  // add an epslon to max possible upper bound of aero_ssa_sw
-
-  add_postcondition_check<FieldWithinIntervalCheck>(get_field_out("aero_g_sw"),m_grid,0.0,1.0,true);
-  add_postcondition_check<FieldWithinIntervalCheck>(get_field_out("aero_ssa_sw"),m_grid,0.0,1.0,true);
-  add_postcondition_check<FieldWithinIntervalCheck>(get_field_out("aero_tau_sw"),m_grid,0.0,1.0,true);
-  add_postcondition_check<FieldWithinIntervalCheck>(get_field_out("aero_tau_lw"),m_grid,0.0,1.0,true);
+  using FWIC = FieldWithinIntervalCheck;
```
jeff-cohere, line 191: Consider `WithinInterval` here if you adopt my suggestion above.

bartgol:
@ambrad good point. I can actually do a first mini-pr that simply changes the LB for positivity (turning it into a "non-negativity" check). Then do the rest of this PR on top of that.

components/scream/src/share/atm_process/atmosphere_process.cpp
```diff
@@ -185,31 +185,35 @@ void AtmosphereProcess::run_precondition_checks () const {
   for (const auto& it : m_precondition_checks) {
     const auto& pc = it.second;
 
-    auto check_result = pc->check();
-    if (check_result.pass) {
+    auto res_and_msg = pc->check();
```
jeff-cohere, line 188: You likely know this, but C++17 has a neat way to handle multiple return values. (It also fixes some `constexpr`-related defects in C++14, which has pushed us to adopt it in our EAGLES work, which is why I mention this).

bartgol: Yes, I think it's called "structured binding", or something like that. I think we officially require only C++14 though, so we can't use it (yet).

jeff-cohere: Fair enough. But I think we should consider moving to C++17 when practical. It's a measurable improvement over C++14 (and I say this as a C++ skeptic!).

components/scream/src/share/atm_process/atmosphere_process.cpp
```diff
@@ -185,31 +185,35 @@ void AtmosphereProcess::run_precondition_checks () const {
...
+    if (res_and_msg.result==CheckResult::Pass) {
       continue;
-    }
-
-    // Failed check.
-    if (pc->can_repair()) {
-      // Ok, just fix it
+    } else if (res_and_msg.result==CheckResult::Repairable && pc->can_repair()) {
```
jeff-cohere, line 191: Considering that `pc` emits `res_and_msg`, would it ever be the case that `res_and_msg.result==CheckResult::Repairable` if `pc->can_repair` is `false`? If so, why?

jeff-cohere (commented):
Looks good to me. I made a few comments/suggestions. I'll hold off on approving till the mini PR goes through.

components/scream/src/share/atm_process/atmosphere_process.cpp
bartgol: I think next Kokkos release will be a good time to do it. IIRC, Kokkos 3.7 will _require_ C++17. The efforts to support HIP in ekat/scream might push us to adopt latest kokkos anyways.

components/scream/src/physics/spa/atmosphere_prescribed_aerosol.cpp
```diff
@@ -189,19 +188,15 @@ void SPA::initialize_impl (const RunType /* run_type */)
...
-  // add an epslon to max possible upper bound of aero_ssa_sw
-
-  add_postcondition_check<FieldWithinIntervalCheck>(get_field_out("aero_g_sw"),m_grid,0.0,1.0,true);
-  add_postcondition_check<FieldWithinIntervalCheck>(get_field_out("aero_ssa_sw"),m_grid,0.0,1.0,true);
-  add_postcondition_check<FieldWithinIntervalCheck>(get_field_out("aero_tau_sw"),m_grid,0.0,1.0,true);
-  add_postcondition_check<FieldWithinIntervalCheck>(get_field_out("aero_tau_lw"),m_grid,0.0,1.0,true);
+  using Interval = FieldWithinIntervalCheck;
+  const auto eps = std::numeric_limits<double>::epsilon();
+
+  add_postcondition_check<Interval>(get_field_out("nccn"),m_grid,0,1e11,true,-1e11*eps);
```
ambrad, line 194: So this line will (1) fail if values are < -1e11*eps or > 1e11 and (2) otherwise clip to >= 0 , right?

bartgol: Correct.

AaronDonahue:
Just to confirm that there won't be any conflicts if we merge #1838 and this PR together?

AaronDonahue (dismissed):

AaronDonahue:
@bartgol , looking at the AT results it looks like we are getting some legit fails that we may want to check out before a manual merge.  It also looks like `blake` didn't run for the tests.

bartgol:
> @bartgol , looking at the AT results it looks like we are getting some legit fails that we may want to check out before a manual merge. It also looks like `blake` didn't run for the tests.

I'll check the fails. Blake did run. When you don't see one of the build links, it is usually b/c the output triggered some markdown syntax on github, and got incorporated in the previous build. If you click on weaver, you'll see blake's output. It happens sometimes, but it's rare enough that we're ok with it.

AaronDonahue:
@bartgol , any progress on the fails reported for this PR?

bartgol:
> @bartgol , any progress on the fails reported for this PR?

Yes.

AaronDonahue:
It looks like the only failed test was on blake, but the error log looks like a build failure.  Maybe we should just retest?

bartgol:
Yeah, it looks like a FS error.
```
FATAL: Unable to delete script file /tmp/jenkins8608864540643239763.sh
```

bartgol:
Skipping mappy cause it's always super busy, and everything passed there.

Speaking of: I think ne4 ERP tests might not be deep enough to catch PE-dependent issues. I guess we can keep them to catch "gross" errors...

Note: even PEM on ne4 was passing before this PR, despite failing in master. I suspect  there is too little PE-variability when running with 96 elements.

ambrad:
We run PE tests (ERP or PEM) on ne30 in the nightlies. We could move one to the AT test set. That said, the property-check-based PE variance was in fact pretty hard to catch. For example, ERP ne30 Ln90 was fine on Chrysalis, IIRC.

jeff-cohere (approved):

PR #1838, opened 2022-08-04
Change lower bound of FieldPositivityCheck to zero

bartgol:
I believe this is the single piece of non-BFBness that #1837 introduced. I'm turning it into its own PR, to make non-BFB changes as isolated as possible.

### DIFF
--- a/components/scream/src/share/property_checks/field_positivity_check.hpp
+++ b/components/scream/src/share/property_checks/field_positivity_check.hpp
@@ -16,9 +16,7 @@ public:
   FieldPositivityCheck (const Field& f,
                         const std::shared_ptr<const AbstractGrid>& grid,
                         const bool can_repair = false)
-   : FieldLowerBoundCheck (f,grid,
-                           std::numeric_limits<double>::epsilon(),
-                           can_repair)
+   : FieldLowerBoundCheck (f,grid,0,can_repair)
   {
     // Nothing to do here
   }


AaronDonahue:
@bartgol , do you have a preferred order between this PR and #1837 ?

ambrad (approved):
Thanks!

bartgol:
> @bartgol , do you have a preferred order between this PR and #1837 ?

Oh yeah, this one goes in first, of course.

AaronDonahue:
I'm keen to have this PR be the next non-BFB PR that goes in.  I'm adding RETEST to let the AT go for it and once that finishes then I will manually merge, unless there is any push back for us to merge a separate PR instead.

ambrad:
@AaronDonahue I agree, but note that this PR won't fix the PEM/ERP tests; only https://github.com/E3SM-Project/scream/pull/1837 can.

bartgol:
@ambrad , that's true, the other PR ends up being non-BFB, because of the removal of CheckAndRepairWrapper and it's faulty PE-dependent logic. So at this point, this PR makes little sense, and we should probably fix everything in one, with that PR.

AaronDonahue:
Great. I added RETEST to #1837 .  Also wrote a comment that confirming that we won't cause any conflicts by merging the two together.  I don't want to manually merge both back-to-back and cause weird issues because master changed.

bartgol:
@AaronDonahue that PR is built on top of this one. So we should simply close this one.

Issue #1807, opened 2022-07-21
SHOC: calculate_dx_from_area assumes radian lat but gets degree lat input


ambrad:
I suspect this doesn't really matter because it just contributes to a small ellipsoid correction to the sphere, but it would be better to do the calculation correctly or just switch to a sphere assumption.

PeterCaldwell:
I'd like @bogensch comment on whether we should just delete this correction. 

If it is needed (or maybe even if it is just easier to switch radian->deg), I think this might be a good task for Hannah (whose git handle I can't find).

bogensch:
I will say that I have extensively verified the v0 version of this calculation (which also uses degree lat) produces very reasonable and representative values of dx input to SHOC for a variety of ne resolutions (and RRM) that I would expect.  Aaron and I also verified this when in v1 when he made PR #1598, so I'm very comfortable with the values currently being produced from this routine.  Thus, I don't think this is super critical for September 1st.... but of course, I think it should probably eventually be corrected to be consistent...  In summary, fixing this will likely have negligible impact on the model solution.

PR #1841, opened 2022-08-04
Remove extra rad2deg conversion in MCT interface layer

bartgol:
The lat/lon data produced by scream are already in degrees, so they do not need an extra multiplication by 180/pi. This rad->deg conversion was necessary when we were still storing lat/lon in scream in radians. At some point, those views got converted to deg, but the extra conversion in the MCT interface was not removed.

I am not 100% sure what MCT uses these for, but I'm guessing to interpolate surface fluxes among components. Hence, I expect this PR to be non-BFB.

### DIFF
--- a/components/scream/src/mct_coupling/atm_comp_mct.F90
+++ b/components/scream/src/mct_coupling/atm_comp_mct.F90
@@ -289,7 +289,6 @@ CONTAINS
     use seq_flds_mod,   only: seq_flds_dom_coord, seq_flds_dom_other
     use shr_const_mod,  only: SHR_CONST_PI
     use scream_f2c_mod, only: scream_get_cols_latlon, scream_get_cols_area
-
     !-------------------------------------------------------------------
     !
     ! Arguments
@@ -303,8 +302,6 @@ CONTAINS
     real(r8), pointer :: data1(:)
     real(r8), pointer :: data2(:)     ! temporary
     integer , pointer :: idata(:)     ! temporary
-
-    real(r8), parameter:: rad2deg = 180.0_r8 / SHR_CONST_PI
     !-------------------------------------------------------------------
 
     allocate(data1(lsize))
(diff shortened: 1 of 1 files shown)


components/scream/src/mct_coupling/atm_comp_mct.F90
```diff
@@ -322,8 +319,10 @@ subroutine atm_domain_mct( lsize, gsMap_atm, dom_atm )
     call scream_get_cols_latlon(c_loc(data1),c_loc(data2))
     call mct_gGrid_importRAttr(dom_atm,"lat",data1,lsize) 
     call mct_gGrid_importRAttr(dom_atm,"lon",data2,lsize) 
-    data1(:) = data1(:) * rad2deg
-    data2(:) = data2(:) * rad2deg
+
+    call mpi_comm_rank(mpicom_atm, rank, ierr)
```
AaronDonahue, line 323: This seems unrelated to the fix, but I'm guessing is just another chunk of code we need so might as well put it here?

AaronDonahue (dismissed):

components/scream/src/mct_coupling/atm_comp_mct.F90
bartgol: Ah, I put it in while debugging. Let me prune it.

PeterCaldwell:
Does this fix https://github.com/E3SM-Project/scream/issues/1807 (calculate_dx_from_area assumes radian lat but gets degrees)? Heads up to @bogensch in the unlikely event that I'm right.

ambrad:
No, I think these lines do not affect what scream sees for lat/lon.

bartgol:
I'm a bit surprised that this PR is bfb. I guess I don't really know what the coupler uses lat/lon for. @rljacob do you know?

rljacob:
Those are part of initalizing the "domain" data structure which is supposed to have useful info about the grid like lat lon even if it isn't used. Any area weights are read in by the coupler and not recalculated from lat,lon.

bartgol:
> Those are part of initalizing the "domain" data structure which is supposed to have useful info about the grid like lat lon even if it isn't used. Any area weights are read in by the coupler and not recalculated from lat,lon.

So is it plausible that fixing lat/lon values would not yield any DIFF?

rljacob:
Yes it is.

bartgol:
Great. @AaronDonahue , you're ready for review then.

AaronDonahue (approved):

PR #1843, opened 2022-08-04
Bug fix to calculate_dx

bogensch:
Fixes issue #1807.  Converts latitude from degrees to radians to be used in calculation estimating distance between two points on an ellipsoid.  As expected, the impact of this change is very minor.  This was verified in an offline calculation to compute dx for the areas provided by initial condition files for ne4, ne30, ne120, and ne1024.

This is a (very minor) answer changing PR for ne1024 but probably will be b4b for ne30 and ne4 as SHOC's length scale likely never gets large enough to be clipped at those resolutions.


AaronDonahue:
@ambrad , this looks ready to merge if you want to give it the green-light.

ambrad (approved):

ambrad:
@bogensch would you summarize why this PR is BFB? I believe it's because there's some length threshold in SHOC that isn't being triggered in any AT simulation-type test, so these tests, other than the comparisons in the unit test, are insensitive to the mod. Right?

bogensch:
Typical values for SHOC's length scale are 2.5 km or less, though it can occasionally get as high as 10 km in some instances.  Therefore, for ne30 and ne4 cases the length scale will never be clipped by the grid length value.  I also wouldn't be surprised if ne120 was b4b, however, ne1024 will most certainly be non b4b as large length scale values will likely be clipped.

PR #1849, opened 2022-08-08
Clean up summit machines

jgfouca:
Merge summit-gpu and summit-cpu. Use gpu substring in compiler names to identify GPU builds.


jgfouca:
Tested via `create_test e3sm_scream_v1 --compiler=gnugpu`. Just occurred to me that I should probably test `--compiler=gnu` too so I will do that now.

jgfouca:
Adding @bartgol because I touched some homme stuff.

ambrad:
I think a lot of this -- at least the mods in the Homme directory -- should go in the upstream repo.

jgfouca:
@ambrad , the changes to cime_config in this PR are basically undoing changes that were made in SCREAM that were never made upstream so this PR should bring us into closer alignment with upstream E3SM.

I'm less familiar with how `summit-gpu` spread through homme. I was thinking it must have happened in SCREAM since there never was a `summit-gpu` in upstream e3sm.

ambrad:
I might be missing something, but I spot-checked a few Homme changes, and to me it looks like the mods you're making are in places that are identical to the upstream.

jgfouca:
@ambrad , ah ok. I can maybe throw out the homme mods if homme is managing its own machine names and not getting them from CIME or SCREAM. @bartgol ?

jgfouca:
Summit+gnu (no gpu) was 100% PASS for e3sm_scream_v1

bartgol:
I agree with @ambrad : all mods in `cime_config` and `components/homme` should come from upstream.

jgfouca:
@bartgol , I'm going to need a bit more info. Does homme manage it's own it own machine names? Where did summit-gpu come from?

The changes to cime_config cannot go into upstream master because this PR is undoing changes that were made on the SCREAM side that were never made upstream.

ambrad:
The standalone Homme build does. Looks like summit-gpu came into the standalone Homme build system in this commit: https://github.com/E3SM-Project/E3SM/commit/d3483227ee015d7d9a13fe6e076f67a998f70ca1

bartgol:
> The changes to cime_config cannot go into upstream master because this PR is undoing changes that were made on the SCREAM side that were never made upstream.

Ah, I understand.

jgfouca:
Can all the homme changes just be tossed, then? @bartgol ?

bartgol:
> Can all the homme changes just be tossed, then? @bartgol ? Is there any connection between CIME machine names and homme machine names?

I think you can toss them, yes. The macros/files you touched should only be used in Homme standalone testing, which scream doesn't do.

jgfouca:
OK, squashed out the homme changes so they are gone and no longer in the history.

PR #1850, opened 2022-08-09
Always ensure csm_share is available

bartgol:
Even in standalone builds. This replace the ad-hoc built lib `share_util` that we were building inside rrtmgp (which was a duplicate in CIME builds).


brhillman (approved):
Looks good to me.

PR #1853, opened 2022-08-10
Fix output redirection in CIME runs

bartgol:
The way we handled homme output was wrong, in the sense that we let homme print to screen, and tried to capture stdout inside scream. This resulted in a clash with e3sm.log, and some output could later be found at random locations.

Fixing it required some changes in Homme, which are currently in a PR upstream. I included the same commits here, but I will rebase the PR on master once the upstream one goes in and we do an upstream merge.

With this PR the output is as follows:
- homme_atm.log: contains all the ~~print~~ `write(iulog,*)` statements from homme
- atm.log: contains all the output from the atm logger
- e3sm.log: contains e3sm stuff PLUS all explicit print/write/printf/cout/cerr statements (i.e., prints that go directly to stdout/stderr)

The warnings about prop checks can be moved in atm.log easily, if we wanted to. We still have to work on a mechanism to limit the number of times the check fails get printed, to keep the pollution of the logs under control.


ambrad:
Would you tell us where these cases will go:
1. print * in any Homme or SCREAM F90 file on any rank. Based on the above, it seems like Homme print * output will appear in homme_atm.log. What about SCREAM?
2. C++ output to stderr in any Homme or SCREAM C++ file on any rank. Same as 1 for Homme?
3. C++ output to stdout in any Homme or SCREAM C++ file on any rank. Same as 1 for Homme?

Thanks.

bartgol:
> 1. print * in any Homme or SCREAM F90 file on any rank. Based on the above, it seems like Homme print * output will appear in homme_atm.log. What about SCREAM?
> 2. C++ output to stderr in any Homme or SCREAM C++ file on any rank. Same as 1 for Homme?
> 3. C++ output to stdout in any Homme or SCREAM C++ file on any rank. Same as 1 for Homme?

`print` commands in F90 are outputing to stdout, which remains linked to e3sm.log. Similarly, `printf(...)/fprint(stdout,...)/fprintf(stderr,...)` in C++ are not ending in atm.log.

I tried the approach of redirecting stdout to a homme log while in homme, but was bug prone. If you want to log something to homme_atm.log, then `write (iulog, *) blah` is the right way.
Even for atm.log, I tried to "disconnect" stdout/stderr from e3sm.log, but reattaching it was problematic. I played with some std/system calls, but eventually gave up.

ambrad:
That's fine. e3sm.log is perfectly fine with me. That's how EAM does things, too. Just wanted to make sure all those cases would end up somewhere.

bartgol:
Now that the Homme fixes have been merged in the upstream, and upstream merged in scream/master, this PR is ready for testing/review.

bartgol:
Looks like this PR is exposing some issues with the csm_share library and homme. Will have to wait until another upstream PR goes in for Homme.

AaronDonahue (approved):

PR #1855, opened 2022-08-10
Switch PEM test from ne30 to ne30pg2.

ambrad:
This fits within the allocated nodes on ascent (based on testing on summit).

I've also lengthened the test to 90 steps, which runs in ~200s. I think a longer
PEM test is useful.

### DIFF
--- a/cime_config/tests.py
+++ b/cime_config/tests.py
@@ -474,7 +474,7 @@ _TESTS = {
         "tests" : (
             #  "SMS_D_Ln2.ne30_ne30.F2000-SCREAMv1-AQP1", # Uncomment once IC file for ne30 is ready
             "ERS_Ln22.ne30_ne30.F2010-SCREAMv1",
-            "PEM_Ln9.ne30_ne30.F2010-SCREAMv1",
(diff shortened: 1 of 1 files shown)


ambrad:
This should get us some useful and passing PEM tests on ascent and pm-gpu.

ambrad:
For the reason I noted above, this new test also fails with out of memory on pm-gpu. Still, getting this going on ascent seems worthwhile.

ambrad:
@ndkeen, are you in charge of the pm-gpu PE layouts? Note my comment above that ne30pg2 gets only 1 node in testing, while ne30 gets 2. We'd really like twice that for PEM/ERP testing.

ndkeen:
Yes we can increase nodes used for this new grid. Should it be in this PR or in upstream e3sm?

ambrad:
I think it's better to make changes to the upstream repo when possible.

bartgol (approved):

ambrad:
I'm going to manually merge this because it would get this v1 CIME test on the dashboard tonight. The fails on weaver and blake were system issues. This is a one-line change, so I have reasonable confidence I haven't broken too much.

PR #1861, opened 2022-08-14
Change to use mosart and mpassi for F2010-SCREAMv1 and update the tests

wlin7:
This PR changes to use MOSART for river and replaces CICE with MPASSI
 for compsets F2010-SCREAMv1 and F2010-SCREAMv1-noAero. The tests
for F2010-SCREAMv1 lowres and medres are updated accordingly. The 
original compset is retained as F2010-SCREAMv1-CICE along with a test.

With mpassi for F2010-SCREAMv1, the grids used for the tests
must be replaced because mpassi can not run on atm se grid.
With mosart as river model for the compset and because of its
much larger time stepping, ERS_Ln# and ERP_Ln# tests would always
end up having mosart restart point being ahead of the stop time of
the restart run, which would fail to proceed. Changes to ERS_Ld3
and ERP_Ld3 could work, but it would take longer time. Furthermore
with _Ld3, REST_N=2days is being treated as 2 steps, leading to
much frequent writing of scream.r and mismatched atm restart time stamp
for ERP test's case2run.

To overcome these restart issue with the updated F2010-SCREAMv1, given
river model does not impact the atm results, a new test mod,
atmlndactive-rtm_off, is introduced to turn off the river model. This mode
is used for the updated ERS and ERP tests.

[non-BFB] for all tests using F2010-SCREAMv1 and F2010-SCREAMv1-noAero
                because of the grid changes and the use of the new testmod,
                the updated F2010-SCREAMv1 lowres and medres tests will become
                new tests and the original ones become missing.

### DIFF
--- a/cime_config/testmods_dirs/atmlndactive/rtm_off/README
+++ b/cime_config/testmods_dirs/atmlndactive/rtm_off/README
@@ -1,0 +1,9 @@
+Modifications (in shell_commands) to enable a hybrid run for SSP compsets
+(e.g., SSp370 or SSP585), starting from 2015-01-01 of v2.LR.historical_0101
+
+Other modifications (same as for wcprod) should result in a case that has the 
+same namelist settings as the water cycle production sims
+
+Run these for at least 1 day to see all output.
+Also use the CMIP6 compsets.
+If running longer, change the nhtfrq for the first history file.
--- a/cime_config/testmods_dirs/atmlndactive/rtm_off/README
+++ b/cime_config/testmods_dirs/atmlndactive/rtm_off/README
@@ -1,9 +1,4 @@
-Modifications (in shell_commands) to enable a hybrid run for SSP compsets
-(e.g., SSp370 or SSP585), starting from 2015-01-01 of v2.LR.historical_0101
-
-Other modifications (same as for wcprod) should result in a case that has the 
-same namelist settings as the water cycle production sims
-
-Run these for at least 1 day to see all output.
-Also use the CMIP6 compsets.
-If running longer, change the nhtfrq for the first history file.
+This test mod ican be used to disable river model mosart during runtime
+for compsets that use mosart as river model. In practice, this is used 
+for ERS and ERP tests with F2010-SCREAMv1 after it was changed to use 
+mosart in place of srof. 
--- a/cime_config/testmods_dirs/atmlndactive/rtm_off/README
+++ b/cime_config/testmods_dirs/atmlndactive/rtm_off/README
@@ -1,4 +1,5 @@
-This test mod ican be used to disable river model mosart during runtime
+This test mod can be used to disable river model mosart during runtime
 for compsets that use mosart as river model. In practice, this is used 
 for ERS and ERP tests with F2010-SCREAMv1 after it was changed to use 
 mosart in place of srof. 
+
--- a/cime_config/testmods_dirs/atmlndactive/rtm_off/user_nl_mosart
+++ b/cime_config/testmods_dirs/atmlndactive/rtm_off/user_nl_mosart
@@ -1,0 +1,6 @@
+! disable rtm to allow short ERS and ERP run to complete. 
+! If rtm active, due to large rtm time step, restart pointer for ROF
+! is always ahead of that for atm at the end of the first run. And in 
+! restart run, the restart point for ROF becomes later than the run stop time
+
+ do_rtm =.false.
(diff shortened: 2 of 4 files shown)


cime_config/testmods_dirs/atmlndactive/rtm_off/README
```diff
@@ -0,0 +1,4 @@
+This test mod ican be used to disable river model mosart during runtime
```
PeterCaldwell, line 1: I think "ican" is a typo, right Wuyin?

cime_config/tests.py
```diff
@@ -461,21 +461,21 @@
     "e3sm_scream_v1_lowres" : {
         "time"  : "01:00:00",
         "tests" : (
-            "ERP_D_Ln9.ne4_ne4.F2010-SCREAMv1",
+            "ERP_D_Ln9.ne4_oQU240.F2010-SCREAMv1.atmlndactive-rtm_off",
             "ERS_Ln9.ne4_ne4.F2000-SCREAMv1-AQP1",
```
PeterCaldwell, line 465: sanity check question: I think it would be good to keep at least one test using CICE (until CICE is completely removed from E3SM). That probably isn't accomplished very well from the aquaplanet ERS test above since that doesn't have any ice, right?

cime_config/tests.py
```diff
@@ -461,21 +461,21 @@
...
-            "SMS_D_Ln9.ne4_ne4.F2010-SCREAMv1-noAero",
-            "ERP_Ln22.ne4pg2_ne4pg2.F2010-SCREAMv1",
-            "ERS_D_Ln22.ne4pg2_ne4pg2.F2010-SCREAMv1",
+            "SMS_D_Ln9.ne4_oQU240.F2010-SCREAMv1-noAero",
```
PeterCaldwell, line 466: I think it is good to have at least one test that still exercises MOSART. I think the SMS noAero test above does that, right?

wlin7: Yes, SMS test will do that (as it is not using the rtm_off testmod).

cime_config/testmods_dirs/atmlndactive/rtm_off/README
wlin7: Good catch. I will fix it.

cime_config/tests.py
PeterCaldwell: That sounds good. We have ERP tests with Debug and non-debug mode, so let's just use CICE for one of them.

cime_config/testmods_dirs/atmlndactive/rtm_off/user_nl_mosart
```diff
@@ -0,0 +1,6 @@
+! disable rtm to allow short ERS and ERP run to complete. 
+! If rtm active, due to large rtm time step, restart pointer for ROF
+! is always ahead of that for atm at the end of the first run. And in 
+! restart run, the restart point for ROF becomes later than the run stop time
+
+ do_rtm =.false.
```
singhbalwinder, line 6: To get past this limitation (missing rpointer for rtm), I used to enforce RTM=NULL in env_build.xml. I think that approach doesn't even build RTM, so we may save some time there. I am not sure how `do_rtm=.false.` is implemented. If it is cleaner to use `do_rtm=.false`, we should stick with this way.

wlin7: Good point, Balwinder. Both would work for the sake of disabling mosart. It may still be useful to build rtm.

singhbalwinder (approved):

cime_config/testmods_dirs/atmlndactive/rtm_off/user_nl_mosart
wlin7: Balwinder, looks like RTM_MODE (taking value ACTIVE or NULL) is no more supported. It was used when we have component/rtm. I think mosart introduced this runtime option do_rtm to do the equivalent.

singhbalwinder: Ok, That is good to know. Thanks, @wlin7 !

PeterCaldwell (approved):

AaronDonahue:
I'm going to go ahead and manually merge this now.

AaronDonahue:
@wlin7 , It looks like there was a FAIL on mappy for the `ERP_Ln22.ne4pg2_oQU480.F2010-SCREAMv1.mappy_gnu9.atmlndactive-rtm_off` test.  Any reason to expect this to fail?

wlin7:
Hi @AaronDonahue , I am looking at the error log. There is a convergent error in mpassi. Look like it is a compiler/optimization specific issue. Only the tests on debug mode can run.  I am going to create an issue for this.

Issue #1732, opened 2022-06-10
radiation_tend aer_asm_bnd_sw above threshold with screamv0 ne30pg2_ne30pg2.F2010-SCREAM-LR and gnu 


ndkeen:
I was trying to run v0 on pm-cpu to get basic performance comparison with v1, but am running into errors with most PE layouts. I also tried on cori-knl -- with intel tests work, but with gnu I also see errors.

For example, this test:
`SMS_P512x1.ne30pg2_ne30pg2.F2010-SCREAM-HR.pm-cpu_gnu`

produced error:
```
  0:  ERROR:  WARNING: radiation_tend aer_asm_bnd_sw:            1  values above threshold ; max =    3.2653680558757496E+211
  0: #0  0x131e6ea in __shr_abort_mod_MOD_shr_abort_backtrace
  0:    at /global/cfs/cdirs/e3sm/ndk/se08-jun6/share/util/shr_abort_mod.F90:104
  0: #1  0x131e8bc in __shr_abort_mod_MOD_shr_abort_abort
  0:    at /global/cfs/cdirs/e3sm/ndk/se08-jun6/share/util/shr_abort_mod.F90:61
  0: #2  0x5fb189 in __radiation_utils_MOD_handle_error
  0:    at /global/cfs/cdirs/e3sm/ndk/se08-jun6/components/eam/src/physics/rrtmgp/radiation_utils.F90:563
  0: #3  0x5f52e8 in __radiation_MOD_radiation_tend
  0:    at /global/cfs/cdirs/e3sm/ndk/se08-jun6/components/eam/src/physics/rrtmgp/radiation.F90:1397
  0: #4  0x6db500 in tphysbc
  0:    at /global/cfs/cdirs/e3sm/ndk/se08-jun6/components/eam/src/physics/cam/physpkg.F90:2768
  0: #5  0x6e248d in __physpkg_MOD_phys_run1
  0:    at /global/cfs/cdirs/e3sm/ndk/se08-jun6/components/eam/src/physics/cam/physpkg.F90:1091
  0: #6  0x4b9b91 in __cam_comp_MOD_cam_run1
  0:    at /global/cfs/cdirs/e3sm/ndk/se08-jun6/components/eam/src/control/cam_comp.F90:258
  0: #7  0x4b4c44 in __atm_comp_mct_MOD_atm_init_mct
[...]

PR #1863, opened 2022-08-15
Pass in smaller extent (to ncol) for certain arrays to clip_values (affecting v0)

ndkeen:
In a few places, for calls to clip_values, change the extent of the array sent in for clipping.

This avoid problems running v0 cases with GNU and intel/debug when presumably trying to work with uninit values.

call handle_error(clip_values(aer_tau_bnd_sw(1:ncol,:,:)...

Fixes https://github.com/E3SM-Project/scream/issues/1732

[bfb]

### DIFF
--- a/components/eam/src/physics/rrtmgp/radiation.F90
+++ b/components/eam/src/physics/rrtmgp/radiation.F90
@@ -1389,9 +1389,9 @@ contains
                end if
 
                ! Check (and possibly clip) values before passing to RRTMGP driver
-               call handle_error(clip_values(cld_tau_gpt_sw,  0._r8, huge(cld_tau_gpt_sw), trim(subname) // ' cld_tau_gpt_sw', tolerance=1e-10_r8))
-               call handle_error(clip_values(cld_ssa_gpt_sw,  0._r8,                1._r8, trim(subname) // ' cld_ssa_gpt_sw', tolerance=1e-10_r8))
-               call handle_error(clip_values(cld_asm_gpt_sw, -1._r8,                1._r8, trim(subname) // ' cld_asm_gpt_sw', tolerance=1e-10_r8))
+               call handle_error(clip_values(cld_tau_gpt_sw(1:ncol,:,:),  0._r8, huge(cld_tau_gpt_sw), trim(subname) // ' cld_tau_gpt_sw', tolerance=1e-10_r8))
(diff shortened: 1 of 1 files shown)


components/eam/src/physics/rrtmgp/radiation.F90
```diff
@@ -1392,9 +1392,9 @@ subroutine radiation_tend(state,   ptend,    pbuf,          cam_out, cam_in,  &
                call handle_error(clip_values(cld_tau_gpt_sw,  0._r8, huge(cld_tau_gpt_sw), trim(subname) // ' cld_tau_gpt_sw', tolerance=1e-10_r8))
                call handle_error(clip_values(cld_ssa_gpt_sw,  0._r8,                1._r8, trim(subname) // ' cld_ssa_gpt_sw', tolerance=1e-10_r8))
                call handle_error(clip_values(cld_asm_gpt_sw, -1._r8,                1._r8, trim(subname) // ' cld_asm_gpt_sw', tolerance=1e-10_r8))
```
PeterCaldwell, line 1394: dumb question, but shouldn't we (1:ncol,:,:) all of these handle_error calls rather than just the aer_ ones? Or are the cld ones not 3d variables?

ndkeen: I only know that these were the line numbers of crashes -- ie, I changed one line, ran again, and got error at another location with `aer_*` variable clipping.

brhillman: Yeah, you’ll likely run into the same problem here.

components/eam/src/physics/rrtmgp/radiation.F90
```diff
@@ -1483,7 +1483,7 @@ subroutine radiation_tend(state,   ptend,    pbuf,          cam_out, cam_in,  &
 
                ! Check (and possibly clip) values before passing to RRTMGP driver
                call handle_error(clip_values(cld_tau_gpt_lw,  0._r8, huge(cld_tau_gpt_lw), trim(subname) // ': cld_tau_gpt_lw', tolerance=1e-10_r8))
```
brhillman, line 1485: We should also add `(1:ncol,:,:)` here.

brhillman (changes requested):
This looks great. While you’re doing this though, could you also please add `(1:ncol,:,:)` to the cloud optics in the same places?

ndkeen:
I pushed a commit -- is that what we want?

brhillman (approved):
This looks great, thanks Noel!

PR #1864, opened 2022-08-15
Increased postcondition tolerances to avoid nuisance crashes

PeterCaldwell:
In our INCITE push we hit errors where PS was slightly above our postcondition check max allowable value of 1100mb. This morning we hit nr>1e9 #/kg even though the model seems to run fine with larger values. In both cases these max allowable values were based on what we expect to see on Earth, but there's no real reason why larger values couldn't occur. Thus I'm just increasing the max allowable values in this PR. Perhaps with better tuning we can avoid having these un-Earthlike values, but we should be allowed to get the model running before making it perfectly tuned. It's possible that we should just remove upper bounds for these variables completely... I'm opening this PR as a conversation starter.


ambrad (approved):

ambrad:
Nothing controversial here, so I approved the PR so it can automerge.

PR #1865, opened 2022-08-16
EAMxx: make radiation process columns in chunks

bartgol:
Allows to reduce (YAKL) memory footprint in RRTMGP. The size of column chunks is configurable via yaml/xml file.

Passes standalone testing on my laptop. I will let the AT run v1 testing (or do it myself in the morning if AT times out on mappy, as it often happens at night). I will also run memcheck tests in the morning.

### DIFF
--- a/components/scream/src/physics/rrtmgp/atmosphere_radiation.hpp
+++ b/components/scream/src/physics/rrtmgp/atmosphere_radiation.hpp
@@ -53,6 +53,9 @@ public:
 
   // Keep track of number of columns and levels
   int m_ncol;
+  int m_num_col_chunks;
+  int m_col_chunk_size;
+  std::vector<int> m_col_chunk_beg;
   int m_nlay;
   view_1d_real m_lat;
   view_1d_real m_lon;
@@ -84,7 +87,7 @@ public:
   int m_ngas;
   std::vector<ci_string>   m_gas_names;
   view_1d_real             m_gas_mol_weights;
-  GasConcs gas_concs;
+  std::vector<GasConcs> gas_concs;
 
   // Rad frequency in number of steps
   int m_rad_freq_in_steps;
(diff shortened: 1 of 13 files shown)


ambrad:
Will the standalone test verify the correctness of the chunked vs non-chunked cases? If not, you could consider creating a testmod for the PEM ne30pg2 case that atmchanges the chunk parameter such that with 6 GPUs, there are two chunks, and with 12, just one. That would catch any errors where `beg` should be used but isn't or the opposite. Even making both runs of the PEM test use multiple chunks, without worrying about making one run use just one chunk, would be a good test, since the `beg` values will be different for a given column in the two runs.

bartgol:
That's a good idea. I'll create a standalone rad-only test that compares chunked-vs-nonchunked. It's probably the fastest way to check this.

ambrad (approved):
This is BFB, so I'm going to review and approve and take Ben off. This works for the ne1024pg2 1000-node run on Summit, so let's bring it in to master. We can always iterate on it later.

components/scream/src/physics/rrtmgp/atmosphere_radiation.hpp
```diff
@@ -84,7 +87,7 @@ class RRTMGPRadiation : public AtmosphereProcess {
   int m_ngas;
   std::vector<ci_string>   m_gas_names;
   view_1d_real             m_gas_mol_weights;
-  GasConcs gas_concs;
+  std::vector<GasConcs> gas_concs;
```
brhillman, line 90: Do all of the chunks have the same size? If so, and the loop over chunks is serial, then why not just init a single `gas_concs` object with size of each chunk?

bartgol: Good point. Working on this improvement.

bartgol:
@ambrad If you think it's ok to merge as is, I can make the improvement on the gas concs array a separate PR.

ambrad:
@bartgol i'm fine with merging and then improving the GasConcs handling in a separte PR.

PR #1867, opened 2022-08-16
Fixes to random single-prec test fails in shoc and p3

### LINKED ISSUE #1452: A test for p3 main consistently fails on blake
For the single precision build, with the proper random number generator seed, we can consistently reproduce a fail on blake.

To reproduce:

-  run `scripts/test-all-scream -m blake -t sp --baseline-dir AUTO --config-only`
-  run `eval $(scripts/scream-env-cmd blake)`
- `cd ctest-build/full_sp_debug/src/physics/p3/tests`
- `make -j64`
- `ctest -R setup` (to pre-build p3 tables)
- ./p3_tests --rng-seed 4126930317`

You should see an error of the form

```
/ascldap/users/e3sm-jenkins/blake/workspace/SCREAM_NightlyTests_Blake/scream/components/scream/src/physics/p3/tests/p3_main_unit_tests.cpp:218: FAILED:
  REQUIRE( isds_fortran[i].qi[k] == isds_cxx[i].qi[k] )
with expansion:
  0.0f == 0.0f
```

This might just be an unlucky input, but the fact that it produces actual number (rather than NaN) that are non-bfb between f90 and cxx makes me a bit suspicious, that some operation in p3 uses something like `1.0` instead of `sp(1.0)` to create a floating point constant.

### LINKED ISSUE #1506: List of unit tests that fail with specific random seeds

jgfouca:
Fixes #1506 
Fixes #1452

### DIFF
--- a/components/scream/src/physics/p3/p3_ice_collection_impl.hpp
+++ b/components/scream/src/physics/p3/p3_ice_collection_impl.hpp
@@ -21,9 +21,9 @@ void Functions<S,D>
   constexpr Scalar tmelt  = C::Tmelt;
 
   // set up masks
-  const auto t_is_negative        = temp < tmelt;
-  const auto qi_incld_gt_small = qi_incld > qsmall;
-  const auto qc_incld_gt_small    = qc_incld > qsmall;
+  const auto t_is_negative        = temp <= tmelt;
+  const auto qi_incld_gt_small    = qi_incld >= qsmall;
+  const auto qc_incld_gt_small    = qc_incld >= qsmall;
   const auto both_gt_small        = qi_incld_gt_small && qc_incld_gt_small && context;
   const auto both_gt_small_pos_t  = both_gt_small && !t_is_negative;
 
--- a/components/scream/src/physics/p3/p3_ice_collection_impl.hpp
+++ b/components/scream/src/physics/p3/p3_ice_collection_impl.hpp
@@ -22,24 +22,24 @@ void Functions<S,D>
 
   // set up masks
   const auto t_is_negative        = temp <= tmelt;
-  const auto qi_incld_gt_small    = qi_incld >= qsmall;
-  const auto qc_incld_gt_small    = qc_incld >= qsmall;
-  const auto both_gt_small        = qi_incld_gt_small && qc_incld_gt_small && context;
-  const auto both_gt_small_pos_t  = both_gt_small && !t_is_negative;
+  const auto qi_incld_ge_small    = qi_incld >= qsmall;
+  const auto qc_incld_ge_small    = qc_incld >= qsmall;
+  const auto both_ge_small        = qi_incld_ge_small && qc_incld_ge_small && context;
+  const auto both_ge_small_pos_t  = both_ge_small && !t_is_negative;
 
   constexpr auto eci = C::eci;
   constexpr auto inv_dropmass = C::ONE/C::dropmass;
 
-  qc2qi_collect_tend.set(both_gt_small && t_is_negative,
+  qc2qi_collect_tend.set(both_ge_small && t_is_negative,
             rhofaci*table_val_qc2qi_collect*qc_incld*eci*rho*ni_incld);
-  nc_collect_tend.set(both_gt_small, rhofaci*table_val_qc2qi_collect*nc_incld*eci*rho*ni_incld);
+  nc_collect_tend.set(both_ge_small, rhofaci*table_val_qc2qi_collect*nc_incld*eci*rho*ni_incld);
 
   // for T_atm > 273.15, assume cloud water is collected and shed as rain drops
   // sink for cloud water mass and number, note qcshed is source for rain mass
-  qc2qr_ice_shed_tend.set(both_gt_small_pos_t, rhofaci*table_val_qc2qi_collect*qc_incld*eci*rho*ni_incld);
-  nc_collect_tend.set(both_gt_small_pos_t, rhofaci*table_val_qc2qi_collect*nc_incld*eci*rho*ni_incld);
+  qc2qr_ice_shed_tend.set(both_ge_small_pos_t, rhofaci*table_val_qc2qi_collect*qc_incld*eci*rho*ni_incld);
+  nc_collect_tend.set(both_ge_small_pos_t, rhofaci*table_val_qc2qi_collect*nc_incld*eci*rho*ni_incld);
   // source for rain number, assume 1 mm drops are shed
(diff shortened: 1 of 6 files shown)


components/scream/src/physics/shoc/tests/shoc_update_prognostics_implicit_tests.cpp
```diff
@@ -338,10 +338,10 @@ struct UnitWrap::UnitTest<D>::TestUpdatePrognosticsImplicit {
     auto engine = setup_random_test();
 
     UpdatePrognosticsImplicitData f90_data[] = {
-      UpdatePrognosticsImplicitData(10, 71, 72, 19, 5),
-      UpdatePrognosticsImplicitData(10, 12, 13, 7, 2.5),
-      UpdatePrognosticsImplicitData(7, 16, 17, 2, 1),
-      UpdatePrognosticsImplicitData(2, 7, 8, 1, 1)
+      UpdatePrognosticsImplicitData(10, 71, 72, 19, .5),
```
jgfouca, line 341: Making the timestep smaller reduces the chance of an FPE in the triangular solve. Thanks @ambrad !

ambrad (approved):

components/scream/src/physics/p3/p3_ice_collection_impl.hpp
```diff
@@ -21,9 +21,9 @@ ::ice_cldliq_collection(
   constexpr Scalar tmelt  = C::Tmelt;
 
   // set up masks
-  const auto t_is_negative        = temp < tmelt;
-  const auto qi_incld_gt_small = qi_incld > qsmall;
-  const auto qc_incld_gt_small    = qc_incld > qsmall;
+  const auto t_is_negative        = temp <= tmelt;
```
jgfouca, line 24: If you look at the fortran, it's doing `.ge.` and `.le.` , so this port was not exactly true to the reference. I'm not sure which implementation is better, but I went with the fortran.

jgfouca (commented):
Annotations.

components/scream/src/physics/p3/p3_ice_collection_impl.hpp
ambrad: Yes, we should follow the Fortran or otherwise correct it. In this case, let's follow it.

jgfouca:
As a side note, finding the non-bfb error in ice_collection was quite the chore without the ability to easily redirect the subroutines in micro_p3.F90 to call the cxx implementation. We had this ability a couple years ago but we decided to remove it. That saved thousands of lines of tedious bridging code, but it would have made my task today a lot easier. Since we still have the Cxx->F90 bridges, which are necessary for bfb unit testing, I was able to hack the p3 cxx functions, one-by-one until I was bfb with f90, to call f90 without too much difficulty.

components/scream/src/physics/p3/p3_ice_collection_impl.hpp
ambrad: One tiny cosmetic thing is you could change `_gt_` on the LHS to `_ge_`, following the C++ naming convention elsewhere, e.g., in the routine following this one.

ambrad:
@jgfouca now that you mentioned this > vs >= issue, I'm seeing it elsewhere, e.g. qi_incld in ice_self_collection.

jgfouca:
@ambrad , yikes, I wonder if we need to do a quick audit of many of the converted functions. An automated audit would be nice but I can't think of how to do it.

ambrad:
Probably wise. You can make it an issue rather than take it on in thsi PR, though.

Might use something like this to make the search somewhat organized:
```
grep "subroutine\|\.gt\.\|\.ge\.\|\.lt\.\|\.le\." micro_p3.F90
```

components/scream/src/physics/p3/p3_ice_sed_impl.hpp
```diff
@@ -240,27 +240,27 @@ ::homogeneous_freezing(
     const auto range_pack    = ekat::range<IntSmallPack>(pk*Spack::n);
     const auto range_mask    = range_pack >= kmin_scalar && range_pack <= kmax_scalar;
     const auto t_lt_homogf   = T_atm(pk) < T_homogfrz;
-    const auto qc_gt_small   = range_mask && t_lt_homogf && qc(pk) > qsmall;
-    const auto qr_gt_small   = range_mask && t_lt_homogf && qr(pk) > qsmall;
+    const auto qc_ge_small   = range_mask && t_lt_homogf && qc(pk) >= qsmall;
```
jgfouca, line 243: Found more mistakes here.

components/scream/src/physics/p3/p3_prevent_liq_supersaturation_impl.hpp
```diff
@@ -27,7 +27,7 @@ void Functions<S,D>::prevent_liq_supersaturation(const Spack& pres, const Spack&
   Spack qv_sinks, qv_sources, qv_endstep, T_endstep, A, frac;
 
   qv_sources.set(context, qi2qv_sublim_tend + qr2qv_evap_tend);
-  const auto has_sources = (qv_sources>qsmall && context); //if nothing to rescale, no point in calculations.
+  const auto has_sources = (qv_sources>=qsmall && context); //if nothing to rescale, no point in calculations.
```
jgfouca, line 30: Another mistake here.

components/scream/src/physics/p3/p3_ice_collection_impl.hpp
```diff
@@ -99,7 +99,7 @@ ::ice_self_collection(
 
   // Set up masks
   const auto qm_incld_positive = qm_incld > zero && context;
-  const auto qi_incld_gt_small = qi_incld > qsmall && context;
+  const auto qi_incld_ge_small = qi_incld >= qsmall && context;
```
jgfouca, line 102: Another mistake here. @ambrad also noticed this.

jgfouca (commented):
Audit complete with a couple more mistakes found.

bartgol (approved):

ambrad (approved):

PR #1871, opened 2022-08-17
Add rayleigh friction to dynamics AD

tcclevenger:
Adds Rayleigh friction computation from `scream/components/eam/src/physics/cam/rayleigh_friction.F90` to `homme_post_process()` in `atmosphere_dynamics` atm process.

### DIFF
--- a/components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
+++ b/components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
@@ -481,7 +481,7 @@ void HommeDynamics::run_impl (const int dt)
 
     // Post process Homme's output, to produce what the rest of Atm expects
     Kokkos::fence();
-    homme_post_process ();
+    homme_post_process (dt);
   } catch (std::exception& e) {
     EKAT_ERROR_MSG(e.what());
   } catch (...) {
@@ -656,7 +656,7 @@ void HommeDynamics::homme_pre_process (const int dt) {
   }
 }
 
-void HommeDynamics::homme_post_process () {
+void HommeDynamics::homme_post_process (const int dt) {
   const auto& pgn = m_phys_grid->name();
   const auto& c = Homme::Context::singleton();
   const auto& params = c.get<Homme::SimulationParams>();
@@ -686,6 +686,7 @@ void HommeDynamics::homme_post_process () {
   using Pack = ekat::Pack<Real,N>;
   using ColOps = ColumnOps<DefaultDevice,Real>;
   using PF = PhysicsFunctions<DefaultDevice>;
+  using WS = ekat::WorkspaceManager<Pack,DefaultDevice>;
 
   if (fv_phys_active()) return;
   
@@ -715,6 +716,17 @@ void HommeDynamics::homme_post_process () {
   const auto& hvcoord = c.get<Homme::HybridVCoord>();
   const auto ps0 = hvcoord.ps0 * hvcoord.hybrid_ai0;
 
+  // Rayleigh friction paramaters
+
+  // Vertical level at which rayleigh friction term is centered.
+  const int rayk0 = m_params.get<int>("Rayleigh Friction Vertical Level", 2);
+  // Range of rayleigh friction profile.
+  const Real raykrange = m_params.get<Real>("Rayleigh Friction Range", 0.0);
+  // Approximate value of decay time at model top (days)
+  // if set to 0, no rayleigh friction is applied
+  const Real raytau0 = m_params.get<Real>("Rayleigh Friction Decay Time", 5.0);
+
+  WS wsm(npacks,1,policy);
   Kokkos::parallel_for(policy, KOKKOS_LAMBDA (const KT::MemberType& team) {
     const int& icol = team.league_rank();
 
@@ -760,6 +772,39 @@ void HommeDynamics::homme_post_process () {
       // Store T at end of the dyn timestep (to back out tendencies later)
       T_prev(ilev) = T_val;
     });
+
+    // Calculate decay rate profile, otau,
+    // and apply Rayleigh friction
+    Real krange; // range of rayleigh friction profile
+    Real tau0;   // approximate value of decay time at model top
+    Real otau0;  // inverse of tau0
+
+    krange = raykrange;
+    if (raykrange == 0) krange = (rayk0 - 1.0)/2.0;
+
+    tau0 = 86400.0*raytau0; // convert to seconds
+    otau0 = 0;
+    if (tau0 != 0) otau0 = 1/tau0;
+
+    if (otau0 != 0) {
+      auto ws = wsm.get_workspace(team);
+      ekat::Unmanaged<WS::view_1d<Pack> > otau = ws.take("otau");
+
+      Kokkos::parallel_for(Kokkos::TeamThreadRange(team, npacks),
+                           [&](const int ilev) {
+        const auto range_pack = ekat::range<Pack>(ilev*Pack::n);
+        const Pack x = (rayk0 - range_pack)/krange;
+        otau(ilev) = otau0*(1 + ekat::tanh(x))/2;
+      });
+
+      // Apply Rayleigh friction.
+      auto u_wind = ekat::subview(v_view, icol, 0);
+      auto v_wind = ekat::subview(v_view, icol, 1);
+      auto T_mid = ekat::subview(T_view, icol);
+      PF::apply_rayleigh_friction(team, npacks, dt, otau, u_wind, v_wind, T_mid);
+
+      ws.release(otau);
+    }
   });
 
   // If ftype==FORCING_2, also set FQ_phys=Q_phys. Next step's Q_dyn will be set
--- a/components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
+++ b/components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
@@ -788,10 +788,10 @@ void HommeDynamics::homme_post_process (const int dt) {
       // Store T at end of the dyn timestep (to back out tendencies later)
       T_prev(ilev) = T_val;
     });
-
-    // Apply Rayleigh friction to update temperature and horiz_winds
-    rayleigh_friction_apply(dt);
   });
+
+  // Apply Rayleigh friction to update temperature and horiz_winds
+  rayleigh_friction_apply(dt);
 
   // If ftype==FORCING_2, also set FQ_phys=Q_phys. Next step's Q_dyn will be set
   // as Q_dyn = Q_dyn_old + PD_remap(Q_phys-Q_phys_old)
--- a/components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
+++ b/components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
@@ -247,10 +247,8 @@ void HommeDynamics::set_grids (const std::shared_ptr<const GridsManager> grids_m
 
 size_t HommeDynamics::requested_buffer_size_in_bytes() const
 {
-  using Pack = ekat::Pack<Real,HOMMEXX_PACK_SIZE>;
   const int nlevs = m_dyn_grid->get_num_vertical_levels();
-  const int npacks = ekat::PackInfo<HOMMEXX_PACK_SIZE>::num_packs(nlevs);
-  const size_t interface_request = Buffer::num_1d_scalar_nlev*npacks*sizeof(Pack);
+  const size_t interface_request = Buffer::num_1d_scalar_nlev*nlevs*sizeof(Real);
 
   using namespace Homme;
 
@@ -303,16 +301,10 @@ void HommeDynamics::init_buffers(const ATMBufferManager &buffer_manager)
                    "Error! Buffers size not sufficient.\n");
 
   // Local memory used by interface
-  using Pack = ekat::Pack<Real,HOMMEXX_PACK_SIZE>;
   const int nlevs = m_dyn_grid->get_num_vertical_levels();
-  const int npacks = ekat::PackInfo<HOMMEXX_PACK_SIZE>::num_packs(nlevs);
-  Pack* packed_mem = reinterpret_cast<Pack*>(buffer_manager.get_memory());
-
-  m_buffer.otau = decltype(m_buffer.otau)(packed_mem, npacks);
-  packed_mem += m_buffer.otau.size();
-  std::cout << "OTAU SIZE: " << m_buffer.otau.size() << std::endl;
-
-  Real* mem = reinterpret_cast<Real*>(packed_mem);
+  Real* mem = reinterpret_cast<Real*>(buffer_manager.get_memory());
+  m_buffer.otau = decltype(m_buffer.otau)(mem, nlevs);
+  mem += m_buffer.otau.size();
 
   // Memory used by homme buffer
   using namespace Homme;
--- a/components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
+++ b/components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
@@ -475,38 +475,8 @@ void HommeDynamics::initialize_impl (const RunType run_type)
   add_postcondition_check<Interval>(get_field_out("horiz_winds",pgn),m_phys_grid,-400.0, 400.0,false);
   add_postcondition_check<Interval>(get_field_out("ps"),m_phys_grid,40000.0, 120000.0,false);
 
-  constexpr int N = HOMMEXX_PACK_SIZE;
-  using KT = KokkosTypes<DefaultDevice>;
-  using Pack = ekat::Pack<Real,N>;
-
-  // Rayleigh friction paramaters
-  m_rayk0     = m_params.get<int>("Rayleigh Friction Vertical Level", 2);
-  m_raykrange = m_params.get<Real>("Rayleigh Friction Range", 0.0);
-  m_raytau0   = m_params.get<Real>("Rayleigh Friction Decay Time", 5.0);
-
-  // Calculate decay rate profile, otau.
-  // If m_raytau0 == 0 no Rayleigh friction is applied.
-  if (m_raytau0 != 0) {
-    Real krange; // range of rayleigh friction profile
-    Real tau0;   // approximate value of decay time at model top
-    Real otau0;  // inverse of tau0
-
-    krange = m_raykrange;
-    if (m_raykrange == 0) krange = (m_rayk0 - 1.0)/2.0;
-
-    tau0 = 86400.0*m_raytau0; // convert to seconds
-    otau0 = 0;
-    if (tau0 != 0) otau0 = 1.0/tau0;
-
-    auto otau = m_buffer.otau;
-    const int npacks = ekat::PackInfo<N>::num_packs(nlevs);
-    Kokkos::parallel_for(KT::RangePolicy(0, npacks),
-                         KOKKOS_LAMBDA (const int ilev) {
-      const auto range_pack = ekat::range<Pack>(ilev*Pack::n+1);
-      const Pack x = (m_rayk0 - range_pack)/krange;
-      otau(ilev) = otau0*(1.0 + ekat::tanh(x))/2.0;
-    });
-  }
+  // Initialize Rayleigh friction variables
+  rayleigh_friction_init();
 }
 
 void HommeDynamics::run_impl (const int dt)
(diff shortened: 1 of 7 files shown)


tcclevenger:
Still need to add a test.

ambrad:
My personal preference is to encapsulate things a little more rather than keep adding to an existing subroutine. (See the ..._fv_phys.cpp file for an example of what I mean.) I recommend moving the dynamics interface Rayleigh friction code to one or more separate subroutines. Also grab the settings just once from the PL and save them as object data.

components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
```diff
@@ -760,6 +772,39 @@ void HommeDynamics::homme_post_process () {
...
+
+    tau0 = 86400.0*raytau0; // convert to seconds
+    otau0 = 0;
+    if (tau0 != 0) otau0 = 1/tau0;
+
+    if (otau0 != 0) {
+      auto ws = wsm.get_workspace(team);
+      ekat::Unmanaged<WS::view_1d<Pack> > otau = ws.take("otau");
+
+      Kokkos::parallel_for(Kokkos::TeamThreadRange(team, npacks),
```
ambrad, line 793: One rank could store a single 1D view init'ed once at init time.

ambrad: Alternatively, since this is our impl, the RF impl could just compute otau during its pass over the data and no extra view input would be needed.

ambrad: Separately, please make sure to verify your otau profile matches the EAM one.

components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
```diff
@@ -710,6 +711,17 @@ void HommeDynamics::homme_post_process () {
...
+
+  // Vertical level at which rayleigh friction term is centered.
+  const int rayk0 = m_params.get<int>("Rayleigh Friction Vertical Level", 2);
+  // Range of rayleigh friction profile.
+  const Real raykrange = m_params.get<Real>("Rayleigh Friction Range", 0.0);
+  // Approximate value of decay time at model top (days)
+  // if set to 0, no rayleigh friction is applied
+  const Real raytau0 = m_params.get<Real>("Rayleigh Friction Decay Time", 5.0);
+
+  WS wsm(npacks,1,policy);
```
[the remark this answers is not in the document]
tcclevenger: Yeah, it does it in one other place in this file with `p_int` and `p_mid`. I can't remember why this couldn't be moved to the buffer.. Maybe I'll try and move all of it there.

components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
tcclevenger: Yes, I verified the profiles matched. I may just change the `PF::apply_rayleigh_friction()` to compute the `otau` vals.

tcclevenger: Actually, do you have a preference: 1d view in the mem buffer vs. internally computing profile? Adding to the mem buffer closer resembles eam.

ambrad: Either is fine, but if you keep it as a view, I recommend a single 1D view per rank computed at init, since otau is independent of dynamic and spatially varying quantities.

tcclevenger:
The compile error is 
```
home/e3sm-jenkins/weaver/workspace/SCREAM_PullRequest_Autotester_Weaver/3639/
scream/externals/YAKL/cub/cub/block/block_load.cuh(174): error: expected an identifier
```
That line looks like 
```
enum
    {
        TOTAL_WORDS = sizeof(items) / sizeof(DeviceWord),
 ->     VECTOR_SIZE = (TOTAL_WORDS % 4 == 0) ?
            4 :
            (TOTAL_WORDS % 2 == 0) ?
                2 :
                1,
        VECTORS_PER_THREAD = TOTAL_WORDS / VECTOR_SIZE,
    };
```
I've reproduced it on summit as well as weaver. The error only happens when compiling `tests/coupled/dynamics_physics/homme_shoc_cld_spa_p3_rrtmgp/`
`tests/coupled/dynamics_physics/homme_shoc_cld_spa_p3_rrtmgp_128levels/`
but does not happen with any other test (even `homme_shoc_cld_p3_rrtmgp/`)

ambrad:
@bartgol is the above maybe a cpp symbol collision on VECTOR_SIZE?

tcclevenger:
> @bartgol is the above maybe a cpp symbol collision on VECTOR_SIZE?

I bet that is exactly what is happening. I include `homme_dimensions.hpp` in the header `atmosphere_dynamics.hpp`, and in `homme_dimensions.hpp` there is the line
```
#ifdef CUDA_BUILD
...
#define HOMMEXX_PACK_SIZE VECTOR_SIZE
...
#else
...
#endif
```

ambrad:
Great. We should definitely avoid including Homme internal headers in the AD header files when possible.

tcclevenger:
The issue is that the `Buffer` struct that lives in `atmosphere_dynamics.hpp` defines the local view `otau` which I had as a packed view. The issue was that I needed the homme packsize which required me to include `homme_dimensions.hpp` which caused the conflicting definitions of VECTOR_SIZE. 

My quickfix solution was to just make `otau` a view of `Real` instead of `Pack`. This isn't ideal since we are now not using packs to apply rayleigh friction which is called every `run_impl()` call, but it's also a small chunk of code, so I doubt it'd be noticeable. 

@ambrad Let me know what you think. I've manually verified that we can compile.

tcclevenger:
Also, this is interface code, should I add the CIME label? There will be diffs, but we could make sure no crashes.

bartgol:
> The issue is that the `Buffer` struct that lives in `atmosphere_dynamics.hpp` defines the local view `otau` which I had as a packed view. The issue was that I needed the homme packsize which required me to include `homme_dimensions.hpp` which caused the conflicting definitions of VECTOR_SIZE.

When we configure Homme, we set homme's VECTOR_SIZE equal to SCREAM_PACK_SIZE. So inside our interface you can safely rely on those to be the same.

bartgol:
> Also, this is interface code, should I add the CIME label? There will be diffs, but we could make sure no crashes.

The CIME label is "old". We introduced it when testing CIME cases was "rare", and the label signaled to run v0 tests. It's basically pointless now, and should be kept for stuff that has something to do with cime (e.g., modifying our cime_config folder).

tcclevenger:
Looks like a CIME test actually fails. I need to investigate

tcclevenger:
I ran on summit, and the failing CIME test was only a DIFF there. Mappy seems to not be allowing user login, so I'll just retest here. Not sure why mappy would fail if summit is ok..

tcclevenger:
Looks like the baseline dir doesn't exist on mappy
```
FAIL ERP_D_Ln9.ne4_ne4.F2010-SCREAMv1.mappy_gnu9 COMPARE_base_rest
FAIL ERP_D_Ln9.ne4_ne4.F2010-SCREAMv1.mappy_gnu9 BASELINE master: ERROR BFAIL baseline directory '/sems-data-store/ACME/baselines/mappy/gnu/master/ERP_D_Ln9.ne4_ne4.F2010-SCREAMv1.mappy_gnu9' does not exist
```

bartgol:
> '/sems-data-store/ACME/baselines/mappy/gnu/master/ERP_D_Ln9.ne4_ne4.F2010-SCREAMv1.mappy_gnu9'

That looks like the wrong compiler in the folder name. it should be `[..]/mappy/gnu9/master` rather than `[..]/mappy/gnu/master`.

tcclevenger:
I just ran using "-c -b master" which I think is what the test scripts use?

bartgol:
> I just ran using "-c -b master" which I think is what the test scripts use?
> 
> Also it looks like the AT can't access mappy either?

Since v1 uses a compiler which is not the "default" one for mappy, you need to specify it, via `--compiler gnu9`.

bartgol:
As for AT accessing mappy, it looks like at the time the job started, mappy was busy with the `mappy_cime` job (nightly testing for cime itself), and eventually AT timed out. Other than that, AT is doing fine on mappy.

tcclevenger:
Ok, checked the case dir and the "FAIL" listed is in AT console output is actually just a "DIFF". Not sure why it didn't report "DIFF"? Either way, looked at the diff files and  everything is as expected.

@ambrad @bartgol Should be ok to merge, let me know if there is anything else.

ambrad:
@AaronDonahue I'm fine with merging this when you're ready for another non-BFB PR unless @bartgol objects.

components/scream/src/dynamics/homme/atmosphere_dynamics.cpp
```diff
@@ -287,22 +293,42 @@ size_t HommeDynamics::requested_buffer_size_in_bytes() const
...
                    "Error! Buffers size not sufficient.\n");
 
+  // Local memory used by interface
+  constexpr int N = HOMMEXX_PACK_SIZE;
+  using Pack = ekat::Pack<Real,N>;
+  const int nlevs = m_dyn_grid->get_num_vertical_levels();
+  const int npacks = ekat::PackInfo<N>::num_packs(nlevs);
+  Pack* packed_mem = reinterpret_cast<Pack*>(buffer_manager.get_memory());
+
+  m_buffer.otau = decltype(m_buffer.otau)(packed_mem, npacks);
```
ambrad, line 313: Do these buffers persist beyond one run_impl call? I thought they are supposed to hold only temporary data.

ambrad: @bartgol can weigh in, but in any case, I think a view that is written once at startup and then read repeatedly during time stepping should be kept as a simple View object in the HommeDynamics object.

tcclevenger: Wait, you are right. This is overwritten. I'll move it.

components/scream/src/dynamics/homme/atmosphere_dynamics_rayleigh_friction.cpp
```diff
@@ -14,8 +14,12 @@ void HommeDynamics::rayleigh_friction_init()
   using KT = KokkosTypes<DefaultDevice>;
   using Pack = ekat::Pack<Real,N>;
 
+  const int nlevs = m_dyn_grid->get_num_vertical_levels();
+  const auto npacks= ekat::PackInfo<Pack::n>::num_packs(nlevs);
+  m_otau = decltype(m_otau)("otau", npacks);
+
   // Rayleigh friction paramaters
-  m_rayk0     = m_params.get<int>("Rayleigh Friction Vertical Level", 2);
+  m_rayk0     = m_params.get<int>("Rayleigh Friction Vertical Level", 1);
```
tcclevenger, line 22: @ambrad I went ahead and converted this to c++ indexing (set this `[0,numlevs-1)`). Let me know if you would like it reverted back. I verified that the profile is the same as before (which matches what you posted in scream slack from F90)

tcclevenger:
Thank for finding that @ambrad. Now it is a class var initialized and computed in `atmosphere_dynamics_rayleigh_friction.cpp`.

components/scream/src/dynamics/homme/atmosphere_dynamics_rayleigh_friction.cpp
ambrad: That's a matter of taste. @AaronDonahue @PeterCaldwell do we use 0-based or 1-based indexing in user-facing parameters?

Separately, we're no longer using PL entries of the form "Rayleigh Friction Vertical Level"; switch to something like Rayleigh_friction_vertical_level, where I purpose capitalize Rayleigh since it's a name and lowercase everything else.

components/scream/src/dynamics/homme/atmosphere_dynamics_rayleigh_friction.cpp
```diff
@@ -0,0 +1,83 @@
...
+  using Pack = ekat::Pack<Real,N>;
+
+  const int nlevs = m_dyn_grid->get_num_vertical_levels();
+  const auto npacks= ekat::PackInfo<Pack::n>::num_packs(nlevs);
+  m_otau = decltype(m_otau)("otau", npacks);
+
+  // Rayleigh friction paramaters
+  m_rayk0     = m_params.get<int>("Rayleigh Friction Vertical Level", 1);
+  m_raykrange = m_params.get<Real>("Rayleigh Friction Range", 0.0);
+  m_raytau0   = m_params.get<Real>("Rayleigh Friction Decay Time", 5.0);
```
ambrad, line 24: Same for these re: removing unnecessary caps and all spaces.

components/scream/src/dynamics/homme/atmosphere_dynamics_rayleigh_friction.cpp
bartgol: Since we went with snake_case, I would keep lowercase also for names, acronyms, and whatnots. But again, taste.

components/scream/src/dynamics/homme/atmosphere_dynamics_rayleigh_friction.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+  const auto horiz_winds_view = get_field_out("horiz_winds").get_view<Pack***>();
+  const auto T_mid_view       = get_field_out("T_mid").get_view<Pack**>();
+
+  const auto policy = ESU::get_default_team_policy(ncols, npacks);
+  Kokkos::parallel_for(policy, KOKKOS_LAMBDA (const KT::MemberType& team) {
+    const int& icol = team.league_rank();
+
+    // If m_raytau0 == 0 no Rayleigh friction is applied.
+    if (m_raytau0 != 0) {
```
ambrad, line 73: m_raytau0 is a class var; this can crash on the GPU. Copy to a local for lambda capture.

components/scream/src/dynamics/homme/atmosphere_dynamics_rayleigh_friction.cpp
```diff
@@ -0,0 +1,83 @@
...
+      const auto otau = m_otau;
```
ambrad, line 74: Ditto re: lambda capture.

components/scream/src/dynamics/homme/atmosphere_dynamics_rayleigh_friction.cpp
AaronDonahue: uggg, thats a tricky one.  I'm inclined to say we want to stick with 1-based since most users will be familiar with that.  Although the use of the variable inside C++ should be 0-based to avoid confusion when developing or just plain reading the code.  If we stick with outward 1-based can we just subtract 1 from what we read from the parameter list?  also would be good to put a comment into the code explaining what is going on here.  Also a comment in the default parameter list...

ambrad: @bartgol I'd like to disagree re: case. Snake case (vs. camel case) is the best convention b/c you can capitalize what should be cap'ed and nothing else.

bartgol: Fair point.

ambrad: That said, I'm realizing this is potentially going to cause user experience issues, so all lowercase is fine. We can use a capital letter if a situation arises where actual ambiguity can occur due to constant-naming conventions.

tcclevenger: RE: 0- vs 1- based, I changed the param to be 1-based, then immediately convert to 0-based with a comment. Let me know what y'all think.

components/scream/src/dynamics/homme/atmosphere_dynamics_rayleigh_friction.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+  const auto ncols = m_phys_grid->get_num_local_dofs();
+  const auto nlevs = m_phys_grid->get_num_vertical_levels();
+  const auto npacks= ekat::PackInfo<N>::num_packs(nlevs);
+
+  const auto horiz_winds_view = get_field_out("horiz_winds").get_view<Pack***>();
+  const auto T_mid_view       = get_field_out("T_mid").get_view<Pack**>();
+
+  // locals for lambda captures to avoid issues on GPU
+  auto raytau0 = m_raytau0;
```
ambrad, line 76: It would be good to do an early return here if m_raytau0 == 0.

components/scream/src/dynamics/homme/atmosphere_dynamics_rayleigh_friction.cpp
```diff
@@ -0,0 +1,93 @@
...
+  auto otau = m_otau;
+
+  const auto policy = ESU::get_default_team_policy(ncols, npacks);
+  Kokkos::parallel_for(policy, KOKKOS_LAMBDA (const KT::MemberType& team) {
+    const int& icol = team.league_rank();
+
+    // If raytau0 == 0 no Rayleigh friction is applied.
+    if (raytau0 != 0) {
```
ambrad, line 84: An early return above means you don't need to check here.

components/scream/src/dynamics/homme/atmosphere_dynamics_rayleigh_friction.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+void HommeDynamics::rayleigh_friction_init()
+{
+  constexpr int N = HOMMEXX_PACK_SIZE;
+  using KT = KokkosTypes<DefaultDevice>;
+  using Pack = ekat::Pack<Real,N>;
+
+  const int nlevs = m_dyn_grid->get_num_vertical_levels();
+  const auto npacks= ekat::PackInfo<Pack::n>::num_packs(nlevs);
+  m_otau = decltype(m_otau)("otau", npacks);
```
ambrad, line 19: If m_raytau == 0, you don't need to allocate this view. Not a big deal since it's a small view, but might as well skip the allocation.

components/scream/src/dynamics/homme/atmosphere_dynamics_rayleigh_friction.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+  const auto ncols = m_phys_grid->get_num_local_dofs();
+  const auto nlevs = m_phys_grid->get_num_vertical_levels();
+  const auto npacks= ekat::PackInfo<N>::num_packs(nlevs);
+
+  const auto horiz_winds_view = get_field_out("horiz_winds").get_view<Pack***>();
+  const auto T_mid_view       = get_field_out("T_mid").get_view<Pack**>();
+
+  // locals for lambda captures to avoid issues on GPU
+  auto raytau0 = m_raytau0;
```
ambrad, line 77: This local is not needed now.

ambrad:
This looks good now. Would you run some tests that diff on Mappy with raytau0 set to 0 to confirm this PR is BFB if raytau0 is 0? You might even consider submitting this PR as a BFB one with raytau0 set to 0 by default, then submitting a short second non-BFB PR that sets its intended default value.

tcclevenger:
@ambrad I like the idea that this PR be BFB and then a PR with 1-line change be non-BFB. I've manually verified BFB, but we'll let the AT do it's thing. I'll add automerge.

ambrad (approved):

tcclevenger:
Weaver is failing with 
```
11:46:35 WARNING: Failed to configure baselines:
11:46:35 CMake Error: The source "/home/e3sm-jenkins/weaver/workspace/SCREAM_PullRequest_Autotester_Weaver/3683/scream/components/scream/CMakeLists.txt" 
does not match the source "/home/e3sm-jenkins/weaver/workspace/SCREAM_PullRequest_Autotester_Weaver/3682/scream/components/scream/CMakeLists.txt" used to generate cache.  Re-run cmake with a different source directory.
```
Not sure what this error is, could this just go away with retesting?

bartgol:
Argh, that's an old acquaintance. It comes up when something happened while generating baselines. Best thing is to wipe baselines folders, and let test-all regenerate them.

tcclevenger:
How to wipe the baseline folders?

bartgol:
Actually, the baselines are "clean", so I don't understand why we're getting the error. Let's see if the AT craps out again.

ambrad (approved):

PR #1873, opened 2022-08-17
EAMxx: use a single GasConcs object in RRTMGPRadiation

bartgol:
Small improvement regarding RRTMGP chunking of the columns: we don't need nchunks separate GasConcs objects. One single object is enough, provided we correctly size it in case the last chunk has smaller size (to avoid some rrtmgp errors when checking array sizes).


bartgol:
@mrnorman pointed me to a new commit in YAKL, which I used to get the current (actual) memory alloc size in the YAKL pool, and I verified that this PR reduces the memory usage by exactly what you'd expect. The memory reduction (on the YAKL side) is not huge, but non-negligible: about 7.5% for a config with 866 cols/rank (ne4 with 1 MPI rank).

Note: I ran rrtmgp_standalone, but the YAKL mem usage should be unchanged, even if we runmore atm processes (rrtmgp is the only YAKL user).

brhillman (approved):
Looks good to me, thanks for putting this together!

PR #1875, opened 2022-08-18
EAMxx: make buildnml leave original output yaml files unchanged

bartgol:
This PR fixes some undesired behavior related to buildnml and the output streams yaml files. In particular,

- the yaml files are copied to $rundir/data/, from wherever they are
- the original files are left unchanged, wherever they are, while the files in $rundir/data are subject to CIME var expansion (and comments removal)

This allows to leave the user original file unchanged. E.g., it still contains useful comments, and also it contains CIME vars, so that the same file can be reused across cases, with CIME vars being correctly expanded each time.


bartgol:
@jgfouca  As a side note, I think `atmchange` should have an `--append` option. This would allow to append to existing arguments, without having to do
```
$ ./atmquery VAR
  VAR: blah
$ ./atmchange VAR="blah, new_blah"
```
and simply do
```
$ ./atmchange --append VAR="new_blah"
```

jgfouca (approved):

PR #1876, opened 2022-08-18
Add calendar information to time variable in output NC files

### LINKED ISSUE #1872: Add calendar (no leap year) info in the metadata of time output for v1
In the v1 output, the time dimension should include information that denotes that there are no leap years in the SCREAMv1 model. Including this info will avoid any confusion down the road when analyzing data (where the timestamp on the output file doesn't match up with what `time` gives when using analysis software like xarray. 
What the `time` variable metadata shows in v1:
```
	double time(time) ;
		time:units = "days since 2010-01-01 00:00:00" ;
		time:long_name = "time" ;
```
For reference, here is what the v0 (EAM) output of time gives:
```
	double time(time) ;
		time:long_name = "time" ;
		time:units = "days since 1979-01-01 00:00:00" ;
		time:calendar = "noleap" ;
		time:bounds = "time_bnds" ;
```

bartgol:
Adds the `calendar: noleap` or `calendar: gregorian` information to the output nc file generated by eamxx. Also, fix a missing include which caused the update of a TimeStamp to never consider leap years.

I verified that generated files do contain the metadata, and that it's `gregorian` if `SCREAM_HAS_LEAP_YEAR=ON` (the default) and `noleap` if `SCREAM_HAS_LEAP_YEAR=OFF`.

Fix #1872

### DIFF
--- a/components/scream/src/share/io/scream_output_manager.cpp
+++ b/components/scream/src/share/io/scream_output_manager.cpp
@@ -206,6 +206,11 @@ void OutputManager::run(const util::TimeStamp& timestamp)
       // Register time as a variable.
       auto time_units="days since " + m_case_t0.get_date_string() + " " + m_case_t0.get_time_string();
       register_variable(filename,"time","time",time_units,1,{"time"},  PIO_REAL,"time");
+#ifdef SCREAM_HAS_LEAP_YEAR
+      set_variable_metadata (filename,"time","calendar","gregorian");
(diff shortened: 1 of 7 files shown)


crterai (approved):
From what I can follow, looks good to me. And also confirmed that running with this PR will indeed add Feb 29th date on leap-years. Thanks for getting this PR in so quickly!

PR #1877, opened 2022-08-19
Some simplifications in the grids manager classes

bartgol:
This PR simplifies a bit some aspects related to grids/grids_manager.

- Grids managers build all the grids they can, depending on input ParameterList options
- Hence, no need for atm procs to tell the AD what grids they need. The GM builds them all, then they simply grab what they need.
- Offer a "Physics" alias for the physics grid, which works regardless of whether the grid is "Physics GLL" or "Physics PG2". Most physics processes don't care.
- Removed the "Grid" namelist entry for Physics processes, since they simply grab "Physics"
- Removed concept of "Reference Grid". In all practical purposes, this will always be "Physics", so simply use that.


tcclevenger (approved):
This is great. It's nice being able to just ask for physics grid, instead of the whole name from params and then get grid.

AaronDonahue (approved):
This is a great clean up of our grids manager infrastructure.  Thank you Luca.  While we are on the subject, one future development effort will be in setting up regridding fields from the simulation grid to either a coarser grid or to a regional grid.  Will any of the changes here impact that?  Make it easier?  Harder?  Not related? :-)

AaronDonahue:
Worth noting that when this PR goes in, since it touches a lot of code, we may need to fix conflicts in a bunch of our PRs.  The pre-inspection should catch it though and we can cross those bridges when they come.

bartgol:
> This is a great clean up of our grids manager infrastructure. Thank you Luca. While we are on the subject, one future development effort will be in setting up regridding fields from the simulation grid to either a coarser grid or to a regional grid. Will any of the changes here impact that? Make it easier? Harder? Not related? :-)

You mean during IO? I don't know, but I suspect it would be unrelated.

bartgol:
Standalone fails are due to a bug in ekat (see e3sm-project/ekat#244). V1 fails are under investigation.

tcclevenger:
It ended up passing. Is it still good to review?

bartgol:
Ah, I think I posted on the wrong PR. This PR does not update ekat, so the bug I linked is not affecting this. Yes, ready to review, thanks.

PR #1880, opened 2022-08-19
Better handling of floating point precision in scorpio interface layer.

bartgol:
In particular:
- for model restart output, always match the precision of the fields (single or double, depending on the build);
- for normal model output, default to single precision, but allow the user to change it from yaml file.

This should allow to save on model output storage for very large grids cases.


components/scream/src/share/io/scream_scorpio_interface.F90
```diff
@@ -1449,9 +1450,13 @@ subroutine grid_read_darray_1d(filename, varname, var_data_ptr, time_index)
     var_size = SIZE(var%compdof)
 
     ! Now we know the exact size of the array, and can shape the f90 pointer
-    if (var%dtype .eq. pio_rtype) then
-      call c_f_pointer (var_data_ptr, var_data_real, [var_size])
-      call pio_read_darray(pio_atm_file%pioFileDesc, var%piovar, var%iodesc, var_data_real, ierr)
+    if (var%dtype .eq. PIO_DOUBLE) then
```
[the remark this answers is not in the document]
bartgol: I think we're already doing that a lot. Our input files have a certain precision (and some fields are double while other single), but we use it for both single and double precision testing. Someone mentioned (prob on slack) that PIO is allowing this to happen, so that the user can read in fp data without worrying about matching the precision of the input file.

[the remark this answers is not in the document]
bartgol: Correction: the issue is not reading a var with precision X when in the file it has precision Y. The issue was telling scorpio that we wanted to read/write a var with precision X, and then passing a pointer to Y. What's in the file is not the problem.

bartgol:
I am not sure what's the reason of the two DIFF. I _did_ bless the current master lowres baselines (blessing the last merged PR) when this PR was half way through testing, but then I would have expected DIFFs in the first few tests, and PASS in the last ones. I will dig a bit.

components/scream/src/share/io/scream_output_manager.cpp
```diff
@@ -215,16 +209,20 @@ void OutputManager::run(const util::TimeStamp& timestamp)
 
       // Register time as a variable.
       auto time_units="days since " + m_case_t0.get_date_string() + " " + m_case_t0.get_time_string();
-      register_variable(filename,"time","time",time_units,1,{"time"},  PIO_REAL,"time");
+      register_variable(filename,"time","time",time_units,1,{"time"}, scorpio::nctype("double"),"time");
```
AaronDonahue, line 212: Is this set to only be double for restarts? I can't tell from the snippet here. I'm reviewing on my phone...

components/scream/src/share/io/scream_scorpio_interface.F90
```diff
@@ -1359,22 +1355,55 @@ end function get_dimlen
...
+
+    ! Local variables
+
+    real(kind=c_float) :: sbuf(buf_size)
+    type(pio_atm_file_t), pointer :: pio_atm_file
+    type(hist_var_t), pointer     :: var
+    integer                       :: ierr,var_size
+    logical                       :: found
+
+    print *, "write_darray_double"
```
AaronDonahue, line 1406: Leftover from debugging?

AaronDonahue (approved):
Looks good. I had two questions. I'm approving knowing that this will need to be manually merged because the single prec should be non-BFB in CIME tests. 
To

components/scream/src/share/io/scream_output_manager.cpp
bartgol: I kept time as a double all the time, since it's a scalar, so turning it to float doesn't really make a dent anyways...

components/scream/src/share/io/scream_scorpio_interface.F90
bartgol: Ah, yes. We'll remove it whenever we put up a PR next.

bartgol:
> Looks good. I had two questions. I'm approving knowing that this will need to be manually merged because the single prec should be non-BFB in CIME tests. To

Actually, I did not set a specific precision for the default model output, which means it will default to whatever "Real" is. Since our CIME runs are all double precision, we should get the same output as before. In order to turn on single-precision output, one needs to add `Floating Point Precision: float` (or `single`) to the model output yaml file.

ambrad:
I'm seeing that on both pm-gpu and ascent, today's nightlies, which ran with this PR merged, take substantially longer than yesterday's. No other PR was merged today.

pm-gpu:
* before: https://my.cdash.org/viewTest.php?buildid=2212826&onlypassed
* after: https://my.cdash.org/viewTest.php?buildid=2213296&onlypassed

ascent:
* before: https://my.cdash.org/viewTest.php?buildid=2212834&onlypassed
* after: https://my.cdash.org/viewTest.php?onlypassed&buildid=2213338

Would you mind doing a v1 test on summit with and without this PR, e.g. PEM_Ln90.ne30pg2_ne30pg2.F2010-SCREAMv1, and looking at the timers?

AaronDonahue:
@ambrad , I'm putting a ninja fix in now.  There was an errant `print` statement that was left in the F90 code which I think is slowing things down.

PR #1886, opened 2022-08-22
Changing naming convention of scream parameters

jgfouca:
No more whitespaces or double underscores. All impacted items were also converted into lowercase.

I included the script I used since it could be useful in the future if we decide to systematically rename things again.

### DIFF
--- a/components/scream/cime_config/namelist_defaults_scream.xml
+++ b/components/scream/cime_config/namelist_defaults_scream.xml
@@ -290,7 +290,7 @@ tweaking their $case/namelist_scream.xml file.
   <!-- List of yaml files containing I/O output specs -->
   <Scorpio>
     <output_yaml_files type="array(string)">${SRCROOT}/components/scream/data/scream_default_output.yaml</output_yaml_files>
-    <Model__Restart>
+    <model_restart>
       <Casename>./${CASE}.scream</Casename>
       <Output__Control>
         <Frequency>${REST_N}</Frequency>
@@ -300,7 +300,7 @@ tweaking their $case/namelist_scream.xml file.
         <AVG__Type__in__Filename>false</AVG__Type__in__Filename>
         <Frequency__in__Filename>false</Frequency__in__Filename>
       </Output__Control>
-    </Model__Restart>
+    </model_restart>
   </Scorpio>
 
   <!-- Debug options for scream -->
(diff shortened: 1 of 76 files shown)


components/scream/cime_config/namelist_defaults_scream.xml
```diff
@@ -47,7 +47,7 @@ tweaking their $case/namelist_scream.xml file.
       <sections>ctl_nl</sections>
     </file>
     <file name="data/scream_input.yaml" format="yaml">
-      <sections>Debug,atmosphere_processes,Grids__Manager,Initial__Conditions,Scorpio,E3SM__Parameters</sections>
+      <sections>Debug,atmosphere_processes,grids_manager,initial_conditions,Scorpio,e3sm_parameters</sections>
```
bartgol, line 50: while at it, we could change `Scorpio` to `scorpio` (lowercase) as well.

bartgol (approved):
I still see several single-word parameters with Uppercase, like Scorpio here. Should we take the chance to make things uniform (and use lowercase everywhere)?

jgfouca:
@bartgol  , 

> Should we take the chance to make things uniform (and use lowercase everywhere)?

That would be nice. Unfortunately, changing something like `<Grid>` to `<grid>` will likely defeat my script because there will be too many false matches in sed. I can try changing these by hand.

components/scream/src/share/io/scream_output_manager.cpp
```diff
@@ -98,16 +98,16 @@ setup (const ekat::Comm& io_comm, const ekat::ParameterList& params,
...
       m_checkpoint_control.nsteps_since_last_write = 0;
 
       // File specs
       m_checkpoint_file_specs.max_snapshots_in_file = 1;
       m_checkpoint_file_specs.num_snapshots_in_file = 0;
       m_checkpoint_file_specs.filename_with_time_string = pl.get("Timestamp in Filename",true);
       m_checkpoint_file_specs.filename_with_mpiranks    = pl.get("MPI Ranks in Filename",false);
-      m_checkpoint_file_specs.filename_with_avg_type    = pl.get("AVG Type in Filename",true);
-      m_checkpoint_file_specs.filename_with_frequency   = pl.get("Frequency in Filename",true);
+      m_checkpoint_file_specs.filename_with_avg_type    = pl.get("avg_type_in_filename",true);
```
AaronDonahue, line 109: @jgfouca and @bartgol , did we also want to change the two parameters above this one?  Timestep in Filenam and MPI Ranks in Filename?

jgfouca: @AaronDonahue , those parameters aren't in `cime_config/namelist_defaults_scream.xml` so they weren't processed by my script.

PR #1891, opened 2022-08-24
Add a general horizontal remap utility

AaronDonahue:
This commit introduces a horizontal remapping structure fashioned on the approach used in the component coupler.

The new horizontal remapper has better memory management, enabling the use of very fine resolution remapping (i.e. ne30->ne1024) without hitting OOM errors.

This commit also introduces a robust randomized test which compares the use of the remapper given local information or remapping data provided by file and comparing everything against a baseline.


AaronDonahue:
Keeping this as WIP, there are still a few steps to do to cleanup the PR before review.  Although if anyone likes looking through changed code they are absolutely welcome to offer preliminary comments that I can incorporate.

components/scream/src/share/io/scream_scorpio_interface.cpp
```diff
@@ -121,8 +121,8 @@ void grid_read_data_array(const std::string &filename, const std::string &varnam
   grid_read_data_array_c2f(filename.c_str(),varname.c_str(),time_index,hbuf);
 }
 /* ----------------------------------------------------------------- */
-void grid_write_data_array(const std::string &filename, const std::string &varname, const Real* hbuf) {
-  grid_write_data_array_c2f_real(filename.c_str(),varname.c_str(),hbuf);
+void grid_write_data_array(const std::string &filename, const std::string &varname, void *hbuf) {
```
AaronDonahue, line 124: @bartgol , I noticed in the scorpio Slack space you were asking about something that may be related to this change.  So I thought I would ping you just in case you are working on something similar.

bartgol: Last weekend, I was playing around with scorpio C (to bypass flib). The C interface uses a non-const buffer, which forces some const cast. In our master, however, we fwd the call to F90, where const doesn't matter, so we don't need this change.

AaronDonahue: I think we still need this change here.  I added it to allow us to write both REAL and INT output.  Currently in master we can only write REAL.  I copied the syntax from read_array which can support both formats.

bartgol: Ah, sorry, yes, the `void` part you do need it. But you should keep `const` in front, to allow to be able to write read-only fields.

components/scream/src/share/grid/remap/horizontal_remap_utility.hpp
```diff
@@ -0,0 +1,564 @@
...
+using view_3d = typename KT::template view_3d<S>;
+
+template <typename S>
+using view_1d_host = typename KT::template view_1d<S>::HostMirror;
+
+template <typename S>
+using view_2d_host = typename KT::template view_2d<S>::HostMirror;
+
+public:
+  virtual ~RemapSegment() = default;
```
jeff-cohere, line 41: Do you intend for this to be a base class with subclasses inheriting from it? If not, you don't need a `virtual` destructor.

components/scream/src/share/grid/remap/horizontal_remap_utility.hpp
```diff
@@ -0,0 +1,564 @@
...
+using view_3d = typename KT::template view_3d<S>;
+
+template <typename S>
+using view_1d_host = typename KT::template view_1d<S>::HostMirror;
+
+template <typename S>
+using view_2d_host = typename KT::template view_2d<S>::HostMirror;
+
+public:
+  virtual ~GSMap() = default;
```
jeff-cohere, line 192: No virtual destructor needed if this isn't a base class.

components/scream/src/share/grid/remap/horizontal_remap_utility.hpp
```diff
@@ -0,0 +1,564 @@
...
+      for (int ii=0;ii<m_dofs_gids.size();ii++) {
+        if (!found[ii]) {
+          printf("No segment found for DOF = %d\n",dofs_gids_h(ii));
+        }
+      }
+    }
+    EKAT_REQUIRE(pass);
+  };
+/*---------------------------------------------*/
+  // Useful if needing to grab source data from somewhere, like a data file.
```
jeff-cohere, line 268: It might be helpful to document exactly what this function does (in terms of the members of the `GSMap`, with preconditions and postconditions, etc), since it takes no arguments.

components/scream/src/share/grid/remap/horizontal_remap_utility.hpp
```diff
@@ -0,0 +1,564 @@
...
+
+template <typename S>
+using view_2d_host = typename KT::template view_2d<S>::HostMirror;
+
+public:
+  virtual ~GSMap() = default;
+  GSMap() {};
+
+/*---------------------------------------------*/
+  void set_name(const std::string& name) { map_name = name; }
```
jeff-cohere, line 196: Just curious--what's the significance of this name?

components/scream/src/share/grid/remap/horizontal_remap_utility.hpp
```diff
@@ -0,0 +1,564 @@
...
+/*---------------------------------------------*/
+  Int get_num_of_segs() {return map_segments.size();}
```
jeff-cohere, line 198: Consider `num_segments` and `num_dofs`. :-)

components/scream/src/share/grid/remap/horizontal_remap_utility.hpp
```diff
@@ -0,0 +1,564 @@
...
+      for (int ii=0; ii<seg.m_length; ii++) {
+        auto idx = std::find(unique_dofs.begin(), unique_dofs.end(), src_dofs_h(ii));
+        int idx_ii = idx - unique_dofs.begin();
+        src_idx_h(ii) = idx - unique_dofs.begin();
+      }
+      Kokkos::deep_copy(src_idx,src_idx_h);
+    }
+  }
+/*---------------------------------------------*/
+  RemapSegment get_segment_impl(const gid_type dof) {
```
jeff-cohere, line 302: These `_impl` functions should probably be `private`, to keep from cluttering the interface.

components/scream/src/share/grid/remap/horizontal_remap_utility.hpp
```diff
@@ -0,0 +1,564 @@
...
+  for (int iseg=0; iseg<get_num_of_segs(); iseg++) {
+    auto seg = map_segments[iseg];
+    ScalarT remap_val = seg.apply_segment<ScalarT>(source_data);
+    remap_data_h(iseg) = remap_val;
+  }
+  Kokkos::deep_copy(remapped_data,remap_data_h);
+}
+/*---------------------------------------------*/
+
+protected:
```
jeff-cohere, line 551: More C++ trivia: in classes that don't belong to an inheritance hierarchy (like this one), `protected` is the same as `private`, so it might be less confusing to use the latter.

components/scream/src/share/grid/remap/horizontal_remap_utility.hpp
```diff
@@ -0,0 +1,564 @@
...
+  }
+/*---------------------------------------------*/
+  std::vector<gid_type> get_unique_dofs() { return unique_dofs; };
+/*---------------------------------------------*/
+  void add_segment(RemapSegment& segment) {
+    add_segment_impl(segment);
+  }; 
+/*---------------------------------------------*/
+  // Useful for debugging, print all the GSMap info.
+  void print_GSMap() {
```
jeff-cohere, line 221: Consider making this function and the `check` function `const`.

components/scream/src/share/grid/remap/horizontal_remap_utility.hpp
```diff
@@ -0,0 +1,564 @@
...
+      auto src_data_sub = ekat::subview(source_data_h,idx);
+      for (int kk=0; kk<src_data_sub.extent(0); kk++) {
+        ret_h(kk) = ret_h(kk) + src_data_sub(kk)*weights_h(ii);
+      }
+    }
+    Kokkos::deep_copy(ret,ret_h);
+    return ret;
+  }
+/*---------------------------------------------*/
+  template <typename ScalarT>
```
jeff-cohere, line 125: Let's add a brief description of what this is doing. As simple as it is, it will help folks understand what's going on.

components/scream/src/share/grid/remap/horizontal_remap_utility.hpp
```diff
@@ -0,0 +1,564 @@
...
+
+  gid_type          m_dof;       // The degree of freedom (dof) on target mesh this segment maps to.
+  view_1d<gid_type> source_dofs; // A view of the dofs on the source grid that map to this target dof.
+  view_1d<int>      source_idx;  // Local index in the set of unique columns for source data - set by GSMap
+  view_1d<Real>     weights;     // A view of the associated weights in the mapping.
+  Int               m_length;    // The size of this segment.
+
+/*---------------------------------------------*/
+  // Helper function to check a segment is valid.
+  bool check() {
```
jeff-cohere, line 60: Consider making this and the `print_X` function below `const`.

components/scream/src/share/grid/remap/horizontal_remap_utility.hpp
```diff
@@ -0,0 +1,564 @@
...
+template <typename S>
+using view_3d = typename KT::template view_3d<S>;
+
+template <typename S>
+using view_1d_host = typename KT::template view_1d<S>::HostMirror;
+
+template <typename S>
+using view_2d_host = typename KT::template view_2d<S>::HostMirror;
+
+public:
```
jeff-cohere, line 40: In a `struct`, everything is `public`, so you don't need this.

components/scream/src/share/grid/remap/horizontal_remap_utility.hpp
```diff
@@ -0,0 +1,564 @@
...
+    printf("\n--------------------\n");
+  }
+/*---------------------------------------------*/
+
+}; // end struct RemapSegment
+
+/* --------------------------------------------------------------------------------------------- */
+/* A structure to store all of the map information 
+ * --------------------------------------------------------------------------------------------- */
+struct GSMap { // TODO: Following the name in component coupler, could be a different name
```
jeff-cohere, line 169: This type has behaviors implemented by things that you have marked `private`, so arguably it should be a `class` instead of a `struct`.

components/scream/src/share/grid/remap/horizontal_remap_utility.hpp
AaronDonahue: Useful for error messages when you may have more than one horizontal remapper defined

AaronDonahue:
Thanks for all the comments Jeff.  I'm in the process of cleaning up the GSMap structure (which might be better as a class :-)) So I can implement these fixes.

components/scream/src/share/grid/remap/horizontal_remap_utility.hpp
```diff
@@ -0,0 +1,564 @@
...
+
+namespace scream {
+
+
+/* --------------------------------------------------------------------------------------------- */
+/* A simple structure to store remap information for a single target column
+ * --------------------------------------------------------------------------------------------- */
+struct RemapSegment {
+
+using gid_type = AbstractGrid::gid_type;
```
bartgol, line 20: Indenting these types helps to quickly realize they are local to the struct.

components/scream/src/share/grid/remap/horizontal_remap_utility.hpp
AaronDonahue: Any candidates for a better name?  `eamxx_horizontal_remap`?  `horizontal_map`?  `HorizMap`?

jeff-cohere: `ColumnMap`, perhaps? Since it maps from a source column space to a target column space?

components/scream/cmake/machine-files/quartz.cmake
```diff
@@ -20,3 +20,5 @@ set(SCREAM_MPIRUN_EXE "srun" CACHE STRING "")
 set(SCREAM_MPI_NP_FLAG "-n" CACHE STRING "")
 set(SCREAM_MPI_EXTRA_ARGS "" CACHE STRING "")
 set(SCREAM_INPUT_ROOT "/usr/gdata/climdat/ccsm3data/inputdata" CACHE STRING "")
+
+set(SCREAM_ENABLE_STATM TRUE CACHE BOOL "")
```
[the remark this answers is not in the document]
jeff-cohere: Ah, never mind. This is a machine file.

AaronDonahue:
Closing, being replaced by #1945

PR #1892, opened 2022-08-24
Use reduced spectral resolution for RRTMGP

brhillman:
Use reduced spectral resolution for RRTMGP. This uses the reduced spectral resolution absorption data developed externally by RRTMGP developers, which uses half the number of "g-points". This is consistent with the spectral resolution used in the RRTMG (no P) implementation in EAMv2, and should speed up the execution and reduce the memory footprint of RRTMGP in EAMxx by about a factor of two, with minimal impact to the solution accuracy. This also adds the capability to configure the spectral resolution by specifying the input file to use for radiation in the case directory via `atmchange`, rather   than hard-coding the path in the RRTMGP interface. This will be non-BFB for CIME cases that now use the reduced spectral resolution by default, but BFB for standalone tests, which use a mix of the full and reduced resolution depending on the test. This is done both for additional test coverage (to make sure the actual code is agnostic to input data size), but also because the `rrtmgp_tests` and `rrtmgp_standalone_unit` tests evaluate a solution against a reference problem that use the full spectral resolution.

### DIFF
--- a/components/scream/src/physics/rrtmgp/CMakeLists.txt
+++ b/components/scream/src/physics/rrtmgp/CMakeLists.txt
@@ -94,8 +94,8 @@ target_include_directories(scream_rrtmgp SYSTEM PUBLIC
 
 # Ensure RRTMGP lookup tables are present in the data dir
 set (RRTMGP_TABLES
-  init/rrtmgp-data-sw-g224-2018-12-04.nc
-  init/rrtmgp-data-lw-g256-2018-12-04.nc
+  init/rrtmgp-data-sw-g112-210809.nc #rrtmgp-data-sw-g224-2018-12-04.nc
+  init/rrtmgp-data-lw-g128-210809.nc #rrtmgp-data-lw-g256-2018-12-04.nc
   init/rrtmgp-allsky.nc
   init/rrtmgp-cloud-optics-coeffs-sw.nc
   init/rrtmgp-cloud-optics-coeffs-lw.nc
--- a/components/scream/src/physics/rrtmgp/CMakeLists.txt
+++ b/components/scream/src/physics/rrtmgp/CMakeLists.txt
@@ -94,8 +94,8 @@ target_include_directories(scream_rrtmgp SYSTEM PUBLIC
 
 # Ensure RRTMGP lookup tables are present in the data dir
 set (RRTMGP_TABLES
-  init/rrtmgp-data-sw-g112-210809.nc #rrtmgp-data-sw-g224-2018-12-04.nc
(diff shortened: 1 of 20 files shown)


components/scream/src/physics/rrtmgp/CMakeLists.txt
```diff
@@ -76,8 +76,8 @@ target_include_directories(scream_rrtmgp SYSTEM PUBLIC
 
 # Ensure RRTMGP lookup tables are present in the data dir
 set (RRTMGP_TABLES
-  init/rrtmgp-data-sw-g224-2018-12-04.nc
-  init/rrtmgp-data-lw-g256-2018-12-04.nc
+  init/rrtmgp-data-sw-g112-210809.nc #rrtmgp-data-sw-g224-2018-12-04.nc
```
AaronDonahue, line 79: any reason to keep the commented out part?

brhillman:
Some quick timing results from ne4pg2 on 24 procs of cori-knl 10 days with no output, with original full resolution data:
```
name                                                on  processes  threads        count      walltotal   wallmax (proc   thrd  )   wallmin (proc   thrd  )
"a:EAMxx::Radiation::run"                            -         24       24 5.760000e+03   5.404421e+03   262.649 (    17      0)   185.220 (     9      0)
```
and with reduced resolution data:
```
name                                                on  processes  threads        count      walltotal   wallmax (proc   thrd  )   wallmin (proc   thrd  )
"a:EAMxx::Radiation::run"                            -         24       24 5.760000e+03   2.779081e+03   135.143 (    16      0)    95.803 (     9      0)
```

AaronDonahue (approved):
Just has a question, but more a style question so approving.

components/scream/src/physics/rrtmgp/CMakeLists.txt
brhillman: Nope, good catch will fix.

brhillman: Fixed.

AaronDonahue:
@brhillman This PR might have conflicts with master now which is holding up the AT.  Can you take a look and merge master if needed?  Note, we just reverted the IO PR, so if that is where the conflict lies it might have been fixed.

AaronDonahue (dismissed):

brhillman:
@AaronDonahue just updated.

brhillman:
Diffs in v1 CIME cases are expected, I think this is ready to merge.

AaronDonahue (approved):

AaronDonahue:
manually merging this PR

PR #1894, opened 2022-08-24
EAMxx: in test-all-scream, cat the correct file when ctest fails

bartgol:
If the build fails at config time, cat `LastConfigure_XYZ.log`, if it fails at build time, cat `LastBuild_XYZ.log`, if it fails at testing time, cat `LastTestsFailed_XYZ.log`.

### DIFF
--- a/components/scream/scripts/test_all_scream.py
+++ b/components/scream/scripts/test_all_scream.py
@@ -922,10 +922,6 @@ remove existing baselines first. Otherwise, please run 'git fetch $remote'.
             for file in files:
                 file_no_path = file.name
                 tokens = re.split('_|-|\.',str(file_no_path))
-                print (tokens)
-                #  if len(tokens)!=4:
-                #      print (f"tokens: {','.join('{}'.format(t) for t in tokens)}")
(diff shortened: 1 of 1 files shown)


components/scream/scripts/test_all_scream.py
```diff
@@ -888,14 +889,61 @@ def run_all_tests(self):
...
+            curr_tag=0
+            curr_idx=0
+            latest = None
+            # ctest creates files of the form Last{phase}_$TIMESTAMP-$IDX.log
+            # we split the name, and sue $TIMESTAMP to pick the newest, and, in case
+            # of tie, $IDX as tiebreaker
+            for file in files:
+                file_no_path = file.name
+                tokens = re.split('_|-|\.',str(file_no_path))
+                print (tokens)
```
jgfouca, line 925: Looks like maybe some debugging code was left in here?

jgfouca (changes requested):

jgfouca (dismissed):

jgfouca (approved):

PR #1895, opened 2022-08-25
Propagate flags set by CIME to scream

bartgol:
Up to now, we were completely ignoring the flags set by CIME in the vars `CPPDEFS`, `CXXFLAGS`, `CFLAGS`, and `FFLAGS`. Honestly, I always assumed e3sm would stuff these inside `CMAKE_<LANG>_FLAGS`, and expected them to be already there when scream is processed. However, e3sm chooses to manually set flags on source files, which means we need to modify the flags ourselves.

Notice that, upon returning from the `build_scream` macro, the `CMAKE_<LANG>_FLAGS` vars are restored to their original value (which, as of now, is empty).

I verified that:
- `<LANG>FLAGS` (and `CPPDEFS`) are indeed propagated to scream's targets
- `<LANG>FLAGS` (and `CPPDEFS`) are unchanged outside the scream folder (i.e., scream does not modify them inadvertently).
- `CMAKE_<LANG>_FLAGS` are unchanged outside the scream folder (currently, e3sm never sets these, so they're empty)


ambrad:
@jgfouca's review is more important than mine, so I removed my name from the reviewer list.

ambrad:
For clarity, can you confirm that you intend to update the EKAT module in this PR?

bartgol:
> For clarity, would you confirm that you intend to update the EKAT module in this PR?

Yes, this PR is supposed to do that too. Most likely, the ekat change will have no effect (as is, EKAT was really not adding the warning flags to the compile line), but for the sake of known fixing flags-related issues, I included it.

components/scream/scripts/test_all_scream.py
```diff
@@ -108,7 +108,8 @@ def __init__(self):
             self,
             "debug_nopack_fpe",
             "debug pksize=1 floating point exceptions on",
-            DBG.CMAKE_ARGS + [("SCREAM_PACK_SIZE", "1")],
+            DBG.CMAKE_ARGS + [("SCREAM_PACK_SIZE", "1"), ("EKAT_ENABLE_FPE","True")],
+            uses_baselines=False
```
bartgol, line 112: @jgfouca This one is a bit unrelated, but on my machine it was throwing since I did not have fpe baselines. And I started wondering if we do care about baselines for FPE builds. I think in FPE builds, we are really testing code correctness (much like a memcheck build), so I removed baselines for this.

But I can roll this back if you disagree.

jgfouca:
@bartgol , I agree with this change. Changing SCREAM flags within a CIME build is very difficult currently. That said, this is a major philosophical change for the SCREAM project. We have always said that "scream manages its own flags" but now that's no longer true. I'm not sure how SCREAM will play with the various nuances of the CIME build system, like Depends files and all the things you can do in the macros. At a cursory glance, I think most of that stuff should work?

@PeterCaldwell any thoughts?

I think this is a good first step so I'm going to approve.

jgfouca (approved):

bartgol:
> @bartgol , I agree with this change. Changing SCREAM flags within a CIME build is very difficult currently. That said, this is a major philosophical change for the SCREAM project. We have always said that "scream manages its own flags" but now that's no longer true. I'm not sure how SCREAM will play with the various nuances of the CIME build system, like Depends files and all the things you can do in the macros. At a cursory glance, I think most of that stuff should work?

I think scream can still nudge things, by appending flags. IIRC, flags work in a "last-one-wins" way, in case of conflicting options. So scream, by appending to `CMAKE_<LANG>_FLAGS`, can still achieve some degree of independence.

In general, users should be able to set flags via case management. SCREAM should only set technical flags needed for particular reasons (e.g., disable fmad for GPU runs in debug mode). Perhaps scream should set all other flags (e.g., common flags like `-assume byterecl -ftz`) only in standalone mode. I don't know, we should think about this.

bartgol:
Ah, the fails on mappy are a good source of discussion: the flags from CIME contain `-ffpe-trap=zero,overflow`, and something in SCREAM is throwing. I suspect it's some padded value in a pack. SCREAM is supposed to set the fpe mask to 0, and the log shows it:
```
-------- EKAT CONFIGS --------

 ExecSpace name: OpenMP
 ExecSpace initialized: yes 
 active avx set: -AVX2-AVX
 compiler id: GCC 
 default FPE mask: 0 (NONE) 
```
Inside every call to scream from the `atm_comp_mct.F90` module, we do something like
```
disable_fpes();
run_whatever_was_to_be_run();
re_instate_fpes();
```
However, it seems that this fpe guard system is not working properly, and FPEs are still generated. I suspect that the compile-time flag `-ffpe-trap` trumps the runtime fpe mask set in scream.

I will dig a bit, to see if maybe it's simply a bug in how we impl-ed the fpe guard (in which case I'll fix it). But if there's no way around this, we will probably have to make scream override the `-ffpe-trap` flag.

bartgol:
Re: the possibility of overriding CIME's settings, from GCC manual
```
If the option is used more than once in the command line, the lists will be joined: ’ffpe-trap=list1 ffpe-trap=list2’ is equivalent to ffpe-trap=list1,list2.

Note that once enabled an exception cannot be disabled (no negative form).
```
So, current options:
- check if this is simply a bug in our fpe guard in scream_cxx_f90_interface.cpp, and if so, fix it.
- if the above fails, we need to remove `-ffpe-trap` from FFLAGS in f90 if scream is to be built without FPEs on.

bartgol:
Ok, I tried a simple program, with an f90 main that calls a cxx function, in which I do a div by zero. I added `-ffpe-trap=zero` to the Fortran flags, and tried to disable the exception on the C++ side. The fpe is correctly disabled, and the code prints `inf`. If I comment out the line that disables exceptions, the f90 code crashes with the FPE.
I also tried to have the C++ code retrieve the current fe mask (which should be set in f90 via the compiler option), and it captures it correctly.

So in theory, our approach that disables fpes _should_ work in our mct interface. Perhaps we're doing something wrong.

bartgol:
Ok, the issue was that, in the mct coupling layer, we relied on EKAT for both enabling FPE support as well as for getting the default fpe mask flag.

So I changed our approach: we *always* set EKAT_ENABLE_FPE=ON, but we *never* use ekat's default fpes. Instead, we use scream's default FPEs, which are nontrivial if SCREAM_FPE=ON.

jgfouca (approved):

jgfouca (approved):

PR #1899, opened 2022-08-26
EAMxx: reduce cost of memcheck build

bartgol:
For multi-rank system tests, only build the test with most ranks. Also, for memchek/coverage builds, set the test size (which most coupled tests use to determine the number of time steps) to short.

### DIFF
--- a/components/scream/scripts/test_all_scream.py
+++ b/components/scream/scripts/test_all_scream.py
@@ -155,8 +155,6 @@ def test_factory(user_req_tests, machine, mem_check):
     if mem_check:
         for test in result:
             test.longname += "_cuda_mem_check" if is_cuda_machine(machine) else "_valgrind"
-            test.uses_baselines = False
-
 
     return result
 
--- a/components/scream/scripts/test_all_scream.py
+++ b/components/scream/scripts/test_all_scream.py
@@ -155,6 +155,8 @@ def test_factory(user_req_tests, machine, mem_check):
     if mem_check:
         for test in result:
             test.longname += "_cuda_mem_check" if is_cuda_machine(machine) else "_valgrind"
+            test.uses_baselines = False
+
 
     return result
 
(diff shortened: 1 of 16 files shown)


components/scream/scripts/test_all_scream.py
```diff
@@ -155,6 +155,8 @@ def test_factory(user_req_tests, machine, mem_check):
     if mem_check:
         for test in result:
             test.longname += "_cuda_mem_check" if is_cuda_machine(machine) else "_valgrind"
+            test.uses_baselines = False
```
jgfouca, line 158: Is this needed? I see around line 418 that baselines are already disabled if mem_check is on.

components/scream/tests/coupled/dynamics_physics/homme_shoc_cld_p3_rrtmgp/CMakeLists.txt
```diff
@@ -63,17 +63,17 @@ configure_file(${SCREAM_SRC_DIR}/dynamics/homme/tests/theta.nl
 GetInputFile(init/${EAMxx_tests_IC_FILE_72lev})
 
 ## Finally compare all MPI rank output files against the single rank output as a baseline, using CPRNC
-## Only if max mpi ranks is >1
+## Only if running with 2+ ranks configurations
 # This test requires CPRNC
-if (SCREAM_TEST_MAX_RANKS GREATER 1)
+if (TEST_RANK_START GREATER TEST_RANK_END)
```
jgfouca, line 68: I see the following code pattern was repeated in at least 6 other places. Might be nice to encapsulate.

jgfouca (changes requested):
Couple minor questions.

components/scream/scripts/test_all_scream.py
bartgol: I thought I was getting an error due to missing baselines on my machine. Perhaps the checks before line 418? Let me verify.

bartgol: You're right, the line was not needed. I'm not sure why I thought I had to add it. I purged it.

jgfouca (approved):

PR #1906, opened 2022-08-30
EAMxx: fix time handling in SPA

bartgol:
Spa was correctly computing the current time as `ts.frac_of_year_in_days()+dt`. However, it was passing `ts` rather than `ts+dt` to the main routine. In the case where `dt` takes you to the next month, SPA was using beg/end from the previous month. This was especially bad when the next month has more days, since we'd get `t_now-t_beg>days_in_month`, leading to a non-convex interpolation.

However, this should have caused a problem _only if_ the start date/time and ATM_NCPL are such that we have a timestep that crosses midnight. E.g., if start time is XX:15 and we use a timestep of 30 minutes.

### DIFF
--- a/components/scream/src/physics/spa/atmosphere_prescribed_aerosol.cpp
+++ b/components/scream/src/physics/spa/atmosphere_prescribed_aerosol.cpp
@@ -203,9 +203,9 @@ void SPA::initialize_impl (const RunType /* run_type */)
 void SPA::run_impl (const int dt)
 {
   /* Gather time and state information for interpolation */
-  auto ts = timestamp();
+  auto ts = timestamp()+dt;
   /* Update the SPATimeState to reflect the current time, note the addition of dt */
-  SPATimeState.t_now = ts.frac_of_year_in_days() + dt/86400.;
+  SPATimeState.t_now = ts.frac_of_year_in_days();
   /* Update time state and if the month has changed, update the data.*/
   SPAFunc::update_spa_timestate(m_spa_data_file,m_nswbands,m_nlwbands,ts,SPAHorizInterp,SPATimeState,SPAData_start,SPAData_end);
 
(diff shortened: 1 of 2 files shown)


AaronDonahue (approved):
Good catch and I like the new assert to make sure we are not doing concave interpolation.  That should at least provide a better error message to figure out.

ndkeen:
I tried a case with this branch.  It fails at the day *before* it was failing earlier. 

```
 tStamp_write: model date =   00010227       0 wall clock = 2022-08-30 10:12:20 avg dt =    32.70 dt =    32.36
 tStamp_write: model date =   00010228       0 wall clock = 2022-08-30 10:12:53 avg dt =    32.69 dt =    32.23
```


```
385: terminate called after throwing an instance of 'std::logic_error'
385:   what():  /global/cfs/cdirs/e3sm/ndk/bartgol_spa-timestate-fix/components/scream/src/physics/spa/spa_functions_impl.hpp:173: FAIL:
385: delta_t_fraction>=0 && delta_t_fraction<=1
385: Error! Convex interpolation with coefficient out of [0,1].
385:   t_now  : 59.000000
385:   t_beg  : 60.000000
385:   delta_t: 31.000000
```

AaronDonahue:
okay, so t_beg=60 implies March 1st non-leap year right... 31+28+1
t_now=59 implies similarly that we are at Feb 28th

bartgol:
Ugh, I get this error also in the standalone test. I'll fix it up.

ndkeen:
With your last commit, it will run at least 5months.  Submitting for longer.

ndkeen (approved):
I hope this was not waiting for my review

bartgol:
The PR passes all standalone tests, as well as the noAero v1 test. The v1 tests that use SPA are expected to fail, since we fixed the interpolation coefficient.

@AaronDonahue are we clear for a non-BFB merge?

bartgol:
The PR passes all standalone tests, as well as the noAero v1 test. The v1 tests that use SPA are expected to fail, since we fixed the interpolation coefficient.

@AaronDonahue are we clear for a non-BFB merge?

bartgol:
The PR passes all standalone tests, as well as the noAero v1 test. The v1 tests that use SPA are expected to fail, since we fixed the interpolation coefficient.

@AaronDonahue are we clear for a non-BFB merge?

AaronDonahue:
@bartgol Yes, we are clear for a non-BFB merge.

PR #1910, opened 2022-08-31
Added F2010-SCREAMv1-MPASSI and tests

wlin7:
Added compset F2010-SCREAMv1-MPASSI to use mpassi for
prescribed sea-ice mode and mosart for river model.  A test suite
e3sm_scream_v1_mpassi is added for testing the new compset.

Also Made 128 level default for high resolution (ne256,ne512, ne1024).

[BFB] for existing tests. new tests (scream_v1_mpassi) added.

### DIFF
--- a/cime_config/tests.py
+++ b/cime_config/tests.py
@@ -454,7 +454,7 @@ _TESTS = {
 
     "e3sm_scream_v1" : {
         "time"    : "03:00:00",
-        "inherit" : ("e3sm_scream_v1_lowres", "e3sm_scream_v1_medres"),
+        "inherit" : ("e3sm_scream_v1_lowres", "e3sm_scream_v1_medres","e3sm_scream_v1_mpassi"),
     },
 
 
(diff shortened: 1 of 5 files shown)


AaronDonahue (approved):
Thanks Wuyin!

wlin7:
This PR mirrors #1861 but with a new compset name for the configuration with mosart and mpassi.
Basic SMS, ERS and ERP tests passed on cori-knl with intel compiler.

cime_config/tests.py
```diff
@@ -488,6 +488,19 @@
...
+    "e3sm_scream_v1_mpassi" : {
+        "time"  : "01:00:00",
+        "tests" : (
+            "ERP_D_Ln9.ne4_oQU240.F2010-SCREAMv1-MPASSI.atmlndactive-rtm_off",
+            "SMS_D_Ln9.ne4_oQU240.F2010-SCREAMv1-MPASSI-noAero",
+            "ERP_Ln22.ne4pg2_oQU480.F2010-SCREAMv1-MPASSI.atmlndactive-rtm_off",
+            "ERS_D_Ln22.ne4pg2_oQU480.F2010-SCREAMv1-MPASSI.atmlndactive-rtm_off",
+            "ERS_Ln22.ne30_oECv3.F2010-SCREAMv1-MPASSI.atmlndactive-rtm_off",
+            "PEM_Ln90.ne30pg2_EC30to60E2r2.F2010-SCREAMv1-MPASSI",
+            "ERS_Ln22.ne30pg2_EC30to60E2r2.F2010-SCREAMv1-MPASSI.atmlndactive-rtm_off",
```
bartgol, line 500: That's a lot of cases to add to our nightlies. Can we trim it down to 4-5 tests?

bartgol: The 3rd and 4th are very similar. We can run ERS manually as part of debugging, if the ERP one fails.

components/scream/cime_config/config_component.xml
```diff
@@ -26,8 +26,9 @@
     <type>char</type>
     <valid_values></valid_values>
     <default_value></default_value>
-    <values modifier='additive'>
+    <values match='last'>
```
bartgol, line 29: @jgfouca is prob better than me to review XML stuff, but isn't this preventing people from appending CMake config options via xmlchange? Or can they still do `xmlchange --append SCREAM_CMAKE_OPTIONS="SOME_VAR my_value"`?

cime_config/tests.py
wlin7: Sounds good. Some of the tests were intended to replace those in lowres and medres suite (when the compset is adopted as default). For current purpose,  no more than three necessary (ERP, ERS_D and PEM).

components/scream/cime_config/config_component.xml
wlin7: Yes, it is for that purpose and make it work out of box.  This field currently uses only one long string as value, so 'last' is appropriate and high-res grid could match both entries.

bartgol (approved):

bartgol:
The diffs in this PR are just due to master having just been updated, and not yet blessed. We can merge.

PR #1911, opened 2022-08-31
Allow to reduce size of e3sm log, by optionally suppressing pre/post condition checks warnings

bartgol:
A couple of mods:

- Change construction of pre/post condition checks: do not pass CheckFailHandling as template arg. It's pointless, it can be a runtime arg.
- Each atm proc can receive the parameter `repair_log_level`, which establishes the level of logging for when the pre/post condition checks fail, and are repaired. Default is 'warn' for standalone, and 'trace' for v1 runs. This means that, with likely choices of atm log level, standalone will print the warnings, while v1 will not, keeping logs sizes under control.

### DIFF
--- a/components/scream/src/share/atm_process/atmosphere_process.cpp
+++ b/components/scream/src/share/atm_process/atmosphere_process.cpp
@@ -32,6 +32,21 @@ AtmosphereProcess (const ekat::Comm& comm, const ekat::ParameterList& params)
       "  - Num subcycles: " + std::to_string(m_num_subcycles) + "\n");
 
   m_timer_prefix = m_params.get<std::string>("Timer Prefix","EAMxx::");
+
+  const auto& repair_log_level = m_params.get<std::string>("repair_log_level","warn");
+  if (repair_log_level=="off") {
+    m_repair_log_level = LogLevel::off;
+  } else if (repair_log_level=="trace") {
+    m_repair_log_level = LogLevel::trace;
+  } else if (repair_log_level=="debug") {
+    m_repair_log_level = LogLevel::debug;
+  } else if (repair_log_level=="info") {
+    m_repair_log_level = LogLevel::info;
+  } else if (repair_log_level=="warn") {
+    m_repair_log_level = LogLevel::warn;
+  } else {
+    EKAT_ERROR_MSG ("Invalid value for 'repair_log_level': " + repair_log_level + "\n");
+  }
 }
 
 void AtmosphereProcess::initialize (const TimeStamp& t0, const RunType run_type) {
@@ -182,6 +197,7 @@ void AtmosphereProcess::set_computed_group (const FieldGroup& group) {
 
 void AtmosphereProcess::run_precondition_checks () const {
   // Run all pre-condition property checks
+
   for (const auto& it : m_precondition_checks) {
     const auto& pc = it.second;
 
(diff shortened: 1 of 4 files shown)


jgfouca (dismissed):

components/scream/src/share/atm_process/atmosphere_process.hpp
```diff
@@ -220,20 +220,21 @@ class AtmosphereProcess : public ekat::enable_shared_from_this<AtmosphereProcess
...
-  template<typename FPC, CheckFailHandling CFH = CheckFailHandling::Fatal, typename... Args>
+  template<typename FPC, typename... Args>
   void add_precondition_check (const Args... args);
-  template<typename FPC, CheckFailHandling CFH = CheckFailHandling::Fatal, typename... Args>
+  template<typename FPC, typename... Args>
   void add_postcondition_check (const Args... args);
-  template<typename FPC, CheckFailHandling CFH = CheckFailHandling::Fatal, typename... Args>
+  template<typename FPC, typename... Args>
   void add_invariant_check (const Args... args);
+  template<typename FPC, typename... Args>
```
tcclevenger, line 237: This line should not be here.

components/scream/src/share/atm_process/atmosphere_process.cpp
```diff
@@ -450,6 +468,49 @@ get_internal_field(const std::string& field_name) const {
...
+  // update a field, without that appearing in the dag.
+  for (const auto& ptr : pc->repairable_fields()) {
+    const auto& fid = ptr->get_header().get_identifier();
+    EKAT_REQUIRE_MSG (
+        has_computed_field(fid) || has_computed_group(fid.name(),fid.get_grid_name()),
+        "Error! Input property check can repair a non-computed field.\n"
+        "  - Atmosphere process name: " + name() + "\n"
+        "  - Property check name: " + name() + "\n");
+  }
+  m_precondition_checks.push_back(std::make_pair(pc,cfh));
```
tcclevenger, line 493: This pair (and one below) is backwards. Should be `(cfh,pc)`. Maybe the structure of `m_precondition_checks` and `add_precondition_check (pc, cfh)` should match.

components/scream/src/share/atm_process/atmosphere_process.cpp
```diff
@@ -450,6 +468,49 @@ get_internal_field(const std::string& field_name) const {
...
+  // update a field, without that appearing in the dag.
+  for (const auto& ptr : pc->repairable_fields()) {
+    const auto& fid = ptr->get_header().get_identifier();
+    EKAT_REQUIRE_MSG (
+        has_computed_field(fid) || has_computed_group(fid.name(),fid.get_grid_name()),
+        "Error! Input property check can repair a non-computed field.\n"
+        "  - Atmosphere process name: " + name() + "\n"
+        "  - Property check name: " + name() + "\n");
+  }
+  m_postcondition_checks.push_back(std::make_pair(pc,cfh));
```
tcclevenger, line 511: Same as above.

tcclevenger (commented):
The only other thing keeping me from compiling is line 540-544 in `src/share/tests/atm_process_test.cpp` need to be updated (remove the template and add as 2nd arg).

jgfouca (approved):

PR #1912, opened 2022-08-31
Add ne1024pg2_ne1024pg2 grid to have ocnice on the same grid as atm

wlin7:
lnd is also on ne1024pg2 grid while using ICOS10 as mask.
ICOS10 mask is the highest resolution available. With this mask,
domain files and elm input files (fsurdat and finidat) for ne1024pg2_ICOS10 
can be used for this new grid.

### DIFF
--- a/cime_config/config_grids.xml
+++ b/cime_config/config_grids.xml
@@ -1517,6 +1517,16 @@
       <mask>ICOS10</mask>
     </model_grid>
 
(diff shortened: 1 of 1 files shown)


AaronDonahue (approved):
Probably need Ben to really give final approval.

brhillman (approved):

PR #1913, opened 2022-08-31
On pm-cpu and pm-gpu, use CFS as output dir temporarily

ndkeen:
On pm-cpu and pm-gpu, use CFS as output dir instead of scratch as a temporary fix until scratch is back on PM

[bfb]


ndkeen:
Output will default to 
```
/global/cfs/cdirs/e3sm/e3sm_scratch/pm-cpu/
and
/global/cfs/cdirs/e3sm/e3sm_scratch/pm-gpu/
```
which may confuse testing?

AaronDonahue (dismissed):
This will probably conflict with any upstream merges.  Or at least be different.  But since we plan on only making this temporary and reverting to the original setup in a week or so it should be alright.

As for confusing testing... @jgfouca would have to answer that one.

ndkeen:
Actually, maybe I should add a $USER in there so it's not all in same dir?

ndkeen:
OK I pushed a commit, now the output dirs will be (for user ndk):
```
/global/cfs/cdirs/e3sm/ndk/e3sm_scratch/pm-cpu/
and
/global/cfs/cdirs/e3sm/ndk/e3sm_scratch/pm-gpu/
```

ndkeen:
Can this be merged?

jeff-cohere (approved):

PR #1914, opened 2022-09-01
Link YAKL against MPI if HAVE_MPI is defined

### LINKED ISSUE #1905: Build error on pm-gpu after PR to use CIME flags
After `[#1895 Propagate flags set by CIME to scream](https://github.com/E3SM-Project/scream/pull/1895)`, a test worked (built/ran) on pm-cpu, but failed on pm-gpu.

```
cd /global/cfs/cdirs/e3sm/ndk/e3sm_scratch/pm-gpu/se20-aug29/f30.F2010-SCREAMv1.ne30_ne30.se20-aug29.gnugpu.5d.n004a4xX14444/bld/cmake-bld/scream/src/dynamics/homme/homme/composec++ && /global/cfs/cdirs/e3sm/ndk/se20-aug29/externals/ekat/extern/kokkos/bin/kokkos_launch_com$
In file included from /global/cfs/cdirs/e3sm/ndk/se20-aug29/externals/YAKL/src/YAKL.h:4,
                 from /global/cfs/cdirs/e3sm/ndk/se20-aug29/externals/YAKL/src/YAKL.cpp:2:
/global/cfs/cdirs/e3sm/ndk/se20-aug29/externals/YAKL/src/YAKL_header.h:26:10: fatal error: mpi.h: No such file or directory
   26 | #include <mpi.h>
      |          ^~~~~~~
compilation terminated.
```

I see that for the compilation of YAKL.cpp, there are some flags used in previous repo, that are not here now. But I'm not yet sure if that's actually the issue or not. For example, I don't see these flags thrown for YAKL.cpp: `-DGPU -DHAVE_MPI -DLinux -DMCT_INTERFACE  -DNO_R16 -D_PNETCDF` in older version.

Note also that YAKL.cpp is the only file built with nvcc.

One hack that allows the build to complete for me is to change 2 sources to force not having MPI. We probably need to do something like: change the compiler used for that YAKL.cpp, change only the flags passed to that one file, allow HAVE_MPI by also adding a `-I`, etc

bartgol:
If HAVE_MPI is defined, YAKL uses mpi.h, but does not link against it. Therefore, if that macro is available to YAKL, we must ensure we provide MPI include/link stuff to the yakl target.

Fix #1905 .

@ndkeen this fixes it for me on pm-gpu.

### DIFF
--- a/components/cmake/build_scream.cmake
+++ b/components/cmake/build_scream.cmake
@@ -12,7 +12,16 @@ function(build_scream)
     set (CMAKE_C_FLAGS ${CFLAGS})
     set (CMAKE_CXX_FLAGS ${CXXFLAGS})
     set (CMAKE_Fortran_FLAGS ${FFLAGS})
-    add_compile_definitions (${CPPDEFS})
+
+    # Strip leading space, or else add_compile_definitions will add
+    # an empty definition '-D `, which will cause compiler errors
+    # string (REGEX REPLACE "^ " "" CPPDEFS "${CPPDEFS}")
+    # add_compile_definitions expects a list of definitinos, each
+    # one provided without -D. So 1) replace ' ' with ';', and
+    # 2) strip '-D'/
+    string (REPLACE " " ";" CPPDEFS "${CPPDEFS}")
+    string (REPLACE "-D" "" CPPDEFS "${CPPDEFS}")
+    add_compile_definitions ("${CPPDEFS}")
 
     # Include machine file here
     include(${CMAKE_SOURCE_DIR}/scream/cmake/machine-files/${MACH}.cmake)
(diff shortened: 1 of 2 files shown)


jgfouca (approved):

ndkeen:
It looks like you already tested it on pm-gpu, but confirming it is building for me. Does it make sense why it was building elsewhere? I would think we would see HAVE_MPI ... At some point, I think Matt suggested that I would need HAVE_MPI to get yakl profiling data, but I might be getting something mixed up. Certainly I didn't specifically add HAVE_MPI for PM.

bartgol:
Yeah, all machines should now see HAVE_MPI, which is added by E3SM build system.

I don't have a definite answer. I wonder whether it has to do with compiler wrappers. On other machines, they are mpicxx/mpicc/mpifort, which automatically add mpi include paths and link flags/libs to the compile/link line. Maybe on pm these paths/flags are not automatically added, and one needs to manually do it? Seems unlikely, but it something I can check.

PR #1915, opened 2022-09-01
Fix quartz building of kokkos

jgfouca:
Do not use KOKKOS_OPTIONS in E3SM macros, use SCREAM macros
to configure Kokkos with non-deprecated options.


AaronDonahue (approved):
Thanks for getting on this so quickly @jgfouca

PR #1918, opened 2022-09-02
SHOC: Add a missing team_barrier.

### LINKED ISSUE #1917: Nondeterministic COMPARE_base_* FAILs on Ascent and Summit.
This issue is to track the nondeterministic FAILs that have appeared in the Ascent e3sm_scream_v1_master_gnugpu dashboard entry for ERP, ERS, and PEM tests from the first day we had v1 nightlies running on GPU machines.

ambrad:
From a comment in the commit:

  A note on team_barrier after linear_interp: One does not need a barrier
  between two parallel_for loops with the same index set: e.g., two
  [0,nlev_pack) parallel_for ranges. But one does if e.g. a linear_interp is
  over nlevi_pack and the next parallel_for is over nlev_pack.

Partially or fully fixes #1917. There's another bug in the dycore that affects ftype=0 runs, but it's not yet clear to me whether it's partly responsible for the nondeterministic fails.

[BFB] except for addressing the nondeterministic fails

### DIFF
--- a/components/scream/src/physics/shoc/shoc_update_prognostics_implicit_impl.hpp
+++ b/components/scream/src/physics/shoc/shoc_update_prognostics_implicit_impl.hpp
@@ -91,6 +91,11 @@ void Functions<S,D>::update_prognostics_implicit(
   linear_interp(team,zt_grid,zi_grid,tkh,tkh_zi,nlev,nlevi,0);
   linear_interp(team,zt_grid,zi_grid,tk,tk_zi,nlev,nlevi,0);
   linear_interp(team,zt_grid,zi_grid,rho_zt,rho_zi,nlev,nlevi,0);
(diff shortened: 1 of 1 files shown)


tcclevenger (approved):

tcclevenger:
This was a common issue in the shoc conversion when using liner_interp. I wonder if it would be smart for me to run through all the calls and just see if they could be potentially affected.

ambrad:
@tcclevenger sure, go ahead; it's always better to have more eyes. But I checked them all and found only this one. Once I saw that SHOC was a problem, I visually read through all of it and found this one. I then also instrumented the code with exclusive-or hashing and again pinpointed this one missing barrier and found nothing else.

ambrad:
Rather than waste the AT resources on a team_barrier, I'm going to manually merge this PR now.

PR #1921, opened 2022-09-06
SHOC: Add some team_barriers that resolve a problem but don't make sense.

ambrad:
From a comment in the commit:

  The team_barriers protect what we think is unexpected behavior in
  Kokkos::parallel_reduce. We expect not to need these based on the semantics
  of parallel_reduce. However, we speculate that in the Cuda implementation
  of the team parallel_reduce, it's possible something in the bookkeeping at
  the end of one parallel_reduce may sometimes be affecting the start of the
  next, leading to nondeterminism. Then again, there might be something
  subtly wrong in our code, and we're just not seeing it. In any case, these
  team_barriers fix nondeterminism that was repeatedly and with high
  confidence traced to these energy integral computations. We checked that
  the view_reduction wrapper is not the cause by simplifying these to be bare
  Kokkos::parallel_reduce calls acting on doubles and saw the same results.

[BFB] except possibly finally fully fixing #1917.

### DIFF
--- a/components/scream/src/physics/shoc/shoc_energy_integrals_impl.hpp
+++ b/components/scream/src/physics/shoc/shoc_energy_integrals_impl.hpp
@@ -26,29 +26,45 @@ void Functions<S,D>
   using ExeSpaceUtils = ekat::ExeSpaceUtils<typename KT::ExeSpace>;
   const auto ggr = C::gravit;
 
+  // The team_barriers protect what we think is unexpected behavior in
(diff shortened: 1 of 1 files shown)


ambrad:
One noteworthy test: I added a team_barrier at the start of this routine, then removed all but the last one, thus sandwiching the reductions between barriers. Repeat testing produced a failure. Thus, a barrier between immediately adjacent reductions is evidently necessary.

PR #1923, opened 2022-09-06
Another try at fixing quartz kokkos

jgfouca:


components/cmake/build_scream.cmake
```diff
@@ -9,10 +9,18 @@ function(build_scream)
...
+    # Transfer CIME-set flags into Cmake-style
     set (CMAKE_C_FLAGS ${CFLAGS})
     set (CMAKE_CXX_FLAGS ${CXXFLAGS})
     set (CMAKE_Fortran_FLAGS ${FFLAGS})
 
+    # SCREAM manages its own kokkos settings. We can think about
+    # removing this once the cime cmake_macros aren't using obsolete
+    # Kokkos settings like KOKKOS_OPTIONS.
+    if (KOKKOS_OPTIONS)
+      unset(KOKKOS_OPTIONS)
```
bartgol, line 21: I'm confused. The CIME mach file for quartz sets this var, but then we unset it? Is it not used then? If so, why do we need this var?

AaronDonahue:
I tested this branch on quartz and it went back to being able to build SCREAMv0 and SCREAMv1.  So I can confirm it accomplishes what it sets out to do w.r.t. quartz.

AaronDonahue (approved):

PR #1925, opened 2022-09-07
EAMxx: fix shoc standalone output yaml file

bartgol:
The file was using some old syntax, which caused no field to be saved.


AaronDonahue (approved):

bartgol:
Argh, this should have not been merged. I thought I pushed the last fix, but didn't. The shoc test continues to produce no fields in the output file.

PR #1926, opened 2022-09-07
Revert "Merge Pull Request #1880 from E3SM-Project/scream/bartgol/model-output-precision`

bartgol:
Revert the PR, to avoid performance hits in production runs, while we figure out how to avoid local copies double->float in our scorpio layer.


AaronDonahue (approved):

AaronDonahue:
Once this finishes the AT I think we can go ahead and manually merge.

PR #1927, opened 2022-09-08
Fix IO classes handling of grids with alias names

bartgol:


components/scream/src/share/io/scorpio_output.cpp
```diff
@@ -64,14 +64,17 @@ AtmosphereOutput (const ekat::Comm& comm, const ekat::ParameterList& params,
...
+    for (const auto& grid_name : io_grid_aliases) {
+      if (f_pl.isSublist(grid_name)) {
+        const auto& pl = f_pl.sublist(grid_name);
+        m_fields_names = pl.get<vos_t>("Field Names");
+
+        // Check if the user wants to remap fields on a different grid first
+        if (pl.isParameter("IO Grid Name")) {
+          io_grid = grids_mgr->get_grid(pl.get<std::string>("IO Grid Name"));
+        }
+        break;
```
AaronDonahue, line 77: Do we want to throw an error if we never get to the break statement? I.e if the grid doesn't match any of the alias'

AaronDonahue (commented):
I just had one question about the loop where we check alias' against the io parameter list.

components/scream/src/share/io/scorpio_output.cpp
bartgol: Yeah, that's a good idea.

AaronDonahue:
side question, do we expect any conflicts with the SP IO PR?  If so, we should probably be careful with the order we merge the two.

AaronDonahue (approved):

PR #1928, opened 2022-09-08
Add subcolumn sampling for cloud overlap in radiation

brhillman:
Add subcolumn sampling for cloud overlap in radiation interface.


AaronDonahue:
@brhillman , just wanted to check the status on this PR, is it ready to review?

bartgol (dismissed):

bartgol (approved):

brhillman:
Example column from a ne4 sim demonstrating subcolumns:
subcols_example
This shows, on the left, the subcolumn mask (derived from a newly added output, cld_tau_sw_gpt, which has dimensions ncol, nlev, nswgpt; the nswgpt dimension is the subcol dimension, always matching the number of quadrature points in the radiation), in the middle the cloud fraction from the `cldfrac_tot` field in the model and and cloud fraction derived by taking the domain average of the subcolumn mask (the comparison showing that the subcolumn mask approximately reproduces the domain mean cloud fraction input to this column), and on the right the cloud water and ice mixing ratios. 

This example demonstrates a few key points. First, the left shows what the subcolumns actually look like for this column. This initial approach here was to implement maximum-random overlap, consistent with what EAM/SCREAMv0 use. This means that vertically-contiguous cloudy layers will have their cloudy subcolumns maximally-overlapped, but layers separated by at least one clear layer are randomly overlapped. This is evident by comparing the middle partially cloudy layer with the lower-level, barely cloudy layers. The middle layer has a number of contiguous model levels that contain cloud, and the cloudy parts of these subcolumns all line up. But the few layers below this with very small cloud fractions are separated from these layers, and thus do not line up with the middle band of cloudy layers.

The second point this example demonstrates is that the domain mean cloud fraction is (approximately) reproduced by the subcolumn mask. Meaning, if you take the mean of the binary mask over all of the subcolumns (the orange curve), you should arrive back at the cloud fraction the code was given as input (the blue curve). The cloud fractions are not guaranteed to always match up perfectly, because a) we are reproducing a continuous real number with the mean of a discrete number of integers, and b) the subcolumn cloud mask is generated by comparing a set of psuedo-random numbers to the cloud fraction, and thus will not be guaranteed to always exactly reproduce the cloud fraction (but should converge in the limit of many samples).

brhillman:
Some performance on summit, for a 7-day ne30 case with no output, with gnugpu on 5 nodes, 6 MPI tasks per node, with current master:
```
"a:EAMxx::Radiation::run"                            -         30       30 1.008000e+04   2.673618e+02     9.895 (    24      0)     8.054 (    11      0)
```
and with this branch with subcolumn sampling on:
```
"a:EAMxx::Radiation::run"                            -         30       30 1.008000e+04   2.874769e+02    10.606 (    24      0)     8.704 (    11      0)
```

bartgol:
> Some performance on summit, for a 7-day ne30 case with no output, with gnugpu on 5 nodes, 6 MPI tasks per node, with current master:
> 
> ```
> "a:EAMxx::Radiation::run"                            -         30       30 1.008000e+04   2.673618e+02     9.895 (    24      0)     8.054 (    11      0)
> ```
> 
> and with this branch with subcolumn sampling on:
> 
> ```
> "a:EAMxx::Radiation::run"                            -         30       30 1.008000e+04   2.874769e+02    10.606 (    24      0)     8.704 (    11      0)
> ```

I thiought subsampling was supposed to be faster. Is that not the case?

brhillman:
> I thiought subsampling was supposed to be faster. Is that not the case?

No, it will alway be a little slower. There's the same number of integration points in the low-level rad calcs, the difference in using subcolumns or not is that with subcolumns, we account for the fact that at 3 km scale, you'd expect part of each column to be cloudy and part of it to be clear, and also how the cloudy parts of cloudy layers would overlap vertically. There's no more or less computation being done, the little bit of slowdown is because we have generate the subcolumns themselves.

Getting into the weeds a little bit, the way we implement subcolumns uses a trick that makes it far faster than other approaches to subcolumns would be, in that we match up a single subcolumn with a single integration point, as opposed to running the full radiation calculations on each subcolumn. That's getting into the weeds a little bit, but maybe that's where the confusion arises.

components/scream/src/physics/rrtmgp/scream_rrtmgp_interface.cpp
```diff
@@ -304,13 +328,175 @@ namespace scream {
...
+            subsampled_optics.alloc_2str(ncol, nlay);
+            // Check that we do not have clouds with no optical properties; this would get corrected
+            // when we assign optical props, but we want to use a "radiative cloud fraction"
+            // for the subcolumn sampling too because otherwise we can get vertically-contiguous cloud
+            // mask profiles with no actual cloud properties in between, which would just further overestimate
+            // the vertical correlation of cloudy layers. I.e., cloudy layers might look maximally overlapped
+            // even when separated by layers with no cloud properties, when in fact those layers should be
+            // randomly overlapped.
+            auto cldfrac_rad = real2d("cldfrac_rad", ncol, nlay);
+            memset(cldfrac_rad, 0.0);  // Start with all zeros
```
bartgol, line 416: Curious (as it's not a perf issue): isn't yakl already init-ing to 0? Or is it leaving memory uninitialized?

bartgol (approved):

components/scream/src/physics/rrtmgp/scream_rrtmgp_interface.cpp
brhillman: Good question. In reading through the YAKL source, it looks like "no", the constructor looks like it just allocates the memory and sets up the dims. But I could be missing something.

bartgol: It's not important, I was just wondering (since Kokkos always inits to 0, unless you use a special constructor that leaves mem uninited). Thanks.

brhillman:
A little more analysis. I ran three ne30 cases, one with v0 configured to use a cloud optics scheme more consistent with what we are using in v1, one with this branch with subcolumn sampling enabled, and one with this branch with subcolumn sampling disabled. I computed what I'm calling a column "cloud effect" on SW and LW up and down fluxes, which is computed just by taking, e.g. for SW up, the differences in all-sky and clear-sky (no clouds) SW upwelling radiative flux. After the first timestep (step 0 for v1, step 1 for v0 since v0 writes out t=0 first):
SW_cloud_effect_up_subcolumns_vs_v0_ne30
SW_cloud_effect_dn_subcolumns_vs_v0_ne30
LW_cloud_effect_up_subcolumns_vs_v0_ne30
LW_cloud_effect_dn_subcolumns_vs_v0_ne30
The comparison shows that the effect of clouds on the shortwave upwelling flux is *lower* with subcolumn sampling than without. This means that clouds appear less bright overall in v1 with subcolumn sampling enabled. I think this is consistent with our expectations; without subcolumns we assume that each column is either completely clear or completely cloudy (if there's any cloud at all, it's completely cloudy), but with subcolumns we allow for both cloudy *and* clear parts of each column, thus partially cloudy scenes (e.g., columns/layers with cloud fraction < 1) will have at least some clear (non-reflective) subcolumns. v1 without subcolumns appears to have overestimated the reflectiveness of cloudy columns relative to v0, while the addition of subcolumns makes the difference *negative* (clouds less reflective than in v0). This is a difficult comparison to make though, because these two cases actually have different amounts of globally-averaged condensed water, with v0 have MORE condensed water than v1 at this point:
qc_subcolumns_vs_v0_ne30
More condensed cloud water in v0 would tend to increase the cloud effect as well, so the underestimate relative to v0 in cloud effect in v1 with subcolumns should not be taken as alarming.

Likewise, the cloud effect on LW fluxes is generally smaller as well, consistent with more clear columns.

AaronDonahue (approved):

AaronDonahue:
Just to be on the safe side I am going to set the AT to retest and then manually merge when it finishes.

AaronDonahue (approved):

AaronDonahue:
AT finished, DIFFs are expected.  I am going to manually merge.

PR #1930, opened 2022-09-08
EAMxx: fix buggy logic in npX vs np1 tests

bartgol:
We were not testing npX vs np1 at all, due to a backward comparison in our cmake logic.


AaronDonahue (approved):
I'm surprised we were not testing np1 vs. npX.  I thought I often saw things like BLAH_np1_vs_BLAH_npX in our set of tests.

bartgol:
Nope, we were not testing them. :/

https://my.cdash.org/viewTest.php?onlypassed&buildid=2213735

jgfouca:
@bartgol , would it be possible to add something to scripts-test that would have caught this?

PR #1931, opened 2022-09-08
Re-instate single precision model output

bartgol:
I reverted the revert of the single precision PR, and added fixes that remove the temporary arrays copies.

### DIFF
--- a/components/scream/src/share/io/scream_output_manager.cpp
+++ b/components/scream/src/share/io/scream_output_manager.cpp
@@ -146,16 +146,10 @@ setup (const ekat::Comm& io_comm, const ekat::ParameterList& params,
       if (has_restart_data && m_output_control.nsamples_since_last_write>0) {
         auto output_restart_filename = find_filename_in_rpointer(hist_restart_casename,false,m_io_comm,m_run_t0);
 
-        ekat::ParameterList res_params("Input Parameters");
-        res_params.set<std::string>("Filename",output_restart_filename);
-        AtmosphereInput output_restart (m_io_comm,res_params);
-
         // Also restart each stream
         for (auto stream : m_output_streams) {
           stream->restart(output_restart_filename);
         }
-
-        output_restart.finalize();
       }
     }
   }
@@ -215,16 +209,20 @@ void OutputManager::run(const util::TimeStamp& timestamp)
 
       // Register time as a variable.
       auto time_units="days since " + m_case_t0.get_date_string() + " " + m_case_t0.get_time_string();
-      register_variable(filename,"time","time",time_units,1,{"time"},  PIO_REAL,"time");
+      register_variable(filename,"time","time",time_units,1,{"time"}, scorpio::nctype("double"),"time");
 #ifdef SCREAM_HAS_LEAP_YEAR
       set_variable_metadata (filename,"time","calendar","gregorian");
 #else
       set_variable_metadata (filename,"time","calendar","noleap");
 #endif
 
+      std::string fp_precision = is_checkpoint_step
+                               ? "real"
+                               : m_params.get<std::string>("Floating Point Precision");
+
       // Make all output streams register their dims/vars
       for (auto& it : m_output_streams) {
-        it->setup_output_file(filename);
+        it->setup_output_file(filename,fp_precision);
       }
 
       // Set degree of freedom for "time"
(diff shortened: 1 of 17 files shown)


ambrad:
I recommend running PEM_Ln90.ne30pg2_ne30pg2.F2010-SCREAMv1 (or an equivalent create_newcase) on Summit without and with this PR to verify the current performance is unaffected. You could also do a third run with single-precision output to check the performance change in that case.

AaronDonahue:
@bartgol , two questions.  

Did you run the test that @ambrad suggested above?  If so, did it confirm that this PR doesn't noticeably slow down the run?

Looking at the commits, I don't see one where you remove the array copies that slowed things down in the first place.  Could you point me to that specific commit in the commit history.

Everything looks good, and I like the cleanup of not passing both the vec of dims and it's size when we don't need to.

ambrad:
> Did you run the test that @ambrad suggested above? If so, did it confirm that this PR doesn't noticeably slow down the run?

Thanks for the reminder; I meant to updated my request above but forgot. The dashboards are back to normal; in particular, ascent was back to normal the first time it reported after the original PR was reverted. So no need to check.

bartgol:
> Did you run the test that @ambrad suggested above? If so, did it confirm that this PR doesn't noticeably slow down the run?
I have, but the results are mixed, so I'm rerunning everything.

> Looking at the commits, I don't see one where you remove the array copies that slowed things down in the first place. Could you point me to that specific commit in the commit history.
[37bea2b51b2cee11a5c206c3f2e68e013e0a9b7c](https://github.com/E3SM-Project/scream/commit/37bea2b51b2cee11a5c206c3f2e68e013e0a9b7c)

ambrad:
@bartgol you might have better luck getting a good signal by looking at the CPL:RUN_LOOP and CPL_ATM_RUN timers in the timer summary files. Note also that if you're using a particular test to look, e.g., PEM_Ln90.ne30pg2_ne30pg2, you can make code changes and then `cd bld/cmake-bld; make -j8` and then `cd ../../; ./case.submit` to rerun a test w/o making a new test case.

components/scream/src/share/io/scream_output_manager.cpp
```diff
@@ -392,6 +392,12 @@ set_params (const ekat::ParameterList& params,
 
     m_output_file_specs.max_snapshots_in_file = m_params.get<int>("Max Snapshots Per File");
     m_casename = m_params.get<std::string>("Casename");
+
+    // Allow user to ask for higher precision for normal model output,
+    // but default to single to save on storage
```
AaronDonahue, line 397: Since this line still applies, why does this pass baseline comparison.  Shouldn't the new output all be SP?

AaronDonahue (commented):
Thanks @bartgol for showing me where the extra copy was removed.  I'm happy to approve this PR, although it is set to Automerge and has passed all tests, which is the source of one last question I have.

bartgol:
Ok, I ran a SMS test to be faster, and timers are roughly the same:

<table>
<tr>
	<td> timer
	<td> master
	<td> branch (DP)
	<td> branch (SP)
<tr>
	<td> CPL:RUN_LOOP
	<td> 49.137
	<td> 47.483
	<td> 48.600
<tr>
	<td> CPL:ATM_RUN
	<td> 46.838
	<td> 45.055
	<td> 46.161
</table>
where, of course, the output files are half the size. There is still a small uptick when going from DP to SP, but I doubt it has anything to do with our code. Perhaps the discrepancy between input-array/nc-file data types causes scorpio to do some extra work to do the conversion (I don't think that's the case though). However, I would have to run both cases a few more times to get a more reliable signal, since also other timers went up a couple of % points (e.g., LND_RUN). However, it is fair to say that the branch is unlikely to hinder master's performance.

components/scream/src/share/io/scream_output_manager.cpp
bartgol: Good question. I don't know.

bartgol: Perhaps cprnc casts everything to whatever the worst precision is. Or maybe it uses the precision of the 1st file (which happens to be from this PR)

bartgol: AH, indeed! Running cprnc with the files swapped causes the test to fail!

bartgol:
@AaronDonahue Given that tests _do_ diff (but cprnc is not called symmetrically by CIME), and that timers are roughly as in master, I believe we can merge this.

AaronDonahue:
@bartgol i saw your message on Slack. Interesting behavior. Thanks for manually confirming the diffs. I agree we can merge this now

AaronDonahue (approved):

PR #1936, opened 2022-09-12
Add cloning capability to AbstractGrid

bartgol:
Together with `reset_num_vertical_lev()`, one can use the `clone` method to create a grid object with a different number of levels (to be used, e.g., for I/O regridding).

### DIFF
--- a/components/scream/src/share/tests/grid_tests.cpp
+++ b/components/scream/src/share/tests/grid_tests.cpp
@@ -55,6 +55,9 @@ TEST_CASE("point_grid", "") {
   for (int i=0; i<grid->get_num_local_dofs(); ++i) {
     REQUIRE (deep_copy->get_dofs_gids_host()[i]==grid->get_dofs_gids_host()[i]);
   }
+
+  shallow_copy->reset_num_vertical_lev(4);
+  REQUIRE (shallow_copy->get_num_vertical_levels()==4);
 }
 
 TEST_CASE("se_grid", "") {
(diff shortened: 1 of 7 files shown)


components/scream/src/share/tests/grid_tests.cpp
```diff
@@ -46,6 +46,18 @@ TEST_CASE("point_grid", "") {
   auto layout = grid->get_2d_scalar_layout();
   REQUIRE(layout.tags().size() == 1);
   REQUIRE(layout.tag(0) == COL);
+
+  auto shallow_copy = grid->clone("shallow",true);
+  auto deep_copy    = grid->clone("shallow",false);
```
AaronDonahue, line 51: Maybe change the name to "deep"

components/scream/src/share/tests/grid_tests.cpp
```diff
@@ -77,6 +89,15 @@ TEST_CASE("se_grid", "") {
   const auto max_gid = se_grid->get_global_max_dof_gid();
   const auto min_gid = se_grid->get_global_min_dof_gid();
   REQUIRE( (max_gid-min_gid+1)==se_grid->get_num_global_dofs() );
+
+  auto shallow_copy = se_grid->clone("shallow",true);
+  auto deep_copy    = se_grid->clone("shallow",false);
```
AaronDonahue, line 94: Same here with grid name

AaronDonahue (commented):
Looks good to me. Just two minor comments, not crucial. I'll pause accepting now in case you want to make the changes, so we don't automerge

components/scream/src/share/tests/grid_tests.cpp
bartgol: Ah, classic copy-paste...

AaronDonahue (approved):

PR #1937, opened 2022-09-12
Add general vertical interpolation utility

rebassoo:
This PR is meant to create the general utility to interpolate a variable from a set of source levels to a set of target levels. It can be used for instance to interpolate temperature (or another field) from source pressure levels to target pressure levels (which could be for instance constant across the globe). However, it is general enough that you could for instance interpolate a variable from source geometric heights to target geometric heights. The utility also provides the ability to return a mask of target levels that are outside the range of the source levels. In the case where the target level is outside the range of the source levels the default interpolated value is set to -9.99e17, but the function is overloaded to a user can provide a different value for this.

It also adds a test that reads in input files for pressure target levels, pressure source levels, temperature at pressure source levels, and then interpolates the temperature to the target pressure levels. Once doing so it compares the output to a log file with expected outputs for this interpolation and fails if it finds any differences. The test also checks that masking is done correctly for the case where the maximum level of the pressure target is above the maximum level of the pressure source.


ambrad:
One high-level comment: In E3SM we generally avoid making baseline files part of the repo. Is it possible to rework the testing so you test properties (things that should be true) rather than compare against a baseline file? Alternatively, if baselines are necessary, they should be generated by the test for subsequent comparison, which is how baseline testing is done in E3SM. For a low-level operation like remapping/interpolation, I recommend property checking rather than baseline comparison.

jgfouca (changes requested):
I agree with @ambrad , we cannot accept the addition of large baseline files into the repo.

components/scream/src/share/util/scream_vertical_interpolation.hpp
```diff
@@ -0,0 +1,77 @@
...
+using ESU = ekat::ExeSpaceUtils<ExeSpace>;
+using LIV = ekat::LinInterp<Real,Spack::n>;
+
+ 
+template <typename S>
+using view_1d = typename KT::template view_1d<S>;
+template <typename S>
+using view_2d = typename KT::template view_2d<S>;
+
+const Real masked_val = -9.99e17;
```
bartgol, line 37: This seems like an arbitrary number. If the goal is to put some invalid number, prefer using `ekat::ScalarTraits<Real>::invalid()`.

components/scream/src/share/util/scream_vertical_interpolation.hpp
```diff
@@ -0,0 +1,77 @@
...
+#include "ekat/ekat_pack_utils.hpp"
+#include "ekat/kokkos/ekat_subview_utils.hpp"
+
+namespace scream {
+
+//
+// ------- Types --------
+//
+
+template <typename S>
```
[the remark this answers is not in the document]
bartgol: On top of that, it seems like you don't even use all the names defined below (e.g., I don't see Sbool used).

components/scream/src/share/util/scream_vertical_interpolation.hpp
```diff
@@ -0,0 +1,77 @@
...
+
+ 
+template <typename S>
+using view_1d = typename KT::template view_1d<S>;
+template <typename S>
+using view_2d = typename KT::template view_2d<S>;
+
+const Real masked_val = -9.99e17;
+  
+void perform_vertical_interpolation (const view_2d<const Spack>&,
```
bartgol, line 39: I think these functions could be templated on scalar type and pack size, so that they could potentially be used even if some parametrization used a different pack size.

rebassoo:
This PR has been updated to remove any text files and perform the test with fake data.

components/scream/src/share/util/scream_vertical_interpolation.hpp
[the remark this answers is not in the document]
bartgol: Note: you could encapsulate names in a struct, and use that struct to access names later. Alternatively, as done in physics, you can encapsulate names and methods (making methods static) in a `VerticalInterpolation` struct.

components/scream/src/share/util/scream_vertical_interpolation.hpp
```diff
@@ -0,0 +1,77 @@
...
+				     const view_1d<const Spack>&,
+				     const view_2d<const Spack>&,
+				     const view_2d<      Spack>&,
```
bartgol, line 42: Headers are what people look at to figure out how to use a function. I would request that you name each variable, so that the users know what each of them is.

components/scream/src/share/util/scream_vertical_interpolation.hpp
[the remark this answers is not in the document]
rebassoo: Thanks for the comments. I have removed the Sbool (it was from a previous iteration of the code). With respect to the typedefs, would making an additional namespace, i.e. scream::vi be acceptable, or would the preference be to put this in a struct. Also, when looking in physics it looks like in atmosphere_microphysics.hpp the typedefs are at the scream namespace level, not sure if this an exception to the rule or just missed previously.

components/scream/src/share/util/scream_vertical_interpolation.hpp
rebassoo: In development this number was discussed with Peter and Aaron as being preferable to a nan when being able to analyze offline. Also, capability to be able to set it to a different number was requested, which is currently in the code. I can revisit this with them.

components/scream/src/share/util/scream_vertical_interpolation.hpp
rebassoo: Done, named all variables in the header.

components/scream/src/share/util/scream_vertical_interpolation.hpp
PeterCaldwell: Yeah, @bartgol - the idea was that we didn't want to have to store both 3d mask and 3d data for each variable in the netcdf output. it sounds like `ekat::ScalarTraits<Real>::invalid()` might be preferable, but I wonder if python analysis code would get confused by this? I guess my question is - what is this invalid() thing? Is it an actual number like -999? Or will it write NaN in the netcdf file? Or is it something else?

rebassoo: This was updated to be: -std::numeric_limits<Real>::max() since a nan is not preferred

components/scream/src/physics/p3/atmosphere_microphysics.hpp
```diff
@@ -11,6 +11,8 @@
 namespace scream
 {
 
+ using namespace p3;
```
bartgol, line 14: This should still stay inside the class declaration. No need to inject all of p3's names in the scream namespace.

components/scream/src/share/util/scream_vertical_interpolation.hpp
rebassoo: Have modified the functions to be templated on scalar type and pack size. To do source moved function implementation into a file scream_vertical_interpolation_impl.hpp and deleted scream_vertical_interpolation.cpp

components/scream/src/physics/p3/atmosphere_microphysics.hpp
rebassoo: Putting it inside the class declaration causes an error: "error: expected an identifier
    using namespace p3;"
To resolve this I removed the line "using namespace p3;" and just added the p3 namespace as a prefix to the needed lines:
line 24 of atmosphere_microphysics.hpp:
using P3F          = p3::Functions<Real, DefaultDevice>; 
and line 214 of atmosphere_microphysics.cpp:
p3::p3_init();

components/scream/src/share/util/scream_vertical_interpolation.hpp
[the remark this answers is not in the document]
rebassoo: With respect to the recommended tests proposed above. The vertical interpolation test has been moved to a new file vertical_interp_tests.cpp and several new tests were added to address the comments above. Integer inputs were already being used and were used for the new added tests. The original test checked that data out of bounds got the assigned masked value. Tests were added to confirm that interpolating on the same grid returns same data, interpolating on a subsampled grid returns the same data, interpolating on midpoints of the initial grid returns the average. In addition, a test was added that calls the 1D interpolator function and confirms that that is working properly. Lastly, tests were added to ensure things work in cases where the number of levels for the source/target are different than the pack size as well as if the pack size is different between the source and target

PeterCaldwell:
Is there a reason why this PR has stalled out? @rebassoo , @AaronDonahue , @bartgol ?

bartgol (changes requested):

components/scream/src/share/util/scream_vertical_interpolation.hpp
```diff
@@ -0,0 +1,104 @@
...
+    
+const Real masked_val = -std::numeric_limits<Real>::max();
+  
+template<typename T, int N> 
+void perform_vertical_interpolation(
+  const view_2d<Pack<T,N>>& x_src,
+  const view_1d<Pack<T,N>>& x_tgt,
+  const view_2d<Pack<T,N>>& input,
+  const view_2d<Pack<T,N>>& output,
+  const int& nlevs_src,
```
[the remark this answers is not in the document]
rebassoo: Have updated this to remove passed by reference for the ints.

components/scream/src/share/util/scream_vertical_interpolation.hpp
```diff
@@ -0,0 +1,104 @@
...
+//
+template<typename T, int N>
+using Pack = ekat::Pack<T, N>;
+
+template <typename S>
+using SmallPack = Pack<S,SCREAM_SMALL_PACK_SIZE>;
+using Spack = SmallPack<Real>;
+
+template<int N>
+using Pmask = ekat::Mask<N>;
```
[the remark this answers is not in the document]
rebassoo: I have changed Pmask to Mask.

components/scream/src/share/util/scream_vertical_interpolation.hpp
```diff
@@ -0,0 +1,104 @@
...
+namespace vinterp {
+  
+//
+// ------- Types --------
+//
+template<typename T, int N>
+using Pack = ekat::Pack<T, N>;
+
+template <typename S>
+using SmallPack = Pack<S,SCREAM_SMALL_PACK_SIZE>;
```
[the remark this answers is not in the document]
rebassoo: I have removed these from the header.

components/scream/src/share/util/scream_vertical_interpolation.hpp
```diff
@@ -0,0 +1,104 @@
...
+  const int& nlevs_src,
+  const int& nlevs_tgt);
+
+template<typename T, int N> 
+void perform_vertical_interpolation(
+  const view_2d<Pack<T,N>>& x_src,
+  const view_1d<Pack<T,N>>& x_tgt,
+  const view_2d<Pack<T,N>>& input,
+  const view_2d<Pack<T,N>>& output,
+  const view_2d<Pmask<N>>& mask,
```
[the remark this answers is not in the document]
rebassoo: I added several comments in the header file. A masked value means that you are out-of-bounds, i.e. it is an extrapolation not an interpolation

components/scream/src/share/util/scream_vertical_interpolation_impl.hpp
```diff
@@ -0,0 +1,139 @@
...
+    auto x1=ekat::subview(x_src, icol);
+    auto in=ekat::subview(input, icol);
+    auto out=ekat::subview(output, icol);
+    auto msk=ekat::subview(mask, icol);
+    
+    vert_interp.setup(team, x1, x_tgt);
+    vert_interp.lin_interp(team, x1, x_tgt, in, out, icol);
+    const auto x_src_s = ekat::scalarize(x1);
+    const auto x_tgt_s = ekat::scalarize(x_tgt);
+    const auto range_boundary = KT::RangePolicy(0, x_tgt.extent(0));
```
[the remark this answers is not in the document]
rebassoo: Changed to TeamThreadRange

components/scream/src/share/util/scream_vertical_interpolation_impl.hpp
```diff
@@ -0,0 +1,139 @@
...
+  const MemberType& team,
+  const LIV<T,N>& vert_interp)
+{
+  //Setup linear interpolation
+  vert_interp.setup(team, x_src, x_tgt);
+  //Run linear interpolation
+  vert_interp.lin_interp(team, x_src, x_tgt, input, output, icol);
+  const auto x_src_s = ekat::scalarize(x_src);
+  const auto x_tgt_s = ekat::scalarize(x_tgt);
+  const auto range_boundary = KT::RangePolicy(0, x_tgt.extent(0));
```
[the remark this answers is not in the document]
rebassoo: Changed to TeamThreadRange

components/scream/src/share/tests/vertical_interp_tests.cpp
```diff
@@ -0,0 +1,309 @@
...
+#include "share/util/scream_vertical_interpolation.hpp"
+
+TEST_CASE("main_vertical_interpolation_test"){
+
+  using namespace scream;
+  using namespace vinterp;
+  const int n_layers_src = 3;
+  const int n_layers_tgt = 4;
+  
+  auto npacks_src = ekat::PackInfo<Spack::n>::num_packs(n_layers_src);
```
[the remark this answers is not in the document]
rebassoo: To resolve this I changed the pack size to 1 so that it would run with more than one pack

components/scream/src/share/util/scream_vertical_interpolation.hpp
```diff
@@ -0,0 +1,104 @@
...
+template <typename S>
+using view_2d = typename KT::template view_2d<S>;
+    
+const Real masked_val = -std::numeric_limits<Real>::max();
+  
+template<typename T, int N> 
+void perform_vertical_interpolation(
+  const view_2d<Pack<T,N>>& x_src,
+  const view_1d<Pack<T,N>>& x_tgt,
+  const view_2d<Pack<T,N>>& input,
```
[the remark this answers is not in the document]
rebassoo: It gives an error when trying to do `const view_2d<const Pack<T,N>>& input`: 
error: no instance of overloaded function "perform_vertical_interpolation" matches the argument list
To resolve this did away with the input view types completely and templated on it:
template<typename Src, typename Tgt, typename Input, typename T, int N>
void perform_vertical_interpolation(
  const Src& x_src,
  const Tgt& x_tgt,
  const Input& input,
  const view_2d<Pack<T,N>>& output,
  const int nlevs_src,
  const int nlevs_tgt);

components/scream/src/share/util/scream_vertical_interpolation_impl.hpp
```diff
@@ -0,0 +1,138 @@
...
+      out(k).set(combined_m,msk_val);
+    });
+    Kokkos::fence();
+
+  });
+  Kokkos::fence();   
+}
+
+template<typename Src, typename Tgt, typename Input, typename T, int N> 
+void perform_vertical_interpolation_impl_1d(
```
bartgol, line 105: Since this is a function that is meant to be called _inside_ a parallel loop, it needs to be marked with the `KOKKOS_FUNCTION` macro.

components/scream/src/share/util/scream_vertical_interpolation.hpp
```diff
@@ -0,0 +1,126 @@
...
+  const Tgt& x_tgt,
+  const Input& input,
+  const view_2d<Pack<T,N>>& output,
+  const view_2d<Mask<N>>& mask,
+  const int nlevs_src,
+  const int nlevs_tgt,
+  const Real& msk_val);
+
+template<typename Src, typename Tgt, typename Input, typename T, int N> 
+void perform_vertical_interpolation_impl_1d(
```
bartgol, line 108: Since this is a function that is meant to be called inside a parallel loop, it needs to be marked with the `KOKKOS_FUNCTION` macro.

bartgol (changes requested):

components/scream/src/share/tests/vertical_interp_tests.cpp
```diff
@@ -0,0 +1,315 @@
...
+  //For each scenario target levels are at the midpoint and so should be the average
+  //of the source layers. 
+  //However, the last target level is not at a midpoint. 
+  //For scenario 1) it is at the last source level
+  //For scenario 2) and 3) it is at the 2nd to last source level
+  //For scenario 4) it is at the 3rd to last source level
+  //using namespace scream;
+  //using namespace vinterp;
+
+  const int N = SCREAM_SMALL_PACK_SIZE;
```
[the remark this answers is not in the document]
rebassoo: I have implemented a "run" function, and in cases of pack size =1 it will just run once. In cases of pack size > 1 it will run with pack size of 1 and whatever the pack size is defined to be. I have kept the first TEST_CASE because it tests aspects that are not in this test, for instance the masking and whether calling the 1D interpolation function gives you the same answer. I have updated the first TEST_CASE to have more target levels so that >1 pack will be used, and I also updated the first TEST_CASE to have pack size = SCREAM_PACK_SIZE.

components/scream/src/share/tests/vertical_interp_tests.cpp
```diff
@@ -0,0 +1,315 @@
...
+				   mask,
+				   n_layers_src[i],
+				   n_layers_tgt[i]);
+
+    for(int col=0; col<2; col++){
+      for(int lev=0; lev<(n_layers_tgt[i]-1); lev++){
+        auto avg_val = (tmp_src_s(col,lev+1)+tmp_src_s(col,lev))/2.;
+        if (lev < (n_layers_tgt[i]-1)){
+          //EKAT_REQUIRE_MSG(out_s(col,lev) == avg_val,"Failed on test: "+std::to_string(i)+"\n");
+          EKAT_REQUIRE_MSG(out_s(col,lev) == avg_val,
```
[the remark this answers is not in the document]
rebassoo: This has been updated.

components/scream/src/share/tests/vertical_interp_tests.cpp
```diff
@@ -0,0 +1,315 @@
...
+  //using namespace scream;
+  //using namespace vinterp;
+
+  const int N = SCREAM_SMALL_PACK_SIZE;
+
+  int n_layers_src[4] = {2*N,2*N+1,2*N,2*N+1};
+  int n_layers_tgt[4] = {2*N,2*N,2*N-1,2*N-1};
+
+  for (int i=0; i<4; i++){
+    auto npacks_src = ekat::PackInfo<N>::num_packs(n_layers_src[i]);
```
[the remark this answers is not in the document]
rebassoo: This has been added.

components/scream/src/share/tests/vertical_interp_tests.cpp
```diff
@@ -0,0 +1,315 @@
...
+                           +", n_layers_tgt: "+std::to_string(n_layers_tgt[i])<<"\n");
+        }
+      }//end of looping over levels
+    }//end of looping over columns
+
+  }//end of looping over tests
+
+
+}
+
```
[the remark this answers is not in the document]
rebassoo: Understood and removed. Also went through the rest of the code to check that there is not other issues with empty lines.

components/scream/src/share/util/scream_vertical_interpolation_impl.hpp
```diff
@@ -0,0 +1,138 @@
...
+    const auto range = Kokkos::TeamThreadRange(team, x_tgt.extent(0));
+    //Mask out values above (below) maximum (minimum) source grid
+    Kokkos::parallel_for(range, [&] (const Int & k) {
+      const auto above_max = x_tgt[k] > x_src_s[nlevs_src-1];
+      const auto below_min = x_tgt[k] < x_src_s[0];
+      const auto combined_m = above_max || below_min;
+      msk(k) = combined_m;
+      out(k).set(combined_m,msk_val);
+    });
+    Kokkos::fence();
```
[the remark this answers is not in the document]
rebassoo: Updated this here, as well as updated a similar issue in the function: perform_vertical_interpolation_impl_1d()

components/scream/src/share/util/scream_vertical_interpolation.hpp
```diff
@@ -0,0 +1,126 @@
...
+using LIV = ekat::LinInterp<T,N>;
+ 
+template <typename S>
+using view_1d = typename KT::template view_1d<S>;
+template <typename S>
+using view_2d = typename KT::template view_2d<S>;
+    
+const Real masked_val = -std::numeric_limits<Real>::max();
+
+//This function call does not have a mask value provided by user
```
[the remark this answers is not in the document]
rebassoo: Moved the generic function to the top and updated the explanations

components/scream/src/share/util/scream_vertical_interpolation_impl.hpp
rebassoo: Added.

components/scream/src/share/util/scream_vertical_interpolation.hpp
rebassoo: Added.

components/scream/src/share/tests/vertical_interp_tests.cpp
```diff
@@ -0,0 +1,315 @@
...
+
+
+
+TEST_CASE("same_output_grid_test"){
```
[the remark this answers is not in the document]
rebassoo: I integrated this last test into a previous test based on this suggestion.

components/scream/src/share/tests/vertical_interp_tests.cpp
```diff
@@ -11,46 +11,192 @@ using Spack = SmallPack<Real>;
...
+                                     out,
+                                     mask,
+                                     n_layers_src[i],
+                                     n_layers_tgt[i]);
+
+      //Check that output of interpolation is as expected
+      for(int col=0; col<2; col++){
+        for(int lev=0; lev<(n_layers_tgt[i]-1); lev++){
+          if (perturb == 0){
+            REQUIRE(out_s(col,lev) == tmp_src_s(col,lev));
```
bartgol, line 89: Might be good to check also the `mask` values, no?

components/scream/src/share/tests/vertical_interp_tests.cpp
```diff
@@ -11,46 +11,192 @@ using Spack = SmallPack<Real>;
...
+      } 
+      //Set source levels and source input (tmp_src)
+      for (int col=0; col<2; col++){
+        for (int lev=0; lev<n_layers_src[i]; lev++){
+          p_src_s(col,lev) = 2.*lev;
+          tmp_src_s(col,lev) = 100.*(col+1) + 2.*lev;
+        }//end of looping over levels
+      }//end of looping over columns
+  
+      perform_vertical_interpolation(p_src,
```
bartgol, line 77: You need to deep copy the source views to device. Otherwise this call will only see views full of zeros (when running on GPU).

components/scream/src/share/tests/vertical_interp_tests.cpp
```diff
@@ -11,46 +11,192 @@ using Spack = SmallPack<Real>;
...
+  	  }
+        }//end of looping over levels
+      }//end of looping over columns
+  
+      //Take subview and run through the 1d interpolator function 
+      //and make sure get same thing back
+      ekat::LinInterp<Real,N> vert_interp(2,n_layers_src[i],n_layers_tgt[i]);
+      const int num_vert_packs = p_tgt.extent(0);
+      const auto policy = ESU::get_default_team_policy(1, num_vert_packs);
+      Kokkos::parallel_for("scream_vert_interp_setup_loop", policy,
```
bartgol, line 103: As a procaution, I would fill `out` with garbage before running this second test. This would prevent seeing the correct results from the first test above in case for some reason this loop doesn't run.

bartgol: Same for `mask`.

components/scream/src/share/tests/vertical_interp_tests.cpp
```diff
@@ -11,46 +11,192 @@ using Spack = SmallPack<Real>;
...
+                                                 n_layers_tgt[i],
+                                                 icol,
+                                                 masked_val,
+                                                 team,
+                                                 vert_interp); 
+      });
+      Kokkos::fence();
+
+      //Check that 1d interpolator output is consistent with what is expected
+      for(int col=0; col<2; col++){
```
bartgol, line 125: Again, you need to deep copy back to host before you can check the results.

components/scream/src/share/tests/vertical_interp_tests.cpp
```diff
@@ -11,46 +11,192 @@ using Spack = SmallPack<Real>;
...
+      perform_vertical_interpolation(p_src,
+                                     p_tgt,
+                                     tmp_src,
+                                     out,
+                                     mask,
+                                     n_layers_src[i],
+                                     n_layers_tgt[i]);
+
+      //Check that output of interpolation is as expected
+      for(int col=0; col<2; col++){
```
bartgol, line 86: These checks won't work on GPU, since `out_s` does not contain the updated output. You need a deep_copy back to host in order to read the correct results.

components/scream/src/share/util/scream_vertical_interpolation_impl.hpp
```diff
@@ -33,19 +34,17 @@ perform_vertical_interpolation_impl_2d(x_src, x_tgt, input, output, mask,
...
 void perform_vertical_interpolation(
-  const Src& x_src,
-  const Tgt& x_tgt,
-  const Input& input,
-  const view_2d<Pack<T,N>>& output,
-  const view_2d<Mask<N>>& mask,
-  const int nlevs_src,
-  const int nlevs_tgt,
-  const Real& msk_val)
+const Src& x_src,
```
bartgol, line 37: Looks like indentation is off here. Would help to indent the fcn inputs.

components/scream/src/share/util/scream_vertical_interpolation_impl.hpp
```diff
@@ -33,19 +34,17 @@ perform_vertical_interpolation_impl_2d(x_src, x_tgt, input, output, mask,
...
+const Tgt& x_tgt,
+const Input& input,
+const view_2d<Pack<T,N>>& output,
+const int nlevs_src,
+const int nlevs_tgt)
 {
+const view_2d<Mask<N>> mask("",x_src.extent(0),x_tgt.extent(0));
```
bartgol, line 44: Indentation off here too.

components/scream/src/share/tests/vertical_interp_tests.cpp
```diff
@@ -11,46 +11,192 @@ using Spack = SmallPack<Real>;
...
+    p_tgt_s(i) = 25.0 + i*5.0;
+  }
+ 
+  //Set source levels:
+  //For column 1 from 20-100
+  //For column 2 from 30-110
+  //Set input variable, in this case
+  //set tmp_src from 200-280 by 10, with final level at 240 so 
+  //can test interpolation with increasing/decreasing values
+  for (int i=0; i<9; i++){
```
bartgol, line 189: After this loop, you have to deep copy the inputs to the device views.

components/scream/src/share/tests/vertical_interp_tests.cpp
```diff
@@ -61,41 +207,41 @@ TEST_CASE("main_vertical_interpolation_test"){
...
+  }
+
+  correct_val[0][14] = 255.;
+  correct_val[0][15] = 240.;
+  correct_val[0][16] = masked_val;
+  correct_val[1][0]  = masked_val;
+  correct_val[1][15] = 270.;
+  correct_val[1][16] = 255.;
+
+  for(int col=0; col<2; col++){
```
bartgol, line 223: Need to deep copy output back to host before this loop.

bartgol: Also, check mask maybe?

components/scream/src/share/tests/vertical_interp_tests.cpp
```diff
@@ -110,16 +256,16 @@ TEST_CASE("main_vertical_interpolation_test"){
   });
   Kokkos::fence();
   
-  for(int col=0; col<3; col++){
-    for(int lev=0; lev<4; lev++){
+  for(int col=0; col<2; col++){
```
bartgol, line 259: Deep copy back to host.

components/scream/src/share/tests/vertical_interp_tests.cpp
```diff
@@ -110,16 +256,16 @@ TEST_CASE("main_vertical_interpolation_test"){
...
+    for(int lev=0; lev<17; lev++){
       REQUIRE(out_1d_test_s(col,lev) == correct_val[col][lev]);
     }
   }
 
   //Check to see if choose different masked value than default that it returns as expected
-  auto out_same = view_2d<Pack<Real,1>>("",3,npacks_tgt);
+  auto out_same = view_2d<Pack<Real,N>>("",2,npacks_tgt);
```
bartgol, line 266: I'm not sure what this test is doing. Are you testing that a mask value of -999 would yield the same result as the default mask value? If that's the case, aren't you basically checking that no value gets masked? And if so, you could just do that by checking the mask view above.

bartgol (changes requested):
Looks better. You're missing a few deep copies for GPU runs. Also, I have a couple of small suggestions on what you are testing.

components/scream/src/share/tests/vertical_interp_tests.cpp
rebassoo: I've added checks of the masks values throughout the test file

components/scream/src/share/tests/vertical_interp_tests.cpp
rebassoo: I have added the deep copy here

components/scream/src/share/tests/vertical_interp_tests.cpp
rebassoo: I made new variables instead of using the ones from the first test, so this should be resolved

components/scream/src/share/tests/vertical_interp_tests.cpp
rebassoo: Added the deep copy back

components/scream/src/share/util/scream_vertical_interpolation_impl.hpp
rebassoo: Fixed indentation

components/scream/src/share/tests/vertical_interp_tests.cpp
rebassoo: Added deep copy

components/scream/src/share/tests/vertical_interp_tests.cpp
rebassoo: Added deep copy output back to host as well as check on mask

components/scream/src/share/tests/vertical_interp_tests.cpp
rebassoo: Added deep copy back to host

components/scream/src/share/tests/vertical_interp_tests.cpp
rebassoo: This test is testing that a user can specify a mask value different than the default one in scream_vertical_interpolation.hpp. In this case the user provides -999. and this test checks that the function properly uses this and masks out-of-bounds values to -999 (not the default value in the scream_vertical_interpolation.hpp file). I don't think you can check this by checking the mask view above. The mask view above will be the same but your output view will have different values depending on what the user specifies (or doesn't) for the mask value.

bartgol (dismissed):

jgfouca (approved):

bartgol (dismissed):

AaronDonahue (approved):
the last commit was a simple single line bugfix.  Since it was approved before I am going to approve again.

AaronDonahue:
@bartgol or @jgfouca , could there be a reason the AT hasn't started for this PR?  Unless the AT is just slow today.  But I wanted to check since we have a WIP PR waiting on this going in to be ready.

bartgol:
> @bartgol or @jgfouca , could there be a reason the AT hasn't started for this PR? Unless the AT is just slow today. But I wanted to check since we have a WIP PR waiting on this going in to be ready.

This PR comes from a fork. As such, the AT always considers it "unsafe", and a pre-test inspection must be performed. I'll set the pre-test inspected label.

PR #1938, opened 2022-09-13
SHOC: Use another team_barrier based on view_reduction pattern.

ambrad:
This one is based just on grepping. We know there is still a failure somewhere. I've isolated it again to SHOC on ascent but haven't yet gotten direct evidence of the detailed source. But grepping the text showed these adjacent reductions, which I identified in #1921 as a problem.

Associated with #1917.


ambrad:
The blake build failed for system reasons. This is a one liner, so I'm going to merge it manually.

PR #1940, opened 2022-09-13
SHOC support for small kernels

### LINKED ISSUE #1903: Support for both monolithic and many small kernels to address Crusher/Frontier challenges.
The AMD compilers for the AMD GPUs, e.g. hippcc, are having a hard time with large kernels, SHOC in particular. Error modes include spurious segmentation faults, hanging, and two types of failures to allocate registers during compilation. Right now we have a workaround for SHOC, but it makes SHOC very slow, much slower than the dycore for example.

Fortunately, our column-physics s/w design accommodates both monolithic (LK) and many small kernels (SK). Recall that LK is generally best on the CPU because it encourages good caching, it minimizes memory footprint due to temporaries, and it has been fine on Nvidia GPUs; thus we have favored it so far. LK is also harder to design for, so it's good to design it in up front and retreat to SKs as needed with optional configuration flags. Here I outline what we should do to help SCREAMv1 run well on Crusher/Frontier.

The keys to accommodating both LK and SK are:

1. function decomposition and
2. the workspace manager.

Consider shoc_main_internal as an example. In the LK approach, the workspace manager manages nslot work slots for the ncol columns. On the GPU, it's possible by design (to decrease the memory footprint) for nslot < ncol. shoc_main_internal acts on one column at a time, calling a long sequence of subroutines on the same column data, with the work slots holding temporary data.

In the SK approach, we will set nslot = ncol always. Each subroutine call in shoc_main_internal will be replaced by a top-level Kokkos loop (KOKKOS_LAMBDA, a kernel dispatch) over ncol. Work slots will hold data persistently between kernel dispatches.

To enable both LK and SK, we need to do the following:

1. Introduce a new SCREAM CMake variable, SCREAM_MONOLITHIC_KERNELS (or something like that), defaulting to ON.

2. Introduce a new EKAT CMake variable, EKAT_WORKSPACE_NSLOT_EQ_NCOL (or something like that), defaulting to OFF.

3. In SCREAM, if SCREAM_MONOLITHIC_KERNELS is OFF, force EKAT_WORKSPACE_NSLOT_EQ_NCOL to be ON.

4. In shoc_main, ifdef SCREAM_MONOLITHIC_KERNELS, do what is currently done; else, add new code to call an alternative version of shoc_main_internal.

5. In the alternative shoc_main_internal, replace each subroutine call with a dispatch loop calling the subroutine. The local scalars should be stored in an extra work slot.

6. Optionally, make each dispatch its own subroutine and put it in its own cpp file. This will have two benefits: (1) shorter compilation time (many small translation units to build in parallel) and (2) if there is an internal compiler error (ICE), it will be immediately obvious what bit of code is triggering the ICE. To support this capability, the CMake code will have to add a bunch of cpp files to the source file list if SCREAM_MONOLITHIC_KERNELS is OFF.

We should start with SHOC because it is the problem at the moment, but P3 should eventually support SCREAM_MONOLITHIC_KERNELS, too.

jgfouca:
This is a major refactor of SHOC to support small kernels which is needed on frontier/crusher.

The only significant challenge so far has been how to maintain workspaces across kernels. This is a new challenge for us because all previous implementation were monolithic kernels. The approach that appears to be working the best is to use integer slot-ids as a handle to the workspace.

Another challenge is that temporary scalars in the monolithic kernel have to become views of schol scalars for small kernels.

I've only transitioned a couple SHOC functions to supporting small kernels. I wanted to get approval for the pattern I'm using before doing the rest since that's going to be a considerable amount of work.

There a few minor changes needed to EKAT in order for this to build that you can't see here. The only interesting change was the addition of this WSM::Workspace method:
```
template <typename T, typename D>
template <typename S>
KOKKOS_FORCEINLINE_FUNCTION
Unmanaged<typename WorkspaceManager<T, D>::template view_1d<S> >
WorkspaceManager<T, D>::Workspace::get_space_in_slot(
  const int slot) const
{
  EKAT_KERNEL_ASSERT_MSG(slot < m_parent.m_max_used, "Asked for out-of-bounds slot\n");
  auto result = m_parent.get_space_in_slot<S>(m_ws_idx, slot);
  EKAT_KERNEL_ASSERT_MSG(is_active<S>(result), "Asked for inactive slot\n");

  return result;
}
```

Fixes #1903


components/scream/src/physics/shoc/shoc_main_impl.hpp
```diff
@@ -285,7 +286,267 @@ void Functions<S,D>::shoc_main_internal(
...
+  using ExeSpace = typename KT::ExeSpace;
 
+  static constexpr Int num_local_scalars = 13;
+  static constexpr Int num_local_ws      = 5;
+  static constexpr Int num_total_local   = num_local_scalars + num_local_ws;
+
+  // Need a second workspace for storing shcol local scalars. Need workspaces
+  // of size schol, not nlev, to store temporary scalars for small kernels
+  const auto policy = ekat::ExeSpaceUtils<ExeSpace>::get_default_team_policy(shcol, 1);
+  ScalarWorkspaceMgr wsm_local(shcol, num_local_scalars, policy, 1.e10); // huge overprovision => no ws sharing
```
jgfouca, line 358: Since both the workspace size and type are different, I decided to make a new WSM for managing local scalars instead of trying to leverage the existing WSM.

components/scream/CMakeLists.txt
```diff
@@ -192,6 +197,10 @@ set(SCREAM_LIB_ONLY ${DEFAULT_LIB_ONLY} CACHE BOOL "Only build libraries, no exe
 set(NetCDF_Fortran_PATH ${DEFAULT_NetCDF_Fortran_PATH} CACHE FILEPATH "Path to netcdf fortran installation")
 set(NetCDF_C_PATH ${DEFAULT_NetCDF_C_PATH} CACHE FILEPATH "Path to netcdf C installation")
 set(SCREAM_MACHINE ${DEFAULT_SCREAM_MACHINE} CACHE STRING "The CIME/SCREAM name for the current machine")
+set(SCREAM_MONOLITHIC_KERNELS ${DEFAULT_MONOLITHIC_KERNELS} CACHE STRING "Use monolithic kokkos kernels")
+if (NOT SCREAM_MONOLITHIC_KERNELS)
+  set(EKAT_DISABLE_WORKSPACE_SHARING TRUE CACHE STRING "")
```
jgfouca, line 202: I'm not 100% sure this is needed or even a good design choice. I think we should leave it to SCREAM to set up their WSMs properly. Any WSM that might be used for small kernels should have their over provision factor set to a huge number. This will prevent any workspace sharing. The top level functions of small kernels should assert that their WSMs have one slot per team.

components/scream/src/physics/shoc/CMakeLists.txt
```diff
@@ -62,6 +62,14 @@ if (NOT CUDA_BUILD OR Kokkos_ENABLE_CUDA_RELOCATABLE_DEVICE_CODE)
   ) # SHOC ETI SRCS
 endif()
 
+# Add dispatch source files if monolithic kernels are off
+if (NOT SCREAM_MONOLITHIC_KERNELS)
+  list(APPEND SHOC_SRCS
+    shoc_energy_integrals_disp.cpp
```
bartgol, line 68: Not sure if you want to do it in this PR, and it might be a personal taste thing, but in case you agree: the shoc (and p3) folders are quite big, given the large number of `_impl.cpp` and `.cpp` files (96 files in the shoc folder). This makes navigating the repo a bit hard (imho). I would vote for adding subfolders: one `impl` subfolder for the impl files, one `eti` subfolder for the ETI cpp files, and one `dispatch` for the new files you are adding. That will also allow the less experienced users to recognize the difference between eti and dispatch `*.cpp` files...

components/scream/src/physics/shoc/shoc_energy_fixer_disp.cpp
```diff
@@ -0,0 +1,81 @@
+#include "shoc_functions.hpp"
+
+#include "ekat/kokkos/ekat_subview_utils.hpp"
+
+namespace scream {
+namespace shoc {
+
+template<>
+void Functions<Real,DefaultDevice>
+::shoc_energy_fixer_disp(
```
bartgol, line 10: My personal taste here would be for spelling out `dispatch`, but that's very subjective.

jgfouca: Those extra 4 keystrokes would give me carpal tunnel!

components/scream/src/physics/shoc/shoc_energy_fixer_disp.cpp
```diff
@@ -0,0 +1,81 @@
...
+  const Int&                   shcol,
+  const Int&                   nlev,
+  const Int&                   nlevi,
+  const Scalar&                dtime,
+  const Int&                   nadv,
+  const view_2d<const Spack>&  zt_grid,
+  const view_2d<const Spack>&  zi_grid,
+  const Int&                   se_b_slot,
```
bartgol, line 18: Maybe we could group these slot indices in a verbose struct, like `shoc::SmallKernelsSlotsIds` or something.

components/scream/src/physics/shoc/CMakeLists.txt
jgfouca: I agree. This directory now has 98 files and will be growing to near 150 when this is done.

components/scream/src/physics/shoc/shoc_functions.hpp
```diff
@@ -611,6 +659,65 @@ struct Functions
     const uview_1d<Spack>&       wqls_sec,
     const uview_1d<Spack>&       brunt,
     const uview_1d<Spack>&       isotropy);
+#else
+  static void shoc_main_internal(
```
bartgol, line 663: Maybe we could give these functions two different names, like `shoc_main_monolithic` and `shoc_main_multiple_kernels`, or something.

bartgol (commented):
I think overall looks fine. My biggest concern is potential confusion from code duplication and bloating of files. Which is why at some point (perhaps at the end of the refactor), I would prefer if we spend a tiny bit of effort on organizing code/files in a self-documenting way.

components/scream/CMakeLists.txt
bartgol: I agree. I'm not 100% sure what this option would do in EKAT. But I feel like we don't want to set _global_ ekat configurations just b/c _a few_ use cases need it. If we can make SCREAM create WSMs differently, each with its own config/ctor params, I think that would be better.

ambrad: EKAT's WSM should support a new option to make the workspace size proportional to league size (total number of physics columns) rather than use the overprovision factor. This will follow what Hommexx does.

ambrad:
Good start. I'd like to offer two alternatives for handling persistent workspace slots.

(1) Rather than getting pointers (*_slot indices) and passing those around, make a setup/teardown pair of routines that just does the usual thing, e.g.
```C++
  workspace.template take_many_and_reset<5>(
    {"rho_zt", "shoc_qv", "dz_zt", "dz_zi", "tkh"},
    {&rho_zt, &shoc_qv, &dz_zt, &dz_zi, &tkh});
```
and then releases it all at the end of the kernel. Each small kernel always calls this setup/teardown pair and thus always gets the right persistent memory for each variable. You could package the persistent arrays in a structure. This pattern would be useful in the case of more complicated partial-persistence patterns where the setup/teardown would change throughout the parameterization. SHOC doesn't seem to need that, though. But this pattern is what I had in mind in #1903, where I was thinking about the general case.

(2) Instead, this immediately suggests a simpler option for SHOC: In the case of nonmono kernels, just have one additional buffer struct for these arrays, and when requesting buffer space, include this additional buffer struct in the request. Then no extra workspace gymnastics are needed in SHOC.

For the scalars, I recommend making an array to hold all scalars and then providing an enum in the new buffer struct that indexes this array, like this:
```C++
enum : int { se_b = 0, ke_b, wv_b, ... };
view1d<Scalar> scals;
// usage
b.scals[b.se_b]
```

bartgol:
I'm not a fan of the setup/teardown approach, which relies on the order of fields to be always the same. It is a somewhat subtle assumption, which has the risk of being forgotten 3yy down the road when another developer is maybe adding a feature to shoc. Of course, inline documentation would help, but it would still be a bit fragile.

Option 2 seems more attractive. If I understand correctly, you want to use "normal" buffer views for persistent data, without using WSM at all, right? Then, inside each top-level kernel, you would subview those buffers, rather than grab slots from the WSM. Imho, this is much simpler than hacking the WSM to get non-scratch memory.

ambrad:
@bartgol re: option 2, yes. Since it's all we need for SHOC, we can table discussion of more complicated temp-buffer use cases.

jgfouca:
@ambrad , @bartgol , OK, see if you like this better. I agree that, if you aren't sharing or releasing workspaces, then it raises the question of why use the WSM at all? The code is definitely simpler without it.

components/scream/src/physics/shoc/shoc_main_impl.hpp
```diff
@@ -285,7 +286,242 @@ void Functions<S,D>::shoc_main_internal(
...
+  const view_2d<Spack>&       wtke_sec,
+  const view_2d<Spack>&       uw_sec,
+  const view_2d<Spack>&       vw_sec,
+  const view_2d<Spack>&       w3,
+  const view_2d<Spack>&       wqls_sec,
+  const view_2d<Spack>&       brunt,
+  const view_2d<Spack>&       isotropy)
+{
+  // Create space for temporary scalars
+  view_1d<Scalar>
```
ambrad, line 350: These views shouldn't be allocated each time SHOC is called. They should be part of the init-time allocation using buffer memory. I suggest using a struct to hold these.

jgfouca:
@ambrad, @bartgol , see if you like this latest commit. I know that I still need to reorganize the directory (impls, eti, and dispatch subdirectories).

components/scream/src/physics/shoc/shoc_functions_f90.cpp
```diff
@@ -2693,7 +2693,12 @@ int shoc_init_f(Int nlev, Real *pref_mid, Int nbot_shoc, Int ntop_shoc)
   ekat::host_to_device({pref_mid}, nlev, temp_d);
   view_1d pref_mid_d(temp_d[0]);
 
+#ifndef SCREAM_MONOLITHIC_KERNELS
+  SHF::SHOCTemporaries temporaries; // we won't keep these
```
jgfouca, line 2697: This is a little hacky but fine IMO. The f90->cxx bridge is already slow so it doesn't matter if we reallocate temporaries for each shoc_main_f call.

components/scream/src/physics/shoc/shoc_main_impl.hpp
```diff
@@ -378,6 +645,44 @@ Int Functions<S,D>::shoc_main(
...
+    shoc_input.wtracer_sfc, shoc_input.inv_exner, shoc_input.phis, // Input
+    workspace_mgr, // Workspace Manager
+    shoc_input_output.host_dse, shoc_input_output.tke, shoc_input_output.thetal, shoc_input_output.qw, u_wind_s, v_wind_s, // Input/Output
+    shoc_input_output.wthv_sec, shoc_input_output.qtracers, shoc_input_output.tk, shoc_input_output.shoc_cldfrac, // Input/Output
+    shoc_input_output.shoc_ql, // Input/Output
+    shoc_output.pblh, shoc_output.shoc_ql2, // Output
+    shoc_history_output.shoc_mix, shoc_history_output.w_sec, shoc_history_output.thl_sec, shoc_history_output.qw_sec, shoc_history_output.qwthl_sec, // Diagnostic Output Variables
+    shoc_history_output.wthl_sec, shoc_history_output.wqw_sec, shoc_history_output.wtke_sec, shoc_history_output.uw_sec, shoc_history_output.vw_sec, // Diagnostic Output Variables
+    shoc_history_output.w3, shoc_history_output.wqls_sec, shoc_history_output.brunt, shoc_history_output.isotropy, // Diagnostic Output Variables
+    shoc_temporaries.se_b,
```
jgfouca, line 667: I followed the existing pattern of unpacking the struct before calling shoc_main_internal even though this function has a crazy number of arguments.

components/scream/src/physics/shoc/shoc_main_impl.hpp
```diff
@@ -58,6 +63,33 @@ Int Functions<S,D>::shoc_init(
   const auto host_view = Kokkos::create_mirror_view(npbl_d);
   Kokkos::deep_copy(host_view, npbl_d);
 
+  // Allocate temporaries if using small kernels
+#ifndef SCREAM_MONOLITHIC_KERNELS
```
ambrad, line 67: If I understand correctly, this part is not following the established pattern for processes to allocate workspace. I believe this code should be in `SHOCMacrophysics::init_buffers` and be allocated using the `buffer_manager.get_memory()` pointer, as is done for `m_buffer`. The reason for this is then no process is holding onto a bunch of device memory that can't be used by any other process.

jgfouca: @ambrad , thanks, I was wondering if I should use that (Buffer). I have pushed a commit that does this.

components/scream/src/physics/shoc/atmosphere_macrophysics.cpp
```diff
@@ -165,68 +168,33 @@ void SHOCMacrophysics::init_buffers(const ATMBufferManager &buffer_manager)
...
-  m_buffer.thv = decltype(m_buffer.thv)(s_mem, m_num_cols, nlev_packs);
-  s_mem += m_buffer.thv.size();
-  m_buffer.dz = decltype(m_buffer.dz)(s_mem, m_num_cols, nlev_packs);
-  s_mem += m_buffer.dz .size();
-  m_buffer.zt_grid = decltype(m_buffer.zt_grid)(s_mem, m_num_cols, nlev_packs);
-  s_mem += m_buffer.zt_grid.size();
-  m_buffer.zi_grid = decltype(m_buffer.zi_grid)(s_mem, m_num_cols, nlevi_packs);
-  s_mem += m_buffer.zi_grid.size();
+  using spack_2d_view_t = decltype(m_buffer.z_mid);
+  spack_2d_view_t* _2d_spack_mid_view_ptrs[Buffer::num_2d_vector_mid] = {
```
jgfouca, line 172: I took the opportunity to clean up some of the repetitive code in this method. The changes to the 1d code should be semantically identical to the previous impl. The 2d code does not behave exactly as before because the order of the views in the buffer is different with all the "mid" views and "int" views grouped separately now. I assume this order doesn't matter. If it does, I will refactor this to preserve the original order.

jgfouca: @tcclevenger ^

ambrad:
@jgfouca, after the last changes, this code looks great. Are you wanting to merge this PR, or are you looking for approval from everyone following our discussions above and then you'll continue working on this branch using the patterns you established?

jgfouca:
@ambrad , thanks for the good feedback. 

> Are you wanting to merge this PR, or are you looking for approval from everyone following our discussions above and then you'll continue working on this branch using the patterns you established?

I'm going to continue working on this using the established pattern.

ambrad:
Great. I say go for it. Luca's comments about file organization can be folded in as you proceed.

jgfouca:
This passes test-all-scream on mappy and weaver. I think it's ready to merge.

ambrad:
Wow, this is great. What testing will there be in the nightlies for this new capability? In addition to unit tests, I'd like one new v1 test: ERS_Ln90.ne30pg2_ne30pg2.F2010-SCREAMv1, but with a testmod to set the small-kernels flag.

jgfouca:
@ambrad , 

> What testing will there be in the nightlies for this new capability?

Oof, I knew I would forget something. If we want really thorough testing, we could add another mode to test-all-scream. That may be overkill since only SHOC is currently impacted by setting SCREAM_MONOLITHIC_KERNELS=off. @bartgol , thoughts?

bartgol:
> @ambrad ,
> 
> > What testing will there be in the nightlies for this new capability?
> 
> Oof, I knew I would forget something. If we want really thorough testing, we could add another mode to test-all-scream. That may be overkill since only SHOC is currently impacted by setting SCREAM_MONOLITHIC_KERNELS=off. @bartgol , thoughts?

A full new build just to run shoc_tests is indeed an overkill. But we could build shoc_tests twice on CUDA, the second time with SCREAM_MONOLITHIC_KERNELS=ON. The "global" macro would not be defined in `scream_config.h`, but we can pass additional compiler definitions to CreateUnitTest, when building shoc_tests. Notice that this only works on CUDA, where ETI is off, so that the impl files are built while we build the test. On CPU, with ETI on, we would have to rebuild the shoc lib, with using `target_compile_definitions` to set the macro in the second build.

Long story short: I vote to build a second version of `shoc_tests` on GPU builds, passing `COMPILER_CXX_DEFS SCREAM_MONOLITHIC_KERNELS` the second time we build the tests.

jgfouca:
@ambrad , @bartgol , OK, we now have a small kernel test that should always run.

components/scream/src/physics/shoc/tests/CMakeLists.txt
```diff
@@ -73,7 +74,8 @@ set(SHOC_TESTS_SRCS
 
 # NOTE: tests inside this if statement won't be built in a baselines-only build
 if (NOT SCREAM_BASELINES_ONLY)
-  CreateUnitTest(shoc_tests "${SHOC_TESTS_SRCS}" "${NEED_LIBS}" THREADS 1 ${SCREAM_TEST_MAX_THREADS} ${SCREAM_TEST_THREAD_INC} DEP shoc_tests_ut_np1_omp1)
+  CreateUnitTest(shoc_tests    "${SHOC_TESTS_SRCS}" "${NEED_LIBS}"    THREADS 1 ${SCREAM_TEST_MAX_THREADS} ${SCREAM_TEST_THREAD_INC} DEP shoc_tests_ut_np1_omp1)
+  CreateUnitTest(shoc_sk_tests "${SHOC_TESTS_SRCS}" "${SK_NEED_LIBS}" THREADS 1 ${SCREAM_TEST_MAX_THREADS} ${SCREAM_TEST_THREAD_INC} DEP shoc_tests_ut_np1_omp1 EXE_ARGS shoc_main_bfb)
```
jgfouca, line 78: New test here.

ambrad: Would you also add something like ERS_Ln90.ne30pg2_ne30pg2.F2010-SCREAMv1 with a testmod to set the small-kernels flag? That can go in the e3sm_scream_v1_lowres tests.py group.

components/scream/src/physics/shoc/CMakeLists.txt
```diff
@@ -73,10 +92,16 @@ if (NOT SCREAM_DEBUG)
 endif()
 
 add_library(shoc ${SHOC_SRCS})
-target_include_directories(shoc PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/../share)
-set_target_properties(shoc PROPERTIES Fortran_MODULE_DIRECTORY ${SCREAM_F90_MODULES})
-target_link_libraries(shoc physics_share scream_share)
-target_compile_options(shoc PUBLIC $<$<COMPILE_LANGUAGE:Fortran>:${SCREAM_Fortran_FLAGS}>)
+add_library(shoc_sk ${SHOC_SRCS} ${SHOC_SK_SRCS})
```
bartgol, line 95: I think we should build the SK lib only if bot SCREAM_SMALL_KERNELS and SCREAM_LIB_ONLY are OFF. If the former is ON, then shoc itself will contain small kernels (right?), while if the latter is ON, then we are not going to run tests anyways...

jgfouca: @bartgol , I thought about this too. If SCREAM_SMALL_KERNELS is on, then shoc and shoc_sk are testing the same thing. The shoc_sk test only runs shoc_main_bfb, so I assumed this was an acceptable level of redundancy given that SCREAM_SMALL_KERNELS will usually be off, so why complicate the CMake files even more. If you feel strongly, I will change it to eliminate unneeded builds/tests.

components/scream/src/physics/shoc/tests/CMakeLists.txt
jgfouca: Yes, I can add.

components/scream/src/physics/shoc/shoc_energy_integrals_impl.hpp
```diff
@@ -37,34 +37,47 @@ ::shoc_energy_integrals(
...
+  const auto srcm      = ekat::scalarize(rcm);
+  const auto su_wind   = ekat::scalarize(u_wind);
+
+  const auto sv_wind   = ekat::scalarize(v_wind);
   // Compute se_int
-  ExeSpaceUtils::view_reduction(team,0,nlev,
-                                [&] (const int k) -> Spack {
-    return host_dse(k)*pdel(k)/ggr;
-  }, se_int);
+  ExeSpaceUtils::parallel_reduce(team,0,nlev,
```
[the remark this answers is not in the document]
jgfouca: Serialize defaults to True when EKAT_BFB is on, which it should be if we expect to be bfb with f90.

components/scream/src/physics/shoc/CMakeLists.txt
bartgol: I don't feel strongly. We can leave it as is, and perhaps change later if need be.

jgfouca: @bartgol , ah, I hit link problems with SMALL_KERNELS on. You are right, this chunk of code needs a refactor.

ambrad (approved):

jgfouca:
OK, I believe this is now 100% ready.

components/scream/cime_config/testdefs/testmods_dirs/scream/small_kernels/shell_commands
```diff
@@ -0,0 +1 @@
+./xmlchange SCREAM_CMAKE_OPTIONS='SCREAM_SMALL_KERNELS On'
```
ambrad, line 1: Does this line definitely work? I don't remember whether you then need to then reset tracer count to 10.

jgfouca: @ambrad , wow, good catch. I have fixed this and updated the documentation for making SCREAM_CMAKE_OPTIONS test mods.

ambrad (dismissed):

components/scream/cime_config/testdefs/testmods_dirs/scream/small_kernels/shell_commands
bartgol: Yeah, I think we need `--append` here.

jgfouca:
Darn, looks like I might have some real DIFFs in e3sm_scream_v1 to sort out.

ambrad:
> Darn, looks like I might have some real DIFFs in e3sm_scream_v1 to sort out.

Do you mean the mappy ones in the autotester? They probably are real, but I'll nonetheless note that these are the same tests that caused issues in other PRs I think on Monday due to blessing issues.

jgfouca:
@PeterCaldwell , @bartgol , @ambrad ,

This has been a nightmare of a PR. Now I cannot get Release builds of CIME cases to be BFB with upstream. I have been working with the case `SMS_Ln9.ne4_ne4.F2010-SCREAMv1.mappy_gnu9` which builds and runs quickly and demonstrates the problem. I've been investigating for days and here's what I know:
1. In standalone (IE, not CIME) testing, I also struggled with the `shoc_energy_integrals` function being non-BFB. I eventually dodged the problem by scalarizing the views and using `ExeSpaceUtils::parallel_reduce` instead of `ExeSpaceUtils::view_reduction`. With this change, CXX is BFB with fortran in SCREAM standalone on all CI platforms, release and debug, and with or without small kernels.
2. In CIME testing, the switch to `ExeSpaceUtils::parallel_reduce` is not BFB with the previous implementation of `shoc_energy_integrals` for Release builds. This is different from the problems I had in step 1 which were exposed by turning on small kernels and setting OMP_NUM_THREADS>1. Here, small kernels are off and the thing that exposes the problem is switching from a Debug to a Release build. 
3. Even more surprising, this problem persists even with OMP_NUM_THREADS=1 and SCREAM_PACK_SIZE=1, which should be the easiest scenario to achieve BFB-ness.
4. Forcing serialization of `ExeSpaceUtils::parallel_reduce` does not fix the problem, nor does setting EKAT_DEFAULT_BFB to true.
5. The DIFF is caused by any of the first 3 reductions being switched to `ExeSpaceUtils::parallel_reduce`. The 4th one seems to be BFB either way and so it's the only one working as expected.
6. The only difference in compilation flags between SCREAM standalone and CIME is the addition of `-ffp-contract=fast` on the CIME side, which I think is the default so shouldn't matter.

Options:
1. The differences appear to be small, so we can just bless and move on. I think that would be my preferred choice at this time.
2. I can keep digging and start printing the outputs at each step in the reduction and try to see if anything can be learned.
3. Something else one of you can think to try?

jgfouca:
@ambrad , I don't mind waiting, but I'd hate to add another thing to your plate. What do you think about just blessing the DIFF?

ambrad:
It's a small thing, but we don't want to move away from packed reductions. I want to see if the ekat packed reduction code is broken. Thus, I'm a little reluctant to merge this PR right now.

ambrad (commented):
Switching approval status until I get a chance to look at the reduction code.

bartgol:
@jgfouca, wild guess: have you tried to init result to 0 in view_reduction? I see that if `Serialize=false`, we _do_ init the packed result to 0, but if `Serialize=true` we don't. If there is garbage at the beginning, btw, we never init result to 0.

Note: I don't think we should init to 0 in `view_reduction`, cause user might want to accumulate in result. So maybe you should init se_int to 0 before calling view_reduction.

bartgol:
Ah, I think I might have something: in here
```
  if (pack_loop_begin != pack_loop_end) {
    if (Serialize) {
      impl::parallel_reduce<Serialize>(team, pack_loop_begin, pack_loop_end,
                                       [&](const int k, ValueType& local_sum) {
          // Sum over pack entries and add to local_sum
          ekat::reduce_sum<Serialize>(input(k),local_sum);
      }, result);
    } else {
    ...
```
This is not the same as a sum from begin to end. In fact, we reduce each pack first, and _then_ we sum those contributions. So with beg=0, end=4, and pack size 2, we are doing
```
(v(0)+v(1)) + (v(2)+v(3))
```
instead of
```
v(0)+v(1) + v(2)+v(3)
```

jgfouca:
@ambrad , @bartgol , since we appear to be leaning towards going back to the original `view_reduction` implementation of `shoc_energy_integrals`:

1. I use this cake config: `cmake -DCMAKE_CXX_COMPILER=$(which mpicxx) -DCMAKE_C_COMPILER=$(which mpicc) -DCMAKE_Fortran_COMPILER=$(which mpif90) -C ../../../components/scream/cmake/machine-files/mappy.cmake -DSCREAM_SMALL_KERNELS=On -DCMAKE_BUILD_TYPE=Debug ..`
2. `cd src/physics/shoc/tests`
3. `OMP_NUM_THREADS=X ./shoc_tests shoc_main_bfb`
4. If X>1 then I start to see random large (not roundoff) diffs. I see this even when SCREAM_PACK_SIZE=1

bartgol:
I realize my excitement about the "bug" was too premature. What I thought was a bug is not, in fact, a bug.

jgfouca:
It looks like everything will pass with the most recent commit.

bartgol:
I'm still puzzled on why the automatic var changes things. Also, if this is the fix we needed, then we should have needed it for the monolithic kernel as well.

jgfouca:
@bartgol , in the monolithic kernel, those values are already locals:

```
  // Local scalars                                                                                                                                                                                                            
  Scalar se_b{0},   ke_b{0}, wv_b{0},   wl_b{0},
         se_a{0},   ke_a{0}, wv_a{0},   wl_a{0},
         ustar{0},  kbfs{0}, obklen{0}, ustar2{0}, wstar{0};
```

In the small kernels, they are stored in views, so shared.

jgfouca:
This also PASSes the CIME test I was having trouble with.

bartgol:
Ok, that explains the difference with monolithic kernels. But I still don't see _where_ the fact that the var is shared should cause problems. Only team rank 0 performs += on the input var. The shared memory ops are performed in `impl::parallel_reduce`, which already operates on a local var.

ambrad:
I think I see the issue, and sorry I didn't spot this sooner. The straightforward replacement for the scalars in the monolithic kernel would be a 2D view indexed on (i,k), not just i. However, using a 1D view is fine as long as you take measures to handle the fact that what was a thread-local scalar in the monolithic kernel is now a global variable shared among all threads in the team.

bartgol:
I think using `PerThread` instead of `PerTeam` in the singles is incorrect, since on CPU we have all team members running that single. But I swapped those with `PerTeam`'s, and it's still non bfb.

ambrad:
@bartgol this is an example of where a global var shared among all threads in a team will lead to an error:
https://github.com/E3SM-Project/EKAT/blob/dbe3d1c2e3d6242ce998865b2d4676676d80bb69/src/ekat/kokkos/ekat_kokkos_utils.hpp#L98
This should be protected by a single in that case.

bartgol:
> @bartgol this is an example of where a global var shared among all threads in a team will lead to an error: https://github.com/E3SM-Project/EKAT/blob/dbe3d1c2e3d6242ce998865b2d4676676d80bb69/src/ekat/kokkos/ekat_kokkos_utils.hpp#L98 This should be protected by a single in that case.

Yes, that's a pb. We also have another += on the shared var in the view_reduction function, for Serialize=false. But for Serialize=true, all += on shared vars are only performed by team rank 0.

bartgol:
For clarity, I tried to add `team_barrier()` and `team_broadcast(result,0)` calls all over the two fcns, and changed `PerThread` into `PerTeam`, to no avail. I don't see where += can be performed on 2+ threads on the shared mem var.

bartgol:
Also, tbc, I'm fine with Jim's mod. I just want to understand where the shared memory was causing a problem with Serialize=true.

ambrad:
@bartgol re: Serialize, e.g. here: https://github.com/E3SM-Project/EKAT/blob/dbe3d1c2e3d6242ce998865b2d4676676d80bb69/src/ekat/kokkos/ekat_kokkos_utils.hpp#L155

bartgol:
> @bartgol re: Serialize, e.g. here: https://github.com/E3SM-Project/EKAT/blob/dbe3d1c2e3d6242ce998865b2d4676676d80bb69/src/ekat/kokkos/ekat_kokkos_utils.hpp#L155

Yes, that's the one I was referring to with "We also have another += on the shared var in the view_reduction function". But that code only happens if Serialize=false. If Serialize=true, none of the += on shared memory happens.

ambrad:
Ok, what about https://github.com/E3SM-Project/EKAT/blob/dbe3d1c2e3d6242ce998865b2d4676676d80bb69/src/ekat/kokkos/ekat_kokkos_utils.hpp#L133? However, this applies only if pack size > 1.

bartgol:
> Ok, what about [E3SM-Project/EKAT@`dbe3d1c`/src/ekat/kokkos/ekat_kokkos_utils.hpp#L133](https://github.com/E3SM-Project/EKAT/blob/dbe3d1c2e3d6242ce998865b2d4676676d80bb69/src/ekat/kokkos/ekat_kokkos_utils.hpp#L133)? However, this applies only if pack size > 1.

Yes, that is a problem, since all team members execute that line. But I tried to switch PerThread with PerTeam, adding a team.team_broadcast(result,0) right after the single, and still got non bfb.

ambrad:
Ah, ok, maybe this somewhat more subtle situation: In parallel_reduce, thread 1 writes to result before thread 2 executes `local_tmp = result`.

bartgol:
> Ah, ok, maybe this somewhat more subtle situation: In parallel_reduce, thread 1 writes to result before thread 2 executes `local_tmp = result`.

Maybe. I stuffed the function with plenty of team_barrier() calls though, so I am not really sure. I will raise my hands and give up. Though I think we should fix this in ekat, to avoid doing the same mistake again somewhere else.

bartgol:
Uhm, I modified view_reduction to do
```
{
  auto temp = result;
  [same code but updating temp instead of result]
  result = temp;
}
```
and called it passing `se_int` (which is from a view) as arg for result, but it's not passing the tests. I think this _should_ be the same as doing
```
Scalar se_tmp(0);
view_reduction(...,se_tmp);
se_int = se_tmp;
```
with the original code, no?

ambrad:
Luca, I don't think it's the same. I believe the thread race condition I mentioned is the key here. In the first case you quote, `temp = result` in thread 2 could be executed after thread 1 has set `result = temp`. In the second case you quote, `se_tmp` is not set to `se_int` initially; rather, it's set to 0. Thus, no race condition occurs.

bartgol:
> Luca, I don't think it's the same. I believe the thread race condition I mentioned is the key here. In the first case you quote, `temp = result` in thread 2 could be executed after thread 1 has set `result = temp`. In the second case you quote, `se_tmp` is not set to `se_int` initially; rather, it's set to 0. Thus, no race condition occurs.

Mmm, ok, but then adding a team_barrier() after `temp=result` should restore bfb, while it doesn't.

Anyway, I should prob stop worring about this. I was just trying to emulate Jim's fix inside EKAT, so that we would not have this issue again in the future when calling `view_reduction` with a shared mem output variable.

jgfouca:
@bartgol , @ambrad , thanks for the deep dive on `view_reduction`. I made an EKAT issue a while back https://github.com/E3SM-Project/EKAT/issues/254 so maybe we should document some of our discoveries there.

In the meantime, can one of you approve this PR?

bartgol (approved):

ambrad:
https://github.com/E3SM-Project/scream/pull/1797 is using this new capability; see https://github.com/E3SM-Project/scream/pull/1797#issuecomment-1278205887 for a preliminary report of success.

PR #1941, opened 2022-09-13
Bugfix to re-initialize views in output after each write step

AaronDonahue:
This commit is a bugfix that was discovered when writing output for averaging type MIN over multiple output snaps.  We were not re-init the local views so the total minimum remained the same from snap to snap.

This commit has the output manager re-initialize local views in an output stream following each write of a timesnap.

### DIFF
--- a/components/scream/src/share/io/scorpio_output.cpp
+++ b/components/scream/src/share/io/scorpio_output.cpp
@@ -434,23 +434,33 @@ void AtmosphereOutput::register_views()
       m_dev_views_1d.emplace(name,view_1d_dev("",size));
       m_host_views_1d.emplace(name,Kokkos::create_mirror(m_dev_views_1d[name]));
 
-      // Init dev view with an "identity" for avg_type
-      switch (m_avg_type) {
-        case OutputAvgType::Instant:
-          // No averaging
-          break;
-        case OutputAvgType::Max:
-          Kokkos::deep_copy(m_dev_views_1d[name],-std::numeric_limits<Real>::infinity());
-          break;
-        case OutputAvgType::Min:
-          Kokkos::deep_copy(m_dev_views_1d[name],std::numeric_limits<Real>::infinity());
-          break;
-        case OutputAvgType::Average:
-          Kokkos::deep_copy(m_dev_views_1d[name],0);
-          break;
-        default:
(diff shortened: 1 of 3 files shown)


AaronDonahue:
B4B because we don't have any runs that use MIN and MAX output for the baselines.

components/scream/src/share/io/scorpio_output.cpp
```diff
@@ -434,23 +434,33 @@ void AtmosphereOutput::register_views()
...
-        case OutputAvgType::Average:
-          Kokkos::deep_copy(m_dev_views_1d[name],0);
-          break;
-        default:
-          EKAT_ERROR_MSG ("Unrecognized averaging type.\n");
-      }
+    }
+  }
+  // Initialize the local views
+  initialize_dev_views();
```
bartgol, line 440: May I suggest "reset" instead of "initialize". The latter sounds like something you do only once in the lifetime of the views, while this is (in fact, _has to be_) done after every write as well.

components/scream/src/share/io/scream_output_manager.cpp
```diff
@@ -287,12 +287,15 @@ void OutputManager::run(const util::TimeStamp& timestamp)
...
     ++filespecs.num_snapshots_in_file;
 
-    // Now that we've written output to this file we need reset the nsteps.
-    control.nsamples_since_last_write = 0;
-
-    // Since we wrote to file we need to reset the nsamples_since_last_write and the timestamp
+    // Since we wrote to file we need to reset the nsamples_since_last_write, the timestamp
     control.nsamples_since_last_write = 0;
     control.timestamp_of_last_write = timestamp;
+    // and the local views - unless it is a checkpoint step, then we keep local views.
```
bartgol, line 293: Seems like the comment sentence is broken up by the two lines of code. I would either put all comments together, or put some `...` at the end of the first and beginning of the latter, to show that it's the same comment.

bartgol (approved):
Looks good. I have a couple of style comments, feel free to ignore.

AaronDonahue:
@bartgol , thanks for the quick review.  I'm going to add automerge back as a label and if its ready you can approve and it will go right in.

bartgol (approved):

PR #1943, opened 2022-09-14
Sliced output diagnostic

bartgol:
Adds a diagnostic for sliced output. Still need to setup syntax to request sliced output in yaml file, but the diagnostic itself works.

### DIFF
--- a/components/scream/src/share/field/field_manager.cpp
+++ b/components/scream/src/share/field/field_manager.cpp
@@ -36,15 +36,11 @@ void FieldManager::register_field (const FieldRequest& req)
     m_fields[id.name()] = std::make_shared<Field>(id);
   } else {
     // Make sure the input field has the same layout and units as the field already stored.
-    // NOTE: we allow the input units to be "invalid", in case the customer requesting
-    //       does not really care about them. E.g., a diagnostic that deals with the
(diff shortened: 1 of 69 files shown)


components/scream/src/share/field/field_manager.cpp
```diff
@@ -53,11 +36,15 @@ void FieldManager::register_field (const FieldRequest& req)
     m_fields[id.name()] = std::make_shared<Field>(id);
   } else {
     // Make sure the input field has the same layout and units as the field already stored.
+    // NOTE: we allow the input units to be "invalid", in case the customer requesting
```
AaronDonahue, line 39: In the usage of the new diagnostic isn't it possible to add the field units to the parameter list that is passed at construction, so that we can retain the units?  Presumably, if someone is only looking a output file with surface values they won't want to have to query the default output to see what the units should have been.

AaronDonahue (commented):
Looks good, I just had one question before I approve and let the AT automerge.

components/scream/src/share/field/field_manager.cpp
bartgol: I can add that.

AaronDonahue (approved):

PR #1945, opened 2022-09-14
Introduce a general horizontal remapper v2

AaronDonahue:
This commit introduces a horizontal remapping structure fashioned on the approach used in the component coupler.

The new horizontal remapper has better memory management, enabling the use of very fine resolution remapping (i.e. ne30->ne1024) without hitting OOM errors.

This commit also introduces a robust randomized test which compares the use of the remapper given local information or remapping data provided by file and comparing everything against a baseline.

This PR appears to be Non-BFB, but at a low relative difference.  This is attributable to applying the weights for remapping in a slightly different order than the original simulation.

Replaces #1891.


AaronDonahue:
Some memory usage and timing info is coming.  Also a better measure of the non-BFB ness.  In the meantime this can still be reviewed.

components/scream/src/physics/spa/spa_functions_impl.hpp
```diff
@@ -300,32 +301,30 @@ perform_vertical_interpolation(
...
-  });
-  // Determine the set of unique columns in this remapping
-  spa_horiz_interp.set_unique_cols();
+  auto& spa_gsmap = spa_horiz_interp.gsmap;
+  spa_gsmap = GSMap(spa_horiz_interp.m_comm,"SPA 1-1 Remap",dofs_gids,min_dof);
+  view_1d<gid_type> src_dofs("",1);
+  view_1d<Real>     src_wgts("",1);
+  auto dofs_gids_h = Kokkos::create_mirror_view(dofs_gids);
+  Kokkos::deep_copy(dofs_gids_h,dofs_gids);
+  for (int ii=0;ii<num_local_cols;ii++) {
```
ambrad, line 319: This is done once at model init, right?

components/scream/src/physics/spa/spa_functions_impl.hpp
```diff
@@ -334,113 +333,18 @@ template <typename S, typename D>
 void SPAFunctions<S,D>
 ::get_remap_weights_from_file(
```
ambrad, line 334: And, tbc, this routine is all run once and only at model init?

components/scream/src/physics/spa/spa_functions_impl.hpp
```diff
@@ -496,62 +400,60 @@ ::update_spa_data_from_file(
           SPAHorizInterp&       spa_horiz_interp,
           SPAInput&             spa_data)
 {
+  start_timer("EAMxx::SPA::update_spa_data_from_file");
```
ambrad, line 403: This routine is run once a month and is the possible performance bottleneck, right?

One thing that would help is to add a bunch of timers throughout this routine to assess the cost of each piece and report performance results in a comment. If the slowdown is in something other than reading the data from the file, then there's quite a lot that can be done to speed up the code.

I'm also fine with merging this PR and then doing a separate speedup PR.

ambrad (commented):
Is this supposed to be BFB for v1 tests? If so, please mark it as such in the PR header comment. It looks like a relevant test is failing on weaver, so I won't approve yet.

components/scream/src/physics/spa/spa_functions_impl.hpp
AaronDonahue: Yes, we only gather remap weights once.

components/scream/src/physics/spa/spa_functions_impl.hpp
AaronDonahue: I'll add more timers for my timing check and report back.

components/scream/src/physics/spa/spa_functions_impl.hpp
AaronDonahue: yes, just once, for cases where there is no remap.  Such as in an ne30np4 case where the data is already on the native grid.

components/scream/src/physics/spa/spa_functions.hpp
```diff
@@ -149,46 +150,14 @@ struct SPAFunctions
     // The weights stores the remapping weight to be applied to the source grid data for this location
     //   in the target data.
     SPAHorizInterp() = default;
-    SPAHorizInterp(const int length_)
+    SPAHorizInterp(const ekat::Comm& comm)
```
jeff-cohere, line 153: Let's add the `explicit` keyword to this single-argument constructor.

components/scream/src/physics/spa/spa_functions.hpp
```diff
@@ -149,46 +150,14 @@ struct SPAFunctions
...
     {
-      length = length_;
-      weights = view_1d<Real>("",length_);
-      source_grid_loc = view_1d<Int>("",length_);
-      target_grid_loc = view_1d<Int>("",length_);
+      m_comm = comm;
     }
+    // Horizontal Remap
+    GSMap gsmap;
```
jeff-cohere, line 158: Should we call this type `HorizontalRemap` or `HorizontalMapping` or ...?

components/scream/src/physics/spa/spa_functions_impl.hpp
jeff-cohere: Worth documenting in the comment above the function?

components/scream/src/physics/spa/spa_functions_impl.hpp
jeff-cohere: als0 als0 doc

components/scream/src/share/grid/remap/horizontal_remap_utility.hpp
```diff
@@ -0,0 +1,163 @@
...
+#include "share/scream_types.hpp"
+#include "ekat/kokkos/ekat_subview_utils.hpp"
+
+#include <numeric>
+
+namespace scream {
+
+/*===============================================================================================*/
+/*                                        GSSegment
+ * Lightwieght structure to represent a single remapping of source data to a single column.
```
jeff-cohere, line 15: Lightwieght -> Lightweight. Also, worth saying "single target column" instead of "single column"?

components/scream/src/share/grid/remap/horizontal_remap_utility.hpp
```diff
@@ -0,0 +1,163 @@
...
+ *     Y_target = sum_(n=1)^N ( w_n * Y_source_n )
+ * See description of GSMap structure for more details on the mapping.
+ *
+ * This structure is used to organize the overall horizontal remapping stored in the GSMap structure.
+ * Each segment represents a single target column in the GSMap.
```
jeff-cohere, line 20: Ah, I see you used the term here. Great.

components/scream/src/share/grid/remap/horizontal_remap_utility.hpp
```diff
@@ -0,0 +1,163 @@
...
+  // Getter Functions
+  gid_type get_dof()     const { return m_dof; }
+  Int      get_dof_idx() const { return m_dof_idx; }
+  Int      get_length()  const { return m_length; }
+  view_1d<gid_type> get_source_dofs() const { return m_source_dofs; }
+  view_1d<Int>      get_source_idx()  const { return m_source_idx; }
+  view_1d<Real>     get_weights()     const { return m_weights; }
+ 
+  // TODO: Not sure why, but this can't be set as private, otherwise `set_dof_idx` doesn't work. 
+  Int      m_dof_idx = -999; // The degree of freedom w.r.t. to the local index for this map
```
jeff-cohere, line 64: Why are you initializing this variable here and not in the constructor?

jeff-cohere: Also, is there a different between `int` and `Int`? If not, let's ditch `Int`.

components/scream/src/share/grid/remap/horizontal_remap_utility.hpp
```diff
@@ -0,0 +1,163 @@
...
+  template <typename S>
+  using view_2d = typename KT::template view_2d<S>;
+  
+  template <typename S>
+  using view_3d = typename KT::template view_3d<S>;
+
+  
+public:
+  // Constructors/Destructor
+  virtual ~GSMap() = default;
```
jeff-cohere, line 112: Doesn't need to be virtual

components/scream/src/share/grid/remap/horizontal_remap_utility.hpp
```diff
@@ -0,0 +1,163 @@
...
+
+  // Getter functions
+  view_1d<gid_type>      get_unique_source_dofs() const { return m_unique_dofs; }
+  Int                    get_num_unique_dofs() const { return m_num_unique_dofs; }
+  Int                    get_num_of_dofs() const { return m_num_dofs; }
+  std::vector<GSSegment> get_map_segments() const { return m_map_segments; }
+  Int                    get_num_of_segments() const { return m_num_segments; }
+  
+
+protected:
```
jeff-cohere, line 141: Can be `private`

components/scream/src/share/grid/remap/horizontal_remap_utility.hpp
```diff
@@ -0,0 +1,163 @@
...
+  Int                    get_num_of_dofs() const { return m_num_dofs; }
+  std::vector<GSSegment> get_map_segments() const { return m_map_segments; }
+  Int                    get_num_of_segments() const { return m_num_segments; }
+  
+
+protected: 
+
+  // Global degrees of freedom information on target grid
+  view_1d<gid_type> m_dofs_gids;
+  Int               m_num_dofs = 0;
```
jeff-cohere, line 145: Use `int` unless there's a reason to use `Int`

components/scream/src/share/grid/remap/horizontal_remap_utility.cpp
```diff
@@ -0,0 +1,543 @@
...
+    auto seg = m_map_segments[iseg];
+    auto seg_dof_idx = seg.get_dof_idx();
+    auto remap_sub = Kokkos::subview(remapped_data,seg_dof_idx,Kokkos::ALL);
+    seg.apply_segment(num_levs, source_data, remap_sub);
+  }
+  stop_timer("EAMxx::GSMap::apply_remap_2d");
+}
+/*-----------------------------------------------------------------------------------------------*/
+// This overload of apply remap assumes a set of horizontal slices of source data being mapped onto
+// a set horizontal slices of remapped data.  The assumption is that there are levels and one other
```
jeff-cohere, line 416: a set "of" horizontal slices?

jeff-cohere (approved):
Looks good to me. I left several minor comments. Up to you whether to incorporate my suggestions before you merge. Approving, assuming the test failing only because of the minor changes in data from the reordered ops you mentioned.

components/scream/src/share/grid/remap/horizontal_remap_utility.hpp
bartgol: Re `int` vs `Int`. I think the odds we'll ever switch the typedef to something different from `int` are miniscule. I always use `int`, but that's my preference. OTOH, I'm in favor of using `intXYZ_t` (X=64,32.16...) when the size is important.

bartgol: As for init-ing at declaration time (inside the class), I like it, actually. It is especially useful when the value is something true, regardless of which ctor is called (I use it a lot for stuff like `m_is_inited`, which should always be false upon construction).

More info: https://isocpp.org/wiki/faq/cpp11-language-classes#member-init

AaronDonahue:
>Is this supposed to be BFB for v1 tests? If so, please mark it as such in the PR header comment. It looks like a relevant test is failing on weaver, so I won't approve yet.
@ambrad it is in principle BFB but it turns out to be technically non-BFB at a relatively small error.  My hypothesis is that because we looping through the weights in the cumulative sum slightly differently that we get small changes and hence cprnc reports DIFF.  Interestingly that only happens for the two ERS tests.  So maybe that's a legit fail that needs to be investigated.

As for the `weaver` fail, that is for a spa unit test apply the remap and only happens for SP and looks like a tolerance thing in a X==Y check.  Although the tolerance appears to be higher than 100*macheps but smaller than 1000*macheps.  So again, maybe a real error, although this doesn't happen for any other SP tests and doesn't happen at all in the DP tests.  And only happens on weaver.

components/scream/src/share/grid/remap/horizontal_remap_utility.hpp
AaronDonahue: Yes, my motivation for init'ing like this is because this is the init value regardless of constructor.  As for `int` vs. `Int`.  I'm pretty agnostic on the syntax.  I can switch everything to `int`.  I recall we tried sticking with `Int` as a declared SCREAM type near the beginning of the project, like 'Real', and I've tried to stick with that convention ever since.

components/scream/src/physics/spa/spa_functions.hpp
AaronDonahue: Great. I'll change the name.  We hadn't settled the debate in the old PR so I stuck with GSMap and made sure it would be something I could easily find/replace.  I'll use `HorizontalRemap` unless there is a strong favor for the other suggestion.

AaronDonahue:
@bartgol , @ambrad , @jeff-cohere with regards to the failed test on weaver in sp.  I'm in favor of using the 1000*macheps tolerance to get it passing rather than chase down exactly why it isn't exactly ==.  More for the sake of time.  TBH, this test is old and I don't think I did a really good job setting it up.  It checks that SPA read data and remapped it, which is essentially a repeat of the newer better test this PR brings in.  It's good to check if someone misues the remapper in the SPA infrastructure, so we should keep it, but I'm not bothered that at SP it's not perfect.

I'm open to push back from you though.

AaronDonahue:
Re: the two DIFFs for ERP and ERS respectively.  The main difference between passing and DIFFing tests is that both optimized runs don't pass baselines, the other three are debug `_D` builds.  Could someone w/ access to `mappy` check if those DIFFs with the baseline are large, or relatively small?  Is it reasonable to expect that when we do something like change loop order an optimized build may have a small DIFF?

bartgol:
> /ascldap/users/e3sm-jenkins/acme/scratch/ERS_Ln9.ne4_ne4.F2000-SCREAMv1-AQP1.mappy_gnu9.C.20220915_131857_lxaqd5

They're large:

```
RMS PotentialTemperature             2.8247E-02            NORMALIZED  5.6301E-05
```

ambrad:
@bartgol I'm thinking that error is actually fine. Any perturbation in a simulation tends to lead to errors of that magnitude in the final time step.

ambrad (approved):
Sorry I missed the bit in the PR header comment describing the change in summation order. I agree that will lead to non-BFB results.

bartgol:
Is the read from data test comparing the value computed by `update_spa_data_from_file` with the analytic formula used to generate the data? Does the update do time/space interpolation in the process? And does interpolation perform any reduction? If so, then I guess we could expect the order of ops to be different when it runs on GPU. But I would not expect *huge* diffs.

One thing we could do is to use small integers for weights and data (few non-integers powers of 2 also work). That way we can be sure that the order of ops is not impacting the result.

AaronDonahue:
> 

Yes, I created a netCDF file of SPA data using python and a netCDF remap file of weights following the convection w_n = 1/2^(n+1) with the last weight adjusted so sum(w_n)=1.0.  But in SP with something like 20 columns in the mapping I'm guessing that 1/2^n is a bad choice for weight.  Is that a correct assumption.  I would be in favor of overhauling this test altogether.  Probably following the example in the horizontal_remap test I created here.  Making the map and weights on the fly, writing to file and then using SPA and that file to get SPA values.  Then we can compare directly and in the appropriate precision.

AaronDonahue:
This now looks like its passing everywhere we expect and we just have the two DIFFs on mappy CIME cases which we've discussed.  I have one more requested change - the name of the structure.  Then, if it passes another round of testing I will manually merge, considering that those changes are cosmetic and shouldn't impact the approval status.

Note, I went ahead and loosened the tolerance for the one test that failed SP on weaver.  There is still time to push back on that.

components/scream/src/physics/spa/spa_functions_impl.hpp
AaronDonahue: @ambrad , @jeff-cohere and @bartgol .  The preliminary timer info is in and it looks pretty bad for the apply_remap call for the 4 banded variables which are 3D.   Here's a taste of timing on summit, ne30pg2

```
"a_i:EAMxx::SPA::update_spa_data_from_file"                -         30       30 6.000000e+01   6.618477e+03   220.617 (    10      0)   220.614 (     1      0)
"a_i:EAMxx::SPA::update_spa_data_from_file::read_data"     -         30       30 6.000000e+01   3.921619e+02    13.074 (    11      0)    13.070 (    19      0)
"a_i:EAMxx::SPA::update_spa_data_from_file::apply_remap"   -         30       30 6.000000e+01   6.055629e+03   207.534 (     4      0)   196.535 (    24      0)
"a_i:EAMxx::GSMap::apply_remap_1d"                         -         30       30 6.000000e+01   1.459733e+00     0.050 (     4      0)     0.047 (    23      0)
"a_i:EAMxx::GSMap::apply_remap_2d"                         -         30       30 6.000000e+01   1.038210e+02     3.574 (     4      0)     3.381 (    24      0)
"a_i:EAMxx::GSMap::apply_remap_3d"                         -         30       30 2.400000e+02   5.949352e+03   203.876 (     4      0)   193.073 (    24      0)
"a_i:EAMxx::SPA::update_spa_data_from_file::copy_and_pad"  -         30       30 6.000000e+01   1.470893e-02     0.001 (    16      0)     0.000 (    21      0)
```

There are 14 swbands and 16 lwbands in the third dimension, and 4 of those in total, so 3*14 + 16 = 58.  So 3D remapping is 58X more work.  So that explains why 3D remapping is around 60X more time.

I'm running more timing so I can compare master.  But I mention this now in case it can spur creative ideas about speeding up this chunk of code.  It is the biggest bottleneck that I can see.

AaronDonahue: Here is current `master`, same run

```
"a_i:EAMxx::SPA::update_spa_data_from_file"               -         30       30 6.000000e+01   4.059888e+02    13.538 (    25      0)    13.531 (     3      0)
"a_i:EAMxx::SPA::update_spa_data_from_file::read_data"    -         30       30 6.000000e+01   3.368266e-01     0.016 (    25      0)     0.009 (     0      0)
"a_i:EAMxx::SPA::update_spa_data_from_file::apply_remap"  -         30       30 6.000000e+01   3.573152e+02    13.512 (     0      0)    10.239 (     1      0)
```

Note, here remapping is one monolithic set of for loops, so there is no breakdown by dimension like above.  Still, looking at more than 10X a slowdown overall.  Current hypothesis is the host/device copies in the new remapper.  This can be improved, so I am going to investigate next.

ambrad: Just to be clear, these timers are for init (a_i prefix => atm init). 1. Do these same timers apply when time stepping? 2. That aside, my sense is that this init time is really bad, perhaps the vast majority of the init time for this case (ne30pg2 on 5 nodes of Summit, I suspect). True?

AaronDonahue: The times I report above are the sum of all times the "update_spa_data" is called.  In this case twice, at initialization.  If I ran for at least a month there would be another 3rd call then to update to the next month of data.   But more to the point, yes, a huge % of init is in this one section.  Total init times are

```
master:
"a_i:EAMxx::init"                                         -         30       30 1.800000e+02   7.822712e+02    26.098 (    10      0)    26.053 (    12      0)
new remap:
"a_i:EAMxx::init"                                          -         30       30 1.800000e+02   6.977105e+03   232.574 (    10      0)   232.555 (     1      0)
```

Just to be clear.  SPA initialization calls the data update twice to get things started.  But updating data is called once per month.

ambrad: Ok, thanks. Staying completely on host, as discussed on slack, seems like the obvious alternative to try.

AaronDonahue:
@ambrad , I have confirmed that switching to doing everything on Host for applying the remapper sped everything up by ~20X.  It may be possible to eventually switch everything to Device.  But in the interest of time I am going to push my changes for an all-Host application of the remapper and then we can build from there.

AaronDonahue:
@jeff-cohere , I just made the changes to the syntax.  I think this PR should be ready to merge pending one last AT run.

bartgol:
Can I ask what "Segment" is referring to in this context?

AaronDonahue:
> Can

Absolutely.  It is just organizational.  In the remapping we have N source cols which map to 1 target column.  Here I am calling this part of the mapping a "remap segment".   The motivation came from reading what is done in CPL for their mapping.  That being said, I'm not strictly wedded to this feature if we want to go a different route.  

Although, in the interest of the science campaign runs it might be good to keep segments for now and revisit in the future, so that this PR can get in sooner than later.

jeff-cohere:
Looks like the build/test workflow timed out under Blake. Is this something we've been seeing? Do we need to trim some fat or add to the 180 minute threshold?

components/scream/src/share/grid/remap/horizontal_remap_utility.hpp
```diff
@@ -0,0 +1,165 @@
...
+namespace scream {
+
+/*===============================================================================================*/
+/*                                        HorizontalMapSegment
+ * Lightweight structure to represent a single remapping of source data to a single target column.
+ *     Y_target = sum_(n=1)^N ( w_n * Y_source_n )
+ * See description of HorizontalMap structure for more details on the mapping.
+ *
+ * This structure is used to organize the overall horizontal remapping stored in the HorizontalMap structure.
+ * Each segment represents a single target column in the HorizontalMap.
```
jeff-cohere, line 20: Thanks for these comments!

components/scream/src/share/grid/remap/horizontal_remap_utility.hpp
```diff
@@ -0,0 +1,165 @@
...
+
+  // Getter functions
+  view_1d<gid_type>      get_unique_source_dofs() const { return m_unique_dofs; }
+  int                    get_num_unique_dofs() const { return m_num_unique_dofs; }
+  int                    get_num_of_dofs() const { return m_num_dofs; }
+  std::vector<HorizontalMapSegment> get_map_segments() const { return m_map_segments; }
+  int                    get_num_of_segments() const { return m_num_segments; }
+  long long              get_peak_memory_usage() const { return m_peak_mem_usage; }
+  
+protected:
```
jeff-cohere, line 140: Can be changed to `private` since this is not a base class.

jeff-cohere (dismissed):
Thanks for all the comments, Aaron. I made a pass and I believe it's easier for someone to figure out what's going on now. Approving subject to resolving test failures (and the comments of others, of course).

AaronDonahue:
> Looks like the build/test workflow timed out under Blake. Is this something we've been seeing? Do we need to trim some fat or add to the 180 minute threshold?

There were some mem usage statistics left in the PR that were slowing things down due to an all-reduce.  I've removed that commit so things should run faster now.

AaronDonahue:
@jeff-cohere , @bartgol and/or @ambrad .  I think this is finally ready.  I've stripped the mem usage changes, `blake` now finishes, when before it timed out, and the only FAILS are two DIFFs on `mappy`.  I suppose my only outstanding concern is that we are not non-BFB for all of the CIME based tests.  But maybe I'm missing something?

AaronDonahue:
For context.  I ran master and this branch for the `homme_shoc_cld_spa_p3_rrtmgp` test and did a cprnc comparison on the two output files and they match.  So it would seem to be BFB, at least in that context.

ambrad:
> I suppose my only outstanding concern is that we are not non-BFB for all of the CIME based tests. But maybe I'm missing something?

I think the key question is whether in the matrix-vector product `y = A*x`, `y(i) = A(i,:)*x` is being computed with a different column order than in the original SPA impl. If yes, then we should expect all nontrivial tests to diff a little.

AaronDonahue:
> > I suppose my only outstanding concern is that we are not non-BFB for all of the CIME based tests. But maybe I'm missing something?
> 
> I think the key question is whether in the matrix-vector product `y = A*x`, `y(i) = A(i,:)*x` is being computed with a different column order than in the original SPA impl. If yes, then we should expect all nontrivial tests to diff a little.

Yes, I think that is safe to say.

ambrad (approved):

jeff-cohere (approved):

AaronDonahue:
@bartgol are you good with me merging this now?

components/scream/src/share/grid/remap/horizontal_remap_utility.cpp
```diff
@@ -0,0 +1,543 @@
...
+  chunk_displacement[0] = 0;
+  total_num_chunks = num_chunks_per_rank[0];
+  for (int ii=1; ii<num_ranks; ii++) {
+    chunk_displacement[ii] = total_num_chunks;
+    total_num_chunks += num_chunks_per_rank[ii];
+  }
+  int* buff_dof = (int*)calloc(total_num_chunks, sizeof(int));
+  int* buff_sta = (int*)calloc(total_num_chunks, sizeof(int));
+  int* buff_len = (int*)calloc(total_num_chunks, sizeof(int));
+  MPI_Allgatherv(chunk_dof.data(),  chunk_dof.size(),MPI_INT,buff_dof,num_chunks_per_rank,chunk_displacement,MPI_INT,m_comm.mpi_comm());
```
bartgol, line 123: FYI, `ekat::Comm` has some interfaces that make MPI calls a bit more understandable.

AaronDonahue: I looked into that for `allgather` but if I recall either the `allgatherv` wasn't an option, or I kept hitting an error when I tried the implementation in this context.

bartgol: Ah, I missed the `v` part, sorry.

PR #1948, opened 2022-09-15
Add initial condition and defaults for ne256

brhillman:
Add initial condition and defaults for ne256 EAMxx configurations.


ndkeen (approved):
I tried a ne256 case using this default IC on cori-knl. Ran about 22 steps before i timed out as the IO is really slow with master now.

PR #1949, opened 2022-09-16
Minor bugfixes for solar insolation

brhillman:
Minor bugfixes for solar insolation inconsistencies with EAM/SCREAMv0. This includes an off-by-one bug in setting the time used by the orbital parameter routines, which expect time `calday` to be in day + fraction of day, where time 0Z on Jan 01 corresponds to `calday = 1.0` while our `ts.frac_of_year_in_days()` routine gives `0` on 0Z Jan 01. Also set orbital year to a constant value of 2010 for F2010 compsets. Before we let this be set by the simulation date, which meant by default we were setting the orbital year to be 0001 for F2010 cases. For real cases, we do probably want this to be set by the simulation timestamp, or else manually set to be consistent with the simulated year, but for cyclic forcing experiments we want to set orbital year to a constant. Also add output for `cosine_solar_zenith_angle`, and allow using reduced gpoint data with SCREAMv0 configurations for easier comparisons.

### DIFF
--- a/components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
+++ b/components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
@@ -144,6 +144,7 @@ void RRTMGPRadiation::set_grids(const std::shared_ptr<const GridsManager> grids_
   add_field<Computed>("sfc_flux_dif_vis", scalar2d_layout, Wm2, grid_name);
   add_field<Computed>("sfc_flux_sw_net" , scalar2d_layout, Wm2, grid_name);
   add_field<Computed>("sfc_flux_lw_dn"  , scalar2d_layout, Wm2, grid_name);
+  add_field<Computed>("cosine_solar_zenith_angle",scalar2d_layout,nondim,grid_name);
 }  // RRTMGPRadiation::set_grids
 
 size_t RRTMGPRadiation::requested_buffer_size_in_bytes() const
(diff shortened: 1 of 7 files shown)


brhillman:
Some analysis. Before this PR, the solar zenith angle would be slightly off at each timestep. Example, v0 vs v1 on the first timestep:
image
while after implementing this fix:
image
This leads to very small differences in globally-averaged solar insolation, but up to on the order of a W/m^2 locally:
image
These errors are reduced to a fraction of a W/m^2 with the bugfix:
image

components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
```diff
@@ -144,6 +144,7 @@ void RRTMGPRadiation::set_grids(const std::shared_ptr<const GridsManager> grids_
   add_field<Computed>("sfc_flux_dif_vis", scalar2d_layout, Wm2, grid_name);
   add_field<Computed>("sfc_flux_sw_net" , scalar2d_layout, Wm2, grid_name);
   add_field<Computed>("sfc_flux_lw_dn"  , scalar2d_layout, Wm2, grid_name);
+  add_field<Computed>("cosine_solar_zenith_angle",scalar2d_layout,nondim,grid_name);
```
AaronDonahue, line 147: Has this been added as a "computed" field for a reason?  Was it just to be able to see it as output in debugging?  If so, do we want to keep it like this or put it back in the mem buffer?

AaronDonahue (dismissed):
Looks good.  I had one question, it shouldn't hold up merging this though.  Although think it is not BFB so we shouldn't have automerge occur anyway.

components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
brhillman: I wanted to be able to output it. It's an output from EAM as well, so allows comparing with SCREAMv0/EAM. We probably won't regularly look at it, but it was pretty useful to have for this exercise. We can revert it back to a non-tracked field in the mem buffer if you think that's better. I don't think it will impact performance either way.

brhillman: Pushed a fix. I don't think it helps mem right now, but would if we were to use a unified buffer. No sense polluting the FM with fields we won't use anyways.

AaronDonahue (approved):

AaronDonahue:
This is expected as a non-BFB PR.  Looks like its passing everywhere else and is approved.  I'm going manually merge.

PR #1952, opened 2022-09-19
EAMxx: fix detection of CIME build in scorpio F90 interface

bartgol:
While removing some unused code I accidentally removed the scream config file inclusion, so the `SCREAM_CIME_BUILD` macro was not picked up.

### DIFF
--- a/components/scream/src/share/CMakeLists.txt
+++ b/components/scream/src/share/CMakeLists.txt
@@ -35,6 +35,14 @@ target_include_directories(scream_share PUBLIC ${SCREAM_SRC_DIR} ${SCREAM_BIN_DI
 target_link_libraries(scream_share PUBLIC ekat gptl)
 set_target_properties(scream_share PROPERTIES Fortran_MODULE_DIRECTORY ${SCREAM_F90_MODULES})
 target_compile_options(scream_share PUBLIC $<$<COMPILE_LANGUAGE:Fortran>:${SCREAM_Fortran_FLAGS}>)
+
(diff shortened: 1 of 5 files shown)


components/scream/src/share/CMakeLists.txt
```diff
@@ -35,6 +35,14 @@ target_include_directories(scream_share PUBLIC ${SCREAM_SRC_DIR} ${SCREAM_BIN_DI
 target_link_libraries(scream_share PUBLIC ekat gptl)
 set_target_properties(scream_share PROPERTIES Fortran_MODULE_DIRECTORY ${SCREAM_F90_MODULES})
 target_compile_options(scream_share PUBLIC $<$<COMPILE_LANGUAGE:Fortran>:${SCREAM_Fortran_FLAGS}>)
+
+# This used to be in `scream_config.h/scream_config.f`, but we did accidentally remove
+# the includes at least twice. To avoid this from happening again, we add this crucial
+# macro as a compile definition of the scream_share, which is linked to all other
+# scream libraries.
+if (SCREAM_CIME_BUILD)
+  target_compile_definitions (scream_share PUBLIC SCREAM_CIME_BUILD)
```
bartgol, line 44: This should get propagated to all targets linking against scream_share (which is all atm targets, plus E3SM executable). I think it is safer than including a header, since we're highly unlikely to accidentally delete this line, while it is possible (in fact, it happened!) to delete a line in a source file, or even to forget to add the include in a new source file (this also happened!).

jgfouca (approved):
Good change!

PR #1954, opened 2022-09-20
Update files used w/ ne1024 runs in a CIME case.

AaronDonahue:
This commit updates the initial condition file to be compatible with the coarse pressue gradient formulation.

It also update the default namelist parameters to have a default SPA mapping file for getting ne1024pg2 data from the ne30np4 dataset.


AaronDonahue:
Is B4B for everything except out-of-the-box ne1024pg2 runs.

brhillman (changes requested):
I think this line should be changed to point to the new initial condition as well:
```
    <Filename hgrid="ne1024np4">${DIN_LOC_ROOT}/atm/scream/init/screami_ne1024np4L128_era5_20131001_20220823.nc</Filename> 
```

brhillman (approved):

PR #1956, opened 2022-09-23
HighResMIP 0.25x0.25 daily sstice for 2010 SCREAMv1 at ne120 or higher

wlin7:
Changes to use HighResMIP 2013 daily sstice data for SCREAMv1 F compset at
ne120, ne256, ne512 and ne1024 resolutions. The current default uses the standard
1 degree monthly climatology centered around 2010 sstice data for EAM, 
sst_ice_CMIP6_DECK_E3SM_1x1_2010_clim_c20190821.nc, which will still be used 
for lower resolution configuration after this PR.

[BFB] for SCREAMv1 tests at resolution lower than ne120.

### DIFF
--- a/components/data_comps/docn/cime_config/config_component.xml
+++ b/components/data_comps/docn/cime_config/config_component.xml
@@ -185,6 +185,7 @@
       <value compset="1850_" grid="a%0.23x0.31.*_oi%0.23x0.31"					>$DIN_LOC_ROOT/atm/cam/sst/sst_HadOIBl_bc_0.23x0.31_clim_pi_c101028.nc</value>
       <value compset="1950_" grid=".+"								>$DIN_LOC_ROOT/ocn/docn7/SSTDATA/sst_ice_CMIP6_DECK_E3SM_1x1_1950_clim_c20180910.nc</value>
       <value compset="2010.*_" grid=".+"					        >$DIN_LOC_ROOT/ocn/docn7/SSTDATA/sst_ice_CMIP6_DECK_E3SM_1x1_2010_clim_c20190821.nc</value>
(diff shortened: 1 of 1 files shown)


PeterCaldwell (approved):
Good catch, Wuyin!

wlin7:
> @wlin7 , there are two such SST files I created. One is for year 2013 only, and the other is for the climatology at that decade. Which one is intended to be used here?

To use year 2013 only, Qi. Thanks for reviewing.

tangq:
@wlin7 , I dug out this [PR](https://github.com/E3SM-Project/E3SM/pull/3119/files), and this [simulation](https://acme-climate.atlassian.net/wiki/spaces/SIM/pages/978714703/cori-knl.20190214+maint-1.0.F2010-CMIP6-HR.dailySST.noCNT.ne120+oRRS18v3) for the high-res (ne120) paper. I confirmed the following:

The year 2013 high-res file
/global/cfs/cdirs/e3sm/inputdata/ocn/docn7/SSTDATA/sst_ice_CMIP6_HighResMIP_E3SM_0.25x0.25_2010_clim_c20190125_intoisst.nc

The 2015-2014 climatology low-res file (under the same directory)
sst_ice_CMIP6_DECK_E3SM_1x1_2010_clim_c20190821.nc

It turns out that we didn't create a high-res climatology file.

In conclusion, the SST file you changed to use is the right one.

tangq (approved):
It's the right 0.25x0.25 daily SST/sea ice file with data from year 2013.

wlin7:
> It's the right 0.25x0.25 daily SST/sea ice file with data from year 2013.

Thanks for the careful review, @tangq . Yes, I can see this sstice file is specified in the config_component.xml file in the maint-1.0 branch for ne120 configuration.

PR #1957, opened 2022-09-24
This commit fixes the zonal and meridonal vapor flux diagnostics.

AaronDonahue:
In the original implementation the u and v wind profile were specifically defined as a required field from the field manager.  In EAMxx these two fields are packed inside the `horiz_winds` field.  This commit switches the diagnostics to require `horiz_winds` and then uses the specfic velocity component associated with the diagnostic.

This commit also updates the tests to reflect the change in the diagnostics themselves.

Tested that this fix allows a typical EAMxx run to use these diagnostics as output.


hassanbeydoun (approved):
Thanks Aaron for taking care of this!

PR #1962, opened 2022-09-26
Fix pioc/piof interface libraries in CIME builds

bartgol:
The interface libraries were linking against the netcdf imported libraries, but were not adding the corresponding include directories. This was ok so long as we were not including pioc headers in scream. As soon as pio.h started to be included in scream, it caused issues with finding netcdf headers (included indirectly via pio.h).

I verified that this fix does indeed allow e3sm.exe to compile on Chrysalis.


jgfouca (approved):

AaronDonahue:
I'm testing this branch on `quartz` now to see if it will solve the issues there.

bartgol:
I have no clue why this PR has diffs in two tests. The only possibility is if somehow this changed what netcdf gets linked/included. But I ran on mappy, and looked at the generated cmake files, and they looked the same. Even a pre-processed cpp.i file in the scream_io lib looked the same.

ambrad:
@bartgol aren't these the same diffs @tcclevenger mentioned in the context of another PR?

tcclevenger:
Yes, these are the same ones.

AaronDonahue:
could be pointing to the DIFFs in the horiz remap not being simply baseline diffs.

AaronDonahue:
I've confirmed this fixes the issue on quartz with building

bartgol:
For the curious: mystery solved. If you're curious, the last non-bfb PR was merged yesterday, but its last AT run was days old (before _other_ non-bfb prs got merged). So when I blessed its AT run results, I blessed something that did not include stuff that was already in master, causing diffs. In fact, the same two diffs appear in the nightlies. I'm reblessing/retesting, and we should be good to go.

PR #1965, opened 2022-09-27
Fix passage of views to parallel Kokkos loops in horizontal_remap_utility.

AaronDonahue:
This commit fixes a bug in how views that existed within a structure were handled within Kokkos for loops.  Should fix errors we are seeing in cuda-mem-check fails on weaver.

### DIFF
--- a/components/scream/src/share/grid/remap/horizontal_remap_utility.cpp
+++ b/components/scream/src/share/grid/remap/horizontal_remap_utility.cpp
@@ -187,12 +187,11 @@ void HorizontalMap::set_dof_gids(const view_1d<const gid_type>& dofs_gids, const
   start_timer("EAMxx::HorizontalMap::set_dof_gids");
   EKAT_REQUIRE(dofs_gids.size()>0);
   m_dofs_gids = view_1d<gid_type>("",dofs_gids.size());
-  view_1d<gid_type> l_dofs_gids("",dofs_gids.size());
+  const auto l_dofs_gids = m_dofs_gids;
   m_num_dofs = m_dofs_gids.extent(0);
   Kokkos::parallel_for("", m_num_dofs, KOKKOS_LAMBDA (const int& ii) {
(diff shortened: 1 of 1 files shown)


components/scream/src/share/grid/remap/horizontal_remap_utility.cpp
```diff
@@ -187,10 +187,12 @@ void HorizontalMap::set_dof_gids(const view_1d<const gid_type>& dofs_gids, const
   start_timer("EAMxx::HorizontalMap::set_dof_gids");
   EKAT_REQUIRE(dofs_gids.size()>0);
   m_dofs_gids = view_1d<gid_type>("",dofs_gids.size());
+  view_1d<gid_type> l_dofs_gids("",dofs_gids.size());
   m_num_dofs = m_dofs_gids.extent(0);
   Kokkos::parallel_for("", m_num_dofs, KOKKOS_LAMBDA (const int& ii) {
-    m_dofs_gids(ii) = dofs_gids(ii)-min_dof;
+    l_dofs_gids(ii) = dofs_gids(ii)-min_dof;
   });
+  Kokkos::deep_copy(m_dofs_gids,l_dofs_gids);
```
AaronDonahue, line 195: would something like
```
m_dof_gids = view_1d<gid_type>(l_dofs_gids,m_num_dofs);
```
be better in this context?

ambrad: You could do this:
```C++
const auto l_dofs_gids = m_dofs_gids;
// then use l_dofs_gids as you already do in the parallel_for
// then no deep_copy
```
All you need to do here is avoid the implicit `this` access. So you just need to get a local view of m_dofs_gids for the lambda to capture, not create a separate array.

AaronDonahue: done

tcclevenger (approved):

ambrad (approved):

PR #1967, opened 2022-09-27
EAMxx/SHOC: Fix new nondeterminism.

ambrad:
We started seeing new nondeterministic FAILs in debug-build tests on ascent after a recent PR. I believe the PR itself, though it had a bug that was fixed in a subsequent PR, was not in fact responsible; I think it just revealed latent issues.

I traced the new nondeterminism to the calculation of wpthlp_sfc in SHOC's preprocess phase.

In this commit, I've cleaned up SHOC's preprocess phase. I've removed multiple spurious references. I've also combined parallel_for's that were needlessly separate loops. The spurious refs seemed to be the issue, as repeat testing of a reproducer switched to passing after those changes.

I've also added a new category of v1 tests to run only on GPU machines in our nightlies. These run three tests for many steps, increasing the probability we'll get nondeterministic FAILs to show up more quickly.

[BFB]

### DIFF
--- a/cime_config/tests.py
+++ b/cime_config/tests.py
@@ -509,6 +509,15 @@ _TESTS = {
             )
     },
 
+    "e3sm_scream_v1_long" : {
+        "time"  : "01:00:00",
+        "tests" : (
+            "ERP_D_Lh182.ne4pg2_ne4pg2.F2010-SCREAMv1",
+            "ERS_D_Ln182.ne4_ne4.F2010-SCREAMv1",
+            "ERS_Ln362.ne30pg2_ne30pg2.F2010-SCREAMv1"
+            )
+    },
+
     "e3sm_gpuacc" : {
         "tests"    : (
                  "SMS_Ld1.T62_oEC60to30v3.CMPASO-NYF",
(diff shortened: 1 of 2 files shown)


bartgol (approved):
I'm a bit puzzled on why splitting a pfor in two consecutive pfors can give  nondeterministic results, but if one pfor works, then it's obviously ok to use a single one.

ambrad:
My reproducer was ERP_D_Ln90.ne4_ne4.F2010-SCREAMv1.ascent_gnugpu, where I switched to Ln90 to increase the chance of failure. At the start, it always failed. Now 31 runs with 100% pass finished.

bartgol:
Uhm, const refs to temporaries are std compliant. I guess the problematic refs where those created outside a TTR pfor, and used inside it? I don't fully grasp why this would be a problem. I thought all threads in the team would execute those lines, so each of them would have a thread-local copy of the ref. Maybe I'm missing some cuda threads shenanigans...

ambrad:
Sorry, I should've been clearer. I suspect there was nothing wrong with the original code. However, using exclusive-or hashing at extremely fine granularity, I was able to isolate a nondeterministic diff to the wpthlp_sfc computation, despite all the inputs to that computation being BFB. Seeing nothing obviously wrong, I just started playing around with things. The first thing I tried was removing the const lvalue refs, just guessing there was something buggy in the lifetime analysis, and that worked.

PR #1968, opened 2022-09-28
Add ne30 L128 configuration

brhillman:
Add defaults for ne30 L128 and make it easier to change vertical resolution for other configurations as well. This also revamps how vertical coordinate files are specified, keying off `SCREAM_NUM_VERTICAL_LEV` instead of by horizontal grid and compset and adding a single L72 and L128 vertical coordinate file to be used for all horizontal resolutions. This makes it easier to run different horizontal resolutions with different vertical grids without having to create entirely new compsets and without having to specify the initial condition file twice, once for the vertical coordinate file and once for the actual initial condition filename. Two testmods are added to support forcing L72 or L128 for arbitrary horizontal resolutions, but a default alternative initial condition file is offered only for ne30 at the moment. To run ne30 with L128 instead of the default L72, simply use the `scream-L128` testmod, which will set `SCREAM_NUM_VERTICAL_LEV 128`, which will in turn set the proper `vertical_coordinate_filename` and `initial_conditions::Filename`. That is, something like:
```
cime/scripts/create_test SMS.ne30pg2_ne30pg2.F2010-SCREAMv1.pm-gpu_gnugpu.scream-L128
```
ought to work out of the box. If you are instead using `create_newcase`, you just need to change `SCREAM_CMAKE_OPTIONS` to use `SCREAM_NUM_VERTICAL_LEV 128`.

[B4B]


ambrad (approved):

ambrad:
1. Are the new files on the public_html data server?
2. Just to be clear, if using create_newcase, do the XML selectors mean that running `./xmlchange SCREAM_CMAKE_OPTIONS="SCREAM_NUM_VERTICAL_LEV 128` in the run script will make the right IC file get selected?

ambrad:
@oksanaguba fyi

brhillman:
> Are the new files on the public_html data server?

@ambrad yes.

AaronDonahue (approved):
so happy to see we have designated vertical coordinate files now.

oksanaguba:
It seems it requires also SCREAM_NUM_TRACERS 10 .

ambrad:
@oksanaguba re: "It seems it requires also SCREAM_NUM_TRACERS 10": Do you mean in your xmlchange line? I believe that is consistent with typical usage. I haven't tested --append, but w/o --append, if you mod the number of levels, you also have to reset the tracers to 10.

oksanaguba:
thanks -- i did not realize i am overwriting all options. i wonder what else i overwrote.

PR #1969, opened 2022-09-29
EAMxx: Skip calling free_decomp when a file is closed.

ambrad:
We're going to try keeping decomps around indefinitely for maximal reuse.

[BFB]

### DIFF
--- a/components/scream/src/share/grid/remap/horizontal_remap_utility.cpp
+++ b/components/scream/src/share/grid/remap/horizontal_remap_utility.cpp
@@ -57,8 +57,8 @@ void HorizontalMap::set_remap_segments_from_file(const std::string& remap_filena
     my_chunk += my_rank<remainder ? 1 : 0;
   }
   // now determine where this rank start reading the data.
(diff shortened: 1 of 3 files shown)


AaronDonahue (approved):

AaronDonahue:
@bartgol Is the AT timing out because of the overload you noted on mappy, on the infastructure Slack channel?

bartgol:
> @bartgol Is the AT timing out because of the overload you noted on mappy, on the infastructure Slack channel?

Yes.

ambrad:
@AaronDonahue when you get a chance, would you re-review this? I fixed a uniqueness issue with tag names. I also noticed some memory leaks in the SPA map-reading routine and fixed them by using `std::vector` instead of raw arrays.

AaronDonahue (approved):
Thanks for catching the remap issues and fixing those too.

ambrad:
In #1980 I reinstated the free_decomp line because there are tag uniqueness issues in specific runs. We'll have to investigate those before we can keep decomps persistent. But the memory leak is at least fixed.

PR #1970, opened 2022-09-29
Add vertically-projected cloud area output from radiation

brhillman:
Add a new output for vertically-projected cloud area (`cldtot`) computed from the subcolumn cloud optical depth used in the actual radiation computation. This provides a 2D cloud area fraction that is guaranteed to be consistent with the overlap assumption used in the radiation.

[B4B]

### DIFF
--- a/components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
+++ b/components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
@@ -851,7 +851,7 @@ void RRTMGPRadiation::run_impl (const int dt) {
 
     // Compute diagnostic total cloud area (vertically-projected cloud cover)
     auto cldtot = real1d("cldtot", ncol);
-    rrtmgp::compute_cloud_area(ncol, nlay, nswgpts, cld_tau_sw_gpt, cldtot);
+    rrtmgp::compute_cloud_area(ncol, nlay, nlwgpts, cld_tau_lw_gpt, cldtot);
 
     // Copy output data back to FieldManager
     if (update_rad) {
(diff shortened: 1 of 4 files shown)


bartgol (approved):

brhillman:
Performance penalty should be small from this extra calculation. Comparing timers from current master with this branch, on summit with ne30, for a 7 day simulation with 5 nodes and 6 MPI tasks per node:
```
[brhillman@login4.summit scripts]$ less ne30pg2_ne30pg2.F2010-SCREAMv1.master.a814294717aa10a4d101e0258f162adfd07f6cb2.before-cldtot-output/timing/e3sm_timing_stats.2478851.220929-153959.gz | grep Radiation::run
"a:EAMxx::Radiation::run"                                  -         30       30 1.008000e+04   2.574207e+02     9.787 (    25      0)     6.528 (    11      0)
[brhillman@login4.summit scripts]$ less ne30pg2_ne30pg2.F2010-SCREAMv1.add-cldtot-output.8ca31c665fdd033722005d451b2dbee498a03d87.add-cldtot-output/timing/e3sm_timing_stats.2478972.220929-160829.gz | grep Radiation::run
"a:EAMxx::Radiation::run"                                  -         30       30 1.008000e+04   2.574054e+02     9.820 (    25      0)     6.486 (    11      0)
```
so wallmax for the overall `Radiation::run` timer increases by only 0.3%.

crterai (approved):
Looks good from what I can tell and if unit test passes, I believe it's doing what we want it to.

brhillman:
Here's a quick comparison between cloud area fraction in SCREAMv0, EAMxx with this new diagnostic cloud fraction computed from the mean of vertically-projected sub column cloud (middle) and EAMxx with a crude "all or nothing" cloud fraction in which a column is determined cloudy if any level above has partial cloud fraction greater than 0 (bottom; i.e., what we would get for cloud area if we assumed every column was either clear or completely cloudy, with no partial cloud):
cldtot_test
These are from ne30pg2 simulations, instantaneous snapshots after 1 day.

PR #1971, opened 2022-09-29
Add temperature limiters for gaseous absorption calculations

brhillman:
Add temperature limiters to make sure that the temperatures that RRTMGP gaseous absorption calculations see is within the bounds of the lookup tables. This *should* be BFB unless our test cases are going off the rails already. It turns out though that this is non-BFB for the aquaplanet test because our aquaplanet case reaches temperatures below the minimum temperature defined in the absorption coefficient lookup table data, so these limiters will come into effect for the aquaplanet case.

[non-B4B]

### DIFF
--- a/components/scream/src/physics/rrtmgp/scream_rrtmgp_interface.cpp
+++ b/components/scream/src/physics/rrtmgp/scream_rrtmgp_interface.cpp
@@ -699,12 +699,16 @@ namespace scream {
             OpticalProps2str optics;
             optics.alloc_2str(nday, nlay, k_dist);
 
+            // Limit temperatures for gas optics look-up tables
+            auto t_lay_limited = real2d("t_lay_limited", nday, nlay);
+            limit_to_bounds(t_lay_day, k_dist_sw.get_temp_min(), k_dist_sw.get_temp_max(), t_lay_limited);
+
             // Do gas optics
             real2d toa_flux("toa_flux", nday, ngpt);
             auto p_lay_host = p_lay.createHostCopy();
             bool top_at_1 = p_lay_host(1, 1) < p_lay_host(1, nlay);
 
-            k_dist.gas_optics(nday, nlay, top_at_1, p_lay_day, p_lev_day, t_lay_day, gas_concs_day, optics, toa_flux);
+            k_dist.gas_optics(nday, nlay, top_at_1, p_lay_day, p_lev_day, t_lay_limited, gas_concs_day, optics, toa_flux);
 
 #ifdef SCREAM_RRTMGP_DEBUG
             // Check gas optics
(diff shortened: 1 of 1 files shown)


components/scream/src/physics/rrtmgp/scream_rrtmgp_interface.cpp
```diff
@@ -807,9 +811,14 @@ namespace scream {
...
 
+            // Limit temperatures for gas optics look-up tables
+            auto t_lay_limited = real2d("t_lay_limited", ncol, nlay);
+            auto t_lev_limited = real2d("t_lev_limited", ncol, nlay+1);
+            limit_to_bounds(t_lay, k_dist_lw.get_temp_min(), k_dist_lw.get_temp_max(), t_lay_limited);
+            limit_to_bounds(t_lev, k_dist_lw.get_temp_min(), k_dist_lw.get_temp_max(), t_lev_limited);
 
-            // Do gas optics
-            k_dist.gas_optics(ncol, nlay, top_at_1, p_lay, p_lev, t_lay, t_sfc, gas_concs, optics, lw_sources, real2d(), t_lev);
+            // Do gas opticay+1s
```
ambrad, line 820: Cat crossed the keyboard?

brhillman: Fixed 🐱

brhillman:
The only fail appears to be a DIFF on mappy for the aquaplanet test:
```
DIFF ERS_Ln9.ne4_ne4.F2000-SCREAMv1-AQP1.mappy_gnu9 (phase BASELINE)
```
I think this is expected, as we know that the aquaplanet case tends produce temperatures that dip below the lookup table bounds for RRTMGP, so this PR *should* be adjusting temperatures that are input to those lookup table routines and thus answers should change somewhat. I think this is evidence that this PR is doing what it should be.

brhillman:
Unsurprisingly, if I add a `field_invariant_check` that makes sure temperature is always within RRTMGP's lookup table limits like this:
```
  add_invariant_check<FieldWithinIntervalCheck>(get_field_out("T_mid"),m_grid,rrtmgp::k_dist_sw.get_temp_min(), 500.0,false);
```
the pre-condition check catches the low temperature in the aquaplanet run:
```
Error! Failed pre-condition check (cannot be repaired).
  - Atmosphere process name: Radiation
  - Property check name: T_mid within interval [160, 500]
  - Atmosphere process MPI Rank: 42
  - Message: Check failed.
  - check name: T_mid within interval [160, 500]
  - field id: T_mid[Physics GLL] <double:COL,LEV>(4,72) [K]
  - minimum:
    - value: 150.781
    - entry: (742,3)
    - lat/lon: (-52.2292, 45)
  - maximum:
    - value: 276.714
    - entry: (745,71)
    - lat/lon: (-41.3496, 45)
```
Thus, I think this PR is doing what it is supposed to, and we can ignore the aquaplanet fail and simply bless the DIFF once this is merged.

bartgol (approved):

PR #1974, opened 2022-09-30
Add vertically integrated low, mid, and high-level cloud area

brhillman:
Add vertically integrated low (cldlow), mid (cldmed) and high-level (cldhgh) cloud area. This also adds these fields (along with cldtot) to the default output yaml file so that these diagnostics will be output by default for EAMxx runs. This should be BFB other than the addition of the three extra default output fields.

### DIFF
--- a/components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
+++ b/components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
@@ -133,6 +133,9 @@ void RRTMGPRadiation::set_grids(const std::shared_ptr<const GridsManager> grids_
   // Cloud optical properties added as computed fields for diagnostic purposes
   add_field<Computed>("cld_tau_sw_gpt", scalar3d_swgpts_layout, nondim, grid_name, ps);
   add_field<Computed>("cld_tau_lw_gpt", scalar3d_lwgpts_layout, nondim, grid_name, ps);
+  add_field<Computed>("cldlow"        , scalar2d_layout, nondim, grid_name);
+  add_field<Computed>("cldmed"        , scalar2d_layout, nondim, grid_name);
+  add_field<Computed>("cldhgh"        , scalar2d_layout, nondim, grid_name);
   add_field<Computed>("cldtot"        , scalar2d_layout, nondim, grid_name);
 
   // Translation of variables from EAM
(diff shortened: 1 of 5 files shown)


brhillman:
Some comparisons between these fields in a one-day EAMxx ne30 run with SCREAMv0.

Low-level cloud:
screamv0_vs_eamxx_CLDLOW

Mid-level cloud:
screamv0_vs_eamxx_CLDMED

High-level cloud:
screamv0_vs_eamxx_CLDHGH

components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
```diff
@@ -850,8 +856,20 @@ void RRTMGPRadiation::run_impl (const int dt) {
...
+    // NOTE: limits for low, mid, and high clouds are mostly taken from EAM F90 source, with the
+    // exception that I removed the restriction on low clouds to be above (numerically lower pressures)
+    // 1200 hPa, and on high clouds to be below (numerically high pressures) 50 hPa. This probably
+    // does not matter in practice, as clouds probably should not be produced above 50 hPa and we
+    // should not be encountering surface pressure above 1200 hPa, but in the event that things go off
+    // the rails we might want to look at these still.
+    rrtmgp::compute_cloud_area(ncol, nlay, nlwgpts, 700e2, std::numeric_limits<Real>::max(), p_lay, cld_tau_lw_gpt, cldlow);
+    rrtmgp::compute_cloud_area(ncol, nlay, nlwgpts, 400e2,                            700e2, p_lay, cld_tau_lw_gpt, cldmed);
+    rrtmgp::compute_cloud_area(ncol, nlay, nlwgpts,     0,                            400e2, p_lay, cld_tau_lw_gpt, cldhgh);
+    rrtmgp::compute_cloud_area(ncol, nlay, nlwgpts,     0, std::numeric_limits<Real>::max(), p_lay, cld_tau_lw_gpt, cldtot);
```
bartgol, line 872: Isn't cldtot equal to the sum of the other three? I don't know if this fcn is expensive, but maybe it seems like we have some redundant stuff?

bartgol (approved):
Looks good. I only have one comment, but probably not too important.

components/scream/src/physics/rrtmgp/atmosphere_radiation.cpp
brhillman: Actually, it's not necessarily, that depends on how the subgrid cloud overlaps in the vertical. Imagine this example 3-layer, 3-subcolumn case, where 1s indicate cloud, 0s no cloud, and --- the boundary between low, mid, and high:
```
1 0 0 
-----
1 0 0
-----
1 0 0
```
In this case, each of cldlow, cldmed, and cldhgh would be 1/3, but cldtot would also be 1/3 (project cloud straight down). The only time you'd get cldtot = cldow + cldmed + cldhgh would be the case where each layer had no overlap:
```
1 0 0
-----
0 1 0
-----
0 0 1
```
In which case cldhgh = cldmed = cldow = 1/3, and cldtot = 1.

bartgol: Ok, I don't fully understand the subgrid thingy then. Thanks! =)

brhillman:
Timing results from a 7 day ne30 run with no output, with this branch:
```
"a:EAMxx::Radiation::run"                                  -         30       30 1.008000e+04   2.643063e+02     9.949 (    25      0)     6.662 (    11      0)
```
and before:
```
"a:EAMxx::Radiation::run"                                  -         30       30 1.008000e+04   2.574207e+02     9.787 (    25      0)     6.528 (    11      0)
```
So this is 1.7% increase in the wallmax timer between these two runs.

brhillman:
Merged and blessed on mappy.

PR #1979, opened 2022-10-03
Add a coarsening remapper

bartgol:
This remapper allows to remap fields from a fine grid to a coarse one, using a map file. Some comments:

- The remapper is not constructible from the grids manager, since it requires an extra bit of information (the map file name). This made it impossible to use the grids manager `create_remapper` method. Not a big deal, as I think I want eventually remove that method from the grids manager class(es), and also b/c this remapper is probably only useful for I/O.
- The remapper only works in the "forward" direction. Backward remap is disabled.
- The remapper performs local mat-vec multiplication (on device) with the sparse matrix loaded from the map file.
- The remapper is not yet used anywhere, since I thought it was better to introduce the feature, and then use it. It is however unit tested.
- I introduced another scream cmake option: `SCREAM_MPI_ON_DEVICE` (default ON). This is the same option we have in Homme. If ON, we pass device pointers to the MPI routines, otherwise we manually deep copy buffers to/from host before/after MPI calls.


components/scream/src/share/grid/remap/coarsening_remapper.cpp
```diff
@@ -0,0 +1,951 @@
...
+    case 1:
+    {
+      auto x_view = x.get_view<const Real*>();
+      auto y_view = y.get_view<      Real*>();
+      Kokkos::parallel_for(RangePolicy(0,nnz),
+                           KOKKOS_LAMBDA(const int& i) {
+        const auto row = row_col_lids(i,0);
+        const auto col = row_col_lids(i,1);
+        const auto w = weights(i);
+        y_view(row) += w*x_view(col);
```
ambrad, line 344: On GPU this is a write race condition on `y_view(row)`. For this reason, during init, I recommend formatting the local matrices as CRS. Then you can do the ||4 over rows rather than nonzeros.

components/scream/src/share/grid/remap/coarsening_remapper.cpp
```diff
@@ -0,0 +1,951 @@
...
+  const int nlweights = io_grid_linear->get_num_local_dofs();
+  const auto& raw_comm = m_comm.mpi_comm();
+
+  gid_t offset = nlweights;
+  m_comm.scan(&offset,1,MPI_SUM);
+  offset -= nlweights; // scan is inclusive, but we need exclusive
+
+  // Read chunk of triplets col indices
+  std::vector<gid_t> cols(nlweights);
+  scorpio::get_variable(map_file, "col", "col", {"n_s"}, "int", "int-nnz");
```
ambrad, line 622: So that we can reuse PIO decomps, we should make tag names unique to a decomp or else explicitly deallocate them. Here, "int-nnz" is associated with a different decomp than "int-nnz" later in this routine.

components/scream/src/share/grid/remap/coarsening_remapper.cpp
bartgol: Ah, good point.

components/scream/src/share/grid/remap/coarsening_remapper.cpp
bartgol: Yes, good point. I was indeed getting an issue with decomps getting reused, which I fixed by closing the file every time. But in view of decomps no longer being freed at file closure, I need to make them unique.

bartgol:
Still failing on cuda. Will fix it, and then un-wip it.

bartgol:
I'm getting a cryptic error on CUDA for the new test:
```
coarsening_remapper: memory/memtype_cache.c:90: ucs_memtype_cache_delete: Assertion `pgt_region != ((void *)0)' failed.
```
Google-ing this error only gives me one meaningful hit, related to a [trilinos](https://github.com/trilinos/Trilinos/issues/5036) issue (with an [issue](https://github.com/openucx/ucx/issues/3550) open in ucx repo as well). The solution they propose (setting the env var `UCX_MEMTYPE_CACHE=n`) _does_ indeed remove that error, but they say it is needed if there are cuda mallocs _before_ MPI_Init, which we don't have (MPI_Init is the very first line we execute in our tests). So I'm a bit hesitant introducing an env var without knowing if it's indeed the culprit.

@ambrad @tcclevenger @jgfouca does the error above ring any bell for you?

ambrad:
Might it be related to your MPI windows?

bartgol:
> Might it be related to your MPI windows?

It definitely is. But I'm using them "by the book". Even creating a trivial window of 1 element, in which nothing gets written/read to/from, causes the crash.

I can switch to a sequential round of bcasts, given that it's just a setup cost. Windows were handy since recv procs did not know where the data came. But, as I said, given that it's an init cost, I can prob swallow the pill and do a round of bcast or maybe one/two gathers followed by p2p send/recv.

ambrad:
You might make a standalone reproducer to investigate.

That aside, is the elem GID -> proc map not already available from the old F90 grid setup code? If so, the DOF GID -> proc map can be worked out from that map.

bartgol:
> You might make a standalone reproducer to investigate.
> 
> That aside, is the elem GID -> proc map not already available from the old F90 grid setup code? If so, the DOF GID -> proc map can be worked out from that map.

I'm not sure what setup code you refer to. This MPI communication schedule is unrelated from other maps. In particular, the gid->proc map is on a new grid (the coarse grid used for IO).

I ended up doing two all_gather followed by two rounds of p2p send/recv. That's opposed to 3 MPI_Window communications. I think that's fine for an init-only communication.

ambrad:
That's fine. But if you have a rule of the sort "lowest contributing fine-grid GID's owner owns the coard-grid GID", then I think you can back out everything you need from the fine-grid GID->rank map.

bartgol:
> That's fine. But if you have a rule of the sort "lowest contributing fine-grid GID's owner owns the coard-grid GID", then I think you can back out everything you need from the fine-grid GID->rank map.

I see. I don't think I have any such rule.

components/scream/src/share/grid/remap/coarsening_remapper.cpp
```diff
@@ -0,0 +1,1002 @@
...
+    } else if (can_pack && src_ap.is_compatible<RPack<4>>() &&
+                           ov_tgt_ap.is_compatible<RPack<4>>()) {
+      local_mat_vec<4>(f_src,f_ov_tgt);
+    } else {
+      local_mat_vec<1>(f_src,f_ov_tgt);
+    }
+  }
+
+  // Pack, then fire off the sends
+  pack ();
```
AaronDonahue, line 280: see comment about name of this routine in `.hpp`

AaronDonahue (commented):
Holding on "approving" so that it won't automerge in case you wanted to respond to my comments.  Otherwise, this looks great! Thank @bartgol

components/scream/src/share/tests/coarsening_remapper_tests.cpp
```diff
@@ -0,0 +1,417 @@
...
+  
+  scorpio::eam_pio_enddef(filename);
+
+  // Create triplets: tgt entry K is the avg of src entries K and K+ngdofs_tgt
+  // NOTE: add 1 to row/col indices, since e3sm map files indices are 1-based
+  std::vector<Real> col,row,S;
+  for (int i=0; i<nldofs_tgt; ++i) {
+    row.push_back(1+i+nldofs_tgt*comm.rank());
+    col.push_back(1+i+nldofs_tgt*comm.rank());
+    S.push_back(0.5);
```
AaronDonahue, line 120: Here you are essentially making the weights 0.5 everywhere which doesn't necessarily check what happens when they are not uniform.  We could make this value 0.7 and the next one 0.3, or any combo that adds to 1.0.  Then if there is a mismatch in weight order we might catch it.  You could even have a random real between 0 and 1 and make this one `wgt` and the other `1-wgt`.

components/scream/src/share/grid/abstract_grid.cpp
```diff
@@ -249,6 +249,204 @@ void AbstractGrid::reset_num_vertical_lev (const int num_vertical_lev) {
...
+  }
+
+  dofs_list_type unique_gids_d("",unique_dofs.size());
+  auto unique_gids_h = Kokkos::create_mirror_view(unique_gids_d);
+  std::memcpy(unique_gids_h.data(),unique_dofs.data(),sizeof(gid_type)*unique_dofs.size());
+  Kokkos::deep_copy(unique_gids_d,unique_gids_h);
+  return unique_gids_h;
+}
+
+auto AbstractGrid::
```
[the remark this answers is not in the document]
bartgol: It actually does have an explanation, in the hpp file. I tend to put a comment explaining what the fcn does in the hpp, since that's where the API is.

components/scream/src/share/grid/abstract_grid.cpp
```diff
@@ -249,6 +249,204 @@ void AbstractGrid::reset_num_vertical_lev (const int num_vertical_lev) {
   //       invalidate all geo data whose FieldLayout contains LEV/ILEV
 }
 
+auto
```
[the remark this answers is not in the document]
bartgol: I'll add a brief comment in the header.

components/scream/src/share/grid/remap/coarsening_remapper.hpp
```diff
@@ -0,0 +1,227 @@
...
+
+  std::map<int,std::vector<int>>
+  recv_gids_from_pids (const std::map<int,std::vector<int>>& send_lids_to_pids) const;
+
+#ifdef KOKKOS_ENABLE_CUDA
+public:
+#endif
+  template<int N>
+  void local_mat_vec (const Field& f_src, const Field& f_tgt) const;
+  void pack () const;
```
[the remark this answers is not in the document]
bartgol: I think there would be no confusion, given the context (even MPI has a `MPI_Pack` and `MPI_Unpack` methods). But to meet you half way, I will mirror Homme's methods `pack_and_send` and `recv_and_unpack`, by incorporating send/recv in the pack/unpack methods.

components/scream/src/share/grid/remap/coarsening_remapper.cpp
```diff
@@ -0,0 +1,1002 @@
...
+  view_1d<gid_t>::HostMirror  row_gids_h("",nlweights);
+  view_1d<gid_t>::HostMirror  col_gids_h("",nlweights);
+  view_1d<double>::HostMirror S_h ("",nlweights);
+
+  // scream's gids are of type int, while scorpio wants long int as offsets.
+  std::vector<scorpio::offset_t> dofs_offsets(nlweights);
+  for (int i=0; i<nlweights; ++i) {
+    dofs_offsets[i] = gids_h[i];
+  }
+  const std::string idx_decomp_tag = "int-nnz" + std::to_string(nlweights);
```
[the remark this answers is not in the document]
bartgol: I thought including the nnz value was enough, but in the remote case that the nnz read at the beginning (when each rank has yet to figure out which triplets it needs) and the nnz read here have are the same, it would not be enough. I'll change it.

components/scream/src/share/tests/coarsening_remapper_tests.cpp
bartgol: I picked 0.5 b/c multiplying by powers of two is an exact op in floating point arithmetic (so long as we don't over/underflow, it's just an exponent shift). I suppose multiplying by 0.25 is too, and the counterpart 0.75 should be fine too.

bartgol: But to be precise, I want to avoid weights that do not have an exact FP representation, since I want the checks to be exact, rather than require some tolerance.

components/scream/src/share/field/field.hpp
```diff
@@ -178,11 +178,11 @@ class Field {
 
   // Set the field to a constant value (on host or device)
   template<typename T, HostOrDevice HD = Device>
-  void deep_copy (const T value);
+  void deep_copy (const T value) const;
```
jeff-cohere, line 181: I am a little worried that we're abusing the meaning of `const` here. Strictly speaking, a `const` View can have its data mutated because Views use reference semantics, but this is never going to be obvious to anyone who looks at this thing to try to gain an understanding of what changes what.

I don't think this warrants a change. Just complaining. :-)

components/scream/src/share/grid/remap/coarsening_remapper.cpp
```diff
@@ -0,0 +1,965 @@
...
+      {
+        auto v = f.get_view<Real*>();
+        Kokkos::parallel_for(RangePolicy(0,m_total_num_recv_gids),
+                             KOKKOS_LAMBDA(const int& i){
+          const int lid = lids_pids(i,0);
+          const int pid = lids_pids(i,1);
+          const int lidpos = i - pid_lid_start(pid);
+          const int offset = f_pid_offsets(pid);
+
+          v(lid) += buf (offset + lidpos);
```
ambrad, line 531: In general, multiple remotes will update the owner's `v(lid)`. Thus, isn't this loop (and the subsequent similar ones) a write race on `v(lid)`?

components/scream/src/share/grid/remap/coarsening_remapper.cpp
```diff
@@ -0,0 +1,965 @@
...
+  m_ov_tgt_grid = ov_tgt_grid;
+  const int num_ov_row_gids = m_ov_tgt_grid->get_num_local_dofs();
+
+  // Now we have to create the weights CRS matrix
+  m_row_offsets = view_1d<int>("",num_ov_row_gids+1);
+  m_col_lids    = view_1d<int>("",nlweights);
+  m_weights     = view_1d<Real>("",nlweights);
+
+  // Sort col_gids_h and row_gids_h by row gid. It is easier to sort
+  // the array [0,...,n), and use it later to index the
```
ambrad, line 98: To index the...? Such a cliffhanger!

jeff-cohere: Asynchronous documentation!

components/scream/src/share/grid/remap/coarsening_remapper.cpp
```diff
@@ -0,0 +1,965 @@
...
+  m_tgt_fields.push_back(field_type(tgt));
+}
+
+void CoarseningRemapper::
+do_bind_field (const int ifield, const field_type& src, const field_type& tgt)
+{
+  EKAT_REQUIRE_MSG (
+      src.get_header().get_identifier().get_layout().rank()>1 ||
+      src.get_header().get_alloc_properties().get_padding()==0,
+      "Error! We don't support 2d scalar fields that are padded.\n");
```
jeff-cohere, line 219: We support rank 3 field layouts, too, right? Maybe "2d/3d" instead of "2d" here?

components/scream/src/share/grid/remap/coarsening_remapper.hpp
```diff
@@ -0,0 +1,227 @@
...
+  view_1d<int>          m_send_pid_lids_start;
+  view_1d<int>          m_recv_pid_lids_start;
+
+  // Send/recv requests
+  std::vector<MPI_Request>  m_recv_req;
+  std::vector<MPI_Request>  m_send_req;
+
+  // Unfortunately, MPI's startall/waitall needs a ptr to nonconst requests.
+  // Since do_remap_fwd is a const method, calling m_recv_req.data() would
+  // return a pointer to const. So store the nonconst ptrs.
```
jeff-cohere, line 211: You could also use the `mutable` keyword on the above vectors here if you preferred.

components/scream/src/share/grid/remap/coarsening_remapper.cpp
```diff
@@ -0,0 +1,965 @@
...
+            const int idim = idx / nlevs;
+            const int ilev = idx % nlevs;
+            buf(offset + lidpos*ndims*nlevs + idim*nlevs + ilev) = v(lid,idim,ilev);
+          });
+        });
+      } break;
+
+      default:
+        EKAT_ERROR_MSG ("Unexpected field rank in CoarseningRemapper::pack.\n"
+            "  - MPI rank  : " + std::to_string(m_comm.rank()) + "\n"
```
jeff-cohere, line 475: Two different kinds of ranks!

components/scream/src/share/grid/remap/coarsening_remapper.cpp
jeff-cohere: 👆

components/scream/src/share/grid/remap/coarsening_remapper.cpp
jeff-cohere: Gotcha. Thanks for setting me straight.

components/scream/src/share/grid/remap/coarsening_remapper.cpp
bartgol: I know, I was struggling to make the message clear, without any confusion.

components/scream/src/share/grid/remap/coarsening_remapper.cpp
bartgol: Ah, good catch! It is indeed a race condition. I need to rearrange the unpacking, and use a team ||for.

components/scream/src/share/field/field.hpp
bartgol: Yeah, it would also remove the need for `mutable` for the MPI request vectors.

jeff-cohere:
Looks good to me. I will let others weigh in before approving.

ambrad:
I'm fine with this PR. It's low risk to merge it since the feature defaults to off.

AaronDonahue (dismissed):

PR #1980, opened 2022-10-04
EAMxx/IO: Reinstate free_decomp for now.

ambrad:
I was too ambitious in PR #1969. For now, call free_decomp. I'm not doing a revert because 1969 has a memory-leak fix we want to keep plus some tag uniqueness work that will be needed later and is fine to have now.


ambrad:
It's not clear yet that this is actually the issue, so I'll hold on merging this PR until we're sure.

ambrad:
A reproducer is SMS_P512_Ld1.ne30pg2_ne30pg2.F2010-SCREAMv1.pm-cpu_gnu.

PR #1981, opened 2022-10-05
Switch to new ne30L128 IC with hydrometeors clipped

brhillman:
Switch to new ne30L128 initial condition file with qc, qi, qr, nc, ni, and nr clipped to prevent negative values.


brhillman:
New input file has been uploaded to blues and permissions should be correct, so file should auto-download at case run.

ambrad (approved):
5 days and running on Chrys. The issue was in the first step, so this IC file clearly fixes that. Thanks!

ndkeen (approved):
I was going to test on PM, but it's down. I see Andrew has verified it works on Chrys, so I'm sure it's ok.

brhillman:
@ndkeen I actually tested on perlmutter yesterday before it went down and it worked for me.

Issue #1982, opened 2022-10-06
In scorpio_input, get_io_decomp should record the precision.


ambrad:
PIO's data types include the values PIO_FLOAT 5, PIO_DOUBLE 6, PIO_INT 4. PIO decomps depend on the data type. We are successfully reading IC files with a mix of float and double fields having the same dimensions. Yet https://github.com/E3SM-Project/scream/blob/ae9bf7399ba0e3673c8269bbe2e0bf17ca88059e/components/scream/src/share/io/scorpio_input.cpp#L460 does not distinguish between float and double when making a tag for lookup purposes. It appears that we _should_ make different decomps for each, but right now PIO doesn't care.

bartgol:
Our input class should always uses the scream precision (whatever `Real` is). So the precision _does_ end up in the decomp name.

ambrad:
I might be confused. What happens in the case of an IC file having both float and double fields with the same dimensions?

bartgol:
You mean an IC file where two copies of the same field (meaning same name) are present? That I don't know. Though I wasn't aware that we had such a scenario in our IC files.

ambrad:
No, I mean the following. Suppose the IC file has this:
```
  float phis(time, ncol) ;
  double ps(time, ncol) ;
```
Then the routine I pointed to above will associate a decomp having tag something like "Real-ncol_N" with each. Thus, it might create the decomp for phis and then reuse it for ps, since the tag uses "Real" as the precision. Yet in principle the decomps should be different because the PIO data type is different, float vs double.

bartgol:
Ah, I see. I think you don't need two decomps. From what I understand, the decomp data type needs to match the data type of the array passed in during read/write ops. If the data type of the field in the nc database is different, scorpio will convert on the fly. That is, something similar to `output[idx] << get_val(idx)' happens at read time. Obv, not that, but something that allows, say, values of type `float` to be stuffed into an array of `double`s. The decom data type is used only to correctly `reinterpret_cast` the array passed by the user.

@jayeshkrishna, please, correct me if I'm wrong.

jayeshkrishna:
Yeah, the I/O decompositions represent the data in memory and the data is converted, if needed, to the type of the variable when it is written out.

ambrad:
I see; good to know. Thanks.

PR #1985, opened 2022-10-10
EAMxx: Use simulation data type rather than I/O data type in decomp tag.

ambrad:
Consistent with the resolution to issue #1982, the decomp tag is the simulation data type, not the I/O data type. Thus, the same decomp can be used for both double and float outputs of a field of the same dimension.

[BFB]


AaronDonahue (approved):

ambrad:
This PR tidies up a line of code now that I better understand what PIO needs in its initdecomp routine; it has no effect on what decomps are created in simulations.

PR #1989, opened 2022-10-13
Add a diagnostic to take variables at a specific pressure level

AaronDonahue:
This commit will add a new capability to output a field at a specific pressure level designated in `mb`, `hPa` or `Pa`.

The interface follows the example of the FieldAtLevel diagnostic but uses the vertical interpolation routine to interpolate a field onto a specific pressure level.

In order to request a specific pressure level the syntax is:

```
  - VARIABLE@###mb
  - VARIABLE@###hPa
  - VARIABLE@###Pa
```
where `VARIABLE` is the variable name, `###` is some integer for the pressure level.  The diagnostic will convert `mb` and `hPa` to the appropriate value to satisfy `Pa` which is the unit of pressure in EAMxx.  The output variable name will use the requested units.

### DIFF
--- a/components/scream/src/diagnostics/CMakeLists.txt
+++ b/components/scream/src/diagnostics/CMakeLists.txt
@@ -3,7 +3,7 @@ set(DIAGNOSTIC_SRCS
   dry_static_energy.cpp
   exner.cpp
   field_at_level.cpp
-  field_at_single_pressure.cpp
+  field_at_pressure_level.cpp
   ice_cloud_mask.cpp
   ice_water_path.cpp
   liquid_water_path.cpp
--- a/components/scream/src/diagnostics/CMakeLists.txt
+++ b/components/scream/src/diagnostics/CMakeLists.txt
@@ -26,7 +26,7 @@ set(DIAGNOSTIC_SRCS
 
 add_library(diagnostics ${DIAGNOSTIC_SRCS})
 target_include_directories(diagnostics PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/../share)
-target_link_libraries(diagnostics PUBLIC scream_share physics_share)
+target_link_libraries(diagnostics PUBLIC scream_share)
 
 if (NOT SCREAM_LIB_ONLY)
   add_subdirectory(tests)
--- a/components/scream/src/diagnostics/CMakeLists.txt
+++ b/components/scream/src/diagnostics/CMakeLists.txt
@@ -3,6 +3,7 @@ set(DIAGNOSTIC_SRCS
   dry_static_energy.cpp
   exner.cpp
   field_at_level.cpp
+  field_at_single_pressure.cpp
   ice_cloud_mask.cpp
   ice_water_path.cpp
   liquid_water_path.cpp
--- a/components/scream/src/diagnostics/field_at_pressure_level.cpp
+++ b/components/scream/src/diagnostics/field_at_pressure_level.cpp
@@ -1,4 +1,4 @@
-#include "diagnostics/field_at_single_pressure.hpp"
+#include "diagnostics/field_at_pressure_level.hpp"
 
 #include "ekat/std_meta/ekat_std_utils.hpp"
 #include "ekat/util/ekat_units.hpp"
@@ -7,7 +7,7 @@ namespace scream
 {
 
 // =========================================================================================
-FieldAtSinglePressure::FieldAtSinglePressure (const ekat::Comm& comm, const ekat::ParameterList& params)
+FieldAtPressureLevel::FieldAtPressureLevel (const ekat::Comm& comm, const ekat::ParameterList& params)
   : AtmosphereDiagnostic(comm,params)
   ,m_field_layout(m_params.get<FieldLayout>("Field Layout"))
   ,m_field_units(m_params.get<ekat::units::Units>("Field Units"))
@@ -16,13 +16,13 @@ FieldAtSinglePressure::FieldAtSinglePressure (const ekat::Comm& comm, const ekat
 {
   using namespace ShortFieldTagsNames;
   EKAT_REQUIRE_MSG (ekat::contains(std::vector<FieldTag>{LEV,ILEV},m_field_layout.tags().back()),
-      "Error! FieldAtSinglePressure diagnostic expects a layout ending with 'LEV'/'ILEV' tag.\n"
+      "Error! FieldAtPressureLevel diagnostic expects a layout ending with 'LEV'/'ILEV' tag.\n"
       " - field name  : " + m_field_name + "\n"
       " - field layout: " + to_string(m_field_layout) + "\n");
 }
 
 // =========================================================================================
-void FieldAtSinglePressure::set_grids(const std::shared_ptr<const GridsManager> grids_manager)
+void FieldAtPressureLevel::set_grids(const std::shared_ptr<const GridsManager> grids_manager)
 {
   using namespace ShortFieldTagsNames;
   using namespace ekat::units;
@@ -52,7 +52,7 @@ void FieldAtSinglePressure::set_grids(const std::shared_ptr<const GridsManager> 
   m_diagnostic_output.allocate_view();
 }
 // =========================================================================================
-void FieldAtSinglePressure::compute_diagnostic_impl()
+void FieldAtPressureLevel::compute_diagnostic_impl()
 {
   using namespace scream::vinterp;
 
@@ -61,7 +61,7 @@ void FieldAtSinglePressure::compute_diagnostic_impl()
   const auto pressure = pressure_f.get_view<const Pack**>();
 
   //This is the 1D target pressure
-  view_1d<Pack> p_tgt = view_1d<Pack>("",1);  // We only plan to map onto a single pressure level
+  view_1d<Pack> p_tgt = view_1d<Pack>("",1);  // We only plan to map onto a pressure level
   Kokkos::deep_copy(p_tgt, m_pressure_level);
 
   //input field
--- a/components/scream/src/diagnostics/field_at_pressure_level.cpp
+++ b/components/scream/src/diagnostics/field_at_pressure_level.cpp
@@ -32,52 +32,40 @@ void FieldAtPressureLevel::set_grids(const std::shared_ptr<const GridsManager> g
   m_num_cols = m_grid->get_num_local_dofs();
   auto num_levs = m_grid->get_num_vertical_levels();
 
-  constexpr int ps = Pack::n;
-
   add_field<Required>(m_field_name, m_field_layout, m_field_units, gname);
-  if (m_field_layout.tags().back()==LEV) {
-    m_pres_name = "p_mid";
-    m_num_levs = num_levs;
-    FieldLayout pres_layout { {COL,LEV}, {m_num_cols,m_num_levs} };
-    add_field<Required>(m_pres_name, pres_layout, Pa, gname, ps);
-  } else {
-    m_pres_name = "p_int";
-    m_num_levs = num_levs+1;
-    FieldLayout pres_layout { {COL,ILEV}, {m_num_cols,m_num_levs} };
-    add_field<Required>(m_pres_name, pres_layout, Pa, gname, ps);
-  }
+
+  m_pres_name = m_field_layout.tags().back()==LEV ? "p_mid" : "p_int";
+  add_field<Required>(m_pres_name, m_field_layout, Pa, gname);
+  m_num_levs = m_field_layout.dims().back();
 
   FieldLayout diag_layout { {COL}, {m_num_cols} };
-  FieldIdentifier fid (name(),diag_layout, m, gname);
+  FieldIdentifier fid (name(),diag_layout, m_field_units, gname);
   m_diagnostic_output = Field(fid);
-  auto& C_ap = m_diagnostic_output.get_header().get_alloc_properties();
   m_diagnostic_output.allocate_view();
 }
 // =========================================================================================
 void FieldAtPressureLevel::compute_diagnostic_impl()
 {
   using namespace scream::vinterp;
+  using P = ekat::Pack<Real,1>;
 
   //This is 2D source pressure
   const Field& pressure_f = get_field_in(m_pres_name);
-  const auto pressure = pressure_f.get_view<const Pack**>();
+  const auto pressure = pressure_f.get_view<const P**>();
 
   //This is the 1D target pressure
-  view_1d<Pack> p_tgt = view_1d<Pack>("",1);  // We only plan to map onto a pressure level
+  view_1d<P> p_tgt = view_1d<P>("",1);  // We only plan to map onto a pressure level
   Kokkos::deep_copy(p_tgt, m_pressure_level);
 
   //input field
   const Field& f = get_field_in(m_field_name);
-  const auto f_data_src = f.get_view<const Pack**>();
+  const auto f_data_src = f.get_view<const P**>();
 
   //output field on new grid
   auto d_data_tgt = m_diagnostic_output.get_view<Real*>();
-  view_2d<Pack> data_tgt_tmp("",m_num_cols,1);  // Note, vertical interp wants a 2D view, so we create a temporary one
+  view_2d<P> data_tgt_tmp(reinterpret_cast<P*>(d_data_tgt.data()),m_num_cols,1);  // Note, vertical interp wants a 2D view, so we create a temporary one
 
   perform_vertical_interpolation(pressure,p_tgt,f_data_src,data_tgt_tmp,m_num_levs,1);
-  Kokkos::parallel_for("", m_num_cols, KOKKOS_LAMBDA (const int& icol) {
-    d_data_tgt(icol) = data_tgt_tmp(icol,0)[0];
-  });
 
 }
 
--- a/components/scream/src/diagnostics/field_at_pressure_level.cpp
+++ b/components/scream/src/diagnostics/field_at_pressure_level.cpp
@@ -12,7 +12,7 @@ FieldAtPressureLevel::FieldAtPressureLevel (const ekat::Comm& comm, const ekat::
   ,m_field_layout(m_params.get<FieldLayout>("Field Layout"))
   ,m_field_units(m_params.get<ekat::units::Units>("Field Units"))
   ,m_field_name(m_params.get<std::string>("Field Name"))
-  ,m_pressure_level(m_params.get<int>("Field Target Pressure"))
+  ,m_pressure_level(m_params.get<Real>("Field Target Pressure"))
 {
   using namespace ShortFieldTagsNames;
   EKAT_REQUIRE_MSG (ekat::contains(std::vector<FieldTag>{LEV,ILEV},m_field_layout.tags().back()),
--- a/components/scream/src/diagnostics/field_at_pressure_level.cpp
+++ b/components/scream/src/diagnostics/field_at_pressure_level.cpp
@@ -38,12 +38,14 @@ void FieldAtPressureLevel::set_grids(const std::shared_ptr<const GridsManager> g
   if (m_field_layout.tags().back()==LEV) {
     m_pres_name = "p_mid";
     m_num_levs = num_levs;
+    FieldLayout pres_layout { {COL,LEV}, {m_num_cols,m_num_levs} };
+    add_field<Required>(m_pres_name, pres_layout, Pa, gname, ps);
   } else {
     m_pres_name = "p_int";
     m_num_levs = num_levs+1;
+    FieldLayout pres_layout { {COL,ILEV}, {m_num_cols,m_num_levs} };
+    add_field<Required>(m_pres_name, pres_layout, Pa, gname, ps);
   }
-  FieldLayout pres_layout { {COL,ILEV}, {m_num_cols,m_num_levs} };
-  add_field<Required>(m_pres_name, pres_layout, Pa, gname, ps);
 
   FieldLayout diag_layout { {COL}, {m_num_cols} };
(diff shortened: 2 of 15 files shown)


AaronDonahue:
WIP pending #1937.

components/scream/src/diagnostics/field_at_single_pressure.cpp
```diff
@@ -0,0 +1,85 @@
...
+  const auto& gname  = m_params.get<std::string>("Grid Name");
+  auto m_grid = grids_manager->get_grid(gname);
+  const auto& grid_name = m_grid->name();
+  m_num_cols = m_grid->get_num_local_dofs();
+  auto num_levs = m_grid->get_num_vertical_levels();
+
+  constexpr int ps = Pack::n;
+
+  add_field<Required>(m_field_name, m_field_layout, m_field_units, gname);
+  if (ekat::contains(std::vector<FieldTag>{LEV},m_field_layout.tags().back())) {
```
bartgol, line 40: This is an overkill. You can simply do
```
if (m_field_layout.tags().back()==LEV) {
```

components/scream/src/diagnostics/field_at_single_pressure.cpp
```diff
@@ -0,0 +1,85 @@
...
+  m_diagnostic_output.allocate_view();
+}
+// =========================================================================================
+void FieldAtSinglePressure::compute_diagnostic_impl()
+{
+  using namespace scream::vinterp;
+
+  //This is 2D source pressure
+  Field& pressure = get_field_in(m_pres_name);
+  const auto p_data = pressure.get_view<const Pack**>();
```
bartgol, line 64: Style: in general, the word "data" makes me think to raw pointers. That's what the `data()` method in a `Kokkos::View` or `std::vector` does.

components/scream/src/diagnostics/field_at_single_pressure.cpp
```diff
@@ -0,0 +1,85 @@
...
+  C_ap.request_allocation(ps);
+  m_diagnostic_output.allocate_view();
+}
+// =========================================================================================
+void FieldAtSinglePressure::compute_diagnostic_impl()
+{
+  using namespace scream::vinterp;
+
+  //This is 2D source pressure
+  Field& pressure = get_field_in(m_pres_name);
```
bartgol, line 63: This could be `const Field&`, since it's an input.

bartgol: Side comment: I do not know why I added a non-const getter for input fields. It is puzzling, at best.

components/scream/src/diagnostics/field_at_single_pressure.cpp
```diff
@@ -0,0 +1,85 @@
...
+  const auto p_data = pressure.get_view<const Pack**>();
+
+  //This is the 1D target pressure
+  view_1d<Pack> p_tgt = view_1d<Pack>("",1);  // We only plan to map onto a single pressure level
+  Kokkos::deep_copy(p_tgt, m_pressure_level);
+
+  //input field
+  Field& f = get_field_in(m_field_name);
```
bartgol, line 71: Could also be const.

components/scream/src/diagnostics/field_at_single_pressure.cpp
```diff
@@ -0,0 +1,85 @@
...
+    m_num_levs = num_levs+1;
+  }
+  FieldLayout pres_layout { {COL,ILEV}, {m_num_cols,m_num_levs} };
+  add_field<Required>(m_pres_name, pres_layout, Pa, gname);
+
+  FieldLayout diag_layout { {COL}, {m_num_cols} };
+  FieldIdentifier fid (name(),diag_layout, m, gname);
+  m_diagnostic_output = Field(fid);
+  auto& C_ap = m_diagnostic_output.get_header().get_alloc_properties();
+  C_ap.request_allocation(ps);
```
bartgol, line 54: I don't think you want a pack size here. We never pack along horizontal direction. You're likely better off not using a pack for the diagnostic.

components/scream/src/diagnostics/field_at_single_pressure.cpp
```diff
@@ -0,0 +1,85 @@
...
+  add_field<Required>(m_field_name, m_field_layout, m_field_units, gname);
+  if (ekat::contains(std::vector<FieldTag>{LEV},m_field_layout.tags().back())) {
+    m_pres_name = "p_mid";
+    m_num_levs = num_levs;
+  } else {
+    m_pres_name = "p_int";
+    m_num_levs = num_levs+1;
+  }
+  FieldLayout pres_layout { {COL,ILEV}, {m_num_cols,m_num_levs} };
+  add_field<Required>(m_pres_name, pres_layout, Pa, gname);
```
bartgol, line 48: You are not requesting a pack size for the pressure and for the source field. Yet, during the run_impl, you access them with a `Pack` scalar type. Either request for pack size when you call `add_field`, or use the field with Real as scalar type.

components/scream/src/diagnostics/field_at_single_pressure.hpp
```diff
@@ -0,0 +1,52 @@
...
+
+
+
+  // Keep track of field dimensions
+  std::string         m_field_name;
+  FieldLayout         m_field_layout;
+  ekat::units::Units  m_field_units;
+  std::string         m_pres_name;
+
+  int                 m_pressure_level;
```
bartgol, line 43: This is supposed to be the value of the pressure, and not the level index, right? I understand we will not use decimal pressures, but I think it might be best if you use `Real` here. It doesn't change much, but helps disambiguate from a level index.

components/scream/src/diagnostics/field_at_single_pressure.hpp
```diff
@@ -0,0 +1,52 @@
...
+  int                 m_num_levs;
+  int                 m_num_cols;
+  //scream::vinterp::view_1d<scream::vinterp::Spack> m_pressure_levels;
```
bartgol, line 46: Leftover to be removed.

components/scream/src/diagnostics/register_diagnostics.hpp
```diff
@@ -23,13 +23,16 @@
 #include "diagnostics/zonal_vapor_flux.hpp"
 #include "diagnostics/meridional_vapor_flux.hpp"
 #include "diagnostics/ice_cloud_mask.hpp"
+#include "diagnostics/ice_cloud_mask.hpp"
```
bartgol, line 26: Duplicate include.

components/scream/src/diagnostics/tests/CMakeLists.txt
```diff
@@ -2,11 +2,14 @@
 if (NOT SCREAM_BASELINES_ONLY)
   include(ScreamUtils)
 
-  set( NEED_LIBS scream_share diagnostics physics_share)
+  set( NEED_LIBS scream_share diagnostics physics_share )
```
bartgol, line 5: Why `physics_share`? I am not sure that we need it in all the tests.

components/scream/src/diagnostics/tests/field_at_single_pressure_tests.cpp
```diff
@@ -0,0 +1,166 @@
...
+    for (int icol=0; icol<ncols; icol++){
+      f_mid_v_h(icol,ilev) = icol*100 + ilev;
+      p_mid_v_h(icol,ilev) = 100*ilev;
+      f_int_v_h(icol,ilev) = icol*100 + ilev;
+      p_int_v_h(icol,ilev) = 100*ilev;
+      f_int_v_h(icol,ilev+1) = icol*100 + ilev+1;
+      p_int_v_h(icol,ilev+1) = 100*(ilev+1);
+    }
+  }
+  Kokkos::deep_copy(f_mid_v, f_mid_v_h);
```
bartgol, line 139: The `Field` class conveniently offers `sync_to_host` and `sync_to_dev`. So you can just do
```
auto f_mid_v_h = f_mid.get_view<Real**,Host>();
// fill view
f_mid.sync_to_dev();
```

components/scream/src/diagnostics/tests/field_at_single_pressure_tests.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+  // Test that output at a single pressure level works as expected.
+  // For this test we set a field "M" to be defined as 100*i + k,
+  // where i=column and k=level
+  //
+  // We then set the pressure levels to be 100*k where again k=level.
+  //
+  // Lastly we define a pressure_at_single_level to be for p=150, thus
+  // the output should be (M_1+M_2)/2, or halfway between level 1 and level
+  // 2 of the data.  Given the formula above the output should be exactly
```
bartgol, line 39: I would prefer a bunch of tests covering multiple cases, rather than a single one duplicated over cols/levs. You can keep the pressure levels as you did, but you should test the diagnostic at left/right ends (p=100, p=200), middle (p=150), and also random entry (which might require to use `Catch::Approx` to account for roundings).

components/scream/src/diagnostics/tests/field_at_single_pressure_tests.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+  constexpr int packsize = SCREAM_PACK_SIZE;
+
+  ekat::Comm comm(MPI_COMM_WORLD);
+
+  auto engine = scream::setup_random_test(&comm);
+
+  // Create a grids manager
+  const int ncols = 3;
+  const int nlevs = packsize*2 + 1;  // Note, we need at least 3 levels for the test to work
```
bartgol, line 54: Why does the test require 3+ levels?

components/scream/src/diagnostics/CMakeLists.txt
```diff
@@ -3,6 +3,7 @@ set(DIAGNOSTIC_SRCS
   dry_static_energy.cpp
   exner.cpp
   field_at_level.cpp
+  field_at_single_pressure.cpp
```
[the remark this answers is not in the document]
AaronDonahue: I suppose the idea was that we plan to have a separate diagnostic that will interpolate onto a set of pressure levels.  That being said, we can certainly still get rid of the "single" in the name.

components/scream/src/diagnostics/field_at_single_pressure.cpp
```diff
@@ -0,0 +1,85 @@
...
+namespace scream
+{
+
+// =========================================================================================
+FieldAtSinglePressure::FieldAtSinglePressure (const ekat::Comm& comm, const ekat::ParameterList& params)
+  : AtmosphereDiagnostic(comm,params)
+  , m_field_layout(m_params.get<FieldLayout>("Field Layout"))
+  , m_field_units(m_params.get<ekat::units::Units>("Field Units"))
+{
+  m_field_name  = m_params.get<std::string>("Field Name");
```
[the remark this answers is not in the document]
AaronDonahue:
```
FieldAtLevel::FieldAtLevel (const ekat::Comm& comm, const ekat::ParameterList& params)
  : AtmosphereDiagnostic(comm,params)
  , m_field_layout(m_params.get<FieldLayout>("Field Layout"))
  , m_field_units(m_params.get<ekat::units::Units>("Field Units"))
{ 
  m_field_name  = m_params.get<std::string>("Field Name");
```

Following the example of `FieldAtLevel`.  But I'll make the change for clarity.  Do you want me to make a similar change in FieldAtLevel?

AaronDonahue:
@bartgol , this PR should be ready for review now

components/scream/src/diagnostics/field_at_pressure_level.cpp
```diff
@@ -0,0 +1,84 @@
...
+  m_num_cols = m_grid->get_num_local_dofs();
+  auto num_levs = m_grid->get_num_vertical_levels();
+
+  constexpr int ps = Pack::n;
+
+  add_field<Required>(m_field_name, m_field_layout, m_field_units, gname);
+  if (m_field_layout.tags().back()==LEV) {
+    m_pres_name = "p_mid";
+    m_num_levs = num_levs;
+    FieldLayout pres_layout { {COL,LEV}, {m_num_cols,m_num_levs} };
```
bartgol, line 41: Isn't `pres_layout` simply `m_field_layout`? I think you could just do
```
m_pres_name = m_field_layout.tags().back()==LEV ? "p_mid" : "p_int";
add_field<Required>(m_pres_name,m_field_layout, Pa, gname, ps);
m_num_levs = m_field_layout.dims().back();
```

components/scream/src/diagnostics/field_at_pressure_level.cpp
```diff
@@ -0,0 +1,84 @@
...
+    m_pres_name = "p_int";
+    m_num_levs = num_levs+1;
+    FieldLayout pres_layout { {COL,ILEV}, {m_num_cols,m_num_levs} };
+    add_field<Required>(m_pres_name, pres_layout, Pa, gname, ps);
+  }
+
+  FieldLayout diag_layout { {COL}, {m_num_cols} };
+  FieldIdentifier fid (name(),diag_layout, m, gname);
+  m_diagnostic_output = Field(fid);
+  auto& C_ap = m_diagnostic_output.get_header().get_alloc_properties();
```
bartgol, line 53: Seems like `C_ap` is unused. Can be removed.

components/scream/src/diagnostics/field_at_pressure_level.cpp
```diff
@@ -0,0 +1,84 @@
...
+  const Field& pressure_f = get_field_in(m_pres_name);
+  const auto pressure = pressure_f.get_view<const Pack**>();
+
+  //This is the 1D target pressure
+  view_1d<Pack> p_tgt = view_1d<Pack>("",1);  // We only plan to map onto a pressure level
+  Kokkos::deep_copy(p_tgt, m_pressure_level);
+
+  //input field
+  const Field& f = get_field_in(m_field_name);
+  const auto f_data_src = f.get_view<const Pack**>();
```
bartgol, line 71: If you get rid of packs, you can grab a view with scalar type `const Pack<Real,1>**`. Would be better to simply grab a view with scalar type `const Real**`, but I see that the vert interpolation requires packs.

components/scream/src/diagnostics/field_at_pressure_level.cpp
```diff
@@ -0,0 +1,84 @@
...
+{
+  using namespace ShortFieldTagsNames;
+  using namespace ekat::units;
+  const auto& gname  = m_params.get<std::string>("Grid Name");
+  auto m_grid = grids_manager->get_grid(gname);
+  const auto& grid_name = m_grid->name();
+  m_num_cols = m_grid->get_num_local_dofs();
+  auto num_levs = m_grid->get_num_vertical_levels();
+
+  constexpr int ps = Pack::n;
```
bartgol, line 35: Do you need packs? You're going to get the field at a single level. No point doing packed operations, no?

components/scream/src/diagnostics/field_at_pressure_level.cpp
```diff
@@ -0,0 +1,84 @@
...
+  view_1d<Pack> p_tgt = view_1d<Pack>("",1);  // We only plan to map onto a pressure level
+  Kokkos::deep_copy(p_tgt, m_pressure_level);
+
+  //input field
+  const Field& f = get_field_in(m_field_name);
+  const auto f_data_src = f.get_view<const Pack**>();
+
+  //output field on new grid
+  auto d_data_tgt = m_diagnostic_output.get_view<Real*>();
+  view_2d<Pack> data_tgt_tmp("",m_num_cols,1);  // Note, vertical interp wants a 2D view, so we create a temporary one
```
bartgol, line 75: You could be more efficient then:
```
using P = Pack<Real,1>;
ekat::Unmanaged<view_2d<P>> data_tmp(reinterpret_cast<P*>(d_data_tgt.data()),m_num_cols,1);
```
Then you don't need the parallel for after the interpolation, since `data_tmp` is already aliasing `d_data_tgt`.

bartgol: This requires that you also use P as scalar type for the src/tgt pressure fields.

components/scream/src/diagnostics/tests/field_at_pressure_level_tests.cpp
```diff
@@ -0,0 +1,258 @@
...
+  PressureBnds pressure_bounds;
+  auto engine = scream::setup_random_test(&comm);
+  using RPDF = std::uniform_real_distribution<Real>;
+  RPDF pdf_plev(pressure_bounds.p_top,pressure_bounds.p_surf);
+  
+
+  // Create a grids manager w/ a point grid
+  int ncols = 3;
+  int nlevs = 10;
+  REQUIRE(2*nlevs>=ncols+1);
```
bartgol, line 64: What's the reason behind this requirement?

components/scream/src/diagnostics/tests/field_at_pressure_level_tests.cpp
```diff
@@ -0,0 +1,258 @@
...
+  auto grid = gm->get_grid("Point Grid");
+  auto fm   = get_test_fm(grid);
+  util::TimeStamp t0 ({2022,1,1},{0,0,0});
+
+  // Get a diagnostic factory
+  auto& diag_factory = AtmosphereDiagnosticFactory::instance();
+  diag_factory.register_product("FieldAtPressureLevel",&create_atmosphere_diagnostic<FieldAtPressureLevel>);
+
+  {
+    // Test 1: Take a slice at 500mb (50000Pa) for variable defined at midlevel.
```
bartgol, line 77: It seems to me that test 3 and 4 already cover any use case, so 1 and 2 can be removed?

components/scream/src/diagnostics/tests/field_at_pressure_level_tests.cpp
```diff
@@ -0,0 +1,258 @@
...
+    auto diag_f = diag->get_diagnostic();
+    diag_f.sync_to_host();
+    auto test2_diag_v = diag_f.get_view<const Real*, Host>();
+    for (int icol=0;icol<ncols;icol++) {
+      REQUIRE(test2_diag_v(icol)==get_test_data(plevel));
+    }
+  } 
+  {
+    // Test 3: Take a slice at a random value for variable defined at midpoint.
+    Real plevel = pdf_plev(engine);
```
bartgol, line 104: You might want to put this in a loop, so that you draw plevel from pdf_plev a few times. 10 runs should be good enough, I think.

components/scream/src/diagnostics/field_at_pressure_level.cpp
AaronDonahue: Good point.

components/scream/src/diagnostics/field_at_pressure_level.cpp
AaronDonahue: I'm not sure.  I was under the impression that I needed those for vertical interpolation to work.  But maybe not.

components/scream/src/diagnostics/field_at_pressure_level.cpp
AaronDonahue: using `Unmanaged` didn't compile, but removing that part of the line worked.

components/scream/src/diagnostics/tests/field_at_pressure_level_tests.cpp
AaronDonahue: nope, that was from an older version of the test.

components/scream/src/diagnostics/field_at_pressure_level.cpp
AaronDonahue: This change throws a different error that I need to resolve.  There is a check inside the vertical interpolation that makes sure the extent(1) of the source data matches the expected number of packs.  The test has 10 levels in it, so based on the definition of `P=ekat::Pack<Real,1>` there are 10 packs expected.  But, with a packsize of 16 it looks like `f_data_src` has an `extent(1)` of 16.  The check then throws an error.

components/scream/src/diagnostics/field_at_pressure_level.cpp
```diff
@@ -0,0 +1,72 @@
...
+{
+  using namespace scream::vinterp;
+  using P = ekat::Pack<Real,1>;
+
+  //This is 2D source pressure
+  const Field& pressure_f = get_field_in(m_pres_name);
+  const auto pressure = pressure_f.get_view<const P**>();
+
+  //This is the 1D target pressure
+  view_1d<P> p_tgt = view_1d<P>("",1);  // We only plan to map onto a pressure level
```
bartgol, line 57: This allocates a view at every execution, while we could easily create p_tgt at init time.

AaronDonahue: oh good point. I can create a ninja fix.

AaronDonahue: Done, can you check my change to see if it matches what you were envisioning?

AaronDonahue:
serendipitously this didn't get merged.  My CIME runs uncovered a slight bug where IO didn't like the default mask value output.  I've pushed a change to fix that.

@rebassoo , you may want to take a look too, I added an overload to the vertical interpolation code to accommodate the needed changes.

AaronDonahue:
I'm not sure why this failed to build.  I just built it on blake as a test.

bartgol:
```
error: no matching function for call to approx(const float&, double)
129 |         REQUIRE(approx(test2_diag_v(icol),-99999.0));
```
This is an error of sp build. If you use `-99999` instead, it should probably work. Even better, `Real(-99999)`.

bartgol (approved):

PR #1990, opened 2022-10-14
Fix race condition for compiled f90 modules in shoc

bartgol:
If the same module is compiled twice (usually b/c of different choices of CPP macros), then we should make sure the two modules end up in _different_ folders, to avoid a race condition when make is relocating the compiled files.

In this PR, I make each target use a different folder for their compiled modules. We then make the target add their module folder to the list of their include directories, so that downstream targets can still use those modules.


jgfouca (approved):

jgfouca:
@bartgol , I pushed a fix for the cmake error.

jgfouca (approved):

PR #1993, opened 2022-10-17
Added SCREAMv1-DYAMOND1 compset and related settings

wlin7:
Two compsets are added, F2010-SCREAMv1-DYAMOND1 to go with smooth 16x topography and the associated atm IC, and F2010-SCREAMv1-DYAMOND1-X6T with x6t topography and atm IC. RUN_STARTDATE and land IC automatically set to match DYAMOND-1 starting date of 2016-08-01.

[BFB]


PeterCaldwell:
Thanks Wuyin and Ben!

wlin7:
Note that DYAMOND-1 compsets for ne102pg2 (with 16x or x6t topo) will work out of box, but ICs for generic F2010-SCREAMv1 (for startdate of 01/01, 04/01, 07/01, 10/01) still are not flexible for now. Since the ICs for the runs with the generic F2010-SCREAMv1 are being specified in run scripts, there would be no impact from this PR.

wlin7:
With commit [f92fee3](https://github.com/E3SM-Project/scream/pull/1993/commits/f92fee35a513fc59290870baac362397bb74aacf), the land IC specification is now flexible: proper land IC will be picked up for DYNMOND-1, as well as 01/01, 04/01, 07/01 and 10/01 start dates. The YEAR in RUN_STARTDATE will have no impact.

wlin7:
@bishtgautam , can you please take a look if the change to elm/cime_config/buildnml could have any implication for other tests? For SCREAMv1 simulations at ne1024pg2 grid, we want to have proper lnd IC set for RUN_STARTDATE at the start of each season. To make that work,  I have to disable the possibility of 'ignore_ic_date' being passed to ELM's build-namelist. I don't see impact on tests at other grids. But if it is not the case and the finidat for certain low-res tests could be affected, I could make that change applicable only to SCREAMv1.

wlin7:
@AaronDonahue , can we set to AutoMerge this PR now that all tests PASSED?

AaronDonahue:
for sure we can.  It still needs review.  I can do that.

AaronDonahue:
@PeterCaldwell and/or @crterai , can one of you check this PR.  I'm happy to approve but I am not a Compset expert so I'm not the best candidate to catch errors.

components/scream/cime_config/namelist_defaults_scream.xml
```diff
@@ -263,6 +263,8 @@ tweaking their $case/namelist_scream.xml file.
     <Filename hgrid="ne256np4">${DIN_LOC_ROOT}/atm/scream/init/screami_ne256np4L128_ifs-20200120_20220914.nc</Filename>
     <Filename hgrid="ne512np4">${DIN_LOC_ROOT}/atm/scream/init/screami_ne512np4L128_20220823.nc</Filename> 
     <Filename hgrid="ne1024np4">${DIN_LOC_ROOT}/atm/scream/init/screami_ne1024np4L128_era5-20131001-topoadj-16x_20220914.nc</Filename>
+    <Filename hgrid="ne1024np4" COMPSET=".*SCREAM%DYAMOND-1_ELM">${DIN_LOC_ROOT}/atm/scream/init/screami_ne1024np4L128_ifs-20160801-topoadj16x_20221011.nc</Filename>
+    <Filename hgrid="ne1024np4" COMPSET=".*SCREAM%DYAMOND-1-X6T_ELM">${DIN_LOC_ROOT}/atm/scream/init/screami_ne1024np4L128_ifs-20160801-topoadjx6t_20221011.nc</Filename>
```
PeterCaldwell, line 267: Sanity check question - topography is still included in the screami file, right? If so, then this X6T line is perfect.

components/elm/cime_config/buildnml
```diff
@@ -181,7 +181,10 @@ def buildnml(case, caseroot, compname):
         if elm_nml_use_case != "UNSET": usecase = "-use_case {}".format(elm_nml_use_case)
 
         start_ymd = run_startdate.replace("-", "")
-        ignore = "-ignore_ic_year" if ("-01-01" in run_startdate or "-09-01" in run_startdate) else "-ignore_ic_date"
+
+        # change to always ignore_ic_year to allow more flexibility in specifying IC for different start date of year
+        # ignore = "-ignore_ic_year" if ("-01-01" in run_startdate or "-09-01" in run_startdate) else "-ignore_ic_date"
+        ignore = "-ignore_ic_year"
```
PeterCaldwell, line 187: I think this is fine for now, but eventually we will have a case for the same month/day but a different year and we'll be confused why it isn't working right. Does "101" get expanded to "year=00"? If so, we could ignore_ic_year only when year=00 to avoid future problems...

components/scream/cime_config/namelist_defaults_scream.xml
wlin7: Yes, Peter, topo included in in the screami. The file to be used will be automatically set via compset, F2010-SCREAMv1-DYAMOND1 or F2010-SCREAMv1-DYAMOND1-X6T.

components/elm/cime_config/buildnml
wlin7: True, Peter. If we come to that one day, we need to make it even more flexible, with modifications inside build-namelist that this python script invokes. In the original script, it goes with either 'ignore_ic_year' or 'ignore_ic_date', not suitable for our current purpose. Also don't see this change to allow more flexibility will break any existing tests.

wlin7: BTW, good thought about the ignore option for year=00 only. Changes inside the perl build-namelist would still be needed for make it work. I did test before creating this PR by ignoring none. It did not work out for all our land IC options we plan to run (e.g,  DYAMOND-1, DYAMOND-2, and the four seasons for a year).

PeterCaldwell (approved):

PR #1994, opened 2022-10-17
Add mass and energy checker

tcclevenger:
Adds a property check that checks for mass and energy conservation. Modeled after `check_energy_chng()` in `check_energy.F90`. For any physics atm process that updates any value used in the mass/energy computation, the check compared the computed mass/energy after the process is run to the expected mass/energy (mass/energy _before_ the processes was run plus boundary flux terms) and compares to a user specified tolerance.

The tolerance is defaulted to max Real value (test is effectively off). The current relative errors match what we see in EAM (comparing `SMS_D.ne4pg2_ne4pg2.F2010-SCREAM-LR` (v0) and `SMS_D_Ln9.ne4_ne4.F2010-SCREAMv1-noAero` (v1) CIME cases).

Errors after...
SHOC-mass: ~1e-5, SHOC-energy: ~1e-7
P3-mass: ~1e-5, P3-energy: ~1e-7
Rad-mass: 0 (not updated), Rad-energy: ~1e-16


components/scream/src/control/atmosphere_driver.cpp
```diff
@@ -312,6 +313,8 @@ void AtmosphereDriver::setup_surface_coupling_processes () const
     EKAT_ASSERT_MSG(exporter_found, "Error! SurfaceCoupling exporter data was setup, but no atm process "
                                     "of type AtmosphereProcessType::SurfaceCouplingExporter exists.\n");
   }
+  (void)importer_found;
+  (void)exporter_found;
```
tcclevenger, line 317: Unrelated but was causing warnings

components/scream/src/control/atmosphere_driver.cpp
```diff
@@ -312,6 +313,8 @@ void AtmosphereDriver::setup_surface_coupling_processes () const
     EKAT_ASSERT_MSG(exporter_found, "Error! SurfaceCoupling exporter data was setup, but no atm process "
```
ambrad, line 313: One thing I noticed recently is that EKAT_ASSERT_MSG (disabled ifdef NDEBUG) is being used where EKAT_REQUIRE_MSG (never disabled) is probably intended.

components/scream/src/share/atm_process/atmosphere_process_group.cpp
```diff
@@ -190,6 +190,57 @@ gather_internal_fields  () {
...
+    // If any field used in the mass or energy calculate is updated by this process,
+    // add postcondition check.
+    // Assert that, if a process adds the postcondition check, it also defines
+    // the boundary fluxes.
+    if (mass_or_energy_is_updated) {
+      const bool has_all_boundary_fluxes = atm_proc->has_computed_field("vapor_flux", phys_grid_name) &&
+                                           atm_proc->has_computed_field("water_flux", phys_grid_name) &&
+                                           atm_proc->has_computed_field("ice_flux",   phys_grid_name) &&
+                                           atm_proc->has_computed_field("heat_flux",  phys_grid_name);
+      EKAT_ASSERT_MSG(has_all_boundary_fluxes,
```
ambrad, line 232: Probably meant to be EKAT_REQUIRED_MSG.

components/scream/src/share/property_checks/mass_and_energy_conservation_check.cpp
```diff
@@ -0,0 +1,336 @@
...
+  m_fields["heat_flux"]      = heat_flux_ptr;
+
+  // Assert that all fields needed for mass and energy computation are not null.
+  const bool all_computation_fields_exist =
+      !is_field_null("pseudo_density") && !is_field_null("ps")          &&
+      !is_field_null("phis")           && !is_field_null("horiz_winds") &&
+      !is_field_null("T_mid")          && !is_field_null("qv")          &&
+      !is_field_null("qc")             && !is_field_null("qr")          &&
+      !is_field_null("qi");
+  EKAT_ASSERT_MSG(all_computation_fields_exist,
```
ambrad, line 53: Probably meant to be EKAT_REQUIRE_MSG.

components/scream/src/share/property_checks/mass_and_energy_conservation_check.cpp
```diff
@@ -0,0 +1,336 @@
...
+                  "Error! Currently we require mass and energy conservation "
+                  "check to contain all fields related to the mass and energy "
+                  "computation.\n");
+  (void)all_computation_fields_exist;
+
+  // Require any process that add this checker to define all fluxes.
+  // Fluxes which are not relevant to a certain process should be set to 0.
+  EKAT_ASSERT_MSG(!is_field_null("vapor_flux") && !is_field_null("water_flux") &&
```
ambrad, line 61: Same here.

components/scream/src/share/property_checks/mass_and_energy_conservation_check.cpp
```diff
@@ -0,0 +1,336 @@
...
+                  !is_field_null("ice_flux")   && !is_field_null("heat_flux"),
+                  "Error! If a process adds this check, it must define all " 
+                  "boundary fluxes. Fluxes which are not relevant to a "
+                  "certain process should be set to 0.\n");
+}
+
+bool MassAndEnergyConservationCheck::is_field_null(const std::string fname) const
+{
+  EKAT_ASSERT_MSG(m_fields.find(fname) != m_fields.end(),
```
ambrad, line 70: And here.

components/scream/src/share/property_checks/mass_and_energy_conservation_check.cpp
```diff
@@ -0,0 +1,336 @@
...
+
+PropertyCheck::ResultAndMsg MassAndEnergyConservationCheck::check() const
+{
+  auto mass   = m_current_mass;
+  auto energy = m_current_energy;
+
+  view_1d<Real> rel_err_mass  ("rel_err_mass",   m_num_cols);
+  view_1d<Real> rel_err_energy("rel_err_energy", m_num_cols);
+
+  EKAT_ASSERT_MSG(!std::isnan(m_dt), "Error! Timestep dt must be set in MassAndEnergyConservationCheck "
```
ambrad, line 137: And here.

components/scream/src/control/atmosphere_driver.cpp
```diff
@@ -327,6 +330,48 @@ void AtmosphereDriver::set_precipitation_fields_to_zero ()
   }
 }
 
+void AtmosphereDriver::setup_conservation_checks ()
```
ambrad, line 333: This routine looks like it always runs, regardless of a user option to run or not run the conservation checks. I'd like this routine to exit immediately if checks are not requested.

Separately, I think these conservation checks should have a different input control variable than enable_postcondition_checks, e.g., enable_column_conservation_checks.

components/scream/src/physics/p3/atmosphere_microphysics.hpp
```diff
@@ -243,6 +243,13 @@ class P3Microphysics : public AtmosphereProcess
       } // for ipack
       precip_liq_surf_mass(icol) += precip_liq_surf_flux(icol) * PC::RHO_H2O * m_dt;
       precip_ice_surf_mass(icol) += precip_ice_surf_flux(icol) * PC::RHO_H2O * m_dt;
+
+      // Set appropriate boundary fluxes for energy and mass conservation checks.
+      // Any boundary fluxes not included in P3 interface are set to 0.
```
ambrad, line 248: Might be good to provide a comment here explaining why these fluxes are not accumulated like the ones above.

components/scream/src/share/property_checks/mass_and_energy_conservation_check.cpp
```diff
@@ -0,0 +1,336 @@
...
+    const auto qi_i             = ekat::subview(qi, i);
+    const auto qr_i             = ekat::subview(qr, i);
+
+    // Calculate total mass
+    const Real tm = compute_total_mass_on_column(team, pseudo_density_i, qv_i, qc_i, qi_i, qr_i);
+    const Real previous_tm = mass(i);
+
+    // Calculate expected total mass
+    const Real tm_exp = previous_tm +
+                        compute_mass_boundary_flux_on_column(vapor_flux(i), water_flux(i))*dt;
```
ambrad, line 175: Would you explain how you resolved the issue with dtime vs dtime/numsubcycles being used here?

components/scream/src/physics/p3/atmosphere_microphysics.cpp
```diff
@@ -300,12 +312,14 @@ void P3Microphysics::initialize_impl (const RunType /* run_type */)
...
-      diag_outputs.precip_liq_surf,diag_outputs.precip_ice_surf,
-      precip_liq_surf_mass,precip_ice_surf_mass);
+  p3_postproc.set_variables(m_num_cols,nk_pack,
+                            prog_state.th,pmid,T_atm,t_prev,
+                            prog_state.qv, prog_state.qc, prog_state.nc, prog_state.qr,prog_state.nr,
+                            prog_state.qi, prog_state.qm, prog_state.ni,prog_state.bm,qv_prev,
+                            diag_outputs.diag_eff_radius_qc,diag_outputs.diag_eff_radius_qi,
+                            diag_outputs.precip_liq_surf,diag_outputs.precip_ice_surf,
+                            precip_liq_surf_mass,precip_ice_surf_mass,
+                            vapor_flux, water_flux, ice_flux, heat_flux);
```
bartgol, line 322: If these fields are not always present, you should find a different way to pass them. Perhaps using pointers, which are null if fields are not there?

components/scream/src/physics/p3/atmosphere_microphysics.hpp
```diff
@@ -243,6 +243,13 @@ class P3Microphysics : public AtmosphereProcess
...
+      vapor_flux(icol) = 0.0;
```
bartgol, line 249: Still related to fields not being present when checks are off: this code should be conditional to conservation checks being on.

components/scream/src/physics/shoc/atmosphere_macrophysics.hpp
```diff
@@ -358,6 +358,13 @@ class SHOCMacrophysics : public scream::AtmosphereProcess
             qtracers(i,convert_wet_dry_idx_d(iq),k) = PF::calculate_wetmmr_from_drymmr(qtracers(i,convert_wet_dry_idx_d(iq),k), qv(i,k));
           qv(i,k) = PF::calculate_wetmmr_from_drymmr(qv(i,k), qv(i,k));
       });
+
+      // Set appropriate boundary fluxes for energy and mass conservation checks.
+      // Any boundary fluxes not included in SHOC interface are set to 0.
+      flx_vap(i) = surf_evap(i);
```
bartgol, line 364: Same here: what if checks are off, and these fields are not present.

components/scream/src/physics/p3/atmosphere_microphysics.hpp
```diff
@@ -243,6 +243,13 @@ class P3Microphysics : public AtmosphereProcess
...
+      water_flux(icol) = precip_liq_surf_flux(icol)+precip_ice_surf_flux(icol);
+      ice_flux(icol)   = precip_ice_surf_flux(icol);
```
bartgol, line 251: What's the point of `ice_flux`, if it's set to `precip_ice_surf_flux`? Seems like a pointless field?

components/scream/src/share/atm_process/atmosphere_process_group.cpp
```diff
@@ -190,6 +190,57 @@ gather_internal_fields  () {
...
+    const bool mass_or_energy_is_updated = updates_static_energy || updates_kinetic_energy ||
+                                           updates_water_vapor   || updates_water_liquid ||
+                                           updates_water_ice;
+
+    // If any field used in the mass or energy calculate is updated by this process,
+    // add postcondition check.
+    // Assert that, if a process adds the postcondition check, it also defines
+    // the boundary fluxes.
+    if (mass_or_energy_is_updated) {
+      const bool has_all_boundary_fluxes = atm_proc->has_computed_field("vapor_flux", phys_grid_name) &&
```
bartgol, line 228: Why do you require all physics parametrizations to compute all these fields? I saw some fluxes are set to 0 in SHOC. Is this to avoid having the property check do a bunch of if/else based on whether fields were just updated?

components/scream/src/share/property_checks/mass_and_energy_conservation_check.cpp
```diff
@@ -0,0 +1,336 @@
...
+  m_fields["qc"]             = qc_ptr;
+  m_fields["qr"]             = qr_ptr;
+  m_fields["qi"]             = qi_ptr;
+  m_fields["vapor_flux"]     = vapor_flux_ptr;
+  m_fields["water_flux"]     = water_flux_ptr;
+  m_fields["ice_flux"]       = ice_flux_ptr;
+  m_fields["heat_flux"]      = heat_flux_ptr;
+
+  // Assert that all fields needed for mass and energy computation are not null.
+  const bool all_computation_fields_exist =
```
bartgol, line 47: Might be just me, but given that shared_ptr has the `operator bool ()` overload, having simply
```
const bool all_valid =
	pseudo_density_ptr &&  ps_ptr &&	phis_ptr && horiz_winds_ptr &&
	T_mid_ptr && ...
```
and doing away with `is_field_null` would be more readable.

components/scream/src/physics/p3/atmosphere_microphysics.hpp
tcclevenger: `precip_ice_surf_flux` is in the P3 interface buffer, not in the FM. Also, other processors need to set it to 0 (the checker relies on each process to correctly set all 4 fluxes). I could add `ice_flux` always in P3, and conditionally in SHOC and Radiation.

components/scream/src/share/atm_process/atmosphere_process_group.cpp
tcclevenger: The issue is that if one process includes it, it exists in the checker for all processes. So the only way to be sure it doesn't interfere is to set it to 0. I could zero all of them out before a process runs, and then have the process set the ones it needs.

components/scream/src/physics/p3/atmosphere_microphysics.hpp
tcclevenger: ahh Technically they should be accumulated, I think. And then reset to 0 after P3 is run (like `precip_liq/ice_surf_mass`). Having P3 internal subcycle be 1 hides this issue. I'll update.

tcclevenger: Wait, I'm actually not sure now. In `micro_p3_interface` these values appear to be set (and not accumulated) each time `micro_p3_tend` is run, even during subcycling.

ambrad: Right, it's fine not to accumulate these; I just meant you should provide a comment explaining why, since right above these lines, fluxes are accumulated.

components/scream/src/share/atm_process/atmosphere_process.hpp
```diff
@@ -120,6 +120,9 @@ class AtmosphereProcess : public ekat::enable_shared_from_this<AtmosphereProcess
   // Return the parameter list
   const ekat::ParameterList& get_params () const { return m_params; }
 
+  // Return non-const parameter list
+  ekat::ParameterList& get_params_non_const () { return m_params; }
```
tcclevenger, line 124: I didn't like adding this function, but the first time I ask about `enable_column_conservation_checks` for some processes (those that don't set any boundary fluxes) is in `atmosphere_process_group.cpp` where I need to get the parameters from a different process, and need it to have a default value. Let me know if there is a better way to do this.

tcclevenger (commented):
@ambrad @bartgol Thanks for the reviews. 

I changed things up a bit. Added the new `enable_column_conservation_checks` parameter defaulted to `false` with default `tol=1e-10`. In `add_conservation_checks()` I verify that the checks are not enabled by anything that cannot use them (non-physics process, process group), and I only build the check object if there are processes that will use it.

components/scream/src/physics/p3/atmosphere_microphysics.hpp
tcclevenger: Ok, I'll add a comment.

components/scream/src/share/atm_process/atmosphere_process.cpp
```diff
@@ -828,21 +826,35 @@ ::remove_group (const std::string& group_name, const std::string& grid_name) {
 
 void AtmosphereProcess::setup_conservation_checks_for_this_process (const int dt)
 {
-  const Real tol = m_params.get<Real>("conservation check tolerance",
-                                      std::numeric_limits<Real>::max());
+  // Only setup the check if this process call it.
+  if (not m_params.get<bool>("enable_column_conservation_checks", false)) {
```
ambrad, line 830: Since this is being done at every process call, I recommend calling the PL get function once at init and saving the value to a class boolean that is then used during time stepping.

components/scream/src/share/atm_process/atmosphere_process.cpp
```diff
@@ -818,4 +824,37 @@ ::remove_group (const std::string& group_name, const std::string& grid_name) {
...
+  // Loop through postcondition checks, find the MassAndEnergyConservationCheck,
+  // set dt and tolerance, and compute current mass and energy.
+  bool conservation_check_exists = false;
+  for (const auto& it : m_postcondition_checks) {
+    const auto& pc = it.second;
+    if (pc->name() == "Energy conservation check") {
+      conservation_check_exists = true;
+
+      const auto conservation_check = std::dynamic_pointer_cast<MassAndEnergyConservationCheck>(pc);
+      conservation_check->set_dt(dt);
```
ambrad, line 847: If I understand correctly, you've resolved the dt issue by setting it each time the process is called in the run_impl loop. The dt passed here is the one the run_impl function is given, and that value is dtime divided by the proper number of subcycles. Right?

ambrad (commented):
This is looking really good.

One thing I'd like you to do before merging this PR is to verify that when the checks are off, there is no performance impact. I use this setup when checking the performance implications of my PRs:
* Summit
* 4 nodes, 6 GPUs/node, 6 ranks/node
* ne30pg2_ne30pg2.F2010-SCREAMv1
* 4 day run.

I can send you a script for doing this if you'd like. My approach is to build and run master, then switch to my PR branch and rebuild and run again, and finally compare the model_timing_stats files from the two runs.

components/scream/src/share/atm_process/atmosphere_process_group.cpp
```diff
@@ -190,6 +190,106 @@ gather_internal_fields  () {
...
+      if (atm_proc_group->are_column_conservation_checks_enabled()) {
+        return true;
+      } else {
+        continue;
+      }
+    }
+
+    // If this process is not a group, query enable_column_conservation_checks
+    // and return true if true.
+    if (atm_proc->get_params_non_const().get<bool>("enable_column_conservation_checks", false)) {
```
bartgol, line 212: You could do
```
if (not atm_proc->get_params().isParameter("enable_column_conservation_checks") ||
    not atm_proc->get_params().get<bool>("enable_column_conservation_checks")) {
    return true;
 }
```
C++'s boolean short-circuit will ensure the get method is not called if the param is not stored.

components/scream/src/share/atm_process/atmosphere_process.hpp
bartgol: See comment on the line where you use it.

components/scream/src/share/property_checks/mass_and_energy_conservation_check.cpp
```diff
@@ -170,7 +161,10 @@ PropertyCheck::ResultAndMsg MassAndEnergyConservationCheck::check() const
     const Real tm = compute_total_mass_on_column(team, pseudo_density_i, qv_i, qc_i, qi_i, qr_i);
     const Real previous_tm = mass(i);
 
-    // Calculate expected total mass
+    // Calculate expected total mass. Here, dt should be set to the timestep of the 
+    // subcycle for the process call this check. This effectively scales the boundary
```
bartgol, line 165: "for the process _that called_" this check.

components/scream/src/share/atm_process/atmosphere_process.cpp
```diff
@@ -76,6 +77,11 @@ void AtmosphereProcess::run (const int dt) {
     run_precondition_checks();
   }
 
+  // Column-wise mass and energy checks requires the total mass and energy
+  // to be computed directly before the atm process is run, as well and store
+  // the correct timestep for the process.
+  setup_conservation_checks_for_this_process(dt);
```
bartgol, line 83: I don't see a conservation-check-specific line down below, so I think you can only do the mass/energy checks if `enable_postcondition_checks=true`, right? We're already using a separate PL entry for mass/energy checks, we could probably store the check separate from the other checks, and manually run it based on the value of `enable_column_conservation_checks`. We don't want to have to enable _all_ checks in order to turn on conservation checks...

components/scream/src/share/atm_process/atmosphere_process.cpp
tcclevenger: Correct. So the dt passed to the checker is `atm_dt/num_current_subcycles`

components/scream/src/physics/p3/atmosphere_microphysics.hpp
tcclevenger: I'm actually not sure why this is the case (outside of "it's what eam does"). I'm guessing it has to with the fact that we are only checking the mass/energy change during a single step, so we don't need to accumulate from the previous steps.

components/scream/src/share/atm_process/atmosphere_process.hpp
```diff
@@ -483,6 +498,19 @@ class AtmosphereProcess : public ekat::enable_shared_from_this<AtmosphereProcess
...
+  // Column local mass and energy conservation check
+  std::pair<CheckFailHandling,prop_check_ptr> m_column_conservation_check;
+
+  // Store data related to this processes conservation check.
+  struct ColumnConservationCheckData {
+    // Boolean which dictates whether or not this process
+    // contains the mass and energy conservation checks.
+    bool has_check;
+    // Tolerance used for the conservation check
+    Real tolerance;
```
bartgol, line 510: Do you think it makes sense to use the same tolerance for all conservation checks? Even for the MassEnergy check itself, is it ok to use the same tolerance for mass and energy?

components/scream/src/share/property_checks/mass_and_energy_column_conservation_check.cpp
```diff
@@ -0,0 +1,331 @@
...
+  Kokkos::parallel_reduce("energy_reduce", m_num_cols, KOKKOS_LAMBDA(int i, maxloc_value_t& result) {
+    if (rel_err_energy(i) > result.val) {
+      result.val = rel_err_energy(i);
+      result.loc = i;
+    }
+  }, maxloc_t(maxloc_energy));
+
+  PropertyCheck::ResultAndMsg res_and_msg;
+  if (maxloc_mass.val < m_tol && maxloc_energy.val < m_tol) {
+    res_and_msg.result = CheckResult::Pass;
```
bartgol, line 213: If the check passed, do we really want to spend time creating a message that will never be printed anyways? Imho, we can do an early return here...

components/scream/src/share/property_checks/mass_and_energy_column_conservation_check.cpp
```diff
@@ -0,0 +1,331 @@
...
+                                                                           const Real                  phis) const
+{
+  using PC = scream::physics::Constants<Real>;
+  const Real LatVap = PC::LatVap;
+  const Real LatIce = PC::LatIce;
+  const Real gravit = PC::gravit;
+  const Real Cpair  = PC::Cpair;
+
+  Real total_energy(0);
+  ekat::ExeSpaceUtils<typename KT::ExeSpace>::parallel_reduce(team, 0, m_num_levs,
```
bartgol, line 302: Super minor (and perhaps subjective), but in case you're going to add a commit with other mods, you might as well consider adding a class-level typedef for this type, and then use it in all the fcns in the cpp file to keep lines shorter and more readable (I see `ESU` a short name in other areas of the code).

bartgol (approved):
I only have very minor comments. Feel free to disregard.

components/scream/src/physics/p3/atmosphere_microphysics.hpp
ambrad: Yes, exactly: In both EAM and EAMxx, the macro- and microphysics routines can be subcycled together, so their fluxes have to be accumulated over the physics time step. But the energy and mass checking is done at each call to each physics process, so no accumulation is done. This is what you should document for future readers.

ambrad (commented):
Looks good. I agree with Luca about the message in the case the checks passed. I also responded to your question about accumulating the fluxes vs not. Once you address those two points and do a performance check, I'll approve.

components/scream/src/share/atm_process/atmosphere_process.hpp
ambrad: I'm OK with this if EAM uses just one tol. In principle a process should conserve mass and energy to nearly machine precision, so we ought to be able to set a tolerance of 1e-15 or 1e-14 eventually.

components/scream/src/physics/p3/atmosphere_microphysics.hpp
tcclevenger: What if we have internal p3-only subcycles? In that case the checker is run after multiple p3 calls.
```
In input.yaml:

  p3:
     num_subcycles: n

In atmosphere_process.cpp:run(dt):

  compute_column_conservation_check_data()
  for (num_subcycles) {
     run_impl(dt/num_subcycles)
  }
  run_column_conservation_check()
```
We would need to accumulate in the case, right? Either that or move the conservation stuff inside the subcycle loop:
```
  for (num_subcycles) {
     compute_column_conservation_check_data()
     run_impl(dt/num_subcycles)
     run_column_conservation_check()
  }
```

components/scream/src/share/atm_process/atmosphere_process.hpp
tcclevenger: EAM hard codes its tolerances. 1e-10 for mass, 1e-14 for energy.

components/scream/src/physics/p3/atmosphere_microphysics.hpp
ambrad: > move the conservation stuff inside the subcycle loop

Yes.

But there is perhaps more complexity in the AD than is really necessary. There isn't a need to support process-level subcycling, just grouped-process subcycling. (Process-level subcycling is better handled in the process.) Considering the hierarchy of AtmosphereProcess(Group)::run(_impl) calls, have you thought about removing conservation-check calls from the AtmosphereProcess run routine and instead making it something that a client process would call in their run_impl routine? Doing so would have the advantages of (1) making clear in the code the valid time extent of the mass and energy fluxes and (2) avoiding conservation-related calls in nodes of the AP(G) tree that clearly don't logically support the check. @bartgol, thoughts?

components/scream/src/share/atm_process/atmosphere_process.hpp
tcclevenger: I'll separate mass and energy into their own tolerance, and make them global parameters.

components/scream/src/physics/p3/atmosphere_microphysics.hpp
tcclevenger: Right now I require that a process calling the check logically should call the check. I.e., is a physics process that defines boundary fluxes. So if the user tries to enable the conservation check on a group, or dynamics, or a physics process that doesn't define boundary fluxes, it throws an error. 

I think your suggestion would require a lot of code duplication, since each individual process would need to do the same process of asking "do I have the check? If so, precompute data" and then "do I have the check? If so, run the check".

Another option in a completely different direction would be to allow the user to set `enable_column_conservation_checks" at the physics group level (with a tolerance), and then atmosphere_process_group decides which individual process gets the check. A downside here is that, the a process may want to run the check, but the creator of the process doesn't realize that they needed to define boundary fluxes and is never told.

ambrad: Ok, I'm convinced. The critical driver here is to make sure a conservation check is done for a single process call and nothing else or, alternatively, to accumulate fluxes for the conservation check at the right level of granularity. So your suggestion of moving the calls into the subcycling loop is probably best.

components/scream/src/share/atm_process/atmosphere_process_group.cpp
```diff
@@ -190,6 +190,106 @@ gather_internal_fields  () {
...
+    const bool updates_static_energy  = atm_proc->has_computed_field("T_mid", phys_grid_name);
+    const bool updates_kinetic_energy = atm_proc->has_computed_field("horiz_winds", phys_grid_name);
+    const bool updates_water_vapor    = atm_proc->has_computed_field("qv", phys_grid_name);
+    const bool updates_water_liquid   = atm_proc->has_computed_field("qc", phys_grid_name) ||
+                                        atm_proc->has_computed_field("qr", phys_grid_name);
+    const bool updates_water_ice      = atm_proc->has_computed_field("qi", phys_grid_name);
+    const bool mass_or_energy_is_updated = updates_static_energy || updates_kinetic_energy ||
+                                           updates_water_vapor   || updates_water_liquid ||
+                                           updates_water_ice;
+    EKAT_REQUIRE_MSG(mass_or_energy_is_updated, "Error! enable_column_conservation_checks=true for "
```
tcclevenger, line 271: @ambrad Should I instead throw a warning in the atm process group log here instead of throwing an error? Technically there is nothing wrong with running the check on a process that doesn't update mass or energy, it just wastes time. Radiation for instance doesn't update mass, but still runs the check on mass since it needs the check on energy.

Maybe this is overkill.

components/scream/src/physics/p3/atmosphere_microphysics.hpp
bartgol: It is true that, if a single individual process had to be subcycled, then it can be done inside run_impl. That's precisely what HommeDynamics currently does, with `nsplit` (although homme has a pre-subcycling phase (`homme_pre_process`) and a post-subcycling phase (`homme_post_process`) that would not otherwise be performed had it used subcycling from the base class, so this is a bit different from a pure subcycling). OTOH, keeping it at the AD level seemed like an easy enough thing to do, which costed nothing to do. I am also thinking that, if/when we support parallel splitting for processes, we _might_ need new atm proc classes to wrap a "remote" process. I don't have a design in mind, but keeping things general might prove useful down the road.

bartgol: But I don't understand @tcclevenger point. Why can't we do what andrew suggested? I don't understand `each individual process would need to do the same process of asking "do I have the check? If so, precompute data"`. If a developer for parametrization FOO _knows_ that FOO should conserve mass/energy, they simply call the setup/check method at the top/bottom of run_impl.

tcclevenger: My only point is that, the code in the individual processes interface will all be identical copies of each other (and to what is currently in `atmosphere_process.cpp::run()`).  Whereas right now the developer just adds the input.yaml line `enable_column_conservation_check: true` if they want a certain process checked for conservation, and the rest is handled internally.

We definitely can do what Andrew suggested, but I think it would result in more lines of duplicated code between the processes.

components/scream/src/share/atm_process/atmosphere_process.hpp
tcclevenger: Yeah, I like that idea. I see some inconsistencies with formatting yaml section names: `Time Stepping` and `Scorpio` (spaces, capital letters) vs. `initial_conditions`, `atmosphere_processes`, etc. Should I add tolerances to Debug, and then in a separate PR make the section names consistent?

tcclevenger:
Ran some summit timings with Andrew's suggestions. I ran this CIME case on 4 nodes with conservation checks off:
```
ERS_Ld4.ne30pg2_ne30pg2.F2010-SCREAMv1
```
Timings were similar. I rebased my branch off master to make sure we were comparing only with this branch difference. Not sure why the runtimes were shorter in my branch, but could just be chance?
```
EAMxx::initialize_atm_procs (where all the setup for conservation checks are done)
min time:  6.087 (master), 6.089 (this branch) 
max time:  6.862 (master), 6.922 (this branch)

EAMxx::run
min time: 388.473 (master), 382.159 (this branch) 
max time: 439.625 (master), 432.883 (this branch)
```

@ambrad @bartgol Addressed the final comments. This should be good to go.

ambrad:
Yes, variability. GPU kernels tend to have little variability, but comm, even on on just 4 nodes, can have a lot.

ambrad (approved):

AaronDonahue:
@tcclevenger , if you think this is ready you can add the Automerge label and whatever labels you think would apply.

PR #1995, opened 2022-10-18
Add the capability to produce regional output using a remap file.

AaronDonahue:
This commit will apply the new coarsening remapper to the scorpio output interface to give the user the ability to produce remapped output on the fly.

In order to trigger regional output the output control YAML file needs to have the "remap_file" parameter given.  The value should be a string that points to the remap file to be used.

Additionally all of the fields that will be written as output are provided using "Field Names" as follows:

```
Field Names:
 - var1
 - var2
 - etc
```

NOT:

```
Fields:
  GRID:
    Field Names:
      - var1
      - var2
      - etc.
```

### DIFF
--- a/components/scream/src/share/grid/abstract_grid.cpp
+++ b/components/scream/src/share/grid/abstract_grid.cpp
@@ -366,7 +366,7 @@ get_owners (const hview_1d<const gid_type>& gids) const
   //  - 1.a Figure where each local dof specs will be written in the linearly
   //        distributed global array
   for (int i=0; i<get_num_local_dofs(); ++i) {
-    const auto gid = m_dofs_gids_host[i];
+    const auto gid = m_dofs_gids_host[i]-get_global_min_dof_gid();
     const auto pidlid = pid_and_lid (gid);
     const auto pid = pidlid.first;
     const auto lid = pidlid.second;
--- a/components/scream/cime_config/namelist_defaults_scream.xml
+++ b/components/scream/cime_config/namelist_defaults_scream.xml
@@ -305,7 +305,7 @@ tweaking their $case/namelist_scream.xml file.
 
   <!-- List of yaml files containing I/O output specs -->
   <Scorpio>
-    <output_yaml_files type="array(string)">${SRCROOT}/components/scream/data/scream_default_output.yaml</output_yaml_files>
+    <output_yaml_files type="array(string)">${SRCROOT}/components/scream/data/scream_default_output.yaml,${SRCROOT}/components/scream/data/scream_default_remap.yaml</output_yaml_files>
     <model_restart>
       <Casename>./${CASE}.scream</Casename>
       <output_control>
--- a/components/scream/data/scream_default_remap.yaml
+++ b/components/scream/data/scream_default_remap.yaml
@@ -1,0 +1,71 @@
+%YAML 1.1
+---
+Casename: ${CASE}.screami.remap.hi
+Averaging Type: Instant
+Max Snapshots Per File: 744 # One output every 31 days
+remap_file: ${DIN_LOC_ROOT}/atm/scream/maps/map_ne4np4_to_ne2np4_mono.nc
+Fields:
+  Physics ${PHYSICS_GRID_TYPE}:
+    Field Names:
+      # HOMME
+      - ps
+      - pseudo_density
+      - omega
+      - p_int
+      - p_mid
+      # SHOC + HOMME
+      - horiz_winds
+      # SHOC
+      - cldfrac_liq
+      - eddy_diff_mom
+      - sgs_buoy_flux
+      - tke
+      - pbl_height
+      # CLD
+      - cldfrac_ice
+      - cldfrac_tot
+      # P3
+      - bm
+      - nc
+      - ni
+      - nr
+      - qi
+      - qm
+      - qr
+      - eff_radius_qc
+      - eff_radius_qi
+      - precip_ice_surf_mass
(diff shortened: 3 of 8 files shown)


AaronDonahue:
WIP pending two things:

1. Doesn't work for "horiz_winds", or maybe more generally any field with CMP2 in the layout.
2. Needs robust tests.  I've tested anecdotally using a stand-alone run w/ fabricated regional output and it passes.  But still need an actual unit test.

bartgol:
When you say 'produce regional output', are you interested in just "masking" a bunch of columns, or also changing the resolution? If it's the former, perhaps we can make a better implementation, that does not do mat-vec + MPI...

bartgol:
For sure. I was just thinking "regional" as in "restricted to a lat/lon box", for instance. That could be done faster than the current remapper, since it only involves local copies.

AaronDonahue:
Actually, that is a good point.  That would be another good application and would be cheaper.  I envisioned this also being used to produce global output on a coarse mesh as well.  Like ne120 from ne1024.  In this implementation that would work too.

PeterCaldwell:
Interesting point, Luca. I suspect there will definitely be a common use case where folks are just cookie-cutter-ing out a chunk of the ne1024 domain and another use case where people want global output at coarser resolution. I'm not sure many people will want regional output at coarser resolution. Nonetheless, let's start by producing a single tool that accomplishes both use cases slowly and add the "just subset quickly" task to the v1.1 pile.

AaronDonahue:
Do we know what will happen if the output grid has `ncols` < # of ranks in the job?

PeterCaldwell:
> Do we know what will happen if the output grid has ncols < # of ranks in the job?

Can't we create a test to explore this? Seems like a good unit test to keep forever!

AaronDonahue:
> Are you afraid of some ranks having 0 cols? That should not be a problem. The coarsening remapper test already has this happening (with 2 ranks, the tgt grid happens to be all on rank 0), so the field and remapper stacks are ok with that. And I suspect scorpio will be fine too. Though we should unit test it, of course.

@bartgol I was concerned about the case where the target mesh would be smaller than the number of ranks, thus some ranks will have 0 dof's defined on it for the target mesh.  If I understand the coarsening correctly, we take the target mesh and create a grid over all ranks the same way we would create the grid for the simulation.  But lets say I wanted regional output on 100 columns for a run that used 1000 ranks.  Would the coarsening remapper, or maybe more directly, would the grid constructor be okay w/ 90% of the mpi ranks having 0 degrees-of-freedom?

bartgol:
@AaronDonahue , as I said, this is already happening in the coarsening remapper test, at least for the case with 2 ranks; in that case, the target grid happens to be distributed so that all dofs are on rank 0, with rank 1 owning no dofs at all. The mesh object is _defined_ on all ranks, but some have an empty list of gids.

AaronDonahue:
@bartgol , I still want to add at least one more test.  But for the most part this PR is ready to review, at least on the infrastructure side.  I'm also debating changing up the testing infrastructure for the unit tests, but the essence of the tests will stay the same.  If you have a chance can you take a look at this PR, mainly the fixes to scorpio_output.cpp?  That way, if there are needed changes I can get on them sooner than later.

bartgol (approved):
Just a couple of comments, nothing big.

AaronDonahue:
@bartgol , I made two changes based on your comments.  This PR may need to be re-approved.

bartgol (approved):

AaronDonahue:
This should be ready to go.  I want to finish my full CIME simulations w/ regional output before adding the AUTOMERGE label.

AaronDonahue:
Running this w/ a CIME case remapping from ne4 -> ne2 fails w/ the following error message:
```
32: terminate called after throwing an instance of 'std::logic_error'
32:   what():  /global/u1/a/adonahue/Code/scream/components/scream/src/share/grid/abstract_grid.cpp:334: FAIL:
32: pid>=0 && pid<comm.size()
32: Error! Failed to retrieve owner of GID in the linear grid.
```

AaronDonahue:
I'm making this WIP until I can figure out why it isn't working in CIME cases.

components/scream/src/share/grid/abstract_grid.cpp
```diff
@@ -366,7 +366,7 @@ get_owners (const hview_1d<const gid_type>& gids) const
   //  - 1.a Figure where each local dof specs will be written in the linearly
   //        distributed global array
   for (int i=0; i<get_num_local_dofs(); ++i) {
-    const auto gid = m_dofs_gids_host[i];
+    const auto gid = m_dofs_gids_host[i]-get_global_min_dof_gid();
```
bartgol, line 369: While this works, I do not think it is the right thing to do. This is a GID, not an offset, and this fix might cause trouble down the road if something else in the lines below changes. Instead, I think we should subtract the min GID in the `pid_and_lid` lambda.

bartgol: Side comment, you should cache the value of the min GID. The routine involves global comm, and you don't want to do that multiple times!

AaronDonahue: yes, I agree.  This is in a debug commit at the moment that I added so I could bring the changes back over to blake to do some unit testing.  But I already see this causes some hangs, which I don't fully understand TBH.  Still digging.

AaronDonahue:
Closing due to updated PR in #2051

PR #1998, opened 2022-10-21
Update PM kokkos target and add 2 scream machine files

ndkeen:
Change the kokkos target on Perlmutter from ZEN2 to ZEN3 for CPU.
Add machine file for alvarez and gcp based on pm-cpu

[bfb]


ndkeen:
This was suggested by kokkos developer. The arch of the CPUs on PM is zen3, not zen2, but I originally had trouble trying to use zen3. Now it's working. The actual differences might only be specific, rarely-used optimizations, but just wanted to be more accurate about the target to avoid confusion.

Alvarez is same as pm-cpu.
For GCP, the actual HW used might not be known -- and I found that if I left it blank, it seemed to work fine.

ndkeen:
Actually, for the zen3 arch to work, I need to add `externals/ekat/cmake/machine-files/kokkos/amd-zen3.cmake`.  Does that mean I need a separate PR in that repo?

bartgol:
> Actually, for the zen3 arch to work, I need to add `externals/ekat/cmake/machine-files/kokkos/amd-zen3.cmake`. Does that mean I need a separate PR in that repo?

Yes, if you want to add a new arch file in the ekat submodule, you need a PR in ekat, then update the ekat submodule in scream.

ndkeen:
OK I think last time this happened (to add A100), Jeff did it for me. I think he may be off this week, can someone else help add this file to ekat? Or what's the process?

bartgol:
> OK I think last time this happened (to add A100), Jeff did it for me. I think he may be off this week, can someone else help add this file to ekat? Or what's the process?

I can add the mach file in ekat. I am planning to update ekat in scream (to bring in some changes in some utils) tomorrow. I'll make sure the mach file pr is in first.

AaronDonahue:
Will this PR no longer be needed after #2012 ?

ndkeen:
Still want this, but it won't work until 2012

bartgol:
PR was merged this morning, so this should not be resolved. If not, feel free to reopen.

ambrad:
@bartgol I believe this PR requires EKAT PR 2012 but provides additional capabilities for Alvarez. Thus, it still needs to be merged.

bartgol:
Ah, sorry, I thought this was an issue. My bad.

bartgol:
There's no reason this PR should fail any test. This makes me wonder if we accidentally merged something buggy. I'll check master.

AaronDonahue:
The pressure level PR also failed w/ the same fail in `cprnc` diffs detected.

bartgol:
I'm a bit concerned by the fact that this PR had a fail on blake in the second to last run, but it did not fail on the last run. Unfortunately, I can't reproduce the fail. I suspect this might be partly due to our inability of set the same seed on all ranks, when running with 2+ ranks...

AaronDonahue:
> I'm a bit concerned by the fact that this PR had a fail on blake in the second to last run, but it did not fail on the last run. Unfortunately, I can't reproduce the fail. I suspect this might be partly due to our inability of set the same seed on all ranks, when running with 2+ ranks...

Couldn't we create a helper that only calculates the seed on the master rank and then bcasts it to all ranks?

PR #2003, opened 2022-10-27
Make grids store data inside Field objects rather than bare views

bartgol:
This has a couple of advantages: will allow to remap geometric fields from one grid to another, and also makes dealing with host/dev copies of gids/geo_data easier.

A list of changes:
- Remove dependency of `scream_io` lib on `diagnostics` lib. The IO does not need to know the details about specific diagnostics, it only needs to know what a diagnostic is, which is provided by the `atmosphere_diagnostic.hpp` header. The diags registration in the factory can be done like the registration of atm procs and grids managers, namely before creating an AD instance in tests or mct interface.
- Make grids store gids, lid2idx, and geometric data as Field. The instances of Field returned by getters are const if the grid is const. This means that all Fields need to be populated at grid construction time (or else we need to cast away const).
- B/c of the above point, I moved the loading of lat/lon and hybrid coeffs inside the grids manager, removing it from the AD as well as HommeDynamics.
- Since grids managers now need to use `scream_io`, I moved `mesh_free_grids_manager.*pp` out of `scream_share`, and into its own library `scream_test_support` (which also contains `scream_test_session.cpp`). This lib is built only if `SCREAM_LIB_ONLY=FALSE` (i.e., not a CIME run), and is automatically linked to all our unit tests.

### DIFF
--- a/components/scream/src/dynamics/homme/homme_grids_manager.cpp
+++ b/components/scream/src/dynamics/homme/homme_grids_manager.cpp
@@ -5,6 +5,7 @@
 
 #ifndef NDEBUG
 #include "share/property_checks/field_nan_check.hpp"
+#include "share/property_checks/field_lower_bound_check.hpp"
 #endif
 
 #include "share/io/scorpio_input.hpp"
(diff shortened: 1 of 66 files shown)


tcclevenger (approved):
This looks great. Just some minor comments, feel free to ignore.

components/scream/src/dynamics/homme/homme_grids_manager.cpp
```diff
@@ -204,54 +206,112 @@ build_physics_grid (const ci_string& type, const ci_string& rebalance) {
...
+  auto lat  = phys_grid->create_geometry_data("lat",layout2d,nondim);
+  auto lon  = phys_grid->create_geometry_data("lon",layout2d,nondim);
+  auto area = phys_grid->create_geometry_data("area",layout2d,nondim);
+
+  using gid_t = AbstractGrid::gid_type;
+
+  auto dofs_h = dofs.get_view<gid_t*,Host>();
+  auto lat_h  = lat.get_view<Real*,Host>();
+  auto lon_h  = lon.get_view<Real*,Host>();
+  auto area_h = lon.get_view<Real*,Host>();
```
tcclevenger, line 223:
```suggestion
  auto area_h = area.get_view<Real*,Host>();
```

bartgol:
After discussion with @tcclevenger , we decided to put this on hold, since we're no longer sure this is needed/worth the time. But I'm taking out a few useful commits in a separate PR.

PR #2006, opened 2022-10-28
Refactor how input files are handling in namelist_scream.xml

### LINKED ISSUE #1955: Make it simpler to denote input files in namelist XML
Add another meta attribute, "file", that indicates that the data contained by the element is a file and should be added to the input_list for CIME to download.

jgfouca:
Remove special input_files element and instead add file attribute to parameters that specify input files. This eliminates duplication and allows selectors to filter-out unneeded files.

Fixes #1955

### DIFF
--- a/components/scream/cime_config/buildnml
+++ b/components/scream/cime_config/buildnml
@@ -725,7 +725,7 @@ def get_file_parameters(caseroot):
     for item in raw_xml.findall('.//*[@type="array(file)"]'):
         result.extend(refine_type(item.text, force_type="array(file)"))
 
-    # Remove duplicates. Note sure if an error would be warranted if dupes exist
+    # Remove duplicates. Not sure if an error would be warranted if dupes exist
     return list(OrderedDict.fromkeys(result))
 
 ###############################################################################
@@ -747,12 +747,6 @@ def create_input_data_list_file(case, caseroot):
     with open(input_data_list_file, "w") as fd:
         for idx, file_path in enumerate(files_to_download):
             fd.write("scream_dl_input_{} = {}\n".format(idx, file_path))
-
-            # SCREAM is currently hardcoded to expect many inputs to be in $rundir/data
-            file_name = os.path.basename(file_path)
-            tgt_file = os.path.join(target, file_name)
-            if not os.path.exists(tgt_file) and not os.path.islink(tgt_file):
-                os.symlink(file_path, tgt_file)
 
 ###############################################################################
 def do_cime_vars_on_yaml_output_files(case,caseroot):
--- a/components/scream/cime_config/buildnml
+++ b/components/scream/cime_config/buildnml
@@ -37,7 +37,7 @@ from CIME.buildnml import parse_input
 
 # SCREAM imports
 from eamxx_buildnml_impl import get_valid_selectors, get_child, refine_type, \
-        resolve_all_inheritances, gen_atm_proc_group, find_node, check_all_values
+        resolve_all_inheritances, gen_atm_proc_group, check_all_values
 
 from utils import ensure_yaml
 ensure_yaml()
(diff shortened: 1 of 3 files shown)


jgfouca:
Just noticed @bartgol 's suggestion to use `type` instead of a new attribute. I will do that real quick.

components/scream/cime_config/buildnml
```diff
@@ -675,13 +671,13 @@ def create_input_files(caseroot, screamroot, rundir):
         generated_files = get_child(tree.getroot(),"generated_files")
 
     result = {}
-    for file in generated_files:
-        expect("name" in file.attrib, "file element missing required 'name' attribute")
-        expect("format" in file.attrib, "file element missing required 'format' attribute")
-        filename    = os.path.join(rundir, file.attrib["name"])
-        file_format = file.attrib["format"]
+    for file_ in generated_files:
```
bartgol, line 674: Out of curiosity, why this change?

components/scream/cime_config/buildnml
```diff
@@ -706,40 +702,57 @@ def create_input_files(caseroot, screamroot, rundir):
...
+        raw_xml = tree.getroot()
 
+    result = []
+    for item in raw_xml.findall('.//*[@type="file"]'):
+        result.append(item.text.strip())
+
+    for item in raw_xml.findall('.//*[@type="array(file)"]'):
+        result.extend(refine_type(item.text, force_type="array(file)"))
+
+    # Remove duplicates. Note sure if an error would be warranted if dupes exist
```
bartgol, line 728: Not*

components/scream/cime_config/buildnml
```diff
@@ -706,40 +702,57 @@ def create_input_files(caseroot, screamroot, rundir):
...
-        src_table = do_cime_vars(table.strip(),case)
-        table_name = os.path.basename(src_table)
-        tgt_table = os.path.join(target, table_name)
-        if not os.path.exists(tgt_table) and not os.path.islink(tgt_table):
-            os.symlink(src_table, tgt_table)
+        for idx, file_path in enumerate(files_to_download):
+            fd.write("scream_dl_input_{} = {}\n".format(idx, file_path))
 
+            # SCREAM is currently hardcoded to expect many inputs to be in $rundir/data
+            file_name = os.path.basename(file_path)
```
bartgol, line 752: Is this still necessary? I think we used to put the tables in `run/data` when we had them stored in our repo, but now that we get them from the machine's input data folder, we might not need this anymore. If that's the case, I would remove this altogether (and leave our run/data folder skinnier).

components/scream/cime_config/namelist_defaults_scream.xml
```diff
@@ -74,29 +74,41 @@ tweaking their $case/namelist_scream.xml file.
...
            are matched, then the parameter will be omitted from the case XML
            file ($case/namelist_scream.xml).
+
         4) Parameter types will be inferred. You can override the inferred type
            via the 'type' metadata attribute. Types are:
              logical: will be inferred from any case of true/false
              integer: will be inferred from any sequence of 0-9
              real   : will be inferred from any numeric expression with a decimal point or scientific notation
+             file   : A string that denotes a file path that CIME should download
+                      NOTE: All file parameters will be symlinked to ${RUNDIR}/data from the download area.
```
bartgol, line 93: Related to my comment above: if we no longer need this and remove the links, this comment needs to be removed.

components/scream/cime_config/namelist_defaults_scream.xml
```diff
@@ -74,29 +74,41 @@ tweaking their $case/namelist_scream.xml file.
...
              string : everything else
+
+             array(one of the above): An array of one of the above types
```
bartgol, line 96: We might say that both `[a,b,c]` and `(a,b,c)` syntaxes are supported. At least I think they are...

bartgol (approved):
This looks great. I have a few suggestions, not really change requests though. I'm removing automerge in case you want to act on them.

components/scream/cime_config/buildnml
jgfouca: `file` is a special name in python that usually refers to the class that handles files. The language allows you to override names like this, but it's considered poor practice and pylint flagged it.

bartgol: Ah, I was wondering if that was the case. I tried to open a python file and type `for file in blah` but my vim did not highlight `file`, so I thought that wasn't the case. Good to know. Thanks.

components/scream/cime_config/buildnml
jgfouca: You're right. Just ran a test and this apparently is not needed anymore.

bartgol (approved):

PR #2007, opened 2022-10-28
Add a diagnostic that can vertically remap a field to new pressure levels   

rebassoo:
This commit will add the ability for a field to be output on different pressure levels than the nominal grid. The pressure levels can be defined by the user and can be any number of bins.


AaronDonahue:
It looks like the fails on this have something to do w/ trying to get a clone of the repo and not the actual code changes themselves.  Maybe something wrong w/ the AT starting up again.  I'll add RETEST to see if the problem sorted itself out.

AaronDonahue:
@bartgol , I think this can be reviewed now.  It passed the AT.

bartgol (approved):

PR #2009, opened 2022-10-31
Prevent SHOC temperature from going negative

bogensch:
Temporary fix to prevent individual Gaussian temperatures from going negative.  If so, replace with grid mean temperature to prevent crash.  

PR needs:  examination of cpp conversion (no confidence) and warning messages if clipping does occur.

### DIFF
--- a/components/scream/src/physics/shoc/shoc_assumed_pdf_impl.hpp
+++ b/components/scream/src/physics/shoc/shoc_assumed_pdf_impl.hpp
@@ -222,17 +222,18 @@ void Functions<S,D>::shoc_assumed_pdf(
       }
 
       // Begin to compute cloud property statistics
+      const Spack Tl1_g = thl_first/(ekat::pow(basepres/pval,(rair/cp)));
       const Spack Tl1_1 = thl1_1/(ekat::pow(basepres/pval,(rair/cp)));
       const Spack Tl1_2 = thl1_2/(ekat::pow(basepres/pval,(rair/cp)));
 
       const auto index_range = ekat::range<IntSmallPack>(k*Spack::n);
       const Smask active_entries = (index_range < nlev);
 
-      // Check to ensure Tl1_1 and Tl1_2 are not negative, endrun otherwise
+      // Check to ensure Tl1_1 and Tl1_2 are not negative, temporarily set to grid mean value if so.
+      Tl1_1.set(Tl1_1<0, Tl1_g);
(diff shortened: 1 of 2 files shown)


components/scream/src/physics/shoc/shoc_assumed_pdf_impl.hpp
```diff
@@ -222,17 +222,36 @@ void Functions<S,D>::shoc_assumed_pdf(
...
 
-      // Check to ensure Tl1_1 and Tl1_2 are not negative, endrun otherwise
+      // Check to ensure Tl1_1 and Tl1_2 are not negative, temporarily set to grid mean value if so.
       const auto is_neg_Tl1_1 = (Tl1_1 <= 0) && active_entries;
       const auto is_neg_Tl1_2 = (Tl1_2 <= 0) && active_entries;
-      EKAT_KERNEL_REQUIRE_MSG(!(is_neg_Tl1_1.any()), "Error! Tl1_1 has <= 0 values.\n"); //exit with an error message
-      EKAT_KERNEL_REQUIRE_MSG(!(is_neg_Tl1_2.any()), "Error! Tl1_2 has <= 0 values.\n"); //exit with an error message
+      Tl1_1.set(is_neg_Tl1_1, Tl1_g);
+      Tl1_2.set(is_neg_Tl1_2, Tl1_g);
+      if( is_neg_Tl1_1.any() ) {
```
ambrad, line 237: This is a bit of overkill since the pack size is 1 on the GPU platforms. But doesn't matter either way.

ambrad (approved):

components/scream/src/physics/shoc/shoc_assumed_pdf_impl.hpp
```diff
@@ -230,10 +230,28 @@ void Functions<S,D>::shoc_assumed_pdf(
       const Smask active_entries = (index_range < nlev);
 
       // Check to ensure Tl1_1 and Tl1_2 are not negative, temporarily set to grid mean value if so.
-      Tl1_1.set(Tl1_1<0, Tl1_g);
-      Tl1_2.set(Tl1_2<0, Tl1_g);
       const auto is_neg_Tl1_1 = (Tl1_1 <= 0) && active_entries;
       const auto is_neg_Tl1_2 = (Tl1_2 <= 0) && active_entries;
+      Tl1_1.set(is_neg_Tl1_1, Tl1_g);
```
bartgol, line 235: You might as well move these `set` calls inside the if statements.

Issue #2004, opened 2022-10-28
Iniital runs with DYAMOND1 compset at ne1024


ndkeen:
An issue to help us track problems testing DYAMOND1 after https://github.com/E3SM-Project/scream/pull/1993

On pm-cpu, I tried 

```
set compset=F2010-SCREAMv1-DYAMOND1 
set mach=pm-cpu
set res=ne1024pg2_ne1024pg2
./xmlchange SCREAM_CMAKE_OPTIONS="SCREAM_NP 4 SCREAM_NUM_VERTICAL_LEV 128 SCREAM_NUM_TRACERS 10"
```

and see:

`ERROR: GETFIL: FAILED to get /global/cfs/cdirs/e3sm/inputdata/lnd/clm2/initdata/20220928.I2010CRUELM.ne1024pg2_ICOS10.elm.r.2016-08-01-00000.nc`

And when I instead try:
`set compset=F2010-SCREAMv1-DYAMOND1-X6T`

I see:

```
Model scream missing file ic_file_0 = '/global/cfs/cdirs/e3sm/inputdata/atm/scream/init/screami_ne1024np4L128_ifs-20160801-topoadjx6t_20221011.nc'
Trying to download file: 'atm/scream/init/screami_ne1024np4L128_ifs-20160801-topoadjx6t_20221011.nc' to path '/global/cfs/cdirs/e3sm/inputdata/atm/scream/init/screami_ne1024np4L128_ifs-20160801-topoadjx6t_20221011.nc' using SVN protocol.
svn export failed with output:  and errput svn: E170000: URL 'https://svn-ccsm-inputdata.cgd.ucar.edu/trunk/inputdata/atm/scream/init/screami_ne1024np4L128_ifs-20160801-topoadjx6t_20221011.nc' doesn't exist
```

Note that same setup runs with `set compset=F2010-SCREAMv1`

wlin7:
Noel, at this point, we should be using compset F2010-SCREAMv1-DYAMOND1, without '-X6T'. But nevertheless, we should have the x6t screami in place. F2010-SCREAMv1-DYAMOND1 uses the following screami on nersc.
/global/cfs/cdirs/e3sm/inputdata/atm/scream/init/screami_ne1024np4L128_ifs-20160801-topoadj16x_20221011.nc

The error on getting elm.r is confusing, as the file exists at the exact path.

ndkeen:
As Wuyin notes, the file that the error message points to is there:
```
[...]

PR #2011, opened 2022-11-01
Set the year values for DYAMOND-1 SSTICE data stream

wlin7:
This is to match both the year values in the SSTICE data and the model year to use for DYAMOND-1 simulation, which is 2016.

Fix issue #2004.

[BFB]


ndkeen (approved):
I ran 36 steps with this branch on pm-cpu and pm-gpu, however, I mistakenly left the xmlchanges in my script. So I'm submitting again without those to double-check no issues.
I will go ahead and approve.

ndkeen:
The 36-step case on pm-cpu completed (without any xmlachnges)

PR #2016, opened 2022-11-02
Avoid very expensive $phys_functions_f90.cpp compile if SCREAM_LIB_ONLY

### LINKED ISSUE #547: Many fewer things are needed when SCREAM_LIB_ONLY=True
The cxx->f90 bindings and probably most or all of the *_f90.[ch]pp files.

jgfouca:
Fixes #547


bartgol (approved):
Nice

PR #2017, opened 2022-11-04
SHOC: Fix use before assign of kbfs.

ambrad:


ambrad:
Compare with the F90: https://github.com/E3SM-Project/scream/blob/6141dea55154bbf6f4b1a9c28bccdfae22530eba/components/eam/src/physics/cam/shoc.F90#L4148

AaronDonahue (approved):

tcclevenger (approved):

jgfouca:
@ambrad , nice catch! I'm wondering how that cxx implementation ever passed our bfb unit tests?

jgfouca:
Ah, I bet the ranges for the randomized inputs were such that the negative value for kbfs was never hit.

PR #2018, opened 2022-11-07
Cleanup shoc and p3 dirs

### LINKED ISSUE #1966: Reorg physics directories
There are loads of files, most falling into one of three categories:
1) .cpp ETI files
2) _impl.hpp files
3) _disp.cpp small-kernel dispatch files

We should split these into different subdirectories to avoid having an overwhelming number of files in a single directory.

jgfouca:
Fixes #1966


bartgol (approved):
Nice! Thanks for the cleanup! It looks so much easier to navigate now.

It's curious that git did not pick up that a couple of files in both p3 and shoc were moved, while it did pick that up for all other cpp files.

jgfouca:
@bartgol , some of them were probably too-modified for git to detect the move. I cleaned up whitespace, unnecessary includes etc in some files.

ambrad:
I can symlink everything back to where it was, but I just want to record that devs have different opinions on deep vs. shallow dir structs, and I'm one who likes shallow ones. I like being able to e.g. `C-x-f shoc_pblintd*pp` to open everything related to that set of routines. It's never clear to me what deep hierarchies offer in trade.

ambrad:
I'm removing the automerge label because we want to be sure to get some small PRs in for the next production run before this, to avoid having to do conflict resolution on those.

jgfouca:
@ambrad , @bartgol , @tcclevenger , @AaronDonahue, @PeterCaldwell : we should discuss if this is worth it. I also usually prefer shallower directory trees. Looking at shoc, we have 110 source files in the same directory without this PR.

Arguments (IMO) for this reorg:
1) 100+ files in the same directory is too many
2) The ETI files are uninteresting and unlikely to change, so developers will rarely want to interact with these files. Having them adjacent to the "real" source files clutters things
3) If a developer is looking at shoc to see what API shoc offers, they will be overwhelmed by dozens of *impl.hpp files which aren't intended to included by outside packages.

Arguments against:
1) A flatter structure reduces navigation typing when developing/debugging as Andrew explained
2) Most of the benefits of the reorg can be achieved with file globbing since we are using consistent file names
3) Intra-shoc includes are a bit less convenient (`#include file` becomes `#include ../file` or `#include impl/file`)

Thoughts?

ambrad:
What about the option of keeping the `eti` directory from the PR but removing the `impl` and `disp` directories?

jgfouca:
@ambrad , yes, I think we can probably all agree on that one.

ambrad:
If someone were to make SHOC a standalone library, they would use CMake to create the installation directory structure best suited to library use. The source dir structure that devs work on should favor development ease, since it doesn't have to be the same as the installation structure.

bartgol:
I am very much in favor of the reorg.

Imho, long term, the benefit of code organization is very much worth the extra keystrokes in the folder navigation and/or header include.

tcclevenger:
I also like the reorg. It becomes much easier to decipher what is what for anyone coming in new in the future if there is a little more organization, whereas now it just looks like a mess when you `ls`. 

Also, I would vote against relative paths in includes, but don't feel too strongly if no one else has an issue with them.

ambrad:
I acquiesce. But I agree with Conrad that we should _not_ use relative paths. We use filename suffixes to indicate impl, so that distinction is already clear in the include line.

jgfouca:
@ambrad , @tcclevenger , thanks! Let me confirm that I understand what you want wrt includes. My first thought was to add a private include path to the shoc library target, so I could do 
`#include "p3_functions.hpp"`
instead of
`#include "../p3_functions.hpp"`

That would take care of the includes in `impl` and `disp` which leaves the `impl/*.hpp` includes in the main shoc_functions.hpp file. I could add an additional private include to the shoc library to `impl`.

jgfouca:
@ambrad , is this safe to merge now?

jgfouca:
I am cleaning up the relative includes now.

jgfouca:
Since this passed and Luca had approved before, I'm going to merge to avoid conflicts.

PR #2021, opened 2022-11-09
EAMxx: process rrtmgp sources with YAKL's macros to select compile language

### LINKED ISSUE #2020: RRTMGP: Summit SCREAMv1 build shows -arch=sm_35 for RRTMGP files
Look for
```
nvcc warning : The 'compute_35', 'compute_37', 'compute_50', 'sm_35', 'sm_37' and 'sm_50' architectures are deprecated, and may be removed in a future release (Use -Wno-deprecated-gpu-targets to suppress warning).
```
in e3sm.bldlog. When I print the nvcc command in nvcc_wrapper, I see in more detail that it's only the RRTMGP files that have sm_35 rather than sm_70.

bartgol:
In particular, this sets the compile language of rrtmgp sources to match whatever compile language was set for YAKL's c++ sources. This ensures that the flags of target rrtmgp inherit the same flags used to compile the yakl target.

Fix #2020

### DIFF
--- a/components/scream/src/physics/rrtmgp/CMakeLists.txt
+++ b/components/scream/src/physics/rrtmgp/CMakeLists.txt
@@ -57,6 +57,12 @@ set(EXTERNAL_SRC
   ${EAM_RRTMGP_DIR}/external/cpp/rte/kernels/mo_rte_solver_kernels.cpp
   ${EAM_RRTMGP_DIR}/external/cpp/extensions/fluxes_byband/mo_fluxes_byband_kernels.cpp
 )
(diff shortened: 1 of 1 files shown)


bartgol:
Removed automerge, in case Ben has any comment. @brhillman , if you approve, please, re-add automerge (or merge directly if AT already ran)

ambrad:
Let's automerge then. @brhillman, feel free to suggest something else in another PR. This one solves the key problem, so it would be good to get it in for the next runs on Summit.

PR #2022, opened 2022-11-10
EAMxx: minor mod to random test setup, to enhance reproducibility

bartgol:
When trying to re-run a test with 2+ ranks, we can only pass the random seed used by rank 0 (which was printed by the test setup), but that will make all ranks use what rank 0 used in the first run, which is not what they originally used.

This PR makes the rng seed for ranks>0 dependent on the rng seed of rank 0 _always_. They will still use different seeds, to increase randomization, but this way we can fully reproduce an MPI run.

Note: in order to take advantage of this feature, we should always pass a comm to `setup_random_test`. @jgfouca I wonder if we should make the comm arg mandatory...


jgfouca (approved):
Looks good. I think the comm is fine as-is since we want to support tests that don't use MPI.

bartgol:
Actually, `./horizontal_remap --rng-seed 747842163` fails also with one rank.

Issue #2015, opened 2022-11-02
3 Diagnostic fields appear to now work on weaver when output


AaronDonahue:
This issue was discovered when developing PR #1989 .  This PR also added an output stream to the `homme_shoc_cld_spa_p3_rrtmgp` unit test to output just all of the available diagnostics.  This test failed on `weaver` for the three diagnostics:
      - VerticalLayerInterface
      - VerticalLayerMidpoint
      - DryStaticEnergy

We should investigate this error and fix it.  Note, there were changes to the output stream in this PR so it's possible the PR itself broke these diagnostics, although it is not clear why these three would be uniquely effected.

AaronDonahue:
I've confirmed we get the same behavior w/ master.

AaronDonahue:
Here is a sample of the error message:
```
KERNEL CHECK FAILED:
   !ekat::OnGpu<typename device_%s:%u: %s: block:pace>::value || is_pow_of_2(team.team_size())
   Error! Team-level parallel_scan on CUDA only works for team sizes that are power of 2.
       You could try to reduce the team size to the previous pow of 2.
```

AaronDonahue:
I've made some progress on tracking down the issue here.  @bartgol , @tcclevenger , @ambrad or @jgfouca  you may be able to offer some further insight because I think it has to do w/ team policy.  The three diagnostics that fail on `weaver` w/ the above error all use `calculate_dz` from the physics functions: https://github.com/E3SM-Project/scream/blob/master/components/scream/src/share/util/scream_common_physics_functions_impl.hpp#L299

There are two other diagnostics that use the same function and they do work, `atm_density` and `vertical_layer_thickness`.  Looking at the usage in both working and non-working cases there is a pattern.

WORKING: https://github.com/E3SM-Project/scream/blob/master/components/scream/src/diagnostics/atm_density.cpp#L55
```
  Kokkos::parallel_for("VerticalLayerThicknessDiagnostic",
                       Kokkos::RangePolicy<>(0,m_num_cols*npacks),
                       KOKKOS_LAMBDA(const int& idx) {
      const int icol  = idx / npacks;
      const int jpack = idx % npacks;
      dz(icol,jpack) = PF::calculate_dz(pseudo_density_mid(icol,jpack), p_mid(icol,jpack), T_mid(icol,jpack), qv_mid(icol,jpack));
  });
```

[...]

PR #2023, opened 2022-11-10
This commit fixes a bug in the team policy declaration for diagnostics.

AaronDonahue:
There are three diagnostics that failed when run in a typical AD case on a GPU machine.  These three diagnostics;
 - DryStaticEnergy
 - VerticalLayerInterface
 - VerticalLayerMidpoint all shared one thing in common, under-the-hood they used a Kokkos parallel scan.  The way the Kokkos policy was set up didn't guarantee a team size of a power of 2 which caused fails in certain cases.  This commit solves that by using the appropriate team policy setup.

This commit also solves a separate bug which was discovered where temporary variables were not being handled properly in the diagnostics which could lead to non-BFB behavior.

Solves #2015 

[BFB] - in general, does solve a potential race condition.

### DIFF
--- a/components/scream/src/diagnostics/dry_static_energy.cpp
+++ b/components/scream/src/diagnostics/dry_static_energy.cpp
@@ -43,6 +43,13 @@ void DryStaticEnergyDiagnostic::set_grids(const std::shared_ptr<const GridsManag
   auto& C_ap = m_diagnostic_output.get_header().get_alloc_properties();
   C_ap.request_allocation(ps);
   m_diagnostic_output.allocate_view();
+
+  // Initialize a 2d view of dz to be used in compute_diagnostic
+  const auto npacks     = ekat::npack<Pack>(m_num_levs);
+  const auto npacks_p1  = ekat::npack<Pack>(m_num_levs+1);
+  m_dz    = view_2d("",m_num_cols,npacks);
+  m_z_int = view_2d("",m_num_cols,npacks_p1);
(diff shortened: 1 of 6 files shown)


components/scream/src/diagnostics/dry_static_energy.cpp
```diff
@@ -63,27 +70,30 @@ void DryStaticEnergyDiagnostic::compute_diagnostic_impl()
   const Real surf_geopotential = 0.0;
 
   const int num_levs = m_num_levs;
-  view_1d dz("",npacks);
-  view_1d z_int("",npacks_p1);
-  view_1d z_mid("",npacks);
+  auto      dz       = m_dz;
```
bartgol, line 73: If you really wanted, you could store dz in `m_z_mid`. By the time you compute z_mid, you no longer need dz...

components/scream/src/diagnostics/vertical_layer_midpoint.cpp
```diff
@@ -57,19 +63,21 @@ void VerticalLayerMidpointDiagnostic::compute_diagnostic_impl()
   const Real surf_geopotential = 0.0;
 
   const int num_levs = m_num_levs;
-  view_1d dz("",npacks);
-  view_1d z_int("",npacks_p1);
+  auto dz    = m_dz;
```
bartgol, line 66: Same here. You can store dz in z_mid, and avoid this temporary.,

bartgol (approved):

PR #2024, opened 2022-11-10
EAMxx: link kokkos to rrtmgp, to get correct compiler and flags on CUDA

bartgol:
This seems to fix the rrtmgp tests, while not reintroducing the warning that #2021 addressed.

In short:
- do not use YAKL macro to mark C++ source files with lang=CUDA.
- link kokkos to rrtmgp, simply to get nvcc_wrapper and CUDA flags.

This ensures we build rrtmgp in the same way we build the rest of scream (nvcc_wrapper and correct flags).

### DIFF
--- a/components/scream/src/physics/rrtmgp/CMakeLists.txt
+++ b/components/scream/src/physics/rrtmgp/CMakeLists.txt
@@ -57,24 +57,22 @@ set(EXTERNAL_SRC
   ${EAM_RRTMGP_DIR}/external/cpp/rte/kernels/mo_rte_solver_kernels.cpp
   ${EAM_RRTMGP_DIR}/external/cpp/extensions/fluxes_byband/mo_fluxes_byband_kernels.cpp
 )
+
 add_library(rrtmgp ${EXTERNAL_SRC})
 EkatDisableAllWarning(rrtmgp)
 SetCudaFlags(rrtmgp)
-target_link_libraries(rrtmgp PUBLIC yakl)
+
+# NOTE: rrtmgp does not use kokkos, but by linking to it, we a) have rrtmgp be built
+#       using nvcc_wrapper, and b) all the (public) CUDA flags for kokkos are fwd-ed
+#       to rrtmgp (e.g., the arch flag).
+target_link_libraries(rrtmgp PUBLIC yakl kokkos)
 target_include_directories(rrtmgp PUBLIC
-    ${SCREAM_BASE_DIR}/../../externals/YAKL
     ${EAM_RRTMGP_DIR}/external/cpp
     ${EAM_RRTMGP_DIR}/external/cpp/rte
     ${EAM_RRTMGP_DIR}/external/cpp/rte/kernels
     ${EAM_RRTMGP_DIR}/external/cpp/rrtmgp
     ${EAM_RRTMGP_DIR}/external/cpp/rrtmgp/kernels
 )
-
-# The lines below are needed to ensure that kokkos_launch_compiler injects
-# nvcc into compilations. rrtmgp uses YAKL, not kokkos, so the wrapper
-# didn't know to add nvcc without these lines.
-target_compile_definitions(rrtmgp PRIVATE KOKKOS_DEPENDENCE)
-target_link_options(rrtmgp PRIVATE -DKOKKOS_DEPENDENCE)
 
 # Build RRTMGP interface; note that we separate the SCREAM-specific RRTMGP interface
 # from the external core RRTMGP library because, ideally, the RRTMGP library has its


brhillman:
I'm going to ping @mrnorman on this; Matt, do you see any danger in not using the yakl macro to process the RRTMGPXX source files?

mrnorman:
It's hard to say, so do this at your own risk. Without knowing what flags you're setting and how, it's hard to say. You'll need to make sure you set the same flags that YAKL would've set.

mrnorman:
You could end up missing some of the needed CPP defs such as `YAKL_DEBUG`, `YAKL_B4B`, or `YAKL_ARCH_CUDA`, so pay attention to those. Also, you'll need to ensure it's c++17.

bartgol:
> You could end up missing some of the needed CPP defs such as `YAKL_DEBUG`, `YAKL_B4B`, or `YAKL_ARCH_CUDA`, so pay attention to those. Also, you'll need to ensure it's c++17.

We should still get all the YAKL cpp/flags settings, since they are PUBLIC properties of the yakl target, so CMake will transfer them on downstream targets.

jgfouca (approved):

mrnorman:
I might be missing something, but YAKL doesn't set source file flags or properties on the YAKL CMake target to automatically propagate to all source files that link it. It expects you to do that with the `yakl_utils.cmake` file's macros either through `yakl_process_cxx_source_files` or `yakl_process_target` to *only* the files / targets that actually use YAKL.

This is because, typically, not every C++ file in a CMake target that links to YAKL needs to be processed the same as the files that *actually use* YAKL. It separates things out better.

mrnorman:
These links will help you figure out what's most appropriate for SCREAM based on how you want to handle it:
https://github.com/mrnorman/YAKL/wiki/Using-and-Compiling-with-YAKL
https://github.com/mrnorman/YAKL/blob/main/CMakeLists.txt
https://github.com/mrnorman/YAKL/blob/main/yakl_utils.cmake

The documentation is fairly detailed, and the CMake files are small and simple.

brhillman:
I'm a little bit out of the loop here, so I'm not 100% sure issue this is fixing. From other conversations, it sounds like this is fixing a build error that resulted from another PR to optimize compile flags for another system. I think Matt's macro was intended to keep YAKL flags from polluting the rest of the build environment, so that unrelated files would not need to be compiled with the same flags YAKL needed for whatever reason. Thus, YAKL source files could have one set of flags, Kokkos source files could have another. If it's desired to have all of the YAKL build flags and Kokkos build flags combined into a single global CXXFLAGS, then maybe this PR is what we want. But I think Matt was cautioning against that just to keep from polluting the build environment. But again, I'm a bit out of the loop here.

ambrad:
@brhillman the original issue was that the rrtmgp files were being compiled with -arch=sm_35 rather than sm_70.

bartgol:
> I might be missing something, but YAKL doesn't set source file flags or properties on the YAKL CMake target to automatically propagate to all source files that link it. It expects you to do that with the `yakl_utils.cmake` file's macros either through `yakl_process_cxx_source_files` or `yakl_process_target` to _only_ the files / targets that actually use YAKL.
> 
> This is because, typically, not every C++ file in a CMake target that links to YAKL needs to be processed the same as the files that _actually use_ YAKL. It separates things out better.

Ah, I forgot to mention an important detail: we are setting those on the yakl target ourselves:
```
target_compile_options(yakl PUBLIC $<$<COMPILE_LANGUAGE:CXX>:${YAKL_CXX_FLAGS}>)
```
yakl does not use CXX as compile language for its sources (well, its only source, `YAKL.cpp`), but the property is still stored, and passed along to other targets. I suspect we did this b/c in scream we don't change the language of our C++ files to CUDA, so we needed to attach these to yakl to get them propagated to rrtmgp (I suppose we could have just set these props on rrtmgp itself).

bartgol:
> I'm a little bit out of the loop here, so I'm not 100% sure issue this is fixing. From other conversations, it sounds like this is fixing a build error that resulted from another PR to optimize compile flags for another system. I think Matt's macro was intended to keep YAKL flags from polluting the rest of the build environment, so that unrelated files would not need to be compiled with the same flags YAKL needed for whatever reason. Thus, YAKL source files could have one set of flags, Kokkos source files could have another. If it's desired to have all of the YAKL build flags and Kokkos build flags combined into a single global CXXFLAGS, then maybe this PR is what we want. But I think Matt was cautioning against that just to keep from polluting the build environment. But again, I'm a bit out of the loop here.

The idea that "one target can use certain flags, while another target can use a different set of flags" is very nice, but in the case of header-rich libs, it is not as clean. Most of YAKL/Kokkos stuff is inside templated functions in headers. These will be compiled every time they are used, and in order to get the correct behavior, you need the correct flags every time. In particular, if you use YAKL in `rrtgmp` and `scream_rrtmgp`, they will both need YAKL flags/cpp settings. However, in `scream_rrtmgp` you also need access to kokkos headers, so the flags/cpp stuff has to mix in there.

The best thing we can do is to link yakl "privately" to CMake targets that use it, to prevent its settings to spill over to other targets. We are currently not doing that, but it appears not to cause issues (though we can certainly fix that). But, as I said, we cannot prevent kokkos and YAKL flags to mix in our rrtmgp interface, since we need to use headers from both libs.

mrnorman:
Currently in E3SM, we have an `ArrayIR` class that holds metadata for `Kokkos::View` and `yakl::Array` objects to transition between them so that one never needs to include both Kokkos and YAKL headers in the same target.

AaronDonahue:
I've been following the conversation, but I'm not sure where things landed.  @bartgol , @mrnorman is this ready to merge?

mrnorman:
Looks good to me

ambrad:
@AaronDonahue looks like there's a build error in the current state of this PR, so we shouldn't merge it.

AaronDonahue:
@ambrad , oh yes, you are correct.  Thanks.  I wasn't planning to manually merge anyway.  Just wanted to make sure that the AUTOMERGE label was appropriate.

bartgol:
I've seen the horizontal remap failure before. It seems to happen randomly, and only in the np>1 case. The fail is due to a `REQUIRE_THROWS(..)` where the exception is not thrown.

jgfouca (dismissed):

AaronDonahue:
If that is the only fail we see then I am okay with manually merging so that the other PRs can start passing the AT.  We can address the horizontal remap fail in a separate PR.

bartgol:
@AaronDonahue the fail was due to a `blah / 2` not generating diff big enough to trigger a check failure, in case `blah` was already small. I think I fixed it, but please, take a look, since I'm not familiar with the horiz remap stuff.

components/scream/src/share/tests/horizontal_remap_test.cpp
```diff
@@ -440,13 +440,19 @@ TEST_CASE("horizontal_remap_units", "") {
...
     Kokkos::deep_copy(weights_h,check_wgts);
-    weights_h(0) = weights_h(0)/2.0;
-    Real wgt_to_add = weights_h(0);
+    Real wgt_to_add = weights_h(0)/2;
+    weights_h(0) = 1-weights_h(0);
     Kokkos::deep_copy(check_wgts,weights_h);
     REQUIRE_THROWS(test_map.check());
+    // Restore w/2, so that adding a complementary weight segment should
+    // restore the valid map
+    weights_h(0) = wgt_to_add;
```
AaronDonahue, line 454: Do we need this line if we have the same 4 lines later?

AaronDonahue: oh wait, i understand now.

AaronDonahue (dismissed):

bartgol:
Closing, in favor of #2091

PR #2034, opened 2022-11-14
SCREAMv1 nightlies: Remove two MPASSI-based tests.

ambrad:
We're not using MPASSI right now. Remove two MPASSI-based tests that are failing in the pm-gpu nightly.


AaronDonahue (approved):

bartgol:
These all look like MPI finalization issues, like if OpenMPI RTE did not correctly release some resource. Nothing under our control. I'm tempted to merge as is, but I'll let the AT run standalone tests on mappy once more, just to be sure.

PR #2037, opened 2022-11-16
Precipitation mass flux diagnostics

tcclevenger:
Add precipitation mass flux diagnostics. Also correct units/unit conversions for precipitation mass.

### DIFF
--- a/components/scream/src/diagnostics/precip_ice_surf_mass_flux.cpp
+++ b/components/scream/src/diagnostics/precip_ice_surf_mass_flux.cpp
@@ -1,0 +1,49 @@
+#include "diagnostics/precip_ice_surf_mass_flux.hpp"
+
+namespace scream
+{
+
+// =========================================================================================
+PrecipIceSurfMassFluxDiagnostic::PrecipIceSurfMassFluxDiagnostic (const ekat::Comm& comm, const ekat::ParameterList& params)
+  : AtmosphereDiagnostic(comm,params)
+{
+  // Nothing to do here
+}
+
+// =========================================================================================
+void PrecipIceSurfMassFluxDiagnostic::set_grids(const std::shared_ptr<const GridsManager> grids_manager)
+{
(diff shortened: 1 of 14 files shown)


AaronDonahue (approved):
Looks good, thanks @tcclevenger .  I'm tagging @bartgol just for him to take a look at the change to the atm_diagnostic signature (i.e. passing `dt`), just in case he has an opinion.  He can take a look when he gets back to the office.  In the meantime I'll approve so it can be merged.

tcclevenger:
Thanks @AaronDonahue. I'll include my reasons here:

The `run_impl(dt)` already takes in `dt`, it just was not being used, so nothing extra needs to be passed. I chose to set `dt` inside `compute_diagnostics(dt)` so that it could be easily done by the tests, and the input `dt` is defaulted to 0, so the only extra work done is setting `m_dt=0` if a diagnostic does not use dt. But there could also be other ways to do this.

components/scream/src/diagnostics/precip_ice_surf_mass_flux.cpp
```diff
@@ -0,0 +1,49 @@
...
+// =========================================================================================
+void PrecipIceSurfMassFluxDiagnostic::compute_diagnostic_impl()
+{
+  const auto& precip_ice_surf_mass      = get_field_in("precip_ice_surf_mass").get_view<const Real*>();
+  const auto& precip_ice_surf_mass_flux = m_diagnostic_output.get_view<Real*>();
+
+  Kokkos::parallel_for("PrecipIceMassFluxDiagnostic",
+                       KT::RangePolicy(0,m_num_cols),
+                       KOKKOS_LAMBDA(const Int& icol) {
+    precip_ice_surf_mass_flux(icol) = precip_ice_surf_mass(icol)/PC::RHO_H2O/m_dt;
```
ambrad, line 45: Here's one class variable, m_dt.

PR #2041, opened 2022-11-17
Improve qv_sat to allow more caller info

### LINKED ISSUE #2040: Provide caller's name in physics_saturation_impl.hpp error message.
Callers of `qv_sat` should pass a label to the function that can be printed when a `check_temperature` check fails. For P3, which calls `qv_sat` in multiple spots, make unique labels for each, e.g. "P3_X", where "X" describes the specific call site.

jgfouca:
Fixes #2040

### DIFF
--- a/components/scream/src/physics/share/physics_saturation_impl.hpp
+++ b/components/scream/src/physics/share/physics_saturation_impl.hpp
@@ -14,7 +14,7 @@ namespace physics {
 template <typename S, typename D>
 KOKKOS_FUNCTION
 void  Functions<S,D>
-::check_temperature(const Spack& t_atm, const char* /* func_name */, const Smask& range_mask)
+::check_temperature(const Spack& t_atm, const char* caller, const Smask& range_mask)
 {
 
   //"range_mask" masks out packs which are padded with uninitialized values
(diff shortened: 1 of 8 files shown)


ambrad:
These need to be handled, too:
```
./src/physics/shoc/shoc_assumed_pdf_impl.hpp:242:        const Spack esval1_1 = scream::physics::Functions<S,D>::MurphyKoop_svp(Tl1_1,liquid,active_entries);
./src/physics/shoc/shoc_assumed_pdf_impl.hpp:243:        const Spack esval1_2 = scream::physics::Functions<S,D>::MurphyKoop_svp(Tl1_2,liquid,active_entries);
```

components/scream/src/physics/share/physics_saturation_impl.hpp
```diff
@@ -27,22 +27,22 @@ ::check_temperature(const Spack& t_atm, const char* /* func_name */, const Smask
...
 template <typename S, typename D>
 KOKKOS_FUNCTION
 typename Functions<S,D>::Spack
-Functions<S,D>::MurphyKoop_svp(const Spack& t_atm, const bool ice, const Smask& range_mask)
+Functions<S,D>::MurphyKoop_svp(const Spack& t_atm, const bool ice, const Smask& range_mask, const char* caller)
 {
 
   //First check if the temperature is legitimate or not
-  check_temperature(t_atm, "MurphyKoop_svp", range_mask);
+  check_temperature(t_atm, ekat::impl::strlen(caller) == 0 ? "MurphyKoop_svp" : caller, range_mask);
```
ambrad, line 45: A default to `nullptr` I think is more natural. Then here you wouldn't call `strlen` but instead just check if `caller == `nullptr` (or, idiomatically, simply `caller ? caller : "MurphyKoop_svp"`).

components/scream/src/physics/share/physics_saturation_impl.hpp
```diff
@@ -84,12 +84,12 @@ Functions<S,D>::MurphyKoop_svp(const Spack& t_atm, const bool ice, const Smask&
...
 KOKKOS_FUNCTION
 typename Functions<S,D>::Spack
-Functions<S,D>::polysvp1(const Spack& t, const bool ice, const Smask& range_mask)
+Functions<S,D>::polysvp1(const Spack& t, const bool ice, const Smask& range_mask, const char* caller)
 {
   // REPLACE GOFF-GRATCH WITH FASTER FORMULATION FROM FLATAU ET AL. 1992, TABLE 4 (RIGHT-HAND COLUMN)
 
   //First check if the temperature is legitimate or not
-  check_temperature(t, "polysvp1", range_mask);
+  check_temperature(t, ekat::impl::strlen(caller) ? "polysvp1" : caller, range_mask);
```
ambrad, line 92: Same here re: `caller`. Note that this is a function call, so every call to MurphyKoop_svp will encounter this line, and the args to `check_temperature` will be evaluated always. So we don't want to call `strlen` every time.

jgfouca:
@ambrad , fixed.

ambrad (approved):

ambrad:
Thanks!

AaronDonahue:
@jgfouca , is there any reason this PR is failing to build on `blake` and `weaver`, but not `mappy`?

PR #2044, opened 2022-11-18
Jgfouca/upstream merge 2022 11 18

jgfouca:
Conflicts were trivial except for config_machines.xml which I collaborated with Aaron on slack to resolve.


jgfouca:
The DIFFs are not real: `diff_test: the two files DIFFER only in their field lists` . I think this can be merged.

AaronDonahue (approved):

PR #2046, opened 2022-11-21
Fix CMake -O1 code for cori-knl.

ambrad:
Fixes Cori-KNL build failures on the dashboard.


jgfouca:
I think this maybe got merged as part of another PR? In any case, I approve.

PR #2050, opened 2022-11-22
Remove RMA dependency from coarsening remapper

AaronDonahue:
This commit fixes a number of issues with the implementation of the coarsening remapper
that were discovered after more extensive testing.  In particular, a few of the machines that
support scream don't support RMA operations in MPI.  The main body of this commit is to
revert the use of RMA to more basic MPI functions.

Over the course of developing this commit a handful of bugs were also discovered which have
been fixed.  In particular, the treatment of global id's in both the remap triplets and the source/
target grid.

This commit will allow us to develop the regionally refined output capability which will be in a
subsequent PR.

### DIFF
--- a/components/eamxx/src/share/grid/remap/coarsening_remapper.cpp
+++ b/components/eamxx/src/share/grid/remap/coarsening_remapper.cpp
@@ -68,6 +68,20 @@ CoarseningRemapper (const grid_ptr_type& src_grid,
   scorpio::grid_read_data_array(map_file,"S",  -1,S_h.data(),       nlweights);
   scorpio::eam_pio_closefile(map_file);
 
+  // Offset the cols ids to match the source grid.
+  // Determine the min id among the cols array, we
+  // also add the min_dof for the grid.
+  int remap_min_dof = col_gids_h[0];
+  for (int id=1; id<nlweights; id++) {
+    remap_min_dof = std::min(col_gids_h[id],remap_min_dof);
+  }
+  int global_remap_min_dof;
+  m_comm.all_reduce(&remap_min_dof,&global_remap_min_dof,1,MPI_MIN);
+
+  for (int ii=0; ii<nlweights; ii++) {
+    col_gids_h(ii) = col_gids_h(ii) - global_remap_min_dof + src_grid->get_global_min_dof_gid();
+  }
+
   // Create an "overlapped" tgt grid, that is, a grid where each rank
   // owns all tgt rows that are affected by at least one of the cols
   // in its src_grid
@@ -110,7 +124,7 @@ CoarseningRemapper (const grid_ptr_type& src_grid,
   auto weights_h     = Kokkos::create_mirror_view(m_weights);
 
   for (int i=0; i<nlweights; ++i) {
-    col_lids_h(i) = gid2lid(col_gids_h(id[i])-1,src_grid);
+    col_lids_h(i) = gid2lid(col_gids_h(id[i]),src_grid);
     weights_h(i)  = S_h(id[i]);
   }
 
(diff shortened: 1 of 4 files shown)


components/scream/src/share/grid/remap/coarsening_remapper.cpp
```diff
@@ -259,6 +273,8 @@ void CoarseningRemapper::do_remap_fwd ()
     // x is the src field, and y is the overlapped tgt field.
     const auto& f_src    = m_src_fields[i];
     const auto& f_ov_tgt = m_ov_tgt_fields[i];
+    const auto& src_layout = f_src.get_header().get_identifier().get_layout();
```
bartgol, line 276: This seems to be unused.

components/scream/src/share/grid/remap/coarsening_remapper.cpp
```diff
@@ -304,8 +320,10 @@ local_mat_vec (const Field& x, const Field& y) const
   using PackInfo    = ekat::PackInfo<PackSize>;
 
   const auto& src_layout = x.get_header().get_identifier().get_layout();
+  const auto& tgt_layout = y.get_header().get_identifier().get_layout();
```
bartgol, line 323: This seems to be unused.

components/scream/src/share/grid/remap/coarsening_remapper.cpp
```diff
@@ -304,8 +320,10 @@ local_mat_vec (const Field& x, const Field& y) const
...
   const int rank = src_layout.rank();
   const int nrows = m_ov_tgt_grid->get_num_local_dofs();
+  const int nrows_src = m_src_grid->get_num_local_dofs();
```
bartgol, line 326: This seems to be unused.

components/scream/src/share/grid/remap/coarsening_remapper.cpp
```diff
@@ -645,8 +663,19 @@ get_my_triplets_gids (const std::string& map_file,
...
+  int remap_min_dof = cols[0];
+  for (int id=1; id<nlweights; id++) {
+    remap_min_dof = std::min(cols[id],remap_min_dof);
+  }
+  int global_remap_min_dof;
+  m_comm.all_reduce(&remap_min_dof,&global_remap_min_dof,1,MPI_MIN);
+
   for (auto& id : cols) {
-    --id; // Subtract 1 to get 0-based indices
+    id = id - global_remap_min_dof + src_grid->get_global_min_dof_gid();
```
bartgol, line 678: This is fine, but I think that `-=` is more cplusplus-y than `id = id -` .... However, given that you do - and +, perhaps precompute the offset for legibility.

```
int offset = global_remap_min_dof - src_grid->get_global_min_dof_gid();
for (auto& id : cols) {
  id -= offset;
}
```

bartgol (changes requested):
Just some minor cleanups.

components/scream/src/share/grid/remap/coarsening_remapper.cpp
AaronDonahue: oh yes, good catch, these are leftover from some debugging.

components/scream/src/share/grid/remap/coarsening_remapper.cpp
AaronDonahue: good suggestion

components/eamxx/src/share/grid/remap/coarsening_remapper.cpp
```diff
@@ -68,6 +68,21 @@ CoarseningRemapper (const grid_ptr_type& src_grid,
   scorpio::grid_read_data_array(map_file,"S",  -1,S_h.data(),       nlweights);
   scorpio::eam_pio_closefile(map_file);
 
+  // Offset the cols ids to match the source grid.
+  // Determine the min id among the cols array, we
+  // also add the min_dof for the grid.
+  int remap_min_dof = col_gids_h[0];
```
bartgol, line 74: This might be an issue if we have a rank which reads no triplets, since `col_gids_h` will have size 0. I think you should init `remap_min_dof` with `std::numeric_limits<int>::max()`.

AaronDonahue:
All of these changes and more are included already in #2051

PR #2051, opened 2022-11-22
Add the capability to produce regional output.

AaronDonahue:
This commit will apply the new coarsening remapper to the scorpio output interface to give the user the ability to produce remapped output on the fly.
    
In order to trigger regional output the output control YAML file needs to have the "remap_file" parameter given. The value should be a string that points to the remap file to be used.

### DIFF
--- a/components/eamxx/src/share/grid/remap/coarsening_remapper.cpp
+++ b/components/eamxx/src/share/grid/remap/coarsening_remapper.cpp
@@ -68,6 +68,20 @@ CoarseningRemapper (const grid_ptr_type& src_grid,
   scorpio::grid_read_data_array(map_file,"S",  -1,S_h.data(),       nlweights);
   scorpio::eam_pio_closefile(map_file);
 
+  // Offset the cols ids to match the source grid.
+  // Determine the min id among the cols array, we
+  // also add the min_dof for the grid.
+  int remap_min_dof = col_gids_h[0];
+  for (int id=1; id<nlweights; id++) {
+    remap_min_dof = std::min(col_gids_h[id],remap_min_dof);
+  }
+  int global_remap_min_dof;
+  m_comm.all_reduce(&remap_min_dof,&global_remap_min_dof,1,MPI_MIN);
+
+  for (int ii=0; ii<nlweights; ii++) {
+    col_gids_h(ii) = col_gids_h(ii) - global_remap_min_dof + src_grid->get_global_min_dof_gid();
+  }
+
   // Create an "overlapped" tgt grid, that is, a grid where each rank
   // owns all tgt rows that are affected by at least one of the cols
   // in its src_grid
@@ -110,7 +124,7 @@ CoarseningRemapper (const grid_ptr_type& src_grid,
   auto weights_h     = Kokkos::create_mirror_view(m_weights);
 
   for (int i=0; i<nlweights; ++i) {
-    col_lids_h(i) = gid2lid(col_gids_h(id[i])-1,src_grid);
+    col_lids_h(i) = gid2lid(col_gids_h(id[i]),src_grid);
     weights_h(i)  = S_h(id[i]);
   }
 
@@ -158,7 +172,7 @@ create_src_layout (const FieldLayout& tgt_layout) const
   const auto lt = get_layout_type(tgt_layout.tags());
   auto src = FieldLayout::invalid();
   const bool midpoints = tgt_layout.has_tag(LEV);
-  const int vec_dim = tgt_layout.is_vector_layout() ? tgt_layout.get_vector_dim() : -1;
+  const int vec_dim = tgt_layout.is_vector_layout() ? tgt_layout.dim(CMP) : -1;
   switch (lt) {
     case LayoutType::Scalar2D:
       src = m_src_grid->get_2d_scalar_layout();
@@ -184,7 +198,7 @@ create_tgt_layout (const FieldLayout& src_layout) const
   const auto lt = get_layout_type(src_layout.tags());
   auto tgt = FieldLayout::invalid();
   const bool midpoints = src_layout.has_tag(LEV);
-  const int vec_dim = src_layout.is_vector_layout() ? src_layout.get_vector_dim() : -1;
+  const int vec_dim = src_layout.is_vector_layout() ? src_layout.dim(CMP) : -1;
   switch (lt) {
     case LayoutType::Scalar2D:
       tgt = m_tgt_grid->get_2d_scalar_layout();
@@ -259,6 +273,8 @@ void CoarseningRemapper::do_remap_fwd ()
     // x is the src field, and y is the overlapped tgt field.
     const auto& f_src    = m_src_fields[i];
     const auto& f_ov_tgt = m_ov_tgt_fields[i];
+    const auto& src_layout = f_src.get_header().get_identifier().get_layout();
+    const auto& tgt_layout = f_ov_tgt.get_header().get_identifier().get_layout();
 
     // Dispatch kernel with the largest possible pack size
     const auto& src_ap = f_src.get_header().get_alloc_properties();
@@ -304,8 +320,10 @@ local_mat_vec (const Field& x, const Field& y) const
   using PackInfo    = ekat::PackInfo<PackSize>;
 
   const auto& src_layout = x.get_header().get_identifier().get_layout();
+  const auto& tgt_layout = y.get_header().get_identifier().get_layout();
   const int rank = src_layout.rank();
   const int nrows = m_ov_tgt_grid->get_num_local_dofs();
+  const int nrows_src = m_src_grid->get_num_local_dofs();
   auto row_offsets = m_row_offsets;
   auto col_lids = m_col_lids;
   auto weights = m_weights;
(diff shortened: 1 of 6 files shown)


AaronDonahue:
WIP pending #2050

components/scream/src/share/grid/remap/coarsening_remapper.cpp
```diff
@@ -304,8 +320,10 @@ local_mat_vec (const Field& x, const Field& y) const
   using PackInfo    = ekat::PackInfo<PackSize>;
 
   const auto& src_layout = x.get_header().get_identifier().get_layout();
+  const auto& tgt_layout = y.get_header().get_identifier().get_layout();
```
bartgol, line 323: This appears to be unused.

components/scream/src/share/grid/remap/coarsening_remapper.cpp
```diff
@@ -259,6 +273,8 @@ void CoarseningRemapper::do_remap_fwd ()
     // x is the src field, and y is the overlapped tgt field.
     const auto& f_src    = m_src_fields[i];
     const auto& f_ov_tgt = m_ov_tgt_fields[i];
+    const auto& src_layout = f_src.get_header().get_identifier().get_layout();
+    const auto& tgt_layout = f_ov_tgt.get_header().get_identifier().get_layout();
```
bartgol, line 277: These appear to be unused.

components/scream/src/share/grid/remap/coarsening_remapper.cpp
```diff
@@ -304,8 +320,10 @@ local_mat_vec (const Field& x, const Field& y) const
   using PackInfo    = ekat::PackInfo<PackSize>;
 
   const auto& src_layout = x.get_header().get_identifier().get_layout();
+  const auto& tgt_layout = y.get_header().get_identifier().get_layout();
   const int rank = src_layout.rank();
   const int nrows = m_ov_tgt_grid->get_num_local_dofs();
+  const int nrows_src = m_src_grid->get_num_local_dofs();
```
bartgol, line 326: This appears to be unused.

components/scream/src/share/grid/remap/coarsening_remapper.cpp
```diff
@@ -645,8 +663,19 @@ get_my_triplets_gids (const std::string& map_file,
...
+  int remap_min_dof = cols[0];
+  for (int id=1; id<nlweights; id++) {
+    remap_min_dof = std::min(cols[id],remap_min_dof);
+  }
+  int global_remap_min_dof;
+  m_comm.all_reduce(&remap_min_dof,&global_remap_min_dof,1,MPI_MIN);
+
   for (auto& id : cols) {
-    --id; // Subtract 1 to get 0-based indices
+    id = id - global_remap_min_dof + src_grid->get_global_min_dof_gid();
```
bartgol, line 678: This is fine, but I think that `-=` is more cplusplus-y than `id = id - ...`. However, given that you do - and +, perhaps precompute the offset for legibility.
```
int offset = global_remap_min_dof + src_grid->get_global_min_dof_gid();
for (auto& id : cols) {
  id -= offset;
}
```

components/scream/src/share/io/scorpio_output.cpp
```diff
@@ -540,8 +550,14 @@ AtmosphereOutput::get_var_dof_offsets(const FieldLayout& layout)
   //       All we need to do in this routine is to compute the offset of all the entries
   //       of the MPI-local array w.r.t. the global array. So long as the offsets are in
   //       the same order as the corresponding entry in the data to be read/written, we're good.
+  // NOTE: In the case of regional output this rank may have 0 columns to write, thus, var_dof
+  //       should be empty, we check for this special case and return an empty var_dof.
   if (layout.has_tag(ShortFieldTagsNames::COL)) {
     const int num_cols = m_io_grid->get_num_local_dofs();
+    if (num_cols==0) {
+      var_dof = std::vector<scorpio::offset_t>(0);
```
bartgol, line 558: This line is not necessary. If `num_cols==0`, the layout size is 0, so the vector was already created with the correct length.

components/scream/src/share/io/scorpio_output.cpp
```diff
@@ -570,6 +586,10 @@ AtmosphereOutput::get_var_dof_offsets(const FieldLayout& layout)
     const int num_my_elems = layout2d.dim(0);
     const int ngp = layout2d.dim(1);
     const int num_cols = num_my_elems*ngp*ngp;
+    if (num_cols==0) {
+      var_dof = std::vector<scorpio::offset_t>(0);
```
bartgol, line 590: Same here.

bartgol (changes requested):
Just some minor cleanup things.

components/eamxx/src/share/grid/remap/coarsening_remapper.cpp
```diff
@@ -68,6 +68,25 @@ CoarseningRemapper (const grid_ptr_type& src_grid,
...
   scorpio::eam_pio_closefile(map_file);
 
+  // Offset the cols ids to match the source grid.
+  // Determine the min id among the cols array, we
+  // also add the min_dof for the grid.
+  // Note: The cols field is not guaranteed to have the
+  // min offset value, but the rows will be numbered 
+  // by 1 from least to greatest.  So we use row_gids to calculate
+  // the remap min.
+  int remap_min_dof = io_grid->get_num_global_dofs()*100;  // Really big INT
```
bartgol, line 78: This is not really safe. If the number of global dofs is by chance larger than 20M (which happens with ne1024), then the result of the RHS will overflow, causing min dof to be inited to some neg number (which will mess up the min logiv below).

I think the safest thing to do is to init the int with `std::numeric_limits<int>::max()`.

components/eamxx/src/share/grid/remap/coarsening_remapper.cpp
```diff
@@ -645,8 +667,20 @@ get_my_triplets_gids (const std::string& map_file,
...
+  // Determine the min id among the cols array, we
+  // also add the min_dof for the grid.
+  int remap_min_dof = ngweights*100;
+  for (int id=0; id<nlweights; id++) {
+    remap_min_dof = std::min(cols[id],remap_min_dof);
+  }
+  int global_remap_min_dof;
+  m_comm.all_reduce(&remap_min_dof,&global_remap_min_dof,1,MPI_MIN);
+
+  gid_t col_offset = global_remap_min_dof + src_grid->get_global_min_dof_gid();
```
bartgol, line 681: I think this should have a `-` instead of a `+`.

components/eamxx/src/share/io/scorpio_output.cpp
```diff
@@ -359,10 +369,11 @@ set_grid (const std::shared_ptr<const AbstractGrid>& grid)
       (grid->get_global_max_dof_gid()-grid->get_global_min_dof_gid()+1)==grid->get_num_global_dofs(),
       "Error! In order for IO to work, the grid must (globally) have dof gids in interval [gid_0,gid_0+num_global_dofs).\n");
 
-  EKAT_REQUIRE_MSG(m_comm.size()<=grid->get_num_global_dofs(),
-      "Error! PIO interface requires the size of the IO MPI group to be\n"
-      "       no greater than the global number of columns.\n"
-      "       Consider decreasing the size of IO MPI group.\n");
+//ASD  IS THIS STILL TRUE vv
```
bartgol, line 372: We should verify this and, if no longer true, remove these lines.

components/eamxx/src/share/io/scorpio_output.cpp
```diff
@@ -540,8 +551,14 @@ AtmosphereOutput::get_var_dof_offsets(const FieldLayout& layout)
   //       All we need to do in this routine is to compute the offset of all the entries
   //       of the MPI-local array w.r.t. the global array. So long as the offsets are in
   //       the same order as the corresponding entry in the data to be read/written, we're good.
+  // NOTE: In the case of regional output this rank may have 0 columns to write, thus, var_dof
+  //       should be empty, we check for this special case and return an empty var_dof.
   if (layout.has_tag(ShortFieldTagsNames::COL)) {
     const int num_cols = m_io_grid->get_num_local_dofs();
+    if (num_cols==0) {
+      var_dof = std::vector<scorpio::offset_t>(0);
```
bartgol, line 559: This line can be removed. `var_dof` is already created with size equal to layout size. If num_cols=0, the size is already 0.

components/eamxx/src/share/io/scorpio_output.cpp
```diff
@@ -570,6 +587,10 @@ AtmosphereOutput::get_var_dof_offsets(const FieldLayout& layout)
     const int num_my_elems = layout2d.dim(0);
     const int ngp = layout2d.dim(1);
     const int num_cols = num_my_elems*ngp*ngp;
+    if (num_cols==0) {
+      var_dof = std::vector<scorpio::offset_t>(0);
```
bartgol, line 591: Same here.

components/eamxx/src/share/tests/coarsening_remapper_tests.cpp
```diff
@@ -90,6 +90,230 @@ void print (const std::string& msg, const ekat::Comm& comm) {
...
+  // source and target grid.  This is typical in monotone remappers from
+  // fine to coarse meshes.
+
+  // -------------------------------------- //
+  //           Init MPI and PIO             //
+  // -------------------------------------- //
+
+  ekat::Comm comm(MPI_COMM_WORLD);
+
+  MPI_Fint fcomm = MPI_Comm_c2f(comm.mpi_comm());
```
bartgol, line 169: Not for this PR, perhaps, but we should change our interface fcn, and do the c2f conversion in there. It's kind of silly to have our C interfaces use a Fortran MPI comm (an int)...

bartgol (changes requested):

components/eamxx/src/share/grid/remap/coarsening_remapper.cpp
AaronDonahue: agreed

components/eamxx/src/share/io/scorpio_output.cpp
AaronDonahue: I think intuitively we can see that in regional output we will violate this.  The IO grid will be a subset of the main grid and if the regional output is small enough, say just a handful of ARM sites we can expect the number of mpi ranks to exceed the number of output points.  We may want to do something like have the `m_comm` just be a sub-comm of the main the comm group with just those ranks that are involved in writing the output.  Then we can keep this check.

bartgol (approved):

PR #2053, opened 2022-11-22
Crusher-GPU: Fix one source of build errors for some tests.

ambrad:
Other sources of errors will require SCORPIO updates and an upstream merge to get https://github.com/E3SM-Project/E3SM/pull/5323.


ambrad:
@oksanaguba FYI

PR #2064, opened 2022-11-29
Lower the T_mid minimum cutoff to 100 from 130

ndkeen:
Lower the T_mid minimum cutoff to 100 from 130
Allows cases to continue running if T is under 130 (but over 100)
[bfb]


ndkeen:
I merged this, but wanted to note here that we don't want T to ever dip below 130 and would prefer to not make this change. But while we debug the low temps, allowing cases to continue running is best option for now. Hopefully we can find a way to keep T within more normal range and then re-visit these min/max error-cutoffs.

Issue #2057, opened 2022-11-24
Crusher: ELM reports a very large number of "Ball-Berry" errors.


ambrad:
Now that SCREAMv1 is running deterministically on Crusher, I'm starting to look at some anomalies in the runs. The most prominent is a huge number of lines of this sort in e3sm.log:
```
7:  Ball-Berry error check - stomatal conductance error:
7:  0.,  52577.053632404488
```
A 10-day ne30pg2 run has a 421M e3sm.log file essentially 100% due to lines of this type.

ambrad:
@amametjanov @grnydawn have you ever seen the above warning from ELM in your Crusher runs? This is with crayclang, and LND is running with no OpenACC on the CPU. I'm wondering if you've disocvered some F90 flags that might solve this issue. Thanks.

ambrad:
I've confirmed an identical run on Ascent does not have this warning in e3sm.log.

ambrad:
`-O0` does not solve the issue. The build line with the `-O0` modification and excluding irrelevant inclusion stuff, is this:
```
python3 /autofs/nccs-svm1_home1/ambradl/crusher/ne30pg2_ne30pg2.F2010-SCREAMv1.1/Tools/e3sm_compile_wrap.py \
 /opt/cray/pe/craype/2.7.16/bin/ftn -DCNL -DCPRCRAY -DFORTRANUNDERSCORE -DHAVE_MPI -DHAVE_SLASHPROC -DLINUX -DMCT_INTERFACE -DMODAL_AER -DNDEBUG -DNO_R16 -D_PNETCDF \ 
... \
-I/autofs/nccs-svm1_home1/ambradl/repo/SCREAM/components/elm/src/utils -I/autofs/nccs-svm1_home1/ambradl/repo/SCREAM/components/elm/src/cpl -I/gpfs/alpine/cli133/proj-shared/ambradl/e3sm_scratch/crusher/ne30pg2_ne30pg2.F2010-SCREAMv1.1/bld/lnd/obj -I/opt/cray/pe/mpich/8.1.16/ofi/crayclang/10.0/include \
-em -J.   -f free -N 255 -h byteswapio -em -M1077 -hnoacc -I/opt/cray/pe/mpich/8.1.16/ofi/crayclang/10.0/include \
[...]

PR #2066, opened 2022-11-29
Crusher: Remove -hzero from crayclang line.

ambrad:
-hzero 0-inits stack-allocated variables, including intent(out) arguments. E3SM conventionally does not 0-init stack variables, as we want code to properly init variables explicitly.

-hzero can be useful to find bugs, however. Upstream PR 5334 fixes a bug in ELM in which intent(out) args should be intent(inout).

In our production configuration, we don't want -hzero. Remove it here.


ambrad:
See https://github.com/E3SM-Project/scream/issues/2057 for discussion.

ambrad:
ERS_Ln362.ne30pg2_ne30pg2.F2010-SCREAMv1 passes on Crusher, and its e3sm.log file doesn't have the LND Ball-Berry messages, as expected.

ambrad:
No need for testing since this is a config file for a specific machine that is not tested by the AT. I manually tested, as noted above.

PR #2072, opened 2022-11-30
Crusher: Enable threading specification for use in LND.

ambrad:
In our production configuration, we want to run nrank x nthread = 8x8 for LND and 8x1 for everything else. Fix crusher-gpu config_machines.xml entry so we can do that.


ambrad:
Tested on Crusher.

PR #2073, opened 2022-11-30
Read topography from file

tcclevenger:
Read topography from its own file (not the IC file) and in the grid manager. Also changes lat/lon and vertical coords to be read in the grid manager as well.

Currently just setting the topography file as the IC file until I figure out which topo files are correct to use.

### DIFF
--- a/components/eamxx/cime_config/namelist_defaults_scream.xml
+++ b/components/eamxx/cime_config/namelist_defaults_scream.xml
@@ -278,6 +278,20 @@ tweaking their $case/namelist_scream.xml file.
     <vertical_coordinate_filename type="file">UNSET</vertical_coordinate_filename>
     <vertical_coordinate_filename nlev="72">${DIN_LOC_ROOT}/atm/scream/init/vertical_coordinates_L72_20220927.nc</vertical_coordinate_filename>
     <vertical_coordinate_filename nlev="128">${DIN_LOC_ROOT}/atm/scream/init/vertical_coordinates_L128_20220927.nc</vertical_coordinate_filename>
+    <!-- TODO: Change to using Topo files -->
+    <!-- Current issue is what to do for aquaplanet -->
+    <topography_filename type="file">UNSET</topography_filename>
+    <topography_filename hgrid="ne4np4">${DIN_LOC_ROOT}/atm/scream/init/screami_ne4np4L72_20220823.nc</topography_filename>
+    <topography_filename hgrid="ne30np4" nlev="72">${DIN_LOC_ROOT}/atm/scream/init/screami_ne30np4L72_20220823.nc</topography_filename>
+    <topography_filename hgrid="ne30np4" nlev="128">${DIN_LOC_ROOT}/atm/scream/init/screami_ne30np4L128_20221004.nc</topography_filename>
+    <topography_filename hgrid="ne120np4">${DIN_LOC_ROOT}/atm/scream/init/screami_ne120np4L72_20220823.nc</topography_filename>
+    <topography_filename hgrid="ne256np4">${DIN_LOC_ROOT}/atm/scream/init/screami_ne256np4L128_ifs-20200120_20220914.nc</topography_filename>
+    <topography_filename hgrid="ne512np4">${DIN_LOC_ROOT}/atm/scream/init/screami_ne512np4L128_20220823.nc</topography_filename>
+    <topography_filename hgrid="ne1024np4">${DIN_LOC_ROOT}/atm/scream/init/screami_ne1024np4L128_era5-20131001-topoadj-16x_20220914.nc</topography_filename>
+    <topography_filename hgrid="ne1024np4" COMPSET=".*SCREAM%DYAMOND-1_ELM">${DIN_LOC_ROOT}/atm/scream/init/screami_ne1024np4L128_ifs-20160801-topoadj16x_20221011.nc</topography_filename>
+    <topography_filename hgrid="ne1024np4" COMPSET=".*SCREAM%DYAMOND-1-X6T_ELM">${DIN_LOC_ROOT}/atm/scream/init/screami_ne1024np4L128_ifs-20160801-topoadjx6t_20221011.nc</topography_filename>
+    <topography_filename hgrid="ne4np4" COMPSET=".*SCREAM%AQUA.*">${DIN_LOC_ROOT}/atm/scream/init/screami_aquaplanet_ne4np4L72_20220823.nc</topography_filename>
+    <topography_filename hgrid="ne30np4" COMPSET=".*SCREAM%AQUA.*">${DIN_LOC_ROOT}/atm/scream/init/screami_aquaplanet_ne30np4L128_20220823.nc</topography_filename>
   </grids_manager>
 
   <!-- List of nc files for loading inputs on specified grids -->
--- a/components/eamxx/cime_config/namelist_defaults_scream.xml
+++ b/components/eamxx/cime_config/namelist_defaults_scream.xml
@@ -278,20 +278,15 @@ tweaking their $case/namelist_scream.xml file.
     <vertical_coordinate_filename type="file">UNSET</vertical_coordinate_filename>
     <vertical_coordinate_filename nlev="72">${DIN_LOC_ROOT}/atm/scream/init/vertical_coordinates_L72_20220927.nc</vertical_coordinate_filename>
     <vertical_coordinate_filename nlev="128">${DIN_LOC_ROOT}/atm/scream/init/vertical_coordinates_L128_20220927.nc</vertical_coordinate_filename>
-    <!-- TODO: Change to using Topo files -->
-    <!-- Current issue is what to do for aquaplanet -->
+    <!-- Use non-pg topo files to avoid variable naming issues. For aquaplanet runs, set to "none". -->
     <topography_filename type="file">UNSET</topography_filename>
-    <topography_filename hgrid="ne4np4">${DIN_LOC_ROOT}/atm/scream/init/screami_ne4np4L72_20220823.nc</topography_filename>
-    <topography_filename hgrid="ne30np4" nlev="72">${DIN_LOC_ROOT}/atm/scream/init/screami_ne30np4L72_20220823.nc</topography_filename>
-    <topography_filename hgrid="ne30np4" nlev="128">${DIN_LOC_ROOT}/atm/scream/init/screami_ne30np4L128_20221004.nc</topography_filename>
-    <topography_filename hgrid="ne120np4">${DIN_LOC_ROOT}/atm/scream/init/screami_ne120np4L72_20220823.nc</topography_filename>
-    <topography_filename hgrid="ne256np4">${DIN_LOC_ROOT}/atm/scream/init/screami_ne256np4L128_ifs-20200120_20220914.nc</topography_filename>
-    <topography_filename hgrid="ne512np4">${DIN_LOC_ROOT}/atm/scream/init/screami_ne512np4L128_20220823.nc</topography_filename>
-    <topography_filename hgrid="ne1024np4">${DIN_LOC_ROOT}/atm/scream/init/screami_ne1024np4L128_era5-20131001-topoadj-16x_20220914.nc</topography_filename>
-    <topography_filename hgrid="ne1024np4" COMPSET=".*SCREAM%DYAMOND-1_ELM">${DIN_LOC_ROOT}/atm/scream/init/screami_ne1024np4L128_ifs-20160801-topoadj16x_20221011.nc</topography_filename>
-    <topography_filename hgrid="ne1024np4" COMPSET=".*SCREAM%DYAMOND-1-X6T_ELM">${DIN_LOC_ROOT}/atm/scream/init/screami_ne1024np4L128_ifs-20160801-topoadjx6t_20221011.nc</topography_filename>
-    <topography_filename hgrid="ne4np4" COMPSET=".*SCREAM%AQUA.*">${DIN_LOC_ROOT}/atm/scream/init/screami_aquaplanet_ne4np4L72_20220823.nc</topography_filename>
-    <topography_filename hgrid="ne30np4" COMPSET=".*SCREAM%AQUA.*">${DIN_LOC_ROOT}/atm/scream/init/screami_aquaplanet_ne30np4L128_20220823.nc</topography_filename>
+    <topography_filename hgrid="ne4np4">${DIN_LOC_ROOT}/atm/cam/topo/USGS-gtopo30_ne4np4_16x.c20160612.nc</topography_filename>
+    <topography_filename hgrid="ne30np4">${DIN_LOC_ROOT}/atm/cam/topo/USGS-gtopo30_ne30np4_16xdel2-PFC-consistentSGH.nc</topography_filename>
+    <topography_filename hgrid="ne120np4">${DIN_LOC_ROOT}/atm/cam/topo/USGS-gtopo30_ne120np4_16xdel2-PFC-consistentSGH.nc</topography_filename>
+    <topography_filename hgrid="ne256np4">${DIN_LOC_ROOT}/atm/cam/topo/USGS-gtopo30_ne256np4_16xdel2_consistentSGH_20220429.nc</topography_filename>
+    <topography_filename hgrid="ne512np4">${DIN_LOC_ROOT}/atm/cam/topo/USGS-gtopo30_ne512np4_16xconsistentSGH_20190212.nc</topography_filename>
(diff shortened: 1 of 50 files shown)


tcclevenger:
Regarding topography files in https://web.lcrc.anl.gov/public/e3sm/inputdata/atm/cam/topo/, I'm curious which to use for which CIME-cases in eamxx. Here are some questions I have:
- Seems to be two or more options for many of the cases. 
Ex. `ne4pg2`: `USGS-gtopo30_ne4np4pg2_16x_converted.c20200527.nc` and `USGS-gtopo30_ne4pg2_16xdel2-PFC-consistentSGH.c..>`. I can elaborate more on each case if that is helpful.
- Seem to be no "non-pg" files for ne30. Not sure if from the web view I just don't see all the options?
- Aqua doesn't have files (I don't think). phis=0 here, so maybe we can pass through info to the grid if it is aqua-planet and set topo, or create aqua files?

ambrad:
Re: available files, they all have `np4pg2` in their names:
```
-rw-rwsr-- 1 ac.mtaylor   E3SM 1.6M Jan 24  2020 USGS-gtopo30_ne30np4pg2_16xdel2.c20200108.nc
-rw-rwsr-- 1 ac.mtaylor   E3SM 112M Feb 20  2020 USGS-gtopo30_ne256np4pg2_16xdel2_20200213.nc
-rw-rwsr-- 1 ac.mtaylor   E3SM 1.8G Feb 20  2020 USGS-gtopo30_ne1024np4pg2_16xconsistentSGH_20190528_converted.nc
-rw-rwsr-- 1 ac.mtaylor   E3SM 445M Feb 20  2020 USGS-gtopo30_ne512np4pg2_16xconsistentSGH_20190212_converted.nc
-rw-r-Sr-- 1 ac.ambradl   E3SM 446K May 27  2020 USGS-gtopo30_ne16np4pg2_16xdel2_20200527.nc
-rw-r-Sr-- 1 ac.ambradl   E3SM  31K May 28  2020 USGS-gtopo30_ne4np4pg2_16x_converted.c20200527.nc
-rw-rwSr-- 1 ac.whannah   E3SM  25M Jun 23  2020 USGS-gtopo30_ne120np4pg2_16xdel2.nc
-rw-rwSr-- 1 ac.whannah   E3SM 3.5M Jul 22  2020 USGS-gtopo30_ne45np4pg2_16xdel2.c20200615.nc
-rw-r--r-- 1 ac.mtaylor   E3SM 1.8G Jun 14  2021 USGS-gtopo30_ne1024np4pg2_x6t-SGH.nc
-rw-r--r-- 1 ac.mtaylor   E3SM 112M Jun 14  2021 USGS-gtopo30_ne256np4pg2_x6t-SGH.nc
-rw-r--r-- 1 ac.mtaylor   E3SM 1.6M Jun 14  2021 USGS-gtopo30_ne30np4pg2_x6t-SGH.nc
```

ambrad:
Re: ne30 case, USGS-gtopo30_ne30np4pg2_16xdel2.c20200108.nc should have the same phi_s values on the dynamics grid as the SCREAM IC files. That may not be true, however, since the provenance of SCREAM topo is not available, which is one of the issues this PR is solving.

mt5555:
For now, lets leave the default as the "16x"  files.   But we eventually want to switch to "x6t" combined with the other improvements (pgrad_correction and hv_ref_profiles).

components/eamxx/src/dynamics/homme/atmosphere_dynamics.cpp
```diff
@@ -175,14 +167,14 @@ void HommeDynamics::set_grids (const std::shared_ptr<const GridsManager> grids_m
...
-  add_field<Updated >("phis",          FL({COL         },{ncols           }),m2/s2, pgn);
+  add_field<Updated> ("qv",            FL({COL,     LEV},{ncols,  nlev_mid}),Q,     pgn,"tracers",N);
   add_field<Computed>("p_int",         FL({COL,    ILEV},{ncols,  nlev_int}),Pa,    pgn,N);
   add_field<Computed>("p_mid",         FL({COL,     LEV},{ncols,  nlev_mid}),Pa,    pgn,N);
   add_field<Computed>("p_dry_int",     FL({COL,    ILEV},{ncols,  nlev_int}),Pa,    pgn,N);
   add_field<Computed>("p_dry_mid",     FL({COL,     LEV},{ncols,  nlev_mid}),Pa,    pgn,N);
   add_field<Computed>("omega",         FL({COL,     LEV},{ncols,  nlev_mid}),Pa/s,  pgn,N);
-  add_group<Updated>("tracers",pgn,N, Bundling::Required);
+  add_group<Updated> ("tracers",pgn,N, Bundling::Required);
+  create_helper_field("phis",{COL},{ncols},pgn);
```
bartgol, line 177: Aren't we storing phis in the grids? Can't we just grab it from there, instead of creating another field?

components/eamxx/src/dynamics/homme/atmosphere_dynamics.cpp
```diff
@@ -194,11 +186,21 @@ void HommeDynamics::set_grids (const std::shared_ptr<const GridsManager> grids_m
     add_field<Required>("horiz_winds",   FL({COL,CMP, LEV},{nc,2,nlev_mid}),m/s,   rgn,N);
     add_field<Required>("T_mid",         FL({COL,     LEV},{nc,  nlev_mid}),K,     rgn,N);
     add_field<Required>("ps",            FL({COL         },{nc           }),Pa,    rgn);
-    add_field<Required>("phis",          FL({COL         },{nc           }),m2/s2, rgn);
     add_group<Required>("tracers",rgn,N, Bundling::Required, DerivationType::Import, "tracers", pgn);
     fv_phys_rrtmgp_active_gases_init(grids_manager);
     // This is needed for the dp_ref init in initialize_homme_state.
     add_field<Computed>("pseudo_density",FL({COL,     LEV},{nc,  nlev_mid}),Pa,    rgn,N);
+    create_helper_field("phis_cgll",{COL},{nc},rgn);
```
bartgol, line 193: Same here.

components/eamxx/src/dynamics/homme/homme_grids_manager.cpp
```diff
@@ -268,6 +283,120 @@ build_pg_codes () {
...
+void HommeGridsManager::load_topography (const nonconstgrid_ptr_type& phys_grid,
+                                         const ci_string&             type) const {
+  using geo_view_device = AbstractGrid::geo_view_type;
+  using geo_view_host   = geo_view_device::HostMirror;
+
+  const int nlcols = phys_grid->get_num_local_dofs();
+
+  if (type == "PG2") {
+
+    // For PG2, homme AD interface will compute
```
bartgol, line 351: I'm not sure I fully understand why we can't load it. For PG2 cases, isn't the topo file storing both`phis` and `phis_d`? Or are we computing `phis` remapping `phis_d`?

components/eamxx/src/share/grid/mesh_free_grids_manager.cpp
```diff
@@ -2,6 +2,9 @@
 #include "share/grid/point_grid.hpp"
 #include "share/grid/se_grid.hpp"
 #include "share/grid/remap/do_nothing_remapper.hpp"
+#include "share/io/scorpio_input.hpp"
+
+#include "physics/share/physics_constants.hpp"
```
bartgol, line 7: I am reluctant to include physics/share stuff in the scream_share library. This is a header only, so no circular deps, but it's still confusing.

I see that you only need it for pi. Perhaps pi could be moved in `share/utils/scream_universal_constants.hpp`? It's fine to store it as a double, and let the assignment operator cast it down to float when doing `Real pi = constants::pi;`.

bartgol: Note: for bwd compatibility, we can put
```
static constexpr Scalar Pi = constants::pi;
```
in `physics/share/physics_constants.hpp`, so that we don't have to change all the files using pi.

components/eamxx/src/dynamics/homme/homme_grids_manager.cpp
tcclevenger: We could accurately load it, but then `fv_phys` code in `HommeDynamics` would immediately overwrite it.

The procedure is 
1. Load phis on GLL
2. Remap phys_gll -> phys_dyn
3. Compute phys_pg2 in `fv_phys` code

tcclevenger: Maybe in the future we want to load phis on pg2? I can change it to load phis for both GLL and PG2, and then just allow it to be overwritten. It's a bit more computation, but only in the setup phase so no big deal.

components/eamxx/src/dynamics/homme/atmosphere_dynamics.cpp
tcclevenger: I believe I need to use fields because I want to use them in the `m_ic_remapper` in `initialize_homme_state()`.

components/eamxx/src/dynamics/homme/homme_grids_manager.cpp
ambrad: No, we should not load phis_phys. It's in that file for use in other contexts, such as analysis, that don't have access to the projection routine. Load just phis_gll, and the projection will occur, as it already does, during the rest of initialization.

ambrad: I should add, as a reminder, that in the future some physics parameterizations may need data that exist only on the physics grid in the topo file, e.g., the SGH fields. (These are not L2 projections from a dynamics-grid field.)

tcclevenger: I reverted back to not loading PG2 topo, but I've kept the distinction between pg2 and non-pg2 files for topography_filename.

components/eamxx/src/dynamics/homme/atmosphere_dynamics.cpp
ambrad: My opinion: phi_s should remain a field like ps, unless there was some crucial need that motivated this change. A grid object most naturally has as data the topology and geometry of the nodes and edges, and nothing else. In principle, multiple models could use it, only some of which know about phi_s.

bartgol: Ah, yes, we need to be able to write them, and the grid fields are read only. I have a commit in a follow up PR that allows to extract nonconst geo data from a const grid. In that PR, I can remove this helper field then.

components/eamxx/src/dynamics/homme/homme_grids_manager.cpp
```diff
@@ -355,24 +353,24 @@ void HommeGridsManager::load_topography (const nonconstgrid_ptr_type& phys_grid,
...
 
-    // TODO: Use topo files instead of IC files. This hack is currently needed since
-    //       the topo files use uppercase PHIS, whereas IC files use lower case.
-    //       Right now the issue is that not all CIME compsets have equiv. topo
-    //       files. Example: aquaplanet file is needed with phis=0.
-#if 0
-    // If this is the GLL grid of a PG2 run, then the
-    // field name is PHIS_d in the topography file.
-    // Else the name is PHIS.
+    // If this is a PG2 run, the naming is "PHIS" for the PG2 grid,
```
bartgol, line 366: The PG2 case is already handled by the first branch of the if, so I think this branch can just hard-code PHIS, no?

bartgol (changes requested):

components/eamxx/src/share/grid/mesh_free_grids_manager.cpp
```diff
@@ -84,11 +87,168 @@ build_grids ()
...
+  lat_lon_reader.read_variables();
+  lat_lon_reader.finalize();
+
+  // Set geometry data
+  grid->set_geometry_data("lat", cmvc(host_views["lat"]));
+  grid->set_geometry_data("lon", cmvc(host_views["lon"]));
+}
+
+void MeshFreeGridsManager::
+load_vertical_coordinates (const nonconstgrid_ptr_type& grid) const
```
ambrad, line 185: There seems to be some code duplication. I'm wondering if this code duplication is a symptom of the vertical data not naturally living in a grid object.

components/eamxx/src/dynamics/homme/atmosphere_dynamics.cpp
ambrad: @bartgol I'm saying something a bit stronger than that. I think it's a design mistake to put phi_s in the grids class. What is the motivation for moving all these data there? As far as I can tell, it's not improving anything. In my opinion, if one just thinks of phi_s as ps, then it becomes clear where phi_s should live. As for the vertical coordinate, if you want to move it out of the dynamics AD object, then I suggest an object separate from "grid". Note that, e.g., a pg2 and GLL grid both have the same reference vertical coordinate, implying the vertical coordinate data are conceptually separate from the horizontal grids.

components/eamxx/cime_config/namelist_defaults_scream.xml
```diff
@@ -278,6 +275,23 @@ tweaking their $case/namelist_scream.xml file.
...
+    <topography_filename hgrid="ne120np4">${DIN_LOC_ROOT}/atm/cam/topo/USGS-gtopo30_ne120np4_16xdel2-PFC-consistentSGH.nc</topography_filename>
+    <topography_filename hgrid="ne120np4.pg2">${DIN_LOC_ROOT}/atm/cam/topo/USGS-gtopo30_ne120np4pg2_16xdel2.nc</topography_filename>
+    <topography_filename hgrid="ne256np4">${DIN_LOC_ROOT}/atm/cam/topo/USGS-gtopo30_ne256np4_16xdel2_consistentSGH_20220429.nc</topography_filename>
+    <topography_filename hgrid="ne256np4.pg2">${DIN_LOC_ROOT}/atm/cam/topo/USGS-gtopo30_ne256np4pg2_16xdel2_20200213.nc</topography_filename>
+    <topography_filename hgrid="ne512np4">${DIN_LOC_ROOT}/atm/cam/topo/USGS-gtopo30_ne512np4_16xconsistentSGH_20190212.nc</topography_filename>
+    <topography_filename hgrid="ne512np4.pg2">${DIN_LOC_ROOT}/atm/cam/topo/USGS-gtopo30_ne512np4pg2_16xconsistentSGH_20190212_converted.nc</topography_filename>
+    <topography_filename hgrid="ne1024np4">${DIN_LOC_ROOT}/atm/cam/topo/USGS-gtopo30_ne1024np4_16xconsistentSGH_20190528.nc</topography_filename>
+     <topography_filename hgrid="ne1024np4.pg2">${DIN_LOC_ROOT}/atm/cam/topo/USGS-gtopo30_ne1024np4pg2_16xconsistentSGH_20190528_converted.nc</topography_filename>
```
ambrad, line 293: mappy will show whether the ne30 and ne30pg2 cases are BFB, as they should be. But none of our testing will check that this PR is BFB for the production runs with ne1024pg2. Thus, please check that the phi_s field in a production ne1024pg2 IC file is BFB with the phi_s field in this topo file.

components/eamxx/src/dynamics/homme/atmosphere_dynamics.cpp
tcclevenger: @ambrad
> If this PR were to do only 2, would there be a lot fewer code changes?
 
Yes there would be a lot less changes, but even more duplicated code. The grid seemed like a common place to gather these things (lat/lon and vert coords were already stored as geometry data), but I guess it fell short of truly getting rid of almost all code duplication, which can easily be done. 

As far as keeping `phis` as a FM field, that is fine with me. There is no reason it can't be, I just had it as geometry data since it fit in with other geometry data lat/lon and vert coords in that it is loaded from file in the same way, but I see your point that it is not really grid data like lat/lon (though I don't quite understand why the vertical coords are not?).

I do think a separate PR needs to be done to remove the code duplication in loading lat/lon and vert coords from file, but that does not need to be done here. Maybe a separate utility function that gets a filename and list of fields and hands back the loaded data. Shouldn't that already exist for IC? Anyways, I'll trim this PR back to only be loading topo from file.

tcclevenger:
@bartgol @ambrad Ended up making a new PR since this one was very cluttered with other changes, and the comments here are to parts of the code that are no longer changed in the new PR. Exception is the verification of ne1024 phis values, which I still need to do.

tcclevenger:
@bartgol @ambrad Ended up making a new PR since this one was very cluttered with other changes, and the comments here are to parts of the code that are no longer changed in the new PR. Exception is the verification of ne1024 phis values, which I still need to do.

PR #2074, opened 2022-11-30
Crusher: Work around SCORPIO threading issue.

ambrad:
SCORPIO needs only the link line; avoiding threading flags in the compile lines works around the SCORPIO bool and extern "C" issues.


ambrad:
The AT can't test this. I manually tested it on Crusher.

PR #2075, opened 2022-11-30
Crusher: Set default PE layout for ne30* to get more tests passing.

ambrad:


ambrad:
@jgfouca are you OK with my making this change in this repo instead of upstream?

jgfouca (approved):

ambrad:
Some PEM and ERP nightly tests are failing b/c the default is 1 node. Change to 2. Also run with 8 threads for LND, as we will in our production runs.

PR #2076, opened 2022-11-30
Crusher: Fix MAX_TASKS_PER_NODE to align with default PEs layout.

ambrad:
I realized ne30 tests were running on 16 nodes instead of 2. 2 is enough except for probably PEM/ERP.ne30_ne30, and we don't have any of those (and don't want them, since pg2 is the grid used in practice).


ambrad:
ERS.ne30_ne30 has the right layout now, although it now fails during time stepping with a seg fault.

PR #2077, opened 2022-11-30
Remove hard-coded disabling of prescribed aero perturbation

brhillman:
Remove hard-coded disabling of prescribed aerosol perturbation that was added for SCREAMv0. This would cause an unintended DIFF in MMF tests when the scream repo is merged upstream to E3SM, and does not appear to actually be used by SCREAMv0 tests (tests appear to be BFB).

[B4B]


crterai (approved):
Looks good to me.

PR #2081, opened 2022-12-02
Set phis from separate topography file

tcclevenger:
Read topography from its own file (not the IC file). Starting over from comments in https://github.com/E3SM-Project/scream/pull/2073.

### DIFF
--- a/components/eamxx/src/control/atmosphere_driver.cpp
+++ b/components/eamxx/src/control/atmosphere_driver.cpp
@@ -315,7 +315,7 @@ void AtmosphereDriver::setup_surface_coupling_processes () const
   }
 }
 
-void AtmosphereDriver::set_precipitation_fields_to_zero () 
+void AtmosphereDriver::set_precipitation_fields_to_zero ()
 {
   auto phys_grid = m_grids_manager->get_grid("Physics");
 
@@ -612,7 +612,7 @@ void AtmosphereDriver::initialize_output_managers () {
     }
     om.setup_globals_map(m_atm_process_group->get_restart_extra_data());
   }
-  
+
   // Build one manager per output yaml file
   using vos_t = std::vector<std::string>;
   const auto& output_yaml_files = io_params.get<vos_t>("output_yaml_files",vos_t{});
@@ -857,6 +857,10 @@ void AtmosphereDriver::set_initial_conditions ()
   std::vector<FieldIdentifier> ic_fields_to_copy;
   std::map<std::string,std::vector<std::string>> fields_inited;
 
+  // Check which fields should be loaded from the topography file
+  std::map<std::string,std::vector<std::string>> topography_fields_names_nc;
+  std::map<std::string,std::vector<std::string>> topography_fields_names_eamxx;
+
   // Helper lambda, to reduce code duplication
   auto process_ic_field = [&](const Field& f) {
     const auto& fid = f.get_header().get_identifier();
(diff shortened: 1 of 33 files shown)


components/eamxx/src/control/atmosphere_driver.cpp
```diff
@@ -1027,9 +1045,79 @@ void AtmosphereDriver::set_initial_conditions ()
...
+                      "topography_filename or entry matching the field name "
+                      "was given in IC parameters.\n");
+    }
+  }
+
   m_atm_logger->info("  [EAMxx] set_initial_conditions ... done!");
 }
 
+void AtmosphereDriver::
+read_fields_from_file (const std::vector<std::string>& field_names_nc,
```
bartgol, line 1076: There's a lot of code duplication between this fcn and the one below. Can we make the one below call this one, like

```
read_field_from_file (field_names,field_names,grid_name,file_name,t0);
```
?

components/eamxx/tests/coupled/dynamics_physics/homme_shoc_cld_p3_rrtmgp/CMakeLists.txt
```diff
@@ -61,6 +61,7 @@ configure_file(${SCREAM_SRC_DIR}/dynamics/homme/tests/theta.nl
 
 # Ensure test input files are present in the data dir
 GetInputFile(init/${EAMxx_tests_IC_FILE_72lev})
+GetInputFile(../cam/topo/USGS-gtopo30_ne4np4_16x.c20160612.nc)
```
bartgol, line 64: Can we use `${TOPO_DATA_DIR}` instead of `../cam/topo`, here and also in the other CMakeLists files?

bartgol (changes requested):

components/eamxx/src/control/atmosphere_driver.cpp
tcclevenger: Yeah, let me look at this again. I would also like to add in the stuff from read lat/lon/vcoords from this file as well.

components/eamxx/tests/coupled/dynamics_physics/homme_shoc_cld_p3_rrtmgp/CMakeLists.txt
bartgol: Ah, I see now. I think the change of the base dir for the search path is a good idea.

bartgol (approved):

ambrad (approved):
Very clean. Backing up one directory in the `inputdata` specifications will definitely be useful in the future.

tcclevenger:
@ambrad Attempting to compare `phis` between files for ne1024. Using `cprnc` doesn't work (on it's face) since the variable names don't match. You know of a different tool that will allow me to compare variables with different names? If not I can at least compare the last few values..

ambrad:
Some example approaches: You can use python (xarray or netCDF4) with E3SM unified env on most machines and diff in python. Or you can `ncdump -v [variable name]` to text files and diff the text files after removing or modifying the variable names in the text files.

tcclevenger:
@ambrad The only .nc file I currently have access to (on summit) for ne1024 is the PG2 one. I verified that `USGS-gtopo30_ne1024np4pg2_16xconsistentSGH_20190528_converted.nc` match the base ne1024 IC file `screami_ne1024np4L128_era5-20131001-topoadj-16x_20220914.nc`. I can try and download all the others onto my machine, but that could take a while.

ambrad:
Great. That's the check that was needed.

ambrad:
Did you run some SCREAMv1 tests with this PR on a GPU machine?

PR #2082, opened 2022-12-02
Print all in/out fields upon Fatal property check fail in AtmosphereProcess

bartgol:
This PR adds a few concepts/features:

- Allow to print a field at a particular hyperslab. E.g., print `f(1,:,2,3,:)`.
- Allow property checks to return a location where the check failed
- When a property check fail (and fail checks are fatal), if the property check provided a location, we print all input/output fields of the atm process at that location. We only print fields with a "compatible" layout (e.g., if the location of the fail has tags `<COL,LEV>`, we won't print fields with layout `<EL,GP,GP,LEV>`.
- Add the XML option `driver_options::check_all_computed_fields_for_nans`, which defaults to true, and causes the AD to add a `FieldNaNCheck` property check for _all_ computed fields after _each_ atm process.

@ambrad, @PeterCaldwell @crterai @AaronDonahue  @brhillman Question for the model experts: I did not add a feature like in Homme, where we print a whole column worth of data when EOS errors out. Right now, if there's a failed check for a field at col=A,lev=B, I print all other fields at col=A,lev=B. I could probably prune the LEV/ILEV tag/index (if present) before calling the method that prints the hyperslab. Would that be preferable? Or is the current impl ok?

As an example, I hardcoded a NaN in shoc, and ran a test. It generated the following error message.

[snippet.txt](https://github.com/E3SM-Project/scream/files/10145054/snippet.txt)

Of course, formatting is up for debate.

### DIFF
--- a/components/eamxx/src/share/atm_process/atmosphere_process.cpp
+++ b/components/eamxx/src/share/atm_process/atmosphere_process.cpp
@@ -1,6 +1,7 @@
 #include "share/atm_process/atmosphere_process.hpp"
 #include "share/util/scream_timing.hpp"
 #include "share/property_checks/mass_and_energy_column_conservation_check.hpp"
+#include "share/field/field_utils.hpp"
 
 #include "ekat/ekat_assert.hpp"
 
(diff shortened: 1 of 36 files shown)


ambrad:
Because the output is for fatal errors and so output verbosity is not an issue, a dump of the whole column would be useful and not a burden in the logs. That's what Homme (and now Hommexx) do.

bartgol:
Ok, I can easily add that for 3d fields.

components/eamxx/src/share/atm_process/atmosphere_process.cpp
```diff
@@ -498,6 +545,35 @@ add_precondition_check (const prop_check_ptr& pc, const CheckFailHandling cfh)
...
+        s = "Fatal";
+        break;
+      case CheckFailHandling::Warning:
+        s = "Warning";
+        break;
+      default:
+        EKAT_ERROR_MSG ("Unexpected/unsupported CheckFailHandling value.\n");
+    }
+
+    return "";
```
AaronDonahue, line 561: why are we returning anything for this `void` function?

components/eamxx/src/share/field/field_utils_impl.hpp
```diff
@@ -594,6 +594,244 @@ ST field_min(const Field& f, const ekat::Comm* comm)
...
+  constexpr int max_per_line = 5;
+  if (curr_idx==tags.size()) {
+    // All slices have been taken. Print the whole input field.
+    const auto& orig_layout = get_orig_header()->get_identifier().get_layout();
+    const auto dims_left = get_dims_left(orig_layout);
+    auto dims_str = get_dims_str(orig_layout);
+
+    f.sync_to_host();
+    const int rank = layout.rank();
+    out << " -----------------------------------------------------------------------\n";
```
AaronDonahue, line 680: See my comment about format of output in the "conversation" section.  I think there are two many `-----....` lines which make it difficult to parse the error message.

AaronDonahue (dismissed):
All in all this looks good, thank you for putting it in Luca.  I have one comment on the format which I will put in the conversation boxes and open up to comment from others.

AaronDonahue:
I think there may be two many `------....` dividing lines, it ends up being that even the variable description gets separated from the value, for example
```
   -----------------------------------------------------------------------
   -----------------------------------------------------------------------
       qc<COL,LEV>(218,72)
   -----------------------------------------------------------------------
  
  qc(0,0)
      0, 
   -----------------------------------------------------------------------
   -----------------------------------------------------------------------
```

It also close to triples the number of lines needed for each entry.  Could we try something where the the variables are lumped together, like

```
qc<COL,LEV>(218,72)
   qc(0,0) = 
      0, 
```

So the indentation makes it possible to parse the output?  Or maybe just a `-------.....` type line above the variable layout line, which will have the effect of separating all variables by a single line.

components/eamxx/src/share/atm_process/atmosphere_process.cpp
bartgol: Yeah, that's not right. I'll fix it.

bartgol: Oh, no, wait, that's right. The return is for the lambda function, not the class method.

AaronDonahue (dismissed):
Thanks for the changes @bartgol

AaronDonahue (approved):

PR #2083, opened 2022-12-05
Revised SHOC TL Rails

bogensch:
Revise clipping condition for SHOC clipping on individual temperature Gaussians.  If they go below a specified minimum threshold then set to that threshold.


ambrad:
Blake fails because of a system problem, but Weaver and Mappy pass, so this is ready to go.

AaronDonahue (approved):

PR #2086, opened 2022-12-06
Crusher: Simplify modules.

ambrad:
Use default PrgEnv-cray rather than hardcoded version. Remove non-default libsci and cray-mpich; the defaults are loaded with PrgEnv-cray.


ambrad:
Tested on Crusher.

PR #2087, opened 2022-12-06
Fix treatment of diagnostics in IO

AaronDonahue:
This commit fixes an issue where the treatment of diagnostics in IO led to errors when writing diagnostics as regional output.  This redefines the `get_field` function in `scorpio_output` to accomodate multiple field managers (simulation and IO) which we get when doing regional output.

Note, diagnostics are a special case because they must be calculated on the simulation grid using field managed variables from the simulation, before they are remapped.

With this commit in it should be possible to do regional output for all available variables.

### DIFF
--- a/components/eamxx/src/share/io/scorpio_output.cpp
+++ b/components/eamxx/src/share/io/scorpio_output.cpp
@@ -53,7 +53,7 @@ AtmosphereOutput (const ekat::Comm& comm, const ekat::ParameterList& params,
       "Error! Unsupported averaging type '" + avg_type + "'.\n"
       "       Valid options: Instant, Max, Min, Average. Case insensitive.\n");
 
-  set_field_manager (field_mgr);
+  set_field_manager (field_mgr,"both");
 
   // By default, IO is done directly on the field mgr grid
   m_grids_manager = grids_mgr;
@@ -88,6 +88,9 @@ AtmosphereOutput (const ekat::Comm& comm, const ekat::ParameterList& params,
   // Try to set the IO grid (checks will be performed)
   set_grid (io_grid);
 
+  // Register any diagnostics needed by this output stream
+  set_diagnostics();
+
   bool remap_from_file = params.isParameter("remap_file");
 
   if (io_grid->name()!=fm_grid->name() || remap_from_file) {
@@ -104,7 +107,7 @@ AtmosphereOutput (const ekat::Comm& comm, const ekat::ParameterList& params,
     // Register all output fields in the remapper.
     m_remapper->registration_begins();
     for (const auto& fname : m_fields_names) {
-      auto f = get_field(fname);
+      auto f = get_field(fname,m_sim_field_mgr);
       const auto& src_fid = f.get_header().get_identifier();
       EKAT_REQUIRE_MSG(src_fid.data_type()==DataType::RealType,
           "Error! I/O supports only Real data, for now.\n");
@@ -123,8 +126,8 @@ AtmosphereOutput (const ekat::Comm& comm, const ekat::ParameterList& params,
 
     // Now that fields have been allocated on the io grid, we can bind them in the remapper
     for (const auto& fname : m_fields_names) {
-      auto src = get_field(fname);
-      auto tgt = io_fm->get_field(fname);
+      auto src = get_field(fname,m_sim_field_mgr);
+      auto tgt = io_fm->get_field(src.name());
       m_remapper->bind_field(src,tgt);
     }
 
@@ -133,7 +136,7 @@ AtmosphereOutput (const ekat::Comm& comm, const ekat::ParameterList& params,
         "Error! Something went wrong while building the scorpio input remapper.\n");
 
     // Reset the field manager
-    set_field_manager(io_fm);
+    set_field_manager(io_fm,"io");
   }
 
   // Setup I/O structures
@@ -161,9 +164,6 @@ void AtmosphereOutput::restart (const std::string& filename)
 
 void AtmosphereOutput::init()
 {
-  // Register any diagnostics needed by this output stream
-  set_diagnostics();
-
   for (const auto& var_name : m_fields_names) {
     register_dimensions(var_name);
   }
@@ -210,7 +210,7 @@ void AtmosphereOutput::run (const std::string& filename, const bool is_write_ste
   // Take care of updating and possibly writing fields.
   for (auto const& name : m_fields_names) {
     // Get all the info for this field.
-    const auto  field = get_field(name);
+    const auto  field = get_field(name,m_io_field_mgr);
     const auto& layout = m_layouts.at(name);
     const auto& dims = layout.dims();
     const auto  rank = layout.rank();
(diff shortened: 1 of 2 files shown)


components/eamxx/src/share/io/scorpio_output.cpp
```diff
@@ -662,23 +676,34 @@ setup_output_file(const std::string& filename,
   set_degrees_of_freedom(filename);
 }
 /* ---------------------------------------------------------- */
+// This routine will evaluate the diagnostics stored in this
+// output instance.
+void AtmosphereOutput::update_field(const std::string& name) const
+{
+  // Check if the field is a diagnostic, if so evaluate it. 
+  if (m_diagnostics.find(name) != m_diagnostics.end()) {
```
bartgol, line 684: We could remove this if, and call this method only for diag fields.

components/eamxx/src/share/io/scorpio_output.cpp
```diff
@@ -184,6 +184,12 @@ void AtmosphereOutput::run (const std::string& filename, const bool is_write_ste
 
   using namespace scream::scorpio;
 
+  // Update all diagnostics, we need to do this before applying the remapper
+  // to make sure that the remapped fields are the most up to date.
+  for (auto const& name : m_fields_names) {
+    update_field(name); // Note, if the field is a diagnostic this will update it.
```
bartgol, line 190: Can we make this more self explanatory? Namely:
- move the if statement in here and out of the `update_field` function
- rename to `compute_diagnostic` or something that makes it clear we are only processing a diag field

So something like this
```
for (...) {
  auto it = m_diagnostic.find(name);
  if (it != m_diagnostics.end()) {
    compute_diagnostic(*it);
  }
}
```

components/eamxx/src/share/io/scorpio_output.cpp
```diff
@@ -662,23 +676,34 @@ setup_output_file(const std::string& filename,
...
+  // Check if the field is a diagnostic, if so evaluate it. 
+  if (m_diagnostics.find(name) != m_diagnostics.end()) {
+    const auto& diag = m_diagnostics.at(name);
+    // Check if the diagnostics has any dependencies, if so, evaluate
+    // them as well.  Needed if a diagnostic relies on another
+    // diagnostic.
+    for (const auto& dep : m_diag_depends_on_diags.at(name)) {
+      update_field(dep);
+    }
+    diag->compute_diagnostic();
```
bartgol, line 692: I feel like we have to add some book keeping for which diags are computed, in order to avoid computing the same diag twice. This might happen if diag D1 depends on diag D2, and we are asking for both D1 and D2 in our output. In that case, D2 is computed twice: once when processing D2, and once as a dep of D1 when processing D1.

We should keep a simple `std::map<std::string,bool> m_diag_computed` stored in the class. At each run step, we do `m_diag_computed.clear()` right away, so that when we do `m_diag_computed[diag_name]` the first time it returns false. Then, inside this fcn, we update it to true once we're done computing it.

components/eamxx/src/share/io/scorpio_output.cpp
```diff
@@ -689,7 +714,7 @@ void AtmosphereOutput::set_diagnostics()
 {
   // Create all diagnostics
   for (const auto& fname : m_fields_names) {
-    if (!m_field_mgr->has_field(fname)) {
+    if (!m_io_field_mgr->has_field(fname)) {
```
bartgol, line 717: I thought we agreed on computing diags on the model grid, rather than the IO grid. Shouldn't this be `m_sim_field_mgr`?

components/eamxx/src/share/io/scorpio_output.cpp
```diff
@@ -53,7 +53,7 @@ AtmosphereOutput (const ekat::Comm& comm, const ekat::ParameterList& params,
       "Error! Unsupported averaging type '" + avg_type + "'.\n"
       "       Valid options: Instant, Max, Min, Average. Case insensitive.\n");
 
-  set_field_manager (field_mgr);
+  set_field_manager (field_mgr,"both");
```
bartgol, line 56: Assuming my comment inside `set_diagnostic` is correct, I believe we don't need the "both" mode. I believe we only need to set the sim FM at the beginning, since that's what we need to create diagnostics and, if needed, the remapper.

components/eamxx/src/share/io/scorpio_output.cpp
AaronDonahue: why does this need to be a `map`, doesn't `.clear()` just remove everything from the map?  If just being present in the map is enough for us to know it has been computed couldn't be use a `std::vector`?

bartgol: A map is better than a std::vector, since we would access it with a string key. If the key was an int (with small range), then a vector would do.

As for `clear()`, we need to reset all entries to "false" every time we run the output class. We could loop through the map and set all entries to `false`, but clearing the map works just fine (and it's less code), since calling `auto b = map[name];` when `name` is not found would create a new entry, using default ctor (which, for builtin `bool` would init the entry to `false`).

components/eamxx/src/share/io/scorpio_output.cpp
AaronDonahue: would it make more sense to just cycle through `m_diagnostics` rather than all of `m_field_names` and then check if they are a diagnostic?

components/eamxx/src/share/io/scorpio_output.cpp
AaronDonahue: okay, thanks for the explanation.  I'll use `map`

components/eamxx/src/share/io/scorpio_output.cpp
bartgol: Yes, that's fine too.

AaronDonahue:
Looks like this passed the AT, just needs the final round of review.

components/eamxx/src/share/io/scorpio_output.cpp
```diff
@@ -662,23 +678,37 @@ setup_output_file(const std::string& filename,
...
 // of available diagnostics.  If neither of these two options it
 // will throw an error.
-Field AtmosphereOutput::get_field(const std::string& name, const bool eval_diagnostic) const
+Field AtmosphereOutput::get_field(const std::string& name, const std::shared_ptr<const fm_type>& field_mgr) const
 {
-  if (m_field_mgr->has_field(name)) {
-    return m_field_mgr->get_field(name);
+  if (field_mgr->has_field(name)) {
+    return field_mgr->get_field(name);
   } else if (m_diagnostics.find(name) != m_diagnostics.end()) {
```
bartgol, line 710: Since we compute diags on the sim grid, we should be careful returning a diag here, in case field_mgr is not the sim FM. I think we should only check the diags if we're asking for a field from the sim FM, so we should do something ilke

```
} else {
   EKAT_REQUIRE_MSG (
     m_diagnostics.find(name) != m_diagnostics.end() && field_mgr==m_sim_field_mgr,
     "Field " + name + " not found in output field manager or diagnostics list");
```
This should protect us in case we accidentally ask a field that is not in the io FM, and we end up returning the diagnostic. It might be a corner case, but with some fields names changing (with diags), it might happen we don't register (or register with wrong name) the field in the remapper, leading to the field not being in the IO fm...

bartgol (changes requested):
Looks good. I have one last safety check request.

bartgol (approved):

PR #2091, opened 2022-12-07
EAMxx: separate rrtmgp into yakl and kokkos libraries

bartgol:
By that, I mean

- yakl libraries: targets that undergo the `yakl_process_target` macro (which changes the compiler language to CUDA on cuda gpu builds)
- kokkos libraries: libraries that are built like any other target in scream


components/eamxx/src/physics/rrtmgp/CMakeLists.txt
```diff
@@ -2,40 +2,41 @@ include(EkatUtils)
...
+  set(YAKL_CUDA_FLAGS "-DYAKL_ARCH_CUDA --expt-extended-lambda --expt-relaxed-constexpr -ccbin ${CMAKE_CXX_COMPILER}")
 endif()
 if (HIP_BUILD)
   set(YAKL_ARCH "HIP")
-  list(APPEND YAKL_CXX_FLAGS "-DYAKL_ARCH_HIP -O3 -D__HIP_ROCclr__ -D__HIP_ARCH_GFX90A__=1 --rocm-path=${ROCM_PATH} --offload-arch=gfx90a -x hip ")
+  set(YAKL_HIP_FLAGS "-DYAKL_ARCH_HIP -O3 -D__HIP_ROCclr__ -D__HIP_ARCH_GFX90A__=1 --rocm-path=${ROCM_PATH} --offload-arch=gfx90a -x hip")
 endif()
 
+set (YAKL_SOURCE_DIR ${SCREAM_BASE_DIR}/../../externals/YAKL)
+add_subdirectory(${YAKL_SOURCE_DIR} ${CMAKE_BINARY_DIR}/externals/YAKL)
```
brhillman, line 31: This might conflict with how YAKL is built in `build_model.cmake` for CIME-based builds. Have you tried a CIME-based case?

components/eamxx/src/physics/rrtmgp/CMakeLists.txt
```diff
@@ -59,8 +67,13 @@ set(EXTERNAL_SRC
 )
 add_library(rrtmgp ${EXTERNAL_SRC})
 EkatDisableAllWarning(rrtmgp)
-SetCudaFlags(rrtmgp)
-target_link_libraries(rrtmgp PUBLIC yakl)
+SetCudaFlags(rrtmgp CUDA_LANG)
```
brhillman, line 70: Is this necessary if we are using `yakl_process_target` below?

components/eamxx/src/physics/rrtmgp/CMakeLists.txt
```diff
@@ -81,25 +94,53 @@ target_link_options(rrtmgp PRIVATE -DKOKKOS_DEPENDENCE)
...
   mo_load_coefficients.cpp
   mo_load_cloud_coefficients.cpp
-  rrtmgp_test_utils.cpp
-  shr_orb_mod_c2f.F90
 )
-add_library(scream_rrtmgp ${INTERFACE_SRC})
+
+add_library(scream_rrtmgp_yakl ${SCREAM_RRTMGP_SOURCES_YAKL})
+yakl_process_target(scream_rrtmgp_yakl)
+SetCudaFlags(scream_rrtmgp_yakl CUDA_LANG)
```
brhillman, line 115: Again curious if we still need the `SetCudaFlags` here.

components/eamxx/src/physics/rrtmgp/atmosphere_radiation.cpp
```diff
@@ -845,7 +846,12 @@ void RRTMGPRadiation::run_impl (const int dt) {
     const int kbot = nlay+1;
 
     // Compute diffuse flux as difference between total and direct; use YAKL parallel_for here because these are YAKL objects
-    parallel_for(Bounds<3>(nswbands,nlay+1,ncol), YAKL_LAMBDA(int ibnd, int ilev, int icol) {
```
brhillman, line 848: Since this is no longer a YAKL parallel_for, we should edit the comment that says “use YAKL parallel for”

components/eamxx/src/physics/rrtmgp/atmosphere_radiation.cpp
```diff
@@ -845,7 +846,12 @@ void RRTMGPRadiation::run_impl (const int dt) {
...
+    Kokkos::parallel_for(Kokkos::RangePolicy<ExeSpace>(0,nswbands*(nlay+1)*ncol),
+                         KOKKOS_LAMBDA (const int idx) {
+      // CAREFUL: these are YAKL arrays, with "LayoutLeft". So make the indices stride accordingly
+      const int ibnd = (idx / ncol) / (nlay+1);
+      const int ilev = (idx / ncol) % (nlay+1);
+      const int icol =  idx % ncol;
       sw_bnd_flux_dif(icol,ilev,ibnd) = sw_bnd_flux_dn(icol,ilev,ibnd) - sw_bnd_flux_dir(icol,ilev,ibnd);
```
brhillman, line 855: Since these are YAKL arrays, I think they use 1-based instead of 0-based indexing, so should the indices here be `(icol+1,ilev+1,ibnd+1)`? I think the Kokkos parallel_for is looping over idx [0, nswbands*(nlay+1)*ncol - 1], so, e.g., for idx = 0, this gives icol = 0, but I think sw_bnd_flux_dif indexing starts at 1.

components/eamxx/src/physics/rrtmgp/CMakeLists.txt
bartgol: No, but this is exactly how it was built before (I just put the path in a var, to make it clearer). Why would it conflict though?

components/eamxx/src/physics/rrtmgp/CMakeLists.txt
bartgol: I think so. In debug mode, we add `--fmad=false` with EKAT, but YAKL doesn't. We could just add `--fmad=false` to `YAKL_CUDA_FLAGS` if it's a debug build, I guess.

components/eamxx/src/physics/rrtmgp/CMakeLists.txt
bartgol: I see what you mean, and now I'm wondering how it managed to work so far...

bartgol: Ah wait, that usage of YAKL is inside `if (COMP_NAME STREQUAL "eam")`, so it won't happen if we build eamxx. Still, I am adding a guard in case other components (e.g., Omega ocean) are building YAKL. Later, we might have to add a check to see if YAKL was pre-built as part of the sharedlib phase...

brhillman (dismissed):
Thanks for doing this Luca. We should make sure this doesn't break the CIME-based builds, but other than that I think it looks good.

bartgol:
I'm trying to figure out the v1 errors. The tests are failing at config time, with  an error I've never seen before:
```
CMake Error: Could not find cmake module file: CMakeDetermineFortanCompiler.cmake
CMake Error: Error required internal CMake variable not set, cmake may not be built correctly.
Missing variable is: 
CMAKE_Fortan_COMPILER_ENV_VAR
CMake Error: Error required internal CMake variable not set, cmake may not be built correctly.
Missing variable is: 
CMAKE_Fortan_COMPILER
CMake Error: Could not find cmake module file: /home/e3sm-jenkins/acme/scratch/SMS_D_Ln9.ne4_ne4.F2010-SCREAMv1-noAero.mappy_gnu9.C.20221207_173807_q67359/bld/cmake-bld/CMakeFiles/3.19.1/CMakeFortanCompiler.cmake
CMake Error at eamxx/CMakeLists.txt:76 (enable_language):
  No CMAKE_Fortan_COMPILER could be found.

  Tell CMake where to find the compiler by setting the CMake cache entry
  CMAKE_Fortan_COMPILER to the full path to the compiler, or to the compiler
  name if it is in the PATH.


CMake Error: Could not find cmake module file: CMakeFortanInformation.cmake
CMake Error: CMAKE_Fortan_COMPILER not set, after EnableLanguage
```

bartgol:
One day computers will be able to see that when I type `Fortan` I actually meant `Fortran`...

jgfouca (approved):

jgfouca:
Merging because we need builds to work again. @brhillman , feel free to make a follow on PR.

PR #2092, opened 2022-12-07
Crusher: Specialize v1_long test group to Crusher.

ambrad:
On ascent and pm-gpu, ERP_D_Lh182.ne4pg2_ne4pg2.F2010-SCREAMv1 takes < 8 min to run, but on crusher-gpu, it takes hours. Thus, make a special v1_long group for Crusher.


ambrad:
Tested on Crusher.

PR #2094, opened 2022-12-09
Add a way to track alternate field names in output for certain diagnostics

AaronDonahue:
This commit adds a map to scorpio_output that tracks alternate names that may exist for a specific output stream.  This is useful for sliced fields which have a name like X@tom for example but would take the name X_lev_0 under-the-hood.

When using a remapper the io_fm stores the latter name for the field while the field name itself is queried in `get_field` using the former name.  We didn't catch this before because the second check would default to calculating the diagnostic as expected.  It wasn't until we added the check that we only grab diagnostics on the sim_field_mgr that this issue was found.

### DIFF
--- a/components/eamxx/src/share/io/scorpio_output.cpp
+++ b/components/eamxx/src/share/io/scorpio_output.cpp
@@ -710,6 +710,8 @@ Field AtmosphereOutput::get_field(const std::string& name, const std::shared_ptr
   } else if (m_diagnostics.find(name) != m_diagnostics.end() && field_mgr==m_sim_field_mgr) {
     const auto& diag = m_diagnostics.at(name);
     return diag->get_diagnostic();
+  } else if (m_fields_alt_name.find(name) != m_fields_alt_name.end() && field_mgr->has_field(m_fields_alt_name.at(name))) {
+    return field_mgr->get_field(m_fields_alt_name.at(name));
   } else {
     EKAT_ERROR_MSG ("Field " + name + " not found in output field manager or diagnostics list, or requesting a diag not on the simualation field manager.");
   }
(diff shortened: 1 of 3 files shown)


AaronDonahue:
This may not be the best approach to solve the issue, so I am open to recommendations.  But we will need to fix the problem eventually if we want to start using regional output.  The summary of the problem is that calling `get_field(name,io_fm)` for a sliced diagnostic hits the EKAT_ERROR_MSG, the reason being that the io_fm stores the field in its altered name, like `T_mid_lev_0` rather than `T_mid@tom`.  So the string `T_mid@tom` is not found in the io_fm, and the second check requires the fm to be the sim_fm, so we throw the error.  

All this commit does is track any potential alternate names for the field in the field manager, and then try that before throwing an error.  I checked and it does resolve the problem with sliced output.

PeterCaldwell:
I'm curious whether this capability is also needed (or at least potentially helpful) for our goal of eventually writing out CMIP variable names rather than our internal variable names? Just a half-baked thought...

bartgol (approved):
If this works, then it's fine. One possible different solution could be a 2 level indirection, storing a `map<string,string> m_field_alias`, mapping any field alias whatsoever into the internal name that is actually used in the FM or diagnostic. So at the top of this fcn, one would do `auto name_internal = m_field_alias.at(name);`, and use that in the fcn body.

bartgol:
> I'm curious whether this capability is also needed (or at least potentially helpful) for our goal of eventually writing out CMIP variable names rather than our internal variable names? Just a half-baked thought...

I think that's a different issue. IO has no knowledge of "what the CF compliant names are", unless one stores a huge map `scream_name->CDF_name" for all possible fields in the ATM.

I think what we should do for tackling that issue is to optionally store a CF compliant name in FieldIdentifier, and, if present, use that in our output files.

PR #2097, opened 2022-12-12
SHOC assumed PDF NaN Output dump

bogensch:
Print out (hopefully) helpful information should a NaN be detected in Tl1_1 or Tl1_2.

### DIFF
--- a/components/eamxx/src/physics/shoc/impl/shoc_assumed_pdf_impl.hpp
+++ b/components/eamxx/src/physics/shoc/impl/shoc_assumed_pdf_impl.hpp
@@ -102,6 +102,7 @@ void Functions<S,D>::shoc_assumed_pdf(
     const auto qwthlsec = qwthl_sec_zt(k);
     const auto wqwsec = wqw_sec_zt(k);
     const auto wthlsec = wthl_sec_zt(k);
+    const auto w2sec = w_sec(k);
 
     // Compute square roots of some variables so we don't have to compute these again
     const auto sqrtw2 = ekat::sqrt(w_sec(k));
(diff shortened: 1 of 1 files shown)


components/eamxx/src/physics/shoc/impl/shoc_assumed_pdf_impl.hpp
```diff
@@ -229,6 +230,27 @@ void Functions<S,D>::shoc_assumed_pdf(
...
       const Smask active_entries = (index_range < nlev);
 
+      // Do NaN Checking here
+      const Smask is_nan_Tl1_1 = (Tl1_1 != Tl1_1) && active_entries;
+      const Smask is_nan_Tl1_2 = (Tl1_2 != Tl1_2) && active_entries;
+      if (is_nan_Tl1_1.any() || is_nan_Tl1_2.any()) {
+        printf("WARNING: NaN Detected in Tl1_1 or Tl1_2! \n");
+	for (int i=0; i<is_nan_Tl1_1.n; i++) {
+          if (is_nan_Tl1_1[i] || is_nan_Tl1_2[i]) {
+	    printf("Tl1 NaN Detected: Tl1_1, Tl1_2: \n");
```
AaronDonahue, line 240: I think this set of `printf` statements might allow for multiple ranks to be writing at the same time which may make it confusing to dechiper the warning messages.  Would an `osstream` be the better approach here?  @bartgol do you have any thoughts?

bartgol: It's technically possible. Putting everything together in one printf statement may help. However, ostream cannot be used, since this fcn is to be used on device. If you really want to avoid ranks overlap, you need to use a single printf statement. That doesn't have to make the code ugly, since compile time C strings are cat-ed across lines. So this works:

```
printf ("Tl1 NaN Detected: Tl1_1, Tl1_2: \n"
         "%4.9f,  %4.9f \n"
         "Tl1 NaN Detected, thetal, qw, temp, pressure, thl_sec, qw_sec, w2sec, lev: \n"
         " %4.9f, %2.9f, %4.9f, %8.9f, %4.9f, %2.9f, %4.9f %d \n"
         ...
         " %2.9f, %2.9f, %2.9f, %2.9f \n",
        Tl1_1, Tl1_2,
        thl_first[i], qw_first[i], Tl1_g[i], pval[i], thlsec[i], qwsec[i], w2sec[i], k,
        ...
        qw1_1[i], qw1_2[i], qw2_1[i], qw2_2[i]);
```
Notice *no commas* between consecutive C strings. The compiler splices them together right away.

bogensch: Thanks much, @bartgol, this is very helpful.

ambrad:
@bartgol @AaronDonahue what do you think of the consolidated printf statement?

ambrad:
Blake failed with a system problem. Weaver and Mappy passed. Thus, if @bogensch and others are fine with my commit, this can be merged.

bogensch:
looks good, thanks.

ambrad:
Great. I'll fix the conflict, retest, then merge.

bartgol:
> @bartgol @AaronDonahue what do you think of the consolidated printf statement?

Works for me.

bartgol (approved):

ambrad:
@bartgol what do you think of the above where we see "Timed out waiting for job SCREAM_PullRequest_Autotester_Blake to start" despite having the label "AT: Skip blake"? There was a reasonable time gap between when that label was applied and the AT started its round.

bartgol:
> @bartgol what do you think of the above where we see "Timed out waiting for job SCREAM_PullRequest_Autotester_Blake to start" despite having the label "AT: Skip blake"? There was a reasonable time gap between when that label was applied and the AT started its round.

See [this comment](https://github.com/E3SM-Project/scream/pull/1704#issuecomment-1351676927). The label is perhaps misleading, but that's the best we could do with AT1.0. Hopefully AT2.0 will be more powerful in this regard.

Issue #2080, opened 2022-12-02
Crusher: Nondeterministic seg faults in SCORPIO's flush_output_buffer's call to PnetCDF's ncmpi_wait_all


ambrad:
The current test failures on the Crusher dashboard are all in the RUN phase, except for one irrelevant build error. The e3sm.log file typically shows a seg fault with no further information.

I've traced this seg fault to flush_output_buffer in pio_darray_int.c. The seg fault occurs in this routine's call to PnetCDF's ncmpi_wait_all. Print statements right before and after this call show ncmpi_wait_all does not return in the case of a seg fault.

I often see this failure occur in the cpl.r write, after scream.r has already been written, but I don't think that's significant.

I'm using SCORPIO @ 7a1a1b3, the current version of SCORPIO in both E3SM and SCREAM repos.

The CDash line of interest is this one: https://my.cdash.org/viewTest.php?buildid=2252574

ambrad:
FYI, @jayeshkrishna @dqwu

jayeshkrishna:
@ambrad in your runs are there any tests that reliably reproduces (more than other tests) the crash/failures (I tried running ERP_D_Lh4.ne4_ne4.F2010-SCREAMv1.crusher-gpu_crayclang three times and it always passed for me)?

ambrad:
As far as I can tell, no. I experienced the same in the course of analyzing the problem. We can watch the dashboard entry for crusher at the link in my original post to see if there are tests that end up failing more frequently than others. I got the dashboard to its current state only on Wed night, so I haven't yet seen enough data to detect any patterns.

Fortunately, we can run ./case.submit from the test directory many times pretty quickly to get some failures. My approach to homing in on the ncmpi_wait_all call was to put in a bunch of print statements, ./case.submit repeatedly until I got a failure, read the log, then put in more statements based on what I determined, etc., until I had narrowed it down fully.

I use a script like the following to run repeats in a test directory until failure:
```bash
pass=0
fail=0
while true; do
    rm -f TestStatus
    # If running ERS/P, cp the original (not post-run) env_run.xml to
    # env_run.xml.save and then cp env_run.xml.save back to env_run.xml prior to
    # submitting. This assures the file is in the 'initial' state.
    #cp env_run.xml.save env_run.xml
    ./case.submit
    while true; do
        if [ -f TestStatus ]; then
            ln=`grep COMPARE_ TestStatus`
            echo $ln
[...]

PR #2102, opened 2022-12-14
Crusher: Update CDash configuration.

ambrad:
* Use 2 nodes instead of 1 for ne30(pg2) tests to catch any inter-node issues.
* Restore use of PnetCDF.
* Work around Cray-MPICH MPI_File* bug. See issue #2080.

Previously, I set ne30(pg2) tests to run with 8 LND threads per rank. Combined with this PR, the ne30(pg2) tests on the dashboard will run in the current best science configuration.


ambrad:
These two tests pass and have the desired configuration changes:
* ERS_Ln362.ne30pg2_ne30pg2.F2010-SCREAMv1.crusher-gpu_crayclang
* ERS_Ln9.ne4_ne4.F2000-SCREAMv1-AQP1.crusher-gpu_crayclang

PR #2104, opened 2022-12-15
Update vertical remapping to support 3+ dimensional variables

AaronDonahue:
This commit updates the vertical remapper to handle N dimension arrays.
    
    The previous implementation was limited to only working with 2D arrays
    with layout COL x LEV/ILEV.  This meant we could not easily vertically
    remap fields with 3+ dimensions, such as horiz_winds.  Not was it
    straightforward to remap a single vertical column.  This set of commits
    will address that.
    
    This first commit creates a signature for the main vertical interpolation
    routine that supports an Nd view rather than just 2d.  Subsequent commits
    will introduce the special 1d view case and will change the signature of
    the main routine to be Nd, currently it is still set to 2D.


AaronDonahue:
@bartgol and @rebassoo .  This PR is the first one in a series to get us to using vertical remapping in output.  I had to update the vertical remapper interface to allow me to pass arrays that were not just 2D, consider the case when we will want to remap `horiz_winds` for example, which is technically a 3D view.  I put you both as reviewers in case I missed something important, made an obvious mistake or could do better than I did here.  Feel free to add any comments.  I'm working on adding a test for the 3D and 4D versions then the PR should be ready to review/merge.

components/eamxx/src/share/util/scream_vertical_interpolation.hpp
```diff
@@ -10,20 +10,23 @@
...
+ * Also a user can decide to provide a view_Nd mask which will be filled
+ * based on whether each point is masked or not.
+ * Masking occurs when a value is out-of-bounds (i.e. requires an 
+ * extrapolation).
+ * The main function assumes that the user is providing as input:
+ *   a 2D view for the source vertical levels to interpolate from
+ *   a 1D view for the target vertical levels to interpolate onto
+ *   a set of Nd views for the input data, output data and an optional mask
+ *     where Nd is a N-dimensional view of dimension 2-4.
+ * There is a function, perform_vertical_interpolation_impl_1d which
```
bartgol, line 27: "which where"

AaronDonahue: :-)

AaronDonahue:
Is it safe to say that the fails on `weaver` are not related to this PR.  They seem to be in tests that shouldn't care about these changes.

bartgol:
> Is it safe to say that the fails on `weaver` are not related to this PR. They seem to be in tests that shouldn't care about these changes.

They are the same failures that we observed in our nightlies last night, so I would say yes.

components/eamxx/src/share/util/scream_vertical_interpolation.hpp
```diff
@@ -128,16 +119,62 @@ void perform_vertical_interpolation_impl_1d(
...
+  const int                         nlevs_tgt);
+
+template<typename T, int P> 
+void apply_interpolation(
+  const                        int  num_levs,
+  const                          T  mask_val,
+  const                   LIV<T,P>& vert_interp,
+  const view_2d<const Pack<T,P>>&   x_src,
+  const view_2d<const Pack<T,P>>&   x_tgt,
+  const view_Nd<const Pack<T,P>,2>& input,
```
bartgol, line 141: Any reason not to use `view_2d<T>` instead of `view_Nd<T,2>`?

components/eamxx/src/share/util/scream_vertical_interpolation.hpp
```diff
@@ -128,16 +119,62 @@ void perform_vertical_interpolation_impl_1d(
...
+  const view_Nd<        Mask<P>,2>& mask);
+
+template<typename T, int P> 
+void apply_interpolation(
+  const                        int  num_levs,
+  const                          T  mask_val,
+  const                   LIV<T,P>& vert_interp,
+  const view_2d<const Pack<T,P>>&   x_src,
+  const view_2d<const Pack<T,P>>&   x_tgt,
+  const view_Nd<const Pack<T,P>,3>& input,
```
bartgol, line 152: `view_3d` is also available in ekat.

components/eamxx/src/share/util/scream_vertical_interpolation.hpp
```diff
@@ -128,16 +119,62 @@ void perform_vertical_interpolation_impl_1d(
...
+  const view_Nd<        Mask<P>,3>& mask);
+
+template<typename T, int P> 
+void apply_interpolation(
+  const                        int  num_levs,
+  const                          T  mask_val,
+  const                   LIV<T,P>& vert_interp,
+  const view_2d<const Pack<T,P>>&   x_src,
+  const view_2d<const Pack<T,P>>&   x_tgt,
+  const view_Nd<const Pack<T,P>,4>& input,
```
bartgol, line 163: Do we really have a use case for 4d?

components/eamxx/src/share/util/scream_vertical_interpolation_impl.hpp
```diff
@@ -118,39 +52,248 @@ void perform_vertical_interpolation_impl_1d(
...
+  // The input data and x_src data should match in the appropriate size
+  EKAT_REQUIRE(x_src.extent_int(0) == input.extent_int(0));
+  EKAT_REQUIRE(x_src.extent_int(1) == input.extent_int(input.rank-1));
+  // The output data and x_tgt data should match in the appropriate size
+  EKAT_REQUIRE(x_tgt.extent_int(0) == output.extent_int(0));
+  EKAT_REQUIRE(x_tgt.extent_int(1) == output.extent_int(input.rank-1));
+
+
+  // The output and input data should match in rank
+  EKAT_REQUIRE(input.rank == output.rank);
```
bartgol, line 79: This check is poinltess. The two template args are already view_Nd with the same N. There's no way the compiler would pick this template signature if input/output had different ranks.

components/eamxx/src/share/util/scream_vertical_interpolation_impl.hpp
```diff
@@ -118,39 +52,248 @@ void perform_vertical_interpolation_impl_1d(
...
-  const view_Nd<const Pack<T,P>,2>& x_src,
+void perform_checks(
+  const view_2d<const Pack<T,P>>&   x_src,
+  const view_2d<const Pack<T,P>>&   x_tgt,
+  const view_Nd<const Pack<T,P>,N>& input,
+  const view_Nd<      Pack<T,P>,N>& output,
+  const int                         nlevs_src,
+  const int                         nlevs_tgt)
+{
+  auto rank = input.rank;
```
bartgol, line 67: Rank is N by construction, so you could just use N here.

components/eamxx/src/share/util/scream_vertical_interpolation_impl.hpp
```diff
@@ -118,39 +52,248 @@ void perform_vertical_interpolation_impl_1d(
...
+  EKAT_REQUIRE(x_src.extent_int(1) == input.extent_int(input.rank-1));
+  // The output data and x_tgt data should match in the appropriate size
+  EKAT_REQUIRE(x_tgt.extent_int(0) == output.extent_int(0));
+  EKAT_REQUIRE(x_tgt.extent_int(1) == output.extent_int(input.rank-1));
+
+
+  // The output and input data should match in rank
+  EKAT_REQUIRE(input.rank == output.rank);
+  // The output data and the input data should match in all sizes except the last one
+  for (int ii=0;ii<rank-1;ii++) {
```
bartgol, line 81: if you want, you could do
```
EKAT_REQUIRE (input.layout()==output.layout());
```

components/eamxx/src/share/util/scream_vertical_interpolation.hpp
AaronDonahue: No reason then just a bunch of copy/pastes and changing a single number.  I can change this and all others to `view_2d`, `view_3d`, etc. if that is desired.

components/eamxx/src/share/util/scream_vertical_interpolation.hpp
AaronDonahue: Not that I can think of offhand - but it was easy to add.  In IO we support up to rank 6, so I was tempted to go that high but I think case 5 and 6 are even more unlikely.  I assume if we ever hit that someone can add the equivalent vertical interpolation case.

So TLDR; I went up to 4 because IO supports up to 6, but I thought 5 and 6 were pretty unlikely.

If you want I can remove this case.

components/eamxx/src/share/util/scream_vertical_interpolation.hpp
bartgol: It's not really a big deal, but since you are already using view_2d, I think it looks cleaner. At least to me. No pressure though, do what you like best.

components/eamxx/src/share/util/scream_vertical_interpolation.hpp
bartgol: IO will _really_ need only up to 5, for dyn grid stuff (tracers and horiz_winds). 6 is never really used other than for dyn internal fields with time levels (but the outside world never sees the time level index).

Since dyn output is usually needed only for debug or restart purposes, I don't think we need to support rank>3. Each of these fcns need to be tested, and tests must be maintained. So if we realistically don't use it, I would just get rid of 4d.

components/eamxx/src/share/util/scream_vertical_interpolation_impl.hpp
AaronDonahue: but we want to skip checking the last dimension because we expect output to be on a different set/number of vertical levels.

components/eamxx/src/share/util/scream_vertical_interpolation.hpp
AaronDonahue: okay cool, I'll remove it.

AaronDonahue:
I can't access the `mappy` build fails, so I'm not sure why this is failing the AT.  It seems to build on my LC machines, and last Friday I was able to build on weaver.  Is it possible that the fails on the AT are unrelated to this PR?  If not, can someone w/ access to `mappy` investigate and let me know what the problem may be.

Note, I was able to track down the weaver fail to the lines which add a default argument to `perform_vertical_interpolation`, so maybe that is the issue.  But with weaver not working in general at the moment (see CDash) I'm not able to test.

bartgol:
```
/home/e3sm-jenkins/jenkins-ws/workspace/SCREAM_PullRequest_Autotester_Mappy/3461/scream/components/eamxx/src/share/util/scream_vertical_interpolation_impl.hpp:87:6: error: redeclaration of ‘template<class T, int P, int N> void scream::vinterp::perform_vertical_interpolation(scream::vinterp::view_2d<const ekat::Pack<T, n> >&, scream::vinterp::view_2d<const ekat::Pack<T, n> >&, scream::vinterp::view_Nd<const ekat::Pack<T, n>, N>&, scream::vinterp::view_Nd<ekat::Pack<T, n>, N>&, scream::vinterp::view_Nd<ekat::Mask<n>, N>&, int, int, scream::Real)’ may not have default arguments [-fpermissive]
   87 | void perform_vertical_interpolation(
      |      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

AaronDonahue:
@bartgol , thanks for your help.  This is passing the AT now.  When you get a chance can you re-review?

bartgol (dismissed):

components/eamxx/src/diagnostics/register_diagnostics.hpp
```diff
@@ -23,7 +23,7 @@
 #include "diagnostics/zonal_vapor_flux.hpp"
 #include "diagnostics/meridional_vapor_flux.hpp"
 #include "diagnostics/ice_cloud_mask.hpp"
-#include "diagnostics/vertically_remapped_field.hpp"
+//#include "diagnostics/vertically_remapped_field.hpp"
```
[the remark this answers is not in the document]
bartgol: Diagnostics are mostly for the sake of I/O, I think, right? So at this point, I would vote to remove it, yes.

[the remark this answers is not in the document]
AaronDonahue: 👍

bartgol (approved):

PR #2107, opened 2022-12-17
Do less rad work

brhillman:
Do less rad work on timesteps where rad heating is not updated. This bypasses the entire chunk loop and all rad calls (like cloud optics and gas concentrations) when fluxes and heating rates are not updated. Previously there was still a non-trivial amount of data wrangling and computation on each timestep.


ambrad (approved):

ambrad:
@brhillman any reason not to merge this today and get some nightly testing on it this weekend?

brhillman:
@ambrad nope, should be good to go I’ll add automerge.

PR #2109, opened 2022-12-20
Add a vertical remapper to EAMxx IO

AaronDonahue:
This commit introduces a vertical remapper to `scorpio_output.cpp/hpp` similar to the current coarsening remapper.

The purpose of the new remapper is to allow users to output simulation data on a different vertical mesh.  This is controlled by passing a vertical_remap file within the output YAML file, which triggers vertical remapping.

There is support to also output fields with no vertical extent, in that case the vertical remapping step will be skipped and the regular field will be in the output.

Note, with this commit there is not yet support for both vertical and horizontal remapping.  That feature will be added in a subsequent PR.

### DIFF
--- a/components/eamxx/src/share/grid/grids_manager.hpp
+++ b/components/eamxx/src/share/grid/grids_manager.hpp
@@ -51,6 +51,12 @@ public:
 
   const grid_repo_type& get_repo () const { return m_grids; }
 
+  void print() const {
+    auto avail_grids = print_available_grids();
+    printf("List of available grids in grids manager:\n");
+    printf("   - %s\n",avail_grids.c_str());
+  }
+
 protected:
   using nonconstgrid_ptr_type     = std::shared_ptr<grid_type>;
   using nonconstgrid_repo_type    = std::map<std::string, nonconstgrid_ptr_type>;
--- a/components/eamxx/src/share/grid/grids_manager.hpp
+++ b/components/eamxx/src/share/grid/grids_manager.hpp
@@ -51,12 +51,6 @@ public:
 
   const grid_repo_type& get_repo () const { return m_grids; }
 
-  void print() const {
-    auto avail_grids = print_available_grids();
-    printf("List of available grids in grids manager:\n");
-    printf("   - %s\n",avail_grids.c_str());
-  }
-
 protected:
   using nonconstgrid_ptr_type     = std::shared_ptr<grid_type>;
   using nonconstgrid_repo_type    = std::map<std::string, nonconstgrid_ptr_type>;
--- a/components/eamxx/src/share/grid/remap/coarsening_remapper.cpp
+++ b/components/eamxx/src/share/grid/remap/coarsening_remapper.cpp
@@ -814,7 +814,7 @@ void CoarseningRemapper::create_ov_tgt_fields ()
   m_ov_tgt_fields.reserve(m_num_fields);
   const int num_ov_cols = m_ov_tgt_grid->get_num_local_dofs();
   const auto ov_gn = m_ov_tgt_grid->name();
-  // for (const auto& f : m_tgt_fields) {
+  // for (const auto& f : m_tgt_fields) { //TODO should we delete this line?
   for (int i=0; i<m_num_fields; ++i) {
     const auto& f_src = m_src_fields[i];
     const auto& f_tgt = m_tgt_fields[i];
--- a/components/eamxx/src/share/grid/remap/coarsening_remapper.cpp
+++ b/components/eamxx/src/share/grid/remap/coarsening_remapper.cpp
@@ -814,7 +814,6 @@ void CoarseningRemapper::create_ov_tgt_fields ()
   m_ov_tgt_fields.reserve(m_num_fields);
   const int num_ov_cols = m_ov_tgt_grid->get_num_local_dofs();
   const auto ov_gn = m_ov_tgt_grid->name();
-  // for (const auto& f : m_tgt_fields) { //TODO should we delete this line?
   for (int i=0; i<m_num_fields; ++i) {
     const auto& f_src = m_src_fields[i];
     const auto& f_tgt = m_tgt_fields[i];
--- a/components/eamxx/src/share/grid/remap/vertical_remapper.cpp
+++ b/components/eamxx/src/share/grid/remap/vertical_remapper.cpp
@@ -51,7 +51,6 @@ VerticalRemapper (const grid_ptr_type& src_grid,
   // the correct number of level
   auto tgt_grid = src_grid->clone("Point Grid",false);  //<-- ?? For Luca, why does the 1st arg have to be PointGrid?  Shouldn't the tgt_grid be of the same type as the src_grid if being cloned?  Note, choosing any other string threw an error for an "unsupported grid".
   tgt_grid->reset_num_vertical_lev(m_num_remap_levs);
-  printf("ASD - tgt_grid num of levs = %d\n",tgt_grid->get_num_vertical_levels());
   this->set_grids(src_grid,tgt_grid);
 
   // Set the LEV and ILEV vertical profiles for interpolation from
@@ -144,6 +143,10 @@ create_tgt_layout (const FieldLayout& src_layout) const
 
 void VerticalRemapper::
 set_pressure_levels(const std::string& map_file) {
+  // TODO: I have concerns here with packsize.  I am using SCREAM_PACK_SIZE, which appears to work
+  //       but maybe it is possible for different variables to have different packsizes, and if so
+  //       then we can't hardcode the packsize for the target_levels.  Can we define as REAL and
+  //       still get_view using a different packsize?  Or some other strategy?
   scorpio::register_file(map_file,scorpio::FileMode::Read);
   m_num_remap_levs = scorpio::get_dimlen_c2f(map_file.c_str(),"nlevs");
   std::vector<Real> remap_pres_levs(m_num_remap_levs);
@@ -210,14 +213,16 @@ do_register_field (const identifier_type& src, const identifier_type& tgt)
 void VerticalRemapper::
 do_bind_field (const int ifield, const field_type& src, const field_type& tgt)
 {
+  auto name = src.name();
+  EKAT_REQUIRE_MSG(src.get_header().get_identifier().get_layout()==src.get_header().get_identifier().get_layout(),"ERROR! vert_remap:do_bind_field:" + name + ", tgt and src do not have the same layout");
   EKAT_REQUIRE_MSG (
       src.get_header().get_identifier().get_layout().rank()>1 ||
       src.get_header().get_alloc_properties().get_padding()==0,
-      "Error! We don't support 2d scalar fields that are padded.\n");
+      "Error! vert_remap:do_bind_field:check_src:" + name + ", We don't support 2d scalar fields that are padded.\n");
   EKAT_REQUIRE_MSG (
       tgt.get_header().get_identifier().get_layout().rank()>1 ||
       tgt.get_header().get_alloc_properties().get_padding()==0,
-      "Error! We don't support 2d scalar fields that are padded.\n");
+      "Error! vert_remap:do_bind_field:check_tgt:" + name + ", We don't support 2d scalar fields that are padded.\n");
   m_src_fields[ifield] = src;
   m_tgt_fields[ifield] = tgt;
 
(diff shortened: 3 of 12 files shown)


AaronDonahue:
Note: This PR is dependent on #2104 , so many of the changes shown here reflect the changes introduced in that PR.  When 2104 is merged I will rebase off of master and that will simplify the changes shown in this PR.

components/eamxx/src/share/io/scorpio_output.cpp
```diff
@@ -196,20 +254,36 @@ void AtmosphereOutput::run (const std::string& filename, const bool is_write_ste
   }
 
   // If needed, remap fields from their grid to the unique grid, for I/O
-  if (m_remapper) {
-    start_timer("EAMxx::IO::remap");
-    m_remapper->remap(true);
+  if (m_vert_remapper) {
```
AaronDonahue, line 257: Would it be better/cleaner to add a `map` to scorpio_output.cpp that stores all remappers in a specific order and calls them in a loop?  I ask because this chunk of code matches the horiz_remapper one below, just with a different remapper.

bartgol: You can turn the body of these if statements into a lambda, and call the lambda passing the two remappers. I don't think we need a map for remappers.

AaronDonahue: oh good idea.

components/eamxx/src/share/grid/remap/coarsening_remapper.cpp
```diff
@@ -814,7 +814,7 @@ void CoarseningRemapper::create_ov_tgt_fields ()
   m_ov_tgt_fields.reserve(m_num_fields);
   const int num_ov_cols = m_ov_tgt_grid->get_num_local_dofs();
   const auto ov_gn = m_ov_tgt_grid->name();
-  // for (const auto& f : m_tgt_fields) {
+  // for (const auto& f : m_tgt_fields) { //TODO should we delete this line?
```
bartgol, line 817: Yes.

components/eamxx/src/share/grid/remap/vertical_remapper.cpp
```diff
@@ -0,0 +1,286 @@
...
+
+  // This is a vertical remapper. We only go in one direction
+  m_bwd_allowed = false;
+
+  // Gather the pressure level data for vertical remapping
+  set_pressure_levels(map_file);
+
+  // Create tgt_grid that is a clone of the src grid but with
+  // the correct number of level
+  auto tgt_grid = src_grid->clone("Point Grid",false);  //<-- ?? For Luca, why does the 1st arg have to be PointGrid?  Shouldn't the tgt_grid be of the same type as the src_grid if being cloned?  Note, choosing any other string threw an error for an "unsupported grid".
```
bartgol, line 52: The first arg is just a name to give to this grid. It should not throw. You should run through gdb, and figure out where it was throwing.

components/eamxx/src/share/grid/remap/vertical_remapper.cpp
```diff
@@ -0,0 +1,286 @@
...
+  tgt_grid->reset_num_vertical_lev(m_num_remap_levs);
+  this->set_grids(src_grid,tgt_grid);
+
+  // Set the LEV and ILEV vertical profiles for interpolation from
+  register_vertical_source_field(lev_prof,"mid");
+  register_vertical_source_field(ilev_prof,"int");
+}
+
+VerticalRemapper::
```
bartgol, line 61: If the default destructor is fine, you should just do
```
~VerticalRemapper () = default;
```
in the header file. It makes your intent clear to whomever reads the header.

components/eamxx/src/share/grid/remap/vertical_remapper.hpp
```diff
@@ -0,0 +1,137 @@
+#ifndef EAMXX_VERTICAL_REMAPPER_HPP
+#define EAMXX_VERTICAL_REMAPPER_HPP
+
+#include "share/field/field_tag.hpp"
+#include "share/grid/remap/abstract_remapper.hpp"
+#include "share/util/scream_vertical_interpolation.hpp"
+
+#include "scream_config.h"
+
+#include <mpi.h>
```
bartgol, line 10: I don't think you need mpi in this header.

components/eamxx/src/share/grid/remap/vertical_remapper.hpp
```diff
@@ -0,0 +1,137 @@
+#ifndef EAMXX_VERTICAL_REMAPPER_HPP
+#define EAMXX_VERTICAL_REMAPPER_HPP
+
+#include "share/field/field_tag.hpp"
+#include "share/grid/remap/abstract_remapper.hpp"
+#include "share/util/scream_vertical_interpolation.hpp"
```
bartgol, line 6: You don't need this in this header, so include it in the cpp file.

components/eamxx/src/share/grid/remap/vertical_remapper.hpp
```diff
@@ -0,0 +1,137 @@
...
+
+#include "scream_config.h"
```
bartgol, line 8: Also not needed here.

components/eamxx/src/share/io/scorpio_output.cpp
```diff
@@ -331,7 +400,7 @@ void AtmosphereOutput::run (const std::string& filename, const bool is_write_ste
 long long AtmosphereOutput::
 res_dep_memory_footprint () const {
   long long rdmf = 0;
-  if (m_remapper) {
+  if (m_horiz_remapper) {
```
bartgol, line 403: The vertical remapper also contributes to memory footprint, no? Perhaps we could just do
```
if (m_io_field_mgr != m_sim_field_mgr)
```

components/eamxx/src/share/io/scorpio_output.hpp
```diff
@@ -11,7 +11,10 @@
 
 #include "ekat/ekat_parameter_list.hpp"
 #include "ekat/mpi/ekat_comm.hpp"
+#include "ekat/ekat_pack_utils.hpp"
+#include "ekat/ekat_pack_kokkos.hpp"
 
+#include <numeric>
```
bartgol, line 17: I am not sure these three new headers are used in this file. If you need them in the cpp file, you can include them there. But I don't think you need them there either...

components/eamxx/src/share/grid/remap/vertical_remapper.cpp
```diff
@@ -0,0 +1,286 @@
...
+  if (mode=="mid") {
+    auto layout = src_fid.get_layout();
+    auto name   = src_fid.name();
+    EKAT_REQUIRE_MSG(ekat::contains(std::vector<FieldTag>{LEV},layout.tags().back()),
+      "Error::VerticalRemapper::register_vertical_source_field,\n"
+      "mode = 'mid' expects a layour ending with LEV tag.\n"
+      " - field name  : " + name + "\n"
+      " - field layout: " + to_string(layout) + "\n");
+    EKAT_REQUIRE_MSG(src.is_allocated(), "Error! LEV source field is not yet allocated.\n");
+    src_mid = src;
```
bartgol, line 184: To follow other member vars, you could prepend `m_` to member var names. It helps to distinguish immediately if a var is local or a member, when looking at a code snippet.

components/eamxx/src/share/grid/remap/vertical_remapper.cpp
```diff
@@ -0,0 +1,286 @@
...
+  m_src_fields.push_back(field_type(src));
+  field_type tgt_f(tgt);
+  m_tgt_fields.push_back(tgt_f);
+}
+
+void VerticalRemapper::
+do_bind_field (const int ifield, const field_type& src, const field_type& tgt)
+{
+  auto name = src.name();
+  EKAT_REQUIRE_MSG(src.get_header().get_identifier().get_layout()==src.get_header().get_identifier().get_layout(),"ERROR! vert_remap:do_bind_field:" + name + ", tgt and src do not have the same layout");
```
bartgol, line 212: I think one of the two was supposed to be `tgt` here.

components/eamxx/src/share/grid/remap/vertical_remapper.cpp
```diff
@@ -0,0 +1,286 @@
...
+  register_vertical_source_field(ilev_prof,"int");
+}
+
+VerticalRemapper::
+~VerticalRemapper ()
+{
+  // Nothing to do
+}
+
+void VerticalRemapper::
```
bartgol, line 67: I guess you used this method while debugging, but it's no longer used. I would prune it completely then.

components/eamxx/src/share/grid/remap/vertical_remapper.cpp
```diff
@@ -0,0 +1,286 @@
...
+      break;
+    default:
+      EKAT_ERROR_MSG ("Layout not supported by VerticalRemapper: " + e2str(lt) + "\n");
+  }
+  return tgt;
+}
+
+void VerticalRemapper::
+set_pressure_levels(const std::string& map_file) {
+  // TODO: I have concerns here with packsize.  I am using SCREAM_PACK_SIZE, which appears to work
```
bartgol, line 146: You need to pick a packsize that you're ok with. Then, during the do_remap method, you should check to see if the src/tgt fields are compatible with your pack size. That's what we do in the PD remapper. Something like

```
  if (src and tgt compatible with packsize=N1)
     call_my_remap<N1>(...);
   else if (src and tgt compatible with packsize N2)
     call_my_remap<N2>(...);
   ...
   else
     call_my_remap<1>();
```
You decide how many pack sizes you want to support. The easiest choice is to support N=SCREAM_PACK_SIZE and N=1.

components/eamxx/src/share/grid/grids_manager.hpp
```diff
@@ -51,6 +51,12 @@ class GridsManager
 
   const grid_repo_type& get_repo () const { return m_grids; }
 
+  void print() const {
```
bartgol, line 54: This seems like a method you used while debugging, but no longer needed/used. Let's prune it.

bartgol (changes requested):
There's one bug to fix (src vs tgt typo). I also have a few suggestions, you decide which one is important.

components/eamxx/src/share/grid/remap/vertical_remapper.cpp
AaronDonahue: It appears to be a legit error thrown by a check in the grids manager code.  As far as I can tell I would need the grids manager to `add_grid` the grid from the vertical remapper in order to not throw this error.
```
  /g/g17/donahue5/Code/e3sm/scream/scream_sandbox/components/eamxx/src/share/
  grid/grids_manager.cpp:34: FAIL:
  has_grid(to_grid->name())
  Error! Target grid 'P Grid' is not supported.
```

Note, this error is thrown during a call to `create_remapper` in the grids manager.

components/eamxx/src/share/grid/remap/vertical_remapper.cpp
bartgol: It looks like you are only using SCREAM_PACK_SIZE. Are you ok not supporting any other pack size (or even packsize=1, i.e., no packing)?

bartgol (commented):
It would be nice to add some unit tests for this remapper. I am assuming you will soon incorporate it in the IO classes, and test it there, but when bugs show up (as they always do), it is helpful to have a test for each individual component..

AaronDonahue:
> It would be nice to add some unit tests for this remapper. I am assuming you will soon incorporate it in the IO classes, and test it there, but when bugs show up (as they always do), it is helpful to have a test for each individual component..

The newest commit adds a suite of unit tests similar to the coarsening remapper tests.

components/eamxx/src/share/util/scream_vertical_interpolation_impl.hpp
```diff
@@ -46,8 +46,8 @@ void apply_interpolation_impl_1d(
   auto mask   = Kokkos::subview(mask_col,  Kokkos::pair<int,int>(0,num_tgt_packs));
 
   // The input/output data and x_src/x_tgt data should match in the appropriate size, respectively.
-  EKAT_REQUIRE_MSG(x_tgt.size() == output.size(),"Error! vertical_interpolation::perform_checks " + std::to_string(x_tgt.size()) + " != " + std::to_string(output.size()) + ".");
-  EKAT_REQUIRE_MSG(x_src.size() == input.size(), "Error! vertical_interpolation::perform_checks " + std::to_string(x_src.size()) + " != " + std::to_string(input.size()) + ".");
+  EKAT_KERNEL_REQUIRE_MSG(x_tgt.size() == output.size(),"Error! vertical_interpolation::perform_checks " + std::to_string(x_tgt.size()) + " != " + std::to_string(output.size()) + ".");
```
bartgol, line 49: I don't think this will work. The KERNEL assertions require C style strings, not std::string. Given that you can't allocate a C string here to create the message (e.g., via sprintf), your only option is to not print any info regarding the data, just a plain error message string.

AaronDonahue: Huh, that's probably why the fail in the AT.  Interestingly I kept that because it was working on my weaver build.  Oh well.  I'll go ahead and take away the message, it isn't strictly necessary anyways.

components/eamxx/src/share/util/scream_vertical_interpolation_impl.hpp
```diff
@@ -46,8 +46,8 @@ void apply_interpolation_impl_1d(
   auto mask   = Kokkos::subview(mask_col,  Kokkos::pair<int,int>(0,num_tgt_packs));
 
   // The input/output data and x_src/x_tgt data should match in the appropriate size, respectively.
-  EKAT_KERNEL_REQUIRE_MSG(x_tgt.size() == output.size(),"Error! vertical_interpolation::perform_checks " + std::to_string(x_tgt.size()) + " != " + std::to_string(output.size()) + ".");
-  EKAT_KERNEL_REQUIRE_MSG(x_src.size() == input.size(), "Error! vertical_interpolation::perform_checks " + std::to_string(x_src.size()) + " != " + std::to_string(input.size()) + ".");
+  EKAT_KERNEL_REQUIRE(x_tgt.size() == output.size());
```
bartgol, line 49: This is fine, but notice that you _could_ use `EKAT_KERNEL_REQUIRE_MSG` passing a message that gives the user a bit more context. The catch is that it must be a single "blah" C-style string...

components/eamxx/src/share/tests/vertical_remapper_tests.cpp
```diff
@@ -99,7 +97,7 @@ void create_remap_file(const std::string& filename, const int nlevs, const std::
 
   scorpio::register_file(filename, scorpio::FileMode::Write);
 
-  scorpio::register_dimension(filename,"nlevs","nlevs",nlevs);
+  scorpio::register_dimension(filename,"nlevs","nlevs",nlevs, true);
```
bartgol, line 100: Are you sure nlevs should be marked as a partitioned dimension?

AaronDonahue: Oh probably not

bartgol (approved):

PR #2115, opened 2023-01-03
Crusher: Use -S 0 to solve -c 8 issue.

### LINKED ISSUE #2112: Crusher: "srun: error: Unable to create step for job ...: More processors requested than permitted"
At first I thought this was associated with node count, but it's not. It seems that 8 tasks by 8 threads, 1 thread/core, stopped working on 29 Dec. I've tried various things of this sort to no avail:
```
export SLURM_JOB_CPUS_PER_NODE=64
srun  -l -K -n 8 -N 1  --gpus-per-node=8 --gpu-bind=closest -c 8 --cores-per-socket=64 --sockets-per-node=1 --ntasks-per-core=1 ...
```
`-c 7` works, however. I'll use this to work around the problem, but I probably need to email with some systems folks to get to the bottom of this.

ambrad:
Also move crusher-related entries in XML file to OLCF area and revert the PE-layout workaround.

Fixes #2112.

### DIFF
--- a/cime_config/allactive/config_pesall.xml
+++ b/cime_config/allactive/config_pesall.xml
@@ -2275,18 +2275,18 @@
       <pes compset="any" pesize="any">
         <comment>"crusher-gpu ne30np4 and ne30np4.pg2"</comment>
         <ntasks>
-          <ntasks_atm>-1</ntasks_atm>
-          <ntasks_lnd>-1</ntasks_lnd>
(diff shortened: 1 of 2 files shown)


ambrad:
ERS_Ln362.ne30pg2_ne30pg2.F2010-SCREAMv1.crusher-gpu_crayclang passes, and the PE layout is back to what we want.

Issue #2114, opened 2023-01-02
Pre/post-condition output for bounds checking prints the column associated with the wrong bound.


ambrad:
In this chunk of code:
https://github.com/E3SM-Project/scream/blob/f63890fbb7ce70374568cb5f4366ff11dd3069c9/components/eamxx/src/share/property_checks/field_within_interval_check.cpp#L201-L223
`pass_lower` and `pass_upper` should be initialized to `true` rather than `false`. Currently, if `pass_upper` is truly `false` and `pass_lower` _should_ be `true`, `pass_lower` ends up being `false`, and so `res_and_msg` ends up recording data associated with the lower bound. Then the nifty new column output is for the wrong column.

ambrad:
Probably should add a unit test for this case, showing first that it fails and then that the fix makes it pass.

PR #2116, opened 2023-01-04
Aarondonahue/fix bug in prepost cond output

AaronDonahue:
Fix bug in the within_interval_check for property checks where if the maximum was violated the failed column
index still took the location of the minimum value.

This commit also adds a unit test to ensure that the column index matches the expectation.

Addresses #2114

### DIFF
--- a/components/eamxx/src/share/tests/property_checks.cpp
+++ b/components/eamxx/src/share/tests/property_checks.cpp
@@ -205,6 +205,27 @@ TEST_CASE("property_checks", "") {
     interval_check->repair();
     res_and_msg = interval_check->check();
     REQUIRE(res_and_msg.result==CheckResult::Pass);
+
+    // Re-assign an out-of-bounds value to the field, but only in the upper-bound.  Check if it
+    // fails and reports just the max fail.
+    std::vector<int> exp_fail_loc = {0,0,1};
+    f_data[1] = -2;
+    f.sync_to_dev();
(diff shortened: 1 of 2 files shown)


AaronDonahue:
@ambrad , thanks for catching this bug.  Since @bartgol is on vacation I decided to put in the fix.

components/eamxx/src/share/tests/property_checks.cpp
```diff
@@ -205,6 +205,27 @@ TEST_CASE("property_checks", "") {
     interval_check->repair();
     res_and_msg = interval_check->check();
     REQUIRE(res_and_msg.result==CheckResult::Pass);
+
+    // Re-assign an out-of-bounds value to the field, but only in the upper-bound.  Check if it
```
ambrad, line 209: Should "upper" here be "lower" in the comment?

ambrad (dismissed):

components/eamxx/src/share/tests/property_checks.cpp
AaronDonahue: oh yes, copy/paste error.

AaronDonahue:
@ambrad , I fixed the typo in the comments so this will probably need to be re-approved.

ambrad:
@AaronDonahue I think there's something up with the AT or Weaver. If you've tested this PR reasonably, I suggest you merge it and let the nightlies take care of the rest.

Issue #2014, opened 2022-11-01
Units of __WaterPath and ____VapFlux appear incorrect


crterai:
The current units of `VapWaterPath`, `LiqWaterPath`, and other vertically integrated water paths in v1 are set to `m`, but looking at the values, it looks like they should be `mm`  (or kg m**-2) rather than `m`. 
```
	float LiqWaterPath(time, ncol) ;
		LiqWaterPath:units = "m" ;
		LiqWaterPath:long_name = "LiqWaterPath" ;
	float IceWaterPath(time, ncol) ;
		IceWaterPath:units = "m" ;
		IceWaterPath:long_name = "IceWaterPath" ;
	float RainWaterPath(time, ncol) ;
		RainWaterPath:units = "m" ;
		RainWaterPath:long_name = "RainWaterPath" ;
	float RimeWaterPath(time, ncol) ;
		RimeWaterPath:units = "m" ;
		RimeWaterPath:long_name = "RimeWaterPath" ;
	float VapWaterPath(time, ncol) ;
		VapWaterPath:units = "m" ;
		VapWaterPath:long_name = "VapWaterPath" ;
```
ncvis of the ne1024 data:
<img width="923" alt="image" src="https://user-images.githubusercontent.com/12244479/199360292-f658dccd-89c5-4362-8142-68a24bea52fe.png">

crterai:
The units of the following variables also should be updated: `ZonalVapFlux` and `MeridionalVapFlux`. 
Instead of `m`, I think they should be `kg m**-1 s**-1`.

crterai:
I'd forgotten to follow up about this issue. @AaronDonahue or @hassanbeydoun, would you be able to make this fix? I tried to look through the code, but couldn't figure out where the units are specified.

AaronDonahue:
Hi @crterai , I'll put in a quick fix

crterai:
Thanks, @AaronDonahue !

ambrad:
Fixed by #2119.

PR #2119, opened 2023-01-04
This commit will fix the units for water path diagnostics.

AaronDonahue:
The units for `ICE` `LIQ` `RAIN` `RIME` and `VAP` water path were incorrect.  This commit fixes that.

Additionally, the flux for `meridional` and `zonal` vapor flux were incorrect and have been fixed.

Addresses #2014


crterai (approved):
Looks good to me!

PR #2129, opened 2023-01-17
[wip] bring in changes for crusher machine

oksanaguba:


cime_config/machines/cmake_macros/crayclanggpu_crusher.cmake
```diff
@@ -0,0 +1,52 @@
...
+if (NOT DEBUG)
+  string(APPEND CFLAGS   " -O2")
+  string(APPEND CXXFLAGS " -O2")
+  string(APPEND FFLAGS   " -O2")
+endif()
+
+if (COMP_NAME STREQUAL elm)
+  string(APPEND FFLAGS " -hfp0")
+endif()
+string(APPEND FFLAGS " -hipa0 -hzero -em -ef -hnoacc")
```
ambrad, line 40: Note we don't want to use `-hzero`, as that zeros all stack-allocated arrays. This can lead to uncaught uninitialized-variable bugs.

oksanaguba: yes, i am consolidating/comparing changes from Az and what scream uses now. hzero is one of many questions.

oksanaguba: tbc the 1st 2 commits were to bring Az's changes into this branch blindly. the last commit is where i tried to compare everything in his 3 files and in scream's gpu file and put comments.

oksanaguba: that is, no one is bringing these changes to scream as they are, yet

oksanaguba:
While i am still cleaning and testing this, what would be good tests to run on crusher for this PR?

ambrad:
Run these two groups: e3sm_scream_v1_long_crusher and e3sm_scream_v1. These run nightly on Crusher; you can see results on cdash.

oksanaguba:
The purpose of this PR was to have the same machine+compiler pair for scream and for e3sm to ease upstream merge process. This is actually ready, except that i could not keep some of changes in Az's PR https://github.com/E3SM-Project/E3SM/pull/5229 . Some of the changes i did not keep are these and similar:
```
set(MPICC "cc")
set(MPICXX "CC")
set(MPIFC "ftn")
set(SCC "cc")
set(SCXX "CC")
set(SFC "ftn")
```
and i used `-h omp` instead of `-fopenmp` (i did not really investigate omp scaling with either flag, but from googling, I think cray takes in -homp).
For compilers, i had to comment those out because otherwise pio is being compiled with hipcc, which did not accept -homp flag. There were other changes that i made while debugging tests, preserving scream setup as much as possible.

**However, following the discussion from today's scream telecon, i am not sure if this PR should be closed without merging. The current setup in scream is to use crayclang compiler and crusher-gpu, crusher-gpu machines. Thus e3sm and scream share only crayclang.cmake file in such setup,  and the compiler in scream setup can be renamed to avoid any intersection. Another issue is config_pesall file, which needs a patch. So I assume this PR should be abandoned?**

oksanaguba:
tagging @ambrad, please, see my comment above

oksanaguba:
closing till there is better understanding how merges to and from e3sm repo are handled and how scream is run.

Issue #2120, opened 2023-01-05
Create rough topography compset for v1


tcclevenger:
Create ne30 rough topography compset using USGS-gtopo30_ne30np4pg2_x6t-SGH.nc en route to switching to rough topo for ne1024.

tcclevenger:
@mt5555 On this older issue https://github.com/E3SM-Project/scream/issues/903 about running scream high-res with rough topo, you recommended for a rough topo run:

- pgrad_correction=1
- hv_ref_profiles=0
- hv_theta_correction=0

I think this has changed to using `hv_ref_profiles=6`. I see that `hv_ref_profile=0` corresponds to only keeping `phi_ref`. What does `hv_ref_profiles=6` mean? And is it pointless to try and run until this option is available, or is `hv_ref_profiles=0` good enough for now?

Then for `hv_theta_correction=0`, I'm not seeing this used or set in any default namelist files in EAMxx. Is this defaulted to 0? Is 0 still recomended?

Are there any other non-default namelist variables that we need?

mt5555:
hv_ref_profiles=0 is good enough for now.  

the hv_ref_profiles=6 option is on a branch - I'll merge that into HOMME after your hv_ref_profiles sync branch goes in.  (or maybe I could add that code to your branch to minimize PRs)

tcclevenger:
@mt5555 Feel free to add it to my branch. There is still some open questions about that PR an option to not change theta-l code, but that should be resolved soon.

PR #2130, opened 2023-01-19
Add rough topo compset and test for ne30

tcclevenger:
What this PR currently does:
- Adds rough topo compset for ne30 screamv1, with `USGS-gtopo30_ne30np4pg2_x6t-SGH.nc` topo file and `pgrad_correction=1` set in the input yaml.
- Adds midres test `ERS_Ln22.ne30pg2_ne30pg2.F2010-SCREAMv1-X6T`.
- Fixes ne1024 namelist defaults, so that `*X6T_ELM*` compsets get `USGS-gtopo30_ne1024np4pg2_x6t-SGH.nc` topo file and `pgrad_correction=1`.

What I've verified is that `ERS_Ln22.ne30pg2_ne30pg2.F2010-SCREAMv1-X6T` runs to completion and includes the correct topo file in the input yaml and `pgrad_correction=1` in the namelist file.

Should anything else be tested? Or output files be looked at?

Closes https://github.com/E3SM-Project/scream/issues/2120

### DIFF
--- a/components/eamxx/cime_config/config_compsets.xml
+++ b/components/eamxx/cime_config/config_compsets.xml
@@ -81,6 +81,13 @@
       <support_level>Experimental, under development</support_level>
   </compset>
 
+  <compset>
+      <alias>F2010-SCREAMv1-X6T</alias> <!-- Coupled land-atm compset (using SCREAMv1), with X6T topography-->
+      <lname>2010_SCREAM%X6T_ELM%SPBC_CICE%PRES_DOCN%DOM_SROF_SGLC_SWAV_SIAC_SESP</lname>
+      <support_level>Experimental, under development</support_level>
+  </compset>
+
+
   <entries>
 
     <entry id="RUN_STARTDATE">
--- a/cime_config/tests.py
+++ b/cime_config/tests.py
@@ -486,6 +486,7 @@ _TESTS = {
             "PEM_Ln90.ne30pg2_ne30pg2.F2010-SCREAMv1",
             "ERS_Ln22.ne30pg2_ne30pg2.F2010-SCREAMv1",
             "ERS_Ln90.ne30pg2_ne30pg2.F2010-SCREAMv1.scream-small_kernels",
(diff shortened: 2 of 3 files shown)


components/eamxx/cime_config/config_compsets.xml
```diff
@@ -81,6 +81,13 @@
       <support_level>Experimental, under development</support_level>
   </compset>
 
+  <compset>
+      <alias>F2010-SCREAMv1-X6T</alias> <!-- Coupled land-atm compset (using SCREAMv1), with X6T topography-->
+      <lname>2010_SCREAM%X6T_ELM%SPBC_CICE%PRES_DOCN%DOM_SROF_SGLC_SWAV_SIAC_SESP</lname>
```
tcclevenger, line 86: I saw a lot of repetitive keys here from the screamv1, so I copied. Is this correct to do?

PeterCaldwell:
"Should anything else be tested? Or output files be looked at?"

I know this PR is about ne30, but is there anything that needs to be done for us to use rough topo at ne1024pg2? That's our ultimate goal...

tcclevenger:
@PeterCaldwell I updated the `namelist_defaults_scream.xml` so that the ne1024 compsets which include "X6T" set `pgrad_correction=1` and include the ne1024np4pg2-x6t topo files. I've not attempted to actually run this, but it _should_ run.

Also, soon the `hv_ref_profiles=6` should be merged soon into Homme, but I'm guessing next week at the earliest. Not sure how important that is and if it matters more for ne1024?

PeterCaldwell:
Ok. Let's discuss whether anything still needs to be done to switch ne1024 to rough topography during the v1 call today.

ambrad (changes requested):
I'm fine merging this as is, except for oen small change. You can follow up later with changes.

Since we already have multiple ne30pg2 tests with smooth topo and I recommend against adding another build to our nightlies, I suggest removing ERS_Ln22.ne30pg2_ne30pg2.F2010-SCREAMv1 or, in other words, just adding -X6T to it and not adding a new test.

ambrad (approved):

PR #2135, opened 2023-01-25
Add vertical + horizontal remapping as an option for output

AaronDonahue:
This commit is a follow up to #2109.  It unlocks the capability to designate both vertical and horizontal remapping for a single output stream.  

In order to designate that an output stream should do both we just combine the syntax for each approach in the YAML file that controls output.  So if a control file has both `horiz_remap_file` and `vertical_remap_file` listed than both of these remap steps will be taken.

Note, the order is hard-coded to be vertical first, horizontal next. 

Furthermore, if both are declared than the horizontal remapper will account for any masking in the vertical remapping.  Note that vertical remapping could result in target pressure levels that are outside the bounds of the simulation pressure profile.  When this happens the vertical remapper will assign a masking array to the field that can be used by the horizontal remapper.  Masking values by default are a large negative number, -999999.  If we didn't account for masking in the horizontal remapper we would get some values that are incorrectly skewed towards -999999.  Furthermore, the when a masked value is detected, not only is it not included in the horizontal remapping, but the overall weights used for that column are rescaled such that the total sum of weights used is still 1.0.  This avoids unrealistic values that are skewed towards 0.0 when masking is present.

This commit also introduces an IO test suite to test vertical remapping, horizontal remapping and vertical+horizontal remapping when used as output.


AaronDonahue:
WIP pending #2109

AaronDonahue:
@bartgol , this PR is ready for review when you have a chance

components/eamxx/src/share/io/scorpio_output.cpp
```diff
@@ -473,16 +515,23 @@ void AtmosphereOutput::
 set_field_manager (const std::shared_ptr<const fm_type>& field_mgr, const std::string& mode)
 {
   // Sanity checks
-  EKAT_REQUIRE_MSG (mode=="io" || mode=="sim", "Error! Invalid arg to set_field_manager, must be 'io' or 'sim'");
+  EKAT_REQUIRE_MSG (mode =="all" || mode=="int" || mode=="io" || mode=="sim", "Error! Invalid arg to set_field_manager, must be 'io' or 'sim'");
```
bartgol, line 518: Suggestion: you start to have quite a few FM here, so maybe a `FMType` enum might be more appropriate than a string.

components/eamxx/src/share/io/scorpio_output.cpp
```diff
@@ -814,8 +863,8 @@ Field AtmosphereOutput::get_field(const std::string& name, const std::shared_ptr
   } else if (m_diagnostics.find(name) != m_diagnostics.end() && field_mgr==m_sim_field_mgr) {
     const auto& diag = m_diagnostics.at(name);
     return diag->get_diagnostic();
-  } else if (m_fields_alt_name.find(name) != m_fields_alt_name.end() && field_mgr->has_field(m_fields_alt_name.at(name))) {
-    return field_mgr->get_field(m_fields_alt_name.at(name));
+  } else if (m_fields_alt_name.find(name) != m_fields_alt_name.end()) {
+    return get_field(m_fields_alt_name.at(name),field_mgr);
   } else {
     EKAT_ERROR_MSG ("Field " + name + " not found in output field manager or diagnostics list, or requesting a diag not on the simualation field manager.");
```
bartgol, line 869: Suggestion: given that we have _several_ FM's in this class, we should print _which_ FM we were querying, cause I suspect it might help debugging what stage of the execution went wrong. We could pass the string with the FM type rather than the FM itself, and retrieve the FM inside the fcn, so that we can print the FM type in the message.

components/eamxx/src/share/io/scorpio_output.hpp
```diff
@@ -178,6 +178,7 @@ class AtmosphereOutput
   // sim_field_manager points to the simulation field manager
   // when remapping horizontally these two field managers may be different.
   std::shared_ptr<const fm_type>      m_io_field_mgr;
+  std::shared_ptr<const fm_type>      m_int_field_mgr;
```
bartgol, line 181: I'm guessing `int` stands for `intermediate`, but it's awfully reminding of the `int` data type. Perhaps we can spell "intermediate" out? Or something more descriptive, like `m_field_mgr_after_vinterp`?

bartgol (changes requested):
Lots of stuff! I have a few comments. Some are style, but I think some are changes we should do (e.g., views alloc on device at runtime).

components/eamxx/src/share/grid/remap/coarsening_remapper.cpp
```diff
@@ -291,6 +304,33 @@ do_bind_field (const int ifield, const field_type& src, const field_type& tgt)
 
 void CoarseningRemapper::do_registration_ends ()
 {
+  // Before finishing the registration we need to also register the set of masks, if needed.
+  if (m_track_mask) {
+    for (int idx=0; idx<m_mask_fields_src.size(); ++idx) {
+      auto f   = m_mask_fields_src[idx];
+      auto fid = f.get_header().get_identifier();
+      register_field_from_src(fid);
+      m_mask_map_tgt.emplace(idx,m_num_registered_fields-1);
```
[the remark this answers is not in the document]
AaronDonahue: doesn't `register_field_from_src` add 1 to the m_num_registered_fields tally?  I think this will always grab the last field in the map.  There may be a better way to do that though.

components/eamxx/src/share/grid/remap/coarsening_remapper.cpp
```diff
@@ -291,6 +304,33 @@ do_bind_field (const int ifield, const field_type& src, const field_type& tgt)
...
+    }
+    // Update the number of fields.
+    m_num_fields = m_num_registered_fields;
```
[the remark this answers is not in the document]
AaronDonahue: My understanding was that the number of registered fields is calculated first, before `do_registration_ends()` is called.  Since here we are actually adding fields we have to update the number.  See: https://github.com/E3SM-Project/scream/blob/master/components/eamxx/src/share/grid/remap/abstract_remapper.cpp#L97

Maybe the more correct thing would be to change `do_registration_ends()` in the abstract_remapper class?

components/eamxx/src/share/grid/remap/coarsening_remapper.cpp
```diff
@@ -347,11 +414,119 @@ void CoarseningRemapper::do_remap_fwd ()
...
+  using MemberType  = typename KT::MemberType;
+  using ESU         = ekat::ExeSpaceUtils<typename KT::ExeSpace>;
+  using Pack        = ekat::Pack<Real,PackSize>;
+  using PackInfo    = ekat::PackInfo<PackSize>;
+
+  const auto& layout = x.get_header().get_identifier().get_layout();
+  const int rank = layout.rank();
+  const int ncols = m_ov_tgt_grid->get_num_local_dofs();
+  const Real mask_val = -999999.0;   // TODO: Maybe not hard code what a fully masked value will be set too.
+  const Real mask_threshold = 1e-8;  // TODO: Should we not hardcode the threshold for simply masking out the column.
```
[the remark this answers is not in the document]
AaronDonahue: We certainly can set it to 0.  I don't have a reason to choose 1e-8, that's for sure.  The argument for caring about a threshold at all is to avoid say a single column in a map of 60 to dominate the whole column as though it shouldn't be masked.  For example, say column 1-60 map to column A in the coarse mesh.  If 2-60 are masked, and column 1 has a weight of 1e-12, we may want to see this and say that this is actually a masked column fully, throwing away the incredibly small 1e-12 value rather than scale it up to 1.0.

In a real life example, that would be like a columns 2-60 are inside a mountain, but column 1 is just on the edge, so not masked but barely adding anything.  Do we want the coarse mesh to take that value anyway?

The problem is setting the threshold though.  So hard-coding to 0.0 for now is probably the best choice and letting it be changed later  might be okay.

components/eamxx/src/share/grid/remap/coarsening_remapper.cpp
```diff
@@ -316,20 +358,45 @@ void CoarseningRemapper::do_remap_fwd ()
...
+      mask_idx = m_mask_map_src.at(f_src.name());
+    }
+
+    if (mask_idx != -1) {
+      // Pass the mask to the local_mat_vec routine
+      auto mask = m_mask_fields_src[mask_idx];
+      // Dispatch kernel with the largest possible pack size
+      const auto& src_ap = f_src.get_header().get_alloc_properties();
+      const auto& ov_tgt_ap = f_ov_tgt.get_header().get_alloc_properties();
+      if (can_pack && src_ap.is_compatible<RPack<16>>() &&
```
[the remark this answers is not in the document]
AaronDonahue: should we do the same for `loc_mat_vec` above?

components/eamxx/src/share/grid/remap/coarsening_remapper.cpp
```diff
@@ -388,6 +569,12 @@ local_mat_vec (const Field& x, const Field& y) const
...
       auto y_view = y.get_view<      Pack**>();
+      view_2d<Real> mask_view("",x_view.extent_int(0),x_view.extent_int(1));
+      if (mask != NULL) {
+        mask_view = mask->get_view<Real**>();
+      } else {
+        Kokkos::deep_copy(mask_view,1.0);
+      }
       const int dim1 = PackInfo::num_packs(src_layout.dim(1));
       auto policy = ESU::get_default_team_policy(nrows,dim1);
       Kokkos::parallel_for(policy,
```
AaronDonahue, line 580: Note to self, the mask_view doesn't appear to be used here.  Need to investigate.

components/eamxx/src/share/grid/remap/coarsening_remapper.cpp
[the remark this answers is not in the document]
bartgol: Ah, yes, you're right. Can you add a 5 words comment about that? I can see myself being confused by this in the future...

components/eamxx/src/share/grid/remap/coarsening_remapper.cpp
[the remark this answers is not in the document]
bartgol: I think so, yeah.

components/eamxx/src/share/grid/remap/coarsening_remapper.cpp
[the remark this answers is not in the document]
bartgol: I see your point. Though 1e-7 is prob also a similar scenario. Perhaps a reasonable rule of thumb could be 1/N, with being the number of nonzeros for that row (which is fixed, and same for all fields, so can be precomputed). It would mean that the non-masked cols add up to more than a single contribution of a uniform average of the patch.

But yeah, we can leave 0, and decide this factor later.

components/eamxx/src/share/grid/remap/coarsening_remapper.cpp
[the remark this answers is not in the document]
bartgol: Ah right, yes.

components/eamxx/src/share/grid/remap/vertical_remapper.cpp
```diff
@@ -212,9 +213,64 @@ do_bind_field (const int ifield, const field_type& src, const field_type& tgt)
...
+      m_tgt_masks.push_back(mask_tgt_fld);
+    }
+  } else {
+    // If a field does not have LEV or ILEV it may still have mask tracking assigned from somewhere else.
+    // In those cases we want to copy that mask tracking to the target field.
+    const auto src_extra = src.get_header().get_extra_data();
+    if (src_extra.count("mask_data")) {
+      auto& f_tgt = m_tgt_fields[ifield];
+      auto f_mask = ekat::any_cast<Field>(src_extra.at("mask_data"));
+      f_tgt.get_header().set_extra_data("mask_data",f_mask);
```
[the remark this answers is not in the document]
AaronDonahue: Good point, I think it should throw an error, we would want developers thinking they have assigned a specific mask to a field only to have the vertical remapper override it. At the very least they should be aware and could always change the name of their special mask to something other than "mask_data"

components/eamxx/src/share/grid/remap/vertical_remapper.cpp
```diff
@@ -272,21 +328,77 @@ void VerticalRemapper::do_remap_fwd ()
...
+                             src_pres_ap.is_compatible<RPack<2>>() &&
+                             tgt_pres_ap.is_compatible<RPack<2>>()) {
+        apply_vertical_interpolation<2>(f_src,f_tgt,true); 
+      } else {
+        apply_vertical_interpolation<1>(f_src,f_tgt,true); 
+      }
+    } else {
+      // There is nothing to do, this field cannot be vertically interpolated,
+      // so just copy it over.
+      f_tgt.deep_copy(f_src);
```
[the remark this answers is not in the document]
AaronDonahue: This is a great idea.  I do run into a problem in the simplest implementation.

We run into an issue in scorpio_output.cpp.  The new IO field manager is created by getting all target fields from the vertical remapper.  Those are created in the `register_field` step, and at that point we are dealing with field_id's not actual fields.  The point where we can go ahead and alias, when the fields are treated as fields, is in the `bind_fields` step.  But at that point the IO field manager is pointing to this empty field while the vertical remapper is just aliasing the source field in the simulation field manager.

components/eamxx/src/diagnostics/field_at_pressure_level.cpp
```diff
@@ -71,21 +82,36 @@ void FieldAtPressureLevel::compute_diagnostic_impl()
     auto fdat = view_Nd<mPack,2>("",f_data_src.extent_int(0),f_data_src.extent_int(1));
```
bartgol, line 82: Allocating memory on device at run time is not great. Can this helper view be created at init time? Perhaps when field_src is set? If not, you could have it as a member var, and do
```
if (m_fdat.size()==0) {
  m_dat = view_Nd<mPack,2>(...);
}
```
so that the allocation happens only once.

components/eamxx/src/share/grid/remap/vertical_remapper.cpp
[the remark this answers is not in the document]
AaronDonahue: ping

components/eamxx/src/share/grid/remap/vertical_remapper.cpp
```diff
@@ -51,10 +53,32 @@ VerticalRemapper (const grid_ptr_type& src_grid,
   scorpio::register_file(map_file,scorpio::FileMode::Read);
   m_num_remap_levs = scorpio::get_dimlen_c2f(map_file.c_str(),"nlevs");
   scorpio::eam_pio_closefile(map_file);
-  auto tgt_grid = src_grid->clone("Point Grid",false);  //<-- ?? For Luca, why does the 1st arg have to be PointGrid?  Shouldn't the tgt_grid be of the same type as the src_grid if being cloned?  Note, choosing any other string threw an error for an "unsupported grid".
-  tgt_grid->reset_num_vertical_lev(m_num_remap_levs);
+
+  auto tgt_grid_gids = src_grid->get_unique_gids();
```
bartgol, line 57: I would add a comment here, explaining that this new grid has same 2d gids as src grid, and only differs in the number of levels.

bartgol: Btw, you could just do
```
auto tgt_grid = src_grid->clone("vertical_remap_tgt_grid", /*shallow = */ false);
tgt_grid->reset_num_vertical_lev  (m_num_remap_levs);
this->set_grids(src_grid,tgt_grid);
```
which is prob a bit more self-documenting.

bartgol: Actually, you could do a shallow copy, since all the grid fields (geo data, gids, ...) are the same, and you only need to change nlevs. This would save us some storage (the hard copy of geo data and gids).

components/eamxx/src/share/grid/remap/vertical_remapper.cpp
```diff
@@ -51,10 +53,32 @@ VerticalRemapper (const grid_ptr_type& src_grid,
...
+  const int ngids = tgt_grid_gids.size();
+  auto tgt_grid = std::make_shared<PointGrid>("vertical_remap_tgt_grid",ngids,m_num_remap_levs,m_comm);
+  auto tgt_grid_gids_h = tgt_grid->get_dofs_gids().get_view<gid_t*,Host>();
+  std::memcpy(tgt_grid_gids_h.data(),tgt_grid_gids.data(),ngids*sizeof(gid_t));
+  tgt_grid->get_dofs_gids().sync_to_dev();
   this->set_grids(src_grid,tgt_grid);
 
+  // Replicate the src grid geo data in the tgt grid.
+  const auto& src_geo_data_names = src_grid->get_geometry_data_names();
```
bartgol, line 66: Again, if you use `clone`, with shallow=true, you don't need this.

components/eamxx/src/share/grid/remap/vertical_remapper.cpp
```diff
@@ -212,9 +236,80 @@ do_bind_field (const int ifield, const field_type& src, const field_type& tgt)
...
   m_src_fields[ifield] = src;
   m_tgt_fields[ifield] = tgt;
 
+  // Add mask tracking to the target field
+  using namespace ShortFieldTagsNames;
+  using namespace ekat::units;
+  auto nondim = Units::nondimensional();
+  if (src_layout.has_tag(LEV) || src_layout.has_tag(ILEV)) {
+    auto& f_tgt = m_tgt_fields[ifield];
+    // Use create_source_layout to make a copy of the source layout.
```
bartgol, line 249: Looks like this comment is outdated?

components/eamxx/src/share/grid/remap/vertical_remapper.cpp
```diff
@@ -243,29 +338,50 @@ void VerticalRemapper::do_remap_fwd ()
...
+  for (int i=0; i<m_tgt_masks.size(); ++i) {
+          auto& f_src    = m_src_masks[i];
+          auto& f_tgt    = m_tgt_masks[i];
+    const auto& layout   = f_src.get_header().get_identifier().get_layout();
+    const auto  src_tag  = layout.tags().back();
+    const bool  do_remap = ekat::contains(std::vector<FieldTag>{ILEV,LEV},src_tag);
+    if (do_remap) {
+      // If we are remapping then we need to initialize the mask source values to 1.0
+      f_src.deep_copy(1.0);
+      f_src.sync_to_host();
```
bartgol, line 372: Is this sync to host really needed?

components/eamxx/src/share/grid/remap/vertical_remapper.cpp
```diff
@@ -276,17 +392,18 @@ void VerticalRemapper::do_remap_fwd ()
...
     using Pack = ekat::Pack<Real,Packsize>;
     using namespace ShortFieldTagsNames;
     using namespace scream::vinterp;
     const auto& layout = f_src.get_header().get_identifier().get_layout();
     const auto  rank   = f_src.rank();
-    auto src_tag = layout.tags().back();
-    auto src_num_levs = layout.dim(src_tag);
-    const bool do_remap = ekat::contains(std::vector<FieldTag>{ILEV,LEV},src_tag);
+    const auto src_tag = layout.tags().back();
+    const auto src_num_levs = layout.dim(src_tag);
```
bartgol, line 404: It's probably simpler to just do `layout.dims().back()`...

components/eamxx/src/share/io/scorpio_output.cpp
```diff
@@ -130,12 +130,13 @@ AtmosphereOutput (const ekat::Comm& comm, const ekat::ParameterList& params,
...
-    // TODO - support both horizontal and vertical remapping together.
-    EKAT_ERROR_MSG("Error! scorpio_output:: We currently do not support both vertical and horizontal remapping.");
-  }
+  // Note: We currently support three remappers
+  //   - vertical remapping from file
+  //   - horizontal remapping from file
+  //   - online remapping which is setup using the create_remapper function
+  const bool use_vertical_remap_from_file = params.isParameter("vertical_remap_file");
+  const bool use_horiz_remap_from_file = params.isParameter("horiz_remap_file");
+  const bool use_online_remapper = io_grid->name()!=fm_grid->name();  // TODO: QUESTION, Do we anticipate online remapping w/ horiz_remap_from file?
```
bartgol, line 139: Since I assume we don't support it now, I would add a check to ensure we don't simultaneously request online remap AND horiz/vertical.

components/eamxx/src/share/io/scorpio_output.cpp
```diff
@@ -440,24 +493,38 @@ void AtmosphereOutput::run (const std::string& filename, const bool is_write_ste
...
-    // not shared with anyone else, so we can safely add its footprint
-    for (const auto& it : *m_io_field_mgr) {
-      const auto& fap = it.second->get_header().get_alloc_properties();
-      if (fap.is_subfield()) {
-        continue;
+  const auto sim_field_mgr = get_field_manager("sim");
+
+  // Cycle through all unique field managers in this output,
+  // first make a copy so we can grab just unique pointers.
+  std::vector<std::shared_ptr<const fm_type>> unique_fms;
```
bartgol, line 500: This sorting is overly complicated, IMHO. All you need is to track which FM you already parsed. You could do:
```
std::set<std::string> grids;
for (auto fm : m_field_mgrs) {
  const auto& gn = fm->get_grid()->name();
  if (grids.count(gn)>0) {
    continue; // grid manager already parsed
  }
  grids.insert (gn);
  // do stuff on this FM.
}
```

components/eamxx/src/share/io/scorpio_output.hpp
```diff
@@ -177,14 +184,12 @@ class AtmosphereOutput
...
-  std::shared_ptr<const fm_type>      m_io_field_mgr;
-  std::shared_ptr<const fm_type>      m_sim_field_mgr;
-  std::shared_ptr<const grid_type>    m_io_grid;
-  std::shared_ptr<remapper_type>      m_horiz_remapper;
-  std::shared_ptr<remapper_type>      m_vert_remapper;
-  std::shared_ptr<const gm_type>      m_grids_manager;
-  bool                                m_horiz_remap_from_file = false;
-  bool                                m_vert_remap_from_file = false;
+  std::vector<std::shared_ptr<const fm_type>> m_field_mgrs;
+  std::map<std::string,int>                   m_field_mgrs_idx;
```
bartgol, line 188: Can't you just map the string to the FM?
```
std::map<std::string,std::shared_ptr<const fm_type>> 	m_field_mgrs;
```
and do away with the indices?

components/eamxx/src/diagnostics/field_at_pressure_level.cpp
```diff
@@ -48,6 +48,27 @@ set_grids(const std::shared_ptr<const GridsManager> grids_manager)
   FieldIdentifier fid (name(),diag_layout, m_field_units, gname);
   m_diagnostic_output = Field(fid);
   m_diagnostic_output.allocate_view();
+
+  // Add a field representing the mask as extra data to the diagnostic field.
```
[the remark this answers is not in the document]
AaronDonahue: I added a comment.  We need two, one if the actual mask and the other is the preallocated view of 1's that is used to generate the mask.

components/eamxx/src/diagnostics/field_at_pressure_level.cpp
```diff
@@ -48,6 +48,27 @@ set_grids(const std::shared_ptr<const GridsManager> grids_manager)
...
+  Field diag_mask(mask_fid);
+  diag_mask.allocate_view();
+  m_diagnostic_output.get_header().set_extra_data("mask_data",diag_mask);
+  m_diagnostic_output.get_header().set_extra_data("mask_value",m_mask_val);
+
+  // Allocate helper views
+  // Note that mPack is by design a pack of size 1, so we need to take into consideration the size of
+  // the source data which view. which  will be equal to the product of the number of source packs
+  // and the simulation pack size.
+  const int num_src_mask_lev = m_field_layout.tags().back()==LEV ? m_num_levs : m_num_levs+1;
```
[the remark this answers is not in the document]
AaronDonahue: Ahh yes, you are correct.

components/eamxx/src/share/grid/remap/vertical_remapper.cpp
AaronDonahue: The `abstract_grid` class takes issue with both a shallow and deep copy in this case:
```
/ascldap/users/asdonah/sandbox_1/scream/components/eamxx/tests/uncoupled/shoc/shoc_standalone.cpp:29: FAILED:
  {Unknown expression after the reported line}
due to unexpected exception with message:
  /ascldap/users/asdonah/sandbox_1/scream/components/eamxx/src/share/grid/
  abstract_grid.cpp:167: FAIL:
  not has_geometry_data(name)
  Error! Cannot create geometry data, since it already exists.
    - grid name: vertical_remap_tgt_grid
    - geo data name: area
    - geo data layout: <COL>(433)
    - input layout: <COL>(433)
```

AaronDonahue: if I follow the comment below then I hit a separate error:
```
/ascldap/users/asdonah/sandbox_1/scream/components/eamxx/tests/uncoupled/shoc/shoc_standalone.cpp:29: FAILED:
  {Unknown expression after the reported line}
due to unexpected exception with message:
  /ascldap/users/asdonah/sandbox_1/scream/components/eamxx/src/share/grid/
  remap/abstract_remapper.hpp:130: FAIL:
  src_fid.get_grid_name()==m_src_grid->name()
  Error! Input FieldIdentifier has the wrong grid name:
     - input src fid grid name: Point Grid
     - remapper src grid name:  vertical_remap_tgt_grid
```

bartgol: Ah, right. Would be nice to add `void Field::share_data (const Field& src)` to circumvent this problem.

bartgol (approved):

PR #2138, opened 2023-01-26
Crusher: Add -scream identifier to machine configuration.

ambrad:
This identifier will deconflict SCREAM's configuration code blocks from anything upstream. Work to consolidate settings, if desired, can then proceed at a natural pace rather than be rushed by our SCREAM->E3SM merge effort.


ambrad:
@oksanaguba @rljacob: Do either of you object to this approach?

jgfouca:
@ambrad , this change will have some downstream effects. I would have to make some minor changes to some of the test scripts in E3SM_test_scripts and it would change the "Site" field on Cdash.

rljacob:
Can you also implement the decision we made here:  https://acme-climate.atlassian.net/wiki/spaces/EIDMG/pages/3446079573/How+to+describe+heterogenous+node+machines+with+CIME
Namely, don't have "gpu" and "cpu" machine entries for single-node-architecture systems like Summit/crusher.  Instead use the compiler entires for a cpu-only compile.

oksanaguba (dismissed):

oksanaguba:
@rljacob i was in the middle of this PR for that suggestion (no -cpu, no -gpu machine) https://github.com/E3SM-Project/scream/pull/2129 , but i thought after today that it would be abandoned and the focus would be on keeping scream machines setup separate and functioning.

ambrad:
@rljacob your request in my opinion is orthogonal to the spirit of this effort, which is to get EAMxx cleanly into upstream ASAP for the benefit of MMF while preserving current SCREAM functionality. In my opinion, the "-scream" identifier is sufficient to make clear that the machine configs are for a specific purpose. Subsequent work on improving compliance with E3SM's approach can occur without a deadline.

ambrad:
@jgfouca yes I was planning to make the script changes myself and then ask you for a cdash entry update once this PR goes in.

jgfouca:
Sounds good.

rljacob:
@oksanaguba adding "scream" is what keeps them separate.   A follow-up PR that gets rid of "-cpu" and "-gpu" would be welcome but it can wait.

oksanaguba:
thanks, however, now i do not know what to do with that PR even after Ben's merge happens. I will close it for now.

ambrad:
> thanks, however, now i do not know what to do with that PR even after Ben's merge happens. I will close it for now.

@oksanaguba one thing you might consider is the following: Once Ben's PR is merged into E3SM, SCREAM will run just as well from that repo as the SCREAM repo. You can then work with Az and others, as necessary, to consolidate the Crusher configurations in E3SM. When the consolidated configuration works for SCREAMv1 tests on Crusher, the -scream one can be removed from E3SM. The next upstream merge to SCREAM will then bring in the consolidated configuration and remove the -scream one.

rljacob (dismissed):

ambrad:
@oksanaguba The CPU test was failing due to an invalid project ID. I've fixed it here: https://github.com/E3SM-Project/scream/commit/8dbac84373ce4d9a663b96ec547eea7f30fa77fd. Have you experienced that before, and are you OK with this change?

oksanaguba:
Andrew, i received your comment " Have you experienced that before, and are you OK with this change?" in email, do not see it here. Yes, i changed project id in the PR i was working on.

ambrad:
I ran several GPU and one CPU test with these modifications and got all PASSes. Manually merging.

PR #2140, opened 2023-01-30
Add a getter to IO interface to grab the timestamp of a file.

AaronDonahue:
This commit introduces the function read_time_at_index to the IO interface for EAMxx.  The purpose of this routine is to read the specific "time" variable to get the timestamp from a file.

This commit also adds the read_time_at_index calls to the io_test unit test which verifies that the output variables store the correct timestamp and verifies that the read_time_at_index routine is working correctly.

### DIFF
--- a/components/eamxx/src/share/io/scream_scorpio_interface.F90
+++ b/components/eamxx/src/share/io/scream_scorpio_interface.F90
@@ -1374,14 +1374,21 @@ contains
     
     type(pio_atm_file_t), pointer :: pio_atm_file
     logical                       :: found
-    integer                       :: ierr
+    integer                       :: dim_id, time_len, ierr
     type(var_desc_t)              :: varid ! netCDF variable ID
     integer                       :: strt(1), cnt(1)
 
     call lookup_pio_atm_file(trim(filename),pio_atm_file,found)
-    if (.not.found) call errorHandle("pio_inq_dimlen ERROR: File "//trim(filename)//" not found",-999)
+    if (.not.found) call errorHandle("read_time_at_index ERROR: File "//trim(filename)//" not found",-999)
     ierr = PIO_inq_varid(pio_atm_file%pioFileDesc,"time",varid)
     call errorHandle('read_time_at_index: Error finding variable ID for "time" in file '//trim(filename)//'.',ierr);
+
+    ierr = pio_inq_dimid(pio_atm_file%pioFileDesc,trim("time"),dim_id)
+    call errorHandle("read_time_at_index ERROR: dimension 'time' not found in file "//trim(filename)//".",ierr)
+    ierr = pio_inq_dimlen(pio_atm_file%pioFileDesc,dim_id,time_len)
+    if (time_index .gt. time_len) then
+      call errorHandle("read_time_at_index ERROR: time_index arg larger than length of time dimension",-999)
+    end if
 
     if (time_index .gt. 0) then
       strt(1) = time_index
--- a/components/eamxx/src/share/io/scream_scorpio_interface.F90
+++ b/components/eamxx/src/share/io/scream_scorpio_interface.F90
@@ -77,6 +77,7 @@ module scream_scorpio_interface
             get_int_attribute,           & ! Retrieves an integer global attribute from the nc file
             set_int_attribute,           & ! Writes an integer global attribute to the nc file
             get_dimlen,                  & ! Returns the length of a specific dimension in a file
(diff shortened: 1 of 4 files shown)


AaronDonahue:
@rebassoo , this PR will allow you to grab a timestamp.  Note, for EAMxx the timestamps are in "days".  So 1 hour is 1/24. days.

components/eamxx/src/share/io/scream_scorpio_interface.F90
```diff
@@ -1359,6 +1360,40 @@ subroutine get_pio_atm_file(filename,pio_file,purpose)
...
+    real(c_double)               :: val_buf(1)
+    
+    type(pio_atm_file_t), pointer :: pio_atm_file
+    logical                       :: found
+    integer                       :: ierr
+    type(var_desc_t)              :: varid ! netCDF variable ID
+    integer                       :: strt(1), cnt(1)
+
+    call lookup_pio_atm_file(trim(filename),pio_atm_file,found)
+    if (.not.found) call errorHandle("pio_inq_dimlen ERROR: File "//trim(filename)//" not found",-999)
```
bartgol, line 1382: Function name in the string is wrong.

components/eamxx/src/share/io/scream_scorpio_interface.F90
```diff
@@ -1359,6 +1360,40 @@ subroutine get_pio_atm_file(filename,pio_file,purpose)
...
+    ierr = PIO_inq_varid(pio_atm_file%pioFileDesc,"time",varid)
+    call errorHandle('read_time_at_index: Error finding variable ID for "time" in file '//trim(filename)//'.',ierr);
+
+    if (time_index .gt. 0) then
+      strt(1) = time_index
```
bartgol, line 1387: Should we inquire the time dim to ensure the time_index is not too large (at least in debug build)?

components/eamxx/src/share/io/scream_scorpio_interface.hpp
```diff
@@ -69,6 +69,8 @@ extern "C" {
   void set_int_attribute_c2f (const char*&& filename, const char*&& attr_name, const int& value);
   int get_dimlen_c2f(const char*&& filename, const char*&& dimname);
   bool has_variable_c2f (const char*&& filename, const char*&& varname);
+  /* Query a netCDF file for the time variable */
+  double read_time_at_index_c2f(const char*&& filename, const int& time_index = -1);
```
bartgol, line 73: How about not defaulting time index to -1, and instead offering `read_curr_time_c2f` as well? It could simply call `get_dimlen_c2f` and then call `read_time_at_index_c2f` with the current time length...

bartgol (changes requested):
Just a minor error msg string fix, plus some up for debate suggestions.

components/eamxx/src/share/io/scream_scorpio_interface.hpp
AaronDonahue: Great suggestion.  Should I get rid of the option to even pass a negative or 0 value to this function, now that the `read_curr_time` is in?

AaronDonahue:
@bartgol , thanks for the review and suggestions.  I've made the changes and had one follow-up question.

components/eamxx/src/share/io/scream_scorpio_interface.hpp
bartgol: Yes, I would say no default for `time_index` anymore, and negative values should be considered an error.

bartgol: @AaronDonahue Are you going to change `read_time_at_index_c2f`, enforcing non-negative time index? Or do you want to leave it like this?

AaronDonahue:
@bartgol I think this is ready for re-review

components/eamxx/src/share/io/scream_scorpio_interface.hpp
```diff
@@ -69,6 +69,9 @@ extern "C" {
   void set_int_attribute_c2f (const char*&& filename, const char*&& attr_name, const int& value);
   int get_dimlen_c2f(const char*&& filename, const char*&& dimname);
   bool has_variable_c2f (const char*&& filename, const char*&& varname);
+  /* Query a netCDF file for the time variable */
+  double read_time_at_index_c2f(const char*&& filename, const int& time_index);
```
AaronDonahue, line 73: @bartgol , This has been changed,  That review was outdated.

bartgol (approved):

PR #2150, opened 2023-02-04
Homme: Fix SCREAM-side discrepancies in Homme, including a bug fix.

ambrad:
Bug: Sx, Sy should not be set to 0 in namelist_mod, as doing so overwrites non-0 namelist values. They are already correctly zero'ed in control_mod.F90 at initialization. These spurious lines (that appear in just the SCREAM repo) are why homme_integration doesn't pass in the SCREAM repo.

Remove redundant MPI_bcast lines, also only in the SCREAM repo.

Fix indentation to agree with upstream.


ambrad:
homme_integration on Chrysalis fails against baselines in the planar tests without this commit and passes with it.

brhillman (approved):

PR #2153, opened 2023-02-07
EAMxx: Use type annotation in mesh_file entry.

ambrad:
This should make the grid auto-download through CIME.


bartgol (approved):

jgfouca (approved):

jgfouca:
@bartgol , I assume our system is smart enough to not try to download `none`?

ambrad:
I'll check. Maybe it needs to be `UNSET` rather than `none`.

ambrad:
I tried an ne30pg2 run and see no evidence of spurious downloading.

PR #2154, opened 2023-02-07
Allow using non-integer timestep inside atmosphere processes

bartgol:
There are only two places where we need an integer dt: when we interface to the component coupler, and when we update a timestamp. All other places in the code, can use a non-integer value.

With this mod, atmosphere processes can use a number of subcycles that yields non-integer dt. Notice that, if we're in the middle of a subcyle, the atm proc class will not attempt to update the computed fields' timestamps, so the non-integer dt is not a problem. The TimeStamp class checks that the input dt is "close to be integer" when updating the time; if it's not, it throws, otherwise it rounds dt to an int, and then updates the time. When all the subcycles are completed, we should be back to an integer dt (modulo some roundoffs), so updating time is again ok.

### DIFF
--- a/components/eamxx/src/dynamics/homme/atmosphere_dynamics.cpp
+++ b/components/eamxx/src/dynamics/homme/atmosphere_dynamics.cpp
@@ -461,8 +461,17 @@ void HommeDynamics::run_impl (const double dt)
 
     // NOTE: we did not have atm_dt when we inited homme, so we set nsplit=1.
     //       Now we can compute the actual nsplit, and need to update its value
-    //       in Hommexx's data structures
-    const int nsplit = get_homme_nsplit_f90(dt);
+    //       in Hommexx's data structures.
+    //       Also, nsplit calculation requires an integer atm timestep, so we
+    //       check to ensure that's the case
+    EKAT_REQUIRE_MSG (std::abs(dt-std::round(dt))<std::numeric_limits<double>::epsilon()*10,
+        "[HommeDynamics] Error! Input timestep departure from integer above tolerance.\n"
+        "  - input dt : " << dt << "\n"
+        "  - tolerance: " << std::numeric_limits<double>::epsilon()*10 << "\n");
+
+    const int dt_int = static_cast<int>(std::round(dt));
+
+    const int nsplit = get_homme_nsplit_f90(dt_int);
     const auto& c = Homme::Context::singleton();
     auto& params = c.get<Homme::SimulationParams>();
     params.nsplit = nsplit;
@@ -514,7 +523,7 @@ void HommeDynamics::set_computed_group_impl (const FieldGroup& group)
   }
 }
 
-void HommeDynamics::homme_pre_process (const int dt) {
+void HommeDynamics::homme_pre_process (const double dt) {
   // T and uv tendencies are backed out on the ref grid.
   // Homme takes care of turning the FT tendency into a tendency for VTheta_dp.
   using KT = KokkosTypes<DefaultDevice>;
@@ -618,7 +627,7 @@ void HommeDynamics::homme_pre_process (const int dt) {
   }
 }
 
-void HommeDynamics::homme_post_process (const int dt) {
+void HommeDynamics::homme_post_process (const double dt) {
   const auto& pgn = m_phys_grid->name();
   const auto& c = Homme::Context::singleton();
   const auto& params = c.get<Homme::SimulationParams>();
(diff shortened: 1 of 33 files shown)


components/eamxx/src/physics/shoc/atmosphere_macrophysics.cpp
```diff
@@ -461,7 +461,7 @@ void SHOCMacrophysics::run_impl (const int dt)
   // number of SHOC timesteps (nadv) to be 1.
   // TODO: input parameter?
   hdtime = dt;
-  m_nadv = std::max(hdtime/dt,1);
+  m_nadv = std::max(static_cast<int>(hdtime/dt),1);
```
ambrad, line 464: Need to be careful here. Casting to an int will always round down. `round` should be used for proper rounding.

components/eamxx/src/physics/spa/tests/spa_main_test.cpp
```diff
@@ -192,9 +192,9 @@ TEST_CASE("spa_main")
   std::cout << "  -> time interp, t_beg<t<t_end\n";
 
   RPDF pdf(0.001,0.999);
-  auto dt = int ( (t_end-t_beg) * pdf(engine) );
+  auto dt = (t_end-t_beg) * pdf(engine);
 
-  auto t_mid = t_beg + dt;
+  auto t_mid = t_beg + static_cast<int>(dt);
```
ambrad, line 197: `round` here, too.

components/eamxx/src/physics/shoc/atmosphere_macrophysics.cpp
ambrad: That said, `hdtime = dt` means this will always be 1, anyway.

components/eamxx/src/physics/spa/tests/spa_main_test.cpp
ambrad: But do you actually need this to be an `int`? Below I see the time class can handle `double`.

bartgol: This is for a unit test, so I wasn't really concerned about the +/- one. But I can change it to the nearest one.

components/eamxx/src/physics/shoc/atmosphere_macrophysics.cpp
bartgol: Good point. We don't want to round 2.9999 to 2 if/when we will allow hdtime!=dt. I can fix this.

ambrad:
It would be good to run e3sm_scream_v1_master_gnu9 on mappy against baselines before merging.

bartgol:
Running `e3sm_scream_v1_medres` and `e3sm_scream_v1_mpassi` now (the lowres were run by AT).

ambrad (changes requested):

components/eamxx/src/dynamics/homme/atmosphere_dynamics.cpp
```diff
@@ -452,7 +452,7 @@ void HommeDynamics::initialize_impl (const RunType run_type)
   rayleigh_friction_init();
 }
 
-void HommeDynamics::run_impl (const int dt)
+void HommeDynamics::run_impl (const double dt)
```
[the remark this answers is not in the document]
ambrad: Whoops, sorry, I forgot to "finish" the review, so this didn't get posted until after my comment abut mappy.

ambrad (approved):
I'm approving, but please post the results of the mappy run before merging.

bartgol:
> I'm approving, but please post the results of the mappy run before merging.

```
$ ~/acme/scratch/cs.status.20230209_114714_miaf3u -s
20230209_114714_miaf3u: 5 tests
  PASS ERS_D_Ln22.ne4pg2_oQU480.F2010-SCREAMv1-MPASSI.mappy_gnu9.atmlndactive-rtm_off
  PASS ERS_Ln22.ne30_ne30.F2010-SCREAMv1.mappy_gnu9
  PASS ERS_Ln22.ne30pg2_ne30pg2.F2010-SCREAMv1-X6T.mappy_gnu9
  PASS ERS_Ln90.ne30pg2_ne30pg2.F2010-SCREAMv1.mappy_gnu9.scream-small_kernels
  PASS PEM_Ln90.ne30pg2_ne30pg2.F2010-SCREAMv1.mappy_gnu9
```
The lowres were already ran by the AT. These were midres+mpassi.

PR #2155, opened 2023-02-07
EAMxx: PE layouts for RRM nightly test.

ambrad:


ambrad:
The RRM test passes on Ascent and Crusher. I'm guessing the layout for cori-knl and will fix it with another PR if I got it wrong (out of memory). PM-GPU is also a guess, but we'll need the PM-GPU workaround in before I test. Mappy should be fixed by #2153.

ambrad:
I'm merging manually because this PR changes only PE layouts, and I've tested the results on two machines.

PR #2156, opened 2023-02-08
Add ne120pg2 and ne256pg2 monogrids

brhillman:
Add ne120pg2 and ne256pg2 monogrid configurations. These are run by specifying resolution `ne120pg2_ne120pg2` and `ne256pg2_ne256pg2`. Note that `EPS_AGRID` tolerance needed to be increased to avoid failing the domain grid check (differences in coordinates less than 1e-10, but default tolerance is 1e-12).

[B4B]

### DIFF
--- a/cime_config/config_grids.xml
+++ b/cime_config/config_grids.xml
@@ -2552,8 +2552,8 @@
       <ny>1</ny>
       <file grid="atm|lnd" mask="oRRS18to6v3">$DIN_LOC_ROOT/share/domains/domain.lnd.ne256pg2_oRRS18to6v3.200212.nc</file>
       <file grid="ice|ocn" mask="oRRS18to6v3">$DIN_LOC_ROOT/share/domains/domain.ocn.ne256pg2_oRRS18to6v3.200212.nc</file>
-      <file grid="atm|lnd" mask="ICOS10">/global/cfs/cdirs/e3sm/bhillma/grids/ne256pg2/domain_files/domain.lnd.ne256pg2_ICOS10.230201.nc</file>
-      <file grid="ice|ocn" mask="ICOS10">/global/cfs/cdirs/e3sm/bhillma/grids/ne256pg2/domain_files/domain.ocn.ne256pg2_ICOS10.230201.nc</file>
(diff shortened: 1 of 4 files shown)


mt5555:
EPS_AGRID issue in longstanding (but harmless in this case):

https://github.com/E3SM-Project/E3SM/issues/4437

ndkeen (dismissed):
I tested smoke test of 5 days of `ne256pg2_ne256pg2` and `ne256pg2_r0125_oRRS18to6v3` on pm-gpu

ambrad:
I think merging master into this branch will resolve the mappy issue with submodules.

brhillman:
@ambrad I don't understand the issue with submodules; and I thought the autotester was working on a merge of the candidate branch into master?

ambrad:
@brhillman I don't understand it either, but doing that on another PR either resolved the issue, or the issue resolved randomly independent of that on the next AT iteration.

brhillman:
@ndkeen @ambrad can I get a re-review since I rebased so this will automerge?

bartgol:
For the curious, here's the submodule issue.

Jenkins works in two phases: in the pre-build phase, the repo is fetched using a git plugin, which uses @jgfouca public ssh key to fetch. In the actual build phase, a shell script is executed by the `e3sm-jenkins` user. Now, when the repo is cloned, only master and the PR branch are downloaded, with the latter being checked out. The scripts then perform a master merge, followed by a submodule update. If there's a submodule in master that was not in the branch, the submodule needs to be refetched. But now we're in a shell script, using `e3sm-jenkins` ssh credentials, which are nowhere to be found on github, so github declines the fetch.

Jim and I talked about this, and the solution is to either copy @jgfouca 's  (or anyone of us, really) ssh public key into `e3sm-jenkins`'s ssh folder, OR to create a new ssh key for `e3sm-jenkins`, and add it to the github account of `E3SM-Autotester`. I would vote for the latter (seems cleaner, and more stable), but the former is probably the fastest. @jgfouca could you please take care of the ssh pubkey when you have a chance?

jgfouca:
@bartgol , I put my ssh keys in e3sm-jenkins' .ssh. It looks like weaver/blake already had them.

ndkeen (approved):

PR #2159, opened 2023-02-08
Adding nudging capability for fields

rebassoo:
This request adds a physics module called nudging that will read in an external netcdf file and will nudge the internal T_mid, qv, and horiz_winds fields to the values in the external netcdf file. Interpolation in time and in pressure levels is required. The interpolation in pressure levels is necessary because the pressure levels from the external netcdf file are not necessarily the same as for the internal field.


components/eamxx/src/physics/nudging/atmosphere_nudging.cpp
```diff
@@ -0,0 +1,264 @@
...
+
+  auto T_mid       = get_field_out("T_mid").get_view<mPack**>();
+  auto qv          = get_field_out("qv").get_view<mPack**>();
+  auto hw          = get_field_out("horiz_winds").get_view<mPack***>();
+
+  const auto& p_mid    = get_field_in("p_mid").get_view<const mPack**>();
+  auto p_mid_v = view_Nd<mPack,2>("",p_mid.extent_int(0),p_mid.extent_int(1));
+  Kokkos::deep_copy(p_mid_v,p_mid);
+  
+  view_Nd<mPack,2> T_mid_out("T_mid_out",m_num_cols,T_mid.extent_int(1));
```
bartgol, line 178: Creating views on device at runtime is not advised, due to cost of allocation. Since I see you eventually copy this back into T_mid, why don't you just use T_mid? Same goes for qv, and hw.

bartgol: As for the `_p` variables, I don't think you need them. Can't you just replace every use of, say, `T_mid_ext_p` with `T_mid_ext`?

components/eamxx/src/physics/nudging/atmosphere_nudging.cpp
```diff
@@ -0,0 +1,264 @@
...
+  data_input=data_input0;
+
+  T_mid_ext = fields_ext["T_mid"];
+  p_mid_ext = fields_ext["p_mid"];
+  horiz_winds_ext = fields_ext_3d["horiz_winds"];
+  qv_ext = fields_ext["qv"];
+  ts0=timestamp();
+}
+
+void NUDGING::time_interpolation (const int time_s) {
```
bartgol, line 88: It seems that this fcn runs at every time step. I'm assuming the input data is on a coarser time grid than that where the simulation runs, so this ends up re-reading the same data for a few time steps. I/O is not cheap, so we should probably store X_bef and X_aft as class members, and update them only when time_index changes.

In fact, I think this class should share a lot of code with SPA. Perhaps we should factor out some of that code? @AaronDonahue

components/eamxx/src/physics/nudging/atmosphere_nudging.cpp
```diff
@@ -0,0 +1,264 @@
+#include "atmosphere_nudging.hpp"
+
+namespace scream
+{
+
+  //using namespace spa;
+// =========================================================================================
+NUDGING::NUDGING (const ekat::Comm& comm, const ekat::ParameterList& params)
```
bartgol, line 8: I would not use all uppercase names. Uppercase are associated with either acronyms or compile-time constants. I would just use `Nudging`.

components/eamxx/src/physics/nudging/atmosphere_nudging.cpp
```diff
@@ -0,0 +1,264 @@
...
+  : AtmosphereProcess(comm, params)
+{
+  m_fnames=m_params.get<std::vector<std::string>>("Field_Names");
```
bartgol, line 11: Are the field names really something that the user can choose? It seems that the impl assumes we nudge T, qv, and horiz_winds. If that's always the case, I would not expose this entry in the input param list.

components/eamxx/src/physics/nudging/atmosphere_nudging.cpp
```diff
@@ -0,0 +1,264 @@
...
+  datafile=m_params.get<std::string>("Nudging_Filename");
+  time_step_file=m_params.get<Int>("Time_Step_File",-9999);
```
bartgol, line 13: Is this the timestep in the input NC file? Maybe we could read it straight from the NC file, rather than giving the impression to the user that they can change it? Perhaps, we could even read all the timesteps in the input file at init time, and store them in a std::vector, and compute the bef/aft time index based on the current time?

components/eamxx/src/physics/nudging/atmosphere_nudging.cpp
```diff
@@ -0,0 +1,264 @@
...
+
+  //Now perform the vertical interpolation from the external file to the internal
+  //field levels
+  perform_vertical_interpolation<Real,1,2>(p_mid_ext_p,
+                                           p_mid_v,
+                                           T_mid_ext_p,
+                                           T_mid_out,
+                                           m_num_src_levs,
+                                           m_num_levs);
+  Kokkos::deep_copy(T_mid,T_mid_out);
```
bartgol, line 204: I thought nudging was meant to give a little push toward the "reference" solution. Here, I see that we set the solution to _match_ the reference one. I guess I'm showing my newbiness here, but is that what nudging really is?

components/eamxx/src/physics/nudging/atmosphere_nudging.cpp
rebassoo: I will remove it, this was originally there because we wanted to make it generic to fields that a user can choose. However, I later changed to just doing the fields we care about for the nudging.

components/eamxx/src/physics/nudging/atmosphere_nudging.cpp
rebassoo: Fixed

components/eamxx/src/physics/nudging/atmosphere_nudging.cpp
rebassoo: Yes, this is the timestep in the input NC file. This is a good suggestion, I will implement this, thanks

components/eamxx/src/physics/nudging/atmosphere_nudging.cpp
rebassoo: Yes, it is meant to match the reference one not just give it a little push towards the reference one. Maybe it is a bit of a misnomer, but it is the nomenclature used in the literature.  You are only matching a few of the reference fields and keeping the rest of the internal fields unchanged, so maybe this is the explanation as to why it is denoted as nudging. You are replacing some part of the model but not the whole thing. Maybe there is a better name

components/eamxx/src/physics/nudging/atmosphere_nudging.cpp
rebassoo: The reason I had to convert these is that the vertical interpolation is expecting a view_Nd, so if you try to put T_mid in directly or T_mid_ext in directly it gives the error below. Previously I think the vertical interpolation was templated so that this would have been ok, but with the recent updates to the vertical interpolation by Aaron I think you have to specify it in terms of a view_Nd. I will see if there is a way around this or if the vertical interpolation needs to be updated. 
error: no instance of overloaded function "perform_vertical_interpolation" matches the argument list
            argument types are: (scream::vinterp::view_Nd<scream::Nudging::mPack, 2>, scream::vinterp::view_Nd<scream::Nudging::mPack, 2>, scream::Nudging::view_2d<scream::Real>, scream::vinterp::view_Nd<scream::Nudging::mPack, 2>, int, int)

components/eamxx/src/physics/nudging/atmosphere_nudging.cpp
rebassoo: Thanks for the comment, I will talk to Aaron about this.

components/eamxx/src/physics/nudging/atmosphere_nudging.cpp
bartgol: Thanks!

components/eamxx/src/physics/nudging/atmosphere_nudging.cpp
bartgol: I see that `T_mid_ext` is a view of Real, while `T_mid_ext_p` is a view of Packs. First, the deep copy below looks dangerous. I suspect it ony works because you are using packs of size 1.

Regardless, if you need to turn a view of Real into a view of Pack, you can just do
```
auto packed_data = reinterpret_cast<mPack*>(T_mid_ext.data());
int num_packs = ekat::PackInfo<mPack::n>::num_packs(num_levs); // <- num_levs: whatever num levs T_mid_ext uses
view_Nd<mPack,2> T_mid_ext_p ( packed_data, num_cols,num_packs); // <- num_cols: whatever num_cols T_mid_ext uses
```

rebassoo: mPack always has a pack size of 1. The reason I did this was because of this issue when converting from a view of Real to a view of Pack and I thought there would be problems with a pack size different from 1 (and wasn't sure how to resolve). In the case of pack size 1 I think you could simplify this to one line(?):
view_Nd<mPack,2> T_mid_ext_p( reinterpret_cast<mPack*> (T_mid_ext.data()),num_cols,num_levs);

An alternative solution is whether it might be possible to get fields instead of views from a file using AtmosphereInput. If this is possible then I could get the fields from the external field and then just do field.get_view<mPack**>. But I am not sure if this is possible. It looks like I might be able to create a field manager and fill it with the fields and then feed it to AtmosphereInput.

rebassoo: One additional comment with respect to "why don't you just use T_mid? Same goes for qv, and hw." Eventually we want to actually output to a file the nudging tendencies, i.e. |original T_mid - T_mid from external file|. So we don't want to overwrite T_mid. But this will probably come in a later PR, so for now I can just do it like you suggest with the idea that this will be updated later.

bartgol: Yes, the reintepret cast line you posted should work.

bartgol: As for AtmInput, we do not offer a way to read into fields, but we _could_ think of adding such capability. After all, the setup version that takes layouts and views could be easily called by a version that takes Fields. We may even benefit a tiny bit from this simplification in other parts of the EAMxx, like when we read geo data from IC file or topo from topofile.

components/eamxx/src/physics/nudging/atmosphere_nudging.cpp
rebassoo: Talking to Aaron, I think that this would be an end goal but not for this PR. It would take a bit of work to share the code and we want to get something working to provide some collaborators with nudged data. For now I will create a struct that holds the data. I think there would be a nudging 2.0 at some point that would share code with SPA and then also possibly not hard code the parameters you want to nudge but have the user be able to specify that. But we think these updates are better to hold off on for now. thanks

components/eamxx/src/physics/nudging/atmosphere_nudging.cpp
rebassoo: Ok, for now I will implement the reinterpret cast and we can think about if we want to add the AtmInput in the future.

components/eamxx/src/physics/nudging/atmosphere_nudging.cpp
bartgol: ok

components/eamxx/src/physics/nudging/atmosphere_nudging.cpp
rebassoo: Removed

components/eamxx/src/physics/nudging/atmosphere_nudging.cpp
rebassoo: I initially implemented this std::vector, but after I implemented the struct with the before and after time_state data I actually didn't need it so I removed it. The NudgingData struct contains time information as well which is what I use to figure out if you need to load new data or not.

components/eamxx/src/physics/nudging/atmosphere_nudging.cpp
```diff
@@ -40,9 +38,21 @@ void Nudging::set_grids(const std::shared_ptr<const GridsManager> grids_manager)
...
   double time_value_0= scorpio::read_time_at_index_c2f(datafile.c_str(),0);
-  std::cout<<"time_value_0: "<<time_value_0<<std::endl;
   double time_value_1= scorpio::read_time_at_index_c2f(datafile.c_str(),1);
-  std::cout<<"time_value_1: "<<time_value_1<<std::endl;
+  double time_value_2= scorpio::read_time_at_index_c2f(datafile.c_str(),2);
+
+  //The times for time_steps are in units of day, not seconds
+  if (time_value_1>0){
+    time_steps.push_back(time_value_1);
+    int n_time_steps = time_value_0/time_value_1;
```
bartgol: I don't get this line. Why would the ratio t1/t0 be a number of time steps?

components/eamxx/src/physics/nudging/atmosphere_nudging.cpp
```diff
@@ -0,0 +1,275 @@
...
+  double w_aft = (time_s-(time_index)*time_step_file) / time_step_file_d;
+  
+  const int num_cols = NudgingData_aft.T_mid.extent(0);
+  const int num_vert_packs = NudgingData_aft.T_mid.extent(1);
+  const auto policy = ESU::get_default_team_policy(num_cols, num_vert_packs);
+  Kokkos::parallel_for("nudging_time_interpolation", policy,
+     	       KOKKOS_LAMBDA(MemberType const& team) {
+      const int icol = team.league_rank();
+      auto T_mid_1d = ekat::subview(T_mid_ext,icol);
+      auto T_mid_bef_1d = ekat::subview(NudgingData_bef.T_mid,icol);
```
bartgol, line 122: All accesses to `NudgingData_xyz` are accessing members of this class. You are not technically allowed to do this inside a `KOKKOS_LAMBDA`. If/when we upgrade to C++17, you can use `KOKKOS_CLASS_LAMBDA`, but until then, you need to do one of the following:
a) use a functor
b) create local copies of `NudgingData_xyz` (as well as any other class-member) right before the pfor call (views are cheap to copy, so that's not expensive).

components/eamxx/src/physics/nudging/atmosphere_nudging.cpp
```diff
@@ -0,0 +1,275 @@
...
+
+  perform_vertical_interpolation<Real,1,3>(p_mid_ext_p,
+                                           p_mid_v,
+                                           hw_ext_p,
+                                           hw,
+                                           m_num_src_levs,
+                                           m_num_levs);
+
+  //This code removes any masked values and sets them to the corresponding
+  //values at the lowest/highest pressure levels from the external file
```
bartgol, line 234: Wait, I thought we were using ekat's linear interpolation in such a way that we were not extrapolating. That is, if you pad x1 and y1 passed to lin interp as (pseudo code)
```
L = some-very-large-x (for pressure, it could be 2.0)
x1 = [x1 L]
y1 = [y1 y1(end)]
```
then you will be guaranteed that, ix x2 is larger than the max of the original (pre-padding) values of x1, then y2 will be set to the last entry of y1.

components/eamxx/src/share/io/scorpio_input.cpp
```diff
@@ -9,6 +9,10 @@
 namespace scream
 {
 
+AtmosphereInput::AtmosphereInput()
```
bartgol, line 12: Looks like you added this empty ctor b/c you added a member of type `AtmosphereInput` to `Nudging`. Instead of adding an interface just for the sake of that class, I would have `Nudging` store a `std::shared_ptr<AtmosphereInput>`, which you can then init once you have all the data, and I would remove this empty ctor.

bartgol (changes requested):

components/eamxx/src/physics/nudging/atmosphere_nudging.cpp
rebassoo: It is because t0 is actually the last time in the file. This is how scorpio::read_time_at_index_c2f was implemented that scorpio::read_time_at_index_c2f(filename,0) will give you the last time in the file. However, this code is actually unnecessary and I removed this block (8 lines from L44-L51). It isn't necessary because I store the time information in the NudgingData struct and compare to the current time.

components/eamxx/src/physics/nudging/atmosphere_nudging.cpp
rebassoo: Ok, thanks for the details. I have gone for the local copy route by adding this before the pfor call:
  auto T_mid_bef = view_2d<Real>("",NudgingData_bef.T_mid.extent_int(0),                               NudgingData_bef.T_mid.extent_int(1));
  Kokkos::deep_copy(T_mid_bef,NudgingData_bef.T_mid);
And similar for the other data.

components/eamxx/src/physics/nudging/atmosphere_nudging.cpp
rebassoo: Yes, ekat's linear interpolation is not extrapolating, so you could pre-pad and maybe I should have done this. I have to say it seemed easier to remove the masked values after, but maybe it is better to pre-pad. Just so I understand this would entail taking for instance T_mid_ext_p (see below) and adding two more levels, one at the bottom and one at the top and then setting the bottom value to the bottom+1 value and the top to the top-1 value
  const view_Nd<mPack,2> T_mid_ext_p(reinterpret_cast<mPack*>(T_mid_ext.data()),
                                     m_num_cols,m_num_src_levs);
And then as you say for the pressure add two levels with one very large and one very small so that there are never any layers outside those bounds. I guess what was making it difficult was having to define a new T_mid field with two extra levels. Is there an easy way to add levels to a view as opposed to creating a new view with +2 levels and looping over this new one and filling in values from the original (which doesn't seem any easier than looping over to remove masked values)?

components/eamxx/src/share/io/scorpio_input.cpp
rebassoo: Thanks for the input. This is a better solution. I put std::shared_ptr<AtmosphereInput> data_input; 
in the Nudging header file and then these lines where the data is filled. I believe this is what you were suggesting.
 AtmosphereInput data_input0(m_comm,data_in_params);
  data_input0.init(grid_l,host_views,layouts);
  data_input=std::make_shared<AtmosphereInput>(data_input0);

rebassoo: I also removed the empty ctor

components/eamxx/src/physics/nudging/atmosphere_nudging.cpp
```diff
@@ -82,7 +82,7 @@ void Nudging::initialize_impl (const RunType /* run_type */)
   data_in_params.set("Skip_Grid_Checks",true);  // We need to skip grid checks because multiple ranks may want the same column of source data.
   AtmosphereInput data_input0(m_comm,data_in_params);
   data_input0.init(grid_l,host_views,layouts);
-  data_input=data_input0;
+  data_input=std::make_shared<AtmosphereInput>(data_input0);
```
bartgol, line 85: This is prob overly complicated. You could just do
```
data_input = std::make_shared<AtmosphereInput>(m_comm,data_in_params);
data_input->init(grid_1,host_views,layouts);
```

components/eamxx/src/physics/nudging/atmosphere_nudging.cpp
rebassoo: Will keep as is after conversations via Slack

components/eamxx/src/physics/nudging/atmosphere_nudging.cpp
rebassoo: I have updated, thanks

components/eamxx/src/physics/nudging/atmosphere_nudging.cpp
```diff
@@ -23,27 +23,27 @@ void Nudging::set_grids(const std::shared_ptr<const GridsManager> grids_manager)
...
   m_num_src_levs = scorpio::get_dimlen_c2f(datafile.c_str(),"lev");
   double time_value_1= scorpio::read_time_at_index_c2f(datafile.c_str(),1);
   double time_value_2= scorpio::read_time_at_index_c2f(datafile.c_str(),2);
-
+    
   //Here we are assuming that the time in the netcdf file is in days
   //Internally we want this in seconds so need to convert
   //Only consider integer time steps in seconds to resolve any roundoff error
   time_step_file=int((time_value_2-time_value_1)*86400);
+  std::cout<<"time_step_file: "<<time_step_file<<std::endl;
```
bartgol, line 46: Avoid printing stuff to cout, and use the `log` method instead. If instead this was just some print statement during debugging, you can just prune it.

components/eamxx/src/physics/nudging/atmosphere_nudging.hpp
```diff
@@ -37,6 +36,9 @@ class Nudging : public AtmosphereProcess
   template <typename S>
   using view_2d = typename KT::template view_2d<S>;
 
+  //template <typename S>
```
bartgol, line 39: If you don't need this, let's remove it.

components/eamxx/src/physics/nudging/atmosphere_nudging.hpp
```diff
@@ -82,11 +96,15 @@ class Nudging : public AtmosphereProcess
   std::map<std::string,FieldLayout>  layouts;
   std::vector<std::string> m_fnames;
   std::map<std::string,view_2d<Real>> fields_ext;
+  std::map<std::string,view_2d_host<Real>> fields_ext_h;
   std::map<std::string,view_3d<Real>> fields_ext_3d;
+  std::map<std::string,view_3d_host<Real>> fields_ext_3d_h;
   view_2d<Real> T_mid_ext;
   view_2d<Real> p_mid_ext;
-  view_3d<Real> horiz_winds_ext;
+  //view_3d<Real> horiz_winds_ext;
```
bartgol, line 104: Commented code to remove.

components/eamxx/src/physics/nudging/nudging_functions.hpp
```diff
@@ -31,8 +31,10 @@ struct NudgingFunctions
       if (allocate){
         T_mid = view_2d<Real>("",ncols,nlevs);
         p_mid = view_2d<Real>("",ncols,nlevs);
-        hw = view_3d<Real>("",ncols,2,nlevs);
+        //hw = view_3d<Real>("",ncols,2,nlevs);
```
bartgol, line 34: Commented code to remove

components/eamxx/src/physics/nudging/nudging_functions.hpp
```diff
@@ -41,8 +43,10 @@ struct NudgingFunctions
     int time;
     view_2d<Real> T_mid;
     view_2d<Real> p_mid;
-    view_3d<Real> hw;
+    //view_3d<Real> hw;
```
bartgol, line 46: Commented code to remove

components/eamxx/src/physics/nudging/atmosphere_nudging.cpp
```diff
@@ -54,50 +54,97 @@ void Nudging::initialize_impl (const RunType /* run_type */)
...
+  int start_date=scorpio::get_int_attribute_c2f(datafile.c_str(),"start_date");
+  int start_time=scorpio::get_int_attribute_c2f(datafile.c_str(),"start_time");
+  int start_year=int(start_date/10000);
+  int start_month=int((start_date-start_year*10000)/100);
+  int start_day=int(start_date-start_year*10000-start_month*100);
+  int start_hour=int(start_time/10000);
+  int start_min=int((start_time-start_hour*10000)/100);
+  int start_sec=int(start_time-start_hour*10000-start_min*100);
+
+  EKAT_REQUIRE_MSG(start_year==ts0.get_date()[0],
```
bartgol, line 114: For clarity (and to avoid relying on internal impl details), instead of `get_date()[k]` and `get_time()[k]`, you can use `get_year()`, `get_month()`, ...

components/eamxx/src/physics/nudging/atmosphere_nudging.cpp
```diff
@@ -111,7 +158,10 @@ void Nudging::time_interpolation (const int time_s) {
   double time_step_file_d = time_step_file;
   double w_bef = ((time_index+1)*time_step_file-time_s) / time_step_file_d;
   double w_aft = (time_s-(time_index)*time_step_file) / time_step_file_d;
-  
+  std::cout<<"w_bef: "<<w_bef<<std::endl;
```
bartgol, line 161: If this is output you needed while debugging, please remove it. Otherwise, use the `log` method.

components/eamxx/src/physics/nudging/atmosphere_nudging.cpp
```diff
@@ -122,34 +172,49 @@ void Nudging::time_interpolation (const int time_s) {
   auto T_mid_aft = view_2d<Real>("",NudgingData_aft.T_mid.extent_int(0),
```
bartgol, line 172: These views are created on device at every timestep. That's not great, especially on GPUs. You should allocate them once and for all at init time, and reuse them over and over.

bartgol: To be fair though, I am not sure why you need these local deep copies. Can't you just use, say, `NudgingData_aft.T_mid` everywhere you use `T_mid_aft` (and similarly for other vars)?

components/eamxx/src/physics/nudging/atmosphere_nudging.cpp
```diff
@@ -177,26 +241,27 @@ void Nudging::update_time_step (const int time_s)
...
-      Kokkos::deep_copy(NudgingData_bef.hw,horiz_winds_ext);
-      Kokkos::deep_copy(NudgingData_bef.qv,qv_ext);
-      NudgingData_bef.time = time_step_file*(time_index);
-
-      data_input->read_variables(time_index);
-      Kokkos::deep_copy(NudgingData_aft.T_mid,T_mid_ext);
-      Kokkos::deep_copy(NudgingData_aft.p_mid,p_mid_ext);
-      Kokkos::deep_copy(NudgingData_aft.hw,horiz_winds_ext);
-      Kokkos::deep_copy(NudgingData_aft.qv,qv_ext);
+      Kokkos::deep_copy(NudgingData_bef.T_mid,NudgingData_aft.T_mid);
```
bartgol, line 244: Instead of a deep copy, you could consider doing a swap

```
std::swap (NudgingData_bef,NudringData_aft);
```
This will simply swap a few pointers, without doing unnecessary deep copies.

bartgol: Note: all you need to do to use std::swap is to implement `NudgingData& operator=(const NudgingData&)` (which should do a shallow assignment).

bartgol: Actually, the compiler should already implicitly provide the correct one, but you may consider making it clear via
```
NudgingData& operator=(const NudgingData&) = default;
```
in the class declaration.  Explicit intent is always nice.

components/eamxx/src/physics/nudging/atmosphere_nudging.cpp
```diff
@@ -212,26 +277,29 @@ void Nudging::run_impl (const int dt)
...
 
   auto T_mid       = get_field_out("T_mid").get_view<mPack**>();
   auto qv          = get_field_out("qv").get_view<mPack**>();
-  auto hw          = get_field_out("horiz_winds").get_view<mPack***>();
+  auto u          = get_field_out("u").get_view<mPack**>();
+  auto v          = get_field_out("v").get_view<mPack**>();
 
   const auto& p_mid    = get_field_in("p_mid").get_view<const mPack**>();
   auto p_mid_v = view_Nd<mPack,2>("",p_mid.extent_int(0),p_mid.extent_int(1));
   Kokkos::deep_copy(p_mid_v,p_mid);
```
bartgol, line 291: I'm not sure I understand why you have this deep copy. Can't you do the same thing you do for the other views right below?

components/eamxx/src/physics/nudging/atmosphere_nudging.cpp
```diff
@@ -281,15 +359,14 @@ void Nudging::run_impl (const int dt)
         T_mid_1d(k).set(below_min,T_mid_ext_1d(0));
         qv_1d(k).set(above_max,qv_ext_1d(num_vert_packs_ext-1));
         qv_1d(k).set(below_min,qv_ext_1d(0));
-        hw_2d(0,k).set(above_max,hw_ext_2d(0,num_vert_packs_ext-1));
-        hw_2d(0,k).set(below_min,hw_ext_2d(0,0));
-        hw_2d(1,k).set(above_max,hw_ext_2d(1,num_vert_packs_ext-1));
-        hw_2d(1,k).set(below_min,hw_ext_2d(1,0));
+        u_1d(k).set(above_max,u_ext_1d(num_vert_packs_ext-1));
```
bartgol, line 362: You may want to wrap all this set calls in a `if (above_max.any())` and `if (below_min.any())` respectively.

components/eamxx/src/physics/nudging/tests/nudging_tests.cpp
```diff
@@ -99,67 +97,69 @@ TEST_CASE("nudging") {
...
     fm->init_fields_time_stamp(time);
     // Sync back to device
+    f1.sync_to_dev();
     f2.sync_to_dev();
     f3.sync_to_dev();
     f4.sync_to_dev();
     f5.sync_to_dev();
 
-    fm->init_fields_time_stamp(t0);
+    //fm->init_fields_time_stamp(t0);
```
bartgol, line 155: Remove commented code.

bartgol (changes requested):

components/eamxx/src/physics/nudging/atmosphere_nudging.cpp
bartgol: Ah, sorry, I should have followed up on this. By copy I meant shallow copy, not deep copy. So
```
auto T_mid_bef = NudgingData_bef.T_mid;
```
is perfect to get around the `this` issue of `KOKKOS_LAMBDA`.

components/eamxx/src/physics/nudging/atmosphere_nudging.cpp
rebassoo: Yes, I just noticed that I thought I had committed the removal of these debug statements but they are only in my local repo. This has been removed in the latest commit.

components/eamxx/src/physics/nudging/atmosphere_nudging.hpp
rebassoo: removed

components/eamxx/src/physics/nudging/atmosphere_nudging.hpp
rebassoo: removed

components/eamxx/src/physics/nudging/nudging_functions.hpp
rebassoo: removed

components/eamxx/src/physics/nudging/nudging_functions.hpp
rebassoo: removed

components/eamxx/src/physics/nudging/atmosphere_nudging.cpp
rebassoo: removed

components/eamxx/src/physics/nudging/tests/nudging_tests.cpp
rebassoo: removed

components/eamxx/src/physics/nudging/atmosphere_nudging.cpp
rebassoo: Done

components/eamxx/src/physics/nudging/atmosphere_nudging.cpp
rebassoo: This is nice, thanks. I have updated to do this

components/eamxx/src/physics/nudging/atmosphere_nudging.cpp
rebassoo: fixed

components/eamxx/src/physics/nudging/atmosphere_nudging.cpp
rebassoo: Actually, there is a simpler way, just:
const auto& p_mid    = get_field_in("p_mid").get_view<const mPack**>();
And then p_mid_v can be just changed to p_mid everywhere

components/eamxx/src/physics/nudging/atmosphere_nudging.cpp
bartgol: I don't think this has been addressed yet.

rebassoo: I have updated this, thanks

bartgol (dismissed):

bartgol:
@jgfouca it's happening here too. Is it possible that tests not run b/c disabled would cause a non-zero exit code?

bartgol (approved):

PR #2172, opened 2023-02-14
Add U and V subfields of horiz_winds to the field manager

bartgol:
Allows to save U and V separately in the output NC files.


AaronDonahue (changes requested):
Thanks for putting this in so quickly Luca.  Just one question, would it be possible to add a unit test for u/v?  Maybe a quick run where we write `horiz_winds` and `u` and `v` and check that u=horiz_winds(0) and v=horiz_winds(1)?

bartgol:
> Thanks for putting this in so quickly Luca. Just one question, would it be possible to add a unit test for u/v? Maybe a quick run where we write `horiz_winds` and `u` and `v` and check that u=horiz_winds(0) and v=horiz_winds(1)?

That would be nice. I'm not sure how to proceed though, since it probably requires manipulating output files with NCO tools. I'll think about it.

bartgol:
@AaronDonahue Crafting a script that compares vars in two nc files (possibly subsliced) wasn't really hard, especially since some of the boilerplate netcdf stuff was already in other scripts we have.

I added U and V in the output of  shoc_standalone, and added unit tests that verify U/V are correctly subviewing the horiz_winds field. I used uppercase for the fields, but we can change to lowercase if that's what existing postprocessing scripts would expect.

components/eamxx/tests/uncoupled/shoc/CMakeLists.txt
```diff
@@ -43,3 +46,49 @@ if (TEST_RANK_END GREATER TEST_RANK_START)
...
+            WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
+  set_tests_properties (check_U_V_slices_np${RANK} PROPERTIES
+            FIXTURES_REQUIRED shoc_generate_output_nc_files)
+endforeach()
+
+################################
+#        MUST FAIL tests       #
+################################
+
+# These tests ensure that the compare-nc-files script does always pass
```
AaronDonahue, line 66: do you mean "doesn't" always pass?

AaronDonahue (approved):
@bartgol , thank you for adding the test and a tool that we can use for other unit tests.  It looks like weaver still has some issues with installing netCDF, but I approve of the changes.  Ping me to reapprove whenever the netCDF weaver stuff gets sorted out.

bartgol:
@jgfouca the weaver fail is b/c pip cannot install the netCDF4 package:

```
16:33:54 Detected missing package netCDF4. Will attempt to install locally
16:33:54 Looking in links: /tmp/tmp_o7dtcb5
16:33:54 Requirement already satisfied: setuptools in /projects/ppc64le/python/3.7.3/lib/python3.7/site-packages (40.8.0)
16:33:54 Requirement already satisfied: pip in /home/e3sm-jenkins/.local/lib/python3.7/site-packages (21.3.1)
16:33:54 ERROR: Failed to install netCDF4, cannot continue:
16:33:54 ERROR: Could not install packages due to an OSError: [Errno 2] No such file or directory: '/tmp/pip-req-tracker-uviojvk9/025205f9fe213fcace1cc098dc0aae7fd9d310f12047df50e65f43a8'
16:33:54 
16:33:54 WARNING: You are using pip version 21.3.1; however, version 23.0 is available.
16:33:54 You should consider upgrading via the '/home/projects/ppc64le/python/3.7.3/bin/python3 -m pip install --upgrade pip' command.
```
The `No such file or directory` is something I don't understand. Does it ring any bell to you?

jgfouca:
@bartgol , the ensure_* functions have been flaky ever since I added them. I have repeatedly seen this "missing /tmp" error. I will investigate.

jgfouca:
@bartgol , it looks like updating the version of pip for the user e3sm-jenkins fixed the problem. I will try to incorporate some kind of update check in ensure_pip.

AaronDonahue (approved):

PR #2174, opened 2023-02-15
Save initial state, to adjust timestamp/content of files for postprocessing

### LINKED ISSUE #2166: Write output on 1st time step to make # of samples/day constant for all days.
Currently if we ask for 6 hr output, the first output time will be 6 UTC and the name of the first output file will include 6 UTC in the timestep. This is annoying. It would be better and more standard if all output streams write their first output in timestep 0 in order to have file names start on daily boundaries and to have {0 UTC, 6 UTC, 12 UTC, 18 UTC} in that first day’s file instead of {6 UTC, 12 UTC, 18 UTC, 0 UTC on next day} in that first file. Note that ELM and maybe EAMf90 already do this. 

I think we’d also need to make sure ave, max, and min use the beginning rather than end of the time  interval as its timestamp in the filename. I’m not sure if this is already done.

Tagging @paullric because he's the one most excited about this. I still don't really understand why this is a big deal.

bartgol:
I tried to run p3_standalone with the following settings:

- start time: midnight
- timestep: 12h
- num steps: 6
- output freq: 1
- max snapshots per file: 2

The following output files were generated, with the reported times
```
p3_standalone_output.INSTANT.nsteps_x1.np4.2021-10-12-00000.nc   time = 0, 0.5 ;
p3_standalone_output.INSTANT.nsteps_x1.np4.2021-10-13-00000.nc   time = 1, 1.5 ;
p3_standalone_output.INSTANT.nsteps_x1.np4.2021-10-14-00000.nc   time = 2, 2.5 ;
p3_standalone_output.INSTANT.nsteps_x1.np4.2021-10-15-00000.nc   time = 3;
```
Furthermore, variables that were not in the IC file (or inited by P3 during init, which I don't think happens anyways) contained all zeros for the t=0 hyperslab.

@paullric is this compatible with what the postprocess scripts expect? Anything else I'm missing?

Fixes #2166 

@AaronDonahue I took the liberty of fixing a few compiler warnings. The _real_ changes are in `scream_output_manager.cpp`, and `scorpio_output.*pp`.


AaronDonahue (approved):
Looks great! I'll let @paullric respond to your question in the PR description.

bartgol:
@AaronDonahue I added quite some stuff to this PR. Here's a (hopefully comprehensive) summary.

- I broke the io.cpp test into three tests, for ease of debugging/maintainability:`io_basic` (no packs, no diags), `io_packed`, and `io_diags`.
- Only `io_basic` tests all avg options and all freq units. The other two do simple INSTANT output, since they only need to check that packs/diags work.
- The non-INSTANT output contains `time_bnds` variable, where `time_bnds(k) = a b` means that the k-th average used snaps at `t=a+dt, a+2*dt,... a+freq*dt=b`.
- The INSTANT output always outputs at t=0, possibly filling the field with 0 if not yet inited.
- The timestamp of an AVG file correspond to the `a` of the first entry of `time_bnds` (so, technically, it's a timestamp that was NOT used in the output file). This is conformal with what EAM does.
- We now restart `timestamp_of_last_write` correctly (before, we always set it to `m_case_t0`, since that was fine for what it was used, namely compute the number of steps since last write). However, now we need to use it as timestamp for the filename of the next output (see above). I think I might have retrieved this info from 'current_time - dt*freq', so maybe I can roll this back. But I guess it doesn't hurt to have the _correct_ last write timestamp.
- `Max Snapshots Per File` is now optional. If not provided (or negative), we interpret as "put all snaps in a single file like there's no tomorrow".
- Misc warnings cleanup and bug fixes.

What I want to do still, but perhaps in a second PR (what do you think?) is the following.
- Add a fill value attribute for output, configurable from yaml file.
- Pass fill value to the vremap (if present), so that it is used as mask value.
- Add a test that ensures we indeed write out zeros for avg=INSTANT if field is not initialized.

bartgol:
@AaronDonahue @PeterCaldwell The way `scream_default_output.yaml` was set up, our ERS tests would fail with this PR. The reason was that the first run (the monolithic one) would output t=0, hence, create the nc file at t=0 (hence, with -00000 in the filename), and save t=0 and t=9h snaps (2 snaps), while the restarted run (picking up at t=5h) would only write one snap (t=9h), opening the nc file at that time (hence, with -32400 in the filename). This would cause CIME to issue a FAIL, since the `xyz-00000.nc.base` file had no 'rest' counterpart, and the `xyz-32400.nc.rest` file has no 'base' counterpart.

I had a few options to make them pass:

- Change the default avg type to AVERAGE. Non-instantaneous output does not trigger a snap at t=0, so both base and rest would have their first snap at t=9h, hence creating the same `xyz-32400.nc` file.
- Add a new yaml file for output, a testmod that uses this output file, and add the testmod to our ers/erp tests.

I wend with option 1. This, however, means that if we want an INSTANT output in our production runs, we now have to manually change the output yaml file, which, tbh, is a healthy habit (defaults are not always what you need, so better always create your own yaml output file).

I think this is the best short term solution. Once this is in, I could work a bit on allowing the user to atmchange output configuration, without having to edit the yaml file and/or use a new one. This, btw, would allow me to implement option 2 above, but without having to create a new default yaml file (I would just need to create a `shell_commands` testmod, and attach it to our ERS files). But I don't know how important it is.

Finally, for all of you plus @czender , one last question (which I think we already answered at some point, but best to double check). What should the timestamp in the filename be for an average output: the beginning or the end of the 1st averaging window? E.g., if we run run 1h starting from midnight, and produce hourly average, should the file be called `xyz-YYYY-MM-DD-00000.nc` or `xyz-YYYY-MM-DD-03600.nc`? Notice that I'm only asking about the filename. The time and time_bnds in the file would be consistent with EAM, namely there would be 1 record, with `time=0.04166` and `time_bnds= 0,  0.04166` (1/24th of a day).

I think we agreed on "timestamp should reflect the beginning of the averaging window", which is what EAM does and also what EAMxx does in this PR.

czender:
@bartgol I'm glad you are double-checking what it means to name/stamp consistently with EAM. My understanding of EAM conventions is not in agreement with your description above. I am looking at a file containing thirty daily-mean snaps right now. It is the first file of the simulation that started at 2001-01-01 00:00:00. The filename is `caseid.cam.h1.2001-01-01-00000.nc`. The first timestep is instantaneous at t = 0, as we discussed on Slack. The remaining 29 records contain daily averages for Jan 1-29. The next 30 daily averages are in the file `caseid.cam.h1.2001-01-31-00000.nc`. As you can see, the files are named by the date/time at the end of the first timestep in the file. For the first file, this is the end of the instantaneous step, i.e., 2001-01-01 00:00:00. For the second file, the filename is the end of the averaging period for its first record, i.e., 2001-01-31 00:00:00. The is the end of the averaging period for calendar day Jan 30. To be consistent with EAM/CAM/CCM, name the files by the end of the period bounding the first timestep in the file. NB: Please let me know if you find exceptions to this rule. I'm just inferring the "rules" from the patterns I see in datasets close to hand, not from the source code. There may well be exceptions...

bartgol:
> @bartgol I'm glad you are double-checking what it means to name/stamp consistently with EAM. My understanding of EAM conventions is not in agreement with your description above. I am looking at a file containing thirty daily-mean snaps right now. It is the first file of the simulation that started at 2001-01-01 00:00:00. The filename is `caseid.cam.h1.2001-01-01-00000.nc`. The first timestep is instantaneous at t = 0, as we discussed on Slack. The remaining 29 records contain daily averages for Jan 1-29. The next 30 daily averages are in the file `caseid.cam.h1.2001-01-31-00000.nc`. As you can see, the files are named by the date/time at the end of the first timestep in the file. For the first file, this is the end of the instantaneous step, i.e., 2001-01-01 00:00:00. For the second file, the filename is the end of the averaging period for its first record, i.e., 2001-01-31 00:00:00. The is the end of the averaging period for calendar day Jan 30. To be consistent with EAM/CAM/CCM, name the files by the end of the period bounding the first timestep in the file. NB: Please let me know if you find exceptions to this rule. I'm just inferring the "rules" from the patterns I see in datasets close to hand, not from the source code. There may well be exceptions...

Ok, thanks. So the case you quoted is a bit of a mix (both instand and avg output). In EAMxx, we don't mix avg and instant output in the same file. So I'm wondering what EAM would do if one asked for ONLY one type of output. I have to leave for the day, but I'll re-read in the morning, perhaps analyzing more datasets.

My concern with using the end of the 1st avg window is that, without the instant output in the same file (which, from discussion on slack, I think others would not like to see in an avg output file), we would get the first file called `xyz.2001-01-02-00000.nc`, which might not be intuitive. If we keep only avg output in the file, then, in order to get the initial timestamp in the 1st file, we'd have to use the timestamp of the beg of the 1st window. I think the second option is more intuitive. E.g., a run producing files containing 30 daily avg only (no instant output), would yield `x.2001-01-01-00000.nc`, `x.2001-01-31-00000.nc',...

czender:
Sounds good. One clarification: for the EAM h1 files I described above, the user requested only daily averages. The instantaneous step is something that EAM added without being asked, presumably to make it possible to include instantaneous timeseries that would use the same time coordinate. I concur that the convention you advocate above makes more sense. I fear this means that implementing your proposal necessarily involves producing filenames that would not agree with EAM filenames for the same type of averaged output :( Good luck developing an implementation that satisfies everyone :)

PeterCaldwell:
Thanks for digging into this, Luca. A couple comments:

1. In the big picture, the point of an ERS test is to make sure a restarted run is BFB with a monolithic run. Thus behavior on the first timesnap isn't relevant. All you really need is the instantaneous state at the last timestep (or the average over the whole period, or anything depending on the state at the end of the simulation). 
2. I'd prefer to have our default output be the basic set of variables we'd like to see rather than hijack it for testing such that users need to always remember to change the output in order to be able to interpret their run... From experience, I know that even seasoned screamers often forget to do things like change the output request and it would be a bummer for that to result in totally useless runs. Thus I'd prefer ERS to have a testmod in order to write special output (if possible). 
3. I suspect EAMv2/CESM/E3SM writes that initial instantaneous snapshot in timestep 0 entirely to force the filename to reflect the start of the simulation rather than because anyone would use that data. The resulting convention is bonkers though - the first file is named by the beginning of the first averaging period and the other files are named by the end of the first averaging period. I recognize that changing things causes a lot of headaches for analysis script writers, but I think creating a new model should be seen as license to fix rather than perpetuate dumb design ideas from the past. So - no instant snap at the beginning of averaged datasets and always use the beginning of the averaging period in the file name.

bartgol:
@PeterCaldwell I can create a testmod just for the sake of ERS tests. This would make ERS test do AVG output (which does not include t=0), or maybe to normal INSTANT output but with some extra flag that disable t=0 output. The only concern, though, is that if anyone wanted to add a new ERS/ERP test in the future, they would have to remember to add this testmod.

Perhaps there's a third option: do not have any model output by default. This would a) make all ERS/ERP tests compare only cpl fields (that's what happens in EAM), and b) force all production runs to run atmchange to specify the yaml file for the output (they can of course point to `scream_default_output.yaml`).

bartgol:
> I fear this means that implementing your proposal necessarily involves producing filenames that would not agree with EAM filenames for the same type of averaged output :( Good luck developing an implementation that satisfies everyone :)

Yes, I share your fear. However, as Peter mentioned, the EAM convention seems a bit confusing, and overly complicated, so (assuming my understanding of it is right), I would be inclined to not stick to it. The only reason to stick with it would be if postprocessing tools _expect_ that convention. Put another way: do you think ncxyz postprocessing tools would work fine if we had no t=0 output, and filenames always reflecting start of first avg window?

czender:
Count me in for adopting more rational convention! I will change the NCO tools (and try to maintain backwards compatibility) so they expect the conventions according to the datasets they are analyzing (e.g., EAM vs EAMxx). Once you establish the new convention and the dust settles I can start that process.

PeterCaldwell:
>Perhaps there's a third option: do not have any model output by default.

I still want the default output to be what a user would want for a generic run rather than something weird or nothing. So I vote for testmod. I think people who add ERS tests in the future will figure out to add the testmod once their PR fails the AT and/or nightlies!

bartgol:
@PeterCaldwell ok, I think I found a way that should leave default output as is for normal runs, and make ERS/ERP pass. Let me know if that's acceptable.

Basically, inside our buildnml script, if we detect that we are in an ERS/ERP test, we change the output setting to do a single AVG step at the end of the run. This should allow the default output yaml file to remain the same whenever people create new cases, or even if they run SMS/PEM tests. This last mod only affects ERS/ERP tests.

PeterCaldwell:
Sounds good. Can we add a comment somewhere that people might actually read it pointing out that we're modifying ERS/ERP output variables?

bartgol:
> Sounds good. Can we add a comment somewhere that people might actually read it pointing out that we're modifying ERS/ERP output variables?

Yeah, I can add a comment at the top of the default output yaml file

bartgol:
Still working out a little kink regarding CIME automatic handling of restart files in ERS/ERP tests. WIP for a bit to save AT time.

bartgol:
I ran an ERP test on mappy and it passed, so this should be ready for the AT and the review.

AaronDonahue:
@bartgol , the only fails I see are DIFFs in `mappy`, which should be expected right?  Since filenames are changed in this PR.  If so, is this ready to be manually merged?

components/eamxx/data/scream_default_output.yaml
```diff
@@ -1,6 +1,7 @@
 %YAML 1.1
 ---
 Casename: ${CASE}.scream.hi
+# WARNING: ERS/ERP tets will oveerride this with AVERAGE
```
AaronDonahue, line 4: minor: "oveeride" -> "override"

components/eamxx/data/scream_default_output.yaml
```diff
@@ -58,6 +59,7 @@ Fields:
       # Diagnostics
       - PotentialTemperature
 output_control:
+# WARNING: ERS/ERP tets will oveerride this with STOP_N/STOP_OPTION
```
AaronDonahue, line 62: minor: same here

components/eamxx/src/share/io/scorpio_output.cpp
```diff
@@ -80,6 +80,10 @@ AtmosphereOutput (const ekat::Comm& comm, const ekat::ParameterList& params,
 {
   using vos_t = std::vector<std::string>;
 
+  if (params.isParameter("Fill Value")) {
+    m_fill_value = static_cast<float>(params.get<double>("Fill Value"));
```
AaronDonahue, line 84: Commenting here because code I reference isn't in the PR:

For vertical and horizontal remapping we can specify the "mask" value, currently set to std::numeric_limits<float>.  No that we allow for a fill_value can we have the mask value overridden if `m_fill_value` is set?

AaronDonahue (approved):
One comment that could be part of a separate PR if desired.

components/eamxx/src/share/io/scorpio_output.cpp
bartgol: We could. Do you want me to do it in this PR?

bartgol:
The diffs were expected. For several reasons:

- ERP/ERS tests now do avg instead of instant output
- avg timestamp reflects the start of the first avg window
- when doing INSTANT output, we output at t=0, so a) filename changes, and b) there is one more snapshot in the file

For the SMS test (which still does INSTANT output), I verified that the 2nd (and last) snapshot of some fields is BFB with the 1st (and only) snapshot in the master baseline. For ER* tests, we do a different avg type, so I can't verify anything.

bartgol:
@AaronDonahue I wend ahead and set `Fill Value` as mask for vertical remap. That's the only place where I need to specify the mask value, right?

AaronDonahue (approved):

bartgol:
The diff are expected, due to the change in filenames (for all tests) as well as what output we're checking (for ERS/ERP). Merging.

Issue #2173, opened 2023-02-14
bad dphi error with ne120pg2_ne120pg2.F2010-SCREAMv1 when using 128 vertical levels -- solved: needs 128 level IC file


ndkeen:
All previous ne120 cases I've run were using default `SCREAM_NUM_VERTICAL_LEV=72` (monogrid or trigird).
And with scream repo of Feb13th (after PR2156), I can run `ne120pg2_ne120pg2.F2010-SCREAMv1` with 72 levels, but when I try to change this to 128 levels, I'm seeing an error. This was on pm-cpu.

There are many `hommexx.errlog.*` files. One of them has:

```
login11% head f120.F2010-SCREAMv1.ne120pg2_ne120pg2.se53-feb13.gnu.5d.n043a128x1cX.pk8.L128/run/hommexx.errlog.5440.103
label: CaarFunctorImpl::run TagPreExchange
time-level 0
lat  5.894483702078075e-01 lon  6.165375582669969e+00
ie 6 igll 1 jgll 3 lev 1: bad dphi
level                   dphi                   dp3d              vtheta_dp
    0 -1.664116547164507e+03  6.500425338745119e+01  9.356703122407671e+04
    1  4.915497854467249e+03  6.500422954559328e+01  7.490565450493728e+04
    2 -7.312708120998798e+04  6.500422954559323e+01  7.749004002065660e+05
    3 -2.050754506114795e+04  7.836956977844241e+01  2.270458452057926e+05
    4 -1.090808836318721e+04  9.670791625976567e+01  1.148631115326998e+05
```

The e3sm log does not show an obvious error message, but has many warnings:
```
4006: WARNING:CAAR: dp3d too small. k=1, dp3d(k)=1.130438, dp0=65.004253
4006: WARNING:CAAR: dp3d too small. k=1, dp3d(k)=0.660907, dp0=65.004253
4006: WARNING:CAAR: dp3d too small. k=1, dp3d(k)=0.660907, dp0=65.004253

/pscratch/sd/n/ndk/e3sm_scratch/pm-cpu/se53-feb13/f120.F2010-SCREAMv1.ne120pg2_ne120pg2.se53-feb13.gnu.5d.n043a128x1cX.pk8.L128

Case that runs with 72 levels:
[...]

PR #2176, opened 2023-02-15
Add default IC for ne120 with 128 vertical levels

brhillman:
Add default initial condition for ne120 with 128 vertical levels. This can be ran the same way as ne30 L128 by using the testmod scream-L128, which changes SCREAM_NUM_VERTICAL_LEV.

[B4B]


ndkeen:
Why can't 128 levels be the default?

brhillman:
It can be, and then we could make exceptions for the cases we want L72 (i.e., I think we use ne4pg2 L72 for testing). I don't really feel strongly either way, but can easily add that to this PR.

ndkeen (approved):
I tested ne120pg2_ne120pg2 using 128 levels and the current default 72 levels. Both completed 1 month on pm-gpu. With 32 nodes, the SYPD was 6.5 for 72 levels and 4.5 with 128 levels.

brhillman:
I'm going to let this merge as-is and we can make a separate PR to change the default for all resolutions to use L128.

ndkeen:
Fixes #2173

Issue #2070, opened 2022-11-29
Use TeamVectorRange.


ambrad:
Use TeamVectorRange, which was not available when the project first started. This will clarify the threading intent in column physics (which currently uses TeamThreadRange with the assertion that vector size is 1) and remove architecture-dependent workarounds in the tridiag solver and GLL-FV remap.

tcclevenger:
@ambrad I can pick this up, but I do have some questions.

1. Is it correct that we want to use TVR instead of TTR as top level execution within a team? For all of SCREAM? Where in SCREAM is vector size asserted to be 1?
2. You bring up Gll-fv remap, is this change for Homme as well? The only thing I'm finding there that is architecture-dependent is:
```
388    // 2D -> 1D thread index space. Need to make vector dim the faster for                                                  
389    // coalesced memory access on the GPU. Kokkos doesn't expose the number of                                              
390    // threads in a team, so we have to go to the lower-level API here.                                                     
391    const int nthr_per_team =
392#if defined __CUDA_ARCH__ || defined __HIP_DEVICE_COMPILE__
393      blockDim.x,
394#else
395      1,
396#endif
397      team_niter = (niter + nthr_per_team - 1)/nthr_per_team;
398    assert(OnGpu<ExecSpace>::value || nthr_per_team == 1);
399    parallel_for(Kokkos::TeamThreadRange(team, team_niter), [&] (const int team_idx) {
400      parallel_for(Kokkos::ThreadVectorRange(team, nthr_per_team), [&] (const int vec_idx) {
401        const int k = team_idx*nthr_per_team + vec_idx;
402        if (k >= niter) return;
403        fn(k);
404      }); });
405  }
```
3. For the tridiag solver, the only parallel_fors within a team that I'm seeing is for `bfb()` calls. I see architecture-dependent code for getting the number of threads and so on, but they are only used in serial `for()` statements in `cr()` and `thomas()`. I need to understand better what is happening with these solves.. Also, a bit of a tangent, but why does homme reference `scream_tridiag.hpp` but shoc uses `ekat_tridiag.hpp`. Seems like these files do the same thing...

ambrad:
1. Yes. Note we can't use the abbrev TVR anymore since it could mean Team- or ThreadVectorRange. Let's abbreviate TeVR and ThVR. Not sure if there's an explicit assertion anywhere. That might appear only in our very old P3 micro-app.
[...]

PR #2178, opened 2023-02-15
Use TeamVectorRange for flat-loops

tcclevenger:
Use `TeamVectorRange` when no nested parallelism is needed. Updates EKAT to include similar updates. Continuation of https://github.com/E3SM-Project/scream/issues/2070.

Sanity performance test `F2010-SCREAMv1.ne30pg2_ne30pg2` on 4 summit nodes:
**Old (TeamThreadRange)**
```
name                     processes          count      walltotal  wallmax  wallmin
"a:EAMxx::physics::run"         24   1.152000e+03   1.821062e+02    8.160    7.076 
```
**New (TeamVectorRange)**
```
name                     processes          count      walltotal  wallmax  wallmin
"a:EAMxx::physics::run"         24   1.152000e+03   1.751569e+02    7.738    6.878 
```
From `wallmax`, we see a 5% decrease in runtime using `TeamVectorRange`. Maybe this is just coincidence/negligible, but they match 2 of the same simulations I did a couple of days ago, so slight performance gain in physics.

### DIFF
--- a/components/eamxx/src/diagnostics/dry_static_energy.cpp
+++ b/components/eamxx/src/diagnostics/dry_static_energy.cpp
@@ -79,7 +79,7 @@ void DryStaticEnergyDiagnostic::compute_diagnostic_impl()
     const auto& dz_s    = ekat::subview(tmp_mid,icol);
     const auto& z_int_s = ekat::subview(tmp_int,icol);
     const auto& z_mid_s = dz_s; // Reuse the memory for z_mid, but set a new variable for code readability.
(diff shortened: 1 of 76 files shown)


bartgol:
Maybe TeVR has less index arithmetic and/or registers usage than TTR?

jgfouca (approved):

ambrad (approved):
Great! Slight speedup is definitely possible for the reasons Luca gave.

bartgol (approved):

PR #2180, opened 2023-02-16
Add total precipitation diagnostic

tcclevenger:
Adding new diagnostic for for total precipitation surface flux that we want to measure for rough topo runs.

### DIFF
--- a/components/eamxx/src/diagnostics/CMakeLists.txt
+++ b/components/eamxx/src/diagnostics/CMakeLists.txt
@@ -17,6 +17,7 @@ set(DIAGNOSTIC_SRCS
   rime_water_path.cpp
   sea_level_pressure.cpp
   shortwave_cloud_forcing.cpp
+  total_precip_surf_mass_flux.cpp
   vapor_water_path.cpp
   vertical_layer_interface.cpp
   vertical_layer_midpoint.cpp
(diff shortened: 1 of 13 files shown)


PeterCaldwell:
Dumb/minor comment, but maybe we should call it `precip_total_surf_mass_flux` instead of `total_precip_surf_mass_flux` in order to be consistent with the liquid and ice precip variable names?

AaronDonahue (changes requested):
@tcclevenger , Looks great!  I'm happy to approve but I had a question about testing.  Would it be hard to add to the test a check that the new diagnostic matches the PrecipLiq and PrecipIce diagnostics added together?  IT would be a two-fer, checking that PrecipTotal is correct and that the relationship with Liq and Ice also matches our expectations.

tcclevenger:
@PeterCaldwell No problem changing the name.

@AaronDonahue Yeah, that will be a nice check. BFB may fail, I'll probably just change the way I calculate the total flux to match.

tcclevenger:
Name changed and test extended to include comparison to the sum of liq/ice diagnostics.

AaronDonahue (approved):
Thanks for adding the test @tcclevenger

AaronDonahue (dismissed):

tcclevenger:
Thanks @AaronDonahue. I'm back now, I'll start to investigate the fails.

AaronDonahue (approved):
Welcome back @tcclevenger , looks good.  Thanks for taking care of this right away.

PR #2182, opened 2023-02-17
update-all-pip: New script for updating pip packages

jgfouca:


components/eamxx/scripts/update-all-pip
```diff
@@ -0,0 +1,16 @@
...
+# dependencies, we have a set of ensure_$depname functions available
+# in utils.py. These functions will not reliably work unless the set
+# of existing python packages are relatively up-to-date. This script
+# will update all pip packages.
+#
+# Note: netcdf4 1.6.2 would not install cleanly on weaver so we had
+# to force a downgrade to 1.6.0.
+
+python3 -m pip install --upgrade pip --user
+for item in $(python -m pip list --outdated | grep -v Package | grep -v -- '-----' | cut -f 1 -d ' '); do
```
bartgol, line 14: This bash command is a bit cryptic, and I can't understand which item names we're trying to gather. Can you add a comment?

bartgol: Ok, after running the command on my laptop, I figured out what it does. But I'm puzzled: won't this also update netcdf4?

bartgol (approved):

Issue #2160, opened 2023-02-09
Output diagnostics issues (ps, PrecipLiqSurfMassFlux, PrecipIceSurfMassFlux)


golaz:
I ran a short test (1 month) on chrysalis with the latest code (b6314dad6440c284e8121cc402f9e723c5ec7640, COMPSET="F2010-SCREAMv1", RESOLUTION="ne30pg2_ne30pg2", 128 levels). Thanks to @crterai for helping me configure this run.

I noticed some issues with output variables.

**ps**

The surface pressure magnitude is way off, topping around 2052 hPa. 

![ps](https://user-images.githubusercontent.com/15743046/217918712-141f3cf6-0ed3-4626-90f2-9736da9563d5.png)

But the lowest model p_mid looks fine (topping at 1024 hPa), so this must be a diagnostic issue.

![p_mid-127](https://user-images.githubusercontent.com/15743046/217918798-d4884b0c-f925-4736-8e65-7e68948a5783.png)

**PrecipLiqSurfMassFlux**, **PrecipIceSurfMassFlux**

Values are NaNs

```
ncdump -v PrecipLiqSurfMassFlux output.scream.monthly.AVERAGE.nmonths_x1.0001-02-01-00000.nc
[...]
    NaNf, NaNf, NaNf, NaNf, NaNf, NaNf, NaNf, NaNf, NaNf, NaNf, NaNf, NaNf, 
    NaNf, NaNf, NaNf, NaNf, NaNf, NaNf, NaNf, NaNf, NaNf, NaNf, NaNf, NaNf, 
    NaNf, NaNf, NaNf, NaNf, NaNf, NaNf, NaNf, NaNf, NaNf, NaNf, NaNf, NaNf, 
    NaNf, NaNf, NaNf, NaNf, NaNf ;
}
```

Attached are the run script and output yaml to reproduce.

[run_20230209.F2010-SCREAMv1.ne30.chrysalis.sh.txt](https://github.com/E3SM-Project/scream/files/10700940/run_20230209.F2010-SCREAMv1.ne30.chrysalis.sh.txt)
[scream_output.monthly.20230209.yaml.txt](https://github.com/E3SM-Project/scream/files/10700941/scream_output.monthly.20230209.yaml.txt)

ndkeen:
When I try this setup on pm-gpu, I see fails:

```
[...]

PR #2188, opened 2023-02-24
Better handle accumulated fields

### LINKED ISSUE #1767: Reset accumulated variables in a standardized way

bartgol:
This PR adds some metadata to FieldTracking, namely the timestamp where accumulation started. Customers of the field can use this info to deduce how much time has passed, like

```
  auto t_start = tracking.get_accum_start_time();
  auto t_now = tracking.get_time_stamp();
  auto dt = t_now - t_start;
```
The driver queries the field managers for field group "ACCUMULATED", and resets them to 0 at the beginning of each time step, (setting their timestamp as well).

So far, the driver assumes the accumulation is a SUM. I thought about adding MAX/MIN, but a) I didn't know if it was useful, and b) it was not as immediate to add.

Fixes #1767 
Fixes part of #2160

### DIFF
--- a/components/eamxx/src/diagnostics/precip_ice_surf_mass_flux.cpp
+++ b/components/eamxx/src/diagnostics/precip_ice_surf_mass_flux.cpp
@@ -36,14 +36,21 @@ void PrecipIceSurfMassFluxDiagnostic::set_grids(const std::shared_ptr<const Grid
 // =========================================================================================
 void PrecipIceSurfMassFluxDiagnostic::compute_diagnostic_impl()
 {
-  const auto& precip_ice_surf_mass      = get_field_in("precip_ice_surf_mass").get_view<const Real*>();
-  const auto& precip_ice_surf_mass_flux = m_diagnostic_output.get_view<Real*>();
-  const auto dt = m_dt;
+  const auto& mass_field = get_field_in("precip_ice_surf_mass");
+  const auto& mass_view  = mass_field.get_view<const Real*>();
+  const auto& flux_view  = m_diagnostic_output.get_view<Real*>();
 
+  const auto& mass_track = mass_field.get_header().get_tracking();
+
+  const auto& t_start = mass_track.get_accum_start_time ();
+  const auto& t_now   = mass_track.get_time_stamp ();
+  const auto dt = t_now - t_start;
+
+  auto rhodt = PC::RHO_H2O*dt;
   Kokkos::parallel_for("PrecipIceMassFluxDiagnostic",
                        KT::RangePolicy(0,m_num_cols),
                        KOKKOS_LAMBDA(const Int& icol) {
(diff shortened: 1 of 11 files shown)


components/eamxx/src/share/field/field_tracking.cpp
```diff
@@ -31,4 +31,10 @@ void FieldTracking::update_time_stamp (const TimeStamp& ts) {
   }
 }
 
+void FieldTracking::set_accum_start_time (const TimeStamp& t_start) {
+  EKAT_REQUIRE_MSG (not m_time_stamp.is_valid() || m_time_stamp<=t_start,
```
AaronDonahue, line 35: will this be **not** "m_time_stamp.is_valid() or m_time_stamp<=t_start" or will it be "not m_time_stamp_is_valid()" **or** "m_time_stamp<=t_start"?  In other words, what is the `not` effecting?  I'm guessing the former, but wanted to make sure.

AaronDonahue (changes requested):
just one comment about tests, otherwise everything else looks pretty good.

components/eamxx/src/diagnostics/precip_ice_surf_mass_flux.cpp
```diff
@@ -36,14 +36,21 @@ void PrecipIceSurfMassFluxDiagnostic::set_grids(const std::shared_ptr<const Grid
...
+  const auto& t_start = mass_track.get_accum_start_time ();
+  const auto& t_now   = mass_track.get_time_stamp ();
+  const auto dt = t_now - t_start;
+
+  auto rhodt = PC::RHO_H2O*dt;
   Kokkos::parallel_for("PrecipIceMassFluxDiagnostic",
                        KT::RangePolicy(0,m_num_cols),
                        KOKKOS_LAMBDA(const Int& icol) {
-    precip_ice_surf_mass_flux(icol) = precip_ice_surf_mass(icol)/PC::RHO_H2O/dt;
+    flux_view(icol) = mass_view(icol) / rhodt;
```
AaronDonahue, line 53: Can you update the diagnostic tests which are currently calling `compute_diagnostic` by passing a `dt`?

AaronDonahue (approved):

PR #2190, opened 2023-02-24
Move MPI knowledge into EKAT and refactor cmake mach files

jgfouca:
Change list:
1) Change SCREAM_MPI* to EKAT_MPI* and handle them in EKAT
2) Remove confusing CUDA/HIP prints in CMakeLists.txt
3) Remove unused SCREAM_MPI_ERRORS_ARE_FATAL
4) MPI global configs no longer need to be passed to CUT
5) New common.cmake with common_setup macro that all machine files should use
6) Redo rank/thread spreading test. We have to make mappy go back to pre-slurm.
7) Make SCREAM_MACHINE setting a bit more robust.


components/eamxx/CMakeLists.txt
```diff
@@ -204,11 +200,6 @@ endif()
 set(SCREAM_MIMIC_GPU ${DEFAULT_MIMIC_GPU} CACHE BOOL "Mimic GPU to correctness-test inter-column parallelism on non-GPU platform")
 set(SCREAM_PACK_CHECK_BOUNDS FALSE CACHE BOOL "If defined, scream::pack objects check indices against bounds")
 set(SCREAM_TEST_DATA_DIR ${CMAKE_CURRENT_BINARY_DIR}/data CACHE PATH "Location of data files generated by tests")
-set(SCREAM_MPI_ERRORS_ARE_FATAL TRUE CACHE BOOL "Whether MPI errors should abort (default TRUE). If false, errors should be handled.")
```
bartgol, line 207: I think this got deleted by accident.

components/eamxx/cmake/machine-files/ascent.cmake
```diff
@@ -1,13 +1,15 @@
...
+common_setup()
+
 # Copied from summit.cmake, only change is to SCREAM_MACHINE
 # and remove support for CPU builds.
 #
 # Load all kokkos settings from Ekat's mach file
-set (EKAT_MACH_FILES_PATH ${CMAKE_CURRENT_LIST_DIR}/../../../../externals/ekat/cmake/machine-files)
 include (${EKAT_MACH_FILES_PATH}/kokkos/nvidia-v100.cmake)
 include (${EKAT_MACH_FILES_PATH}/kokkos/cuda.cmake)
+include (${EKAT_MACH_FILES_PATH}/mpi/other.cmake)
```
bartgol, line 10: I would add a comment here, specifying that this is sets all defaults to empty strings.

components/eamxx/cmake/machine-files/crusher.cmake
```diff
@@ -1,7 +0,0 @@
-# Load all kokkos settings from Ekat's mach file
-set (EKAT_MACH_FILES_PATH ${CMAKE_CURRENT_LIST_DIR}/../../../../externals/ekat/cmake/machine-files)
```
bartgol, line 2: Why are we removing this file?

components/eamxx/CMakeLists.txt
jgfouca: > Remove unused SCREAM_MPI_ERRORS_ARE_FATAL

components/eamxx/cmake/machine-files/crusher.cmake
jgfouca: There is no crusher machine. we have crusher-gpu and crusher-cpu machines.

components/eamxx/CMakeLists.txt
bartgol: Why did you choose to remove it?

jgfouca: It wasn't being used anywhere.

jgfouca: To clarify, the string SCREAM_MPI_ERRORS_ARE_FATAL only occurs here, so it doesn't do anything.

bartgol: No, it is used to set the default for EKAT_MPI_ERRORS_ARE_FATAL, which is used in the `ekat_comm.*pp` files. It's true we don't make use of this feature so far, but one may set `SCREAM_MPI_ERRORS_ARE_FATAL=OFF` in its config, and things would change.

jgfouca: Yeah, it's confusing. There's nothing wrong with scream builds setting ekat stuff directly. Should I restore this parameter or keep it removed?

bartgol (approved):

PR #2193, opened 2023-02-27
fix wet-dry issues, conservation

### LINKED ISSUE #2167: Incorrect `dz` calculation in P3
I mentioned this on Slack late (for me) yesterday, but I'm more confident now that this is a bug. The value of `dz` passed to P3 is calculated [by this call](https://github.com/E3SM-Project/scream/blob/f90b4e036706f1a7fbd6a67080771862d5f79074/components/eamxx/src/physics/p3/atmosphere_microphysics.hpp#L110)

```C++
        dz(icol,ipack) = PF::calculate_dz(pseudo_density(icol,ipack), pmid_pack, T_atm_pack, qv(icol,ipack));
```

where `pseudo_density` and `pmid_pack` [are set to the "dry" values](https://github.com/E3SM-Project/scream/blob/f90b4e036706f1a7fbd6a67080771862d5f79074/components/eamxx/src/physics/p3/atmosphere_microphysics.cpp#L229-L230) and `calculate_dz` [uses virtual temperature](https://github.com/E3SM-Project/scream/blob/f90b4e036706f1a7fbd6a67080771862d5f79074/components/eamxx/src/share/util/scream_common_physics_functions_impl.hpp#L303). But the virtual temperature is only correct if "wet" values are being input.

If we denote dry pseudo-density by $\Delta \widetilde{p}_d$ and dry density as $\rho_d$, then my understanding is that for each grid cell $\Delta \widetilde{p}_d = \rho_d g \Delta z$ by definition. The dry pressure can also be expressed in terms of dry density $p_d = \rho_d R_d T$. So $\Delta z = (R_d T \Delta \widetilde{p}_d) / (p_d g)$, i.e. the same formula as `calculate_dz` uses, but with actual rather than virtual temperature. (Maybe this is obvious, but I'm not used to writing things in terms of pseudo-density, so wanted to be explicit about it.)

There are a few possible ways to fix this, but the easiest would be to just pass zero rather than `qv` to `calculate_dz`:

```C++
        dz(icol,ipack) = PF::calculate_dz(pseudo_density(icol,ipack), pmid_pack, T_atm_pack, 0.);
```

All that said, I don't know that this bug would cause cause mass conservation problems. An incorrect `dz` would lead to an incorrect value calculated for `rho`, which means that the process rate and fall speed calculations will generally be off in P3. But in nearly every case, this would simply mean an error in the amount of mass transferred between categories (e.g. cloud to rain) or in the magnitude of fluxes between vertical levels, neither of which affects total mass.

In the sedimentation code, the conversion of fluxes to a change in `qc`/`qr`/`qi` relies on both `dz` and `rho`. However, in this case the errors in `dz` and `rho` should cancel each other, so this does not seem to cause problems for conservation either.

### LINKED ISSUE #2168: Inconsistent use of dry/wet variables in SHOC
Figured I'd post the other issue related to the conservation errors @oksanaguba has been tracking down.

For SHOC to conserve water mass, it really needs to conserve $\rho r_t$, where $r_t$ is total water mixing ratio and $\rho$ is dry air density if $r_t$ is a dry mixing ratio, or total density if $r_t$ is a wet mixing ratio. In particular, the net flux of water vapor across an interface due to sub-grid-scale mixing is

$$ \rho \overline{w'r'_t} $$

where $\rho$ again has to be dry mixing ratio if $r_t$ is a dry mixing ratio. However, the surface value of $\overline{w'r'_t}$ passed to SHOC is currently calculated using

```C++
      wprtp_sfc(i)  = surf_evap(i)/rrho_i(i,nlevi_v)[nlevi_p];
```

(and something equivalent in the Fortran code), where `rrho_i` is total ("wet") density at the interface, and thus inconsistent with the dry mixing ratios currently being sent to SHOC.

This flux miscalculation is only an issue at the surface, but there is another problem, which is with the turbulent advection term affecting water mass, defined as:

$$ \frac{1}{\rho} \frac{\partial \rho \overline{w'r'_t}}{\partial z} $$

Using a "wet" $\rho$ here could cause conservation violations internal to the column, wherever there is a vertical gradient of humidity, not just at the interface.

It's not really clear to me what the correct way to deal with this is. A plausible solution to this would be to simply use a "dry" `rho` everywhere in SHOC, but there might be side effects of this because of other uses of `rho` in the code. For instance, I think that changing the `rho` used in the momentum budget would amount to neglecting momentum and kinetic energy associated with water mass (would this be more or less consistent with what the dycore does?). I don't know what the effect is of changing the `rho` used in equations related to potential temperature or buoyancy, or anywhere else it might show up. It seems like it would affect energy conservation.

Another plausible solution would be to just switch everything back to wet mixing ratios, although if that's not consistent with the energy we want the model to conserve, that feels like just kicking the can down the road.

### LINKED ISSUE #2169: Inconsistency between wet MMR/dry MMR conversion functions and AD assumptions?
Adding another issue, with appreciation for @mt5555 and @oksanaguba for explaining what's going on here.

For a level pseudodensity $\Delta \widetilde{p}$, dry pseudodensity $\Delta \widetilde{p}_d$, and (wet) water vapor mixing ratio $q_v$,

$$ \Delta \widetilde{p} = \frac{\Delta \widetilde{p}_d}{1 - q_v} $$

Whenever an atmospheric process changes $q_v$, we have four options:
1) Update $\Delta \widetilde{p}$ while keeping $\Delta \widetilde{p}_d$ constant; this is effectively asserting that the process is not changing the mass of dry air in the grid cell. This is valid under a "Lagrangian" view of things where the pressure levels are allowed to shift around a bit. (Or in the case of P3, for instance, this would be consistent because the grid is basically treated as Eulerian in $z$ coordinates, and the dry mass in each grid cell is never changed by P3.)
2) Update $\Delta \widetilde{p}_d$ while keeping constant $\Delta \widetilde{p}$. This would potentially be correct for a model with an "Eulerian" grid using pressure coordinates such that the "wet" mass in each grid cell has to be constant.
3) Do something more complicated while updating both.
4) Update neither, introducing an inconsistency such that the equation above does not hold. In this case there are now two different ways of calculating the (wet or dry) mass in each grid cell, and those two methods give different answers.

Since nothing is ever easy, CAM and EAM take option 4. When using wet mixing ratios to calculate the total mass of water in a column, we therefore have to choose whether to multiply the mixing ratios by $\Delta \widetilde{p}$ or $\Delta \widetilde{p}_d/(1 - q_v)$. The "obvious", simpler choice is the former, so that's what Oksana has been using (and I'm assuming that any other conservation checks in the code do the same).

Now as an example, consider what's done by the microphysics in P3, ignoring sedimentation. (I think this problem affects the sedimentation and SHOC too, but this is the simplest example because there's no vertical transport.) The microphysical processes in P3 essentially conserve whatever they think $\Delta \widetilde{p} (q_v + q_c + q_r + q_i)$ is in each grid cell. P3 is intended to accept dry mixing ratios and dry mass, but if you input wet mixing ratios and wet mass instead, it will happily conserve that as well. The process rates will be a bit wrong if wet values are input, but no conservation laws will be broken. I believe this is why Oksana has found that P3 does successfully conserve mass if given "wet" inputs.

In a world where all our variables were consistent with each other, this would also work if P3 was given dry quantities; water mass is water mass, and it should be conserved regardless of whether you use wet or dry "units" to describe them. However, the [two](https://github.com/E3SM-Project/scream/blob/f90b4e036706f1a7fbd6a67080771862d5f79074/components/eamxx/src/share/util/scream_common_physics_functions_impl.hpp#L253) [routines](https://github.com/E3SM-Project/scream/blob/f90b4e036706f1a7fbd6a67080771862d5f79074/components/eamxx/src/share/util/scream_common_physics_functions_impl.hpp#L276) that convert between wet and dry MMR currently use $q_v$ to do the conversion. As a result, when P3 is run using "dry" quantities, and then the result is converted back to "wet" units, the quantity it conserves (expressed in wet units) is $(\Delta \widetilde{p}_d / (1 - q_v)) (q_v + q_c + q_r + q_i)$, not $\Delta \widetilde{p} (q_v + q_c + q_r + q_i)$. These two quantities should be the same, but due to the inconsistency noted above they are not, so we get a conservation violation.

As far as fixes go, I think that the best short-term fix is to change `calculate_wetmmr_from_drymmr` and `calculate_drymmr_from_wetmmr` to use `pseudo_density` and `pseudo_density_dry` to do the conversion rather than using `qv`. [This is how Fortran EAM does it.](https://github.com/E3SM-Project/E3SM/blob/1ceacbc87a183a532bfbe7c64e117e4f45b29530/components/eam/src/physics/cam/physics_types.F90#L1491-L1547) An equivalent "hack" would be to store the value of `qv` used to convert quantities to dry mixing ratios before P3, and use the same value to convert them back at the end (rather than using the new `qv` that has been updated by P3). (I guess this could also be needed for SHOC?)

Either way, it's kind of cruddy that the code works this way in the first place, and I gather that it has caused problems in the past. We shouldn't really be allowing these variables to become inconsistent with each other like this, if we can help it.

oksanaguba:
Fixing wet/dry conversion issues in shoc and p3: 
in shoc, revert MMRs to wet, in p3, use a new function for conversion. 

Using dry MMRs causes also energy errors in P3, it is fixed by using a factor in T tendency. Changes in this PR remove column energy leak messages from F and CXX code in P3. Shoc energy column errors are still present, shoc energy fixer is complicated and shoc energy leaks will be resolved in subsequent PRs.

Besides, in p3 change Exner, theta, and dz (thanks to Sean Santos and others for looking at these issues) to be computed from full wet pressure.

Fixes #2169 
Fixes #2168 
Fixes #2167

[nonbfb] and potentially climate-changing.


oksanaguba:
The current state wrt global mass diagnostics looks ok, also, there are no mass column conservation errors, but there are column energy errors from both p3 and shoc. It seems that those can be genuine, in T tendencies, not in diagnostics. (the rest is notes to myself) There seem to be no sensitivity wrt qv_prev in p3, also, no sensitivity wrt how dz is computed, which by some reason is used with pmid_dry to compute rho.

components/eamxx/src/physics/p3/atmosphere_microphysics.hpp
```diff
@@ -68,46 +68,45 @@ class P3Microphysics : public AtmosphereProcess
...
 
-        // Exner
+        // Exner from full pressure?
         const auto& exner = PF::exner_function(pmid_pack);
         inv_exner(icol,ipack) = 1.0/exner;
-        // Potential temperature
+        // Potential temperature, from full pressure?
         th_atm(icol,ipack) = PF::calculate_theta_from_T(T_atm_pack,pmid_pack);
-        // DZ
-        dz(icol,ipack) = PF::calculate_dz(pseudo_density(icol,ipack), pmid_pack, T_atm_pack, qv(icol,ipack));
```
oksanaguba, line 110: note that this call should never be used like this, computing dz from dry quantities, but using qv for Tv. also, while we can separate p (assuming only vapor as water form) and rho into dry and wet components, dz is not linear function wrt p and ro and cannot be "broken" into wet and dry.

oksanaguba:
@AaronDonahue or @bartgol  i need to change F code too, again, at the interface level of p3 and shoc. there are cxx-f bfb tests, do they test interface code?

AaronDonahue:
@oksanaguba , when you say interface code do you mean the atmosphere driver interface?  If so, then no they don't.  Only full CIME and any of the "standalone" tests that have p3 in them would test that interface.

AaronDonahue:
Will this fix #2168 and/or #2169 ?

oksanaguba:
though the last commit fixes water leak in p3, i left Tv, dz, and exner calculations to be fixed later. i now think it would be better to fix energy in p3 at the same time, to avoid questions whether these new changes make energy leaks worse. another possibility is to attach global energy diagnostics.

oksanaguba:
I am opening this for reviews/tests. Ignore changes that were only done in AT for global energy, i will revert them once this is approved. I still need to attach global energy diagn to F code to confirm it is not worse.

oksanaguba:
Status of v0 global diagnostics runs for this branch (after cherry-picking commit 1e80b6ab1f4):
```
[ac.onguba@chrlogin1 scream-b2]$ zgrep "nstep, W d" ~/ess/sky-scream-v0-ne4pg2-mar10/run/atm.log.304308.230325-203024.gz 
...
 nstep, W dflux, dstep        3   0.39623099036405783E-01   0.39669803638121226E-01  -0.46704601715442939E-04
 nstep, W dflux, dstep        4   0.31638649889857186E-01   0.31638649889854799E-01   0.23869795029440866E-14
 nstep, W dflux, dstep        5   0.38991870712564117E-01   0.39163458223310854E-01  -0.17158751074673739E-03
 nstep, W dflux, dstep        6   0.39351387258159794E-01   0.39351387258157261E-01   0.25326962749261384E-14
 nstep, W dflux, dstep        7   0.52281790466550189E-01   0.52289156446588443E-01  -0.73659800382541718E-05
 nstep, W dflux, dstep        8   0.25702057116483785E-01   0.25702057116483962E-01  -0.17694179454963432E-15
 nstep, W dflux, dstep        9   0.28817781079708477E-01   0.28822752091214002E-01  -0.49710115055251469E-05
 nstep, W dflux, dstep       10   0.83992964003951828E-02   0.84501485614041534E-02  -0.50852161008970581E-04
 nstep, W dflux, dstep       11   0.51427485529631249E-02   0.51427485529637832E-02  -0.65832755913319829E-15
[ac.onguba@chrlogin1 scream-b2]$ zgrep "nstep, E d(te)/dt-rr" ~/ess/sky-scream-v0-ne4pg2-mar10/run/atm.log.304308.230325-203024.gz 
...
 nstep, E d(te)/dt-rr        2   0.15617955703741515E+00
 nstep, E d(te)/dt-rr        3   0.15940018327735039E+00
 nstep, E d(te)/dt-rr        4   0.18637354024744113E+00
 nstep, E d(te)/dt-rr        5   0.15938202348297281E+00
 nstep, E d(te)/dt-rr        6   0.16528497494733507E+00
 nstep, E d(te)/dt-rr        7   0.69625967302747505E-01
 nstep, E d(te)/dt-rr        8   0.47704130669985467E-01
 nstep, E d(te)/dt-rr        9  -0.23470295421221143E-01
 nstep, E d(te)/dt-rr       10  -0.81117632374043680E-02
 nstep, E d(te)/dt-rr       11  -0.39519309839896266E-01
```
and run from branch oksanaguba/eam/temp-global-diagn to compare with master:
```
[ac.onguba@chrlogin1 scream-master]$ zgrep "nstep, W d" ~/ess/sky-scream-v0-global-diagn/run/atm.log.303737.230324-155312.gz 
...
 nstep, W dflux, dstep        2   0.88845177258516081E-01   0.87395789093993009E-01   0.14493881645230716E-02
 nstep, W dflux, dstep        3   0.68959526810131497E-01   0.67153290430019802E-01   0.18062363801116949E-02
 nstep, W dflux, dstep        4   0.51992519953754632E-01   0.50380859765918502E-01   0.16116601878361295E-02
 nstep, W dflux, dstep        5   0.54784182300168127E-01   0.53345281861891240E-01   0.14389004382768869E-02
 nstep, W dflux, dstep        6   0.55864859897725112E-01   0.54242280306761079E-01   0.16225795909640336E-02
 nstep, W dflux, dstep        7   0.47000588631186009E-01   0.45254369222761534E-01   0.17462194084244750E-02
 nstep, W dflux, dstep        8   0.41637506615859451E-01   0.39971180864862045E-01   0.16663257509974064E-02
 nstep, W dflux, dstep        9   0.36536051357833180E-01   0.34843380589041476E-01   0.16926707687917042E-02
 nstep, W dflux, dstep       10   0.20016699736420014E-01   0.18469636563420933E-01   0.15470631729990816E-02
 nstep, W dflux, dstep       11   0.11049641247647689E-01   0.93775782268552769E-02   0.16720630207924126E-02
 nstep, W dflux, dstep       12   0.39093892238548211E-02   0.22664466606889923E-02   0.16429425631658287E-02
 nstep, W dflux, dstep       13   0.25842243117125598E-02   0.88010734843937824E-03   0.17041169632731816E-02
[ac.onguba@chrlogin1 scream-master]$ zgrep "nstep, E d(te)/dt-rr" ~/ess/sky-scream-v0-global-diagn/run/atm.log.303737.230324-155312.gz 
...
 nstep, E d(te)/dt-rr        2  -0.61966435906638040E+00
 nstep, E d(te)/dt-rr        3  -0.49380844106951116E+00
 nstep, E d(te)/dt-rr        4  -0.42664811451162876E+00
 nstep, E d(te)/dt-rr        5  -0.41492436397235011E+00
 nstep, E d(te)/dt-rr        6  -0.45167157236956434E+00
 nstep, E d(te)/dt-rr        7  -0.58335059264946487E+00
 nstep, E d(te)/dt-rr        8  -0.56726378425247503E+00
 nstep, E d(te)/dt-rr        9  -0.63208244660355106E+00
 nstep, E d(te)/dt-rr       10  -0.60704437668561095E+00
 nstep, E d(te)/dt-rr       11  -0.65254498374951941E+00
 nstep, E d(te)/dt-rr       12  -0.70506223939286095E+00
 nstep, E d(te)/dt-rr       13  -0.75595213878888501E+00
```

Overall this branch is an improvement compared to master. Water leaks should not happen at all, but @bogensch mentioned that he knows where in clubb the issue is. The issue is also in cxx code for some time steps.

components/eam/src/physics/cam/micro_p3_interface.F90
```diff
@@ -984,38 +986,41 @@ subroutine micro_p3_tend(state, ptend, dtime, pbuf)
...
-    qv_prev_dry(:ncol,:pver) = calculate_drymmr_from_wetmmr(ncol, pver, qv_prev_wet,            qv_wet_in)
-    cldliq(:ncol,:pver)      = calculate_drymmr_from_wetmmr(ncol, pver, state%q(:,:,ixcldliq),  qv_wet_in)
-    numliq(:ncol,:pver)      = calculate_drymmr_from_wetmmr(ncol, pver, state%q(:,:,ixnumliq),  qv_wet_in)
-    rain(:ncol,:pver)        = calculate_drymmr_from_wetmmr(ncol, pver, state%q(:,:,ixrain),    qv_wet_in)
-    numrain(:ncol,:pver)     = calculate_drymmr_from_wetmmr(ncol, pver, state%q(:,:,ixnumrain), qv_wet_in)
-    ice(:ncol,:pver)         = calculate_drymmr_from_wetmmr(ncol, pver, state%q(:,:,ixcldice),  qv_wet_in)
-    qm(:ncol,:pver)          = calculate_drymmr_from_wetmmr(ncol, pver, state%q(:,:,ixcldrim),  qv_wet_in) !Aaron, changed ixqm to ixcldrim to match Kai's code
-    numice(:ncol,:pver)      = calculate_drymmr_from_wetmmr(ncol, pver, state%q(:,:,ixnumice),  qv_wet_in)
-    rimvol(:ncol,:pver)      = calculate_drymmr_from_wetmmr(ncol, pver, state%q(:,:,ixrimvol),  qv_wet_in)
+
```
PeterCaldwell, line 991: A comment explaining the algebra which makes wet->dry conversion equivalent to multiplying by dp_wet/dp_dry would be welcome here.

components/eam/src/physics/cam/micro_p3_interface.F90
```diff
@@ -1024,6 +1029,8 @@ subroutine micro_p3_tend(state, ptend, dtime, pbuf)
     ite     = state%ncol
     kts     = 1
     kte     = pver
+
+!OG do we want dry or wet pressure here?
     pres    = state%pmiddry(:,:)
     ! Initialize the raidation dependent variables.
```
PeterCaldwell, line 1035: Let's fix the "radiation" typo above ^

components/eam/src/physics/cam/micro_p3_interface.F90
```diff
@@ -1184,19 +1191,24 @@ subroutine micro_p3_tend(state, ptend, dtime, pbuf)
...
+    cldliq     (:ncol,:pver) = cldliq (:ncol,:pver) *ratio_local(:ncol,:pver)
+    numliq     (:ncol,:pver) = numliq (:ncol,:pver) *ratio_local(:ncol,:pver)
+    rain       (:ncol,:pver) = rain   (:ncol,:pver) *ratio_local(:ncol,:pver)
+    numrain    (:ncol,:pver) = numrain(:ncol,:pver) *ratio_local(:ncol,:pver)
+    ice        (:ncol,:pver) = ice    (:ncol,:pver) *ratio_local(:ncol,:pver)
+    qm         (:ncol,:pver) = qm     (:ncol,:pver) *ratio_local(:ncol,:pver)
+    numice     (:ncol,:pver) = numice (:ncol,:pver) *ratio_local(:ncol,:pver)
+    rimvol     (:ncol,:pver) = rimvol (:ncol,:pver) *ratio_local(:ncol,:pver)
+
+    !new - old
```
PeterCaldwell, line 1207: It isn't clear to me why dT would need wet/dry conversion since I don't think of temperature as being wet or dry or even having units of "per kg air". Could you add a comment to the code providing hints about what you're doing here?

components/eam/src/physics/cam/micro_p3_interface.F90
```diff
@@ -984,38 +986,41 @@ subroutine micro_p3_tend(state, ptend, dtime, pbuf)
...
 
     ! COMPUTE GEOMETRIC THICKNESS OF GRID & CONVERT T TO POTENTIAL TEMPERATURE
     !==============
     ! TODO: Create a general function to calculate Exner's formula that can be
     ! used by all parameterizations, such as P3 and SHOC.
     ! This would take a bit more work, so we have decided to delay this task
     ! until a later stage of code cleanup.
-    inv_exner(:ncol,:pver) = 1._rtype/((state%pmiddry(:ncol,:pver)*1.e-5_rtype)**(rair*inv_cp))
+
+    inv_exner(:ncol,:pver) = 1._rtype/((state%pmid(:ncol,:pver)*1.e-5_rtype)**(rair*inv_cp))
```
quantheory, line 1014: Just to clarify, is this change being made because it matters for conservation, or is it just because using dry mass for the exner function doesn't make much sense in the first place. I think this change looks good, just want to understand.

components/eam/src/physics/cam/micro_p3_interface.F90
```diff
@@ -984,38 +986,41 @@ subroutine micro_p3_tend(state, ptend, dtime, pbuf)
...
        do k = 1,pver
           ! Note, there is a state%zi variable that could be used to calculate
           ! dz, but that is in a wet coordinate frame rather than dry.  Now that
           ! P3 is using dry MMR we instead calculated dz using virtual
           ! temperature and pressure.
-          T_virtual  = state%t(icol,k) * (1.0 + qv_dry(icol,k)*(1.0*mwdry/mwh2o - 1.0))
-          dz(icol,k) = (rair/gravit) * state%pdeldry(icol,k) * T_virtual / state%pmiddry(icol,k) 
+
+          T_virtual  = state%t(icol,k) * (1.0 + state%q(icol,k,1)*(mwdry/mwh2o - 1.0))
+          dz(icol,k) = (state%zi(icol,k) - state%zi(icol,k+1))/gravit
```
quantheory, line 1023: Since `T_virtual` is not being used to calculate `dz`, it doesn't seem to be used at all, so I think you should just remove `T_virtual` completely.

PeterCaldwell:
Should we delete the `calculate wet_mmr_from_dry` function and its inverse now that we use the new the `dp` versions of these functions?

components/eam/src/physics/cam/micro_p3_interface.F90
```diff
@@ -1024,6 +1029,8 @@ subroutine micro_p3_tend(state, ptend, dtime, pbuf)
     ite     = state%ncol
     kts     = 1
     kte     = pver
+
+!OG do we want dry or wet pressure here?
```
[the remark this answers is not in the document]
quantheory: I don't think that this is related to any bigger changes, because this variable I think is only used within P3's parameterizations. I looked into this and the answer seems to be, weirdly, that wet pressure would be correct here.

The reason is that, while we do want `pdeldry` to be passed in as `dpres` so that P3 is calculating densities in terms of dry air, `pres` is actually only used for two things, the calculation of water vapor diffusivity (`dv`) and as an argument for `qv_sat`. For `dv`, I don't think that that wet vs. dry matters, though my guess is that this formula comes from an empirical fit using measurements of total (wet) pressure. However, `qv_sat` is weird, because it requires a wet pressure as an input, but produces a dry mixing ratio as an output. You can see that [here](https://github.com/E3SM-Project/scream/blob/c814965fe67b7c307d5c39b615430781d1afca67/components/eam/src/physics/cam/wv_sat_scream.F90#L51), where the water vapor pressure is being subtracted from the input pressure (which only makes sense if the input is wet pressure). And the formula is correct if you are calculating dry saturation mixing ratio, from a wet pressure.

So wet pressure is good for both of the uses of `pres`, but it would be weird to pass both wet pressure and `pdeldry`. If this `qv_sat` is only used by P3, I would suggest that we modify it to accept dry rather than wet pressure, and then continue to pass only dry pressure variables to P3. At least, to me that feels like the least obnoxious solution.

components/eamxx/src/physics/p3/atmosphere_microphysics.cpp
```diff
@@ -275,8 +282,11 @@ void P3Microphysics::initialize_impl (const RunType /* run_type */)
   }
   diag_inputs.ni_activated    = get_field_in("ni_activated").get_view<const Pack**>();
   diag_inputs.inv_qc_relvar   = get_field_in("inv_qc_relvar").get_view<const Pack**>();
+
+//OG why is this not using p3_preproc?
+//OG do we want dry or wet pressure here?
   diag_inputs.pres            = get_field_in("p_dry_mid").get_view<const Pack**>();
```
quantheory, line 288: This has the same problem as the Fortran code I think; because `qv_sat` wants a wet pressure, passing a dry pressure here is a bit wrong, even though `dpres` *does* need to be dry.

components/eam/src/physics/cam/micro_p3_interface.F90
oksanaguba: This does not affect conservation, but yes, "dry" Exner does not make sense to me here.

oksanaguba:
> Should we delete the `calculate wet_mmr_from_dry` function and its inverse now that we use the new the `dp` versions of these functions?

The old function is used in tests, like below, it would take longer to clean that up.
```
./eamxx/src/share/tests/common_physics_functions_tests.cpp:  drymmr0 = PF::calculate_drymmr_from_wetmmr(wetmmr0,qv_wet0);//get drymmr from wetmmr0 using qv_wet0
./eamxx/src/share/tests/common_physics_functions_tests.cpp:  qv_dry0 = PF::calculate_drymmr_from_wetmmr(qv_wet0, qv_wet0);
```

components/eam/src/physics/cam/micro_p3_interface.F90
oksanaguba: thanks, i added a comment, though tbh it seems to be a trivial thing to explain.

components/eam/src/physics/cam/micro_p3_interface.F90
oksanaguba: thanks, done

components/eam/src/physics/cam/micro_p3_interface.F90
oksanaguba: Thanks, i added a comment.

quantheory:
@oksanaguba Actually, what you would need to do to accept dry pressure is just to get rid of the subtraction in the denominator. So changing the `qv_sat` assignment like this would be sufficient:
```f90
qv_sat = ep_2*e_pres/p_atm
```
I believe that if you make this change, then passing `pmiddry` to P3 should be fine.

In the C++ version (defined in `src/physics/share/physics_saturation_impl.hpp`), it's more complicated, because `qv_sat` is called by the relative humidity diagnostic as well as P3. For P3, you could correct the formula to use dry pressure like in the Fortran code, and it would be fine:
```c++
return ep_2 * e_pres / p_atm;
```
However, the relative humidity diagnostic (defined in `src/diagnostics/relative_humidity.cpp`) seems to use all "wet" variables. That means that there's an entirely separate bug, where it is expecting the *output* of `qv_sat` to be wet, but in the current code it is dry. Fixing that would require a different, incompatible change to `qv_sat`:
```c++
return ep_2 * e_pres / (ep_2 * e_pres + p_atm - e_pres)
```
This latter choice is (I think) the one most often used by the old CAM5 code and by MG2, because it is correct for wet mixing ratios.

Maybe the correct change in the C++ version is to split `qv_sat` into separate functions `qv_sat_wet` (used for diagnostic code) and `qv_sat_dry` (used for P3)? At least it would force users to think about which version they actually want.

oksanaguba:
@PeterCaldwell can we have a different PR to sort out issues with qsat? Or do you want it done in this PR?

oksanaguba:
@quantheory thanks.

PeterCaldwell:
> @PeterCaldwell can we have a different PR to sort out issues with qsat? Or do you want it done in this PR?

Either way is totally fine with me.

components/eam/src/physics/cam/micro_p3_interface.F90
```diff
@@ -984,38 +985,41 @@ subroutine micro_p3_tend(state, ptend, dtime, pbuf)
...
+    !The conversion is done via calculation drymmr = (wetmmr * wetdp) / drydp
+    ratio_local(:ncol,:pver) = state%pdel(:ncol,:pver)/state%pdeldry(:ncol,:pver)
+
+    qv_dry     (:ncol,:pver) = state%q(:ncol,:pver,         1) *ratio_local(:ncol,:pver)
+    cldliq     (:ncol,:pver) = state%q(:ncol,:pver,  ixcldliq) *ratio_local(:ncol,:pver)
+    numliq     (:ncol,:pver) = state%q(:ncol,:pver,  ixnumliq) *ratio_local(:ncol,:pver)
+    rain       (:ncol,:pver) = state%q(:ncol,:pver,    ixrain) *ratio_local(:ncol,:pver)
+    numrain    (:ncol,:pver) = state%q(:ncol,:pver, ixnumrain) *ratio_local(:ncol,:pver)
+    ice        (:ncol,:pver) = state%q(:ncol,:pver,  ixcldice) *ratio_local(:ncol,:pver)
+    !Aaron, changed ixqm to ixcldrim to match Kai's code
```
AaronDonahue, line 1000: thanks for noting that, I think you can remove this comment now that I have seen it.

components/eam/src/physics/cam/micro_p3_interface.F90
```diff
@@ -1184,19 +1190,27 @@ subroutine micro_p3_tend(state, ptend, dtime, pbuf)
...
+    rain       (:ncol,:pver) = rain   (:ncol,:pver) *ratio_local(:ncol,:pver)
+    numrain    (:ncol,:pver) = numrain(:ncol,:pver) *ratio_local(:ncol,:pver)
+    ice        (:ncol,:pver) = ice    (:ncol,:pver) *ratio_local(:ncol,:pver)
+    qm         (:ncol,:pver) = qm     (:ncol,:pver) *ratio_local(:ncol,:pver)
+    numice     (:ncol,:pver) = numice (:ncol,:pver) *ratio_local(:ncol,:pver)
+    rimvol     (:ncol,:pver) = rimvol (:ncol,:pver) *ratio_local(:ncol,:pver)
+
+    !compute temperature tendency as calculated by P3
+    dtemp(:ncol,:pver) = th(:ncol,:pver)/inv_exner(:ncol,:pver) - state%t(:ncol,:pver)
+    !rescale temperature tendency to conserve entahly:
```
AaronDonahue, line 1208: is this supposed to be "enthalpy"

components/eamxx/src/physics/p3/atmosphere_microphysics.cpp
```diff
@@ -275,8 +282,11 @@ void P3Microphysics::initialize_impl (const RunType /* run_type */)
   }
   diag_inputs.ni_activated    = get_field_in("ni_activated").get_view<const Pack**>();
   diag_inputs.inv_qc_relvar   = get_field_in("inv_qc_relvar").get_view<const Pack**>();
+
+//OG why is this not using p3_preproc?
```
AaronDonahue, line 286: I think for simplicity, since we can use the field pointer directly for this value there's no need to store a copy in the preproc.

components/eamxx/src/share/atm_process/atmosphere_process_group.cpp
```diff
@@ -340,6 +348,80 @@ void AtmosphereProcessGroup::run_impl (const double dt) {
...
+
+ using C = scream::physics::Constants<Real>;
+  constexpr Real gravit = C::gravit;
+  constexpr Real Pi = C::Pi;
+  std::setprecision(20);
+
+  auto& c = scream::ScreamContext::singleton();
+  auto ad = c.getNonConst<scream::control::AtmosphereDriver>();
+  const auto gn = "Physics";
+  //const auto gn = "Physics GLL";
```
AaronDonahue, line 363: delete?

components/eamxx/src/share/atm_process/atmosphere_process_group.cpp
```diff
@@ -340,6 +348,80 @@ void AtmosphereProcessGroup::run_impl (const double dt) {
...
+
+//lets find all grids //does not work for grid_name
+//  const auto fms = ad.m_field_mgrs;
+//  for (const auto& it : ad.m_field_mgrs) {
+//      const auto& grid_name = it.first;
+//  m_atm_logger->info("OG find all grids: "+std::to_string(grid_name));
+//  }
+
+
+#if 0
```
AaronDonahue, line 376: delete?

components/eamxx/src/share/atm_process/atmosphere_process_group.cpp
```diff
@@ -340,6 +348,80 @@ void AtmosphereProcessGroup::run_impl (const double dt) {
...
+  const auto qc_ptr             = phys_field_mgr->get_field_ptr("qc");
+  const auto qr_ptr             = phys_field_mgr->get_field_ptr("qr");
+  const auto qi_ptr             = phys_field_mgr->get_field_ptr("qi");
+  const auto vapor_flux_ptr     = phys_field_mgr->get_field_ptr("vapor_flux");
+  const auto water_flux_ptr     = phys_field_mgr->get_field_ptr("water_flux");
+  const auto ice_flux_ptr       = phys_field_mgr->get_field_ptr("ice_flux");
+  const auto heat_flux_ptr      = phys_field_mgr->get_field_ptr("heat_flux");
+#endif
+
+  //auto ff = fm->get_field("ps");
```
AaronDonahue, line 392: delete?

components/eamxx/src/share/atm_process/atmosphere_process_group.cpp
```diff
@@ -340,6 +348,80 @@ void AtmosphereProcessGroup::run_impl (const double dt) {
...
+  m_atm_logger->info("OG   size of real = "+std::to_string(sizeof(Real)));
+
+  Real aaa = 0.0;
+  for (int ii = 0; ii < ncols; ii++){
+       aaa+= area(ii);
+  }
+
+printf("OG area is %.20f \n",aaa);
+
+#if 0
```
AaronDonahue, line 408: delete?

AaronDonahue (commented):
Looks good, there seem to be a lot of commented out lines or `#ifdef 0` lines in the components/eamxx/src/share/atm_process/atmosphere_process_group.cpp file that should be taken care of before we can merge this PR.

oksanaguba:
Thanks, @AaronDonahue  -- those lines that you mentioned, most of them are from temp. commits for energy/mass diagnostics. I will revert them before this goes in.

oksanaguba:
Some notes for my own benefit:
Dry qsat is given by (assume anything with superscript v is at saturation point)
```
qsat_dry = \rho^v/\rho^d = (p^d/Rd/T) / (p^v/Rv/T) = ep2 * p^v / p^d
```
Using the same idea, wet qsat is given by
```
qsat_wet = \rho^v/ (\rho^v+\rho^d) = //some algebra // = Rd*p^v/(Rv*p^d + Rd*p^v)
```
which is the same CAM5 formula as @quantheory provided after some algebra (ev2 = Rd/Rv). 
Though while we have const moist pressure assumption, i would find using
```
qsat_wet = qsat_dry * pdeldry / pdelwet
```
easier to use, because CAM5 formula uses "updated" moist pressure, while the interface does not update it.

I think i will try to switch to new names, _wet and _dry and see where instances of qsat appear in the code.

@quantheory please, let me know if you have comments, thanks.

oksanaguba:
An update: two commits above introduce wet and dry qsat functions as was recommended by Sean and dry version is used in p3. I ran both v0 and v1 and did not see column checks leaks. I switched to fixing standalone tests (they use qsat, too), but there are conflicts with global diagnostics. I need to revert global diagnostics commits. So I looked at the global diagnostics again and noticed that in v0 there are still significant water leaks at random steps (but not at column levels). There seem to be no leaks in v1 (global or local). I will try to chase these global leaks in v0. It is a rather tedious process.

PeterCaldwell:
> I will try to chase these global leaks in v0. It is a rather tedious process.

Since we want to abandon SCREAMv0 in a few months, I don't think that fixing its leaks should be a high priority. Is this needed in order to pass BFB tests or something?

oksanaguba:
No, there are no tests for global leaks at the moment.

PeterCaldwell:
Ok, let's not worry about fixing SCREAMv0 then.

oksanaguba:
Regarding fails in src/physics/share/tests/physics_saturation_unit_tests.cpp -- these tests compare sauration values with pre-recorded values, so when i replaced qv_sat with qv_sat_dry, it caused fails. replacing the call with wet version won't help either, since the only version that matches recorded values was removed per discussion above. i am not sure what to do in this case, @PeterCaldwell ? thanks!

PeterCaldwell:
@oksanaguba - if I understand correctly, you're getting fails due to the blocks of data in https://github.com/E3SM-Project/scream/blob/9827aa35ca24e011fcb9d2d14372bbd26d95ecb6/components/eamxx/src/physics/share/tests/physics_saturation_unit_tests.cpp#L295 (for example). These blocks consist of packs of 8 entries of the form {Flatau ice SVP, Flatau liq SVP, Flatau ice qsat, Flatau liq qsat, MK ice SVP, MK liq SVP, MK ice qsat, MK liq qsat}. 

Note that changing qsat from wet to dry shouldn't affect SVP (=saturation vapor pressure) so shouldn't cause fails. Can you confirm that the fails you're getting are for saturation *mixing ratio*?

These values are calculated in https://github.com/E3SM-Project/scream-docs/blob/master/analysis-scripts/test_qv_sat.py. When I run this code, it is giving slightly different answers and only gives 4 of the needed 8 entries. I've added the task of figuring out why to my todo list. If/when this code is providing correct answers, getting new data will just be a matter of changing the qsat equation in this python code to get consistent answers. Otherwise, if you're sure your impl is correct you could just print out these values from when you actually run the test in C++ and copy them into the test script.

oksanaguba:
the test uses qsat:
```
    //last argument "0" of qv_sat function below forces qv_sat to call "polysvp1"
    const Spack mix_ice_fr = physics::qv_sat(temps, pres, true, Smask(true), physics::Polysvp1);
    const Spack mix_liq_fr = physics::qv_sat(temps, pres, false,Smask(true), physics::Polysvp1);
```
and
```
    // Now check that computed vs expected values are small enough (Flatau).
    if (std::abs(mix_ice_fr[0] -  expected_mix_ice_fr) > Cond_ice_r*tol ) {
      printf("qsi_fp: abs(calc-expected)=%e %e\n",std::abs(mix_ice_fr[0] -  expected_mix_ice_fr),tol*Cond_ice_r);
      errors++;
    }
```
etc. So it is not only about saturation values.

PeterCaldwell:
I don't understand the last comment. It's true that the code tests both qv_sat and polysvp. My point is that the stuff testing just polysvp should still pass and the stuff testing qv_sat - which uses polysvp (which is still correct) _and_ SVP->qsat conversion (which has changed) should fail. 

The example you show above seems to just be testing qv_sat.

oksanaguba:
Here is a set of fails:
```
65: qsi_fp: abs(calc-expected)=2.338062e-05 4.993094e-11
65: qsl_fp: abs(calc-expected)=2.338062e-05 4.405308e-11
65: qsi_mkp: abs(calc-expected)=2.337853e-05 4.993094e-11
65: qsl_mkp: abs(calc-expected)=2.337853e-05 4.405308e-11
65: qsi_fp: abs(calc-expected)=8.996424e-08 5.609145e-11
65: qsl_fp: abs(calc-expected)=1.620640e-07 4.948838e-11
65: qsi_mkp: abs(calc-expected)=8.990428e-08 5.609145e-11
65: qsl_mkp: abs(calc-expected)=1.614486e-07 4.948838e-11
65: qsi_fp: abs(calc-expected)=1.170588e-03 4.498973e-11
65: qsl_fp: abs(calc-expected)=1.170588e-03 3.969355e-11
65: qsi_mkp: abs(calc-expected)=1.171502e-03 4.498973e-11
65: qsl_mkp: abs(calc-expected)=1.171502e-03 3.969355e-11
65: qsl_fp: abs(calc-expected)=1.001323e-10 5.730047e-11
65: qsl_mkp: abs(calc-expected)=9.464188e-11 5.730047e-11
65: qsi_fp: abs(calc-expected)=4.630267e-08 5.682765e-11
65: qsl_fp: abs(calc-expected)=8.879397e-08 5.013791e-11
65: qsi_mkp: abs(calc-expected)=4.627367e-08 5.682765e-11
65: qsl_mkp: abs(calc-expected)=8.827862e-08 5.013791e-11
65: qsi_fp: abs(calc-expected)=2.341471e-05 4.992911e-11
65: qsl_fp: abs(calc-expected)=2.341471e-05 4.405147e-11
65: qsi_mkp: abs(calc-expected)=2.341264e-05 4.992911e-11
65: qsl_mkp: abs(calc-expected)=2.341264e-05 4.405147e-11
65: qsi_fp: abs(calc-expected)=8.058991e-04 4.546212e-11
65: qsl_fp: abs(calc-expected)=8.058991e-04 4.011033e-11
65: qsi_mkp: abs(calc-expected)=8.065296e-04 4.546212e-11
65: qsl_mkp: abs(calc-expected)=8.065296e-04 4.011033e-11
65: qsi_fp: abs(calc-expected)=3.599939e-07 5.609145e-11
65: qsl_fp: abs(calc-expected)=6.485873e-07 4.948838e-11
65: qsi_mkp: abs(calc-expected)=3.597539e-07 5.609145e-11
65: qsl_mkp: abs(calc-expected)=6.461236e-07 4.948838e-11
```
while the block with sat pressure ("esi_"... etc) is not reported (so prob pass). 

Sorry i misunderstood your 1st message.

oksanaguba:
I looked at the python script, there are a couple of obvious places where to change it, i will try and see if it works. 

Separately, while we have constant moist pressure assumption in physics, it seems to me that ambiguity about how to do phase changes will remain.

oksanaguba:
@PeterCaldwell i spent some time with python and i would need your help, please. As you said, it prints 4 numbers, not 8. It has only 2 commits, and there is no version with 8 numbers. In scream test, all 8 numbers are used and there is a hint that they are MK-related numbers:
```
    //extract values from the array
    const auto expected_sat_ice_fp = expected_vals[0];
    const auto expected_sat_liq_fp = expected_vals[1];
    const auto expected_mix_ice_fr = expected_vals[2];
    const auto expected_mix_liq_fr = expected_vals[3];

    const auto expected_sat_ice_mkp = expected_vals[4];
    const auto expected_sat_liq_mkp = expected_vals[5];
    const auto expected_mix_ice_mkr = expected_vals[6];
    const auto expected_mix_liq_mkr = expected_vals[7];
```
So I printed all available possible numbers in python like this:
```
    print('  Data to copy Flatau: ',e_ice,e_liq,qsat_i,qsat_l )
    print('  Data to copy MK: ',e_ice_MK,e_liq_MK,qsat_i_MK,qsat_l_MK )
    print('  Data to copy GG: ',e_ice_GG,e_liq_GG,qsat_i_GG,qsat_l_GG )
    print('  Data to copy exact: ',qsat_exact_i,qsat_exact_l )
```
but i am not getting a match for the last 4 numbers in this entry in cxx:
```
      stargs[0] = {tmelt,
                   atm_pres,
                   //dp_data
                  {611.23992100000009, 611.23992100000009,
                   0.0038251131382843278, 0.0038251131382843278,
                   611.2126978267946, 611.2126978267946,
                   0.0038249417291628678, 0.0038249417291628678},
                   //sp_data
                  {611.2399, 611.2399, 0.0038251132, 0.0038251132,
                   611.2123, 611.2123, 0.003824939, 0.003824939}
      };
```
Instead i am getting (see how the 1st 4 numbers match)
```
  Data to copy Flatau:  611.2399210000001 611.2399210000001 0.0038251131382843278 0.0038251131382843278
  Data to copy MK:  611.1535914292243 611.2126978267946 0.003824569569503023 0.0038249417291628678
  Data to copy GG:  610.2072697609785 610.3360999334138 0.0038186111757935247 0.003819422331873977
  Data to copy exact:  0.003825113138284327 0.003825113138284327
```

thanks!

oksanaguba:
Unrelated to that, but related to the unit-test issue, I believe the only change that is needed in python is to replace the 1st line with the 2nd line
```
    mix_ratio = scal_fact*vap_pres/(pres - vap_pres)
    #mix_ratio = scal_fact*vap_pres/pres
```
and then of course regenerate all numbers. There is also a minor python2 vs python3 issue, i can push it together.

PeterCaldwell:
Recap: one piece of Oksana's PR is to change qvsat from producing wet saturation mixing ratios to dry mixing ratios. This broke unit tests which were comparing against hard-coded values for certain cases. Those hard-coded values seem to have been taken from an old python script mimicking the calculations in our C++.

Thoughts:
1. Comparing against hard-coded values is really a primitive form of a regression test: if the answer changes, the test will fail. It would be much slicker and more reasonable to have this test create a baseline using master, then compare against it (like our run_and_cmp codes do). Is that hard to do in this case? Note that the hard-coded values approach originally made some sense because, at least for T=273.15 K using the Flatau scheme, the saturation pressure could be read directly out of the publication describing our qvsat calculation. We've since switched from the Flatau to the Murphy-Koop impl, so this awesome direct connection to the original scheme is lost.
2. I don't think we use the Flatau scheme any more, but it's still in the code base and we still test it. Murphy-Koop (MK) is always more accurate, but is a bit slower. If I recall correctly, MK is negligibly slower for full-model tests. Should we just delete all mention of Flatau?

oksanaguba:
@PeterCaldwell  thanks. Obviously, I do not have any opinion on any of this, so it would be nice if anyone else comment. From my primitive perspective, I do not see much in using that test. It seems to me it compares a function value as coded in cxx with the same value as coded in python, except that python is using exact functions and integrated functions. Maybe i am missing what exactly is going on.

quantheory:
I would suggest keeping Flatau in, partly because it shouldn't be a lot of code, and partly because I do think there are some potential uses. People do try switching schemes occasionally just to see how sensitive results are to a change. Also, Flatau might be significantly cheaper in an iterative method that has to call the SVP function a lot per time step.

PeterCaldwell:
> I would suggest keeping Flatau in, partly because it shouldn't be a lot of code, and partly because I do think there are some potential uses. People do try switching schemes occasionally just to see how sensitive results are to a change. Also, Flatau might be significantly cheaper in an iterative method that has to call the SVP function a lot per time step.

Ok.

oksanaguba:
With RH tests fixed again (it was comparing zeros and it wasn't bfb), it seems there are only fails from saturation. I will wait for suggestions how to approach saturation test. I will also check if i can access weaver report.

oksanaguba:
So far i see nothing alarming on weaver. Once we clear saturation test, it will be easier to browse outputs.

AaronDonahue:
I think the plan for saturation is for us to make a baseline comparison test based on the current master branch of the code.   The plan from last Thursdays telecom was for @jgfouca to take this on.  Jim, let us know if you need any guidance on how this test should work.

jgfouca:
@AaronDonahue , yes, @PeterCaldwell gave me guidance over slack. I will start work on this soon.

components/eam/src/physics/p3/scream/micro_p3_interface.F90
```diff
@@ -1184,19 +1190,27 @@ subroutine micro_p3_tend(state, ptend, dtime, pbuf)
...
+    rain       (:ncol,:pver) = rain   (:ncol,:pver) *ratio_local(:ncol,:pver)
+    numrain    (:ncol,:pver) = numrain(:ncol,:pver) *ratio_local(:ncol,:pver)
+    ice        (:ncol,:pver) = ice    (:ncol,:pver) *ratio_local(:ncol,:pver)
+    qm         (:ncol,:pver) = qm     (:ncol,:pver) *ratio_local(:ncol,:pver)
+    numice     (:ncol,:pver) = numice (:ncol,:pver) *ratio_local(:ncol,:pver)
+    rimvol     (:ncol,:pver) = rimvol (:ncol,:pver) *ratio_local(:ncol,:pver)
+
+    !compute temperature tendency as calculated by P3
+    dtemp(:ncol,:pver) = th(:ncol,:pver)/inv_exner(:ncol,:pver) - state%t(:ncol,:pver)
+    !rescale temperature tendency to conserve entahly:
```
whannah1, line 1208: just noticed this typo "entahly" => "enthalpy"

PeterCaldwell:
@oksanaguba - Note that #2301 just went in, which fixes the saturation tests - I think that unblocks this PR from being integrated.

oksanaguba:
@PeterCaldwell thanks, i merged master into this branch and let tests run. i assume the new saturation test will need changes in this PR, will look tomorrow.

bartgol (dismissed):
From the AD point of view, this seems fine. I'll let other atm folks talk about the physical correctness.

oksanaguba:
On mappy
```
	 21 - physics_saturation_run_and_cmp (Failed)
	 25 - p3_run_and_cmp_cxx (Failed)
	 26 - p3_run_and_cmp_f90 (Failed)
	 23 - physics_saturation_run_and_cmp (Failed)
	 27 - p3_run_and_cmp_cxx (Failed)
	 28 - p3_run_and_cmp_f90 (Failed)
```
My understanding is that all are baseline tests. I scanned the log and did not see any crashes.

oksanaguba:
On blake the report was
```
/home/e3sm-jenkins/blake/workspace/SCREAM_PullRequest_Autotester_Blake/4485
[onguba@blake 4485]$ grep Faile JENKINS_2023-05-04_160945 
 33/228 Test  #72: physics_saturation_run_and_cmp .....***Failed    4.27 sec
 48/305 Test  #80: physics_saturation_run_and_cmp ............................***Failed    2.68 sec
 75/294 Test #170: relative_humidity .........................................***Failed    3.25 sec
210/228 Test #106: p3_run_and_cmp_cxx_omp16 ...........***Failed    7.53 sec
211/228 Test #107: p3_run_and_cmp_f90_omp16 ...........***Failed    2.86 sec
	 72 - physics_saturation_run_and_cmp (Failed)
	106 - p3_run_and_cmp_cxx_omp16 (Failed)
	107 - p3_run_and_cmp_f90_omp16 (Failed)
 57/304 Test  #80: physics_saturation_run_and_cmp ............................***Failed    2.90 sec
141/304 Test  #43: coarsening_remapper_np2 ...................................***Failed    4.05 sec
281/305 Test #114: p3_run_and_cmp_cxx_omp16 ..................................***Failed    5.59 sec
	 43 - coarsening_remapper_np2 (Failed)
	 80 - physics_saturation_run_and_cmp (Failed)
282/305 Test #115: p3_run_and_cmp_f90_omp16 ..................................***Failed    2.49 sec
	 80 - physics_saturation_run_and_cmp (Failed)
	114 - p3_run_and_cmp_cxx_omp16 (Failed)
	115 - p3_run_and_cmp_f90_omp16 (Failed)
	170 - relative_humidity (Failed)
```
with RH failing with FPE . I think i fixed it (ran on blake manually). The other fail on blake, which is run crash, is
coarsening_remapper_np2 from release

```
   -> Checking field with layout <COL,CMP,LEV>(10,3,20) ... OK!
   -> Checking field with layout <COL,CMP,ILEV>(10,3,21) ..
   -> Checking field with layout <COL,CMP,ILEV>(10,3,21) .. OK!
check tgt fields ... done!
^[[1;31m^[[0m^[[1;33m^[[0m^[[1;32m===============================================================================^[[0m
^[[1;32mAll tests passed^[[0m (59 assertions in 2 test cases)

[blake08:105777] *** An error occurred in MPI_Reduce
[blake08:105777] *** reported by process [4089577473,0]
[blake08:105777] *** on communicator MPI COMMUNICATOR 3 DUP FROM 0
[blake08:105777] *** MPI_ERR_TRUNCATE: message truncated
[blake08:105777] *** MPI_ERRORS_ARE_FATAL (processes in this communicator will now abort,
[blake08:105777] ***    and potentially your MPI job)
```
I do not see such fail in master, maybe, I need to merge master into this branch.

oksanaguba:
Regarding leaks -- as before, this branch does not show mass leaks in v1 runs (exception is 1 timestep due to shoc) with my 1-rank diagnostics (which won't be usable much longer, since master has some incompatible changes in it now), and energy leaks are coming only from shoc.

```
 1017  zgrep "OG phy" ~/ess/sky-scream-v1-ne4pg2-apr28/run/e3sm.log.324822.230501-141712 
 1018  zgrep "OG mic" ~/ess/sky-scream-v1-ne4pg2-apr28/run/e3sm.log.324822.230501-141712 
 1019  zgrep "OG mac" ~/ess/sky-scream-v1-ne4pg2-apr28/run/e3sm.log.324822.230501-141712 
 1020  vi ~/ess/sky-scream-v1-ne4pg2-apr28/run/atm.log.324822.230501-141712 
 1021  zgrep "name: shoc" ~/ess/sky-scream-v1-ne4pg2-apr28/run/atm.log.324822.230501-141712 | wc -l
 1022  zgrep "nergy rela" ~/ess/sky-scream-v1-ne4pg2-apr28/run/atm.log.324822.230501-141712 | wc -l
 1023  zgrep "Warning" ~/ess/sky-scream-v1-ne4pg2-apr28/run/atm.log.324822.230501-141712 | wc -l
 1024  vi ~/ess/sky-scream-v1-ne4pg2-apr28/run/atm.log.324822.230501-141712 
```

oksanaguba:
Status of this PR for now -- if blake and weaver tests after the last commit look like above except for 2 blake tests, then this is ready.

oksanaguba:
Looking at weaver and blake AT logs, only _cmp tests fail and i see no crashes in logs. This is ready. @bartgol and/or @AaronDonahue can you approve this? thanks!

bartgol (approved):
@AaronDonahue @PeterCaldwell  you should weigh in regarding the actual physics of this test, and if all is good, merge. From the AD point of view, this is good.

components/eamxx/src/physics/share/tests/physics_saturation_run_and_cmp.cpp
```diff
@@ -50,15 +50,17 @@ struct UnitWrap::UnitTest<D>::TestSaturation
     sat_ice_fp  = physics::polysvp1(temps, true, Smask(true))[0];
     sat_liq_fp  = physics::polysvp1(temps, false, Smask(true))[0];
 
-    mix_ice_fr = physics::qv_sat(temps, pres, true, Smask(true), physics::Polysvp1)[0];
-    mix_liq_fr = physics::qv_sat(temps, pres, false,Smask(true), physics::Polysvp1)[0];
+    //Functions<S,D>::qv_sat_dry(const Spack& t_atm, const Spack& p_atm_dry, const bool ice, const Smask& range_mask,
```
PeterCaldwell, line 53: Unless we completely delete qv_sat, we should probably continue testing it here... so *add* new tests instead of just replacing the qv_sat tests with qv_sat_dry tests.

PeterCaldwell:
I see mappy was disabled for the last several AT cycles, so no info there. Weaver and blake fail  physics_saturation_run_and_cmp, p3_run_and_cmp_cxx, and p3_run_and_cmp_f90. 

My understanding is that p3 switched from using qv_sat (wet) to qv_sat_dry in this PR, so it makes sense that p3_run_and_cmps would be non-BFB. physics_saturation_run_and_cmp were changed to use dry rather than moist qv_sat, so it makes sense that it also failed... but I would have liked to see qv_sat_dry added as an additional test rather than replacing qv_sat (since qv_sat is still in the code, we should keep testing it!). 

Also, just as a sanity check - shoc_run_and_cmp is passing because shoc is still using wet rather than dry qv_sat, right?

In short, I think test behavior is as expected but I think we should add physics_saturation_run_and_cmp tests for qv_sat in addition to qv_sat_dry before we integrate.

PeterCaldwell:
> qv_sat_dry is tested in phys_sat_run_cmp test (baseline test), while qv_sat_wet is tested in rel_humidity (not baseline test).
Grepping in relative_humidity_tests.cpp, I think only qv_sat_dry is tested??? In any case, I think it makes most sense to test both qv_sat_wet and qv_sat_dry in phys_st_run_cmp since that's the smallest unit that encapsulates that code. I think this test is particularly important since I don't think qv_sat_wet is actually used anywhere (except maybe SHOC?) so if it has no tests at all it is bound to give unexpected answers eventually! Apologies for the extra work - I know it's annoying.

I also can't find qv_sat in SHOC, but I could have sworn it was used there so I'm continuing to investigate.

oksanaguba:
I see your newest comment with some formatting and not sure if i need to respond. but assuming i do, then relative_humidity.cpp has line 
```
auto qv_sat_l = physics::qv_sat_wet( ...
```
which means the test will call this function, too, in compute_diagnostic() (why not diagnostics with -s at the end?).

PeterCaldwell:
Regarding qv_sat_wet in relhum tests: ugh, I didn't see this until I refreshed my browser. Sorry! I agree that qv_sat_wet is tested in relhum. I still think we should test qv_sat_wet in the saturation tests, but I guess I don't care enough to argue about it.

PeterCaldwell:
I figured out the SHOC qv_sat mystery (with Peter B's help). Check out https://github.com/E3SM-Project/scream/blob/06a178c79a2f2842e2a5b364f34f97a13e4c4dda/components/eamxx/src/physics/shoc/impl/shoc_assumed_pdf_impl.hpp#L288

SHOC computes the saturation vapor pressure and converts it to a mixing ratio on its own. The only place saturation vapor pressure is used is for converting it into qv_sat. I consider this to be dangerous because we could be using different definitions for qv_sat in different parts of the code. The actual qv_sat conversion is 
https://github.com/E3SM-Project/scream/blob/06a178c79a2f2842e2a5b364f34f97a13e4c4dda/components/eamxx/src/physics/shoc/impl/shoc_assumed_pdf_impl.hpp#L292 
while in the actual qv_sat routine we use 
https://github.com/E3SM-Project/scream/blob/acf554f4dfafa41225aa2cc5c0a45fe29a63bc37/components/eamxx/src/physics/share/physics_saturation_impl.hpp#L156

I think we should replace the more complicated formula in SHOC with a call to qv_sat_wet(?) unless @bogensch has a good argument for why not. We can do this in a separate PR if Oksana doesn't want to do it here...

oksanaguba:
@PeterCaldwell thanks! i have a slight preference to do these 2 things in a new PR, but i will do whatever you suggest.

PeterCaldwell:
Yeah, new PR sounds good to me. Let's integrate this now. I'll let someone else push the button in case I'm not current on my integration training :-)

AaronDonahue:
Done

PR #2199, opened 2023-02-28
P3/Shoc run_and_cmp: Cleanup cmake and improve robustness

### LINKED ISSUE #894: Cannot use -jN to generate baselines on GPU on summit
The command `make -j6 baseline` fails on summit, for a clear reason: the two `shoc_run_and_cmp` and `p3_run_and_cmp` both try to run in parallel, and both try to use the first GPU. Cuda on summit (but not on weaver, apparently) is refusing to let kokkos init the device if it's busy, so whomever comes first (between `shoc_run_and_cmp` and `p3_run_and_cmp`) can successfully run, while the other gets a runtime exception.

I am not sure what the best solution here is. The easiest one is to run `make baseline` in test-all-scream, without `-jN` flag.

Pinging run_and_cmp experts; @jgfouca @tcclevenger .

jgfouca:
Have baseline generation command use the same format as the ECUT tests. This will mean batch machines will use srun which should prevent oversubcribing GPUs etc.

Fixes #894


components/eamxx/src/physics/p3/tests/CMakeLists.txt
```diff
@@ -61,36 +61,57 @@ if (NOT SCREAM_BASELINES_ONLY)
...
+
+  get_test_property(p3_baseline_f90_fake_omp${SCREAM_TEST_MAX_THREADS} FULL_TEST_COMMAND P3_F90_GEN)
+  get_test_property(p3_baseline_cxx_fake_omp${SCREAM_TEST_MAX_THREADS} FULL_TEST_COMMAND P3_CXX_GEN)
+
+  separate_arguments(P3_F90_GEN_ARGS UNIX_COMMAND "${P3_F90_GEN}")
+  separate_arguments(P3_CXX_GEN_ARGS UNIX_COMMAND "${P3_CXX_GEN}")
 
   add_custom_target(p3_baseline_f90
-    COMMAND ${CMAKE_COMMAND} -E env OMP_NUM_THREADS=${SCREAM_TEST_MAX_THREADS} $<TARGET_FILE:p3_run_and_cmp_f90> -f -g -b ${SCREAM_TEST_DATA_DIR}/p3_run_and_cmp.baseline)
+    COMMAND ${CMAKE_COMMAND} -E env OMP_NUM_THREADS=${SCREAM_TEST_MAX_THREADS} ${P3_F90_GEN_ARGS})
```
bartgol, line 108: I'm puzzled. Doesn't this run the same command that the original version would?

jgfouca: No, ECUT adds a bunch of stuff like test-launcher, mpi stuff, etc.

bartgol: Ah, so batch machines will run the generation through the queue. Ok, thanks.

bartgol (approved):

jgfouca:
The diffs are not from this PR.

PR #2203, opened 2023-03-01
Ensure output is always flushed when program terminates

### LINKED ISSUE #2185: On model error, flush output before exiting.
That way we get all output in files at a model crash.

bartgol:
In case of abnormal termination (e.g., exception thrown), the driver was not finalized, causing output streams to not be flushed.

The fix was trivial for standalone runs:. replace the default dtor of the driver with one that calls the finalize method. The dtor is guaranteed to be called for all automatic (i.e., stack) variables.

For CIME runs, not so much. The issue is that the AD is no longer a stack var, but it's held by the `ScreamContext` object. My first take was "well, let's register a cleanup function with `atexit`". That sounds reasonable, but the fortran execution does not call `exit` like a C/C++ program would, and the fcn registered with atexit never gets called, and Fortran does not have an equivalent of atexit.

My current solution is to handle the exception inside our F90<->C++ interface layer, by catching the exception, and doing the cleanup:
```
  try {
    scream_blah
  } catch (...) {
    finalize_singleton
    throw; // <- this will be were all our stacktraces will point to
  }
```
The downside is that the exception that kills the run is now originating from that `throw` statement. So the backtrace of our runs will not be very informative, as it will always point to that line.

I am open to different strategies. If anyone knows of a way to intercept the termination of the program in fortran (similar to the `atexit` way in C/C++), I'm all ears. Meanwhile, I think this is the best we can do to guarantee our output streams are flushed.

Well, that's not the whole truth: we do not have the ability to flush streams if the error does not originate in our C/C++ code. E.g., we do not intercept errors originating in other components, so if CICE crashes, we will not have a chance to flush our output. That said, I think this is still a good improvement, in that we can flush output if EAMxx generates an error (e.g., a field property check fails).

I verified (by injecting a `EKAT_ERROR_MSG("whoops!")` call in the code) that current master did not properly flush files, while this branch does, both for standalone and CIME runs.

@PeterCaldwell I would like you to weigh in on the limitations and drawback(s) of the approach I implemented. To summarize, the limitations/drawbacks are
- limitation: won't work if crash originates outside of EAMxx or in EAMxx's F90 code
- drawback: stacktrace of all EAMxx errors will point to the same line in `scream_cxx_f90_interface.cpp`.

Fixes #2185

### DIFF
--- a/components/eamxx/src/control/atmosphere_driver.cpp
+++ b/components/eamxx/src/control/atmosphere_driver.cpp
@@ -94,6 +94,11 @@ AtmosphereDriver(const ekat::Comm& atm_comm,
   set_params(params);
 }
 
+AtmosphereDriver::~AtmosphereDriver ()
+{
+  finalize();
+}
+
 void AtmosphereDriver::
 set_comm(const ekat::Comm& atm_comm)
 {
--- a/components/eamxx/src/control/atmosphere_driver.cpp
+++ b/components/eamxx/src/control/atmosphere_driver.cpp
@@ -1327,6 +1327,10 @@ void AtmosphereDriver::run (const int dt) {
 void AtmosphereDriver::finalize ( /* inputs? */ ) {
   start_timer("EAMxx::finalize");
 
+  if (m_ad_status==0) {
+    return;
+  }
+
   m_atm_logger->info("[EAMxx] Finalize ...");
 
   // Finalize and destroy output streams, make sure files are closed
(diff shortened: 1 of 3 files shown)


ambrad:
A possible idea: Use basic_stacktrace in the EKAT error-handling code. Before throwing, print the stacktrace. Then throw as usual. This depends on basic_stacktrace being implemented in a meaingful way on the various platforms.

bartgol:
> A possible idea: Use basic_stacktrace in the EKAT error-handling code. Before throwing, print the stacktrace. Then throw as usual. This depends on basic_stacktrace being implemented in a meaingful way on the various platforms.
> 
> Oh, yeah, and it depends on having a time machine, since this is a C++23 feature... Oh, well.

Yeah, I was getting excited when I digged up something, only to find it was a particular platform or compiler feature (forgot what it was). At the very least, the assert macros in EKAT will print `__FILE__` and `__LINE__`, so we will have a tad more info. Although, if the exception is thrown in a base class such as `AtmosphereProcess`, we still have little info.

Another option would be to use boost::stacktrace...

PeterCaldwell:
Yeah, I think this PR is great - just because it isn't everything all the time doesn't mean it isn't a huge improvement!!!

AaronDonahue (approved):

PR #2204, opened 2023-03-01
Add subfield attributes to output of field groups - add more control over netCDF metadata

### LINKED ISSUE #2194: Add ordered list of variable names included in 'tracers' in scream.r files
In scream.r files (i.e. our restarts), advected quantities are bundled in a variable called `tracers`. It is occasionally useful to analyze or use data from the restart file, so it would be nice if scream.r listed the order in which variables are contained in the `tracers` array. Ideally, this would occur as an additional attribute of the tracers array, but I have some vague recollection that adding an attribute for a single variable in a netcdf file is hard. In that case, making a global attribute with this info would be fine. Note that there's also a bit of trickery here in that different simulations may have different number or order of tracers in the future (e.g. with prognostic aerosol), so we need to get the list of tracers online rather than hardcoding it.

AaronDonahue:
Add the capability to add more attributes and metadata to EAMxx output files.

1. Add an interface to set a string attribute in the netCDF metadata.  This will be useful for adding metadata like the averaging type of the output, or the casename.  This PR begins this process.  We will also be able to remove the boiler plate metadata which is currently hard-coded in the F90 interface and have the output manager set that information.
2. Add an attribute to any variable that is actually a set of subfields.  Examples include `tracers` and `horiz_winds`.  All registered variables will automatically be checked for any children (aka subfields).  If any are found than a new attribute called `sub_fields` will be attached to that variable.  The attribute will list a human readable ordered list of what the subfields are.  For example:
```
	float tracers(time, ncol, dim3, lev) ;
		tracers:units = "1" ;
		tracers:long_name = "tracers" ;
		tracers:sub_fields = "[ qv, tke, qc ]" ;
	float T_mid(time, ncol, lev) ;
		T_mid:units = "K" ;
		T_mid:long_name = "T_mid" ;
```

Fixes #2194


PeterCaldwell:
Sweet, thanks Aaron!

components/eamxx/src/share/io/scorpio_output.cpp
```diff
@@ -742,6 +742,29 @@ register_variables(const std::string& filename,
...
+      int it = 0;
+      for (const auto& child_weak : children) {
+        auto child = child_weak.lock();
+        it ++;
+        children_list << child->get_identifier().name().c_str();
+        if (it < children.size()) {
+          children_list << ", ";
+        }
+      }
+      children_list << " ]";
```
bartgol, line 764: Instead of using the it variable (which makes the impl more puzzling to read), consider a simpler approach
```
std::string children_list = "[";
for (auto ch_w : children) {
  auto child = ch_w.lock();
  children_list += child->get_identifier().name() + ", ";
}
// Remove last ", "
children_list.pop_back();
children_list.pop_back();
set_variable_attribute_c2f(...);
```

AaronDonahue: Excellent, I had just the same thought last night before going to bed.

components/eamxx/src/share/io/scream_scorpio_interface.F90
```diff
@@ -1257,6 +1262,121 @@ subroutine set_int_attribute (file_name, attr_name, val)
...
+    enddef_needed = .false.
+    if (ierr .ne. PIO_NOERR .and. pio_atm_file%is_enddef) then
+      ! In theory, there are several reason why this could fail. However, pio.F90
+      ! does *not* expose all the nc error codes like pio.h does (e.g., no PIO_ENOTATT).
+      ! So we just *assume* that the attribute was not found, and try to define it
+      ! NOTE: We also check if we have ended the definition phase for this file,
+      ! if not we don't need to open it again, and we don't need to flag it to
+      ! be ended when we are done.
+
+      ! TODO - LUCA QUESTION - Is the above statement true?
```
AaronDonahue, line 1355: @bartgol - tagging you because I wrote this comment inline while adding code.

bartgol: The comment refers to the call to PIO_inq_att right above. When I coded this if statement, the above statement was true (I don't know if it still is, but probably).

bartgol: Uhm, but I guess you are right. We might need to simply check if the file is closed, regardless of whether the att is already defined or not.

AaronDonahue:
@bartgol this is passing the AT, it should be ready for review.

components/eamxx/src/share/io/scream_scorpio_interface.F90
```diff
@@ -76,6 +76,8 @@ module scream_scorpio_interface
             eam_update_time,             & ! Update the timestamp (i.e. time variable) for a given pio netCDF file
             get_int_attribute,           & ! Retrieves an integer global attribute from the nc file
             set_int_attribute,           & ! Writes an integer global attribute to the nc file
+            set_str_attribute,           & ! Writes an string global attribute to the nc file
+            set_variable_attribute,      & ! Writes an string attribute for a variable in the nc file
```
bartgol, line 80: This seems very similar to `set_variable_metadata`. Maybe we can use that one?

bartgol (changes requested):

components/eamxx/src/share/io/scream_scorpio_interface.F90
AaronDonahue: Good point.  I don't think `set_variable_metadata` will work if the "definition" phase has been closed.  I think we can probably use `set_variable_metadata` but enhanced to reopen the definition phase if it has been closed, with the `set_X_attribute` functions do.

Quick question, I didn't see `set_variable_metadata` before you mentioned it here because when I started looking for ways to update metadata I first saw `set/get_int_attribute`.  Should we adopt the `set_x_attribute` naming convention for all functions that set metadata?  Or change the name of `set_int_attribute` to `set_int_metadata`?

bartgol: I have redone these a bit more neatly in the branch where I resume filling the output files. Here, you can just get rid of one of the two, to avoid redundancies, and then I'll deal with conflicts when issuing the next PR.

AaronDonahue: sounds good.

bartgol (approved):

PR #2206, opened 2023-03-01
Must use FORCE to set cache vars that have already been set

jgfouca:


bartgol:
It's not that simple, imho. Both
```
set (FOO "one" CACHE STRING "") 
set (FOO "two" CACHE STRING "" FORCE)
message ("FOO: ${FOO}")
```
and
```
set (FOO "one" CACHE STRING "") 
set (FOO "two")
message ("FOO: ${FOO}")
```
run with `cmake -DFOO=three ...` prints `FOO: two`. Which means the user is no longer able to configure the variable, and the cmake is always picking what was hard coded.

Imho, the best way of fixing this is to not use `other.cmake`. Or, maybe we can turn it into a macro
```
macro (SetMpiVars)
  set(options)
  set(args_1v MPIEXEC MPI_NP_FLAGS MPI_EXTRA_ARGS MPI_THREAD_FLAGS)
  set(args_Mv)
  cmake_parse_arguments(PARSE "${options}" "${args_1v}" "${args_Mv}")
  
  if (PARSE_MPIEXEC)
    set (EKAT_MPIRUN_EXEC ${PARSE_MPIEXEC} CACHE STRING "")
  else ()
    set (EKAT_MPIRUN_EXEC "" CACHE STRING "")
  endif ()
  ...
```
and then in mach files we can call it with only the ones we need to set
```
SetMpiVars (MPIEXEC "mpirun" MPI_NP_FLAGS "-n")
```
and the rest will be set to empty strings...

jgfouca:
@bartgol , this crossed my mind too. Will a user ever elect to use a machine file but then want to make changes to MPI stuff? I think it's unlikely.

bartgol:
> @bartgol , this crossed my mind too. Will a user ever elect to use a machine file but then want to make changes to MPI stuff via the cmake command line? I think it's unlikely.

That's a fair point.

bartgol (approved):

PR #2207, opened 2023-03-01
Allow CIME to use ${mach}-${compiler}.cmake machine files

### LINKED ISSUE #1665: Change build_scream.cmake to include a mach file matching mach-compiler combo
By including `components/scream/cmake/machine-files/${MACH}.cmake`, we complicate our lives, because we need the mach file to switch between compiler choices. Instead, we should include `${MACH}_${COMPILER}.cmake`. For instance, on Summit, this would make switching between P9 and V100 much easier, with `summit_gnugpu.cmake` containing all v100/cuda options, and `summit_gnu.cmake` contianing all p9/openmp options.

jgfouca:
Add some documentation for the machine file system.

Fixes #1665

### DIFF
--- a/components/cmake/build_eamxx.cmake
+++ b/components/cmake/build_eamxx.cmake
@@ -31,8 +31,13 @@ function(build_eamxx)
     string (REPLACE "-D" "" CPPDEFS "${CPPDEFS}")
     add_compile_definitions ("${CPPDEFS}")
 
-    # Include machine file here
-    include(${CMAKE_SOURCE_DIR}/eamxx/cmake/machine-files/${MACH}.cmake)
+    # Include machine file here. Prefer a compiler-specific file.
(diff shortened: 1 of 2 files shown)


sarats:
> When eamxx is being build as a component within an E3SM case,
these machine files will still be used. It will prefer a
${mach}-${compiler}.cmake over a plain ${mach}.cmake file

I was wondering what if E3SM had a similar file. In general for E3SM, we have $compiler_$machine.cmake? Is that the way we avoid a conflict?

sarats:
Oh, are these files going to be under eamxx and not under cime_config/machines?

jgfouca:
@sarats , yes, this is for eamxx machine files. The E3SM machine files already support this.

bartgol (approved):

bartgol (approved):

PR #2211, opened 2023-03-03
Update some comments related to output code

tcclevenger:
1. `scorpio_output.hpp`: Seems that comment for `Max Snapshots Per File` was truncated. Added some additional text.
2. `scream_default_output.yaml`: comment for `Max Snapshots Per File` wasn't taking into account that output freq/units are not fixed.


tcclevenger:
@AaronDonahue I'm just now getting into the output options and saw these inconsistencies. Let me know if anything here is way off, or if you have better wording suggestions.

PR #2215, opened 2023-03-06
Add a new CIME based EAMxx test to check output coarsening

AaronDonahue:
This commit adds a new test `scream-output_coarsening` that will test that our horizontal and vertical output capability hasn't been broken for CIME cases.


AaronDonahue:
The last task for this PR is to upload the map file to the blues server so it is available for download.

components/eamxx/cime_config/testdefs/testmods_dirs/scream/output_coarsening/shell_commands
```diff
@@ -0,0 +1 @@
+./xmlchange SCREAM_ATMCHANGE_BUFFER='Scorpio::output_yaml_files=${SRCROOT}/components/eamxx/data/scream_coarsening_output.yaml'
```
jgfouca, line 1: You'll want xmlchange to append this value (`--append`) so as to not flush whatever might have already been in the buffer.

jgfouca (changes requested):

jgfouca (dismissed):

AaronDonahue:
Fails pending the upload of a new file to the data server.

AaronDonahue:
@jgfouca , I think I know why this is still failing.  The automatic download of the remap file it needs isn't being triggered because it isn't listed in the namelist_defaults file, instead it is set in an output control yaml file, which I don't think `buildnml` check for required files.  Is there a CIME command I can add to the `testdefs` that will trigger `wget` if a specific file is missing?

jgfouca:
@AaronDonahue , which yaml file is this? I can add it to the list of things buildnml scans for input files.

jgfouca:
I'm going to do an upstream merge to bring in recent buildnml changes.

jgfouca:
@AaronDonahue , @bartgol , see what you think of my latest commit.

bartgol:
This assumes that all streams requiring remap will use the same tgt grid, right? I thought @AaronDonahue said he may want 2+ separate coarsening remaps.

AaronDonahue:
> This assumes that all streams requiring remap will use the same tgt grid, right? I thought @AaronDonahue said he may want 2+ separate coarsening remaps.
@jgfouca , yes, I'm not sure we can anticipate what remap files a user will want ahead of time and it might be confusing if we make it a part of the scream_namelist file.  If that makes sense.  I wonder if there is a `shell_commands` that we can use that does the equivalent of CIME's file check.  In other words, a command or set of commands that will attempt to touch this specific file and if it is missing issue a `wget`.  Is that approach possible?

bartgol:
Taking care of #2414 would partially address the issues with this PR. @AaronDonahue another issue with this PR is that it adds both horiz and vert remap, while in the IO classes we explicitly prohibit to do horiz+vert (we only allow one of them).

Separately, @jgfouca what do you think about using ATM vars (not CIME XML vars) as selectors? Meaning, allow to atmchange (some) selectors. It seems pointless to set stuff like EAMXX_OUTPUT_GRID in the main CIME vars, when it is clearly just an atm setting... So, ideally, I'd like to do
```
./atmchange output_remap_grid=ne2pg2
./case.setup
```
and see the value of `horiz_remap_file` in the output yaml file changed. I think this would take two steps:
1. allow to atmchange selectors (perhaps we should revamp the "locked" attribute, to distinguish selectors that are linked to the case, and cannot be changed, like hgrid, from others)
2. in `do_cime_vars_on_yaml_output_files` (or in a separate analogue routine), also expand `${blah}` if `blah` is an EAMxx selector.
```

jgfouca:
@bartgol , I'd have to think about all the implications of that. It seems like it should be doable although it opens us up to circular dependencies.

AaronDonahue:
addressed by #2727

PR #2217, opened 2023-03-07
Add new field to track ice cld frac for output

### LINKED ISSUE #2132: Cloud fraction outputs use wrong ice cloud fraction
The 3d cldfrac_tot field and the 2d cldhgh, cldmed, and cldlow fields are all based on the "internal ice cloud fraction" which needs to be 1 whenever any ice is present in the cell at all in order for microphysics to act at all. As a result, our cloud fraction variables are 100% almost always whenever qi is present. People evaluating models want ice cloud fraction to mean that there's enough ice in the cell to be visible (or something like that). We have an output field "ice_cloud_mask" which uses the diagnostic definition. We need all the total (liquid+ice) cloud fractions output by the model to use this "evaluation" ice cloud fraction rather than the "internal" ice cloud fraction. 

In addition, we should change our default output request to output the 3d *liquid* cloud fraction and the total cloud fraction rather than the total cloud fraction and ice_cloud_mask since ice will always be 0 or 1 in a snapshot while liquid is generally less than 1. Thus we can reconstruct the ice cloud fraction from the total and the liquid... but we can't recover reasonable cloud fractions from what we have for the 4 season runs. @AaronDonahue - can you make this change?

My understanding is that ice_cloud_mask is a "diagnostic output" so it is calculated then immediately written out rather than being available as an ingredient for other output, particularly those calculated within physical processes and passed via the field manager to the output process. At least some of these other cloud outputs (e.g. cldhgh) need to be calculated within radiation because that's where we have the overlap calculator. I think the natural solution is to compute ice_cloud_mask within the ice cloud fraction process and to make it available to the AD. Maybe we could call it ice_cloud_fraction_4OUTPUTONLY or something to make sure people don't try to use it in calculations in the future?

AaronDonahue:
This commit introduces two new fields to the cloud fraction process which track the ice and total cloud fraction using a more restrictive threshold for ice cloud fraction.  These new fields are clearly labeled as "_for_output" to discern them from the actual cloud fractions to be used by microphysics.

Note, this is needed because the current cloud fraction calculation considers a cell layer to have 100% ice cloud cover given any amount of cloud ice present. When considering ice cloud for output variables this has the effect of drowning out the cloud fractions for any trace amount of cloud ice including falling snow. So, as issue #2132 illustrates we need to have a separate cloud fraction calculation with a stricter threshold that can be used for output variables.

Fixes #2132


AaronDonahue:
Will fail pending the upload of two new initial condition files for unit tests to the data server.

AaronDonahue:
@PeterCaldwell and @brhillman , the fails will be in all of the standalone unit tests, which will attempt to download a file not yet on the server.

That file renames the `cldfrac_tot` variable in the initial condition files to `cldfrac_tot_for_output`.  Before we commit to that naming and usage I would like you two to chime in on the PR itself.   If we decide to change the name of the variable or change how RRTMGP uses the variable then it might requires changes to that file.

components/eamxx/src/physics/rrtmgp/atmosphere_radiation.cpp
```diff
@@ -386,7 +387,7 @@ void RRTMGPRadiation::run_impl (const double dt) {
   auto d_qv = get_field_in("qv").get_view<const Real**>();
   auto d_qc = get_field_in("qc").get_view<const Real**>();
   auto d_qi = get_field_in("qi").get_view<const Real**>();
-  auto d_cldfrac_tot = get_field_in("cldfrac_tot").get_view<const Real**>();
+  auto d_cldfrac_tot = get_field_in("cldfrac_tot_for_output").get_view<const Real**>();
```
PeterCaldwell, line 390: Question for @brhillman here: does radiation use total cldfrac for anything other than writing output? Aaron is changing the definition of cloud fraction here...

components/eamxx/src/physics/cld_fraction/cld_fraction_main_impl.hpp
```diff
@@ -31,11 +33,15 @@ ::main(
     const auto oqi   = ekat::subview(qi,   i);
     const auto oliq_cld_frac = ekat::subview(liq_cld_frac, i);
     const auto oice_cld_frac = ekat::subview(ice_cld_frac, i);
-    const auto otot_cld_frac = ekat::subview(tot_cld_frac,  i);
+    const auto otot_cld_frac = ekat::subview(tot_cld_frac, i);
+    const auto oice_cld_frac_4out = ekat::subview(ice_cld_frac_4out, i);
+    const auto otot_cld_frac_4out = ekat::subview(tot_cld_frac_4out, i);
 
-    calc_icefrac(team,nk,oqi,oice_cld_frac);
+    calc_icefrac(team,nk,1e-14,oqi,oice_cld_frac);
```
PeterCaldwell, line 40: our threshold used to be 1e-12 before this PR, right? Did you mean to change the threshold here?

AaronDonahue: yes it was, this is a typo, good catch!

components/eamxx/src/physics/rrtmgp/atmosphere_radiation.cpp
AaronDonahue: right.  This was my main concern.  I do see that cloud fraction is passed to `rrtmgp::main` as well as used for `cldlow`, `cldmed` and `cldhigh`.  It would not be difficult to use both definitions of cloud frac total if that is warranted.

brhillman:
> @PeterCaldwell and @brhillman , the fails will be in all of the standalone unit tests, which will attempt to download a file not yet on the server.
> 
> That file renames the `cldfrac_tot` variable in the initial condition files to `cldfrac_tot_for_output`. Before we commit to that naming and usage I would like you two to chime in on the PR itself. If we decide to change the name of the variable or change how RRTMGP uses the variable then it might requires changes to that file.

Why would this need a new inputfile? If this is just a diagnostic output, then that shouldn’t be read from an initial condition or anything, should it?

components/eamxx/tests/CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 # Some tests for checking testing works
-set(EAMxx_tests_IC_FILE_72lev "screami_unit_tests_ne4np4L72_20220822.nc")
-set(EAMxx_tests_IC_FILE_128lev "screami_unit_tests_ne2np4L128_20220822.nc")
+set(EAMxx_tests_IC_FILE_72lev "screami_unit_tests_ne4np4L72_20230307.nc")
```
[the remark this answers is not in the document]
AaronDonahue: For the rrtmgp standalone unit tests, which have `cldfrac_tot` as REQUIRED but no `cld_fraction` process is running to set it.

components/eamxx/src/physics/rrtmgp/atmosphere_radiation.cpp
brhillman: Okay, I misunderstood a little. Cloud fraction is still all or nothing in the presence of ice (0 or 1), this just changes how much ice needs to be present to consider a layer ice filled vs absent of ice. I’m tempted to say we should stick with the same definition as micro and any other processes use for determining whether or not a layer contains ice. But what would be really nice is to have some estimate of how much of the layer actually contains ice, because this will affect the heating. Note that sticking with the mask p3 uses means that cldlow, cldmid, and cldhgh will be trivially 1 in the case of any ice, but using the new “for_output” fields will mean they are sometimes 1, but also sometimes not. I guess that’s the point, but it does change the behavior of radiative heating (i.e., sometimes ice will have no impact on rad now).

PeterCaldwell: Sorry for tying you in knots, Ben. I was trying to ask the simpler question of "does radiation use cloud fraction for any of its internal calculations". Since it does, we should definitely pass it the old version of cloud fraction in addition to the "for output only" cloud fraction to use for diagnostics. 

Put another way, the goal of this PR is to fix some bad model outputs so if it is answer changing we've screwed up.

brhillman: Looking at it a bit more, I think we should keep the original threshold for considering a layer to be cloudy in terms of what rad sees. Arbitrarily lowering this threshold would just mean rad would completely ignore more layers that actually contain some amount of ice. With that though, this will keep our cldlow, cldmid, cldhigh definitions the same. We constructed those to be exactly consistent with what radiation sees by vertically projecting the subcolumn cloud mask used to get the subcolumn optical depth profiles that are used for computing rad fluxes. When we did this originally, we hailed this as a great improvement in consistency over how this used to be handled in other models, that used separate params for cloud area. I understand that keeping the current threshold means that we will continue to have a lot of, in particular, cldhigh columns that are entirely cloudy, but since this is what radiation actually sees, I might argue this isn’t such a bad thing. It seems to me that arbitrarily changing this threshold would just “mask” the real issue here, which is that we do not have an estimate of partial cloud ice fraction. At high enough resolution, this probably ceases to be a problem, but at coarser resolutions it is probably a big problem.

brhillman: > Sorry for tying you in knots, Ben. I was trying to ask the simpler question of "does radiation use cloud fraction for any of its internal calculations". Since it does, we should definitely pass it the old version of cloud fraction in addition to the "for output only" cloud fraction to use for diagnostics.
> 
> Put another way, the goal of this PR is to fix some bad model outputs so if it is answer changing we've screwed up.

No worries, I think it’s good to think through these things until we completely understand the choices we have made and whether or not they (still) make sense.

whannah1:
For the record - based on several conversations with Ben over the years I'm starting to think that we actually should not even allow cloud fraction to be a history output. It is not a valid quantity for comparison with observations, and it's often not even valid for comparison between models. I cannot think of any use case where analyzing cloud fraction would be superior to other quantities. So if it were up to me, I think I would get rid of all cloud fraction output variables and force people to use more meaningful quantities.

brhillman:
Another thought I had on this, I assume these are the names of the outputs that would exist in the output files, so, for example, in a simulation we would have output variables called `cldfrac_tot` and `cldfrac_tot_for_output`. This seems a little misleading, wouldn’t `cldmask` be more appropriate for `cldfrac_tot`, or `cldfrac_for_p3`, and then something like `diagnostic_cldfrac` or `thresholded_cldfrac`? These are definitely not good names, but hopefully this illustrates the point that really we just have two different cloud fraction outputs with somewhat arbitrary thresholds. Would it be possible to add what those threshold are to the netcdf variable attributes so there is less confusion at least?

AaronDonahue:
> Would it be possible to add what those threshold are to the netcdf variable attributes so there is less confusion at least?

yes, that shouldn't be too hard.

brhillman:
> For the record - based on several conversations with Ben over the years I'm starting to think that we actually should not even allow cloud fraction to be a history output. It is not a valid quantity for comparison with observations, and it's often not even valid for comparison between models. I cannot think of any use case where analyzing cloud fraction would be superior to other quantities. So if it were up to me, I think I would get rid of all cloud fraction output variables and force people to use more meaningful quantities.

To be clear, this is why satellite simulators exist.

AaronDonahue:
Based on the conversation so far I think my next task is to revert all the changes to radiation back to the original code.

Re: having cloud fraction as output @whannah1 - The way we have it setup now any field that is shared between processes is part of the field manager and thus also available as output.  It would be better to use a cloud simulator, but in the current infrastructure we can't stop a user from requesting the cloud fraction.  Hopefully they use it wisely if they do.

Final question.  @PeterCaldwell , @brhillman and @whannah1 , preference on variable names?  That seems to be the last open question before this is ready to be approved and merged.

PeterCaldwell:
In terms of names, I'd like to have 1Eminus5 in the name to be explicit about the arbitrariness and clear in our choice. I don't like "diagnostic_" because we often talk about variables in the model as being diagnostic or prognostic. I agree that "for_output" is also a bit vague, but I'm not sure what it could mean other than it is being used to write output. We could say "_4outputwrite" to be more specific? I'd be ok calling ice cloud fraction a cloud mask instead of a cldfraction, but we'd have to do that for both the "internal" value and the "for output" version because it is the zero vs one nature that makes it a cloud mask, not whether it's the output or internal threshold. Also, having ice be a mask and liq be a frac would make the relationship between these 2 variables unclear. 

I'm not sure how to turn this into a constructive suggestion rather than a rant. I guess I'd like to keep using cldfrac when talking about ice phase. I don't want to use "diagnostic" in the name. Are people ok with ice_cldfrac_1Eminus5QiCutoff or something like that?

bartgol:
If something is _just_ for output, maybe we can add a Diagnostic class, just like we do for other derived quantities?

Regarding cldfrac vs cldfrac_mask. I understand that there are downstream scripts that may depend on names and/or make assumptions on what things are, but I find it underwhelming that a Next-Generation software uses the term "fraction" for a mask.

> .. and force people to use more meaningful quantities.

I totally agree. I also believe that we can spend an hour going through the code and replace all usage of `cldfrac` with `cld_mask` (or `cldmsk`, since e3sm likes no-vowels-names...). As a person that had to put his hands in the code without knowing much (anything?) about atm modeling, I find it frustrating to not be able to understand what the vars are b/c of cryptic names, but I find it even more frustrating if the names are misleading unless you're privy to some "peculiar" model naming convention...

PeterCaldwell:
>uses the term "fraction" for a mask

The problem is that the liquid cloud fraction really is a fraction - it would be wrong to call it a mask - and the ice cloud fraction is really a mask. Then P3 makes the situation even worse because it insists on being handed the combined liquid+ice cloud fraction because otherwise each process would need to act separately on the subcloud, the liquid-only, the ice-only, and the liquid+ice portions of the cell which would get really complicated. So now we need a combined variable for the liquid cloud fraction and ice cloud mask. Do we call that a fraction or a mask? I guess we could call the combined and the liquid variables fractions and the ice a mask, but that gets pretty complicated. Basically, I think it is going to be confusing and/or misleading any way we manage it.

bartgol:
> Then P3 makes the situation even worse because it insists on being handed the combined liquid+ice cloud fraction because otherwise each process would need to act separately on the subcloud, the liquid-only, the ice-only, and the liquid+ice portions of the cell which would get really complicated.

How do we combine liquite+ice, if one is a fraction and the other a mask? It seems to me that we're adding/combining apples and oranges.

AaronDonahue:
@PeterCaldwell and @brhillman , I have stripped all the changes to the `rrtmgp` interface from this PR and made the threshold for the different definitions of ice cloud fraction a runtime option.  The final consideration is the naming convention, and how to let the user know the threshold that was used for the ice cloud fraction.

My vote is for a standard name and adding the threshold as variable metadata.  But I am open to any consensus outside of that.  Note, I think regardless of what name we choose we should add the metadata about the threshold to the output to make things more clear.

PeterCaldwell:
It seems like this PR has stalled out because there's no good name for what we're trying to do, so we can't agree on a name. 

Thus I'm going to make the executive decision that we should just integrate it as-is (i.e. using cldfrac_ice_for_output). If we come up with a better idea later, we can submit another PR.

PeterCaldwell (dismissed):

AaronDonahue:
@PeterCaldwell I had to fix an issue with single precision builds so this will need to be re-reviewed.  I also took the opportunity of new commits to change the name to `for_analysis` since I think we determined that was a slight improvement in the naming convention during our Thurs. telecom.

PeterCaldwell (approved):

AaronDonahue:
The DIFF's on the `mappy` CIME tests are expected.  This PR changes the names of 2 of the output fields.  I'm going to manually merge since this PR pass all units tests and has been approved.

Issue #2195, opened 2023-02-27
Surface pressure (ps) output appears to be double the correct size.


AaronDonahue:
Branched off from Issue #2160 , Here is the original text submitted by @golaz .

I ran a short test (1 month) on chrysalis with the latest code (b6314dad6440c284e8121cc402f9e723c5ec7640, COMPSET="F2010-SCREAMv1", RESOLUTION="ne30pg2_ne30pg2", 128 levels). Thanks to @crterai for helping me configure this run.

I noticed some issues with output variables.

**ps**

The surface pressure magnitude is way off, topping around 2052 hPa. 

![ps](https://user-images.githubusercontent.com/15743046/217918712-141f3cf6-0ed3-4626-90f2-9736da9563d5.png)

But the lowest model p_mid looks fine (topping at 1024 hPa), so this must be a diagnostic issue.

![p_mid-127](https://user-images.githubusercontent.com/15743046/217918798-d4884b0c-f925-4736-8e65-7e68948a5783.png)

AaronDonahue:
Bringing in comments from #2160 relevant to this issue:

Aaron Donahue:
Re: ps, have we been able to recreate this effect? I'm looking at a number of my recent ne30 runs and ps always seems to be right in the 100hPa 1000hPa range. @crterai , @golaz ?

Chris Terai:
One of my recent test shows relatively okay looking ps values so @golaz is better positioned to chime in here.

AaronDonahue:
I reran this case w/ all four output types:  AVERAGE, INSTANT, MAX, MIN and here are the results for the largest value that `ps` takes:
```
In [8]: np.amax(favg["ps"][:])
Out[8]: 205140.17

In [9]: np.amax(fins["ps"][:])
Out[9]: 103572.484

In [10]: np.amax(fmin["ps"][:])
Out[10]: 101642.055

In [11]: np.amax(fmax["ps"][:])
Out[11]: 104350.78
```

Based on these results it looks like something make be wrong with how we are handling the AVERAGE flag.  My first concern was that maybe we hit a large `ps` peak that when averaged out looks like 2x the normal values.  But the MAX output proves that over the entire simulation we don't go over a reasonable value.

@bartgol , you might have some insight.  Do you see any chance there is a bug in AVERAGE flag?  I would think our unit tests would catch this, but it might be worth noting that `ps` is a homme variable that is remapped to physics maybe just for writing as output.

AaronDonahue:
[...]

PR #2219, opened 2023-03-08
And a helper function to scorpio_output to check field list for duplicates

AaronDonahue:
This commit adds a helper function `sort_and_check` which can be called on the list of output fields and throw an error if the list has a duplicate name.  The function also sorts the list of fields, which has no big impact other than to put the list into alphabetical order.

#2195


bartgol (dismissed):

AaronDonahue:
Looks like the changes already found a few other places w/ duplicates in our testing.  I've removed these and it should now hopefully pass the AT.  It will then need a re-approval.

bartgol (approved):

PR #2221, opened 2023-03-09
EAGLES: Third-party libraries and stub aerosol microphysics process

jeff-cohere:
Hey, everyone--EAGLES did something!

This PR brings in two new submodules (Haero, our aerosol package interface thingy, and mam4xx, our C++ port of MAM4 [so far, anyway]), and establishes a stub aerosol microphysics `AtmosphereProcess` that @pbosler and I can develop in feature branches. It may not look like much, but it reflects the work we've done in integrating these libraries into EAMxx's build system.

MAM4's aerosol microphysics parameterizations have been ported to C++, so this `AtmosphereProcess` will come to contain and manipulate those parameterizations in the most appropriate way for EAMxx's timestepping. In the short term, we will establish a test that demonstrates our ability to nucleate H2SO4 gas to sulfate aerosol. More to come.

Everything in this branch is protected by the `SCREAM_ENABLE_MAM` build option, which is `OFF` by default. No work has yet been done to show that this branch builds on anything but my workstation, so if you want to flip this switch, be prepared for sparks to fly!


components/eamxx/src/physics/mam4/aerosol_microphysics.cpp
```diff
@@ -0,0 +1,171 @@
...
+  // atmospheric quantities
+  add_field<Required>("T_mid", scalar3d_layout_mid, K, grid_name);
+  add_field<Required>("p_mid", scalar3d_layout_mid, Pa, grid_name); // total pressure
+  add_field<Required>("qv", scalar3d_layout_mid, q_unit, grid_name, "tracers");
+  add_field<Required>("pbl_height", scalar2d_layout_col, m, grid_name);
+  add_field<Required>("pseudo_density", scalar3d_layout_mid, q_unit, grid_name); // pdel
+  add_field<Required>("cldfrac_tot", scalar3d_layout_mid, nondim, grid_name);
+
+  // tracer group (stores all aerosol prognostics)
+  add_group<Updated>("tracers", grid_name, Bundling::Required);
```
bartgol, line 64: I'm assuming mam does not update all tracers, but just the aerosols ones? At some point, we may want to restrict it to those only, and see how robust it the FieldManager ability to rearrange fields inside groups (which aims at keeping as many as possible subgroups contiguous).

components/eamxx/CMakeLists.txt
```diff
@@ -51,7 +51,7 @@ endif()
 # to be on. For now, simply ensure Kokkos Serial is enabled
 option (Kokkos_ENABLE_SERIAL "" ON)
 
-set (CMAKE_CXX_STANDARD 14)
+set (CMAKE_CXX_STANDARD 17)
```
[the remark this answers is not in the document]
jeff-cohere: Nice catch, @bartgol ! (I forgot to add you to the reviewer list.) We actually do use some very very basic C++17 features, and I believe they're related to Kokkos, but I'd have to review. If this causes problems, we can probably revert to C++14 and work around them in our feature branches, but it would be nice if we could switch without incident.

components/eamxx/src/physics/mam4/aerosol_microphysics.cpp
jeff-cohere: Your assumption is mostly correct--we'll also need to update some gas tracers that interact with aerosols. I agree it would be nice if we can devise a way to restrict updates to tracer subgroups.

components/eamxx/CMakeLists.txt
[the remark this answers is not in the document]
bartgol: I don't know if it will cause issues. My main concern is to be able to compile correctly on all target machines. I think they should all be C++17 compliant by now, or else Kokkos would not be forcing C++17, I think. The caveat is to make sure the chain of TPL libs we need on those machines is fully supported with the C++17 compliant compiler suite.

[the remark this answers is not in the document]
jeff-cohere: I'll go along with whatever the EAMxx team thinks is best. Let me know if you'd like this reverted back to C++14, or whether I can help determine whether C++17 is practical.

components/eamxx/src/physics/mam4/aerosol_microphysics.hpp
```diff
@@ -0,0 +1,247 @@
...
+      qv_ = qv;
+      q_soag_ = q_soag;
+      q_h2so4_ = q_h2so4;
+      q_nh3_ = q_nh3;
+      q_aitken_so4_ = q_aitken_so4;
+    } // set_variables
+  }; // MAM4AerosolMicrophysics::Postprocess
+
+  // MAM4 aerosol particle size description
+  mam4::AeroConfig aero_config_;
```
PeterCaldwell, line 226: Is there a standard meaning to a variable ending in an underscore? Is this standard across all of EAMxx, or just within MAMxx? This is probably a matter of me not spending enough time in code rather than a real problem.

jeff-cohere: Ah, it's just something I use to indicate that a variable or method is `private` or `protected`. I don't think we have C++ style rules for EAMxx yet, or if we do, I'm not aware of them (or have forgotten about them!). It's easy enough for us to change these to fit a spec. But I like some indication that something belongs to an object, because it makes analyzing code in methods easier.

PeterCaldwell:
This looks very nice! Good job! A couple of questions/comments:
1. My understanding is that MAM needs to be broken up into a couple of atm_process instances rather than executing as a single process (i.e. in a single location in the timestep call loop). For example, aerosol microphysics, transport, activation, etc probably need to occur before/after different procs. You've done a nice job of creating the aerosol microphysics process, but I think the code structure should be `scream/components/eamxx/src/physics/aero_microphys/` instead of scream/components/eamxx/src/physics/mam4/` because the last subdirectory is meant to be the atm_process name, not a drawer full of a bunch of different procs. If you really want to have all mam4 procs in one place, maybe we should put aerosol on the same level as physics and dynamics.
2. I think calling your directory mam4 is an unfortunate choice because eventually you might want to have more than 4 modes. Maybe call it mam or just aerosol?
3. EAMxx's (or maybe just my) vision is that each process has a). an*interface* layer which grabs all the variables needed for the proc from the field manager and b). the actual parameterization code. The actual parameterizations should always pass their argument lists explicitly instead of using trickery to make the variables magically appear in scope for the inner processes. One goal here is to make unit tests easier. The other is to make it easier to track how/where variables are changed throughout the code. My bigger comment is that I don't see lists of vars passed to procs in your interface layer. Maybe just because those procs aren't wired up yet? My secondary comment - which is a complaint about EAMxx overall which you've exacerbated - is that your interface functions aren't named anything that makes their role as interface clear. We use `atmosphere_<name>` to indicate an interface function in EAMxx and you're using `aerosol_<name>` for your code. I think the solution here is to switch everything to `interface_<name>` where MAM's name includes both `aerosol` and `microphysics` (or whatever proc is needed).

jeff-cohere:
Thanks, @PeterCaldwell ! To your comments:
>     1. My understanding is that MAM needs to be broken up into a couple of atm_process instances rather than executing as a single process (i.e. in a single location in the timestep call loop). For example, aerosol microphysics, transport, activation, etc probably need to occur before/after different procs. You've done a nice job of creating the aerosol microphysics process, but I think the code structure should be `scream/components/eamxx/src/physics/aero_microphys/` instead of scream/components/eamxx/src/physics/mam4/` because the last subdirectory is meant to be the atm_process name, not a drawer full of a bunch of different procs. If you really want to have all mam4 procs in one place, maybe we should put aerosol on the same level as physics and dynamics.
>     2. I think calling your directory mam4 is an unfortunate choice because eventually you might want to have more than 4 modes. Maybe call it mam or just aerosol?

I've followed the convention of `src/physics/name-of-thingy` (e.g. `p3`, `shoc`, `rrtmgp`, none of which give any clue as to what they do). It is true that our port of MAM will end up comprising a small number (~3) of atmosphere processes, but there's no prior art, so I just assumed that things would live in the `src/physics/name-of-aerosol-thingy` directory. But I'm happy to rearrange as needed. If we want to move toward a `src/physics/physics-related-name` scheme, I can call this `aero_microphys`.

>     3. EAMxx's (or maybe just my) vision is that each process has a). an_interface_ layer which grabs all the variables needed for the proc from the field manager and b). the actual parameterization code. The actual parameterizations should always pass their argument lists explicitly instead of using trickery to make the variables magically appear in scope for the inner processes. One goal here is to make unit tests easier. The other is to make it easier to track how/where variables are changed throughout the code. My bigger comment is that I don't see lists of vars passed to procs in your interface layer. Maybe just because those procs aren't wired up yet? My secondary comment - which is a complaint about EAMxx overall which you've exacerbated - is that your interface functions aren't named anything that makes their role as interface clear. We use `atmosphere_<name>` to indicate an interface function in EAMxx and you're using `aerosol_<name>` for your code. I think the solution here is to switch everything to `interface_<name>` where MAM's name includes both `aerosol` and `microphysics` (or whatever proc is needed).

The "interface layer" (a) is going to be the `AerosolMicrophysics` class, and the parameterizations (b) are implemented in `mam4xx`, so we're all good, vision-wise. In fact, `mam4xx` has its own set of tests, so the parameterizations will be testable outside of EAMxx, even!

Meanwhile, if `AerosolMicrophysics` looks empty, that's because it is! This is just a foothold for us to begin collaborative development on (a) in earnest without having to do a bunch of rebasing to keep up with the `master` branch.

I can rename the class name to `AtmosphereAerosolMicrophysics` and the files to `atmosphere_aero_microphysics.{hc}pp` if that works.

Have I addressed your comments?

jeff-cohere:
Blake said:

```
/home/e3sm-jenkins/blake/workspace/SCREAM_PullRequest_Autotester_Blake/4320/scream/externals/ekat/extern/spdlog/src/fmt.cpp(29): error: incomplete type is not allowed
FMT_API static constexpr const char hex_digits[] = "0123456789abcdef";
^
detected during instantiation of class "fmt::v8::detail::basic_data [with T=void]" at line 106
```

Maybe we have our first indication of a C++17-related issue here?

components/eamxx/src/physics/mam4/aerosol_microphysics.hpp
bartgol: I usually use `m_` for member variables, and `s_` for static member variables. I've seen the trailing underscore used in the opposite way in other libs, to distinguish the input var `foo_` from the stored var `foo`, usually in constructor/setter methods, so that the shorter version is available in other methods. I personally dislike no prefix/suffix for member vars, since you never know inside an impl whether the var is local or not (but i'm ok with it in case of tiny small POD structs).

I think we briefly discussed coding conventions at the beginning, but the consensus was, IIRC, that we didn't want ppl to get lost amidst conventions.

jeff-cohere: Happy to go with the flow here, too. It's pretty easy to make these kinds of changes whenever we need to, since we're already following a convention.

bartgol: No need to, since you did implement some "class-member-marker" (the underscore). We can do a formatting PR if/when we ever decide is important/needed.

components/eamxx/tpls/CMakeLists.txt
```diff
@@ -16,3 +16,61 @@ if (SCREAM_CIME_BUILD)
   include (${SCREAM_BASE_DIR}/cmake/tpls/Mct.cmake)
   CreateMctTarget()
 endif()
+
+# MAM aerosol support
+if (SCREAM_ENABLE_MAM)
```
bartgol, line 21: I'm curious, why did you use ExternalProject_Add rather than just to add_subdirectory for haero/mam4xx, like we do for other submodules?

jeff-cohere: I prefer `ExternalProject_Add` to in-tree builds because it's more of a self-contained system that plays more nicely with other things. You can build a software package and have it install its things in a given location without having to understand what variables it sets or any of that business, and you can just look at the installed config CMake file (or pkgconfig or whatever) to fish out the information.

By contrast, I find in-tree builds harder to understand, because they are sucked inside their parent projects, and it's not as clear what is dictating the behavior of the compilation. Also, if you have packages that assume that another software package is already installed in some prefix, it's harder to convince them that an in-tree software package is actually installed as it would be "in the field." I've had this experience with Sandia's TChem library, which wants a pre-installed version of Kokkos somewhere, and the way we build EKAT (and Kokkos) in-tree creates a lot of complexity.

Happy to discuss, adjust, etc. But I hope you agree with me that installing Haero and mam4xx this way is pretty simple!

bartgol: Good points. Relying on an installation of some of our TPLs would make CIME-vs-standalone similar. I might revisit my stance on subdirs...

jeff-cohere:
Okay, since #2224 indicates that switching to C++17 might be nontrivial, I've added logic that enables C++17 if `SCREAM_ENABLE_MAM` is `ON`, and otherwise uses C++14. Is that okay with everyone here?

PeterCaldwell:
> Okay, since https://github.com/E3SM-Project/scream/pull/2224 indicates that switching to C++17 might be nontrivial, I've added logic that enables C++17 if SCREAM_ENABLE_MAM is ON, and otherwise uses C++14. Is that okay with everyone here?

Ok with me. It seems like we will need to get C++ 17 working in EAMxx eventually in order to be able to upgrade Kokkos. Should we make a git issue for this (or do we already have one)?

jeff-cohere:
Well, we could just keep banging away at PR #2224. :-) But I'll create an issue if we think that's needed to track it.

jeff-cohere:
All right! 🎉 

@bartgol, @PeterCaldwell, @AaronDonahue , are you okay with merging this PR?

PeterCaldwell (approved):

jeff-cohere:
I'll set this to AUTOMERGE, since the config option protects you all from these changes... for now!

PR #2224, opened 2023-03-09
Switch to C++17 from C++14(!)

### LINKED ISSUE #2233: Make the switch to C++17
Because MAM and because Kokkos 4, we know that C++17 is in our near future. We depend on lots of software and lots of machines, so it could be a little tricky switching from C++14 to C++17 (see the related PR #2224), so this issue is here so we have a nice box to tick when we succeed.

jeff-cohere:
This is a single-character-change PR, tied for the smallest in this repository's history.

This single character switches to C++17, which brings (checks notes) improved support for lambdas, which I believe our EAGLES-related work happens to need. But also, Kokkos is switching over to C++17, so this change is in our future even if it's a bit too early yet.

Let's see what the testing system has to say.

Closes #2233

### DIFF
--- a/components/eamxx/CMakeLists.txt
+++ b/components/eamxx/CMakeLists.txt
@@ -51,12 +51,7 @@ endif()
 # to be on. For now, simply ensure Kokkos Serial is enabled
 option (Kokkos_ENABLE_SERIAL "" ON)
 
-# MAM support requires C++17 -- hopefully SCREAM itself will get there soon
-if (SCREAM_ENABLE_MAM)
-  set(CMAKE_CXX_STANDARD 17)
-else()
-  set(CMAKE_CXX_STANDARD 14)
-endif()
+set(CMAKE_CXX_STANDARD 17)
 
 if (NOT SCREAM_CIME_BUILD)
   project(SCREAM CXX C Fortran)


bartgol (approved):
C++17 also comes with one of my favorite features: [structured binding](C++17 brings also one of my favorite features: https://en.cppreference.com/w/cpp/language/structured_binding).

Aside from being useful in itself, structured binding would work beautifully in combination with zip iterator pattern, to reduce tedious boilerplate patterns that make the code less readable.

jeff-cohere:
Yeah, structured binding is very cool, too. I feel like C++17 is actually an improvement to the language and not just a bunch of new features.

AaronDonahue:
Looks like this PR hit a snag on `blake` with a build error:
```
/home/e3sm-jenkins/blake/workspace/SCREAM_PullRequest_Autotester_Blake/4330/scream/externals/ekat/extern/spdlog/src/fmt.cpp(29): error: incomplete type is not allowed
FMT_API static constexpr const char hex_digits[] = "0123456789abcdef";
^
detected during instantiation of class "fmt::v8::detail::basic_data [with T=void]" at line 106
```
@jeff-cohere ^^

jeff-cohere:
Yep, this is the same failure encountered by the EAGLES PR. It surprises me that fmt isn't able to build with C++17. Maybe we can configure it differently to get past this.

@bartgol

bartgol:
That's bizarre. The macro evaluates to nothing (I checked the pre-processed cpp.i file), so the offending file is simply
```
...
        {'9', '4'}, {'9', '5'}, {'9', '6'}, {'9', '7'}, {'9', '8'}, {'9', '9'}};
     static constexpr const char hex_digits[] = "0123456789abcdef"; // <! this line
...
```
which is fine.

jeff-cohere:
The only sense I can make of it is that `FMT_API` isn't actually expanded (because it might not be defined) in that translation unit. But I can't understand why that would be the case building with C++17 and not with C++14.

bartgol:
> The only sense I can make of it is that `FMT_API` isn't actually expanded (because it might not be defined) in that translation unit. But I can't understand why that would be the case building with C++17 and not with C++14.

It expands to nothing, I verified that building and looking at `fmt.cpp.i`. Another "fix" is to replace `[]` with `*`. The two are equivalent in every aspect, except that `const char []` can be used to init another constexpr, while `const char *` can't.

bartgol:
Ok, another proposal: switching to `v1.11.0` of spdlog in EKAT fixes it (I only built ekat, but I assume that's enough).

@jeff-cohere , what do you think?

jeff-cohere:
> Ok, another proposal: switching to `v1.11.0` of spdlog in EKAT fixes it (I only built ekat, but I assume that's enough).
> 
> @jeff-cohere , what do you think?

I think it's definitely worth a shot!

bartgol:
> > Ok, another proposal: switching to `v1.11.0` of spdlog in EKAT fixes it (I only built ekat, but I assume that's enough).
> > @jeff-cohere , what do you think?
> 
> I think it's definitely worth a shot!

Great. I put up a PR in EKAT. If/when it passes, you should update the submodule in this PR (as well as fix the merge conflicts, so it can test again), and we can check if everything works.

bartgol (approved):

Issue #1851, opened 2022-08-09
EAMxx_tests_IC_FILE_72lev needs to be updated


AaronDonahue:
During some separate work I noticed that the default test initial condition file is for an ne4 initial condition, but the tests are all set to be ne2.  So the initial condition file for our tests in the `components/scream/tests` directory needs to be updated.

This means a remap of the ne4 IC file for testing to an ne2 file version.
then make the change in `components/scream/tests/CMakeLists.txt`

Making this issue so I don't lose sight of this bugfix.

bartgol:
Our IC files for unit tests are now ne2.

PR #2226, opened 2023-03-09
Verify input file has same dimensions as populated fields

### LINKED ISSUE #1329: IO input silently continues when the nc file has the wrong dimensions
You can read an IC file with N columns while running a grid with M columns, and IO never complains. We need to figure out why the code does not crash, and then make it crash if the dimensions of a file in the nc file do not match the dimensions of the scream field.

AaronDonahue:
This commit introduces a check to scorpio_input.h/cpp that checks the dimension lengths in the input file against the dimension lengths of individual fields to be populated.  If they don't match then it throws an error.

This should remove the hacky/buggy feature that allowed some tests to use an initial condition file with number of columns = X but declare the problem size to have number of columns = Y, where X != Y.

It also stops a user from accidentally using an initial condition file that does not match the problem they are running.

Lastly, this commit updates the input.yaml control files for all of the standalone tests that were doing exactly that.

Fixes #1329

### DIFF
--- a/components/eamxx/src/share/grid/remap/horizontal_remap_utility.hpp
+++ b/components/eamxx/src/share/grid/remap/horizontal_remap_utility.hpp
@@ -38,7 +38,7 @@ struct HorizontalMapSegment {
   HorizontalMapSegment(const gid_type dof_gid, const int length, const view_1d<const gid_type>& source_dofs, const view_1d<const Real>& weights);
 
   // Helper Functions
-  bool check() const;      // Check if this segment is valid
+  bool check() const;  // Check if this segment is valid
   void print() const;  // Useful for debugging, print the segment mapping info
 
   // Setter Functions
--- a/components/eamxx/src/share/io/scorpio_input.cpp
+++ b/components/eamxx/src/share/io/scorpio_input.cpp
@@ -205,8 +205,8 @@ set_grid (const std::shared_ptr<const AbstractGrid>& grid)
   // Sanity checks
   EKAT_REQUIRE_MSG (not m_io_grid, "Error! Grid pointer was already set.\n");
   EKAT_REQUIRE_MSG (grid, "Error! Input grid pointer is invalid.\n");
-  const bool skip_grid_chk = m_params.get<bool>("Skip_Grid_Checks",false);
-  if (!skip_grid_chk) {
+  m_skip_grid_chk = m_params.get<bool>("Skip_Grid_Checks",false);
+  if (!m_skip_grid_chk) {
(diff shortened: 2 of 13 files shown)


AaronDonahue:
@bartgol , this PR bought up a point about our stand alone testing.  We have been psuedo running ne2 by specifying the grid size to be 218 but using the ne4 init condition file.  Currently I've bumped the tests up to 866 columns matching the init but we should probably keep the 218 and just generate a new init condition file with an ne2 configuration.

I'm pinging you to get some feedback on that thought.

bartgol:
> @bartgol , this PR bought up a point about our stand alone testing. We have been psuedo running ne2 by specifying the grid size to be 218 but using the ne4 init condition file. Currently I've bumped the tests up to 866 columns matching the init but we should probably keep the 218 and just generate a new init condition file with an ne2 configuration.
> 
> I'm pinging you to get some feedback on that thought.

To be honest, I've been thinking for a while that we should instead init our fields on the fly for our standalone tests. I don't think there's _a ton_ of conditions that the inputs need to verify, in order to avoid nonsense calculations. Some of these can be handled by using a proper pdf (e.g., non-negative values, or use perturbations from a constant value), some others are slightly more complex, but not too bad (e.g., pressure increasing is simply the column scan of a non-negative quantity).

That way we would stress test a bit more our physics (e.g., what if one of the if statements in p3 is never triggered in standalone system tests?), and we could also change the number of cols (for physics-only tests) to whatever we want (e.g., run physics only tests with a specific set of values for ncols, to study perf dependence on workload).

AaronDonahue:
>That way we would stress test a bit more our physics (e.g., what if one of the if statements in p3 is never triggered in standalone system tests?), and we could also change the number of cols (for physics-only tests) to whatever we want (e.g., run physics only tests with a specific set of values for ncols, to study perf dependence on workload).

Yes, that would probably be much better for all of the physics based tests, especially since we don't have to worry about consistency across adjacent columns.  

It's a bit different for any of the tests w/ dynamics.  Maybe the right thing to do would be to focus on those tests for now since they will need a valid initial condition file (right?) and then create a git-issue to improve the physics-only stand-alone tests to have in-house values.  Thoughts?

bartgol:
Yeah, we don't need to do that now. Feel free to choose the approach you think it's best for this PR.

AaronDonahue:
Fixes #1851

bartgol:
@AaronDonahue This PR fell behind quite a bit. #2361 will take care of checking dimensions though, so maybe we can close this and push for that one to complete?

PR #2227, opened 2023-03-09
atmchange: Add support for empty lists, reset option

jgfouca:
You can set a list/array variable to empty via:
`./atmchange varname=''`

### DIFF
--- a/components/eamxx/cime_config/buildnml
+++ b/components/eamxx/cime_config/buildnml
@@ -20,13 +20,13 @@ Some import errors are expected and can be ignored
 
 It is encouraged that any changes to this system be tested with
 eamxx/cime-nml-tests
-
 """
 
-import os
+import os, sys
 
 from CIME.case import Case
 from CIME.utils import expect, safe_copy, SharedArea, run_cmd_no_fail
+from CIME.buildlib import parse_input
 
 from eamxx_buildnml import create_raw_xml_file, create_input_files, create_input_data_list_file, \
     do_cime_vars_on_yaml_output_files
--- a/components/eamxx/scripts/atmchange
+++ b/components/eamxx/scripts/atmchange
@@ -17,10 +17,17 @@ from atm_manip import get_xml_node, atm_config_chg_impl
 from utils import run_cmd_no_fail, expect
 
 ###############################################################################
-def atm_config_chg(changes, no_buffer=False):
+def atm_config_chg(changes, no_buffer=False, reset=False):
 ###############################################################################
     expect(os.path.exists("namelist_scream.xml"),
-           "No pwd/namelist_scream.xml file is present. Please rum from a case dir that has been setup")
+           "No pwd/namelist_scream.xml file is present. Please run from a case dir that has been set up")
+
+    if reset:
+        run_cmd_no_fail(f"./xmlchange SCREAM_ATMCHANGE_BUFFER=''")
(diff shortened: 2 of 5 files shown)


components/eamxx/scripts/atmchange
```diff
@@ -17,10 +17,17 @@ from atm_manip import get_xml_node, atm_config_chg_impl
...
-def atm_config_chg(changes, no_buffer=False):
+def atm_config_chg(changes, no_buffer=False, reset=False):
 ###############################################################################
     expect(os.path.exists("namelist_scream.xml"),
-           "No pwd/namelist_scream.xml file is present. Please rum from a case dir that has been setup")
+           "No pwd/namelist_scream.xml file is present. Please run from a case dir that has been set up")
+
+    if reset:
+        run_cmd_no_fail(f"./xmlchange SCREAM_ATMCHANGE_BUFFER=''")
+        print("All buffered atmchanges have been removed. A fresh namelist_scream.xml will be generated the next time buildnml (case.setup) is run.")
```
bartgol, line 27: Is this true? Don't you need to rm namelist_scream.xml in order to get a new one generated by buildnml?

components/eamxx/scripts/atmchange
```diff
@@ -66,7 +73,14 @@ OR
         help="Used by buildnml to replay buffered commands",
     )
 
-    parser.add_argument("changes", nargs="+", help="Values to change")
+    parser.add_argument(
+        "--reset",
+        default=False,
+        action="store_true",
+        help="Forget all previous atmchanges",
```
bartgol, line 80: But does not reset the XML to whatever case.setup would generate, right? I think of `--reset` as restoring the default/original values, while that's not happening here.

components/eamxx/cime_config/buildnml
```diff
@@ -20,13 +20,13 @@ Some import errors are expected and can be ignored
 
 It is encouraged that any changes to this system be tested with
 eamxx/cime-nml-tests
-
 """
 
-import os
+import os, sys
```
bartgol, line 25: Not really important, but is sys needed at all?

bartgol (commented):
The empty list looks good. I'm a bit confused on the reset option.

components/eamxx/scripts/atmchange
AaronDonahue: This came up during the v1 telecon.  With the recent change to our buildnml which buffers all `atmchange` calls into an XML entry it isn't easy to reset that buffer to be empty if you want to return to defaults.  So this option is supposed to take care of that.

bartgol: Ah, nvm, I thought we did not re-generate the XML file, but Jim's last PR on buildnml made it so that the atm xml is always rebuilt. Unless SCREAM_HACK_XML=ON.

@jgfouca should we print a warning if SCREAM_HACK_XML=ON  and --reset is used? As things stand, resetting the buffer won't do anything if SCREAM_HACK_XML=ON, since the next buildnml will not regenerate the XML.

components/eamxx/cime_config/buildnml
bartgol: Interesting. I only looked at changes, and didn't see anything with sys. How did this manage to work before this fix then?

components/eamxx/scripts/atmchange
bartgol: Nah, no need to re-rerun case.setup. I forgot about your recent change that XML is _always_ regenerated.

bartgol (approved):

PR #2231, opened 2023-03-10
Fix FieldAtLevel diagnostic

### LINKED ISSUE #2230: qv@bot not working right
@crterai and Yunyan report that qv@bot looks really weird in our recent Oct 1 ne1024 simulation - it has values of 1e7. Looking at the bottom layer of the full 3d qv variable looks normal though. Chris reports that T_mid@bot looks fine for this run. There's also a possibility that there was a bug in our code in Dec when we did this run that we fixed in recent months (but I don't remember one).

Bad values from qv@bot:
image001

Reasonable values from qv[127]:
image002

bartgol:
The previous impl was implicitly assuming the input field alloc was contiguous (which is not true for subfields).

I changed also the unit test, so that the midpoint field is a subfield of another field, and verified that in such condition the original impl was giving the wrong answer.

Fixes #2230

### DIFF
--- a/components/eamxx/src/diagnostics/field_at_level.cpp
+++ b/components/eamxx/src/diagnostics/field_at_level.cpp
@@ -13,6 +13,10 @@ FieldAtLevel::FieldAtLevel (const ekat::Comm& comm, const ekat::ParameterList& p
 {
   m_field_name  = m_params.get<std::string>("Field Name");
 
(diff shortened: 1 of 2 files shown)


AaronDonahue (approved):
Looks good, record for fastest resolved issue!

PeterCaldwell (approved):
Wow, you're on fire Luca! Thanks! I don't fully understand the details of the code here, but the conceptual problem and the solution make sense.

Issue #2222, opened 2023-03-09
Issues with namelist_default_scream.xml and rough topography


tcclevenger:
A user can specify `F2010-SCREAMv1-X6T` compset, but not receive an x6t topo file.

__Example:__
 No rough topography file exists for ne30np4 case in https://web.lcrc.anl.gov/public/e3sm/inputdata/atm/cam/topo/. If one runs a 
`F2010-SCREAMv1-X6T.ne30_ne30` V1 case, the `-X6T` is ignored and the topography file is just considered to be
`USGS-gtopo30_ne30np4_16xdel2-PFC-consistentSGH.nc` (smooth topo).

__Proposed solution:__
- Since `ne${x}np4gp2` topo files have both `PHIS` (pg2) and `PHIS_d` (GLL), switch to only using these files and having scream logic that chooses the correct field name based on the physics grid type. 
- Error out if a user asks for `F2010-SCREAMv1-X6T` compset for a grid that doesn't have a rough topo file (e.g., ne4, ne120, ne512)

__Potential Issues__
- If the user wants any other field from the topo file, during an np4 run, only pg2 data exists.

PR #2237, opened 2023-03-13
Only use np4pg2 topography files

tcclevenger:
Previously, topography files of type `*np4_*` and `*np4pg2_*` were being used in `namelist_defaults_scream.yaml`, although the `*np4pg2_*` files contain both `PHIS` (PG2 nodes) and `PHIS_d` (GLL nodes). This is not ideal as, rough topography files only exist for `*np4pg2_*` type files in https://web.lcrc.anl.gov/public/e3sm/inputdata/atm/cam/topo/ collection. This caused a silent bug documented in issue https://github.com/E3SM-Project/scream/issues/2222 (fixed here).

Additionally, unset `topography_filename` if user request `*-X6T` compset, but on a grid which no rough topography file exists. This will throw an error if the user does not provide `phis`.

I tested on summit: 
- `ERS_Ln22.ne30_ne30.F2010-SCREAMv1` (fail, diffs)
- `ERS_Ln22.ne30pg2_ne30pg2.F2010-SCREAMv1-X6T` (pass)
 
The reason for the `ERS_Ln22.ne30_ne30.F2010-SCREAMv1` diffs is that the topo files change and `PHIS` in the np4-only file (used on master) does not match `PHIS_d` in the np4pg2 file (used in this PR). The ne4 tests in the AT should pass since I've verified `PHIS` in the np4-only file matches `PHIS_d` in the np4pg2 file for ne4.

### DIFF
--- a/components/eamxx/src/control/atmosphere_driver.cpp
+++ b/components/eamxx/src/control/atmosphere_driver.cpp
@@ -877,17 +877,18 @@ void AtmosphereDriver::set_initial_conditions ()
 
       if (fname == "phis") {
         // Topography (phis) is a special case that should
-        // be loaded from the topography file, where the 
-	// eamxx field "phis" corresponds to the name 
-	// "PHIS_d" on the GLL grid and "PHIS" on the PG2 
-	// grid in the topography file.
+        // be loaded from the topography file, where the
+        // eamxx field "phis" corresponds to the name
+        // "PHIS_d" on the GLL and Point grids and "PHIS"
+        // on the PG2 grid in the topography file.
         if (grid_name == "Physics PG2") {
           this_grid_topo_file_fnames.push_back("PHIS");
-        } else if (grid_name == "Physics GLL") {
+        } else if (grid_name == "Physics GLL" ||
+                   grid_name == "Point Grid") {
           this_grid_topo_file_fnames.push_back("PHIS_d");
         } else {
           EKAT_ERROR_MSG ("Error! Requesting phis on an unknown grid: " + grid_name + ".\n");
-	}
+        }
         this_grid_topo_eamxx_fnames.push_back("phis");
       } else if (c.size()==0) {
         // If this field is the parent of other subfields, we only read from file the subfields.
--- a/components/eamxx/src/control/atmosphere_driver.cpp
+++ b/components/eamxx/src/control/atmosphere_driver.cpp
@@ -843,8 +843,8 @@ void AtmosphereDriver::set_initial_conditions ()
   std::map<std::string,std::vector<std::string>> fields_inited;
 
   // Check which fields should be loaded from the topography file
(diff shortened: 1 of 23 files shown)


tcclevenger:
@AaronDonahue This changes the way ne1024 runs grab default input.yaml values, both np4, pg2, and rough topo. We may want to run a debug test to verify everything is correct.

mt5555:
One thought: if you made the topo reading code flexible to look for PHIS_d, but if not found, use PHIS, then all the NP4 tests can keep using their older topo files, and you dont have to change all the test code, and then there will be many fewer DIFFs.

I just mention it if avoid DIFFs is worth the trouble.  If not, it's cleaner if all simulations always require a modern topography file with both PHIS and PHIS_d in it.

tcclevenger:
@mt5555 I'm wondering what the solution will be with grids which do not have a rough topo file. For instance, ne4 is used for the CIME tests in the AT, but no `x6t` topo file exists for ne4. We could keep the topo file the same for ne4 (use `x16`) and then have logic to reset the `pgrad_correction=hv_ref_profiles=0`.

`x6t` files also do not exist for ne120 and ne512. Some options"
1. UNSET the topo file in those case (simulation would error out),
2.  Use smooth topo for these cases,
3.  Upload `x6t` files for these cases (if they exist somewhere)?

Either way, I agree that this PR should change to "use np4pg2 files for all cases" (BFB) and then changing the default behavior should be in a separate PR.

components/eamxx/src/control/atmosphere_driver.cpp
```diff
@@ -870,22 +870,26 @@ void AtmosphereDriver::set_initial_conditions ()
...
-          this_grid_topo_fnames_nc.push_back("PHIS_d");
-          this_grid_topo_fnames_eamxx.push_back("phis");
+        // be loaded from the topography file, where the
+        // eamxx field "phis" corresponds to the name
+        // "PHIS_d" on the GLL and Point grids and "PHIS"
+        // on the PG2 grid in the topography file.
+        if (grid_name == "Physics PG2") {
+          this_grid_topo_file_fnames.push_back("PHIS");
+        } else if (grid_name == "Physics GLL" ||
+                   grid_name == "Point Grid") {
```
bartgol, line 887: Side comment: `Point Grid` is just for unit tests, right? At some point we should make this more explicit, so we clearly recognize pieces of code that are not relevant for normal runs...

bartgol (approved):

components/eamxx/src/control/atmosphere_driver.cpp
tcclevenger: Yeah, I was going to have a separate `else if (PointGrid)`, but it duplicated code. I'm willing to put it back in.

bartgol: No, it's fine. I just wanted to make sure we were not using this kind of generic/confusing names in production code.

PR #2247, opened 2023-03-15
Include surface quantities in nudging process

AaronDonahue:
This commit adds the set of fields passed to the surface models as available nudging fields.  Note, this PR is mid-point towards having a more general nudging process that can have it's nudged fields set at runtime.

That will be taken care of in a subsequent PR.


AaronDonahue:
WIP for now, I would like to make the testing of the new variables more robust.  In the meantime, @rebassoo - if you could verify that what i did looks correct, and @PeterCaldwell if you can verify the list of surface variables are truly the ones we want be nudging (and that I haven't missed any).

PeterCaldwell:
At first glance, I'm surprised you are only replacing radiation variables and I'm dubious that you need to replace *net* radiation variables since I think the land model needs to supply the upward fluxes at the surface. I would have expected that you'd need to pass lowest-level T, qv, and winds. Basically, I think we need to look in the coupler files for what the land model needs from the atm. I think that info is in the first 2 tables of https://acme-climate.atlassian.net/wiki/spaces/NGDNA/pages/961184622/Fields+passed+to+from+component+coupler (?).

AaronDonahue:
@PeterCaldwell I created this Confluence page to assist in the surface coupling vars audit: https://acme-climate.atlassian.net/wiki/spaces/EAMxxML/pages/3715924076/List+of+Nudged+Variables

rebassoo:
Looks good to me, I think you need to add the surface fields to the time interpolation function as well.

AaronDonahue:
I'm closing this, after discussion with @PeterCaldwell we determined a better strategy for dealing with the surface quantities that will need to be replaced when nudging the state.

PR #2251, opened 2023-03-18
EAMxx: fix timestamp of FV fields for Initial run in HommeDynamics

### LINKED ISSUE #2250: PEM test failure in nightlies
After merging #2216, the PEM test in our nightlies started to fail across the board. I reorganized the commits in that branch into few self-consistent commits (meaning each commit contains one feature add/change, and code would compile at every commit), and via bisection landed on [this commit](https://github.com/E3SM-Project/scream/pull/2216/commits/b1485a4e64f6c08210003c950a3f35751b080745) as possible culprit.

That commit is in theory innocuous, in that it moves the init of output streams till after atm procs have been inited. The reason being that, otherwise, any field inited inside atm procs during `initialize_atm_procs` (and [tke is an example](https://github.com/E3SM-Project/scream/blob/2d07997c12ac74fc00df86fafadc88366afc7c64/components/eamxx/src/physics/shoc/atmosphere_macrophysics.cpp#L281)) would not be reflected in the t=0 output (which, to be fair, only happens for INSTANT output streams).

Now, the initialization of output mgrs _can_ set a field to `m_fill_value` if the field was not yet inited (true for most fields that are not in the IC file). My first guess was that by initing these fields to 0 during `initialize_output_managers`, we were hiding the fact that one of those fields was actually supposed to be inited via IC file but somehow wasn't. However, if that was the case, all tests should fail, including ERS/ERP/SMS.

bartgol:
The tendencies PR introduced a bug by doing something seemingly innocuous (in [this](https://github.com/E3SM-Project/scream/pull/2216/commits/b1485a4e64f6c08210003c950a3f35751b080745) commit). This PR fixes the exposed inconsistency. I verified the fix works by running SMS_Ln90.ne30pg2_ne30pg2.F2010-SCREAMv1 on mappy.

What happened is summarized in the following points.

- For PG2 runs, AD reads state on a physics GLL grid, rather than on PG2 grid.
- HommeDynamics takes care of initing FV state fields, but was not initializing the timestamp of those fields.
- This used to be innocuous, since the timesamp would have been updated at the very first timestep.
- The linked commit changed the order of Output Mgr (OM) init and Atm Procs (AP) init, so that OM init would happen after AP init.
- INSTANT output always outputs fields at t=0, and if it finds an uninited field, it _assumes_ it's just something not yet computed (e.g. some fields computed by physics, like `tke`), and proceeds to set the field equal to `m_fill_value`.
- Result: when OM ran after Homme init, it would see FV copies of state as un-inited, and set them to `m_fill_value`, wreacking havoc.

Fixes #2250

### DIFF
--- a/components/eamxx/src/dynamics/homme/atmosphere_dynamics_fv_phys.cpp
+++ b/components/eamxx/src/dynamics/homme/atmosphere_dynamics_fv_phys.cpp
@@ -124,6 +124,21 @@ void HommeDynamics::fv_phys_dyn_to_fv_phys (const bool restart) {
     const auto T  = get_field_out("T_mid",pgn).get_view<const Pack**>();
     const auto uv = get_field_out("horiz_winds",pgn).get_view<const Pack***>();
     copy_prev(ncols, npacks, T, uv, FT, FM);
+
+    // In an initial run, the AD only reads IC for the Physics GLL fields,
+    // and this class has just taken care of remapping them to the FV grid.
+    // Therefore, the timestamp of the FV fields has *not* been set yet,
+    // which can cause serious issues downstream. For details, see
+    //   https://github.com/E3SM-Project/scream/issues/2250
+    // To avoid any problem, we simply set the timestamp of FV state fields here.
+    // NOTE: even if the init sequence *ever* changed, and the AD *did* read
+    // IC for FV fields, this step remains safe (we're setting the same t0)
+    for (auto n : {"T_mid","horiz_winds","ps","phis","omega","pseudo_density"}) {
+      auto f = get_field_out(n,pgn);
+      f.get_header().get_tracking().update_time_stamp(timestamp());
+    }
+    auto Q = get_group_out("tracers",pgn).m_bundle;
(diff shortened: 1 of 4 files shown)


ambrad (approved):

ambrad:
Am I correct that until the recent tendencies PR, this particular timestamp was inert in the code, i.e., had no effect on the solution?

bartgol:
> Am I correct that until the recent tendencies PR, this particular timestamp was inert in the code, i.e., had no effect on the solution?

Yes, there should have been no impact in the solution. The swap of OM_init - AP_init (together with the recent PR that added t=0 output) is what made the non-init of that timestamp relevant. All results obtained so far can still be trusted (to the best of my knowledge).

ambrad:
One note in case this wasn't clear. The routine modified in this PR is called in the Hommexx interface's post-processing phase at each time step. It's possible the intention here was to set the time stamp only in initialization, in which case the code here should be moved to the calling function in atmosphere_dynamics.cpp. I suspect the time step modifications during time stepping are redundant and thus not harmful.

PR #2259, opened 2023-03-23
Fix restart with radiation super-stepping

### LINKED ISSUE #2201: Non-BFB across restarts due to radiation superstepping
EAMxx is not BFB across restarts when radiation supersteps don’t result in radiation being called on the first restarted timestep

brhillman:
Fix restart with radiation super-stepping. Previously, restarts would be non-bfb if super-stepping radiation and trying to restart on a step in which radiation is not computed. This was due to the heating and surface fluxes not being written to restart files, since they were tagged as computed rather than updated fields. These fields are added to the list of updated fields here so that they will automatically be restarted. A testmod is also added to change the radiation frequency, and a corresponding test modified in the v1 low-res test suite to catch this case. This also changes the values of the fields that were changed to updated on the zeroth step (because they are initialized to zero instead of fill values), so tests that include output on the zeroth step will show DIFFs. These changes should otherwise be BFB. Fixes #2201 

[Non-B4B]

### DIFF
--- a/components/eamxx/src/physics/rrtmgp/atmosphere_radiation.cpp
+++ b/components/eamxx/src/physics/rrtmgp/atmosphere_radiation.cpp
@@ -129,7 +129,7 @@ void RRTMGPRadiation::set_grids(const std::shared_ptr<const GridsManager> grids_
   add_field<Computed>("SW_clrsky_flux_dn_dir", scalar3d_layout_int, Wm2, grid_name, ps);
   add_field<Computed>("LW_clrsky_flux_up", scalar3d_layout_int, Wm2, grid_name, ps);
   add_field<Computed>("LW_clrsky_flux_dn", scalar3d_layout_int, Wm2, grid_name, ps);
-  add_field<Computed>("rad_heating_pdel", scalar3d_layout_mid, Pa*K/s, grid_name, ps);
+  add_field<Updated>("rad_heating_pdel", scalar3d_layout_mid, Pa*K/s, grid_name, ps);
   // Cloud properties added as computed fields for diagnostic purposes
   add_field<Computed>("cldlow"        , scalar2d_layout, nondim, grid_name);
   add_field<Computed>("cldmed"        , scalar2d_layout, nondim, grid_name);
@@ -147,12 +147,13 @@ void RRTMGPRadiation::set_grids(const std::shared_ptr<const GridsManager> grids_
   // netsw      sfc_flux_sw_net    net (down - up) SW flux at surface
   // flwds      sfc_flux_lw_dn     downwelling LW flux at surface
   // --------------------------------------------------------------
-  add_field<Computed>("sfc_flux_dir_nir", scalar2d_layout, Wm2, grid_name);
-  add_field<Computed>("sfc_flux_dir_vis", scalar2d_layout, Wm2, grid_name);
-  add_field<Computed>("sfc_flux_dif_nir", scalar2d_layout, Wm2, grid_name);
-  add_field<Computed>("sfc_flux_dif_vis", scalar2d_layout, Wm2, grid_name);
-  add_field<Computed>("sfc_flux_sw_net" , scalar2d_layout, Wm2, grid_name);
-  add_field<Computed>("sfc_flux_lw_dn"  , scalar2d_layout, Wm2, grid_name);
+  // These need to be added to restarts in the case of super-stepping
+  add_field<Updated>("sfc_flux_dir_nir", scalar2d_layout, Wm2, grid_name);
+  add_field<Updated>("sfc_flux_dir_vis", scalar2d_layout, Wm2, grid_name);
+  add_field<Updated>("sfc_flux_dif_nir", scalar2d_layout, Wm2, grid_name);
+  add_field<Updated>("sfc_flux_dif_vis", scalar2d_layout, Wm2, grid_name);
+  add_field<Updated>("sfc_flux_sw_net" , scalar2d_layout, Wm2, grid_name);
+  add_field<Updated>("sfc_flux_lw_dn"  , scalar2d_layout, Wm2, grid_name);
 
   // Boundary flux fields for energy and mass conservation checks
   if (has_column_conservation_check()) {
--- a/components/eamxx/src/physics/rrtmgp/atmosphere_radiation.cpp
+++ b/components/eamxx/src/physics/rrtmgp/atmosphere_radiation.cpp
@@ -129,7 +129,7 @@ void RRTMGPRadiation::set_grids(const std::shared_ptr<const GridsManager> grids_
   add_field<Computed>("SW_clrsky_flux_dn_dir", scalar3d_layout_int, Wm2, grid_name, ps);
   add_field<Computed>("LW_clrsky_flux_up", scalar3d_layout_int, Wm2, grid_name, ps);
   add_field<Computed>("LW_clrsky_flux_dn", scalar3d_layout_int, Wm2, grid_name, ps);
-  add_field<Updated>("rad_heating_pdel", scalar3d_layout_mid, Pa*K/s, grid_name, ps);
+  add_field<Computed>("rad_heating_pdel", scalar3d_layout_mid, Pa*K/s, grid_name, "RESTART", ps);
   // Cloud properties added as computed fields for diagnostic purposes
   add_field<Computed>("cldlow"        , scalar2d_layout, nondim, grid_name);
   add_field<Computed>("cldmed"        , scalar2d_layout, nondim, grid_name);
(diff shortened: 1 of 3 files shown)


components/eamxx/cime_config/namelist_defaults_scream.xml
```diff
@@ -348,6 +348,15 @@ be lost if SCREAM_HACK_XML is not enabled.
     <nc_nuceat_tend    >0.0</nc_nuceat_tend>
     <!-- SHOC internally inits TKE; just set to 0 here to signal it's not in the IC file -->
     <tke               >0.0</tke>
+    <!-- Rad heating and sfc fluxes are defined as "updated" so it will be written to restart files, so we   -->
```
AaronDonahue, line 351: I'm adding a comment here just to ping @bartgol with a question.

Did we ever fix the code so that a variable can be initialized in the `init_impl` of a process and thus the AD won't try to load it from file.  I recall a long conversation about this when it came to SHOC, and we have this chunk of code: https://github.com/E3SM-Project/scream/blob/master/components/eamxx/src/physics/shoc/atmosphere_macrophysics.cpp#L278

AaronDonahue (dismissed):
I pinged @bartgol with a question, but the answer shouldn't hold up this fix.

brhillman:
Unit tests need updates to yaml to handle initialization of Updated fields, fix coming.

components/eamxx/cime_config/namelist_defaults_scream.xml
brhillman: Yeah, this would be handy. I just had to add initialization for all of these changed fields to all of the input.yaml files in all of the tests that involve radiation. Not really a huge deal, but being able to tell the AD not to try to initialize fields that will be initialized by the individual processes would be nice. The way this is handled in EAM is that fields are assumed to be initialized in the individual parameterizations if they are not explicitly initialized via the initial condition file. Assuming is probably never the best idea, but maybe one solution would be to add some operation that tells the AD that a field has been initialized once it has been called. The AD could then check *after* all processes have ran init whether every field tagged as "updated" or "required"  has been initialized, either by IC, yaml constant initialization, or by calling this yet un-named initialization routine.

brhillman: By the way, I discovered what I think is kind of a weird bug in working on this. It seems that with a pg2 grid, the AD will NOT complain if a field tagged as "Updated" is not initialized, but with a np4 grid, I would get runtime fails without the extra initialization in the input.yaml for fields that were not in the IC. That is, this would fail:
```
ERS_D_Ln21.ne4_ne4.F2010-SCREAMv1.mappy_gnu9.scream-rad_frequency_2
```
but this would pass:
```
ERS_D_Ln21.ne4pg2_ne4pg2.F2010-SCREAMv1.mappy_gnu9.scream-rad_frequency_2
```

brhillman:
mappy is reporting DIFFs. The first DIFF is expected due to the test change, but the second does not really make sense to me. This is in the TestStatus.log in the failing case directory:
```
    cat /ascldap/users/e3sm-jenkins/acme/scratch/SMS_D_Ln9.ne4_ne4.F2010-SCREAMv1-noAero.mappy_gnu9.C.20230323_172527_g2eurq/run/SMS_D_Ln9.ne4_ne4.F20
10-SCREAMv1-noAero.mappy_gnu9.C.20230323_172527_g2eurq.scream.hi.0001-01-01-00000.nc.base.cprnc.out
  comparing model 'elm'
    no hist files found for model elm
  comparing model 'cice'
    no hist files found for model cice
  comparing model 'docn'
    no hist files found for model docn
  comparing model 'srof'
    no hist files found for model srof
  comparing model 'sglc'
    no hist files found for model sglc
  comparing model 'swav'
    no hist files found for model swav
  comparing model 'siac'
    no hist files found for model siac
  comparing model 'sesp'
    no hist files found for model sesp
  comparing model 'cpl'
    SMS_D_Ln9.ne4_ne4.F2010-SCREAMv1-noAero.mappy_gnu9.C.20230323_172527_g2eurq.cpl.hi.0001-01-01-32400.nc matched cpl.hi.0001-01-01-32400.nc
FAIL
  Most recent bless: sha:unknown date:2023-03-16_16:30:29

 ---------------------------------------------------
2023-03-23 17:31:36: tail -n20 /ascldap/users/e3sm-jenkins/acme/scratch/SMS_D_Ln9.ne4_ne4.F2010-SCREAMv1-noAero.mappy_gnu9.C.20230323_172527_g2eurq/run/SMS_D_Ln9.ne4_ne4.F2010-SCREAMv1-noAero.mappy_gnu9.C.20230323_172527_g2eurq.cpl.hi.0001-01-01-32400.nc.cprnc.out

                 866  (     1,     1,     1) (     1,     1,     1)
                 722   0.000000000000000E+00   0.000000000000000E+00
                 722   0.000000000000000E+00   0.000000000000000E+00
                 866  (     1,     1,     1) (     1,     1,     1)
          avg abs field values:    0.000000000000000E+00
                                   0.000000000000000E+00
************************************************************************************************************************************
  
SUMMARY of cprnc:
 A total number of    423 fields were compared
          of which      0 had non-zero differences
               and      0 had differences in fill patterns
               and      0 had different dimension sizes
 A total number of      0 fields could not be analyzed
 A total number of      0 time-varying fields on file 1 were not found on file 2.
 A total number of      0 time-constant fields on file 1 were not found on file 2.
 A total number of      0 time-varying fields on file 2 were not found on file 1.
 A total number of      0 time-constant fields on file 2 were not found on file 1.
  diff_test: the two files seem to be IDENTICAL 
```

brhillman:
Some more evidence:
```
[bhillma@mappy rrtmgp]$ ncmax sfc_flux_sw_net /ascldap/users/e3sm-jenkins/acme/scratch/SMS_D_Ln9.ne4_ne4.F2010-SCREAMv1-noAero.mappy_gnu9.C.20230323_172527_g2eurq/run/SMS_D_Ln9.ne4_ne4.F2010-SCREAMv1-noAero.mappy_gnu9.C.20230323_172527_g2eurq.scream.hi.0001-01-01-00000.nc.base
1013.32 

[bhillma@mappy rrtmgp]$ ncmax sfc_flux_sw_net /sems-data-store/ACME/baselines/mappy/gnu9/master/SMS_D_Ln9.ne4_ne4.F2010-SCREAMv1-noAero.mappy_gnu9/scream.hi.0001-01-01-00000.nc.base
3.40282e+33 
```
So the diff is because in current master, `sfc_flux_sw_net` and `sfc_flux_lw_dn` get garbage values in some spots for some reason, but initializing these fields to zero fixes this. Note that this case does NOT use super-stepping, so this should not be due to skipped rad steps.

brhillman:
Ah, so the garbage values are on the first output step, which looks like it is maybe output before radiation is run (time = 0). So initializing the fluxes to zero makes these outputs all zero on the first output step, rather than garbage uninitialized data. This didn't show up in the other tests in this suite because those tests do not appear to write scream history files with outputs on step 0, only rhist files. So I think the diffs here are all expected and should be blessed after merging. This PR should be ready to merge.

AaronDonahue:
> Ah, so the garbage values are on the first output step, which looks like it is maybe output before radiation is run (time = 0). So initializing the fluxes to zero makes these outputs all zero on the first output step, rather than garbage uninitialized data. This didn't show up in the other tests in this suite because those tests do not appear to write scream history files with outputs on step 0, only rhist files. So I think the diffs here are all expected and should be blessed after merging. This PR should be ready to merge.

Re: garbage values.  I believe those are initialized to the Fill-Value which is a machine precision number.  Just for clarification :-)

AaronDonahue (approved):

components/eamxx/cime_config/namelist_defaults_scream.xml
AaronDonahue: Did you ever get to the bottom of this weird bug?  If not, can you make an issue so we don't lose sight of it?

AaronDonahue:
@brhillman , looks good to me.  Quick question on the DIFFs.  Only a few tests DIFF, do you know why the others are not impacted by the initial condition value being 0 vs. Fill_Value?

brhillman:
> @brhillman , looks good to me. Quick question on the DIFFs. Only a few tests DIFF, do you know why the others are not impacted by the initial condition value being 0 vs. Fill_Value?

Because the others do not write scream.hi files, only scream.rhist files, which would not contain the first (Fill) values.

brhillman:
> > Ah, so the garbage values are on the first output step, which looks like it is maybe output before radiation is run (time = 0). So initializing the fluxes to zero makes these outputs all zero on the first output step, rather than garbage uninitialized data. This didn't show up in the other tests in this suite because those tests do not appear to write scream history files with outputs on step 0, only rhist files. So I think the diffs here are all expected and should be blessed after merging. This PR should be ready to merge.
> 
> Re: garbage values. I believe those are initialized to the Fill-Value which is a machine precision number. Just for clarification :-)

Okay, two points then. 1) I was surprised that that the fill value was something like the machine max for that data type. I.e., 3.40282e+33. 2) I missed this because the _FillValue attribute is not populated in these files, so the values end up looking like garbage rather than fill values. If we do not already, we should have an issue to track this (_FillValue should be written as a netcdf attribute for any field that might contain fill values. Safer to do it for all).

components/eamxx/cime_config/namelist_defaults_scream.xml
bartgol: Yes, this is due to the currently slightly hacky impl of PG2 initialization. I have to fix this at some point...

bartgol (changes requested):

components/eamxx/src/physics/rrtmgp/atmosphere_radiation.cpp
```diff
@@ -129,7 +129,7 @@ void RRTMGPRadiation::set_grids(const std::shared_ptr<const GridsManager> grids_
   add_field<Computed>("SW_clrsky_flux_dn_dir", scalar3d_layout_int, Wm2, grid_name, ps);
   add_field<Computed>("LW_clrsky_flux_up", scalar3d_layout_int, Wm2, grid_name, ps);
   add_field<Computed>("LW_clrsky_flux_dn", scalar3d_layout_int, Wm2, grid_name, ps);
-  add_field<Computed>("rad_heating_pdel", scalar3d_layout_mid, Pa*K/s, grid_name, ps);
+  add_field<Updated>("rad_heating_pdel", scalar3d_layout_mid, Pa*K/s, grid_name, ps);
```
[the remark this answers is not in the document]
brhillman: That seems to have done it. I removed the xml/yaml mods and just added these fields to the RESTART list, reverting them all to Computed rather than Updated.

bartgol (approved):
This looks good from the AD point of view. Whether or not the diffs are expected, I'll let you/others decide.

brhillman:
@bartgol if a field is in the RESTART group, is it automatically initialized to zero? The diff in the no-aero test is because that test is writing output, and on the first step, our baselines have missing values for the fields affected here (i.e., `sfc_flux_sw_net`), but on this branch these end up getting filled with zeros I believe. Everything is the same after the initial step, hence why the runs are still BFB (coupler history matches, and outputs match if you exclude the first step).

bartgol:
I have to check, but I would guess the asnwer is "no". In `Initial` runs, the AD should not init any field that is not explicitly requested to init (via input/xml file). In restarted runs, it should simply read the field from restart file.

brhillman:
Ah, I got confused, the diff in suspect was from before I had pushed my changes. Will see what AT reports from the running tests.

brhillman:
Okay great, looks like the only DIFF is the new/renamed test. This should be ready to merge.

AaronDonahue (approved):

AaronDonahue:
This PR has been approved and the DIFF reported by `mappy` is expected so I am going to manually merge.

PR #2261, opened 2023-03-24
add vtheta_thresh as homme parameter to eamxx namelist

AaronDonahue:
Add `vtheta_thresh` as a namelist parameter to the EAMxx namelist options for homme.

This should make it possible to change this value using `atmchange`


ndkeen (approved):
wow that was fast

PR #2269, opened 2023-03-28
Force avg type, freq/freq_units, and timestamp to be in the output NC file name

### LINKED ISSUE #2202: Insist on including ave type and freq in output filename
Currently, users can elect to omit output frequency or averaging type from filenames. If they did so, changing freq or ave type wouldn’t start a new output stream. It would also result in name conflicts between (for example) attempts to output daily max and daily min T. Thus, we will disable the option to suppress freq and ave type in output filenames.

bartgol:
Fixes #2202


components/eamxx/data/scream_default_output.yaml
```diff
@@ -63,15 +63,7 @@ output_control:
 # WARNING: ERS/ERP tets will override this with STOP_N/STOP_OPTION
   Frequency: ${HIST_N}
   frequency_units: ${HIST_OPTION}
-  MPI Ranks in Filename: false
```
AaronDonahue, line 66: I think we should keep this line in, it alerts new users to the option exisiting even if it is default false.

bartgol: Well, I don't really want them to use it... That option is really just for unit tests...

AaronDonahue: That's fair

AaronDonahue (dismissed):

AaronDonahue:
The AT failed in `mappy` with DIFFs, I think those should be expected right?  The names of all of the output files have now changed to have the AVG type in them by default right?  @bartgol can you confirm this is the cause of the failure, and if so I think you can manually merge.

AaronDonahue:
@bartgol note that #2259 just went in and it will cause a DIFF on mappy as well.  Not sure if we want to hold off on manually merging 2 PR's that will need blessed baselines after the nightlies.  I think this PR is different enough that it shouldn't be an issue, but wanted to make sure I made a note just in case.

bartgol:
> The AT failed in `mappy` with DIFFs, I think those should be expected right? The names of all of the output files have now changed to have the AVG type in them by default right? @bartgol can you confirm this is the cause of the failure, and if so I think you can manually merge.

One ERP test failed in the RUN phase, which is not related to that, I think. I have to check what the deal is. Hold off any merge.

bartgol:
Yeah, looking at our nightlies, it seems master was not aware of any scream hist file:
```
2023-03-28 22:15:16: Comparing hists for case 'ERP_Ln22.ne4pg2_ne4pg2.F2010-SCREAMv1.mappy_gnu9.C.JMasterScream_v120230328_211943' dir1='/ascldap/users/e3sm-jenkins/acme/scratch/ERP_Ln22.ne4pg2_ne4pg2.F2010-SCREAMv1.mappy_gnu9.C.JMasterScream_v120230328_211943/run', suffix1='base',  dir2='/ascldap/users/e3sm-jenkins/acme/scratch/ERP_Ln22.ne4pg2_ne4pg2.F2010-SCREAMv1.mappy_gnu9.C.JMasterScream_v120230328_211943/run' suffix2='rest'
  comparing model 'scream'
    no hist files found for model scream
```
Looking at master, I suspect the regex `hi` for hist files should have been changed to `hist\.`. I'm trying that now.

bartgol:
> The ERP test failed cause I missed a `*` in the regex pattern for rest/hist files in config_archive. It should work now.
> 
> Regarding the other diffs, I noticed that the baseline folders of ERP/ERS/PEM tests do not contain the scream history files I'd expect. The ERS/ERP tests only gontain one `.rhist` nc file, while I would expect a `.hi` file (or maybe two, one `.base` and one `.rest`). The PEM test contains only the `.modpes` nc file.
> 
> I think this might have been a semi-bug in master, so this PR should be fine, but I'd like to be sure.

I verified that this issue is not (entirely) our fault. Although there _is_ a way for us to specify a regex for hist file names that CIME can use, it is quite ugly:
```
<hist_file_extension>(INSTANT|AVERAGE|MAX|MIN)\.n(step|sec|min|hour|day|month|year)s_x\d*\.\d{4}-\d{2}-\d{2}-\d{5}\.nc</hist_file_extension>
```
If we ever changed our format of the output file names, this regex would have to change as well. I can implement this regex in this PR, to get full check of base/rest files, and get all of our baselines copied to the baseline dir (I think, I can't be sure this would work, until I try to bless a test). But long run, it'd be neat if CIME allowed to specify an extension capturing only _part_ of the hist file name (e.g., `hist` or something).

See #esmci/cime#4382 for some more insight on this.

AaronDonahue (dismissed):
@bartgol , if you are now comfortable w/ the DIFFs on `mappy` then you can manually merge.

bartgol:
Ok, ERS/ERP tests now pass when I run them on mappy. Let's see what the AT has to say...

bartgol:
I decided to remove the commits that removed pack size requests for RRTMGP fields. To be clear, RRTMGP does not use packs, so it makes sense that doesn't request packs, but some downstream places did not like this (trcmix was one, longwave_cloud_forcing diagnostic was another). Whether or not we should remove those pack size requests, it's a topic orthogonal to this PR, and should not slow down this PR.

components/eamxx/src/physics/share/scream_trcmix.cpp
```diff
@@ -14,24 +14,34 @@ namespace physics {
...
   const Real co2vmr, const Real n2ovmr, const Real ch4vmr, const Real f11vmr, const Real f12vmr)
 {
   using C = Constants<Real>;
   using KT = KokkosTypes<DefaultDevice>;
   using ExeSpace = KT::ExeSpace;
   using MemberType = KT::MemberType;
 
-  const auto ncols = clat.extent(0);
-  const auto nlevs = pmid.extent(1);
+  const auto& p_layout = pmid.get_header().get_identifier().get_layout();
```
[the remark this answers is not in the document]
bartgol: NVM, I removed the last two commits, since I was opening a can of worms, and did not need that stuff for this PR.

bartgol:
All DIFFs on mappy are due to atm hist files not having a counterpart in the baselines dirs, as expected. This is ready for final review and merge.

components/eamxx/tests/coupled/dynamics_physics/homme_shoc_cld_p3_rrtmgp/homme_shoc_cld_p3_rrtmgp.cpp
```diff
@@ -29,18 +29,15 @@ TEST_CASE("scream_homme_physics", "scream_homme_physics") {
   // Load ad parameter list
   std::string fname = "input.yaml";
   ekat::ParameterList ad_params("Atmosphere Driver");
-  REQUIRE_NOTHROW ( parse_yaml_file(fname,ad_params) );
-
-  // Time stepping parameters
+  parse_yaml_file(fname,ad_params);
   ad_params.print();
```
AaronDonahue, line 33: Not important, but do we really need to print the input.yaml to screen?

AaronDonahue (approved):

AaronDonahue:
@bartgol , I'm assuming that the only fails are expected DIFFs on `mappy` because this PR changes the format of output file names right?  If so, I think you can manually merge when you are ready.

PR #2272, opened 2023-03-29
Allow user to prescribe atmosphere surface forcing to other components

AaronDonahue:
This PR introduces an option to override the exported surface fluxes from the ATM to CPL with constant values.

This is a first step to prescribing exported values using data from a file.  A task that will be tackled in a subsequent PR.

This is a functionality that has been requested by the ML nudging effort.  A part of that workflow will require the surface
to be forced by values taken from a file.

This functionality may be generally useful for any case where we want to explicitly force some or all of the ATM
surface fluxes to be a specific value.

### DIFF
--- a/components/eamxx/cime_config/namelist_defaults_scream.xml
+++ b/components/eamxx/cime_config/namelist_defaults_scream.xml
@@ -165,7 +165,7 @@ be lost if SCREAM_HACK_XML is not enabled.
     <sc_export inherit="atm_proc_base">
       <prescribed_constants>
         <fields type="array(string)">NONE</fields>
-        <values type="array(double)">NONE</values>
+        <values type="array(real)">0</values>
       </prescribed_constants>
     </sc_export>
 
--- a/components/eamxx/cime_config/namelist_defaults_scream.xml
+++ b/components/eamxx/cime_config/namelist_defaults_scream.xml
@@ -162,7 +162,9 @@ be lost if SCREAM_HACK_XML is not enabled.
 
     <!-- Surface coupling (import and export) -->
     <sc_import inherit="atm_proc_base"/>
-    <sc_export inherit="atm_proc_base"/>
+    <sc_export inherit="atm_proc_base">
+      <prescribed_export_control_file type="string">NONE</prescribed_export_control_file>
+    </sc_export>
 
     <!-- Homme dynamics -->
     <homme inherit="atm_proc_base">
--- a/components/eamxx/cime_config/namelist_defaults_scream.xml
+++ b/components/eamxx/cime_config/namelist_defaults_scream.xml
@@ -163,7 +163,10 @@ be lost if SCREAM_HACK_XML is not enabled.
     <!-- Surface coupling (import and export) -->
     <sc_import inherit="atm_proc_base"/>
     <sc_export inherit="atm_proc_base">
-      <prescribed_export_control_file type="string">NONE</prescribed_export_control_file>
+      <prescribed_constants>
+        <fields type="array(string)">NONE</fields>
+        <values type="array(double)">NONE</values>
+      </prescribed_constants>
     </sc_export>
 
     <!-- Homme dynamics -->
--- a/components/eamxx/src/control/atmosphere_surface_coupling_exporter.cpp
+++ b/components/eamxx/src/control/atmosphere_surface_coupling_exporter.cpp
@@ -1,5 +1,6 @@
 #include "atmosphere_surface_coupling_exporter.hpp"
 
+#include "ekat/ekat_parse_yaml_file.hpp"
 #include "ekat/ekat_assert.hpp"
 #include "ekat/util/ekat_units.hpp"
 
@@ -135,10 +136,6 @@ void SurfaceCouplingExporter::setup_surface_coupling_data(const SCDataManager &s
   m_num_cpl_exports    = sc_data_manager.get_num_cpl_fields();
   m_num_scream_exports = sc_data_manager.get_num_scream_fields();
 
-  // Allocate enum export source view
-  m_export_source = view_1d<DefaultDevice,ExportType>("",m_num_scream_exports);
-  Kokkos::deep_copy(m_export_source,EAMXX);  // The default is that all export variables will be derived from the EAMxx state.
-
   EKAT_ASSERT_MSG(m_num_scream_exports <= m_num_cpl_exports,
                   "Error! More SCREAM exports than actual cpl exports.\n");
   EKAT_ASSERT_MSG(m_num_cols == sc_data_manager.get_field_size(), "Error! Surface Coupling exports need to have size ncols.");
@@ -172,10 +169,6 @@ void SurfaceCouplingExporter::setup_surface_coupling_data(const SCDataManager &s
 void SurfaceCouplingExporter::initialize_impl (const RunType /* run_type */)
 {
   bool any_initial_exports = false;
-
-  // Set the number of exports from eamxx or set to a constant
-  m_num_eamxx_exports = m_num_scream_exports;
-  m_num_const_exports = 0;
 
   for (int i=0; i<m_num_scream_exports; ++i) {
 
@@ -211,6 +204,36 @@ void SurfaceCouplingExporter::initialize_impl (const RunType /* run_type */)
   // Copy data to device for use in do_export()
   Kokkos::deep_copy(m_column_info_d, m_column_info_h);
 
+  // Set the number of exports from eamxx or set to a constant
+  m_export_source = view_1d<DefaultDevice,ExportType>("",m_num_scream_exports);
+  Kokkos::deep_copy(m_export_source,EAMXX);  // The default is that all export variables will be derived from the EAMxx state.
+  m_num_eamxx_exports = m_num_scream_exports;
+  auto prescribed_export_control = m_params.get<std::string>("prescribed_export_control_file","NONE");
+  if (prescribed_export_control != "NONE") {
+    // There is a separate file that controls how some export fields will be handled.
+    ekat::ParameterList export_params("Export Control");
+    parse_yaml_file(prescribed_export_control,export_params);
+    export_params.print(); //ASD
+    if (export_params.isSublist("set_export_to_constant")) {
+      m_export_constants = view_1d<DefaultDevice,Real>("",m_num_scream_exports);
+      // Some of the vars will be set to a constant value
+      auto constant_exp = export_params.sublist("set_export_to_constant");
+      for (int i=0; i<m_num_scream_exports; ++i) {
+        std::string fname = m_export_field_names[i];
+        if (constant_exp.isParameter(fname)) {
+          EKAT_REQUIRE_MSG(m_export_source(i)==EAMXX,"Error! surface_coupling_exporter - '" + fname + "' is being set in " + prescribed_export_control + " more than once.");
+          m_export_source(i) = CONSTANT;
+          m_num_const_exports += 1;
+          m_num_eamxx_exports -= 1;
+          m_export_constants(i) = constant_exp.get<Real>(fname);
+        }
+      }
+    }
+  } 
+  // Final sanity check
+  EKAT_REQUIRE_MSG(m_num_scream_exports = m_num_const_exports+m_num_eamxx_exports,"Error! surface_coupling_exporter - Something went wrong set the type of export for all variables.");
+  EKAT_REQUIRE_MSG(m_num_eamxx_exports>=0,"Error! surface_coupling_exporter - The number of exports derived from EAMxx < 0, something must have gone wrong in assigning the types of exports for all variables.");
+
   // Perform initial export (if any are marked for export during initialization)
   if (any_initial_exports) do_export(0, true);
 }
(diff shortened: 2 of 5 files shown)


components/eamxx/cime_config/namelist_defaults_scream.xml
```diff
@@ -162,7 +162,9 @@ be lost if SCREAM_HACK_XML is not enabled.
 
     <!-- Surface coupling (import and export) -->
     <sc_import inherit="atm_proc_base"/>
-    <sc_export inherit="atm_proc_base"/>
+    <sc_export inherit="atm_proc_base">
+      <prescribed_export_control_file type="string">NONE</prescribed_export_control_file>
```
bartgol, line 166: I know you're still in dev phase, but to avoid rethinking this later when code is bigger: I would prefer keeping these export options _inside_ this XML section, rather than delegating to another input file.

AaronDonahue: I thought about this too and hopefully set up the code so if we ultimately decide to go this route it's not a big change.  For now I went w/ a separate file because I still don't know how complex the parameters will be for the case where we will force the export values w/ data from a set of files.

AaronDonahue: okay, I changed it.

AaronDonahue:
Looks like the only test fail is on `weaver`, which is not surprising since I haven't tested the changes on GPU.  I'm having difficulties working on weaver, I'll try to sort those out this morning so I can fix the broken test.

components/eamxx/src/control/atmosphere_surface_coupling_exporter.cpp
```diff
@@ -1,5 +1,6 @@
 #include "atmosphere_surface_coupling_exporter.hpp"
 
+#include "ekat/ekat_parse_yaml_file.hpp"
```
bartgol, line 3: Stranded include?

components/eamxx/src/control/atmosphere_surface_coupling_exporter.cpp
```diff
@@ -46,28 +47,41 @@ void SurfaceCouplingExporter::set_grids(const std::shared_ptr<const GridsManager
...
-  add_field<Required>("sfc_flux_sw_net" ,     scalar2d_layout,      Wm2,   grid_name);
-  add_field<Required>("sfc_flux_lw_dn"  ,     scalar2d_layout,      Wm2,   grid_name);
+  add_field<Required>("p_int",                scalar3d_layout_int,  Pa,     grid_name);
+  add_field<Required>("pseudo_density",       scalar3d_layout_mid,  Pa,     grid_name, ps);
+  add_field<Required>("phis",                 scalar2d_layout,      m2/s2,  grid_name);
+  add_field<Required>("p_mid",                scalar3d_layout_mid,  Pa,     grid_name, ps);
+  add_field<Required>("qv",                   scalar3d_layout_mid,  Qunit,  grid_name, "tracers", ps);
+  add_field<Required>("T_mid",                scalar3d_layout_mid,  K,      grid_name, ps);
+  add_field<Required>("horiz_winds",          vector3d_layout,      m/s,    grid_name);
+//ASD  add_field<Required>("u",                    scalar3d_layout_mid,  m/s,    grid_name); // TODO: These appear to not work yet
```
bartgol, line 57: Yeah, the u/v fields are created by the AD _after_ all fields have been created. If you declare them here, the AD will think that they are unrelated to `horiz_winds`. I don't have a good way to fix this yet.

components/eamxx/src/control/atmosphere_surface_coupling_exporter.cpp
```diff
@@ -196,6 +204,40 @@ void SurfaceCouplingExporter::initialize_impl (const RunType /* run_type */)
...
+    EKAT_REQUIRE_MSG(export_constant_fields.size()==export_constant_values.size(),"Error! surface_coupling_exporter::init - prescribed_constants 'fields' and 'values' are not the same size");
+    if (export_constant_fields.size()>0) {
+      // Determine which fields need constants
+      m_export_constants = view_1d<DefaultDevice,Real>("",m_num_scream_exports);
+      for (int i=0; i<m_num_scream_exports; ++i) {  // TODO: This loop would probably be simpler if we just checked which "i" corresponded to each name in the fields list.
+        std::string fname = m_export_field_names[i];
+        auto loc = std::find(export_constant_fields.begin(),export_constant_fields.end(),fname);
+        if (loc != export_constant_fields.end()) {
+          const auto pos = loc-export_constant_fields.begin();
+          m_export_source(i) = CONSTANT;
```
bartgol, line 229: This one and the one 3 lines below are likely the reasons of the fail on weaver. You're accessing dev views from host. You need a host mirror, and then deep copy.

components/eamxx/src/control/atmosphere_surface_coupling_exporter.cpp
```diff
@@ -207,47 +249,94 @@ void SurfaceCouplingExporter::run_impl (const double dt)
...
+
+  // Finish up exporting vars
+  do_export_to_cpl(called_during_initialization);
+}
+// =========================================================================================
+void SurfaceCouplingExporter::do_export_constant(const double dt, const bool called_during_initialization)
+{
+  // Cycle through those fields that will be set to a constant value:
+  for (int i=0; i<m_num_scream_exports; ++i) {
+    if (m_export_source(i)==CONSTANT) {
```
bartgol, line 267: Another access of dev memory from host.

components/eamxx/src/control/atmosphere_surface_coupling_exporter.cpp
```diff
@@ -207,47 +249,94 @@ void SurfaceCouplingExporter::run_impl (const double dt)
...
+      std::string fname = m_export_field_names[i];
+      const auto field_view = m_helper_fields.at(fname).get_view<Real*>();
+      Kokkos::deep_copy(field_view,m_export_constants(i));
```
bartgol, line 270: Another one here. It looks like `m_export_constants` does not need to be a dev view, btw. You could just use a std::vector, or a std::map.

components/eamxx/src/control/atmosphere_surface_coupling_exporter.cpp
```diff
@@ -268,6 +357,7 @@ void SurfaceCouplingExporter::do_export(const double dt, const bool called_durin
     PF::calculate_z_mid(team, num_levs, z_int_i, z_mid_i);
```
bartgol, line 357: This chunk of code above is only needed if you need to export `Sa_z` and/or `Sa_tbot`, so you could do something like
```
bool compute_z = h_export_source(idx_Sa_tbot)==EAMXX || h_export_source(idx_Sa_z)==EAMXX);
parfor (..) {
  if (compute_z) {
    const Real z_surf = 0.0;
    PF::calculate_z_int(team, num_levs, dz_i, z_surf, z_int_i);
    ...
   }
};
```

components/eamxx/src/control/atmosphere_surface_coupling_exporter.hpp
```diff
@@ -111,8 +122,13 @@ class SurfaceCouplingExporter : public AtmosphereProcess
   // Number of fields in cpl data
   Int m_num_cpl_exports;
 
-  // Number of exports from SCREAM
-  Int m_num_scream_exports;
+  // Number of exports from EAMxx and how they will be handled
+  Int                               m_num_scream_exports;
+  view_1d<DefaultDevice,ExportType> m_export_source;
+  view_1d<DefaultDevice,Real>       m_export_constants;
+  std::map<std::string,int>         m_export_indices;
```
bartgol, line 129: This is unused.

bartgol (changes requested):

components/eamxx/src/control/atmosphere_surface_coupling_exporter.hpp
```diff
@@ -69,7 +77,10 @@ class SurfaceCouplingExporter : public AtmosphereProcess
   // If calling in initialize_impl(), set
   // called_during_initialization=true to avoid exporting fields
   // which do not have valid entries.
-  void do_export(const double dt, const bool called_during_initialization=false);
+  void do_export(const double dt, const bool called_during_initialization=false);             // Main export routine
+  void do_export_from_eamxx(const double dt, const bool called_during_initialization=false);  // Export vars are derived from eamxx state
+  void do_export_constant(const double dt, const bool called_during_initialization=false);    // Export vars are set to a constant
+  void do_export_to_cpl(const bool called_during_initialization=false);                       // Finish export by copying data to cpl structures.
```
tcclevenger, line 83: For naming, I think something like `compute_eamxx_exports()` `set_constant_exports()` `do_export_to_cpl()` would be better, but feel free to ignore.

bartgol: Agreed.

AaronDonahue:
@bartgol and @tcclevenger I think I've addressed all of your comments.  Feel free to re-review when you have a chance.

components/eamxx/src/control/atmosphere_surface_coupling_exporter.cpp
```diff
@@ -207,89 +249,218 @@ void SurfaceCouplingExporter::run_impl (const double dt)
...
+      const auto u_wind_i = ekat::subview(horiz_winds, i, 0); // TODO, when U and V work switch to using here instead of horiz_winds.
+      Sa_u(i)             = u_wind_i(num_levs-1);
+    }
+
+    if (m_export_source(idx_Sa_v)==EAMXX) {
+      const auto v_wind_i = ekat::subview(horiz_winds, i, 1);
+      Sa_v(i)             = v_wind_i(num_levs-1);
+    }
+
+    if (m_export_source(idx_Sa_tbot)==EAMXX) {  /// HERE
```
tcclevenger, line 403: What does the `/// HERE` signify, or was it left over?

components/eamxx/src/control/atmosphere_surface_coupling_exporter.hpp
```diff
@@ -21,6 +21,14 @@ namespace scream
  * in its list of subcomponents (the AD should make sure of this).
 */
 
+// enum to track how exported fields will be set.
+enum ExportType {
+  EAMXX      =  0,  // Variable will be derived from EAMxx state
+  PRESCRIBED =  1,  // Variable will be set given data from a file
+  CONSTANT   =  2   // Set variable to a constant value
```
tcclevenger, line 28: Would `FILE = 1` make more sense because several places we have `prescribed_constants`. I'm guessing when this option becomes available, the parameter list will have `prescribed_constant` and `prescribed_from_file` or something like that, so "prescribed" applies to both.

components/eamxx/src/control/atmosphere_surface_coupling_exporter.cpp
```diff
@@ -220,20 +219,21 @@ void SurfaceCouplingExporter::initialize_impl (const RunType /* run_type */)
...
       // Determine which fields need constants
-      m_export_constants = view_1d<DefaultDevice,Real>("",m_num_scream_exports);
       for (int i=0; i<m_num_scream_exports; ++i) {  // TODO: This loop would probably be simpler if we just checked which "i" corresponded to each name in the fields list.
         std::string fname = m_export_field_names[i];
         auto loc = std::find(export_constant_fields.begin(),export_constant_fields.end(),fname);
         if (loc != export_constant_fields.end()) {
           const auto pos = loc-export_constant_fields.begin();
-          m_export_source(i) = CONSTANT;
+          export_source_h(i) = CONSTANT;
           m_num_const_exports += 1;
```
bartgol, line 228: I think `++blah` and `--blah` are nicer to see for a C++ developer ... :)

components/eamxx/src/control/atmosphere_surface_coupling_exporter.hpp
AaronDonahue: Good point, I change to `FROM_FILE`, I wanted to avoid `FILE` alone since I think that file is a key word for other things too.

components/eamxx/src/control/atmosphere_surface_coupling_exporter.cpp
```diff
@@ -196,6 +202,42 @@ void SurfaceCouplingExporter::initialize_impl (const RunType /* run_type */)
   // Copy data to device for use in do_export()
   Kokkos::deep_copy(m_column_info_d, m_column_info_h);
 
+  // Set the number of exports from eamxx or set to a constant, default type = EAMXX
+  using vos_type = std::vector<std::string>;
+  using vor_type = std::vector<Real>;
+  m_export_source     = view_1d<DefaultDevice,ExportType>("",m_num_scream_exports);
+  auto export_source_h = Kokkos::create_mirror_view(m_export_source);
+  Kokkos::deep_copy(m_export_source,EAMXX);  // The default is that all export variables will be derived from the EAMxx state.
+  m_num_eamxx_exports = m_num_scream_exports;
```
bartgol, line 211: I think this is confusing, especially in view of EAMxx in E3SMv4 (where there will be no "scream"). But even for scream, it might be confusing ("what's the difference between eamxx and scream?" on may ask).

I think the `eamxx` part was meant to reflect the enum value `EAMXX`. Perhaps that's a bit ambiguous? Maybe we could change `eamxx/EAMXX` with `computed/COMPUTED`?

AaronDonahue: That's a fair point, I can change to `COMPUTED`, or `FROM_MODEL` may also be descriptive, since `computed` is also a bit ambiguous.  The prescribed ones are computed from somewhere too, and not all fluxes are computed in surface coupling, some are just taken from model state.

bartgol: Can we change `m_num_eamxx_exports` too then? I see this line becoming puzzling when scream becomes eamxx...

bartgol (approved):

tcclevenger (approved):

PR #2277, opened 2023-04-06
Crusher: Update netcdf modules associated with PrgEnv-cray default update.

ambrad:
Update modules after Crusher system update.


ambrad:
Tests pass on Crusher.

PR #2278, opened 2023-04-07
Add support for ADIOS

jayeshkrishna:
Adding support for ADIOS by including fixes that are specific
to SCREAM.

### DIFF
--- a/components/eamxx/src/share/io/CMakeLists.txt
+++ b/components/eamxx/src/share/io/CMakeLists.txt
@@ -14,6 +14,11 @@ set_target_properties(scream_io PROPERTIES
   Fortran_MODULE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/modules
 )
 target_include_directories(scream_io PUBLIC ${CMAKE_CURRENT_BINARY_DIR}/modules)
+
+if (DEFINED ENV{ADIOS2_DIR})
+  target_include_directories(scream_io PRIVATE $ENV{ADIOS2_DIR}/include)
+endif ()
+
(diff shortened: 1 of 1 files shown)


jayeshkrishna:
After this fix the only change required in SCREAM to start using ADIOS out of the box would be to update to the latest CIME master (or update CIME to the version being merged to E3SM in https://github.com/E3SM-Project/E3SM/pull/5581)

components/eamxx/src/share/io/CMakeLists.txt
```diff
@@ -14,6 +14,11 @@ set_target_properties(scream_io PROPERTIES
   Fortran_MODULE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/modules
 )
 target_include_directories(scream_io PUBLIC ${CMAKE_CURRENT_BINARY_DIR}/modules)
+
+if (DEFINED ENV{ADIOS2_DIR})
```
bartgol, line 18: Does scorpio handle this in the same way, namely that ADIOS support is enabled if the env var ADIOS2_DIR is defined at config time? I just want to make sure we don't run into the case where scorpio thinks it's using adios, but eamxx did not see the env var, so doesn't add the include dir.

jayeshkrishna: Yeah, SCORPIO also requires this env var set to configure with ADIOS

bartgol (approved):

bartgol:
@jayeshkrishna Unless you're planning to wait (perhaps to add more stuff) feel free to push the shiny green button whenever you're ready.

Issue #2142, opened 2023-01-31
NetCDF global attributes incomplete


ambrad:
I was looking at an EAMxx history file and noticed a number of the global attributes are potentially misleading. For example, IC and topo files were used in the simulation, but here they show as "NONE FOR NOW". The case and git_hash attributes would also be useful.
```
// global attributes:
    :source = "E3SM Atmosphere Model Version 4" ;
    :case = "TEST 1" ;
    :title = "EAMv4 History File" ;
    :git_hash = "THE GIT LOG HASH" ;
    :host = "THE HOST" ;
    :Version = "1.0" ;
    :revision_Id = "None" ;
    :initial_file = "NONE FOR NOW" ;
    :topography_file = "NONE FOR NOW" ;
    :contact = "e3sm-data-support@llnl.gov" ;
    :institution_id = "E3SM-Project" ;
    :product = "model-output" ;
    :realm = "atmos" ;
    :Conventions = "None yet" ;
```

PeterCaldwell:
@jgfouca - this "help wanted" task has your name on it (now that I've mentioned you) ;-). You don't need to do it - it isn't urgent - but I think you're probably best equipped for it since it is at the intersection between CIME and SCREAM.

PeterCaldwell:
@czender wrote on slack: "FYI the last time SCREAM files I looked at had a global attribute Conventions = "None yet" . For EAMxx to leverage some ncclimo capabilities, that attribute value will need to contain a string of the form "CF-1.X", e.g., Conventions = "CF-1.8" . Once CF is listed as a convention, NCO will understand that it should look for and propagate coordinate bounds variables like time_bnds discussed above. It might also help with other software, not sure..."

Thus we should make sure the conventions flag we add in this issue is consistent with some of the work we need to do in other output git issues...

rljacob:
"EAMv4 History File" should be "EAMxx History File"
"E3SM Atmosphere Model Version 4"  should be something like "E3SM C++ Atmosphere Model"

czender:
[...]

PR #2282, opened 2023-04-12
Add output global metadata for 'conventions'

AaronDonahue:


AaronDonahue:
Not sure, but it may cause a DIFF in cprnc in that the metadata for the baseline will be different.

PeterCaldwell:
Grepping confirms that CF-1.0 is appropriate. From Charlie in "NetCDF global attributes incomplete (Issue #2142)":

> The EAMxx output files already implement enough CF Conventions (units, calendar, standard_name, etc) to merit adding a Conventions = "CF-1.0" attribute. I do not think bounds were included in that version. Later, when EAMxx has as much CF metadata as EAM, you can bump the version to 1.7 or higher. I cannot think of any downside to havine CF-1.0 in Conventions right now, and it would help folks like @golaz who are trying EAM workflow infrastructure right now on (heavily massaged) EAMxx datasets.

components/eamxx/src/share/io/scream_output_manager.cpp
```diff
@@ -560,7 +560,7 @@ void set_file_header(const std::string& filename)
   set_str_attribute_c2f(filename.c_str(),"institution_id","");  // TODO
   set_str_attribute_c2f(filename.c_str(),"product","");  // TODO
   set_str_attribute_c2f(filename.c_str(),"component","ATM");
-  set_str_attribute_c2f(filename.c_str(),"conventions","");  // TODO
+  set_str_attribute_c2f(filename.c_str(),"conventions","CF-1.0");  // TODO: It might be nice to have this passed from CIME.  We hard-code for now though, because post-processing needs something in this global attribute. 2023-04-12
```
PeterCaldwell, line 563: One comment on a one-line PR: I disagree with the comment that CIME should tell us the CF convention value... that value is determined by what output units, what time bounds, and what other metadata we have. That's EAMxx's to determine, not CIME.

golaz (approved):

components/eamxx/src/share/io/scream_output_manager.cpp
bartgol: Agreed. CIME probably doesn't even know what CF is, not should it.

bartgol:
The fail on weaver was due to an issue with the scripts detecting all 4 GPUs (they only saw 2, for some reason). I checked the node they ran on, to make sure no GPU was temporarily removed/disabled, and it looks fine. Retesting.

Also, removing automerge, since it seems the conversation on CF version is not over, and don't want AT to prematurely merge.

czender:
@golaz and @bartgol `ncclimo` will work fine with `conventions=CF-1.0`. Thanks for adding it. No reason to increment the CF version or delay this PR on account of `ncclimo` compatibility.

components/eamxx/src/share/io/scream_output_manager.cpp
AaronDonahue: >One comment on a one-line PR:

It's okay, every revision arbitrarily buffs my git stats with an extra commit ;-)

golaz (approved):

czender (changes requested):
The official name of the conventions attribute is `Conventions` (uppercase, as in EAM) not `conventions` (lowercase). Please change. I'm also looking into the value, and whether it should be bumped to > 1.0. Will provide more feedback soon.

AaronDonahue:
> The official name of the conventions attribute is `Conventions` (uppercase, as in EAM) not `conventions` (lowercase). Please change. I'm also looking into the value, and whether it should be bumped to > 1.0. Will provide more feedback soon.

@czender great! Thanks for looking into that.  I will make that change once you report back on what the value should be.

czender:
The file from the April 12th repo that @PeterCaldwell pointed me to, `/pscratch/sd/n/ndk/e3sm_scratch/pm-cpu/se63-apr12/f30cpu.F2010-SCREAMv1.ne30pg2_ne30pg2.se63-apr12.gnu.1m.n008a128x1.pk8.wr/run/f30cpu.F2010-SCREAMv1.ne30pg2_ne30pg2.se63-apr12.gnu.1m.n008a128x1.pk8.wr.scream.hi.INSTANT.nsteps_x480.0001-01-01-00000.nc` passes the CEDA CF-compliance checker for CF-1.8 with no warnings or errors. Presuming the rest of the EAMxx files have similar headers/metadata, then I recommend to change the global attribute to `Conventions=CF-1.8`. There's still a lot of useful CF attributes that could be added to EAMxx (and EAM, and ELM). Since CF does not mandate most attributes (they are almost all opt-in), the checkers really only tell us whether there are outright violations of CF, and there are none in the current repo.

PeterCaldwell:
> passes the CEDA CF-compliance checker for CF-1.8 with no warnings or errors

Oh that's excellent!

czender (approved):

PR #2286, opened 2023-04-13
fix pointer to device vs. host memory in surface coupling run_impl

AaronDonahue:
Fixes a bug introduced in PR #2272

### DIFF
--- a/components/eamxx/src/control/atmosphere_surface_coupling_exporter.cpp
+++ b/components/eamxx/src/control/atmosphere_surface_coupling_exporter.cpp
@@ -206,8 +206,8 @@ void SurfaceCouplingExporter::initialize_impl (const RunType /* run_type */)
   using vos_type = std::vector<std::string>;
   using vor_type = std::vector<Real>;
   m_export_source     = view_1d<DefaultDevice,ExportType>("",m_num_scream_exports);
-  auto export_source_h = Kokkos::create_mirror_view(m_export_source);
-  Kokkos::deep_copy(export_source_h,FROM_MODEL);  // The default is that all export variables will be derived from the EAMxx state.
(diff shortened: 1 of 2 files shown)


components/eamxx/src/control/atmosphere_surface_coupling_exporter.cpp
```diff
@@ -436,13 +437,13 @@ void SurfaceCouplingExporter::compute_eamxx_exports(const double dt, const bool
...
-      if (m_export_source(idx_Faxa_rainl)==FROM_MODEL) { Faxa_rainl(i) = precip_liq_surf_mass(i)/dt*(1000.0/PC::RHO_H2O); }
-      if (m_export_source(idx_Faxa_snowl)==FROM_MODEL) { Faxa_snowl(i) = precip_ice_surf_mass(i)/dt*(1000.0/PC::RHO_H2O); }
+      if (export_source(idx_Faxa_rainl)==FROM_MODEL) { Faxa_rainl(i) = precip_liq_surf_mass(i)/dt*(1000.0/PC::RHO_H2O); }
+      if (export_source(idx_Faxa_snowl)==FROM_MODEL) { Faxa_snowl(i) = precip_ice_surf_mass(i)/dt*(1000.0/PC::RHO_H2O); }
     }
   });
   // Variables that are already surface vars in the ATM can just be copied directly.
-  auto export_source_h = Kokkos::create_mirror_view(m_export_source);
-  Kokkos::deep_copy(export_source_h,m_export_source);
+  auto export_source_h = Kokkos::create_mirror_view(export_source);
```
bartgol, line 445: You might as well save a host copy of export source in the class, so you don't need to allocate/copy every timestep..

PR #2290, opened 2023-04-17
Updating submodules and continuing work on MAM4 AtmosphereProcess

jeff-cohere:
Things are coming along. We had to make some changes to our data structures to support tracer subviews, but I think we've sorted that out now. This AtmosphereProcess still doesn't do much, but @pbosler and I are going to be working on it in a more focused way moving forward.

Because we're still playing around and getting things to work, and because our work still requires SCREAM to be built with C++17, this functionality is still protected by the `SCREAM_ENABLE_MAM` option, which is `OFF` by default.

@AaronDonahue and @bartgol , I've added you as reviewers to keep this on your respective radars. Let me know if you want me to add anyone else to these reviews, or to only poke you when we've got something "interesting." :-)

### DIFF
--- a/components/eamxx/src/physics/mam/eamxx_mam_microphysics.cpp
+++ b/components/eamxx/src/physics/mam/eamxx_mam_microphysics.cpp
@@ -105,6 +105,64 @@ set_computed_group_impl(const FieldGroup& group) {
     "Error! MAM4 requires at least " << num_aero_tracers << " aerosol tracers.");
 }
 
+size_t MAMMicrophysics::requested_buffer_size_in_bytes() const
+{
+  // number of Reals needed by local views in interface
+  const size_t iface_request = sizeof(Real) *
+    (Buffer::num_2d_mid * ncol_ * nlev_ +
+     Buffer::num_2d_iface * ncol_ * (nlev_+1));
+
+  // FIXME: Need to figure out whether we need this stuff
+  /*
+  // Number of Reals needed by the WorkspaceManager passed to shoc_main
+  const auto policy       = ekat::ExeSpaceUtils<KT::ExeSpace>::get_default_team_policy(m_num_cols, nlev_packs);
+  const int n_wind_slots  = ekat::npack<Spack>(2)*Spack::n;
+  const int n_trac_slots  = ekat::npack<Spack>(m_num_tracers+3)*Spack::n;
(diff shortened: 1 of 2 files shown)


components/eamxx/src/physics/mam/eamxx_mam_microphysics.cpp
```diff
@@ -60,10 +62,19 @@ void MAMMicrophysics::set_grids(const std::shared_ptr<const GridsManager> grids_
...
+  // aerosol tracers of interest: mass (q) and number (n) mixing ratios
+  add_field<Updated>("q_aitken_so4", scalar3d_layout_mid, q_unit, grid_name, "tracers");
+  add_field<Updated>("n_aitken_so4", scalar3d_layout_mid, n_unit, grid_name, "tracers");
+
+  // aerosol-related gases: mass mixing ratios
+  add_field<Updated>("q_soag", scalar3d_layout_mid, q_unit, grid_name, "tracers");
+  add_field<Updated>("q_h2so4", scalar3d_layout_mid, q_unit, grid_name, "tracers");
+
+  // Tracer group -- do we need this in addition to the tracers above?
+  add_group<Updated>("tracers", grid_name, 1, Bundling::Required);
```
bartgol, line 74: It doesn't look like you're using tracers... You could add the 4 fields above to a new group, and call it `aero_tracers`. I'd be interested in whether the AD can correctly bundle those 4. So far, we only have one group, the "tracers" group. This would be the first instance of a nested group of tracers being bundled _within_ the tracers group.

bartgol: You can, of course, still add these to the tracers group as well. You can include a field in as many groups as you want. So you can do
```
add_field<Updated>("q_aitken_so4", scalar3d_layout_mid, q_unit, grid_name, {"tracers","aero_tracers"});
...
add_group<Updated>("aero_tracers", grid_name, 1, Bundling::Required);
```

To be clear, the latter is really only important if a) you want to handle any aero tracer that may be declared by other atm procs, or b) you want them bundled (perhaps so you can loop over them and do some common ops). If you _expect_ that the only aero tracers are those declared in this class _and_ you don't need the aero tracers to be bundled, then you don't really need the group, and can simply work with the individual fields.

bartgol (approved):

pbosler (approved):

PR #2280, opened 2023-04-11
Jgfouca/upstream merge 2023 04 11

jgfouca:
No conflicts!

This PR brings in the Adios stuff that I think some of us were waiting for in scream.

PR #2291, opened 2023-04-17
Always write rhist files if checkpoint is enabled

bartgol:
This PR ensures that a `.rhist` file is _always_ written, even if the avg type is INSTANT, or if frequency is every step, or if the checkpoint step coincides with an output step. Previously, we were not generating a rhist file in these scenarios, since there's no history to track. However, this made history restarts a bit more complicated, as we had to make some assumptions depending on the output spec and timestamp. For instance, for an INSTANT output, we had to determine how many steps were passed since the last write based on the output freq and the timestamp nsteps attribute.

With this modifications, we always save _some_ data. If the checkpoint falls in one of the three scenarios listed above, we do not store any time record, and the file simply contains global metadata. This metadata can however be useful to correctly restart the history, without making any assumption. For instance, for the case listed above, the restart file will contain the time of the last write, as well as how many snapshots had already been taken since the last write, which is sufficient for the output stream to determine when the next output step will happen.

This will also help to resume filling an output file upon restart.

I manually verified that an INSTANT output generates nc files with 0 time records, with only metadata.


AaronDonahue (approved):
Looks good to me, I especially like the first step into calling PIOc directly instead of wrapping F90.  I saw there was a set of FAILED CIME based tests in the AT on mappy.  Once those are all sorted this should be good to go.

AaronDonahue (approved):

bartgol:
Same NL diffs as #2280 , but otherwise tests pass. Merging.

PR #2232, opened 2023-03-10
Add scream dockerfile 

elynnwu:
This PR adds a dockerfile for eamxx. It is built on top of CIME's docker container with additional packages for eamxx and google cloud platform sdk for interfacing with AI2's ML workflow. Full model run and eamxx unit tests were completed inside the container and should be added to CI in the future.

PR #2293, opened 2023-04-19
Rename docker to docker-scream to avoid conflicts

jgfouca:


elynnwu (approved):

elynnwu:
> @elynnwu , I am unsure if I caught all the places impacted by this rename. In the original PR (https://github.com/E3SM-Project/scream/pull/2232/files) I don't see where the scream machine is being selected.

There's currently no test that selects the docker machine, but we do use this setup for running simulation in the cloud. I will make sure to update anything downstream. This change looks fine.

PR #2294, opened 2023-04-19
Add a bit more info to what is printed under SimulationParams

jgfouca:
This output is the most useful info in atm.log so I decided to add a bit of extra info even though this info is not tied to the params object.


components/eamxx/src/dynamics/homme/atmosphere_dynamics.cpp
```diff
@@ -815,6 +823,13 @@ void HommeDynamics::init_homme_views () {
   msg << "   disable_diagnostics: " << (params.disable_diagnostics ? "yes" : "no") << "\n";
   msg << "   theta_hydrostatic_mode: " << (params.theta_hydrostatic_mode ? "yes" : "no") << "\n";
   msg << "   prescribed_wind: " << (params.prescribed_wind ? "yes" : "no") << "\n";
+  msg << "   ncols: " << ncols << "\n";
```
bartgol, line 826: These are physics grid parameters, and the policy-related ones refer to our default policy over columns and level packs. 

The CXX simulation params are meant to reflect the Hommexx configuration. I added this impl back when `SimulationParams::print` did not accept an ostream as input. That has since changed, and we should probably just call `SimulationParams::print(msg)`.

Either way, I think it should be made clear that the ones you added are *not* Homme parameters.

jgfouca:
@bartgol , does my latest commit make it better?

bartgol (approved):

PR #2296, opened 2023-04-20
Fix issue where some restart configurations were failing.

AaronDonahue:
This commit fixes an issue where restarts that involved history files with frequency and frequency units that didn't match the simulation frequency and units would go looking for a restart history file even if it was needed.  For example if we restart a monthly run with an output stream that took average output every 3 hours then no restart history file is needed but EAMxx would still go look for one in the rpointer.atm file and throw an error.

### DIFF
--- a/components/eamxx/src/share/io/scream_output_manager.cpp
+++ b/components/eamxx/src/share/io/scream_output_manager.cpp
@@ -162,7 +162,7 @@ setup (const ekat::Comm& io_comm, const ekat::ParameterList& params,
       // NOTE: if you change the output frequency when you restart, this could lead to wonky behavior
       m_output_control.nsamples_since_last_write = m_run_t0.get_num_steps() % m_output_control.frequency; 
 
-      if (has_restart_data) {
+      if (has_restart_data && m_output_control.nsamples_since_last_write>0) {
         using namespace scorpio;
         auto fn = find_filename_in_rpointer(hist_restart_casename,false,m_io_comm,m_run_t0);
         register_file(fn.c_str(),FileMode::Read);


AaronDonahue:
#2291 may already address this in part, but I made this quick bugfix just in case.  I think we tried the branch for #2291 to see if the error went away and it didn't.

PeterCaldwell:
Wow, esoteric error. Nice catch!

bartgol (commented):
I'm merging #2291 right now (only NL fails, brought in by #2280). I would retest master after the merge, to see if the problem persists. I'm guessing/hoping it won't, so this PR won't be needed anymore.

components/eamxx/src/share/io/scream_output_manager.cpp
```diff
@@ -162,7 +162,7 @@ setup (const ekat::Comm& io_comm, const ekat::ParameterList& params,
       // NOTE: if you change the output frequency when you restart, this could lead to wonky behavior
       m_output_control.nsamples_since_last_write = m_run_t0.get_num_steps() % m_output_control.frequency; 
 
-      if (has_restart_data) {
+      if (has_restart_data && m_output_control.nsamples_since_last_write>0) {
```
AaronDonahue, line 165: Actually, I think line 163 isn't correct either.  If frequency was 1 / month then `get_num_steps() % m_output_control.frequency==0` for all steps because frequency is 1, even for a run that restarts after 1 week.

bartgol: Yes, that's true. It is implicitly assumed that the number of samples is always the number of steps. Imho, this points to the need of distinguishing the sample size from the output frequency.

bartgol: Actually, so long as we use all timesteps when performing averages, the two are the same.

elynnwu:
I ran master and it created a `rhist` file for the output stream but the restart still does not work:
```
Error! Restart requested, but no restart file found in 'rpointer.atm'.
   restart case name: ./test-restart-3.F2010-SCREAMv1.ne4_ne4.ruby.scream
   restart file type: history restart
   rpointer content:
./test-restart-3.F2010-SCREAMv1.ne4_ne4.ruby.scream.r.INSTANT.nhours_x6.2010-01-01-21600.nc
output.scream.coarsened.rhist.AVERAGE.nhours_x6.2010-01-01-21600.nc
```

Think we still need this PR

bartgol:
Closing, since #2299 should address the root cause.

PR #2297, opened 2023-04-21
Use map to avoid issue with grid data corruption

tcclevenger:
In my testing on summit the grids stored in these sets have data that gets corrupted when `total_mpi_tasks>2` . Here is the error:
```
FAIL:
grid->is_unique()
Error! I/O only supports grids which are 'unique', meaning that the  map dof_gid->proc_id is well defined.
```
To try and summarize when this occurs, when adding output streams for geo data:
```
// Store io grids (PG2 and GLL) from m_output_streams to std::set of grids
std::set<std::shared_ptr<const AbstractGrid>> grids;
for (auto& it : m_output_streams) {
  it->get_io_grid()->is_unique() // Always TRUE
  grids.insert(it->get_io_grid());  // Store the ptr in a std::set
}
// Check if girds in grids set are still unique
for (auto grid : grids) {
  grid->is_unique(); // Almost always FALSE, triggered when setting up new geo output
}
```
Basically, the GIDs on the grids were getting corrupted in the form of switching between the 2 grids on some processors. So for example processor 2's GLL grid could have GIDs from the PG2 grid. Very odd, but structured.

Using `std::map` to hold the grid ptr fixed it. 

Thought on why this could be the case? @ambrad @jgfouca @bartgol. I'm a bit confused.

From @bartgol, we see this same fail in valgrid tests on mappy https://my.cdash.org/test/78411956. Currently running those tests  with this change, but not hopeful since I don't think that test uses the same code (need to verify).

### DIFF
--- a/components/eamxx/src/share/io/scream_output_manager.cpp
+++ b/components/eamxx/src/share/io/scream_output_manager.cpp
@@ -106,25 +106,25 @@ setup (const ekat::Comm& io_comm, const ekat::ParameterList& params,
   // For normal output, setup the geometry data streams, which we used to write the
   // geo data in the output file when we create it.
   if (m_output_file_specs.save_grid_data) {
-    std::set<std::shared_ptr<const AbstractGrid>> grids;
-    for (auto& it : m_output_streams) {
-      grids.insert(it->get_io_grid());
+    std::map<std::string, std::shared_ptr<const AbstractGrid>> grids;
+    for (const auto& it : m_output_streams) {
+      grids[it->get_io_grid()->name()] = it->get_io_grid();
     }
 
     // If 2+ grids are present, we mandate suffix on all geo_data fields,
     // to avoid clashes of names.
     bool use_suffix = grids.size()>1;
-    for (auto grid : grids) {
+    for (const auto& grid : grids) {
       std::vector<Field> fields;
-      for (const auto& fn : grid->get_geometry_data_names()) {
-        const auto& f = grid->get_geometry_data(fn);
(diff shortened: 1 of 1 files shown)


bartgol:
I have a vague recollection of some issue when using std::shared_ptr as key in ordered std containers. I do not recall the details (and quick google searches seem to suggest it's ok to use them), so this may be unrelated.

However, _IF_ this PR fixes the model_restart tests (which, mind, only fail with valgrind, and not consistently), there is a slight chance that they may fix some random behavior we are observing with some ne30 tests. It's a very small chance, but worth checking if @tcclevenger confirms valg tests pass after this.

ambrad:
The issue is that the iteration through the set has an order determined by the shared_ptr operator<, which in turn depends on the arbitrary address values of the object it holds. Thus, in the case of multiple grids, each rank may have a different iteration order through the set. Result: is_unique, which contains comm, is called for different grids on different ranks, mixing GIDs in the comm. The map fix you put in uniquifies the order of iteration through the container, so is_unique is called for the same grid on every rank.

bartgol:
Ah, yes, that's it! It makes a lot of sense now.

I should probably check that we do not do something like this anywhere else. This kind of error has the potential of generating random errors, and cannot be picked up by valgrind, since it's perfectly valid code.

tcclevenger:
Thanks @ambrad, definitely clear now.

@bartgol `model_baseline` passes for mappy `full_debug_valgrind`. I didn't see earlier that it is running with multiple grid, so it would touch that piece of code. So I expect this to be a complete fix.

PR #2299, opened 2023-04-21
Fix bug related to restart

bartgol:
@elynnwu helped me track down a couple of bugs related to history restart files. This PR addresses both of them.

Closes #2296


AaronDonahue (approved):
Thanks for finding and fixing the bug so quickly!

PR #2301, opened 2023-04-24
Convert physics_saturation test to a baseline test

jgfouca:
Change list:

1) Transform physics_saturation test to a baseline test
2) Encapsulate scream::compare, which was about to be duplicated a third time
3) There is no point in doing thread sweeping on physics_saturation since none of the physics functions do any kokkos operations


components/eamxx/src/physics/share/tests/physics_saturation_run_and_cmp.cpp
```diff
@@ -0,0 +1,320 @@
...
+    //Convert Scalar inputs to Spacks because that's what polysvp1 and qv_sat expect as inputs.
+    //--------------------------------------
+    const Spack temps(temperature);
+    const Spack pres(pressure);
+
+    //Get values from polysvp1 and qv_sat (qv_sat calls polysvp1 here) to test against "expected" values
+    //--------------------------------------
+    sat_ice_fp  = physics::polysvp1(temps, true, Smask(true))[0];
+    sat_liq_fp  = physics::polysvp1(temps, false, Smask(true))[0];
+    //last argument "0" of qv_sat function below forces qv_sat to call "polysvp1"
```
PeterCaldwell, line 52: I think this comment about last arg being 0 and the comment ~8 lines down about last arg of 1 aren't true any more because you're passing physics::MurphyKoop and physics_Polysvp1 now? Delete them? Or is physics::MurphyKoop==1?

PeterCaldwell:
PR looks great to me. So much cleaner than my old version! Thanks. I'm not approving just because I didn't fully understand the make baseline/compare against baseline code so want someone with more savvy to look at that.

components/eamxx/src/physics/share/tests/physics_saturation_run_and_cmp.cpp
jgfouca: Yes, looks like it's an enum now instead of a bool. The comments aren't needed since the enum name makes it clear what's going on. I will remove them.

bartgol (approved):

jgfouca:
Only fails are due to missing baselines for physics_saturation_run_and_cmp, which is expected. Merging.

PR #2302, opened 2023-04-24
Add ocnfrac and landfrac to SC importer

tcclevenger:
Add ocnfrac and landfrac to SC importer. Added to default output yaml file.


tcclevenger:
@AaronDonahue I'm guessing CIME tests will fail since I added to output.yaml. Should I just remove and whoever is curious about these fields can add on their own?

PeterCaldwell:
I'd vote for not having these be default outputs - the landfrac in particular is constant in time, which greatly offends my sense of data compression.

PeterCaldwell (approved):

PR #2306, opened 2023-04-26
test-all-scream: Add support for compute-sanitizer test types

### LINKED ISSUE #2125: Cuda Compute Sanitizer
Cuda 11 has a new tool set that looks pretty nifty: https://docs.nvidia.com/compute-sanitizer/ComputeSanitizer/index.html

This task is to use the Memcheck, Racecheck, Initcheck, and Synccheck commands on a one-node v1 test to see if they find anything interesting.

jgfouca:
Also, refactor and clean up some things in test_all_scream and Cmake:
1) Remove --mem-check flag. Replace with specific test types
2) SCREAM_TEST_SIZE=SHORT is now one of the determining factors whether BFB layer is on. If test size is short, bfb layer will be off so things can run faster.
3) Tests types can now override the default test size
4) Can now pass --test-size to test-all-scream to force a certain test size
5) "mem" test now means chose either valgrind or cuda-memcheck depending on machine
6) Clean up lots of one-off special cases due to --mem-check

Fixes #2125


ambrad (approved):

components/eamxx/scripts/test_all_scream.py
```diff
@@ -129,34 +141,126 @@ def __init__(self):
...
+            [("CMAKE_BUILD_TYPE", "Debug"),
+             ("EKAT_ENABLE_COMPUTE_SANITIZER", "True"),
+             ("EKAT_COMPUTE_SANITIZER_OPTIONS", "--tool initcheck")],
+            uses_baselines=False,
+            on_by_default=False,
+            default_test_len="short"
+        )
+
+###############################################################################
+class CSS(TestProperty):
```
bartgol, line 235: I'm guessing that compute-sanitizer does not allow 2+ tools to run at the same time, right?

bartgol (approved):

components/eamxx/scripts/jenkins/jenkins_common_impl.sh
```diff
@@ -108,7 +108,7 @@ if [ $skip_testing -eq 0 ]; then
 
       # Add a memcheck test for mappy/weaver for nightlies
       if [[ "$SCREAM_MACHINE" == "mappy" || "$SCREAM_MACHINE" == "weaver" ]]; then
-        ./scripts/gather-all-data "./scripts/test-all-scream -t dbg --mem-check ${TAS_ARGS} -c SCREAM_TEST_SIZE=SHORT" -l -m $SCREAM_MACHINE
+        ./scripts/gather-all-data "./scripts/test-all-scream -t mem ${TAS_ARGS}" -l -m $SCREAM_MACHINE
```
bartgol, line 111: Should we also add builds on weaver for CMS, CMR, CMI?

components/eamxx/scripts/test_all_scream.py
jgfouca: Yep, I specifically checked for that.

components/eamxx/scripts/jenkins/jenkins_common_impl.sh
jgfouca: Good question. I was wondering that too.

bartgol: It looks like our current memcheck build on weaver takes 45-60min. Assuming similar weight for the other 3, and the fact that weaver is not very used at night (I think), we may be able to afford it.

bartgol: Also, we may launch the 4 compute-sanitizer builds in ||, since there are 4 gpus on each weaver node...

jgfouca: Nice, so we can do 4 for the price of 1? Seems like an obvious win although it might have us chasing down false positives all the time.

bartgol: Well, not exactly. When running a single build, we can use 4 gpus as well, so running 4 builds we make each of them slower. The build phase will however be slightly faster (IIRC, most of the build time on weaver is spent on a couple of files, shoc/p3_functions_f90.cpp). I guess we'd have to try it, to see how expensive they really are...

bartgol: Regarding false positives, that might be true, but what's the point of having them if we don't run them? I would say "let's activate them", and if we see that they hit too many false positives, we can remove them from our nightlies (leaving them as an option, to debug stuff).

PR #2307, opened 2023-04-26
Remove possibility to remap fields on the fly when using AtmosphereInput

bartgol:
This feature was never used, seems unlikely to be needed (one can always remap after calling the input class), and was going to complicate a bit things for work that is coming up in IO.


tcclevenger (approved):
Looks good

PR #2309, opened 2023-04-27
A few mods related to IO

bartgol:
- Make the `e2str(FieldTag)` utility return the same strings that `get_nc_tag_name` returned, and remove the latter.
- Allow to query a grid for the string name of a tag. Defaults to `e2str(FieldTag)`, but a grid can store 'special' names for some tags. This helps when checking number of GLL columns in topo file when PG2 is not enabled.
- Add convenience constructor for AtmosphereInput accepting a list of Fields. Use this in grids managers as well as to read topo file.


tcclevenger (approved):
Looks good. Definitely fits with what I'm doing for GLL output.

tcclevenger (approved):

tcclevenger (approved):

PR #2310, opened 2023-04-27
test-all-scream: Improve valgrind support 

jgfouca:
Use persistent valgrind suppression files if they exist.

### DIFF
--- a/components/eamxx/scripts/test_all_scream.py
+++ b/components/eamxx/scripts/test_all_scream.py
@@ -156,7 +156,7 @@ class COV(TestProperty):
 class VALG(TestProperty):
 ###############################################################################
 
-    def __init__(self, _):
+    def __init__(self, tas):
         TestProperty.__init__(
             self,
             "valgrind",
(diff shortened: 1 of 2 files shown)


components/eamxx/scripts/test_all_scream.py
```diff
@@ -166,6 +166,11 @@ def __init__(self, _):
             on_by_default=False,
             default_test_len="short"
         )
+        if tas is not None:
```
bartgol, line 169: Rather than storing this in the repo, why don't we store it in the $baseline_dir folder for this machine? this way we can manually go and edit the file if need be, without the need for a PR?

We could store the supp file location inside our `machine_specs.py`, so that if someone runs on their personal machine (i do on my workstation/laptop quite often), they can have their own copy (stored in `~/.cime/scream_mach_specs.py`).

jgfouca: @bartgol interesting idea. There's some advantages to having the supp file be version controlled but it is probably better to have the flexibility of keeping it outside the repo.

bartgol (approved):
I removed automerge, so you have time to decide whether to change where we store the valg file (repo or baseline dir).

jgfouca:
@bartgol , I think it may be better to have it in the repo. It means we'll always know what suppression file goes with which version of the code and it's always preferable to have data under version control (so you can revert, examine history etc).

PR #2312, opened 2023-04-28
Crusher: Use GPU buffers in MPI.

ambrad:
Now that Lustre is active on Crusher, try GPU buffers again.

Temporarily disable ADIOS because there appears to be a conflict between the current ADIOS libs and module craype-accel-amd-gfx90a that manifests as a link error in the SCORPIO build phase.


ambrad:
Tested on Crusher.

PR #2313, opened 2023-04-28
Crusher: Work around tcmalloc issue.

ambrad:
cee/15.0.0 with GPU MPI buffers can crash in a system lib. Work around this by using cee/14.0.3.

### DIFF
--- a/cime_config/machines/config_machines.xml
+++ b/cime_config/machines/config_machines.xml
@@ -857,6 +857,8 @@
         <!-- See SCREAM issue #2080. -->
         <command name="load">craype-accel-amd-gfx90a</command>
         <command name="load">rocm/5.1.0</command>
+        <!-- Work around tcmalloc crash. -->
+        <command name="load">cce/14.0.3</command>
       </modules>
       <modules>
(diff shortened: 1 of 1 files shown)


ndkeen:
Not sure if this same thing, but I'm using cce15 and was using this:

```
#3: /home/jenkins/crayftn/craylibs/google-perftools/src/tcmalloc.cc:647] Attempt to free invalid pointer: 0xa15d810                                                                                                                         
#This is a signature for a known problem with tcmalloc, that the CCE Fortran compiler uses by default for the releases currently on Tioga.                                                                                                  
#They next craype release should make tcmalloc non-default, I believe.                                                                                                                                                                      
#In the meantime, could you rebuild adding "-hsystem_alloc" to the fortran compiler flags? (and link flags!)                                                                                                                                
string(APPEND FFLAGS " -hsystem_alloc")
string(APPEND LDFLAGS " -hsystem_alloc")
```

ambrad:
Yes, that works too, but I figured the one-line 14.0.3 change would be easier to back out when the next version solves this issue.

PR #2315, opened 2023-05-01
Small fix to VerticalRemapper

bartgol:
The class was passing a device pointer to the scorpio routines. I'm wondering how this did not cause any crash so far. @AaronDonahue is the vertical remapper used at all in our common runs?


AaronDonahue (approved):
Thanks for the catch Luca.

AaronDonahue:
@bartgol , to answer your question.  We do use the vertical remapper for our unit tests.  So I am also surprised that the unit test didn't flag this fail when run on `weaver`.

PR #2317, opened 2023-05-03
Add compute sanitizer tests to weaver nightlies

jgfouca:


components/eamxx/scripts/jenkins/jenkins_common_impl.sh
```diff
@@ -106,14 +106,23 @@ if [ $skip_testing -eq 0 ]; then
...
+      if [[ "$SCREAM_MACHINE" == "mappy" ]]; then
         ./scripts/gather-all-data "./scripts/test-all-scream -t mem ${TAS_ARGS}" -l -m $SCREAM_MACHINE
         if [[ $? != 0 ]]; then
           fails=$fails+1;
           memcheck_fail=1
         fi
       fi
+
+      if [[ "$SCREAM_MACHINE" == "weaver" ]]; then
+        ./scripts/gather-all-data "./scripts/test-all-scream -t cmc -t csr -t csi -t css ${TAS_ARGS}" -l -m $SCREAM_MACHINE
```
bartgol, line 119: Should we run `-csm` instead of `-cmc`? If that works for a few days, then we can also get rid of the CMC build type, which would be obsolete now that we can rely on compute-sanitizer.

jgfouca: I was thinking keep the cmc build in order to keep that row on cdash. We could either have csm tests report themselves as cuda-memcheck, even though it's really compute sanitizer, or just start a new cdash entry and lose the history of the old build.

bartgol: Ok, then let's keep cmc for now. If/when we are confident csr/csi/css are working fine, we can decide what to do (imho, start new cdash, and prune cmc entry).

bartgol (approved):

PR #2321, opened 2023-05-05
EAMxx: State hasher to track BFBness in production simulations.

ambrad:
Default is off (BfbHash = 0). A number N > 0 means that a hash is printed every N physics time steps.

The hash line appears in e3sm.log for the root rank and looks like this:
```
0: bfbhash>             14 81f1508e28e52796 (Hommexx)
```
Here, "14" is the physics time step. The hasher is called in homme_dynamics.cpp at the start of the run.

A testmod used in the RRM test exercises the hasher. In the future I might add a python POSTRUN_SCRIPT to check the hash values, but for now I don't.

Two future PRs will add an internal diagnostics capability to the EAMxx AD and Hommexx for debugging purposes.

### DIFF
--- a/components/eamxx/src/dynamics/homme/atmosphere_dynamics.cpp
+++ b/components/eamxx/src/dynamics/homme/atmosphere_dynamics.cpp
@@ -65,7 +65,9 @@ HommeDynamics::HommeDynamics (const ekat::Comm& comm, const ekat::ParameterList&
   const char* logname = m_atm_logger->get_logfile_name().c_str();
   set_homme_log_file_name_f90 (&logname);
 
-  m_bfb_hash_nstep = std::max(0, params.get<int>("BfbHash"));
+  m_bfb_hash_nstep = 0;
+  if (params.isParameter("BfbHash"))
+    m_bfb_hash_nstep = std::max(0, params.get<int>("BfbHash"));
 }
 
 HommeDynamics::~HommeDynamics ()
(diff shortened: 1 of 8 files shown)


ambrad:
Tested with BfbHash = 0, 1, and 6 on Crusher and Ascent.

components/eamxx/src/dynamics/homme/atmosphere_dynamics.cpp
```diff
@@ -64,6 +64,10 @@ HommeDynamics::HommeDynamics (const ekat::Comm& comm, const ekat::ParameterList&
   // Set the log filename in the F90 interface
   const char* logname = m_atm_logger->get_logfile_name().c_str();
   set_homme_log_file_name_f90 (&logname);
+
```
AaronDonahue, line 67: I think with `params.get<int>` you can pass a default value and if the key isn't present it takes the default.  So these lines could be reduced to
```
m_bfb_has_nstep = params.get<int>("BfbHash",0);
```

Of course, if someone passes a negative number we won't catch that, so maybe add a `EKAT_ERROR_MSG(m_bfb_has_nstep>=0,..);`

ambrad: The `get` that takes a default is non-const and so can't be called on a `const PL&`. I suspect that is a typo in the PL API.

bartgol: I suppose we could allow the default on non-const PL, by simply returning the input (without setting it in the PL).

bartgol (approved):

components/eamxx/src/dynamics/homme/atmosphere_dynamics.cpp
ambrad: Sure. I suggest rather than debating the nuances of PL behavior here, we accept these three lines of code as is and merge the PR.

PR #2322, opened 2023-05-05
Fix issues with GLL output for PG2 CIME cases

tcclevenger:
- Update output manager to deal with conflicting `ncol` lengths (PG2 and GLL).
- Add homme states to internal fields so they can be int output
- Add GLL output of homme states to test/default output

### DIFF
--- a/components/eamxx/data/scream_default_output.yaml
+++ b/components/eamxx/data/scream_default_output.yaml
@@ -59,6 +59,13 @@ Fields:
       - surf_sens_flux
       # Diagnostics
       - PotentialTemperature
+  # GLL output for homme states.
+  Dynamics:
+    Field Names:
+      - ps_dyn
+      - dp3d_dyn
+      - omega_dyn
+    IO Grid Name: Physics GLL
 output_control:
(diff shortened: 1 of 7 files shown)


bartgol:
Yes, this will cause some diffs for v1 tests. The standalone fails are another issue though.

tcclevenger:
@bartgol All fails as expects (CIME tests have 0 DIFFS, just additional fields in output)

bartgol (approved):

AaronDonahue (approved):

tcclevenger:
I'm currently attempting to plot some field data to verify GLL output matches what we expect. I'll update when we can force merge (also will let the AT run again).

tcclevenger:
@AaronDonahue Verified `_dyn` output is as expected for both np4 and pg2 case. Blake passes, mappy fails with expected diffs, weaver passed before the AT issues (not changes to the code since). I'm ok to merge this.

bartgol:
Looks like we need to fix conflicts, so we may as well get another AT round when you push.

AaronDonahue:
@tcclevenger , Looks like the fails are all DIFFs on mappy.  Are these expected.  Can this PR be merged?

Issue #2225, opened 2023-03-09
change `atmosphere_<name>` to `eamxx_<name>`


PeterCaldwell:
For each process, the interface files connecting the AD and FM to the actual parameterization have been named `atmosphere_<name>.Xpp`. This is confusing and we decided to change these file names to `eamxx_<name>.Xpp` instead.

bartgol:
Also, change <generic_name> to <pkg_name>, right? So e.g., `atmosphere_microphysics.hpp`->`eamxx_p3.hpp`.

PeterCaldwell:
Yeah, Luca's correct. ^^^^

AaronDonahue:
When addressing this we should look at #1322 to make sure it isn't possible to fix both with the same PR.

Issue #1322, opened 2021-11-18
Improve register_physics.hpp logic


bartgol:
Right now, the header pulls in all the physics packages. So far, except for unit testing, it's fine, since we don't have multiple choices for each parametrization. But if/when the SCREAM AD becomes the main atm driver for e3sm, this might make `register_physics.hpp` unusable.

**A possible approach**

First, rename all files avoiding the generic prefix `atmosphere`. E.g., `atmosphere_macrophysics.*pp` -> `shoc_atm_process.*pp`. Then, make the inclusion and registration of processes guarded by ifdefs:
```
#ifdef SCREAM_HAS_SHOC
#include "physics/shoc/shoc_atm_process.hpp
#endif
#ifdef SCREAM_HAS_CLUBB
#include "physics/clubb/clubb_atm_process.hpp
#endif
#ifdef SCREAM_HAS_P3
#include "physics/p3/p3_atm_process.hpp
#endif
...
void register_physics () {
  auto& proc_factory = AtmosphereProcessFactory::instance();
#ifdef SCREAM_HAS_SHOC
  proc_factory.register_product("shoc",&create_atmosphere_process<SHOCMacrophysics>);
#endif
#ifdef SCREAM_HAS_CLUBB
  proc_factory.register_product("clubb",&create_atmosphere_process<CLUBBMacrophysics>);
#endif
...
}
```
Now, in order to be able to call `register_physics()` from the test setup, we need to forward the macros we want, so that they are defined when `register_physics.hpp` is parsed. The naive approach would be to add the compile definitions when building the tests. A much more powerful (and less intrusive) method is to have the libraries define the macro. So

Inside p3's CMakeLists.txt:
```
add_library(p3 ...)
target_compile_definitions (p3 PUBLIC SCREAM_HAS_P3)
```
Inside the test (or lib) CMakeLists.txt:
```
# use scream's macro to create test
CreateUnitTest (my_test ... LIBS "p3;share;...")

# create a library for mct
add_library (atm ...)
target_link_libraries (atm PUBLIC p3)
```
Done. Nothing else is required, since `target_link_libraries(<target> PUBLIC p3)` already propagates the PUBLIC (and INTERFACE) compile definition (and more) of the `p3` library target to the exec/lib currently built. So when building the `atm` library, `SCREAM_HAS_P3` _will_ be defined, so that `register_physics.hpp` is registering P3 in the factory.

[...]

PR #2324, opened 2023-05-09
Rename interface files and improve register_physics logic

tcclevenger:
Closes https://github.com/E3SM-Project/scream/issues/2225 and https://github.com/E3SM-Project/scream/issues/1322.

Seemed like some low hanging fruit. Let me know if there is any reason to punt this down the road, or if a more sophisticated approach with CMakeLists is wanted (I followed Luca's suggestion in https://github.com/E3SM-Project/scream/issues/1322, but there he mentions things I don't implement, namely generic macros like `${SCREAM_MACROPHYSICS_PKG}`).

Questions:
1. I renamed MAM to follow this convention (removing "microphysics" from `eamxx_mam_micrphysics.xpp`). Since it is the most recent named file, we may want to instead follow its lead of including the generic category. E.g., `eamxx_shoc_macrophysics.xpp`. I think I like not including the generic name, but open to discussion.
2. Should we stop here? Or should `atmosphere_process` -> `eamxx_process`? And likewise with all `atmosphere_` naming.

I'm tagging those I remember in the original discussion, feel free to add others if I forgot any. Or remove yourself if you aren't interested in the discussion.


bartgol (dismissed):
Looks good. I think we can merge this as is, if we decide that every `atmosphere_blah` has to become `eamxx_blah`, we can always do that later.

PeterCaldwell:
>I renamed MAM to follow this convention (removing "microphysics" from eamxx_mam_microphysics.xpp).

I think @jeff-cohere will agree with me that `mam_microphysics` is as special thing in the sense that mam (the aerosol module) has lots of individual processes. Aerosol microphysics is one of them, so we need to keep the microphysics in this particular name (but not in shoc, p3, or anything else).

jeff-cohere:
I do indeed agree with @PeterCaldwell on this. Aerosols are a cross-cutting phenomenon, it turns out, so we do expect to have a handful of MAM-related processes.

jeff-cohere (dismissed):
This looks good to me! Thanks for reverting the MAM name changes, @tcclevenger .

AaronDonahue (approved):
@tcclevenger thanks for doing this.  It is great for us to move towards the EAMxx convention got general files.

I only have one comment that should not hold up approval.  For the `eamxx_PROC.hpp/cpp` files, would it be more descriptive to make these `eamxx_PROC_interface.hpp/cpp`?  This is more of a general question for the group - @PeterCaldwell , @bartgol ?

bartgol:
> For the `eamxx_PROC.hpp/cpp` files, would it be more descriptive to make these `eamxx_PROC_interface.hpp/cpp`? This is more of a general question for the group - @PeterCaldwell , @bartgol ?

I don't have a strong optinion one way or another.

PeterCaldwell:
> I only have one comment that should not hold up approval. For the `eamxx_PROC.hpp/cpp` files, would it be more descriptive to make these `eamxx_PROC_interface.hpp/cpp`? This is more of a general question for the group - @PeterCaldwell , @bartgol ?

I'm a big fan of adding `_interface` to these files because dividing our processes into "interface" and "implementation" is a fundamental aspect of our design and I think it would be good to be as clear about that distinction as possible. I remember last time this issue came up, I was outvoted on this because (as Jeff says) "interface" is such an overloaded term that most people will see it as meaningless and redundant. I'm not sure that's true for our atm scientist user base though. If nobody else cares, I have a strong preference for including `_interface`. But I'm ok with being outvoted (again ;-) )

tcclevenger:
I don't have much of a preference about using interface. 

There is a place where interface is a bit overloaded in `src/physics/rrtmgp`. If we changed to using `_interface.xpp` we would have `eamxx_rrtmgp_interface.xpp` and `scream_rrtmgp_interface.xpp`. I looks like the functions in `scream_rrtmgp_interface.xpp` interface for rrtmgp<->eamxx, and `eamxx_rrtmgp_interface.xpp` interface between AD<->`scream_rrtmgp_interface.xpp`. Maybe `scream_rrtmgp_interface` needs to be renamed anyways, or moved into `eamxx_rrtmgp(_interface).xpp`.

AaronDonahue:
> To be clear, I don't have strong feelings about the use of "interface" here. I just want to urge caution about using buzzwords (see also "framework") that tend to be used in tandem with wild hand waving.

No buzzwords? I was going to suggest adding "synergy" next... alas. :-)

But in all seriousness.  I do think this is a case where `interface` is actually appropriately used.  We designed the AD to have an interface between EAMxx and a given atmosphere process.  Maybe there is a better word that can be used which is less ambiguous.  'eamxx_ap_PROC.xpp`, `eamxx_PROC_bridge.xpp`?

Either way, I brought the question up with a minds eye to future developers and our goal to be less ambiguous than EAM in our naming conventions.  Perhaps our enforcement of consistency in the naming of files makes the point moot since developers will learn the pattern `eamxx_PROC.xpp` means "interface".

tcclevenger:
As per discussion in SCREAMv1 meeting: `eamxx_${proc}_ad_interface.xpp` (or maybe `_driver_` to avoid asking what "ad" is? I'll make some decision..)

bartgol:
> As per discussion in SCREAMv1 meeting: `eamxx_${proc}_ad_interface.xpp` (or maybe `_driver_` to avoid asking what "ad" is? I'll make some decision..)

Imho, the `ad`/`driver` part is redundant. I interpret `eamxx_blah_interface` as "the interface between eamxx and blah".

tcclevenger:
@bartgol My issue is that there is a naming convention already for `interface`: `scream_${proc}_interface.xpp`. I think it's present for all processes whose code exists outside the eamxx repo (homme, rrtmgp), and it provides the interface to make a process' external code callable from eamxx. This should probably be renamed `eamxx_$proc_interface.xpp`?

Whereas `eamxx_${proc}.xpp` (or `eamxx_$proc_ad_interface.xpp`) is the interface between the process and the AD (the atm process). So maybe we leave as is? Or `eamxx_$proc_atm_process.xpp`?

PeterCaldwell:
I find this discussion confusing, so maybe this is a dumb idea... If ${proc}_interface conflicts with interface:hommexx stuff, maybe ${proc}_driver.xpp is the way to go? I also prefer not including ad in the name because folks might not know what it means and it feels a bit redundant (especially when one unrolls the acronym to atmos driver driver :-) ).

bartgol:
This is becoming more complicated than it should be... So long for the S in SCREAM! :)

At this point I don't feel strongly about any abbreviation. I don't really care about distinguishing AD-AtmProc interfaces from raw interfaces (i.e., I'm ok with `eamxx_scorpio_interface.hpp` and `eamxx_homme_interface.hpp` sharing a similar name despite being different kind of interfaces). For homme/rrtmgp, we do have the issue of an existing interface file/lib though, so to accommodate those, we either add another prefix/postfix to atm procs interfaces (I think `eamxx_$pkg_atm_proc_interface.hpp` or `eamxx_$pkg_process_interface.hpp` may be the cleanest) , or we change the name of current interfaces (perhaps `eamxx_$pkg_interface_details.hpp`, or `eamxx_$pkg_helpers.hpp`).

tcclevenger:
Went with `eamxx_$proc_process_interface`. Setting to WIP until weaver is up.

jeff-cohere (approved):

PR #2326, opened 2023-05-11
Change name of FieldAt(Pressure)Level diagnostic fields

bartgol:
List of mods:

* Use '_at_' instead of '@', to for the sake of current postprocessing scripts
* Move logic to determine level/pressure_level inside the diagnostic
* Change `tom`/`bot` suffix strings to `model_top`/`model_bot`
* If user specifies `model_top`/`model_bot`, keep that in the output field name (i.e., do not convert into `levN`)
* Minor clean up of diagnostic code.

This was requested by e3sm postprocessing scripts developers, since the symbol `@` was causing some trouble to the scripts.


AaronDonahue (changes requested):
Looks good @bartgol , thanks for such a rapid response.  The only requested change is to fix the output.yaml files for any tests that currently use the `@` syntax.  A quick look and I found at least one place: https://github.com/E3SM-Project/scream/blob/master/components/eamxx/tests/coupled/dynamics_physics/homme_shoc_cld_spa_p3_rrtmgp/homme_shoc_cld_spa_p3_rrtmgp_diagnostics.yaml

There may be more.

bartgol:
Ah, good catch. I checked the v1 ones, but not the standalone tests. Will fix it/them.

AaronDonahue (approved):

AaronDonahue:
@czender - Just a heads up on the changing convention for output names in EAMxx

golaz:
@bartgol : thanks for working on this change. This will greatly help downstream processing of EAMxx output files.

AaronDonahue (approved):

AaronDonahue (approved):

AaronDonahue (approved):

PR #2327, opened 2023-05-12
Allow to resume filling existing output files upon restart

bartgol:
This PR has the main goal to allow to continue writing in the same output file when we perform a restart (provided that there is still room in the file, and that the user does not explicitly ask _not_ to resume).

Although the task per se was not too complicated, it required some changes which cascaded into more and more changes. Here's a summary of the main ones.

- When we open an existing file for input (or append), we check that the variables/dimensions in the file have the expected specs (length, units, data type,...).
- The above, forced me to add ne2 IC and topo files. I still have to upload them on the server (though they are on mappy), since the ANL server was down.
- Refactored a bit how we pass specs for the mesh-free GM. This was in turn needed b/c of how the AD looks for the topography fields (and their dims) from the topo file. So far, things "worked", but we were really reading a ne4pg2 phis field into the ne2gll one (but our IO interface was ok with that, since we never checked dimensions).
- Unrelated to the task: update EKAT, to take advantage of E3SM-Project/EKAT#292

Note: I still have to fully test this with v1 cases. I tried an ERS test, and it passed, but the initial run (up to the rhist write) generates no output, so it might be a corner case. I need to test an INSTANT output, and an AVERAGE output that _does_ write some snapshots before getting to the restart point. Standalone tests suggest that INSTANT should be working already (model_restart tests are already resuming the fill).


bartgol:
A manual restarted case with AVERAGE or INSTANT output on mappy seems to suggest the resume option works as expected.

bartgol:
@AaronDonahue no, they are not expected.

bartgol:
One of the failing tests passes when I run manually. I'm gonna trigger another round of testing, and see what happens.

AaronDonahue (approved):

components/eamxx/src/share/io/tests/output_restart.cpp
```diff
@@ -39,125 +43,110 @@ void time_advance (const FieldManager& fm,
...
+        std::cout << std::left << std::setw(line_len) << std::setfill('.') << s;
+      }
+    }
+  };
+  // Run test for different avg type choices
+  for (const std::string& avg_type : {"INSTANT","AVERAGE"}) {
+    print("   -> Averaging type: " + avg_type + " ", 40);
+    output_params.set<std::string>("Averaging Type",avg_type);
+
+    // 1. Run for full 20 days, no restarts needed
```
AaronDonahue, line 121: is this "days" or "steps" that you run for?

AaronDonahue (approved):
Nice big PR.  It looks good, I had one comment about a comment in the code, but that shouldn't hold up merging.

AaronDonahue:
@bartgol , It looks like this PR hit DIFFs in mappy.  Are these the same ones you tested offline and they passed?

bartgol:
I am a bit puzzled. I think they are. I will re-test again, and see what happens.

AaronDonahue:
Is it possible we fail baselines for some tests because in some cases we are resuming to fill files, which we were not in the past.  Probably could check the cprnc log to see, I think cprnc will throw an error if the time dimensions don't match.

bartgol:
> Is it possible we fail baselines for some tests because in some cases we are resuming to fill files, which we were not in the past. Probably could check the cprnc log to see, I think cprnc will throw an error if the time dimensions don't match.

No, that's not it. Our ERS/ERP tests only do AVERAGE, and they average over the whole run, so we only have 1 snap in the file. These are legit diffs, as the averages are completely off. I'll have to dig a bit.

bartgol:
After inspecting the output of branch and master, I concluded that master was buggy, and this PR fixes it. In particular, for avg output, master was saving in rhist files the current average rather than the current sum. This caused the 2nd run to restart the sum with an avg instead.

For instance, i looked at ps (which should be of the order of 1e5) for a ERS_Ln9 test (which writes a restart after 5 steps, and should compute the avg over all 9 steps). The rhist file contained ps approx 1e5 for master and 5e5 for the branch (which is the correct tally after 5 steps). The final output contained 5.5e5 for master, which is roughly equal to (1e6 (the rhist value) + 4*e6  (the remaining 4 steps) divided by 9 (the total number of steps). The branch, otoh, has a correct ~1e6 avg value for ps at the end of the 9 steps.

@AaronDonahue @PeterCaldwell was any of our production runs saving AVG values, with restarts written in the middle of an avg interval? If so, I'm afraid that output has to be thrown away. Instant output otoh (including model restart files themselves) is correct.

If there is no objection, I am going to merge this.

PR #2331, opened 2023-05-15
Update weaver mach file and mach specs to adapt to new system modules

bartgol:
I also had to do a minor C++ fix, to accommodate a finicky compiler. With these mods, I can compile scream (with test-all), but I cannot yet run the tests, since there are still issues with propagating modules on the compute nodes

### DIFF
--- a/components/eamxx/cmake/machine-files/weaver.cmake
+++ b/components/eamxx/cmake/machine-files/weaver.cmake
@@ -1,8 +1,9 @@
 include(${CMAKE_CURRENT_LIST_DIR}/common.cmake)
 common_setup()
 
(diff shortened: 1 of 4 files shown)


jgfouca (approved):

bartgol:
@jgfouca The tests are running on weaver, and passing. I get a couple of fails, but for separate reasons:
- run_and_cmp have old baselines (pre qv sat PR merge), and I did not run with `-i`
- some tests require the netCDF4 package, but pip is erroring out (see attachment)

I am for merging this, since testing cannot even start on weaver right now, due to python 3.7 module not existing anymore...

[pip-error.txt](https://github.com/E3SM-Project/scream/files/11493039/pip-error.txt)

bartgol:
I'm going to merge this, so we can at least get _some_ testing going on weaver tonight.

PR #2333, opened 2023-05-16
EAMxx: Add BFB hasher with internal_diagnostics_level control.

ambrad:
When an atmosphere process's internal_diagnostics_level > 0, print global hash values before and after running the process. The default level is 0.

Add a testmod that sets the level to 1 for all processes and runs a POSTRUN_SCRIPT to check if the hashes are all the same.

Append this testmod to an existing test.


ambrad:
Some machines prefix each line of e3sm.log with "1:" and "2:" in ERS tests, so I had to do more manual diff'ing in the python script than I had originally done. That's why I don't just use zdiff.

ambrad:
The old test with the new testmod passes on Chrysalis and Ascent.

golaz:
@ambrad : if I understand this new feature correctly, this will be extremely useful to have. But I would prefer if it was on by default. My experience with global sums in EAM atm.log is that one never knows when they will be needed, so it's good practice to have them on no matter what. 

In fact, I was just using them today as I was trying (and struggling) to BFB reproduce an older E3SMv2 simulation.

ambrad:
> But I would prefer if it was on by default.

This PR is for fine-grained and expensive diagnostics. A previous PR put in what you'd like. I can switch the default value for that hash computation to non-0.

golaz:
> This PR is for fine-grained and expensive diagnostics. A previous PR put in what you'd like.

Thanks, @ambrad. I didn't know my use-case this was already taken care of.

components/eamxx/cime_config/testdefs/testmods_dirs/scream/internal_diagnostics_level/shell_commands
```diff
@@ -0,0 +1,9 @@
+# Can't change atm_proc_base::internal_diagnostics_level, so change each one
```
bartgol, line 1: I think this has been fixed in #2334 , so you should be able to just do
```
./xmlchange --append SCREAM_ATMCHANGE_BUFFER=" --all internal_diagnostics_level=1"
```
You should probably try it first, though.

components/eamxx/tests/postrun/check_hashes_ers.py
```diff
@@ -0,0 +1,98 @@
+#!/usr/bin/env python3
```
bartgol, line 1: Would be nice to check this script in our nightlies. I'm not really sure how to do it though. I guess we'd have to introduce some bug in some atm proc, to trigger a diff in the hash. Perhaps not too important, unless @jgfouca has some clear idea of how to do it (and assuming it's not too time consuming, and doesn't require tons of additions in the code).

ambrad: I may not follow, but to be clear, this script will run at the end of the RUN phase for ERS_Ln22.ne30_ne30.F2010-SCREAMv1.scream-internal_diagnostics_level in our nightlies as a result of the POSTRUN_SCRIPT setting in the internal_diagnostics testmod. If the script detects no hash diffs, it will return 0, permitting the RUN phase to pass. If it detects diffs, it will return non-0, forcing the RUN phase to fail.

components/eamxx/cime_config/testdefs/testmods_dirs/scream/internal_diagnostics_level/shell_commands
ambrad: Looks like it's not working for my use case. I need internal_diagnostics_level for the EAMxx-labeled process to be 0, but I can't seem to specify anything like that. For example, this doesn't seem to work:
```
./atmchange --all internal_diagnostics_level=1 atmosphere_processes::internal_diagnostics_level=1
```
And `EAMxx::internal_diagnostics_level` doesn't exist; I seem to recall it did previously. My original shell_commands still works, though, so I'm going to stick with that.

components/eamxx/tests/postrun/check_hashes_ers.py
bartgol: What I meant is to have a way to ensure that this script works. I was thinking of having an ERS test where the hash _does_ change, and verify that the script indeed triggers a fail. Along the line of what we already do in our scripts-tests for standalone tests.

But as I said, this is not too important. I'm assuming you already checked the script does fail when it's supposed to anyways.

ambrad: Yes. One easy way to make it fail is to exclude `atmosphere_processes::internal_diagnostics_level=0` from the `atmchange` command. Something in the "EAMxx" process is including buffers that aren't actually filled, leading to run-to-run diffs at every time step. I found and fixed a similar problem in the Homme AD interface but haven't looked at the EAMxx process.

ambrad: This isn't what you were asking for, but I was reminded that I wanted to separate out and unit test the core hashing routines, which I've now done.

bartgol (approved):

PR #2334, opened 2023-05-16
Upgrades to atmquery/atmchange

bartgol:
There are two separate upgrades:

- `atmquery --grep str`: allows to look for all the XML entries whose name (partially) matches `str`. E.g.,
```
$ ./atmquery --grep precond
    atmosphere_processes::enable_precondition_checks: false
    sc_import::enable_precondition_checks: false
    homme::enable_precondition_checks: false
    physics::enable_precondition_checks: false
    mac_aero_mic::enable_precondition_checks: false
    shoc::enable_precondition_checks: false
    cldFraction::enable_precondition_checks: false
    spa::enable_precondition_checks: false
    p3::enable_precondition_checks: false
    rrtmgp::enable_precondition_checks: false
    sc_export::enable_precondition_checks: false
```
- `atmchange --all <change-list>`: allows to apply the given changes to all matches. Eg.,
```
$ ./atmchange --all enable_precondition_checks=true --all
$ ./atmquery --grep precond
    atmosphere_processes::enable_precondition_checks: true
    sc_import::enable_precondition_checks: true
    homme::enable_precondition_checks: true
    physics::enable_precondition_checks: true
    mac_aero_mic::enable_precondition_checks: true
    shoc::enable_precondition_checks: true
    cldFraction::enable_precondition_checks: true
    spa::enable_precondition_checks: true
    p3::enable_precondition_checks: true
    rrtmgp::enable_precondition_checks: true
    sc_export::enable_precondition_checks: true
```

### DIFF
--- a/components/eamxx/scripts/atm_manip.py
+++ b/components/eamxx/scripts/atm_manip.py
@@ -2,7 +2,7 @@
 Retrieve nodes from EAMxx XML config file.
 """
 
-import sys, os
+import sys, os, re
 
 # Used for doctests
 import xml.etree.ElementTree as ET # pylint: disable=unused-import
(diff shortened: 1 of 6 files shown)


ambrad:
One additional thing that might be nice, in case it doesn't work already, is for this to work:
```
/atmchange --all enable_precondition_checks=true p3::enable_precondition_checks=false
```
This sets a global default and then changes the value for one process.

bartgol:
> One additional thing that might be nice, in case it doesn't work already, is for this to work:
> 
> ```
> /atmchange --all enable_precondition_checks=true p3::enable_precondition_checks=false
> ```
> 
> This sets a global default and then changes the value for one process.

I think this should work as expected. The `--all` applies to all changes, but obviously only the first one has 2+ matches. And yes, you need to set the general first, then do the exceptions.

bartgol:
Ok, I tested the feature on this simplified XML:
```
<namelist_defaults>
    <spa>
      <enable_precondition_checks type="logical">true</enable_precondition_checks>
    </spa>
    <p3>
      <enable_precondition_checks type="logical">true</enable_precondition_checks>
    </p3>
    <rrtmgp>
      <enable_precondition_checks type="logical">true</enable_precondition_checks>
    </rrtmgp>
</namelist_defaults>
```
This is what I get:
```
$ ./andquery --grep precon
    spa::enable_precondition_checks: true
    p3::enable_precondition_checks: true
    rrtmgp::enable_precondition_checks: true
$ ./atmchange --all enable_precondition_checks=false p3::enable_precondition_checks=true
$ ./atmquery --grep precond
    spa::enable_precondition_checks: false
    p3::enable_precondition_checks: true
    rrtmgp::enable_precondition_checks: false
```
which is what we wanted.

jgfouca (approved):

components/eamxx/scripts/atm_manip.py
```diff
@@ -250,71 +136,100 @@ def atm_config_chg_impl(xml_parent, xml_node, change, all_matches=False):
...
+
+    return any_change
 
-        node_found = node is not None
-        if node is not None:
-            any_change = apply_change (node,new_value,append_this)
+###############################################################################
+def create_parent_map(root):
+###############################################################################
+    return {c: p for p in root.iter() for c in p}
```
bartgol, line 215: The Python is strong in you. The double list comprehension is giving me headaches.

components/eamxx/scripts/atm_manip.py
```diff
@@ -105,81 +30,42 @@ def get_xml_node(xml_root,name,allow_not_found=False):
...
 
-    # If user specified selectors via namespace, recurse over them
-    for s in selectors[1:]:
-        expect (allow_not_found or num_nodes_with_name(node,s,recurse=False)>0,
-            f"Error! XML entry {s} not found in section {node.tag}")
-        expect (allow_not_found or num_nodes_with_name(node,s,recurse=False)==1,
-            f"Error! Multiple XML entries with name {s} found in section {node.tag}")
+    try:
+        xpath_str = prefix + name.replace("::", "/")
+        result = xml_root.findall(xpath_str)
```
bartgol, line 60: This is MUCH cleaner. Thanks!

jgfouca (approved):

PR #2336, opened 2023-05-17
EAMxx: workaround to make testing work on weaver

bartgol:
IT folks are examining why the module command is not found from within a script on weaver. Meanwhile, they suggested this reasonable workaround.


jgfouca (approved):

bartgol:
Tested manually, it works.

PR #2338, opened 2023-05-18
Add a time interpolation utility to EAMxx

AaronDonahue:
This commit introduces a new class, `TimeInterpolation` that can be used to manage data used for time interpolation.
The class can have data assigned directly, or be passed a list of files that contain all the data to be used for interpolation.

Furthermore, when using files, the initialization steps will ensure that the timestamps associated with the interpolation data
are sorted chronologically, avoiding weird answers if the user accidentally jumbles file order in the list.

This new class will be useful for nudging and forcing the surface export variables, two tasks that will be covered in subsequent PR's.


components/eamxx/src/share/util/eamxx_time_interpolation.cpp
```diff
@@ -0,0 +1,343 @@
...
+) : TimeInterpolation(grid)
+{
+  set_file_data_triplets(list_of_files);
+}
+/*-----------------------------------------------------------------------------------------------*/
+/* A function to perform time interpolation using data from all the fields stored in the local
+ * field managers.
+ * Conducts a simple linear interpolation between two points using
+ *        y* = w*y0 + (1-w)*y1
+ * where w = (t1-dt)/(t1-t0)
```
bartgol, line 33: Minor: seeing `dt` in the comment seems confusing. That's not a timestep, but a _timestamp_. So maybe `w=(t1-t) / (t1-t0)`?

components/eamxx/src/share/util/eamxx_time_interpolation.cpp
```diff
@@ -0,0 +1,343 @@
...
+  const Real w_num = m_time1 - time_in;
+  const Real w_den = m_time1 - m_time0;
+  const Real weight = w_num/w_den;
+
+  // Cycle through all stored fields and conduct the time interpolation
+  for (auto name : m_field_names)
+  {
+    const auto& field0 = m_fm_time0->get_field(name);
+    const auto& field1 = m_fm_time1->get_field(name);
+    Field field_out    = field0.clone();
```
bartgol, line 61: This would allocate a new field every time this routine is called. Perhaps, we should alloc the fields once, and reuse the same data structure over and over again.

components/eamxx/src/share/util/eamxx_time_interpolation.cpp
```diff
@@ -0,0 +1,343 @@
...
+  return interpolated_fields;
+}
+/*-----------------------------------------------------------------------------------------------*/
+/* Function which registers a field in the local field managers.
+ * Input:
+ *   field_in - Is a field with the appropriate dimensions and metadata to match the interpolation
+ * Output:
+ *   None
+ */
+void TimeInterpolation::add_field(const Field& field_in)
```
bartgol, line 76: This routine may be the right place to store the output fields: on top of `m_fm_time0` and `m_fm_time1`, we could have `m_fm_curr` (or `m_fm_out`, or whatever). After all, it's likely that the user of this routine is passing as `field_in` precisely the field that they will use at runtime, no?

components/eamxx/src/share/util/eamxx_time_interpolation.cpp
```diff
@@ -0,0 +1,343 @@
...
+/*-----------------------------------------------------------------------------------------------*/
+/* Function to shift the data of a single field from time1 to time0
+ * Input:
+ *   name - The name of the field to be shifted.
+ */
+void TimeInterpolation::shift_data(const std::string& name)
+{
+        auto  field0 = m_fm_time0->get_field(name);
+  const auto& field1 = m_fm_time1->get_field(name);
+  field0.deep_copy(field1);
```
bartgol, line 99: Rather than a deep copy, how about doing `std::swap(field0,field1)`? This should do a (soft) swap of internal metadata pointers, which means just a handful of pointers...

components/eamxx/src/share/util/eamxx_time_interpolation.cpp
```diff
@@ -0,0 +1,343 @@
...
+}
+/*-----------------------------------------------------------------------------------------------*/
+/* Function to shift all data from time1 to time0, update timestamp for time0
+ */
+void TimeInterpolation::shift_data()
```
bartgol, line 104: Can we remove the above method, and put its impl directly in here then? Will we ever need to shift data only for some field but not others? If not, there's no point in having 2 methods, I think.

components/eamxx/src/share/util/eamxx_time_interpolation.hpp
```diff
@@ -0,0 +1,86 @@
...
+  fm_type  m_fm_time1;
+  vos_type m_field_names;
+
+  // Store the timestamps associated with the two time snaps
+  TimeStamp m_time0;
+  TimeStamp m_time1;
+
+  // Variables related to the case where we use data from file
+  std::vector<DataFromFileTriplet>           m_file_data_triplets;
+  std::vector<DataFromFileTriplet>::iterator m_triplet_iterator;
```
bartgol, line 77: Doesn't look like you need to store this in the class

components/eamxx/src/share/util/eamxx_time_interpolation.cpp
```diff
@@ -0,0 +1,343 @@
...
+      ts_ref = ts_file_start;
+    }
+    // Gather information about time in this file
+    scorpio::register_file(filename,scorpio::Read);
+    const int ntime = scorpio::get_dimlen(filename,"time");
+    for (int tt=0; tt<ntime; tt++) {
+      auto time_snap = scorpio::read_time_at_index_c2f(filename.c_str(),tt+1);
+      TimeStamp ts_snap = ts_file_start;
+      if (time_snap>0) {
+        ts_snap += (time_snap*86400); // note, time is assumed to be in days.
```
bartgol, line 250: You could double check the `units` attribute of the `time` variable...

components/eamxx/src/share/util/eamxx_time_interpolation.cpp
```diff
@@ -0,0 +1,343 @@
...
+      map_of_times_to_vector_idx.emplace(time,running_idx);
+      filenames_tmp.push_back(filename);
+      timestamps_tmp.push_back(ts_snap);
+      time_idx_tmp.push_back(tt);
+      ++running_idx;
+    }
+  }
+  // Now that we have gathered all of the timesnaps we can arrange them in order as DataFromFileTriplet objects.
+  // Taking advantage of maps automatically self-sorting by the first arg.
+  for (auto a : map_of_times_to_vector_idx) {
```
bartgol, line 262: I think all you need is to store the first (and last, for safety) timestamp in each file. Then, using std::upper_bound or std::lower_bound, you can efficiently find which file you need to look in.

bartgol: Also, if we can assume we will alwasy "move forward" with time, and that our timestep will not be so large as to skip a file, then we don't really need to search.

components/eamxx/src/share/tests/eamxx_time_interpolation_tests.cpp
```diff
@@ -0,0 +1,404 @@
...
+std::shared_ptr<FieldManager> get_fm (const std::shared_ptr<const AbstractGrid>& grid, const util::TimeStamp& t0, const int seed);
+util::TimeStamp init_timestamp();
+std::vector<std::string> create_test_data_files(const ekat::Comm& comm, const std::shared_ptr<const GridsManager>& gm,const util::TimeStamp& t0, const int seed);
+
+// Functions needed to run the test
+void update_field_data(const Real slope, const Real dt, Field& field);
+bool views_are_approx_equal(const Field& f0, const Field& f1, const Real tol);
+
+// Helper randomizer
+auto my_pdf = [&](std::mt19937_64& engine) -> Real {
```
bartgol, line 42: Since you're at file scope, this can simply be a normal function, rather than a lambda...

components/eamxx/src/share/util/eamxx_time_interpolation.cpp
AaronDonahue: I didn't want to assume that our timestep will always be smaller than the data.  I think 99% of the time that will be true.  But there is no guarantee that this is the case.

components/eamxx/src/share/util/eamxx_time_interpolation.hpp
AaronDonahue: This may be an incorrect usage of an iterator, if so I can change how it is used elsewhere, but the use here is to store the location in the `m_file_data_triplets` vector that is the most recent data snap we read.  So at any given time in the simulation `m_triplet_iterator` is pointing to a different element of the vector, and being updated.

Alternatively (and in a previous version) I had an int that stored the index of the vector and would update that.  And then grab the element of `m_file_data_triplets` using the index.

components/eamxx/src/share/util/eamxx_time_interpolation.cpp
AaronDonahue: i don't see how we avoid a search one way or the other.  If we store upper and lower bounds that will make it easy to quickly know which file grab, but then we will still need to search through the `time` vector of that file to figure out which snap to take.

Unless storing the triplet information is expensive, in which case it would probably be better to search on demand rather then preorganize all the triplets.

components/eamxx/src/share/util/eamxx_time_interpolation.cpp
AaronDonahue: I'm not sure this will work, since the field manager will still be storing the swapped pointers.  Thus when the data is read as input it will end up going to field0 when it is supposed to go to field1.  Note - when I do the swap the unit test fails at the timestep where shift data should have been called.

bartgol: True, the fm returns a copy, not a reference. But if you call `get_field_ptr`, then you can swap the pointers content.

components/eamxx/src/share/util/eamxx_time_interpolation.cpp
AaronDonahue: That's a good question.  Currently, a user is allowed to update field data on a field by field basis (if they want).  This means we would want to only "shift_data" for that field at that time.  I envisioned a use case where we want to deal with internal interpolation, for example for an implicit solve.  So we march individual fields forward in time to populate the interpolation bounds and this might have to be done on a field by field basis.  But maybe this is a case where we add this use case when we need it, rather than design with it in mind.

bartgol: I think it's safe to assume we update all of them. If we have two groups of fields, we can create two TimeInterpolation obj.

components/eamxx/src/share/util/eamxx_time_interpolation.cpp
```diff
@@ -0,0 +1,340 @@
...
+  }
+}
+/*-----------------------------------------------------------------------------------------------*/
+/* Function which registers a field in the local field managers.
+ * Input:
+ *   field_in - Is a field with the appropriate dimensions and metadata to match the interpolation
+ * Output:
+ *   None
+ */
+void TimeInterpolation::add_field(Field& field_in, const bool deep)
```
bartgol, line 73: `deep` is not a very clear name, imho. At first, I actually thought that if `deep=true`, you should create a deep copy of the field, while it's used the other way around in the code.

Perhaps you could call the arg something like `store_soft_copy`.

bartgol: Also, you can take a const ref to `field_in`, since you are not going to change it.

components/eamxx/src/share/util/eamxx_time_interpolation.cpp
```diff
@@ -0,0 +1,340 @@
...
+/*-----------------------------------------------------------------------------------------------*/
+/* Function to shift all data from time1 to time0, update timestamp for time0
+ */
+void TimeInterpolation::shift_data()
+{
+  for (auto name : m_field_names)
+  {
+    auto field0 = m_fm_time0->get_field(name);
+    auto field1 = m_fm_time1->get_field(name);
+    field0.deep_copy(field1);
```
bartgol, line 104: Using std::swap should be the way to go here. We should find a way to make it work.

components/eamxx/src/share/util/eamxx_time_interpolation.cpp
```diff
@@ -0,0 +1,340 @@
...
+ *
+ * It is assumed that any updated data is meant to replace the data at time1 and that the time1
+ * data should be shifted to time0
+ */
+void TimeInterpolation::update_data_from_field(const Field& field_in)
+{
+  const auto name = field_in.name();
+  auto field0 = m_fm_time0->get_field(name);
+  auto field1 = m_fm_time1->get_field(name);
+  field0.deep_copy(field1);
```
bartgol, line 188: Another place where std::swap should be used, to avoid the f1->f0 deep copy (we'll still need the deep copy f_in->f1.

components/eamxx/src/share/util/eamxx_time_interpolation.cpp
```diff
@@ -0,0 +1,340 @@
...
+/* Function to check the current set of interpolation data against a timestamp and, if needed,
+ * update the set of interpolation data to ensure the passed timestamp is within the bounds of
+ * the interpolation data.
+ * Input:
+ *   ts_in - A timestamp that we intend to interpolate onto.
+ */
+void TimeInterpolation::check_and_update_data(const TimeStamp& ts_in)
+{
+  // First check if the passed timestamp is within the bounds of time0 and time1.
+  bool current_data_okay = (ts_in.seconds_from(m_time0) >= 0) and (m_time1.seconds_from(ts_in) >= 0);
```
bartgol, line 297: Looks like the first check should always be true (unless we start from the left of the first time interval). I think the first check should be inside an `EKAT_REQUIRE_MSG`, to ensure we are not extrapolating on the left.

components/eamxx/src/share/util/eamxx_time_interpolation.cpp
bartgol: The way it is now, we store a triplet for each timestamp, which seems an overkill. Storing t0,tN for each file, and the timestamps in each file should be enough. Using something like std::upper_bound (twice), we should be able to a) locate the file, and b) locate the time index.

If you want, I can take a stab at this (next week).

AaronDonahue:
WIP pending #2340

components/eamxx/src/share/util/eamxx_time_interpolation.cpp
```diff
@@ -0,0 +1,345 @@
...
+  if (!current_data_okay) {
+    // The timestamp is out of bounds, need to load new data.
+    // First cycle through the DataFromFileTriplet's to find a timestamp that is greater than this one.
+    bool found = false;
+    int step_cnt = 0; // Track how many triplets we passed to find one that worked. 
+    while (m_triplet_iterator != m_file_data_triplets.end() and step_cnt < m_file_data_triplets.size()) {
```
bartgol, line 308: The second check is redundant. Since you start from 0, and since the iterator cannot go past the end, the count will always be less than `m_file_data_triplets.size()`.

components/eamxx/src/share/io/scorpio_input.hpp
```diff
@@ -103,10 +103,15 @@ class AtmosphereInput
   // Cleans up the class
   void finalize();
 
+  // Getters
+  std::string get_filename() { return m_filename; } // Simple getter to query the filename for this stream.
+
+  // Expose the ability to set field manager for cases like time_interpolation where we swap fields
+  // between field managers to avoid deep_copy.
+  void set_field_manager (const std::shared_ptr<const fm_type>& field_mgr);
```
bartgol, line 111: If you want to allow to reset a FM, we should add some checks, to ensure that the fields in the new FM have the same layout (across all ranks). If accidentally they don't, we'd end up with reading errors, since the data structure in scorpio have already been set (e.g., the dofs offsets have already been set).

components/eamxx/src/share/util/eamxx_time_interpolation.cpp
AaronDonahue: I thought so to, but I have been conditioned to add a fail-safe to while loops since they can go infinite so easily and the code will just burn through the allocation without warning.  Then time-out, which is useless for debugging.

AaronDonahue:
@bartgol , I think I've addressed all comments.  Would you mind re-reviewing when you have a chance?  Once the FieldMan PR goes in I can toggle this for AUTOMERGE, assuming it passes the AT.

components/eamxx/src/share/util/eamxx_time_interpolation.cpp
bartgol: I would still remove it. Redundant checks can be misleading (someone may think step_cnt _could_ go past the triplets size).

bartgol (changes requested):
Once you remove the redundant check, this can be merged.

bartgol (approved):

PR #2340, opened 2023-05-19
Some small simplifications to FieldManager

bartgol:
In particular:

* Allow to call `add_field` without having to call `registration_begins/ends` back to back. If repo is Clean when `add_field` is called, it will be immediately set to Closed (cannot "register" fields anymore, only add pre-built ones). This simplifies things in a few places that only need a FM with pre-build fields.
* Move `get_field_ptr` to protected section. The only outside customer of this method was the mass/energy check, which did not need pointers anyways.
* Add non-const `get_field` returning a reference to the Field. This can be useful with "local" field managers (which are non-const), so we can modify fields on the fly.

@AaronDonahue I think point 3 may make swapping fields between two FM's easier for you.

### DIFF
--- a/components/eamxx/src/control/atmosphere_driver.cpp
+++ b/components/eamxx/src/control/atmosphere_driver.cpp
@@ -346,32 +346,20 @@ void AtmosphereDriver::setup_column_conservation_checks ()
 
   // Get fields needed to run the mass and energy conservation checks. Require that
   // all fields exist.
-  const auto pseudo_density_ptr = phys_field_mgr->get_field_ptr("pseudo_density");
-  const auto ps_ptr             = phys_field_mgr->get_field_ptr("ps");
-  const auto phis_ptr           = phys_field_mgr->get_field_ptr("phis");
-  const auto horiz_winds_ptr    = phys_field_mgr->get_field_ptr("horiz_winds");
-  const auto T_mid_ptr          = phys_field_mgr->get_field_ptr("T_mid");
-  const auto qv_ptr             = phys_field_mgr->get_field_ptr("qv");
-  const auto qc_ptr             = phys_field_mgr->get_field_ptr("qc");
-  const auto qr_ptr             = phys_field_mgr->get_field_ptr("qr");
-  const auto qi_ptr             = phys_field_mgr->get_field_ptr("qi");
-  const auto vapor_flux_ptr     = phys_field_mgr->get_field_ptr("vapor_flux");
-  const auto water_flux_ptr     = phys_field_mgr->get_field_ptr("water_flux");
-  const auto ice_flux_ptr       = phys_field_mgr->get_field_ptr("ice_flux");
-  const auto heat_flux_ptr      = phys_field_mgr->get_field_ptr("heat_flux");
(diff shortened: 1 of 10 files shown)


tcclevenger (approved):
A lot cleaner.

components/eamxx/src/control/atmosphere_driver.cpp
```diff
@@ -381,14 +369,28 @@ void AtmosphereDriver::setup_column_conservation_checks ()
...
-                                                           pseudo_density_ptr, ps_ptr, phis_ptr,
-                                                           horiz_winds_ptr, T_mid_ptr, qv_ptr,
-                                                           qc_ptr, qr_ptr, qi_ptr,
-                                                           vapor_flux_ptr, water_flux_ptr,
-                                                           ice_flux_ptr, heat_flux_ptr);
+                                                           pseudo_density, ps, phis,
+                                                           horiz_winds, T_mid, qv,
+                                                           qc, qr, qi,
+                                                           vapor_flux, water_flux,
+                                                           ice_flux, heat_flux);
```
AaronDonahue, line 393: Line 473 below: 
```
auto f = rel_fm->get_field_ptr(fname);
```

AaronDonahue (changes requested):

components/eamxx/src/control/atmosphere_driver.cpp
AaronDonahue:
```[donahue5@quartz770:scream_sandbox]$ git diff
diff --git a/components/eamxx/src/control/atmosphere_driver.cpp b/components/eamxx/src/control/atmosphere_driver.cpp
index 2f1e3e6fc5..b3fd88202e 100644
--- a/components/eamxx/src/control/atmosphere_driver.cpp
+++ b/components/eamxx/src/control/atmosphere_driver.cpp
@@ -470,12 +470,12 @@ void AtmosphereDriver::create_fields()
         // Loop over all fields in group src_name on grid src_grid.
         for (const auto& fname : rel_info->m_fields_names) {
           // Get field on src_grid
-          auto f = rel_fm->get_field_ptr(fname);
+          auto f = rel_fm->get_field(fname);
 
           // Build a FieldRequest for the same field on greq's grid,
           // and add it to the group of this request
           if (fvphyshack) {
-            const auto& sfid = f->get_header().get_identifier();
+            const auto& sfid = f.get_header().get_identifier();
             auto dims = sfid.get_layout().dims();
             dims[0] = fm->get_grid()->get_num_local_dofs();
             FieldLayout fl(sfid.get_layout().tags(), dims);
@@ -483,7 +483,7 @@ void AtmosphereDriver::create_fields()
             FieldRequest freq(fid,req.name,req.pack_size);
             fm->register_field(freq);
           } else {
-            const auto fid = r->create_tgt_fid(f->get_header().get_identifier());
+            const auto fid = r->create_tgt_fid(f.get_header().get_identifier());
             FieldRequest freq(fid,req.name,req.pack_size);
             fm->register_field(freq);
           }
```

worked for me.

AaronDonahue (approved):

AaronDonahue (approved):

PR #2341, opened 2023-05-19
Fix ML correction unit test

elynnwu:
When `SCREAM_FPE=True`, importing numpy causes FPE. For the unit test, we can just disable it. In the production ML correction, we would want to enable fpes after python calls are done.

### DIFF
--- a/components/eamxx/tests/uncoupled/ml_correction/ml_correction_standalone.cpp
+++ b/components/eamxx/tests/uncoupled/ml_correction/ml_correction_standalone.cpp
@@ -21,11 +21,11 @@ TEST_CASE("ml_correction-stand-alone", "") {
   ekat::ParameterList ad_params("Atmosphere Driver");
   parse_yaml_file(fname, ad_params);
 
-  const auto& ts     = ad_params.sublist("time_stepping");
-  const auto  dt     = ts.get<int>("time_step");
-  const auto  nsteps = ts.get<int>("number_of_steps");
-  const auto  t0_str = ts.get<std::string>("run_t0");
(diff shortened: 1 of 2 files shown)


components/eamxx/tests/uncoupled/ml_correction/ml_correction_standalone.cpp
```diff
@@ -62,6 +62,7 @@ TEST_CASE("ml_correction-stand-alone", "") {
   reference += 0.1;
   Real reference2 = qv(1, 30);
   reference2 += 0.1;
+  ekat::disable_all_fpes();  // required for importing numpy
```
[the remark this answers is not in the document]
elynnwu: Yes, that is the plan. I didn't do it here because there's no more computation after coming back from python in this test. For consistency, I will re-enable it so that we don't forget to do that in the real ML correction module.

components/eamxx/tests/uncoupled/ml_correction/ml_correction_standalone.cpp
```diff
@@ -71,6 +73,8 @@ TEST_CASE("ml_correction-stand-alone", "") {
           num_cols * num_levs, qv.data(), py::str{}),
       num_cols, num_levs);
   py::gil_scoped_release no_gil;
+  int fpe_mask = ekat::get_enabled_fpes();
```
jeff-cohere, line 76: What's this fpe_mask for?

elynnwu: It should get passed to `ekat::enable_fpes`, updated

jeff-cohere (approved):

PR #2343, opened 2023-05-22
Update machine config for Ruby TOSS4

elynnwu:
Same update as in quartz. Still need `pnetcdf` for saving output from larger jobs, pending LC to expose the library.


AaronDonahue (approved):
Great job @elynnwu , thanks for getting on this so quickly.

PR #2344, opened 2023-05-23
Create standalone pg2 unit test

tcclevenger:
CIME now defaults to pg2 physics grid in most cases, so standalone EAMxx should have a pg2 test. 

The main change needed was to remove some `#ifdef SCREAM_CIME_BUILD` guards in the driver to expose PG2 related code in standalone runs. Also, now the driver needs to create a dynamics library to compile, and guards needed to be added if user does not define homme dycore.

Currently the test is ne2. Takes about 1.5 min to run (Debug, serial, 1 rank). Equivalent ne4 takes 8 min. Let me know if there are preferences for what this unit test should be.


components/eamxx/src/control/CMakeLists.txt
```diff
@@ -13,7 +35,7 @@ set(SCREAM_CONTROL_HEADERS
 )
 
 add_library(scream_control ${SCREAM_CONTROL_SOURCES})
-target_link_libraries(scream_control PUBLIC scream_share scream_io)
+target_link_libraries(scream_control PUBLIC scream_share scream_io ${dynLibName})
```
bartgol, line 38: If I understand correctly, you need the dyn lib for `fv_phys_rrtmgp_active_gases_init` and `fv_phys_rrtmgp_active_gases_set_restart`, right? Those free functions do not seem to have any strict dependency on the dyn lib, so maybe we should move those to something like `share/util/eamxx_fv_phys_rrtmgp_active_gases_workaround.*pp`, along with the type `TraceGasesWorkaround`. Then, we would not have to link dyn to the control lib.

components/eamxx/tests/coupled/dynamics_physics/homme_shoc_cld_spa_p3_rrtmgp_pg2/homme_shoc_cld_spa_p3_rrtmgp_pg2_gll_output.yaml
```diff
@@ -0,0 +1,26 @@
+%YAML 1.1
```
bartgol, line 1: We should be able to write both gll and pg2 fields in the same file (restart already does that). Perhaps we should do it here too, to have one more instance of 2+ grids fields written on the same mesh?

components/eamxx/tests/coupled/dynamics_physics/homme_shoc_cld_spa_p3_rrtmgp_pg2/homme_shoc_cld_spa_p3_rrtmgp_pg2_diagnostics.yaml
```diff
@@ -0,0 +1,41 @@
+%YAML 1.1
```
bartgol, line 1: This output stream could be merged with the others, no?

components/eamxx/tests/coupled/dynamics_physics/homme_shoc_cld_spa_p3_rrtmgp_pg2/CMakeLists.txt
```diff
@@ -0,0 +1,123 @@
...
+# This test requires CPRNC
+if (TEST_RANK_END GREATER TEST_RANK_START)
+  include (BuildCprnc)
+  BuildCprnc()
+
+  set (BASE_TEST_NAME "homme_shoc_cld_spa_p3_rrtmgp_pg2")
+  math (EXPR CMP_RANK_START ${TEST_RANK_START}+1)
+  foreach (MPI_RANKS RANGE ${CMP_RANK_START} ${TEST_RANK_END})
+
+    set (SRC_FILE "${BASE_TEST_NAME}_output.INSTANT.nsteps_x${NUM_STEPS}.np${MPI_RANKS}.${RUN_T0}.nc")
```
bartgol, line 93: If we only had one output nc file, this would be easier, no?

components/eamxx/src/control/CMakeLists.txt
bartgol: We could make TraceGasesWorkaround a singleton, so that it's accessible from both the `fv_phys_rrtmgp_active_gases_xyz` routines, as well as from HommeDynamics.

components/eamxx/tests/coupled/dynamics_physics/homme_shoc_cld_spa_p3_rrtmgp_pg2/homme_shoc_cld_spa_p3_rrtmgp_pg2_gll_output.yaml
tcclevenger: Yes, I'll combine. I did this because ncvis doesn't know how to output nicely when `ncol_d` is used (instead of `ncol`) and I was checking output. Didn't think to switch it back.

components/eamxx/src/control/CMakeLists.txt
tcclevenger: Yeah, this is a good idea. I'll work on it.

components/eamxx/tests/coupled/dynamics_physics/homme_shoc_cld_spa_p3_rrtmgp_pg2/CMakeLists.txt
tcclevenger: This is the convention in some of the other tests. I'll combine for them as well all in a standalone commit. Unless you object since unrelated to this PR.

components/eamxx/tests/coupled/dynamics_physics/homme_shoc_cld_spa_p3_rrtmgp_pg2/homme_shoc_cld_spa_p3_rrtmgp_pg2_gll_output.yaml
tcclevenger: No, I'll combine. Most people doing actual output are using better tools.

tcclevenger:
Investigating the fails.

bartgol (approved):

PR #2347, opened 2023-05-24
Bug Fix - Fix ml_correction standalone test header declaration

AaronDonahue:
Fix to ml_correction standalone test to changes introduced in #2324 

This is a quick fix to a header file name change that was introduced in a previous commit, but missed because the ml_correction test is only run on a few machines for testing.


AaronDonahue:
Manually merging, only impacts failed tests on `quartz` which I've manually checked and is a quick fix sure to not impact the other test suites.

PR #2349, opened 2023-05-24
EAMxx: Add a tolerance on #diff lines in exxhash checker.

ambrad:
Flushed prints to e3sm.log can sometimes conflict with other output. Permit up to 'thr' diffs so we don't fail due to badly printed lines. This isn't a big loss in checking because an ERS_Ln22 second run writes > 1000 hash lines, and a true loss of BFBness is nearly certain to propagate to a large number of subsequent hashes.


ambrad:
Merging manually and without review b/c this PR touches just the POSTRUN_SCRIPT file for a non-AT-run test. Tested on Chrysalis.

PR #2330, opened 2023-05-15
Default ne120 and ne512 to use rough topo

tcclevenger:
Update `namelist_defaults_scream.xml`

PR #2351, opened 2023-05-26
Add geopotential height diagnostic

tcclevenger:
Previously, `VerticalLayer` diags were computed from sea level (i.e., `surf_geopotential=0`). This adds two diagnostics, `z_*` and `geopotential_*` where `z_*` is calculated with `surf_geopotential=0` and `geopotential_*` with `surf_geopotential=phis(icol)`.


AaronDonahue (commented):
@tcclevenger This is great.  Thanks for putting it together so quickly and I like the elegant solution.  I am happy to approve but I want to get feedback from @PeterCaldwell and/or @golaz on the name of the output which is switching to z_int and z_mid.

PeterCaldwell:
I like the naming convention, but I'd also like Chris's opinion.

AaronDonahue:
@golaz , when you get a chance can you comment on the naming convention in this PR?

golaz:
@tcclevenger , @AaronDonahue  : thanks for implementing this new variable. The naming convention is clear and reasonable.

components/eamxx/src/diagnostics/vertical_layer_interface.cpp
```diff
@@ -7,9 +7,16 @@ namespace scream
 VerticalLayerInterfaceDiagnostic::VerticalLayerInterfaceDiagnostic (const ekat::Comm& comm, const ekat::ParameterList& params)
```
bartgol, line 7: Do you think it's worth it to also merge the int and mid version into a single `VerticalLayer` diagnostic? There seem to be a lot of code duplication, and the only difference is an extra call to `calculate_z_mid` in the mid case (something that we could handle with a simple if statement)...

tcclevenger: Yeah, a lot of duplicated code. I'll add that.

tcclevenger:
@bartgol Combined all the VerticalLayer* diags. Using `dz` for the old `VerticalLayerThickness`.

components/eamxx/src/diagnostics/vertical_layer.cpp
```diff
@@ -0,0 +1,124 @@
...
+      }
+    }
+  });
+
+  // Copy data to the diagnostic output. For dz, z_mid, and geopotential_mid,
+  // z_mid contains the correct values. For z_int and geopotential_int, z_int.
+  if (m_is_interface_layout) Kokkos::deep_copy(diag_view, z_int);
+  else                       Kokkos::deep_copy(diag_view, z_mid);
+
+  const auto ts = get_field_in("qv").get_header().get_tracking().get_time_stamp();
```
bartgol, line 120: The base class already takes care of setting the timestamp in the diag field. so this is not needed.

components/eamxx/src/diagnostics/vertical_layer.hpp
```diff
@@ -9,10 +9,10 @@ namespace scream
 {
 
 /*
- * This diagnostic will produce the potential temperature.
+ * This diagnostic will produce the vertical layer height at midpoint.
```
bartgol, line 12: The class does much more now. We should write in the comment all 5 possible quantities that it can compute.

bartgol (changes requested):
Nice cleanup! There's definitely one thing to fix (string used on device), and a storage/runtime easy optimization in the diag implementation.

bartgol (dismissed):

tcclevenger:
@bartgol accidentally undid some changes with master rebase. I'll need another approval.

bartgol (dismissed):

bartgol (dismissed):

tcclevenger:
@bartgol It looks like https://github.com/E3SM-Project/scream/pull/2327 has added fails to master. Commit d35ea9d1938c66cac6a3a6314091a08236d5cbf7 here fixes the issues with PG2 test, the other is an issue with time_iterpolation test.

bartgol:
@tcclevenger it should have not added fails to master. It was passing everything, except for v1 baselines cmp. But it did change our tests to use ne2 IC/topo files, so yes, your commit should fix that.

tcclevenger:
I saw the same fails for `time_interpolation_np1` in https://github.com/E3SM-Project/scream/pull/2330 which doesn't touch any src code. It looks like it doesn't like the way the PointGrid is setup in the test.

AaronDonahue:
> I saw the same fails for `time_interpolation_np1` in #2330 which doesn't touch any src code. It looks like it doesn't like the way the PointGrid is setup in the test.

I'm working on a fix now.  Should be in this morning.  On a separate note, I also see fails on the CDashboard in the `[homme_shoc_cld_spa_p3_rrtmgp_pg2_np2](https://my.cdash.org/test/81705760)` tests that I think are unrelated.

tcclevenger:
Setting WIP until AT fixes are merged.

bartgol (approved):

PR #2357, opened 2023-06-02
Bugfix to TimeInterpolation that was introduced by recent changes to IO

AaronDonahue:
This commit fixes a few issues in master that arose from conflicts between the TimeInterpolation class and recent changes to IO introduced by #2327.

It brings the IO calls in the TimeInterpolation unit test in alignment with the changes to IO.

It also introduces a destructor and a finalize routine to TimeInterpolation to ensure that any open AtmosphereInput objects in use by a TimeInterpolation object are properly finalized before the end of the run.

### DIFF
--- a/components/eamxx/src/share/io/scream_scorpio_interface.F90
+++ b/components/eamxx/src/share/io/scream_scorpio_interface.F90
@@ -395,7 +395,16 @@ contains
     ! Get a new variable pointer in var_list
     if (len_trim(shortname)>max_hvarname_len) call errorHandle("PIO Error: variable shortname "//trim(shortname)//" is too long, consider increasing max_hvarname_len or changing the variable shortname",-999)
     curr => pio_atm_file%var_list_top
-
+    do while ( associated(curr) )
+      if (associated(curr%var)) then
+        if (trim(curr%var%name)==trim(shortname) .and. curr%var%is_set) then
+          exit
+        end if
+      end if
+      prev => curr
+      curr => prev%next
+    end do
+    
     allocate(prev%next)
     curr => prev%next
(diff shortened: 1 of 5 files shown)


components/eamxx/src/share/io/scream_scorpio_interface.F90
```diff
@@ -400,6 +400,10 @@ subroutine register_variable(filename,shortname,longname,units, &
     do while (associated(curr))
       if (associated(curr%var)) then
         if (trim(curr%var%name)==trim(shortname) .and. curr%var%is_set) then
+          ! TODO: For Luca, not sure why this is getting triggered for the time_interpolation test.  We are registering two
```
bartgol, line 403: We should remove this check. Our current impl allows to call register_variable multiple times, provided that the metadata is _all_ the same. In fact, in the cpp file, we do perform all these checks.

Notice that we cannot remove the actual call to `register_variable_c2f` in the cpp function if the var is already in the file, since we don't check our F90 metadata, but rather ping the file directly via scorpio calls (so we _must_ call register even if the var is in the file, since we may still need to setup the F90 hist var metadata).

bartgol: The reason why this only appears here is that time interp is the only place that opens the same file twice at the same time (other places close the file before reopening again, like for IC and vert coord).

AaronDonahue: Yes, that's what I figured.  Thanks for confirming.  I'll remove the check.

bartgol (approved):

bartgol (approved):

PR #2358, opened 2023-06-05
Update PG2 standalone test IC files

tcclevenger:
Fixes error in AT.


tcclevenger:
@AaronDonahue This is the fix for the PG2 fail.

AaronDonahue (approved):

AaronDonahue:
This is likely to still fail the AT because of the time_interpolation issue.  That being said, when the AT reports back if the PG2 tests report passing we can manually merge this so we are part of the way to having our master working again.

tcclevenger:
@AaronDonahue checked all 3 machines, only time_interpolation failing. I'll merge.

PR #2364, opened 2023-06-07
Use parallel-netcdf on TOSS4 machines

elynnwu:
During the initial TOSS4 machine upgrade, `pnetcdf` became unavailable and we reverted to using regular `netcdf` for quartz and ruby. However, it would result in `NetCDF: One or more variable sizes violate format constraints` when writing out restart files for large simulations (ne120 errors out but not ne30). LC has now exposed `parallel-netcdf` for both quartz and ruby. In this PR, we change the default behavior to use `parallel-netcdf`.


AaronDonahue:
Thanks @elynnwu for putting this together so quickly.  Have you tested it on Ruby and Quartz to make sure it works?

AaronDonahue:
failed because of an unrelated bug on master.  Retesting - which should pass now that the bugfix is in.

elynnwu:
> Thanks @elynnwu for putting this together so quickly. Have you tested it on Ruby and Quartz to make sure it works?

Yes, worked on quartz. Checking Ruby now and will merge after.

PR #2365, opened 2023-06-07
Remove host copy of uninitialized device memory

tcclevenger:
Should fix [scream_unit_tests_compute_santizer_initcheck](https://my.cdash.org/build/2347016) fails. Two issue:
1. We were requesting host copy of uninitialized device array in YAKL. We only need the host object, that we read values into from netcdf file, and then we deep copy back to device.
2. `rrtmgp_standalone_unit` test allocated an array that was unset. Uninitialized memory read occurred when copying to new array.

### DIFF
--- a/components/eamxx/src/physics/rrtmgp/simple_netcdf.hpp
+++ b/components/eamxx/src/physics/rrtmgp/simple_netcdf.hpp
@@ -81,7 +81,7 @@ namespace simple_netcdf {
                 std::vector<int> dimSizes(ndims);
                 size_t dimsize;
                 for (int i = 0; i < ndims; i++) {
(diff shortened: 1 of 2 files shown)


jgfouca (approved):
Nice! Does this change actually fix a bug or does it just make compute-sanitizer happy?

tcclevenger:
Yeah, I guess its not actually a bug. Just makes compute sanitizer happy. I'll remove the label.

tcclevenger:
@jgfouca Found another initcheck fail. Manually tested the rest so I think this should be the last one.

jgfouca:
@tcclevenger , my previous comment was not so much about the label; i'm trying to figure out if the things sanitizer is catching are useful.

jgfouca (approved):

PR #2367, opened 2023-06-08
EAMxx: allow CoarseningRemapper to handle strided subfields

bartgol:
This PR allows to run the CoarseningRemapper on fields that are strided subfields of other fields. It also fixes an issue within the Field class impl, that caused errors with strided fields.

To confirm things work as expected, I ran an ne30 case, with two output streams (one on ne30pg2 grid, and one remapped to ne4pg2). I then ran ncremap on the ne30 output and compared it to the ne4 output. The two are not bfb, due to how accumulations are done as well as to the fact that EAMxx runs in double precision, while the nc file contain single precision fields. The diffs, however, are within roundoff:
```
23c23
< -0.06780888
---
> -0.06780889
28c28
< -0.001327516
---
> -0.001327515
67c67
< -0.0007315932
---
> -0.0007315931
...
```

One last thing to do: add a remapped output stream to one of our ne30 nightly tests.

### DIFF
--- a/components/eamxx/src/share/field/field_impl.hpp
+++ b/components/eamxx/src/share/field/field_impl.hpp
@@ -200,8 +200,10 @@ deep_copy_impl (const Field& src) {
   //       field might be a subfield of another. We need the reshaped view.
   //       Also, don't call Kokkos::deep_copy if this field and src have
   //       different pack sizes.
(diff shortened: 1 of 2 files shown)


tcclevenger (approved):
Looks good.

bartgol:
I will add the coarsening test in a followup PR.

PR #2369, opened 2023-06-08
Update nudging atm process member variables

tcclevenger:
Fixes [scream_unit_tests_cuda_mem_check ](https://my.cdash.org/build/2348064) fails for the nudging test. Issue was member variables were being directly used inside `parallel_for()` functors, instead of local copies. 

Changed the following
- Use `m_*` naming for all member variables (naming made the error hard to detect)
- Removed some member variables that were only used locally inside a single function
- Remove member views, only keeping map of views that then local copies are made directly before parallel_fors.

### DIFF
--- a/components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
+++ b/components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
@@ -7,9 +7,8 @@ namespace scream
 // =========================================================================================
 Nudging::Nudging (const ekat::Comm& comm, const ekat::ParameterList& params)
   : AtmosphereProcess(comm, params)
-{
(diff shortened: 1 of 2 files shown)


jgfouca (approved):
Nice! So this caught a real problem?

tcclevenger:
> Nice! So this caught a real problem?

Yeah, I would say memcheck definitely catches real issues. And initcheck from the other PR caught 2 things that should have been fixed (unused view, unneeded deep_copy) although maybe that was by coincidence? I.e., that could still be happening many places and won't get caught.

racecheck is one that I'm not sure the current errors are useful. I'll look at the `--print-level` option to see if maybe changing to only printing errors is better (right now I _think_ we are failing because we are printing warnings).

PR #2370, opened 2023-06-08
Got mam4 standalone test working

jeff-cohere:
This PR incorporates lots of changes reflecting work done by me and @pbosler to continue integrating mam4xx into eamxx. We've created a standalone uncoupled test driver similar to those for P3 and SHOC and are using it to characterize the behavior of aerosol nucleation over several columns with a configuration that is easy to understand. We've also been working on an E3SM (F-case) simulation using this branch and are hoping to have something interesting to show for the E3SM all-hands meeting later this month.

As with prior EAGLES-related PRs, all of this code is protected by the `SCREAM_ENABLE_MAM` CMake option, which is `FALSE` by default. This is mainly because mam4xx requires C++17, and SCREAM isn't there yet. @pbosler and I will continue our integration process using feature branches, but this PR is our way of gradually introducing code into SCREAM.

The uncoupled test itself runs to completion, meaning that all required and updated fields are properly requested from the atmosphere driver. The data itself isn't correct yet--that's the next step! We had to make a few more disruptive changes to haero and mam4xx to get all these field requests to work properly, but it looks like we're through those obstacles.

After this PR is merged, we'll continue our integration process in one or more feature branches. At any time, please let @pbosler and I know if you have questions or concerns.


components/eamxx/src/physics/mam/eamxx_mam_microphysics_process_interface.hpp
```diff
@@ -0,0 +1,369 @@
...
+
+  Logger logger;
+
+  // number of horizontal columns and vertical levels
+  int ncol_, nlev_;
+
+  // Atmosphere processes often have a pre-processing step that constructs
+  // required variables from the set of fields stored in the field manager.
+  // This functor implements this step, which is called during run_impl.
+  struct Preprocess {
```
bartgol, line 90: I know we do it in P3 and SHOC too, so this is fine, but the pre/post process structs (and their impl) are not really well fit in this header. Perhaps a pimpl would be appropriate here.

components/eamxx/src/physics/mam/eamxx_mam_microphysics_process_interface.hpp
```diff
@@ -0,0 +1,369 @@
...
+  using view_2d       = typename KT::template view_2d<Real>;
+  using const_view_1d = typename KT::template view_1d<const Real>;
+  using const_view_2d = typename KT::template view_2d<const Real>;
+
+  // unmanaged views (for buffer and workspace manager)
+  using uview_1d = Unmanaged<typename KT::template view_1d<Real>>;
+  using uview_2d = Unmanaged<typename KT::template view_2d<Real>>;
+
+  // a quantity stored in a single vertical column with a single index
+  using ColumnView = mam4::ColumnView;
```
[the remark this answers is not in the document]
bartgol: Ah, I missed `aero_config_`. Though, tbh, that could be stored as a ptr, and the type simply fwd declared...

components/eamxx/src/physics/mam/CMakeLists.txt
```diff
@@ -1,11 +1,11 @@
 add_library(mam eamxx_mam_microphysics_process_interface.cpp)
 target_compile_definitions(mam PUBLIC EAMXX_HAS_MAM)
 add_dependencies(mam mam4xx_proj)
-target_include_directories(mam PRIVATE
+target_include_directories(mam PUBLIC
```
bartgol, line 4: Since you link mam4xx and haero below, do you really need these includes here?

components/eamxx/src/physics/mam/eamxx_mam_microphysics_process_interface.hpp
jeff-cohere: Yeah, that might be best in the end. Do you mind if we wait till things are further along before we do this sort of stuff? The pimpl pattern would require us to move more code out of the header (which is good, but I prefer getting things to work "at all" before cleaning up).

components/eamxx/src/physics/mam/CMakeLists.txt
jeff-cohere: I just removed this line and the headers weren't found, so evidently we do need them.

components/eamxx/src/physics/mam/eamxx_mam_microphysics_process_interface.hpp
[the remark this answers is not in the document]
jeff-cohere: Yeah, I'd like to clean this stuff up as we get more comfortable with all the moving parts.

components/eamxx/tpls/CMakeLists.txt
```diff
@@ -22,6 +22,24 @@ if (SCREAM_ENABLE_MAM)
...
+  endif()
+
+  # Normalize CMAKE_BUILD_TYPE.
+  string(TOLOWER "${CMAKE_BUILD_TYPE}" lc_build_type)
+  if (${lc_build_type} STREQUAL "debug")
+    set(mam_build_type "Debug")
+  elseif(${lc_build_type} STREQUAL "release")
+    set(mam_build_type "Release")
+  else()
+    set(mam_build_type "Debug") # when in doubt...
```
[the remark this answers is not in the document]
jeff-cohere: Removed.

components/eamxx/src/physics/mam/CMakeLists.txt
bartgol: It's obv a minor thing, so not really important. But I suspect the problem is with the imported target, which do not store any information regarding include dirs. If [here](https://github.com/E3SM-Project/scream/blob/580beae6b310c11c0f34ba9041ae63fdda99bbac/components/eamxx/tpls/CMakeLists.txt#L72) and [here](https://github.com/E3SM-Project/scream/blob/580beae6b310c11c0f34ba9041ae63fdda99bbac/components/eamxx/tpls/CMakeLists.txt#L97) you also do something like
```
target_include_directories (mam4xx PUBLIC ${MAM4XX_INSTALL_PREFIX}/${CMAKE_INSTALL_INCLUDEDIR})
```
then you don't need these. Though you may argue it's a potato potatoe kind of difference...

bartgol (approved):
My comments were more style than anything else. This PR is ok to merge.

jeff-cohere:
Thanks, Luca. I'll put it on my list to give this the pimpl treatmenr in a later PR.

PR #2374, opened 2023-06-09
Adds an EAMXX fully coupled compset

singhbalwinder:
A fully coupled compset that uses EAMXX as an atmospheric component is added. The compset runs fine for 5 days in full debug mode on Compy.

### DIFF
--- a/cime_config/allactive/config_compsets.xml
+++ b/cime_config/allactive/config_compsets.xml
@@ -14,7 +14,7 @@
     TIME_ATM[%phys]_LND[%phys]_ICE[%phys]_OCN[%phys]_ROF[%phys]_GLC[%phys]_WAV[%phys][_ESP%phys][_BGC%phys]
     Where for the EAM specific compsets below the following is supported
     TIME = Time period (e.g. 2000, HIST, RCP8...)
-    ATM  = [EAM, SATM, SCREAM]
+    ATM  = [EAM, EAMXX, SATM, SCREAM]
     LND  = [ELM,  SLND]
     ICE  = [MPASSI, CICE, DICE, SICE]
     OCN  = [MPASO, DOCN, SOCN]
@@ -329,6 +329,14 @@
   <alias>BGWCYCL1850</alias>
   <lname>1850_EAM%CMIP6_ELM%SPBC_MPASSI_MPASO_MOSART_MALI%STATIC_SWAV</lname>
 </compset>
+
+<!-- EAMXX fully coupled compset -->
+
+<compset>
+  <alias>WCYCLXX2010</alias>
+  <lname>2010_EAMXX_ELM%SPBC_MPASSI_MPASO_MOSART_SGLC_SWAV</lname>
+</compset>
+
 
 <!-- OCN + CICE + GLC Only Compsets -->
 <compset>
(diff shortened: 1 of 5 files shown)


cime_config/allactive/config_compsets.xml
```diff
@@ -330,6 +330,14 @@
   <lname>1850_EAM%CMIP6_ELM%SPBC_MPASSI_MPASO_MOSART_MALI%STATIC_SWAV</lname>
 </compset>
 
+<!-- EAMXX fully coupled compset -->
+
+<compset>
+  <alias>WCYCLXX2010</alias>
```
singhbalwinder, line 336: I chose this name arbitrarily, so I am open to changing it.

PeterCaldwell: @golaz and @rljacob  - do you have an opinion about how we name coupled compsets using EAMxx?

PeterCaldwell (dismissed):
Looks good to me. Others should review too, and we should get more input on what to name this compset.

jgfouca (dismissed):

jgfouca:
It struck me as a little odd that we have both "eamxx" and "scream" as supported atm component names even though they both refer to the same thing. I suppose we need this for now for backwards compatibility in our test suites.

PeterCaldwell:
good point Jim. I'd prefer to just always call it eamxx and delete the scream option, but maybe that will require a lot of work.

cime_config/allactive/config_compsets.xml
rljacob: We want do distinguish between a WCYCL case with EAM and one with EAMXX.  I think adding "XX" somewhere makes sense.

rljacob:
> good point Jim. I'd prefer to just always call it eamxx and delete the scream option, but maybe that will require a lot of work.

You don't have to change the name of any compset aliases,  just change the longname to use "eamxx" instead of "scream" and delete "scream" from the list of possible atmosphere models.

singhbalwinder:
> You don't have to change the name of any compset aliases, just change the longname to use "eamxx" instead of "scream" and delete "scream" from the list of possible atmosphere models.

That's a great suggestion @rljacob! I can do this in this PR and run all the tests to make sure all are working as expected.

PS. I WIPed this PR.

singhbalwinder:
@PeterCaldwell and I decided to merge this PR (after review) considering that the original objective of executing this compset has already been achieved. That is, this compset can run successfully for several time steps (currently till 538 time steps) in full debug mode on PM-GPU. I will create an issue so that we don't forget about unifying EAMxx and SCREAM in the ATM model definition in various config files.

PeterCaldwell:
Why isn't this merged already? Do we just need someone to manually merge it? @singhbalwinder - if you have a PR which you are waiting for someone to merge, I think it's your job to ask a merger to help you rather than letting the PR side idle for weeks... (which you probably know - I'm just trying to state the obvious in case we were on different pages).

singhbalwinder:
I have added the automerge label but it looks like it needs a fresh review before it can be merged. I should have asked you for a review. Would you please review it so that it is not blocked?

PeterCaldwell (approved):

PR #2378, opened 2023-06-13
Extension of the nudging process to be more general

AaronDonahue:
This commit makes a number of changes to the nudging process to support a more general implementation.

This includes:
1. Adding the nudging process to the namelist defaults so it can be used for CIME cases.
2. A runtime option to change the nudging timescale.
3. Make helper fields part of the generic atmosphere class instead of just surface coupling export.
4. Switch to runtime option for designating which fields will be nudged 
5. Use the TimeInterpolation class to handle the nudge data management 
6. Set masked values to the nearest column neighbor, to avoid strange nudging values
7. Option for cases where source data is stored on static pressure profiles, so not require `p_mid` in the source data file. 
8. Expand the unit testing to check the above changes (NOT DONE YET)


bartgol:
I'm not a fan of having both helper and internal fields as part of the AP interface. The two are very similar. I think we can merge them into a single thing.

AaronDonahue:
This PR appears to have an issue just with CIME based runs where `nudging` is turned on.  @elynnwu is helping investigate since I will be going on vacation for 2.5 weeks.

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
```diff
@@ -27,329 +41,206 @@ void Nudging::set_grids(const std::shared_ptr<const GridsManager> grids_manager)
...
+    auto pmid_ext = get_helper_field("p_mid_ext");
+    m_time_interp.add_field(pmid_ext,"p_mid",true);
+  } else if (m_src_pres_type == STATIC) {
+    // Load p_lev from source data file
+    create_helper_field("p_mid_ext", scalar2d_layout_mid, grid_name, ps);
+    auto pmid_ext = get_helper_field("p_mid_ext");
+    auto pmid_ext_v = pmid_ext.get_view<Real*,Host>();
+    ekat::ParameterList in_params;
+    in_params.set<std::vector<std::string>>("Field Names",{"p_lev"});
+    in_params.set("Filename",m_datafiles[0]);
```
elynnwu, line 101: The variable name used in vertical remapping is `p_levs`. I see that you're using the first `m_datafiles` to read in this field. By default, coarsening does not actually write `p_levs` as part of the output stream. Can we add another file name field in the name list that takes in the vertical remapping file, then use that to read `p_levs`?

AaronDonahue:
Newer, cleaner PR has been posted: #2484

PR #2379, opened 2023-06-13
Update compute-sanitizer tests

tcclevenger:
A few changes: 
- use `compute-sanitizer --tool=memcheck ./exe` instead of `cuda-memcheck ./exe`. Reason: when running `cuda-memcheck`...
```
========= CUDA-MEMCHECK
========= This tool is deprecated and will be removed in a future release of the CUDA toolkit
========= Please use the compute-sanitizer tool as a drop-in replacement
```
- Add `--racecheck-detect-level=error` for racecheck. I've tested manually, and when using the default (`=warn`) you get many many many "hazards" which cause the tester to fail. Every instance of these hazards I've looked at revolve around `Kokkos::parallel_reduce` and appear to be internal to Kokkos. The most basic reduce gives these hazards:
```
CODE:

double result;
Kokkos::parallel_reduce("simple_reduce", 1, KOKKOS_LAMBDA(int i, double& lsum) {
  lsum += i;
}, result);

OUTPUT:

========= COMPUTE-SANITIZER
========= Warning: Race reported between Read access at 0x570 in void Kokkos::Impl::cuda_parallel_launch_local_memory<Kokkos::Impl::ParallelReduce<Kokkos::Impl::CombinedFunctorReducer<main::[lambda(int, double &) (instance 1)], Kokkos::Impl::FunctorAnalysis<Kokkos::Impl::FunctorPatternInterface::REDUCE, Kokkos::RangePolicy<Kokkos::Cuda>, main::[lambda(int, double &) (instance 1)], double>::Reducer, void>, Kokkos::RangePolicy<Kokkos::Cuda>, Kokkos::Cuda, Kokkos::Cuda>>(T1)
=========     and Write access at 0x5a0 in void Kokkos::Impl::cuda_parallel_launch_local_memory<Kokkos::Impl::ParallelReduce<Kokkos::Impl::CombinedFunctorReducer<main::[lambda(int, double &) (instance 1)], Kokkos::Impl::FunctorAnalysis<Kokkos::Impl::FunctorPatternInterface::REDUCE, Kokkos::RangePolicy<Kokkos::Cuda>, main::[lambda(int, double &) (instance 1)], double>::Reducer, void>, Kokkos::RangePolicy<Kokkos::Cuda>, Kokkos::Cuda, Kokkos::Cuda>>(T1) [6272 hazards]
========= ....more lines of the same warning....
...
========= RACECHECK SUMMARY: 8 hazards displayed (0 errors, 8 warnings)
```
Adding `--racecheck-detect-level=error` removes these and the test cases pass.
```
========= COMPUTE-SANITIZER
...
========= RACECHECK SUMMARY: 0 hazards displayed (0 errors, 0 warnings)
```
Not sure if racecheck is going to turn out to be useful. These warning I don't think are useful, but maybe the errors are? 
- typo in test name: `santizer -> sanitizer`

### DIFF
--- a/components/eamxx/scripts/jenkins/jenkins_common_impl.sh
+++ b/components/eamxx/scripts/jenkins/jenkins_common_impl.sh
@@ -116,7 +116,7 @@ if [ $skip_testing -eq 0 ]; then
       fi
 
       if [[ "$SCREAM_MACHINE" == "weaver" ]]; then
-        ./scripts/gather-all-data "./scripts/test-all-scream -t cmc -t csr -t csi -t css ${TAS_ARGS}" -l -m $SCREAM_MACHINE
+        ./scripts/gather-all-data "./scripts/test-all-scream -t csm -t csr -t csi -t css ${TAS_ARGS}" -l -m $SCREAM_MACHINE
         if [[ $? != 0 ]]; then
           fails=$fails+1;
           memcheck_fail=1
--- a/components/eamxx/scripts/scripts-tests
+++ b/components/eamxx/scripts/scripts-tests
@@ -349,11 +349,12 @@ class TestTestAllScream(TestBaseOuter.TestBase):
             cmd = self.get_cmd("./test-all-scream -m $machine {}".format(options),
                                 self._machine, dry_run=False)
             run_cmd_assert_result(self, cmd, from_dir=TEST_DIR)
-            builddir = "cuda_mem_check" if is_cuda_machine(self._machine) else "valgrind"
+            builddir = "compute_sanitizer_memcheck" if is_cuda_machine(self._machine) else "valgrind"
             test_cmake_cache_contents(self, builddir, "CMAKE_BUILD_TYPE", "Debug")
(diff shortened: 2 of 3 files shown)


jgfouca (approved):
Thanks! This looks good. I will update the cdashboard tomorrow with the new build name

tcclevenger:
Thanks @ambrad I can let them know.

tcclevenger:
@ambrad racecheck warnings are expected. They pointed to an example from openmp https://github.com/kokkos/kokkos/issues/2355#issuecomment-534331799. Not sure if this means that these sanitizer checks will be unstable from Kokkos point of view. The memcheck and initcheck have already revealed issues for EAMxx though, so I'm thinking they are still useful.

tcclevenger:
@ambrad Actually, heard from someone else that while that behavior is expected for openmp, it is not expected for cuda. We should be passing sanitizer racecheck tests. The problem is in `parallel_reduce()` so outside of SCREAMs control, but we are working on a fix. Obviously that is in the development branch which SCREAM won't use that code anytime soon I'm guessing. I'll add a comment in some future SCREAM PR for why we are setting the detect level to warn.

PR #2383, opened 2023-06-14
test-all-scream: Fix nested quoting

jgfouca:
The extra layer of shell evaluation when running `taskset` was breaking some cmake_args.

Did a little extra work to ensure that the string quoting in the source code was consistent (" for outer quotes, ' for inner quotes).


tcclevenger (approved):
Thanks, Jim!

PR #2388, opened 2023-06-16
Add field at height diagnostic

bartgol:
Add equivalent of FieldAtPressureLevel, but for height rather than pressure coordinate.

### DIFF
--- a/components/eamxx/src/diagnostics/field_at_height.cpp
+++ b/components/eamxx/src/diagnostics/field_at_height.cpp
@@ -1,0 +1,164 @@
+#include "diagnostics/field_at_height.hpp"
+
+#include "ekat/std_meta/ekat_std_utils.hpp"
+#include "ekat/util/ekat_units.hpp"
+
+namespace
+{
+// Find first position in array pointed by [beg,end) that is below z
+// If all z's in array are >=z, return end
+template<typename T>
+KOKKOS_INLINE_FUNCTION
+const T* find_first_smaller_z (const T* beg, const T* end, const T& z)
+{
+  int count = end - beg;
+  if (count==1) {
+    return z < *beg ? beg : end;
+  } else {
+    auto mid = beg + count/2;
+    return z>= *mid ? find_first_smaller_z (mid,end,z)
+                    : find_first_smaller_z (beg,mid,z);
+  }
+}
+
+} // anonymous namespace
+
+namespace scream
+{
+
+// =========================================================================================
+FieldAtHeight::
+FieldAtHeight (const ekat::Comm& comm, const ekat::ParameterList& params)
+ : AtmosphereDiagnostic(comm,params)
+{
+  const auto& f = params.get<Field>("Field");
+  const auto& fid = f.get_header().get_identifier();
+  m_field_name = f.name();
+
+  // Sanity checks
+  using namespace ShortFieldTagsNames;
+  const auto& layout = fid.get_layout();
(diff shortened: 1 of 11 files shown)


components/eamxx/src/diagnostics/field_at_height.cpp
```diff
@@ -0,0 +1,164 @@
...
+template<typename T>
+KOKKOS_INLINE_FUNCTION
+const T* find_first_smaller_z (const T* beg, const T* end, const T& z)
+{
+  int count = end - beg;
+  if (count==1) {
+    return z < *beg ? beg : end;
+  } else {
+    auto mid = beg + count/2;
+    return z>= *mid ? find_first_smaller_z (mid,end,z)
```
ambrad, line 19: Recursion on the GPU is generally troublesome because the compiler can't know in advance how many registers to allocate.

components/eamxx/src/diagnostics/field_at_height.cpp
```diff
@@ -0,0 +1,164 @@
...
+          // We just extapolate *end
+          d_view(i) = *(--end);
+        } else {
+          auto pos = it-beg;
+          auto z0 = z_i(pos-1);
+          auto z1 = z_i(pos);
+          auto f0 = f_i(pos-1);
+          auto f1 = f_i(pos);
+
+          d_view(i) = ( (z_tgt-z0)*f1 + (z1-z_tgt)*f0 ) / (z1-z0);
```
ambrad, line 127: Is there a reason not to use EKAT's linear interpolation utility here? That would provide the benefit of a unit-tested routine.

bartgol: I thought about it, but given that there is one grid point, I thought the hierarchical ||ism was a bit of a waste, and since the data structures won't be reused, there was no advantage in reusing internal data computed by LinInterp.

PeterCaldwell:
I just had a big-picture thought about this PR (which shouldn't change the PR, but might change our interpretation): variables sampled very close to the surface (e.g. T_2m or U10) are within the Monin-Obukhov layer where quantities change exponentially instead of linearly. Thus using our new "output on height levels" capability introduced here might give different answers than the traditional implementation for these variables which sits somewhere in the coupler or surface models. Is this worth worrying about @bogensch and @golaz ?

components/eamxx/src/diagnostics/field_at_height.cpp
bartgol: I can turn this into a while loop, like in ekat's upper_bound (which, in case someone wondered, I could not use (without hacking it) b/c z is decreasing rather than increasing).

bartgol:
> I just had a big-picture thought about this PR (which shouldn't change the PR, but might change our interpretation): variables sampled very close to the surface (e.g. T_2m or U10) are within the Monin-Obukhov layer where quantities change exponentially instead of linearly. Thus using our new "output on height levels" capability introduced here might give different answers than the traditional implementation for these variables which sits somewhere in the coupler or surface models. Is this worth worrying about @bogensch and @golaz ?

Everywhere else in EAMxx (and homme), vertical integral use linear approximations, so I am not entirely sure this would be consistent. That said, it's a diagnostic output, so I guess we can do whatever extrapolation we want. I would perhaps defer to a second PR, in case you decide to go for exp extrapolation...

AaronDonahue:
@ambrad , would you mind taking a look at this now that I believe Luca addressed your comment?

tcclevenger (approved):

PR #2395, opened 2023-06-20
EAMxx: fix dependency calculation in atm DAG utility

bartgol:
The DAG we were saving to file was incorrect. In fact, it was not a DAG, since it contains loops, and self-dependencies. This nano PR fixes it.

### DIFF
--- a/components/eamxx/src/share/atm_process/atmosphere_process_dag.cpp
+++ b/components/eamxx/src/share/atm_process/atmosphere_process_dag.cpp
@@ -31,6 +31,11 @@ create_dag(const group_type& atm_procs)
   end_ts.id = m_nodes.size()-1;
   m_unmet_deps[end_ts.id].clear();
 
+  // Now all nodes are created, we need to create edges, by checking
+  // which is the first node (if any) that computes any of the input
+  // fields of each node
+  add_edges ();
+
   // Next, check if some unmet deps are simply coming from previous time step.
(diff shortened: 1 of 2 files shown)


PeterCaldwell:
Oh cool, I've been wanting this for a long time but couldn't justify prioritizing it. It's like Christmas! Thanks for this.

bartgol:
Tbc, the DAG does not yet show a full picture. E.g., it reports some inputs as "missing", if they are simply constant field read from IC file (no other atm proc computes them). But at least is not plain wrong (with deps that were completely non-existent).

PeterCaldwell:
Would it make sense to change "missing" to "read from file" or "from init" or something?

bartgol:
> Would it make sense to change "missing" to "read from file" or "from init" or something?

The DAG has no concept of "init", so it would be a bit like sweeping under the rug. What I have in mind would bring back the DAG also as a dependency check tool:

1. avoid our IC file read to error out if a field is not found.
2. create the DAG _after_ IC read
3. in the DAG, identify the fields inited (from IC file) by checking their time stamp (it will not be invalid)
4. mark as "missing" only the fields that a) have no previous atm proc that computes them (in the time step sequence), and b) have not yet been inited
5. crap out if there are missing fields, printing some useful message.

PeterCaldwell:
Yeah, your plan is way better than mine. Carry on :-)

PR #2400, opened 2023-06-21
Some minor fixes to the AtmosphereDriver

jeff-cohere:
@pbosler and I discovered a couple of things that needed adjusting in our MAM-related adventures. This PR has two minor fixes:

1. It's possible to call `finalize` on the AD before it has defined its process group, which was producing a segmentation fault by calling `finalize` on a null pointer. Now the AD checks to see whether it has a process group before attempting to finalize it.
2. <strike>The `fvphyshack`, which evidently does something special when the PG2 grid is used in tandem with RRTMGP, previously assumed that RRTMGP was enabled whenever a PG2 grid was found. Now, the AD determines whether RRTMGP is enabled and sets `fvphyshack` only if both PG2 and RRTMGP are present.</strike>The `fvphyshack`-related initialization to RRTMGP is performed only when RRTMGP is present.

### DIFF
--- a/components/eamxx/src/control/atmosphere_driver.cpp
+++ b/components/eamxx/src/control/atmosphere_driver.cpp
@@ -131,16 +131,9 @@ set_params(const ekat::ParameterList& atm_params)
     // To figure out whether we're using RRTMGP, we create a process group and
     // see whether RRTMGP is in it. This seems heavy-handed, but it reduceѕ code
     // duplication.
-    bool using_rrtmgp = false;
     auto& atm_proc_params = m_atm_params.sublist("atmosphere_processes");
     auto group = std::make_shared<AtmosphereProcessGroup>(m_atm_comm,atm_proc_params);
-    for (int i = 0; i < group->get_num_processes(); ++i) {
-      if (group->get_process(i)->name() == "rrtmgp") {
-        using_rrtmgp = true;
-        break;
-      }
-    }
-    if (using_rrtmgp) {
+    if (group->has_process("rrtmgp")) {
       fv_phys_rrtmgp_active_gases_init(m_atm_params);
     }
   }
--- a/components/eamxx/src/control/atmosphere_driver.cpp
+++ b/components/eamxx/src/control/atmosphere_driver.cpp
@@ -124,8 +124,24 @@ set_params(const ekat::ParameterList& atm_params)
 
   m_ad_status |= s_params_set;
 
+  // Apply the FV-RRTMGP hack if we're using a PG2 grid and RRTMGP.
   const auto pg_type = "PG2";
-  fvphyshack = m_atm_params.sublist("grids_manager").get<std::string>("physics_grid_type", "None") == pg_type;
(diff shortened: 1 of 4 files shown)


ambrad:
fvphyshack is needed if the physics pg2 grid is being used, which is essentially always in practice, independent of RRTMGP.

Likely the issue with RRTMGP has to do with handling a special input for RRTMGP, done here: https://github.com/E3SM-Project/scream/blob/e6c2838b826ac3041e7c289b8d916014b09dea5e/components/eamxx/src/share/util/eamxx_fv_phys_rrtmgp_active_gases_workaround.cpp.

My guess is the solution is not to call the  fv_phys_rrtmgp_active_gases_* routines if RRTMGP is not enabled.

I think the fvphyshack global is in that file due to some recent changes for building tests. But, again, conceptually that variable has nothing to do with RRTMGP except that RRTMGP has a special input it needs processed.

jeff-cohere:
Good to know! Thanks @ambrad . I'll rework that fix to do the check to skip the calls to RRTMGP if it's not present.

components/eamxx/src/control/atmosphere_driver.cpp
```diff
@@ -128,7 +128,21 @@ set_params(const ekat::ParameterList& atm_params)
...
     // See the [rrtmgp active gases] note in share/util/eamxx_fv_phys_rrtmgp_active_gases_workaround.hpp
-    fv_phys_rrtmgp_active_gases_init(m_atm_params);
+    // To figure out whether we're using RRTMGP, we create a process group and
+    // see whether RRTMGP is in it. This seems heavy-handed, but it reduceѕ code
+    // duplication.
+    bool using_rrtmgp = false;
+    auto& atm_proc_params = m_atm_params.sublist("atmosphere_processes");
+    auto group = std::make_shared<AtmosphereProcessGroup>(m_atm_comm,atm_proc_params);
+    for (int i = 0; i < group->get_num_processes(); ++i) {
+      if (group->get_process(i)->name() == "rrtmgp") {
```
bartgol, line 138: This assumes that there are no nested groups. In practice, we _do_ have nested groups. In particular, rrtmgp is part of the "physics" atm proc group, so this check will not detect rrtmgp.

I think the best way to deal with this through recursion. You can either wrap this check in a lambda, or add a method to the atm proc group class.

bartgol (changes requested):

ambrad:
We may be overcomplicating this. Is the problem simply that the PL getter sequence here:
https://github.com/E3SM-Project/scream/blob/e6c2838b826ac3041e7c289b8d916014b09dea5e/components/eamxx/src/share/util/eamxx_fv_phys_rrtmgp_active_gases_workaround.cpp#L8-L9
fails? If so, you might try to break it up into a layered sequence of queries and gets and return without doing anything if the rrtmgp->active_gases entry isn't there.

jeff-cohere:
It's true that I've been using the Principle of Least Knowledge to address this issue (I didn't even open that file, for example). If that fixes it, maybe that's the best solution.

jeff-cohere:
We can take or leave the `AtmosphereProcessGroup::has_process` method.

bartgol (approved):

PR #2415, opened 2023-06-30
Prefer HINTS to PATHS when looking for tpls

bartgol:
That's b/c HINTS is processed *before* system paths, while PATHS is processed *after* them. Hence, on systems that have a system-installed version, PATHS would always pick up the system one. While this is likely irrelevant for e3sm-shipped tpls like mct or csm_share, netcdf is sometimes found in the system lib/inc dirs on some machines, and it is usually not the right version (e.g., not the parallel one).


rljacob:
Should we do something similar in the E3SM cime/cmake build ?

jgfouca (approved):

jgfouca:
@rljacob , we aren't using find_library in E3SM's system. The E3SM build system is less flexible and more prescriptive when it comes to locating TPLs.

Issue #2387, opened 2023-06-15
Convert DP-SCREAM to C++


bogensch:
DP-SCREAM needs to be converted into C++ and integrated into v1.  This list is currently not complete and will be updated.

**v0 Routines that need to be converted to C++.**  All subroutines in all of these files need to be converted (to my knowledge, this list is complete).
1) components/eam/src/control/apply_iop_forcing.F90 
2) components/eam/src/dynamics/se/se_iop_intr_mod.F90
3) components/eam/src/control/iop_data_mod.F90
4) components/eam/src/control/history_iop.F90

**Pieces/snippets of code in v0 that needs to be incorporated into v1 AD**:
NEEDS TO BE ADDED/UPDATED

**Running v0 DP-SCREAM to generate baseline case/data:**
https://github.com/E3SM-Project/scmlib/wiki/Running-the-DP-SCREAM-Configuration
My recommendation is to use the DYCOMS-RF01 case to start with.  Alternatively, run the CIME smoke test:

**CIME smoke DP-SCREAM smoke test:**
SMS_R_Ln10.ne4_ne4.FDPSCREAM-ARM97

tcclevenger:
https://github.com/E3SM-Project/scream/pull/2587 merged. Added IOP class, DP unit test, and hooks to AD for reading IC file of arbitrary size (test is ne4) into grid for IOP case.

Next PR is for reading IOP data file and finish setting all initial conditions using this data (surface coupler, homme states, shoc grid size, among others).

bartgol:
@tcclevenger @bogensch I'm considering this done. Feel free to open a new issue in E3SM for the remaining work.

PR #2417, opened 2023-07-05
Generate all boilerplate for DP scream functions

jgfouca:
Subroutines were gathered based on #2387 

I suspect a lot of this generated code will need to be changed but it at least gives us something to start with.

Discussion points:
1) There are a number of static constants in the fortran (plev, pcnst, etc). I assume we want these to be runtime values in cxx so I treated them like function inputs.
2) The fortran files have extensive dependencies on the rest of eam which defeats the approach we used in p3 and shoc to add a file or two from eam to our build. This means we can't realistically call f90 implementations from scream for BFB unit testing without doing a full E3SM build of eam. This could also complicate the conversion since there's potentially a lot more functionality to convert beyond the dozen or so functions.
3) The iop_data_mod module has lots of data. This will probably require an additional new struct/class in Cxx.
4) What's the story with MODEL_THETA_L? Do we need to preserve this in Cxx?
5) There appear to be a number of dimensions involved (nelemd, plev, pcnst, np). Do we want to follow the same pattern of p3/shoc and have a monolithic kernel that launches a parallel_for over one of these dimensions (I assume nelemd)?
6) Which of fortran files is considered the "entry point" to DP? In other words, the equivalent of p3_main?

Made a progress table here to track things:
https://acme-climate.atlassian.net/wiki/spaces/EAMXX/pages/3736895503/2023.Q2+C+DP-SCREAM


components/eamxx/src/physics/dp/impl/dp_apply_iop_forcing_impl.hpp
```diff
@@ -0,0 +1,25 @@
...
+namespace dp {
+
+/*
+ * Implementation of dp apply_iop_forcing. Clients should NOT
+ * #include this file, but include dp_functions.hpp instead.
+ */
+
+template<typename S, typename D>
+KOKKOS_FUNCTION
+void Functions<S,D>::apply_iop_forcing(const Int& nelemd, const uview_1d<element_t>& elem, hvcoord_t& hvcoord, const hybrid_t& hybrid, const timelevel_t& tl, const Int& n, const bool& t_before_advance, const Int& nets, const Int& nete)
```
tcclevenger, line 16: In F90, this function (and some others) live in the `eam/src/dynamics/` folder with calls to homme. Not sure exactly how the hooks to homme work in eamxx, but will this need to live in the `eamxx/src/dynamics` subdirectory instead of dp?

tcclevenger (commented):
For 2., it also looks like there are homme/ dependencies in the `se_iop_intr_mod.F90`. Are all the eam dependencies contained in that file as well? It seems that the routines in that file acts as a sort of an interface between `eam/src/control/*iop*.F90` routines and the rest of the model, and almost exclusively called around calls to dynamics. Maybe the routines in that file make more sense in `eamxx/src/dynamics`? 

Then, all the routines from `eam/src/control/*iop*.F90` files might be able to have traditional bfb tests? But I am a bit out of my depth right now with exactly how dp works..

I can approve if would be helpful for us to start porting simpler functions now, or we can wait on @bogensch to answer some of the structure questions.

components/eamxx/src/physics/dp/dp_functions_f90.cpp
```diff
@@ -0,0 +1,308 @@
...
+{
+#if 0
+  using PF = Functions<Real, DefaultDevice>;
+
+  using Spack   = typename PF::Spack;
+
+  Kokkos::parallel_for(1, KOKKOS_LAMBDA(const Int&) {
+    PF::iop_broadcast();
+  });
+#endif
```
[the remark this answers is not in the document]
jgfouca: These are auto-generated implementations that are not guaranteed to work.

components/eamxx/src/physics/dp/impl/dp_apply_iop_forcing_impl.hpp
jgfouca: Yeah, I expect some of this stuff will need to be moved but I'm not worried aboutt that right now.

tcclevenger (approved):

bogensch:
Thanks @jgfouca this looks good as a start; as you said will probably need to be modified as conversion takes place.  My replies to some of your questions:

4.  anything in the MODEL_THETA_L is what we want to convert since that's SCREAM's default dycore.  Anything outside of these dependencies we will not need to convert since EAMxx doesn't know about those dycores (and we don't need to preserve the ifdef MODEL_THETA_L statements).
6.  While there is no one entry point the se_iop_intr_mod.F90 serves as the main interface of DP with the rest of the model and iop_data_mod.F90 serves as the main interface with the data needed to be used by DP.

bogensch (approved):

jgfouca:
Thanks, @bogensch !

tcclevenger (approved):

Issue #2372, opened 2023-06-08
Convert qneg4 routine to C++


bogensch:
The qneg4 routine in the v0 Fortran code (located at components/eam/src/physics/cam/qneg4.F90) needs to be converted to c++.  This code block checks to see if moisture flux into the ground (negative value of latent heat flux) exceeds the total moisture content of the lowest model layer.  Currently, v1 produces negative values of water vapor near the surface (primarily during model spin-up) and this will provide a patch for that in an energy conserving manner.  

Lines 64 through 77 represent the essential code that needs to be converted.  The rest of the code is for bookkeeping purposes to keep track of the number incidences of clipping (not sure if we want to convert that portion).

In v1 this block of code should be called in the macrophysics (SHOC) interface directly before shoc_main is called.

https://github.com/E3SM-Project/scream/blob/1084bdc66d2f60855de49420226580405a56673c/components/eam/src/physics/cam/qneg4.F90#L58-L77

oksanaguba:
A couple of suggestions: IIRC the code in qneg4 creates a leak is there is not enough mass to compensate for negative CFLX. There is no reason not to try to compensate for negative CFLX in the first few bottom levels, or even the whole column to make differences smaller. Also, let's not change SHFLX in this code is it is done above, this just creates yet another energy leak.

I wrote a simple substitute for qneg4 as below for the whole column, but to have it. I did not try to experiment using only a few bottom levels, Peter B or Peter C or anyone else will know if we should use just a subset of levels and how many. My simple code is below

```
do i=1,ncol
  cc = cam_in%cflx(i,1)*ztodt*gravit
  if (cc < 0.0_r8) then
    cc=abs(cc)
    mm=0.0_r8
    do k=1,pver
      mm = mm + state%q(i,k,1)*state%pdel(i,k)
    enddo
    cam_in%cflx(i,1) = 0.0_r8
    if (mm < cc) then
      state%q(i,1:pver,1) = 0.0_r8 !allow leak
    else
      do k=1,pver
        qq = state%q(i,k,1)
        pp = state%pdel(i,k)
        adjust = cc *  qq*pp/mm
        state%q(i,k,1) = (qq*pp - adjust)/pp
      enddo
    endif
!print *, "OG cc, mm, new mm",cc,mm,sum(state%q(i,:,1)*state%pdel(i,:))
!print *, 'OG before, after',mm-cc, sum(state%q(i,:,1)*state%pdel(i,:))
  endif
enddo
```

[...]

PR #2418, opened 2023-07-07
Add check_flux_state_consistency to shoc interface

tcclevenger:
New `check_flux_state_consistency()` from @oksanaguba suggestion https://github.com/E3SM-Project/scream/issues/2372#issuecomment-1583832926. Called directly before `shoc_main()`.

### DIFF
--- a/components/eamxx/src/physics/shoc/eamxx_shoc_process_interface.cpp
+++ b/components/eamxx/src/physics/shoc/eamxx_shoc_process_interface.cpp
@@ -469,14 +469,17 @@ void SHOCMacrophysics::check_flux_state_consistency(const double dt)
 {
   using PC = scream::physics::Constants<Real>;
   const Real gravit = PC::gravit;
+  const Real qmin   = 1e-12; // minimum permitted constituent concentration (kg/kg)
 
   const auto& pseudo_density = get_field_in ("pseudo_density").get_view<const Spack**>();
   const auto& surf_evap      = get_field_out("surf_evap").get_view<Real*>();
   const auto& qv             = get_field_out("qv").get_view<Spack**>();
 
-  const auto nlevs      = m_num_levs;
-  const auto nlev_packs = ekat::npack<Spack>(nlevs);
-  const auto policy     = ekat::ExeSpaceUtils<KT::ExeSpace>::get_default_team_policy(m_num_cols, nlev_packs);
+  const auto nlevs           = m_num_levs;
+  const auto nlev_packs      = ekat::npack<Spack>(nlevs);
+  const auto last_pack_idx   = (nlevs-1)/Spack::n;
+  const auto last_pack_entry = (nlevs-1)%Spack::n;
+  const auto policy          = ekat::ExeSpaceUtils<KT::ExeSpace>::get_default_team_policy(m_num_cols, nlev_packs);
   Kokkos::parallel_for("check_flux_state_consistency",
                        policy,
                        KOKKOS_LAMBDA (const KT::MemberType& team) {
@@ -485,15 +488,21 @@ void SHOCMacrophysics::check_flux_state_consistency(const double dt)
     const auto& pseudo_density_i = ekat::subview(pseudo_density, i);
     const auto& qv_i             = ekat::subview(qv, i);
 
-    if (surf_evap(i) < 0) {
+    // reciprocal of pseudo_density at the bottom layer
+    const auto rpdel = 1.0/pseudo_density_i(last_pack_idx)[last_pack_entry];
+
+    // Check if the negative surface latent heat flux can exhaust
+    // the moisture in the lowest model level. If so, apply fixer.
+    const auto condition = surf_evap(i) - (qmin - qv_i(last_pack_idx)[last_pack_entry])/(dt*gravit*rpdel);
+    if (condition < 0) {
       const auto cc = abs(surf_evap(i)*dt*gravit);
 
-      auto qv_x_pdel = [&](const int k) {
+      auto tracer_mass = [&](const int k) {
         return qv_i(k)*pseudo_density_i(k);
       };
-      Real mm = ekat::ExeSpaceUtils<KT::ExeSpace>::view_reduction(team, 0, nlevs, qv_x_pdel);
+      Real mm = ekat::ExeSpaceUtils<KT::ExeSpace>::view_reduction(team, 0, nlevs, tracer_mass);
 
-      EKAT_KERNEL_ASSERT_MSG(mm >= cc, "Error! Accumulated column moisture should be greater than surf_evap.\n");
+      EKAT_KERNEL_ASSERT_MSG(mm >= cc, "Error! Total mass of column vapor should be greater than mass of surf_evap.\n");
 
       Kokkos::parallel_for(Kokkos::TeamVectorRange(team, nlev_packs), [&](const int& k) {
         const auto adjust = cc*qv_i(k)*pseudo_density_i(k)/mm;
--- a/components/eamxx/src/physics/shoc/eamxx_shoc_process_interface.cpp
+++ b/components/eamxx/src/physics/shoc/eamxx_shoc_process_interface.cpp
@@ -64,13 +64,13 @@ void SHOCMacrophysics::set_grids(const std::shared_ptr<const GridsManager> grids
   const auto s2 = s*s;
 
   // These variables are needed by the interface, but not actually passed to shoc_main.
-  add_field<Required>("omega",            scalar3d_layout_mid,  Pa/s,    grid_name, ps);
-  add_field<Required>("surf_sens_flux",   scalar2d_layout_col,  W/m2,    grid_name);
-  add_field<Required>("surf_evap",        scalar2d_layout_col,  kg/m2/s, grid_name);
-  add_field<Required>("surf_mom_flux",    surf_mom_flux_layout, N/m2,    grid_name);
+  add_field<Required>("omega",          scalar3d_layout_mid,  Pa/s, grid_name, ps);
+  add_field<Required>("surf_sens_flux", scalar2d_layout_col,  W/m2, grid_name);
+  add_field<Required>("surf_mom_flux",  surf_mom_flux_layout, N/m2, grid_name);
 
-  add_field<Updated> ("T_mid",            scalar3d_layout_mid, K,       grid_name, ps);
-  add_field<Updated> ("qv",               scalar3d_layout_mid, Qunit,   grid_name, "tracers", ps);
+  add_field<Updated>("surf_evap", scalar2d_layout_col, kg/m2/s, grid_name);
+  add_field<Updated> ("T_mid",    scalar3d_layout_mid, K,       grid_name, ps);
+  add_field<Updated> ("qv",       scalar3d_layout_mid, Qunit,   grid_name, "tracers", ps);
 
   // Input variables
   add_field<Required>("p_mid",          scalar3d_layout_mid, Pa,    grid_name, ps);
@@ -91,7 +91,7 @@ void SHOCMacrophysics::set_grids(const std::shared_ptr<const GridsManager> grids
   add_field<Computed>("inv_qc_relvar", scalar3d_layout_mid, Qunit*Qunit, grid_name, ps);
 
   // Tracer group
-  add_group<Updated>("tracers",grid_name,ps,Bundling::Required);
+  add_group<Updated>("tracers", grid_name, ps, Bundling::Required);
 
   // Boundary flux fields for energy and mass conservation checks
   if (has_column_conservation_check()) {
@@ -371,7 +371,7 @@ void SHOCMacrophysics::initialize_impl (const RunType run_type)
     const auto& ice_flux   = get_field_out("ice_flux").get_view<Real*>();
     const auto& heat_flux  = get_field_out("heat_flux").get_view<Real*>();
     shoc_postprocess.set_mass_and_energy_fluxes (surf_evap, surf_sens_flux,
-      					         vapor_flux, water_flux,
+                                                 vapor_flux, water_flux,
                                                  ice_flux, heat_flux);
   }
 
@@ -432,6 +432,10 @@ void SHOCMacrophysics::run_impl (const double dt)
                        shoc_preprocess);
   Kokkos::fence();
 
+  if (m_params.get<bool>("check_flux_state_consistency", false)) {
+    check_flux_state_consistency(dt);
+  }
+
   // For now set the host timestep to the shoc timestep. This forces
   // number of SHOC timesteps (nadv) to be 1.
   // TODO: input parameter?
@@ -461,5 +465,44 @@ void SHOCMacrophysics::finalize_impl()
   // Do nothing
 }
 // =========================================================================================
+void SHOCMacrophysics::check_flux_state_consistency(const double dt)
+{
+  using PC = scream::physics::Constants<Real>;
+  const Real gravit = PC::gravit;
 
+  const auto& pseudo_density = get_field_in ("pseudo_density").get_view<const Spack**>();
+  const auto& surf_evap      = get_field_out("surf_evap").get_view<Real*>();
+  const auto& qv             = get_field_out("qv").get_view<Spack**>();
+
+  const auto nlevs      = m_num_levs;
+  const auto nlev_packs = ekat::npack<Spack>(nlevs);
+  const auto policy     = ekat::ExeSpaceUtils<KT::ExeSpace>::get_default_team_policy(m_num_cols, nlev_packs);
+  Kokkos::parallel_for("check_flux_state_consistency",
+                       policy,
+                       KOKKOS_LAMBDA (const KT::MemberType& team) {
+    const auto i = team.league_rank();
+
+    const auto& pseudo_density_i = ekat::subview(pseudo_density, i);
+    const auto& qv_i             = ekat::subview(qv, i);
+
+    if (surf_evap(i) < 0) {
+      const auto cc = abs(surf_evap(i)*dt*gravit);
+
+      auto qv_x_pdel = [&](const int k) {
+        return qv_i(k)*pseudo_density_i(k);
+      };
+      Real mm = ekat::ExeSpaceUtils<KT::ExeSpace>::view_reduction(team, 0, nlevs, qv_x_pdel);
+
+      EKAT_KERNEL_ASSERT_MSG(mm >= cc, "Error! Accumulated column moisture should be greater than surf_evap.\n");
+
+      Kokkos::parallel_for(Kokkos::TeamVectorRange(team, nlev_packs), [&](const int& k) {
+        const auto adjust = cc*qv_i(k)*pseudo_density_i(k)/mm;
+        qv_i(k) = (qv_i(k)*pseudo_density_i(k) - adjust)/pseudo_density_i(k);
+      });
+
+      surf_evap(i) = 0;
+    }
+  });
+}
+// =========================================================================================
 } // namespace scream
--- a/components/eamxx/cime_config/namelist_defaults_scream.xml
+++ b/components/eamxx/cime_config/namelist_defaults_scream.xml
@@ -204,7 +204,7 @@ be lost if SCREAM_HACK_XML is not enabled.
     <!-- SHOC macrophysics -->
     <shoc inherit="atm_proc_base">
       <enable_column_conservation_checks>false</enable_column_conservation_checks>
-      <check_flux_state_consistency>false</check_flux_state_consistency>
+      <check_flux_state_consistency>true</check_flux_state_consistency>
     </shoc>
 
     <!-- CLD fraction -->
--- a/components/eamxx/cime_config/namelist_defaults_scream.xml
+++ b/components/eamxx/cime_config/namelist_defaults_scream.xml
@@ -204,6 +204,7 @@ be lost if SCREAM_HACK_XML is not enabled.
     <!-- SHOC macrophysics -->
     <shoc inherit="atm_proc_base">
       <enable_column_conservation_checks>false</enable_column_conservation_checks>
+      <check_flux_state_consistency>false</check_flux_state_consistency>
     </shoc>
 
     <!-- CLD fraction -->
--- a/components/eamxx/src/physics/shoc/eamxx_shoc_process_interface.hpp
+++ b/components/eamxx/src/physics/shoc/eamxx_shoc_process_interface.hpp
@@ -102,7 +102,7 @@ public:
         // to tracer group in postprocessing.
         // TODO: remove *_copy views once SHOC can request a subset of tracers.
         tke_copy(i,k) = tke(i,k);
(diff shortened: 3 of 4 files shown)


components/eamxx/src/physics/shoc/eamxx_shoc_process_interface.cpp
```diff
@@ -432,6 +432,10 @@ void SHOCMacrophysics::run_impl (const double dt)
                        shoc_preprocess);
   Kokkos::fence();
 
+  if (m_params.get<bool>("check_flux_state_consistency", false)) {
```
tcclevenger, line 435: Should this be called always? Or toggled on and off via input.yaml?

components/eamxx/src/physics/shoc/eamxx_shoc_process_interface.cpp
```diff
@@ -461,5 +465,44 @@ void SHOCMacrophysics::finalize_impl()
...
+
+      Kokkos::parallel_for(Kokkos::TeamVectorRange(team, nlev_packs), [&](const int& k) {
+        const auto adjust = cc*qv_i(k)*pseudo_density_i(k)/mm;
+        qv_i(k) = (qv_i(k)*pseudo_density_i(k) - adjust)/pseudo_density_i(k);
+      });
+
+      surf_evap(i) = 0;
+    }
+  });
+}
```
tcclevenger, line 506: Questions about this function:
1. What is the significance of `surf_evap < 0`? I want to add a comment.
2. What does `surf_evap*dt*gravit` represent? Would like to use a more descriptive name.
3. Same as above for `qv*pseudo_density`.
4. For `mm>=cc` does my error message make sense? Is there something better to say?

tcclevenger (commented):
I've left some questions I have on the implementation. An overarching question: How should I test this is improving simulations?

bogensch:
Thanks for working on this @tcclevenger .  I took a look at your code and the fortran snippet that Oksana added in issue #2372.  I have big concerns with one of the lines here.

`do i=1,ncol
  cc = cam_in%cflx(i,1)*ztodt*gravit
  if (cc < 0.0_r8) then
    cc=abs(cc)
    mm=0.0_r8
    do k=1,pver
      mm = mm + state%q(i,k,1)*state%pdel(i,k)
    enddo
    cam_in%cflx(i,1) = 0.0_r8
    if (mm < cc) then
      state%q(i,1:pver,1) = 0.0_r8 !allow leak
    else
      do k=1,pver
        qq = state%q(i,k,1)
        pp = state%pdel(i,k)
        adjust = cc *  qq*pp/mm
        state%q(i,k,1) = (qq*pp - adjust)/pp
      enddo
    endif
!print *, "OG cc, mm, new mm",cc,mm,sum(state%q(i,:,1)*state%pdel(i,:))
!print *, 'OG before, after',mm-cc, sum(state%q(i,:,1)*state%pdel(i,:))
  endif
enddo`

It seems as if we are applying adjustment anytime the surface latent flux is less than zero (i.e. anytime cc < 0)?  Is that true @oksanaguba?  If so then I do not agree with this as cc < 0 (surface latent flux less than zero) is something that can and does happen quite often in nature and the model and SCREAM can handle this gracefully in 99.999999% of `instances...` However there are a few instances (primarily during spin-up) when the negative surface latent heat flux can exhaust the moisture in the lowest model level.  It is only in these outlier cases where the fixer/adjust should be applied.  Therefore I propose replacing line 

```
if (cc < 0.0_r8) then
```

with a trigger something in the spirit in qneg4.F90

`       excess(i) = qflx(i,1) - (qmin(1) - qbot(i,1))/(ztodt*gravit*srfrpdel(i)) 
 ! 
       if (excess(i) < 0._r8) then `

Where it checks to see if the bottom layer vapor will be exhausted or not.  If it is, then carry on with Oksana's fixer.  Else, if we do this whenever cflx < 0 if we will be overly applying this adjustment and I anticipate that simulation quality will be deteriorated.

Apologies if I've misinterpreted the c++ code or oksana's code.

components/eamxx/src/physics/shoc/eamxx_shoc_process_interface.cpp
oksanaguba: 1) regarding surf_evap < 0 -- it means mass of vapor in the bottom layer will be reduced. i heard arguments that it is due to dew formation. you can say "surf_evap<0 may drive Q1 to negative values".

2) surf_evap units are kg/m^2/sec, so surf_evap*dt is change of mass in kg/m^2. since for tracers we often operate with quantities Q1*dp, which are in units of [dp]=kg/m/sec^2, the conversion between Pa and kg/m2 is 1/g factor. in other words, [surf_evap*dt*gravit] = Pa.

3) similarly, qv*pseudo_density is tracer 'mass' in Pa.

4) you can leave the message as is, or say ""Error! Total mass of column vapor should be greater than mass of QFLX.\n""

components/eamxx/src/physics/shoc/eamxx_shoc_process_interface.cpp
oksanaguba: i would keep on/off switch if you already implemented it.

oksanaguba (approved):

oksanaguba:
Replying to Peter's @bogensch comment:
If i understood it correctly, Peter suggests to not redistribute QFLX<0 values along the whole column, but to instead redistribute only a leftover that would not drive Q1 to less than Q1_min value. I have no opinion (as i asked about my crude code before, i do not know the best way to do it), but i can redo my F code following this suggestion to make it easier for Conrad to change his.

Regarding how often QFLX<0 happens, while i don't think it is relevant to how we need to do the limiter, I think it happens often enough: check Figure 5 panel b for averaged QFLX, it seems there is enough of light green color in the plot (which indicates QFLX<0).

thanks!

components/eamxx/src/physics/shoc/eamxx_shoc_process_interface.cpp
```diff
@@ -461,5 +465,44 @@ void SHOCMacrophysics::finalize_impl()
...
+  const auto policy     = ekat::ExeSpaceUtils<KT::ExeSpace>::get_default_team_policy(m_num_cols, nlev_packs);
+  Kokkos::parallel_for("check_flux_state_consistency",
+                       policy,
+                       KOKKOS_LAMBDA (const KT::MemberType& team) {
+    const auto i = team.league_rank();
+
+    const auto& pseudo_density_i = ekat::subview(pseudo_density, i);
+    const auto& qv_i             = ekat::subview(qv, i);
 
+    if (surf_evap(i) < 0) {
```
bogensch, line 488: It seems as if we are applying adjustment anytime the surface latent flux is less than zero (i.e. anytime cc < 0)?  Is that true @oksanaguba?  If so then I do not agree with this as cc < 0 (surface latent flux less than zero) is something that can and does happen quite often in nature and the model and SCREAM can handle this gracefully in 99.999999% of instances... However there are a few instances (primarily during spin-up) when the negative surface latent heat flux can exhaust the moisture in the lowest model level.  It is only in these outlier cases where the fixer/adjust should be applied.  Therefore I propose replacing line 

```
if (cc < 0.0_r8) then
```

with a trigger something in the spirit in qneg4.F90

```
 excess(i) = qflx(i,1) - (qmin(1) - qbot(i,1))/(ztodt*gravit*srfrpdel(i)) 
  
 if (excess(i) < 0._r8) then 
```

Where it checks to see if the bottom layer vapor will be exhausted or not.  If it is, then carry on with Oksana's fixer.  Else, if we do this whenever cflx < 0 if we will be overly applying this adjustment and I anticipate that simulation quality will be deteriorated.

bogensch (changes requested):

bogensch:
Thanks @oksanaguba but my issue is that as written we are calling this adjustment WAY more often than it is necessary and will likely have detrimental impacts on our simulation.  This does not need to be triggered/called every instance the QFLX < 0 as the vast majority of the time no adjustment needs to be done.  See my comment above.  All I suggest is that the trigger be revised to be something similar to what qneg4 has (only call the adjustment when the application of the surface latent heat flux will exhaust/exceed all vapor at the surface).  So just a one/two line change and the rest of the fixer I don't have a problem with.

bogensch:
What I propose (using Oskana's code from issue #2372 as a basis)

```
do i=1,ncol
  trigger = cam_in%cflx(i,1) - (qmin(1) - state%q(i,pver,1))/(ztodt*gravit*state%rpdel(1,pver))
  if (trigger < 0.0_r8) then
    cc = cam_in%cflx(i,1)*ztodt*gravit
    cc=abs(cc)
    mm=0.0_r8
    do k=1,pver
      mm = mm + state%q(i,k,1)*state%pdel(i,k)
    enddo
    cam_in%cflx(i,1) = 0.0_r8
    if (mm < cc) then
      state%q(i,1:pver,1) = 0.0_r8 !allow leak
    else
      do k=1,pver
        qq = state%q(i,k,1)
        pp = state%pdel(i,k)
        adjust = cc *  qq*pp/mm
        state%q(i,k,1) = (qq*pp - adjust)/pp
      enddo
    endif
!print *, "OG cc, mm, new mm",cc,mm,sum(state%q(i,:,1)*state%pdel(i,:))
!print *, 'OG before, after',mm-cc, sum(state%q(i,:,1)*state%pdel(i,:))
  endif
enddo
```

Note that only the first three lines within the ncol loop has changed.  Everything else stays the same.

oksanaguba:
thanks, @bogensch , i believe i understood your concern the first time :)

however, i have a follow up question -- qneg4 in eam is called at the beginning of tphysAC, which implies that Q1 does not change between the call to qneg4 and the call to clubb/shoc in tphysBC. This is probably not correct -- between qneg4 and clubb there are at least GWD and ZM that could potentially change Q1 at the bottom layer. Also, clubb itself is called in an iteration. Do i understand it right that clubb as turbulence model redistributes mass in the lowest levels? So that each iteration of clubb can reduce the amount of Q1 at the bottom and potentially lead to violation of condition mass>0 if cflx remains <0 even after the limiter?

One can say GWD and ZM issues are not applied to scream as we can call the new limiter right before clubb, but would the iteration issue apply to scream anyway? There are a few ways to deal with this then.

bogensch:
@oksanaguba thanks and my apologies for misinterpreting your previous message (mondays!).  

Yes, SHOC/CLUBB redistribute mass in the vertical.  Ideally the call to qneg4 (or comparable fixer) should be directly before SHOC/CLUBB are called because this is where the surface fluxes are applied (for the exact reasons you state).  Yes it is true that CLUBB is iterated in EAM (and SHOC has the capability in EAMxx to be iterated, but not by default at 3 km since the SHOC time step is equal to the physics time step).  In EAM, the most consistent/safest way may be to call qneg4 for each step of the macro_micro substep loop, at the beginning of this loop (using the sub stepped time step).  I think calling it once before the macro_micro substep loop could be a good enough approximation (using the physics time step)... though it may lead to the fixer being called a few more instances than it needs to be.

tcclevenger:
@bogensch I'm a little confused exactly where this should be called. Is it still directly before SHOC main code, or elsewhere? What is the `macro_micro` loop? In eamxx we have something called `mac_aero_mic` which consists of `shoc,cldFraction,spa,p3`, I think in that order.

bogensch:
@tcclevenger sorry for the confusion.  I think placing the call in the SHOC interface before shoc_main is called is the best place.

AaronDonahue:
The FAIL's I see in the AT are DIFF's in `mappy`.  This is a non-BFB PR right so this is expected?  

I think this PR is ready to merge given a final round of reviews - @bogensch and @oksanaguba ?

tcclevenger:
@AaronDonahue I'm currently running simulation on Summit to see what difference is made to the energy errors. I'll update here with what I find.

tcclevenger:
@oksanaguba @bogensch Running `COMPSET=F2010-SCREAMv1` at `RES=ne30pg2 for 6 months, both master and this branch exhibit roughly the same magnitude energy errors following SHOC (around ~1e-9 relative error). I have verified that the fixer is actually being used. Is this what we expect? Is there a different config I should run that is producing higher errors?

tcclevenger:
Thanks, @bogensch. @bogensch and @oksanaguba, let me know if there are any other comments you have (and I'll need an approval). @AaronDonahue I'm going to re-run the AT since last was over 2 weeks ago and I can no longer check diffs. Then we can probably merge when that finishes.

tcclevenger:
@AaronDonahue Only diffs, no nans or massively different values (rel errs ~1e-8 to ~1e0).

bogensch (approved):
Thanks for your efforts on this @tcclevenger.  This looks good to me.

components/eamxx/src/physics/shoc/eamxx_shoc_process_interface.cpp
```diff
@@ -461,5 +465,53 @@ void SHOCMacrophysics::finalize_impl()
...
+    const auto& qv_i             = ekat::subview(qv, i);
+
+    // reciprocal of pseudo_density at the bottom layer
+    const auto rpdel = 1.0/pseudo_density_i(last_pack_idx)[last_pack_entry];
+
+    // Check if the negative surface latent heat flux can exhaust
+    // the moisture in the lowest model level. If so, apply fixer.
+    const auto condition = surf_evap(i) - (qmin - qv_i(last_pack_idx)[last_pack_entry])/(dt*gravit*rpdel);
+    if (condition < 0) {
+      const auto cc = abs(surf_evap(i)*dt*gravit);
```
ambrad, line 498: `abs` rather than `std::abs` potentially calls C's `int abs(int)` rather than, essentially, `fabs`. You might check if that's happening here. For me, this reproducer shows it's something to watch out for:
```
$ cat abs.cpp
#include <iostream>
int main () { std::cout << abs(-3.14) << "\n"; }
$ g++ -std=c++14 abs.cpp; if [ $? == 0 ]; then ./a.out; fi
3
$ cat abs.cpp
#include <iostream>
int main () { std::cout << std::abs(-3.14) << "\n"; }
$ g++ -std=c++14 abs.cpp; if [ $? == 0 ]; then ./a.out; fi
3.14
```

If I'm right about this, here are some others I'm seeing:
```
./tests/uncoupled/rrtmgp/rrtmgp_standalone_unit.cpp:209:            p_del(icol,ilay) = abs(p_lev(icol,ilay+1) - p_lev(icol,ilay));
./src/physics/nudging/tests/nudging_tests.cpp:312:          REQUIRE(abs(T_mid_v_h_o(icol,ilev) - val_tim_avg)<0.001);
./src/physics/nudging/tests/nudging_tests.cpp:313:          REQUIRE(abs(qv_h_o(icol,ilev) - val_tim_avg)<0.001);
./src/physics/nudging/tests/nudging_tests.cpp:314:          REQUIRE(abs(u_h_o(icol,ilev) - val_tim_avg)<0.001);
./src/physics/nudging/tests/nudging_tests.cpp:315:          REQUIRE(abs(v_h_o(icol,ilev) - val_tim_avg)<0.001);
./src/physics/nudging/tests/nudging_tests.cpp:328:          REQUIRE(abs(T_mid_v_h_o(icol,ilev) - val_tim_avg)<0.001);
./src/physics/nudging/tests/nudging_tests.cpp:329:          REQUIRE(abs(qv_h_o(icol,ilev) - val_tim_avg)<0.001);
./src/physics/nudging/tests/nudging_tests.cpp:330:          REQUIRE(abs(u_h_o(icol,ilev) - val_tim_avg)<0.001);
./src/physics/nudging/tests/nudging_tests.cpp:331:          REQUIRE(abs(v_h_o(icol,ilev) - val_tim_avg)<0.001);
./src/physics/nudging/tests/nudging_tests.cpp:348:  REQUIRE(abs(T_mid_v_h_o(icol,ilev) - val_avg)<0.001);
./src/physics/nudging/tests/nudging_tests.cpp:349:  REQUIRE(abs(qv_h_o(icol,ilev) - val_avg)<0.001);
./src/physics/nudging/tests/nudging_tests.cpp:350:  REQUIRE(abs(u_h_o(icol,ilev) - val_avg)<0.001);
./src/physics/nudging/tests/nudging_tests.cpp:351:  REQUIRE(abs(v_h_o(icol,ilev) - val_avg)<0.001);
./src/physics/p3/impl/p3_calc_rime_density_impl.hpp:48:      Spack V_impact = abs(vtrmi1-Vt_qc);
./src/physics/shoc/tests/shoc_pdf_qw_parameters_tests.cpp:88:      REQUIRE(abs(SDS.w1_1) > abs(SDS.w1_2));
./src/physics/shoc/tests/shoc_pdf_qw_parameters_tests.cpp:92:      REQUIRE(abs(SDS.w1_1) < abs(SDS.w1_2));
./src/physics/shoc/tests/shoc_pdf_qw_parameters_tests.cpp:96:      REQUIRE(abs(SDS.w1_1) == abs(SDS.w1_2));
./src/physics/shoc/tests/shoc_pdf_qw_parameters_tests.cpp:104:    Real qwgaus_diff_result1 = abs(qvconv*SDS.qw1_2 - qvconv*SDS.qw1_1);
./src/physics/shoc/tests/shoc_pdf_qw_parameters_tests.cpp:113:    Real qwgaus_diff_result2 = abs(qvconv*SDS.qw1_2 - qvconv*SDS.qw1_1);
./src/physics/shoc/tests/shoc_diag_third_tests.cpp:159:        REQUIRE(abs(SDS.w3[offset]) < 1);
./src/physics/shoc/tests/shoc_energy_integral_tests.cpp:102:          REQUIRE(abs(SDS.u_wind[offsets]) > abs(SDS.u_wind[offset]));
./src/physics/shoc/tests/shoc_energy_integral_tests.cpp:103:          REQUIRE(abs(SDS.v_wind[offsets]) > abs(SDS.v_wind[offset]));
./src/physics/shoc/tests/shoc_compute_diag_third_tests.cpp:166:        REQUIRE(abs(SDS.w3[offset]) < 10);
./src/physics/shoc/tests/shoc_clip_third_moms_tests.cpp:76:        if (abs(SDS.w3[offset]) > 1000){
./src/physics/shoc/tests/shoc_clip_third_moms_tests.cpp:94:        if (abs(w3_in[n]) > 1000){
./src/physics/shoc/tests/shoc_clip_third_moms_tests.cpp:95:          REQUIRE(abs(SDS.w3[offset]) < abs(w3_in[n]));
./src/physics/shoc/tests/shoc_pdf_thl_parameters_tests.cpp:88:      REQUIRE(abs(SDS.w1_1) > abs(SDS.w1_2));
./src/physics/shoc/tests/shoc_pdf_thl_parameters_tests.cpp:92:      REQUIRE(abs(SDS.w1_1) < abs(SDS.w1_2));
./src/physics/shoc/tests/shoc_pdf_thl_parameters_tests.cpp:96:      REQUIRE(abs(SDS.w1_1) == abs(SDS.w1_2));
./src/physics/shoc/tests/shoc_pdf_thl_parameters_tests.cpp:151:    Real thlgaus_diff_result1 = abs(SDS.thl1_2 - SDS.thl1_1);
./src/physics/shoc/tests/shoc_pdf_thl_parameters_tests.cpp:160:    Real thlgaus_diff_result2 = abs(SDS.thl1_2 - SDS.thl1_1);
./src/physics/shoc/tests/shoc_pdf_computetemp_tests.cpp:61:    Int num_tests = SDS.pval/abs(presincr);
./src/physics/shoc/tests/shoc_pdf_vv_parameters_tests.cpp:95:    REQUIRE(abs(SDS.w1_1) > abs(SDS.w1_2));
./src/physics/shoc/tests/shoc_pdf_vv_parameters_tests.cpp:128:    REQUIRE(abs(SDS.w1_1) < abs(SDS.w1_2));
./src/physics/shoc/impl/shoc_assumed_pdf_impl.hpp:139:      const Smask condition =  thlsec > (thl_tol*thl_tol) && ekat::abs(w1_2 - w1_1) > w_thresh;
./src/physics/shoc/impl/shoc_assumed_pdf_impl.hpp:146:        const auto tsign = ekat::abs(tmp_val_1 - tmp_val_2);
./src/physics/shoc/impl/shoc_assumed_pdf_impl.hpp:177:      const Smask condition = qwsec > (rt_tol*rt_tol) && ekat::abs(w1_2 - w1_1) > w_thresh;
./src/physics/shoc/impl/shoc_assumed_pdf_impl.hpp:182:      const auto tsign = ekat::abs(tmp_val_1 - tmp_val_2);
./src/share/tests/common_physics_functions_tests.cpp:28:    return not ( k<NumLevels && abs(lhs-rhs)>=tol );
./src/share/tests/common_physics_functions_tests.cpp:32:    return abs(computed-expected)/abs(expected) < tol;
./src/share/tests/common_physics_functions_tests.cpp:52:    return (range_mask && abs(lhs-rhs)>=tol).none();
./src/share/tests/common_physics_functions_tests.cpp:55:    return (abs(computed-expected)/abs(expected) < tol).all();
```

ambrad, line 498: For this case of a double, rather than pack, argument, I believe `abs` is C's `int abs(int)`, not C++'s `std::abs` call through to, essentially, `fabs`. Demo:
```C++
$ cat abs.cpp
#include <iostream>
int main () { std::cout << abs(-3.14) << "\n"; }
$ g++ -std=c++14 abs.cpp; if [ $? == 0 ]; then ./a.out; fi
3
$ cat abs.cpp
#include <iostream>
int main () { std::cout << std::abs(-3.14) << "\n"; }
$ g++ -std=c++14 abs.cpp; if [ $? == 0 ]; then ./a.out; fi
3.14
```

components/eamxx/src/physics/shoc/eamxx_shoc_process_interface.cpp
```diff
@@ -461,5 +465,53 @@ void SHOCMacrophysics::finalize_impl()
...
+      EKAT_KERNEL_ASSERT_MSG(mm >= cc, "Error! Total mass of column vapor should be greater than mass of surf_evap.\n");
+
+      Kokkos::parallel_for(Kokkos::TeamVectorRange(team, nlev_packs), [&](const int& k) {
+        const auto adjust = cc*qv_i(k)*pseudo_density_i(k)/mm;
+        qv_i(k) = (qv_i(k)*pseudo_density_i(k) - adjust)/pseudo_density_i(k);
+      });
+
+      surf_evap(i) = 0;
+    }
+  });
```
oksanaguba, line 514: i am confused about this code:

1. i think it is being called in each macmic iteration, which means only surf_evap/macmic_loop_size is applied in shoc iteration. does the code operate with quantity surf_evap/macmic_loop_size or does it operate with surf_evap?
2. the way i read the code, compared to the original version (before 'condition' line was introduced) is that nothing in the distribution of extra negative mass changed. i think Peter's objection was that if there is potentially negative mass in the bottom layer, then the excess would be redistributed above it. the way i read the code, right now not excess (value connected to 'condition'), but the whole surf_evap mass (value connected to 'cc') is redistributed, and not to the top of the very bottom layer, but to all layers (parallel_for loop).
3. related to that, i think the algorithm that Peter B. wanted to see would leave surf_evap not with 0, but with the biggest by the abs value mass that is allowed in the bottom layer.

Issue #2420, opened 2023-07-12
Port tbr_mtn_stress.F90


tcclevenger:
Port turbulent mountain flux code from eam and make it it's own AtmosphereProcess.

PR #2421, opened 2023-07-12
Port trb_mtn_stress.F90 with BFB test

tcclevenger:
Create physics directory `tms`, port `compute_tms()` from `eam/src/physics/cam/trb_mtn_stress.F90`. Add BFB test with original fortran code.

Part of https://github.com/E3SM-Project/scream/issues/2420.


components/eamxx/src/physics/share/physics_constants.hpp
```diff
@@ -107,6 +107,10 @@ struct Constants
   static constexpr int VTABLE_DIM1    = 10;
   static constexpr int MU_R_TABLE_DIM = 150;
 
+  // Turbulent Mountain Stress constants
+  static constexpr Scalar orocnst = 1;     // Converts from standard deviation to height [ no unit ]
+  static constexpr Scalar z0fac   = 0.075; // Factor determining z_0 from orographic standard deviation [ no unit ]
```
tcclevenger, line 112: I found these namelist defaults in eam and made them physics constants. Do we need these as variables? That's an easy switch, perhaps they can be AP params.

components/eamxx/src/physics/tms/tests/compute_tms_tests.cpp
```diff
@@ -0,0 +1,106 @@
...
+namespace scream {
+namespace tms {
+namespace unit_test {
+
+template <typename D>
+struct UnitWrap::UnitTest<D>::TestComputeTMS {
+
+  static void run_property()
+  {
+    // Should property tests be created?
```
tcclevenger, line 21: Do we need property tests?

components/eamxx/src/physics/tms/tests/compute_tms_tests.cpp
```diff
@@ -0,0 +1,106 @@
...
+      //             ncols, nlevs
+      ComputeTMSData(12,    72),
+      ComputeTMSData(8,     12),
+      ComputeTMSData(7,     16),
+      ComputeTMSData(2,     7)
+    };
+
+    // Generate random input data
+    for (auto& d : f90_data) {
+      d.randomize(engine, { {d.sgh, {0.5, 1.5}} });
```
tcclevenger, line 38: Allows equal chance of `if (horo < horomin)` being true/false in impl.

tcclevenger (commented):
Pinging @PeterCaldwell @mt5555 @bogensch since y'all were discussing on the email. I'm headed on vacation till next Wednesday (7/19/23), so if you want to wait to look at this I can ping for reviews when I get back.

In a follow up PR I'll create the AtmosphereProcess, this is only meant to be porting the compute_tms function.

tcclevenger:
I have left a few questions about property tests and parameters. With the BFB tests I'm confident we are matching F90.

jgfouca (approved):
LGTM

PR #2422, opened 2023-07-16
Add surface latent heat flux (upward) for evaporation

### LINKED ISSUE #2362: create derived output surf_latent_flux
EAMxx internally stores the flux of moisture from the surface as a water flux (units kg/m2/s). Diagnosticians are more used to the energy flux version of this variable (aka the latent heat flux, units W/m2). We should create a derived output variable to make these folks' lives easier. This is really trivial because:

surf_latent_flux = surf_evap * latent_heat_of_vaporization

pbosler:
Adds the surface latent heat flux diagnostic requested by #2362.

Adds a unit test for the diagnostic.  

Closes #2362


jeff-cohere (approved):
Looks great!

PR #2423, opened 2023-07-17
Speed up rain_sed bfb unit test on GPU

jgfouca:
The f90 implementation was taking a very long time, the CXX impl is super fast.


AaronDonahue (approved):
Does this PR impact the slowdown seen in the F90 test, as referenced in the PR description?  Since it only change cpp code I'm thinking no.

jgfouca:
@AaronDonahue , it changes `dt` on GPU so that it does way fewer steps. This affects both f90 and cxx but the primary goal was to speed up f90.

jgfouca:
That reminded me to add a comment.

jgfouca:
Passed on Blake and weaver, retesting on mappy.

PR #2424, opened 2023-07-18
fpe_guard_wrapper() now prints any exception it encounters.

jeff-cohere:
In the current master branch, any exception emitted by a function invoked using `fpe_guard_wrapper` gets swallowed, making it hard to understand what went wrong. This PR just catches the exception and sends its string representation to `stderr`. It seems like you'd want this (or something like it!) if you're trying to debug a new case.


bartgol (approved):
Yes! I ended up adding precisely this lines of code while debugging some odd behavior of the nudging branch, but forgot to share them. Thanks for the quick patch!

PR #2429, opened 2023-07-24
EAGLES: mam4 aerosol microphysics "proto-process" runs within EAMxx

jeff-cohere:
This PR allows us to create a case with "aerosol microphysics" (currently only nucleation, case configurable with instructions and materials [here](https://github.com/eagles-project/eamxx-mam4xx-ne4-case)) replacing SPA, and to run the case within EAMxx. It also adds a standalone nucleation test that demonstrates the nucleation results shown at the E3SM meeting:

mam4xx_nucleation_time_series

Tests for the mam4xx parameterizations can be found in the [mam4xx repo](https://github.com/eagles-project/mam4xx), which can be built stand-alone, and is provided as a submodule to EAMxx.

Caveat: the case with the MAM4 aerosol microphysics process doesn't yet run to completion, because RRTMGP expects aerosol optics fields that are to be handled in a different aerosol process. I have another PR lined up after this one that adds a "stub" aerosol optics package that provides those fields with some constant values that do allow EAMxx to run to completion when SPA is replaced by the MAM4 aerosol microphysics and optics processes.

These initial EAGLES PRs are going to be a little weird, because we're still standing things up and getting cases to run to completion. Very little analysis has been performed on the ne4 cases we've been using to make progress, mostly because we've only recently gotten things to run to completion. Happy to answer any questions anyone might have. Please add any other reviewers you think may be interested in this line of work (though I think @susburrows is on vacation this week).


jeff-cohere:
@AaronDonahue and @bartgol : can we get Naser Mahfouz (Github username: `mahf708`) added to the SCREAM repo? Susannah asked for him to be a reviewer on EAGLES PRs.

AaronDonahue:
> @AaronDonahue and @bartgol : can we add @mahf708 to the SCREAM repo? Susannah asked for him to be a reviewer on EAGLES PRs.

@mahf708 do you already have access.  @rljacob can correct me if I'm wrong but I thought everyone w/ E3SM access also had access to the SCREAM repo.  Actually, I recall in the past having someone completely outside of the E3SM project team comment once on one of our PR's so perhaps the repo is publicly available?

jeff-cohere:
Thanks for the very detailed comments, Naser. I'll try to address them in-line from the perspective of a software engineer involved in the porting of MAM4's Fortran implementation to C++. I want to make clear that the porting team was asked to do a "direct port" of MAM4, without addressing any issues related to the model itself. This has guided our strategy in certain ways, such as the integration with EAMxx (including this PR) effectively being separate from the activity of porting the Fortran code to C++ itself, which happens entirely within the mam4xx repo.
 
> Mathematically, I view (aerosol) new-particle formation (NPF) as a boundary condition represented as flux, very strictly speaking. (New-particle formation is likely a better term to use for nucleation in this context, but that's besides the point here.) Essentially, nucleation is a flux bringing particles from nowhere (i.e., smaller than their arbitrarily defined "fully realized" size) to a particle distribution of some sort. It so happens that it really doesn't matter how exactly we represent nucleation especially in a modal representation like this (as a boundary condition or as a source term or as whatever) as long as the accounting makes sense. As an aside, I actually had the most scientific fun [writing a paper on nucleation fluxes vis-a-vis condensation fluxes a few years ago!](https://agupubs.onlinelibrary.wiley.com/doi/abs/10.1029/2021GL092758) Maybe my views on nucleation will be controversial in these corners, but hey, I stand by them ... I am mentioning this because I believe the nucleation parameterization here is a "nucleation + growth" parameterization, strictly speaking, because it has to be so in a modal representation (unlike in more detailed models).

That is correct--the nucleation process you see here reflects the nucleation and growth of sulfate aerosol formed under binary collisions between H2SO4 gas molecules, which are then grown to a size appropriate for them to be deposited into the aitken mode.
 
> In the PBL case, nucleation is represented as a phase-conversion process. It seems we have something like: H2SO4 --> SO4a with a first-order (`pbl_nuc_wang2008_user_choice = 1`[1](#user-content-fn-1-4fc6477d76b1e9b83f5befdb04bcc0e0)) rate coefficient of 1e-6 /time, so that in equilibrium, the rate of H2SO4 loss is the same as the rate of SO4a production and it is 1e-6 * [H2SO4] in units of concentration per time. The time unit appears to be second. Why am I going on about this? Because we should be able to see the exact constant in the plots you're sharing, but I am having difficulty seeing it! To this end, and if deemed possible and in-scope for this PR, I have two minor requests (but feel free to ignore them if there is no time or if you think they won't help).
> 
>     1. Could you please extend the time span further and plot it (x-axis) in seconds?
> 
>     2. And could you also plot the y-axis as a natural logarithm of concentration (or mass mixing ratio)?
> 
> 
> _(EDIT: Something I didn't think through carefully was the PBL vs tropospheric/stratospheric parameterizations. The above only applies for the PBL one (I added a note to be clearer). In order for my two items to be meaningful, the plots will have to be in the PBL only (if that's actually doable). If not doable, it would still be helpful to see how much the effective (total) reaction rate coefficient (as guessed from slope of the ln(mmr) vs time) of the two separate parameterizations together deviates from 1e-6. The tropospheric/stratospheric parameterization is technically a binary one, between sulfuric acid and water --- the binary effective rate coefficient appears to vary quite a bit depending on sulfuric acid concentration from 1e-9/s to 1e-3/s, so it may be close to 1e-6 anyway, depending on the starting concentration...)_

This PR (and the next, which involves adding an aerosol optics package that returns constant values for aerosol optical properties until we replace them with the real ones) are really "software plumbing" PRs with minimal scientific content. Their purpose is to make sure that all the C++/Kokkos stuff works between EAMxx and the standalone mam4xx column model. So instead of plugging in all of MAM4's aerosol microphysics, we started with only nucleation, which gives us a simple set of tendencies among tracers decoupled from the rest of the atmospheric physics. Personally, I had intended the nucleation effects to be a qualitative indication that things are or aren't hooked up properly, so we didn't incorporate a workflow that allowed us to study the actual binary nucleation rates, which are computed using fit coefficients for various quantities. We have actually validated these nucleation rates themselves against reference data computed by the Fortran version of MAM4. The Fortran and C++ code take the same inputs and produce output in the same form, so the comparison there is pretty straightforward. This is in line with our porting validation strategy, which I'd be happy to discuss with you. @pressel and @singhbalwinder can also provide information about this strategy.

We could set up a more rigorous test of the nucleation in the EAMxx setting as you propose, but it would take some consideration and work to figure out how to do it in a meaningful way. Since we haven't really discussed the validation of mam4xx running within EAMxx yet, I expect the scientists on the EAGLES project might have opinions about how we do this.

For what it's worth: after we merge this PR and the "stand-in" optics process, my next objective is to incorporate all of the MAM4 aerosol microphysics parameterizations as implemented by the "interface" function [here](https://github.com/eagles-project/mam4xx/blob/main/src/tests/mam4_amicphys_1gridcell.cpp#L1513). This function is structured identically to its Fortran counterpart, and calls all of the aerosol microphysics parameterizations in a specific order, handling any intermediate quantities the same way they are handled in the original implementation.

Once we've got aerosol microphysics in, it becomes more essential for us (those on EAGLES, and those interested from the E3SM side) to think harder about how we start to evaluate MAM4's behavior within EAMxx. One could argue that we should pause before bringing in aerosol microphysics so we can more carefully evaluate the nucleation process itself. But given that EAMxx won't even run a simulation without this PR and the optics one, I believe it will be necessary to get these two PRs merged before we start asking probing questions.

components/eamxx/tests/uncoupled/mam4/input.yaml
```diff
@@ -24,10 +24,13 @@ grids_manager:
 initial_conditions:
   # The name of the file containing the initial conditions for this test.
   Filename: ${SCREAM_DATA_DIR}/init/${EAMxx_tests_IC_FILE_72lev}
-  q_aitken_so4 : 0.0
-  n_aitken : 0.0
-  q_h2so4 : 0.0000000002
-  pbl_height : 400.0
+  q_aitken_so4: 0.0
+  n_aitken: 0.0
+  q_h2so4: 1.9186478479542893e-011 # 0.65e-10 is from namelist, but this is what gets to nucleation
```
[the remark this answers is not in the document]
jeff-cohere: Yes: `e-011` is the same as `e-11`, and this code was likely machine-generated or copy-pasted.

components/eamxx/src/physics/mam/eamxx_mam_microphysics_process_interface.cpp
```diff
@@ -357,8 +356,10 @@ void MAMMicrophysics::run_impl(const double dt) {
       ekat::subview(w_updraft, icol),
       pblh_(icol));
 
+    // set surface state data
+    haero::Surface sfc{};
```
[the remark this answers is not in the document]
jeff-cohere: I don't believe MAM4's nucleation process interacts with the surface directly at all, so we just create a "blank" `Surface` representation here to pass in to the process, as required by its interface. Down the line, this surface representation will actually need to be filled with data that is used by other aerosol processes. Since all these processes have the same interface, the Surface gets passed to each one whether or not it actually is used. The uniformity of the interface makes it easier to change the implementation/underlying model in an aerosol process without cascading code changes.

components/eamxx/src/physics/mam/eamxx_mam_microphysics_process_interface.cpp
```diff
@@ -307,6 +295,17 @@ void MAMMicrophysics::initialize_impl(const RunType run_type) {
...
+  mam4::NucleationProcess::ProcessConfig nuc_config{};
+  nuc_config.dens_so4a_host = 1770.0;
+  nuc_config.mw_so4a_host = 115.0;
+  nuc_config.newnuc_method_user_choice = 2;
+  nuc_config.pbl_nuc_wang2008_user_choice = 1;
+  nuc_config.adjust_factor_pbl_ratenucl = 1.0;
+  nuc_config.accom_coef_h2so4 = 1.0;
+  nuc_config.newnuc_adjust_factor_dnaitdt = 1.0;
+  nucleation_->init(nuc_config);
+
```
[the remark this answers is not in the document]
jeff-cohere: I agree with you that whatever ends up here should be documented/annotated. I didn't do it here because the nucleation parameterization is just a stand-in for aerosol microphysics as we get the basic transfer of data sorted. In general, I think it would be helpful to develop naming conditions too. Something else for us to discuss, perhaps.

To your questions:

* The "a" in `so4a` is indeed for "aerosol".
* "Host" has two meanings in the context of this project
    1. We refer to the atmosphere model in which mam4xx runs (EAMxx, in this case) as a "host model".
    2. HPC software engineers refer to "host" and "device" when they want to distinguish between a CPU that is dispatching parallel work and a compute device that does the work (usually either a CPU via threads or a GPU). `so4a_host` refers to a property (density or molecular weight) of SO4 aerosol as defined by the host. I am told this is to allow for properties describing characteristic mixtures of SO4 with other things in the air.
    3. The accommodation coefficient does seem to be involved with particle growth (specifically [here](https://github.com/eagles-project/mam4xx/blob/main/src/mam4xx/nucleation.hpp#L509))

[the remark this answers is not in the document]
jeff-cohere: Right. I should have said "so4a_host refers to a property (density or molecular weight) of SO4 aerosol as defined by the host _model_". It's a mechanism for EAMxx to set these properties according to its desired "mixtures".

components/eamxx/src/mct_coupling/CMakeLists.txt
```diff
@@ -32,6 +32,7 @@ set (SCREAM_LIBS
      scream_share
      scream_control
      ${dynLibName}
+     mam
```
bartgol, line 35: Shouldn't this be guarded by an `if (SCREAM_ENABLE_MAM)` clause?

components/eamxx/tests/coupled/dynamics_physics/homme_mam4xx_pg2/homme_mam4xx_pg2.cpp
```diff
@@ -0,0 +1,70 @@
...
+  const auto  nsteps = ts.get<int>("number_of_steps");
+  const auto  t0_str = ts.get<std::string>("run_t0");
+  const auto  t0     = util::str_to_time_stamp(t0_str);
+
+  logger.info("running HOMME/MAMMicrophysics coupled test with dt = {} for {} steps.", dt, nsteps);
+
+  // Register all atm procs and the grids manager in the respective factories
+  register_dynamics();
+  register_physics();
+// TODO: register_diagnostics();
```
bartgol, line 46: This line is not needed (and actually misleading, since diags should not be registered here).

components/eamxx/tests/uncoupled/mam4/mam4_nucleation_standalone.cpp
```diff
@@ -38,10 +39,11 @@ TEST_CASE("mam4-nucleation-standalone", "") {
   const auto  t0_str = ts.get<std::string>("run_t0");
   const auto  t0     = util::str_to_time_stamp(t0_str);
 
+  logger.info("running MAMMicrophysics standalone test with dt = {} for {} steps.", dt, nsteps);
+
   // Need to register products in the factory *before* we create any atm process or grids manager.
-  auto& proc_factory = AtmosphereProcessFactory::instance();
+  register_physics();
   auto& gm_factory = GridsManagerFactory::instance();
```
bartgol, line 46: To mimic the line above, where you used `register_physics` to hide tedious lines that the end user should not care about, you could simply do
```
register_mesh_free_grids_manager();
```
rather than do it manually.

bartgol (dismissed):
Looks good. I have a couple of comments, but nothing major. Perhaps just the comment on the cmake if statement, assuming I followed correctly.

components/eamxx/src/mct_coupling/CMakeLists.txt
jeff-cohere: Ah, good eye! I'll fix that.

jeff-cohere:
Thanks for the comments, @bartgol . I've addressed them with my last commit. Do you and @AaronDonahue think we should have others look at this PR before it goes in, or just proceed for now?

I have another PR ready to go after this one that adds a stub aerosol optics AtmosphereProcess that feed RRTMGP the optics fields it needs. That's the last one we need to be able to run simulations, identify and fix problems, and start adding in more aerosol physics.

bartgol (approved):

bartgol:
> Thanks for the comments, @bartgol . I've addressed them with my last commit. Do you and @AaronDonahue think we should have others look at this PR before it goes in, or just proceed for now?
> 
> I have another PR ready to go after this one that adds a stub aerosol optics AtmosphereProcess that feeds RRTMGP the optics fields it needs. That's the last one we need to be able to run simulations, identify and fix problems, and start adding in more aerosol physics.

I'd say go ahead and merge. As far as I understand, this PR is about plumbing of MAM4xx into EAMxx, so the aerosol details are not key. However, if you want to have more aerofolks look at it, that's fine too.

susburrows:
@jeff-cohere I would ideally like to review this PR, but I won't have time until after the ARM/ASR meeting (so, starting Aug 14).  I don't want to hold up your process if waiting for me would slow you down on other tasks.  So, please feel free to go ahead without me.  Or, if it can wait until mid-August, I can look at it then (we will need to reconnect around that time in any case).

jeff-cohere:
@susburrows, thanks. Given that this PR and the next are "just" plumbing, I'd like to try to get them in sooner rather than later so we can discuss next steps (where the physics begins!). I'm planning on setting up a configurable microphysics process that allows scientists to switch specific parameterizations on and off, set up extra diagnostics, etc.

components/eamxx/src/physics/register_physics.hpp
```diff
@@ -51,7 +51,7 @@ inline void register_physics () {
   proc_factory.register_product("Nudging",&create_atmosphere_process<Nudging>);
 #endif
 #ifdef EAMXX_HAS_MAM
-  proc_factory.register_product("MAMMicrophysics",&create_atmosphere_process<MAMMicrophysics>);
+  proc_factory.register_product("MAM4Microphysics",&create_atmosphere_process<MAMMicrophysics>);
```
pbosler, line 54: Should this string be `mam4_micro`, to match the output of the `name()` function?

components/eamxx/tests/uncoupled/mam4/input.yaml
```diff
@@ -9,8 +9,8 @@ time_stepping:
   number_of_steps: ${NUM_STEPS}
 
 atmosphere_processes:
-  atm_procs_list: (MAMMicrophysics)
-  MAMMicrophysics:
+  atm_procs_list: (mam4microphysics)
+  mam4_microphysics:
```
pbosler, line 13: Should these match `mam4_micro`, or are they different for a reason?

pbosler (approved):

components/eamxx/src/physics/register_physics.hpp
jeff-cohere: If I had my way, this process would have a single name everywhere. I don't really understand the reasoning for all the names I see floating around in the code. I believe we can fix this in the next PR.

components/eamxx/tests/uncoupled/mam4/input.yaml
bartgol: yeah, this does seem like a bug. The name within the parentheses must match verbatim the name on the next line.

jeff-cohere: The real issue is that we have names for `AtmosphereProcess`s that are independent of how these processes are registered in `register_physics`. This issue is only going to cause more confusion as time passes.

For now, I'm going to change all these names to be `mam4_micro`, but in my view the right way to solve this is to enforce that the registered name and the `name()` method are always the same.

bartgol: Ideally, we would like to call `P3::name()` in the factory registration, but the `name` method is not static (and cannot be, since it's virtual).

We could have derived classes have a "name_static()" method, and have their `name()` function call the static one, and use the static method in the factory (to avoid bugs due to typos). Unfortunately we _do_ need the name method, since the base class uses it when it's printing messages inside generic interfaces (e.g., put the ap name in the logger, or in the timings file).

bartgol (approved):

PR #2431, opened 2023-07-25
Make one scorpio_input.cpp error message for robust, remove another check.

AaronDonahue:
Just adds more context to an error message so that when the error is thrown it is easier to decipher why.  This commit also removes a bad check that throws an error on legitimate runs.


sarats:
This PR fixed a runtime abort on Frontier when running ne1024 on 8k nodes (64k MPI ranks) apparently as the number of ranks exceeded the DOF in the ne30np4 SPA data (48602 DOF).

sarats (approved):
Removes false alarm/abort and adds detail to error message.

PR #2432, opened 2023-07-26
add small kernel for p3

xyuan:
add small kernel option for p3

### DIFF
--- a/components/eamxx/src/physics/p3/CMakeLists.txt
+++ b/components/eamxx/src/physics/p3/CMakeLists.txt
@@ -61,19 +61,46 @@ if (NOT EAMXX_ENABLE_GPU OR Kokkos_ENABLE_CUDA_RELOCATABLE_DEVICE_CODE OR Kokkos
   ) # P3 ETI SRCS
 endif()
 
-add_library(p3 ${P3_SRCS})
+# List of dispatch source files if monolithic kernels are off
+set(P3_SK_SRCS
+    disp/p3_check_values_impl_disp.cpp  
+    disp/p3_ice_sed_impl_disp.cpp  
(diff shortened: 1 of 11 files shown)


jgfouca (dismissed):
Nice job! This appears to have followed the pattern in shoc, which is what I expected.

sarats:
Curious: Have you tested this on Frontier? What kind of performance improvements did you see?

xyuan:
> Curious: Have you tested this on Frontier? What kind of performance improvements did you see?

not tested on Frontier yet

AaronDonahue:
@xyuan , could you try merging `master` into this branch and pushing that change?  I'm still confused on why you are seeing build errors on `weaver` when I can build manually.  Note, the AT should already be doing the merge of master, so I'm not confident this will solve the problem.  But it is a low-hanging fruit to try.

jgfouca:
I'm going to jump on weaver and see if I can get this to build.

jgfouca:
OK, there's a race in the build system:

```
f951: Fatal Error: Cannot rename module file 'modules/micro_p3.mod0' to 'modules/micro_p3.mod': No such file or directory
```

Appears about 1/3 of time. I'm trying to figure out why. We are using the same pattern as shoc small kernels and that has always worked fine.

bartgol:
The two libs should use different FORTRAN_MODULE_DIRECTORY, but they are using the same. p3 should do what shoc does in the CMakeLists.txt

jgfouca:
@bartgol , yep, I see it now. I will add commit

jgfouca:
OK, lets hope tests pass now.

jgfouca (approved):

PR #2433, opened 2023-07-26
Fix time averaged output to account for filled or masked values

AaronDonahue:
Add support for masked/filled values in scorpio_output
    
    This commit enhances the `combine` function in scorpio_output.cpp
    to support cases where some values may be masked or filled.  This
    is essential to properly handle output that utilizes vertical remapping,
    which may include some filled values.
    
    There is a different treatment of filled values depending on the averaging
    type for the output.
     - INSTANT: no change
     - MIN and MAX: Only use the new_val to update if it is unfilled.
     - AVERAGE: Only add the contribution of the new_val if it is unfilled.
    
    *Note on AVERAGE.  Because the AVERAGE output is later scaled by the
    number of points that contribute to the average this commit also tracks
    the number of times a new value was used to update the output value over
    the averaging period.  This is then used with AVERAGE to scale the output.


components/eamxx/src/share/io/scorpio_output.cpp
```diff
@@ -20,23 +20,36 @@ namespace scream
...
-      curr_val += new_val;
-      break;
-    default:
-      EKAT_KERNEL_ERROR_MSG ("Unexpected value for m_avg_type. Please, contact developers.\n");
+  const bool new_fill  = new_val == fill_value;
+  const bool curr_fill = curr_val == fill_value;
+  if (!new_fill) {
+    avg_coeff += 1;
+  }
+  if (curr_fill & new_fill) {
```
tcclevenger, line 30: Did you mean `&&`?

components/eamxx/src/share/io/scorpio_output.cpp
```diff
@@ -89,8 +102,13 @@ AtmosphereOutput (const ekat::Comm& comm, const ekat::ParameterList& params,
...
-  if (params.isParameter("Fill Value")) {
-    m_fill_value = static_cast<float>(params.get<double>("Fill Value"));
+  if (params.isParameter("fill_value")) {
+    m_fill_value = static_cast<float>(params.get<double>("fill_value"));
+  }
+  if (params.isParameter("fill_threshold")) {
+    m_avg_coeff_threshold = params.get<Real>("fill_threshold");
+  } else {
+    m_avg_coeff_threshold = 0.5; // default to 0.5
   }
```
tcclevenger, line 112: No need for if-else, There is a built in default arg
```
m_avg_coeff_threshold = params.get<Real>("fill_threshold", 0.5);
```

components/eamxx/src/share/io/scorpio_output.cpp
```diff
@@ -89,8 +102,13 @@ AtmosphereOutput (const ekat::Comm& comm, const ekat::ParameterList& params,
 {
   using vos_t = std::vector<std::string>;
 
-  if (params.isParameter("Fill Value")) {
-    m_fill_value = static_cast<float>(params.get<double>("Fill Value"));
+  if (params.isParameter("fill_value")) {
+    m_fill_value = static_cast<float>(params.get<double>("fill_value"));
```
tcclevenger, line 106: Is there a default for `m_fill_value`?

components/eamxx/src/share/io/scorpio_output.cpp
AaronDonahue: Good catch

components/eamxx/src/share/io/scorpio_output.cpp
AaronDonahue: yes, it is assigned in the `.hpp` file: https://github.com/E3SM-Project/scream/pull/2433/files#diff-47d669cb9ede18b477ba793844458e7c86ec56590372ce145d17412eaed0d742R202

components/eamxx/tests/uncoupled/surface_coupling/surface_coupling.cpp
```diff
@@ -260,20 +262,25 @@ void test_imports(const FieldManager& fm,
     EKAT_REQUIRE(surf_evap(i)       == import_constant_multiple_view(13)*import_data_view(i, import_cpl_indices_view(13)));
 
     // The following are only imported during run phase. If this test is called
-    // during initialization, all values should still be 0.
+    // during initialization, all values should be the default value.
+    // TODO: Why are some of these FillValue and others are 0.0?  For the former,
```
AaronDonahue, line 266: @tcclevenger , do you have any thoughts on the behavior I am describing in this TODO comment?

tcclevenger: This is certainly a bug. Maybe it's related to the `curr_fill & new_fill` if statement above?

tcclevenger (changes requested):
I think `curr_fill & new_fill` still needs to be fixed, and may be the reason for the bug in the SC test.

components/eamxx/src/share/io/scorpio_output.cpp
bartgol: FYI, both `&` and `&&` work correctly for bool types in C++ (not necessarily in C!). However, it's true that someone may be puzzled by this, and maybe think that some nifty bitwise check is needed here, so the `&&` operator is probably cleaner.

components/eamxx/src/share/io/scorpio_output.cpp
bartgol: That's b/c you are calling `get` on the input parameter list, which is const. The default adds an entry if one is not found, which is why you can't call it on a const list. If you call it on m_params, you should not get the error.

An alternative (which I have thought about a few times) is to change the impl of `get` with a default. The current impl, mimics the behavior of `ParameterList` in Trilinos, but one may argue that it's somewhat of a stealth list modification. Perhaps, when a param is not found, we should simply _return_ the default, without adding an entry to the plist. This would be an ekat mod, but I'm afraid it would not be bwd compatible (which may or may not be a problem, if we agree that the current behavior is not intuitive).

components/eamxx/tests/uncoupled/surface_coupling/surface_coupling.cpp
AaronDonahue: so weird, I swear I changed that in the latest commit.... oh well, git mysteries.  I'll fix it again and see if that also fixes the SC test.

components/eamxx/src/share/io/scorpio_output.cpp
AaronDonahue: ahhh thanks for the explanation.  I'll make the change to use m_params to keep it clean

AaronDonahue: nevermind, this approach didn't work.  So I kept it as is.

tcclevenger (approved):

components/eamxx/src/share/io/scorpio_output.cpp
```diff
@@ -20,23 +20,36 @@ namespace scream
...
-      curr_val = ekat::impl::min(curr_val,new_val);
-      break;
-    case OutputAvgType::Average:
-      curr_val += new_val;
-      break;
-    default:
-      EKAT_KERNEL_ERROR_MSG ("Unexpected value for m_avg_type. Please, contact developers.\n");
+  const bool new_fill  = new_val == fill_value;
+  const bool curr_fill = curr_val == fill_value;
+  if (!new_fill) {
```
bartgol, line 27: There are lots of conditionals here. For "normal" output, all of these are checked, which may slow down things, since the conditions may change across iterations (except for the avg type switch). Can we attach a flag to the fields that "may" contain fill values, and only do all these checks for those? Perhaps we should have `combine` and `combine_with_fill`.

Beside performance, my real concern is that all these cases of fill values are making our code quite complex and hard to follow, in a lot of places, all for the sake (if I understand correctly) of one particular use case (vertical remap) which we don't use that often.

I know we all have projects to follow, and deadlines, so I won't make a stand, but I'm going to open an issue, and see if we can clean up a bit.

components/eamxx/src/share/io/scorpio_output.cpp
```diff
@@ -269,12 +285,15 @@ void AtmosphereOutput::restart (const std::string& filename)
...
   hist_restart.read_variables();
+  auto coeff_rest = scorpio::get_attribute<int>(filename,"num_snapshots_since_last_write");
   hist_restart.finalize();
   for (auto& it : m_host_views_1d) {
     const auto& name = it.first;
     const auto& host = it.second;
     const auto& dev  = m_dev_views_1d.at(name);
+    const auto& coeff = m_avg_coeff_views_1d[name];
     Kokkos::deep_copy(dev,host);
+    Kokkos::deep_copy(coeff,Real(coeff_rest));
```
bartgol, line 296: Wait, why do you restart with coeff being the num snaps since last write? Isn't the point of `m_avg_coeff_views_1d` to keep track of the number of valid snaps for each entry? Shouldn't this simply be a field we read from the rhist file?

components/eamxx/src/share/io/tests/io_basic.cpp
```diff
@@ -224,20 +244,42 @@ void read (const std::string& avg_type, const std::string& freq_units,
...
+        f0 = fm0->get_field("f_3").clone();
+      }
       auto f  = fm->get_field(fn);
       if (avg_type=="MIN") {
         // The 1st snap in the avg window (the smallest)
         // is one past window_start=n*freq
         add(f0,n*freq+1);
+	// TODO: The MIN function ignores the initial condition, so the filling
+	// doesn't change any results for this test.  
+	// QUESTION: Should MIN include initial condition, should MAX as well?
```
bartgol, line 257: I don't think so. We do average (including min/max) by picking the rhs point of each interval.

components/eamxx/src/share/io/scorpio_output.cpp
AaronDonahue: yes, thats the last commit I'm working on now.

components/eamxx/src/share/io/scorpio_output.cpp
```diff
@@ -481,7 +481,9 @@ run (const std::string& filename,
           auto avg_coeff_1d = view_Nd_dev<1>(avg_cnt_data,avg_cnt_dims[0]);
           Kokkos::parallel_for(policy, KOKKOS_LAMBDA(int i) {
 	    if (m_track_avg_cnt && m_add_time_dim) {
+	      printf("ASD - %s\t, %d, %e, %e, %f, %e  -->  ",name.c_str(),i,new_view_1d(i), avg_view_1d(i), avg_coeff_1d(i), m_fill_value);
```
AaronDonahue, line 484: this was for debugging, I'll remove in the next commit.

components/eamxx/src/share/io/scorpio_output.cpp
```diff
@@ -384,6 +382,43 @@ run (const std::string& filename,
...
+    for (const auto name : m_avg_cnt_names) {
+      auto& dev_view = m_local_tmp_avg_cnt_views_1d.at(name);
+      Kokkos::deep_copy(dev_view,1.0);
+    }
+    // Now we cycle through all the fields
+    for (const auto name : m_fields_names) {
+      auto field    = get_field(name,"io");
+      auto lookup   = m_field_to_avg_cnt_map.at(name);
+      auto dev_view = m_local_tmp_avg_cnt_views_1d.at(lookup);
+      update_avg_cnt_view(field,dev_view);
```
bartgol, line 408: Aren't you calling this fcn many times, if there are 2+ fields with the same avg count field?

AaronDonahue: yes, the idea being that we want consistency over each layout.  So one variable may have a fill value, while others won't.  I think I outline in a comment that for such a case we will mask that value for all fields that share that layout.  Thus, I need to loop over all fields and check for fill values because I won't know ahead of time which one(s) have been filled.

essentially, if I have variable A and B on layout L and only A has a fill value at point `i`.  In this current implementation we would flag point `i` as "filled" and skip it for averaging calculations in both field `A` and field `B`.  In order to detect the fill value at point `i` I need to check both `A` and `B`.

I'm open to suggestions on how to approach this differently.  This was the only way I could think of that safely follows the assumption that all layouts are "filled" with the same pattern, and account for an edge case where some variable doesn't actually have a fill value.

bartgol: Well, if all fields with the same layout have the same mask, then you only need to process the lookup field once. The way it is now, I think you end up doing lookup+=N on the non-masked entries, where N is the number of fields that share that mask field...

AaronDonahue: yes, my concern here is for cases where there are two fields with the same layout, but only 1 actually has the filled value.  Maybe the other has a reasonable value.  If I only check a single field for that layout then won't I be running the risk that I won't catch this case consistently.

I'm trying to anticipate future use cases that would be hard to debug.  If I only check one field and for some reason a user sets up a process that fills only a subset of variables on a specific layout then I run the risk of "averaging" over filled values.  I essentially don't trust that a in the future we won't have cases where the same layout see variables with and without fill values.  And there won't be an obvious error to throw in that case.

bartgol: If two fields have the same mask field attached to them, then they are masked the same way. If parametrization XYZ fills field "blah" only at even cols, then parametrization XYZ is responsible for attaching a different mask to "blah".

It's not a matter of "layout", but more a matter of "mask": if two fields are masked in the same way, then at every given entry they are either both valid or both invalid.

bartgol: Regardless of the motivation, with the code you have, if two fields have the same "lookup" field, you end up with avg_cnt+=N in some entries, which doesn't really make sense. The avg_cnt may even be different at different entries, no? What does the lookup field even represent here then?

AaronDonahue: ahh yes, I see the confusion.  the lookup is a map of a `field` to the corresponding `mask` field for it.  So if N fields are on layout L then the lookup will have N pointing to 1 unique dev_view.  So the idea here is to cycle through every field associated with that dev_view and ensure that if that field has a masked value that the corresponding location in the mask view is represented as "masked".  The function "update_avg_cnt_view" might be a poor name for the function, if you look at it, all it does is set values to 0 if a mask is present.  So at the end you should only have a view of 0's and 1's.

AaronDonahue: down below, the for loop there updates the true average count over the span of averaged time steps by adding the final view from this step.

bartgol: Oh, I see, then discard my +=N comment.

Nevertheless: you must be assuming that all field with the same mask have invalid values at the same location, right? Otherwise it would not make much sense: if field A is invalid where field B is valid and viceversa, the count will be 0 for both fields, even though each of them is valid in half of the entries...

Therefore, we should only have to call update_avg_cnt once for each mask field...

AaronDonahue: Right, I agree, although w/ masked values things are a bit weird.  I'm trying to hedge against the case where field A has a valid value at a point where field B is invalid.  If we only check one field we have the potential of it being A, thus we will flag the value as valid then B gets an incorrect value in the averaging.  It shouldn't happen, but if it does it would a tricky bug to catch.

AaronDonahue: The alternative would be to throw an error in this case.  The final alternative is to have each field get it's own masking array, but that will double the size of the restart history file, so I think that's a non-starter.

bartgol: I don't understand why field A would be valid if field B would be invalid, if they share the same mask field. If field A not only is remapped (like B) but also, say, masked out by radiation (to set it to invalid values over night columns), then field B should have a different mask attached to it.

So if two fields have the same mask field attached, they must have valid/invalid values in the _same_ locations.

AaronDonahue: If we are talking only about vertical remapping, then yes, vertical remapping will apply the same mask to all fields on the same layout.  With this redundancy I'm attempting to tackle a more general application, if a specific atmospheric process ends up setting a value to the fill value for field A, because that is the only field it specifically cares about, we find ourselves in the case I described above.

TBH - I'm not excited about this approach, so I am open to push back or suggestions.  We could, for example, just support masking in the case of online remapping .  We don't currently have any processes that mask field values.  That being said, I don't think it is unreasonable to assume that in the future someone will introduce a process that wants to do masking.  For example, nudging may want to mask some nudging tendencies that are of bounds so we output nudging tendencies we find ourselves in this case.

components/eamxx/src/share/io/scorpio_output.cpp
```diff
@@ -116,6 +117,7 @@ AtmosphereOutput (const ekat::Comm& comm, const ekat::ParameterList& params,
  : m_comm         (comm)
  , m_add_time_dim (true)
 {
+  params.print();
```
bartgol, line 120: This (and the ASD print statements below) should probably be removed before merging.

components/eamxx/src/share/io/tests/io_basic.cpp
```diff
@@ -27,7 +27,7 @@
 namespace scream {
 
 constexpr int num_output_steps = 5;
-constexpr Real FillValue = DEFAULT_FILL_VALUE;
+constexpr Real FillValue = constants::DefaultFillValue<float>().value;
```
bartgol, line 30: I think we could maybe move the testing of fill value in a separate test? I think it's clearer if we keep the `io_basic.cpp` tests...well...basic. :)

There may be some code duplication, but that's prob justified by the fact that it makes it easier to debug I/O failures, since we're not mixing up many use cases in one.

PR #2435, opened 2023-07-27
Add COSP to EAMxx

brhillman:
Add interface to COSP to EAMxx. This adds COSP as an additional atmosphere process. So far, this just adds the ability to run the ISCCP simulator, but brings in all of the infrastructure to be able to run the full suite of simulators once we populate more of the fields. Adds three new field managed variables available for output: `isccp_cldtot`, `isccp_ctptau`, and `isccp_mask`. These outputs are weighted by `isccp_mask`, which is 0 for night columns and 1 for valid day columns. Masked means can then be computed from these outputs via something like `isccp_cldtot_masked_average = avg(isccp_cldtot) / avg(isccp_mask)`. This is needed since we do not yet have support for time-aggregation of output fields with missing values, but the resulting means come out to be the same.

[B4B]


components/eamxx/src/physics/cosp/cosp_functions.cpp
```diff
@@ -0,0 +1,12 @@
+#include "cosp_functions.hpp"
+namespace scream {
+    namespace CospFunc {
+        void initialize(int ncol, int nsubcol, int nlay) {
+            std::cout << "brhdebug: call cosp_c2f_init()" << std::endl;
```
bartgol, line 5: On massively large simulations, this is going to inject lots of lines in e3sm.log. Perhaps you could pass a flag to the init function, telling it whether to print to screen or not?

components/eamxx/src/physics/cosp/cosp_functions.hpp
```diff
@@ -0,0 +1,96 @@
...
+        using lview_host_3d = typename ekat::KokkosTypes<HostDevice>::template lview<Real***>;
+        template <typename S>
+        using view_1d = typename ekat::KokkosTypes<HostDevice>::template view_1d<S>;
+        template <typename S>
+        using view_2d = typename ekat::KokkosTypes<HostDevice>::template view_2d<S>;
+        template <typename S>
+        using view_3d = typename ekat::KokkosTypes<HostDevice>::template view_3d<S>;
+
+        void initialize(int ncol, int nsubcol, int nlay) {
+            cosp_c2f_init(ncol, nsubcol, nlay);
```
bartgol, line 27: If you provide an impl for initialize/finalize, then
1. you don't need the cosp_functions.cpp file anymore (otherwise you have multiple def of the same fcn within each translation unit)
2. you need to make these functions inline (otherwise you will have multiple def of the same fcn across all translation units)

components/eamxx/src/physics/cosp/cosp_functions.hpp
```diff
@@ -0,0 +1,96 @@
...
+            // Make host copies and permute data as needed
+            lview_host_2d
+                  T_mid_h("T_mid_h", ncol, nlay), p_mid_h("p_mid_h", ncol, nlay), p_int_h("p_int_h", ncol, nlay+1),
+                  qv_h("qv_h", ncol, nlay), cldfrac_h("cldfrac_h", ncol, nlay),
+                  reff_qc_h("reff_qc_h", ncol, nlay), reff_qi_h("reff_qi_h", ncol, nlay),
+                  dtau067_h("dtau_067_h", ncol, nlay), dtau105_h("dtau105_h", ncol, nlay);
+            lview_host_3d isccp_ctptau_h("isccp_ctptau_h", ncol, ntau, nctp);
+            // NOTE: these should already be host views, so we could probably
+            // skip creating the mirror views here
+            auto sunlit_h = Kokkos::create_mirror_view(sunlit);
```
bartgol, line 50: Your `view_1d`/`view_2d` types are already on host. Why do you need to create mirrors for sunlit, skt, and isccp_cldtot?

components/eamxx/src/physics/cosp/cosp_functions.hpp
```diff
@@ -0,0 +1,96 @@
...
+            auto skt_h = Kokkos::create_mirror_view(skt);
+            auto isccp_cldtot_h = create_mirror_view(isccp_cldtot);
+
+            // Copy to layoutLeft host views
+            Kokkos::deep_copy(sunlit_h, sunlit);
+            Kokkos::deep_copy(skt_h, skt);
+            for (int i = 0; i < ncol; i++) {
```
bartgol, line 57: This is fine, but in case you would like to hide this manual transposition, ekat offers `ekat::transpose`, which can be used like this:
```
ekat::transpose(T_mid.data(),T_mid_h.data(),ncol,nlay,ekat::c2f);
```
Again, this is fine though.

components/eamxx/src/physics/cosp/eamxx_cosp.cpp
```diff
@@ -0,0 +1,190 @@
...
+
+// =========================================================================================
+void Cosp::run_impl (const double dt)
+{
+  // Determine if we should update COSP this timestep; use rad function to compare COSP frequency with this timestep
+  auto ts = timestamp();
+  auto cosp_freq_in_steps = m_cosp_frequency;
+  if (m_cosp_frequency < 0) {
+      // interpret cosp_freq as nhours
+      cosp_freq_in_steps = -3600.0 * m_cosp_frequency / dt;
```
bartgol, line 99: I would double check that `dt` divides 3600 here, or else you may get some unexpected behavior. I'm sure this is (currently) always the case, but just in case someone accidentally (or willingly) changed, say, some subcycling parameter, better be safe than sorry.

components/eamxx/src/physics/cosp/eamxx_cosp.hpp
```diff
@@ -0,0 +1,77 @@
...
+  AtmosphereProcessType type () const { return AtmosphereProcessType::Physics; }
+
+  // The name of the subcomponent
+  std::string name () const { return "cosp"; }
+
+  // Set the grid
+  void set_grids (const std::shared_ptr<const GridsManager> grids_manager);
+
+  // Scratch space for local variables
+  struct Buffer {
```
bartgol, line 38: Is this needed? If not, I would prune it.

components/eamxx/tests/uncoupled/CMakeLists.txt
```diff
@@ -8,6 +8,7 @@ if (NOT SCREAM_BASELINES_ONLY)
   add_subdirectory(cld_fraction)
   add_subdirectory(spa)
   add_subdirectory(surface_coupling)
+  add_subdirectory(cosp)
```
bartgol, line 11: Shouldn't you put this inside the SCREAM_DOUBLE_PRECISION if, as you did in the src folder?

components/eamxx/tests/uncoupled/cosp/cosp_standalone.cpp
```diff
@@ -0,0 +1,72 @@
...
+  }
+  for (int i=0; i<nsteps; ++i) {
+    ad.run(dt);
+    if (atm_comm.am_i_root()) {
+      std::cout << "  - Iteration " << std::setfill(' ') << std::setw(3) << i+1 << " completed";
+      std::cout << "       [" << std::setfill(' ') << std::setw(3) << 100*(i+1)/nsteps << "%]\n";
+    }
+  }
+
+  // TODO: get the field repo from the driver, and go get (one of)
```
bartgol, line 62: Can you remove these comments? They come from a copy-paste, and may confuse someone...

components/eamxx/tests/uncoupled/cosp/cosp_standalone.cpp
```diff
@@ -0,0 +1,72 @@
...
+  //       the output(s) of SHOC, to check its numerical value (if possible)
+
+  // Finalize 
+  ad.finalize();
+
+  // If we got here, we were able to run shoc
+  REQUIRE(true);
```
bartgol, line 69: I know this is prob done in other tests and that's why it's here, but it really makes no sense to do such a check. We can remove it.

components/eamxx/cime_config/namelist_defaults_scream.xml
```diff
@@ -285,8 +285,14 @@ be lost if SCREAM_HACK_XML is not enabled.
       <number_of_subcycles hgrid="ne0np4_conus_x4v1_lowcon">5</number_of_subcycles>
     </mac_aero_mic>
 
+    <cosp inherit="atm_proc_base">
+      <cosp_subcolumns>10</cosp_subcolumns>
+      <!-- Frequency at which to call COSP; positive values interpreted as number of steps, negative as number of hours -->
```
bartgol, line 290: Can we avoid this CAM-like way of passing parameters? I would rather have something like
```
<cosp_frequency>1</cosp_frequency>
<cosp_frequency_units>nhours</cosp_frequency_units>
```

components/eamxx/src/physics/cosp/cosp_functions.cpp
brhillman: This needs to be removed sorry.

brhillman: Removed.

components/eamxx/src/physics/cosp/cosp_functions.hpp
brhillman: Carryover from when I was grabbing device views from the FM before you told me to just grab host views. Will fix.

brhillman: Fixed.

components/eamxx/tests/uncoupled/cosp/cosp_standalone.cpp
brhillman: Removed.

components/eamxx/tests/uncoupled/cosp/cosp_standalone.cpp
brhillman: Removed.

components/eamxx/src/physics/cosp/cosp_functions.hpp
brhillman: Yeah, the cpp source was already not being used. I added the inline specifier to the function decls, please let me know if this is not what you meant here.

components/eamxx/src/physics/cosp/eamxx_cosp.hpp
brhillman: Removed.

components/eamxx/src/physics/cosp/eamxx_cosp.cpp
brhillman: Added an EKAT_REQUIRE_MSG here for this.

components/eamxx/cime_config/namelist_defaults_scream.xml
brhillman: Yes, I think I have this implemented right now. Currently supported `cosp_frequency_units` are “steps” and “hours” now. I don’t really see needing any additional units, and throw an error if units are not steps or hours.

components/eamxx/src/physics/cosp/eamxx_cosp.cpp
```diff
@@ -78,8 +78,9 @@ void Cosp::set_grids(const std::shared_ptr<const GridsManager> grids_manager)
 // =========================================================================================
 void Cosp::initialize_impl (const RunType /* run_type */)
 {
-  // Determine how often to call COSP, specified as number of atm steps (>0) or number of hours (<0)
+  // Determine how often to call COSP; units can be steps or hours
   m_cosp_frequency = m_params.get<Int>("cosp_frequency", 1);
+  m_cosp_frequency_units = m_params.get<std::string>("cosp_frequency_units", "steps");
```
bartgol, line 83: I would move this in the constructor, and move the check that you added in run_impl (checking that units is either "steps" or "hours") in the constructor too.

The reason is that initialize_impl for atm processes is called toward the end of the init sequence. If the user specified invalid units, we would like to catch that asap. Esp for ne1024 runs, it may save quite a few core hours.

components/eamxx/cime_config/namelist_defaults_scream.xml
```diff
@@ -288,7 +288,8 @@ be lost if SCREAM_HACK_XML is not enabled.
     <cosp inherit="atm_proc_base">
       <cosp_subcolumns>10</cosp_subcolumns>
       <!-- Frequency at which to call COSP; positive values interpreted as number of steps, negative as number of hours -->
-      <cosp_frequency>-1</cosp_frequency>
+      <cosp_frequency>1</cosp_frequency>
+      <cosp_frequency_units>hours</cosp_frequency_units>
```
bartgol, line 292: Related to the comment below. I think you should take advantage of the XML attribute "valid_values", so that any invalid value will be caught during atmchange, rather than runtime. Here's an example of how to use it
```
 <atm_log_level valid_values="trace,debug,info,warn,error">info</atm_log_level>
```

bartgol (changes requested):

components/eamxx/src/physics/cosp/eamxx_cosp.cpp
brhillman: Is there any reason not to put ALL the param reads into the constructor then?

components/eamxx/cime_config/namelist_defaults_scream.xml
brhillman: Good call. Done.

bartgol (dismissed):

bartgol (approved):

bartgol (dismissed):

brhillman:
@bartgol mind taking a quick look for approval since this is passing AT now?

bartgol (approved):

PR #2436, opened 2023-08-01
Update set_variable_metadata, add a get_variable_metadata

AaronDonahue:
This commit extends the set_variable_metadata function to accept Real valued input, and introduces a new  get_variable_metadata function.

Both are needed to allow users to set things like a `_FillValue` as variable metadata, and if needed be able to query that value from a source data file.


AaronDonahue:
@bartgol this should be ready for review

bartgol (approved):

PR #2440, opened 2023-08-03
Fix bug when writing rhist files

### LINKED ISSUE #2438: Monthly average files are not computed correctly
In our recent Cess control simulation, the monthly files look as if the monthly-averaged output are just averages from the last X-days of simulation before the monthly output file is produced. Note that the spatial structure looks reasonable, but the Temperature values (which should be on order of 280K) are just very small:
image

It had originally seemed that the issue was fixed from merging master into the machines/frontier branch, but it seems to still exist in this simulation. 
https://github.com/E3SM-Project/scream/issues/2391

The run directory of this simulation is: 
```
/lustre/orion/cli115/proj-shared/terai/e3sm_scratch/cess-v1-cntl.ne1024pg2_ne1024pg2.F2010-SCREAMv1.20230721/run/
```

It was run with the https://github.com/E3SM-Project/scream/tree/simulations/cess-v1 branch, which merged master in July 19, '23: https://github.com/E3SM-Project/scream/commit/587f02628a5ac36508c851c1a3060d90ec49a244

bartgol:
This pr fixes (hopefully) a bug related to incorrect writing of rhist files. The issue boils down to the fact that output manager was not correctly detecting that a full checkpoint (meaning not just metadata, but also variables values) is needed in case of output freq=1 BUT freq units different from "nsteps". This PR fixes that, and forces a full checkpoint step if units!=nsteps or freq>1. Notice that, in theory, something like units="nhours" and freq=1 may still mean "output at every model step", if the model dt is 3600s. However, that's a corner case that will probably only show up in unit tests, so writing variable data in rhist even if not strictly necessary should not be too bad, considering that tests are usually at low resolutions.

Fixes #2438 

@crterai Could you try to merge this branch in your Cess experiments branch and verify that it fixes the issue, before we merge it?

### DIFF
--- a/components/eamxx/src/share/io/scream_output_manager.cpp
+++ b/components/eamxx/src/share/io/scream_output_manager.cpp
@@ -331,10 +331,17 @@ void OutputManager::run(const util::TimeStamp& timestamp)
   //       Since we *always* write a history restart file, we can have a non-full checkpoint, if the average
   //       type is Instant and/or the frequency is every step. A non-full checkpoint will simply write some
   //       global attribute, such as the time of last write.
+  //       Also, notice that units="nhours" and freq=1 would still output evey step if dt=3600s. However,
+  //       it is somewhat hard to figure out if output happens every step, without having a dt to compare
+  //       against. Therefore, we simply assume that if units!=nsteps OR freq>1, then we don't output every
+  //       timestep. If, in fact, we are outputing every timestep, it's likely a small test, so it's not too
+  //       bad if we write out some extra data.
+  const bool output_every_step       = m_output_control.frequency_units=="nsteps" &&
+                                       m_output_control.frequency==1;
   const bool is_t0_output            = timestamp==m_case_t0;
   const bool is_output_step          = m_output_control.is_write_step(timestamp) || is_t0_output;
   const bool is_checkpoint_step      = m_checkpoint_control.is_write_step(timestamp) && not is_t0_output;
-  const bool has_checkpoint_data     = (m_avg_type!=OutputAvgType::Instant && m_output_control.frequency>1);
+  const bool has_checkpoint_data     = m_avg_type!=OutputAvgType::Instant && not output_every_step;
   const bool is_full_checkpoint_step = is_checkpoint_step && has_checkpoint_data && not is_output_step;
   const bool is_write_step           = is_output_step || is_checkpoint_step;
 
@@ -741,9 +748,6 @@ push_to_logger()
   m_atm_logger->info("      Includes Grid Data ?: " + bool_to_string(m_output_file_specs.save_grid_data));
   // List each GRID - TODO
   // List all FIELDS - TODO
-
-
-
 }
 
 } // namespace scream


crterai:
Yup, I can cherry-pick this change to the `simulations/cess-v1` branch. Will need to test at low-resolution, since getting a month at ne1024 takes time.

AaronDonahue (dismissed):
I wonder if we could have the check compare the `freq*freq_units` converted to seconds and divide it by the `dt`.  That should avoid the edge case you mentioned in the PR description.  I'm happy with this PR as it stands now as it should fix the issue we are seeing in the CESS runs.

crterai:
I applied this fix and tried the ne30 test and it still gave me some odd behaviors, where the monthly average output looks to be still smaller than what it should be:
image
The case is here:
```
/lustre/orion/cli115/proj-shared/terai/e3sm_scratch/t.20230711.F2010-SCREAMv1.ne30pg2_ne30pg2.frontier-scream-gpu.n001t8x1.vth200.test_writing_monthly/run
```
And used the most up-to-date version of this branch (`323efd91e8`):
https://github.com/E3SM-Project/scream/tree/simulations/cess-v1

crterai:
Follow up from previous comment: 
The ne30 case that I ran did produce non-small rhist files, so i don't think it's a matter of not writing rhist files:
```
[terai@login14.frontier run]$ ls -lah output.scream.Cess.monthly_*
-rwxr-xr-t 1 terai cli115 4.5M Aug  3 21:44 output.scream.Cess.monthly_COSP_ne1024.AVERAGE.nmonths_x1.0001-01-01-00000.nc
-rwxr-xr-t 1 terai cli115 8.5M Aug  3 20:59 output.scream.Cess.monthly_COSP_ne1024.rhist.AVERAGE.ndays_x11.0001-01-12-00000.nc
-rwxr-xr-t 1 terai cli115 8.5M Aug  3 21:22 output.scream.Cess.monthly_COSP_ne1024.rhist.AVERAGE.ndays_x11.0001-01-23-00000.nc
-rwxr-xr-t 1 terai cli115 8.5M Aug  3 21:44 output.scream.Cess.monthly_COSP_ne1024.rhist.AVERAGE.ndays_x11.0001-02-03-00000.nc
-rwxr-xr-t 1 terai cli115 204M Aug  3 21:44 output.scream.Cess.monthly_ne1024.AVERAGE.nmonths_x1.0001-01-01-00000.nc
-rwxr-xr-t 1 terai cli115 407M Aug  3 20:59 output.scream.Cess.monthly_ne1024.rhist.AVERAGE.ndays_x11.0001-01-12-00000.nc
-rwxr-xr-t 1 terai cli115 407M Aug  3 21:22 output.scream.Cess.monthly_ne1024.rhist.AVERAGE.ndays_x11.0001-01-23-00000.nc
-rwxr-xr-t 1 terai cli115 407M Aug  3 21:44 output.scream.Cess.monthly_ne1024.rhist.AVERAGE.ndays_x11.0001-02-03-00000.nc
```

bartgol:
> I wonder if we could have the check compare the `freq*freq_units` converted to seconds and divide it by the `dt`. That should avoid the edge case you mentioned in the PR description. I'm happy with this PR as it stands now as it should fix the issue we are seeing in the CESS runs.

The output manager does not know the timestep.

bartgol:
Ugh, so there were 2+ issues. This PR still does fix one of them, but not the whole thing. I admit that I did not verify on the standalone pb that the avg came up correctly. I was convinced that the issue was solely in the rhist file, so when I saw that it came out correctly, I called it a day. I will go back on the standalone pb for debugging; hopefully that will exhibit the same bug.

bartgol:
Actually, it was fairly easy to track it down. Turns out the original commit I pushed fixed only one of the two places that only checked freq=1 (and not also freq units). Fixing the other as well makes things work correctly on the standalone test.

I took the liberty to fix another couple issues, one having to do with restart file timestamp, and the other with a red herring error msg from homme, which messed up stack unwinding while handling an exception thrown.

components/eamxx/src/dynamics/homme/interface/homme_driver_mod.F90
```diff
@@ -242,11 +242,16 @@ subroutine prim_run_f90 (nsplit_iteration) bind(c)
   end subroutine prim_run_f90
 
   subroutine prim_finalize_f90 () bind(c)
-    use homme_context_mod,    only: is_model_inited, elem, dom_mt, close_homme_log
+    use homme_context_mod,    only: is_model_inited, elem, par, dom_mt, close_homme_log
     use prim_cxx_driver_base, only: prim_finalize
 
     if (.not. is_model_inited) then
-      call abortmp ("Error! prim_init_model_f90 was not called yet (or prim_finalize_f90 was already called).\n")
+      if (par%masterproc) then
```
bartgol, line 249: @brhillman @ndkeen this should fix the red herring error message you (and probably others) observed. SInce the error happened during the stack unwinding while handling a thrown exception, we were completely hiding the actual error message. I verified that this fix brings back error messages.

bartgol:
@crterai when you have a chance, can you try again with this new version of the branch?

crterai:
It took a bit of time for the test sim to get through the queue, but I can confirm that the output after applying the most recent commits looks good. 
image
Thanks, @bartgol! 

For reference, the case that worked can be found here:
```
/lustre/orion/cli115/proj-shared/terai/e3sm_scratch/t.20230711.F2010-SCREAMv1.ne30pg2_ne30pg2.frontier-scream-gpu.n001t8x1.vth200.test_writing_monthly_try3
```

bartgol:
Great! If Aaron has no further comment, I think this can be merged.

AaronDonahue (approved):
Looks good!! Thanks for fixing the bug @bartgol !

PR #2441, opened 2023-08-03
Default energy fixer to off

tcclevenger:
Some issues were raised by @ambrad and @oksanaguba about the PR https://github.com/E3SM-Project/scream/pull/2418 that need to be investigated. For now, easiest to just default off for CIME runs until we are confident in the computation.


ambrad (approved):
Sure, turning it off for now makes sense.

tcclevenger:
CIME baselines haven't been blessed, so no need to test.

oksanaguba:
just a minor comment -- this PR is not for an energy fixer, this PR is to turn off new CFLX limiter by default.

PR #2442, opened 2023-08-05
Add implementaion for ML correction 

elynnwu:
ML correction is now a valid physics process- opens a pre-trained ML model, predicts the nudging tendencies, add them back to the state variables. It is currently hard-corded to apply to `T_mid` and `qv`. We should make this more flexible and allow any combination of [`T_mid`, `qv`, `u`, `v`]. Another shortcoming now is that we need to open the ML model per timestep on all processors, this might prove to be a performance bottleneck. We may want to change this to open during init time on one node and distribute it to all others.

In order to run this process bare-metal on quartz, we have to use custom `hdf5` library because `tensorflow` uses `h5py` which is built against a different version of `hdf5` available on quartz. Custom modulefiles are added in  `/usr/gdata/climdat/install/quartz/modulefiles`. The python packages required to run ML correction are installed as part of an virtual environment in `/usr/WS1/climdat/python_venv/3.9.2/screamML`.


AaronDonahue:
Perhaps @bartgol , @ambrad or @jeff-cohere have thoughts on how to address your comment:
```
Another shortcoming now is that we need to open the ML model per timestep on all processors, this might prove to be a performance bottleneck. We may want to change this to open during init time on one node and distribute it to all others.
```

bartgol:
Well, I'm not sure what "opening the ML model" entails. If it requires some non-trivial initialization that is independent of the current state/timestep, then yes, I would move that to init time. I would recommend adding some timers around the "opening" phase, and see what the cost is on an ne30 case.

jeff-cohere:
It sounds like the "ML Model" might be a data file, in which case I would second @elynnwu 's recommendation for opening it only on one rank and broadcasting it to the others, to alleviate pressure on the file system. And if the file doesn't change for the duration of a simulation, this would ideally be done at init time instead of every timestep.

elynnwu:
Yes, opening the ML model is independent of the current time step, it does take in the current state but the model is constant throughout the simulation. I think it's a good idea to add some timer to understand the overhead.

@bartgol and @jeff-cohere, do you think the performance optimization is necessary for this PR? I'm hoping to get this PR through because the unit test has been updated significantly from master so would like this to be tested as part of CI asap, and the optimization can go in subsequent PRs.

jeff-cohere:
Personally, I think it's okay to save the optimizations for a later PR, but I'll defer to @bartgol on this one.

bartgol:
Yeah, no need to optimize right away.

components/eamxx/cime_config/namelist_defaults_scream.xml
```diff
@@ -209,6 +209,12 @@ be lost if SCREAM_HACK_XML is not enabled.
     <!-- CLD fraction -->
     <cldFraction inherit="atm_proc_base"/>
 
+    <!-- ML correction -->
+    <mlcorrection inherit="atm_proc_base">
+      <ML_model_path type="array(string)">NONE</ML_model_path>
```
bartgol, line 214: Is `array(string)` intended? "model_path" makes me think this is just a string...

components/eamxx/src/physics/ml_correction/CMakeLists.txt
```diff
@@ -5,8 +5,22 @@ set(MLCORRECTION_SRCS
...
+endif()
+find_package(Python REQUIRED COMPONENTS Interpreter Development)
 
 add_library(ml_correction ${MLCORRECTION_SRCS})
+target_compile_definitions(ml_correction PUBLIC EAMXX_HAS_ML_CORRECTION)
+target_compile_definitions(ml_correction PRIVATE -DML_CORRECTION_CUSTOM_PATH="${CMAKE_CURRENT_SOURCE_DIR}")
 target_include_directories(ml_correction PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/../share)
-target_link_libraries(ml_correction physics_share scream_share)
+target_include_directories(ml_correction SYSTEM PRIVATE ${PYTHON_INCLUDE_DIRS})
+target_link_libraries(ml_correction physics_share scream_share pybind11::pybind11 Python::Python)
```
bartgol, line 25: since we do it everywhere else, you should probably declare the link to be PUBLIC, for consistency in cmake style.

components/eamxx/src/physics/ml_correction/CMakeLists.txt
```diff
@@ -5,8 +5,22 @@ set(MLCORRECTION_SRCS
...
 target_compile_options(ml_correction PUBLIC)
```
bartgol, line 26: This line is pointless, and should be removed.

components/eamxx/src/physics/ml_correction/eamxx_ml_correction_process_interface.cpp
```diff
@@ -1,15 +1,21 @@
...
 #include "ekat/util/ekat_units.hpp"
+#include "share/field/field_utils.hpp"
 
 namespace scream {
 // =========================================================================================
 MLCorrection::MLCorrection(const ekat::Comm &comm,
                            const ekat::ParameterList &params)
     : AtmosphereProcess(comm, params) {
-  // Nothing to do here
+  m_ML_model_path = m_params.get<std::string>("ML_model_path");
```
bartgol, line 17: This confirms my comment on the XML file, that this param should be marked as `string` in the XML, and not `array(string)`.

components/eamxx/src/physics/ml_correction/eamxx_ml_correction_process_interface.cpp
```diff
@@ -22,40 +28,90 @@ void MLCorrection::set_grids(
   // Nevertheless, for output reasons, we like to see 'kg/kg'.
   auto Q = kg / kg;
   Q.set_string("kg/kg");
-
+  constexpr int ps = SCREAM_SMALL_PACK_SIZE;
```
bartgol, line 31: I'm curious, why do you pick the small pack size rather than the default one?

components/eamxx/src/physics/ml_correction/eamxx_ml_correction_process_interface.cpp
```diff
@@ -22,40 +28,90 @@ void MLCorrection::set_grids(
...
-  // - There are no fields used as both input and output.
+  /* ----------------------- WARNING --------------------------------*/
+  /* The following is a HACK to get things moving, we don't want to
+   * add all fields as "updated" long-term.  A separate stream of work
+   * is adapting the infrastructure to allow for a generic "add_field" call
+   * to be used here which we can then setup using the m_fields_ml_output_variables variable
+   */
+  add_field<Updated>("T_mid", scalar3d_layout_mid, K, grid_name, ps);
+  add_field<Updated>("qv",    scalar3d_layout_mid, Q, grid_name, "tracers", ps);
+  add_field<Updated>("u",     scalar3d_layout_mid, m/s, grid_name, ps);
```
bartgol, line 51: This may not work as you indended. EAMxx can only create the u/v fields _after_ all fields have been created, by slicing the `horiz_wind` vector field. But it calls them `U` and `V` (uppercase). The two fields you declare here are two separate fields, independent of `horiz_winds` (or `U` and `V`). However, it would still not work if you used `U` and `V`, with the AD that will crap out (I can explain why offline).

If you want to grab the two components of horiz_winds separately, you need to request `horiz_winds`, and then simply grab the two components in `run_impl`:
```
const auto& u = get_field_out("horiz_winds").get_component(0);
const auto& v = get_field_out("horiz_winds").get_component(1);
```

components/eamxx/src/physics/ml_correction/eamxx_ml_correction_process_interface.cpp
```diff
@@ -22,40 +28,90 @@ void MLCorrection::set_grids(
...
+  const auto &qv       = qv_field.get_view<Real **, Host>();
+  const auto &T_mid_field = get_field_out("T_mid");
+  const auto &T_mid       = T_mid_field.get_view<Real **, Host>();
+  const auto &u_field = get_field_out("u");
+  const auto &u       = u_field.get_view<Real **, Host>();
+  const auto &v_field = get_field_out("v");
+  const auto &v       = v_field.get_view<Real **, Host>();
+
+  // having m_lat and m_lon in set_grids breaks the standalone unit test
+  m_lat  = m_grid->get_geometry_data("lat");
```
bartgol, line 77: The geometry data fields should really be grabbed just once during `set_grids`.

components/eamxx/src/physics/ml_correction/eamxx_ml_correction_process_interface.hpp
```diff
@@ -18,14 +25,15 @@ namespace scream {
 
 class MLCorrection : public AtmosphereProcess {
  public:
+  using mPack = ekat::Pack<Real,SCREAM_PACK_SIZE>;
```
bartgol, line 28: This type alias is unused, and can be removed.

components/eamxx/cime_config/namelist_defaults_scream.xml
elynnwu: Yes you're right, this should be just a string.

components/eamxx/src/physics/ml_correction/eamxx_ml_correction_process_interface.cpp
elynnwu: Frankly no reason, I only learned about pack size recently. This is also used in the unit test, should it be the default pack size?

components/eamxx/src/physics/ml_correction/eamxx_ml_correction_process_interface.cpp
elynnwu: Okay, I was following what Aaron did in this [PR](https://github.com/E3SM-Project/scream/pull/2378), but admittedly I haven't actually tried ML correcting `u` and `v`. Maybe this explains why the nudging tendencies for `u` and `v` seem off. Pinging you and Aaron offline for this.

components/eamxx/src/physics/ml_correction/eamxx_ml_correction_process_interface.cpp
elynnwu: Is there an example of this working alongside the unit test using mesh free grid? I originally had it in `set_grids` but it was breaking the unit test so I had to move it.

components/eamxx/src/physics/ml_correction/CMakeLists.txt
elynnwu: Do you mean to say this line?
```
target_compile_options(ml_correction PUBLIC)
```

Because I think line 25 is required.

bartgol: Yes, that's the line I meant. It wasn't part of your changes, but I figured we could remove it while at it...

components/eamxx/src/physics/ml_correction/eamxx_ml_correction_process_interface.cpp
bartgol: In practice, it probably does not matter, since we default SMALL_PACK_SIZE=PACK_SIZE, but it does look a bit confusing. I think at the beginning we had a small pack size, which was more suitable for stuff with lots of conditionals (like microphysics). But unless you have lots of if statements concerning the vertical direction, you should use the largest pack size.

components/eamxx/src/physics/ml_correction/eamxx_ml_correction_process_interface.cpp
bartgol: Odd. The (standalone) unit tests do have some peculiarity, since they don't have dynamics building the grids. But if the geo data was set at all in the grid, then it should already be available during set_grid. In other words, if it failed there, it should fail here too...

elynnwu: This is the error when I move these lines to `set_grids`:
```
  abstract_grid.cpp:173: FAIL:
  has_geometry_data(name)
  Error! Geometry data 'lat' not found.
```

components/eamxx/src/physics/ml_correction/ml_correction.py
```diff
@@ -0,0 +1,51 @@
...
+from scream_run.steppers.machine_learning import (
+    MachineLearningConfig,
+    open_model,
+    predict,
+)
+
+
+def get_ML_correction(model_path, T_mid, qv, cos_zenith, dt):
+    config = MachineLearningConfig(models=[model_path])
+    model = open_model(config)
```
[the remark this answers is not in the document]
elynnwu: Thanks for the tips! I will look into this for the next step.

elynnwu:
>     * It seems the underlying model architecture, model weights, inference calculation, and other details are hidden in a separate python package, correct? Is that publicly available for review or still not released? It may be helpful to see that for potential issues/optimization (obviously it doesn't need to happen now)

Yes, the python package is publicly available in a separate repo [here](https://github.com/ai2cm/fv3net). Specific details for handling scream ML correction is under a subpackage [here](https://github.com/ai2cm/fv3net/tree/master/workflows/prognostic_scream_run).

>     * You mention this is a TensorFlow model (hence the need for different HDF library). In my experience, there are two things to a trained model in TensorFlow --- the model architecture (usually light-ish) and model weights (usually where the true value is). Is that your case too? Or is the model architecture itself heavy? One can define and redefine the model (and its layers) in-python-code without having to (re)load it (but obviously the weights will have to be loaded). Potential optimization here includes sharding and caching the weights (not sure if they optimized this as much as they optimized caching and tfrecords for training/input data... my experience is from ~1.5 years ago)

Yes, this is using a TensorFlow model. The saved model has the architecture and model weights. I agree that we should be able to cache the weights in scream, I just haven't figured out a good way to do this.

>     * I suspect eventually the back-and-forth between C++ (EAMxx), Python (TensorFlow wrapper), and C/C++ (libtensorflow) may be the area where most of the overhead will be. Have you considered using libtensorflow directly? Most people only use the Python API of TensorFlow (understandably) and so there may not be good docs on this, but I can help you get the C++ library in isolation if you'd like it, or help you build an optimized version of it for whatever architecture mix where this is being run. As a side note: This is one thing I learned the hard way in the past: Using custom-optimized versions of packages like TensorFlow can increase performance significantly

Yes, initially we considered using the C/C++ API directly. That is still a route we can explore. The reason we choose to go through Python is because our group has spent quite a bit of time building up the process to do ML correction - including things like ensuring the model levels are aligned during training vs. online, sanity checks on physical constraints (qv > 0), etc. We also believe it's better to do training/validation/online in the same setup, which is less prone to errors. However, if it's still too slow in the end of our python/C++ optimization attempt, we can certainly port those functionality to scream itself directly and skip the python layer.

components/eamxx/src/mct_coupling/CMakeLists.txt
```diff
@@ -39,6 +39,7 @@ set (SCREAM_LIBS
      spa
      nudging
      diagnostics
+     ml_correction
```
AaronDonahue, line 42: Does this need to only happen if we are on `quartz`?  Could this be the cause of the fails on `mappy`?

elynnwu: Hmm, but I need this right? But should it have worked on weaver?

bartgol: Weaver does not build the coupled model, it only does standalone testing.

bartgol: If you want to merge this in master, I would strongly encourage you to make this work on supported E3SM machines (at least on chrysalis). If this only works on an internal LLNL machines, then it should probably remain in a branch...

bartgol (approved):

PR #2444, opened 2023-08-08
Chrysalis: Exclude chr-0512 for one test.

ambrad:
A good lead in my nondeterminism hunt on Chrysalis is that node chr-0512 produces the failure very frequently, while other nodes I have tried so far never do. Since the small_kernels test happens to fail the most, probably because of how nodes get allocated to the tests, I want to see what happens when it isn't permitted to use this node.


ambrad:
@jgfouca please don't merge this. I'll merge it after manual testing is done.

jgfouca (approved):

ambrad:
Chrysalis runs with the desired line, while Ascent runs with no batch-line modification, as desired.

PeterCaldwell:
@ambrad - my recollection is that asking the scheduler to run a job on a nonstandard set of nodes like this often results in very long queue wait times. Will that be a problem here?

ambrad:
No. This one test will just wait a bit until debug-queue nodes free up, if that's even necessary. It's giving up just 1 of 20 nodes and uses 4. To find this issue, I've been running repeat testing on a one-node test using --nodelist with exactly one node specified, and I've gotten through many tens of runs with little wait time.

PR #2445, opened 2023-08-10
Fixes a segfault in the finalization of F90 SCORPIO system in certain cases

jeff-cohere:
**EDIT: Only the second bullet still applies for this PR. See discussion below.**

These changes prevent exceptions from being swallowed in cases where the model has not been completely initialized. Specifically:

* AtmosphereProcess objects now only run their finalization if they have been initialized
* The SCREAM SCORPIO interface is a bit more careful about how it finalizes itself (as a result of the first refinement).

I've been trying to debug a standalone coupled HOMME-MAM4xx test that needs some extra fields in its initial conditions, and these changes have made it much easier to figure out which fields are needed, so I hope this can help others stand up new AtmosphereProcess subclasses.

### DIFF
--- a/components/eamxx/src/share/io/scream_scorpio_interface.F90
+++ b/components/eamxx/src/share/io/scream_scorpio_interface.F90
@@ -890,11 +890,11 @@ contains
             if (associated(var%iodesc_list)) then
               ! Remove this variable as a customer of the associated iodesc
               var%iodesc_list%num_customers = var%iodesc_list%num_customers - 1
-              ! Dellocate select memory from this variable.  Note we can't just
-              ! deallocate the whole var structure because this would also
-              ! deallocate the iodesc_list.
-              call deallocate_hist_var_t(var)
             end if ! associated(var%iodesc_list)
+            ! Dellocate select memory from this variable.  Note we can't just
+            ! deallocate the whole var structure because this would also
+            ! deallocate the iodesc_list.
+            call deallocate_hist_var_t(var)
           end if ! associated(var)
           curr_var_list => curr_var_list%next  ! Move on to the next variable
         end do ! associated(curr_var_list)
--- a/components/eamxx/src/share/io/scream_scorpio_interface.F90
+++ b/components/eamxx/src/share/io/scream_scorpio_interface.F90
@@ -887,12 +887,14 @@ contains
         do while (associated(curr_var_list))
           var => curr_var_list%var  ! The actual variable pointer
           if (associated(var)) then
-            ! Remove this variable as a customer of the associated iodesc
-            var%iodesc_list%num_customers = var%iodesc_list%num_customers - 1
-            ! Dellocate select memory from this variable.  Note we can't just
-            ! deallocate the whole var structure because this would also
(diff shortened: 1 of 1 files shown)


AaronDonahue (approved):
looks good to me, I'll defer to @bartgol if regarding final approval though.

bartgol:
Follow up: if some atm procs do need to perform some clean-up ops _only if initialize has been called_, then I think those process should have an internal flag, and handle things accordingly in their finalize method. I don't think the base class should handle whether or not finalize should be called. E.g., some atm procs _may_ perform some initialization during construction, and additional setup during initialize (when fields have been created/inited). In the finalization, it may have to _always_ do something (like undo whatever the ctor did), and some additional cleanup if initialize had been called.

bartgol:
> Fair enough. In this case, HOMME's `AtmosphereProcess` has some finalization in Fortran that complains when it hasn't been initialized, and in doing so it calls `abort`, which wipes out any exceptions that were being propagated. Do you want me to add some logic to HOMME to prevent this from happening and remove the base class checks that I put in?

Ah, I think I fixed that in #2440 . See [this](https://github.com/E3SM-Project/scream/pull/2440/files#r1284883152) comment. It was indeed annoying, but it should be fixed now.

jeff-cohere:
Oh, awesome, thanks! Let me make sure I have that commit in my branch and try it without my fix.

jeff-cohere:
Okay, the error does get propagated, but the changes to `scream_scorpio_interface.F90` still seem necessary, since my test now segfaults when it tries to access a field in a history variable. Should I revert the changes to `AtmosphereProcess` and keep these ones?

bartgol:
Yeah, you can use the force, if you want, to avoid X and revert X in the history...

Btw: the annoying (and to me, sometimes ununderstandable) issues with memory management in the F90 scorpio interface are the main reason behind #2361. I had a similar error related to some f90 var not being associated, but I never figured out whether it was a real issue, or simply a by-product of an early termination due to an upstream error...

jeff-cohere:
Okay, I used my force powers!

components/eamxx/src/share/io/scream_scorpio_interface.F90
```diff
@@ -887,12 +887,14 @@ subroutine eam_pio_closefile(fname)
...
-            ! deallocate the whole var structure because this would also
-            ! deallocate the iodesc_list.
-            call deallocate_hist_var_t(var)
+            if (associated(var%iodesc_list)) then
+              ! Remove this variable as a customer of the associated iodesc
+              var%iodesc_list%num_customers = var%iodesc_list%num_customers - 1
+              ! Dellocate select memory from this variable.  Note we can't just
+              ! deallocate the whole var structure because this would also
+              ! deallocate the iodesc_list.
+              call deallocate_hist_var_t(var)
```
bartgol, line 896: Shouldn't this deallocation be moved outside of the if block?

jeff-cohere: Ah, probably so, yes! One moment, please.

bartgol (approved):

PR #2446, opened 2023-08-10
add select AeroCOM cloud-top diagnsotics in RRTMGP

### LINKED ISSUE #2394: Add functionality to output cloud droplet number at cloud top
To analyze albedo susceptibilities to aerosol and calculate other ACI-related diagnostics, we (@mahf708, @jmuelmen, and myself) would like EAMxx to produce a CDNC diagnostic variable similar to that of Aerocom. This variable should provide the average cloud droplet number concentration in the first liquid cloud level (from the model top) that contains "sufficient" total cloud water (above a pre-defined threshold).  The rationale is to simulate the cloud droplet number that would be retrieved from a satellite such as MODIS.

The calculation of this diagnostic can be simplified in EAMxx (as long as we assume we’re running at 3km) compared to how it is done in EAM.  This is because at 3km, we can, to a first approximation, ignore some of the complexities related to the interpretation of cloud fractions when comparing MODIS data with the model in coarse resolution models.

The algorithm to calculate cloud-top CDNC can then be simplified to the following (the calculation is done per column):

- Iterate through model levels beginning with the second level from the top (the Aerocom procedure in EAM assumes that there is no cloud in the top level)

- The diagnostic CDNC equals the cloud droplet number concentration (grid-box mean) in the “stratocumulus cloud-top” level.  When we reach a level where three criteria are all met, we call this the “cloud top”.    The criteria are:
    1. No cloud ice is present
    2. Cloud liquid cloud water content exceeds a pre-defined threshold value
    3. Liquid cloud fraction exceeds a pre-defined threshold value

- After reaching the cloud top, exit the iteration (the computation in the coarse-resolution model also looks at levels below the cloud top and estimates the additional cloud fraction that a satellite would see there, but we can ignore that at 3km resolution).

The pre-defined threshold values have not yet been determined, but we can provide some placeholder values.  It might be useful to implement them as parameters that are accessible through the namelist.

mahf708:
Adds select AeroCOM cloud-top diagnostics inside the RRTMGP interface.

For now, I selected a few key diagnostics (3 different flavors) and we can expand this to any parameter of interest. Additionally, for now, this implementation is the "vanilla" aeroCOM recommendation, and we likely need to adjust it for our needs. This PR is to register the intention to start iterating on this and build from here.

Fixes #2394

### DIFF
--- a/components/eamxx/src/physics/rrtmgp/scream_rrtmgp_interface.cpp
+++ b/components/eamxx/src/physics/rrtmgp/scream_rrtmgp_interface.cpp
@@ -912,9 +912,9 @@ namespace scream {
                 A.  and if we assume randomn overalp, we can write pp(k+1) =
                     pp(k) * (1-cld(k+1)),
                 B.  but if we assume maximum overlap, we can write pp(k+1) =
-                    pp(k) * (1-max(cld(k+1),cld(k)),
+                    pp(k) * (1-max(cld(k+1),1-pp(k)),
                 C.  instead, we can assume maximum-random overlap, pp(k+1) =
-                    pp(k) * (1-max(cld(k+1),cld(k))) / (1-cld(k)),
+                    pp(k) * (1-max(cld(k+1),1-pp(k))) / pp(k),
                 which we use. We use a minimum in the denominator to avoid
                 division by 0. The maximum-random overlap assumption is designed to form
                 maximum overlap for contiguous cloudy layers and random overlap
@@ -925,9 +925,9 @@ namespace scream {
                 B.  consider the case of three contiguous cloudy layers after the first,
                     where cld(2) < cld(3) > cld(4) (i.e., the middle layer is highest).
                     pp(1) = 1 (by assumption, also cld(1) = 0)
-                    pp(2) = pp(1) * (1-cld(2))/(1)        = 1-cld(2)
-                    pp(3) = pp(2) * (1-cld(3))/(1-cld(2)) = 1-cld(3)
-                    pp(4) = pp(3) * (1-cld(3))/(1-cld(3)) = 1-cld(3)
+                    pp(2) = pp(1) * (1-cld(2))/pp(1) = 1-cld(2)
+                    pp(3) = pp(2) * (1-cld(3))/pp(2) = 1-cld(3)
+                    pp(4) = pp(3) * pp(3)/pp(3)      = 1-cld(3)
                     Effectively, the denominator in C removes the randomness
                     introduced in A because we do not want it in the case of contiguous
                     cloudy layers, instead simplifying to B. On the other hand,
@@ -1009,9 +1009,9 @@ namespace scream {
                        * minimum in the denominator is to avoid division by 0 */
                       aerocom_tmp(icol) =
                           aerocom_clr(icol) *
-                          (1 - ekat::impl::max(aerocom_cld(icol, ilay - 1),
+                          (1 - ekat::impl::max(1-aerocom_clr(icol),
                                                aerocom_cld(icol, ilay))) /
-                          (1 - ekat::impl::min(aerocom_cld(icol, ilay - 1),
+                          (1 - ekat::impl::min(1-aerocom_clr(icol),
                                                1 - 0.001));  // BAD_CONSTANT!
                       /* PART II: The inferred properties */
                       /* T_mid and p_mid are the "other properties" and so we
--- a/components/eamxx/src/physics/rrtmgp/tests/rrtmgp_unit_tests.cpp
+++ b/components/eamxx/src/physics/rrtmgp/tests/rrtmgp_unit_tests.cpp
@@ -696,6 +696,20 @@ TEST_CASE("rrtmgp_aerocom_cloudtop") {
         cldfrac_liq_at_cldtop, cldfrac_tot_at_cldtop, nc_at_cldtop);
     REQUIRE(cldfrac_tot_at_cldtop.createHostCopy()(1) == .7);
 
+    // Case 3xtra: test max overlap (sneaky bug)
+    // This case should still produce 0.7 because it is all contiguous
+    yakl::fortran::parallel_for(
+        1, YAKL_LAMBDA(int /* dummy */) {
+          cldfrac_tot(1, 5) = 0.3;
+          cldfrac_tot(1, 6) = 0.4;
+          cldfrac_tot(1, 7) = 0.2;
+        });
+    scream::rrtmgp::compute_aerocom_cloudtop(
+        ncol, nlay, tmid, pmid, p_del, z_del, qc, qi, cldfrac_tot, nc,
+        tmid_at_cldtop, pmid_at_cldtop, cldfrac_ice_at_cldtop,
+        cldfrac_liq_at_cldtop, cldfrac_tot_at_cldtop, nc_at_cldtop);
+    REQUIRE(cldfrac_tot_at_cldtop.createHostCopy()(1) == .7);
+
     // Case 4: test random overlap (if non-contiguous cloudy layers, then
     // random)
     yakl::fortran::parallel_for(


components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
```diff
@@ -135,6 +137,13 @@ void RRTMGPRadiation::set_grids(const std::shared_ptr<const GridsManager> grids_
   add_field<Computed>("cldmed"        , scalar2d_layout, nondim, grid_name, "RESTART");
   add_field<Computed>("cldhgh"        , scalar2d_layout, nondim, grid_name, "RESTART");
   add_field<Computed>("cldtot"        , scalar2d_layout, nondim, grid_name, "RESTART");
+  // Cloud-top diagnostics following AeroCOM recommendation
+  add_field<Computed>("aerocom_ttop"  , scalar2d_layout, nondim, grid_name, "RESTART");
```
bartgol, line 141: Are these really needed in the restart file? Their calculation below seems to be indep of their previous value...

components/eamxx/src/physics/rrtmgp/scream_rrtmgp_interface.cpp
```diff
@@ -878,5 +879,79 @@ namespace scream {
...
+            using physconst = scream::physics::Constants<Real>;
+            // Loop over all columns in parallel
+            yakl::fortran::parallel_for(
+                SimpleBounds<1>(ncol), YAKL_LAMBDA(int icol) {
+                  // Loops over all layers in serial (due to accumulative
+                  // product) Start at 2 (second highest) layer because the
+                  // highest one has no clouds
+                  for(int ilay = 2; ilay <= nlay; ++ilay) {
+                    // We only do the calculation if certain conditions are met
+                    if(qc(icol, ilay) + qi(icol, ilay) > 0 &&  // BAD_CONSTANT!
```
jeff-cohere, line 908: What's the bad constant in the above line? Are you saying `qc + qi > 0` is a heuristic constraint for deciding whether to perform the calculation?

jeff-cohere (commented):
Looks good so far! I'll give it a proper review when you add your unit test.

components/eamxx/src/physics/rrtmgp/scream_rrtmgp_interface.cpp
brhillman: You might consider using variables with descriptive names for these thresholds instead of the magic numbers to make this a little more obvious.

components/eamxx/src/physics/rrtmgp/scream_rrtmgp_interface.hpp
```diff
@@ -112,6 +112,16 @@ namespace scream {
         void compute_cloud_area(
                 int ncol, int nlay, int ngpt, Real pmin, Real pmax,
                 const real2d& pmid, const real3d& cld_tau_gpt, real1d& cld_area);
+        /*
```
brhillman, line 115: This is probably fine for now, but I think this might be where my code is a bad example to follow. It looks like your diagnostics don’t use anything inherently tied to RRTMGP, and thus to YAKL, so it would probably be preferable to implement this in Kokkos and keep separate from the `scream_rrtmgp_interface` code. I think the general guidance is “implement everything we can in Kokkos, except when we can’t or it doesn’t make sense to”. Our fork of RRTMGP was written in YAKL, so functions which need to work directly with RRTMGP are stuck in here to share data types (we have to do deep copies between YAKL arrays and Kokkos views generally due to differences in layout I think).

components/eamxx/src/physics/rrtmgp/scream_rrtmgp_interface.cpp
```diff
@@ -878,5 +879,79 @@ namespace scream {
             });
         }
 
+        void compute_aerocom_cloudtop(
```
brhillman, line 882: See below comment, but this could probably be impl’d in Kokkos and avoid some extra copies back and forth between YAKL arrays and Kokkos views. I.e., could probably use Kokkos views directly from the Field Manager rather than relying on the YAKL copies.

components/eamxx/src/physics/rrtmgp/scream_rrtmgp_interface.cpp
bartgol: Side comment: the syntax here is correct, since `+` has precedence over `&&`. Alas, for a second I was wondering "is this doing + or > first? Is this a bug?", and had to google c++ ops precedence rules (I google that page way more than I should). Perhaps we could use some parentheses to help readers avoid confusion.

components/eamxx/src/physics/rrtmgp/scream_rrtmgp_interface.cpp
```diff
@@ -878,5 +879,79 @@ namespace scream {
...
+                  // product) Start at 2 (second highest) layer because the
+                  // highest one has no clouds
+                  for(int ilay = 2; ilay <= nlay; ++ilay) {
+                    // We only do the calculation if certain conditions are met
+                    if(qc(icol, ilay) + qi(icol, ilay) > 0 &&  // BAD_CONSTANT!
+                       cldfrac_tot(icol, ilay) > 0.001) {      // BAD_CONSTANT!
+                      aerocom_cld(icol, ilay) = cldfrac_tot(icol, ilay);
+                      aerocom_tmp(icol) =
+                          aerocom_clr(icol) *
+                          (1 - std::max(aerocom_cld(icol, ilay - 1),
```
bartgol, line 913: I am not sure this is standard compliant code, since std functions are in general not device friendly. I've seen some std::blah stuff compile and run, but I think it may be preferable to avoid them. An alternative option could be `ekat::impl::max(a,b)`.

jeff-cohere: In this situations, if the code doesn't compile, it's usually in a GPU configuration. If you have access to a GPU-enabled system, you can build there to see whether things of this sort would cause problems. It is inconvenient, though, that the CPU and GPU variants of this dialect of "C++" require this sort of scrutiny.

components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
bartgol: Uhm, I forgot about that. This is b/c rad can be supercycled, right? I guess you can do that, but we really need to fix our I/O classes to avoid this stuff.

jeff-cohere:
The AT seems to be having trouble cloning your fork/branch, which explains (among other things) why it failed so quickly.

```
b"ERROR: Error cloning remote repo 'origin'"
b'hudson.plugins.git.GitException: Command "/home/projects/ppc64le/git/2.10.1/bin/git fetch --tags --progress -- https://github.com/mahf708/scream.git +refs/heads/master:refs/remotes/origin/master +refs/pull/2446/head:refs/heads/pr/2446" returned status code 128:'
b'stdout: '
b"stderr: fatal: Couldn't find remote ref refs/pull/2446/head"
```

@bartgol , I know you've been looking at the AT recently. Might this be related to the fact that this PR comes from a fork and not the E3SM's own SCREAM repo?

bartgol:
> The AT seems to be having trouble cloning your fork/branch, which explains (among other things) why it failed so quickly.
> @bartgol , I know you've been looking at the AT recently. Might this be related to the fact that this PR comes from a fork and not E3SM's own SCREAM repo?

Yeah, that was it. I had configured our Jenkins job to use the "source" repo for the PR (which the AT code grabs from github), which in this case is a fork. But our testing infrastructures actually exploits the fact that the PR branch is available in the target repo (even if the PR comes from a fork), via the `refs/pull/N` refspec. I changed our jenkins scripts to always use the scream repo for the source code.

components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
```diff
@@ -844,6 +867,18 @@ void RRTMGPRadiation::run_impl (const double dt) {
       rrtmgp::compute_cloud_area(ncol, nlay, nlwgpts,     0,                            400e2, p_lay, cld_tau_lw_gpt, cldhgh);
       rrtmgp::compute_cloud_area(ncol, nlay, nlwgpts,     0, std::numeric_limits<Real>::max(), p_lay, cld_tau_lw_gpt, cldtot);
 
+      // Compute cloud-top diagnostics following AeroCOM recommendation
+      auto T_mid_at_cldtop       = real1d("T_mid_at_cldtop", ncol);
```
bartgol, line 871: Personally, I find code like `auto x = blah(...);` confusing when `blah` is a type and not a function (well, it is also a fcn, namely the constructor, but you get the point). When I see this I think that `real1d` is a function, and it's not immediately clear that `x` is indeed an object of type `real1d`. I think writing `real1d x (...);` is much clearer (there's no way of misinterpreting what x is here).

That said, this is perfectly valid code. And considering that you are planning to move it to kokkos later one (IIUC what you said earlier), there's absolutely no need to change this. Just wanted to share some thought on formatting (since you showed interested in the topic earlier ;) ).

components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
```diff
@@ -859,6 +894,12 @@ void RRTMGPRadiation::run_impl (const double dt) {
         d_cldmed(icol) = cldmed(i+1);
         d_cldhgh(icol) = cldhgh(i+1);
         d_cldtot(icol) = cldtot(i+1);
+        d_T_mid_at_cldtop(icol) = T_mid_at_cldtop(i+1);
+        d_p_mid_at_cldtop(icol) = p_mid_at_cldtop(i+1);
+        d_cldfrac_ice_at_cldtop(icol) = cldfrac_ice_at_cldtop(i+1);
+        d_cldfrac_liq_at_cldtop(icol) = cldfrac_liq_at_cldtop(i+1);
```
bartgol, line 900: These are all 1d views, so there's no C vs F90 ordering. I assume YAKL does offer unmanaged arrays, so one could simply create `cldfrac_XYZ_at_cldtop` using the same data ptr from the kokkos views, and skip this copy after the calculation...

bartgol: Done. #2466

bartgol (dismissed):

jeff-cohere (approved):
This looks good to me. The comments on approximations and implementations to revisit were helpful. Let's get this thing in.

components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
brhillman: Sounds good.

brhillman: YAKL does offer unmanaged arrays/views which can use the kokkos data directly.

components/eamxx/src/physics/rrtmgp/scream_rrtmgp_interface.cpp
```diff
@@ -878,5 +880,87 @@ namespace scream {
...
+                SimpleBounds<1>(ncol), YAKL_LAMBDA(int icol) {
+                  // Loops over all layers in serial (due to accumulative
+                  // product), starting at 2 (second highest) layer because the
+                  // highest one has no clouds
+                  for(int ilay = 2; ilay <= nlay; ++ilay) {
+                    // We only do the calculation if certain conditions are met
+                    if((qc(icol, ilay) + qi(icol, ilay)) >
+                           0 &&                              // BAD_CONSTANT!
+                       (cldfrac_tot(icol, ilay) > 0.001)) {  // BAD_CONSTANT!
+                      aerocom_cld(icol, ilay) = cldfrac_tot(icol, ilay);
```
brhillman, line 919: it looks like the top level aerocom_cld starts as 0 and is never set to cldfrac_tot at top, is this intentional? I'm sure doesn't matter.

components/eamxx/src/physics/rrtmgp/scream_rrtmgp_interface.cpp
```diff
@@ -878,5 +880,87 @@ namespace scream {
...
+                      aerocom_tmp(icol) =
```
brhillman, line 920: I know this is supposed to be "temporary", but in general it would be good to see some documentation here describing the algorithm or at minimum referring to an equation in a paper or something.

components/eamxx/src/physics/rrtmgp/scream_rrtmgp_interface.cpp
```diff
@@ -878,5 +880,87 @@ namespace scream {
...
+                          T_mid_at_cldtop(icol) +
+                          tmid(icol, ilay) *
+                              (aerocom_clr(icol) - aerocom_tmp(icol));
+                      // p_mid_at_cldtop
+                      p_mid_at_cldtop(icol) =
+                          p_mid_at_cldtop(icol) +
+                          pmid(icol, ilay) *
+                              (aerocom_clr(icol) - aerocom_tmp(icol));
+                      // cldfrac_ice_at_cldtop
+                      cldfrac_ice_at_cldtop(icol) =
```
brhillman, line 937: I don't follow, the calculation of cldfrac_ice_at_cldtop looks identical to cldfrac_liq_at_cldtop

components/eamxx/src/physics/rrtmgp/scream_rrtmgp_interface.cpp
```diff
@@ -878,5 +880,87 @@ namespace scream {
...
+                       (cldfrac_tot(icol, ilay) > 0.001)) {  // BAD_CONSTANT!
+                      aerocom_cld(icol, ilay) = cldfrac_tot(icol, ilay);
+                      aerocom_tmp(icol) =
+                          aerocom_clr(icol) *
+                          (1 - ekat::impl::max(aerocom_cld(icol, ilay - 1),
+                                               aerocom_cld(icol, ilay))) /
+                          (1 - ekat::impl::min(aerocom_cld(icol, ilay - 1),
+                                               1 - 0.001));  // BAD_CONSTANT!
+                      // T_mid_at_cldtop
+                      T_mid_at_cldtop(icol) =
```
brhillman, line 927: Sorry for being dense, but I'm having a hard time understanding these calculations and what they mean physically. Can you add some more description here, at least so I can follow along?

components/eamxx/src/physics/rrtmgp/scream_rrtmgp_interface.cpp
```diff
@@ -878,5 +880,182 @@ namespace scream {
...
+            photonic approach from above the model top. Let's say pp(k) is
+            the probability of a photon passing through the layer k. We follow
+            the maximum-random overlap assumption.
+
+            1.  Assume the highest layer has no clouds, thus the pp(k) = 1 for
+                the highest layer. Note that pp(k) is initialized as 1 for all
+                layers.
+            2.  Starting at the second highest layer, k+1, if some "cloudy"
+                conditions are met, and if we assume complete randomness, we can
+                naively write pp(k+1) = pp(k) * (1-cldfrac(k+1)) /
```
bartgol, line 911: I'm curious: why do you have `1 - cldfrac(k)` in the denom? Once it gets there, whether or not the photon can go through lev=k+1 should not depend on what's on lev k, no? Besides, this probability could  very well be larger than 1, depending on the two cldfrac values...

bartgol: Follow up: it seems that, expanding the formula for `pp(k)`, one gets
```
pp(k+1) = pp(k) * (1-cldfrac(k+1)) / (1-cldfrac(k)) = pp(k-1) * (1-cldfrac(k)) / (1-cldfrac(k-1)) * (1-cldfrac(k+1)) / (1-cldfrac(k))  = ... = 1 - cldfrac(k+1)
```
since `cldfrac(1)=0`, `pp(1)=1` and all other terms cancel...

components/eamxx/src/physics/rrtmgp/scream_rrtmgp_interface.cpp
```diff
@@ -878,5 +880,182 @@ namespace scream {
...
+            "calculating properties" at said cloud top.
+            */
+
+            /* PART I: Probabilistically  determining cloud top
+            We treat model columns independently, so we loop over all columns
+            in parallel. We then loop over all layers in serial (due to
+            needing an accumulative product), starting at 2 (second highest)
+            layer because the highest is assumed to have no clouds. Let's take a
+            photonic approach from above the model top. Let's say pp(k) is
+            the probability of a photon passing through the layer k. We follow
```
brhillman, line 903: This also assumes that cloudy portions of the layers are completely opaque, right? Because technically I think you'd have a non-zero probability of a photon passing through even cloudy layers.

PR #2448, opened 2023-08-11
Move extra RRTMGP temporary arrays to scratch buffer

### LINKED ISSUE #2447: Two temporaries in RRTMGP are allocated at every time step
The two views at [this line](https://github.com/E3SM-Project/scream/blob/f06407f30f27557d8ae6581ca7d6fd058f3450d1/components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp#L567) in RRTMGP are allocated on device at every time step.

Device allocation is not cheap, and should be done only once during init. It's also possible to simply use unmanaged views from the atm scratch buffer, but as a first iteration it's ok to just use managed views, but allocated at init time.

brhillman:
Move extra RRTMGP temporary arrays to scratch buffer. These were previously being allocated each timestep in `run_impl`, which is just silly. Fixes #2447.

[B4B]

### DIFF
--- a/components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
+++ b/components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
@@ -176,7 +176,9 @@ size_t RRTMGPRadiation::requested_buffer_size_in_bytes() const
     Buffer::num_3d_nlay_nswbands*m_col_chunk_size*(m_nlay)*m_nswbands +
     Buffer::num_3d_nlay_nlwbands*m_col_chunk_size*(m_nlay)*m_nlwbands +
     Buffer::num_3d_nlay_nswgpts*m_col_chunk_size*(m_nlay)*m_nswgpts +
-    Buffer::num_3d_nlay_nlwgpts*m_col_chunk_size*(m_nlay)*m_nlwgpts;
+    Buffer::num_3d_nlay_nlwgpts*m_col_chunk_size*(m_nlay)*m_nlwgpts +
+    Buffer::num_3d_nlay_nswbands_ncol * m_ncol * m_nlay * m_nswbands +
+    Buffer::num_3d_nlay_nlwbands_ncol * m_ncol * m_nlay * m_nlwbands;
 
   return interface_request * sizeof(Real);
 } // RRTMGPRadiation::requested_buffer_size
@@ -237,11 +239,15 @@ void RRTMGPRadiation::init_buffers(const ATMBufferManager &buffer_manager)
   mem += m_buffer.sw_heating.totElems();
   m_buffer.lw_heating = decltype(m_buffer.lw_heating)("lw_heating", mem, m_col_chunk_size, m_nlay);
   mem += m_buffer.lw_heating.totElems();
-  // 3d arrays
   m_buffer.p_lev = decltype(m_buffer.p_lev)("p_lev", mem, m_col_chunk_size, m_nlay+1);
   mem += m_buffer.p_lev.totElems();
   m_buffer.t_lev = decltype(m_buffer.t_lev)("t_lev", mem, m_col_chunk_size, m_nlay+1);
   mem += m_buffer.t_lev.totElems();
+  m_buffer.d_tint = decltype(m_buffer.d_tint)(mem, m_col_chunk_size, m_nlay+1);
+  mem += m_buffer.d_tint.size();
+  m_buffer.d_dz  = decltype(m_buffer.d_dz )(mem, m_col_chunk_size, m_nlay);
+  mem += m_buffer.d_dz.size();
+  // 3d arrays
   m_buffer.sw_flux_up = decltype(m_buffer.sw_flux_up)("sw_flux_up", mem, m_col_chunk_size, m_nlay+1);
   mem += m_buffer.sw_flux_up.totElems();
   m_buffer.sw_flux_dn = decltype(m_buffer.sw_flux_dn)("sw_flux_dn", mem, m_col_chunk_size, m_nlay+1);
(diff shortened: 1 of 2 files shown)


brhillman:
> It may be worth putting `dz` and `tint` back to the main field manager/buffer. However, these types of things should really be calculated elsewhere... (should be calculated somewhere in the dynamic core maybe?)

Yeah if these were calculated elsewhere it’d be great to just pull them out of the FM. But since they aren’t, I thought it made sense for RRTMGP to compute what it needs here. That is, I don’t think we want to get in the habit of telling other processes to compute specialized quantities because one other process requires them. Are other processes also computing and using Tint and dz? That might be motivation for moving these up if so.

components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
```diff
@@ -393,12 +408,18 @@ void RRTMGPRadiation::run_impl (const double dt) {
...
   // Output fields
   auto d_tmid = get_field_out("T_mid").get_view<Real**>();
-  using SmallPack = ekat::Pack<Real,SCREAM_SMALL_PACK_SIZE>;
-  const int n_lay_w_pack = SCREAM_SMALL_PACK_SIZE*ekat::npack<SmallPack>(m_nlay);
-  view_3d_real d_aero_tau_sw("aero_tau_sw",m_ncol,m_nswbands,n_lay_w_pack);
-  view_3d_real d_aero_ssa_sw("aero_ssa_sw",m_ncol,m_nswbands,n_lay_w_pack);
-  view_3d_real d_aero_g_sw  ("aero_g_sw"  ,m_ncol,m_nswbands,n_lay_w_pack);
-  view_3d_real d_aero_tau_lw("aero_tau_lw",m_ncol,m_nlwbands,n_lay_w_pack);
+
+  // This is annoying to have to have to make extra copies of these, but we
```
bartgol, line 412: If I understand correctly, these views are only used to init the yakl arrays. inside the pfor down below, right? If that's correct, I would do
```
if (m_do_aerosol_rad) {
  ...
} else {
   // init yakl arrays to 0 (rather than kokkos views)
}
```
Then, inside the pfor, we could do
```
        // Note that RRTMGP expects ordering (col,lay,bnd) but the FM keeps things in (col,bnd,lay) order
        if (do_aerosol_rad) {
          Kokkos::parallel_for(Kokkos::TeamVectorRange(team, nswbands*nlay), [&] (const int&idx) {
              auto b = idx / nlay;
              auto k = idx % nlay;
              aero_tau_sw(i+1,k+1,b+1) = d_aero_tau_sw(icol,b,k);
              aero_ssa_sw(i+1,k+1,b+1) = d_aero_ssa_sw(icol,b,k);
              aero_g_sw  (i+1,k+1,b+1) = d_aero_g_sw  (icol,b,k);
          }); 
        }
```
ans similarly for the `d_aero_tau_lw` loop.

bartgol (dismissed):
Looks good. I left a comment regarding your comment on the waste of temporaries, but it can be done at a later time.

components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
brhillman: Yeah so the problem I'm wrestling with is, if `m_do_aerosol_rad` is false, then `aero_tau_sw`, etc., won't be provided by any other process, and they are "required" fields when `m_do_aerosol_rad` is true. So we can't put something like 'auto d_aero_tau_sw = get_field_in("aero_tau_sw").get_view<const Real***>();' outside the logical, but we likewise can't do
```
if (m_do_aerosol_rad) {
  auto d_aero_tau_sw = get_field_in("aero_tau_sw").get_view<const Real***>();
}
```
because then when we try to use it later, we get a build fail because `d_aero_tau_sw` was not declared in scope. Am I missing something?

bartgol: Yes, you can do that.

brhillman: Okay, I think I did that.

components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
```diff
@@ -643,18 +623,33 @@ void RRTMGPRadiation::run_impl (const double dt) {
...
-              aero_tau_sw(i+1,k+1,b+1) = d_aero_tau_sw(icol,b,k);
-              aero_ssa_sw(i+1,k+1,b+1) = d_aero_ssa_sw(icol,b,k);
-              aero_g_sw  (i+1,k+1,b+1) = d_aero_g_sw  (icol,b,k);
-          });
-          Kokkos::parallel_for(Kokkos::TeamVectorRange(team, nlwbands*nlay), [&] (const int&idx) {
-              auto b = idx / nlay;
-              auto k = idx % nlay;
-              aero_tau_lw(i+1,k+1,b+1) = d_aero_tau_lw(icol,b,k);
-          });
+          if (m_do_aerosol_rad) {
```
bartgol, line 626: You can't access a class member from inside a kernel [^1]. You need to create a local copy before the kernel dispatch, and use that in the kernel.

[^1] Soon we will switch to C++17, and we will be able to use `KOKKOS_CLASS_LAMBDA`, which will allow to capture members of the class. For now, we're still stuck with this tedious workaround.

bartgol (changes requested):

components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
bartgol: Also, It appears some markdown syntax doesn't work in PR comments...

brhillman: Doh, I knew that. Pushing a fix.

brhillman: Okay should be fixed.

bartgol (approved):

PR #2450, opened 2023-08-14
EAGLES: adds a second MAM-related AtmosphereProcess with constant optical properties

jeff-cohere:
This PR is the last in a series of "software plumbing only" PRs to set up the development environment for properly integrating the C++ port of MAM4 with EAMxx.

* It adds a "MAM optics" process that supplies aerosol optical properties to RRTMGP, which allows us to run simple simulations with MAM4's nucleation parameterization and optics that are constant but "reasonable" enough to allow the simulations to complete.
* The failing HOMME+MAM4 standalone test has been fixed.
* Commit `07d46e9` includes an I/O-related fix that prevents a double deallocation of a SCORPIO Fortran variable in the case that a simulation fails during the initialization step. I can cherry-pick this commit into a smaller PR if necessary. Hopefully Luca's [rewrite of the SCORPIO interfaces to call the C libraries directly](https://github.com/E3SM-Project/scream/pull/2361) comes in soon, rendering these kinds of fixes unnecessary.

My next step after this PR is to move all of MAM4's C++ microphysics logic into the microphysics AtmosphereProcess so we can start assessing its correctness and performance with EAMxx and E3SM stakeholders. Exciting stuff!

FYI @susburrows


PeterCaldwell:
I'm glad @brhillman is a reviewer. @hassanbeydoun might be good as well since he was so involved in this stuff for SPA.

components/eamxx/src/physics/mam/eamxx_mam_optics_process_interface.hpp
```diff
@@ -0,0 +1,84 @@
...
+  void set_grids(const std::shared_ptr<const GridsManager> grids_manager) override;
+
+  // process behavior
+  void initialize_impl(const RunType run_type) override;
+  void run_impl(const double dt) override;
+  void finalize_impl() override;
+
+private:
+
+  Logger logger;
```
bartgol, line 64: The base class already stores a logger (provided by the AD). Is there a special need for an additional one?

pbosler (dismissed):

components/eamxx/src/physics/mam/eamxx_mam_optics_process_interface.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+  auto aero_nccn   = get_field_out("nccn"); // FIXME: get rid of this
+
+  // Compute optical properties on all local columns.
+  // (Strictly speaking, we don't need this parallel_for here yet, but we leave
+  //  it in anticipation of column-specific aerosol optics to come.)
+  Kokkos::parallel_for(policy, KOKKOS_LAMBDA(const ThreadTeam& team) {
+    const Int icol = team.league_rank(); // column index
+
+    auto g_sw = ekat::subview(aero_g_sw.get_view<Real***>(), icol);
```
bartgol, line 93: `Field::get_view` is not a device-friendly method. You should extract the view outside the kernel:
```
auto aero_g_sw = get_field_out("aero_g_sw").get_view<Real***>();
```

components/eamxx/src/physics/mam/eamxx_mam_optics_process_interface.cpp
```diff
@@ -0,0 +1,124 @@
...
+    Kokkos::deep_copy(tau_sw, 0.0);
+    Kokkos::deep_copy(tau_lw, 0.0);
+
+    // FIXME: Get rid of this
+    auto nccn = ekat::subview(aero_nccn.get_view<Real**>(), icol);
+    Kokkos::deep_copy(nccn, 50.0);
+  });
+
+  /*
+  // update the timestamps for the fields
```
bartgol, line 110: I know this is commented, but just for clarity: the base class already does this _for all computed fields_. I'm not 100% sure that's the right thing to do, and we may change it in the future, but so far that's how it works.

components/eamxx/src/physics/mam/eamxx_mam_optics_process_interface.hpp
```diff
@@ -0,0 +1,84 @@
...
+
+#include <ekat/ekat_parameter_list.hpp>
+#include <ekat/ekat_workspace.hpp>
+#include <ekat/logging/ekat_logger.hpp>
+#include <mam4xx/mam4.hpp>
+
+#include <string>
+
+#ifndef KOKKOS_ENABLE_CUDA
+#define protected public
```
bartgol, line 16: This is a bit too hacky, I think. `protected` and `private` are two C++ keywords, so I would recommend against redefining those for the CPP inside a header file. You could use alternative names, such as `ProtectedExceptCuda` or something along that line.

components/eamxx/tests/coupled/dynamics_physics/homme_mam4xx_pg2/CMakeLists.txt
```diff
@@ -30,7 +30,7 @@ configure_file(${CMAKE_CURRENT_SOURCE_DIR}/homme_mam4xx_pg2_output.yaml
 
 # Set homme's test options, so that we can configure the namelist correctly
 # Discretization/algorithm settings
-set (HOMME_TEST_NE 2)
+set (HOMME_TEST_NE 4)
```
bartgol, line 33: Any reason not to use NE2? It may help with valgrind/memcheck builds. Having both NE2 and NE4 tests forces us to keep an up to date IC file for both resolutions...

bartgol (changes requested):

bartgol:
The call to get_view on device must change. The rest is up for debate.

components/eamxx/src/physics/mam/eamxx_mam_optics_process_interface.hpp
jeff-cohere: Probably not! I'll remove it.

components/eamxx/src/physics/mam/eamxx_mam_optics_process_interface.cpp
jeff-cohere: Thanks for the tip!

components/eamxx/src/physics/mam/eamxx_mam_optics_process_interface.hpp
jeff-cohere: Fair enough--I'll do it.

components/eamxx/tests/coupled/dynamics_physics/homme_mam4xx_pg2/CMakeLists.txt
jeff-cohere: No reason--I think ne4 was just easier for me and Pete to get working, but now that it does work we can switch it back.

components/eamxx/src/physics/mam/eamxx_mam_microphysics_process_interface.cpp
```diff
@@ -31,8 +27,6 @@ std::string MAMMicrophysics::name() const {
 void MAMMicrophysics::set_grids(const std::shared_ptr<const GridsManager> grids_manager) {
   using namespace ekat::units;
 
-  logger.trace("entering MAMMicrophysics::set_grids");
```
bartgol, line 34: Note: you can still log stuff from this process, using the base class protected method `log`:
```
this->log(LogLevel::trace,"entering MAMMicrophysics::set_grids");
```

bartgol (approved):

components/eamxx/src/physics/mam/eamxx_mam_optics_process_interface.cpp
```diff
@@ -0,0 +1,107 @@
...
+  auto aero_g_sw   = get_field_out("aero_g_sw").get_view<Real***>();
+  auto aero_ssa_sw = get_field_out("aero_ssa_sw").get_view<Real***>();
+  auto aero_tau_sw = get_field_out("aero_tau_sw").get_view<Real***>();
+  auto aero_tau_lw = get_field_out("aero_tau_lw").get_view<Real***>();
+
+  auto aero_nccn   = get_field_out("nccn").get_view<Real**>(); // FIXME: get rid of this
+
+  // Compute optical properties on all local columns.
+  // (Strictly speaking, we don't need this parallel_for here yet, but we leave
+  //  it in anticipation of column-specific aerosol optics to come.)
```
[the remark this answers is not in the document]
jeff-cohere: Your understanding is correct--the parallel_for is unnecessary because we set all optical properties to constants (but we'll need it soon!).

components/eamxx/src/physics/mam/eamxx_mam_optics_process_interface.cpp
```diff
@@ -0,0 +1,107 @@
...
+void MAMOptics::set_grids(const std::shared_ptr<const GridsManager> grids_manager) {
+  using namespace ekat::units;
+
+  grid_ = grids_manager->get_grid("Physics");
+  const auto& grid_name = grid_->name();
+
+  ncol_ = grid_->get_num_local_dofs();      // number of columns on this rank
+  nlev_ = grid_->get_num_vertical_levels(); // number of levels per column
+  nswbands_ = 14;                           // number of shortwave bands
+  nlwbands_ = 16;                           // number of longwave bands
```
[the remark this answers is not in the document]
jeff-cohere: The appended underscore is my informal style for expressing that a variable or function is attached to a context (either it's a member of an object or a static variable within some namespace). EAMxx doesn't have a style guide, so you'll see other things, like Luca's Hungarian-notation-like `m_var` for member variables. I like these contextual "tags" because it makes it less confusing for the reader to figure out where the variable is coming from, and it prevents one from declaring a symbol that occludes a contextual variable.

The EAMxx team has discussed whether to have a style guide in the past, and it was decided to be unnecessary. As the code matures, we might reconsider whether to revisit this decision. There are many auto-formatters out there that eliminate the "holy wars" that often come with adopting styles, which changes the nature of the decision.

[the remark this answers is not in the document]
bartgol: It is one of the common ways to signal that the variable is a class member. That way, you can immediately tell if the var is local to the fcn or a class member. Common choices are `blah_` (this is google's guideline), `_blah`, `m_blah`, `M_blah`, `_M_blah` (this is what std lib uses)`. I like `m_blah` (m for member), which becomes `s_blah` if the member is static. But it is really a personal taste issue.

components/eamxx/src/physics/mam/eamxx_mam_optics_process_interface.cpp
```diff
@@ -0,0 +1,107 @@
...
+    // populate these fields with reasonable representative values
+    Kokkos::deep_copy(g_sw, 0.5);
+    Kokkos::deep_copy(ssa_sw, 0.7);
+    Kokkos::deep_copy(tau_sw, 0.0);
+    Kokkos::deep_copy(tau_lw, 0.0);
+
+    // FIXME: Get rid of this
+    auto nccn = ekat::subview(aero_nccn, icol);
+    Kokkos::deep_copy(nccn, 50.0);
+  });
```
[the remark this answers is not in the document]
jeff-cohere: Definitely just a stopgap. I don't know how MAM computes CCN, but I think it's probably going to be done in the third MAM-related AtmosphereProcess, which will handle aerosol-cloud interactions. This code is still being ported and seems to me to be the most complicated part of MAM.

jeff-cohere:
Any thoughts on this PR, @AaronDonahue , @brhillman , and @hassanbeydoun ? It's not that exciting, so it's okay to say "no," but I want you to have the chance to weigh in if I've inadvertently piqued your interest.

components/eamxx/tests/coupled/dynamics_physics/homme_mam4xx_pg2/CMakeLists.txt
jeff-cohere: We'll have to rework this test in the next PR anyway, so I'll switch it back to NE2 then.

jeff-cohere:
I'm going to get the tests running for this PR while everyone mulls their remaining thoughts. :-)

components/eamxx/src/physics/mam/eamxx_mam_optics_process_interface.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+void MAMOptics::set_grids(const std::shared_ptr<const GridsManager> grids_manager) {
+  using namespace ekat::units;
+
+  grid_ = grids_manager->get_grid("Physics");
+  const auto& grid_name = grid_->name();
+
+  ncol_ = grid_->get_num_local_dofs();      // number of columns on this rank
+  nlev_ = grid_->get_num_vertical_levels(); // number of levels per column
+  nswbands_ = 14;                           // number of shortwave bands
```
AaronDonahue, line 35: We may want to have a check somewhere that confirms that the number of bands in MAM and in RRTMGP are consistent.  Not sure how to do this, without having something like `get_num_swbands` or equivalent somewhere in the code.

jeff-cohere: Great point. I don't even think we'll know the number of bands before the real MAM optics code makes it in. But when it does, this is something we'll want. (I think I got this number from SPA.)

bartgol: If, say, MAM uses a field (with `nswbands` in its layout) that RRTMGP computes, inside `initialize_impl`, when the fields have been set in all atm procs, MAM can check that there is a provider for that field, by pinging the FieldTracking metadata.

bartgol: Better: we can have a common physics data structure in `physics/share`, storing constants like `nswbands`. The two classes will then ping that data structure for this information.

AaronDonahue (approved):

components/eamxx/src/physics/mam/eamxx_mam_optics_process_interface.cpp
bartgol: Yet another thought: perhaps we should make the number of bands a CMake configurable parameter for EAMxx, which could end up in our config.h file... @brhillman may have some thought on this.

bartgol: I don't know when the builds were attempted. The timeout may be b/c another (long) job was running (maybe e3sm testing). Adding the retest label should be enough.

bartgol: Ah, wait, we had some issues with mappy over the weekend. I see that all PRs were having issues starting its job. This is likely not an AT issue, but rather a mappy issue. We've been struggling to keep up with its issues... :/

jeff-cohere:
Any objections to AUTOMERGE-ing this PR?

PR #2454, opened 2023-08-15
EAMxx: revamped some mkdocs stuff, moved into eamxx folder

bartgol:
I thought if we start to bring the mkdocs stuff in our master branch, people are epsilon-more likely to eventually write some documentation. I started to write some documentation for some of the main data structures. I think my next stab will be the testing infrastructure.

I verified that even if the docs is confined to `components/eamxx`, it gets correctly deployed to the gh-pages branch, which is visualized [here](https://e3sm-project.github.io/scream/). I am not sure how we can deal with this once we merge back to E3SM, since there is only one gh page for each repo. Perhaps we can get E3SM to use mkdocs, and we will simply be one of the pages? Otherwise, mkdocs is basically markdown, so we can export it elsewhere (e.g., confluence). But having it in the repo, we can at least "force" people to add some documentation when they add/remove/change major things.

### DIFF
--- a/components/eamxx/docs/mkdocs/docs/user/index.md
+++ b/components/eamxx/docs/mkdocs/docs/user/index.md
@@ -1,0 +1,2 @@
+# SCREAM User Guide
+
--- a/components/eamxx/docs/mkdocs/docs/user/index.md
+++ b/components/eamxx/docs/mkdocs/docs/user/index.md
@@ -1,2 +1,3 @@
 # SCREAM User Guide
 
+For the time being, see our [public confluence EAMxx user guide](https://acme-climate.atlassian.net/wiki/spaces/DOC/pages/3858890786/EAMxx+User+s+Guide)
--- a/components/eamxx/docs/mkdocs/mkdocs.yml
+++ b/components/eamxx/docs/mkdocs/mkdocs.yml
@@ -1,0 +1,59 @@
+site_name: The C++ E3SM Atmosphere Model (EAMxx)
+
+nav:
+  - 'Home': 'index.md'
+  - 'User Guide':
+    - 'Overview': 'user/index.md'
+    - 'Installation': 'common/installation.md'
+  - 'Developer Guide':
+    - 'Overview': 'dev/index.md'
+    - 'Installation': 'common/installation.md'
+    - 'Style Guide': 'dev/style_guide.md'
+    - 'Kokkos and EKAT': 'dev/kokkos_ekat.md'
+    - 'Important Data Structures':
+      - 'Fields': 'dev/field.md'
+      - 'Grids and Remappers': 'dev/grid.md'
+      - 'Atmosphere Processes': 'dev/processes.md'
+      - 'Managers': 'dev/managers.md'
+    - 'I/O': 'dev/io.md'
+    - 'Testing':
+      - 'Standalone': 'dev/standalone_testing.md'
+      - 'Full model': 'dev/cime_testing.md'
+      - 'CI and Nightly Testing': 'dev/ci_nightly.md'
+
+edit_uri: ""
+
+theme:
+  name: material
+  palette:
+  - media: "(prefers-color-scheme: light)"
+    scheme: default
+    toggle:
+      icon: material/weather-sunny
+      name: Switch to dark mode
+  - media: "(prefers-color-scheme: dark)"
+    scheme: slate
+    toggle:
+      icon: material/weather-night
+      name: Switch to light mode
+  features:
+    - navigation.indices
+    - navigation.instant
+    - navigation.sections
+    - navigation.top
+#    - navigation.tabs
+
+markdown_extensions:
+  - pymdownx.highlight
+  - pymdownx.superfences
+  - pymdownx.tabbed:
+      alternate_style: true
+  - pymdownx.arithmatex:
+      generic: true
+
+extra_javascript:
+  # - javascript/mathjax.js
+  - https://polyfill.io/v3/polyfill.min.js?features=es6
+  - https://cdn.jsdelivr.net/npm/mathjax@3/es5/tex-mml-chtml.js
+
+repo_url: https://github.com/E3SM-Project/scream
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,6 @@ buildlib_cmakec
 
 # Ignore emacs backup files
 *~
+
+# Ignore mkdocs site-generated files in eamxx
+components/eamxx/docs/mkdocs/site/*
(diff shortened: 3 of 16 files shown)


components/eamxx/docs/mkdocs/docs/user/index.md
```diff
@@ -0,0 +1,2 @@
+# SCREAM User Guide
```
jeff-cohere, line 1: Maybe we should get rid of this, because E3SM's User Guide strategy is all-Confluence, all the time (which I think is fine).

components/eamxx/docs/mkdocs/mkdocs.yml
```diff
@@ -0,0 +1,56 @@
+site_name: The C++ E3SM Atmosphere Model (EAMxx)
+
+nav:
+  - 'Home': 'index.md'
+  - 'User Guide':
```
jeff-cohere, line 5: Delete and delegate to Confluence User Guide?

jeff-cohere (dismissed):
A great start! I'll try to help flesh some things out soon--maybe I'll attack the build/install stuff.

@PeterCaldwell , since we have a Confluence "user guide", should we get rid of the User Guide here?

PeterCaldwell:
Thanks Luca! My main question with mkdocs is how folks will flag bad explanations or ask for additions. I really like @susburrows ' idea of prioritizing responses rather than getting a continuous stream of requests. Is the vision that users would identify the offending chunk of markdown code, then make a git issue about it which we would put into the appropriate stack to solve? Does anyone have a better idea?

I'm also curious whether we can merge the automated tools we've been talking about for listing the compile-time and run-time options and their allowable values into the webpages generated by mkdocs. Like, would we trigger a function which collects all the atmchange variables and their options and writes a markdown page which mkdocs then compiles and stuffs with our other git pages?

Apologies if this is way outside the scope of your humble PR. I'm trying to understand whether I like mkdocs as our solution.

components/eamxx/docs/mkdocs/docs/user/index.md
bartgol: You mean completely remove the user guide part?

PeterCaldwell:
> since we have a Confluence "user guide", should we get rid of the User Guide here?

I'm ambivalent. It would be nice to have all our documentation in one place and I'm unclear whether the current confluence documentation is a permanent solution. It certainly isn't complete. The user guide will also have to change as we change the code(?), so having it live with the code makes sense. But if the rest of E3SM has its users' guide in Confluence, we should to. In short, maybe we should leave the mkdocs pages as they are now, but add a link to the confluence EAMxx user guide on the appropriate mkdocs page (for the time being).

bartgol:
@jeff-cohere I'm ok either way. I can remove the user guide folder, and delegate to confluence.

@PeterCaldwell if we remove user guide, this documentation will be a devs only guide. That said, new devs can still flag stuff as unclear by opening an issue, possibly adding a github link to the file and linenumber of the doc section that needs clarification.

Regarding auto-generation of params documentation, we can work on a script, and make it so the script spits out a md file in the `components/eamxx/docs/mkdocs/dev` folder, which mkdocs process every time it runs. So when someone adds/removes/changes some params, they simply do
```
$ ./scripts/update-eamxx-params-docs
$ cd docs/mkdocs
$ mkdocs build && mkdocs gh-deploy
```
et voila`, the github page is updated.

bartgol:
> > since we have a Confluence "user guide", should we get rid of the User Guide here?
> 
> I'm ambivalent. It would be nice to have all our documentation in one place and I'm unclear whether the current confluence documentation is a permanent solution. It certainly isn't complete. The user guide will also have to change as we change the code(?), so having it live with the code makes sense. But if the rest of E3SM has its users' guide in Confluence, we should to. In short, maybe we should leave the mkdocs pages as they are now, but add a link to the confluence EAMxx user guide on the appropriate mkdocs page (for the time being).

That's a good solution. The github.io page is persistent, so we can simply link to it on our confluence user guide...

jeff-cohere:
I'm also fine with having the Dev and User Guides both in mkdocs, of course. :-)

I'll think about how we might incorporate suggestions (from users--devs can probably just open issues). We might be able to get away with something simple, like a link to a Confluence page intended for documentation requests. That would let people comment or write, whatever they feel more comfortable doing.

jeff-cohere:
And Luca's suggestion for auto-generated documentation for parameters illustrates one of the nicest things about having documentation in a plain-text format. :-)

jeff-cohere (approved):

PeterCaldwell:
I just pushed a change to the user-docs index.html providing the link to the confluence documentation. Revert it if you don't like it. 

Given that we now have the capability of auto-generating mkdocs pages for all our atmchange options, I'm in favor of putting our user guide in mkdocs and (eventually) just making the confluence user guide just a stub linking to our github pages content. 

I also think we should force folks who have complaints or suggestions about our documentation to provide that feedback via git issues. Anyone running the code should have a github account and should be able to create an appropriate git issue if we provide documentation on how to do so...

jeff-cohere:
I agree with @mahf708 that it would be cool for us to make sure that the new docs are published if there are updates in the `docs/` directory, but I don't think we need to do that in this PR. I also think we can work out some guidelines for development practices in and around documentation, but that probably needs a bit more discussion.

If we get this and #2455 in, we can start to experiment, and use either GitHub Actions or the Sandia AT to publish doc changes based on some criteria--in the latter case, we'd just use the `mkdocs` command correctly.

Thanks for resurrecting this effort, Luca! I was keeping a very old branch with this documentation stuff around in case we ever got back to it, but you beat me to the punch!

jeff-cohere (approved):

jeff-cohere:
I say let's merge it!

PeterCaldwell:
Yeah, merging sounds good. However, one comment about:

> we can start pointing people to the docs, "encouraging" them to contribute?

I see getting the Cess and aerosol runs going and finishing DP++ and regional nudging as higher priorities than writing documentation. I don't want to damp the momentum we've generated so definitely do documentation stuff if you think of something you're really excited about, but please don't neglect your more time-sensitive tasks!

I'm also a little worried that we're jumping into mkdocs - which is a huge long-term decision - without careful consideration. I'm pretty sure this is what we want to do, but I want to check with Rob and I need to ask the E3SM execs if this is ok. So don't burn too many hours on it quite yet.

jeff-cohere:
Definitely agree about priorities, but I do think we can take advantage of selecting a documentation "platform" that uses Markdown, even if it's not our final platform. We'll be able to create documentation that is easy to use in other situations, available in plain text, and easy to work on in bits.

PR #2456, opened 2023-08-15
Initialize char variable to fix valgrind error

tcclevenger:
This appears to be the cause of the valgrind error on mappy. I believe the call to `std::string(buffptr)` requires `buffptr` to be initialized, which does not happen yet since the DP function is not yet implemented. Does this make sense, @jgfouca?


jgfouca (approved):
Thanks!

PR #2457, opened 2023-08-15
DP-SCREAM (v0): Add wind nudging options

bogensch:
Some IOP files contain either large-scale or geostrophic wind profiles.  This PR:

1. Provides optional inputs for large-scale or geostrophic winds (denoted in IOP files as "u_ls" and "v_ls").  Winds are assumed to be geostrophic if user sets "iop_apply_coriolis = .true." and large scale (or background) winds otherwise.
2. Adds a routine "iop_apply_coriolis" in which the user has the option to nudge the winds to the geostrophic winds by applying Coriolis force, as is sometimes specified in certain cases (enabled by setting iop_coriolis = .true.).

This PR only effects the v0 code base as DP-SCREAM is not yet converted to c++.  Functionality of these routines have been tested extensively in the COMBLE and CASS cases which are being added to the DP-SCREAM library via the THREAD project.

PR is B4B as the test cases do not have "u_ls" or "v_ls" in their IOP files.


jgfouca (approved):
Thanks, I will regenerate the boilerplate when this goes in.

xuezhengllnl (commented):
I just reviewed the changed files. All changes look good to me.

PR #2458, opened 2023-08-15
Autogenerate markdown documentation file for all the atmchange-able parameters of EAMxx

bartgol:
This script parses our `namelist_defaults_scream.xml` file, and produces a markdown file in the folder `components/eamxx/docs/mkdocs/common` (which will exist as soon as #2454 is merged). The generated file contains all (and only) the parameters that are atmchange-able, listing the name, type, description, and, if present, valid values and/or constraints.

The generated file (see attached) clearly shows that we have some work to do in adding more metadata to our XML defaults file.

Ideally, when someone adds a new configurable parameters, they should  run the script in their branch, and commit also the changed newly generated markdown file. Then, upon PR merge, deploy the new documentation via
```
$ cd components/eamxx/docs/mkdocs
$ mkdocs build
$ mkdocs gh-deploy
```
Note: the above folder does not yet exist in master, it will be created only when #2454 is merged.

If you don't have mkdocs, simply run `$ pip install mkdocs pymdown-extensions mkdocs-material`. That should be enough (as of today) to run mkdocs.

[eamxx_params.md](https://github.com/E3SM-Project/scream/files/12353778/eamxx_params.md)

### DIFF
--- a/components/eamxx/cime_config/eamxx_buildnml.py
+++ b/components/eamxx/cime_config/eamxx_buildnml.py
@@ -43,7 +43,7 @@ CIME_VAR_RE = re.compile(r'[$][{](\w+)[}]')
 #    Examples:
 #      - constraints="ge 0; lt 4" means the value V must satisfy V>=0 && V<4.
 #      - constraints="mod 2 eq 0" means the value V must be a multiple of 2.
-METADATA_ATTRIBS = ("type", "valid_values", "locked", "constraints", "inherit")
+METADATA_ATTRIBS = ("type", "valid_values", "locked", "constraints", "inherit", "doc")
 
 ###############################################################################
 def do_cime_vars(entry, case, refine=False, extra=None):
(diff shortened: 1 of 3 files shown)


PeterCaldwell:
This is great! We may not explain what any of our parameters do, but we certainly have a lot of them! Just seeing the list will be very useful.

jgfouca (changes requested):
Looks good! My only thought is that eamxx-params-docs-autogen is not quite following the pattern of our other tools, specifically, it's not doing any option parsing/handling. I think it's probably worth adding this, even if `--help` is the only current supported option.

bartgol:
> Looks good! My only thought is that eamxx-params-docs-autogen is not quite following the pattern of our other tools, specifically, it's not doing any option parsing/handling. I think it's probably worth adding this, even if `--help` is the only current supported option.

I thought about it, but since there's no real option to parse, I figured I'd skip it. I can add that layer.

jgfouca (approved):

PR #2459, opened 2023-08-16
Add TurbulentMountainStress AtmosphereProcess

tcclevenger:
Creates `TurbulentMountainStress` ATM process which produces `surf_drag_coef_tms` and `wind_stress_tms`. Then in `SHOCMacrophysics` ATM process, apply by
```
upwp_sfc = (wind_stress - surf_drag_coeff_tms*u(srf))/rrhoi(surf)
```
and similar for `v`. If no TMS is present, then set`surf_drag_coeff=0` and `upwp_sfc` is calculated as before.

This PR also renames the field `surf_mom_flux` -> `wind_stress` which should more accurately describe the quantity.

Questions/comments on the application side @PeterCaldwell @bogensch @oksanaguba @mt5555:
1. Does anyone want `wind_stress_tms` right now? I have it as a field, but I can remove that.
2. How do I test this? Any ne30pg2 `n`-months simulation I can do and look at some output to see if it is working how I think it is?
3. Just to clarify, we use `sgh30` not `sgh`?

Also, @PeterCaldwell, I did end up including `surf_drag_coeff_tms` as it's own field. I don't think it's possible to just update the SHOC input in TMS since the input isn't calculated until SHOC's processor (which runs after TMS), and the input itself is a local variable. We could make `{u|v}pwp_srf` a global field that TMS would have access to, but that doesn't get around adding an extra field and I think makes it harder to follow how it's updated.


tcclevenger:
Pinging @crterai. TMS can be used by just adding `tms` to input.yaml processes.

bartgol (dismissed):
Looks good on the AD side.

components/eamxx/tests/uncoupled/surface_coupling/surface_coupling_output.yaml
```diff
@@ -13,7 +13,7 @@ Field Names:
   - surf_lw_flux_up
   - surf_sens_flux
   - surf_evap
-  - surf_mom_flux
+  - wind_stress
```
oksanaguba, line 16: you renamed surf_mom_flux everywhere to wind_stress. it is a little confusing, since the old name was connected to surf_sens_flux and surf_evap (and all 3 are coming from the same codes). not sure whether i contributed to this confusion, last we talked about naming iirc it was only in context of QFLX limiter, not TMS.

mt5555:
some comments about the design:  

Are you sure this should be it's own process?  IIUC, it is computing a field, but then that field is used by the SHOC process.  So the TMS process isn't a typical process since it doesn't update any state variables.   Does our current design support some way to communicate that the TMS process has computed something that will then be used by SHOC?  Or is this creating a hidden connection between these two processes?  (which the AD tries to avoid)

It seems more logical that TMS would be a subroutine in the SHOC process.  SHOC computes several momentum fluxes (including TMS) and then adds them all together.  Are the other momentum fluxes in SHOC their own processes?   

I'm not well versed in the design of the atmosphere driver - so feel free to ignore this comment if it doesn't make sense.

oksanaguba:
@mt5555  Not sure if you mean that shoc computes stresses -- it does not. it receives stresses from the coupler and then converts them into its own variables. unrelated to that, but does it matter when TMS (or its coefficient) is computed wrt dycore or physics?

tcclevenger:
@mt5555 I agree that in its current form it acts much more like a SHOC preprocess. The way SHOC handles on/off TMS is basically setting this drag coeff to 0 in initialization, so if TMS is off, it has no effect on SHOC computation. This is not ideal as it is hidden. I could have the driver query all the processes and ask if TMS is present and then set a `bool` in SHOC to use the drag coeff, but that also seems hacky (but maybe better).

The reason it is it's own process is from @PeterCaldwell saying that in the future it would be a different impl (ECMWF implementation, I think?) that would fit much more into it's own process? I think this version is so simple that I would be in favor of moving to SHOC for now with a param `compute_tms: true/false`. Makes it super easy to use. But I'm also ok keeping it as a process if that is preferred.

components/eamxx/tests/uncoupled/surface_coupling/surface_coupling_output.yaml
tcclevenger: The variable name seemed wrong to me. Basically before this change, `surf_mom_flux_u` (same for `_v`) was imported from cpl variables `Faxx_taux`, which in `cam_cpl_indices.F90` in EAM is called "wind stress" with units `N/m2`. In EAM SHOC the input `uw_sfc` is referred to as "surface momentum flux" with units `m2/s2`, and is preprocessed in the interface as `uw_sfc = surf_mom_flux_u/rrho`. This seemed inconsistent, but I'm happy to change it back if the old naming made more sense.

PeterCaldwell:
Regarding whether TMS should be its own process: for this current implementation, no. The current impl is such a small chunk of code, is tightly coupled to SHOC, and is very similar in nature to a lot of other functions which make up SHOC. 

Recall, however, that this version is just a stopgap fix. TMS is tightly related to orographic gravity waves, and future implementations might make more sense as part of that parameterization, or as their own process in order to connect better to both SHOC and GWD. In addition, future impls may be more complex or we might want to be able to swap where they are called or which impl we use. At that point, it will make sense for them to be their own process. In addition, it will be nice at that point for users to be able to quickly find "the TMS process" rather than doing a global grep to realize it is part of SHOC.

bartgol:
It seems to me to be a similar case to `cld_fraction`. But I forgot the reason why that is its own pkg rather than being included in p3 or shoc...

PeterCaldwell:
Maybe we should have included cld_fraction in SHOC. It is distinct than SHOC because SHOC predicts our liquid cloud fraction and cld_fraction predicts ice and (from this) total cloud fraction. I think we made a separate process for it because Peter B didn't want ice polluting his perfect little scheme and because we'll eventually need to come up with a better implementation for ice cloud fraction.

tcclevenger:
From EAMxx dev telecon: keeping as it's own process. I'm going to look into maybe changing the logic for how SHOC knows (or doesn't) about TMS, but I'll update here when it's ready to be looked at again.

@oksanaguba Let me know your thoughts on the `surf_mom_flux`->`wind_stress` name change. I'm happy to change back (I put it in it's own commit so it would be easy if there was objection).

PeterCaldwell:
> The variable name seemed wrong to me. Basically before this change, surf_mom_flux_u (same for _v) was imported from cpl variables Faxx_taux, which in cam_cpl_indices.F90 in EAM is called "wind stress" with units N/m2. In EAM SHOC the input uw_sfc is referred to as "surface momentum flux" with units m2/s2, and is preprocessed in the interface as uw_sfc = surf_mom_flux_u/rrho. This seemed inconsistent, but I'm happy to change it back if the old naming made more sense.

https://en.wikipedia.org/wiki/Wind_stress and other webpage points out that wind stress has the same units as momentum flux (N/m2). Thus even though calling <u'w'>=mom_flux/density a momentum flux is wrong, calling it a wind stress is also wrong. It is really the covariance between u and w.

components/eamxx/src/control/atmosphere_driver.cpp
```diff
@@ -232,6 +232,14 @@ void AtmosphereDriver::create_grids()
 
   m_atm_logger->debug("  [EAMxx] Grids created.");
 
+  // If TMS process is enabled, SHOC needs to know to request tms' surface drag coefficient
+  // as a required field during the set_grid() call below, but SHOC does not have knowledge
+  // of other processes. The driver needs propgate this information to SHOC.
+  if(m_atm_process_group->has_process("tms") &&
+     m_atm_process_group->has_process("shoc")) {
+    setup_shoc_tms_links();
+  }
```
tcclevenger, line 241: `SHOCMacrophysics::set_grids()` is the first place that needs to know if TMS is a process, so call here to set `apply_tms=true` is shoc params.

components/eamxx/src/control/atmosphere_driver.cpp
```diff
@@ -414,6 +422,19 @@ void AtmosphereDriver::setup_column_conservation_checks ()
...
+{
+  EKAT_REQUIRE_MSG(m_atm_process_group->has_process("tms"),
+                   "Error! Attempting to setup link between "
+                   "SHOC and TMS, but TMS is not defined.\n");
+  EKAT_REQUIRE_MSG(m_atm_process_group->has_process("shoc"),
+                   "Error! Attempting to setup link between "
+                   "SHOC and TMS, but SHOC is not defined.\n");
+
+  auto shoc_process = m_atm_process_group->get_process_nonconst("shoc");
+  shoc_process->get_params().set<bool>("apply_tms", true);
```
tcclevenger, line 435: This is the only use I can find for `get_params()`, so I changed it to be nonconst.

components/eamxx/src/physics/shoc/eamxx_shoc_process_interface.cpp
```diff
@@ -432,6 +437,10 @@ void SHOCMacrophysics::run_impl (const double dt)
                        shoc_preprocess);
   Kokkos::fence();
 
+  if (m_params.get<bool>("apply_tms", false)) {
+    apply_turbulent_mountain_stress();
+  }
```
tcclevenger, line 441: After SHOC processes inputs, update `upwp_sfc` input with TMS coeff

components/eamxx/src/share/atm_process/atmosphere_process.hpp
```diff
@@ -115,7 +115,7 @@ class AtmosphereProcess : public ekat::enable_shared_from_this<AtmosphereProcess
   const ekat::Comm& get_comm () const { return m_comm; }
 
   // Return the parameter list
-  const ekat::ParameterList& get_params () const { return m_params; }
+  ekat::ParameterList& get_params () { return m_params; }
```
tcclevenger, line 118: See comment where this is used. Was previously unused, but now I'm using and need to set a param (nonconst function).

components/eamxx/src/share/atm_process/atmosphere_process_group.cpp
```diff
@@ -149,6 +149,24 @@ AtmosphereProcessGroup (const ekat::Comm& comm, const ekat::ParameterList& param
...
+    if (process->type() == AtmosphereProcessType::Group) {
+      const auto* group = dynamic_cast<const AtmosphereProcessGroup*>(process.get());
+      return group->get_process_nonconst(name);
+    } else if (process->name() == name) {
+        return_process = process;
+        break;
+    }
+  }
+  return return_process;
+}
```
tcclevenger, line 168: New helper function for finding a field by name. Could just implement these lines in the driver where this function is called if we don't want this version.

components/eamxx/tests/CMakeLists.txt
```diff
@@ -10,9 +10,9 @@ if (NOT DEFINED ENV{SCREAM_FAKE_ONLY})
   set(TEST_RANK_END ${SCREAM_TEST_MAX_RANKS})
 
   # Initial condition files used in the tests
-  set(EAMxx_tests_IC_FILE_72lev   "screami_unit_tests_ne2np4L72_20220822.nc")
-  set(EAMxx_tests_IC_FILE_128lev  "screami_unit_tests_ne2np4L128_20220822.nc")
-  set(EAMxx_tests_TOPO_FILE       "USGS-gtopo30_ne2np4pg2_16x_converted.c20200527.nc")
+  set(EAMxx_tests_IC_FILE_72lev  "screami_unit_tests_ne2np4L72_20220822.nc")
+  set(EAMxx_tests_IC_FILE_128lev "screami_unit_tests_ne2np4L128_20220822.nc")
+  set(EAMxx_tests_TOPO_FILE      "USGS-gtopo30_ne2np4pg2_x6t_20230331.nc")
```
tcclevenger, line 15: The previously used topo files has incorrect dims for PG2 data...

tcclevenger (commented):
I've updated how SHOC AP knows to apply TMS that should be much more clear. Basically the driver sees if TMS and SHOC are listed as processes, and the set a parameter in SHOC AP `apply_tms`.

I'll let the AT run over the weekend and then ping for reviews.

components/eamxx/cime_config/namelist_defaults_scream.xml
```diff
@@ -276,6 +279,8 @@ be lost if SCREAM_HACK_XML is not enabled.
 
     <mac_aero_mic inherit="atm_proc_group">
       <atm_procs_list>(shoc,cldFraction,spa,p3)</atm_procs_list>
+      <atm_procs_list hgrid=".*pg2">(tms,shoc,cldFraction,spa,p3)</atm_procs_list> <!-- TMS only available for PG2 -->
+      <atm_procs_list COMPSET=".*SCREAM.*noAero" hgrid=".*pg2">(tms,shoc,cldFraction,p3)</atm_procs_list>
```
bartgol, line 283: @jgfouca may correct me, but I think this should come _after_ the next line. CIME should use the last matching entry, so you want the most specific ones after the general one.

components/eamxx/src/share/atm_process/atmosphere_process_group.cpp
bartgol: I think it's fine.

bartgol (approved):
AD/infrastructure looks good to me.

bartgol (changes requested):
...but I think you need that XML change.

components/eamxx/cime_config/namelist_defaults_scream.xml
jgfouca: Agree.

bartgol (approved):

tcclevenger:
Everything passes (outside DIFFs in CIME tests, which are expected).

@oksanaguba @PeterCaldwell I see chat in eamxx slack channel, seems that TMS isn't necessarily making things better? Let me know if TMS should be merged as is (turned on by default), merged turned off by default, or wait on merging.

AaronDonahue:
@tcclevenger , it looks like this PR is no in confllct w/ `master`.  Do you mind fixing the conflicts and pushing a new branch?

AaronDonahue:
Also, in this current configuration is TMS on by default?  Could we change it to be off by default, with an option to turn it on?  We can always turn it on by default in the future.  @oksanaguba , @PeterCaldwell I'm asking this based on the conversation we had on the eval call, does this sound like the proper plan to move forward on this PR?

tcclevenger:
@AaronDonahue It's currently turned on. Let me know, I'll turn it off.

PeterCaldwell:
Yeah, I want this PR merged ASAP and it seems like concern about whether it will cause bad behavior is preventing that from happening. So let's merge it with TMS turned off by default.

tcclevenger:
@PeterCaldwell done. I kept it in the standalone test so that the code path is still touched in the AT.

bartgol (approved):

tcclevenger:
@bartgol Can I get that approval again. Something ended up merging right before creating conflicts.

bartgol (approved):

PR #2460, opened 2023-08-16
Get eamxx to build on lassen.

jgfouca:
A few changes were needed.
1) A working netcdf toolchain. I built it and put it in a public place
2) Fix lapack linking
3) Fix thrust version check
4) Use newer cuda module

I assume tests will work, but I lack the necessary access on Lassen to run things at the moment.


AaronDonahue (approved):

AaronDonahue:
I ran into a few issues running the tests, but not building.  I suspect the test fails are less related to machine config and maybe an issue with lassen having access to specific files.  I'll have to hunt those down.

PR #2462, opened 2023-08-17
EAMxx: revert change in xml defaults

bartgol:
I am responsible for the wrong change, which caused all v1 tests to fail.


bartgol:
I'm merging this, since it's an urgent hot fix.

PR #2465, opened 2023-08-21
Add workflow to run Mkdocs at every master updated

bartgol:
Add a github action workflow to automatically update the gh-pages branch every time that master is updated.

This should allow people to not even know how Mkdocs works, or even that it exists. Github will handle everything under the hood.

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -15,7 +15,7 @@ jobs:
           ref: master
           fetch-depth: 0 # Needed, or else gh-pages won't be fetched, and push rejected
       - name: Install python deps
-        run: pip install mkdocs pymdown-extensions mkdocs-material
+        run: python3 -m pip install mkdocs pymdown-extensions mkdocs-material
       - name: Build and deploy
         working-directory: components/eamxx/docs/mkdocs
         run: mkdocs build && mkdocs gh-deploy
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -1,0 +1,21 @@
+name: Build and deploy gh-pages branch with Mkdocs
+
+on:
+  # Runs every time master branch is updated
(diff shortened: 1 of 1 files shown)


.github/workflows/gh-pages.yml
```diff
@@ -0,0 +1,21 @@
...
+  Build-and-Deploy-docs:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          repository: bartgol/scream
+          ref: master
+          fetch-depth: 0 # Needed, or else gh-pages won't be fetched, and push rejected
+      - name: Install python deps
+        run: pip install mkdocs pymdown-extensions mkdocs-material
```
jeff-cohere, line 18: Might want to change to `pip3` here, depending on the AT system. It'll be nice when `pip` means "pip for Python 3" at some distant point in the future.

jeff-cohere (approved):
Added a minor `pip` -> `pip3` suggestion. Looks good overall. Thanks, @bartgol !

.github/workflows/gh-pages.yml
bartgol: How about `python3 -m pip` just to be sure?

jeff-cohere (approved):

.github/workflows/gh-pages.yml
```diff
@@ -0,0 +1,21 @@
...
+  push:
+    branches: ["master"]
+
+jobs:
+  Build-and-Deploy-docs:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          repository: bartgol/scream
```
rljacob, line 14: Should this be E3SM-Project/scream ?

jeff-cohere: Ah, yeah, probably should!

PR #2468, opened 2023-08-22
Some upgrades to the XML defaults syntax

bartgol:
After #2458 , we will (hopefully!) start to add documentation to our parameters. Since putting all on one line may be hard to visualize on some text editors, splitting the XML defaults file on multiple lines is probably a good idea. For instance, we may do something like

```
    <atm_log_level type="string"
                   valid_values="trace,debug,info,warn,error"
                   doc="Verbosity level for the atm logger">
      info
    </atm_log_level>
```
However, the current XML parsing would consider as parameter value the whole string between `>` and `</`, including spaces and newlines. Hence, it would cause an error of the form
```
ERROR: Invalid value '
      info
    ' for element 'atm_log_level'. Value not in the valid list ('['trace', 'debug', 'info', 'warn', 'error']')
```
This PR fixes this, by stripping space/newline chars inside the param values.

### DIFF
--- a/components/eamxx/cime_config/eamxx_buildnml.py
+++ b/components/eamxx/cime_config/eamxx_buildnml.py
@@ -807,7 +807,7 @@ def do_cime_vars_on_yaml_output_files(case, caseroot):
         # Hence, change default output settings to perform a single AVERAGE step at the end of the run
         if case.get_value("TESTCASE") in ["ERP", "ERS"]:
             test_env = case.get_env('test')
-            stop_n = test_env.get_value("STOP_N")
+            stop_n = int(test_env.get_value("STOP_N"))
             stop_opt = test_env.get_value("STOP_OPTION")
             content['output_control']['Frequency'] = stop_n
             content['output_control']['frequency_units'] = stop_opt
--- a/components/eamxx/cime_config/eamxx_buildnml.py
+++ b/components/eamxx/cime_config/eamxx_buildnml.py
@@ -238,15 +238,16 @@ def evaluate_selectors(element, case, ez_selectors):
     child_values = {} # elem_name -> evaluated XML element
     children_to_remove = []
     for child in element:
-        child_name = child.tag
-        child_val = child.text
-
         # Note: in our system, an XML element is either a "node" (has children)
         # or a "leaf" (has a value).
         has_children = len(child) > 0
         if has_children:
             evaluate_selectors(child, case, ez_selectors)
         else:
+            child_name = child.tag
+            child.text = str(child.text).strip(' \n')
+            child_val = child.text
+
             selectors = child.attrib
             if selectors:
                 all_match = True
--- a/components/eamxx/cime_config/eamxx_buildnml.py
+++ b/components/eamxx/cime_config/eamxx_buildnml.py
@@ -245,7 +245,7 @@ def evaluate_selectors(element, case, ez_selectors):
             evaluate_selectors(child, case, ez_selectors)
         else:
             child_name = child.tag
-            child.text = str(child.text).strip(' \n')
+            child.text = None if child.text is None else child.text.strip(' \n')
             child_val = child.text
 
             selectors = child.attrib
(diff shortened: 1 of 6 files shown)


bartgol:
@jgfouca I changed the PR name, since I added another feature. Namely, we no longer need to provide "special" null values for arrays (so that the yaml parser can figure out the type of the array items). For instance, instead of
```
<compute_tendencies type="array(string)">NONE</compute_tendencies>
```
we can now just do
```
<compute_tendencies type="array(string)"></compute_tendencies>
```
or even
```
<compute_tendencies type="array(string)"/>
```
This removes the need in the code to handle the special case where the array has length 1, but contains the special value for "null". Instead, the code can simply check the array size, with size>0 implying valid values.

In order to do this I had to add stuff to our buildnml infrastructure. Namely, I had to create new user-defined YAML tags, signaling a sequence of bool/int/float/string values. I also needed to update EKAT, to get the correct version of the parser for these new tags.

components/eamxx/cime_config/eamxx_buildnml.py
```diff
@@ -238,15 +246,16 @@ def evaluate_selectors(element, case, ez_selectors):
...
-        child_val = child.text
-
         # Note: in our system, an XML element is either a "node" (has children)
         # or a "leaf" (has a value).
         has_children = len(child) > 0
         if has_children:
             evaluate_selectors(child, case, ez_selectors)
         else:
+            child_name = child.tag
+            child.text = None if child.text is None else child.text.strip(' \n')
```
jgfouca, line 256: Text should never be `None` now, right?

components/eamxx/cime_config/eamxx_buildnml.py
```diff
@@ -745,6 +754,8 @@ def create_input_data_list_file(caseroot):
 ###############################################################################
 def do_cime_vars_on_yaml_output_files(case, caseroot):
 ###############################################################################
+    from yaml_utils import array_constructor
```
jgfouca, line 757: I'd prefer this be at the top with the other imports.

components/eamxx/cime_config/eamxx_buildnml_impl.py
```diff
@@ -190,6 +190,8 @@ def has_child (root,name):
 ###############################################################################
 def refine_type(entry, force_type=None):
 ###############################################################################
+    from yaml_utils import make_array
```
jgfouca, line 193: Ditto

components/eamxx/cime_config/yaml_utils.py
```diff
@@ -0,0 +1,68 @@
...
+        return Bools(vals)
+    elif etype=="int":
+        return Ints(vals)
+    elif etype=="float" or etype=="real":
+        return Floats(vals)
+    elif etype=="string" or etype=="file":
+        return Strings(vals)
+    else:
+        raise ValueError (f"Unsupported element type '{etype}' for arrays.")
+###############################################################################
```
jgfouca, line 38: Move this second hash line directly below `def make_array`

jgfouca (changes requested):
Some minor nits. Looks good overall.

components/eamxx/cime_config/eamxx_buildnml.py
bartgol: If there is a parameter like `<my_param/>` (open and closed in one set of brackets), then the text will indeed be set to None by the XML parser.

components/eamxx/cime_config/eamxx_buildnml.py
bartgol: ok

components/eamxx/cime_config/eamxx_buildnml.py
jgfouca: @bartgol 

That's is unfortunate.

I thought `<x/>` was just semantic sugar for `<x></x>`. Are you saying the former gives text=None and the latter gives text=""?

jgfouca (approved):

components/eamxx/cime_config/eamxx_buildnml.py
bartgol: I was about to say "yes", then I checked with this mini script
```python
import xml.etree.ElementTree as ET


s = """ 
<tree>
    <a/>
    <b></b>
</tree>
"""
xml = ET.fromstring(s)

for a in xml:
    print(f"{a.tag} = {a.text}")
```
which printed
```bash
a = None
b = None
```
so now I'm not sure anymore...

PR #2469, opened 2023-08-22
buildnml: Use correct version of parse_input

jgfouca:
buildnml scripts should be using the one from CIME.buildnml, not CIME.buildlib which is for build scripts. The former returns one argument, the caseroot, not a tuple. This explains the bug when calling our buildnml from the command line.


bartgol (approved):
Thanks!

PR #2472, opened 2023-08-23
DP: convert advance_iop_forcing

jgfouca:
Issues:
1) A couple of the fortran arrays have dims `(plev, pcnst)`. I wasn't sure which dimension to pack and which index should be the fast index. I think, in fortran for the above dims, plev is the fast index. But the fortran code using these arrays had the pcnst loop as the inner loop. I decide to pack on the plev dimension and keep plev indexing as the fast index.
2) Other than running the code and seeing that it doesn't crash or have memory problems, there's no way to test these conversions.

### DIFF
--- a/components/eamxx/src/physics/dp/impl/dp_advance_iop_forcing_impl.hpp
+++ b/components/eamxx/src/physics/dp/impl/dp_advance_iop_forcing_impl.hpp
@@ -121,13 +121,6 @@ void Functions<S,D>::advance_iop_forcing(
       //  considered if using the preq-x dycore and if three dimensional
       //  forcing is not provided by IOP forcing file.
       Spack t_expan = 0;
-#ifndef MODEL_THETA_L
-      // this term is already taken into account through
-      //  LS vertical advection in theta-l dycore
-      if (!use_3dfrc) {
-        t_expan = scm_dt*wfld(k)*t_in(k)*rair/(cpair*pmidm1(k));
-      }
-#endif
       t_update(k) = t_in(k) + t_expan + scm_dt*(t_phys_frc(k) + t_lsf(k));
       for (Int m = 0; m < pcnst; ++m) {
         q_update(m, k) = q_in(m, k) + scm_dt*q_lsf(m, k);
--- a/components/eamxx/src/physics/dp/impl/dp_advance_iop_forcing_impl.hpp
+++ b/components/eamxx/src/physics/dp/impl/dp_advance_iop_forcing_impl.hpp
@@ -137,14 +137,22 @@ void Functions<S,D>::advance_iop_forcing(
   ////////////////////////////////////////////////////////////
   //  Set U and V fields
 
+  uview_1d<const Spack> u_src, v_src;
+
   if (have_v && have_u && !dp_crm) {
-    Kokkos::deep_copy(u_update, uobs);
-    Kokkos::deep_copy(v_update, vobs);
+    u_src = uobs;
+    v_src = vobs;
   }
   else {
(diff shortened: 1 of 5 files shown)


components/eamxx/src/physics/dp/impl/dp_advance_iop_forcing_impl.hpp
```diff
@@ -13,10 +13,141 @@ namespace dp {
...
+  }
+
+  const Int plev_pack = ekat::npack<Spack>(plev);
+  Kokkos::parallel_for(
+    Kokkos::TeamVectorRange(team, plev_pack), [&] (Int k) {
+      // Initialize thermal expansion term to zero.  This term is only
+      //  considered if using the preq-x dycore and if three dimensional
+      //  forcing is not provided by IOP forcing file.
+      Spack t_expan = 0;
+#ifndef MODEL_THETA_L
```
tcclevenger, line 124: I can't remember the result of the discussion, but don't we only use THETAL?

components/eamxx/src/physics/dp/tests/dp_advance_iop_forcing_tests.cpp
```diff
@@ -48,10 +60,13 @@ struct UnitWrap::UnitTest<D>::TestAdvanceIopForcing {
...
     for (auto& d : cxx_data) {
       d.transpose<ekat::TransposeDirection::c2f>(); // _f expects data in fortran layout
-      advance_iop_forcing_f(d.plev, d.pcnst, d.scm_dt, d.ps_in, d.u_in, d.v_in, d.t_in, d.q_in, d.t_phys_frc, d.u_update, d.v_update, d.t_update, d.q_update);
+      advance_iop_forcing_f(d.plev, d.pcnst, d.scm_dt, d.ps_in, d.have_u, d.have_v, d.dp_crm, d.use_3dfrc, d.u_in, d.v_in, d.t_in, d.q_in, d.t_phys_frc, d.divt3d, d.divq3d, d.divt, d.divq, d.wfld, d.uobs, d.vobs, d.hyai, d.hyam, d.hybi, d.hybm, d.u_update, d.v_update, d.t_update, d.q_update);
       d.transpose<ekat::TransposeDirection::f2c>(); // go back to C layout
     }
 
+    // We can't call into fortran. Due to all the dependencies it has, it's not possible
+    // to build it in standalone eamxx. Without fortran, we cannot do BFB tests.
+#if 0
```
tcclevenger, line 69: For now are we just ignoring all testing? Should we maybe assert non-nan? I don't know how much that would be worth..

tcclevenger (commented):
For 1., I believe in SHOC.F90, the inner-most loop was cols, and outer was levs, so it seems to follow the same here. But I guess there the arrays were ordered (cols, levs).

components/eamxx/src/physics/dp/impl/dp_advance_iop_forcing_impl.hpp
bogensch: That is correct.  In v0 this block of code was in place to provide support for the preqx dycore, which v1 does not have.  Thus, this (and other THETA_L if defs) can be removed.

components/eamxx/src/physics/dp/tests/dp_advance_iop_forcing_tests.cpp
jgfouca: We are testing that the code runs without memory/bounds errors. That's about it. I think this is probably adequate until we can do a regression test.

tcclevenger (approved):

PR #2477, opened 2023-08-24
Jgfouca/revert wet dry restore shoc

jgfouca:


components/eam/src/physics/cam/shoc_intr.F90
```diff
@@ -929,36 +873,12 @@ subroutine shoc_tend_e3sm( &
...
-      if (lq(ixind))  then
-         icnt = icnt + 1
-         if(convert_back_to_wet(icnt)) then !convert from wet to dry mmr if true
-            state1%q(:,:,ixind) = calculate_wetmmr_from_drymmr(ncol, pver, state1%q(:,:,ixind), qv_dry)
-         endif
-      endif
-     enddo
-   rcm(:,:) = calculate_wetmmr_from_drymmr(ncol, pver, rcm, qv_dry)
-   rtm(:,:) = calculate_wetmmr_from_drymmr(ncol, pver, rtm, qv_dry)
+
```
PeterCaldwell, line 876: I think the next few lines are dumb (though I don't think they cause any bugs). If you grep for `edsclr_` you will find that `edsclr_in` is a local variable in shoc_intr.F90 which just gets assigned to values from state%q then passed as an inout variable to shoc_main. After shoc_main, the resulting edsclr_in value is just assigned here to `edsclr_out`... which is all `edsclr_out` is used for. Why not just have one variable named `edsclr` which contains both the input and output versions of edsclr? As an aside, edsclr is the name for the subset of state%q (the advected tracers) which are also turbulently diffused. Can we please come up with a better variable name for this!?!

PR #2479, opened 2023-08-25
Reorganize the docs folder

bartgol:
This PR makes the `components/eamxx/docs` folder a bit less messy.

- Most of the old content of the docs folder is now in `docs/old`. A couple of files (already in md format) are in our docs folder (e.g., `build.md`). Some files have been deleted (empty tex files). The shoc folder in the old docs contained a lengthy tex file, which we may want to convert to md at some point.
- The main `mkdocs.yaml` file is now in the eamxx folder, and the content of `docs/mkdocs/docs` is now in `docs`.

@rljacob I think if we adopt the [monorepo plugin](https://github.com/backstage/mkdocs-monorepo-plugin) (or something equivalent) in the upstream, the only thing to change would be adding the "main" `mkdocs.yml` file, with no changes to the eamxx/docs folder. Of course, we also need to ensure that the workflow for auto-gen the gh-pages that is currently in the scream fork does not end up in the upstream, since the scream version has workflows to deploy docs for `e3sm-project/scream`, while the upstream will have workflows to deploy `e3sm-project/e3sm`...


AaronDonahue:
Just to confirm, this change keeps the SHOC docs, but removes the rest, right?  I'm pretty sure that's what I'm seeing in the code changes but no harm in getting an extra layer of confirmation.

bartgol:
Yes, I kept all the non-trivial files in `docs/old`, with the idea of converting them to md format at some point.

rljacob:
Yes merges back and forth will have to be sure to keep the right mkdocs.yaml file.   Eventually I'd still like to see this separate repo go away so there would be just one web location for scream docs.

bartgol:
> Eventually I'd still like to see this separate repo go away so there would be just one web location for scream docs.

Yes, I think we all agree on that (not sure if there is a tentative timeline for pulling the plug on this fork). But we are doing some merges back upstream (at least we did one, to allow the mmf+scream-physics work), so until then we need to keep our guard up.

AaronDonahue (approved):

PR #2480, opened 2023-08-25
DP: Port advance_iop_nudging

tcclevenger:
Port advance_iop_nudging function.

### DIFF
--- a/components/eamxx/src/physics/dp/impl/dp_advance_iop_nudging_impl.hpp
+++ b/components/eamxx/src/physics/dp/impl/dp_advance_iop_nudging_impl.hpp
@@ -61,7 +61,7 @@ void Functions<S,D>::advance_iop_nudging(
     const auto condition = pmidm1(k) <= iop_nudge_tq_low*100
                            &&
                            pmidm1(k) >= iop_nudge_tq_high*100;
-    rtau(k).set(condition, std::max(scm_dt, iop_nudge_tscale));
+    rtau(k).set(condition, ekat::impl::max(scm_dt, iop_nudge_tscale));
     relaxt(k).set(condition, (t_in(k) - tobs(k))/rtau(k));
     relaxq(k).set(condition, (q_in(k) - qobs(k))/rtau(k));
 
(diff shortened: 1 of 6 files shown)


components/eamxx/src/physics/dp/impl/dp_advance_iop_nudging_impl.hpp
```diff
@@ -13,10 +13,59 @@ namespace dp {
...
+
+    const auto condition = pmidm1(k) <= iop_nudge_tq_low*100
+                           &&
+                           pmidm1(k) >= iop_nudge_tq_high*100;
+    rtau(k).set(condition, std::max(scm_dt, iop_nudge_tscale));
+    relaxt(k).set(condition, (t_in(k) /*- tobs(k)*/)/rtau(k));
+    relaxq(k).set(condition, (q_in(k) /*- qobs(k)*/)/rtau(k));
+
+    t_update(k).set(condition, t_in(k) + relaxt(k)*scm_dt);
+    q_update(k).set(condition, q_in(k) + relaxq(k)*scm_dt);
```
tcclevenger, line 67: @bogensch 2 questions:
1. In lines 63, 64 and 66, 67, I use `t_in` and `q_in` but in the .F90 `t_update` and `q_update` are used. In F90 these `*_update` are `intent(out)` variables, and the `*_in` variables are never used which makes me think it was a typo. Either way, where the function is called, the same variable was used for both inputs, so it wouldn't change the answer. 
2. What is `tobs` and `qobs`?

tcclevenger (commented):
Setting WIP while I ask questions

tcclevenger:
@bogensch I made `tobs` and `qobs` inputs since I think they will need to be provided from dynamics. What are good descriptions for these variables?

components/eamxx/src/physics/dp/impl/dp_advance_iop_nudging_impl.hpp
jgfouca: @tcclevenger , I'm having to deal with `tobs` and `qobs` in my current function as well:

```
    const uview_1d<Spack>& tobs,       // actual temperature, dims=(plev)                                                                                                                                                      
    const uview_1d<Spack>& qobs);      // actual W.V. Mixing ratio, dims=(plev) 
```

bogensch: Sorry for the late reply, for some reason just seeing this.  

1.  Whoops, yes that is a typo in the F90 but yes wouldn't change the answer.
2. tobs = observed temperature [K] ; qobs = observed water vapor [kg/kg]

PR #2481, opened 2023-08-28
Convert advance_iop_subsidence

jgfouca:
It's important to add tests for which the plev falls right on a pack boundary. This helps catch errors where index_and_shfit might go out of bounds and plev+1, a common array size, would also go out of bounds.


bogensch (approved):
Looks good to me!

PR #2482, opened 2023-08-28
Transplanted and adapted some existing docs.

jeff-cohere:
I had written some words about how to configure/build EAMxx (back in the SCREAM days), and also some stuff on testing. I've moved that text into the right places here, and also hooked up the auto-generated  runtime parameters to the navigation bar generated by mkdocs.


jeff-cohere:
This stuff definitely isn't perfect, but I think we should leave the polishing for when we have more stuff. :-) Of course, I guess I'll try to understand why everything I changed has a conflict first, though.

jeff-cohere:
Aha, seems like @bartgol moved stuff around, and I didn't grab that commit when I started this work. Stand by.

jeff-cohere:
I'll try to rework my contributions to the new structure in another PR.

PR #2484, opened 2023-08-28
Update nudging process in EAMxx to be more flexible

AaronDonahue:
This commit makes a number of changes to the nudging process to support a more general implementation.

This includes:

- Adding the nudging process to the namelist defaults so it can be used for CIME cases.
- A runtime option to change the nudging timescale.
- Make helper fields part of the generic atmosphere class instead of just surface coupling export.
- Switch to runtime option for designating which fields will be nudged
- Use the TimeInterpolation class to handle the nudge data management
- Set masked values to the nearest column neighbor, to avoid strange nudging values
- Option for cases where source data is stored on static pressure profiles, so not require p_mid in the source data file.


components/eamxx/cime_config/namelist_defaults_scream.xml
```diff
@@ -214,6 +214,14 @@ be lost if SCREAM_HACK_XML is not enabled.
     <!-- CLD fraction -->
     <cldFraction inherit="atm_proc_base"/>
 
+    <!-- nudging -->
+    <nudging inherit="atm_proc_base">
+      <nudging_filename type="array(string)">NONE</nudging_filename>
```
bartgol, line 219: We no longer have to put a placeholder for arrays. For an empty default, you can simply do
```
<nudging_filename type="array(string)"/>
```
which avoids maintenance burden of handling the default invalid string...

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
```diff
@@ -7,8 +7,22 @@ namespace scream
...
+  m_datafiles  = m_params.get<std::vector<std::string>>("nudging_filename");
+  m_timescale = m_params.get<int>("nudging_timescale",0);
+  m_fields_nudge = m_params.get<std::vector<std::string>>("nudging_fields");
+  auto src_pres_type = m_params.get<std::string>("source_pressure_type","DYNAMIC");
+  if (src_pres_type=="DYNAMIC") {
+    m_src_pres_type = DYNAMIC;
+  } else if (src_pres_type=="STATIC") {
+    m_src_pres_type = STATIC;
+  } else {
+    EKAT_ERROR_MSG("ERROR! Nudging::parameter_list - unsupported source_pressure_type provided.  Current options are [DYNAMICS,STATIC].  Please check");
```
bartgol, line 20: DYNAMIC~S~

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
```diff
@@ -27,329 +41,271 @@ void Nudging::set_grids(const std::shared_ptr<const GridsManager> grids_manager)
...
+	  fill_value = int_state_view_1d(ipack)[iidx];
+	} else {
+	  int_state_view_1d(ipack)[iidx] = fill_value;
+	}
+      }
+    });
+
+
+    // Vertical Interpolation onto atmosphere state pressure levels
+    if (m_src_pres_type == DYNAMIC) {
```
bartgol, line 208: Is the difference between STATIC and DYNAMIC that the former is a 1d pressure view (same for all cols), while the latter is a horiz varying field? If so, I think we should choose different names for the option. Static/dynamic make me think of something constant/varying _in time_, rather than space (and IIUC, in both these cases the pressure is changing in time).

Perhaps something like `UNIFORM` vs `HORIZ_VARYING`, or `VERTICAL_PROFILE` vs `3D_PROFILE`...or something conveying the meaning of 1d vs 3d.

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
```diff
@@ -27,329 +41,271 @@ void Nudging::set_grids(const std::shared_ptr<const GridsManager> grids_manager)
...
+  } else if (m_src_pres_type == STATIC) {
+    p_mid_ext_1d   = get_helper_field("p_mid_ext").get_view<mPack*>();
+  }
+  for (auto name : m_fields_nudge) {
+    auto atm_state_field = get_field_out(name);
+    auto int_state_field = get_helper_field(name);
+    auto ext_state_field = get_helper_field(name+"_ext").get_view<mPack**>();
+    auto atm_state_view  = atm_state_field.get_view<mPack**>();  // TODO: Right now assume whatever field is defined on COLxLEV
+    auto int_state_view  = int_state_field.get_view<mPack**>();
+    view_Nd<mMask,2> int_mask_view("",m_num_cols,m_num_levs); // Track mask of interpolated values
```
bartgol, line 156: We should avoid allocating views inside the run method. It has a non trivial cost.

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
```diff
@@ -27,329 +41,271 @@ void Nudging::set_grids(const std::shared_ptr<const GridsManager> grids_manager)
...
+      auto int_mask_view_1d  = ekat::subview(int_mask_view,icol);
+      auto int_state_view_1d = ekat::subview(int_state_view,icol);
+      Real fill_value;
+      int  fill_idx = -1;
+      // Scan top to surf and backfill all values near TOM that are masked.
+      for (int kk=0; kk<m_num_levs; ++kk) {
+        const auto ipack = kk / mPack::n;
+	const auto iidx  = kk % mPack::n;
+        // Check if this index is masked
+	if (!(int_state_view_1d(ipack)[iidx]==var_fill_value)) {
```
bartgol, line 181: Two things:
- instead of `if (!(a==b))`, you can just do `if (a!=b)`, which is more readable.
- I don't get the logic: if the entry is not the fill value, then you set fill_value=entry...doesn't that set the view to a _valid_ value in the entries that were previously filled?

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
```diff
@@ -27,329 +41,271 @@ void Nudging::set_grids(const std::shared_ptr<const GridsManager> grids_manager)
...
+	  break;
+	}
+      }
+      // Now fill the rest, the fill_idx should be non-negative.  If it isn't that means
+      // we have a column that is fully masked 
+      for (int kk=fill_idx+1; kk<m_num_levs; ++kk) {
+        const auto ipack = kk / mPack::n;
+	const auto iidx  = kk % mPack::n;
+        // Check if this index is masked
+	if (!(int_state_view_1d(ipack)[iidx]==var_fill_value)) {
```
bartgol, line 198: Also here I'm not sure what is happening. If the entry is valid, then you set fill_value to be that entry. Why are we removing the invalid values and setting them to something valid?

AaronDonahue: Similar response as above.  In cases where we have filled/invalid data passed to nudging we are taking the nearest neighbor approach for vertical interpolation.  So here we fill all invalid data with nearest neighbor.

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
AaronDonahue: Great suggestion! Yeah, I'm not a fan of these names I chose, but didn't want to spin my wheels on names so I just moved on.

Re: your question.  DYANAMIC (as named now) would also involve time varying because we will be grabbing p_mid from the nudging data which could be time varying.  It might be better to define 3 modes: `SINGLE_COL_UNIFORM`, `HORIZ_VARYING` and `TIME_DEPENDENT`

elynnwu: Maybe `SINGLE_VERTICAL_PROFILE` instead of `SINGLE_COL_UNIFORM`? I like the other two names.

bartgol (approved):

components/eamxx/cime_config/namelist_defaults_scream.xml
```diff
@@ -219,9 +219,15 @@ be lost if SCREAM_HACK_XML is not enabled.
       <nudging_filename type="array(string)"/>
       <nudging_fields type="array(string)"/>
       <nudging_timescale type="integer">0</nudging_timescale>
-      <source_pressure_type type="string">TIME_DEPENDENT</source_pressure_type>
+      <source_pressure_type type="string">TIME_DEPENDENT_3D_PROFILE</source_pressure_type>
```
bartgol, line 222: You could add `"valid_values=TIME_DEPENDENT_3D_PROFILE,STATIC_1D_VERTICAL_PROFILE"`, which would catch at submit time if someone has a typo in its atmchange...

bartgol: Also, since we have automatic docs generation, do you mind adding `doc="some descirption"` to the new parameters? :)

components/eamxx/cime_config/namelist_defaults_scream.xml
```diff
@@ -219,9 +219,15 @@ be lost if SCREAM_HACK_XML is not enabled.
...
     </nudging>
 
+    <!-- ML correction -->
+    <mlcorrection inherit="atm_proc_base">
+      <ML_model_path type="string">NONE</ML_model_path>
+      <ML_output_fields type="array(string)">NONE</ML_output_fields>
```
bartgol, line 228: For arrays, you don't need to use a special value for invalid. Just leave an empty param:
```
<ML_output_fields type="array(string)"/>
```

bartgol: You'll have to change the place where this param is pinged in the cpp code as well. Any code of the form
```
if (vec.size()==1 && vec[0]=="NONE") {
  return;
}
```
should be replaced with the more self-explanatory (and less bug prone)
```
if (vec.size()==0) {
  return;
}
```

bartgol (changes requested):

components/eamxx/cime_config/namelist_defaults_scream.xml
AaronDonahue: @elynnwu , my changes flagged as changes to the ML correction namelist section.  When we switch to turning ML correction back on for all machines could you also incorporate these suggested changes?

components/eamxx/cime_config/namelist_defaults_scream.xml
AaronDonahue: >You could add "valid_values=TIME_DEPENDENT_3D_PROFILE,STATIC_1D_VERTICAL_PROFILE", which would catch (at submit time) if someone has a typo in its atmchange...

Should I remove the EKAT_ERROR_MSG associated with checking this for a valid value internally then?

bartgol: No, that should stay there, for standalone purposes (as well as for extra protection against xml/inputs tampering). But we want to catch XML errors at buildnml time if possible.

components/eamxx/cime_config/namelist_defaults_scream.xml
elynnwu: Will do.

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
```diff
@@ -27,329 +41,289 @@ void Nudging::set_grids(const std::shared_ptr<const GridsManager> grids_manager)
...
+    // Load p_lev from source data file
+    ekat::ParameterList in_params;
+    in_params.set("Filename",m_datafiles[0]);
+    in_params.set("Skip_Grid_Checks",true);  // We need to skip grid checks because multiple ranks may want the same column of source data.
+    std::map<std::string,view_1d_host<Real>> host_views;
+    std::map<std::string,FieldLayout>  layouts;
+    create_helper_field("p_mid_ext", scalar2d_layout_mid, grid_name, ps);
+    auto pmid_ext = get_helper_field("p_mid_ext");
+    auto pmid_ext_v = pmid_ext.get_view<Real*,Host>();
+    in_params.set<std::vector<std::string>>("Field Names",{"p_lev"});
```
elynnwu, line 104: Sorry to bother you on this again, but the vertical remapped file uses convention `p_levs`.

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
```diff
@@ -27,329 +41,289 @@ void Nudging::set_grids(const std::shared_ptr<const GridsManager> grids_manager)
...
 // =========================================================================================
 void Nudging::initialize_impl (const RunType /* run_type */)
 {
   using namespace ShortFieldTagsNames;
 
-  std::map<std::string,view_1d_host<Real>> host_views;
-  std::map<std::string,FieldLayout>  layouts;
-
+  // Initialize the time interpolator
+  auto grid_ext = m_grid->clone("Point Grid", false);
```
elynnwu, line 82: I'm getting an error on this that:
```
Error! Input field to 'add_field' is defined on a grid different from the one stored.
  - field manager grid: Point Grid
  - input field grid:   Physics PG2
```

Switching to `Physics PG2` works.

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
```diff
@@ -27,329 +41,289 @@ void Nudging::set_grids(const std::shared_ptr<const GridsManager> grids_manager)
...
+       	       KOKKOS_LAMBDA(MemberType const& team) {
       const int icol = team.league_rank();
-      auto T_mid_1d = ekat::subview(T_mid,icol);
-      auto T_mid_ext_1d = ekat::subview(T_mid_ext,icol);
-      auto u_1d = ekat::subview(u,icol);
-      auto u_ext_1d = ekat::subview(u_ext,icol);
-      auto v_1d = ekat::subview(v,icol);
-      auto v_ext_1d = ekat::subview(v_ext,icol);
+      auto int_mask_view_1d  = ekat::subview(int_mask_view,icol);
+      auto int_state_view_1d = ekat::subview(int_state_view,icol);
```
elynnwu, line 173: I think the bug somehow got merged back in. This should be:
```
auto ext_state_view_1d = ekat::subview(ext_state_view,icol);
```
and all the following should be `ext` instead of `int`

elynnwu: I found this note:
image

bartgol (approved):

components/eamxx/docs/developer/processes.md
```diff
@@ -15,3 +15,4 @@ TODO: add links to papers/github-repos, and a SMALL description
 * spa: prescribed aerosols, blah blah
 * surface coupling: blah
 * mam: prognostic aerosols, blah blah
+nudging: This process is responsible for nudging the model simulation given a set of files with a target nudged state.
```
bartgol, line 18: The markdown syntax needs a `*` here.

bartgol (changes requested):

components/eamxx/docs/developer/processes.md
bartgol: You can click on the three dots next to the file and select `View file`. Github will render md syntax, so you can see here how the page will look like (links may not work, but formatting will)

AaronDonahue:
I cannot figure out why this PR has DIFFs on `mappy`.  @bartgol since you have access to `mappy` would it be difficult to look at the cprnc out and figure out which fields are diff-ing?

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.hpp
```diff
@@ -13,6 +13,7 @@
 #include "share/grid/point_grid.hpp"
 #include "share/util/scream_vertical_interpolation.hpp"
 #include "share/util/scream_time_stamp.hpp"
+#include "share/util/scream_universal_constants.hpp"
```
bartgol, line 16: IIUC, this is needed for the def fill value in the cpp file. If so, the include should be moved to the cpp file.

components/eamxx/src/share/field/field_impl.hpp
```diff
@@ -372,9 +372,10 @@ update (const Field& x, const ST alpha, const ST beta)
 
   // Determine if there is a FillValue that requires extra treatment.
   ST fill_val = constants::DefaultFillValue<ST>().value;
-  const auto& xtra_data = get_header().get_extra_data();
+  const auto& xtra_data = x.get_header().get_extra_data();
+
   if (xtra_data.count("mask_value")) {
-    fill_val = ekat::any_cast<ST>(xtra_data.at("mask_value"));
+    fill_val = ekat::any_cast<float>(xtra_data.at("mask_value"));
```
bartgol, line 378: Since `update_impl` expects `fill_value` to be of same type of alpha/beta, we should at least check that the field data type is not `IntType`, otherwise when `update_impl` is called there will be a narrowing conversion.

components/eamxx/src/share/io/scream_scorpio_interface.F90
```diff
@@ -575,7 +575,7 @@ function get_variable_metadata_float(filename, varname, metaname) result(metaval
     ! Find the pointer for this file
     call lookup_pio_atm_file(trim(filename),pio_file,found)
     if (.not.found ) then
-      call errorHandle("PIO ERROR: error getting metadata for variable "//trim(varname)//" in file "//trim(filename)//".\n PIO file not found or not open.",-999)
+      call errorHandle("PIO ERROR: error getting metadata for file "//trim(filename)//".\n PIO file not found or not open.",-999)
```
bartgol, line 578: Why removing the var name in the message? It seemed useful.

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
```diff
@@ -55,9 +57,15 @@ void Nudging::set_grids(const std::shared_ptr<const GridsManager> grids_manager)
   /* ----------------------- WARNING --------------------------------*/
 
   //Now need to read in the file
-  scorpio::register_file(m_datafiles[0],scorpio::Read);
-  m_num_src_levs = scorpio::get_dimlen(m_datafiles[0],"lev");
-  scorpio::eam_pio_closefile(m_datafiles[0]);
+  if (m_src_pres_type == TIME_DEPENDENT_3D_PROFILE) {
+    scorpio::register_file(m_datafiles[0],scorpio::Read);
+    m_num_src_levs = scorpio::get_dimlen(m_datafiles[0],"lev");
+    scorpio::eam_pio_closefile(m_datafiles[0]);
```
elynnwu, line 63: Actually if it's time dependent, it should never be hard-coded to the first one right?

bartgol: The number of levels should not change with time, only the pressure values. So using the first one should be fine.

components/eamxx/src/share/io/scream_scorpio_interface.F90
bartgol: It's true that the issue is that a file is not found, but knowing what var was pinged imho can help us to track where in the code this error is originating. For instance, this can be helpful if there are two streams for the same file (like pg2 and dyn streams).

AaronDonahue: sounds good, I'm happy to revert this change.

components/eamxx/src/share/field/field_impl.hpp
```diff
@@ -372,9 +372,14 @@ update (const Field& x, const ST alpha, const ST beta)
 
   // Determine if there is a FillValue that requires extra treatment.
   ST fill_val = constants::DefaultFillValue<ST>().value;
-  const auto& xtra_data = get_header().get_extra_data();
+  const auto& xtra_data = x.get_header().get_extra_data();
```
bartgol, line 375: I don't think this will compile anymore. We changed how we access the field extra data.

bartgol (changes requested):

bartgol:
@AaronDonahue Notice that the PASSED status is old. The PR is no longer mergeable (so no point in re-running the AT, cause the master auto-merge would fail anyways). You need to merge/rebase-on master first.

AaronDonahue:
@bartgol and @elynnwu , this should be ready for rereview now.  I've updated w/ master and added a new test for more coverage.  Feel free to wait until after the AT reports.  I just wanted to ping you guys.

bartgol (approved):

PR #2485, opened 2023-08-29
add cloud-top estimation following AeroCOM

### LINKED ISSUE #2394: Add functionality to output cloud droplet number at cloud top

### LINKED ISSUE #2466: In RRTMGP interface, avoid Kokkos<->YAKL deep copies for 1d arrays
Since there's no C-vs-Fortran ordering issue with 1d arrays, there really is no need to have two copies of the same data around. YAKL arrays can be created as "unmanaged", using the pointer from the Kokkos views.

See [this](https://github.com/E3SM-Project/scream/pull/2446/files#r1299276889) discussion

mahf708:
Adds select AeroCOM cloud-top diagnostics inside the RRTMGP interface.

Fixes #2394
Fixes #2466

--

For background on the specific algorithm, please refer to the comment in the already existing subroutine `get_subcolumn_mask`. Specifically, the algorithm in `compute_aerocom_cloudtop` is equation 13 (for model column) whereas the one in `get_subcolumn_mask` is equation 14 (for subcolumns). 

```cpp
            // Maximum-random overlap:
                // Uses essentially the algorithm described in eq (14) in Raisanen et al. 2004,
                // https://rmets.onlinelibrary.wiley.com/doi/epdf/10.1256/qj.03.99. Also the same
                // algorithm used in RRTMG implementation of maximum-random overlap (see
                // https://github.com/AER-RC/RRTMG_SW/blob/master/src/mcica_subcol_gen_sw.f90)
                //
```

### DIFF
--- a/components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
+++ b/components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
@@ -902,10 +902,10 @@ void RRTMGPRadiation::run_impl (const double dt) {
       );
 
       // Compute diagnostic total cloud area (vertically-projected cloud cover)
-      auto cldlow = real1d("cldlow", ncol);
-      auto cldmed = real1d("cldmed", ncol);
-      auto cldhgh = real1d("cldhgh", ncol);
-      auto cldtot = real1d("cldtot", ncol);
+      real1d cldlow = real1d("cldlow", d_cldlow.data(), ncol);
+      real1d cldmed = real1d("cldmed", d_cldmed.data(), ncol);
+      real1d cldhgh = real1d("cldhgh", d_cldhgh.data(), ncol);
+      real1d cldtot = real1d("cldtot", d_cldtot.data(), ncol);
       // NOTE: limits for low, mid, and high clouds are mostly taken from EAM F90 source, with the
       // exception that I removed the restriction on low clouds to be above (numerically lower pressures)
       // 1200 hPa, and on high clouds to be below (numerically high pressures) 50 hPa. This probably
@@ -957,10 +957,6 @@ void RRTMGPRadiation::run_impl (const double dt) {
         d_sfc_flux_dif_vis(icol) = sfc_flux_dif_vis(i+1);
         d_sfc_flux_sw_net(icol)  = sw_flux_dn(i+1,kbot) - sw_flux_up(i+1,kbot);
         d_sfc_flux_lw_dn(icol)   = lw_flux_dn(i+1,kbot);
-        d_cldlow(icol) = cldlow(i+1);
-        d_cldmed(icol) = cldmed(i+1);
-        d_cldhgh(icol) = cldhgh(i+1);
-        d_cldtot(icol) = cldtot(i+1);
         Kokkos::parallel_for(Kokkos::TeamVectorRange(team, nlay+1), [&] (const int& k) {
           d_sw_flux_up(icol,k)            = sw_flux_up(i+1,k+1);
           d_sw_flux_dn(icol,k)            = sw_flux_dn(i+1,k+1);


components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
```diff
@@ -883,6 +907,18 @@ void RRTMGPRadiation::run_impl (const double dt) {
       // Get IR 10.5 micron band for COSP
       auto idx_105 = rrtmgp::get_wavelength_index_lw(10.5e-6);
 
+      // Compute cloud-top diagnostics following AeroCOM recommendation
+      auto T_mid_at_cldtop       = real1d("T_mid_at_cldtop", ncol);
```
bartgol, line 911: Since these are all 1d arrays, why not making all these views use the pointer from the kokkos views? Something like
```
real1d T_mid_at_cldtop = real1d("T_mid_at_cldtop,d_T_mid_at_cldtop.data(),ncol);
```
This should allow you to remove the deep copies like `d_T_mid_at_cldtop(icol) = T_mid_at_cldtop(i+1);` below. Same goes for the cldlow/med/hgh/tot arrays...

brhillman: Agreed.

components/eamxx/src/physics/rrtmgp/scream_rrtmgp_interface.cpp
```diff
@@ -908,5 +910,90 @@ namespace scream {
...
+          memset(cldfrac_liq_at_cldtop, 0.0);
+          memset(cldfrac_tot_at_cldtop, 0.0);
+          memset(cdnc_at_cldtop, 0.0);
+          // Initialize the 1D "clear fraction" as 1 (totally clear)
+          auto aerocom_clr = real1d("aerocom_clr", ncol);
+          memset(aerocom_clr, 1.0);
+          // Get gravity acceleration constant from constants
+          using physconst = scream::physics::Constants<Real>;
+          // Set non-zero q threshold
+          constexpr real q_threshold = 0.0;  // BAD_CONSTANT!
```
brhillman, line 939: Note that, even with the best intentions, things tend to left in the code longer than we think they will...rather than "BAD_CONSTANT", should we instead drop a note like "TODO: this should be tuned"? Just an idea.

components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
```diff
@@ -883,6 +907,18 @@ void RRTMGPRadiation::run_impl (const double dt) {
...
+      auto p_mid_at_cldtop       = real1d("p_mid_at_cldtop", ncol);
+      auto cldfrac_ice_at_cldtop = real1d("cldfrac_ice_at_cldtop", ncol);
+      auto cldfrac_liq_at_cldtop = real1d("cldfrac_liq_at_cldtop", ncol);
+      auto cldfrac_tot_at_cldtop = real1d("cldfrac_tot_at_cldtop", ncol);
+      auto cdnc_at_cldtop        = real1d("cdnc_at_cldtop", ncol);
+      rrtmgp::compute_aerocom_cloudtop(
+          ncol, nlay, t_lay, p_lay, p_del, z_del, qc, qi, cldfrac_tot, nc,
```
brhillman, line 918: I feel like I'm constantly puzzling over where we should use "cldfrac_tot" vs "cldfrac_tot_for_analysis", but note that "cldfrac_tot" is very sensitive to the presence of even very small amounts of ice (I think the threshold for cldfrac_tot = 1 is something like qi > 1e-12). So, I think in using cldfrac_tot here, you're making sure that your cldfrac is consistent with radiation (probably a good thing), but I do wonder if this is going to make your diagnostics look a little funny in the presence of ice. Have you thought about this?

components/eamxx/src/physics/rrtmgp/scream_rrtmgp_interface.cpp
```diff
@@ -908,5 +910,90 @@ namespace scream {
...
+            const real2d &p_del, const real2d &z_del, const real2d &qc,
+            const real2d &qi, const real2d &cldfrac_tot, const real2d &nc,
+            real1d &T_mid_at_cldtop, real1d &p_mid_at_cldtop,
+            real1d &cldfrac_ice_at_cldtop, real1d &cldfrac_liq_at_cldtop,
+            real1d &cldfrac_tot_at_cldtop, real1d &cdnc_at_cldtop) {
+          /* The goal of this routine is to calculate properties at cloud top
+           * based on the AeroCOM recommendation. See reference for routine
+           * get_subcolumn_mask above, where equation 14 is used for the
+           * maximum-random overlap assumption for subcolumn generation. We use
+           * equation 13, the column counterpart.
```
brhillman, line 924: Thanks for this reference, that's really helpful. I think this is what I missing last time looking at this.

components/eamxx/src/physics/rrtmgp/scream_rrtmgp_interface.cpp
```diff
@@ -908,5 +910,90 @@ namespace scream {
...
+                        (1.0 - ekat::impl::max(cldfrac_tot(icol, ilay - 1),
+                                               cldfrac_tot(icol, ilay))) /
+                        (1.0 - ekat::impl::min(cldfrac_tot(icol, ilay - 1),
+                                               1.0 - cldfrac_tot_threshold));
+                    // Temporary variable for probability "weights"
+                    auto aerocom_wts = aerocom_clr(icol) - aerocom_tmp;
+                    // Temporary variable for liquid "phase"
+                    auto aerocom_phi =
+                        qc(icol, ilay) / (qc(icol, ilay) + qi(icol, ilay));
+                    /* PART II: The inferred properties */
```
brhillman, line 967: Is there a paper reference for this part?

components/eamxx/src/physics/rrtmgp/scream_rrtmgp_interface.cpp
```diff
@@ -908,5 +910,90 @@ namespace scream {
...
+                    /* In general, convering a 3D property X to a 2D cloud-top
```
brhillman, line 968: "convering" -> "converting"?

components/eamxx/src/physics/rrtmgp/scream_rrtmgp_interface.cpp
jeff-cohere: The BAD_CONSTANT terminology likely comes from EAGLES, where we have been stamping all over MAM whenever we see one (which is often!).

jeff-cohere (commented):
Looks pretty good to me so far! I don't have anything to add to what Ben and Luca have said.

crterai:
If it's relatively easy to add, it'd be great to have `eff_radius_qc` as a variable that can be diagnosed at liquid cloud top. It's been used in observational studies as a proxy for likelihood of precipitation, and its maps can be compared to existing passive satellite retrievals.

bartgol (dismissed):

components/eamxx/src/physics/rrtmgp/tests/rrtmgp_unit_tests.cpp
```diff
@@ -619,3 +619,225 @@ TEST_CASE("rrtmgp_cloud_area") {
...
+      eff_radius_qc_at_cldtop, eff_radius_qi_at_cldtop);
+
+  REQUIRE(cldfrac_tot_at_cldtop.createHostCopy()(1) == .7);
+
+  // Case 3xtra: test max overlap
+  // This case produces >0.7 due to slight enhancement in the presence of a
+  // local minimum (0.1 is the local minimum between 0.2 and 0.4)
+  yakl::fortran::parallel_for(
+      1, YAKL_LAMBDA(int /* dummy */) {
+        cldfrac_tot(1, 5) = 0.1;
```
brhillman, line 719: Minor point, it would be clearer to include setting the other levels here too so that this test was self contained. There's nothing wrong with this though as written.

components/eamxx/src/physics/rrtmgp/tests/rrtmgp_unit_tests.cpp
```diff
@@ -619,3 +619,225 @@ TEST_CASE("rrtmgp_cloud_area") {
...
+        cldfrac_tot(1, 5) = 0.0;
+        cldfrac_tot(1, 6) = 0.1;
+      });
+  scream::rrtmgp::compute_aerocom_cloudtop(
+      ncol, nlay, tmid, pmid, p_del, z_del, qc, qi, rel, rei, cldfrac_tot, nc,
+      tmid_at_cldtop, pmid_at_cldtop, cldfrac_ice_at_cldtop,
+      cldfrac_liq_at_cldtop, cldfrac_tot_at_cldtop, cdnc_at_cldtop,
+      eff_radius_qc_at_cldtop, eff_radius_qi_at_cldtop);
+
+  REQUIRE(cldfrac_tot_at_cldtop.createHostCopy()(1) >
```
brhillman, line 744: Since it's random, shouldn't this be >= .7? I think it won't matter with how you've written the algorithm but would matter if you applied the overlap rules in a stochastic fashion.

brhillman (approved):

PR #2487, opened 2023-08-29
Tune max_ni

hassanbeydoun:
Changed max_total_ni to 740e3 from 500e3 to achieve desired decrease in global longwave cooling


crterai (approved):
Changes look good to me and consist with multiplying the ni_max threshold by 1.48. Thanks, @hassanbeydoun!

AaronDonahue (changes requested):
This needs an equivalent change in the F90 code to pass p3_run_and_cmp

AaronDonahue (approved):

AaronDonahue:
@jgfouca , @bartgol - it's been awhile since we've made a change that impacted run_and_cmp.  Remind me, we would expect p3_run_and_cmp to actually fail if we made a non-BFB change to C++ and F90 right?  Even if they actually match.  Because the comparison is based on baselines generated from the current `master` right?

jgfouca:
@AaronDonahue , yes, i think that's right

AaronDonahue:
okay cool, then I think this is ready to merge

PR #2489, opened 2023-08-30
More improvements to gh-pages docs workflow

mahf708:
Motivation: There was still a failure after merging the ml_correction PR (perhaps due to lack of sync? see https://github.com/E3SM-Project/scream/actions/runs/6019310104/job/16328947146) and this workflow adjustment should hopefully be more robust in the future. 

- reduces fetch-depth (slightly quicker); we still need to get rid of CIME dependency in the doc generation logic, but that will wait for later
- use more standard and automated action to push generated docs: https://github.com/peaceiris/actions-gh-pages?tab=readme-ov-file#readme (instead of previous manual pushing via mkdocs deploy)

⚠️ before we merge, @bartgol, we need to go over the `secrets.GITHUB_TOKEN` to 1) set the permissions correctly and 2)  see if it is okay with hierarchy to enable it (it is relatively safe and automated, but we can fine-grain it). More info: https://docs.github.com/en/actions/security-guides/automatic-token-authentication

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -10,6 +10,7 @@ on:
   # Runs every time a PR is open against master
   pull_request:
     branches: [ master ]
+
   workflow_dispatch:
 
 jobs:
@@ -21,7 +22,7 @@ jobs:
       - name: Check out the repository
         uses: actions/checkout@v3
         with:
-          fetch-depth: 0 # Needed, or else gh-pages won't be fetched, and push rejected
+          # TODO: git rid of dependency on CIME
           submodules: recursive
   
       - name: Show action trigger
@@ -47,9 +48,10 @@ jobs:
         run: |
           mkdocs build --strict --verbose
 
-      # only deploy when there is a push to master
       - if: ${{ github.event_name == 'push' }}
-        name: Deploy docs
-        working-directory: components/eamxx
-        run: |
-          mkdocs gh-deploy --verbose
+        name: GitHub Pages action
+        uses: peaceiris/actions-gh-pages@v3.9.3
+        with:
+          # information about GITHUB_TOKEN are in settings
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          publish_dir: ./components/eamxx/site
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -15,7 +15,7 @@ on:
 
 jobs:
 
-  emaxx-docs:
+  eamxx-docs:
     runs-on: ubuntu-latest
 
     steps:
@@ -23,6 +23,9 @@ jobs:
         uses: actions/checkout@v3
         with:
           # TODO: git rid of dependency on CIME
+          # TODO: another option to investigate is a sparse checkout. 
+          # In the scream repo, all other components do not need to be checked out. 
+          # And even in the upstream, we mainly need only components/xyz/docs (and a few more places).
           submodules: true
   
       - name: Show action trigger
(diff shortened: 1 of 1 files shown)


bartgol (approved):
I thought that the push event on master would trigger an action on the server side. It's almost as if the action was triggered by the entity that pushed, but even then, the push rejection message is suggesting that the remote was updated on the meantime. I'm not sure who could have updated it, unless we forcefully merged a PR while the at tested that one. Who knows. Either way, I agree that using an existing action to deploy should be more robust.

As for the token, I'm not sure what's there to do. That's not a token associated to any account btw.

jeff-cohere (commented):
This seems fine to me, as long as it doesn't create new areas of expertise. :-P

bartgol:
> We just need someone with admin access to check on the permissions in the settings panel, see bottom of screenshot above from my fork
The permissions should already be read/write (otherwise the workflow would not have been able to update the gh-pages branch all along)

bartgol:
> Screenshot 2023-08-29 at 9 39 14 PM
> Side note: the action triggers a legit "deployment" action as well so we get this clean way to look at what it does: [mahf708/scream/actions/runs/6019678385](https://github.com/mahf708/scream/actions/runs/6019678385)

I am curious to see if this happens in this repo. The current workflow was triggering the deployment action on my fork too, but not on the main repo. I bet it has to do with organization-vs-user repo. Either way, we'll see what happens once it's merged.

bartgol (approved):
Looks good. I have a suggestion for submodules, but not necessary. Speed is not _that_ important for docs generation.

jeff-cohere (dismissed):

jeff-cohere (dismissed):

.github/workflows/gh-pages.yml
```diff
@@ -10,6 +10,7 @@ on:
   # Runs every time a PR is open against master
   pull_request:
     branches: [ master ]
+
   workflow_dispatch:
 
 jobs:
```
bartgol, line 16: Can we rename `emaxx-docs` to `eamxx-docs`?

bartgol (changes requested):

.github/workflows/gh-pages.yml
bartgol: Could not comment on the actual line, since it was not a touched one, but it's the job name

jeff-cohere: For a second there, I pictured Emaxx: the C++/Kokkos port of GNU Emacs. <shudder>

.github/workflows/gh-pages.yml
```diff
@@ -21,7 +22,7 @@ jobs:
       - name: Check out the repository
         uses: actions/checkout@v3
         with:
-          fetch-depth: 0 # Needed, or else gh-pages won't be fetched, and push rejected
+          # TODO: git rid of dependency on CIME
```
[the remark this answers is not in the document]
bartgol: TODO: another option to investigate is a sparse checkout. In the scream repo, all other components do not need to be checked out. And even in the upstream, we mainly need only `components/xyz/docs` (and a few more places).

.github/workflows/gh-pages.yml
```diff
@@ -49,7 +53,9 @@ jobs:
...
-        name: Deploy docs
-        working-directory: components/eamxx
-        run: |
-          mkdocs gh-deploy --verbose
+        name: GitHub Pages action
+        uses: peaceiris/actions-gh-pages@v3.9.3
+        with:
+          # information about GITHUB_TOKEN are in settings
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          publish_dir: ./components/eamxx/site
```
rljacob, line 61: why replace the simpler "mkdocs gh-deploy" with this?

bartgol (approved):

PR #2492, opened 2023-08-30
Remove ML correction from coupler CMakelist

elynnwu:
PR #2442 broke nightly tests since it was not tested properly on all machines. This PR temporary disable ML correction from being compiled and should allow nightly tests to run normally.

We will re-enable this once all machines can build ML correction properly.


AaronDonahue (approved):

AaronDonahue (approved):

bartgol:
I don't think this is enough. I am still getting a python error on frontier, since the ml folder is still parsed by cmake. Will add a follow-up PR.

PR #2494, opened 2023-08-31
EAMxx: better handle enabling/disabling ml_correction

bartgol:
Two things:

* Add cmake option to enable it (OFF by default)
* If disabled, do not even parse the parametrization folder

Current master was not linking ml_correction to libatm, but was still configuring/building it, which on certain machines is currently still causing problems.


elynnwu (approved):
Thanks! This looks a lot cleaner.

elynnwu:
Quick question, would this break ML correction unit test on quartz? It uses the flag `RUN_ML_CORRECTION_TEST` to determine whether to build/run the test.

bartgol:
> Quick question, would this break ML correction unit test on quartz? It uses the flag `RUN_ML_CORRECTION_TEST` to determine whether to build/run the test.

Ah, yes, let me fix that.

elynnwu (approved):

PR #2500, opened 2023-08-31
Fix leap year logic in EAMxx

bartgol:
When running through CIME, we get the type of calendar from CIME. I manually verified that this works.

### DIFF
--- a/components/eamxx/src/mct_coupling/atm_comp_mct.F90
+++ b/components/eamxx/src/mct_coupling/atm_comp_mct.F90
@@ -86,12 +86,13 @@ CONTAINS
     character(CS)                    :: run_type       ! type of run
     type(c_ptr) :: x2a_ptr, a2x_ptr
     character(len=256)               :: atm_log_fname  ! name of ATM log file
+    character(CL)                    :: calendar       ! calendar string
 
     ! TODO: read this from the namelist?
     character(len=256)                :: yaml_fname = "./data/scream_input.yaml"
     character(kind=c_char,len=256), target :: yaml_fname_c, atm_log_fname_c
     character(len=256) :: caseid, username, hostname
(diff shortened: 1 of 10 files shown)


AaronDonahue (approved):
I'm a bit surprised we have DIFFs in the mappy tests.  None of those tests run through a leap year right?

bartgol:
Yeah, diffs were not expected. I will investigate.

bartgol:
Ah, wait, this could be due to #2487 that got merged ~3h before this PR got tested. We'll bless baselines, and see if it passes.

bartgol:
@tcclevenger The failed test on weaver is `tms_tests_omp1`, due to a diff

```
/home/e3sm-jenkins/weaver/workspace/SCREAM_PullRequest_Autotester_Weaver/4863/scream/components/eamxx/src/physics/tms/tests/compute_tms_tests.cpp:79: FAILED:
  REQUIRE( d_f90.ksrf[i] == d_cxx.ksrf[i] )
with expansion:
  0.0001042416 == 0.0001042416
```
The PR does not change anything that should affect tms, so I wasn't expecting this. If one of the two sides was a NaN, I may say we hit an unlucky set of random input values. But here, the diff is at the roundoff level, which makes me wonder if there are some places in the tms code where we re-arranged inputs. Do you know if that's possible? It's not that obvious that's the case though, since I would have assumed we would have gotten a fail in our nightlies by now, if that was the case. But maybe this special value triggered some additional chunk of code? I don't know. If it helps, here's the rng seed to reproduce it.
```
For test compute_tms_bfb, random number generator seed: 1417415078
```
Could you please take a quick look?

tcclevenger:
@bartgol Taking a look now

tcclevenger:
Reproducible on master, so you are just getting unlucky here. Should be fine to merge if that's the only fail.

Issue #2501, opened 2023-08-31
When linking pyhton, `undefined symbol: EVP_KDF_CTX_new_id, version OPENSSL_1_1_1d`


ndkeen:
There was an issue on pm-cpu nightly tests after attempt to add in something new by @elynnwu that would link in python to exe. The tests work normally.  But for nightly testing, we actually first load e3sm_unified to run. The error is on compute node with immediate: `bld/e3sm.exe: symbol lookup error: /usr/lib64/libssh.so.4: undefined symbol: EVP_KDF_CTX_new_id, version OPENSSL_1_1_1d`  Debugging, I can reproduce the error by first loading e3sm_unified in terminal and launching the test. We see that there is a module at NERSC named evp-patch.  Doing a module show evp-patch:
```
Helper module that provides a workaround for cray-mpich / libfabric / conda env issues in SLES 15 SP4.

This module prepends a consistent set of openssl and libssh libraries (cray-mpich/libfabric dependencies) 
to LD_LIBRARY_PATH to avoid an incompatibility issue with openssl from conda envs and the libssh
on the system which can occur when using cray-mpich/libfabric inside a conda environment.
```

If I then load evp-patch and try test again, it works.  I see on NERSC channel: "HPE has said they plan to remove the library dependency from libfabric that is not compatible with the openssl library in some conda environments. Once that is done, I believe we can retire the evp-patch module"
I *think* evp-patch is auto-loaded when user requests a python module.

Potential solutions:
1) see if there's a solution around how the original link/test is working (I don't know details)
2) stop loading e3sm_unified to do nightly tests (i've tried suggesting this)
3) see if we can add the evp-patch module to e3sm_unified
4) add evp-patch module to our set of modules we load on perlmutter (temporarily)

I made a PR for (4). I think evp-patch is inert/safe, but as it would be used in normal runs (ie not nightly tests), we should do some tests.

PR #2502, opened 2023-08-31
Add evp-patch module to pm-cpu and pm-gpu

ndkeen:
For pm-cpu/pm-gpu, add evp-patch module temporarily.
Fixes https://github.com/E3SM-Project/scream/issues/2501
[bfb]


elynnwu (commented):
Thanks for opening this PR. I think you need to add this in `components/eamxx/scripts/machines_specs.py` too.

elynnwu (approved):

PR #2504, opened 2023-09-05
Use bfb math functions in TMS F90 code

tcclevenger:
BFB fail found in https://github.com/E3SM-Project/scream/pull/2500 (unrelated to that PR). I didn't use the bfb functions and wasn't caught until now. ~~This PR will not be BFB to previous eam code.~~

### DIFF
--- a/components/eam/src/physics/cam/trb_mtn_stress.F90
+++ b/components/eam/src/physics/cam/trb_mtn_stress.F90
@@ -1,7 +1,16 @@
+! Include bit-for-bit math macros.
+#include "bfb_math.inc"
+
 module trb_mtn_stress
 
+  ! Bit-for-bit math functions.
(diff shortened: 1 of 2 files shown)


bartgol (approved):
This should be bfb when building SCREAMv0 cases, since `SCREAM_CONFIG_IS_CMAKE` should not be defined, and all the bfb_XYZ calls will resolve to XYZ itself.

bartgol:
I'm puzzled by the diffs. I thought the F90 code was only run during the BFB unit tests. Also, we do not have a v1 test that enables tms anyways, so this is not clear to me.

tcclevenger:
@bartgol Oh yes, I forget that needs to be done. I can bless on mappy if you haven't already

bartgol:
I just fired off the bless job

PR #2505, opened 2023-09-05
atmchange: do not fail on missing entry when not buffering

### LINKED ISSUE #2427: When a new atm proc is added via atmchange, regenerate the intermediate xml file, to include its defaults
Currently, when one does something like
```
./atmchange physics::atm_procs_list=(shoc,cld,spa,p3,foo)
./atmchange foo::blah=2
```
we would get an error, since `foo::blah` is not found. That's because `atmchange` looks for the parameter in the intermediate XML file (in the case dir), which, if `foo` was not included in the current compset, would not contain the `foo` atm proc defaults. In order to re-generate the intermediate XML file (which would then contain the `foo` process section) one needs to re-run `buildnml`, e.g. via `case.setup`. This can be confusing, since it means we must run `case.setup` twice (the first time to get the atmchange script linked in the case folder, and the second to get the `foo` settings in the XML file in the case folder). The average user may be confused by this, and their productivity will be impacted.

There are a few options to fix this.

- Re-generate the intermediate XML file every time atmchange runs. This may make manual case handling a bit slower.
- When we create the intermediate XML file, we could also put the defaults for all atm procs in it, even if not currently included in the ATM ap list. Obv, we would put them in an XML section that would not be included in the final yaml input file.
- When we run atmchange, if an atm proc list is changed, we add some code that, if a new atm proc has been added, it grabs the defaults from the XML defaults file in the src tree.

I think the only use case that requires `case.setup` to be re-run is precisely the one described above. Adding a new atm proc (or even swapping one for another) is the only scenario where a new XML section needs to be added to the intermediate XML file. Therefore, I think we can handle this inside atmchange: if we detect that the entry to be changed is an `atm_procs_list`, we can scan the current XML file in the case folder, and update the sections of the atm procs. E.g., if we swap shoc with clubb, we should remove the shoc section, and add the clubb section, grabbing it from the defaults XML file in the src tree.

jgfouca:
If the user changed fields of an atm_proc, then removed the atm_proc, we don't want buffer applies to fail, so turn on missing_ok when no_buffer is on.

Always regenerate namelist.xml file after atmchange is run (unless it's buffer-only or no-buffer).

I also had to fix a number of doctests which appear to have been broken in previous PRs. We need to remember to run `cime-buildnml-tests` when we change these scripts.

Fixes #2427

### DIFF
--- a/components/eamxx/cime_config/eamxx_buildnml_impl.py
+++ b/components/eamxx/cime_config/eamxx_buildnml_impl.py
@@ -246,13 +246,22 @@ def refine_type(entry, force_type=None):
             expect (force_type is None or is_array_type(force_type),
                     "Error! Invalid type '{}' for an array.".format(force_type))
 
-            elem_type = force_type if force_type is None else array_elem_type(force_type)
-            result = [refine_type(item.strip(), force_type=elem_type) for item in entry.split(",") if item.strip() != ""]
-            result = make_array(result,'string' if elem_type is None else elem_type)
+            if force_type is None:
+                # Try to derive type from first element
+                elem_type = derive_type([item.strip() for item in entry.split(",")][0])
+            else:
+                elem_type = force_type if force_type is None else array_elem_type(force_type)
+
+            try:
+                result = [refine_type(item.strip(), force_type=elem_type) for item in entry.split(",") if item.strip() != ""]
+            except ValueError:
+                expect(False, "List '{}' has inconsistent types inside".format(entry))
+
+            result = make_array(result, "string" if elem_type is None else elem_type)
             expected_type = type(result[0])
             for item in result[1:]:
                 expect(isinstance(item, expected_type),
-                      "List '{}' has inconsistent types inside".format(entry))
+                       "List '{}' has inconsistent types inside".format(entry))
 
             return result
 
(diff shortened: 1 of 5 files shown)


components/eamxx/cime_config/eamxx_buildnml_impl.py
```diff
@@ -246,13 +246,22 @@ def refine_type(entry, force_type=None):
...
                     "Error! Invalid type '{}' for an array.".format(force_type))
 
-            elem_type = force_type if force_type is None else array_elem_type(force_type)
-            result = [refine_type(item.strip(), force_type=elem_type) for item in entry.split(",") if item.strip() != ""]
-            result = make_array(result,'string' if elem_type is None else elem_type)
+            if force_type is None:
+                # Try to derive type from first element
+                elem_type = derive_type([item.strip() for item in entry.split(",")][0])
+            else:
+                elem_type = force_type if force_type is None else array_elem_type(force_type)
```
bartgol, line 253: Since you're in the `else` branch, you already know force_type is not None.

bartgol (changes requested):

components/eamxx/cime_config/eamxx_buildnml_impl.py
jgfouca: Good catch, thx

bartgol (dismissed):

bartgol (approved):

bartgol (approved):

PR #2506, opened 2023-09-05
[WIP] Verify array is allocated before accessing size

tcclevenger:
In EAMxx, if this output array is not allocated, then this resulted in a valgrind error when attempting to access `size(cospOUT%isccp_meantb)`. This is a possible solution, just asking if `cospOUT%isccp_meantb` is associated with space in memory (follows format of surrounding code).

@brhillman Does this work?

### DIFF
--- a/components/eam/src/physics/cosp2/local/cosp.F90
+++ b/components/eam/src/physics/cosp2/local/cosp.F90
@@ -478,7 +478,9 @@ CONTAINS
        Lrttov_column    = .true.
 
     ! Set flag to deallocate rttov types (only done on final call to simulator)
-    if (size(cospOUT%isccp_meantb) .eq. stop_idx) lrttov_cleanUp = .true.
+    if (associated(cospOUT%isccp_meantb)) then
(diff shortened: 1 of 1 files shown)


brhillman:
Yeah, this should fix it, but this does affect EAM code. We should probably plan to do an upstream merge after this, or issue the change to E3SM-Project/E3SM as well.

brhillman (approved):
This looks good to me; this should also be fixed upstream in EAM, and we should plan to either do a merge up to E3SM after this PR goes in, or issue an additional PR to E3SM repo.

brhillman:
Removing automerge for a moment...since this changes EAM code rather than EAMxx code, should this PR instead go to E3SM repo and then pull down with an upstream merge?

tcclevenger:
This is incorporated as part of upstream merge https://github.com/E3SM-Project/scream/pull/2511#issuecomment-1710707247

tcclevenger:
This is incorporated as part of upstream merge https://github.com/E3SM-Project/scream/pull/2511#issuecomment-1710707247

PR #2507, opened 2023-09-05
Fix bug in coarsening remapper

### LINKED ISSUE #2439: regridded ne120 output lat and lon coordinates are zero
I noticed recently that the lat and lon coordinates in the regridded ne120 v1 output are all set to zero everywhere.

bartgol:
The bug was subtle, in that whether it exhibit itself or not depended heavily on machine/compiler settings, and on runtime "luck". Basically, we were firing up MPI sends without ensuring that all the packing operations had completed. Adding a fence after the packing loop ensures that we are sending the correct data.

Fixes #2439 

I also threw in fixes for a couple of bugs I ran into while debugging this.

Note: I would _not_ trust _any_ remapped output before this PR. While you may see non-zero values, I cannot guarantee that the values where correct. It is possible that the output at a certain timestep used coarsened values from the previous time step, b/c this timestep's MPI sends were fired off _before_ the packing operation completed.

### DIFF
--- a/components/eamxx/src/dynamics/homme/eamxx_homme_process_interface.cpp
+++ b/components/eamxx/src/dynamics/homme/eamxx_homme_process_interface.cpp
@@ -426,6 +426,12 @@ void HommeDynamics::initialize_impl (const RunType run_type)
     restart_homme_state ();
   }
 
+  // Since we just inited them, ensure p_mid/p_int (dry and wet) timestamps are valid
+  get_field_out("p_int")    .get_header().get_tracking().update_time_stamp(timestamp());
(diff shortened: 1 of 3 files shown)


AaronDonahue:
wow! Nice catch @bartgol.  I'm glad you caught this before the Cess runs.

tcclevenger (approved):
Nice

bartgol:
The diffs are expected: since p_mid/p_int did not have a valid stamp, they were not saved at t=0, and neither was PotentialTemperature (since its required field p_mid was not "valid"). Notice that the ERS/ERP tests are using AVERAGE output, which does not do t=0 output, so the diff only appears for SMS/PET tests.

PR #2508, opened 2023-09-06
Fix cloud diagnostics when using chunks

brhillman:
Fix cloud diagnostic outputs when using chunks. PR #2485 modified these to avoid having to do deep copies of the data, but neglected to account for the fact that in the chunk loop in radiation they are treated as having size equal to current chunk size rather than size m_ncol. We can still directly use the data from the field manager, but we need to offset the pointer based on where we are in the chunk loop. This PR also adds the aerocom diagnostics to the RESTART group. This should be required (for now, at least until I/O can handle masked aggregation/averages) to account for the possibility that restarts do not line up with radiation update steps.


brhillman:
Illustration of the problem this PR fixes. With the recent changes to cloud diagnostics, when NTASKS was small enough to use chunks, big blocks would be left undefined:
image
After the fix, things look like we "expect" (other than being overly large due to using the internal ice cloud mask that is sensitive to even very small amounts of ice):
image
After the fix, cloud top diagnostics also look reasonable:
image

bartgol (approved):

PR #2510, opened 2023-09-06
Change in atm proc group specification, and upgrade to atmchange

### LINKED ISSUE #2305: Switch `atm_procs_list` in EAMxx control file to a vector of strings
Currently EKAT is set to parse the `atm_procs_list` namelist variable in EAMxx as a single string, which is then further parsed internally into a list of processes.  This can be counterintuitive to users who likely think of this as a list that needs to be changed as a list would be.  Similar to how the `output_yaml_files` namelist parameter is changed.

The task is to update the parsing of the EAMxx namelist file to treat `atm_procs_list` as a vector of strings.

bartgol:
This PR does a few things:

- changes the type of `atm_procs_list` parameter for `AtmosphereProcessGroup` to `array(string)`. As discussed in #2305 , this prevents us from declaring arbitrary sub groups on the fly, via something like `./atmchange group::atm_procs_list=(homme,(shoc,p3),rrtmgp)`. This is why this PR also...
- upgrades atmchange to handle an arbitrary change of an atm proc group. In particular, the user can declare a new atm proc group, and add/remove atm processes from an existing group via `./atmchange group=ap1,ap2,ap3`. If a process in the list is in our XML defaults, buildnml will simply grab that. If not, we interpret that as a request to declare a new group with the given name. The new group will have NO atm procs inside, and will have to be updated later via a similar call. For instance, the current list of atm procs in physics is `mac_aero_mic, rrtmgp`. To change radiation with, say, the subgroup `rrtmgp,cosp`, one can do
```
$ ./atmchange physics=mac_aero_mic,my_rad
$ ./atmchange my_rad=rrtmgp,cosp
```
Note: this _can_ be risky, if the user simply misspelled an atm proc. E.g, the following
```
$ ./atmchange physics=mac_aero_mic,rtmgp,cosp
```
(notice the misspelled rrtmgp) will be interpreted as `mac_aero_mic`, new (currently empty) group called `rtmgp`, and `cosp`, which is not what the user wants. For this reason, I recommend using `./atmchange group_name=ap1,ap2,ap3,...` only if you intend to _declare a new group on the fly`. If you just want to change the list, using existing atm procs, you should do what you were already doing, namely `./atmchange group_name::atm_procs_list=ap1,ap2,ap3`. Notice that the call `./atmchange physics::atm_procs_list=mac_aero_mic,rtmgp,cosp` would fail, since `rtmgp` is not a valid process, and this syntax assumes all the ap's in the list are existing in the defaults XML file.

- `atm_procs_list` was the only use case of parameters with parentheses/brackets. Therefore, I removed support for parsing xml entries that start/end with `[`/`]` or `(`/`)` as arrays. I also don't parse automatically a string such as `a,b,c` as an array. We _must_ set `type="array(X)"` in the XML default, in order to trigger parsing as an array (this also indirectly helps with our automatic parameters doc generation, since it specifically states the type rather than relying on automatic detection, buried deep in the code).

Closes #2305


bartgol:
I have considered adding some flag to make explicit that we want to define a new group, something like `--add-apg`, but the way we buffer changes makes this flag parsed as a change itself, which ends up failing. Another option would be to use some special character to declare a new group. Something like `physics=mac_aero_mic,$my_rad`. I may actually do this, since it reduce the chance of bugs.

bartgol:
@jgfouca I think I should add some doctests to the scripts. I also want to extract all our XML utils and put them in a module, since I think there's some overlap between stuff in `cime_config` and stuff in `scripts`.

jgfouca (changes requested):
Thanks! It has bugged me for a long time that proc lists were strings and not lists. You will need to run cime-nml-tests and make fixes there because it is still using the old syntax in a few places.

bartgol:
Added some docstring test for the new python fcn, and `cime-nml-tests` now passes.

jgfouca (dismissed):

bartgol:
Shout out to github action that automatically updates our docs, which helped finding a bug in one of the buildnml scripts:

https://github.com/E3SM-Project/scream/actions/runs/6113064621/job/16591795995?pr=2510

jgfouca (approved):

PR #2514, opened 2023-09-07
EAMxx: fix bug in scorpio output

bartgol:
Always use the field name from the field object as key in maps. That's b/c field.name() may differ from the string we received in the input file if the field is a diagnostic.


AaronDonahue (dismissed):

ndkeen (dismissed):
Using these changes, I can run a ne256 case on frontier for 5 days that was previously failing with the "map" error.

AaronDonahue (approved):

AaronDonahue (approved):

PR #2518, opened 2023-09-11
Removing old PIO modules in EAMXX

jayeshkrishna:
Ensure that all PIO functions are included via the "pio" module.


jayeshkrishna:
This change should work with the existing SCORPIO release included in the source and the SCORPIO master (The SCORPIO master includes refactoring of the SCORPIO Fortran interface and requires this change)

bartgol (approved):

PR #2520, opened 2023-09-11
Unify implementation of some diagnostics and their tests

bartgol:
In particular,

- `[Liq|Ice|Rain|Rime|Vapor]WaterPath`: they only differ in the tracer they use
- `precip_[liq|ice|total]_surf_mass_flux`: these were already unified, but their unit tests were not
- `[Meridional|Zonal]VaporFlux`: they only differ in the component of `horiz_winds` used.

A single impl reduces the amount of code to maintain.


AaronDonahue (dismissed):
I like this simplification.  Should be much easier to follow.  Thanks Luca!

AaronDonahue (approved):

PR #2522, opened 2023-09-12
Fixes an inconsistency between standard IO and vertical remap files.

AaronDonahue:
This commit fixes an inconsistency between the typical EAMxx output which uses the dimension name `lev` for levels and the vertical remapping files which uses `nlevs`.  This isn't a major bug, but the inconsistency can cause some confusion and more importantly, if someone uses EAMxx output for nudging they would have a pre-processing step of adding a dimension `levs` to the file containing `p_levs` just to work.

### DIFF
--- a/components/eamxx/src/share/grid/remap/vertical_remapper.cpp
+++ b/components/eamxx/src/share/grid/remap/vertical_remapper.cpp
@@ -54,7 +54,7 @@ VerticalRemapper (const grid_ptr_type& src_grid,
   // as the source field, but will have a different number of 
   // vertical levels.
   scorpio::register_file(map_file,scorpio::FileMode::Read);
-  m_num_remap_levs = scorpio::get_dimlen(map_file,"nlevs");
+  m_num_remap_levs = scorpio::get_dimlen(map_file,"lev");
 
   auto tgt_grid_gids = src_grid->get_unique_gids();
   const int ngids = tgt_grid_gids.size();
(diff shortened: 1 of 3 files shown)


AaronDonahue:
This commit will require changes to our vertical remapping source files.  In the long-run I think we want this fix, but it might be best to hold off on merging this until the CESS simulations are finished.  Pinging @crterai for his opinion.

Note, the change is pretty straightforward.  We would need to change our vertical remapping file to use the name `lev` instead of `nlevs` for the dimension.

crterai:
Thanks for pinging me, @AaronDonahue. I'm hesitant to include this in our Cess simulations, since it's another thing we need to create (vertical levels files), add in yaml files, and test. At the same time, I realize post-processing data can take time. Is the post-processing step of changing from lev to nlev currently a big post-processing step? 

In either case, I think it should go into master. At this point, with Luca's scream output fixes having gone into master, I can merge current master into machines/frontier and then cherry-pick the needed PRs.

elynnwu:
> At the same time, I realize post-processing data can take time. Is the post-processing step of changing from lev to nlev currently a big post-processing step?

I think we don't even need to rename for the CESS simulation output. This change is just to make it consistent with the nudging process where we read the `p_levs` from a static file that is supposed to have dimension of `lev` instead of `nlev`.

bartgol:
If we incorporate this, I think all we need to do is update the vert remap nc file. But that means we need to change all the yaml files that use it (since it must have a timestamp in the filename).

@crterai I don't think it's a lot of work, I can help with that. But if you don't want to use this in the CESS runs, it's ok.

elynnwu (approved):
Completed a run with ne30 coarsened to ne4 using an updated map file `/usr/gdata/climdat/eamxx-ml/maps/map_ne30pg2_to_ne4pg2_20230912_vert_enhanced.nc` with `p_levs(lev)`.

AaronDonahue:
I'm totally okay tabling this until after the CESS runs.  Elynn caught the inconsistency today and I wanted to make sure the fix got on the board before we forgot.  After our eval call today I think this PR is a perfect example of making a change that disrupts the workflow and could potentially slow things down leading into Cess runs this/next week.  There's no rush on this.

AaronDonahue:
Now that we have a CESS production branch we can go ahead and merge this into master without impacting those runs.  @crterai and @tcclevenger , pinging both of you to make you aware that if this does go into `master` we want to make sure it doesn't accidentally get into the cess production cherry-pick.

Issue #2381, opened 2023-06-13
Model crash at ne30 -- T above 500k (likely fixed with macmic=12)


golaz:
My latest simulation with EAMxx at ne30 crashed a little over after one year. I have performed a number of ne30 simulations before, none of which failed. I've verified that the crash is reproducible.

The model fails because of excessively large temperature 514 K) in BC, Canada on April 1

From atm.log

```
Atmosphere step = 21839
  model time = 0002-03-31 23:30:00

[EAMxx::output_manager] - Writing model output:
[EAMxx::output_manager]      CASE: 20230607.F2010-SCREAMv1.ne30pg2_ne30pg2.chrysalis.scream.monthly
[EAMxx::output_manager]      FILE: 20230607.F2010-SCREAMv1.ne30pg2_ne30pg2.chrysalis.scream.monthly.AVERAGE.nmonths_x1.0002-03-01-00000.nc
Atmosphere step = 21840
  model time = 0002-04-01 00:00:00

```

From e3sm.log

```
 858: terminate called after throwing an instance of 'std::logic_error'
 858:   what():  /gpfs/fs1/home/ac.golaz/E3SM/EAMxx/code/20230607/components/eamxx/src/share/atm_process/atmosphere_process.cpp:442: FAIL:
 858: false
 858: Error! Failed post-condition property check (cannot be repaired).
 858:   - Atmosphere process name: shoc
 858:   - Property check name: T_mid within interval [100, 500]
 858:   - Atmosphere process MPI Rank: 858
 858:   - Message: Check failed.
 858:   - check name: T_mid within interval [100, 500]
 858:   - field id: T_mid[Physics PG2] <double:ncol,lev>(12,128) [K]
 858:   - minimum:
 858:     - value: 211.163
 858:     - entry: (20541,60)
 858:     - lat/lon: (56.7783, 238.776)
 858:   - maximum:
 858:     - value: 514.47
 858:     - entry: (20663,127)
 858:     - lat/lon: (54.7812, 232.505)
```

Simulation output is available on chrysalis:
 `/lcrc/group/e3sm/ac.golaz/E3SM/EAMxx/20230607.F2010-SCREAMv1.ne30pg2_ne30pg2.chrysalis`

Run script is attached: 
[...]

PR #2523, opened 2023-09-12
Change default ne30 macmic to 12 (from 6)

ndkeen:
For ne30, we found that using macmic=12 allowed long-running cases to avoid high T error.

NBFB at ne30

Fixes https://github.com/E3SM-Project/scream/issues/2381

### DIFF
--- a/components/eamxx/cime_config/namelist_defaults_scream.xml
+++ b/components/eamxx/cime_config/namelist_defaults_scream.xml
@@ -288,7 +288,7 @@ be lost if SCREAM_HACK_XML is not enabled.
       <atm_procs_list>(shoc,cldFraction,spa,p3)</atm_procs_list>
       <atm_procs_list COMPSET=".*SCREAM.*noAero">(shoc,cldFraction,p3)</atm_procs_list>
       <number_of_subcycles hgrid="ne4np4">24</number_of_subcycles>
-      <number_of_subcycles hgrid="ne30np4">6</number_of_subcycles>
+      <number_of_subcycles hgrid="ne30np4">12</number_of_subcycles>
       <number_of_subcycles hgrid="ne120np4">3</number_of_subcycles>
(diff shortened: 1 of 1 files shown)


AaronDonahue (dismissed):

components/eamxx/cime_config/namelist_defaults_scream.xml
```diff
@@ -288,7 +288,7 @@ be lost if SCREAM_HACK_XML is not enabled.
       <atm_procs_list>(shoc,cldFraction,spa,p3)</atm_procs_list>
       <atm_procs_list COMPSET=".*SCREAM.*noAero">(shoc,cldFraction,p3)</atm_procs_list>
       <number_of_subcycles hgrid="ne4np4">24</number_of_subcycles>
-      <number_of_subcycles hgrid="ne30np4">6</number_of_subcycles>
+      <number_of_subcycles hgrid="ne30np4">12</number_of_subcycles>
       <number_of_subcycles hgrid="ne120np4">3</number_of_subcycles>
```
PeterCaldwell, line 292: We also need to change the ne120 value from 3 to 6 since default full-model dt for ne120 is 15 min so 3 substeps = 5 min, which is too long. We also have evidence of ne120 crashes from this problem as well (from Elynn).

ndkeen: Yes, I can also push a commit to do that, but I don't want to hold this up as currently, ne30 is broken on master and has been for a long time.  Just change above 
`<number_of_subcycles hgrid="ne120np4">3</number_of_subcycles>`
to
`<number_of_subcycles hgrid="ne120np4">6</number_of_subcycles>`
?

PeterCaldwell: Yup, perfect. It looks like you already made the change. Thanks! ne120 was just as broken as ne30, so it is good to fix both.

PeterCaldwell (approved):

PR #2525, opened 2023-09-12
Update check_flux_state_consistency() computation

### LINKED ISSUE #2372: Convert qneg4 routine to C++

tcclevenger:
@oksana I update the computation as we talked about. 
Notes/Questions: 
1. What is a good way for me to test this? I ran two 5 day simulation with these checks on and off. Neither has mass errors over `1e-10`, and both simulations had energy errors around `1e-8`, no difference between with or without checks.
2. Where these checks are performed I don't think we have access to the number of mac_aero_mic steps. The number of steps aren't used anywhere in the shoc code to my knowledge (outside of it being contained in `dt`). It would be easy to propagate this information, but where are you suggesting we use it?
3. Should we be using the updated `surf_evap_intput` in the mass and energy checks for shoc, or the original `surf_evap`?

### DIFF
--- a/components/eamxx/src/physics/shoc/eamxx_shoc_process_interface.cpp
+++ b/components/eamxx/src/physics/shoc/eamxx_shoc_process_interface.cpp
@@ -530,18 +530,27 @@ void SHOCMacrophysics::check_flux_state_consistency(const double dt)
       return qv_i(k)*pseudo_density_i(k);
     };
 
+    // Define column_vapor_mass on surface
+    const auto column_vapor_mass_sfc = column_vapor_mass_on_level(last_pack_idx)[last_pack_entry];
+
     // Define surface evaporative mass
-    auto surf_evap_mass = modified_surf_evap(i)*dt*gravit;
+    const auto surf_evap_mass = modified_surf_evap(i)*dt*gravit;
 
     // Check if the negative surface latent heat flux can exhaust
     // the moisture in the lowest model level. If so, apply fixer.
-    const auto excess_mass = surf_evap_mass + column_vapor_mass_on_level(last_pack_idx)[last_pack_entry];
+    const auto excess_mass = surf_evap_mass + column_vapor_mass_sfc;
     if (excess_mass < 0) {
       // Calculate the total column vapor mass
       Real total_column_vap_mass = ekat::ExeSpaceUtils<KT::ExeSpace>::view_reduction(team, 0, nlevs, column_vapor_mass_on_level);
-      EKAT_KERNEL_ASSERT_MSG(total_column_vap_mass >= surf_evap_mass,
+      EKAT_KERNEL_ASSERT_MSG(total_column_vap_mass >= std::abs(surf_evap_mass),
                             "Error! Total mass of column vapor should be greater "
-                            "than mass of surf_evap.\n");
+                            "than the magnitude of  mass of surf_evap.\n");
+
+      // Set modified_surf_evap to exactly exhaust the moisture at the bottom layer
+      modified_surf_evap(i) = -1.0*column_vapor_mass_sfc/(dt*gravit);
+
+      // Set total_column_vap_mass to only be from levels 0:nlev-2 (exclude sfc value)
+      total_column_vap_mass -= column_vapor_mass_sfc;
 
       // Redistribute excess mass
       Kokkos::parallel_for(Kokkos::TeamVectorRange(team, nlev_packs), [&](const int& k) {
@@ -549,12 +558,9 @@ void SHOCMacrophysics::check_flux_state_consistency(const double dt)
         const auto range_pack = ekat::range<IntSmallPack>(k*Spack::n);
         const Smask index_above_sfc = range_pack < nlevs-1;
 
-        const auto adjust = std::abs(surf_evap_mass)*column_vapor_mass_on_level(k)/total_column_vap_mass;
-        qv_i(k).set(index_above_sfc, (column_vapor_mass_on_level(k) - adjust)/pseudo_density_i(k));
+        const auto adjust = excess_mass*column_vapor_mass_on_level(k)/total_column_vap_mass;
+        qv_i(k).set(index_above_sfc, (column_vapor_mass_on_level(k) + adjust)/pseudo_density_i(k));
       });
-
-      // Set modified_surf_evap to exactly exhaust the moisture at the bottom layer
-      modified_surf_evap(i) = -1.0*column_vapor_mass_on_level(last_pack_idx)[last_pack_entry]/(dt*gravit);
     }
   });
 }
--- a/components/eamxx/src/physics/shoc/eamxx_shoc_process_interface.cpp
+++ b/components/eamxx/src/physics/shoc/eamxx_shoc_process_interface.cpp
@@ -64,18 +64,18 @@ void SHOCMacrophysics::set_grids(const std::shared_ptr<const GridsManager> grids
   const auto s2 = s*s;
 
   // These variables are needed by the interface, but not actually passed to shoc_main.
-  add_field<Required>("omega",               scalar3d_layout_mid,  Pa/s,    grid_name, ps);
-  add_field<Required>("surf_sens_flux",      scalar2d_layout_col,  W/m2,    grid_name);
-  add_field<Required>("surf_mom_flux",       surf_mom_flux_layout, N/m2, grid_name);
-
-  add_field<Updated>("surf_evap",           scalar2d_layout_col,  kg/m2/s, grid_name);
-  add_field<Updated> ("T_mid",               scalar3d_layout_mid,  K,       grid_name, ps);
-  add_field<Updated> ("qv",                  scalar3d_layout_mid,  Qunit,   grid_name, "tracers", ps);
+  add_field<Required>("omega",          scalar3d_layout_mid,  Pa/s,    grid_name,            ps);
+  add_field<Required>("surf_sens_flux", scalar2d_layout_col,  W/m2,    grid_name);
+  add_field<Required>("surf_evap",      scalar2d_layout_col,  kg/m2/s, grid_name);
+  add_field<Required>("surf_mom_flux",  surf_mom_flux_layout, N/m2,    grid_name);
 
   // If TMS is a process, add surface drag coefficient to required fields
   if (m_params.get<bool>("apply_tms", false)) {
     add_field<Required>("surf_drag_coeff_tms", scalar2d_layout_col,  kg/s/m2, grid_name);
   }
+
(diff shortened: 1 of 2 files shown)


oksanaguba:
tagging here previous discussions https://github.com/E3SM-Project/scream/pull/2418 .

components/eamxx/src/physics/shoc/eamxx_shoc_process_interface.cpp
```diff
@@ -239,7 +240,6 @@ void SHOCMacrophysics::initialize_impl (const RunType run_type)
   const auto& pseudo_density      = get_field_in("pseudo_density").get_view<const Spack**>();
   const auto& omega               = get_field_in("omega").get_view<const Spack**>();
   const auto& surf_sens_flux      = get_field_in("surf_sens_flux").get_view<const Real*>();
-  const auto& surf_evap           = get_field_in("surf_evap").get_view<const Real*>();
```
oksanaguba, line 242: needs to be grabbed back

components/eamxx/src/physics/shoc/eamxx_shoc_process_interface.cpp
```diff
@@ -376,7 +377,7 @@ void SHOCMacrophysics::initialize_impl (const RunType run_type)
     const auto& water_flux = get_field_out("water_flux").get_view<Real*>();
     const auto& ice_flux   = get_field_out("ice_flux").get_view<Real*>();
     const auto& heat_flux  = get_field_out("heat_flux").get_view<Real*>();
-    shoc_postprocess.set_mass_and_energy_fluxes (surf_evap, surf_sens_flux,
+    shoc_postprocess.set_mass_and_energy_fluxes (surf_evap_input, surf_sens_flux,
```
oksanaguba, line 380: use original surf_evap here.

components/eamxx/src/physics/shoc/eamxx_shoc_process_interface.cpp
```diff
@@ -499,17 +506,16 @@ void SHOCMacrophysics::check_flux_state_consistency(const double dt)
 {
   using PC = scream::physics::Constants<Real>;
   const Real gravit = PC::gravit;
-  const Real qmin   = 1e-12; // minimum permitted constituent concentration (kg/kg)
```
oksanaguba, line 502: @bogensch @AaronDonahue @PeterCaldwell this lower bound was taken from EAM setup (iirc EAM won't run if this bount is set to 0.0). Do we need a certain nontrivial bound in scream or can we always assume lower acceptable bounds for tracers are zero. In this case it is only for vapor. If a certain bound is needed, Conrad will put it in physics constants.

components/eamxx/src/physics/shoc/eamxx_shoc_process_interface.cpp
```diff
@@ -518,28 +524,35 @@ void SHOCMacrophysics::check_flux_state_consistency(const double dt)
...
-      };
-      Real mm = ekat::ExeSpaceUtils<KT::ExeSpace>::view_reduction(team, 0, nlevs, tracer_mass);
-
-      EKAT_KERNEL_ASSERT_MSG(mm >= cc, "Error! Total mass of column vapor should be greater than mass of surf_evap.\n");
-
-      Kokkos::parallel_for(Kokkos::TeamVectorRange(team, nlev_packs), [&](const int& k) {
-        const auto adjust = cc*qv_i(k)*pseudo_density_i(k)/mm;
-        qv_i(k) = (qv_i(k)*pseudo_density_i(k) - adjust)/pseudo_density_i(k);
+    const auto excess_mass = surf_evap_mass + column_vapor_mass(last_pack_idx)[last_pack_entry];
+    if (excess_mass < 0) {
```
oksanaguba, line 538: qmin?

components/eamxx/src/physics/shoc/eamxx_shoc_process_interface.cpp
```diff
@@ -518,28 +524,35 @@ void SHOCMacrophysics::check_flux_state_consistency(const double dt)
     const auto& pseudo_density_i = ekat::subview(pseudo_density, i);
     const auto& qv_i             = ekat::subview(qv, i);
 
-    // reciprocal of pseudo_density at the bottom layer
-    const auto rpdel = 1.0/pseudo_density_i(last_pack_idx)[last_pack_entry];
+    // Define lambda to calculate column vapor mass
+    auto column_vapor_mass = [&](const int k) {
```
oksanaguba, line 528: add function in the mname then?

components/eamxx/src/physics/shoc/eamxx_shoc_process_interface.cpp
```diff
@@ -518,28 +524,35 @@ void SHOCMacrophysics::check_flux_state_consistency(const double dt)
...
+    const auto excess_mass = surf_evap_mass + column_vapor_mass(last_pack_idx)[last_pack_entry];
+    if (excess_mass < 0) {
+      // Calculate the total column vapor mass
+      Real total_column_vap_mass = ekat::ExeSpaceUtils<KT::ExeSpace>::view_reduction(team, 0, nlevs, column_vapor_mass);
+      EKAT_KERNEL_ASSERT_MSG(total_column_vap_mass >= surf_evap_mass,
+                            "Error! Total mass of column vapor should be greater "
+                            "than mass of surf_evap.\n");
+
+      // Redistribute excess mass
+      Kokkos::parallel_for(Kokkos::TeamVectorRange(team, nlev_m_1_packs), [&](const int& k) {
```
oksanaguba, line 546: excess mass needs to be used here

components/eamxx/src/physics/shoc/eamxx_shoc_process_interface.cpp
oksanaguba: maybe it is a question for @brhillman instead, since in EAM the comment mentions radiation
```
    real(r8),intent(in)    :: qminc  ! minimum value of mass mixing ratio (kg/kg)
                                     ! normally 0., except water 1.E-12, for radiation.
```

oksanaguba:
@tcclevenger  here is the code i ran in EAM (in place of qneg4 call in physpkg line 1767)
```
!vapor hard set bound
bb = 1e-12_r8
!convert to/from Pa
ffactor = ztodt*gravit

do i=1,ncol
  !convert cflx in Pa units
  cc = cam_in%cflx(i,1)*ffactor

  qqdp1 = state%q(i,pver,1)*state%pdel(i,pver)
  !if adding cflx to the bottom violates the bound
  if ( (cc + qqdp1) < bb) then

    !print *, 'OG cc value is ', cc
    cc=abs(cc)

    !comute total Pa mass of water vapor in column
    mm=0.0_r8
    do k=1,pver
      mm = mm + state%q(i,k,1)*state%pdel(i,k)
    enddo
    mass_before = mm + cam_in%cflx(i,1)*ffactor

    !set cflx to the value that won't violate bb bound
    cam_in%cflx(i,1) = (bb - qqdp1)/ffactor

    if (mm < cc) then
      !abort, cflx value is negative and by abs is larger than total mass of vapor in column
      stop
    else
      !compute mass ot 1:plev-1 column
      mm = mm - qqdp1
      !redistribute excess = -cc - (bb - qqdp1)
      !check: excess + cflx*ffactor = -cc (this is what we want)
      !cc is negative excess now
      cc = -cc - (bb - qqdp1)
      do k=1,pver-1
        qq = state%q(i,k,1)
        pp = state%pdel(i,k)
        adjust = cc *  qq*pp/mm
        state%q(i,k,1) = (qq*pp + adjust)/pp
      enddo
    endif
!print *, "OG cc, mm, new mm",cc,mm,sum(state%q(i,:,1)*state%pdel(i,:))
!print *, 'OG mass before, after',mass_before, (sum(state%q(i,:,1)*state%pdel(i,:)) + (bb - qqdp1))
!print *, 'OG mass before - after',mass_before - (sum(state%q(i,:,1)*state%pdel(i,:)) + (bb - qqdp1))
!with right code rel error is ~1e-16
!print *, 'OG rel mass before - after',(mass_before - (sum(state%q(i,:,1)*state%pdel(i,:)) + (bb - qqdp1)))/mass_before
  endif
enddo
```

oksanaguba:
The code sets cflx to an acceptable value and redistributes leftovers in the column 1:plev-1. The mass check for the code is rel error ~1e-16.
Diagnostics for qneg4 and the new code is in this plot
[WATER-aaa.pdf](https://github.com/E3SM-Project/scream/files/12771891/WATER-aaa.pdf)

Here green line is the old code, blue line is the new. Diagnostics is dWater minus (expected fluxes). Fluxes values for these runs are ~1e-3 to 1e-1. One interesting detail is that the run above is for macmic=1, not for the default macmic=6. With macmic=6, the diagnostics for the new code is actually worse. In EAM case the code does not take into account that there are many iterations of macmic loop; i did not try to understand it better than that, i just switched to 1.

oksanaguba:
i forgot about mass_borrower for the runs above, will repeat them.

AaronDonahue:
@oksanaguba and @tcclevenger , what is the status of this PR?  Is it still WIP?

tcclevenger:
@AaronDonahue For now still WIP, I need to implement Oksana's suggestions. I'll try and get to that this week.

tcclevenger:
@oksanaguba Could you take a look at this again and maybe test? I updated the computation based on your code.

oksanaguba:
@tcclevenger thanks, i cannot do it immediately, maybe in 1-2 weeks.

AaronDonahue:
@tcclevenger , is there any reason to keep this PR open?  If so, can it be revived and updated w/ the latest master?

tcclevenger:
> @tcclevenger , is there any reason to keep this PR open? If so, can it be revived and updated w/ the latest master?

@oksanaguba Is there any reason to keep this open? Can't remember if we were going a different direction.

PR #2526, opened 2023-09-13
minor fixes and formatting in output docs

mahf708:
Hopefully better formatting of output docs

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -10,7 +10,7 @@ on:
     # Only if docs-related files are touched
     paths:
       - components/eamxx/mkdocs.yml
-      - components/eamxx/docs/*
+      - components/eamxx/docs/**
       - components/eamxx/cime_config/namelist_defaults_scream.xml
   # Runs every time a PR is open against master
   pull_request:
@@ -18,7 +18,7 @@ on:
     # Only if docs-related files are touched
     paths:
       - components/eamxx/mkdocs.yml
-      - components/eamxx/docs/*
+      - components/eamxx/docs/**
       - components/eamxx/cime_config/namelist_defaults_scream.xml
 
   workflow_dispatch:
@@ -34,7 +34,7 @@ jobs:
 
     steps:
       - name: Check out the repository
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
         with:
           # TODO: git rid of dependency on CIME
           # TODO: another option to investigate is a sparse checkout. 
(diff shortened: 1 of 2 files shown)


.github/workflows/gh-pages.yml
```diff
@@ -10,15 +10,15 @@ on:
     # Only if docs-related files are touched
     paths:
       - components/eamxx/mkdocs.yml
-      - components/eamxx/docs/*
+      - components/eamxx/docs/**
```
bartgol, line 13: Why the double star?

components/eamxx/docs/user/model_output.md
```diff
@@ -52,48 +54,52 @@ The other parameters are
...
+    - `IceWaterPath`
+    - `VapWaterPath`
+    - `RainWaterPath`
+    - `RimeWaterPath`
+    - `ShortwaveCloudForcing`
+    - `LongwaveCloudForcing`
+    - `RelativeHumidity`
+    - `ZonalVapFlux`
+    - `MeridionalVapFlux`
+    - `precip_surf_mass_flux`
```
bartgol, line 88: We should update this. I mean, I should have when I submitted the or, but forgot :)

Can you remove this and add 3 entries for `precip_X_surf_mass_flux`, with X `liq`, `ice`, and `total`?

bartgol (changes requested):

bartgol:
> I noticed that the ruleset around two spaces only for nested lists doesn't render correctly in mkdocs (it renders correctly usually in regular markdown, so I have no idea what's going on 🤷)

What's a double space supposed to be rendered as?

components/eamxx/docs/user/model_output.md
bartgol: Oh wow, I didn't notice the lack of list for that page. Googling, I found [this cheat sheet](https://yakworks.github.io/docmark/cheat-sheet/) for mkdocs, and they state clearly that "need a blank line above to start new list", and that you need "4 spaces or 1 tab to indent".

.github/workflows/gh-pages.yml
bartgol: Ah, didn't know about the `**` needed to include `/` in the matched pattern. Good to know.

bartgol: I have no clue why that PR triggered the action. Maybe _at the time the PR was issued_ it included a few commits (from another PR) that touched them, but as soon as that PR got merged, they went away?

PR #2527, opened 2023-09-13
fix shoc energy fixer

oksanaguba:
Fixing energy leaks for both F and CXX: the issue was the same in both -- instead of using raw coupler fluxes for the energy fixer, shoc recomputes them back from wpthlp_sfc (problematic) and wprtp_sfc (not an issue). The first one was missing exner factor after we transitioned to the unapproximated version of thetal. The factor was missing only in how the energy diff is computed for the fixer.

Separately, there was a SW bug in the F version, uninit-ed det_ice and det_s, which lead to false energy check errors. i don't think that issue was affecting anything besides diagnostics.

[nonbfb] for baseline tests.

### DIFF
--- a/components/eam/src/physics/cam/shoc.F90
+++ b/components/eam/src/physics/cam/shoc.F90
@@ -14,6 +14,7 @@ module shoc
 
   use physics_utils, only: rtype, rtype8, itype, btype
   use scream_abortutils, only: endscreamrun
+  use physconst,     only: rair, cpair
 
 ! Bit-for-bit math functions.
 #ifdef SCREAM_CONFIG_IS_CMAKE
@@ -30,6 +31,9 @@ logical :: use_cxx = .true.
 
 real(rtype), parameter, public :: largeneg = -99999999.99_rtype
 real(rtype), parameter, public :: pi = 3.14159265358979323_rtype
+
+!repeated from shoc_intr!
+real(rtype), parameter         :: p0_shoc = 100000._rtype
 
 !=========================================================
 ! Physical constants used in SHOC
@@ -224,6 +228,7 @@ subroutine shoc_main ( &
      wthl_sfc, wqw_sfc, uw_sfc, vw_sfc, & ! Input
      wtracer_sfc,num_qtracers,w_field, &  ! Input
      inv_exner,phis, &                        ! Input
+     sens_heat,cflx, &                        ! Input
      host_dse, tke, thetal, qw, &         ! Input/Output
      u_wind, v_wind,qtracers,&            ! Input/Output
      wthv_sec,tkh,tk,&                    ! Input/Output
@@ -291,6 +296,9 @@ subroutine shoc_main ( &
   real(rtype), intent(in) :: inv_exner(shcol,nlev)
   ! Host model surface geopotential height
   real(rtype), intent(in) :: phis(shcol)
+
+  real(rtype), intent(in) :: sens_heat(shcol)
+  real(rtype), intent(in) :: cflx(shcol)
 
 ! INPUT/OUTPUT VARIABLES
   ! prognostic temp variable of host model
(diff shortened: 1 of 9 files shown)


tcclevenger:
Tested ne30pg2 CIME case with F2010 compset, 5 day simulation outputting all energy relative errors in SHOC over 1e-14 at each step. On master, energy errors ~1e-9, with this branch no errors.

oksanaguba:
For F runs, with use_mass_borrower i do not see error messages in the logs for the latest commit. I did not try to run without mass borrower.

components/eam/src/physics/cam/shoc.F90
```diff
@@ -31,6 +32,9 @@ module shoc
 real(rtype), parameter, public :: largeneg = -99999999.99_rtype
 real(rtype), parameter, public :: pi = 3.14159265358979323_rtype
 
+!repeated from shoc_intr!
+real(rtype), parameter         :: p0_shoc = 100000._rtype
```
tcclevenger, line 36: We should probably pass this through `shoc_init` as well to make sure everything matches. It is treated as an EAMxx constant so that way no sneaky BFB issues pop up. 

This requires some EAMxx/EAM interface changes, let me know if you want and I can add a quick commit.

tcclevenger (commented):
AT fails are from build issues with constants. If you don't want to worry about messing with the EAM/EAMxx interface stuff I don't mind adding a quick commit to the PR with all the changes. 

Otherwise looks good to me.

tcclevenger:
@oksanaguba Updated. Reran the CIME test and confirmed the energy errors are gone. 

There should be DIFFs when the AT runs and we probably need to find someone else to review the F90 changes.

oksanaguba:
@tcclevenger thanks for fixing constants interface, Exner bfb, and unit tests!

tcclevenger:
I cant remember with shoc_run_and_cmp (failing test) if it compares to previously built baselines or not. I'll make sure that fail isn't blocking this PR.

tcclevenger:
Yes the run_and_cmp fails are expected.

oksanaguba:
Climo results for v1 with this fix https://acme-climate.atlassian.net/wiki/spaces/EAMXX/pages/3922559112/Fixing+shoc+energy+fixer+PR+2527

AaronDonahue:
@oksanaguba , @tcclevenger do you guys think this is ready to merge?

oksanaguba:
@AaronDonahue i got distracted by other tasks. i will talk to Conrad about it today and we will decide. thanks.

tcclevenger:
@AaronDonahue I'm retesting since it was 2 weeks ago, but all fails were expected. I'll merge once the AT runs again and I check the output.

tcclevenger:
Retesting, mappy timed out

AaronDonahue:
@oksanaguba and @tcclevenger can you confirm that the DIFFs are what we expected?  It looks like we have DIFFs in `run_and_cmp` on `weaver` and DIFFs in the `mappy` CIME based tests...

If they are what we expect is this ready to merge?

tcclevenger:
@AaronDonahue yes, diffs in run_and_cmp around `1e-9` and in CIME ne4 tests around `1e-5 - 1e0`, which seem appropriate for the change (i.e., nothing blows up). I believe @oksanaguba has confirmed expected output in simulations, and I have confirmed energy errors disappear  in ne30 runs that previously contained them.

We can get confirmation from Oksana, but good to merge in my opinion.

oksanaguba:
Relevant climo statistics:
https://acme-climate.atlassian.net/wiki/spaces/EAMXX/pages/3922559112/Fixing+shoc+energy+fixer+PR+2527
@AaronDonahue  This can be merged, thanks.

AaronDonahue:
Great! Thanks @oksanaguba and @tcclevenger

PR #2528, opened 2023-09-13
EAMxx: switch to CXX17

bartgol:
C++17 seems to give no problem on mappy and weaver. It compiles fine on frontier (test pending). Now I'm going to run v1 tests on chrysalis and pm-gpu. I do not have access to ascent, but given that Summit will be decommissioned soon, I don't know if it's _that_ important.

Note: this upgrade is needed by other E3SM project. In particular, MALI: it links to "recent" Trilinos, which uses Kokkos 4.0, which in turn requires C++17. So even if we (as SCREAM) could punt on C++17 for a bit, it's a good idea to do it sooner than later, so that other projects can move forward.

### DIFF
--- a/components/eamxx/CMakeLists.txt
+++ b/components/eamxx/CMakeLists.txt
@@ -55,12 +55,8 @@ endif()
 # to be on. For now, simply ensure Kokkos Serial is enabled
 option (Kokkos_ENABLE_SERIAL "" ON)
 
-# MAM support requires C++17 -- hopefully SCREAM itself will get there soon
-if (SCREAM_ENABLE_MAM)
-  set(CMAKE_CXX_STANDARD 17)
-else()
-  set(CMAKE_CXX_STANDARD 14)
-endif()
+# We want to use C++17 in EAMxx
+set(CMAKE_CXX_STANDARD 17)
 
 if (NOT SCREAM_CIME_BUILD)
   project(SCREAM CXX C Fortran)


ambrad:
I recommend testing this change on Summit unless it's been decided that the Ascent CDash line and Summit support are no longer needed. I have two arguments for keeping EAMxx running on Summit: (1) GPU test machines are flaky, so having three (Ascent, Crusher, pm-gpu) helps keep us informed of GPU-relevant problems even when one or two test machines are having problems. (2) It's possible we'd want to do science runs on Summit if Frontier was ever giving us problems we couldn't work around.

bartgol:
I can def hop on Summit, where I do have access. I thought it was going to be decommissioned in Sept, but I may be mixing up different things/deadlines.

ambrad:
Mid Dec: https://docs.olcf.ornl.gov/systems/2023_olcf_system_changes.html

xyuan:
it also works on aurora using c++17 and sycl kokkos backend (required c++17 or above)

bartgol:
> it also works on aurora using c++17 and sycl kokkos backend (required c++17 or above)

Thanks for checking that.

bartgol:
On Frontier:

- the 3 DEBUG tests fail to build, due to `error: unhandled SGPR spill to memory` while building `eamxx_rrtmgp_process_interface.cpp`. We've never seen this before, but the issue is present in master as well.
- ERS_Ln9.ne4_ne4.F2000-SCREAMv1-AQP1 is a PASS
- ERP_Ln22.ne4pg2_ne4pg2.F2010-SCREAMv1 is a FAIL. Interestingly, the base-vs-rest check for the scream output passes, but we get a diff for the cpl output, for 6 out of 423 fields:
	- i2x_Fioi_meltw
	- i2x_Faii_evap
	- i2x_Faii_sen
	- i2x_Faii_lat
	- i2x_Si_snowh
	- x2a_Si_snowh
  I will try to run the same test for master, and see if we have a diff there too. It may take a while to get through the Q though.

bartgol:
As it often happens with ICEs, the Summit ICE went away by simply re-organizing the code (in this case, moving a certain code snippet inside its own function). Re-running on Summit.

bartgol:
> On Frontier:
> 
> * the 3 DEBUG tests fail to build, due to `error: unhandled SGPR spill to memory` while building `eamxx_rrtmgp_process_interface.cpp`. We've never seen this before, but the issue is present in master as well.
> * ERS_Ln9.ne4_ne4.F2000-SCREAMv1-AQP1 is a PASS
> * ERP_Ln22.ne4pg2_ne4pg2.F2010-SCREAMv1 is a FAIL. Interestingly, the base-vs-rest check for the scream output passes, but we get a diff for the cpl output, for 6 out of 423 fields:
>   
>   * i2x_Fioi_meltw
>   * i2x_Faii_evap
>   * i2x_Faii_sen
>   * i2x_Faii_lat
>   * i2x_Si_snowh
>   * x2a_Si_snowh
>     I will try to run the same test for master, and see if we have a diff there too. It may take a while to get through the Q though.

master exhibits the same behavior. :/

bartgol:
Update: recents mods got Summit to build, although I had to reduce GMAKE_J from 8 to 4, b/c I was hitting an nvcc error which I found out, with some online help, was due to too much stuff buiidling at the same time, making nvcc run out of memory (or something like that).

I am re-running e3sm_scream_v1_lowres on all other archs as well. Will report soon

```
| Machine  | Status |
|--- -----|
|Summit | PASS
|Frontier | PEND |
|Chrysalis | PEND |
|PM-GPU | PEND |
|PM-CPU  | PEND |
```

jgfouca (approved):

bartgol:
I'm going to wait for Frontier to come back online, so I can inspect the results of the tests I launched last night. If all is good, I will merge.

ambrad:
A build on Ascent works.

bartgol:
All machines are passing. Merging.

Issue #2471, opened 2023-08-23
Add ability to output precipitation areal fractions as a 3D diagnostic variable


susburrows:
Add the ability to output precipitation areal fractions as a 3D diagnostic variable.  I.e., the areal fraction of the grid box taken up by rain and by snow (separately).

susburrows:
@quantheory

PR #2530, opened 2023-09-14
Add effective radius of rain and rain precipitation fraction outputs

quantheory:
This commit adds the effective radius of rain drops (`eff_radius_qr`)
and rain precipitation fraction (`rainfrac`) outputs, as calculated by
P3. Both are also added to the list of default outputs.

These outputs are not communicated to COSP or any other
parameterization that might hypothetically use these fields.
It only makes them available in the field manager.

This solves the rain part of #2471.

BFB, except for the new outputs themselves.

### DIFF
--- a/components/eam/src/physics/p3/scream/micro_p3.F90
+++ b/components/eam/src/physics/p3/scream/micro_p3.F90
@@ -960,7 +960,8 @@ contains
       inv_exner, cld_frac_l, cld_frac_r, cld_frac_i, &
       rho, inv_rho, rhofaci, qv, th_atm, qc, nc, qr, nr, qi, ni, qm, bm, latent_heat_vapor, latent_heat_sublim, &
       mu_c, nu, lamc, mu_r, lamr, vap_liq_exchange, &
-      ze_rain, ze_ice, diag_vm_qi, diag_eff_radius_qi, diag_diam_qi, rho_qi, diag_equiv_reflectivity, diag_eff_radius_qc)
+      ze_rain, ze_ice, diag_vm_qi, diag_eff_radius_qi, diag_diam_qi, rho_qi, diag_equiv_reflectivity, &
+      diag_eff_radius_qc, diag_eff_radius_qr)
 
    implicit none
 
@@ -974,7 +975,8 @@ contains
         qv, th_atm, qc, nc, qr, nr, qi, ni, qm, bm, latent_heat_vapor, latent_heat_sublim, &
         mu_c, nu, lamc, mu_r, &
         lamr, vap_liq_exchange, &
-        ze_rain, ze_ice, diag_vm_qi, diag_eff_radius_qi, diag_diam_qi, rho_qi, diag_equiv_reflectivity, diag_eff_radius_qc
+        ze_rain, ze_ice, diag_vm_qi, diag_eff_radius_qi, diag_diam_qi, rho_qi, diag_equiv_reflectivity, &
+        diag_eff_radius_qc, diag_eff_radius_qr
 
    ! locals
    integer :: k, dumi, dumii, dumjj, dumzz
@@ -1028,6 +1030,7 @@ contains
          ze_rain(k) = nr(k)*(mu_r(k)+6._rtype)*(mu_r(k)+5._rtype)*(mu_r(k)+4._rtype)*           &
               (mu_r(k)+3._rtype)*(mu_r(k)+2._rtype)*(mu_r(k)+1._rtype)/bfb_pow(lamr(k), 6._rtype)
          ze_rain(k) = max(ze_rain(k),1.e-22_rtype)
+         diag_eff_radius_qr(k) = 1.5_rtype/lamr(k)
       else
          qv(k) = qv(k)+qr(k)
          th_atm(k) = th_atm(k)-inv_exner(k)*qr(k)*latent_heat_vapor(k)*inv_cp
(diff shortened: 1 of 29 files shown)


quantheory:
Not sure if this is too late for Cess, but in any case, here it is. I also wanted to add the rain fall speeds as an output, but it looked like it would take some work, and I'm out of time to work at this. I got to understand how P3 interacts with the AD much better, though.

@AaronDonahue, we should also probably talk about how best to deal with missing values in averaged outputs, at some point...

PeterCaldwell:
If we were to include these outputs in cess, what horizontal resolution, time frequency, ave or instant, and native grid or pressure levels do you want?

AaronDonahue (dismissed):
Looks good to me.  @PeterCaldwell , when you have a chance can you check the formula for rain radius and the default rain radius just to verify it is correct?

@brhillman , this PR doesn't pass any info to rrtmgp, but would rrtmgp or COSP want the rain radius?  This might be a naive question.

components/eam/src/physics/p3/scream/micro_p3.F90
```diff
@@ -1028,6 +1030,7 @@ subroutine p3_main_part3(kts, kte, kbot, ktop, kdir, &
          ze_rain(k) = nr(k)*(mu_r(k)+6._rtype)*(mu_r(k)+5._rtype)*(mu_r(k)+4._rtype)*           &
               (mu_r(k)+3._rtype)*(mu_r(k)+2._rtype)*(mu_r(k)+1._rtype)/bfb_pow(lamr(k), 6._rtype)
          ze_rain(k) = max(ze_rain(k),1.e-22_rtype)
+         diag_eff_radius_qr(k) = 1.5_rtype/lamr(k)
```
AaronDonahue, line 1033: @PeterCaldwell , here's the formula for eff_radius_qr

components/eam/src/physics/p3/scream/micro_p3.F90
```diff
@@ -1297,6 +1301,7 @@ SUBROUTINE p3_main(qc,nc,qr,nr,th_atm,qv,dt,qi,qm,ni,bm,   &
     ze_rain   = 1.e-22_rtype
     diag_eff_radius_qc = 10.e-6_rtype ! default value
     diag_eff_radius_qi = 25.e-6_rtype ! default value
+    diag_eff_radius_qr = 500.e-6_rtype ! default value
```
AaronDonahue, line 1304: @PeterCaldwell , and the default value

quantheory:
@PeterCaldwell So, to be frank, I did this more as practice working with SCREAM than because I'm thinking about the CESS runs myself, and because @susburrows was interested in these as site-specific outputs. So that would mean `eff_radius_qr` and `rainfrac` added to ARM outputs.

Personally, I think the most likely situation where I'd use this data is if I had high-horizontal-resolution snapshots (even if relatively infrequent ones) that had at least `rainfrac`, as well as `qr`, `nr`, and thermodynamic variables (`T`, `q`, `p`, or enough information to calculate those). I'd suggest `rainfrac` to the 3-hourly ne120 snapshots where I think you're already planning to write those out, since I don't expect you're outputting all of those at native resolution.

quantheory:
To be clear, definitely more interested in instantaneous than time averaged; I'm interested in variability in the state before averaging.

components/eam/src/physics/p3/scream/micro_p3.F90
quantheory: To be clear, I chose this fairly arbitrarily, as a value that's kind of agnostic between the category being drizzle or true rain. (I think we should consider using an alternate approach that will get rid of these defaults entirely, but that's a conversation for next week I think.)

susburrows:
@quantheory, thanks for working on this!

I had requested these as 3D fields every 15 minutes.  We'd want them as instantaneous outputs, but only in the domains surrounding the ARM sites.  @quantheory , is it possible/easy to also add the effective radius and precip fraction for snow?

quantheory:
@susburrows P3 only knows that ice crystals exist with certain properties, but it does not categorize them into separate categories like "cloud ice", "snow", "hail", etc. Since P3 has no definition of what "snow" is, there's no built-in way to measure a separate effective radius or precipitation fraction for snow.

However, if ice fields are being output (`qi`, `ni`, `qm`, and `bm`, `cld_frac_i`, and maybe `eff_radius_qi`), **and** you have some definition for what "snow" should be (e.g. ice crystals above a particular mass or radius), then it should be possible to perform an offline calculation to determine what fraction of the ice is snow. In particular, where the mean mass is large, you could assume that everything in the grid cell is "snow", and treat `eff_radius_qi` and `cld_frac_i` as the effective radius and area fraction for snow.

quantheory:
@AaronDonahue @mahf708
I forgot to mention, because I added new fields, one of which was also output from the Fortran version, the P3 baselines need to be regenerated before standalone tests will pass. I assume this is also why the `DIFF` failures occur above; I added these fields to the default output (I think that `rainfrac` should definitely be a default output at least).

AaronDonahue:
@jgfouca and/or @bartgol , can you confirm that if @quantheory changed the default output in this PR we would expect the unit tests to fail w/ the p3_run_and_cmp tests because the baselines have changed.  If so, then I think we can manually merge this PR.

bartgol:
Few things:
- It's been 2w since last test, so we need to retest
- There are conflicts to resolve with master
- The tests failed on weaver too, so it's not a matter of baselines.

AaronDonahue:
@quantheory ^^ See Luca's comment above.

quantheory:
Rebasing to master now, just re-running a couple of tests on my end, and then will push it.

@bartgol I think the failures on Weaver are still essentially "baseline" failures, just for the standalone P3 tests rather than the full model tests. That is, I had to redo the `make baseline` and `make baseline_cxx` when testing standalone to get things to pass.

This is expected, because I'm outputting an extra field from both the C++ and Fortran codes, and I added that field to the baseline comparisons. (In retrospect, this was probably not necessary, but I think it's basically consistent with what's already being done at least. And it made me slightly more confident that I knew how things worked when I got those tests to pass with the extra field.)

bartgol:
Ah, I see. Then yes, rebase-on/merge master and push. You don't even need to run tests yourself, the AT will take care of that. We do want the AT to re-run tests (even if we know the last failures were expected) since it's been 2w since that testing, and stuff has changed in master.

bartgol:
@quantheory is it still appropriate to consider this BFB?

quantheory:
@bartgol For every previously existing field, it should be BFB. The only difference in the output should be adding those new variables.

bartgol:
Just logging for posterity: the DIFFs in v1 tests are indeed b/c of new fields (I could have checked that rather than ask Sean)
```
SUMMARY of cprnc:
 A total number of     62 fields were compared
          of which      0 had non-zero differences
               and      0 had differences in fill patterns
               and      0 had different dimension sizes
 A total number of      0 fields could not be analyzed
 A total number of      2 time-varying fields on file 1 were not found on file 2.
 A total number of      0 time-constant fields on file 1 were not found on file 2.
 A total number of      0 time-varying fields on file 2 were not found on file 1.
 A total number of      0 time-constant fields on file 2 were not found on file 1.
  diff_test: the two files DIFFER only in their field lists
```
Same deal for all v1 DIFFs. 

Merging.

PR #2532, opened 2023-09-14
convert iop setinitial

jgfouca:
This was a very tedious conversion and I'm not too happy with the result. We were already struggling with the issue of BFB testing not being available. Now we are starting to convert functions that pull in quite a lot of dependencies from around eam, the main culprit being iop_data_mod.


components/eamxx/src/physics/dp/CMakeLists.txt
```diff
@@ -7,6 +9,15 @@ set(DP_SRCS
   #${SCREAM_BASE_DIR}/../eam/src/control/history_iop.F90"
 )
 
+# Set cmake config options for Homme
+if (NOT "${SCREAM_DYNAMICS_DYCORE}" STREQUAL "HOMME")
+  message(FATAL_ERROR "Requires homme")
```
jgfouca, line 14: In order to pick up Homme::Tracers and Homme::Elements, we establish a dependency on Homme. These Homme objects are being unused as the CXX equivalent of element_mod::element_t.

With this requirement, SCREAM_DYNAMICS_DYCORE must be on for dp to be able to be built.

components/eamxx/src/physics/dp/dp_functions.hpp
```diff
@@ -193,8 +198,45 @@ struct Functions
...
+    const uview_1d<const Spack>& vobs,      // actual v wind
+    const uview_1d<const Spack>& numliqobs, // actual ???
+    const uview_1d<const Spack>& numiceobs, // actual ???
+    const uview_1d<const Spack>& cldliqobs, // actual ???
+    const uview_1d<const Spack>& cldiceobs, // actual ???
+    const Scalar& psobs,                    // ???
+    const uview_1d<const Scalar>& dx_short, // short length scale in km (nelemd)
+    // Input/Output arguments
+    Scalar& dyn_dx_size,                    // for use in doubly periodic CRM mode
+    tracer_t& tracers,                      // tracers
```
jgfouca, line 236: We could potentially avoid using tracers and elements directly by taking the individual items used from those objects and just piling them in to the function signature.

It looks like we use:
elem.m_forcing.m_ft
elem.m_state.m_ps_v
elem.m_state.m_v
elem.m_derived.m_omega_p

And
tracers.Q

components/eamxx/src/physics/dp/dp_functions.hpp
```diff
@@ -193,8 +198,45 @@ struct Functions
...
+  //   when running SCM or DP-CRM.
+  //----------------------------------------------------------
+  static void iop_setinitial(
+    // Input arguments
+    const Int& plev,                        // number of vertical levels
+    const Int& pcnst,                       // number of advected constituents including cloud water
+    const Int& nelemd,                      // number of elements per MPI task
+    const Int& np,                          // NP
+    const Int& nstep,                       // the timestep number
+    const bool& use_replay,                 // use e3sm generated forcing
```
jgfouca, line 212: A lot of these params were coming from iop_data_mod, which has over 100 fields. I'm not sure what the best approach is here. We could make an iop_data_mod struct but that would require a lot more porting from f90. How would we initialize all that data?

components/eamxx/src/physics/dp/dp_functions_f90.hpp
```diff
@@ -84,15 +84,59 @@ struct AdvanceIopSubsidenceData : public PhysicsTestData {
...
+        {&uobs, &vobs, &numliqobs, &numiceobs, &cldliqobs, &cldiceobs, &tobs, &qobs}
+      }),
+    plev(plev_), pcnst(pcnst_), nelemd(nelemd_), np(np_), nstep(nstep_), psobs(psobs_),
+    use_replay(use_replay_), dynproc(dynproc_), have_t(have_t_), have_q(have_q_), have_ps(have_ps_), have_u(have_u_), have_v(have_v_), have_numliq(have_numliq_), have_cldliq(have_cldliq_), have_numice(have_numice_), have_cldice(have_cldice_), scm_zero_non_iop_tracers(scm_zero_non_iop_tracers_), is_first_restart_step(is_first_restart_step_) { }
+
+  PTD_STD_DEF(IopSetinitialData, 19, plev, pcnst, nelemd, np, nstep, psobs, use_replay, dynproc, have_t, have_q, have_ps, have_u, have_v, have_numliq, have_cldliq, have_numice, have_cldice, scm_zero_non_iop_tracers, is_first_restart_step);
+
+  void init()
+  {
+    tracers.init(nelemd, 10); // what is num tracers?
```
jgfouca, line 120: Guessed at few things here. I don't know where I'd get num tracers from. I also didn't know what to put for the Element scale factor.

components/eamxx/src/physics/dp/impl/dp_iop_setinitial_impl.hpp
```diff
@@ -12,11 +12,203 @@ namespace dp {
...
+    element_t& elem,
+    const uview_1d<Spack>& tobs,
+    const uview_1d<Spack>& qobs)
 {
-  // TODO
-  // Note, argument types may need tweaking. Generator is not always able to tell what needs to be packed
+  // Made these up
+
+  //FieldGroup g = ...;
+  //const auto& names = g.m_info->m_field_names;
```
jgfouca, line 52: Didn't know how to get indices for field names.

components/eamxx/src/physics/dp/impl/dp_iop_setinitial_impl.hpp
```diff
@@ -12,11 +12,203 @@ namespace dp {
...
+      policy,
+      KOKKOS_LAMBDA(const MemberType& team) {
+
+      // We cannot support parallelism at the element level when nstep<=1 because the thelev
+      // computation of tobs is dependent on prior iterations that may have alterted tobs
+      for (Int ie = 0; ie < nelemd; ++ie) {
+        for (Int j = 0; j < np; ++j) {
+          for (Int i = 0; i < np; ++i) {
+
+          // Find level where tobs is no longer zero
```
jgfouca, line 79: This part of the algorithm is inherently non-parallel because of races with tobs when nstep <=1.

components/eamxx/src/physics/dp/impl/dp_iop_setinitial_impl.hpp
```diff
@@ -12,11 +12,203 @@ namespace dp {
...
+            }, Kokkos::Min<Int>(thelev));
+          }
+
+          if (nstep <= 1) {
+            Kokkos::parallel_for(
+              Kokkos::TeamVectorRange(team, thelev+1), [&] (int k) {
+              auto zmask = k < thelev ? Smask(true) : tobs(k) == 0;
+              vector_simd for (Int p = 0; p < Spack::n; ++p) {
+                if (zmask[p]) {
+                  tobs(k)[p] = elem.m_forcing.m_ft(ie,i,j,k)[p];
```
jgfouca, line 97: This tobs write can change subsequent computations of `thelev`

components/eamxx/src/physics/dp/impl/dp_iop_setinitial_impl.hpp
```diff
@@ -12,11 +12,203 @@ namespace dp {
...
+                  qobs(k)[p] = tracers.Q(ie,0,i,j,k)[p]; // Tracer index 0 is qobs
+                }
+              }
+            });
+          }
+          else {
+            Kokkos::parallel_for(
+              Kokkos::TeamVectorRange(team, plev_packs), [&] (int k) {
+              // Ekat packs don't know how to assign themselves to KokkosKernels::Batched::Experimental::SIMD
```
jgfouca, line 106: Ekat packs don't know how to assign themselves to KokkosKernels::Batched::Experimental::SIMDs and vice versa, making the code even uglier.

components/eamxx/src/physics/dp/impl/dp_iop_setinitial_impl.hpp
```diff
@@ -12,11 +12,203 @@ namespace dp {
...
+                vector_simd for (Int p = 0; p < Spack::n; ++p) {
+                  if (zmask[p]) {
+                    tracers.Q(ie,0,i,j,k)[p] = qobs(k)[p];
+                  }
+                }
+              }
+            });
+
+            if (have_ps) {
+              elem.m_state.m_ps_v(ie, 0, i, j) = psobs; // what is [NUM_TIME_LEVELS]?
```
jgfouca, line 144: How to handle the NUM_TIME_LEVELS dimension? f90 didn't use it as far as I can tell.

components/eamxx/src/physics/dp/impl/dp_iop_setinitial_impl.hpp
```diff
@@ -12,11 +12,203 @@ namespace dp {
...
+            }
+
+            Kokkos::parallel_for(
+              Kokkos::TeamVectorRange(team, plev_packs), [&] (int k) {
+              if (have_u) {
+                vector_simd for (Int p = 0; p < Spack::n; ++p) {
+                  elem.m_state.m_v(ie, 0, 0, i, j, k)[p] = uobs(k)[p]; // [2] is 0?
```
jgfouca, line 151: Wasn't sure about the [2] dimension in 
 `ExecViewManaged<Scalar * [NUM_TIME_LEVELS][2][NP][NP][NUM_LEV]>;` for `m_v`.

components/eamxx/src/physics/dp/impl/dp_iop_setinitial_impl.hpp
```diff
@@ -12,11 +12,203 @@ namespace dp {
...
+  }
+
+  // If DP-CRM mode then SHOC/CLUBB needs to know about grid
+  // length size.  The calculations of this based on a sphere in the
+  // SHOC and CLUBB interefaces are not valid for a planar grid, thus
+  // save the grid length from the dycore. Note that planar dycore
+  // only supports uniform grids, thus we only save one value.
+  // Set this if it is the first time step or the first restart step
+  if ( (nstep == 0 || is_first_restart_step) && dynproc) {
+    // for (Int ie = 0; ie < nelemd; ++ie) {
```
jgfouca, line 199: The f90 code here made no sense to me. It was iterating over all of dx_short, assigning (not adding) to dyn_dx_size. If that was the intent, why not just assign it to the last item?

components/eamxx/src/physics/share/physics_test_data.hpp
```diff
@@ -70,7 +79,16 @@ struct SHOCGridData : public PhysicsTestData {
 #define  PTD_ASS8(a, b, c, d, e, f, g, h          )  PTD_ASS7(a, b, c, d, e, f, g)          ; h = rhs.h
 #define  PTD_ASS9(a, b, c, d, e, f, g, h, i       )  PTD_ASS8(a, b, c, d, e, f, g, h)       ; i = rhs.i
 #define PTD_ASS10(a, b, c, d, e, f, g, h, i, j    )  PTD_ASS9(a, b, c, d, e, f, g, h, i)    ; j = rhs.j
-#define PTD_ASS11(a, b, c, d, e, f, g, h, i, j, k ) PTD_ASS10(a, b, c, d, e, f, g, h, i, j) ; k = rhs.k
+#define PTD_ASS11(a, b, c, d, e, f, g, h, i, j, k                            ) PTD_ASS10(a, b, c, d, e, f, g, h, i, j)                            ; k = rhs.k
```
jgfouca, line 82: Having to support so many scalars is making these macros ridiculous.

jgfouca (commented):
Annotations.

components/eamxx/src/physics/dp/CMakeLists.txt
PeterCaldwell: >are being unused as the CXX equivalent
I think you mean they *are* used, right?

It seems reasonable for dp to need HOMME and dycore built since dp uses the dycore, right? 

Is the concern that we end up needing to build the dycore in order to do unit tests on unrelated stuff?

jgfouca: Yep, typo: `used` is what I meant. 

I'm not worried too much about the homme dependency but I am wondering if DP belongs under `physics`.

tcclevenger: It seems to me that DP is called from dynamics to alter dynamics elements. Why is DP considered a physics process? Should it just be a directory of callable function inside eamxx/src/dynamics/homme?

tcclevenger: Didn't see your response, Jim. I wonder the same thing. I think it is the case too with `apply_iop_forcing()` that it's called directly inside dynamics.

components/eamxx/src/physics/dp/impl/dp_iop_setinitial_impl.hpp
```diff
@@ -12,11 +12,203 @@ namespace dp {
...
+
+      // We cannot support parallelism at the element level when nstep<=1 because the thelev
+      // computation of tobs is dependent on prior iterations that may have alterted tobs
+      for (Int ie = 0; ie < nelemd; ++ie) {
+        for (Int j = 0; j < np; ++j) {
+          for (Int i = 0; i < np; ++i) {
+
+          // Find level where tobs is no longer zero
+          Int thelev=-1;
+          if (nstep <= 1) {
```
tcclevenger, line 81: All these `if (nstep <= 1)` and `if (nstep == 0)` seem unnecessary since part of the if-statement these are contained in is `if (... && nstep == 0 && ...)`. I may be missing something.

components/eamxx/src/physics/dp/CMakeLists.txt
PeterCaldwell: What about having a "dp_interface" layer which gathers all this stuff and organizes everything into a couple data structures, then having the actual functions which do stuff just ingest these data structures. Then initializing all these things for unit tests would just involve explicitly setting these things in the test code...

components/eamxx/src/physics/dp/dp_functions.hpp
PeterCaldwell: I like the idea of just passing the specific variables we use rather than big structures with a bunch of stuff we never use. I guess so long as we're just passing pointers there isn't any real memory impact, but it has annoyed me in the past that tons of unused variables are floating around in certain EAMf90 procs.

On the other hand, using elem directly gathers at least some of these variables into collections, minimizing your pain.

components/eamxx/src/physics/dp/CMakeLists.txt
PeterCaldwell: Ah, I agree that dp doesn't belong in physics. Putting it there is a throwback to the fact that we created dp from the single-column model code and the whole point of the single column model was that it only called physics routines. Perhaps DP should be in EAMxx/src/dp/ since it uses both physics and dynamics?

components/eamxx/src/physics/dp/dp_functions.hpp
PeterCaldwell: As I noted above, maybe having a dp_interface layer which collects all this stuff and bundles it into structs which we pass to these routines would be helpful. Then unit tests would just hardcode values for all these structs as needed, while the interface would collect it from elsewhere in the code (so no init needed).

components/eamxx/src/physics/dp/impl/dp_iop_setinitial_impl.hpp
PeterCaldwell: Agreed, the F90 makes no sense. dyn_dx_size should just be dx_short(nelemd-1)*1000

PeterCaldwell: Ah - from reading the comments, I think dyn_dx_size *was* multidimensional for global runs and got converted to a scalar for DP because (as the comments say) the planar dycore only supports uniform grids. The for loop probably made sense for global grids and should have been removed, but wasn't. I'm unclear why dx_short is still multi-dimensional in DP, but maybe it is inherently part of the dycore. If so, is dx_short(i) the same for all i? Otherwise how do we know that setting dyn_dx_size to the last instead of first entry is the way to go?

components/eamxx/src/physics/share/physics_test_data.hpp
PeterCaldwell: Yeah, this is absurd. Maybe clumping stuff into structs would solve? Or can we create a little helper function which returns a macro with a given number of dims?

components/eamxx/src/physics/dp/impl/dp_iop_setinitial_impl.hpp
bartgol: that's for the 2 components of the horiz velocity.

components/eamxx/src/physics/dp/impl/dp_iop_setinitial_impl.hpp
bartgol: On my todo list there is to make Homme use ekat::Pack (and other utils).

components/eamxx/src/physics/dp/impl/dp_iop_setinitial_impl.hpp
bartgol: F90 does have it too. It's called `timelevels`, declared in `src/theta-l_kokkos/element_state.F90`.

bartgol: The time levels are used to store copies of the state that are used during the timestepping. For the currently used DIRK method, a 5 stages method, turns out that 3 vectors are enough to store all stages (since you can recycle old stages as the RK method progresses).

components/eamxx/src/physics/dp/CMakeLists.txt
bartgol: What about `src/dynamics/dp`, kind of as a dycore choice? Or maybe directly in `src/dynamics/homme`, as `eamxx_homme_dp_process_interface.*pp`?

components/eamxx/src/physics/dp/impl/dp_iop_setinitial_impl.hpp
PeterCaldwell: I think DP probably just wants the most recent (up-to-date) value, right? Is that the last value, or is there a secret decoder ring needed to figure out the order of the time levels?

components/eamxx/src/physics/dp/CMakeLists.txt
PeterCaldwell: I don't want dp inside the dycore because it is as much part of physics as it is part of the dycore... it doesn't fit in either.

components/eamxx/src/physics/dp/impl/dp_iop_setinitial_impl.hpp
bartgol: Homme stores the index for the last value in the `TimeLevel` data structure. The index of the most up to date solution should be stored in the `np1` member of `TimeLevel`. However, you need to be careful _when_ you get that. If you get it _before_ `TimeLevel::update_dynamics_levels` has been called, then yes, `np1` is the most recent one. If `update_dynamics_levels` has already been called, then `n0` is the most recent. The update method simply does `np1=nm1`, `n0=np1`, `nm1 = n0`.

If you let `prim_run_subcycle` complete, then the most up to date time slice index is stored in `n0` (since update levels is called toward the end of prim_run_subcycle).

components/eamxx/src/physics/dp/CMakeLists.txt
bartgol: Ok. I thought dp was mostly a hack for dynamics (which is the only process that knows about cols interconnection), to make it see a planar periodic mesh, and then simply used whatever column physics you wanted.

Then yeah, `eamxx/src/dp` may be best. I may go a step further, and suggest `eamxx/src/doubly-periodic` ;)

components/eamxx/src/physics/dp/impl/dp_iop_setinitial_impl.hpp
tcclevenger: I *think* this is called directly before `prim_run_subcycle` in eam.

components/eamxx/src/physics/dp/CMakeLists.txt
PeterCaldwell: No, DP is way more than just a dynamics hack. It is a way of life (for ARM/ASR people ;-) ) . I like your version with doubly-periodic spelled out!

components/eamxx/src/physics/dp/impl/dp_iop_setinitial_impl.hpp
bartgol: Then `n0` shold be ok.

components/eamxx/src/physics/dp/dp_functions_f90.hpp
```diff
@@ -87,15 +87,59 @@ struct AdvanceIopSubsidenceData : public PhysicsTestData {
...
+    elem.init(nelemd, true, true, 2);
+  }
+
+  void randomize(std::mt19937_64& engine)
```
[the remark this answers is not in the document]
jgfouca: Should be off for production:

```
if (NOT SCREAM_LIB_ONLY)
  list(APPEND DP_SRCS
    dp_functions_f90.cpp
    ) # Add f90 bridges needed for testing                                                                                                                                                                                     
endif()
```

Same deal as p3/shoc.

components/eamxx/src/physics/dp/CMakeLists.txt
jgfouca: I am happy to move it. Is that OK @tcclevenger ? That would disrupt and dp change you have on branches.

components/eamxx/src/physics/dp/impl/dp_iop_setinitial_impl.hpp
jgfouca: Nice catch. That will reduce some of mess inside.

jgfouca:
@bartgol ,

> that's for the 2 components of the horiz velocity.

OK, I assume U is first, then V?

components/eamxx/src/physics/dp/impl/dp_iop_setinitial_impl.hpp
jgfouca: @bartgol , any thoughts here?

components/eamxx/src/physics/dp/CMakeLists.txt
tcclevenger: Ok for sure, I'll be able to copy and paste important stuff easily

components/eamxx/src/doubly-periodic/impl/dp_iop_setinitial_impl.hpp
```diff
@@ -0,0 +1,207 @@
...
+  const Int plev_packs = ekat::npack<Spack>(plev);
+  const auto policy = ekat::ExeSpaceUtils<ExeSpace>::get_default_team_policy(1, plev_packs);
+
+  assert(np <= NP);
+  //assert(plev_packs <= Homme::NUM_LEV);
+  assert(tracers.inited());
+  assert(elem.inited());
+
+  // Get time info
+  const Int n0 = 0; //Homme::Context::singleton().get<Homme::TimeLevel>().n0;
```
jgfouca, line 70: This attempt to use TimeLevel failed at runtime. I assume due to not being initialized:
`Error! Context member 'N5Homme9TimeLevelE' not found.`

components/eamxx/src/doubly-periodic/CMakeLists.txt
```diff
@@ -49,11 +49,10 @@ set_target_properties(dp PROPERTIES
   Fortran_MODULE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/modules
 )
 target_include_directories(dp PUBLIC
-  ${CMAKE_CURRENT_SOURCE_DIR}/../share
+  ${CMAKE_CURRENT_SOURCE_DIR}/../physics/share
```
bartgol, line 52: This line should not be needed, right? linking the `physics_share` lib should bring in all its include dirs...

components/eamxx/src/doubly-periodic/CMakeLists.txt
```diff
@@ -49,11 +49,10 @@ set_target_properties(dp PROPERTIES
...
   ${CMAKE_CURRENT_BINARY_DIR}/modules
   ${CMAKE_CURRENT_SOURCE_DIR}
   ${CMAKE_CURRENT_SOURCE_DIR}/impl
-  ${SCREAM_BASE_DIR}/../eam/src/physics/cam
 )
 target_link_libraries(dp physics_share scream_share ${dynLibName})
```
bartgol, line 57: Since we do it elsewhere, we may as well specify PUBLIC tgt dep here...

components/eamxx/src/doubly-periodic/impl/dp_iop_setinitial_impl.hpp
bartgol: yes, homme has a routine to init the time level structure. For the `theta-l_kokkos` target, that's `init_time_level_c`.

bartgol (dismissed):

jgfouca:
I think I addressed all blocking concerns except one, how do this on the CXX side:
```
    call cnst_get_ind('NUMLIQ', inumliq, abrtf=.false.)
    call cnst_get_ind('NUMICE', inumice, abrtf=.false.)
    call cnst_get_ind('CLDLIQ', icldliq)
    call cnst_get_ind('CLDICE', icldice)
```
So, looking up tracer ids by name which will later be used to index into `Q`. @bartgol mentioned that normally you'd do this on the CXX side by using a FieldGroup, but I don't know how to obtain this object from here without adding a bunch of stuff that probably only belongs in the process interface.

bartgol (approved):

PR #2533, opened 2023-09-14
Updating to SCORPIO 1.5.0

jayeshkrishna:
Updating SCORPIO from v1.4.1 to v1.5.0.

SCORPIO v1.5.0 includes,
* Refactored Fortran interface (The API remains the same but
  the code is refactored for better user error checking/reporting,
  code maintenance/correctness etc)
* Performance tuning/settings for Frontier/Crusher
* Performance enhancements for ADIOS
* Code refactoring and fixes for the HDF5 I/O type


jayeshkrishna:
@AaronDonahue / @sarats : For the CESS runs (if you are planning to use a branch for the runs) we will need to include this PR and PR https://github.com/E3SM-Project/scream/pull/2518 (which is a prereq for this PR)

bartgol (approved):

bartgol:
These failures are quite bizarre. Running the `time_interpolation` test manually on mappy, I get a fail in release mode, with the error

```
 FAIL:
  map_of_times_to_vector_idx.count(time)==0
  /home/lbertag/workdir/scream/scream-src/branch/components/eamxx/src/share/
  util/eamxx_time_interpolation.cpp:273
  Error! TimeInterpolation::set_file_data_triplets - The same time step has
  been encountered more than once in the data files, please check
      TimeStamp: 2023-05-16-00000
       Filename: source_data_for_time_interpolation.INSTANT.nsteps_x4.np1.2023-
  05-16-01200.nc
```

In the function `eam_update_time` from `scream_scorpio_interface.F90`, there is the following line

```fortran
    if (time>=0) ierr = pio_put_var(pio_atm_file%pioFileDesc,var%piovar,(/ pio_atm_file%numRecs /), (/ 1 /), (/ time /))
```
Changing it to this, makes the test pass
```fortran
    if (time>=0) then
      ierr = pio_put_var(pio_atm_file%pioFileDesc,var%piovar,(/ pio_atm_file%numRecs /), (/ 1 /), (/ time /)) 
      call errorHandle("PIO ERROR: could not update time on file "//trim(filename),ierr)
    endif
```
BUT, if I comment out the `call errorHandle` line, then the tests goes back to failing.

This smells a lot like a mem corruption error, but running the test through valgrind reveals nothing...

bartgol:
More intel. Even adding a simple `print *, ierr` instead of the `call errorHandle` line seems to make the code run. Printing other stuff (like filename) doesn't.

This makes me think: is it possible that _using_ `ierr` after the `pio_put_att` call forces the compiler to not throw away `ierr`, and that this, in turn, forces the compiler to _actually_ call `pio_put_att`? It is almost as if, without using ierr, the compiler sees no reason to actually call `pio_put_att`, since `ierr` is not needed.

@jayeshkrishna @dqwu does the `pio_put_att` function get declared as a no-side-effect function? That _may_ hint the compiler that, without using `ierr` afterwards, calling it is pointless...

jayeshkrishna:
This is indeed strange. pio_put_var() is not declared pure.

bartgol:
> This is indeed strange. pio_put_var() is not declared pure.

Exactly. But the behavior is _exactly_ as if it was. I will try a very small reproduced, only linking pio. If I can reproduce, maybe we can look at the binary strings, to see if the fcn call is there.

jayeshkrishna:
Ok, meanwhile I cannot recreate the issue on anl gce nodes (linux workstation, gcc 11.1.0). The test ERP_Ln22.ne4pg2_ne4pg2.F2010-SCREAMv1.anlgce_gnu passes (archive/screamv1-frontier-feb2023-2256-g66a067f024 + scorpio 1.5.0).

We just tagged a patch release, scorpio-v1.4.2, that only includes fixes that we need for the SCREAM runs (patch release over scorpio-v1.4.1). We are testing it right now, hopefully we have better luck with it.

bartgol:
Unfortunately, a small test that creates a file, adds an unlimited var, and tries to write, does not reproduce the error. I will investigate the main failure in a different way.

bartgol:
Ok, the call to `pio_put_var` is definitely not doing what it should. I wrapped it in this:
```
      ierr2 = pio_inq_dimid(pio_atm_file%pioFileDesc,"time",time_id)
      ierr2 = pio_inq_dimlen(pio_atm_file%pioFileDesc,time_id,tlen)
      print *, "tlen pre:", tlen
      print *, "calling pio_put_var with start:",pio_atm_file%numRecs
      ierr = pio_put_var(pio_atm_file%pioFileDesc,var%piovar,(/ pio_atm_file%numRecs /), (/ 1 /), (/ time /)) 
      ierr2 = pio_inq_dimlen(pio_atm_file%pioFileDesc,time_id,tlen)
      print *, "tlen post:", tlen
      ! print *, "ierr:", ierr
```
When the fcn is called for the first time, I see this 
```
 tlen pre:           0
 calling pio_put_var with start:           1
 tlen post:           0
```
but if I uncomment the last line, I see this
```
 tlen pre:           0
 calling pio_put_var with start:           1
 tlen post:           1
 ierr:           0
```

The only explanation I can find to this is that the compiler is completely removing the call to `pio_put_var` if the last line is commented, inferring (somehow) that it has no side effect, and is therefore useless.

jayeshkrishna:
Can you copy your reproducer to github gist and link it here?

bartgol:
> Can you copy your reproducer to github gist and link it here?

The small reproducer does not reproduce the problem though. Only the eamxx tests do. :/

rljacob:
I assume this can be closed since #2754  SCORPIO to 1.6.2

jayeshkrishna:
Closing this PR, we are going to update SCREAM directly to SCORPIO 1.6.* (see PR #2783)

PR #2540, opened 2023-09-15
Fix TMS race condition and remove packs form TMSFunctions

### LINKED ISSUE #2534: non-bfb issues with Cess-like cases on frontier
Starting with a sep6th checkout of machines/frontier, I have run many tests at ne30/ne256/ne1024 while debugging various issues as well as testing performance. I noticed that several cases were not BFB (looking at the hashes). Looking closer, I don't see any two ne256 or ne1024 cases that are BFB. They are all the same at 0th hash, but different at 18th.

With ne30, there is more consistency -- where I see many cases are the same, but at least 2 cases differ at hash 468. Additionally, even at ne30, using threads in ICE is not BFB and it should be (as we already compile with openmp since we use threading in LND). I just tried 2 ne30 cases without threads, and still see non-BFB.

On pm-gpu, using scream master of sep12, I have two ne256 cases that are identical after 5 days (only diff between the 2 cases is threading in ICE).

The cases on frontier are at `/lustre/orion/cli115/proj-shared/noel/e3sm_scratch/maf-sep6`

bartgol:
Jury is still out on whether this fixes #2534 , but I'm feeling confident it will. 

List of changes:

- add missing team barrier: there were two team-level pfor's with dependencies one after the other.
- remove use of Pack from TMSFunctions: the only fcn doesn't use packs at all, so no point in using them just to call `ekat::scalarize` inside (for every col, btw)
- change policy to a range policy: the team was not really used; in fact, all threads in the team were re-doing the same calculations, resulting in a (benign) race condition.

### DIFF
--- a/components/eamxx/src/physics/tms/eamxx_tms_process_interface.cpp
+++ b/components/eamxx/src/physics/tms/eamxx_tms_process_interface.cpp
@@ -115,8 +115,13 @@ void TurbulentMountainStress::run_impl (const double /* dt */)
   });
 
   // Compute TMS
-  TMSFunctions::compute_tms(ncols, nlevs, horiz_winds, T_mid, p_mid,
-                            exner, z_mid, sgh30, landfrac,
+  TMSFunctions::compute_tms(ncols, nlevs,
+                            ekat::scalarize(horiz_winds),
+                            ekat::scalarize(T_mid),
+                            ekat::scalarize(p_mid),
+                            ekat::scalarize(exner),
+                            ekat::scalarize(z_mid),
+                            sgh30, landfrac,
                             surf_drag_coeff_tms, wind_stress_tms);
 }
 
--- a/components/eamxx/src/physics/tms/eamxx_tms_process_interface.cpp
+++ b/components/eamxx/src/physics/tms/eamxx_tms_process_interface.cpp
@@ -108,6 +108,7 @@ void TurbulentMountainStress::run_impl (const double /* dt */)
     // Calculate z_mid
     PF::calculate_dz(team, pseudo_density_i, p_mid_i, T_mid_i, qv_i, dz_i);
     const Real z_surf = 0.0; // For now, set z_int(i,nlevs) = z_surf = 0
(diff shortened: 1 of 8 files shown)


ndkeen:
When I just add the team.barrier(), I get BFB results with ne30 after 10 days on frontier with TMS enabled.

components/eamxx/src/physics/tms/eamxx_tms_process_interface.cpp
```diff
@@ -108,14 +108,20 @@ void TurbulentMountainStress::run_impl (const double /* dt */)
     // Calculate z_mid
     PF::calculate_dz(team, pseudo_density_i, p_mid_i, T_mid_i, qv_i, dz_i);
     const Real z_surf = 0.0; // For now, set z_int(i,nlevs) = z_surf = 0
+    team.team_barrier();
```
tcclevenger, line 111: Ah, good catch. I remember have to search for a bug in the shoc interface because no team barrier here..

components/eamxx/src/physics/tms/impl/compute_tms_impl.hpp
```diff
@@ -31,18 +33,15 @@ void Functions<S,D>::compute_tms(
   const Scalar rair    = C::Rair;    // Gas constant for dry air
 
   // Loop over columns
-  const auto nlev_packs = ekat::npack<Spack>(nlevs);
-  const auto policy = ekat::ExeSpaceUtils<typename KT::ExeSpace>::get_default_team_policy(ncols, nlev_packs);
-  Kokkos::parallel_for(policy, KOKKOS_LAMBDA(const MemberType& team) {
-    const int i = team.league_rank();
-
+  const typename KT::RangePolicy policy (0,ncols);
+  Kokkos::parallel_for(policy, KOKKOS_LAMBDA(const int& i) {
```
tcclevenger, line 37: Yep, no need for team policy.

tcclevenger (approved):
Thanks for finding this, Luca!

PR #2546, opened 2023-09-20
Allow locally-empty layouts in AtmosphereInput

bartgol:
This was already done for writing, but the lack of fix in the read part caused restart crashes.

Note: this bug only shows up when one requests output coarsened to the point where there are less columns than ranks in the coarsened output.


AaronDonahue (approved):

ndkeen:
I see this already merged. I was finally able to test it with my 36-step ne1024 case on frontier. I can run with 512 nodes and run a short 36-step case, write restart, then read in for another 36 steps. This was not possible before adding this change. This now allows us to test write/read of restarts a little cheaper -- 2 jobs using 512 nodes running in under 30min each.

PR #2547, opened 2023-09-21
add calls to enable PBL diagnostic output (v0)

bogensch:
Add addfld and outfld calls in shoc_intr.F90 to enable diagnostic output of the planetary boundary layer (PBL) depth.  Change is made in v0 only as it is needed for DP simulations for THREAD project.


crterai (approved):
Looks good to me.

PR #2549, opened 2023-09-21
Crusher: Point to cprnc rather than build it.

ambrad:
The location of the pre-built cprnc has changed. Point to the new location. This might solve the mysterious problem we see in the nightlies where the test-built cprnc core dumps in the second comparison. This is a guess because I haven't been able to reproduce the problem in manual runs of the tests.


ambrad:
Tested on Crusher.

PR #2550, opened 2023-09-26
Ensure IO decomp tags do not clash

bartgol:
In current master, it was possible for two IO decomposition to have the same name on some ranks but not all. That is because we built the decomp ids with rank-local information, without considering that two decomps may have the same extents on _some_ ranks, but not all.

For instance, one decomp on CoarseningRemapper had tag
```
const std::string idx_decomp_tag = "coarsening_remapper::constructor_int_nnz" + std::to_string(nlweights);
```
where `nlweights` was the _local_ number of non-zeros to read from the map file. However, two different map files can cause some ranks to get the same `nlweights` for both, while other ranks will have two different extents. As a consequence, the ranks that had a match would not participate to the creation of the new decomp, and simply recycle the existing one. However, ranks where `nlweights` is different between the two map files would create a new decomp. Since creating the decomp is a collective operation, this would cause a hang.

This PR introduces some unique id in the tag name. In particular, if the decomp is built from a grid object, we add a unique grid id to the tag name. If it's built from a map file (without a grid object), we associate a unique id to each map filename.

@ndkeen Once this PR passes all tests, it would be nice to test it for the cases that were hanging before.

### DIFF
--- a/components/eamxx/src/share/grid/abstract_grid.cpp
+++ b/components/eamxx/src/share/grid/abstract_grid.cpp
@@ -28,6 +28,11 @@ AbstractGrid (const std::string& name,
 
   // This grid name is also an alias
   m_aliases.push_back(m_name);
+
+  // Ensure each grid object gets a different id
+  static int counter = 0;
+  m_unique_grid_id = counter;
+  ++counter;
 }
 
 AbstractGrid::
--- a/components/eamxx/src/share/grid/abstract_grid.hpp
+++ b/components/eamxx/src/share/grid/abstract_grid.hpp
@@ -161,6 +161,8 @@ public:
   // NOTE: we'd need setter/getter for this, so we might as well make it public
   std::string m_short_name = "";
 
+  int get_unique_grid_id () const { return m_unique_grid_id; }
+
 protected:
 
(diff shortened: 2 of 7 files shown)


AaronDonahue (approved):

ndkeen:
Ran a quick ne30 test with your branch on pm-gpu.

```
12:  FAIL:
12: err!=(-46)
12: /global/cfs/cdirs/e3sm/ndk/repos/bartgol_make-scorpio-decomps-unique/components/eamxx/src/share/io/scream_scorpio_interface.cpp:121
12: Error! Could not find dimension in the file.
12:  - filename : /global/cfs/cdirs/e3sm/inputdata/atm/scream/maps/map_ne1024pg2_to_ne30pg2_20230630_vert_enhanced.nc
12:  - dimname  : lev
12:  - pio error: -46
```

ndkeen:
After Chris made new vertical remap netcdf files, I tried this branch again on pm-gpu and I see that ne30 works -- running 1 month, writing restarts, reading those restarts and running another month.  And this morning, I see that can also do the same for a ne256 case (where I only ran 5 days, restart, another 5 days)

PR #2553, opened 2023-09-27
Add special "io: string attributes" field extra data, which IO will write to file

bartgol:
- Atm procs, diagnostics, driver, or other parts of the code can add string attributes to field extra data. IO will take all of them and add them to the nc file. E.g., usage by an atm proc:
```c++
using stratts_t = std::map<std::string,std::string>;
auto& f = get_field_out("blah");
auto& atts = f.get_header().get_extra_data<stratts_t>("io: string attributes");
atts["my_att_name"] = "some string";
```
- This extra data is preserved during output remap.
- The FieldAtXYZ diagnostics also preserve this extra data.
- Tested in `io_basic` and `io_remap_tests`. I also tested it by temporarily hacking p3, adding a dummy attribute to "T_mid", and observing it in both "T_mid" and  "T_mid_at_2m" in the output file.

@brhillman Let me know if this suits your needs or if it needs adjustment.

### DIFF
--- a/components/eamxx/src/physics/p3/eamxx_p3_process_interface.cpp
+++ b/components/eamxx/src/physics/p3/eamxx_p3_process_interface.cpp
@@ -340,11 +340,6 @@ void P3Microphysics::initialize_impl (const RunType /* run_type */)
   // Setup WSM for internal local variables
   const auto policy = ekat::ExeSpaceUtils<KT::ExeSpace>::get_default_team_policy(m_num_cols, nk_pack);
   workspace_mgr.setup(m_buffer.wsm_data, nk_pack_p1, 52, policy);
-
-
-  using stratts_t = std::map<std::string,std::string>;
-  auto& io_atts = get_field_out("T_mid").get_header().get_extra_data<stratts_t>("io: string attributes");
-  io_atts["test"] = "blah";
 }
 
 // =========================================================================================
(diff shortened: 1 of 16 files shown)


components/eamxx/src/physics/p3/eamxx_p3_process_interface.cpp
```diff
@@ -340,6 +340,11 @@ void P3Microphysics::initialize_impl (const RunType /* run_type */)
   // Setup WSM for internal local variables
   const auto policy = ekat::ExeSpaceUtils<KT::ExeSpace>::get_default_team_policy(m_num_cols, nk_pack);
   workspace_mgr.setup(m_buffer.wsm_data, nk_pack_p1, 52, policy);
+
+
+  using stratts_t = std::map<std::string,std::string>;
+  auto& io_atts = get_field_out("T_mid").get_header().get_extra_data<stratts_t>("io: string attributes");
+  io_atts["test"] = "blah";
```
AaronDonahue, line 347: Did you mean to delete this after testing?

components/eamxx/src/share/io/scorpio_output.cpp
```diff
@@ -202,6 +202,18 @@ AtmosphereOutput (const ekat::Comm& comm, const ekat::ParameterList& params,
   // Register any diagnostics needed by this output stream
   set_diagnostics();
 
+  // Helper lambda, to copy io string attributes. This will be used if any
+  // remapper is created, to ensure atts set by atm_procs are not lost
+  const std::string io_string_atts_key ="io: string attributes";
```
AaronDonahue, line 207: any reason this is not defined inside the lambda?  I don't think you use it anywhere else in the constructor.

AaronDonahue (changes requested):
Just a couple minor comments.  Otherwise this looks great.

components/eamxx/src/physics/p3/eamxx_p3_process_interface.cpp
bartgol: whoops!

PR #2554, opened 2023-09-27
Add runtime options to control shoc - isotropic_ts

AaronDonahue:
This commit makes it possible to pass the parameters used in the SHOC routine isotropic_ts as runtime options.  This is the first step in exposing a number of shoc tuning parameters to be runtime options.


AaronDonahue:
@crterai - here's the tuning PR for SHOC.  Still need to work on the P3 one.

AaronDonahue:
Note: This PR assumes that we will eventually drop support for F90 in EAMxx.  So I did not go through all of the steps of adding the new tuning params as args for the F90 <-> C++ bridges.  This PR is still BFB w/ F90 because the default parameters match F90.

If anyone has an objection to this step I can add the extra lines of code.  @tcclevenger , @jgfouca , @ambrad , @bartgol , @PeterCaldwell pinging you since you may have an opinion on this approach.

components/eamxx/src/physics/shoc/eamxx_shoc_process_interface.cpp
```diff
@@ -230,6 +230,11 @@ void SHOCMacrophysics::init_buffers(const ATMBufferManager &buffer_manager)
 // =========================================================================================
 void SHOCMacrophysics::initialize_impl (const RunType run_type)
 {
+  // Gather runtime options
+  runtime_options.lambda_low    = m_params.get<Real>("lambda_low",0.001);
```
bartgol, line 234: I would not provide a default here. Instead, I would put a default value in the XML file. Users won't know about this default just looking at the docs, but they will see any default value that is present in the XML file.

components/eamxx/src/physics/shoc/disp/shoc_tke_disp.cpp
```diff
@@ -39,7 +43,12 @@ ::shoc_tke_disp(
 
     auto workspace       = workspace_mgr.get_workspace(team);
 
+    const Scalar lambda_low    = 0.001;  //ASD
```
bartgol, line 46: Why these? Shouldn't you just use the values passed to the function?

components/eamxx/src/physics/shoc/impl/shoc_main_impl.hpp
```diff
@@ -434,7 +446,13 @@ void Functions<S,D>::shoc_main_internal(
                      brunt,shoc_mix);       // Output
 
     // Advance the SGS TKE equation
-    shoc_tke_disp(shcol,nlev,nlevi,dtime,wthv_sec,    // Input
+    const Scalar lambda_low    = 0.001;  //ASD
```
bartgol, line 449: Again, why these? You can just use whatever is passed to the function, no?

components/eamxx/src/physics/shoc/shoc_functions.hpp
```diff
@@ -69,6 +69,16 @@ struct Functions
   using WorkspaceMgr = typename ekat::WorkspaceManager<Spack,  Device>;
   using Workspace    = typename WorkspaceMgr::Workspace;
 
+  // This struct stores runtime options for shoc_main
+ struct SHOCRuntime {
+   SHOCRuntime() = default;
+   // Runtime options for isotropic_ts
+   Scalar lambda_low    = 0.001;
```
bartgol, line 76: Let's avoid putting defaults in several places in the source code, especially for configurable parameters. Let's just put the defaults in the XML file.

bartgol (changes requested):

components/eamxx/src/physics/shoc/shoc_functions.hpp
AaronDonahue: I've made the change in the latest version.  I had added the defaults here for the sake of unit tests, so that users wouldn't have to know the order of the parameters when initializing a SHOCRuntime struct.  But presumably we won't be adding new runtime options often so probably not necessary.

components/eamxx/src/physics/shoc/impl/shoc_isotropic_ts_impl.hpp
```diff
@@ -28,11 +32,11 @@ ::isotropic_ts(
...
   //Declare constants
-  static constexpr Scalar lambda_low   = 0.001;
-  static constexpr Scalar lambda_high  = 0.04;
-  static constexpr Scalar lambda_slope = 2.65;
-  static constexpr Scalar lambda_thresh= 0.02;
-  static constexpr Scalar maxiso       = 20000; // Return to isotropic timescale [s]
+         const     Scalar lambda_low    = lambda_low_in; 
+         const     Scalar lambda_high   = lambda_high_in;
+         const     Scalar lambda_slope  = lambda_slope_in;
+         const     Scalar lambda_thresh = lambda_thresh_in;
```
tcclevenger, line 38: If you are making other changes, then add a fix to indent here.

components/eamxx/src/physics/shoc/shoc_functions_f90.cpp
```diff
@@ -2856,6 +2861,7 @@ Int shoc_main_f(Int shcol, Int nlev, Int nlevi, Real dtime, Int nadv, Int npbl,
                                              qwthl_sec_d, wthl_sec_d, wqw_sec_d, wtke_sec_d,
                                              uw_sec_d,    vw_sec_d,   w3_d,      wqls_sec_d,
                                              brunt_d,     isotropy_d};
+  SHF::SHOCRuntime shoc_runtime_options;
```
tcclevenger, line 2864: From comments above, I think it would be better here to set the default value since this function is only used for BFB tests with F90, so it shouldn't need variable parameters.

tcclevenger (dismissed):
Looks good to me. Don't worry about the indent if Luca also has no other changes.

bartgol (approved):

AaronDonahue:
I had to commit a small fix to an input.yaml file, so now I need a re-review.  @bartgol or @tcclevenger

tcclevenger (dismissed):

bartgol (approved):

crterai:
@AaronDonahue - Thanks for creating this PR! we'll want to set `lambda_high` to 40.0 using this runtime option for our Cess simulations. Does this parameter show up in the namelist_scream.xml file and can we change it using `./atmchange ...` ?

PR #2555, opened 2023-09-27
Set default macmic=6 for ne256 and macmic=2 for ne512

ndkeen:
Changing default values of macmic for ne256 and ne512.
We encountered crashes at ne256 with macmic=3, and using 6 seems to resolve.


crterai (approved):
Looks good to me. Thanks, @ndkeen!

crterai:
Changes only affect ne256 and ne512 simulations and should be BFB with all our current tests.

ndkeen:
We also discussed making macmic=4 for ne256 (and initial results seem to suggest it's OK).

Issue #2556, opened 2023-09-27
Lower case `u` and `v` should not be allowed in output yaml


elynnwu:
We found discrepancies between `horiz_winds` and lowercase `u` and `v` when no remapping is performed. This is not an issue if we requested uppercase `U` and `V`, or if we remap both horizontally and vertically.

Here are the 4 experiments done:
1. Remapped horizontally and vertically. Output `horiz_winds` , `u` , and `v` . No difference between `horiz_winds[dim=0]` and `u`
2. Remapped horizontally and vertically. Output `horiz_winds`, `U`,  and `V`. No difference between `horiz_winds[dim=0]` and `U`
3. No remapping, output native grid. Output `horiz_winds`, `u` , and `v` . Difference between `horiz_winds[dim=0]` and `u` observed:
```
np.testing.assert_allclose(f2.horiz_winds.isel(dim2=1,time=0), f2.u.isel(time=0))
...
AssertionError: 
Not equal to tolerance rtol=1e-07, atol=0

Mismatched elements: 2764800 / 2764800 (100%)
Max absolute difference: 95.25008
Max relative difference: 8619548.
 x: array([[  1.160124,  -0.207151,  -0.787418, ..., -11.84242 , -11.617801,
        -11.110476],
       [  3.189489,   2.263207,   1.753559, ..., -12.107219, -11.874316,...
 y: array([[-29.751326, -27.446182, -25.305601, ...,  -1.167284,  -1.154717,
         -1.114421],
       [-28.3936  , -25.810406, -23.467003, ...,   0.335149,   0.314825,.. 
```
4. No remapping, output native grid. Output `horiz_winds`, `U` , and `V`. No difference between `horiz_winds[dim=0]` and `U`

It is unclear where `u` and `v` came from, atmos driver only declares `U` and `V` from `horiz_winds` ([here](https://github.com/E3SM-Project/scream/blob/master/components/eamxx/src/control/atmosphere_driver.cpp#L793)).

Proposed solution: always request `U` and `V`, do not allow outputting `u` and `v`

bartgol:
@elynnwu did you run this with _and_ without the nudging process?

elynnwu:
This is all without nudging.

elynnwu:
This issue is fixed in PR #2574

PR #2557, opened 2023-09-28
Update nudging to use correct wind fields `U` and `V`

elynnwu:
Per #2556, we conclude that `U` and `V` are the appropriate wind fields to use.

### DIFF
--- a/components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
+++ b/components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
@@ -52,7 +52,7 @@ void Nudging::set_grids(const std::shared_ptr<const GridsManager> grids_manager)
    * to be used here which we can then setup using the m_fields_nudge variable
    * For now we check if a field is intended to be nudged via the m_fields_nudge
    * vector, if it is we register it.  For now we are limited to just T_mid, qv,
-   * u and v
+   * U and V
    */
   if (std::find(m_fields_nudge.begin(),m_fields_nudge.end(),"T_mid") != m_fields_nudge.end()) {
     add_field<Updated>("T_mid", scalar3d_layout_mid, K, grid_name, ps);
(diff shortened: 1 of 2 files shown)


bartgol:
I am reluctant to simply brush the issue under the rug, changing u to U and v to V. There clearly is something that makes u/v valid names, even though they are never used. More importantly, in #2556 you showed that while u/v are bfb with horiz_winds, and so are U/V, mixing cases causes diffs. That's not clear why, so I am not sure U/V is the solution, simply put, because I don't understand what the _real_ problem is.

That said, feel free to merge this, as it is probably ok to switch u/v to U/V. But we are not solving the problem, we are hiding it.

elynnwu:
> I am reluctant to simply brush the issue under the rug, changing u to U and v to V. There clearly is something that makes u/v valid names, even though they are never used. More importantly, in #2556 you showed that while u/v are bfb with horiz_winds, and so are U/V, mixing cases causes diffs. That's not clear why, so I am not sure U/V is the solution, simply put, because I don't understand what the _real_ problem is.
> 
> That said, feel free to merge this, as it is probably ok to switch u/v to U/V. But we are not solving the problem, we are hiding it.

I agree. I think it can be a separate PR that actually resolves the issue, we'll keep the issue open.

AaronDonahue:
@bartgol and @elynnwu , After some digging I found a simple explanation for why `u` and `U` both work.  Fields are stored in the field manager w/ a ekat::case_insensitive name for the `map` of fields.  Thus, when a user requests a `get_field` it doesn't matter what the case of the string is, you will get the correct pointer to the field.

Of course, then the field will be saved in the netCDF file with the wrong case, so this can lead to confusion if some users use `U` while others use `u`.

@bartgol can speak to whether or not there are places for other bugs, like when registering a field, as this PR attempts to address. Maybe we should drop the case-insensitive declaration for the field manager map of fields (?)

Note, you can verify this is true by asking for T_MID for example.  That will work too.

bartgol:
I think I originally used CI strings because I thought we wanted to allow, e.g., both `Temperature` and `temperature`. It would be annoying to see a run crash because of a different case. That said, if we really wanted our fields to be case insensitive, we should propagate the usage of CI strings in a lot of places. In the long run, some regular std::string usage will creep up, and may cause this kind of confusion again (or, worse, some odd crash due to some string not being found).

Perhaps @PeterCaldwell may want to chime in, but personally I think it's best to remove CI strings usage from the code. I don't feel that strongly about it though, so I could go either way.

bartgol:
> I imagine a future where folks get confused about the proper capitalization of variable names or - worse yet - try to add variables of identical spelling but different capitalization and are surprised when chaos ensues. Although maybe they would deserve it - that would be dumb.

Yes, this is a legit concern. I can foresee someone testing some sort of algorithmic mod to variable "blah", and saving its new version as "BLAH", thinking they will be two different outputs. I agree that relying on name casing is a bad habit, and as you say, such users would somewhat ask for it. Still, let's gain karma points and save them some stress.

PeterCaldwell:
I definitely prefer U and V to u and v because the former is actually the correct name for the variables and folks might see the latter in our simulations and figure that's the right name for future simulations... which will cause failures because we're going to get rid of case insensitive behavior. 

That said, I think the Cess runs have already started and I'm not sure they have been updated to use U and V... So maybe it's too late.

bartgol:
It may be too late for the Cess runs, but not for master. Let's stick with uppercase, as Peter said. I'll take care of removing CI strings and make everything consistent.

elynnwu:
It sounds like we're in agreement to use uppercase `U` and `V`. Can either one of you approve this PR @bartgol @AaronDonahue?

AaronDonahue (approved):

PR #2558, opened 2023-09-28
Add max_total_ni as a runtime option in P3

AaronDonahue:
Similar to #2554 , this commit adds a new runtime option to a physics process, in this case P3.  The new option is for `max_total_ni`.  This commit establishes a structure in the P3 interface with the AD that can handle additional runtime options as they become needed.

Note: Similar to the work of #2554 this PR assumes that we will eventually drop support for the F90 version of SCREAM so there are no changes to the P3 F90 <-> C++ bridges.  It is still, however, BFB as long as the user sets max_total_ni to the default value.

### DIFF
--- a/components/eamxx/cime_config/namelist_defaults_scream.xml
+++ b/components/eamxx/cime_config/namelist_defaults_scream.xml
@@ -196,6 +196,7 @@ be lost if SCREAM_HACK_XML is not enabled.
       <do_predict_nc>true</do_predict_nc>
       <do_predict_nc COMPSET=".*SCREAM.*noAero">false</do_predict_nc>
       <enable_column_conservation_checks>false</enable_column_conservation_checks>
+      <max_total_ni type="real" doc="maximum total ice concentration (sum of all categories) (m)">740.0e3</max_total_ni>
       <tables type="array(file)">
         ${DIN_LOC_ROOT}/atm/scream/tables/p3_lookup_table_1.dat-v4.1.1,
         ${DIN_LOC_ROOT}/atm/scream/tables/mu_r_table_vals.dat8,
--- a/components/eamxx/cime_config/namelist_defaults_scream.xml
+++ b/components/eamxx/cime_config/namelist_defaults_scream.xml
@@ -196,7 +196,7 @@ be lost if SCREAM_HACK_XML is not enabled.
       <do_predict_nc>true</do_predict_nc>
       <do_predict_nc COMPSET=".*SCREAM.*noAero">false</do_predict_nc>
       <enable_column_conservation_checks>false</enable_column_conservation_checks>
-      <max_total_ni type="real" doc="maximum total ice concentration (sum of all categories) (m)">740.0e3</max_total_ni>
+      <max_total_ni type="real" doc="maximum total ice concentration (sum of all categories)" constraints="gt 0">740.0e3</max_total_ni>
       <tables type="array(file)">
         ${DIN_LOC_ROOT}/atm/scream/tables/p3_lookup_table_1.dat-v4.1.1,
         ${DIN_LOC_ROOT}/atm/scream/tables/mu_r_table_vals.dat8,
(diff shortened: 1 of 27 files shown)


AaronDonahue:
@crterai Here's the change for `max_total_ni`.

AaronDonahue:
There is a chance that this PR may conflict w/ 2554 for a few of the input.yaml file changes.  So we should be sure to not merge them back-to-back.

components/eamxx/cime_config/namelist_defaults_scream.xml
```diff
@@ -196,6 +196,7 @@ be lost if SCREAM_HACK_XML is not enabled.
       <do_predict_nc>true</do_predict_nc>
       <do_predict_nc COMPSET=".*SCREAM.*noAero">false</do_predict_nc>
       <enable_column_conservation_checks>false</enable_column_conservation_checks>
+      <max_total_ni type="real" doc="maximum total ice concentration (sum of all categories) (m)">740.0e3</max_total_ni>
```
bartgol, line 199: What's the trailing `(m)`?

components/eamxx/src/physics/p3/p3_functions.hpp
```diff
@@ -107,6 +107,13 @@ struct Functions
   using WorkspaceManager = typename ekat::WorkspaceManager<Spack, Device>;
   using Workspace        = typename WorkspaceManager::Workspace;
 
+  // Structure to store p3 runtime options
+  struct P3Runtime {
+    P3Runtime() = default;
```
bartgol, line 112: FYI, this kind of defaulted constructor for POD types is a bit of an overkill. In general, if a struct has only built in types, or very well known types (like std::string) that are known to be default constructible, you can omit the constructor.

components/eamxx/src/physics/p3/p3_functions_f90.cpp
```diff
@@ -1685,6 +1685,7 @@ void p3_main_part2_f(
 
   const Int nk = (kte - kts) + 1;
   const Int nk_pack = ekat::npack<Spack>(nk);
+  const Real max_total_ni = 740.0e3;
```
bartgol, line 1688: Might be worth adding a comment, that this is the value hard-coded in Fortran. The user may be confused as of why this is hard coded, while eamxx has it as a runtime parameter...

components/eamxx/src/physics/p3/p3_functions_f90.cpp
```diff
@@ -1840,6 +1841,7 @@ void p3_main_part3_f(
...
```
bartgol, line 1844: Same dealio

components/eamxx/cime_config/namelist_defaults_scream.xml
bartgol: Are there constraints on this param? I'm guessing it should be non negative? If so, you could add `constraints="ge 0"` (or gt if it should be strictly positive). This would prevent an accidental typo like `atmchange max_total_ni=-800e3` from going through (the `=` and `-` keys are close together, so it's not unlikely to accidentally type =- instead of =).

components/eamxx/tests/coupled/physics_only/shoc_p3_nudging/input_nudging.yaml
```diff
@@ -11,6 +11,8 @@ time_stepping:
 atmosphere_processes:
   schedule_type: Sequential
   atm_procs_list: [shoc,p3,nudging]
+  p3:
+    max_total_ni: 740.0e3
```
bartgol, line 15: Any reason not to try another value in all the unit tests? It's quite unlikely that we'd trigger a bug, but we may as well. Something that may trigger an edge case may be interesting to test though... Like what happens if we use 1e0 or something small? Will certain special path of p3 get triggered more/less?

AaronDonahue: @crterai or @hassanbeydoun , any thoughts on a good value to use here other than 740?  Could we try something super small like 1.0 as Luca suggests?

bartgol (approved):

AaronDonahue:
I had to rebase to resolve merge conflicts.  So this will need a re-review.

AaronDonahue:
@bartgol , it looks like this is passing now.  So I think we can go ahead and go the usual route and let the AT automerge.  Would you mind re-approving?

bartgol (approved):

PR #2561, opened 2023-09-29
Default TMS on in CIME runs

tcclevenger:
Also adds an error message if trying to use TMS with Physics GLL.

@crterai Any interest in this going in the cess-production branch? Shouldn't change anything about the runs, could remove some lines in scripts I suspect.

### DIFF
--- a/components/eamxx/cime_config/namelist_defaults_scream.xml
+++ b/components/eamxx/cime_config/namelist_defaults_scream.xml
@@ -141,7 +141,7 @@ be lost if SCREAM_HACK_XML is not enabled.
       so it must be of the form (a,b,...).
       NOTE: *CANNOT* be changed.
     -->
-    <!-- <atm_procs_list>(sc_import,homme,((shoc,cldFraction,spa,p3),rrtmgp,sc_export))</atm_procs_list> -->
+    <!-- <atm_procs_list>(sc_import,homme,((tms,shoc,cldFraction,spa,p3),rrtmgp,sc_export))</atm_procs_list> -->
     <atm_procs_list type="array(strring)">sc_import,homme,physics,sc_export</atm_procs_list>
 
     <!-- Basic options for each atm process -->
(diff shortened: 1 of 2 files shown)


components/eamxx/cime_config/namelist_defaults_scream.xml
```diff
@@ -317,6 +319,9 @@ be lost if SCREAM_HACK_XML is not enabled.
       <cosp_frequency_units valid_values="steps,hours">hours</cosp_frequency_units>
     </cosp>
 
+    <!-- Turbulent Mountain Stress -->
+    <tms inherit="atm_proc_base"/>
```
bartgol, line 323: If this was not in master, how were noel/chris able to add tms via atmchange?

components/eamxx/cime_config/namelist_defaults_scream.xml
```diff
@@ -141,7 +141,7 @@ be lost if SCREAM_HACK_XML is not enabled.
       so it must be of the form (a,b,...).
       NOTE: *CANNOT* be changed.
     -->
-    <!-- <atm_procs_list>(sc_import,homme,((shoc,cldFraction,spa,p3),rrtmgp,sc_export))</atm_procs_list> -->
+    <!-- <atm_procs_list>(sc_import,homme,((tms,shoc,cldFraction,spa,p3),rrtmgp,sc_export))</atm_procs_list> -->
```
bartgol, line 144: We should remove this line. It uses an old syntax, so it is misleading.

bartgol (changes requested):

ambrad:
We expect to see that all pg2 nightly tests will require rebaselining (on baseline machines), right?

tcclevenger:
@ambrad Correct.

components/eamxx/cime_config/namelist_defaults_scream.xml
tcclevenger: That is a good question. It doesn't look like on the cess-production branch it is included either.

tcclevenger: Looking back through the eamxx-simulation chat, I think that Noel encountered the error of no TMS registered and commented out `./atmchange physics::mac_aero_mic::atm_procs_list="tms,..."` so most likely running without TMS, and Oksana was most likely running with a version of the code before TMS was removing .

ndkeen:
I would be ok with it going in as long as there is a clear way of turning it off in the script (if needed).

bartgol (approved):

bartgol:
> I would be ok with it going in as long as there is a clear way of turning it off in the script (if needed).

I think this should work
```
./atmchange mac_aero_mic::atm_procs_list=shoc,cldFraction,p3
```

crterai:
@tcclevenger - Yes, let's add this to the cess-production branch as well.

tcclevenger:
@AaronDonahue DIFFs on Mappy expected, there is a random fail on weaver in DP unit tests which has nothing to do with this PR. If you are ok with it, feel free to merge. 

I'm going to disable DP tests in a PR since we are going a different route with that conversion. Don't want fails in those test to mess with AT anymore.

bartgol:
@tcclevenger I say go ahead and merge.

tcclevenger:
@bartgol @ambrad We'll need new baselines. I can take care of Mappy AT, do either of you have access to the others?

bartgol:
We can't create baselines on Chrysalis until nightlies run. I don't think pm does baseline tests. I don't know about crusher/ascent, since they are not handled through the jenkins farms.

bartgol:
> I'm going to disable DP tests in a PR since we are going a different route with that conversion. Don't want fails in those test to mess with AT anymore.

@tcclevenger are you still planning on doing this? All PRs are failing b/c of this test, so I'd like to disable it asap.

tcclevenger:
@bartgol Yes, completely forgot to submit the PR. Will do it now.

PR #2564, opened 2023-10-02
Fix attributes for ISCCP outputs

brhillman:
Fix attributes for ISCCP outputs. This sets the units attribute to "%" for `isccp_cldtot` and `isccp_ctptau` to reflect the fact that these quantities are expressed as percent area cover, and also adds a note that time averages of these quantities need to be divided by the time averaged `isccp_mask`, since these are only valid during daytime and we set values to zero during nighttime. This is a temporary workaround because we cannot yet aggregate masked time averages, but this information should be in the netcdf files to avoid confusing potential users of the data who may not peek inside the code to figure out how these quantities are handled.


bartgol (approved):

AaronDonahue:
@brhillman , is this alright to add the AUTOMERGE label to?

PR #2566, opened 2023-10-02
Disable DP tests

tcclevenger:
DP unit tests causing AT to fail. Disabling.

DP will change a lot after dev meeting, so not even sure these test will still exist. No risk in disabling them.


tcclevenger:
@bartgol @jgfouca This probably doesn't need to be tested. Feel free to just merge if this is the correct fix.

bartgol (approved):

PR #2571, opened 2023-10-04
Allow to request fields simpy by name and grid

bartgol:
@AaronDonahue this one was sitting on the server for a few weeks, so I decided to run the tests, and polish it. I remember you needed something like this (maybe for  nudging?), not sure if you still do.


AaronDonahue (commented):
This is great! and yes, this will be helpful for making nudging more general.  Also helpful for ML I believe.  I would approve but it looks like the `field_at_X` diagnostics tests are failing.

AaronDonahue (approved):

PR #2572, opened 2023-10-05
In buildnml, split apply_buffer in two parts

bartgol:
Namely

- The first call is done at the beginning of the xml file generation. It runs on the xml tree of the defaults, and only applies changes that alter the atm procs used. We need this early on, since below we build the atm proc group, so we need to have the correct list of atm procs.
-  The second call is done at the end, when the xml has been fully created. In this pass, we will skip all changes that alter the atm procs used, since we already took care of those in the first call.

Note: this pr prevents us from doing something like this:
```
$ ./atmchange mac_aero_mic::atm_procs_list=shoc,spa,p3,blah
$ ./atmchange blah:foo=bar
$ ./atmchange mac_aero_mic::atm_procs_list=shoc,spa,p3
```
That's b/c the 1st and 3rd line will be done in the first call, while the 2nd line is done in the second call. By the time the 2nd atmchange is issues, there is no longer a "blah" node in the XML. This is a mild limitation: if you do something like the above, you may as well issue `./atmchange --reset` and start over...

This PR should fix all the fails we are currently seeing with atmchange. The reason for the fails is that the buffer was unpacked before resolving inheritances.

Note: wip while I test the scripts locally.

### DIFF
--- a/components/eamxx/scripts/atmchange
+++ b/components/eamxx/scripts/atmchange
@@ -14,7 +14,7 @@ sys.path.append(os.path.dirname(os.path.realpath(__file__)))
 
 from eamxx_buildnml_impl import check_value, is_array_type
 from eamxx_buildnml import create_raw_xml_file
-from atm_manip import atm_config_chg_impl, buffer_changes, reset_buffer, get_xml_nodes
+from atm_manip import atm_config_chg_impl, buffer_changes, reset_buffer, get_xml_nodes, parse_change
 from utils import run_cmd_no_fail, expect
 
 # Add path to cime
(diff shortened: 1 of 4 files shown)


jgfouca (changes requested):
Can we add some tests to catch this in cime-buildnml-tests?

components/eamxx/scripts/atmchange
```diff
@@ -52,6 +52,12 @@ def atm_config_chg(changes, reset=False, all_matches=False, buffer_only=False):
     else:
         expect(changes, "Missing <param>=<val> args")
 
+    # Before applying/buffering changes, at the very least check the syntax
+    for c in changes:
+        # This will throw if the syntax is bad
+        _, _, _ = parse_change(c)
+
+    any_change = True
```
jgfouca, line 60: I think `any_change` should be initialized to `False`. Since we are OR-ing them together, initializing to `True` will mean it will always be `True`.

jgfouca (changes requested):
Looks good! Just one minor change needed.

components/eamxx/scripts/atmchange
bartgol: It is inited to False inside the `if not buffer_only` case. But if `buffer_only=True`, we must assume there are changes, since we have no way to check it (unless we go through them all). It is a bit redundant. I can remove one line and init it as `any_change = buffer_only`.

jgfouca (approved):

PR #2574, opened 2023-10-09
Fix issue w/ U and V in nudging process

AaronDonahue:
In the current setup, the fields U and V are added to the field manager as subfields of horiz_winds *after* the field manager has already registerd all field requests from atmospheric processes.  As a result, nudging, which may register U and V disrupts the assumptions about horiz_winds currently hard-coded.  When U and V are used in nudging they are added as independent fields and are no longer tied to horiz_winds.

This commit is a quick fix to address the immediate issue in nudging.  A longer term fix will be require a bit more work and will be added in a subsequent PR.

### DIFF
--- a/components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
+++ b/components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
@@ -54,14 +54,13 @@ void Nudging::set_grids(const std::shared_ptr<const GridsManager> grids_manager)
    * vector, if it is we register it.  For now we are limited to just T_mid, qv,
    * U and V
    */
-  if (std::find(m_fields_nudge.begin(),m_fields_nudge.end(),"T_mid") != m_fields_nudge.end()) {
+  if (ekat::contains(m_fields_nudge,"T_mid")) {
     add_field<Updated>("T_mid", scalar3d_layout_mid, K, grid_name, ps);
   }
-  if (std::find(m_fields_nudge.begin(),m_fields_nudge.end(),"qv") != m_fields_nudge.end()) {
+  if (ekat::contains(m_fields_nudge,"qv")) {
     add_field<Updated>("qv",    scalar3d_layout_mid, Q, grid_name, "tracers", ps);
   }
-  if (std::find(m_fields_nudge.begin(),m_fields_nudge.end(),"U") != m_fields_nudge.end() or
-      std::find(m_fields_nudge.begin(),m_fields_nudge.end(),"V") != m_fields_nudge.end()) {
+  if (ekat::contains(m_fields_nudge,"U") or ekat::contains(m_fields_nudge,"V")) {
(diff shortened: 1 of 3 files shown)


components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
```diff
@@ -348,5 +346,22 @@ void Nudging::init_buffers(const ATMBufferManager& buffer_manager) {
...
+// =========================================================================================
+Field Nudging::get_field_out_wrap(const std::string& field_name) {
+  if (field_name == "U" or field_name == "V") {
+    auto hw = get_field_out("horiz_winds");
+    const auto& fid = hw.get_header().get_identifier();
+    const auto& layout = fid.get_layout();
+    const int vec_dim = layout.get_vector_dim();
+    const auto& units = fid.get_units();
+    if (field_name == "U") {
+      return hw.subfield("U",units,vec_dim,0);
```
bartgol, line 358: You can do this much more easily by using `Field::get_component`.

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
```diff
@@ -60,11 +60,9 @@ void Nudging::set_grids(const std::shared_ptr<const GridsManager> grids_manager)
   if (std::find(m_fields_nudge.begin(),m_fields_nudge.end(),"qv") != m_fields_nudge.end()) {
     add_field<Updated>("qv",    scalar3d_layout_mid, Q, grid_name, "tracers", ps);
   }
-  if (std::find(m_fields_nudge.begin(),m_fields_nudge.end(),"U") != m_fields_nudge.end()) {
-    add_field<Updated>("U",     scalar3d_layout_mid, m/s, grid_name, ps);
-  }
-  if (std::find(m_fields_nudge.begin(),m_fields_nudge.end(),"V") != m_fields_nudge.end()) {
-    add_field<Updated>("V",     scalar3d_layout_mid, m/s, grid_name, ps);
+  if (std::find(m_fields_nudge.begin(),m_fields_nudge.end(),"U") != m_fields_nudge.end() or
```
bartgol, line 63: For brevity, you could do
```
if (ekat::contains(m_fields_nudge,"U") or ekat::contains(m_fields_nudge,"V"))
```

bartgol (commented):
I have some suggestions. I will review again when AT passes.

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
```diff
@@ -355,9 +354,9 @@ Field Nudging::get_field_out_wrap(const std::string& field_name) {
     const int vec_dim = layout.get_vector_dim();
```
bartgol, line 354: fid, layout, vec_dim, and units are all unused vars now, so we should remove them.

bartgol (approved):

PR #2575, opened 2023-10-11
Add team_barriers to water path tests

tcclevenger:
Adding `team_barrier` in between `parallel_reduce()` calls in the `water_path` test clears the compute-sanatizer racecheck fails. I'm not exactly sure why, my guess is that a team-level `parallel_reduce()` reuses some metadata. 

This is only in test code so there is not a negative impact of adding these barriers, but if we want to wait to merge until we understand exactly what is happening that is also ok.

### DIFF
--- a/components/eamxx/src/diagnostics/tests/water_path_tests.cpp
+++ b/components/eamxx/src/diagnostics/tests/water_path_tests.cpp
@@ -217,6 +217,7 @@ void run(std::mt19937_64& engine)
             }
           }, Kokkos::Max<Real>(qv_mass_max));
         EKAT_KERNEL_REQUIRE(vwp_v(icol)>qv_mass_max);
+        team.team_barrier();
         // Test qc mass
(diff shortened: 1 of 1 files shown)


ambrad:
I think Luca's referring to this issue: https://github.com/kokkos/kokkos/issues/5475.

ambrad:
Looks like our current Kokkos version doesn't have the fix (3.7.00 vs 3.7.01), so this fix compensates for that bug.

tcclevenger:
@bartgol @ambrad Thanks, I could add a `TODO` here so we can remove when we update Kokkos, but probably isn't necessary.

bartgol:
> @bartgol @ambrad Thanks, I could add a `TODO` here so we can remove when we update Kokkos, but probably isn't necessary.

Yeah, no need for that. This is a unit test, and an extra barrier does not matter.

PR #2579, opened 2023-10-17
Add more robust log entries for IO

AaronDonahue:
This commit adds atmosphere log capability to the AtmosphereInput and AtmosphereOutput classes.  This should allow for more robust logging of IO in the atmosphere log.  This commit adds log messages for both of the "run" routines for Input and Output respectively, including timers to record how long the run routines have taken.

[BFB]

### DIFF
--- a/components/eamxx/src/share/io/scorpio_output.cpp
+++ b/components/eamxx/src/share/io/scorpio_output.cpp
@@ -366,6 +366,12 @@ run (const std::string& filename,
   if (not is_write_step and m_avg_type==OutputAvgType::Instant) {
     return;
   }
+  auto func_start = std::chrono::steady_clock::now();
+  if (is_write_step) {
+    if (m_atm_logger) {
+      m_atm_logger->info("[EAMxx::scorpio_output] Writing variables to file:\n\t " + filename + " ...\n");
+    }
+  }
 
   using namespace scream::scorpio;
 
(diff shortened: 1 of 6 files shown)


AaronDonahue:
This PR is a result of the conversation we had today on the Eval call where we discussed the possibility that hangs we are seeing on frontier could be in IO.   The idea is that if we could see in the log file when IO starts and ends we could rule out IO as the cause of the hang... or rule it in.

AaronDonahue:
@ndkeen I also added simple timers to the new log entries as you had suggested.

components/eamxx/src/share/io/scorpio_output.cpp
```diff
@@ -638,6 +644,13 @@ run (const std::string& filename,
       grid_write_data_array(filename,name,view_host.data(),view_host.size());
     }
   }
+  auto func_finish = std::chrono::steady_clock::now();
+  if (is_write_step) {
+    if (m_atm_logger) {
+      auto duration = std::chrono::duration_cast<std::chrono::milliseconds>(func_finish - func_start)/1000.0;
+      m_atm_logger->info("[EAMxx::scorpio_output] Writing variables to file:\n\t " + filename + " ...done! (Elapsed time = " + std::to_string(duration.count()) +" seconds)\n");
```
bartgol, line 651: To be fair, this is not just measuring the time it takes to write to file, but also the time it took to remap and do the accumulation. You could either always print this (regardless of the value of `is_write_step`) or maybe do a more accurate clocking of the pio write function.

bartgol (dismissed):
Looks good. I have one suggestion on the clocking part, but I guess it depends on what you wanted to clock in the first place...

components/eamxx/src/share/io/scorpio_output.cpp
AaronDonahue: Good point.  I could probably just assign `func_start` before the grid_write_data_array call to capture just the time in writing the file to more accurately capture the time in pio.

bartgol: Yeah, that's why I said I didn't know what you wanted to capture. PIO already logs the time it takes to write, but I guess this would give you a sense of how much each write costs (e.g., restart vs output). Up to you.

bartgol (approved):

PR #2580, opened 2023-10-18
regional nudging using user provided nudging weights

xyuan:
regional nudging using user provided nudging weights file with the same gridset and vertical level


PeterCaldwell:
@xyuan - is this PR ready to review?

xyuan:
> @xyuan - is this PR ready to review?

yes, I tested ne4pg2, and ne30pg2 case, and both are working fine

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
```diff
@@ -40,6 +45,8 @@ void Nudging::set_grids(const std::shared_ptr<const GridsManager> grids_manager)
   FieldLayout scalar3d_layout_mid { {COL,LEV}, {m_num_cols, m_num_levs} };
   FieldLayout horiz_wind_layout { {COL,CMP,LEV}, {m_num_cols,2,m_num_levs} };
 
+printf("set grids: %d, %d\n",m_num_cols, m_num_levs);
```
AaronDonahue, line 48: Is this left over from debugging?  If we do want to keep this message then I think we should use the atmosphere logger instead of printf.

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
```diff
@@ -118,9 +154,14 @@ void Nudging::initialize_impl (const RunType /* run_type */)
...
     auto pmid_ext = get_helper_field("p_mid_ext");
     auto pmid_ext_v = pmid_ext.get_view<Real*,Host>();
-    in_params.set<std::vector<std::string>>("Field Names",{"p_levs"});
-    host_views["p_levs"] = pmid_ext_v;
-    layouts.emplace("p_levs",scalar2d_layout_mid);
+//    in_params.set<std::vector<std::string>>("Field Names",{"p_levs"});
+//    host_views["p_levs"] = pmid_ext_v;
+//    layouts.emplace("p_levs",scalar2d_layout_mid);
+
+    in_params.set<std::vector<std::string>>("Field Names",{"hybm"});
```
AaronDonahue, line 161: I don't think we can expect all nudging source files to use the hybrid coordinate system for the vertical pressure profile.  In fact, `p_levs` is used as the variable name for our vertical remapping because we pass explicit pressure levels for that.  

What is the goal with switching to `hybm` here?  I don't see this being used to build the pressure profile using the hybrid coordinate system anywhere, and wouldn't you need surface pressure as well?

If RRM nudging needs to use the hybrid coordinate system I would suggest adding another pressure type option to the nudging interface, in addition to `TIME_DEPENDENT_3D_PROFILE` and `STATIC_1D_VERTICAL_PROFILE`

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
```diff
@@ -294,9 +350,19 @@ void Nudging::run_impl (const double dt)
       Kokkos::deep_copy(atm_state_view,int_state_view);
     } else {
       // Back out a tendency and apply it.
-      apply_tendency(atm_state_field, int_state_field, dt);
+      if (m_use_weights <= 0) {
+        apply_tendency(atm_state_field, int_state_field, dt);
+      } else {
+        // get nudging weights field
```
AaronDonahue, line 356: Question, do we apply weights by column or is it fully 3D?  If it's by column then we shouldn't need to vertically interpolate the weights as you suggested.

AaronDonahue (changes requested):
Thanks for putting this together @xyuan!  I had a few questions about the PR, see below.

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
AaronDonahue: Maybe a name like "TIME_DEPENDENT_3D_HYBRID" for one that uses horizontally varying surface pressure.  Of course you would also need to load `ps` from file.

xyuan: sorry, this is also debug test, I should remove it

xyuan: since I don't have the plev data, so I tested it using hybm as an alternative option.

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
xyuan: yeah, let me remove it

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
AaronDonahue: Ahh okay, that makes sense.  Then I would think we need to vertically interpolate the weights if we are already vertically interpolating the actual nudging data, to be consistent.  Also, one could imagine nudging data that is on a different vertical grid than the simulation, for example 72 vs. 128 levels.

xyuan: sure, we can call the online mapping function to do the interpolation on vertical grid

components/eamxx/cime_config/namelist_defaults_scream.xml
```diff
@@ -231,6 +231,8 @@ be lost if SCREAM_HACK_XML is not enabled.
       <nudging_filename type="array(string)"/>
       <nudging_fields type="array(string)" doc="List of fields to be nudged.  Note, syntax of 'A:B' represents nudging field A with data from field B in files, syntax of 'A' assumes that nudging file has the same variables name as EAMxx"/>
       <nudging_timescale type="integer" doc="Timescale to apply nudging tendencies, 0: full replacement, >0: actual timescale">0</nudging_timescale>
+      <use_nudging_weights type="integer" doc="Flag for nudging weights option"/>
```
PeterCaldwell, line 234: Would it make more sense to have "use_nudging_weights" be true or false rather than an integer? Is zero false and 1 true?

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
```diff
@@ -91,6 +101,30 @@ void Nudging::apply_tendency(Field& base, const Field& next, const int dt)
   base.update(tend,dtend,Real(1.0));
 }
 // =========================================================================================
+void Nudging::apply_weighted_tendency(Field& base, const Field& next, const Field& weights, const int dt)
+{
+  // Calculate the weight to apply the tendency
+  const Real dtend = Real(dt)/Real(m_timescale);
```
PeterCaldwell, line 107: is dt the model timestep? I'm surprised it has integer type.

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
PeterCaldwell: My vision was that the user would have to supply nudging weights on exactly the same grid as used for the physics parameterizations in the target simulation - no interpolation in horizontal, vertical or time (because the same weights would be used for all times). My logic is that it should be super-trivial to create a weights file for your target grid (just open an output file with a 3d output on that grid and the lat/lon dimensions, create a variable of the same size minus the time dimension, then loop over all the columns and assign a weight based on its lat/lon... Easier to create a new weights file than to add code for interpolation.

PeterCaldwell:
Wouldn't it be simpler to have initialize_impl read in the nudging weights if m_use_weights is true and otherwise just initialize all the entries in nudging_weights_field to 1? Then instead of having an if/then for whether we have weights or not in apply_tendency you would always have weights. And wouldn't it be easier and equivalent to do base.update(tend,dtend*nudging_weights,1.0) in apply_tendency?

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
yuenxq: @PeterCaldwell, sure, the code is working in the way you described at the moment

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
AaronDonahue: Yes, the model timestep is passed as an integer to atmosphere processes.  I can't quite remember why, @bartgol may.

PeterCaldwell: I think E3SM/CPL7 may be specifying dt as an integer (which saves a bit of memory), so it would be natural for us to do so as well. Still, I'm surprised we don't recast it to double. In particular, the ability to substep procs at runtime seems like it would naturally result in non-integer timesteps. For example, what if full-model dt is 150 sec and a user asks for 7 micro substeps (=21.43 sec micro dt)?

bartgol: The timestep is _not_ an integer. It is passed by the cpl to the AD as an integer, but atm processes expect a double.

bartgol:
> Wouldn't it be easier and equivalent to do base.update(tend,dtend*nudging_weights,1.0) in apply_weighted_tendency?

The `update` method does not support the 2nd argument (alpha) to be a field itself, it must be a scalar.

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
```diff
@@ -63,6 +68,10 @@ void Nudging::set_grids(const std::shared_ptr<const GridsManager> grids_manager)
   if (ekat::contains(m_fields_nudge,"U") or ekat::contains(m_fields_nudge,"V")) {
     add_field<Updated>("horiz_winds",   horiz_wind_layout,   m/s,     grid_name, ps);
   }
+
+  if (m_use_weights) 
+    add_field<Updated>("nudging_weights", scalar3d_layout_mid, A, grid_name, "weights", ps);
```
bartgol, line 73: Why is this field exposed? It seems to me that this should be a field local to the atm proc.

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
```diff
@@ -91,6 +101,30 @@ void Nudging::apply_tendency(Field& base, const Field& next, const int dt)
...
+  // Calculate the weight to apply the tendency
+  const Real dtend = Real(dt)/Real(m_timescale);
+  EKAT_REQUIRE_MSG(dtend>=0,"Error! Nudging::apply_tendency - timescale tendency of " << std::to_string(dt)
+                  << " / " << std::to_string(m_timescale) << " = " << std::to_string(dtend)
+                  << " is invalid.  Please check the timescale and/or dt");
+  // Now apply the tendency.
+  Field tend = base.clone();
+
+  // Use update internal to set tendency, will be (weights*next - weights*base), note tend=base at this point.
+  auto base_view = base.get_view<const mPack**>();
```
bartgol, line 115: Since `mPack` is just a pack of size 1, why can't we just use `Real` and avoid the confusion of packs?

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
```diff
@@ -139,6 +173,20 @@ void Nudging::initialize_impl (const RunType /* run_type */)
   }
   m_time_interp.initialize_data_from_files();
 
+  // load nudging weights from file
+  if (m_use_weights)
+  {
+    auto nudging_weights = get_field_out("nudging_weights");
+    auto weights_layout = nudging_weights.get_header().get_identifier().get_layout();
+    create_helper_field("nudging_weights_ext", weights_layout, grid_name, ps);
```
bartgol, line 181: Looks like `nudging_weights_ext` is never used? Can we prune it?

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
AaronDonahue: Yeah, this is something that drove me crazy.  In short, the vertical interpolation function in EAMxx only works on Packs, so we cannot pass it an array of Reals.  I think the proper fix is to overhaul the vertical interpolation function to accept Real arrays.

bartgol: Yeah, but there's no interpolation here...

AaronDonahue: Ahh right, so we should change all for of these `get_view`'s to `Real`.

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
xyuan: yes, this can be the local field, if we have the time-independent nudging weights, so it doesn't need to be the global.

xyuan:
> > Wouldn't it be easier and equivalent to do base.update(tend,dtend*nudging_weights,1.0) in apply_weighted_tendency?
> 
> The `update` method does not support the 2nd argument (alpha) to be a field itself, it must be a scalar.

if the field update method support arguments of view type, it can remove the parallel_for loop. but it looks quite a lot work by doing so

components/eamxx/cime_config/namelist_defaults_scream.xml
xyuan: yeah, zero is false, and 1 is true, 
xml doesn't support _bool_, and _boolean_, see the error information below:
NameError: ERROR: Invalid/unsupported force type 'boolean', and 
NameError: ERROR: Invalid/unsupported force type 'bool'

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
xyuan: ok, that looks less confusion here

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
PeterCaldwell: Ok, so this should be fixed.

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
PeterCaldwell: I think this issue can be resolved by just deleting the comment lines directly below.

bartgol:
> > > Wouldn't it be easier and equivalent to do base.update(tend,dtend*nudging_weights,1.0) in apply_weighted_tendency?
> > 
> > 
> > The `update` method does not support the 2nd argument (alpha) to be a field itself, it must be a scalar.
> 
> if the field update method support arguments of view type, it can remove the parallel_for loop. but it looks quite a lot work by doing so

We can add that capability, but I think this would be the only use case. Since it's a small ||for, and adding it requires to also add testing, I'm inclined to not add the overload. If more places start to need a "weighted" update method, we can def think about it.

xyuan:
@AaronDonahue the test case fails because the YAML nudging file is missing, could you tell where I can find these files, so I can generated the weight file based on your input

xyuan:
add create_nudging_weights for the test, and it passes on ANL gpu machine using cuda/11.8.0

xyuan@gpu07:/gpfs/jlse-fs0/projects/climate/xyuan/scream.nudging/components/eamxx/build/tests/coupled/physics_only/shoc_p3_nudging> ctest
Test project /gpfs/jlse-fs0/projects/climate/xyuan/scream.nudging/components/eamxx/build/tests/coupled/physics_only/shoc_p3_nudging
    Start 1: create_vert_remap
1/5 Test #1: create_vert_remap ................   Passed    0.81 sec
    Start 2: create_nudging_weights
2/5 Test #2: create_nudging_weights ...........   Passed    0.57 sec
    Start 3: shoc_p3_source
3/5 Test #3: shoc_p3_source ...................   Passed    1.42 sec
    Start 4: shoc_p3_nudged
4/5 Test #4: shoc_p3_nudged ...................   Passed    1.50 sec
    Start 5: shoc_p3_nudged_remapped
5/5 Test #5: shoc_p3_nudged_remapped ..........   Passed    1.52 sec

100% tests passed, 0 tests failed out of 5

Total Test time (real) =   5.83 sec

components/eamxx/tests/coupled/physics_only/shoc_p3_nudging/CMakeLists.txt
```diff
@@ -26,7 +29,9 @@ configure_file(${CMAKE_CURRENT_SOURCE_DIR}/output_remapped.yaml
                ${CMAKE_CURRENT_BINARY_DIR}/output_source_data_remapped.yaml)
 CreateUnitTestFromExec (shoc_p3_source shoc_p3_nudging
       EXE_ARGS "--use-colour no --ekat-test-params ifile=input_source_data.yaml"
-      PROPERTIES FIXTURES_SETUP source_data FIXTURES_REQUIRED create_vertical_remap_file)
+      PROPERTIES FIXTURES_SETUP source_data 
+      FIXTURES_REQUIRED create_vertical_remap_file 
+      FIXTURES_REQUIRED create_nudging_weights_file)
```
bartgol, line 34: This won't work. Passing the same property twice causes the second instance to override the first one. To set multiple fixtures, you should do this:
```
  CreateUnitTestFromExec (....
     FIXTURES_REQUIRED "create_vertical_remap_file;create_nudging_weights_file")
```

bartgol (changes requested):

components/eamxx/tests/coupled/physics_only/shoc_p3_nudging/CMakeLists.txt
xyuan: but I have this error, when I use FIXTURES_REQUIRED "create_vertical_remap_file;create_nudging_weights_file")

CMake Error at /gpfs/jlse-fs0/projects/climate/xyuan/scream.nudging/externals/ekat/cmake/EkatCreateUnitTest.cmake:375 (set_tests_properties):
  set_tests_properties called with incorrect number of arguments.
Call Stack (most recent call first):
  cmake/ScreamUtils.cmake:130 (EkatCreateUnitTestFromExec)
  tests/coupled/physics_only/shoc_p3_nudging/CMakeLists.txt:30 (CreateUnitTestFromExec)

bartgol: Ah, yes, we have some issue with passing some properties that are lists. I tried this with `add_test`, but not with the Ekat utility. Fortunately, we have a special handling for FIXTURES...but unfortunately there is a bug in it. If you can hold this PR for a bit more, I can fix it in Ekat, then update ekat in scream repo?

xyuan: sure, please go ahead, it is possibly the reason that the cpu test failed, it shows _"Failed test depencies: shoc_p3_source"_

bartgol: The PR was merged in ekat. #2586 is undergoing testing right now. Once merged, we can re-test this, and should hopefully pass.

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
bartgol: @xyuan we still need to fix dt here. The method should accept a Real, not an integer, to avoid silent truncation by the compiler.

components/eamxx/cime_config/namelist_defaults_scream.xml
bartgol: Ah, we are using `logical` in our XML settings. Can you try with that?

bartgol (changes requested):

components/eamxx/tests/coupled/physics_only/shoc_p3_nudging/create_vert_remap_and_weights.cpp
```diff
@@ -0,0 +1,85 @@
...
+  scorpio::eam_pio_enddef(filename);
+  scorpio::grid_write_data_array(filename,"nudging_weights",&weights[0][0][0],ntimes*ncols*nlevs);
+  scorpio::eam_pio_closefile(filename);
+  scorpio::eam_pio_finalize();
+}
+
+TEST_CASE("create_vert_remap_and_weights","create_vert_remap_and_weights")
+{
+  create_vert_remap();
+  create_nudging_weights_ncfile(1, 218, 72, "nudging_weights_remapped.nc");
```
bartgol, line 82: These two calls have the same parameters, so I'm guessing it will create the same output. Do we really need two files that are identicals?

bartgol (dismissed):

bartgol:
Github is blocking the PR merge b/c Aaaron requested changes, so it waits for his approval. I think he may be out today. Since his comments have been addressed, I have removed him as a reviewer, so that the integration can continue. I am not sure if gh will still prevent the PR from merging even if he's no longer a reviewer. Maybe. If so, I can force-merge the PR.

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.hpp
```diff
@@ -115,8 +118,12 @@ class Nudging : public AtmosphereProcess
   int m_num_levs;
   int m_num_src_levs;
   int m_timescale;
+  int m_use_weights;
```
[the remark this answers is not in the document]
AaronDonahue: It looks like this is still an `int`.  Should it b e a `bool`?

AaronDonahue (changes requested):

bartgol (approved):

components/eamxx/cime_config/namelist_defaults_scream.xml
```diff
@@ -231,6 +231,8 @@ be lost if SCREAM_HACK_XML is not enabled.
       <nudging_filename type="array(string)"/>
       <nudging_fields type="array(string)" doc="List of fields to be nudged.  Note, syntax of 'A:B' represents nudging field A with data from field B in files, syntax of 'A' assumes that nudging file has the same variables name as EAMxx"/>
       <nudging_timescale type="integer" doc="Timescale to apply nudging tendencies, 0: full replacement, >0: actual timescale">0</nudging_timescale>
+      <use_nudging_weights type="logical" doc="Flag for nudging weights option"/>
```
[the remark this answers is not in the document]
bartgol: Another option would be to actually set a default... Empty parameters kind of make sense for lists/strings, while for other types, leaving them empty makes the interpretation somewhat arbitrary...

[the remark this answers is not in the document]
bartgol: I now see that that's exactly what you did in the PR you linked. :)

PR #2581, opened 2023-10-24
add clean-clear-sky and clean-sky radiation calls

mahf708:
Adds two additional radiation diagnostic calls following the logic of clear-sky calls. These are clean-clear-sky (neither clouds nor aerosols) and clean-sky (no aerosols). The latter in enabled by instantiating a second optics and never endowing it with aerosols, while the former is called before the other optics are endowed with aerosols and clouds.

TODO:
- [x] some tests
- [x] performance guard
- [x] add docs


jeff-cohere:
Looks good to me so far!

bartgol (dismissed):

components/eamxx/src/physics/rrtmgp/scream_rrtmgp_interface.cpp
```diff
@@ -728,6 +766,8 @@ namespace scream {
             bool top_at_1 = p_lay_host(1, 1) < p_lay_host(1, nlay);
 
             k_dist.gas_optics(nday, nlay, top_at_1, p_lay_day, p_lev_day, t_lay_limited, gas_concs_day, optics, toa_flux);
+            k_dist.gas_optics(nday, nlay, top_at_1, p_lay_day, p_lev_day, t_lay_limited, gas_concs_day, optics_no_aerosols, toa_flux);
```
brhillman, line 769: This is a fairly expensive operation, and will be identical to the operation with `optics` with aerosol. I think it might be preferable to simple do a deep copy of the `optics` computed, unless this is meant to allow for changing gas optics as well (but we'd probably want to implement that a little differently to mimic the EAM-style diag calls).

components/eamxx/src/physics/rrtmgp/scream_rrtmgp_interface.cpp
```diff
@@ -838,12 +901,16 @@ namespace scream {
 
             // Do gas optics
             k_dist.gas_optics(ncol, nlay, top_at_1, p_lay, p_lev, t_lay_limited, t_sfc, gas_concs, optics, lw_sources, real2d(), t_lev_limited);
+            k_dist.gas_optics(ncol, nlay, top_at_1, p_lay, p_lev, t_lay_limited, t_sfc, gas_concs, optics_no_aerosols, lw_sources, real2d(), t_lev_limited);
```
brhillman, line 904: Again, I think this is fairly expensive, and we might think about just doing a deep copy of the `optics` object.

components/eamxx/src/physics/rrtmgp/scream_rrtmgp_interface.cpp
```diff
@@ -838,12 +901,16 @@ namespace scream {
...
 
 #ifdef SCREAM_RRTMGP_DEBUG
             // Check gas optics
             check_range(optics.tau,  0, std::numeric_limits<Real>::max(), "rrtmgp_lw:optics.tau");
 #endif
 
+            // Compute clean-sky fluxes before we add in aerosols
```
brhillman, line 911: Isn't this one clean-clear-sky fluxes?

brhillman:
Minor comments. This PR will have an impact on radiation performance. This is probably fairly minimal relative to total throughput since we do not run radiation every timestep, but we should evaluate the impact. Note that the additional calls to gas_optics are unnecessary, and we could probably alternatively just deep copy the optics object after the first call to gas optics. This would recover some performance, and if I remember right the gas_optics call is not exactly cheap. We might also consider NOT doing these extra calls if we are not outputting the additional diag calls. That's the strength of the EAM approach that does these in a loop, which we may consider in the future, but this is certainly the simplest approach for getting this functionality in ASAP.

PeterCaldwell:
@mahf708 - thanks for these heuristic results. I'd also be curious how much the total wall clock time for the simulations increases from double-call radiation.

components/eamxx/src/physics/rrtmgp/scream_rrtmgp_interface.cpp
```diff
@@ -760,6 +820,9 @@ namespace scream {
             // Combine gas and cloud optics
             clouds_day.delta_scale();
             clouds_day.increment(optics);
+            if (extra_clnsky_diag) {
+                clouds_day.increment(optics_no_aerosols);
```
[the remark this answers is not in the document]
bartgol: I'm not familiar with the `OpticalProps2str` data structure, so I can't comment. I'll let Ben decide.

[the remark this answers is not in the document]
brhillman: My initial reaction to this is that I don't personally like the idea of adding a decrement method to the external optics class, but I'll have to think about that a little more. I think it makes more sense to create the combinations we want here by addition rather than subtraction, and in the future based on a set of lists of constituents for each diag call.

components/eamxx/cime_config/namelist_defaults_scream.xml
```diff
@@ -308,6 +308,8 @@ be lost if SCREAM_HACK_XML is not enabled.
       <do_aerosol_rad>true</do_aerosol_rad>
       <do_aerosol_rad COMPSET=".*SCREAM.*noAero">false</do_aerosol_rad>
       <enable_column_conservation_checks>false</enable_column_conservation_checks>
+      <extra_clnclrsky_diag>false</extra_clnclrsky_diag>
```
bartgol, line 311: Would you mind adding the `type="logical"` specifier here, and perhaps, also a `doc="blah"` string? It helps with the auto-generated documentation.

bartgol: If you don't mind, and you have time to spare (haha), you could also add it for some of the other vars here (not the ones inherited from `atm_proc_base`, only the rrtmgp-specific ones).

components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
```diff
@@ -120,18 +120,32 @@ void RRTMGPRadiation::set_grids(const std::shared_ptr<const GridsManager> grids_
     add_field<Required>("aero_tau_lw", scalar3d_lwband_layout, nondim, grid_name, ps);
   }
 
+  // Whether we do extra clean/clear sky calculations
+  m_extra_clnclrsky_diag = m_params.get<bool>("extra_clnclrsky_diag", false);
```
bartgol, line 124: Don't change it, but FYI, if the type can be inferred from the default, you don't need the template. In practice, the _only_ case where you need to specify the type is when you pass a C-string as default, since we store strings as C++ strings. So this
```C++
params.get("my_string","my_value");
```
will fail at compile time (there's some overload that is not specified for C-strings), and you need to instead do
```C++
params.get<std::string>("my_string","my_value");
```
For virtually any other type we use, the provided default value is enough.

components/eamxx/src/physics/rrtmgp/scream_rrtmgp_interface.cpp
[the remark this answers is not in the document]
bartgol: I'm not really following. If/when you want to implement this, you can of course ping me and go over the details, if you need.

bartgol (dismissed):

jeff-cohere:
This PR has accumulated some history and, while it's not huge, has enough detail in it that I suspect I'm not following it as closely as I need to in order to understand the details. I didn't see anything that concerns me that hasn't been pointed out by others. @mahf708 , let me know if there's anything in particular you'd like me to examine more closely.

components/eamxx/src/physics/rrtmgp/scream_rrtmgp_interface.cpp
```diff
@@ -222,12 +222,17 @@ namespace scream {
                 real3d &cld_tau_lw_gpt,
                 real2d &sw_flux_up, real2d &sw_flux_dn, real2d &sw_flux_dn_dir,
                 real2d &lw_flux_up, real2d &lw_flux_dn,
+                real2d &sw_clnclrsky_flux_up, real2d &sw_clnclrsky_flux_dn, real2d &sw_clnclrsky_flux_dn_dir,
```
brhillman, line 225: This is probably not worth changing (especially since I want for us to refactor this soon to generalize to N-diag calls), but I think it might have been better to order the args to these calls as allsky, clrsky, clnsky, clnclrsky just to keep the "always on" fields first in the list (the cln versions are only computed if the extra bool args are true).

components/eamxx/src/physics/rrtmgp/scream_rrtmgp_interface.cpp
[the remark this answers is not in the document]
brhillman: Naser and I talked about this a week or two ago, I actually have a slightly different idea about how to handle this.

components/eamxx/src/physics/rrtmgp/scream_rrtmgp_interface.cpp
```diff
@@ -760,6 +820,9 @@ namespace scream {
             // Combine gas and cloud optics
             clouds_day.delta_scale();
             clouds_day.increment(optics);
+            if (extra_clnsky_diag) {
```
brhillman, line 823: I would move this to the extra_clnsky_diag logical chunk below to keep everything related to the extra_cldsky_diag call together so it's more obvious what goes into that calculation.

components/eamxx/src/physics/rrtmgp/tests/rrtmgp_tests.cpp
```diff
@@ -217,6 +234,20 @@ int run(int argc, char** argv) {
...
+    if (!rrtmgpTest::all_close(sw_flux_dn , sw_clnsky_flux_dn , 0.0000000001)) nerr++;
+    if (!rrtmgpTest::all_close(sw_clrsky_flux_dn , sw_clnclrsky_flux_dn , 0.0000000001)) nerr++;
+    if (!rrtmgpTest::all_close(sw_flux_dir , sw_clnsky_flux_dir , 0.0000000001)) nerr++;
+    if (!rrtmgpTest::all_close(sw_clrsky_flux_dir , sw_clnclrsky_flux_dir , 0.0000000001)) nerr++;
+    if (!rrtmgpTest::all_close(lw_flux_up , lw_clnsky_flux_up , 0.0000000001)) nerr++;
+    if (!rrtmgpTest::all_close(lw_clrsky_flux_up , lw_clnclrsky_flux_up , 0.0000000001)) nerr++;
+    if (!rrtmgpTest::all_close(lw_flux_dn , lw_clnsky_flux_dn , 0.0000000001)) nerr++;
+    if (!rrtmgpTest::all_close(lw_clrsky_flux_dn , lw_clnclrsky_flux_dn , 0.0000000001)) nerr++;
+
+    logger->info(nerr);
```
brhillman, line 249: Should this have any additional text string other than just the integer number with no context?

components/eamxx/src/physics/rrtmgp/tests/rrtmgp_unit_tests.cpp
```diff
@@ -841,3 +842,207 @@ TEST_CASE("rrtmgp_aerocom_cloudtop") {
 
   yakl::finalize();
 }
+
+/* comment out for now
```
brhillman, line 846: I would pull this block out of the PR. If it isn't ready to include, let's not commit commented out code.

components/eamxx/src/physics/rrtmgp/scream_rrtmgp_interface.cpp
```diff
@@ -741,6 +789,18 @@ namespace scream {
                 toa_flux(iday,igpt) = tsi_scaling * toa_flux(iday,igpt);
             });
 
+            if (extra_clnclrsky_diag) {
+                // Compute clear-clean-sky (just gas) fluxes on daytime columns
+                rte_sw(optics, top_at_1, mu0_day, toa_flux, sfc_alb_dir_T, sfc_alb_dif_T, fluxes_day);
```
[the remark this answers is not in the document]
brhillman: I prefer simpler and clearer to fine-grain control, personally.

components/eamxx/src/physics/rrtmgp/scream_rrtmgp_interface.cpp
```diff
@@ -838,12 +951,20 @@ namespace scream {
...
+            }
 
 #ifdef SCREAM_RRTMGP_DEBUG
             // Check gas optics
             check_range(optics.tau,  0, std::numeric_limits<Real>::max(), "rrtmgp_lw:optics.tau");
 #endif
 
+            if (extra_clnclrsky_diag) {    
+                // Compute clean-clear-sky fluxes before we add in aerosols and clouds
+                rte_lw(max_gauss_pts, gauss_Ds, gauss_wts, optics, top_at_1, lw_sources, emis_sfc, clnclrsky_fluxes);
```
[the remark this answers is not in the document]
brhillman: Again, I prefer simple/clear to fine-grained control (too many options and duplicate code blocks clutters an interface unnecessarily unless there's good reason for it). But, that shouldn't hold up the PR.

brhillman (approved):

PR #2584, opened 2023-10-25
Fix calculation of tendencies when parent atm proc is subcycled

### LINKED ISSUE #2583: Possible bug in how atm procs compute their tendencies
While discussing another issue related to subcycling, I realized we may be computing the wrong (or, better, incomplete) tendency when an atm proc is part of a subcycled group. Taking `mac_aero_mic` as an example (which contains `shoc, cld_fraction, p3`), we usually subcycle the whole group, while each proc is not subcycled within the group.

My fear is that [this line](https://github.com/E3SM-Project/scream/blob/9080ef7e7245a2b8b5af812bc187e64ea066a3c6/components/eamxx/src/share/atm_process/atmosphere_process.cpp#L484) is wrong. In particular, every time shoc is invoked, it loses all the previously accumulated tendency.

The best solution is probably to add another internal field (for each requested tendency), storing the value of the field at the beginning of the step. The AD already takes care of zeroing out the tendencies at the beginning of the time step, since they are declared as part of the "ACCUMULATED" group (which the AD treats accordingly). Hence, summing into the tendency field should always be safe.

bartgol:
The current tendency calculation assumed that, when an atm proc run method was called, it had the right to overwrite the content of the tend field with the start-of-step value of the field. However, this is wrong if the atm proc is part of a subcycled group. In that case, when the run method is called the second time, the tend field already contains a contribution from the previous subcycle iteration of the parent atm proc.

The simplest solution is to store the start-of-step value of the field in a separate object, to avoid overwriting the tend field.

Fixes #2583

### DIFF
--- a/components/eamxx/src/control/atmosphere_driver.cpp
+++ b/components/eamxx/src/control/atmosphere_driver.cpp
@@ -825,9 +825,6 @@ initialize_fields ()
     }
   }
 
-  // Zero out accumulated fields
-  reset_accummulated_fields();
-
 #ifdef SCREAM_HAS_MEMORY_USAGE
   long long my_mem_usage = get_mem_usage(MB);
   long long max_mem_usage;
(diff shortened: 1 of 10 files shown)


AaronDonahue (changes requested):
Looks good.  My only question would be, is there a way to create a test (or add to an existing one) to check for proper accumulation of tendencies when subcycling?

bartgol:
> Looks good. My only question would be, is there a way to create a test (or add to an existing one) to check for proper accumulation of tendencies when subcycling?

Thanks for making me look into this. Turned out that doing `t+=(fnew-fold)/dt` from within the proc yielded the wrong tendency (dt was dt/N in case of subcycling). I fixed it, and added a tend check test in the shoc_p3 subcycling test.

AaronDonahue (approved):

PR #2587, opened 2023-10-30
DP-SCREAM: Add IOP class and unit test 

tcclevenger:
Adds the `IntensiveObservationPeriod` (IOP) class with functions to handle IC/topo file inputs.

IOP is the generic name, where doubly periodic is the current implementation (@bogensch is this a correct characterization?). The class currently has 3 associated parameters: `doubly periodic mode` (must be set `true`), and `target_latitude,`target_longitude` which give the lat/lon for looking through the IC/topo files to find which column we use to initialize all field's columns. 

Also in the PR: hooks in AD for loading inputs for IOP case and a unit test which uses DP-SCREAM.

This is the minimum needed to run dp-scream. I'll update the [confluence page](https://acme-climate.atlassian.net/wiki/spaces/EAMXX/pages/3736895503/2023.Q2+C+DP-SCREAM) with all the tasks remaining.


components/eamxx/src/control/intensive_observation_period.cpp
```diff
@@ -0,0 +1,329 @@
...
+  }
+}
+
+namespace scream {
+namespace control {
+
+// Helper functions for reading files
+namespace {
+// Currently EAMxx scorpio interface doesn't have a way to inquire about dimension names
+bool file_has_dim(const std::string& filename, const std::string& dimname)
```
bartgol, line 25: I'm in favor of moving this in the IO layer.

components/eamxx/src/control/intensive_observation_period.cpp
```diff
@@ -0,0 +1,329 @@
...
+{
+  m_comm = comm;
+  m_params = params;
+  EKAT_REQUIRE_MSG(m_params.get<bool>("doubly_periodic_mode", false),
+                   "Error! Currently doubly_periodic_mode is the only use case for "
+	           "intensive observation period files.\n");
+
+  EKAT_REQUIRE_MSG(m_params.isParameter("target_latitude") && m_params.isParameter("target_longitude"),
+                   "Error! Using intensive observation period files requires "
+                   "target_lat and target_lon be gives as parameters in "
```
bartgol, line 62: I would spell out `lat` and `lon` in the message, since those are the names of the expected parameters.

components/eamxx/src/control/intensive_observation_period.cpp
```diff
@@ -0,0 +1,329 @@
...
+    const auto field_rank = fl.rank();
+    const auto fname = field_names_eamxx[i];
+
+    // If I am the rank with the minimum lat/lon distance,
+    // store data of values to broadcast.
+    const auto my_rank = m_comm.rank();
+    const auto col_mpi_rank = m_lat_lon_info[grid_name].mpi_rank_of_closest_column;
+    if (my_rank == col_mpi_rank) {
+      const auto col_idx = m_lat_lon_info[grid_name].local_column_index_of_closest_column;
+      switch (field_rank) {
```
bartgol, line 245: These fields were just created above. Since none of these is a subfield of another field, you actually _know_ that they are stored contiguously in memory. Therefore, you can simply use the field's internal pointers for MPI. Something like
```
for (const auto& f : fields) {
  const auto& fl = f.get_header().get_identifier().get_layout();
  const auto col_mpi_rank = m_lat_lon_info[grid_name].mpi_rank_of_closest_column;
  m_comm.broadcast(f.get_internal_view_data<Real,Host>(), fl.size(), col_mpi_rank);
}
```

components/eamxx/tests/coupled/dynamics_physics/homme_shoc_cld_spa_p3_rrtmgp_dp/homme_shoc_cld_spa_p3_rrtmgp_dp.cpp
```diff
@@ -0,0 +1,64 @@
+#include "catch2/catch.hpp"
```
bartgol, line 1: We ought to create some sort of common-ad-testing lib, so that we build this file only once. We have almost the exact identical file in all coupled/uncoupled test directories...

components/eamxx/src/control/intensive_observation_period.cpp
tcclevenger: Let me know what you think.

We can't be guaranteed that all ranks own the same amount of columns, so for the mpi rank that owns the column data, I transferred that data to the first column, then broadcasted the first column of data to all the other rank's first column, then copied to field manager data.

bartgol: Ah, yeah. You can't even assume that the other ranks own any column at all... You could create a one-col field, and bcast that, then loop over ncols and copy. Like this:
```
auto col_layout = fl.strip_dim(COL);
FieldIdentifier fid_tmp("temp",col_layout,dummy_units,"");
Field single_col (fid_tmp);
single_col.allocate_view();
if (comm.rank()==mpi_rank_of_closest_col) {
  single_col.deep_copy(f.subfield(0,index_of_closest_col),Host);
}
comm.broadcast(single_col.get_internal_view_data<Real,Host>(),col_layout.size(),mpi_rank_of_closest_col);
for (int icol=0; icol<fl.dim(COL); ++icol) {
   f.subfield(0,icol).deep_copy(single_col,Host);
}
f.sync_to_dev();
```

tcclevenger: I didn't think about subfielding and using deep_copy. I'll update, this is much nicer than the case approach I'm using.

tcclevenger: Update it. The only hiccup was using Field::deep_copy with subfield of Column-only fields, but I just added a `if (rank==1)` case when copying to and from `col_data` field.

components/eamxx/src/control/intensive_observation_period.cpp
```diff
@@ -189,137 +168,82 @@ read_fields_from_file_for_iop (const std::string& file_name,
...
+    auto& fm_field = field_mgr->get_field(fname);
+
+    // Create a temporary field to store the data from the
+    // single column of the closest lat/lon pair
+    const auto io_fid = io_field.get_header().get_identifier();
+    auto dims = io_fid.get_layout().dims();
+    dims[0] = 1;
+    FieldLayout col_data_fl(io_fid.get_layout().tags(), dims);
+    FieldIdentifier col_data_fid("col_data", col_data_fl, dummy_units, "");
+    Field col_data(col_data_fid);
```
bartgol, line 214: This is fine, but fyi, rank-0 fields are supported in EAMxx (if you tried and they didn't, let me know!). So you could have done
```
auto col_lt = io_fid.get_layout().strip_dim(FieldTag::Column);
```
This can potentially give you a rank 0 field, but those should work. Then, later on, you don't need checks on `col_data.rank()`, you should be able to simply do, e.g.
```
col_data.deep_copy(io_field.subfield(0,col_indx_with_data);
```
and
```
fm_field.subfield(0,icol).deep_copy<Host>(col_data);
```

bartgol (dismissed):

components/eamxx/src/control/intensive_observation_period.cpp
tcclevenger: I think then the field.deep_copy()is what complained. Let me check, if that is the case I think it should be implemented and I'll switch designs.

tcclevenger:
I'll set WIP until https://github.com/E3SM-Project/scream/pull/2590 is merged, or we decided to revert back to the old way (using rank-1 fields with a single column). Until then this PR will fail.

bartgol (approved):

components/eamxx/src/share/field/field_impl.hpp
```diff
@@ -197,6 +197,15 @@ deep_copy_impl (const Field& src) {
...
           " to field " + get_header().get_identifier().name() + ".  Layouts don't match.");
   const auto  rank = layout.rank();
+
+  // For rank 0 view, we only need to copy a single value and return
+  if (rank == 0) {
+    auto v     =     get_view<      ST,HD>();
+    auto v_src = src.get_view<const ST,HD>();
+    v() = v_src();
+    return;
+  }
```
tcclevenger, line 207: @bartgol No rush, just when you have a chance, let me know what you think about me exiting early here to avoid computing extents and policies etc. Rank 0 can just copy it's value and return.

bartgol: Yeah, early return is always ok (even though you should never trust people saying "always" in programming...).

bartgol (approved):

bartgol:
@tcclevenger I took the liberty of removing the automerge label, in case you wanted to wait for other reviews. If not, of course, feel free to re-add.

tcclevenger:
@jgfouca @bogensch Could I get one more set of eyes on this? The only thing here is setting up the class and loading in the ICs (finding column with the closest lat/lon to the parameters and copying that column to all other columns). And also setting up the unit test.

bogensch:
@tcclevenger this looks good to my naive c++ mind.  Thanks for your hard work on this thus far!

bogensch (approved):

PR #2588, opened 2023-10-31
ML correction for `u` and `v`

elynnwu:
In this PR, we add the capability of ML correction for `u` and `v`. Currently, this is done in a separate ML model, one for temperature and specific humidity, and the other for winds. Additional input fields are needed (`phis`) and will be passed to python regardless of whether `ML_model_path_uv` is specified. 

We also added `ML_correction_unit_test` in the namelist to accommodate the unit test that does not have access to lat/lon and surface geopotential. This allows us to move get lat/lon and phis to `set_grids`.

### DIFF
--- a/components/eamxx/cime_config/namelist_defaults_scream.xml
+++ b/components/eamxx/cime_config/namelist_defaults_scream.xml
@@ -242,8 +242,10 @@ be lost if SCREAM_HACK_XML is not enabled.
 
     <!-- ML correction -->
     <mlcorrection inherit="atm_proc_base">
-      <ML_model_path type="string" doc="Path to pre-trained ML model"/>
+      <ML_model_path_tq type="string" doc="Path to pre-trained ML model for temperature and specific humidity"/>
+      <ML_model_path_uv type="string" doc="Path to pre-trained ML model for wind fields"/>
       <ML_output_fields type="array(string)" doc="ML correction output variables, the following variables are supported: T_mid,qv,u,v"/>
+      <ML_correction_unit_test type="logical">false</ML_correction_unit_test>
     </mlcorrection>
 
     <!-- For internal testing only -->
--- a/components/eamxx/src/physics/ml_correction/eamxx_ml_correction_process_interface.cpp
+++ b/components/eamxx/src/physics/ml_correction/eamxx_ml_correction_process_interface.cpp
@@ -33,6 +33,8 @@ void MLCorrection::set_grids(
   // Layout for 3D (2d horiz X 1d vertical) variable defined at mid-level and
   // interfaces
   FieldLayout scalar3d_layout_mid{{COL, LEV}, {m_num_cols, m_num_levs}};
+  FieldLayout horiz_wind_layout { {COL,CMP,LEV}, {m_num_cols,2,m_num_levs} };
+
   
   /* ----------------------- WARNING --------------------------------*/
   /* The following is a HACK to get things moving, we don't want to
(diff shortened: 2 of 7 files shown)


components/eamxx/src/physics/ml_correction/ml_correction.py
```diff
@@ -8,23 +8,51 @@
...
+
+
+def ensure_correction_ordering(correction):
+    for key in correction:
+        correction[key] = correction[key].transpose("ncol", "z")
+    return correction
 
-def get_ML_correction(model, T_mid, qv, cos_zenith, dt):
+
+def get_ML_correction_dQ1_dQ2(model, T_mid, qv, cos_zenith, dt):
```
AaronDonahue, line 26: I wonder if this name could be changed to `dQT_dQqv` or something like that.  I noticed that T and qv are often referred to as q1 and q2 in our ML which is a bit difficult to understand to an outside observer.  But maybe q1 and q2 are a convention?

components/eamxx/cime_config/namelist_defaults_scream.xml
```diff
@@ -242,8 +242,10 @@ be lost if SCREAM_HACK_XML is not enabled.
 
     <!-- ML correction -->
     <mlcorrection inherit="atm_proc_base">
-      <ML_model_path type="string" doc="Path to pre-trained ML model"/>
+      <ML_model_path_tq type="string" doc="Path to pre-trained ML model for temperature and specific humidity"/>
```
AaronDonahue, line 245: Is there a reason we need to split the variables into two ML models.

components/eamxx/src/physics/ml_correction/ml_correction.py
elynnwu: `dQ1` and `dQ2` are the convention for heating and moistening, and they are heavily used throughout training and diagnostics. We can change it here though I find `dQT` and `dQqv` somewhat odd?

components/eamxx/cime_config/namelist_defaults_scream.xml
elynnwu: Yes, they are trained in separate models because the inputs are not the same. The team has in the past tried to train one model for Tquv but did not have good success. Having them separate also makes it easier to test Tq vs. Tquv. Later on, we will add another one for surface fluxes.

components/eamxx/src/physics/ml_correction/eamxx_ml_correction_process_interface.cpp
```diff
@@ -32,8 +34,17 @@ void MLCorrection::set_grids(
 
   // Layout for 3D (2d horiz X 1d vertical) variable defined at mid-level and
   // interfaces
+  FieldLayout scalar2d_layout{ {COL}, {m_num_cols}};
   FieldLayout scalar3d_layout_mid{{COL, LEV}, {m_num_cols, m_num_levs}};
-  
+  FieldLayout horiz_wind_layout { {COL,CMP,LEV}, {m_num_cols,2,m_num_levs} };
+  if (not m_ML_correction_unit_test) {
```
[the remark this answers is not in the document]
elynnwu: It is also used in the unit test input yaml. The unit test does not have access to surface geopotential or lat/lon, and since it adds `ml_correction` it goes through the init (not implementation) and crashes.

On second thought, do we have to register MLCorrection in the unit test?

components/eamxx/src/physics/ml_correction/ml_correction.py
AaronDonahue: Ahh, okay.  That makes sense then.  Could we add a comment here then explaining that dQ1 and dQ2 are the conventions?

components/eamxx/src/physics/ml_correction/eamxx_ml_correction_process_interface.cpp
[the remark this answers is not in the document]
elynnwu: Okay I thought about this more, we definitely want to register MLCorrection. I tried experimenting removing `m_ML_correction_unit_test` but kept running into different issues. Essentially, we have to be able to allocate all the fields MLCorrection needs during initialization, which includes variables such as `lat`, `phis`, and radiation fluxes (not this PR, but the next PR with ML surface fluxes). I can make it work for `lat` and `phis` by using actual scream IC and topography file, but it means slower unit test (which is probably okay). But, we will need a lot of dependencies on radiation in the next PR when we need fields like `SW_flux_dn`.

One obvious solution is to dynamically update pybind arguments based on runtime option, for example, only ask for surface fluxes and pass them if we intend to ML correct for those fields. The problem is that at the pybind11 interface, we cannot dynamically pass different number of variables. The only way would be to split it into 3 separate python calls (`Tq`, `Tquv`, `Tquv+surface fluxes`) but this will have a significant performance hit. 

So I'm inclined to keep `m_ML_correction_unit_test` flag and have a simple/fast unit test.

AaronDonahue (approved):

PR #2590, opened 2023-11-02
Changes needed for creating, allocating, and deepcpoying rank-0 fields

tcclevenger:
Two issues I've run into when using rank-0 fields in https://github.com/E3SM-Project/scream/pull/2587:
1. When running `Field::allocate_view()`, in `FieldAllocProp::commit()` we ask for `m_layout->dims().back()`, but no dims exist, so we error out. My solution here is to set `m_last_extent=0` (there is no "last dimension"), and `m_alloc_size=m_scalar_type_size` (field will contain a single value)
2. `Field::deep_copy()` (and `randomize(field,...)` which is needed in the test) need a `case 0:` implementation.

@bartgol These are the issues I ran into, maybe others exist.

### DIFF
--- a/components/eamxx/src/share/field/field_alloc_prop.cpp
+++ b/components/eamxx/src/share/field/field_alloc_prop.cpp
@@ -162,8 +162,12 @@ void FieldAllocProp::commit (const layout_type& layout)
       m_last_extent = std::max(m_last_extent, num_st);
     }
 
-    m_alloc_size = m_layout.size() / last_phys_extent  // All except the last dimension
-                 * m_last_extent * m_scalar_type_size; // Last dimension must account for padding (if any)
+    if (m_layout.size()>0) {
+      m_alloc_size = m_layout.size() / last_phys_extent  // All except the last dimension
+                   * m_last_extent * m_scalar_type_size; // Last dimension must account for padding (if any)
+    } else {
+      m_alloc_size = 0;
+    }
   }
 
   m_contiguous = true;
--- a/components/eamxx/src/share/field/field_alloc_prop.cpp
+++ b/components/eamxx/src/share/field/field_alloc_prop.cpp
@@ -37,9 +37,9 @@ subview (const int idim, const int k, const bool dynamic) const {
       "Error! Subview requires alloc properties to be committed.\n");
   EKAT_REQUIRE_MSG (idim==0 || idim==1,
       "Error! Subviewing is only allowed along first or second dimension.\n");
-  EKAT_REQUIRE_MSG (idim<m_layout->rank(),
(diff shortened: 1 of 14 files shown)


components/eamxx/src/share/field/field_alloc_prop.cpp
```diff
@@ -161,31 +161,33 @@ void FieldAllocProp::commit (const layout_ptr_type& layout)
...
-  m_last_extent = 0;
-  int last_phys_extent = m_layout->dims().back();
-  for (auto vts : m_value_type_sizes) {
-    // The number of scalar_type in a value_type
-    const int vt_len = vts / m_scalar_type_size;
+  if (m_layout->size()==0) {
+    // Zero-dimensional fields are supported, so we may end up with a
+    // layout of size 0. In this case allocate a single scalar, but
+    // set the last extent to 0 since we have no dimension.
+    m_alloc_size = m_scalar_type_size;
```
bartgol, line 168: What if the user (for some reason) asked for this field to accommodate packing? In that case `m_alloc_size = m_scalar_type_size` seems wrong...

OTOH, I don't think anyone will ever ask for a rank-0 field that is packed, so we could just error out in that case? So perhaps just add a check that `m_value_type_sizes` has size 1 (we always have `m_scalar_type_size` in it)?

bartgol: Actually, we _might_ hit that case. We can create a rank-0 subfield of a packed field. In that case, we may have to allow a non-trivial pack size...

bartgol:
I'm seeing a couple of things that need to change in `FieldAllocProp::subview`. I think we may have just gotten lucky so far...

bartgol (approved):

tcclevenger:
@bartgol I need to look at some errors (it looks like with SP). If you have the changes before I get those fixed then feel free to push here, if not a 2nd PR is fine. Doesn't matter to me. Thanks for looking at it!

components/eamxx/src/share/field/field_alloc_prop.cpp
tcclevenger: Yeah, padding is probably an issue. And we probably don't just want to error out in case we are calling `get_padding()` on multiple fields, not just ones we think will have padding. Could be annoying.

Something I was thinking about as well is subviewing an already subviewed 1d field (or any case where we aren't guaranteed contiguous data). Maybe that could be added to the tests if we think it works (or maybe I'm overthinking).

tcclevenger:
@bartgol We were returning `1` for FieldLayout size for rank 0 fields (I forgot to check in the change to return `0`).

jgfouca (approved):

PR #2591, opened 2023-11-04
Increase resolution of rain fall speed lookup tables.

quantheory:
This commit increases the resolution and range of the rain fall speed lookup tables, since the previous resolution caused significant artifacts in the vertical profile of rain. The previous tables used 10 micron resolution for small drop sizes and 30 micron resolution for larger drop sizes, covering a range between 5 and 8595 microns. The new tables have a uniform 1 micron resolution between 1 and 9000 microns, which has proven sufficient to remove the numerical artifacts in simple offline tests.

Since linear interpolation is used to extract fall speeds from the lookup tables, and since the fall speed has negative second derivative with respect to diameter over almost all of its range, this commit systematically increases rain fall speed globally. However, the climate impact of this effect is not expected to be large, since it will generally only result in a small increase in rain evaporation.

This commit will require regeneration of P3 baselines, as well as copying or regenerating the affected lookup tables to the input data directory on all machines.

[CC]

### DIFF
--- a/components/eam/src/physics/p3/scream/micro_p3.F90
+++ b/components/eam/src/physics/p3/scream/micro_p3.F90
@@ -80,7 +80,7 @@ module micro_p3
   real(rtype), protected, dimension(150) :: mu_r_table_vals
 
   ! lookup table values for rain number- and mass-weighted fallspeeds and ventilation parameters
-  real(rtype), protected, dimension(300,10) :: vn_table_vals,vm_table_vals,revap_table_vals
+  real(rtype), protected, dimension(9000,10) :: vn_table_vals,vm_table_vals,revap_table_vals
 
   type realptr
      real(rtype), dimension(:), pointer :: p
@@ -191,7 +191,7 @@ contains
     ! This can be called after p3_init_b.
     implicit none
     real(rtype), dimension(150), intent(out) :: mu_r_user
-    real(rtype), dimension(300,10), intent(out) :: vn_user, vm_user, revap_user
+    real(rtype), dimension(9000,10), intent(out) :: vn_user, vm_user, revap_user
     mu_r_user(:) = mu_r_table_vals(:)
     revap_user(:,:) = revap_table_vals(:,:)
     vn_user(:,:) = vn_table_vals(:,:)
@@ -205,7 +205,7 @@ contains
     ! This can be called instead of p3_init_b.
     implicit none
     real(rtype), dimension(150), intent(in) :: mu_r_user
-    real(rtype), dimension(300,10), intent(in) :: vn_user, vm_user, revap_user
+    real(rtype), dimension(9000,10), intent(in) :: vn_user, vm_user, revap_user
     mu_r_table_vals(:) = mu_r_user(:)
     revap_table_vals(:,:) = revap_user(:,:)
     vn_table_vals(:,:) = vn_user(:,:)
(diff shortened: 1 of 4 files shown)


quantheory:
Had some tests complete on this while I was out today, and figured I'd throw the pull request out here for review. Note that this will require both regenerating baselines and providing the new lookup tables by running `p3_tables_setup` (and/or just uploading the new files to the input data repository at ANL). My guess is that the larger lookup tables are still not large enough to cause a noticeable slowdown (<1MB), though I don't have hard data on this for GPU systems.

I distinguished the new lookup tables from the old ones by adding `_res1micron` to the file name.

For context regarding the problem being solved here, here's a plot of rain mixing ratio in a simple rainshaft model we've been using for testing purposes.

image

The jagged solid grey line is the vertical profile when using the current code (with a 50 second P3 time step), and the dotted blue line is the result with a 1 micron resolution lookup table. (The other lines mostly involve changes to the time stepping and are not so relevant for this PR.) The vertical grid here is the same as that used in EAM. So you can see that the lookup table resolution has a pretty noticeable effect on both the smoothness and mean value of the solution, and using a finer resolution lookup table gets rid of a lot of numerical noise.

PeterCaldwell:
Thanks Sean. It would be great to get this integrated in time for the decadal runs. Sean - what would the effect of increased vertical resolution be on your results? Scream has much finer resolution than EAM.

quantheory:
@PeterCaldwell Vertical resolution doesn't matter much. I think that the figure I linked earlier was from a run from several months ago which had a couple of minor bugs and some different initial/boundary conditions from what we eventually settled on for our "standard" case.

So I reran it just now, with the updated/fixed code, and this time I used SCREAM resolutions (pressure levels taken from the 128 level vertical grid and a 100 second time step). Here's the results after allowing it to run for half an hour (which is close to the "steady state":

qr_profile_screamlevs

quantheory:
I marked this WIP just to stop the autotester spam. Until the input data files for the changed lookup tables are made available, it's just going to keep failing.

ambrad:
Sean, is it possible the wild behavior of the original is due to a bug in the original table rather than its coarser resolution? To me, it seems the correct curves are robustly monotonic, while the original curves are not. That then makes me wonder if the table is supposed to have some property that implies this monotonicity, and the original table simply lacked that property.

quantheory:
@ambrad I'm very confident in the table values because we have calculated the fall speeds two completely different ways (with a very simple numerical integration and with an analytical formula using incomplete gamma functions) and gotten the same results both ways. I doubt there is an issue with the lookup code either, but I can come up with a plot of the fall speeds coming out of the interpolation to be sure.

My understanding of the issue is probably not complete, but from writing down the microphysical equations for the steady state, I'm able to get them in a form where everything should be continuous with height, except there's one term proportional to the derivative of fall speed with height, which is simply the derivative of fall speed with respect to diameter multiplied by derivative of diameter with respect to height. The derivative of the fall speed with respect to diameter will not be continuous, because it comes from a table lookup involving simple linear interpolation, and that necessarily produces "jumps" in the derivative at each table value.

Increasing the resolution of the table by an order of magnitude does not get rid of these jump discontinuities in the derivative, but it does reduce the magnitude of those discontinuities to the point where the introduced noise appears relatively small (particularly given that there's a bit of smoothing from numerical diffusion at this vertical resolution).

I'm considering changing the interpolation from linear to cubic monotone to try to get rid of the discontinuities as much as possible, but this was the easily-implemented, low-risk solution.

quantheory:
@ambrad I am very glad that you prompted me to double-check the lookup table code, because it turns out that there was a copy-paste typo in the version of the code used to generate those plots, which ended up adding an additional error proportional to the table width (basically, the linear interpolated lines were too steep and "overshooting" the point where they should connect, producing a bit of a sawtooth noise overlaying the fall speeds). This bug does not appear to be in the existing P3 code, just my copy.

I feel like there's an important lesson here; I've just started a new test suite for this P3 driver and gotten up to about 75% unit test coverage on this code, but I apparently need to bump that up to ~100% before I start trusting it enough to push PRs.

@PeterCaldwell Apologies! I think this was largely a false alarm. This PR would have been harmless I think, but I don't think it's needed after all. With the aforementioned bug fixed, the two curves line up right on top of each other:

qr_profile_screamlevs

Issue #1722, opened 2022-06-05
P3 performance analysis


ambrad:
This issue documents some basic findings about P3's performance on the CPU and suggests action items for future performance work.

I was curious what the primary cost in the C++ P3 code is. It turns out to be 
1. https://github.com/E3SM-Project/scream/blob/46ff6b3cdabd0b8e86d1e05ce89f63f5e51ec53b/components/scream/src/physics/p3/p3_rain_sed_impl.hpp#L111
2. the equivalent in ice sedimentation.

In particular, while one might guess that the upwind impl could be slow, it is not: calc_first_order_upwind_step is < 4% of the total P3 cost. In contrast, the rain and ice fall velocity calculations are very roughly 80%.

Possible action items:
1. Profile using an Intel tool at the line level, starting with rain sedimentation. (1) Are there a few costly lines, e.g., a slow tgamma impl, or instead (2) is the cost per line fairly uniform over the whole velocity computation?
2. If 2, then try a few different modifications to the Mask implementation: different integer sizes for the mask slots; different implementations (e.g. ternary op vs `if`) for the masked ops.
3. If there is no big change, profile with pack size 1 to see if that reveals anything.
4. Try a pack-free impl, using `scalarize` to produce 1D views of reals as inputs. This is a mask-intensive region of code, and the C++ compiler might not be able to handle it well.

PeterCaldwell:
@hr203 - this might be a good topic for you!

PR #2592, opened 2023-11-06
Speed up `get_rain_dsd2` calculations.

### LINKED ISSUE #2565: Inefficient coding in P3
So, @ambrad pointed me to #1722 the other day, and I realized that a lot of the cost of the P3 rain sedimentation is probably in the many `get_rain_dsd2` calls in the rain fall velocity calculations (which is a prerequisite for the lookup table stuff). There's a lot of inefficient stuff in there!

First, note these lines:
https://github.com/E3SM-Project/scream/blob/f412870ca05da69fa26698aa22bb47b4fc24ef20/components/eamxx/src/physics/p3/impl/p3_dsd2_impl.hpp#L100-L107
In this case, `inv_dum `doesn't actually seem to be used in the current version, so I'm hoping that the compiler is able to optimize out its calculation completely. But also, since `mu_r` is currently treated as a constant, `inv_dum` is just a constant divided by `lamr`. So calculating the cube root twice would be unnecessary anyway if `cbrt((mu_r + 3) * (mu_r + 2) * (mu_r + 1) / 6.)` was calculated once at initialization, and the value saved for all further calculations.

Second, just below that:
https://github.com/E3SM-Project/scream/blob/f412870ca05da69fa26698aa22bb47b4fc24ef20/components/eamxx/src/physics/p3/impl/p3_dsd2_impl.hpp#L123-L127
This looks like it is just a translation from the Fortran version of P3, but the original P3 code itself is very weird. This code is calculating the mathematical equivalent of:
```c++
nr[s] = lamr[s]*lamr[s]*lamr[s] * qr[s] / ((mu_r+1) * (mu_r+2) * (mu_r+3) * cons1);
```
That's it; none of the `exp`, `log`, or `tgamma` calls are necessary here. Also, you can see that the denominator is an expression that has already been used above, so if you really wanted to optimize, that could be saved and reused.

Third, right below that, you have:
https://github.com/E3SM-Project/scream/blob/f412870ca05da69fa26698aa22bb47b4fc24ef20/components/eamxx/src/physics/p3/impl/p3_dsd2_impl.hpp#L130-L132
If `mu_r` is a constant, then `tgamma(mu_r + 1.)` can be calculated at initialization and saved, thus removing `tgamma` from this performance-critical code. Furthermore, the second line can be reduced to:
```c++
logn0r.set(qr_gt_small, log10(cdistr) + (mu_r + 1) * log10(lamr));
```
This reduces the number of `log10` calls by one. Even better, the calculation of these variables could be split out into a separate function from `get_rain_dsd2` entirely, since I do not believe `cdistr` or `logn0r` are needed in the sedimentation loop at all.

I am **not** planning to make any of these changes myself, at least right now. However, since the above optimizations should collectively remove *all* of the transcendental function calls from the performance-critical sedimentation loop (leaving only a single cube root and a few arithmetic operations), it might be worth it for someone to try to implement these changes, and see if there are any similar improvements are to be found in the ice code.

(Of course these optimizations also are, or will be, in the modified P3 code we're developing as part of our SciDAC project.)

quantheory:
Fixes #2565. This commit makes a set of changes to the Fortran and C++ versions of P3 used by SCREAM, all of which are intended to avoid redundant calculations and/or minimize special function evaluations. These changes are limited to `get_rain_dsd2` and its callers, and made due to the belief that `get_rain_dsd2` is relatively performance- critical code, since it is evaluated at each rain sedimentation time step (often much smaller than the overall physics time step).

The changes are:
 1. Remove calculation of `inv_dum`, since it is not used.
 2. Simplify the conversions between `lamr` and `nr`. In particular, all special function calls can be removed from the calculation of the "adjusted" `nr` produced when `lamr` is limited.
 3. Remove extra `log10` and `tgamma` calls when calculating `logn0r`, since it can be calculated more simply from the value of `cdistr` on the line above.

The C++ version also has an additional change, which is separating the calculation of `cdistr` and `logn0r` into a separate function from `get_rain_dsd2`. This is done because these two values require a `tgamma` call and two `log10` calls to calculate, but they are only needed in a few places in the code. Thus it is better to only calculate them in the one place where they are actually needed, rather than in every `get_rain_dsd2` call. In particular, they are not needed in the rain sedimentation loop.

These changes cause roundoff-level changes in P3.

[non-BFB]

### DIFF
--- a/components/eam/src/physics/p3/scream/micro_p3.F90
+++ b/components/eam/src/physics/p3/scream/micro_p3.F90
@@ -1846,7 +1846,8 @@ contains
     real(rtype),     intent(out)           :: lamr,mu_r,cdistr,logn0r
 
     !local variables:
-    real(rtype)                            :: inv_dum,lammax,lammin
+    real(rtype)                            :: lammax,lammin
+    real(rtype)                            :: mass_to_d3_factor
 
     !--------------------------------------------------------------------------
 
@@ -1858,25 +1859,25 @@ contains
        ! find spot in lookup table
        ! (scaled N/q for lookup table parameter space_
        nr      = max(nr,nsmall)
-       inv_dum = bfb_cbrt(qr/(cons1*nr*6._rtype))
 
        ! Apply constant mu_r:  Recall the switch to v4 tables means constant mu_r
        mu_r = mu_r_constant
-       lamr   = bfb_cbrt(cons1*nr*(mu_r+3._rtype)*(mu_r+2._rtype)*(mu_r+1._rtype)/(qr))  ! recalculate slope based on mu_r
+       mass_to_d3_factor = cons1*(mu_r+3._rtype)*(mu_r+2._rtype)*(mu_r+1._rtype)
+       lamr   = bfb_cbrt(mass_to_d3_factor*nr/qr)  ! recalculate slope based on mu_r
        lammax = (mu_r+1._rtype)*1.e+5_rtype   ! check for slope
        lammin = (mu_r+1._rtype)*500._rtype  !500=1/(2mm) is inverse of max allowed number-weighted mean raindrop diameter
        
        ! apply lambda limiters for rain
        if (lamr.lt.lammin) then
           lamr = lammin
-          nr   = bfb_exp(3._rtype*bfb_log(lamr)+bfb_log(qr)+bfb_log(bfb_gamma(mu_r+1._rtype))-bfb_log(bfb_gamma(mu_r+4._rtype)))/(cons1)
+          nr   = lamr * lamr * lamr * qr / mass_to_d3_factor
        elseif (lamr.gt.lammax) then
           lamr = lammax
-          nr   = bfb_exp(3._rtype*bfb_log(lamr)+bfb_log(qr)+bfb_log(bfb_gamma(mu_r+1._rtype))-bfb_log(bfb_gamma(mu_r+4._rtype)))/(cons1)
+          nr   = lamr * lamr * lamr * qr / mass_to_d3_factor
        endif
 
        cdistr  = nr/bfb_gamma(mu_r+1._rtype)
-       logn0r  = bfb_log10(nr)+(mu_r+1._rtype)*bfb_log10(lamr)-bfb_log10(bfb_gamma(mu_r+1._rtype)) !note: logn0r is calculated as log10(n0r)
+       logn0r  = bfb_log10(cdistr)+(mu_r+1._rtype)*bfb_log10(lamr) !note: logn0r is calculated as log10(n0r)
 
     else
 
--- a/components/eamxx/src/physics/p3/impl/p3_dsd2_impl.hpp
+++ b/components/eamxx/src/physics/p3/impl/p3_dsd2_impl.hpp
@@ -76,7 +76,7 @@ KOKKOS_FUNCTION
 void Functions<S,D>::
 get_rain_dsd2 (
   const Spack& qr, Spack& nr, Spack& mu_r,
-  Spack& lamr, Spack& cdistr, Spack& logn0r, 
+  Spack& lamr,
   const physics::P3_Constants<S> & p3constants,
   const Smask& context)
 {
@@ -85,8 +85,6 @@ get_rain_dsd2 (
   constexpr auto cons1  = C::CONS1;
 
   lamr.set(context  , 0);
-  cdistr.set(context, 0);
-  logn0r.set(context, 0);
 
   const auto qr_gt_small = qr >= qsmall && context;
 
@@ -99,14 +97,12 @@ get_rain_dsd2 (
     // find spot in lookup table
     // (scaled N/q for lookup table parameter space)
     const auto nr_lim = max(nr, nsmall);
-    Spack inv_dum(0);
-    inv_dum.set(qr_gt_small, cbrt(qr / (cons1 * nr_lim * 6)));
 
     // Apply constant mu_r:  Recall the switch to v4 tables means constant mu_r
     mu_r.set(qr_gt_small, mu_r_const);
     // recalculate slope based on mu_r
-    lamr.set(qr_gt_small, cbrt(cons1 * nr_lim * (mu_r + 3) *
-                                     (mu_r + 2) * (mu_r + 1)/qr));
+    const auto mass_to_d3_factor = cons1 * (mu_r + 3) * (mu_r + 2) * (mu_r + 1);
+    lamr.set(qr_gt_small, cbrt(mass_to_d3_factor * nr_lim / qr));
 
     // check for slope
     const auto lammax = (mu_r+1.)*sp(1.e+5);
@@ -123,15 +119,31 @@ get_rain_dsd2 (
       lamr.set(lt, lammin);
       lamr.set(gt, lammax);
       ekat_masked_loop(either, s) {
-        nr[s] = std::exp(3*std::log(lamr[s]) + std::log(qr[s]) +
-                         std::log(std::tgamma(mu_r[s] + 1)) - std::log(std::tgamma(mu_r[s] + 4)))
-          / cons1;
+        nr[s] = lamr[s]*lamr[s]*lamr[s] * qr[s] / mass_to_d3_factor[s];
       }
     }
+  }
+}
 
+template <typename S, typename D>
+KOKKOS_FUNCTION
+void Functions<S,D>::
+get_cdistr_logn0r (
+  const Spack& qr, const Spack& nr, const Spack& mu_r,
+  const Spack& lamr, Spack& cdistr, Spack& logn0r,
+  const Smask& context)
+{
+  constexpr auto qsmall = C::QSMALL;
+
+  cdistr.set(context, 0);
+  logn0r.set(context, 0);
+
+  const auto qr_gt_small = qr >= qsmall && context;
+
+  if (qr_gt_small.any()) {
     cdistr.set(qr_gt_small, nr/tgamma(mu_r + 1));
     // note: logn0r is calculated as log10(n0r)
-    logn0r.set(qr_gt_small, log10(nr) + (mu_r + 1) * log10(lamr) - log10(tgamma(mu_r+1)));
+    logn0r.set(qr_gt_small, log10(cdistr) + (mu_r + 1) * log10(lamr));
   }
 }
 
--- a/components/eamxx/src/physics/p3/impl/p3_main_impl_part2.hpp
+++ b/components/eamxx/src/physics/p3/impl/p3_main_impl_part2.hpp
@@ -216,7 +216,8 @@ void Functions<S,D>
                      lamc(k), cdist(k), cdist1(k), not_skip_micro);
       nc(k).set(not_skip_micro, nc_incld(k) * cld_frac_l(k));
 
-      get_rain_dsd2(qr_incld(k), nr_incld(k), mu_r(k), lamr(k), cdistr(k), logn0r(k), p3constants, not_skip_micro);
+      get_rain_dsd2(qr_incld(k), nr_incld(k), mu_r(k), lamr(k), p3constants, not_skip_micro);
+      get_cdistr_logn0r(qr_incld(k), nr_incld(k), mu_r(k), lamr(k), cdistr(k), logn0r(k), not_skip_micro);
       nr(k).set(not_skip_micro, nr_incld(k) * cld_frac_r(k));
 
       impose_max_total_ni(ni_incld(k), max_total_ni, inv_rho(k), not_skip_micro);
--- a/components/eamxx/src/physics/p3/impl/p3_main_impl_part3.hpp
+++ b/components/eamxx/src/physics/p3/impl/p3_main_impl_part3.hpp
@@ -108,7 +108,7 @@ void Functions<S,D>
       auto nr_incld = nr(k)/cld_frac_r(k); //nr_incld is updated in get_rain_dsd2 but isn't used again
 
       get_rain_dsd2(
-        qr_incld, nr_incld, mu_r(k), lamr(k), ignore1, ignore2, p3constants, qr_gt_small);
+        qr_incld, nr_incld, mu_r(k), lamr(k), p3constants, qr_gt_small);
 
       //Note that integrating over the drop-size PDF as done here should only be done to in-cloud
       //quantities but radar reflectivity is likely meant to be a cell ave. Thus nr in the next line
--- a/components/eamxx/src/physics/p3/impl/p3_rain_sed_impl.hpp
+++ b/components/eamxx/src/physics/p3/impl/p3_rain_sed_impl.hpp
@@ -22,8 +22,7 @@ void Functions<S,D>
   const Smask& context)
 {
   Table3 table;
-  Spack tmp1, tmp2; //ignore
-  get_rain_dsd2(qr_incld, nr_incld, mu_r, lamr, tmp1, tmp2, p3constants, context);
+  get_rain_dsd2(qr_incld, nr_incld, mu_r, lamr, p3constants, context);
 
   if (context.any()) {
     lookup(mu_r, lamr, table, context);
--- a/components/eamxx/src/physics/p3/p3_functions.hpp
+++ b/components/eamxx/src/physics/p3/p3_functions.hpp
@@ -625,8 +625,15 @@ struct Functions
   KOKKOS_FUNCTION
   static void get_rain_dsd2 (
     const Spack& qr, Spack& nr, Spack& mu_r,
-    Spack& lamr, Spack& cdistr, Spack& logn0r,
+    Spack& lamr,
     const physics::P3_Constants<ScalarT> & p3constants,
+    const Smask& context = Smask(true) );
+
+  // Computes and returns additional rain size distribution parameters
+  KOKKOS_FUNCTION
+  static void get_cdistr_logn0r (
+    const Spack& qr, const Spack& nr, const Spack& mu_r,
+    const Spack& lamr, Spack& cdistr, Spack& logn0r,
     const Smask& context = Smask(true) );
 
   // Calculates rime density
--- a/components/eamxx/src/physics/p3/tests/p3_dsd2_unit_tests.cpp
+++ b/components/eamxx/src/physics/p3/tests/p3_dsd2_unit_tests.cpp
@@ -161,7 +161,8 @@ struct UnitWrap::UnitTest<D>::TestDsd2 {
       }
 
       Spack mu_r(0.0), lamr(0.0), cdistr(0.0), logn0r(0.0);
-      Functions::get_rain_dsd2(qr, nr, mu_r, lamr, cdistr, logn0r, physics::P3_Constants<Real>());
+      Functions::get_rain_dsd2(qr, nr, mu_r, lamr, physics::P3_Constants<Real>());
+      Functions::get_cdistr_logn0r(qr, nr, mu_r, lamr, cdistr, logn0r);
 
       // Copy results back into views
       for (Int s = 0, vs = offset; s < Spack::n; ++s, ++vs) {


quantheory:
Thought I'd make up for the confusion with my other PR this weekend by throwing this PR together. These changes appear to be roundoff as intended according to the test suite output (e.g. `nr` relative differences less than $10^{-14}$, signifying absolute differences generally well below $10^{-8}$). I'm confident that this code is faster, but not sure by how much in practice.

bartgol:
@quantheory You can try to run `p3_run_and_cmp`. I don't recall exactly its options (running with `-h` should tell you), but it should allow you to run the whole p3_main a bunch of times, and report the timing. If you do that for master and the branch, you should get a sense of the speedup of p3_main itself. I don't think we have a way to get speedups of a single p3 function.

quantheory:
@bartgol That doesn't seem to work too well. If I run this code vs. master, there's no timing difference (on my desktop's CPU, can run both in about 4.1 seconds). However, if I comment out rain sedimentation entirely (which should speed things up a lot), there's *still* no timing difference, which is inconsistent with #1722. I think that in that test driver, the P3 runs are either running in a regime that's not representative of what happens in the global model, or because the runs are so short, the overhead from I/O and other operations in `p3_run_and_cmp` swamp the cost of P3 itself. Or both.

quantheory:
Oops, it is in Debug mode, so that's probably it. Lemme try again.

quantheory:
OK, I ran again with:

```./src/physics/p3/tests/p3_run_and_cmp -r 2 -s 100```

The output for the current master:

```
Generating to 8
Running P3 with ni=3, nk=72, dt=300, ts=100, predict_nc=0, prescribed_CCN=0, small_packn=16
Time = 3.392e-02 seconds
Running P3 with ni=3, nk=72, dt=300, ts=100, predict_nc=0, prescribed_CCN=1, small_packn=16
Time = 6.948e-02 seconds
Running P3 with ni=3, nk=72, dt=300, ts=100, predict_nc=1, prescribed_CCN=0, small_packn=16
Time = 9.820e-02 seconds
Running P3 with ni=3, nk=72, dt=300, ts=100, predict_nc=1, prescribed_CCN=1, small_packn=16
Time = 1.318e-01 seconds
```

With all sedimentation turned off (cloud, rain and ice):

```
Generating to 8
Running P3 with ni=3, nk=72, dt=300, ts=100, predict_nc=0, prescribed_CCN=0, small_packn=16
Time = 2.403e-02 seconds
Running P3 with ni=3, nk=72, dt=300, ts=100, predict_nc=0, prescribed_CCN=1, small_packn=16
Time = 4.556e-02 seconds
Running P3 with ni=3, nk=72, dt=300, ts=100, predict_nc=1, prescribed_CCN=0, small_packn=16
Time = 6.432e-02 seconds
Running P3 with ni=3, nk=72, dt=300, ts=100, predict_nc=1, prescribed_CCN=1, small_packn=16
Time = 8.639e-02 seconds
```

With these changes:

```
Generating to 8
Running P3 with ni=3, nk=72, dt=300, ts=100, predict_nc=0, prescribed_CCN=0, small_packn=16
Time = 2.756e-02 seconds
Running P3 with ni=3, nk=72, dt=300, ts=100, predict_nc=0, prescribed_CCN=1, small_packn=16
Time = 5.508e-02 seconds
Running P3 with ni=3, nk=72, dt=300, ts=100, predict_nc=1, prescribed_CCN=0, small_packn=16
Time = 7.876e-02 seconds
Running P3 with ni=3, nk=72, dt=300, ts=100, predict_nc=1, prescribed_CCN=1, small_packn=16
Time = 1.074e-01 seconds
```

To summarize, in these tests, sedimentation of all variables is about a third of the total cost, which is honestly still much less than I expected, since IIRC @ambrad found that well over half the total cost was in the sedimentation. Maybe this has to do with the different vertical grid in this simple driver (the driver's dt is also 300, but if anything that should increase the proportion of the cost from the sedimentation). Of that cost, these changes reduce the total cost of the sedimentation by over half.

**Review by bartgol — APPROVED** [review]

**Comment by ambrad** [comment]
> well over half the total cost was in the sedimentation

For context, this statement refers to https://github.com/E3SM-Project/scream/issues/1722. I didn't record the specific runs I did there, but it's possible I ran with dt much greater than 300, which I think would make sedimentation relatively more important than the water phase processes. It's also likely that I ran on a platform where 512-bit vectorization matters a lot, likely amplifying the inefficiency of the fall-speed calculations relative to the rest of the code.

Those details aside, I read your results as suggesting you get an overall ~23% speedup on your laptop, which is great.

**Comment by bartgol** [comment]
The diffs in V1 cases are quite substantial. Most fields have diffs in all digits in the Ln9 tests. It seems a bit extreme that all fields go out of whack that quickly. And looking at p3_run_and_cmp output suggests something might indeed be amiss:

```
--- checking case # 3, timestep # 5 of 6 ---
qv > tol 5 times. Max rel diff= 6.98492e-09 normalized by ref impl val=0.05.
th_atm > tol 1 times. Max rel diff= 7.06724e-07 normalized by ref impl val=1079.54.
qc > tol 26 times. Max rel diff= 9.81562e-07 normalized by ref impl val=1.11189e-05.
nc > tol 26 times. Max rel diff= 5.52599e-07 normalized by ref impl val=650336.
qr > tol 29 times. Max rel diff= 1.05796e-05 normalized by ref impl val=0.000349196.
nr > tol 31 times. Max rel diff= 4.61097e-08 normalized by ref impl val=254149.
qi > tol 10 times. Max rel diff= 0.000763301 normalized by ref impl val=0.000813633.
ni > tol 10 times. Max rel diff= 0.000224525 normalized by ref impl val=465567.
qm > tol 10 times. Max rel diff= 0.000801534 normalized by ref impl val=0.000772498.
bm > tol 10 times. Max rel diff= 0.000801527 normalized by ref impl val=8.58332e-07.
precip_liq_surf > tol 3 times. Max rel diff= 0.000352202 normalized by ref impl val=2.97999e-06.
diag_eff_radius_qc > tol 11 times. Max rel diff= 2.37815e-07 normalized by ref impl val=2.29463e-05.
diag_eff_radius_qi > tol 10 times. Max rel diff= 0.00160528 normalized by ref impl val=0.002923.
diag_eff_radius_qr > tol 28 times. Max rel diff= 0.000188127 normalized by ref impl val=0.0005.
rho_qi > tol 9 times. Max rel diff= 2.55669e-05 normalized by ref impl val=900.
precip_liq_flux > tol 30 times. Max rel diff= 0.00035605 normalized by ref impl val=0.068168.
liq_ice_exchange > tol 1 times. Max rel diff= 0.00170493 normalized by ref impl val=4.37002e-06.
vap_liq_exchange > tol 8 times. Max rel diff= 1.0671e-06 normalized by ref impl val=1.385e-06.
Ref impl failed.
```
This was the SP build. The DP build has relative errors close or below eps, but it has _a lot_ of warnings that look like the following
```
** WARNING IN P3_MAIN -- src, gcol, lon, lat, lvl, tstep, T:  900,        1,   1.0000,   1.0000,    1,        6,   0.14999470E+03
** WARNING IN P3_MAIN -- src, gcol, lon, lat, lvl, tstep, T:  900,        1,   1.0000,   1.0000,    2,        6,   0.15208006E+03
** WARNING IN P3_MAIN -- src, gcol, lon, lat, lvl, tstep, T:  900,        1,   1.0000,   1.0000,    3,        6,   0.15416387E+03
** WARNING IN P3_MAIN -- src, gcol, lon, lat, lvl, tstep, T:  900,        1,   1.0000,   1.0000,    4,        6,   0.15624748E+03
** WARNING IN P3_MAIN -- src, gcol, lon, lat, lvl, tstep, T:  900,        1,   1.0000,   1.0000,    5,        6,   0.15833101E+03
** WARNING IN P3_MAIN -- src, gcol, lon, lat, lvl, tstep, Qv  900,        1,   1.0000,   1.0000,   72,        6,   0.50000000E-01
...
```
I'm not sure if they are related. We may as well get these in master too, our testing framework won't print any test output unless the test fails...

**Comment by quantheory** [comment]
@bartgol Those warnings appear in master as well. Just doing a `make baseline` using master (which also runs `p3_run_and_cmp`), I get dozens of these as well:

```
** WARNING IN P3_MAIN -- src, gcol, lon, lat, lvl, tstep, Qv  900,        2,   2.0000,   2.0000,   72,        6,   0.50000000E-01
** WARNING IN P3_MAIN -- src, gcol, lon, lat, lvl, tstep, T:  900,        3,   3.0000,   3.0000,    1,        6,   0.14999470E+03
** WARNING IN P3_MAIN -- src, gcol, lon, lat, lvl, tstep, T:  900,        3,   3.0000,   3.0000,    2,        6,   0.15210089E+03
** WARNING IN P3_MAIN -- src, gcol, lon, lat, lvl, tstep, T:  900,        3,   3.0000,   3.0000,    3,        6,   0.15420553E+03
** WARNING IN P3_MAIN -- src, gcol, lon, lat, lvl, tstep, T:  900,        3,   3.0000,   3.0000,    4,        6,   0.15630998E+03
** WARNING IN P3_MAIN -- src, gcol, lon, lat, lvl, tstep, T:  900,        3,   3.0000,   3.0000,    5,        6,   0.15841434E+03
** WARNING IN P3_MAIN -- src, gcol, lon, lat, lvl, tstep, Qv  900,        3,   3.0000,   3.0000,   72,        6,   0.50000000E-01
```

Those seem like very cold temperatures (150K) at the top of the model and a very high specific humidity (50 g/kg) at the lowest level, which suggests that this test case setup is unrealistic, probably in how the initial and boundary conditions are set up. But that's a preexisting issue unrelated to this PR.

Regarding the other issues, I don't actually think this is a problem. The tiny amount of change when running `p3_run_and_cmp` in double precision suggests that the changes really are affecting answers only due to limited machine precision. Even in single precision, I only get large changes for that particular case 3 (I think that's prognostic `nc` and `CCN`?). Probably the most "extreme" change that affects answers is this diff:

```diff
-        nr[s] = std::exp(3*std::log(lamr[s]) + std::log(qr[s]) +
-                         std::log(std::tgamma(mu_r[s] + 1)) - std::log(std::tgamma(mu_r[s] + 4)))
-          / cons1;
+        nr[s] = lamr[s]*lamr[s]*lamr[s] * qr[s] / mass_to_d3_factor[s];
```
I would not be surprised if the evaluation of the old code was much less accurate in single precision than the revised code, simply due to the inaccuracy produced by the seven special function calls (five if the `tgamma` implementation is smart enough to evaluate integer arguments exactly).

But I also suspected that P3 is just extremely sensitive to the `nr` value here when running in single precision, so I tested that. Specifically, I generated a baseline using current master with single precision, and then I added the following line directly after the `nr` limiter code quoted above:

```c++
nr[s] = nr[s] * sp(1 + std::numeric_limits<float>::epsilon());
```

With just this machine epsilon change, I get the following output for the same case 3 time step you quoted from `p3_run_and_cmp`:

```text
qv > tol 18 times. Max rel diff= 0.00285448 normalized by ref impl val=0.05.
th_atm > tol 26 times. Max rel diff= 0.000334478 normalized by ref impl val=1079.54.
qc > tol 34 times. Max rel diff= 0.0730696 normalized by ref impl val=1.17271e-05.
nc > tol 34 times. Max rel diff= 0.0458858 normalized by ref impl val=717636.
qr > tol 39 times. Max rel diff= 0.0245211 normalized by ref impl val=0.000343321.
nr > tol 39 times. Max rel diff= 0.001393 normalized by ref impl val=245398.
qi > tol 40 times. Max rel diff= 0.460071 normalized by ref impl val=0.000973015.
ni > tol 40 times. Max rel diff= 0.0695089 normalized by ref impl val=475081.
qm > tol 40 times. Max rel diff= 0.469813 normalized by ref impl val=0.000973015.
bm > tol 40 times. Max rel diff= 0.469813 normalized by ref impl val=1.08113e-06.
precip_liq_surf > tol 3 times. Max rel diff= 0.00132982 normalized by ref impl val=2.35303e-06.
precip_ice_surf > tol 2 times. Max rel diff= 3.20736e+06 normalized by ref impl val=5.49395e-16.
diag_eff_radius_qc > tol 30 times. Max rel diff= 0.00767626 normalized by ref impl val=2.29463e-05.
diag_eff_radius_qi > tol 40 times. Max rel diff= 11.7658 normalized by ref impl val=0.000162998.
diag_eff_radius_qr > tol 38 times. Max rel diff= 0.0207639 normalized by ref impl val=0.00054714.
rho_qi > tol 40 times. Max rel diff= 0.985317 normalized by ref impl val=900.
qv2qi_depos_tend > tol 5 times. Max rel diff= 0.00248039 normalized by ref impl val=7.24513e-07.
precip_liq_flux > tol 39 times. Max rel diff= 0.00304799 normalized by ref impl val=0.0537522.
liq_ice_exchange > tol 25 times. Max rel diff= 0.005667 normalized by ref impl val=4.33695e-06.
vap_liq_exchange > tol 4 times. Max rel diff= 0.00192103 normalized by ref impl val=1.38499e-06.
vap_ice_exchange > tol 5 times. Max rel diff= 0.00248039 normalized by ref impl val=7.24513e-07.
Ref impl failed.
```

So I don't think there's an issue with this PR. It's just that P3 is so sensitive to roundoff level changes when run at single precision, that within a few time steps a machine epsilon sized perturbation leads to pretty significant changes.

bartgol:
@quantheory thanks for the in depth analysis. I'm a bit puzzled that run_and_cmp has very tiny errors on dp, and yet we get O(1) diffs in v1 cases after 9 steps in most fields. Is this something you would expect? I guess we can try to run a single step, and see how this PR compares to master.

quantheory:
No, I don't really understand why this perturbation causes simulations to diverge so quickly from the baseline, even with double precision. I am wondering how many columns would contain large deviations after only one or two time steps. I.e. does every column experience this, or is it more localized.

bartgol:
I ran `SMS.ne4pg2_ne4pg2.F2010-SCREAMv1` with both master and the branch. At step 0 and 1, they are bfb, but at step 2 there are some *wild* differences. There are a lot of diffs, but P3 vars like `nr` and `nc` immediately caught my eye (perhaps b/c they are not usually small numbers). E.g,  I saw for `nr` the line
```
    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, -2.985072, -18.8812, -40.69811, -27.98344, -26.21632, -15.32353, -9.36927, -4.683715, -2.17664, -0.8259934, 0.019797, 0.3120894, 0.2998486, 0.2346968, 0.1831921, 0.1566313, 0.1471648,
```
and for `nc` the line
```
    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 646660, -130424, -225600, -230232, -126656, 205756, 1295136, 4281888, 1.14502e+07, 2.412143e+07, 4.378973e+07, 292728.6, 13.0688, 0, 0, 0, 0, 
```
I don't know anything about microphysics, but looking at the output file from master it seems like these are O(1) errors. I noticed that `nc` diffs were all in the bottom half of the atmosphere (the first diff happened at lev=50, ncol=112, which is lat=25.3N, lon=118.3W). I don't know if this means anything to you...

bartgol:
Good point about single precision output (that is indeed the default for eamxx). I will try to output at full (double) model precision.

bartgol:
Indeed, with double precision output, we do get diffs after one timestep. I'm attaching two dumps at t=1st_step: the output from the branch, and the diff with master. The diffs seem all small (at least in a relative sense), but I don't know how small rel errors propagate with the impl in the std library. I tried to google the ulp errors of std math functions, and I was able to find [this](https://www.gnu.org/software/libc/manual/html_node/Errors-in-Math-Functions.html) for GCC, which shows 9ulps error for tgamma. It is possible that master was losing some precision, due to computing tgamma twice (for no reason)...


[diff_1st_step.txt](https://github.com/E3SM-Project/scream/files/13301632/diff_1st_step.txt)
[branch_1st_step.txt](https://github.com/E3SM-Project/scream/files/13301647/branch_1st_step.txt)

quantheory:
@bartgol I'm at a workshop today/tomorrow. From a quick look at this output, it's not really clear to me how the differences cause a large increase in the diffs at the next time step. I'm not sure how deep you want to go into this, but I think that in order to really understand this I would need more fine-grained information about the outputs, e.g. to determine whether the first well-above-ulp changes are coming from SHOC or P3. My suspicion is that at least one of those parameterizations is just sensitive to small `qc`/`nc` changes in general, and it's not specific to anything this PR is doing, but I don't know where exactly that's happening. I'm wondering if about limiters in the code that may be causing a significant bifurcation based on whether one of those variables hits a specific threshold.

quantheory:
@bartgol I suppose we both kind of forgot about this PR.

The question in my mind is, to merge this PR, do we need to track down where in the code we have rapid perturbation growth? Or is it sufficient to say that if this PR is only introducing roundoff-level differences at each time step, then it should be OK?

I'm generally in favor of figuring out why the model does unexpected things, in this case being extremely sensitive to changes in rain number even after only one or two time steps. But given the above tests, the rapid perturbation growth is a pre-existing feature of SCREAM that's orthogonal to this PR.

quantheory:
(And yes, I know I need to fix merge conficts now. Should be pretty trivial.)

bartgol:
@quantheory Personally, if a PR is small enough that one can verify visually the only differences are math ops rearrangements, I don't need to see any other study/test (no perturbation growth, e.g.).

When you're done rebasing, ping me again. I can give another fresh look, to make sure we didn't accidentally introduce answer-changing (on paper) mods.

bartgol:
Oh, i often rely on the autotester to do the testing. :)

And yes, any mathematical property that is not answer-changing on paper is fine to use.

Coming to think of it, though, what I said earlier is not entirely correct. When two expressions are equivalent on paper, one should also keep into account the stability of each of the two with respect to perturbations (I guess that's what you were getting at earlier). In this case, however, I don't think we are doing calculations that are more ill-conditioned than what was in master...

quantheory:
Well, there are parameterizations that use logarithms and `loggamma` functions, and there can be problems with naively removing logarithms in some of those cases (e.g. converting $e^{3 \log(a)-\log(b)}$ to $a^3/b$), particularly on single precision. E.g. maybe there's a division where both the numerator and denominator are too large (or small) for single precision, but their ratio is representable, so calculating exp of the log helps to get around overflow (or underflow) problems.

You can probably see how this is related to the conditioning problem you mentioned, but in this case it can be a simpler and more dire problem. You don't just get inaccurate results; you can get an immediate floating-point exception.

That said, part of the reason I mentioned my experience in applying this optimization, is because I know this rarely happens for this application. It is often the case that parameterizations set $\mu=0$ as a constant (in fact, I think SCREAM sets it to $0$, and EAM sets it to $1$, or vice versa). In that case, the formula
$$e^{loggamma(\mu+4) - \loggamma(\mu+1)}$$
is an extremely slow and low-precision method for calculating the number $6$ (or $24$ if $\mu=1$). For P3 this should be fine, but in 3-moment (or more) parameterizations, single or even double precision can be limiting because particle diameters are pretty small in SI units. However, a temporary change of units (e.g. from meters to microns) is sometimes a much more efficient solution than working with log-variables.

bartgol:
> For P3 this should be fine to fix with identities, but in 3-moment (or more) parameterizations, single or even double precision can be limiting because particle diameters are pretty small in SI units. However, a temporary change of units (e.g. from meters to microns) is sometimes a much more efficient solution than working with log-variables.

@quantheory I agree with the last point. In EAMxx, we use SI units when passing fields around, but that's just for pure convenience. We can a) use scaled version for certain vars, or b) locally rescale vars during calculations, and rescale them back before returning control to the AD.

Regarding this PR, I think diffs are expected, but I wonder why of all the AD tests, only `homme_shoc_cld_spa_p3_rrtmgp_128levels` and `p3_standalone` are failing comparison against baselines. There are other tests (with p3 in them) that do baselines comparison:

- homme_shoc_cld_p3_rrtmgp
- homme_shoc_cld_p3_rrtmgp_pg2
- homme_shoc_cld_spa_p3_rrtmgp
- homme_shoc_cld_spa_p3_rrtmgp_dp

quantheory:
@bartgol I'm not sure about the unexpectedly passing AD tests.

My guess is that the tests are short and the output is being written in single precision. So in any given grid cell, at any given time step, there's a very low chance that a change of a few ulps at double precision actually affects the output, and the tests are just too short for most of them to overcome that low probability. Especially since they may be checking `qc` and `qr`, but not `nc` and `nr`. So you could check this by running those tests longer and/or in double precision.

If that's what's happening, it's kind of weird, because above you ran an F case and found that this change causes big output differences in just 2 time steps. Maybe the AD tests are different, because they don't run some of the coupler code that would be exercised in an F case? Or maybe my speculation is just wrong? Not sure.

quantheory:
Tangentially, I do know that one of our group's postdocs, Justin Dong over at Livermore, has found that the equations used to calculate fluxes over ocean seem to be ill-posed in near-convectively-neutral conditions, potentially leading to kind of weird/noisy behavior from the iterative method that's supposed to solve those equations. It's very recent/preliminary work, but so far it suggests pretty significant errors over the midlatitudes and parts of the subtropics. So it's quite possible that a set of code changes in EAMxx could appear roundoff-level in AD tests, but much larger when the E3SM driver is active and calculating ocean fluxes.

bartgol:
@quantheory ok, given your explanation I'm inclined to merge the PR and bless. What do you think?

quantheory:
@bartgol Sounds good to me, I don't think there are any outstanding issues.

PR #2594, opened 2023-11-07
Add ML surface fluxes override

elynnwu:
This PR is a followup from #2588. This includes implementation for overriding surface fluxes in MLCorrection - specifically, net shortwave flux and downward longwave flux. This requires additional radaitive fluxes in MLCorrection where we need to be careful with how we add those fields. We do not pass packsize to 2d fields when requesting them.


AaronDonahue (dismissed):

AaronDonahue:
We can add the AUTOMERGE label once #2588 is merged.

AaronDonahue (approved):

components/eamxx/src/physics/ml_correction/eamxx_ml_correction_process_interface.cpp
```diff
@@ -69,6 +77,7 @@ void MLCorrection::initialize_impl(const RunType /* run_type */) {
   py_correction = pybind11::module::import("ml_correction");
   ML_model_tq = py_correction.attr("get_ML_model")(m_ML_model_path_tq);
   ML_model_uv = py_correction.attr("get_ML_model")(m_ML_model_path_uv);
+  ML_model_sfc_fluxes = py_correction.attr("get_ML_model")(m_ML_model_path_sfc_fluxes);
   ekat::enable_fpes(fpe_mask);
```
bartgol, line 81: I don't see `fpe_mask` being initialized. Perhaps you are missing a line like
```c++
fpe_mask = ekat::get_enabled_fpes();
```
before you start using it?

elynnwu: It's on line 70 in this file.

bartgol:
@elynnwu I submitted the review as comment instead of requested changes, so after aaron's approval, it got merged. Could you please open a ninja PR with the fix regarding fpe_mask? I'm afraid that, as it stands, ml correction will simply cause FPE to be disabled and never be re-enabled...

PR #2598, opened 2023-11-13
Fixes scream case builds on a couple systems

jgfouca:
Some minor changes to the macro files for ascent and pm-gpu were needed now that we are better about using the sharedlib kokkos build.


ambrad:
I'll test this on Ascent.

ambrad (approved):

ndkeen:
Should we add the same to the nvidia build?  nvidiagpu_pm-gpu.cmake

ambrad (approved):

PR #2599, opened 2023-11-15
Disable snan init for cice

jgfouca:
It causes things to break for some eamxx cases. We don't want to spend resources debugging cice.


jgfouca:
@amametjanov ^

ambrad:
It might be enough to do this for just ice_grid.F90, but since CICE is going away soon, it doesn't really matter. Keeps the nightlies running until we switch fully to MPASSI.

ambrad (approved):

PR #2600, opened 2023-11-17
Fix chrysalis cmake macro to fix DIFFs

jgfouca:
The shared kokkos config did not quite match the config that was being used for the eamxx build of kokkos.


cime_config/machines/cmake_macros/intel_chrysalis.cmake
```diff
@@ -23,3 +23,7 @@ if (MPILIB STREQUAL impi)
   set(MPICXX "mpiicpc")
   set(MPIFC "mpiifort")
 endif()
+string(APPEND KOKKOS_OPTIONS " -DKokkos_ARCH_ZEN2=On")
```
bartgol, line 26: I think our next step (to keep code duplication under control) should be to include in the cmake mach files other cmake files for kokkos options, much like we do in EKAT.

oksanaguba:
If i can ask about this -- we used to have 2 kokkos submodules, was there a recent PR to get rid of one of them (i did not find it in recent PRs)? From Luca's comments, it seems the one from ekat is going away? How will it work in scream builds, will standalone still need external/ekat/extern/kokkos, but cime build will need external/kokkos?

cime_config/machines/cmake_macros/intel_chrysalis.cmake
jgfouca: Agree.

jgfouca:
@oksanaguba , yes, there used to be two kokkos sumbodules. Now, the one under EKAT is the only one. `externals/kokkos` should be gone if you are on a recent master commit. Both eamxx and e3sm will used the EKAT kokkos.

ambrad (approved):

PR #2603, opened 2023-11-20
Allow to flush output files before they are closed

bartgol:
Usually, if EAMxx crashes, we should intercept the crash, and flush any open file. However, it is possible that this doesn't work (e.g., crash inside another component, or an exception is thrown during stack unwinding). In this case, it can help to have EAMxx flush output files more frequently.

This PR adds this capability. By default, we only flush when we close the file, but the user can request a different frequency by setting the `flush_frequency` paraeter in the output yaml file.

### DIFF
--- a/components/eamxx/src/share/io/scream_output_manager.cpp
+++ b/components/eamxx/src/share/io/scream_output_manager.cpp
@@ -182,6 +182,7 @@ setup (const ekat::Comm& io_comm, const ekat::ParameterList& params,
 
       // File specs
       m_checkpoint_file_specs.max_snapshots_in_file = 1;
+      m_checkpoint_file_specs.flush_frequency = 1;
       m_checkpoint_file_specs.filename_with_mpiranks    = pl.get("MPI Ranks in Filename",false);
       m_checkpoint_file_specs.save_grid_data = false;
       m_checkpoint_file_specs.hist_restart_file = true;
@@ -487,6 +488,8 @@ void OutputManager::run(const util::TimeStamp& timestamp)
         eam_pio_closefile(filespecs.filename);
         filespecs.num_snapshots_in_file = 0;
         filespecs.is_open = false;
+      } else if (filespecs.file_needs_flush()) {
+        eam_flush_file (filespecs.filename);
       }
     };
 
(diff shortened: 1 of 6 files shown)


AaronDonahue:
@elynnwu ^^

components/eamxx/src/share/io/scream_output_manager.cpp
```diff
@@ -583,6 +586,11 @@ set_params (const ekat::ParameterList& params,
         "Error! For restart output, max snapshots per file must be 1.\n"
         "   Note: you don't have to specify this parameter for restart output.\n");
 
+    m_output_file_specs.flush_frequency = m_params.get("flush_frequencyuency",1);
```
AaronDonahue, line 589: There's a typo in the key "frequencyuency"

components/eamxx/src/share/io/scream_output_manager.cpp
```diff
@@ -611,6 +619,7 @@ set_params (const ekat::ParameterList& params,
     constexpr auto large_int = 1000000;
     m_output_file_specs.max_snapshots_in_file = m_params.get<int>("Max Snapshots Per File",large_int);
     m_filename_prefix = m_params.get<std::string>("filename_prefix");
+    m_output_file_specs.flush_frequency = m_params.get("flush_frequencyuency",m_output_file_specs.max_snapshots_in_file);
```
AaronDonahue, line 622: same here w/ typo

AaronDonahue (changes requested):
I don't suppose there is a way to create a unit test for this capability.  Maybe we could have a test that runs 5 steps w/ output every 1 step, flush every 2 steps and a max snaps = 10.  Then trigger a forced fail after step 4, like `if nstep > 4; FORCED FAIL`.  Then have a cprnc compare with a file we have set to have max snaps = 4 and see if they match.

AaronDonahue:
@bartgol , thanks for getting to this so quickly and having a solution in place.

bartgol:
> I don't suppose there is a way to create a unit test for this capability. Maybe we could have a test that runs 5 steps w/ output every 1 step, flush every 2 steps and a max snaps = 10. Then trigger a forced fail after step 4, like `if nstep > 4; FORCED FAIL`. Then have a cprnc compare with a file we have set to have max snaps = 4 and see if they match.

I thought about it, but it seems very complicated. I could see if I can cook it up real quick, but given that it is a very niche usage, which only matters for some debugging effort, I'm tempted to leave as is...

AaronDonahue (approved):

PR #2606, opened 2023-11-22
EAMxx: rename cxx_blah wrapper to avoid conflict with Homme

### LINKED ISSUE #2604: Multiple definitions of cxx_log
I'm not sure why this is not picked up in master, but I'm getting this error on a branch were I was doing some cmake changes.

The function cxx_log is declared in `physics_share.cpp` with C linkage (so no namespace qualification), to make it Fortran interoperable. However, in Homme, we also declare a cxx_log function with C linkage, in `BfbUtils.cpp`. I am not sure how these two are not already colliding in master. It looks to me as if they should.

I'm going to investigate this a bit more.

bartgol:
This solves a link error that for some reason does not always appear. Both homme and scream define a function with C linkage with the same name.

We could change the name of the fcn in Homme, but that requires lengthier process (due to PR workflow in E3SM)

Fixes #2604

### DIFF
--- a/components/eam/src/physics/cam/bfb_math.inc
+++ b/components/eam/src/physics/cam/bfb_math.inc
@@ -6,8 +6,8 @@
 ! Make sure to place the following lines at the top of any modules
 ! that use these macros:
 !
-! use physics_share_f2c, only: cxx_pow, cxx_sqrt, cxx_cbrt, cxx_gamma, cxx_log, &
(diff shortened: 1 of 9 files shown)


ambrad (approved):

bartgol:
I verified that this does indeed solve the link problem in the branch that exposed the issue. Merging.

PR #2609, opened 2023-11-22
Some mods to CrateUnitTest suite

bartgol:
This PR mostly wanted to avoid having multiple source files scattered around the `tests` folder, which ultimately do the same thing: crate AD, register products, run time loop. In the process, I ended up changing quite a few things, trying to clean up stuff and reduce copy+paste errors. I'm going to list them here.

- Change CreateUnitTest macro, so that libs are passed via LIBS argument. This removes the need to bundle up libs in a string representing a list, which was needed since the 2nd and 3rd arg of the macro were both lists.
- Always link `scream_share` to all our unit tests, so no need to pass it in the LIBS list
- Add macro to creat AD unit test. This automatically links `scream_control scream_io diagnostics`, and adds the `driver` label. User can of course still pass other libs (it usually will) or labels.
- Add `eamxx_ad_test.cpp` template source file for ad-driven system tests. Notice that, by linking in only the libs needed for the test, `register_physics.hpp` will only provide those processes (same for dynamics). This leverages the fact that each package adds the `EAMXX_HAS_<PKGNAME>` CPP macro as PUBLIC define on the target.
- Add shortcut macro to create diagnostic unit test, which makes `src/diagnostics/test/CMakeLists.txt` a tad easier to read (especially if text editor is "narrow", which often happens with multiple vertically split windows).
- Renamed most output yaml files in ad test as just "output.yaml". This makes tab completion faster, since calling them, say, `shoc_standalone_output.yaml` will make them clash with the output nc file name. Faster tab completion when doing lengthy debug campaigns can help.
- Use some CMake vars to store fixtures/base test names in the tests folder. This reduces the chance of test name misspelling, causing a dependency to "disappear" (CMake does not complain if a required fixture is not provided anywhere, and simply assumes that all dependencies are satisfied).

Personally, I would have also liked to pass test source files via a SOURCES argument, like
```
CreateUnitTest (my_test
	SOURCES foo.cpp bar.cpp baz.cpp
	LIBS scream_io diagnostics
	...
)
```
I find it cleaner/clearer, since the inputs have a clear well spelled out meaning also at the call point, but I don't know if it's a shared feeling. If you guys feel the same way, I can quickly add a commit for that though.


components/eamxx/cmake/ScreamUtils.cmake
```diff
@@ -45,40 +45,27 @@ set(SCREAM_CUT_TEST_MV_ARGS ${CUT_TEST_MV_ARGS})
...
-  set(TEST_INCLUDE_DIRS
-    ${SCREAM_INCLUDE_DIRS}
-    ${CMAKE_CURRENT_SOURCE_DIR}
-    ${CMAKE_CURRENT_BINARY_DIR}
-  )
-
-  set(test_libs "${scream_libs};scream_test_support")
-  list(APPEND test_libs "${SCREAM_TPL_LIBRARIES}")
-
-  if (SCREAM_Fortran_FLAGS)
```
bartgol, line 68: This is not needed, since project-wide F flags are already attached to `scream_share`.

components/eamxx/cmake/ScreamUtils.cmake
```diff
@@ -45,40 +45,27 @@ set(SCREAM_CUT_TEST_MV_ARGS ${CUT_TEST_MV_ARGS})
...
 ###############################################################################
-function(CreateUnitTestExec exec_name test_srcs scream_libs)
+function(CreateUnitTestExec exec_name test_srcs)
 ###############################################################################
-  cmake_parse_arguments(cute "${SCREAM_CUT_EXEC_OPTIONS}" "${SCREAM_CUT_EXEC_1V_ARGS}" "${SCREAM_CUT_EXEC_MV_ARGS}" ${ARGN})
-  CheckMacroArgs(CreateUnitTestExec cute "${SCREAM_CUT_EXEC_OPTIONS}" "${SCREAM_CUT_EXEC_1V_ARGS}" "${SCREAM_CUT_EXEC_MV_ARGS}")
-
-  separate_cut_arguments(cute "${SCREAM_CUT_EXEC_OPTIONS}" "${SCREAM_CUT_EXEC_1V_ARGS}" "${SCREAM_CUT_EXEC_MV_ARGS}" options)
-
-  set(TEST_INCLUDE_DIRS
```
bartgol, line 59: `SCREAM_INCLUDE_DIRS` is never used. And I thought that current source/binary dirs can be added when needed, via the `INCLUDE_DIRS` optional arg.

components/eamxx/src/physics/shoc/tests/CMakeLists.txt
```diff
@@ -75,27 +73,36 @@ set(SHOC_TESTS_SRCS
...
 # NOTE: tests inside this if statement won't be built in a baselines-only build
 if (NOT SCREAM_BASELINES_ONLY)
-  CreateUnitTest(shoc_tests    "${SHOC_TESTS_SRCS}" "${NEED_LIBS}"    THREADS 1 ${SCREAM_TEST_MAX_THREADS} ${SCREAM_TEST_THREAD_INC} DEP shoc_tests_ut_np1_omp1)
+  CreateUnitTest(shoc_tests    "${SHOC_TESTS_SRCS}"
+    LIBS shoc
+    THREADS 1 ${SCREAM_TEST_MAX_THREADS} ${SCREAM_TEST_THREAD_INC}
+  )
+
   if (NOT SCREAM_SMALL_KERNELS)
-    CreateUnitTest(shoc_sk_tests "${SHOC_TESTS_SRCS}" "${SK_NEED_LIBS}" THREADS 1 ${SCREAM_TEST_MAX_THREADS} ${SCREAM_TEST_THREAD_INC} DEP shoc_tests_ut_np1_omp1 EXE_ARGS shoc_main_bfb)
```
bartgol, line 80: After discussing with @tcclevenger , I'm pretty sure all usages of `DEP` were caused by copy+paste previous code. I suspect this was originally needed (eons ago) when p3 used to compute some tables on the fly. In that case, we had to ensure only one test wrote to file. A while ago we moved the tables to the inputdata folder, so we no longer compute them. The p3 tests are indeed free of the DEP keyword, but I suspect it got copy+pasted in shoc, and from there in tms and doubly_periodic tests. If you can think of a reason why we still need DEP, let me know, and I'll revert the commit.

bartgol:
I'm still seeing a few failures in some `blah_np1_vs_npX_bfb` tests, probably due to some typo in the fixtures and/or filenames. I will fix them shortly.

tcclevenger (approved):

tcclevenger (approved):

tcclevenger (approved):

PR #2610, opened 2023-11-25
fix pm-gpu machine in eamxx scripts

mahf708:
Someone still needs to add pm-cpu... 

Without this edit, I cannot use the test launcher with either pm-gpu or perlmutter because of the inconsistency between perlmutter here and pm-gpu.cmake elsewhere, so unifying under pm-gpu.


ndkeen:
I have not needed any changes in that file -- I'm sure it's fine to change. It will be slightly different than what pm-gpu uses via CIME.

jgfouca (approved):
This seems like a good cleanup of an inconsistency. I don't see the `perlmutter` name used anywhere but I do see `pm-gpu` in the machine files.

PR #2611, opened 2023-11-25
remove psutil from machines_specs

mahf708:
Get rid of psutil in machine_specs, but not elsewhere.

The motivation behind this is that the earliest call to test-launcher, etc. will go through machine_specs, and if psutil is not installed, it will necessitate the user to install it, etc. which can be quite annoying. The psutil logic in here wasn't used to calculate anything fancy (despite the options) but a plain logical cpu count available. The builtin os module has an easy alternative. Note that if we want to exclude hyperthreads, we will have to revert back to psutil. However, because this is only for building/compiling the tests, I don't think it matters. This will be a decent "quality of life" improvement for my workflow (which strictly follows Luca's recommendation for how I should do the tests via test-launcher).

### DIFF
--- a/components/eamxx/scripts/machines_specs.py
+++ b/components/eamxx/scripts/machines_specs.py
@@ -1,8 +1,6 @@
-from utils import expect, run_cmd_no_fail, ensure_psutil
+from utils import expect, run_cmd_no_fail
 
 import os, sys, pathlib
-ensure_psutil()
-import psutil
 
 # MACHINE -> (env_setup,                      # list of shell commands to set up scream-approved env
 #             compilers,                      # list of compilers [CXX, F90, C]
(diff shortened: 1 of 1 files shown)


components/eamxx/scripts/machines_specs.py
```diff
@@ -161,23 +159,13 @@ def get_mach_baseline_root_dir(machine):
     return MACHINE_METADATA[machine][3]
 
 ###############################################################################
-def logical_cores_per_physical_core():
```
bartgol, line 164: This funciton was used in `test_all_scream.py` as well.

bartgol (commented):
I'm confused: if you use `test-launcher`, you are using `test-all-scream` (there's no point in using `test-launcher` otherwise). And since `test-all-scream` _does_ need psutil, I'm not sure what we gain by removing the dependency here...

TBC, if we can just use `os` instead of `psutil`, I'm happy (the former is always installed, unlike the latter), but I don't think we can do away with psutil anyways...

PR #2613, opened 2023-11-27
nudging from coarse data

mahf708:
adds the ability to nudge from coarse data. Both the vertical interpolation and regional nudging (weighting) should be unaffected. However, I did have to make edits in code related to vertical interp and regional nudging, so please take a look. The tests pass on chrysalis. The vertical and horizontal interpolation should work well together. The weighting should also work well with both, but only if applied at the model grid (in the future, we could contemplate applying it to the external grid if we need to).

Ultimately, I think we will need to rewrite this whole interface soon to make it easier to maintain. Currently, there are areas of the code that code be improved.

A current limitation is the need to provide an explicit list of data files. In my testing, this didn't work quite well (it was not possible to provide a list of more than ~5 files).

### DIFF
--- a/components/eamxx/cime_config/namelist_defaults_scream.xml
+++ b/components/eamxx/cime_config/namelist_defaults_scream.xml
@@ -240,7 +240,6 @@ be lost if SCREAM_HACK_XML is not enabled.
     STATIC_1D_VERTICAL_PROFILE: The dataset uses a fixed in time single pressure profile, variable name 'p_lev' with dimension (nlev).">TIME_DEPENDENT_3D_PROFILE</source_pressure_type>
       <source_pressure_file type="string" doc="If STATIC_1D_VERTICAL_PROFILE, this is an optional arg to point to a file with the source pressure levels defined.  
     Default is to look for p_levs in the first nudging_filename file"/>
-      <do_nudging_refine_remap type="logical" doc="Flag for whether to remap nudging fields to the physics grid">false</do_nudging_refine_remap>
       <nudging_refine_remap_mapfile type="string" doc="File containing the remapping weights for nudging fields to the physics grid">"no-file-given"</nudging_refine_remap_mapfile>
     </nudging>
 
--- a/components/eamxx/cime_config/namelist_defaults_scream.xml
+++ b/components/eamxx/cime_config/namelist_defaults_scream.xml
@@ -231,7 +231,7 @@ be lost if SCREAM_HACK_XML is not enabled.
       <nudging_filename type="array(string)"/>
       <nudging_fields type="array(string)" doc="List of fields to be nudged.  Note, syntax of 'A:B' represents nudging field A with data from field B in files, syntax of 'A' assumes that nudging file has the same variables name as EAMxx"/>
       <nudging_timescale type="integer" doc="Timescale to apply nudging tendencies, 0: full replacement, >0: actual timescale">0</nudging_timescale>
-      <use_nudging_weights type="logical" doc="Flag for nudging weights option"/>
+      <use_nudging_weights type="logical" doc="Flag for nudging weights option">false</use_nudging_weights>
       <nudging_weights_file type="string" doc="weights that relax the nudging fields update"/>
       <source_pressure_type type="string" 
 	                    valid_values="TIME_DEPENDENT_3D_PROFILE,STATIC_1D_VERTICAL_PROFILE"
--- a/components/eamxx/cime_config/namelist_defaults_scream.xml
+++ b/components/eamxx/cime_config/namelist_defaults_scream.xml
@@ -240,7 +240,7 @@ be lost if SCREAM_HACK_XML is not enabled.
     STATIC_1D_VERTICAL_PROFILE: The dataset uses a fixed in time single pressure profile, variable name 'p_lev' with dimension (nlev).">TIME_DEPENDENT_3D_PROFILE</source_pressure_type>
       <source_pressure_file type="string" doc="If STATIC_1D_VERTICAL_PROFILE, this is an optional arg to point to a file with the source pressure levels defined.  
     Default is to look for p_levs in the first nudging_filename file"/>
-      <nudging_refine_remap_mapfile type="string" doc="File containing the remapping weights for nudging fields to the physics grid">"no-file-given"</nudging_refine_remap_mapfile>
+      <nudging_refine_remap_mapfile type="string" doc="Refine-remapping mapfile from the source nudging dataset to the physics grid">"no-file-given"</nudging_refine_remap_mapfile>
     </nudging>
 
     <!-- ML correction -->
--- a/components/eamxx/cime_config/namelist_defaults_scream.xml
+++ b/components/eamxx/cime_config/namelist_defaults_scream.xml
@@ -240,6 +240,8 @@ be lost if SCREAM_HACK_XML is not enabled.
     STATIC_1D_VERTICAL_PROFILE: The dataset uses a fixed in time single pressure profile, variable name 'p_lev' with dimension (nlev).">TIME_DEPENDENT_3D_PROFILE</source_pressure_type>
       <source_pressure_file type="string" doc="If STATIC_1D_VERTICAL_PROFILE, this is an optional arg to point to a file with the source pressure levels defined.  
     Default is to look for p_levs in the first nudging_filename file"/>
+      <do_nudging_refine_remap type="logical" doc="Flag for whether to remap nudging fields to the physics grid">false</do_nudging_refine_remap>
+      <nudging_refine_remap_mapfile type="string" doc="File containing the remapping weights for nudging fields to the physics grid">"no-file-given"</nudging_refine_remap_mapfile>
     </nudging>
 
     <!-- ML correction -->
--- a/components/eamxx/cime_config/namelist_defaults_scream.xml
+++ b/components/eamxx/cime_config/namelist_defaults_scream.xml
@@ -244,13 +244,13 @@ be lost if SCREAM_HACK_XML is not enabled.
         type="string"
         doc="Refine-remapping mapfile from the source nudging dataset to the physics grid"
       >
-      "no-file-given"
+        "no-file-given"
       </nudging_refine_remap_mapfile>
       <nudging_refine_remap_vert_cutoff
         type="real"
         doc="A vertical cutoff to go with refine-remap logic (in units of p_mid) where the nudging is turned off above it (closer to the surface)"
       >
-      0.0
+        0.0
       </nudging_refine_remap_vert_cutoff>
     </nudging>
 
--- a/components/eamxx/cime_config/namelist_defaults_scream.xml
+++ b/components/eamxx/cime_config/namelist_defaults_scream.xml
@@ -240,7 +240,18 @@ be lost if SCREAM_HACK_XML is not enabled.
     STATIC_1D_VERTICAL_PROFILE: The dataset uses a fixed in time single pressure profile, variable name 'p_lev' with dimension (nlev).">TIME_DEPENDENT_3D_PROFILE</source_pressure_type>
       <source_pressure_file type="string" doc="If STATIC_1D_VERTICAL_PROFILE, this is an optional arg to point to a file with the source pressure levels defined.  
     Default is to look for p_levs in the first nudging_filename file"/>
-      <nudging_refine_remap_mapfile type="string" doc="Refine-remapping mapfile from the source nudging dataset to the physics grid">"no-file-given"</nudging_refine_remap_mapfile>
+      <nudging_refine_remap_mapfile
+        type="string"
+        doc="Refine-remapping mapfile from the source nudging dataset to the physics grid"
+      >
+      "no-file-given"
+      </nudging_refine_remap_mapfile>
+      <nudging_refine_remap_vert_cutoff
+        type="real"
+        doc="A vertical cutoff to go with refine-remap logic (in units of p_mid) where the nudging is turned off above it (closer to the surface)"
+      >
+      0.0
+      </nudging_refine_remap_vert_cutoff>
     </nudging>
 
     <!-- ML correction -->
--- a/components/eamxx/docs/user/coarse_nudging.md
+++ b/components/eamxx/docs/user/coarse_nudging.md
@@ -1,6 +1,6 @@
 # Nudging from coarse data
 
-Because EAMxx is designed to run at ultra high resolution, it is not feasible to produce nudging data at the same resolution.
+Because EAMxx is designed to support ultra-high resolutions (in fact, that was the initial reason for its inception), it is not feasible to produce nudging data at the same resolution.
 Instead, in EAMxx, it is possible to nudge from coarse data.
 This is done by remapping the coarse data provided by the user to the runtime physics grid of EAMxx.
 In order to enable nudging from coarse data, the user must provide nudging data at the coarse resolution desired and an appropriate     ncremap-compatible mapping file.
@@ -13,7 +13,7 @@ A limitation for now is that the nudging data must be provided explicitly, eithe
 This can be problematic for long list of files, but we are working on a solution to this problem.
 
 Let's say that the nudging data is provided as one file in the following path: `/path/to/nudging_data_ne4pg2_L72.nc`.
-Then, a mapping file is provided as `/path/to/mapping_file_ne4pg2_to_ne120pg2.nc`.
+Then, a mapping file is provided as `/another/path/to/mapping_file_ne4pg2_to_ne120pg2.nc`.
 Then if the physics grid is ne120pg2, the user must enable the nudging process, specify the nudging files, and provide the specifies the nudging data and a remap file.
 In other words, the following options are needed:
 
@@ -21,5 +21,5 @@ In other words, the following options are needed:
 ./atmchange atm_procs_list=(sc_import,nudging,homme,physics,sc_export)
 ./atmchange nudging_fields=U,V
 ./atmchange nudging_filename=/path/to/nudging_data_ne4pg2_L72.nc
-./atmchange nudging_refine_remap_mapfile=/path/to/mapping_file_ne4pg2_to_ne120pg2.nc
+./atmchange nudging_refine_remap_mapfile=/another/path/to/mapping_file_ne4pg2_to_ne120pg2.nc
 ```
--- a/components/eamxx/docs/user/coarse_nudging.md
+++ b/components/eamxx/docs/user/coarse_nudging.md
@@ -1,0 +1,25 @@
+# Nudging from coarse data
+
+Because EAMxx is designed to run at ultra high resolution, it is not feasible to produce nudging data at the same resolution.
+Instead, in EAMxx, it is possible to nudge from coarse data.
+This is done by remapping the coarse data provided by the user to the runtime physics grid of EAMxx.
+In order to enable nudging from coarse data, the user must provide nudging data at the coarse resolution desired and an appropriate     ncremap-compatible mapping file.
+
+## Example setup
+
+A user can produce coarse nudging data from running EAMxx or EAM at a ne30pg2 or any other applicable resolution.
+Additionally, several users in the E3SM projects have produced nudging data at the ne30pg2 resolution from the MERRA2 and ERA5 datasets.
+A limitation for now is that the nudging data must be provided explicitly, either as one file or as a list of files.
+This can be problematic for long list of files, but we are working on a solution to this problem.
+
+Let's say that the nudging data is provided as one file in the following path: `/path/to/nudging_data_ne4pg2_L72.nc`.
+Then, a mapping file is provided as `/path/to/mapping_file_ne4pg2_to_ne120pg2.nc`.
+Then if the physics grid is ne120pg2, the user must enable the nudging process, specify the nudging files, and provide the specifies the nudging data and a remap file.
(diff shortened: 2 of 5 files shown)


components/eamxx/cime_config/namelist_defaults_scream.xml
```diff
@@ -240,6 +240,8 @@ be lost if SCREAM_HACK_XML is not enabled.
     STATIC_1D_VERTICAL_PROFILE: The dataset uses a fixed in time single pressure profile, variable name 'p_lev' with dimension (nlev).">TIME_DEPENDENT_3D_PROFILE</source_pressure_type>
       <source_pressure_file type="string" doc="If STATIC_1D_VERTICAL_PROFILE, this is an optional arg to point to a file with the source pressure levels defined.  
     Default is to look for p_levs in the first nudging_filename file"/>
+      <do_nudging_refine_remap type="logical" doc="Flag for whether to remap nudging fields to the physics grid">false</do_nudging_refine_remap>
```
bartgol, line 243: Can't you just deduce this from the size of the two grids (nudging data and eamxx physics)?

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
```diff
@@ -268,24 +342,39 @@ void Nudging::run_impl (const double dt)
...
                                                m_num_levs);
     }
 
+    // Refine-remap onto target atmosphere state horiz grid ("int")
+    // Note that we are going from hxt to int here
+    if(m_refine_remap) {
+      // We have to register the fields
+      refine_remapper->registration_begins();
+      refine_remapper->register_field(hxt_state_field, int_state_field);
+      refine_remapper->registration_ends();
```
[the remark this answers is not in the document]
bartgol: You can restart field registration once ended. Simply move the registration begin/end to right before/after the for loop.

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.hpp
```diff
@@ -132,6 +133,14 @@ class Nudging : public AtmosphereProcess
 
   std::vector<std::string> m_fields_nudge;
 
+  /* Nudge from coarse data */
+  // if true, remap coarse data to fine grid
+  bool m_refine_remap;
+  // file containing coarse data mapping
+  std::string m_refine_remap_file;
+  // (refining) remapper object
+  std::shared_ptr<scream::RefiningRemapperP2P> refine_remapper;
```
bartgol, line 142: Nitpicky, but for consistency, you could add the `m_` prefix to this var too...

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
```diff
@@ -268,24 +342,39 @@ void Nudging::run_impl (const double dt)
...
-                                               int_state_view,
+                                               hxt_state_view,
                                                int_mask_view,
                                                m_num_src_levs,
                                                m_num_levs);
     }
 
+    // Refine-remap onto target atmosphere state horiz grid ("int")
+    // Note that we are going from hxt to int here
+    if(m_refine_remap) {
```
bartgol, line 366: Up to you, but if you want, you could do away with `m_refine_remap`, and simply do `if (refine_remapper)`, since `std::shared_ptr` has an `operator bool` overload...

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
```diff
@@ -125,17 +138,67 @@ void Nudging::initialize_impl (const RunType /* run_type */)
...
+  // Initialize the time interpolator
   m_time_interp = util::TimeInterpolation(grid_ext, m_datafiles);
 
   constexpr int ps = SCREAM_PACK_SIZE;
-  const auto& grid_name = m_grid->name();
+  // To be extra careful, this should be the ext_grid
+  const auto& grid_ext_name = grid_ext->name();
   if (m_src_pres_type == TIME_DEPENDENT_3D_PROFILE) {
-    create_helper_field("p_mid_ext", scalar3d_layout_mid, grid_name, ps);
+    create_helper_field("p_mid_ext", scalar3d_layout_mid, grid_ext_name, ps);
```
bartgol, line 201: I noticed you have a few places where you create a helper field, and immediately retrieve it. You may consider changing `create_helper_field` to return the newly created field, to make the code simpler.

Instead of
```
create_helper_field("blah",...);
auto blah = get_helper_field("blah");
```
simply do
```
auto blah = create_helper_field("blah");
```

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
```diff
@@ -268,24 +364,38 @@ void Nudging::run_impl (const double dt)
...
     }
 
+    // Refine-remap onto target atmosphere state horiz grid ("int")
+    // Note that we are going from hxt to int here
+    if(m_refine_remap) {
+      // Call the remapper
+      print_field_hyperslab (ext_state_field);
+      print_field_hyperslab (hxt_state_field);
+      print_field_hyperslab (int_state_field);
+      refine_remapper->remap(true);
```
bartgol, line 393: You are calling the remap method inside a fol loop, causing all fields to be remapped multiple times. The fields not yet remapped vertically will likely still have invalid values, no?

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
```diff
@@ -437,13 +406,67 @@ void Nudging::run_impl (const double dt)
...
-	  int_state_view_1d(ipack)[iidx] = fill_value;
+	  hxt_state_view_1d(ipack)[iidx] = fill_value;
 	}
       }
     });
 
+  }
+
+  // Open the registration!
+  if(m_refine_remap) {
```
bartgol: Looks like you could do just one if statement, and be more concise.

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
```diff
@@ -125,17 +139,67 @@ void Nudging::initialize_impl (const RunType /* run_type */)
...
+  // Initialize the refining remapper stuff at the outset,
+  // because we need to know the grid information
+  if(m_refine_remap) {
+    // For now, we are doing the horizontal interpolation last,
+    // so we use the m_grid (model physics) as the target
+    // TODO: maybe clean this up?
+    auto grid_tgt = m_grid->clone(m_grid->name(), false);
+    auto refine_remapper_p2p =
+        std::make_shared<RefiningRemapperP2P>(grid_tgt, m_refine_remap_file);
+    refine_remapper = refine_remapper_p2p;
```
bartgol, line 151: IF you ever get bored of doing lots of `if (refine_remapper)`, you could consider _always_ have a remapper, with the catch that, if remap is not needed, it is of type [DoNothingRemapper](https://github.com/E3SM-Project/scream/blob/master/components/eamxx/src/share/grid/remap/do_nothing_remapper.hpp), which is kind of a placeholder. But that way, you can _always_  call stuff like `refine_remapper->registration_ends();` (it's a no-op for `DoNothingRemapper`). You can even call the remap methods (which are also no-op methods), or use it to get the ext grid (which would be the remapper src grid).

This is really not necessary, but I'm pointing it out in case you notice that all these if statements make the code less readable.

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
```diff
@@ -156,24 +220,50 @@ void Nudging::initialize_impl (const RunType /* run_type */)
...
+    create_helper_field(name,     layout,              grid_int_name, ps);
+    create_helper_field(name_ext, scalar3d_layout_mid, grid_ext_name, ps);
+    create_helper_field(name_hxt, scalar3d_layout_hid, grid_hxt_name, ps);
+    // No need to follow with hxt because we are not reading it externally
     auto field_ext = get_helper_field(name_ext);
+    if(m_refine_remap) {
+      auto field_hxt = get_helper_field(name_hxt);
+      auto field_int = get_helper_field(name);
+      refine_remapper->register_field(field_hxt, field_int);
+    }
```
bartgol, line 249: E.g., using DoNothingRemapper when remap is not necessary, here you could do:
```
auto field_int = create_helper_field(name, ...);
auto field_hxt = create_helper_field(name_hxt);
refine_remapper->register_field(field_hxt,field_int);
```
In case the remapper is the do nothing one, this call is almost a no op.

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
```diff
@@ -125,18 +138,57 @@ void Nudging::initialize_impl (const RunType /* run_type */)
...
   using namespace ShortFieldTagsNames;
 
-  // Initialize the time interpolator
-  auto grid_ext = m_grid->clone(m_grid->name(), false);
-  grid_ext->reset_num_vertical_lev(m_num_src_levs);
+  // Initialize the refining remapper stuff at the outset,
+  // because we need to know the grid information.
+  // For now, we are doing the horizontal interpolation last,
+  // so we use the m_grid (model physics) as the target grid
+  grid_int = m_grid->clone(m_grid->name(), false);
```
bartgol, line 145: I think you want the last argument to be `true`, since I think shallow copy is enough for you.

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
```diff
@@ -125,18 +138,57 @@ void Nudging::initialize_impl (const RunType /* run_type */)
...
+    grid_ext->reset_num_vertical_lev(m_num_src_levs);
+    // The second grid is grid_hxt (external horiz grid, but model physics
+    // vert grid, so potentially a bit of a mess)
+    grid_hxt = grid_ext->clone(grid_ext->name(), false);
+    grid_hxt->reset_num_vertical_lev(m_num_levs);
+  } else {
+    // DoNothingRemapper
+    // If not refine-remapping, then use whatever was used before,
+    // i.e., deep clone the physics grid
+    grid_ext = m_grid->clone(m_grid->name(), false);
```
bartgol, line 167: Here as well. Shallow clone should be ok.

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
```diff
@@ -124,19 +136,77 @@ void Nudging::apply_weighted_tendency(Field& base, const Field& next, const Fiel
...
+    /* quick check here to ensure we are in good stnading */
+    // If the user gives a mapfile, we assume we are refine-remapping,
+    // but we should check if the mapfile is actually remapping stuff,
+    // so we compare the global columns of the target and source grids
+    auto grid_int_global_cols = grid_int->get_num_global_dofs();
+    auto grid_ext_global_cols = grid_ext->get_num_global_dofs();
+    EKAT_REQUIRE_MSG(
+        grid_int_global_cols != grid_ext_global_cols,
+        "Error! Nudging::initialize_impl - the mapfile given for "
+        "refine-remapping does not remap anything.  Please check the mapfile.");
```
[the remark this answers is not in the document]
bartgol: I think this is backwards, compared to what I had in mind. You set `m_refine_remap` if an entry is found in the input file specifying the map file and this line checks that indeed the two grids are different (no neX->neX remap).

However, I think the logic should be backwards:
1. if the nudging file has ncol different from eamxx phys grid, set m_refine_remap=true;
2. if refine_remap=true, you _must_ pass a map file, otherwise there should be _no_ map file provided. if map file provided when ncols matches, you could decide to error out or just issue a warning (it may be the user specified wrong nudging file, or maybe the nudging file is correct, and they forgot to delete the map file entry)
3. If a map file is specified, you should check that the two ncols match the nudging file and eamxx grid values.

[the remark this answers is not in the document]
bartgol: Uhm, no you can just do the checks in the `set_grids` method, when all grids are available. And you only need to check the map file if the model and nudging ncols are different (which you would have to do anyways, since you need to create a remapper).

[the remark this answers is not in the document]
bartgol: Notice that `set_grids` happens quite early in the init sequence. Atm procs creation is usually fast, and not much can happen in the constructor, since no grid info is yet available. I think it's perfectly ok to put this kind of checks in `set_grids`.

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
```diff
@@ -93,6 +86,46 @@ void Nudging::set_grids(const std::shared_ptr<const GridsManager> grids_manager)
...
     scorpio::eam_pio_closefile(m_static_vertical_pressure_file);
   }
+
+  /* Check for consistency between nudging files, map file, and remapper */
+
+  // Number of columns globally
+  auto m_num_cols_global = m_grid->get_num_global_dofs(); 
+
+  // Get the information from the first nudging data file
+  scorpio::register_file(m_datafiles[0],scorpio::Read);
```
bartgol, line 96: FYI, you don't need to manually open/close the file for this operation. `get_dimlen` will automatically register/close the file if not yet open.

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
```diff
@@ -93,6 +86,46 @@ void Nudging::set_grids(const std::shared_ptr<const GridsManager> grids_manager)
...
+                     << "nudging data file and/or the mapfile.");
+    EKAT_REQUIRE_MSG(num_cols_remap_b == m_num_cols_global,
+                      "Error! Nudging::set_grids - the number of columns in the model grid "
+                      << std::to_string(m_num_cols_global) << " does not match the number of columns in the "
+                      << "mapfile " << std::to_string(num_cols_remap_b) << ".  Please check the "
+                      << "model grid and/or the mapfile.");
+    // If we get here, we are good to go!
+    m_refine_remap = true;
+  } else {
+    m_refine_remap = false;
```
bartgol, line 127: I would still add a check on whether a map file was provided. It may be that the user _wanted_ remap, and added the map file, but forgot to change the nudging input file. You may just issue a warning in this case, rather than error out...

bartgol (approved):

bartgol (approved):

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
bartgol: I don't recall, I would have to check.

components/eamxx/docs/user/coarse_nudging.md
```diff
@@ -0,0 +1,25 @@
...
+Instead, in EAMxx, it is possible to nudge from coarse data.
+This is done by remapping the coarse data provided by the user to the runtime physics grid of EAMxx.
+In order to enable nudging from coarse data, the user must provide nudging data at the coarse resolution desired and an appropriate     ncremap-compatible mapping file.
+
+## Example setup
+
+A user can produce coarse nudging data from running EAMxx or EAM at a ne30pg2 or any other applicable resolution.
+Additionally, several users in the E3SM projects have produced nudging data at the ne30pg2 resolution from the MERRA2 and ERA5 datasets.
+A limitation for now is that the nudging data must be provided explicitly, either as one file or as a list of files.
+This can be problematic for long list of files, but we are working on a solution to this problem.
```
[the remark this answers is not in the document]
bartgol: ah, yes, that's true, we probably are constrained in the length of a string, either in python, bash, or XML, not sure which one.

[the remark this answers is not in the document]
bartgol: I _really_ need to learn to read the _whole_ message before replying. This sparse-read feature of my brain is not working well.

Since you noticed the error at run time, it may not be an issue of bash/python/xml, but rather something else. There shouldn't be any restriction to a std::string size that could hurt us here though (there is a hard limit, but it's like in the billions of chars...

xyuan (dismissed):
this looks good, thanks.

components/eamxx/docs/user/coarse_nudging.md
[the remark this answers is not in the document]
AaronDonahue: does EKAT have a hard limit?  And can't we switch to a vector of strings rather than one long string to parse?

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
```diff
@@ -124,19 +171,61 @@ void Nudging::apply_weighted_tendency(Field& base, const Field& next, const Fiel
...
+  grid_tmp = grid_ext->clone(grid_ext->name(), true);
+  grid_tmp->reset_num_vertical_lev(m_num_levs);
+
+  // Declare the layouts for the helper fields (int: internal)
+  FieldLayout scalar2d_layout_mid { {LEV}, {m_num_levs} };
+  FieldLayout scalar3d_layout_mid { {COL,LEV}, {m_num_cols, m_num_levs} };
+
+  // Get the number of external cols on current rank
+  auto m_num_cols_ext = grid_ext->get_num_local_dofs();
+  // Declare the layouts for the helper fields (tmp: temporay))
```
AaronDonahue, line 214: not important, but misspelled "temporary"

AaronDonahue (commented):
Looks good to me, although I think there may still be some outstanding comments from @bartgol that need to be addressed.

components/eamxx/docs/user/coarse_nudging.md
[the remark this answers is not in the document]
AaronDonahue: I do not, but I believe our restart files are rather large.  So we should be able to handle O(100GB), @bartgol (?).

That being said, I wonder if there is a performance cost in reading in such single large files, rather than a number of smaller ones. I'm inclined to believe that the overhead of opening a file using SCORPIO outweighs opening a larger file.

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
```diff
@@ -523,8 +523,8 @@ void Nudging::run_impl (const double dt)
         apply_weighted_tendency(atm_state_field, int_state_field, nudging_weights_field, dt);
       } else if (m_refine_remap_vert_cutoff > 0.0) {
```
AaronDonahue, line 524: Would it make more sense to set the default `cutoff` value to something huge rather than 0.0?  In it's current implementation everything works because the `if` conditional checks if `cutoff>0.0` to begin with, but if someone messes that up the code would run fine and cutoff all nudging (so do nothing).  Maybe that's an argument for keeping it, so that if someone disrupts the code the baseline is not do any nudging at all...  I welcome push back.

components/eamxx/docs/user/coarse_nudging.md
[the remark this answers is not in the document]
AaronDonahue: Bigger picture, this still concerns me.  We should be this limited in the number of files we can feed to nudging.  I'm going to try to create a simple recreator and try to figure out the issue.

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
bartgol: One hand, I agree with Aaron, in the sense that not doing cutoff is the same as having the cutoff set to something very large (larger than any possible surface pressure), so having it set to 200k should be a good safeguard. However, that means you should change the check in this line (you can't check the cutoff itself), and you to check something else.

All in all, I think the current impl is fine.

Side comment though: @mahf708 I noticed going through the file that there are some indented blocks that use tabs instead of spaces. That can be messy to visualize, since different editors may use different tab length. If you have a second, can you sed `\t` with two spaces (which seems to be the indentation across the file)?

(btw: the `\t`->`  ` replacement thing is one of those things we ought to automate via actions...)

bartgol: E.g., starting at line 417, I see this
```c++
      for (int kk=fill_idx+1; kk<num_src_levs; ++kk) {
        const auto ipack = kk / mPack::n;
  const auto iidx  = kk % mPack::n;
        // Check if this index is masked
  if (!(ext_state_view_1d(ipack)[iidx]==var_fill_value)) {
    fill_value = ext_state_view_1d(ipack)[iidx];
  } else {
    ext_state_view_1d(ipack)[iidx] = fill_value;
  }
      }   
    }); 
```
It is all outside your mods, I was just hoping that we could piggy back that mod in... No big deal though.

bartgol: Ok, I thought this was a 5sec edit, but if it's raising other issues, forget about it. Feel free to force push to prev version...

bartgol: It still displays wrong on my editor, so I'd say reset HEAD~1 and use the force ...

bartgol: We actually _do_ have a `.clang_format` file in the eamxx folder. It's been there since the start of scream, but we never used it. Maybe we can revive it, and add a git hook (or gh action) to apply formatting when PR are merged...

bartgol (approved):

PR #2616, opened 2023-11-29
Add cmake utility to generate "PEM" like tests for EAMxx standalone

bartgol:
I had this branch sitting there for quite some time, so after last week's CreateUnitTest upgrades, I decided to spend a couple of hours to revive it and put up the PR.

The cmake code we used to generate the "np1_vs_npX" tests was not super obvious. This PR puts it in a CMake function, with the hope to hide some unnecessary detail, and make code maintenance as well as future tests development easier (and less bug prone).


components/eamxx/tests/coupled/dynamics_physics/homme_mam4xx_pg2/CMakeLists.txt
```diff
@@ -72,25 +72,16 @@ foreach (file IN ITEMS ${TEST_INPUT_FILES})
...
-    set (TEST_NAME "${BASE_TEST_NAME}_np${TEST_RANK_START}_vs_np${MPI_RANKS}_output_bfb")
-    add_test (NAME ${TEST_NAME}
-              COMMAND cmake -P ${CMAKE_BINARY_DIR}/bin/CprncTest.cmake ${SRC_FILE} ${TGT_FILE}
-              WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
-    set_tests_properties(${TEST_NAME} PROPERTIES
-      LABELS "dynamics;physics;mam4xx"
-      FIXTURES_REQUIRED "${FIXTURES_BASE_NAME}_np${MPI_RANKS}_omp1;${FIXTURES_BASE_NAME}_np1_omp1")
-  endforeach()
-endif()
+CompareNCFilesFamily (
```
bartgol, line 80: I was pondering whether I should add a higher level macro, called something like `CompareNCFilesFamilyPEM`, which takes `RANK_BEG`, `RANK_END`, and `RANK_INC` and internally creates the list of MAGIC_VALUES to then fwd to a call to `CompareNCFilesFamily`. It would not make the code _much_ smaller, but it would allow to not have to manually call `CreateRange(...)`.

Let me know if you think it would make sense to add it.

components/eamxx/tests/coupled/dynamics_physics/homme_mam4xx_pg2/CMakeLists.txt
```diff
@@ -72,25 +72,16 @@ foreach (file IN ITEMS ${TEST_INPUT_FILES})
...
+  TEST_META_NAME ${TEST_BASE_NAME}_npMPIRANKS_vs_np1
+  FILE_META_NAME ${TEST_BASE_NAME}_output.INSTANT.nsteps_x${NUM_STEPS}.npMPIRANKS.${RUN_T0}.nc
+  MAGIC_STRING "MPIRANKS"
+  MAGIC_VALUES "${MpiRanks}"
+  LABELS dynamics physics mam4xx PEM
```
bartgol, line 85: I used the label "PEM" to follow the CIME tests names.

components/eamxx/tests/coupled/dynamics_physics/homme_mam4xx_pg2/CMakeLists.txt
tcclevenger: Yeah, I think it makes sense to remove the manual call to `CreateRange(...)`. Makes the tests CMakeList clearer/simpler/easier when it comes to adding new tests. Not a big deal though if you decide you want to leave it out.

components/eamxx/tests/coupled/dynamics_physics/homme_mam4xx_pg2/CMakeLists.txt
tcclevenger: What does "PEM" stand for?

tcclevenger (approved):
I'll go ahead and approve and then you can add automerge if you decide not to add the PEM specific function.

components/eamxx/tests/coupled/dynamics_physics/homme_mam4xx_pg2/CMakeLists.txt
bartgol: This is what CIME has to say about it:
```
PEM    modified pe counts mpi bfb test (seq tests)
       do an initial run with default pe layout                               (suffix: base)
       do another initial run with modified pes (NTASKS_XXX => NTASKS_XXX/2)  (suffix: modpes)
       compare base and single_thread
```
If you think the name is too cryptic, I can use something like MPI-BFB or something along that way? Any idea?

tcclevenger: I guess I would prefer less cryptic, but maybe PEM isn't cryptic to the majority of the project? The current test name/cmake function name with comment is probably good enough.

bartgol:
@tcclevenger I implemented the shorthand version for the Mpi-bfb tests. I kept PEM as label, but being in a single place (the cmake function) is now easier to remove if we decide to to so later on.

tcclevenger (approved):

Issue #2602, opened 2023-11-20
``FieldLayout`` with new FieldTag


odiazib:
In order to use the ``AtmosphereInput class`` to read a variable from a netcdf file, one needs to pass the variable’s layout, which is an instantiation of the ``FieldLayout`` class. To construct a FieldLayout instance, we need to pass a FieldTag (an enum) and an integer. 

For example, 
```
FieldLayout scalar_refindex_real_sw_layout { {SWBAND, NREFINDEX_REAL},
                                     {nswbands,
                                      refindex_real} };
```
The limitation with the FieldTag enum defined [here]( https://github.com/E3SM-Project/scream/blob/7cb3c208c7a9367b55ee656de794533ed2e9dbfe/components/eamxx/src/share/field/field_tag.hpp#L27) is that one needs to add a new enum if the name of a dimension from a variable in the netcdf file is not listed in this part of the code. 

For example, for the layout of the variable `refindex_real_sw` with the following information:
```
double refindex_real_sw(sw_band, refindex_real) ;
                refindex_real_sw:long_name = "real refractive index of aerosol - shortwave" ;
```
One needs to add two new enums to the list of FieldTag. Note that in FiledTag ,there is also an enum with name of [ShortWaveBand](https://github.com/E3SM-Project/scream/blob/7cb3c208c7a9367b55ee656de794533ed2e9dbfe/components/eamxx/src/share/field/field_tag.hpp#L101) with name ``swband``, which is equivalent to the first dimension of `refindex_real_sw`, ``sw_band``, but one cannot use this tag because of the underscore symbol “_”.  

As the comment [here]( https://github.com/E3SM-Project/scream/blob/7cb3c208c7a9367b55ee656de794533ed2e9dbfe/components/eamxx/src/share/field/field_tag.hpp#L36), is there a better way or more flexible way of adding new field tags? Or can one have the option of using a std:string instead of FliedTag to set a new layout?
```
For example, 
FieldLayout scalar_refindex_real_lw_layout { {“sw_band”, “refindex_real”},
                                     {nlwbands,
                                      refindex_real} };  
```

bartgol:
I'm going to start working on this next.

bartgol:
[...]

PR #2620, opened 2023-12-01
Store tag names inside FieldLayout

### LINKED ISSUE #2601: AtmosphereInput and non-time dependent variables
I am getting the following error when using the ``AtmosphereInput class`` to read non-time dependent variables with more than two dimensions.  

``PIO: FATAL ERROR: Aborting... Extra outermost dimensions must have lengths of 1 (e3sm_repo/externals/scorpio/src/clib/pio_darray_int.c: 1358)``

As example, here are more details for a non-time dependent variable (``refindex_real_sw``) of two dimensions. 

NetCDF file is used by mam4 in e3sm v2: ``inputdata/atm/cam/physprops/mam4_mode2_rrtmg_c130628.nc``
This is the code that I am using: 

```
std::string tables_filename_mode4 = "mam4_mode4_rrtmg_c130628.nc";
ekat::ParameterList rrtmg_params;
rrtmg_params.set("Filename",tables_filename_mode4);
rrtmg_params.set<strvec_t>("Field Names",{"refindex_real_sw"});
// Note: Field tags were added to the code. 
FieldLayout scalar_refindex_real_lw_layout { {LWBAND, NREFINDEX_REAL},
                                     {nlwbands,
                                      refindex_real} };
std::map<std::string,FieldLayout>  layouts;
layouts.emplace("refindex_real_sw", scalar_refindex_real_sw_layout);
std::map<std::string,view_1d_host> host_views;
auto refindex_real_lw_host = view_2d_host("refrtablw_real_host", nlwbands, refindex_real);
 host_views["refindex_real_sw"]
   = view_1d_host(refindex_real_sw_host.data(),
   refindex_real_sw_host.size());
AtmosphereInput mode4_rrtmg(rrtmg_params, grid_, host_views, layouts);
mode4_rrtmg.read_variables();
mode4_rrtmg.finalize();
```

The above code compiled without issues, but produced the error mentioned above. 

Checking the PIO FATAL ERROR:[see here](https://github.com/E3SM-Project/scorpio/blob/de0b1ca2200f62c6eb5e3fd40147965409e97123/src/clib/pio_darray_int.c#L1357)

First, scorpio computes the number of extra dimensions with this equation:

``int num_extra_dims = (vdesc->record >= 0 && fndims > 1)? (ndims - (fndims - 1)) : (ndims - fndims);``
[see here](https://github.com/E3SM-Project/scorpio/blob/de0b1ca2200f62c6eb5e3fd40147965409e97123/src/clib/pio_darray_int.c#L1352
)

In this equation, if the condition ``(vdesc->record >= 0 && fndims > 1)`` is true, scorpio assumes that the variable is time dependent and if it is false, scorpio assumes a non-time dependent variable. 

For the example variable: `vdesc->record` is 1 and `fndims` is 2 and `ndims` is 2. Therefore, scorpio assumes that this variable is time dependent with one extra dimension with length of 16. 

To fix this issue, we need to set `vdesc->record` to a negative value for variables with more than two dimensions. Note that for variables with one dimension (`fndims=1`), the number of extra variables is computed by the right side of the conditional operation. 

Second, scorpio uses a different block of code to read non-time dependent variables; [see here](https://github.com/E3SM-Project/scorpio/blob/de0b1ca2200f62c6eb5e3fd40147965409e97123/src/clib/pio_darray_int.c#L1379C33-L1379C42). To execute this block of code this if statement `if (vdesc->record >= 0 && fndims > 1)` need to be false; [see here](https://github.com/E3SM-Project/scorpio/blob/de0b1ca2200f62c6eb5e3fd40147965409e97123/src/clib/pio_darray_int.c#L1361C17-L1361C54). Therefore, we need to set `vdesc->record` to a negative value to execute the correct block of code in scorpio. 

A temporal solution to fix this issue is to set `vdesc->record` to a negative number. For the code above, vdesc->record is set
[here](https://github.com/E3SM-Project/scream/blob/7cb3c208c7a9367b55ee656de794533ed2e9dbfe/components/eamxx/src/share/io/scream_scorpio_interface.F90#L1660) with `pio_atm_file%numRecs`. Thus, I did the following.

```
 else if (time_index == -999) then
            ! The user wants to read data set with non-time dependent variable
 call PIO_setframe(pio_atm_file%pioFileDesc,var%piovar,int(-1,kind=pio_offset_kind))
```
and 

``
mode4_rrtmg.read_variables(-1000);
``
with this temporal fix, ``AtmosphereInput`` is able to read a non-time dependent variables with more than 2 dimensions.

bartgol:
This PR fixes a few issues regarding `FieldTag`. The main goal was to address #2602 , but ended up doing a few more things. Here's a summary of the changes:

- `FieldLayout` stores names of the tags. If no names are provided at construction time, use the output of `e2str(FieldTag)`. Names can be reset after construction as well.
- Increased number of methods of `AbstractGrid` to create layouts. E.g., `grid->get_3d_scalar_layout_mid()`, which is more self-documenting (at the call site) than `grid->get_3d_scalar_layout(true)`.
- Prefer relying on `AbstractGrid`'s methods to create layout, rather than manually creating them.
- Get rid of all the rad-related `FieldTag`'s: `NGAS` (unused), `SWBND`, `LWBND`, `SWGPT`, `LWGPT`, `ISCCPTAU`, `ISCCPPRS`. The only reason for their existence was to provide a different return value for `e2str`, but now that tag names are stored in the field layout, they are no longer needed
- The `get_Nd_vector_layout` methods of `AbstractGrid` no longer accept a `FieldTag`, hard coding the vec tag to `CMP`. However, one can pass a specific name for the tag.
- Simplified a method in `VerticalRemapper`.
- Store type of layout inside `FieldLayout`. The utility `get_layout_type(tags)` was always called when the layout was available. So to make things easier, we store the type in the layout, and simply query the layout when the info is needed.

Fixes #2601 

@odiazib


tcclevenger (dismissed):
This will work for https://github.com/E3SM-Project/scream/pull/2619. I'll start testing with it.

odiazib:
@jeff-cohere

bartgol:
I'm closing this for now, while I think of a strategy to circumvent the IC read errors.

PR #2624, opened 2023-12-06
Remove compute-sanitizer --tool=synccheck from AT

tcclevenger:
The `synccheck` tool tests the use of CUDA synchronization primitives `__syncwarp()` and `__syncthreads()`, which we do not use directly in EAMxx code. Kokkos' CUDA back end obviously makes extensive use of these, but in the AT we are pretty much exclusively getting false and/or non-reproducible fails with this specific `compute-sanitizer` test. So I propose removing from the AT. This PR would leave it as a test def in `test_all_scream` so we could easily add back or run manually (`-t css`) if we wanted to.

Thoughts? Objections?

### DIFF
--- a/components/eamxx/scripts/jenkins/jenkins_common_impl.sh
+++ b/components/eamxx/scripts/jenkins/jenkins_common_impl.sh
@@ -121,7 +121,7 @@ if [ $skip_testing -eq 0 ]; then
       fi
 
       if [[ "$SCREAM_MACHINE" == "weaver" ]]; then
-        ./scripts/gather-all-data "./scripts/test-all-scream -t csm -t csr -t csi -t css ${TAS_ARGS}" -l -m $SCREAM_MACHINE
+        ./scripts/gather-all-data "./scripts/test-all-scream -t csm -t csr -t csi ${TAS_ARGS}" -l -m $SCREAM_MACHINE
         if [[ $? != 0 ]]; then
           fails=$fails+1;
           memcheck_fail=1


bartgol:
I'm not entirely sure we want to remove this. I think an error that synccheck could catch is something like this
```
  Kokkos::parallel_for(Kokkos::TeamVectorRange(team,N),[&](int k) {
     if (N==0) {
       team.team_barrier();
     }
  });
```
Even though we don't use the native CUDA sync functions, we can still use Kokkos syncronization utils badly...

One may argue that this kind of errors is unlikely to happen, since the code is obviously wrong. But the one above is of course an oversimplification; the code inside the if statement may call a utility funciton, which was designed to not be called from inside a TeamVectorRange... This scenario is not so remote, especially for the common physics utility functions, where we have 2+ versions of the same utility, each meant to be called from a different context (i.e., within or outside a pfor).

ambrad:
I agree with Luca. Are we sure the AT failures are false positives?

tcclevenger:
The last error looked like this
```
[weaver3:2484085:0:2484093] Caught signal 11 (Segmentation fault: address not mapped to object at address 0x70)
==== backtrace (tid:2484093) ====
 0  /projects/ppc64le-pwr9-rhel8/tpls/ucx/1.13.1/gcc/11.3.0/base/jx6zoju/lib/libucs.so.0(ucs_handle_error+0x354) [0x200005267f24]
 1  /projects/ppc64le-pwr9-rhel8/tpls/ucx/1.13.1/gcc/11.3.0/base/jx6zoju/lib/libucs.so.0(+0x380b0) [0x2000052680b0]
 2  /projects/ppc64le-pwr9-rhel8/tpls/ucx/1.13.1/gcc/11.3.0/base/jx6zoju/lib/libucs.so.0(+0x38544) [0x200005268544]
 3  linux-vdso64.so.1(__kernel_sigtramp_rt64+0) [0x2000000604d8]
 4  /projects/ppc64le-pwr9-rhel8/tpls/cuda/11.8.0/gcc/11.3.0/base/ztdfrze/compute-sanitizer/libsanitizer-collection.so(+0x450a2c) [0x2000004d0a2c]
 5  /projects/ppc64le-pwr9-rhel8/tpls/cuda/11.8.0/gcc/11.3.0/base/ztdfrze/compute-sanitizer/libsanitizer-collection.so(+0x441b34) [0x2000004c1b34]
 6  /projects/ppc64le-pwr9-rhel8/tpls/cuda/11.8.0/gcc/11.3.0/base/ztdfrze/compute-sanitizer/libsanitizer-collection.so(+0x43ee88) [0x2000004bee88]
 7  /projects/ppc64le-pwr9-rhel8/tpls/cuda/11.8.0/gcc/11.3.0/base/ztdfrze/compute-sanitizer/libsanitizer-collection.so(+0x450f14) [0x2000004d0f14]
 8  /projects/ppc64le-pwr9-rhel8/tpls/cuda/11.8.0/gcc/11.3.0/base/ztdfrze/compute-sanitizer/libsanitizer-collection.so(+0x450f60) [0x2000004d0f60]
 9  /projects/ppc64le-pwr9-rhel8/tpls/cuda/11.8.0/gcc/11.3.0/base/ztdfrze/compute-sanitizer/libsanitizer-collection.so(+0x44b698) [0x2000004cb698]
10  /projects/ppc64le-pwr9-rhel8/tpls/cuda/11.8.0/gcc/11.3.0/base/ztdfrze/compute-sanitizer/libsanitizer-collection.so(+0x44bc14) [0x2000004cbc14]
11  /projects/ppc64le-pwr9-rhel8/tpls/cuda/11.8.0/gcc/11.3.0/base/ztdfrze/compute-sanitizer/libsanitizer-collection.so(+0x44b154) [0x2000004cb154]
12  /projects/ppc64le-pwr9-rhel8/tpls/cuda/11.8.0/gcc/11.3.0/base/ztdfrze/compute-sanitizer/libsanitizer-collection.so(+0x1c36b0) [0x2000002436b0]
13  /projects/ppc64le-pwr9-rhel8/tpls/cuda/11.8.0/gcc/11.3.0/base/ztdfrze/compute-sanitizer/libsanitizer-collection.so(+0x6f92a0) [0x2000007792a0]
14  /lib64/glibc-hwcaps/power9/libpthread-2.28.so(+0x9678) [0x2000033f9678]
15  /lib64/glibc-hwcaps/power9/libc-2.28.so(clone+0x74) [0x200003d68738]
=================================
========= Error: process didn't terminate successfully
=========     The application may have hit an error when dereferencing Unified Memory from the host. Please rerun the application under cuda-gdb or a host debugger to catch host side errors.
========= Target application returned an error
========= ERROR SUMMARY: 0 errors
--------------------------------------------------------------------------
Primary job  terminated normally, but 1 process returned
a non-zero exit code. Per user-direction, the job has been aborted.
--------------------------------------------------------------------------
--------------------------------------------------------------------------
mpiexec detected that one or more processes exited with non-zero status, thus causing
the job to be terminated. The first process to do so was:

  Process name: [[11151,1],0]
  Exit code:    1
--------------------------------------------------------------------------

```

If I remember correctly this is how they have all gone. At least searching slack eamxx channel for "synccheck", the last 3 have all been this type of failure.
https://my.cdash.org/test/107625543
https://my.cdash.org/test/100458179
https://my.cdash.org/test/108579964

This type of fail can also happen in memcheck and the others, but seems to be much more common in synccheck. That said, I expected to search and find more times this happened, so maybe this is not as much of an issue as I thought. I don't mind closing the PR. Maybe I'll see if I can trigger an error using Luca's idea.

bartgol:
Yeah, I think that error points to some issue unrelated to threads synchronization. And google-ing a bit, I see other folks struggling with this sort of error when using compute-sanitizer. So maybe it's an issue in the tool itself

ambrad:
I think we should keep this tool running and just be aware that this seg fault issue might be a problem for a while.

PR #2626, opened 2023-12-07
Fix 2 bugs in how IO (output) handles AVG pressure sliced vars

AaronDonahue:
Fix two bugs in IO regarding average tracking and sliced output.
    
    This commit fixes two bugs in how we handle average snap tracking
    in cases where sliced output is one of the fields.
    
      1. We were not automatically tracking the number of filled values
         in cases where fields where output on a set pressure level.
         This would result in some columns being the average of the Fill
         Value and real values.
    
      2. When tracking the fill values we were using a single tracking
         field for all outputs on the same layout.  This would result
         in sliced outputs causing other 2D fields (ncol) to also be
         masked, even when they shouldn't have been.  The most direct
         example would be surface pressure.

### DIFF
--- a/components/eamxx/src/share/io/scorpio_output.cpp
+++ b/components/eamxx/src/share/io/scorpio_output.cpp
@@ -846,7 +846,7 @@ void AtmosphereOutput::register_views()
   reset_dev_views();
 }
 /* ---------------------------------------------------------- */
-void AtmosphereOutput::set_avg_cnt_tracking(const std::string& name, const std::string& name_ext, const FieldLayout& layout)
+void AtmosphereOutput::set_avg_cnt_tracking(const std::string& name, const std::string& avg_cnt_suffix, const FieldLayout& layout)
 {
     // Make sure this field "name" hasn't already been regsitered with avg_cnt tracking.
     // Note, we check this because some diagnostics need to have their own tracking which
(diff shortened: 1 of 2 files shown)


AaronDonahue:
ugg.... I think this got messed up in the log history when I squashed two commits.  Let me fix.

components/eamxx/src/share/io/scorpio_output.cpp
```diff
@@ -839,10 +839,28 @@ void AtmosphereOutput::register_views()
...
+      return;
+    }
+    // Now create and store a dev view to track the averaging count for this layout (if we are tracking)
+    // We don't need to track average counts for files that are not tracking the time dim
+    const auto size = layout.size();
     const auto tags = layout.tags();
     auto lt = get_layout_type(tags);
     if (m_add_time_dim && m_track_avg_cnt) {
-      std::string avg_cnt_name = "avg_count_" + e2str(lt);
+      std::string avg_cnt_name = "avg_count_" + e2str(lt) + "_" + name_ext;
```
bartgol, line 863: If `name_ext=""`, which is the case for all but the `FieldAtPressureLevel` diagnostic, you get two underscores in the name (one here, and one from the for loop below). Can we jsut do `+ e2str(lt) + name_ext`, and put the underscore in the `name_ext` when it's computed (if not empty)?

bartgol: Also, maybe `name_ext` is not super self-explanatory. How about "avg_cnt_suffix"?

bartgol: Also, since we add the tag names to `avg_cnt_name` below, we can keep the name shorter and not add `e2str(lt)` maybe? It doesn't add any info not already in the list of tags...

bartgol (commented):
I got some appearance comments. Feel free to disregard them.

components/eamxx/src/share/io/scorpio_output.cpp
AaronDonahue: all good points.  I'll make the changes..

bartgol (approved):

PR #2628, opened 2023-12-08
Reorganize infrastructure code

bartgol:
First of all, don't get discouraged by the number of files changed in this PR. Most mods are just changing the location of files included.

The main goal of this PR is to reduce the spaghetti-ness of the share folder, which has grown to a very large amount of files. In particular, I tried to split it into smaller libs, which should not have cyclic dependencies (hopefully). The libs are the following (the real name of the lib has the `eamxx_` prefix), with their deps listed in parentheses (they should depend only on stuff above).

- `core` (`ekat`): the config defs and basic types of eamxx
- `utils` (`core`, `gptl`): some common utils
- `field` (`core`, `utils`): the field stack (except for field manager)
- `grid` (`field`): the grid stack (no remappers, not grid manager)
- `scorpio_interface` (`utils`, `pioc`, `piof`): all low-level interfaces to scorpio (performing I/O calls with low-level data structures)
- `remap` (`grid`, `scorpio_interface`): all the remappers infrastructure
- `property_checks` (`'field`, `grid`, `utils`): the property checks classes
- `managers` (`field`, `grid`, `remap`): the grid and field manager classes
- `atm_process` (`field`, `managers`, `util`, `property_check`): the atm process interfaces
- `io` (`scorpio_interface`, `atm_process`, `remap`, `managers`): the high-level input/output classes (performing I/O calls with eamxx data structures)
- `fvphyshack` (`remap`): hacks to handle pg2 grids in the ad (hopefully to go away at some point)
- `diagnostics` (`atm_process`): all diagnostic classes
- `unit_test_support` (`managers`): the mesh-free GM, the test session, the random test setup.

All the tests for these libs (if any) are in the homoninous folder in `share/tests`. In particular, io and diags tests are no longer inside those folders.

The `scream_share` lib has also been renamed to `eamxx_share` and it is an INTERFACE lib, linking to all the above (except for `unit_test_support` and `diagnostics`, but _including_ `io`).

Note: I moved the time interpolation utility in the `share/io` folder, along with other IO stuff. That's because it is mainly used to interpolate data from a set of input files. At some point, we may as well call it `InputFileTimeInterpolation` or something...


bartgol:
> Definitely in favor of this change.
Ok, then I may move fwd with this, although not a big priority right now.

> Only question: there are still a few "scream" in file names, `cmake/ScreamUtils.cmake` and `share/util/scream_bfbhash.hpp` are ones I found. Should we change these to "eamxx"?
I don't know if changing all names should be done in this PR. I did change _some_ of them, but maybe I should hold back, and we could do a single scream->eamxx PR when there's not much work going on (to avoid hard-to-fix conflicts). OTOH, this PR may already create conflicts for other development, so maybe we should kill two birds with a stone?

bartgol:
Closing, becasue this has fallen so much behind that it's too complicated to sync up with master. If we decide to do something like this again, it will be from scratch, and probably done in steps.

PR #2631, opened 2023-12-08
Utility for applying random perturbation to IC

tcclevenger:
Creates a `perturb()` in `field_utils.cpp` which applies a random perturbation
```
field_view(i0,...,k) *= (1+pertval) when level_mask(k)=true
```
The value `pertval` is randomly selected from a user defined RNG.

Uses in AD: Directly after ICs are loaded, perturb fields (if requested). Introduces some new parameters under `initial_conditions`:
  - `perturbed_fields`: set of fields to be perturbed
  - `perturbation_limit`: defines range [-perturbed_limit, perturbed_limit] for RNG
  - `perturbation_pressure_lower_bound`: lower bound s.t. perturbed field at k if `ref_pres(k)>perturbation_pressure_lower_bound`
  - `perturbation_random_seed`: seed for RNG, allows us to recreate runs

We need this for DP-SCREAM, where we perturb `T_mid`

### DIFF
--- a/components/eamxx/src/control/atmosphere_driver.cpp
+++ b/components/eamxx/src/control/atmosphere_driver.cpp
@@ -1293,15 +1293,33 @@ void AtmosphereDriver::set_initial_conditions ()
                      "the GLL grid, but no Physics GLL FieldManager was defined.\n");
     const auto& fm = m_field_mgrs.at("Physics GLL");
 
-    // Setup RNG. If no random seed is requested in
-    // initial_conditions params, generate one using rand().
-    const auto seed = ic_pl.get<int>("perturbation_random_seed", rand());
-    std::mt19937_64 engine(seed);
+    // Setup RNG. There are two relevant params: generate_perturbation_random_seed and
+    // perturbation_random_seed. We have 3 cases:
+    //   1. Parameter generate_perturbation_random_seed is set true, assert perturbation_random_seed
+    //      is not given and generate a random seed using std::rand() to get an integer random value.
+    //   2. Parameter perturbation_random_seed is given, use this value for the seed.
+    //   3. Parameter perturbation_random_seed is not given and generate_perturbation_random_seed is
+    //      not given, use 0 as the random seed.
+    // Case 3 is considered the default (using seed=0).
+    int seed;
+    if (ic_pl.get<bool>("generate_perturbation_random_seed", false)) {
+      EKAT_REQUIRE_MSG(not ic_pl.isParameter("perturbation_random_seed"),
+                       "Error! Param generate_perturbation_random_seed=true, and "
+                       "a perturbation_random_seed is given. Only one of these can "
+                       "be defined for a simulation.\n");
+      std::srand(std::time(nullptr));
+      seed = std::rand();
+    } else {
+      seed = ic_pl.get<int>("perturbation_random_seed", 0);
+    }
     m_atm_logger->info("      For IC perturbation, random seed: "+std::to_string(seed));
+    std::mt19937_64 engine(seed);
 
-    // Get perturbation limit. Defines a range [-perturbation_limit, perturbation_limit]
-    // for which the perturbation value will be randomly generated from.
-    const auto perturbation_limit = ic_pl.get<Real>("perturbation_limits", 0.001);
+    // Get perturbation limit. Defines a range [1-perturbation_limit, 1+perturbation_limit]
+    // for which the perturbation value will be randomly generated from. Create a uniform
+    // distribution for this range.
+    const auto perturbation_limit = ic_pl.get<Real>("perturbation_limit", 0.001);
+    std::uniform_real_distribution<Real> pdf(1-perturbation_limit, 1+perturbation_limit);
 
     // Define a level mask using reference pressure and the perturbation_minimum_pressure parameter.
     // This mask dictates which levels we apply a perturbation.
@@ -1325,9 +1343,9 @@ void AtmosphereDriver::set_initial_conditions ()
                        "  - Grid:  "+fm->get_grid()->name()+"\n");
 
       auto field = fm->get_field(fname);
-      std::uniform_real_distribution<Real> pdf(-1.0*perturbation_limit, perturbation_limit);
-      perturb(field, engine, pdf, pressure_mask);
+      perturb(field, engine, pdf, seed, pressure_mask, fm->get_grid()->get_dofs_gids());
     }
+
     m_atm_logger->info("    [EAMxx] Adding random perturbation to ICs ... done!");
   }
 
--- a/components/eamxx/src/control/atmosphere_driver.cpp
+++ b/components/eamxx/src/control/atmosphere_driver.cpp
@@ -1303,17 +1303,17 @@ void AtmosphereDriver::set_initial_conditions ()
     // for which the perturbation value will be randomly generated from.
     const auto perturbation_limit = ic_pl.get<Real>("perturbation_limits", 0.001);
 
-    // Define a level mask using reference pressure and the pressure_lower_bound parameter.
+    // Define a level mask using reference pressure and the perturbation_minimum_pressure parameter.
     // This mask dictates which levels we apply a perturbation.
     const auto gll_grid = m_grids_manager->get_grid("Physics GLL");
     gll_grid->get_geometry_data("hyam").sync_to_host(), gll_grid->get_geometry_data("hybm").sync_to_host();
     const auto hyam = gll_grid->get_geometry_data("hyam").get_view<const Real*, Host>();
     const auto hybm = gll_grid->get_geometry_data("hybm").get_view<const Real*, Host>();
     constexpr auto ps0 = physics::Constants<Real>::P0;
-    const auto pressure_lower_bound = ic_pl.get<Real>("perturbation_pressure_lower_bound", 1050.0);
+    const auto min_pressure = ic_pl.get<Real>("perturbation_minimum_pressure", 1050.0);
     auto pressure_mask = [&] (const int ilev) {
       const auto pref = (hyam(ilev)*ps0 + hybm(ilev)*ps0)/100; // Reference pressure ps0 is in Pa, convert to millibar
-      return pref > pressure_lower_bound;
+      return pref > min_pressure;
     };
 
     // Loop through fields and apply perturbation.
--- a/components/eamxx/src/control/atmosphere_driver.cpp
+++ b/components/eamxx/src/control/atmosphere_driver.cpp
@@ -1324,13 +1324,12 @@ void AtmosphereDriver::set_initial_conditions ()
     // Define a level mask using reference pressure and the perturbation_minimum_pressure parameter.
     // This mask dictates which levels we apply a perturbation.
     const auto gll_grid = m_grids_manager->get_grid("Physics GLL");
-    gll_grid->get_geometry_data("hyam").sync_to_host(), gll_grid->get_geometry_data("hybm").sync_to_host();
-    const auto hyam = gll_grid->get_geometry_data("hyam").get_view<const Real*, Host>();
-    const auto hybm = gll_grid->get_geometry_data("hybm").get_view<const Real*, Host>();
+    const auto hyam_h = gll_grid->get_geometry_data("hyam").get_view<const Real*, Host>();
+    const auto hybm_h = gll_grid->get_geometry_data("hybm").get_view<const Real*, Host>();
     constexpr auto ps0 = physics::Constants<Real>::P0;
     const auto min_pressure = ic_pl.get<Real>("perturbation_minimum_pressure", 1050.0);
     auto pressure_mask = [&] (const int ilev) {
-      const auto pref = (hyam(ilev)*ps0 + hybm(ilev)*ps0)/100; // Reference pressure ps0 is in Pa, convert to millibar
+      const auto pref = (hyam_h(ilev)*ps0 + hybm_h(ilev)*ps0)/100; // Reference pressure ps0 is in Pa, convert to millibar
       return pref > min_pressure;
     };
 
--- a/components/eamxx/src/control/atmosphere_driver.cpp
+++ b/components/eamxx/src/control/atmosphere_driver.cpp
@@ -2,6 +2,8 @@
 #include "control/atmosphere_surface_coupling_importer.hpp"
 #include "control/atmosphere_surface_coupling_exporter.hpp"
 
+#include "physics/share/physics_constants.hpp"
+
 #include "share/atm_process/atmosphere_process_group.hpp"
 #include "share/atm_process/atmosphere_process_dag.hpp"
 #include "share/field/field_utils.hpp"
@@ -29,6 +31,7 @@
 #endif
 
 #include <fstream>
+#include <random>
 
 namespace scream {
 
@@ -1001,7 +1004,6 @@ void AtmosphereDriver::set_initial_conditions ()
       if (ic_pl.isType<double>(fname) or ic_pl.isType<std::vector<double>>(fname)) {
         // Initial condition is a constant
         initialize_constant_field(fid, ic_pl);
-        fields_inited[grid_name].push_back(fname);
 
         // Note: f is const, so we can't modify the tracking. So get the same field from the fm
         auto f_nonconst = m_field_mgrs.at(grid_name)->get_field(fid.name());
@@ -1009,11 +1011,11 @@ void AtmosphereDriver::set_initial_conditions ()
       } else if (ic_pl.isType<std::string>(fname)) {
         // Initial condition is a string
         ic_fields_to_copy.push_back(fid);
-        fields_inited[grid_name].push_back(fname);
       } else {
         EKAT_ERROR_MSG ("ERROR: invalid assignment for variable " + fname + ", only scalar "
                         "double or string, or vector double arguments are allowed");
       }
+      fields_inited[grid_name].push_back(fname);
     } else if (fname == "phis" or fname == "sgh30") {
       // Both phis and sgh30 need to be loaded from the topography file
       auto& this_grid_topo_file_fnames = topography_file_fields_names[grid_name];
@@ -1032,6 +1034,7 @@ void AtmosphereDriver::set_initial_conditions ()
           EKAT_ERROR_MSG ("Error! Requesting phis on an unknown grid: " + grid_name + ".\n");
         }
         this_grid_topo_eamxx_fnames.push_back(fname);
+	fields_inited[grid_name].push_back(fname);
       } else if (fname == "sgh30") {
         // The eamxx field "sgh30" is called "SGH30" in the
         // topography file and is only available on the PG2 grid.
@@ -1040,6 +1043,7 @@ void AtmosphereDriver::set_initial_conditions ()
                         " topo file only has sgh30 for Physics PG2.\n");
         topography_file_fields_names[grid_name].push_back("SGH30");
         topography_eamxx_fields_names[grid_name].push_back(fname);
+	fields_inited[grid_name].push_back(fname);
       }
     } else if (not (fvphyshack and grid_name == "Physics PG2")) {
       // The IC file is written for the GLL grid, so we only load
@@ -1051,6 +1055,7 @@ void AtmosphereDriver::set_initial_conditions ()
         // If this field is the parent of other subfields, we only read from file the subfields.
         if (not ekat::contains(this_grid_ic_fnames,fname)) {
           this_grid_ic_fnames.push_back(fname);
+	  fields_inited[grid_name].push_back(fname);
         }
       } else if (fvphyshack and grid_name == "Physics GLL") {
         // [CGLL ICs in pg2] I tried doing something like this in
@@ -1064,10 +1069,10 @@ void AtmosphereDriver::set_initial_conditions ()
           const auto& fname = fid.name();
           if (ic_pl.isParameter(fname) and ic_pl.isType<double>(fname)) {
             initialize_constant_field(fid, ic_pl);
-            fields_inited[grid_name].push_back(fname);
           } else {
             this_grid_ic_fnames.push_back(fname);
           }
+	  fields_inited[grid_name].push_back(fname);
         }
       }
     }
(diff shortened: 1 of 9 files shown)


components/eamxx/src/control/atmosphere_driver.cpp
```diff
@@ -1071,10 +1076,10 @@ void AtmosphereDriver::set_initial_conditions ()
           const auto& fname = fid.name();
           if (ic_pl.isParameter(fname) and ic_pl.isType<double>(fname)) {
             initialize_constant_field(fid, ic_pl);
-            fields_inited[grid_name].push_back(fname);
           } else {
             this_grid_ic_fnames.push_back(fname);
           }
+	  fields_inited[grid_name].push_back(fname);
```
tcclevenger, line 1082: We were only using `fields_inited` for fields set as params, but I think it makes sense to use for any field we init, including from file. I use this to verify random perturbation was requested from an appropriate field.

components/eamxx/src/control/intensive_observation_period.cpp
```diff
@@ -133,7 +133,6 @@ IntensiveObservationPeriod(const ekat::Comm& comm,
   if (not m_params.isParameter("iop_nudge_tq_low"))     m_params.set<Real>("iop_nudge_tq_low",     1050);
   if (not m_params.isParameter("iop_nudge_tq_high"))    m_params.set<Real>("iop_nudge_tq_high",    0);
   if (not m_params.isParameter("iop_nudge_tscale"))     m_params.set<Real>("iop_nudge_tscale",     10800);
-  if (not m_params.isParameter("iop_perturb_high"))     m_params.set<Real>("iop_perturb_high",     1050);
```
tcclevenger, line 136: In F90 this was a IOP param, no longer need it since AD takes care of it.

components/eamxx/src/share/field/field_utils.hpp
```diff
@@ -43,6 +43,41 @@ void randomize (const Field& f, Engine& engine, PDF&& pdf)
   impl::randomize<ST>(f,engine,pdf);
 }
 
+// Compute a random perturbation of a field for all view entries
+// field_view(:, k) when level_mask(k)==true. The field
+// must have level midpoint tag as last dimension, and level mask
+// should have size of last field dimension.
+template<typename Engine, typename PDF, typename MaskType>
+void perturb (const Field& f, Engine& engine, PDF&& pdf,
+	      const MaskType& level_mask)
```
tcclevenger, line 52: I chose to use a level mask. I could make it even more generic and make a mask for every entry. @PeterCaldwell can you envision any use case where we want to perturb based on anything but level index?

components/eamxx/tests/coupled/dynamics_physics/homme_shoc_cld_spa_p3_rrtmgp_dp/input.yaml
```diff
@@ -22,6 +22,10 @@ initial_conditions:
   surf_sens_flux: 0.0
   precip_liq_surf_mass: 0.0
   precip_ice_surf_mass: 0.0
+  perturbed_fields: [T_mid]
+  perturbation_limit: 0.001
+  perturbation_pressure_lower_bound: 900.0 # in milibar
+  perturbation_random_seed: 1805289383
```
tcclevenger, line 28: @PeterCaldwell Do you see needed a different perturbation range (defined by `perturbation_limit`) for each field? Currently I just have it set to a single value used for every field.

tcclevenger:
@bartgol I know this will conflict with the IC work you are doing. Feel free to request changes that will fit in with your design/make your life easier.

bartgol:
Quick comment for now. I will read in detail on Monday. And yes, it has quite some conflict with the IC refactor, but I can deal with it.

Why not perturbing fields as
```
Field perturb (...);
randomize(perturb, engine, pdf);
f.update(perturb,1.0,1.0);
```
If you only want to perturb certain levels, you can do
```
Field perturb (...); // with layout stripped of levels
for (int ilev=start; ilev<end; ++ilev) {
   randomize(perturb,engine,pdf);
   f.subfield(lev_dim, ilev).update(perturb,1.0,1.0);
}
```
It may not be as efficient (you need to create the perturb field for each field), but it definitely reduces the amount of code needed.

tcclevenger:
@bartgol I like the field idea. I'll work on implementing it and then re-request a review when I'm done.

components/eamxx/src/control/atmosphere_driver.cpp
PeterCaldwell: Super minor point - if you're using fields_inited for everything, why not call fields_inited 1x just after all these if/elseif conditions instead of repeating the same line in each if/elseif case?

components/eamxx/src/control/atmosphere_driver.cpp
```diff
@@ -1283,6 +1288,56 @@ void AtmosphereDriver::set_initial_conditions ()
...
+                     "the GLL grid, but no Physics GLL FieldManager was defined.\n");
+    const auto& fm = m_field_mgrs.at("Physics GLL");
+
+    // Setup RNG. If no random seed is requested in
+    // initial_conditions params, generate one using rand().
+    const auto seed = ic_pl.get<int>("perturbation_random_seed", rand());
+    std::mt19937_64 engine(seed);
+    m_atm_logger->info("      For IC perturbation, random seed: "+std::to_string(seed));
+
+    // Get perterbation limit. Defines a range [-perturbation_limit, perturbation_limit]
```
PeterCaldwell, line 1309: Typo "perterbation"

components/eamxx/src/control/atmosphere_driver.cpp
```diff
@@ -1283,6 +1288,56 @@ void AtmosphereDriver::set_initial_conditions ()
...
+    // Define a level mask using reference pressure and the pressure_lower_bound parameter.
+    // This mask dictates which levels we apply a perturbation.
+    const auto gll_grid = m_grids_manager->get_grid("Physics GLL");
+    gll_grid->get_geometry_data("hyam").sync_to_host(), gll_grid->get_geometry_data("hybm").sync_to_host();
+    const auto hyam = gll_grid->get_geometry_data("hyam").get_view<const Real*, Host>();
+    const auto hybm = gll_grid->get_geometry_data("hybm").get_view<const Real*, Host>();
+    constexpr auto ps0 = physics::Constants<Real>::P0;
+    const auto pressure_lower_bound = ic_pl.get<Real>("perturbation_pressure_lower_bound", 1050.0);
+    auto pressure_mask = [&] (const int ilev) {
+      const auto pref = (hyam(ilev)*ps0 + hybm(ilev)*ps0)/100; // Reference pressure ps0 is in Pa, convert to milibar
```
PeterCaldwell, line 1322: typo in comment: milibar->millibar

PeterCaldwell: More importantly though - why are we reconstructing hydrostatic pressure from hyam, hybm? Do we not have the full pressure at this point? I'm a bit worried about this because we hope to eventually start using non-hydrostatic pressure for physics, so it would be better to just use pressure where we can. But if doing that isn't convenient right now, we can just leave this as something to fix if/when we switch to full pressure. Or maybe we never change this because it's just in initialization and is approximately correct anyways.

components/eamxx/src/control/atmosphere_driver.cpp
```diff
@@ -1283,6 +1288,56 @@ void AtmosphereDriver::set_initial_conditions ()
...
+    for (size_t f=0; f<perturbed_fields.size(); ++f) {
+      const auto fname = perturbed_fields[f];
+      EKAT_REQUIRE_MSG(ekat::contains(fields_inited[fm->get_grid()->name()], fname),
+                       "Error! Attempting to apply perturbation to field not in initial_conditions.\n"
+                       "  - Field: "+fname+"\n"
+                       "  - Grid:  "+fm->get_grid()->name()+"\n");
+
+      auto field = fm->get_field(fname);
+      std::uniform_real_distribution<Real> pdf(-1.0*perturbation_limit, perturbation_limit);
+      perturb(field, engine, pdf, pressure_mask);
```
PeterCaldwell, line 1336: Can we specify the random seed? Doing so would allow us to BFB repeat a particular simulation if needed, which would be nice. Like, the random seed for each run could be either a user-input or randomly generated. If it is alwayas printed to the log file, we could repeat any run we want. Though I'm not sure if we'd need a different seed for each variable we're randomizing or even each col and lev within a field... that would be a hassle.

components/eamxx/src/control/atmosphere_driver.cpp
PeterCaldwell: More importantly - awesome to print the random seed to the log file. Could we add the option for a user to specify the random seed for a given run? Then we could repeat a simulation if it did something weird or we didn't get the output we needed or something.

components/eamxx/src/control/atmosphere_driver.cpp
tcclevenger: This is just how EAM constructed it. At this point we would have the pressure from the IC file if that is better to use.

components/eamxx/src/control/atmosphere_driver.cpp
tcclevenger: The user can request put `perturbation_random_seed:` in the input.yaml. But I think the field values will only line up if the user also requests the same fields to be perturbed.

components/eamxx/src/share/field/field_utils.hpp
PeterCaldwell: > can you envision any use case where we want to perturb based on anything but level index?
No. Maybe someday someone will come up with a use case, but let's let them add that functionality if/when they do in order to make our code simpler now.

components/eamxx/src/control/atmosphere_driver.cpp
tcclevenger: Yeah, maybe we can. There are some fields that are skipped, but that should be easy to account for. I'll look at it.

components/eamxx/src/control/atmosphere_driver.cpp
```diff
@@ -1283,6 +1288,56 @@ void AtmosphereDriver::set_initial_conditions ()
...
+    const auto perturbation_limit = ic_pl.get<Real>("perturbation_limits", 0.001);
+
+    // Define a level mask using reference pressure and the pressure_lower_bound parameter.
+    // This mask dictates which levels we apply a perturbation.
+    const auto gll_grid = m_grids_manager->get_grid("Physics GLL");
+    gll_grid->get_geometry_data("hyam").sync_to_host(), gll_grid->get_geometry_data("hybm").sync_to_host();
+    const auto hyam = gll_grid->get_geometry_data("hyam").get_view<const Real*, Host>();
+    const auto hybm = gll_grid->get_geometry_data("hybm").get_view<const Real*, Host>();
+    constexpr auto ps0 = physics::Constants<Real>::P0;
+    const auto pressure_lower_bound = ic_pl.get<Real>("perturbation_pressure_lower_bound", 1050.0);
```
PeterCaldwell, line 1320: kind of a dumb comment, but I initially understood "pressure_lower_bound" as saying you're going to use all pressures above a certain geometric height level instead of all pressures larger (i.e. closer to the surface) than this pressure level. Maybe change this name to min_perturbed_pressure or something?

components/eamxx/src/share/tests/field_utils.cpp
```diff
@@ -172,6 +173,59 @@ TEST_CASE("utils") {
...
+    auto check_level = [&] (const int ilev, const Real val) {
+      if (ilev < nlevs-3) REQUIRE(val == 1);
+      else REQUIRE((2 <= val && val <= 3));
+    };
+    for (int icol=0; icol<ncols; ++icol) {
+      for (int icmp=0; icmp<ncmps; ++icmp) {
+        for (int ilev=0; ilev<nlevs; ++ilev) {
+          if (icol==0 && icmp==0) check_level(ilev, v1(ilev));
+          if (icmp==0) check_level(ilev, v2(icol,ilev));
+          check_level(ilev, v3(icol,icmp,ilev));
```
PeterCaldwell, line 225: Maybe add a check that if you have the same random seed, you get the same answer 2x in a row and if you have a different random seed, you get 2 different answers? 

Also make sure that different columns get different values for a given model level?

components/eamxx/tests/coupled/dynamics_physics/homme_shoc_cld_spa_p3_rrtmgp_dp/input.yaml
PeterCaldwell: We probably do need different perturbations for different variables because T has values of ~300 K while nc has units of something like 1e5 #/kg. Maybe just a different scaling factor for each variable?

bartgol:
> @bartgol I like the field idea. I'll work on implementing it and then re-request a review when I'm done.

I just noticed that `randomize` assumes contiguous view. We may want to replace stuff like
```
    case 2:
      {   
        auto v = f.template get_view<ST**,Host>();
        for (int i=0; i<fl.dim(0); ++i) {
          for (int j=0; j<fl.dim(1); ++j) {
            v(i,j) = pdf(engine);
        }}
      }  
```
with something like
```
    case 2:
      { 
        auto v = f.template get_strided_view<ST**,Host>();
        for (int i=0; i<fl.dim(0); ++i) {
          for (int j=0; j<fl.dim(1); ++j) {
            v(i,j) = pdf(engine);
        }}
      }  
 ```
 We don't use `randomize` often, usually only in tests, so strided views are not a big deal. But if we want to be fancy, we can add more code, and do something like
 ```
 if (f.get_header().get_alloc_properties().contiiguous()) {
   do_randomize (f.get_view<ST**,Host>(),pdf,engine);
 } else {
   do_randomize(f.get_strided_view<ST**,Host>(),pdf,engine);
 }
 ```
 where `do_randomize` is templated on the view type, so that we avoid writing the same loops twice.

File: components/eamxx/src/control/atmosphere_driver.cpp (hunk 6 above)
Comment (PeterCaldwell) [reply components/eamxx/src/control/atmosphere_driver.cpp:1322]: Using the actual pressure from the IC file seems better BUT would give different number of masked levs for each different col. Which could cause weird stuff to happen if, for example, we only want to perturb below 700mb but we have a Mt Everest grid cell whose surface is higher (in z) than 700mb so no cells get perturbed at all. I think that's fine for our uses, but I'm curious what @bogensch thinks.

File: components/eamxx/src/control/atmosphere_driver.cpp (hunk 5 above)
Comment (PeterCaldwell) [reply components/eamxx/src/control/atmosphere_driver.cpp:1309]: >The user can request put perturbation_random_seed: in the input.yaml. But I think the field values will only line up if the user also requests the same fields to be perturbed.

That's fine. Someone who asks for different fields to be perturbed and expects to get the same answer has bigger problems.

File: components/eamxx/src/control/atmosphere_driver.cpp (hunk 1 above)
Comment (PeterCaldwell) [reply components/eamxx/src/control/atmosphere_driver.cpp:1082]: Whatever's easiest and cleanest. Just something to consider.

File: components/eamxx/tests/coupled/dynamics_physics/homme_shoc_cld_spa_p3_rrtmgp_dp/input.yaml (hunk 4 above)
Comment (tcclevenger) [reply components/eamxx/tests/coupled/dynamics_physics/homme_shoc_cld_spa_p3_rrtmgp_dp/input.yaml:28]: This value isn't an absolute but I guess is like a "percentage perturbed". So it will scale with variable magnitude. Basically `pert_val` is in `[-pertlim, pertlim]` and then `f_val_new = f_val*(1+pertval)`. Maybe this could be better described in the name..

Comment (PeterCaldwell) [reply components/eamxx/tests/coupled/dynamics_physics/homme_shoc_cld_spa_p3_rrtmgp_dp/input.yaml:28]: >This value isn't an absolute but I guess is like a "percentage perturbed"
Ah, percentage perturbed is perfect. I think it is very reasonable to just use the same percentage perturbation for all variables. Much better than asking the user to specify a zillion different scaling factors.

File: components/eamxx/src/control/atmosphere_driver.cpp (hunk 1 above)
Comment (tcclevenger) [reply components/eamxx/src/control/atmosphere_driver.cpp:1082]: I'm going to leave this as is since Luca is reworking this whole framework anyways.

File: components/eamxx/src/control/atmosphere_driver.cpp (hunk 6 above)
Comment (tcclevenger) [reply components/eamxx/src/control/atmosphere_driver.cpp]: For DP, the pressure will be the same in each column, but obviously not true in the general case.

Comment (PeterCaldwell) [reply components/eamxx/src/control/atmosphere_driver.cpp]: > For DP, the pressure will be the same in each column, but obviously not true in the general case.
I'm leaning towards using the current impl a) out of laziness and b). because I think what we really want is perturbations in the first N levels rather than below a strict pressure cutoff. Specifying a pressure is just shorthand for the number of levels. Though maybe we should just directly specify the number of levels then? @bogensch ?

**Comment by tcclevenger** [comment]
@bartgol Looking back at this, `get_strided_view()` is only implemented for rank-1 fields. So we cannot simply use subfields here. 

I'll probably go the route of 
```
Field perturb (...); // with layout stripped of levels
for (int ilev=start; ilev<end; ++ilev) {
   randomize(perturb,engine,pdf);
   f.subfield(lev_dim, ilev).update(perturb,1.0,1.0);
}
```

**Comment by tcclevenger** [comment]
AT bringing up an interesting point: np1 vs np${n} tests are failing (perturbations different when you have multiple processesors). I will think about this because I don't want to disable this test since IOP classes uses MPI in a way that is very different that typical AD unit tests. Should be tested IMO.

**Comment by PeterCaldwell** [comment]
> np1 vs np${n} tests are failing (perturbations different when you have multiple processesors)

Disambiguation: by np1 vs npN you are referring to the number of procs in the submit script rather than "ne30np4" kind of thing, right?

Assuming so, I'm confused why switching processor counts would change the random numbers. Isn't that the point of specifying a seed?

**Comment by tcclevenger** [comment]
@PeterCaldwell It's because for a single rank, pert random vals are computed for every column/level based on the seed, but for multiple rank, each rank starts it's owned column/level with that seed. This is actually a bug I'm realizing, because each MPI domain will have the exact same values at the beginning of it's owned columns, so the fields will have this odd repeating structure. I need to make sure that each MPI rank gets a unique random seed.

But that still would lead rank 1 vs. rank n to be different. In EAM it looks like they reset the random seed to be the global element ID (unless a random seed is specified). Maybe this makes sense?

**Comment by PeterCaldwell** [comment]
> In EAM it looks like they reset the random seed to be the global element ID (unless a random seed is specified).

Maybe we should set the random seed for each element to be the global random seed times the global element ID? Then everything would be deterministic based on a single input number but would be independent of MPI decomposition. Or would that not be random enough from element to element? My intuition is that it would...

**Comment by tcclevenger** [comment]
Yeah, there should be no relationship from element to element, and it would match runs with any number of MPI ranks. If we want to be able to also have it non-deterministic, I can still take a seed and pre process the seed based on global element ID, then it would be same for all MPI ranks, and deterministic to the seed value (not just global element ID).

**Comment by bartgol** [comment]
> Yeah, there should be no relationship from element to element, and it would match runs with any number of MPI ranks. If we want to be able to also have it non-deterministic, I can still take a seed and pre process the seed based on global element ID, then it would be same for all MPI ranks, and deterministic to the seed value (not just global element ID).

You mean to combine the overall seed from Catch2 with the col GID? I think that's a good idea (keep the ability to have different values every run), but you need to enforce the same Catch2 seed for np1 and npN runs...

You need to pass `--rnd-seed XYZ` as exec args when you create the test, and you need to create the seed via CMake. Something like this
```
string (RANDOM LENGTH 5 ALPHABET "0123456789" rnd_seed)
CreateUnitTest (
  ....
  EXE_ARGS "--rng-seed ${rnd_seed}"
)
```

**Comment by PeterCaldwell** [comment]
> you need to enforce the same Catch2 seed for np1 and npN runs

@bartgol - why? np1 vs npN would be non-BFB anyways so who cares if changing the number of GLL nodes per element changes the random seed?

**Comment by bartgol** [comment]
> > you need to enforce the same Catch2 seed for np1 and npN runs
> 
> @bartgol - why? np1 vs npN would be non-BFB anyways so who cares if changing the number of GLL nodes per element changes the random seed?

Sorry, I'm using `npX` as `number_of_mpi_ranks_X`, not as `number_of_gll_points_X`...

**Comment by PeterCaldwell** [comment]
Update from @bogensch - for DP and SCM usage, we actually want every run to use the same random seed in order to have DP/SCM give the same answer every time we run. But we also need the ability to provide different random seeds for different simulations in order to create ensembles of perturbed initial condition runs. I think this is what EAMf90 does, but I think it also has the ability to actually apply truly random initial noise. I'm not sure what that would be good for - perhaps to increase the envelope of testing? So I guess we should include the actually random init option as well? In other words - just copy EAMf90?

**Comment by PeterCaldwell** [comment]
> I'm using npX as number_of_mpi_ranks_X, not as number_of_gll_points_X... That's the string we append to the test to differentiate tests with different number of ranks.

Maybe we should change one of those terms to disambiguate? Or maybe we don't have control of either of these names?

**Comment by bartgol** [comment]
> > I'm using npX as number_of_mpi_ranks_X, not as number_of_gll_points_X... That's the string we append to the test to differentiate tests with different number of ranks.
> 
> Maybe we should change one of those terms to disambiguate? Or maybe we don't have control of either of these names?

We can of course disambiguate. `-np X` is an `mpirun` flag (though mpirun is not the official MPI runner, which is `mpiexec` and which doesn't support `-np`). That's why we used it probably. We can try to think about new non-confusing choices. 
- `myTest_mpiX`: I'm worried someone may interpret this as "MPI version X"
- `myTest_nrX`: with `nr`="number of ranks"
- `myTest_Xranks`: a woeful 3 chars longer than now
- `myTest_mpX`: may be too close to "omp", making it a tad harder to parse `myTesp_mp1_omp4`

I also want to point out that we use `npX` also in our nc output files. Would you like to change that as well?

**Comment by tcclevenger** [comment]
I'll work on an impl with all this in mind and then ping you all to look at it. Probably will change the `perturb()` impl quite a bit since `randomize()` is not aware of global element IDs and such. Maybe there will be a nice extension that can also go in `field_utils.hpp`, but I'm not sure where it will go just yet.

**Comment by tcclevenger** [comment]
@bartgol I guess I would have a slight preference for option 2 or 3.

**Comment by PeterCaldwell** [comment]
Yeah, I like nrX or Xrange instead of np X. Maybe we should ask on the eamxx slack though so we don't have everyone surprised that we made this git-backroom deal.

**Comment by bartgol** [comment]
> I'll work on an impl with all this in mind and then ping you all to look at it. Probably will change the `perturb()` impl quite a bit since `randomize()` is not aware of global element IDs and such. Maybe there will be a nice extension that can also go in `field_utils.hpp`, but I'm not sure where it will go just yet.

We have a few options, I think
- add `randomize (Field& f, Engine& e, Pdf& pdf, const AbstractGrid& g)` in field_utils (which uses the grid to get the gids). Drawback, if we ever implement #2628, we are injecting grid lib into the upstream field lib.
- same as before, but simply pass `const Field& dof_gids` to the routine (instead of grid). It avoids libs deps in #2628, but still injects knowledge about what dofs are in the field lib.
- manually loop over cols, take subfiled, and call randomize (a bit clunky, since we need to recreate the `Engine` obj for each col)
- add `reproducible_randomize (Field& f, Engine& e, PDF& pdf, int base_seed)` as a method to `AbstractGrid`, which takes care of changing the seed via the `seed` method of `Engine`.

File: components/eamxx/src/share/tests/field_utils.cpp (hunk 9 above)
Comment (tcclevenger) [reply components/eamxx/src/share/tests/field_utils.cpp:240]: Added checks below which adds these

**Comment by tcclevenger** [comment]
@PeterCaldwell @bartgol I updated the PR. Main things
- Two random seed related params: `generate_perturbation_random_seed` and `perturbation_random_seed`. If nothing is given, use `seed=0` (default case), if `generate_perturbation_random_seed=true` generate a random seed (assert no `perturbation_random_seed` is given), and if `perturbation_random_seed` is given use that. For all cases, seed is in log file so anything can be recreated.
- Perturbation field is now defined, randomized, and applied to field we are perturbing. No more looping through field view manually.
- Random seed is reset at each column to be `base_seed + gids(icol)`. So test is BFB between MPI ranks used.
- `Field::scale(Field x)` function added with test. I called this `scale` instead of `multiply` because `multiply` for a multi-dim field would maybe imply matrix mult?

I'm gone for the break after today, so if this doesn't get looked at today, I'll re-ping in January.

**Comment by PeterCaldwell** [comment]
@tcclevenger - Looks good, thanks! My only question is about the new scale function... it looks like it sets the scaling factor to the fill value defined for a given variable. I don't recall what the fill_value means. Typically, I've seen "fill_value" used to mean the value data is set to when we want to mask that cell but don't want to create a whole new masked array. I think we are using it here to mean the value data gets initialized to if we don't read in an actual initial condition. Is this true? If so, maybe we should call it default_value, or init_value, or default_init_value?

Also, I missed the request to re-review so I expect @bartgol did too. I'm happy to approve once I understand fill_value.

**Comment by tcclevenger** [comment]
@PeterCaldwell It looks like if the either the field we are scaling or scaling with has an entry value equal to the mask value in the fields header, then the result at that index will be set to `fill_val` instead of computed. I've not run into this "mask val" part of fields before, so I was just copying the behavior of the other `scale()` and `update()` functions.

**Comment by PeterCaldwell** [comment]
> if the either the field we are scaling or scaling with has an entry value equal to the mask value in the fields header, then the result at that index will be set to `fill_val` instead of computed. 

Hi @tcclevenger - I must have had a good holiday break because I don't remember why you're telling me this. It makes sense to me and seems slick to keep entries which are already fill value at the fill value regardless of rescaling. I don't understand why we would ever have a field or a scaling factor with a number we want to use which is equal to the fill value - that's why the fill value is usually -999 or something.

File: components/eamxx/src/share/field/field_utils.hpp [hunk 10]
```diff
@@ -43,6 +43,69 @@ void randomize (const Field& f, Engine& engine, PDF&& pdf)
...
+                   (std::is_same_v<ST,float> && data_type==DataType::FloatType) or
+                   (std::is_same_v<ST,double> && data_type==DataType::DoubleType),
+                   "Error! Field data type incompatible with input PDF.\n");
+
+  using namespace ShortFieldTagsNames;
+  const auto& fl = f.get_header().get_identifier().get_layout();
+
+  // Field we are perturbing should have a level midpoint dimension,
+  // and it is required to be the last dimension
+  EKAT_REQUIRE_MSG(fl.has_tag(LEV),
```
bartgol, line 85: You could condensate these two checks in one, and add a printout of the field layout. Something like

```suggestion
  EKAT_REQUIRE_MSG(fl.rank()>0 and (fl.tags().back()==LEV or fl.tags().back()==ILEV),
    "Error! Trying to perturb a field with no vertical levels dimension dimension.\n"
    "  - field name: " + f.name() + "\n"
    "  - field layout: " + to_string(fl) + "\n");
```

components/eamxx/src/share/field/field_utils.hpp
```diff
@@ -43,6 +43,69 @@ void randomize (const Field& f, Engine& engine, PDF&& pdf)
...
+                   "Error! Trying to perturb field \""+f.name()+"\", but field "
+                   "has no LEV dimension.\n");
+  EKAT_REQUIRE_MSG(fl.tags().back() == LEV,
+                   "Error! Trying to perturb field \""+f.name()+"\", but field "
+	                 "does not have LEV as last dimension.\n");
+
+  if (fl.has_tag(COL)) {
+    // If field has a column dimension, it should be the first dimension
+    EKAT_REQUIRE_MSG(fl.tag(0) == COL,
+                     "Error! Trying to perturb field \""+f.name()+"\", but field "
```
bartgol, line 95: I would add the layout of the field to the error msg here as well.

components/eamxx/src/share/field/field_utils_impl.hpp
```diff
@@ -214,6 +214,72 @@ void randomize (const Field& f, Engine& engine, PDF&& pdf)
...
+  // Check to see if field has a column dimension
+  using namespace ShortFieldTagsNames;
+  const bool has_column_dim = fl.has_tag(COL);
+
+  if (has_column_dim) {
+    // Because Column is the partitioned dimension, we must reset the
+    // RNG seed to be the same on every column so that a column will
+    // have the same value no matter where it exists in an MPI rank's
+    // set of local columns.
+    dof_gids.sync_to_host();
```
bartgol, line 236: No need to sync gids to host. They are immutable during a simulation, so we make sure they are synced when we construct the grid.

tcclevenger:
@bartgol This is ready to be reviewed again. Fail from yesterday was unrelated to PR (I think mappy timed out?).

bartgol (approved):

PR #2632, opened 2023-12-09
Overhaul of the initial conditions (Initial run type)

bartgol:
The initial conditions setup is one of the hackiest parts in EAMxx. The reason is that when EAMxx started, there was no PG2 support, and everything was easy: the fields to read from file are all the inputs to the atmosphere. With PG2, things got complicated, and forced us to add some hacks to get things to work.

This PR is an attempt to "hide" a bit better the hack, without using existing tools in a way that they were not meant to. In particular, rather than funneling the PG2 particular case through the non-PG2 framework (input to atm is what we read), this PR creates an interface `ModelInit` (all names up for discussions, ofc) to be used for IC read. The interface is abstract, and different implementations will be used depending on whether PG2 is active or not. This allows to be more explicit in the PG2 case, without trying to shove a square peg in a round hole.

So far I implemented the non-PG2 case. I tried it on a couple of non-PG2 tests (`p3_standalone` and `homme_shoc_cld_spa_p3_rrtmgp`), and it seems to work (meaning that it runs to completion...I have no baselines testing for standalone scream). I will work on the PG2 case next week. Basically, the code will do the same stuff (read on physGLL, remap to dyn, then remap to PG2), but it won't have to create dummy physGLL version of the input fields and add them as "Required" fields jsut so that the AD can add them to the list of fields to read. Instead, the PG2 model init instance will directly access the IC file, and read on GLL grid, without adding any additional field to the main AD field manager.

@AaronDonahue @tcclevenger this is still incomplete (there's no PG2 support!), but I'd like you to start taking a look, to see if the framework makes sense or it is even more confusing than before.

@tcclevenger I would particularly appreciate if you can take a look at how I handle IOP-vs-nonIOP in the init sequence. I hope I am doing things correctly (I haven't tried to run the dp test yet, maybe I should do that first).

### DIFF
--- a/components/eamxx/src/control/atmosphere_driver.cpp
+++ b/components/eamxx/src/control/atmosphere_driver.cpp
@@ -1,4 +1,5 @@
 #include "control/atmosphere_driver.hpp"
+#include "control/eamxx_default_model_init.hpp"
 #include "control/atmosphere_surface_coupling_importer.hpp"
 #include "control/atmosphere_surface_coupling_exporter.hpp"
 
@@ -11,6 +12,7 @@
 #include "share/util/scream_timing.hpp"
 #include "share/util/scream_utils.hpp"
 #include "share/io/scream_io_utils.hpp"
+#include "share/eamxx_model_init.hpp"
 #include "share/property_checks/mass_and_energy_column_conservation_check.hpp"
 
 #include "ekat/ekat_assert.hpp"
@@ -163,8 +165,6 @@ init_time_stamps (const util::TimeStamp& run_t0, const util::TimeStamp& case_t0)
   m_run_t0 = m_current_ts = run_t0;
   m_case_t0 = case_t0;
 }
-
-
 
 void AtmosphereDriver::
 setup_intensive_observation_period ()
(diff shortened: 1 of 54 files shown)


components/eamxx/src/share/eamxx_model_init.cpp
```diff
@@ -0,0 +1,190 @@
...
+          "Error! Bad name specification for constant field initialization.\n"
+          " - input name: " + fnames[i] + "\n");
+      try {
+        int cmp = std::stoi (tail);
+        init_constant_field(fnames[i],fvalues[i],cmp);
+      } catch (std::exception&) {
+        EKAT_ERROR_MSG (
+          "Error! Bad name specification for constant field initialization.\n"
+          " - input name: " + fnames[i] + "\n");
+      }
```
tcclevenger, line 79: Is this only catching the case that `std::stoi(tail)` gives error? If so, should the error message mention this? I only say this because there is a high likelihood that a user reaches this error message if they are specifying constant fields. Something like 
```
"Error! Bad name specification for constant field initialization. "
"Cannot deduce field component from specified \"cmp_"+tail+"\"\n"
" - input name: " + fnames[i] + "\n"
```
The same for `EKAT_REQUIRE_MSG (tail.size()>0,...)` above.

components/eamxx/src/control/atmosphere_driver.cpp
```diff
@@ -985,291 +985,109 @@ void AtmosphereDriver::set_initial_conditions ()
...
+    m_atm_logger->info("        filename: " + filename);
+    m_atm_params.sublist("provenance").set("topography_file",filename);
+    if (iop) {
+      // For IOP, we load from file and copy data from the closest
+      // lat/lon column to every other column
+      auto topo_fnames_file  = model_init->get_topo_fields_names_file();
+      auto topo_fnames_eamxx = model_init->get_topo_fields_names_eamxx();
+      for (const auto& it : topo_fnames_file) {
+        const auto& gname = it.first;
+        auto fm = get_field_mgr(gname);
```
tcclevenger, line 1058: I think here we will need to insert
```
if (gname == "Physics PG2") {
  iop->setup_io_info(filename, fm->get_grid());
}
```

tcclevenger (commented):
The IOP looks like it will work fine for the current test. I added a fix we will need for PG2, but there are other things holding us back from PG2 working. Still, I think that is worth adding.

And then I added a comment about error messages, but I don't feel too strongly about that.

components/eamxx/src/control/atmosphere_driver.cpp
bartgol: Looks like it's safe to call `setup_io_info` multiple times (the if statements should prevent code from running twice), so maybe I don't need to check the grid name? (I'm trying to keep explicit knowledge about grids names as much as possible outside the driver...)

tcclevenger: The issue is that if `setup_io_info()` is called on GLL grid with the topo file it should error out, because topo file has `lat`/`lon` for PG2 only. Maybe if GLL is already setup (it will be because ICs are read first) we will just return, but that kinda hides that it is an incompatible call.

Another option would be to change `setup_io_info()` to take in a list of grids that require read from file, and the IC file and topo file (if exists), and then call before ICs/topo are read. Then internally IOP can error out if an incorrect request is made (e.g., read from PG2 but no topo file) and setup every grid needed at once.

tcclevenger: And from the AD's perspective, they just hand over all the info they have agnostic of exactly which grids they are.

bartgol: Ok, I will think of how to best handle this.

AaronDonahue:
@bartgol can we turn this into a git-issue and tag it such that it is considered in the I/O epic?  We are doing a major overhaul of I/O now and maybe that can include how we handle initial conditions? Maybe that's a stretch, but I'm concerned that this PR is now so far out-of-sync w/ master that it would be difficult to revive it from this branch.

PR #2635, opened 2023-12-12
This commit fixes the way the time interpolation determines the startdate

AaronDonahue:
This commit uses the string units associated with the "time" variable in the source data rather than rely on case_t0 in the file metadata.

This is more accurate because the case_t0 may not always reflect the actual reference date as specified in the time variable.

### DIFF
--- a/components/eamxx/src/share/util/eamxx_time_interpolation.cpp
+++ b/components/eamxx/src/share/util/eamxx_time_interpolation.cpp
@@ -251,8 +251,6 @@ void TimeInterpolation::set_file_data_triplets(const vos_type& list_of_files) {
   int running_idx = 0;
   for (size_t ii=0; ii<list_of_files.size(); ii++) {
     const auto filename = list_of_files[ii];
-    // Reference TimeStamp
-    auto ts_file_start = scorpio::read_timestamp(filename,"case_t0");
     // Gather the units of time
     auto time_units_tmp = scorpio::get_any_attribute(filename,"time","units");
     auto& time_units = ekat::any_cast<std::string>(time_units_tmp);
@@ -268,6 +266,19 @@ void TimeInterpolation::set_file_data_triplets(const vos_type& list_of_files) {
     } else {
       EKAT_ERROR_MSG("Error!! TimeInterpolation::set_file_triplets - unsupported units of time = (" << time_units << ") in source data file " << filename << ", supported units are: seconds, minutes, hours and days");
     }
+    auto date_start = time_units.find_first_of("0123456789");
+    auto time_units_date = time_units.substr(date_start);
+    // Parse the time_units_date into a vector of date and time
+    // Reference TimeStamp
+    auto tokens = ekat::split(time_units_date," ");
+    auto date_tkn = ekat::split(tokens[0],"-");
+    auto time_tkn = ekat::split(tokens[1],":");
+    std::vector<int> ts_date, ts_time;
+    for (int jj=0; jj<3; jj++) {
+      ts_date.push_back(stoi(date_tkn[jj]));
+      ts_time.push_back(stoi(time_tkn[jj]));
+    } 
+    const TimeStamp ts_file_start(ts_date,ts_time);
     // Gather information about time in this file
     if (ii==0) {
       ts_ref = ts_file_start;


AaronDonahue:
@mahf708 , @elynnwu and @bartgol - I would welcome any feedback on this change.  It changes how time interpolation determines the date/time associated w/ the `time` variable in a source data file.
OLD METHOD: Use `case_t0` which is in the netCDF header of E3SM outputs.  It was noted that when we start nudging to observed data it's unlikely that `case_t0` will be in the header, thus adding an extra step for users.
NEW METHOD: Grab the reference date directly from the `time` variable units.  There is an assumption here that the format for the units will follow the pattern `UNIT since YYYY-MM-DD HH:MM:SS`.  Which may, or may not be reasonable.

components/eamxx/src/share/util/eamxx_time_interpolation.cpp
```diff
@@ -268,6 +266,19 @@ void TimeInterpolation::set_file_data_triplets(const vos_type& list_of_files) {
...
+    // Reference TimeStamp
+    auto tokens = ekat::split(time_units_date," ");
+    auto date_tkn = ekat::split(tokens[0],"-");
+    auto time_tkn = ekat::split(tokens[1],":");
+    std::vector<int> ts_date, ts_time;
+    for (int jj=0; jj<3; jj++) {
+      ts_date.push_back(stoi(date_tkn[jj]));
+      ts_time.push_back(stoi(time_tkn[jj]));
+    } 
+    const TimeStamp ts_file_start(ts_date,ts_time);
```
bartgol, line 281: Is this really a file start? The units will likely be "nXYZ from <case t0>", so even if you have multiple files, they should all have the same units, no? Perhaps you should add the first value of the time var to `ts_ref`?

components/eamxx/src/share/util/eamxx_time_interpolation.cpp
```diff
@@ -268,6 +266,19 @@ void TimeInterpolation::set_file_data_triplets(const vos_type& list_of_files) {
...
     // Gather information about time in this file
     if (ii==0) {
       ts_ref = ts_file_start;
```
bartgol, line 284: Assuming my comment above is correct, we should prob only compute `ts_file_start` via string parsing for the first file. Every other file should get `ts_file_start = ts_ref + <first time var value>`...

components/eamxx/src/share/util/eamxx_time_interpolation.cpp
```diff
@@ -268,6 +266,19 @@ void TimeInterpolation::set_file_data_triplets(const vos_type& list_of_files) {
     } else {
       EKAT_ERROR_MSG("Error!! TimeInterpolation::set_file_triplets - unsupported units of time = (" << time_units << ") in source data file " << filename << ", supported units are: seconds, minutes, hours and days");
     }
+    auto date_start = time_units.find_first_of("0123456789");
```
bartgol, line 269: Perhaps, if you want to _assume_ that the units are "XYZ since YYYY-MM-DD HH:MM:SS", you should check that the string has indeed that format. You should use a regex here. Since you found out what the units are, you can do
```
#include <regex>
...
std::regex r { units + " since [0-9]{4,}-[0-9]{2,}-[0-9]{2,} [0-9]{2,}:[0-9]{2,}:[0-9]{2,}"};
EKAT_REQUIRE_MSG (std::regex_search(time_units,r),
  "Error! Time units should be of the form '<untis> since YYYY-MM-DD HH:MM:SS'\n"
  "  - time units: " + time_units + "\n");
```

bartgol (commented):
I'm con

elynnwu:
I'm wondering how people feel about a "shift time" option in the namelist. Instead of having to change `case_t0` or the actual `time` index, we would add an option to treat the timestamp by an offset specified by the user. This is convenient since the users won't have to actually change the input files and instead it's being taken care of from within time interpolation. Another use case I could think of it's for some sort of data assimilation where the observation data may be in a local time zone and needs to be shifted to UTC, having this option would also eliminate modifying the source file. Though I don't know that this use case even makes sense.

I'm open to see what others think. If we were to go with reading the actual `time` index and not `case_t0`, I would still suggest that during post processing, one would change both so that there's no confusion when you look back to the source file.

PeterCaldwell:
I kinda like the "time shift" option *for reading in data*. I can imagine a lot of cases where inputdata uses a time convention that we don't like. Allowing an offset avoids having to rewrite the files. 

I don't want people to be able to change what variables mean as they *write out* data because then we could never trust that we know what our output means... but I don't think that's what we're discussing..

bartgol:
> I'm wondering how people feel about a "shift time" option in the namelist. Instead of having to change `case_t0` or the actual `time` index, we would add an option to treat the timestamp by an offset specified by the user. This is convenient since the users won't have to actually change the input files and instead it's being taken care of from within time interpolation. Another use case I could think of it's for some sort of data assimilation where the observation data may be in a local time zone and needs to be shifted to UTC, having this option would also eliminate modifying the source file. Though I don't know that this use case even makes sense.
> 
> I'm open to see what others think. If we were to go with reading the actual `time` index and not `case_t0`, I would still suggest that during post processing, one would change both so that there's no confusion when you look back to the source file.

Are you saying the user could pass via yaml input a parameter that basically says what the start time is in the first file, and subsequent time slabs are assumed to be at `start_time + index*dt`, where `dt`=time[index]-time[index-1]` (which should be constant wrt `index`)?

elynnwu:
> 
> Are you saying the user could pass via yaml input a parameter that basically says what the start time is in the first file, and subsequent time slabs are assumed to be at `start_time + index*dt`, where `dt`=time[index]-time[index-1]`(which should be constant wrt`index`)?

Yes, exactly. So the users don't have to post process the source files.

bartgol:
> > Are you saying the user could pass via yaml input a parameter that basically says what the start time is in the first file, and subsequent time slabs are assumed to be at `start_time + index*dt`, where `dt`=time[index]-time[index-1]`(which should be constant wrt`index`)?
> 
> Yes, exactly. So the users don't have to post process the source files.

I think that's ok. Would you still parse the nc file to see if there is `case_t0` or if the time units contain `XYZ since YYYY-MM-DD HH:MM:SS` or would you simply _always_ require to provide the start date?

AaronDonahue:
I think the issue we have to consider is when a user nudges using multiple files with a different reference date/time for the `time` variable.  @mahf708 illustrates such a case with the MERRA data up above.  @elynnwu 's suggestion would work well when nudging from E3SM data which has the same reference start date for all output in a simulation campaign, that is `case_t0`.

We could let a user override the reference start date using Elynn's suggestion of a runtime option.  But I think making it a requirement would invalidate observed data for nudging, or at the very least add extra work to the person using that data as they would have to make sure every file had a `time` variable that conformed to the same start date.

Additionally, I think @mahf708 's example illustrates the other motivation for this code change.  I would suspect that the MERRA data does not contain a `case_t0` field in the header metadata.

AaronDonahue:
It also might help to describe the two different applications that led to this change:
1. @elynnwu 's application is to use E3SM output data that has had the time axis shifted by half of the averaging window.  So she needs a way to make sure the time interpolation recognizes this shift in the time axis.  Her suggestion of adding a runtime option would make this pretty straightforward.
2. @mahf708 intends to use observed data in the aerosol nudging runs, and we cannot assume that observed data has the same netCDF header metadata as an E3SM output file, and as noted we cannot assume that all files use the same reference date for the time axis.

AaronDonahue:
All that being said, I am open to suggestions on a better application.  I just think we want to address this while we have it in mind rather than work around the current implementation.

bartgol:
Can't we have multiple instances of the nudging class? Like this
```
atmosphere_processes: [ sc_import, all_nudgings, homme, physics, sc_export ]
all_nudgings:
  Type: Group
  atm_procs_list: [ nudge1, nudge2]
  nudge1:
    Type: nudging
    files: [file1.nc]
    ref_date: 1950-01-01 00:00:00
  nudge2:
    Type: nudging
    files: [file2.nc]
    ref_date: 1850-01-01 00:00:00
...
```

elynnwu:
> Can't we have multiple instances of the nudging class? Like this
> 
> ```
> atmosphere_processes: [ sc_import, all_nudgings, homme, physics, sc_export ]
> all_nudgings:
>   Type: Group
>   atm_procs_list: [ nudge1, nudge2]
>   nudge1:
>     Type: nudging
>     files: [file1.nc]
>     ref_date: 1950-01-01 00:00:00
>   nudge2:
>     Type: nudging
>     files: [file2.nc]
>     ref_date: 1850-01-01 00:00:00
> ...
> ```

I like the flexibility of this. If one were to nudge to more than one dataset, it would be quite a pain to have to combine those source files. We can implement this in a separate PR though. The question then is how does `nudge1` determine the time axis frequency? If it reads the `ref_date` from namelist, does it then read the time `days since xxxx` as time index is in the unit of days since the `ref_date` provided and ignore `xxxx`?

bartgol:
> > Can't we have multiple instances of the nudging class? Like this
> > ```
> > atmosphere_processes: [ sc_import, all_nudgings, homme, physics, sc_export ]
> > all_nudgings:
> >   Type: Group
> >   atm_procs_list: [ nudge1, nudge2]
> >   nudge1:
> >     Type: nudging
> >     files: [file1.nc]
> >     ref_date: 1950-01-01 00:00:00
> >   nudge2:
> >     Type: nudging
> >     files: [file2.nc]
> >     ref_date: 1850-01-01 00:00:00
> > ...
> > ```
> 
> I like the flexibility of this. If one were to nudge to more than one dataset, it would be quite a pain to have to combine those source files. We can implement this in a separate PR though. The question then is how does `nudge1` determine the time axis frequency? If it reads the `ref_date` from namelist, does it then read the time `days since xxxx` as time index is in the unit of days since the `ref_date` provided and ignore `xxxx`?

I used `ref_date` to echo your idea of specifying the time shift via input file, I wasn't suggesting how to use it.

I think every file should have some sort of units defined for the time. They may not specified as an nc attribute, but the time var must have some sort of units, so we should be able to compute `dt_file=time(1)-time(0)` in some units (either read the 'units' attribute, or maybe pass it via input yaml file). Then, the ref_date should be interpret as "assume that time(0) (in whatever units it is) coincides with this date". With this info, one can reconstruct the timestamp of all slices in the file, by doing `timestamp(N) = ref_date + N*dt_file`.

elynnwu:
> 
> I used `ref_date` to echo your idea of specifying the time shift via input file, I wasn't suggesting how to use it.
> 
> I think every file should have some sort of units defined for the time. They may not specified as an nc attribute, but the time var must have some sort of units, so we should be able to compute `dt_file=time(1)-time(0)` in some units (either read the 'units' attribute, or maybe pass it via input yaml file) and convert it to seconds (which is what we need in eamxx for timestep sizes). Then, the ref_date should be interpret as "assume that time(0) (in whatever units it is) coincides with this date". With this info, one can reconstruct the timestamp of all slices in the file, by doing `timestamp(N) = ref_date + N*dt_file`.

I like this idea. It makes a lot of sense to read `dt` from the source file timestamps. Then we can default `ref_date` to `case_t0` if it's present in the file, or use the first time index, or read in `ref_date` from namelist if it's specified.

bartgol:
> hen we can default `ref_date` to `case_t0` if it's present in the file, or use the first time index, or read in `ref_date` from namelist if it's specified.

Yes, we need to think carefully about which one should have the priority though.

components/eamxx/src/share/util/eamxx_time_interpolation.cpp
bartgol: Are you saying we could have a list of files with different units? If we allow this, for safety, I think we should check that the files are "not overlapping", which is kind of tedious/annoying.

bartgol: Well, I see both a different case_t0 as well as a different scale (minutes vs days).. I'm ok leaving to the user the burden of verifying nonoverlap-ness, but mixing units may be confusing...

bartgol:
I don't think the convo about different files having different units ever got resolved. In particular, did anyone check that nudging works correctly if file2 has different units and/or ref date than file1?

AaronDonahue:
@mahf708 did you happen to run a case that would address @bartgol 's question?

bartgol (approved):
Since Naser said this PR solves the issues he had, let's merge it. We'll revisit the whole "read-data-from-file-and-interpolate" infrastructure soon anyways, so no need to polish corner cases here...

bartgol:
Ok, I removed automerge. If this PR is no longer needed, then we should hold off merge, maybe. We are thinking of pulling out the data-from-file interpolation framework, and make it more structured. So if this is not needed now, we may just do changes when we do the refactor. I'm thinking that when we design the common interfaces for data interpolation, we will dig up other issues, and we'll have to think through all the assumptions as well as restrictions on the class usage.

PeterCaldwell:
Note that Elynn is on maternity leave for the next several months, so don't hold your breath waiting for her.

bartgol:
I'm closing this PR, since there's no consensus on whether it should be merged or not, and the AT keeps parsing this PR, posting messages and spamming our github notifications. We can re-open later, if we think it's needed. Most likely though, when we will get to work on unifying the time-data from file approach we will (re)consdier how we handle the time dimension, and decide what's the most portable/robust approach then.

PR #2636, opened 2023-12-12
Some fixes related to radiation interface and restarts

bartgol:
In particular:
 - remove any mention of packs: rrtmgp does not use them, so they are misleading at best
 - init active gases workaround inside rrtmgp: allows less spaghetti code from the driver
 - enforce consistency between rad_frequency and model restart: no longer allow restart to happen in the middle of a rad superstep
 - remove no longer needed RESTART group names from some rad fields
 
 @jgfouca can you please review the buildnml part?
 @brhillman can you please review rad internal consistency?
 @tcclevenger can you please double check pg2 hack consistency?


components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
```diff
@@ -771,48 +815,13 @@ void RRTMGPRadiation::run_impl (const double dt) {
...
 
         // We read o3 in as a vmr already
         if (name=="o3") continue;
 
         auto d_vmr = get_field_out(name + "_volume_mix_ratio").get_view<Real**>();
-        if (name == "h2o") {
-          // h2o is (wet) mass mixing ratio in FM, otherwise known as "qv", which we've already read in above
-          // Convert to vmr
-          const auto policy = ekat::ExeSpaceUtils<ExeSpace>::get_default_team_policy(ncol, m_nlay);
-          Kokkos::parallel_for(policy, KOKKOS_LAMBDA(const MemberType& team) {
```
bartgol, line 781: The calculation of h2o was fine, since it only computed the value for the cols in the current chunk. However, for the other gases, we were recomputing all the `m_ncol` entries for every chunk.

components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
```diff
@@ -771,48 +815,13 @@ void RRTMGPRadiation::run_impl (const double dt) {
...
-            const int i = team.league_rank();
-            const int icol = i + beg;
-            Kokkos::parallel_for(Kokkos::TeamVectorRange(team, nlay), [&] (const int& k) {
-              d_vmr(icol,k) = PF::calculate_vmr_from_mmr(gas_mol_weights[igas],d_qv(icol,k),d_qv(icol,k));
-            });
-          });
-          Kokkos::fence();
-        } else if (name == "n2") {
-          // n2 prescribed as a constant value
-          Kokkos::deep_copy(d_vmr, m_params.get<double>("n2vmr", 0.7906));
```
bartgol, line 793: These two could probably be done during init.

components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
```diff
@@ -771,48 +815,13 @@ void RRTMGPRadiation::run_impl (const double dt) {
       // n2 and co are set to constants and are not handled by trcmix;
       // the rest are handled by trcmix
       real2d tmp2d = subview_2d(m_buffer.tmp2d);
-      const auto gas_mol_weights = m_gas_mol_weights;
       for (int igas = 0; igas < m_ngas; igas++) {
         auto name = m_gas_names[igas];
 
         // We read o3 in as a vmr already
         if (name=="o3") continue;
```
bartgol, line 822: This skips the line 
```
m_gas_concs.set_vmr(name, tmp2d);
```
down below. Is this intended? The comment `we read o3 in as a vmr already` makes me think the answer is NO, and we were simply trying to skip the trcmix calculation...

components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
```diff
@@ -565,6 +565,50 @@ void RRTMGPRadiation::run_impl (const double dt) {
...
+        // Convert to vmr
+        const auto policy = ekat::ExeSpaceUtils<ExeSpace>::get_default_team_policy(m_ncol, m_nlay);
+        Kokkos::parallel_for(policy, KOKKOS_LAMBDA(const MemberType& team) {
+          const int icol = team.league_rank();
+          Kokkos::parallel_for(Kokkos::TeamVectorRange(team, nlay), [&] (const int& k) {
+            d_vmr(icol,k) = PF::calculate_vmr_from_mmr(gas_mol_weights[igas],d_qv(icol,k),d_qv(icol,k));
+          });
+        });
+        Kokkos::fence();
+      } else if (name == "n2") {
```
bartgol, line 593: These two could probably be done during init. The perf gain may not be noticeable anyways though.

jgfouca (dismissed):
Python changes LGTM

components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
```diff
@@ -83,88 +85,88 @@ void RRTMGPRadiation::set_grids(const std::shared_ptr<const GridsManager> grids_
...
+  add_field<Required>("qv",scalar3d_layout_mid,kgkg,grid_name);
   add_field<Required>("surf_lw_flux_up",scalar2d_layout,W/(m*m),grid_name);
   // Set of required gas concentration fields
   for (auto& it : m_gas_names) {
     // Add gas VOLUME mixing ratios (moles of gas / moles of air; what actually gets input to RRTMGP)
     if (it == "o3") {
       // o3 is read from file, or computed by chemistry
-      add_field<Updated >(it + "_volume_mix_ratio", scalar3d_layout_mid, molmol, grid_name, ps);
+      add_field<Required >(it + "_volume_mix_ratio", scalar3d_layout_mid, molmol, grid_name);
+      TraceGasesWorkaround::singleton().add_active_gas(it + "_volume_mix_ratio");
```
tcclevenger, line 110: In the old version, wasn't this only call `if (fvphyshack)`? Now is it being called all the time?

bartgol: Well, rrtmgp does not know about what pg2 is. It simply fills the list of trace gases. Then, it's up to dyn what to do with this. If phys=GLL, dyn won't do anything, otherwise it will.

tcclevenger (dismissed):
physhack stuff looks good to me.

components/eamxx/src/physics/share/scream_trcmix.cpp
```diff
@@ -14,6 +14,7 @@ namespace physics {
 
 void trcmix(
   const std::string& name,
+  const int nlevs,
```
brhillman, line 17: If we are going to pass nlevs, why not also pass ncols for consistency?

components/eamxx/src/physics/share/scream_trcmix.hpp
```diff
@@ -16,8 +16,10 @@ using trcmix_view1d = typename ekat::KokkosTypes<DefaultDevice>::template view_1
 template<typename Scalar>
 using trcmix_view2d = typename ekat::KokkosTypes<DefaultDevice>::template view_2d<Scalar>;
 
+// NOTE: nlevs be smaller than pmid/q extent, in case one/both of them are padded
```
brhillman, line 19: Typo: nlevs *can* be smaller

components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
```diff
@@ -564,6 +576,50 @@ void RRTMGPRadiation::run_impl (const double dt) {
     shr_orb_decl_c2f(calday, eccen, mvelpp, lambm0,
                      obliqr, &delta, &eccf);
 
+    // Precompute VMR for all gases, on all cols, before starting the chunks loop
```
brhillman, line 579: I think at some point we should maybe think about moving this out of radiation, or at least moving out of run_impl to clean this up. I think I could do this when I tackle generalizing the diagnostic rad calls (clear sky, clean sky, etc).

brhillman:
Looks good to me. I left a few comments, but probably nothing that needs to be changed.

brhillman (approved):

components/eamxx/src/physics/share/scream_trcmix.cpp
bartgol: `ncols` can safely be inferred from the view dimensions. `nlevs` on the other hand cannot be inferred, since we don't know if the views are padded or not.

components/eamxx/src/physics/share/scream_trcmix.hpp
bartgol: To avoid another round of testing and approval (in order to merge with all checks passed), I'll fix it in my next PR.

PR #2638, opened 2023-12-14
Change STOP_N for rad_frequency_2 cime test

bartgol:
In view of PR that forces restarts to coincide with rad steps, we change the time horizon of the rad supercycling test, so that when we test PR #2636 we don't risk to ignore DIFFs for the wrong reason.


bartgol:
DIFF is expected (we run the code one extra step).

brhillman (approved):

PR #2640, opened 2023-12-14
Add "append" attribute to namelist defaults

bartgol:
The new parameter is meant to simplify the input file in case of very long arrays, whose entries depend on case settings. As an example, consider this parameter
```
<my_param type="array(string)">string1, ..., string 10</my_param>
<my_param hgrid="foo">string1, ..., string 10, string11, string12</my_param>
<my_param COMPSET="ONE" hgrid="foo">string1, ..., string 10, string11, string13, string14</my_param>
<my_param COMPSET="TWO" hgrid="foo">string1, ..., string 10, string11, string15, string16</my_param>
<my_param COMPSET="ONE" hgrid="bar">string1, ..., string 10, string12, string13, string14</my_param>
<my_param COMPSET="TWO" hgrid="bar">string1, ..., string 10, string12, string15, string16</my_param>
```
This can easily make the nameslist defaults xml file long. With this PR, we can now do
```
<my_param type="array(string)">string1, ..., string 10</my_param>
<my_param hgrid="foo" append="base">string11</my_param>
<my_param hgrid="bar" append="base">string12</my_param>
<my_param COMPSET="ONE" append="last">string13, string14</my_param>
<my_param COMPSET="TWO" append="last">string15, string16</my_param>
```
The value `append="base"` is probably not needed in most cases (all?), since you can probably rearrange the entries in such a way that `last` will be ok. But I can imagine we may have lots of selectors, with so many combinations that you _may_ want to "reset" to the base case before appending new params, to make life easier. If that never happens, we may consider removing the 'base' version, and only have `append="true"`.

Note: this new attribute is only supported if `type="array(T)"` for some T (I didn't think it made sense for other parameters.

### DIFF
--- a/components/eamxx/cime_config/buildlib_cmake
+++ b/components/eamxx/cime_config/buildlib_cmake
@@ -25,9 +25,16 @@ def buildlib(bldroot, installpath, case):
     expect (len(tokens) % 2 == 0, "Error! SCREAM_CMAKE_OPTIONS should contain a string of the form 'option1 value1 option2 value2 ...'\n")
     it = iter(tokens)
 
-    cmake_args = ""
+    # Parse all options and put them in a dict first. This allows to overwrite options via
+    #  ./xmlchange --append SCRAM_CMAKE_OPTIONS="NAME VALUE"
+    # rather than having to reset them all, running ./xmlquery first to see what the others are
+    cmake_args_dict = {}
     for item in it:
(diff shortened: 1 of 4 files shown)


jgfouca (approved):

bartgol:
Since Jim approved, and the last commit was just a pylint fix, I'm going to merge.

PR #2641, opened 2023-12-16
EAMxx, Ascent: Work around apparent --fmad=false bug.

ambrad:
Nondeterminism on Ascent in RRMTGP in debug CIME-based tests appears to result from --fmad=false, counterintuitively. Possibly there is a bug in nvcc. In any case, work around this issue.

### DIFF
--- a/components/eamxx/src/dynamics/homme/CMakeLists.txt
+++ b/components/eamxx/src/dynamics/homme/CMakeLists.txt
@@ -117,7 +117,22 @@ macro (CreateDynamicsLib HOMME_TARGET NP PLEV QSIZE)
     if (HOMME_ENABLE_COMPOSE)
       target_link_libraries (${hommeLibName} composec++)
     endif()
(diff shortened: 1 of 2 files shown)


ambrad:
I'm not sure I want to merge this, but I want at least to check how the AT tests do. I might check if knowledge of the machine is available at the build-system level at which this workaround is done.

bartgol:
Diffs are likely due to the upstream merge. I just launched the job to regenerate baselines. We'll see what the next AT run says.

ambrad:
I've added a commit to isolate the workaround to just Ascent.

ambrad:
ERS_D_Ln21_P1x1.ne4pg2_ne4pg2.F2010-SCREAMv1.ascent_gnugpu.scream-rad_frequency_2 passes ~36 times in a row now (still running), whereas before it would fail in almost always <4 tries.

bartgol (approved):

PR #2644, opened 2023-12-21
Allow to construct a Field from a pre-existing view

bartgol:
This can be handy when we have pre-existing views, but still want to use routines/data-structures that require a Field.

### DIFF
--- a/components/eamxx/src/share/field/field_impl.hpp
+++ b/components/eamxx/src/share/field/field_impl.hpp
@@ -10,8 +10,7 @@
 namespace scream
 {
 
-template<typename ViewT,
-         typename = typename std::enable_if<Kokkos::is_view<ViewT>::value>::type>
+template<typename ViewT, typename>
 Field::
 Field (const identifier_type& id,
        const ViewT& view_d)
--- a/components/eamxx/src/share/field/field_impl.hpp
+++ b/components/eamxx/src/share/field/field_impl.hpp
@@ -5,8 +5,85 @@
 #include "share/util/scream_array_utils.hpp"
 #include "share/util/scream_universal_constants.hpp"
 
+#include <ekat/ekat_type_traits.hpp>
+
 namespace scream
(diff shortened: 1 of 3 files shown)


components/eamxx/src/share/field/field_impl.hpp
```diff
@@ -5,9 +5,86 @@
...
+        " - field name: " + id.name() + "\n"
+        " - layout last dim: " + std::to_string(fl.dims()[N-1]) + "\n"
+        " - view last dim: " + std::to_string(view_d.extent(N-1)) + "\n"); 
+
+    // We have a padded view. We don't know what the pack size was, so we pick the largest
+    // power of 2 that divides the last extent
+    auto last_view_dim = view_d.extent_int(N-1);
+    int last_fl_dim    = fl.dims().back();
+
+    // This should get the smallest pow of 2 that gives npacks*pack_size==view_last_dim
```
AaronDonahue, line 60: Won't this always be pow2 = 1?  Since no matter what a packsize of 2 should always fit?

bartgol: No. Say that the physical dim is 24, and the view was allocated with pack size 16, so it has a last extent 32. With `ps=2` you get
```
npacks = (24 + 2 - 1) / 2 = 12;
packed_length = 12 * 2 = 24;
```
Since `24!=36`, `ps=2` is not accepted. Everything up to 8 is rejected, since with `ps=1,2,4,8` you can fit 24 exactly, with no padding, so you would get `packed_length=24`. Only for `ps=16` you start to have padding, and you indeed get 2 packs, with `packed_length = 16 * 2 = 32`, which matches the view extent, and is therefore accepted.

bartgol:
The unit test seems to pass on mappy when I run it, so this is ready for final testing and review.

AaronDonahue (approved):

PR #2646, opened 2023-12-22
Refactor of SPA data structures

### LINKED ISSUE #2645: Switch SPA remap to use the Refining/Coarsening remappers, rather than HorizontalMap
Overall, the Refining/Coarsening remappers are much better, since they do things on device (using a CRS storage for the weights matrix).

In general, this is not that important for "normal (long) runs", since the remap only happens when a new slice is read in, which is once per simulated month. However, for short runs, this becomes expensive. Consider scream-autotuning runs, where we run for 2 days. We need to remap two sets of spa data (start of first month and start of second month). @ndkeen can post the exact data, but it is definitely non-negligible.

The solution is to switch to the Coarsening/Refining remappers, depending on the model res compared to the spa data res (most of the time it will be Refining, but for ultra-low-res testing, we can use coarsening).

bartgol:
This PR addresses issue #2645. In particular,

1. SPA class now stores a permanent `AtmosphereInput` object, so it doesn't have to re-create it every time it needs new data.
2. The horiz remap is now handled via an `AbstractRemapper`. There are 3 scenarios:
	1.	`ncols_data>ncols_model`: ~in this case, the remapper will be a `CoarseningRemapper`~ I changed the code to *not* allow this case. It is a corner case, probably only important for testing (where we currently use a single ne30 spa data file even for ne4 runs).
	2.	`ncols_data<ncols_model`: in this case, the remapper will be a `RefiningRemapperP2P`.
	3.	`ncols_data=ncols_model`: in this case, the remapper will be a `IdentityRemapper`. In this case, we also alias src and tgt field in the remapper, to a) avoid waste of memory, b) make the call to `remap` a no-op.
	In cases 1-2, the internal impl of the remapper is faster than the old `HorizontalMap` implementation, since it does most of the calculation (the mat-vec product for the interpolation) on device.
3. When updating the SPA state, we don't need to read in 2 states (the beg and end of the time window); we can simply copy the current "end" into "beg", and load a new state in "end".

Most of the mods are not really important in regular SCREAM runs, where dt is so small that these ops happen rarely. However, for very short runs, the I/O and remap costs are non-negligible.

Closes #2645.


bartgol:
I was getting an odd fail in testing, and realized that SPA _can_ use the coarsening remapper in a way that is *not* compatible with its original design. In particular, it can use it if the spa file has more cols than the model. But the coarsening remapper was meant to be used when the fine grid was the model grid, while here it's the opposite.

I'm working on an extension of the remapper, which should cover this. I suspect it won't be super performant, but I don't think it's a big deal, considering that we run in this scenario only when running ne<=30: our spa data is on ne30, gll grid, so running the model with ne<30 or ne30pg2 yields less cols, triggering the "backward coarsening.

I hope to finish this tomorrow, so that I don't need to reboot my thought process after the break...

bartgol:
@AaronDonahue I was discussing with @mahf708 offline, and we were wondering: how restrictive is it to _require_ that spa data is at the same or lower resolution as the model? Notice that if the model uses pg2, I would require spa data to be pg2 as well. So the current spa data (at ne30np4) would not be usable at all with our pg2.

I am having some issues with trying to use CoarseningRemapper "backward" (meaning where the model grid is the coarse one, rather than the fine one). I think restricting us to the case where spa data is coarse and model is fine would make life much easier...

bartgol:
An additional comment: why don't we use ne30pg2 data for spa? It would be a smaller file, and so would the map file. E.g., from the e3sm data server,

```
map_ne30np4_to_ne1024pg2_intbilin_20221012.nc      12-Oct-2022 17:18      4G
map_ne30pg2_to_ne1024pg2_20231201.nc               01-Dec-2023 23:33      3G
```
(the latter is actually ~2.7G, and the former is 4.0G).

bartgol:
> You should check the actual datafile though... lemme get the numbers... nvm, I deleted them to avoid spamming the inputdata server. They're about 50% less expensive (the ne30np4 file is ~9G while the pg2 file is ~4.5G)

Which makes sense, since pg2 has 4/9th of the data of gll.

AaronDonahue:
>An additional comment: why don't we use ne30pg2 data for spa? It would be a smaller file, and so would the map file. E.g., from the e3sm data server,
@bartgol I think this is absolutely reasonable.  I believe the current reason we are on np4 is just based on how the SPA data was created in the first place, using an ne30np4 simulation.  That being said, I don't see any reason why we couldn't process the files and regrid from np4 to pg2 and call it a day.  @hassanbeydoun , do you see any reason why we would want SPA data to stay on the Gauss grid (i.e. np4)?

AaronDonahue:
>@AaronDonahue I was discussing with @mahf708 offline, and we were wondering: how restrictive is it to require that spa data is at the same or lower resolution as the model? 

I would give @hassanbeydoun the final word on this since it is his model we are using.  But first impression, I don't see why this would be a problem.  When running an ne4 simulation there really isn't any benefit in sticking with ne30 for SPA data.  It just means using a big file and a big remapper in the SCREAM run.  And remapping to ne4 offline is pretty easy.

The only argument I would see for allowing ne30 SPA data in an ne4 run would be for cases where we make new SPA data and want to allow for rapid deployment in our runs.  But again, since remapping from ne30 to ne4 is relatively easy I don't think this is a good arg.

bartgol:
After talking with Aaron, I decided to enforce the requirement that SPA data must be on the same grid as the model, or on a coarser grid. The reason for this is that using the CoarseningRemapper in a configuration where the model grid is the coarse one rather than the fine one makes for a lot of headaches. I tried to accommodate it, but kept running into issues. Since in practice, spa will always be on a coarse grid, I think this is ok. If we run into a scenario where SPA data is on a grid finer than the model, we will have to pre-process the spa data file, coarsening down to the model grid (or coarser).

In terms of init costs, I ran an ne30pg2 case on mappy, using SPA data from ne4pg2, remapped online up to ne30pg2. These are the existing SPA timers.

master
```
"a_i:EAMxx::spa::init"                                       8.533
"a_i:EAMxx::SPA::get_remap_weights_from_file"                0.075
"a_i:EAMxx::SPA::update_spa_data_from_file"                  8.440
"a_i:EAMxx::SPA::update_spa_data_from_file::read_data"       8.258
"a_i:EAMxx::SPA::update_spa_data_from_file::apply_remap"     0.141
"a_i:EAMxx::SPA::update_spa_data_from_file::copy_and_pad"    0.058
"a:EAMxx::spa::run"                                          0.920
```

branch
```
"a_i:EAMxx::spa::init"                                       0.049
"a_i:EAMxx::SPA::update_spa_data_from_file"                  0.049
"a_i:EAMxx::SPA::update_spa_data_from_file::read_data"       0.022
"a_i:EAMxx::SPA::update_spa_data_from_file::horiz_remap"     0.021
"a_i:EAMxx::SPA::update_spa_data_from_file::copy_and_pad"    0.012
"a:EAMxx::spa::run"                                          1.041
"a:EAMxx::SPA::update_spa_data_from_file"                    0.042
"a:EAMxx::SPA::update_spa_data_from_file::read_data"         0.016
"a:EAMxx::SPA::update_spa_data_from_file::horiz_remap"       0.017
"a:EAMxx::SPA::update_spa_data_from_file::copy_and_pad"      0.012
```
It is not an apples-apples comparison, since the branch is using spa data on a much coarser grid (you can tell the difference in the `read_data` timer). Hence, we need to test this in a scenario where master and branch are on the same ground. E.g., we can run ne1024, using the existing ne30 spa data (it is coarser than the model grid, so it's ok).

bartgol:
The pg2 test is quite bizarre. It fails due to a postcondition check in p3, where `nc` is out of the interval `[0,1e11]`, with a value of `5e11`. The bizarre part is that if I remove `nccn` from the list of output vars, the post condition does not fail.

This is quite suspicious, since output vars should _NOT_ affect the state of the problem. @AaronDonahue , do you have any clue what the issue could be here?

Note: I changed the pg2 unit test to be without SPA, since I did not want to create an ne2pg2 spa input file (recall we no longer allow to coarsen spa data, only refine).

bartgol:
I think I know what's happening: when doing INSTANT output, we output t=0 data, and we must allow some invalid data (since the atm did not run yet, so some fields may be invalid). In this process, we fill the fields with the fillValue. Since I did not specify `do_prescribed_ccn: false` in the input, p3 assumes it is prescribed, which means it is an input field. I forgot to add it to the init list (e.g., to 0), so its timestep was never set, hence IO considered it invalid and overwrote its value with fillValue. Adding `nccn: 0` in the IC sublist fixed the problem.

bartgol:
I am still not positive that things are correct. In the `homme_shoc_cld_p3_rrtmgp_pg2` test I get lots of warnings regarding qv being repaired during shoc:
```
WARNING: Failed and repaired post-condition property check.
  - Atmosphere process name: shoc
  - Property check name: qv within interval [0, 0.2]
  - Message: Check failed.
  - check name: qv within interval [0, 0.2]
  - field id: qv[Physics PG2] <double:ncol,lev>(96,72) [kg/kg]
  - minimum:
    - value: -1.69407e-21
    - indices (w/ global column index): (78,32)
    - lat/lon: (-54.744, 72.5463)
  - maximum:
    - value: 0.0170544
    - indices (w/ global column index): (39,71)
    - lat/lon: (-11.2246, 191.446)
```
This error does not show up in the np4 case. Furthermore, I get the slightly more worrying warning
```
rrtmgp_main::t_lev: 96.0000 values outside range [160.000,355.000]; minval = 0.00000; maxval = 300.342
```
It's a tad cold when the minimum temperature reaches 0K...

bartgol:
Ok, I found out the issues, and they all related to the lack of values for certain surface fields. The problem was present in master already, for the pg2 unit test, even with spa on.

In short, while our IC files _do_ have data for the surf fields that usually come from the surface coupling, the PG2 handling of IC does not allow to remap them from GLL to PG2. That's b/c PG2 was developed for CIME cases, where we always get the surf data from the coupler, so we don't have to read it from the atm IC file.

I circumvent this issue by setting some fields to a constant value, whose value I chosed by looking at the values stored in the IC file for the corresponding field, and picking something in the right range.

bartgol:
The DIFFs are expected, since we are changing the way we do interpolation in SPA. The noAero compset is BFB, as expected.

components/eamxx/src/share/grid/remap/horiz_interp_remapper_base.cpp
```diff
@@ -61,23 +61,24 @@ create_src_layout (const FieldLayout& tgt_layout) const
...
     case LayoutType::Scalar3D:
       src = m_src_grid->get_3d_scalar_layout(midpoints);
       break;
     case LayoutType::Vector3D:
-      src = m_src_grid->get_3d_vector_layout(midpoints,CMP,vec_dim);
+      src = m_src_grid->get_3d_vector_layout(midpoints,vec_tag,vec_dim);
       break;
     default:
-      EKAT_ERROR_MSG ("Layout not supported by CoarseningRemapper: " + e2str(lt) + "\n");
+      EKAT_ERROR_MSG ("Layout not supported by HorizInterpRemapperBase: " + e2str(lt) + "\n");
```
AaronDonahue, line 81: Could we add `Target` to the beginning of this message and `Source` to the one below?  That way it is easier to see which of these two function threw the error.

components/eamxx/src/share/grid/remap/horiz_interp_remapper_base.cpp
```diff
@@ -203,9 +205,30 @@ get_my_triplets (const std::string& map_file) const
   scorpio::set_dof(map_file,"S"  ,nlweights,dofs_offsets.data());
   scorpio::set_decomp(map_file);
 
-  scorpio::grid_read_data_array(map_file,"col",-1,cols.data(),cols.size());
-  scorpio::grid_read_data_array(map_file,"row",-1,rows.data(),rows.size());
-  scorpio::grid_read_data_array(map_file,"S"  ,-1,S.data(),S.size());
+  // Figure out if we are readint the right map, that is:
```
AaronDonahue, line 208: typo in "reading"

AaronDonahue: also typo in "matches" below.

AaronDonahue (approved):
Thank you for all of these improvements in SPA.  I had a few comments, but they are non-blocking.  So I am approving.

bartgol:
As suspected, the speedup is more noticeable on GPU. I went on weaver and ran the `homme_shoc_cld_spa_p3_rrtmgp` test on an ne4np4 grid, with ne2np2 SPA data (to trigger remapping), in both master and this branch, with a single MPI rank. The relevant spa timers are as follows:

master
```
EAMxx::spa::init                                     -          1        1 1.000000e+00   9.988400e-01     0.999 (     0      0)     0.999 (     0      0)
EAMxx::SPA::get_remap_weights_from_file              -          1        1 1.000000e+00   5.383470e-01     0.538 (     0      0)     0.538 (     0      0)
EAMxx::SPA::update_spa_data_from_file                -          1        1 2.000000e+00   4.534800e-01     0.453 (     0      0)     0.453 (     0      0)
EAMxx::SPA::update_spa_data_from_file::read_data     -          1        1 2.000000e+00   2.355700e-01     0.236 (     0      0)     0.236 (     0      0)
EAMxx::SPA::update_spa_data_from_file::apply_remap   -          1        1 2.000000e+00   2.144670e-01     0.214 (     0      0)     0.214 (     0      0)
EAMxx::SPA::update_spa_data_from_file::copy_and_pad  -          1        1 2.000000e+00   5.320000e-04     0.001 (     0      0)     0.001 (     0      0)
EAMxx::spa::run                                      -          1        1 2.400000e+01   2.411800e-02     0.024 (     0      0)     0.024 (     0      0)
```
branch
```
EAMxx::SPA::create_horiz_remapper                    -          1        1 1.000000e+00   1.378930e-01     0.138 (     0      0)     0.138 (     0      0)
EAMxx::SPA::create_data_reader                       -          1        1 1.000000e+00   5.802300e-02     0.058 (     0      0)     0.058 (     0      0)
EAMxx::spa::init                                     -          1        1 1.000000e+00   7.960200e-02     0.080 (     0      0)     0.080 (     0      0)
EAMxx::SPA::update_spa_data_from_file                -          1        1 2.000000e+00   1.624780e-01     0.162 (     0      0)     0.162 (     0      0)
EAMxx::SPA::update_spa_data_from_file::read_data     -          1        1 2.000000e+00   1.308480e-01     0.131 (     0      0)     0.131 (     0      0)
EAMxx::SPA::update_spa_data_from_file::horiz_remap   -          1        1 2.000000e+00   2.969200e-02     0.030 (     0      0)     0.030 (     0      0)
EAMxx::SPA::update_spa_data_from_file::copy_and_pad  -          1        1 2.000000e+00   1.925000e-03     0.002 (     0      0)     0.002 (     0      0)
EAMxx::spa::run                                      -          1        1 2.400000e+01   1.071000e-01     0.107 (     0      0)     0.107 (     0      0)
```

Some considerations
- It is not immedate why `read_data` should be faster, but upon inspection, one realizes that in master we were creating an AtmosphereInput object every time, while here we always recycle the same one (saving on pio setup times, views allocations, cleanup,...).
- The run timer has gone up. But this is simply because we no longer read the first two months at init time. In order to keep the code generic, we read only the first month at init time, storing it in the data structure for "t_end" (leaving the one for "t_beg" empty). Then, at run time, if the input time is past "t_end", we move the spa data of t_end into the spa data of t_beg, and load a new month for t_end. Hence, the 1st call of `SPA::run` now includes roughly half of the cost of `update_spa_data_from_file`, so ~63ms. Taking this out of the timer, we get 24ms, just like in master, corresponding to the cost of running `spa_main`.
- The remapper and IO reader objects are created inside `SPA::set_grids`, since they don't depend on any FM quantity.
- We no longer have the `get_remap_weights_from_file`. This phase is basically incorporated in the construction of the horiz remapper.
- Applying the remapper is ~85% faster, since the sparse mat-vec product is done on GPU. With one rank, the comm part is fast, since it's on the same processor.
- The SPA::init cost on master includes the remap setup, and two calls to `update_spa_data_from_file`. To compare, we can consider in the branch `create_horiz_remapper+create_data_reader+init+2*update_spa_data_from_file`. This gives 700ms for the branch, which means a 30% reduction of the SPA init cost compared to master. Note: the timers `create_horiz_remapper` and `create_data_reader` are not in the branch, I simply added them momentarily to perform this comparison.

Note: these are ne4 timers...We need to run some more realistic scenario on a larger node count to get a feeling of the actual saving for production runs. But the bottom line is that we expect things to be faster.

bartgol:
@AaronDonahue How can we check that the DIFFs are indeed benign?  Looking at CPRNC output, I can't be sure everything is fine, since the diffs are large.

What I am concerned about is the possibility of some accidental aliasing of internal data structure, or maybe something wrong when reading new spa data and "shifting" t_end->t_beg. I will think of some unit tests for this, but meanwhile, can we run an ne30 case for a while (ideally 1y, but maybe a few days, crossing a month boundary, may be enough) and check that the output looks reasonable, compared to master? I don't know how to check climate...

bartgol:
@AaronDonahue Another question: if we merge this, there will no longer be any customer of the horiz_remap_utility stuff. Are you ok removing it altogether?

AaronDonahue:
> @AaronDonahue How can we check that the DIFFs are indeed benign? Looking at CPRNC output, I can't be sure everything is fine, since the diffs are large.
> 
> What I am concerned about is the possibility of some accidental aliasing of internal data structure, or maybe something wrong when reading new spa data and "shifting" t_end->t_beg. I will think of some unit tests for this, but meanwhile, can we run an ne30 case for a while (ideally 1y, but maybe a few days, crossing a month boundary, may be enough) and check that the output looks reasonable, compared to master? I don't know how to check climate...

Yes, I can start that now.  And then post comparisons of the two climates on this PR for discussion.

AaronDonahue:
> @AaronDonahue Another question: if we merge this, there will no longer be any customer of the horiz_remap_utility stuff. Are you ok removing it altogether?

Absolutely.  If there is no one using the code we should cut it.

components/eamxx/cime_config/namelist_defaults_scream.xml
```diff
@@ -274,19 +274,25 @@ be lost if SCREAM_HACK_XML is not enabled.
...
       <spa_remap_file hgrid="ne120np4.pg2">${DIN_LOC_ROOT}/atm/scream/maps/map_ne30np4_to_ne120pg2_intbilin_20221012.nc</spa_remap_file>
       <spa_remap_file hgrid="ne256np4.pg2">${DIN_LOC_ROOT}/atm/scream/maps/map_ne30np4_to_ne256pg2_intbilin_20221011.nc</spa_remap_file>
       <spa_remap_file hgrid="ne512np4.pg2">${DIN_LOC_ROOT}/atm/scream/maps/map_ne30np4_to_ne512pg2_intbilin_20221012.nc</spa_remap_file>
       <spa_remap_file hgrid="ne1024np4.pg2">${DIN_LOC_ROOT}/atm/scream/maps/map_ne30np4_to_ne1024pg2_intbilin_20221012.nc</spa_remap_file>
 
-      <spa_data_file type="file">${DIN_LOC_ROOT}/atm/scream/init/spa_file_unified_and_complete_ne30_20220428.nc</spa_data_file>
+      <spa_data_file type="file"
+        doc="File containing aerosol data. Must be on same grid as the atm, or a coarser one"
+        >${DIN_LOC_ROOT}/atm/scream/init/spa_file_unified_and_complete_ne30_20220428.nc</spa_data_file>
+      <spa_data_file hgrid="ne30np4.pg2" type="file">${DIN_LOC_ROOT}/atm/scream/init/spa_file_unified_and_complete_ne4pg2_20231222.nc</spa_data_file>
```
AaronDonahue, line 293: We should probably remap the ne30np4 data to ne30pg2 and use it here.  ne4pg2 is significantly smoothed compared to ne30np4.

AaronDonahue:
This looks good to me.  I'm waiting on the monthly ne30 tests that @bartgol requested before approving.  @PeterCaldwell and/or @bartgol let me know if you want me to approve earlier than that so that this can get into master sooner than later.

components/eamxx/cime_config/namelist_defaults_scream.xml
bartgol: Yeah, I was trying to trigger a remap in test though. Perhaps I can use a testmod, to force a nightly test to use ne4pg2 data, and remap...

PeterCaldwell:
Yeah, an extra day or two would be fine.

bartgol:
Testing is still good after the last commits, with standalone passing, and v1 diffing only for compsets with SPA on. I will do manual testing, to ensure SPA updates correctly when crossing a month, and @AaronDonahue will verify that the horiz remapper is doing the correct thing. Once that's done, we can merge.

bartgol:
I manually verified that, upon month change, the SPA states are correctly updated, meaning that spa_data_end is swapped with spa_data_beg, and a the new month is loaded into spa_data_end. I will now verify that the remapping procedure produces fields similar to the data.

bartgol:
I plotted some aero_XYZ fields from an ne4 run and an ne30 run (which uses ne4 data with remap). The two are very similar in terms of structures. The ne30 one has some topography features visible near the surface for `aero_ssa_sw` and `aero_g_sw`, but that makes sense since SPA also performs vinterp, and the difference in topography creates some visible different features over Himalayaa and Andes (see below: ne30 is left, ne4 is right). The vars `aero_tau_lw` and `aero_tau_sw` are identical even near surface, since from what I can tell they have very small values above land.
aero_ssa_sw_bnd3_lvl30
aero_ssa_sw_bnd1_lvl60

@AaronDonahue I think we can merge this. Any objection?

AaronDonahue (approved):

PR #2648, opened 2023-12-29
EAMxx: See if removing --fmad=false fixes nondeterminism in pm-gpu nightlies.

ambrad:
pm-gpu is showing nondeterminism in debug builds. Try the `--fmad=false` fix that works on Ascent. I'm not going to do any analysis on pm-gpu right now; I'm just seeing if I get lucky.


ambrad:
I'm going to merge this without review.

ambrad:
Update: It looks like this workaround indeed fixes the restart-compare failures on pm-gpu.

PR #2650, opened 2024-01-03
time-varying inputs for radiation

mahf708:
GHG file for SCREAM: `atm/cam/ggas/GHG_CMIP-1-2-0_Annual_Global_0000-2014_scream_c20240104.nc`

### DIFF
--- a/components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
+++ b/components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
@@ -389,12 +389,12 @@ void RRTMGPRadiation::initialize_impl(const RunType /* run_type */) {
   // TODO: Harden this impl; get the EKAT msg/warn stuff and use it here to avoid bad logic at the outset
   // TODO: Maybe also move this to the very top?
   m_time_varying_active_gases_list =
-      m_params.get<std::vector<std::string>>("time_varying_active_gases_list", {"no-list-given"});
+      m_params.get<std::vector<std::string>>("time_varying_active_gases_list", {"nolistgiven"});
   m_time_varying_active_gases_file =
-      m_params.get<std::vector<std::string>>("time_varying_active_gases_file", {"no-file-given"});
+      m_params.get<std::vector<std::string>>("time_varying_active_gases_file", {"nofilegiven"});
 
   // Initialize the time interpolator
-  if (m_time_varying_active_gases_file[0] != "no-file-given") {
+  if (m_time_varying_active_gases_file[0] != "nofilegiven") {
     m_time_interp =
         util::TimeInterpolation(m_grid, m_time_varying_active_gases_file);
 
@@ -406,7 +406,7 @@ void RRTMGPRadiation::initialize_impl(const RunType /* run_type */) {
       if (name == "co2vmr") {
         FieldLayout scalar0d_layout{{}, {}};
         Field field_transient = create_helper_field(
-            name_transient, scalar0d_layout, m_grid->name(), ps);
+            name_transient, scalar0d_layout, m_grid->name());
       } else {
         // Not supported, ignore
         // TODO: get a proper logger/warner here (EKAT msg?) 
@@ -584,11 +584,12 @@ void RRTMGPRadiation::run_impl (const double dt) {
   auto ts = timestamp();
   auto update_rad = scream::rrtmgp::radiation_do(m_rad_freq_in_steps, ts.get_num_steps());
 
+  if (m_time_varying_active_gases_file[0] != "nofilegiven") {
+    // Perform time interpolation
+    m_time_interp.perform_time_interpolation(ts);
+  }
+
   if (update_rad) {
-    if (m_time_varying_active_gases_file[0] != "no-file-given") {
-      // Perform time interpolation
-      m_time_interp.perform_time_interpolation(ts);
-    }
     // On each chunk, we internally "reset" the GasConcs object to subview the concs 3d array
     // with the correct ncol dimension. So let's keep a copy of the original (ref-counted)
     // array, to restore at the end inside the m_gast_concs object.
@@ -646,7 +647,7 @@ void RRTMGPRadiation::run_impl (const double dt) {
         });
         Kokkos::fence();
       } else {
-        if (m_time_varying_active_gases_file[0] != "no-file-given") {
+        if (m_time_varying_active_gases_file[0] != "nofilegiven") {
           // For time varying inputs, we should read them from the interpolator
           // The interpolator creates helper fields with _transient suffix
           for (std::string name : m_time_varying_active_gases_list) {
@@ -1207,8 +1208,10 @@ void RRTMGPRadiation::finalize_impl  () {
   m_gas_concs.reset();
   rrtmgp::rrtmgp_finalize();
 
-  // Finalzie time interpolator
-  m_time_interp.finalize();
+  // Finalize time interpolator
+  if (m_time_varying_active_gases_file[0] != "nofilegiven") {
+    m_time_interp.finalize();
+  }
   // Finalize YAKL
   yakl_finalize();
 }
--- a/components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
+++ b/components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
@@ -385,30 +385,31 @@ void RRTMGPRadiation::initialize_impl(const RunType /* run_type */) {
   // Not sure what to do about this
   constexpr int ps = SCREAM_PACK_SIZE;
 
-  // Get time-varying input fields
-  m_time_varying_input_fields_datafile = m_params.get<std::vector<std::string>>(
-      "time_varying_input_fields_datafile");
-  m_time_varying_input_fields =
-      m_params.get<std::vector<std::string>>("time_varying_input_fields");
+  // Get time-varying input fields and their associated file(s)
+  m_time_varying_active_gases_list =
+      m_params.get<std::vector<std::string>>("time_varying_active_gases_list");
+  m_time_varying_active_gases_file =
+      m_params.get<std::vector<std::string>>("time_varying_active_gases_file");
 
   // Initialize the time interpolator
   m_time_interp =
-      util::TimeInterpolation(m_grid, m_time_varying_input_fields_datafile);
+      util::TimeInterpolation(m_grid, m_time_varying_active_gases_file);
 
   // To create helper fields for to interpolate their values
-  for(std::string name : m_time_varying_input_fields) {
+  for (std::string name : m_time_varying_active_gases_list) {
     // Helper fields that will temporarily store the input values
     std::string name_transient = name + "_transient";
     Field field_transient;
-    if(name == "co2vmr") {
+    if (name == "co2vmr") {
       FieldLayout scalar0d_layout{{}, {}};
       Field field_transient = create_helper_field(
           name_transient, scalar0d_layout, m_grid->name(), ps);
     } else {
       // Not supported, ignore
+      // TODO: get a proper logger/warner here (EKAT msg?) 
       std::cout << "Skipping time-varying input field: " << name << std::endl;
-      // For O3, we can do the following, but more work is needed downstream
       // TODO: Add time-varying O3 when/if desired
+      // For O3, we can do the following, but more work is needed downstream
       // auto field = create_helper_field(name, scalar3d_layout_mid, m_grid,
       // ps);
     }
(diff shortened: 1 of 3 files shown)


components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
```diff
@@ -105,9 +108,13 @@ void RRTMGPRadiation::set_grids(const std::shared_ptr<const GridsManager> grids_
   // Set of required gas concentration fields
   for (auto& it : m_gas_names) {
     // Add gas VOLUME mixing ratios (moles of gas / moles of air; what actually gets input to RRTMGP)
-    if (it == "o3") {
-      // o3 is read from file, or computed by chemistry
-      add_field<Required>(it + "_volume_mix_ratio", scalar3d_layout_mid, molmol, grid_name);
+    if (it == "o3" && !m_transient_ozone) {
+      if (!m_transient_ozone) {
```
brhillman, line 112: We should probably call this `m_prescribed_ozone` instead of transient, and then always require this input is present until we have ozone computed by chemistry. I want to completely abandon the practice of sticking it in the initial condition file and remove it from the ICs.

brhillman, line 112: It looks like you have redundant logicals here...`if(!m_transient_ozone)` is inside the `if (it == "o3" && !m_transient_ozone)` block.

components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
```diff
@@ -382,6 +402,48 @@ void RRTMGPRadiation::initialize_impl(const RunType /* run_type */) {
   // Determine whether or not we are using a fixed solar zenith angle (positive value)
   m_fixed_solar_zenith_angle = m_params.get<double>("Fixed Solar Zenith Angle", -9999);
 
+  // Not sure what to do about this
+  constexpr int ps = SCREAM_PACK_SIZE;
```
brhillman, line 406: Do we need packs anywhere in this?

components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
```diff
@@ -382,6 +402,48 @@ void RRTMGPRadiation::initialize_impl(const RunType /* run_type */) {
...
+  // TODO: Maybe also move this to the very top?
+  m_time_varying_active_gases_list =
+      m_params.get<std::vector<std::string>>("time_varying_active_gases_list", {"nolistgiven"});
+  m_time_varying_active_gases_file =
+      m_params.get<std::vector<std::string>>("time_varying_active_gases_file", {"nofilegiven"});
+
+  // Initialize the time interpolator
+  if (m_time_varying_active_gases_file[0] != "nofilegiven") {
+    m_time_interp =
+        util::TimeInterpolation(m_grid, m_time_varying_active_gases_file);
```
brhillman, line 419: As a matter of style, I'd avoid this line break

components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
```diff
@@ -382,6 +402,48 @@ void RRTMGPRadiation::initialize_impl(const RunType /* run_type */) {
   // Determine whether or not we are using a fixed solar zenith angle (positive value)
   m_fixed_solar_zenith_angle = m_params.get<double>("Fixed Solar Zenith Angle", -9999);
 
+  // Not sure what to do about this
+  constexpr int ps = SCREAM_PACK_SIZE;
+
+  // Get time-varying input fields and their associated file(s)
+  // TODO: Harden this impl; get the EKAT msg/warn stuff and use it here to avoid bad logic at the outset
+  // TODO: Maybe also move this to the very top?
```
brhillman, line 410: This block does seems like something that should maybe be pulled out of the main run impl

components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
```diff
@@ -382,6 +402,48 @@ void RRTMGPRadiation::initialize_impl(const RunType /* run_type */) {
...
+      // Helper fields that will temporarily store the input values
+      std::string name_transient = name + "_transient";
+      Field field_transient;
+      if (name == "co2" || name == "n2o" || name == "ch4" || name == "f11" || name == "f12") {
+        // These are all scalar (0d) time-varying
+        FieldLayout scalar0d_layout{{}, {}};
+        field_transient = create_helper_field(
+            name_transient, scalar0d_layout, m_grid->name(), ps);
+      } else if (name =="o3") {
+        // TODO: add O3 later because we need to decide about remapping it from ne30pg2 to target
```
brhillman, line 432: Actually, now that I think about it, I think EAM has a prescribed O3 file option that it can use, where O3 is not assumed to exist on a 3d grid. I'll check on that, but maybe we can just steal that and not worry about remap from ne30pg2

components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
```diff
@@ -382,6 +402,48 @@ void RRTMGPRadiation::initialize_impl(const RunType /* run_type */) {
...
+    // To create helper fields for to interpolate their values
+    for (std::string name : m_time_varying_active_gases_list) {
+      // Helper fields that will temporarily store the input values
+      std::string name_transient = name + "_transient";
+      Field field_transient;
+      if (name == "co2" || name == "n2o" || name == "ch4" || name == "f11" || name == "f12") {
+        // These are all scalar (0d) time-varying
+        FieldLayout scalar0d_layout{{}, {}};
+        field_transient = create_helper_field(
+            name_transient, scalar0d_layout, m_grid->name(), ps);
```
brhillman, line 430: Is pack size an optional argument? You might be able to avoid packs here (nothing else in rad uses packs)

components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
brhillman: Actually, I think I misread this...

components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
```diff
@@ -823,10 +925,19 @@ void RRTMGPRadiation::run_impl (const double dt) {
       real2d tmp2d = subview_2d(m_buffer.tmp2d);
       for (int igas = 0; igas < m_ngas; igas++) {
         auto name = m_gas_names[igas];
-        auto full_name = name + "_volume_mix_ratio";
-
-        // 'o3' is marked as 'Required' rather than 'Computed', so we need to get the proper field
-        auto f = name=="o3" ? get_field_in(full_name) : get_field_out(full_name);
+        auto full_name = (name=="o3" && m_transient_ozone) ? name + "_transient" : name + "_volume_mix_ratio";
```
brhillman, line 928: If it hasn't become obvious from previous comments by now, I'm confused by the transient ozone flag. Is this just to allow hooking up prognostic ozone? If so, I vote we remove all of this extra future-proofing logic to simplify this interface.

components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
```diff
@@ -1109,12 +1220,45 @@ void RRTMGPRadiation::run_impl (const double dt) {
   }
 
 }
+
+// =========================================================================================
+// This is a copy of a the helper field functionality from nduging...
```
brhillman, line 1225: typo in nudging

components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
AaronDonahue: Yes, `ps` is optional I believe so you should be able to remove that line and not worry about defining the field as packed.  Note, that you can always request a view of Real from any field, packed or not.

components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
```diff
@@ -382,6 +402,48 @@ void RRTMGPRadiation::initialize_impl(const RunType /* run_type */) {
...
+        field_transient = create_helper_field(
+            name_transient, scalar0d_layout, m_grid->name(), ps);
+      } else if (name =="o3") {
+        // TODO: add O3 later because we need to decide about remapping it from ne30pg2 to target
+        // field_transient = create_helper_field(name, scalar3d_layout_mid, m_grid->name(), ps);
+        // m_transient_ozone = true;
+        std::cout << "Skipping time-varying input field: " << name << std::endl;
+      } else {
+        // Not supported, ignore
+        // TODO: get a proper logger/warner here (EKAT msg?)
```
AaronDonahue, line 438: You should be able to use the `log` function that is part of every atmosphere process.  See: https://github.com/E3SM-Project/scream/blob/master/components/eamxx/src/share/atm_process/atmosphere_process.hpp#L282

components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
```diff
@@ -545,6 +607,11 @@ void RRTMGPRadiation::run_impl (const double dt) {
   auto ts = timestamp();
   auto update_rad = scream::rrtmgp::radiation_do(m_rad_freq_in_steps, ts.get_num_steps());
 
+  if (m_time_varying_active_gases_file[0] != "nofilegiven") {
```
AaronDonahue, line 610: For simplicity of the code you could make a single bool variable that you initialize in the init_impl and use that for all of these logicals.  Maybe something like
```
m_time_vary_active_gas_from_file = m_time_varying_active_gases_file[0] != "nofilegiven" ? true : false;
```

components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
```diff
@@ -603,6 +670,41 @@ void RRTMGPRadiation::run_impl (const double dt) {
...
+                Kokkos::deep_copy(m_f12vmr, value_transient);
+              }
+            } else if (name=="o3") {
+              // Not applicable
+              // The user shouldn't give us n2 or co here anyway,
+              // but for for o3, it is handled separately later with
+              // an additional flag m_transient_ozone
+            } else {
+              // Not supported; warn and ignore
+              // TODO: get a proper logger/warner here (EKAT msg?)
```
AaronDonahue, line 702: See comment about using `log` above.

components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
```diff
@@ -1109,12 +1220,45 @@ void RRTMGPRadiation::run_impl (const double dt) {
   }
 
 }
+
+// =========================================================================================
+// This is a copy of a the helper field functionality from nduging...
+// We should abstract this out of these places to somewhere more shared...
+// TODO: Ask Luca what he prefers to do (somewhere in share/util maybe?)
+Field RRTMGPRadiation::create_helper_field(const std::string &name,
```
AaronDonahue, line 1228: We do have a way to create "internal" fields, see: https://github.com/E3SM-Project/scream/blob/master/components/eamxx/src/share/atm_process/atmosphere_process.hpp#L192, however there are other processes where an internal field is not appropriate and they too have a create_helper_field defined.  I think this task might be to consolidate internal and helper fields, if that is possible.

AaronDonahue: @bartgol ^^. Maybe you have some insight on the use of internal fields.

components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
```diff
@@ -382,6 +402,48 @@ void RRTMGPRadiation::initialize_impl(const RunType /* run_type */) {
...
+        util::TimeInterpolation(m_grid, m_time_varying_active_gases_file);
+
+    // To create helper fields for to interpolate their values
+    for (std::string name : m_time_varying_active_gases_list) {
+      // Helper fields that will temporarily store the input values
+      std::string name_transient = name + "_transient";
+      Field field_transient;
+      if (name == "co2" || name == "n2o" || name == "ch4" || name == "f11" || name == "f12") {
+        // These are all scalar (0d) time-varying
+        FieldLayout scalar0d_layout{{}, {}};
```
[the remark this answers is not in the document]
bartgol: This is subtle. The brace initialization causes some things to trigger. In this case
```c++
FieldLayout layout0d { {}, {} };
```
the compiler interprets the input as an initializer list of size 2. It then scans the constructors of `FieldLayout` and finds `FieldLayout (const std::initializer_list<FieldTag>& tags);`, so it converts the input to an initializer list of tags, of size 2. In this case, otoh,
```c++
FieldLayout layout1d { {LEV}, {nlevs} };
```
the compiler cannot parse the input as a single initializer list, since the two entries would have different type. Scanning the class, it finds
```c++
  FieldLayout (const std::vector<FieldTag>& tags, const std::vector<int>& dims);
```
and since it can construct the two vectors from the two inner init lists, it realizes it can call this constructor. Notice that, if you do
```c++
FieldLayout layout0d ( {}, {} );
```
the compiler can no longer interpret this as a single init list with two entries, and is forced to pick the constructor that takes two vectors (like for layout1d).

components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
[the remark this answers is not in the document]
bartgol: FWIW, I changed this check in another branch. Instead of checking the grid name, I made it to check that the field's grid name is a valid alias of the FM's grid. Not sure if it would fix the pb here.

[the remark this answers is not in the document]
bartgol: Also, I have been wanting to consolidate all FM's into a single FM. If I ever do that, then there will be no point in the check above, since the FM would store fields from all over the place, corresponding to whatever grids..

PR #2653, opened 2024-01-05
Add option to output a field at height above the surface.

AaronDonahue:
The current implementation of FieldAtHeight assumes the user wants to output the field at a specific height above sealevel.  In practice this means the reference height used for interpolation is either z_mid or z_int.

It is likely that users would also like the option to output a field at a specific height above the surface, for example the 2 meter temperature.

This commit adds that capability by including a reference surface to the FieldAtHeight diagnostic.  This can either be "sealevel" or "surface". If a user specifies "sealevel" then z_mid or z_int are used as the level heights.  If a user specifies "surface" than geopotential_mid or geopotential_int are used instead.

This is controlled by the diagnostic name.  The old way to trigger a FieldAtHeight was to list the variable X at N meters like follows:
  - X_at_Nm
  
 The new approach allows for

  - X_at_Nm_above_sealevel, or
  - X_at_Nm_above_surface.

If a user still uses the old way without specifying "above_Y" then it defaults to "sealevel" and prints a warning to the ATM log.

### DIFF
--- a/components/eamxx/src/diagnostics/field_at_height.cpp
+++ b/components/eamxx/src/diagnostics/field_at_height.cpp
@@ -42,8 +42,9 @@ FieldAtHeight (const ekat::Comm& comm, const ekat::ParameterList& params)
   EKAT_REQUIRE_MSG(surf_ref == "sealevel" or surf_ref == "surface",
       "Error! Invalid surface reference for FieldAtHeight.\n"
       " -        field name: " + m_field_name + "\n"
-      " - surface reference: " + surf_ref + "\n");
-  m_surf_ref = (surf_ref == "sealevel") ? "z" : "geopotential";
+      " - surface reference: " + surf_ref + "\n"
+      " -     valid options: sealevel, surface\n");
+  m_z_name = (surf_ref == "sealevel") ? "z" : "geopotential";
   const auto& location = m_params.get<std::string>("vertical_location");
   auto chars_start = location.find_first_not_of("0123456789.");
   EKAT_REQUIRE_MSG (chars_start!=0 && chars_start!=std::string::npos,
@@ -68,8 +69,8 @@ set_grids (const std::shared_ptr<const GridsManager> grids_manager)
   add_field<Required>(m_field_name,gname);
 
   // We don't know yet which one we need
-  add_field<Required>(m_surf_ref+"_mid",gname);
-  add_field<Required>(m_surf_ref+"_int",gname);
+  add_field<Required>(m_z_name+"_mid",gname);
+  add_field<Required>(m_z_name+"_int",gname);
 }
 
 void FieldAtHeight::
@@ -96,7 +97,7 @@ initialize_impl (const RunType /*run_type*/)
       " - field layout: " + to_string(layout) + "\n");
 
   // Figure out the z value
-  m_z_name = tag==LEV ? m_surf_ref+"_mid" : m_surf_ref+"_int";
+  m_z_suffix = tag==LEV ? "_mid" : "_int";
 
   // All good, create the diag output
   FieldIdentifier d_fid (m_diag_name,layout.strip_dim(tag),fid.get_units(),fid.get_grid_name());
@@ -117,7 +118,7 @@ initialize_impl (const RunType /*run_type*/)
 // =========================================================================================
 void FieldAtHeight::compute_diagnostic_impl()
 {
-  const auto z_view = get_field_in(m_z_name).get_view<const Real**>();
+  const auto z_view = get_field_in(m_z_name + m_z_suffix).get_view<const Real**>();
   const Field& f = get_field_in(m_field_name);
   const auto& fl = f.get_header().get_identifier().get_layout();
 
--- a/components/eamxx/src/diagnostics/field_at_height.cpp
+++ b/components/eamxx/src/diagnostics/field_at_height.cpp
@@ -38,6 +38,12 @@ FieldAtHeight (const ekat::Comm& comm, const ekat::ParameterList& params)
  : AtmosphereDiagnostic(comm,params)
 {
   m_field_name = m_params.get<std::string>("field_name");
+  auto surf_ref = m_params.get<std::string>("surface_reference");
+  EKAT_REQUIRE_MSG(surf_ref == "sealevel" or surf_ref == "surface",
+      "Error! Invalid surface reference for FieldAtHeight.\n"
+      " -        field name: " + m_field_name + "\n"
+      " - surface reference: " + surf_ref + "\n");
+  m_surf_ref = (surf_ref == "sealevel") ? "z" : "geopotential";
   const auto& location = m_params.get<std::string>("vertical_location");
   auto chars_start = location.find_first_not_of("0123456789.");
   EKAT_REQUIRE_MSG (chars_start!=0 && chars_start!=std::string::npos,
@@ -52,7 +58,7 @@ FieldAtHeight (const ekat::Comm& comm, const ekat::ParameterList& params)
       "Error! Invalid string for height value for FieldAtHeight.\n"
       " - input string   : " + location + "\n"
       " - expected format: Nm, with N integer\n");
-  m_diag_name = m_field_name + "_at_" + m_params.get<std::string>("vertical_location");
+  m_diag_name = m_field_name + "_at_" + m_params.get<std::string>("vertical_location") + "_above_" + surf_ref;
 }
 
 void FieldAtHeight::
@@ -62,8 +68,8 @@ set_grids (const std::shared_ptr<const GridsManager> grids_manager)
   add_field<Required>(m_field_name,gname);
 
   // We don't know yet which one we need
-  add_field<Required>("z_mid",gname);
-  add_field<Required>("z_int",gname);
+  add_field<Required>(m_surf_ref+"_mid",gname);
+  add_field<Required>(m_surf_ref+"_int",gname);
 }
 
 void FieldAtHeight::
@@ -90,7 +96,7 @@ initialize_impl (const RunType /*run_type*/)
       " - field layout: " + to_string(layout) + "\n");
 
   // Figure out the z value
-  m_z_name = tag==LEV ? "z_mid" : "z_int";
+  m_z_name = tag==LEV ? m_surf_ref+"_mid" : m_surf_ref+"_int";
 
   // All good, create the diag output
   FieldIdentifier d_fid (m_diag_name,layout.strip_dim(tag),fid.get_units(),fid.get_grid_name());
--- a/components/eamxx/src/diagnostics/tests/field_at_height_tests.cpp
+++ b/components/eamxx/src/diagnostics/tests/field_at_height_tests.cpp
@@ -22,7 +22,7 @@ TEST_CASE("field_at_height")
   // Get an MPI comm group for test
   ekat::Comm comm(MPI_COMM_WORLD);
 
-  constexpr int nruns = 10;
+  constexpr int nruns = 100;
 
   util::TimeStamp t0 ({2022,1,1},{0,0,0});
 
@@ -99,9 +99,8 @@ TEST_CASE("field_at_height")
   using RPDF = std::uniform_real_distribution<Real>;
 
   IPDF pdf_fields (0,1000);
-  RPDF pdf_m  (0,10);
+  RPDF pdf_m  (1,10);
   RPDF pdf_y0 (0,5);
-  IPDF pdf_levs (1,nlevs-1);
 
   // Lambda to create and run a diag, and return output
   auto run_diag = [&](const Field& f, const Field& z,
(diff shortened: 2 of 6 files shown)


AaronDonahue:
@bartgol and @brhillman , this still needs a commit to test for the case where `geopotential_mid/int` is used instead of `z_mid/int`.  But I wanted to submit the PR so I could ask a design question and get feedback while I'm working on the PR.

As the PR description states, users now have 3 options to get a Field at Height;
```
  X_at_Nm
  X_at_Nm_above_sealevel
  X_at_Nm_above_surface
```

Two question:
1. Should we support that first case, which currently just defaults to `_above_sealevel`?
2. I've set it up so that now the variable name in the file will always have the `_above_Y` in it, even if the `X_at_Nm` option is used.  I.e. all var names in file will either be `X_at_Nm_above_sealevel` or `X_at_Nm_above_surface`.  Will this mess up post-processing?

brhillman:
Thanks for putting this together Aaron, I think this is great capability to have and really nice to make things explicit. Responding to your questions:

> 1. Should we support that first case, which currently just defaults to `_above_sealevel`?  My concern is that case 1 is a bit lazy and ambiguous.  Although I do currently force the actual output name to explicitly state "above sealevel", see the next question.
> 2. I've set it up so that now the variable name in the file will always have the `_above_Y` in it, even if the `X_at_Nm` option is used.  I.e. all var names in file will either be `X_at_Nm_above_sealevel` or `X_at_Nm_above_surface`.  Will this mess up post-processing?

Regarding 1), the only reason to support the first case is backward compatibility. If we remove that option, then old output yaml lists will break. I think I'm personally comfortable with that, and think we should probably take the cleanest/simplest/easiest to understand approach. If that means only supporting the two options, then I think I'm fine with that.

Regarding 2), I think this is further justification for dropping the ambiguous option. It seems a little weird to give the user something that isn't exactly what they asked for. I.e., I asked for `windspeed_at_100m`, but I got `windspeed_at_100m_above_sea_level`. I do not think this breaks anything though, because I do not think we really have anything that automatically parses the output yaml files to then do any kind of meaningful postprocessing with the outputs themselves and would get confused, but maybe there are things that happen at runtime. Luca would need to answer that. In general though, I guess we are "fortunate" in that we do not have established post-processing toolchains that this would break, so we can kind of do whatever we want at this point.

AaronDonahue:
@brhillman , thanks for responding.  I am in favor of explicit and dropping the ambiguous option.  It would be a part of this PR to to fix all of the YAML files we already have to state "above_sealevel" and I can update the `scream-docs` repo to reflect the new standard.

components/eamxx/src/diagnostics/field_at_height.cpp
```diff
@@ -38,6 +38,12 @@ FieldAtHeight (const ekat::Comm& comm, const ekat::ParameterList& params)
  : AtmosphereDiagnostic(comm,params)
 {
   m_field_name = m_params.get<std::string>("field_name");
+  auto surf_ref = m_params.get<std::string>("surface_reference");
+  EKAT_REQUIRE_MSG(surf_ref == "sealevel" or surf_ref == "surface",
+      "Error! Invalid surface reference for FieldAtHeight.\n"
```
bartgol, line 43: I would also point out to the user what the valid options are. E.g.,

```suggestion
  auto surf_ref = m_params.get<std::string>("surface_reference");
    EKAT_REQUIRE_MSG(surf_ref == "sealevel" or surf_ref == "surface",
      "Error! Invalid surface reference for FieldAtHeight.\n"
     " -        field name: " + m_field_name + "\n"
     " - surface reference: " + surf_ref + "\n"
     " - valid options: sealevel, surface\n");
```

components/eamxx/src/diagnostics/field_at_height.cpp
```diff
@@ -38,6 +38,12 @@ FieldAtHeight (const ekat::Comm& comm, const ekat::ParameterList& params)
...
+      " -        field name: " + m_field_name + "\n"
+      " - surface reference: " + surf_ref + "\n");
+  m_surf_ref = (surf_ref == "sealevel") ? "z" : "geopotential";
```
bartgol, line 46: This is minor, but I think you are abusing the term "surf_ref" here. Perhaps you could use `m_vert_coord_name`, since you are storing the coord name, rather than the surface reference name...

bartgol: Although, that may seem confusingly similar to `m_z_name`. Perhaps you should store "sealevel"/"surface" in `m_surf_name`, and do a switch on it when you need it, like

```
m_z_name = (m_surf_ref=="sealevel" ? "z" : "geopotential") + (tag==LEV" ? "_mid" : "_int");
```

bartgol: Another possibility: do away with `m_surf_ref`, and instead:
- set `m_z_name` to either `z` or `geopotential`
- store `m_z_suffix` to be either `_mid` or `_int`.

Then you can do stuff like
```
add_field<Required>(m_z_name+m_z_suffix,gname);
```
or
```
const auto z_view = get_field_in(m_z_name+m_z_suffix).get_view<const Real**>();
```

components/eamxx/src/diagnostics/field_at_height.cpp
```diff
@@ -62,8 +68,8 @@ set_grids (const std::shared_ptr<const GridsManager> grids_manager)
   add_field<Required>(m_field_name,gname);
 
   // We don't know yet which one we need
-  add_field<Required>("z_mid",gname);
-  add_field<Required>("z_int",gname);
+  add_field<Required>(m_surf_ref+"_mid",gname);
```
bartgol, line 71: I don't think we need to add both mid and int as deps, right?

components/eamxx/src/diagnostics/tests/field_at_height_tests.cpp
```diff
@@ -128,22 +148,22 @@ TEST_CASE("field_at_height")
     print(" -> Testing with z_tgt coinciding with a z level\n");
     {
       print("    -> scalar midpoint field...............\n");
-      auto d = run_diag (s_mid,z_mid,loc);
+      auto d = run_diag (s_mid,z_mid,loc,"sealevel");
```
bartgol, line 151: Looks like you only test `"sealevel"`. Can we add some tests for the `"surface"` case?

components/eamxx/src/diagnostics/tests/field_at_height_tests.cpp
```diff
@@ -128,22 +148,22 @@ TEST_CASE("field_at_height")
     print(" -> Testing with z_tgt coinciding with a z level\n");
```
bartgol, line 148: Can you add a test like
```suggestion
REQUIRE_THROWS (run_diag(s_mid,z_mid,"foobar"));
```
to verify the check on the surf_ref type works correctly?

bartgol (commented):
I have some suggestions. I agree with Ben that we should take the chance to prune the `f_at_Xm` syntax, since there are no postprocessing tools yet using these fields. If we waited a bit, giving some grace period, that may no longer be the case...

components/eamxx/src/diagnostics/field_at_height.cpp
AaronDonahue: I think we do, since we don't what layout the source field has until we get to initialization.  Unless there has been an update in how we can access fields in the field manager since this diagnostic was written...

bartgol: Ah, yes. I didn't realize that the request for the input field was created just with field/grid name, without layout.

components/eamxx/src/diagnostics/tests/field_at_height_tests.cpp
```diff
@@ -120,106 +143,181 @@ TEST_CASE("field_at_height")
...
+Field get_ref_field(const int lev, const Field& src_data, const std::string& surf_ref)
+{
+  // A single level implies this is a test where the height matches a single level.
+  if (surf_ref == "sealevel") {
+    auto tgt = src_data.subfield(1,lev).clone();
+    const auto& tgt_view = tgt.get_view<Real*,Host>();
+    const auto dims = src_data.get_header().get_identifier().get_layout().dims();
+    const auto src_view = src_data.get_view<const Real**,Host>();
+    for (int i = 0; i < dims[0]; ++i) {
+      int ilev = lev + i >= dims[1] ? dims[1]-1 : lev + i;
```
bartgol, line 288: I'm confused by this line. The index `i` is over columns, while here you use it to choose at what level to subview the input field?

components/eamxx/src/diagnostics/tests/field_at_height_tests.cpp
```diff
@@ -120,106 +143,181 @@ TEST_CASE("field_at_height")
...
+    auto tgtp = src_data.subfield(1,levp).clone();
+    auto tgtm = src_data.subfield(1,levm).clone();
+
+    const auto& tgtp_view = tgtp.get_view<Real*,Host>();
+    const auto& tgtm_view = tgtm.get_view<Real*,Host>();
+
+    const auto dims = src_data.get_header().get_identifier().get_layout().dims();
+    const auto src_view = src_data.get_view<const Real**,Host>();
+    for (int i = 0; i < dims[0]; ++i) {
+      int ilevp = levp + i >= dims[1] ? dims[1]-1 : levp + i;
```
bartgol, line 310: Same here. Why are column indices mixed up with the level indices?

components/eamxx/src/diagnostics/tests/field_at_height_tests.cpp
```diff
@@ -98,7 +110,18 @@ TEST_CASE("field_at_height")
     const auto& dims = f.get_header().get_identifier().get_layout().dims();
     for (int i=0; i<dims[0]; ++i) {
       for (int j=0; j<dims[1]; ++j) {
-        v(i,j) = nlevs-j;
+        v(i,j) = dims[1]-j+i;
```
bartgol, line 113: Ah, now I understand why you mix j and i in `get_ref_field`. The surf elevation is basically `z_surf(i) = 1 + i`.

I would make the test clearer (or else it may be a bit of a headache for those who will maintain it down the road):
1. create the z_surf field.
2. create z_int as z_surf + ilev
3. create z_mid as z_surf + ilev + 0.5 (the half is to make it consistent with z_int, for clarity).
4. pass z (mid or int) to `get_ref_field`, and do something like
```
  if (z_lev < z_surf) {
    ...
```

bartgol:
@AaronDonahue FYI, this is the output of the test that fails on weaver:
```
17:11:26 -------------------------------------------------------------------------------
17:11:26 /home/e3sm-jenkins/weaver/workspace/SCREAM_PullRequest_Autotester_Weaver/5267/scream/components/eamxx/src/diagnostics/tests/field_at_height_tests.cpp:15
17:11:26 ...............................................................................
17:11:26 
17:11:26 /home/e3sm-jenkins/weaver/workspace/SCREAM_PullRequest_Autotester_Weaver/5267/scream/components/eamxx/src/diagnostics/tests/field_at_height_tests.cpp:226: FAILED:
17:11:26   REQUIRE( views_are_equal(d,tgt,&comm) )
17:11:26 with expansion:
17:11:26   false
```
Not a huge help, but at least you know which check if failing...

components/eamxx/src/diagnostics/tests/field_at_height_tests.cpp
AaronDonahue: I can do that.  I think having a +0.5 for `z_mid` will break the "simplicity" of the test just taking a subfield of the data at the z_tgt level.  But I think that is okay, the test might be more robust if we actually triggered random locations (rather than integers) for the z_tgt and thus for the interpolation.

bartgol: yeah, that's fine. You can keep `z_mid = z_surf + ilev`. For the sake of this testing, we don't need to have mid to be in the middle of interfaces...

AaronDonahue:
> @AaronDonahue FYI, this is the output of the test that fails on weaver:
> 
> ```
> 17:11:26 -------------------------------------------------------------------------------
> 17:11:26 /home/e3sm-jenkins/weaver/workspace/SCREAM_PullRequest_Autotester_Weaver/5267/scream/components/eamxx/src/diagnostics/tests/field_at_height_tests.cpp:15
> 17:11:26 ...............................................................................
> 17:11:26 
> 17:11:26 /home/e3sm-jenkins/weaver/workspace/SCREAM_PullRequest_Autotester_Weaver/5267/scream/components/eamxx/src/diagnostics/tests/field_at_height_tests.cpp:226: FAILED:
> 17:11:26   REQUIRE( views_are_equal(d,tgt,&comm) )
> 17:11:26 with expansion:
> 17:11:26   false
> ```
> 
> Not a huge help, but at least you know which check if failing...

Thanks @bartgol, I think I may have been missing a sync_to_dev.  Hopefully the latest version has remedied all of these.

components/eamxx/src/diagnostics/tests/field_at_height_tests.cpp
```diff
@@ -92,135 +123,200 @@ TEST_CASE("field_at_height")
...
-        REQUIRE (views_are_equal(di,tgt,&comm));
+      // Randomize fields using f_z_src function defined above:
+      auto slope = pdf_m(engine); 
+      auto inter = pdf_y0(engine);
+      f_z_src(inter, slope, mid_src, s_mid);
+      f_z_src(inter, slope, mid_src, v_mid);
+      f_z_src(inter, slope, int_src, s_int);
+      f_z_src(inter, slope, int_src, v_int);
+
+      // Set target z-slice for testing to a random value.
```
bartgol, line 188: This only tests the case where `z_tgt` matches exeactly one of the levels though. Since you compute src and tgt via analytic function, you can probably switch `pdf_levs` to be a RPDF, and test also intermediate cases.

AaronDonahue: oh yes, good point.  Thanks.  I'll make the change.

AaronDonahue: @bartgol , actually it still needs to be an INT not a REAL because FieldAtHeight assumes an integer number of meters above the reference.  I.e. doesn't support decimals in the height.  So instead I will bump up the `z_top` and made the PDF go from 1 to z_top.

bartgol: Ok, but we need to test the case where `z_tgt` does not coincide with any of the values in `z_mid`/`z_int`. Perhaps you could make them have a stride >=2, so that it is possible that the IPDF generates an integer that is not a value of z?

components/eamxx/src/diagnostics/tests/field_at_height_tests.cpp
```diff
@@ -133,29 +132,39 @@ TEST_CASE("field_at_height")
...
+  int         min_col_thickness = z_top;
+  int         max_surf = 0;
   for (int ii=0; ii<ncols; ++ii) {
-    zsurf_v(ii) = ii;
-    const float dz = (z_top - zsurf_v(ii))/nlevs;
+    zsurf_v(ii) = ii*surf_slope;
+    max_surf = zsurf_v(ii) > max_surf ? zsurf_v(ii) : max_surf;
+    const Real col_thickness = z_top - zsurf_v(ii);
+    min_col_thickness = min_col_thickness < col_thickness ? col_thickness : min_col_thickness;
+    const Real dz = (z_top - zsurf_v(ii))/nlevs;
```
bartgol, line 149: I believe your approach works, since `ztop-zsurf` is larger than `nlevs`. However, I would like to make sure this won't change in the future (e.g., someone changes this test 1y from now, making ztop=10).

Can you add something like this?

```suggestion
REQUIRE (dz>1);
```

If I understand correctly, given your choices, right now we have
- slope = 100/ncols=33.3
- max(zsurf) = (ncols-1)*slope=66.6
- min(dz) = (1000-max(zsurf))/nlevs = 933.3/10 = 93.3

So we're good. But better make sure this stays like this.

bartgol: Since committing now would abort the AT run, you can do this change later, and merge without another round of testing. I'm confident that the data we have would make the `REQUIRE(dz>1)` check pass.

bartgol (approved):

components/eamxx/src/diagnostics/tests/field_at_height_tests.cpp
bartgol: You could maybe run the test locally on your machine, to make sure our math is correct, and dz>1 always (meaning that check will pass, and we're not merging something that will cause fails)

PR #2654, opened 2024-01-05
Gets Frontier-GPU build working again

jgfouca:
Note: I did not look at getting it to run. That will be a follow-on PR.

Fix list:
1) e3sm_remove_flags is gone, so the frontier Depends file that was using it had to be modified. I don't think the call was necessary since later compiler options override prior ones.
2) Change fortran flags to `-em -ef` to ensure that mod files do not have all-caps names. The all-caps was throwing off Cmake in a couple places were lowercase mod file names were assumed. Make this change at the compiler macro level so that it affects all machines.
3) Be sure to use fortran as the linker language. This change was on the cess-production branch but not master.
4) Set compilers in the standard way, including MPICXX and use MPICH_CXX to make sure hipcc is used.
5) Set KOKKOS_OPTIONS so that HIP is on.
6) Unset HDF5_ROOT in config_machines, this will disable the `find_package(hdf5)` call. Frontier fails to find HDF5 even though the module is loaded and HDF5_ROOT was being set correctly.
7) Adds new mechanism for hardcoding BLAS_LIBRARIES and set it to the CRAY_LIBSCI libs. This workaround is needed because the CMake native findBlas module does not support Cray as a blas vendor yet.
8) Skip CUDA/HIP compiler checks (much like how language compiler checks are skipped via CMAKE_<LANG>_COMPILER_WORKS=1.
9) csm_share does not use HIP or CUDA, so there's no need to enable those language in its CMakeLists.txt file.

Tested with:
`./create_test SMS.ne4pg2_ne4pg2.F2010-SCREAMv1.frontier-scream-gpu_crayclang-scream --no-run`

Fixes https://github.com/E3SM-Project/E3SM/issues/6140


ambrad:
I recommend getting tests to run in this same PR. That's because I think it's best to spend some time running tests on a currently working GPU system like pm-gpu before merging this. As long as you need to put in that effort, you might as well get tests running on Frontier, too.

It looks like some of these changes could affect upstream E3SM. I'm wondering if a subset of this PR should be done there instead of in this repo so that they get the test coverage of the `next` nightlies. You're the build-system lead, though, so this is just a thought, not a request.

Is 8 necessary or, rather, something you decided would be nice? Ascent, Summit, and pm-gpu are all fine with the CUDA compiler checks.

Am I correct that some or most of these changes are needed for the new build system, while a smaller set of changes are to make the master branch consistent with the cess-production branch?

jgfouca:
@ambrad , 

> It looks like some of these changes could affect upstream E3SM. I'm wondering if a subset of this PR should be done there instead of in this repo so that they get the test coverage of the next nightlies. You're the build-system lead, though, so this is just a thought, not a request.

That's correct. With the imminent unification of the repos, I'm not too worried about it.

> Is 8 necessary or, rather, something you decided would be nice? Ascent, Summit, and pm-gpu are all fine with the CUDA compiler checks.

Just nice.

> Am I correct that some or most of these changes are needed for the new build system, while a smaller set of changes are to make the master branch consistent with the cess-production branch?

1, 2, 4, 5, 6, and 7 were all due to upstream build system changes. 3 was the only item where cess branch stuff needed to be moved over to this branch. 8 and 9 are "nice to have" changes.

cime_config/machines/cmake_macros/crayclang-scream_frontier-scream-gpu.cmake
```diff
@@ -1,6 +1,6 @@
 set(MPICC "mpicc")
-set(MPICXX "hipcc")
-set(MPIFC "ftn")
+set(MPICXX "mpicxx") # Needs MPICH_CXX to use hipcc
```
sarats, line 2: Need to make sure mpicxx is picking up the right compiler (hipcc in this case). To avoid this, we previously had hipcc and then added MPI include and lib flags appropriately.

cime_config/machines/config_machines.xml
```diff
@@ -1269,23 +1269,25 @@
...
       <env name="NETCDF_PATH">$ENV{NETCDF_DIR}</env>
       <env name="PNETCDF_PATH">$ENV{PNETCDF_DIR}</env>
+      <env name="HDF5_ROOT"/> <!-- frontier cmake cannot find_package(HDF5) for some reason, so disable it -->
       <env name="MPICH_GPU_SUPPORT_ENABLED">1</env>
       <env name="MPICH_VERSION_DISPLAY">1</env>
       <env name="MPICH_OFI_CXI_COUNTER_REPORT">2</env>
+      <env name="MPICH_CXX">$SHELL{which hipcc}</env>
       <env name="LD_LIBRARY_PATH">$ENV{CRAY_LD_LIBRARY_PATH}:$ENV{LD_LIBRARY_PATH}</env>
-
+      <env name="BLAS_LIBRARIES">$ENV{CRAY_LIBSCI_PREFIX_DIR}/lib/libsci_cray.a</env> <!-- CMake does not currently support Cray as a blas vendor -->
```
sarats, line 1290: CMake tends to be schizophrenic with LibSci, used to know about LibSci.

Anyway, you don't want to link a specific variant as compiler flags (with OpenMP, accelerator support etc.) dictate which BLAS lib gets linked. 

```
$ ls /opt/cray/pe/libsci/22.12.1.1/CRAY/9.0/x86_64/lib/
libsci_cray.a         libsci_cray_mpi_mp.so      libsci_cray_mpi.so.5    libsci_cray_mp.so.5.0  pkgconfig
libsci_cray_mp.a      libsci_cray_mpi_mp.so.5    libsci_cray_mpi.so.5.0  libsci_cray.so
libsci_cray_mpi.a     libsci_cray_mpi_mp.so.5.0  libsci_cray_mp.so       libsci_cray.so.5
libsci_cray_mpi_mp.a  libsci_cray_mpi.so         libsci_cray_mp.so.5     libsci_cray.so.5.0
```

cime_config/machines/cmake_macros/crayclang-scream_frontier-scream-gpu.cmake
sarats: To elaborate, hipcc will pick it up from the amd/rocm modules while Cray wrapper (CC) will use the Cray's C++ compiler. Depending on what the path ends up being, mpicxx might pickup the Cray's compiler which lags behind the AMD's LLVM featureset/

oksanaguba:
It seems that libsci/blas issues from scream builds/runs resurfaced. Correct me if i am wrong, iirc this hasn't been resolved. In case it helps, scream for production runs on frontier had to have a mod in land https://github.com/E3SM-Project/scream/commit/24b1a20b76154e90a846a66392fdae3ae4317569 to replace blas calls with raw implementations that Andrew found. Otherwise with threads, no threads, or 1 thread, land model would crash during runs around those blas calls, erratically. Maybe it is safer to keep this change in mmf runs, too?

ambrad:
Oksana is referring to https://github.com/E3SM-Project/scream/commit/24b1a20b76154e90a846a66392fdae3ae4317569, where we call the LAPACK reference version of the band-diagonal solver instead of Cray's (which seems to use OpenBLAS internally (?) b/c linking to OpenBLAS didn't fix the problem). This commit works around a run-time problem.

It's true that this problem will probably still cause issues for any runs that don't have this workaround. However, my understanding is that the specific issue Jim was having was at link time. My understanding is also that this branch is currently failing on the tcmalloc issue at run time, so it isn't yet supporting a run that's long enough to hit the band-diag solver problem.

cime_config/machines/config_machines.xml
```diff
@@ -1269,23 +1269,25 @@
...
-    
+
     <RUNDIR>$CIME_OUTPUT_ROOT/$CASE/run</RUNDIR>
     <EXEROOT>$CIME_OUTPUT_ROOT/$CASE/bld</EXEROOT>
     <TEST_TPUT_TOLERANCE>0.1</TEST_TPUT_TOLERANCE>
     <MAX_GB_OLD_TEST_DATA>0</MAX_GB_OLD_TEST_DATA>
     <environment_variables>
       <env name="NETCDF_PATH">$ENV{NETCDF_DIR}</env>
       <env name="PNETCDF_PATH">$ENV{PNETCDF_DIR}</env>
+      <env name="HDF5_ROOT"/> <!-- frontier cmake cannot find_package(HDF5) for some reason, so disable it -->
```
oksanaguba, line 1284: How does *xml work with cmake? Naively, I would think setting HDF5_ROOT here is the way to point the build to hdf5 in case find_package() fails, but it is actually not the case?

cime_config/machines/config_machines.xml
```diff
@@ -1269,23 +1269,25 @@
...
       <env name="MPICH_GPU_SUPPORT_ENABLED">1</env>
       <env name="MPICH_VERSION_DISPLAY">1</env>
       <env name="MPICH_OFI_CXI_COUNTER_REPORT">2</env>
+      <env name="MPICH_CXX">$SHELL{which hipcc}</env>
```
oksanaguba, line 1288: why is mpich connected to the version of hipcc?

cime_config/machines/cmake_macros/crayclang-scream_frontier-scream-gpu.cmake
jgfouca: The env var `MPICH_CXX` should take care of this. See the changes in the config_machines.xml section.

components/CMakeLists.txt
```diff
@@ -82,9 +82,11 @@ endif()
 
 project(E3SM C CXX Fortran)
 
-if(USE_CUDA)
+if (USE_CUDA)
+  set(CMAKE_CUDA_COMPILER_FORCED True)
   enable_language(CUDA)
-elseif(USE_HIP)
+elseif (USE_HIP)
+  set(CMAKE_HIP_COMPILER_FORCED True)
```
oksanaguba, line 89: what does CMAKE_HIP_COMPILER_FORCED do? adds some special hipcc flags?

cime_config/machines/config_machines.xml
jgfouca: Note that the text section for this element is blank, this tells CIME to UNSET this env var. When HDF5_ROOT is not set, our cmake will not try to find_package it.

cime_config/machines/config_machines.xml
jgfouca: Same reason we use OMPI_CXX=nvcc_wrapper on CUDA machines. We want to use hipcc as the compiler underlying mpicxx.

components/CMakeLists.txt
jgfouca: > Skip CUDA/HIP compiler checks (much like how language compiler checks are skipped via CMAKE__COMPILER_WORKS=1.

cime_config/machines/config_machines.xml
oksanaguba: i re-read the PR and i see there was already a comment on MPICH_CXX, but i am still confused how exactly this variable is processed, if you can elaborate please? thanks.

jgfouca:
@ambrad ,

> My understanding is also that this branch is currently failing on the tcmalloc issue at run time, so it isn't yet supporting a run that's long enough to hit the band-diag solver problem.

Correct. Current error:

```
3: /home/jenkins/crayftn/craylibs/google-perftools/src/tcmalloc.cc:651] Attempt to realloc invalid pointer: 0x204beb20
7: /home/jenkins/crayftn/craylibs/google-perftools/src/tcmalloc.cc:651] Attempt to realloc invalid pointer: 0x204beb20
5: /home/jenkins/crayftn/craylibs/google-perftools/src/tcmalloc.cc:651] Attempt to realloc invalid pointer: 0x204beb20
4: /home/jenkins/crayftn/craylibs/google-perftools/src/tcmalloc.cc:651] Attempt to realloc invalid pointer: 0x204beb20
6: /home/jenkins/crayftn/craylibs/google-perftools/src/tcmalloc.cc:651] Attempt to realloc invalid pointer: 0x204beb20
0: /home/jenkins/crayftn/craylibs/google-perftools/src/tcmalloc.cc:651] Attempt to realloc invalid pointer: 0x204beb20
1: /home/jenkins/crayftn/craylibs/google-perftools/src/tcmalloc.cc:651] Attempt to realloc invalid pointer: 0x204beb20
2: /home/jenkins/crayftn/craylibs/google-perftools/src/tcmalloc.cc:651] Attempt to realloc invalid pointer: 0x204beb20
srun: error: frontier00127: tasks 0-2,4-7: Aborted
```

I get the same error on debug builds. I diffed against the key cime_config files from the cess_production branch, config_machines.xml, Depends.cray.., cmake_macros and it was all basically the same. I also used my build differ tool and didn't see anything interesting, though there was a lot to sift through so it's possible I missed something.

One thing that jumps out at me from the error is this path: `/home/jenkins/crayftn/craylibs/google-perftools/src/tcmalloc.cc` . I don't even have read access there.

ambrad:
Re: tcmalloc, doing something like https://github.com/E3SM-Project/scream/commit/218ccf228d061b28427aeebbf499ed4e148efca5 didn't help?

jgfouca:
@ambrad , I tried hacking that tcmalloc into my link line and it didn't help. I will try making those changes at the macro level.

ambrad:
>  into my link line and it didn't help. I will try making those changes at the macro level.

No, if it didn't work in a hacked link.txt, then there must be some other issue.

One thing that struck me in the --fmad=false/true nondeterminism bug hunt on pm-gpu and Ascent is that the build lines in the old and new build systems looked essentially the same and yet the behavior was different. Is it possible a CIME python wrapper has changed in an unintended way?

ambrad:
Noting Sarat's comment above and also more generally, have you compared the `module list` for the cess-production and current branches?

jgfouca:
@ambrad ,

> One thing that struck me in the --fmad=false/true nondeterminism bug hunt on pm-gpu and Ascent is that the build lines in the old and new build systems looked essentially the same and yet the behavior was different. Is it possible a CIME python wrapper has changed in an unintended way?

That's frustrating. I have not seen this myself (until now?). The python build wrapper, `e3sm_compile_wrap.py` (currently just used to add timings to the build logs), hasn't changed in years, so I don't think that could be the problem.

jgfouca:
@ambrad ,

> Noting Sarat's comment above and also more generally, have you compared the module list for the cess-production and current branches?

Yes. I copied the module/environment XML sections for frontier-scream-gpu for the cess branch and my branch into two files and did a diff. There were identical except for the changes I made in this PR (unsetting HDF5_ROOT and setting MPICH_CXX).

ambrad:
> did a diff.

Even so, it might be good to source the two .env_machine...xml files (or look at the env txt file that I think exists in the case directory) to be sure the environments are identical.

jgfouca:
https://github.com/E3SM-Project/scream/commit/218ccf228d061b28427aeebbf499ed4e148efca5 leads to a new error:

```
0: MPICH ERROR [Rank 0] [job id 1563975.0] [Mon Jan  8 13:50:23 2024] [frontier07425] - Abort(-1) (rank 0 in comm 0): MPIDI_CRAY_init: GPU_SUPPORT_ENABLED is requested, but GTL library is not linked
0:  (Other MPI error)
0: 
0: aborting job:
0: MPIDI_CRAY_init: GPU_SUPPORT_ENABLED is requested, but GTL library is not linked
0: 
srun: error: frontier07425: task 0: Exited with exit code 255
srun: Terminating StepId=1563975.0
srun: error: frontier07425: tasks 1-3,5-7: Segmentation fault
0: slurmstepd: error: *** STEP 1563975.0 ON frontier07425 CANCELLED AT 2024-01-08T13:50:24 ***
srun: error: frontier07425: task 4: Segmentation fault (core dumped)
```

Not sure if this means I made it past the tcmalloc error. This looks like it should be easy to fix, I just need to figure out what additional library to link.

ambrad:
Jim, I think you're right: you're past the tcmalloc error. The GTL issue looks like something basic like not having craype-accel-amd-gfx90a loaded.

jgfouca:
Based on googling, this error is because CRAY_ACCEL_TARGET is not set in the env, but i see it in mine:
```
CRAY_ACCEL_TARGET=amd_gfx90a
```

I'm going to try setting it explicitly via compiler argument `-target-accel=amd_gfx90a`

oksanaguba:
what i find confusing is that we already have a frontier config in cess branch that works, why are there these issues with target and env and stuff?

jgfouca:
@oksanaguba , I haven't been able to identify what the difference is between the cess branch and this branch that is causing these extra problems.

oksanaguba:
From googling, here is another user's reference to tcmalloc and jenkins (so "jenkins" comes from the system) in the path name, they changed linker flags to get around it. i understand you already bypassed this issue in another way.

https://gitlab.com/QEF/q-e/-/merge_requests/2080

jgfouca:
@oksanaguba , thanks, that appears to be helpful. Just the presence of `-hsystem_alloc` on the link line gets from the tmalloc error to the `no GPU_SUPPORT_ENABLED` error. I think this is a step forward.

jgfouca:
OK, minor breakthrough. If I take the e3sm.exe link line from the cess2 branch and apply it to my branch, making the necessary tweaks for kokkos, path changes etc, it PASSes. I just need to figure out what link flag specifically makes the difference.

jgfouca:
Weird, simply using `ftn` instead of `mpif90` (and explicitly linking `-lmpich`) fixes everything.

ambrad:
> Weird, simply using ftn instead of mpif90 (and explicitly linking -lmpich) fixes everything.

`ftn` is the Cray compiler wrapper, so it actually makes sense that it works. Typically, on Cray systems, we have to use their wrapper to successfully link because the wrapper brings in a bunch of extra stuff.

jgfouca:
OK, this branch builds and runs PASS now. I still need to investigate the blas issue that @sarats brought up.

cime_config/machines/cmake_macros/crayclang-scream_frontier-scream-gpu.cmake
```diff
@@ -1,6 +1,6 @@
...
+set(MPIFC "ftn") # Linker needs to be plain ftn, not mpif90
```
ambrad, line 3: Here I would remove the "plain ftn" part because `ftn` is a Cray wrapper, not the bare Fortran compiler.

jgfouca: @ambrad , good call. I have improved the text of the comment.

jgfouca:
@sarats , you were right. Skipping the cmake blas stuff entirely seems to work fine. I think this branch is ready for final review and merge.

ambrad (dismissed):
Looks good. I removed AT:AUTOMERGE so my approval wouldn't trigger the merge, in case Oksana or Sarat have final comments.

ambrad:
We should note that this PR gets tests running, but the ELM diag-solve BLAS issue, if it still exists, will show up randomly. The only fix or workaround that we've thought of that works is to use the LAPACK reference impl instead of the system blas library. But that can't go into the upstream code.

oksanaguba (dismissed):

cime_config/machines/cmake_macros/crayclang-scream_frontier-scream-gpu.cmake
sarats: >       <env name="MPICH_CXX">$SHELL{which hipcc}</env>

Clarify the purpose/usage of MPICXX and MPICH_CXX. Trying to understand why we need both of them and if there is way to improve clarity.

bartgol: `MPICXX` is a CMake var, used by E3SM CMake build system to specify the compiler. We could do away with this, and just do `set(CMAKE_CXX_COMPILER mpicxx CACHE STRING "")`, or something like that (maybe with FORCE), but I think for bwd reasons we need to keep these vars around for CIME to pick up the compiler (that's a guess). `MPICH_CXX`, otoh, is an env var that it used by `mpicxx` to decide which backend CXX compiler to use.

jgfouca: Yep, same concept as OMPI_CXX=nvcc for openmpi on cuda machines.

PeterCaldwell:
Awesome, great to have this in! Thanks Jim!

PR #2655, opened 2024-01-08
expose p3 parameters

oksanaguba:
Expose some of P3 parameters as namelist variables, put them in one struct.

Here are vars exposed:
P3   p3_autoconversion_prefactor = 1350.000000
P3   p3_mu_r_constant = 1.000000
P3   p3_spa_to_nc = 1.000000
P3   p3_k_accretion = 67.000000
P3   p3_eci = 0.500000
P3   p3_eri = 1.000000
P3   p3_rho_rime_min = 50.000000
P3   p3_rho_rime_max = 900.000000
P3   p3_a_imm = 0.650000
P3   p3_dep_nucleation_exponent = 0.304000
P3   p3_ice_sed_knob = 1.000000
P3   p3_d_breakup_cutoff = 0.000280

[nonbfb] for physics_saturation_run_and_cmp for weaver because of inlining line in physics/share/tests/physics_saturation_run_and_cmp.cpp .

### DIFF
--- a/components/eamxx/cime_config/namelist_defaults_scream.xml
+++ b/components/eamxx/cime_config/namelist_defaults_scream.xml
@@ -206,6 +206,7 @@ be lost if SCREAM_HACK_XML is not enabled.
       <p3_pre_autoconversion_factor type="real" doc="P3 pre_autoconversion_factor">1350.0</p3_pre_autoconversion_factor>
       <p3_mu_r_constant type="real" doc="P3 mu_r_constant">1.0</p3_mu_r_constant>
       <p3_spa_to_nc type="real" doc="P3 spa_to_nc">1.0</p3_spa_to_nc>
+      <p3_k_accretion type="real" doc="P3 k_accretion">67.0</p3_k_accretion>
     </p3>
 
     <!-- SHOC macrophysics -->
--- a/components/eamxx/cime_config/namelist_defaults_scream.xml
+++ b/components/eamxx/cime_config/namelist_defaults_scream.xml
@@ -207,6 +207,11 @@ be lost if SCREAM_HACK_XML is not enabled.
       <p3_mu_r_constant type="real" doc="P3 mu_r_constant">1.0</p3_mu_r_constant>
       <p3_spa_to_nc type="real" doc="P3 spa_to_nc">1.0</p3_spa_to_nc>
       <p3_k_accretion type="real" doc="P3 k_accretion">67.0</p3_k_accretion>
+      <p3_eci type="real" doc="P3 eci">0.5</p3_eci>
+      <p3_eri type="real" doc="P3 eri">1.0</p3_eri>
+      <p3_rho_rime_min type="real" doc="P3 rho_rime_min">50.0</p3_rho_rime_min>
+      <p3_rho_rime_max type="real" doc="P3 rho_rime_max">900.0</p3_rho_rime_max>
+      <p3_a_imm type="real" doc="P3 a_imm">0.65</p3_a_imm>
     </p3>
 
     <!-- SHOC macrophysics -->
--- a/components/eamxx/cime_config/namelist_defaults_scream.xml
+++ b/components/eamxx/cime_config/namelist_defaults_scream.xml
@@ -203,7 +203,7 @@ be lost if SCREAM_HACK_XML is not enabled.
         ${DIN_LOC_ROOT}/atm/scream/tables/vn_table_vals.dat8,
         ${DIN_LOC_ROOT}/atm/scream/tables/vm_table_vals.dat8
       </tables>
-      <p3_autoconversion_factor type="real" doc="P3 autoconversion factor">1350.0</p3_autoconversion_factor>
+      <p3_pre_autoconversion_factor type="real" doc="P3 pre_autoconversion_factor">1350.0</p3_pre_autoconversion_factor>
     </p3>
 
     <!-- SHOC macrophysics -->
--- a/components/eamxx/cime_config/namelist_defaults_scream.xml
+++ b/components/eamxx/cime_config/namelist_defaults_scream.xml
@@ -203,15 +203,15 @@ be lost if SCREAM_HACK_XML is not enabled.
         ${DIN_LOC_ROOT}/atm/scream/tables/vn_table_vals.dat8,
         ${DIN_LOC_ROOT}/atm/scream/tables/vm_table_vals.dat8
       </tables>
-      <p3_pre_autoconversion_factor type="real" doc="P3 pre_autoconversion_factor">1350.0</p3_pre_autoconversion_factor>
-      <p3_mu_r_constant type="real" doc="P3 mu_r_constant">1.0</p3_mu_r_constant>
-      <p3_spa_to_nc type="real" doc="P3 spa_to_nc">1.0</p3_spa_to_nc>
-      <p3_k_accretion type="real" doc="P3 k_accretion">67.0</p3_k_accretion>
-      <p3_eci type="real" doc="P3 eci">0.5</p3_eci>
-      <p3_eri type="real" doc="P3 eri">1.0</p3_eri>
-      <p3_rho_rime_min type="real" doc="P3 rho_rime_min">50.0</p3_rho_rime_min>
-      <p3_rho_rime_max type="real" doc="P3 rho_rime_max">900.0</p3_rho_rime_max>
-      <p3_a_imm type="real" doc="P3 a_imm">0.65</p3_a_imm>
+      <p3_pre_autoconversion_factor type="real" doc="P3 pre_autoconversion_factor (scale factor in autoconversion)">1350.0</p3_pre_autoconversion_factor>
+      <p3_mu_r_constant type="real" doc="P3 mu_r_constant (rain “shape parameter” in gamma drop-size distribution)">1.0</p3_mu_r_constant>
+      <p3_spa_to_nc type="real" doc="P3 spa_to_nc (scaling factor for turning CCN into nc in SPA)">1.0</p3_spa_to_nc>
+      <p3_k_accretion type="real" doc="P3 k_accretion (scaling factor on accretion)">67.0</p3_k_accretion>
+      <p3_eci type="real" doc="P3 eci (liquid/ice collision/collection coefficient)">0.5</p3_eci>
+      <p3_eri type="real" doc="P3 eri (ice/rain collision/collection coefficient)">1.0</p3_eri>
+      <p3_rho_rime_min type="real" doc="P3 rho_rime_min (riming density maximum)">50.0</p3_rho_rime_min>
+      <p3_rho_rime_max type="real" doc="P3 rho_rime_max (riming density minimum)">900.0</p3_rho_rime_max>
+      <p3_a_imm type="real" doc="P3 a_imm (drop freezing exponent )">0.65</p3_a_imm>
     </p3>
 
     <!-- SHOC macrophysics -->
--- a/components/eamxx/cime_config/namelist_defaults_scream.xml
+++ b/components/eamxx/cime_config/namelist_defaults_scream.xml
@@ -203,7 +203,7 @@ be lost if SCREAM_HACK_XML is not enabled.
         ${DIN_LOC_ROOT}/atm/scream/tables/vn_table_vals.dat8,
         ${DIN_LOC_ROOT}/atm/scream/tables/vm_table_vals.dat8
       </tables>
-      <p3_pre_autoconversion_factor type="real" doc="P3 pre_autoconversion_factor (scale factor in autoconversion)">1350.0</p3_pre_autoconversion_factor>
+      <p3_autoconversion_prefactor type="real" doc="P3 autoconversion_prefactor (scale factor in autoconversion)">1350.0</p3_autoconversion_prefactor>
       <p3_mu_r_constant type="real" doc="P3 mu_r_constant (rain “shape parameter” in gamma drop-size distribution)">1.0</p3_mu_r_constant>
       <p3_spa_to_nc type="real" doc="P3 spa_to_nc (scaling factor for turning CCN into nc in SPA)">1.0</p3_spa_to_nc>
       <p3_k_accretion type="real" doc="P3 k_accretion (scaling factor on accretion)">67.0</p3_k_accretion>
@@ -212,6 +212,9 @@ be lost if SCREAM_HACK_XML is not enabled.
       <p3_rho_rime_min type="real" doc="P3 rho_rime_min (riming density maximum)">50.0</p3_rho_rime_min>
       <p3_rho_rime_max type="real" doc="P3 rho_rime_max (riming density minimum)">900.0</p3_rho_rime_max>
       <p3_a_imm type="real" doc="P3 a_imm (drop freezing exponent )">0.65</p3_a_imm>
+      <p3_dep_nucleation_exponent type="real" doc="P3 dep_nucleation_exponent (deposition nucleation)">0.304</p3_dep_nucleation_exponent>
+      <p3_ice_sed_knob type="real" doc="P3 ice_sed_knob (ice fall speed)">1.0</p3_ice_sed_knob>
+      <p3_d_breakup_cutoff type="real" doc="P3 d_breakup_cutoff (rain self collection and breakup)">0.00028</p3_d_breakup_cutoff>
     </p3>
 
     <!-- SHOC macrophysics -->
--- a/components/eamxx/cime_config/namelist_defaults_scream.xml
+++ b/components/eamxx/cime_config/namelist_defaults_scream.xml
@@ -205,6 +205,7 @@ be lost if SCREAM_HACK_XML is not enabled.
       </tables>
       <p3_pre_autoconversion_factor type="real" doc="P3 pre_autoconversion_factor">1350.0</p3_pre_autoconversion_factor>
       <p3_mu_r_constant type="real" doc="P3 mu_r_constant">1.0</p3_mu_r_constant>
+      <p3_spa_to_nc type="real" doc="P3 spa_to_nc">1.0</p3_spa_to_nc>
     </p3>
 
     <!-- SHOC macrophysics -->
--- a/components/eamxx/cime_config/namelist_defaults_scream.xml
+++ b/components/eamxx/cime_config/namelist_defaults_scream.xml
@@ -203,6 +203,7 @@ be lost if SCREAM_HACK_XML is not enabled.
         ${DIN_LOC_ROOT}/atm/scream/tables/vn_table_vals.dat8,
         ${DIN_LOC_ROOT}/atm/scream/tables/vm_table_vals.dat8
       </tables>
+      <p3_autoconversion_factor type="real" doc="P3 autoconversion factor">1350.0</p3_autoconversion_factor>
     </p3>
 
     <!-- SHOC macrophysics -->
--- a/components/eamxx/cime_config/namelist_defaults_scream.xml
+++ b/components/eamxx/cime_config/namelist_defaults_scream.xml
@@ -204,6 +204,7 @@ be lost if SCREAM_HACK_XML is not enabled.
         ${DIN_LOC_ROOT}/atm/scream/tables/vm_table_vals.dat8
       </tables>
       <p3_pre_autoconversion_factor type="real" doc="P3 pre_autoconversion_factor">1350.0</p3_pre_autoconversion_factor>
+      <p3_mu_r_constant type="real" doc="P3 mu_r_constant">1.0</p3_mu_r_constant>
     </p3>
 
     <!-- SHOC macrophysics -->
--- a/components/eamxx/src/physics/p3/eamxx_p3_process_interface.cpp
+++ b/components/eamxx/src/physics/p3/eamxx_p3_process_interface.cpp
@@ -13,8 +13,6 @@
 namespace scream
 {
 
-using CP3 = physics::P3_Constants<Real>; 
-
 // =========================================================================================
 P3Microphysics::P3Microphysics (const ekat::Comm& comm, const ekat::ParameterList& params)
   : AtmosphereProcess(comm, params)
@@ -209,7 +207,7 @@ void P3Microphysics::initialize_impl (const RunType /* run_type */)
 
   //1350 will be from nl...
   //what is the
-  CP3::set_p3_autoconversion_factor(m_params.get<double>("p3_autoconversion_factor"));
+  CP3::p3_autoconversion_factor = m_params.get<double>("p3_autoconversion_factor");
 
   // Set property checks for fields in this process
   add_invariant_check<FieldWithinIntervalCheck>(get_field_out("T_mid"),m_grid,100.0,500.0,false);
--- a/components/eamxx/src/physics/p3/eamxx_p3_process_interface.cpp
+++ b/components/eamxx/src/physics/p3/eamxx_p3_process_interface.cpp
@@ -13,6 +13,8 @@
 namespace scream
 {
 
+using CP3 = physics::P3_Constants<Real>; 
+
 // =========================================================================================
 P3Microphysics::P3Microphysics (const ekat::Comm& comm, const ekat::ParameterList& params)
   : AtmosphereProcess(comm, params)
@@ -204,6 +206,11 @@ void P3Microphysics::initialize_impl (const RunType /* run_type */)
 {
   // Gather runtime options
   runtime_options.max_total_ni = m_params.get<double>("max_total_ni");
+
+  //1350 will be from nl...
+  //what is the
+  CP3::set_p3_autoconversion_factor(m_params.get<double>("p3_autoconversion_factor"));
+
   // Set property checks for fields in this process
   add_invariant_check<FieldWithinIntervalCheck>(get_field_out("T_mid"),m_grid,100.0,500.0,false);
   add_invariant_check<FieldWithinIntervalCheck>(get_field_out("qv"),m_grid,1e-13,0.2,true);
--- a/components/eamxx/src/physics/p3/eamxx_p3_process_interface.cpp
+++ b/components/eamxx/src/physics/p3/eamxx_p3_process_interface.cpp
@@ -206,8 +206,11 @@ void P3Microphysics::initialize_impl (const RunType /* run_type */)
   // Gather runtime options
   runtime_options.max_total_ni = m_params.get<double>("max_total_ni");
 
-  CP3::set_p3_defaults();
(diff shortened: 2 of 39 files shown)


oksanaguba:
@AaronDonahue suggested that instead of each individual set and get functions for each parameter, i use set(name, value) and get(name,value), to make it even easier to add more exposed params in future. my only reservation is string comparison in get function, in performant code. iirc @ambrad said before that string comparison does not come with big penalty.

components/eamxx/src/physics/share/physics_constants.hpp
```diff
@@ -126,6 +126,29 @@ struct Constants
   static constexpr Scalar earth_ellipsoid3 = 1.175;     // third expansion coefficient for WGS84 ellipsoid
 };
 
+template <typename Scalar>
+struct P3_Constants
+{
+
+private:
+   static Scalar p3_autoconversion_factor;
```
ambrad, line 134: One thing to watch out for is static variables (I don't mean `constexpr`s, but actual variables) that are accessed on both host and GPU. Without Cuda/HIP-specific annotations, that won't work, and it's a little messy to do that. The simplest portable way to handle a struct of set-able parameters is to make them non-static. You can pass a struct of non-static variables around and to device code like any other variable.

ambrad:
> string comparison does not come with big penalty

It's not a big deal at the top level of code and at initialization, which is probably the intent here. For clarity: Within kernel code, first, it can't be done b/c I'm pretty sure most or all of `std::stream` is still not implemented for the device; second, it shouldn't be done deep in loops, anyway.

components/eamxx/src/physics/share/physics_constants.hpp
```diff
@@ -126,6 +126,29 @@ struct Constants
...
+
+public:
+static void set_p3_defaults(){
+  set_p3_autoconversion_factor(1350.0);
+};
+
+static Scalar get_p3_autoconversion_factor(){
```
bartgol, line 141: Btw, there is no difference between having `p3_autoconversion_factor` public and having it private with get/set methods. If set/get are to be provided (and they don't perform any side-effect work, but simply set/get the value of the member), then it is much simpler to make the parameter a public member of the class.

oksanaguba:
@bartgol @ambrad  -- could you please see if this is acceptable? if it is, i will put the rest of parameters in here (i think more work might be needed for standalone tests, i haven't checked yet). my issues yesterdays were resolved after a fresh build.

components/eamxx/src/physics/share/physics_constants.hpp
```diff
@@ -126,6 +126,19 @@ struct Constants
   static constexpr Scalar earth_ellipsoid3 = 1.175;     // third expansion coefficient for WGS84 ellipsoid
 };
 
+template <typename Scalar>
+struct P3_Constants
+{
+  public:
+  Scalar p3_autoconversion_factor;
+
+  void set_p3_defaults(){
```
bartgol, line 135: You can put the defaults in the decl, and get rid of this method:
```
template <typename Scalar>
struct P3_Constants
{
  public:
  Scalar p3_autoconversion_factor = 1350.0;
};
```

components/eamxx/src/physics/p3/impl/p3_autoconversion_impl.hpp
```diff
@@ -13,18 +14,25 @@ void Functions<S,D>
...
   const Smask& context)
 {
+
   // Khroutdinov and Kogan (2000)
   const auto qc_not_small = qc_incld >= 1e-8 && context;
   constexpr Scalar CONS3 = C::CONS3;
+
+  Scalar p3_autoconversion_factor = loc_p3constants.p3_autoconversion_factor;
+
+//printf("  hey inside  AAAAAAAAAAAAAAAA %13.6f \n", p3_autoconversion_factor);
```
bartgol, line 27: When you're done with debugging, we can remove this line too.

components/eamxx/src/physics/p3/eamxx_p3_process_interface.cpp
```diff
@@ -204,6 +205,15 @@ void P3Microphysics::initialize_impl (const RunType /* run_type */)
...
   // Gather runtime options
   runtime_options.max_total_ni = m_params.get<double>("max_total_ni");
+
+  m_p3constants.set_p3_defaults();
+
+  //does this work when the value is not in the nl?
+  m_p3constants.p3_autoconversion_factor = m_params.get<double>("p3_autoconversion_factor");
+
+//temp to check   
+std::cout << "AAAAAAAAAAAAAANG "<<m_p3constants.p3_autoconversion_factor <<"\n" << std::flush;
```
bartgol, line 215: When you are done with debugging, we can remove this line.

bartgol: Or actually, we could keep it, but make it better. You could do something like this
```
  m_atm_logger->info("P3 Constants:");
  m_atm_logger->info("  p3 autoconversion factor: " + std::to_string(m_p3constants.p3_autoconversion_factor));
  // same for other params
```
This way we can always verify that the params were correct by looking at the atm log file.

bartgol (approved):
Looks good. I have a couple of suggestions, feel free to incorporate/ignore.

components/eamxx/src/physics/p3/impl/p3_autoconversion_impl.hpp
```diff
@@ -13,18 +14,25 @@ void Functions<S,D>
...
   const Spack& inv_qc_relvar, Spack& qc2qr_autoconv_tend, Spack& nc2nr_autoconv_tend, Spack& ncautr,
+  const physics::P3_Constants<S> & loc_p3constants,
   const Smask& context)
 {
+
   // Khroutdinov and Kogan (2000)
   const auto qc_not_small = qc_incld >= 1e-8 && context;
   constexpr Scalar CONS3 = C::CONS3;
+
+  Scalar p3_autoconversion_factor = loc_p3constants.p3_autoconversion_factor;
```
ambrad, line 25: Declare this as `const Scalar`, i.e., with `const`.

components/eamxx/src/physics/p3/impl/p3_main_impl.hpp
```diff
@@ -253,7 +254,7 @@ ::p3_main_internal(
       nr_incld, ni_incld, bm_incld, mu_c, nu, lamc, cdist, cdist1, cdistr,
       mu_r, lamr, logn0r, oqv2qi_depos_tend, precip_total_tend, nevapr, qr_evap_tend,
       ovap_liq_exchange, ovap_ice_exchange, oliq_ice_exchange,
-      pratot, prctot, hydrometeorsPresent, nk);
+      pratot, prctot, hydrometeorsPresent, nk, loc_p3constants);
```
ambrad, line 257: What does `loc_` mean?

ambrad:
This approach is sound. I see you haven't added the code yet to set the constants based on the YAML, but other than that, it looks like the approach is complete. I left a note about using `const` whenever possible and reconsidering the `loc_` part of the argument name.

components/eamxx/src/physics/p3/impl/p3_main_impl.hpp
oksanaguba: "local", but i see it is now not the right prefix here.

oksanaguba:
@ambrad @bartgol  not sure about YAML comment: the code right now has an entry in default namelist and 
```
  //does this work when the value is not in the nl?
  m_p3constants.p3_autoconversion_factor = m_params.get<double>("p3_autoconversion_factor");
```
Is more needed? Though i did have a question about namelist, what happens to the line m_params.get when the namelist does not have corresponding value? For now i am setting defaults before that.

ambrad:
Oh, sorry, I missed that line. Yes, that's right. I don't remember the syntax exactly, but there's a query method you can call to see if the parameter is in the list, like this:
```
a = 0.5; // default;
if (pl.present("a"))  // something like this
  a = pl.get ...
```

bartgol:
> @ambrad @bartgol not sure about YAML comment: the code right now has an entry in default namelist and
> 
> ```
>   //does this work when the value is not in the nl?
>   m_p3constants.p3_autoconversion_factor = m_params.get<double>("p3_autoconversion_factor");
> ```
> 
> Is more needed? Though i did have a question about namelist, what happens to the line m_params.get when the namelist does not have corresponding value? For now i am setting defaults before that.

The method Andrew refers to is `params.isParameter ("blah")`. You could also consider adding a `set_params` method to the `P3_Constant` struct. Something like this:

```
void P3_Constants::set_params (ekat::ParameterList& params) {
  p3_autoconversion_factor = params.get<double>("p3_autoconversion_factor",1350.0); // <-- If not found in the param list, set 1350 in the param list, and return it.
  ...
}
```
which you can then call from P3's constructor.

components/eamxx/cime_config/namelist_defaults_scream.xml
```diff
@@ -203,6 +203,9 @@ be lost if SCREAM_HACK_XML is not enabled.
         ${DIN_LOC_ROOT}/atm/scream/tables/vn_table_vals.dat8,
         ${DIN_LOC_ROOT}/atm/scream/tables/vm_table_vals.dat8
       </tables>
+      <p3_pre_autoconversion_factor type="real" doc="P3 pre_autoconversion_factor">1350.0</p3_pre_autoconversion_factor>
```
bartgol, line 206: Can we make the doc string a little bit more explanatory? So far they seem to just restate the name of the param, but I have no idea what "spa_to_nc" means...

I know we're in a rush to get these params exposed for the autotuning work, so I understand if you want to punt. If you want to punt, maybe some other dev with P3 knowledge could help? @AaronDonahue maybe?

oksanaguba: i do not know what these parameters mean either, if anyone wants to add another column for descriptions in https://acme-climate.atlassian.net/wiki/spaces/EAMXX/pages/4055498787/Auto-Calibration+Tuning+Parameters i will put them in the pr. thanks!

components/eamxx/src/physics/p3/impl/p3_autoconversion_impl.hpp
```diff
@@ -13,18 +14,25 @@ void Functions<S,D>
...
+//printf("  hey inside  AAAAAAAAAAAAAAAA %13.6f \n", p3_autoconversion_factor);
+
   if(qc_not_small.any()){
     Spack sgs_var_coef;
     // sgs_var_coef = subgrid_variance_scaling(inv_qc_relvar, sp(2.47) );
     sgs_var_coef = 1;
 
     qc2qr_autoconv_tend.set(qc_not_small,
-              sgs_var_coef*1350*pow(qc_incld,sp(2.47))*pow(nc_incld*sp(1.e-6)*rho,sp(-1.79)));
+              sgs_var_coef*p3_pre_autoconversion_factor*pow(qc_incld,sp(2.47))*pow(nc_incld*sp(1.e-6)*rho,sp(-1.79)));
```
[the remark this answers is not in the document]
oksanaguba: will do, thanks

oksanaguba:
About unit tests and F90 code, for example there is code
```
void ice_sedimentation_f(
  Int kts, Int kte, Int ktop, Int kbot, Int kdir,
  Real* rho, Real* inv_rho, Real* rhofaci, Real* cld_frac_i, Real* inv_dz,
  Real dt, Real inv_dt,
  Real* qi, Real* qi_incld, Real* ni, Real* qm, Real* qm_incld, Real* bm, Real* bm_incld,
  Real* ni_incld, Real* precip_ice_surf, Real* qi_tend, Real* ni_tend)
{
  using P3F  = Functions<Real, DefaultDevice>;
...
    P3F::ice_sedimentation(
      rho_d, inv_rho_d, rhofaci_d, cld_frac_i_d, inv_dz_d,
      team, wsm.get_workspace(team),
      nk, ktop, kbot, kdir, dt, inv_dt,
      qi_d, qi_incld_d, ni_d, ni_incld_d, qm_d, qm_incld_d, bm_d, bm_incld_d,
      qi_tend_d, ni_tend_d, ice_table_vals,
      precip_ice_surf_k);
...
}
```
So F function calls cxx function, which now needs an extra argument, p3constants struct. 
What is the best way to handle it here? i am not familiar how unit tests are organized in eamxx.

oksanaguba:
Could i use an ifdef and avoid this for standalone execs?

oksanaguba:
@bartgol @ambrad please see my question above. i now think the easiest way forward is ifdef (there must be a scream cime idfef). this PR needs to be merged in a week, thanks.

ambrad:
If I understand the issue, I think you should be able to just do something like this:
```c++
    P3F::ice_sedimentation(
      rho_d, inv_rho_d, rhofaci_d, cld_frac_i_d, inv_dz_d,
      team, wsm.get_workspace(team),
      nk, ktop, kbot, kdir, dt, inv_dt,
      qi_d, qi_incld_d, ni_d, ni_incld_d, qm_d, qm_incld_d, bm_d, bm_incld_d,
      qi_tend_d, ni_tend_d, ice_table_vals,
      precip_ice_surf_k, P3_Constants<Real>());
```

bartgol:
This would only do the C-vs-F comparison for the default values of the parameters. I guess that's just what we are doing now, so maybe that's ok.

Perhaps a good question (maybe for @PeterCaldwell) is "how long are we going to keep the bulky C-vs-F intercomparison testing? It seems to me that, after all the production runs we have done, we "trust" the C++ implementation to do the right thing. If we are not going to backport changes to F90, maybe we should pull the plug sooner than later?

PeterCaldwell:
> If we are not going to backport changes to F90, maybe we should pull the plug sooner than later?

I agree that the C++ code is "trusted" these days. The plan has always been to drop all the Fortran/C++ testing once DP runs in C++. The reason we might want to wait until DP is done is that we often use DP for debugging and if the only DP version we had was totally disconnected from the global code we care about, we would be in trouble. Though I feel like our trust that F90 DP is a good proxy for C++ global runs has been waning, so maybe we don't care. 

I'm totally fine with just disabling this *single* test that is giving us problems and keeping the rest of the F/C++ testing until DP is done. Though maybe/probably Oksana's example is just one of many examples and this change would force us to get rid of all P3 unit tests?

ambrad:
@bartgol @PeterCaldwell I might be wrong in my suggested fix, but I'm thinking what I wrote is right. If so, then there is no need to drop any tests. The tests will continue to run with the same, now default, values. But I think that's fine: adding additional F90-vs-C++ testing in the standalone tests specifically for non-default values has essentially 0 priority, IMO. And we'll test the key thing: that this PR, with no changes to the XML, will reproduce master.

oksanaguba:
for this commit, v1 cime run ne4pg2 5 days is bfb for restart files:
```
[onguba@login12.frontier run]$ pwd
/ccs/home/onguba/eff/v1-expose-jan16/run
[onguba@login12.frontier run]$ /lustre/orion/cli115/world-shared/e3sm/tools/cprnc/cprnc -m v1-expose-jan16.cpl.r.0001-01-06-00000.nc ~/eff/v1-master-jan17/run/v1-master-jan17.cpl.r.0001-01-06-00000.nc
```
sanity checks:
```
[onguba@login12.frontier run]$ pwd
/ccs/home/onguba/eff/v1-expose-jan16/run
[onguba@login12.frontier run]$ zgrep "P3" atm.log.157
atm.log.1576971.240116-192629.gz  atm.log.1576981.240116-194413     
atm.log.1576976.240116-193734     atm.log.1577016.240116-195234.gz  
[onguba@login12.frontier run]$ zgrep "P3" atm.log.1577016.240116-195234.gz 
P3 Constants:
P3   p3_autoconversion_prefactor = 1350.000000
P3   p3_mu_r_constant = 1.000000
P3   p3_spa_to_nc = 1.000000
P3   p3_k_accretion = 67.000000
P3   p3_eci = 0.500000
P3   p3_eri = 1.000000
P3   p3_rho_rime_min = 50.000000
P3   p3_rho_rime_max = 900.000000
P3   p3_a_imm = 0.650000
P3   p3_dep_nucleation_exponent = 0.304000
P3   p3_ice_sed_knob = 1.000000
P3   p3_d_breakup_cutoff = 0.000280
[onguba@login12.frontier run]$ zgrep "P3" ~/eff/v1-master-jan17/run/atm.log.1577902.240117-105249.gz 
[onguba@login12.frontier run]$ zgrep "P3" ~/eff/v1-master-jan17/run/atm.log.1577902.240117-105249.gz | wc -l
0
```

ambrad:
Is this ready for review? If so, don't forget to remove "[wip]" so that the AT will run.

components/eamxx/src/physics/p3/eamxx_p3_process_interface.cpp
```diff
@@ -204,6 +205,12 @@ void P3Microphysics::initialize_impl (const RunType /* run_type */)
 {
   // Gather runtime options
   runtime_options.max_total_ni = m_params.get<double>("max_total_ni");
+
+  //////////// setting P3 constants in a struct
```
ambrad, line 209: These comment decorations are a little excessive, but I'm not strongly opposed.

ambrad (approved):
Looks great!

components/eamxx/src/physics/p3/eamxx_p3_process_interface.cpp
oksanaguba: fixed, thanks

oksanaguba:
removed [wip] comment to test

ambrad (dismissed):

oksanaguba:
i am not sure what/if there is an issue with tests on mappy, folder is 
```
/home/e3sm-jenkins/jenkins-ws/workspace/SCREAM_PullRequest_Autotester_Mappy/4778
```
i see a few lines like `100% tests passed, 0 tests failed out of 279` for standalone and a bunch of passes(?) for v1,
```
PASS ERP_D_Lh4.ne4_ne4.F2010-SCREAMv1.mappy_gnu9 RUN
    Case dir: /ascldap/users/e3sm-jenkins/acme/scratch/ERP_D_Lh4.ne4_ne4.F2010-SCREAMv1.mappy_gnu9.C.20240118_123540_rrwwdh
PASS ERP_Ln22.ne4pg2_ne4pg2.F2010-SCREAMv1.mappy_gnu9 RUN
    Case dir: /ascldap/users/e3sm-jenkins/acme/scratch/ERP_Ln22.ne4pg2_ne4pg2.F2010-SCREAMv1.mappy_gnu9.C.20240118_123540_rrwwdh
PASS ERS_D_Ln22.ne4pg2_ne4pg2.F2010-SCREAMv1.mappy_gnu9.scream-rad_frequency_2 RUN
    Case dir: /ascldap/users/e3sm-jenkins/acme/scratch/ERS_D_Ln22.ne4pg2_ne4pg2.F2010-SCREAMv1.mappy_gnu9.scream-rad_frequency_2.C.20240118_123540_rrwwdh
PASS ERS_Ln9.ne4_ne4.F2000-SCREAMv1-AQP1.mappy_gnu9 RUN
    Case dir: /ascldap/users/e3sm-jenkins/acme/scratch/ERS_Ln9.ne4_ne4.F2000-SCREAMv1-AQP1.mappy_gnu9.C.20240118_123540_rrwwdh
PASS PET_Ln9_P32x2.ne4pg2_ne4pg2.F2010-SCREAMv1.mappy_gnu9 RUN
    Case dir: /ascldap/users/e3sm-jenkins/acme/scratch/PET_Ln9_P32x2.ne4pg2_ne4pg2.F2010-SCREAMv1.mappy_gnu9.C.20240118_123540_rrwwdh
PASS SMS_D_Ln9.ne4_ne4.F2010-SCREAMv1-noAero.mappy_gnu9 RUN
    Case dir: /ascldap/users/e3sm-jenkins/acme/scratch/SMS_D_Ln9.ne4_ne4.F2010-SCREAMv1-noAero.mappy_gnu9.C.20240118_123540_rrwwdh
```

So is mappy ok or not?

oksanaguba:
On weaver, 
```
cd /home/e3sm-jenkins/weaver/workspace/SCREAM_PullRequest_Autotester_Weaver/5246
...
30 - physics_saturation_run_and_cmp (Failed)
```
Not sure why it is ok on mappy, but not on weaver.

bartgol:
The mappy build passed. If you click on the line `Pull Request Auto Testing has FAILED (click to expand)`, you can see a summary, and it says that Mappy PASSED. The issue is just the run_and_cmp fail on weaver. Looking at the full log on the jenkins web interface, I see all diff messages saying something like this

```
14:47:04 --- checking physics saturation case # 10
14:47:04 sat_ice_fp > tol 1 times. Max rel diff= inf normalized by ref impl val=0.
14:47:04 sat_liq_fp > tol 1 times. Max rel diff= inf normalized by ref impl val=0.
14:47:04 mix_ice_fr > tol 1 times. Max rel diff= inf normalized by ref impl val=0.
14:47:04 mix_liq_fr > tol 1 times. Max rel diff= inf normalized by ref impl val=0.
14:47:04 sat_ice_mkp > tol 1 times. Max rel diff= inf normalized by ref impl val=0.
14:47:04 sat_liq_mkp > tol 1 times. Max rel diff= inf normalized by ref impl val=0.
14:47:04 mix_ice_mkr > tol 1 times. Max rel diff= inf normalized by ref impl val=0.
14:47:04 mix_liq_mkr > tol 1 times. Max rel diff= inf normalized by ref impl val=0.
14:47:04 Ref impl failed.
```

ambrad:
> Andrew hinted before that he would know @ambrad thanks!

I was referring to diffs in the single-precision build that could result from not using `sp(number)`. The output Luca reports suggests something else is amiss.

components/eamxx/src/physics/share/tests/physics_saturation_run_and_cmp.cpp
```diff
@@ -189,6 +189,7 @@ struct UnitWrap::UnitTest<D>::TestSaturation
     Scalar mix_ice_mkr;
     Scalar mix_liq_mkr;
 
+    KOKKOS_INLINE_FUNCTION
```
oksanaguba, line 192: @ambrad @bartgol -- could this be a nonbfb change? i am not seeing direct connection between this pr and saturation code (there may be, but i do not see anything obvious), except for this line. Fro Jenkins output
```
[onguba@weaver11 5246]$ grep "saturation.baseline" JENKINS_2024-01-18_124134 | grep "at tol"
Comparing with /home/projects/e3sm/scream/pr-autotester/master-baselines/weaver/full_sp_debug/data/physics_saturation.baseline4 at tol 1.2e-07
Comparing with /home/projects/e3sm/scream/pr-autotester/master-baselines/weaver/release/data/physics_saturation.baseline8 at tol 2.2e-12
Comparing with /home/projects/e3sm/scream/pr-autotester/master-baselines/weaver/full_debug/data/physics_saturation.baseline8 at tol 2.2e-16
Comparing with /home/projects/e3sm/scream/pr-autotester/master-baselines/weaver/full_sp_debug/data/physics_saturation.baseline4 at tol 1.2e-07
Comparing with /home/projects/e3sm/scream/pr-autotester/master-baselines/weaver/release/data/physics_saturation.baseline8 at tol 2.2e-12
Comparing with /home/projects/e3sm/scream/pr-autotester/master-baselines/weaver/full_debug/data/physics_saturation.baseline8 at tol 2.2e-16
```
So errors maybe not too small for SP runs.

ambrad: Anything's possible, but in principle it shouldn't change answers. But it also seems like it should have not have built without it. My only recommendation is to study this test on Weaver, using `printf`s, etc. One thing that might provide clues is looking at why KOKKOS_INLINE_FUNCTION was apparently not needed on Weaver previously.

oksanaguba:
I reverted KOKKOS_INLINE_FUNCTION declaration for overloaded operator and now 
```All Jobs Finished; status = PASSED, However Inspection must be performed before merge can occur...```
I agree the questions Andrew asked are interesting and worth discovering, but also this PR is needed now. Would everyone agree I should put back KOKKOS_INLINE_FUNCTION and declare this is the reason for nonbfb and be done with the PR?
@ambrad @bartgol  ? thanks

**Comment by ambrad** [comment]
I agree. It's peculiar, but those tests are among the least important, so it's hard to argue for allocating time to investigate further.

**Review by ambrad — APPROVED** [review]

**Comment by ambrad** [comment]
Actually, I realized one issue is that this test won't be fixed by baseline regeneration. @bartgol, isn't that right? Oksana, since we can't have a test failing in the AT, I think you need to disable the section of the test that is failing if you can't spend time to investigate the failure further.

**Comment by bartgol** [comment]
Could it be that the problem is the lack of a `Kokkos::reduction_identity` (or whatever it is called) specialization for the OutputData class?

**Comment by bartgol** [comment]
LoL, it appears that the _reference_ values are all zeros:

```c++
      std::cout << "--- checking physics saturation case # " << case_num << std::endl;
      read(fid, ref);
      printf("sat_ice_fp: %f\n",ref.sat_ice_fp);
      printf("sat_liq_fp: %f\n",ref.sat_liq_fp);
      printf("mix_ice_fr: %f\n",ref.mix_ice_fr);
      printf("mix_liq_fr: %f\n",ref.mix_liq_fr);
```
prints
```
--- checking physics saturation case # 10
sat_ice_fp: 0.000000
sat_liq_fp: 0.000000
mix_ice_fr: 0.000000
mix_liq_fr: 0.000000
```

Is it possible our baselines got corrupted somehow. I am not sure how other PRs are not seeing this...

bartgol:
I'm thinking this is wrong
```c++
    ekat::read(&d.sat_ice_fp, 1, fid);
    ekat::read(&d.sat_liq_fp, 1, fid);
    ekat::read(&d.mix_ice_fr, 1, fid);
    ekat::read(&d.mix_liq_fr, 1, fid);
```
The function `ekat::read` is declared as
```c++
template <typename Scalar>
void read (const char* filename, Scalar* a, const int n) {
```
I think types are mixed up, but the compiler can convert one to the other. However, I would expect this
```c++
  FILEPtr fid(fopen(filename, "r"));
  EKAT_REQUIRE_MSG( fid, "Could not open " << filename << " for reading.");
```
to throw... anyways, I'm going to try changing the read calls.

bartgol:
Cat-ing the physics baselines returns nothing, as opposed to the p3 baseline which returns a bunch of gibberish symbols (since the files are in binary format). I think the baselines got corrupted, but I don't have the time to investigate how that happened, so I would not hold off the merge based on that test. The next AT run _should_ regenerate the baselines, and other PRs should test correclty.

I'm very puzzled by the fact that other PRs are not producing an error for this test though.

bartgol (approved):

oksanaguba:
the button to merge is not active in here, what's the procedure -- to wait? i will ask to re-test.

PR #2658, opened 2024-01-09
DP-SCREAM v0 diagnostic coordinate fix

bogensch:
Fixes DP SCREAM output coordinate issue where x-coordinates should be associated with spherep lon values and y-coordinates should be associated with lat values.  This is a diagnostic output fix and thus B4B.

@tcclevenger if this section of code has already been converted to v1 could you let me know and I will make the fix.

### DIFF
--- a/components/eam/src/dynamics/se/stepon.F90
+++ b/components/eam/src/dynamics/se/stepon.F90
@@ -585,8 +585,8 @@ subroutine stepon_run3(dtime, cam_out, phys_state, dyn_in, dyn_out)
    if (dp_crm) then
 
      do ie=1,nelemd
-       out_gridx(:,:) = dyn_in%elem(ie)%spherep(:,:)%lat
-       out_gridy(:,:) = dyn_in%elem(ie)%spherep(:,:)%lon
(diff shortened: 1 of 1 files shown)


tcclevenger (approved):
I'm actually not sure how this will relate to the EAMxx version since we do not have the same `outfld` method. I'm assuming this is defining a field to be output? What exactly is being output here?

bogensch:
The planar grid coordinates are being output in units of meters; which are essential coordinates for diagnostic purposes.

tcclevenger:
@bogensch I can't really tell what the mappy fails are from the output here, and I think the build isn't on jenkins dashboard anymore. I'll retest and then make sure to look at the log when it finishes. Odd that it would fail since I don't think there are any V1 F90 tests that set `dp_crm=true`..

tcclevenger:
@bogensch Looks like there is a DP v0 test and that is what is diffing (because you changed the output of 2 fields). This should be expected. If this sounds right to you, I can manually merge.

bogensch:
Thanks @tcclevenger. Yes, that would be expected.

PR #2659, opened 2024-01-09
Avoid pointless work in FieldWithinIntervalCheck

### LINKED ISSUE #2657: In property checks, do not add expensive info to the msg unless requested.
Right now, in our property checks, when a check fails, we print a message with a bunch of information, including the check specs, the location of the fail, and possibly some additional fields at the location of the failure.

Some of this data is small and already available (e.g., the computed max/min of the field in `FieldWithinIntervalCheck`), and we can always add it to the msg. Other data, however, is not. E.g., in `FieldWithinIntervalCheck`, any additional data fields (like phis) that is stored in the class, is first synced to host, then fully printed at the location of the max/min values (which includes a sync to host call to ensure data is up to date before calling `print_field_hyperslab`).

If the check failed and cannot be repaired, the fact that the extra data added to the msg was expensive is not a big deal, since we're about to crash, and we _do_ want maximum info in the log. However, if the check passed, or if the property will be repaired, we _don't_ want to spend time adding info that we are not interested in anyways.

One possibility is to only add the extra info if the test is Failed, while skip the extra info if the test result is Passed or Repairable.

bartgol:
Fixes #2657

### DIFF
--- a/components/eamxx/src/share/atm_process/atmosphere_process.cpp
+++ b/components/eamxx/src/share/atm_process/atmosphere_process.cpp
@@ -458,28 +458,37 @@ void AtmosphereProcess::run_property_check (const prop_check_ptr&       property
 
 void AtmosphereProcess::run_precondition_checks () const {
   m_atm_logger->debug("[" + this->name() + "] run_precondition_checks...");
+  start_timer(m_timer_prefix + this->name() + "::run-precondition-checks");
   // Run all pre-condition property checks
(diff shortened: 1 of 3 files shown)


bartgol:
On CPU, the perf gain appear to be minimal, if anything at all. I suspect though that on GPU the sync_to_host call we are avoiding for the additional fields in the prop check will prove non-negligible. Small, perhaps, but measurable.

tcclevenger (approved):

PR #2665, opened 2024-01-11
Fix handling of avg count in scorpio output

### LINKED ISSUE #2663: Problem with avg count in scorpio output class
One of the output yaml files I got from @ndkeen exposed an issue in the way we handle the average count in our output. The yaml file is the following:
```
Averaging Type: Instant
Fields:
  Physics PG2:
    Field Names:
    - SeaLevelPressure
    - wind_speed_10m
    - T_mid_at_400hPa
    - z_mid_at_500hPa
    - z_mid_at_200hPa
Max Snapshots Per File: 24
Restart:
  force_new_file: true
filename_prefix: output.scream.Cess.3hourlyINST_ne1024
output_control:
  Frequency: 3
  MPI Ranks in Filename: false
  frequency_units: nhours
```
As you can see, there are only 2d fields registered, but two of them, `z_mid_at_***Pa` are a diagnostic of another diagnostic (namely `FieldAtPressureLevel` composed with `VerticalLayer`).

When we create diagnostics, we also register the avg count for that diagnostic. So in this case, we register an avg count for `z_mid` and one for `z_mid_at_***Pa`. The former has layout `(ncol,lev)`, while the latter has layout `(ncol)`. When we register variables in the file, we try to stuff both these avg counts in the file, but we hit a problem: `avg_count_ncol_lev` needs a dimension `lev` which was _never_ registered in the NC file, since no output vars had the `lev` dimension.

The best solution would be to not register `avg_count_ncol_lev` in the output file, since no output var has that layout.

bartgol:
Fixes handling of avg count extra variable in scorpio output, to avoid adding avg count for layouts that belong to intermediate diagnostic quantities. See #2663 for more details. I verified running the yaml file in that issue that this PR fixes the error.

Also, fix a bug in the fvphys layer in dynamics: when remapping rrtmgp active gases, we were not initializing the field timestamp. If the field was outputed in an INSTANT output stream, I/O would overwrite the field with a fill_value, causing havoc at rrtmpg's first step.

Fixes #2663

### DIFF
--- a/components/eamxx/src/share/io/scorpio_output.cpp
+++ b/components/eamxx/src/share/io/scorpio_output.cpp
@@ -848,31 +848,41 @@ void AtmosphereOutput::register_views()
 /* ---------------------------------------------------------- */
 void AtmosphereOutput::set_avg_cnt_tracking(const std::string& name, const std::string& avg_cnt_suffix, const FieldLayout& layout)
 {
-    // Make sure this field "name" hasn't already been regsitered with avg_cnt tracking.
-    // Note, we check this because some diagnostics need to have their own tracking which
-    // is created at the 'create_diagnostics' function.
-    if (m_field_to_avg_cnt_map.count(name)>0) {
-      return;
+  // Make sure this field "name" hasn't already been regsitered with avg_cnt tracking.
+  // Note, we check this because some diagnostics need to have their own tracking which
+  // is created at the 'create_diagnostics' function.
+  if (m_field_to_avg_cnt_map.count(name)>0) {
+    return;
+  }
+
+  // If the field is not an output field, do not register the avg count. This can happen
+  // if a diag depends on another diag. In this case, the inner diag is never outputed,
+  // so we don't want to create an avg count for its layout, since it may contain dims
+  // that are not in the list of registered dims (the dims of the output vars).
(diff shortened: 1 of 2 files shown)


components/eamxx/src/share/io/scorpio_output.cpp
```diff
@@ -848,31 +848,41 @@ void AtmosphereOutput::register_views()
 /* ---------------------------------------------------------- */
 void AtmosphereOutput::set_avg_cnt_tracking(const std::string& name, const std::string& avg_cnt_suffix, const FieldLayout& layout)
 {
-    // Make sure this field "name" hasn't already been regsitered with avg_cnt tracking.
-    // Note, we check this because some diagnostics need to have their own tracking which
-    // is created at the 'create_diagnostics' function.
-    if (m_field_to_avg_cnt_map.count(name)>0) {
-      return;
+  // Make sure this field "name" hasn't already been regsitered with avg_cnt tracking.
```
AaronDonahue, line 851: There is a typo in "registered" which is held over from when I put this comment in the code.

AaronDonahue (approved):

components/eamxx/src/share/io/scorpio_output.cpp
bartgol: We can put it in one of our next PRs.

PR #2668, opened 2024-01-13
Some fixes/upgrades to avg count logic in output classes

bartgol:
There were mainly two things that could be done better:

- Ensure that avg_cnt logic is OFF for Instant output, or if the time dimension is not added to the vars (e.g., the geo_data streams used when a new file is setup). Currently, we were adding avg count logic also to instant output, and we were not correctly guarding all uses of avg cnt views (e.g., we were always creating an avg cnt for diagnostics, regardless of whether the avg cnt track was ON or OFF).
- No need to store a view of 1/0 to establish whether a certain field was filled or not. We can simply:
  - keep track of which avg_cnt fields were processed (so we don't update it twice when another field that share the same avg_cnt view is processed)
  - check the field new value to determine if it was filled
  
This allows to save
 - time: one less deep copy per avg cnt, as well as one less pfor for each field
 - space: one less view for each avg cnt

### DIFF
--- a/components/eamxx/src/share/io/scorpio_output.cpp
+++ b/components/eamxx/src/share/io/scorpio_output.cpp
@@ -1379,105 +1379,98 @@ AtmosphereOutput::create_diagnostic (const std::string& diag_field_name) {
 // Helper function to mark filled points in a specific layout
 void AtmosphereOutput::
 update_avg_cnt_view(const Field& field, view_1d_dev& dev_view) {
-  // If the dev_view_1d is aliasing the field device view (must be Instant output),
-  // then there's no point in copying from the field's view to dev_view
   const auto& name   = field.name();
   const auto& layout = m_layouts.at(name);
   const auto& dims   = layout.dims();
-  const auto  rank   = layout.rank();
         auto  data   = dev_view.data();
-  const bool is_diagnostic = (m_diagnostics.find(name) != m_diagnostics.end());
-  const bool is_aliasing_field_view =
-      m_avg_type==OutputAvgType::Instant &&
-      field.get_header().get_alloc_properties().get_padding()==0 &&
-      field.get_header().get_parent().expired() &&
-      not is_diagnostic;
   const auto fill_value = m_fill_value;
-  if (not is_aliasing_field_view) {
-    KT::RangePolicy policy(0,layout.size());
-    const auto extents = layout.extents();
-    switch (rank) {
-      case 1:
-      {
-        // For rank-1 views, we use strided layout, since it helps us
-        // handling a few more scenarios
-        auto src_view_1d = field.get_strided_view<const Real*,Device>();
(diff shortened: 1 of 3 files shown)


bartgol:
@AaronDonahue I removed the possibility of disabling fill tracking in IO. My rationale is that it makes no sense to not track fill if we are in a scenario where fill is active. If we didn't, the output avg would have no meaning, since it would be a mix of filled and real values. I also don't think we were even using this feature, right?

components/eamxx/src/share/io/scorpio_output.cpp
```diff
@@ -418,37 +425,36 @@ run (const std::string& filename,
...
+      auto field = get_field(name,"io");
+      update_avg_cnt_view(field,m_dev_views_1d.at(avg_cnt_name));
+
+      // Make sure we don't double update this avg cnt
+      avg_updated.insert(avg_cnt_name);
     }
   }
 
   // Take care of updating and possibly writing fields.
+  // These are needed inside kernels, so crate local copies
```
AaronDonahue, line 453: misspelled "create"

components/eamxx/src/share/io/scorpio_output.cpp
```diff
@@ -1293,7 +1272,10 @@ AtmosphereOutput::create_diagnostic (const std::string& diag_field_name) {
       } else if (units=="mb" or units=="Pa" or units=="hPa") {
         diag_name = "FieldAtPressureLevel";
         diag_avg_cnt_name = "_" + tokens[1]; // Set avg_cnt tracking for this specific slice
-        m_track_avg_cnt = true; // If we have pressure slices we need to be tracking the average count.
+
+        // If we have pressure slices we need to be tracking the average count,
```
AaronDonahue, line 1276: Can we also add for FieldAtHeight?

AaronDonahue, line 1276: We don't seem to mask for FieldAtHeight above sealevel, but presumably, 2m above sealevel should be masked in mountainous areas... right?  Maybe something to consider for a different PR.

AaronDonahue (approved):

components/eamxx/src/share/io/scorpio_output.cpp
bartgol: Ah, yes, `FieldAtHeight` with "sealevel" as reference would trigger filling, I think. But I don't think we have that PR merged yet. Perhaps you could do the same for `FieldAtHeight` in your PR, if the ref is "sealevel" and the height is less than 9km?

AaronDonahue: I think the current FieldAtHeight diagnostic extrapolates, which is why we don't actually have an issue.  We would need to change the vertical interpolation routine in the `run_impl` to mask output.  For the sake of getting the FieldAtHeight PR in before the decadal runs I think it would be better to make this change in a separate PR.

Issue #2671, opened 2024-01-13
scorpio interface doesn't seem to handle reading scalar0d fields...


mahf708:
Consider this check:

https://github.com/E3SM-Project/scream/blob/cc36493037d40d577bfe0bec8fc075dd6c109f3b/components/eamxx/src/share/io/scream_scorpio_interface.cpp#L342

What do we expect the second conditional to be in the case of a scalar0d layout? Hint: SEGFAULT (!)

bartgol:
I am a bit puzzled about why one would save a 0d quantity that doesn't depend on time as a PIO variable rather than a PIO global attribute. In my mind, a var must have some sort of dimension.

PR #2672, opened 2024-01-13
in scorpio interface, only access array when it is not empty

mahf708:
See #2671 for details

### DIFF
--- a/components/eamxx/src/share/io/scream_scorpio_interface.cpp
+++ b/components/eamxx/src/share/io/scream_scorpio_interface.cpp
@@ -338,11 +338,20 @@ void register_variable(const std::string &filename, const std::string& shortname
           " - pio error: " + std::to_string(err) + "\n");
     }
 
-
-    if (mode==Read && (dims_from_file[0]=="time" && var_dimensions[0]!="time")) {
-      // For Read operations, we may not consider "time" as a field dimension, so if the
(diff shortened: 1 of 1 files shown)


bartgol (approved):

components/eamxx/src/share/io/scream_scorpio_interface.cpp
```diff
@@ -338,11 +338,20 @@ void register_variable(const std::string &filename, const std::string& shortname
...
     }
 
-
-    if (mode==Read && (dims_from_file[0]=="time" && var_dimensions[0]!="time")) {
-      // For Read operations, we may not consider "time" as a field dimension, so if the
-      // input file has "time", simply disregard it in this check.
-      dims_from_file.erase(dims_from_file.begin());
+    // Here, let's only try to access var_dimensions[0] when we know for sure
+    // that var_dimensions is actually dimensioned (i.e., .size()>0)
+    if (var_dimensions.size()>0) {
```
bartgol, line 343: Up to you, but you could also do this:
```
if (mode==Read and dims_from_file[0]=="time" and ekat::contains(var_dimensions,"time")) {
 ...
}
```
I wrote the `ekat::contains(x,y)` function as a shortcut to the usual
```
std::find(x.begin(),x.end(),y)!=x.end()
```
If the input container is empty, it simply returns false.

PR #2675, opened 2024-01-19
Enable mlcorrection unit test on pm-cpu

elynnwu:
We extend the ability to run MLCorrection unit test to pm-cpu. A shared python venv is installed under `$CFS/e3sm/eamxx-ml/` and can be loaded through module load to play nicely with CIME when running the full model.

To build and run this test, assuming WORK_DIR is the build folder and CODE_ROOT is scream's top directory:
```
mkdir -p $WORK_DIR/baselines/full_debug/data
$CODE_ROOT/components/eamxx/scripts/test-all-scream -w $WORK_DIR -k -m pm-cpu -t dbg --limit-test-regex=ml_correction
```

There's something wrong with the srun commands because I'm getting:
```
srun: error: Job request does not match any supported policy.
srun: error: Unable to allocate resources: Unspecified error
```

However, once `test-all-scream` compiles the test, running directly inside the ml_correction test works:
```
cd $WORK_DIR/full_debug/tests/uncoupled/ml_correction
# make sure to load the same env test-all-scream uses
SCREAM_MACHINE=pm-cpu . ${CODE_ROOT}/components/eamxx/scripts/source_to_load_scream_env.sh
./ml_correction_standalone
# or alternatively submit through salloc
salloc --time 00:30:00 --nodes=1 --constraint=cpu -q debug --account ACCOUNT ml_correction_standalone
```


ndkeen:
Why would we want to source `/global/cfs/cdirs/e3sm/eamxx-ml/python_venv/3.9.13/screamML/bin/activate`?  Can this only be done for certain experiments? It makes me uneasy to change the environment

bartgol (approved):

components/eamxx/scripts/machines_specs.py
```diff
@@ -57,6 +57,10 @@
                 ["mpicxx","mpifort","mpicc"],
                 "bsub -I -q batch -W 0:30 -P cli115 -nnodes 1",
                 "/gpfs/alpine/cli115/proj-shared/scream/master-baselines"),
+    "pm-cpu" : (["module purge", "module use --append /global/cfs/cdirs/e3sm/eamxx-ml/install/install/modulefiles", "module load PrgEnv-gnu gcc-native/12.3 cray-libsci craype cray-mpich cray-hdf5-parallel cray-netcdf-hdf5parallel cray-parallel-netcdf cmake evp-patch cray-python/3.9.13.1 screamML-venv/0.0.1", "export LD_LIBRARY_PATH=/opt/cray/pe/python/3.9.13.1/lib:/global/common/software/nersc/pm-2022q4/sw/evp-patch/lib64:/opt/cray/libfabric/1.15.2.0/lib64:/usr/lib64/gcc/x86_64-suse-linux/12"],
```
elynnwu, line 60: I believe there's a bug in Perlmutter's `LD_LIBRARY_PATH`, or perhaps I'm loading an extra module that I shouldn't be. When using `gcc-native/12.3`, it should look under `/usr/lib64/gcc/x86_64-suse-linux/12` but `/opt/cray/pe/gcc-libs` keeps showing up in `LD_LIBRARY_PATH` despite module purge first. This is the error:
```
/opt/cray/pe/gcc-libs/libstdc++.so.6: version `GLIBCXX_3.4.30' not found
/opt/cray/pe/gcc-libs/libgcc_s.so.1: version `GCC_12.0.0' not found
```

As a workaround, I'm manually overwriting `LD_LIBRARY_PATH` to remove `/opt/cray/pe/gcc-libs`. Either we fix this by finding out which module causes this path to be added and removing that, or we should open a ticket to ask support if this is an unexpected behavior.

bartgol:
@AaronDonahue @elynnwu what do we want to do with this PR? If you think it's valuable, we can convert to non-daft, and let the AT merge. But I think the convo between Elynn and Noel above was not really resolved...

AaronDonahue:
Closing this pending some work on the ML side

PR #2678, opened 2024-01-23
Share some mapping data across horizontal remappers

bartgol:
In particular, we share the coarse/ov_coarse grids, and the CRS matrix views. The goal is not much about saving memory, but rather computing time, since reading the map files and creating the CRS matrices can have a nontrivial cost.


tcclevenger (dismissed):
Looks good to me.

tcclevenger (approved):

PR #2682, opened 2024-01-24
add inputdata unit test

mahf708:
Adds a job to monitor consistency between inputdata server files and what we have in our configs. The job runs at every PR, every push to master, and everyday at 000000 (not sure which time zone, maybe UTC). If the job fails, it will show up as failing (thus hopefully stopping the PR from going in, not sure how that plays with the AT though). Additionally, if the scheduled job fails (i.e., not the PR or push to master), then a bot will create an issue with details for follow-up assigning Naser and Luca for now. What's more the workflow can be triggered manually. If an issue exists, the bot will simply append to it by adding more comments. We reserve the notify-file-gh-action for this specific workflow.

To test the automated issue-creation part, we should delete the following file and see when/if this workflow catches the discrepancy.

```xml
<spa_remap_file hgrid="ne120np4">${DIN_LOC_ROOT}/atm/scream/maps/map_ne30np4_to_ne120np4_mono_20220502.nc</spa_remap_file>
```

Accidental example of what the bot will create: https://github.com/E3SM-Project/scream/issues/2683

### DIFF
--- a/.github/workflows/eamxx_default_files.yml
+++ b/.github/workflows/eamxx_default_files.yml
@@ -36,24 +36,12 @@ jobs:
     steps:
       - name: Create issue
         run: |
-          if [[ $CLOSE_PREVIOUS == true ]]; then
-            previous_issue_number=$(gh issue list \
-              --label "$LABELS" \
-              --json number \
-              --jq '.[0].number')
-            if [[ -n $previous_issue_number ]]; then
-              gh issue close "$previous_issue_number"
-              gh issue unpin "$previous_issue_number"
-            fi
-          fi
-          new_issue_url=$(gh issue create \
-            --title "$TITLE" \
-            --assignee "$ASSIGNEES" \
-            --label "$LABELS" \
-            --body "$BODY")
-          if [[ $PINNED == true ]]; then
-            gh issue pin "$new_issue_url"
-          fi
+          previous_issue_number=$(gh issue list \
+          --label "$LABELS" \
+          --json number \
+          --jq '.[0].number')
+          gh issue edit "$previous_issue_number" \
+            --body "$BODY"
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           GH_REPO: ${{ github.repository }}
@@ -61,12 +49,7 @@ jobs:
           ASSIGNEES: mahf708,bartgol
           LABELS: bug,input file,notify-file-gh-action
           BODY: |
-
             Workflow failed! There's likely a missing file specified in the configs!
-
             For more information, please see:
               - Workflow URL: https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}
               - Workflow SHA: ${{ github.sha }}
-    
-          PINNED: false
-          CLOSE_PREVIOUS: false
--- a/.github/workflows/eamxx_default_files.yml
+++ b/.github/workflows/eamxx_default_files.yml
@@ -62,7 +62,7 @@ jobs:
           LABELS: bug
           BODY: |
 
-            [Workflow](${{ github.sha }}) failed! There's likely a missing file specified in the configs!
+            Workflow failed! There's likely a missing file specified in the configs!
 
             For more information, please see:
               - Workflow URL: https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}
--- a/.github/workflows/eamxx_default_files.yml
+++ b/.github/workflows/eamxx_default_files.yml
@@ -40,8 +40,16 @@ jobs:
           --label "$LABELS" \
           --json number \
           --jq '.[0].number')
-          gh issue edit "$previous_issue_number" \
-            --body "$BODY"
+          if [[ -n $previous_issue_number ]]; then
+            gh issue edit "$previous_issue_number" \
+              --body "$BODY"
+          else
+            gh issue create \
+              --title "$TITLE" \
+              --assignee "$ASSIGNEES" \
+              --label "$LABELS" \
+              --body "$BODY"
+          fi
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           GH_REPO: ${{ github.repository }}
--- a/.github/workflows/eamxx_default_files.yml
+++ b/.github/workflows/eamxx_default_files.yml
@@ -30,7 +30,6 @@ jobs:
           python -m unittest tests/eamxx_default_files.py -v
 
   notify-scream-defaults:
-    name: Notify if failure
     needs: scream-defaults
     if: failure() && needs.scream-defaults.outputs.event_name == 'pull_request'
     runs-on: ubuntu-latest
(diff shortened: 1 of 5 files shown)


components/eamxx/cime_config/tests/test_inputdata_namelist_defaults_scream.py
```diff
@@ -0,0 +1,55 @@
...
+        file and extract the files of interest based on the DIN_LOC_ROOT
+        variable. Assign the extracted files to the 'my_files' attribute
+        of the test instance.
+        """
+        os.environ["DIN_LOC_ROOT"] = "https://web.lcrc.anl.gov/public/e3sm/inputdata/"
+
+        tree = ET.parse('namelist_defaults_scream.xml')
+        root = tree.getroot()
+
+        file_of_interest = [
```
bartgol, line 25: We can use the nodes attributes to grab the arrays of files...
```suggestion
files = child.text for child in root.findall(".//") if child.text and "type" in child.attrib.keys() and child.attrib["type"]=="file"
files.append (child.text.split(',') for child in root.findall(".//") if child.text and "type" in child.attrib.keys() and child.attrib["type"]=="array(file)")

files_of_interest = [f for f in files if f.startswith("${DIN_LOC_ROOT}")]
```

rljacob:
I need more context.  Don't really know what you're asking.

components/eamxx/cime_config/tests/test_inputdata_namelist_defaults_scream.py
bartgol: But the last line should only pick files that start with `${DIN_LOC_ROOT}`, just like you did.

.github/workflows/eamxx_default_files.yml
```diff
@@ -0,0 +1,72 @@
...
+        with:
+          python-version: "3.11"
+      - name: Run unit tests
+        working-directory: components/eamxx/cime_config/
+        run: |
+          python -m unittest tests/eamxx_default_files.py -v
+
+  notify-scream-defaults:
+    needs: scream-defaults
+    if: failure() && needs.scream-defaults.outputs.event_name == 'pull_request'
```
bartgol, line 34: Where does "failure()" come from?

bartgol: Also, I think we should run this on master, not on the PR. The PR already has merge blocked, since the checks don't pass. Also, opening an issue b/c a PR branch is missing data seems wrong. OTOH, it is appropriate if master is missing data on the server.

.github/workflows/eamxx_default_files.yml
```diff
@@ -0,0 +1,72 @@
...
+            --body "$BODY")
+          if [[ $PINNED == true ]]; then
+            gh issue pin "$new_issue_url"
+          fi
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GH_REPO: ${{ github.repository }}
+          TITLE: Inputdata server file missing
+          ASSIGNEES: mahf708,bartgol
+          LABELS: bug
```
bartgol, line 62: We also have `input file`, which I think is appropriate here.

.github/workflows/eamxx_default_files.yml
```diff
@@ -0,0 +1,72 @@
...
+          python -m unittest tests/eamxx_default_files.py -v
+
+  notify-scream-defaults:
+    needs: scream-defaults
+    if: failure() && needs.scream-defaults.outputs.event_name == 'pull_request'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Create issue
+        run: |
+          if [[ $CLOSE_PREVIOUS == true ]]; then
```
bartgol, line 39: This seems overly fancy... Why not simply appending to the existing issue?

.github/workflows/eamxx_default_files.yml
```diff
@@ -0,0 +1,72 @@
...
+            previous_issue_number=$(gh issue list \
+              --label "$LABELS" \
```
bartgol, line 41: We should probably add a new label, that is only used by this issue, so that we avoid finding an issue with the same labels that is however not the one that was generated by this action...

.github/workflows/eamxx_default_files.yml
bartgol: I mean, you can still check if the issue is open, and if so, append to it. I just think we don't need to close the old issue and open a new one...

.github/workflows/eamxx_default_files.yml
bartgol: Ok. I was puzzled by the == instead of != typo. Also, I didn't know where `failure()` is defined. I guess it's a built-in function of the action framework.

.github/workflows/eamxx_default_files.yml
```diff
@@ -0,0 +1,55 @@
...
+    if: ${{ failure() && needs.scream-defaults.outputs.event_name != 'pull_request' }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Create issue
+        run: |
+          previous_issue_number=$(gh issue list \
+          --label "$LABELS" \
+          --json number \
+          --jq '.[0].number')
+          gh issue edit "$previous_issue_number" \
```
bartgol: Two things:
- we may also do `gh issue comment`, so we keep track of how the missing list evolved (assuming nobody fixed it in the meantime) rather than overwriting. Not sure if it'd help, but can't hurt...
- we still need ot check if the issue exists, otherwise I guess this may fail? something like
```suggestion
if [ -z "$previous_issue_number" ]; then
  gh issue open <labels, assignees, body...>
else
  gh issue comment <new comment>
fi
```

components/eamxx/cime_config/tests/eamxx_default_files.py
```diff
@@ -0,0 +1,79 @@
...
+
+        for i_file in range(len(self.my_files)):
+            with self.subTest(i_file=i_file):
+                try:
+                    request_return = urllib.request.urlopen(
+                        f"{os.environ['DIN_LOC_ROOT']}{self.my_files[i_file]}"
+                    )
+                    self.assertIsInstance(request_return, http.client.HTTPResponse)
+                except urllib.error.HTTPError:
+                    file_name = f"{os.environ['DIN_LOC_ROOT']}{self.my_files[i_file]}"
```
bartgol, line 60: We should print the file name, so we get it in the action output.

components/eamxx/cime_config/tests/eamxx_default_files.py
```diff
@@ -0,0 +1,79 @@
...
+                    self.assertTrue(
```
bartgol, line 61: We should wait to assert until after the for loop, when all files have been processed, so we can fix all of them.

components/eamxx/cime_config/tests/eamxx_default_files.py
bartgol: Whoops, I missed that.

bartgol (approved):

bartgol:
I think this looks good now. I wonder whether you should "reinstate" some of the missing files, so we can verify that the action works as expected when it runs on master...

PR #2687, opened 2024-01-26
Add WindSpeed diagnostic

bartgol:
It computes the (3d field) magnitude of the "horiz_winds" field. Can probably be generalized to compute magnitude of any field if need be in the future.

Note: the most important use case is to get the wind speed at 100m of elevation. This can be achieved by requesting `wind_speed_at_100m_above_surface` in the output yaml file. Our I/O infrastructure will automatically pipe the WindSpeed diag with the FieldAtHeight one.


AaronDonahue (approved):
Not bad for 10min work ;-)

AaronDonahue (approved):

PR #2689, opened 2024-01-30
enable reading nudging_filename in pattern and bypass vert_interp for nudging

jsbamboo:
Two small features:

1. Add a glob function to enable reading nudging_filename in pattern (idea from Aaron).
2. Use a TIME_DEPENDENT_3D_HYBRID condition to bypass vert_interp and the following fill_mask in nudging.

A unit test has been added for them. Besides, a short run when a glob pattern matches 30 filename strings was passed. 

The glob function (or similar capability) is intended to address the needs of the long-term nudged runs, so that the user does not have to change the nudging_filename in each submission cycle. The TIME_DEPENDENT_3D_HYBRID condition can only be used if the nudging data has the same vertical coordinate as the model (i.e. vertical interpolation for nudging data is done offline). These changes should not affect existing nudging capabilities.

Not sure if these changes will apply to the needs of most users, but I am opening this PR since Aaron said "being able to use a pattern for nudging file names is a useful feature others would want". Please feel free to close/modify it.

### DIFF
--- a/components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
+++ b/components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
@@ -2,6 +2,8 @@
 #include "share/util/scream_universal_constants.hpp"
 #include "share/grid/remap/refining_remapper_p2p.hpp"
 #include "share/grid/remap/do_nothing_remapper.hpp"
+#include <glob.h>
+
 
 namespace scream
 {
@@ -10,7 +12,7 @@ namespace scream
 Nudging::Nudging (const ekat::Comm& comm, const ekat::ParameterList& params)
   : AtmosphereProcess(comm, params)
 {
-  m_datafiles  = m_params.get<std::vector<std::string>>("nudging_filename");
+  m_datafiles  = filename_glob(m_params.get<std::vector<std::string>>("nudging_filename"));
   m_timescale = m_params.get<int>("nudging_timescale",0);
   m_fields_nudge = m_params.get<std::vector<std::string>>("nudging_fields");
   m_use_weights   = m_params.get<bool>("use_nudging_weights",false);
@@ -26,8 +28,11 @@ Nudging::Nudging (const ekat::Comm& comm, const ekat::ParameterList& params)
     m_src_pres_type = STATIC_1D_VERTICAL_PROFILE;
     // Check for a designated source pressure file, default to first nudging data source if not given.
     m_static_vertical_pressure_file = m_params.get<std::string>("source_pressure_file",m_datafiles[0]);
+  } else if (src_pres_type=="TIME_DEPENDENT_3D_HYBRID") {
+    m_src_pres_type = TIME_DEPENDENT_3D_HYBRID;
   } else {
-    EKAT_ERROR_MSG("ERROR! Nudging::parameter_list - unsupported source_pressure_type provided.  Current options are [TIME_DEPENDENT_3D_PROFILE,STATIC_1D_VERTICAL_PROFILE].  Please check");
+    EKAT_ERROR_MSG("ERROR! Nudging::parameter_list - unsupported source_pressure_type provided.  "
+                   "Current options are [TIME_DEPENDENT_3D_PROFILE,STATIC_1D_VERTICAL_PROFILE,TIME_DEPENDENT_3D_HYBRID].  Please check");
   }
   // use nudging weights
   if (m_use_weights) 
@@ -37,6 +42,37 @@ Nudging::Nudging (const ekat::Comm& comm, const ekat::ParameterList& params)
   // 1. if m_timescale is <= 0 we will do direct replacement.
   // 2. if m_fields_nudge is empty or =NONE then we will skip nudging altogether.
 }
+
+// =========================================================================================
+std::vector<std::string> Nudging::filename_glob(const std::vector<std::string>& patterns) {
+  std::vector<std::string> all_files;
+  for (const auto& pattern : patterns) {
+      auto files = globloc(pattern);
+      all_files.insert(all_files.end(), files.begin(), files.end());
+  }
+  return all_files;
+}
+
+std::vector<std::string> Nudging::globloc(const std::string& pattern) {
+  // glob struct resides on the stack
+  glob_t glob_result;
+  memset(&glob_result, 0, sizeof(glob_result));
+
+  int return_value = ::glob(pattern.c_str(), GLOB_TILDE, NULL, &glob_result);
+  if (return_value != 0) {
+    globfree(&glob_result);
+    EKAT_REQUIRE_MSG(return_value == 0, "glob() failed with return value " + std::to_string(return_value));
+  }
+
+  std::vector<std::string> filenames;
+  for (size_t i = 0; i < glob_result.gl_pathc; ++i) {
+      filenames.push_back(std::string(glob_result.gl_pathv[i]));
+  }
+
+  globfree(&glob_result);
+  return filenames;
+}
+
 
 // =========================================================================================
 void Nudging::set_grids(const std::shared_ptr<const GridsManager> grids_manager)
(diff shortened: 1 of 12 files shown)


components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
```diff
@@ -10,7 +12,7 @@ namespace scream
 Nudging::Nudging (const ekat::Comm& comm, const ekat::ParameterList& params)
   : AtmosphereProcess(comm, params)
 {
-  m_datafiles  = m_params.get<std::vector<std::string>>("nudging_filename");
+  m_datafiles  = filename_glob(m_params.get<std::vector<std::string>>("nudging_filename"));
```
bartgol, line 15: This may be personal, but since we use patterns, we should probably change the param name to something like `nudging_filenames_patterns`? The param name, as of now, makes it sound like you must pass actual filenames (and it's not even clear on the fact that 2+ are allowed).

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
```diff
@@ -79,10 +115,16 @@ void Nudging::set_grids(const std::shared_ptr<const GridsManager> grids_manager)
   /* ----------------------- WARNING --------------------------------*/
 
   //Now need to read in the file
-  if (m_src_pres_type == TIME_DEPENDENT_3D_PROFILE) {
-    m_num_src_levs = scorpio::get_dimlen(m_datafiles[0],"lev");
-  } else {
+  if (m_src_pres_type == STATIC_1D_VERTICAL_PROFILE) {
     m_num_src_levs = scorpio::get_dimlen(m_static_vertical_pressure_file,"lev");
+  } else {
+    m_num_src_levs = scorpio::get_dimlen(m_datafiles[0],"lev");
```
bartgol, line 121: Is it worth checking that all files have the same "lev" dimension?

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
```diff
@@ -461,43 +505,45 @@ void Nudging::run_impl (const double dt)
...
-	const auto iidx  = kk % mPack::n;
-        // Check if this index is masked
-	if (!int_mask_view_1d(ipack)[iidx]) {
-	  fill_value = tmp_state_view_1d(ipack)[iidx];
-	} else {
-	  tmp_state_view_1d(ipack)[iidx] = fill_value;
-	}
-      }
-    });
+    if (m_src_pres_type != TIME_DEPENDENT_3D_HYBRID) {
```
bartgol, line 508: This is fine, but just to share a different perspective (for future reference): you could consider early return instead of a long if statement. It gets a bit hard to track nesting when there are several layers. Imho, something like this is clearer (in addition of only causing a 4 lines diff instead of ~40).
```c++
  if (m_src_pres_type == TIME_DEPENDENT_3D_HYBRID) {
    // Nothing left to do
    return;
  }
  
  const int num_levs = m_num_levs;
  ...
```
Again, what you did is fine. Just sharing this, in case it looks better to you.

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.hpp
```diff
@@ -63,6 +64,12 @@ class Nudging : public AtmosphereProcess
   // The name of the subcomponent
   std::string name () const { return "Nudging"; }
 
+  // Find the full filename list from patterns
+  std::vector<std::string> filename_glob(const std::vector<std::string>& patterns);
```
bartgol, line 68: These two functions do not change the state of the class. In fact, they don't even use any internals of the class. You could consider moving them inside `share/util/scream_util.hpp`. You could also simply declare them as local functions inside the cpp file, rather than methods. They don't conceptually belong to the public interface of the class...

bartgol (commented):
I have some thoughts on a couple of things. Feel free to push back on them.

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
jsbamboo: Thank you for telling me this! I don't want unnecessary diff either

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
jsbamboo: I feel just checking the first file should be enough - I would expect it's a consensus for nudging users that nudging data should be prepared in the same format because they should come from the same set of source data (I can't currently imagine a situation where different formats of nudging data are used interchangeably). Moreover, the TIME_DEPENDENT_3D_HYBRID condition is even stricter (hybrid coordinate same as the model coordinate)

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
jsbamboo: Thanks Luca, I just didn't intend to change any existing user-level parameters... the user should know that 2+ strings are allowed for nudging_filename as shown in the array type specified in its definition? How about just adding a doc to indicate that nudging_filename can accept patterns in namelist_defaults_scream.xml? something like "one or more strings to indicate the actual filenames or patterns"?

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.hpp
jsbamboo: Thanks again for teaching me the basics.. will try to move them to scream_util.hpp

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
bartgol: I agree that the user can see that they can pass more than 1 string, but the more pressing thing is that it's not very clear that these strings can be patterns, rather than actual file names.

I don't think it's a bad idea to change a parameter name and make it more clear. Imho, it will pay off in the long run.

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
bartgol: What if the user accidentally uses a wrong pattern for, say, the 2nd file? The run will start, proceed for a while, and then crash. If a restart file was written sufficiently close, that may not be _too_ bad, but it would be certainly more convenient to find this out early during init time, no? And btw, you could probably do this check on the file inside the Nudging constructor, so that the error happens even earlier in the init sequence.

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.hpp
bartgol: Tbc, what you have is fine, and it works. It's just a matter of concept/design. Keeping large code bases tidy can help maintenance (especially when the maintainer is not the person that wrote the code).

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
jsbamboo: i see i'll change all nudging_filename to nudging_filenames_patterns

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
jsbamboo: thanks! please see if the changes make sense

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.hpp
jsbamboo: i like your suggestions, in fact I thought it would be more appropriate to put it in a more generic interface (but didn't know know where to put it at the time), since the glob function is a very generic function.. please check if the changes make sense

bartgol (approved):

bartgol:
I'm afraid this PR lost the race for integration, and is now in conflict with master. It needs to be rebased on master, and there may be some difficulty, since the nudging interface did undergo a non-trivial restructuring.

jsbamboo:
@bartgol let's close this one.

if you could help add the glob function then? or do you want me to open a new PR?

bartgol:
> @bartgol let's close this one.
> 
> if you could help add the glob function then? or do you want me to open a new PR?

You can open the new PR from a new branch. Feel free to tag me as reviewer when ready.

jsbamboo:
okay! i tried to use rebase while many changes directly disappeared. close this and i will open a new PR

PR #2691, opened 2024-01-30
Sort gids of ov coarse grid in horiz remappers, to ensure same results regardless of PES count

### LINKED ISSUE #2690: PEM test failure debug
Opening an issue to track the debugging of the PEM failure, so I know where I left off when I come back tomorrow.

Here's a summary of stuff I found out

- The fail is reproducible on mappy.
- The fail does not happen with ne4, but happens with ne30.
- The fail seem to develop at the first time step
- The first diff appears right after spa, hinting at issue

Next checks:
 - [x] verify with an Ln1 test that the test fails. It does.
 - [x] check whether the test fails without the `spa_remap` testmod. It doesn't.

bartgol:
I ran `PEM_Ln1.ne30pg2_ne30pg2.F2010-SCREAMv1.mappy_gnu9.scream-spa_remap` with master and with this branch. On master, the test fails, on this branch it passes. I am moderately confident this may fix our PEM failures in the nightlies.

Fixes #2690

### DIFF
--- a/components/eamxx/src/share/grid/remap/horiz_interp_remapper_base.cpp
+++ b/components/eamxx/src/share/grid/remap/horiz_interp_remapper_base.cpp
@@ -329,6 +329,10 @@ create_coarse_grids (const std::vector<Triplet>& triplets)
   for (const auto& it : ov_gid2lid) {
     ov_coarse_gids_h[it.second] = it.first;
   }
+  auto beg = ov_coarse_gids_h.data();
+  auto end = beg+ov_coarse_gids_h.size();
+  std::sort(beg,end);
+
   ov_coarse_grid->get_dofs_gids().sync_to_dev();
   m_ov_coarse_grid = ov_coarse_grid;
 


ambrad:
What happens if GIDs get fragmented, e.g.,
* run 1 is fully serial: GIDs [0, 1, 2, 3, 4]
* run 2 has two ranks: GIDs [0, 2, 3], [1, 4].

Just to rule out the obvious: This isn't a case of doing a matvec with accumulation in an output row being done over multiple ranks, is it?

bartgol:
> What happens if GIDs get fragmented, e.g.,
> 
> * run 1 is fully serial: GIDs [0, 1, 2, 3, 4]
> * run 2 has two ranks: GIDs [0, 2, 3], [1, 4].
> 
> Just to rule out the obvious: This isn't a case of doing a matvec with accumulation in an output row being done over multiple ranks, is it?

I think we need some clarity. The ov coarse grid is created by adding all coarse points that affect (via the map sparse matrix) any of the dofs in the fine grid. The fine grid is the model grid, which is uniquely owned across ranks (no duplicates). So doesn't really matter how the fine grid is partitioned.

This commit ensures that the dofs in the overlapped coarse grid are ordered in the same way, regardless of how they were gathered. The issue stemmed (I think) from how we were gathering these ov coarse gids:
 - we were reading in the triplets of the sparse matrix we need
 - we were adding the triplet's col GID to the ov coarse grid gids list, in the order in which we parsed the triplets.
 
 This fix makes it so that regardless of what triplets we need on each rank, the sets of GIDs affeting each of this rank's rows appear in the same order in the ov coarse grid. Thus, during the sparse mat-vec, we accumulate contributions _in the same order_. Also, we really don't care how the coarse grid is generated from the ov coarse grid (we use `AbstractGrid::get_unique_gids`, which assings shared dofs to the lowest rank, I think). That's b/c the coarse grid is only responsible for reading in data from file. What matters, is the order of dofs in the ov coarse grid.
 
Sorry if my explanation is not 100% rigorous, perhaps a bit hand wavy here and there. I hope the idea makes sense though.

ambrad:
> The matvec product for the refine case is done after the MPI comm: we import the src field, then perform a fully local computation of each row.

Ok, great. So then this was an issue of having all the source data available for a dot product in one rank but then doing the local accumulation in a PE-layout-dependent order?

bartgol:
> > The matvec product for the refine case is done after the MPI comm: we import the src field, then perform a fully local computation of each row.
> 
> Ok, great. So then this was an issue of having all the source data available for a dot product in one rank but then doing the local accumulation in a PE-layout-dependent order?

Basically, yes. We were ordering the content of the overlapped map in a PES-dependent way, which caused accumulation order to vary with NTASKS.

ambrad (approved):

PR #2696, opened 2024-01-31
Add support for tensor fields in horizontal interp remappers

bartgol:
Currently, we were allowing 2d/3d scalar/vector fields only. This PR add tensor fields (of rank 2). Here, the rank of the tensor field is what is left in the layout after removing the COL and, possibly, the LEV/ILEV dims.

While a "tensor" layout, from the FieldLayout and AbstractGrid point of view, can have an arbitrary number of dimensions, I only added support for rank2 tensors for the remap, since I think this is all that is needed here.

I added testing for the CoarseningRemapper for 2d/3d tensor fields. I did not add any testing for RefiningRemapperP2P, even though I added support for 2d/3d tensor fields in it. If any parametrization needs it (usually this remapper is just for reading in data from a coarser grid dataset), I can add testing to ensure things works as expected.

### DIFF
--- a/components/eamxx/src/share/field/field_layout.cpp
+++ b/components/eamxx/src/share/field/field_layout.cpp
@@ -166,17 +166,17 @@ LayoutType get_layout_type (const std::vector<FieldTag>& field_tags) {
     std::vector<FieldTag> lev_tags = {LEV,ILEV};
     return ekat::contains(lev_tags,t);
   };
-  auto is_vec_tag = [](const FieldTag t) {
-    std::vector<FieldTag> vec_tags = {CMP,NGAS,SWBND,LWBND,SWGPT,ISCCPTAU,ISCCPPRS};
-    return ekat::contains(vec_tags,t);
+  auto is_cmp_tag = [](const FieldTag t) {
+    std::vector<FieldTag> cmp_tags = {CMP,NGAS,SWBND,LWBND,SWGPT,ISCCPTAU,ISCCPPRS};
+    return ekat::contains(cmp_tags,t);
   };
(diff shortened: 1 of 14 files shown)


components/eamxx/src/share/field/field_layout.cpp
```diff
@@ -126,36 +166,38 @@ LayoutType get_layout_type (const std::vector<FieldTag>& field_tags) {
     std::vector<FieldTag> lev_tags = {LEV,ILEV};
     return ekat::contains(lev_tags,t);
   };
-  auto is_vec_tag = [](const FieldTag t) {
-    std::vector<FieldTag> vec_tags = {CMP,NGAS,SWBND,LWBND,SWGPT,ISCCPTAU,ISCCPPRS};
-    return ekat::contains(vec_tags,t);
+  auto is_cmp_tag = [](const FieldTag t) {
```
AaronDonahue, line 169: Does this mean we are relying on Tensor layouts to be like "DYNAMIC" above, i.e. we don't trigger a `return` statement from the conditional?  If so we might want to change the comment to include "A Dynamics or Tensor layout"

AaronDonahue (approved):
Just one comment about comments.  Otherwise looks good to me.

components/eamxx/src/share/field/field_layout.cpp
bartgol: No, it can be a dynamics layout or a physics one. The if/else above takes care of eliminating EL/COL/GP tags, so that we are left with just vector/tensor tags and possibly LEV/ILEV/TL.

PR #2698, opened 2024-01-31
atmchange: enforce locked items

jgfouca:
It will now be an error to try to change a locked parameter.

I added some unit tests for this and also verified that `Frequency` is no longer atmchangeable due to it being the child of a locked element:

```
      <output_control locked="true">
        <frequency_units>${REST_OPTION}</frequency_units>
      </output_control>
```


bartgol (approved):

bartgol:
As a follow up to this PR, we could modify the script `eamxx-params-docs-autogen`, to either completely skip params that are not configurable (i.e., locked), or report that they are not atmchange-able.

You don't have to do it in this PR (but of course can, if you want).

jgfouca:
@bartgol , yeah, it looks like it was basically already done. I just cleaned up a couple things.

bartgol (approved):

PR #2700, opened 2024-02-01
Update EKAT submodule

jgfouca:
to eeaac1a32de15c25b0fdb895b9872ff31702ff7b

Changes:
1) Update YAML-cpp submodule to remove non-portable warning flags
2) Fix GetMpiDistributionName for Cray mpi compilers
3) Change default constructor of Pack to init to 0
4) Use check_symbol_exists instead of try_compile to detect MPI distro name
5) Fix GetMpiDistributionName
6) Allow forcing build of tpls from internal submodules
7) Add native support for WILL_FAIL in EkatCreateUnitTest
8) Ensure that yaml is always linked using the scoped target name
9) Do not use new find_package option with old cmake
10) [SYCL] Minor cleanup for IntelLLVM cmake
11) Some fixes that allow to build with raw compilers instead of mpi wrap…
12) Error out if the found kokkos installation is not compatible with the requested options
13) Add missing quotes for FIXTURES to cmake function create unit test.

To sum up, it looks like nearly all the changes are build/cmake related. This is probably safe to merge IMO.

### DIFF
--- a/components/eamxx/src/physics/p3/tests/CMakeLists.txt
+++ b/components/eamxx/src/physics/p3/tests/CMakeLists.txt
@@ -40,9 +40,9 @@ set(P3_TESTS_SRCS
     ) # P3_TESTS_SRCS
 
 if (SCREAM_DEBUG AND NOT SCREAM_TEST_SIZE STREQUAL "SHORT")
-  set (FORCE_RUN_DIFF_FAILS TRUE)
+  set (FORCE_RUN_DIFF_FAILS "WILL_FAIL")
 else ()
-  set (FORCE_RUN_DIFF_FAILS FALSE)
+  set (FORCE_RUN_DIFF_FAILS "")
 endif()
 
 # NOTE: tests inside this if statement won't be built in a baselines-only build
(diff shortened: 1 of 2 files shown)


bartgol (approved):
I agree, these should be safe to merge. If, by chance, we break the build on some machine, we should be able to ninja fix it. It's not like causing hard-to-find non bfb diffs...

bartgol:
Ah, we need to update the tests that should fail, since WILL_FAIL is now an ecut option.

jgfouca:
@bartgol , fixed. Also fixed an issue with cprnc injecting its ctests into ours.

bartgol:
> @bartgol , fixed. Also fixed an issue with cprnc injecting its ctests into ours.

Sorry, I just saw this. Do you want to still do E3SM-Project/ekat#321? Or should I take back the change?

jgfouca:
@bartgol , I actually like `WILL_FAIL` better than `WILL_FAIL TRUE`. I would prefer you close that PR.

bartgol:
> @bartgol , I actually like `WILL_FAIL` better than `WILL_FAIL TRUE`. I would prefer you close that PR.

Ok. I will change it back. But that PR has another fix, which would be nice to incorporate. I will close and reopen just with the fix.

bartgol (approved):

PR #2703, opened 2024-02-05
Fix issue with small_kernels testmod

ndkeen:
Only impacts testing with small_kernels testmod.
Remove lines from shell command specific to chrysalis that are no longer needed (and caused error on at least pm-cpu).

[bfb]


ambrad (approved):

ndkeen:
Tested with `ERS_Ln90.ne30pg2_ne30pg2.F2010-SCREAMv1.pm-gpu_gnugpu.scream-small_kernels`

PR #2707, opened 2024-02-07
ensure that lines are ascii in scream defaults xml

mahf708:
In a future iteration, I could add more sophisticated testing. For now, this line-by-line check will catch obvious issues. We could ensure individual entries are appropriately formatted later (e.g., doc must be string, type must be some sort of thing from a list of approved types, etc.)


jgfouca (approved):
Thanks!

bartgol (approved):

PR #2710, opened 2024-02-08
Some DP cleanup to prepare for DP forcing port PR

tcclevenger:
Separated some cleanup work before submitting a PR for porting the forcing functions to make things easier to read. 

Changes:
- Rename `m_intensive_observation_period` -> `m_iop`
- Changed back to `atmosphere_process.hpp` storing `m_iop` pointer. Both surface coupling and dynamics processes will need access to the pointer, might as well just store it in the base class than both processes needing functions to set and get iop pointer.
- Need to check for a 3d forcing IOP file variable (does not affect the current test case)
- Allocate IOP file data using a pack size which can be nontrivial
- Change test DT to match the v0 case I'm testing against.
- phys grid was not syncing geo data to device, so the random perturbations were never happening for Cuda testing
- `eamxx/src/doubly-periodic` is an unused directory, removing.


components/eamxx/src/share/atm_process/atmosphere_process.hpp
```diff
@@ -276,6 +280,13 @@ class AtmosphereProcess : public ekat::enable_shared_from_this<AtmosphereProcess
   // For BFB tracking in production simulations.
   void print_fast_global_state_hash(const std::string& label) const;
 
+  // Set IOP object
+  virtual void set_intensive_observation_period(const iop_ptr& iop) {
+    m_iop = iop;
+  }
+
+  const iop_ptr& get_intensive_observation_perdiod() { return m_iop; }
```
bartgol, line 288: I would just make `m_iop` protected, so that derived classes can access it without the lengthy `get_intensive_operation_period()` call. After all, if they can set and get, they should just have access to the member.

components/eamxx/src/control/intensive_observation_period.cpp
```diff
@@ -191,6 +191,10 @@ initialize_iop_file(const util::TimeStamp& run_t0,
       EKAT_REQUIRE_MSG(field_rank <= 1,
                        "Error! Unexpected field rank "+std::to_string(field_rank)+" for iop file fields.\n");
       Field field(fid);
+      if (fl.has_tag(FieldTag::LevelMidPoint) or fl.has_tag(FieldTag::LevelInterface)) {
+        // Request packsize allocation for level layout
+        field.get_header().get_alloc_properties().request_allocation(Pack::n);
```
bartgol, line 196: It looks like you never really use `Pack`. The only time IOP uses packs, is with the "fake" `Pack1d`, just for the sake of being able to call the LinInterp interfaces (we should allow calling with a scalar view, btw).

If that's the case, then you should prob remove all the request (here and below) for a n-sized pack allocation, since IOP doesn't need it.

bartgol (approved):
Looks good. I have a couple of comments, but not necessarily road-blocking.

components/eamxx/src/control/intensive_observation_period.cpp
tcclevenger: We will use pack size n in the next PR.

And yes, scalar would be nice for lininterp. I'll add that to my list since I have a case here I can test.

bartgol (approved):

PR #2714, opened 2024-02-13
enable reading nudging_filename in pattern and bypass vert_interp for nudging (v1)

jsbamboo:
Redo #2689 based on the updated master (after the reconstructed nudging interface #2701 was merged into the master)

==========

Two small features:

Add a glob function to enable reading nudging_filename in pattern (idea from Aaron).
Use a TIME_DEPENDENT_3D_HYBRID condition to bypass copy_and_pad and vert_interp in nudging.
A unit test has been added for them. Besides, a short run when a glob pattern matches 30 filename strings was passed.

The glob function (or similar capability) is intended to address the needs of the long-term nudged runs, so that the user does not have to change the nudging_filename in each submission cycle. The TIME_DEPENDENT_3D_HYBRID condition can only be used if the nudging data has the same vertical coordinate as the model (i.e. vertical interpolation for nudging data is done offline). These changes should not affect existing nudging capabilities.

### DIFF
--- a/components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
+++ b/components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
@@ -33,8 +33,17 @@ Nudging::Nudging (const ekat::Comm& comm, const ekat::ParameterList& params)
     m_src_pres_type = STATIC_1D_VERTICAL_PROFILE;
     // Check for a designated source pressure file, default to first nudging data source if not given.
     m_static_vertical_pressure_file = m_params.get<std::string>("source_pressure_file",m_datafiles[0]);
+  } else if (src_pres_type=="TIME_DEPENDENT_3D_HYBRID") {
+    m_src_pres_type = TIME_DEPENDENT_3D_HYBRID;
+    int first_file_levs = scorpio::get_dimlen(m_datafiles[0],"lev");
+    for (const auto& file : m_datafiles) {
+        int current_file_levs = scorpio::get_dimlen(file,"lev");
+        EKAT_REQUIRE_MSG(current_file_levs == first_file_levs,
+                         "Error! Inconsistent 'lev' dimension found in nudging data files.");
+    }
   } else {
-    EKAT_ERROR_MSG("ERROR! Nudging::parameter_list - unsupported source_pressure_type provided.  Current options are [TIME_DEPENDENT_3D_PROFILE,STATIC_1D_VERTICAL_PROFILE].  Please check");
+    EKAT_ERROR_MSG("ERROR! Nudging::parameter_list - unsupported source_pressure_type provided.  "
+                   "Current options are [TIME_DEPENDENT_3D_PROFILE,STATIC_1D_VERTICAL_PROFILE,TIME_DEPENDENT_3D_HYBRID].  Please check");
   }
   // use nudging weights
   if (m_use_weights) 
@@ -86,8 +95,14 @@ void Nudging::set_grids(const std::shared_ptr<const GridsManager> grids_manager)
   /* ----------------------- WARNING --------------------------------*/
 
   //Now need to read in the file
-  if (m_src_pres_type == TIME_DEPENDENT_3D_PROFILE) {
+  if (m_src_pres_type == TIME_DEPENDENT_3D_PROFILE || m_src_pres_type == TIME_DEPENDENT_3D_HYBRID) {
     m_num_src_levs = scorpio::get_dimlen(m_datafiles[0],"lev");
+    // HYBRID needs the same vertical grid as model for now to bypass vert_interp
+    if (m_src_pres_type == TIME_DEPENDENT_3D_HYBRID) {
+      EKAT_REQUIRE_MSG(m_num_src_levs == m_num_levs,
+                     "Error! TIME_DEPENDENT_3D_HYBRID requires the vertical level to be "
+                     << " the same as model vertical level ");
+    }
   } else {
     m_num_src_levs = scorpio::get_dimlen(m_static_vertical_pressure_file,"lev");
   }
@@ -381,6 +396,19 @@ void Nudging::run_impl (const double dt)
 
   // Perform horizontal remap (if needed)
   m_horiz_remapper->remap(true);
+
+  // bypass copy_and_pad and vert_interp for hybrid coordinate:
+  if (m_src_pres_type == TIME_DEPENDENT_3D_HYBRID) {
+    for (const auto& name : m_fields_nudge) {
+      auto tmp_state_field = get_helper_field(name+"_tmp");
+
+      if (m_timescale > 0) {
+        auto atm_state_field = get_field_out_wrap(name);
+        apply_tendency(atm_state_field,tmp_state_field,dt);
+      }
+    }
+    return;
+  }
 
   // Copy remapper tgt fields into padded views, to allow extrapolation at top/bot,
   // then call remapping routines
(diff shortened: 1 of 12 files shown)


components/eamxx/tests/coupled/physics_only/shoc_p3_nudging/input_nudging.yaml
```diff
@@ -14,7 +14,7 @@ atmosphere_processes:
   p3:
     max_total_ni: 720.0e3
   nudging:
-    nudging_filename: [shoc_p3_source_data_${POSTFIX}.INSTANT.nsteps_x${NUM_STEPS}.${RUN_T0}.nc]
+    nudging_filenames_patterns: [shoc_p3_source_data_${POSTFIX}.INSTANT.nsteps_x${NUM_STEPS}.${RUN_T0}.nc]
```
bartgol, line 17: One thought, feel free to implement or not: the "patterns" feature is basically not used here, cause there is just one file. What if we generated nudging data with 1 time slice per file? That way, we could specify a pattern, and see if the nudging interface picks up all the files correctly..

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
```diff
@@ -32,8 +33,17 @@ Nudging::Nudging (const ekat::Comm& comm, const ekat::ParameterList& params)
     m_src_pres_type = STATIC_1D_VERTICAL_PROFILE;
     // Check for a designated source pressure file, default to first nudging data source if not given.
     m_static_vertical_pressure_file = m_params.get<std::string>("source_pressure_file",m_datafiles[0]);
+  } else if (src_pres_type=="TIME_DEPENDENT_3D_HYBRID") {
+    m_src_pres_type = TIME_DEPENDENT_3D_HYBRID;
+    int first_file_levs = scorpio::get_dimlen(m_datafiles[0],"lev");
```
bartgol, line 38: I think this check should be done regardless of `src_pres_type`, no? If there are 2+ files, we should always verify that their dimensions match.

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
```diff
@@ -381,6 +397,19 @@ void Nudging::run_impl (const double dt)
   // Perform horizontal remap (if needed)
   m_horiz_remapper->remap(true);
 
+  // bypass copy_and_pad and vert_interp for hybrid coordinate:
```
bartgol, line 400: I'm confused. I thought the hybrid approach was based on reconstructing p_mid_src from ps/hyam/hybm (after horiz interp), and then proceed to do vert interp. That's also what SPA does. I don't see ps/hyam/hybm used here, while in the hpp I saw the comment
```cpp
  // hybrid data includes hyai(ilev),hybi(ilev),hyam(lev),hybm(lev),ilev(ilev),lev(lev),P0,PS(time, ncol)
```
To me, it seems like this hybrid approach is simply an horiz interpolation, assuming the data levels are _the same_ as the model levels.

This, btw, brings back a discussion that @AaronDonahue started: horiz interpolation should happen on pressure levels, not model levels.

bartgol (commented):
The globbing part is fine. I'm a bit puzzled by the hybrid approach. In particular, I'm wondering if we're missing a piece, where we reconstruct p_mid_src from ps/hyam/hybm and then do vinterp.

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
jsbamboo: you mean for all time-varying src_pres_type (not including STATIC_1D_VERTICAL_PROFILE) right?

components/eamxx/tests/coupled/physics_only/shoc_p3_nudging/input_nudging.yaml
jsbamboo: i dont quite understand this problem.. nudging_filenames_patterns can read {a single filename, a single pattern name, a list of filename, a list of patterns}, so in reading a single filename in this case is just a subset of the ``patterns'' feature ..? (or are you discussing nomenclature?)

oh, the test i added is another one, this is the default input for other nudging tests. 
please see `components/eamxx/tests/coupled/physics_only/shoc_p3_nudging/input_nudging_glob.yaml`

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
jsbamboo: or should i change the name for this condition? HYBRID does make people think it's need reconstructing p_mid_src from the hybrid coordinate...

bartgol: What if, instead of specifying a new value for the src pres type, we use a _separate_ input option. After all, the src pressure is still a time dependent 3d profile. What about adding the following?
```
  <skip_vert_interpolation type="logical" doc="blahblah">false</skip_vertical_interpolation>
```
Clearly, if this is set to `true`, we need to check that `m_num_levs==m_src_num_levs`, otherwise we _can't_ skip vert interpolation. We could also check that this option is not set to `true` if the static 1d profile is used (makes no sense to not interpolate if src press is uniform and constant).

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
bartgol: Well, also for static, no? What if the user specifies two files, and accidentally they are from different runs?

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
jsbamboo: thanks for the suggestion! it makes sense to me. I will change them

bartgol (approved):

components/eamxx/tests/coupled/physics_only/shoc_p3_nudging/input_nudging.yaml
bartgol: I mean, I like the motto "if it's not tested, it doesn't exist". We can certainly to a follow up PR, but I'd like to test the case where we use 2+ files in our nightlies at some point...

components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp
```diff
@@ -381,6 +396,19 @@ void Nudging::run_impl (const double dt)
   // Perform horizontal remap (if needed)
   m_horiz_remapper->remap(true);
 
+  // bypass copy_and_pad and vert_interp for skip_vert_interpolation:
+  if (m_skip_vert_interpolation) {
+    for (const auto& name : m_fields_nudge) {
+      auto tmp_state_field = get_helper_field(name+"_tmp");
```
bartgol, line 402: The field with `_tmp` suffix is not created during init if `m_skip_vert_interpolation=true`...

jsbamboo: the !m_skip_vert_interpolation is only for *p_mid* fields right (updated L281-306)

jsbamboo: the tmp fields for others (T_mid, qv, horiz_winds) are set here for all cases:
```
    // Second copy of the field: after horiz interp (alias "ext" if no remap)
    Field field_tmp;
    if (m_refine_remap) {
      field_tmp = create_helper_field(name_tmp, layout_tmp, grid_tmp->name());
    } else {
      field_tmp = field_ext.alias(name_tmp);
      m_helper_fields[name_tmp] = field_tmp;
    }
```

bartgol: Ah right.

bartgol (approved):

PR #2715, opened 2024-02-14
Port advance_iop functions

tcclevenger:
Port `apply_iop_forcing()` (partially) which is the main function called by dynamics post process. This function is ported fully for the current test case (still need to port `iop_domain_relaxation()` for other cases).

**Other changes:**
- Consolidate `Pack` typedefs in Homme-AP interface
- Add parameter `iop_dosubsidence=true` for this case

**Testing:**
I ran F90 case and C++ unit test with hacked in input directly before `apply_iop_forcing()` and the output matches up to machine precision on CPU (serial and openmp) and GPU.

**Next PRs up:**
- Correctly load SPA data, add SPA back to test
- Add CIME case to lowres test suite
- Port `iop_domain_relaxation()` and add test cases which use that code

### DIFF
--- a/components/eamxx/src/dynamics/homme/eamxx_homme_iop.cpp
+++ b/components/eamxx/src/dynamics/homme/eamxx_homme_iop.cpp
@@ -112,37 +112,43 @@ advance_iop_subsidence(const KT::MemberType& team,
     Pack omega_int_k, omega_int_kp1;
     ekat::index_and_shift<1>(s_omega_int, range_pack, omega_int_k, omega_int_kp1);
 
-    auto at_top = range_pack==0;
-    auto at_bot = range_pack==nlevs-1;
+    const auto at_top = range_pack==0;
+    const auto at_bot = range_pack==nlevs-1;
+    const auto at_mid = not at_top and not at_bot;
 
-    auto fac = dt/(2*pdel(k));
+    const auto fac = (dt/2)/pdel(k);
 
     // Update u
-    u(k).set(at_top,                    u(k) - fac*omega_int_kp1*delta_u_k);
-    u(k).set(at_bot,                    u(k) - fac*omega_int_k*delta_u_km1);
-    u(k).set(not at_top and not at_bot, u(k) - fac*(omega_int_kp1*delta_u_k + omega_int_k*delta_u_km1));
+    auto& u_k = u(k);
+    u_k.set(at_top, u_k - fac*omega_int_kp1*delta_u_k);
+    u_k.set(at_bot, u_k - fac*omega_int_k*delta_u_km1);
+    u_k.set(at_mid, u_k - fac*(omega_int_kp1*delta_u_k + omega_int_k*delta_u_km1));
 
     // Update v
-    v(k).set(at_top,                    v(k) - fac*omega_int_kp1*delta_v_k);
-    v(k).set(at_bot,                    v(k) - fac*omega_int_k*delta_v_km1);
-    v(k).set(not at_top and not at_bot, v(k) - fac*(omega_int_kp1*delta_v_k + omega_int_k*delta_v_km1));
+    auto& v_k = v(k);
+    v_k.set(at_top, v_k - fac*omega_int_kp1*delta_v_k);
+    v_k.set(at_bot, v_k - fac*omega_int_k*delta_v_km1);
+    v_k.set(at_mid, v_k - fac*(omega_int_kp1*delta_v_k + omega_int_k*delta_v_km1));
+
+    // Before updating T, first scale using thermal
+    // expansion term due to LS vertical advection
+    auto& T_k = T(k);
+    T_k *= 1 + (dt*Rair/Cpair)*omega(k)/pmid(k);
 
     // Update T
-    const auto T_k = T(k);
-    T(k).set(at_top,                    T_k - fac*omega_int_kp1*delta_T_k);
-    T(k).set(at_bot,                    T_k - fac*omega_int_k*delta_T_km1);
-    T(k).set(not at_top and not at_bot, T_k - fac*(omega_int_kp1*delta_T_k + omega_int_k*delta_T_km1));
-    // Add thermal expansion term due to LS vertical advection
-    T(k) += dt*omega(k)*T_k*Rair/(Cpair*pmid(k));;
+    T_k.set(at_top, T_k - fac*omega_int_kp1*delta_T_k);
+    T_k.set(at_bot, T_k - fac*omega_int_k*delta_T_km1);
+    T_k.set(at_mid, T_k - fac*(omega_int_kp1*delta_T_k + omega_int_k*delta_T_km1));
 
     // Update Q
+    Pack delta_tracer_k, delta_tracer_km1;
     for (int iq=0; iq<n_q_tracers; ++iq) {
       auto s_delta_tracer = Kokkos::subview(s_delta_Q, iq, Kokkos::ALL());
-      Pack delta_tracer_k, delta_tracer_km1;
       ekat::index_and_shift<-1>(s_delta_tracer, range_pack_m1, delta_tracer_k, delta_tracer_km1);
-      Q(iq, k).set(at_top,                    Q(iq, k) - fac*omega_int_kp1*delta_tracer_k);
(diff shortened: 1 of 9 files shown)


tcclevenger:
@bogensch I'm going to let Luca look this over for C++ related changes, then I'll request a review from you with most likely some impl questions.

components/eamxx/src/dynamics/homme/eamxx_homme_iop.cpp
```diff
@@ -0,0 +1,408 @@
...
+    v(k).set(at_bot,                    v(k) - fac*omega_int_k*delta_v_km1);
+    v(k).set(not at_top and not at_bot, v(k) - fac*(omega_int_kp1*delta_v_k + omega_int_k*delta_v_km1));
+
+    // Update T
+    const auto T_k = T(k);
+    T(k).set(at_top,                    T_k - fac*omega_int_kp1*delta_T_k);
+    T(k).set(at_bot,                    T_k - fac*omega_int_k*delta_T_km1);
+    T(k).set(not at_top and not at_bot, T_k - fac*(omega_int_kp1*delta_T_k + omega_int_k*delta_T_km1));
+    // Add thermal expansion term due to LS vertical advection
+    T(k) += dt*omega(k)*T_k*Rair/(Cpair*pmid(k));;
```
bartgol, line 136: Again, it's best to do all the scalar ops at once, since they are cheaper than vector ops. Any chance we can turn this into this?
```suggestion
  T(k) += (dt*Rair/Cpair) * omega(k)*T_k/pmid(k) ;
```
If you need to keep BFB with F90, for a <5 lines snippet of code, it's ok to do something like
```c++
#ifdef EAMXX_BFB_TESTING
  // impl with same order of ops as in f90
#else
  // impl with better order of ops
#endif
```
Note: I don't recall if we have that macro, but we should have something equivalent.

bartgol: Also, in the update of `T(k)`, is it intended to use `T_k` rather than `T(k)` on the rhs?

components/eamxx/src/dynamics/homme/eamxx_homme_iop.cpp
```diff
@@ -0,0 +1,408 @@
...
+    Pack omega_int_k, omega_int_kp1;
+    ekat::index_and_shift<1>(s_omega_int, range_pack, omega_int_k, omega_int_kp1);
+
+    auto at_top = range_pack==0;
+    auto at_bot = range_pack==nlevs-1;
+
+    auto fac = dt/(2*pdel(k));
+
+    // Update u
+    u(k).set(at_top,                    u(k) - fac*omega_int_kp1*delta_u_k);
```
bartgol, line 121: One general thing for this for loop. I would consider grabbing a ref out of the view, so you don't index the view 3+ times. Something like
```suggestion
auto& uk = u(k);
uk.set(...);
uk.set(...);
uk.set(...);
```

components/eamxx/src/dynamics/homme/eamxx_homme_iop.cpp
bartgol: Finally, from the Tk update formula, putting the pieces together, it seems we do
```
Tk = Tk - fac*blah + dt*Tk*blah2
```
Maybe we could then do
```
auto& Tk = T(k);
Tk *= (1+dt*blah2);
ekat_masked_loop (at_top,s)
  Tk[s] -= fac*blah_top;
ekat_masked_loop (at_bot,s)
  Tk[s] -= fac*blah_bot;
ekat_masked_loop (at_mid,s)
  Tk[s] -= fac*blah_mid;
```

bartgol: If we impl-ed an `update` method for `Pack` similar to `Pack::set`, we could do
```
auto& Tk = T(k);
Tk *= (1+dt*blah2);
Tk.update<Subtract>(not at_top,fac*blah_not_top);
Tk.udpate<Subtract>(not at_bot,fac*blah_not_bot);
```

components/eamxx/src/dynamics/homme/eamxx_homme_iop.cpp
```diff
@@ -0,0 +1,408 @@
...
+  constexpr int NGP   = HOMMEXX_NP;
+  constexpr int NLEV  = HOMMEXX_NUM_LEV;
+  constexpr int NLEVI = HOMMEXX_NUM_LEV_P;
+  const auto nelem    = m_dyn_grid->get_num_local_dofs()/(NGP*NGP);
+  const auto total_levels = m_dyn_grid->get_num_vertical_levels();
+  const auto qsize = params.qsize;
+
+  // Sanity checks since we will be switching between ekat::Pack
+  // and Homme::Scalar view types
+  EKAT_ASSERT(NLEV  == ekat::npack<Pack>(total_levels));
```
bartgol, line 200: You could use `EKAT_ASSERT_MSG`, and add an explanation of what the root cause could be

components/eamxx/src/dynamics/homme/eamxx_homme_iop.cpp
tcclevenger: What would be the point of going with the `ekat_masked_loop` instead of using `Pack::set(mask,)`?

And yes, an update method would be great. I think there are a few places that could already use it.

tcclevenger: To answer an earlier question, the original `T(k)` needed to be used in
```
T(k) += dt*omega(k)*T_k*Rair/(Cpair*pmid(k));
```
but now I followed your suggestion and just first scaled by `(1+dt*blah2)`.

components/eamxx/src/dynamics/homme/eamxx_homme_iop.cpp
```diff
@@ -0,0 +1,408 @@
...
+  auto s_delta_T = ekat::scalarize(delta_T);
+  auto s_delta_Q = ekat::scalarize(delta_Q);
+  auto s_omega_int = ekat::scalarize(omega_int);
+
+  // Compute omega on the interface grid by using a weighted average in pressure
+  const int pack_begin = 1/Pack::n, pack_end = (nlevs-1)/Pack::n;
+  Kokkos::parallel_for(Kokkos::TeamVectorRange(team, pack_begin, pack_end+1), [&] (const int k){
+    auto range_pack = ekat::range<IntPack>(k*Pack::n);
+    range_pack.set(range_pack<1, 1);
+    Pack pmid_k, pmid_km1, omega_k, omega_km1;
```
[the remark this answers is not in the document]
tcclevenger: Would this be legal to move Pack definition outside of TeamVectorRange loop? The only reason I question is in https://kokkos.org/kokkos-core-wiki/ProgrammingGuide/HierarchicalParallelism.html#team-loops 

> Note that it is not legal to make a write access to POD data outside the closure of a nested parallel layer. This is a conscious choice to prevent difficult-to-debug issues related to thread private, team shared and globally shared variables. A simple way to enforce this is by using the “capture by value”’ clause with lambdas, but “capture by reference” is recommended for release builds...

I guess `Pack` isn't exactly a POD type, but capturing by value would fail.

Testing though, for openMP it looks like it works and give identical answers.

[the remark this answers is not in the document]
bartgol: Uhm, that may be a valid point. I need to double check this with the kokkos folks.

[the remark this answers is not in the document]
bartgol: It appears you are correct. The temp needs to be declared locally.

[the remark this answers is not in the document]
tcclevenger: How significant is the performance hit?

bartgol (approved):

tcclevenger:
@bogensch Only question I have for this one is to verify, we don't need anything that is `if (.not. scm_multcols)`, correct? I did not port `advance_iop_nudging()`.

bogensch:
@tcclevenger correct; you don't need to port anything with if (.not. scm_multcols) which means that advance_iop_nudging() need not be ported.

bogensch (approved):
Looks good to me,  thanks again for your hard work on this!

PR #2719, opened 2024-02-16
Load SPA data in IOP cases

tcclevenger:
Add spa process to DP unit test and allow for correctly loading on IOP grid.


components/eamxx/src/physics/spa/spa_functions.hpp
```diff
@@ -124,6 +127,33 @@ struct SPAFunctions
...
+
+    void read_variables(const int time_index) {
+      iop->read_fields_from_file_for_iop(file_name, field_names, util::TimeStamp(), field_mgr, time_index);
+    }
+
+    iop_ptr_type iop;
+    std::string file_name;
+    std::vector<std::string> field_names;
+    std::shared_ptr<FieldManager> field_mgr;
+  };
```
tcclevenger, line 155: Reader struct for IOP. Analogous to the `AtmosphereInput` class where we give a list of fields to read, and define a `read_variables()` function.

components/eamxx/src/physics/spa/spa_functions_impl.hpp
```diff
@@ -91,12 +92,14 @@ create_horiz_remapper (
...
   std::shared_ptr<AbstractRemapper> remapper;
-  EKAT_REQUIRE_MSG (ncols_data<=ncols_model,
-      "Error! We do not allow to coarsen spa data to fit the model. We only allow\n"
-      "       spa data to be at the same or coarser resolution as the model.\n");
-  if (ncols_data==ncols_model) {
+  if (ncols_data==ncols_model
+      or
+      use_iop /*IOP class defines it's own remapper for file data*/) {
     remapper = std::make_shared<IdentityRemapper>(horiz_interp_tgt_grid,IdentityRemapper::SrcAliasTgt);
   } else {
```
tcclevenger, line 99: For IOP, force remaper to be identity. `IOPReader` will automatically read file data onto the remapper target grid.

components/eamxx/src/physics/spa/spa_functions_impl.hpp
```diff
@@ -425,12 +444,13 @@ perform_vertical_interpolation(
...
  */
 template<typename S, typename D>
+template<typename DataReader>
 void SPAFunctions<S,D>
 ::update_spa_data_from_file(
-          AtmosphereInput&  spa_data_reader,
+          DataReader&       spa_data_reader,
     const int               time_index, // zero-based
           AbstractRemapper& spa_horiz_interp,
           SPAInput&         spa_input)
```
tcclevenger, line 456: Abstract the SPA update functions (this one and the one below) on `DataReader` type, where the `spa_data_reader` object only needs to define a function `read_variables(time_index)` that reads file data to fields defined on the remapper source grid. Standard grids use `AtmosphereInput` and can have a remapper, IOP uses `IOPReader` and is required to have identity remapper.

tcclevenger (commented):
Added some comments to explain the changes.

components/eamxx/src/control/intensive_observation_period.cpp
```diff
@@ -502,8 +503,10 @@ read_fields_from_file_for_iop (const std::string& file_name,
     // Sync fields to device
     fm_field.sync_to_dev();
 
-    // Set the initial time stamp on FM fields
-    fm_field.get_header().get_tracking().update_time_stamp(initial_ts);
+    // Set the initial time stamp on FM fields (if a valid TimeStamp is given)
+    if (initial_ts.is_valid()) {
```
bartgol, line 507: Why would initial_ts not be valid?

components/eamxx/src/physics/spa/spa_functions.hpp
```diff
@@ -124,6 +127,33 @@ struct SPAFunctions
...
+        field_mgr->add_field(f);
+        field_names.push_back(f.name());
+      }
+
+      // Set IO info for this grid and file in IOP object
+      iop->setup_io_info(file_name, io_grid_);
+    }
+
+    void read_variables(const int time_index) {
+      iop->read_fields_from_file_for_iop(file_name, field_names, util::TimeStamp(), field_mgr, time_index);
```
bartgol, line 148: I guess this is the case where the timestamp is invalid?

components/eamxx/src/physics/spa/spa_functions_impl.hpp
bartgol: When I wrote the interface, I used a reference (and deref-ed the pointer at call site) for the remapper, but you could easily do
```
update_spa_data_from_file(const std::shared_ptr<AtmosphereInput>& scorpio_reader,
						 const std::shared_ptr<IOPReader>& iop_reader,
						 ...)
{
  if (iop_reader) {
    iop_reader->blah(...);
  } else {
    scorpio_reader->blah2(...);
  }
}
```
so that the two readers don't necessarily need to have the same interface.

Up to you. If you like the template version better, that's fine with me. Just wanted to point out you can change the interface, if you think it's better.

components/eamxx/src/physics/spa/spa_functions.hpp
bartgol: You could make SPA (the atm proc) pass the current year to `update_spa_data_from_file`, and then we can create the timestamp, since `time` in the spa data is the days since the start of the year:
```
TimeStamp t({yy,1,1},{0,0,0}); // yy is the year passed by SPA
auto time = scorpio::read_time_at_index_c2f(filename,time_index);
t += time*86400;
```
Up to you. This way you'd remove the check on the timestamp in IOP. Not sure if you think it's worth it though.

bartgol (approved):
Looks good. I commented with some possible alternatives, but nothing that seems objectively better. Just throwing there options, and you decide based on what you think looks cleaner/easier to read/maintain.

components/eamxx/src/control/intensive_observation_period.cpp
tcclevenger: Good point, there is not reason we can't set the timestamp for SPA fields (it just wasn't being done). I'll remove this check and pass the current timestamp.

components/eamxx/src/physics/spa/spa_functions_impl.hpp
tcclevenger: I'll make this change so that I can pass in a timestamp for the IOP reader.

components/eamxx/src/physics/spa/spa_functions.hpp
tcclevenger: Everywhere we call this function the timestamp is available, so I've changed it so that it is just passed in.

bartgol (approved):

PR #2720, opened 2024-02-17
Some changes and upgrades to IO utils

bartgol:
The two most important changes are

- At init, and at every output step, precompute the next write timestamp. At run time, check if it's an output step by comparing the current timestamp with the next write timestamp. This is more stable than "checking backwards", and requires less complicated timestamp manipulation when checking if it's an output step. In particular, if the output is, say, monthly, and we start from Jan30th, the master version would have completely skipped Feb. With this branch, we get an output at Feb 28th. The drawback (but acceptable, and better than current master) is that from Feb on, output will always happen on the 28th day of the month, since we have no memory of the fact that before Feb it happened on the 30th.
- Allow to save one output file per month or year. This feature creates output files with a different time string, namely `blah.YYYY-MM.nc` and `blah.YYYY.nc`. To request this version of output, one needs to use `file_max_storage_type: one_month` (or `one_year`). The option `Max Snapshots Per File` will be ignored.

There are a few other changes, which are mostly a reorg or how we store data.


components/eamxx/src/share/io/scream_io_control.hpp
```diff
@@ -0,0 +1,103 @@
...
+      auto date = last_write_ts.get_date();
+      if (frequency_units=="nmonths") {
+        int temp = date[1] + frequency - 1;
+        date[1] = temp % 12 + 1;
+        date[0] += temp / 12;
+      } else {
+        date[0] += frequency;
+      }
+
+      // Fix day, in case we moved to a month/year where current days. E.g., if last_write
```
AaronDonahue, line 83: Note, I think the "HOWEVER" here should be fine.  I think this edge case will be exceedingly rare, it requires a monthly output frequency for a simulation that has started at the end of the month.

components/eamxx/src/share/io/scream_output_manager.cpp
```diff
@@ -790,13 +825,25 @@ push_to_logger()
...
+  m_atm_logger->info("              Reference t0: " + m_output_control.last_write_ts.to_string());
   m_atm_logger->info("         Is Restart File ?: " + bool_to_string(m_is_model_restart_output));
   m_atm_logger->info("        Is Restarted Run ?: " + bool_to_string(m_is_restarted_run));
   m_atm_logger->info("            Averaging Type: " + e2str(m_avg_type));
   m_atm_logger->info("          Output Frequency: " + std::to_string(m_output_control.frequency) + " " + m_output_control.frequency_units);
-  m_atm_logger->info("         Max snaps in file: " + std::to_string(m_output_file_specs.max_snapshots_in_file));  // TODO: add "not set" if the value is -1
-  m_atm_logger->info("      Includes Grid Data ?: " + bool_to_string(m_output_file_specs.save_grid_data));
+  switch (m_output_file_specs.storage.type) {
+    case NumSnaps:
+      m_atm_logger->info("             File Capacity: " + std::to_string(m_output_file_specs.storage.max_snapshots_in_file) + " snapshots");  // TODO: add "not set" if the value is -1
```
AaronDonahue, line 835: Is there a reason the TODO here isn't added in the PR?

AaronDonahue (commented):
I had just one question.

components/eamxx/src/share/io/scream_output_manager.cpp
bartgol: I can add that.

AaronDonahue (approved):

PR #2721, opened 2024-02-19
Change perf suite. Add test mod

jgfouca:


cime_config/tests.py
```diff
@@ -602,12 +602,11 @@
...
 
+    # Used to track performance
     "e3sm_scream_v1_hires" : {
-        "time"  : "03:00:00",
+        "time"  : "01:00:00",
         "tests" : (
-            "SMS_D_Ln12.ne120_r0125_oRRS18to6v3.F2010-SCREAMv1",
-            "SMS_Ln12.ne120_ne120.F2010-SCREAMv1",
-#            "SMS_Ln12.ne120_r0125_oRRS18to6v3.F2000-SCREAMv1-AQP1", add when aquap 120 inputs available
+            "SMS_Ln90.ne30pg2_ne30pg2.F2010-SCREAMv1.scream-perf_test"
```
bartgol, line 609: Wouldn't a `_P8` have worked without the need to use a testmod?

jgfouca: Yes, but that commits us to P8 on all platforms. If we want this test to use a full single node, that will mean different pelayout on different machines, so I think a testmod is a better approach.

bartgol: Oh, I guess on other machines we still want to use whatever many nodes ne30 maps to... Ok.

bartgol (approved):

ambrad (commented):
At a future time, we might include certain types of output/maps to include, too, which makes the testmod approach here useful.

Suggestion: Rename the group to "e3sm_scream_v1_perf_small" and run it nightly. Later we might add something like "e3sm_scream_v1_perf_large" and run it weekly.

jgfouca:
OK, I have turned up the frequency of this run since it's cheap. It will run nightly (early AM) now.

PR #2722, opened 2024-02-20
Set boundary of shifted packs correctly in IOP forcing

tcclevenger:
Fixes the compute sanitizer initcheck error where we were accessing uninitialized data because the shifted Packs were not set up correctly at model bottom.

Additionally, this fixes a bug that wasn't caught where`delta_u_k` not defined correctly at model top
```
u_k.set(at_top, u_k - fac*omega_int_kp1*delta_u_k);
```
This was not caught because `omega_int=0` at top of model, so update was 0 anyways.

### DIFF
--- a/components/eamxx/src/dynamics/homme/eamxx_homme_iop.cpp
+++ b/components/eamxx/src/dynamics/homme/eamxx_homme_iop.cpp
@@ -97,6 +97,8 @@ advance_iop_subsidence(const KT::MemberType& team,
     const auto at_top = range_pack==0;
     const auto at_bot = range_pack==nlevs-1;
     const auto at_mid = not at_top and not at_bot;
+    const bool any_at_top = at_top.any();
+    const bool any_at_bot = at_bot.any();
 
     // Get delta(k-1) packs. The range pack should not
     // contain index 0 (so that we don't attempt to access
@@ -115,10 +117,15 @@ advance_iop_subsidence(const KT::MemberType& team,
     // At the top and bottom of the model, set the end points for
     // delta_*_k and delta_*_km1 to be the first and last entries
     // of delta_*, respectively.
-    if (at_top.any() or at_bot.any()) {
-      delta_u_k.set(at_top, s_delta_u(0)); delta_u_km1.set(at_bot, s_delta_u(nlevs-2));
-      delta_v_k.set(at_top, s_delta_v(0)); delta_v_km1.set(at_bot, s_delta_v(nlevs-2));
-      delta_T_k.set(at_top, s_delta_T(0)); delta_T_km1.set(at_bot, s_delta_T(nlevs-2));
+    if (any_at_top) {
+      delta_u_k.set(at_top, s_delta_u(0));
+      delta_v_k.set(at_top, s_delta_v(0));
+      delta_T_k.set(at_top, s_delta_T(0));
+    }
+    if (any_at_bot) {
+      delta_u_km1.set(at_bot, s_delta_u(nlevs-2));
+      delta_v_km1.set(at_bot, s_delta_v(nlevs-2));
+      delta_T_km1.set(at_bot, s_delta_T(nlevs-2));
     }
 
     // Get omega_int(k+1) pack. The range pack should not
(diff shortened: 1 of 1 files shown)


tcclevenger:
@AaronDonahue this *should* fix compute sanitizer, but weaver is so slow right now so I haven't been able to test. I'll confirm this before it merges.

components/eamxx/src/dynamics/homme/eamxx_homme_iop.cpp
```diff
@@ -94,28 +94,40 @@ advance_iop_subsidence(const KT::MemberType& team,
...
+         delta_v_k, delta_v_km1,
+         delta_T_k, delta_T_km1;
+    ekat::index_and_shift<-1>(s_delta_u, range_pack_for_m1_shift, delta_u_k, delta_u_km1);
+    ekat::index_and_shift<-1>(s_delta_v, range_pack_for_m1_shift, delta_v_k, delta_v_km1);
+    ekat::index_and_shift<-1>(s_delta_T, range_pack_for_m1_shift, delta_T_k, delta_T_km1);
+
+    // At the top and bottom of the model, set the end points for
+    // delta_*_k and delta_*_km1 to be the first and last entries
+    // of delta_*, respectively.
+    if (at_top.any() or at_bot.any()) {
```
bartgol, line 118: You may break this into two if statements. There's no point in running the `at_top` set, when only `at_top` has any masked entry, and viceversa.

components/eamxx/src/dynamics/homme/eamxx_homme_iop.cpp
```diff
@@ -144,7 +156,12 @@ advance_iop_subsidence(const KT::MemberType& team,
     Pack delta_tracer_k, delta_tracer_km1;
     for (int iq=0; iq<n_q_tracers; ++iq) {
       auto s_delta_tracer = Kokkos::subview(s_delta_Q, iq, Kokkos::ALL());
-      ekat::index_and_shift<-1>(s_delta_tracer, range_pack_m1, delta_tracer_k, delta_tracer_km1);
+      ekat::index_and_shift<-1>(s_delta_tracer, range_pack_for_m1_shift, delta_tracer_k, delta_tracer_km1);
+      if (at_top.any() or at_bot.any()) {
+        delta_tracer_k.set(at_top, s_delta_tracer(0));
```
bartgol, line 161: This `at_top` seems like a typo. I guess it should be `at_bot`?

components/eamxx/src/dynamics/homme/eamxx_homme_iop.cpp
```diff
@@ -144,7 +156,12 @@ advance_iop_subsidence(const KT::MemberType& team,
     Pack delta_tracer_k, delta_tracer_km1;
     for (int iq=0; iq<n_q_tracers; ++iq) {
       auto s_delta_tracer = Kokkos::subview(s_delta_Q, iq, Kokkos::ALL());
-      ekat::index_and_shift<-1>(s_delta_tracer, range_pack_m1, delta_tracer_k, delta_tracer_km1);
+      ekat::index_and_shift<-1>(s_delta_tracer, range_pack_for_m1_shift, delta_tracer_k, delta_tracer_km1);
+      if (at_top.any() or at_bot.any()) {
```
bartgol, line 160: Here too, I would break the if into two, to avoid running the `set` method of the one you don't have (since we never have both of them true at the same time).

bartgol: Do you think it makes sense to cache the value of `at_top.any()` in a local var? It may be a micro (or nano) opt, so probably not really needed? Although, for MAM4xx cases, we may have close to 100 tracers...so maybe precomputing `const bool at_top_any = at_top.any();` and similarly for `at_bot`, may prove helpful?

bartgol (commented):
I have a few comments, and one possible typo.

Also, I think we should update ekat, and take advantage of `Pack::update`. It would allow to avoid computing temporaries like `uk-fac*delta_uk`, and it would also allow to run only 2 update calls, rather than 3 set calls, since all we need is 
```
uk.update(not_top, delta_one,-fac,1);
uk.update(not_bot, delta_two,-fac,1);
```
Note that in `Q_k.set(at_top, Q_k - fac*omega_int_kp1*delta_tracer_k);` we are computing 3 temps:
 - `fac*omega_int_kp1`
 - `-fac*omega_int_kp1*delta_tracer_k` (the unary `-` has higher precedence than the bynary `*`)
 - `Q_k - fac*omega_int_kp1*delta_tracer_k`

components/eamxx/src/dynamics/homme/eamxx_homme_iop.cpp
tcclevenger: I think this is correct? At top (at least how I'm using it) is level=0, and at bottom is level=nlevs-1. The top and bottom being reversed index always confuses me though, so I could be missing something..

bartgol: Ah right. sorry.

tcclevenger:
@bartgol I wouldn't mind updating ekat in this PR if you have no objections

bartgol:
> @bartgol I wouldn't mind updating ekat in this PR if you have no objections. Not sure if there is anything major not related that would need its own PR..

No, I think we should do it separately. First, let's get the sanitizer to pass. We can update ekat in a follow up PR, and do this (CPU) optimization then.

bartgol (approved):

PR #2725, opened 2024-02-21
Increase length of performance test to 300 steps

jgfouca:
We need to run long enough to get at least 4 memory usage dumps, otherwise the perf comparison stuff in CIME won't work.

### DIFF
--- a/cime_config/tests.py
+++ b/cime_config/tests.py
@@ -606,7 +606,7 @@ _TESTS = {
     "e3sm_scream_v1_hires" : {
         "time"  : "01:00:00",
         "tests" : (
-            "SMS_Ln90.ne30pg2_ne30pg2.F2010-SCREAMv1.scream-perf_test"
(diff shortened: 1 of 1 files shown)


cime_config/tests.py
```diff
@@ -606,7 +606,7 @@
     "e3sm_scream_v1_hires" : {
         "time"  : "01:00:00",
         "tests" : (
-            "SMS_Ln90.ne30pg2_ne30pg2.F2010-SCREAMv1.scream-perf_test"
+            "SMS_Ln300.ne30pg2_ne30pg2.F2010-SCREAMv1.scream-perf_test"
```
bartgol, line 609: Should we also add higher res (with higher node count) like done in the `e3sm_bench_hires_g` testsuite?

jgfouca: I can add whatever the team wants. This is just a starting point.

PR #2726, opened 2024-02-21
Improve jenkins cleanup scripts

jgfouca:
Mappy has been filling up quickly lately, so we need to be more aggressive and thorough when it comes to cleaning up old test files. This PR reduces the number of preserved AT runs to 12 and adds a new script for cleaning up the scratch area.

### DIFF
--- a/components/eamxx/scripts/jenkins/scratch_cleanup.py
+++ b/components/eamxx/scripts/jenkins/scratch_cleanup.py
@@ -1,0 +1,76 @@
+#!/usr/bin/env python3
+
+"""
+Clean up old files in the scratch area for mappy.
+"""
+
+from pathlib import Path
+import re, time, shutil, sys, argparse
+
+###############################################################################
+def parse_command_line(args, description):
+###############################################################################
+    parser = argparse.ArgumentParser(
+        usage="""\n{0} [-c HOURS]
+OR
+{0} --help
+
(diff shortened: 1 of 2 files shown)


components/eamxx/scripts/jenkins/scratch_cleanup.py
```diff
@@ -0,0 +1,76 @@
...
+    parser.add_argument("-d", "--dry-run", action="store_true", help="Do a dry run, don't actually remove files")
+
+    args = parser.parse_args(args[1:])
+
+    return args
+
+###############################################################################
+def scratch_cleanup(cutoff, dry_run):
+###############################################################################
+    scratch = Path('/ascldap/users/e3sm-jenkins/acme/scratch')
```
bartgol, line 37: This script is pruning more than just scream test cases. It's pruning all e3sm-jenkins cases.

I wonder if we should just add a jenkins job that runs every 12h or so, and cleans up its scratch space. It seems more of an e3sm-jenkins duty than something that belongs to eamxx's scripts...

jgfouca: E3SM has it's own test archiving system. I did a dry run and it's only cleanup up our AT stuff.

bartgol (approved):

PR #2727, opened 2024-02-22
Add a script to generate/edit output stream yaml files

bartgol:
This script allows to create and edit output yaml files via scripting. The main reason for this PR is to be able to create testmods that add/edit output streams for our nightly testing, avoiding the combinatorial explosion due to all the possible output stream options.

So far I added support for the basic options, but I still need to add the not-so-commonly-used ones, like fp precision or restart power-user options.


bartgol:
@AaronDonahue @jgfouca This PR does quite a few changes, and I'm open to break it apart if you feel like it. Let me summarize them here

- Add a script to generate/edit an output yaml file. It does not support _all_ the options available (e.g., no support for file max snapshots, or for power-user restart options), but it is definitely enough for our testing purposes
- Add a bunch of testmods, which create different output configurations. I basically have 3 streams: dyn, phys, diags. Other testmods can set average avg type, horiz, or vert remap. I added a bunch of fields to the streams, but probably could add more. @AaronDonahue , feel free to suggest more.
- I added some "preset" testmods, which combine the testmods above, for a variety of output configurations. They are not exhaustive, but definitely an improvement w.r.t. what we had. Right now there are 6 presets. I don't think we should add more, cause we don't have enough tests to use more than those. If we think there has to be more stress-testing, we can make the presets more diverse, so that we test more configs with those 6 testmods.
- I removed the default output yaml. By default, scream generates no output, unless told to do so. I find it clearer than just shoving all possible fields in a single file. Prod runs all use their own sets of yaml files anyways...
- The above led me to a nano fix to atm_manip, when it comes to appending an option to an empty node.

jgfouca (dismissed):
Cool feature!

bartgol:
Removed automerge, in case @AaronDonahue wants to share some thoughts.

components/eamxx/cime_config/testdefs/testmods_dirs/scream/output/diags/shell_commands
```diff
@@ -0,0 +1,30 @@
+# This script generates a (single) yaml file for EAMxx output.
+# The output will be INSTANT, with only some diags fields as ouput
```
AaronDonahue, line 2: "output" misspelled.

components/eamxx/cime_config/testdefs/testmods_dirs/scream/output/README
```diff
@@ -0,0 +1,9 @@
+The testmods in this folder contain different configuration
+of output to request to eamxx. All the presets create a single
+output file, all with a different name, which means one can
+use 2+ presets at the same time.
+
+
+Each preset does basically two things:
+ 1. create a yaml file in the case folder
+ 2. add the yaml file to the yaml_output_files xml setting of eamxx
```
AaronDonahue, line 9: It looks like we have two macros, `YAML_GEN` and `YAML_EDIT`.  Could this README have a few comments briefly describing the use of each.  Namely that `YAML_GEN` is always needed, and `YAML_EDIT` can be used to add more changes.

Side question I had while reading test mods.  If we wanted a test like the current default output, which has output on both the Dynamics and Physics grids in a single output stream, then I'm guessing the way to set that up would be to use YAML_GEN to add the Physics output and then use YAML_EDIT to add Dynamics output?  Is the tool sophisticated enough to add a subfield to the main `Fields` key in the yaml file?

AaronDonahue (dismissed):

AaronDonahue:
This is a great tool @bartgol !  Thanks for putting this together.

AaronDonahue (dismissed):

bartgol:
Note: I'm waiting to merge this, since I want to fix a bug in the AT reportgen, which shows up if the overall testing fails, and one build was skipped (like in this case).

bartgol:
WIP, while I figure out how to make the ERP/ERS tests work with the testmods.

bartgol:
@AaronDonahue The diffs with baselines are expected, since we are changing the output file names. If you're ok, we can merge this so we can bless tomorrow.

jgfouca (approved):
This looks great!

AaronDonahue:
@bartgol , I think this is ready to merge.  Do you think it makes sense to rebase off of master since the last round of AT is about a week old?

bartgol:
> @bartgol , I think this is ready to merge. Do you think it makes sense to rebase off of master since the last round of AT is about a week old?

Yeah, def needs a rebase and another round of testing.

bartgol:
Since Jim already approved before the rebsae and the only fails are diffs due to missing baselines, I'm going to merge this.

PR #2730, opened 2024-02-22
Add F20TR-SCREAMv1 compset

brhillman:
Add F20TR-SCREAMv1 compset. This contains the minimum changes required to get a 20th century transient case created, but it is up to the user (or runscript) to configure further. Required to make sure MOSART is turned on for SCREAM decadal simulations, and to pull the correct ELM use case files at case creation.

### DIFF
--- a/components/elm/bld/namelist_files/namelist_defaults.xml
+++ b/components/elm/bld/namelist_files/namelist_defaults.xml
@@ -268,6 +268,7 @@ ic_tod="0" sim_year="2000" glc_nec="0" use_crop=".true." irrigate=".true." nu_co
 <finidat hgrid="ne1024np4.pg2" mask="ICOS10" ic_ymd="20160801">lnd/clm2/initdata/20220928.I2010CRUELM.ne1024pg2_ICOS10.elm.r.2016-08-01-00000.nc</finidat>
 <finidat hgrid="ne1024np4.pg2" mask="ICOS10" ic_ymd="20200120">lnd/clm2/initdata/20221218.F2010-CICE.ne30pg2_ne1024pg2.elm.r.2020-01-20-00000.nc</finidat>
 <finidat hgrid="ne1024np4.pg2" mask="ICOS10" ic_ymd="19941001">lnd/clm2/initdata/20231226.I2010CRUELM.ne1024pg2_ICOS10.elm.r.1994-10-01-00000.nc</finidat>
+<finidat hgrid="ne256np4.pg2"  mask="ICOS10" ic_ymd="19941001">lnd/clm2/initdata/20240104.I2010CRUELM.ne256pg2.elm.r.1994-10-01-00000.nc</finidat>
 <finidat hgrid="ne1024np4.pg2" mask="ICOS10" ic_ymd="101">lnd/clm2/initdata/20220717.I2010CRUELM.ne1024pg2_ICOS10.elm.r.2013-01-01-00000.nc</finidat>
 <finidat hgrid="ne1024np4.pg2" mask="ICOS10" ic_ymd="401">lnd/clm2/initdata/20220717.I2010CRUELM.ne1024pg2_ICOS10.elm.r.2013-04-01-00000.nc</finidat>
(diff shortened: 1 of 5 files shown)


whannah1 (approved):

components/elm/bld/namelist_files/use_cases/20thC_transient.xml
```diff
@@ -30,5 +30,10 @@
 <stream_year_last_popdens  phys="elm" use_cn=".true."   >2010</stream_year_last_popdens>
 <model_year_align_popdens  phys="elm" use_cn=".true."   >1850</model_year_align_popdens>
 
+<!-- Hack to override files for ne1024 for SCREAM, even though the sim years do not match up -->
+<fsurdat hgrid="ne1024np4.pg2">lnd/clm2/surfdata_map/surfdata_ne1024pg2_simyr2010_c211021.nc</fsurdat>
+<flanduse_timeseries hgrid="ne1024np4.pg2">lnd/clm2/surfdata_map/landuse.timeseries_ne1024pg2_historical_simyr1990-2014_c240109.nc</flanduse_timeseries>
+<check_finidat_year_consistency>.false.</check_finidat_year_consistency>
+<check_dynpft_consistency>.false.</check_dynpft_consistency>
```
crterai, line 37: Would you want the two statements above to be false by default, even for non-SCREAM simulations? Can we add the `hgrid="ne1024np4.pg2"` condition to these lines?

brhillman: Yeah, good point. For what it's worth, these lines were already in the 20thC_CMIP6_transient.xml file.

crterai: Thanks. I didn't know that there was a separate xml file for CMIP6, but figured someone had kept those consistency checks for some reason.

crterai (approved):
Looks good to me!

components/elm/bld/namelist_files/use_cases/20thC_transient.xml
wlin7: The 'false' flags are just to allow more tolerance when there is a mismatch. finidat_year should be ok for the the planned AMIP as the initial year will match. dynpft_consistency needs to be false for the the SCREAM amip because we set to use an finidat that was spun-up  using simyr2010 settings.

components/elm/bld/namelist_files/namelist_defaults.xml
```diff
@@ -267,10 +267,12 @@ ic_tod="0" sim_year="2000" glc_nec="0" use_crop=".true." irrigate=".true." nu_co
 <!-- Initial conditions for DYAMOND1, DYAMOND2 and for seasons of ENSO neutral year 2013 -->
 <finidat hgrid="ne1024np4.pg2" mask="ICOS10" ic_ymd="20160801">lnd/clm2/initdata/20220928.I2010CRUELM.ne1024pg2_ICOS10.elm.r.2016-08-01-00000.nc</finidat>
 <finidat hgrid="ne1024np4.pg2" mask="ICOS10" ic_ymd="20200120">lnd/clm2/initdata/20221218.F2010-CICE.ne30pg2_ne1024pg2.elm.r.2020-01-20-00000.nc</finidat>
+<finidat hgrid="ne1024np4.pg2" mask="ICOS10" ic_ymd="19941001">lnd/clm2/initdata/20231226.I2010CRUELM.ne1024pg2_ICOS10.elm.r.1994-10-01-00000.nc</finidat>
```
wlin7, line 270: Could also put in a default finidat for ne256pg2 for 1994-01-01-00000.nc.

wlin7 (approved):
The changes to enable the new compset look ok. Other settings for running AMIP could be specified in the run script.

components/elm/bld/namelist_files/namelist_defaults.xml
brhillman: Yeah, I could do that too, but we don't rely on these at all since the compset keeps the default start date and I elected to just set via runscript. I had hoped to actually avoid creating a new compset altogether but needed to in order to turn on MOSART.

brhillman: Okay, pushed a commit to add the default for ne256pg2. It seems to get picked up.

PR #2733, opened 2024-02-27
Aarondonahue/add avg cnt track to field at height

AaronDonahue:
This commit fixes an oversight where we were only uniquely tracking the averaging counting for field slices at pressure level.  We also need to include any fields sliced at a specific height above sealevel, because those could also have unique masking.


bartgol:
Given what gh is doing on the other PR, you may want to rebase this on master as well...

bartgol (dismissed):

PR #2737, opened 2024-02-29
Fix handling of cloud fraction for COSP

brhillman:
Fix handling of cloud fraction for COSP. This makes two changes to how cloud fraction is treated in COSP. First, we switch to using the cloud fraction that is consistent with that used in radiation and the mask used in microphysics. Previously, we used the "analysis" cloud fraction (`cldfrac_tot_for_analysis`) to set the cloud fraction for subcolumn sampling in COSP for all configurations of EAMxx, in a misguided attempt to use a cloud fraction that is more consistent with observations. This was inconsistent with the cloud fraction used for subcolumn sampling in radiation, however, which uses a less stringent criteria for considering a cell to contain ice cloud. This also led to a gross underestimation/reporting of ISCCP-simulated cloud area at ne1024. Second, we disable subcolumn sampling in COSP by default for resolutions finer than ne256, which is on the order of the resolution of ISCCP satellite pixels. The point of the subcolumn sampling in COSP is to address the mis-match between model resolutions and satellite retrieval scales, accounting for vertical overlap of unresolved clouds. However, at resolutions finer than ne256 we approach the scale of satellite pixels, so performing pseudo-retrievals on gridbox-mean optical properties is more consistent with the satellite retrievals, and is less ambiguous. Making this change further improves the ISCCP-simulated cloud cover.

Note that these changes will be non-BFB only for ISCCP-simulated outputs (isccp_ctptau, isccp_cldtot).

### DIFF
--- a/components/eamxx/cime_config/namelist_defaults_scream.xml
+++ b/components/eamxx/cime_config/namelist_defaults_scream.xml
@@ -372,7 +372,7 @@ be lost if SCREAM_HACK_XML is not enabled.
     </mac_aero_mic>
 
     <cosp inherit="atm_proc_base">
-      <cosp_subcolumns>10</cosp_subcolumns>
+      <cosp_subcolumns doc="Number of subcolumns to use for COSP simulators; cosp_subcolumns=1 implies no subcolumn sampling">10</cosp_subcolumns>
       <cosp_subcolumns hgrid="ne1024np4">1</cosp_subcolumns>  <!-- No subcolumn sampling at satellite pixel scales -->
       <cosp_subcolumns hgrid="ne512np4" >1</cosp_subcolumns>  <!-- No subcolumn sampling at satellite pixel scales -->
       <cosp_subcolumns hgrid="ne256np4" >1</cosp_subcolumns>  <!-- No subcolumn sampling at satellite pixel scales -->
--- a/components/eamxx/cime_config/namelist_defaults_scream.xml
+++ b/components/eamxx/cime_config/namelist_defaults_scream.xml
@@ -354,7 +354,9 @@ be lost if SCREAM_HACK_XML is not enabled.
       >
         false
       </extra_clnsky_diag>
-      <do_subcol_sampling type="logical">true</do_subcol_sampling>
+      <do_subcol_sampling type="logical" doc="Flag to turn on/off subcolumn sampling of optical properties; if false treat cells as either completely clear or cloudy">
+          true
+      </do_subcol_sampling>
     </rrtmgp>
 
     <mac_aero_mic inherit="atm_proc_group">
--- a/components/eamxx/cime_config/namelist_defaults_scream.xml
+++ b/components/eamxx/cime_config/namelist_defaults_scream.xml
@@ -373,6 +373,9 @@ be lost if SCREAM_HACK_XML is not enabled.
 
     <cosp inherit="atm_proc_base">
       <cosp_subcolumns>10</cosp_subcolumns>
+      <cosp_subcolumns hgrid="ne1024np4">1</cosp_subcolumns>  <!-- No subcolumn sampling at satellite pixel scales -->
+      <cosp_subcolumns hgrid="ne512np4" >1</cosp_subcolumns>  <!-- No subcolumn sampling at satellite pixel scales -->
+      <cosp_subcolumns hgrid="ne256np4" >1</cosp_subcolumns>  <!-- No subcolumn sampling at satellite pixel scales -->
       <!-- Frequency at which to call COSP; positive values interpreted as number of steps, negative as number of hours -->
       <cosp_frequency>1</cosp_frequency>
       <cosp_frequency_units valid_values="steps,hours">hours</cosp_frequency_units>
--- a/components/eamxx/cime_config/namelist_defaults_scream.xml
+++ b/components/eamxx/cime_config/namelist_defaults_scream.xml
@@ -354,6 +354,7 @@ be lost if SCREAM_HACK_XML is not enabled.
       >
         false
       </extra_clnsky_diag>
+      <do_subcol_sampling type="logical">true</do_subcol_sampling>
     </rrtmgp>
 
     <mac_aero_mic inherit="atm_proc_group">
--- a/components/eamxx/src/physics/cosp/cosp_c2f.F90
+++ b/components/eamxx/src/physics/cosp/cosp_c2f.F90
@@ -295,41 +295,27 @@ contains
     ! In-cloud values are assumed. If ncolumns = 1, then convert in-cloud values to gridbox
     if (ncolumns == 1) then
        tca(:npoints,:nlevels) = cldfrac(:npoints,:nlevels)
-       cca(:npoints,:nlevels) = 0
-       mr_lsliq(:npoints,:nlevels) = 0
-       mr_ccliq(:npoints,:nlevels) = 0
-       mr_lsice(:npoints,:nlevels) = 0
-       mr_ccice(:npoints,:nlevels) = 0
-       dtau_c(:npoints,:nlevels) = 0
        dtau_s(:npoints,:nlevels) = cldfrac(:npoints,:nlevels) * dtau067(:npoints,:nlevels)
-       dem_c (:npoints,:nlevels) = 0
        dem_s (:npoints,:nlevels) = 1._wp - exp(-cldfrac(:npoints,:nlevels) * dtau105(:npoints,:nlevels))
-       mr_lsrain(:npoints,:nlevels) = 0
-       mr_ccrain(:npoints,:nlevels) = 0
-       mr_lssnow(:npoints,:nlevels) = 0
-       mr_lssnow(:npoints,:nlevels) = 0
-       mr_ccsnow(:npoints,:nlevels) = 0
-       mr_lsgrpl(:npoints,:nlevels) = 0
-       reff = 0  ! FIXME
     else
        tca(:npoints,:nlevels) = cldfrac(:npoints,:nlevels)
-       cca(:npoints,:nlevels) = 0
-       mr_lsliq(:npoints,:nlevels) = 0
-       mr_ccliq(:npoints,:nlevels) = 0
-       mr_lsice(:npoints,:nlevels) = 0
-       mr_ccice(:npoints,:nlevels) = 0
-       dtau_c(:npoints,:nlevels) = 0
        dtau_s(:npoints,:nlevels) = dtau067(:npoints,:nlevels)
-       dem_c (:npoints,:nlevels) = 0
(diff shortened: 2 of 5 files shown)


brhillman:
Some comparisons of ISCCP-simulated quantities for a 1-day ne1024 test follow. First, the ISCCP-simulated total cloud area:

isccp_cldtot_masked_average ne1024-analysisCldfrac
isccp_cldtot_masked_average ne1024-ncolumns10
isccp_cldtot_masked_average ne1024-ncolumns1

The first figure shows the current behavior, using `cldfrac_tot_for_analysis` and doing subcolumn sampling in COSP with 10 subcolumns. Numbers in the titles indicate area-weighted global means over the 1-day simulations. The second figure shows cloud area after changing the code to use `cldfrac_tot` (the same cloud fraction used in radiation) for the subcolumn sampling in COSP. This increases the area-weighted global mean by about 8% cloud area, which is a positive change in itself. The third figure shows the change to use no subcolumns in COSP. This takes the in-cloud optical depth computed in radiation and converts to a grid-mean optical depth by scaling by the cloud fraction radiation used to compute in-cloud condensate mixing ratios from the grid-mean condensate passed from microphysics. This means that subgrid partial cloud fraction and vertical overlap of unresolved cloud is still considered when computing optical properties, but COSP simulators only see the grid-mean optical properties, consistent with how a satellite-based radiometer would view the gridbox. This change results in a further increase in ISCCP-simulated cloud area. This is again a positive change, as we know that ISCCP-retrieved cloud area is about 68% in the global climatological mean, but our SCREAM ISCCP-simulated cloud cover from the Cess simulations was less than 50%.

Joint histogram comparisons are maybe difficult to interpret from these one-day simulations, but are shown for documentation:

isccp_ctptau_masked_average ne1024-analysisCldfrac
isccp_ctptau_masked_average ne1024-ncolumns10
isccp_ctptau_masked_average ne1024-ncolumns1

Numbers in the titles indicate 1) the sum of all bins (total cloud fraction disregarding optical depth) and 2) the sum of only those bins with optical depth greater than 0.3 (more consistent with ISCCP retrievals). Both fixes result in noticeable changes in the joint histograms, primarily to add cloud occurrence to the lower optical depth bins. This is expected: switching to the internal (less stringent) ice cloud mask means that more cells will be considered cloudy that have low ice concentrations. These extra cells would likewise have low optical depths, so we would expect these extra occurrences to show up in the low optical depth bins. These cases would also likely show up as high altitude clouds, since any change to low-level clouds would likely already be dominated by liquid phase condensate so wouldn't have a significant impact on the occurrence. This is exactly what we see, with cloud occurrence increased for low cloud top pressures in the ncolumns10 case. The change to using no subcolumns for COSP leads to further increases in the frequency of low to intermediate optical depths, but also a shift to or increase in the occurrence of gridboxes with diagnosed high cloud top pressures. I think this is just a consequence of the simulator now seeing grid-mean optical depths with no regard for overlap; high-topped (low cloud top pressure) occurrences are diagnosed less frequency due to contributions from lower-level clouds when the gridbox is taken as a whole (each pseudo-retrieval can only be single layered, so by eliminating subcolumn sampling we eliminate the possibility to diagnose separately the thin high-level ice cloud overlying lower level liquid cloud).

In all, I think these results are consistent with our expectations.

components/eamxx/src/physics/cosp/eamxx_cosp.cpp
```diff
@@ -140,7 +140,7 @@ void Cosp::run_impl (const double dt)
   get_field_in("T_mid").sync_to_host();
   get_field_in("p_mid").sync_to_host();
   get_field_in("p_int").sync_to_host();
-  get_field_in("cldfrac_tot_for_analysis").sync_to_host();
+  get_field_in("cldfrac_rad").sync_to_host();
```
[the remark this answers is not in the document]
brhillman: It's not always an alias to `cldfrac_tot`:
```
      if (not do_subcol_sampling) {
        const auto policy = ekat::ExeSpaceUtils<ExeSpace>::get_default_team_policy(ncol, m_nlay);
          const int i = team.league_rank();
          const int icol = i + beg;
          Kokkos::parallel_for(Kokkos::TeamVectorRange(team, nlay), [&] (const int& k) {
            if (d_cldfrac_tot(icol,k) > 0) {
              cldfrac_tot(i+1,k+1) = 1;
            } else {
              cldfrac_tot(i+1,k+1) = 0;
            }
            d_cldfrac_rad(icol,k) = cldfrac_tot(i+1,k+1);
          });
        });
      } else {
        const auto policy = ekat::ExeSpaceUtils<ExeSpace>::get_default_team_policy(ncol, m_nlay);
          const int i = team.league_rank();
          const int icol = i + beg;
          Kokkos::parallel_for(Kokkos::TeamVectorRange(team, nlay), [&] (const int& k) {
            cldfrac_tot(i+1,k+1) = d_cldfrac_tot(icol,k);
            d_cldfrac_rad(icol,k) = d_cldfrac_tot(icol,k);
          });
        });
      }
```
The danger is if `do_subcol_sampling == false` in the radiation interface, the radiation uses a binary cloud fraction. The point of adding this is to make sure we are treating cloud fraction consistently between radiation and cosp.

components/eamxx/cime_config/namelist_defaults_scream.xml
```diff
@@ -372,6 +373,9 @@ be lost if SCREAM_HACK_XML is not enabled.
 
     <cosp inherit="atm_proc_base">
       <cosp_subcolumns>10</cosp_subcolumns>
+      <cosp_subcolumns hgrid="ne1024np4">1</cosp_subcolumns>  <!-- No subcolumn sampling at satellite pixel scales -->
+      <cosp_subcolumns hgrid="ne512np4" >1</cosp_subcolumns>  <!-- No subcolumn sampling at satellite pixel scales -->
+      <cosp_subcolumns hgrid="ne256np4" >1</cosp_subcolumns>  <!-- No subcolumn sampling at satellite pixel scales -->
       <!-- Frequency at which to call COSP; positive values interpreted as number of steps, negative as number of hours -->
```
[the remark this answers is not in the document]
brhillman: Added.

hassanbeydoun (approved):
Glad you found a fix for COSP cloud fraction woes! Looks good to me!

crterai (dismissed):
It's nice to see that the fix doesn't mean that we just see lots of high-level clouds. Thanks for running the tests and finding a fix, Ben.

components/eamxx/cime_config/namelist_defaults_scream.xml
```diff
@@ -354,6 +354,7 @@ be lost if SCREAM_HACK_XML is not enabled.
       >
         false
       </extra_clnsky_diag>
+      <do_subcol_sampling type="logical">true</do_subcol_sampling>
```
bartgol, line 357: Would you mind adding a `doc="blah"` string explaining what this parameter does?

components/eamxx/src/physics/cosp/cosp_c2f.F90
```diff
@@ -292,23 +292,44 @@ subroutine cosp_c2f_run(npoints, ncolumns, nlevels, ntau, nctp, &
...
-    dem_s (:npoints,:nlevels) = 1._wp - exp(-dtau105(:npoints,:nlevels))
-    mr_lsrain(:npoints,:nlevels) = 0
-    mr_ccrain(:npoints,:nlevels) = 0
-    mr_lssnow(:npoints,:nlevels) = 0
-    mr_lssnow(:npoints,:nlevels) = 0
-    mr_ccsnow(:npoints,:nlevels) = 0
-    mr_lsgrpl(:npoints,:nlevels) = 0
-    reff = 0  ! FIXME
+    ! In-cloud values are assumed. If ncolumns = 1, then convert in-cloud values to gridbox
+    if (ncolumns == 1) then
```
bartgol, line 296: Up to you, but since only 2 arrays are treated differently, you may just do all the `=0` assignments outside the if, and leave only `dtau_s` and `dem_s` in the conditional.

components/eamxx/src/physics/cosp/cosp_c2f.F90
```diff
@@ -292,23 +292,44 @@ subroutine cosp_c2f_run(npoints, ncolumns, nlevels, ntau, nctp, &
...
+       dtau_s(:npoints,:nlevels) = cldfrac(:npoints,:nlevels) * dtau067(:npoints,:nlevels)
+       dem_c (:npoints,:nlevels) = 0
+       dem_s (:npoints,:nlevels) = 1._wp - exp(-cldfrac(:npoints,:nlevels) * dtau105(:npoints,:nlevels))
+       mr_lsrain(:npoints,:nlevels) = 0
+       mr_ccrain(:npoints,:nlevels) = 0
+       mr_lssnow(:npoints,:nlevels) = 0
+       mr_lssnow(:npoints,:nlevels) = 0
+       mr_ccsnow(:npoints,:nlevels) = 0
+       mr_lsgrpl(:npoints,:nlevels) = 0
+       reff = 0  ! FIXME
```
bartgol, line 313: I know it was like this in the original, so no big deal, but what does this `FIXME` refer to? Is it something we should be concerned about? Should we beef up the comment, so that other devs may understand what needs to be fixed?

bartgol (changes requested):
I have one request regarding the new parameter in the XML. Then a couple of comments, which you can feel free to disregard.

components/eamxx/src/physics/cosp/cosp_c2f.F90
brhillman: I left the rest here because in the near future these will not be all zero, and wanted to keep all of the like fields together to organize things. @whannah1 brought this up too.

components/eamxx/src/physics/cosp/cosp_c2f.F90
brhillman: I can add more context here. This will need to be fixed when we turn on the MODIS, CloudSat, and CALIPSO simulators I believe, but is not used for the ISCCP and MISR simulators.

components/eamxx/cime_config/namelist_defaults_scream.xml
brhillman: Done.

bartgol (approved):

components/eamxx/src/physics/cosp/cosp_c2f.F90
brhillman: Okay I lied, I just decided to add doc about all these fields. Updated.

bartgol (dismissed):

components/eamxx/src/physics/cosp/cosp_c2f.F90
brhillman: Decided to fix this anyways. Pulled the unused fields out of the logical and added doc about all of them.

bartgol (approved):

PR #2739, opened 2024-03-04
Changes for running through CIME

tcclevenger:
Adds a compset `F2010-SCREAMv1-DP-DYCOMSrf01` and a test for Mappy AT (roughly the size of ne2 unit test). Have both `DP-EAMxx` and `DYCOMSrf01` in the compset name since some setting are general to DP-EAMxx, and some are case specific for DYCOMSrf01.

Other changes
- additions to namelist defaults
- Reload IOP data after restart
- SPA remap file to check "None" for IOP

### DIFF
--- a/components/eamxx/cime_config/namelist_defaults_scream.xml
+++ b/components/eamxx/cime_config/namelist_defaults_scream.xml
@@ -48,7 +48,7 @@ be lost if SCREAM_HACK_XML is not enabled.
       <sections>ctl_nl</sections>
     </file>
     <file name="data/scream_input.yaml" format="yaml">
-      <sections>driver_options,intensive_observation_period_options,atmosphere_processes,grids_manager,initial_conditions,Scorpio,e3sm_parameters</sections>
+      <sections>driver_options,iop_options,atmosphere_processes,grids_manager,initial_conditions,Scorpio,e3sm_parameters</sections>
     </file>
   </generated_files>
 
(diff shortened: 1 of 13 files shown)


components/eamxx/cime_config/namelist_defaults_scream.xml
```diff
@@ -48,7 +48,7 @@ be lost if SCREAM_HACK_XML is not enabled.
       <sections>ctl_nl</sections>
     </file>
     <file name="data/scream_input.yaml" format="yaml">
-      <sections>driver_options,atmosphere_processes,grids_manager,initial_conditions,Scorpio,e3sm_parameters</sections>
+      <sections>driver_options,intensive_observation_period_options,atmosphere_processes,grids_manager,initial_conditions,Scorpio,e3sm_parameters</sections>
```
bartgol, line 51: Have you thought about calling this just `iop_options`? It would make manual `atmchange` commands faster to type.

components/eamxx/cime_config/namelist_defaults_scream.xml
```diff
@@ -462,6 +465,12 @@ be lost if SCREAM_HACK_XML is not enabled.
     <nc hgrid="ne256np4|ne1024np4">0.0</nc>
     <ni hgrid="ne256np4|ne1024np4">0.0</ni>
     <o3_volume_mix_ratio hgrid="ne0np4_conus_x4v1_lowcon">0.0</o3_volume_mix_ratio>
+
+    <!-- Information about perturbed fields-->
+    <perturbed_fields type="array(string)"/>
```
bartgol, line 470: Can you add a `doc` attrib for the new parameters?

Note: if the same param appears 2+ times (like `perturbed_fields`), put the doc string only in the first occurrence.

components/eamxx/cime_config/namelist_defaults_scream.xml
```diff
@@ -490,18 +499,37 @@ be lost if SCREAM_HACK_XML is not enabled.
...
+    <enable_intensive_observation_period COMPSET=".*DP-EAMxx">true</enable_intensive_observation_period>
   </driver_options>
 
   <!-- E3SM Simulation Settings -->
   <e3sm_parameters>
     <e3sm_casename>${CASE}</e3sm_casename>
   </e3sm_parameters>
 
+  <!-- Settings for IOP -->
+  <intensive_observation_period_options>
```
bartgol, line 512: Can you add doc strings for all these params too? You should be able then to verify the docs is correct by looking at [this](https://e3sm-project.github.io/scream/pr-preview/pr-2739/common/eamxx_params/) auto-generated page.

components/eamxx/src/physics/spa/eamxx_spa_process_interface.cpp
```diff
@@ -68,7 +68,7 @@ void SPA::set_grids(const std::shared_ptr<const GridsManager> grids_manager)
   // where a single column of data corresponding to the closest lat/lon pair to
   // the IOP lat/lon parameters is read from file, and that column data is mapped
   // to all columns of the IdentityRemapper source fields.
-  EKAT_REQUIRE_MSG(spa_map_file == "" or not m_iop,
+  EKAT_REQUIRE_MSG(spa_map_file == "" or spa_map_file == "None" or not m_iop,
```
bartgol, line 71: Does `None` really end up in the yaml file? It's supposed to just have an empty string if the compset doesn't require a map file...

bartgol (changes requested):

components/eamxx/src/physics/spa/eamxx_spa_process_interface.cpp
tcclevenger: Yeah, a bit annoying but the way that it treats 
```
<spa_remap_file COMPSET="*.DP-EAMxx*"/>
```
is to fill in "None". Is there another way to delete the entry? Else it will get whatever is dictated by the `hgrid` value.

tcclevenger:
@bartgol went ahead and renamed a few other functions with "intensive_observation_period" to "iop".

bartgol (approved):

PR #2740, opened 2024-03-05
Delete cmake files for 2 machines we don't support on LC.

AaronDonahue:


AaronDonahue:
Pretty sure I can just merge this, it just removes unused files.

PR #2741, opened 2024-03-05
check-hashes-ers: Bring in to eamxx/scripts ecosystem

jgfouca:
Use standard script format. Some minor fixes. Make sure it will work with py3.6.

I've confirmed this works on chrysalis. The main issue there was CIME does a `module purge` to the env, which potentially leaves an older system python as the python that was being used to run check-hashes. There was at least one thing in there that was not available in py3.6 (the version of chrysalis' system python).


ambrad (approved):

components/eamxx/scripts/check-hashes-ers
```diff
@@ -0,0 +1,171 @@
...
+    glob_pat = get_log_glob_from_atm_modelio(case_dir_p)
+    e3sm_fns = glob.glob(glob_pat)
+    if len(e3sm_fns) == 0:
+        print('Could not find e3sm.log files with glob string {}'.format(glob_pat))
+        return False
+    e3sm_fns.sort()
+    if len(e3sm_fns) == 1:
+        # This is the first run. Exit and wait for the second
+        # run. (POSTRUN_SCRIPT is called after each of the two runs.)
+        print('Exiting on first run.')
```
bartgol, line 124: Minor, but maybe we could print more descriptive messages, so that someone looking at TestStatus.log would understand what this message refers to. Something like
```python
print ("[check-hashes-ers postrun script log] <rest of msg here>")
```
For ppl familiar with this postrun script, the messages are already enough, but the avg user may not understand what's going on.

components/eamxx/scripts/check-hashes-ers
```diff
@@ -0,0 +1,171 @@
...
+    # output. Permit up to 'thr' diffs so we don't fail due to badly printed
+    # lines. This isn't a big loss in checking because an ERS_Ln22 second run
+    # writes > 1000 hash lines, and a true loss of BFBness is nearly certain to
+    # propagate to a large number of subsequent hashes.
+    thr = 5
+    if len(lns[0]) < 100: thr = 0
+
+    ok = True
+    if len(diffs) > thr:
+        print('DIFF')
```
bartgol, line 154: In the same spirit of the other comment, we may inform the avg user of what "DIFF" refers to (namely, that the eamxx bfbhash between base/rest did not match). Maybe even print the first thr+1 occurrences? It may help to immediately see how long it took for non bfb to develop, since I'm guessing we'd see a timestamp in the log line?

bartgol (approved):

bartgol:
@jgfouca Looks good. I removed automerge in case you want to act on my comments. But I still approved, so it doesn't block in case you don't feel like acting on it.

components/eamxx/scripts/check-hashes-ers
jgfouca: @bartgol , all postrun script output goes to a separate log, `$RUNDIR/postrun.external.log.$testid`. Does that change your thoughts on this?

bartgol: So it doesn't end up in TestStatus.log? Then I suppose it's fine.

PR #2742, opened 2024-03-05
Make nudging log which file it's reading

bartgol:
With these small mods, the log from a test that uses nudging contains snippets of the form

```
[EAMxx::Nudging] Reading nudging data
[EAMxx:time_interpolation] Reading data at time 2021-10-12-46500
[EAMxx::scorpio_input] Reading variables from file
  file name: shoc_p3_source_data_nudged.INSTANT.nsteps_x5.2021-10-12-45000.nc
  var names: T_mid, qv, p_mid
  time idx : 1
  Done! Elapsed time: 0.050000 seconds

```

PR #2744, opened 2024-03-06
Fix how we read/write timestamp to file

bartgol:
A previous PR added the possibility of writing/reading also the global attribute `${timestamp_name}_nsteps`, which is used to correctly restart INSTANT output. However, this attributed is not present in older output files, so we should only read it if it's present and/or needed.

### DIFF
--- a/components/eamxx/src/share/io/scream_io_control.hpp
+++ b/components/eamxx/src/share/io/scream_io_control.hpp
@@ -22,30 +22,14 @@ struct IOControl {
   util::TimeStamp next_write_ts;
   util::TimeStamp last_write_ts;
 
-  mutable int dt = 0;
-
   bool output_enabled () const {
     return frequency_units!="none" && frequency_units!="never";
   }
 
-  void compute_dt (const util::TimeStamp& ts) {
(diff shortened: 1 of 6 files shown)


AaronDonahue (dismissed):

ndkeen:
For my test case (nudged seed case at ne256), I may need this fix after https://github.com/E3SM-Project/scream/pull/2742/ was merged, else I get: 

```
  4:  FAIL:
  4: err==0
  4: /lustre/orion/cli115/proj-shared/noel/wacmy/bartgol_eamxx_nudging-add-logging/components/eamxx/src/share/io/scream_scorpio_interface.cpp:496
  4: [get_any_attribute] Error! Something went wrong while inquiring global attribute.
  4:  - filename : /lustre/orion/cli115/proj-shared/hannah6/scream_scratch/nudge_data/DY1/HICCUP.nudging_uv_era5.2016-08-01.ne128pg2.L128.nc
  4:  - variable : GLOBAL
  4:  - attribute: case_t0_nsteps
  4:  - pio error: -43
  ```

tcclevenger (approved):

PR #2745, opened 2024-03-06
test-all-scream: Move baseline build dir

jgfouca:
Doing it in the baseline dir seems like a bad idea. One, potential races with other TAS runs. Two, on some machines like pmgpu, the baseline filesystem is MUCH slower than the scratch area.

### DIFF
--- a/components/eamxx/scripts/test_all_scream.py
+++ b/components/eamxx/scripts/test_all_scream.py
@@ -853,7 +853,8 @@ remove existing baselines first. Otherwise, please run 'git fetch $remote'.
                f"Something is off. generate_baseline should have not be called for test {test}")
 
         baseline_dir = self.get_test_dir(self._baseline_dir, test)
-        test_dir = self.get_test_dir(self._work_dir, test, "tas_baseline_build")
+        test_dir = self.get_test_dir(self._work_dir, test) / "tas_baseline_build"
+        test_dir.mkdir()
 
         cmake_config = self.generate_cmake_config(test)
         cmake_config += " -DSCREAM_BASELINES_ONLY=ON"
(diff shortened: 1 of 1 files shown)


components/eamxx/scripts/test_all_scream.py
```diff
@@ -887,10 +888,8 @@ def generate_baselines(self, test, commit):
                     success = False
 
         finally:
-            # Clean up the directory, by removing everything but the 'data' subfolder. This must
-            # happen unconditionally or else subsequent runs could be corrupted
-            run_cmd_no_fail(r"find -maxdepth 1 -not -name data ! -path . -exec rm -rf {} \;",
-                            from_dir=test_dir, verbose=True, dry_run=self._dry_run)
+            # Clean up the directory by removing everything
+            run_cmd_no_fail(f"/bin/rm -rf {test_dir}/*")
```
bartgol, line 892: If I understand correctly, the location of SCREAM_DATA_DIR when generating baselines is in the baseline dir, while the actual code is built and run in the usual ctest-build subdir, right?

If that's the case, I would _not_ clean up the folder at the end. We already nuke them at the beginning of the test phase anyways. The advantage of not nuking them is that, if something fails, we can cd in the folder and see what was going on (e.g., inspect the cache, check flags, etc.).

bartgol (commented):
Assuming I understood what this PR does, I like it. I have a suggestion though, so I'm holding off approval to avoid a possibly premature automerge. I'll approve after we settle that.

components/eamxx/scripts/test_all_scream.py
jgfouca: @bartgol , I had to move the "make test dirs and nuke them" call earlier in TAS in order to ensure this directory existed for generate_baseline. I don't think they get nuked again, so leaving the files around could case problems with the test build. That said, I don't think it would be hard to put the baseline build in a subdirectory that would be harmless to the main build. Something like `ctest-build/debug/baseline_build`.

components/eamxx/scripts/test_all_scream.py
```diff
@@ -886,17 +887,14 @@ def generate_baselines(self, test, commit):
...
-            run_cmd_no_fail(r"find -maxdepth 1 -not -name data ! -path . -exec rm -rf {} \;",
-                            from_dir=test_dir, verbose=True, dry_run=self._dry_run)
-
         if success:
             # Store the sha used for baselines generation
             self.set_baseline_file_sha(test, commit)
             test.missing_baselines = False
 
+            # Clean up the directory by removing everything
+            shutil.rmtree(test_dir)
```
bartgol, line 896: We may need to add one of these also _before_ building baselines. I'm thinking of this scenario:

1. run tas: baseline build fails, so we don't nuke the dir
2. change something, run tas again: baseline dir is not clean, so we may not get a clean build

bartgol (approved):

Issue #1996, opened 2022-10-18
Change implementation of get_vector_dim in FieldLayout


bartgol:
Right now, it returns the _index_ of the `CMP` tag. Compare this with the method `int dim(const FieldTag t)`, which returns the extent of the dimension with the given tag (assuming there's only one). This is confusing. On top of that, in all the places where we call `get_vector_dim` we're actually interested in the _extent_ of that dimension.

Hence, we should change the impl to return the extent of the `CMP` dimension. Basically, replace impl with `return dim(FieldTag::Component);`.

mjs271:
closed by #2746

PR #2746, opened 2024-03-06
Change Implementation of get_vector_dim() in FieldLayout

mjs271:
This PR addresses #1996 and makes the following changes:

1. The function previously-named `get_vector_dim()` was re-named `get_vector_component_idx()` to align with its functionality. Namely, when a `FieldLayout` object is a vector, which is to say

    ```
    fv.is_vector_layout() == true
    ```

    then this function returns the index that corresponds to the vector of _**Components**_, which may be a vector holding any of these field tags: `{CMP, NGAS, SWBND, LWBND, SWGPT, ISCCPTAU, ISCCPPRS}`.
3. A replacement version of `get_vector_dim()` has been added and provides the extent (length) of the dimension with the index provided by `get_vector_component_idx()`. E.g., in pseudocode
    ```
    fv(fv.get_vector_component_idx()).extent() == fv.get_vector_dim().
    ```
4. Tests were added for `get_vector_component_idx()`.
5. Necessary modifications were made to other parts of the code that called the previous version of `get_vector_dim()`.

### DIFF
--- a/components/eamxx/src/share/field/field_layout.cpp
+++ b/components/eamxx/src/share/field/field_layout.cpp
@@ -27,7 +27,10 @@ bool FieldLayout::is_tensor_layout () const {
   return lt==LayoutType::Tensor2D || lt==LayoutType::Tensor3D;
 }
 
-int FieldLayout::get_vector_dim () const {
+// get the index of the CMP (Components) tag in the FieldLayout
+// e.g., for FieldLayout f({COL, CMP, LEV}, {...});
+// we have get_component_idx(f) = 1
+int FieldLayout::get_vector_component_idx () const {
   EKAT_REQUIRE_MSG (is_vector_layout(),
       "Error! 'get_vector_dim' available only for vector layouts.\n"
       "       Current layout: " + e2str(get_layout_type(m_tags)) + "\n");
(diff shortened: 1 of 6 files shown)


components/eamxx/src/share/field/field_layout.cpp
```diff
@@ -43,8 +46,17 @@ int FieldLayout::get_vector_dim () const {
   return std::distance(m_tags.cbegin(),it);
 }
 
+// get the extent of the CMP (Components) tag in the FieldLayout
+// e.g., for FieldLayout f({COL, CMP, LEV}, {ncol, ncmp, nlev});
+// we have get_component_idx(f) = ncmp
```
bartgol, line 51: I think this comment got messed up during copy+paste. you probably meant `we have get_vector_dim() = ncmp`?

components/eamxx/src/share/field/field_layout.cpp
```diff
@@ -27,7 +27,10 @@ bool FieldLayout::is_tensor_layout () const {
   return lt==LayoutType::Tensor2D || lt==LayoutType::Tensor3D;
 }
 
-int FieldLayout::get_vector_dim () const {
+// get the index of the CMP (Components) tag in the FieldLayout
+// e.g., for FieldLayout f({COL, CMP, LEV}, {...});
+// we have get_component_idx(f) = 1
```
bartgol, line 32: Minor, but prob best to change `get_component_idx(f)` to `f.get_vector_component_idx()`?

components/eamxx/src/share/field/field_layout.cpp
mjs271: oof, twice in a row 🤦‍♂️

fixed!

bartgol (approved):

PR #2749, opened 2024-03-08
Fix bug in IO when doing a restarted run without restarting history

bartgol:
The start ref in that case must be run_t0, not case_t0. Otherwise, the output manager will set `last_write_ts=case_t0` which may cause `next_write_ts` to be _before_ `run_t0`...


ndkeen (approved):
Looks to be working with ne256 case on frontier as part of the PPE cases. Where we run 5-day seed case, write restarts, then run many member cases that read in from those restarts and write output.  We don't write/read "history" files.

PR #2750, opened 2024-03-08
Add baseline tests for EAMxx standalone ad tests

bartgol:
This PR adds baseline testing for AD-driven tests in test-all-scream.


components/eamxx/CMakeLists.txt
```diff
@@ -1,7 +1,18 @@
+####################################################################
```
bartgol, line 1: The changes here are mostly a reorg of the CMakeLists, to group together settings that are conceptually related, to make it easier to navigate the file.

components/eamxx/src/physics/p3/tests/CMakeLists.txt
```diff
@@ -103,42 +107,13 @@ if (SCREAM_ENABLE_BASELINE_TESTS)
       EXCLUDE_MAIN_CPP
       ${FORCE_RUN_DIFF_FAILS})
 
-  #
```
bartgol, line 106: By using a ctest + a test label (i.e., `ctest -L baseline` rather than `make baseline`), we don't really need to jump hoops to get the test launch line and args from the test.

components/eamxx/src/physics/p3/tests/CMakeLists.txt
```diff
@@ -77,7 +77,11 @@ if (NOT SCREAM_BASELINES_ONLY)
 endif()
 
 if (SCREAM_ENABLE_BASELINE_TESTS)
-  set(BASELINE_FILE_ARG "-b ${SCREAM_TEST_DATA_DIR}/p3_run_and_cmp.baseline")
+  if (SCREAM_GENERATE_BASELINES)
+    set(BASELINE_FILE_ARG "-g -b ${SCREAM_BASELINES_DIR}/data/p3_run_and_cmp.baseline")
```
bartgol, line 81: Here, I'm assuming that if `SCREAM_GENERATE_BASELINES=ON`, we are not interested in running tests, but just in creating baselines. I think it's a fair assumption.

components/eamxx/tests/coupled/dynamics_physics/homme_mam4xx_pg2/CMakeLists.txt
```diff
@@ -85,3 +85,32 @@ CompareNCFilesFamilyMpi (
   LABELS dynamics physics mam4xx
   META_FIXTURES_REQUIRED ${FIXTURES_BASE_NAME}_npMPIRANKS_omp1
 )
+
+if (SCREAM_ENABLE_BASELINE_TESTS)
```
bartgol, line 89: I thought about making a cmake function to create the baseline test. I may add it to this PR maybe.

components/eamxx/tests/uncoupled/cosp/CMakeLists.txt
```diff
@@ -20,3 +24,33 @@ configure_file (${CMAKE_CURRENT_SOURCE_DIR}/input.yaml
                 ${CMAKE_CURRENT_BINARY_DIR}/input.yaml)
 configure_file (${CMAKE_CURRENT_SOURCE_DIR}/output.yaml
                 ${CMAKE_CURRENT_BINARY_DIR}/output.yaml)
+
+if (SCREAM_ENABLE_BASELINE_TESTS)
+  # Compare one of the output files with the baselines.
+  # Note: for other tests we do np1-vs-npX bfb tests, which is why one is enough.
+  #       COSP, is not bfb w.r.t. num ranks though. Still, we only check 1 output files against baselines
```
bartgol, line 31: @brhillman side question for you. I had originally added np1-vs-npX tests for cosp_standalone, but they failed, and I remembered we found this with PEM/ERP tests before. What did we decide? Are we planning to make COSP pem-invariant?

components/eamxx/CMakeLists.txt
bartgol: There are also a couple of vars that changed name, and I removed some pointless DEFAULT_xyz vars, using their value directly in the main cache var declaration.

bartgol:
Note: all the `XYZ_baseline_cmp` tests are supposed to fail, since there are no baselines stored. I did however hack my baselines folder on my machine, copying the baselines, and managed to get those tests to pass.

components/eamxx/scripts/test_all_scream.py
```diff
@@ -872,25 +870,46 @@ def generate_baselines(self, test, commit):
...
+            test.missing_baselines = False
+
+            # Read list of nc files to copy to baseline dir
+            with open(test_dir/"data/baseline_list","r") as fd:
+                ncfiles = fd.read().splitlines()
+
+                for fn in ncfiles:
+                    # In case appending to the file leaves an empty line at the end
+                    src = Path(fn)
+                    dst = Path(baseline_dir) / "data" / src.name
```
jgfouca, line 907: baseline_dir should already be a Path.

jgfouca (changes requested):
Some minor nits. Looks good overall. I'm happy to have better baseline coverage for eamxx.

bartgol:
@jgfouca This PR fails the scripts-test, I believe because the scripts used for testing are from the branch, but when it fake generates baselines it switches to origin/master. Now, the baseline generation in this branch generates a file called `$tas_build_dir/$test/data/baseline_list`, which test-all-scream uses to copy nc files from different places in the bld tree back to the baseline dir. Since origin/master does not create this file, tas craps out.

I think we can merge (assuming everything else is fine), and monitor if scripts tests work overnight. If they still crap out, I can investigate what's happening.

jgfouca:
Go for it.

bartgol:
Among the baseline_cmp tests, only the AD tests failed, due to lack of baselines, while all the run_and_cmp tests passed. The scripts tests are expected to fail, b/c of incompatibility between the scripts in this branch and master. Given Jim's approval, I'll merge.

PR #2753, opened 2024-03-13
Some changes to test-all-scream and scripts-tests

bartgol:
This PR does a few changes

For test-all-scream:
- Remove `-f`/`--fast-fail`. This option was never used. It only matters if we have 2+ builds _not_ built in parallel. This likely doesn't happen anymore
- Remove `-b`/`--baseline-ref` option. Instead, we always generate from current HEAD, _except_ for integration tests, when we temporarily switch to `origin/master`.
- Remove `-k`/`--keep-tree`. Instead, always allow to test from a dirty tree (create temp commit if we need to switch to `origin/master` to create baselines when doing `-i` tests)
- Request to use `-g` to generate baselines. If not passed, and baselines are not found, we error out. This avoids overwriting by accident local baselines in `ctest-build/baselines`. Using `-i` or `-u` will automatically add `-g`.
- Add "LOCAL" as possible choice for `--baseline-dir`. This is resolved to `ctest-build/baselines`.
- When not specifying a baseline-dir, we default to:
	- "LOCAL": if `-g` was used but not `-i`
	- "AUTO": otherwise
- When running with `-g` without `--baseline-dir` (i.e., LOCAL is used), always assume baselines are expired, so they will be regenerated.

For scripts-tests:
- Do not build ekat when SCREAM_FAKE_ONLY is defined in the env. The scripts tests don't really care about running an executable via ekat's `ekat_catch_main` main. All they need is an executable that does what expected (e.g., print a msg to screen, pass, fail, ...). Running `time ./scripts-tests -m local -f TestTestAllScream.test_baseline_handling` on my laptop with this branch gives ~25s, compared to 5m25s on master.
- Propagate changes to test-all-scream to this script

### DIFF
--- a/components/eamxx/scripts/git_utils.py
+++ b/components/eamxx/scripts/git_utils.py
@@ -193,7 +193,7 @@ def get_git_toplevel_dir(repo=None):
     return output if stat == 0 else None
 
 ###############################################################################
-def cleanup_repo(orig_branch, orig_commit, repo=None, dry_run=False):
+def cleanup_repo(orig_branch, orig_commit, has_backup_commit=False, repo=None, dry_run=False):
 ###############################################################################
     """
     Discards all unstaged changes, as well as untracked files
@@ -217,4 +217,10 @@ def cleanup_repo(orig_branch, orig_commit, repo=None, dry_run=False):
     # NOTE: if you reset the branch, don't forget to re-update the modules!!
     if curr_commit != orig_commit and not dry_run:
         run_cmd_no_fail("git reset --hard {}".format(orig_commit), from_dir=repo)
+        if has_backup_commit:
+            # This can happen if we ran an integration test with a dirty repo.
+            # test_all_scream will create a temporary backup commit, which we
+            # need to undo, but leaving the changed files in the workspace.
+            # So DON'T add --hard to this call!
+            run_cmd_no_fail("git reset {HEAD~1}", from_dir=repo)
         update_submodules(repo=repo)
--- a/components/eamxx/scripts/git_utils.py
+++ b/components/eamxx/scripts/git_utils.py
@@ -137,6 +137,17 @@ def merge_git_ref(git_ref, repo=None, verbose=False, dry_run=False):
             print_last_commit()
 
 ###############################################################################
+def create_backup_commit (repo=None, dry_run=False):
+###############################################################################
+
+    bkp_cmd = "git add -A && git commit -m 'WARNING: test-all-scream backup commit'"
+    if dry_run:
+        print (f"Woudl run: {bkp_cmd}")
+    else:
+        run_cmd_no_fail(bkp_cmd, from_dir=repo)
+        expect(is_repo_clean(repo=repo), "Something went wrong while performing the backup commit")
+
+###############################################################################
 def print_last_commit(git_ref=None, repo=None, dry_run=False):
 ###############################################################################
     """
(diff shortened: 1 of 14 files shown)


PeterCaldwell:
This seems like an appropriate PR to add the "coupled/uncoupled" directories => "multiproc/single-proc" (or whatever you prefer) to?

bartgol:
> This seems like an appropriate PR to add the "coupled/uncoupled" directories => "multiproc/single-proc" (or whatever you prefer) to?

I don't think so. There are other things that could go wrong with this PR regarding to testing, so I don't want to throw in a folder reorg in the mix. I can do that PR soon-ish (there are a couple more testing PRs in line).

components/eamxx/scripts/scripts-tests
```diff
@@ -518,20 +520,12 @@ class TestTestAllScream(TestBaseOuter.TestBase):
...
         """
         if self._full and self._machine == "mappy":
-            spread_test_opts = "-x -e SCREAM_TEST_THREAD_SPREAD=True -c EKAT_TEST_LAUNCHER_MANAGE_RESOURCES=True -c EKAT_MPIRUN_EXE=mpiexec -c EKAT_MPI_EXTRA_ARGS='-bind-to core' -c EKAT_MPI_NP_FLAG='--map-by' -c EKAT_MPI_THREAD_FLAG='' -m $machine -b HEAD -k -t dbg"
+            spread_test_opts = "-x -e SCREAM_TEST_THREAD_SPREAD=True -e SCREAM_TEST_RANK_SPREAD=True -c EKAT_TEST_LAUNCHER_MANAGE_RESOURCES=True -c EKAT_MPIRUN_EXE=mpiexec -c EKAT_MPI_EXTRA_ARGS='-bind-to core' -c EKAT_MPI_NP_FLAG='--map-by' -c EKAT_MPI_THREAD_FLAG='' -m $machine -t dbg"
 
             cmd = self.get_cmd(f"./test-all-scream {spread_test_opts} --ctest-parallel-level=40 ", self._machine, dry_run=False)
             output = run_cmd_assert_result(self, cmd, from_dir=TEST_DIR)
             test_omp_spread(self, output, 40)
 
-            cmd = self.get_cmd(f"./test-all-scream {spread_test_opts} --ctest-parallel-level=40 ", self._machine, dry_run=False)
```
bartgol, line 527: Long time ago, we factored `SCREAM_TEST_XYZ_SPREAD` out, but didn't realize that 1) the 4 tests below were now two equal pairs of tests and 2) we did not add the rank spread in the test opts string.

components/eamxx/scripts/test-all-scream
```diff
@@ -64,17 +64,14 @@ OR
     parser.add_argument("-p", "--parallel", action="store_true",
                         help="Launch the different build types stacks in parallel")
 
-    parser.add_argument("-f", "--fast-fail", action="store_true",
```
bartgol, line 67: With `-f` gone, I was tempted to add `-f` as shortcut for `--force-baseline-regen`...

components/eamxx/scripts/git_utils.py
```diff
@@ -136,6 +136,17 @@ def merge_git_ref(git_ref, repo=None, verbose=False, dry_run=False):
             print ("git ref {} successfully merged.".format(git_ref))
             print_last_commit()
 
+###############################################################################
+def create_backup_commit (repo=None, dry_run=False):
+###############################################################################
+
+    bkp_cmd = "git add -A && git commit -m 'WARNING: test-all-scream backup commit'"
+    if dry_run:
+        print (f"Woudl run: {bkp_cmd}")
```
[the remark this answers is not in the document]
bartgol: I'll fix it in the baselines PR, since this is ready for merging.

PR #2754, opened 2024-03-13
Support different I/O types for history and restart output

jayeshkrishna:
Adding support for using different I/O types (pnetcdf, adios, netcdf) for history and restart output.

The I/O type for restart output is picked up from CIME for E3SM runs, while the I/O type for the history output is controlled via a new YAML key "iotype" in the output YAML configuration file.

This PR also updates SCORPIO to 1.6.2 and CIME to cime6.0.226

### DIFF
--- a/cime_config/customize/case_post_run_io.py
+++ b/cime_config/customize/case_post_run_io.py
@@ -74,6 +74,12 @@ def _convert_adios_to_nc(case):
     # Load the environment
     case.load_env(reset=True)
 
+    # Reset MPICH/MPI GPU support, if enabled
+    is_mpich_gpu_enabled = os.environ.get('MPICH_GPU_SUPPORT_ENABLED')
+    if int(0 if is_mpich_gpu_enabled is None else is_mpich_gpu_enabled) == 1:
+      logger.info("Resetting support for GPU in MPICH/MPI library (since its not used by the tool)")
+      os.environ['MPICH_GPU_SUPPORT_ENABLED'] = str(0);
+
     run_func = lambda: run_cmd(cmd, from_dir=rundir)[0]
 
     # Run the modified case
--- a/cime_config/machines/config_machines.xml
+++ b/cime_config/machines/config_machines.xml
@@ -1472,6 +1472,10 @@
       <env name="OMP_PROC_BIND">spread</env>
       <env name="OMP_PLACES">threads</env>
     </environment_variables>
+
+    <environment_variables compiler="crayclang-scream" mpilib="mpich">
+      <env name="ADIOS2_ROOT">$SHELL{if [ -z "$ADIOS2_ROOT" ]; then echo /lustre/orion/cli115/world-shared/frontier/3rdparty/adios2/2.9.1/cray-mpich-8.1.26/crayclang-scream-14.0.0; else echo "$ADIOS2_ROOT"; fi}</env>
+    </environment_variables>
   </machine>
 
   <!-- Skylake nodes of Stampede2 at TACC -->
--- a/cime_config/machines/config_workflow.xml
+++ b/cime_config/machines/config_workflow.xml
@@ -37,7 +37,7 @@
       <template>template.case.test</template>
       <prereq>$BUILD_COMPLETE and $TEST</prereq>
     </job>
-    <job name="case.post_run_io">
+    <!--job name="case.post_run_io">
       <template>template.post_run_io</template>
       <dependency>case.run</dependency>
       <prereq>case.get_value("PIO_TYPENAME_ATM") == 'adios' or \
@@ -53,11 +53,12 @@
       <runtime_parameters>
         <walltime>0:30:00</walltime>
       </runtime_parameters>
-    </job>
+    </job-->
     <job name="case.st_archive">
       <template>template.st_archive</template>
       <!-- If DOUT_S is true and case.run (or case.test) exits successfully then run st_archive-->
-      <dependency>(case.run and case.post_run_io) or case.test</dependency>
+      <!--dependency>(case.run and case.post_run_io) or case.test</dependency-->
+      <dependency>case.run or case.test</dependency>
       <prereq>$DOUT_S</prereq>
       <runtime_parameters>
         <task_count>1</task_count>
--- a/components/eamxx/cmake/tpls/Scorpio.cmake
+++ b/components/eamxx/cmake/tpls/Scorpio.cmake
@@ -42,6 +42,17 @@ macro (CreateScorpioTargets)
     option (PIO_ENABLE_FORTRAN "Enable the Fortran library builds" ON)
 
     add_subdirectory (${E3SM_EXTERNALS_DIR}/scorpio ${CMAKE_BINARY_DIR}/externals/scorpio)
+
+    set (SCORPIO_Fortran_INCLUDE_DIRS ${CMAKE_BINARY_DIR}/externals/scorpio/src/flib CACHE INTERNAL "SCORPIO Fortran include dirs")
+    set (C_INCLUDE_DIRS "${E3SM_EXTERNALS_DIR}/scorpio/src/clib")
+    list(APPEND C_INCLUDE_DIRS "${CMAKE_BINARY_DIR}/externals/scorpio/src/clib")
+    set (SCORPIO_C_INCLUDE_DIRS "${C_INCLUDE_DIRS}" CACHE INTERNAL "SCORPIO C include dirs")
+
+    # Add GPTL from SCORPIO
+    if (NOT GPTL_PATH)
+      set (GPTL_PATH ${E3SM_EXTERNALS_DIR}/scorpio/src/gptl CACHE INTERNAL "Path to GPTL library")
+    endif ()
+
     EkatDisableAllWarning(pioc)
     EkatDisableAllWarning(piof)
     EkatDisableAllWarning(gptl)
--- a/components/eamxx/src/share/io/scorpio_input.cpp
+++ b/components/eamxx/src/share/io/scorpio_input.cpp
@@ -351,7 +351,7 @@ void AtmosphereInput::finalize()
 void AtmosphereInput::init_scorpio_structures() 
 {
   std::string iotype_str = m_params.get<std::string>("iotype", "default");
-  int iotype = str2iotype(iotype_str);
+  int iotype = scorpio::str2iotype(iotype_str);
 
   scorpio::register_file(m_filename,scorpio::Read,iotype);
 
--- a/components/eamxx/src/share/io/scorpio_input.cpp
+++ b/components/eamxx/src/share/io/scorpio_input.cpp
@@ -350,7 +350,10 @@ void AtmosphereInput::finalize()
 /* ---------------------------------------------------------- */
 void AtmosphereInput::init_scorpio_structures() 
 {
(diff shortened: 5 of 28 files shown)


cime_config/customize/case_post_run_io.py
```diff
@@ -74,6 +74,12 @@ def _convert_adios_to_nc(case):
     # Load the environment
     case.load_env(reset=True)
 
+    # Reset MPICH/MPI GPU support, if enabled
+    is_mpich_gpu_enabled = os.environ.get('MPICH_GPU_SUPPORT_ENABLED')
+    if int(0 if is_mpich_gpu_enabled is None else is_mpich_gpu_enabled) == 1:
+      logger.info("Resetting support for GPU in MPICH/MPI library (since its not used by the tool)")
+      os.environ['MPICH_GPU_SUPPORT_ENABLED'] = str(0);
+
```
jayeshkrishna, line 82: A CIME config change.
This change is not in E3SM master yet, but we will be adding this change there soon

cime_config/machines/config_machines.xml
```diff
@@ -1295,6 +1295,10 @@
       <env name="OMP_PROC_BIND">spread</env>
       <env name="OMP_PLACES">threads</env>
     </environment_variables>
+
+    <environment_variables compiler="crayclang-scream" mpilib="mpich">
+      <env name="ADIOS2_ROOT">$SHELL{if [ -z "$ADIOS2_ROOT" ]; then echo /lustre/orion/cli115/world-shared/frontier/3rdparty/adios2/2.9.1/cray-mpich-8.1.26/crayclang-scream-14.0.0; else echo "$ADIOS2_ROOT"; fi}</env>
+    </environment_variables>
```
jayeshkrishna, line 1301: A CIME config change
This change also needs to be ported to E3SM master too

cime_config/machines/config_workflow.xml
```diff
@@ -53,11 +53,12 @@
...
         <walltime>0:30:00</walltime>
       </runtime_parameters>
-    </job>
+    </job-->
     <job name="case.st_archive">
       <template>template.st_archive</template>
       <!-- If DOUT_S is true and case.run (or case.test) exits successfully then run st_archive-->
-      <dependency>(case.run and case.post_run_io) or case.test</dependency>
+      <!--dependency>(case.run and case.post_run_io) or case.test</dependency-->
+      <dependency>case.run or case.test</dependency>
```
jayeshkrishna, line 61: A CIME config change
This is a SCREAM repo only change for now. This change disables the ADIOS to NetCDF conversion job from running after the E3SM job completes.

We will need to add a way to disable it via XML changes in E3SM.

components/eamxx/src/share/io/scorpio_input.cpp
```diff
@@ -350,7 +350,13 @@ void AtmosphereInput::finalize()
 /* ---------------------------------------------------------- */
 void AtmosphereInput::init_scorpio_structures() 
 {
-  scorpio::register_file(m_filename,scorpio::Read);
+  int iotype = scorpio::str2iotype("default");
+  if(m_params.isParameter("iotype")){
```
bartgol, line 354: If you want, you can be more concise:
```suggestion
auto iotype_str = m_params.get<std::string>("iotype","default");
int iotype = scorpio::str2iotype(iotype_str);
```

components/eamxx/src/share/io/scorpio_input.hpp
```diff
@@ -2,6 +2,7 @@
 #define SCREAM_SCORPIO_INPUT_HPP
 
 #include "share/io/scream_scorpio_interface.hpp"
+#include "share/io/scream_io_utils.hpp"
```
bartgol, line 5: I don't think this include is really needed. If it's needed in the cpp, you can just include it in the cpp.

components/eamxx/src/share/io/scream_scorpio_interface.cpp
```diff
@@ -94,8 +94,8 @@ void eam_pio_finalize() {
   eam_pio_finalize_c2f();
 }
 /* ----------------------------------------------------------------- */
-void register_file(const std::string& filename, const FileMode mode) {
-  register_file_c2f(filename.c_str(),mode);
+void register_file(const std::string& filename, const FileMode mode, const int iotype) {
```
bartgol, line 97: Up to you, but since you often call this with `iotype=IoType::DefaultIoType`, you could consider adding the default value in the fcn signature, so that you don't need to add `, scorpio::IOType::DefaultIOType` to all `register_file` calls..

bartgol (dismissed):
This looks great! I have a couple of comments, but nothing too important.

components/eamxx/src/share/io/scorpio_input.cpp
jayeshkrishna: Thanks, your suggestion looks better/concise. I will make that change.

components/eamxx/src/share/io/scorpio_input.hpp
jayeshkrishna: ok, will remove it (I think it must have been added while I was modifying the code)

components/eamxx/src/share/io/scream_scorpio_interface.cpp
jayeshkrishna: I think its a good idea to add the default value here (On the other hand forcing users to specify the args can prevent copy-paste errors :) ). I will make that change

bartgol: Yeah, it's not an obvious choice. As I said, up to you. If you think in the vast majority of cases the "default" value is fine, then adding defaults to the signature makes sense. But if the cost of using default when you shouldn't is very high, then we can just force users to always pass it.

jayeshkrishna: I will modify the code to use default args since there are many instances where the default io type is used (including tests etc)

jayeshkrishna:
@bartgol : This PR should now have all the changes to build/run with CIME and non-CIME/standalone builds. Can you verify that the branch now builds on mappy?

bartgol:
@jayeshkrishna we'll know soon enough, when the Autotester builds it.

bartgol:
I think the failure here is due to us not using pnetcdf in test-all-scream:
```
10:45:32 PIO: FATAL ERROR: Aborting... An error occured, Creating file (homme_shoc_cld_p3_rrtmgp_output.INSTANT.nsteps_x4.np1.2021-10-12-45000.nc) failed. Invalid iotype (PIO_IOTYPE_PNETCDF:1) specified. Available iotypes are : PIO_IOTYPE_NETCDF (2), PIO_IOTYPE_NETCDF4C (3), PIO_IOTYPE_NETCDF4P (4). Bad IO type (err=-500). Aborting since the error handler was set to PIO_INTERNAL_ERROR... (/home/e3sm-jenkins/jenkins-ws/workspace/SCREAM_PullRequest_Autotester_Mappy/5084/scream/externals/scorpio/src/clib/pioc_support.c: 2756)
```

@jayeshkrishna I'm going to ninja submit a PR to add pnetcdf support in our AT runs, so that we can then test this PR correctly. The good news is that it configured correctly, so the GPTL cmake issue is gone.

jayeshkrishna:
@bartgol : Do you know what is causing the failures on mappy for this PR (Do you know if all the tests PASS with the latest master on mappy?)?

bartgol:
So bizarre the standalone error. PIO craps out b/c the "units" attribute is not found attached to the "time" var. And that's true, ncdump shows no "units" for "time". BUT, for the debug build, the attribute _is_ there. It may be another issue like the one we had for update time (i.e., we need to "use" the return code from pio)

bartgol:
Yup, that was it. I went through our scorpio interface file, and fixed all remaining cases where the return error code was not used.

@jayeshkrishna  I'm attaching the patch here.
[patch.txt](https://github.com/E3SM-Project/scream/files/14792206/patch.txt)

jayeshkrishna:
ok, I will rebase the branch with the patch above. Thanks!

bartgol:
The v1 fails give the same error I was observing in the PR where I switch to pioc, which was also updating scorpio to v1.5:
```
ERROR: (seq_infodata_Check) : invalid continue restart case name =
```
This suggests that the root of the error may come from a scorpio update. I am not sure how/why, but it's the only thing these two PRs have in common...

jayeshkrishna:
This could also be related to the version of gcc on mappy

bartgol:
I added WIP to prevent pointless testing until we upgrade mappy's compiler. Jim said he may get to do it tomorrow.

jgfouca:
@jayeshkrishna , yes, I'm testing E3SM with the new toolchain now.

jayeshkrishna:
Waiting on https://github.com/E3SM-Project/scream/pull/2784 (Will rebase the branch to latest master after this PR goes in)

components/eamxx/cmake/tpls/Scorpio.cmake
```diff
@@ -42,6 +42,17 @@ macro (CreateScorpioTargets)
...
+    set (SCORPIO_Fortran_INCLUDE_DIRS ${CMAKE_BINARY_DIR}/externals/scorpio/src/flib CACHE INTERNAL "SCORPIO Fortran include dirs")
+    set (C_INCLUDE_DIRS "${E3SM_EXTERNALS_DIR}/scorpio/src/clib")
+    list(APPEND C_INCLUDE_DIRS "${CMAKE_BINARY_DIR}/externals/scorpio/src/clib")
+    set (SCORPIO_C_INCLUDE_DIRS "${C_INCLUDE_DIRS}" CACHE INTERNAL "SCORPIO C include dirs")
+
+    # Add GPTL from SCORPIO
+    if (NOT GPTL_PATH)
+      set (GPTL_PATH ${E3SM_EXTERNALS_DIR}/scorpio/src/gptl CACHE INTERNAL "Path to GPTL library")
+    endif ()
+
```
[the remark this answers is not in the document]
jayeshkrishna: Yeah, eventually we will move to just using find_library(scorpio...) mode. The basic pieces are there in SCORPIO now, we will start using it in E3SM/EAMXX in upcoming PRs.

(PS: Note that this hack is needed due to the way EAMXX currently adds SCORPIO source for standalone builds)

jayeshkrishna:
@bartgol / @jgfouca : Are the diffs on mappy ok (I am assuming the DIFFs in the latest round of tests were due to the upstream merge - PR 2784)? How can I restart the tests on this PR?

bartgol:
@jayeshkrishna there was a problem with the jenkins job that blessed the AT baselines (it still used gnu9 as compiler), so it didn't bless the baselines. I fixed it and I'm re-running it. I also added the RETEST label so that the testing will be triggered again.

bartgol (approved):

AaronDonahue:
@jayeshkrishna , with the upstream merge in place and this PR passing the AT and approved is it safe to say this is ready to go in?

jayeshkrishna:
Yes, this PR should be ready to merge.

dqwu (approved):

bartgol:
I added automerge, since I believe there was no reason to hold off the merge. The AT will test once more (since master has changed since last PASS), then will merge.

PR #2758, opened 2024-03-13
Run e3sm gh action only if repo name is e3sm

bartgol:
Ideally, if a gh action file starts with `e3sm-`, then it should only run on the `e3sm` repo (and have a check inside that that's the case). On top of that, the singularity tests seem to have little meaning for the SCREAM fork, since we virtually never modify code outside the eamxx folder, and in the rare instances where we do, it's probably in EAM, which _does_ get tested via SCREAMv0 label.

### DIFF
--- a/.github/workflows/e3sm-gh-pages.yml
+++ b/.github/workflows/e3sm-gh-pages.yml
@@ -15,7 +15,7 @@ concurrency:
   
 jobs:
   Build-and-Deploy-docs:
-    if: ${{ github.event.repository.name != 'scream' }}
+    if: ${{ github.event.repository.name == 'e3sm' }}
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


.github/workflows/e3sm-gh-pages.yml
```diff
@@ -15,7 +15,7 @@ concurrency:
   
 jobs:
   Build-and-Deploy-docs:
-    if: ${{ github.event.repository.name != 'scream' }}
+    if: ${{ github.event.repository.name == 'e3sm' }}
```
bartgol, line 18: Better compare to `e3sm`, or else this would run on any non-scream fork of e3sm...

bartgol: Ah, good point. Let me check

bartgol: I think they are not, as stated [here](https://docs.github.com/en/actions/learn-github-actions/expressions):

```
GitHub ignores case when comparing strings.
```

PR #2762, opened 2024-03-18
IOP fixes (shoc cell length and iop file read)

tcclevenger:
Two fixes:
1. SHOC cell length computation should be converted to meters.
2. IOP file read in EAMxx was interpolating endpoints different from EAM.

### DIFF
--- a/cime_config/tests.py
+++ b/cime_config/tests.py
@@ -627,7 +627,7 @@ _TESTS = {
     "e3sm_scream_v1_dp-eamxx" : {
         "time"  : "01:00:00",
         "tests" : (
-            "ERS_Ln22.ne30_ne30.F2010-SCREAMv1-DP-DYCOMSrf01", # 225 phys cols, roughly size of ne2
+            "ERS_Ln22.ne30_ne30.F2010-SCREAMv1-DP-DYCOMSrf01.scream-ntasks-16", # 225 phys cols, roughly size of ne2
             )
     },
 
--- a/cime_config/tests.py
+++ b/cime_config/tests.py
@@ -627,7 +627,7 @@ _TESTS = {
     "e3sm_scream_v1_dp-eamxx" : {
         "time"  : "01:00:00",
         "tests" : (
-            "ERS_Ln22.ne30_ne30.F2010-SCREAMv1-DP-DYCOMSrf01.scream-ntasks-16", # 225 phys cols, roughly size of ne2
+            "ERS_P16_Ln22.ne30_ne30.F2010-SCREAMv1-DP-DYCOMSrf01", # 225 phys cols, roughly size of ne2
             )
     },
 
--- a/cime_config/tests.py
+++ b/cime_config/tests.py
@@ -635,7 +635,7 @@ _TESTS = {
     # should be fast, so we limit it to low res and add some thread tests
     # specifically for mappy.
     "e3sm_scream_v1_at" : {
-        "inherit" : ("e3sm_scream_v1_lowres"),
+        "inherit" : ("e3sm_scream_v1_lowres", "e3sm_scream_v1_dp-eamxx"),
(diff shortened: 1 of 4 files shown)


components/eamxx/src/control/intensive_observation_period.cpp
```diff
@@ -725,18 +725,14 @@ read_iop_file_data (const util::TimeStamp& current_ts)
...
-                               KOKKOS_LAMBDA (const int ilev) {
-            iop_field_v(ilev) = iop_field_v(model_start);
-          });
-        }
-        if (model_end < total_nlevs) {
-          Kokkos::parallel_for(Kokkos::RangePolicy<>(model_end, total_nlevs),
-                               KOKKOS_LAMBDA (const int ilev) {
-            iop_field_v(ilev) = iop_field_v(model_end-1);
-          });
-        }
```
tcclevenger, line 739: @bogensch Previously, EAMxx was doing 2 things different to EAM for this subset of iop fields (in EAM, fields that have `fill_ends=true`)
i. EAMxx was setting the non-interpolated region to the first/last interpolated value in the IOP field, whereas in EAM we set it to the first/last value in the IOP file.
ii. EAM always overwrites the first and last value in the interpolated region to the IOP file start and end values, whereas EAMxx was only setting end points outside the interpolated region.

Let me know if this is the intended behavior. I've verified the IOP fields now match exactly between EAM and EAMxx.

tcclevenger:
Forgot to add DP test to CIME AT list in last PR. Adding here.

components/eamxx/cime_config/testdefs/testmods_dirs/scream/ntasks/16/shell_commands
```diff
@@ -0,0 +1,2 @@
+./xmlchange NTASKS=16 
+
```
tcclevenger, line 2: @jgfouca @bartgol Is this the best way to change the ntasks for CIME tests in mappy? DP test has 25 elements, so the default mpi tasks on mappy (64) was too large.

cime_config/tests.py
```diff
@@ -627,15 +627,15 @@
     "e3sm_scream_v1_dp-eamxx" : {
         "time"  : "01:00:00",
         "tests" : (
-            "ERS_Ln22.ne30_ne30.F2010-SCREAMv1-DP-DYCOMSrf01", # 225 phys cols, roughly size of ne2
+            "ERS_Ln22.ne30_ne30.F2010-SCREAMv1-DP-DYCOMSrf01.scream-ntasks-16", # 225 phys cols, roughly size of ne2
```
jgfouca, line 630: If you're going to hardcode the test to 16 tasks, you don't need a testmod. Just do:
`ERS_P16_Ln22.ne30_ne30.F2010-SCREAMv1-DP-DYCOMSrf01`

tcclevenger: ahh good to know! Thanks.

components/eamxx/cime_config/testdefs/testmods_dirs/scream/ntasks/16/shell_commands
bartgol: Yes, this is ok. I am not sure, however, how an xmlchange of ntasks plays with ERP/PEM tests. Maybe @jgfouca knows. I think it's ok, but there may be some gotchas for those tests.

cime_config/tests.py
bartgol: Ah, just read this after the prev comment. I was going to suggest the same thing.

components/eamxx/cime_config/testdefs/testmods_dirs/scream/ntasks/16/shell_commands
bartgol: NVM, this is outdated by the last commit.

bartgol (approved):

tcclevenger:
Looked at the output, fails are expected (diffs in baseline tests, and CIME case doesn't have baselines yet, but runs without failing). I'll manually merge re-run mappy baseline.

PR #2767, opened 2024-03-21
test-all-scream: Get scripts-tests working again

jgfouca:
Main change: remove support for dry runs. Dry runs can no longer be supported because CMake has to actually run to create key baseline files in order for TAS to work. We had a lot of test infrastructure based on comparing dry runs, which was never used, so I had to remove all that as well.

Other changes: `-b` is now free to be used as a shortcut for `--baseline-dir`.

### DIFF
--- a/components/eamxx/scripts/scripts-tests
+++ b/components/eamxx/scripts/scripts-tests
@@ -3,14 +3,7 @@
 """
 Script containing python test suite for SCREAM test
 infrastructure. This suite should be run to confirm overall
-correctness. You should run this test once in generation mode to
-generate baseline results using your reference commit (common
-ancestor) and once in comparison mode to compare against these
-baselines using your development commit. Baseline and compare runs
-will use dry-run modes so we are only comparing hypothetical shell
-commands, not actually running them.
-
-You can also do a full run which will actually execute the commands.
+correctness.
 
 If you are on a batch machine, it is expected that you are on a compute node.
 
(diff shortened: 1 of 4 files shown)


components/eamxx/scripts/scripts-tests
```diff
@@ -592,7 +534,7 @@ class TestGatherAllData(TestBaseOuter.TestBase):
 ###############################################################################
 
     CMDS_TO_TEST = [
-        "./gather-all-data './scripts/test-all-scream -m $machine' -l -m $machine",
+        "./gather-all-data 'echo -m $machine' -l -m $machine",
```
bartgol, line 537: Nice! Using a simple cmd like echo is much better to test the outer script. No need to add the inner script complexity to the mix.

components/eamxx/scripts/scripts-tests
```diff
@@ -667,12 +592,6 @@ OR
     parser.add_argument("-m", "--machine",
                         help="Provide machine name. This is required for full (not dry) runs")
 
-    parser.add_argument("-g", "--generate", action="store_true",
```
bartgol, line 670: Does it make sense to keep `-f` now that `-g` and `-c` have been removed? Seems like we do run full regardless now, no? I mean, what's the point of running without `-f`?

bartgol (approved):

bartgol:
I rm-ed automerge in case you want to act on my last comment. But I approved in case you want to pass on it.

components/eamxx/scripts/scripts-tests
jgfouca: We still need to distinguish between "full" and "jenkins" runs.

jgfouca: Actually, we could just assume no `-j` implies `-f`. Let me look

jgfouca: There's still some value here since, without `-f` or `-j`, it just runs very quick pylint and doctests, which is useful.

PR #2770, opened 2024-03-25
Add compiler flags to perlmutter cmake to enable use of mlcorrection

AaronDonahue:
This commit adds the cmake flags to turn on ml-correction, already used on Livermore Computing machines, to the perlmutter machine files.  I have manually tested that this compiles and works on perlmutter.


AaronDonahue:
There are some issues w/ permissions that need to be resolved before this can be unWIP

AaronDonahue:
Closing pending some work on ML

PR #2774, opened 2024-03-27
Update EKAT

tcclevenger:
Update ekat.


bartgol (approved):

jgfouca (approved):

bartgol:
We never blessed the AT v1 cases after merging the p3 non-bfb PR. I am blessing now, and we can verify on the next run that everything is fine.

PR #2776, opened 2024-03-30
Enhance handling of tags in FieldLayout

bartgol:
This PR should make it easier to add new field layouts without having to add new FieldTag enum's.

The main modifications are:
- Store tag names inside the layout. By default, we use `e2str(tag)` for each tag, but the user can overwrite with specific values.
- Remove Rad-related FieldTag values. Radiation will use `CMP`, setting the tag name to "swband"/"lwband" and the likes
- Minor change in some layout methods. The `strip_dim` method now modifies the current layout, so one has to do `fl.clone().strip_dim(CMP)` to obtain a copy of `fl` with the dimension stripped.
- Some fixes for handling tensor fields

### DIFF
--- a/components/eamxx/src/physics/mam/impl/compute_o3_column_density.cpp
+++ b/components/eamxx/src/physics/mam/impl/compute_o3_column_density.cpp
@@ -3,19 +3,18 @@ namespace scream::impl {
 KOKKOS_INLINE_FUNCTION
 void compute_o3_column_density(const ThreadTeam& team, const haero::Atmosphere& atm,
                                const mam4::Prognostics &progs, ColumnView o3_col_dens) {
-  constexpr int nabscol = mam4::gas_chemistry::nabscol;     // number of absorbing densities
   constexpr int gas_pcnst = mam4::gas_chemistry::gas_pcnst; // number of gas phase species
   constexpr int nfs = mam4::gas_chemistry::nfs;             // number of "fixed species"
-  constexpr Real mwdry = 1.0/haero::Constants::molec_weight_dry_air;
+  // constexpr Real mwdry = 1.0/haero::Constants::molec_weight_dry_air;
 
   Real o3_col_deltas[mam4::nlev+1] = {}; // o3 column density above model [1/cm^2]
   // NOTE: if we need o2 column densities, set_ub_col and setcol must be changed
   Kokkos::parallel_for(Kokkos::TeamThreadRange(team, atm.num_levels()), [&](const int k) {
 
-    Real temp = atm.temperature(k);
-    Real pmid = atm.pressure(k);
+    // Real temp = atm.temperature(k);
+    // Real pmid = atm.pressure(k);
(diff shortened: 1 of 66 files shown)


components/eamxx/src/physics/mam/impl/compute_o3_column_density.cpp
```diff
@@ -3,28 +3,27 @@ namespace scream::impl {
 KOKKOS_INLINE_FUNCTION
 void compute_o3_column_density(const ThreadTeam& team, const haero::Atmosphere& atm,
                                const mam4::Prognostics &progs, ColumnView o3_col_dens) {
-  constexpr int nabscol = mam4::gas_chemistry::nabscol;     // number of absorbing densities
   constexpr int gas_pcnst = mam4::gas_chemistry::gas_pcnst; // number of gas phase species
   constexpr int nfs = mam4::gas_chemistry::nfs;             // number of "fixed species"
-  constexpr Real mwdry = 1.0/haero::Constants::molec_weight_dry_air;
+  // constexpr Real mwdry = 1.0/haero::Constants::molec_weight_dry_air;
```
bartgol, line 8: All these commented lines are to get rid of compiler warnings due to unused stuff.

bartgol: Also, sometimes I removed the line, sometimes I commented, the rational being that
- if the  variable would be used in a function down below, but said function call is commented since it's not yet ported, then I commented the line declaring the var
- if the variable is _never_ used down below, then I removed the line altogether.

components/eamxx/tests/multi-process/dynamics_physics/homme_shoc_cld_p3_rrtmgp_pg2/CMakeLists.txt
```diff
@@ -10,7 +10,7 @@ CreateDynamicsLib("theta-l_kokkos"  4   72   10)
 # Create the test
 CreateADUnitTest(${TEST_BASE_NAME}
   LABELS dynamics tms shoc cld p3 rrtmgp physics pg2
-  LIBS cld_fraction nudging tms shoc p3 scream_rrtmgp ${dynLibName} diagnostics
+  LIBS cld_fraction tms shoc p3 scream_rrtmgp ${dynLibName} diagnostics
```
tcclevenger, line 13: What is the reason for this change?

tcclevenger (approved):
Just a question, but not blocking merging.

components/eamxx/tests/multi-process/dynamics_physics/homme_shoc_cld_p3_rrtmgp_pg2/CMakeLists.txt
bartgol: The test does not use nudging. So if you want to build just this test (while debugging), you would have also had to build nudging, for no reason.

PR #2781, opened 2024-04-03
Fix rrtmgp valgrind tests

jgfouca:
They were handing due to YAKL pool running out, so make it bigger by setting GATOR_INITIAL_MB to 4GB in both the standalone and cime env for mappy.


AaronDonahue:
@jgfouca ,did you see @mrnorman 's comment on Slack?  I think he suggested we should disable the pool altogether for the valgrind test.

AaronDonahue (approved):
I'm going to approve this now so that we can get `valgrind` coverage running again as soon as tonight.

PR #2782, opened 2024-04-03
Fix standalone eamxx build on pm-gpu

jgfouca:
The critical `-arch sm_80` flag was not being added for the compilation of `scream_rrtmgp_interface.cpp` on pm-gpu. This PR sets YAKL_CUDA_FLAGS based on KOKKOS_CUDA_OPTIONS, which I think should be a reliable way to get the correct cuda flags assuming the cmake machine files are correct.

Also, add a kokkos dependence to stuff in eamxx/src/physics/rrtmgp. This will prep for the transition to kokkos and means we don't have to add hacks in order to use kokkos_launch_compiler.


jgfouca:
The rrtmgp baseline diff is expected, merging.

PR #2783, opened 2024-04-04
Adding support for PIO_REARR_ANY

jayeshkrishna:
Adding support for new SCORPIO rearranger, PIO_REARR_ANY.

This rearranger lets the SCORPIO library decide on the optimal
rearranger to use. The current default settings for this rearranger
gives us the best performance for ne1024 runs (when using the
PnetCDF I/O type for history output and ADIOS I/O type for
restart output)

### DIFF
--- a/driver-mct/cime_config/config_component.xml
+++ b/driver-mct/cime_config/config_component.xml
@@ -2692,10 +2692,10 @@
 
   <entry id="PIO_REARRANGER">
     <type>integer</type>
-    <valid_values>1,2</valid_values>
+    <valid_values>1,2,3</valid_values>
     <group>run_pio</group>
     <file>env_run.xml</file>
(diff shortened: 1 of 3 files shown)


jayeshkrishna:
This PR needs to go in after https://github.com/E3SM-Project/scream/pull/2754

bartgol:
May I suggest the word `auto` rather than `any`? I think it conveys better the meaning that "it will be chosen automatically". It's also a more common keyword in others libraries for this kind of behavior. 

That said, you guys are the Scorpio devs, we'll use whatever you guys like best.

bartgol (approved):

dqwu (approved):

bartgol:
Merging, since I need to rebase on master to bring up to speed the PR that rids us of Fortran in the IO interface

PR #2784, opened 2024-04-04
Jgfouca/upstream merge 2024 04 04

jgfouca:
Also, update all eamxx mappy stuff to use new gnu-11.2 toolchain.


AaronDonahue (approved):
@jgfouca , this looks good.  I see that the FAILS are due to the nonBFB changes.  Can you confirm these are expected?  Also, do you know why only the 128 level unit test is failing on `weaver` in the baseline comparison, but not the other physics/dynamics unit tests?

jgfouca:
Baseline fails are expected, merging.

PR #2785, opened 2024-04-04
Enable Subfield Slices Spanning a Range of indices (Issue #1312)

### LINKED ISSUE #1312: Allow subfield method of Field to extract 2+ slices.
Basically, I would like to be able to do
```
  Field<Real> vel3d(...); // vector field, with 3 components
  auto uv = vel3d.get_components(0,2); // take 2 slices, starting at component 0
  auto w  = vel3d.get_component(2); // take 3rd component
```
Currently, this is not possible. A subfield is always an n-1 dimensional field. So one has to take components u and v separately.

Note: this feaature might actually be more useful for tracers, where some parametrizations might be interested to get a subset of Q, like doing `my_tracers=Q(:, 3:, :)`, in F90 syntax.

mjs271:
This PR adds functionality that allows a user to create a "multi-slice" subfield and gain access to the underlying view. That is, a multi-slice subfield is in contrast to the previous capability that sliced a single entry of a chosen dimension--e.g., `view(:, 7, :)` to use notation common to fortran/matlab, and expands this capability to allow for slicing a chosen dimension across a range of values--e.g., `view(:, 2:7, :)`.

Summary of changes and capabilities:

- Added overloads for `Field::subfield()` that takes the extra argument(s) `index_beg, index_end` to indicate the bounds of the slicing range.
- Added `Field::get_multi_sliced_view()` that provides the view underlying the subfield as a multi-dimensional view with the same number of dimensions as the parent's associated view (as opposed to a single-sliced view that is (N - 1)-dimensional.
  - _Note:_ as implemented, this is only returned as a `LayoutStride` view, rather than `LayoutRight`.

I do need to add a few more tests and will push those changes soon, but I wanted to get this out there as a draft. Any feedback about my approach would be much-appreciated, since this is my first time doing a deep-dive into this code.

Closes #1312


mjs271:
Converting from draft to full PR after adding tests for `f.subfield()` and `sf.get_multi_sliced_view()`. Also cleaned up `ekat::subview()` , added tests over there, and brought the ekat submodule up to date.

AaronDonahue:
>Note: the ekat branch for the PR is over on [my fork](https://github.com/mjs271/EKAT), since I don't yet have push access to E3SM-Project/EKAT

@mjs271 , does this mean we will need a separate EKAT PR to accompany this one?

mjs271:
@AaronDonahue @mahf708 
That's correct--I've got a PR open on ekat, but I'm not a member of that repo, so this PR has the submodule commit pointed at my personal fork. Happy to resolve that however seems best

And, @mahf708, in response to your question, I've it a  little thought, and I don't think it'd be an **_enormous_** undertaking, but I do think it would require a new approach. The reason being that, as far as I know, kokkos allows slicing only with scalars, `{std,Kokkos}::pair<>()`, and `Kokkos::ALL`. As I was working on this code, my thought for the noncontiguous slicing approach was that it'd require taking individual contiguous slices and then concatenating. However, there's a nonzero chance there's another clever approach I'm not thinking of

It also occurred to me that it could be nice to be able to slice multiple dimensions at once, but I'm not sure whether there's a ready use case and also whether that might be approaching too much complexity

AaronDonahue:
@mjs271 , it looks like this is failing the `weaver` test suite for the AT.  Can you take a look?

mjs271:
@AaronDonahue I'll give it a look!

components/eamxx/src/share/field/field_alloc_prop.hpp
```diff
@@ -61,21 +61,37 @@ namespace scream
...
-  int slice_idx  = -1;  // Slice along dimension $dim_idx
-  int dim_extent = -1;  // Extent of dimension $dim_idx
-  bool dynamic   = false; // Whether this is a dynamic subview (slice_idx can change)
+  SubviewInfo() = default;
+  SubviewInfo(const int dim, const int slice, const int extent,
+              const bool is_dynamic)
+      : dim_idx(dim), slice_idx(slice), dim_extent(extent),
+        dynamic(is_dynamic) {}
+  // multi-slice subview across contiguous indices
+  SubviewInfo(const int dim, const int slice_beg, const int slice_end,
```
bartgol, line 70: As I said in another comment, let's remove the `dynamic` option from this overload. I can't see when/why it would be needed.

components/eamxx/src/share/field/field_alloc_prop.hpp
```diff
@@ -61,21 +61,37 @@ namespace scream
...
+              const int extent, const bool is_dynamic)
+      : dim_idx(dim), slice_idx(slice_beg), slice_idx_end(slice_end),
+        dim_extent(extent), dynamic(is_dynamic) {}
+  SubviewInfo(const SubviewInfo&) = default;
+  SubviewInfo& operator=(const SubviewInfo&) = default;
+
+  int dim_idx = -1;    // Dimension along which slicing happened
+  // Slice along dimension $dim_idx if taking single-index slice for subfield
+  // If taking multi-slice subfield, then this is the starting slice index
+  // e.g., slicing (:, slice_idx_end : slice_idx_end, :)
```
bartgol, line 80: This comment seems wrong (why end:end rather than beg:end?). I also don't really understand what it was trying to show.

bartgol: On a related note: do you _need_ to have `slide_idx` store a valid value for multi-slice subfields? You don't store beg=end=slice_idx for the single slice case, after all.

components/eamxx/src/share/field/field_alloc_prop.hpp
```diff
@@ -61,21 +61,37 @@ namespace scream
...
+  int slice_idx_end = -1;
+  int dim_extent = -1; // Extent of dimension $dim_idx
+  bool dynamic =
+      false; // Whether this is a dynamic subview (slice_idx can change)
 };
 
 inline bool operator== (const SubviewInfo& lhs, const SubviewInfo& rhs) {
   return lhs.dim_idx==rhs.dim_idx &&
          lhs.slice_idx==rhs.slice_idx &&
+        //  FIXME: change for multi-slice (throw error to check?)
```
bartgol, line 94: I think we need to fix this check before we merge.

components/eamxx/src/share/field/field_alloc_prop.cpp
```diff
@@ -77,6 +77,56 @@ subview (const int idim, const int k, const bool dynamic) const {
...
+  props.m_scalar_type_size = m_scalar_type_size;
+  props.m_layout =
+      m_layout.clone_with_different_extent(idim, k_end - k_beg + 1);
+
+  // Output is contiguous if either
+  //  - this->m_contiguous=true AND idim==0
+  //  - m_layout.dim(i)==1 for all i<idim
+  //  - m_layout.rank()==1 (we end up with a rank-0 subview)
+  auto is_one = [](int i) { return i == 1; };
+  props.m_contiguous = (m_contiguous and idim == 0) || m_layout.rank() == 1 ||
```
bartgol, line 107: I'm not sure this is right. As of now, multi-slice subfields will always require strided views. If `m_contiguous` is true, then the `get_view` method of `Field` becomes ok to call, which may yield garbage. I would always consider multi-slice views as non-contiguous, until we figure out a way to be smarter in the `Field::get_view` method(s).

components/eamxx/src/share/field/field_impl.hpp
```diff
@@ -205,6 +203,56 @@ auto Field::get_strided_view () const
   return DstView(get_ND_view<HD,DstValueType,1>());
 }
 
+// NOTE: multi-slicing a view is only supported for strided view return type
+// TODO: N doesn't actually need to be passed in, since the subview must have
+// the same rank as its parent
+template <typename DT, int N, HostOrDevice HD>
+auto Field::get_multi_sliced_view() const
```
bartgol, line 210: I don't see this function called anywhere, which means the user must be aware that this field is a mutli-slice subfield of another, and call this function manually. That's probably asking a bit too much to the user. It would be nice if `get_strided_view` called this method internally (or maybe just add bits of this method to `get_strided_view`). 

Right now, `get_strided_view` _assumes_ that the output view is 1d, since we only have the need for strided views for 1d subfields. But that assumption is not really _needed_ (more of a correctness check, since we didn't expect any strided view usage for rank 2+ fields).

In short: let's make `get_strided_view` handle the multi-slice case. That will require allowing rank 2+ views for that method.

components/eamxx/src/share/tests/subfield_tests.cpp
```diff
@@ -0,0 +1,389 @@
+#include <catch2/catch.hpp>
+#include <numeric>
+
+#include "ekat/kokkos/ekat_subview_utils.hpp"
```
bartgol, line 4: Some of these includes are not needed in this test. E.g., `field_manager.hpp` or the ekat pack headers.

components/eamxx/src/share/tests/subfield_tests.cpp
```diff
@@ -0,0 +1,389 @@
...
+TEST_CASE("field", "") {
+  using namespace scream;
+  using namespace ShortFieldTagsNames;
+  using namespace ekat::units;
+
+  auto engine = setup_random_test ();
+  using RPDF = std::uniform_real_distribution<Real>;
+  RPDF pdf(0.01,0.99);
+
+  std::vector<FieldTag> tags = {COL,LEV};
```
bartgol, line 30: These two vectors and `fid` seem to be unused.

components/eamxx/src/share/field/field_impl.hpp
bartgol: It dawned on me that some Field utilities call either `get_view` or `get_strided_view`, depending on whether `FieldAllocProp::contiguous()` returns true or false. My previous comment becomes more important in this regard: moving this case into `get_strided_view` allows to keep the downstream code simpler, since they only have two possible methods to call: `get_view` or `get_strided_view`.

Furthermore, since multi-slice view is strided but may have rank>1, we need to add more if statements to those Field utilities. Namely, for rank 2+, those utilities need to do the same thing we do for rank-1: check if allocation is contiguous, and if so, call `get_view`, otherwise call `get_strided_view`.

components/eamxx/src/share/field/field.hpp
```diff
@@ -241,11 +247,23 @@ class Field {
   Field subfield (const std::string& sf_name, const int idim,
                        const int index, const bool dynamic = false) const;
   Field subfield (const int idim, const int k, const bool dynamic = false) const;
-
+  // subfield fxn to extract multiple slices in a continuous range of indices
```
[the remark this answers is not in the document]
mjs271: haha, I was going for "function"--changed in favor of a more descriptive wording

components/eamxx/src/share/field/field.hpp
```diff
@@ -241,11 +247,23 @@ class Field {
...
+  // e.g., (in matlab syntax) subf = f.subfield(:, 1:3, :)
+  // but NOT subf = f.subfield(:, [1, 3, 4], :)
+  Field subfield (const std::string& sf_name, const ekat::units::Units& sf_units,
```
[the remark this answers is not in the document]
mjs271: Done! I'm assuming this should be removed from `get_components()` as well?

components/eamxx/src/share/field/field_alloc_prop.cpp
```diff
@@ -77,6 +77,56 @@ subview (const int idim, const int k, const bool dynamic) const {
...
+  FieldAllocProp props(m_scalar_type_size);
+  props.m_committed = true;
+  props.m_scalar_type_size = m_scalar_type_size;
+  props.m_layout =
+      m_layout.clone_with_different_extent(idim, k_end - k_beg + 1);
+
+  // Output is contiguous if either
+  //  - this->m_contiguous=true AND idim==0
+  //  - m_layout.dim(i)==1 for all i<idim
+  //  - m_layout.rank()==1 (we end up with a rank-0 subview)
```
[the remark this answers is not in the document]
mjs271: thank you!

components/eamxx/src/share/field/field.hpp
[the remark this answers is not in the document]
bartgol: Yes

components/eamxx/src/share/field/field_alloc_prop.cpp
```diff
@@ -77,6 +77,56 @@ subview (const int idim, const int k, const bool dynamic) const {
...
 }
 
+FieldAllocProp FieldAllocProp::subview(const int idim,
+                                       const int k_beg,
+                                       const int k_end,
+                                       const bool dynamic) const {
+  EKAT_REQUIRE_MSG(
+      is_committed(),
+      "Error! Subview requires alloc properties to be committed.\n");
+  EKAT_REQUIRE_MSG(idim <= m_layout.rank(),
```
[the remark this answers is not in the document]
mjs271: absolutely correct--thank you!

components/eamxx/src/share/field/field_alloc_prop.hpp
mjs271: done!

components/eamxx/src/share/field/field_alloc_prop.hpp
mjs271: Thanks for catching that--`slice_beg` was meant to be removed earlier 🤦‍♂️

components/eamxx/src/share/field/field_impl.hpp
mjs271: Ah, ok! That makes a lot of sense--will knock that out asap

mjs271: I believe we should be just about finished after the recent pushes. All tests passing locally for me, and  I'll need to do a rebase before merge, but otherwise let me know if this approach works for you--thanks!

tcclevenger:
Looks like some conflicts exist.

AaronDonahue:
@mjs271 Can you rebase this PR and remove the WIP label?

mjs271:
@AaronDonahue Sure thing! I had an issue with an expired yubikey certificate and haven't been able to access the SNL network for ~1.5 weeks (where my dev machine and all of my work lives). However, I've been promised the new one arrives tomorrow, so I'll get right on this!

components/eamxx/src/share/field/field_impl.hpp
```diff
@@ -143,66 +142,80 @@ auto Field::get_view () const
...
+                   "requested value type.\n");
 
   // Check if this field is a subview of another field
   const auto parent = m_header->get_parent().lock();
-  if (parent!=nullptr) {
+  if (parent != nullptr) {
     // Parent field has correct layout to reinterpret the view into N+1-dim view
-    // So create the parent field on the fly, use it to get the N+1-dim view, then subview it.
-    // NOTE: we can set protected members, since f is the same type of this class.
+    // So create the parent field on the fly, use it to get the N+1-dim view,
```
bartgol, line 175: This comment should be updated: the parent view may not be N+1 dim, if we are taking multiple slices, but it will have rank N as well (just potentially a bigger extent).

mjs271: Thanks for the catch!

components/eamxx/src/share/field/field.cpp
```diff
@@ -115,6 +115,46 @@ subfield (const int idim, const int index, const bool dynamic) const {
   return subfield(m_header->get_identifier().name(),idim,index,dynamic);
 }
 
+// slice at index idim, entries \in [index_beg, index_end]
```
bartgol, line 118: I would make end be the index _past_ the last slice. This would reflect what happens with the `end()` iterator from std containers and algorithms. Many devs would see "end" and immediately think that the required slices are `[beg,end)`. Also, it is more common to know the 1st index and the count, so that one would like to do `subfield(..., first, first+N)`, rather than add `-1` in the formula.

components/eamxx/src/share/field/field.cpp
```diff
@@ -127,11 +167,31 @@ get_component (const int i, const bool dynamic) {
   EKAT_REQUIRE_MSG (i>=0 && i<layout.dim(idim),
       "Error! Component index out of bounds [0," + std::to_string(layout.dim(idim)) + ").\n");
 
-  // Add _$idim to the field name, to avoid issues if the subfield is stored
+  // Add _$i to the field name, to avoid issues if the subfield is stored
   // in some structure that requires unique names (e.g., a remapper)
   return subfield (fname + "_" + std::to_string(i),idim,i,dynamic);
 }
 
+Field Field::get_components(const int i1, const int i2) {
```
bartgol, line 175: I would change the input vars names, since i1/i2 do not really explain what happens. One may misunderstand them for "oh, i only get 2 slices, i1 and i2".  I would stick with beg/end, which are more self-documenting.

components/eamxx/src/share/field/field.hpp
```diff
@@ -241,11 +241,20 @@ class Field {
...
                   const int index, const bool dynamic = false) const;
   Field subfield (const int idim, const int k, const bool dynamic = false) const;
-
+  // extracts a subfield composed of multiple slices in a continuous range of indices
+  // e.g., (in matlab syntax) subf = f.subfield(:, 1:3, :)
+  // but NOT subf = f.subfield(:, [1, 3, 4], :)
+  Field subfield (const std::string& sf_name, const ekat::units::Units& sf_units,
+                  const int idim, const int index_beg, const int index_end) const;
+  Field subfield (const std::string& sf_name, const int idim,
+                  const int index_beg, const int index_end) const;
```
bartgol, line 250: Can we clean up the signature of the subfield methods? We use sometimes `k` and sometime `index`. Let's use always the same. I don't care which one, but it makes reading the code faster.

components/eamxx/src/share/field/field.hpp
```diff
@@ -241,11 +241,20 @@ class Field {
...
+  Field subfield (const int idim, const int k_beg, const int k_end) const;
   // If this field is a vector field, get a subfield for the ith component.
   // If dynamic = true, it is possible to "reset" the component index at runtime.
   // Note: throws if this is not a vector field.
   Field get_component (const int i, const bool dynamic = false);
+  // version for slicing across multiple, contiguous indices
+  Field get_components (const int i1, const int i2);
```
bartgol, line 257: Here too

components/eamxx/src/share/field/field_alloc_prop.cpp
```diff
@@ -77,6 +77,53 @@ subview (const int idim, const int k, const bool dynamic) const {
...
+  props.m_committed = true;
+  props.m_scalar_type_size = m_scalar_type_size;
+  props.m_layout = m_layout.clone();
+  props.m_layout.reset_dim(idim, k_end - k_beg);
+
+  // Output is contiguous if either
+  //  - this->m_contiguous=true AND idim==0
+  //  - m_layout.dim(i)==1 for all i<idim
+  auto is_one = [](int i) { return i == 1; };
+  // we do not currently have the capability for contiguous multi-slice views
```
bartgol, line 105: Then we could prune the is_one lambda and the comment above.

components/eamxx/src/share/field/field_alloc_prop.cpp
```diff
@@ -77,6 +77,53 @@ subview (const int idim, const int k, const bool dynamic) const {
...
+  props.m_contiguous = false;
+
+  props.m_subview_info =
+      SubviewInfo(idim, k_beg, k_end, m_layout.dim(idim));
+
+  // Figure out strides/packs
+  if (idim == m_layout.rank()) {
+    // We're slicing the possibly padded dim, so everything else is as in the
+    // layout, and there is no packing
```
bartgol, line 114: With multislice, we can still have packing. We should maintain the same pack size as the parent view.

components/eamxx/src/share/field/field_alloc_prop.cpp
```diff
@@ -77,6 +77,53 @@ subview (const int idim, const int k, const bool dynamic) const {
...
+    props.m_last_extent = m_layout.dim(idim);
```
bartgol, line 115: This doesn't seem right. We should have m_last_extent = k_end-k_beg, no?

components/eamxx/src/share/field/field_alloc_prop.cpp
bartgol: Actually, packing becomes problematic once we slice, because of memory alignment. So yeah, let's keep pack_size_max=1 for safety.

components/eamxx/src/share/field/field_utils_impl.hpp
```diff
@@ -29,6 +29,8 @@ bool views_are_equal(const Field& f1, const Field& f2, const ekat::Comm* comm)
   f2.sync_to_host();
 
   // Reshape based on the rank, then loop over all entries.
+  // NOTE: because views_are_equal() is only used for testing, we generalize by
+  // always calling get_strided_view(), even if it could be contiguous
```
bartgol, line 33: I would add another line:
```
  // This allows to call this function on any type of field, including multi-slice subfiedls
```
Otherwise the user may wonder why we cannot use `get_view`...

components/eamxx/src/share/field/field_utils_impl.hpp
```diff
@@ -740,6 +742,9 @@ void print_field_hyperslab (const Field& f,
     const auto dims_left = get_dims_left(orig_layout);
     auto dims_str = get_dims_str(orig_layout);
 
+    // NOTE: because print_field_hyperslab() is only used for testing, we
+    // generalize by always calling get_strided_view(), even if it could be
+    // contiguous
```
bartgol, line 747: Same here.

components/eamxx/src/share/property_checks/field_nan_check.cpp
```diff
@@ -47,7 +47,7 @@ PropertyCheck::ResultAndMsg FieldNaNCheck::check_impl() const {
   switch (layout.rank()) {
```
bartgol, line 47: Can you add a comment here too, saying that "we can't be sure that the field has contiguous allocation, so we use get_strided_view"?

components/eamxx/src/share/field/field.cpp
mjs271: Good call! My brain always defaults to math notation, but not the best choice here...

components/eamxx/src/share/field/field_alloc_prop.cpp
mjs271: thanks for the catch--revealed I needed `if (idim == (m_layout.rank() - 1))`, and that's why a test didn't catch this earlier 😬

tcclevenger (commented):
Are AT fails expected?

bartgol:
I think the PR might have been tested before we blessed baselines.

components/eamxx/src/share/field/field_alloc_prop.cpp
```diff
@@ -77,6 +77,50 @@ subview (const int idim, const int k, const bool dynamic) const {
...
+      SubviewInfo(idim, index_beg, index_end, m_layout.dim(idim));
+
+  // Figure out strides/packs
+  if (idim == (m_layout.rank() - 1)) {
+    // We're slicing the possibly padded dim, so everything else is as in the
+    // layout, and there is no packing
+    // props.m_last_extent = m_layout.dim(idim);
+    props.m_last_extent = index_end - index_beg;
+    props.m_pack_size_max = 1;
+    props.m_alloc_size = m_alloc_size / m_last_extent;
```
bartgol, line 114: I don't think m_alloc_size is used in anything other than for allocating the view, so here is probably fine. However, it is odd to see this value. I would have expected `props.m_alloc_size = (m_alloc_size / m_last_extent) * props.m_last_extent`.

That said, I think we should change this to be preserved upon subview. So can you change this to be `props.m_alloc_size = m_alloc_size`? And since it should be the same in both cases of this if, you can put it outside the if statement.

components/eamxx/src/share/field/field_alloc_prop.cpp
```diff
@@ -77,6 +77,50 @@ subview (const int idim, const int k, const bool dynamic) const {
...
+  } else {
+    // We are keeping the last dim, so same last extent and max pack size
+    props.m_last_extent = m_last_extent;
+    props.m_pack_size_max = m_pack_size_max;
+    props.m_alloc_size = m_alloc_size / m_layout.dim(idim);
```
bartgol, line 119: This is the other case. When you do the change above, don't forget to prune this.

bartgol (changes requested):
Looks good. I have one last request, to avoid confusion.

jgfouca (approved):

bartgol (approved):

bartgol:
NL fails are due to the upstream merge (we have to bless). Merging.

PR #2786, opened 2024-04-05
add new default production spa file

mahf708:
This is based on a v3 LR (ne30pg2) experiment, with everything corrected in the recently identified bug. Some experiments were done. Documentation to follow.

[Internal documentation of creating the new file](https://acme-climate.atlassian.net/wiki/spaces/EAMXX/pages/4239851548/spa+v3.LR.F2010.2011-2025.c+20240405).

[Non-B4B]

### DIFF
--- a/components/eamxx/cime_config/namelist_defaults_scream.xml
+++ b/components/eamxx/cime_config/namelist_defaults_scream.xml
@@ -312,10 +312,10 @@ be lost if SCREAM_HACK_XML is not enabled.
       <spa_remap_file COMPSET=".*DP-EAMxx"/>
 
       <spa_data_file type="file" doc="File containing aerosol data. Must be on same grid as the atm, or a coarser one"/>
-      <spa_data_file hgrid="ne.*np4.pg2">${DIN_LOC_ROOT}/atm/scream/init/spa_from_eamv3.ne30pg2.F2010.2011-2025.c_20240405.nc</spa_data_file>
       <spa_data_file hgrid="ne.*np4">${DIN_LOC_ROOT}/atm/scream/init/spa_from_eamv3.ne30np4.F2010.2011-2025.c_20240405.nc</spa_data_file>
-      <spa_data_file hgrid="ne4np4.pg2">${DIN_LOC_ROOT}/atm/scream/init/spa_from_eamv3.ne4pg2.F2010.2011-2025.c_20240405.nc</spa_data_file>
+      <spa_data_file hgrid="ne.*np4.pg2">${DIN_LOC_ROOT}/atm/scream/init/spa_from_eamv3.ne30pg2.F2010.2011-2025.c_20240405.nc</spa_data_file>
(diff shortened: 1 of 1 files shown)


AaronDonahue:
@hassanbeydoun and/or @brhillman can you confirm this is the correct SPA data file?  I'm inclined to believe it is, but since the production runs will be impacted I want to check before merging to master.

bartgol (dismissed):

components/eamxx/cime_config/namelist_defaults_scream.xml
```diff
@@ -296,7 +296,7 @@ be lost if SCREAM_HACK_XML is not enabled.
 
       <spa_data_file type="file" doc="File containing aerosol data. Must be on same grid as the atm, or a coarser one"/>
       <spa_data_file hgrid="ne.*np4">${DIN_LOC_ROOT}/atm/scream/init/spa_file_unified_and_complete_ne30_20220428.nc</spa_data_file>
-      <spa_data_file hgrid="ne.*np4.pg2">${DIN_LOC_ROOT}/atm/scream/init/spa_file_unified_and_complete_ne30pg2_20240111.nc</spa_data_file>
+      <spa_data_file hgrid="ne.*np4.pg2">${DIN_LOC_ROOT}/atm/scream/init/spa_v3.LR.F2010.2011-2025.c_20240405.nc</spa_data_file>
```
bartgol, line 299: Should we maybe add the grid name to the file? It may help to recognize immediately what the resolution is, without having to open the file...

AaronDonahue:
I'm retesting just because the last test was nearly a week ago.  Once we see the expected DIFFs can we say this is ready to merge?  @mahf708?

components/eamxx/cime_config/namelist_defaults_scream.xml
bartgol: We can't remap ne30->ne4 _online_. Unfortunately, the online remappers hard coded the assumption that the fine grid is the model grid (so you can coarsen output, or read in coarse inputs). It is of course possible to extend them, but when I tried I quickly realized it wasn't going to be super quick, so I gave up (since we didn't really need it).

For this case (ne4 input) I would simply remap offline, and upload the new ne4 spa data on the server.

bartgol: I'm always happy if someone else does the work... ;-)

bartgol:
> @bartgol @AaronDonahue: should be ready. Tests should fail now. I dropped support for np4, which means likely more tests will fail. We should move our testing to be strictly pg2 (like we did in upstream e3sm).

Dropping support for np4, albeit reasonable, is a big step. We should either keep np4 and integrate this PR, or drop np4 in a separate PR first.

@PeterCaldwell @mt5555  what are your thoughts about supporting only pg2 in EAMxx?

bartgol (dismissed):

mt5555:
for the record: i'm against dropping support for np4.  And I think the issue above has been resolved (so we can keep np4).

bartgol:
@mahf708 It appears this PR got a bit stale. When you get back, ping me if you still think we should integrate it, otherwise just go ahead and close it.

PR #2787, opened 2024-04-05
calculate aodvis as a model-level diag

mahf708:
Adds ability to calculate aodvis as part of diagnostics

### DIFF
--- a/components/eamxx/src/diagnostics/aodvis.cpp
+++ b/components/eamxx/src/diagnostics/aodvis.cpp
@@ -31,7 +31,6 @@ void AODVis::set_grids(
   add_field<Required>("aero_tau_sw", scalar3d_swband_layout, nondim, grid_name);
 
   // Construct and allocate the aodvis field
-  // We are going to assume we have nondim units here for ease
   FieldIdentifier fid("AerosolOpticalDepth550nm", scalar1d_layout, nondim,
                       grid_name);
   m_diagnostic_output = Field(fid);
@@ -45,7 +44,7 @@ void AODVis::compute_diagnostic_impl() {
 
   const auto aod     = m_diagnostic_output.get_view<Real *>();
   const auto tau_vis = get_field_in("aero_tau_sw")
-                           .subfield(1, n_vis_bnd)
+                           .subfield(1, m_vis_bnd)
                            .get_view<const Real **>();
 
   const auto num_levs = m_nlevs;
--- a/components/eamxx/src/diagnostics/aodvis.hpp
+++ b/components/eamxx/src/diagnostics/aodvis.hpp
@@ -32,7 +32,7 @@ class AODVis : public AtmosphereDiagnostic {
   int m_nlevs;
 
   int m_swbands = eamxx_swbands();
-  int n_vis_bnd = eamxx_vis_swband_idx();
+  int m_vis_bnd = eamxx_vis_swband_idx();
 };
 
 }  // namespace scream


AaronDonahue (commented):
Based on the structure alone I would approve, however, I can't speak to the actual calculation of AODvis in the diagnostic.  I'll let others verify that the correct band is used and that the equation is correct.

In the meantime, I do have one question.  Are there specific AODvis properties that we could test for?  One that comes to mind is if `tau` is uniformly zero we would expect AODvis to be zero too right?  Could we add to the test this property test, and any others that come to mind?

AaronDonahue:
>@AaronDonahue what do you think about the name here? I don't think aodvis (what I have now) aligns with other SCREAM variables, so maybe I should change it... AODvis? AerosolOpticalDepthVis?

Let's chat about this tomorrow and finalize the name.  My vote is `AerosolOpticalDepthVis`, but I'm not a user of the diag so maybe others have better feedback.

susburrows:
On the dev call today, we discussed the idea of adding the wavelength [band] to the variable name to make information more visible to users, e.g, Aerosol_Optical_Depth_533nm

susburrows:
The CMIP6 variable name is od550aer https://clipc-services.ceda.ac.uk/dreq/u/8f2fb9e812c26ee6cb8d9673e09d2644.html

components/eamxx/src/diagnostics/aodvis.cpp
```diff
@@ -0,0 +1,68 @@
...
+void AODVis::compute_diagnostic_impl() {
+  using KT  = KokkosTypes<DefaultDevice>;
+  using MT  = typename KT::MemberType;
+  using ESU = ekat::ExeSpaceUtils<typename KT::ExeSpace>;
+
+  const auto aod = m_diagnostic_output.get_view<Real *>();
+  // TODO: don't hardcode swbnd 10
+  // Get slice of tau at swbnd 10
+  const auto tau_vis =
+      get_field_in("aero_tau_sw").subfield(1, 10).get_view<Real **>();
```
bartgol, line 52: Minor: you could use `const Real`, for safety and self-documentation.

components/eamxx/src/diagnostics/aodvis.cpp
```diff
@@ -0,0 +1,68 @@
...
+  const auto policy   = ESU::get_default_team_policy(m_ncols, m_nlevs);
+  Kokkos::parallel_for(
+      "Compute " + name(), policy, KOKKOS_LAMBDA(const MT &team) {
+        const int icol = team.league_rank();
+        auto tau_icol  = ekat::subview(tau_vis, icol);
+        Kokkos::parallel_reduce(
+            Kokkos::TeamVectorRange(team, num_levs),
+            [&](const int &ilev, Real &lsum) { lsum += tau_icol(ilev); },
+            aod(icol));
+        team.team_barrier();
```
bartgol, line 64: This barrier is unnecessary.

components/eamxx/src/diagnostics/tests/aodvis_test.cpp
```diff
@@ -0,0 +1,110 @@
...
+
+    for(int icol = 0; icol < grid->get_num_local_dofs(); ++icol) {
+      auto aod_temp = 0.0;
+      for(int ilev = 0; ilev < nlevs; ++ilev) {
+        // TODO: Don't hardcode the 10!
+        aod_temp += tau_h(icol, 10, ilev);
+        aod_t(icol) += tau_h(icol, 10, ilev);
+      }
+      // Allow for sp tolerance?
+      REQUIRE(std::abs(aod_temp - aod_h(icol)) < 1e-6);
```
bartgol, line 103: You allow a tolerance here, but not in the check below. Imho, one check is enough. You choose which one you want to keep.

components/eamxx/src/diagnostics/aodvis.cpp
```diff
@@ -0,0 +1,68 @@
...
+  const auto tau_vis =
+      get_field_in("aero_tau_sw").subfield(1, 10).get_view<Real **>();
+
+  const auto num_levs = m_nlevs;
+  const auto policy   = ESU::get_default_team_policy(m_ncols, m_nlevs);
+  Kokkos::parallel_for(
+      "Compute " + name(), policy, KOKKOS_LAMBDA(const MT &team) {
+        const int icol = team.league_rank();
+        auto tau_icol  = ekat::subview(tau_vis, icol);
+        Kokkos::parallel_reduce(
```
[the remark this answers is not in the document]
bartgol: I think on host we're ok, since the diag does not use packs, so no issue there. I don't recall if the ekat wrapper was also important for GPU...

components/eamxx/src/diagnostics/tests/aodvis_test.cpp
bartgol: Uhm, no. `views_are_equal` does exactly what you'd expect: it checks that all entries are the same. There's no tolerance in the test.

bartgol: Ah wait. `views_are_equal` does a deep copy to host first. Since you never modified the device view, you prob get the host view full of 0's. You should add `aod_tf.sync_to_dev()` before calling `views_are_equal`.

components/eamxx/src/diagnostics/aodvis.cpp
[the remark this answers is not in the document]
tcclevenger: That is correct (if we need BFB). You could use 
```
aod(icol) = ekat::ExecSpaceUtils::view_reduction(team, 0, num_levs, tau_icol).
```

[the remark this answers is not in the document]
tcclevenger: I can't remember either about GPU, but I'm guessing yes it was important. `Kokkos::parallel_reduce()` shouldn't be able to guarantee summation order, right?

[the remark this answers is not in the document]
tcclevenger: Maybe I am misunderstanding..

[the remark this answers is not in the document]
bartgol: That's right. We don't have C-F bfbf needs, but the concern here is with two consecutive runs. I don't think `Kokkos::parallel_reduce` is BFB within two consecutive runs...

@mahf708 I think Conrad's suggestion is the cleanest thing.

components/eamxx/src/diagnostics/tests/aodvis_test.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+    for(int icol = 0; icol < grid->get_num_local_dofs(); ++icol) {
+      for(int ilev = 0; ilev < nlevs; ++ilev) {
+        // TODO: Don't hardcode the 10!
+        aod_t(icol) += tau_h(icol, 10, ilev);
+      }
+    }
+    aod_hf.sync_to_dev();
+    aod_tf.sync_to_dev();
+    REQUIRE(views_are_equal(aod_hf, aod_tf));
```
tcclevenger, line 102: The fail is here in release mode. I think it's because in release, EKAT_DEFAULT_BFB=False and so `view_reduction()` doesn't use the serialized version. Let me take a look and see if we have any workarounds other places in the code. We could always test views to a tolerance in release.

tcclevenger, line 102: The only places we use `view_reduction()` all tests are guarded by `SCREAM_BFB_TESTING`, so there isn't this issue in release mode. A possible workaround is the following, which would mean we aren't testing the diagnostic in release, which I'd be ok with. Maybe @bartgol has other thoughts?
```suggestion
    // Workaround for non-bfb behavior of view_reduction() in release builds
    if (SCREAM_BFB_TESTING) { 
      REQUIRE(views_are_equal(aod_hf, aod_tf));
    }
```

components/eamxx/src/diagnostics/aodvis.cpp
```diff
@@ -0,0 +1,64 @@
...
+  m_diagnostic_output.allocate_view();
+}
+
+void AODVis::compute_diagnostic_impl() {
+  using KT  = KokkosTypes<DefaultDevice>;
+  using MT  = typename KT::MemberType;
+  using ESU = ekat::ExeSpaceUtils<typename KT::ExeSpace>;
+
+  const auto aod = m_diagnostic_output.get_view<Real *>();
+  // TODO: don't hardcode swbnd 10
```
bartgol, line 49: Have you thought about exposing the band number as part of the diag API? In practice, it would be the same, but it may be more self-documenting, and less hidden in the code... E.g., you could add this to the class header:
```cpp
  static constexpr int vis_band_idx = 10;
```
and then use it here:
```c++
  const auto tau_vis = ... subfield(1,vis_band_idx) ...
  ```

**Review by bartgol — DISMISSED** [review]

**Comment by bartgol** [comment]
@mahf708 I removed AUTOMERGE so you get a chance to read my last comment before the AT takes over. Feel free to disregard it, and merge manually (or re-add the automerge label).

File: components/eamxx/src/diagnostics/aodvis.cpp (hunk 6 above)
Comment (bartgol) [reply components/eamxx/src/diagnostics/aodvis.cpp:49]: Then it's wise to put it in a header that all three can access. My preference is for `scream_universal_constants.hpp`, even though the file name suggests it should contain model-indep stuff (like "speed of light" kind of stuff), rather than the index of the speed of light.

Another alternative is to use a function, maybe to be put in `src/util/scream_common_physics_functions.hpp`:
```cpp
constexpr int eamxx_vis_band_idx () {
  // Some comment explaining what this is, and that the idx itself
  // may vary if we increase the number/position of bands
  return 10;
}
```

bartgol (approved):

components/eamxx/src/diagnostics/aodvis.hpp
```diff
@@ -0,0 +1,40 @@
...
+ protected:
+#ifdef KOKKOS_ENABLE_CUDA
+ public:
+#endif
+  void compute_diagnostic_impl();
+
+  int m_ncols;
+  int m_nlevs;
+
+  int m_swbands = eamxx_swbands();
```
bartgol, line 34: TODO: do the same thing in mam4xx and rrtmgp interfaces.

PR #2792, opened 2024-04-19
docs: write top-level and tech doc overview; add .bib file

PeterCaldwell:
First attempt at writing overview text for the whole EAMxx component and for the tech doc. Required adding EAMxx's .bib file, which so far includes citations in alphabetical order. We'll see how long that lasts!


PeterCaldwell:
I think my first docs PR is ready for review. Do we have a protocol for these kinds of things?

components/eamxx/docs/index.md
```diff
@@ -1,10 +1,11 @@
-# The C++ E3SM Atmosphere Model (EAMxx)
+# The E3SM Atmosphere Model in C++ (EAMxx)
```
whannah1, line 1: How about "the E3SM Atmosphere Model ++ (EAMxx)" so it sounds like an incremental improvement on EAM rather than just a rewrite?

components/eamxx/docs/index.md
```diff
@@ -1,10 +1,11 @@
...
 
-Some nice introductory text goes here! Maybe some figures, too. Who knows?!
+EAMxx is an entirely new atmosphere model for E3SM. It is written in C++ using the Kokkos performance portability library to enable it to run efficiently on CPUs, GPUs, and (hopefully) whatever comes next. Currently only the km-scale explicit-convection version called SCREAM (the Simple Cloud-Resolving E3SM Atmosphere Model) is available, but a low-resolution version is in the works.
```
whannah1, line 3: Phrases like "entirely new" seem destined to go out of date by not being updated when we start to feel like the model is "not new" anymore, so I would use a tone more apt for documentation rather than an advertisement.

components/eamxx/docs/refs.bib
```diff
@@ -0,0 +1,125 @@
...
+pages = {e2020MS002419},
+keywords = {computational efficiency, E3SM, effective resolution, grid remap methods},
+doi = {https://doi.org/10.1029/2020MS002419},
+url = {https://agupubs.onlinelibrary.wiley.com/doi/abs/10.1029/2020MS002419},
+eprint = {https://agupubs.onlinelibrary.wiley.com/doi/pdf/10.1029/2020MS002419},
+note = {e2020MS002419 2020MS002419},
+year = {2021}
+}
+
+@article {Morrison_Milbrandt15,
```
whannah1, line 72: Do we care about consistent formatting here?
Or are we ok with it turning into a "dumping ground" of bibtex info?
(personally I'm ok relaxing my standards here)

components/eamxx/docs/technical/index.md
```diff
@@ -1,3 +1,23 @@
-# SCREAM Technical Guide
+# EAMxx Technical Guide
 
-SCREAM contributors and maintainers will add detailed technical information about SCREAM here.
+The goal of this document is to describe the specific equations, parameterizations, and numerical methods used in the current version of EAMxx. Because our master-branch implementation changes every time we make a new commit, this documentation will also evolve continuously. As such, documentation for master should always be considered to be preliminary and under construction. If you want trustworthy documentation, pull it from an official model release.
```
whannah1, line 3: seems like "specific equations, parameterizations, and numerical methods" should be put into a list.

Also, maybe "relevant equations" is more accurate to what we hope to put here? I'm imaging there might be a page that uses a simple derivation to explain a fundamental concept of a particular scheme?

On that note - I'm curious how well we can write equations in markdown... how amazing would it be to write a block of latex!?

components/eamxx/docs/technical/index.md
```diff
@@ -1,3 +1,23 @@
...
+5. the effects of aerosol are prescribed via the Simple Prescribed Aerosol (SPA) scheme, which is very similar to MACv2-SP [@Stevens_et17]
+5. the **P3 microphysics** scheme from [@Morrison_Milbrandt15] modified as described by [@Caldwell_et21] to assume instantaneous liquid saturation adjustment for consistency with SHOC
+6. **RTE/RRTMGP radiation** from [@Pincus_et19] rewritten in C++ for consistency and performance
+
+By default processes are called in this order, but which processes to include and in what order is modifiable at run time. These parameterizations are described in more detail in Caldwell et al 2021. The CFMIP Observation Simulator Package (COSP) is also integrated into EAMxx, but currently only the ISCCP output is enabled. As in EAM, dynamics operates on a spectral element grid and all other processes use a finite-volume grid that divides each spectral element into 4 quadrilaterals. This physics grid is described in [@Hannah_et21].
```
whannah1, line 23: - "By default processes are called in this order [following the coupling with surface component models]"
- Need bibtex citation/link for Caldwell et al 2021?
- should COSP be added to the list above rather than touched on here?

components/eamxx/docs/technical/index.md
```diff
@@ -1,3 +1,23 @@
...
+2. the **Simple Higher-Order Closure (SHOC)** parameterization from [@Bogenschutz_Krueger13], which handles turbulent diffusion, condensation/evaporation, and liquid cloud fraction
+3. **turbulent mountain stress** is crudely parameterized following [@Fiedler_Panofsky72] to reduce excessive winds around topography
+4. an **all-or-nothing ice cloud fraction** parameterization that sets ice cloud fraction to 100% whenever cloud ice mass q<sub>i</sub> is less than a user-specified threshold set by default to 1e-5 kg/kg. This scheme also sets the total cloud fraction (used by microphysics) to the maximum of the liquid and ice cloud fraction.
+5. the effects of aerosol are prescribed via the Simple Prescribed Aerosol (SPA) scheme, which is very similar to MACv2-SP [@Stevens_et17]
+5. the **P3 microphysics** scheme from [@Morrison_Milbrandt15] modified as described by [@Caldwell_et21] to assume instantaneous liquid saturation adjustment for consistency with SHOC
+6. **RTE/RRTMGP radiation** from [@Pincus_et19] rewritten in C++ for consistency and performance
```
whannah1, line 21: Do you plan to have links here that point to pages on each?
If so, I would opt for a less-verbose list - or at least a single word at the front that can contain the link, like:
- [P3 Microphysics](link here) -  from [@Morrison_Milbrandt15] modified as described by [@Caldwell_et21]

components/eamxx/docs/index.md
```diff
@@ -1,10 +1,11 @@
...
-  its standalone configuration and within E3SM.
-* The [Developer Guide](developer/index.md) contains all the information needed
-  to contribute to the development of EAMxx.
-* The [Technical Guide](technical/index.md) contains all the technical
-  information about EAMxx.
+Like the documentation for other component models, EAMxx documentation is divided into:
+
+* The [User Guide](user/index.md) which explains how to run EAMxx and describes all the compile-time and run-time options available for modifying a simulation
+* The [Developer Guide](developer/index.md) which contains all the information needed to contribute to the development of EAMxx
+* The [Technical Guide](technical/index.md) which describes the equations and numerical methods used by each of the specific process implementations and diagnostics available in the current version of EAMxx
```
whannah1, line 9: I think these could be less verbose:

* The [User Guide](user/index.md) - how to run EAMxx and options for modifying a simulation
* The [Developer Guide](developer/index.md) -  information needed to contribute to EAMxx development 
* The [Technical Guide](technical/index.md) - equations and numerical methods used in EAMxx

Also, should the technical guide really contain info about diagnostics? Seems like we should have a separate area for that. For the higher level E3SM docs we've been talking about dividing the guides into pre and post processing guides. EAMxx has special considerations for each - so maybe we will eventually want a set of notes here? idk

Or am I totally misunderstanding what "diagnostics" means and it's all about run-time diagnostics for debugging or online checks for numerical instability?

components/eamxx/docs/index.md
PeterCaldwell: Yeah, that's a good point, but it's official name is "EAM in C++". Hindsight 20/20 but I think it is too late for yet another rebranding.

PeterCaldwell: Yeah, that's a good point, but it's official name is "EAM in C++". Hindsight 20/20 but I think it is too late for yet another rebranding.

components/eamxx/docs/index.md
PeterCaldwell: I meant "entirely new" in the sense that it doesn't share any parameterizations with EAMf90... though I suspect that will change with some of the low-res parameterizations. How about calling it a "clean-slate atmosphere model"?

components/eamxx/docs/index.md
PeterCaldwell: Ok, I took you "less verbose" suggestions.

I think you are misunderstanding what diagnostics means. EAMxx has a "diagnostics" class that consists of variables you can ask to be written out which aren't calculated anywhere in the actual processes, but instead are assembled from other available field-managed variables. This is as opposed to diagnostics one would compute or look at by postprocessing data. Maybe we could come up with a better name for the diagnostics class?

components/eamxx/docs/refs.bib
PeterCaldwell: What consistency rules would you suggest? I think it doesn't really matter how you format bibtex entries because latex rules will make everything look the same once compiled. Or do you mean conventions for article labels?

components/eamxx/docs/technical/index.md
PeterCaldwell: regarding putting stuff in a list versus inline - I think this is a matter of preference and I'm not going to change it. 

regarding "relevant equations" - I wanted to emphasize that the tech doc describes the particular version of the code associated with the current git hash rather than talking about generic best practices for writing a process (which should be in the dev guide). So "specific" here refers to equations, params, and methods rather than just equations. Thus I don't want to change it. 

regarding latex equations - we can! See https://github.com/E3SM-Project/E3SM/blob/master/components/elm/docs/tech-guide/top_solar_parameterization.md as an example.

components/eamxx/docs/technical/index.md
whannah1: ok, sound good.

components/eamxx/docs/refs.bib
whannah1: I was referring to all of that. The indentation is what made me think of it, but then I got to thinking whether should have a note about the format and provide a loose "template".

But I also think that's kinda dumb because it's likely people won't conform, so let's forget that I suggested this.

components/eamxx/docs/index.md
whannah1: ok, so we just want to convey that it's not an iterative evolution of EAM.

Here's an idea for a rewrite:

```
EAMxx is one of the available atmospheric models for E3SM that is being built from the ground up using C++ in order to embrace modern software practices and allow "performance portability" across various super computers using the Kokkos library for optimal efficiency on CPUs, GPUs, and whatever comes next.
```

components/eamxx/docs/index.md
whannah1: 👍

bartgol:
We should look into mkdocs extensions like [this](https://github.com/shyamd/mkdocs-bibtex), to make usage of the bib file easier...

bartgol:
> This makes me think. We should ideally revamp how we are compiling our docs. Why don't we just compile everything like upstream does? (Having two repos is kinda difficult 💀)

Well, I don't know. On one hand, it would be way easier to have a single job that builds the whole thing. OTOH, while we're still on a fork, users may be more interested in finding out about EAMxx alone, so a project-specific docs makes sense.

Besides, using the e3sm job would make us overwrite upstream docs, which is probably not  right (unless we just did an upstream merge, we may be missing some docs in our fork).

components/eamxx/docs/index.md
PeterCaldwell: Ok, new version (uploaded soon) will be:
>EAMxx was built from the ground up using C++ in order to embrace modern software practices and to allow "performance portability" across various supercomputers. The latter goal is achieved by using the Kokkos library. EAMxx is a "clean-start" model with almost no similarity to the E3SM atmosphere model used in versions 1-3.  Currently only the km-scale explicit-convection version called SCREAM (the Simple Cloud-Resolving E3SM Atmosphere Model) is available, but a low-resolution version is in the works.

components/eamxx/docs/index.md
PeterCaldwell: Since I took your suggestions for shorter bullets, diagnostics aren't mentioned here. In the actual tech doc where we describe them... we should be more clear. I guess diagnostic outputs is ok but I suspect people will still be confused. That's a then problem.

components/eamxx/docs/technical/index.md
PeterCaldwell: Ok, updated version should fix all of this. Good catches!

components/eamxx/docs/technical/index.md
```diff
@@ -1,3 +1,24 @@
...
+
+Currently, EAMxx is only configured for km-scale convection-permitting runs. In order to provide scientifically-credible simulations at lower resolution, parameterizations for the following processes would be needed:
+
+1. deep convection
+2. gravity-wave drag
+3. energy fixer
+
+The only configuration of EAMxx that is currently implemented is the convection-permitting version, commonly known as the Simple Cloud-Resolving E3SM Atmosphere Model (SCREAM). Processes in EAMxx-SCREAM are:
+
+1. a non-hydrostatic version of the **spectral-element dynamical core** used by other E3SM Atmosphere Model versions [@Taylor_et20] with semi-Lagrangian tracer advection as described by [@Bradley_et22]
```
crterai, line 15: Here and other places: 
For citations, we still need to write out the citation. The references in brackets only give the numbers. 
For example, in the [ELM TOP description](https://github.com/E3SM-Project/E3SM/blob/master/components/elm/docs/tech-guide/top_solar_parameterization.md), the citation is written out as: 
```
  Fig 1.: A schematic diagram of the five components of incoming solar radiation over mountains adapted from Lee et al. (2011)[@lee2011parameterization]. 
```

crterai (changes requested):
Apart from the citation, it looks good to me.

crterai (approved):
Looks great!

PR #2793, opened 2024-04-22
mam4xx - updating the mam4xx and haero submodules.

odiazib:
Updating the mam4xx and haero submodules to includes a fix for compute-sanitizer fail in EAMxx.

### DIFF
--- a/components/eamxx/src/physics/mam/mam_coupling.hpp
+++ b/components/eamxx/src/physics/mam/mam_coupling.hpp
@@ -76,7 +76,7 @@ constexpr int num_aero_tracers() {
 
 // Given a MAM aerosol mode index, returns a string denoting the symbolic
 // name of the mode.
-KOKKOS_INLINE_FUNCTION
+inline
 const char* aero_mode_name(const int mode) {
   static const char *mode_names[num_aero_modes()] = {
     "1",
@@ -89,7 +89,7 @@ const char* aero_mode_name(const int mode) {
 
 // Given a MAM aerosol-related gas ID, returns a string denoting the symbolic
 // name of the gas species.
-KOKKOS_INLINE_FUNCTION
+inline
 const char* gas_species_name(const int gas_id) {
   static const char *species_names[num_aero_gases()] = {
     "O3",
@@ -110,14 +110,14 @@ constexpr int max_field_name_len() {
   return 128;
 }
 
-KOKKOS_INLINE_FUNCTION
+inline
 size_t gpu_strlen(const char* s) {
   size_t l = 0;
   while (s[l]) ++l;
   return l;
 }
 
-KOKKOS_INLINE_FUNCTION
+inline
 void concat_2_strings(const char *s1, const char *s2, char *concatted) {
   size_t len1 = gpu_strlen(s1);
   for (size_t i = 0; i < len1; ++i)
@@ -128,7 +128,7 @@ void concat_2_strings(const char *s1, const char *s2, char *concatted) {
   concatted[len1+len2] = 0;
 }
 
-KOKKOS_INLINE_FUNCTION
+inline
 void concat_3_strings(const char *s1, const char *s2, const char *s3, char *concatted) {
   size_t len1 = gpu_strlen(s1);
   for (size_t i = 0; i < len1; ++i)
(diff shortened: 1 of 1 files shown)


odiazib:
After upgrading the mam4xx submodule, the baselines test for the single process test of mam4xx-optics failed with significant differences between the baselines and the current PR values. I reviewed the github log in mam4xx and found that these differences were produced by [PR#305]( https://github.com/eagles-project/mam4xx/pull/305) in mam4xx. In this mam4xx PR, we changed the order of species in each mode in mam4xx to match the species order in mam4 (Fortran version). Furthermore, I reviewed the emaxx interface for mam4xx-optics, and found a mistake in the data transfer between the prognostic data structure and temporal variables in mam4xx (i.e., `state_q` and `qqwc`). The mistake (bug) was that I assumed that mam4xx and eamxx interface have same species ordering; however, we are using the species ordering from mam4 (Fortran version) in the emaxx interface. Note that after upgrading the mam4xx submodule, this mistake was resolved. Hence, I have changed the label BFB to non-BFB of PR #2793

singhbalwinder (approved):

singhbalwinder:
Tagging @susburrows so that she knows about this submodule update.

tcclevenger (dismissed):
Looks good to me.

tcclevenger:
We will have to manually merge since it's non-BFB. I can do that if you are confident the output is all expected.

components/eamxx/src/physics/mam/mam_coupling.hpp
```diff
@@ -238,7 +223,8 @@ const char* cld_aero_mmr_field_name(const int mode, const int species) {
   if (!cld_aero_mmr_names(mode, species)[0]) {
     const auto aero_id = mam4::mode_aero_species(mode, species);
     if (aero_id != mam4::AeroId::None) {
-      concat_3_strings(aero_species_name(static_cast<int>(aero_id)),
+      auto aerosol_species_name =mam4::aero_id_short_name(aero_id);
```
odiazib, line 226: @jeff-cohere , I am using `std::string `for this species name return from` mam4::aero_id_short_name`.  It should work. However, I noted you are using `char*` in `concat_3_strings`. Should we switch `std::string ` to `char* in  `mam4::aero_id_short_name`?

odiazib:
> We will have to manually merge since it's non-BFB. I can do that if you are confident the output is all expected.

Yes, the output is all expected. Let's wait for @jeff-cohere to answer one minor question and can merge this PR.

components/eamxx/src/physics/mam/mam_coupling.hpp
jeff-cohere: I think when I wrote this I was just favoring C strings over C++ strings because I wasn't sure exactly where all the code would end up being executed (i.e. on CPU or GPU). Switching to C++ strings is fine for anything that doesn't need to run on the GPU.

odiazib: This part of the code should not be invoked by GPUs. Is this correct @bartgol ?

bartgol: This routine is marked with KOKKOS_INLINE_FUNCTION, so I'm assuming you are planning to call it from device. Maybe the better question is why you need to deal with strings on device? Usually, we address different variables by an integer idx when we're in device (i.e., performance critical) code...

odiazib: We are not planning to call this routine from device. I will removed KOKKOS_INLINE_FUNCTION.

odiazib: @bartgol , what is the routine that you mentioned in your previous comment?

odiazib: It is `cld_aero_mmr_field_name`. @jeff-cohere, we are not planning to invoke this routine from the device. is this correct?

jeff-cohere: It should be possible to avoid calling this function on the GPU, unless we're using it there already and don't realize it. Maybe the easiest thing is to switch it over to using C++ strings, get rid of `KOKKOS_INLINE_FUNCTION`, and see whether things still work.

odiazib: I will remove `KOKKOS_INLINE_FUNCTION` from routines that return  a` char*`.  In  a future PR, we can switch `char*`  to C++ strings.

odiazib:
@tcclevenger @AaronDonahue. Can you merge this PR? 

I ran the automated tests (AT), and two tests are producing differences (DIFFs):
•	`mam4_optics_standalone_baseline_cmp`: This failure is expected.
•	`SMS_D_Ln5.ne4pg2_oQU480.F2010-SCREAMv1-MPASSI.mappy_gnu.scream-mam4xx-optics`: These DIFFs are also expected; however, this case is producing non-deterministic values. We are working on fixing this issue, but the fix will be part of another pull request (PR).
@jgfouca , could you update the baselines for mam4_optics_standalone_baseline_cmp after this pull request (PR) is merged.

tcclevenger:
@odiazib My only concern is that, if we merge, every mappy AT instance will fail because `SMS_D_Ln5.ne4pg2_oQU480.F2010-SCREAMv1-MPASSI.mappy_gnu.scream-mam4xx-optics` will DIFF (either we don't update baselines, or non-determinism). How far away is the non-determinism fix do you think?

odiazib:
> `SMS_D_Ln5.ne4pg2_oQU480.F2010-SCREAMv1-MPASSI.mappy_gnu.scream-mam4xx-optics`

We will need to update the baselines for SMS_D_Ln5.ne4pg2_oQU480.F2010-SCREAMv1-MPASSI.mappy_gnu.scream-mam4xx-optics.
Regarding the non-deterministic issue observed in chrysalis, we have identified that the issue occurs when this case is run in debug (D) mode with more than one thread. In contrast, the SMS_D_Ln5 configuration in mappy uses only one thread. To confirm, I have just run this case in mappy.

tcclevenger:
Oh ok, I didn't realize it was only for Chrysalis. Yeah, I'll merge.

PR #308, opened 2020-03-21
Aarondonahue/lc testing

AaronDonahue:
Update machine config files for Livermore Computing to facilitate running E3SM tests from CIME.

Also fixed an issue in the Yaml CMAKE file that broke the FSCREAM-SA tests when run in CIME.

PR #2794, opened 2024-04-23
Adds MAM4xx: Aerosol Cloud Interactions (ACI) interface

singhbalwinder:
This PR connects MAM4xx’s Aerosol Cloud Interactions (ACI) codes to EAMxx using an interface. Since the cmake option, `SCREAM_ENABLE_MAM` is always `ON`, this code is always compiled with the EAMxx codebase. ACI is added as an EAMxx process with the ability to turn it on/off using the label `mam4_aci` in the ` namelist_scream.xml` file.  The atmchange command to invoke this process in a CIME simulation is:
 
`./atmchange mac_aero_mic::atm_procs_list="tms,shoc,cldFraction,mam4_aci,p3"`

To set up a test with the above configuration, use the following commands:
```
cd cime/scripts
./create_test SMS_D_P32x1_Ln5.ne4pg2_oQU480.F2010-SCREAMv1-MPASSI.<machine>_<compiler>.scream-mam4xx-aci -t <test id> -p <project id>
```
Where,
`<machine>` is the machine name
`<compiler>` is the compiler to use on the machine
`<test id >` is a unique string test identifier
`<project id>` is the allocation project to charge

`scream-mam4xx-aci ` is the test modifier that adds `mam4_aci` to the `mac_aero_mic` process list and increase the number of tracers to 41.
 
ACI invokes the following sub-processes:
1.     Ice nucleation
2.     Droplet activation
3.     Heterogeneous freezing
 
The output from the processes above is used by the P3 microphysics scheme:
1.     Number of activated aerosol for ice nucleation[#/kg]
2.     Tendency in droplet number mixing ratio [#/kg/s]
3.     Heterogeneous freezing by immersion nucleation [cm^-3 s^-1]
4.     Heterogeneous freezing by contact nucleation [cm^-3 s^-1]
5.     Heterogeneous freezing by deposition nucleation [cm^-3 s^-1]

Other outputs:
 Cloudborne and interstitial aerosols
 
ACI also needs input from the SHOC process, so a new code has been added to get those inputs from SHOC.
 
### Testing
 We added 1 single process, 4 multi process, and 1 CIME low-resolution test in EAMxx. These are all smoke tests. MAM4xx has its validation testing suites where we validate all MAM4xx codes. Manual end-to-end validation was done for the entire ACI interface and the results were exactly the same as EAMv2 grid cells up to single precision. We don’t expect an exact match till double precision as the computing environments are different for both runs.
 
### Figures from model analysis
[figure_mam_aci_850hPa_ncdact_daily_avg.pdf](https://github.com/E3SM-Project/scream/files/15441981/figure_mam_aci_850hPa_ncdact_daily_avg.pdf)
[figure_mam_aci_850hPa_mass_SO4_daily_avg.pdf](https://github.com/E3SM-Project/scream/files/15441982/figure_mam_aci_850hPa_mass_SO4_daily_avg.pdf)
[figure_mam_aci_850hPa_mass_POM_daily_avg.pdf](https://github.com/E3SM-Project/scream/files/15441983/figure_mam_aci_850hPa_mass_POM_daily_avg.pdf)
[figure_mam_aci_850hPa_ccn_0p5_daily_avg.pdf](https://github.com/E3SM-Project/scream/files/15441984/figure_mam_aci_850hPa_ccn_0p5_daily_avg.pdf)
[figure_mam_aci_850hPa_mass_SOA_daily_avg.pdf](https://github.com/E3SM-Project/scream/files/15441985/figure_mam_aci_850hPa_mass_SOA_daily_avg.pdf)
[figure_mam_aci_850hPa_ccn_0p1_daily_avg.pdf](https://github.com/E3SM-Project/scream/files/15441986/figure_mam_aci_850hPa_ccn_0p1_daily_avg.pdf)
[figure_mam_aci_850hPa_ccn_0p2_daily_avg.pdf](https://github.com/E3SM-Project/scream/files/15441987/figure_mam_aci_850hPa_ccn_0p2_daily_avg.pdf)
[figure_mam_aci_850hPa_nc_daily_avg.pdf](https://github.com/E3SM-Project/scream/files/15441988/figure_mam_aci_850hPa_nc_daily_avg.pdf)

### Standalone Test Timings (Compy):

Label Time Summary:
PEM                 =  33.83 sec*proc (60 tests)
bfbhash             =   4.82 sec*proc (1 test)
check               =   6.40 sec*proc (1 test)
cld                 = 11128.37 sec*proc (81 tests)
cld_fraction        =   4.26 sec*proc (1 test)
cosp                =  53.82 sec*proc (4 tests)
diagnostics         = 494.54 sec*proc (4 tests)
dp                  = 1386.93 sec*proc (7 tests)
driver              = 13326.49 sec*proc (90 tests)
dynamics            = 7971.84 sec*proc (53 tests)
fail                =   5.29 sec*proc (1 test)
**mam4_aci**            = **2367.73** sec*proc **(35** **tests)**
mam4_optics         = 2755.88 sec*proc (21 tests)
nudging             = 717.58 sec*proc (4 tests)
p3                  = 11309.07 sec*proc (92 tests)
pg2                 = 497.80 sec*proc (10 tests)
physics             = 13090.69 sec*proc (153 tests)
rrtmgp              = 11457.10 sec*proc (90 tests)
shoc                = 11505.12 sec*proc (102 tests)
spa                 = 4093.34 sec*proc (34 tests)
surface_coupling    =  49.41 sec*proc (4 tests)
tms                 = 1884.11 sec*proc (14 tests)

### Important notes:
1.     The code that uses output from the heterogeneous freezing is currently missing in P3. We will port and add that code in a follow-up PR.
2.     This PR uses a branch of MAM4xx. Once we have all the changes from the branch merged in MAM4xx main, we will switch the submodule


components/eamxx/src/physics/shoc/disp/shoc_assign_2d_view_disp.cpp
```diff
@@ -0,0 +1,29 @@
...
+  const view_2d<Spack>&        output_view)
+{
+  using ExeSpace = typename KT::ExeSpace;
+
+  const auto nlev_packs = ekat::npack<Spack>(nlev);
+  const auto policy = ekat::ExeSpaceUtils<ExeSpace>::get_default_team_policy(shcol, nlev_packs);
+  Kokkos::parallel_for(policy, KOKKOS_LAMBDA(const MemberType& team) {
+    const Int i = team.league_rank();
+    Kokkos::parallel_for(Kokkos::TeamVectorRange(team, nlev_packs), [&] (const Int& k) {
+        output_view (i,k) = input_view(i,k);
```
singhbalwinder, line 23: We can probably use `deep_copy` here but I am not sure if we have a different way to use deep copy with spacks.

susburrows:
Great to see this come in, @singhbalwinder .  Is it correct that this should be labeled as "stealth"?

singhbalwinder:
Yes, this feature is not invoked until we include `mam4_aci` in the namelist via the `atmchange` command.

singhbalwinder:
@kaizhangpnl : I have added instructions to run a test with `mam4_aci` turned on in the PR description.

susburrows:
Thanks @singhbalwinder .  I added the "Stealth" tag (it seems we didn't have that tag before).

The description is clear.  I am wondering if there are any meaningful evaluation plots that could be provided to help with understanding the behavior of this feature inside of EAMxx, similar to the AOD plots that were shared for the optics code?

singhbalwinder:
>One more question: are there any new input datasets required for this PR (e.g., to add hygroscopicity properties)?

No, we didn't add any new datasets in this PR. 

We did add a new initial condition (IC) file for the unit tests. The old file had some missing values, so we added a new IC file. The new file has more realistic values directly from an E3SMv2 run with each column having a unique set of values.

singhbalwinder:
The evaluation process for the global simulations has just kicked off and we will be adding plots pretty soon of all meaningful evaluations we can do with this process.

susburrows:
Thanks @singhbalwinder !  Based on the information in the current PR description and the comments, as well as our discussion in the EAMxx eval call this morning, I added the WIP tag to this PR.  We discussed (and it's mentioned above) that there is some missing code that still needs to be added, as well as the timings and some basic sanity-test evaluation to demonstrate that the feature is functioning as expected at the process level.  In the meantime, it would be great if @kaizhangpnl could start looking at the code and providing any relevant feedback from the aerosol modeling perspective.

components/eamxx/src/physics/mam/eamxx_mam_aci_process_interface.cpp
```diff
@@ -0,0 +1,1299 @@
...
+        haero::Atmosphere haero_atm =
+            atmosphere_for_column(dry_atmosphere, icol);
+
+        // Update the updraft velocity needed by nucleation to be "wsubice"
+        // in the haero_atm object
+        haero_atm.updraft_vel_ice_nucleation = ekat::subview(wsubice, icol);
+
+        // All the output from this process is diagnotics; creates "diags" with
+        // nlev column length
+        mam4::Diagnostics diags(nlev);
```
odiazib, line 217: I wonder if we should remove `mam4::Diagnostics `or replace by a struct with diagnostic variables for each process? We have many variables in` mam4::Diagnostics` and I find sometime difficult to know why variables I need to allocate.

components/eamxx/src/physics/mam/eamxx_mam_aci_process_interface.cpp
```diff
@@ -0,0 +1,1299 @@
...
+  MAMAci::view_2d raertend     = dropmixnuc_scratch_mem[13];
+  MAMAci::view_2d qqcwtend     = dropmixnuc_scratch_mem[14];
+
+  //---------------------------------------------------------------------------
+  // ## Initialize the ndrop class.
+  //---------------------------------------------------------------------------
+  const int ntot_amode        = mam_coupling::num_aero_modes();
+  const int maxd_aspectype    = mam4::ndrop::maxd_aspectype;
+  const int nspec_max         = mam4::ndrop::nspec_max;
+  int nspec_amode[ntot_amode] = {};
```
odiazib, line 412: We may need to move the computation of these parameter inside of the parallel_for.

components/eamxx/src/physics/mam/eamxx_mam_aci_process_interface.cpp
```diff
@@ -0,0 +1,1299 @@
...
+  grid_ = grids_manager->get_grid("Physics");
+
+  // Name of the grid
+  const auto &grid_name = grid_->name();
+
+  ncol_ = grid_->get_num_local_dofs();       // Number of columns on this rank
+  nlev_ = grid_->get_num_vertical_levels();  // Number of levels per column
+
+  // Define the different field layouts that will be used for this process
+  using namespace ShortFieldTagsNames;
```
odiazib, line 736: We may need to update this code before merging into master; see PR[2776](https://github.com/E3SM-Project/scream/pull/2776)

components/eamxx/src/physics/mam/eamxx_mam_aci_process_interface.cpp
```diff
@@ -0,0 +1,1299 @@
...
+// ================================================================
+//  INITIALIZE_IMPL
+// ================================================================
+void MAMAci::initialize_impl(const RunType run_type) {
+  // ------------------------------------------------------------------------
+  // ## Runtime options
+  // ------------------------------------------------------------------------
+
+  wsubmin_ = m_params.get<double>("wsubmin");
+  top_lev_ = m_params.get<int>("top_level_mam4xx");
```
odiazib, line 937: How is `top_lev` in E3SM v2 obtained? is it also passed as input parameter?

components/eamxx/src/physics/mam/eamxx_mam_aci_process_interface.cpp
```diff
@@ -0,0 +1,1299 @@
...
+      get_field_out("hetfrz_contact_nucleation_tend").get_view<Real **>();
+  hetfrz_depostion_nucleation_tend_ =
+      get_field_out("hetfrz_depostion_nucleation_tend").get_view<Real **>();
+
+  //---------------------------------------------------------------------------------
+  // Allocate memory for the class members
+  // (Kokkos::resize only works on host to allocates memory)
+  //---------------------------------------------------------------------------------
+
+  Kokkos::resize(rho_, ncol_, nlev_);
```
odiazib, line 1052: Does it make a difference if this view is allocated as  `rho_= view_2d("rho", ncol_, nlev_); `

components/eamxx/src/physics/mam/mam_coupling.hpp
```diff
@@ -549,7 +550,8 @@ mam4::Prognostics interstitial_aerosols_for_column(const AerosolState& dry_aero,
     progs.n_mode_i[m] = ekat::subview(dry_aero.int_aero_nmr[m], column_index);
     for (int a = 0; a < num_aero_species(); ++a) {
       if (dry_aero.int_aero_mmr[m][a].data()) {
-        progs.q_aero_i[m][a] = ekat::subview(dry_aero.int_aero_mmr[m][a], column_index);
+        int prg_idx = mam4::utils::get_prognostics_index(m,a);//Index of HAERO's prognostics
```
odiazib, line 553: I believe, we do not need `mam4::utils::get_prognostics_index(m,a)`. The species ordering in each mode  is equal in  mam4xx and emaxx.

odiazib: In addition, see my comment in[ PR #308](https://github.com/eagles-project/mam4xx/pull/308) in mam4xx. 
> I reviewed this part of the code and found that we cannot use ` mam4::mode_aero_species(mode, species_index)` to obtain the index of the species in each mode. ` mam4::mode_aero_species(mode, species_index)` returns the aero_id, which is an enum from [here](https://github.com/eagles-project/mam4xx/blob/7d14b8a71166c48327f1839c46c0f974134a631f/src/mam4xx/aero_modes.hpp#L130).
> 
> Because mam4xx prognostics structure has same ordering of species per node that state_q and qqcw, we do not need to add this function; however, we could use the function that @jeff-cohere have implemented [here](https://github.com/eagles-project/mam4xx/blob/7d14b8a71166c48327f1839c46c0f974134a631f/src/mam4xx/aero_modes.hpp#L249) and it is used [here](https://github.com/eagles-project/mam4xx/blob/7d14b8a71166c48327f1839c46c0f974134a631f/src/mam4xx/utils.hpp#L163).

components/eamxx/src/physics/mam/mam_coupling.hpp
```diff
@@ -574,7 +576,8 @@ mam4::Prognostics aerosols_for_column(const AerosolState& dry_aero,
     progs.n_mode_c[m] = ekat::subview(dry_aero.cld_aero_nmr[m], column_index);
     for (int a = 0; a < num_aero_species(); ++a) {
       if (dry_aero.cld_aero_mmr[m][a].data()) {
-        progs.q_aero_c[m][a] = ekat::subview(dry_aero.cld_aero_mmr[m][a], column_index);
+        int prg_idx = mam4::utils::get_prognostics_index(m,a); //Index of HAERO's prognostics array
```
odiazib, line 579: we do not need `mam4::utils::get_prognostics_index(m,a)`.

components/eamxx/src/physics/mam/eamxx_mam_aci_process_interface.cpp
```diff
@@ -0,0 +1,1299 @@
...
+void compute_tke_using_w_sec(const haero::ThreadTeam &team,
+                             const MAMAci::const_view_2d w_sec, const int icol,
+                             const int nlev,
+                             // output
+                             MAMAci::view_2d tke) {
+  Kokkos::parallel_for(
+      Kokkos::TeamThreadRange(team, 0u, nlev + 1),
+      KOKKOS_LAMBDA(int kk) { tke(icol, kk) = (3.0 / 2.0) * w_sec(icol, kk); });
+}
+void compute_tke_using_w_sec(haero::ThreadTeamPolicy team_policy,
```
kaizhangpnl, line 98: This function name is the same as the above one (for KOKKOS). Is it a problem?

components/eamxx/src/physics/mam/eamxx_mam_aci_process_interface.cpp
```diff
@@ -0,0 +1,1299 @@
...
+  // Set to be zero at the surface by initialization.
+  Kokkos::parallel_for(
+      Kokkos::TeamThreadRange(team, 0u, top_lev), KOKKOS_LAMBDA(int kk) {
+        wsub(icol, kk)    = wsubmin;
+        wsubice(icol, kk) = 0.001;
+        wsig(icol, kk)    = 0.001;
+      });
+  Kokkos::parallel_for(
+      Kokkos::TeamThreadRange(team, top_lev, nlev), KOKKOS_LAMBDA(int kk) {
+        wsub(icol, kk) = haero::sqrt(0.5 * (tke(icol, kk) + tke(icol, kk + 1)) *
```
kaizhangpnl, line 126: Is the "tke" here at the interface (interpolated using mid-layer TKE from SHOC)?

kaizhangpnl (commented):
The vertical diffusion of aerosol tracers (interstitial and cloud-borne) need to be treated with care. Vertically diffusion of (interstitial aerosols and cloud droplet number) tracers is supposed to be done in SHOC by default, but the "ndrop" routine will also diffuse the aerosol tracers and cloud droplet number as in MAM/E3SM. According to our EAGLES MAMxx meeting discussion, it would be better to keep the vertical diffusion calculation  (for interstitial aerosols and cloud droplet number) in ndrop and turn it off in SHOC for initial porting and testing. This needs to be addressed either in this PR or a following PR. 

The comments here are only for routines inside the scream repo. Note that there are external routines that are important for this PR. Not sure what is the best way to review/comment on these routines. 

https://github.com/eagles-project/mam4xx/blob/main/src/mam4xx/ndrop.hpp
https://github.com/eagles-project/mam4xx/blob/main/src/mam4xx/hetfrz.hpp

components/eamxx/src/physics/mam/eamxx_mam_aci_process_interface.cpp
```diff
@@ -0,0 +1,1299 @@
...
+  Kokkos::resize(nidep_, ncol_, nlev_);
+
+  // number conc of ice nuclei due to meyers deposition [1/m3]
+  Kokkos::resize(nimey_, ncol_, nlev_);
+
+  // number of activated aerosol for ice nucleation(homogeneous frz only)[#/kg]
+  Kokkos::resize(naai_hom_, ncol_, nlev_);
+
+  //---------------------------------------------------------------------------------
+  // Diagnotics variables from the droplet activation scheme
```
kaizhangpnl, line 1084: It would be nice to have the CCN diagnostic output fields (e.g., CCN@0.1%, 0.2%, and 0.5%). Perhaps this can be addressed in a future PR.

components/eamxx/src/physics/mam/eamxx_mam_aci_process_interface.cpp
```diff
@@ -0,0 +1,1299 @@
...
+  dry_atm_.z_mid = buffer_.z_mid;
+
+  // total cloud fraction
+  dry_atm_.cldfrac = get_field_in("cldfrac_tot").get_view<const Real **>();
+
+  // computed updraft velocity
+  dry_atm_.w_updraft = buffer_.w_updraft;
+
+  // ------------------------------------------------------------------------
+  // Output fields to be used by other processes
```
kaizhangpnl, line 988: It would be nice to have the CCN diagnostic output fields (e.g., CCN@0.1%, 0.2%, and 0.5%). Perhaps this can be addressed in a future PR.

components/eamxx/cime_config/testdefs/testmods_dirs/scream/mam4xx/aci/shell_commands
```diff
@@ -0,0 +1,9 @@
+
+#Default scream has 10 tracers, MAM4xx adds another 31 making a total of 41 tracer
+#Set total number of tracers to 41. We are using append here as last entry wins while parsing xml options
+./xmlchange --append SCREAM_CMAKE_OPTIONS="SCREAM_NUM_TRACERS 41"
+
+$CIMEROOT/../components/eamxx/scripts/atmchange initial_conditions::Filename='$DIN_LOC_ROOT/atm/scream/init/screami_mam4xx_ne4np4L72_c20240208.nc' -b
```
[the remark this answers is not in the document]
singhbalwinder: @bartgol might have a better answer but I assumed this is for "buffering". I added this because sometimes these shell commands were not executed or were not executed at the right time in the `create_test` process. Adding the `-b` flag fixed that issue.

components/eamxx/cime_config/namelist_defaults_scream.xml
```diff
@@ -234,6 +234,12 @@ be lost if SCREAM_HACK_XML is not enabled.
       <Ckm type="real" doc="Eddy diffusivity coefficient for momentum">0.1</Ckm>
     </shoc>
 
+    <!-- MAM4xx-ACI -->
+    <mam4_aci inherit="atm_proc_base">
+      <wsubmin type="real" doc="Minimum diagnostic sub-grid vertical velocity">0.001</wsubmin>
```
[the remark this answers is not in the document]
singhbalwinder: That is a good point. I will fix it in the C++ code.

components/eamxx/src/physics/shoc/disp/shoc_assign_2d_view_disp.cpp
```diff
@@ -0,0 +1,29 @@
+#include "shoc_functions.hpp"
+
+#include "ekat/kokkos/ekat_subview_utils.hpp"
+
+namespace scream {
+namespace shoc {
+
+template<>
+void Functions<Real,DefaultDevice>
+::shoc_assign_2d_view_disp(
```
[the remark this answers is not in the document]
singhbalwinder: This function copies data from one view (or array) to another. There may be a better way to do it, that is why I left a comment here for others to take a look.

components/eamxx/src/physics/shoc/eamxx_shoc_process_interface.hpp
```diff
@@ -84,6 +84,9 @@ class SHOCMacrophysics : public scream::AtmosphereProcess
 
       Kokkos::parallel_for(Kokkos::TeamVectorRange(team, nlev_packs), [&] (const Int& k) {
 
+        
+        cldfrac_liq_prev(i,k)=cldfrac_liq(i,k);
```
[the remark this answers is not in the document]
singhbalwinder: `cldfrac_liq_prev` is the previous (or old value) of liquid stratiform cloud fraction. It is captured at each time step before the SHOC and cloud fraction processes modify it. The new (or modified) value is stored in `cldfrac_liq`. In the Fortran code, these are represented by `liqcldfo` and `liqcldf`.

components/eamxx/tests/multi-process/physics_only/mam/shoc_cldfrac_mam4_aci_p3/output.yaml
```diff
@@ -0,0 +1,44 @@
...
+  - eddy_diff_mom
+  - horiz_winds
+  - sgs_buoy_flux
+  - tke
+  - inv_qc_relvar
+  - pbl_height
+  # CLD
+  - cldfrac_ice
+  - cldfrac_tot
+  # P3
```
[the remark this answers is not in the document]
singhbalwinder: This is the standalone multi-process test. We might want to do that in the CIME simulation.

components/eamxx/src/physics/mam/eamxx_mam_aci_process_interface.cpp
singhbalwinder: No, C++ (like Fortran) allows function overloading. In Fortran, we have to use an "interface" but C++ allows it without interfaces.

components/eamxx/src/physics/mam/eamxx_mam_aci_process_interface.cpp
singhbalwinder: Yes, `tke` is assumed to be at interfaces here. The level loop here goes from `top_lev` to `nlev` and we use `kk+1` index of `tke` to fetch data from the last interface level.

components/eamxx/src/physics/mam/eamxx_mam_aci_process_interface.cpp
singhbalwinder: Yes, we can do that as a diagnostics in a future PR.

components/eamxx/cime_config/testdefs/testmods_dirs/scream/mam4xx/aci/shell_commands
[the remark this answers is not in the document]
bartgol: That's correct. The `-b` option is basically needed only for testmods: all atmchanges MUST be buffered and NOT directly applied (since case.setup has not run yet, so we can't yet do anything). Buffering means they are stored in a special XML CIME var, which later allows buildnml to replay all the atmchanges if an xml change happens.

components/eamxx/src/physics/mam/eamxx_mam_aci_process_interface.cpp
```diff
@@ -0,0 +1,1326 @@
...
+      // mass mixing ratio of species "a" of mode "m"
+      auto aero_mmr = dry_aero.int_aero_mmr[m][a];
+
+      if(aero_mmr.data()) {
+        Kokkos::parallel_for(
+            team_policy, KOKKOS_LAMBDA(const haero::ThreadTeam &team) {
+              const int icol = team.league_rank();
+              // update values for all levs at this column
+              update_interstitial_aerosols_levs(team, nlev, icol, s_idx, dt,
+                                                ptend_q,
```
bartgol, line 595: I am not 100% sure that it's ok to capture by value a static C array. I've had errors in the past when using them across a kernel boundary, so we should be extra safe and check with cuda compute-sanitizer as well as running on AMD gpus as well. Maybe I'm misremembering, or maybe it was a different use case. But just to be safe...

bartgol: The alternative to static arrays is to use a "View of Views", which is a bit annoying, but we already do that in Homme, for instance.

components/eamxx/src/physics/mam/eamxx_mam_aci_process_interface.cpp
```diff
@@ -0,0 +1,1326 @@
...
+  naai_ = get_field_out("ni_activated").get_view<Real **>();
+
+  // droplet activation output
+  tendnd_ = get_field_out("nc_nuceat_tend").get_view<Real **>();
+
+  // interstitial and cloudborne aerosol tracers of interest: mass (q) and
+  // number (n) mixing ratios
+  for(int m = 0; m < mam_coupling::num_aero_modes(); ++m) {
+    // interstitial aerosol tracers of interest: number (n) mixing ratios
+    const char *int_nmr_field_name = mam_coupling::int_aero_nmr_field_name(m);
```
bartgol, line 1027: Any reason to use C strings instead of std::string?

components/eamxx/src/physics/mam/eamxx_mam_aci_process_interface.cpp
```diff
@@ -0,0 +1,1326 @@
...
+      KT::ExeSpace>::get_thread_range_parallel_scan_team_policy(ncol_, nlev_);
+
+  // preprocess input -- needs a scan for the calculation of local derivied
+  // quantities
+  Kokkos::parallel_for("preprocess", scan_policy, preprocess_);
+  Kokkos::fence();
+
+  haero::ThreadTeamPolicy team_policy(ncol_, Kokkos::AUTO);
+
+  compute_w0_and_rho(team_policy, dry_atm_, top_lev_, nlev_,
```
bartgol, line 1222: I'm curious: why are you dispatching several kernels rather than a single monolithic one? You may be able to use smaller temporaries if you have a single kernel instead of many.

components/eamxx/src/physics/shoc/disp/shoc_assign_2d_view_disp.cpp
bartgol: Deep copy works fine with packs. We should not reinvent the wheel, and use Kokkos::deep_copy instead of this routine.

components/eamxx/src/physics/shoc/eamxx_shoc_process_interface.hpp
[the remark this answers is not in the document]
bartgol: I don't see how shoc uses this "prev" variable. Is it really needed? @tcclevenger .

components/eamxx/src/physics/shoc/impl/shoc_main_impl.hpp
```diff
@@ -310,6 +311,12 @@ void Functions<S,D>::shoc_main_internal(
           workspace,                      // Workspace
           pblh);                          // Output
 
+  // Assign tkh to the tkh output variable
+  const Int nlev_pack = ekat::npack<Spack>(nlev);
+  Kokkos::parallel_for(Kokkos::TeamVectorRange(team, nlev_pack), [&] (const Int& k) {
+    tkh_out(k) = tkh(k);
```
bartgol, line 317: If you have tkh as inout arg (b/c presumably you need it outside), then there's no point in using a slot from the workspace manager for `tkh`. We could remove that slot, and rename the inout arg `tkh`...

@tcclevenger may also know whether removing this slot would entail a reduction in the overall WSM needs for shoc or not...

bartgol (changes requested):
I think there is at least one needed fix for a fcn called in a kernel. The other are suggestions/comments, which you should feel free to disagree with.

components/eamxx/src/physics/shoc/eamxx_shoc_process_interface.hpp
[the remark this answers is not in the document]
singhbalwinder: I added the `prev` variable as I need it for the ACI process.

kaizhangpnl:
> The vertical diffusion of aerosol tracers and cloud droplet number needs to be treated with care. Vertically diffusion of (interstitial aerosols and cloud droplet number) tracers is supposed to be done in SHOC by default, but the "ndrop" routine will also diffuse the aerosol tracers and cloud droplet number as in MAM/E3SM. According to our discussion in the EAGLES MAMxx meeting, it's better to keep the vertical diffusion calculation (for interstitial aerosols and cloud droplet number) in "ndrop" and turn it off in SHOC for initial porting and testing. (Note that the vertical diffusion of cloud-borne aerosols is considered in "ndrop", but not in SHOC since cloud-borne aerosols are not in the tracer array). This needs to be addressed either in this PR or an additional PR.
> 
> The comments here are only for routines inside the scream repo. Note that there are external routines that are important for this PR. Not sure what is the best way to review/comment on these routines.
> 
> https://github.com/eagles-project/mam4xx/blob/main/src/mam4xx/ndrop.hpp https://github.com/eagles-project/mam4xx/blob/main/src/mam4xx/hetfrz.hpp

There are some hard-coded indices in ndrop.hpp: 

https://github.com/eagles-project/mam4xx/blob/42ea4d6dd65d560f79554b2a3c24d18c4ad9dfa7/src/mam4xx/ndrop.hpp#L85
https://github.com/eagles-project/mam4xx/blob/42ea4d6dd65d560f79554b2a3c24d18c4ad9dfa7/src/mam4xx/ndrop.hpp#L102

Do they match the tracer indices in EAMxx (e.g., 41 tracers in EAMxx vs 40 in EAMv2)?

components/eamxx/src/physics/shoc/eamxx_shoc_process_interface.cpp
```diff
@@ -80,16 +80,19 @@ void SHOCMacrophysics::set_grids(const std::shared_ptr<const GridsManager> grids
   add_field<Required>("phis",           scalar2d    , m2/s2, grid_name, ps);
 
   // Input/Output variables
-  add_field<Updated>("tke",           scalar3d_mid, m2/s2,   grid_name, "tracers", ps);
-  add_field<Updated>("horiz_winds",   vector3d_mid,   m/s,     grid_name, ps);
-  add_field<Updated>("sgs_buoy_flux", scalar3d_mid, K*(m/s), grid_name, ps);
-  add_field<Updated>("eddy_diff_mom", scalar3d_mid, m2/s,    grid_name, ps);
-  add_field<Updated>("qc",            scalar3d_mid, Qunit,   grid_name, "tracers", ps);
-  add_field<Updated>("cldfrac_liq",   scalar3d_mid, nondim,  grid_name, ps);
```
tcclevenger, line 88: I think these layout changes might be reverting to an old version of master. Is this not causing errors?

components/eamxx/src/physics/shoc/eamxx_shoc_process_interface.cpp
```diff
@@ -80,16 +80,19 @@ void SHOCMacrophysics::set_grids(const std::shared_ptr<const GridsManager> grids
...
+  add_field<Updated>("cldfrac_liq",   scalar3d_layout_mid, nondim,  grid_name, ps);
 
   // Output variables
-  add_field<Computed>("pbl_height",    scalar2d    , m,           grid_name);
-  add_field<Computed>("inv_qc_relvar", scalar3d_mid, Qunit*Qunit, grid_name, ps);
+  add_field<Computed>("pbl_height",       scalar2d_layout_col, m,           grid_name);
+  add_field<Computed>("inv_qc_relvar",    scalar3d_layout_mid, Qunit*Qunit, grid_name, ps);
+  add_field<Computed>("eddy_diff_heat",   scalar3d_layout_mid, m2/s,        grid_name, ps);
+  add_field<Computed>("w_variance",       scalar3d_layout_mid, m2/s2,       grid_name, ps);
+  add_field<Computed>("cldfrac_liq_prev", scalar3d_layout_mid, nondim,  grid_name, ps);
```
tcclevenger, line 95: Same with these layouts.

singhbalwinder: Yes, I rebased my branch and accidentally copied the old code. I have fixed it now and currently testing the code again.

components/eamxx/src/physics/shoc/impl/shoc_main_impl.hpp
tcclevenger: Yeah, I believe this reduces WSM shoc needs since `tkh` is global to `shoc_main()`, which means when we get rid of it we will need to reduce the WSM request size for the internal buffer in `eamxx_shoc_process_interface`. 

Also for small kernels I'm guessing we can remove the `tkh` temporary.

components/eamxx/cime_config/namelist_defaults_scream.xml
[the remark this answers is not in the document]
kaizhangpnl: The (characteristic) updraft velocity (wsub) is dynamically calculated in the model. wsubmin is used to apply a lower bound for wsub. In E3SMv2 aerosol activation code (ndrop), there is an additional lower bound 0.1 m/s for wsub. It would be nice to have a single namelist switch to control this lower bound, which is often used for model tuning and testing.

singhbalwinder:
>There are some hard-coded indices in ndrop.hpp:

>https://github.com/eagles-project/mam4xx/blob/42ea4d6dd65d560f79554b2a3c24d18c4ad9dfa7/src/mam4xx/ndrop.hpp#L85
https://github.com/eagles-project/mam4xx/blob/42ea4d6dd65d560f79554b2a3c24d18c4ad9dfa7/src/mam4xx/ndrop.hpp#L102

>Do they match the tracer indices in EAMxx (e.g., 41 tracers in EAMxx vs 40 in EAMv2)?

That's a good question. EAMxx has 41 tracers but when we call MAM4xx routines, we construct an array like EAMv2's `state_q` with 40 species. The hardcoded indices should work as MAM4xx sees an array similar to EAMv2's `state%q`.

I ran validation tests and the results were exactly the same (up to single precision) for all the species (cloud-borne and interstitial), so I am positive that these indices are correct.

components/eamxx/src/physics/shoc/disp/shoc_assign_2d_view_disp.cpp
singhbalwinder: I have completely removed this function and used deep copy

components/eamxx/src/physics/shoc/impl/shoc_main_impl.hpp
```diff
@@ -643,6 +654,7 @@ Int Functions<S,D>::shoc_main(
     const auto shoc_cldfrac_s = ekat::subview(shoc_input_output.shoc_cldfrac, i);
     const auto shoc_ql_s      = ekat::subview(shoc_input_output.shoc_ql, i);
     const auto shoc_ql2_s     = ekat::subview(shoc_output.shoc_ql2, i);
+    const auto tkh_out_s          = ekat::subview(shoc_output.tkh, i);
```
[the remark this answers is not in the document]
singhbalwinder: Done.

components/eamxx/cime_config/namelist_defaults_scream.xml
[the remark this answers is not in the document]
bartgol: So `wsub>=max(wsubmin,0.1)`? Then it makes no sense to set `wsubmin` to `0.001` in the defaults, no? I would just set `0.1`.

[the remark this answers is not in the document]
kaizhangpnl: Yes. The changes needed are 1) remove the wsub lower bound (0.1) in ndrop; and 2) set wsubmin = 0.1 in the default namelist.

components/eamxx/src/physics/mam/eamxx_mam_aci_process_interface.cpp
odiazib: @bartgol, will std::string produce the warning of calling host function from device?

bartgol: This is not device code though, so std::string is fine.

components/eamxx/cime_config/namelist_defaults_scream.xml
```diff
@@ -489,6 +489,31 @@ be lost if SCREAM_HACK_XML is not enabled.
...
+    <nacl_c2 type="real">0.0</nacl_c2>
+    <nacl_c3 type="real">0.0</nacl_c3>
+    <mom_c1 type="real">0.0</mom_c1>
+    <mom_c2 type="real">0.0</mom_c2>
+    <mom_c3 type="real">0.0</mom_c3>
+    <mom_c4 type="real">0.0</mom_c4>
+    <num_c1 type="real">0.0</num_c1>
+    <num_c2 type="real">0.0</num_c2>
+    <num_c3 type="real">0.0</num_c3>
+    <num_c4 type="real">0.0</num_c4>
```
singhbalwinder, line 516: Note that this is a temporary change for testing purposes only. I will remove it.

components/eamxx/src/physics/shoc/eamxx_shoc_process_interface.cpp
```diff
@@ -332,11 +336,12 @@ void SHOCMacrophysics::initialize_impl (const RunType run_type)
   // Output Variables
   output.pblh     = get_field_out("pbl_height").get_view<Real*>();
   output.shoc_ql2 = shoc_ql2;
+  output.tkh      = get_field_out("eddy_diff_heat").get_view<Spack**>();
 
   // Ouput (diagnostic)
   history_output.shoc_mix  = m_buffer.shoc_mix;
   history_output.isotropy  = m_buffer.isotropy;
-  history_output.w_sec     = m_buffer.w_sec;
```
tcclevenger, line 339: We need to remove this from the buffer if we aren't using it.

components/eamxx/src/physics/shoc/impl/shoc_main_impl.hpp
```diff
@@ -119,6 +119,7 @@ void Functions<S,D>::shoc_main_internal(
   // Output Variables
   Scalar&                      pblh,
   const uview_1d<Spack>&       shoc_ql2,
+  const uview_1d<Spack>&       tkh_out,
```
tcclevenger, line 122: We can just call this `tkh` and remove the other `tkh` from the workspace manager. Then no need to copy down below from `tkh`->`tke_out`.

components/eamxx/src/physics/shoc/impl/shoc_main_impl.hpp
```diff
@@ -371,6 +378,7 @@ void Functions<S,D>::shoc_main_internal(
   // Output Variables
   const view_1d<Scalar>&      pblh,
   const view_2d<Spack>&       shoc_ql2,
+  const view_2d<Spack>&       tkh_out,
```
tcclevenger, line 381: Same here, call this `tkh` and remove the other `tkh` input.

components/eamxx/src/physics/shoc/impl/shoc_main_impl.hpp
```diff
@@ -565,6 +573,9 @@ void Functions<S,D>::shoc_main_internal(
                kbfs,shoc_cldfrac,              // Input
                workspace_mgr,                  // Workspace mgr
                pblh);                          // Output
+  
+  // Assign tkh to the tkh output variable
+  Kokkos::deep_copy(tkh_out,tkh);
```
tcclevenger, line 578: This deep copy will be unecessary when `tkh_out`->`thk` is the only `tkh`.

components/eamxx/src/physics/shoc/shoc_functions_f90.cpp
```diff
@@ -2910,7 +2909,7 @@ Int shoc_main_f(Int shcol, Int nlev, Int nlevi, Real dtime, Int nadv, Int npbl,
     tabs    ("shoc_tabs", shcol, nlev_packs),
     dz_zt   ("dz_zt",   shcol, nlevi_packs),
     dz_zi   ("dz_zi",   shcol, nlevi_packs),
-    tkhv    ("tkh",     shcol, nlevi_packs);
+    tkhv    ("tkh",     shcol, nlev_packs);
```
tcclevenger, line 2912: We can remove this from `SHOCTemporaries`.

kaizhangpnl:
With commits [42d657f](https://github.com/E3SM-Project/scream/pull/2794/commits/42d657f3e04f0fbb7c4955a6e51065b6846e5eae) and  [c022bf2](https://github.com/E3SM-Project/scream/pull/2794/commits/c022bf2d34c7fbbcf3cf2eab9030e7e266a5cfec), all previous fields with missing values look good now: 
https://github.com/kaizhangpnl/MAMxx_share/tree/main/run/aci/test_20240512

tcclevenger (commented):
SHOC changes look good.

AaronDonahue:
@singhbalwinder , is this PR ready for re-review?

singhbalwinder:
@AaronDonahue: We are evaluating global simulation results. Most of the ACI output fields look okay but there are a few fields that are stronger than expected. These can be due to multiple reasons since we are missing some processes and some processes are done twice (this will be fixed in a follow-up PR). I am currently looking into it with help from @kaizhangpnl, @polunma, and Jerome. Once we fully evaluate and understand it, I will ask for a re-review before merging.

components/eamxx/cime_config/namelist_defaults_scream.xml
singhbalwinder: It looks like we need to keep this change unless there is a way to do the same in the code. I tried initializing all these variables in `initialize_impl` but it didn't work.

components/eamxx/src/physics/mam/eamxx_mam_aci_process_interface.cpp
singhbalwinder: I have already rebased the code and fixed these field layout tags.

components/eamxx/src/physics/mam/eamxx_mam_aci_process_interface.cpp
singhbalwinder: We will eventually do this cleanup. For now, let's keep the existing data structures as is.

components/eamxx/src/physics/mam/eamxx_mam_aci_process_interface.cpp
singhbalwinder: I don't think so. @overfelt  uses resize and I think both methods may be equivalent.

components/eamxx/src/physics/mam/mam_coupling.hpp
singhbalwinder: You are right. I have removed it.

components/eamxx/src/physics/mam/eamxx_mam_aci_process_interface.hpp
```diff
@@ -0,0 +1,270 @@
...
+#define EAMXX_MAM_ACI_HPP
+
+// For MAM4 aerosol configuration
+#include <physics/mam/mam_coupling.hpp>
+
+// For declaring ACI class derived from atm process class
+#include <share/atm_process/atmosphere_process.hpp>
+
+// For EKAT units package
+#include "ekat/util/ekat_units.hpp"
```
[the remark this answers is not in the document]
singhbalwinder: Thanks. I have moved this to the CPP file.

components/eamxx/src/physics/mam/eamxx_mam_aci_process_interface.cpp
singhbalwinder: I tested it on both GPUs and CPUs. I get the right values whether these are in or out of the Kokkos loop. As we discussed, I am planning to keep the code as is for now.

components/eamxx/src/physics/mam/mam_coupling.hpp
singhbalwinder: I have removed it.

kaizhangpnl (dismissed):

susburrows:
Thanks for adding your review, @kaizhangpnl .  The evaluation outputs look very reasonable at this stage.  @singhbalwinder , have all the technical comments now been addressed?

components/eamxx/src/physics/mam/eamxx_mam_aci_process_interface.cpp
singhbalwinder: I talked with @overfelt and he suggested that we may need to revisit this when we optimize the code. If ACI is not taking too much memory, we can have these small kernels launched simultaneously (i.e., without Kokkos::fence) and improve performance. If we are limited by memory, we might combine these kernels.

singhbalwinder:
@susburrows: I have addressed all the major comments. I will address the minor comments in a follow-up PR (see git issue: #2850 ). I have created another issue (see #2849 )as a reminder to fix the double counting of aerosol species mixing by SHOC.

I am waiting for the test results. Once all the expected tests `PASS`, I will ask to merge this PR (unless there are additional reviews).

components/eamxx/src/physics/mam/eamxx_mam_aci_process_interface.cpp
singhbalwinder: I will fix it in a follow-up PR (see issue #2850 )

components/eamxx/src/physics/mam/eamxx_mam_aci_process_interface.cpp
singhbalwinder: I will fix it in a follow-up PR (see issue https://github.com/E3SM-Project/scream/issues/2850 )

components/eamxx/src/physics/mam/eamxx_mam_aci_process_interface.hpp
```diff
@@ -0,0 +1,270 @@
...
+  // droplet activation inputs and outputs
+  view_2d kvh_int_;  // Eddy diffusivity of heat at the interfaces
+  const_view_2d liqcldf_;
+  const_view_2d liqcldf_prev_;
+  const_view_2d kvh_mid_;  // Eddy diffusivity of heat at the midpoints
+
+  view_2d cloud_frac_;
+  view_2d cloud_frac_prev_;
+  view_2d qcld_;
+  view_2d ptend_q_[mam4::aero_model::pcnst];
```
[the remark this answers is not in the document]
singhbalwinder: We will revisit this (see issue #2850 )

components/eamxx/src/physics/mam/eamxx_mam_aci_process_interface.cpp
overfelt: Yes, both methods are the same since we know that rho_ is a view_2d.  In general Kokkos::resize would work if it was a view of ints or some other declaration that would need to be duplicated if resized without using Kokkos::resize.

components/eamxx/src/physics/mam/eamxx_mam_aci_functions.hpp
```diff
@@ -0,0 +1,726 @@
...
+  // Get physical constants
+  using C                      = physics::Constants<Real>;
+  static constexpr auto gravit = C::gravit;  // Gravity [m/s2]
+  // Gas constant for dry air [J/(kg*K) or J/Kg/K]
+  static constexpr auto rair = C::Rair;
+  Kokkos::parallel_for(
+      Kokkos::TeamThreadRange(team, 0u, top_lev), KOKKOS_LAMBDA(int kk) {
+        w0(icol, kk)  = 0;
+        rho(icol, kk) = -999.0;
+      });
```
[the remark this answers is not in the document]
overfelt: I'll own up to this one.  I thought that since the loop indexes do not overlap, two loops looked better than one loop with an if statement.  Just a style choice.

components/eamxx/src/physics/shoc/eamxx_shoc_process_interface.cpp
```diff
@@ -85,6 +85,9 @@ void SHOCMacrophysics::set_grids(const std::shared_ptr<const GridsManager> grids
   // Output variables
   add_field<Computed>("pbl_height",    scalar2d    , m,            grid_name);
   add_field<Computed>("inv_qc_relvar", scalar3d_mid, pow(kg/kg,2), grid_name, ps);
+  add_field<Computed>("eddy_diff_heat",   scalar3d_mid, m2/s,        grid_name, ps);
+  add_field<Computed>("w_variance",       scalar3d_mid, m2/s2,       grid_name, ps);
+  add_field<Computed>("cldfrac_liq_prev", scalar3d_mid, nondim,      grid_name, ps);
```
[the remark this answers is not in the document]
singhbalwinder: I agree, it will be great if we can get these quantities using a function with some inputs rather than adding code in other processes. It makes the logic simpler. 
In ACI, we need `tke` and we derive it from `w_sec_`. We should probably use the `tke` directly as it is one of SHOC's outputs. I won't change the code at this time as it will trigger another cycle of evaluations. But I will do some side calculations to see how close the results are and later use `tke` directly if we can.

components/eamxx/src/physics/mam/eamxx_mam_aci_functions.hpp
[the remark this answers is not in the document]
singhbalwinder: I guess avoiding the if-check is the main driver here (@odiazib and @overfelt can comment more about it). Although, it may not make a dramatic change in performance, avoiding if-check should help these kernels to run faster.

components/eamxx/src/physics/mam/eamxx_mam_aci_process_interface.cpp
```diff
@@ -0,0 +1,676 @@
...
+namespace scream {
+MAMAci::MAMAci(const ekat::Comm &comm, const ekat::ParameterList &params)
+    : AtmosphereProcess(comm, params) {
+  // Asserts for the runtime or namelist options
+  EKAT_REQUIRE_MSG(m_params.isParameter("wsubmin"),
+                   "ERROR: wsubmin is missing from mam_aci parameter list.");
+  EKAT_REQUIRE_MSG(
+      m_params.isParameter("top_level_mam4xx"),
+      "ERROR: top_level_mam4xx is missing from mam_aci parameter list.");
+}
```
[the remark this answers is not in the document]
singhbalwinder: I didn't know that these checks are there by default. Thanks. I added these mimicking other interfaces.

components/eamxx/src/physics/mam/eamxx_mam_aci_functions.hpp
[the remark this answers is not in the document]
overfelt: Yes, the 0u is assumed.  

The use of KOKKOS_LAMBDA evaluates to [=] __host__ __device__ but since we are already in device code, the Cuda decorations might be redundant and [&] better if there is no warning about calling _host_ from _device_? I'm not sure.

components/eamxx/src/physics/shoc/eamxx_shoc_process_interface.cpp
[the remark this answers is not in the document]
singhbalwinder: ACI codes need `tke` at the interfaces. I think SHOC has an internal `tke` at the interfaces.

components/eamxx/src/physics/mam/eamxx_mam_aci_functions.hpp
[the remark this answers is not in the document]
singhbalwinder: I changed the inner kernels as I got the following error on GPU tests:
```
An extended __host__ __device__ lambda cannot be defined inside an extended __host__ __device__  lambda expression("operator()").
```
I googled and google suggested to use `[=]` instead of `KOKKOS_LAMBDA`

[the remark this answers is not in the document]
bartgol: Since you are already on device, you should use `[&]` instead of `[=]`. The latter is _needed_ for the outermost lambda (you can't take references of host objects). But once you are on device, all the visible stuff is already on device, so might as well use references rather than copies.

components/eamxx/src/physics/shoc/eamxx_shoc_process_interface.cpp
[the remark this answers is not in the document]
bartgol: If the prev field is only exposed b/c it's needed for restarts, then it should be created separately, and declared as internal field. You can see an example in [homme](https://github.com/E3SM-Project/scream/blob/6557feb759786a7ed602a22d350e16bdca11ef77/components/eamxx/src/dynamics/homme/eamxx_homme_process_interface.cpp#L241).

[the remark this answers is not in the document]
bartgol: NOTE: internal fields are treated by the AD as follows:
- should not be used by other atm procs (errors out if they are)
- are added to the RESTART group, so that they are written to restart file
- they are not needed for normal initialization (not looked for in the IC file).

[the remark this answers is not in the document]
bartgol: Ah, I didn't realize what's happening. Here we have SHOC computing a field just for the sake of ACI. That is indeed underwhelming. Can't ACI compute this field?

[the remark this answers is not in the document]
singhbalwinder: ACI needs to know the values of liquid cloud fraction before SHOC alters it. The `_prev` field is not added for the restarts, it is added because ACI needs it.

`w_variance` is already computed by SHOC. I just exposed it as ACI codes need to use it. I mimicked what was being done on EAMv2. In EAMv2, CLUBB computes w_variance and the ACI codes use it. We can revisit it as we may be able to use the `tke` directly in the ACI codes.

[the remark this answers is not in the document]
bartgol: I don't think using tendencies is a good approach, as it requires one to remember adding the tendency request to shoc from input file (besides adding more rounding errors).

Looking at mam4xx's `dropmixnuc` inline docs, I see that the variable `cldo` (the _prev var here) is "cloud fraction on previous time step". I'm a bit confused then: is it "from the previous timestep" or "whatevet was before shoc"? If the latter, why does ACI needs to know the value right before shoc? Can you explain what's the rationale behind it?

I want to try to avoid creating teleconnections between different packages, so understanding the underlying reason may help find a better way...

overfelt (approved):

components/eamxx/src/physics/mam/eamxx_mam_aci_functions.hpp
```diff
@@ -0,0 +1,726 @@
...
+      Kokkos::TeamThreadRange(team, top_lev, nlev), KOKKOS_LAMBDA(int kk) {
+        aitken_dry_dia(icol, kk) = dgnum(icol, aitken_idx, kk);
+      });
+}
+void compute_aitken_dry_diameter(haero::ThreadTeamPolicy team_policy,
+                                 const MAMAci::const_view_3d dgnum,
+                                 const int top_lev, const int nlev,
+                                 // output
+                                 MAMAci::view_2d aitken_dry_dia) {
+  Kokkos::parallel_for(
```
bartgol, line 156: Does MAM _need_ the output to be a contiguous view? Otherwise, why not just do
```
aitken_dry_dia = ekat::subview_1(dgnum,mam4::ModeIndex::Aitken);
```
at the call site? No need to a) create an additional view and b) dispatch a kernel.

That is, unless you _need_ the output view to be contiguous (which you probably don't, since all cols are treated independently anyways inside kernels).

components/eamxx/src/physics/mam/eamxx_mam_aci_functions.hpp
```diff
@@ -0,0 +1,726 @@
...
+}
+void compute_recipical_pseudo_density(haero::ThreadTeamPolicy team_policy,
+                                      MAMAci::const_view_2d pdel,
+                                      const int nlev,
+                                      // output
+                                      MAMAci::view_2d rpdel) {
+  Kokkos::parallel_for(
+      team_policy, KOKKOS_LAMBDA(const haero::ThreadTeam &team) {
+        const int icol = team.league_rank();
+        compute_recipical_pseudo_density(team, pdel, icol, nlev,
```
bartgol, line 298: Maybe it's just me, but I see you have plenty of functions that have 2 versions, one to dispatch the outer loop, and one to be called from inside the loop. If the inner one is only used in the outer one, you could consider just using a lambda in the outer one, to keep the code smaller.

OTOH, dispatching lots of kernels is not great. If in all outer loops you just do
```
  const int icol = team.league_rank();
  inner_function_name(...);
```
you could consider merging all these kernels together, and doing all these calculations at once, if possible (it may not be, if they must happen at different times during the ACI step).

components/eamxx/src/physics/mam/eamxx_mam_aci_functions.hpp
```diff
@@ -0,0 +1,726 @@
...
+        mam4::Prognostics progs_at_col = aerosols_for_column(dry_aero, icol);
+
+        // get atmospheric quantities
+        haero::Atmosphere haero_atm =
+            atmosphere_for_column(dry_atmosphere, icol);
+
+        // Construct state_q (interstitial) and qqcw (cloud borne) arrays
+        Kokkos::parallel_for(
+            Kokkos::TeamThreadRange(team, 0u, mam4::ndrop::pver),
+            [=](int klev) {
```
bartgol, line 461: Prefer `[&]` over `[=]` for inner loops lambdas. Also, you should use `TeamVectorRange` rather than `TeamThreadRange`. The former ensures each iteration is handled by a different thread in a warp, while the latter would be problematic if by chance the team policy changes and has a vector_length!=1.

components/eamxx/src/physics/mam/eamxx_mam_aci_functions.hpp
```diff
@@ -0,0 +1,726 @@
...
+              for(int icnst = istart; icnst < mam4::aero_model::pcnst;
+                  ++icnst) {
+                state_q_work_loc(icol, klev, icnst) = state_q_at_lev_col[icnst];
+              }
+
+              // get qqcw at a grid cell (col,lev)
+              // NOTE: The layout for qqcw array is based on mam_idx in
+              // dropmixnuc. To mimic that, we are using the following for-loops
+              int ind_qqcw = 0;
+              for(int m = 0; m < mam_coupling::num_aero_modes(); ++m) {
```
bartgol, line 483: These for loops will deteriorate performance on GPU, due to non-coalesced access. You should try to rework the code, with multiple TeamVectorRange loops (that's fine, especially if there are no deps among them, so no barriers needed). E.g., this for loop could be a TVR over nlevs*num_aero_mods, where inside you do modulo arithmetic to recover ilev/imode (in a way that still ensures coalesced access)

components/eamxx/src/physics/mam/eamxx_mam_aci_functions.hpp
```diff
@@ -0,0 +1,726 @@
...
+            ekat::subview(eddy_diff_km, icol), ekat::subview(qncld, icol),
+            ekat::subview(srcn, icol), ekat::subview(source, icol),
+            ekat::subview(dz, icol), ekat::subview(csbot_cscen, icol),
+            ekat::subview(raertend, icol), ekat::subview(qqcwtend, icol));
+      });
+}
+
+// Update cloud borne aerosols
+void update_cloud_borne_aerosols(
+    haero::ThreadTeamPolicy team_policy,
```
bartgol, line 532: Why passing a policy, if all work is done serially on host?

components/eamxx/src/physics/mam/eamxx_mam_aci_functions.hpp
```diff
@@ -0,0 +1,726 @@
...
+      if(dry_aero.cld_aero_mmr[m][a].data()) {
+        Kokkos::deep_copy(dry_aero.cld_aero_mmr[m][a], qqcw_fld_work[ind_qqcw]);
+        ++ind_qqcw;
+      }
+    }
+  }
+}
+
+// Update interstitial aerosols using tendencies - levels
+void update_interstitial_aerosols_levs(
```
bartgol, line 550: This needs to be marked as a KOKKOS_INLINE_FUNCTION, since you are assuming to be already on device

components/eamxx/src/physics/mam/eamxx_mam_aci_functions.hpp
```diff
@@ -0,0 +1,726 @@
...
+    const haero::ThreadTeam &team, const int nlev, const int icol,
+    const int s_idx, const Real dt,
+    const MAMAci::view_2d ptend_q[mam4::aero_model::pcnst],
+    // output
+    MAMAci::view_2d aero_mr) {
+  Kokkos::parallel_for(Kokkos::TeamThreadRange(team, nlev), [=](int kk) {
```
bartgol, line 556: Prefer `[&]` over `[=]` when you are already on device.

components/eamxx/src/physics/mam/eamxx_mam_aci_process_interface.cpp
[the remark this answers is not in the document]
bartgol: The check is done the first time you access the parameters (unless you provide a default when yo call `get`). I think this is fine though. This check happens early on in init, so it allows to error out sooner.

components/eamxx/src/physics/mam/eamxx_mam_aci_process_interface.cpp
```diff
@@ -0,0 +1,676 @@
...
+  // Layer thickness(pdel) [Pa] at midpoints
+  add_field<Required>("pseudo_density", scalar3d_layout_mid, Pa, grid_name);
+
+  // planetary boundary layer height
+  add_field<Required>("pbl_height", scalar2d_layout_col, m, grid_name);
+
+  // cloud fraction [nondimensional] computed by eamxx_cld_fraction_process
+  add_field<Required>("cldfrac_tot", scalar3d_layout_mid, nondim, grid_name);
+
+  auto m2 = m * m;
```
bartgol, line 118: If you wand to see `m^2` in the output, use `auto m2 = pow(m,2);`

components/eamxx/src/physics/mam/eamxx_mam_aci_process_interface.cpp
```diff
@@ -0,0 +1,676 @@
...
+  compute_tke_using_w_sec(team_policy, w_sec_int_, nlev_,
+                          // output
+                          tke_);
+
+  Kokkos::fence();  // wait for for tke_ to be computed.
+  compute_subgrid_scale_velocities(team_policy, tke_, wsubmin_, top_lev_, nlev_,
+                                   // output
+                                   wsub_, wsubice_, wsig_);
+
+  Kokkos::fence();  // wait for wsig_ to be computed.
```
bartgol, line 577: Does aitken diameter need the result from the previous kernel? If not, you don't need the fence. Similar for other fences: only add them if you _need_ the previous kernel(s) to complete before the next starts.

components/eamxx/tests/single-process/mam/aci/CMakeLists.txt
```diff
@@ -0,0 +1,57 @@
...
+CompareNCFilesFamilyMpi (
+  TEST_BASE_NAME ${TEST_BASE_NAME}
+  FILE_META_NAME ${TEST_BASE_NAME}_output.INSTANT.nsteps_x1.npMPIRANKS.${RUN_T0}.nc
+  MPI_RANKS ${TEST_RANK_START} ${TEST_RANK_END}
+  LABELS mam4_aci physics
+  META_FIXTURES_REQUIRED ${FIXTURES_BASE_NAME}_npMPIRANKS_omp1
+)
+
+# Check tendency calculation
+foreach (NRANKS RANGE ${TEST_RANK_START} ${TEST_RANK_END})
```
bartgol, line 40: This looks like a copy+paste overlook. Your test does not compute any tendencies, so this test will likely fail. Let's remove it.

bartgol (changes requested):
I have a few comments.

Also, I didn't comment on all of the occurrences, but you should replace TeamThreadRange with TeamVectorRange, and for the lambdas in those loops, use `[&]` instead of `[=]`.

components/eamxx/src/physics/shoc/eamxx_shoc_process_interface.cpp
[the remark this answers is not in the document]
susburrows: The EAGLES team in planning to look into this issue in more detail, but their top priority right now is to simply get an implementation of the original mam4xx code running.  My two cents: I think it's going to be very hard to test whether changes to the treatment of vertical velocity variances are improvements (in terms of model behavior) until the full aerosol life cycle is implemented and longer tests can be performed... so, kicking the can down the road seems like a reasonable approach here to me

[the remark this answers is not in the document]
singhbalwinder: >Looking at mam4xx's dropmixnuc inline docs, I see that the variable cldo (the _prev var here) is "cloud fraction on previous time step". I'm a bit confused then: is it "from the previous timestep" or "whatevet was before shoc"? If the latter, why does ACI needs to know the value right before shoc? Can you explain what's the rationale behind it?

Here is an explanation by @kaizhangpnl about the use of this field:
```
liqcldfo and liqcldf are used to determine whether there is growing or shrinking 
cloud in the grid box. liqcldfo is the cloud fraction before CLUBB updates it - same 
as the old cloud fraction used for cloud microphysics and radiation in the previous 
time step (after CLUBB update).
```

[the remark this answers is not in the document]
singhbalwinder: We will remove this field from the SHOC and compute it in the ACI interface in a follow-up PR.

components/eamxx/tests/single-process/mam/aci/CMakeLists.txt
singhbalwinder: Thanks for catching that. I have removed this section.

components/eamxx/src/physics/mam/eamxx_mam_aci_functions.hpp
singhbalwinder: Done.

components/eamxx/src/physics/mam/eamxx_mam_aci_functions.hpp
singhbalwinder: Thanks. Done

components/eamxx/src/physics/mam/eamxx_mam_aci_functions.hpp
singhbalwinder: Done

components/eamxx/src/physics/mam/eamxx_mam_aci_functions.hpp
singhbalwinder: I implemented it but it didn't work (as we discussed) but the following worked:
```
Kokkos::deep_copy(
      aitken_dry_dia_,
      ekat::subview_1(dgnum_, static_cast<int>(mam4::ModeIndex::Aitken)));
```

components/eamxx/src/physics/mam/eamxx_mam_aci_functions.hpp
```diff
@@ -0,0 +1,726 @@
...
+
+KOKKOS_INLINE_FUNCTION
+void compute_tke_using_w_sec(const haero::ThreadTeam &team,
+                             const MAMAci::const_view_2d w_sec, const int icol,
+                             const int nlev,
+                             // output
+                             MAMAci::view_2d tke) {
+  Kokkos::parallel_for(
+      Kokkos::TeamThreadRange(team, 0u, nlev + 1),
+      KOKKOS_LAMBDA(int kk) { tke(icol, kk) = (3.0 / 2.0) * w_sec(icol, kk); });
```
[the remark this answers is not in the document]
singhbalwinder: We discussed this and (fortunately), it is *not* a bug as it is currently implemented (thanks @mahf708 for looking into it). The `w2tune` is 1.0 in the code currently. If it were anything other than 1, this calculation would have been inconsistent. I will create a git issue, we decided to get `tke` directly from SHOC and remove the code to get `w_variance`.

components/eamxx/cime_config/namelist_defaults_scream.xml
```diff
@@ -234,6 +234,12 @@ be lost if SCREAM_HACK_XML is not enabled.
       <Ckm type="real" doc="Eddy diffusivity coefficient for momentum">0.1</Ckm>
     </shoc>
 
+    <!-- MAM4xx-ACI -->
+    <mam4_aci inherit="atm_proc_base">
+      <wsubmin type="real" doc="Minimum diagnostic sub-grid vertical velocity">0.001</wsubmin>
+      <top_level_mam4xx type="integer" doc="Top level for MAM4xx">6</top_level_mam4xx>
```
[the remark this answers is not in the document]
singhbalwinder: I will include a small code to compute `top_lev` in a follow up PR.

components/eamxx/src/physics/mam/eamxx_mam_aci_process_interface.cpp
singhbalwinder: We will include a code to compute it. It varies with different vertical resolutions.

components/eamxx/src/physics/mam/eamxx_mam_aci_functions.hpp
singhbalwinder: Thanks. We will fix it in a follow-up PR (see issue #2858 )

components/eamxx/src/physics/mam/eamxx_mam_aci_functions.hpp
singhbalwinder: Thanks. We will fix it in a follow-up PR (see issue #2858 )

singhbalwinder:
@bartgol and @mahf708: I have now addressed most of the comments and created git issues for the remaining ones (see #2858 and #2849). I or someone from our team will address those comments soon.

I see only the expected fails (baseline "_cmp") from the AT. This PR is ready to merge if everything looks good to you.

components/eamxx/tests/multi-process/physics_only/mam/shoc_cldfrac_mam4_aci_p3_rrtmgp/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
+INCLUDE (ScreamUtils)
+
+set (TEST_BASE_NAME shoc_cldfrac_mam4_aci_p3_rrtmgp)
+set (FIXTURES_BASE_NAME ${TEST_BASE_NAME}_generate_output_nc_files)
+
+# Create the test
+CreateADUnitTest(${TEST_BASE_NAME}
+  LIBS shoc cld_fraction p3 scream_rrtmgp mam
+  LABELS shoc cld p3 rrtmgp physics mam
```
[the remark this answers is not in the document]
singhbalwinder: Thanks, Oscar for catching that. I have fixed the labels.

AaronDonahue:
@kaizhangpnl @odiazib @tcclevenger @mahf708 
Can you take a look and confirm that the latest version addresses your comments?

@singhbalwinder 
Can you look at the fails on `weaver`?

kaizhangpnl:
> @kaizhangpnl @odiazib @tcclevenger @mahf708 Can you take a look and confirm that the latest version addresses your comments?
> 
> @singhbalwinder Can you look at the fails on `weaver`?

Yes, all my comments have been addressed.

bartgol (dismissed):

singhbalwinder:
@overfelt has fixed the bug (and a lot of compiler warnings) and now we only have expected fails:

```
DIFF SMS_D_Ln5.ne4pg2_oQU480.F2010-SCREAMv1-MPASSI.mappy_gnu.scream-mam4xx-aci (phase BASELINE)

104:mam4_optics_standalone_baseline_cmp
106:mam4_aci_standalone_baseline_cmp
139:homme_shoc_cld_p3_mam_optics_rrtmgp_baseline_cmp
90:mam4_optics_standalone_baseline_cmp
92:mam4_aci_standalone_baseline_cmp
```

`mam4_optics_standalone_baseline_cmp` is failing the baseline test as we have updated the IC file for all the MAM4xx tests to use more realistic inputs from EAMv2.

@AaronDonahue : Could you please help merge this? Thanks!

bartgol (approved):

PR #2796, opened 2024-04-24
move rad calls docs from tech to user, update nudging and aerocom docs

mahf708:
A slight reorg in docs added in earlier PRs, with more concrete and current examples. The rad calls docs was moved from technical to user.


components/eamxx/docs/user/aerocom_cldtop.md
```diff
@@ -0,0 +1,30 @@
+# The AeroCom algorithm
+
```
PeterCaldwell: My vision is for the user guide to explain available options to control a simulation without changing code, the dev guide should provide general info about how to change code, and the tech doc should describe the equations solved and algorithms used for the particular version of EAMxx we have now. 

In this context, I think aerocom_top.md is a tech doc and clean_clear_sky.md is mostly a user guide thing except for describing the addition of a new class which seems like it will be outdated soon (when the new class isn't new) and maybe doesn't need to be described?

bartgol (dismissed):

components/eamxx/docs/technical/aerocom_cldtop.md
```diff
@@ -24,4 +24,7 @@ $$c\Phi_{i,k} = \frac{cQ_{i,k}}{iQ_{i,k} + cQ_{i,k}}$$
...
 
-A helpful references: Räisänen, P., Barker, H. W., Khairoutdinov, M. F., Li, J., & Randall, D. A. (2004). Stochastic generation of subgrid‐scale cloudy columns for large‐scale models. Quarterly Journal of the Royal Meteorological Society: A journal of the atmospheric sciences, applied meteorology and physical oceanography, 130(601), 2047-2067.
+Helpful pointers:
+
+- Räisänen, P., Barker, H. W., Khairoutdinov, M. F., Li, J., & Randall, D. A. (2004). Stochastic generation of subgrid‐scale cloudy columns for large‐scale models. Quarterly Journal of the Royal Meteorological Society: A journal of the atmospheric sciences, applied meteorology and physical oceanography, 130(601), 2047-2067.
+- References in this ECMWF report: Tiedtke, M., Geleyn, J.-F., Hollingsworth, A., and Louis, J.-F.: ECMWF model parameterisation of sub-grid scale processes, Technical Report, ECMWF, Shinfield Park, Reading, 10, 1979.
```
PeterCaldwell, line 30: Note that we're supposed to use bibtex for citations now. See my https://github.com/E3SM-Project/scream/pull/2792/ for the (apparently official) way to do this. But my PR brings in the .bib file, creating a PR order dependency. We could integrate yours as is, then fix as a follow up?

components/eamxx/docs/user/clean_clear_sky.md
```diff
@@ -0,0 +1,83 @@
+# Clean- and clean-clear-sky diagnostics
+
+In order to decompose the aerosol effective radiative forcing, additional diagnostic radiation calls are needed.
+These extra diagnostics are optionally added to the main radiation call. The extra diagnostics are:
+
+- Clean-clear-sky fluxes: the fluxes that would be present if there were neither aerosols nor clouds, and are calculated by adding an additional radiation call at the very beginning of the logic before the optics class is endowed with aerosol and cloud properties.
+- Clean-sky fluxes: the fluxes that would be present if there were no aerosols, and are calculated by adding an additional radiation call after substantiating an additional optics class, but not endowing it with aerosol properties.
+
+It was necessary to add an additional optics class because the original optics class is endowed with aerosols before clouds (in order to calculate the clear-sky fluxes).
```
PeterCaldwell, line 9: the line above is the one I was complaining about in terms of being TMI for a user doc. I have a slight preference for deleting this sentence, but am fine with keeping it if you prefer.

components/eamxx/docs/user/clean_clear_sky.md
```diff
@@ -0,0 +1,83 @@
...
+
+The extra calls are controlled by runtime flags `extra_clnclrsky_diag` and `extra_clnsky_diag` (they take either `true` or `false` as their values).
+
+```shell
+    ./atmchange extra_clnclrsky_diag=true
+    ./atmchange extra_clnsky_diag=true
+```
+
+An example output file with the additional radiation diagnostics requested.
+
```
PeterCaldwell, line 21: I think it would make more sense to include an example that *only* includes clnclrsky variables. As is, the important bits are hidden in a pile of irrelevant stuff and the example will take more screen real estate than it needs to. But again, I'm fine with keeping it as is if you want.

components/eamxx/docs/user/coarse_nudging.md
```diff
@@ -5,21 +5,23 @@ Instead, in EAMxx, it is possible to nudge from coarse data.
 This is done by remapping the coarse data provided by the user to the runtime physics grid of EAMxx.
```
PeterCaldwell, line 5: I'm kind of weirded out by calling the file name "coarse_nudging", which always makes me wonder if it is something different than what I think of as nudging. Like maybe it is something called coarse that is your nudging target. Could you get rid of the coarse mentions except having a sentence saying that right now the only data you can nudge to is at either the same or coarser resolution than the model grid?

components/eamxx/docs/user/coarse_nudging.md
```diff
@@ -5,21 +5,23 @@ Instead, in EAMxx, it is possible to nudge from coarse data.
...
-./atmchange nudging_filenames_patterns=/path/to/nudging_data_ne4pg2_L72.nc
-./atmchange nudging_refine_remap_mapfile=/another/path/to/mapping_file_ne4pg2_to_ne120pg2.nc
+    ./atmchange nudging::nudging_filenames_patterns="${PSCRATCH}/mo_ne45pg2_nudg_trim_merra/*.nc"
+    ./atmchange nudging::source_pressure_type=TIME_DEPENDENT_3D_PROFILE
+    ./atmchange nudging::nudging_fields=U,V
+    ./atmchange nudging::nudging_timescale=21600 #  6-hr
+    ./atmchange nudging::nudging_refine_remap_mapfile="${PSCRATCH}/map_ne45pg2_to_ne256pg2.trbilin.20240410.nc"
 ```
+
+To gain a deeper understanding of these parameters and options, please refer to the overarching nudging documentation.
```
PeterCaldwell, line 27: is there a link to the overarching nudging documentation? And by overarching you mean the nudging info in the EAMf90 user guide, right?

components/eamxx/docs/user/aerocom_cldtop.md
PeterCaldwell: I think we're in agreement now. I'm only complaining about mentioning classes because I think 99% of users won't understand what a class is (and I think you've deleted the class mention now). I totally agree with "I think it is important people know what we are calling "clean" and "clean-clear" at a brief technical level"

components/eamxx/docs/refs.bib
```diff
@@ -123,3 +123,27 @@ @article{Taylor_et20
...
+	language = {en},
+	institution = {ECMWF},
+	author = {Tiedtke, M. and Geleyn, J.-F. and Hollingsworth, A. and Louis, J.-F.},
+	month = jan,
+	year = {1979},
+	note = {10},
+	pages = {146},
+}
+
+@article{raisanen2004stochastic,
```
PeterCaldwell, line 140: I was trying to add references to the .bib in alphabetical order in order to make them easier to find, but in my .bib PR I asked the question "how long is this going to last?". Apparently 1 PR ;-). I'd prefer you put raisanen in alphabetical order, but am resigned to losing alphabetization.

components/eamxx/docs/user/nudging.md
```diff
@@ -0,0 +1,25 @@
+# Nudging in EAMxx
+
+Nudging is supported in EAMxx.
```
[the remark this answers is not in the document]
PeterCaldwell: Some things we might want to add (in this PR or another):
1. to nudge to ERA5 data, we need to be able to dictionary between the EAMxx name and the ERA5 name for each variable, right? Would be worth mentioning how to do that. 
2. the fact that we can only nudge coarser grid data makes me think that regridding is done using an internally calculated bilinear remap rather than by passing the mapping file you want the code to use like we've done everywhere else. Otherwise nudging to finer data would just require passing conservative mapping weights. Sorry for being out of the loop, but why are we handling this case differently?
3. it's fine if Naser doesn't include info on weighted nudging here, but we definitely need it. We also need to include the specific commands needed to do remap nudging.
4. We should mention how data to nudge to can be split across many files. 
5. I kinda like how instead of long descriptions of what each command does, there's just an example with all of the commands you'd need to tinker with to get things working. But I guess we need to supply a bit of extra info like what the units of the nudging timescale are, whether nudging field names need to be eamxx format or reanalysis format, what the source_pressure_type is about, etc.

bartgol (approved):

components/eamxx/docs/user/nudging.md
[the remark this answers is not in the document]
jsbamboo: i'm definitely not good at fully thinking through a wide range of user needs ahead of time.. you guys have very cool suggestions
 
just two questions - i thought the weighted nudging && nudging-to-coarse data simultaneously has been addressed since Luca's PR (the one changed the order of horizontal and vertical mapping)? as long as the weight file is on the model grid (horz+vert)

another thing is Aaron's PR to change the startdate in TimeInterpolation (https://github.com/E3SM-Project/scream/pull/2635) - it's very helpful for my own tests, just curious about its progress

bartgol (approved):

PeterCaldwell:
Oh, I thought I did review it. In any case, I'm happy with it.

components/eamxx/docs/user/nudging.md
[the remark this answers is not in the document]
jsbamboo: oh i see, thanks! i feel this check may can be removed after #2701 (it seems like the interpolation order caused the previous failures since the previous perform_vertical_interpolation needs to be on the same horiz grid) 

> We still have a runtime check to forbid this: https://github.com/E3SM-Project/scream/blob/305bde73db9d3c641e8808caa8371ad399256d9d/components/eamxx/src/physics/nudging/eamxx_nudging_process_interface.cpp#L138C1-L142C60

PR #2797, opened 2024-04-24
test unification of docs

mahf708:


.github/workflows/e3sm-gh-pages.yml
```diff
@@ -15,7 +15,7 @@ concurrency:
   
 jobs:
   Build-and-Deploy-docs:
-    if: ${{ github.event.repository.name == 'e3sm' }}
+    if: ${{ github.event.repository.name == 'e3sm' || github.event.repository.name == 'scream' }}
```
bartgol, line 18: I don't think we should deploy if the repo is 'scream', so we need some checks down at the `GitHub Pages action` step.

bartgol: Ah, yes, the deployment url is taken from the current repo name.

bartgol: > Regardless, this doesn't look like it's working like expected, so let's not waste more time on it...

Uh? It seemed to have done what you wanted.

PR #2799, opened 2024-04-25
For machine `frontier-scream-gpu`, upgrade to rocm 5.4

ndkeen:
Only for machine `frontier-scream-gpu`, upgrade rocm from 5.1 to 5.4.

Note that other similar machines (frontier and crusher) already have rocm 5.4.
We could also upgrade `crusher-scream-gpu` to 5.4 as well.

The reason for upgrade is that we found it's needed to complete the kokkos version upgrade to 4.2 (`bartgol/eamxx/kokkos-4.2`)
So I wanted to make this change first as a clean PR.

With at least a simple ne30 test, results are not BFB.
[NBFB]

### DIFF
--- a/cime_config/machines/config_machines.xml
+++ b/cime_config/machines/config_machines.xml
@@ -1433,7 +1433,7 @@
         <command name="reset"></command>
         <command name="load">PrgEnv-cray</command>
         <command name="load">craype-accel-amd-gfx90a</command>
-        <command name="load">rocm/5.1.0</command>
+        <command name="load">rocm/5.4.0</command>
         <command name="load">libunwind/1.6.2</command>
       </modules>
       <modules>


bartgol (approved):

mt5555:
Regarding the crusher-scream-gpu I think "crusher" went away with the end of the ECP project - it became an internal OLCF machine and it was unclear what it's future would be.  

Regarding rocm5.4 - in our SC2023 paper, there were some issues with 5.4 (it was slower than 5.1, unless other code changes were made).  Not sure what the status is now, but @sarats might remember the details.

mt5555:
from 1 year ago:  "The original code demonstrates a significant performance degradation moving from ROCm 5.1.0 to ROCm 5.4.0, and the “caar compute” region accounts for the bulk of this difference. The optimizations not only decrease the “caar compute” runtime significantly; but also improves the robustness by minimizing performance variability associated with ROCm versions. "

The optimizations referred to above are the ones @trey-ornl is working on to get into SCREAM - not sure the status on that.

sarats:
Last year:
Without Trey's optimizations, Scream had a performance degradation when we switched from 5.1 to to 5.4. 
Additionally, there were other MPI flags that further degraded performance and have since been removed.

Anyway, we can't cling to old rocm versions, so this needs to be updated.

ndkeen:
Mark also reminded us of this. I tried a ne256 case which showed 5% decrease in perf using rocm 5.4. But a ne1024 case on 2048 nodes did not show any difference. 

I vote we move to rocm 5.4 to allow the newer kokkos version branch to be merged. Then work towards upgrading all module versions on frontier.

grnydawn:
@ndkeen, I have tested this PR with the "PEM_Ln90.ne30pg2_ne30pg2.F2010-SCREAMv1" case, and it passed. From the comments above, I think this PR is ready to merge.

@rljacob, I cannot assign myself to this PR on GitHub, possibly because my Github id not included in the admin group. Could you add me so that I can proceed with the merging?

PeterCaldwell:
I think @grnydawn can't merge because he's not a privileged member of the EAMxx team and this is the SCREAM rather than E3SM repo. Since Youngsung doesn't  know EAMxx's CI and integration policies, we should get someone else to do the integration (once others agree it is time).

PR #2800, opened 2024-04-25
add "NumberPath" diags

mahf708:
Like WaterPath, but scaling by mixing ratios (aka "numbers"?!)... An example:

$$
\mathcal{L} = \sum q\frac{\rho}{g}
$$

$$
\mathcal{N} = \sum q\frac{\rho}{g}n
$$

What I did:
1. I copied WaterPath diags and edited to suit the above
2. I trimmed WaterPath tests to focus on one test only

### DIFF
--- a/components/eamxx/src/diagnostics/tests/number_paths_tests.cpp
+++ b/components/eamxx/src/diagnostics/tests/number_paths_tests.cpp
@@ -1,0 +1,243 @@
+#include <iomanip>
+
+#include "catch2/catch.hpp"
+#include "diagnostics/register_diagnostics.hpp"
+#include "ekat/kokkos/ekat_kokkos_utils.hpp"
+#include "ekat/util/ekat_test_utils.hpp"
+#include "physics/share/physics_constants.hpp"
+#include "share/field/field_utils.hpp"
+#include "share/grid/mesh_free_grids_manager.hpp"
+#include "share/util/scream_common_physics_functions.hpp"
+#include "share/util/scream_setup_random_test.hpp"
+#include "share/util/scream_utils.hpp"
+
+namespace scream {
+
+std::shared_ptr<GridsManager> create_gm(const ekat::Comm &comm, const int ncols,
+                                        const int nlevs) {
+  const int num_global_cols = ncols * comm.size();
+
+  using vos_t = std::vector<std::string>;
+  ekat::ParameterList gm_params;
+  gm_params.set("grids_names", vos_t{"Point Grid"});
+  auto &pl = gm_params.sublist("Point Grid");
+  pl.set<std::string>("type", "point_grid");
+  pl.set("aliases", vos_t{"Physics"});
+  pl.set<int>("number_of_global_columns", num_global_cols);
+  pl.set<int>("number_of_vertical_levels", nlevs);
+
+  auto gm = create_mesh_free_grids_manager(comm, gm_params);
+  gm->build_grids();
+
+  return gm;
+}
+
+//-----------------------------------------------------------------------------------------------//
+template <typename DeviceT>
+void run(std::mt19937_64 &engine) {
+  using PC         = scream::physics::Constants<Real>;
+  using KT         = ekat::KokkosTypes<DeviceT>;
+  using ESU        = ekat::ExeSpaceUtils<typename KT::ExeSpace>;
+  using MemberType = typename KT::MemberType;
+  using view_1d    = typename KT::template view_1d<Real>;
+
+  constexpr int num_levs = 33;
+  constexpr Real gravit  = PC::gravit;
+  constexpr Real macheps = PC::macheps;
+
+  // A world comm
+  ekat::Comm comm(MPI_COMM_WORLD);
+
+  // Create a grids manager - single column for these tests
+  const int ncols = 10;
+  auto gm         = create_gm(comm, ncols, num_levs);
+
+  // Kokkos Policy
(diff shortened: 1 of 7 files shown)


components/eamxx/src/diagnostics/tests/number_paths_tests.cpp
```diff
@@ -0,0 +1,243 @@
...
+    const auto &pseudo_dens_v = pseudo_dens_f.get_view<Real **>();
+    for(int icol = 0; icol < ncols; icol++) {
+      const auto &qc_sub = ekat::subview(qc_v, icol);
+      const auto &nc_sub = ekat::subview(nc_v, icol);
+      const auto &qi_sub = ekat::subview(qi_v, icol);
+      const auto &ni_sub = ekat::subview(ni_v, icol);
+      const auto &qr_sub = ekat::subview(qr_v, icol);
+      const auto &nr_sub = ekat::subview(nr_v, icol);
+      const auto &dp_sub = ekat::subview(pseudo_dens_v, icol);
+      ekat::genRandArray(pseudo_density, engine, pdf_pseudodens);
```
bartgol, line 143: Why not calling `genRandArray` directly on `xyz_sub`? Seems less lines of code, and easier to maintain/follow.

components/eamxx/src/diagnostics/tests/number_paths_tests.cpp
```diff
@@ -0,0 +1,243 @@
...
+      Kokkos::deep_copy(nr_sub, nr);
+    }
+    // Grab views for each of the number path diagnostics
+    const auto &lnp   = diags["lnp"]->get_diagnostic();
+    const auto &inp   = diags["inp"]->get_diagnostic();
+    const auto &rnp   = diags["rnp"]->get_diagnostic();
+    const auto &lnp_v = lnp.get_view<const Real *>();
+    const auto &inp_v = inp.get_view<const Real *>();
+    const auto &rnp_v = rnp.get_view<const Real *>();
+    const auto &lnp_h = lnp.get_view<const Real *, Host>();
```
bartgol, line 166: These three host views are unused.

components/eamxx/src/diagnostics/tests/number_paths_tests.cpp
```diff
@@ -0,0 +1,243 @@
...
+            Real qndc_prod = 0.0;
+            auto qc_icol   = ekat::subview(qc_v, icol);
+            auto nc_icol   = ekat::subview(nc_v, icol);
+            Kokkos::parallel_reduce(
+                Kokkos::TeamVectorRange(team, num_levs),
+                [&](Int idx, Real &lsum) {
+                  lsum += nc_icol(idx) * qc_icol(idx) * dp_icol(idx) / gravit;
+                },
+                qndc_prod);
+            REQUIRE(std::abs(lnp_v(icol) - qndc_prod) < macheps);
```
bartgol, line 189: REQUIRE is a Catch2 macro, that does not work on device. You have two options:
- turn `parallel_for` into `parallel_reduce`, where you reduce a counter of how many failed comparisons you get.
- just do a simple for loop without kokkos parallel stuff (and compare against host views from the diags).

Option two would be my favorite, but you may run into issues with order of ops in the reduction.

bartgol (changes requested):
Looks good. I only have some minor comments. Feel free to disregard them. The only one that I think needs to be addressed is the use of REQUIRE in a kernel.

components/eamxx/src/diagnostics/tests/number_paths_tests.cpp
```diff
@@ -0,0 +1,216 @@
...
+      const auto &nr_sub = ekat::subview(nr_v, icol);
+      const auto &dp_sub = ekat::subview(pd_v, icol);
+      ekat::genRandArray(qc_sub, engine, pdf_qx);
+      ekat::genRandArray(nc_sub, engine, pdf_qx);
+      ekat::genRandArray(qi_sub, engine, pdf_qx);
+      ekat::genRandArray(ni_sub, engine, pdf_qx);
+      ekat::genRandArray(qr_sub, engine, pdf_qx);
+      ekat::genRandArray(nr_sub, engine, pdf_qx);
+      ekat::genRandArray(dp_sub, engine, pdf_pd);
+    }
```
bartgol, line 157: The random data is generated on the device view only. If you want to use `qr_h`, ... in the checks below, you need to call `sync_to_host()` on the fields.

components/eamxx/src/diagnostics/tests/number_paths_tests.cpp
```diff
@@ -0,0 +1,216 @@
...
+    const auto &inp   = diags["inp"]->get_diagnostic();
+    const auto &rnp   = diags["rnp"]->get_diagnostic();
+    const auto &lnp_h = lnp.get_view<const Real *, Host>();
+    const auto &inp_h = inp.get_view<const Real *, Host>();
+    const auto &rnp_h = rnp.get_view<const Real *, Host>();
+
+    for(const auto &dd : diags) {
+      dd.second->compute_diagnostic();
+    }
+    // test manual calculation vs one provided by diags
```
bartgol, line 169: Here too: before comparing the diags with the manually computed ones, you need to sync `inp`, `rnp`, `lnp` to host first.

bartgol (approved):
I think you are missing a few syncs.

bartgol (changes requested):
Whoops, clicked on the wrong button.

bartgol (approved):

PR #2801, opened 2024-04-25
Aarondonahue/io add longname metadata

AaronDonahue:
Add an option to specify a 'longname' for output variables.
Using the new longname feature we add the variable 'lev' to the grid data so that it can be output.
    
This commit adds a structure to io_utils that allows the developer to specify a longname to go into the metadata of an output variable.  Building on the longname capability this commit adds a new grid variable 'lev' which is a representative pressure at each level built from hyam and hybm.  It is necessary for some post-processing.

### DIFF
--- a/components/eamxx/src/dynamics/homme/homme_grids_manager.cpp
+++ b/components/eamxx/src/dynamics/homme/homme_grids_manager.cpp
@@ -258,17 +258,31 @@ build_physics_grid (const ci_string& type, const ci_string& rebalance) {
     auto layout_mid = phys_grid->get_vertical_layout(true);
     auto layout_int = phys_grid->get_vertical_layout(false);
     const auto nondim = ekat::units::Units::nondimensional();
+    auto lev_unit = ekat::units::Units::nondimensional();;
+    lev_unit.set_string("mb");
 
     auto hyai = phys_grid->create_geometry_data("hyai",layout_int,nondim);
     auto hybi = phys_grid->create_geometry_data("hybi",layout_int,nondim);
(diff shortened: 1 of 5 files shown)


AaronDonahue:
Note, this PR is only non-BFB because we add a new default variable to all output.  Otherwise it is BFB.

bartgol (dismissed):

AaronDonahue:
This will probably conflict with #2802 , if this goes in first we might need to update for #2802 , or vice versa if that one goes in first.

bartgol (approved):

kaizhangpnl:
@AaronDonahue Thanks for this useful PR. The code compiles and runs well on Chrysalis, but the "lev" variable is still missing in my simulation output (with monthly mean and daily mean 3D fields). Is it in your test simulations? Thanks! 

My code, case, and run directories are 

```
/lcrc/group/e3sm/ac.kzhang/share/scream_lev 
/lcrc/group/e3sm/ac.kzhang/share/case_scripts
/lcrc/group/e3sm/ac.kzhang/share/run 
```

bartgol (approved):

kaizhangpnl (approved):
Tested on Chrysalis with the following setup: 

```
readonly COMPSET="F2010-SCREAMv1"
readonly RESOLUTION="ne30_ne30" 
readonly PELAYOUT="320x1"
readonly WALLTIME="00:59:00"
```

"lev" variable is now in the output: 

```
float lev(lev) ;
        lev:units = “mb” ;
        lev:long_name = “hybrid level at midpoints (1000*(A+B))” ;
        lev:_FillValue = 3.402824e+33f ;
```

PR #2802, opened 2024-04-26
Use only Scorpio's C interfaces in EAMxx

bartgol:
This PR removes the F90 interfaces to scorpio. A summary of changes:

- some interfaces name changes. E.g., remove `eam_pio_` prefixes (since we are already in scream::scorpio namespace), `read_var`/`write_var` rather than `grid_read_data_array`/`grid_write_data_array`.
- read/write of vars support read/write of time-dep vars that are not partitioned (should address #2601)
- users must provide decompositions only along the partitioned dimension. The IO interface will _internally_ build PIO decomp objects for all variables. This simplifies life a lot, especially in the case of non-partitioned vars, where we no longer need to provide the list of dofs offsets.
- allow handling input files where the "time" dimension is not unlimited. We do this by "pretending" that a dim called "time" is unlimited (which triggers a different handling of the var layout). I am not fond of this, but as far as I understand, we want to support this type of files, correct? If that's not true, I'm more than happy to revert this, and update all existing nc files for unit tests on the servers. But I suspect we can't guarantee all future input files will have unlimted time dim...

@tcclevenger @AaronDonahue I'm marking the PR WIP, since I want to verify everything works also on our production machines. But I want to start the review process, in case some design choices are to be revisited.


AaronDonahue:
This is a general question, since we are touching all of the scream_scorpio code, is this an opportunity for us to switch all of the uses of 'scream' for 'eamxx'?  If it's too much trouble no worries, it isn't essential.  But this PR does offer an opportunity to drop "scream" when we mean "eamxx".

components/eamxx/src/share/io/scorpio_input.cpp
```diff
@@ -351,51 +351,42 @@ void AtmosphereInput::finalize()
...
-
-/* ---------------------------------------------------------- */
-void AtmosphereInput::register_variables()
-{
-  // Register each variable in IO stream with the SCORPIO interface.
-  // This allows SCORPIO to lookup vars in the nc file with the correct
-  // dof decomposition across different ranks.
+  // Some input files have the "time" dimension as non-unlimited. This messes up our
+  // scorpio interface. To avoid trouble, if a dim called 'time' is present we
+  // treat is as unlimited, even though it isn't.
```
AaronDonahue, line 360: typo: "is" -> "it"

components/eamxx/src/share/io/scream_output_manager.cpp
```diff
@@ -801,20 +797,24 @@ void OutputManager::set_file_header(const IOFileSpecs& file_specs)
...
+  set_str_att("case",p.get<std::string>("caseid","NONE"));
+  set_str_att("source","E3SM Atmosphere Model (EAMxx)");
+  set_str_att("eamxx_version",EAMXX_VERSION);
+  set_str_att("git_version",p.get<std::string>("git_version",EAMXX_GIT_VERSION));
+  set_str_att("hostname",p.get<std::string>("hostname","UNKNOWN"));
+  set_str_att("username",p.get<std::string>("username","UNKNOWN"));
+  set_str_att("atm_initial_conditions_file",p.get<std::string>("initial_conditions_file","NONE"));
+  set_str_att("topography_file",p.get<std::string>("topography_file","NONE"));
+  set_str_att("contact","e3sm-data-support@llnl.gov");
+  set_str_att("institution_id","E3SM-Projet");
```
AaronDonahue, line 813: Leftover typo from before: "project"

components/eamxx/src/share/io/scream_scorpio_types.hpp
```diff
@@ -0,0 +1,154 @@
...
+  std::string name;   // In EAMxx, we prefer to use a string name than a number id
+};
+
+// An entity that is linked to a specific file
+// This allows an entity to retrieve its file
+struct PIOFileEntity : public PIOEntity {
+  int fid = -1;       // Allow retrieving file from the entity
+};
+
+// struct PIOAtt : public PIOFileEntity {
```
AaronDonahue, line 62: any reason this is commented out?

components/eamxx/src/share/io/scream_scorpio_types.hpp
```diff
@@ -0,0 +1,154 @@
...
+//       A PIODecomp is associated with a Nd layout, which includes decomp_dim
+//       among its dimensions. PIODecomp::offsets are the offsets of the full
+//       array layout owned by this rank. Hence, there can be many PIODecomp
+//       all storing the same dim
+struct PIODecomp : public PIOEntity {
+  std::vector<offset_t>           offsets;  // Owned offsets
+  std::shared_ptr<const PIODim>   dim; 
+};
+
+// // A decomposition of a single dimension
```
AaronDonahue, line 91: Any reason the next 10 lines are commented out?

components/eamxx/src/share/io/scream_scorpio_types.hpp
```diff
@@ -0,0 +1,154 @@
...
+//   std::shared_ptr<const PIODimDecomp>   dim_decomp; // The decomposition along the dimension
+//   std::vector<offset_t>                 offsets;    // Owned offsets along the full var layout
+// };
+
+// A variable
+struct PIOVar : public PIOFileEntity {
+  // Note: if time_dep=true, we will add it to the list of dims passed
+  // to scorpio, but the time dim will not appear in this list.
+  std::vector<std::shared_ptr<const PIODim>>          dims;
+  // std::map<std::string,std::shared_ptr<const PIOAtt>> atts;
```
AaronDonahue, line 108: commented out, not used?

components/eamxx/src/share/io/scream_scorpio_types.hpp
```diff
@@ -0,0 +1,154 @@
...
+  // Used only if a) var is not decomposed, and b) dtype!=nc_dtype
+  int size = -1; // Product of all dims
+  std::vector<char> buf;
+};
+
+// A file, which is basically a container for dims and vars
+struct PIOFile : public PIOEntity {
+  std::map<std::string,std::shared_ptr<PIODim>>   dims;
+  std::map<std::string,std::shared_ptr<PIOVar>>   vars;
+  // std::map<std::string,std::shared_ptr<PIOAtt>>   atts;
```
AaronDonahue, line 139: here too

AaronDonahue (commented):
Looks good to me! Epic PR that I'm quite happy is finally going in.  Thanks for your work on this @bartgol

components/eamxx/src/share/io/scream_scorpio_types.hpp
bartgol: Yeah, I was thinking of adding a PIOAtt ministruct, to allow to easy query if a certain att existed (much like for PIOVar/PIODim). But I realized that for atts there's really not much metadata, so it's prob not worth it. I will delete these lines when I'm ready to clean it up.

components/eamxx/src/share/io/scream_scorpio_types.hpp
bartgol: Yeah, these were early iterations, which I forgot to clean up.. will clean it up once everything passes.

tcclevenger (commented):
This works well with DP, should fix any issues with IOP files where we don't have to make any changes.

components/eamxx/src/control/intensive_observation_period.cpp
```diff
@@ -291,11 +234,14 @@ initialize_iop_file(const util::TimeStamp& run_t0,
...
-  else if (scorpio::has_variable(iop_file, "nbdate"))   bdate_name = "nbdate";
+  if      (scorpio::has_var(iop_file, "bdate"))    bdate_name = "bdate";
+  else if (scorpio::has_var(iop_file, "basedate")) bdate_name = "basedate";
+  else if (scorpio::has_var(iop_file, "nbdate"))   bdate_name = "nbdate";
   else EKAT_ERROR_MSG("Error! No valid name for bdate in "+iop_file+".\n");
-  read_dimensionless_variable_from_file<int>(iop_file, bdate_name, &bdate);
+
+  // Just in case bdate_name is stored with a type that is not "int" (e.g., it's int64)
+  scorpio::change_var_dtype(iop_file,bdate_name,"int");
+  scorpio::read_var(iop_file, bdate_name, &bdate);
```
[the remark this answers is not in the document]
bartgol: `read_data()` already knows the data type (it's the last arg). Originally, I had `change_var_dtype` called inside `read_var` automatically if the input ptr type did not match the dtype from the nc file. However, for decomposed vars, changing dtype requires to reset the PIO decomposition (since the data type is part of it). Therefore, I decided to force the user to set the var dtype manually.

That said, if you call `change_var_dtype` N times with the same dtype, only the first call does anything, all other calls are no-ops. So maybe I can just call `change_var_dtype` from `read_var` after all...

components/eamxx/src/share/io/scream_scorpio_interface.hpp
```diff
@@ -1,168 +1,234 @@
...
+//  float  -> float
+//  double -> double
+//  real   -> float or double (depending on SCREAM_DOUBLE_PRECISION)
+std::string refine_dtype (const std::string& dtype);
+
+template<typename T>
+std::string get_dtype () {
+  using raw_t = typename std::remove_cv<T>::type;
+  std::string s;
+  if (std::is_same<raw_t,int>::value) {
```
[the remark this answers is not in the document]
bartgol: Yes, I thought about using that. You're right, we should use better constructs where possible.

components/eamxx/src/control/intensive_observation_period.cpp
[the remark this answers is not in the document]
bartgol: I don't think we'll ever alternate calls to `read_var<T>("file","var",ptr)`, with different data types T, so changing var dtype on the fly is fine.

tcclevenger (approved):

bartgol:
I can't reproduce the failure on mappy for the sp build. Let's see if they show up again in the AT...

AaronDonahue:
WIP pending performance tests on frontier

bartgol:
@ndkeen tested this branch against master on frontier, using rocm 5.4, and the IO timings appear to be substantially the same:

branch:
```
"a:EAMxx::IO::restart"                                           35.353     24.245 
"a:EAMxx::IO::restart::get_new_file"                             18.749     18.677 
"a:EAMxx::IO::restart::run_output_streams"                        2.174      2.172 
"a:EAMxx::IO::standard"                                         191.585    121.289 
"a:EAMxx::IO::standard::get_new_file"                            19.708      3.860 
"a:EAMxx::IO::standard::run_output_streams"                     174.152    116.068 
"a:EAMxx::IO::vert_remap"                                        20.134     15.586 
"a:EAMxx::IO::horiz_remap"                                       85.079     22.506 
```
master:
```
"a:EAMxx::IO::restart"                                           35.886     24.299 
"a:EAMxx::IO::restart::get_new_file"                             18.922     18.850 
"a:EAMxx::IO::restart::run_output_streams"                        2.096      2.094 
"a:EAMxx::IO::standard"                                         193.661    122.588 
"a:EAMxx::IO::standard::get_new_file"                            20.949      4.156 
"a:EAMxx::IO::standard::run_output_streams"                     174.976    116.125 
"a:EAMxx::IO::vert_remap"                                        20.243     15.651 
"a:EAMxx::IO::horiz_remap"                                       85.374     22.604 
```

@AaronDonahue unless you have any objection, I may merge this today, so it gets tested in the nightlies.

AaronDonahue (approved):

tcclevenger (approved):

PR #2803, opened 2024-04-29
Bring various DPxx options out to the namelist

bogensch:
Brings various DPxx related options out the namelist to allow modification at runtime.


tcclevenger (approved):
I removed automerge in case you wanted to make more changes before requesting reviews. But this looks good to me, I should have added them on my last PR. We could also make case-specific options, but that will depend on how we want to set up the compsets.

PR #2804, opened 2024-04-30
AeroComCldTop/Bot diagnostics

mahf708:
Several TODO items remain:
- [x] add test for cldtop (not sure if I have enough bandwidth to add a test for cldbot anytime soon, so that one would be "use at your own" risk type of deal)
- [x] come up with a smarter strategy for self-documenting via attrs
- [x] ... other stuff?

### DIFF
--- a/components/eamxx/src/diagnostics/aerocom_cld.cpp
+++ b/components/eamxx/src/diagnostics/aerocom_cld.cpp
@@ -45,6 +45,10 @@ void AeroComCld::set_grids(
   EKAT_REQUIRE_MSG(m_index_map.size() == m_units_map.size(),
                    "Error! Some inconsistency in AeroComCld: index and units "
                    "maps do not match!\n");
+  // Ensure m_index_map and m_ndiag match
+  EKAT_REQUIRE_MSG(static_cast<int>(m_index_map.size()) == m_ndiag,
+                   "Error! Some inconsistency in AeroComCld: index and units "
+                   "maps do not match!\n");
 
   m_ncols = grid->get_num_local_dofs();
   m_nlevs = grid->get_num_vertical_levels();
@@ -133,17 +137,17 @@ void AeroComCld::compute_diagnostic_impl() {
   const int nlevs = m_nlevs;
   bool is_top     = (m_topbot == "Top");
 
-  // Get indices of each call we care about
-  const int T_mid_idx         = m_index_map["T_mid"];
-  const int p_mid_idx         = m_index_map["p_mid"];
-  const int cldfrac_ice_idx   = m_index_map["cldfrac_ice"];
-  const int cldfrac_liq_idx   = m_index_map["cldfrac_liq"];
-  const int cdnc_idx          = m_index_map["cdnc"];
-  const int eff_radius_qc_idx = m_index_map["eff_radius_qc"];
-  const int eff_radius_qi_idx = m_index_map["eff_radius_qi"];
-  const int nc_idx            = m_index_map["nc"];
-  const int ni_idx            = m_index_map["ni"];
-  const int cldfrac_tot_idx   = m_index_map["cldfrac_tot"];
+  // subview the out field for each variable
+  auto o_tmid          = ekat::subview_1(out, m_index_map["T_mid"]);
+  auto o_pmid          = ekat::subview_1(out, m_index_map["p_mid"]);
+  auto o_cldfrac_ice   = ekat::subview_1(out, m_index_map["cldfrac_ice"]);
+  auto o_cldfrac_liq   = ekat::subview_1(out, m_index_map["cldfrac_liq"]);
+  auto o_cdnc          = ekat::subview_1(out, m_index_map["cdnc"]);
+  auto o_eff_radius_qc = ekat::subview_1(out, m_index_map["eff_radius_qc"]);
+  auto o_eff_radius_qi = ekat::subview_1(out, m_index_map["eff_radius_qi"]);
+  auto o_nc            = ekat::subview_1(out, m_index_map["nc"]);
+  auto o_ni            = ekat::subview_1(out, m_index_map["ni"]);
+  auto o_cldfrac_tot   = ekat::subview_1(out, m_index_map["cldfrac_tot"]);
 
   Kokkos::parallel_for(
       "Compute " + name(), policy, KOKKOS_LAMBDA(const MT &team) {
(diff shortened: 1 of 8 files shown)


components/eamxx/src/diagnostics/tests/aerocom_cld_test.cpp
```diff
@@ -0,0 +1,115 @@
...
+    diag->set_required_field(nc);
+    diag->initialize(t0, RunType::Initial);
+
+    // Run diag
+    diag->compute_diagnostic();
+
+    // Check result
+    qc.sync_to_host();
+    nc.sync_to_host();
+    diag->get_diagnostic().sync_to_host();
```
bartgol, line 90: I don't think you need this sync, since you never access the diag values on host here.

components/eamxx/src/diagnostics/tests/aerocom_cld_test.cpp
```diff
@@ -0,0 +1,115 @@
...
+    out_tf.deep_copy<double, Host>(0.0);
+    auto out_t = out_tf.get_view<Real **, Host>();
+
+    for(int icol = 0; icol < grid->get_num_local_dofs(); ++icol) {
+      for(int ilev = 0; ilev < nlevs; ++ilev) {
+        out_t(icol, 0) += qc_h(icol, ilev);
+        out_t(icol, 1) += nc_h(icol, ilev);
+      }
+    }
+    out_hf.sync_to_dev();
```
bartgol, line 106: This is also not needed, since you did not modify the host view

components/eamxx/src/diagnostics/aerocom_cld.cpp
```diff
@@ -0,0 +1,208 @@
...
+
+        // Initialize the 1D "clear fraction" as 1 (totally clear)
+        auto clr_icol = 1.0;
+
+        // Loop over all layers in serial (due to accumulative
+        // product), starting at 2 (second highest) layer because the
+        // highest is assumed to hav no clouds for cldtop, but starting
+        // at m_nlevs-1 for cldbot
+
+        for(int ilay : level_vector) {
```
tcclevenger, line 154: If I am reading the code correctly (various column sums), we could do all this in a `Kokkos::parallel_reduce()` instead of serially. It is a bit more complex, but could be a good kokkos learning experience (if you are interested).

components/eamxx/src/diagnostics/aerocom_cld.cpp
```diff
@@ -0,0 +1,208 @@
...
+
+  std::vector<int> level_vector(m_nlevs);
+  for(int i = 1; i < m_nlevs; ++i) {
+    level_vector[i] = m_topbot == "Top" ? i : m_nlevs - i;
+  }
+
+  const auto out = m_diagnostic_output.get_view<Real **>();
+
+  // zero out the outputs
+  Kokkos::deep_copy(out, 0.0);
```
tcclevenger, line 98: Just a personal preference, but I would move lines 90-98 directly before the `Kokkos::parallel_for()` since that is the only place that uses these.

components/eamxx/src/diagnostics/aerocom_cld.cpp
tcclevenger: I see, I was missing the product in there. Let me take a closer look.

tcclevenger: Yeah, there isn't really an easy solution I see. No "prefix product" parallel_scan exists in Kokkos. This is fine.

components/eamxx/src/diagnostics/aerocom_cld.cpp
```diff
@@ -0,0 +1,208 @@
...
+  const auto qv   = get_field_in("qv").get_view<const Real **>();
+  const auto qc   = get_field_in("qc").get_view<const Real **>();
+  const auto qi   = get_field_in("qi").get_view<const Real **>();
+  const auto rel  = get_field_in("eff_radius_qc").get_view<const Real **>();
+  const auto rei  = get_field_in("eff_radius_qi").get_view<const Real **>();
+  const auto cld  = get_field_in("cldfrac_tot").get_view<const Real **>();
+  const auto nc   = get_field_in("nc").get_view<const Real **>();
+
+  // Hack: phony field to store dz
+  auto dz_f = get_field_in("pseudo_density").clone("dz");
```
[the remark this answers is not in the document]
bartgol: Yes. That snippet is fine. The issue is just to not allocate memory at every call. Pre-allocating the `dz_f` field during init should be enough.

components/eamxx/src/diagnostics/tests/aerocom_cld_test.cpp
```diff
@@ -0,0 +1,288 @@
...
+    cd.deep_copy(0.0);
+    diag->compute_diagnostic();
+    diag->get_diagnostic().sync_to_host();
+    Field diag_f = diag->get_diagnostic();
+    auto diag_v  = diag_f.get_view<Real **, Host>();
+    for(int icol = 0; icol < grid->get_num_local_dofs(); ++icol) {
+      for(int idiag = 0; idiag < m_ndiag; ++idiag) {
+        REQUIRE(diag_v(icol, idiag) == 0.0);
+      }
+    }
```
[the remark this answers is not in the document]
tcclevenger: Which line is the segfault? The REQUIRE()?

components/eamxx/src/diagnostics/tests/aerocom_cld_test.cpp
```diff
@@ -0,0 +1,303 @@
...
+    diag->set_required_field(pm);
+    diag->set_required_field(qv);
+    diag->set_required_field(qc);
+    diag->set_required_field(qi);
+    diag->set_required_field(ec);
+    diag->set_required_field(ei);
+    diag->set_required_field(cd);
+    diag->set_required_field(nc);
+
+    // tm.sync_to_dev();
```
bartgol, line 147: When you're ready to merge, let's also rm these commented lines.

tcclevenger: `randomize()` above syncs to device already.

bartgol: They aren't. The `randomize` function randomizes the host copy of the field, and then call `sync_to_host()`. So the fields are already synced.

components/eamxx/src/diagnostics/aerocom_cld.cpp
```diff
@@ -0,0 +1,214 @@
...
+        auto qv_icol   = ekat::subview(qv, icol);
+        auto qc_icol   = ekat::subview(qc, icol);
+        auto qi_icol   = ekat::subview(qi, icol);
+        auto rel_icol  = ekat::subview(rel, icol);
+        auto rei_icol  = ekat::subview(rei, icol);
+        auto cld_icol  = ekat::subview(cld, icol);
+        auto nc_icol   = ekat::subview(nc, icol);
+
+        // We need dz too
+        // TODO: deduce dz_icol type, etc. on the spot here w/o field cloning?
```
tcclevenger, line 142: This comment is outdated?

components/eamxx/src/diagnostics/tests/aerocom_cld_test.cpp
tcclevenger: I think he is accessing on host below. I find this ordering a little confusing. I would do
```
diag_f = diag->get_diagnostic();
diag_f.sync_to_host();
diag_v = diag_f.get_view<Real **, Host>();
```

bartgol: I agree that ordering like you suggested is better. But I also think the deep copy is not needed. The only use of the diag is via `out_hf`, which is only used in `views_are_equal`.  That routine will do a deep copy to host before starting to compare host views.

bartgol: Ah crap. I clicked on the file name above, and gh only showed me changes from one commit. Seems like a silly behavior... I'll review the whole thing.

components/eamxx/src/diagnostics/aerocom_cld.cpp
bartgol: I'm not sure what you need to deduce. The comment says you want to deduce the type of dz_icol, but why? You know it's a 1d view, and that's all you need, right?

As for temp views: inside kernel, you can't declare managed views, you can only create unmanaged views of existing memory (like subviews). Outside a kernel, you can declare any view you like, though be aware that, again, allocating views at runtime is not very efficient; you should create any scratch view at init time (unless they are 'scratch' in the kokkos sense, meaning they use the `ScratchMemorySpace`, which is a per-kernel scratch memory; but we don't really use that in EAMxx (yet)).

bartgol: I mean, doesn't have to be a `Field` obj, you really only need a view. You could use an `ekat::WorkspaceManager`, to be a tad more efficient, if you wanted.

bartgol (dismissed):

bartgol (approved):

bartgol (dismissed):

tcclevenger (approved):

components/eamxx/src/diagnostics/aerocom_cld.cpp
```diff
@@ -45,6 +45,10 @@ void AeroComCld::set_grids(
   EKAT_REQUIRE_MSG(m_index_map.size() == m_units_map.size(),
                    "Error! Some inconsistency in AeroComCld: index and units "
                    "maps do not match!\n");
+  // Ensure m_index_map and m_ndiag match
+  EKAT_REQUIRE_MSG(static_cast<int>(m_index_map.size()) == m_ndiag,
```
tcclevenger, line 49: Why static_cast here? You could `(int)m_index_map.size()` instead.

tcclevenger: Never mind, looking into it more, `static_cast` is probably the better option.

tcclevenger (approved):

components/eamxx/src/diagnostics/aerocom_cld.cpp
tcclevenger: Yeah, static_cast will ensure the types are compatible, whereas I don't think `(int)` has any sort of checks. It shouldn't matter here, but `static_cast` is safer.

PR #2806, opened 2024-05-01
A recent change to get_X_layout didn't make it to ML

AaronDonahue:


AaronDonahue:
Skipping testing because mlcorrection is only able to be run on Livermore machines currently, so this change won't be seen on our AT machines.

bartgol (approved):

PR #2808, opened 2024-05-02
add a banner to scream

mahf708:
add a basic banner to scream to indicate level of support.

### DIFF
--- a/components/eamxx/docs/overrides/main.html
+++ b/components/eamxx/docs/overrides/main.html
@@ -1,0 +1,8 @@
+{% extends "base.html" %}
+
+{% block announce %}
+You're not viewing the latest version.
+<a href="{{ docs.e3sm.org }}"> 
+  <strong>Go to docs.e3sm.org instead!</strong>
+</a>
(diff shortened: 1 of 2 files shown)


components/eamxx/docs/overrides/main.html
```diff
@@ -0,0 +1,8 @@
+{% extends "base.html" %}
+
+{% block announce %}
+EAMxx SCREAM is not supported yet. Read more about this here.
```
bartgol, line 4: I would not add "SCREAM" here. SCREAM is more of a resolution/compset combo, while the model itself is EAMxx.

bartgol: Also, what do you mean with "Read more about this here"?

bartgol (dismissed):

AaronDonahue:
@mahf708 Should we add the more detailed text to this PR?  That way we can get this merged sooner than later?  @bartgol would you be open to adding the more detailed context?

bartgol (approved):

PR #2811, opened 2024-05-04
remove aerocom algo from rrtmgp (breaking change)

mahf708:
After #2804, there's no need for the cloud-top routine inside rrtmgp, so this reverts #2485 (as long promised)

This PR includes a **BREAKING CHANGE** as it will no longer allow users to request eight FM-provided fields. Users needing these fields should plan accordingly if/when this PR is merged.)


bartgol:
> This PR includes a BREAKING CHANGE as it will no longer allow users to request eight FM-provided fields. Users needing these fields should plan accordingly if/when this PR is merged.)

What does "planning accordingly" mean? How do they get those fields now?

bartgol (approved):

bartgol:
I could not verify it on mappy, since the build was too old, and got removed. But on weaver I could verify that the diffs were just missing fields:
```
10:54:07 SUMMARY of cprnc:
10:54:07  A total number of     19 fields were compared
10:54:07           of which      0 had non-zero differences
10:54:07                and      0 had differences in fill patterns
10:54:07                and      0 had different dimension sizes
10:54:07                and      0 had different data types
10:54:07  A total number of      0 fields could not be analyzed
10:54:07  A total number of      0 time-varying fields on file 1 were not found on file 2.
10:54:07  A total number of      0 time-constant fields on file 1 were not found on file 2.
10:54:07  A total number of      8 time-varying fields on file 2 were not found on file 1.
10:54:07  A total number of      0 time-constant fields on file 2 were not found on file 1.
10:54:07   diff_test: the two files DIFFER only in their field lists
```

AaronDonahue:
@brhillman I think this PR might be waiting on you to review too, see Naser's comment above.

PR #2814, opened 2024-05-06
Fix for setting up lat/lon info in PG2 DP case

tcclevenger:
Fix the lat/lon info setup for PG2 grid in IOP run. Basically, if we have a PG2 field (only possible case at the moment is `SGH30` when TMS is turned on), we need to call the IO setup on the topography file so that PG2 lat/lon is loaded correctly.

I changed the DP unit test to PG2 with TMS. The CIME case is still GLL, so both are covered. Eventually everything will be PG2.


components/eamxx/src/control/atmosphere_driver.cpp
```diff
@@ -1031,19 +1031,19 @@ void AtmosphereDriver::set_initial_conditions ()
...
       if (fname == "phis") {
-        // The eamxx field "phis" corresponds to the name
-        // "PHIS_d" on the GLL and Point grids and "PHIS"
-        // on the PG2 grid in the topography file.
+        // For GLL points, phis corresponds to "PHIS_d" in the
+        // topography file. On PG2 grid, dynamics will take care
+        // of computing phis, so do not add to initialized fields.
         if (grid_name == "Physics PG2") {
-          this_grid_topo_file_fnames.push_back("PHIS");
+          // Skip
```
tcclevenger, line 1038: This is a minor "fix". We don't need to load in `phis` on the PG2 grid, since dynamics internal dynamics->pg2 remapper will always overwrite this. Not a bug since nothing was going wrong, just unnecessary.

bartgol (approved):

tcclevenger:
Fails are expected, I'm going to manually merge.

PR #2816, opened 2024-05-07
Improve model output documentation

bartgol:
Add more information (e.g., atm process tendencies, power-user options).

### DIFF
--- a/components/eamxx/docs/user/model_output.md
+++ b/components/eamxx/docs/user/model_output.md
@@ -1,11 +1,20 @@
 # Model output
 
 EAMxx allows the user to configure the desired model output via [YAML](https://yaml.org/) files,
-with each YAML file associated to a different output file.
+with each YAML file associated to a different output file. In order to add an output stream,
+one needs to run `atmchange output_yaml_files+=/path/to/my/output/yaml` (more information on how
+to use `atmchange` can be found [here](./model_input.md#changing-model-inputs-atmchange)).
+During the `buildnml` phase of the case management system, a copy of these YAML files will be copied
+into the RUNDIR/data folder. During this process, the files will be parsed, and any CIME-related
+variable will be resolved accordingly. Therefore, it is not advised to put the original YAML files in RUNDIR/data,
+since upon `buildnml` execution, all the CIME vars will no longer be available in the YAML file,
+making it harder to tweak it, and even harder to share with other users/cases. Another consequence
+of this is that the user should not modify the yaml files in RUNDIR/data, since any modification will
+be lost on the next run of `buildnml`.
 
-## Basic output YAML file syntax
+## Basic output
 
-The following is an example of a simple output request.
+The following is a basic example of an output request.
 
 ```yaml
 %YAML 1.1
@@ -30,7 +39,7 @@ output_control:
 Notice that lists can be equivalently specified in YAML as `Field Names: [f1, f2, f3]`.
 The user can specify fields to be outputted from any of the grids used in the simulation.
 In the example above, we requested fields from both the Physics and Dynamics grid.
-The other parameters are
+The meaning of the other parameters is as follows:
 
 - `Averaging Type`: how the fields are integrated in time before being saved. Valid
   options are
@@ -61,7 +70,7 @@ request to output derived quantities, which will be computed on the fly by the
 I/O interface of EAMxx. There are two types of diagnostic outputs:
 
 - quantities computed as a function of EAMxx fields. These are simply physical quantities
-  that EAMxx does not keep in persistent storage. As of August 2023, the available
+  that EAMxx does not keep in persistent storage. As of May 2024, the available
   derived quantities are (case sensitive):
 
     - `PotentialTemperature`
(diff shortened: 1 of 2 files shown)


components/eamxx/docs/user/model_output.md
```diff
@@ -89,6 +98,12 @@ I/O interface of EAMxx. There are two types of diagnostic outputs:
     - `precip_ice_surf_mass_flux`
     - `precip_total_surf_mass_flux`
     - `surface_upward_latent_heat_flux`
+    - `wind_speed`
```
AaronDonahue, line 101: Peter C. was curious if there was a way we could automate this list.  Like have a small script that parses register_diagnostics.cpp for example and records every option...

Maybe too complicated to implement but this is a good place to ask.

components/eamxx/docs/user/model_output.md
```diff
@@ -89,6 +98,12 @@ I/O interface of EAMxx. There are two types of diagnostic outputs:
...
+    - `AerosolOpticalDepth550nm`
+    - `NumberPath`
+    - `AeroComCld`
+
+  TODO: add some information about what each diagnostic is, perhaps a formula
```
AaronDonahue, line 106: We should make a git-issue for this todo.

components/eamxx/docs/user/model_output.md
```diff
@@ -116,38 +132,75 @@ they are computed on.
...
+
+## Tendencies output
+
+It is also possible to request tendencies of fields that are computed by atmosphere processes,
+on a per-process basis. Since the term "tendency" can be used with slightly different connotations,
+we clarify what we mean by that when it comes to model output: if process P updates field A,
+by the tendency of A from process P we mean `(A_after_P - A_before_P) / dt`, where `dt` is
+the atmosphere timestep.
+
+As of today, the user needs two things in order to get tendencies from a process. E.g., to get
```
AaronDonahue, line 152: Can we say "as of May 2024" ?

components/eamxx/docs/user/model_output.md
```diff
@@ -116,38 +132,75 @@ they are computed on.
...
+the tendencies of `T_mid` and `horiz_winds` from the process `shoc`, one needs:
+
+- `atmchange shoc::compute_tendencies=T_mid,horiz_winds`
+- add `shoc_T_mid_tend` and `shoc_horiz_winds_tend` to the list of fields in the output YAML file
```
AaronDonahue, line 156: maybe "in at least one YAML file"

components/eamxx/docs/user/model_output.md
```diff
@@ -116,38 +132,75 @@ they are computed on.
...
+
+## Additional options
+
+There YAML file shown at the top of this section, together with the remap options in the following
```
AaronDonahue, line 160: "The" YAML file shown...

AaronDonahue (commented):
This is great! Thanks @bartgol .  I had a few comments and minor corrections, but this is essentially ready to go.

components/eamxx/docs/user/model_output.md
bartgol: I think we discussed this on slack: we can certainly produce a script that automates this, but it is not as easy, so I would punt on it for now.

bartgol: To elaborate more, parsing the diagnostics folder is not enough, since some diagnostic classes are able to produce more than one diagnostic (e.g., `WaterPathDiagnostic` can produce Liq/Rain/Rime/Ice/Vap water paths). Also, I don't foresee us adding tens of diagnostics in the future, so I think manual edit of the docs whenever we add one is probably good enough.

components/eamxx/docs/user/model_output.md
bartgol: Ah, right.

components/eamxx/docs/user/model_output.md
bartgol: I went with option C. :)

components/eamxx/docs/user/model_output.md
```diff
@@ -116,38 +132,75 @@ they are computed on.
...
+  is the number of Gauss points along each axis in the 2d spectral element, and `nelems`
+  is the number of horizontal elements. However, due to continuity, the values on the
+  Gauss points on the element boundary must match the values on the neighboring element,
+  resulting in duplicated data. By remapping to a "unique" version of the dynamics grid
+  (which in EAMxx is referred to as "Physics GLL"), we can save roughly 45% of storage.
+  Note: this feature cannot be used along with the horizontal/vertical remap.
+
+## Tendencies output
+
+It is also possible to request tendencies of fields that are computed by atmosphere processes,
```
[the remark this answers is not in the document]
bartgol: It's technically "Updated". I thought this was getting down to a level of detail that the user may not understand. But it's better to be precise, so I'll edit.

components/eamxx/docs/user/model_output.md
bartgol: Yeah, We could probably just put the formula in the header files, and then just say "for details on the formula for each diagnostic, see the comment at the top of their header file in `components/eamxx/src/diagnostics`.

AaronDonahue:
Not sure why this is still testing in the AT since it is `integrate-without-testing` but I think I will will just manually merge since it just updates docs.

PR #2817, opened 2024-05-08
Add a precipitation adjustment to the ml-correction process.

AaronDonahue:
This commit adds a precipitation adjustment in cases where the ml-correction is applied to qv.  We interpret the hard adjustment of qv in the column as an attempt by ML to represent the missing precipitation in the coarse resolution run vs. the fine resolution run.  In order to have balance with the land models we need to add this adjustment to the precipitation, otherwise we risk drying out the land.

### DIFF
--- a/components/eamxx/src/physics/ml_correction/eamxx_ml_correction_process_interface.cpp
+++ b/components/eamxx/src/physics/ml_correction/eamxx_ml_correction_process_interface.cpp
@@ -104,7 +104,7 @@ void MLCorrection::run_impl(const double dt) {
   const auto &v               = get_field_out("horiz_winds").get_component(1).get_view<Real **, Host>();
 
   // For precipitation adjustment we need to track the change in column integrated 'qv'
-  auto qv_told = Kokkos::create_mirror_view(qv);
+  host_view2d_type qv_told("",qv.extent(0),qv.extent(1));
   Kokkos::deep_copy(qv_told,qv);
 
 
@@ -149,7 +149,7 @@ void MLCorrection::run_impl(const double dt) {
   ekat::enable_fpes(fpe_mask);   
 
   // Now back out the qv change abd apply it to precipitation, only if Tq ML is turned on
-  if (not (m_ML_model_path_tq == "None")) {
+  if (m_ML_model_path_tq != "None") {
     using PC  = scream::physics::Constants<Real>;
     using KT  = KokkosTypes<DefaultDevice>;
     using MT  = typename KT::MemberType;
@@ -168,12 +168,12 @@ void MLCorrection::run_impl(const double dt) {
       auto qold_icol = ekat::subview(qv_told,icol);
       auto qnew_icol = ekat::subview(qv_tnew,icol);
       auto rho_icol  = ekat::subview(pseudo_density,icol);
-      Real wvp = 0;
+      Real net_column_moistening = 0;
       // Compute WaterVaporPath Difference
       Kokkos::parallel_reduce(Kokkos::TeamVectorRange(team, num_levs),
                               [&] (const int& ilev, Real& lsum) {
         lsum += (qnew_icol(ilev)-qold_icol(ilev)) * rho_icol(ilev) / g;
-      },wvp);
+      },net_column_moistening);
       team.team_barrier();
       // Adjust Precipitation
       //  - Note, we subtract the water vapor path because positive precip represents
@@ -181,24 +181,32 @@ void MLCorrection::run_impl(const double dt) {
       auto tot_precip = precip_liq_surf_mass(icol)+precip_ice_surf_mass(icol);
       if (tot_precip>0) {
         // adjust precip by weighted avg of both phases
-        auto liq_w = precip_liq_surf_mass(icol)/tot_precip;
-        auto ice_w = precip_ice_surf_mass(icol)/tot_precip;
-        precip_liq_surf_mass(icol) -= liq_w*wvp;
-        precip_ice_surf_mass(icol) -= ice_w*wvp;
+        Kokkos::single(Kokkos::PerTeam(team), [&] {
+          auto liq_frac = precip_liq_surf_mass(icol)/tot_precip;
+          auto ice_frac = precip_ice_surf_mass(icol)/tot_precip;
+          precip_liq_surf_mass(icol) -= liq_frac*net_column_moistening;
+          precip_ice_surf_mass(icol) -= ice_frac*net_column_moistening;
+	});
       } else {
         // Apply all the adjustment to a single phase based on surface temperature
-        auto T_icol = ekat::subview(T_mid,icol);
-        if (T_icol(m_num_levs-1)>273.15) {
-          precip_liq_surf_mass(icol) -= wvp;
-        } else {
-          precip_ice_surf_mass(icol) -= wvp;
-        }
+        Kokkos::single(Kokkos::PerTeam(team), [&] {
+          auto T_icol = ekat::subview(T_mid,icol);
+          if (T_icol(m_num_levs-1)>273.15) {
+            precip_liq_surf_mass(icol) -= net_column_moistening;
+          } else {
+            precip_ice_surf_mass(icol) -= net_column_moistening;
+          }
+	});
       }
       if (precip_liq_surf_mass(icol)<0) {
-        precip_liq_surf_mass(icol) = 0.0;
+        Kokkos::single(Kokkos::PerTeam(team), [&] {
+          precip_liq_surf_mass(icol) = 0.0;
+	});
       }
       if (precip_ice_surf_mass(icol)<0) {
-        precip_ice_surf_mass(icol) = 0.0;
+        Kokkos::single(Kokkos::PerTeam(team), [&] {
+          precip_ice_surf_mass(icol) = 0.0;
+	});
       }
     });
   }
--- a/components/eamxx/src/physics/ml_correction/eamxx_ml_correction_process_interface.cpp
+++ b/components/eamxx/src/physics/ml_correction/eamxx_ml_correction_process_interface.cpp
@@ -2,6 +2,7 @@
 #include "ekat/ekat_assert.hpp"
 #include "ekat/util/ekat_units.hpp"
 #include "share/field/field_utils.hpp"
+#include "physics/share/physics_constants.hpp"
 
 namespace scream {
 // =========================================================================================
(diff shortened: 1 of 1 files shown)


AaronDonahue:
I've asked Andre from AI2 to take a look at this PR and give comments.  I don't believe he has been added to the SCREAM group w.r.t. reviewing and adding PR's to SCREAM.  We may want to add him, @PeterCaldwell ?

components/eamxx/src/physics/ml_correction/eamxx_ml_correction_process_interface.cpp
```diff
@@ -135,6 +147,61 @@ void MLCorrection::run_impl(const double dt) {
...
+    using MT  = typename KT::MemberType;
+    using ESU = ekat::ExeSpaceUtils<typename KT::ExeSpace>;
+    const auto &pseudo_density       = get_field_in("pseudo_density").get_view<const Real**>();
+    const auto &precip_liq_surf_mass = get_field_out("precip_liq_surf_mass").get_view<Real *, Host>();
+    const auto &precip_ice_surf_mass = get_field_out("precip_ice_surf_mass").get_view<Real *, Host>();
+    constexpr Real g = PC::gravit;
+    const auto num_levs = m_num_levs;
+    const auto policy = ESU::get_default_team_policy(m_num_cols, m_num_levs);
+    
+    const auto &qv_tnew = get_field_in("qv").get_view<const Real **>();
```
frodre, line 164: What is the difference between `get_field_in` and `get_field_out`? Above we use the `get_field_out` to access the `qv` field.

components/eamxx/src/physics/ml_correction/eamxx_ml_correction_process_interface.cpp
```diff
@@ -135,6 +147,61 @@ void MLCorrection::run_impl(const double dt) {
...
+    Kokkos::parallel_for("Compute WVP diff", policy,
+                         KOKKOS_LAMBDA(const MT& team) {
+      const int icol = team.league_rank();
```
frodre, line 167: This section by using the `parellel_for` operates on each column individually, right?

components/eamxx/src/physics/ml_correction/eamxx_ml_correction_process_interface.cpp
```diff
@@ -135,6 +147,61 @@ void MLCorrection::run_impl(const double dt) {
...
+      auto qold_icol = ekat::subview(qv_told,icol);
+      auto qnew_icol = ekat::subview(qv_tnew,icol);
+      auto rho_icol  = ekat::subview(pseudo_density,icol);
+      Real wvp = 0;
```
frodre, line 171: Since this is a key variable here, I would use a more descriptive name. e.g., `net_column_moistening`. Then it meshes with the variables `tot_precip` below and describes why we're subtracting.

components/eamxx/src/physics/ml_correction/eamxx_ml_correction_process_interface.cpp
```diff
@@ -135,6 +147,61 @@ void MLCorrection::run_impl(const double dt) {
...
+      // Compute WaterVaporPath Difference
+      Kokkos::parallel_reduce(Kokkos::TeamVectorRange(team, num_levs),
+                              [&] (const int& ilev, Real& lsum) {
+        lsum += (qnew_icol(ilev)-qold_icol(ilev)) * rho_icol(ilev) / g;
```
frodre, line 175: Just for personal knowledge.  The syntax here within `{}` specifies the reduction which is then saved to `wvp`?

components/eamxx/src/physics/ml_correction/eamxx_ml_correction_process_interface.cpp
```diff
@@ -135,6 +147,61 @@ void MLCorrection::run_impl(const double dt) {
...
+      },wvp);
+      team.team_barrier();
+      // Adjust Precipitation
+      //  - Note, we subtract the water vapor path because positive precip represents
+      //    a descrease in qv.
+      auto tot_precip = precip_liq_surf_mass(icol)+precip_ice_surf_mass(icol);
+      if (tot_precip>0) {
+        // adjust precip by weighted avg of both phases
+        auto liq_w = precip_liq_surf_mass(icol)/tot_precip;
```
frodre, line 184: `liquid_frac`, `ice_frac`

components/eamxx/src/physics/ml_correction/eamxx_ml_correction_process_interface.cpp
```diff
@@ -135,6 +147,61 @@ void MLCorrection::run_impl(const double dt) {
...
+      } else {
+        // Apply all the adjustment to a single phase based on surface temperature
+        auto T_icol = ekat::subview(T_mid,icol);
+        if (T_icol(m_num_levs-1)>273.15) {
+          precip_liq_surf_mass(icol) -= wvp;
+        } else {
+          precip_ice_surf_mass(icol) -= wvp;
+        }
+      }
+      if (precip_liq_surf_mass(icol)<0) {
```
frodre, line 197: I confirmed that the general precip adjustment strategy is how we implemented it in `fv3` as well.  I was curious whether there was any handling specific to columns that do or do not have physics precip, but there's no special handling.

frodre (approved):
LGTM, thanks @AaronDonahue.  I had a few minor requests about variable naming and a couple general questions for my understanding's sake.

components/eamxx/src/physics/ml_correction/eamxx_ml_correction_process_interface.cpp
AaronDonahue: There is added protection for the `get_field_in` call such that it must be `const`.  In other words, a field can be gathered either by get_field_in or get_field_out within an atm. process.  With the following rules about when the field was registered w/ the process:

as Required: only get_field_in
as Computed: only_get_field_out
as Updated: both get_field_in and get_field_out

For protection though, if you are grabbing a field with no intention of changing it you use get_field_in, even if you could use get_field_out.

components/eamxx/src/physics/ml_correction/eamxx_ml_correction_process_interface.cpp
AaronDonahue: yes

components/eamxx/src/physics/ml_correction/eamxx_ml_correction_process_interface.cpp
AaronDonahue: right, this is actually a lambda definition for what is being reducted.

components/eamxx/src/physics/ml_correction/eamxx_ml_correction_process_interface.cpp
AaronDonahue: great! thanks

components/eamxx/src/physics/ml_correction/eamxx_ml_correction_process_interface.cpp
bartgol: Some more historical context: in the early SCREAM days, fields were templated on data type, which was `Real` for output fields and `const Real` for input fields. At that time, we had no choice but to store them separately, and therefore have two separate `get_field_in/out` methods. Now, fields are no longer templated on the data type (they handle read-only constraints at runtime), so we _could_ merge the `get_field_in/out` methods in one (and I thought about doing it).

components/eamxx/src/physics/ml_correction/eamxx_ml_correction_process_interface.cpp
```diff
@@ -86,16 +91,23 @@ void MLCorrection::run_impl(const double dt) {
...
-  const auto &phis            = get_field_in("phis").get_view<const Real *, Host>();
   const auto &SW_flux_dn      = get_field_out("SW_flux_dn").get_view<Real **, Host>();
-  const auto &sfc_alb_dif_vis = get_field_in("sfc_alb_dif_vis").get_view<const Real *, Host>();  
   const auto &sfc_flux_sw_net = get_field_out("sfc_flux_sw_net").get_view<Real *, Host>();
   const auto &sfc_flux_lw_dn  = get_field_out("sfc_flux_lw_dn").get_view<Real *, Host>();
   const auto &u               = get_field_out("horiz_winds").get_component(0).get_view<Real **, Host>();
   const auto &v               = get_field_out("horiz_winds").get_component(1).get_view<Real **, Host>();
 
+  // For precipitation adjustment we need to track the change in column integrated 'qv'
+  auto qv_told = Kokkos::create_mirror_view(qv);
```
bartgol, line 107: This won't work. `qv_told` will just be an alias for `qv`, since they are both on host. What you need is something like
```c++
host_view2d_type qv_told("",qv.extent(0),qv.extent(1));
Kokkos::deep_copy(qv_told,qv);
```

components/eamxx/src/physics/ml_correction/eamxx_ml_correction_process_interface.cpp
```diff
@@ -135,6 +147,61 @@ void MLCorrection::run_impl(const double dt) {
       ML_model_tq, ML_model_uv, ML_model_sfc_fluxes, datetime_str);
   pybind11::gil_scoped_release no_gil;  
   ekat::enable_fpes(fpe_mask);   
+
+  // Now back out the qv change abd apply it to precipitation, only if Tq ML is turned on
+  if (not (m_ML_model_path_tq == "None")) {
```
bartgol, line 152: I find backward conditionals to be harder to read. Why not just do
```suggestion
if (m_ML_mod_path_tq != "None") {
```

components/eamxx/src/physics/ml_correction/eamxx_ml_correction_process_interface.cpp
```diff
@@ -135,6 +147,61 @@ void MLCorrection::run_impl(const double dt) {
...
+      team.team_barrier();
+      // Adjust Precipitation
+      //  - Note, we subtract the water vapor path because positive precip represents
+      //    a descrease in qv.
+      auto tot_precip = precip_liq_surf_mass(icol)+precip_ice_surf_mass(icol);
+      if (tot_precip>0) {
+        // adjust precip by weighted avg of both phases
+        auto liq_w = precip_liq_surf_mass(icol)/tot_precip;
+        auto ice_w = precip_ice_surf_mass(icol)/tot_precip;
+        precip_liq_surf_mass(icol) -= liq_w*wvp;
```
bartgol, line 186: This is not thread safe. I know you are running this only on Host, for now, but we should avoid issues down the road, if ML becomes device friendly. You should wrap all these blocks in a `Kokkos::single` (with `Kokkos::PerTeam`.

bartgol (changes requested):

components/eamxx/src/physics/ml_correction/eamxx_ml_correction_process_interface.cpp
AaronDonahue: thanks for the fix.

bartgol (approved):

PR #2819, opened 2024-05-09
Use IOP lat/lon in rrtmgp interface

tcclevenger:
Use IOP lat/lon in rrtmgp interface.

Unrelated change: `dx_short` code no longer needed in iop.

### DIFF
--- a/components/eamxx/src/control/atmosphere_driver.cpp
+++ b/components/eamxx/src/control/atmosphere_driver.cpp
@@ -199,9 +199,6 @@ setup_iop ()
                                                        hyam,
                                                        hybm);
 
-  auto dx_short_f = phys_grid->get_geometry_data("dx_short");
-  m_iop->set_grid_spacing(dx_short_f.get_view<const Real,Host>()());
-
   // Set IOP object in atm processes
   m_atm_process_group->set_iop(m_iop);
 }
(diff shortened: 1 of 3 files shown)


tcclevenger:
Setting this as WIP until @bogensch is backfrom vacation.

bogensch:
Thanks @tcclevenger for your work on this.  I will do some testing with this branch and report back.

bogensch:
I have tested this branch and it absolutely solves the problem pertaining to the incoming solar radiation.... Thus I approve of this PR and this should be merged.  

Note that, unfortunately, this did not solve the bulk of the differences I am seeing in v0 and v1 DP multi-day runs (like I thought it would) and will continue to debug to track down what the source of the issue is.

bogensch (approved):

tcclevenger:
The hunt continues.. Thanks, @bogensch!

bogensch:
@tcclevenger I consider this PR ready to go and should be merged in.  Thanks.

bartgol:
> @bartgol Can you take a quick look before I manual merge. DIFF in AT is expected.
> 
> Basically for rrtmgp we need to use the lat/lon from IOP file instead of the one from dynamics in the geometry data, but we don't want to overwrite geometry data for the rest of the run. So I just clone and deep copy if we have IOP object.

I wonder if it would make sense to have an iop_grid, holding just one dof. Maybe it could simplify some IO stuff. I think we should explore that route when we take on the unification of the interpolation of data from files.

bartgol (approved):

tcclevenger:
@bartgol Yes, I completely agree.

PR #2820, opened 2024-05-10
Port eamxx rrtmgp to kokkos

jgfouca:
Overall port strategy is very similar to what was done for RRTMGP standalone. Duplication of impls was done at the lower levels, validation is done at the eamxx_rrtmgp_process_interface level.

Change list:
1) Add new SCREAM_RRTMGP_ENABLE_YAKL and SCREAM_RRTMGP_ENABLE_KOKKOS. SCREAM_RRTMGP_ENABLE_YAKL is on by default and should give us the behavior we've had. If both are on, then kokkos results will validate against YAKL results.
2) Remove some files from eamxx's rrtmgp that were duplicated from the rrtmgp standalone
3) Cleaned up code indentation across all files (please view PR with whitespace changes off)
4) Points the rrtmgp submodule to our kokkos conversion branch.

Concerns:

My main concern is with eamxx_rrtmgp_process_interface.cpp and how the buffer memory is wrapped by YAKL arrays. The original author left comments indicating that they thought the various subview_ND functions were copying the data. What is actually happening is they are creating the YAKL equivalent of an unmanaged View, wrapping the provided raw pointer. The data coming in from the buffer may not be layout left but it is being treated as layout left in the yakl arrays.

I also had some concerns with m_gas_concs and the gas_concs temporary. Because subview_ND wraps the original pointer, changes to m_gas_concs will change gas_concs; not sure if that was the intent.

Testing:

Tested TAS standalone with
1) YAKL only
2) Kokkos only
3) YAKL and Kokkos (with validation)


components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
```diff
@@ -630,6 +804,9 @@ void RRTMGPRadiation::run_impl (const double dt) {
 
       // Create YAKL arrays. RRTMGP expects YAKL arrays with styleFortran, i.e., data has ncol
       // as the fastest index. For this reason we must copy the data.
```
jgfouca, line 806: The original comment is wrong.  The semantics of these subview_Nd functions have a big impact in behavior.

components/eamxx/CMakeLists.txt
```diff
@@ -189,6 +189,9 @@ if (SCREAM_DEBUG)
   endif ()
 endif()
 
+set(DEFAULT_RRTMGP_ENABLE_YAKL TRUE)
```
bartgol, line 192: Since these are hardcoded, and never change, you could prob just use TRUE/FALSE in the `option(...)` lines down below...

components/eamxx/CMakeLists.txt
```diff
@@ -203,13 +206,24 @@ if (NOT SCREAM_SMALL_KERNELS)
...
 
-# Add RRTMGP debug checks. Note, we might consider also adding RRTMGP_EXPENSIVE_CHECKS
+# Add RRTMGP settings. Note, we might consider also adding RRTMGP_EXPENSIVE_CHECKS
 # to turn on the RRTMGP internal checks here as well, via
 #     option (RRTMGP_EXPENSIVE_CHECKS "Turn on internal RRTMGP error checking" ${SCREAM_DEBUG})
 # and then adding to scream_config.h:
 #     #cmakedefine RRTMGP_EXPENSIVE_CHECKS
 option (SCREAM_RRTMGP_DEBUG "Turn on extra debug checks in RRTMGP" ${SCREAM_DEBUG})
 
+option(SCREAM_RRTMGP_ENABLE_YAKL "Use YAKL under rrtmgp" ${DEFAULT_RRTMGP_ENABLE_YAKL})
```
bartgol, line 216: I'm wondering: is there a reason not to move all RRTMGP cmake configurations (new as well as existing ones) down in the src/physics/rrtmgp folder? It seems like a more appropriate place, and helps to keep our main CMakeLists.txt more readable.

components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
```diff
@@ -394,19 +546,18 @@ void RRTMGPRadiation::initialize_impl(const RunType /* run_type */) {
   m_do_subcol_sampling = m_params.get<bool>("do_subcol_sampling",true);
 
   // Initialize yakl
-  yakl_init();
+  init_kls();
```
bartgol, line 549: What is `init_kls`?

components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
bartgol: I may have written the comment myself. I think I wasn't referring to the moment when we do the subview, but rather to the fact that we have all those manual copies down below, from kokkos views to yakl arrays. Stuff like
```c++
            p_lay(i+1,k+1)       = d_pmid(icol,k);
            t_lay(i+1,k+1)       = d_tmid(icol,k);
            z_del(i+1,k+1)       = d_dz(i,k);
            ...
```

bartgol: I agree that the comment is not very clear about that. Perhaps we could rephrase, and maybe move it down near the place where we do the data copy...

components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
```diff
@@ -697,13 +874,112 @@ void RRTMGPRadiation::run_impl (const double dt) {
...
       auto cld_tau_sw_gpt  = subview_3d(m_buffer.cld_tau_sw_gpt);
       auto cld_tau_lw_gpt  = subview_3d(m_buffer.cld_tau_lw_gpt);
+#endif
+#ifdef RRTMGP_ENABLE_KOKKOS
+      // If YAKL is on, we don't want aliased memory in both the yakl and kokos
+      // subviews.
+      auto subview_1dk = [&](const real1dk v) -> real1dk {
+        real1dk subv(v, std::make_pair(0, ncol));
+#ifdef RRTMGP_ENABLE_YAKL
+        real1dk rv(v.label(), ncol);
```
bartgol, line 884: I don't think we need to deep copy here. I was really just taking a subview, so that all the data copy down below could be done.

components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
tcclevenger: Yeah, the buffer variables are there to be the YAKL layout left copy of Kokkos field views. This process controls all the data in the buffer, so no data is coming from anywhere else.

I think the tough part with the comment is that some of the buffer arrays are copied from input fields before running rrtmgp_main(), and some buffer arrays are used as output from rrtmgp_main() and then copied to field views. I think we'll need at least some comment here as like "these are the yakl fields for rrtmgp internal to use" and then comments where input and output data is copied.

components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
```diff
@@ -1056,11 +1609,68 @@ void RRTMGPRadiation::run_impl (const double dt) {
...
     // Restore the refCounted array.
+#ifdef RRTMGP_ENABLE_YAKL
     m_gas_concs.concs = gas_concs;
-
+    m_gas_concs.ncol = orig_ncol;
+#endif
+#ifdef RRTMGP_ENABLE_KOKKOS
+    m_gas_concs_k.concs = gas_concs_k;
+    m_gas_concs_k.ncol = orig_ncol_k;
+    //VALIDATE_KOKKOS(m_gas_concs, m_gas_concs_k);
```
tcclevenger, line 1672: Same here.

components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
```diff
@@ -835,13 +1176,25 @@ void RRTMGPRadiation::run_impl (const double dt) {
...
         Kokkos::fence();
 
         // Populate GasConcs object
+#ifdef RRTMGP_ENABLE_YAKL
         m_gas_concs.set_vmr(name, tmp2d);
+#endif
+#ifdef RRTMGP_ENABLE_KOKKOS
+        COMPARE_WRAP(tmp2d, tmp2d_k);
+        m_gas_concs_k.set_vmr(name, tmp2d_k);
+        //VALIDATE_KOKKOS(m_gas_concs, m_gas_concs_k); broken for some reason
```
tcclevenger, line 1196: This is checked above? If so we can probably remove comment.

components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
```diff
@@ -931,8 +1337,52 @@ void RRTMGPRadiation::run_impl (const double dt) {
...
+        sw_clnclrsky_flux_up_k, sw_clnclrsky_flux_dn_k, sw_clnclrsky_flux_dn_dir_k,
+        sw_clrsky_flux_up_k, sw_clrsky_flux_dn_k, sw_clrsky_flux_dn_dir_k,
+        sw_clnsky_flux_up_k, sw_clnsky_flux_dn_k, sw_clnsky_flux_dn_dir_k,
+        lw_clnclrsky_flux_up_k, lw_clnclrsky_flux_dn_k,
+        lw_clrsky_flux_up_k, lw_clrsky_flux_dn_k,
+        lw_clnsky_flux_up_k, lw_clnsky_flux_dn_k}));
+
+      COMPARE_ALL_WRAP(std::vector<real3d>({sw_bnd_flux_up, sw_bnd_flux_dn, sw_bnd_flux_dir, lw_bnd_flux_up, lw_bnd_flux_dn}),
+                       std::vector<real3dk>({sw_bnd_flux_up_k, sw_bnd_flux_dn_k, sw_bnd_flux_dir_k, lw_bnd_flux_up_k, lw_bnd_flux_dn_k}));
+      //VALIDATE_KOKKOS(m_gas_concs, m_gas_concs_k);
```
tcclevenger, line 1381: Same here.

components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
tcclevenger: Comments out of order, see the one below.

components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
jgfouca: Thanks, that makes sense. So once we have things in a better state, we would just pass d_pmid , or a subview of it, directly to rrtmgp.

components/eamxx/CMakeLists.txt
jgfouca: I can see advantages to both. The current approach allows developers to find all cache var options in one place.

components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
jgfouca: initializes all kernel launchers, yakl and kokkos.

components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
jgfouca: We do need a deep copy if YAKL is enabled:
```
      // If YAKL is on, we don't want aliased memory in both the yakl and kokos                                                                                                                                                 
      // subviews.
```

components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.hpp
```diff
@@ -119,18 +124,33 @@ class RRTMGPRadiation : public AtmosphereProcess {
...
+    real1dk mu0_k;
+    real1dk sfc_alb_dir_vis_k;
+    real1dk sfc_alb_dir_nir_k;
+    real1dk sfc_alb_dif_vis_k;
+    real1dk sfc_alb_dif_nir_k;
+    real1dk sfc_flux_dir_vis_k;
+    real1dk sfc_flux_dir_nir_k;
+    real1dk sfc_flux_dif_vis_k;
+    real1dk sfc_flux_dif_nir_k;
+#endif
```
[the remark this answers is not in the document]
jgfouca: There's a couple members mixed in here that are not duplicated, cosine_zenith and d_dz.

components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
bartgol: Why do you need them deep copied though? What's wrong with an unmanaged one?

jgfouca: Oh, you're assuming there's never anything in the input view that would be used later? It's possible that's true but I didn't verify and I don't think it hurts anything.

We can't use unmanaged views because RRTMGP is not currently coded to accept those.

tcclevenger (approved):

components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
```diff
@@ -812,15 +1089,79 @@ void RRTMGPRadiation::run_impl (const double dt) {
...
+            rel_k(i,k)         = d_rel(icol,k);
+            rei_k(i,k)         = d_rei(icol,k);
+            p_lev_k(i,k)       = d_pint(icol,k);
+            t_lev_k(i,k)       = d_tint(i,k);
+          });
+
+          p_lev_k(i,nlay) = d_pint(icol,nlay);
+          t_lev_k(i,nlay) = d_tint(i,nlay);
+
+          // Note that RRTMGP expects ordering (col,lay,bnd) but the FM keeps things in (col,bnd,lay) order
```
brhillman, line 1117: Should we change the storage in the FM at some point?

brhillman (approved):

PR #2821, opened 2024-05-11
Refactor vertical interpolation in EAMxx

bartgol:
Before, the way we did vertical interpolation, for every field we were redoing the setup of LinInterp, as well as reallocating some views (not really good).

This PR changes a few things:

1. Remove the `scream_vertical_interpolation*.hpp` files altogether.
2. In `VerticalRemapper`, call `ekat::LinInterp` manually. We create up to 4 `ekat::LinInterp` objects, depending on fields to remap: midpoints-vs-interfaces based and packed-nonpacked. At run time, we setup the interpolation objects and reuse them for all the fields of the same type. Also, no runtime device views allocations.
3. In `FieldAtPressureLevel`, we don't even need `ekat::LinInterp`, since the tgt grid has a single entry, and we don't need to reuse the interpolation for multiple evaluations. Simply call `ekat::upper_bound` and manually interpolate.

I still need to chek the performance, but I'm fairly sure that, for a moderately large number of fields to remap (when using the vertical remap feature in IO), we would see a noticeable improvement.


components/eamxx/src/share/grid/remap/vertical_remapper.hpp
```diff
@@ -124,13 +126,25 @@ class VerticalRemapper : public AbstractRemapper
...
+  struct FType {
+    bool packed = false;
+    bool midpoints = true;
+  };
+  std::map<std::string,FType> m_field2type;
+
+  std::shared_ptr<ekat::LinInterp<Real,SCREAM_PACK_SIZE>> m_lin_interp_mid_packed;
+  std::shared_ptr<ekat::LinInterp<Real,SCREAM_PACK_SIZE>> m_lin_interp_int_packed;
+  std::shared_ptr<ekat::LinInterp<Real,1>>                m_lin_interp_mid_scalar;
+  std::shared_ptr<ekat::LinInterp<Real,1>>                m_lin_interp_int_scalar;
```
tcclevenger, line 147: It may not be too important, but this is another time where it would be nice to be able to have a scalar version (and not just pack size 1). Should be a quick task, may look into this over leave if you haven't already.

tcclevenger (approved):
Looks good. 

I need to check if IOP can use the vertical remapper because there we are manually setting up ekat::LinInterp for each field as well.

components/eamxx/src/share/grid/remap/vertical_remapper.hpp
bartgol: Yes, it would be nice. But then you'd have 2 impls for LinInterp. I think at some point we need to a) think about switching to Kokkos::simd, and b) go through our basic functions, and make sure we can call them with either packs or scalars. E.g., index_and_shift (whose name I would probably change) only works for packed views, forcing different impls. If we had something like
```
template<typename ScalarT, typename IndexT>
ScalarT eval_at_index (const view_1d<ScalarT>& v, const IndexT& i);
```
with specializations for Packs[^1], then we may be able to have a single impl for LinInterp, which works for both packs and scalars.

[^1]: since partial specialization is not possible for function, we'd need a helper struct, but that's a minor detail.

bartgol:
The diff is b/c we forgot to bless baselines after merging the mam4xx non-bfb pr. Merging.

ndkeen:
Would it be surprising to see 2.8x speedup with cess-style ne1024 case?

Trying to explain a large speedup between a run with checkout before this change and one with early July checkout.

```
< "a:EAMxx::IO::vert_remap"                                     -       2048     2048 1.064141e+07   3.724505e+05   197.325 (   508      0)   138.164 (   740      0)
> "a:EAMxx::IO::vert_remap"                                     -       2048     2048 1.064141e+07   2.071911e+04    70.652 (    76      0)     9.645 (  1931      0)
```

where horiz is about the same:
```
< "a:EAMxx::IO::horiz_remap"                                    -       2048     2048 1.147699e+07   3.588312e+05   243.145 (  1654      0)   122.785 (   739      0)
> "a:EAMxx::IO::horiz_remap"                                    -       2048     2048 1.147699e+07   2.804111e+05   239.984 (  1435      0)    92.531 (  1534      0)
```

bartgol:
@ndkeen It's great to see that the speedup is quite noticeable. The feature was not heavily used (if used at all), so I didn't have much data. But yes, I was expecting some speedup in vertical remap (it was the whole point of this PR, basically). 2.8x in that timer is a yummy treat.

I'd have to check how many streams have vremap enabled, to realize how important it is in absolute terms. But given that vremap is 100% rank-local, it was disappointing to see it in the same ball park as hremap. 70s vs 240s is more satisfying.

PR #2822, opened 2024-05-13
pbl entrainment budget diags

mahf708:
adding pbl entrainment budget diags to eamxx

### DIFF
--- a/components/eamxx/src/diagnostics/pbl_entrainment_budget.cpp
+++ b/components/eamxx/src/diagnostics/pbl_entrainment_budget.cpp
@@ -149,6 +149,7 @@ void PBLEntrainmentBudget::compute_diagnostic_impl() {
   using ESU = ekat::ExeSpaceUtils<typename KT::ExeSpace>;
 
   constexpr Real g = PC::gravit;
+  Real fill_value  = constants::DefaultFillValue<Real>().value;
 
   // Before doing anything, subview the out field for each variable
   auto out = m_diagnostic_output.get_view<Real **>();
@@ -227,7 +228,7 @@ void PBLEntrainmentBudget::compute_diagnostic_impl() {
         // biggest positive jump in theta_l. Third, we can find the level which
         // has the biggest negative jump in qt.
 
-        int opt_tm_grad_lev = 0;
+        int opt_tm_grad_lev = 1;
         // Find tm_grad (tm_grad is a catch-all for the 3 methods)
         Kokkos::parallel_for(
             Kokkos::TeamVectorRange(team, 1, num_levs), [&](int k) {
@@ -269,49 +270,60 @@ void PBLEntrainmentBudget::compute_diagnostic_impl() {
         team.team_barrier();
         opt_tm_grad_lev = minloc.loc;
 
-        // Save some outputs just above the "mixed" PBL
-        o_pm_hplus(icol) = pm_icol(opt_tm_grad_lev - 1);
-        o_tl_hplus(icol) = tl_icol(opt_tm_grad_lev - 1);
-        o_qt_hplus(icol) = qt_icol(opt_tm_grad_lev - 1);
+        if(opt_tm_grad_lev < 2 || opt_tm_grad_lev > num_levs - 1) {
+          // Weird stuff can happen near the top and bottom of atm, so fill_val
+          o_pm_hplus(icol) = fill_value;
+          o_tl_hplus(icol) = fill_value;
+          o_qt_hplus(icol) = fill_value;
+          o_df_inpbl(icol) = fill_value;
+          o_tl_caret(icol) = fill_value;
+          o_qt_caret(icol) = fill_value;
+          o_tl_ttend(icol) = fill_value;
+          o_qt_ttend(icol) = fill_value;
+        } else {
+          // Save some outputs just above the "mixed" PBL
+          o_pm_hplus(icol) = pm_icol(opt_tm_grad_lev - 1);
+          o_tl_hplus(icol) = tl_icol(opt_tm_grad_lev - 1);
+          o_qt_hplus(icol) = qt_icol(opt_tm_grad_lev - 1);
 
-        // Save the dF term (F(h) - F(0))
-        o_df_inpbl(icol) = (sd_icol(opt_tm_grad_lev - 1) - sd_icol(0)) -
-                           (su_icol(opt_tm_grad_lev - 1) - su_icol(0)) +
-                           (ld_icol(opt_tm_grad_lev - 1) - ld_icol(0)) -
-                           (lu_icol(opt_tm_grad_lev - 1) - lu_icol(0));
+          // Save the dF term (F(h) - F(0))
+          o_df_inpbl(icol) = (sd_icol(opt_tm_grad_lev - 1) - sd_icol(0)) -
+                             (su_icol(opt_tm_grad_lev - 1) - su_icol(0)) +
+                             (ld_icol(opt_tm_grad_lev - 1) - ld_icol(0)) -
+                             (lu_icol(opt_tm_grad_lev - 1) - lu_icol(0));
 
-        // Now only need to compute below from opt_tm_grad_lev to num_levs
-        // Integrate through the PBL, mass-weighted
-        // TODO:
-        // combine/refactor this once inner parallel_reduce
-        // with multiple results/sums is supported...
-        Kokkos::parallel_reduce(
-            Kokkos::TeamVectorRange(team, opt_tm_grad_lev, num_levs),
-            [&](const int &k, Real &result) {
-              result += tl_icol(k) * pd_icol(k) / g;
-            },
-            o_tl_caret(icol));
-        Kokkos::parallel_reduce(
-            Kokkos::TeamVectorRange(team, opt_tm_grad_lev, num_levs),
-            [&](const int &k, Real &result) {
-              result += qt_icol(k) * pd_icol(k) / g;
-            },
-            o_qt_caret(icol));
-        Kokkos::parallel_reduce(
-            Kokkos::TeamVectorRange(team, opt_tm_grad_lev, num_levs),
-            [&](const int &k, Real &result) {
-              auto tl_tend = (tl_icol(k) - prev_tliq_icol(k)) / dt;
-              result += tl_tend * pd_icol(k) / g;
-            },
-            o_tl_ttend(icol));
-        Kokkos::parallel_reduce(
-            Kokkos::TeamVectorRange(team, opt_tm_grad_lev, num_levs),
-            [&](const int &k, Real &result) {
-              auto qt_tend = (qt_icol(k) - prev_qtot_icol(k)) / dt;
-              result += qt_tend * pd_icol(k) / g;
-            },
-            o_qt_ttend(icol));
-
+          // Now only need to compute below from opt_tm_grad_lev to num_levs
+          // Integrate through the PBL, mass-weighted
+          // TODO:
+          // combine/refactor this once inner parallel_reduce
+          // with multiple results/sums is supported...
+          Kokkos::parallel_reduce(
+              Kokkos::TeamVectorRange(team, opt_tm_grad_lev, num_levs),
+              [&](const int &k, Real &result) {
+                result += tl_icol(k) * pd_icol(k) / g;
+              },
+              o_tl_caret(icol));
+          Kokkos::parallel_reduce(
+              Kokkos::TeamVectorRange(team, opt_tm_grad_lev, num_levs),
+              [&](const int &k, Real &result) {
+                result += qt_icol(k) * pd_icol(k) / g;
+              },
+              o_qt_caret(icol));
+          Kokkos::parallel_reduce(
+              Kokkos::TeamVectorRange(team, opt_tm_grad_lev, num_levs),
+              [&](const int &k, Real &result) {
+                auto tl_tend = (tl_icol(k) - prev_tliq_icol(k)) / dt;
+                result += tl_tend * pd_icol(k) / g;
+              },
+              o_tl_ttend(icol));
+          Kokkos::parallel_reduce(
+              Kokkos::TeamVectorRange(team, opt_tm_grad_lev, num_levs),
+              [&](const int &k, Real &result) {
+                auto qt_tend = (qt_icol(k) - prev_qtot_icol(k)) / dt;
+                result += qt_tend * pd_icol(k) / g;
+              },
+              o_qt_ttend(icol));
+        }
         // release stuff from wsm
         ws.release_many_contiguous<wsms>({&tm_grad});
       });
--- a/components/eamxx/src/diagnostics/tests/pbl_entrainment_budget_test.cpp
+++ b/components/eamxx/src/diagnostics/tests/pbl_entrainment_budget_test.cpp
@@ -200,6 +200,10 @@ TEST_CASE("entrainment_budget") {
     ld.get_header().get_tracking().update_time_stamp(t2);
     lu.get_header().get_tracking().update_time_stamp(t2);
 
+    // This diag call ensures that random values are
+    // correctly set in the diagnostic, with no numerical bugs
+    diag->compute_diagnostic();
+
     auto tm_v = tm.get_view<Real **, Host>();
     auto pm_v = pm.get_view<Real **, Host>();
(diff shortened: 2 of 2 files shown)


components/eamxx/src/diagnostics/extraaci.cpp
```diff
@@ -0,0 +1,255 @@
...
+                               [&](int k) {
+                                 tm_grad(k) = (tm_icol(k - 1) - tm_icol(k)) /
+                                              (pm_icol(k - 1) - pm_icol(k));
+                               });
+          team.team_barrier();
+
+          // Find minimum gradient, because d(p_mid) < 0 in definition above
+          // Starting from the surface, and ensuring p_mid > 70000.0 Pa,
+          // to avoid resolving to some odd place higher up in the atmosphere.
+          for(int k = num_levs; k > 1; k--) {
```
bartgol, line 222: This part of the code should be contained in a `Kokkos::single`:
```c++
Kokkos::singLe(Kokkos::PerTeam(team),[&]{
   for (...) {
     ...
   }
});
```

Technically, the way you wrote it, it works, since `min_tm_grad_lev` is an automatic var, which is thread-local, and all threads in the team will compute the same thing. Still...

bartgol: You could even think of using a `parallel_reduce`, if you feel like it, and avoid the single.

components/eamxx/src/diagnostics/extraaci.cpp
```diff
@@ -0,0 +1,255 @@
...
+
+        const auto qt_icol = ekat::subview(qt_v, icol);
+        const auto tl_icol = ekat::subview(tl_v, icol);
+
+        auto prev_qtot_icol = ekat::subview(prev_qtot_v, icol);
+        auto prev_tliq_icol = ekat::subview(prev_tliq_v, icol);
+
+        auto ws = wsm.get_workspace(team);
+        ekat::Unmanaged<WSMgr::view_1d<Real>> tm_grad;
+        ws.take_many_contiguous_unsafe<wsms>({"tm_grad"}, {&tm_grad});
```
bartgol, line 194: It's not wrong, but since you use only one slot, you can prob just use `auto tm_grad = ws.take("tm_grad")`.

components/eamxx/src/diagnostics/extraaci.cpp
```diff
@@ -0,0 +1,255 @@
...
+              // the integrated (averaged) quantities
+              o_tl_caret(icol) += tl_icol(k) * pd_icol(k) / g;
+              o_tl_ttend(icol) += tl_tend * pd_icol(k) / g;
+              o_qt_caret(icol) += qt_icol(k) * pd_icol(k) / g;
+              o_qt_ttend(icol) += qt_tend * pd_icol(k) / g;
+            });
+        team.team_barrier();
+
+        // release stuff from wsm
+        ws.release_many_contiguous<wsms>({&tm_grad});
```
bartgol, line 251: If you switch to `take`, then this can become a simple `release()`

components/eamxx/src/diagnostics/extraaci.cpp
```diff
@@ -0,0 +1,255 @@
...
+        }
+
+        // Save some outputs just above the "mixed" PBL
+        o_pm_hplus(icol) = pm_icol(min_tm_grad_lev - 1);
+        o_tl_hplus(icol) = tl_icol(min_tm_grad_lev - 1);
+        o_qt_hplus(icol) = qt_icol(min_tm_grad_lev - 1);
+
+        // Now only need to compute below from min_tm_grad_lev to num_levs
+        Kokkos::parallel_for(
+            Kokkos::TeamThreadRange(team, min_tm_grad_lev, num_levs),
```
bartgol, line 237: Again, when you don't have a third nested layer of parallelism, prefer TVR over TTR.

components/eamxx/src/diagnostics/extraaci_util.hpp
```diff
@@ -0,0 +1,45 @@
...
+  std::map<std::string, std::string> units_map;
+  int size;
+  std::string pblinvalg;
+
+  ExtraAciDiagUtil() {
+    // Start (post) incrementing size from 0
+    size = 0;
+
+    // pblp
+    index_map["p+"] = size++;
```
bartgol, line 19: What are `+` and `^` in the diags names?

components/eamxx/src/diagnostics/tests/extraaci_test.cpp
```diff
@@ -0,0 +1,184 @@
...
+  using namespace ShortFieldTagsNames;
+  using namespace ekat::units;
+  // A world comm
+  ekat::Comm comm(MPI_COMM_WORLD);
+
+  scorpio::init_subsystem(comm);
+
+  std::string iop_file =
+      "/global/cfs/cdirs/e3sm/inputdata/atm/cam/scam/iop/"
+      "DYCOMSrf02_iopfile_4scam.nc";
```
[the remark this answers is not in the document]
bartgol: You can pass input args to your unit test. In your CMakeLists.txt, do
```cmake
GetInputFile(cam/scam/iop/DYCOMSrf02_iopfile_4scam.nc)
CreateDiagTest(extraaci "extraaci_test.cpp"
  EXE_ARGS "--ekat-test-params data=${SCREAM_INPUT_ROOT}//atm/cam/scam/iop/DYCOMSrf02_iopfile_4scam.nc)
```
and in the cpp file for the test, you can do
```cpp
const auto& data = ekat::TestSession::get().params.at("data");
```
You can use whatever string you want instead of `data`, so long as it's the same in the cpp file and in CMakeLists.txt.

components/eamxx/src/diagnostics/tests/extraaci_test.cpp
```diff
@@ -0,0 +1,184 @@
...
+    diag->get_diagnostic().sync_to_host();
+
+    const auto qc_h = qc.get_view<const Real **, Host>();
+    const auto qv_h = qv.get_view<const Real **, Host>();
+    const auto tm_h = tm.get_view<const Real **, Host>();
+    const auto pm_h = pm.get_view<const Real **, Host>();
+
+    const auto out_hf = diag->get_diagnostic();
+    auto out_hv       = out_hf.get_view<Real **, Host>();
+    Field out_tf      = diag->get_diagnostic().clone();
```
bartgol, line 173: I'm guessing this is leftover from previous iterations, since `out_t` is never really used...

components/eamxx/src/diagnostics/extraaci_util.hpp
bartgol: Are these "common" names in this context? If not, I would rather use some more wordy names...

components/eamxx/src/diagnostics/extraaci.cpp
```diff
@@ -0,0 +1,268 @@
...
+          // Find minimum gradient, because d(p_mid) < 0 in definition above
+          // Starting from the surface, and ensuring p_mid > 70000.0 Pa,
+          // to avoid resolving to some odd place higher up in the atmosphere.
+          using minloc_t       = Kokkos::MinLoc<Real, int>;
+          using minloc_value_t = typename minloc_t::value_type;
+          minloc_value_t minloc;
+          Kokkos::parallel_reduce(
+              Kokkos::TeamVectorRange(team, 1, num_levs),
+              [&](const int &k, minloc_value_t &result) {
+                if(tm_grad(k) < result.val  && pm_icol(k) > 70000.0) {
```
bartgol, line 227: I think you also need `&& k<result.loc`.

components/eamxx/src/diagnostics/extraaci.cpp
```diff
@@ -0,0 +1,268 @@
...
+                  result.val = tm_grad(k);
+                  result.loc = k;
+                }
+              },
+              minloc_t(minloc));
+          opt_tm_grad_lev = minloc.loc;
```
bartgol, line 233: You need a team barrier before this assignment, to make sure the reduction has fully completed before any thread attempts to read minloc.

components/eamxx/src/diagnostics/entrainment_budget.cpp
```diff
@@ -0,0 +1,300 @@
...
+  const auto &prev_tliq_v = m_prev_tl.get_view<Real **>();
+
+  view_2d qt_v("qt_v", m_ncols, m_nlevs);
+  view_2d tl_v("tl_v", m_ncols, m_nlevs);
+  calc_tl_qt(tm_v, pm_v, qv_v, qc_v, tl_v, qt_v);
+
+  const auto &curr_ts =
+      get_field_in("qc").get_header().get_tracking().get_time_stamp();
+
+  std::int64_t dt = curr_ts - m_start_t;
```
tcclevenger, line 185: I'd use `auto` here.

components/eamxx/src/diagnostics/entrainment_budget.cpp
```diff
@@ -0,0 +1,300 @@
...
+            Kokkos::TeamVectorRange(team, opt_tm_grad_lev, num_levs),
+            [&](int k) {
+              // The time-based tendencies
+              auto tl_tend = (tl_icol(k) - prev_tliq_icol(k)) / dt;
+              auto qt_tend = (qt_icol(k) - prev_qtot_icol(k)) / dt;
+              // the integrated (averaged) quantities
+              o_tl_caret(icol) += tl_icol(k) * pd_icol(k) / g;
+              o_tl_ttend(icol) += tl_tend * pd_icol(k) / g;
+              o_qt_caret(icol) += qt_icol(k) * pd_icol(k) / g;
+              o_qt_ttend(icol) += qt_tend * pd_icol(k) / g;
```
tcclevenger, line 292: I think we need reductions to compute these columns sums. We can run into trouble if different threads are attempting to update the same value.

You can do an arbitrary amount of reducers in the same call
```
Real sum1, sum2, ...;
Kokkos::parallel_reduce(Kokkos::TeamVectorRange(team, opt_tm_grad_lev, num_levs),
                        [&](int k, Real& local_sum1, Real& local_sum2, ...) {
  local_sum1 += ...;
  .
  .
  .
}, sum1, sum2, ...);
```
And the after well need a Kokkos:single to add in the column sums
```
Kokkos::single(Kokkos::PerTeam(team), [&]() {
  o_tl_caret(icol) += sum1;
  .
  .
  .
});
```

tcclevenger: I believe it overwrites the value you already have set. You could do the reduction (as you have it written) and then add in the other part, but you would still need a Kokkos::single.

tcclevenger: If `wp` contains stuff you want to keep, then I think so. I need to check this, maybe if you only hand a scalar `sum` (and not a `Kokkos::Sum<Real>(sum)`) as the last arg it won't overwrite? But I'm 95% sure it used the default `init` for sum reduction which sets the value to 0 at the start.

Also, that team_barrier is unnecessary since it's the end of the kernel.

tcclevenger: The only issue in your current code is that you set the value 
```
o_tl_caret(icol) = initial_value
```
and then want to add a column sum to that,
```
o_tl_caret(icol) = initial_value + column_val(opt_tm_grad_lev) + ... + column_val(nlevs-1)
``` 
but the `parallel_reduce()` is going to overwrite `o_tl_caret(icol)` to just be whatever the result of the reduction is, so you lose any prior info you had
```
o_tl_caret(icol) = 0 + column_val(opt_tm_grad_lev) + ... + column_val(nlevs-1)
```

In the water path example, if you want 
```
wp(icol) = q_icol(0)*rho_icol(0)/g + ... + q_icol(nlevs-1)*rho_icol(nlevs-1)/g
```
then it is doing exactly what you want.

tcclevenger: ahh I completely read it wrong!

tcclevenger: My bad! I was just looking at another output variable. So yes, you will need a reduction, but can just hand in the o_tl_caret(icol) to the reducer.

components/eamxx/src/diagnostics/entrainment_budget.cpp
bartgol: Yes. The return type of op- may or may not change in the future, so unless you _need_ to use a special type, auto is best.

tcclevenger:
What is the status of this? It's labeled WIP, is it a part of the output fails you are trying to debug?

tcclevenger:
If you aren't using this version of the diag (and no one else is needing it) then I guess we should leave it unmerged. Re-request reviews if/when you want to merge it.

components/eamxx/src/diagnostics/pbl_entrainment_budget.cpp
```diff
@@ -0,0 +1,332 @@
...
+  // Get qc and qv
+  add_field<Required>("qc", scalar2d_layout, kg / kg, grid_name);
+  add_field<Required>("qv", scalar2d_layout, kg / kg, grid_name);
+  // Get T_mid, p_mid
+  add_field<Required>("T_mid", scalar2d_layout, K, grid_name);
+  add_field<Required>("p_mid", scalar2d_layout, Pa, grid_name);
+  // Get pseudo_density
+  add_field<Required>("pseudo_density", scalar2d_layout, Pa, grid_name);
+  // Get radiation up and down terms
+  add_field<Required>("SW_flux_dn", scalar2d_layout, W / m * m, grid_name);
```
bartgol, line 73: Watch out, these units here will resolve to `W`, since you don't have parentheses!

Note: this will have _zero_ impact on this code working, since I/O does not check the units of the requested fields. But we should fix this, a) to avoid confusion and b) in case I/O starts to check units of the diags required fields.

components/eamxx/src/diagnostics/pbl_entrainment_budget.cpp
```diff
@@ -0,0 +1,332 @@
...
+
+  const auto nondim = Units::nondimensional();
+
+  // Set the index map and units map
+  PBLEntrainmentBudgetDiagUtil eadu;
+  m_index_map = eadu.index_map;
+  m_units_map = eadu.units_map;
+  m_ndiag     = eadu.size;
+
+  if(eadu.pblinvalg == "temperature-inversion") {
```
bartgol, line 32: This is the only possible scenario, since `eadu` is default-constructed, so it will always have `"temperature-inversion"` as `pblinvalg`. In the long run, I'm guessing this code should change into something like
```cpp
  PBLEntrainmentBudgetDiagUtil eadu;
  m_index_map = eadu.index_map;
  m_units_map = eadu.units_map;
  m_ndiag     = eadu.size;
  eadu.pblinvalg = m_params.get<std::string>("pblinvalg");
  if (eadu.pblinvalg=="temperature-inversion") {
    m_pblinvalg = 0;
  } else if (...) {
  } else {
    EKAT_ERROR_MSG ("Watch out, silly: invalid value for pblinvalg!\n");
  }
```

bartgol (commented):
I don't see any mods to `scorpio_output.cpp`. The user can still request it using `PBLEntrainmentBudget` as field in the output yaml, but there's no control over the options. If you will ever want to support the other `pblinvalg` options, you will need to have multiple strings to map to that diag, which means that `scorpio_output.cpp` will have to handle them, setting the proper values for the diag parameter list.

I'm assuming for now you're just interested in the default `temperature-inversion` case. Perhaps add a comment in the header file, explaining that the only supported case is `temperature-inversion` for now?

components/eamxx/src/diagnostics/pbl_entrainment_budget.cpp
bartgol: I have no clue what these variables represent, so I can't advise on this. But it's confusing to have code that is _guaranteed_ to never be executed. I would pick _some_ name for those diagnostics, and just do the mods to support them all. Alternatively, remove all other implementations, and hard-code the version for "temperature-inversion", without giving the idea that this is a configurable option...

bartgol: Ugh, if there's no other way, then ok. But I would prefer "Temperature" or "T", as you suggested. The name length is already super long, so no point in being cheap (and cryptic) with the last word...

PR #2824, opened 2024-05-15
This commit addresses two fixes in Corrective-ML

AaronDonahue:
1. Fixes precipitation adjustment in Corrective-ML, and it should now do the adjustment on Device when possible.
2. Fixes the interface to the TQ trained model to include phis and lat.


bartgol (commented):
I'll let someone else take a look, since I don't know anything about the actual ml code. The c++ interface changes look fine though.

frodre (approved):
Thanks, LGTM

PR #2827, opened 2024-05-16
Update ekat submodule

bartgol:
Update the ekat submodule to the commit just before the one that updates to kokkos 4.2. This takes care of some pervasive changes related to ekat's Units, so that we don't need those changes when we do the upstream merge.

The main change is that `Units` no longer offers `set_string`. The good news is that any arithmetic expression, such as `kg/kg` will correspond verbatim to the stored string (possibly with parentheses, for legibility). If one still wants a peculiar string (suche as "m2" for squared meters), they can pass it _at consturction time_, like `Units m2 (m*m,"m2");`. That string will then be used in downstream expressions, so that `auto Wm2 = W/m2;` will automatically display as "W/m2".

This takes away the need of the almost ubiquitous snippet
```
auto Q = kg/kg;
Q.set_string("kg/kg");
```
and we can simply just use `kg/kg` directly.

Note: we can still get a string in the fundamental SI units:
```
auto Wm2 = W/m2;
std::cout << Wm2.to_string() << "\n"; // prints "W/m2"
std::cout << Wm2.get_si_string() << "\n"; // prints something like kg^a m^b s^c
```

### DIFF
--- a/components/eamxx/src/control/atmosphere_surface_coupling_exporter.cpp
+++ b/components/eamxx/src/control/atmosphere_surface_coupling_exporter.cpp
@@ -25,15 +25,8 @@ void SurfaceCouplingExporter::set_grids(const std::shared_ptr<const GridsManager
   m_num_cols = m_grid->get_num_local_dofs();       // Number of columns on this rank
   m_num_levs = m_grid->get_num_vertical_levels();  // Number of levels per column
 
-  const auto m2 = m*m;
-  const auto s2 = s*s;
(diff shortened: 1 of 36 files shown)


bartgol:
The failures are due to the HDF error on weaver. Those tests should not be impacted by this PR anyways, so I'm leaning toward merging. I will go ahead and fix the files on weaver anyways.

tcclevenger (approved):

PR #2830, opened 2024-05-16
EAMxx: enable kokkos bounds checks in full_debug standalone build

### LINKED ISSUE #2828: are we not checking OOB?
This bug was causing an elusive (to Naser) segfault. The test was passing on pm-gpu just fine, but failed on other machines.

Are we not checking OOB as rigorously as we could?

              ```suggestion
        const int icol   = idx / nlevs;
```

ht @bartgol

_Originally posted by @mahf708 in https://github.com/E3SM-Project/scream/pull/2822#discussion_r1603644915_

### PR DESCRIPTION
**Pull request by bartgol**
Fixes #2828

### DIFF
--- a/components/eamxx/scripts/test_factory.py
+++ b/components/eamxx/scripts/test_factory.py
@@ -81,15 +81,13 @@ class TestProperty(object):
 class DBG(TestProperty):
 ###############################################################################
 
-    CMAKE_ARGS = [("CMAKE_BUILD_TYPE", "Debug"), ("EKAT_DEFAULT_BFB", "True"),
-                  ("Kokkos_ENABLE_DEBUG_BOUNDS_CHECK", "True")]
-
     def __init__(self, _):
         TestProperty.__init__(
             self,
             "full_debug",
             "debug",
-            self.CMAKE_ARGS,
+            [("CMAKE_BUILD_TYPE", "Debug"), ("EKAT_DEFAULT_BFB", "True"),
+             ("Kokkos_ENABLE_DEBUG_BOUNDS_CHECK", "True")]
         )
 
 ###############################################################################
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: components/eamxx/scripts/test_factory.py [hunk 1]
```diff
@@ -81,7 +81,8 @@ def __str__(self):
 class DBG(TestProperty):
 ###############################################################################
 
-    CMAKE_ARGS = [("CMAKE_BUILD_TYPE", "Debug"), ("EKAT_DEFAULT_BFB", "True")]
+    CMAKE_ARGS = [("CMAKE_BUILD_TYPE", "Debug"), ("EKAT_DEFAULT_BFB", "True"),
```
jgfouca, line 84: Several of the other TestProperty subclasses use the DBG.CMAKE_ARGS list.

SP: `DBG.CMAKE_ARGS + [("SCREAM_DOUBLE_PRECISION", "False")],`
FPE: `DBG.CMAKE_ARGS + [("SCREAM_PACK_SIZE", "1"), ("SCREAM_FPE","True")]`

So this turns on bounds checking for DBG, SP, and FPE. Is that the intent?

bartgol: Ah, no I didn't want to do that. Although, should we? I def don't want it on in the valg build, so I need to edit that. I would argue that bounds can be checked in just one build, since access patterns do not depend on precision or FPE handling... I will make sure other builds don't inherit that.

jgfouca: Probably not. You can just explicitly disable it in the SP and FPE tests.

jgfouca: VALG does not use DBG.CMAKE_ARGS, so we are fine there.

bartgol:
This PR brought up an issue in EKAT. I submitted a PR there, but we then need to wait for an upstream merge, since ekat will have kokkos 4.2 in it now. So this PR is on hold until E3SM-Project/E3SM#6421 is merged, we do an upstream merge, E3SM-Project/EKAT#339 is merged, and we update EKAT in the scream repo.

jgfouca (approved):

PR #2834, opened 2024-05-17
Add pgrad_correction to caar_ut

trey-ornl:
Add an outer loop over values of `pgrad_correction` (0/false and 1/true) to `caar_ut` (the CAAR Unit Test). The existing `caar_ut` uses the value `pgrad_correction=0`, while most Scream runs now use `pgrad_correction=1`. The logical changes are minor, but the addition of the outer loop to `caar_ut.cpp` makes the diff a mess.

Update the system software versions for compiler `crayclang-scream`. These changes may break other builds.

Add a machine file for BFB tests on Frontier, `frontier-bfb`.

Make one-line minor changes to other files to make the Frontier build happy.

I confirmed that the updated `caar_ut` passes on Frontier, and I confirmed that the test does indeed fail if `pgrad_correction` is set differently for Fortran and for C++, for both true/false and false/true. The updated test sets `pgrad_correction` to 0 and 1, thus doubling the number of calls to `CaarFunctor`. We can change the loop in `caar_ut.cpp` to run just one of the values (line 223).

I need `caar_ut` to test `pgrad_correction=1` to make sure that my pending GPU optimizations are correct for this important use case.


trey-ornl:
Will do. And I'll drop the changes to `config_machines.xml` that may be controversial.

PR #2835, opened 2024-05-20
Upstream merge 2024-05-20

bartgol:
This PR brings in kokkos 4.2 in EAMxx.

No conflicts (except git choking a bit with the ekat submodule, so had to update it manually).


jgfouca (dismissed):

jgfouca (approved):

AaronDonahue:
@bartgol is there a reason we haven't merged this yet?

AaronDonahue:
Do we know why this is failing the AT?

bartgol:
> Do we know why this is failing the AT?

Lots of stuff happened in the last week, so I rewrote the branch, to be up to date with master, and replayed the upstream merge on top.

jgfouca (dismissed):

jgfouca:
@bartgol , I'm trying to move this along. I'm not worried about NMLDIFFs on mappy CIME cases. I see a homme_shoc_cld_p3_rrtmgp_baseline_cmp fails on weaver and mappy, which is likely a diff for this test. Is that expected?

bartgol:
Thanks for helping Jim!

I looked at the baseline cmp test, and the failure is odd. CPRNC reports ` NetCDF: Index exceeds dimension bound`. I think that means the number of time slices is different, and I don't know why that would be the case. Perhaps the baselines are corrupted. I will quickly investigate. If that's the case, we can merge.

bartgol:
Looking at the console output, I see that the time dim does indeed differ: `Dimension time differs            2  /=            1`. The first file is the one generated by the PR, while the second is the baseline. I don't know why the PR should do something different from the baselines, but this console snapshot suggests we may need to correct the output yaml file for the test:

```
Attribute max_snapshots_per_file from file1:          744  does not match that found on file2            1
```

bartgol:
@jgfouca A quick inspection showed that the upstream merge somehow brought in an old version of an input yaml file, causing the fail. I fixed it. Since the upstream merge was already a bit old (~20days), I wend ahead and redid the merge from scratch, to get the newest e3sm master. I don't expect diffs.

bartgol:
@jgfouca I think we're good to merge, what do you think?

PR #2837, opened 2024-05-20
Fix and cleanup vertical layer diagnostic

bartgol:
A few mods were needed, since there was an inconsistency in the handling of the surface value, as well as on how this diagnostic was used in the FieldAtHeight diagnostic. Summary:

* Fix usage of surface value (the b.c. for the vertical integral)
* Added "height" as a version of "z" but from surface
* Reworked the unit tests, so that they are simpler, and they do not just copy+paste the implementation of the class. In particular, no need to ensure `calculate_dz` works correctly, since it's already tested elsewhere. We just need to test that the diag computes the correct diagnostic type, and has the right behavior w.r.t. level index.

Also, one thing we should do in _all_ the diagnostics: we are not free to request a pack size. By the time diags are created, all fields have been allocated, so the diag does not have a saying in pack size. Instead, the diag should defer the creation of the output field until all inputs are set (i.e.., do it inside `initialize_impl`), and pick as a pack size for the output whatever can work with all its inputs (i.e., the smallest among the pack sizes of the inputs).

### DIFF
--- a/components/eamxx/src/diagnostics/field_at_height.cpp
+++ b/components/eamxx/src/diagnostics/field_at_height.cpp
@@ -44,7 +44,7 @@ FieldAtHeight (const ekat::Comm& comm, const ekat::ParameterList& params)
       " -        field name: " + m_field_name + "\n"
       " - surface reference: " + surf_ref + "\n"
       " -     valid options: sealevel, surface\n");
-  m_z_name = (surf_ref == "sealevel") ? "z" : "geopotential";
+  m_z_name = (surf_ref == "sealevel") ? "z" : "altitude";
   const auto& location = m_params.get<std::string>("vertical_location");
   auto chars_start = location.find_first_not_of("0123456789.");
   EKAT_REQUIRE_MSG (chars_start!=0 && chars_start!=std::string::npos,
--- a/components/eamxx/src/diagnostics/field_at_height.cpp
+++ b/components/eamxx/src/diagnostics/field_at_height.cpp
@@ -44,7 +44,7 @@ FieldAtHeight (const ekat::Comm& comm, const ekat::ParameterList& params)
       " -        field name: " + m_field_name + "\n"
       " - surface reference: " + surf_ref + "\n"
       " -     valid options: sealevel, surface\n");
-  m_z_name = (surf_ref == "sealevel") ? "z" : "altitude";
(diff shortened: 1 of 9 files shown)


components/eamxx/src/share/io/scream_io_utils.hpp
```diff
@@ -80,9 +80,9 @@ struct LongNames {
   std::map<std::string,std::string> name_2_longname = {
```
AaronDonahue, line 80: Would it make sense to do away with this approach to long names altogether and just use the same syntax for `hya`, `hyb`, `lev`, etc. as we do for the diagnostics?  I.e. add longname as metadata?

AaronDonahue (commented):
Thank you for putting this fix in so quickly, and improving the functionality quite a bit.  I had a few comments, but for the most part I think this is nearly ready to go in.

components/eamxx/src/diagnostics/field_at_height.cpp
```diff
@@ -44,7 +44,7 @@ FieldAtHeight (const ekat::Comm& comm, const ekat::ParameterList& params)
       " -        field name: " + m_field_name + "\n"
       " - surface reference: " + surf_ref + "\n"
       " -     valid options: sealevel, surface\n");
-  m_z_name = (surf_ref == "sealevel") ? "z" : "geopotential";
+  m_z_name = (surf_ref == "sealevel") ? "z" : "altitude";
```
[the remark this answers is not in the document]
bartgol: ah right, i forgot to fix this entry

components/eamxx/src/share/io/scream_io_utils.hpp
bartgol: Probably. But we can do it in a separate PR.

bartgol:
@AaronDonahue I fixed the only issue left, the p3_standalone test. The baseline_cmp tests are expected to fail, since we fixed the bug in the vertical integral. If on the next AT run the only fails are the baseline_cmp tests, we're good to merge if you don't have any more comments.

AaronDonahue (approved):
@bartgol , once you confirm the fails in the AT are the expected ones go ahead and manually merge.

bartgol:
The failures are the expected ones. Merging.

PR #2840, opened 2024-05-21
Allow adding geo fields to grid objects, even if they are const

bartgol:
This can be used by atm procs to add fields corresponding to the values of dimensions peculiar to vars of that process. For instance, in this PR I made RRTMGP load the sw/lw bands bounds and save them in the physics grid, so that they can later be saved to output file by adding `save_grid_data: true` to the output yaml file (it actually is true by default, so if you don't see this param, data will be saved)

As a follow up PR, we could make RRTMGP _not_ hardcode the values for `m_nswbands` and `m_nlwbands`, but rather read them from file. I understand that other atm procs may need this info, so at some point we may add a singleton-like data structure in `physics/share`, which upon construction will load this info from file. First atm proc to be created will init the singleton, while the ones after that will simply read the data from it.

I also patched up some use cases of LayoutType that were not really correctly supported.


brhillman (approved):
This looks good to me, and appears to work as expected for a test case.

PR #2846, opened 2024-05-23
Fix valgrind errors

bartgol:
Only one error was truly our fault, while most of the fails are due to some error reported inside pnetcdf, which appears to be not our responsibility. I added an entry in mappy's suppressions file for valgrind, so that we can ignore it.

### DIFF
--- a/components/eamxx/src/share/io/scream_scorpio_interface.cpp
+++ b/components/eamxx/src/share/io/scream_scorpio_interface.cpp
@@ -1378,12 +1378,50 @@ T get_attribute (const std::string& filename,
   if (varname=="GLOBAL") {
     varid = PIO_GLOBAL;
   } else {
-    varid = impl::get_var(filename,varname,"scorpio::set_any_attribute").ncid;
+    varid = impl::get_var(filename,varname,"scorpio::get_attribute").ncid;
   }
 
+  // If the attribute type does not match T, we need a temporary, since we can't pass T* where pio expects
+  // a different type of pointer
+  int att_type, err;
+  err = PIOc_inq_atttype(pf.file->ncid,varid,attname.c_str(),&att_type);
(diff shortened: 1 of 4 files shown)


components/eamxx/src/share/io/scream_scorpio_interface.cpp
```diff
@@ -1378,12 +1378,50 @@ T get_attribute (const std::string& filename,
   if (varname=="GLOBAL") {
     varid = PIO_GLOBAL;
   } else {
-    varid = impl::get_var(filename,varname,"scorpio::set_any_attribute").ncid;
+    varid = impl::get_var(filename,varname,"scorpio::get_attribute").ncid;
   }
 
+  // If the attribute type does not match T, we need a temporary, since we can't pass T* where pio expects
```
bartgol, line 1384: This was the real bug. Basically, when the var has float dtype, also the FillValue has float dtype, but we were trying to access it passing a double pointer. When PIO does a reinterpret_cast of the input pointer to a float* and writes to it, it only writes the first 32bits (and the meaning of those 32 bits is not the same as in the full double), leaving some bits uninitialized. In fact, reading the fillvalue gave a completely random number (like 1e-300), rather than the expeted 1e30 (or whatever we had in the file).

components/eamxx/src/share/io/tests/scorpio_interface_tests.cpp
```diff
@@ -20,18 +20,14 @@ TEST_CASE ("io_subsystem") {
 TEST_CASE ("write_and_read") {
   ekat::Comm comm (MPI_COMM_WORLD);
 
-  EKAT_REQUIRE_MSG (comm.size()<=4,
```
bartgol, line 23: These mods are just to allow running this test with an arbitrary number of ranks (rather than hardcoding 1-4).

jgfouca (approved):

PR #2847, opened 2024-05-23
SHOC interface bug fix - TKE clipping

bogensch:
In the v1 SHOC interface the TKE minimum allowable value was being clipped an order of magnitude higher than intended (0.004 vs the 0.0004 value in v0).  This value has been modified to use the SHOC constant "mintke".  

This was discovered while validating DPxx where I noticed that turbulent mixing in the free troposphere was an order of magnitude higher than DPv0.  

Since this bug was in the SHOC interface I anticipate that all SHOC tests will be b4b, but global simulations will not be b4b.  However, I do not anticipate a large impact from this bug fix since turbulent mixing in the free troposphere is quite small.


tcclevenger (approved):

tcclevenger:
It looks like all our CIME cases are diffing and only one of the unit tests. I looked at a couple of the CIME cases and the errors do seem relatively small (normalized errors between 1e-5 and 2e0 for most fields). Let me know if you are ready to merge.

bogensch:
Thanks @tcclevenger for investigating.  Yes, that sounds like what I would expect and ready to merge.

PR #2851, opened 2024-05-29
Initial implementation of pyscream module(s) for running eamxx from python

bartgol:
The desire is to run a single parametrization from a python script. This requires a few hooks, that this PR tries to put together. Work is still ongoing, and the PR may be rebased/reset/closed/reopen.

The idea is to create a separate python module for each scream library:
- pyscream: contains all infrastructure hooks. Currently it has wrappers for
  - ekat's Units
  - scream Field
  - scream Grid
- pyp3: contains a wrapper of P3Microphysics.

The wrappers do NOT expose all the methods of the wrapped class, nor they are built in the same way as the wrapped class. Instead, we expose the minimum necessary to be able to create/setup/use scream objects.

If you want to try building this, you need to
- pip install mpi4py pybind11
- run test-all-scream with `-c EAMXX_ENABLE_PYBIND=ON -c BUILD_SHARED_LIBS=ON`

Lots of stuff to do/decide
- Add method to extract numpy array from PyField
- Add wrappers for timestamp (needed for init)
- Add wrappers for reading fields from IC files
- Add hooks to (re)set parameters inside parametrizations. What if this requires to re-run initialize? Can we re-init an atm proc? I don't think so. Alternative: re-create the atm proc every time.
- Run init/finalize of eamxx session when pyscream is imported/unloaded (the latter may be done with atexit)
- Allow to set geometry data in the grid? E.g., lat/lon/area?
- Do units really matter? Can we set in/out fields with bad units? Do atm procs notice it or do they just check the field name/layout?
- Other?


bartgol:
Here's an example of how it can be used (still crashes, but gives the idea):

```python
#!/usr/bin/env python3

import sys

sys.path.append('/home/lbertag/workdir/scream/scream-src/master/components/eamxx/ctest-build/full_debug/src/share/python')
sys.path.append('/home/lbertag/workdir/scream/scream-src/master/components/eamxx/ctest-build/full_debug/src/physics/p3/python')

#  import mpi4py.rc
#  mpi4py.rc.finalize = True
#  mpi4py.rc.threads = False
#  mpi4py.rc.thread_level = "single"

from mpi4py import MPI
import pyscream as ps
import pyp3 as P3
from pyscream.units import m,s,kg,K,N,J,W,Pa,one

#######################################
def run ():
#########################################

    # Create a comm, and print specs
    comm = MPI.COMM_WORLD
    print (f"hello from rank{comm.Get_rank()}/{comm.Get_size()}")

    # Create a physics-like grid
    ncols = 100
    nlevs = 128
    grid = ps.Grid("Physics",ncols,nlevs)

    # Check that units operators overloads work
    m2 = m*m
    Wm2 = W / m2
    print (f"Wm2: {Wm2.str()}")

    # Create a bunch of (managed) fields
    T_mid = grid.scalar3d_mid("T_mid",K)
    qv = grid.scalar3d_mid("qv",kg/kg)
    qc = grid.scalar3d_mid("qc",kg/kg)
    qi = grid.scalar3d_mid("qi",kg/kg)
    qr = grid.scalar3d_mid("qr",kg/kg)
    qm = grid.scalar3d_mid("qm",kg/kg)
    nc = grid.scalar3d_mid("nc",one/kg)
    nr = grid.scalar3d_mid("nr",one/kg)
    ni = grid.scalar3d_mid("ni",one/kg)
    bm = grid.scalar3d_mid("bm",one/kg)
    precip_l = grid.scalar2d("precip_liq_surf_mass",kg)
    precip_i = grid.scalar2d("precip_ice_surf_mass",kg)
    eff_r_qc = grid.scalar3d_mid("eff_radius_qc",m)
    eff_r_qi = grid.scalar3d_mid("eff_radius_qi",m)
    eff_r_qr = grid.scalar3d_mid("eff_radius_qr",m)
    p_mid = grid.scalar3d_mid("p_mid",m)
    p_dry_mid = grid.scalar3d_mid("p_dry_mid",m)
    rho = grid.scalar3d_mid("pseudo_density",m)
    rho_dry = grid.scalar3d_mid("pseudo_density_dry",m)
    cldfrac_tot = grid.scalar3d_mid("cldfrac_tot",m)
    qv_prev = grid.scalar3d_mid("qv_prev_micro_step",m)
    T_prev = grid.scalar3d_mid("T_prev_micro_step",m)
    rain_frac = grid.scalar3d_mid("rainfrac",m)
    nc_nuceat_tend = grid.scalar3d_mid("nc_nuceat_tend",m)
    nccn = grid.scalar3d_mid("nccn",m)
    ni_activated = grid.scalar3d_mid("ni_activated",m)
    inv_qc_relvar = grid.scalar3d_mid("inv_qc_relvar",m)
    liq_ice_exchange = grid.scalar3d_mid("micro_liq_ice_exchange",m)
    vap_liq_exchange = grid.scalar3d_mid("micro_vap_liq_exchange",m)
    vap_ice_exchange = grid.scalar3d_mid("micro_vap_ice_exchange",m)

    # Create an init p3
    p3 = P3.P3(grid)
    p3.set_fields([qv,T_mid,qc,qi,qr,qm,nc,nr,ni,bm,precip_l,precip_i,eff_r_qc,eff_r_qi,eff_r_qr,p_mid,p_dry_mid,rho,rho_dry,cldfrac_tot,qv_prev,rain_frac,nc_nuceat_tend,nccn,ni_activated,inv_qc_relvar,T_prev,liq_ice_exchange,vap_liq_exchange,vap_ice_exchange])
    p3.initialize()

#######################################
def main ():
#######################################
    # This level of indirection ensures all pybind structs are destroyed
    # before we finalize eamxx (and hence kokkos)
    ps.init()
    run ()
    ps.finalize()

####################################
if  __name__  == "__main__":
    main()
```

PeterCaldwell:
quick thought - we should probably call the library pyeamxx or something like that since folks will be using it for low-res applications as well. Though pyeamxx sounds lame. py3sm? eampy?

bartgol:
Uhm, that may be a better approach. Do not let the user create fields, just access the ones they need. I'll work on that. It may also allow to trim down the pygrid file, since we don't need to expose those capabilities anymore.

PeterCaldwell:
Ignore if this is a dumb comment: we're using pybind11 for our AI work and I like the symmetry of using the same thing for running EAMxx from python and for running python from EAMxx.

bartgol:
- pybind: I think a good old `find_package(pybind11 REQUIRED)` is all we need. If a CMake var called `<packagename>_ROOT` exists, cmake already uses it to find the package.
- I don't know about the array specifics. What's bad with the stuff in numpy.h from pybind11?

bartgol:
I think this PR is getting to a point where we should try to integrate it. I added a unit test, that runs p3_standalone both in cxx (the usual AD test) and from python, and then compares the generated nc files for all ranks.

The cxx test is a re-run of p3_standalone. I thought about just using the existing one, but ctest does not allow to set dependencies on tests from other folders (the FIXTURE thingy does not work outside of the folder). I guess we could put the p3_standalone_py script in the folder `tests/single-process/p3`, and run it only if `EAMXX_ENABLE_PYBIND=ON`. If you think that's cleaner (maybe it is), I can do that.

I have not enabled the build of pyeamxx in any of the test-all-scream builds, so it isn't really tested (unless you add `-c EAMXX_ENABLE_PYBIND=ON -c BUILD_SHARED_LIBS`). I didn't for a couple of reasons

1. I had to manually hack the mam4xx and haero submodules, to allow building shared libs. We should wait for those repos to change their cmake logic, then update them in the scream repo.
2. I wasn't entirely sure if all this stuff would work on mappy, so I wanted to test there first (manually). And running on weaver may be harder still. I think it should work on GPU, but I can't be sure.

Also, I haven't really triggered anything interesting in the test, like modifying an input or reading an output from python. That may be the next step.

components/eamxx/tests/python/pyp3/p3_standalone_py
```diff
@@ -0,0 +1,64 @@
...
+import sys
+
+# Add path to scream libs
+sys.path.append('@SCREAM_BASE_DIR@/scripts')
+
+# Add path to pyeamxx libs
+sys.path.append('@CMAKE_BINARY_DIR@/src/share/python')
+sys.path.append('@CMAKE_BINARY_DIR@/src/physics/p3/python')
+
+# Without these, and manual init/finalize, on my laptop I get
```
bartgol, line 12: If/when we move MPI/eamxx init/final in the module internals, this may become moot.

bartgol: Btw, I think [this approach](https://github.com/sandialabs/Albany/blob/master/pyAlbany/python/__init__.py) is quite neat. The temporary is created at module level, so it is destroyed when the module is unloaded, and its destructor will take care of finalizing kokkos. We can do something similar with MPI/eamxx init/final.

.gitmodules
```diff
@@ -83,3 +83,6 @@
 	path = externals/mam4xx
 	url = git@github.com:eagles-project/mam4xx.git
 
+[submodule "externals/pybind11"]
```
bartgol, line 86: @mahf708 I don't think we should add pip-installable modules as git submodules. Users should be in charge of verifying that dependencies are in place. For standalone testing, we can pip install missing modules locally on the fly.

components/eamxx/src/python/pyproject.toml
```diff
@@ -0,0 +1,17 @@
...
+dependencies = ["numpy", "mpi4py"]
+
+[tool.setuptools.packages.find]
+where = ["."]
+include = ["pyeamxx", "libpyeamxx"]
+exclude = ["build_src", "tests"]
+namespaces = true
+
+[tool.setuptools.package-data]
+"libpyeamxx" = ["*.so*"]
```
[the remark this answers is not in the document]
bartgol: Why can't we put everything in `pyeamxx`? Why do we need `libpyeamxx`?

components/eamxx/tests/python/pyp3/p3_standalone_py
```diff
@@ -0,0 +1,63 @@
...
+    ncols = 218
+    nlevs = 72
+    grid = ps.Grid("Physics",ncols,nlevs)
+
+    ic_file = Path('@SCREAM_DATA_DIR@/init/screami_unit_tests_ne2np4L72_20220822.nc')
+
+    p3params = ps.ParameterList(yaml_input['atmosphere_processes']['p3'])
+    p3 = pyp3.P3(grid,p3params)
+    missing = p3.read_ic(str(ic_file))
+    print (f"WARNING! The following input fields were not found in the IC file, and must be manually initialized: {missing}")
```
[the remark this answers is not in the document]
bartgol: Manual initialization is done simply in python, no cpp knowledge. If field `X` is not in the IC file, the user can do
```
X = p3.get_field('X')
np.deep_copy(X,0)
```
or whatever is the way we assign 0 to all entries of a np nd-array.

bartgol:
@mahf708 I was puzzled an confused by the two modules (pyeamxx and libpyeamxx_ext), so I renamed the second to pyeamxx. My guess is that the two clash when you try to package it as a conda package? I'm not sure. We can discuss when you get back.

Meanwhile, pyeamxx can now be used with any of the physics parametrizations. The only limitation is that the user must create a grid for which we have an IC file. We can work on adding some logic (in eamxx or pyeamxx) to allow reading an existing IC file (with N cols) into a grid with 1 col, perhaps maybe recycling IOP implementation: specify the lat/lon of the col, and reading in the closest col to that lat/lon.

When Naser returns, we can iterate on the pyeamxx conda package impl, which I'm confident I broke. Meanwhile, pyeamxx is available as follows:
```bash
$ cd components/eamxx
$ ./scripts/test-all-scream -t opt -m MACHINE --config-only
$ cd ctest-build/release/src/python
$ make -j
```
And then in your python script, do something like
```python
import sys
sys.path.append('/path/to/eamxx/ctest-build/full_debug/src/python')
sys.path.append('/path/to/eamxx/scripts')

import mpi4py
mpi4py.rc.initialize = False  # do not initialize MPI automatically
mpi4py.rc.finalize = False    # do not finalize MPI automatically

from mpi4py import MPI
import pyeamxx
from pathlib import Path
from utils import ensure_yaml
ensure_yaml()
import yaml

def run():
    # Get timestepping params
    with open('input.yaml','r') as fd:
        yaml_input = yaml.load(fd,Loader=yaml.SafeLoader)
    nsteps = yaml_input['time_stepping']['number_of_steps']
    dt     = yaml_input['time_stepping']['time_step']
    t0_str = yaml_input['time_stepping']['run_t0']

    # Create the grid
    ncols = 218
    nlevs = 72
    grid = pyeamxx.Grid("Physics",ncols,nlevs)

    ic_file = Path('/home/lbertag/workdir/e3sm/e3sm-data/inputdata/atm/scream/init/screami_unit_tests_ne2np4L72_20220822.nc')

    rrtmgp_params = pyeamxx.ParameterList(yaml_input['atmosphere_processes']['rrtmgp'],'rrtmgp')
    rrtmgp = pyeamxx.AtmProc(rrtmgp_params,grid)
    missing = rrtmgp.read_ic(str(ic_file))
    print (f"WARNING! The following input fields were not found in the IC file, and must be manually initialized: {missing}")
    rrtmgpinitialize(t0_str)
    rrtmgp.setup_output("my_output_rrtmgp.yaml")

    # Time looop
    for n in range(0,nsteps):
        rrtmgp.run(dt)

if  __name__  == "__main__":
    # This level of indirection ensures all pybind structs are destroyed
    # before we finalize eamxx (and hence kokkos)
    MPI.Init()
    pyeamxx.init()
    main ()
    pyeamxx.finalize()
    MPI.Finalize()
```

The above assumes that `input.yaml` is structured similarly to our standalone input tests. Of course, you are welcome to change that, and then also change the python example above accordingly.

PR #2853, opened 2024-05-29
Remove unused interface for ZM deep convection

AaronDonahue:
The ZM interface we currently have is out-of-date and no longer used. If we decide to start using ZM again it is likely that we need to rewrite the interface to reflect the current paradigm.  As a result, this commit removes the old interface to avoid confusion.


AaronDonahue:
@bartgol , this PR just removes the unused directory for ZM that we created a few years back.  So hopefully it is pretty simple to review and approve.

bartgol (approved):

PR #2856, opened 2024-05-30
Move IOP from control/ to share/iop

tcclevenger:
Since IOP only depends on `src/share` code, move from `src/control/` to `src/share/` to remove spaghetti connections causing linking errors when using shared libs. 

I was between `src/share/atm` and `src/share/iop`.  I chose to have it's own folder since not everything in IOP is directly used in an atmosphere process (e.g., reading ICs is all done in the driver, but I guess you could argue that is for the processes). I also think future additions to IOP might not be directly related to atm processes (IOP specific output and/or diagnostics related to IOP). I'd be fine either way though if there are strong opinions.


bartgol (approved):
Thanks. This seems to make shared libs build fine. I have no idea why static worked and shared didn't... Alas, a fix is a fix.

PR #2857, opened 2024-06-03
generalize table reading in p3_iso_c

mahf708:
Two issues that appeared in cases where the produced shared object was patched to make the hard links softer.
1. The length of the paths was more than 256 characters
2. Somehow the `write` logic in the append_precision wasn't working, so fixing that with a more hardcoded char concat

In both cases, the issues weren't reproducible in supported machines or supported builds (the issue is essentially when the shared objects are built on one machine and used on another, hence the need to change the hard links to softer links, so that the paths can be different on different machines)

### DIFF
--- a/components/eamxx/src/physics/p3/p3_iso_c.f90
+++ b/components/eamxx/src/physics/p3/p3_iso_c.f90
@@ -70,15 +70,15 @@ contains
     ok = .false.
 
 #ifdef SCREAM_DOUBLE_PRECISION
-    mu_r_filename  = lookup_file_dir//'/mu_r_table_vals.dat8'//C_NULL_CHAR
-    revap_filename = lookup_file_dir//'/revap_table_vals.dat8'//C_NULL_CHAR
-    vn_filename    = lookup_file_dir//'/vn_table_vals.dat8'//C_NULL_CHAR
-    vm_filename    = lookup_file_dir//'/vm_table_vals.dat8'//C_NULL_CHAR
+    mu_r_filename  = lookup_file_dir(1:len)//'/mu_r_table_vals.dat8'//C_NULL_CHAR
+    revap_filename = lookup_file_dir(1:len)//'/revap_table_vals.dat8'//C_NULL_CHAR
+    vn_filename    = lookup_file_dir(1:len)//'/vn_table_vals.dat8'//C_NULL_CHAR
+    vm_filename    = lookup_file_dir(1:len)//'/vm_table_vals.dat8'//C_NULL_CHAR
 #else
-    mu_r_filename  = lookup_file_dir//'/mu_r_table_vals.dat4'//C_NULL_CHAR
-    revap_filename = lookup_file_dir//'/revap_table_vals.dat4'//C_NULL_CHAR
-    vn_filename    = lookup_file_dir//'/vn_table_vals.dat4'//C_NULL_CHAR
-    vm_filename    = lookup_file_dir//'/vm_table_vals.dat4'//C_NULL_CHAR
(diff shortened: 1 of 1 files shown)


components/eamxx/src/physics/p3/p3_iso_c.f90
```diff
@@ -16,8 +16,8 @@ module p3_iso_c
 contains
   subroutine append_precision(string, prefix)
 
-    character(kind=c_char, len=256), intent(inout) :: string
-    character(*), intent(in) :: prefix
+    character(kind=c_char, len=256), intent(out) :: string
+    character(kind=c_char, len=*), intent(in) :: prefix
```
ambrad, line 20: Technically, `prefix` is not `c_char`; only `string` is. In practice it won't matter. But I suggest not specifying it.

ambrad (dismissed):

ambrad:
Looks fine. Definitely let the AT run all its tests before merging.

ambrad (approved):
For the record, note that the change from calling append_precision to hardcoded string concatenation is because something in the old append_precision code path in the container environment Naser is working on is not working. A small reproducer does not have the problem, so something subtle is going wrong.

bartgol:
@jgfouca @tcclevenger  @AaronDonahue  and other p3 devs: It seems to me that we read p3 tables at init, but we never use them in the cxx code. Is that right? I see them explicitly computed in `init_kokkos_ice_lookup_tables` and `init_kokkos_tables`. It seems to me that reading the f90 tables only matters for the f90 code, which in turns only matters for unit testing.

Maybe it's time we separate the f90 stuff in p3 that is needed for unit tests only, and put it somewhere else (maybe in p3/tests). Or, at the very least, avoid doing this step in CIME runs, since it's pointless?

jgfouca:
P3 definitely needs those tables. I think we decided it was simpler to read them in fortran and then pass the raw data over to the cxx side, but that data is used heavily all over p3. Just do a grep for `table` under `eamxx/src/physics/p3/impl` and you'll seem them used all over.

bartgol:
I know we use the tables, but it looks like we are _computing them_ in [this function](https://github.com/E3SM-Project/scream/blob/3a0cee9be388187492bd9447d35f34d17bd44b94/components/eamxx/src/physics/p3/impl/p3_table3_impl.hpp#L87) and [this function](https://github.com/E3SM-Project/scream/blob/3a0cee9be388187492bd9447d35f34d17bd44b94/components/eamxx/src/physics/p3/impl/p3_table_ice_impl.hpp#L18), which are called during P3's [initialize_impl](https://github.com/E3SM-Project/scream/blob/3a0cee9be388187492bd9447d35f34d17bd44b94/components/eamxx/src/physics/p3/eamxx_p3_process_interface.cpp#L343). All of this makes me wonder if P3 itself needs to load those files in a CIME run...

jgfouca:
OK, my memory is starting to jog. I remember now that I did port the table reads from f90 to cxx. The second function you linked does file I/O, so the tables are still being read in cxx just like in f90 unless I misunderstood what you are saying.

bartgol:
Ah, yes, I completely missed the ifstream part in that file.

jgfouca:
@PeterCaldwell , it's been so long that my memory is foggy, but I am pretty sure I remember that some of these tables were quite expensive to create which is why we changed them to file reads.

bartgol:
> @PeterCaldwell , it's been so long that my memory is foggy, but I am pretty sure I remember that some of these tables were quite expensive (over 10 seconds) to create which is why we changed them to file reads.

I wonder if they were expensive to build b/c we built them on host in F90 (I assume that's what we did). Maybe if we built them on device things would be better? Although, this would only matter for a GPU build...

PR #2859, opened 2024-06-04
Fix valgrind errors in eamxx_homme_iop.cpp

jgfouca:
We were running off the end of a couple views.

Example from my debugging printing:
`JGF lev=79 hyam.size=72, hybm.size=72`

### DIFF
--- a/components/eamxx/src/dynamics/homme/eamxx_homme_iop.cpp
+++ b/components/eamxx/src/dynamics/homme/eamxx_homme_iop.cpp
@@ -594,9 +594,9 @@ apply_iop_forcing(const Real dt)
             // Restrict nudging of T and qv to certain levels if requested by user
             // IOP pressure variable is in unitis of [Pa], while iop_nudge_tq_low/high
             // is in units of [hPa], thus convert iop_nudge_tq_low/high
-            Mask nudge_level;
-            for (int p=0; p<Pack::n; ++p) {
-              const auto lev = k*Pack::n + p;
+            Mask nudge_level(false);
+            int max_size = hyam.size();
+            for (int lev=k*Pack::n, p = 0; p < Pack::n && lev < max_size; ++lev, ++p) {
               const auto pressure_from_iop = hyam(lev)*ps0 + hybm(lev)*ps_iop;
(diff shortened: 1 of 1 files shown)


components/eamxx/src/dynamics/homme/eamxx_homme_iop.cpp
```diff
@@ -594,9 +594,9 @@ apply_iop_forcing(const Real dt)
             // Restrict nudging of T and qv to certain levels if requested by user
             // IOP pressure variable is in unitis of [Pa], while iop_nudge_tq_low/high
             // is in units of [hPa], thus convert iop_nudge_tq_low/high
-            Mask nudge_level;
-            for (int p=0; p<Pack::n; ++p) {
-              const auto lev = k*Pack::n + p;
+            Mask nudge_level(false);
```
jgfouca, line 597: I assume I should default this mask to all false?

AaronDonahue (approved):

components/eamxx/src/dynamics/homme/eamxx_homme_iop.cpp
```diff
@@ -594,9 +594,9 @@ apply_iop_forcing(const Real dt)
...
+            int max_size = hyam.size();
+            for (int lev=k*Pack::n, p = 0; p < Pack::n && lev < max_size; ++lev, ++p) {
               const auto pressure_from_iop = hyam(lev)*ps0 + hybm(lev)*ps_iop;
               nudge_level.set(p, pressure_from_iop <= iop_nudge_tq_low*100
```
jgfouca, line 601: @bogensch , this looks like a bug. How can pressure ever be both low and high? This `and` should be an `or` I think.

bogensch: It's not a bug, just confusing names.....  Valid example values: iop_nudge_tq_low = 1000 hPa; iop_nudge_tq_high=700 hPa.  Thus low/high refers to the relative position in the atmosphere with respect to the surface rather than the value.

jgfouca: Thanks! Makes sense now.

PR #2860, opened 2024-06-06
[WIP] adds support st archive

mahf708:
for now to start a discussion only... need some actual testing with a small case

### DIFF
--- a/cime_config/config_archive.xml
+++ b/cime_config/config_archive.xml
@@ -29,7 +29,19 @@
     <rest_file_extension>r\.(INSTANT|AVERAGE|MAX|MIN)\.n(step|sec|min|hour|day|month|year)s_x\d*</rest_file_extension>
     <rest_file_extension>rhist\.(INSTANT|AVERAGE|MAX|MIN)\.n(step|sec|min|hour|day|month|year)s_x\d*</rest_file_extension>
     <!-- The following matches "hi.AVGTYPE.FREQUNITS_xFREQ.TIMESTAMP.nc"-->
+    <!-- what's the "hi" for? I don't think we follow that convention (maybe we should?) -->
     <hist_file_extension>hi\.(INSTANT|AVERAGE|MAX|MIN)\.n(step|sec|min|hour|day|month|year)s_x\d*\.\d{4}-\d{2}-\d{2}-\d{5}\.nc$</hist_file_extension>
+    <!-- The following will actually match all files produced by EAMxx IO, which may prove problematic; need some testing -->
+    <hist_file_extension>\.(INSTANT|AVERAGE|MAX|MIN)\.n(step|sec|min|hour|day|month|year)s_x\d*\.\d{4}-\d{2}-\d{2}-\d{5}\.nc$</hist_file_extension>
+    <!-- I think we likely need this: https://github.com/E3SM-Project/E3SM/issues/4615 -->
+    <rest_history_varname>nhfil</rest_history_varname>
+    <rpointer>
+      <rpointer_file>rpointer.atm$NINST_STRING</rpointer_file>
+      <!-- Currently a period and a slash exist in rpointer.atm in EAMxx cases -->
+      <rpointer_content>./$CASE.scream$NINST_STRING.r.$DATENAME.nc </rpointer_content>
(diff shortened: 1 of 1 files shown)


cime_config/config_archive.xml
```diff
@@ -29,7 +29,19 @@
     <rest_file_extension>r\.(INSTANT|AVERAGE|MAX|MIN)\.n(step|sec|min|hour|day|month|year)s_x\d*</rest_file_extension>
     <rest_file_extension>rhist\.(INSTANT|AVERAGE|MAX|MIN)\.n(step|sec|min|hour|day|month|year)s_x\d*</rest_file_extension>
     <!-- The following matches "hi.AVGTYPE.FREQUNITS_xFREQ.TIMESTAMP.nc"-->
+    <!-- what's the "hi" for? I don't think we follow that convention (maybe we should?) -->
     <hist_file_extension>hi\.(INSTANT|AVERAGE|MAX|MIN)\.n(step|sec|min|hour|day|month|year)s_x\d*\.\d{4}-\d{2}-\d{2}-\d{5}\.nc$</hist_file_extension>
```
PeterCaldwell, line 33: I think "hi" is mean to refer to E3SM's "h for history" followed by "i" for index, which can be h0, h1, etc. If I recall correctly, coupler history files literally use the name "hi" because there's only 1 of them. We specifically chose *not* to follow the "hi" formatting for EAMxx output names because it is cryptic and led to a limit on how many output files we could have. So I'm strongly opposed to hist file extension "hi" unless it refers to other components than EAMxx.

cime_config/config_archive.xml
```diff
@@ -29,7 +29,19 @@
...
+    <!-- The following will actually match all files produced by EAMxx IO, which may prove problematic; need some testing -->
+    <hist_file_extension>\.(INSTANT|AVERAGE|MAX|MIN)\.n(step|sec|min|hour|day|month|year)s_x\d*\.\d{4}-\d{2}-\d{2}-\d{5}\.nc$</hist_file_extension>
+    <!-- I think we likely need this: https://github.com/E3SM-Project/E3SM/issues/4615 -->
+    <rest_history_varname>nhfil</rest_history_varname>
+    <rpointer>
+      <rpointer_file>rpointer.atm$NINST_STRING</rpointer_file>
+      <!-- Currently a period and a slash exist in rpointer.atm in EAMxx cases -->
+      <rpointer_content>./$CASE.scream$NINST_STRING.r.$DATENAME.nc </rpointer_content>
```
PeterCaldwell, line 41: we should probably change the rpointers to be $CASE.eamxx rather than $CASE.scream :-(

cime_config/config_archive.xml
rljacob: I think the "i" for EAM is "initial".  Its a snapshot of the initial condition dumped at the start of a run.  But in the coupler "i" is instantaneous.   EAM outputs a file with "hi" in the name along with h0, h1, h2, etc.

PeterCaldwell: oh yeah, we called our scream initial conditions screami files because we thought it was cute.

bartgol: Here is where Jim's knowledge is important. But trying to remember, I think we had to add something, otherwise rest files were also picked up as hist files. I'm not 100% sure about that though, we should check (and ask Jim!).

bartgol:
I removed myself as a reviewer, as I don't know much about this stuff. I think Jim/Rob can give better feedback.

cime_config/config_archive.xml
jgfouca: Everything looks ok to me except maybe this second `hist_file_extension`. It looks exactly the same as the one above except without the leading `hi`, this means the files you want to match are identical to the ones it was already matching except with no `hi` in the file name. Is that what is intended here?

chengzhuzhang:
@mahf708 thank you for taking the initiative on this effort! 
Originally, I thought it is straightforward to extend st_archive through CIME to cover eamxx files. Based on your exploration, it requires changing eamxx output file names. In addition, there is potential performance issue. The slowness is mostly from copying large size files, right? I think folks have being using move (`--force-move` ), which is much faster.

Issue #2813, opened 2024-05-04
SHOC and HOMME use different definitions of vtheta


mahf708:
(This was inspired by a conversation I had with @oksanaguba about the physics--dynamics interface of HOMME and the assumptions made. In particular, I was interested in retrieving theta_l (liquid theta) for some diagnostics I am working on, and @oksanaguba warned me that SHOC may have it somewhere, but it may not be what I wanted. The definition of theta_v in the SHOC code is just below the definition of theta_l. The definition for theta_l was the same as the one I had in mind, but I noticed that SHOC uses a different definition for theta_v than the one used in HOMME.)

In SHOC (and also Wikipedia), 
$$\theta_{\text{v}} = \theta (1 + \beta q_{\text{v}} - q_{\text{c}}) $$

In HOMME,
$$\theta_{\text{v}} = \theta (1 + \gamma q_{\text{v}}) $$

However, both SHOC and HOMME *almost* converge on $\beta$ and $\gamma$. I calculated them to be:

$$ \beta = 0.6077948168566272 $$

<details><summary>SHOC</summary>
<p>

```ipython
>>> (((6.02214e26*1.38065e-23)/18.016)/287.042)-1 # SHOC
0.6077948168566272
```

</p>
</details> 

$$ \gamma = 0.6077930728241567 $$

<details><summary>HOMME</summary>
<p>

```ipython
>>> 1/(18.016/28.966)-1 # HOMME
0.6077930728241567
```

</p>
</details> 

So, HOMME ignores $q_{\text{c}}$ and SHOC (dis)counts it.

oksanaguba:
I always thought clubb and shoc use theta l (liquid), not virtual PT. Besides that There are other inconsistencies in thermodynamics of eam and homme. We can document them in more detail.

mt5555:
It looks to me like the Wikipedia/SHOC formula is assuming mixing ratios relative to total mass (dry + vapor + condensates).    So the Wikipedia formula is removing the q_c contribution when computing the dry + vapor mass used in the theta/theta_v relation.  

SCREAM and EAM mixing ratios are written with respect to a total mass that is only dry + vapor.   This suggests the HOMME formula is correct when written in terms of SCREAM mixing ratios.

PeterCaldwell:
I never thought I'd say - we should correct the Wikipedia article about us! ^^ @bogensch !

oksanaguba:
[...]

PR #2861, opened 2024-06-10
DPxx - Use consistent formulation of vtheta_dp

bogensch:
This PR updates the definition of vtheta_dp in the DP/HOMME interface to be consistent with that used by HOMME in v1.  Previously we were using the v0 formulation, which was causing a drift in long DPxx simulations due to the inconsistency.  

PR should only be answer changing for DP simulations; otherwise B4B.


components/eamxx/src/dynamics/homme/eamxx_homme_iop.cpp
```diff
@@ -333,67 +338,60 @@ apply_iop_forcing(const Real dt)
...
+    Kokkos::parallel_for("compute_temperature_for_iop", policy_homme, KOKKOS_LAMBDA (const KT::MemberType& team) {
       KV kv(team);
       const int ie  =  team.league_rank();
 
       // Get temp views from workspace
-      auto ws = homme_wsm.get_workspace(team);
-      auto pnh_slot   = ws.take_macro_block("pnh"  , NGP*NGP);
-      uview_2d<Pack> pnh(reinterpret_cast<Pack*>(pnh_slot.data()), NGP*NGP, NLEV);
+      auto ws = eamxx_wsm.get_workspace(team);
+      uview_1d<Pack> pmid, pint, pdel;
```
tcclevenger, line 352: We can remove `pint` and `pdel` here, we are computing these values like above.

components/eamxx/src/dynamics/homme/eamxx_homme_iop.cpp
```diff
@@ -333,67 +338,60 @@ apply_iop_forcing(const Real dt)
...
+
         // Reinterperate into views of Homme::Scalar for calling Hommexx function.
         Homme::ExecViewUnmanaged<Homme::Scalar[NLEV]> dp3d_scalar(reinterpret_cast<Homme::Scalar*>(dp3d_i.data()), NLEV);
         Homme::ExecViewUnmanaged<Homme::Scalar[NLEV]> vtheta_dp_scalar(reinterpret_cast<Homme::Scalar*>(vtheta_dp_i.data()), NLEV);
-        Homme::ExecViewUnmanaged<Homme::Scalar[NLEVI]> phi_int_scalar(reinterpret_cast<Homme::Scalar*>(phi_int_i.data()), NLEVI);
         Homme::ExecViewUnmanaged<Homme::Scalar[NLEV]> qv_scalar(reinterpret_cast<Homme::Scalar*>(qv_i.data()), NLEV);
-        Homme::ExecViewUnmanaged<Homme::Scalar[NLEV]> pnh_scalar(reinterpret_cast<Homme::Scalar*>(pnh_i.data()), NLEV);
-        Homme::ExecViewUnmanaged<Homme::Scalar[NLEV]> exner_scalar(reinterpret_cast<Homme::Scalar*>(exner_i.data()), NLEV);
-        Homme::ExecViewUnmanaged<Homme::Scalar[NLEV]> rstar_scalar(reinterpret_cast<Homme::Scalar*>(rstar_i.data()), NLEV);
         Homme::ExecViewUnmanaged<Homme::Scalar[NLEV]> temperature_scalar(reinterpret_cast<Homme::Scalar*>(temperature_i.data()), NLEV);
```
tcclevenger, line 378: I think all these homme version of views are not needed any longer.

components/eamxx/src/dynamics/homme/eamxx_homme_iop.cpp
```diff
@@ -333,67 +338,60 @@ apply_iop_forcing(const Real dt)
...
   WorkspaceMgr eamxx_wsm(NLEVI, 7+qsize, policy_eamxx);
-  WorkspaceMgr homme_wsm(NLEV, 16 + (theta_hydrostatic_mode ? 16 : 0), policy_homme);
   view_Nd<Pack, 4>
-    rstar      ("rstar",       nelem, NGP, NGP, NLEV),
-    exner      ("exner",       nelem, NGP, NGP, NLEV),
     temperature("temperature", nelem, NGP, NGP, NLEV);
 
-  // Lambda for computing rstar, exner, and temperature from Hommexx
+  // Lambda for computing temperature from Hommexx
   auto compute_homme_states = [&] () {
```
tcclevenger, line 345: We're no longer computing multiple homme states, I would change the name here to something like `compute_temperature`. I would also remove "from Hommexx" in the comment, since we are using physics functions.

components/eamxx/src/dynamics/homme/eamxx_homme_iop.cpp
```diff
@@ -494,7 +509,7 @@ apply_iop_forcing(const Real dt)
     // and observed quantities of T, Q, u, and
 
     if (iop_nudge_tq) {
-      // Compute rstar, exner and temperature from Hommexx
+      // Compute temperature from Hommexx
```
tcclevenger, line 512: We can remove "from Hommexx" here too.

components/eamxx/src/dynamics/homme/eamxx_homme_iop.cpp
```diff
@@ -576,19 +591,31 @@ apply_iop_forcing(const Real dt)
       KV kv(team);
       const int ie  =  team.league_rank();
 
+      // Get temp views from workspace
+      auto ws = eamxx_wsm.get_workspace(team);
+      uview_1d<Pack> pmid, pint, pdel;
```
tcclevenger, line 596: Same thing, we don't need pint or pdel.

components/eamxx/src/dynamics/homme/eamxx_homme_iop.cpp
```diff
@@ -333,67 +338,60 @@ apply_iop_forcing(const Real dt)
...
         auto qv_i          = ekat::subview(Q_dyn, ie, 0, igp, jgp);
-        auto pnh_i         = ekat::subview(pnh, idx);
-        auto rstar_i       = ekat::subview(rstar, ie, igp, jgp);
-        auto exner_i       = ekat::subview(exner, ie, igp, jgp);
         auto temperature_i = ekat::subview(temperature, ie, igp, jgp);
 
+        // Compute reference pressures and layer thickness.
+        // TODO: Allow geometry data to allocate packsize
+        auto s_pmid = ekat::scalarize(pmid);
+        Kokkos::parallel_for(Kokkos::TeamVectorRange(team, total_levels), [&](const int& k) {
```
tcclevenger, line 368: This is related to my overall comment about these policy change below. But we should not use a `TeamVectorRange` here, because we are inside a `TeamThreadRange`. We should use a `ThreadVectorRange`.

The possible patterns for team policies are 
```
TeamPolicy {
  TeamThreadRange {
    ThreadVectorRange {
  }
}
```
or
```
TeamPolicy {
  TeamVectorRange {
}
```
But ignore this comment if you decide to change all the policies to the `policy_eamxx` where there would be no need for the `TeamThreadRange`.

components/eamxx/src/dynamics/homme/eamxx_homme_iop.cpp
```diff
@@ -457,18 +455,32 @@ apply_iop_forcing(const Real dt)
...
       auto temperature_i = ekat::subview(temperature, ie, igp, jgp);
 
-      // Reinterperate into views of Homme::Scalar for calling Hommexx function.
-      Homme::ExecViewUnmanaged<Homme::Scalar[NLEV]> qv_scalar(reinterpret_cast<Homme::Scalar*>(qv_i.data()), NLEV);
-      Homme::ExecViewUnmanaged<Homme::Scalar[NLEV]> rstar_scalar(reinterpret_cast<Homme::Scalar*>(rstar_i.data()), NLEV);
+      // Compute reference pressures and layer thickness.
+      // TODO: Allow geometry data to allocate packsize
+      auto s_pmid = ekat::scalarize(pmid);
+      auto s_pint = ekat::scalarize(pint);
+      Kokkos::parallel_for(Kokkos::TeamVectorRange(team, total_levels+1), [&](const int& k) {
```
tcclevenger, line 476: Same comment as above for `TeamVectorRange` -> `ThreadVectorRange`.

tcclevenger (commented):
A large drive-by comment before you asked for a review.. Feel free to ignore...

So the only incorrect code here is the `TeamVectorRange` being used for computing `pmid` (need to use `ThreadVectorRange` if we are inside a `TeamThreadRange`). But there is a change to team policies that we should make that would take care of that bug. If you want to you can change in this PR, or I can make a PR after this if you do not want worry with it.

Change: Because we are no longer using hommexx functions we don't need the distinction of `policy_homme` (loops over elements) and `policy_eamxx` (loops over all columns). We can just loop over all columns (and name change `policy_eamxx` -> `policy` and `eamxx_wsm` -> `wsm`). Basically change
```
Kokkos::parallel_for("compute_temperature_for_iop", policy_homme, KOKKOS_LAMBDA (const KT::MemberType& team) {
      KV kv(team);
      const int ie  =  team.league_rank();

      // Get temp views from workspace
      auto ws = eamxx_wsm.get_workspace(team);
      uview_1d<Pack> pmid;
      ws.take_many_contiguous_unsafe<1>({"pmid"},{&pmid});

      Kokkos::parallel_for(Kokkos::TeamThreadRange(kv.team, NGP*NGP), [&] (const int idx) {
        const int igp = idx/NGP;
        const int jgp = idx%NGP;
        
        auto ps_i          = ps_dyn(ie, igp, jgp);
        auto dp3d_i        = ekat::subview(dp3d_dyn, ie, igp, jgp);
        auto vtheta_dp_i   = ekat::subview(vtheta_dp_dyn, ie, igp, jgp);
        auto qv_i          = ekat::subview(Q_dyn, ie, 0, igp, jgp);
        auto temperature_i = ekat::subview(temperature, ie, igp, jgp);
```
to
```
Kokkos::parallel_for("compute_temperature_for_iop", policy_eamxx, KOKKOS_LAMBDA (const KT::MemberType& team) {
    const int ie  =  team.league_rank()/(NGP*NGP);
    const int igp = (team.league_rank()/NGP)%NGP;
    const int jgp =  team.league_rank()%NGP;

    // Get temp views from workspace
    auto ws = eamxx_wsm.get_workspace(team);
    uview_1d<Pack> pmid;
    ws.take_many_contiguous_unsafe<1>({"pmid"},{&pmid});

    auto ps_i          = ps_dyn(ie, igp, jgp);
    auto dp3d_i        = ekat::subview(dp3d_dyn, ie, igp, jgp);
    auto vtheta_dp_i   = ekat::subview(vtheta_dp_dyn, ie, igp, jgp);
    auto qv_i          = ekat::subview(Q_dyn, ie, 0, igp, jgp);
    auto temperature_i = ekat::subview(temperature, ie, igp, jgp);
```
We can also remove most Hommexx includes, typedefs (or `using ...`) for Hommexx, and eos and elem_ops objects. Let me know what you think.

oksanaguba:
@bogensch do i understand it right that the main purpose of this PR is the change around line 484 in components/eamxx/src/dynamics/homme/eamxx_homme_iop.cpp ? If so, i am getting from algebra that the old code is the same as the new code (though i haven't tried to run).

Is this PR related to https://github.com/E3SM-Project/scream/issues/2813 or not? in other words, does this PR intend to 'fix' water loading problem as described in the issue?

A side note, the issue mentioned an old change in the definition of virt temp in PR https://github.com/E3SM-Project/scream/pull/1059/files . That change is about switching the input of calculate_virtual_temperature(), qv, from being a dry MR to wet MR, it does not address water loading.

Which brings another question -- if we have functions like calculate_virtual_temperature() operate in terms of wet MRs, how do we know that p3 (which operates with dry MRs) is not calling it with dry MRs? and if it does not know, how to we prevent other packages from using wrong wet/dry MRs in such functions? should we reconsider names/interfaces for these functions?

bartgol:
I'm not sure what the issue is, but the test `homme_shoc_cld_spa_p3_rrtmgp_pg2_dp` runs insanely slow in this PR. In master (and other PRs), it runs in less than 30s. With this PR, it does not complete in the time limits of the job.

@jgfouca The thing that catches my eye here is some new use of WSM. I remember we had some issues with WSM in the past, with some tests taking an insanely long amount of time. I wonder if the WSM usage in this PR is exposing an issue in the impl. Another thought is that we may need to update the scratch memory requirements for homme, since we use more wsm (though I see eamxx_wsm being created with 7+qsize slots, so that may not be it).

tcclevenger:
@bartgol Could this be related to TeamPolicy, containing TeamThreadRange, containing TeamVectorRange? Although I guess I would expect that to hang over just running for a long time..

bartgol:
> @bartgol Could this be related to TeamPolicy, containing TeamThreadRange, containing TeamVectorRange? Although I guess I would expect that to hang over just running for a long time..

That may actually be the case. In particular [this](https://github.com/E3SM-Project/scream/blob/771fa4ee77919b0a102e0e92e6e45ffe1f254a00/components/eamxx/src/dynamics/homme/eamxx_homme_iop.cpp#L353) line seems wrong. It request a nlevi-long slot, but we actually need NGP*NGP copies of that. 

The test _does_ indeed hang. It doesn't complete within the jenkins job time limit. My guess is that 16 team members (which means 16 sets of 32 threads) are all using the same memory, causing a huge congestion. Even if no errors ensued, the calculations would definitely be wrong, since all 16 gauss points will be using the same pmid, which may contain a whole lot of garbage.

I think we should roll back to what was done before:
```
      auto pnh_slot   = ws.take_macro_block("pnh"  , NGP*NGP);
      uview_2d<Pack> pnh(reinterpret_cast<Pack*>(pnh_slot.data()), NGP*NGP, NLEV);
 ```
Maybe simply renamed pmid, now that we changed formulation.

**Comment by bogensch** [comment]
@tcclevenger thanks for your feedback.  I believe I have now made all of your suggested changes including the change to policy_eamxx.

**Comment by bogensch** [comment]
> @bogensch do i understand it right that the main purpose of this PR is the change around line 484 in components/eamxx/src/dynamics/homme/eamxx_homme_iop.cpp ? If so, i am getting from algebra that the old code is the same as the new code (though i haven't tried to run).
> 
> Is this PR related to #2813 or not? in other words, does this PR intend to 'fix' water loading problem as described in the issue?
> 
> A side note, the issue mentioned an old change in the definition of virt temp in PR https://github.com/E3SM-Project/scream/pull/1059/files . That change is about switching the input of calculate_virtual_temperature(), qv, from being a dry MR to wet MR, it does not address water loading.
> 
> Which brings another question -- if we have functions like calculate_virtual_temperature() operate in terms of wet MRs, how do we know that p3 (which operates with dry MRs) is not calling it with dry MRs? and even if we do know, how do we prevent other packages from using wrong wet/dry MRs in such functions? should we reconsider names/interfaces for these functions?

@oksanaguba the bulk of the changes are on line 484 and line 383.  This routine applies the prescribed large scale forcing (IOP forcing) for DP runs.  This routine is called in homme post process.  The IOP forcing tendency is for absolute temperature.  Thus, before we can apply the forcing we need to covert vtheta_dp to temperature... apply the forcing... then convert back to vtheta_dp before the end of homme post process.  I didn't work out the algebra but my tests showed that these two definitions are not b4b and was causing a slow drift in long simulations.  Using the consistent definition of vtheta_dp that homme v1 uses solves this issue. 

No, this PR is not related to #2813.    

I haven’t had a chance to digest your last two paragraphs but they seem unrelated to this PR and discussion on this should probably be placed somewhere more relevant?

**Comment by oksanaguba** [comment]
@bogensch thanks. in your simulations with a drift, were those hydrostatic or nonhydrostatic runs?

separately, since you mentioned that it is about forcing, homme has a routine for forcing application, to be consistent, should that routine be called here instead? there are details about how forcing is applied to geopotential in NH runs, for example.

File: components/eamxx/src/dynamics/homme/eamxx_homme_iop.cpp [hunk 8]
```diff
@@ -324,15 +310,10 @@ apply_iop_forcing(const Real dt)
...
 
-  // Team policy and workspace manager for both homme and scream
-  // related loops. We need separate policies since hommexx functions used here
-  // assume they are called inside nested loops for elements and Gaussian points,
-  // whereas EAMxx function we use expects a single level of parallelism
-  // for elements and Guassian points.
+  // Team policy and workspace manager for eamxx
   // TODO: scream::ColumnOps functions could take an arbitary loop boundary
   //       (TeamVectorRange, TeamThreadRange, ThreadVectorRange) so that
   //       all 3 kernel launches here could be combined.
```
tcclevenger, line 316: This TODO no longer applies either, can you remove?

components/eamxx/src/dynamics/homme/eamxx_homme_iop.cpp
```diff
@@ -324,15 +310,10 @@ apply_iop_forcing(const Real dt)
...
-  const auto policy_homme = ESU::get_default_team_policy(nelem, NLEV);
   const auto policy_eamxx = ESU::get_default_team_policy(nelem*NGP*NGP, NLEV);
```
tcclevenger, line 317: I'd name this "policy" now since there is no eamxx specific policy. Same with "wsm" below.

components/eamxx/src/dynamics/homme/eamxx_homme_iop.cpp
```diff
@@ -341,49 +322,38 @@ apply_iop_forcing(const Real dt)
...
+      // Compute reference pressures and layer thickness.
+      // TODO: Allow geometry data to allocate packsize
+      auto s_pmid = ekat::scalarize(pmid);
+      Kokkos::parallel_for(Kokkos::TeamVectorRange(team, total_levels), [&](const int& k) {
+        s_pmid(k) = hyam(k)*ps0 + hybm(k)*ps_i;
+      });
+      team.team_barrier();
 
+      // Compute temperature from virtual potential temperature
+      Kokkos::parallel_for(Kokkos::ThreadVectorRange(team, NLEV), [&] (const int k) {
```
tcclevenger, line 352: This now should be a `TeamVectorRange` since this is the 1st level of ||ism inside `TeamPolicy` loop, and there are no more interior loops to ||ize. You could also use a `TeamThreadRange`, but that could potentially leave some ||ism on the table. 

But `ThreadVectorRange` is assuming that we are inside a `TeamThreadRange` which we no longer are. So this is a bug.

components/eamxx/src/dynamics/homme/eamxx_homme_iop.cpp
```diff
@@ -447,70 +417,66 @@ apply_iop_forcing(const Real dt)
...
 
-      team.team_barrier();
+    team.team_barrier();
 
-      // Compute Qdp from updated Q
-      Kokkos::parallel_for(Kokkos::ThreadVectorRange(team, NLEV*qsize), [&] (const int k) {
-        const int ilev = k/qsize;
-        const int q = k%qsize;
+    // Compute Qdp from updated Q
+    Kokkos::parallel_for(Kokkos::ThreadVectorRange(team, NLEV*qsize), [&] (const int k) {
```
tcclevenger, line 453: Change to `TeamVectorRange`

components/eamxx/src/dynamics/homme/eamxx_homme_iop.cpp
```diff
@@ -447,70 +417,66 @@ apply_iop_forcing(const Real dt)
...
+    });
       team.team_barrier();
 
-      // Convert updated temperature back to psuedo density virtual potential temperature
-      Kokkos::parallel_for(Kokkos::ThreadVectorRange(team, NLEV), [&] (const int k) {
-          const auto th = PF::calculate_theta_from_T(temperature_i(k),pmid(k));
-          vtheta_dp_i(k) = PF::calculate_virtual_temperature(th,qv_i(k))*dp3d_i(k);
-      });
+    // Convert updated temperature back to psuedo density virtual potential temperature
+    Kokkos::parallel_for(Kokkos::ThreadVectorRange(team, NLEV), [&] (const int k) {
```
tcclevenger, line 464: Change to `TeamVectorRange`

components/eamxx/src/dynamics/homme/eamxx_homme_iop.cpp
```diff
@@ -586,65 +552,63 @@ apply_iop_forcing(const Real dt)
...
+
+      // Compute reference pressures and layer thickness.
+      // TODO: Allow geometry data to allocate packsize
+      auto s_pmid = ekat::scalarize(pmid);
+      Kokkos::parallel_for(Kokkos::TeamVectorRange(team, total_levels), [&](const int& k) {
+        s_pmid(k) = hyam(k)*ps0 + hybm(k)*ps_i;
+      });
+      team.team_barrier();
+
+      Kokkos::parallel_for(Kokkos::ThreadVectorRange(team, NLEV), [&](const int& k) {
```
tcclevenger, line 583: Change to `TeamVectorRange`

components/eamxx/src/dynamics/homme/eamxx_homme_iop.cpp
```diff
@@ -319,82 +310,59 @@ apply_iop_forcing(const Real dt)
...
+        s_pmid(k) = hyam(k)*ps0 + hybm(k)*ps_i;
+      });
+      team.team_barrier();
+
+      // Compute temperature from virtual potential temperature
+      Kokkos::parallel_for(Kokkos::ThreadVectorRange(team, NLEV), [&] (const int k) {
+        auto T_val = vtheta_dp_i(k);
+        T_val /= dp3d_i(k);
+        T_val = PF::calculate_temperature_from_virtual_temperature(T_val,qv_i(k));
+        temperature_i(k) = PF::calculate_T_from_theta(T_val,pmid(k));
```
oksanaguba, line 356: @bogensch a guess -- could it be that the differences you see are in NH runs? this line above, new, uses Exner pressure from hydro pressure, while the old version uses NH pressure. if you agree, then the old version is 
 consistent with homme EOS.

tcclevenger (changes requested):
This looks good, just a few more conversions to `TeamVectorRange` needed. And then some cosmetic suggestions.

components/eamxx/src/dynamics/homme/eamxx_homme_iop.cpp
bartgol: For clarity: you _can_ use TeamThreadRange, and you don't leave any ||ism on the table _as long as the policy was created with vector length 1_. That's usually the case for policies created in eamxx (homme's policies for instance are not like that, since they are tailored for 3 layers of ||ism).

If the policy has vector length>1, not only TeamThreadRange is wasting threads, but it may give bad answers, since all threads in the warp will run the same code (so stuff like += will be executed many times over).

components/eamxx/src/dynamics/homme/eamxx_homme_iop.cpp
bogensch: @oksanaguba yes, I was running NH.

oksanaguba:
i removed automerge label from this for now till we sort out diffs between old and new versions and communicate about forcing algorithms.

oksanaguba:
@bogensch could we possibly talk about this? i might be mistaken but the way i read the code, for NH flow, the new code replaces NH pressure with HY pressure in homme EOS. The old code looks correct to me (and it used homme EOS routine), but the new one may behave better because of 'better' HY pressure. 

separately, the old code could have called `          eos.compute_pnh_and_exner(kv, vtheta_dp_scalar, phi_int_scalar, pnh_scalar, exner_scalar);` for both HY and NH parts.

another issue is that if you are applying forcings (i must admit i haven't followed what IOP taks is for), then it probably should be done with a call to one (or both) of homme forcings routines, where we worked out how forcings are used for now.

thanks!

bogensch:
> @bogensch could we possibly talk about this? i might be mistaken but the way i read the code, for NH flow, the new code replaces NH pressure with HY pressure in homme EOS. The old code looks correct to me (and it used homme EOS routine), but the new one may behave better because of 'better' HY pressure.
> 
> separately, the old code could have called ` eos.compute_pnh_and_exner(kv, vtheta_dp_scalar, phi_int_scalar, pnh_scalar, exner_scalar);` for both HY and NH parts.
> 
> another issue is that if you are applying forcings (i must admit i haven't followed what IOP taks is for), then it probably should be done with a call to one (or both) of homme forcings routines, where we worked out how forcings are used for now.
> 
> thanks!

Thanks @oksanaguba , you are probably right.  I will update the information on this pull request to reflect that, so it is not misleading.  I still prefer to use this new code since it uses consistent formulations that most of the v1 model uses and it's a cleaner code.  

Regarding the homme forcing routines, I'm not familiar with these.  Perhaps @tcclevenger would be better to comment on this.

oksanaguba:
thanks @bogensch . What is this PR for? HOMME already has an established algorithm to apply physics forcings, so i am sure we should keep yet another code here. thanks again!

oksanaguba:
oh i see -- the description says 'DPxx'. i was confused by the file name, what is IOP? anyway, for DP we cannot use two different forcing applications. we should try using homme existing routines and see if it helps.

the forcings should be stored in FQ, FT, FM fields just like EAM/EAMxx does it. then homme routines will be called in dp-coupling layer (is that part still active in DP). i can talk to @tcclevenger about how it should be implemented. thanks!

PeterCaldwell:
@oksanaguba : 

1. This PR *only* affects DP mode. That's what the "iop" designation means. IOP stands for intensive observation period which is old terminology from CESM for single column and DP configurations. We're using iop to denote DP/SCM-specific code because we can't come up with a better name.
2. Our current goal for DPxx is to have it perfectly reproduce the behavior of DPf90 (modulo the fact that the 2 implementations can't be BFB). This *needs* to happen for us to be able to verify that porting to C++ didn't introduce errors. Once we verify that DPxx is working as expected, we can make additional changes in future PRs. You're probably right that it would be better to use the existing HOMME conversion functions (but now isn't the time to do that).
3. As additional motivation for not being too careful with our current implementation, we're already planning to eventually move these DP-specific calculations completely out of the dycore into their own "iop" process. Since the dycore changes we're making are temporary, it doesn't make sense to make them perfect.

Does this seem reasonable? Basically we need to do the "wrong" thing here in order to be consistent with the way SCREAMv0 operated in DP mode and we can switch to the "right" way once we create a baseline that agrees with v0.

oksanaguba:
thanks, Peter, if i may ask a few more questions -- how was coupling of DPff was done? surely it used the same dp-coupling layer in F EAM code as before? if not, could someone (Peter B?) point me to the code. If DPff was using the same coupling as EAM uses, then it called the same routines that I am suggesting to call here. That same code handles hydrostatic and nonhydrostatic flows. Therefore DPxx should use the same code, not the new code, if your intention is to have the same results.

Unless DPff really used something else, but then just please point me to the code, thanks!

separately, i do not know if it matters here, but when homme nonhydrostatic is used, T tendency ends up affecting theta AND geopotential, something that is not done in this PR.

bogensch:
The fortran equivalent of this code can be found at: 
components/eam/src/dynamics/se/se_iop_intr_mod.F90 

Specifically the "apply_iop_forcing" routine.  This may be messer than the c++ converted routine as that contained hooks for the single column model which v1 will not have.

oksanaguba:
thanks, both Peters, i appreciate your patience. from looking quickly, it seems to me that F code is consistent with the old version of DPxx coupling, not the one that is proposed in this PR?

separately, if DP model is sensitive to the coupling mechanism, just like Peter B said, when DPxx code is used, one should keep it in mind that full eamxx and DPxx use different phys-dyn coupling, and the results from DP might not apply to eamxx?

oksanaguba:
... also, i removed my 'do not merge' label, thanks

PeterCaldwell:
> it seems to me that F code is consistent with the old version of DPxx coupling, not the one that is proposed in this PR?
Odd since empirically the new DPxx coupling is needed to get the same answers as the F code.

> full eamxx and DPxx use different phys-dyn coupling [so] the results from DP might not apply to eamxx?
This is a great point and good motivation to switch implementations later!!!

oksanaguba:
We were brainstorming this with Mark and Andrew and one suggestion was that while DP uses the usual dp-coupling layer logic from eamxx, it also uses a separate mechanism for coupling in apply_iop_forcing for some other forcing (forcing from a file? nudging?). One of our confusions is that dp_crm and single_column variables are not mutually exclusive, does it mean in DP variable single_column is used for this extra nudging(?) forcing?

I ran FDPSCREAM-ARM97 compset with print statements to see which vars are on in this compset. Outputs are below. It seems that both codes, usual logic for pd coupling and apply_iop_forcing, are active, but for this particular case, which depends on condition ".not. use_3dfrc .or. dp_crm", temperature tendencies are set to zero. However, I did not try to track what happens to Q or momentum tendencies.

Do we have a DP case where coupling is done via dp-coupling mechanism only? In which case is extra forcing applied in apply_iop_forcing? What happens with pressure adjustment when this extra mechanism is active?

```
0:  OG in cam_tracers FT  2.169649261199358E-005  3.361880573371314E-005

0:  OG in stepon doiopupdate T
0:  OG in stepon dp_crm T
0:  OG in dyn_comp in dyn_run single_column= T
0:  OG in dyn_comp in dyn_run dp_crm= T
0:  OG in dyn_comp in dyn_run about to call apply_iop_forcing
0:  OG in apply_iop_forcing
0:  OG scm_multcols T
0:  OG dp_crm T
0:  OG iop_nudge_tq F
0:  OG iop_dosubsidence F
0:  OG NOTdefined E3SM_SCM_REPLAY

//this one means that temp_tend = 0 in apply_iop_forcing
0:  OG we are in if-statement of \.not. use_3dfrc .or. dp_crm\ 
```

oksanaguba:
Regarding Peter's comment `Odd since empirically the new DPxx coupling is needed to get the same answers as the F code.` -- if apply_iop_forcing routine in F and cxx does not have a unit test (question to Conrad, probbaly, but also considering this PR is taking place, there is probably no test), then there may be other parts of the code that do not match between F and cxx. While it is less trivial, since unit test may be too hard to achieve, one could try to run both codes side by side with print statements.

bogensch:
> We were brainstorming this with Mark and Andrew and one suggestion was that while DP uses the usual dp-coupling layer logic from eamxx, it also uses a separate mechanism for coupling in apply_iop_forcing for some other forcing (forcing from a file? nudging?). One of our confusions is that dp_crm and single_column variables are not mutually exclusive, does it mean in DP variable single_column is used for this extra nudging(?) forcing?
> 
> I ran FDPSCREAM-ARM97 v0 compset with print statements to see which vars are on in this compset. Outputs are below. It seems that both codes, usual mechanism for pd coupling and apply_iop_forcing, are active, but for this particular case, which depends on condition ".not. use_3dfrc .or. dp_crm", temperature tendencies in apply_iop_forcing are set to zero. However, I did not try to track what happens to Q or momentum tendencies.
> 
> Do we have a DP case where coupling is done via dp-coupling mechanism only? In which case is extra forcing applied in apply_iop_forcing? What happens with pressure adjustment when this extra mechanism is active? Do we really need a different implementation of coupling in apply_iop_forcing?
> 
> ```
> 0:  OG in cam_tracers FT  2.169649261199358E-005  3.361880573371314E-005
> 
> 0:  OG in stepon doiopupdate T
> 0:  OG in stepon dp_crm T
> 0:  OG in dyn_comp in dyn_run single_column= T
> 0:  OG in dyn_comp in dyn_run dp_crm= T
> 0:  OG in dyn_comp in dyn_run about to call apply_iop_forcing
> 0:  OG in apply_iop_forcing
> 0:  OG scm_multcols T
> 0:  OG dp_crm T
> 0:  OG iop_nudge_tq F
> 0:  OG iop_dosubsidence F
> 0:  OG NOTdefined E3SM_SCM_REPLAY
> 
> //this one means that temp_tend = 0 in apply_iop_forcing
> 0:  OG we are in if-statement of \.not. use_3dfrc .or. dp_crm\ 
> ```

Oksana, yes, the naming convention is a bit misleading due to having to support both a SCM and DP CRM, both of which use the same infrastructure but need to be treated differently in certain instances.  This confusion will not exist in v1 (which won’t have an official SCM), but this naming convention needs to be cleared up in E3SM (something that I am to blame for and is on my plate) code base and v0.

single_column flag means that the original SCM infrastructure is being utilized; which is something that both the SCM and DP-SCREAM utilize since DP-SCREAM is a “single-point CRM”.  Probably something like “iop_mode” would be more appropriate name for this.

Forcing comes from IOP files (Intensive Observation Period).  When you run a case look for the variable "iopfile" in your atm_in.  That will be the file where the forcing (and initial sounding for the run) are coming from.  For more info please check out the wiki for the SCM and DP-SCREAM modes (https://github.com/E3SM-Project/scmlib/wiki).  There you will find links to the SCM and DP-SCREAM papers.  Perhaps checking out the intro and design sections of those papers would be helpful?

I don’t think I understand your question about DP case where coupling is done vi dp-coupling mechanism only.  Do you mean a case where the forcing in the IOP file is zero everywhere so that the apply_iop_forcing routine is not doing any work?  If so, I would suggest the Radiative Convective Equilibrium (RCE) case.  Details on that case and how to run can be found at the above wiki.  The ARM97 case has large scale forcing provided for T & Q so that is a case where apply_iop_forcing is being exercised.

oksanaguba:
thanks @bogensch . 

Regarding `question about DP case where coupling is done vi dp-coupling mechanism only.` -- in eam/eamxx we have what we call a "dp-coupling layer", where we have logic about how to compute physics tendencies for homme. this layer also may be calling homme routines to apply these tendencies (and also some of these tendencies may be applied by dribbling during dycore timestepping). we are interested in preserving these mechanisms in DP, because this is how phys-dyn coupling is done in eamxx. to me, DP is a model that can be run with the same physics as eam/eamxx, just cheaper. So having the same mechanisms for forcing in DP and in eamxx seems to be important.

it seems though that you have another way of using DP, by using "nudging" forcing from a file, let's call it nudging coupling. in the case i ran, i saw that both codes, dp-coupling and nudging coupling, are active. in case i ran, `FDPSCREAM-ARM97`, i see that "nudging" part has zero T tendencies (details are below). I did not try to trace what happens to "nudging" Q tendencies in this compset. I do have "iopfile" in atm_in. In the same case, i see that "dp-coupling" tendencies are not zero.

Did you intend to have "dp-coupling" tendencies ever active in DP runs? is the usual eamxx physics active in DP cases? considering that both "dp-coupling" and "nudging" couplings operate with elem%derived%FT, i am also puzzled how they both could potentially be active in some other compset (not the one i ran).

Here are more details on what i ran:
I put this debug code in
```
   @@ -333,10 +340,23 @@ subroutine apply_iop_forcing(elem,hvcoord,hybrid,tl,n,t_before_advance,nets,nete
 
         if (.not. use_3dfrc .or. dp_crm) then
           temp_tend(:) = 0.0_real_kind
+
+print *, 'OG we are in if-statement of \.not. use_3dfrc .or. dp_crm\ ' 
+print *, 'OG temp_tend', temp_tend(:)
+
+
         else
           temp_tend(:) = elem(ie)%derived%fT(i,j,:)
+
+print *, 'OG we are in ELSE of if-statement \.not. use_3dfrc .or. dp_crm\ ' 
+if( (i==1) .and. (j==1))then
+print *, 'OG in apply_iop_forcing FT is ',elem(ie)%derived%fT(i,j,:)
+endif
         endif
 
+
+
+
 #ifdef MODEL_THETA_L
         ! At first time step the tendency term is set to the
         !  initial condition temperature profile.  Do NOT use
@@ -364,6 +384,9 @@ subroutine apply_iop_forcing(elem,hvcoord,hybrid,tl,n,t_before_advance,nets,nete
 
         ! Call the main subroutine to update t, q, u, and v according to
         !  large scale forcing as specified in IOP file.
+
+print *, 'OG about to call advance_iop_forcing'
+print *, 'OG before advance_iop_forcing temp_tend', temp_tend(:)
         call advance_iop_forcing(dt,elem(ie)%state%ps_v(i,j,t1),& ! In
```
and i see in the output lines like 
```
0:  OG we are in if-statement of \.not. use_3dfrc .or. dp_crm\ 
0:  OG temp_tend  0.000000000000000E+000  0.000000000000000E+000
0:  OG about to call advance_iop_forcing
0:  OG before advance_iop_forcing temp_tend  0.000000000000000E+000
```
so not only tendencies are zero, they will be zero indep. of input file because a certain if-statement is triggered.

bogensch:
@oksanaguba I think I might be confused so let me clarify a few things.....  in DP mode the SCREAM physics is active as usual and these tendencies are applied via dp coupling as usual (an exception is the single column model (SCM) of E3SM where large portions of the dycore are bypassed so the physics forcing is applied a bit different... that is the only time that the code block utilizing temp_tend(:) = elem(ie)%derived%fT(i,j,:) is active ).  The forcing we are applying in advance_iop_forcing is a prescribed large scale forcing that comes from the IOP file (to account for changes in the meteorology coming into our domain); thus it is an additional forcing term.  In some/many cases this is just set to zero everywhere and this code is therefore not exercised.  If it is, this forcing is assumed to be horizontally homogenous and is applied to all columns.... and note that there is no reason why this has to be done in homme post-processing in v1 (in v0 it had to be to satisfy some SCM requirements); so our goal is to eventually move this IOP forcing module into it's own process and get it out of homme post-processing, which I think will clear up a lot of confusion.

mt5555:
adding my interpretation:   This code is applying nudging directly to the dycore variables (for historical reasons).  Thus it needs to reproduce a lot of the code in the dycore's "ApplyCAMForcing".   This is complicated code because Temperature tendencies have to be converted to virtual pot. temperatere and geopotential height tendicies, using energy conserving and moist EOS consistent formulas.  

Some of the formulas used appear (from Oksana's comments) to not quite match what the dycore uses (causing some concern among dycore developers - duplicating code but with a slightly implementation that is hard to tell if it is the same, and might be missing tests of that property.)      Eventually, this code will become it's own process.  If it becomes it's own process, it woudl be good if it works on the physics state (not the dycore state), and then it will just be added into the physics and can be applied to the dycore state via the normal physics/dynamics coupling, and we will avoid duplicating code?

bogensch:
> Eventually, this code will become it's own process. If it becomes it's own process, it woudl be good if it works on the physics state (not the dycore state), and then it will just be added into the physics and can be applied to the dycore state via the normal physics/dynamics coupling, and we will avoid duplicating code?

Exactly, that is my intention... for it to be added into the physics state.

oksanaguba:
thanks, everyone. a couple of questions, why is FDPSCREAM-ARM97 not applying these large-scale tendencies (i thought it is supposed to), and why is the new code (this PR) does not match F code for apply_iop_forcing routine (there i mean conversion T->vtheta and back).

bogensch:
@oksanaguba I'm not sure what you are seeing in ARM97 case, but I just verified that large scale tendencies are being applied on my end for ARM97.  The reason why we changed the conversion from T->vtheta and back is to use the consistent formulations/functions of vtheta_dp that homme v1 was using to define vtheta_dp.  While we strive to keep DPxx and DPv0 as close to each other as possible, we did change a few things to be consistent with v1.

oksanaguba:
@bogensch i made a copy of relevant code below. the old version of the code is consistent with homme as it calls homme EOS routines, and the new one is not, it is alos not consistent with Fortran? regarding what i see in forcing routines -- i will push my branch with debug statements for you to examine tonight (i do not have access to a server right now). thanks!

```

NEW CODE

      auto s_pmid = ekat::scalarize(pmid);
      Kokkos::parallel_for(Kokkos::TeamVectorRange(team, total_levels), [&](const int& k) {
        s_pmid(k) = hyam(k)*ps0 + hybm(k)*ps_i;
      });
      team.team_barrier();

      // Compute temperature from virtual potential temperature
      Kokkos::parallel_for(Kokkos::TeamVectorRange(team, NLEV), [&] (const int k) {
        auto T_val = vtheta_dp_i(k);
        T_val /= dp3d_i(k);
        T_val = PF::calculate_temperature_from_virtual_temperature(T_val,qv_i(k));
        temperature_i(k) = PF::calculate_T_from_theta(T_val,pmid(k));
      });


OLD CODE

        // Compute exner from EOS
        if (theta_hydrostatic_mode) {
          auto hydro_p_int = ws.take("hydro_p_int");
          Homme::ExecViewUnmanaged<Homme::Scalar[NLEVI]> hydro_p_int_scalar(reinterpret_cast<Homme::Scalar*>(hydro_p_int.data()), NLEVI);
          elem_ops.compute_hydrostatic_p(kv, dp3d_scalar, hydro_p_int_scalar, pnh_scalar);
          eos.compute_exner(kv, pnh_scalar, exner_scalar);
          ws.release(hydro_p_int);
        } else {
          eos.compute_pnh_and_exner(kv, vtheta_dp_scalar, phi_int_scalar, pnh_scalar, exner_scalar);
        }

        // Get the temperature from dynamics states
        elem_ops.get_temperature(kv, eos, use_moisture, dp3d_scalar, exner_scalar, vtheta_dp_scalar, qv_scalar, rstar_scalar, temperature_scalar);
      });


FORTRAN CODE

#ifdef MODEL_THETA_L
    ! If using the theta-l dycore then need to get the exner function
    !   and reference levels "dp", so we can convert the SCM updated
    !   temperature back potential temperature on reference levels.
    dp(:,:,:) = elem(ie)%state%dp3d(:,:,:,t1)
    call pnh_and_exner_from_eos(hvcoord,elem(ie)%state%vtheta_dp(:,:,:,t1),&
        dp,elem(ie)%state%phinh_i(:,:,:,t1),pnh,exner,dpnh_dp_i)
#endif

    dt=dtime

    ! Get temperature from dynamics state
    call get_temperature(elem(ie),temperature,hvcoord,t1)

```

bogensch:
> @bogensch i made a copy of relevant code below. the old version of the code is consistent with homme as it calls homme EOS routines, and the new one is not, it is alos not consistent with Fortran? regarding what i see in forcing routines -- i will push my branch with debug statements for you to examine tonight (i do not have access to a server right now). thanks!
> 
> ```
> 
> NEW CODE
> 
>       auto s_pmid = ekat::scalarize(pmid);
>       Kokkos::parallel_for(Kokkos::TeamVectorRange(team, total_levels), [&](const int& k) {
>         s_pmid(k) = hyam(k)*ps0 + hybm(k)*ps_i;
>       });
>       team.team_barrier();
> 
>       // Compute temperature from virtual potential temperature
>       Kokkos::parallel_for(Kokkos::TeamVectorRange(team, NLEV), [&] (const int k) {
>         auto T_val = vtheta_dp_i(k);
>         T_val /= dp3d_i(k);
>         T_val = PF::calculate_temperature_from_virtual_temperature(T_val,qv_i(k));
>         temperature_i(k) = PF::calculate_T_from_theta(T_val,pmid(k));
>       });
> 
> 
> OLD CODE
> 
>         // Compute exner from EOS
>         if (theta_hydrostatic_mode) {
>           auto hydro_p_int = ws.take("hydro_p_int");
>           Homme::ExecViewUnmanaged<Homme::Scalar[NLEVI]> hydro_p_int_scalar(reinterpret_cast<Homme::Scalar*>(hydro_p_int.data()), NLEVI);
>           elem_ops.compute_hydrostatic_p(kv, dp3d_scalar, hydro_p_int_scalar, pnh_scalar);
>           eos.compute_exner(kv, pnh_scalar, exner_scalar);
>           ws.release(hydro_p_int);
>         } else {
>           eos.compute_pnh_and_exner(kv, vtheta_dp_scalar, phi_int_scalar, pnh_scalar, exner_scalar);
>         }
> 
>         // Get the temperature from dynamics states
>         elem_ops.get_temperature(kv, eos, use_moisture, dp3d_scalar, exner_scalar, vtheta_dp_scalar, qv_scalar, rstar_scalar, temperature_scalar);
>       });
> 
> 
> FORTRAN CODE
> 
> #ifdef MODEL_THETA_L
>     ! If using the theta-l dycore then need to get the exner function
>     !   and reference levels "dp", so we can convert the SCM updated
>     !   temperature back potential temperature on reference levels.
>     dp(:,:,:) = elem(ie)%state%dp3d(:,:,:,t1)
>     call pnh_and_exner_from_eos(hvcoord,elem(ie)%state%vtheta_dp(:,:,:,t1),&
>         dp,elem(ie)%state%phinh_i(:,:,:,t1),pnh,exner,dpnh_dp_i)
> #endif
> 
>     dt=dtime
> 
>     ! Get temperature from dynamics state
>     call get_temperature(elem(ie),temperature,hvcoord,t1)
> ```

Yes,  see my comment above "The reason why we changed the conversion from T->vtheta and back is to use the consistent formulations/functions of vtheta_dp that homme v1 was using to define vtheta_dp. While we strive to keep DPxx and DPv0 as close to each other as possible, we did change a few things to be consistent with v1."

oksanaguba:
@bogensch thanks, sorry for the confusion, i thought that either this pr is making dpxx consistent with dp F, or it is making its vtheta definition consistent with homme. i see now that it is neither. 

regarding my debugging of when iop forcing is called (or not), please, see the script below and my branch oksanaguba/dp/debug-forcing and let me know what i am doing wrong. thanks!

```
#!/bin/bash -e
#The -e option means "if any pipeline ever ends with a non-zero ('error') exit status, terminate the script immediately". 

which=scream-master
fol=xxne4-june20-dpscream
nn=1

fcomp=FDPSCREAM-ARM97

#FDPSCREAM-ARM97

~/${which}/cime/scripts/create_newcase --case $fol --res ne4_ne4 --mach chrysalis --compiler intel --compset ${fcomp}

echo "I am ${0}"
cp ${0} ${fol}/
cd $fol

./xmlchange SCREAM_CMAKE_OPTIONS="SCREAM_NP 4 SCREAM_NUM_VERTICAL_LEV 72 SCREAM_NUM_TRACERS 10 SCREAM_HAS_LEAP_YEAR OFF"

if true; then
./xmlchange NTASKS=${nn}
./xmlchange JOB_WALLCLOCK_TIME=3:00:00
./xmlchange PIO_NETCDF_FORMAT="64bit_data"
ncore=64
./xmlchange NTHRDS=1
./xmlchange MAX_MPITASKS_PER_NODE=$ncore
./xmlchange MAX_TASKS_PER_NODE=$ncore

fi

./case.setup --reset
./case.setup


./case.build
./xmlchange STOP_N=5 ; ./xmlchange STOP_OPTION=nsteps ; 

./xmlchange --id JOB_QUEUE --val 'debug,compute' --force

cp ${HOME}/aaa.yaml .

./atmchange hypervis_subcycle_q=1
./atmchange dt_tracer_factor=2
./atmchange tstep_type=5
./atmchange theta_hydrostatic_mode=True
./atmchange transport_alg=0
#./atmchange se_tstep=2

./case.submit
```

bogensch:
> @bogensch thanks, sorry for the confusion, i thought that either this pr is making dpxx consistent with dp F, or it is making its vtheta definition consistent with homme. i see now that it is neither.
> 
> regarding my debugging of when iop forcing is called (or not), please, see the script below and my branch oksanaguba/dp/debug-forcing and let me know what i am doing wrong. thanks!
> 

@oksanaguba First, I don’t see how your experiment is showing that IOP forcing is not being applied.  IOP forcing is applied in advance_iop_forcing routine and the large-scale forcing is stored in the t_lsf and q_lsf (temperature and moisture respectively) arrays; which you are not exploiting in your branch.  I have verified that everything is working as intended.

Second, it is not clear to me how this inquiry pertains to this PR and I worry it is distracting from the goal of getting this merged.  If you are having trouble running DP-SCREAM or have general questions, I think it makes the most sense to send me a direct message on Slack or an email.  Thanks.

oksanaguba:
@bogensch feel free to merge, thanks. i will follow up with slack.

PR #2863, opened 2024-06-10
EAMxx: avoid nullptr when reading horiz remap sparse matrix entries

bartgol:
A std vector of length 0 is not allocated, resulting in invalid memory


brhillman (approved):

bartgol:
The standalone fail must be a fluke of mappy. All sandalone builds reported 100% passed. Perhaps a broken pipe in the suprocess dispatched by test-all-scream returned nonzero. I'm manually merging.

PR #2864, opened 2024-06-12
Aarondonahue/corrective ml/20240611 fix

AaronDonahue:
Fix precipitation adjustment in corrective-ML interface and implement field property checks.


components/eamxx/src/physics/ml_correction/eamxx_ml_correction_process_interface.cpp
```diff
@@ -81,6 +84,16 @@ void MLCorrection::initialize_impl(const RunType /* run_type */) {
...
+
+  // Enforce bounds on quantities adjusted by ML using Field Property Checks
+  using LowerBound = FieldLowerBoundCheck;
+  add_postcondition_check<LowerBound>(get_field_out("precip_liq_surf_mass"),m_grid,0,true);
+  add_postcondition_check<LowerBound>(get_field_out("precip_ice_surf_mass"),m_grid,0,true);
+  // Add sanity property check for other variables.
+  using Interval = FieldWithinIntervalCheck;
+  add_postcondition_check<Interval>(get_field_out("qv"),m_grid,0,0.2,true);
+  add_postcondition_check<Interval>(get_field_out("T_mid"),m_grid,100.0,500.0,false);
+  add_postcondition_check<Interval>(get_field_out("horiz_winds"),m_grid,-400.0,400.0,false);
```
elynnwu, line 96: I think it's great to have bound checks, but should this be done from within ML prediction? There's threshold limits on the predicted tendencies, but I think the issue here is that the actual field can go crazy?

components/eamxx/src/physics/ml_correction/eamxx_ml_correction_process_interface.cpp
```diff
@@ -187,23 +214,16 @@ void MLCorrection::run_impl(const double dt) {
         // Apply all the adjustment to a single phase based on surface temperature
         Kokkos::single(Kokkos::PerTeam(team), [&] {
           auto T_icol = ekat::subview(T_mid,icol);
-          if (T_icol(m_num_levs-1)>273.15) {
+          if (T_icol(num_levs-1)>273.15) {
```
elynnwu, line 217: Is `num_levs` interchangeable with `m_num_levs`?

components/eamxx/cmake/machine-files/ruby-intel.cmake
```diff
@@ -1,7 +1,8 @@
 include(${CMAKE_CURRENT_LIST_DIR}/ruby.cmake)
 set(CMAKE_EXE_LINKER_FLAGS "-L/usr/tce/packages/mkl/mkl-2022.1.0/lib/intel64/ -qmkl" CACHE STRING "" FORCE)
 set(PYTHON_EXECUTABLE "/usr/tce/packages/python/python-3.9.12/bin/python3" CACHE STRING "" FORCE)
+set(Python_EXECUTABLE "/usr/tce/packages/python/python-3.9.12/bin/python3" CACHE STRING "" FORCE)
```
bartgol, line 4: This line seems to be the same as the one above...

components/eamxx/src/physics/ml_correction/eamxx_ml_correction_process_interface.cpp
```diff
@@ -101,8 +114,11 @@ void MLCorrection::run_impl(const double dt) {
   const auto &v               = get_field_out("horiz_winds").get_component(1).get_view<Real **, Host>();
 
   // For precipitation adjustment we need to track the change in column integrated 'qv'
-  decltype(qv) qv_told("", qv.extent(0), qv.extent(1));
-  Kokkos::deep_copy(qv_told,qv);
+  // So we clone the original qv before ML changes the state so we can back out a qv_tend
+  // to use with precip adjustment.
+  auto qv_src = get_field_in("qv");
+  auto qv_in = qv_src.clone();
+  qv_in.deep_copy(qv_src);
```
bartgol, line 121: The clone method already performs a deep copy, so this is redundant work.

components/eamxx/cmake/machine-files/ruby-intel.cmake
elynnwu: It should but I've run into issues where it's looking for the lowercase Python and I'm not sure why

bartgol: Ah, I missed the upper/lower case difference.

That said, since you have `find_package(Python ..)` in `ml_correction/CMakeLists.txt`, this variable [should be set by CMake](https://cmake.org/cmake/help/latest/module/FindPython.html#result-variables).

components/eamxx/src/physics/ml_correction/eamxx_ml_correction_process_interface.cpp
AaronDonahue: We need to save the local variable `num_levs=m_num_levs` for running on Device.

components/eamxx/cmake/machine-files/ruby-intel.cmake
AaronDonahue: I tested it removing both of these python executable lines and it appeared to work.  So the latest branch version has these both gone.

elynnwu (approved):

PR #2865, opened 2024-06-12
Beydoun/fix rh diag

hassanbeydoun:
My understanding of the ice flag in the qv_sat_wet function is that it allows for using qv_sat_ice when T < 0 C. If the reviewers can confirm this, that would be great!


singhbalwinder:
It has been some time since I worked on this but your understanding seems correct. [Here is how we determine](https://github.com/E3SM-Project/scream/blob/b7f853cf108dc55a33a69ae5986711f54cf3fe9b/components/eamxx/src/physics/share/physics_saturation_impl.hpp#L54) whether to use the ice implementation or not:
```
const Smask ice_mask = (t_atm < tmelt) && ice;
```
So, the temperature should be < tmelt, and the `ice` should be `True`.

singhbalwinder (approved):

bartgol (approved):

bartgol:
I think the fail is unrelated (we need to regen the baselines). Merging.

PR #2867, opened 2024-06-13
Fix usage of code deprecated Kokkos code, and updates ekat

bartgol:
We can now build with `Kokkos_ENABLE_DEPRECATED_CODE_4=OFF`. Also, brings in an update in ekat that suppresses the zillion warnings we were getting.

@jgfouca I did not set `Kokkos_ENABLE_DEPRECATED_CODE_4=OFF` in our test-all-scream builds. But maybe we should consider doing it at some point?

### DIFF
--- a/components/eamxx/src/dynamics/homme/physics_dynamics_remapper.cpp
+++ b/components/eamxx/src/dynamics/homme/physics_dynamics_remapper.cpp
@@ -413,7 +413,7 @@ do_remap_fwd()
 
   using TeamPolicy = typename KT::TeamTagPolicy<RemapFwdTag>;
 
-  const auto concurrency = KT::ExeSpace::concurrency();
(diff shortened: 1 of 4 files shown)


jgfouca:
Am i going crazy? I don't see the ekat submodule being updated?

bartgol:
> Am i going crazy? I don't see the ekat submodule being updated?

I forgot to push the commit, but you should see it now.

jgfouca (approved):

jgfouca:
Looks like weaver failed

bartgol:
We still have the issue that test-all-scream does not really updates baselines when running through the AT (something is amiss with the master merge and the ahead/behind calculation). I am manually forcing baseline regen now.

jgfouca (approved):

PR #2869, opened 2024-06-17
test-all-scream: Simplify the baseline model

### LINKED ISSUE #2852: Clean up test-all-scream handling of baselines generation
There's something wrong in test-all-scream handling of baselines. Perhaps it's not TAS itself, but how we use it in our jenkins jobs. The problem is that master baselines do not get updated, but the sha file in the baseline dir does. As a result, we get old baselines which later runs of TAS see has up-to-date.

@jgfouca  and I discussed some changes. We think we need to better separate the generation of baselines from the testing. Some thoughts (Jim, feel free to edit my post).

- Do not perform the master merge inside TAS. The user is responsible for taking care of that.
- When -g is passed, do not run tests
- Other?

jgfouca:
By default, test-all-scream will not run baseline tests. If you set -b AUTO, baseline tests will be done with the pre-existing public baselines in the location specified by the machine spec. You can regenerate baselines any time by using the -g flag, but be aware this will impact everyone if you regenerate the public baselines. You can change the target baseline area using -b $path. If -g is provided, no tests will be run; -g means generate only.

The general workflow for baseline-changing PRs is:
1) Issue PR
2) AT will fail with differences in the baseline tests
3) Review and merge.
4) Ask JimF or LucaB for a bless to baselines on mappy and weaver.

If you are developing a branch and baselines tests are failing unexpectedly, it is likely that your branch has fallen out of date. You should upstream merge or rebase your branch.

To sum up, the baseline handling for test-all-scream should basically match what we do for create_test tests, the only difference is that test-all-scream does baseline comparison tests by default.

Fixes #2852

### DIFF
--- a/components/eamxx/scripts/jenkins/jenkins_common_impl.sh
+++ b/components/eamxx/scripts/jenkins/jenkins_common_impl.sh
@@ -72,7 +72,7 @@ if [ $skip_testing -eq 0 ]; then
   # IF such dir is not found, then the default (ctest-build/baselines) is used
   BASELINES_DIR=AUTO
 
-  TAS_ARGS="--baseline-dir $BASELINES_DIR \$compiler -p -c EKAT_DISABLE_TPL_WARNINGS=ON -i -m \$machine"
+  TAS_ARGS="--baseline-dir $BASELINES_DIR \$compiler -p -c EKAT_DISABLE_TPL_WARNINGS=ON -m \$machine"
   # pm-gpu needs to do work in scratch area in order not to fill home quota
   if [[ "$SCREAM_MACHINE" == "pm-gpu" ]]; then
       TAS_ARGS="${TAS_ARGS} -w /pscratch/sd/e/e3smtest/e3sm_scratch/pm-gpu/ctest-build"
@@ -90,6 +90,16 @@ if [ $skip_testing -eq 0 ]; then
       if [ -z "$SCREAM_FAKE_ONLY" ]; then
           # Run EKAT tests for real nightly runs
           TAS_ARGS="${TAS_ARGS} --submit -c EKAT_ENABLE_TESTS=ON"
+      fi
+  fi
+
+  # AT runs may need an upstream merge in order to ensure that any DIFFs
+  # are caused by this PR and not simply because the PR is too far behind master.
+  if [ -z "$SCREAM_FAKE_ONLY" && $is_at_run == 1 ]; then
+      ./scripts/git-merge-ref origin/master
+      if [[ $? != 0 ]]; then
+          echo "MERGE FAILED! Please resolve conflicts"
+          exit 1
       fi
   fi
 
@@ -167,15 +177,6 @@ if [ $skip_testing -eq 0 ]; then
     if [[ -z "$SCREAM_FAKE_ONLY" && $is_at_run == 1 ]]; then
 
       if [[ $test_v0 == 1 || $test_v1 == 1 ]]; then
-        # AT CIME runs may need an upstream merge in order to ensure that any DIFFs
-        # are caused by this PR and not simply because the PR is too far behind master.
-        if [ -n "$PULLREQUESTNUM" ]; then
-          ./scripts/git-merge-ref origin/master
-          if [[ $? != 0 ]]; then
-              echo "MERGE FAILED! Please resolve conflicts"
-              exit 1
-          fi
-        fi
       fi
 
       if [[ $test_v0 == 1 ]]; then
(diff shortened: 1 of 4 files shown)


components/eamxx/scripts/jenkins/jenkins_common_impl.sh
```diff
@@ -93,6 +93,16 @@ if [ $skip_testing -eq 0 ]; then
       fi
   fi
 
+  # AT runs may need an upstream merge in order to ensure that any DIFFs
+  # are caused by this PR and not simply because the PR is too far behind master.
+  if [ -z "$SCREAM_FAKE_ONLY" && $is_at_run == 1 ]; then
+      ./scripts/git-merge-ref origin/master
```
jgfouca, line 99: Do the merge here so that test-all-scream doesn't have to worry about it.

bartgol: I think we are now handling the merge inside the jenkins plugin, so we may as well remove it here... One less thing to worry about in the scripts...

jgfouca: oh, excellent

components/eamxx/scripts/test-all-scream
```diff
@@ -27,26 +48,26 @@ def parse_command_line(args, description):
...
+    \033[1;32m# Run all tests on current machine with default behavior except using non-default baselines \033[0m
     > cd $scream_repo/components/eamxx
-    > ./scripts/{0} -m melvin --baseline-dir=PATH_TO_BASELINES
+    > ./scripts/{0} -m mappy --baseline-dir=PATH_TO_BASELINES
 
-    \033[1;32m# Run only the dbg test on current machine with default behavior otherwise\033[0m
+    \033[1;32m# Run all tests on current machine with default behavior except using local baselines \033[0m
     > cd $scream_repo/components/eamxx
-    > ./scripts/{0} -m melvin -t dbg
+    > ./scripts/{0} -m mappy --baseline-dir=LOCAL
```
bartgol, line 66: Should we support LOCAL at all? I'm thiking that, now that the user must explicitly specify `-g` to generate baselines, we should also ask for a folder where to put them, no? It would be one less thing to keep track/support/test in TAS. We'd only keep AUTO as special value for `-b`...

components/eamxx/scripts/test-all-scream
```diff
@@ -133,9 +147,6 @@ OR
                         help="Set the testing level. Defaults to medium unless the test is cov or mem_check"
                         "(short is default in those cases).")
 
-    parser.add_argument("--force-baseline-regen", action="store_true",
```
bartgol, line 136: Does this mean that if the user passes `-g`, we update the baselines regardless of how the current SHA compares to the baselines' SHA?

components/eamxx/scripts/test_all_scream.py
```diff
@@ -831,14 +716,11 @@ def baselines_to_be_generated(self):
         """
         Return list of baselines to generate. Baselines need to be generated if
          - they are missing
-         - they are expired and we asked to update expired baselines
```
bartgol, line 834: How do we force new baselines now that you removed `--force-baseline-regen`? I think we should either
- keep `--force-baseline-regen`, which generates baselines even if not missing
- always generate baselines, even if not missing.

I think the former may be more robust against accidental typos (like `-g` instead of `-h`). But the latter reduces the code in TAS... I'm ok either way, but I would like to have to manually rm baselines if I want them regenerated...

bartgol (commented):
One thing is not clear/right, I think. If baselines are not missing, TAS will not generate, even if `-g` is used (I think). I don't think that's what we want.

components/eamxx/scripts/test-all-scream
jgfouca: I think LOCAL is kind of useful. If you want a temporary stash of baselines, it saves a little bit of typing.

components/eamxx/scripts/test-all-scream
jgfouca: Yes.

components/eamxx/scripts/test_all_scream.py
jgfouca: Good catch, this was a mistake.

jgfouca:
@bartgol , fixed.

components/eamxx/scripts/test_all_scream.py
```diff
@@ -1,8 +1,6 @@
 from utils import run_cmd, run_cmd_no_fail, expect, check_minimum_python_version, ensure_psutil, \
     SharedArea, safe_copy
-from git_utils import get_current_head, get_current_commit, get_current_branch, is_repo_clean, \
```
bartgol, line 3: I wonder if we should rm the git utils we no longer use, and keep the code base leaner...

bartgol (approved):

bartgol (approved):

PR #2870, opened 2024-06-18
Option to do ML correction on temperature only 

elynnwu:
This PR adds the option to do ML correction on temperature only.

### DIFF
--- a/components/eamxx/src/physics/ml_correction/eamxx_ml_correction_process_interface.cpp
+++ b/components/eamxx/src/physics/ml_correction/eamxx_ml_correction_process_interface.cpp
@@ -17,9 +17,6 @@ MLCorrection::MLCorrection(const ekat::Comm &comm,
   m_ML_model_path_sfc_fluxes = m_params.get<std::string>("ML_model_path_sfc_fluxes");
   m_fields_ml_output_variables = m_params.get<std::vector<std::string>>("ML_output_fields");
   m_ML_correction_unit_test = m_params.get<bool>("ML_correction_unit_test");
-  EKAT_ASSERT_MSG(!(m_ML_model_path_tq != "None" && m_ML_model_path_t_only != "None"),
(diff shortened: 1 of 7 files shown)


components/eamxx/src/physics/ml_correction/eamxx_ml_correction_process_interface.cpp
```diff
@@ -12,10 +12,14 @@ MLCorrection::MLCorrection(const ekat::Comm &comm,
                            const ekat::ParameterList &params)
     : AtmosphereProcess(comm, params) {
   m_ML_model_path_tq = m_params.get<std::string>("ML_model_path_tq");
+  m_ML_model_path_t_only = m_params.get<std::string>("ML_model_path_temperature");
   m_ML_model_path_uv = m_params.get<std::string>("ML_model_path_uv");
   m_ML_model_path_sfc_fluxes = m_params.get<std::string>("ML_model_path_sfc_fluxes");
   m_fields_ml_output_variables = m_params.get<std::vector<std::string>>("ML_output_fields");
   m_ML_correction_unit_test = m_params.get<bool>("ML_correction_unit_test");
+  EKAT_ASSERT_MSG(!(m_ML_model_path_tq != "None" && m_ML_model_path_t_only != "None"),
```
elynnwu, line 20: For some reason, I can't trigger this error in the full run but the same logic worked in the unit test. Maybe there's a better to handle this?

elynnwu:
Confirmed that we the model check works in python and throws an exception.
```
19: Program received signal SIGABRT: Process abort signal.
19:
19: Backtrace for this error:
12: terminate called after throwing an instance of 'pybind11::error_already_set'
12:   what():  ValueError: Only one of ML_model_path_tq,             ML_model_path_temperature can be specified
```

frodre (approved):
Thanks, LGTM

PR #2875, opened 2024-06-21
nvidiagpu_pm-gpu.cmake: Turn on correct kokkos settings

jgfouca:


bartgol:
Since he's the POC for the machine, I'll let @ndkeen review it, to avoid triggering automerge before he can see this.

ndkeen:
i was thinking something more like:
```
string(APPEND CONFIG_ARGS " --host=cray")
set(USE_CUDA "TRUE")
string(APPEND CPPDEFS " -DGPU -DMPAS_OPENACC")
if (COMP_NAME STREQUAL gptl)
  string(APPEND CPPDEFS " -DHAVE_NANOTIME -DBIT64 -DHAVE_SLASHPROC -DHAVE_GETTIMEOFDAY")
endif()
string(APPEND CPPDEFS " -DTHRUST_IGNORE_CUB_VERSION_CHECK")
string(APPEND CMAKE_CUDA_FLAGS " -ccbin CC -O2 -arch sm_80 --use_fast_math")
string(APPEND CMAKE_Fortran_FLAGS " -acc -gpu=cc80 -Minfo=accel")
string(APPEND CMAKE_EXE_LINKER_FLAGS " -acc -gpu=cc80 -Minfo=accel")

if (compile_threaded)
  string(APPEND KOKKOS_OPTIONS " -DKokkos_ARCH_AMPERE80=On -DKokkos_ENABLE_CUDA=On -DKokkos_ENABLE_CUDA_LAMBDA=On -DKokkos_ENABLE_IMPL_CUDA_MALLOC_ASYNC=Off")
else()
  string(APPEND KOKKOS_OPTIONS " -DKokkos_ARCH_AMPERE80=On -DKokkos_ENABLE_CUDA=On -DKokkos_ENABLE_CUDA_LAMBDA=On -DKokkos_ENABLE_SERIAL=ON -DKokkos_ENABLE_OPENMP=Off -DKokkos_ENABLE_IMPL_CUDA_MALLOC_ASYNC=Off")
endif()

set(SCC "cc")
set(SCXX "CC")
set(SFC "ftn")
```

but as I'm still having build issues, I wanted to address that first as it may involve changes here

bartgol:
@ndkeen did you mean to add `Kokkos_ENABLE_OPENMP=ON` in the `compile_threaded` case? Also, while it's ok to add `Kokkos_ENABLE_Serial=ON`, notice that Kokkos _always_ enables Serial for GPU builds, to ensure that a host space is available.

That said, I think we should think about what we want to do with kokkos in a GPU build when `compile_threaded=ON`. As of now, if building on a GPU machine (with a GPU compiler), we most likely only need the GPU backend, even if `compile_threaded=ON`. It is possible that, down the road, we may have other components using Kokkos::OpenMP while ATM uses Kokkos::Cuda. In that case, yes, we'd need both spaces. However, as of now, I think it's ok (better?) to just enable the Cuda backend in Kokkos.

bartgol:
As a follow-up, enabling 2+ backends (not counting Serial) is delicate for the following reason: some codes may just pick the "default" kokkos space. However, if you enable Cuda and OpenMP, the "default" space in Kokkos is Cuda. So if we want 2+ components to use different kokkos backends, we need to make sure they do not rely on the "default" kokkos device, but instead explicitly ask for a specific one...

ndkeen:
Yea I now realize we probably dont want openmp enable for kokkos at all (we don't have it on for gnugpu).
However, I'm still unable to build. We can make a change here, but might be better to wait until we can build with nvidia.

bartgol:
Ok, I added the WIP label, so we can sort the build issue out first.

ndkeen:
I'm still making progress, but it is already involving upgrading nvidia/cuda versions and removing some flags that are thrown in at various places.

We can either close this and make a general "building with nvidagpu" issue.
Or, as we are pretty sure of some changes needed in this file, make some improvements here?
The file is otherwise not used...

ndkeen:
At one point, I was able to build/run scream with gnu and nvidia for GPU. But as we did not test it often, things have devolved.  We don't test it now.  The test with nvidiagpu you noted above is _only_ testing MPAS cases (and they use ACC).  It's true the exes built with nvidia typically do not perform better than with GNU. However, it improves the quality of our code to test with as many compilers as we can (within effort constraints).  We can close this PR and make an issue with branch.

ndkeen:
Closing with pointer to issue here https://github.com/E3SM-Project/scream/issues/2879

PR #2876, opened 2024-06-21
Rework rrtmgp interface based on recent improvements to rrtmgp standalone interface

jgfouca:
rrtmgp standalone underwent a major re-work of it's interface for the Kokkos-enabled API, making all View arguments templates:
https://github.com/E3SM-Project/rte-rrtmgp/pull/34

This removes the assumptions about the incoming Views that forced us to make copies in our rrtmgp interface.

This PR removes many of the View copies in rrtmgp interface. Also, some of the assumptions about views that rrtmgp was making were also being made in our code, especially in scream_rrtmgp_interface.hpp, so I had to make similar to changes to what was done in rrtmgp standalone to make things more generic.

A lot of things can be removed from RRTMGPRadiation::Buffer once we drop YAKL support.

If `-DSCREAM_RRTMGP_ENABLE_KOKKOS=Off` (this is the default setting), then this PR should not have any impact.


components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
```diff
@@ -955,67 +976,67 @@ void RRTMGPRadiation::run_impl (const double dt) {
...
-      auto sfc_flux_dif_vis_k = subview_1dk(m_buffer.sfc_flux_dif_vis_k);
-      auto sfc_flux_dif_nir_k = subview_1dk(m_buffer.sfc_flux_dif_nir_k);
-      auto aero_tau_sw_k     = subview_3dk(m_buffer.aero_tau_sw_k);
-      auto aero_ssa_sw_k     = subview_3dk(m_buffer.aero_ssa_sw_k);
-      auto aero_g_sw_k       = subview_3dk(m_buffer.aero_g_sw_k);
-      auto aero_tau_lw_k     = subview_3dk(m_buffer.aero_tau_lw_k);
+      // Note, ncol will not necessary be m_col_chunk_size because the number of cols
+      // will not always be evenly divided by m_col_chunk_size. In most cases, the
+      // extra space will not cause any problems, but it does sometimes.
+      auto p_lay_k           = subview_2dkc(d_pmid, m_nlay);
```
jgfouca, line 982: This code is worth looking at since it shows how well we were able to leverage the existing fields.

components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
```diff
@@ -955,67 +976,67 @@ void RRTMGPRadiation::run_impl (const double dt) {
...
+      auto t_lay_k           = subview_2dkc(d_tmid, m_nlay);
+      auto p_lev_k           = subview_2dkc(d_pint, m_nlay+1);
+      auto p_del_k           = subview_2dkc(d_pdel, m_nlay);
+      auto sfc_alb_dir_k     = m_buffer.sfc_alb_dir_k;
```
jgfouca, line 986: In some cases, like here, we continue to use the buffer memory because these items are needed as temporary workspace; there is no corresponding field.

components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
```diff
@@ -955,67 +976,67 @@ void RRTMGPRadiation::run_impl (const double dt) {
...
+      auto sfc_alb_dif_k     = m_buffer.sfc_alb_dif_k;
+      auto sfc_alb_dir_vis_k = subview_1dkc(d_sfc_alb_dir_vis);
+      auto sfc_alb_dir_nir_k = subview_1dkc(d_sfc_alb_dir_nir);
+      auto sfc_alb_dif_vis_k = subview_1dkc(d_sfc_alb_dif_vis);
+      auto sfc_alb_dif_nir_k = subview_1dkc(d_sfc_alb_dif_nir);
+      auto qc_k              = subview_2dkc(d_qc, m_nlay);
+      auto nc_k              = subview_2dkc(d_nc, m_nlay);
+      auto qi_k              = subview_2dkc(d_qi, m_nlay);
+      auto cldfrac_tot_k     = m_buffer.cldfrac_tot_k;
```
jgfouca, line 995: cldfrac_tot has special handling if `do_subcol_sampling` is off, so we have to use the buffer.

components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
```diff
@@ -955,67 +976,67 @@ void RRTMGPRadiation::run_impl (const double dt) {
...
+      auto sw_clnsky_flux_up_k      = subview_2dk(d_sw_clnsky_flux_up, m_nlay+1);
+      auto sw_clnsky_flux_dn_k      = subview_2dk(d_sw_clnsky_flux_dn, m_nlay+1);
+      auto sw_clnsky_flux_dn_dir_k  = subview_2dk(d_sw_clnsky_flux_dn_dir, m_nlay+1);
+      auto lw_clnclrsky_flux_up_k   = subview_2dk(d_lw_clnclrsky_flux_up, m_nlay+1);
+      auto lw_clnclrsky_flux_dn_k   = subview_2dk(d_lw_clnclrsky_flux_dn, m_nlay+1);
+      auto lw_clrsky_flux_up_k      = subview_2dk(d_lw_clrsky_flux_up, m_nlay+1);
+      auto lw_clrsky_flux_dn_k      = subview_2dk(d_lw_clrsky_flux_dn, m_nlay+1);
+      auto lw_clnsky_flux_up_k      = subview_2dk(d_lw_clnsky_flux_up, m_nlay+1);
+      auto lw_clnsky_flux_dn_k      = subview_2dk(d_lw_clnsky_flux_dn, m_nlay+1);
+      auto sw_bnd_flux_up_k  = m_buffer.sw_bnd_flux_up_k;
```
jgfouca, line 1018: This group are all needed as temporary workspaces.

components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
```diff
@@ -955,67 +976,67 @@ void RRTMGPRadiation::run_impl (const double dt) {
...
+      auto sw_bnd_flux_dn_k  = m_buffer.sw_bnd_flux_dn_k;
+      auto sw_bnd_flux_dir_k = m_buffer.sw_bnd_flux_dir_k;
+      auto sw_bnd_flux_dif_k = m_buffer.sw_bnd_flux_dif_k;
+      auto lw_bnd_flux_up_k  = m_buffer.lw_bnd_flux_up_k;
+      auto lw_bnd_flux_dn_k  = m_buffer.lw_bnd_flux_dn_k;
+      auto sfc_flux_dir_vis_k = subview_1dk(d_sfc_flux_dir_vis);
+      auto sfc_flux_dir_nir_k = subview_1dk(d_sfc_flux_dir_nir);
+      auto sfc_flux_dif_vis_k = subview_1dk(d_sfc_flux_dif_vis);
+      auto sfc_flux_dif_nir_k = subview_1dk(d_sfc_flux_dif_nir);
+      auto aero_tau_sw_k     = m_buffer.aero_tau_sw_k;
```
jgfouca, line 1028: These 4 are unfortunately needed because RRTMGP expects ordering (col,lay,bnd) but the FM keeps things in (col,bnd,lay) order.

components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
```diff
@@ -955,67 +976,67 @@ void RRTMGPRadiation::run_impl (const double dt) {
...
+      auto aero_ssa_sw_k     = m_buffer.aero_ssa_sw_k;
+      auto aero_g_sw_k       = m_buffer.aero_g_sw_k;
+      auto aero_tau_lw_k     = m_buffer.aero_tau_lw_k;
       auto cld_tau_sw_bnd_k  = subview_3dk(m_buffer.cld_tau_sw_bnd_k);
```
jgfouca, line 1032: These 4 3d views are needed as temporary workspaces.

components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
```diff
@@ -1658,34 +1642,8 @@ void RRTMGPRadiation::run_impl (const double dt) {
       Kokkos::parallel_for(policy, KOKKOS_LAMBDA(const MemberType& team) {
         const int i = team.league_rank();
         const int icol = i + beg;
-        d_sfc_flux_dir_nir(icol) = sfc_flux_dir_nir_k(i);
```
jgfouca, line 1661: You can see here most of the output copies are gone.

components/eamxx/src/physics/rrtmgp/rrtmgp_test_utils.cpp
```diff
@@ -46,28 +38,7 @@ bool all_close(real2d &arr1, real2d &arr2, double tolerance) {
   }
   return true;
 }
-#endif
-#ifdef RRTMGP_ENABLE_KOKKOS
```
jgfouca, line 50: Templates force us to move a lot of this stuff to the header.

components/eamxx/src/physics/rrtmgp/rrtmgp_utils.hpp
```diff
@@ -98,28 +99,31 @@ bool check_range(T x, Real xmin, Real xmax, std::string msg, std::ostream& out=s
...
+bool check_range_k(T x, typename T::const_value_type xmin, typename T::const_value_type xmax,
+                   std::string msg, std::ostream& out=std::cout) {
   bool pass = true;
   auto _xmin = conv::minval(x);
   auto _xmax = conv::maxval(x);
   if (_xmin < xmin or _xmax > xmax) {
     // How many outside range?
-    bool1dk bad_mask("bad_mask", x.size());
-    Kokkos::parallel_for(x.size(), KOKKOS_LAMBDA (int i) {
-      if (x.data()[i] < xmin or x.data()[i] > xmax) {
```
jgfouca, line 121: When you have subviews, it's no longer safe to use data(). If you do, you'll end up iterating over items that aren't in the subview.

components/eamxx/src/physics/rrtmgp/scream_rrtmgp_interface.cpp
```diff
@@ -106,36 +79,7 @@ OpticalProps2str get_cloud_optics_sw(
   // Return optics
   return clouds;
 }
-#endif
-#ifdef RRTMGP_ENABLE_KOKKOS
```
jgfouca, line 110: Tons of code had to be moved to the header due to templating unfortunately.

components/eamxx/src/physics/rrtmgp/scream_rrtmgp_interface.hpp
```diff
@@ -303,10 +144,1035 @@ void mixing_ratio_to_cloud_mass(
...
+
+int get_wavelength_index(OpticalProps &kdist, double wavelength);
+int get_wavelength_index_sw(double wavelength);
+int get_wavelength_index_lw(double wavelength);
+#endif // RRTMGP_ENABLE_YAKL
+
+// New interface for Kokkos and flexible types
 #ifdef RRTMGP_ENABLE_KOKKOS
+template <typename RealT=Real, typename LayoutT=Kokkos::LayoutRight, typename DeviceT=DefaultDevice>
+struct rrtmgp_interface {
```
jgfouca, line 163: Wrap the API in a struct so that we can define and reuse common types.

components/eamxx/src/physics/rrtmgp/scream_rrtmgp_interface.hpp
```diff
@@ -303,10 +144,1035 @@ void mixing_ratio_to_cloud_mass(
...
+  pool_t::init(2e6 * (float(my_size_ref) / base_ref));
+
+  // We are now initialized!
+  initialized_k = true;
+}
+
+/*
+ * Compute band-by-band surface albedos from broadband albedos.
+ */
+template <typename SfcAlbDirVisT, typename SfcAlbDirNirT, typename SfcAlbDifVisT,
```
jgfouca, line 241: Since we already know the device and layout, it's possible these views didn't have to be templates.

components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
```diff
@@ -1308,12 +1293,12 @@ void RRTMGPRadiation::run_impl (const double dt) {
...
-      scream::rrtmgp::mixing_ratio_to_cloud_mass(qi, cldfrac_tot, p_del, iwp);
+      rrtmgp::mixing_ratio_to_cloud_mass(qc, cldfrac_tot, p_del, lwp);
+      rrtmgp::mixing_ratio_to_cloud_mass(qi, cldfrac_tot, p_del, iwp);
 #endif
 #ifdef RRTMGP_ENABLE_KOKKOS
-      scream::rrtmgp::mixing_ratio_to_cloud_mass(qc_k, cldfrac_tot_k, p_del_k, lwp_k);
-      scream::rrtmgp::mixing_ratio_to_cloud_mass(qi_k, cldfrac_tot_k, p_del_k, iwp_k);
+      interface_t::mixing_ratio_to_cloud_mass(qc_k, cldfrac_tot_k, p_del_k, lwp_k);
+      interface_t::mixing_ratio_to_cloud_mass(qi_k, cldfrac_tot_k, p_del_k, iwp_k);
       COMPARE_ALL_WRAP(std::vector<real2d>({lwp, iwp}),
```
bartgol, line 1302: Do we need to run all these checks every time? There are ~10 of these in this file. This seems like a unit-test kind of check. At the very least, can we disable these checks in RELEASE builds? I looked at the rrtmgp submodule, and it seems like this check is always active..

bartgol (commented):
I'm confused, do we still have both yakl and kokkos arrays? I see we still run the COMPARE_WRAP macro, so I think the answer is yes?

jgfouca:
@bartgol ,

> I'm confused, do we still have both yakl and kokkos arrays? I see we still run the COMPARE_WRAP macro, so I think the answer is yes?

Yes, we have not yet dropped support for YAKL so the YAKL arrays are still there (if SCREAM_RRTMGP_ENABLE_YAKL is On). Kokkos is used if SCREAM_RRTMGP_ENABLE_KOKKOS is On. If both are on, then we do both and can compare results.

components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
jgfouca: By check, do you mean COMPARE_ALL? That only does something in a very unusual configuration (both YAKL and Kokkos are on) in which performance would already be awful since we doing everything twice.

bartgol:
> @bartgol ,
> 
> > I'm confused, do we still have both yakl and kokkos arrays? I see we still run the COMPARE_WRAP macro, so I think the answer is yes?
> 
> Yes, we have not yet dropped support for YAKL so the YAKL arrays are still there (if SCREAM_RRTMGP_ENABLE_YAKL is On). Kokkos is used if SCREAM_RRTMGP_ENABLE_KOKKOS is On. If both are on, then we do both and can compare results.

Do you think we should keep the cmp check on even in RELEASE mode?

jgfouca:
> > @bartgol ,
> > > I'm confused, do we still have both yakl and kokkos arrays? I see we still run the COMPARE_WRAP macro, so I think the answer is yes?
> > 
> > 
> > Yes, we have not yet dropped support for YAKL so the YAKL arrays are still there (if SCREAM_RRTMGP_ENABLE_YAKL is On). Kokkos is used if SCREAM_RRTMGP_ENABLE_KOKKOS is On. If both are on, then we do both and can compare results.
> 
> Do you think we should keep the cmp check on even in RELEASE mode?

By check, do you mean COMPARE_ALL? That only does something in a very unusual configuration (both YAKL and Kokkos are on) in which performance would already be awful since we doing everything twice.

bartgol:
Ah, I see. Can we add (or do we already have) a CMake check to ensure we don't accidentally turn both on in a CIME build?

bartgol (approved):

bartgol:
@jgfouca I removed automerge in case you want to wait for others to review.

components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
brhillman: Is this not true for the cld_ ones below?

components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
```diff
@@ -1384,11 +1369,11 @@ void RRTMGPRadiation::run_impl (const double dt) {
...
 #endif
 #ifdef RRTMGP_ENABLE_KOKKOS
-      rrtmgp::rrtmgp_main(
+      interface_t::rrtmgp_main(
         ncol, m_nlay,
-        p_lay_k, t_lay_k, p_lev_k, t_lev_k,
+        p_lay_k, t_lay_k, p_lev_k, d_tint,
         m_gas_concs_k,
-        sfc_alb_dir_k, sfc_alb_dif_k, mu0_k,
+        sfc_alb_dir_k, sfc_alb_dif_k, d_mu0,
```
brhillman, line 1376: I'm a little confused about use of `d_mu0` here. It looks like `d_mu0` refers to the m_buffer field cosine_zenith (`auto d_mu0 = m_buffer.cosine_zenith;`), but I don't see it populated anywhere above. Should this be another subview slice for this chunk?

brhillman: Okay never mind, I think I found it. But I think maybe use of `d_mu0` for naming is a little confusing here. I had originally taken the `d_` prefix convention from some other process (don't remember which) to refer to the device views that come directly from the FM. In this case though, I think `d_mu0` is being treated like the other subviews that hold data for just the current "chunk" of ncol. Should we rename this to be consistent with how the others are used? Or am I missing something here?

components/eamxx/src/physics/rrtmgp/rrtmgp_utils.hpp
brhillman: That's fine, but does this still check over all elements in the subview? That was the only reason to use data before, to flatten the view and make it easy to check all values (i.e., over ncol, nlay, nbnd, etc).

brhillman: The use of `extent(0)` makes me think not, but this is a little out of my wheelhouse.

brhillman: Okay, it looks like you implemented separate templates for rank = 1, 2, 3 below. Is there a way to avoid this? Seems like a lot of duplicated code.

components/eamxx/src/physics/rrtmgp/scream_rrtmgp_interface.hpp
brhillman: Aren't these all the same type? Why template on each individual type if in use they are all the same? I.e., why couldn't we do:
```
template<typename T, typename S> static void compute_band_by_band_surface_albedos(
  const int incol, const int nswbands,
  T & sfc_alb_dir_vis, T &sfc_alb_dir_nir, T &sfc_alb_dif_vis, T &sfc_alb_dif_nir,
  S &sfc_alb_dir, S &sfc_alb_dif)
```

components/eamxx/src/physics/rrtmgp/scream_rrtmgp_interface.hpp
```diff
@@ -303,10 +144,1035 @@ void mixing_ratio_to_cloud_mass(
...
+      sfc_alb_dir(icol,ibnd) = 0.5*(sfc_alb_dir_vis(icol) + sfc_alb_dir_nir(icol));
+      sfc_alb_dif(icol,ibnd) = 0.5*(sfc_alb_dif_vis(icol) + sfc_alb_dif_nir(icol));
+    }
+  });
+}
+
+/*
+ * Compute broadband visible/UV and near-infrared surface fluxes.
+ */
+template <typename SwBndFluxDirT, typename SwBndFluxDifT, typename SfcFluxDirVisT,
```
brhillman, line 292: Same comment here as above, templating on each view seems overkill. It is perfectly reasonable to assume/require sw_bnd_flux_dir and sw_bnd_flux_dif to have the same type, sfc_flux_dir_vs, sfc_flux_dir_nir, sfc_flux_dif_vis, sfc_flux_dif_nir to have the same type.

components/eamxx/src/physics/rrtmgp/scream_rrtmgp_interface.hpp
```diff
@@ -303,10 +144,1035 @@ void mixing_ratio_to_cloud_mass(
...
+    }
+  });
+}
+
+/*
+ * Main driver code to run RRTMGP.
+ * The input logger is in charge of outputing info to
+ * screen and/or to file (or neither), depending on how it was set up.
+ */
+template <typename PlayT, typename TlayT, typename PlevT, typename TlevT,
```
brhillman, line 352: I guess you will probably answer this in the above comments I made, but again here we shouldn't have to assume that each input can have different types. The only types we should need to template on are those with different ranks, and there should only be maybe 4-5 different layouts I believe.

components/eamxx/src/physics/rrtmgp/scream_rrtmgp_interface.hpp
```diff
@@ -303,10 +144,1035 @@ void mixing_ratio_to_cloud_mass(
...
+  k_dist_lw_k.finalize();
+  cloud_optics_sw_k.finalize(); //~CloudOptics();
+  cloud_optics_lw_k.finalize(); //~CloudOptics();
+  pool_t::finalize();
+}
+
+/*
+ * Shortwave driver (called by rrtmgp_main)
+ */
+template <typename PlayT, typename TlayT, typename PlevT, typename TlevT,
```
brhillman, line 552: See above. Redundant types.

components/eamxx/src/physics/rrtmgp/scream_rrtmgp_interface.hpp
```diff
@@ -303,10 +144,1035 @@ void mixing_ratio_to_cloud_mass(
...
+      clnsky_flux_dn    (icol,ilev) = flux_dn_day    (iday,ilev);
+      clnsky_flux_dn_dir(icol,ilev) = flux_dn_dir_day(iday,ilev);
+    });
+  }
+}
+
+/*
+ * Longwave driver (called by rrtmgp_main)
+ */
+template <typename PlayT, typename TlayT, typename PlevT, typename TlevT>
```
brhillman, line 814: See above; redundant types

components/eamxx/src/physics/rrtmgp/scream_rrtmgp_interface.hpp
```diff
@@ -303,10 +144,1035 @@ void mixing_ratio_to_cloud_mass(
...
+    for (int igpt = 0; igpt < ngpt; ++igpt) {
+      cld_area(icol) += subcol_mask(icol,igpt) * ngpt_inv;
+    }
+  });
+}
+
+/*
+ * Return select cloud-top diagnostics following AeroCom recommendation
+ */
+template <typename TmidT, typename PmidT, typename PdelT, typename ZdelT, typename QcT,
```
brhillman, line 1061: See above; redundant types

components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
jgfouca: It might be but those don't need to be initialized from field vars.

components/eamxx/src/physics/rrtmgp/eamxx_rrtmgp_process_interface.cpp
jgfouca: Any name is fine with me, I was just using the names that were already there. mu0_k was just a copy of d_mu0.

components/eamxx/src/physics/rrtmgp/rrtmgp_utils.hpp
jgfouca: The specializations are needed since we are no longer using the .data() approach that allowed us to treat all ranks of views the same. .data() assumes the view is contiguous and that is not true for some kinds of subviews.

jgfouca:
@bartgol , @brhillman , I think we need to discuss the issue of how to handle view arguments in a way that doesn't limit us in the future. I agree with Ben that just making them all separate templates is kind of ugly (I would know, I had to type all that crap!). It also obfuscates the API a bit to code reader since the rank and scalar type is no longer obvious.

The rationale was that I wanted to avoid the situation we hit with rrtmgp standalone where the API made so many assumptions about the views that it forced us to make copies. In scream, I wrapped all the rrtmgp interfaces in a struct that is templated on scalar, device, and layout, so it's possible that we have enough knowledge to get rid of the templates.

A view type has 4 components:
1) data_type, e.g. real***
2) layout (right vs. left)
3) device/memory_space (host vs device)
4) traits, e.g managed vs unmanaged

Thanks to the wrapping struct, 1-3 are already taken care of, which leaves (4) and this is the sticking point. The most likely trait that is going to vary is managed vs unmanaged. I wrote a simple program:

```
using view_t = Kokkos::View<int**>;
using uview_t = Unmanaged<view_t>;

void func1(const view_t& arg)
{
  for (size_t i = 0; i < arg.size(); ++i) {
    std::cout << arg.data()[i] << " ";
  }
  std::cout << std::endl;
}

void func2(const uview_t& arg)
{
  for (size_t i = 0; i < arg.size(); ++i) {
    std::cout << arg.data()[i] << " ";
  }
  std::cout << std::endl;
}

int main(int argc, char** argv)
{
  Kokkos::initialize(argc, argv); {
    view_t data("data", 2, 4);
    uview_t data2(data.data(), 2, 4);
    Kokkos::deep_copy(data, 42);
    func1(data);
    func2(data);
    func1(data2);
    func2(data2);
  }
  Kokkos::finalize();
  return 0;
}

```

^^ This program compiles and runs, so it appears that a managed view is convertible to an unmanaged view reference argument and vice versa. So that's good. I also tried atomic vs. non-atomic, and those were also convertible, so maybe the templates aren't needed?

jgfouca:
@brhillman , I removed the excessive templating in scream_rrtmgp_interface.hpp

bartgol (approved):

PR #2877, opened 2024-06-24
Add timers on a per-yaml-file-basis

bartgol:
On top of global IO timers (adding up all equivalent timers from all output streams), this PR adds timers that show the overall time spent on each output stream. These additional timers are not split internally (e.g., no per-yaml-file `update_snapshot_tally` timer); they simply show what's the overall run time spent on an output stream. I tested it duplicating the output of the p3_standalone test, and this is what I see in the timings file:

```
EAMxx::IO::output.yaml                      -          1        1 6.000000e+00   3.732990e-01     0.373 (     0      0)     0.373 (     0      0)
...
EAMxx::IO::output2.yaml                     -          1        1 6.000000e+00   2.916770e-01     0.292 (     0      0)     0.292 (     0      0)
```


AaronDonahue (approved):
Looks good.  Thanks for putting this in, it should be helpful understanding where we spend time in IO

PR #2878, opened 2024-06-25
Fix how we forward iotype to scorpio routines

bartgol:
So far, we were simply casting our IOType enum value to the underlying int. This was buggy, since the entries of IOType were ordered differently from the scorpio PIO_IOTYPE enum, so that, e.g., our IOType::PnetCDF was mapped to PIO_IOTYPE_NETCDF4C.

This PR adds a small function to map scream's enum into the pio one explicitly, withouth relying on the internal ordering of the enum values.

### DIFF
--- a/components/eamxx/src/share/io/scream_scorpio_interface.cpp
+++ b/components/eamxx/src/share/io/scream_scorpio_interface.cpp
@@ -207,6 +207,21 @@ size_t dtype_size (const std::string& dtype) {
   return 0;
 }
 
+int pio_iotype (IOType iotype) {
+  int iotype_int;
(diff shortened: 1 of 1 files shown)


brhillman (approved):

bartgol:
The fail on mappy was an interesting internal compiler error. I am going to blame cosmic rays, unless it fails again. I really don't want to chase ICEs, we've been blessed by a long period of not having them...

ndkeen:
fwiw, i checked out this branch and ran some ne256 cases on frontier -- verified that I see expected performance with all yaml streams using `iotype:pnetcdf`.

PR #2880, opened 2024-06-25
Improve grids handling in pyeamxx

bartgol:
Summary:

- Make user create a GridsManager rather than a single grid, to prepare for more complex use cases (e.g., dyn-phys use cases)
- Create a PySession singleton to store some data (so far, the comm and grids manager)
- Misc fixes to p3_standalone_py test

With these mods, we should now be able to load lat/lon from an NC file containing them. So the user would simply do something like this:
```python
my_latlon_file = "/path/to/my/latlon/file.nc"
pyeamxx.create_grids_manager(num_cols,num_levs,my_latlon_file)
```
And the C++ atm procs will then have access to lat/lon from the grid.

Note: the next step is to create a "single-column" grids manager, with possibility of specifying the lat/lon of such column. This feature is in the works.


bartgol:
> All good, but can you finish your thought in the PR description (or give an example)? 😝

Whoops! Done.

PR #2881, opened 2024-06-26
pyeamxx: allow to set atm procs parameters from python

bartgol:
This PR adds some interfaces to PyParamList, which allow to change the value of a parameter list from python. The [p3_standalone_py](https://github.com/E3SM-Project/scream/blob/54c1ece88d58a427d951dfef7db723344698e1ca/components/eamxx/tests/python/pyp3/p3_standalone_py#L45) script shows an example of how this is done.

### DIFF
--- a/components/eamxx/src/python/pyparamlist.hpp
+++ b/components/eamxx/src/python/pyparamlist.hpp
@@ -7,19 +7,54 @@
 #include <pybind11/numpy.h>
 #include <pybind11/stl.h>
 
+#include <functional>
+
 namespace scream {
 
 struct PyParamList {
   ekat::ParameterList pl;
+  std::reference_wrapper<ekat::ParameterList> pl_ref;
 
-  PyParamList(const pybind11::dict& d) {
+  PyParamList (ekat::ParameterList& src)
+   : pl_ref(src)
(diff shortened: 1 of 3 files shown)


components/eamxx/src/python/pyparamlist.hpp
```diff
@@ -7,19 +7,54 @@
...
 
-  PyParamList(const pybind11::dict& d, const std::string& name) {
-    pl.rename(name);
-    parse_dict(d,pl);
+  PyParamList sublist (const std::string& name) {
+    PyParamList spl(pl.sublist(name));
+    return spl;
+  }
+
+  int get_int (const std::string& name) const {
```
bartgol, line 38: Template methods where the template arg cannot be deduced from inputs are not py-interoperable. That is, there is no equivalent of `auto blah = get<int>("my_param");` that you can call from python. Hence, the templated overloads need to have different names.

bartgol:
Yeah, I think being able to change one may save some time over a "recreate the atm proc from scratch" approach, in case one is running MANY ensembles. If the atm proc does stuff during construction time that depends on the parameters, however, then this doesn't help much.

PR #2882, opened 2024-06-26
Add scorpio reader to fill single-column fields with closest column in input file

bartgol:
This PR takes what was done in IOP, and makes it available as a generic tool. The reason for this is that I'm planning to use it in the pyeamxx interface, to allow running parametrizations multiple times on a single column.

@tcclevenger for now this is enough for my pyeamxx needs. However, I think we could make IOP use this class, perhaps after polishing it to accommodate any peculiar IOP needs.

### DIFF
--- a/components/eamxx/src/share/io/scorpio_input.hpp
+++ b/components/eamxx/src/share/io/scorpio_input.hpp
@@ -1,10 +1,8 @@
 #ifndef SCREAM_SCORPIO_INPUT_HPP
 #define SCREAM_SCORPIO_INPUT_HPP
 
-#include "share/io/scream_scorpio_interface.hpp"
 #include "share/field/field_manager.hpp"
 #include "share/grid/abstract_grid.hpp"
-#include "share/grid/grids_manager.hpp"
 
 #include "ekat/ekat_parameter_list.hpp"
 #include "ekat/logging/ekat_logger.hpp"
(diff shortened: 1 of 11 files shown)


components/eamxx/src/share/io/scorpio_input.hpp
```diff
@@ -1,10 +1,8 @@
 #ifndef SCREAM_SCORPIO_INPUT_HPP
 #define SCREAM_SCORPIO_INPUT_HPP
 
-#include "share/io/scream_scorpio_interface.hpp"
```
bartgol, line 4: This header does not need scorpio interfaces. Removing this caused compiler errors, so I had to add it in a few cpp files.

components/eamxx/src/share/io/scorpio_scm_input.hpp
```diff
@@ -0,0 +1,71 @@
...
+#include <ekat/logging/ekat_logger.hpp>
+
+namespace scream
+{
+
+// Similar to AtmosphereInput, but reads in a single column from
+// a file with N columns. A few assumptions:
+//  - lat and lon variables are present in the file
+//  - fields have layout <COL [, ...]>
+class SCMInput
```
bartgol, line 15: I opted for SCMInput, since it should recall the acronym Single Column Model (SCM), used elsewhere. I thought about SCInput (Single Column Input), but SC was also very close to Surface Coupling. I'm open to change this to a better name though.

tcclevenger (approved):
This looks good to me. I'll work to incorperate into IOP where maybe some changes are needed, but I think we'll only figure those out once we try and use it.

Issue #2883, opened 2024-06-28
Revert back to using string hex format for writing hashes


ndkeen:
After https://github.com/E3SM-Project/scream/pull/2871/, we now see hashes printed as overflowing integers like so:
```
  0: bfbhash>              0 977529120635437145 (Hommexx)
  0: bfbhash>             18 5740692760551287164 (Hommexx)
  0: bfbhash>             36 3532359390829707060 (Hommexx)
  0: bfbhash>             54 308155417158379321 (Hommexx)
  0: bfbhash>             72 2146756769357887193 (Hommexx)
  0: bfbhash>             90 -2728738275833850477 (Hommexx)
  0: bfbhash>            108 273843384767190905 (Hommexx)
  0: bfbhash>            126 4947377104589463742 (Hommexx)
  0: bfbhash>            144 -4517632968632715674 (Hommexx)
``` 

Seems better to go back to having them as string hashes:

```
0: bfbhash>              0 c08de33d90a54e3b (Hommexx)
0: bfbhash>              1 98fb7e0227daaf07 (Hommexx)
0: bfbhash>              2 242d7a55e6fc07b9 (Hommexx)
0: bfbhash>              3 f257e0ccdf3e98ba (Hommexx)
0: bfbhash>              4 3bae25ea0d4fe1a9 (Hommexx)
0: bfbhash>              5 d0fe4ef4356df255 (Hommexx)
0: bfbhash>              6  74f506e78c0b398 (Hommexx)
0: bfbhash>              7 72f8a0e629eae1b8 (Hommexx)
0: bfbhash>              8 4c8f609e3adaa810 (Hommexx)
0: bfbhash>              9 c83caacbdbcb91e8 (Hommexx)
0: bfbhash>             10 dc4e5191e151f956 (Hommexx)
```

Also, if we don't do this, makes it very difficult to compare hashes between cases run with previous checkouts.

Testing branch `ndk/scream/bfbhash-back-to-hex`

```
<     fprintf(stderr, "bfbhash> %14d %16lld (%s)\n",
<             timestamp().get_num_steps(), (long long int) gaccum, label.c_str());
---
>     fprintf(stderr, "bfbhash> %14d %16lx (%s)\n",
>             timestamp().get_num_steps(), gaccum, label.c_str());
```

PR #2884, opened 2024-06-28
Revert back to using string hex format for writing hashes

ndkeen:
Use string hex values for hash prints that are used to compare cases.

Fixes https://github.com/E3SM-Project/scream/issues/2883

[bfb]

### DIFF
--- a/components/eamxx/src/share/atm_process/atmosphere_process_hash.cpp
+++ b/components/eamxx/src/share/atm_process/atmosphere_process_hash.cpp
@@ -129,9 +129,9 @@ void AtmosphereProcess
   if (m_comm.am_i_root())
     for (int i = 0; i < nslot; ++i)
       if (show[i])
-        fprintf(stderr, "exxhash> %4d-%9.5f %1d %16lld (%s)\n",
+        fprintf(stderr, "exxhash> %4d-%9.5f %1d %16lx (%s)\n",
                 timestamp().get_year(), timestamp().frac_of_year_in_days(),
-                i, (long long int)gaccum[i], label.c_str());
+                i, gaccum[i], label.c_str());
 }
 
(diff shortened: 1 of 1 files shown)


ndkeen:
Tested with `ERP_Ln22.conusx4v1pg2_r05_oECv3.F2010-SCREAMv1-noAero.pm-gpu_gnugpu.scream-bfbhash--scream-output-preset-6`
and confirmed it passes, and is BFB with same test using checkout of June 18

components/eamxx/src/share/atm_process/atmosphere_process_hash.cpp
```diff
@@ -129,9 +129,9 @@ ::print_global_state_hash (const std::string& label, const bool in, const bool o
   if (m_comm.am_i_root())
     for (int i = 0; i < nslot; ++i)
       if (show[i])
-        fprintf(stderr, "exxhash> %4d-%9.5f %1d %16lld (%s)\n",
+        fprintf(stderr, "exxhash> %4d-%9.5f %1d %16lx (%s)\n",
```
bartgol, line 132: Nitpicky: I would use `%16llx` as format (notice the double ell). The hash is a 64 bit integer, so we want to be sure we are using a 64 bit format integer in the print statement, to avoid compiler warnings.

Yes, on most 64bit systems, `long` _is_ 64 bits, but technically it is still implementation dependent; so you _could_ find a system where `long` is 32 bits, causing warnings.

bartgol (changes requested):
I have a minor suggestion.

components/eamxx/src/share/atm_process/atmosphere_process_hash.cpp
ndkeen: Agreed.  I tested again with same test and verified identical behaviour.

bartgol (approved):

PR #2885, opened 2024-06-28
Enable pyeamxx testing in SP builds

bartgol:
We were not unit testing pyeamxx, so enable it in SP builds (which are faster, since there's no dynamics or rrtmgp).

~Also, add a documentation page for pyeamxx to our docs~ Removed docs commit, since we'll do it in another PR.


bartgol:
> Looks good.
> 
> Just curios: Is it struggling to find pybind11 in mappy/weaver? It often "just" finds it...

On mappy I had to pip install pybind11 and mpi4py, while on weaver it failed b/c of a legit code error.

PR #2886, opened 2024-07-02
Adds description for top_lev and adds its value for 128 levs

singhbalwinder:
The top level for the MAM4xx ACI codes is the level corresponding to the top of troposphere clouds. I have added this info to the namelist and added its value for 128 levels. Currently, these values are hardwired but we have an issue open to add code that will compute these levels during the runtime.


singhbalwinder:
Forgot to tag @susburrows. I thought the testing would take some time but just realized that it was skipped due to the nature of the changes.
Thanks for the quick reviews @bartgol and @mahf708 !

PR #2887, opened 2024-07-02
Add P3 vars to FM

tcclevenger:
Eventually MAM4xx will need these variables, so adding them to the FieldManager.


singhbalwinder (approved):
Looks great! Thanks @tcclevenger !

PR #2893, opened 2024-07-08
Update module versions on pm-gpu

ndkeen:
On pm-gpu, there are new default versions of several modules and this PR updates all of them. What we really want is the newer cudatoolkit version 12, but to get that version, need to also update the compiler (as NERSC made this a package update). This change updates to GNU version 12.

All of the tests we run on cdash are passing, but of course, we don't expect BFB with a newer compiler.
We don't see any major performance differences yet. 
Tried basic ne30 cases as well as a basic ne256 Cess-style case.
For the NESAP benchmark (ne1024 cess-style case) -- it actually ran quite a bit faster, but looking at timing data, the compute intensive timers were a bit slower (1-2% ?), where the IO timers were faster -- suspect that is from another PR.

I'm also updating nvidia compiler version, but we already know this currently has trouble building.
In debugging that issue, I found I needed to update the compiler, so hopefully this is a step in the direction of getting nvidia working again with SCREAM.

Note we are not updating the versions on pm-cpu in this PR.

[NBFB] only pm-gpu/muller-gpu

### DIFF
--- a/cime_config/machines/config_machines.xml
+++ b/cime_config/machines/config_machines.xml
@@ -678,6 +678,7 @@
         <command name="unload">nvidia</command>
         <command name="unload">aocc</command>
         <command name="unload">cudatoolkit</command>
+        <command name="unload">cray-libsci</command>
         <command name="unload">climate-utils</command>
         <command name="unload">matlab</command>
         <command name="unload">craype-accel-nvidia80</command>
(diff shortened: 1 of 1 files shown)


ndkeen:
Right -- thanks Naser. It was actually Naser who figured out how to download/install our own cudatoolkit version, make a module, and we verified it works with SCREAM. It's still not clear if NERSC will be removing cuda11 in the near future (of course, eventually they will). These newer versions have been the machine default for some time now, so we should update asap.

ndkeen:
Yes, I should also update muller-gpu at the same time to keep them consistent.  But as no tests run there, it's not necessary in this PR. If it doesn't slow us down, I can update muller-gpu here.

PR #2894, opened 2024-07-09
MAM4xx Dry Deposition process interface for EAMxx

overfelt:
This PR connects MAM4xx’s dry deposition codes to EAMxx using an interface. Since the cmake option, `SCREAM_ENABLE_MAM` is always `ON`, this code is always compiled with the EAMxx codebase. Dry deposition is added as an EAMxx process with the ability to turn it on/off using the label `mam4_drydep` in the ` namelist_scream.xml` file.  The atmchange command to invoke this process in a CIME simulation is:
 
`./atmchange physics::atm_procs_list="mac_aero_mic,rrtmgp,mam4_drydep"`

To set up a test with the above configuration, use the following commands:
```
cd cime/scripts
./create_test SMS_D_P32x1_Ln5.ne4pg2_oQU480.F2010-SCREAMv1-MPASSI.<machine>_<compiler>.scream-mam4xx-drydep -t <test id> -p <project id>
```
Where,
`<machine>` is the machine name
`<compiler>` is the compiler to use on the machine
`<test id >` is a unique string test identifier
`<project id>` is the allocation project to charge

`scream-mam4xx-drydep` is the test modifier that adds `mam4_drydep` to the `atm_procs_list` process list and increases the number of tracers to 41.
 
The output from this process includes:
1.  Updated interstitial aerosols
2. Updated cloud-borne aerosols
3. Deposition flux of interstitial aerosols
4. Deposition flux of cloud-borne aerosols

 
### Testing
 We added 1 single process, 2 multi process, and 1 CIME low-resolution test in EAMxx. These are all smoke tests. MAM4xx has its validation testing suites where we validate all MAM4xx codes. Manual end-to-end validation was done for the entire dry deposition interface and the results were exactly the same as EAMv2 grid cells up to single precision. We don’t expect an exact match till double precision as the computing environments are different for both runs.
 
### Figures from model analysis
image
image

### Standalone Test Timings (Compy):
(Coming soon)

### Important notes:
1. Fractional land use is fixed. It involves a file reading, we will add that as a separate feature.


components/eamxx/cime_config/testdefs/testmods_dirs/scream/mam4xx/drydep/shell_commands
```diff
@@ -0,0 +1,13 @@
+
+#Default scream has 10 tracers, MAM4xx adds another 31 making a total of 41 tracer
+#Set total number of tracers to 41. We are using append here as last entry wins while parsing xml options
+./xmlchange --append SCREAM_CMAKE_OPTIONS="SCREAM_NUM_TRACERS 41"
```
AaronDonahue, line 4: I left a similar comment on @singhbalwinder 's ACI PR:
Here we assume a default of 10 tracers in EAMxx, but as the model evolves we may expect some configurations that could have >10.  @singhbalwinder did you put something in your PR that queried the number of tracers before adding 31?  Maybe there is no easy solution other than what you have done here by including a comment.

AaronDonahue: Actually, this is for a test, so maybe this is sufficient since we can assume that the test is setup to use 10 tracers for SCREAM.

AaronDonahue (commented):
Looks good to me.  I will let Luca take a look specifically at the Kokkos loops, etc to make sure I didn't miss anything.

components/eamxx/src/physics/mam/eamxx_mam_dry_deposition_process_interface.cpp
```diff
@@ -0,0 +1,430 @@
...
+  const FieldLayout scalar2d{{COL}, {ncol_}};
+
+  // Layout for 3D (2d horiz X 1d vertical) variable defined at mid-level and
+  // interfaces
+  const FieldLayout scalar3d_mid{{COL, LEV}, {ncol_, nlev_}};
+  const FieldLayout scalar3d_int{{COL, ILEV}, {ncol_, nlev_ + 1}};
+
+  // layout for 2D (ncol, pcnst)
+  constexpr int pcnst = mam4::aero_model::pcnst;
+  const FieldLayout scalar2d_pcnct =
```
bartgol, line 53: Since you use the grid to build layouts (and that's the best way, imho), you could have used it also to build the layouts above.

components/eamxx/src/physics/mam/eamxx_mam_dry_deposition_process_interface.cpp
```diff
@@ -0,0 +1,430 @@
...
+
+  // Layout for 4D (2d horiz X 1d vertical x number of modes) variables
+  // at mid points
+  auto make_layout = [](const std::vector<int> &extents,
+                        const std::vector<std::string> &names) {
+    std::vector<FieldTag> tags(extents.size(), CMP);
+    return FieldLayout(tags, extents, names);
+  };
+  const int num_aero_modes = mam_coupling::num_aero_modes();
+  FieldLayout scalar4d_mid =
```
bartgol, line 64: This seems like a "3d vector" layout. You can just build it using the grid interfaces, like you did for pcnst above
```c++
auto scalar4d_mid = grid->get_3d_vector_layout(true,num_aero_modes,"num_modes");
```

components/eamxx/src/physics/mam/eamxx_mam_dry_deposition_process_interface.cpp
bartgol: Also, the name is "scalar2d", but you call `get_2d_vector_layout`. Perhaps the name should be changed, to be less confusing?

components/eamxx/src/physics/mam/mam_coupling.hpp
```diff
@@ -592,21 +592,26 @@ void compute_vertical_layer_heights(const Team& team,
...
   const auto qv = ekat::subview(dry_atm.qv, column_index);
   const auto p_mid = ekat::subview(dry_atm.p_mid, column_index);
   const auto T_mid = ekat::subview(dry_atm.T_mid, column_index);
   const auto pseudo_density = ekat::subview(dry_atm.p_del, column_index);
+  
   // NOTE: we are using dry qv. Does calculate_dz require dry or wet?
   PF::calculate_dz(team, pseudo_density, p_mid, T_mid, qv, // inputs
             dz);//output
   team.team_barrier();
+  EKAT_KERNEL_ASSERT_MSG(dry_atm.z_surf == 0, "dry_atm.z_surf must be zero");
```
bartgol, line 609: I don't know anything about mam, or what z_surf represents, but...does this mean you're assuming no topography?

bartgol (commented):
I have a few comments, mostly on readability.

components/eamxx/src/physics/mam/mam_coupling.hpp
singhbalwinder: I mimicked this from EAM but @kaizhangpnl might know more about it. [Here ](https://github.com/eagles-project/scream/blob/3005a4b8ef245c06f0a769a87c3c85bf0c29e519/components/eam/src/physics/cam/geopotential.F90#L83)is the EAM code from where I assumed that it should be zero

bartgol: It depends where it comes from. Grepping in the mam folder, it seems it is always set to zero (and never changed):
```bash
$ git grep z_surf
eamxx_mam_aci_process_interface.cpp:  dry_atm_.z_surf = 0;
eamxx_mam_dry_deposition_process_interface.cpp:  dry_atm_.z_surf    = 0.0;  // FIXME: for now
eamxx_mam_microphysics_process_interface.cpp:  dry_atm_.z_surf = 0.0; // FIXME: for now
eamxx_mam_optics_process_interface.cpp:  dry_atm_.z_surf    = 0.0;
mam_coupling.hpp:  Real          z_surf;    // height of bottom of atmosphere [m]
mam_coupling.hpp:  EKAT_KERNEL_ASSERT_MSG(dry_atm.z_surf == 0, "dry_atm.z_surf must be zero");
mam_coupling.hpp:  PF::calculate_z_int(team, mam4::nlev, dz, dry_atm.z_surf, //inputs
```
But the comment "FIXME: for now" makes me think this is just temporary.

bartgol (approved):

components/eamxx/src/physics/mam/mam_coupling.hpp
singhbalwinder: I should remove the `FIXME` comments for `z_surf`. I just grepped the EAMxx code and it is set to zero at all places there as well (e.g. SHOC, TMS, surface coupling, etc.). So it should be zero everywhere.

bartgol:
@singhbalwinder while github claims the PR is not mergeable, I don't see any conflict listed. It may be related to submodules. In these cases, the easiest thing is to simply rebase and force-push.

components/eamxx/src/physics/mam/mam_coupling.hpp
kaizhangpnl: state%zi in EAM is the geopotential height (at the layer interface) above the surface (not above the sea level). At the bottom of the model, it is zero by definition. On the other hand, I am not sure how z_surf is used exactly in MAMxx (just to calculate dz?).

bartgol: You should not need z_surf to compute dz though...

kaizhangpnl: By searching "z_surf" in the repo, it seems that it has different definitions in different parameterizations: 

https://github.com/search?q=repo%3AE3SM-Project%2Fscream%20z_surf&type=code 

Maybe it (how it is defined and used) needs to be described more clearly in the code.

singhbalwinder: In MAM4xx codes, we use `z_surf` to compute `z_iface` (or `z_int` or state%zi in EAM or z interface values). We followed EAMxx to compute `z_iface` using the `PF::calculate_z_int` function. In [EAM](https://github.com/eagles-project/scream/blob/3005a4b8ef245c06f0a769a87c3c85bf0c29e519/components/eam/src/physics/cam/geopotential.F90#L83), we also set it to zero at the surface.

bartgol:
Looks like the only tests that fail are due to baselines cmp. I'm assuming all tests are new, correct?

singhbalwinder:
The only unexpected fail is:
`mam4_wetscav_standalone_baseline_cmp`

The baselines were not generated for this test earlier. I will ask for a retest and expect the test above to `pass`.

The following tests are all new tests and are expected to fail due to missing baselines:
```
DIFF SMS_D_Ln5.ne4pg2_oQU480.F2010-SCREAMv1-MPASSI.mappy_gnu.scream-mam4xx-drydep (phase BASELINE)
mam4_drydep_standalone_baseline_cmp
homme_shoc_cld_mam_aci_p3_mam_optics_rrtmgp_mam_drydep_baseline_cmp
homme_shoc_cld_spa_p3_rrtmgp_mam4_wetscav_baseline_cmp
```

singhbalwinder:
All the failing tests in the recent re-run of the tests are expected fails (missing baselines).

kaizhangpnl:
Below is the checklist required by the PR review guidance: https://acme-climate.atlassian.net/wiki/spaces/EAMXX/pages/4128768032/Short+description+of+EAGLES+PR+review+process 

> Does the PR description accurately and adequately summarize the content of the PR?

Yes. 

> Are the expected benefits of the new feature adequately demonstrated by the supplied documentation?

Yes, the PR summary does the job. 

> Does the feature add new prognostic variables?  If yes, have they been sufficiently documented and compared with benchmarks?

The new feature (when turned on) will 

1. read gas and aerosol tracers from the new initial condition file; 
2. transport the these tracers (homme & shoc); 
3. make dry removal calculation 

The aerosol and gas tracers are "prognostic" since their values are affected by transport. When this feature (drydep) is turned on, the aerosol tracers (both interstitial and cloud-borne) are affected by dry removal. 

Due to the idealized configuration with this feature, there is no benchmark to compare with. 

> Does the feature add new couplings?  If yes, has the behavior of the variables mediating the coupling been documented, and compared with any static fields used previously, and with other benchmarks if appropriate?

No new coupling (that affects meteorological fields) is added, except that aerosol concentrations will be changed by drydep and this will affect "aci" calculation in the next model time step. 

> Are there any notable impacts on other aspects of climate?

The "drydep" feature alone only changes aerosol concentrations and does not directly affect meteorological fields. Also, the current tests are short idealized simulations and the impact on climate is not evaluated. 

> Are any additional diagnostics or evaluations needed before deciding on feature acceptance? 

Due to the idealized configuration of the current model feature, we can only evaluate the feature impact qualitatively. It would be more meaningful to perform the aerosol mass and number budget analysis when we can run a longer simulation with more realistic aerosol fields in MAMxx/EAMxx. 

> Additional evaluation of the new feature

The simulated aerosol burden changes and dry deposition fluxes are evaluated (here we only consider dry removal of aerosols). The simulated BC aerosol burden decreases by 21% [(figure)](https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.kzhang/TMP/drydep/figure_mam_drydep_surf_BURDENBC_daily_avg.pdf) and dust burden by 43% in 5 days [figure](https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.kzhang/TMP/drydep/figure_mam_drydep_surf_BURDENDST_daily_avg.pdf). The higher dry removal rate for dust is reasonable, since the dry deposition flux also includes sedimentation that is a major sink for larger dust particles. The dry removal rate for sea salt seems to be underestimated (burden decreases by 18% in 5 days [(figure)](https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.kzhang/TMP/drydep/figure_mam_drydep_surf_BURDENSS_daily_avg.pdf). 

Additional diagnostics can be found here: https://web.lcrc.anl.gov/public/e3sm/diagnostic_output/ac.kzhang/TMP/drydep/ 

> Additional evaluation of expected impacts on other parts of the model (due to direct or indirect couplings)

N/A. 

> Additional notes

Available diagnostics output fields: 

```
      - deposition_flux_of_cloud_borne_aerosols
      - deposition_flux_of_interstitial_aerosols
```

   The dry deposition flux includes both turbulent dry deposition at the surface and sedimentation flux. In the future, it would 
   be nice to diagnose the two (turbulent dry deposition and sedimentation) separately.

bartgol (dismissed):

components/eamxx/src/physics/mam/eamxx_mam_dry_deposition_functions.hpp
```diff
@@ -0,0 +1,205 @@
...
+        Kokkos::deep_copy(
+            dry_aero.cld_aero_mmr[m][a],
+            ekat::subview(qqcw, mam4::ConvProc::lmassptrcw_amode(a, m)));
+      }
+    }
+  }
+}  // Update cloud borne aerosols ends
+
+// FIXME: remove the following function after implementing file read for landuse
+void populated_fraction_landuse(MAMDryDep::view_2d flu, const int ncol) {
```
kaizhangpnl, line 185: What is the relationship between this and the following (mam4xx/drydep.hpp) code? 

https://github.com/eagles-project/mam4xx/blob/2eee1988b1e6488c904e0a28564bdcadfa190d34/src/mam4xx/drydep.hpp#L32C10-L32C26

kaizhangpnl (approved):
Other than the overall comment, I only have two other comments/questions that could be easily addressed, so I'm going to approve this PR.

components/eamxx/src/mct_coupling/scream_cpl_indices.F90
```diff
@@ -106,7 +109,10 @@ subroutine scream_set_cpl_indices (x2a, a2x)
     import_cpl_indices(13) = mct_avect_indexra(x2a,'Faxx_sen')
     import_cpl_indices(14) = mct_avect_indexra(x2a,'Faxx_evap')
     import_cpl_indices(15) = mct_avect_indexra(x2a,'Sf_ofrac')
-    import_cpl_indices(16) = mct_avect_indexra(x2a,'Sf_lfrac') 
+    import_cpl_indices(16) = mct_avect_indexra(x2a,'Sf_lfrac')
+    import_cpl_indices(17) = mct_avect_indexra(x2a,'Sf_ifrac') 
+    import_cpl_indices(19) = mct_avect_indexra(x2a,'Sl_fv')
```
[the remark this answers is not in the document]
singhbalwinder: Thanks for catching that. I will fix this.

components/eamxx/src/physics/mam/eamxx_mam_dry_deposition_functions.hpp
singhbalwinder: Variable `fraction_landuse[n_land_type]` at https://github.com/eagles-project/mam4xx/blob/2eee1988b1e6488c904e0a28564bdcadfa190d34/src/mam4xx/drydep.hpp#L32C10-L32C26  is only used if we invoke ` DryDeposition::compute_tendencies` (see [here](https://github.com/eagles-project/mam4xx/blob/2eee1988b1e6488c904e0a28564bdcadfa190d34/src/mam4xx/drydep.hpp#L1129C5-L1129C39)) function. This function is only used in MAM4xx validation tests. EAMxx never invokes this function.

The `populated_fraction_landuse` in the EAMxx dry deposition interface is a temporary function that specifies fractional land use values for now. We will replace it with a reader that will read the values of fractional land use later.

singhbalwinder:
We have the expected fails. @kaizhangpnl, if everything looks good to you,  should we go ahead and merge it?

kaizhangpnl:
> We have the expected fails. @kaizhangpnl, if everything looks good to you, should we go ahead and merge it?

Thanks for addressing my comments. Yes, it looks good to me.

singhbalwinder:
Updated the PR description to add timing information for all the unit tests (including this process).

PR #2899, opened 2024-07-11
Fix ThreadVectorRange -> TeamVectorRange in DPxx

tcclevenger:
One usage of `ThreadVectorRange` leaked through. This should be BFB (I don't think the CIME case uses these nudging variables).

Also, small optimization: only launch kernel if one of `iop_nudge_tq` or `iop_nudge_uv` is true.


tcclevenger:
Hide whitespace option shows the relative changes more clearly.

bogensch (approved):

Issue #2898, opened 2024-07-11
IO changes dtypes between rhist and h, leading to fails


mahf708:
This issue was encountered by @brhillman. Reporting the details here for better provenance.

```
  699: 
  699:  FAIL:
  699: var->dtype==refine_dtype(dtype)
  699: /autofs/nccs-svm1_home1/brhillman/codes/scream/branches/decadal-production-run6/components/eamxx/src/share/io/scream_scorpio_interface.cpp:997
  699: Error! Attempt to redefine variable with different data type.
  699:  - filename : output.scream.decadal.monthlyAVG_ne30pg2.AVERAGE.nmonths_x1.1994-10-01-00000.nc
  699:  - varname  : avg_count_ncol
  699:  - old dtype: float
  699:  - new dtype: double
  699: 
```

Some notes from internal messaging:

- The problem stems from the fact that the attr in question was written as float in the ACTUAL output, but was written as DOUBLE in the rhist files

- In all rhist files (there are 4 of them) it is always double, I have no clue how it was written as FLOAT in the output

- This commit is v sus: https://github.com/E3SM-Project/scream/commit/29dd63b8e0eb13e7153f5f5805c0ae3ff7661570

- Setting `force_new_file: true` (default false) seems to overcome this issue

highly likely xref #2890 #2892

PR #2902, opened 2024-07-11
Fix bug in io with registering avg_cnt vars in Append mode.

AaronDonahue:
This commit addresses a bug where average count tracking variables may be registered again when a file is in Append mode.  There is logic in the registration of typical output variables that was missing for average count variables.

Addresses #2898

### DIFF
--- a/components/eamxx/src/share/io/scorpio_output.cpp
+++ b/components/eamxx/src/share/io/scorpio_output.cpp
@@ -1038,11 +1038,42 @@ register_variables(const std::string& filename,
   }
   // Now register the average count variables
   if (m_track_avg_cnt) {
+    std::string unitless = "unitless";
     for (const auto& name : m_avg_cnt_names) {
       const auto layout = m_layouts.at(name);
       auto vec_of_dims   = set_vec_of_dims(layout);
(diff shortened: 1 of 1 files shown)


AaronDonahue:
@ndkeen , this fix appeared to address the issue of the double vs. float error in my local recreator.  If you have a chance could you verify independently that it works for your tests?

components/eamxx/src/share/io/scorpio_output.cpp
```diff
@@ -1038,11 +1038,42 @@ register_variables(const std::string& filename,
   }
   // Now register the average count variables
   if (m_track_avg_cnt) {
+    std::string unitless = "unitless";
     for (const auto& name : m_avg_cnt_names) {
       const auto layout = m_layouts.at(name);
       auto vec_of_dims   = set_vec_of_dims(layout);
-      scorpio::define_var(filename, name, "unitless", vec_of_dims,
-                          "real",fp_precision, m_add_time_dim);
+      if (mode==scorpio::FileMode::Append) {
```
AaronDonahue, line 1045: This is essentially a copy/paste of the same check used above for `m_field_names`

PR #2903, opened 2024-07-12
DPxx: IOP pressure level fix and fix to lat/lon relative error checking

bogensch:
Fix to read in IOP pressure level data each time the IOP file is read.  This prevents the persistence of IOP pressure level data, which may be adjusted due to the surface pressure and occasionally may result in pressure increasing with height and hence crashes.

In addition, fix an issue where the IOP lat and lon is checked to be consistent with that provided by case specifications.  The relative error checker divides by the lat and lon, which creates NaNs when either of these is 0.  Thus introduce a minimum threshold (0.1 degree).

This PR will not be b4b for DPxx, answer changes are expected.

### DIFF
--- a/components/eamxx/src/share/iop/intensive_observation_period.cpp
+++ b/components/eamxx/src/share/iop/intensive_observation_period.cpp
@@ -541,6 +541,15 @@ read_iop_file_data (const util::TimeStamp& current_ts)
     scorpio::read_var(iop_file,"Ps",ps_data,iop_file_time_idx);
     surface_pressure.sync_to_dev();
 
+    // Read in IOP lev data
+    auto data = iop_file_pressure.get_view<Real*, Host>().data();
+    scorpio::read_var(iop_file,"lev",data);
+
+    // Convert to pressure to millibar (file gives pressure in Pa)
+    for (int ilev=0; ilev<file_levs; ++ilev) data[ilev] /= 100;
+    iop_file_pressure.sync_to_dev();
+    m_helper_fields.insert({"iop_file_pressure", iop_file_pressure});
+
     // Pre-process file pressures, store number of file levels
     // where the last level is the first level equal to surface pressure.
     const auto iop_file_pres_v = iop_file_pressure.get_view<Real*>();
--- a/components/eamxx/src/share/iop/intensive_observation_period.cpp
+++ b/components/eamxx/src/share/iop/intensive_observation_period.cpp
@@ -268,9 +268,9 @@ initialize_iop_file(const util::TimeStamp& run_t0,
   scorpio::read_var(iop_file,"lon",&iop_file_lon);
 
   const Real rel_lat_err = std::fabs(iop_file_lat - m_params.get<Real>("target_latitude"))/
-                             m_params.get<Real>("target_latitude");
+                             ekat::impl::max(m_params.get<Real>("target_latitude"),0.1);
   const Real rel_lon_err = std::fabs(std::fmod(iop_file_lon + 360.0, 360.0)-m_params.get<Real>("target_longitude"))/
-                             m_params.get<Real>("target_longitude");
+                             ekat::impl::max(m_params.get<Real>("target_longitude"),0.1);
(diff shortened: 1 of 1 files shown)


oksanaguba:
@bogensch is it correct that for the IOP forcing in DPxx there is no pressure adjustment for active Q forcing?

components/eamxx/src/share/iop/intensive_observation_period.cpp
```diff
@@ -268,9 +268,9 @@ initialize_iop_file(const util::TimeStamp& run_t0,
   scorpio::read_var(iop_file,"lon",&iop_file_lon);
 
   const Real rel_lat_err = std::fabs(iop_file_lat - m_params.get<Real>("target_latitude"))/
-                             m_params.get<Real>("target_latitude");
+                             ekat::impl::max(m_params.get<Real>("target_latitude"),0.1);
   const Real rel_lon_err = std::fabs(std::fmod(iop_file_lon + 360.0, 360.0)-m_params.get<Real>("target_longitude"))/
-                             m_params.get<Real>("target_longitude");
+                             ekat::impl::max(m_params.get<Real>("target_longitude"),0.1);
```
tcclevenger, line 273: Same here.

components/eamxx/src/share/iop/intensive_observation_period.cpp
```diff
@@ -541,6 +541,15 @@ read_iop_file_data (const util::TimeStamp& current_ts)
...
     surface_pressure.sync_to_dev();
 
+    // Read in IOP lev data
+    auto data = iop_file_pressure.get_view<Real*, Host>().data();
+    scorpio::read_var(iop_file,"lev",data);
+
+    // Convert to pressure to millibar (file gives pressure in Pa)
+    for (int ilev=0; ilev<file_levs; ++ilev) data[ilev] /= 100;
+    iop_file_pressure.sync_to_dev();
+    m_helper_fields.insert({"iop_file_pressure", iop_file_pressure});
```
tcclevenger, line 551: Remove this `insert()` call. It will do nothing since the key "iop_file_pressure" already exists in the `m_helper_fields` map object. And we are updating the data of the field stored in the map, so all the info is automatically in the map object when we update the data.

tcclevenger (changes requested):
Some minor changes.

tcclevenger (dismissed):

components/eamxx/src/share/iop/intensive_observation_period.cpp
```diff
@@ -268,9 +268,9 @@ initialize_iop_file(const util::TimeStamp& run_t0,
   scorpio::read_var(iop_file,"lon",&iop_file_lon);
 
   const Real rel_lat_err = std::fabs(iop_file_lat - m_params.get<Real>("target_latitude"))/
-                             m_params.get<Real>("target_latitude");
+                             std::max(m_params.get<Real>("target_latitude"),0.1);
```
tcclevenger, line 271: We need to cast to `Real` because `std::max` requires both arguments to be the same type.
```suggestion
                             std::max(m_params.get<Real>("target_latitude"),(Real)0.1);
```

components/eamxx/src/share/iop/intensive_observation_period.cpp
```diff
@@ -268,9 +268,9 @@ initialize_iop_file(const util::TimeStamp& run_t0,
...
   const Real rel_lon_err = std::fabs(std::fmod(iop_file_lon + 360.0, 360.0)-m_params.get<Real>("target_longitude"))/
-                             m_params.get<Real>("target_longitude");
+                             std::max(m_params.get<Real>("target_longitude"),0.1);
```
tcclevenger, line 273: Same here.
```suggestion
                             std::max(m_params.get<Real>("target_longitude"),(Real)0.1);
```

tcclevenger (approved):

tcclevenger:
A comment since this merged B4B: The CIME case used most likely didn't trigger the DIFF since only one time slice is loaded.

PR #2909, opened 2024-07-18
ML correction on CPU and GPU

elynnwu:
In this PR, we add the capability of running ML correction on GPU. The call to the python code differs due to how pybind11 handles the exchange. On cpu, we continue to rely on pybind11 to pass the array pointer using its integration with numpy. On gpu, we pass the pointer then rebuild it as a cupy array on the python side (gpu allows unmanaged memory access). By doing so, the rest of the code is unchanged since xarray can work with both numpy and cupy arrays. As a result, the actual calls to do ML correction is identical between cpu and gpu.

We also introduce a few features in this PR:
 - new option for only correcting for temperature and not specific humidity is added.  
 - surface radiative flux correction now overwrites all direct and diffused component of `sfc_flux` as well as `sfc_flux_sw_net` and `sfc_flux_sw_dn`

We have also started focusing on using perlmutter cpu and gpu as our main machine for ML corrective work. A shared python env is now maintained at: `/global/common/software/m4492/fv3net-shared-py39`


ndkeen:
I don't think we want to add modules for pm-gpu builds?
```
<command name="load">cudnn/8.9.3_cuda12</command>
```
Would be better to find a conditional way to do these sorts of things for special cases.

I'm not seeing a great example right now. We could come up with a new CIME variable similar
to `BUILD_THREADED` to be `BUILD_ML` or something. 

Alternatives:
Make an additional machine configuration, which I would recommend against as it's awkward and overkill.
You can also load these modules and set these env variables yourself in the launch script. They should get picked
up by the job.

elynnwu:
I opted to remove ML specific settings from `config_machines`, we will include them in our launch script as Noel suggested. A short doc is also added.

components/eamxx/docs/user/ml_correction.md
```diff
@@ -0,0 +1,35 @@
...
+
+```shell
+./atmchange physics::atm_procs_list="mac_aero_mic,rrtmgp,mlcorrection"
+```
+
+The following options can be specified:
+
+```shell
+./atmchange ML_model_path_tq=/path/to/pretrained/temperature_and_specific_humidity_model
+./atmchange ML_model_path_uv=/path/to/pretrained/u_and_v_model
```
frodre, line 17: Do we still have the Tuv only option, or did we drop that since it wasn't showing positive results?

components/eamxx/src/physics/ml_correction/CMakeLists.txt
```diff
@@ -15,6 +15,10 @@ include(ScreamUtils)
 else()
     message(FATAL_ERROR "pybind11 is missing. Use CMake >= 3.11 or download it")
 endif()
+
+message(STATUS "Python_EXECUTABLE: ${Python_EXECUTABLE}")
```
frodre, line 19: Note: I was originally using this for debugging when there was a conflict between the discovered CMAKE Python and the Pybind python.  Maybe no longer necessary?

components/eamxx/tests/single-process/ml_correction/CMakeLists.txt
```diff
@@ -12,7 +12,13 @@ else()
     message(FATAL_ERROR "pybind11 is missing. Use CMake >= 3.11 or download it")
 endif()
 
+message(STATUS "Python_EXECUTABLE: ${Python_EXECUTABLE}")
```
frodre, line 15: Same as before, these were just for debugging CMAKE vs Pybind found python libraries and could be omitted.

frodre (commented):
Thanks for the updated PR for this.  It's too bad we couldn't have a unified interface, but this is the best we can do given the setup!  Just a few minor comments about some leftovers from my debugging days.

components/eamxx/docs/user/ml_correction.md
elynnwu: We do, including in the doc now.

components/eamxx/cmake/machine-files/pm-cpu.cmake
```diff
@@ -5,6 +5,9 @@ include (${EKAT_MACH_FILES_PATH}/kokkos/amd-zen3.cmake)
 include (${EKAT_MACH_FILES_PATH}/kokkos/openmp.cmake)
 
 set(CMAKE_CXX_FLAGS "-DTHRUST_IGNORE_CUB_VERSION_CHECK" CACHE STRING "" FORCE)
+set(PYBIND11_PYTHON_VERSION 3.9 CACHE STRING "")
```
AaronDonahue, line 8: @elynnwu or @frodre , I think there is consensus that we can merge this in as long as we put guard rails on the environment call.  Can you add a compiler flag to turn these options "OFF" by default and only "ON" when the user specifies they want ML?  I would be in favor of making the flag explicit in name, like `FV3NET` or `CorrectiveML`.  You would want to add the flags to both gpu and cpu config files.

@mahf708 have I characterized what needs to be done correctly?

AaronDonahue:
Closing this until we finalize the ML approach

PR #2912, opened 2024-07-19
fix capture this warning in mam micro

mahf708:
I am not entirely sure if this step_ gets ever updated; it doesn't look like it. Anyway, we cannot use members deep inside kernels automatically (it is only a warning, but sometimes it becomes an error). Fixing here to avoid the warning.


jeff-cohere (approved):
I'm also not sure we need this, but this fix makes sense. Thanks @mahf708

singhbalwinder (approved):
Thanks for taking care of it, Naser!

PR #2913, opened 2024-07-19
add output variable for rain effective diameter

hassanbeydoun:
Rain effective diameter is needed to run the offline radar simulator (https://github.com/columncolab/EMC2). This PR adds rain effective diameter as an optional output variable from P3 for v0 only. A future PR will be needed for v1.


AaronDonahue (approved):

AaronDonahue:
@hassanbeydoun can you add a description to the PR?  And then you can add labels for it, including the Automerge label

PeterCaldwell:
@hassanbeydoun and others - I think it is very important that we flag variables we're adding to the FM solely for output so when we have a better approach to writing these variables out, we can easily see which variables we should remove from FM. Can you add a comment after the 3 lines of code you add saying this variable is for output only?

Or- question for others - is identifying these variables from the DAG so easy that we don't need inline flagging?

PR #2916, opened 2024-07-19
filter aodvis by day indices

### LINKED ISSUE #2910: AerosolOpticalDepth550nm should be day-filtered
AerosolOpticalDepth550nm must be day-filtered because aero_tau_sw is not.

mahf708:
fix #2910

### DIFF
--- a/components/eamxx/src/diagnostics/tests/aodvis_test.cpp
+++ b/components/eamxx/src/diagnostics/tests/aodvis_test.cpp
@@ -1,10 +1,11 @@
+#include <share/field/field_utils.hpp>
+
 #include "catch2/catch.hpp"
 #include "diagnostics/register_diagnostics.hpp"
 #include "share/field/field_utils.hpp"
 #include "share/grid/mesh_free_grids_manager.hpp"
 #include "share/util/scream_setup_random_test.hpp"
-#include "share/util/scream_utils.hpp"
-
+#include "share/util/scream_universal_constants.hpp"
 namespace scream {
 
 std::shared_ptr<GridsManager> create_gm(const ekat::Comm &comm, const int ncols,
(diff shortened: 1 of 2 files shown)


components/eamxx/src/diagnostics/tests/aodvis_test.cpp
```diff
@@ -71,18 +83,33 @@ TEST_CASE("aodvis") {
...
+    }
+    // sync to device
+    sunlit.sync_to_dev();
+
     // Run diag
     diag->compute_diagnostic();
 
     // Check result
     tau.sync_to_host();
+    sunlit.sync_to_host();
```
bartgol, line 112: Harmless, but also pointless, since sunlit is up to date on host.

bartgol (dismissed):

bartgol:
> @AaronDonahue @bartgol could you confirm the the IO will take care of correctly averaging the output here, given that ~half of its values are going to be fillval? I was hoping of doing a small test but I never got around to it, so a review before a test would be good, thanks!

You need to enable the tracking of avg count in the yaml file, but if you do, then yes, it will only average non-filled values. Of course, this means that stuff like hourly output may make no sense, but I'm quite confident you will only need daily output (or less), which means every column will get 1+ valid values to average.

To enable the avg count tracking, you need `track_avg_cnt: true` in the yaml file.

@AaronDonahue should we maybe turn on avg count tracking automatically for this diag? Or, more in general, for any diag that can have a certain extra data set in the header (something like "can_have_invalid_data" or something)?

AaronDonahue:
>@AaronDonahue should we maybe turn on avg count tracking automatically for this diag? Or, more in general, for any field that can have a certain extra data set in the header (something like "can_have_invalid_data" or something)?
Yes, I think we would want to do that rather than have users try to figure out why some fields are nan, or otherwise bad values.

bartgol (approved):

PR #2917, opened 2024-07-22
Fix how IOP reads time

tcclevenger:
Basically, if time is unlimited, you cannot read the time array all at once using `read_var()`. Fix: treat time as unlimited and then read each time index in sequence using `time_idx` input for `read_vars()`.

### DIFF
--- a/components/eamxx/src/share/iop/intensive_observation_period.cpp
+++ b/components/eamxx/src/share/iop/intensive_observation_period.cpp
@@ -248,13 +248,15 @@ initialize_iop_file(const util::TimeStamp& run_t0,
   else if (scorpio::has_dim(iop_file, "tsec")) time_dimname = "tsec";
   else EKAT_ERROR_MSG("Error! No valid dimension for tsec in "+iop_file+".\n");
 
-  const auto ntimes = scorpio::get_dimlen(iop_file, time_dimname);
-  m_time_info.iop_file_times_in_sec = view_1d_host<int>("iop_file_times", ntimes);
-  scorpio::read_var(iop_file,"tsec",m_time_info.iop_file_times_in_sec.data());
-
-  // From now on, when we read vars, "time" must be treated as unlimited, to avoid issues
+  // When we read vars, "time" must be treated as unlimited, to avoid issues
   if (not scorpio::is_dim_unlimited(iop_file,time_dimname)) {
     scorpio::pretend_dim_is_unlimited(iop_file,time_dimname);
+  }
+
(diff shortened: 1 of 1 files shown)


bartgol:
We also have the routine [get_all_times](https://github.com/E3SM-Project/scream/blob/a1b89eb8e0db61100e03c9770a18e11ed307c8d4/components/eamxx/src/share/io/scream_scorpio_interface.hpp#L189), if that helps...

tcclevenger:
> We also have the routine [get_all_times](https://github.com/E3SM-Project/scream/blob/a1b89eb8e0db61100e03c9770a18e11ed307c8d4/components/eamxx/src/share/io/scream_scorpio_interface.hpp#L189), if that helps...
> 
> Edit: although, this only work if the time dimension is unlimited. But you can first check that with `is_dim_unlimited`, or maybe call `pretend_dim_is_unlimited` before calling `get_all_times`...

Yeah, I thought about `get_all_times` but that only works if the time dimension is the same name as the time variable. In IOP files the time variable is named `tsec` and the time dimension could be `tsec` or `time`.

We could have a `change_var_name` function, but that is probably overkill for just this situation.

bartgol:
> > We also have the routine [get_all_times](https://github.com/E3SM-Project/scream/blob/a1b89eb8e0db61100e03c9770a18e11ed307c8d4/components/eamxx/src/share/io/scream_scorpio_interface.hpp#L189), if that helps...
> > Edit: although, this only work if the time dimension is unlimited. But you can first check that with `is_dim_unlimited`, or maybe call `pretend_dim_is_unlimited` before calling `get_all_times`...
> 
> Yeah, I thought about `get_all_times` but that only works if the time dimension is the same name as the time variable. In IOP files the time variable is named `tsec` and the time dimension could be `tsec` or `time`.
> 
> We could have a `change_var_name` function, but that is probably overkill for just this situation.

Ah, now I see the issue. Thanks!

bartgol (approved):

bogensch:
thanks @tcclevenger; I verified that this PR solves the issues I was seeing in many cases.

bogensch (approved):

AaronDonahue:
this is failing on weaver for the aci standalone baseline test.  I don't think anything about this PR would make us think it impacted ACI.  Should we rebase the branch and see if the fail goes away?

tcclevenger:
> this is failing on weaver for the aci standalone baseline test. I don't think anything about this PR would make us think it impacted ACI. Should we rebase the branch and see if the fail goes away?

I can.

PR #2919, opened 2024-07-25
Add runtime and buildnml-time checks on output frequency

### LINKED ISSUE #2891: IO doesn't O if freq < timestep
To repro, take ne4pg2_ne4pg2 res whose timestep is 1 hour, set an instant output stream with frequency of 30 mins. 

The result: the very first step is written, but nothing afterwards. 

This seems kind of weird because obviously every step should in theory be an output step. In the code, I thought the output manager checks if the step is an output step by doing that logic, irrespective of freq < timestep or freq > timestep (but I could have misunderstood).

If the code is indeed working like intended, then we can document the behavior and close this issue. 

See internal outputs https://acme-climate.atlassian.net/wiki/spaces/EAMXX/pages/4334223933/EAMxx+ERFaer+production from a recent run using commit https://github.com/E3SM-Project/scream/commit/29bdb81 on branch https://github.com/E3SM-Project/scream/tree/mahf708-ff-a73d48a

bartgol:
Ensures dt_output is not smaller than dt_atm, which would cause empty output files. I ran case.setup with an output file with frequency=1800s, for a case with dt_atm=3600s, and got the error
```
ERROR: Cannot have output frequency faster than atm timestep.
  yaml file: /home/lbertag/acme/scratch/SMS.ne4pg2_ne4pg2.F2010-SCREAMv1.groppello_gnu.20240725_082320_8zpnc3/output.yaml
   Frequency: 1800
   frequency_units: nsecs
   ATM_NCPL: 24
 This yields dt_atm=3600.0 > dt_output=1800. Please, adjust 'Frequency' and/or 'frequency_units'
```

Fixes #2891 .

### DIFF
--- a/components/eamxx/cime_config/eamxx_buildnml.py
+++ b/components/eamxx/cime_config/eamxx_buildnml.py
@@ -1017,6 +1017,23 @@ def do_cime_vars_on_yaml_output_files(case, caseroot):
             print ("  - setting skip_t0_output=true\n")
             print ("  - setting freq and freq_units to HIST_N and HIST_OPTION respectively\n")
 
+        # If frequency_units is not nsteps, verify that we don't request
+        # a frequency faster than the model timestep
+        if content['output_control']['frequency_units'] in ['nsecs','nmins','nhours']:
+            freq  = content['output_control']['Frequency']
+            units = content['output_control']['frequency_units']
+            dt_out = 1 if units=="nsecs" else 60 if units=="nmins" else 3600
+            dt_out = dt_out*int(freq)
+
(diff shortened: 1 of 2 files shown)


AaronDonahue:
@bartgol this looks like it hit a number of fails both building on weaver and CIME based tests on `mappy`.  Is it possible our mappy testing all has a frequency < dt and therefore is triggering the new error?

jgfouca (approved):

bartgol:
@AaronDonahue no, the fails were all build errors. Should be fixed now.

PR #2920, opened 2024-07-25
Add support for ne512pg2 bi-grid

wlin7:
Enable ne512pg2_ne512pg2 and be512pg2_ICOS10 grids. 
The latter has ocean and ice on ICOS10 mesh. 
Both use ICOS10 ocean mask.

[BFB] no other grids are affected

### DIFF
--- a/cime_config/config_grids.xml
+++ b/cime_config/config_grids.xml
@@ -1683,6 +1683,26 @@
       <mask>oRRS18to6v3</mask>
     </model_grid>
 
+    <model_grid alias="ne512pg2_ICOS10">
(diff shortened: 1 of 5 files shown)


crterai:
I think we'd want to demonstrate that we can run a ne512 simulation with the changes in this PR before we merge. I'm currently running a test with a change you mentioned above and another change. I can report its pass/fail here when it completes.

AaronDonahue:
@crterai , can you update this PR once your tests have finished?

crterai:
@AaronDonahue The test was successful after applying the latest commit that Wuyin added, so I think it's ready to merge.

PR #2923, opened 2024-07-27
Frontier: Additional post-maintenance updates

dqwu:
This PR is a follow-up to PR #2915, which optionally runs hipInit
prior to MPI_Init to prevent occasional segmentation faults during
MPI_Init (see OLCFDEV-1655).


ambrad (approved):
Nice!

@bartgol do you see anything that isn't safe to merge? I'm not seeing anything.

jayeshkrishna:
I will let the SCREAM devs review the PR but a quick question : Why is the macro used to enable the workaround called SCREAM_SYSTEM_WORKAROUND (Wouldn't naming it to say SCREAM_HIP_INIT_WORKAROUND be better? If the SCREAM_SYSTEM_WORKAROUND is supposed to have different values based on the workaround used, is there a plan to document it?)?

dqwu:
> I will let the SCREAM devs review the PR but a quick question : Why is the macro used to enable the workaround called SCREAM_SYSTEM_WORKAROUND (Wouldn't naming it to say SCREAM_HIP_INIT_WORKAROUND be better? If the SCREAM_SYSTEM_WORKAROUND is supposed to have different values based on the workaround used, is there a plan to document it?)?

@jayeshkrishna This macro name was suggested by @ambrad in PR #2918

> What do you think of adding a more generic flag that we can use in all future instances for this same purpose? I'm thinking something like SCREAM_SYSTEM_WORKAROUND. To make this approach maximally flexible (example in a moment), we'd want to permit setting integer values. Then any place in general E3SM code that requires a workaround would have something like
> 
> #if !defined SCREAM_SYSTEM_WORKAROUND || SCREAM_SYSTEM_WORKAROUND != 1
> ... this code should run in general ...
> #endif
> 
> Once that flag is available, we would be able to use it as needed without messing with the build system again.
> 
> The purpose for the integer complication is as follows. In the future one can imagine needing separate workarounds for Aurora and Frontier. We might then say that Frontier is assigned 1 and Aurora is assigned 2.

Also, @jgfouca has already applied it in this commit: https://github.com/E3SM-Project/scream/pull/2918/commits/ccf8f1eb5fab37caa1d4c4ae95a13f33897d211e

ambrad:
I'm fine with either. I had suggested SCREAM_SYSTEM_WORKAROUND before Jim had done anything; thus, I didn't know it would be as simple as adding the -D flag in one spot to affect all components. Since it is that simple, anything is fine.

cime_config/machines/cmake_macros/crayclang-scream_frontier-scream-gpu.cmake
```diff
@@ -5,7 +5,7 @@ set(SCC "cc")
 set(SCXX "hipcc")
 set(SFC "ftn")
 
-string(APPEND CPPDEFS " -DLINUX")
+string(APPEND CPPDEFS " -DLINUX -DSCREAM_SYSTEM_WORKAROUND=1")
```
bartgol, line 8: I don't think you ever use the case where it's 0, so prob you could just do `-DSCREAM_SYSTEM_WORKAROUND`, and remove `&& (SCREAM_SYSTEM_WORKAROUND == 1)` from all CPP checks.

bartgol (approved):
Looks good to me. I have one minor comment, but feel free to disregard.

cime_config/machines/cmake_macros/crayclang-scream_frontier-scream-gpu.cmake
dqwu: > I don't think you ever use the case where it's 0, so prob you could just do `-DSCREAM_SYSTEM_WORKAROUND`, and remove `&& (SCREAM_SYSTEM_WORKAROUND == 1)` from all CPP checks.

According to @ambrad, the purpose for the integer complication is as follows. In the future one can imagine needing separate workarounds for Aurora and Frontier. We might then say that Frontier is assigned 1 and Aurora is assigned 2.

bartgol: Makes sense, thanks!

ambrad:
@dqwu feel free to merge this when you're ready.

PR #2927, opened 2024-07-30
fix dev/host warnings in shoc

mahf708:
copied logic from haero (h/t @jeff-cohere, xref https://github.com/eagles-project/haero/pull/482)

happy to change "safe_min" to something else. Also, should we get rid of the remaining 5 `std::` calls in this? I thought they would trigger warnings too, but they don't, so I left them untouched

hopefully bfb, but if not, god help us

(also hope I got the right people to review this...)

### DIFF
--- a/components/eamxx/src/physics/shoc/impl/shoc_assumed_pdf_impl.hpp
+++ b/components/eamxx/src/physics/shoc/impl/shoc_assumed_pdf_impl.hpp
@@ -10,6 +10,20 @@ namespace scream {
 namespace shoc {
 
 /*
+Add some functions to avoid cuda compilation warnings
+TODO: move this to ekat or something
+*/
+#ifdef KOKKOS_ENABLE_CUDA
+KOKKOS_INLINE_FUNCTION constexpr Real safe_min() {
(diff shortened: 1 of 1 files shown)


jeff-cohere (approved):
👍 Thanks, Naser!

components/eamxx/src/physics/shoc/impl/shoc_assumed_pdf_impl.hpp
```diff
@@ -9,6 +9,20 @@
 namespace scream {
 namespace shoc {
 
+/*
+Add some functions to avoid cuda compilation warnings
+TODO: move this to ekat or something
+*/
+#ifdef KOKKOS_ENABLE_CUDA
+KOKKOS_INLINE_FUNCTION constexpr Real safe_min() {
+  return Kokkos::Experimental::norm_min_v<Real>;
```
bartgol, line 18: You may as well use this every time...

bartgol (approved):

tcclevenger (approved):

jgfouca (approved):

Issue #1377, opened 2022-01-24
SHOC assumed_pdf routine not converted to CPP as intended


bogensch:
The "shoc_assumed_pdf" routine was not converted from F90 to cpp as intended.  Recall that prior to cpp conversion the F90 code was heavily refactored so that long routines were broken into smaller and testable routines.  However, PR #704 converted the "shoc_assumed_pdf" as one function instead of the paradigm in the F90 code.  The only reason why our property tests are passing is because the SHOC property tests are still being called on the F90 code (see issue #1376) and not cpp.  

Note that it is possible that other SHOC routines were not converted as intended; I only noticed this one as I was making a modification to the assumed_pdf cpp code.  Probably the best way to check to see which routines were converted as intended is to first address PR #1376.

bartgol:
@tcclevenger @bogensch is this still an issue?

PeterCaldwell:
Yes, this is still an issue.

tcclevenger:
All shoc property tests call F90 code, but the F90 code is verified to be BFB with C++. It would be easy to change the property tests to call C++ code I think.

In EAM, `shoc_assumed_pdf()` in shoc.F90 calls a lot of very small functions that are only used once inside `shoc_assumed_pdf()`, so in C++ I just converted them inline to save files/lines of code. Are there additional property checks that we need on those other small functions? C++ should be BFB with the most current F90, so I would expect the property tests run through C++ to pass.

tcclevenger:
I can refactoring shoc_assumed_pdf in C++ to contain the nested functions if that is preferred.

tcclevenger:
`shoc_energy_fixer` is another function that had sub-routines which were not individually ported. Just writing down in case I go back and refactor.

PR #2932, opened 2024-08-01
Port shoc_assumed_pdf functions individually (like in shoc.F90)

tcclevenger:
Closes https://github.com/E3SM-Project/scream/issues/1377. This is needed before all property tests can be switched to C++ from F90.

I didn't make individual BFB tests since I think I would need a work around for randomizing a single pack of data, and the BFBness is tested in `shoc_assumed_bfb`.


components/eamxx/src/physics/shoc/impl/shoc_assumed_pdf_compute_buoyancy_flux_impl.hpp
```diff
@@ -0,0 +1,41 @@
...
+namespace scream {
+namespace shoc {
+
+/*
+ * Implementation of shoc_assumed_pdf_compute_buoyancy_flux. Clients should NOT
+ * #include this file, but include shoc_functions.hpp instead.
+ */
+
+template<typename S, typename D>
+KOKKOS_FUNCTION
```
bartgol, line 17: Given how tiny these functions are, we should probably use KOKKOS_INLINE_FUNCTION. Although the compiler will likely inline them anyways, we don't want to risk them to not be inlined.

tcclevenger:
Investigating CPU fail for baseline tests.

tcclevenger:
@bartgol We get compile time warning about inlining functions without definition on CPU which I believe is because we are interfacing through `eti/func_name.cpp`, so I need to revert.

bartgol:
> @bartgol We get compile time warning about inlining functions without definition on CPU which I believe is because we are interfacing through `eti/func_name.cpp`, so I need to revert.

You mean that there's no available definition, and the code is linked? How does that even work on GPU then, workout the relocatable device code flags?

tcclevenger:
> You mean that there's no available definition, and the code is linked? How does that even work on GPU then, workout the relocatable device code flags?

In `shoc_functions.hpp`, after all the function declarations, we have 
```
#ifdef ENABLE_GPU and not RELOCATABLE_DEVICE
  #include func_name_impl.hpp
#endif
```

bartgol:
Oh, I see. Our ETI strategy is conflicting with the inline declaration. That makes sense. I wonder if ETI makes sense for such small functions...

tcclevenger:
ETI probably doesn't make sense for these. Considered just adding the definitions to the .hpp declaration, but wouldn't want to clutter it up.

bartgol:
> ETI probably doesn't make sense for these. Considered just adding the definitions to the .hpp declaration, but wouldn't want to clutter it up.

You can simply include the impl files, but, unlike other functions, you don't guard with any ifdef, so they always get included. Then, you can mark them inline.

bartgol (approved):

PR #2933, opened 2024-08-01
Remove multiprocess tests for compute sanitizer

tcclevenger:
The multi-process unit tests were quite a strain on the compute sanitizer racecheck. Even without dynamics, they would take around 20-40 min each. I think with the addition of some mam tests, this has made the CSR test start timing out consistently.

A quick solution is to just disable them for compute sanitizer tests. All processes should be touched in the single process tests.

Let me know thoughts. Are we missing out on anything valuable by doing this?

Other options:
- still run a subset of the multi-process tests
- only do this for racecheck (instead of init, sync, and mem, which do not have problems running all these tests)

### DIFF
--- a/components/eamxx/tests/CMakeLists.txt
+++ b/components/eamxx/tests/CMakeLists.txt
@@ -68,9 +68,13 @@ if (NOT DEFINED ENV{SCREAM_FAKE_ONLY})
   # Testing individual atm processes
   add_subdirectory(single-process)
 
-  # Testing multiple atm processes coupled together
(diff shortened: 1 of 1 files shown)


bartgol (approved):

jgfouca (approved):

tcclevenger:
Failing test is a mam baseline test. Shouldn't be from this PR. I'll merge tomorrow unless there are any objections.

tcclevenger:
Baseline test fail not related to this PR. Merging.

PR #2935, opened 2024-08-04
Frontier: use cpe/22.12 in the post-maintenance configuration

dqwu:
The change in the default cpe version (22.12 -> 23.12) may be one
of the main breaking changes for post-maintenance SCREAM builds.

Trey White recommended using cpe/22.12 to get as close as possible
to the pre-maintenance configuration.


ambrad (approved):

dqwu:
@ambrad @trey-ornl 
I have run some ne1024 DECADAL simulations on Frontier using cpe/22.12.

It seems that the workaround calling hipInit prior to MPI_Init is now
optional. However, we can still keep this workaround as it is a known
OLCF issue.

Additionally, the DIRK errors are not reproducible at this time (whether
or not the hipInit workaround is applied). More simulations are needed to
confirm this.

PR #2940, opened 2024-08-06
Update DPxx tests to run with PG2 grid

bogensch:
At long last we are able to run DP mode with PG2.  This PR modifies all DPxx tests so that they use PG2 by default.

Note that I verified that DPxx simulations with PG2 perform scientifically as expected and produce credible results in all the cases I test (about 10).

### DIFF
--- a/cime_config/tests.py
+++ b/cime_config/tests.py
@@ -657,10 +657,10 @@ _TESTS = {
         "time"  : "01:00:00",
         # each test 225 phys cols, roughly size of ne2
         "tests" : (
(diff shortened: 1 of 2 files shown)


bogensch:
Baseline fails expected for all DP tests since the tests are essentially new due to the transition to pg2.

tcclevenger:
Awesome! I'll merge.

tcclevenger (approved):

ambrad:
Note this PR also switch DPxx to use SL transport because that transport method is default for the pg2 grid. That's great!

PR #2943, opened 2024-08-09
Frontier: disable hipInit before MPI_Init

dqwu:
This change reflects that the hipInit workaround is no longer
needed with cpe/22.12. DIRK NaN errors have been observed only
in runs where hipInit was invoked. Disabling hipInit appears
promising, though more production runs are needed to confirm.


trey-ornl (approved):
It seems reasonable to turn off the `hipInit`, since the old environment didn't seem to need it, and future Rocms should not need it.

I do wonder if it would be best to explicitly initialize Kokkos where you currently have the `hipInit`. Seems like that would be portable, should not hurt, and might avoid future issues. It is beyond the scope of the pull request, however.

dqwu:
Compared to the pre-maintenance configuration, the minimal changes we made (after this PR is applied to the SCREAM master) are:

- Included cstdint in scream_scorpio_types.hpp to fix some compile errors.
- Added the GCC 12.2.0 library path to CMAKE_EXE_LINKER_FLAGS to address some linking errors.
- Explicitly used cpe/22.12, which is also used in the pre-maintenance environment.

ambrad (approved):

PR #2945, opened 2024-08-14
update Livermore Computing machine config for eamxx

jsbamboo:
update `/usr/gdata/e3sm` → `/usr/workspace/e3sm` and add dane machine for LC in components/eamxx/


AaronDonahue:
Thanks @jsbamboo !

PR #2950, opened 2024-08-16
move pyscream to nb

mahf708:
The only two tricky parts were the casting (nb does a lot of automatic casting if the correct headers are included) and arrays (for now, I am hardcoding returning double arrays)

### DIFF
--- a/components/eamxx/src/python/libpyscream/pyfield.hpp
+++ b/components/eamxx/src/python/libpyscream/pyfield.hpp
@@ -4,9 +4,11 @@
 #include "share/field/field.hpp"
 #include "share/field/field_utils.hpp"
 
-#include <pybind11/pybind11.h>
-#include <pybind11/numpy.h>
-#include <pybind11/stl.h>
+#include <nanobind/nanobind.h>
+#include <nanobind/ndarray.h>
+#include <nanobind/stl/list.h>
+
+namespace nb = nanobind;
 
 namespace scream {
 
@@ -24,7 +26,8 @@ struct PyField {
     f.allocate_view();
   }
 
-  pybind11::array get () const {
+  template <typename FRAMEWORK>
+  nb::ndarray<FRAMEWORK> get () const {
     const auto& fh  = f.get_header();
     const auto& fid = fh.get_identifier();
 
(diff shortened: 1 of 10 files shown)


components/eamxx/src/python/libpyscream/pyfield.hpp
```diff
@@ -24,7 +26,9 @@ struct PyField {
     f.allocate_view();
   }
 
-  pybind11::array get () const {
+  // FIXME: how to return ndarray?
+  template <typename X, typename Y, typename Z>
```
bartgol, line 30: This seems obscure. There's no way to understand what X/Y/Z mean here, and we only use one template overload anyways. Nanobind (like pybind) does not require to specify any template argument for ndarray. While template args may provide compile/runtime benefits (like type checking and certain compiler optimizations), given our use case, I would advocate for usable and generic code first (we can optimize perf later).

I think you can create an ndarray without template args, and pass data type, shape, and framework as constructor arguments.

components/eamxx/src/python/libpyscream/pyfield.hpp
```diff
@@ -24,7 +26,9 @@ struct PyField {
...
+  nb::ndarray<X,Y,Z> get () const {
```
bartgol, line 31: I would rename this as `array_host` or something. Better to be explicit on the fact that it's host data, so the user remembers to sync later. Also, we may later add `array_dev`, to allow grabbing dev pointers directly.

components/eamxx/src/python/libpyscream/pyfield.hpp
```diff
@@ -59,8 +69,9 @@ struct PyField {
 
     // NOTE: you MUST set the parent handle, or else you won't have view semantic
     auto data = f.get_internal_view_data_unsafe<void,Host>();
-    auto this_obj = pybind11::cast(this);
-    return pybind11::array(dt,shape,strides,data,pybind11::handle(this_obj));
+    auto this_obj = nb::cast(this);
+    // pybind11::array(dt,shape,strides,data,pybind11::handle(this_obj));
+    return nb::ndarray<X,Y,Z>(data, shape_t, shape, nb::handle(this_obj));
```
bartgol, line 74: Related to template arg comments: I think it's fine (for now) to hard-code X to be `nb::numpy`. We can decide how to allow getting different backends later, if needed. The other two args can be removed altogether, I think.

bartgol (commented):
Looks relatively seamless. I think the ndarray issue can be fixed easily.

bartgol (approved):

PR #2955, opened 2024-08-19
Limit compute-sanitizer racecheck resources

tcclevenger:
`compute-sanitizer --tool=racecheck` was attempting to use the entire node worth of threads, which is an issue in a parallel `test-all-scream` build, since all other tests would start to hang. This was causing the non-report from compute-sanitizer nightlies.

Solution: if we are in a parallel build, limit racecheck to using `compile_res_count` threads (approx `total_cpu_threads/num_parallel_tests`).

Running exact `gather-all-scream` command as the compute-sanitizer nightlies took ~70 min which should be plenty fast enough to report.

### DIFF
--- a/components/eamxx/scripts/test_all_scream.py
+++ b/components/eamxx/scripts/test_all_scream.py
@@ -2,7 +2,7 @@ from utils import run_cmd, run_cmd_no_fail, expect, check_minimum_python_version
     SharedArea, safe_copy
 from git_utils import get_current_head, get_current_commit
 
-from test_factory import create_tests, COV
+from test_factory import create_tests, COV, CSR
 
 from machines_specs import get_mach_compilation_resources, get_mach_testing_resources, \
     get_mach_baseline_root_dir, setup_mach_env, is_cuda_machine, \
@@ -385,6 +385,15 @@ class TestAllScream(object):
 
         if "SCREAM_DYNAMICS_DYCORE" not in custom_opts_keys:
             result += " -DSCREAM_DYNAMICS_DYCORE=HOMME"
+
+        # For the compute-sanitizer tool 'racecheck', if no option --racecheck-num-workers
+        # is provided, it will attempt to use all threads available on node. This can cause
+        # issues if other test cases are being run in parallel. If the option was not specified,
+        # limit the number of threads availible to racecheck to the number of compile resources.
+        if self._parallel and isinstance(test, CSR) and not '--racecheck-num-workers' in result:
+            new_option = ' --racecheck-num-workers=' + str(test.compile_res_count)
+            index = result.index('--tool=racecheck') + len('--tool=racecheck')
+            result = result[:index] + new_option + result[index:]
 
         return result
 
--- a/components/eamxx/scripts/test_factory.py
+++ b/components/eamxx/scripts/test_factory.py
@@ -176,7 +176,7 @@ class CSM(TestProperty):
             "debug with compute sanitizer memcheck",
             [("CMAKE_BUILD_TYPE", "Debug"),
              ("EKAT_ENABLE_COMPUTE_SANITIZER", "True"),
-             ("EKAT_COMPUTE_SANITIZER_OPTIONS", "--tool=memcheck")],
+             ("EKAT_COMPUTE_SANITIZER_OPTIONS", "'--tool=memcheck'")],
             uses_baselines=False,
             on_by_default=False,
             default_test_len="short"
@@ -210,7 +210,7 @@ class CSI(TestProperty):
             "debug with compute sanitizer initcheck",
             [("CMAKE_BUILD_TYPE", "Debug"),
              ("EKAT_ENABLE_COMPUTE_SANITIZER", "True"),
-             ("EKAT_COMPUTE_SANITIZER_OPTIONS", "--tool=initcheck")],
+             ("EKAT_COMPUTE_SANITIZER_OPTIONS", "'--tool=initcheck'")],
             uses_baselines=False,
             on_by_default=False,
             default_test_len="short"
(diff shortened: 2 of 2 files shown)


tcclevenger:
This change doesn't affect the AT, but I'll still run AT since this does add code to `test_all_scream.py` that could break things.

components/eamxx/scripts/test_factory.py
```diff
@@ -176,7 +176,7 @@ def __init__(self, _):
             "debug with compute sanitizer memcheck",
             [("CMAKE_BUILD_TYPE", "Debug"),
              ("EKAT_ENABLE_COMPUTE_SANITIZER", "True"),
-             ("EKAT_COMPUTE_SANITIZER_OPTIONS", "--tool=memcheck")],
+             ("EKAT_COMPUTE_SANITIZER_OPTIONS", "'--tool=memcheck'")],
```
tcclevenger, line 179: These changes are unrelated, but if we ever need to add more options to these tests, not having these quotes might lead to hard to find errors. But I can revert this commit if anyone wants.

jgfouca (approved):
Nice find! Weird that race check would use threads beyond the one GPU that it gets allocated?

jgfouca (changes requested):
Actually, now I'm more confused. `compile_res_count` is a build resource tracker, not a run resource tracker. What I thought was happening on GPU is that each test gets it's own GPU to itself.

tcclevenger:
> Nice find! Weird that race check would use threads beyond the one GPU that it gets allocated?

Yeah, I agree. Unfortunately there isn't much documentation on this option.

tcclevenger:
> Actually, now I'm more confused. `compile_res_count` is a build resource tracker, not a run resource tracker.

`compile_res_count` comes from `get_available_cpu_count()`, which gives all the threads available. This is the resource that compute sanitizer is allocating. I could instead recompute the option from `get_available_cpu_count()` which would be more obvious to the person reading `test_all_scream.py`, but the value I want is already in `compile_res_count`.

tcclevenger:
The only documentation on this options is 
```
racecheck-num-workers | {number} | 0 | Number of CPU worker threads used by the tool. Use 0 for automatic.

```
found https://docs.nvidia.com/compute-sanitizer/ComputeSanitizer/index.html.

It appears that the racecheck tool is trying to use all the CPU threads on the entire node when set to "automatic". Basically a large test (`wsm_omp1` from ekat) starts and then every other test cannot complete whatever current test is launch because racecheck is using every thread. Because the test lasts so long, all other tests (synccheck, memcheck, initcheck) time out, and then only racecheck completes the remainder of it's tests.

tcclevenger:
Another solution is the run `-t csr` alone, and then `-t css -t csi -t csm -p`.

jgfouca (approved):

jgfouca:
OK, thanks. This all makes more sense now. I think the current solution is better than having to remember to run racecheck separately.

components/eamxx/scripts/test_all_scream.py
```diff
@@ -386,6 +386,15 @@ def generate_cmake_config(self, test, for_ctest=False):
         if "SCREAM_DYNAMICS_DYCORE" not in custom_opts_keys:
             result += " -DSCREAM_DYNAMICS_DYCORE=HOMME"
 
+        # For the compute-sanitizer tool 'racecheck', if no option --racecheck-num-workers
+        # is provided, it will attempt to use all threads available on node. This can cause
+        # issues if other test cases are being run in parallel. If the option was not specified,
+        # limit the number of threads availible to racecheck to the number of compile resources.
+        if self._parallel and isinstance(test, CSR) and not '--racecheck-num-workers' in result:
+            new_option = ' --racecheck-num-workers=' + str(test.compile_res_count)
+            index = result.index('--tool=racecheck') + len('--tool=racecheck')
```
bartgol, line 395: I see this in test_factory.py:

```python
    ("EKAT_COMPUTE_SANITIZER_OPTIONS", "'--tool=racecheck --racecheck-detect-level=error'")],
```
So I am thinking that this call to `index` may fail... Why not sticking with the version without `'`, and just look for `--tool-racecheck`?

components/eamxx/scripts/test_factory.py
bartgol: See my comment above. I think something is not right.

bartgol (commented):
According to SNL AI bot, we could even make the sanitizer pin its threads to specific cores, by combining it with taskset:
```
If you want to run compute-sanitizer with 4 worker threads on CPU cores 2, 3, 4, and 5, you would use:

taskset -c 2,3,4,5 compute-sanitizer --tool=racecheck --racecheck-num-workers=4 ./your_cuda_application
```
I don't think we should do it, but I think it may help if the CSR build is not "the first" in the bld list for test-all-scream. The idea is that with all 4 CS* builds going, we want the N cores available to CSR to be the same that CSR is using to build. Maybe we can achieve this by simply reordering the builds in jenkins_common_impl.sh?
It's prob not that important though, since you were able to bring down exec time by quite a bit. But I think a single sanitizer build takes less than 70min, so I'm thinking that there's some resource contention with other CS builds...

components/eamxx/scripts/test_all_scream.py
tcclevenger: I think python `str1.index(str2)` takes in a strings, no? That's why I added the `''`, so I'm looking for --tool=racecheck. But my python is rusty so I may misunderstand your question..

tcclevenger: Unless you mean the `'--tool=racecheck --racecheck-detect-level=error'`. If you don't include the outer quotes, cmake only picks up `EKAT_COMPUTE_SANITIZER_OPTIONS=--tool=racecheck` and you get an error
```
CMake Error: Unknown argument --racecheck-detect-level=error
```

bartgol: Ah, sorry, I mixed up the quotes looking at this and the mods in test_factory.py. Yes, this is fine.

tcclevenger:
@bartgol Removing the automerge label. I'll test your suggestions (moving `-t csr` first). 

Also, I will test a little with taskset. The other sanitizer tests still run slower than if they are run without `csr` (but consistently do not timeout). I think maybe the threads are still being taken over. I wonder if we could save a lot of time by pinning threads for each test.

bartgol:
> Also, I will test a little with taskset. The other sanitizer tests still run slower than if they are run without `csr` (but consistently do not timeout). I think maybe the threads are still being taken over. I wonder if we could save a lot of time by pinning threads for each test.

It's possible. But I'd say, don't sweat about it.

tcclevenger:
Same issues with resource contention with putting `-t csr` first. I'll just let this merge as is and maybe test taskset at some point. I am very curious..

Issue #1376, opened 2022-01-24
SHOC property tests are still calling F90 code


bogensch:
It appears that most (if not all... I haven't examined every test file) SHOC property tests are still calling the F90 code and not the cpp implementation.  This should be modified before the F90 code is abandoned to ensure property tests are being called on the relevant cpp code.

bartgol:
@tcclevenger is this still the case?

tcclevenger:
@bartgol Looks that way, we just need to transpose the input data and call the `shoc_*_f()` function. Should be easy. I can find some time this week to put a PR together if you want.

bartgol:
I don't think it's a priority. I was just going through our issues to see what v1 dev issues were still a thing, trying to find some suitable ones for a new eamxx dev.

PeterCaldwell:
Agreed this isn't urgent, but it will mean that once DP can be run in C++ mode and we get rid of F90 support, all of our SHOC testing will disappear unless we fix this. So worth keeping open and doing, but not worth putting ahead of anything else we're already doing or is urgent.

AaronDonahue:
@jgfouca , I suspect that your upcoming work on dropping F90 support for SHOC will also address this issue.  Can you tag this as fixed when that PR goes in?

jgfouca:
@AaronDonahue , yep! I am addressing this exact issue.

jgfouca:
Fixed.

PR #2956, opened 2024-08-19
Use C++ in shoc property tests

tcclevenger:
Call C++ interface functions in shoc property tests. Partly addresses https://github.com/E3SM-Project/scream/issues/1376.

I'm not sure the reason the functions calling C++ code (e.g., `shoc_assumed_pdf_f()`) expect a fortran data layout, and the functions calling F90 (e.g., shoc_assumed_pdf(ShocAssumedPDFData)`) expect C layout. But that can be changed whenever we drop F90.

### DIFF
--- a/components/eamxx/src/physics/shoc/tests/shoc_assumed_pdf_tests.cpp
+++ b/components/eamxx/src/physics/shoc/tests/shoc_assumed_pdf_tests.cpp
@@ -102,8 +102,6 @@ struct UnitWrap::UnitTest<D>::TestShocAssumedPdf {
     }
 
     // Check that the inputs make sense
-
-    // Load input data
     for(Int s = 0; s < shcol; ++s) {
       for(Int n = 0; n < nlev; ++n) {
         const auto offset = n + s * nlev;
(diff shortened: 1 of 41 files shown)


bartgol:
> I'm not sure the reason the functions calling C++ code (e.g., `shoc_assumed_pdf_f()`) expect a fortran data layout, and the functions calling F90 (e.g., shoc_assumed_pdf(ShocAssumedPDFData)`) expect C layout. But that can be changed whenever we drop F90.

My guess: the fcn to call C++ code are expected to be used from F90, and as such they are expected to have the data still in F90 format. And viceversa. That said, it does look confusing.

bartgol:
Not to wash my hands of it, but I think Jim is the best person to review stuff that has to do with physics pkg CXX-F90 testing, since he implemented some of the infrastructure for it back in the P3 refactor days. I get a bit lost in that c vs s ping-pong of data...

jgfouca (dismissed):
LGTM

tcclevenger:
@jgfouca the diffs on the check that I changed were at `1e-23`. I'm guessing there is no other way to avoid checking against a tolerance.

jgfouca (approved):

PR #2969, opened 2024-08-26
Move Frontier build to gfortran and latest rocm

trey-ornl:
Added new `cime_config/machines/cmake_macros/craygnu-hipcc.cmake` file, which uses the Cray compiler wrappers with Gnu compilers, along with `hipcc` for AMD GPUs.
Changed Adios2 path to an appropriate OLCF build.
Deleted all Crusher files. Crusher no longer exists.


trey-ornl:
I used this branch, before the last merge with `master`, to try to reproduce the E3SM failure on the "bad node" of Frontier, which now has the alias `borg106`. The run did fail, while runs on other nodes completed without failing. Diffs of the `e3sm.log.*` files eventually varied in their hashes, however, typically starting with `exxhash` around line 569. Here are the runs:
```
/lustre/orion/world-shared/cli115/trey/olcfhelp-14845/case/2024-08-*
```
I `gunzip`-ed the `e3sm.log` files for the successful runs.

trey-ornl:
A two-node run of `ne30` ran to completion on Frontier with this branch before the final merge with `master`.
```
/lustre/orion/world-shared/cli115/trey/scream-ne30-2node/case
```

ambrad:
Am I correct that the current s/w stack can't be run from this branch?

trey-ornl:
> Am I correct that the current s/w stack can't be run from this branch?

Yes, as written, it replaces the current stack. Shall I modify it so that the previous stack can still be selected? It could key off of the compiler choice.

ambrad:
Tagging @rljacob @ndkeen @jgfouca @PeterCaldwell @AaronDonahue.

Trey, I don't know the right answers to your questions in terms of what everyone wants. But in terms of technical steps:
1. We can give machines/compiler entries uniquifying names to permit parallel stacks.
2. The CMakeLists.txt change could probably be handled with slightly better conditionals. In particular, I think querying the Fortran compiler ID might help, e.g., something based on `CMAKE_Fortran_COMPILER_ID MATCHES "gfortran"`.

ambrad:
I marked this PR as WIP so that the autotester doesn't run on it until it's ready.

ambrad:
@trey-ornl the P3 deoptimization is working. Thus, in your performance testing, if you haven't started already, I suggest adding this:
```
diff --git a/cime_config/machines/cmake_macros/craygnu-hipcc.cmake b/cime_config/machines/cmake_macros/craygnu-hipcc.                                                                                                                        
cmake                                                                                                                                                                                                                                        
index 8322c5d3a9..6cb79c0146 100644                                                                                                                                                                                                          
--- a/cime_config/machines/cmake_macros/craygnu-hipcc.cmake                                                                                                                                                                                  
+++ b/cime_config/machines/cmake_macros/craygnu-hipcc.cmake                                                                                                                                                                                  
@@ -5,7 +5,7 @@ set(SCC "cc")                                                                                                                                                                                                                
 set(SCXX "hipcc")                                                                                                                                                                                                                           
 set(SFC "ftn")                                                                                                                                                                                                                              
                                                                                                                                                                                                                                             
-string(APPEND CPPDEFS " -DLINUX -DFORTRANUNDERSCORE -DNO_R16 -DCPRGNU")                                                                                                                                                                     
+string(APPEND CPPDEFS " -DLINUX -DFORTRANUNDERSCORE -DNO_R16 -DCPRGNU -DSCREAM_SYSTEM_WORKAROUND_P3_PART2")                                                                                                                                 
 if (COMP_NAME STREQUAL gptl)                                                                                                                                                                                                                
     string(APPEND CPPDEFS " -DHAVE_NANOTIME -DBIT64 -DHAVE_SLASHPROC -DHAVE_COMM_F2C -DHAVE_TIMES -DHAVE_GETTIMEOFDA                                                                                                                        
Y")                                                                                                                                                                                                                                          
 endif()                                                                                                                                                                                                                                     
diff --git a/components/eamxx/src/physics/p3/disp/p3_main_impl_part2_disp.cpp b/components/eamxx/src/physics/p3/disp/                                                                                                                        
p3_main_impl_part2_disp.cpp                                                                                                                                                                                                                  
index 2b619d54bf..28445a35c3 100644                                                                                                                                                                                                          
--- a/components/eamxx/src/physics/p3/disp/p3_main_impl_part2_disp.cpp                                                                                                                                                                       
+++ b/components/eamxx/src/physics/p3/disp/p3_main_impl_part2_disp.cpp                                                                                                                                                                       
@@ -9,7 +9,9 @@ namespace p3 {                                                                                                                                                                                                               
  * Implementation of p3 main function. Clients should NOT #include                                                                                                                                                                          
  * this file, #include p3_functions.hpp instead.                                                                                                                                                                                            
  */                                                                                                                                                                                                                                         
-                                                                                                                                                                                                                                            
+#ifdef SCREAM_SYSTEM_WORKAROUND_P3_PART2                                                                                                                                                                                                    
+# pragma clang optimize off                                                                                                                                                                                                                 
+#endif                                                                                                                                                                                                                                      
 template <>                                                                                                                                                                                                                                 
 void Functions<Real,DefaultDevice>                                                                                                                                                                                                          
 ::p3_main_part2_disp(                                                                                                                                                                                                                       
@@ -130,7 +132,9 @@ void Functions<Real,DefaultDevice>                                                                                                                                                                                       
     if (!hydrometeorsPresent(i)) return;                                                                                                                                                                                                    
   });                                                                                                                                                                                                                                       
 }                                                                                                                                                                                                                                           
-                                                                                                                                                                                                                                            
+#ifdef SCREAM_SYSTEM_WORKAROUND_P3_PART2                                                                                                                                                                                                    
+# pragma clang optimize on                                                                                                                                                                                                                  
+#endif                                                                                                                                                                                                                                      
 } // namespace p3                                                                                                                                                                                                                           
 } // namespace scream
```

rljacob:
@trey-ornl you should key off the compiler choice but we want to deprecate all machine entries with "scream" in the title.  Maybe now would be a good time to do that.

rljacob:
Also I think the E3SM convention for this compiler option name is "craygnugpu" .  We have not put the vendor-specific GPU language in the names.  Not sure if we should.

ndkeen:
Why would the compiler name not be just gnugpu?

trey-ornl:
> Also I think the E3SM convention for this compiler option name is "craygnugpu" . We have not put the vendor-specific GPU language in the names. Not sure if we should.

"craygnugpu" says to me that the Cray wrappers will use the Gnu compilers to offload to GPUs. That is not the case here. The intended message for the name to convey is this: "Use Cray wrappers with Gnu compilers for the host, and use the AMD Hip compiler for the GPU (currently named 'hipcc')." Maybe "craygnuamdpgu" or "craygnu-amdgpu"?

trey-ornl:
> Why would the compiler name not be just gnugpu?

The name "gnugpu" says to me that the Gnu compilers are used directly ("g++", "gfortran"), and that they are used to compile for the GPUs. Neither is true in this case.

rljacob:
@trey-ornl good point about the GPU compile.  I guess craygnuamdpgu is what we should use.

Yes "craygnu" implies cray wrappers while "gnu" is just calling gfortran directly. 

Ignoring ones with "scream", we currently have:
amdclang_frontier.cmake				crayclang_frontier.cmake
amdclanggpu_frontier.cmake			crayclanggpu_frontier.cmake
gnu_frontier.cmake		gnugpu_frontier.cmake

So we'd be adding "craygnuamdgpu_frontier.cmake"

rljacob:
Ignore my comment about editing the "frontier" entry.  We'll do that later.

trey-ornl:
@ambrad, are the tests in `components/eamxx/src/physics/p3/tests` useful? Do you know if they detect the issue in `p3_main_part2`?

ambrad:
I don't use those tests for nondeterminism analysis and have not run them in years. I consider a single-node ne30 ERS test to be the most useful test configuration.

rljacob:
Why did you merge master in to this feature branch?  We discourage that unless you have a specific reason.

trey-ornl:
> Why did you merge master in to this feature branch? We discourage that unless you have a specific reason.

Newbie mistake. How do you test your branch with all the latest changes without merging them in first? Is the expectation that we only test our changes relative to where the branch started?

trey-ornl:
> Why did you merge master in to this feature branch? We discourage that unless you have a specific reason.

Ugh, now the build in this branch is broken. Looks like it has something to do with the updated external `components/elm/src/external_models/fates`. I am reaping what I sowed!

rljacob:
To answer your question, I think CI does a merge to master and tests that.  You could also do a merge to master in your local copy, test, then reset.

bartgol:
If you are collaboratively working with someone else, it's even better to use `git push --force-with-lease`.

From the docs:
>--force-with-lease alone, without specifying the details, will protect all remote refs that are going to be updated by requiring their current value to be the same as the remote-tracking branch we have for them.
So if someone else pushed something to the remote branch in the meantime, your force push will fail.

AaronDonahue:
@trey-ornl , is this PR ready to be "un-WIP'ed" and tested?

AaronDonahue:
@trey-ornl , would ou mind manually testing on `frontier` with `master` merged in to make sure everything still works as expected?  Once you report back the all-clear we can merge.

trey-ornl:
> @trey-ornl , would ou mind manually testing on `frontier` with `master` merged in to make sure everything still works as expected? Once you report back the all-clear we can merge.

Just built and ran an ne30 performance test from @ndkeen yesterday on Frontier with these changes merged into master. New build (compiler `craygnuamdgpu`) ran 3% faster than the old (compiler `crayclang-scream`).

ambrad:
Looks good. Visual inspection shows that the current build (crayclang-scream) is available again after the most recent commits to this PR. Trey, have you checked that crayclang-scream in this PR reproduces crayclang-scream on master? One way to check would be to diff the `.env_mach_specific.sh` files from branch and master configurations. If they're the same, I'm ready to approve. Thanks for adding a GNU configuration; that will be great to have and hopefully make default in future simulation campaigns.

trey-ornl:
I confirmed with the decadal run on Frontier that `master` and this branch using `crayclang-scream` generate identical `.env-mach-specific.sh` files.

ambrad (approved):

AaronDonahue:
I'll go ahead and manually merge now that we have approval and confirmation that everything works as expected on frontier.  Thanks for putting this in @trey-ornl !

PR #2971, opened 2024-08-26
Add libfabric/1.15.2.0 module to frontier-scream-gpu.

trey-ornl:
The default version of module `libfabric` on Frontier changed from `1.15.2.0` to `1.20.1`, and this caused hangs in the decadal run. I confirmed that the run hangs even if compiled with `libfabric/1.20.1` when using the current `cpe/22.12`-based software environment, but it runs to completion with the module `libfabric/1.15.2.0`.

I also combined the two `<module>` blocks into one, since they both actually depend on the setting `compiler="crayclang-scream"`.


ambrad (approved):

ambrad:
This is fine to merge with the current AT results.

PR #2972, opened 2024-08-27
Pm-cpu: do static libc++ linking

jgfouca:
There are several libraries we use that were built with gcc-11:

```
    <environment_variables compiler="gnu" mpilib="mpich">
      <env name="ADIOS2_ROOT">$SHELL{if [ -z "$ADIOS2_ROOT" ]; then echo /global/cfs/cdirs/e3sm/3rdparty/adios2/2.9.1/cray-mpich-8.1.25/gcc-11.2.0; else echo "$ADIOS2_ROOT"; fi}</env>
      <env name="BLA_VENDOR">Generic</env>
      <env name="Albany_ROOT">$SHELL{if [ -z "$Albany_ROOT" ]; then echo /global/common/software/e3sm/albany/2024.03.26/gcc/11.2.0; else echo "$Albany_ROOT"; fi}</env>
      <env name="Trilinos_ROOT">$SHELL{if [ -z "$Trilinos_ROOT" ]; then echo /global/common/software/e3sm/trilinos/15.1.1/gcc/11.2.0; else echo "$Trilinos_ROOT"; fi}</env>
    </environment_variables>
```

This PR fixes the error:
```
 92: /pscratch/sd/a/acmetest/e3sm_scratch/pm-cpu/ERS_Ln22.ne30_ne30.F2010-SCREAMv1.pm-cpu_gnu.scream-internal_diagnostics_level--scream-output-preset-3.20240826_132116_eilkpw/bld/e3sm.exe: /opt/cray/pe/gcc-libs/libstdc++.so\
.6: version `GLIBCXX_3.4.30' not found (required by /pscratch/sd/a/acmetest/e3sm_scratch/pm-cpu/ERS_Ln22.ne30_ne30.F2010-SCREAMv1.pm-cpu_gnu.scream-internal_diagnostics_level--scream-output-preset-3.20240826_132116_eilkpw/b\
ld/e3sm.exe)
```

... which started happening after the upstream merge, which changed the pm-cpu compiler to 12.2

### DIFF
--- a/cime_config/machines/cmake_macros/gnu_pm-cpu.cmake
+++ b/cime_config/machines/cmake_macros/gnu_pm-cpu.cmake
@@ -10,3 +10,6 @@ set(MPIFC "ftn")
 set(SCC "gcc")
 set(SCXX "g++")
 set(SFC "gfortran")
+
+string(APPEND CMAKE_EXE_LINKER_FLAGS " -static-libstdc++")
+
--- a/cime_config/machines/config_machines.xml
+++ b/cime_config/machines/config_machines.xml
@@ -218,7 +218,7 @@
 
       <modules compiler="gnu">
         <command name="load">PrgEnv-gnu/8.5.0</command>
-        <command name="load">gcc/12.2.0</command>
+        <command name="load">gcc/11.2.0</command>
         <command name="load">cray-libsci/23.02.1.1</command>
       </modules>
 
--- a/cime_config/machines/config_machines.xml
+++ b/cime_config/machines/config_machines.xml
@@ -218,7 +218,7 @@
 
       <modules compiler="gnu">
         <command name="load">PrgEnv-gnu/8.5.0</command>
-        <command name="load">gcc/11.2.0</command>
+        <command name="load">gcc/12.2.0</command>
         <command name="load">cray-libsci/23.02.1.1</command>
       </modules>
 


ndkeen:
No we moved to gcc 12 now.
If the adios libs need to rebuilt, we can do that.
We can also comment out that section for now as things will still work without adios

jgfouca:
@bartgol , I was wondering that too. I guessed that they aren't linking adios2.

jgfouca:
OK, according to `ldd` the only library that scream is loading that e3sm is not is:
```
	libyaml-cpp.so.0.6 => /usr/lib64/libyaml-cpp.so.0.6
```

jgfouca:
Some updates. I did a comparison of `ldd` between an e3sm.exe from a non-eamxx case that worked and an eamxx case that did not work. They were basically identical with the exception of yaml-cpp, which the eamxx case linked against a system install one. I thought maybe this was the problem, so I built a new yaml-cpp with gcc-12 and made sure to use that one instead. I was surprised to see that did not fix the issue. With no more leads, I decided to statically link libstdc++ and that did fix the problem.

bartgol:
> EKAT was bringing it in, as expected. Changing it to a freshly built yaml-cpp did not fix the issue.

Are you saying that EKAT_SKIP_FIND_YAML_CPP still caused ekat to find yaml instead of locally building it? Or it built locally, but then still linked the system one?

jgfouca:
I built my own yaml-cpp with gcc-12 and then set YAML_CPP_ROOT. It linked mine and I no longer saw the system yaml-cpp in the ldd output.

bartgol:
Oh. So the system yamlcpp was _not_ the responsible one, since you still got the GLIBXX error with your custom yaml. Interesting.

jgfouca:
Hey, @ndkeen , @bartgol , can I merge this?

ndkeen:
I thought we were trying to go away from static linking? Do we know why this is needed in scream, but not e3sm? Was it because adios used differently?

jgfouca:
> I thought we were trying to go away from static linking?

This is news to me. We use .a libraries all over our build system.

> Do we know why this is needed in scream, but not e3sm?

No, it's a mystery at the moment. `ldd` output showed e3sm and scream loading the same libraries.

> Was it because adios used differently?

I don't think so. The adios stuff was all linking statically.

ndkeen:
I chatted with Jim and I don't want to hold up the larger merge. We can experiment later.

I know we have .a's, but I thought the final e3sm.exe was dynamically linked. nbd

PR #2975, opened 2024-08-28
SHOC related diagnostic output

bogensch:
Adds the option to output several SHOC related variables.

### DIFF
--- a/components/eamxx/src/physics/shoc/eamxx_shoc_process_interface.cpp
+++ b/components/eamxx/src/physics/shoc/eamxx_shoc_process_interface.cpp
@@ -91,6 +91,20 @@ void SHOCMacrophysics::set_grids(const std::shared_ptr<const GridsManager> grids
   add_field<Computed>("ustar",            scalar2d,     m/s,         grid_name, ps);
   add_field<Computed>("obklen",           scalar2d,     m,           grid_name, ps);
 
+  // Diagnostic output - mid point grid
+  add_field<Computed>("brunt", scalar3d_mid, pow(s,-1), grid_name, ps);
+  add_field<Computed>("shoc_mix", scalar3d_mid, m, grid_name, ps);
+  add_field<Computed>("isotropy", scalar3d_mid, s, grid_name, ps);
+
+  // Diagnostic output - interface grid
+  add_field<Computed>("wthl_sec", scalar3d_int, K*(m/s), grid_name, ps);
+  add_field<Computed>("thl_sec", scalar3d_int, pow(K,2), grid_name, ps);
+  add_field<Computed>("wqw_sec", scalar3d_int, (kg/kg)*(m/s), grid_name, ps);
+  add_field<Computed>("qw_sec", scalar3d_int, pow(kg/kg,2), grid_name, ps);
+  add_field<Computed>("uw_sec", scalar3d_int, pow(m/s,2), grid_name, ps);
+  add_field<Computed>("vw_sec", scalar3d_int, pow(m/s,2), grid_name, ps);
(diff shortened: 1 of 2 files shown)


components/eamxx/src/physics/shoc/eamxx_shoc_process_interface.cpp
```diff
@@ -91,6 +91,25 @@ void SHOCMacrophysics::set_grids(const std::shared_ptr<const GridsManager> grids
   add_field<Computed>("ustar",            scalar2d,     m/s,         grid_name, ps);
   add_field<Computed>("obklen",           scalar2d,     m,           grid_name, ps);
 
+  // Extra SHOC output diagnostics
+  if (m_params.get<bool>("extra_shoc_diags", false)) {
```
bartgol, line 95: While this option makes sense to avoid polluting the FM with stuff we don't need, it has the drawback that users must atmchange this whenever they add one of these fields to their output yaml files.

I suppose this is the best we can do for now. Hopefully this will become obsolete once we rework a bit the IO layer for optional outputs.

components/eamxx/src/physics/shoc/eamxx_shoc_process_interface.cpp
```diff
@@ -495,6 +514,41 @@ void SHOCMacrophysics::run_impl (const double dt)
                        default_policy,
                        shoc_postprocess);
   Kokkos::fence();
+
+  // Extra SHOC output diagnostics
+  if (m_params.get<bool>("extra_shoc_diags", false)) {
+
+    const auto& shoc_mix = get_field_out("shoc_mix").get_view<Spack**>();
+    Kokkos::deep_copy(shoc_mix,history_output.shoc_mix);
```
bartgol, line 522: This is indeed costly, for no reason. In principle, we would like history_output.shoc_mix to simply use the view stored inside the Field "shoc_mix". We are basically adding 10 3d fields for no real reason...

OTOH, this may add more complexity, since we need to be careful during init, depending on the value of `extra_shoc_diags`:
- false: request enough buffer mem to accommodate these vars.
- true: do NOT request buffer mem for these fields, and grab them from the computed field during initialize_impl

tcclevenger: The buffer needs a bit of a reworked to allow for it to be dynamically sized based on a param and not be unreadable. @bogensch If you do not want to do that I can take a look. But at the very least we can get rid of the deep copies like Luca said, in initialize_impl:
```
if(m_params.get<bool>("extra_shoc_diags")) {
  history_output.shoc_mix = get_field_out("shoc_mix").get_view<Spack**>();
} else {
  history_output.shoc_mix = m_buffer.shoc_mix;
}
```

bartgol: Not too much logic to add. When SHOCMacrophysics::requested_buffer_size_in_bytes and SHOCMacrophysics::init_buffers are called, we can ping m_params, to see if we need the extra mem for those fields or not.

tcclevenger: You're right, it's not too much logic. But we use a `constexpr` for number of fields in the buffer, and then we initialize the buffer views in a way that probably doesn't lend itself to using if-else nicely.

bartgol: @tcclevenger what's your plan regarding these unnecessary copies? Are you planning to fix them here or in another PR? If the latter, you can go ahead and merge, since the cosp failures are orthogonal.

bartgol (approved):

